PR #1544, opened 2016-05-02
Move one level up the sections in the configuration file

monicasarbu:
Based on the changed done in #1490, I am proposing the following structure of the configuration file where I move all the sections one level up all. 

Implements part of https://github.com/elastic/beats/issues/1417.

The changes should affect the following configuration files:

- [x] libbeat.yml
- [x] packetbeat.yml
- [x] topbeat.yml
- [x] filebeat.yml
- [x] winlogbeat.yml

Issue #1417, opened 2016-04-19
Configuration refactoring


tsg:
This is a meta issue for reorganizing our default configuration files for 5.0. The problems we're trying to solve are:

* Getting whitespace wrong in YAML is our number one source of issues on discuss. Now the Beats supports dotted notation (e.g. `output.elasticsearch.hosts`) in the configuration file, which greatly reduces the indentation levels required.
* Inconsistent namespacing around Beats. Filebeat has all its options under `filebeat`, Topbeat has/had `input`, Packetbeat has several options toplevel. Correct namespacing is a prerequisite for building more Beats in the same binary/process.
* The Libbeat options are not well organized, we have some stuff top level and some other stuff under `beat`.
* Because we put all available options in the configuration file, the default files have gotten really large and hard to modify for the common options.

TODOs:

* [x] Rename the `input` section to `topbeat` in Topbeat. #1211.
* [X] Move top level Packetbeat sections (`interfaces`, `flows`, `protocols`, `procs`) under the `packetbeat` namespace #1490
* [x] Move settings from the `beat` namespace to top level. #1570
* [x] Change libbeat configuration file to use the dotted notation #1544 #1624
* [x] Change topbeat configuration file to use the dotted notation #1544 #1624
* [x] Change filebeat configuration file to use the dotted notation #1544 #1624
* [x] Change packetbeat configuration file to use the dotted notation #1544 #1624
* [x] Change winlogbeat configuration file to use the dotted notation #1544 #1624
* [x] Change metricbeat configuration file to use the dotted notation #1544 #1624
* [x] Create migration guide/tool: #1488
* [x] Update the system tests with the latest format of the config files. #1688 
* [x] Create a minimized version of each config file. Make it the default. #1637 #1674 #1682
* [x] Add Kafka and Redis outputs to the "full" config files. #1690 
* [x] Review all docs to update examples from configuration files. #1709

tsg:
All done. 😎

PR #1624, opened 2016-05-12
Configuration files restyling

tsg:
Little changes content wise, but some style changes:

* Use ###, ===, and --- headers to indicate hierarchy. This makes it
  easier to skim the config file. Idea stolen from elasticsearch.yml
* Unindent filebeat prospectors and metricbeat modules by one level.
  This is still valid yaml, and fewer spaces make it harder to mess up the
  white spaces.
* Reorganized a bit the logging config
* Moved the "general" libbeat section before the outputs
* Other fairly minor changes to the beats yaml files.

This is a follow up of #1544 and part of #1417.

### DIFF
--- a/filebeat/etc/beat.yml
+++ b/filebeat/etc/beat.yml
@@ -1,192 +1,189 @@
-################### Filebeat Configuration Example #########################
+###################### Filebeat Configuration Example #########################
 
-######################### Filebeat prospectors #############################
+#=========================== Filebeat prospectors =============================
 
 # List of prospectors to fetch data.
 filebeat.prospectors:
-  # Each - is a prospector. Below are the prospector specific configurations
+# Each - is a prospector. Below are the prospector specific configurations
 
-  # Type of the files. Based on this the way the file is read is decided.
-  # The different types cannot be mixed in one prospector
-  #
-  # Possible options are:
-  # * log: Reads every line of the log file (default)
-  # * stdin: Reads the standard in
-  - input_type: log
+# Type of the files. Based on this the way the file is read is decided.
+# The different types cannot be mixed in one prospector
+#
+# Possible options are:
+# * log: Reads every line of the log file (default)
+# * stdin: Reads the standard in
 
-    # Paths that should be crawled and fetched. Glob based paths.
-    # To fetch all ".log" files from a specific level of subdirectories
-    # /var/log/*/*.log can be used.
-    # For each file found under this path, a harvester is started.
-    # Make sure not file is defined twice as this can lead to unexpected behaviour.
-    paths:
-      - /var/log/*.log
-      #- c:\programdata\elasticsearch\logs\*
+#------------------------------ Log prospector --------------------------------
+- input_type: log
 
-    # Configure the file encoding for reading files with international characters
-    # following the W3C recommendation for HTML5 (http://www.w3.org/TR/encoding).
-    # Some sample encodings:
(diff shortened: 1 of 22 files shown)


tsg:
These are the "full" config files. Creating the "short" ones (which will also become the defaults) will come in a future PR.

filebeat/etc/beat.yml
```diff
@@ -1,192 +1,189 @@
-################### Filebeat Configuration Example #########################
+###################### Filebeat Configuration Example #########################
```
ruflin: Perhaps we can remove the example part and use "Example" for the short one then.

filebeat/filebeat.yml
```diff
@@ -1,205 +1,243 @@
...
 
-  # Name of the registry file. If a relative path is used, it is considered relative to the
-  # data path.
-  #registry_file: registry
+#
+# These config files must have the full filebeat config part inside, but only
+# the prospector part is processed. All global options like spool_size are ignored.
+# The config_dir MUST point to a different directory then where the main filebeat config file is in.
+#filebeat.config_dir:
+#================================ General =====================================
```
ruflin: New line between would help. Probably best to introduce it on top of libbeat.yml to have it in all files always.

metricbeat/etc/beat.yml
```diff
@@ -1,85 +1,87 @@
...
-  - module: apache
-    metricsets: ["status"]
-    enabled: true
-    period: 1s
+#==========================  Modules configuration ============================
+metricbeat.modules:
 
-    # Apache hosts
-    hosts: ["http://127.0.0.1/"]
+#----------------------------- Apache Module ----------------------------------
```
ruflin: As this is auto generated, it will not last long ;-) Has to be adjusted in the script. I can do that for you after merging if you like.

ruflin:
LGTM. See notes. I like the new styling with the headers.

filebeat/etc/beat.yml
tsg: I will do that in the PR adding the short ones, because I'm thinking the full and short should have a header mentioning the other option.

metricbeat/etc/beat.yml
tsg: Ha ha, of course. I added the headers in the `_beat/confg.yml` files and removed the 2 spaces extra indent from the script (not actually needed). Please have a look at the results.

tsg:
Comments should be addressed.

metricbeat/module/apache/_beat/config.yml
```diff
@@ -1,4 +1,4 @@
-# Apache Module
+#----------------------------- Apache Module ----------------------------------
```
ruflin: I hope to auto generate this line in the future based on a name entry in fields.yml.

metricbeat/scripts/config_collector.py
```diff
@@ -2,40 +2,39 @@
...
         # Load module yaml
         with file(module_configs) as f:
-
-            # Add 2 spaces for indentation in front of each line
             for line in f:
-                if len(line.strip()) > 0:
-                    config_yml += "  " + line
-                else:
-                    config_yml += line
+                config_yml += line
```
ruflin: Nice, also reduced script complexity :-)

ruflin:
LGTM. We should probably review all the docs after we finished all the changes.

tsg:
@ruflin good point, I added that to the meta issue (https://github.com/elastic/beats/issues/1417)

PR #1625, opened 2016-05-12
Fix spooler shutdown blocking

ruflin:


urso:
LGTM, waiting for green.

PR #1626, opened 2016-05-12
Introduce ufcg for prospector

ruflin:
In a first step, ucfg was introduced to unpack prospector configs. In next steps the config files also for the prospector types and harvester types will be split up.

* Obsolete tests were removed

### DIFF
--- a/filebeat/config/config.go
+++ b/filebeat/config/config.go
@@ -1,11 +1,9 @@
 package config
 
 import (
-	"fmt"
 	"log"
 	"os"
 	"path/filepath"
-	"regexp"
 	"time"
 
 	"github.com/elastic/beats/libbeat/cfgfile"
@@ -17,9 +15,7 @@ import (
 // Defaults for config variables which are not set
 const (
 	DefaultRegistryFile                      = "registry"
-	DefaultIgnoreOlderDuration time.Duration = 0
-	DefaultCloseOlderDuration  time.Duration = 1 * time.Hour
-	DefaultScanFrequency       time.Duration = 10 * time.Second
+	DefaultCloseOlder          time.Duration = 1 * time.Hour
 	DefaultSpoolSize           uint64        = 2048
 	DefaultIdleTimeout         time.Duration = 5 * time.Second
 	DefaultHarvesterBufferSize int           = 16 << 10 // 16384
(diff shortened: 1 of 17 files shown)


filebeat/config/config.go
```diff
@@ -17,7 +15,7 @@ import (
 // Defaults for config variables which are not set
 const (
 	DefaultRegistryFile                      = "registry"
-	DefaultIgnoreOlderDuration time.Duration = 0
+	DefaultIgnoreOlder         time.Duration = 0
 	DefaultCloseOlderDuration  time.Duration = 1 * time.Hour
```
urso: DefaultCloseOlder?

filebeat/crawler/crawler.go
```diff
@@ -39,8 +39,6 @@ func (c *Crawler) Start(prospectorConfigs []config.ProspectorConfig, eventChan c
 	// Prospect the globs/paths given on the command line and launch harvesters
 	for _, prospectorConfig := range prospectorConfigs {
```
urso: one can unpack here into:

```
prospectorInfo := struct {
    Type string `config:"input_type"  validate:"required"`
    Enabled bool `config:"enabled"`
}{"", true}

err := prospectorConfig.Unpack(&prospectorInfo)
...

if !prospectorInfo.Enabled {
  continue
}

plugin := lookup(prospectorInfo.Type)
```

filebeat/crawler/prospector.go
```diff
@@ -28,9 +31,24 @@ type Prospectorer interface {
 	Run()
 }
 
-func NewProspector(prospectorConfig cfg.ProspectorConfig, registrar *Registrar, spoolerChan chan *input.FileEvent) (*Prospector, error) {
+type prospectorConfig struct {
```
urso: I like to put these into separate config.go file.

Also add `func (p *prospectorConfig) Validate() error { return nil }`. This method is executed by ucfg to capture additional config failures and report these errors on Unpack. If error is reported, ucfg adds some additional context with path to hint which prospector config failed. In future ucfg might try to collect mutliple errors for reporting (think `-configtest`).

filebeat/crawler/prospector.go
```diff
@@ -5,14 +5,17 @@ import (
...
 	cfg "github.com/elastic/beats/filebeat/config"
 	"github.com/elastic/beats/filebeat/harvester"
 	"github.com/elastic/beats/filebeat/input"
+	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/logp"
 )
 
 type Prospector struct {
-	ProspectorConfig cfg.ProspectorConfig
+	ProspectorConfig prospectorConfig
```
urso: maybe name field `config`. having `prospector.ProspectorConfig` is pretty stuttery.

filebeat/harvester/harvester.go
```diff
@@ -37,8 +40,31 @@ type Harvester struct {
...
+	MaxBackoffDuration time.Duration
+	CloseOlder         string `config:"close_older"`
+	CloseOlderDuration time.Duration
+	ForceCloseFiles    bool                   `config:"force_close_files"`
+	ExcludeLines       []*regexp.Regexp       `config:"exclude_lines"`
+	IncludeLines       []*regexp.Regexp       `config:"include_lines"`
+	MaxBytes           int                    `config:"max_bytes"`
+	Multiline          *input.MultilineConfig `config:"multiline"`
+	JSON               *input.JSONConfig      `config:"json"`
+}
```
urso: this struct needs some more cleanup?

filebeat/config/config.go
ruflin: Changed

filebeat/crawler/crawler.go
ruflin: I save that for a later PR ;-)

filebeat/crawler/prospector.go
ruflin: I added the method but currently left it empty. Config stuff was moved to config.go

filebeat/crawler/prospector.go
ruflin: Renamed to config and made private. Now that stuff is better separated it is also not that confusing anymore ;-)

filebeat/harvester/harvester.go
ruflin: Yes, and not only this one, but not in this PR (see note in PR).

ruflin:
@urso Update pushed

PR #1628, opened 2016-05-12
Mb base structure

ruflin:

PR #1631, opened 2016-05-12
Add unique identifier for each event / event batch.

ruflin:


ruflin:
@andrewkroh Changes from https://github.com/elastic/beats/pull/1628 were applied here.

tsg:
We should make sure we have a meta ticket to:

* Implement the uid as in ES
* Use `create` (maybe optionally?) in the ES output so duplicates are removed
* In the ES output, remove the "id" field from the actual dictionary, since it will be stored as the "ID"

tsg:
Alternatively we can leave the id in the doc but have the default mappings not index it at all. But we can discuss that in a follow up ticket :-).

ruflin:
@tsg Just to make sure we are on the same page. The id generated / used here is not identical the the future unique event id. This id is used to correlate event multiple events like processes which were created at the same time.

Besides that, agree with the points above :-)

tsg:
Ah, got it. I saw the reference to the uid field from ES so I imagined it would be for that.

In this case, is it really needed? Couldn't one get the same by doing a filter on metricset + host + the exact timestamp value? I'm asking because the disk usage costs for this field are going to be significant (there are lot of process objects).

ruflin:
This should work if the host is unique. I prefer having the field as the user does not have to know the above logic to get events that belong together and it is one instead of 3 filters. It adds ~36 Bytes to each event. I assume most events will be between 500-100 Bytes. We should definitively get the ES variant in rather sooner then later.

tsg:
But is this really a common use case? I never felt the need for this type of correlation in Topbeat, because you normally want to display the evolution over time, in which case you cannot filter by the Ids, but filtering by the hostname works great from the UI point of view.

It was also never requested for Topbeat, that's why I'm dubious about its benefits. It might also be confusing, people might assume (like me) that it is an unique identifier for the event and do de-duplication based on it.

ruflin:
Closing as currently not a priority.

PR #1636, opened 2016-05-13
Update fields structure to be array

ruflin:
This makes the structure more flexible as it is easy to add new elements by just appending to the file. The sections array was removed as key and title are now part of the fields directly.

* Updated all documents to the new format
* Implemented sorting of entries for docs based on key
* Update scripts to handle the new structure


libbeat/scripts/generate_fields_docs.py
```diff
@@ -82,36 +74,21 @@ def fields_to_asciidoc(input, output, beat):
...
-        if key in docs:
-            section = docs[key]
-            if "type" in section:
-                if section["type"] == "group":
-                    section["name"] = name
-                    section["anchor"] = key
-                    document_fields(output, section, sections, "")
+    # Sort alphabetically by key
+    for v in sorted(docs["fields"], key=lambda field: field["key"]):
+        section = v
```
tsg: Why not `for section in sorted(...`? I think that should be equivalent by simpler.

tsg:
LGTM, minor comment left.

libbeat/scripts/generate_fields_docs.py
ruflin: Adjusted. `section = v` was still a left over.

PR #1637, opened 2016-05-13
Added short versions for the configuration files

tsg:
For now, the short files are named like `beatname.short.yml` and I kept the
`beatname.yml` to be the full one.  The intention is to switch them around. The
short ones to become the default and the long ones to be something like
`beatname.full.yml` or similar.

But before doing that I wanted the others in the team to have a look. The
decision of what to include in the short version is fairly arbitrary, I just
did it based on what I thought makes most sense.

I kept the ES and LS outputs in the configuration file. I considered having
only ES for all the beats except Filebeat, but that would complicate the
generation part and I think it's nice to highlight that we support multiple
outputs.

### DIFF
--- a/filebeat/etc/beat.short.yml
+++ b/filebeat/etc/beat.short.yml
@@ -1,0 +1,58 @@
+###################### Filebeat Configuration Example #########################
+
+# This file is an example configuration file highlighting only the most common
+# options. The filebeat.full.yml file from the same directory contains all the
+# supported options with more comments. You can use it as a reference.
+#
+# You can find the full configuration reference here:
+# https://www.elastic.co/guide/en/beats/filebeat/index.html
+
+#=========================== Filebeat prospectors =============================
+
+filebeat.prospectors:
+
+# Each - is a prospector. Most options can be set at the prospector level, so
+# you can use different prospectors for various configurations.
+# Below are the prospector specific configurations.
+
+- input_type: log
+
+  # Paths that should be crawled and fetched. Glob based paths.
+  paths:
+    - /var/log/*.log
+    #- c:\programdata\elasticsearch\logs\*
+
+  # Exclude lines. A list of regular expressions to match. It drops the lines that are
+  # matching any regular expression from the list.
+  #exclude_lines: ["^DBG"]
+
+  # Include lines. A list of regular expressions to match. It exports the lines that are
+  # matching any regular expression from the list.
+  #include_lines: ["^ERR", "^WARN"]
+
+  # Exclude files. A list of regular expressions to match. Filebeat drops the files that
+  # are matching any regular expression from the list. By default, no files are dropped.
+  #exclude_files: [".gz$"]
+
+  # Optional additional fields. These field can be freely picked
+  # to add additional information to the crawled log files for filtering
+  #fields:
+  #  level: debug
+  #  review: 1
+
+  ### Multiline options
+
+  # Mutiline can be used for log messages spanning multiple lines. This is common
+  # for Java Stack Traces or C-Line Continuation
+
+  # The regexp Pattern that has to be matched. The example pattern matches all lines starting with [
+  #multiline.pattern: ^\[
+
+  # Defines if the pattern set under pattern should be negated or not. Default is false.
+  #multiline.negate: false
+
+  # Match can be set to "after" or "before". It is used to define if lines should be append to a pattern
+  # that was (not) matched before or after or as long as a pattern is not matched based on negate.
+  # Note: After is the equivalent to previous and before is the equivalent to to next in Logstash
+  #multiline.match: after
+
--- a/libbeat/_beat/config.short.yml
+++ b/libbeat/_beat/config.short.yml
@@ -1,0 +1,55 @@
+
+#================================ General =====================================
+
+# The name of the shipper that publishes the network data. It can be used to group
+# all the transactions sent by a single shipper in the web interface.
+#name:
+
+# The tags of the shipper are included in their own field with each
+# transaction published.
+#tags: ["service-X", "web-tier"]
+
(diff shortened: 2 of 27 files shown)


tsg:
Part of #1417.

filebeat/etc/beat.short.yml
```diff
@@ -0,0 +1,89 @@
...
+  paths:
+    - /var/log/*.log
+    #- c:\programdata\elasticsearch\logs\*
+
+  # Configure the file encoding for reading files with international characters
+  # following the W3C recommendation for HTML5 (http://www.w3.org/TR/encoding).
+  # Some sample encodings:
+  #   plain, utf-8, utf-16be-bom, utf-16be, utf-16le, big5, gb18030, gbk,
+  #    hz-gb-2312, euc-kr, euc-jp, iso-2022-jp, shift-jis, ...
+  encoding: plain
```
ruflin: I would leave out encoding

filebeat/etc/beat.short.yml
```diff
@@ -0,0 +1,89 @@
...
+
+  # Defines if the pattern set under pattern should be negated or not. Default is false.
+  #multiline.negate: false
+
+  # Match can be set to "after" or "before". It is used to define if lines should be append to a pattern
+  # that was (not) matched before or after or as long as a pattern is not matched based on negate.
+  # Note: After is the equivalent to previous and before is the equivalent to to next in Logstash
+  #multiline.match: after
+
+  ### JSON configuration
```
ruflin: Leave out JSON

libbeat/_beat/config.short.yml
```diff
@@ -0,0 +1,60 @@
...
+# Optional fields that you can specify to add additional information to the
+# output. Fields can be scalar values, arrays, dictionaries, or any nested
+# combination of these.
+#fields:
+#  env: staging
+
+# If this option is set to true, the custom fields are stored as top-level
+# fields in the output document instead of being grouped under a fields
+# sub-dictionary. Default is false.
+#fields_under_root: false
```
ruflin: Would remvoe fields_under_root

libbeat/_beat/config.short.yml
```diff
@@ -0,0 +1,60 @@
...
+
+  # Optional TLS. By default is off.
+  # List of root certificates for HTTPS server verifications
+  #tls.certificate_authorities: ["/etc/pki/root/ca.pem"]
+
+  # Certificate for TLS client authentication
+  #tls.certificate: "/etc/pki/client/cert.pem"
+
+  # Client Certificate Key
+  #tls.certificate_key: "/etc/pki/client/cert.key"
```
ruflin: One thing I would add is the logging part with the debug level. It happens very frequently that we ask people to enable debugging. Like this they would only have to uncomment it.

metricbeat/module/apache/_beat/config.shart.yml
```diff
@@ -0,0 +1,8 @@
+#----------------------------- Apache Module ----------------------------------
+- module: apache
```
ruflin: Filename: shart instead of short

packetbeat/etc/beat.short.yml
```diff
@@ -0,0 +1,90 @@
...
+  # Configure the ports where to listen for Redis traffic. You can disable
+  # the Redis protocol by commenting out the list of ports.
+  ports: [6379]
+
+packetbeat.protocols.thrift:
+  # Configure the ports where to listen for Thrift-RPC traffic. You can disable
+  # the Thrift-RPC protocol by commenting out the list of ports.
+  ports: [9090]
+
+packetbeat.protocols.mongodb:
```
ruflin: Could this also be `packetbeat.protocols.mongodb.ports: [...]`? Would it then be enabled?

ruflin:
LGTM. +1 on moving forward with this. We can still make small adjustments later if needed. 

Two things for the naming:
* In the repository, I would keep the full one without an extension
* When downloading and starting a beat, it is important to take the short one by default

The above notes unfortunately conflict with each other :-) My suggestion is to not have the short version in the the top level folder, only under _beat or etc. During building the `beat.yml` is renamed to `beat.full.yml` and the short and is then `beat.yml`. The downside is, that this adds complexity. The idea behind is, that everyone who checks out the repo should have the full config available. Probably there is a better way ...

urso:
alternative name for `short`: `min`?

I think these very big comments make the minified version still to verbose. Let's try to have at most 1 or 2 lines commenting an option. Maybe with second line being a link to the docs (like we include link to doc reference at top of file, but it's so easily overlooked)

ruflin:
I like the 2 lines suggestion by @urso with the docs.

metricbeat/module/apache/_beat/config.shart.yml
tsg: Thanks, fixed.

packetbeat/etc/beat.short.yml
tsg: That would work yes, but it will make it more difficult to add future options. I'm also thinking that in order to be easy to copy paste from the long version to the short one, they should use the same conventions.

tsg:
I did the 2 comment lines suggestion (but not the link to the docs, that would be quite an effort to maintain) and removed some more stuff as suggested by @ruflin. Let's discuss the rest in the meeting.

tsg:
Why not just name the short one `beatname.yml` and `beatname.full.yml` the complete one in both the repository and the packages? There are people executing them from source, the principle I used so far in packaging is that running from source feels pretty much the same as running from the binary `tar.gz`.

tsg:
Added logging to the short version. I prefer doing the file rename in a follow up PR, so this is ready for merging unless more comments show up or the build fails.

monicasarbu:
LGTM

PR #1638, opened 2016-05-13
Introduce dot notation for metricset namespace

ruflin:
Before the namespace for each metricset was "moduleName-metricSetName". This was not changed to "moduleName.metricSetName", the dot being a sub document. This notation is more in line with the other beats and is nicer to be used.

* Tests were updated to check for new structure
* Apache available for system tests fixed
* Environment system tests command introduced

This PR depends on https://github.com/elastic/beats/pull/1636

### DIFF
--- a/metricbeat/scripts/fields_collector.py
+++ b/metricbeat/scripts/fields_collector.py
@@ -40,7 +40,7 @@ def collect():
                 # Add 4 spaces for indentation in front of each line
                 for line in f:
                     if len(line.strip()) > 0:
-                        fields_yml += "    " + line
+                        fields_yml += "    " + "    " + line
                     else:
                         fields_yml += line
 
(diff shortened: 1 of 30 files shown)


metricbeat/scripts/fields_collector.py
```diff
@@ -40,7 +40,7 @@ def collect():
                 # Add 4 spaces for indentation in front of each line
                 for line in f:
                     if len(line.strip()) > 0:
-                        fields_yml += "    " + line
+                        fields_yml += "    " + "    " + line
```
tsg: The comment above it out of date :)

tsg:
LGTM, will wait for the tests to finish.

metricbeat/scripts/fields_collector.py
ruflin: Will clean it up in a next PR. Code line is also not too nice ...

andrewkroh:
LGTM. I think there may still be some example events that need to be updated. Like the one in beater/doc.go.

ruflin:
@andrewkroh Definitively. Tracking the migration here. Will update all the docs at the end: https://github.com/elastic/beats/issues/1639

PR #1643, opened 2016-05-15
libbeat: fix logging section in the _beat/config.yml file

cyrilleverrier:
I did the same structure as the other configuration sections.
"logging" is the top section.
The configuration fields within the logging section use the dot notation: files.path, files.name, etc.

### DIFF
--- a/libbeat/_beat/config.yml
+++ b/libbeat/_beat/config.yml
@@ -221,7 +221,7 @@ output.elasticsearch:
 # There are three options for the log output: syslog, file, stderr.
 # Under Windows systems, the log files are per default sent to the file output,
 # under all other system per default to syslog.
-
+logging:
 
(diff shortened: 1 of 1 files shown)


cyrilleverrier:
Probably already fixed by #1637

ruflin:
@cyrilleverrier There is no hard rule on when to use do notation and when not. So it could be that it is not to 100% consistent, but we tried to make it as easy as possible for users. You wrote in the title "fix". Does this mean the previous version did not work?

tsg:
@cyrilleverrier you are right, that was fixed in #1637. Thanks for noticing.

cyrilleverrier:
@ruflin before the fix, the "level" field was on the root level. Therefore it was not possible to configure the logging level.

Issue #1600, opened 2016-05-10
Registry state persistence and cleanup


ruflin:
Filebeat currently lacks an option to cleanup the registry file. The goal of this issue is two answer three questions:

* Which files are ignore for crawling?
* Which states are removed from the registry?
* How long is the state kept for files that cannot be found anymore?

I suggest to introduce two new config options `clean_older` and `clean_removed` with the following behaviour and defaults. This is option 4 below.

| Variable      | Description                                                                                                                          | Default  |
|---------------|--------------------------------------------------------------------------------------------------------------------------------------|----------|
| close_older   | Closes file handler for files which are older then x                                                                                 | 1h       |
| clean_removed | Cleans the registry file from files which are not found anymore after x                                                              | 24h      |
| ignore_older  | Does not crawl files which are older then ignore_older. Still keeps the file state and moves the file offset to the end of the file. | infinity |
| clean_older   | Removes file from the registry which are older then x                                                                                | infinity |

Better name suggestions for the new variables are welcome.

# Option Details

Below is the current behaviour and the different options described in more detail.

## Current Behaviour

The current behaviour is as following:

* File offset for files reaching `ignore_older` are set to the end of the file and persisted.

This brings the problem that the registry is never cleaned up.

## Option 1 - ignore_older double meaning

One option is to use ignore_older to also cleanup the state. This would have the following consequences:

* If a file reaches ignore_older, the state is removed from the registry
* That means if a file that reached ignore_older is updated again, it will be read from the beginning as no state was stored

## Option 2 - clean_older >= ignore_older

The introduction of a `clean_older` variable would allow to set a time when the registry should be cleaned up. In case ignore_older has the same behaviour as now, it requires that clean_older is >= ignore_older as otherwise the two would get into a race condition. As ignore_older is set by default to infinity, it would also mean that clean_older is disabled by default. If clean_older is enabled with a time, it would mean also ignore_older has to be set.

## Option 3 - ignore_older, clean_removed

Introducing a variable `clean_removed` would allow to enable an option, that only files which disappear from disk are cleaned up. `clean_removed` could either be a bool and files that disappear would be directly removed or it could also be time a duration after whicht the files are removed. This means all files which still exist will be kept in the registry.

In case clean_removed is duration, it would require `clean_removed >= close_older` or `force_close_files` to be enable to make sure the files are closed.

## Option 4 - clean_older >= ignore_older, clean_older >= clean_removed

[...]

PR #1650, opened 2016-05-17
Refactor state to extrac from prospector

ruflin:
This allows the States struct to be reused also in the registrar which reduces complexity of registrar and prospector.

### DIFF
--- a/filebeat/crawler/prospector_log.go
+++ b/filebeat/crawler/prospector_log.go
@@ -44,18 +44,21 @@ func (p *ProspectorLog) Init() {
 			// Make sure new harvester is started for all states
 			state.Finished = true
 			// Prospector must update all states as it has to detect also file rotation
-			p.Prospector.updateState(state)
+			p.Prospector.states.Update(state)
 		}
 	}
 
-	logp.Info("Previous states loaded: %v", len(p.Prospector.harvesterStates))
+	logp.Info("Previous states loaded: %v", p.Prospector.states.Count())
 }
 
 func (p *ProspectorLog) Run() {
 	logp.Debug("prospector", "Start next scan")
 
 	p.scan()
-	p.Prospector.cleanupStates()
+	// Only cleanup states if enabled
+	if p.config.IgnoreOlder != 0 {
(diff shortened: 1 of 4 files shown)


ruflin:
jenkins, retest it

filebeat/input/state.go
```diff
@@ -20,3 +26,76 @@ func NewFileState(fileInfo os.FileInfo, path string) FileState {
...
+		states: []FileState{},
+	}
+}
+
+// Update updates a state. If previous state didn't exist, new one is created
+func (s *States) Update(newState FileState) {
+
+	index, oldState := s.FindPrevious(newState)
+
+	// Only block after FindPrevious as FindPrevious has its own lock
```
urso: is there a chance for a race between FindPrevious and actual body updating the state?

filebeat/crawler/prospector_log.go
```diff
@@ -44,18 +44,21 @@ func (p *ProspectorLog) Init() {
...
 }
 
 func (p *ProspectorLog) Run() {
 	logp.Debug("prospector", "Start next scan")
 
 	p.scan()
-	p.Prospector.cleanupStates()
+	// Only cleanup states if enabled
+	if p.config.IgnoreOlder != 0 {
+		p.Prospector.states.Cleanup(p.config.IgnoreOlder)
```
urso: so, by default we never clean up the state?

ruflin: Yes, this PR does not change the behaviour. For changing it see https://github.com/elastic/beats/issues/1600

filebeat/input/state.go
ruflin: Do you mean if it could happen that after FindPrevious an other method gets the lock and updates the state? Potentially yes I would say, but not 100% sure about the affect. I think what would happen is that a newer state could be overwritten by an older one. This would have an affect until a new state comes in which then brings it up-to-date again. So in between it could happen, that an older state exists.

ruflin: I pushed a new version with a public and private FindPrevious method. Only the public applies the lock ...

PR #1651, opened 2016-05-17
Auto generates title for config files

ruflin:
* Adjust nginx to new naming structure
* Add dynamic title to docs

### DIFF
--- a/metricbeat/docs/modules/redis.asciidoc
+++ b/metricbeat/docs/modules/redis.asciidoc
@@ -24,13 +24,12 @@ Redis has two additional config options:
 
 === Example Configuration
 
-The Apache module supports the standard configuration options which can be found
+The Redis Status module supports the standard configuration options which can be found
 here (add link). Below is an example of a configuration option:
 
 [source,yaml]
 ----
 metricbeat.modules:
-  #------------------------------- Redis Module ---------------------------------
   - module: redis
     metricsets: ["info"]
     enabled: true
--- a/metricbeat/scripts/config_collector.py
+++ b/metricbeat/scripts/config_collector.py
@@ -1,5 +1,6 @@
 import os
 import argparse
+import yaml
 
 # Collects config for all modules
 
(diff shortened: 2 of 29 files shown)


metricbeat/scripts/config_collector.py
```diff
@@ -61,6 +70,20 @@ def collect(short=False):
     # output string so it can be concatenated
     print config_yml
 
+# Makes sure every title line is 80 chars long
```
tsg: Actually I was going for 79 chars to leave one for the new line character. That's how it is in the master files so if this is 80 it might not align nicely.

metricbeat/scripts/config_collector.py
```diff
@@ -61,6 +70,20 @@ def collect(short=False):
...
+def get_title_line(title):
+    dashes = (80 - 10 - len(title)) / 2
+
+    line = "#"
+    for i in range(0, dashes):
+        line += "-"
```
tsg: You missed the once in lifetime opportunity to use the the multiply operator for strings: `"-" * dashes` :-). Just kidding.

ruflin: Nooooo, didn't know that exists :-)

metricbeat/scripts/config_collector.py
ruflin: Ok, that means also below I have to take chars from 0:78 to have 80 including the newline.

metricbeat/docs/modules/redis.asciidoc
```diff
@@ -24,13 +24,12 @@ Redis has two additional config options:
 
 === Example Configuration
 
-The Apache module supports the standard configuration options which can be found
+The Redis module supports the standard configuration options which can be found
```
tsg: The others have "Nginx Status", but this one only has "Redis" as a title.

tsg:
LGTM. Minor comments left.

metricbeat/docs/modules/redis.asciidoc
ruflin: Will add it. This depends on the YAML file. Not even sure if we should keep Status for the others, as a module can have metricsets that report mor then just status.

ruflin:
Changes applied and opportunity used to use string multiplier :-)

PR #1659, opened 2016-05-18
Streamline topbeat config loading

urso:
Update topbeat config loading to match most ucfg changes. Also 'period' has become time.Duration and can be initialized as low as '1ms'.

### DIFF
--- a/topbeat/beater/config.go
+++ b/topbeat/beater/config.go
@@ -1,17 +1,48 @@
 package beater
 
-type TopConfig struct {
-	Period *int64
-	Procs  *[]string
-	Stats  struct {
-		System     *bool `config:"system"`
-		Proc       *bool `config:"process"`
-		Filesystem *bool `config:"filesystem"`
-		CpuPerCore *bool `config:"cpu_per_core"`
-	}
-}
+import (
+	"errors"
+	"time"
+)
 
 type ConfigSettings struct {
-	Input   *TopConfig `config:"input"`
-	Topbeat *TopConfig `config:"topbeat"`
+	Topbeat TopConfig `config:"topbeat"`
+}
+
+type TopConfig struct {
+	Period time.Duration     `config:"period"  validate:"min=1ms"`
+	Procs  []string          `config:"procs"`
+	Stats  StatsEnableConfig `config:"stats"`
+}
+
+type StatsEnableConfig struct {
+	System     bool `config:"system"`
+	Proc       bool `config:"process"`
+	Filesystem bool `config:"filesystem"`
+	CPUPerCore bool `config:"cpu_per_core"`
+}
+
+var (
+	defaultConfig = TopConfig{
+		Period: 10 * time.Second,
+		Procs:  []string{".*"}, //all processes
+		Stats: StatsEnableConfig{
+			System:     true,
+			Proc:       true,
+			Filesystem: true,
+			CPUPerCore: false,
+		},
+	}
+)
+
+func (c *TopConfig) Validate() error {
+	return nil
+}
+
+func (c *StatsEnableConfig) Validate() error {
+	if !c.System && !c.Proc && !c.Filesystem {
+		return errors.New("Invalid statistics configuration (enable one of 'system', 'process' or 'filesystem')")
+	}
+
+	return nil
 }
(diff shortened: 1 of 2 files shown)


topbeat/beater/config.go
```diff
@@ -1,17 +1,48 @@
 package beater
 
+import (
+	"errors"
+	"time"
+)
+
+type ConfigSettings struct {
```
ruflin: We should standardise this naming across all beats. Suggestion:

This one we call `Config`, the one inside is `TopbeatConfig`. Same as we have in the cookiecutter: https://github.com/elastic/beat-generator/blob/master/%7B%7Bcookiecutter.beat%7D%7D/config/config.go

@urso I remember you once used even a nested notation, so the "empty object was not needed?

topbeat/beater/config.go
```diff
@@ -1,17 +1,48 @@
...
-		System     *bool `config:"system"`
-		Proc       *bool `config:"process"`
-		Filesystem *bool `config:"filesystem"`
-		CpuPerCore *bool `config:"cpu_per_core"`
+	Period time.Duration     `config:"period"  validate:"min=1ms"`
+	Procs  []string          `config:"procs"`
+	Stats  StatsEnableConfig `config:"stats"`
+}
+
+type StatsEnableConfig struct {
```
ruflin: I quite liked the previous nested notation :-)

topbeat/beater/config.go
```diff
@@ -1,17 +1,48 @@
...
+)
+
+func (c *TopConfig) Validate() error {
+	return nil
 }
 
-type ConfigSettings struct {
-	Input   *TopConfig `config:"input"`
-	Topbeat *TopConfig `config:"topbeat"`
+func (c *StatsEnableConfig) Validate() error {
```
ruflin: Ok, I see why you need the non nested one ...

ruflin:
LGTM

ruflin:
@urso This looks like a BC break? Before you specified 10 for the period, now it has to be "10s". You should update the CHANGELOG.

topbeat/beater/config.go
urso: yes, one can uses nested paramters. But due to topbeat trying to figure out the right section this won't work. `ConfigSettings` being obsolete, how about:

```
type Config struct {
  Period time.Duration    `config:"period"  validate:"min=1ms"`
  Procs []string          `config:"procs"`
  Stats StatsEnableConfig `config:"stats"`
}
```

The beater already finds and unpacks from correct child section.

topbeat/beater/config.go
urso: nested struct is not needed for `Validate` as `Topbeat` config also has validate method. Problem is line 29 defining the defaults.

topbeat/beater/config.go
ruflin: Very nice. We should apply the same to beat-generator and all beats.

urso:
it's no breaking change. Just using numbers (e.g. 10) ucfg will convert value into seconds. This is on purpose, so we don't break all our configs when updating to ucfg.

ruflin:
@urso ok, do we have other places where we didn't use the unit? Does ucfg print a warning or something if it is used without unit? Would be good so we get people to migrate. I still would put it into the CHANGELOG as deprecated, in the hope people adjust there config.

Also the beat.yml should be updated to use `10s` I think?

topbeat/beater/config.go
urso: Not so nice, as topbeat is special. In [topbeat.go:42](https://github.com/elastic/beats/pull/1659/files#diff-478f3b0d71959277c6d22459d829d90aR42) we do the unpack (as 'input' is deprectated alias for 'topbeat'). Maybe in v6 when removing support for deprecated sections we can make libbeat/beater pass correct section.

ruflin: I didn't mean to do it fully automatically, just that we put the section into the unpack for each beat. But if we can fully automate it in the future, even better.

urso:
@ruflin almost every config that used to be time.Duration used to use int. There might be some more left over. That's why I didn't deprecated int/uint, but added support for int/uint/float to be interpreted in seconds.

andrewkroh:
LGTM

PR #1660, opened 2016-05-18
Moved one line to the bug fixes section and added a reference to the …

tsg:
…pull request


ruflin:
I actually put that on purpose into Added it wasn't really a bug. It was an intentional behaviour in the previous version as it handled all ingore_files the same independent of the offset. This is more an improvement.

tsg:
OK, fine with me. I then moved Added after Bug fixes and added the link to the pull request.

ruflin:
LGTM. Directly merging as no builds needed here.

PR #1661, opened 2016-05-18
Automate imports generation

ruflin:
Running `make imports` automatically generates the imports in list.go for all existing modules and metricsets


metricbeat/scripts/generate_imports.py
```diff
@@ -0,0 +1,50 @@
...
+header = """/*
+Package include imports all Module and MetricSet packages so that they register
+their factories with the global registry. This package can be imported in the
+main package to automatically register all of the standard supported Metricbeat
+modules.
+*/
+package include
+
+import (
+	// Every module and metricset must be added here so that they can register
```
andrewkroh: Consider changing this comment in some way to mention that this file is generated and that it should not be hand modified.

ruflin: Done. I added a note that it is auto generated. All the details which were in this comment are also in the package description above, so I removed them.

PR #1663, opened 2016-05-18
Update Changelog with releases from 1.2 branch

tsg:
Since we no longer merge to master, these releases were not present in the
master branch, and we should not lose them.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -73,6 +73,31 @@ https://github.com/elastic/beats/compare/v5.0.0-alpha2...master[Check the HEAD d
 
 ////////////////////////////////////////////////////////////
 
+[[release-notes-1.2.3]]
+=== Beats version 1.2.3
+https://github.com/elastic/beats/compare/v1.2.2...v1.2.3[View commits]
+
+==== Bugfixes
+
+*Topbeat*
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.asciidoc
```diff
@@ -73,6 +73,31 @@ https://github.com/elastic/beats/compare/v5.0.0-alpha2...master[Check the HEAD d
...
+
+==== Bugfixes
+
+*Topbeat*
+
+- Fix high CPU usage when using filtering under Windows. {pull}1598[1598]
+
+*Filebeat*
+
+- Fix rotation issue with ignore_older. {issue}1528[1528]
```
ruflin: It's somehow strange to have this here on top as it feels like this issue existed in 5.0.0-alpha2. Do we sort by date or by release numbers? This is probably also for the download pages ...

tsg: Yeah this is difficult because the Changelog is linear but we have branches in code. I sorted by release date but kept the changelog in both if an issue is fixed in both. So people will see that a particular issue is also fixed in 5.0-alpha2. Not sure if there's a better way.

ruflin: Currently I don't see a better one ...

PR #1665, opened 2016-05-18
add process filters system.process metricset

urso:


andrewkroh:
LGTM. Somewhere (I'm not sure where) this configuration option will need to be documented (we have lots of docs to fix up).

monicasarbu:
LGTM

urso:
Yeah, didn't find the place config options are documented.

ruflin:
The config needs to be updated in `metricbeat/module/system/_beat/config.yml`.

ruflin:
@urso Not too happy about the naming, as a `process` config inside the `process` metricset is somehow confusing. What about:

* selector
* process_filter

This brings up the general discussion again if we should replace everything by filters or have sometimes more efficient "methods". I initially had "selectors" in each configuration block that allowed to define which top level doc entries should be shipped (white listing). We removed this in favor of filtering. The pro is to have one unified experience and setup, the con is that it is sometimes less efficient and more complex to configure.

urso:
@ruflin this is how my config looks like:

```
metricbeat.modules:
- module: system
  metricsets: ["cpu", "memory", "process"]
  enabled: true
  period: 1s
  processes: [ ... ]
```

Due to config being on module level naming it processes sounds quite reasonable to me. I don't see any stuttering.

ruflin:
@urso Ok, lets go with "processes".

PR #1668, opened 2016-05-18
Remove topic about viewing topology in Kibana

dedemorton:
Removing this topic because it's a bit janky at this point to ask people to use a Kibana 3 fork (not to mention old beta dashboards)


ruflin:
Agree. If someone needs the content, it is still in the old versions.

PR #1666, opened 2016-05-18
Updated package names

tsg:
To match the conventions agreed with the other Elastic projects. Summary of the
changes:

* archs are called x86 and x64 except for deb/rpm where we keep the OS
  preferred ways
* Use the ~ from the rpm/deb metadata, but - in the file names
* include `linux` in linux name  and `x86` in darwin name
* added these conventions to the README file

 Here are the resulting filenames:

```
filebeat-5.0.0-amd64.deb
filebeat-5.0.0-i386.deb

filebeat-5.0.0-i686.rpm
filebeat-5.0.0-x86_64.rpm

filebeat-5.0.0-linux-x64.tar.gz
filebeat-5.0.0-linux-x86.tar.gz

filebeat-5.0.0-windows-x64.zip
filebeat-5.0.0-windows-x86.zip

filebeat-5.0.0-darwin-x86.tar.gz
```
and the "snapshot" versions:

```
filebeat-5.0.0-SNAPSHOT-amd64.deb
filebeat-5.0.0-SNAPSHOT-i386.deb

filebeat-5.0.0-SNAPSHOT-i686.rpm
filebeat-5.0.0-SNAPSHOT-x86_64.rpm

filebeat-5.0.0-SNAPSHOT-linux-x64.tar.gz
filebeat-5.0.0-SNAPSHOT-linux-x86.tar.gz

filebeat-5.0.0-SNAPSHOT-windows-x64.zip
filebeat-5.0.0-SNAPSHOT-windows-x86.zip

filebeat-5.0.0-SNAPSHOT-darwin-x86.tar.gz
```

PR #1673, opened 2016-05-19
Pure Go cross-compile

### LINKED ISSUE #960: filebeat dynamically linked
When I download Filebeat 1.1.0 LINUX 64-BIT and do ```file filebeat```:
```
ELF 64-bit LSB  executable, x86-64, version 1 (SYSV), dynamically linked (uses shared libs), not stripped
```
What are the shared libs it's depending on?
Is it possible to be statically linked?

tsg:
Depends on #1666.

This produces statically linked binaries for Filebeat and Winlogbeat (which don't use CGo) by using simple cross compilation, without Cgo enabled. The advantage of the statically linked binaries is that they tend to be more portable.

Closes #960.

This also gets rid of the code duplication between `xgo-image` and `xgo-image-deb6`, at least for the `build.sh` script. It does that by simply copying the one from `xgo-image` over to `xgo-image-deb6`, I didn't have a better idea.


tsg:
@cyrilleverrier you might be interested in this one.

cyrilleverrier:
@tsg I had a quick look. LGTM.

dev-tools/packer/Makefile
```diff
@@ -8,8 +8,8 @@ all: packetbeat/deb packetbeat/rpm packetbeat/darwin packetbeat/win packetbeat/b
 	winlogbeat/win \
 	build/upload/build_id.txt
 
-.PHONY: packetbeat topbeat filebeat metricbeat winlogbeat
-packetbeat topbeat filebeat metricbeat winlogbeat: build/upload
+.PHONY: packetbeat topbeat metricbeat winlogbeat
```
ruflin: winlogbeat should probably be removed here too

ruflin:
LGTM besides one small commented.

dev-tools/packer/Makefile
tsg: thanks, fixed.

andrewkroh:
Fixes #960   👍 

LGTM

andrewkroh:
jenkins, package it

tsg:
jenkins, package it

PR #1674, opened 2016-05-19
Swap full and short configuration files

tsg:
* filebeat.yml is now the short version
* filebeat.full.yml is now the complete version

For Metricbeat, if only config.yml exists, it is used for both
short and full.


ruflin:
LGTM

PR #1676, opened 2016-05-19
Fix flaky topbeat/filtering test

tsg:
I think the issue is that due to flattening `proc.mem` never shows
up as a key. Use `proc.mem.size` instead.


ruflin:
LGTM

PR #1677, opened 2016-05-19
Introduce structure for metricset event metadata

ruflin:


tsg:
This one needs a rebase.

I find `metricset.module` a bit confusing because the hierarchy goes the other way around. But it's also nice to have them grouped.

ruflin:
@tsg I also stumbled over it first. But in the end data that arrives at Elasticsearch is only the metricset metrics and this is part of a module.

Side note: I became a big fan of the name MetricSet. I think it fits really well.

PR #1680, opened 2016-05-19
Add banner indicating that guide for creating new beat is out of date

dedemorton:
This change adds a banner to the page to indicate that the content is out of date. I've also included the URL for the beat generator because I think it would be useful for developers. I hope it's OK to mention the Beat generator in the docs. Let me know if it's not.

The second pass block clears the header so that it doesn't appear on subsequent sections.

@monicasarbu Here's the change we talked about.


monicasarbu:
@dedemorton Great, thanks!

PR #1682, opened 2016-05-20
Install full configs in packages

tsg:
This adds the "full" version of the configuration files to the
config path of all the artifacts.

### DIFF
--- a/dev-tools/packer/xgo-scripts/before_build.sh
+++ b/dev-tools/packer/xgo-scripts/before_build.sh
@@ -33,15 +33,15 @@ cp $BEATNAME.template-es2x.json $PREFIX/$BEATNAME.template-es2x.json
 
 # linux
 cp $BEATNAME.yml $PREFIX/$BEATNAME-linux.yml
-
-# binary
-cp $BEATNAME.yml $PREFIX/$BEATNAME-binary.yml
+cp $BEATNAME.full.yml $PREFIX/$BEATNAME-linux.full.yml
 
 # darwin
 cp $BEATNAME.yml $PREFIX/$BEATNAME-darwin.yml
+cp $BEATNAME.full.yml $PREFIX/$BEATNAME-darwin.full.yml
 
 # win
 cp $BEATNAME.yml $PREFIX/$BEATNAME-win.yml
+cp $BEATNAME.full.yml $PREFIX/$BEATNAME-win.full.yml
 
 # Contains beat specific adjustments. As it is platform specific knowledge, it should be in packer not the beats itself
 PREFIX=$PREFIX make before-build
--- a/winlogbeat/Makefile
+++ b/winlogbeat/Makefile
@@ -12,7 +12,3 @@ gen:
 # This is called by the beats packer before building starts
 .PHONY: before-build
 before-build:
-	# Windows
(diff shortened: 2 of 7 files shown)


winlogbeat/Makefile
```diff
@@ -12,7 +12,3 @@ gen:
 # This is called by the beats packer before building starts
 .PHONY: before-build
 before-build:
-	# Windows
-	sed -i 's|#\{0,1\}\(to_files:\).*|\1 true|' $(PREFIX)/$(BEATNAME)-win.yml
-	sed -i 's|#\{0,1\}\(level:\).*|\1 info|' $(PREFIX)/$(BEATNAME)-win.yml
-	sed -i '/log files/{n;s|#\{0,1\}path:.*|path: C:/ProgramData/winlogbeat/Logs|}' $(PREFIX)/$(BEATNAME)-win.yml
```
tsg: @andrewkroh I removed these as the current defaults should work fine and I want to minimize the number of seds we do on the config files.

tsg:
jenkins, package it

dev-tools/packer/xgo-scripts/before_build.sh
```diff
@@ -33,15 +33,15 @@ cp $BEATNAME.template-es2x.json $PREFIX/$BEATNAME.template-es2x.json
 
 # linux
 cp $BEATNAME.yml $PREFIX/$BEATNAME-linux.yml
```
ruflin: For linux we do not copy full?

dev-tools/packer/xgo-scripts/before_build.sh
```diff
@@ -33,15 +33,15 @@ cp $BEATNAME.template-es2x.json $PREFIX/$BEATNAME.template-es2x.json
...
-
-# binary
-cp $BEATNAME.yml $PREFIX/$BEATNAME-binary.yml
```
ruflin: Don't we also need the short one for the binary?

ruflin: Ok, makes sense now. We don't need the binary anymore as we also use linux for the binary one. Ignore my 2 comments.

ruflin:
LGTM

winlogbeat/Makefile
andrewkroh: The only thing I will miss here is having the info level logging out of the box. When you are new first time user it's nice to have some positive feedback in the logs that its running and indexing events.

tsg: Yeah, I thought so, but that makes it inconsistent with the rest of the Beats. The reason we keep it to ERR is that WARN is too verbose in Packetbeat, maybe in other beats as well. Maybe we should move most of those WARNs to DEBUG and then switch to INFO by default.

I think our logging system, in general, is due for refactoring.

ruflin: I did in the past quite  a few changes to filebeat and libbeat to make "INFO" feasible. +1 on switching to Info and do some refactoring of packetbeat. That would be also a win that the log file gets generated by default as there are always info messages on startup.

ruflin:
@tsg @andrewkroh Merging this one as I assume we agree on improving the logging. I would even suggest to enabled info by default in a small PR and then tackle packetbeat?

PR #1683, opened 2016-05-20
Implement new fields naming for apache

ruflin:


metricbeat/etc/fields.yml
```diff
@@ -47,31 +47,31 @@
...
               type: float
               description: >
                 Bytes per second.
-            - name: bytesPerReq
+            - name: bytes_per_request
               type: float
               description: >
                 Bytes per request.
-            - name: busyWorkers
+            - name: workers.busy
```
monicasarbu: @ruflin do we want a dot here in the field name?

metricbeat/metricbeat.template-es2x.json
```diff
@@ -115,41 +113,51 @@
...
-                    "serverUptimeSeconds": {
+                    "server_uptime": {
                       "type": "integer"
                     },
                     "uptime": {
                       "type": "integer"
                     }
                   }
+                },
+                "workers": {
```
monicasarbu: Great!

metricbeat/etc/fields.yml
monicasarbu: Sorry, I just noticed that the generated template contains a sub-dictionary. Please ignore the question.

PR #1688, opened 2016-05-20
Updated configuration templates used in tests

tsg:
The goal was to bring them closer to the new conventions we
use in the configuration files, so what we test is closer to what
we ship in terms of configuration.

I also made steps in unifying the libbeat portion of these templates so
we could remove the duplication in the future.

Part of #1417.


ruflin:
Didn't check every single line, but changes LGTM and tests pass ;-) I'm looking forward to move the common parts to libbeat. Perhaps we can even use `include` from jina ...

PR #1689, opened 2016-05-20
Add async logstash output mode supporting pipelining

urso:
Requires async client support from go-lumber introduced in #1678

### DIFF
--- a/libbeat/outputs/logstash/config.go
+++ b/libbeat/outputs/logstash/config.go
@@ -13,6 +13,7 @@ type logstashConfig struct {
 	LoadBalance      bool                  `config:"loadbalance"`
 	BulkMaxSize      int                   `config:"bulk_max_size"`
 	Timeout          time.Duration         `config:"timeout"`
+	Pipelining       int                   `config:"pipelining"        validate:"min=0"`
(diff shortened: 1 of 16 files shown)


urso:
blocked by #1715

libbeat/outputs/logstash/config.go
```diff
@@ -13,6 +13,7 @@ type logstashConfig struct {
 	LoadBalance      bool                  `config:"loadbalance"`
 	BulkMaxSize      int                   `config:"bulk_max_size"`
 	Timeout          time.Duration         `config:"timeout"`
+	Pipelining       int                   `config:"pipelining"        validate:"min=0"`
```
ruflin: This should probably also be added to the libbeat.yml file and the docs?

urso: It's not added on purpose, as current logstash-input-beats plugin contains a bug making logstash fail miserably once enabled. Will add pipelining option to docs once it's safe to use.

urso: done

ruflin:
Merged. I think best is to test this out in the wilde.

PR #1690, opened 2016-05-20
Add Redis and Kafka outputs to the full configs

tsg:


filebeat/filebeat.full.yml
```diff
@@ -395,22 +395,206 @@ output.elasticsearch:
...
+  #max_message_bytes: 1000000
+
+  # The ACK reliability level required from broker. 0=no response, 1=wait for
+  # local commit, -1=wait for all replicas to commit. The default is 1.  Note:
+  # If set to 0, no ACKs are returned by Kafka. Messages might be lost silently
+  # on error.
+  #required_acks: 0
+
+  # The number of seconds to wait for new events between two producer API calls.
+  #flush_interval: 1
```
ruflin: Not related to this PR, but for durations in my opinion we should always include the unit, means this would be 1s (@urso).

filebeat/filebeat.full.yml
```diff
@@ -395,22 +395,206 @@ output.elasticsearch:
...
+  #db: 0
+
+  # The Redis data type to use for publishing events. If the data type is list,
+  # the Redis RPUSH command is used. If the data type is channel, the Redis
+  # PUBLISH command is used. The default value is list.
+  #datetype: list
+
+  # The Redis host to connect to when using topology map support. Topology map
+  # support is disabled if this option is not set.
+  #host_topology:
```
ruflin: Do we still need this config option?

filebeat/filebeat.full.yml
```diff
@@ -395,22 +395,206 @@ output.elasticsearch:
...
+  # The URL of the SOCKS5 proxy to use when connecting to the Redis servers. The
+  # value must be a URL with a scheme of socks5://.
+  #proxy_url:
+
+  # This option determines whether Redis hostnames are resolved locally when
+  # using a proxy. The default value is false, which means that name resolution
+  # occurs on the proxy server.
+  #proxy_use_local_resolver: false
+
+  # Optional TLS. By default is off.
```
ruflin: Perhaps better: Optional TLS configuration options. TLS is off by default.

filebeat/filebeat.full.yml
```diff
@@ -395,22 +395,206 @@ output.elasticsearch:
...
   #path: "/tmp/filebeat"
 
-  # Name of the generated files. The default is `filebeat` and it generates files: `filebeat`, `filebeat.1`, `filebeat.2`, etc.
+  # Name of the generated files. The default is `filebeat` and it generates
+  # files: `filebeat`, `filebeat.1`, `filebeat.2`, etc.
   #filename: filebeat
 
-  # Maximum size in kilobytes of each file. When this size is reached, the files are
-  # rotated. The default value is 10240 kB.
+  # Maximum size in kilobytes of each file. When this size is reached, the files
```
ruflin: Should we add a note that the file is also rotated on restart?

ruflin:
LGTM. Some small comments added. @urso Could you also have a quick look, as you know more about the Redis / Kafka configs.

filebeat/filebeat.full.yml
```diff
@@ -395,22 +395,206 @@ output.elasticsearch:
...
+  # Set Kafka topic by event type. If use_type is false, the topic option must
+  # be configured. The default is false.
+  #use_type: false
+
+  # The number of concurrent load-balanced Kafka output workers.
+  #worker: 1
+
+  # The number of times to retry publishing an event after a publishing failure.
+  # After the specified number of retries, the events are typically dropped.
+  # Some Beats, such as Filebeat, ignore the max_retries setting and retry until
```
JarenGlover:
```
Some Beats, such as Filebeat, ignore the max_retries setting and retry until  all events are published. 
```
^^  Is this config file only for Filebeat? If so, can then this section can be deleted? This might remove some confusion when working between different beats.

ruflin: @JarenGlover The problem here is that this general part is coming from the libbeat config file, means it is generated. Like this we make sure we don't have to keep multiple files up-to-date with the config option.

filebeat/filebeat.full.yml
```diff
@@ -395,22 +395,206 @@ output.elasticsearch:
...
+
+  # The number of seconds to wait for responses from the Kafka brokers before
+  # timing out. The default is 30s.
+  #timeout: 30s
+
+  # The maximum duration a broker will wait for number of required ACKs. The
+  # default is 10s.
+  #broker_timeout: 10s
+
+  # Per Kafka broker number of messages buffered in output pipeline. The default
```
JarenGlover: Perhaps:  The number of messages buffered on each Kafka broker. The default is 256.

filebeat/filebeat.full.yml
JarenGlover: @ruflin aha .. makes sense. thx

ruflin: @JarenGlover There are also other options which could need beat specific optimizations. Perhaps we add in the future some logic to the generation to accomodate this. Thanks for the inputs.

filebeat/filebeat.full.yml
tsg: Seems like topology handling is still a feature of the Redis output. We could consider deprecating it or marking it experimental or something like that.

filebeat/filebeat.full.yml
tsg: I updated it to "The number of messages buffered for each Kafka broker. The default is 256." to make clear that it's the Beat that buffers, not the Kafka brokers. @dedemorton, let me know if I made it worse.

tsg:
@ruflin @JarenGlover Thanks for the reviews. I added a commit with the editing you suggested, please have a look.

tsg:
Part of #1417.

urso:
btw. to make it easier to figure the outputs options I did put them into config.go for each output module.

- [Kafka output config](https://github.com/elastic/beats/blob/master/libbeat/outputs/kafka/config.go)
- [Redis output config](https://github.com/elastic/beats/blob/master/libbeat/outputs/redis/config.go)

All network based outputs have `bulk_max_size` and `flush_interval` config options in addition: https://github.com/elastic/beats/blob/master/libbeat/publisher/output.go#L20

filebeat/filebeat.full.yml
```diff
@@ -395,22 +395,206 @@ output.elasticsearch:
...
+  # is 256.
+  #channel_buffer_size: 256
+
+  # The keep-alive period for an active network connection. If 0s, keep-alives
+  # are disabled. The default is 0 seconds.
+  #keep_alive: 0
+
+  # Sets the output compression codec. Must be one of none, snappy and gzip. The
+  # default is snappy.
+  #compression: snappy
```
urso: default is gzip. You have to jump some (small) hoops to get snappy support in your kafka brokers.

filebeat/filebeat.full.yml
```diff
@@ -395,22 +395,206 @@ output.elasticsearch:
...
+  # The maximum permitted size of JSON-encoded messages. Bigger messages will be
+  # dropped. The default value is 1000000 (bytes). This value should be equal to
+  # or less than the broker’s message.max.bytes.
+  #max_message_bytes: 1000000
+
+  # The ACK reliability level required from broker. 0=no response, 1=wait for
+  # local commit, -1=wait for all replicas to commit. The default is 1.  Note:
+  # If set to 0, no ACKs are returned by Kafka. Messages might be lost silently
+  # on error.
+  #required_acks: 0
```
urso: default should be 1

filebeat/filebeat.full.yml
urso: Be super carefull about just changing durations in config files. Quite some code has not been adjusted to time.Duration support in ucfg.

urso: flush_interval uses time.Duration and can be changes to `1s`.

filebeat/filebeat.full.yml
```diff
@@ -395,22 +395,206 @@ output.elasticsearch:
   # Configure curve types for ECDHE based cipher suites
   #tls.curve_types: []
 
+#------------------------------- Kafka output ---------------------------------
+#output.kafka:
+  # The list of Kafka broker addresses to connect to.
+  #hosts: ["localhost:9092"]
```
urso: tl;dr This list is only about fetching metadata, not about actual endpoints events are published too

To be more exact, it's the initial list of configured kafka brokers to fetch cluster metadata from in order to identify all kafka brokers and how topics/partitions are distributed between brokers. The actual connections to kafka brokers is made on advertised names being included in metadata.

Just giving an example this failing badly:
  - kafka broker in host `mykafka`
  - kafka broker advertises `localhost`
  - beats running on different machine with `hosts` being configured to `mykafka:9092`.

Having this setup, beat will connect to mykafka:9092 in order to fetch the metadata. The metadata state, there is one (and only one) kafka broker being available on localhost:9092 -> beats will fail to publish events due to no kafka instance running on `localhost:9200`.

If possible, it's recommended to configure multiple brokers (3 or 5 for example), in case the broker is down.

filebeat/filebeat.full.yml
tsg: OK, it was wrong in the docs then, I'll fix it there as well.

filebeat/filebeat.full.yml
urso: Maybe add: redis itself does not support TLS. If deployed behind TLS endpoint (e.g. stunnel), TLS can be configured directly from within beats

filebeat/filebeat.full.yml
tsg: @urso Hmm, can you suggest an alternate explanation for the setting, please? I took this from the docs, but if this can cause issues we should clarify in both places.

filebeat/filebeat.full.yml
JarenGlover: ^^ this is better wording ... when I first read it, I was thinking the buffering happen on kafka's side not beat's. 💯

filebeat/filebeat.full.yml
ruflin: I was hoping so as this is a new implementation :-)

filebeat/filebeat.full.yml
ruflin: Did we ever test if this still works? :-)

ruflin: I'm ok leaving it in for the moment, we can check this in a later PR.

filebeat/filebeat.full.yml
tsg: Changing it to "The list of Kafka broker addresses from where to fetch the cluster metadata.". What do you say, @urso?

urso: +1.

How about "The list of Kafka broker addresses from where to fetch the cluster metadata. The cluster metadata contain the actual kafka brokers events are published to." 

Should be enough (even for docs). People knowing kafka what's meant, and people having no clue about kafka will run into issues anyway.

Issue #982, opened 2016-02-13
Add step about testing the config file to the gettng started for each beat


dedemorton:
The getting started guide for each beat should show how to test the config file (the `-configtest` option).

Might also want to add a note to the sections about configuring the beat and the intro topic for the config options doc.

PR #1696, opened 2016-05-21
Add info about testing the config file

dedemorton:
Winlogbeat is not included here because the docs already have this info.

This closes issue #982

### DIFF
--- a/packetbeat/docs/gettingstarted.asciidoc
+++ b/packetbeat/docs/gettingstarted.asciidoc
@@ -186,6 +186,9 @@ output:
 +
 If you are sending output to Logstash, see <<config-packetbeat-logstash>> instead.
 
+TIP: To test your configuration file, run Packetbeat in the foreground with the following options specified:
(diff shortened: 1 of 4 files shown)


packetbeat/docs/gettingstarted.asciidoc
```diff
@@ -186,6 +186,9 @@ output:
 +
 If you are sending output to Logstash, see <<config-packetbeat-logstash>> instead.
 
+TIP: To test your configuration file, run Packetbeat in the foreground with the following options specified:
++sudo ./packetbeat -configtest -e+.
```
ruflin: @tsg Do we need the sudo here?

tsg: Just tested. Yes, that's needed because Packetbeat tries to open the sniffing socket on `-configtest`.

ruflin: Thanks for checking.

PR #1702, opened 2016-05-23
Debug level in the default config (commented out)

tsg:
This makes it really easy to go to debug mode. The default in the code (when the option is commented out) is still ERR.


ruflin:
LGTM

PR #1703, opened 2016-05-23
Migration to new state structure

ruflin:
The previous state list was dependent on a file path. This had the potential to lead to overwrite and conflicts on file rotation. As the file path is also stored inside the state object this information was duplicated. Now a pure array is stored in the registry which makes the format more flexibel and brings it close to the format used by Logstash.

Changes:
* Not storing an index with paths as this leaded to duplicates
* Introduction of last_seen to differentiate between entries with same bath which show up multiple times. This introduces also the base to clean up the registry file. Currently it is only used in the prospector, no registry yet.
* Registry can now contain multiple entries for a path
* Refactor registrar to use the same state array as prospector does
* Introduce migration check to migrate from old to new state. Make it backward compatible by checking for old structure on startup
* Decouple registrar from prospector
* Update tests to follow new model

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -55,6 +55,7 @@ https://github.com/elastic/beats/compare/v5.0.0-alpha2...master[Check the HEAD d
 *Topbeat*
 
 *Filebeat*
+- The registry format was changed to an array instead of dict. The migration to the new format will happen automatically at the first startup. {pull}1703[1703]
 
 *Winlogbeat*
 
--- a/filebeat/crawler/prospector.go
+++ b/filebeat/crawler/prospector.go
@@ -17,7 +17,6 @@ type Prospector struct {
 	prospectorer  Prospectorer
 	spoolerChan   chan *input.FileEvent
 	harvesterChan chan *input.FileEvent
-	registrar     *Registrar
 	done          chan struct{}
 	states        *input.States
 	wg            sync.WaitGroup
@@ -28,14 +27,13 @@ type Prospectorer interface {
 	Run()
 }
 
-func NewProspector(cfg *common.Config, registrar *Registrar, spoolerChan chan *input.FileEvent) (*Prospector, error) {
+func NewProspector(cfg *common.Config, states input.States, spoolerChan chan *input.FileEvent) (*Prospector, error) {
 	prospector := &Prospector{
 		config:        defaultConfig,
-		registrar:     registrar,
 		spoolerChan:   spoolerChan,
 		harvesterChan: make(chan *input.FileEvent),
 		done:          make(chan struct{}),
-		states:        input.NewStates(),
+		states:        &states,
 		wg:            sync.WaitGroup{},
 	}
 
@@ -44,7 +42,6 @@ func NewProspector(cfg *common.Config, registrar *Registrar, spoolerChan chan *i
 	}
 
 	err := prospector.Init()
-
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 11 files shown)


filebeat/crawler/prospector.go
```diff
@@ -28,14 +27,13 @@ type Prospectorer interface {
...
-func NewProspector(cfg *common.Config, registrar *Registrar, spoolerChan chan *input.FileEvent) (*Prospector, error) {
+func NewProspector(cfg *common.Config, states input.States, spoolerChan chan *input.FileEvent) (*Prospector, error) {
 	prospector := &Prospector{
 		config:        defaultConfig,
-		registrar:     registrar,
 		spoolerChan:   spoolerChan,
 		harvesterChan: make(chan *input.FileEvent),
 		done:          make(chan struct{}),
-		states:        input.NewStates(),
+		states:        &states,
```
urso: the `states` parameter given to NewProspector is a copy of original one, but here we allocate the copied state onto heap. You want input.States to be shared or to be a copy in here?

filebeat/crawler/registrar.go
```diff
@@ -63,14 +64,84 @@ func (r *Registrar) Init() error {
...
+func (r *Registrar) LoadState() error {
+
+	// Check if files exists
+	_, err := os.Stat(r.registryFile)
+	if err != nil && !os.IsNotExist(err) {
+		return err
+	}
+
+	// Error means no file found
+	if err != nil {
```
urso: can file permissions make `err != nil`? File permission errors should not be ignored, as we won't be able to write the registry file in the future.

ruflin: We should probably catch that even early when we create the data directory? Currently not the case. 

You suggest in case we can't create a registry, we do not run filebeat.

filebeat/crawler/prospector.go
ruflin: It should be a copy as from here on, each prospector treats the state independent.

filebeat/crawler/registrar.go
tsg: I think we do exit if creating the data directory fails, but it's possible that the directory already exists with the wrong permissions.

+1 to stop filebeat in case the registry file cannot be written.

CHANGELOG.asciidoc
```diff
@@ -55,6 +55,7 @@ https://github.com/elastic/beats/compare/v5.0.0-alpha2...master[Check the HEAD d
 *Topbeat*
 
 *Filebeat*
+- The registry format was changed to an array instead of dict. The migration to the new format will happen automatically at the first startup. {pull}1703[1703]
```
tsg: +++ for doing the migration automatically!

filebeat/crawler/registrar.go
ruflin: It could also happen, that filebeat is restarted but does not have file write permission anymore, means registry file aready exist but filebeat was started last time by a different user. Should we also exit? I suggest yes.

I have to find an easy way to determine if we can write to the file / directory :-)

ruflin: @tsg @urso Would it be ok to do this in a second PR? The current behaviour is identical to what happened so far I think.

filebeat/crawler/prospector.go
urso: states is defined like this:

```
 // States handles list of FileState
 type States struct {
 	states []FileState
 	mutex  sync.Mutex
 }
```

The copy of 'states' is mostly shallow. Some code checkers (newer `go vet` should) will complain about copy on mutex. The underlying byte-buffer is shared when slice is copied. The `(*States).Update` method can override a file its state, if file is member of slice. Due to mutex being copied, any 2 go-routines attempting to updates a file state  for the same file might actually race.

urso:
LGTM

PR #1705, opened 2016-05-23
Export cpu usage in percentage by default

monicasarbu:
With this PR, I did the minimal changes that Topbeat and Metricbeat to export (by default) the CPU usage statistics in percentage instead of CPU ticks. The percentages are: 
- `cpu.iowait_p`, `cpu.idle_p`, `cpu.irq_p`, `cpu.nice_p`, `cpu.softirq_p`, `cpu.steal_p`, `cpu.system_p` and `cpu.user_p`.
- `proc.cpu.total_p`

If the user is interested in exporting the CPU usage in ticks , he/she needs to set the configuration option `cpu_usage_in_ticks` to true. The following values are exported in addition to the CPU usage in percentage:

- `cpu.iowait`, `cpu.idle`, `cpu.irq`, `cpu.nice`, `cpu.softirq`, `cpu.steal`
- `proc.cpu.user`, `proc.cpu.system`, `proc.cpu.total`
- `core.iowait`,  `core.idle`, `core.irq`, `core.nice`, `core.softirq`, `core.steal`

We don't need to change the Topbeat dashboard as the values that we are making optional, are not used in the dashboard.

cc-ed: @tsg @simianhacker @ruflin @andrewkroh


metricbeat/etc/fields.yml
```diff
@@ -704,28 +710,57 @@
...
+              description: >
+                The percentage of CPU time spent in wait (on disk).
+
             - name: irq
               type: integer
               description: >
                 The amount of CPU time spent servicing and handling hardware interrupts.
 
+            - name: irq_p
+              type: integer
```
monicasarbu: It should be float. I am fixing it.

metricbeat/docs/fields.asciidoc
```diff
@@ -1083,6 +1083,13 @@ type: integer
 The amount of CPU time spent on low-priority processes.
 
 
+==== system.core.nice_p
```
ruflin: Could we call it in metricbeat nice_pct: https://github.com/elastic/beats/blob/master/metricbeat/module/doc.go#L34 p seems for me too generic. But this is no blocker for this PR.

metricbeat/etc/beat.yml
```diff
@@ -55,4 +55,5 @@ metricbeat.modules:
   period: 2s
   processes: ['.*']
 
-
+  # if true, exports the CPU usage in ticks, together with the percentage values
+  cpu_usage_in_ticks: false
```
ruflin: I would call the option just `ticks` as it is already for the metricset cpu and core.

metricbeat/etc/fields.yml
```diff
@@ -704,28 +710,57 @@
...
+                The percentage of CPU time spent servicing and handling hardware interrupts.
+
+
             - name: softirq
               type: integer
               description:
                 The amount of CPU time spent servicing and handling software interrupts.
 
+            - name: softirq_p
+              type: integer
```
ruflin: same here: float

metricbeat/etc/fields.yml
```diff
@@ -704,28 +710,57 @@
...
                 The amount of CPU time spent servicing and handling hardware interrupts.
 
+            - name: irq_p
+              type: integer
+              description: >
+                The percentage of CPU time spent servicing and handling hardware interrupts.
+
+
             - name: softirq
               type: integer
```
ruflin: Nothing new in this PR, bue how large get these values? Should we go with long?

metricbeat/module/system/cpu/cpu.go
```diff
@@ -25,9 +25,21 @@ type MetricSet struct {
 // New is a mb.MetricSetFactory that returns a cpu.MetricSet.
 func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
 
+	config := struct {
+		CpuTicks bool `config:"cpu_usage_in_ticks"` // export CPU usage in ticks
```
ruflin: see comment above to call it just `ticks`

metricbeat/tests/system/test_system.py
```diff
@@ -3,11 +3,14 @@
...
 import metricbeat
 
-SYSTEM_CPU_FIELDS = ["idle", "iowait", "irq", "load", "nice", "softirq",
-                     "steal", "system", "system_p", "user", "user_p"]
+SYSTEM_CPU_FIELDS = ["idle_p", "iowait_p", "irq_p", "load", "nice_p",
+                     "softirq_p", "steal_p", "system_p", "user_p"]
 
-SYSTEM_CORE = ["core", "idle", "iowait", "irq", "nice", "softirq",
-                "steal", "system", "system_p", "user", "user_p"]
+SYSTEM_CPU_ALL_FIELDS = ["idle_p", "idle", "iowait_p", "iowait", "irq_p", "irq", "load", "nice_p", "nice",
```
ruflin: we can to this also at a later stage, but I was thinking if it would make sense to group the ticks together. `system.cpu.ticks.idle` and `system.cpu.ticks.steal`.

metricbeat/etc/beat.yml
ruflin: It seems like you updated this here directly? This should be updated in the system module and then you have to run `make collect`.

topbeat/docs/reference/configuration/topbeat-options.asciidoc
```diff
@@ -26,8 +26,10 @@ topbeat:
     filesystem: true
 
     # cpu usage per core, by default is false
-    cpu_per_core: false
+    core: false
```
ruflin: Is this a BC break?

topbeat/etc/fields.yml
```diff
@@ -8,6 +8,90 @@
...
+  fields:
+    - name: id
+      type: long
+      description: >
+       The id of the core X.
+
+    - name: user
+      type: long
+      description: >
+       The amount of CPU time spent in user space on core X.
```
andrewkroh: If this is an amount of time, then what are the units? I believe on linux this counter increments once per every 100 cycles (100 Hz) so each increment represents 10ms.

CHANGELOG.asciidoc
```diff
@@ -20,6 +20,7 @@ https://github.com/elastic/beats/compare/v5.0.0-alpha2...master[Check the HEAD d
 *Packetbeat*
 
 *Topbeat*
+- Export CPU stats per core in type=core documents and add `cpu_usage_in_ticks` option {pull}1705[1705]
```
ruflin: I suggest to split this up into two lines, as these are 2 changes (even though it is in on PR).

ruflin:
In general LGTM. Somme suggestions/comments:

* We should group the ticks together under the namespace `ticks`.
* Having ticks grouped together, allows us to call the current `user_p` just `user`. Otherwise I would call it `user_pct`, at least in metricbeat
* I'm not sure if you run the `make collect` script. Please check if that overwrites stuff again. 
* Rename the really long config option to just `ticks`.

CHANGELOG.asciidoc
monicasarbu: Agree

metricbeat/etc/beat.yml
monicasarbu: Sorry, I missed it. Maybe we should put config.yml directly in the module/<module-name> directory, so it's more visible to the users if it's so important.

topbeat/etc/fields.yml
urso: see `man sysconf` and search for ticks. Default value for linux is 100, but do not rely on it. Without knowing the frequency the tick counter is incremented, ticks is quite useless. Better would be some measure in seconds or milliseconds.

urso: Not a windows expert. This how you get the ticks on windows? https://msdn.microsoft.com/en-us/library/system.timespan.tickspersecond(v=vs.110).aspx

Seems to be 10ns (if I didn't miss-count the zeros).

Makes it hardly possible to compare raw tick values between linux and windows systems?

metricbeat/etc/fields.yml
monicasarbu: yes, Topbeat uses long already for the CPU usage values expressed in ticks.

metricbeat/module/system/cpu/cpu.go
monicasarbu: I called it `cpu_ticks`.

metricbeat/tests/system/test_system.py
monicasarbu: That would be a good idea, to group all the values under `ticks`.
cc-ed @tsg

topbeat/docs/reference/configuration/topbeat-options.asciidoc
monicasarbu: yes, it is, as the CPU details per core are placed under the `type=core` documents instead of `type=system` document.s

metricbeat/tests/system/test_system.py
tsg: Aren't "ticks" more like the measurement unit? In that case per the style guide you would have `system.cpu.idle.pct` and `system.cpu.idle.ticks`. In any case, it's probably better to do the changes in a future PR.

monicasarbu:
I updated the PR with the comments. 

@ruflin I kept `user_p` for now, for alpha3 as there are quite many changes for it.

monicasarbu:
jenkins, test it

metricbeat/tests/system/test_system.py
ruflin: @tsg agree on both (suggestion and future PR). This is even better the moving all ticks into a namespace "ticks". Also if you want to filter out `idle` in the example above, you just define `system.cpu.idle.*` and all idle information is removed.

topbeat/tests/system/config/topbeat.yml.j2
```diff
@@ -13,10 +13,11 @@ topbeat.procs: [
 
 # Statistics to collect (all enabled by default)
 topbeat.stats:
-  system: {{ "false" if system_stats == false else "true" }}
```
ruflin: Looks like this is 4 spaces now. Does it still work? Is there a need to change this?

ruflin:
LGTM

topbeat/tests/system/config/topbeat.yml.j2
monicasarbu: Yes, I think it's working. Adding more spaces at each line wasn't on purpose.

monicasarbu:
jenkins, test it

monicasarbu:
jenkins, test it

PR #1709, opened 2016-05-23
Udpate docs for new configs

tsg:
As part of #1417, this updates the docs to use the new conventions from the configuration files.

In addition:

* it "documents" more Packetbeat fields in `packetbeat.full.yml`
* it adds a bit more examples to the Filtering section
* it fixes the Metricbeat Reference page (which was containing Topbeat info)
* a few other fixes

As this has become quite a large PR, I left the individual commits (one per beat), so it is easier to review.
CC @dedemorton.


filebeat/docs/multiline.asciidoc
```diff
@@ -46,10 +46,9 @@ To consolidate these lines into a single event in Filebeat, use the following mu
 
 [source,json]
 -------------------------------------------------------------------------------------
-multiline:
-  pattern: '^[[:space:]]'
-  negate: false
-  match: after
+multiline.pattern: '^[[:space:]]'
```
ruflin: I was thinking if we should have the "correct" indentation for these sub fields, so if someone does a copy paste it is already with the right indentation?

packetbeat/packetbeat.full.yml
```diff
@@ -120,36 +199,165 @@ packetbeat.protocols.memcache:
...
+  # unlimited number of characters per document.
+  #max_doc_length: 5000
+
+  # If this option is enabled, the raw message of the request (`request` field)
+  # is sent to Elasticsearch. The default is false.
+  #send_request: false
+
+  # If this option is enabled, the raw message of the response (`response`
+  # field) is sent to Elasticsearch. The default is false.
+  #send_response: false
```
ruflin: I didn't even know all these options exist :-)

metricbeat/docs/modules/apache.asciidoc
```diff
@@ -18,13 +18,13 @@ here (add link). Below is an example of a configuration option:
 [source,yaml]
 ----
 metricbeat.modules:
-  - module: apache
-    metricsets: ["status"]
-    enabled: true
-    period: 1s
+- module: apache
```
ruflin: This is directly generated from the config.yml file. It looks like you edited it here directly?

ruflin:
Great effort. LGTM. Happy to merge this ASAP to prevent further conflicts and do small fixes in a second stage. WDYT?

metricbeat/docs/modules/apache.asciidoc
tsg: No, I edited the script this time :)

ruflin: Ah, right. Only the indentation changed, not the content. I was searching for the content changes :-)

packetbeat/docs/reference/configuration/packetbeat-options.asciidoc
```diff
@@ -336,30 +325,29 @@ sample configuration section:
...
-    hide_keywords: ["pass", "password", "passwd"]
+  # Uncomment the following to hide certain parameters in URL or forms attached
+  # to HTTP requests. The names of the parameters are case insensitive.
+  # The value of the parameters will be replaced with the 'xxxxx' string.
+  # This is generally useful for avoiding storing user passwords or other
+  # sensitive information.
+  hide_keywords: ["pass", "password", "passwd"]
 
-    # Uncomment the following to export a list of extra HTTP headers. By
-    default is none sent.
```
dedemorton: Word order is off here. Maybe say, "By default, no extra HTTP headers are sent"

packetbeat/docs/reference/configuration/packetbeat-options.asciidoc
```diff
@@ -336,30 +325,29 @@ sample configuration section:
...
-    send_headers: ["User-Agent", "Cookie", "Set-Cookie"]
+  # Uncomment the following to export a list of extra HTTP headers. By
+  default is none sent.
```
dedemorton: Word order is off here. Maybe say, "By default, no extra HTTP headers are sent"

packetbeat/docs/reference/configuration/packetbeat-options.asciidoc
```diff
@@ -336,30 +325,29 @@ sample configuration section:
...
+  send_headers: ["User-Agent", "Cookie", "Set-Cookie"]
 
-    # Uncomment the following to export Cookie or Set-Cookie headers. By
-    # default is false.
-    split_coookie: true
+  # Uncomment the following to export Cookie or Set-Cookie headers. By
+  # default is false.
```
dedemorton: Missing subject... Would say, "The default is false"

packetbeat/docs/reference/configuration/packetbeat-options.asciidoc
```diff
@@ -305,18 +295,17 @@ The `dns` section specifies configuration options for the DNS protocol. The DNS
...
 
-    # include_authorities controls whether or not the dns.authorities field
-    # (authority resource records) is added to messages.
-    # Default: false
-    include_authorities: true
-    # include_additionals controls whether or not the dns.additionals field
-    # (additional resource records) is added to messages.
-    # Default: false
-    include_additionals: true
+  # include_authorities controls whether or not the dns.authorities field
```
dedemorton: General comment/question: We're a bit inconsistent about including comment blocks in example configs. I'm not saying that we have to be consistent, but it seems a bit arbitrary sometimes. Curious to know if other reviewers think that we should standardize on an approach.

packetbeat/docs/reference/configuration/packetbeat-options.asciidoc
dedemorton: Elsewhere, I see split_cookie described this way: "If the Cookie or Set-Cookie headers are sent, this option controls whether they are split into individual values."  This seems more accurate and easier to follow than the description given here.

filebeat/docs/multiline.asciidoc
dedemorton: I'm also wondering if it would help to show more of the config file so users can see where  sub-fields are defined (we can leave out sections and use ..., but some context might help). I'm not suggesting you should fix this now, but if you think it's a good idea, let me know, and I'll create an issue to track the work.

dedemorton:
Doc content changes LGTM. Nice work! BTW, I think we should say more about conditionals in the section about generic filtering (it appears in topics like [this one](https://www.elastic.co/guide/en/beats/filebeat/master/filtering-data.html)). @tsg Do you want me to add something to that section?

dedemorton:
@tsg Also noticed that drop_event is not covered in libbeat/docs/filteringconfig.asciidoc. Want me to add it?

filebeat/docs/multiline.asciidoc
tsg: I agree. The good news is that this only happens for the filebeat prospectors. For everything else, after the dot notation change, the config file sample stand for themselves.

tsg:
@dedemorton I agree about conditionals, I already told @monicasarbu about the need for more docs/examples. If you can jump in to help, that would be highly appreciated :-).

packetbeat/docs/reference/configuration/packetbeat-options.asciidoc
tsg: That's a good question, I'm not sure. I think it most cases we could go without comments as the required explanations usually follow anyway. Would be worth it to have a style guide for docs that we can all follow? Might be worth a discussion at least.

PR #1711, opened 2016-05-24
Move all conversions to helper function

ruflin:


tsg:
There seems to be a test failure for the zookeeper module.

metricbeat/helper/conversion.go
```diff
@@ -0,0 +1,85 @@
...
+	if err != nil {
+		logp.Err("Error converting param to float: %s", key)
+		value = 0.0
+	}
+
+	return value
+}
+
+// ToInt converts value to int. In case of error, returns 0
+func ToInt(key string, data map[string]string) int {
```
tsg: Shouldn't we work on int64 by default (long in Elasticsearch)? I think longs will be pretty common in these datasets.

Btw, I recently learned that there's not much benefit in declaring numbers as integers in Elasticsearch, since integers are actually longs behind the scene (i.e. they take the exact same storage space).

tsg: This is kind of related to the uint64 issue that @andrewkroh raised recently.

ruflin: Very interesting, didn't know about the integer part. This means we could almost always use long without any downsides, correct?

ruflin: @tsg I switched now everything to int64

tsg: Yes, that's my understanding as well, we can just always use longs.

ruflin:
jenkins, test it

PR #1715, opened 2016-05-24
logstash output module base on go-lumber

urso:
Update logstash client to be based on go-lumber. Unit tests for logstash output will use lumberjack server implementation from go-lumber, simplifying test code quite a bit. Custom incomplete lumberjack server has been removed from output plugin.

json encoder used by logstash output is somewhat hand-crafted supporting primitive types without  reflection. It's basically an extension of a bugfix for logstash output. Json support will be generalized   andmoved into libbeat/common in another PR.

Testing [generatorbeat](https://github.com/urso/generatorbeat) with go-lumber test server throughput increased noticably on my dev machine after introducing custom json encoder:

   | no zlib    |   zlib  
--- | --- | ---
json.Encode            |    52.5k   |   47.5k 
json custom | 70.2k | 60.1K


ruflin:
LGTM

PR #1718, opened 2016-05-24
Set version to 5.0.0-alpha3

tsg:
To be closer to what we plan for the unified release process, I'm setting the
version in advance. The artifacts will be named 5.0.0-alpha3-SNAPSHOT during
testing but the version in code is 5.0.0-alpha3. At the time of the release, we
rename the artifacts to remove SNAPSHOT, without having to rebuild.

### DIFF
--- a/dev-tools/packer/version.yml
+++ b/dev-tools/packer/version.yml
@@ -1,2 +1,1 @@
-
 version: "5.0.0-alpha3"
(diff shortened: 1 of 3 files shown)


dev-tools/packer/version.yml
```diff
@@ -1 +1,2 @@
-version: "5.0.0"
+
```
ruflin: Is there a reason for the empty line?

tsg: No, it was generated like that by `dev-tools/set_version`. I fixed it now in the script.

PR #1719, opened 2016-05-24
Use underscores in RPM version metadata

tsg:
The ~ doesn't work with older RPM versions, so the new convention is that we
use - everywhere except for the RPM metadata which uses _. All filenames still
use dashes.


tsg:
jenkins, package it

tsg:
jenkins, package it

PR #1722, opened 2016-05-25
Add alpha3 snapshots for testing

ruflin:
Add script to simplify common parts between elasticsearch docker images

### DIFF
--- a/testing/environments/alpha2.yml
+++ b/testing/environments/alpha2.yml
@@ -4,7 +4,7 @@
 elasticsearch:
   build: ./docker/elasticsearch
   dockerfile: Dockerfile-5.0.0-alpha1    # still using alpha1 because alpha2 doesn't work fine in docker
-  command: elasticsearch -Ees.network.host=0.0.0.0 -Ees.discovery.zen.minimum_master_nodes=1
+  command: elasticsearch -Enetwork.host=0.0.0.0 -Ediscovery.zen.minimum_master_nodes=1
 
 logstash:
   build: ./docker/logstash
--- a/testing/environments/docker/elasticsearch/Dockerfile-5.0.0-alpha1
+++ b/testing/environments/docker/elasticsearch/Dockerfile-5.0.0-alpha1
@@ -1,36 +1,13 @@
 FROM java:8-jre
 
-# grab gosu for easy step-down from root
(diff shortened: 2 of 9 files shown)


testing/environments/docker/elasticsearch/Dockerfile-5.0.0-alpha1
```diff
@@ -1,37 +1,14 @@
...
-	&& chmod +x /usr/local/bin/gosu
-
-RUN apt-key adv --keyserver ha.pool.sks-keyservers.net --recv-keys 46095ACC8548582C1A2699A9D27D666CD88E42B4
-
 ENV ELASTICSEARCH_MAJOR 5.0
 ENV ELASTICSEARCH_VERSION 5.0.0-alpha1
 
-RUN wget https://download.elasticsearch.org/elasticsearch/release/org/elasticsearch/distribution/deb/elasticsearch/5.0.0-alpha1/elasticsearch-5.0.0-alpha1.deb
+COPY setup.sh /
+RUN bash setup.sh https://download.elasticsearch.org/elasticsearch/release/org/elasticsearch/distribution/deb/elasticsearch/5.0.0-alpha1/elasticsearch-5.0.0-alpha1.deb
```
tsg: Do we need to keep alpha1 and alpha2 around?

testing/environments/alpha2.yml
```diff
@@ -4,7 +4,7 @@
 elasticsearch:
   build: ./docker/elasticsearch
   dockerfile: Dockerfile-5.0.0-alpha1    # still using alpha1 because alpha2 doesn't work fine in docker
-  command: elasticsearch -Ees.network.host=0.0.0.0 -Ees.discovery.zen.minimum_master_nodes=1
+  command: elasticsearch -Enetwork.host=0.0.0.0 -Ediscovery.zen.minimum_master_nodes=1
```
tsg: I'm not sure, but I think alpha2 doesn't work because of the system checks. Did you try it? If it doesn't work, we could just remove the files.

ruflin: @tsg Agree, we should just remove alpha1 and alpha2 as we don't need it anymore. I can do it in this PR here: https://github.com/elastic/beats/pull/1724

PR #1726, opened 2016-05-25
Clearly mark Metricbeat as alpha in docs

tsg:
* Add (alpha) in the title, so that shows up on the main index page
* Add a notice that the docs are for a non-GA version. This is done
  via a temporary `index-docinfo.xml` file at Clint's suggestion.
* Set version to alpha3 already (the version bump for the other
  products will come later)

CC @dedemorton and @ruflin. I think with these changes we can push Metricbeat to the live website. What do you say?

### DIFF
--- a/metricbeat/docs/index-docinfo.xml
+++ b/metricbeat/docs/index-docinfo.xml
@@ -1,0 +1,7 @@
+<?page_header <b>PLEASE NOTE:</b> These docs are for a pre-GA version of Metricbeat, {version}.?>
+
+<abstract>
+    <simpara role="page_header">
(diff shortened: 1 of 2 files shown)


ruflin:
LGTM.

Pushing it sounds good to me. @dedemorton mentioned yesterday that she would integrate https://github.com/elastic/beats/blob/master/metricbeat/README.md into the docs for me. Do you have any timeframe on this?

This would mean that docs get always updated with the most recent version from master, correct? Exposing it will also provide us feedback for what is still broken.

tsg:
Yes, every changes we do in master will show up on the web site in ~30 minutes. So we can easily iterate over the docs, I'm not too worried about making them public.

PR #1730, opened 2016-05-25
Removed (alpha) from the Metricbeat docs title

tsg:
This partially reverts #1726. After all, the text on the index page
comes from conf.yaml, and the notice is enough on the main Metricbeat
docs page.


tsg:
The `(alpha)` is added here instead.

PR #1732, opened 2016-05-25
Add system/network MetricSet for network IO

andrewkroh:
An event is generated for each network interface. This MetricSet is available on
Darwin (aka OSX), Windows, Linux, and FreeBSD.


metricbeat/module/system/network/_beat/docs.asciidoc
```diff
@@ -0,0 +1,23 @@
...
+
+By default metrics are reported from all network interfaces. To select which
+interfaces that metrics are reported from, use the `interfaces` configuration
+options. The value must be an array of interface names. For example:
+
+[source,yaml]
+----------------------------
+metricbeat.modules:
+- module: system
+  metricsets: [network]
```
ruflin: In most other examples we use `metricset: ["network"]`. I'm happy with both. Which one should we pick?

metricbeat/module/system/network/network.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	return events, nil
+}
+
+func ioCountersToMapStr(counters net.IOCountersStat) common.MapStr {
+	return common.MapStr{
+		"name":         counters.Name,
+		"bytes_sent":   counters.BytesSent,
+		"bytes_recv":   counters.BytesRecv,
```
ruflin: I suggest to transform this to:

```
received.bytes
received.packets
sent.bytes
sent.packets
```

metricbeat/module/system/network/network.go
```diff
@@ -0,0 +1,95 @@
...
+		"packets_sent": counters.PacketsSent,
+		"packets_recv": counters.PacketsRecv,
+		"errin":        counters.Errin,
```
ruflin: Not sure if here it would make seen to go with:

```
in.error
in.drop
out.error
out.drop
```

ruflin:
Nice 💯 LGTM. I added some small notes for the field naming.

ruflin:
@andrewkroh Merged this one as I have to go over the namins of the system module anyways.

PR #1733, opened 2016-05-25
Update Metricbeat examples in docs

monicasarbu:
Update examples from the system module of Metricbeat.

cc-ed @andrewkroh @ruflin @dedemorton


metricbeat/docs/modules/system/core.asciidoc
```diff
@@ -18,28 +18,30 @@ Below is an example document generated by this metricset.
...
-        "steal":0,
-        "system":7499682,
-        "system_p":0,
-        "user":9354158,
-        "user_p":0
-    },
-    "type":"metricsets"
+  "@timestamp": "2016-05-25T20:49:40.557Z",
+  "beat": {
+    "hostname": "mar.local",
```
andrewkroh: I propose we use the same host in each of these examples. How about `host.example.com`?

andrewkroh:
Thanks for updating these. LGTM

monicasarbu:
@andrewkroh I replaced all the hostnames with `host.example.com`.

ruflin:
LGTM. Thanks for the updates. I'm trying to automate this in the future. Here is a first example that works: https://github.com/elastic/beats/blob/master/metricbeat/module/system/cpu/cpu_test.go#L12

Here is the code that does it and what test commands have to be used: https://github.com/elastic/beats/blob/master/metricbeat/mb/testing/data_generator.go#L15

In a later PR here we should also update the default host: https://github.com/elastic/beats/blob/master/metricbeat/mb/testing/data_generator.go#L38

PR #1738, opened 2016-05-26
Add script to generate MetricSet

ruflin:
Adding a MetricSet requires a create several files with a predefined file structure for automation. To simplify the task of creating a MetricSet and make sure the correct structure is used, this task is now automated with a script. The script can be called with `make create-metricset MODULE=modulename METRICSET=metricsetname`. This generates all files including template, config and docs. After `make imports` must be called to update the import list and `make collect` to aggregate all the new data. The MetricSet can directly be compiled and run with the default content.

This script can also be used in the metricbeat-generator, means someone creating multiple metricsets has also the tooling available. In addition, this simplifies the metricbeat-generator as most of the templates which change frequently are in the beats repository.

As the templates used for the generation are the same as shown in the developer guide, it would be nice to also import these templates to make sure the docs are always up-to-date.

### DIFF
--- a/filebeat/tests/system/test_crawler.py
+++ b/filebeat/tests/system/test_crawler.py
@@ -6,6 +6,7 @@ import codecs
 import os
 import time
 from nose.plugins.skip import Skip, SkipTest
+import shutil
 
 # Additional tests to be added:
 # * Check what happens when file renamed -> no recrawling should happen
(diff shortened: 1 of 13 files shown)


filebeat/tests/system/test_crawler.py
```diff
@@ -270,15 +271,19 @@ def test_file_disappear_appear(self):
...
         )
         os.mkdir(self.working_dir + "/log/")
 
         testfile = self.working_dir + "/log/test.log"
+        testfilenew = self.working_dir + "/log/hiddenfile"
         file = open(testfile, 'w')
 
+        # Creates testfile now, to prevent inode reuse
+        open(testfilenew, 'a').close()
+
```
tsg: Are these filebeat changes related? Or just fixing a flaky test?

ruflin: Fixing a flaky test. Not related. But it started to pop up frequently in the last 2 days (not sure why). The issue was inode reuse.

tsg: Yes, I saw it as well, thanks for fixing it so quickly!

tsg:
LGTM, will wait for appveyor as well due to the Filebeat change.

PR #1743, opened 2016-05-26
Fixing two fields that contain strings and not integers

jjfalling:
Small fix to two fields in the mapping that have the wrong type.


ruflin:
LGTM. Waiting for green.

PR #1749, opened 2016-05-27
Clean up core fields

ruflin:
* Introduce missing ticks config option
* Add auto generation data test

I do the data conversion now in Metricbeat to decouple it from Topbeat field structure.


metricbeat/module/system/core/core.go
```diff
@@ -45,10 +56,47 @@ func (m *MetricSet) Fetch() ([]common.MapStr, error) {
 	cores := []common.MapStr{}
 
 	for core, stat := range cpuCoreStat {
-		coreStat := m.cpu.GetCpuStatEvent(&stat)
```
monicasarbu: Why didn't you do the changes in the `GetCpuStatEvent` function? So, `system` type doc can benefit from it.

PR #1752, opened 2016-05-27
Leave only the system module enabled in Metricbeat

tsg:
So that Metricbeat (default) == Topbeat

Also changed the default interval to 10s. The other modules are
left commented out in the full config and removed completely from the short configs.


monicasarbu:
LGTM, thanks @tsg

PR #1761, opened 2016-05-27
Fix multiple dots in field names in template

ruflin:


tsg:
I'm kind of surprised that this didn't happen automatically via the dedot function. Or is that not called on these?

tsg:
Merging without tests so @monicasarbu can work on dashboards.

PR #1767, opened 2016-05-30
Stop following symlinks.

### LINKED ISSUE #1686: Improve symlink handling in Filebeat
Currently Filebeat treats symlink as normal files. In case a file appears in the glob as symlink and file, the content is read twice. The following changes should be made:

* [ ] By default symlinks should be skipped (BC break)
* [ ] Introduce a configuration option to enable symlinks following
  * Make sure symlinks work properly with overwriting symlinks to new file

Is the second option needed?

For reference also see: https://discuss.elastic.co/t/filebeat-fails-to-harvest-if-a-file-and-a-symlink-to-that-file-is-in-the-same-directory/49743/3

ruflin:
Previously symlinks were followed. This had the consequence if a symlink and the file itself existed, the file was read twice. Now symlinks are not followed anymore.

This closes #1686

### DIFF
--- a/filebeat/crawler/prospector_log.go
+++ b/filebeat/crawler/prospector_log.go
@@ -78,12 +78,12 @@ func (p *ProspectorLog) getFiles() map[string]os.FileInfo {
 				continue
 			}
 
-			// Stat the file, following any symlinks.
 			fileinfo, err := os.Lstat(file)
 			if err != nil {
 				logp.Debug("prospector", "stat(%s) failed: %s", file, err)
 				continue
 			}
+			// Check if file is symlink
 			if fileinfo.Mode()&os.ModeSymlink != 0 {
 				logp.Debug("prospector", "File %s skipped as it is a symlink.", file)
 				continue
--- a/filebeat/crawler/prospector_log.go
+++ b/filebeat/crawler/prospector_log.go
@@ -79,9 +79,13 @@ func (p *ProspectorLog) getFiles() map[string]os.FileInfo {
 			}
 
 			// Stat the file, following any symlinks.
-			fileinfo, err := os.Stat(file)
+			fileinfo, err := os.Lstat(file)
 			if err != nil {
 				logp.Debug("prospector", "stat(%s) failed: %s", file, err)
+				continue
+			}
+			if fileinfo.Mode()&os.ModeSymlink != 0 {
+				logp.Debug("prospector", "File %s skipped as it is a symlink.", file)
 				continue
 			}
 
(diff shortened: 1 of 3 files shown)


filebeat/crawler/prospector_log.go
```diff
@@ -79,11 +79,15 @@ func (p *ProspectorLog) getFiles() map[string]os.FileInfo {
 			}
 
 			// Stat the file, following any symlinks.
```
andrewkroh: The comment here is out of date, right? `os.Lstat` doesn't follow symlinks.

ruflin: Yes, will remove it.

filebeat/tests/system/test_prospector.py
```diff
@@ -514,3 +514,45 @@ def test_close_older_file_rotation_and_removal(self):
...
+        testfile = self.working_dir + "/log/test-2016.log"
+        symlink_file = self.working_dir + "/log/test.log"
+
+        # write first line
+        with open(testfile, 'a') as file:
+            file.write("Hello world\n")
+
+        if os.name == "nt":
+            import win32file
+            win32file.CreateSymbolicLink(symlink_file, testfile, 1)
```
andrewkroh: The link is to a file so the last parameter should be 0. https://msdn.microsoft.com/en-us/library/windows/desktop/aa363866(v=vs.85).aspx

Hopefully that will fix the issue with 7z on AppVeyor.

djschny:
Can we add the option for a fail fast option so that if the path provided in the config is a symlink, that it fails to startup so that way a person knows right away of the issue?

djschny:
Alternatively why not default to it being disabled but allowing it to be enabled if one chooses.

ruflin:
@djschny The plan was to introduce a config option as soon as we have a request for it as we were not sure if this is something even needed. See https://github.com/elastic/beats/issues/1686#issuecomment-222452375 Is following symlink something you need?

About the "fail fast". This is a little bit more complex as the symlink could also appear when Filebeat is already running and there is not difference in "first scan" of file and all the follow up scans. I prefer here the config option as I think it is more predictable and is closer to the principle we have so far. If one file is not readable, all the other files are still read.

djschny:
Cool, thanks @ruflin, I missed that comment linked issue. Sounds good.

In regards to "fail fast" I don't think people would care much about the worry about if it changed "out from under" Filebeat, but if you don't have startup check code in place, I can understand how this would be a heavier implementation than what may be viewed externally. Thanks for fielding my comment.

PR #1773, opened 2016-05-30
Add more conceptual info to metricbeat docs

dedemorton:
@ruflin Here's the first pass at adding content from the readme. I think it works pretty well (decided to let the build break this section into separate topics because it was rather long). 

I don't think the event structure is quite right (and it should probably show where the actual metrics data is returned, imho).


ruflin:
@dedemorton Can you directly remove the README.asciidoc file or at least the content that went into other files?

PR #1778, opened 2016-05-31
Fix 'print' and 'file' for python 3.5 scripts in libbeat

radoondas:
This PR fixes (in my case) issues with python scripts. I was unable to run 'make update' unless I fixed scripts the way as in commit. Environment created in my beat was with python3.5.
Is this a valid fix?

### DIFF
--- a/libbeat/scripts/generate_fields_docs.py
+++ b/libbeat/scripts/generate_fields_docs.py
@@ -76,7 +76,7 @@ grouped in the following categories:
 
     # fields file is empty
     if docs is None:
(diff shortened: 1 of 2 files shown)


ruflin:
I think most of us running everything in Python 2.7.x. It would be nice if the above fixes make it compatible with both version. I will probably have to test this one manually :-)

radoondas:
I thought so. 
I still noticed that there were in one 'print' call parenthesis and in the other not. And the same with 'file' vs 'open' call. I will be happy if you test it when you have time. :)

ruflin:
Seems to work fine under python 2.7. Merging.

Issue #1791, opened 2016-06-03
Move configuration options up in the doc navigation hierarcy


dedemorton:
The current structure of the reference material buries the config information under a reference container. The following changes are required across the Beats doc:

- Move these topics/container topics to appear under  Configuring Beatname:
YAML Tips and Gotchas
Regular Expression support 
Configuration Options container

- Remove the Reference container topic

PR #1794, opened 2016-06-03
Restructure navigation to move config options up

dedemorton:
First pass at restructuring topics for issue #1791 

To what the files will look like in the build, go here (content will be available until this PR is closed):

UPDATED: 06.21.16
https://packetbeatrestructure.firebaseapp.com
https://winlogbeatrestructure.firebaseapp.com
https://filebeatrestructure.firebaseapp.com
https://metricbeatreorg.firebaseapp.com/


dedemorton:
I need to do more work here because the book parts for the module section of metricbeat cause the topics about security and exported fields to be nested below modules. I'll submit a fix soon, but don't merge until I've done that.

dedemorton:
@elastic/beats I've updated the doc builds (links above) to reflect the high-level structural changes. WDYT?

ruflin:
* I like the new config changes. I would put the configs directly after Getting Started, as I think that is much more important than exported fields
* Do we need the overview? Isn't that part of getting started? Or could it be part directly in index.html (landing page). It feels like way too important for not having too much information.

+1 on the changes.

dedemorton:
This should be ready to review now. I've updated the files on Firebase to reflect the latest changes.
@ruflin To move the "Exported Fields" topic, I had to update the script for generating the doc because the topic needs to be a part (which changes the heading levels...arggg). Please make sure the changes look good (not sure I took the best tack, but it works). I like the overview topic where it is and would prefer to keep it there for now.

urso:
Nice. Restructuring looks very good. 

The Command line options and packetbeat directory layout are under 'Getting Started'. Don't think that's a good place for them.

I like the 'How Metricbeat works' section. Would be nice to have this kind of overview for all beats (maybe even with schematic overview). Maybe putting command line options and directory layout in this section would make sense?

dedemorton:
@urso I def like the idea of adding "how this works" sections to all of the Beats (especially for the ones where users keep asking us for more detail about how they work). Not sure about moving the CLI and directory layout stuff into that section, though. If we had a dedicated section about installation, it would make sense to have those topics there, but since the GS provides the installation info, I think those two topics logically belong there. And since they are at the end, I don't think they distract from the getting started experience.

filebeat/docs/index.asciidoc
```diff
@@ -21,6 +21,10 @@ include::../../libbeat/docs/shared-directory-layout.asciidoc[]
 
 include::./configuring-howto.asciidoc[]
 
+include::../../libbeat/docs/yaml.asciidoc[]
+
+include::../../libbeat/docs/regexp.asciidoc[]
+
 include::./filebeat-filtering.asciidoc[]
 
 include::./multiline.asciidoc[]
```
ruflin: Not 100% sure about the ordering here, as somehow further up means more important. I see the logic that regexp sohuld be understood to use mutiline, but I think people will first look for multiline and then jump to regexp. Same for yaml?

metricbeat/docs/index.asciidoc
```diff
@@ -12,10 +12,16 @@ include::./gettingstarted.asciidoc[]
...
 
+include::../../libbeat/docs/shared-directory-layout.asciidoc[]
+
 include::./how-metricbeat-works.asciidoc[]
 
 include::./configuring-howto.asciidoc[]
 
+include::../../libbeat/docs/yaml.asciidoc[]
+
+include::../../libbeat/docs/regexp.asciidoc[]
```
ruflin: See comment filebeat about regexp.

ruflin:
LGTM. Only left some minor comments that could also be addressed / discussed later. 

For command line options. I think there are 2 parts here: The very common onces like `-c` and the more complex ones. Having in the getting started guide `-c` seems to be important, but all the others not sure where to place (but agree not Getting started). It somehow fits more in a section "Running & Debugging Beats" or something like that?

dedemorton:
@elastic/beats I think/hope this is ready to merge now. I've updated the builds at the links above so you can see the changes. The latest round of changes includes:
- moving the topics about YAML and Regexp support lower in the nav (to reflect lower importance)
- adding the name of the config file to each topic in the reference section (I realize this seems kind of repetitive in places, but it helps users who jump to one of the reference topics--or sections--directly from a search...just gives them a bit more context for understanding where the configuration is defined.)
- moved the previously shared topic about runoptions to the Packetbeat docs.
I left all of these as separate commits. Feel free to squash the commits.

I'm not going to mess with the CLI and directory structure topics at the moment because I have another issue open to track changes to the getting started.

PR #1800, opened 2016-06-06
Remove trailing slash from apache/nginx configs

### LINKED ISSUE #1779: Nginx/Apache default configuration doesn't work
In the default config for nginx/apache we have:

```
  hosts: ["http://127.0.0.1/"]
```

That trailing `/` causes issues, because the code then ignores the path configured later. We should either remove that `/` from the configuration or change the code to ignore the trailing slash.

tsg:
The issue with the trailing slash is that the code considers that
a "path", so the real path is ignored when configured.

Fixes #1779.


ruflin:
LGTM

PR #1802, opened 2016-06-06
Add generation script for packer files

ruflin:
These script can be used in all community beats to generate the dev-tools folder with the docker files. This allows to also use these scripts in the generator in a second step. This simplifies getting started with a beat (less files) but still offers all the possibilities. In addition, it allows us to update / improve these files over time and external beats can fetch the updates again through overwriting if needed.

Dev-tools folder will be removed from the templates after this PR and https://github.com/elastic/beats/pull/1799 are merged.

### DIFF
--- a/libbeat/scripts/create_packer.py
+++ b/libbeat/scripts/create_packer.py
@@ -1,0 +1,71 @@
+import os
+import argparse
+
+# Creates a new metricset with all the necessary file
+# In case the module does not exist, also the module is created
+
+
+def generate_packer(es_beats, abs_path, beat, beat_path, version):
+
+    # create dev-tools/packer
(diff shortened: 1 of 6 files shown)


libbeat/scripts/create_packer.py
```diff
@@ -0,0 +1,71 @@
+import os
+import argparse
+
+# Creates a new metricset with all the necessary file
```
andrewkroh: Copy/paste issue with this comment?

ruflin: Yes, fixed

tsg:
CC-ing @cyrilleverrier as he was involved with this part.

PR #1821, opened 2016-06-08
packetbeat: nfs: fix first-class operation code in compound

kofemann:
NFS4 compound operations can be constructed from multiple
single requests. Nevertheless, each compound request as a
main operation.

current code failing to detect main operation if it was in
the middle of compound call.

Signed-off-by: Tigran Mkrtchyan <tigran.mkrtchyan@desy.de>


ruflin:
jenkins, test it

packetbeat/protos/nfs/nfs4.go
```diff
@@ -167,18 +167,20 @@ func (nfs *Nfs) eatData(op int, xdr *Xdr) {
 
 func (nfs *Nfs) getV4Opcode(xdr *Xdr) string {
 
+	// did we found a main operation opcode?
```
urso: shouldn't it say `did we find ...` ?

packetbeat/protos/nfs/nfs4.go
```diff
@@ -167,18 +167,20 @@ func (nfs *Nfs) eatData(op int, xdr *Xdr) {
 
 func (nfs *Nfs) getV4Opcode(xdr *Xdr) string {
```
urso: Do we return the last seen name on purpose or do we want to return 'ILLEGAL' if no matching op has been found? Some test with 'ILLEGAL' opcode would be nice.

proposal for returning 'ILLEGAL' if no op has been found:

```
func (nfs *Nfs) getV4Opcode(cdr *Xdr) string {
    count := int(xdr.getUint())
    for i := 0; i < count; i++ {
        op := int(xdr.getUInt())
        name, ok := nfs_opnum4[op]
        if !ok {
            return fmt.Sprintf(...)
        }

        switch op {
          case ...:
            return name
          default:
            nfs.eatData(op, xdr)
        }
    }

  return "ILLEGAL"
}
```

kofemann: we do return last seen on purpose. Here it the why. Compound request group multiple nfs operations into a sine request. For example:

PUTFH + READDIR + GETATTR

can be part of a single request. In this example READDIR is the main operation, triggered by **ls** command.  On an other hand in:

PUTFH + GETATTR

GETATTR is the main operation, triggered by stat call.

The whole 'main operation' idea is based on fact, that clients more or less combine first-class operations, like OPEN, WRITE, READDIR with some less important opertions - GETATTR, PUTFH and so on. Nevertheless, 'second class' operations can by them self be a 'main' operation in a different context. The function getV4Opcode tries to guess main purpose of the request. I will think how to restructure it to make it more obvious.

urso: Thanks for the explanation. Having a function `findV4MainOp` returning the opcode (not required, but makes intend more clear) + some comment explaining why (similar the text you just wrote) should be sufficient. It's important for maintainers not to rewrite the function by accident.

kofemann:
makes sense, updated.

kofemann:
@ruflin are there more changes required?

packetbeat/protos/nfs/nfs4.go
```diff
@@ -165,20 +165,37 @@ func (nfs *Nfs) eatData(op int, xdr *Xdr) {
...
+//
+// PUTFH + READDIR + GETATTR
+//
+// READDIR is the main operation. while in
+//
+// PUTFH + GETATTR
+//
+// GETATTR is the main operation.
+//
+//  This  function tries to guess main purpose of the request. If not found, then the last operation is used.
```
urso: Very nice explanation. Thanks!!!

some minor details:
- I don't see the caller of getV4Opcode being changed. Does it compile?
- spacing on line 170 and 172 seems to be a little off
- godoc convention kinda mandates putting the function and short description first (before going into detail):
```
findV4MainOpcode finds the main operation in a compound call. If no main operation can be found, the last operation in compound call is returned.

Compount requests group ...
```

You can remove the last line `This function tries...`, as this information should be present at beginning of comment (see my sample).

packetbeat/protos/nfs/nfs4.go
```diff
@@ -165,20 +165,37 @@ func (nfs *Nfs) eatData(op int, xdr *Xdr) {
 	}
 }
 
-func (nfs *Nfs) getV4Opcode(xdr *Xdr) string {
+// Find main operation in a compound call.
+//
+//  Compound request group multiple nfs operations into a single request. Nevertheless, all compound requests are
+// triggered by  end-user activity, line 'ls', 'open', 'stat' and IO calls. Depending on which operations are combined
+//  tougher the main operation can be different. For example, in compound:
```
urso: tougher?

packetbeat/protos/nfs/nfs4.go
```diff
@@ -165,20 +165,37 @@ func (nfs *Nfs) eatData(op int, xdr *Xdr) {
 	}
 }
 
-func (nfs *Nfs) getV4Opcode(xdr *Xdr) string {
+// Find main operation in a compound call.
+//
+//  Compound request group multiple nfs operations into a single request. Nevertheless, all compound requests are
```
urso: Compount requests group ?

packetbeat/protos/nfs/nfs4.go
urso: Question: Is there a chance of having multiple 'main' operations in one compound request?

kofemann: findV4MainOpcode is called in nfs.go:29, it compiles :)

kofemann: Technically, one can build a compound with READDIR and WRITE in it. In practice, nfs clients don't do that, as all nfs requests triggered by  user activities, and there are no posix call that will trigger readdir+write. Well, synthetic clients can....but they used only for server functional testing.

packetbeat/protos/nfs/nfs4.go
```diff
@@ -165,20 +165,36 @@ func (nfs *Nfs) eatData(op int, xdr *Xdr) {
 	}
 }
 
-func (nfs *Nfs) getV4Opcode(xdr *Xdr) string {
+// findV4MainOpcode finds the main operation in a compound call. If no main operation can be found, the last operation
+// in compound call is returned.
+//
+// Compound requests group multiple nfs operations into a single request. Nevertheless, all compound requests are
+// triggered by end-user activity, line 'ls', 'open', 'stat' and IO calls. Depending on which operations are combined
```
andrewkroh: `s/line/like/`

packetbeat/protos/nfs/nfs4.go
```diff
@@ -165,20 +165,36 @@ func (nfs *Nfs) eatData(op int, xdr *Xdr) {
...
+// tougher the main operation can be different. For example, in compound:
```
andrewkroh: `s/tougher/together/` or just remove the word.

kofemann: fixed

packetbeat/protos/nfs/nfs4.go
kofemann: fixed

urso:
Thanks for the fix and all the patience!

PR #1824, opened 2016-06-09
Switch squeeze image to use archive

tsg:
It was set to use snapshots as a temporary (heh) workaround for the
time when squeeze got removed from the main repos but wasn't yet added
to the archive.


tsg:
jenkins, package it

PR #1828, opened 2016-06-09
Fix logstash default bulk max size

### LINKED ISSUE #1662: logstash bulk_max_size only 1024 by default in 1.x releases
by default bulk_max_size for logstash is still 1024, but should be 2048.

Workaround: Set bulk_max_size for logstash output plugin.

urso:
Resolves #1662


ruflin:
LGTM. 

@urso see second comment in #1662 You also want to add it here?

urso:
Oh, thanks. Yeah, will introduce it

ruflin:
@urso ok, thanks. I removed the review label so it doesn't get merged by accident.

urso:
@ruflin done

ruflin:
@urso not sure if we have to forward port this?

PR #1830, opened 2016-06-10
Modify metricset generator to be countbeat

ruflin:
Like this the beat generated by beats generator and metricset generator is the same.
This also gives an example on how state should be handled in a metricset.


ruflin:
@tsg @dedemorton It would be nice if we could just merge this one and afterwards @dedemorton I think the Metricbeat docs are ready for a full review including the Developer Guide.

PR #1421, opened 2016-04-19
Fix handling of @timestamp and type from the JSON decoder

tsg:
If `json.keys_under_root` and `json.overwrite_keys` are set, and the keys
`@timestamp` or `type` show up in the decoded JSON, they can cause issues because
the Elasticsearch output makes some assumptions about them: @timestamp has to
be of type `common.Time` and `type` has to be of type `string`.

The fix attempts to make use of the `@timestamp` and `type` from the JSON, but if
parsing fails or the resulting values are invalid, the fields are not overwritten
and an error key is added to help with troubleshooting.

Fixes #1378.

Also hardens the `getIndex` call in libbeat, to protect against wrong types there
as well.

PR #1831, opened 2016-06-10
filebeat: Add failing tests for @timestamps in JSON with timezone specified

morganchristiansson:
filebeat is currently generating errors for timestamps with timezone specified as +00:00 ('Z' is allowed without suffix). Failing tests added.

This is the default @timestamp format used by kibana json log for example.

Seems to have been introduced here: https://github.com/elastic/beats/pull/1421/files#r60245935

### DIFF
--- a/filebeat/tests/files/logs/json_timestamp.log
+++ b/filebeat/tests/files/logs/json_timestamp.log
@@ -1,3 +1,5 @@
-{"@timestamp":"2016-04-05T18:47:18.444Z","level":"INFO","logger":"iapi.logger","thread":"JobCourier4","appInfo":{"appname":"SessionManager","appid":"Pooler","host":"demohost.mydomain.com","ip":"192.168.128.113","pid":13982},"userFields":{"ApplicationId":"PROFAPP_001","RequestTrackingId":"RetrieveTBProfileToken-6066477"},"source":"DataAccess\/FetchActiveSessionToken.process","msg":"FetchActiveSessionToken process ended"}
-{"@timestamp":"invalid","level":"INFO","logger":"iapi.logger","thread":"JobCourier4","appInfo":{"appname":"SessionManager","appid":"Pooler","host":"demohost.mydomain.com","ip":"192.168.128.113","pid":13982},"userFields":{"ApplicationId":"PROFAPP_001","RequestTrackingId":"RetrieveTBProfileToken-6066477"},"source":"DataAccess\/FetchActiveSessionToken.process","msg":"FetchActiveSessionToken process ended"}
-{"@timestamp":{"hello": "test"},"level":"INFO","logger":"iapi.logger","thread":"JobCourier4","appInfo":{"appname":"SessionManager","appid":"Pooler","host":"demohost.mydomain.com","ip":"192.168.128.113","pid":13982},"userFields":{"ApplicationId":"PROFAPP_001","RequestTrackingId":"RetrieveTBProfileToken-6066477"},"source":"DataAccess\/FetchActiveSessionToken.process","msg":"FetchActiveSessionToken process ended"}
+{"@timestamp":"2016-04-05T18:47:18.444Z"}
+{"@timestamp":"invalid"}
+{"@timestamp":{"hello": "test"}}
+{"@timestamp":"2016-04-05T18:47:18.444+00:00"}
+{"@timestamp":"2016-04-05T18:47:18+00:00"}
--- a/filebeat/tests/system/test_json.py
+++ b/filebeat/tests/system/test_json.py
@@ -188,12 +188,12 @@ class Test(BaseTest):
 
         proc = self.start_beat()
         self.wait_until(
-            lambda: self.output_has(lines=3),
+            lambda: self.output_has(lines=5),
             max_timeout=10)
         proc.check_kill_and_wait()
 
         output = self.read_output()
-        assert len(output) == 3
+        assert len(output) == 5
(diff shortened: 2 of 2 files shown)


ruflin:
@morganchristiansson Thanks a lot for all the details. We will have a look to check what goes wrong here.

tsg:
Filebeat currently doesn't support a configurable timestamp format, and the current format requires the Z at the end. What we could do to support this test case without adding configurable patterns is to switch to  RFC3339 as the format. That accepts both formats: https://play.golang.org/p/boxSimIhq8

We could also just make it configurable.

@morganchristiansson do you need just this format or configurable formats?

morganchristiansson:
Hi.

I have no idea why it's suggested that filebeat should enforce the format of @timestamp when the rest of the stack seemingly supports ISO8601/RFC3339. Why do you even attempt to parse @timestamp other than to pass it to the output as a string? When I ask filebeat to parse the JSON I do not want it to mangle our @timestamp field.

JSON parsing was only recently added and you immediately jump to enforce the format of the @timestamp ?? WHY? This is already handled by both logstash and elasticsearch, it is not the log collectors job to try and parse it!

We switched from nxlog because it was converting @timestamp to a format that logstash/elasticsearch wouldn't recognize and also stripping precision from the @timestamp value and now filebeat is going down the same route!



Also the format you are using as example in the filebeat reference log output is RFC3339 which is not supported by filebeat. Much like kibanas timestamp format is not supported either.

filebeat: 2006-01-02T15:04:05Z07:00 <-- 07:00 is not supported. Only UTC!

kibana: 2016-06-10T16:29:37+00:00 <-- Z missing and +00:00 not supported.

https://www.elastic.co/guide/en/beats/filebeat/master/configuration-logging.html#_logging_format


If you're going to enforce the format you really should make sure you use your own format that all filebeat users should now be using. I shall do the same and reach out to the multiple teams that are now generating invalid timezone formats as only a subset of the ISO8660/RFC3336 is now supported.


Or should we start using an alternative field and write logstash configuration to rename it back just so that filebeat won't muck it up? All other apps including kibana seem to generate documents with (invalid)  @timestamp field.

ruflin:
It's a very good point that we should be consistent across the stack or support at least the other formats. I think the first step we should do is to support both formats which seems to be quite a simple change.

The second question is if we should allow any kind of structure in @timestamp. If JSON is used, should we try to parse the @timestamp value if it is a string or just leave it as it is? All the beats and elasticsearch rely that @timestamp is a valid `date` field so I think it should be a supported timestamp. Making it configurable would also require to make the template potentially configurable so elsaticsearch understands the date format. This seems to add too much complexity I assume.

If someone does want to use his own @timestamp format this is possible, by not using KeysUnderRoot (never tested, but this should work).

morganchristiansson:
Personally I mostly want more support for specifying timezones in UTC and allowing non-UTC. The +00:00/+01:00 format of ISO8660/RFC is required really.

RFC3339 format for parsing suggested by @tsg sounds great. Though I'm curious what the reason is for filebeat attempts to parse the @timestamp and not just pass on the logline to the output as is? I suspect other timeformats will be needed.

As for keys_under_root setting, it changes existing data and is not what I want, although with a logstash filter to move the values under the json key to root it would probably work great. Will consider this.

morganchristiansson:
My dislike for handling dates and times increases. Only somewhat related but thought I'd share. 

ISO860 in PHP

Source: http://www.reddit.com/r/lolphp/comments/4oacnc/_/

ruflin:
@morganchristiansson The reason it is parse is mainly for historical reason. All other implementations actually generate the timestamp and we use internally a fixed type to handle this field. So in case it is not converted, we have to change larger parts in the code base. That doesn't mean it should not be changed, at the same time I feel like as this is almost the only field we required, we should be strict here for the type.

morganchristiansson:
It would be useful to record when the log was processed by filebeat in a secondary field like beat.timeatamp.

Parsing dates is a can of worms best avoided or delegated to libraries that specialise in it.

ruflin:
We first had that in `timestamp` and LS was then adding `@timestamp` which had become really confusing. Now that we have meta data under `beat` this could be worth discussing again. The problem is that in some cases the log lines are directly sent to elasticsearch and then the @timestamp would be missing.

ruflin:
@morganchristiansson Could you have a look at this PR? https://github.com/elastic/beats/pull/1889 I took your changes / tests and built on top the necessary changes.

ruflin:
@morganchristiansson BTW: Could you sign the CLA so I can take your code in my commit? https://www.elastic.co/contributor-agreement/

andrewkroh:
Incorporated into #1889

morganchristiansson:
I've implemented this workaround/solution to fix the events in logstash after they've exited the filebeats pipeline.

```
filter {
  if [json] {
    ruby {
      code => "event['json'].each { |k,v| event[k]=v } ; event.remove('json')"
    }
  }
}
```

and then set ```filebeat.prospectors[].json.keys_under_root: false``` everywhere.

ruflin:
@morganchristiansson Thanks for sharing. I'm not 100% sure in which cases this is needed?

morganchristiansson:
It's a workaround waiting for filebeat-5.0.0-alpha4 release and also handling any ISO8601 dates that aren't RFC3339 dates, such as timezones with missing colon (+0000)

ruflin:
@morganchristiansson Got it. Thanks for the details.

PR #1834, opened 2016-06-10
Combine separate es bulk api implementations

urso:
Until now (*Connection).Bulk and (*Client).PublishEvents seem to be based on
different implementations for dealing with bulk API. Only the former being
tested by unit and integration tests. This refactoring removes duplicate logic.


ruflin:
Some of the bulk tests do not seem to like these changes.

PR #1835, opened 2016-06-11
Elasticsearch output support for gzip compressed content-encoding

urso:
Add support and new configuration option to elasticsearch output to enable gzip based  compression of JSON payload in POST requests.

Requires #1834 plus rebase after merging #1834

### DIFF
--- a/libbeat/outputs/elasticsearch/config.go
+++ b/libbeat/outputs/elasticsearch/config.go
@@ -7,19 +7,20 @@ import (
 )
 
 type elasticsearchConfig struct {
-	Protocol     string             `config:"protocol"`
-	Path         string             `config:"path"`
-	Params       map[string]string  `config:"parameters"`
-	Username     string             `config:"username"`
-	Password     string             `config:"password"`
-	ProxyURL     string             `config:"proxy_url"`
-	Index        string             `config:"index"`
-	LoadBalance  bool               `config:"loadbalance"`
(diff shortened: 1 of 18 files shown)


ruflin:
@urso #1834 is merged. Needs rebase now.

urso:
@ruflin Thanks. It's rebased.

libbeat/outputs/elasticsearch/config.go
```diff
@@ -34,16 +35,17 @@ const (
...
-		LoadBalance: true,
+		Protocol:         "",
+		Path:             "",
+		ProxyURL:         "",
+		Params:           nil,
+		Username:         "",
+		Password:         "",
+		Timeout:          90 * time.Second,
+		MaxRetries:       3,
+		CompressionLevel: 0,
```
ruflin: Is that a good default? I assume it is what we had before?

ruflin:
LGTM

libbeat/outputs/elasticsearch/config.go
urso: Right, I made the default to match the old behavior. Making compression opt-in instead of opt-out. Default config is debatable, though.

ruflin: I somehow expected that some compression as default is good. I assume there is somehow a tradeoff between performance and compression level, but we can figure that out later.

PR #1837, opened 2016-06-13
MongoDB metricbeat module

tsg:
* Contains the status metricset, reporting parts of the `db.serverStatus()`
  output. The locks and metrics sections are not yet reported.
* Includes docs, basic integration tests, and  basic system tests


metricbeat/module/mongodb/_beat/config.yml
```diff
@@ -0,0 +1,6 @@
+- module: mongodb
+  metricsets: ["status"]
+  enabled: true
+  period: 1s
```
ruflin: Should we make the default 10s?

metricbeat/module/doc.go
```diff
@@ -34,7 +34,8 @@ List of standardised words and units across all metricsets. On the left are the
 * pct: precentage
 * request: req
 * sec: seconds, second, s
-*
+* ms: millisecond, millis
+* mb: megabytes
```
ruflin: one more we could add here that I spotted in your code and like is `msg` for `message`.

metricbeat/etc/fields.yml
```diff
@@ -205,6 +205,343 @@
...
+                    The number of times MongoDB requested a commit before the scheduled
+                    journal group commit interval.
+                - name: times
+                  type: group
+                  description: >
+                    Information about the performance of the mongod instance during the
+                    various phases of journaling in the last journal group commit
+                    interval.
+                  fields:
+                    - name: dt.ms
```
ruflin: What is `dt` standing for?

metricbeat/module/mongodb/status/data.go
```diff
@@ -0,0 +1,241 @@
...
+	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/logp"
+)
+
+func eventMapping(status map[string]interface{}) common.MapStr {
+	errs := map[string]error{}
+	event := common.MapStr{
+		"version": mustBeString("version", status, errs),
+		"uptime": common.MapStr{
+			"ms": mustBeInteger("uptimeMillis", status, errs),
```
ruflin: Interesting approach with the errs. So far I used to helper.ToInt( functions etc.

metricbeat/module/mongodb/status/data.go
```diff
@@ -0,0 +1,241 @@
...
+		},
+		"local_time":         mustBeTime("localTime", status, errs),
+		"write_backs_queued": mustBeBool("writeBacksQueued", status, errs),
+	}
+	removeErroredKeys(event, errs)
```
ruflin: The current approach was to set empty / default values for non existing values to not "break" dashboards / queries because of missing values. But this approach could be even more powerful and better cross version compatible.

metricbeat/docker-compose.yml
```diff
@@ -34,6 +37,9 @@ kibana:
 apache:
   build: ${PWD}/module/apache/_beat
 
+mongodb:
+  image: mongo:3.0
```
ruflin: Could we add the tested version somewhere to the docs (see other modules).

ruflin:
LGTM. It is interesting that you introduced for most of the fields the "type" like `ms` or `mb`. So far I left it out most of the time except it was already part of the name. I was thinking to do it in the templates but this feels very natural and makes it easy to read.

metricbeat/module/mongodb/status/data.go
tsg: The reason I couldn't use the ToInt stuff is that the stuff from mongo is already parsed in a `map[string]interface{}` so no conversions from strings are required. Very similar otherwise, though.

About the errs, I wanted to have the key missing in case of errors, rather than filled with the zero value.

metricbeat/module/mongodb/status/data.go
tsg: Ok, I see, it's debatable what's best, but with Rhythm I think we'll generally want missing values in case of errors.

ruflin: By "missing values" you mean missing keys, so your approach?

ruflin:
@tsg Seems like Jenkins hit a flaky test?

ruflin:
jenkins, retest it

metricbeat/etc/fields.yml
tsg: Hmm, not sure, I think "delta t". I could expand it to `delta_t`, not sure if that would be better.

metricbeat/module/mongodb/status/data.go
tsg: Right, IMO kibana does a relatively good job for when there are no keys. I would be more worried if we fill it with zeros, as that might hide real issues and we lose users' trust.

tsg:
@ruflin i addressed most comments and rebased to master to make sure the `_beat` to `_meta` rename happens correctly. To make the rebase easier I squashed to one commit.

metricbeat/module/mongodb/status/data.go
ruflin: Agree. Lets try then to get the other metricsets in line here.

ruflin:
jenkins, test it

ruflin:
LGTM

PR #1842, opened 2016-06-13
Implement ucfg for prospector and harvester 

ruflin:
This separates the config options from prospector and harvester and uses ucfg for it.

Further changes:
* Only start new harvester if offset changed -> more efficient if file didn't change

### DIFF
--- a/filebeat/harvester/config.go
+++ b/filebeat/harvester/config.go
@@ -1,0 +1,65 @@
+package harvester
+
+import (
+	"fmt"
+	"regexp"
+	"time"
+
+	cfg "github.com/elastic/beats/filebeat/config"
+	"github.com/elastic/beats/filebeat/input"
+	"github.com/elastic/beats/libbeat/common"
+)
+
+var (
+	defaultConfig = harvesterConfig{
+		BufferSize:      cfg.DefaultHarvesterBufferSize,
+		DocumentType:    cfg.DefaultDocumentType,
+		InputType:       cfg.DefaultInputType,
+		TailFiles:       cfg.DefaultTailFiles,
+		Backoff:         cfg.DefaultBackoff,
+		BackoffFactor:   cfg.DefaultBackoffFactor,
+		MaxBackoff:      cfg.DefaultMaxBackoff,
+		CloseOlder:      cfg.DefaultCloseOlder,
+		ForceCloseFiles: cfg.DefaultForceCloseFiles,
+		MaxBytes:        cfg.DefaultMaxBytes,
+	}
+)
+
+type harvesterConfig struct {
+	common.EventMetadata `config:",inline"`     // Fields and tags to add to events.
+	BufferSize           int                    `config:"harvester_buffer_size"`
+	DocumentType         string                 `config:"document_type"`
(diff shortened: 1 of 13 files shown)


filebeat/harvester/config.go
```diff
@@ -0,0 +1,65 @@
...
+	BackoffFactor        int                    `config:"backoff_factor" validate:"min=1"`
+	MaxBackoff           time.Duration          `config:"max_backoff" validate:"min=0,nonzero"`
+	CloseOlder           time.Duration          `config:"close_older"`
+	ForceCloseFiles      bool                   `config:"force_close_files"`
+	ExcludeLines         []*regexp.Regexp       `config:"exclude_lines"`
+	IncludeLines         []*regexp.Regexp       `config:"include_lines"`
+	MaxBytes             int                    `config:"max_bytes" validate:"min=0,nonzero"`
+	Multiline            *input.MultilineConfig `config:"multiline"`
+	JSON                 *input.JSONConfig      `config:"json"`
+}
```
urso: normally I'd prefer to have `validate`-tags to be aligned to have them stand out a little more (just add spaces in tags).

filebeat/harvester/harvester.go
```diff
@@ -40,53 +38,39 @@ type Harvester struct {
...
+	}
+	if err := h.Config.Validate(); err != nil {
+		return nil, err
+	}
+
+	encoding, ok := encoding.FindEncoding(h.Config.Encoding)
+	if !ok || encoding == nil {
+		return nil, fmt.Errorf("unknown encoding('%v')", h.Config.Encoding)
+	}
+	h.encoding = encoding
```
urso: Tip: using [Unpacker interface](https://github.com/urso/ucfg/blob/master/ucfg.go#L31) one can unpack custom types when unpacking configs. If an error is reported by unpack (e.g. unknown codec), the error will pass ucfg => ucfg might add context (name of config that failed + source config variable was read from). In future this might allow us to record custom errors in ucfg.

urso: In case Validate-method is not checking for field.

ruflin: Thanks for the details, was not aware of the interface. 

One other question: Could we somehow print the config options to the log that were set? Reason is that in case people use for example envrionment variables it makes finding errors easier when during startup the variables are printed out. Alternatively it can be done manually after the unpack (not so nice).

filebeat/harvester/config.go
ruflin: @urso Is that a manual thing (no auto formatting)? If yes, that will be quite hard to keep it "clean"

filebeat/harvester/harvester.go
andrewkroh: One thing we have to be careful about is making sure we don't log secrets.

ruflin: Perhaps we could solve this with a special "tag/flag" for these kind of fields.

urso: maybe when running beats with `-configtest -v` we can print the evaluated config file (more interesting once ucfg  variable expansion will be added to beats).

PR #1850, opened 2016-06-14
Update and fix minor issues with scripts

radoondas:
I have found following changes very useful while playing with exporting of dashboards:
- update requirements for python env (elasticsearch)
- fix typo and add python env for export-dashboards target in makefile
- fix 'print' call in export_dashboards script


ruflin:
LGTM

ruflin:
jenkins, test it

PR #1857, opened 2016-06-14
Quick Cleanup

ruflin:
I quick went through the TODO's and fixed the easy ones or removed unnecessary ones.

* Remove debug log messages as it spams the log file. This happens every time an event or update is published

### DIFF
--- a/filebeat/tests/system/test_prospector.py
+++ b/filebeat/tests/system/test_prospector.py
@@ -164,9 +164,6 @@ class Test(BaseTest):
             lambda: self.output_has(lines=1),
             max_timeout=15)
 
-        # TODO: Find better solution when filebeat did crawl the file
-        # Idea: Special flag to filebeat so that filebeat is only doing and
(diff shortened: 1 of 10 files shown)


filebeat/tests/system/test_prospector.py
```diff
@@ -164,9 +164,6 @@ def test_exclude_files(self):
             lambda: self.output_has(lines=1),
             max_timeout=15)
 
-        # TODO: Find better solution when filebeat did crawl the file
-        # Idea: Special flag to filebeat so that filebeat is only doing and
-        # crawl and then finishes
         filebeat.check_kill_and_wait()
```
urso: have we found a better solution? If not, let's move comment to `def check_kill_and_wait`.

ruflin: @urso not yet, but we had the comment in at least 10 times. I still left it in once so we remember.

PR #1858, opened 2016-06-14
Add topic about configuring multiline to 1.2 docs

dedemorton:
This PR adds more info about configuring multiline to the 1.2 docs. Since the POSIX compilation is supported in 1.2, this version of the doc shows the older Go Playground.

@ruflin When I cherry-pick this into 1.3, I assume I'll continue to show the POSIX compilation, no?


ruflin:
@dedemorton Not sure but it seems like Jenkins tells me that this is breaking the doc build?

dedemorton:
@ruflin Yah, you're right. I didn't build the topbeat docs. Should be fixed now.

andrewkroh:
jenkins, test it

Issue #1787, opened 2016-06-02
Metricbeat: Cores are all reporting the same value.


simianhacker:
- Install the `metricbeat` from `master`
- Inspect the data in Kibana and you will see the number are all the same for each core

monicasarbu:
I can easily reproduce the bug on my mac.

monicasarbu:
Fixed by #1863

PR #1863, opened 2016-06-15
Fix the CPU values returned for each core

monicasarbu:
Solves bug https://github.com/elastic/beats/issues/1787


monicasarbu:
@ruflin @andrewkroh Can you please have a look to the fix?

ruflin:
LGTM

PR #1865, opened 2016-06-15
Use x86_64 instead of x64 in tar ball names

### LINKED ISSUE #1808: Metricbeat tarball is oddly named.
For config management, it would be really nice to have filenames for all the Beats that are consistent, predictable and composable.

Metricbeat is kinda unique in its naming right now. Specifically, the generic Linux tarball has:

- Architecture: `x64` not `x86_64`. 
- `linux` in the filename, where Filebeat for example does not.

Filebeat:
```
https://download.elastic.co/beats/filebeat/filebeat-1.2.3-x86_64.tar.gz
```

Metricbeat:
```
https://download.elastic.co/beats/metricbeat/metricbeat-5.0.0-alpha3-linux-x64.tar.gz
```

It would be good if the `linux` string was either present of absent from all packages.
Having the architecture be `x86_64` is really useful, since that's what Puppet and Ansible will return if you ask for the host architecture.

I'd like to be able to say this in Ansible:
```yaml
metricbeat_version: 5.0.0-alpha3
metricbeat_file: metricbeat-{{ metricbeat_version }}-{{ ansible_architecture }}.tar.gz
# -> metricbeat-5.0.0-alpha3-x86_64.tar.gz
```

Thanks.

tsg:
* Fixes #1808.
* x86_64 is more common in config management tools, so it is easier
  to work with.
* The DEB/RPM still use their recommended arch names.

### DIFF
--- a/dev-tools/packer/README.md
+++ b/dev-tools/packer/README.md
@@ -71,13 +71,13 @@ For example, here are the artifacts created for Filebeat:
 
 ```
 filebeat-5.0.0-amd64.deb
-filebeat-5.0.0-darwin-x64.tar.gz
+filebeat-5.0.0-darwin-x86_64.tar.gz
 filebeat-5.0.0-i386.deb
 filebeat-5.0.0-i686.rpm
-filebeat-5.0.0-linux-x64.tar.gz
 filebeat-5.0.0-linux-x86.tar.gz
-filebeat-5.0.0-windows-x64.zip
+filebeat-5.0.0-linux-x86_64.tar.gz
 filebeat-5.0.0-windows-x86.zip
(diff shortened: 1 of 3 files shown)


tsg:
jenkins, package it

tsg:
@dedemorton A note on this one: it will probably break the Getting Started links again. However, I couldn't update them in this PR, because that would break the alpha3 links :(. So we should probably do it in the day of the alpha4 release.

ruflin:
LGTM

tsg:
Packaging tests are green: http://build-eu-00.elastic.co/view/Beats/job/beats-package-PR/11/

Issue #420, opened 2015-12-02
Load / cores metric 


monicasarbu:
Duplicated from: https://github.com/elastic/topbeat/issues/128

Knowing the load of a machine, or even the cpu time spent (percentage) is not very useful if you don't know the number of cores. i.e. a load of 3.1 is still healthy if you have a 4-core machine, but very unhealthy if you have a 2-core machine.
Would it be possible to add another metric to the system messages indicating the value of ( load / cores ) ?

pietervogelaar:
What is the status of this issue? I really want this, because a system load that is not divided by cores is not very useful to show on a Kibana dashboard for Linux systems.

monicasarbu:
@pietervogelaar This is implemented in https://github.com/elastic/beats/pull/2101 it will be available in the next release, 5.0.0-alpha5, that is coming out this week.

monicasarbu:
Closing this issue as it's implemented in #2101.

PR #1868, opened 2016-06-15
Export another metric load / number of cores

monicasarbu:
Implements https://github.com/elastic/beats/issues/420

cc @andrewkroh @ruflin


metricbeat/module/system/cpu/cpu.go
```diff
@@ -100,6 +100,11 @@ func (m *MetricSet) Fetch() (common.MapStr, error) {
 		"5":  loadStat.Load5,
 		"15": loadStat.Load15,
 	}
+	cpuStat["load_normalized"] = common.MapStr{
```
ruflin: Should we go here with `load.1` and `load.normlized.1` or `load.1.value` and `load.1.normlized`. I think at least both should be in the namespace `load`. Not sure about the naming below it.

ruflin:
I think the fields.yml updates are missing here (which means probably also the system tests will fail).

metricbeat/module/system/cpu/cpu.go
monicasarbu: I agree that they should be under `load`. I suggest `load.avg.1` and `load.normalized.1`.

metricbeat/module/doc.go
```diff
@@ -26,6 +26,7 @@ The goal is to have a similar experience across all metrics.
 List of standardised words and units across all metricsets. On the left are the ones to be used, on the right the options seen in metricsets.
 
 * avg: average
+* norm: normalized
```
ruflin: I think we should sort the list alphabetical

monicasarbu: sorry, I didn't notice that it's sorted.

metricbeat/etc/kibana/visualization/System-load.json
```diff
@@ -1,5 +1,5 @@
 {
-  "visState": "{\"title\":\"New Visualization\",\"type\":\"line\",\"params\":{\"shareYAxis\":true,\"addTooltip\":true,\"addLegend\":true,\"showCircles\":true,\"smoothLines\":false,\"interpolate\":\"linear\",\"scale\":\"linear\",\"drawLinesBetweenPoints\":true,\"radiusRatio\":9,\"times\":[],\"addTimeMarker\":false,\"defaultYExtents\":false,\"setYExtents\":false,\"yAxis\":{}},\"aggs\":[{\"id\":\"1\",\"type\":\"avg\",\"schema\":\"metric\",\"params\":{\"field\":\"system.cpu.load.1\",\"customLabel\":\"Load\"}},{\"id\":\"2\",\"type\":\"date_histogram\",\"schema\":\"segment\",\"params\":{\"field\":\"@timestamp\",\"interval\":\"auto\",\"customInterval\":\"2h\",\"min_doc_count\":1,\"extended_bounds\":{}}},{\"id\":\"3\",\"type\":\"terms\",\"schema\":\"group\",\"params\":{\"field\":\"beat.name\",\"size\":5,\"order\":\"desc\",\"orderBy\":\"1\"}}],\"listeners\":{}}",
```
ruflin: The kibana files are also in the _meta folder now, so changes should be made there.

andrewkroh:
LGTM

ruflin:
LGTM, but it seems to break the system tests for topbeat on windows.

monicasarbu:
Closed in favor of https://github.com/elastic/beats/pull/2101

PR #1873, opened 2016-06-16
Add build docs script

ruflin:
Building the docs is now part of make testsuite

Building of the generators was also added to the testsuite.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,11 +3,12 @@ COVERAGE_DIR=${BUILD_DIR}/coverage
 BEATS=packetbeat topbeat filebeat winlogbeat metricbeat
 PROJECTS=libbeat ${BEATS}
 
-# Runs complete testsuites (unit, system, integration) for all beats,
-# with coverage and race detection.
+# Runs complete testsuites (unit, system, integration) for all beats with coverage and race detection.
+# Also it builds the docs and the generators
 .PHONY: testsuite
 testsuite:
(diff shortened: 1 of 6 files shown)


ruflin:
@dedemorton This means in the future you can just run `make docs-preview` instead of assembling the command yourself ...

ruflin:
@andrewkroh I removed the doc build in Jenkins to replace it by this PR. Should we backport this to 1.3, at least the overall docs build?

Makefile
```diff
@@ -3,11 +3,13 @@ COVERAGE_DIR=${BUILD_DIR}/coverage
...
 
-# Runs complete testsuites (unit, system, integration) for all beats,
-# with coverage and race detection.
+# Runs complete testsuites (unit, system, integration) for all beats with coverage and race detection.
+# Also it builds the docs and the generators
 .PHONY: testsuite
 testsuite:
 	$(foreach var,$(PROJECTS),$(MAKE) -C $(var) testsuite || exit 1;)
+	#$(MAKE) -C generate test
+	$(MAKE) docs
```
andrewkroh: I would prefer to not have the docs build tied to testsuite. I would rather be able to specify if I want them both to run like `make testsuite docs`.

andrewkroh:
Yeah, it would be good to backport at least the `make docs` part of this.

Makefile
ruflin: I kept it in to not have to change the jenkins command and it is currently incompatible with 1.3. but I will just update 1.3 very soon, so this is not a problem. I removed it.

PR #1875, opened 2016-06-16
Logstash window sizing if pipelining is enabled

urso:
Logstash output to adapt window sizes faster by growing/shrinking windows per
windowed batch being pushed instead of full batch.


andrewkroh:
Is this something for the changelog?

urso:
I'd say no. it's some enhancement for experimental feature which should not be enabled yet. pipelining is for java-based rewrite of beats input plugin.

PR #1879, opened 2016-06-16
CHANGELOG generator

ruflin:
- [ ] Add migration script to generate new changelog from old format

### DIFF
--- a/changelog.yml
+++ b/changelog.yml
@@ -1,0 +1,15 @@
+project: beats
+changes:
+
+- type: added
+  description: >
+    Add changelog generation script to generate changelogs from structured yaml files.
+  issue: 1879
+  target: dev
+
+- type: release
+  version: 5.0.0-alpha3
+  parent: 5.0.0-alpha2
+  date: "2016-05-31"
+
+
--- a/CHANGELOG2.asciidoc
+++ b/CHANGELOG2.asciidoc
@@ -1,0 +1,1326 @@
+////
+This file is generated! See scripts/changelog.py
+////
+
+[[release-notes-v5.0.0-rc1]]
+=== Beats version v5.0.0-rc1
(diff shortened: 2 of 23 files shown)


changelog.yml
```diff
@@ -0,0 +1,55 @@
+project: filebeat
+changes:
+- type: release
+  version: 5.0.0-alpha3
+  date: 2015-12-12
+  diff: https://github.com/elastic/beats/compare/v5.0.0alpha2...v5.0.0-alpha3
```
urso: instead of `diff` one can add the git label and `parent: 5.0.0-alpha2`

CHANGELOG.asciidoc
```diff
@@ -652,13 +652,6 @@ https://github.com/elastic/beats/compare/v1.2.0...v5.0.0-alpha1[View commits]
 - Add `event_id`, `level`, and `provider` configuration options for filtering events {pull}1218[1218]
 - Add `include_xml` configuration option for including the raw XML with the event {pull}1218[1218]
 
-==== Known issues
```
ruflin: @tsg We had "known issues" in one of the past release. I think it is an interesting question if that is something we will need in the future also as part of the CHANGELOG?

ruflin: @tsg Keep them in would mean we have a type "Known issues" in the script. I think it is a good idea as it will allow us to put the known issues into the yml file but use the script to put them where we need it.

ruflin:
I'm closing this PR as I think in the short term we will not make use of this. We need to first figure out what our "end" structure for changelogs should be. As soon as we have plugins or more modules  this could become useful again. In case we want to pick it up again, I stored the source code here: https://github.com/ruflin/generators/tree/master/changelog

PR #1882, opened 2016-06-17
Start a new harvester when file was truncated

ruflin:
Currently file truncation was handeld by the reader. In case of truncation, the same harvester stayed open and the reader just continued reading from the beginning. This is the most efficient way to make sure the truncated file is immidiately read.

From a logical point of view, truncation of a file is very similar to creating a new file. Instead of letting the reader just continue, the harvester is closed and a completely new harvester is started with a new reader. This means the reader only needs to detect a truncation, but doesnt' have to decide what happens.

The consequence of this change is that it can take up to `scan_frequency` to read the truncated file, but that is the same behaviour as for a new file. In addition the file handler will be closed when the old harvester is stopped.

### DIFF
--- a/filebeat/harvester/log.go
+++ b/filebeat/harvester/log.go
@@ -2,7 +2,6 @@ package harvester
 
 import (
 	"errors"
-	"io"
 	"os"
 
 	"golang.org/x/text/transform"
@@ -64,17 +63,9 @@ func (h *Harvester) Harvest() {
 		ts, text, bytesRead, jsonFields, err := readLine(reader)
 		if err != nil {
 			if err == errFileTruncate {
-				seeker, ok := h.file.(io.Seeker)
-				if !ok {
-					logp.Err("can not seek source")
-					return
-				}
-
 				logp.Info("File was truncated. Begin reading file from offset 0: %s", h.Path)
-
 				h.SetOffset(0)
-				seeker.Seek(h.getOffset(), os.SEEK_SET)
-				continue
+				return
 			}
 
 			logp.Info("Read line error: %s", err)
--- a/filebeat/harvester/reader.go
+++ b/filebeat/harvester/reader.go
@@ -11,10 +11,9 @@ import (
 )
 
 type logFileReader struct {
-	fs        FileSource
-	offset    int64
-	config    logFileReaderConfig
-	truncated bool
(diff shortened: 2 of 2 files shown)


filebeat/harvester/log.go
```diff
@@ -74,7 +74,7 @@ func (h *Harvester) Harvest() {
 
 				h.SetOffset(0)
 				seeker.Seek(h.getOffset(), os.SEEK_SET)
-				continue
+				return
```
urso: The complete `if err == errFileTruncate` can be removed. Where/how does prospector detect the file being truncated? Still `h.SetOffset(0)` required?

ruflin: It is the harvester that detects that the file was truncated, not prospector. The prospector only sees a finished harvester and opens a new one if the offset (which is 0).

urso: In this case it's still too much code. There is no need to call `seeker.Seek`.

ruflin: Yep, will remove it.

filebeat/harvester/log.go
```diff
@@ -64,17 +63,9 @@ func (h *Harvester) Harvest() {
 		ts, text, bytesRead, jsonFields, err := readLine(reader)
 		if err != nil {
 			if err == errFileTruncate {
-				seeker, ok := h.file.(io.Seeker)
-				if !ok {
-					logp.Err("can not seek source")
-					return
-				}
-
 				logp.Info("File was truncated. Begin reading file from offset 0: %s", h.Path)
```
urso: Is info message enough? There is a chance we did loose lines.

ruflin: That is a good point. We should make a WARN out of this and add a comment that this means that not all content was read potentially. WDYT?

But I suggest to make this in a second PR, ok?

ruflin: @urso after our conversation yesterday about warn messages, I will leave it as info, ok?

urso: Didn't we remove the requirement to seek most recently?

ruflin: Yes, in this PR. But the logging that the file was truncated is still needed.

PR #1883, opened 2016-06-17
[beats-generator] Update .travis for better support for community beat

radoondas:
During implementation of Travis CI for community beat I found out that some of the generated paths and options don't work for the beat.
  - make command doesn't need directory setting (-C directory) because we talk about single beat
  - paths for travis setup require beat paths and name and not 'beats' paths

This, of course require review. Thanks.

### DIFF
--- a/generate/beat/{{cookiecutter.beat}}/.travis.yml
+++ b/generate/beat/{{cookiecutter.beat}}/.travis.yml
@@ -15,7 +15,7 @@ os:
 env:
   matrix:
     - TARGETS="check"
-    - TARGETS="-C {{cookiecutter.beat}} testsuite"
+    - TARGETS="testsuite"
 
   global:
     # Cross-compile for amd64 only to speed up testing.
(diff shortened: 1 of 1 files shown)


generate/beat/{{cookiecutter.beat}}/.travis.yml
```diff
@@ -28,10 +28,10 @@ addons:
 
 before_install:
   # Redo the travis setup but with the elastic/libbeat path. This is needed so the package path is correct
-  - mkdir -p $HOME/gopath/src/github.com/elastic/beats/
-  - rsync -az ${TRAVIS_BUILD_DIR}/ $HOME/gopath/src/github.com/elastic/beats/
-  - export TRAVIS_BUILD_DIR=$HOME/gopath/src/github.com/elastic/beats/
-  - cd $HOME/gopath/src/github.com/elastic/beats/
+  - mkdir -p $HOME/gopath/src/github.com/{{cookiecutter.github_name|lower}}/{{cookiecutter.beat}}/
```
ruflin: Nice. That part is actually only important for forks of the repo. For contributions that is great to have it in.

ruflin:
LGTM. Thanks a lot for fixing this. So far I didn't test making a community beat and add it to travis (I totally should). Talking about it, I realised I didn't add yet a travis file to the metricset generator: https://github.com/elastic/beats/tree/master/generate/metricbeat/metricset/%7B%7Bcookiecutter.beat%7D%7D We can probably reuse the same.

One idea related to travis is that we could instrument it for community beats to create the release with `make pack`: https://github.com/elastic/beats/blob/master/generate/beat/%7B%7Bcookiecutter.beat%7D%7D/Makefile#L41 This would be called when a build is tagged to generate the binaries (just an idea).

Issue #1792, opened 2016-06-03
Rename config topics to refelect reorganization of the config file


dedemorton:
Now that the config file uses dot notation and the full version of the YAML has been reorganized, the topic titles and organization of the reference doc does not match the YAML file.

The doc should use titles that are similar to the section titles we use in the YAML file so that users can map what they see in the config file to what we describe in the doc.

dedemorton:
Fixed with #1885

PR #1885, opened 2016-06-18
Change titles in config reference topics

dedemorton:
This PR resolves issue #1792 

Section titles are renamed to match the section comments in the long version of each config file (hopefully making it easier for users to find what they're looking for in the reference docs).

Some content has been moved around to better match the organization of the config files.

### DIFF
--- a/filebeat/docs/reference/configuration/filebeat-options.asciidoc
+++ b/filebeat/docs/reference/configuration/filebeat-options.asciidoc
@@ -1,5 +1,5 @@
 [[configuration-filebeat-options]]
-=== Filebeat
+=== Filebeat Prospectors Configuration
 
 The `filebeat` section specifies a list of `prospectors` that Filebeat
 uses to locate and process log files. Each prospector item begins with a dash (-)
@@ -44,6 +44,18 @@ The value that you specify here is used as the `input_type` for each event publi
 
 A list of glob-based paths that should be crawled and fetched. Filebeat starts a harvester for
 each file that it finds under the specified paths. You can specify one path per line. Each line begins with a dash (-).
+
+===== encoding
+
+The file encoding to use for reading files that contain international characters.
+See the encoding names http://www.w3.org/TR/encoding/[recommended by the W3C for use in HTML5].
+
+Here are some sample encodings from W3C recommendation:
+
+    * plain, latin1, utf-8, utf-16be-bom, utf-16be, utf-16le, big5, gb18030, gbk, hz-gb-2312,
+    * euc-kr, euc-jp, iso-2022-jp, shift-jis, and so on
+
+The `plain` encoding is special, because it does not validate or transform any input.
 
 [[exclude-lines]]
 ===== exclude_lines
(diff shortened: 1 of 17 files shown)


filebeat/docs/reference/configuration/filebeat-options.asciidoc
```diff
@@ -329,6 +341,13 @@ You can force Filebeat to close the file as soon as the file name changes by set
 `force_close_files` option to true. The default is false. Turning on this option can lead to loss of data on
 rotated files in case not all lines were read from the rotated file.
 
+[[configuration-global-options]]
+=== Filebeat Global Configuration
+
+You can specify configuration options that control Filebeat behavior at a global level.
+
+==== Options
```
ruflin: I assume we need this title here, because asciidoc does not allow leaving it out?

filebeat/docs/reference/configuration/filebeat-options.asciidoc
```diff
@@ -386,27 +405,14 @@ If the specified path is not absolute, it is considered relative to the configur
...
-===== encoding
-
-The file encoding to use for reading files that contain international characters.
-See the encoding names http://www.w3.org/TR/encoding/[recommended by the W3C for use in HTML5].
-
-Here are some sample encodings from W3C recommendation:
-
-    * plain, latin1, utf-8, utf-16be-bom, utf-16be, utf-16le, big5, gb18030, gbk, hz-gb-2312,
-    * euc-kr, euc-jp, iso-2022-jp, shift-jis, and so on
+include::../../../../libbeat/docs/shipperconfig.asciidoc[]
```
ruflin: We should probably rename this file from shipper config, as shipper part does not exist anymore. This is mainly for the url ...

libbeat/docs/runconfig.asciidoc
```diff
@@ -11,7 +11,7 @@
 //////////////////////////////////////////////////////////////////////////
 
 [[configuration-run-options]]
-=== Run Options (Optional)
+=== Run Options Configuration (Optional)
```
ruflin: I think these options are Packetbeat only?

ruflin:
LGTM. I left some minor comments. We should defintively rename shipperconfig to prevent confusion.

libbeat/docs/runconfig.asciidoc
dedemorton: @ruflin I don't see runoptions in the packetbeat.full.yml (or any of the other full config files). Maybe @monicasarbu knows?

ruflin: Yes, they are not in the config file (but probably should be). I actually checked the source code and these options are part of packetbeat and not libbeat. Lets move it to packetbeat and in a second step either remove them if not needed or add them to the packetbeat full config file (with tests ...)

filebeat/docs/reference/configuration/filebeat-options.asciidoc
dedemorton: @ruflin It's not exactly an asciidoc issue. In most cases, we could remove this heading and move everything up a level, but then we'd get inconsistent formatting for options because in some places we break the options down into further categories. I think it's better to keep all the options at the same heading level so that they appear consistent across the doc (same formatting, etc), and so that you don't get a long list of options under the "On this page" area above the navigation.

dedemorton:
@ruflin I've updated the docs to reflect your feedback. This PR should be ready to merge if my changes look OK to you.

ruflin:
@dedemorton I merged it as is. About the run options: the file should probably be moved into the packetbeat docs folder.

PR #1888, opened 2016-06-20
Script for cleaning up vendor directory

ruflin:
This scripts removes the unnecessary files from the vendor directory. As we check in the vendor directory the goal is to keep the number of files to a minimum and if changes are made to the dependency make sure that it is easy to review the relevant changes.


scripts/clean_vendor.sh
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env bash
+
+# Removes unncessary files from the vendor directories
+#
+# A aist for files to be removed is used instead of excluding files.
```
urso: s/aist/list/

scripts/clean_vendor.sh
```diff
@@ -0,0 +1,41 @@
...
+# The reason is that this that this makes the setup simpler and prevents
+# from removing files by accident
+#
+# In general it should always be removed manually which files were removed.
```
urso: Wording?

scripts/clean_vendor.sh
```diff
@@ -0,0 +1,41 @@
...
+# For example some projects like stretchr/testify have two LICENSE files
+# with different names, where it is ok that one is removed.
+#
+# We keep the CHANGELOG in as this makes it easy visible when updating dependencies
+# on what has changed.
+
+# In general the following files should b kept:
```
urso: s/b kept/be kept/

scripts/clean_vendor.sh
```diff
@@ -0,0 +1,41 @@
...
+# * .go
+# * LICENSE* CHANGELOG* and PATENT*
```
urso: potentially add: Contributors, Readme ?

scripts/clean_vendor.sh
ruflin: s/removed/checked/

scripts/clean_vendor.sh
```diff
@@ -0,0 +1,41 @@
...
+#
+## Removing all golang test files
+find $DIR_LIST -type f -name "*_test.go" -exec rm -r {} \;
+#
+## Removing all files starting with a dot (like .gitignore)
+find $DIR_LIST -type f -name ".*" -exec rm -r {} \;
+#
+## Removing all .txt files which are normally test data or docs
+## Excluding LICENSE as nranchev/go-libGeoIP has the license in this file
+find $DIR_LIST -type f -name "*.txt" -a ! -name "LICENSE.*" -exec rm -r {} \;
```
urso: let's use `-iname` in case file is named `License` or `License.txt` or ...

scripts/clean_vendor.sh
urso: I don't see filters for CHANGELOG, PATENT in line 39.

ruflin: Added

ruflin:
@urso Thanks for the review. New version pushed.

scripts/clean_vendor.sh
ruflin: @urso That here is more a general remark. I have the LICENSES below as there was one project using it. I would only add "exception" if they occur ...

ruflin:
@urso As discussed, all the other files we should not be deleted are now also excluded by default for *.txt files.

ruflin:
jenkins, test it

scripts/clean_vendor.sh
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env bash
+
+# Removes unncessary files from the vendor directories
+#
+# A list for files to be removed is used instead of excluding files.
+# The reason is that this that this makes the setup simpler and prevents
```
andrewkroh: "that this that this"

PR #1889, opened 2016-06-20
Improve timestamp handling

ruflin:


karmi:
Hi @ruflin, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/beats/pull/1889.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

karmi:
Hi @ruflin, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/beats/pull/1889.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

andrewkroh:
LGTM, but still needs CLA

karmi:
@andrewkroh, @ruflin, note this is a bug in the CLA checker, which should have pinged Morgan specifically, because only his commit is not signed. I'll have a look into that.

morganchristiansson:
Apologies about not signing the CLA. I wasn't sure whether it was needed for just the 2 test cases.

My company is very reasonable about this I should have it ready by COB today.

Thank you!

ruflin:
@morganchristiansson Nice, thanks a lot. Can you quickly ping me here as soon as it is signed so I can manually check and merge it then?

morganchristiansson:
CLAs have been signed by both CTO and myself. @ruflin

ruflin:
@morganchristiansson Looks good. Do you mind if I add the `git@...` email address also to our CLA checker? Because that is the one that is checked because it is in the git commit.

morganchristiansson:
Yeah go ahead. Thank you.

ruflin:
@morganchristiansson Done. So future commits will be automatically recognized by the CLA checker. 

With merging this PR I think we resolved the most pressing issues in https://github.com/elastic/beats/pull/1831 Feel free to open a new issue in case you think further changes to the timestamp handling are needed.

morganchristiansson:
Thanks @ruflin! Unfortunately I already have to raise another issue #1893 :-) but it's not urgent this time.

ruflin:
Thanks. Will have a look.

tsg:
@ruflin I'm late to the party here, but we could have also modified the ParseTime functions to accept RFC3339, I think, because it would have been backwards compatible because RFC3339 is a super-set of what we had before. I have a feeling that not using RFC3339 as our default everywhere is going to cause other issues like this one in other parts of the code.

ruflin:
@tsg That's the first thing I did but it broke quite a few things like the logging output changed a little bit or at least some tests broke. So to not bloat up the PR I only implemented the change there. But in general +1 on change it on a global level.

tsg:
Thanks @ruflin, sounds good.

ruflin:
@tsg Should we open an issue to implement it on a general level?

PR #1890, opened 2016-06-20
Removing Topbeat from master

ruflin:
This should remove the confusion that topbeat was still in the beats repo but no further development in master

* We must make sure the docbuild still works. 1.x docs must stay around
* Remove references in docs
* Remove build file
* Remove topbeat from doc examples

TBD: Update dev guide to remove Topbeat


ruflin:
@tsg Should we point this to 1.3 instead of master? https://github.com/elastic/docs/blob/master/conf.yaml#L416

tsg:
Looks like you figured it out, docs look good 👍 .

Issue #1031, opened 2016-02-24
Winlogbeat - Test Usage on Windows Event Collector


andrewkroh:
I want to make sure users have a good experience when deploying Winlogbeat on a event collector machine. In this architecture event logs from multiple machines are forwarded to the collector and stored in the "Forwarded Events" event log.

In particular, verify that the `message` field is rendered without error for events from the source machines.

Setup summary:

"enable winrm on the forwarding hosts, open the firewall ports, add the collector machine account to the forwarding machine's local eventviewers group then create the subscription on the collector"

davidhowell-tx:
No issues with message fields coming from my event collector.

andrewkroh:
@davidhowell-tx Thank you for commenting! Are you using a pull subscription (event collector pulls from event sources) or a push (event sources push to event collector)? And what operating systems are you using?

davidhowell-tx:
My Event Collector is a Windows Server 2012 R2 system, and my source systems are all Windows Server 2012 R2 as well.  I am using Source initiated subscriptions, so push.

Kevin-Valle:
I am having an issue with the message field not being rendered.
Instead, the message_error field is filled with "The system cannot find the file specified".
The other fields are filled in correctly, the message is just not rendered.

I am running WinlogBeat on an event collector with Source-initiated subscriptions that store into the "Forwarded Events" event log

andrewkroh:
I wonder if the format setting could cause that problem. See `/cf:FORMAT` on https://msdn.microsoft.com/en-us/library/bb736545(v=vs.85).aspx. Maybe try toggling that setting and see if it has any effect.

Kevin-Valle:
Found my problem: The event collector did not have the event manifests from the application that created the original events on the source computer. Installing the application on the event collector resolved the issue. Thanks for responding!

andrewkroh:
I did find a small issue in 5.X when testing. It causes message_error to be always be present but otherwise it is working as expected. See my comment here for more details:
https://discuss.elastic.co/t/winlogbeat-message-error-the-system-cannot-find-the-file-specified/48125/11

I am targeting a fix for that issue for 5.0.0-beta1.

andrewkroh:
There's one more thing that should be fixed so I'm leaving this open. I think the state that we persist to disk needs to be more than just a single number in order to be able to correctly resume reading the ForwardedEvents log on restart. There are multiple log sources contained in ForwardedEvents and each source has its own record number counter.

[...]

PR #1891, opened 2016-06-20
Fix for rendering ForwardedEvents in Winlogbeat.

andrewkroh:
When reading events from the ForwardedEvents log, Winlogbeat was always attempting to render the event locally event using messages files on the collector. This was wrong because the collector may not have the same message files as the source machine. It would result in a message_error if the collector machine was missing the application that originally logged the event on the source machine.

This PR changes Winlogbeat to not attempt to render the event on the collector machine, but instead use the event as is. This requires that the subscription for the Windows event collector uses the "RenderedText" format which includes the message string and any description strings with the forwarded event.

Fixes: https://github.com/elastic/beats/issues/1031#issuecomment-225851259

### DIFF
--- a/winlogbeat/winlogbeat.full.yml
+++ b/winlogbeat/winlogbeat.full.yml
@@ -25,8 +25,8 @@
 # dictionaries.
 #
 # The supported keys are name (required), tags, fields, fields_under_root,
-# ignore_older, level, event_id, provider, and include_xml. Please visit the
-# documentation for the complete details of each option.
+# forwarded, ignore_older, level, event_id, provider, and include_xml. Please
+# visit the documentation for the complete details of each option.
 # https://go.es.io/WinlogbeatConfig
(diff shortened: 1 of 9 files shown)


ruflin:
LGTM. I think this config should be added to the config file? Also worth a CHANGELOG entry.

andrewkroh:
Added a changelog entry.

andrewkroh:
Added `forwarded` config option to the config files.

winlogbeat/winlogbeat.full.yml
```diff
@@ -25,8 +25,8 @@
 # dictionaries.
 #
 # The supported keys are name (required), tags, fields, fields_under_root,
-# ignore_older, level, event_id, provider, and include_xml. Please visit the
```
ruflin: @andrewkroh We should probably add the full config details to the winlogbeat.full.yml config file.

PR #1894, opened 2016-06-21
Improve Filebeat organisiation and Cleanup

ruflin:
The overall goal is to decouple all the modules to make them better reusable and extendable.
In addition the current organisation was not very intuitive.

* Move registrar to its own package
* Improve startup handling of filebeat
* Rename state to states in registrar as this is more accurate
* Add warn message for truncated files. See https://github.com/elastic/beats/pull/1882/files#r67523587
* Move ignore older to ProspectorLog as only used there
* Remove Prospector reference from ProspectorStdin
* Remove empty filebeat test file
* Cleanup New function naming
* Implement defer statements to stop running services
* Clean up crawler / prospector stopping

### DIFF
--- a/filebeat/beater/filebeat.go
+++ b/filebeat/beater/filebeat.go
@@ -9,18 +9,15 @@ import (
 	cfg "github.com/elastic/beats/filebeat/config"
 	"github.com/elastic/beats/filebeat/crawler"
 	"github.com/elastic/beats/filebeat/input"
+	"github.com/elastic/beats/filebeat/publish"
+	"github.com/elastic/beats/filebeat/registrar"
+	"github.com/elastic/beats/filebeat/spooler"
 )
 
 // Filebeat is a beater object. Contains all objects needed to run the beat
 type Filebeat struct {
-	FbConfig *cfg.Config
-	// Channel from harvesters to spooler
-	publisherChan chan []*input.FileEvent
-	spooler       *Spooler
-	registrar     *crawler.Registrar
-	crawler       *crawler.Crawler
-	pub           logPublisher
-	done          chan struct{}
+	config *cfg.Config
+	done   chan struct{}
 }
 
 // New creates a new Filebeat pointer instance.
@@ -32,14 +29,13 @@ func New() *Filebeat {
 func (fb *Filebeat) Config(b *beat.Beat) error {
 
 	// Load Base config
-	err := b.RawConfig.Unpack(&fb.FbConfig)
-
+	err := b.RawConfig.Unpack(&fb.config)
 	if err != nil {
 		return fmt.Errorf("Error reading config file: %v", err)
 	}
 
 	// Check if optional config_dir is set to fetch additional prospector config files
-	fb.FbConfig.FetchConfigs()
+	fb.config.FetchConfigs()
 
 	return nil
 }
@@ -47,7 +43,6 @@ func (fb *Filebeat) Config(b *beat.Beat) error {
 // Setup applies the minimum required setup to a new Filebeat instance for use.
 func (fb *Filebeat) Setup(b *beat.Beat) error {
 	fb.done = make(chan struct{})
-
 	return nil
 }
 
@@ -55,50 +50,65 @@ func (fb *Filebeat) Setup(b *beat.Beat) error {
 func (fb *Filebeat) Run(b *beat.Beat) error {
 
 	var err error
-
-	// Init channels
-	fb.publisherChan = make(chan []*input.FileEvent, 1)
+	config := fb.config.Filebeat
 
 	// Setup registrar to persist state
-	fb.registrar, err = crawler.NewRegistrar(fb.FbConfig.Filebeat.RegistryFile)
+	registrar, err := registrar.New(config.RegistryFile)
 	if err != nil {
 		logp.Err("Could not init registrar: %v", err)
 		return err
 	}
 
-	fb.crawler = &crawler.Crawler{
-		Registrar: fb.registrar,
-	}
+	// Channel from harvesters to spooler
+	publisherChan := make(chan []*input.FileEvent, 1)
 
-	// Load the previous log file locations now, for use in prospector
-	err = fb.registrar.LoadState()
-	if err != nil {
-		logp.Err("Error loading state: %v", err)
-		return err
-	}
+	// Publishes event to output
+	publisher := publish.New(config.PublishAsync,
+		publisherChan, registrar.Channel, b.Publisher.Connect())
 
 	// Init and Start spooler: Harvesters dump events into the spooler.
-	fb.spooler = NewSpooler(fb.FbConfig.Filebeat, fb.publisherChan)
-
+	spooler, err := spooler.New(config, publisherChan)
 	if err != nil {
 		logp.Err("Could not init spooler: %v", err)
 		return err
 	}
 
-	fb.registrar.Start()
-	fb.spooler.Start()
-
-	err = fb.crawler.Start(fb.FbConfig.Filebeat.Prospectors, fb.spooler.Channel)
+	crawler, err := crawler.New(spooler, config.Prospectors)
 	if err != nil {
+		logp.Err("Could not init crawler: %v", err)
 		return err
 	}
 
-	// Publishes event to output
-	fb.pub = newPublisher(fb.FbConfig.Filebeat.PublishAsync,
-		fb.publisherChan, fb.registrar.Channel, b.Publisher.Connect())
-	fb.pub.Start()
+	// The order of starting and stopping is important. Stopping is inverted to the starting order.
+	// The current order is: registrar, publisher, spooler, crawler
+	// That means, crawler is stopped first.
 
-	// Blocks progressing
+	// Start the registrar
+	err = registrar.Start()
+	if err != nil {
+		logp.Err("Could not start registrar: %v", err)
+	}
+	// Stopping registrar will write last state
+	defer registrar.Stop()
+
+	// Start publisher
+	publisher.Start()
+	// Stopping publisher (might potentially drop items)
+	defer publisher.Stop()
+
+	// Starting spooler
+	spooler.Start()
+	// Stopping spooler will flush items
+	defer spooler.Stop()
+
+	err = crawler.Start(registrar.GetStates())
+	if err != nil {
+		return err
+	}
+	// Stop crawler -> stop prospectors -> stop harvesters
+	defer crawler.Stop()
+
+	// Blocks progressing. As soon as channel is closed, all defer statements come into play
 	<-fb.done
 
 	return nil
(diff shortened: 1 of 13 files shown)


urso:
> Registrar is now stopped before publisher is closed. Could this have any side affects?

Yes. Potential side effect is, publisher still returning ACKed lines, the registrar can not handle anymore. Effect is lines need to be send again one restart. In general I like to treat pipelines being setup, configured + stopped in a stack like manner (makes it easier to reason about potential resource usage between stages).

Didn't check yet, but when stopping the registrar, is some channel that has been used by publisher to forward results to registrar being closed? In this case we're asking for a panic on shutdown.

filebeat/beater/filebeat.go
```diff
@@ -9,18 +9,17 @@ import (
...
-	// Channel from harvesters to spooler
-	publisherChan chan []*input.FileEvent
-	spooler       *Spooler
-	registrar     *crawler.Registrar
-	crawler       *crawler.Crawler
-	pub           logPublisher
-	done          chan struct{}
+	crawler  *crawler.Crawler
+	pub      publish.LogPublisher
+	done     chan struct{}
```
urso: uh, I like how this struct has become smaller. Can we rename FbConfig to Config at some point?

ruflin: Yes, it also hurts me every time I see the variables ;-)

filebeat/beater/filebeat.go
```diff
@@ -56,46 +55,37 @@ func (fb *Filebeat) Run(b *beat.Beat) error {
 
 	var err error
 
-	// Init channels
-	fb.publisherChan = make(chan []*input.FileEvent, 1)
+	// Channel from harvesters to spooler
+	publisherChan := make(chan []*input.FileEvent, 1)
```
urso: publisherChan is only required by spooler and publisher. With spooler pushing to publisher can we:
- initialize publisher before spooler
- remove publisherChan and instead give some callback/interface into spooler?

Thinking about init/stop order. How about using `defer X.Stop()` right after creating some async component  `(*Filebeat).Run()` method. Like:

```
X, err := X.New(...)
if err != nil {
    logp.Err(...)
    return err
}
defer X.Stop()
```

The Run method blocks until `fb.done` is closed, which is the case when `(*Filebeat).Stop()` is called. Advantage I see here:
- if some component can not be initialized, all components already started will be closed on shutdown
- makes it easier to maintain stack-like setup/shutdown order due to all Stop methods being close to initialization

filebeat/beater/filebeat.go
```diff
@@ -56,46 +55,37 @@ func (fb *Filebeat) Run(b *beat.Beat) error {
...
 
 	// Setup registrar to persist state
-	fb.registrar, err = crawler.NewRegistrar(fb.FbConfig.Filebeat.RegistryFile)
+	registrar, err := registrar.NewRegistrar(fb.FbConfig.Filebeat.RegistryFile)
```
urso: I like the registrar finally being a separate packet. A more golang idiomatic way of initializing the registrar is `registrar.New(...)`.

filebeat/beater/filebeat.go
```diff
@@ -56,46 +55,37 @@ func (fb *Filebeat) Run(b *beat.Beat) error {
...
 	}
 
-	fb.crawler = &crawler.Crawler{
-		Registrar: fb.registrar,
-	}
-
-	// Load the previous log file locations now, for use in prospector
-	err = fb.registrar.LoadState()
+	// Init and Start spooler: Harvesters dump events into the spooler.
+	spooler, err := spooler.NewSpooler(fb.FbConfig.Filebeat, publisherChan)
```
urso: some. too much stuttering:
- spooler.NewSpooler
- fb.FbConfig.Filebeat ... uhm

filebeat/beater/filebeat.go
```diff
@@ -56,46 +55,37 @@ func (fb *Filebeat) Run(b *beat.Beat) error {
...
 	if err != nil {
-		logp.Err("Error loading state: %v", err)
+		logp.Err("Could not init spooler: %v", err)
 		return err
 	}
 
-	// Init and Start spooler: Harvesters dump events into the spooler.
-	fb.spooler = NewSpooler(fb.FbConfig.Filebeat, fb.publisherChan)
-
+	fb.crawler, err = crawler.NewCrawler(registrar, spooler, fb.FbConfig.Filebeat.Prospectors)
```
urso: more stuttering

filebeat/beater/filebeat.go
urso: uh, we don't even need to store crawler/publisher in fb struct. All components + shutdown order can become local to Run method.

filebeat/crawler/crawler.go
```diff
@@ -23,27 +24,50 @@ import (
...
-	wg          sync.WaitGroup
+	registrar         *registrar.Registrar
+	prospectors       []*prospector.Prospector
+	wg                sync.WaitGroup
+	spooler           *spooler.Spooler
+	prospectorConfigs []*common.Config
 }
 
-func (c *Crawler) Start(prospectorConfigs []*common.Config, eventChan chan *input.FileEvent) error {
+func NewCrawler(registrar *registrar.Registrar, spooler *spooler.Spooler, prospectorConfigs []*common.Config) (*Crawler, error) {
```
urso: let's rename this to `New`.

filebeat/crawler/crawler.go
```diff
@@ -23,27 +24,50 @@ import (
...
+	}, nil
+}
+
+func (c *Crawler) Start() error {
+
+	// Start the registrar
+	c.registrar.Start()
+
+	// Starting spooler
+	c.spooler.Start()
```
urso: let's move 'Start' back into `(*Filebeat).Run`. The crawler should not care about initializing the registrar. I would even consider not passing the registrar and spooler itself, but using some interface for abstracting away the presence of spooler/registrar.

Is registrar even needed by crawl beside 'setup' ?

urso: In general, let's get rid of passing channels around. Rather pass some interface abstracting away the presence of channels. It's still ok if actual implementation uses channels.

filebeat/crawler/crawler.go
```diff
@@ -82,4 +106,10 @@ func (c *Crawler) Stop() {
 	}
```
urso: can we change the go routine to:

```
go func(p Prospector){
    defer c.wg.Done()
    p.Stop
}(prospector)
```

The prospector shouldn't need to know someone is waiting for stop having finished. Simple form of abstraction leak.

filebeat/prospector/prospector_stdin.go
```diff
@@ -8,16 +8,15 @@ import (
 )
 
 type ProspectorStdin struct {
-	Prospector *Prospector
-	harvester  *harvester.Harvester
-	started    bool
+	harvester *harvester.Harvester
+	started   bool
```
urso: why we need 'started'? How often do we expect someone to call Run in ProspectorStdin?

filebeat/publish/publish.go
```diff
@@ -61,11 +61,11 @@ const (
 	batchCanceled
 )
 
-func newPublisher(
+func NewPublisher(
```
urso: let's rename `publisher.NewPublisher` to `publisher.New`.

filebeat/spooler/spooler.go
```diff
@@ -34,7 +34,7 @@ type Spooler struct {
 func NewSpooler(
```
urso: rename `spooler.NewSpooler` to `spooler.New`.

urso:
In general I really like the changes. But start/stop handling in `(*Filebeat).Run()` needs some cleanup within this PR. I'm pretty much in favour of using `defer X.Stop()` right after starting some component.

filebeat/beater/filebeat.go
ruflin: Renamed to `config`. We have `beatConfig` in the generator. But I now even prefer config.

filebeat/beater/filebeat.go
ruflin: will change all accordingly. Now that each is int its own package this makes much more sense.

filebeat/prospector/prospector_stdin.go
ruflin: I remember I introduced it because there was an issue in the past that it was started twice. Will have to check in more detail if it is still needed. Lots of changes have happened ;-) Will put in my list for future clean ups.

filebeat/beater/filebeat.go
ruflin: I changed it and Stop now only closes the channel. Really nice ...

urso: lets move `publisherChan := ...` right before `publisher.New`.

filebeat/beater/filebeat.go
```diff
@@ -55,50 +51,62 @@ func (fb *Filebeat) Setup(b *beat.Beat) error {
...
+
+	// Start the registrar
+	registrar.Start()
+	// Stopping registrar will write last state
+	defer registrar.Stop()
+
+	// Start publisher
+	publisher.Start()
+	// Stopping publisher (might potentially drop items)
+	defer publisher.Stop()
```
urso: with spooler being stopped before publisher, we really might drop items? Hmm... maybe if `publish_async: true` in config file.

filebeat/crawler/crawler.go
```diff
@@ -23,27 +24,44 @@ import (
...
+func (c *Crawler) Start() error {
+
+	logp.Info("Loading Prospectors: %v", len(c.prospectorConfigs))
+
+	// Load the previous log file locations now, for use in prospector
+	err := c.registrar.LoadState()
+	if err != nil {
+		logp.Err("Error loading state: %v", err)
+		return err
+	}
```
urso: doing regsitrar.LoadState in Crawler feels a little out of place. Can we load state when starting/initializing registrar? Either in registrar itself or (*Filebeat).Run()

filebeat/crawler/crawler.go
```diff
@@ -23,27 +24,44 @@ import (
...
 
 	// Get existing states
-	states := *c.Registrar.state
+	states := *c.registrar.GetStates()
```
urso: states only required to startup prosepctors? Registrar only required in order to call GetStates on startup? How about passing states as argument to Start instead of passing registrar to `crawler.New` ?

filebeat/crawler/crawler.go
ruflin: done

filebeat/beater/filebeat.go
ruflin: We now have the flexibility to make optimizations here if needed. Currently the logic should be identical to what we had before.

filebeat/crawler/crawler.go
ruflin: Agree, moved it to the start of the Registrar.

filebeat/crawler/crawler.go
ruflin: done

urso:
LGTM. Let's wait for travis/appveyor

PR #1896, opened 2016-06-22
Cleanup Harvester and Input parts of Filebeat

ruflin:
* Cleanup defaults by moving them to config
* Move readers and sources into their own packages
* Add some minor comments
* Move state to its own package
* Move multiline and json to its own file
* Rename FileState to State as now package name is included


filebeat/harvester/config.go
```diff
@@ -10,18 +10,30 @@ import (
...
+	DefaultHarvesterBufferSize int = 16 << 10 // 16384
+	DefaultDocumentType            = "log"
+	DefaultTailFiles               = false
+	DefaultBackoff                 = 1 * time.Second
+	DefaultBackoffFactor           = 2
+	DefaultMaxBackoff              = 10 * time.Second
+	DefaultCloseOlder              = 1 * time.Hour
+	DefaultForceCloseFiles         = false
+	DefaultMaxBytes                = 10 * (1 << 20) // 10MB
+)
```
urso: we still need the consts if defaultConfig already contains all defaults?

filebeat/harvester/log.go
```diff
@@ -30,15 +35,15 @@ func (h *Harvester) Harvest() {
...
+	readerConfig := reader.LogFileReaderConfig{
+		ForceClose:         cfg.ForceCloseFiles,
+		CloseOlder:         cfg.CloseOlder,
+		BackoffDuration:    cfg.Backoff,
+		MaxBackoffDuration: cfg.MaxBackoff,
+		BackoffFactor:      cfg.BackoffFactor,
 	}
 
-	reader, err := createLineReader(
+	rd, err := reader.CreateLineReader(
```
urso: the reader pipeline being build via `CreateLine` is custom to log based harvester. Same goes for `LogFileReaderConfig`. For turning prospectors/harvesters into more modular form, every harvester should setup it's own reader pipeline. Plus have potentially custom file state (for supporting other types then tail). That is, I'd prefer some `havester/logs` package with `LogFileReaderConfig` and `CreateLineReader` being  private to `harvester/log` package.

filebeat/harvester/log.go
```diff
@@ -265,6 +267,6 @@ func (h *Harvester) close() {
 		h.file.Close()
 		logp.Debug("harvester", "Stopping harvester, closing file: %s", h.Path)
 	} else {
-		logp.Debug("harvester", "Stopping harvester, NOT closing file as file info not available: %s", h.Path)
+		logp.Warn("harvester", "Stopping harvester, NOT closing file as file info not available: %s", h.Path)
```
urso: what does this actually mean? Any chance we're leaking file descriptors (besides file being finally closed when GC'ed, but this GC doesn't count)?

filebeat/harvester/reader/reader.go
```diff
@@ -1,42 +1,42 @@
...
 	"os"
 	"time"
 
-	"github.com/elastic/beats/filebeat/input"
+	"github.com/elastic/beats/filebeat/harvester/source"
+	"github.com/elastic/beats/filebeat/input/file"
 	"github.com/elastic/beats/libbeat/logp"
 )
 
 type logFileReader struct {
```
urso: `logFileReader` is super-specialized to `harvester/log`

filebeat/harvester/source/source.go
```diff
@@ -0,0 +1,17 @@
...
+
+type LogSource interface {
+	io.ReadCloser
+	Name() string
+}
+
+type FileSource interface {
+	LogSource
+	Stat() (os.FileInfo, error)
+	Continuable() bool // can we continue processing after EOF?
```
urso: I don't like this Continuable. It's here due to stdin and log-file tailing using same reader implementation. A custom reader pipeline for files and stdin each will not require this 'Continuable' callback. Custom readers will make `FileSource` interface obsolete.

Reason for FileSource to exist (instead of plain LogSource) is requirement of `Stat()` in harvester when dealing with truncated files. I think with recent refactorings, Stat() is not required by harvester loop anymore.

urso: Same reason, pipeSource (now source.Pipe) exists. In order to massage `stdin` to work well `harvester/log` it's reader and truncation checks without generating panics.

filebeat/input/file/state.go
```diff
@@ -8,41 +8,41 @@ import (
 	"github.com/elastic/beats/libbeat/logp"
 )
 
-// FileState is used to communicate the reading state of a file
-type FileState struct {
+// State is used to communicate the reading state of a file
+type State struct {
```
urso: at some point I'd like to turn State into some very minimal interface. different prospector types might have different needs.

filebeat/registrar/registrar.go
```diff
@@ -83,23 +83,23 @@ func (r *Registrar) loadStates() error {
...
+	oldStates := r.loadAndConvertOldState(f)
 	if oldStates {
 		return nil
 	}
 
-	decoder := json.NewDecoder(file)
-	states := []input.FileState{}
+	decoder := json.NewDecoder(f)
+	states := []file.State{}
 	decoder.Decode(&states)
```
urso: add missing error check here. What's the content of states if registry has been corrupted?

filebeat/registrar/registrar.go
```diff
@@ -214,21 +214,21 @@ func (r *Registrar) writeRegistry() error {
...
 	if e != nil {
 		logp.Err("Failed to create tempfile (%s) for writing: %s", tempfile, e)
 		return e
 	}
 
 	states := r.states.GetStates()
 
-	encoder := json.NewEncoder(file)
+	encoder := json.NewEncoder(f)
 	encoder.Encode(states)
```
urso: still check for encoding errors. you never know.

packetbeat/packetbeat.yml
```diff
@@ -11,7 +11,7 @@
 
 # Select the network interface to sniff the data. You can use the "any"
 # keyword to sniff on all connected interfaces.
-packetbeat.interfaces.device: any
+packetbeat.interfaces.device: lo0
```
urso: ups?

ruflin: ups

filebeat/registrar/registrar.go
ruflin: done

filebeat/harvester/processor/line.go
```diff
@@ -0,0 +1,90 @@
...
+
+	"github.com/elastic/beats/filebeat/harvester/encoding"
+	"github.com/elastic/beats/libbeat/common"
+)
+
+// LineSource produces lines by reading lines from an io.Reader
+// through a decoder converting the reader it's encoding to utf-8.
+type LineSource struct {
+	reader *encoding.LineReader
+}
```
urso: `LineSource` is basically an adapter for converting an `io.Reader` into a `LineProcessor` producing lines. Alternative names `LinesAdapter` or `LineProducer`. Perfectly valid to move LineSource into separate go file and only define interfaces in `line.go`.

filebeat/registrar/registrar.go
ruflin: done

filebeat/input/file/state.go
ruflin: Agree, I would also like to reuse it for winlogbeat

filebeat/harvester/processor/line.go
```diff
@@ -0,0 +1,90 @@
...
+// getting a new line.
+// Next returns the line being read or and error. EOF is returned
+// if processor will not return any new lines on subsequent calls.
+type LineProcessor interface {
+	Next() (Line, error)
+}
+
+// StripNewline processor removes the last trailing newline characters from
+// read lines.
+type StripNewline struct {
```
urso: `StripNewLine` is a special `LineProcessor` removing newline character. Can be moved into separate go file.

filebeat/harvester/processor/line.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+// isLine checks if the given byte array is a line, means has a line ending \n
+func isLine(l []byte) bool {
+	return l != nil && len(l) > 0 && l[len(l)-1] == '\n'
+}
+
+// lineEndingChars returns the number of line ending chars the given by array has
+// In case of Unix/Linux files, it is -1, in case of Windows mostly -2
+func lineEndingChars(l []byte) int {
```
urso: StripLines helper.

filebeat/harvester/processor/line.go
```diff
@@ -0,0 +1,90 @@
...
+
+	L := line.Content
+	line.Content = L[:len(L)-lineEndingChars(L)]
+	return line, err
+}
+
+// isLine checks if the given byte array is a line, means has a line ending \n
+func isLine(l []byte) bool {
+	return l != nil && len(l) > 0 && l[len(l)-1] == '\n'
+}
```
urso: is `isLine` even used?

filebeat/harvester/processor/line.go
urso: For getting proper multiline json support one has to implement `JsonProducer` adapter wrapping io.Reader with json.Decoder. dealing with encoding might still be required, as golang requires UTF-8, but json can be UTF-16 as well.

filebeat/input/file/state.go
ruflin: BTW: This interface should be provided / used by registrar then to persist any state. So it is available to all beats.

filebeat/harvester/reader/reader.go
ruflin: My current intention is to group all readers together so each harvester could make use of the readers.

filebeat/harvester/config.go
ruflin: No, removing it.

filebeat/harvester/log.go
ruflin: I'm a little bit undecided here TBH because I think parts could also be used for a binary file harvester for example. I suggest to keep it in a separate package for now and move it to the right place as soon as we have more harvester / reader types?

filebeat/harvester/log.go
ruflin: This normally happens, when a file cannot be opened for example because filebeat does not have the access rights. The error what exactly happened should be printed out previously when the harvester tried to open the file.

I'm also worried about us leaking file descriptions are there seem to be especially on windows related to this, but so far I haven't seen this log messages in the logs when it happened.

filebeat/harvester/source/source.go
ruflin: I prefer not do logic changes when I do restructuring to make potential problems easier to track, so I would keep this for a future PR. To track all this stuff, I see two possibilities: Meta Github issue or TODO in the code. What do you prefer? Github issue is probably better as it is easier to track progress.

I checked Stat() and Continuable() and it seems like both are still used in the reader.

I get the feeling as soon as we fully restructured harvester the above stuff will cleanup up by itself.

filebeat/harvester/processor/line.go
ruflin: WDYT about `LineEncoder` as this is what it actually does?

filebeat/harvester/processor/line.go
ruflin: done

filebeat/harvester/processor/line.go
ruflin: moved to new `strip_newline.go` file.

filebeat/harvester/processor/line.go
ruflin: Yes, in lineEndingChars

ruflin:
jenkins, retest it

ruflin:
jenkins, retest it

filebeat/harvester/config.go
```diff
@@ -6,42 +6,42 @@ import (
...
-	"github.com/elastic/beats/filebeat/input"
+	"github.com/elastic/beats/filebeat/harvester/processor"
 	"github.com/elastic/beats/libbeat/common"
 )
 
 var (
 	defaultConfig = harvesterConfig{
-		BufferSize:      cfg.DefaultHarvesterBufferSize,
-		DocumentType:    cfg.DefaultDocumentType,
+		BufferSize:      16 << 10, // 16384
```
urso: While I can read it, I like `16 * (1 << 10)`. Or give `1<<10` a name like, `sz1K = (1<<10)`

urso: something like time.Second would be nice. e.g. `sizes.K` or `sizes.M`.

filebeat/harvester/log.go
```diff
@@ -151,24 +157,22 @@ func (h *Harvester) shouldExportLine(line string) bool {
...
-	var file *os.File
-	var err error
 	var encoding encoding.Encoding
 
-	file, err = input.ReadOpen(h.Path)
+	f, err := file.ReadOpen(h.Path)
 	if err == nil {
 		// Check we are not following a rabbit hole (symlinks, etc.)
-		if !input.IsRegularFile(file) {
+		if !file.IsRegularFile(f) {
```
urso: `s/file.IsRegularFile/file.IsRegular/`

filebeat/harvester/log.go
urso: we can try by not giving permissions for some file and do lsof. But prospector should have tripped before when getting stats.

filebeat/harvester/config.go
ruflin: I think the easiest solution here is to just write 16384. I don't think it 16 << 10 or any of the other variants brings a benefit.

filebeat/harvester/log.go
ruflin: Adjusting ...

filebeat/harvester/log.go
ruflin: What happens if file rights change during harvesting? It should not lead to panic on the harvester ...

filebeat/harvester/config.go
urso: it's easier to get the number wrong when using `16384` over `16 * units.Kibi` for example.

ruflin: Will change it

PR #1898, opened 2016-06-22
Fix/overlapping tcp segments

urso:


packetbeat/protos/tcp/tcp.go
```diff
@@ -165,9 +173,25 @@ func (tcp *Tcp) Process(id *flows.FlowID, tcphdr *layers.TCP, pkt *protos.Packet
 				}
 			}
 		}
+
+		// cut away packet overlap
+		if tcpSeqBefore(tcpStartSeq, lastSeq) {
+			delta := lastSeq - tcpStartSeq
+
+			// if 'overlap' already covered by previous packet -> return
+			if int(delta) >= len(pkt.Payload) {
```
ruflin: Should we add at least a debug log message here?

ruflin:
LGTM. Worth backporting?

packetbeat/protos/tcp/tcp.go
urso: debug message is ok. but this is perfectly valid behavior.

urso:
Yeah. in presence of overlapping resends, application layer parser are likely to trip or generate panic.

packetbeat/protos/tcp/tcp.go
urso: rethinking this case, it's impossible to trigger this condition due to tcpSeq already compared to lastSeq and if reset, packet is dropped.

ruflin:
I added the backport layer. There seem to be quite some packetbeat system tests that are failing ...

PR #1903, opened 2016-06-24
Config cleanup

ruflin:
* Code cleanup
* Add humanize for buffersize

### DIFF
--- a/filebeat/beater/filebeat.go
+++ b/filebeat/beater/filebeat.go
@@ -22,7 +22,10 @@ type Filebeat struct {
 
 // New creates a new Filebeat pointer instance.
 func New() *Filebeat {
-	return &Filebeat{}
+	config := cfg.DefaultConfig
+	return &Filebeat{
+		config: &config,
+	}
 }
 
 // Config setups up the filebeat configuration by fetch all additional config files
--- a/filebeat/config/config.go
+++ b/filebeat/config/config.go
@@ -14,10 +14,17 @@ import (
 
 // Defaults for config variables which are not set
 const (
-	DefaultRegistryFile string        = "registry"
-	DefaultSpoolSize    uint64        = 2048
-	DefaultIdleTimeout  time.Duration = 5 * time.Second
-	DefaultInputType                  = "log"
+	DefaultInputType = "log"
+)
+
+var (
+	DefaultConfig = Config{
+		FilebeatConfig{
+			RegistryFile: "registry",
+			SpoolSize:    2048,
+			IdleTimeout:  5 * time.Second,
+		},
+	}
 )
 
 type Config struct {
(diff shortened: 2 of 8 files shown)


filebeat/beater/filebeat.go
```diff
@@ -22,7 +22,9 @@ type Filebeat struct {
 
 // New creates a new Filebeat pointer instance.
 func New() *Filebeat {
-	return &Filebeat{}
+	return &Filebeat{
+		config: &cfg.DefaultConfig,
```
urso: don't take the default config as pointer. DefaultConfig normally serves as template to be copied and must not contain any pointers!

Rather allocate new config initialized with DefaultConfig like:

```
config := cfg.DefaultConfig
return &Filebeat{ config: &config }
```

ruflin: Fixed

filebeat/config/config.go
```diff
@@ -26,9 +33,9 @@ type Config struct {
 
 type FilebeatConfig struct {
 	Prospectors  []*common.Config `config:"prospectors"`
-	SpoolSize    uint64           `config:"spool_size"`
+	SpoolSize    uint64           `config:"spool_size" validate:"nonzero,min=0"`
```
urso: nonzero + min=0? Why not `min=1`? Alternative one can use `positive,nonzero`.

filebeat/spooler/spooler.go
```diff
@@ -36,23 +36,13 @@ func New(
...
-	}
 
 	return &Spooler{
 		Channel:       make(chan *input.FileEvent, channelSize),
-		idleTimeout:   idleTimeout,
+		idleTimeout:   config.IdleTimeout,
 		spoolSize:     spoolSize,
 		exit:          make(chan struct{}),
-		nextFlushTime: time.Now().Add(idleTimeout),
+		nextFlushTime: time.Now().Add(config.IdleTimeout),
```
urso: we need nextFlushTime to be in struct?

filebeat/config/config.go
ruflin: changed to `min=1`

ruflin:
@urso New version pushed.

filebeat/config/config.go
```diff
@@ -26,9 +33,9 @@ type Config struct {
 
 type FilebeatConfig struct {
 	Prospectors  []*common.Config `config:"prospectors"`
-	SpoolSize    uint64           `config:"spool_size"`
+	SpoolSize    uint64           `config:"spool_size" validate:"min=1"`
 	PublishAsync bool             `config:"publish_async"`
-	IdleTimeout  time.Duration    `config:"idle_timeout"`
+	IdleTimeout  time.Duration    `config:"idle_timeout" validate:"nonzero,min=0"`
```
urso: `min=1` ?

urso: what if I want to disable `idle_time`, e.g. via -1 or 0?

ruflin: Not sure if that is something that we should encourage. That above allows to set it for example to 0.001s which is almost like disabling it. 

One thing I just realised related to duration and ucfg: it should be `min=1s` and not `min=1` as if the unit is not know, it could be "anything".

urso: that's right. if not setting any unit, the default is 's'. But normally I set units.

urso: but in past `idle_timeout' was disablable, no?

ruflin: @urso didn't know it allows to set units there. Nice :-) Will add units.

ruflin: default before (and now) is 5s

PR #1908, opened 2016-06-24
Remove MongoDB from the default config

tsg:
And comment it out in the full config.

### DIFF
--- a/metricbeat/module/mongodb/_meta/config.yml
+++ b/metricbeat/module/mongodb/_meta/config.yml
@@ -1,9 +1,9 @@
-- module: mongodb
-  metricsets: ["status"]
-  enabled: true
-  period: 10s
+#- module: mongodb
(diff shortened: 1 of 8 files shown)


metricbeat/module/mongodb/_meta/config.yml
```diff
@@ -1,9 +1,9 @@
-- module: mongodb
-  metricsets: ["status"]
-  enabled: true
-  period: 10s
+#- module: mongodb
+  #metricsets: ["status"]
+  #enabled: true
```
andrewkroh: If we have `enabled: false` available here should we just use that instead of commenting out? // cc: @ruflin

tsg: I find the comments more visually clear. Also, this is inline with the rest of the modules.

PR #1909, opened 2016-06-27
Add close_removed and close_renamed, deprecate force_close_files

ruflin:
force_close_files is replaced by the two option close_removed and close_renamed. Force_close_files is deprecated. In case it is enabled, it sets close_removed and close_renamed to true.

This is part of https://github.com/elastic/beats/issues/1600

### DIFF
--- a/filebeat/filebeat.full.yml
+++ b/filebeat/filebeat.full.yml
@@ -166,14 +166,14 @@ filebeat.prospectors:
   # The backoff value will be multiplied each time with the backoff_factor until max_backoff is reached
   #backoff_factor: 2
 
-  # Close renamed means a file handler is close when a file is renamed / rotated. In case the harvester was
-  # not finished reading the roated file, the file will be picked up again after scan_frequency in case it
-  # also matches the prospector patterns.
+  # Close renamed ensures a file handler is closed when the file is renamed or rotated.
+  # Note: Potential data loss if renamed file is not picked up by prospector.
   #close_renamed: false
 
   # When enabling this option, a file handler is closed immidiately in case a file can't be found
-  # any more. In case the file shows up again later, harvesting will continue at the last position
-  # after scan_frequency
+  # any more. In case the file shows up again later, harvesting will continue at the last known position
+  # after scan_frequency.
+  # Note: Potential data loss if file reading was not finished when file was removed.
   #close_removed: false
 
 #----------------------------- Stdin prospector -------------------------------
--- a/filebeat/filebeat.full.yml
+++ b/filebeat/filebeat.full.yml
@@ -166,22 +166,22 @@ filebeat.prospectors:
   # The backoff value will be multiplied each time with the backoff_factor until max_backoff is reached
   #backoff_factor: 2
 
-  # This option closes a file, as soon as the file name changes.
-  # This config option is recommended on windows only. Filebeat keeps the files it's reading open. This can cause
-  # issues when the file is removed, as the file will not be fully removed until also Filebeat closes
-  # the reading. Filebeat closes the file handler after ignore_older. During this time no new file with the
-  # same name can be created. Turning this feature on the other hand can lead to loss of data
-  # on rotate files. It can happen that after file rotation the beginning of the new
-  # file is skipped, as the reading starts at the end. We recommend to leave this option on false
-  # but lower the ignore_older value to release files faster.
-  #force_close_files: false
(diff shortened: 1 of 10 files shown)


filebeat/filebeat.full.yml
```diff
@@ -166,22 +166,22 @@ filebeat.prospectors:
...
-  # This option closes a file, as soon as the file name changes.
-  # This config option is recommended on windows only. Filebeat keeps the files it's reading open. This can cause
-  # issues when the file is removed, as the file will not be fully removed until also Filebeat closes
-  # the reading. Filebeat closes the file handler after ignore_older. During this time no new file with the
-  # same name can be created. Turning this feature on the other hand can lead to loss of data
-  # on rotate files. It can happen that after file rotation the beginning of the new
-  # file is skipped, as the reading starts at the end. We recommend to leave this option on false
-  # but lower the ignore_older value to release files faster.
-  #force_close_files: false
+  # Close renamed means a file handler is close when a file is renamed / rotated. In case the harvester was
```
urso: 'Close ensures a file handler is closed when the file is renamed or rotated'.

'Note: Potential data loss if renamed file is not picked up by prospector'.

filebeat/filebeat.full.yml
```diff
@@ -166,22 +166,22 @@ filebeat.prospectors:
...
+  # not finished reading the roated file, the file will be picked up again after scan_frequency in case it
+  # also matches the prospector patterns.
+  #close_renamed: false
+
+  # When enabling this option, a file handler is closed immidiately in case a file can't be found
+  # any more. In case the file shows up again later, harvesting will continue at the last position
```
urso: s/last position/last known position/

- impact on data loss if file was removed from local fs
- option more interesting for dealing with network shares?

filebeat/harvester/config.go
```diff
@@ -20,8 +21,10 @@ var (
 		BackoffFactor:   2,
 		MaxBackoff:      10 * time.Second,
 		CloseOlder:      1 * time.Hour,
-		ForceCloseFiles: false,
 		MaxBytes:        10 * (1 << 20), // 10MB
+		CloseRemoved:    false,
+		CloseRenamed:    false,
+		ForceCloseFiles: false,
```
urso: still in alpha, let's remove 'ForceCloseFiles' then.

ruflin: @urso I want to make the migration to 5.0 as seamless as possible. That is why I would prefer to keep this in.

urso: maybe we want to add a 'deprecated' option to go-ucfg, which will print a warning when option is used.

We can find deprecated options in code by grepping for `config:.*deprecated`. Maybe even with version like: `config:",deprecated(5.0)"`.

filebeat/harvester/reader/log.go
```diff
@@ -113,16 +116,19 @@ func (r *logFileReader) Read(buf []byte) (int, error) {
 			return n, ErrInactive
 		}
 
-		if r.config.ForceClose {
-			// Check if the file name exists (see #93)
+		if r.config.CloseRenamed {
+			if !file.IsSameFile(r.fs.Name(), info) {
```
urso: let's keep the code comment `Check if the file name still exists`.

urso: Shouldn't it be: `check if file has been renamed, but still exists` ?

filebeat/filebeat.full.yml
ruflin: Added accordingly, and move the change to beat.yml :-)

filebeat/filebeat.full.yml
ruflin: @urso I plan to add a full page on all this configuration with more examples in the docs after implementation.

filebeat/harvester/config.go
ruflin: That would be great TBH. Lets do this as soon as support for this is in ucfg (not a priority).

filebeat/harvester/reader/log.go
ruflin: I added `Check if the file has been renamed`. It does not check if the file still exists, it only checks if the file with the same name is this file or an other file. In case no file under the same name exists, it just assumes rename (it could also be removed). But that is checked in the next check.

ruflin: Sorry, changed it again to: `// Check if the file can still be found under the same path
`

ruflin:
@urso review applied and new version pushed.

PR #1914, opened 2016-06-27
Introduce close_eof harvster option

ruflin:
Depends on https://github.com/elastic/beats/pull/1909

### DIFF
--- a/filebeat/etc/beat.full.yml
+++ b/filebeat/etc/beat.full.yml
@@ -176,6 +176,11 @@ filebeat.prospectors:
   # Note: Potential data loss if file reading was not finished when file was removed.
   #close_removed: false
 
+  # Closes the file handler as soon as the harvesters reaches then end of the file.
+  # The file will be picked up again by the harvester at previous known state
+  # after scan_frequency in case the file can still be discovered by the prospector.
+  #close_eof: false
+
 #----------------------------- Stdin prospector -------------------------------
 # Configuration to use stdin input
 #- input_type: stdin
--- a/filebeat/filebeat.full.yml
+++ b/filebeat/filebeat.full.yml
@@ -176,6 +176,11 @@ filebeat.prospectors:
   # Note: Potential data loss if file reading was not finished when file was removed.
   #close_removed: false
 
+  # Closes the file handler as soon as the harvesters reaches then end of the file.
+  # The file will be picked up again by the harvester at previous known state
+  # after scan_frequency in case the file can still be discovered by the prospector.
+  #close_eof: false
+
 #----------------------------- Stdin prospector -------------------------------
 # Configuration to use stdin input
 #- input_type: stdin
--- a/filebeat/harvester/config.go
+++ b/filebeat/harvester/config.go
@@ -26,6 +26,7 @@ var (
 		MaxBytes:        10 * (1 << 20), // 10MB
 		CloseRemoved:    false,
 		CloseRenamed:    false,
+		CloseEOF:        false,
 		ForceCloseFiles: false,
 	}
 )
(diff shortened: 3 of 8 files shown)


filebeat/filebeat.full.yml
```diff
@@ -166,22 +166,27 @@ filebeat.prospectors:
...
+  #close_renamed: false
+
+  # When enabling this option, a file handler is closed immidiately in case a file can't be found
+  # any more. In case the file shows up again later, harvesting will continue at the last position
+  # after scan_frequency
+  #close_removed: false
+
+  # Closes the file handler when the end of the file is reached. This is useful for files where
+  # content is only written to once. In case further content is added to the file later again
+  # it will be picked up with the previous state latest after scan_frequency.
```
urso: In case reading goes faster then writing, Isn't there always a chance of hitting EOF?

filebeat/harvester/log.go
```diff
@@ -68,12 +72,28 @@ func (h *Harvester) Harvest() {
...
+				return
+			}
+
+			if err == reader.ErrRenamed {
+				logp.Info("File was renamed: %s. Closing because close_renamed is enabled.", h.Path)
+				return
+			}
+
+			if err == io.EOF {
+				logp.Info("End of file reached: %s. Closing because close_eof is enabled.", h.Path)
```
urso: doesn't reader return `io.EOF` if `close_older` is reached?

urso: oh, it's ErrInactive. How about using switch statement on error types (as all errors are variables):

```
switch err {
    case reader.ErrFileTruncate:
      ...
    case reader.ErrRemoved:
      logp.Info(...)
    case reader.ErrRenamed:
      logp.Info(...)
    case reader.ErrInactive:
      logp.Info(...)
    case io.EOF:
      logp.Info(...)
    default:
        logp.Info("Read line error: %s", err)
}
return
```

filebeat/filebeat.full.yml
ruflin: Yes. But I assume for files with 1 event, the hole event is flushed at once.

ruflin: Perhaps I should not make an example here in the config file and only explain what it does. The example should be in the docs pages.

ruflin: I changed it to:

```
  # Closes the file handler as soon as the harvesters reaches then end of the file.
  # The file will be picked up again by the harvester at previous known state
  #close_eof: false
```

filebeat/harvester/log.go
ruflin: yes, that is nicer. changed.

filebeat/harvester/config.go
```diff
@@ -20,8 +21,11 @@ var (
 		BackoffFactor:   2,
 		MaxBackoff:      10 * time.Second,
 		CloseOlder:      1 * time.Hour,
-		ForceCloseFiles: false,
 		MaxBytes:        10 * (1 << 20), // 10MB
+		CloseRemoved:    false,
+		CloseRenamed:    false,
+		CloseEof:        false,
```
ruflin: @urso Would you prefer here CloseEOF?

filebeat/harvester/config.go
```diff
@@ -46,6 +53,13 @@ type harvesterConfig struct {
 
 func (config *harvesterConfig) Validate() error {
 
+	// TODO: remove in 7.0
```
ruflin: Ups, that should be 6.0

ruflin:
@urso Changes applied and rebased.

ruflin:
jenkins, test it

ruflin:
@urso I triggered this build manually as Jenkins didn't want to start it initially: http://build-eu-00.elastic.co/job/Beats-PR/2217/console I will ping you when green.

filebeat/filebeat.full.yml
urso: Maybe add note of potential data loss if file is deleted within close and `scan_frequency`

ruflin: Will add it in https://github.com/elastic/beats/pull/1915

filebeat/etc/beat.full.yml
```diff
@@ -176,6 +176,11 @@ filebeat.prospectors:
   # Note: Potential data loss if file reading was not finished when file was removed.
   #close_removed: false
 
+  # Closes the file handler as soon as the harvesters reaches then end of the file.
```
monicasarbu: @ruflin needs rephrase: Closes the file handler as soon as the harvester reaches <the> end of the file

PR #1915, opened 2016-06-27
Introducing ttl and timestamp in state

ruflin:
Each state has a timestamp and a ttl. -1 means ttl is disable, 0 means it should be directly removed.
This moves the logic on what should happen with a state completely to the state itself and makes it possible to use states in different use cases.

The advantage of the ttl is that it does not depend in filebeat on the modification time which can be incorrect. This means, in case a file is rotated, the timestamp of a file is updated and it also counts as a new state. This makes sense as otherwise it could happen that the state of a rotate file is removed and then after rotation the file is picked up again as a completely new file. The downside is that people using filebeat must understand the difference between the state timestamp and modtime. In general timestamp is neweer then the modtime, as filebeat finishes reading later.

On the registrar side, the cleanup happens every time before the registry is written. On the prospector side the state is cleaned up after each scan. It can happen that the prospector state list and registrar state list are not 100% in sync as they don't cleanup the states at the same time. The prospector state is the one that will always overwrite the registrar state. No cleanup is done before shutdown.

It is important, that on startup first a full scan is done to update the states before the state is cleaned up, otherwise still needed states could be removed.

### DIFF
--- a/filebeat/etc/beat.full.yml
+++ b/filebeat/etc/beat.full.yml
@@ -179,7 +179,13 @@ filebeat.prospectors:
   # Closes the file handler as soon as the harvesters reaches then end of the file.
   # The file will be picked up again by the harvester at previous known state
   # after scan_frequency in case the file can still be discovered by the prospector.
+  # Note: Potential data loss if file is deleted / moved before picked up again after
+  # scan_frequency by prospector
   #close_eof: false
+
+  # Files for the modification data is older then clean_older the state from the registry is removed
+  # By default this is disabled.
+  #clean_older: 0
 
 #----------------------------- Stdin prospector -------------------------------
 # Configuration to use stdin input
--- a/filebeat/input/file/state.go
+++ b/filebeat/input/file/state.go
@@ -15,7 +15,8 @@ type State struct {
 	Finished    bool        `json:"-"` // harvester state
 	Fileinfo    os.FileInfo `json:"-"` // the file info
 	FileStateOS StateOS
-	LastSeen    time.Time `json:"last_seen"`
+	Timestamp   time.Time     `json:"timestamp"`
+	TTL         time.Duration `json:"ttl"`
 }
 
 // NewState creates a new file state
@@ -25,7 +26,8 @@ func NewState(fileInfo os.FileInfo, path string) State {
 		Source:      path,
 		Finished:    false,
 		FileStateOS: GetOSState(fileInfo),
-		LastSeen:    time.Now(),
+		Timestamp:   time.Now(),
+		TTL:         -1 * time.Second, // By default, state does have an infinit ttl
 	}
 }
 
@@ -47,7 +49,7 @@ func (s *States) Update(newState State) {
 	defer s.mutex.Unlock()
 
 	index, _ := s.findPrevious(newState)
-	newState.LastSeen = time.Now()
+	newState.Timestamp = time.Now()
 
 	if index >= 0 {
 		s.states[index] = newState
@@ -81,25 +83,30 @@ func (s *States) findPrevious(newState State) (int, State) {
 }
 
 // Cleanup cleans up the state array. All states which are older then `older` are removed
-func (s *States) Cleanup(older time.Duration) {
+func (s *States) Cleanup() {
+
 	s.mutex.Lock()
 	defer s.mutex.Unlock()
 
-	for i, state := range s.states {
+	currentTime := time.Now()
+	states := s.states[:0]
 
-		// File wasn't seen for longer then older -> remove state
-		if time.Since(state.LastSeen) > older {
-			logp.Debug("prospector", "State removed for %s because of older: %s", state.Source)
-			s.states = append(s.states[:i], s.states[i+1:]...)
+	for _, state := range s.states {
+		ttl := state.TTL
+		if ttl >= 0 && currentTime.Sub(state.Timestamp) > ttl {
+			logp.Debug("state", "State removed for %v because of older: %v", state.Source, ttl)
+			continue // drop state
 		}
+		states = append(states, state) // in-place copy old state
 	}
-
+	s.states = states
 }
 
 // Count returns number of states
 func (s *States) Count() int {
 	s.mutex.Lock()
 	defer s.mutex.Unlock()
+
 	return len(s.states)
 }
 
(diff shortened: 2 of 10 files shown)


filebeat/input/file/state.go
```diff
@@ -81,25 +83,36 @@ func (s *States) findPrevious(newState State) (int, State) {
...
-	}
 
+		// States is older then ttl
+		if int64(time.Since(state.Timestamp).Seconds()) > state.Ttl {
+			// This check prevents a panic
+			if len(s.states) == 1 {
+				s.states = []State{}
+			} else {
+				s.states = append(s.states[:i], s.states[i+1:]...)
+			}
```
urso: filter loop alternative (not as many potential copies on sub-splices when removing multiple elements):

```
states := s.states[:0]
for _, state := range.stats {
    ttl := state.Ttl * time.Second
    if ttl >= 0 && time.Since(state.Timestamp) <= ttl {
          debugf("State removed for %v because of older: %v", state.Source, ttl)
          continue // drop state
    }
    states = append(states, state) // in-place copy old state
}
s.states = states
```

When converting ttl into time.Duration it nicely prints the unit being used. time.Duration can be negative too.

filebeat/input/file/state.go
```diff
@@ -25,7 +26,8 @@ func NewState(fileInfo os.FileInfo, path string) State {
 		Source:      path,
 		Finished:    false,
 		FileStateOS: GetOSState(fileInfo),
-		LastSeen:    time.Now(),
+		Timestamp:   time.Now(),
+		Ttl:         -1, // By default, state does have an infinit ttl
```
urso: s/Ttl/TTL/

filebeat/prospector/prospector.go
```diff
@@ -105,6 +105,10 @@ func (p *Prospector) Run() {
 				logp.Info("Prospector channel stopped")
 				return
 			case event := <-p.harvesterChan:
+				// Add ttl if cleanOlder is enabled
+				if p.config.CleanOlder != 0 {
```
urso: clean older can be <0

filebeat/prospector/prospector_log.go
```diff
@@ -48,10 +50,13 @@ func (p *ProspectorLog) Run() {
 	logp.Debug("prospector", "Start next scan")
 
 	p.scan()
-	// Only cleanup states if enabled
-	if p.config.IgnoreOlder != 0 {
-		p.Prospector.states.Cleanup(p.config.IgnoreOlder)
+
+	// It is important that a first scan is run before cleanup to make sure all new states are read first
+	if p.config.CleanOlder != 0 {
```
urso: `p.config.CleanOlder > 0`

filebeat/input/file/state.go
urso: maybe get timestamp before removing, so all states are compared against same timestamp.

Are removed states somehow reported to registrar for remove? If so, how can we be sure some state is not updated by some prospector?

filebeat/input/file/state.go
ruflin: changed

filebeat/prospector/prospector.go
ruflin: no, changed it to >

ruflin: also added a check to the config loading part: 
```
	CleanOlder    time.Duration    `config:"clean_older" validate:"min=0"`
```

filebeat/prospector/prospector_log.go

filebeat/prospector/prospector.go
```diff
@@ -105,6 +105,10 @@ func (p *Prospector) Run() {
...
+					event.FileState.Ttl = int64(p.config.CleanOlder.Seconds())
```
urso: event.FileState only forwarded to spooler and registrar next or also used as is in prospector for cleanup? If shared with prospector we need atomic or mutex operation.

filebeat/input/file/state.go
ruflin: The registrar does the cleanup himself. As all states are reported to prospector and registrar, they will have the same information -> do the same cleanup. The cleanup on the registrar side will always happen with a delay. It can happen in an edge case, that cleanup removes a state on the registrar side that was not removed on the prospector side because a new event already exist. But the new state will be sent to the registrar and reintroduce the state again. Prospector always overwrites registrar state.

We should discuss in a second step on the requirements for variables as some combinations / variables can lead to very strange situations. If clean_older is too small, filebeat will "stateless".

filebeat/prospector/prospector.go
ruflin: It is used to update the state (not cleanup) on line 117 below. Update already has a mutex operation. Is that enough?

urso: Have to read up if/when/how FileState is shared between multiple go-routines. But in case it's shared, this is not safe. Would be better to set the TTL when FileState is generated (as it's not supposed to change).

ruflin: The reason I set it here is that I didn't want the harvester to know about this.

filebeat/prospector/prospector_log.go
```diff
@@ -48,10 +50,13 @@ func (p *ProspectorLog) Run() {
...
-	// Only cleanup states if enabled
-	if p.config.IgnoreOlder != 0 {
-		p.Prospector.states.Cleanup(p.config.IgnoreOlder)
+
+	// It is important that a first scan is run before cleanup to make sure all new states are read first
+	if p.config.CleanOlder > 0 {
+		p.Prospector.states.Cleanup()
+		logp.Debug("prospector", "Prospector states cleaned up.")
 	}
+	p.lastScan = time.Now()
```
ruflin: This should be removed

filebeat/etc/beat.full.yml
```diff
@@ -179,8 +179,14 @@ filebeat.prospectors:
   # Closes the file handler as soon as the harvesters reaches then end of the file.
   # The file will be picked up again by the harvester at previous known state
   # after scan_frequency in case the file can still be discovered by the prospector.
+  # Note: Potential data loss if file is deleted / moved before picked up again after
+  # scan_frequency by prospector
   #close_eof: false
 
+  # Files for the modification data is older then clean_older the state from the registry is removed
```
monicasarbu: @ruflin we need to rephrase this sentence here.

ruflin: Agree. Will take care of it.

PR #1916, opened 2016-06-27
Close changelog alpha4

tsg:


tsg:
Indeed, fixed now.

CHANGELOG.asciidoc
```diff
@@ -87,6 +68,56 @@ https://github.com/elastic/beats/compare/v5.0.0-alpha3...master[Check the HEAD d
 
 ////////////////////////////////////////////////////////////
 
+[[release-notes-5.0.0-alpha4]]
+=== Beats version 5.0.0-alpha4
+https://github.com/elastic/beats/compare/v5.0.0alpha3...v5.0.0-alpha4[View commits]
```
ruflin: there is a dash `-` missing before alpha3. It seems like this is also the case in previous releases ...

tsg:
Pushed a fix for the missing dashes.

PR #1919, opened 2016-06-27
Unify doc version handling

tsg:
The new file `libbeat/docs/version.asciidoc` contains a single line with the
`stack-version` variable. All other version variables reference this one. This
makes it easy to automate version changing.

`make update` is needed to copy the file to the individual Beats. It would have
been possible to include directly the file from libbeat, but that would have had
a big disadvantage: the docs only get rebuild on file changes in the "book", so
changing the version in libbeat wouldn't trigger a docs rebuild on the public
website.

Also, it's kind of nice to have to run `make update` shortly before each release,
that makes it sure we won't forget.

### DIFF
--- a/filebeat/docs/index.asciidoc
+++ b/filebeat/docs/index.asciidoc
@@ -2,12 +2,12 @@
 
 include::./version.asciidoc[]
 
-:libbeat: http://www.elastic.co/guide/en/beats/libbeat/master
-:packetbeat: http://www.elastic.co/guide/en/beats/packetbeat/master
-:metricbeat: http://www.elastic.co/guide/en/beats/metricbeat/master
-:filebeat: http://www.elastic.co/guide/en/beats/filebeat/master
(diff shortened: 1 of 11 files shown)


tsg:
CC @dedemorton

dedemorton:
@tsg Great idea! But shouldn't we also have a variable for the branch so that we can set the paths too? So you would have a variable called something like doc-branch that sets the branch for each doc. For example:
:libbeat: http://www.elastic.co/guide/en/beats/libbeat/{doc-branch}. 

Then in version.asciidoc, we would have:
:doc-branch: master

I'm just concerned that we might update the version without remembering to update the path. We use variables like {libbeat} in links between different books.

tsg:
@dedemorton ah, yes, we need that as well. I added it in a second commit. Can you have a look again, please?

ruflin:
+1 on this idea.

dedemorton:
LGTM. I did not pull the files and test the build, tho. I'll assume you did that. :-)

PR #1921, opened 2016-06-28
Send field as message

treff7es:
Being able to specify a field to kafka output which will be sent as it is to kafka.
I found it a bit tricky to send loglines as it is to kafka from filbeat so I created this change where you can specify which field you want to send to kafka.

Feel free to reject if you don't find it useful. It is useful for us.

### DIFF
--- a/libbeat/outputs/kafka/client.go
+++ b/libbeat/outputs/kafka/client.go
@@ -14,10 +14,11 @@ import (
 )
 
 type client struct {
-	hosts   []string
-	topic   string
-	useType bool
-	config  sarama.Config
+	hosts         []string
+	topic         string
+	useType       bool
+	config        sarama.Config
+	sendFieldOnly string
 
 	producer sarama.AsyncProducer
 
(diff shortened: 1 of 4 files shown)


tsg:
@treff7es thanks for the PR. Could you explain a bit more why you need this? Is the concern about having too much data into Kafka? Or the fact that the JSON needs to be decoded by whatever is reading from the queue?

treff7es:
We have services which heavily rely on existing loglines which are sometimes not in a strucutred (json format) like tab separated. To be able to send our Kafka cluster non structured logs I needed this modification and I wanted to make sure that nobody will overwrite/change any of existing fields in the json logs as well.
I can understand that from elasticsearch perspective the non structured logs might not be practical but sending in to kafka cluster it is as monitoring/batch processing is relying on this format now.

tsg:
Thanks for the details, let us discuss it a bit internally to see if this is a deployment model that we want to support. Will get back to you shortly.

tsg:
Sorry for getting back to you so late. While discussing this we realized that we like the general idea but we'd like a more generic implementation of it that would work for all Beats, for more outputs and that would allow formatting the output rather than just selecting a field. 

I opened #2000 with some details, and we'll implement it sooner or later depending on how much interest we get in it. If you'd like to have a go at the more generic implementation, we'd be happy to assist you, but I'm closing this one for now in favor of #2000.

PR #1922, opened 2016-06-28
Introduce clean_removed

ruflin:


filebeat/input/file/state.go
```diff
@@ -81,25 +83,36 @@ func (s *States) findPrevious(newState State) (int, State) {
...
-			logp.Debug("prospector", "State removed for %s because of older: %s", state.Source)
-			s.states = append(s.states[:i], s.states[i+1:]...)
+		// -1 means ttl is disabled
+		if state.Ttl == -1 {
+			continue
 		}
-	}
 
+		// States is older then ttl
+		if int64(time.Since(state.Timestamp).Nanoseconds()/int64(time.Microsecond)) > state.Ttl {
```
urso: can we transform state.TTL to be time.Duration. Not happy with conversions like to nanos and devide by Microseconds. Which unit is TTL in?

filebeat/prospector/prospector.go
```diff
@@ -105,6 +105,10 @@ func (p *Prospector) Run() {
 				logp.Info("Prospector channel stopped")
 				return
 			case event := <-p.harvesterChan:
+				// Add ttl if cleanOlder is enabled
+				if p.config.CleanOlder != 0 {
+					event.FileState.Ttl = int64(p.config.CleanOlder.Nanoseconds() / int64(time.Microsecond))
```
urso: cleanOlder is already time.Duration + we're comparing durations. One more reason to have `TTL : time.Duration`

filebeat/input/file/state.go
ruflin: I just converted and now realised that it seems like this breaks the json conversion. Any trick here?

```
	TTL         time.Duration     `json:"ttl"`
```

ruflin: time.Duration changes were done in #1915

filebeat/prospector/prospector.go
ruflin: Agree. See #1915

CHANGELOG.asciidoc
```diff
@@ -56,6 +56,7 @@ https://github.com/elastic/beats/compare/v5.0.0-alpha3...master[Check the HEAD d
 - Introduce close_removed and close_renamed harvester options {issue}1600[1600]
 - Introduce close_eof harvester options {issue}1600[1600]
 
+- Add clean_removed config option
```
urso: newline on purpose?

filebeat/filebeat.full.yml
```diff
@@ -179,12 +179,18 @@ filebeat.prospectors:
...
   # after scan_frequency in case the file can still be discovered by the prospector.
+  # Note: Potential data loss if file is deleted / moved before picked up again after
+  # scan_frequency by prospector
   #close_eof: false
 
   # Files for the modification data is older then clean_older the state from the registry is removed
   # By default this is disabled.
   #clean_older: 0
 
+  # Removes the state for file which cannot be found on disk anymore immidiately
```
urso: s/immidiately/immediately/

filebeat/prospector/prospector_log.go
```diff
@@ -56,6 +56,21 @@ func (p *ProspectorLog) Run() {
...
 		logp.Debug("prospector", "Prospector states cleaned up.")
 	}
+
+	// Cleanup of removed files will only happen after next scan. Otherwise it can happen that not all states
+	// were updated before cleanup is called
+	if p.config.CleanRemoved {
+		for _, state := range p.Prospector.states.GetStates() {
+			_, err := os.Stat(state.Source)
+			if err != nil {
+				state.TTL = 0
```
urso: setting state.TTL to 0 is pretty scary, given the comparison uses `time.Since(state.timestamp) >= ttl` which is pretty much always true. Even if file is present.

ruflin: Not sure I can follow here. Setting it to 0 the intention is that it will be removed.

filebeat/filebeat.full.yml
ruflin: fixed

filebeat/prospector/prospector_log.go
urso: Ah, I missed the `err != nil`. But we can not distinguish between file being unavailable and permissions to have changed.

ruflin: Note for me: Lets distinguish between these two:

```
var (
        ErrPermission = errors.New("permission denied")
        ErrNotExist   = errors.New("file does not exist")
)
```

If it does not exist, I think the behaviour is clear. I suggest if we don't have access, we log an error and close the file, but we don't remove it from the registry. @urso WDYT?

ruflin: @urso During implementation I realised, that the above `Stat(path)` command does not return an error, when the file is not readable (which makes kind of sense). An error is only returned, if we try to open the file.

As I suggest to remove the file anyways only if it is really removed, I think the current implemented behaviour is as expected.

urso: Maybe add a comment how/why os.Stat is doing the right thing here.

filebeat/input/file/state.go
```diff
@@ -93,7 +93,7 @@ func (s *States) Cleanup() {
 
 	for _, state := range s.states {
 		ttl := state.TTL
-		if ttl >= 0 && currentTime.Sub(state.Timestamp) > ttl {
+		if ttl >= 0 && currentTime.Sub(state.Timestamp) >= ttl {
```
urso: how about: if `ttl==0 || currentTime.Sub(state.Timestamp) > ttl` ?

ruflin: I would be ok with `ttl==0 || currentTime.Sub(state.Timestamp) >= ttl` I had one test case where state.Timestamp == currentTimestamp ! Not sure how this was able to happen (windows ...)

PR #1926, opened 2016-06-29
Add close_timeout option

ruflin:
close_timeout will end the harvester after the predefined time. In case the output is blocked, close_timeout will only apply on the next event sent. This is identical with the close_* options.

### DIFF
--- a/filebeat/docs/reference/configuration/filebeat-options.asciidoc
+++ b/filebeat/docs/reference/configuration/filebeat-options.asciidoc
@@ -233,6 +233,17 @@ WARNING: Only use this options if you understand that data loss is a potential s
 
 Close eof closes a file as soon as the end of a file is reached. This is useful in case your files are only written once and not updated from time to time. This case can happen in case you are writing every single log event to a new file.
 
+===== close_timeout
+
+WARNING: Only use this options if you understand the potential side affects with potential data loss. In addition it can happen that multiline events are not sent completely on timeout.
+
+Close timeout gives every harvester a predefined lifetime. Independent of the location of the reader, it will stop the reader after `close_timeout`. This option can be useful, if only a predefine time should be spent on older log files. Using this option in combination with `ignore_older` == `close_timeout` means the file is not picked up again in case it wasn't modified in between. This normally leads to data loss and not the complete file is sent.
+
+In case close_timeout is used in combination with multiline events, it can happen that the harvester will be stopped in the middle of a multiline event, means only parts of the event will be sent. In case the harvester is continued at a later stage again and the file still exists, only the second part of the event will be sent.
+
+Close timeout will not apply, in case your output is stuck and no further events can be sent. At least one event must be sent after close_timeout kicks in so the harvester can be closed afterwards.
+
+
 [[clean-options]]
 ===== clean_*
 
--- a/filebeat/harvester/config.go
+++ b/filebeat/harvester/config.go
@@ -27,6 +27,7 @@ var (
 		CloseRemoved:    false,
 		CloseRenamed:    false,
 		CloseEOF:        false,
+		CloseTimeout:    0,
 		ForceCloseFiles: false,
 	}
 )
@@ -46,6 +47,7 @@ type harvesterConfig struct {
 	CloseRemoved         bool                    `config:"close_removed"`
 	CloseRenamed         bool                    `config:"close_renamed"`
 	CloseEOF             bool                    `config:"close_eof"`
+	CloseTimeout         time.Duration           `config:"close_timeout" validate:"min=0"`
 	ForceCloseFiles      bool                    `config:"force_close_files"`
 	ExcludeLines         []*regexp.Regexp        `config:"exclude_lines"`
 	IncludeLines         []*regexp.Regexp        `config:"include_lines"`
--- a/filebeat/harvester/log.go
+++ b/filebeat/harvester/log.go
@@ -5,6 +5,7 @@ import (
 	"expvar"
 	"io"
 	"os"
+	"time"
 
 	"golang.org/x/text/transform"
 
@@ -260,6 +261,7 @@ func (h *Harvester) close() {
 	if h.file != nil {
 
 		h.file.Close()
+
 		logp.Debug("harvester", "Closing file: %s", h.state.Source)
 		harvesterOpenFiles.Add(-1)
 
@@ -294,12 +296,27 @@ func (h *Harvester) newLogFileReader() (reader.Reader, error) {
 	// TODO: NewLineReader uses additional buffering to deal with encoding and testing
 	//       for new lines in input stream. Simple 8-bit based encodings, or plain
 	//       don't require 'complicated' logic.
-	fileReader, err := NewLogFile(h.file, h.config, h.done)
+	h.fileReader, err = NewLogFile(h.file, h.config)
 	if err != nil {
 		return nil, err
 	}
 
-	r, err = reader.NewEncode(fileReader, h.encoding, h.config.BufferSize)
+	// Closes reader after timeout or when done channel is closed
+	go func() {
+		var closeTimeout <-chan time.Time
+		if h.config.CloseTimeout > 0 {
+			closeTimeout = time.After(h.config.CloseTimeout)
+		}
+
+		select {
+		case <-h.done:
+		case <-closeTimeout:
+			logp.Info("Closing harvester because close_timeout was reached: %s", h.state.Source)
+		}
+		h.fileReader.Close()
+	}()
+
+	r, err = reader.NewEncode(h.fileReader, h.encoding, h.config.BufferSize)
 	if err != nil {
 		return nil, err
 	}
--- a/filebeat/harvester/log_file.go
+++ b/filebeat/harvester/log_file.go
@@ -3,6 +3,7 @@ package harvester
 import (
 	"io"
 	"os"
+	"sync"
 	"time"
 
 	"github.com/elastic/beats/filebeat/harvester/source"
@@ -17,12 +18,12 @@ type LogFile struct {
 	lastTimeRead time.Time
 	backoff      time.Duration
 	done         chan struct{}
+	singleClose  sync.Once
 }
 
 func NewLogFile(
 	fs source.FileSource,
 	config harvesterConfig,
-	done chan struct{},
 ) (*LogFile, error) {
 	var offset int64
 	if seeker, ok := fs.(io.Seeker); ok {
@@ -39,7 +40,7 @@ func NewLogFile(
 		config:       config,
 		lastTimeRead: time.Now(),
 		backoff:      config.Backoff,
-		done:         done,
+		done:         make(chan struct{}),
 	}, nil
 }
 
@@ -163,4 +164,12 @@ func (r *LogFile) wait() {
 			r.backoff = r.config.MaxBackoff
 		}
 	}
+}
+
+func (r *LogFile) Close() {
+	// Make sure reader is only closed once
+	r.singleClose.Do(func() {
+		close(r.done)
+		// Note: File reader is not closed here because that leads to race conditions
+	})
 }
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -94,6 +94,8 @@ https://github.com/elastic/beats/compare/v5.0.0-alpha5...master[Check the HEAD d
 *Topbeat*
 
 *Filebeat*
+- Introduce close_timeout harvester options {issue}1600[1600]
+
 
 - Add harvester_limit option {pull}2417[2417]
 
--- a/filebeat/docs/troubleshooting.asciidoc
+++ b/filebeat/docs/troubleshooting.asciidoc
@@ -22,8 +22,9 @@ There are 4 more configuration options which can be used to close file handlers,
 * close_renamed
 * close_removed
 * close_eof
+* close_timeout
 
-`close_renamed` and `close_removed` can be useful on Windows and issues related to file rotation, see <<windows-file-rotation>>. `close_eof` can be useful in environments with a large number of files with only very few entries. More details can be found in config options, see <<configuration-filebeat-options>>.
+`close_renamed` and `close_removed` can be useful on Windows and issues related to file rotation, see <<windows-file-rotation>>. `close_eof` can be useful in environments with a large number of files with only very few entries. `close_timeout` in environments where it is more important to close file handlers then to send all log lines. More details can be found in config options, see <<configuration-filebeat-options>>.
 
 Before using any of these variables, make sure to study the documentation on each.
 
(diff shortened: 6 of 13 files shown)


filebeat/harvester/log.go
```diff
@@ -105,6 +106,11 @@ func (h *Harvester) Harvest() {
 		if !h.sendEvent(event) {
 			return
 		}
+
+		if h.Config.CloseTTL > 0 && time.Since(h.startTime) > h.Config.CloseTTL {
```
urso: config validation mandates CloseTTL can not be 0 (due to `min=0,nonzero`). This on purpose?

ruflin: yes, because I don't see any use cases for close_ttl = 0, as this would the harvester directly again.

filebeat/harvester/log.go
```diff
@@ -87,6 +88,11 @@ func (h *Harvester) Harvest() {
 		if !h.sendEvent(event) {
 			return
 		}
+
+		if h.config.CloseTimeout > 0 && time.Since(startTime) > h.config.CloseTimeout {
+			logp.Info("Closing harvester because ttl was reached: %s", h.path)
+			return
+		}
```
urso: Checking for CloseTimeout at this places requires an event to be published for closeTimeout to function.

If reader is waiting long for a new line or sendEvent blocks, due to spooler being blocked on publisher, close_timeout might be very very late.

urso: file reader has some 'inactive timeout'? The 'inactive timeout' should be <= `close_timeout`.

What about multiline timeout? hm...

ruflin: Alternative implementation could be to implement the timeout in a separate go routine the close `h.done`?

ruflin:
@urso I pushed an alternative implementation. Can you have a look?

ruflin:
jenkins, retest it

filebeat/harvester/config.go
```diff
@@ -45,6 +46,7 @@ type harvesterConfig struct {
 	CloseRemoved         bool                       `config:"close_removed"`
 	CloseRenamed         bool                       `config:"close_renamed"`
 	CloseEOF             bool                       `config:"close_eof"`
+	CloseTimeout         time.Duration              `config:"close_timeout" validate:"min=0,nonzero"`
```
urso: `validate` enforces `CloseTimeout>=1ns`. This enforces CloseTimeout always enabled. Why disallow 0 or -1?

filebeat/harvester/log.go
```diff
@@ -44,6 +44,29 @@ func (h *Harvester) Harvest() {
...
+		select {
+		case <-h.prospectorDone:
+			close(h.done)
+		case <-h.done:
+			return
+		}
+
+	}()
+
+	if h.config.CloseTimeout > 0 {
```
urso: always true, due to config validation

filebeat/harvester/log.go
```diff
@@ -44,6 +44,29 @@ func (h *Harvester) Harvest() {
...
+		go func() {
+			select {
+			case <-h.done:
+				return
+			case <-time.After(h.config.CloseTimeout):
+				close(h.done)
```
urso: If I remember correctly, h.done is passed to line reader. So this should break the reader loop. But I'm not sure about impact on multiline. Will a multiline event being generated if h.done is closed?

filebeat/harvester/log.go
```diff
@@ -44,6 +44,29 @@ func (h *Harvester) Harvest() {
...
 
+	go func() {
+		select {
+		case <-h.prospectorDone:
+			close(h.done)
+		case <-h.done:
+			return
+		}
+
+	}()
```
urso: why two done channels and 2 workers dealing with done-channels. Things seem to get complicated in harvester loop again.

filebeat/harvester/config.go
ruflin: I agree that nonzero should be removed. But -1 doesn't make sense from my point of view?

filebeat/harvester/log.go
ruflin: The reason here is that this go loop can close the harvester based on a channel. The prospectorDone channel can't be used for that as the same is used for all harvester, means all harvesters are stopped, not just this one.

filebeat/harvester/log.go
ruflin: Also if the default is set to 0?

filebeat/harvester/log.go
ruflin: It is the same channel we use to stop all harvesters. So stopping should also break the multiline part. Would you expect it to properly finish a mulitline fetch? I think it is only important that it stores the correct offset.

urso: on error, multiline might create a multiline event with current content. I'm worried about storing an offset right in the middle of a multiline-event, potentially splitting one multiline event into 2 events.

filebeat/harvester/log.go
urso: can we document this somewhere very well? It's not fully clear + construct of cascading close-workers looks overly complicated (not to say ugly).

ruflin: Will document it for now directly in the code. Agree would be nice if we find later a nicer option.

filebeat/harvester/log.go
ruflin: Suggestion: I will document this potential issue in the docs for the moment. TBH not 100% sure if that is what would happen. I think it is more that the processor will block the timeout as even though the channel is closed, I wont have an affect on the processor currently. It would be nice to enhance the processors in a way that it is possible to stop a processor.

ruflin:
@urso PR updated with adding a note to `docs` about potential multiline issue and added some notes about the channels to the code. We definitively should revise how processors are started / stopped at a later stage.

filebeat/harvester/config.go
urso: yeah, -1 is kinda similar to 0. `CloseTimeout` should only be enabled if >0

ruflin: I think that is currently the case?

filebeat/harvester/log.go
urso: config requires CloseTimeout to be `min=0,nonzero`. This makes `0` an invalid config file option. With default being 0, it can still be 0. Reminds me, ucfg should validate defaults, if value is not configured.

ruflin: As far as I understand you, the current implementation works as I intended, but you would prefer to disable this in ucfg. The question then is what should be use in this case here? or the nonzero requirement should be removed?

filebeat/harvester/log.go
urso: logFileReader is passed a `done` channel (h.done by default). The done channel when closed makes the [reader return `0, nil`](https://github.com/elastic/beats/blob/master/filebeat/harvester/reader/log.go#L67). The line reader (after [reading 0 bytes without error](https://github.com/elastic/beats/blob/master/filebeat/harvester/reader/line.go#L120)) will block in [for loop waiting for newline](https://github.com/elastic/beats/blob/master/filebeat/harvester/reader/line.go#L65). I'm afraid, we're currently waiting on multiline timeout in order to check done channel.

Closing the file handle will ensure [an error is actually returned](https://github.com/elastic/beats/blob/master/filebeat/harvester/reader/log.go#L85). The error will be [propagated by the line reader](https://github.com/elastic/beats/blob/master/filebeat/harvester/reader/line.go#L120) and the optional json processor. The multiline processor will [publish ongoing buffered multiline event](https://github.com/elastic/beats/blob/master/filebeat/harvester/processor/multiline.go#L167) and return an error on next read attempt (no more attempt by harvester due to h.done being closed). With readLine still returning a valid event, the [event might still be published by harvester](https://github.com/elastic/beats/blob/master/filebeat/harvester/log.go#L74), due to [select choosing first channel by random](https://github.com/elastic/beats/blob/master/filebeat/harvester/log.go#L115). That is, even by closing the file handle, there is a chance of having a multiline event being split up between harvester restarts.

filebeat/harvester/config.go
urso: It's not. using `nonzero` in `validate` requires `close_timeout > 0` in config files.

filebeat/harvester/log.go
urso: nonzero requirement should be removed. It's quite contradictory not allowing this to be 0 in config file, but setting default to 0.

ruflin:
In discussion with @urso we realised there is a potential issue in reader / processor which could lead to and endless loop potentially. Before merging this PR, the reader / processor part should be cleaned up to guarantee proper stopping.

ruflin:
I redid the implementation and add a Close function to the log_file reader. This implementation has the following issues:

* In case the output applies back pressure and not at least one more event can be sent, the harvester will not be closed, means the file handler will stay open in case the output service is down
* An alternative implementation was done where the harvester was directly killed not waiting for the output to complete. This had the side affect, that the final state of the harvester (Finished) was not updated properly. The issue here is currently that state updates and data updates go through the same channel. Even though state updates are not sent by the spooler, they still to through the spooler and if the queue is full not further states can be sent either. One potential option here could be to decouple the state updates and event updates. But this will bring lots of new issues with race conditions between state and event updates. In addition, state updates for events should only be applied, if the event was completely sent.

The above implementation is quite simple and works in cases where the output is behaving normally. This is in line with the oder `close_*` options. They all only apply in case the output works as expected. Otherwise file handlers stay open.

filebeat/harvester/log_file.go
```diff
@@ -164,3 +165,10 @@ func (r *LogFile) wait() {
 		}
 	}
 }
+
+func (r *LogFile) Close() {
+	// Make sure reader is only closed once
+	r.singleClose.Do(func() {
+		close(r.done)
```
urso: as the underlying reader might block on syscall, the file must be closed right after closing the channel.

filebeat/harvester/log.go
```diff
@@ -294,11 +295,34 @@ func (h *Harvester) newLogFileReader() (reader.Reader, error) {
...
+			case <-h.done:
+				return
+			case <-time.After(h.config.CloseTimeout):
+				logp.Info("Closing harvester because close_timeout was reached: %s", h.state.Source)
+				fileReader.Close()
+				return
+			}
+		}()
+	}
+
```
urso: alternative implementation:

```
var closeTimeout chan time.Time
    closeTimeout = time.After(h.config.CloseTimeout):
}
go func () {
    select {
        case <- h.done:
        case <- closeTimeout:
            logp.Info("Closing harvester because close_timeout was reached: %s", h.state.Source)
    }
    fileReader.Close()
}
```

urso: Feels kind of like a hack. Can we wrap `*LogFile` in case `close_timeout` is configured?

urso:
which error code is passed up the readers if file is closed by timeout (or done channel)?

what happens to buffered (but incomplete) multiline events if file is closed early?

filebeat/harvester/log_file.go
ruflin: done

filebeat/harvester/log.go
ruflin: I removed the returns and moved the filereader.Close() after the select. Not sure what the advantage is of moving the closeTimeout logic outside. Like this always a go routine is started which is not necessary.

What do you mean by wrap?

ruflin:
If the channel is closed, the Error will be `ErrClosed`. What the error will be after closing the file handler probably depends in which state the file handler was at this moment.

Incomplete multiline events will be sent on `ErrClosed` (see tests). It is handled the same as if the timeout was reached in multiline.

filebeat/harvester/log.go
ruflin: @urso ignore my first comment above. I see you wrapped both go routines into 1. WIll change it accordingly.

urso: wrap like, have another type implementing `io.ReadCloser` doing the timeout on underlying `io.ReadClose`.

filebeat/harvester/log.go
```diff
@@ -257,9 +258,16 @@ func (h *Harvester) close() {
...
-	if h.file != nil {
+	if h.fileReader != nil || h.file != nil {
+
+		// In case file exists but file reader was never created
+		// Using h.fileReader.Close ensures file handler is only closed once
+		if h.fileReader != nil {
+			h.fileReader.Close()
+		} else {
+			h.file.Close()
+		}
```
urso: who is calling `(*Harvester).close()`? like which go-routine? There a chance of race on h.fileReader being set?

filebeat/harvester/log.go
```diff
@@ -294,12 +302,27 @@ func (h *Harvester) newLogFileReader() (reader.Reader, error) {
...
+	h.fileReader, err = NewLogFile(h.file, h.config)
 	if err != nil {
 		return nil, err
 	}
 
-	r, err = reader.NewEncode(fileReader, h.encoding, h.config.BufferSize)
+	var closeTimeout <-chan time.Time
+	if h.config.CloseTimeout > 0 {
+		closeTimeout = time.After(h.config.CloseTimeout)
+	}
```
urso: this block we can even move into the go-routine being started.

filebeat/harvester/log.go
ruflin: The same go routine that calls close is also setting the fileReader, so there should not be any race condition.

urso: hm... why not call `defer fileReader.Close()` right after creating in said go-routine? `fileReader.Close()` already ensures it's doing the magic only once + is thread-safe.

filebeat/docs/reference/configuration/filebeat-options.asciidoc
```diff
@@ -234,6 +234,17 @@ WARNING: Only use this options if you understand that data loss is a potential s
...
 
+===== close_timeout
+
+WARNING: Only use this options if you understand the potential side affects with potential data loss. In addition it can happen that multiline events are not sent completely on timeout.
+
+Close timeout gives every harvester a predefined lifetime. Independent of the location of the reader, it will stop the reader after `close_timeout`. This option can be useful, if only a predefine time should be spent on older log files. Using this option in combination with `ignore_older` == `close_timeout` means the file is not picked up again in case it wasn't modified in between. This normally leads to data loss and not the complete file is sent.
+
+In case close_timeout is used in combination with multiline events, it can happen that the harvester will be stopped in the middle of a multiline event, means only parts of the event will be sent. In case the harvester is continued at a later stage again and the file still exists, only the second part of the event will be sent.
+
+Close timeout will not apply, in case your output is stuck and no further events can be sent. At least one event further event must be sent to make close_timeout apply.
```
urso: > At least one event further event must be sent to make close_timeout apply.

Can you clarify?

ruflin: fixed

urso:
@ruflin needs rebase

ruflin:
@urso Rebase done

ruflin:
jenkins, retest it

PR #1128, opened 2016-03-08
Change Cpu.Get() on Windows to not use floating point artimetic

andrewkroh:
Incorporates elastic/gosigar#19 to hopefully fix an issue with incorrect `cpu.system_p` values on Windows. See [discuss](https://discuss.elastic.co/t/topbeat-sometimes-returns-an-invalid-value-for-cpu-system-p-on-windows/43706/4)

Best case: it fixes the issue
Worst case: instead of incorrect values `cpu.system_p` values, it reports a negative percentage due to the system time value being reported as negative by Windows.

PR #1928, opened 2016-06-29
Fix high values of proc.cpu.user_p

monicasarbu:
Apply the same fix from #1128 to proc.cpu.user_p. This is fixed already in 5.0.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -8,7 +8,7 @@
 // Template, add newest changes here
 
 === Beats version HEAD
-https://github.com/elastic/beats/compare/v1.2.3...1.3[Check the HEAD diff]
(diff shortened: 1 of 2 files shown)


tsg:
Meging, as appveyor fail in Winlogbeat cannot be related.

CHANGELOG.asciidoc
```diff
@@ -87,6 +87,7 @@ https://github.com/elastic/beats/compare/v1.2.2...v1.2.3[View commits]
 *Topbeat*
 
 - Fix high CPU usage when using filtering under Windows. {pull}1598[1598]
+- Fix high values of proc.cpu.user_p. {pull}1928[1928]
```
andrewkroh: @monicasarbu I think this changelog entry was placed/merged to the wrong location, right? Shouldn't it be under the 1.3.0 section?

PR #1933, opened 2016-06-29
Remove scan_frequency config docs for 0s

ruflin:


filebeat/etc/beat.full.yml
```diff
@@ -83,9 +83,9 @@ filebeat.prospectors:
   # in. Default: log
   #document_type: log
 
-  # Scan frequency in seconds.
-  # How often these files should be checked for changes. In case it is set
-  # to 0s, it is done as often as possible. Default: 10s
+  # How often the prospector checks for new files in the paths that are specified
+  # for harvesting. Specify 1s to scan the directory as frequently as possible
```
tsg: Does that mean that `1s` is the minimum we are confortable with? I thought we use sub-seconds in our tests. Perhaps we could word it as "Values under one second are not recommended" or even enforce the 1s with ucfg.

ruflin: I would stick with "Values under one second are not recommended". Lower values are very useful for us to test some edge cases and race conditions.

PR #1934, opened 2016-06-29
Temporary hack for the Kibana index pattern

tsg:
This is an attempt at a workaround for elastic/kibana#7563. The PR
adds code to our load scripts to set the default index to be
`<name-of-the-beat>-*` after loading the index patterns. This makes it work well out of the box but
comes with two disadvantages:

1. it's hardcoded for alpha4
2. we used to not touch the default index. It might be unexpected for
 existing users that we now change it.


tsg:
@elastic/beats what do you think? It's code and a change in behavior that we'll revert with beta1, but I kind of prefer it against things not working OOB. If we go for it, this would obviously also require a re-spin.

andrewkroh:
Will this make the script **only** work with Kibana 5.0.0-alpha4? I imagine people will try to use it with last major version of Kibana.

monicasarbu:
I propose to write about the workaround in the blog post and close this PR.

monicasarbu:
A fix will be provided in beta1 release.

tsg:
Agreed, closing this one as a respin would add to many untested changes in the release at a too late point in time.

PR #1938, opened 2016-06-30
Add system uptime MetricSet

andrewkroh:
This MetricSet reports the OS uptime in milliseconds. It is currently implemented for Linux, Darwin, and OpenBSD. It's not implemented on Windows yet.

Commentary
- I chose milliseconds for the units (abbreviated as ms).
- `uptime` is in its own metricset. Should it be included in some other metricset?

### DIFF
--- a/metricbeat/etc/fields.yml
+++ b/metricbeat/etc/fields.yml
@@ -1634,6 +1634,15 @@
                   type: long
                   description: >
                     The shared memory the process uses.
+        - name: uptime
+          type: group
+          description: >
+            `uptime` contains the operating system uptime metric.
+          fields:
+            - name: uptime_ms
+              type: long
+              description: >
+                The OS uptime in milliseconds.
 - key: zookeeper
   title: "ZooKeeper"
   description: >
--- a/metricbeat/module/system/uptime/_meta/data.json
+++ b/metricbeat/module/system/uptime/_meta/data.json
@@ -1,0 +1,18 @@
+{
+    "@timestamp": "2016-06-29T19:25:08.155Z",
+    "beat": {
+        "hostname": "host.example.com",
+        "name": "host.example.com"
+    },
+    "metricset": {
+        "module": "system",
+        "name": "uptime",
+        "rtt": 33
+    },
+    "system": {
+        "uptime": {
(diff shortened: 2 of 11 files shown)


tsg:
A metricset with only one value feels weird to me, especially since there's quite a bit of overhead added to that doc. But I see we don't have an overall "system" metricset, so I'm not sure what would be the right place. Perhaps under "cpu"?

metricbeat/module/system/uptime/_meta/data.json
```diff
@@ -0,0 +1,18 @@
...
+        "name": "host.example.com"
+    },
+    "metricset": {
+        "module": "system",
+        "name": "uptime",
+        "rtt": 33
+    },
+    "system": {
+        "uptime": {
+            "uptime_ms": 789306155
```
tsg: We generally do `uptime.ms` in this case (not with actual dot, but as a sub-doc).

andrewkroh:
It does seem strange to have only a single number. @simianhacker also suggested putting it under cpu so I'll move it there and change the name to `uptime.ms`.

metricbeat/etc/fields.yml
```diff
@@ -1634,6 +1634,15 @@
                   type: long
                   description: >
                     The shared memory the process uses.
+        - name: uptime
+          type: group
+          description: >
+            `uptime` contains the operating system uptime metric.
+          fields:
+            - name: uptime_ms
```
ruflin: This should be ms only? removing the duplicated `uptime` part.

ruflin:
An alternative here could be to introduce a `system.info` metricset that will also contain other things in the future like `time`, `timezone` (checking if all servers are configurated correctly), `os` and other infos. For me it doesn't really fit under `cpu`.

metricbeat/etc/fields.yml
ruflin: Removing comment as I saw it was already commented.

andrewkroh:
I'm closing this one and will open up a PR (or someone else can) for a system/info metricset containing uptime. See https://github.com/andrewkroh/beats/blob/feature/system-modules/metricbeat/module/system/host/host.go#L31 for an example.

PR #1941, opened 2016-06-30
Use half floats for most floating point numbers

### LINKED ISSUE #1936: Switch percentages and other small floats to use half floats
Elasticsearch now supports [half floats](https://www.elastic.co/guide/en/elasticsearch/reference/master/number.html). ❗ 🎉  🎈  ❤️ 

We can put them to use at least for our percentage numbers, probably useful in a few other cases as well.

tsg:
Starting with 5.0.0-alpha4, Elasticsearch supports [half floats](https://github.com/elastic/elasticsearch/pull/18887).

[Half precision floating-point](https://en.wikipedia.org/wiki/Half-precision_floating-point_format) numbers
have good precision for small numbers, but the precision degrades fast for larger numbers.

This is switching to half floats all the fields that are percentages
(values between 0 and 1) and the fields that were floats but have naturally
small values only. After going through the list, most of our floating point
numbers fit in one of these categories. Only 3 float fields were left untouched.

The template generator automatically switches to "float" when generating the
ES 2.x template files.

This closes #1936.

### DIFF
--- a/libbeat/scripts/generate_template.py
+++ b/libbeat/scripts/generate_template.py
@@ -178,10 +178,14 @@ def fill_field_properties(args, field, defaults, path):
             }
 
     elif field["type"] in ["geo_point", "date", "long", "integer",
-                           "double", "float", "boolean"]:
+                           "double", "float", "half_float", "boolean"]:
         # Convert all integer fields to long
         if field["type"] == "integer":
             field["type"] = "long"
+
+        if args.es2x and field["type"] == "half_float":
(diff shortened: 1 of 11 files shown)


tsg:
jenkins, retest it

ruflin:
Thanks a lot for taking this one.

jpountz:
I'm curious whether this significatly reduces the size of indices?

jpountz:
See also elastic/elasticsearch#9264.

PR #1944, opened 2016-06-30
Rename filters with processors

monicasarbu:
As we are planning to use filters not only to reduce the number of fields from the exported event, but also to enhance the event with external meta data, we decided to rename `filters` with `processors`.

NOTE: In Metricbeat you can also define `filters` per module and this part is not adjusted by this PR.


tsg:
jenkins, retest it

libbeat/beat/beat.go
```diff
@@ -107,16 +107,16 @@ type Beat struct {
...
 type BeatConfig struct {
-	Shipper publisher.ShipperConfig   `config:",inline"`
-	Output  map[string]*common.Config `config:"output"`
-	Logging logp.Logging              `config:"logging"`
-	Filters filter.FilterPluginConfig `config:"filters"`
-	Path    paths.Path                `config:"path"`
+	Shipper    publisher.ShipperConfig          `config:",inline"`
+	Output     map[string]*common.Config        `config:"output"`
+	Logging    logp.Logging                     `config:"logging"`
+	Processors processors.ProcessorPluginConfig `config:"processors"`
```
urso: let's rename `ProcessorPluginConfig` to `PluginConfig` to reduce stuttering.

libbeat/processors/config.go
```diff
@@ -9,22 +9,22 @@ import (
...
+	Range    *ConditionFields `config:"range"`
 }
 
-type ConditionFilter struct {
+type ConditionFields struct {
 	fields map[string]interface{}
 }
 
-type FilterPluginConfig []map[string]common.Config
+type ProcessorPluginConfig []map[string]common.Config
```
urso: let's rename to 'PluginConfig' as the package name is always part of the type name in golang.

libbeat/processors/processor.go
```diff
@@ -0,0 +1,88 @@
...
+
+	logp.Debug("processors", "Processors: %v", processors)
+	return &processors, nil
+}
+
+func (processors *Processors) addProcessor(p ProcessorInterface) {
+
+	if processors.list == nil {
+		processors.list = []ProcessorInterface{}
+	}
```
urso: this if-block is not required. appending to `nil` ensures array is allocated first time.

libbeat/processors/registry.go
```diff
@@ -0,0 +1,28 @@
+package processors
+
+import (
+	"fmt"
+
+	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/logp"
+)
+
+type ProcessorInterface interface {
```
urso: let's rename to `Processor`.

libbeat/processors/registry.go
```diff
@@ -0,0 +1,28 @@
...
+	Process(event common.MapStr) (common.MapStr, error)
+	String() string
+}
+
+type Constructor func(config common.Config) (ProcessorInterface, error)
```
urso: Normally we pass config by pointer.

Today I learned passing config by pointer and by value is exactly equivalent. All modifications to config will still be visible, due to state is held by references to other objects. In this regard common.Config behaves like map[string]interface{}. It's funny how I wasn't aware of it :)

libbeat/processors/registry.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"fmt"
+
+	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/logp"
+)
+
+type ProcessorInterface interface {
+	Process(event common.MapStr) (common.MapStr, error)
```
urso: I'm tempted to rename `Process` to `Exec` or `Run`, which sounds more generic to me. Plus we can write `Processor.Exec` instead of `Processor.Process`.

urso:
LGTM

libbeat/beat/beat.go
monicasarbu: The reason why I used `ProcessorPluginConfig` is that it's also used in Metricbeat as you are able to configure `filters` in a Metricbeat module.

libbeat/processors/registry.go
monicasarbu: Good idea. I would use `Run`.

ruflin:
Similar to metricbeat, I was planning to implement "filters / prospectors" on a prospector level to filter events early and simplify the configuration of different filters for each prospector.

Issue #1929, opened 2016-06-29
Full configuration files are not updated per OS


tsg:
We modify slightly the configuration files at package time depending on the OS. For example, on OS X we replace `device: any` with `device: en0`. This works for the short configs but not for the long ones.

ruflin:
Closing as https://github.com/elastic/beats/pull/1945 was merged.

PR #1945, opened 2016-06-30
Use OS specific network interfaces in full configs

tsg:
The only platform specific adjustments that we make
are for the packetbeat network device. This applies the change
also to the full config.

### DIFF
--- a/packetbeat/Makefile
+++ b/packetbeat/Makefile
@@ -14,9 +14,13 @@ with_pfring:
 before-build:
 	sed -i.bk 's/device: any/device: en0/' $(PREFIX)/packetbeat-darwin.yml
 	rm $(PREFIX)/packetbeat-darwin.yml.bk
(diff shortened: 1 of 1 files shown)


tsg:
jenkins, package it

tsg:
This fixes #1929.

packetbeat/Makefile
```diff
@@ -14,9 +14,13 @@ with_pfring:
 before-build:
 	sed -i.bk 's/device: any/device: en0/' $(PREFIX)/packetbeat-darwin.yml
 	rm $(PREFIX)/packetbeat-darwin.yml.bk
+	sed -i.bk 's/device: any/device: en0/' $(PREFIX)/packetbeat-darwin.full.yml
+	rm $(PREFIX)/packetbeat-darwin.full.yml.bk
```
ruflin: Should we do the same for filebeat and the windows version for the paths? Currently we have:

```
  paths:
    - /var/log/*.log
    #- c:\programdata\elasticsearch\logs\*
```

Issue #1931, opened 2016-06-29
Reduce Warn verbosity, switch to Info by default


tsg:
The current state is that we log at Error level by default. The historical reason is that Packetbeat could generate a lot of valid warnings (e.g. when traffic contains invalid patterns) which result in slowing it down for the rest of traffic, or filling up the disk, or rotating the logs too fast, etc.

But printing only errors by default is also not great, because:

* the user gets no feedback that the Beat correctly started
* in some cases, critical information is hidden. See #1734 for an example
* very low logging defaults means that for almost any troubleshooting we need to ask the user to increase verbosity

In general, we want to avoid warnings all together, as they are [not friendly to the ops people](http://blog.schlomo.schapiro.org/2015/04/warning-is-waste-of-my-time.html). For us, warnings (and to some extent INFO messages) are essentially debug messages that we want to have always on. The plan is to replace Warnings with expvar metrics, which can be inspected in realtime via the `-httpprof` option and for which we can also write periodic summaries to the logs.

This is a meta ticket to:

* [x] Review and reduce over-verbose warnings and infos in each beat:
  * [x] Packetbeat #1948 
  * [x] Filebeat #1970
  * [x] Metricbeat
  * [x] Winlogbeat
  * [x] libbeat #1973 
* [x] Change the default logging level from Error to Info. #1952 #2364 
* [x] Periodically log expvar counters #1955 #1969 
* [x] Rename our expvars from camelCase to dotted.case #1967
* [x] Review winlogbeat expvars, if they can use the naming convention above
* [x] Log total expvar metrics at the end in all beats, like Metricbeat currently does it. #2349

tsg:
Compared to the other Beats, Filebeat is a bit more verbose at INFO level. For example, it prints a message on each bulk publish:

```
2016/06/29 10:50:49.536775 registrar.go:239: INFO Registry file updated. 12 states written.
2016/06/29 10:50:50.406032 publish.go:111: INFO Events sent: 2048
2016/06/29 10:50:50.407090 registrar.go:239: INFO Registry file updated. 12 states written.
2016/06/29 10:50:51.227240 publish.go:111: INFO Events sent: 2048
2016/06/29 10:50:51.228463 registrar.go:239: INFO Registry file updated. 12 states written.
2016/06/29 10:50:52.017300 publish.go:111: INFO Events sent: 2048
2016/06/29 10:50:52.018828 registrar.go:239: INFO Registry file updated. 12 states written.
[...]

PR #1948, opened 2016-07-01
Replace most Packetbeat warnings with expvar metrics

tsg:
Demoted warnings like "Two requests without a response" or
"Response without a request" to the Debug level. To compensate,
expvar counters are added for these.

These types of warnings are normal in some situations (e.g. at startup)
but can indicate serious issues in the other cases.

This is part of #1931, which has the overall goal for switching the
default logging level to INFO.

### DIFF
--- a/packetbeat/protos/dns/dns.go
+++ b/packetbeat/protos/dns/dns.go
@@ -9,6 +9,7 @@ package dns
 
 import (
 	"bytes"
+	"expvar"
 	"fmt"
 	"net"
 	"sort"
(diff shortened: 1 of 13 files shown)


packetbeat/protos/dns/dns.go
```diff
@@ -41,6 +42,11 @@ const (
 // Transport protocol.
 type Transport uint8
 
+var (
+	unmatchedRequests  = expvar.NewInt("dnsUnmatchedRequests")
+	unmatchedResponses = expvar.NewInt("dnsUnmatchedResponses")
```
ruflin: general idea: should we use the same naming rules for expars as we do metricbeat?

packetbeat/protos/thrift/thrift.go
```diff
@@ -157,6 +158,11 @@ type Thrift struct {
 	Idl          *ThriftIdl
 }
 
+var (
+	unmatchedRequests  = expvar.NewInt("thriftUnmatchedRequests")
```
ruflin: Perhaps we can generalise this in the future so the methods return an error and the "stats" gathering happen one level up.

PR #1950, opened 2016-07-01
Add config examples with deprecation text to support alpha4 and beta1

dedemorton:
This PR adds annotations to indicate configs that are required for alpha4 vs beta1. Also includes a few quick edits (I need to do a more thorough edit later).


monicasarbu:
Thank you @dedemorton

PR #1954, opened 2016-07-04
Fix Filebeat rotation issue

ruflin:
In case a file was renamed after the state was read but before the file was opened by the harvester it could happen that the wrong file was opened. This lead to the issue, that the wrong file was read including reporting the wrong state for a file. In general this was rather unlikely to happen, but can happen when scan_frequency is very low and number of files rotated is very high.

This issue is fixed with the harvester now compares again the file after opening with the state given from the prospector. In case the state is not identical with the file, the harvester aborts. This means that the file will only be picked up again after scan_frequency but potential errors are prevented.

Closes https://github.com/elastic/beats/issues/1956

### DIFF
--- a/filebeat/harvester/log.go
+++ b/filebeat/harvester/log.go
@@ -170,25 +170,34 @@ func (h *Harvester) openFile() (encoding.Encoding, error) {
 	var encoding encoding.Encoding
 
 	f, err := file.ReadOpen(h.Path)
-	if err == nil {
-		// Check we are not following a rabbit hole (symlinks, etc.)
-		if !file.IsRegular(f) {
-			return nil, errors.New("Given file is not a regular file.")
-		}
-
-		encoding, err = h.encoding(f)
-		if err != nil {
-
-			if err == transform.ErrShortSrc {
-				logp.Info("Initialising encoding for '%v' failed due to file being too short", f)
-			} else {
-				logp.Err("Initialising encoding for '%v' failed: %v", f, err)
-			}
-			return nil, err
-		}
-
-	} else {
+	if err != nil {
 		logp.Err("Failed opening %s: %s", h.Path, err)
+		return nil, err
+	}
+
+	// Check we are not following a rabbit hole (symlinks, etc.)
+	if !file.IsRegular(f) {
+		return nil, errors.New("Given file is not a regular file.")
+	}
+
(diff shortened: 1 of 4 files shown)


filebeat/harvester/log.go
```diff
@@ -170,25 +170,34 @@ func (h *Harvester) openFile() (encoding.Encoding, error) {
...
 
-			if err == transform.ErrShortSrc {
-				logp.Info("Initialising encoding for '%v' failed due to file being too short", f)
-			} else {
-				logp.Err("Initialising encoding for '%v' failed: %v", f, err)
-			}
-			return nil, err
-		}
+	// Check we are not following a rabbit hole (symlinks, etc.)
+	if !file.IsRegular(f) {
```
tsg: Wouldn't make sense for the new checks to be after the IsRegular check? Not sure if a symlink can cause issues with `Stat` or `Samefile`.

filebeat/tests/system/test_load.py
```diff
@@ -0,0 +1,108 @@
...
+        assert len(entry_list) == total_lines
+
+
+    def test_no_open_files_left(self):
+        """
+        Test that filebeat not keep any files open
+        """
+
+        # This is not implemented yet, here as a reminder
+        raise SkipTest
```
tsg: Is this for another PR?

filebeat/tests/system/test_load.py
```diff
@@ -0,0 +1,108 @@
...
+
+"""
+Test filebeat under different load scenarios
+"""
+
+
+class Test(BaseTest):
+    def test_no_missing_events(self):
+        """
+        Test that filebeat does not loose any events under heavy file rotation and load
```
tsg: These tests are great! Really these bug fixes wouldn't be possible without.

tsg:
LGTM, left some questions.

filebeat/harvester/log.go
ruflin: I move the check up just to be sure.

filebeat/tests/system/test_load.py
ruflin: yes, still trying to figure out how I can "track" open file handlers in python.

ruflin:
@tsg Updated version pushed.

PR #1955, opened 2016-07-04
Log non-zero expvars periodically

tsg:
Prints the delta since the last period, if it's non-zero. The idea is
to help with troubleshooting while still creating relatively little output
in the logs. 30s seems like a good compromise.

Part of/related to #1931

### DIFF
--- a/libbeat/logp/logp.go
+++ b/libbeat/logp/logp.go
@@ -1,11 +1,14 @@
 package logp
 
 import (
+	"expvar"
 	"flag"
 	"fmt"
 	"io/ioutil"
 	"log"
+	"strconv"
 	"strings"
+	"time"
 
 	"github.com/elastic/beats/libbeat/paths"
 )
@@ -21,7 +24,17 @@ type Logging struct {
 	ToSyslog  *bool `config:"to_syslog"`
 	ToFiles   *bool `config:"to_files"`
 	Level     string
+	Metrics   LoggingMetricsConfig `config:"metrics"`
 }
+
+type LoggingMetricsConfig struct {
+	Enabled *bool          `config:"enabled"`
+	Period  *time.Duration `config:"period" validate:"nonzero,min=0s"`
+}
+
+var (
+	defaultMetricsPeriod = 30 * time.Second
+)
 
 func init() {
 	// Adds logging specific flags: -v, -e and -d.
@@ -112,6 +125,8 @@ func Init(name string, config *Logging) error {
 		log.SetOutput(ioutil.Discard)
 	}
 
+	go logExpvars(&config.Metrics)
+
 	return nil
 }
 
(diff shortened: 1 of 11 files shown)


tsg:
Not ready for merging, but discussion/reviews welcome.

tsg:
Example log line:

```
2016/07/04 12:16:10.821047 logp.go:194: INFO Non-zero metrics in the last 30s:  libbeatEsPublishedAndAckedEvents=18 libbeatEsPublishWriteBytes=10045 libbeatEsPublishReadBytes=1032 libbeatMessagesInWorkerQueues=3 libbeatPublishedEvents=18 libbeatEsPublishEventsCallCount=3
```

libbeat/logp/logp.go
```diff
@@ -21,8 +24,18 @@ type Logging struct {
 	ToSyslog  *bool `config:"to_syslog"`
 	ToFiles   *bool `config:"to_files"`
 	Level     string
+	Metrics   LoggingMetricsConfig `config:"metrics"`
```
ruflin: +1. Should we enable this by default?

libbeat/logp/logp.go
```diff
@@ -141,3 +156,45 @@ func getLogLevel(config *Logging) (Priority, error) {
...
+	}
+	if metricsCfg.Period == nil {
+		metricsCfg.Period = &defaultMetricsPeriod
+	}
+	Info("Metrics logging every %s", metricsCfg.Period)
+
+	ticker := time.NewTicker(*metricsCfg.Period)
+	prevVals := map[string]int64{}
+	for {
+		<-ticker.C
```
ruflin: Even though it is running in a independent go routine, should we still make sure it is shut down properly when the beat is stopped?

ruflin:
LGTM

libbeat/logp/logp.go
tsg: It is enabled by default with a 30s interval.

ruflin: Now I see it on line 170. Would probably make sense to also use a ucfg defaultConfig.

ruflin: @tsg For the above: I didn't mean as part of this PR.

tsg: I tried initially but didn't work out due to circular imports (cfgfile imports logp).

ruflin: :-( @urso interesting issue.

libbeat/logp/logp.go
tsg: I thought about that, but I struggle with the API for it. We'd need to have a `Stop()` function for the logging system, right?

ruflin: A simple version could be just passing the done channel to the go routine and using a select statement. But that would require to move the start of the goroutine into the beat starting process.

tsg: Ah, ok, but then it's not self contained in logp. Maybe it doesn't need to be.

tsg:
Added docs and a basic system test. @ruflin, can you have a look before I squash?

ruflin:
LGTM

libbeat/logp/logp.go
```diff
@@ -141,3 +156,57 @@ func getLogLevel(config *Logging) (Priority, error) {
...
 	return level, nil
 }
+
+// snapshotExpvars iterates through all the defined expvars, and for the top
+// level vars that are integers it snapshots the name and value in a separate
+// map.
+func snapshotExpvars(varsMap map[string]int64) {
+	expvar.Do(func(kv expvar.KeyValue) {
+		switch kv.Value.(type) {
+		case *expvar.Int:
```
andrewkroh: I think this PR covers the most important metrics for the time being so no action is required at the moment IMO. We can improve this when we add "metrics" within beats. We have some expvar metrics that are nested within a map that won't be printed by this. For example, in Metricbeat we have a `fetches.mysql-info.failures` metric that is nested two levels deep in maps.

Issue #1958, opened 2016-07-05
Support Cassandra protocol in Packetbeat


medcl:
As Cassandra is popular used as a multi-column storage, i suggest we may add Cassandra protocol in Packetbeat.

Cassandra clients use CQL to Interactive with the Cassandra server, current CQL protocol is v4, 
the protocol is well documented here:
https://github.com/apache/cassandra/blob/trunk/doc/native_protocol_v4.spec

medcl:
Closing now

PR #1959, opened 2016-07-05
Support protocol: Cassandra

medcl:
This PR is for this issue:
https://github.com/elastic/beats/issues/1958


andrewkroh:
jenkins, test it again, please

glide.yaml
```diff
@@ -83,3 +83,8 @@ import:
   - zlib
 - package: github.com/klauspost/cpuid
   version: v1.0
+- package: https://github.com/go-inf/inf
+  version: v0.9.0
+- package: github.com/golang/snappy
+  version: d9eb7a3d35ec988b8585d4a0068e462c27d28380
```
urso: snappy is already included at line 53.

packetbeat/etc/beat.full.yml
```diff
@@ -382,3 +382,22 @@ packetbeat.protocols.nfs:
...
+  send_response: true
+  # If this option is enabled, the raw message of the response (`cassandra_request_header` field)
+  # is sent to Elasticsearch. The default is false.
+  send_request_header: true
+  # If this option is enabled, the raw message of the response (`cassandra_response_header` field)
+  # is sent to Elasticsearch. The default is false.
+  send_response_header: true
+  # If this option is enabled, and also the flag indicates that the frame body was compressed, and
+  # the stream will decode by the compressor, currently support: `snappy` , The default is `nil`.
+  compressor: "snappy"
```
urso: - we normally set default values in config file and comment these out.
- we normally use a newline after each config.
- when saying `is sent to Elasticsearch`, better use `is included in published events`. Packetbeat is not using elasticsearch exclusively

urso:
run `make clean-vendor` to remove some vendored files introduced. e.g. `make clean-vendor` will remove unit tests from vendored packages.

packetbeat/protos/cassandra/cassandra.go
```diff
@@ -0,0 +1,200 @@
...
+	return p, nil
+}
+
+func (cassandra *cassandra) init(results publish.Transactions, config *cassandraConfig) error {
+	if err := cassandra.setFromConfig(config); err != nil {
+		return err
+	}
+	cassandra.pub.results = results
+
+	isDebug = logp.IsDebug("http")
```
urso: replace 'http' with cassandra!

urso: bug in generator!

packetbeat/protos/cassandra/cassandra.go
```diff
@@ -0,0 +1,200 @@
...
+	// set parser configuration
+	parser := &cassandra.parserConfig
+	parser.maxBytes = tcp.TCP_MAX_DATA_IN_STREAM
+
+	// set parser's compressor, only `snappy` supported right now
+	if config.Compressor == Snappy {
+		parser.compressor = SnappyCompressor{}
+	} else {
+		parser.compressor = nil
+	}
```
urso: does network protocol include the compressor name or some id?

packetbeat/protos/cassandra/config.go
```diff
@@ -0,0 +1,27 @@
...
+		ProtocolCommon: config.ProtocolCommon{
+			TransactionTimeout: protos.DefaultTransactionExpiration,
+		},
+		SendRequestHeader:  false,
+		SendResponseHeader: false,
+	}
+)
+
+func (c *cassandraConfig) Validate() error {
+	return nil
```
urso: check here if Compressor is valid (empty or 'snappy'). This will generate a somewhat readable error message pointing to the config option and config file (or CLI overwrite) that is wrong.

packetbeat/protos/cassandra/frame.go
```diff
@@ -0,0 +1,900 @@
...
+      +---------+---------+---------+---------+
+      |                                       |
+      .            ...  body ...              .
+      .                                       .
+      .                                       .
+      +----------------------------------------
+
+  The protocol is big-endian (network byte order).
+
+  some code derived from https://github.com/gocql/gocql
```
urso: makes me wonder if gocql could be used directly instead of copying and modifying code.

packetbeat/protos/cassandra/parser.go
```diff
@@ -0,0 +1,173 @@
...
+	return &message{
+		Message: applayer.Message{
+			Ts: ts,
+		},
+	}
+}
+
+func (p *parser) parse() (*message, error) {
+
+	r := bytes.NewReader(p.buf.Bytes())
```
urso: p.buf already implements io.Reader.

packetbeat/protos/cassandra/parser.go
```diff
@@ -0,0 +1,173 @@
...
+	head, err := readHeader(r, make([]byte, 9))
```
urso: check via p.buf.Available() if enough space is available.

packetbeat/protos/cassandra/frame.go
```diff
@@ -0,0 +1,900 @@
...
+	f.traceID = nil
+
+	return f
+}
+
+type frame interface {
+	Header() frameHeader
+}
+
+func readHeader(r io.Reader, p []byte) (head frameHeader, err error) {
```
urso: readHeader can be replaced with 'func unmarshalHeader(buf *streambuf.Buffer) (head frameHeader, err error)`. No need to allocate additional reader and buffer to be passed to readHeader.

packetbeat/protos/cassandra/parser.go
```diff
@@ -0,0 +1,173 @@
...
+	if err != nil {
+		logp.Err(err.Error())
+		return nil, nil
```
urso: errors in readHeader seem critical. Why supress error here? I think TCP stream should be dropped. Instead of dropping stream consider marking stream as unintersting (e.g. on version mismatch), so we don't wast additional time analyzing packets.

packetbeat/protos/cassandra/parser.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	if logp.IsDebug("cassandra") {
+		logp.Debug("cassandra", fmt.Sprint(head))
+	}
```
urso: header contains complete message length. Check complete message being contained in buffer now (via `p.buf.Avail`). In case complete message is not included we can stop processing here, saving some precious CPU

urso: Consider storing header info + some parsing state in parser for reuse. That is, if we stop here but header is already parsed, we do not have to parse header again. This allows us to use p.buf as io.Reader into gocql based parser among different calls if message has not been complete in the past.

packetbeat/protos/cassandra/parser.go
```diff
@@ -0,0 +1,173 @@
...
+		logp.Err(err.Error())
+		return nil, nil
+	}
+	msg := p.message
+
+	data, err := framer.parseFrame(msg)
+
+	if err != nil {
+		logp.Err(err.Error())
+		return nil, nil
```
urso: what's the impact for the analyzer if parsing fails?

packetbeat/protos/cassandra/parser.go
```diff
@@ -0,0 +1,173 @@
...
+		},
+	}
+}
+
+func (p *parser) parse() (*message, error) {
+
+	r := bytes.NewReader(p.buf.Bytes())
+	head, err := readHeader(r, make([]byte, 9))
+	if err != nil {
+		logp.Err(err.Error())
```
urso: never do `logp.Err(<go-expression>)`. `logp.Err` expects you to pass format string, which is interpreted when being printed. Better: `logp.Err("%v", err)`.

packetbeat/protos/cassandra/parser.go
```diff
@@ -0,0 +1,173 @@
...
+	dir := applayer.NetOriginalDirection
+
+	isRequest := true
+	if head.version.response() {
+		dir = applayer.NetReverseDirection
+		isRequest = false
+	}
+
+	//collect and wait for enough stream
+	_, err = p.buf.Collect(head.length + 9)
```
urso: if parseFrame consumes all bytes on wire and `p.buf` is passed to `newFramer` instead, this call is not required. The reader in `p.buf` keeps a reading pointer.

packetbeat/protos/cassandra/frame.go
urso: This file seems to be mostly a copy of gocql. How about putting the mostly copied code into separate sub-package and export the required symbols?

E.g. `github.com/elastic/beats/packetbeat/protos/cassandra/internal/gocql`.

This package will be used by cassandra parser only + we can add notes including original license and so on.

packetbeat/protos/cassandra/frame.go
urso: while readHeader can be rewritten to accept `*streambuf.Buffer` directly, maybe we shouldn't as this is directly copied from gocql (do not change original code too much).

Instead in parse check the header can be read `p.buf.Avail(9)` and pass the `*streambuf.Buffer` (which implements io.Reader) directly to readHeader. Store important header fields (or complete header if required) in parser + some state-flag checking the header must be parsed or not.

packetbeat/protos/cassandra/frame.go
```diff
@@ -0,0 +1,900 @@
...
+	traceID []byte
+
+	// holds a ref to the whole byte slice for rbuf so that it can be reset to
+	// 0 after a read.
+	readBuffer []byte
+
+	rbuf []byte
+}
+
+func newFramer(r io.Reader, compressor Compressor, version byte) *framer {
```
urso: disadvantage of `framer` as is, it implements it's own buffering, basically copying the complete message being already in memory once again. This results in increased memory and cpu usage. Given we have the complete message available in memory alread, the framer does not require to read the message via io.Reader in `(*framer).readFrame`.

Instead using `bytes, err := p.buf.Collect(head.length)` would contain the complete frame body in `bytes` one can use to initialize the readBuffer too.

packetbeat/protos/cassandra/frame.go
```diff
@@ -0,0 +1,900 @@
...
+	} else {
+		f.readBuffer = make([]byte, head.length)
+		f.rbuf = f.readBuffer
+	}
+
+	// assume the underlying reader takes care of timeouts and retries
+	n, err := io.ReadFull(f.r, f.rbuf)
+	if err != nil {
+		return fmt.Errorf("unable to read frame body: read %d/%d bytes: %v", n, head.length, err)
+	}
```
urso: `io.ReadFull` becomes obsolete if `f.rbuf` is initialized with buffer from `p.buf`.

packetbeat/protos/cassandra/parser.go
urso: Checking readFrame method, a check for available bytes is required, so not to fail the framer (it's using `io.ReadFull`).

urso:
Looks quite good. Just some minor issues.

Have you considered to fork the 'framer' of from gocql (or refactor gocql a little to make Framer available)? Would be nice to have the backing CQL parser vendored instead code being copied right into the module.

packetbeat/protos/cassandra/cassandra.go
medcl: @urso one bit in header info of the network protocol indicates that whether it compressed or not, and the compressor should be configed during the start time

packetbeat/protos/cassandra/parser.go
medcl: supposed to drop the events

packetbeat/protos/cassandra/cassandra.go
urso: naming the compression algorithm is part of the startup-method. There is a chance we never see the startup-message, but if we do, we should check the compressor matches.

If decompression fails (indicator of wrong compressor being configured), we should set the stream state to don't follow stopping processing of all compressed frames (uncompressed frames can still be processed).

packetbeat/protos/cassandra/parser.go
urso: parser failing is an indicator of:

- not really dealing with CQL
- parser not in sync to TCP-stream (e.g. after gap in stream or on packetbeat startup)

The later case requires to drop the complete stream, so parser can be re-initialized. This should happen until parser is eventually in sync to tcp stream.

medcl: @urso how to drop the complete stream, i was thought if i return nil here, the stream will ignored, seems i am wrong

packetbeat/protos/cassandra/parser.go
medcl: @urso agreed to mark rather dropping, and here the question comes: how to mark the stream unintersting?

medcl: should i call "r.Reset()" to clean up the buffer, if something is wrong

packetbeat/protos/cassandra/cassandra.go
medcl: ++, and  if the stream is compressed, i think we have to read the data into the buffer to do decompress, no way to direct to use p.buf to read the data, right?

urso: right, we will need some new buffer to put compressed information into. But I think the parser as is supports decompression (due to managing an internal read buffer). 

In case you want to handle decompression yourself: If compression can not span multiple frames (need to check protocol this is true), it's quite straight forward to use `p.buf.Collect()` to collect the compressed frames into `[]byte` or better wrap `p.buf` with `LimitReader` and pass the reader to `zlib.Reader` (the reader allocated from zlib is resettable with a new reader, such that no new buffers need to be allocated -> reduces garbage to be collected). If using the later approach check for possibly unread bytes in presence of potential padding, though.

From CQL doc:

> As of this version 2 of the protocol, the following compressions are available:
    - lz4 (https://code.google.com/p/lz4/). In that, note that the 4 first bytes
      of the body will be the uncompressed length (followed by the compressed
      bytes).
    - snappy (https://code.google.com/p/snappy/). This compression might not be
      available as it depends on a native lib (server-side) that might not be
      avaivable on some installation.

That is, if lz4 is used, we do know the required buffer size for decompressed content before decompression. Same goes for snappy:

> The stream starts with the uncompressed length (up to a maximum of 2^32 - 1),
stored as a little-endian varint. Varints consist of a series of bytes,
where the lower 7 bits are data and the upper bit is set iff there are
more bytes to be read. In other words, an uncompressed length of 64 would
be stored as 0x40, and an uncompressed length of 2097150 (0x1FFFFE)
would be stored as 0xFE 0xFF 0x7F.

Will check tomorrow how/if decompression can be handled when doing code review.

medcl: yep, the header is not compressed and with fixed 9 bytes, and also have flags to indicate whether the body frame is compressed or not, and also have the length of the body in header,what the the original parser did is read the whole compressed content and do decompression, seems not efficient, but currently i didn't dealing with compressed protocol, i am do more testing with uncompressed protocol, and then working on compressed one

packetbeat/protos/cassandra/parser.go
```diff
@@ -0,0 +1,246 @@
...
+		p.framer = NewFramer(&p.buf, p.config.compressor)
+	}
+
+	// check if the frame header were parsed or not
+	if p.framer.Header == nil {
+		if isDebug {
+			logp.Debug("cassandra", "start to parse header")
+		}
+		if !p.buf.Avail(9) {
+			logp.Err("not enough head bytes, ignore")
```
urso: this is no error, but an indicator no complete frame header being in buffer. downgrade to Debug

packetbeat/protos/cassandra/parser.go
```diff
@@ -0,0 +1,246 @@
...
+		},
+	}
+}
+
+func (p *parser) parse() (*message, error) {
+
+	// if p.frame is nil then create a new framer, or continue to process the last message
+	if p.framer == nil {
+		if isDebug {
+			logp.Debug("cassandra", "start new framer")
```
urso: use `debugf(format-string, ...)` instead of `logp.Debug("cassandra", format-string, ...)`. This reduces the chance of typos in debug selectors + reduces line length + some linters will check the parameters being correct if debugf is used.

packetbeat/protos/cassandra/parser.go
```diff
@@ -0,0 +1,246 @@
...
+		_, err := p.framer.ReadHeader()
+		if err != nil {
+			logp.Err("%v", err)
+			p.framer = nil
+			return nil, err
+		}
+	}
+
+	msg := p.message
+	msg.data = make(map[string]interface{})
```
urso: `msg.data` should be initialized when creating the message. Here, if not enough bytes are available, msg.data will be reinitialized on next parse attempt.

Also prefer initializer-syntax over make:

```
msg.data = map[string]interface{}{}
```

packetbeat/protos/cassandra/parser.go
```diff
@@ -0,0 +1,246 @@
...
+
+	if p.framer.Header.BodyLength > 0 {
+
+		//let's wait for enough buf
+		if !p.buf.Avail(p.framer.Header.BodyLength) {
+			if isDebug {
+				logp.Debug("cassandra", " buf not enough for body, waiting for more, return")
```
urso: debugf

packetbeat/protos/cassandra/parser.go
```diff
@@ -0,0 +1,246 @@
...
+			}
+			return nil, nil
+		} else {
+			//check if the ops already ignored
+			if p.config.ignoredOps != nil && len(p.config.ignoredOps) > 0 {
+				if isDebug {
+					logp.Debug("cassandra", "ignoreOPS configed, let's check")
```
urso: debugf

packetbeat/protos/cassandra/parser.go
```diff
@@ -0,0 +1,246 @@
...
+
+	if p.framer.Header.BodyLength > 0 {
+
+		//let's wait for enough buf
+		if !p.buf.Avail(p.framer.Header.BodyLength) {
+			if isDebug {
+				logp.Debug("cassandra", " buf not enough for body, waiting for more, return")
+			}
+			return nil, nil
+		} else {
```
urso: do not use `else` branch, as `if !p.buf.Avail ...` branch returns early.

packetbeat/protos/cassandra/parser.go
```diff
@@ -0,0 +1,246 @@
...
+			logp.Err("%v", err)
+			p.framer = nil
+			return nil, err
+		}
+	}
+
+	msg := p.message
+	msg.data = make(map[string]interface{})
+
+	if p.framer.Header.BodyLength > 0 {
```
urso: parsing body quite 'long'. consider refactoring this  into `parseBody` method

urso: parsing body quite 'long'. consider refactoring this  into `parseBody` method

urso: parsing body quite 'long'. consider refactoring this  into `parseBody` method

packetbeat/protos/cassandra/cassandra.go
```diff
@@ -0,0 +1,215 @@
...
+
+	// parsed ignored ops
+	if len(config.IgnoredOPs) > 0 {
+		maps := make(map[string]interface{})
+
+		if strings.Contains(config.IgnoredOPs, ",") {
+			array := strings.Split(config.IgnoredOPs, ",")
+			for i := 0; i < len(array); i++ {
```
urso:
```
for _, str := range strings.Split(config.IgnoredOPs, ",") {
}
```

ops might be invalid, due to being strings. Rather convert names to opcodes failing if unknown opcode is found.

Due to all opcodes being integers consider from 0 to 16 consider using a mask like `[16]bool` or a bitset of at least 17 bits (either `[3]uint8` or `uint32`) for checking an opcode shall be ignored.

urso: implementation details for ignoring ops can be hidden with custom type

```
type opsMask struct {
   ...
}

func (m *opMask) Set(op ...) { ... }

func (m *opMask) IsSet(op ...) bool { ... }
```

and methods

packetbeat/protos/cassandra/parser.go
```diff
@@ -0,0 +1,246 @@
...
+					p.framer = nil
+					return nil, err
+				}
+
+				msg.data = data
+
+				// dealing with un-parsed content
+				frameParsedLength := p.buf.BufferConsumed()
+
+				// collect leftover
```
urso: this due to padding? Are message padded to 2, 4 or 8 bytes by default/

medcl: not padding, some message detail is less interested, i just parser the metadata, but the left content also need to read but won't process

packetbeat/protos/cassandra/cassandra.go
```diff
@@ -0,0 +1,218 @@
...
+		parser.compressor = SnappyCompressor{}
+	} else {
+		parser.compressor = nil
+	}
+
+	// parsed ignored ops
+	if len(config.OPsIgnored) > 0 {
+		maps := map[FrameOp]bool{}
+
+		if strings.Contains(config.OPsIgnored, ",") {
```
urso: sorry, didn't notice it earlier. Do not start splitting strings, but rely on structure provided by configuration support. This is mostly about consistency, as we do this for all fields potentially being an array of strings. If we want to support alternative syntax in future, beats config file reader should have to handle this, but no components should have to care.

Use type `[]string` for `config.OpsIgnored`. This way syntax in yaml file config files becomes

```
ignored_ops:
  - OP1
  - OP2
```

or 

```
ignored_ops: [OP1, OP2]
```

or 

```
ignored_ops: ["OP1", "OP2"]
```
    
```

File: packetbeat/protos/cassandra/cassandra.go [hunk 27]
```diff
@@ -0,0 +1,218 @@
...
+			array := strings.Split(config.OPsIgnored, ",")
+			for i := 0; i < len(array); i++ {
+				str := array[i]
+				if len(str) > 0 {
+					op := FrameOpFromString(strings.ToUpper(strings.TrimSpace(str)))
+					maps[op] = true
+				}
+			}
+		} else {
+			op := FrameOpFromString(strings.ToUpper(strings.TrimSpace(config.OPsIgnored)))
```
urso: what if trimmed string is empty?

consider FrameOpFromString having to deal with case-sensitivity + report some error if op is unknown =>

```
func FrameOpFromString(s string) (opcode, error)
```

packetbeat/protos/cassandra/config.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/elastic/beats/packetbeat/protos"
+	"github.com/pkg/errors"
+)
+
+type cassandraConfig struct {
+	config.ProtocolCommon `config:",inline"`
+	SendRequestHeader     bool   `config:"send_request_header"`
+	SendResponseHeader    bool   `config:"send_response_header"`
+	Compressor            string `config:"compressor"`
+	OPsIgnored            string `config:"ignored_ops"`
```
urso: use `[]string` for `OpsIgnored`

packetbeat/protos/cassandra/internal/gocql/marshal.go
```diff
@@ -0,0 +1,701 @@
...
+	case opAuthResponse:
+		return "AUTH_RESPONSE"
+	case opAuthSuccess:
+		return "AUTH_SUCCESS"
+	default:
+		return fmt.Sprintf("UNKNOWN_OP_%d", f)
+	}
+}
+
+func FrameOpFromString(str string) FrameOp {
```
urso: consider:

```
var frameOps = map[string]FrameOP {
  "ERROR": opError,
  ...
}

func FrameOfFromString(s string) (FrameOp, error) {
  s = strings.ToUpper(strings.Trim(s))
  op, found := frameOps[s]
  if !found {
    return opUnknown, fmt.Errorf("unknown frame op: %v", s)
  }
  return op, nil
}

File: packetbeat/protos/cassandra/config.go (hunk 28 above)
Comment (urso) [inline packetbeat/protos/cassandra/config.go]: alternatively you can use the ucfg Unpacker interface to push parsing frame ops into the config loading phase (better error message naming exact faulty setting) like this:

```
type cassandraConfig struct {
    ...
    OpsIgnored []FrameOp
    ...
}
```

with

```
func (f *FrameOp) Unpack(in interface{}) error {
    s, ok := in.(string)
    if !ok {
      return errors.New("expected string")
    }

    op, err := FrameFromOpString(s)
    if err != nil {
        return err
    }

    *f = op
}
```

File: packetbeat/protos/cassandra/parser.go [hunk 30]
```diff
@@ -0,0 +1,266 @@
...
+func (p *parser) feed(ts time.Time, data []byte) error {
+	if err := p.append(data); err != nil {
+		return err
+	}
+
+	for p.buf.Total() > 0 {
+		if p.message == nil {
+			// allocate new message object to be used by parser with current timestamp
+			p.message = p.newMessage(ts)
+			p.message.data = map[string]interface{}{}
```
urso: why not have newMessage initialize the `data` field?

packetbeat/protos/cassandra/parser.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+
+	return nil
+}
+
+func (p *parser) newMessage(ts time.Time) *message {
+	return &message{
+		Message: applayer.Message{
+			Ts: ts,
+		},
```
urso: add here `data: map[string]interface{}{}`

packetbeat/protos/cassandra/parser.go
```diff
@@ -0,0 +1,266 @@
...
+		_, err := p.framer.ReadHeader()
+		if err != nil {
+			logp.Err("%v", err)
+			p.framer = nil
+			return nil, err
+		}
+	}
+
+	//check if the ops need to be ignored
+	if p.FrameOpsIgnored() {
```
urso: consider renaming `CheckFrameOpIgnored` or `IsFrameOpIgnored` or `CheckIgnoreBody` to make flow more clear.

packetbeat/protos/cassandra/parser.go
```diff
@@ -0,0 +1,266 @@
...
+		// as we already ignore the content, we now mark the result is ignored
+		p.message.ignored = true
+		if isDebug {
+			debugf("Ops: %s was marked to be ignored, ignoring, request:%v", p.framer.Header.Op.String(), p.framer.Header.Version.IsRequest())
+		}
+	}
+
+	msg := p.message
+
+	data, err := p.parserBody()
```
urso: consider parserBody to return an additional bool. Creating a map will allocate an empty map, which might not really be used. With `message.data` already being allocated, parseBody should write into `message.data` and return bool if finished.

packetbeat/protos/cassandra/parser.go
```diff
@@ -0,0 +1,266 @@
...
+
+func (p *parser) newMessage(ts time.Time) *message {
+	return &message{
+		Message: applayer.Message{
+			Ts: ts,
+		},
+	}
+}
+
+func (p *parser) parserBody() (map[string]interface{}, error) {
```
urso: somewhat simplified/flattened parserBody:

```
func (p *parser) parserBody() (done bool, err error) {
  headLen := p.framer.Header.HeadLength
  bdyLen := p.framer.Header.BodyLength
  if bdyLen <= 0 {
    return true, nil
  }
  

  //let's wait for enough buf
  debugf("bodyLength: %d", bdyLen)
  if !p.buf.Avail(bdyLen) {
    if isDebug {
      debugf("buf not enough for body, waiting for more, return")
    }
    return false, nil
  }

  //check if the ops already ignored
  if p.message.ignored {
    if isDebug {
      debugf("message marked to be ignored, let's do this")
    }
    p.buf.Collect(bdyLen)
  } else {
    // start to parse body
    data, err := p.framer.ReadFrame()
    if err != nil {
      // if the frame parsed failed, should ignore the whole message
      p.framer = nil
      return false, err
    }

    // dealing with un-parsed content
    frameParsedLength := p.buf.BufferConsumed()

    // collect leftover
    unParsedSize := bdyLen + headLen - frameParsedLength
    if unParsedSize > 0 {
      if !p.buf.Avail(unParsedSize) {
        err := errors.New("should be enough bytes for cleanup,but not enough")
        logp.Err("Finishing frame failed with: %v", err)
        return false, err
      }

      p.buf.Collect(unParsedSize)
    }

    p.message.data = data
  }

  finalCollectedFrameLength := p.buf.BufferConsumed()
  if finalCollectedFrameLength-headLen != bdyLen {
    logp.Err("body_length:%d, head_length:%d, all_consumed:%d",
      bdyLen, headLen, finalCollectedFrameLength)
    return nil, errors.New("data messed while parse frame body")
  }

	return true, nil
}
```

packetbeat/protos/cassandra/parser.go
urso: why initialize p.message.data with valid map at all, why not use nil. `message.data` will be overwritten by parser anyway.

packetbeat/protos/cassandra/internal/gocql/frame.go
```diff
@@ -0,0 +1,530 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		decoder.Data = &dec
+		f.decoder = decoder
+
+		debugf("hit compress flags")
+
+		return nil, errors.New("Compressed content not supported yet")
```
urso: so we first decompress body and return error about not supporting compressed content next? I'm a little confused what's correct here.

packetbeat/protos/cassandra/config.go
medcl: That's nice, i never know that

packetbeat/protos/cassandra/internal/gocql/frame.go
medcl: Ah, this line should be removed. good catch, Thanks.

packetbeat/protos/cassandra/pub.go
```diff
@@ -67,11 +67,17 @@ func (pub *transPub) createEvent(requ, resp *message) common.MapStr {
...
+			}
+
 			if pub.sendRequestHeader {
 				requ.data["request_headers"] = requ.header
 			}
 
-			event["cassandra_request"] = requ.data
+			if len(requ.data) > 0 {
+				event["cassandra_request"] = requ.data
+			}
```
urso: block can be changed to:

```
if pub.sendRequestHeader {
    if requ.data == nil {
        requ.data = map[string]interface{}{}
    }
    requ.data["request_headers"] = requ.header
}

if len(requ.data) > 0 {
    event["cassandra_request"] = requ.data
}
```

`len(requ.data) == 0` if `requ.data == nil`.

packetbeat/protos/cassandra/pub.go
```diff
@@ -97,11 +103,18 @@ func (pub *transPub) createEvent(requ, resp *message) common.MapStr {
...
+		}
+
 		if pub.sendResponseHeader {
 			resp.data["response_headers"] = resp.header
 		}
 
-		event["cassandra_response"] = resp.data
+		if len(resp.data) > 0 {
+			event["cassandra_response"] = resp.data
+		}
```
urso: same as for requ.data

urso:
Code and Tests look really good.

I just figured some documentation needs to be updated:
- [packetbeat-options.asciidoc](https://github.com/elastic/beats/blob/master/packetbeat/docs/reference/configuration/packetbeat-options.asciidoc)
- [shared-protocol-list.asciidoc](https://github.com/elastic/beats/blob/master/packetbeat/docs/shared-protocol-list.asciidoc)

Would be shame if cassandra gets implemented and no one finds it in docs.

packetbeat/docs/reference/configuration/packetbeat-options.asciidoc
```diff
@@ -659,6 +663,56 @@ at the end of the document.
...
+  send_request: true
+  send_request_header: true
+  send_response: true
+  send_response_header: true
+  compressor: "snappy"
+  ignored_ops: ["SUPPORTED","OPTIONS"]
+------------------------------------------------------------------------------
+
+
+===== send_request
```
urso: like `ports, `send_request` and `send_respose` is already part of common protocol options (already documented)

packetbeat/docs/reference/configuration/packetbeat-options.asciidoc
```diff
@@ -659,6 +663,56 @@ at the end of the document.
...
+is included in published events. The default is true. enable `send_response` first before enable this option.
+
+===== ignored_ops
+
+This option indicates which Operator/Operators will be ignored.
+
+===== compressor
+
+If this option is enabled, and also the flag indicates that the frame body was compressed, and
+the stream will decode by the compressor, currently only support `snappy` , The default compressor is nil.
```
urso: `
If this option is enabled, and also the flag indicates that the frame body was compressed, and
the stream will decode by the compressor, currently only support `snappy` , The default compressor is nil.
`

It's unclear if this is a boolean or some string.

e.g.

`Configures the default compression algorithm being used to uncompress compressed frames by name. Currently only `snappy` is can be configured. By default no compressor is configured`.

packetbeat/docs/reference/configuration/packetbeat-options.asciidoc
```diff
@@ -659,6 +663,56 @@ at the end of the document.
 Note that limiting documents in this way means that they are no longer correctly
 formatted JSON objects.
 
+
+[[configuration-cassandra]]
+==== Cassandra Configuration Options
```
urso: please add section in alphabetical order

medcl:
Kibana Dashboard Screenshoot

PR #1963, opened 2016-07-05
How does Filebeat work

ruflin:
Guide on how Filebeat works and some config options.

@dedemorton The intention is to have a general guide for filebeat. The second part about the config options should probably be moved to an other section as it is for "exeperienced users".

### DIFF
--- a/filebeat/docs/how-filebeat-works.asciidoc
+++ b/filebeat/docs/how-filebeat-works.asciidoc
@@ -1,0 +1,43 @@
+== How Filebeat works
+
+This page intends to explain what the key building blocks of Filebeat are and how they work together. The goal is that the available configuration options can be better understood and more informed decisions can be made about the optimal configuration options for the different use cases.
+
+[float]
+=== What is a harvester?
+
+A harvester is responsible to read the content of a single file. This is done by reading each file line by line and send the content to the output. For each file one harvester is started. The harvester is responsible to open and close files. That means, as long as a harvester is running, the file descriptor stays open. Even if a file is removed or renamed, filebeat will keep reading the file. This has the side affect that the space on your disk will be reserved until the harvester is stopped.
+
+Stopping a harvester has the following consequences:
+
+* The file handler is closed, freeing up the underlying resources if the file was deleted
+* The harvesting of the file will only be started again after `scan_frequency`
+* In case the file was moved / removed, harvesting the file will not continue
+
+To define more in detail when a harvester is closed, use the <<close-options>> configuration options.
+
+[float]
+=== What is a prospector?
+
+A prospector is reponsible to manage the harvesters.  The responsibility of the prospector is to find all sources it should read from. In case of the log input type it is to find all files on the drive based on the defined paths and start a harvester for each file that was found. Each prospector runs in its own go routine and is responsible to manage the harvesters. Each prospector configuration looks as following:
+
+[source,yaml]
+-------------------------------------------------------------------------------------
+filebeat.prospectors:
+- input_type: log
+  paths:
+    - /var/log/*.log
+    - /var/path2/*.log
+-------------------------------------------------------------------------------------
+
+Filebeat currently supports two `prospector` types: `log` and `stdin`. Each prospector type can be defined multiple times. The log prospector checks for each file if a harvester has to be started, if one is already running or the file can be ignored because of configuration options which were set. New files are only picked up, if the offset / size of the file changed since the harvester was closed.
+
+[float]
+=== What is a state?
+
+Filebeat keeps the state of each file and frequently flushes the state to disk to the registry file. The state is used to remember the last offset a harvester was reading from and to ensure all log lines are sent. In case Elasticsearch/Logstash are not reachable, Filebeat will keep track of the last lines sent and will continue reading the files as soon Elasicsearch/Logstash becomes available again. As long as filebeat is running, the state information is also kept in memory by each prospector. In case of a filebeat restart, the data from the registry file is used to rebuild the state and filebeat continues each harvester at the last known position.
+
+Each prospector keeps a state for each file it finds. As files can be renamed or moved, the filename and path are not enough to identify a file. For each file unique identifiers are stored to detect if a file the same file that was harvested previously.
+
+For more details on how to configure the state, checkout out the <<reduce-registry-size>> troubleshooting guide.
+
+
--- a/filebeat/docs/index.asciidoc
+++ b/filebeat/docs/index.asciidoc
@@ -17,6 +17,8 @@ include::./overview.asciidoc[]
 
 include::./getting-started.asciidoc[]
 
+include::./how-filebeat-works.asciidoc[]
+
 include::./command-line.asciidoc[]
 
 include::../../libbeat/docs/shared-directory-layout.asciidoc[]
--- a/filebeat/docs/reference/configuration/filebeat-options.asciidoc
+++ b/filebeat/docs/reference/configuration/filebeat-options.asciidoc
@@ -164,38 +164,115 @@ names added by Filebeat, then the custom fields overwrite the other fields.
 [[ignore-older]]
 ===== ignore_older
 
-If this option is specified, Filebeat
-ignores any files that were modified before the specified timespan. This is disabled by default.
+If this option is specified, Filebeat ignores any files that were modified before the specified timespan. This is disabled by default.
 
 You can use time strings like 2h (2 hours) and 5m (5 minutes). The default is 0, which means disable.
 Commenting out the config has the same affect as setting it to 0.
 
-Files which were falling under ignore_older and are updated again, will start
-from the offset the file was at when it was last ignored by ignore_older. As an example:
-A file was not modified for 90 hours and the offset is at 200. Now a new line is added and
-the last modification date is updated. After scan_frequency detects the change the crawling
-starts at the offset 200. In case the file was falling under ignore_older already when filebeat
-was started, the first 200 lines are never sent. In case filebeat was started earlier, the 200
-chars were already sent and it now continues at the old offset.
+There are two different cases for files which fall under ignore_older:
+
+* Files which were never harvested
+* Files which were harvested but weren't updated for longer then ignore_older
+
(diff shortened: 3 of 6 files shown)


dedemorton:
@ruflin I think we need to be careful here not to duplicate info that really belongs in the reference section for config options. This topic should definitely describe how Filebeat works conceptually and mention config settings that control behavior, but any detailed guidance about setting options belongs either in the reference content or in topics under "Configuring Filebeat." Perhaps you are thinking the same thing when you mention having content for experienced users. :-)

ruflin:
@dedemorton Exactly. I'm currently putting everything into 1 document and we should then decide based on the content what fits where. I will need your help here ;-)

filebeat/docs/how-does-filebeat-work.asciidoc
```diff
@@ -2,68 +2,68 @@
 
 This page intends to explain what the key building blocks of Filebeat are and how they work together. The goal is that the available configuration options can be better understood on what they do.
 
-=== What is a Prospector?
+=== What is an Input?
```
dedemorton: From the perspective of working with the config file, "input" is definitely easier to follow and more intuitive.

My concern with renaming "prospector" to "input" is when we start to talk about an input conceptually as a component that does something. We also have the concept of inputs and outputs in Logstash. Our concept of outputs in Beats is pretty similar to Logstash; you specify an output based on the specific type of system that you want to send events to. But in the context of Filebeat, an input does not read from a specific type of system. It's more about specifying the paths to files and defining how you want Filebeat to process the log lines.

When you start explaining things conceptually, I think that "input" starts to sound a bit awkward. The word "input" by itself is almost too generic. It doesn't sound like a component that actually does something. In Logstash, they get around this problem by talking about "input stages" and "input plugins". I think maybe you ran into this problem yourself because you use "input module" below. 

I wonder if we could rename the section without talking about an input as an actual component of functionality within Filebeat.

filebeat/docs/how-does-filebeat-work.asciidoc
```diff
@@ -2,68 +2,68 @@
...
+Filebeat keeps the state of each file and frequently flushs it to disk. The state is used to remember the last offset a reader was reading from and to ensure all log lines are sent. In case Elasticsearch or Logstash are down, Filebeat will keep the track of the last lines sent for each file and will continue reading the file, as soon elasticsearch becomes available again. As long as filebeat is running, all this information is also kept in memory but frequently written to disk to the `registry_file`. In case of a restart of filebeat, the data from the registry file is used to rebuild the state and filebeat continues at the last known position.
```
dedemorton: I'm not sure why you didn't like the idea of explaining the registry To me, state is something that gets stored in the registry. I don't find that confusing. I think it makes sense to describe the registry as a component of the architecture because the registry file is something that users might need to delete.

filebeat/docs/how-does-filebeat-work.asciidoc
```diff
@@ -2,68 +2,68 @@
...
 
-In the registry file, each havested file has an entry. As files can be renamed or moved, the filename and path are not enough to identify a file. For each file unique identifiers are stored to detect if a file the same file that was harvested previously.
+The input keeps a state for each file it finds. As files can be renamed or moved, the filename and path are not enough to identify a file. For each file unique identifiers are stored to detect if a file the same file that was harvested previously.
```
dedemorton: This is an example of where talking about "input" as a component is a bit awkward. Wouldn't it be OK here just to say "Filebeat keeps the state for each file that it finds"?

filebeat/docs/how-does-filebeat-work.asciidoc
```diff
@@ -2,68 +2,68 @@
...
 
  * scan_frequency
  * ignore_older
 
 ===== scan_frequency
 
-The scan frequency defines how often a prospector scans the file system for new and existing files and checks if for each file a harvester is running (if needed). So in case a harvester is closed, it will be latest restarted again after `scan_frequency`. The default value for `scan_frequency` is 10s. This should work in most cases. Setting this value too low, leads to too much resources used from the file system as all the files are constantly scanned, setting it too high will take longer until new files are picked up. We do not recommend to set this value `<1s`.
+The scan frequency defines how often the input scans the file system for new and existing files and checks if for each file a reader is running (if needed). So in case a reader is closed, it will be latest restarted again after `scan_frequency`. The default value for `scan_frequency` is 10s. This should work in most cases. Setting this value too low, leads to too much resources used from the file system as all the files are constantly scanned, setting it too high will take longer until new files are picked up. We do not recommend to set this value `<1s`.
```
dedemorton: Again, here "input" seems awkward. I won't mark all of these, but the problem is that people think of the input as being the file(s) that you're pointing to and not a component that actually does something.

filebeat/docs/how-does-filebeat-work.asciidoc
```diff
@@ -73,27 +73,45 @@ It can happen that a file is renamed. This happens for example when rotating fil
 
 This option allows a file handler to be closed when it is renamed or moved. In case this option is enabled and the file renamed/moved in such a way that is not part of the prospector patterns, the file will not be picked up again. It will not finish reading the file.
 
+WINDOWS: In case under windows your log rotation system shows error because it can't rotated the files, this is the option to enabled.
+
 ===== close_removed
 
-Close removed can be used to close files that were removed from disk but are still open by the harvester. The harvester will  close the file after it was removed.
+Close removed can be used to close files that were removed from disk but are still open by the harvester. The harvester will close the file after it was removed.
```
dedemorton: You still have some references to harvester in this.

filebeat/docs/how-does-filebeat-work.asciidoc
```diff
@@ -0,0 +1,76 @@
...
+
+```
+filebeat.prospectors:
+- input_type: log
+  paths:
+    - /var/log/*.log
+    - /var/path2/*.log
+```
+
+The log prospector is responsible to find all files which were defined under the paths above and check, if for each file a harvester has to be started, if one is already running or the file can be ignored because of some configuration options. Each prospector runs in its own go routine and is responsible to manage the harvesters. New files are only picked up, if the offset / size of the file changed since the harvester was closed.
```
tsg: Would probably be better to avoid "harvester" in this paragraph, since it's not yet introduced.

tsg: If we rename the harvester it to reader, that would make it easier, I think.

filebeat/docs/how-does-filebeat-work.asciidoc
```diff
@@ -0,0 +1,76 @@
...
+
+=== What is a harvester?
+
+A harvester is responsible to read the content of a single file. That is done by reading each file line by line and send the information to the output. Each harvester is started by its prospector and retrieves all configuration options from the prospector. The harvester is responsible to open and close files. That means, as long as a harvester is running, the file descriptor stays open. Even if a file is removed or renamed, filebeat will keep reading the file. This has the side affect that the space on your disk will be reserved until the harvester is stopped. Stopping a harvester means:
+
+* The file handler is closed
```
tsg: Maybe "The file handler is closed, freeing up the underlying resources if the file was deleted".

filebeat/docs/how-does-filebeat-work.asciidoc
```diff
@@ -0,0 +1,76 @@
...
+=== What is the state?
+
+Filebeat keeps the state of each file and frequently flushs it to disk to the registry file. The state is used to remember the last offset a harvester was reading from and to ensure all log lines are sent. In case Elasticsearch or Logstash are down, Filebeat will keep the track of the last lines sent for each file and will continue reading the file as soon elasticsearch becomes available again. As long as filebeat is running, all this information is kept in memory and frequently written to disk to the `registry_file`. In case of a restart of filebeat, the data from the registry file is used to rebuild the state and filebeat continues at the last known position.
+
+Each prospector keeps a state for each file it finds. As files can be renamed or moved, the filename and path are not enough to identify a file. For each file unique identifiers are stored to detect if a file the same file that was harvested previously.
+
+
+== Use cases
+
+There are common use cases and problems which are described below on how they can be solved with the configuration options.
```
tsg: Consider rephrasing this paragraph, it seems hard to understand for me.

filebeat/docs/how-does-filebeat-work.asciidoc
```diff
@@ -0,0 +1,76 @@
...
+
+=== Reduce open file handlers
+
+Filebeat keeps the file handler open in case it reaches the end of a file to be able to read not log lines in near real time. If filebeat is harvesting a large number of files, the number of open files can be become and issue. As in most environments, the files which are actively updated as low, the configuration `close_idle` should be set accordingly to close files which are not active any more.
```
tsg: "to read new log lines"?

tsg: "can become an issue"

filebeat/docs/how-does-filebeat-work.asciidoc
```diff
@@ -0,0 +1,76 @@
...
+
+There are 4 more configuration options which can be used to close file handlers, but all of them should be used carefully as they can side affects. The options are:
```
tsg: Would be nice to at least give a hint on when these are needed.

filebeat/docs/how-does-filebeat-work.asciidoc
ruflin: I solved it by first explaining the harvester

filebeat/docs/how-does-filebeat-work.asciidoc
ruflin: The complete sentence is now not needed anymore, as I moved these paragraphs to troubleshooting, so the context is already given.

filebeat/docs/how-does-filebeat-work.asciidoc
ruflin: done

filebeat/docs/how-does-filebeat-work.asciidoc
ruflin: added

PR #1964, opened 2016-07-05
Remove optional from config topic headings

dedemorton:
Removed "(Optional)" after config reference section titles because it makes it seem like other sections (which aren't marked optional) are required, which is not true.


ruflin:
@dedemorton should this also be applied to 1.3?

dedemorton:
@ruflin Yes, I will cherry pick the changes.

PR #1969, opened 2016-07-05
Snapshot expvar maps for periodic logging

tsg:
Part of #1931, this adds logic to descend into maps. This is useful especially
to support the metricbeat fetches stats. Locking should be safe as the expvar.Map
Do function takes a read lock on the map.

Note that expvars of type Func are still not included, which is why
the memstats are not logged.


andrewkroh:
LGTM

PR #1970, opened 2016-07-05
Reduce filebeat verbosity at INFO level

tsg:
Part of #1931. Replaces two of the "periodic" Infos with Debugs + expvars.

### DIFF
--- a/filebeat/prospector/prospector.go
+++ b/filebeat/prospector/prospector.go
@@ -129,7 +129,7 @@ func (p *Prospector) Run() {
 			logp.Info("Prospector ticker stopped")
 			return
 		case <-time.After(p.config.ScanFrequency):
-			logp.Info("Run prospector")
+			logp.Debug("prospector", "Run prospector")
 			p.prospectorer.Run()
 		}
 	}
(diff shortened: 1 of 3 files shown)


filebeat/prospector/prospector.go
```diff
@@ -111,7 +111,7 @@ func (p *Prospector) Run() {
 				}
 				select {
 				case <-p.done:
-					logp.Info("Prospector channel stopped")
+					logp.Debug("prospector", "Prospector channel stopped")
```
ruflin: I would probably keep these (all in this file) in Info as they show the shutdown messages (are only showed ones)

filebeat/prospector/prospector.go
```diff
@@ -126,17 +126,17 @@ func (p *Prospector) Run() {
 	for {
 		select {
 		case <-p.done:
-			logp.Info("Prospector ticker stopped")
+			logp.Debug("prospector", "Prospector ticker stopped")
 			return
 		case <-time.After(p.config.ScanFrequency):
-			logp.Info("Run prospector")
+			logp.Debug("prospector", "Run prospector")
```
ruflin: This one should stay in debug.

tsg:
@ruflin Addressed comments.

ruflin:
LGTM

Issue #1494, opened 2016-04-26
Clarify intermediate CA usage in Logstash TLS docs


andrewkroh:
We have a note about "Certificates must be signed by your root CA. Intermediate CAs are currently not supported." We should qualify that point about intermediate CAs. They can only be used if you supply the fully CA chain in a `certificate_authorities` entry. This is caused by logstash-plugins/logstash-input-beats#64.

The PEM file containing the CA chain is created by concatenating the root CA cert and the intermediate CA into a single file: `cat root-ca.crt intermediate-ca.crt > chain.crt` then you set `certificate_authoriities: ['chain.crt']`.

This was prompted by: https://discuss.elastic.co/t/winlogbeat-ignores-ca-certificates-in-windows-certificate-store/48209

dedemorton:
Closing this because we've clarified the docs.

PR #1971, opened 2016-07-05
Modify ca statement

dedemorton:
Just a couple of unrelated changes based on confusion I've seen in forum discussions.

### DIFF
--- a/libbeat/docs/shared-tls-logstash-config.asciidoc
+++ b/libbeat/docs/shared-tls-logstash-config.asciidoc
@@ -22,14 +22,18 @@ To use TLS mutual authentication:
 {beatname_uc} and Logstash. Creating a correct SSL/TLS infrastructure is outside the scope of this
 document. There are many online resources available that describe how to create certificates.
 +
-NOTE: Certificates must be signed by your root CA. Intermediate CAs are currently not supported. You may be able to
-use intermediate CAs by merging all the certificates in the chain into one file.
+NOTE: Certificates must be signed by your root CA. Intermediate CAs are currently not supported.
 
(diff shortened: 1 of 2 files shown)


libbeat/docs/shared-tls-logstash-config.asciidoc
```diff
@@ -22,7 +22,8 @@ To use TLS mutual authentication:
 {beatname_uc} and Logstash. Creating a correct SSL/TLS infrastructure is outside the scope of this
 document. There are many online resources available that describe how to create certificates.
 +
-NOTE: Certificates must be signed by your root CA. Intermediate CAs are currently not supported. 
+NOTE: Certificates must be signed by your root CA. Intermediate CAs are currently not supported. You may be able to
+use intermediate CAs by merging all the certificates in the chain into one file.
```
andrewkroh: Given that this statement is being made in the context of signing client certificates, I don't think that last sentence is relevant. When you sign a client certificate, you sign it with one, and only one, CA certificate.

That statement is relevant when talking about trusting a chain of certificate authorities. So it would be more relevant under the beat's `certificate_authorities` configuration option. With the logstash beats input plugin being re-written in Java with Netty, hopefully the server-side problem that required this workaround of concatenating root CA and intermediate CA into one file is no longer required on the Beats side. Probably @ph has details on this.

Related: https://github.com/elastic/beats/issues/1494

andrewkroh: @dede I think the new sentence should be moved down to the `certificate_authorities` bullet. The "NOTE" can remain as it is.

PR #1973, opened 2016-07-06
Log connection attempts

tsg:
Transform a debug in Info for the loadbalancing mode, to make sure
we have at least an Info on any connection failure.

Part of #1931.

### DIFF
--- a/libbeat/outputs/mode/lb/async_worker.go
+++ b/libbeat/outputs/mode/lb/async_worker.go
@@ -82,17 +82,18 @@ func (w *asyncWorker) run() {
 
 func (w *asyncWorker) connect() bool {
 	for {
-		debugf("try to (re-)connect client")
 		err := w.client.Connect(w.ctx.timeout)
-		if !w.backoff.WaitOnError(err) {
-			return true
-		}
-
 		if err == nil {
+			w.backoff.Reset()
 			return false
 		}
 
-		debugf("connect failed with: %v", err)
+		logp.Info("Connect failed with: %v", err)
+
+		cont := w.backoff.Wait()
+		if !cont {
+			return true
+		}
 	}
 }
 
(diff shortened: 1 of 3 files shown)


tsg:
@urso can you check please if we need to do this in any other place? I covered and tested single mode + lb sync & async.

ruflin:
LGTM

urso:
this PR should capture redis, kafka, elasticsearch and logstash. File output already prints error or critical (if guaranteed flag is set) and console output logs a critical if publish fails.

libbeat/outputs/mode/lb/async_worker.go
```diff
@@ -92,7 +92,7 @@ func (w *asyncWorker) connect() bool {
 			return false
 		}
 
-		debugf("connect failed with: %v", err)
+		logp.Info("Connect failed with: %v", err)
```
urso: this will print the info message after waiting for Sleep on backoff handler. Alternatively:

```
for {
    err := w.client.Connect(w.ctx.timeout)
    if err == nil {
        w.backoff.Reset()
        return false
    }
    
    logp.Info("Connect failed with: %v", err)

    stopped := w.backoff.Wait()
    if stopped {
        return true
    }
}
```

libbeat/outputs/mode/lb/sync_worker.go
```diff
@@ -92,7 +92,7 @@ func (w *syncWorker) connect() bool {
 			return false
 		}
 
-		debugf("connect failed with: %v", err)
+		logp.Info("Connect failed with: %v", err)
```
urso: same as for async worker. message printed after waiting for current backoff.

tsg:
@urso Updated with your sample code. Can you check that I got it right, please?

libbeat/outputs/mode/lb/async_worker.go
```diff
@@ -82,17 +82,18 @@ func (w *asyncWorker) run() {
...
-
 		if err == nil {
+			w.backoff.Reset()
 			return false
 		}
 
-		debugf("connect failed with: %v", err)
+		logp.Info("Connect failed with: %v", err)
+
+		stopped := w.backoff.Wait()
```
urso: I mixed up the return value here :(

should be `cont := w.backoff.Wait(); if !cont { return true }`

Issue #1913, opened 2016-06-27
Filebeat Cleanup


ruflin:
This issue is intended as meta issue to track several filebeat cleanups
# Harvester separation

Log and stdin harvester should be separated and made independent potentially with separate readers.
- [x] Remove Continuable() interface as readers are separated (will no happen as not needed at the moment)
# Harvester
- [x] Unify error message and make them more consistent. For example `err_inactive`.
- [ ] Use `common.Backoff` in file harvester.
- [x] Improve handling of encoding https://github.com/elastic/beats/pull/1975#discussion_r70256218
- [x] Merge shared variables between harvester and state (path/Source, offset) https://github.com/elastic/beats/pull/2020
# Prospectors
- [x] Prospectors must be able to send state updates to the registrar directly (no need to initialised harvester) https://github.com/elastic/beats/pull/1978
- [x] `startHarvester` returns an error which is not checked by all implementations #2240
# Testing
- [x] Clean up inconsistent variable naming in `filebeat.yml.j2`. All should be with _ and not cameCase. https://github.com/elastic/beats/pull/1975
# General
- [x] Introduce expvar to track open files https://github.com/elastic/beats/pull/2020
- [x] Add additional expvar to track states
# Docs
- [x] Write "How Filebeat works" documentation guide. Explain crawler, prospector, harvestor https://github.com/elastic/beats/pull/1963

urso:
Use `common.Backoff` in file harvester.

ruflin:
All changes implement except for `common.Backoff`. @urso I suggest if we sitll want to have a `common.Backoff` we should open a separate issue?

PR #1975, opened 2016-07-06
Cleanup Filebeat code

ruflin:
Some simplifications / cleanup to the filebeat code.

Meta issue: https://github.com/elastic/beats/issues/1913

### DIFF
--- a/filebeat/crawler/crawler.go
+++ b/filebeat/crawler/crawler.go
@@ -11,23 +11,11 @@ import (
 	"github.com/elastic/beats/libbeat/logp"
 )
 
-/*
- The hierarchy for the crawler objects is explained as following
-
- Crawler: Filebeat has one crawler. The crawler is the single point of control
- 	and stores the state. The state is written through the registrar
- Prospector: For every FileConfig the crawler starts a prospector
- Harvester: For every file found inside the FileConfig, the Prospector starts a Harvester
- 		The harvester send their events to the spooler
- 		The spooler sends the event to the publisher
- 		The publisher writes the state down with the registrar
-*/
-
 type Crawler struct {
 	prospectors       []*prospector.Prospector
-	wg                sync.WaitGroup
-	spooler           *spooler.Spooler
 	prospectorConfigs []*common.Config
+	spooler           *spooler.Spooler
+	wg                sync.WaitGroup
 }
 
 func New(spooler *spooler.Spooler, prospectorConfigs []*common.Config) (*Crawler, error) {
@@ -58,7 +46,6 @@ func (c *Crawler) Start(states file.States) error {
 
 	logp.Info("Loading Prospectors completed. Number of prospectors: %v", len(c.prospectors))
 
-	c.wg = sync.WaitGroup{}
 	for i, p := range c.prospectors {
 		c.wg.Add(1)
 
--- a/filebeat/harvester/harvester.go
+++ b/filebeat/harvester/harvester.go
@@ -31,11 +31,12 @@ type Harvester struct {
 	offset             int64
 	state              file.State
 	prospectorChan     chan *input.FileEvent
-	encoding           encoding.EncodingFactory
 	file               source.FileSource /* the file being watched */
 	ExcludeLinesRegexp []*regexp.Regexp
 	IncludeLinesRegexp []*regexp.Regexp
 	done               chan struct{}
+	encodingFactory    encoding.EncodingFactory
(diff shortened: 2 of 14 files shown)


filebeat/harvester/reader/log.go
```diff
@@ -28,13 +28,13 @@ type logFileReader struct {
...
-	CloseOlder         time.Duration
-	BackoffDuration    time.Duration
-	MaxBackoffDuration time.Duration
-	BackoffFactor      int
-	CloseRenamed       bool
-	CloseRemoved       bool
-	CloseEOF           bool
+	Backoff       time.Duration
+	MaxBackoff    time.Duration
+	BackoffFactor int
```
urso: I think I like BackoffDuration better, as we've got 3 configs with Backoff here. Alternatively:

```
type BackoffConfig struct {
    Duration time.Duration
    Max time.Duration
    Factor float32
}
```

filebeat/harvester/harvester.go
```diff
@@ -27,17 +26,16 @@ import (
...
-	IncludeLinesRegexp []*regexp.Regexp
-	done               chan struct{}
+	Path            string /* the file path to harvest */
+	Config          harvesterConfig
+	offset          int64
+	State           file.State
+	stateMutex      sync.Mutex
+	SpoolerChan     chan *input.FileEvent
+	encodingFactory encoding.EncodingFactory
+	encoding        encoding.Encoding
```
urso: why promote the encoding itself to Harvester structure?

ruflin: Because it is needed in the openFile method later on. Not sure if we should move it there?

filebeat/harvester/reader/log.go
ruflin: The naming change is to make reader configs consitent with the harvester config names.

+1 on the suggested struct, but then we should have it in libbeat.

filebeat/harvester/harvester.go
urso: I think it makes sense to just pass the encodingFactory and establish an Encoding instance when required.

filebeat/harvester/log.go
```diff
@@ -291,40 +268,49 @@ func (h *Harvester) close() {
...
+
+func (h *Harvester) newLineProcessor() (processor.LineProcessor, error) {
+
+	readerConfig := h.newLogFileReaderConfig()
+
 	var p processor.LineProcessor
 	var err error
 
-	fileReader, err := reader.NewLogFileReader(in, readerConfig, done)
+	fileReader, err := reader.NewLogFileReader(h.file, readerConfig, h.done)
```
urso: having h.file available, we can initialize the encoding before this line. When initializing the encoding, the file pointer still should be at offset 0 though (in case of encoding uses some BOM). But why not set file offset in newLineProcessor?

filebeat/harvester/reader/log.go
urso: This is what we currently have in libbeat: https://github.com/elastic/beats/blob/master/libbeat/common/backoff.go

ruflin: It doesn't have the factor as it is just uses `*2`. We should do a more generic cleanup across all beats in an other PR.

filebeat/harvester/log.go
ruflin: It could be that we continue reading so the offset is not at 0. So that means if we init the encoding too late, we have a problem here?

About setting the offset: I don't think that is a processors tasks, but the lines are still blury THB.

filebeat/crawler/crawler.go
```diff
@@ -39,6 +27,7 @@ func New(spooler *spooler.Spooler, prospectorConfigs []*common.Config) (*Crawler
 	return &Crawler{
 		spooler:           spooler,
 		prospectorConfigs: prospectorConfigs,
+		wg:                sync.WaitGroup{},
```
andrewkroh: This line isn't necessary.

ruflin: removed

filebeat/harvester/reader/log.go
ruflin: @urso this is now also part of this meta issue here: https://github.com/elastic/beats/issues/1913

filebeat/harvester/log.go
ruflin: I added a comment about encoding handling here: https://github.com/elastic/beats/issues/1913

PR #1985, opened 2016-07-07
The '-c' flag can be used multiple times

urso:
Change '-c' flag to load and merge all config files given on command line via
'-c' flag.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -8,7 +8,7 @@
 // Template, add newest changes here
 
 === Beats version HEAD
-https://github.com/elastic/beats/compare/v5.0.0-alpha3...master[Check the HEAD diff]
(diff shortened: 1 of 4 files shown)


urso:
blocked by #1984

tsg:
LGTM. Does the order matter. i.e. options from the second file overwrite the options from the first file? I'm curious what use case you have in mind with this.

urso:
Yes, order does matter. The second file will overwrite options from first file.
I'd like to have this mostly for manual testing releases overwriting configs (e.g. libbeat based) without having to change every single config file. Instead of modifying the configs every so often when testing, this allows me to prepare configurations with outputs and special hosts I want to test with.

This might also make sense in production when running multiple beats, as output configurations can be shared in one config file only.

This PR + #1986 add some additional flexibility for overwriting configs.

PR #1986, opened 2016-07-07
Add '-E' CLI flag

urso:
Add '-D' CLI flag for overwriting single config options via CLI. For example:

`-D output.console.pretty=true` enables console output with pretty printing
enabled.

### DIFF
--- a/libbeat/cfgfile/cfgfile.go
+++ b/libbeat/cfgfile/cfgfile.go
@@ -15,6 +15,7 @@ var (
 	// when this variable is created. See ChangeDefaultCfgfileFlag which should
 	// be called prior to flags.Parse().
 	configfiles = flagArgList("c", "beat.yml", "Configuration file")
+	overwrites  = common.NewFlagConfig(nil, nil, "E", "Configuration overwrite")
 	testConfig  = flag.Bool("configtest", false, "Test configuration and exit.")
 )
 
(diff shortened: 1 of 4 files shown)


urso:
blocked by #1985

urso:
config values overwritten from command line are remembered by config system. e.g. error message from new system test:

`error unpacking config data: required 'object', but found 'string' in field 'output.console' (source:'command line flag')`

tsg:
Niiiiiice! 

I'm not sure about the best flag to use for this. Elasticsearch replaced `-D` with `-E` so they don't get confused with the JVM options. I guess `-E` is from elasticsearch so that doesn't work. Perhaps `-C` would be an option from "config". In any case, it's hard to be consistent here :(.

@elastic/beats any other ideas?

urso:
One of:
```
- D = define
- E = environment
- C = config
- S = set
```

Don't like `-C`, as it might get mistaken for `-c` by accident.

tsg:
Then I'm +1 for `-E` to be consistent with Elasticsearch. We can pretend it's E for "environment" :)

tsg:
@urso there's a test failure.

libbeat/cfgfile/cfgfile.go
```diff
@@ -15,6 +15,7 @@ var (
 	// when this variable is created. See ChangeDefaultCfgfileFlag which should
 	// be called prior to flags.Parse().
 	configfiles = flagArgList("c", "beat.yml", "Configuration file")
+	overwrites  = common.NewFlagConfig(nil, nil, "E", "Configuration overwrite")
```
ruflin: Can the `-E` flag be used multiple times to overwrite multiple variables?

ruflin:
LGTM. Left a minor comment.

libbeat/cfgfile/cfgfile.go
urso: Yes. Functionality provided by go-ucfg. Yes I need to document ucfg capabilities ;)

libbeat/tests/system/test_base.py
```diff
@@ -39,6 +39,22 @@ def test_invalid_config(self):
...
+        """
+        Checks CLI overwrite actually overwrites some config variable by
+        writing an invalid value.
+        """
+
+        self.render_config_template(
+            console={"pretty": "false"}
+        )
+
+        exit_code = self.run_beat(
```
urso: what exactly would you look for? Just relying on exit_code doesn't tell me if configuration was overwritten correctly.

Would love to introduce another test showing `-E` did apply config correctly. Maybe introduce some special config setting for printing a custom message (like `mockbeat.echo`) ?

ruflin: Yes, but it would confirm that the config you set worked without the -E flag.

ruflin:
LGTM. Waiting for green.

PR #1987, opened 2016-07-07
Add enable-setting to all output plugins

urso:


monicasarbu:
Shall we also update the`full` configuration files of the Beats to contain `enable` even if it's not needed?

urso:
Not sure. For me the enable-config mostly makes sense in conjunction with #1985 and #1986, as this will allow to disable modules e.g. from command line.

But having it named `full`-config, maybe it makes sense to add enable.

tsg:
I think it should be in the full configs, as we said the full configs contain all the non-deprecated options and they can be used as a reference.

monicasarbu:
As you already added the `enable` option, I would say to use it in the `full` -config as it's easier to just change the value of `enable` instead of commenting in/out an output.

monicasarbu:
+1 enable/disable one output in the command line.

urso:
makes sense. will add `enable` to `full` config.

andrewkroh:
LGTM

ruflin:
@urso there is one issue with this PR. Metricbeat uses `enabled`, you used `enable`. We should be consistent here.

PR #1992, opened 2016-07-08
Add filebeat JSON config `plain` for don't add beats related meta data to JSON

bohrqiu:
Some scenes do not need to modify the json data


tsg:
Thanks for the PR @bohrqiu. 

Currently the `@timestamp` and `type` fields are strictly required, otherwise some of the outputs can fail. There are also fields that are added later in the code (e.g. `beat.hostname`).

Perhaps it would be better if you would a processors (available in master) configuration to drop the fields that you don't need. With master it would be something like:

```
processors:
  - drop_fields:
      fields: ["source", "offset", "input_type", "beat"]
```

Would this be an option for you?

tsg:
Closing this one for now. @bohrqiu please reopen if you still have interest in this.

PR #1993, opened 2016-07-08
Load template by default, depending on the version

### LINKED ISSUE #1740: Load the template depending on ES version
It was requested [here](https://discuss.elastic.co/t/suggestion-libbeat-to-select-between-xxx-template-json-and-xxx-template-es2x-json/50887): 

> May I suggest you to implement in libbeat a way to push the correct ES template file depending on the ES server.
> I believe it should not be difficult to query the ES version and select xxx.template,json or xxx.template-es2x.json instead of having to configure it in the yaml file.
>

tsg:
This queries Elasticsearch for the version and choses between the
default template file (specified by the `template.path` option) and
the ES 2x version of the template file (specified by the new
`template.path_es2x` option).

There's also a change in the defaults. With all the template
options commented out, the code now loads the template. Previously,
the template loading was the default via the sample configs only.

I also did a small change in the output plugins API, now they receive
the beatname in a clean way.

Fixes #1740.

### DIFF
--- a/filebeat/filebeat.full.yml
+++ b/filebeat/filebeat.full.yml
@@ -357,16 +357,28 @@ output.elasticsearch:
 
   # A template is used to set the mapping in Elasticsearch
   # By default template loading is enabled and the template is loaded.
-  # These settings can be adjusted to load your own template or overwrite existing ones
+  # These settings can be adjusted to load your own template or overwrite existing ones.
+
+  # Set to false to disable template loading.
+  #template.enabled: true
 
   # Template name. By default the template name is filebeat.
-  template.name: "filebeat"
+  #template.name: "filebeat"
 
   # Path to template file
-  template.path: "filebeat.template.json"
+  #template.path: "filebeat.template.json"
 
   # Overwrite existing template
-  template.overwrite: false
+  #template.overwrite: false
+
+  # If set to true, filebeat checks the Elasticsearch version at connect time, and if it
+  # is 2.x, it loads the file specified by the template.versions.2x.path setting. The 
+  # default is true.
+  #template.versions.2x.enabled: true
+
+  # Path to the Elasticsearch 2.x version of the template file.
+  #template.versions.2x.path: "filebeat.template-es2x.json"
+
 
   # TLS configuration. By default is off.
   # List of root certificates for HTTPS server verifications
--- a/filebeat/filebeat.yml
+++ b/filebeat/filebeat.yml
@@ -82,15 +82,6 @@ output.elasticsearch:
   # Array of hosts to connect to.
   hosts: ["localhost:9200"]
 
-  # Template name. By default the template name is filebeat.
-  template.name: "filebeat"
-
-  # Path to template file
-  template.path: "filebeat.template.json"
-
-  # Overwrite existing template
-  template.overwrite: false
-
 #----------------------------- Logstash output --------------------------------
 #output.logstash:
   # The Logstash hosts
--- a/libbeat/_meta/config.full.yml
+++ b/libbeat/_meta/config.full.yml
@@ -139,16 +139,28 @@ output.elasticsearch:
 
   # A template is used to set the mapping in Elasticsearch
   # By default template loading is enabled and the template is loaded.
-  # These settings can be adjusted to load your own template or overwrite existing ones
+  # These settings can be adjusted to load your own template or overwrite existing ones.
+
+  # Set to false to disable template loading.
+  #template.enabled: true
 
   # Template name. By default the template name is beatname.
-  template.name: "beatname"
+  #template.name: "beatname"
 
   # Path to template file
-  template.path: "beatname.template.json"
+  #template.path: "beatname.template.json"
 
   # Overwrite existing template
-  template.overwrite: false
+  #template.overwrite: false
+
+  # If set to true, beatname checks the Elasticsearch version at connect time, and if it
+  # is 2.x, it loads the file specified by the template.versions.2x.path setting. The 
+  # default is true.
+  #template.versions.2x.enabled: true
+
+  # Path to the Elasticsearch 2.x version of the template file.
+  #template.versions.2x.path: "beatname.template-es2x.json"
+
 
   # TLS configuration. By default is off.
   # List of root certificates for HTTPS server verifications
--- a/libbeat/_meta/config.yml
+++ b/libbeat/_meta/config.yml
@@ -24,15 +24,6 @@ output.elasticsearch:
   # Array of hosts to connect to.
   hosts: ["localhost:9200"]
 
-  # Template name. By default the template name is beatname.
-  template.name: "beatname"
-
-  # Path to template file
-  template.path: "beatname.template.json"
-
-  # Overwrite existing template
-  template.overwrite: false
-
 #----------------------------- Logstash output --------------------------------
 #output.logstash:
   # The Logstash hosts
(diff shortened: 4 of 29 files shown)


tsg:
Not complete yet, but I'd like your feedback on the new config options @elastic/beats.

libbeat/outputs/elasticsearch/client.go
```diff
@@ -510,18 +512,35 @@ func (conn *Connection) Connect(timeout time.Duration) error {
...
+		return false, "", fmt.Errorf("Non 2xx response code: %d", status)
+	}
+
+	var response struct {
+		Version struct {
+			Number string
+		}
+	}
+
+	err = json.Unmarshal(body, &response)
```
urso: make sure body is fully consumed and closed correctly. Otherwise underlying connection can not be re-used (go http module uses keep-alive in order to reuse TCP connections) + need to time out underlying connection in order to close socket.

libbeat/outputs/fileout/file.go
```diff
@@ -39,7 +40,7 @@ func (out *fileOutput) init(config config) error {
 	out.rotator.Path = config.Path
 	out.rotator.Name = config.Filename
 	if out.rotator.Name == "" {
-		out.rotator.Name = config.Index
+		out.rotator.Name = out.beatName
```
urso: this means we can remove 'Index' from file config?

libbeat/outputs/elasticsearch/config.go
```diff
@@ -24,9 +24,12 @@ type elasticsearchConfig struct {
...
 type Template struct {
-	Name      string `config:"name"`
-	Path      string `config:"path"`
-	Overwrite bool   `config:"overwrite"`
+	Enabled           bool   `config:"enabled"`
+	Name              string `config:"name"`
+	IgnoreES2xVersion bool   `config:"ignore_es2x_version"`
+	Path              string `config:"path"`
+	PathES2x          string `config:"path_es2x"`
+	Overwrite         bool   `config:"overwrite"`
```
urso: how about versioned configs like.

In config file:

```
template.versions:
    - version: '2.*' # match all ES 2.x.x versions
      path: ...

template.name: ...
template.path: ...
template.overwrite: ...
```

If any of `versions.<idx>.version` matches the version we discovered during ping, `version.<idx>.path` will be used, otherwise default template file via `template.path` will be used.

libbeat/outputs/elasticsearch/client.go
tsg: This should be the case already because the body is coming from a call to `execHTTPRequest` which does `ReadAll` and `Close`.

libbeat/outputs/fileout/file.go
tsg: Currently `index` is treated specially and uses the hack [here](https://github.com/elastic/beats/blob/master/libbeat/outputs/outputs.go#L79). I think we could remove the hack and use the `beatName` in each output as needed. Is that what you mean?

libbeat/outputs/elasticsearch/config.go
tsg: That seems more flexible and future proof, but what I struggle with in that version is that the we need the 2.x version to be "on" by default. So the default for `template.versions` would be one entry with the 2.x path. That all seems a bit too magical, I have to think more about it.

libbeat/_meta/config.full.yml
```diff
@@ -139,13 +139,22 @@ output.elasticsearch:
   # These settings can be adjusted to load your own template or overwrite existing ones
 
   # Template name. By default the template name is beatname.
-  template.name: "beatname"
+  #template.name: "beatname"
```
ruflin: as it now also loads when commented how, we should have an option `template.enabled: false` to make it possible to not load the template

libbeat/outputs/elasticsearch/config.go
```diff
@@ -24,9 +24,12 @@ type elasticsearchConfig struct {
 }
 
 type Template struct {
-	Name      string `config:"name"`
-	Path      string `config:"path"`
-	Overwrite bool   `config:"overwrite"`
+	Enabled           bool   `config:"enabled"`
```
ruflin: @tsg Just spotted this now :-)
@urso `enabled` is also here ;-)

ruflin: We should have this in the default config.

libbeat/outputs/elasticsearch/config.go
tsg: I went for something in between:

```
  # If set to true, filebeat checks the Elasticsearch version at connect time, and if it
  # is 2.x, it loads the file specified by the template.versions.2x.path setting. The 
  # default is true.
  template.versions.2x.enabled: true

  # Path to the Elasticsearch 2.x version of the template file.
  template.versions.2x.path: "filebeat.template-es2x.json"
```

Not as flexible as your version, but pretty easily extensible in the future. Also, I think the flexibility is not needed, as you generally don't want a Beat to talk to multiple ES versions. This is just for making the getting starting experience easier.

tsg:
Comments mostly addressed, let's see if tests pass.

libbeat/outputs/fileout/file.go
tsg: Probably best to do this refactoring in a another PR, not to lose the focus on this one.

tsg:
jenkins, retest it

tsg:
jenkins, retest it

filebeat/filebeat.full.yml
```diff
@@ -357,16 +357,28 @@ output.elasticsearch:
 
   # A template is used to set the mapping in Elasticsearch
   # By default template loading is enabled and the template is loaded.
-  # These settings can be adjusted to load your own template or overwrite existing ones
+  # These settings can be adjusted to load your own template or overwrite existing ones.
+
+  # Set to false to disable template loading.
+  # template.enabled: true
```
ruflin: Remove the space after #

filebeat/filebeat.yml
```diff
@@ -82,14 +82,14 @@ output.elasticsearch:
   # Array of hosts to connect to.
   hosts: ["localhost:9200"]
 
-  # Template name. By default the template name is filebeat.
-  template.name: "filebeat"
+  # Load the Elasticsearch template.
+  # template.enabled: true
```
ruflin: As it is now loaded by default, I'm not even sure if we still needed in the short config?

libbeat/_meta/config.full.yml
```diff
@@ -139,16 +139,28 @@ output.elasticsearch:
 
   # A template is used to set the mapping in Elasticsearch
   # By default template loading is enabled and the template is loaded.
-  # These settings can be adjusted to load your own template or overwrite existing ones
+  # These settings can be adjusted to load your own template or overwrite existing ones.
+
+  # Set to false to disable template loading.
+  # template.enabled: true
```
ruflin: space

libbeat/_meta/config.yml
```diff
@@ -24,14 +24,14 @@ output.elasticsearch:
   # Array of hosts to connect to.
   hosts: ["localhost:9200"]
 
-  # Template name. By default the template name is beatname.
-  template.name: "beatname"
+  # Load the Elasticsearch template.
+  # template.enabled: true
```
ruflin: space

ruflin:
LGTM. Only two minor comments.

ruflin:
LGTM. Waiting for green, but will ignore appveyor.

PR #1996, opened 2016-07-08
Marked ES as disconnected only of 5xx errors

tsg:
Old code was marking the connection as disconnected on any response code
larger than 300. This caused issues for the template loading code, that
expects a 404 when testing if the template is loaded.


tsg:
I opened this in a separate PR as a proposal and so we carefully check for side effects of this change.

tsg:
Replaced with #1997.

PR #2002, opened 2016-07-09
Replace optional FlagsHandler interface with callback list

urso:


ruflin:
LGTM. Not sure if any other beats made use of this. 

The flags handler is also documented in the newbeat.asciidoc. This should be removed / changed.

urso:
@ruflin newbeat doc will be adjusted once all changes are in place.

dedemorton:
@ruflin I'm removing mention of the flags handler from newbeat.asciidoc for now. Let me know what, if anything, I should say in that doc about adding custom flags.

ruflin:
@dedemorton The addition of custom flags should be very rare and config options should be used. So no need to add docs about it.

PR #2003, opened 2016-07-09
Simplify Beater interface

urso:
pass Beater constructor to beat.Run + simplify Beater interface to only require
Run and Stop methods.

### DIFF
--- a/filebeat/config/config.go
+++ b/filebeat/config/config.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"errors"
 	"log"
 	"os"
 	"path/filepath"
@@ -98,13 +99,13 @@ func mergeConfigFiles(configFiles []string, config *Config) error {
 }
 
 // Fetches and merges all config files given by configDir. All are put into one config object
-func (config *Config) FetchConfigs() {
+func (config *Config) FetchConfigs() error {
 
 	configDir := config.Filebeat.ConfigDir
 
 	// If option not set, do nothing
 	if configDir == "" {
-		return
+		return nil
 	}
 
 	// If configDir is relative, consider it relative to the config path
@@ -117,14 +118,20 @@ func (config *Config) FetchConfigs() {
 
 	if err != nil {
 		log.Fatal("Could not use config_dir of: ", configDir, err)
+		return err
 	}
 
 	err = mergeConfigFiles(configFiles, config)
 	if err != nil {
 		log.Fatal("Error merging config files: ", err)
+		return err
 	}
 
 	if len(config.Filebeat.Prospectors) == 0 {
-		log.Fatalf("No paths given. What files do you want me to watch?")
+		err := errors.New("No paths given. What files do you want me to watch?")
+		log.Fatalf("%v", err)
+		return err
 	}
+
+	return nil
 }
--- a/filebeat/publish/publish.go
+++ b/filebeat/publish/publish.go
@@ -127,8 +127,8 @@ func (p *syncLogPublisher) Start() {
 }
 
 func (p *syncLogPublisher) Stop() {
-	close(p.done)
 	p.client.Close()
+	close(p.done)
 	p.wg.Wait()
 }
 
@@ -182,8 +182,8 @@ func (p *asyncLogPublisher) Start() {
 }
 
 func (p *asyncLogPublisher) Stop() {
-	close(p.done)
 	p.client.Close()
+	close(p.done)
 	p.wg.Wait()
 }
 
--- a/generate/beat/{{cookiecutter.beat}}/beater/{{cookiecutter.beat}}.go
+++ b/generate/beat/{{cookiecutter.beat}}/beater/{{cookiecutter.beat}}.go
@@ -13,42 +13,29 @@ import (
 )
 
 type {{cookiecutter.beat|capitalize}} struct {
-	config config.Config
-	done   chan struct{}
-	client publisher.Client
+	done       chan struct{}
+	config     config.Config
+	client     publisher.Client
 }
 
 // Creates beater
-func New() *{{cookiecutter.beat|capitalize}} {
-	return &{{cookiecutter.beat|capitalize}}{
+func New(b *beat.Beat, cfg *common.Config) (beat.Beater, error) {
+	config := config.DefaultConfig
+	if err := cfg.Unpack(&config); err != nil {
+		return nil, fmt.Errorf("Error reading config file: %v", err)
+	}
+
+	bt := &{{cookiecutter.beat|capitalize}}{
 		done: make(chan struct{}),
+		config: config,
 	}
-}
-
-/// *** Beater interface methods ***///
-
-func (bt *{{cookiecutter.beat|capitalize}}) Config(b *beat.Beat) error {
-
-	bt.config = config.DefaultConfig
-
-	// Load beater config
-	err := b.RawConfig.Unpack(&bt.config)
-	if err != nil {
-		return fmt.Errorf("Error reading config file: %v", err)
-	}
-
-	return nil
-}
-
-func (bt *{{cookiecutter.beat|capitalize}}) Setup(b *beat.Beat) error {
-
-	bt.client = b.Publisher.Connect()
-	return nil
+	return bt, nil
 }
 
 func (bt *{{cookiecutter.beat|capitalize}}) Run(b *beat.Beat) error {
 	logp.Info("{{cookiecutter.beat}} is running! Hit CTRL-C to stop it.")
 
+	bt.client = b.Publisher.Connect()
 	ticker := time.NewTicker(bt.config.{{cookiecutter.beat|capitalize}}.Period)
 	counter := 1
 	for {
@@ -69,10 +56,7 @@ func (bt *{{cookiecutter.beat|capitalize}}) Run(b *beat.Beat) error {
 	}
 }
 
-func (bt *{{cookiecutter.beat|capitalize}}) Cleanup(b *beat.Beat) error {
-	return nil
-}
-
 func (bt *{{cookiecutter.beat|capitalize}}) Stop() {
+	bt.client.Close()
 	close(bt.done)
 }
--- a/libbeat/beat/beat.go
+++ b/libbeat/beat/beat.go
@@ -4,20 +4,20 @@ It provides the standard mechanism for launching a Beat. It manages
 configuration, logging, and publisher initialization and registers a signal
 handler to gracefully stop the process.
 
-Each Beat implementation must implement the Beater interface and may optionally
-implement the FlagsHandler interface. See the Beater interface documentation for
-more details.
+Each Beat implementation must implement the `Beater` interface and a `Creator`
+to create and initialize the Beater instance. See the `Beater` interface and `Creator`
+documentation for more details.
 
 To use this package, create a simple main that invokes the Run() function.
 
   func main() {
-  	if err := beat.Run("mybeat", myVersion, beater.New()); err != nil {
+  	if err := beat.Run("mybeat", myVersion, beater.New); err != nil {
   		os.Exit(1)
   	}
   }
 
 In the example above, the beater package contains the implementation of the
-Beater interface and the New() method returns a new instance of Beater. The
+Beater interface and the New method returns a new instance of Beater. The
 Beater implementation is placed into its own package so that it can be reused
 or combined with other Beats.
 
--- a/libbeat/beat/beat.go
+++ b/libbeat/beat/beat.go
@@ -164,7 +164,10 @@ func (b *Beat) launch(bt Creator) error {
 	if err != nil {
 		return fmt.Errorf("error initializing publisher: %v", err)
 	}
-	defer publisher.Stop()
+
+	// TODO: some beats race on shutdown with publisher.Stop -> do not call Stop yet,
+	//       but refine publisher to disconnect clients on stop automatically
+	// defer publisher.Stop()
 
 	b.Publisher = publisher
 	beater, err := bt(b, b.RawConfig)
--- a/libbeat/beat/beat.go
+++ b/libbeat/beat/beat.go
@@ -134,6 +134,10 @@ func Run(name, version string, bt Creator) error {
 
 // newBeat creates a new beat instance
 func newBeat(name, version string) *Beat {
+	if version == "" {
+		version = defaultBeatVersion
+	}
+
 	return &Beat{
 		Name:    name,
 		Version: version,
--- a/libbeat/beat/beat.go
+++ b/libbeat/beat/beat.go
@@ -52,40 +52,30 @@ import (
 	"github.com/satori/go.uuid"
 )
 
-var (
-	printVersion = flag.Bool("version", false, "Print the version and exit")
-)
-
-var debugf = logp.MakeDebug("beat")
-
-// GracefulExit is an error that signals to exit with a code of 0.
-var GracefulExit = errors.New("graceful exit")
-
-// Beater is the interface that must be implemented every Beat. The full
-// lifecycle of a Beat instance is managed through this interface.
+// Beater is the interface that must be implemented by every Beat. A Beater
+// provides the main Run-loop and a Stop method to break the Run-loop.
+// Instantiation and Configuration is normally provided by a Beat-`Creator`.
 //
-// Life-cycle of Beater
-//
-// The four operational methods are always invoked serially in the following
-// order:
-//
-//   Config -> Setup -> Run -> Cleanup
+// Once the beat is fully configured, the Run() method is invoked. The
+// Run()-method implements the beat its run-loop. Once the Run()-method returns,
+// the beat shuts down.
 //
 // The Stop() method is invoked the first time (and only the first time) a
-// shutdown signal is received. The Stop() method is eligible to be invoked
-// at any point after Setup() completes (this ensures that the Beater
-// implementation is fully initialized before Stop() can be invoked).
-//
-// The Cleanup() method is guaranteed to be invoked upon shutdown iff the Beater
-// reaches the Setup stage. For example, if there is a failure in the
-// Config stage then Cleanup will not be invoked.
+// shutdown signal is received. The Stop()-method normally will stop the Run()-loop,
+// such that the beat can gracefully shutdown.
 type Beater interface {
-	Config(*Beat) error  // Read and validate configuration.
-	Setup(*Beat) error   // Initialize the Beat.
-	Run(*Beat) error     // The main event loop. This method should block until signalled to stop by an invocation of the Stop() method.
-	Cleanup(*Beat) error // Cleanup is invoked to perform any final clean-up prior to exiting.
-	Stop()               // Stop is invoked to signal that the Run method should finish its execution. It will be invoked at most once.
+	// The main event loop. This method should block until signalled to stop by an
+	// invocation of the Stop() method.
+	Run(b *Beat) error
+
+	// Stop is invoked to signal that the Run method should finish its execution.
+	// It will be invoked at most once.
+	Stop()
 }
+
+// Creator initializes and configures a new Beater instance used to execute
+// the beat its run-loop.
+type Creator func(*Beat, *common.Config) (Beater, error)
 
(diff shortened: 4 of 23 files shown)


urso:
@ruflin any tip how to fix the beat generator test? It seems the generator uses master branch, but not feature branch. Maybe fix generator after merging the PR?

filebeat/config/config.go
```diff
@@ -98,13 +99,13 @@ func mergeConfigFiles(configFiles []string, config *Config) error {
 }
 
 // Fetches and merges all config files given by configDir. All are put into one config object
-func (config *Config) FetchConfigs() {
+func (config *Config) FetchConfigs() error {
```
ruflin: Thanks for the cleanup here. This part could probably go into a separate PR to keep the changes in this one to a minimum.

filebeat/publish/publish.go
```diff
@@ -18,6 +18,7 @@ type LogPublisher interface {
 }
 
 type syncLogPublisher struct {
+	pub     publisher.Publisher
```
ruflin: Changes to the publisher could also go into a separate filebeat cleanup PR?

generate/beat/{{cookiecutter.beat}}/beater/{{cookiecutter.beat}}.go
```diff
@@ -13,54 +13,32 @@ import (
...
-
-	return nil
-}
+/// *** Beater interface methods ***///
 
 func (bt *{{cookiecutter.beat|capitalize}}) Run(b *beat.Beat) error {
 	logp.Info("{{cookiecutter.beat}} is running! Hit CTRL-C to stop it.")
 
-	ticker := time.NewTicker(bt.period)
+	bt.client = b.Publisher.Connect()
```
ruflin: An idea when I saw this: Should we require each beat to take care of the connec on Run or should we already to it before in libbeat?

generate/beat/{{cookiecutter.beat}}/config/config.go
```diff
@@ -3,10 +3,18 @@
...
 
+import "time"
+
 type Config struct {
 	{{cookiecutter.beat|capitalize}} {{cookiecutter.beat|capitalize}}Config
 }
 
 type {{cookiecutter.beat|capitalize}}Config struct {
-	Period string `config:"period"`
+	Period time.Duration `config:"period"`
```
ruflin: +1 on these changes. Could go into a cookiecutter cleanup PR?

libbeat/beat/beat.go
```diff
@@ -150,70 +139,113 @@ func init() {
...
-
-	return err
+	return handleFlags(b)
 }
 
 // config reads the configuration file from disk, parses the common options
 // defined in BeatConfig, initializes logging, and set GOMAXPROCS if defined
 // in the config. Lastly it invokes the Config method implemented by the beat.
-func (bc *instance) config() error {
+func (b *Beat) configure() error {
```
ruflin: I would prefer to keep the name config or if we change it, also comment must be changed.

libbeat/publisher/client.go
```diff
@@ -62,12 +62,12 @@ type Client interface {
...
 	canceler *op.Canceler
 
-	publisher           *Publisher
+	publisher           *BeatPublisher
 	beatMeta            common.MapStr        // Beat metadata that is added to all events.
 	globalEventMetadata common.EventMetadata // Fields and tags that are added to all events.
 }
 
-func newClient(pub *Publisher) *client {
+func newClient(pub *BeatPublisher) *client {
```
ruflin: Renaming *BeatPublisher could go into a libbeat cleanup PR.

packetbeat/beater/packetbeat.go
```diff
@@ -27,7 +27,7 @@ import (
 
 // Beater object. Contains all objects needed to run the beat
 type Packetbeat struct {
-	PbConfig    config.Config
+	Config      config.Config
```
ruflin: Packetbeat cleanup PR

packetbeat/publish/publish.go
```diff
@@ -19,9 +20,13 @@ type Flows interface {
 }
 
 type PacketbeatPublisher struct {
-	pub    *publisher.Publisher
+	pub    publisher.Publisher
```
ruflin: Packetbeat cleanup

winlogbeat/config/config.go
```diff
@@ -31,6 +31,14 @@ type Settings struct {
 	Raw        map[string]interface{} `config:",inline"`
 }
 
+var (
+	DefaultSettings = Settings{
```
ruflin: Windlogbeat cleanup PR

libbeat/beat/beat.go
```diff
@@ -52,15 +52,7 @@ import (
...
-	printVersion = flag.Bool("version", false, "Print the version and exit")
-)
-
-var debugf = logp.MakeDebug("beat")
-
-// GracefulExit is an error that signals to exit with a code of 0.
-var GracefulExit = errors.New("graceful exit")
-
+/*
 // Beater is the interface that must be implemented every Beat. The full
```
ruflin: The docs below must be udpated.

ruflin:
+1 on these changes. In general LGTM. 

For my understanding:
* Config(), Setup(), Cleanup() interface methods were removed
* Config is now part of New
* Setup is part or Run
* Cleanup is part of Run as defer or can be part of Stop

So migrating from the old to the new interfaces in a simplified way is to keep the methods like Setup but call them manually in the "right place"?

This changes make the beat interface very similar to the MetricSet interface which I quite like.

The docs must be updated with the new interfaces. As this change breaks all existing beats, we must definitively document it briefly in the changelog and potentially link to a "migration guide" in the dev guide.

I left a few comments similar to "cleanup PR x". The reasons for this is that most of the changes in this PR are actually not related to the interface changes but some cleanup which is nice. But as most of the beats developers have to migrate their beat I would prefer to have a PR with only the changes from the migration inside. This makes it much easier for people to migrate I hope. I'm aware of the separate commits but I prefer to point people to a PR with the changes. Could you pick some of the commits above and open a "Cleanup PR"?

About beat-generator: Unfortunately yes. I implemented a workaround for the metricset-generator which I should also apply to beat generator. Let me see if I can do that today.

ruflin:
@urso generator problem should be fixed by https://github.com/elastic/beats/pull/2006

generate/beat/{{cookiecutter.beat}}/beater/{{cookiecutter.beat}}.go
urso: touch one. But it's perfectly valid for a beat connect multiple times. So i'd keep it as is for now. We should only make it safe to connect to publisher pipeline in constructor, that is, libbeat launch method will kill/break connection on shutdown automatically (to be done in future PR).

libbeat/beat/beat.go
urso: I named it configure, not to be confused with 'Config' field in Beat.

urso:
> For my understanding:
- Config(), Setup(), Cleanup() interface methods were removed
- Config is now part of New
- Setup is part or Run
- Cleanup is part of Run as defer or can be part of Stop

Yes and no:
- Config(), Setup(), Cleanup() have been removed
- Config must be done in New
- Setup depends on beat. Implementation of Setup has been a little inconsistent among beat. Some beats doing too much work in Setup require to call setup in Run, others require setup to be run in New.
- Cleanup should be part of Run as defer. Like done for filebeat more recently. Stop is supposed to break the run loop.

libbeat/beat/beat.go
urso: yes, just commented out original interface. still needs to be cleaned up here. `in progress` flag ;)

filebeat/config/config.go
```diff
@@ -117,14 +118,20 @@ func (config *Config) FetchConfigs() {
 
 	if err != nil {
 		log.Fatal("Could not use config_dir of: ", configDir, err)
```
ruflin: Instead of logging the message here, we could just return the error. Can be done in a later PR.

filebeat/publish/publish.go
```diff
@@ -127,8 +127,8 @@ func (p *syncLogPublisher) Start() {
 }
 
 func (p *syncLogPublisher) Stop() {
-	close(p.done)
 	p.client.Close()
+	close(p.done)
```
ruflin: I'm curious if that could have led to some issues previously?

generate/beat/{{cookiecutter.beat}}/beater/{{cookiecutter.beat}}.go
```diff
@@ -13,42 +13,29 @@ import (
 )
 
 type {{cookiecutter.beat|capitalize}} struct {
-	config config.Config
-	done   chan struct{}
-	client publisher.Client
+	done       chan struct{}
```
ruflin: It could be nice to have the done channel also in under Beat.Done so simple beats could reuse it without having to create or close one themself.

generate/beat/{{cookiecutter.beat}}/beater/{{cookiecutter.beat}}.go
```diff
@@ -13,42 +13,29 @@ import (
...
+	config     config.Config
```
ruflin: BTW: I assume the build is going to complain here about `go fmt` ...

generate/beat/{{cookiecutter.beat}}/beater/{{cookiecutter.beat}}.go
```diff
@@ -13,42 +13,29 @@ import (
...
-
-func (bt *{{cookiecutter.beat|capitalize}}) Config(b *beat.Beat) error {
 
-	bt.config = config.DefaultConfig
-
-	// Load beater config
-	err := b.RawConfig.Unpack(&bt.config)
-	if err != nil {
-		return fmt.Errorf("Error reading config file: %v", err)
+	bt := &{{cookiecutter.beat|capitalize}}{
```
ruflin: Long term we could also go here with a generic approach as we did for MetricSet, calling it Beater here. But it is currently nice to have the name.

libbeat/beat/beat.go
```diff
@@ -52,53 +52,40 @@ import (
...
+	Run(b *Beat) error
+
+	// Stop is invoked to signal that the Run method should finish its execution.
+	// It will be invoked at most once.
+	Stop()
 }
 
+// Creator initializes and configures a new Beater instance used to execute
+// the beat its run-loop.
+type Creator func(*Beat, *common.Config) (Beater, error)
```
ruflin: Not 100% sure about the name.

libbeat/beat/beat.go
```diff
@@ -141,65 +124,114 @@ func init() {
...
+
+func (b *Beat) launch(bt Creator) error {
+	if err := b.handleFlags(); err != nil {
+		return err
+	}
+
+	svc.BeforeRun()
+	defer svc.Cleanup()
+
+	if err := b.configure(); err != nil {
```
ruflin: We loose some flexibilty here in that no special tasks can be execute by a Beater between configuring the beat and running the beat. This was previously possible. But as the refactoring just worked, I assume this possibility was not needed.

libbeat/beat/beat.go
```diff
@@ -141,65 +124,114 @@ func init() {
...
+		return fmt.Errorf("error initializing processors: %v", err)
+	}
+
+	debugf("Initializing output plugins")
+	publisher, err := publisher.New(b.Name, b.Config.Output, b.Config.Shipper, processors)
+	if err != nil {
+		return fmt.Errorf("error initializing publisher: %v", err)
+	}
+
+	// TODO: some beats race on shutdown with publisher.Stop -> do not call Stop yet,
```
ruflin: What happens to this comment?

filebeat/publish/publish.go
urso: not really. libbeat is not checking connections being closed yet (disabled due to races). There is still a chance events are published besides being shutdown.

generate/beat/{{cookiecutter.beat}}/beater/{{cookiecutter.beat}}.go
urso: hm, with libbeat providing Done and maybe even closing client on shutdown, we still need a stop method?

This might even be a bigger change in API.

for example, if I need to unblock anything once done closes I can write:

```
go func(){
    <-b.Done
    bt.UnblockWhatever()
}()
```

packetbeat/beater/packetbeat.go
```diff
@@ -135,16 +126,76 @@ func (pb *Packetbeat) Setup(b *beat.Beat) error {
...
+
+func (pb *Packetbeat) Run(b *beat.Beat) error {
+	defer func() {
+		if service.WithMemProfile() {
+			logp.Debug("main", "Waiting for streams and transactions to expire...")
+			time.Sleep(time.Duration(float64(protos.DefaultTransactionExpiration) * 1.2))
+			logp.Debug("main", "Streams and transactions should all be expired now.")
+		}
+
+		// TODO:
```
ruflin: What will happen to this one?

generate/beat/{{cookiecutter.beat}}/beater/{{cookiecutter.beat}}.go
urso: build is all greenish. consider adding a 'go fmt' check. Maybe it's some good practice to call `go fmt` as post-processing step to ensure all files generated are in good shape.

libbeat/beat/beat.go
urso: The full name is `beat.Creator` ;)

packetbeat/beater/packetbeat.go
```diff
@@ -202,64 +253,3 @@ func (pb *Packetbeat) makeWorkerFactory(filter string) sniffer.WorkerFactory {
...
-
-func (pb *Packetbeat) Cleanup(b *beat.Beat) error {
-
-	if service.WithMemProfile() {
-		logp.Debug("main", "Waiting for streams and transactions to expire...")
-		time.Sleep(time.Duration(float64(protos.DefaultTransactionExpiration) * 1.2))
-		logp.Debug("main", "Streams and transactions should all be expired now.")
-	}
-
-	// TODO:
```
ruflin: I see, this is only copy/paste :-)

libbeat/beat/beat.go
urso: I think it's good we're loosing this `flexibility`. Setup used to be somewhere in the middle between configure and run, inconsistently used by multiple beats as such.

packetbeat/publish/publish.go
```diff
@@ -99,8 +99,8 @@ func (t *PacketbeatPublisher) Start() {
 }
 
 func (t *PacketbeatPublisher) Stop() {
```
ruflin: ok, there is a separate packetbeat publisher

ruflin:
LGTM

libbeat/beat/beat.go
urso: looks ok to me. comment explains why `defer publisher.Stop()` is still commented out.

generate/beat/{{cookiecutter.beat}}/beater/{{cookiecutter.beat}}.go
ruflin: Good question. Not sure TBH. We should not only look at our beats but also at the community beats on how they use the interfaces.

generate/beat/{{cookiecutter.beat}}/beater/{{cookiecutter.beat}}.go
ruflin: not so easy here because of the template variables. I just remember fixing that manually, but if tests are green, all good. because it checks it as part of the build.

libbeat/beat/beat.go
ruflin: I currently don't have a better suggestion :-)

packetbeat/beater/packetbeat.go
urso: no idea yet. Just copy'n pasted from original location. Gotta check if it can be removed (another PR).

packetbeat/beater/packetbeat.go
```diff
@@ -135,16 +126,76 @@ func (pb *Packetbeat) Setup(b *beat.Beat) error {
...
+	}
+
 	return nil
 }
 
+// Called by the Beat stop function
+func (pb *Packetbeat) Stop() {
+	logp.Info("Packetbeat send stop signal")
+	pb.Sniff.Stop()
+	pb.Pub.Stop()
```
urso: this is a PacketbeatPublisher worker having 'Stop' method.

ruflin:
@urso I seems like the appveyor tests indicated that something changed in the exit behaviour of winlogbeat.

libbeat/beat/beat.go
```diff
@@ -52,53 +52,40 @@ import (
 	"github.com/satori/go.uuid"
 )
```
andrewkroh: ^ The package level comment is out-of-date.

libbeat/beat/beat.go
```diff
@@ -141,65 +124,114 @@ func init() {
...
+// instance.
+func Run(name, version string, bt Creator) error {
+	return handleError(newBeat(name, version).launch(bt))
+}
+
+// newBeat creates a new beat instance
+func newBeat(name, version string) *Beat {
+	return &Beat{
+		Name:    name,
+		Version: version,
```
andrewkroh: For our beats, `version` is empty, so I think this should check for empty and then set version to `defaultBeatVersion`. At startup our beats seem to be printing `Version:` but no number.

winlogbeat/beater/winlogbeat.go
```diff
@@ -114,8 +106,8 @@ func (eb *Winlogbeat) Setup(b *beat.Beat) error {
 
 	// Create the event logs. This will validate the event log specific
 	// configuration.
-	eb.eventLogs = make([]eventlog.EventLog, 0, len(eb.config.Winlogbeat.EventLogs))
-	for _, config := range eb.config.Winlogbeat.EventLogs {
+	eb.eventLogs = make([]eventlog.EventLog, 0, len(config.EventLogs))
+	for _, config := range config.EventLogs {
 		eventLog, err := eventlog.New(config)
```
andrewkroh: There is a lot of useful configuration validation performed as part of this `eventlog.New` method. It would be best if most of what is in this `Setup` method was invoked as part of `-configtest` (with the exception of the part where it launches an http server).

generate/beat/{{cookiecutter.beat}}/beater/{{cookiecutter.beat}}.go
urso: What I have in mind is running `go fmt` via cookiecutter post-generate hook: http://cookiecutter.readthedocs.io/en/latest/advanced/hooks.html

libbeat/beat/beat.go
urso: Creator, Constructor, Builder, Maker, Factory, Initializor.

libbeat/beat/beat.go
urso: Thanks. Updated.

libbeat/beat/beat.go
urso: Thanks. fixed.

winlogbeat/beater/winlogbeat.go
urso: moved eventlog.New into `init` method. Not sure about adding checkpoint to `init`. It checks the checkpoint file is available, can be parsed, can be written to + starts some worker. All of this required by `-configtest`?

andrewkroh: Good point about checkpoint. Validating the checkpoint file isn't strictly required to be done in `-configtest` so this doesn't need to be done in `New` since it starts the worker.

libbeat/beat/beat.go
ruflin: Lets keep it as it is :-)

ruflin:
LGTM

PR #2004, opened 2016-07-09
added PGID field to Process struct to use the recent gosigar addition

nickivanov:
This time for metricbeat.

### DIFF
--- a/metricbeat/docs/fields.asciidoc
+++ b/metricbeat/docs/fields.asciidoc
@@ -2630,6 +2630,14 @@ The process parent pid.
 
 
 [float]
+=== system.process.pgid
+
+type: integer
+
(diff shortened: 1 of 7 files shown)


ruflin:
@nickivanov Thanks for the contribution. This change requires to also update the `fields.yml` with the new field to make sure it exists in the template and docs. After updating `fields.yml` run `make collect` to update all files automatically.

karmi:
Hi @nickivanov, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/beats/pull/2004.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

Issue #2005, opened 2016-07-11
Cleanup and stabilize Beater-API


urso:
Simplify common API for creating a new beat:

- [x] FlagHandler registry instead of beat optionally implementing HandleFlags. FlagHandler will be called after parsing flags, but before reading any config files. #2002 
- [x] simplify Beater interface to consist of: Constructor(Initialization + Config), Run and Stop. #2003
- [x] turn publisher.Publisher into interface  #2009 
- [x] pass beat specific config section to beat on init #2014
- [x] update newbeat docs

PR #2006, opened 2016-07-11
Update beat generator to take local beats version for build

ruflin:


generate/beat/{{cookiecutter.beat}}/beater/{{cookiecutter.beat}}.go
```diff
@@ -41,26 +42,14 @@ func (bt *{{cookiecutter.beat|capitalize}}) Config(b *beat.Beat) error {
...
-	}
-
 	return nil
 }
 
 func (bt *{{cookiecutter.beat|capitalize}}) Run(b *beat.Beat) error {
 	logp.Info("{{cookiecutter.beat}} is running! Hit CTRL-C to stop it.")
 
-	ticker := time.NewTicker(bt.period)
+	ticker := time.NewTicker(bt.config.{{cookiecutter.beat|capitalize}}.Period)
```
ruflin: @urso I would be nice in the future that we don't need the beat name here to access own config variables. Not sure if we should have a `beat.Config` and `beater.Config` with the different variables? Would now be possible with ucfg I think.

urso: yeah, with ucfg we can separate and pass the beat config section to init method. Will add item to #2005

PR #2007, opened 2016-07-11
Validate wrong filebeat config combinations

ruflin:
* [x] Check ignore_older and close_older
* [x] Check close_removed and clean_removed combination
  * This will be indirectly solved by https://github.com/elastic/beats/pull/2075 as states can only be removed if reading is finished.
* [x] Check `clean_idle > ignore_older + scan_frequency`


ruflin:
@urso I have some implementation issue here I was hoping perhaps you could help. `close_removed` is part of the harvester config and the prospector does not know about it. `clean_removed` is part of the prospector config and the harvester does not know about it. The problem is now how to verify that both are set. It would be easy to just add `close_removed` also to the prospector config just for the validation. At the same time I think it is really nice that we have the two separated and would like to keep it this way. Any alternative idea?

PR #2008, opened 2016-07-11
[Proposal] Rename prospectors to inputs

ruflin:
This change would require to update all docs.

### DIFF
--- a/filebeat/beater/filebeat.go
+++ b/filebeat/beater/filebeat.go
@@ -40,6 +40,11 @@ func (fb *Filebeat) Config(b *beat.Beat) error {
 	// Check if optional config_dir is set to fetch additional prospector config files
 	fb.config.FetchConfigs()
 
+	if len(fb.config.Filebeat.Prospectors) > 0 {
+		fb.config.Filebeat.Inputs = fb.config.Filebeat.Prospectors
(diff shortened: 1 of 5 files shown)


filebeat/beater/filebeat.go
```diff
@@ -40,6 +40,11 @@ func (fb *Filebeat) Config(b *beat.Beat) error {
 	// Check if optional config_dir is set to fetch additional prospector config files
 	fb.config.FetchConfigs()
 
+	if len(fb.config.Filebeat.Prospectors) > 0 {
+		fb.config.Filebeat.Inputs = fb.config.Filebeat.Prospectors
```
urso: what if I have both, `inputs` and `prospectors`?

ruflin: Not allowed. In case we decide to proceed, I will add a check for that. In case both are set, and error is returned.

tsg:
Nice, changeset is small enough. But it's still a big change as the new terminology changes a bit the way we think about the way this works. So I'd say let's all sleep over it before pressing merge :-).

ruflin:
Closing this PR as there are too many filebeat changes in the pipeline before beta1 I want to focus on (#1600). I don't think we should rush any changes here as it is not just about renaming. It also affects the docs and knowledge of existing users.

PR #2010, opened 2016-07-12
added PGID field to Process struct to use the recent gosigar addition

nickivanov:
This time for metricbeat and updated field docs.

### DIFF
--- a/metricbeat/docs/fields.asciidoc
+++ b/metricbeat/docs/fields.asciidoc
@@ -2630,6 +2630,14 @@ The process parent pid.
 
 
 [float]
+=== system.process.pgid
+
+type: integer
+
+The process group id.
+
+
+[float]
 === system.process.cmdline
 
 type: keyword
--- a/metricbeat/etc/fields.yml
+++ b/metricbeat/etc/fields.yml
@@ -1575,6 +1575,10 @@
               type: integer
               description: >
                 The process parent pid.
+            - name: pgid
+              type: integer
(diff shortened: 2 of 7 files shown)


ruflin:
It seems like still some updates on the system tests are needed: https://travis-ci.org/elastic/beats/jobs/143615788#L2408

Can you also update the CHANGELOG file with this change?

nickivanov:
That's an old log, by the way.

ruflin:
@nickivanov Any reason that you the PR's instead of just updating them?

nickivanov:
@ruflin Sorry, I didn't understand the question.

ruflin:
You close https://github.com/elastic/beats/pull/2004 and https://github.com/elastic/beats/pull/2010 and opened https://github.com/elastic/beats/pull/2021 Why not just updating the existing PR with the changes? Otherwise we loose all the comments / feedback every time a new one is opened.

nickivanov:
Oh, sorry, I didn't realize I could update an existing pull request with new commits. Should I reopen #2004 then, and close the latest? You can probably tell I'm not very familiar with the workflow.

ruflin:
No worries. Take the PR of the 3 which is best for you to continue :-) New commits will appear in the PR as soon as you push them to your branch. Let me know if you need some help.

nickivanov:
OK then, I think I'll keep #2021 going forward and update it if it becomes necessary.

Issue #1685, opened 2016-05-20
Replace gotestcover as deprecated


ruflin:
Gotestcover got deprecated. It should be either replaced by one other tool or by the command line suggested as a replacement (see here): https://github.com/pierrre/gotestcover#deprecated

Here a solution that seems to work with our environment can be found: https://github.com/mrkschan/nginxbeat/issues/45#issuecomment-219751959

ruflin:
Gotestcover was moved into the internal repo as it just works and is cross platform compatible. No need to replace it. See https://github.com/elastic/beats/pull/2154

PR #2013, opened 2016-07-12
Starting to replace gotestcover

ruflin:
See https://github.com/elastic/beats/issues/1685

Problem will be Windows compatibilty.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,6 @@ RUN set -x && \
 ## Install go package dependencies
 RUN set -x \
   go get \
-	github.com/pierrre/gotestcover \
 	github.com/tsg/goautotest \
 	golang.org/x/tools/cmd/vet
 
--- a/libbeat/Dockerfile
+++ b/libbeat/Dockerfile
@@ -10,7 +10,6 @@ RUN set -x && \
 ## Install go package dependencies
 RUN set -x \
   go get \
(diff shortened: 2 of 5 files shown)


ruflin:
@pierrre One thing that was nice about gotestcover is that the same command worked on all OS, including Windows :-)

pierrre:
Yes I understand.
But it was by accident and was not supported :P
Actually, I used it only on Travis, so I didn't care about compatibility...

Feel free to fork and maintain your own version.
But for me, my personal and professional projects, I will use the bash script :)

By the way, I just found this:
https://github.com/pkg/errors/blob/master/.travis.yml#L2
https://docs.travis-ci.com/user/languages/go#Go-Import-Path
Could it replace "rsync" in your travis config file? (for forks)

ruflin:
I try to maintain as few projects as possible ;-) TBH a great accident with Windows. 

That is great news about the `go_import_path`. I wasn't aware of this and it should solve our rsync hack. Makes my blog post obsolete :-) https://www.ruflin.com/2015/08/13/fix-for-travis-ci-failure-in-forked-golang-repositories/

ruflin:
Closing in favor of https://github.com/elastic/beats/pull/2154 We decided to add gotestcover as it is to our dependencies as currently gotestcover does everything we need and is cross platform compatible.

go_import_path was implemented and works like a charm.

PR #2019, opened 2016-07-12
Provide 'Done' signaling via beat.Done

urso:


ruflin:
After we introduced "waitSignal" in filebeat, this starts to look even more natural to be used. Some thoughts for this PR to go forward:

* I suggest to introduce the new `beat.Done` signal in a separate PR (without changing the existing beats)
* Introduce the Done channel for the publisher logic and make Stop and additional interface
* Introduce it in all beats and remove old channels

The reason for this split up is to make it for community beats creater easier to track what happened. Most of the beats will not be affected by the PR's. The first two PR's are enhancements which are made in the background. Only the last PR makes the "optional" change to our beats. Currently it seems like we can keep all these changes BC compatible.

urso:
I also see this PR as complementary.

Th PR is already fully BC compatible. If Beat implements 'Stop', it's automatically added into list of stop-callbacks.

ruflin:
@urso The change I would propose above is to split the PR in 3 PR's, so we could point beats developer to a much simpler PR which only contains the migration part of beats from old to new.

monicasarbu:
@urso Should we close this PR as it's too old?

urso:
Yeah, it's quite old, let's close it.

Still, I'm planning to get this in, in one way or the other. This will also help with potential races on `Stop` and simplify the beater interface.

ruflin:
+1 on doing this one day :-)

PR #2020, opened 2016-07-12
Prospector and Harvester Cleanups

ruflin:
See elastic#1913

* Make sure ignore_older is not only applied to new files but also existing files
* Improve logging in prospector
* Add expvar to prospector and harvester
* Cleanup offset handling
* Add truncation handling to prospector
* Handle all offsets as part of the state
* Remove path from harvester and use state.Source instead
* Remove obsolete exclude / include options
* Remove unnecessary getOffset()
* Fix potential file handler leak. Close file handler in case of errors.
* Implement cleaner file handler closing
* Cleanup initOffset handling

### DIFF
--- a/filebeat/harvester/log.go
+++ b/filebeat/harvester/log.go
@@ -2,11 +2,11 @@ package harvester
 
 import (
 	"errors"
+	"expvar"
+	"io"
 	"os"
 
 	"golang.org/x/text/transform"
-
-	"io"
 
 	"github.com/elastic/beats/filebeat/config"
 	"github.com/elastic/beats/filebeat/harvester/processor"
@@ -17,8 +17,20 @@ import (
 	"github.com/elastic/beats/libbeat/logp"
 )
 
+var (
+	harvesterStarted   = expvar.NewInt("filebeat.harvester.started")
+	harvesterClosed    = expvar.NewInt("filebeat.harvester.closed")
+	harvesterRunning   = expvar.NewInt("filebeat.harvester.running")
+	harvesterOpenFiles = expvar.NewInt("filebeat.harvester.open_files")
+	filesTruncated     = expvar.NewInt("filebeat.harvester.files.truncated")
+)
+
 // Log harvester reads files line by line and sends events to the defined output
 func (h *Harvester) Harvest() {
+
+	harvesterStarted.Add(1)
+	harvesterRunning.Add(1)
+	defer harvesterRunning.Add(-1)
 
 	// Makes sure file is properly closed when the harvester is stopped
 	defer h.close()
@@ -31,7 +43,7 @@ func (h *Harvester) Harvest() {
 		return
 	}
 
-	logp.Info("Harvester started for file: %s", h.path)
+	logp.Info("Harvester started for file: %s", h.state.Source)
 
 	processor, err := h.newLineProcessor()
 	if err != nil {
@@ -40,6 +52,7 @@ func (h *Harvester) Harvest() {
 	}
 
 	// Always report the state before starting a harvester
+	// This is useful in case the file was renamed
 	if !h.sendStateUpdate() {
 		return
 	}
@@ -56,14 +69,15 @@ func (h *Harvester) Harvest() {
 		if err != nil {
 			switch err {
 			case reader.ErrFileTruncate:
-				logp.Info("File was truncated. Begin reading file from offset 0: %s", h.path)
-				h.SetOffset(0)
+				logp.Info("File was truncated. Begin reading file from offset 0: %s", h.state.Source)
+				h.state.Offset = 0
+				filesTruncated.Add(1)
 			case reader.ErrRemoved:
-				logp.Info("File was removed: %s. Closing because close_removed is enabled.", h.path)
+				logp.Info("File was removed: %s. Closing because close_removed is enabled.", h.state.Source)
 			case reader.ErrRenamed:
-				logp.Info("File was renamed: %s. Closing because close_renamed is enabled.", h.path)
+				logp.Info("File was renamed: %s. Closing because close_renamed is enabled.", h.state.Source)
 			case io.EOF:
-				logp.Info("End of file reached: %s. Closing because close_eof is enabled.", h.path)
+				logp.Info("End of file reached: %s. Closing because close_eof is enabled.", h.state.Source)
 			default:
 				logp.Info("Read line error: %s", err)
 			}
@@ -71,7 +85,7 @@ func (h *Harvester) Harvest() {
 		}
 
 		// Update offset if complete line has been processed
-		h.updateOffset(int64(bytesRead))
+		h.state.Offset += int64(bytesRead)
 
 		event := h.createEvent()
 
@@ -94,12 +108,13 @@ func (h *Harvester) Harvest() {
 // By default the offset is set to 0, means no bytes read. This can be used to report the status
 // of a harvester
 func (h *Harvester) createEvent() *input.FileEvent {
+
 	event := &input.FileEvent{
 		EventMetadata: h.config.EventMetadata,
-		Source:        h.path,
+		Source:        h.state.Source,
 		InputType:     h.config.InputType,
 		DocumentType:  h.config.DocumentType,
-		Offset:        h.getOffset(),
+		Offset:        h.state.Offset,
 		Bytes:         0,
 		Fileinfo:      h.state.Fileinfo,
 		JSONConfig:    h.config.JSON,
@@ -148,12 +163,27 @@ func (h *Harvester) shouldExportLine(line string) bool {
 // the file system is scanned
 func (h *Harvester) openFile() error {
 
-	f, err := file.ReadOpen(h.path)
+	f, err := file.ReadOpen(h.state.Source)
 	if err != nil {
-		logp.Err("Failed opening %s: %s", h.path, err)
+		logp.Err("Failed opening %s: %s", h.state.Source, err)
 		return err
 	}
 
+	harvesterOpenFiles.Add(1)
+
+	// Makes sure file handler is also closed on errors
+	err = h.validateFile(f)
+	if err != nil {
+		f.Close()
+		harvesterOpenFiles.Add(-1)
+		return err
+	}
+
+	h.file = source.File{f}
+	return nil
+}
+
+func (h *Harvester) validateFile(f *os.File) error {
 	// Check we are not following a rabbit hole (symlinks, etc.)
 	if !file.IsRegular(f) {
 		return errors.New("Given file is not a regular file.")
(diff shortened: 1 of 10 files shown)


filebeat/harvester/log.go
```diff
@@ -148,11 +160,12 @@ func (h *Harvester) shouldExportLine(line string) bool {
 // the file system is scanned
 func (h *Harvester) openFile() error {
 
-	f, err := file.ReadOpen(h.path)
+	f, err := file.ReadOpen(h.state.Source)
```
urso: After doing `file.ReadOpen` a number of checks that might return an error are executed. Question: in case of openFile returning an error, who is closing the file handle?

filebeat/harvester/log.go
```diff
@@ -240,15 +236,9 @@ func (h *Harvester) getState() file.State {
...
-	return h.state
-}
-
-// refreshState refreshes the values in State with the values from the harvester itself
-func (h *Harvester) refreshState() {
-	h.state.Source = h.path
-	h.state.Offset = h.getOffset()
+	// refreshes the values in State with the values from the harvester itself
 	h.state.FileStateOS = file.GetOSState(h.state.Fileinfo)
+	return h.state
```
urso: kind of ugly the harvester has to differentiate between stdin and file.

This change means we're refreshing state for every line reported (sounds quite expensive).

filebeat/harvester/log.go
```diff
@@ -258,16 +248,20 @@ func (h *Harvester) close() {
...
-	logp.Debug("harvester", "Stopping harvester for file: %s", h.path)
+	logp.Debug("harvester", "Stopping harvester for file: %s", h.state.Source)
 
 	// Make sure file is closed as soon as harvester exits
 	// If file was never opened, it can't be closed
 	if h.file != nil {
 		h.file.Close()
-		logp.Debug("harvester", "Stopping harvester, closing file: %s", h.path)
+		logp.Debug("harvester", "Stopping harvester, closing file: %s", h.state.Source)
+		harvesterOpenFiles.Add(-1)
```
urso: why not set `h.file=nil` here?

filebeat/harvester/log.go
```diff
@@ -258,16 +248,20 @@ func (h *Harvester) close() {
...
 	} else {
-		logp.Warn("Stopping harvester, NOT closing file as file info not available: %s", h.path)
+		logp.Warn("Stopping harvester, NOT closing file as file info not available: %s", h.state.Source)
 	}
+
+	harvesterClosed.Add(1)
+	harvesterRunning.Add(-1)
```
urso: counters at correct place (`harvesterRunning`)? Maybe add `harvesterRunning.Add(1); defer harvesterRunning.Add(-1)` to harvester run loop.

filebeat/prospector/prospector_log.go
```diff
@@ -5,12 +5,18 @@ import (
...
+	"expvar"
+
 	"github.com/elastic/beats/filebeat/harvester"
 	"github.com/elastic/beats/filebeat/input"
 	"github.com/elastic/beats/filebeat/input/file"
 	"github.com/elastic/beats/libbeat/logp"
 )
 
+var (
+	filesRenamed = expvar.NewInt("filebeat.prospector.files_renamed")
```
urso: consider adding prospector(input) type: `filebeat.prospector.log.files_renamed`

filebeat/prospector/prospector_log.go
```diff
@@ -171,8 +173,18 @@ func (p *ProspectorLog) harvestExistingFile(newState file.State, oldState file.S
 		// One problem with comparing modTime is that it is in seconds, and scans can happen more then once a second
 		logp.Debug("prospector", "Resuming harvesting of file: %s, offset: %v", newState.Source, oldState.Offset)
 		p.Prospector.startHarvester(newState, oldState.Offset)
+		return
+	}
+
+	// File size was reduced -> truncated file
+	if oldState.Finished && newState.Fileinfo.Size() < oldState.Offset {
+		logp.Debug("prospector", "Old file was truncated. Starting from the beginning: %s", newState.Source)
```
urso: truncate should have a counter as renamed has.

filebeat/registrar/registrar.go
```diff
@@ -27,6 +27,7 @@ type Registrar struct {
 
 var (
 	statesUpdated = expvar.NewInt("registrar.state_updates")
+	statesTotal   = expvar.NewInt("registar.states.total")
```
urso: counters for states added and removed would be nice too.

filebeat/harvester/log.go
ruflin: You are right :-( This could be a reason we leak some file descriptors. We must backport this fix to 1.3 @tsg FYI

filebeat/harvester/log.go
ruflin: Yes, we should optimize this. But this was the case before already, it was just no visible very well.

In general the filestate of a Fileinfo should never change as long as the Fileinfo doesn't change? So we would have to read this only once?

ruflin: About harvester stdin/log: Yes, it should be 2 different harvesters in the future.

filebeat/harvester/log.go
ruflin: What would be the effect / advantage of doing this?

filebeat/prospector/prospector_log.go
ruflin: I added a counter. A note here: either the harvester can detect it if it is an open file or the prospector if is is a non actively harvested file. Should we use 2 different counters?

filebeat/registrar/registrar.go
ruflin: These should be counted on a prospector level

ruflin: @urso let me add more counters in a separate PR. I added a note to https://github.com/elastic/beats/issues/1913

filebeat/harvester/log.go
urso: there is a check `h.file != nil` right before `h.file.Close()`. So `h.file == nil` only if file has not been opened before? Can harvester be closed twice?

filebeat/prospector/prospector_log.go
urso: I think one counter should be ok. In general `file truncation` is potential source of data-loss.

filebeat/harvester/log.go
ruflin: `h.file` is `nil` in case the `h.open()` fails as it was never assigned.

filebeat/harvester/log.go
```diff
@@ -186,52 +216,34 @@ func (h *Harvester) openFile() error {
...
 	offset, err := file.Seek(0, os.SEEK_CUR)
 
-	if h.getOffset() > 0 {
+	if h.state.Offset > 0 {
 		// continue from last known offset
-		logp.Debug("harvester",
-			"harvest: %q position:%d (offset snapshot:%d)", h.path, h.getOffset(), offset)
-		_, err = file.Seek(h.getOffset(), os.SEEK_SET)
+		logp.Debug("harvester", "harvest: %q position:%d (offset snapshot:%d)", h.state.Source, h.state.Offset, offset)
+		_, err = file.Seek(h.state.Offset, os.SEEK_SET)
```
urso: missing `offset = h.state.Offset` ?

ruflin: Oh, glad you spotted this as otherwise it gets overwritten below :-(

filebeat/harvester/log.go
```diff
@@ -186,52 +216,38 @@ func (h *Harvester) openFile() error {
 		return err
 	}
 
-	// yay, open file
-	h.file = source.File{f}
 	return nil
 }
 
 func (h *Harvester) initFileOffset(file *os.File) error {
```
ruflin: @urso After your input I decided to rewrite / cleanup this part here. I think it is identical to before, but now max 1 Seek instead of 2 are needed and it seems easier to read. Can you have a close look if it looks sane for you?

filebeat/harvester/log.go
```diff
@@ -186,52 +216,38 @@ func (h *Harvester) openFile() error {
...
-	} else if h.config.TailFiles {
-		// tail file if file is new and tail_files config is set
-		logp.Debug("harvester", "harvest: (tailing) %q (offset snapshot:%d)", h.path, offset)
+	var err error
 
-		offset, err = file.Seek(0, os.SEEK_END)
-		h.SetOffset(offset)
+	// continue from last known offset
+	if h.state.Offset > 0 {
+		h.state.Offset, err = file.Seek(h.state.Offset, os.SEEK_SET)
```
urso: do not set `h.state.Offset` here. Which value will the offset be if seek fails (even if fail is temporary)?

urso: On success value, offsets should be same.

filebeat/harvester/log.go
```diff
@@ -186,52 +216,38 @@ func (h *Harvester) openFile() error {
...
-		h.SetOffset(offset)
+	// tail file if file is new and tail_files config is set
+	if h.config.TailFiles {
+		h.state.Offset, err = file.Seek(0, os.SEEK_END)
+		logp.Debug("harvester", "Setting offset for tailing file: %s. Offset (end of file): %d ", h.state.Source, h.state.Offset)
+		return err
 	}
 
+	// get offset from file in case of encoding factory was required to read some data.
+	h.state.Offset, err = file.Seek(0, os.SEEK_CUR)
```
urso: when querying offset here, we expect file to be new? can h.state.Offset become invalid here, us resending complete log by accident if harvester returns?

urso:
@ruflin needs rebase

filebeat/harvester/log.go
ruflin: Based on your input I made two adjustments:

* offset is only updated if seek does not return any error. I move the update of the offset to the validate method
* state update is only sent in the close method, if file was successfully initialised. I think it should not make a difference as the update would stay the same if no change happened, but it is not necessary and somehow makes it clear, that on setup error no updates are sent.

filebeat/harvester/log.go
```diff
@@ -180,58 +210,42 @@ func (h *Harvester) openFile() error {
...
-	h.file = source.File{f}
+	logp.Debug("harvester", "Setting offset for file: %s. Offset: %d ", h.state.Source, offset)
+	h.state.Offset = offset
+
 	return nil
 }
 
-func (h *Harvester) initFileOffset(file *os.File) error {
-	offset, err := file.Seek(0, os.SEEK_CUR)
+func (h *Harvester) getFileOffset(file *os.File) (int64, error) {
```
urso: I think `initFileOffset` still matches this methods functionality better. `getFileOffset` sounds like re-usable getter.

Issue #1735, opened 2016-05-26
default username/password in *beat.full.yml should be elastic/changeme


LeeDr:
Please post all questions and issues on https://discuss.elastic.co/c/beats
before opening a Github Issue. Your questions will reach a wider audience there,
and if we confirm that there is a bug, then you can open a new issue.

For confirmed bugs, please report:
- Version: 5.0.0-alpha3
- Operating System: Ubuntu 15.10
- Steps to Reproduce: If we're going to have well known default superuser in elasticsearch (in x-pack) then we should also have those as the defaults in the config files of our other products.

PR #2023, opened 2016-07-13
Added https user/pass to the default template

tsg:
Since we removed the template stuff from the default config, we can add
the user/pass instead so it's easy to enabled https. It also updates the
sample user/pass used, to match the default in Shield, closing #1735.

Depends on #1993.


ruflin:
@tsg did you run `make update`?

tsg:
@ruflin heh, totally forgot. Done now.

ruflin:
LGTM

PR #2024, opened 2016-07-13
Use pywin download hosted by us

tsg:
Appveyor caches are invalidated when .appveyor.yml changes, so this
does some minor edits to the file to see if it fixes the build.


tsg:
Changed the PR to use the pywin binary hosted by us on S3.

andrewkroh:
Looks like the AppVeyor build is off to a good start now 👍 .  I will wait for the Winlogbeat tests to run since the version of PyWin is changing.

PR #2025, opened 2016-07-13
Use dotted notation in the ingest node docs

tsg:
Minor edit for our ingest node guide. Now that dots are supported, we
can avoid the curlies and still keep it in one line.

Also switched the example to use localhost instead of the docker machine IP,
which is less likely to be used by the users.


ruflin:
LGTM

PR #2026, opened 2016-07-13
Make full config file optional for beats

ruflin:
Most of the community beats dont make a difference between the full and basic config as it is additional overhead to manage it and most configs are simple. The full config build now falls back to the basic config if it the full doesnt exist.

This also fixes the doc build as the option -web does not exist anymore.


monicasarbu:
@ruflin we need also to update the `vendor` directory, right?

ruflin:
@monicasarbu Can you elaborate on that?

PR #2029, opened 2016-07-13
Fix open file handler issue

### LINKED ISSUE #2028: Fix open file handler issue
If during the start of a harvester and error happens during opening the file, it can happen that the file descriptor is not properly closed. This error was found in 5.x and is fixed in https://github.com/elastic/beats/pull/2020 but applies also all 1.x filebeat versions.

The openFile method must be adjusted so that also in case of an error during the first file checks, the file descriptor is closed properly: https://github.com/elastic/beats/blob/master/filebeat/harvester/log.go#L151

ruflin:
Closes #2028

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -33,6 +33,8 @@ https://github.com/elastic/beats/compare/v1.3.0...1.3[Check the HEAD diff]
 
 *Filebeat*
 
+- Fix open file handler issue {issue}2028[2028]
+
 *Winlogbeat*
 
 ==== Added
--- a/filebeat/harvester/log.go
+++ b/filebeat/harvester/log.go
@@ -227,25 +227,31 @@ func (h *Harvester) openFile() (encoding.Encoding, error) {
 	var encoding encoding.Encoding
 
 	file, err = input.ReadOpen(h.Path)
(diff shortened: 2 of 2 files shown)


filebeat/harvester/log.go
```diff
@@ -227,25 +227,31 @@ func (h *Harvester) openFile() (encoding.Encoding, error) {
...
-			return nil, errors.New("Given file is not a regular file.")
-		}
+	if err != nil {
+		logp.Err("Failed opening %s: %s", h.Path, err)
+		return nil, err
+	}
 
-		encoding, err = h.encoding(file)
+	// Closes file handler in case of follow up error
+	defer func() {
```
ruflin: @urso I used the "short" method here to only change the minimal lines of code.

CHANGELOG.asciidoc
```diff
@@ -33,6 +33,8 @@ https://github.com/elastic/beats/compare/v1.3.0...1.3[Check the HEAD diff]
 
 *Filebeat*
 
+- Fix open file handler issue {issue}2028[2028]
```
urso: close newly opened files faster on error. (file closing is backed by gc by default)

urso:
LGTM

Issue #1997, opened 2016-07-08
Elasticsearch output reconnects on 4xx response codes


tsg:
This causes problems for, e.g., the template loading code, which checks if the template is already loaded, and a 404 is expected if the template is not there.

I proposed a simplistic solution in #1996, but @urso is saying that we might be able to get rid of the `connected` flag altogether and simplify the API in the process.

tsg:
Marked as bug for beta1 as this results in logging a connection error when there was actually none. So it can cause confusion.

PR #2030, opened 2016-07-13
Remove IsConnected from ProtocolClient interface

urso:
- Always close connections on failure and enforce reconnect.
- IsConnected flag gone from elasticsearch output (no more error if template already loaded).

Closes: #1997


ruflin:
LGTM

tsg:
LGTM, and can confirm that this fixes #1997.

PR #2032, opened 2016-07-14
Do not send empty values in case of errors

### LINKED ISSUE #1972: Do not send Metricset values which were not set
Currently most metricsets send a default value in case a value is not set. In contrast the mongodb/status metricset removes all fields which are not set. An example can be found here https://github.com/elastic/beats/blob/master/metricbeat/module/mongodb/status/data.go#L175

The MongoDB implementation has the following advantage:
* Default values can be misleading as they can conflict with the actual values sent (0 can have 2 meanings)
* It makes the metricsets more cross version compatible with services. So if an older service does not contain the value, it is just not sent.

All Metricsets should have the same implementation as MongoDB has. The logic in MongoDB to remove fields should be generalised.

tsg:
It used to be that we were sending zero values (e.g. "" for string, 0 for int)
in case of errors getting a particular value.

Fixes #1972.

I'm not very happy with this implementations, because it requires a lot of repetition when doing the conversions. I want to try another approach, but I'm opening this one for discussion and to have something to compare the new code against.


tsg:
Closing in favor of #2034 which I think is better.

PR #2034, opened 2016-07-14
Don't include empty values in Metricbeat

### LINKED ISSUE #1792: Rename config topics to refelect reorganization of the config file

### LINKED ISSUE #1972: Do not send Metricset values which were not set

tsg:
This is an alternative implementation of #2032. Instead of collecting the errors
and then removing the empty values, we define a list of conversions and we use code
to apply them, so we have opportunity to handle errors.

Fixes #1972.

### DIFF
--- a/metricbeat/helper/conversion.go
+++ b/metricbeat/helper/conversion.go
@@ -1,85 +1,168 @@
-/*
-The conversion functions take a key and a map[string]string. First it checks if the key exists and logs and error
-if this is not the case. Second the conversion to the type is done. In case of an error and error is logged and the
-default values is returned. This guarantees that also if a field is missing or is not defined, still the full metricset
-is returned.
-*/
 package helper
 
 import (
+	"fmt"
 	"strconv"
 
+	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/logp"
 )
 
+// Schema describes how a map[string]string object can be parsed and converted into
+// an event. The conversions can be described using an (optionally nested) common.MapStr
+// that contains Conv objects.
+type Schema struct {
+	conversions common.MapStr
+}
+
+// A Conv object represents a conversion mechanism from the data map to the event map.
+type Conv struct {
+	Func     Convertor // Convertor function
+	Key      string    // The key in the data map
+	Optional bool      // Whether to log errors if the key is not found
+}
+
+// Convertor function type
+type Convertor func(key string, data map[string]string) (interface{}, error)
+
+// NewSchema creates a new converting schema.
+func NewSchema(conversions common.MapStr) Schema {
+	return Schema{conversions}
+}
+
+// ApplyTo adds the fields extracted from data, converted using the schema, to the
+// event map.
+func (s Schema) ApplyTo(event common.MapStr, data map[string]string) common.MapStr {
+	applySchemaToEvent(event, data, s.conversions)
+	return event
+}
+
+// Apply converts the fields extracted from data, using the schema, into a new map.
+func (s Schema) Apply(data map[string]string) common.MapStr {
+	return s.ApplyTo(common.MapStr{}, data)
+}
+
+func applySchemaToEvent(event common.MapStr, data map[string]string, conversions common.MapStr) {
+	for key, conversion := range conversions {
+		switch conversion.(type) {
+		case Conv:
+			conv := conversion.(Conv)
+			value, err := conv.Func(conv.Key, data)
+			if err != nil {
+				if !conv.Optional {
+					logp.Err("Error on field '%s': %v", key, err)
+				}
+			} else {
+				event[key] = value
+			}
+		case common.MapStr:
+			subEvent := common.MapStr{}
+			applySchemaToEvent(subEvent, data, conversion.(common.MapStr))
+			event[key] = subEvent
+		}
+	}
+}
+
 // ToBool converts value to bool. In case of error, returns false
-func ToBool(key string, data map[string]string) bool {
+func ToBool(key string, data map[string]string) (interface{}, error) {
 
 	exists := checkExist(key, data)
 	if !exists {
-		logp.Err("Key does not exist in in data: %s", key)
-		return false
+		return false, fmt.Errorf("Key `%s` not found", key)
 	}
 
 	value, err := strconv.ParseBool(data[key])
 	if err != nil {
-		logp.Err("Error converting param to bool: %s", key)
-		return false
+		return false, fmt.Errorf("Error converting param to bool: %s", key)
 	}
 
-	return value
+	return value, nil
+}
+
+// Bool creates a Conv object for parsing booleans
+func Bool(key string, opts ...SchemaOption) Conv {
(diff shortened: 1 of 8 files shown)


tsg:
Personally, I like this one more than #2032. It's less repetition in the code and the module writer can add their own convertors functions if they wish. Also, we can extend the `Conv` type with more functionality in the future, e.g. marking fields as optionals (in which case we don't log an error).

metricbeat/module/apache/status/data.go
```diff
@@ -15,6 +15,42 @@ var (
 
 	// This should match: "CPUSystem: .01"
 	matchNumber = regexp.MustCompile("(^[0-9a-zA-Z ]+):\\s+(\\d*\\.?\\d+)")
+
+	conversions = common.MapStr{
```
urso: this almost looks like a schema/conversion defintion. Next step: make conversions table loadable from file?

metricbeat/helper/conversion.go
```diff
@@ -7,75 +7,96 @@ is returned.
...
+			} else {
+				event[key] = value
+			}
+		case common.MapStr:
+			subEvent := common.MapStr{}
+			AddConversions(subEvent, data, conversion.(common.MapStr))
+			event[key] = subEvent
+		}
+	}
+}
```
urso: how about (naming it `Schema` here, as `Convertor` is already in use):

```
type Schema struct {
    conversions common.MapStr
}

func Make(c common.MapStr) Schema {
  return Schema{c}
}

func (s Schema) Apply(data map[string]string) common.MapStr {
    return s.ApplyTo(common.MapStr{}, data)
}

func (s Schema) ApplyTo(event common.MapStr, data map[string]string) common.MapStr {
    // apply s.conversions here

    return event
}
```

Instead of doing h.Conv{h.ToInt, ...} these can be 'unified' in `schema`, `typ` package:

```
package schema

func Int(s string) h.Conv { return h.Conv{ h.ToInt, s } }

...
```

then schema definition becomes:

```
schema := schema.Make(common.MapStr{
    "count": schema.Int("count"),
    "sub": common.Mapstr{
        "metric": schema.Float("f"),
    }
})
```

Combine this with re-usable data-scrappers/extractors like:

```
extractor := RegexExtractor(common.MapStr{
    "count": "count: (.*)",
})

...

func Exec(in io.Reader, extractor Extractor, schema Schema) (common.MapStr, error) {
    data, err := extractor.Read(in)
    if err != nil {
        return nil, err
    }
    return schema.Apply(data), nil
}
```

And metricsets implementations become quite 'declarative'

tsg: Thanks for the schema API ideas, I adopted most of them. Please have another look.

metricbeat/helper/conversion.go
```diff
@@ -1,85 +1,168 @@
...
 
+	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/logp"
 )
 
+// Schema describes how a map[string]string object can be parsed and converted into
+// an event. The conversions can be described using an (optionally nested) common.MapStr
+// that contains Conv objects.
+type Schema struct {
+	conversions common.MapStr
```
urso: instead of `common.MapStr` consider `map[string]Mapper`

where

```
type Mapper interface{
  Map( ... ) error
}

// implements Mapper on top of Conv
type Primitive struct {
  Conv
}

// implements Mapper interface for structure
type Object struct {
  map[string]Conv
}
```

This will add some type-safety + no need for type-switch in applySchemaToEvent, as (Mapper).Map basically implements applySchemaToEvent (Schema is basically `type Schema Object`.

This turns schema definition from `NewSchema(common.MapStr{ ... })` into `h.Schema{ ... }`.

metricbeat/module/mysql/status/data.go
```diff
@@ -5,47 +5,48 @@ import (
...
-	}
+		"opened_tables": h.Int("Opened_tables"),
+	})
+)
 
-	return event
+// Map data to MapStr of server stats variables: http://dev.mysql.com/doc/refman/5.7/en/server-status-variables.html
+// This is only a subset of the available values
+func eventMapping(status map[string]string) common.MapStr {
+	return schema.Apply(status)
```
urso: long form for `var eventMapping = schema.Apply`

metricbeat/helper/conversion.go
tsg: +1 on this, especially for type safeness. I'll try it in #2040 or in a cleanup PR after that. The Mongodb code complicates all of this, because it's input is `map[string]interface{}` not `map[string]string`. It would be nice to find a solution that doesn't involve to much repetition of these types.

metricbeat/module/mysql/status/data.go
tsg: Heh, yes :)

metricbeat/module/apache/status/data.go
```diff
@@ -89,40 +125,7 @@ func eventMapping(body io.ReadCloser, hostname string) common.MapStr {
...
-				"keep_alive": h.ToInt("ConnsAsyncKeepAlive", fullEvent),
-				"closing":    h.ToInt("ConnsAsyncClosing", fullEvent),
-			},
-		},
-		"load": common.MapStr{
-			"1":  h.ToFloat("Load1", fullEvent),
-			"5":  h.ToFloat("Load5", fullEvent),
-			"15": h.ToFloat("Load15", fullEvent),
-		},
+		"hostname": hostname,
```
ruflin: Should we also add the string fields to the schema to have all fields in the schema? Reason is that the schema above could also allow us to potentially create things like the mapping?

ruflin:
@tsg Great PR. I really like the separation of schema and data. We should probably add a note about this to the Dev guide.

metricbeat/module/apache/status/data.go
tsg: it's not about it being a string, the hostname is different because it's not coming from the full event, that's why it's hard to put it in the schema. I considered defining the schema in the code rather then one-time, but creating all those objects to use only one time seemed wasteful. 

What we could do is put the hostname in the full event artificially and get it from there with the schema.

ruflin: TBH not 100% sure if we should send the hostname here as it is already part of the event under `metricset.host`.

PR #2036, opened 2016-07-14
Update the README with the latest changes

monicasarbu:
Use `make setup` instead of `make init` and `make commit`.


generate/beat/{{cookiecutter.beat}}/README.md
```diff
@@ -12,16 +12,11 @@ Ensure that this folder is at the following location:
 * [Golang](https://golang.org/dl/) 1.6.2
 
 ### Init Project
-To get running with {{cookiecutter.beat|capitalize}}, run the following command:
+To get running with {{cookiecutter.beat|capitalize}} and install also the
```
andrewkroh: "and also install" sounds better.

PR #2040, opened 2016-07-15
Adopt the new schema approach for the MongoDb module

tsg:
This is applying the ideas from #2034 to the MongoDB module, simplifying the
code significantly. Unfortunately, the same conversions functions cannot be applied
as is to the map[string]interface{} that we get from the mongodb driver, so I had to
reimplement them.

Besides simplifying the logic, this fixes a bug: in the previous version if an error happened on a key that was named differently from the source, the key was not removed.

This depends on #2034.


metricbeat/module/mongodb/status/conversion.go
```diff
@@ -0,0 +1,97 @@
+package status
```
ruflin: Should this code be moved into the helper?

tsg: Yeah, I'm trying to find a way to organize these in a way that makes sense. We basically have two sets of convertors. One set that work on `map[string]string` (one level deep) and one that work on `map[string]interface{}` (nested). I'm thinking to have something like:

```
helper/
  schema.go
  mapstrstr/
    ...
  mapstriface/
    ...
```

The names are not great, I know. We could also either move the whole thing one level down under `helper/schema/` or replace `helper` with `schema` as there are no other helpers.

ruflin: +1 on moving it to schema as a top level package as this is so critical to building metricsets. Also then we can use `s.*` instead of `h.*` which somehow feels more logical. Or would you use the full package name every time?

tsg:
@ruflin @urso: I did all the refactoring I had planned, but it turned out to be quite a long PR. I could try splitting it in smaller ones if you think it's worth it. Please have a look, and I can cleanup the history afterwards.

metricbeat/module/apache/status/data.go
```diff
@@ -16,41 +17,41 @@ var (
...
-	schema = h.NewSchema(common.MapStr{
-		"total_accesses":    h.Int("Total Accesses"),
-		"total_kbytes":      h.Int("Total kBytes"),
-		"requests_per_sec":  h.Float("ReqPerSec", h.Optional),
-		"bytes_per_sec":     h.Float("BytesPerSec", h.Optional),
-		"bytes_per_request": h.Float("BytesPerReq", h.Optional),
-		"workers": common.MapStr{
-			"busy": h.Int("BusyWorkers"),
-			"idle": h.Int("IdleWorkers"),
+	schema_ = schema.Schema{
```
ruflin: Little bit odd naming convention, especially as it will also be used by others. Perhaps going with `s` and using `c` for the s. C would stand for conversion. BTW: if we use `.` for the include `mapstrstr`. Never tried it but that should make it part of the same namespace. So we could use it without prefix.

metricbeat/module/redis/keyspace/data.go
```diff
@@ -37,11 +38,11 @@ func findKeyspaceStats(info map[string]string) map[string]string {
 	return keyspace
 }
 
-var schema = h.NewSchema(common.MapStr{
-	"keys":    h.Int("keys"),
-	"expires": h.Int("expires"),
-	"avg_ttl": h.Int("avg_ttl"),
-})
+var schema_ = schema.Schema{
```
ruflin: I really like that the schema is only created ones.

metricbeat/module/apache/status/data.go
tsg: So we'll have `s.Schema` and `c.Int`, right? Yeah, that could also work.

I'd prefer not going to dot imports, those are usually more messy and upset linters.

metricbeat/schema/mapstriface/mapstriface.go
```diff
@@ -0,0 +1,159 @@
+/*
+Package mapstrstr contains utilities for transforming map[string]string objects
```
ruflin: this is package mapstriface not mapstrstr

metricbeat/schema/mapstriface/mapstriface.go
```diff
@@ -0,0 +1,159 @@
...
+		"test_obj_1": schema.Object{
+			"test": Str("testNonNestedObj"),
+		},
+		"test_obj_2": Dict("testObj", schema.Schema{
+			"test": Str("testObjString"),
+		}),
+	}
+	output := schema_.Apply(input)
+
+Note that this allows for converting, renaming, and restructuring the data.
```
ruflin: Can we have this also in the dev guide? ;-)

metricbeat/module/apache/status/data.go
urso: yeah, please no dot imports if possible. I'd also put all functions/types used to define a schema  into one package (or at least re-export).

metricbeat/schema/mapstriface/mapstriface.go
```diff
@@ -0,0 +1,159 @@
...
+		logp.Err("Error accessing sub-dictionary `%s`", convMap.Key)
+		return
+	}
+
+	subEvent := common.MapStr{}
+	convMap.Schema.ApplyTo(subEvent, subData)
+	event[key] = subEvent
+}
+
+func Dict(key string, schema_ schema.Schema) ConvMap {
```
ruflin: also use `s` instead of `schmea_` here

ruflin:
LGTM. So minor naming comments.

Huge step forward. That will make creating metricsets even easier.

ruflin:
jenkins, retest it

metricbeat/module/apache/status/data.go
tsg: All in one package wouldn't be possible without significant repetition between `mapstrstr` and `mapstriface`. What do you mean by re-export. I tried something like `type Schema schema.Schema` but that doesn't really work because the resulting type is not compatible.

metricbeat/schema/mapstrstr/mapstrstr_test.go
```diff
@@ -0,0 +1,51 @@
...
+		"testBool":       "true",
+		"testFloat":      "42.1",
+		"testObjString":  "hello, object",
+		"testError":      42,     // invalid, only strings are allowed
+		"testErrorInt":   "12a",  // invalid integer
+		"testErrorFloat": "12,2", // invalid float
+		"testErrorBool":  "yes",  // invalid bool
+	}
+
+	schema_ := schema.Schema{
```
ruflin: Still `schema_` but not critical as it is part of the tests.

tsg:
History cleaned up, let's wait for green. I'm preparing another PR for adding some docs to the dev guide.

PR #2042, opened 2016-07-15
Make default path.X available in config file

urso:


filebeat/filebeat.full.yml
```diff
@@ -209,7 +209,7 @@ filebeat.prospectors:
 
 # Name of the registry file. If a relative path is used, it is considered relative to the
 # data path.
-#filebeat.registry_file: registry
+#filebeat.registry_file: ${path.data}/registry
```
tsg: Writing `filebeat.registry_file: ${path.data}/registry` will be equivalent with writing `filebeat.registry_file: registry`, right?

I like the version with `${path.data}` as it's more explicit about where the file is searched. But we'll need to explain in the docs that these variables are automatically set somehow.

tsg:
Code LGTM. There's a unit test failure at the moment.

Unfortunately I think we don't have system tests for these, so I guess we'll need to do some manual tests to check that:

* the defaults still respect this: https://www.elastic.co/guide/en/beats/filebeat/master/directory-layout.html
* CLI flags take precedence over config, which takes precedence over the above defaults

filebeat/filebeat.full.yml
urso: yes, it's equivalent.

tsg:
@urso yeah, we don't have `path.bin` configurable since it is always the location of the binary. Do you need it for a config setting or is it just for consistency with ES?

libbeat/beat/beat.go
```diff
@@ -238,6 +241,8 @@ func (b *Beat) configure() error {
 		return fmt.Errorf("error loading config file: %v", err)
 	}
 
+	// init paths config:
```
ruflin: Does that comment belong on line 251?

urso:
@tsg 
> yeah, we don't have path.bin configurable since it is always the location of the binary. Do you need it for a config setting or is it just for consistency with ES?

Just for consistency. But as there is currently no usage for it, let's keep it out.

libbeat/beat/beat.go
urso: line can be removed. comment left over from original failed (more complicated) approach.

tsg:
Looks good on some brief manual testing. @urso happy to merge it, do you want to cleanup history or should I just squash it?

PR #2045, opened 2016-07-15
Export total metrics in system module

monicasarbu:


monicasarbu:
Here is the result. A Kibana dashboard for Network statistics. 

metricbeat network dashboard

metricbeat/metricbeat.yml
```diff
@@ -14,7 +14,7 @@ metricbeat.modules:
 - module: system
   metricsets:
     # CPU stats
-    - cpu
+    #- cpu
```
andrewkroh: Did you intend to check in these changes?

andrewkroh:
There's a test failure. `Exception: Key 'system.network.total.bytes' found in event is not documented!`

andrewkroh:
Could we use a scripted field in Kibana to derive the totals instead of storing them in ES?

ruflin:
👍  on the proposal of @andrewkroh It would be interesting to find out how "complex" it is to do rather simple stuff (like total) directly in ES instead of sending the data.

monicasarbu:
Yes, I agree it would be great to use the scripted fields. I will give it a try and see how it works.

monicasarbu:
Close it as the total values are done by scripted fields in Kibana

Issue #1376, opened 2016-04-12
enable with_vlan by default


urso:
by default with_vlan is set to false. Setting to true improves out of box experience.

Issue #621, opened 2016-01-05
Duplicate packets when using af_packet


devfacet:
Hi All,

I have configuration like;

``` yml
interfaces:
  device: any
  type: af_packet
  buffer_size_mb: 124
protocols:
  dns:
    ports: [53]
    include_authorities: true
    include_additionals: true
output:
  file:
    filename: output.log
    path: /var/log/packetbeat
    rotate_every_kb: 100000
    number_of_files: 2
  elasticsearch:
    hosts: ["localhot:9200"]
```

Once I ran `nslookup google.com`

```
;; Got SERVFAIL reply from 127.0.0.1, trying next server
Server:     172.16.0.23
Address:    172.16.0.23#53

Non-authoritative answer:
Name:   google.com
Address: 74.125.22.113
Name:   google.com
Address: 74.125.22.138
Name:   google.com
Address: 74.125.22.139
Name:   google.com
Address: 74.125.22.100
Name:   google.com
Address: 74.125.22.101
Name:   google.com
Address: 74.125.22.102
```

`packetbeat` goes infinity loop;

```
2016/01/05 02:38:13.432665 preprocess.go:94: DBG  Forward preprocessed events
2016/01/05 02:38:13.432673 async.go:44: DBG  async forward to outputers (2)
2016/01/05 02:38:13.432693 output.go:50: DBG  output worker: publish single event
2016/01/05 02:38:13.904256 output.go:103: DBG  output worker: publish 2 events
2016/01/05 02:38:14.433497 client.go:55: DBG  send event
2016/01/05 02:38:14.433536 async.go:71: DBG  send async event
2016/01/05 02:38:14.433755 preprocess.go:37: DBG  Start Preprocessing
2016/01/05 02:38:14.433891 publish.go:100: DBG  Publish: {
  "@timestamp": "2016-01-05T02:38:00.915Z",
  "beat": {
    "hostname": "...",
    "name": "..."
  },
  "bytes_out": 28,
  "client_ip": "127.0.0.1",
  "client_port": 59705,
  "client_proc": "",
  "client_server": "...",
  "count": 1,
  "direction": "out",
  "dns": {
    "additionals_count": 0,
    "answers_count": 0,
    "authorities_count": 0,
    "flags": {
      "authoritative": false,
      "recursion_allowed": false,
      "recursion_desired": true,
      "truncated_response": false
    },
    "id": 31763,
    "op_code": "QUERY",
    "question": {
      "class": "IN",
      "name": "google.com",
[...]

Issue #522, opened 2015-12-14
Initializing af_packet sniffer failed


lopaz:
**packetbeat.yml:**
interfaces:
  device: any
  type: af_packet

**ERROR Log:**
2015/12/14 03:35:53.074127 sniffer.go:129: DBG  Sniffer type: af_packet device: any
2015/12/14 03:35:53.079966 packetbeat.go:195: CRIT Initializing sniffer failed: Error creating sniffer: setsockopt packet_rx_ring: cannot allocate memory

**uname -a** output:
Linux testmachine1 2.6.32-504.el6.x86_64 #1 SMP Wed Oct 15 04:27:16 UTC 2014 x86_64 x86_64 x86_64 GNU/Linux

**packetbeat** rpm:
https://download.elastic.co/beats/packetbeat/packetbeat-1.0.0-x86_64.rpm

**free** output:
             total       used       free     shared    buffers     cached
Mem:       3924628    2014656    1909972          8     176064    1584936
-/+ buffers/cache:     253656    3670972
Swap:      4194300       6196    4188104

djschny:
FYI - ran into this on darwin as well

andrewkroh:
@djschny  af_packet isn't available on darwin because it's part of the Linux kernel. https://www.elastic.co/guide/en/beats/packetbeat/current/configuration-interfaces.html#_type

dberuben:
I have the same issue 

interfaces:
  device: any
  type: af_packet

**version** 
packetbeat-1.1.2-1.x86_64

ERROR 
2016/03/23 14:52:15.140040 packetbeat.go:215: CRIT Initializing sniffer failed: Error creating sniffer: setsockopt packet_rx_ring: cannot allocate memory

**Memory**

[root@localhost packetbeat]# free -om
             total       used       free     shared    buffers     cached
Mem:          4009       3224        785          0        314        773
Swap:            0          0          0

**Linux Centos 6.7**
uname -a
Linux localhost 2.6.32-431.5.1.el6.x86_64 #1 SMP Wed Feb 12 00:41:43 UTC 2014 x86_64 x86_64 x86_64 GNU/Linux

hnlq715:
+1

packetbeat version 5.0.0-SNAPSHOT (amd64), libbeat 5.0.0-SNAPSHOT
CentOS release 6.4 (Final)
[...]

PR #2048, opened 2016-07-15
packetbeat: use af_packet by default on linux

urso:


andrewkroh:
LGMT. This reminded me of #1376.

Is there anything in the docs that should be updated for this one?

tsg:
This one is debatable, af_packet requires more memory +  a config option enabled at compile time (luckily on in most distros). I'm not sure it's a good idea to change the default.

andrewkroh:
Additionally there are two issues that we might see more occurrences of if we change to af_packet by default.

#522 
#621

urso:
right, increased memory usage. Advantage, much less chance of packet-loss (libpcap based sniffer has hugher overhead).

@andrewkroh I'm not sure #522 is really an issue. af_packet requires to allocate memory in kernel-space, but if kernel can not allocate a big enough continuous space bad luck. #621 is more interesting. I remember reading about kernel bugs with TPACKET V3 (not sure about V2 anymore).

tsg:
> I'm not sure #522 is really an issue. af_packet requires to allocate memory in kernel-space, but if kernel can not allocate a big enough continuous space bad luck. 

If I remember correctly that can happen when the memory is in disk caches as well, which can easily happen in normal operation. We'd probably need to drop the caches from the init script/systemd file if we are to make this the default, but we'd have to do that only for Packetbeat.

tsg:
Or we could have Packetbeat itself do it only if it detects an allocation error. It might still be surprising for operators that we drop caches on Packetbeat start.

urso:
yeah, with packetbeat potentially installed on application servers flushing all caches on startup might be quite annoying.

Normally it shouldn't be much of a problem, as packetbeat should be started early on by init-scripts. But when installing and testing packetbeat, one might want to try running it in foreground.

PR #2053, opened 2016-07-18
Set negate to true in multiline option.

7lima:
According to the [multiline doc] (https://www.elastic.co/guide/en/beats/filebeat/current/configuration-filebeat-options.html#multiline), "negate: true" means that the pattern should match the lines that are NOT a continuation but the start of a new multiline event. This is the case here like in the following example (Timestamps).

### DIFF
--- a/filebeat/docs/multiline.asciidoc
+++ b/filebeat/docs/multiline.asciidoc
@@ -100,11 +100,11 @@ To consolidate these lines into a single event in Filebeat, use the following mu
 -------------------------------------------------------------------------------------
 multiline:
   pattern: '\\$'
-  negate: false
+  negate: true
   match: after
 -------------------------------------------------------------------------------------
 
-This configuration merges any line that ends with the `\` character with the line that follows.
+This configuration merges any line that ends with the `\` character with the line that follows. The `negate: true` setting means that the pattern matches the lines that aren't a continuation line but the start of a new multiline event.
 
 ===== Timestamps
 


tsg:
@7lima the example in the docs is indeed wrong, good catch and thanks for notifying us about it. To fix it, however, I'd prefer to change it to the following:

```
multiline:
  pattern: '\\$'
  negate: true
  match: before
```

This has the advantage that it works if there are multiple consecutive lines ending in `\`. And we actually have a [test](https://github.com/elastic/beats/blob/master/filebeat/tests/system/test_multiline.py#L43) for this exact case.

Let me know if you want to update your PR with this. If you don't have the time, we can also open another PR to fix it.

7lima:
@tsg Now in your proposed setup we have the case: "Consecutive lines that don’t match the pattern are prepended to the next line that does match."
So in an example:
```
a
b
c\
d\
e
f
```
we would get as the result (as far as I understand the docs):
```
a b c\
d\
e
f
```
instead of (from my setup) (as far as I understand the docs):
```
a
b
c\
d\ e f
```
These are both imperfect solutions for the C-style line continuation.

From the docs there remain some unanswered questions about the exact semantics. In an advance to make the semantics formally clear we might take the opportunity to extend multiline options in a way that brings us a solution for any case of line continuation.

@dedemorton Your view on this?

tsg:
Argh, sorry, what I meant is:

```
multiline:
  pattern: '\\$'
  negate: false
  match: before
```

I would expect this to give you

```
a
b
c\d\e
f
```

7lima:
:-) That's right now. You've taken the cake!
And I've finally understood why there's a before match option at all!

dedemorton:
Closing this in favor of #2228

PR #2054, opened 2016-07-18
Documentation: Clearer example for include_lines and exclude_lines.

7lima:
The note about include_lines and exclude_lines tells about the *order of execution* of these options. So to make clear that it's not about their *order of occurrence* in the file I switched the two options.

### DIFF
--- a/filebeat/docs/reference/configuration/filebeat-options.asciidoc
+++ b/filebeat/docs/reference/configuration/filebeat-options.asciidoc
@@ -80,8 +80,8 @@ NOTE: If both `include_lines` and `exclude_lines` are defined, Filebeat executes
 
 [source,yaml]
 -------------------------------------------------------------------------------------
- include_lines: ["apache"]
  exclude_lines: ["^DBG"]
+ include_lines: ["apache"]
 -------------------------------------------------------------------------------------
 
 See <<regexp-support>> for a list of supported regexp patterns.


tsg:
@7lima Thanks for the PR. I worry that this change might confuse people. Yes, the order in the configuration file doesn't matter, but having it in the same order in which it is executed is more intuitive, IMO.

CC-ing @dedemorton for a second opinion.

7lima:
@tsg Right, it might confuse them. But they will be able to figure out by the example even if it's just clear at a second look.

Hovever if we don't make it very clear in the doc a lot of users might try to change the order of these options expecting it to change the order of execution. And failing by that might be even more confusing for them at a later time.

7lima:
@tsg @ruflin Can you please help me understand:
- Why did the build fail? It's not caused by my changes, right?
- How do I myself add labels to a PR?

tsg:
The build error looks like a temporary download issue. I triggered it. I think Github only allows collaborators to add/remove labels.

@dedemorton is the expert in docs, I'm sure she'll find a way to clarify the docs here. Note that she is offline this week, though.

7lima:
@tsg Thank you a lot!

So let's continue in this talk when @dedemorton returns.

dedemorton:
@7lima I agree with @tsg. I don't want to show a non-intuitive example. Users who skim content for examples might see the example and think they *can* control the order of execution (which is precisely what we don't want them to believe.) Instead, I've added more wording in PR #2127 to explain that the order of execution is not determined by the order of options in the config file.

PR #2059, opened 2016-07-19
Fix Metricbeat field names for redis aof

### LINKED ISSUE #2027: Metricbeat Redis aof stats are named used
See this [line](https://github.com/elastic/beats/blob/master/metricbeat/module/redis/info/data.go#L49). We need to fix it there, but also in `fields.yml`.

tsg:
It was called `used`, probably due to a copy & paste error. Fixes #2027.


ruflin:
Docs should also be updated through `make generate-json`, but we can do that once before releasing.

PR #2065, opened 2016-07-19
Add some format string support to libbeat

urso:
This PR adds some customizable format-string support to libbeat. This feature is not used anywhere yet, but lays some ground-work for features/improvements requiring format-string support dynamically constructing strings right from events.

**EventFormatString** implements format string support on events
of type common.MapStr.

The concrete event expansion requires the field name enclosed by brackets.
For example: '%{[field.name]}'. Field names can be separated by points or
multiple braces. This format `%{[field.name]}` is equivalent to `%{[field][name]}`.

Default values are given by the colon operator. For example:
`%{[field.name]:default value}`.

**EventFormatString** is build on top of generic **StringFormatter** Compile function.

Compile compiles an input format string into a StringFormatter. The variable
compiler `vc` is invoked for every variable expansion found in the input format
string.

Variable expansion are enclosed in expansion braces `%{<expansion>}`.
The `<expansion>` can contain additional parameters separated by ops
introduced by collons ':'. For example the format string `%{value:v1:?v2}`
will be parsed into variable expansion on `value` with variable ops
`[(":", "v1"), (":?", "v2")]`. It's up to the variable compiler to interpret
content and variable ops.

The back-slash character `\` acts as escape character so the sequence '%{' will be written to the resulting string is `\` is used.

### DIFF
--- a/libbeat/common/fmtstr/formatevents.go
+++ b/libbeat/common/fmtstr/formatevents.go
@@ -1,0 +1,276 @@
+package fmtstr
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"reflect"
+	"strconv"
+	"strings"
+
+	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/logp"
+)
+
+// EventFormatString implements format string support on events
+// of type common.MapStr.
+//
+// The concrete event expansion requires the field name enclosed by brackets.
+// For example: '%{[field.name]}'. Field names can be separated by points or
+// multiple braces. This format `%{[field.name]}` is equivalent to `%{[field][name]}`.
+//
+// Default values are given defined by the colon operator. For example:
+// `%{[field.name]:default value}`.
+type EventFormatString struct {
+	formatter StringFormatter
+	ctx       *eventEvalContext
+	fields    []fieldInfo
+}
+
+type eventFieldEvaler struct {
+	ctx   *eventEvalContext
+	index int
+}
+
+type defaultEventFieldEvaler struct {
+	ctx          *eventEvalContext
+	index        int
+	defaultValue string
+}
+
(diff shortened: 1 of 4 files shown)


libbeat/common/fmtstr/formatevents.go
```diff
@@ -0,0 +1,276 @@
...
+
+		fields = append(fields, path)
+		field = field[idx+1:]
+	}
+
+	path := strings.Join(fields, ".")
+	return path, nil
+}
+
+// TODO: move to libbeat/common?
```
tsg: done :)

urso: not done. `common.MapStr` does not directly support querying a field as string-value. Same method might become use-full when extracting compound keys from events (e.g. think lookup-cache)

urso:
```
$ go test -cover
PASS
coverage: 87.0% of statements
ok  	github.com/elastic/beats/libbeat/common/fmtstr	0.017s
```

tsg:
Will the `%{ .. }` parser be able to be used for other uses cases? I have one specific use case in mind:

For the index pattern, ideally we'd support the [logstash syntax](https://www.elastic.co/guide/en/logstash/current/plugins-outputs-elasticsearch.html#plugins-outputs-elasticsearch-index) `filebeat-%{+YYYY.MM.dd}`, meaning that `+` introduces a timestamp formatter. 

Ideally this would be combinable with the event fields, so can e.g. split in different indices by metricset: `metricbeat-%{[metricset.module]}-%{[metricset.name]}-%{+YYYY.MM.dd}`

urso:
Yeah, I have something similar in mind. This particular use-case is not fully supported yet, but I've got some ideas for future PRs to make supported expressions-types more flexible.

ruflin:
LGTM

PR #2068, opened 2016-07-19
Always add conditionals to all processor actions

urso:
processor registry automatically adds conditional support to registered actions
by wrapping processor constructors with NewConditional.

With this PR every processor action being registered automatically gains support for the `when`-conditional clause without having to any any extra code. This simplifies existing processor actions code, by requiring actions to implement only the required actions instead of check, parsing, applying conditional configurations. Plus it ensures conditionals always work exactly the same for all available actions.

### DIFF
--- a/libbeat/processors/actions/include_fields.go
+++ b/libbeat/processors/actions/include_fields.go
@@ -5,37 +5,25 @@ import (
 	"strings"
 
 	"github.com/elastic/beats/libbeat/common"
-	"github.com/elastic/beats/libbeat/logp"
 	"github.com/elastic/beats/libbeat/processors"
 )
 
-type IncludeFields struct {
+type includeFields struct {
 	Fields []string
-	// condition
-	Cond *processors.Condition
-}
-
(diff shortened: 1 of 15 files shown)


libbeat/processors/actions/include_fields.go
```diff
@@ -53,43 +41,12 @@ func newIncludeFields(c common.Config) (processors.Processor, error) {
...
-	return nil
-}
-
-func (f *IncludeFields) Run(event common.MapStr) (common.MapStr, error) {
-
-	if f.Cond != nil && !f.Cond.Check(event) {
-		return event, nil
-	}
-
+func (f includeFields) Run(event common.MapStr) (common.MapStr, error) {
```
ruflin: I assume the reference `*` was removed intentionally.

ruflin:
Nice addition.

libbeat/processors/actions/include_fields.go
urso: yes, includeFields is basically `[]string`.

PR #2069, opened 2016-07-19
Enhancements in system module

monicasarbu:
This PR is a proposal and includes couple of changes:
- set `type` to `<module>.<metricset>` to be easier to define searches in Kibana
- allow the metricset to define what groups of fields is adding instead of placing all metrics under the `system.<meticSet>. For example: for `system.memory`, `mem` and `swap` information are added (move one level up the swap information).
- use gopsutil (instead of gosigar) to get memory statistics and export more information like `memory.available` instead of `actual`, for OSX/BSD (`memory.active`, `memory.inactive`, `memory.wired`), for Linux (`memory.buffers`, `memory.cached`). In addition `swap.in` and `swap.out` are exported.

This PR comes with a few breaking changes:
- `system.cpu.load` -> `system.load`
- `system.memory.swap` -> `system.swap`
- `filesystem.avail` -> `filesystem.available`
- `type: metricsets` -> `type: <module>.<metricSet>`

Things that need to be updated:
- [ ] docs
- [ ] update the `Fetch()` function of the other modules (except system) to export the group of fields as they are added to event. For example an event returned by `Fetch` should contain `memory`: { "total": x, ...} instead of {"total": x, ...}

Here is an example of the JSON object exported by the new version:
```
{
  "@timestamp": "2016-07-19T22:17:44.304Z",
  "beat": {
    "hostname": "mar.local",
    "name": "mar.local"
  },
  "metricset": {
    "module": "system",
    "name": "memory",
    "rtt": 6427
  },
  "system": {
    "memory": {
      "active": 5930377216,
      "available": 5852270592,
      "free": 359858176,
      "inactive": 5492412416,
      "total": 17179869184,
      "used": {
        "bytes": 11327598592,
        "pct": 0.6594
      },
      "wired": 3681193984
    },
    "swap": {
      "free": 559750,
      "in": 0,
      "out": 0,
      "total": 6144000,
      "used": {
        "bytes": 5584250,
        "pct": 0.9089
      }
    }
  },
  "type": "system.memory"
}
```

### DIFF
--- a/metricbeat/beater/event.go
+++ b/metricbeat/beater/event.go
@@ -8,10 +8,6 @@ import (
 	"github.com/elastic/beats/libbeat/processors"
 )
 
-const (
-	defaultType = "metricsets"
-)
-
 // eventBuilder is used for building MetricSet events. MetricSets generate a
 // data in the form of a common.MapStr. This builder transforms that data into
 // a complete event and applies any Module-level filtering.
@@ -38,7 +34,7 @@ func (b eventBuilder) build() (common.MapStr, error) {
 
 	// Get and remove meta fields from the event created by the MetricSet.
 	indexName := getIndex(event, "")
-	typeName := getType(event, defaultType)
+	typeName := getType(event, b.moduleName+"."+b.metricSetName)
 	timestamp := getTimestamp(event, common.Time(b.startTime))
 
 	// Apply filters.
@@ -51,9 +47,7 @@ func (b eventBuilder) build() (common.MapStr, error) {
 		"type":       typeName,
 
 		common.EventMetadataKey: b.metadata,
-		b.moduleName: common.MapStr{
-			b.metricSetName: event,
-		},
+		b.moduleName:            event,
 		"metricset": common.MapStr{
 			"module": b.moduleName,
 			"name":   b.metricSetName,
--- a/metricbeat/beater/event_test.go
+++ b/metricbeat/beater/event_test.go
@@ -45,7 +45,7 @@ func TestEventBuilder(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	assert.Equal(t, defaultType, event["type"])
+	assert.Equal(t, moduleName+"."+metricSetName, event["type"])
 	assert.Equal(t, common.Time(startTime), event["@timestamp"])
 
 	metricset := event["metricset"].(common.MapStr)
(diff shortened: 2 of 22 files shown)


metricbeat/beater/event.go
```diff
@@ -38,7 +34,7 @@ func (b eventBuilder) build() (common.MapStr, error) {
 
 	// Get and remove meta fields from the event created by the MetricSet.
 	indexName := getIndex(event, "")
-	typeName := getType(event, defaultType)
+	typeName := getType(event, b.moduleName+"."+b.metricSetName)
```
ruflin: This currently applies to `type` and `_type`. I think if we are going to change this, we should find a better name and not use `type`.

metricbeat/beater/event.go
```diff
@@ -51,9 +47,7 @@ func (b eventBuilder) build() (common.MapStr, error) {
 		"type":       typeName,
 
 		common.EventMetadataKey: b.metadata,
-		b.moduleName: common.MapStr{
-			b.metricSetName: event,
-		},
+		b.moduleName:            event,
```
ruflin: Having the namespace of module and metricset here is critical to prevent namespace conflicts. This guarantees to no metricset can conflict with an other. In addition we move the "burden" of doing the namespacing correctly to the person implementing the metricset. -1 on doing this change.

ruflin:
It seems like we need `system.load` as load somehow doesn't fit into `cpu`. Instead of changing the logic how metricbeat works for one metricset, I think we should introduce a `load` metricset instead and taking load out of CPU. SystemLoad always felt somehow awkward in the code being in CPU so an additional metricset load seems to make sense.

+1 on the additional memory stats as long as the switch to gopsutil is not going to reduce the supported platforms.

I added the `in progress` label as I think this PR should not be merged yet.

metricbeat/beater/event.go
ruflin: @tsg I remember you wanted to do some testing here?

ruflin:
@monicasarbu Suggestion for the next steps: There are great things in this PR like the memory part which we should definitively get into Metricbeat. But there is also some controversial stuff like the namespacing. For one not to block the other I suggest to split up this PR into multiple PRs.

metricbeat/beater/event.go
monicasarbu: @ruflin I would say for now to keep `type` and just change the value in Metricbeat, so I can use it in the Kibana dashboards. I think it's nice to have the same name across all Beats, so you can have the same story behind searching and only use different values depending on the Beat. We can also change the name of the field later.

monicasarbu:
Closing it in favor of https://github.com/elastic/beats/pull/2072

PR #2072, opened 2016-07-20
Create a metricset for swap and load

monicasarbu:
This PR includes the following changes:
- Create a swap metricset instead of including swap information in the memory metricset
- Create a load metricset instead of including load information in the cpu metricset 

This PR includes few breaking changes:
- `system.memory.actual.free` is replace with `system.memory.available`
- `system.memory.actual.used` is not needed as it can calculated from `system.memory.total` - `system.memory.available`
- the configuration file changes, meaning that you need to add `swap` and `load` if you want to export these data. They are not part of `memory` and `cpu` anymore.

cc-ed @ruflin @andrewkroh


metricbeat/module/system/load/_meta/docs.asciidoc
```diff
@@ -0,0 +1,11 @@
+=== System CPU Metricset
```
ruflin: Should be load?

metricbeat/module/system/memory/memory.go
```diff
@@ -29,43 +31,39 @@ func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
...
 
-	swap := common.MapStr{
-		"total": swapStat.Total,
-		"used": common.MapStr{
-			"bytes": swapStat.Used,
-			"pct":   swapStat.UsedPercent,
-		},
-		"free": swapStat.Free,
+	// OSX /BSD specific
+	if memStat.Active != 0 {
```
ruflin: Would be nice if we could check for something else then `0`. Not sure if this value could be also 0 and be valid?

metricbeat/module/system/swap/_meta/docs.asciidoc
```diff
@@ -0,0 +1,11 @@
+=== System Memory Metricset
```
ruflin: Swap

metricbeat/module/system/swap/swap.go
```diff
@@ -0,0 +1,57 @@
+// +build darwin freebsd linux openbsd windows
+
+// +build darwin freebsd linux openbsd windows
```
ruflin: Line is duplicated

ruflin:
I really like this change as I think it makes finding the right metrics easier and also simplifies the code.

metricbeat/docs/fields.asciidoc
```diff
@@ -2402,111 +2402,121 @@ Total space (used plus free).
...
-type: half_float
+type: long
 
-The percentage of actual used memory.
+Available memory is what it matters. It is calculated based on the OS. On Linux it is a sum of free memory and caches and buffers. On OSX consists of free and inactive memory.
 
 
 [float]
-== swap Fields
+=== system.memory.active
```
andrewkroh: I would like to have a more detailed description for some of these fields and possibly how the data is collected (e.g. read from /proc/meminfo or same values reported by `vmstat -a`).

> Active — The total amount of buffer or page cache memory, in kilobytes, that is in active use. This is memory that has been recently used and is usually not reclaimed for other purposes.

> Inactive — The total amount of buffer or page cache memory, in kilobytes, that are free and available. This is memory that has not been recently used and can be reclaimed for other purposes.

metricbeat/docs/fields.asciidoc
```diff
@@ -2402,111 +2402,121 @@ Total space (used plus free).
...
 
-This group contains statistics related to the swap memory usage on the system.
+type: long
+
+Specific for OSX/BSD systems.
```
andrewkroh: Linux also has active/inactive values reported in `/proc/meminfo` so this shouldn't be specific to OSX/BSD, should it?

andrewkroh:
I think this is going to break openbsd support which was added to gosigar.

metricbeat/tests/system/test_system.py
```diff
@@ -26,7 +28,10 @@
 
 SYSTEM_FSSTAT_FIELDS = ["count", "total_files", "total_size"]
 
-SYSTEM_MEMORY_FIELDS = ["swap", "actual", "free", "total", "used.bytes", "used.pct"]
+SYSTEM_MEMORY_FIELDS = ["available", "free", "total", "used.bytes", "used.pct", "active", "inactive", "wired",
```
monicasarbu: need to remove "active", "inactive" ...

metricbeat/module/system/cpu/helper.go
```diff
@@ -3,12 +3,8 @@
 package cpu
 
 import (
-	"time"
-
 	"github.com/elastic/beats/libbeat/common"
-	"github.com/elastic/beats/libbeat/logp"
 	"github.com/elastic/beats/metricbeat/module/system"
-	"github.com/elastic/beats/metricbeat/module/system/memory"
```
ruflin: Nice that we get rid of these interdependencies.

metricbeat/etc/fields.yml
```diff
@@ -1447,62 +1464,21 @@
...
-                  description: >
-                    Actual used memory. This value is the "used" memory minus the memory used for disk caches and buffers.
-                    Available only on Unix.
-
-                - name: free
-                  type: long
-                  description: >
-                    Actual available memory. This value is the "free" memory plus the memory used for disk caches and
-                    buffers. Available only on Unix.
+                Free memory. For a human readable value use Available.
```
andrewkroh: "human readable" means to me that it's the same value formatted with units like MB or GB. It's like `df -h`.

Consider saying this "The total amount of free memory in bytes. This value does not include memory consumed by system caches and buffers (see `system.memory.available`)."

metricbeat/module/system/load/load.go
```diff
@@ -0,0 +1,46 @@
+// +build darwin freebsd linux openbsd windows
```
andrewkroh: If load is not available on Windows then remove the `windows` build tag. You'll also want to do the same for any tests.

metricbeat/module/system/load/helper_test.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	"runtime"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGetSystemLoad(t *testing.T) {
+
+	if runtime.GOOS == "windows" {
```
andrewkroh: If you remove the `windows` tag, then you can remove this runtime check.

metricbeat/tests/system/test_system.py
```diff
@@ -93,6 +97,33 @@ def test_cpu_ticks_option(self):
             cpuStats = evt["system"]["cpu"]
             self.assertItemsEqual(self.de_dot(SYSTEM_CPU_FIELDS_ALL), cpuStats.keys())
 
+    @unittest.skipUnless(re.match("(?i)win|linux|darwin|freebsd|openbsd", sys.platform), "os")
+    def test_load(self):
+        """
+        Test cpu system output.
```
andrewkroh: `s/cpu/load/`

metricbeat/module/system/swap/swap.go
```diff
@@ -0,0 +1,47 @@
...
```
andrewkroh: Is this really implemented for Windows or does this always just return error events? If not implemented, remove `windows`.

metricbeat/module/system/memory/_meta/fields.yml
```diff
@@ -13,59 +13,18 @@
...
-            Available swap memory.
-
-        - name: used.pct
-          type: half_float
-          description: >
-            The percentage of used swap memory.
+    - name: available
+      type: long
+      description: >
+        Available memory is what it matters. It is calculated based on the OS. On Linux it is a sum of free memory
```
andrewkroh: I think describing why it matters would be better for the description. Also I think we should mention the units are in bytes (we don't currently do this consistently).

metricbeat/etc/beat.yml
```diff
@@ -27,10 +30,13 @@ metricbeat.modules:
 
     # File system summary stats
     #- fsstat
-
+
```
andrewkroh: Trailing whitespace that can be removed.

andrewkroh:
The system/load package must have a `doc.go` file. Otherwise metricbeat will fail to build on operating systems that don't support `load` because `list.go` imports what appears to be an empty package. This is why the travis build failed. https://travis-ci.org/elastic/beats/jobs/146202233

andrewkroh:
I'm not fully convinced that `swap` is completely separate concept from `memory` to necessitate it's own metricset. For example, most things in the operating system group all of this data together under the umbrella term or "memory" like `/proc/meminfo` and `/sys/fs/cgroup/memory/memory.stat`.

monicasarbu:
@andrewkroh I agree `swap` should be in the same document with memory. That's why my initial [PR](https://github.com/elastic/beats/pull/2069) was to have `system.memory` and `system.swap` in the same document, but this is not possible with the current Metricbeat architecture. So, I am closing this PR as it seems that the best possible option is the current one.

ruflin:
@monicasarbu @andrewkroh What about the `load` part of this PR? I still like that load is a separate metricset as it is a summary over time.

andrewkroh:
Yeah, I also liked the `load` separation. The memory field renaming is also nice.

monicasarbu:
I created another PR to add the`load` metricset and change the name of the memory exported fields. https://github.com/elastic/beats/pull/2101

Issue #2074, opened 2016-07-20
Add index pattern, rename index options


tsg:
Meta issue for transforming the Elasticsearch output index option into an index pattern (#921) and for cleaning up the outputs configuration around indices.

Tasks:

* [x] In the Redis output, rename the `index` option to `key`, deprecating `index` for now. #2077.
* [x] In the file output, remove the `index` option. #2077
* [x] In the elasticsearch output, make `index` accept an index pattern #2119
* [x] Remove the index hack from the publisher code, and use the beatName as appropiate
* [x] Update docs

PR #2077, opened 2016-07-20
Renamed redis.index option to redis.key

tsg:
The "index" setting name has historical reasons, from the times that
all outputs shared the same configuration options. Since the meaning and
contents of the "index" setting are starting to vary per output, it's time
to clean this up. The Redis "index" is now called "key".

The "index" setting still works, but it's deprecated.

Before this change, the index was still somehow special, since it was set to
the beatName outside of the output. This removes this hack, and sets index to
beatName only in the Elasticsearch and Logstash modules.

Part of #2074.

This also removes the `output.file.index` setting, but this setting was never documented, so removing it shouldn't affect anyone. The setting was equivalent to setting `output.file.filename`.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -13,9 +13,12 @@ https://github.com/elastic/beats/compare/v5.0.0-alpha4...master[Check the HEAD d
 ==== Breaking changes
 
 *Affecting all Beats*
+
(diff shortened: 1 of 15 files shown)


urso:
LGTM.

missing change in config.full.yml (in libbeat) plus `$ make update`.

tsg:
@urso Thanks for the review. Configs updated and rebased.

PR #2078, opened 2016-07-21
Index pattern

### LINKED ISSUE #921: Allow configurable time for ES output
[Here](https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-configuration-details.html#elasticsearch-output) we state;
```
# Optional index name. The default is "filebeat" and generates
    # [filebeat-]YYYY.MM.DD keys.
    index: "filebeat"
```

But we don't allow the `YYYY.MM.DD` to be configurable.

This isn't ideal as users cannot do anything but daily indices without putting Logstash in the pipeline.

tsg:
This transforms the fixed index into a pattern, somehow similar to how
Logstash is doing it. However, logstash is using the Joda format, for which
we don't have Go libraries. So instead of writing `filebeat-%{+YYYY.MM.dd}` you
would need to write `filebeat-%{+2006.01.02}` (i.e. layout by example).

Because the Go layouts don't support ISO 8601 weeks, in order to support weekly
indices, the special `isoweek` keyword was introduced, which is the equivalent
of Joda `xxxx.ww`. The layout `filebeat-%{+isoweek}` results in `filebeat-2016.29`
for today.

Part of #2074. Closes #921.

Also, depends on #2077.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -13,9 +13,12 @@ https://github.com/elastic/beats/compare/v5.0.0-alpha4...master[Check the HEAD d
 ==== Breaking changes
 
 *Affecting all Beats*
+
 - Rename the `filters` section to `processors`. {pull}1944[1944]
 - Introduce the condition with `when` in the processor configuration. {pull}1949[1949]
 - The Elasticsearch template is now loaded by default. {pull}1993[1993]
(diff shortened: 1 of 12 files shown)


tsg:
Closing, to be reimplemented on top of the format string support, which already supports timestamps.

PR #2088, opened 2016-07-25
Refactor processors and readers

ruflin:
* Rename all processors to readers
* Rename line struct to Message
* Remove "reader" from all structs inside reader package

No logic changes are done in this change.


ruflin:
@urso We are currently using a mix of reader and processor. I see processor as a superset of reader, means I plan to rename all readers to processors and group them in one package. Any thoughts?

urso:
+1 for consistent naming. At some point I intended to all 'LineReader's to be a processor. On the other, we already got processors (generic?) in libbeat.

ruflin:
Alternatively we could rename everything to a reader? Looking at the readers / processors we have, reader would also be accurate I think.

urso:
I think I'd prefer Reader here, to not get confused with libbeat processors.

ruflin:
Agree. I will migrate everything to reader. How should be call the struct? `Read` ?

filebeat/harvester/reader/encoder.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	"io"
+	"time"
+
+	"github.com/elastic/beats/filebeat/harvester/encoding"
+)
+
+// Encode reader produces lines by reading lines from an io.Reader
+// through a decoder converting the reader it's encoding to utf-8.
+type Encode struct {
```
urso: hm... I think we should name the Decode, Conver or Transform. This reader converts input encoding into utf8

filebeat/harvester/reader/line.go
```diff
@@ -16,13 +16,12 @@ type Line struct {
...
-	outBuffer *streambuf.Buffer
-	inOffset  int // input buffer read offset
-	byteCount int // number of bytes decoded from input buffer into output buffer
-	decoder   transform.Transformer
+	nl         []byte
+	inBuffer   *streambuf.Buffer
+	outBuffer  *streambuf.Buffer
+	inOffset   int // input buffer read offset
+	byteCount  int // number of bytes decoded from input buffer into output buffer
+	decoder    transform.Transformer
```
urso: here we even use the name `decoder`

filebeat/harvester/reader/encoder.go
ruflin: I think Encode is still correct, but we could name it `EncodeUTF8` to make it more clear?

filebeat/harvester/reader/line.go
ruflin: Here I think the mess is also related to https://github.com/elastic/beats/pull/2089 and some tricks ;-) We should probably clean this up in combination with https://github.com/elastic/beats/pull/2089

filebeat/harvester/reader/log.go
```diff
@@ -37,11 +37,11 @@ type LogFileReaderConfig struct {
...
 }
 
-func NewLogFileReader(
+func NewLogFile(
 	fs source.FileSource,
-	config LogFileReaderConfig,
+	config LogFileConfig,
 	done chan struct{},
-) (*logFileReader, error) {
+) (*logFile, error) {
```
urso: exported `NewX` should not return unexported type. Can still be assigned using type inference and even converted to interface, but not be used by name.

filebeat/harvester/reader/reader.go
```diff
@@ -0,0 +1,24 @@
...
+
+	"github.com/elastic/beats/libbeat/common"
+)
+
+// Message represents a reader event with timestamp, content and actual number
+// of bytes read from input before decoding.
+type Message struct {
+	Ts      time.Time     // timestamp the line was read
+	Content []byte        // actual line read
+	Bytes   int           // total number of bytes read to generate the line
```
urso: don't use `line` in comments, but say content. Other file types might differ here for example.

filebeat/harvester/reader/log.go
ruflin: Will change it to be exported.

filebeat/harvester/reader/timeout.go
```diff
@@ -11,40 +11,39 @@ var (
 
 // timeoutProcessor will signal some configurable timeout error if no
 // new line can be returned in time.
-type timeoutProcessor struct {
-	reader  LineProcessor
+type timeout struct {
```
urso: would it make sense to export this type?

filebeat/harvester/reader/encoder.go
urso: nah, rather use Encode or Encoding then EncodeUTF8

filebeat/harvester/reader/reader.go
ruflin: fixed

filebeat/harvester/reader/timeout.go
ruflin: Yes, just to be in line with the other reader.

ruflin: Changed

filebeat/harvester/reader/encoder.go
ruflin: Ah right, we pass the codec, so it could be anything not only uft-8 if configured differently. In this case I would stick with `reader.Encode`?

urso: ok, let's still with it then.

ruflin: We can still renamed it in combination with #2089 if something else becomes obvious.

ruflin:
@urso requested changes pushed.

urso:
LGTM. waiting for travis.

PR #2090, opened 2016-07-25
Harvester Cleanup and FileEvent

ruflin:
* Rename FileEvent to Event
* Remove double information between FileEvent and State (Source, Fileinfo)
* Remove readLine function as not needed anymore

### DIFF
--- a/filebeat/harvester/log.go
+++ b/filebeat/harvester/log.go
@@ -44,7 +44,8 @@ func (h *Harvester) Harvest() {
 
 	logp.Info("Harvester started for file: %s", h.state.Source)
 
-	r, err := h.newLineReader()
+	r, err := h.newLogFileReader()
+
 	if err != nil {
 		logp.Err("Stop Harvesting. Unexpected encoding line reader error: %s", err)
 		return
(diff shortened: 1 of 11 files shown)


filebeat/harvester/log.go
```diff
@@ -63,8 +64,7 @@ func (h *Harvester) Harvest() {
 		default:
 		}
 
-		// Partial lines return error and are only read on completion
-		ts, text, bytesRead, jsonFields, err := readLine(r)
+		message, err := r.Next()
```
ruflin: @urso The behaviour here should be identical to before, but hopefully better readable. An event is still created if Byte == 0 but this is filtered out as a state update only by the publisher. I assume this happens when an event only consists of a newline. So the state will be updated no event will be sent to the output. Alternatively we could also check inside shouldExportLine for bytes == 0.

urso: normally new lines characters are added to byte count, so to have correct offset.

urso: +1 for using struct over 4 variables

urso:
LGTM. waiting for travis.

PR #2091, opened 2016-07-25
Unmarshal JSON numbers as ints by default

### LINKED ISSUE #2038: Filebeat unmarshals JSON integers as float64
Filebeat can unmarshal arbitrary JSON data, and when it unmarshals numbers they are of type float64. If you try to use a conditional filter with `equals` to match against a number read from JSON you get the following error.

`condition.go:240: WARN unexpected type float64 in equals condition as it accepts only integers and strings.`

Filebeat version: 5.0.0 git c54b7c4e161abe1f2f3d6aa181f8d18f575a7dd1

Config:

```
$ cat filebeat.dev.yml 
filebeat.prospectors:
  - paths:
    - "./input.json"
    json.keys_under_root: false

processors:
  - drop_event:
      when:
        and:
          - contains:
              json.http_user_agent: "ELB-HealthChecker"
          - equals:
              json.status: 200

output.console:
  pretty: true

logging.level: debug
```

Input file:
```
$ cat input.json 
{"http_user_agent": "ELB-HealthChecker/1.0", "status": 200}
```

tsg:
The change modifies the unmarshaling logic to try parsing the numbers
in an int64 first, and only on error parse to a float64. In practice,
this means that e.g. 1 will become an int64 and 1.0 will become a
float64.

Fixes #2038

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -48,9 +48,10 @@ https://github.com/elastic/beats/compare/v5.0.0-alpha4...master[Check the HEAD d
 *Topbeat*
 
 *Filebeat*
-- Fix potential data loss between filebeat restarts, reporting unpublished lines as published. {issue}2041[2041]
 
+- Fix potential data loss between filebeat restarts, reporting unpublished lines as published. {issue}2041[2041]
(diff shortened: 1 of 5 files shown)


andrewkroh:
LGTM

PR #2092, opened 2016-07-25
Filbeat harvester copy

ruflin:
Instead of creating a harvester every time a harvester is started, the prospector has only one harvester which is copied and then started. The advantage is that config unpacking and encoding factory are called only once. 

The idea is that also readers will process their own config in the future and should be setup as part of the "New" method to make sure this also happens only once.


ruflin:
Closing as this will tackled at a later stage again during further refactoring of harvester / prospector.

PR #2094, opened 2016-07-26
Reader close

ruflin:
@urso Not finished yet, but is that the direction you were thinking for the Stop method?

### DIFF
--- a/filebeat/harvester/log.go
+++ b/filebeat/harvester/log.go
@@ -50,6 +50,12 @@ func (h *Harvester) Harvest() {
 		return
 	}
 
+	// Waits for harvester channel to be closed to close reader
+	go func() {
+		<-h.done
+		h.reader.Close()
+	}()
+
 	// Always report the state before starting a harvester
 	// This is useful in case the file was renamed
 	if !h.sendStateUpdate() {
@@ -59,7 +65,6 @@ func (h *Harvester) Harvest() {
 	for {
 		select {
 		case <-h.done:
-			h.reader.Stop()
 			return
 		default:
 		}
@@ -292,7 +297,7 @@ func (h *Harvester) newLogFileReader() (reader.Reader, error) {
 	var r reader.Reader
 	var err error
 
-	logInput, err := NewLogFile(h.file, readerConfig, h.done)
+	logInput, err := NewLogFile(h.file, readerConfig)
 	if err != nil {
 		return nil, err
 	}
--- a/filebeat/harvester/log.go
+++ b/filebeat/harvester/log.go
@@ -44,8 +44,7 @@ func (h *Harvester) Harvest() {
 
 	logp.Info("Harvester started for file: %s", h.state.Source)
 
-	r, err := h.newLogFileReader()
-
+	h.reader, err = h.newLogFileReader()
 	if err != nil {
 		logp.Err("Stop Harvesting. Unexpected encoding line reader error: %s", err)
 		return
@@ -60,11 +59,12 @@ func (h *Harvester) Harvest() {
 	for {
 		select {
 		case <-h.done:
+			h.reader.Stop()
 			return
 		default:
 		}
 
-		message, err := r.Next()
+		message, err := h.reader.Next()
 		if err != nil {
 			switch err {
 			case ErrFileTruncate:
--- a/filebeat/harvester/log.go
+++ b/filebeat/harvester/log.go
@@ -67,13 +67,13 @@ func (h *Harvester) Harvest() {
 		message, err := r.Next()
 		if err != nil {
 			switch err {
-			case reader.ErrFileTruncate:
+			case ErrFileTruncate:
 				logp.Info("File was truncated. Begin reading file from offset 0: %s", h.state.Source)
 				h.state.Offset = 0
 				filesTruncated.Add(1)
-			case reader.ErrRemoved:
+			case ErrRemoved:
 				logp.Info("File was removed: %s. Closing because close_removed is enabled.", h.state.Source)
-			case reader.ErrRenamed:
+			case ErrRenamed:
 				logp.Info("File was renamed: %s. Closing because close_renamed is enabled.", h.state.Source)
 			case io.EOF:
 				logp.Info("End of file reached: %s. Closing because close_eof is enabled.", h.state.Source)
@@ -270,11 +270,11 @@ func (h *Harvester) close() {
 	harvesterClosed.Add(1)
 }
 
-func (h *Harvester) newLogFileReaderConfig() reader.LogFileConfig {
+func (h *Harvester) newLogFileReaderConfig() LogFileConfig {
 	// TODO: NewLineReader uses additional buffering to deal with encoding and testing
 	//       for new lines in input stream. Simple 8-bit based encodings, or plain
 	//       don't require 'complicated' logic.
-	return reader.LogFileConfig{
+	return LogFileConfig{
 		CloseRemoved:  h.config.CloseRemoved,
 		CloseRenamed:  h.config.CloseRenamed,
 		CloseInactive: h.config.CloseInactive,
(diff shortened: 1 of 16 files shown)


filebeat/harvester/log_file.go
```diff
@@ -107,6 +107,10 @@ func (r *LogFile) Read(buf []byte) (int, error) {
 	}
 }
 
+func (r *LogFile) Stop() error {
+	return nil
```
urso: let's close file handle here

filebeat/harvester/reader/line.go
```diff
@@ -90,6 +90,10 @@ func (l *Line) Next() ([]byte, int, error) {
 	return bytes, sz, nil
 }
 
+func (r *Line) Stop() error {
+	return nil
```
urso: maybe use io.ReadCloser checking reader at construction time implements Close, otherwise wrap via iotuil .

filebeat/harvester/log.go
```diff
@@ -44,13 +44,18 @@ func (h *Harvester) Harvest() {
...
+	h.reader, err = h.newLogFileReader()
 	if err != nil {
 		logp.Err("Stop Harvesting. Unexpected encoding line reader error: %s", err)
 		return
 	}
 
+	// Waits for harvester channel to be closed to close reader
+	go func() {
+		<-h.done
+		h.reader.Close()
```
urso: what if harvester just returns? Who is calling reader.Close() in this case?

filebeat/harvester/log.go
```diff
@@ -292,12 +297,12 @@ func (h *Harvester) newLogFileReader() (reader.Reader, error) {
 	var r reader.Reader
 	var err error
 
-	fileReader, err := reader.NewLogFile(h.file, readerConfig, h.done)
+	logInput, err := NewLogFile(h.file, readerConfig)
```
urso: does `LogFile` reader close file handle? Do we have 2 code-pathes potentially closing h.file now?

urso: if we make LogFile to close h.file, what do we need h.file for then?

filebeat/harvester/log_file.go
```diff
@@ -156,8 +161,15 @@ func (r *LogFile) errorChecks(err error) error {
 }
 
 func (r *LogFile) wait() {
+
 	// Wait before trying to read file wr.ch reached EOF again
-	time.Sleep(r.backoff)
+	timer := time.NewTimer(r.backoff)
```
urso: can we use `common.Backoff` from libbeat here?

filebeat/harvester/reader/encode.go
```diff
@@ -16,20 +16,24 @@ type Encode struct {
 // NewEncode creates a new Encode reader from input reader by applying
 // the given codec.
 func NewEncode(
-	in io.Reader,
+	in io.ReadCloser,
```
urso: just asking for io.Reader might be ok here. In stdlib sometimes this pattern is used:

```
// r is io.Reader
rc, ok := r.(io.ReadCloser)
if !ok {
  rc = ioutil.NopCloser(r)
}

// use rc here
```

That is, Close become optional. but using `io.ReadCloser` is ok too.

filebeat/harvester/reader/line.go
```diff
@@ -13,7 +13,7 @@ import (
 // using the configured codec. The reader keeps track of bytes consumed
 // from raw input stream for every decoded line.
 type Line struct {
-	rawInput   io.Reader
+	rawInput   io.ReadCloser
```
urso: same here, we can make `Close` optional using mentioned pattern.

something like

```
func ensureCloser(r io.Reader) io.ReadCloser {
    if rc, ok := r.(io.ReadCloser) {
      return rc
    }
    return iotuil.NopCloser(r)
}
```

filebeat/harvester/log.go
ruflin: If return is used, no need to close the readers. The close is only to unblock potential hanging readers.

filebeat/harvester/log_file.go
ruflin: Yes, but other PR. Already on the list of https://github.com/elastic/beats/issues/1913

filebeat/harvester/reader/line.go
ruflin: Currently that would only simplify the tests, right? So not sure if it brings a benefit to loosen the requirement.

filebeat/harvester/log.go
ruflin: I have to investigate that further. It somehow also relates tot he previous point.

Questions for me:
* What happens if Close is called twice?
* Readers are closed anyways if harvester is ended. Do we to close them anyways?

filebeat/harvester/log.go
urso: when is h.done closed? Potentially leaking go-routine here?

filebeat/harvester/reader/line.go
urso: it's more of an optional change. users can always opt to wrap io.Reader with ioutil.NopClosed if required.

filebeat/harvester/reader/line.go
```diff
@@ -90,12 +98,22 @@ func (l *Line) Next() ([]byte, int, error) {
 	return bytes, sz, nil
 }
 
+func (r *Line) Close() error {
+	close(r.done)
+	return r.rawInput.Close()
```
urso: the rawInput is the file handle from harvester, right? That is (*Harvester).Close will attempt to close the file a second time?

urso: still see need to close the reader here, in order to break the advance loop.

filebeat/harvester/reader/multiline.go
```diff
@@ -112,8 +114,18 @@ func (mlr *Multiline) Next() (Message, error) {
...
+	return r.reader.Close()
+}
+
 func (mlr *Multiline) readFirst() (Message, error) {
 	for {
+		select {
+		case <-mlr.done:
+			return Message{}, ErrReaderStopped
+		default:
+		}
```
urso: we need to test the done channel here, with us closing the underlying reader anyways? We need the done channel at all?

filebeat/harvester/reader/reader.go
```diff
@@ -1,11 +1,14 @@
 package reader
 
 import (
+	"errors"
 	"time"
 
 	"github.com/elastic/beats/libbeat/common"
 )
 
+var ErrReaderStopped = errors.New("reader was stopped")
```
urso: s/stopped/closed/

filebeat/harvester/reader/multiline.go
```diff
@@ -137,6 +149,11 @@ func (mlr *Multiline) readFirst() (Message, error) {
 
 func (mlr *Multiline) readNext() (Message, error) {
 	for {
+		select {
+		case <-mlr.done:
+			return Message{}, ErrReaderStopped
+		default:
+		}
```
urso: TODO: have reader check the error code from `mlr.reader.Next()`. Let's return error immediately instead of pushing half-finished event before pushing actual error.

filebeat/harvester/reader/timeout.go
```diff
@@ -58,6 +60,8 @@ func (p *Timeout) Next() (Message, error) {
 	}
 
 	select {
+	case <-p.done:
+		return Message{}, ErrReaderStopped
```
urso: with error code being forwarded from `p.reader`, I wonder if we need the channel here.

filebeat/harvester/log.go
ruflin: Yes, leaking go routine. Will have to find an other option.

filebeat/harvester/reader/line.go
ruflin: +1 on the NopClosed option as so far almost all readers needed the close method. The main reason is that readers can be chained.

filebeat/harvester/reader/line.go
ruflin: What do you mean by the second point? What you say is that we should keep `r.rawInput.Close()` here and not have it done by the harvester? I did some quick tests and it doesn't seem to be an issue to call close multiple times.

filebeat/harvester/reader/multiline.go
ruflin: You are right, the channel should get obsolete as Next() will return an error anyways.

About your second comment: Agree. Will have a look, but potentially in a second PR to make sure this PR has in general no "functional" changes. In case we break something with the multiline change above, it is easier to find.

ruflin:
Here is the start of an alternative implementation: https://github.com/elastic/beats/pull/2104 Keeping PR open so we can compare the two:

ruflin:
Closing as https://github.com/elastic/beats/pull/2104 is the better approach. Also reader is now always returning an error https://github.com/elastic/beats/commit/33e7e3f77abe1ed5b8b217fce4d4753fcb3b6ba5#diff-c93710bd58486b7f46dcf6e905c0df74L69

Issue #2000, opened 2016-07-08
Add option to output without the JSON wrapper


tsg:
The log lines generated by Filebeat are wrapped automatically in a JSON object. For some outputs (console, file, kafka, redis) it could be an option to just send the original log lines with wrapping them. The way this could be implemented is by allowing at the output level to build a message via a format string. 

For example:

```
output.kafka:
  format: "%{@timestamp} %{message}"
```

This would be helpful during development (console and file outputs) as well as in some deployment situations.

treff7es:
I like this generic solution and happy to help implement it.

PR #2096, opened 2016-07-26
Add option to output without the JSON wrapper

treff7es:
Initial implementation Add option to output without the JSON wrapper feature based on #2000 .
Please check if you like it in this way.


treff7es:
Please, hold the review, I found bug in it which I will need to fix. I will update this one when it is fixed.

urso:
Set to `in progress` so it doesn't get merged by accident

PR #2098, opened 2016-07-26
Closing readers based on errors

ruflin:


filebeat/harvester/log_file.go
```diff
@@ -157,7 +139,13 @@ func (r *LogFile) errorChecks(err error) error {
 
 func (r *LogFile) wait() {
 	// Wait before trying to read file wr.ch reached EOF again
-	time.Sleep(r.backoff)
+	timer := time.NewTimer(r.backoff)
+
+	select {
+	case <-r.done:
+		return
+	case <-timer.C:
```
urso: one can use `<-time.After(r.backoff)` here

ruflin: changed

filebeat/harvester/log_file.go
```diff
@@ -156,8 +138,12 @@ func (r *LogFile) errorChecks(err error) error {
...
 
 func (r *LogFile) wait() {
-	// Wait before trying to read file wr.ch reached EOF again
-	time.Sleep(r.backoff)
+	// Wait before trying to read file again. File reached EOF.
+	select {
+	case <-r.done:
+		return
+	case <-time.After(r.backoff):
+	}
```
urso: where is `wait` being used? We need to distinguish be return value if done channel was closed or wait was completed?

ruflin: It does not matter, as the first thing after `wait()` returns is that it checks again if the channel was closed.

PR #2099, opened 2016-07-26
Cleanup Multiline Code

ruflin:
* Document functions
* Rename and split up some of the functions for simplicity

No changes to the logic were made.

### DIFF
--- a/filebeat/harvester/reader/multiline.go
+++ b/filebeat/harvester/reader/multiline.go
@@ -49,7 +49,7 @@ const (
 type matcher func(last, current []byte) bool
 
 var (
-	errMultilineTimeout = errors.New("multline timeout")
+	sigMultilineTimeout = errors.New("multline timeout")
 )
 
 // NewMultiline creates a new multi-line reader combining stream of
@@ -93,7 +93,7 @@ func NewMultiline(
 	}
 
 	if timeout > 0 {
-		r = NewTimeout(r, errMultilineTimeout, timeout)
+		r = NewTimeout(r, sigMultilineTimeout, timeout)
 	}
 
 	mlr := &Multiline{
@@ -114,34 +114,35 @@ func (mlr *Multiline) Next() (Message, error) {
 
 func (mlr *Multiline) readFirst() (Message, error) {
 	for {
-		p, err := mlr.reader.Next()
-		if err == nil {
-			if p.Bytes == 0 {
+		message, err := mlr.reader.Next()
+		if err != nil {
+			// no lines buffered -> ignore timeout
+			if err == sigMultilineTimeout {
 				continue
 			}
 
-			mlr.startNewLine(p)
-			mlr.state = (*Multiline).readNext
-			return mlr.readNext()
+			// pass error to caller (next layer) for handling
+			return message, err
 		}
 
-		// no lines buffered -> ignore timeout
-		if err == errMultilineTimeout {
+		if message.Bytes == 0 {
 			continue
 		}
 
-		// something is wrong here
-		return p, err
+		// Start new multiline event
+		mlr.clearBuffer()
+		mlr.resetBuffer(message)
+		mlr.setState((*Multiline).readNext)
+		return mlr.readNext()
 	}
 }
 
 func (mlr *Multiline) readNext() (Message, error) {
 	for {
-		p, err := mlr.reader.Next()
-
+		message, err := mlr.reader.Next()
 		if err != nil {
 			// handle multiline timeout signal
-			if err == errMultilineTimeout {
+			if err == sigMultilineTimeout {
 				// no lines buffered -> ignore timeout
 				if mlr.numLines == 0 {
 					continue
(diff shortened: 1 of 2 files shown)


filebeat/harvester/reader/multiline.go
```diff
@@ -114,31 +114,34 @@ func (mlr *Multiline) Next() (Message, error) {
 
 func (mlr *Multiline) readFirst() (Message, error) {
 	for {
-		p, err := mlr.reader.Next()
-		if err == nil {
-			if p.Bytes == 0 {
+		message, err := mlr.reader.Next()
+		if err != nil {
+			// no lines buffered -> ignore timeout
+			// TODO: Does this only happen on backoff? Normally is EOF?
```
urso: we do not care about backoff or any behavior of underlying reader at all in multiline-reader. `errMultilineTimeout` is always triggered if no line and no error is returned by underlying reader for a duration of `timeout`. The errMultilineTimeout is internal to multiline reader in order to distinguish the timer triggering from actual errors. So to say, errMultilineTimeout is no error(!), but a signal saying: no return from underlying reader for N seconds.

filebeat/harvester/reader/multiline.go
```diff
@@ -114,31 +114,34 @@ func (mlr *Multiline) Next() (Message, error) {
...
+			if err == errMultilineTimeout {
 				continue
 			}
 
-			mlr.startNewLine(p)
-			mlr.state = (*Multiline).readNext
-			return mlr.readNext()
+			// something is wrong here
+			return message, err
```
urso: maybe add to comment: pass error to caller (next layer).

filebeat/harvester/reader/multiline.go
ruflin: Oh, that comment was meant for myself. Should remove it. Thanks also for the details. Probably we should "rename" the error?

filebeat/harvester/reader/multiline.go
```diff
@@ -114,31 +114,34 @@ func (mlr *Multiline) Next() (Message, error) {
...
 		}
 
-		// no lines buffered -> ignore timeout
-		if err == errMultilineTimeout {
+		// Empty starting lines are skipped
+		if message.Bytes == 0 {
```
urso: `message.Bytes` normally contains number of newline characters. A message with `message.Bytes==0` is more like a zero/empty message. Question: what makes an empty line?

an empty line and a zero-message both have `len(msg.Content) == 0`. But empty line has `msg.Bytes>0`, but a zero message has `msg.Bytes==0`.

ruflin: Agree. What is not clear to me is the circumstance when we would get such an empty message?

filebeat/harvester/reader/multiline.go
```diff
@@ -149,86 +152,103 @@ func (mlr *Multiline) readNext() (Message, error) {
...
+	mlr.ts = m.Ts
+	mlr.fields = m.Fields
 }
 
-func (mlr *Multiline) pushLine() Message {
-	content := mlr.content
-	sz := mlr.readBytes
-	fields := mlr.fields
+// flush writes the existing content into the returned message and resets all reader variables.
+func (mlr *Multiline) flush() Message {
```
urso: maybe name `finalize` or `finish`. `flush` makes me assume we push something instead of returning the final event.  (`pushLine` was pretty bad offender too),

filebeat/harvester/reader/multiline.go
```diff
@@ -149,86 +152,103 @@ func (mlr *Multiline) readNext() (Message, error) {
...
+		Bytes:   mlr.readBytes,
+	}
+
+	mlr.reset()
+
+	return m
+}
+
+// reset resets reader variables
+func (mlr *Multiline) reset() {
```
urso: these reader variable are basically the reader buffer state. Maybe name `resetBuffer`?

filebeat/harvester/reader/multiline.go
```diff
@@ -241,9 +261,11 @@ func (mlr *Multiline) addLine(m Message) {
 	}
 
 	space := mlr.maxBytes - sz
-	spaceLeft := (mlr.maxBytes <= 0 || space > 0) &&
-		(mlr.maxLines <= 0 || mlr.numLines < mlr.maxLines)
-	if spaceLeft {
+
+	bytesLeft := (mlr.maxBytes <= 0 || space > 0)
+	linesLeft := (mlr.maxLines <= 0 || mlr.numLines < mlr.maxLines)
```
urso: maybe use names indicating we checking something. `XLeft` sounds like numeric value (got a little confused by if condition).

filebeat/harvester/reader/multiline.go
```diff
@@ -260,8 +282,14 @@ func (mlr *Multiline) addLine(m Message) {
...
 
-func (mlr *Multiline) reset() {
-	mlr.state = (*Multiline).readFirst
+// reset sets state of the reader to readFirst
+func (mlr *Multiline) resetState() {
+	mlr.setState((*Multiline).readFirst)
+}
+
+// setState sets state to the given function
+func (mlr *Multiline) setState(next func(mlr *Multiline) (Message, error)) {
```
urso: we really need this?

let's introduce `type readerState func(mlr *Multiline) (Message, error)`

urso: let's change state handlers from methods to functions: 

`func (m *Multiline) readX() (Message, error)` to `func readX(m *Multiline) (Message, error)`. Both are equivalent when used as function pointers, but we can write `mlr.state = readFirst` instead of `mlr.state = (*Multiline).readFirst`.

filebeat/harvester/reader/multiline.go
```diff
@@ -149,86 +152,103 @@ func (mlr *Multiline) readNext() (Message, error) {
...
 
-func (mlr *Multiline) startNewLine(l Message) Message {
-	retLine := mlr.pushLine()
-	mlr.addLine(l)
-	mlr.ts = l.Ts
-	mlr.fields = l.Fields
-	return retLine
+// inits the reader with the given message. It is recommend to either
+// run reset or flush before.
+func (mlr *Multiline) init(m Message) {
```
urso: I think I did like startNewLine better. `init` always sounds like a module initializer. Alternatives: `initNextLine` or `initNextMessage` ?

filebeat/harvester/reader/multiline.go
urso: let's add descriptive code comment where `errMultilineTimeout` is defined. Alternative names `sigMultilineTimeout` or `sigReaderTimeout` or `sigBufferTimeout`?

filebeat/harvester/reader/multiline.go
urso: empty or zero message?

No idea who is creating a zero message. Potentially some underlying reader can push whatever it wants, though.

ruflin: I was referring to `message.Bytes == 0`. What I try to say is that I'm not sure if we even have to catch this "error". But this kind of disussions should go into https://github.com/elastic/beats/pull/2104 For now I will just remove the comment (no comment) so it is the same as before.

filebeat/harvester/reader/multiline.go
ruflin: `finalize` and `finish` also both sound like they don't return something. I like `flush()` because it somehow implies the reset afterwards, but I agree that it is not obvious where it flushes to ... Alternatives: `fetch`, `consume` ...?

filebeat/harvester/reader/multiline.go
ruflin: Both works for me, as resetting the buffer is kind of resetting the reader. Will rename it to `resetBuffer `

filebeat/harvester/reader/multiline.go
ruflin: Agree. What about `maxLinesReached`, `maxBytesReached`?

filebeat/harvester/reader/multiline.go
ruflin: This would make assigning the state cleaner but it would make it less obvious that the state methods are part of multiline.

I introduce setState and resetState as it would allow to add additional steps in case this is needed.

filebeat/harvester/reader/multiline.go
ruflin: The reasons I went with init is that it initialises timestamp and fields. What about `initBuffer`?

filebeat/harvester/reader/multiline.go
ruflin: I like the `sigMultilineTimeout` naming.

filebeat/harvester/reader/multiline.go
urso: it's no error. multline doesn't know care about underlying reader/layer. But messages with Bytes==0 are not really interesting to multiline reader.

But: support for fields to message has been added much later. So, what shall we do if fields are still set, but Bytes==0? Maybe handling should be changed to store fields of message if Bytes==0. One can even consider passing this message along if buffer is still empty.

filebeat/harvester/reader/multiline.go
urso: to me `finalize` indicates, we finished collecting the lines and finalize will create the final message to be reported.

filebeat/harvester/reader/multiline.go
ruflin: changed

filebeat/harvester/reader/multiline.go
ruflin: Lets go with finalize :-)

filebeat/harvester/reader/multiline.go
urso: one can also do `var readNext = (*Multiline).readNext` and reuse `readNext` on assignment.

urso: I don't really like the long syntax including the full type.

filebeat/harvester/reader/multiline.go
ruflin: Lets continue this in #2104

filebeat/harvester/reader/multiline.go
urso: don't like `initBuffer`, as method is called multiple times to prepare for the next potential multiline message.

filebeat/harvester/reader/multiline.go
ruflin: Should we go with 3 variables?
```
readFirst
readNext
readFailed
```

We still have the ugly syntax but somehow hidden

filebeat/harvester/reader/multiline.go
ruflin: `loadBuffer`?

andrewkroh: > method is called multiple times to prepare for the next potential multiline message

sounds like some kind of "reset" method

PR #2101, opened 2016-07-26
Few changes to the metricbeat exported fields

monicasarbu:
The PR includes the following changes:

- [x] create a `load` metricset in the system module
- [x] `system.memory.actual.free` is replaced with `system.memory.available`
- [x] `system.memory.actual.used` is not needed as it can calculated from `system.memory.total`        - [ ] `system.memory.available`
- [x] `system.filesystem.avail` is replaced with `system.filesystem.available` 
- [x] add `system.load.norm.*`


metricbeat/module/system/load/load.go
```diff
@@ -0,0 +1,70 @@
...
+		return nil, errors.Wrap(err, "load statistics")
+	}
+
+	event := common.MapStr{
+		"avg": common.MapStr{
+			"1":  system.Round(loadStat.Load1, .5, 4),
+			"5":  system.Round(loadStat.Load5, .5, 4),
+			"15": system.Round(loadStat.Load15, .5, 4),
+		},
+		"norm": common.MapStr{
```
ruflin: I had to look up what the difference is between avg and norm here. As far as I know, the periods are in minutes. So it could be for `avg`: `system.load.minute(s).1`. Not sure how per core would fit into this model.

Alternatively what about `system.load.1` and `system.load.core.1` ?

metricbeat/module/system/cpu/helper.go
```diff
@@ -158,7 +159,7 @@ func (cpu *CPU) AddCpuPercentageList(t2 []CpuTimes) {
 }
 
 func (cpu *CPU) GetSystemStats() (common.MapStr, error) {
-	loadStat, err := GetSystemLoad()
+	loadStat, err := load.GetSystemLoad()
```
ruflin: do we still need that here?

metricbeat/module/system/load/load.go
ruflin: Thinking about it again, `norm` is potentially better then I initially thought. it means, all the values here will be between 0-1 ? Third option: `system.load.1` and `system.load.norm.1`. I'm curious if we should ship the number of cores as part of load, so the calculations to "normalise" could happen on the kibana side.

metricbeat/module/system/cpu/_meta/data.json
```diff
@@ -1,55 +1,41 @@
...
+  },
+  "metricset": {
+    "module": "system",
+    "name": "cpu",
+    "rtt": 29
+  },
+  "system": {
+    "cpu": {
+      "idle": {
+        "pct": 0.908
```
ruflin: Seems like we should enable the ticks for the generator just to have the full document with all options

metricbeat/module/system/cpu/helper_test.go
```diff
@@ -63,19 +62,3 @@ func TestCpuPercentage(t *testing.T) {
 	assert.Equal(t, cpu2.UserPercent, 0.9502)
 	assert.Equal(t, cpu2.SystemPercent, 0.0448)
 }
-
-func TestGetSystemLoad(t *testing.T) {
```
ruflin: Is this test not needed anymore?

ruflin:
@monicasarbu For `system.memory.actual.used is not needed as it can calculated from system.memory.total - [ ] system.memory.available` it seems we need 2 out of the three. My initial thought was that the total should be the one that should be calculated and `used` and `available` should be shipped. The results is the same for me it just feels more natural that "total" would be a calculated value. What were your thoughts here?

CHANGELOG.asciidoc
```diff
@@ -21,6 +21,9 @@ https://github.com/elastic/beats/compare/v5.0.0-alpha4...master[Check the HEAD d
 - The undocumented file output `index` setting was removed. Use `filename` instead. {pull}2077[2077]
 
 *Metricbeat*
+- Create a separate metricSet for load under the system module {pull}2101[2101]
```
ruflin: It should also be mentioned, that load was removed from CPU.

metricbeat/module/system/cpu/helper.go
monicasarbu: This function is not used anymore, we could just remove it.

metricbeat/module/system/cpu/helper_test.go
monicasarbu: I moved it to the load directory.

metricbeat/module/system/load/load.go
monicasarbu: That's a good idea, to keep `system.load.1`. I will do the changes.

CHANGELOG.asciidoc
monicasarbu: done!

metricbeat/module/system/filesystem/_meta/data.json
```diff
@@ -1,29 +1,28 @@
...
-            "used": {
-                "bytes": 15242084352,
-                "pct": 0.2405
-            }
-        }
-    },
-    "type": "metricsets"
-}
\ No newline at end of file
+ {
```
andrewkroh: The previous example data had better example values. The new ones are all 0.

andrewkroh:
LGTM. I just left a minor comment that can be addressed at another time if you like.

PR #2104, opened 2016-07-27
Change multiline

ruflin:
This change makes the following assumption:
- In case of a new mulitline event an error is found (including timeout signal) the reader returns the error to be handled by the caller
- Empty events not handled and not seens as an error. Must be handled on a caller level
- Partial content is an error. Next time the reader is started, it must start from scarch again. Partial events are only returned after timeout.
- Error are always directly returned by the reader
- Number of lines == 0 means content is "". Has to be handled by the caller.

This changes the existing behavior that on error partial events are returned. This is now only the case after timeout. EOF by default acts like timeout=0. I would suggest to change the timeout to 0 by default (disabled) and have the EOF check. If timeout > 0 the EOF check would be disabled.

### DIFF
--- a/filebeat/harvester/reader/multiline.go
+++ b/filebeat/harvester/reader/multiline.go
@@ -3,6 +3,7 @@ package reader
 import (
 	"errors"
 	"fmt"
+	"io"
 	"regexp"
 	"time"
 )
@@ -109,28 +110,20 @@ func (mlr *Multiline) Next() (Message, error) {
 }
 
 func (mlr *Multiline) readFirst() (Message, error) {
-	for {
-		message, err := mlr.reader.Next()
-		if err != nil {
-			// no lines buffered -> ignore timeout
-			if err == sigMultilineTimeout {
-				continue
-			}
-
+	message, err := mlr.reader.Next()
+	if err != nil {
+		// no lines buffered -> ignore timeout
+		if err != sigMultilineTimeout {
 			// pass error to caller (next layer) for handling
 			return message, err
 		}
-
-		if message.Bytes == 0 {
-			continue
-		}
-
-		// Start new multiline event
-		mlr.clear()
-		mlr.load(message)
-		mlr.setState((*Multiline).readNext)
-		return mlr.readNext()
 	}
+
+	// Start new multiline event
+	mlr.clear()
+	mlr.load(message)
+	mlr.setState((*Multiline).readNext)
+	return mlr.readNext()
 }
 
 func (mlr *Multiline) readNext() (Message, error) {
@@ -138,11 +131,11 @@ func (mlr *Multiline) readNext() (Message, error) {
 		message, err := mlr.reader.Next()
 		if err != nil {
 			// handle multiline timeout signal
-			if err == sigMultilineTimeout {
+			if err == sigMultilineTimeout || err == io.EOF {
 				// no lines buffered -> ignore timeout
-				if mlr.numLines == 0 {
-					continue
-				}
+				//if mlr.numLines == 0 {
+				//	continue
+				//}
 
 				// return collected multiline event and
 				// empty buffer for new multiline event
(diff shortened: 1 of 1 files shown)


filebeat/harvester/reader/multiline.go
```diff
@@ -6,6 +6,8 @@ import (
 	"regexp"
 	"time"
 
+	"io"
```
urso: what's up with the additional newline?

filebeat/harvester/reader/multiline.go
```diff
@@ -113,40 +115,40 @@ func (mlr *Multiline) Next() (Message, error) {
...
+		// no lines buffered -> ignore timeout
+		//if err == sigMultilineTimeout {
+		//	continue
+		//}
 
-			// pass error to caller (next layer) for handling
-			return message, err
-		}
+		// pass error to caller (next layer) for handling
+		return message, err
```
urso: what if message is not 'empty' ?

filebeat/harvester/reader/multiline.go
```diff
@@ -113,40 +115,40 @@ func (mlr *Multiline) Next() (Message, error) {
...
 
-		// Start new multiline event
-		mlr.clearBuffer()
-		mlr.resetBuffer(message)
-		mlr.setState((*Multiline).readNext)
-		return mlr.readNext()
-	}
+	// Start new multiline event
+	mlr.clearBuffer()
+	mlr.resetBuffer(message)
```
urso: doesn't resetBuffer imply clearBuffer?

Why don't we clear the buffer when finalizing the message? 

Just minor details.

filebeat/harvester/reader/multiline.go
```diff
@@ -113,40 +115,40 @@ func (mlr *Multiline) Next() (Message, error) {
...
-			if err == sigMultilineTimeout {
-				continue
-			}
+	//for {
+	message, err := mlr.reader.Next()
+	if err != nil {
+		// no lines buffered -> ignore timeout
+		//if err == sigMultilineTimeout {
+		//	continue
+		//}
```
urso: so we pass sigMultilineTimeout to reader using Multiline? sigMultilineTimeout is internal to Multiline reader and should not escape the read methods.

filebeat/harvester/reader/multiline.go
```diff
@@ -113,40 +115,40 @@ func (mlr *Multiline) Next() (Message, error) {
...
+	//}
 }
 
 func (mlr *Multiline) readNext() (Message, error) {
 	for {
 		message, err := mlr.reader.Next()
 		if err != nil {
 			// handle multiline timeout signal
-			if err == sigMultilineTimeout {
+			if err == sigMultilineTimeout || err == io.EOF {
```
urso: When pushing the buffer on io.EOF it must be very clear, no new line will ever be read from said file. Otherwise we've the chance of pushing a partial event

Does bottom-most log file reader ever return EOF?

filebeat/harvester/reader/multiline.go
```diff
@@ -113,40 +115,40 @@ func (mlr *Multiline) Next() (Message, error) {
...
 				// no lines buffered -> ignore timeout
-				if mlr.numLines == 0 {
-					continue
-				}
+				//if mlr.numLines == 0 {
+				//	continue
+				//}
```
urso: io.EOF and sigMultilineTimeout should still be treated somewhat different. sigMultilinetimeout should never ever be returned from Multiline read methods, but io.EOF must be forwarded. This code block as is will absorb io.EOF as is?

filebeat/harvester/reader/multiline.go
```diff
@@ -156,37 +158,37 @@ func (mlr *Multiline) readNext() (Message, error) {
...
-					return Message{}, err
-				}
-
-				// lines buffered, return multiline and error on next read
-				m := mlr.finalize()
-				mlr.err = err
-				mlr.setState((*Multiline).readFailed)
-				return m, nil
-			}
+			//if message.Bytes == 0 {
```
urso: is there a chance a reader can return a valid message + some error? If yes, the message must be handled.

filebeat/harvester/reader/multiline.go
```diff
@@ -156,37 +158,37 @@ func (mlr *Multiline) readNext() (Message, error) {
...
+			//	return m, nil
+			//}
+
+			// no match, return current multiline and retry with current line on next
 			// call to readNext awaiting the error being reproduced (or resolved)
 			// in next call to Next
 			m := mlr.finalize()
 			mlr.resetBuffer(message)
-			return m, nil
+			return m, err
```
urso: this line make clear, a reader can return a valid message plus error at the same time? Here potentially partial event will be pushed? Is it valid to reuse multiline reader after error?

filebeat/harvester/reader/multiline.go
ruflin: fixed

filebeat/harvester/reader/multiline.go
ruflin: Agree, we should not return the signal and just handle it as it would be a normal "message"

filebeat/harvester/reader/multiline.go
ruflin: I think it is up to the caller to decide what to do with "half" messages. One thing I'm not sure about is the state of Multiline. The state of multiline should be before the error.

filebeat/harvester/reader/multiline.go
ruflin: Agree, reset should be enough. I removed it.

filebeat/harvester/reader/multiline.go
ruflin: About io.EOF: As suggested in the PR description, I suggest timeout to be 0 by default which has kind of the same affect. It is based on the assumption that one multiline event is flushed at once to disk. 

You are right that the log_file reader does not return the io.EOF by default. It is returned if the reader is not Continutable or CloseEOF is set. 

The problem about setting timeout to 0 is that it will directly return the signal after the first line it read. Instead it should only send a signal if there is no more data.

filebeat/harvester/reader/multiline.go
ruflin: That is a BC change but I would say yes. It should be up to the caller to decide what to do with the partial message. If multiline reader returns an error and is started again, it should continue at the position before the error.

ppf2:
As part of this effort, perhaps we can also [add a debug logging entry](https://github.com/elastic/beats/blob/04cf26dbc1513cdb1cef888664feff5b68f8569a/filebeat/harvester/reader/multiline.go#L146) for when the multiline timeout is hit for debugging purposes? :)

ruflin:
Closing this for now as in the meantime lots of change have happened to the multiline. 

@urso We should still have a discussion on what happens on partial events and if we should improve / modify the multiline reader. But as there were recently no issues, it does not seem urgent.

@ppf2 Could you open an issue for the log entry so we don't forget about it?

PR #2087, opened 2016-07-24
Generic Event String Selector

urso:
preparing for #2064, implement documented logic for customizable keys.

Differs from #2064 that a single-keyword and a multi-keyword can be configured. If both given, the single-keywork is used as default (last selector in list) if all others do fail.

e.g.

```
output.elasticsearch.pipelines:
- pipeline: "%{[key1]}"
- pipeline: "default"
```

will be equivalen to:

```
output.elasticsearch.pipelines:
- pipeline: "%{[key1]}"

output.elasticsearch.pipeline: "default"
```

On the other hand:
```
output.elasticsearch.pipeline: "pipeline"
```

is syntactic sugar for 

```
output.elasticsearch.pipelines:
- pipeline: "pipeline"
```

Being generic, functionality is re-usable for es-index, logstash-index, redis-key, kafka-topic and others.

Issue #2064, opened 2016-07-19
Elasticsearch ingest node pipeline mapping


urso:
Elasticsearch event pipeline can currently only configured via `output.elasticsearch.parameters.pipeline`. Improve pipeline configuration+selection by:

- dynamically choose ingest pipeline per event
- add support for multiple pipelines
- define optional mapping from event to pipeline name
- support for conditionals to select pipeline on more complicated rules

Introduce new setting `output.elasticsearch.pipelines` using an array of pipeline mapping rules. The first rule matching defines the pipeline name to use. If no rule matches, no pipeline will be used for said event:

```
output.elasticsearch.pipelines:
- rule1
- rule2
```

- Rules support conditionals using the `when` key-word (like conditionals support in libbeat processors).
- The pipeline name is set using the `pipeline`-setting. It can be set by events using format strings.
- additional name mappings (based on `pipeline`-setting) requires an additional name mapping table on via `mappings`-setting. The `default`-setting is used, if the pipeline is unknown to the mapping table.
- A pipeline rule only consisting of `pipeline`-keyword matches only if format-string can be applied.

**Sample configurations**:

- read pipeline directly from event `fields.pipeline` using format string (pipelines names can be composed from multiple fields when using format strings):

```
output.elasticsearch.pipelines:
- pipeline: '%{[fields.pipeline]}'
```

- use `fields.pipeline`, but fall back to `default-pipeline` if field is missing:

```
output.elasticsearch.pipelines:
- pipeline: '%{[fields.pipeline]}'
- pipeline: 'default-pipeline'
```

- map event field to pipeline name with default pipeline `default-pipeline`, if no mapping is defined or field is missing:

```
output.elasticsearch.pipelines:
- pipeline: '%{[fields.pipeline]}''
  default: ''default-pipeline'
  mappings:
    'abc': 'pipeline-abc'
    'def': 'pipeline-def'
```

- use conditionals to select pipeline based on HTTP code:

```
output.elasticsearch.pipelines:
- pipeline: 'ok-pipeline'
  when.range:
    http.code: [200, 299]
- pipeline: 'verybad-pipeline'
  when.range:
    http.code: [500, 999]
- pipeline: 'default-pipeline'
```

- use filters + mapping for more complicated scenario:
```
output.elasticsearch.pipelines:
- pipeline: '%{[fields.pipeline]}'
  default: default-pipeline
  mappings:
    abc: pipeline-abc
  when.has_fields: ['event_data.IPAddress']
[...]

PR #2119, opened 2016-07-28
Configurable Elasticsearch index pattern

urso:
Part of #2074 
Depends on #2118 

Implement index selector based on selector support provided by #2087 and new format string support.

Default index pattern changes to `beatname-%{+yyyy.MM.dd}`.

More advanced index selection via selector:

```
output.elasticsearch.indices:
- rule1
- rule2
```

If `output.elasticsearch.indices` doesn't match, default case in `output.elasticsearch.index` is used (which is always set to default value on init).

Rules are similar to #2064 with supported keywords: `index`, `default`, `mapping`, `when`.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -13,6 +13,7 @@ https://github.com/elastic/beats/compare/v5.0.0-alpha5...master[Check the HEAD d
 ==== Breaking changes
 
 *Affecting all Beats*
+- Change Elasticsearch output index configuration to be based on format strings. If index has been configured, no date will be appended anymore to the index name. {pull}2119[2119]
 
 *Metricbeat*
 
--- a/filebeat/filebeat.full.yml
+++ b/filebeat/filebeat.full.yml
@@ -440,7 +440,7 @@ output.elasticsearch:
 
   # Optional index name. The default index name is set to name of the beat
   # in all lowercase.
-  #index: filebeat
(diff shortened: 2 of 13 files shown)


tsg:
LGTM. I think this will require some changes in docs & changelog.

urso:
Yes, requires updates to docs. With change being quite complex and format strings being new I'd like to add some doc on config file format and supported 'types' in config file first.

For example:
- [es configuration style docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/setup-configuration.html#styles)
- [ls config file structure](https://www.elastic.co/guide/en/logstash/current/configuration-file-structure.html) and [types](https://www.elastic.co/guide/en/logstash/current/configuration-file-structure.html#plugin-value-types), [more types](https://www.elastic.co/guide/en/logstash/current/event-dependent-configuration.html) and [environment variable support](https://www.elastic.co/guide/en/logstash/current/environment-variables.html)

ruflin:
Should we change as part of this also our default config file? https://github.com/elastic/beats/blob/master/filebeat/filebeat.full.yml#L337 What I have in mind is that if someone upgrades to 5.0 the default in the config file is still the same as before. Also uncommenting should have the same affect as the default we use.

urso:
@ruflin my bad working on too many branches in parallel. Default config should be updated in this PR.

urso:
full config files have been updated.

PR #2122, opened 2016-07-28
Generate Kibana index pattern

monicasarbu:
This PR creates a python script to generate the index pattern in json format from the fields.yml. The generated index pattern is needed by Kibana to know how to format the fields, think of a percentage value.

As Kibana supports `number`, `string` and `boolean` types, all the variations of `half_float`, `float`, `integer`, `long` become `number` in the index-pattern.

For special fields that require a special format (like `percent`, `bytes`), usually you need to define the format manually for each field in Kibana in order for Kibana to know how to display a number when it's a percentage (using %) or in MB/GB for bytes values. To automate the process, this PR lets you define the format of the field in the `format` option.

```
- name: used.bytes
   type: long
   format: bytes
   description: >
        Used memory.

- name: used.pct
   type: half_float
   format: percent
   description: >
        The percentage of used memory.

```

This PR is another way to solve https://github.com/elastic/beats-dashboards/pull/117


libbeat/scripts/generate_index_pattern.py
```diff
@@ -0,0 +1,115 @@
...
+        "name": path,
+        "count": 0,
+        "scripted": False,
+        "indexed": True,
+        "analyzed": False,
+        "doc_values": True,
+    }
+    if "type" in desc:
+        if desc["type"] in ["half_float", "float", "integer", "long"]:
+            field["type"] = "integer"
```
tsg: Is "integer" correct here? Shouldn't it be "number"?

tsg:
LGTM, great work! I see only the Packetbeat index was re-generated, should we do them all in this PR?

Also, needs a rebase.

libbeat/scripts/generate_index_pattern.py
monicasarbu: Oh, good spot, @tsg. Thanks!

monicasarbu:
Yes, I will do for all the Beats in the same PR.

monicasarbu:
With this change the `import-dashboards.sh` doesn't work anymore to load the dashboards and it needs to be changed in a different PR.

metricbeat/docs/fields.asciidoc
```diff
@@ -1901,6 +1911,8 @@ CPU Core number.
 
 type: half_float
 
+format: percent
```
ruflin: It's nice to have this also in the docs 👍

metricbeat/etc/fields.yml
```diff
@@ -1207,42 +1220,50 @@
...
 
             - name: idle.pct
               type: half_float
+              format: percent
               description: >
                 The percentage of CPU time spent idle.
 
             - name: iowait.pct
               type: half_float
+              format: percent
```
ruflin: If we keep the naming convention with .pct we could potentialy also autodetect this in the future ...

ruflin:
LGTM. Great work. Could you rebase?

metricbeat/etc/fields.yml
monicasarbu: Yes, we could generate the format for all the fields that contain`pct`, but there are only a few `pct` fields. We need to be able to specify the `format` as there are other fields of type `half_float` that don't have `pct`.

ruflin: Definitively. Having the configuration the configuration option is great. Was more thinking of offering both (in the future). If it is pct and no definition is there, it uses `percent`. If a format is there, the format will taken.

monicasarbu:
Rebased.

CHANGELOG.asciidoc
```diff
@@ -45,6 +45,17 @@ https://github.com/elastic/beats/compare/v5.0.0-alpha5...master[Check the HEAD d
 
 *Affecting all Beats*
 
+- Periodically log internal metrics. {pull}1955[1955]
```
ruflin: Where are these changelog entries coming from?

monicasarbu: Yes, it looks like "git rebase master" didn't do a good job. They were part of alpha5. Thanks for noticing it. I am fixing it manually.

PR #2123, opened 2016-07-28
Remove note about Kibana versions

dedemorton:
@monicasarbu Not sure if I got the details quite right in this note. Please review.

### DIFF
--- a/libbeat/docs/dashboards.asciidoc
+++ b/libbeat/docs/dashboards.asciidoc
@@ -6,8 +6,6 @@ the perfect dashboards for your needs. But this flexibility can be a bit
 overwhelming at the beginning, so we have created some sample dashboards to get you
 started and to demonstrate what is possible based on the Beats data.
 
-NOTE: The Beats dashboards require Kibana 4 or later.
-
 Each Beat comes packaged with a script that you can use to import the example dashboards,
 visualizations, and searches for that Beat. The script also creates index patterns, such as 
 `packetbeat-*` for Packetbeat, `metricbeat-*` for Metricbeat, and so on. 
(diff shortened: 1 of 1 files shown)


libbeat/docs/dashboards.asciidoc
```diff
@@ -6,7 +6,14 @@ the perfect dashboards for your needs. But this flexibility can be a bit
 overwhelming at the beginning, so we have created some sample dashboards to get you
 started and to demonstrate what is possible based on the Beats data.
 
-NOTE: The Beats dashboards require Kibana 4 or later.
+NOTE: The Beats dashboards require Kibana 5 or later.
```
monicasarbu: In fact the Beats dashboards work with both Kibana versions, but Kibana 5.0 works only with Elasticsearch 5.0 (needs to be confirmed by Kibana team). At a second thought, not sure if we should include this information in our dashboards. Maybe we should just avoid this information for now as it's not related to Beats.

dedemorton: @monicasarbu OK, so it sounds like you want me to:
- revert the original statement on line 9 (to say "The Beats dashboards require Kibana 4 or later" )
- remove the note under "Known Issue in 5.0.0-alpha5". I thought there was an issue specific to our dashboards (and that users who have an earlier version of the ES alpha installed would need to move to alpha5 to get our dashboards to work). But it sounds like this is a Kibana + ES issue and should probably be documented in the release notes for those products. 

Is that right?

dedemorton:
@monicasarbu Please review my changes. As we discussed, I removed the note about the Kibana version.

monicasarbu:
Thanks @dedemorton

PR #1437, opened 2016-04-20
Added path.logs and made logging to files the default

tsg:
This adds the `-path.logs` CLI flag and corresponding config file
setting. By default it's set to `{path.home}/logs`. The rotating
file writer now writes to this path by default.

For DEB/RPM, the logs path is set to `/var/log/{beatname}` which
matches the previous default location. However, this is a breaking
change because the previous default was to send the logs to syslog.

Part of #1371.

PR #2126, opened 2016-07-28
change defaults shown in doc for to_syslog and to_files

dedemorton:
This was missed in PR #1437

### DIFF
--- a/libbeat/docs/loggingconfig.asciidoc
+++ b/libbeat/docs/loggingconfig.asciidoc
@@ -43,12 +43,12 @@ You can specify the following options in the `logging` section of the +{beatname
 
 ===== to_syslog
 
-Sends all logging output to syslog if enabled. For non-Windows systems, the default
-value is true.
+If enabled, sends all logging output to syslog (on non-Windows systems only). The default
+value is false. 
 
 ===== to_files
 
-Writes all logging output to files subject to file rotation. On Windows systems, the
+Writes all logging output to files subject to file rotation. The
 default value is true.
 
 ===== level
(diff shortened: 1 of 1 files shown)


libbeat/docs/loggingconfig.asciidoc
```diff
@@ -43,12 +43,12 @@ You can specify the following options in the `logging` section of the +{beatname
 
 ===== to_syslog
 
-Sends all logging output to syslog if enabled. For non-Windows systems, the default
-value is true.
+If enabled, sends all logging output to syslog (on non-Windows systems only). The default
```
tsg: I think we can leave out the "(on non-Windows systems only)" part. It's not common to have a syslog server when running on Windows, but it's technically possible.

tsg:
Good catch! I left a minor comment.

libbeat/docs/loggingconfig.asciidoc
dedemorton: Ah, OK. I included it because I didn't think we would be able to write to a Windows syslog server, but it sounds like we would, so I will remove it.

dedemorton:
@tsg Fxed the issue. If it looks OK to you, this is ready to squash and merge now.

libbeat/docs/loggingconfig.asciidoc
andrewkroh: Just a late follow-up with some ancillary info, the Golang [syslog](https://golang.org/src/log/syslog/syslog.go) package isn't implemented on Windows. It has a `// +build !windows` build tag.

PR #2127, opened 2016-07-29
clarify order of execution for inlcude_lines and exclude_lines

dedemorton:
This replaces PR #2054 

Rather than changing the example, I've added text to make it clear that the order in which the two options appears in the config file doesn't matter.


tsg:
LGTM.

Issue #2128, opened 2016-07-29
Improve packaging for community Beats


tsg:
#2125 has revealed some issues with the packaging for the community Beats created with the generator. This is a meta ticket for improvements:
- [ ] The generator doesn't encourage people to use a vendor directory, but it is actually required for packaging. I think we will save people time if we run `glide create` and `glide install --strip-vcs` during the setup phase and we default the ES_BEATS Makefile variable to the vendor dir.
- [ ] #2130 The generation of the dev-tools files currently happens at time of make package, but since these files should generally be committed, it would make more sense to include them in the generator template.
- [x] #2129 The current dev-tools/packer/Makefile assumes Cgo, but most community Beats are pure Go. I think it's better to assume pure go, as the compilation is faster and the resulting binaries are static.

kaiyan-sheng:
Closing this issue, generator has been almost rewritten by @fearful-symmetry.

PR #2129, opened 2016-07-29
Use pure Go compilation for community beats

tsg:
Part of #2128. Most community Beats don't require Cgo, and if Cgo is
required things will probably not work out of the box anyway (extra
libraries will probably be needed).

Also `make package` call `make all` in the `dev-tools/packer` so that
the packages are built, not only the binaries.

### DIFF
--- a/libbeat/scripts/Makefile
+++ b/libbeat/scripts/Makefile
@@ -318,4 +318,4 @@ create-packer:
 package: create-packer
 	cd dev-tools/packer; \
 	export ES_BEATS=${ES_BEATS} ; \
-	make deps images ${BEATNAME}
+	make deps images all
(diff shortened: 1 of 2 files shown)


cyrilleverrier:
@tsg: is it still possible to cross-compile with Cgo ?
My project includes https://github.com/pebbe/zmq4 which requires Cgo for binding go with zmq library

tsg:
Yes, it's still possible, just keep using your version of  dev-tools/packer/Makefile. I consider that kind of configuration. Just the default can be simpler.

ruflin:
I like the change that by default we assume no `cgo`. It would be nice if `cgo` builds could be enabled through setting an environment variable or similar. So @cyrilleverrier could still use our most recent dev-tools package and would only have to add an additional parameter. In addition, our beats would use the same version so only 1 Makefile to apply the changes. This probably needs some more restructuring. @tsg I remember you were thinking about moving all this kind of stuff more to a config file like logic?

PR #2130, opened 2016-07-29
Replaced the create_packer script with generator files

tsg:
Part of #2128. This replaces the create_packer makefile target with
the required files in the beat generator. The result should be the
same, but I think it's good to keep the same mechanism for all
template files.

Depends on #2129.


tsg:
@cyrilleverrier, let me know if I missed an important reason for this to be done from the Makefile.

ruflin:
@tsg My approach was going to opposite way and move more and more files to libbeat. The reasoning here is that after updating for example libbeat to the most recent version, `make create-packer` can be run again and the most recent packager version is available as part of the beat.

In general I try to move away from cookiecutter and have our own scripts as the files are very simple and this gives us more flexibility.

tsg:
Closing this one based on the internal chat with @ruflin. There's still some stuff that I don't like about how this works, but we can fix it later.

PR #2132, opened 2016-07-29
Document config file format

urso:
Document config file format + structure.

config files with duplicate (out of date) docs:
shared-env-vars.asciidoc
yaml.asciidoc (also shared)

### DIFF
--- a/libbeat/docs/config-file-format.asciidoc
+++ b/libbeat/docs/config-file-format.asciidoc
@@ -1,0 +1,367 @@
+[[config-file-format]]
+== Config file format
+
+Beats config files are based on http://www.yaml.org[YAML], a file format that is
+easier to read and write than other common data formats like XML or JSON.
+
+In beats all YAML files start with an dictionary, an unordered collection of
+name/value pairs. In addition to dictionaries, YAML also supports lists, numbers,
+strings, and many other data types. All members of the same list or dictionary must
+have the same indentation level.
+
+Dictionaries are represented by simple `key: value` pairs all having the same
+indentation level. The colon after `key` must be followed by a space.
+
+["source","yaml",subs="attributes"]
+------------------------------------------------------------------------------
+name: John Doe
+age: 34
+country: Canada
+------------------------------------------------------------------------------
+
+Lists are introduced by dashes `- `. All list members will be lines beginning
+with `- ` at the same indentation level.
+
+["source","yaml",subs="attributes"]
+------------------------------------------------------------------------------
+- Red
+- Green
+- Blue
+------------------------------------------------------------------------------
+
+Lists and dictionaries are used in beats to build structured configurations.
+
+["source","yaml",subs="attributes"]
+------------------------------------------------------------------------------
+filebeat:
+  prospectors:
+    - input_type: log
+      paths:
+        - /var/log/*.log
+      multiline:
+        pattern: '^['
+        match: after
+------------------------------------------------------------------------------
+
+Lists and dictionaries can also be represented in abbreviated form. Abbreviated
+form is somewhat similar to JSON using `{}` for dictionaries and `[]` for lists:
+
+["source","yaml",subs="attributes"]
+------------------------------------------------------------------------------
+person: \{name: "John Doe", age: 34, country: "Canada"}
+colors: ["Red", "Green", "Blue"]
+------------------------------------------------------------------------------
+
+
+[[config-file-format-namespacing]]
+=== Namespacing
+
+All settings are structured using dictionaries and lists. Those are collapsed
+into "namespaced" settings, by creating a setting using the full path of the
+settings name and it's parent structures names, when reading the configuration
+file.
+
+For example this setting:
+
+["source","yaml",subs="attributes"]
+------------------------------------------------------------------------------
+
+output:
+  elasticsearch:
+    index: 'beat-%{+yyyy.MM.dd}'
+
+------------------------------------------------------------------------------
+
+gets collapsed into `output.elasticsearch.index: 'beat-%{+yyyy.MM.dd}'`. The
+full name of a setting is based on all parent structures involved.
+
+Lists create numeric names starting with 0.
+
+For example this filebeat setting:
+
+["source","yaml",subs="attributes"]
+------------------------------------------------------------------------------
+
+filebeat:
+  prospectors:
+    - input_type: log
+
+------------------------------------------------------------------------------
+
+Gets collapsed into `filebeat.prospectors.0.input_type: log`.
+
+Alternatively to using indentation, setting names can be used in collapsed form too.
+
+Note: having two settings with same fully collapsed path is invalid.
+
+Simple filebeat example with partially collapsed setting names and use of compact form:
+
+
+["source","yaml",subs="attributes"]
+------------------------------------------------------------------------------
+
+filebeat.prospectors:
+- input_type: log
+  paths: ["/var/log/*.log"] 
+  multiline.pattern: '^['
+  multiline.match: after
+
+output.elasticsearch.hosts: ["http://localhost:9200"]
+
+------------------------------------------------------------------------------
+
+[[config-file-format-type]]
+=== Config file Data Types
+
+Values of configuration settings are interpreted as required by beats.
+If a value can not be correctly interpreted as the required type - for example a
+string is given when a number is required - the beat will fail to start up.
+
+==== Boolean
+
+Boolean values can be either `true` or `false`. Alternative names for `true` are
+`yes` and `on`. Instead of `false` the values `no` and `off` can be used.
+
+["source","yaml",subs="attributes"]
+------------------------------------------------------------------------------
+enabled: true
+disabled: false
+------------------------------------------------------------------------------
+
+==== Number
+
+Number values require you to enter the number to use without using single or
+double quotes. Some settings only support a restricted number range though.
+
+["source","yaml",subs="attributes"]
+------------------------------------------------------------------------------
+integer: 123
+negative: -1
+float: 5.4
+------------------------------------------------------------------------------
+
+==== String
+
+In YAML[http://www.yaml.org], multiple styles of string definitions are supported:
+ double-quoted, single-quoted, unquoted.
+
+The double-quoted style is specified by surrounding the string with `"`. This
+style provides support for escaping unprintable characters using `\`, but comes
+at the cost of having to escape `\` and `"` characters.
+
+The single-quoted style is specified by surrounding the string with `'`. This
+style supports no escaping (use `''` to quote a single quote). Only printable
+characters can be used when using this form.
+
+Unquoted style requires no quotes, but does not support any escaping plus care
+needs to be taken to not use any symbol that has a speacial meaning in YAML.
+
+Note: Single-quoted style is recommended when defining regular expressions,
+event format strings, windows file paths, or non-alphabetical symbolic characters.
+
+==== Duration
+
+Durations require a numeric value with optional fraction and required unit.
+Valid time units are `ns`, `us`, `ms`, `s`, `m`, `h`. Sometimes features based
+on durations can be disabled by using zero or negative durations.
+
+["source","yaml",subs="attributes"]
+------------------------------------------------------------------------------
+duration1: 2.5s
+duration2: 6h
+duration_disabled: -1s
+------------------------------------------------------------------------------
+
+==== Regular Expression
+
+Regular expressions are special strings getting compiled into regular
+expressions at load time.
+
+As regular expressions and YAML use `\` for escaping
+characters in strings, it's highly recommended to use single quoted strings when
+defining regular expressions. When single quoted strings are used, `\` character
+is not interpreted by YAML parser as escape symbol.
+
+==== Format String (sprintf)
+
+Format strings enable you to refer to event field values creating a string based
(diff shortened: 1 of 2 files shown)


libbeat/docs/config-file-format.asciidoc
```diff
@@ -0,0 +1,80 @@
+[[config-file-format]]
+== Config file format
+
+Beats config files are based on http://www.yaml.org[YAML], a file format that is easier to read and write then other common data format like XML or JSON.
```
dedemorton: should be "than other common data formats" (plural)

libbeat/docs/config-file-format.asciidoc
```diff
@@ -0,0 +1,80 @@
...
+
+In beats all YAML files start with a "dictionary" (often called a "hash"). In addition to dictionaries, YAML also supports lists, numbers, strings, and many other data types. All members of lists and dictionaries must have the same indentation level.
+
+Dictionaries are represented by simple `key: value` pairs all having the same indentation level. The collon after `key` must be followed by a space.
```
dedemorton: "The colon after..." (typo)

dedemorton:
Great idea!

libbeat/docs/config-file-format.asciidoc
```diff
@@ -0,0 +1,80 @@
...
+==== Event Format String
+
+==== References
+
+=== YAML Tips and Gotchas
+
+[float]
+==== Use Spaces for Indentation
+
+Indentation is meaningful in YAML. Make sure that you use spaces, rather than tab characters, to indent sections.
```
tsg: We could add something like: "In the default configuration files and in all the examples in the documentation, we use 2 spaces per indentation level. We recommend you do the same."

urso:
How should we add this doc best to our documentation?

On the one hand I like it to be generic (as config file format is same for community-beats) and have beats docs point to this docs, on the other hand it duplicates content in yaml.asciidoc which is just some tips specific per beat (rather would like to include 'Config file format' doc with every beat).

Question: with config file structure having nested namespaces, how can we make namespacing usage more clear beats config reference?

libbeat/docs/config-file-format.asciidoc
```diff
@@ -0,0 +1,355 @@
+[[config-file-format]]
+== Config file format
+
+Beats config files are based on http://www.yaml.org[YAML], a file format that is
+easier to read and write then other common data formats like XML or JSON.
```
dedemorton: change "then" to "than"

libbeat/docs/config-file-format.asciidoc
```diff
@@ -0,0 +1,355 @@
...
+
+In beats all YAML files start with a "dictionary" (often called a "hash"). In
```
dedemorton: Since you introduce "dictionary" as a new term you should provide a short definition.

libbeat/docs/config-file-format.asciidoc
```diff
@@ -0,0 +1,355 @@
...
+addition to dictionaries, YAML also supports lists, numbers, strings, and many
+other data types. All members of lists and dictionaries must have the same
```
dedemorton: This is grammatically ambiguous. Might be better to say, "All members of the same list or dictionary must have the same indentation level."

libbeat/docs/config-file-format.asciidoc
```diff
@@ -0,0 +1,355 @@
...
+------------------------------------------------------------------------------
+person: {name: "John Doe", age: 34, country: "Canada"}
+colors: ["Red", "Green", "Blue"]
+------------------------------------------------------------------------------
+
+
+[[config-file-format-namespacing]]
+=== Namespacing
+
+All settings bases on dictionaries and lists are collapsed into "namespaced"
```
dedemorton: Passive voice here leaves it unclear as to when/how the items are collapsed. Does the YAML parser do the collapsing?

libbeat/docs/config-file-format.asciidoc
```diff
@@ -0,0 +1,355 @@
...
+["source","yaml",subs="attributes"]
+------------------------------------------------------------------------------
+constant-format-string: 'constant string'
+field-format-string: '%{[fieldname]} string'
+format-string-with-date: '%{[fieldname]}-%{+yyyy.MM.dd}'
+------------------------------------------------------------------------------
+
+
+[[config-file-format-env-vars]]
+=== Enironment Variables
```
dedemorton: I'm assuming that you mean to remove this topic, right? https://www.elastic.co/guide/en/beats/packetbeat/current/using-environ-vars.html

libbeat/docs/config-file-format.asciidoc
```diff
@@ -0,0 +1,355 @@
...
+  subnamespace:
+    host: localhost
+    sleep: 1s
+------------------------------------------------------------------------------
+
+when using plain references.
+
+
+[[config-file-format-cli]]
+=== Command line arguments
```
dedemorton: I understand why you want to discuss this here, but the content is out of place in a topic about the structure of the config file. My concern is that users will miss this section because they'll expect the topic to be about the structure of the config file (they might not read this content if they think they have a good handle on config files). Instead, we should have a separate topic that talks about the ability to overwrite config settings from the command line and specify multiple config files (this will raise visibility of this capability). We can mention the capability briefly and then point off to the other topic for more info.

libbeat/docs/index.asciidoc
```diff
@@ -28,6 +28,8 @@ include::./dashboards.asciidoc[]
 
 include::./newbeat.asciidoc[]
 
+include::./config-file-format.asciidoc[]
```
dedemorton: I'd like to add this content to the docs for every beat (using asciidoc include statements) because I think users are likely to overlook it if it's only in the Platform Ref.

dedemorton:
@urso I'm done with high level comments. I'd like to do a detailed technical edit of this content after you've merged it (me adding line notes is tedious for everyone). I can take care of the restructuring that I've suggested.

libbeat/docs/config-file-format.asciidoc
dedemorton: Oh, you can ignore this ^^. I see you mention the shared files above.

libbeat/docs/config-file-format.asciidoc
```diff
@@ -0,0 +1,355 @@
...
+
+[[config-file-format-references]]
+=== References
+
+Beats settings can reference other settings splicing multiple optionally custom
+named settings into new values. References use the same syntax as
+<<config-file-format-env-vars>> do. Only fully collapsed setting names can be
+referenced to.
+
+Example referencing `es.host` in `output.elasticsearch.hosts`:
```
dedemorton: @urso One more thing to add: Before introducing the concept of custom named settings, I would first introduce a simple example like:
`filebeat.registry_file: ${path.data}/registry`

Maybe even call out specifically that you can reference the path settings.

libbeat/docs/config-file-format.asciidoc
urso: Checking yaml specs I see `dictionary` is not really used. `dictionary` doesn't seem a very good option here. Alternatives to `dictionary` are `object`, `hash` (use by logstash I think, thanks to ruby influence), `table`, `map(-ping)` (yaml spec code docs use `mapping`). For users familiar with JSON, `object` will be most natural.

libbeat/docs/config-file-format.asciidoc
urso: done

libbeat/docs/config-file-format.asciidoc
urso: replace with `object` plus short intro what an object is:

> In beats all YAML files start with an object, an unordered collection of
name/value pairs.

libbeat/docs/config-file-format.asciidoc
urso: done

libbeat/docs/config-file-format.asciidoc
urso: Yeah, I didn't find a better place for this. I think the Elasticsearch docs have this also in same section. Maybe we can move it later?

libbeat/docs/config-file-format.asciidoc
urso: Does it really matter who/where this happens from technical point of view?

TBH they are not really collapsed, but split up in structure (technically the exact opposite) + configs are always accessible using full path or sub-pathes.

That is, the 'collapsing' part is more of an idea for users.

libbeat/docs/index.asciidoc
urso: Yes me too. But I still would like to see this in the Platform Reference. Reason is: most community beats seem not to be documented at all, so users at least might be interested in some common knowledge regarding all beats.

libbeat/docs/config-file-format.asciidoc
dedemorton: @urso We need to be careful about introducing new terms to describe an existing technology. I don't see "object" in the YAML spec. Is there a precedent for using this term? If we're talking about a YAML file, we should use industry-standard terminology to describe it. I wonder if you could just start by explaining that YAML is composed of nested data structures (or collections?) and then go on to describe what the structures (or collections) can contain. I mean...your current description is already too restrictive because the collection might be a list rather than name/value pairs, right?

libbeat/docs/index.asciidoc
dedemorton: We can add it to the Platform ref, too...just need to make the content generic enough.

libbeat/docs/config-file-format.asciidoc
urso: good point, will do

libbeat/docs/index.asciidoc
urso: Talking about 'generic enough', there are some sample from filebeat and metricbeat.

dedemorton: We can do some conditional stuff in asciidoc, if necessary.

urso: conditional stuff in asciidoc? And people complain about me making config files turing complete :grin:

urso:
updated PR + fixed build + squashed all commits.

Currently looks like this:

screen shot 2016-08-12 at 17 15 12

libbeat/docs/config-file-format.asciidoc
```diff
@@ -0,0 +1,367 @@
...
+        pattern: '^['
+        match: after
+------------------------------------------------------------------------------
+
+Lists and dictionaries can also be represented in abbreviated form. Abbreviated
+form is somewhat similar to JSON using `{}` for dictionaries and `[]` for lists:
+
+["source","yaml",subs="attributes"]
+------------------------------------------------------------------------------
+person: \{name: "John Doe", age: 34, country: "Canada"}
```
andrewkroh: Is that leading escape on the `{` required?

urso: yes, otherwise doc build breaks for me. `{<name>: ...}` is obviously something special in asciidoc.

libbeat/docs/config-file-format.asciidoc
```diff
@@ -0,0 +1,367 @@
...
+==== String
+
+In YAML[http://www.yaml.org], multiple styles of string definitions are supported:
+ double-quoted, single-quoted, unquoted.
+
+The double-quoted style is specified by surrounding the string with `"`. This
+style provides support for escaping unprintable characters using `\`, but comes
+at the cost of having to escape `\` and `"` characters.
+
+The single-quoted style is specified y surrounding the string with `'`. This
```
andrewkroh: Missing a `b` for "by".

libbeat/docs/config-file-format.asciidoc
andrewkroh: Ok, I thought it might be for asciidoc.

andrewkroh:
@urso This is excellent! I really hope people take the time to read it. 👍

libbeat/docs/config-file-format.asciidoc
urso: fixed

libbeat/docs/config-file-format.asciidoc
dedemorton: It's doing that because you have `subs="attributes"` defined for the code block. You only need to include `subs="attributes"` if the code block includes a variable reference that you want asciidoc to resolve. So if  you had {beatname_lc} in the example block and wanted it to resolve to Packetbeat or something, you would include `subs="attributes"`.

dedemorton:
@urso Intro is much easier to follow now. Looks good! I'll plan to edit the content sometime before the beta release.

dedemorton:
@urso BTW, I think the current location of the content in the platform ref isn't right (users who are reading the platform ref won't even know about the config file until after they've installed their beat). 
If we want to include this doc in the platform ref (so it's available to developers), it belongs under the Developer Guide section.

I can work on getting the content into the appropriate sections after it's merged, if you'd like. Just wanted to flag this issue because the current location isn't right.

urso:
@dedemorton yeah, that's another thing I struggled with. I put it after 'Getting Started...' so users have hopefully seen beats before, but felt like putting it into developers guide would hide the content. Maybe it makes sense to make this one available in Platform Ref. and per beat docs?

PR #2152, opened 2016-08-02
Cleanup travis file

ruflin:
* Use travis goimport path instead of rsync (new feature)
* Remove glide as it is not needed. Vendor directory is part of the source code.

See https://github.com/elastic/beats/pull/2013#issuecomment-232185112


andrewkroh:
Nice 👍

PR #2154, opened 2016-08-02
Add gotestcover to vendor directory

ruflin:
gotestcover is currently deprecated but does exactly what we need, so no need to replace it.

Add gotestcover to the vendor directory removes the need to fetch gotestcover during the building process. This means one less component with external dependencies during the build process.

Remove go vet and autotest from build process as not needed anymore

This PR replaces https://github.com/elastic/beats/pull/2013


tsg:
LGTM.

andrewkroh:
LGTM

PR #2155, opened 2016-08-02
Rewrite import_dashboards.sh script in go

monicasarbu:
This PR includes a rewrite of the `import_dashboards.sh` script in Golang.

Features:
- [x] change the index name by using the `-i` option
- [x] authenticate the connection to Elasticsearch using username:password
- [x] update test to test import dashboards

We need to include this script into the Beat package (in a separate PR).

The `import` accepts the following options:
```
Usage: ./import [options]

Kibana dashboards are stored in a special index in Elasticsearch together with the searches,",visualizations and indexes that they use.")

You can import the dashboards, visualizations, searches and index-pattern for any Beat:
  1. from a local directory:
	./import -dir etc/kibana
  2. from a directory of a local zip archive:
	./import -dir metricbeat -file beats-dashboards-1.2.3.zip
  3. from a directory of zip archive available online:
	./import -dir metricbeat -url http://download.elastic.co/beats/dashboards/beats-dashboards-1.2.3.zip

Options:
  -dir string
    	Directory containing the subdirectories: dashboard, visualization, search, index-pattern. eg. kibana/
  -es string
    	Elasticsearch URL (default "http://127.0.0.1:9200")
  -file string
    	Zip archive file containing the Beats dashboards.
  -i string
    	Overwrites the Elasticsearch index name. For example you can replaces metricbeat-* with custombeat-*
  -k string
    	Kibana index (default ".kibana")
  -pass string
    	Password to connect to Elasticsearch
  -url string
    	URL to the zip archive containing the Beats dashboards
  -user string
    	Username to connect to Elasticsearch
```


CHANGELOG.asciidoc
```diff
@@ -44,7 +44,11 @@ https://github.com/elastic/beats/compare/v5.0.0-alpha5...master[Check the HEAD d
 ==== Added
 
 *Affecting all Beats*
+
+- Add enabled setting to packetbeat protocols. {pull}1988[1988]
```
monicasarbu: I added them manually here after they were removed in a previous PR.

libbeat/.gitignore
```diff
@@ -8,6 +8,7 @@ _output
 /docs/html_docs
 .vagrant
 build
+/dashboards/import_dashboards
```
ruflin: What will this ignore exactly? In case these are the generated dashboards, shouldn't these be in the build directory?

libbeat/dashboards/import_dashboards.go
```diff
@@ -0,0 +1,559 @@
...
+	client *elasticsearch.Client
+}
+
+func ParseCommandLine() (*CommandLine, error) {
+	var cl CommandLine
+
+	cl.flagSet = flag.NewFlagSet("import", flag.ContinueOnError)
+
+	cl.flagSet.Usage = func() {
+		const usage = `
```
ruflin: Could you move this doc string somewhere on the top and just use the variable here? This would make the code more readable.

libbeat/dashboards/import_dashboards.go
```diff
@@ -0,0 +1,559 @@
...
+Usage: ./import_dashboards [options]
+
+Kibana dashboards are stored in a special index in Elasticsearch together with the searches, visualizations, and indexes that they use.
+
+You can import the dashboards, visualizations, searches, and the index pattern for any Beat:
+  1. from a local directory:
+	./import_dashboards -dir etc/kibana
+  2. from a directory of a local zip archive:
+	./import_dashboards -dir metricbeat -file beats-dashboards-1.2.3.zip
```
ruflin: Probably use 5.0.0 for the example here as I assume this script will only ship with 5.x releases?

libbeat/dashboards/import_dashboards.go
```diff
@@ -0,0 +1,559 @@
...
+	err = cl.Read()
+	if err != nil {
+		return nil, err
+	}
+	importer.cl = cl
+
+	/* connect to Elasticsearch */
+	fmtstr, err := fmtstr.CompileEvent(cl.opt.Index)
+	if err != nil {
+		return nil, fmt.Errorf("fail to compile the Elasticsearch index pattern: %s", err)
```
ruflin: I suggest to replace this error message with `fail to build the Kibana index pattern: ...` The reason is that compile sounds to me like a binary is generated and index pattern relates to Kibana and not elasticsearch.

libbeat/dashboards/import_dashboards.go
```diff
@@ -0,0 +1,559 @@
...
+		cl.flagSet.Usage()
+		return nil, err
+	}
+	err = cl.Read()
+	if err != nil {
+		return nil, err
+	}
+	importer.cl = cl
+
+	/* connect to Elasticsearch */
```
ruflin: I think the connection to elasticsearch only happens on line 119?

libbeat/dashboards/import_dashboards.go
```diff
@@ -0,0 +1,559 @@
...
+
+func (imp Importer) ImportJsonFile(_type string, file string) error {
+
+	path := "/" + imp.cl.opt.KibanaIndex + "/" + _type
+
+	reader, err := ioutil.ReadFile(file)
+	if err != nil {
+		return fmt.Errorf("fail to read %s. Error: %s", file, err)
+	}
+	var json_content map[string]interface{}
```
ruflin: Normally we use `jsonContent` as golang naming convention.

libbeat/dashboards/import_dashboards.go
```diff
@@ -0,0 +1,559 @@
...
+				},
+			},
+		})
+	if err != nil {
+		fmt.Printf("fail to set the mapping. Error: %s\n", err)
+	}
+	return nil
+}
+
+func (imp Importer) ImportJsonFile(_type string, file string) error {
```
ruflin: I would recommend to use here `type` to not have variables with _

libbeat/dashboards/import_dashboards.go
```diff
@@ -0,0 +1,559 @@
...
+	}
+
+	return nil
+}
+
+func (imp Importer) ImportDashboard(file string) error {
+
+	fmt.Println("Import dashboard ", file)
+
+	/* load dashboard */
```
ruflin: load and import is somehow used interchangeable in the code and comments. I would make the case that `load` is the one that should be used for the following reasoning:

* import would be if we were in the Kibana UI and would import it from there
* load is as we load it from the local files into kibana / elasticsearch

libbeat/dashboards/import_dashboards.go
```diff
@@ -0,0 +1,559 @@
...
+}
+
+func ParseCommandLine() (*CommandLine, error) {
+	var cl CommandLine
+
+	cl.flagSet = flag.NewFlagSet("import", flag.ContinueOnError)
+
+	cl.flagSet.Usage = func() {
+		const usage = `
+Usage: ./import_dashboards [options]
```
ruflin: See comment below about `import` vs `load`. Should we call it `load_dashboards`?

libbeat/.gitignore
monicasarbu: `import_dashboards` is the exec, built from `import_dashboards.go`. This is generated with `make build-dashboards`.

libbeat/dashboards/import_dashboards.go
```diff
@@ -0,0 +1,559 @@
+package main
```
ruflin: I think for the moment it is great to have it in one file to easily make changes and push the project forward. If we move forward with the project and start to ship it with beats. we should start to think about where to place these files as it feels kind of strange to have a golang project with a main method inside libbeat.

libbeat/outputs/elasticsearch/client.go
```diff
@@ -551,17 +551,18 @@ func (client *Client) PublishEvent(data outputs.Data) error {
 func (client *Client) LoadTemplate(templateName string, template map[string]interface{}) error {
 
 	path := "/_template/" + templateName
-	status, _, err := client.request("PUT", path, "", nil, template)
+	return client.LoadJson(path, template)
```
ruflin: should we check for the error here and "print" a template specific error. Now it seems like it only returns the json error which will be printed out later and does not explain to the user that something with the template didn't work.

libbeat/dashboards/import_dashboards.go
monicasarbu: yes, true. It happens a bit later, but the other steps are required for setting up the connection to Elasticsearch. I could create a separate function that gets a connection to Elasticsearch.

libbeat/outputs/elasticsearch/client.go
```diff
@@ -551,17 +551,18 @@ func (client *Client) PublishEvent(data outputs.Data) error {
...
 
+func (client *Client) LoadJson(path string, json map[string]interface{}) error {
+	status, _, err := client.request("PUT", path, "", nil, json)
 	if err != nil {
-		return fmt.Errorf("Template could not be loaded. Error: %s", err)
+		return fmt.Errorf("couldn't load json. Error: %s", err)
 	}
-	if status != 200 {
-		return fmt.Errorf("Template could not be loaded. Status: %v", status)
+	if status > 300 {
```
ruflin: That means all 200 codes are fine? Why was this change needed? Do we get other codes for loading the kibana files?

libbeat/outputs/elasticsearch/client.go
```diff
@@ -551,17 +551,18 @@ func (client *Client) PublishEvent(data outputs.Data) error {
...
+		return fmt.Errorf("couldn't load json. Status: %v", status)
 	}
 
-	logp.Info("Elasticsearch template with name '%s' loaded", templateName)
```
ruflin: This should be moved to line 554 i no error happened as otherwise we loose this message.

ruflin:
It's really great to have this script as it makes it cross platform compatible without having additional dependencies. I left a few comments in the code.

libbeat/dashboards/import_dashboards.go
monicasarbu: Right, actually I was planning to update the url (from the 3rd point) after we create the archive with the beats dashboards (needs to come in a separate PR) in order to provide a valid url in these examples.

libbeat/dashboards/import_dashboards.go
monicasarbu: This is how I called the scripts first, `load_dashboards`, but based on the feedback I got, I decided to rename it to `import_dashboards`, as people had troubles understanding `load`, especially for someone that it's used with the Kibana terminology.

libbeat/outputs/elasticsearch/client.go
monicasarbu: Yes, I changed not to return an error in case of 2xx http response code as they are considered success. It's debatable if we want to through an error for the 3xx codes. I discovered this error as sometimes I got 201 response code at the first run.

I hope this change is not affecting the functionality of the other Beats.

monicasarbu:
@ruflin I addressed all your comments, thank you!

tsg:
Merging so we can continue with packaging.

libbeat/.gitignore
ruflin: Got it. I got confused as the entry is under the list of excluded directories.

libbeat/dashboards/import_dashboards.go
ruflin: Perfect. Wasn't aware of that there is already a history 👍

libbeat/outputs/elasticsearch/client.go
ruflin: @urso Can you have a look here?

PR #2156, opened 2016-08-02
Use scaled_floats for percentages in ES mapping

tsg:
Elasticsearch has recently added scaled_float as an option for storing floating
point numbers. The scaled floats are stored internally as longs, which means
they can take advantage of the integer compression in Lucene. See
https://github.com/elastic/elasticsearch/pull/19264 for details.

The PR moves all percentages to scaled floats. In our `fields.yml` we assume a
default scaling factor of 1000, which should work well for our percentages
(values between 0 and 1). This scaling factor can also be set to a different
value in `fields.yml`.

### DIFF
--- a/libbeat/scripts/generate_template.py
+++ b/libbeat/scripts/generate_template.py
@@ -178,18 +178,23 @@ def fill_field_properties(args, field, defaults, path):
             }
 
     elif field["type"] in ["geo_point", "date", "long", "integer",
-                           "double", "float", "half_float", "boolean"]:
+                           "double", "float", "half_float", "scaled_float",
+                           "boolean"]:
         # Convert all integer fields to long
         if field["type"] == "integer":
             field["type"] = "long"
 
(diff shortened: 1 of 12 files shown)


libbeat/scripts/generate_template.py
```diff
@@ -178,19 +178,24 @@ def fill_field_properties(args, field, defaults, path):
...
+            # ES 2.x doesn't support half or scaled floats, so convert to float
             field["type"] = "float"
 
         properties[field["name"]] = {
             "type": field.get("type")
         }
 
+        if field["type"] == "scaled_float":
+            properties[field["name"]]["scaling_factor"] = \
+                field.get("scaling_factor", 1000)
```
ruflin: Nice

ruflin:
Flaky appveyor test should be fixed in https://github.com/elastic/beats/pull/2156

ruflin:
LGTM

PR #2157, opened 2016-08-02
Change close_inactive default to 5m

ruflin:
The close_inactive default was set to one hour. Having scan_frequency as default set to 10s a lower value as 5 minutes is reasonable for less active files.

There were many issues in the past with open files. This will close inactive files faster by default now.


tsg:
LGTM.

PR #2159, opened 2016-08-02
Fix flaky filebeat test clean_inactive

ruflin:
Sometimes the test failed because the registry file was not written yet. This change introduces a check to guarantee that the registry file was already written when requested.


andrewkroh:
LGTM

tsg:
LGTM.

PR #2165, opened 2016-08-03
Remove full buffer check as not needed

ruflin:
If the buffer is full and no error is returned, the number of bytes read will be returned. In case the buffer is full and there is an error, the error is already returned

Test added to confirm event longer then buffer works as intended.


filebeat/harvester/log.go
```diff
@@ -271,6 +271,10 @@ func (h *Harvester) close() {
 	harvesterClosed.Add(1)
 }
 
+// newLogFileReader creates a new reader to read log files
+// It creates a chain of readers which looks as following:
+// limit -> (multiline -> timeout) -> strip_newline -> json -> encode -> log_file
```
urso: reverse order to match reading order from lowest-level to highest. This makes actual data flow more clear.

```
logFile -> encoding -> json -> strip_newline -> (timeout -> multiline) -> limit
```

ruflin: I picked the other order an purpose as it is the caller / creation order and in case of an error, it goes back to the caller (left). But both orders work for me. I will update the text to make this even more clear. I see it as really important that this logic is understood when working with the code.

filebeat/harvester/log_file.go
```diff
@@ -57,40 +61,49 @@ func (r *LogFile) Read(buf []byte) (int, error) {
...
+		buf = buf[n:]
 
-		// Stdin is not continuable
-		if !r.fs.Continuable() {
-			logp.Debug("harvester", "Source is not continuable: %s", r.fs.Name())
-			return n, err
+		// Buffer is full. Cannot continue reading. Return.
+		// Can happen if n == bufferSize + io.EOF error
+		if len(buf) == 0 {
+			return totalN, err
```
urso: we still need err to by rewritten by r.errorChecks before returning?

btw. in io.Reader interface it's valid to postpone error and return the `0, err` with next call to read.

PR #2166, opened 2016-08-03
Cleanup Line Reader

ruflin:
* Simplify advance loop
* Move newline check and error return to the end of the loop
* Directly return on reading errors
* Add more docs
* Fix flaky filebeat test

### DIFF
--- a/filebeat/harvester/reader/line.go
+++ b/filebeat/harvester/reader/line.go
@@ -7,6 +7,7 @@ import (
 	"golang.org/x/text/transform"
 
 	"github.com/elastic/beats/libbeat/common/streambuf"
+	"github.com/elastic/beats/libbeat/logp"
 )
 
 // lineReader reads lines from underlying reader, decoding the input stream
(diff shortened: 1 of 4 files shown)


filebeat/harvester/reader/line.go
```diff
@@ -24,55 +24,34 @@ type Line struct {
...
-	for {
-		// read next 'potential' line from input buffer/reader
-		err := l.advance()
-		if err != nil {
-			return nil, 0, err
-		}
 
-		// check last decoded byte really being '\n'
-		buf := l.outBuffer.Bytes()
-		if buf[len(buf)-1] == '\n' {
```
urso: this check+loop is here in case of l.advance() did find a sub-match in input string, which is no newline character if encoding has been applied.

ruflin: I readded the check but left out the loop. Instead I return an error. I would think if that happens this is an error as the comparison in advance is already based on the encoded newline character.

urso: tl;dr It can be perfectly ok advance thinking it found a newline, but in utf8 actually not being a newline due to relying on byte comparisons only.

the reason it has been a loop was: what if encoded newline was found due to being a prefix-/suffix-pattern used in encoding another symbol. The check for newline in both places is based on byte comparisons not taking the actual encoding into account. Checking last symbol in array is actually a newline character would require you to iterate array rune by rune (not byte by byte), taking the current encoding into account.

This check mostly acts as second hopefully-we-catch-this-somehow-heuristics in case of false-positive in `advance`.

filebeat/harvester/reader/line.go
```diff
@@ -24,55 +25,40 @@ type Line struct {
...
+	// read next 'potential' line from input buffer/reader
+	err := l.advance()
+	if err != nil {
+		return nil, 0, err
+	}
+
+	// check last decoded byte really being '\n' also unencoded
+	buf := l.outBuffer.Bytes()
+	if buf[len(buf)-1] != '\n' {
+		return nil, 0, fmt.Errorf("Found new line character that is unencoded not a new line character: %s", buf[len(buf)-1])
```
urso: errors will bubble up. This is no real error, but some heuristics in `advance` having found a false-positive.

filebeat/harvester/reader/line.go
ruflin: I readded the loop and a log message. I'm still curious if that is really a case that can happen or more a theoretical one.

Issue #875, opened 2016-01-28
Error publishing events in filebeat logged as INFO


ddikman:
When setting up filebeat it's easy to get the configuration wrong the first time, not to mention any other issues you might have with connectivity to logstash etc. In my case I had some errors connecting. I had the error level set to "error" meaning no log was created when filebeat started so I couldn't know if it even started, I simply didn't get any error.

The reason nothing got posted (which I found after increasing the log level and verifying the log path) was this:
```
INFO Error publishing events (retrying): write tcp xxx.xxx.xxx.xxx:50377->yyy.yyy.yyy.yyy:9300: write: broken pipe
```

If this would've been logged as an error, which I expect it should since it stops events from getting published I could've picked up on the issue. Would it be possible to change?

Additional log entries:
```
2016-01-28T08:22:33Z DBG  Forward preprocessed events
2016-01-28T08:22:33Z DBG  output worker: publish 6 events
2016-01-28T08:22:33Z DBG  Try to publish %!s(int=6) events to logstash with window size %!s(int=10)
2016-01-28T08:22:33Z DBG  %!s(int=0) events out of %!s(int=6) events sent to logstash. Continue sending ...
2016-01-28T08:22:33Z INFO Error publishing events (retrying): write tcp xxx.xxx.xxx.xxx:50347->yyy.yyy.yyy.yyy:9300: write: broken pipe
2016-01-28T08:22:33Z INFO send fail
2016-01-28T08:22:33Z INFO backoff retry: 1s
2016-01-28T08:22:34Z DBG  Try to publish %!s(int=6) events to logstash with window size %!s(int=10)
2016-01-28T08:22:34Z DBG  %!s(int=0) events out of %!s(int=6) events sent to logstash. Continue sending ...
2016-01-28T08:22:34Z INFO Error publishing events (retrying): write tcp xxx.xxx.xxx.xxx:50349->yyy.yyy.yyy.yyy:9300: write: broken pipe
2016-01-28T08:22:34Z INFO send fail
2016-01-28T08:22:34Z INFO backoff retry: 2s

```

ruflin:
We are currently improving the logging messages further to make it even more readable. As far as I remember I had a discussion with @urso about the exact log message above and we decided to put it into INFO as it is retrying and we want to keep the messages in Error's as minimal as possible. I see very well in your case that this didn't work out as expected. I think I agree and we should change it to Error. @urso WDYT?

We should have again a look at the above messages and decide on the error message and the reporting level.

[...]

PR #2183, opened 2016-08-08
Log event publishing errors as Err (was Info)

devenney:
Hi folks!

Today I ran into an issue regarding the classification of log messages. Our logging fell over and we saw no errors in the Filebeat logs. After further investigation we found that this error was being logged as INFO:

    2016/08/08 10:56:03.814969 single.go:130: INFO Connecting error publishing events (retrying): Head http://confidential-hostname:9200: dial tcp: lookup confidential-hostname on 10.251.0.2:53: no such host

This feels wrong as any event publishing error seems to be a real error. This PR ensures these errors are logged as errors.

____

We also found the following which I have not modified in this PR as I am not sure if it is intentional behaviour:

    2016/08/08 10:56:20.807266 log.go:80: INFO Read line error: No more bytes

This appears to be an issue in `filebeat/harvester/log.go`, lines 67-85. A number of these logging calls appear to be real error cases which I would argue should be logged as such. If this is the case, I'm happy to update this PR to modify that behaviour too.

### DIFF
--- a/libbeat/outputs/mode/lb/async_worker.go
+++ b/libbeat/outputs/mode/lb/async_worker.go
@@ -84,7 +84,7 @@ func (w *asyncWorker) connect() bool {
 			return false
 		}
 
-		logp.Info("Connect failed with: %v", err)
+		logp.Err("Connect failed with: %v", err)
 
 		cont := w.backoff.Wait()
 		if !cont {
--- a/libbeat/outputs/mode/lb/sync_worker.go
+++ b/libbeat/outputs/mode/lb/sync_worker.go
@@ -84,7 +84,7 @@ func (w *syncWorker) connect() bool {
 			return false
 		}
 
-		logp.Info("Connect failed with: %v", err)
+		logp.Err("Connect failed with: %v", err)
 
 		cont := w.backoff.Wait()
 		if !cont {
--- a/libbeat/outputs/mode/single/single.go
+++ b/libbeat/outputs/mode/single/single.go
@@ -137,7 +137,7 @@ func (s *Mode) publish(
 		resetFail := false
 
 		if err := s.connect(); err != nil {
-			logp.Info("Connecting error publishing events (retrying): %s", err)
+			logp.Err("Connecting error publishing events (retrying): %s", err)
 			goto sendFail
 		}
 


ruflin:
Hi @devenney 

Thanks for opening this PR. We had previously quite a few discussion on this log message if it should be error or info. Some discussions can for example be found here: https://github.com/elastic/beats/issues/875 We decided to go with Info as retrying could happen quite often and only log an error if it completely fails. The issues was kind of improved by setting the default log level to info for the 5.0 releases.

Let me know your thoughts.

devenney:
The problem I see is that info is generally too verbose for production. Perhaps it's personal preference but I only care about the errors, whether they're transient or not. In the case described above it was a real error which would retry forever without success as the DNS record was incorrectly entered. Other machines had Filebeat processes which presumably cached the old DNS so the lack of error logs on fresh machines meant we responded too slow - thankfully in staging but others may not be so lucky. 

Perhaps there's a distinction to be drawn between temporary connection issues and those requiring intervention, but I'd argue that it's impossible to distinguish - this should be the user's responsibility. And honestly, in production, seeing a transient connection problem is quite serious and requires addressing IMO. I see this case was put forward a couple of times in the referenced issue. 

So to summarise, I think this strategy is okay for those who leave logging at the new default level (probably most users) but not great for those who want less verbosity. In our case Filebeat appeared to just hang ad infinitum without any error output so it was the last part of the chain we investigated. I strongly believe this should be printed as Error as it logs real errors regardless of their frequency. 

Slightly off-topic: `filebeat.yml` and `filebeat.full.yml` still say the default log level is error btw - I had no idea the default was info. Unless the level is set in two places? I am not solely responsible for our deployment so it's maybe just a gap in my knowledge.

pinkisemils:
I really want to see this error in my logs next time Filebeat can't find it's way to Elasticsearch.

tsg:
> Slightly off-topic: filebeat.yml and filebeat.full.yml still say the default log level is error btw - I had no idea the default was info.

That's an oversight on my part, I'll fix it.

As part of https://github.com/elastic/beats/issues/1931 we reduced the verbosity of Info and Warn levels down to startup information and only important notices like this one, so there shouldn't be any spamming of the logs.

devenney:
@tsg: Is it an important notice or an important error? The thing is that it feels like a miscategorisation of this log output. I think the changes you referenced are actually a very good step towards having the perfect balance in the logging, but would you not agree that logging an _error_ as _info_ invalidates the levels? I feel that this forces everyone to run the tool with a log level of _info_ as _critical_ and _error_ do not actually output all errors as you would expect. This can be deceptive and trip up users as they see no errors and assume the problem lies elsewhere.

Additionally, @ruffin mentioned "[you] decided to go with Info as retrying could happen quite often" which suggests the original decision was made in the knowledge that the default logging level was _error_. As this is no longer the case it would seem that this particular conclusion of the original discussion is no longer relevant.

Apologies for being _that_ guy who turns a four character PR into a philosophical debate but I have a lot of love for this project and feel that this is quite an important issue which presumably reaches deeper than just this one line of code.

tl;dr: Error logging does not log all errors, therefore users must always run with a logging level of Info - thus the Error level feels moot.

tsg:
@devenney Yeah, I understand your point of view, and I'm also on the fence regarding the log level for that message. While it is an "error", it can be one that is corrected automatically without any need for user intervention. Filebeat is designed to handle transient errors without data loss, so in some cases there is no error from the user point of view. It could be that one Logstash instance was taken out of the rotation or that a Logstash instance didn't start yet. In these cases, an Info message seems more appropriate.

As the person that designed the whole outputs code, I know that @urso also has some opinions over this.

urso:
+1000 for this PR

Please also update the load balancing workers:
- https://github.com/elastic/beats/blob/master/libbeat/outputs/mode/lb/async_worker.go#L79
- https://github.com/elastic/beats/blob/master/libbeat/outputs/mode/lb/sync_worker.go#L79

There have been many changes to outputs since #875, which was used on another related error potentially spamming logs (even on transient failures) masking potential problems in logs. This PR about (re-)connect failures will not be printed per event, but per failed connection-attempt subject to exponential backoff. Plus backoff support was improved in last few months to remember state instead of being reset between attempts. 

If reconnect is required and succeeds immediately, no error message will be printed.

tsg:
Ok, I thought @urso would be against it. In this case, I'm also +1 on the change :-).

devenney:
All done - and somehow caused the build environment gods to smile upon me. 😆 Let me know if there is any further work required on my side.

ruflin:
@devenney Thanks for pushing the discussion. I remember that we also set it to Info as some users were confused by it, but I agree it is better to have one error too much then one too few. +1 on this change.

LGTM

PR #2184, opened 2016-08-08
Add cgroup metrics to system/process MetricSet

andrewkroh:
This adds metrics from the cpu, cpuacct, memory, and blkio cgroup subsystems. Cgroup metrics are not reported for processes that are members of the default cgroup ('/').

### DIFF
--- a/metricbeat/etc/fields.yml
+++ b/metricbeat/etc/fields.yml
@@ -1662,6 +1662,350 @@
                   type: long
                   description: >
                     The shared memory the process uses.
+            - name: cgroup
+              type: group
+              description: >
+                Metrics and limits from the cgroup of which the task is a member.
+                cgroup metrics are reported when the process has membership in a
+                non-root cgroup. These metrics are only available on Linux.
+              fields:
+                - name: id
+                  type: keyword
+                  description: >
+                    The ID common to all cgroups associated with this task.
+                    If there isn't a common ID used by all cgroups this field will be
+                    absent.
+
+                - name: path
+                  type: keyword
+                  description: >
+                    The path to the cgroup relative to the cgroup subsystem's mountpoint.
+                    If there isn't a common path used by all cgroups this field will be
+                    absent.
+
+                - name: cpu
+                  type: group
+                  description: >
+                    The cpu subsystem schedules CPU access for tasks in the cgroup.
+                    Access can be controlled by two separate schedulers, CFS and RT.
+                    CFS stands for completely fair scheduler which proportionally
+                    divides the CPU time between cgroups based on weight. RT stands for
+                    real time scheduler which sets a maximum amount of CPU time that
+                    processes in the cgroup can consume during a given period.
+
+                  fields:
+                    - name: id
+                      type: keyword
+                      description: ID of the cgroup.
+
+                    - name: path
+                      type: keyword
+                      description: >
+                        Path to the cgroup relative to the cgroup subsystem's
+                        mountpoint.
+
+                    - name: cfs.period.us
+                      type: long
+                      description: >
+                        Period of time in microseconds for how regularly a
+                        cgroup's access to CPU resources should be reallocated.
+
+                    - name: cfs.quota.us
+                      type: long
+                      description: >
+                        Total amount of time in microseconds for which all
+                        tasks in a cgroup can run during one period (as defined by
+                        cfs.period.us).
+
+                    - name: cfs.shares
+                      type: long
+                      description: >
+                        An integer value that specifies a relative share of CPU time
+                        available to the tasks in a cgroup. The value specified in the
+                        cpu.shares file must be 2 or higher.
+
+                    - name: rt.period.us
+                      type: long
+                      description: >
+                        Period of time in microseconds for how regularly a cgroup's
+                        access to CPU resources is reallocated.
+
+                    - name: rt.runtime.us
+                      type: long
+                      description: >
+                        Period of time in microseconds for the longest continuous period
+                        in which the tasks in a cgroup have access to CPU resources.
+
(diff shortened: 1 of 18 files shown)


metricbeat/module/system/process/_meta/fields.yml
```diff
@@ -82,3 +82,347 @@
...
+              type: keyword
+              description: ID of the cgroup.
+
+            - name: path
+              type: keyword
+              description: >
+                Path to the cgroup relative to the cgroup subsystem's
+                mountpoint.
+
+            - name: cfs.period.us
```
ruflin: We should add this to our conventions: us = microseconds

metricbeat/module/system/process/_meta/fields.yml
```diff
@@ -82,3 +82,347 @@
...
+                that have elapsed.
+
+            - name: stats.throttled.periods
+              type: long
+              description: >
+                Number of times tasks in a cgroup have been throttled (that is,
+                not allowed to run because they have exhausted all of the
+                available time as specified by their quota).
+
+            - name: stats.throttled.nanos
```
ruflin: This one should also be added to our standards

metricbeat/module/system/process/_meta/fields.yml
```diff
@@ -82,3 +82,347 @@
...
+              type: long
+              description: >
+                The total time duration (in nanoseconds) for which tasks in a
+                cgroup have been throttled.
+
+        - name: cpuacct
```
ruflin: Should we use cpu_acct here?

metricbeat/etc/fields.yml
```diff
@@ -1662,6 +1662,350 @@
...
+
+                    - name: stats.user.nanos
+                      type: long
+                      description: CPU time consumed by tasks in user mode.
+
+                    - name: stats.system.nanos
+                      type: long
+                      description: CPU time consumed by tasks in user (kernel) mode.
+
+                    - name: percpu
```
ruflin: As far as I understand this is going to be a list. Access with be in Kibana with `percup.1`, `percup.2` etc.? Does it perhaps make more sense to call it `cpu`?

metricbeat/etc/fields.yml
```diff
@@ -1662,6 +1662,350 @@
...
+                        The maximum amount of user memory in bytes (including file
+                        cache) that tasks in the cgroup are allowed to use.
+
+                    - name: mem.failures
+                      type: long
+                      description: >
+                        The number of times that the memory limit (mem.limit.bytes) was
+                        reached.
+
+                    - name: memsw.usage.bytes
```
ruflin: `mem.swap.usage.bytes` or `mem_swap.usage.bytes`?

metricbeat/etc/fields.yml
```diff
@@ -1662,6 +1662,350 @@
...
+                        The maximum amount for the sum of memory and swap usage
+                        that tasks in the cgroup are allowed to use.
+
+                    - name: memsw.failures
+                      type: long
+                      description: >
+                        The number of times that the memory plus swap space limit
+                        (memsw.limit.bytes) was reached.
+
+                    - name: kmem.usage.bytes
```
ruflin: `mem.kernel.*`?

metricbeat/etc/fields.yml
```diff
@@ -1662,6 +1662,350 @@
...
+                        The maximum amount of kernel memory that tasks in the cgroup are
+                        allowed to use.
+
+                    - name: kmem.failures
+                      type: long
+                      description: >
+                        The number of times that the memory limit (kmem.limit.bytes) was
+                        reached.
+
+                    - name: kmem_tcp.usage.bytes
```
ruflin: `mem.kernel.tcp.*`?

metricbeat/etc/fields.yml
```diff
@@ -1662,6 +1662,350 @@
...
+                        Memory limit for the hierarchy that contains the memory cgroup,
+                        in bytes.
+
+                    - name: stats.hierarchical_memsw_limit.bytes
+                      type: long
+                      description: >
+                        Memory plus swap limit for the hierarchy that contains the
+                        memory cgroup, in bytes.
+
+                    - name: stats.inactive_anon.bytes
```
ruflin: `stats.anon.inactive.bytes` combined with `stats.anon.active.bytes` above?

metricbeat/etc/fields.yml
```diff
@@ -1662,6 +1662,350 @@
...
+                      type: long
+                      description: >
+                        Anonymous and swap cache on inactive LRU list, including tmpfs
+                        (shmem), in bytes
+
+                    - name: stats.inactive_file.bytes
```
ruflin: `stats.file.inactive.bytes` with `stats.file.active.bytes` above

metricbeat/module/system/process/process.go
```diff
@@ -3,12 +3,19 @@
...
+
 	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/logp"
 	"github.com/elastic/beats/metricbeat/mb"
 
+	"github.com/elastic/gosigar/cgroup"
 	"github.com/pkg/errors"
 )
 
+var debugf = logp.MakeDebug("system-process")
```
ruflin: Probably we should call this `system.process` to be consistent with our naming

ruflin:
LGTM. I left some comments related to naming of the fields but I think that is something we can also improve up in at a later stage.

ruflin:
@andrewkroh One thing I realised now looking at the code again is if we should add a config option "cgroup" to only enable the cgroup fields if it is set to true? WDYT?

PR #2185, opened 2016-08-08
Make crosscompiling and packaging part of each beat

ruflin:
* Add crosscompile targets and environment config variables to each beat
* Add descriptions variable to beat Makefile
* Remove old beat specific config files in packer
* Move all building artifacts to ${BUILD_DIR} in each beat
* Introduce `make package` on the top level that does the packaging of all beats and copying of build files into the specific directories for uploading
* Remove obsolete Makefile in packer
* Unify beat dir to be for all beats the same
* Make build for internal beats better documented on why it is a special case
* Apply new packager structure to beat generator
* Move Makefile to dev-tools/packer directory
* Move upload commands to top Makefile
* Adjust path for deb-rpm-s3 script
* Adjust gpg path

Additional changes

* Change build target from `build` to `{beatname}` to make it clearer what is actually build
* Change build target from `buildbeat.test` to `{beatname}.test` to make it clearer what is actually build
* Remove old godep variable as not needed anymore because of vendor directories

This is part of #2251


libbeat/scripts/Makefile
```diff
@@ -43,6 +43,9 @@ DOCKER_COMPOSE?=docker-compose -f ${PWD}/../testing/environments/base.yml -f ${P
 DOCKER_CACHE?=1 # If set to 0, all docker images are created without cache
 GOPACKAGES_COMMA_SEP=$(subst $(space),$(comma),$(strip ${GOPACKAGES}))
 PYTHON_ENV?=${BUILD_DIR}/python-env
+BUILDID?=$(shell git rev-parse HEAD)
+CROSSCOMPILE_TARGETS?="linux/amd64 linux/386 windows/amd64 windows/386 darwin/amd64"
+CROSSCOMPILE_TARGETS_CGO?="linux/amd64 linux/386"
```
tsg: There might be some confusion here. We cross compile Cgo beats to all targets as well. Just that the Linux targets need to be done on an older Debian version to link to an as-old-as-possible libc (so that they work on as many systems as possible).

ruflin: Agree. The name of these variables is not good. I need to find a better name. I think I left somewhere a TODO for it.

libbeat/scripts/Makefile
```diff
@@ -313,13 +320,72 @@ install-home:
...
 # Create binary packages for the beat
-package: create-packer
-	cd dev-tools/packer; \
-	export ES_BEATS=${ES_BEATS} ; \
-	make deps images all
+package:
+	if [ $(CGO) = true ]; then \
+		 $(MAKE) prepare-package-cgo; \
+	else \
+		$(MAKE) prepare-package; \
```
tsg: I think this calls the docker commands in `prepare-package` before building the images.

libbeat/scripts/Makefile
```diff
@@ -313,13 +320,72 @@ install-home:
...
+		-v $(abspath ..):/source \
+		-e PACK=${BEATNAME} \
+		-e BEFORE_BUILD=before_build.sh \
+		-e SOURCE=/source \
+		-e TARGETS=${CROSSCOMPILE_TARGETS_CGO} \
+		-e BUILDID=${BUILDID} \
+		tudorg/beats-builder-deb6 \
+		${BEAT_DIR}
+
+include $(ES_BEATS)/dev-tools/packer/scripts/Makefile
```
tsg: Bringing all the packer targets (`images`, `deps`, etc.) into the main Makefile might confusing at some point.

tsg: Perhaps we could call the Makefile in packer as a subprocess rather than include it?

Makefile
```diff
@@ -70,3 +70,9 @@ beats-dashboards:
 .PHONY: docs
 docs:
 	sh libbeat/scripts/build_docs.sh ${PROJECTS}
+
+.PHONY: package
+package:
+	$(foreach var,$(BEATS),$(MAKE) -C $(var) package || exit 1;)
```
tsg: This breaks building in parallel (i.e. `make -j4 package`) :(

tsg:
Another thing that I'm going to miss is the ability to run a single beat/OS combination. For example, it was possible to do: `make packetbeat/deb` to just build the debs for Packetbeat, which was nice when troubleshooting. Not sure if this is possible with this setup.

libbeat/scripts/Makefile
```diff
@@ -313,13 +320,72 @@ install-home:
...
-.PHONY: create-packer
-create-packer:
-	python ${ES_BEATS}/libbeat/scripts/create_packer.py --es_beats=${ES_BEATS} --beat_path=${BEAT_DIR} --beat=${BEATNAME}
+
+# Prepares for packaging. Builds binaries and creates homedir data
+.PHONY: prepare-package
+prepare-package: build/upload
+	# cross compile on ubuntu
+	docker run --rm \
+		-v $(abspath ${ES_BEATS}/dev-tools/packer/xgo-scripts):/scripts \
```
tsg: We're going to need to allow for the community beats to provide their own scripts folder / before_build script, otherwise they won't be able to install or adjust C deps.

libbeat/scripts/Makefile
ruflin: Yes, that is cleaner. I didn't to it initially as like this I don't have to pass all environment variables over. But this would be much cleaner.

libbeat/scripts/Makefile
```diff
@@ -313,13 +320,72 @@ install-home:
...
+		 $(MAKE) prepare-package-cgo; \
+	else \
+		$(MAKE) prepare-package; \
+	fi
+
+	# Generates the package.yml file with all information needed to create packages
+	echo "beat_name: ${BEATNAME}" > ${BUILD_DIR}/package.yml
+	echo "beat_url: https://${BEAT_DIR}" >> ${BUILD_DIR}/package.yml
+	echo "beat_repo: ${BEAT_DIR}" >> ${BUILD_DIR}/package.yml
+	echo "beat_description: ${BEAT_DESCRIPTION}" >> ${BUILD_DIR}/package.yml
```
ruflin: I would really like such a file in each beat that has the meta information of the beat. I think this is not only for packaging but could for example also be used to automatically fetch information from community beats like version, creator, description etc. 

Based on this I would probably rather call it `.beat.yml`? Other suggestions?

Makefile
ruflin: Wasn't aware we use this. Will look for a solution that allows this.

libbeat/scripts/Makefile
ruflin: Good point. It probably only worked in my tests because I have the images already.

libbeat/scripts/Makefile
ruflin: Isn't that what we have with the `make before-build` command which is part of each beat? Did I change something here which was possible before but now is not anymore?

ruflin:
@tsg To have `make packetbeat/deb` you now have to run `$PACKAGE=packetbeat/deb make package`. I remember I used the above quite frequently in the past when we had issues with the packaging but recently I rarely used it. So I was thinking having a little bit more complicated command would be ok?

libbeat/scripts/Makefile
tsg: TBH I'd rather make it `.package.yml` and put all variables needed in it, including cgo/not-cgo, the targets, the before_build script, etc.

tsg: Then it would probably be worth it to replace the Makefile in packer with a python script that's able to interpret this.

ruflin: I really like this idea as it reduces the complexity in the Makefile and keeps the packaging out of it, but at the same time allows us to do things that were not possible before.

ruflin: Downside is there will be some duplication with the Makefile like BEATNAME

libbeat/scripts/Makefile
ruflin: Done

libbeat/scripts/Makefile
ruflin: I now moved the image building of of package and added a note, that it has to be called first. The reason I didn't make it a dependency is that it only has to be run once and the packaging can be run without running it every time.

libbeat/scripts/Makefile
ruflin: I renamed the varialbes to `TARGETS` and `TARGETS_OLD`. Instead of `TARGETS_OLD` I was also considering `TARGETS_DEBIAN6` to make it more specific? WDYT?

Makefile
ruflin: I added the parallel builds as a TODO here: https://github.com/elastic/beats/issues/2251

dev-tools/deploy
```diff
@@ -12,17 +11,14 @@ def main():
     args = parser.parse_args()
 
     dir = os.path.dirname(os.path.realpath(__file__))
-    os.chdir(dir + "/packer")
+    os.chdir(dir + "/../")
     print("Getting dependencies")
-    check_call("make deps", shell=True)
     check_call("make clean", shell=True)
```
tsg: Is `make clean` still the right target here?

ruflin: Yes, this calls the global `clean` and cleans up all build directories.

ruflin: Hm, looking at `dev-tools/packer/scripts/Makefile` there is also a clean that should probably be called here?

Makefile
```diff
@@ -70,3 +71,12 @@ beats-dashboards:
 .PHONY: docs
 docs:
 	sh libbeat/scripts/build_docs.sh ${PROJECTS}
+
+.PHONY: package
+package:
+	$(MAKE) -C libbeat package-setup
+	$(foreach var,$(BEATS),SNAPSTHOT=$SNAPSHOT $(MAKE) -C $(var) package || exit 1;)
```
tsg: Typo in SNAPSHOT and needs brackets around the variable name, i.e. `$(SNAPSHOT)`.

tsg: Would be good if either here or in the beats Makefile it would print some messages like `Start building packages for ${BEAT}` and `Finished packages for ${BEAT}`. The unified build process can use these to print show some progress indications.

libbeat/scripts/Makefile
```diff
@@ -57,13 +68,13 @@ endif
 ### BUILDING ###
 
 # Builds beat
-.PHONY: build
-build: $(GOFILES)
+.PHONY: ${BEATNAME}
+${BEATNAME}: $(GOFILES)
```
tsg: The `.PHONY` line makes it always build it, even if there were no changes in `$(GOFILES)`. So in the current form it's kind of confusing, because when reading the Makefile the impression one gets is that the build is called only when the sources are changed.

That would be awesome (to do it only when the sources are changed), but it needs to include libbeat sources, which I think currently it doesn't. Should also include the vendor directory.

tsg: We can do that in a future PR as well, seems not related to this one.

Makefile
ruflin: Side note: I plan to make it in a future PR parallel builds again based on one of the tricks here: http://stackoverflow.com/questions/1681006/parallelization-of-recursive-jobs-in-gnu-make

tsg: Btw, parallelism should be only an option, I don't think it will be used in the unified build.

tsg:
jenkins, package it

tsg:
Before merging, we need to make sure these two targets (`package-upload` and `release-upload`) have an alternative: https://github.com/elastic/beats/blob/74c1d512694e02d1d7134d6f6ee42dce9ca02d9c/dev-tools/packer/scripts/Makefile#L71

Makefile
ruflin: Typo fixed and echo message with the above text added. It is part of "package" in the libbeat Makefile.

About parallel builds: Agree.

libbeat/scripts/Makefile
ruflin: Yes, this change kind of does not belong in this PR. As build is normally quite fast, I think it is not too big of an issue that it is always built, especially as having the right implementation with all the files changed is tricky. I suggest we remove $(GOFILES) in a follow up PR.

ruflin:
Should we move the commands `package-upload` and `release-upload` into to Makefile in the `beats` directory? As this also applies to all beats ...

ruflin:
@tsg The same applies for `s3-nightlies-upload`.

ruflin:
Here is the commit that moves the commands to the top level Makefile. I also renamed the targets to be kind of consistent: https://github.com/elastic/beats/pull/2185/commits/2f9600c44588d80e14fe336aa3d4388934318d92

tsg:
@ruflin I think you also need to change the paths here: https://github.com/elastic/beats/blob/master/dev-tools/packer/docker/deb-rpm-s3/deb-rpm-s3.sh#L34

Unfortunately I don't know of a way to test this prior to the release, perhaps @andrewkroh can help?

andrewkroh:
> Unfortunately I don't know of a way to test this prior to the release, perhaps @andrewkroh can help?

Since the script doesn't have a built-in dry-run option, I recommend changing the [destination bucket](
https://github.com/elastic/beats/blob/master/dev-tools/packer/docker/deb-rpm-s3/deb-rpm-s3.sh#L26) to `beats-files` for your testing. You can verify that the contents of the bucket are correct (`aws s3 ls --recursive s3://beats-file/beats`) after publishing. Then when done testing just remove the directory (`aws s3 rm s3://beats-files/beats`).

Makefile
```diff
@@ -70,3 +71,30 @@ beats-dashboards:
 .PHONY: docs
 docs:
 	sh libbeat/scripts/build_docs.sh ${PROJECTS}
+
+.PHONY: package
+package:
+	$(MAKE) -C libbeat package-setup
+	$(foreach var,$(BEATS),SNAPSHOT=$SNAPSHOT $(MAKE) -C $(var) package || exit 1;)
```
tsg: `$SNAPSHOT` should be `$(SNAPSHOT)` here.

tsg:
Currently `make package` is failing for me with:

```
echo 7db14ead16dafa0518ea9c68f6dbe8886fd49d6b > /Users/tsg/src/github.com/elastic/beats/filebeat/build/upload/build_id.txt
echo "Finished packages for filebeat
bash: -c: line 0: unexpected EOF while looking for matching `"'
bash: -c: line 1: syntax error: unexpected end of file
make: *** [package] Error 2
```

ruflin:
@tsg Should be fixed now. Can you try again? " was missing in the added echo command :-(

tsg:
jenkins, package it

tsg:
One more path changed needed. The `gpg_key` path [here](https://github.com/ruflin/beats/blob/5abbd2bd12e8a32f0bfd8019a4646b3ee96f5f7d/dev-tools/packer/docker/deb-rpm-s3/deb-rpm-s3.sh#L29) needs to be `"/beats-packer/dev-tools/packer/docker/deb-rpm-s3/elasticsearch.asc"`. Other than that the `upload-package` target seems to work.

ruflin:
jenkins, package it

Issue #2137, opened 2016-07-30
Using Metricbeat's system module inside Docker


andrewkroh:
# Running Metricbeat in Docker

Goal: We want to be able to monitor the host machine from within a container.

#### The Command:

```
sudo docker run \
  --rm \
  --volume=/proc:/hostfs/proc:ro \
  --volume=/sys/fs/cgroup:/hostfs/sys/fs/cgroup \
  --net=host \
  --name=metricbeat \
  akroh/metricbeat:5.0.0-alpha6-SNAPSHOT \
  -system.hostfs=/hostfs \
  -E output.elasticsearch.enabled=false \
  -E output.console.pretty=false
```

This would mount the host’s filesystem under /rootfs inside the container. This gives Metricbeat’s system module access to data from the host. Normally Metricbeat reads the following data, but it will need to read all of this data relative to /rootfs to capture data from the host.

- cpu
  - /proc/stat
- memory
  - /proc/meminfo
- load
  - /proc/loadavg
- core
  - /proc/stat
- diskio (gopsutil)
  - /proc/diskstats
- filesystem
  - /etc/mtab
  - statfs (syscall)
- network (gopsutil)
  - /proc/net/dev
  - This does not appear to work, /rootfs/proc/net/dev and /proc/net/dev are exactly the same and contain container metrics not host metrics. You must start the container with --net=host to see the host’s network stats.
- process
  - /proc
  - /proc/[pid]/stat
  - /proc/[pid]/status
  - Username lookup from uid will likely fail. (?)
- cgroup
  - /proc/cgroups
  - /proc/self/mountinfo
  - /proc/[pid]/cgroup
  - /sys/fs/cgroup or /cgroup (for some systems like RHEL 6)
    - This filepath is read from /proc/self/mountinfo so it must be interpreted as relative to /rootfs when running inside a container.

#### Required Changes

- [x] For gopsutil, set the path to `/etc`, `/proc`, and `/sys` relative to the `-system.hostfs` value by setting `HOST_ETC`, `HOST_PROC`, and `HOST_SYS` environment variables.
- [x] For gosigar, set the global `Procd` variable relative to the `-system.hostfs` value when `GOOS=linux`.
- [x] For gosigar, make the path to `/etc/mtab` configurable and set it relative to the `-system.hostfs` value.
- [x] Need ability to filter filesystems by mountpoint. (Just use a `drop_event` filter, see system.filesystem docs.)
- [x] Update docs to show how to run Metricbeat inside a container.

[...]

PR #2187, opened 2016-08-08
Add option for configuring the location of the container's host filesystem

andrewkroh:
Commit to review: 4b48e2f (isolates these changes from those added in #2184)

This allows the system module to read metrics related to the host machine rather than the container. When using the -system.hostfs flag, directories like /proc and /sys are assumed to be mounted relative to the path given in the value. Here's an example usage.

`docker run --volume=/:/hostfs:ro --net=host --name=metricbeat -d metricbeat -system.hostfs=/hostfs`

And here is a more restrictive example which does not mount all of the host's filesystems inside the container. system/filesystem will not be able to report metrics for the host's filesystem.

`docker run --volume=/proc:/hostfs/proc:ro --volume=/sys/fs/cgroup:/hostfs/sys/fs/cgroup --net=host --name=metricbeat -d metricbeat -system.hostfs=/hostfs`

### DIFF
--- a/metricbeat/module/system/process/process.go
+++ b/metricbeat/module/system/process/process.go
@@ -3,12 +3,14 @@
 package process
 
 import (
+	"fmt"
 	"runtime"
 	"strconv"
 
 	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/logp"
 	"github.com/elastic/beats/metricbeat/mb"
+	"github.com/elastic/beats/metricbeat/module/system"
 
 	"github.com/elastic/gosigar/cgroup"
 	"github.com/pkg/errors"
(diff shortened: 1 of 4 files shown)


ruflin:
@andrewkroh I merged #2184 so this now needs a rebase.

ruflin:
* Is it required to have this as a flag or could we somehow enabled this also as a config option?
* In the commit message you described very nicely how it works and what the options are. This should be added to the docs.

andrewkroh:
Rebased.

> Is it required to have this as a flag or could we somehow enabled this also as a config option?

I did see your comment about this in issue #2137, but I already had it implemented it like this so I put out the PR, but I do think it could be config file option since it is so easy to set config variables on the CLI now. The one issue I ran into was figuring out how to best route the configuration value down into the module without making a global variable. Probably the best option is to write the value to [`ModuleConfig`](https://github.com/elastic/beats/blob/master/metricbeat/mb/mb.go#L102-L112), but not make it configurable by the individual modules.

> In the commit message you described very nicely how it works and what the options are. This should be added to the docs.

Yeah, I haven't updated any of the docs to discuss cgroups yet. I will create an issue for this.

ruflin:
If somehow possible, we should make all options part of the config file, especially with centralised configuration management in mind. I see the command line flags mainly for overwriting existing config options if needed. If it is a config file option, no special flag is needed as the default "overload flag options" can be used.

Alternatively `metricbeat.system.hostfs` could be used. I think it is fine to use the global namespace for global options?

ruflin:
@andrewkroh I merged this one. I would be nice if we could find a "nice" option to have it in the config file and somehow pass it to the Module.

PR #2188, opened 2016-08-08
Enhance kafka topic selection

urso:


metricbeat/metricbeat.full.yml
```diff
@@ -402,14 +402,10 @@ output.elasticsearch:
   # to.
   #hosts: ["localhost:9092"]
 
-  # The Kafka topic used for produced events. If use_type is set to true, the
```
ruflin: Do we already have some docs about how to use format string? If not, we should definitively create an issue for it.

libbeat/docs/outputconfig.asciidoc
```diff
@@ -543,11 +543,8 @@ The cluster metadata contain the actual Kafka brokers events are published to.
 
 ===== topic
 
-The Kafka topic used for produced events. If `use_type` is set to true, the topic will not be used.
-
-===== use_type
-
-Set Kafka topic by event type. If `use_type` is false, the `topic` option must be configured. The default is false.
+The Kafka topic used for produced events. The setting can be a format string
+using any event field. To set the topic from document type use `%{[type]}`.
```
ruflin: Should type be our default recommendation? Bringing up as we had recently some discussion around type. Perhaps better use beat.name by default?

libbeat/outputs/kafka/config.go
```diff
@@ -14,8 +14,6 @@ type kafkaConfig struct {
 	TLS             *outputs.TLSConfig `config:"tls"`
 	Timeout         time.Duration      `config:"timeout"             validate:"min=1"`
 	Worker          int                `config:"worker"              validate:"min=1"`
-	UseType         bool               `config:"use_type"`
-	Topic           string             `config:"topic"`
```
ruflin: Topic still exist, don't we need it here?

libbeat/outputs/kafka/kafka_integration_test.go
```diff
@@ -57,11 +59,13 @@ func newTestKafkaClient(t *testing.T, topic string) *client {
 
 func newTestKafkaOutput(t *testing.T, topic string, useType bool) outputs.Outputer {
 
+	if useType {
```
ruflin: Is this to test BC compatibility?

metricbeat/metricbeat.full.yml
urso: see #2132

libbeat/docs/outputconfig.asciidoc
urso: Default recommendation should be a static string (we do not set any default and require user to set the topic name). type is commonly used with filebeat + multiple prospectors setting document_type.

libbeat/outputs/kafka/config.go
urso: Yes, Topic still exists, but not required here. `outil.BuildSelectorFrom...` looks for `topic` and `topics` building a topic-selector similar to pipelining/index selection in elasticsearch output.

libbeat/outputs/kafka/kafka_integration_test.go
urso: use_type has been removed from config. This is in order to keep the tests working without much change, as some tests use the event `type` to choose topic dynamically and other tests set static topic name. Alternatively one can change tests to pass format-string in `topic` and remove `useType`.

ruflin:
@urso needs rebase

PR #2190, opened 2016-08-08
Update to sarama v1.10.0

urso:


libbeat/outputs/kafka/config.go
```diff
@@ -14,34 +14,58 @@ type kafkaConfig struct {
...
 	defaultConfig = kafkaConfig{
-		Hosts:           nil,
-		TLS:             nil,
-		Timeout:         30 * time.Second,
-		Worker:          1,
+		Hosts:   nil,
+		TLS:     nil,
+		Timeout: 30 * time.Second,
+		Worker:  1,
+		Metadata: metaConfig{
```
ruflin: What will happen if these settings are used with an older version of Kafka?

urso: sarama lib supports kafka 0.8 - 0.10 . Settings should be valid for all kafka versions. Lib might internally ignore settings if they do not apply to the version being used. AFAIK metadata updates and publisher settings should have same effect for kafka versions.

libbeat/outputs/kafka/client.go
```diff
@@ -108,16 +108,28 @@ func (c *client) AsyncPublishEvents(
 			topic = event["type"].(string)
 		}
 
+		var ts time.Time
+		if c.config.Version.IsAtLeast(sarama.V0_10_0_0) {
```
ruflin: A code comment would be nice that quickly explains why whis is only possible with >= 0.10

libbeat/outputs/kafka/config.go
```diff
@@ -58,5 +82,13 @@ func (c *kafkaConfig) Validate() error {
 		return fmt.Errorf("compression mode '%v' unknown", c.Compression)
 	}
 
+	if _, ok := kafkaVersions[c.Version]; !ok {
+		return fmt.Errorf("unknown/unsupported kafka version '%v'", c.Version)
+	}
+
+	if c.Username != "" && c.Password == "" {
+		return fmt.Errorf("password must not be empty when SASL user is configured")
```
ruflin: I would rephrase this to "empty password. Password config option must be set when Username is configured."

libbeat/outputs/kafka/kafka.go
```diff
@@ -51,6 +51,23 @@ var (
 		"gzip":   sarama.CompressionGZIP,
 		"snappy": sarama.CompressionSnappy,
 	}
+
+	kafkaVersions = map[string]sarama.KafkaVersion{
```
ruflin: I was first thinking something is wrong below as some "entries" were missing because the sorting on the right is not based on the version. I would probably sort based on the values on the right side. But that is more my personal preference. Just wanted to note it as I was first confused.

libbeat/outputs/kafka/kafka.go
```diff
@@ -225,6 +250,7 @@ func newKafkaConfig(config *kafkaConfig) (*sarama.Config, error) {
 		retryMax = 1000
 	}
 	k.Producer.Retry.Max = retryMax
+	// TODO: k.Producer.Retry.Backoff = ?
```
ruflin: Should this stay in?

ruflin:
Doesn't this PR also need an update the libbeat config file?

libbeat/outputs/kafka/client.go
urso: done

libbeat/outputs/kafka/config.go
urso: changed to `password must be set when username is configured`

libbeat/outputs/kafka/kafka.go
urso: change ordering

libbeat/outputs/kafka/kafka.go
urso: yes. It's some samara internal config with default 100ms kinda overlapping with libbeat functionality (not critical, though). Added `TODO` so I will reconsider a parameter maybe another time instead of hiding the fact something might be "incomplete".

urso:
> Doesn't this PR also need an update the libbeat config file?

Ups, yes.

urso:
@ruflin added missing config file updates + added options to docs

PR #2191, opened 2016-08-08
Change field type system.process.cpu.start_time from keyword to date

### LINKED ISSUE #1565: Make Topbeat's proc.cpu.start_time a timestamp
The current value of `proc.cpu.start_time` is not very usable and is possibly broken. I propose that we make this field a timestamp in the index template and format it appropriately. **This will be a breaking change to the data mapping.**

Here's what I see now.

OS|`start_time`
----|---------------
darwin|"Apr30"
linux|"21:25"

andrewkroh:
Fixes #1565


andrewkroh:
Needs merged to the 5.0 branched.

PR #2194, opened 2016-08-09
Remove mention of close_timeout

dedemorton:
This mention of `close_timeout` was overlooked in #2158


ruflin:
Thanks for spotting this.

PR #2195, opened 2016-08-09
Update naming conventions with micro and nano seconds

ruflin:
@andrewkroh I'm somehow tempted to go with ns for nanoseconds but ns could also stand for namespace and other things. WDYT?

### DIFF
--- a/libbeat/docs/event-conventions.asciidoc
+++ b/libbeat/docs/event-conventions.asciidoc
@@ -37,4 +37,7 @@ List of standardised words and units across all Beats:
 * ms: millisecond, millis
 * mb: megabytes
 * msg: message
+* nanos: nanoseconds
+* norm: normalized
+* us: microseconds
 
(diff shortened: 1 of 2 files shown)


tsg:
LGTM. Both `ns` and `nanos` seem fine to me.

andrewkroh:
@ruflin  I like `nanos` more, but `ns` is more consistent. So I'm torn. Both are good with me. I can change my code that uses `nanos` if you want to go with `ns`.

ruflin:
3 people that are torn between the two. @brandonmensing one more opinion? :-D

brandonmensing:
I did a little searching for a precedent in our favorite bigger sibling project and found that they are also inconsistently using both 'ns' and 'nanos' :) 

In this context, I would go with consistency w/r/t the other conventions nearby. So 'ns' has my vote to make it a little easier on users who are using 'ms' and 'us'.

ruflin:
PR for renaming opened here: https://github.com/elastic/beats/pull/2252

PR #1952, opened 2016-07-04
Switch default level to INFO

tsg:
This is part of #1931 and fixes #1734.

Some things to note/discuss:

* `-v` does nothing by default now. However, it's possible that you
set it to error in the config file and then use -v to go back to
info level at the CLI. So I kept the option.
* the `logging.level: debug` from the default configuration also
doesn nothing now, unless you also specify some selectors. So I added
also a `logging.selectors: ["*"]` to the default short config.
* While not all of #1931 is done yet, I think we can proceed with the
default level change to fix #1734 and then we do more INFO/WARN cleanups
in future PRs.

PR #2199, opened 2016-08-09
Correct the default log level in configs

tsg:
Part of #1931. The default log level was changed to INFO in PR #1952,
but the change wasn't reflected in the default configuration files. This
fixes it. The issue was signaled in #2183.

Also fixes the logging level in the docs.


ruflin:
LGTM. 

I was looking in our CHANGELOG for an entry related to the log level but couldn't find one. I think it is worth to mention there.

tsg:
Hmm, yes, good point. But I need to add that in the 5.0 branch since the switch is happening in alpha5.

PR #2215, opened 2016-08-10
Add system test to load dashboards

ruflin:
* Move test_version test to base tests as it should not be in base.py

### DIFF
--- a/libbeat/tests/system/test_dashboard.py
+++ b/libbeat/tests/system/test_dashboard.py
@@ -1,0 +1,26 @@
+from base import BaseTest
+
+import os
+import subprocess
+
+
+class Test(BaseTest):
+    def test_load_dashboard(self):
+        """
+        Test loading dashboards for all beats
+        """
+        beats = ["metricbeat", "packetbeat", "filebeat", "winlogbeat"]
+
+        for beat in beats:
+            command = "../../../dev-tools/import_dashboards.sh -l http://"+ self.get_elasticsearch_host() + " -dir ../../../"+ beat + "/etc/kibana"
+
+            if os.name == "nt":
+                command = "..\..\..\dev-tools\import_dashboards.ps1 -dir ..\..\..\\" + beat + "\etc\kibana"
+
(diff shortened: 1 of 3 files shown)


dev-tools/import_dashboards.sh
```diff
@@ -171,10 +171,10 @@ fi
 if [ -d "${DIR}/index-pattern" ]; then
     for file in ${DIR}/index-pattern/*.json
     do
-        NAME=`awk '$1 == "\"title\":" {gsub(/[",]/, "", $2); print $2}' ${file}`
```
monicasarbu: I wouldn't remove this as it gets the name of the index pattern from the `title` field available in the json file, that can be different than the name of the file.  Think of `packetbeat-*` that is available under packetbeat.json

libbeat/tests/system/test_dashboard.py
```diff
@@ -0,0 +1,23 @@
+from base import BaseTest
+
+import os
+import subprocess
+
+
+class Test(BaseTest):
+    def test_load_dashboard(self):
```
monicasarbu: great idea to add a test for loading the dashboards.

dev-tools/import_dashboards.sh
ruflin: @monicasarbu I changed it as the code did not work anymore. I changed it now to `${NAME}-*`. This is only a temporary fix and this script should be removed when https://github.com/elastic/beats/pull/2155 is merged.

ruflin:
@monicasarbu I just realised that even though I made the tests compatible with windows, it does not run on windows as our libbeat tests are not run on windows. As these tests need the full environment, it is good to run them in libbeat. As the code with https://github.com/elastic/beats/pull/2155 will be identical for windows and others, probably it will be sufficient to run them only on linux systems.

dev-tools/import_dashboards.sh
monicasarbu: I have found the problem. #2217 is fixing this issue.

monicasarbu:
The failing test should be fixed in https://github.com/elastic/beats/pull/2220.

ruflin:
@monicasarbu I merged #2220 and rebased on top of master.

Issue #2208, opened 2016-08-09
Metricbeat dashboards do not import correctly


aleph-zero:
On alpha5, when loading the dashboards from the import_dashboards.sh script, they do not show up in Kibana. 

For confirmed bugs, please report:
- Version: 
Alpha5
- Operating System:
OSX 
- Steps to Reproduce: 
1. cd metricbeat-5.0.0-alpha5-darwin-x86_64/kibana && ./import_dashboards.sh
2. go to Kibana, try to Open a dashboard. There are none.

tsg:
Related ES ticket: https://github.com/elastic/elasticsearch/issues/19893

monicasarbu:
Removing Metricbeat label, as the same error appears when loading any dashboard/index pattern.

tsg:
A simple workaround is to replace one line in the `import_dashboards.sh` script, like this:

```
diff --git a/dev-tools/import_dashboards.sh b/dev-tools/import_dashboards.sh
index 231c97e..f5c5466 100755
--- a/dev-tools/import_dashboards.sh
+++ b/dev-tools/import_dashboards.sh
@@ -8,7 +8,7 @@

 # The default value of the variable. Initialize your own variables here
 ELASTICSEARCH=http://localhost:9200
-CURL=curl
+CURL="curl -H Expect:"
 KIBANA_INDEX=".kibana"
 DIR=.
 BEAT_CONFIG=".beatconfig"
```

tsg:
@dedemorton Do you think we could document this workaround in the getting started for the alpha5, similar to how we had a workaround for alpha4? I can then also add a notice to the blog post.

tsg:
I made a proposal for the docs in #2218.

tsg:
Closing this one as the workaround is in place in master and it's documented in the alpha5 docs. The Elasticsearch fix https://github.com/elastic/elasticsearch/pull/19904 is also merged.

cwurm:
Hm, the workaround doesn't work for me if I run the script with `-u username:password`.

ruflin:
@cwurm We missed that in case of `-u` CURL variable is overwritten locally: https://github.com/tsg/beats/blob/179e8770c6e0843e84a092a12706fdc70a002f19/dev-tools/import_dashboards.sh#L75 I'm going to reopen this issue.

monicasarbu:
@cwurm can you please try to change this line:
```
 CURL="curl --user ${USER}"
```
with:
```
[...]

PR #2216, opened 2016-08-10
Implements workaround for dashboard loading

tsg:
Workaround for #2208. Elasticsearch alpha5 doesn't like the "Expect: 100-continue"
header that is added by curl. We should remove this workaround after the next
release is out.


ruflin:
LGTM.

PR #2217, opened 2016-08-10
Fix indentation in the generated index pattern

monicasarbu:
Use indent=2 when dumping the json into the file, in order for `awk` to work in the import_dashboards.sh script when extracting the index pattern title.


ruflin:
LGTM

PR #2219, opened 2016-08-10
[WIP] Experimental config

ruflin:
Over time we will add additional MetricSets without having a major release. Adding a MetricSet does not necessarly directly mean its data structure is already final. Still we would like to get the community to test it. One option is to have this MetricSets only in master and the nightly build must be used to test it. For most users this is a big hurdle. My suggestion is to introduce an "experimental" config option.

`metricbeat.experimental` can be enabled in the config to also use experimental metricsets. This will allow users to test new Metricsets which are already shipped with the binary but not available by default. Also on our doc pages we would state that these are experimental. In case a Metricset is enabled but is experimental and experimental was not enabled, an error message is printed.

This will allow us to add MetricSets with each new minor release without it being already stable.

Below is a first code draft on how the experimental could look like in a MetricSet. A function `AddExperimentalMetricSet` was introduced. This code is not complete yet and only to show / discuss the idea.

### DIFF
--- a/metricbeat/beater/config.go
+++ b/metricbeat/beater/config.go
@@ -5,5 +5,6 @@ import "github.com/elastic/beats/libbeat/common"
 // Config is the root of the Metricbeat configuration hierarchy.
 type Config struct {
 	// Modules is a list of module specific configuration data.
-	Modules []*common.Config `config:"modules" validate:"required"`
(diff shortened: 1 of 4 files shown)


ruflin:
Closing for the moment as we are still discussing on the best approach here.

PR #2220, opened 2016-08-10
Generate the kibana type from the field type

monicasarbu:
As Kibana supports only few types (number, date, string, geo_ip), it complains when an index contains fields of unknown types, so we need to convert the field types to one of `number`, `date`, `string` or `geo_ip`.

For the field types that don't fit in these types, no `type` is set. (eg. `dict`, `list`).


CHANGELOG.asciidoc
```diff
@@ -31,6 +31,8 @@ https://github.com/elastic/beats/compare/v5.0.0-alpha5...master[Check the HEAD d
 ==== Bugfixes
 
 *Affecting all Beats*
+- Use indentation when generating the index pattern {pull}2217[2217]
```
monicasarbu: This is added here as it's not part of #2117.

ruflin: Do we even have to add it here as the script was not part of alpha5? Bugfix for something that was not release yet?

CHANGELOG.asciidoc
```diff
@@ -31,6 +31,8 @@ https://github.com/elastic/beats/compare/v5.0.0-alpha5...master[Check the HEAD d
...
+- Generate the kibana type of each index pattern field from the field type, used in the template {pull}2220[2220]
```
ruflin: Is that a bug fix?

monicasarbu: yes, both are bug fixes as Kibana complains after you import the index patterns in Kibana.

ruflin:
LGTM, waiting for green.

Issue #2164, opened 2016-08-03
Document common practice for bandwidth throttling in FAQ


brandonmensing:
Users often want a Beat to limit network usage but it makes more sense for the network stack in the OS to handle this for many reasons. Let's provide an explanation of this and potentially an example of how this could be done on a common platform.

tsg:
Googling a bit, [this](http://www.cyberciti.biz/faq/linux-traffic-shaping-using-tc-to-control-http-traffic/) seems to contain what we need, but I didn't test it yet.

Something like:
```
# delete existing rules
tc qdisc del dev eth1 root

# Create a class that limits to 512 Kbps
tc qdisc add dev eth1 root handle 1:0 htb default 10
tc class add dev eth1 parent 1:0 classid 1:10 htb rate 512kbps ceil 640kbps prio 0

# Use iptables to mark the packets going to Logstash
iptables -A OUTPUT -t mangle -p tcp --dport 5044 -j MARK --set-mark 10

# Assign the marked connections to the traffic shaping class
tc filter add dev eth1 parent 1:0 prio 0 protocol ip handle 10 fw flowid 1:10
```

tsg:
After testing I managed to get the right effect with the following commands:

```
tc qdisc add dev $DEV root handle 1: htb
tc class add dev $DEV parent 1:1 classid 1:10 htb rate 50kbps ceil 50kbps
tc filter add dev $DEV parent 1:0 prio 1 protocol ip handle 10 fw flowid 1:10
iptables -A OUTPUT -t mangle -p tcp --dport 5044 -j MARK --set-mark 10
```

This sets a limit of 50 kbps on the TCP connections towards port 5044. I.e. you would normally execute this on the Beat side. It doesn't slow down any other connections.

Where should we put this example? FAQ? Troubleshooting guide?

ruflin:
@tsg I would put it into FAQ. We should add some details there on why it is not part of beats itself and that is the way we recommend to do it.

dedemorton:
Added to the FAQ with in #2225 (and cherry-picked into relevant branches)

PR #2225, opened 2016-08-10
Add FAQ topic about bandwidth throttling

dedemorton:
@tsg Here's my first pass at resolving issue #2164. Please review. I wasn't sure if we wanted a generic topic for all beats, or only for Filebeat, so I added it to all beats. 

@andrewkroh Is there something we can say about bandwidth throttling on Windows (I don't think the Linux example is useful for Winlogbeat), or should I remove this topic from Winlogbeat entirely?

### DIFF
--- a/filebeat/docs/faq.asciidoc
+++ b/filebeat/docs/faq.asciidoc
@@ -54,4 +54,5 @@ Filebeat uses a newline character to detect the end of an event. If lines are ad
 harvested, a newline character is required after the last line, or Filebeat will not read the last line of
 the file.
 
+include::../../libbeat/docs/faq-limit-bandwidth.asciidoc[]
 include::../../libbeat/docs/shared-faq.asciidoc[]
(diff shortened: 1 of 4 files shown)


andrewkroh:
I have never tried this on Windows so maybe we should omit it for Winlogbeat. But it seems like it is possible to implement a traffic shaping using a Group Policy. I found this http://www.windowsnetworking.com/articles-tutorials/windows-server-2012/QoS-Windows-Server-2012-Part3.html

tsg:
LGTM, thanks @dedemorton. Requires a rebase.

dedemorton:
@tsg This should be ready to merge now. In addition to the rebase, I moved the content to a separate asciidoc file to make it easier to exclude the content from Winlogbeat (I'm trying to avoid conditional coding in favor of asciidoc include statements because I think the include statements are easier for maintainers to understand).

Issue #637, opened 2016-01-06
Add an new input type to backfill gzipped logs


lminaudier:
Hi,

Following [this discussion on filebeat forum](https://discuss.elastic.co/t/backfilling-old-logs-with-filebeat/38428/2) I would like to ask if it is possible to implement a solution to easily backfill old gzipped logs with filebeat.

The proposed solution mentioned in the topic is to add a new dedicated `input_type`.

It is also mentioned in the topic that when filebeat reaches the end of input on `stdin` it does not give you the hand back and waits for new lines which makes things hard to script to perform backfilling.

What are your thoughts on this ?

Thanks for your hard work.

ruflin:
I would see the implementation as following:

* Prospector: A gzip prospector would be added. Harvesters would only be opened based on filenames (no inode etc.). It is assume, that the files are not renamed and don't have to be tracked. A file has a completion state and if it is once completed, it is never read again. This simplifies the implementation. If a .gz file with a new filename is found, a harvester is started.
* Harvester: A gzip harvester is added. It unzips and reads the full file only once. After finishing reading the file, the harvester stops and is never started again. The offset is only stored, if it is interrupted in the middle of reading. In a first implementation, this could even be removed for simplicity. 

This would be nice to have but I think it is not on the top of our priority list. A community contribution here would be more then welcome.

For your second issue about running filebeat only until first completion, lets refer to this issue: https://github.com/elastic/filebeat/issues/219

lminaudier:
Thanks for the fast reply and the pointer to the issue.

I will try to look at your implementation proposal. I am still quite new to Golang and the project, so I can't promise anything :)

ruflin:
@lminaudier Always here to help.

mryanb:
This would be a great feature addition. Currently the splunk-forwarder does something similar and will index log rotated files that have been gziped automatically.

Ragsboss:
+1. Is anyone working on this? If not I could possibly take it up..

ruflin:
@Ragsboss I don't think anyone is working on that. Would be great to have a PR for that to discuss the details.

cFire:
@Ragsboss @ruflin I'm delighted to see there's someone looking to pick this up. Is this happening? The reason I ask is because it may be possible for me to spend some time helping out with this in lieu of building another solution for gziped logs to use internally.

Ragsboss:
[...]

PR #2227, opened 2016-08-10
[637] Add support to index gzipped logs

Ragsboss:
This is a proposed approach to fix https://github.com/elastic/beats/issues/637. In this approach, we reuse existing input type 'log' to transparently process gzipped files. So there is no configuration change for filebeat users except for ensuring that the filename patterns match gzipped file names. The filebeat recognizes gzipped files using extension '.gz', any other extension will not work.

Ruflin noted an alternative approach of introducing a brand new input type to deal with gzipped files. I'm ok about either approaches. But reusing existing input type seemed more intuitive from filebeat users viewpoint and for me it was easier to implement. I hope this change gives Ruflin a better view of how this approach looks like. If we still feel a new input type is better, we can certainly go down that path..

Few pending things that we can do once (and if) we agree this approach is acceptable
- Test for cases where a regular file gets log rotated and compressed. In this case compressed file will have a different inode and today this works assuming filename patterns don't match .gz files, but with this support, .gz files will typically be matched.
- Write new tests
- Go's compress/gzip package doesn't support seeking, so resuming is not supported. Need to decide if we want to support this or gracefully handle this
- The tabs/indentation seems to be different. I use Sublime 3 on iMac and tried using both spaces and tab, but still the whitespace diffs are showing up...


ruflin:
Thanks a lot for the contribution. It's is very good to have the code base for discussion. Overall the implementations seems to be simpler then I expected which is great. 

I would prefer that we modify the implementation to a different input_type instead of relying on the file ending. Otherwise I already see the request to support other file endings etc. and I'm sure the support request for zip files is also only around the corner. Like this we can separate the implementations and don't add complexity to the already too complex log harvester. There only problem at the moment is that the separation in the code currently in the code is not 100% clean, but I'm working on this to clean it up and parts of it can also happen after this PR got merged. For inspiration on how to deal with an additional input type, best have a look at `stdin`.

One additional point I would like to have it separate and probably the most important one is the one you mentioned above about duplicated data. If someone wants to feed in the compressed files he should really have to think about how to prevent duplicate data. Currently I don't see an option on how we could prevent duplicated data if someone uses `/var/log/*` as pattern. As soon as a new `.gz` file is generated, all files will be resent. We are thinking about adding some fingerprint support to detect if a file changed or is the same. This could potentially mitigate the problem but it will take some more time until we have this feature. Instead of trying to figure out here how to prevent duplicates I suggest to leave it to the user for the moment.

Which feature should be supported for `.gz` prospector/harvester? My suggestion would be:

Supported:
* Multiline
* Json
* File identifier (in case or renaming)

Not supported:
* Offset

For the Offset this would mean, either the file is full read or it is not read. The state is only updated when the harvester finished the file. It will never opened again (no size comparison etc.). In case filebeat is stopped in the middle of reading the file, it will start over again after the restart.

Let me know your thoughts on the above.

For your indentation issue: We use the golang standard. To do it automatically, just run `make fmt` and it will do it for you.

Ragsboss:
@ruflin thank you for your insightful review. I understand your approach of using a separate input type. But my thought process was to explore the possibility of using a single harvester with differences between file types encapsulated behind a single interface. Fortunately code already had such an interface viz. FileSource. Adding support for newer compression types is achieved by having a new type implement this interface appropriately. So it's modular and extensible. 

As we add more configuration, features etc to Harvester/Prospector etc, it will have two problems with a separate input type approach. One is config duplication for users and the other is code maintenance cost for FileBeat developers. Splunk forwarder also works this way. See http://docs.splunk.com/Documentation/Splunk/latest/Data/Monitorfilesanddirectories#How_Splunk_software_monitors_archive_files. No special configuration is needed to index compressed files.

But if you still feel separate input_type is the way to go, I'll perhaps wait for your refactoring changes to go in before picking this up. I agree with your observation that this is more trickier to implement on top of the current state of the code.

For event duplication, we can use the file name blacklist/whitelist approach. See https://answers.splunk.com/answers/50653/log-rotation-for-compressed-files.html. 

For resumability, it's ok to start with no support. But whether we duplicate events or skip events is a choice. In general, erring on event duplication seems safer as that means we are at-least giving (albeit extra) visibility. But I'm also not comfortable making that hard-coded assumption as valid for all use cases. We could consider exposing a config option like 'errOnEventDuplication' or such.

Good to know about `make fmt` - will do.

ruflin:
@Ragsboss After thinking more about this I came up with a third solution which is a "compressed" param in the config. Like this it is extendable to other compressions. I took part of your code and built this PR out of it: https://github.com/elastic/beats/pull/3070 Would be good to hear your feedback on that.

Ragsboss:
@ruflin I'm sorry for not being active as I had switched jobs and new one is quite packed.. Went through your PR and provided my comments.. It looks good overall.

kaiyan-sheng:
Hello @Ragsboss, seems like there is no progress on this PR for more than 6 months. Will close this PR for now. Thanks!

PR #2233, opened 2016-08-11
Actually it should be compatible with `logstash`

alexanderilyin:
It really unexpected when similar features in tools which were created to be used together have problems with compatibility.

### DIFF
--- a/filebeat/docs/reference/configuration/filebeat-options.asciidoc
+++ b/filebeat/docs/reference/configuration/filebeat-options.asciidoc
@@ -184,6 +184,8 @@ Options that control how Filebeat deals with log messages that span multiple lin
 
 The following example shows how to configure Filebeat to handle a multiline message where the first line of the message begins with a bracket (`[`).
 
+NOTE: `pattern` is not compatible with format used in `logstash`. See <<regexp-support>> for details or you will face with errors like: `ERR Stop Harvesting. Unexpected encoding line reader error: error parsing regexp: invalid escape sequence: '\d'`
+
 [source,yaml]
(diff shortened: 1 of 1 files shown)


alexanderilyin:
I've signed **CLA**. Just saying.

filebeat/docs/reference/configuration/filebeat-options.asciidoc
```diff
@@ -184,6 +184,8 @@ Options that control how Filebeat deals with log messages that span multiple lin
 
 The following example shows how to configure Filebeat to handle a multiline message where the first line of the message begins with a bracket (`[`).
 
+NOTE: `pattern` is not compatible with format used in `logstash`. See <<regexp-support>> for details or you will face with errors like: `ERR Stop Harvesting. Unexpected encoding line reader error: error parsing regexp: invalid escape sequence: '\d'`
```
dedemorton: @alexanderilyin  We've made some changes to our regexp support in Filebeat so that `\d` will be supported in 5.0.
I agree with you that we can be more explicit here, but including the details of the message is too explicit. Can you remove the note here and instead change the description next to pattern (line 213 in your file) to say, "Specifies the regular expression pattern to match. Note that the regexp patterns supported by Filebeat differ somewhat from the patterns supported by Logstash. See <<regexp-support>> for a list of regexp patterns supported by Filebeat." I think it would be also useful to add a similar note to the page about regexp support. You can do that when you resubmit your PR, or I can do that. Thanks for helping us improve the docs!

dedemorton:
Closing because this issue is fixed in #2983

PR #2235, opened 2016-08-11
Improve tail_files documentation

### LINKED ISSUE #1458: [Filebeat] tail_files option doesn't seem to work as specified
Hi all.

I'm setting up filebeat, and I encounter issues with the ```tail_files``` option: it doesn't work as expected.
Instead of starting from the end of each log file, it starts from the beginning.

I wanted to do as described [here](https://discuss.elastic.co/t/filebeat-needs-to-be-restarted-in-order-to-send-logs-to-logstash-dynamically/39994), ie:
- 1st start with ```tail_files: true```
- Restart filebeat with ```tail_files: false``` (once it saved the current offset of each file).

But with my config, it just starts from the beginning of each file.

I used filebeat 1.2.1 on Ubtuntt 14.04 LTS, and my config is as follows:

Main config file ```/etc/filebeat/filebeat.yml```:

```YAML
filebeat:
  prospectors:
    -

  registry_file: /var/lib/filebeat/registry
  config_dir: /etc/filebeat/conf.d
output:
  logstash:
    hosts: ["logstash-host:5044",]
    worker: 2
    loadbalance: true
shipper:
logging:
  files:
    path: /var/log/filebeat
    name: filebeat.log
    rotateeverybytes: 10485760 # = 10MB
  level: warning
```

In ```/etc/filebeat/conf.d```, I have 1 conf file per domain. Here's 1 example:

```YAML
filebeat:
  prospectors:
    -
      paths:
        - "/var/log/apache2/access.log"
      input_type: log
      fields:
        format: "APACHE"
        application: "Appli1"
        index: "logs"
        kafka_topic: "logs"
      fields_under_root: true
      document_type: Apache
      tail_files: true
``` 

So the ```tail_files``` option is set in the "child" config files. Could this impact the behaviour?

Best regards

Geoff

ruflin:
Add details that it doesn't apply to files seen before.

Closes #1458

### DIFF
--- a/filebeat/docs/reference/configuration/filebeat-options.asciidoc
+++ b/filebeat/docs/reference/configuration/filebeat-options.asciidoc
@@ -383,6 +383,8 @@ lines are discarded. The default is 500.
 
 If this option is set to true, Filebeat starts reading new files at the end of each file instead of the beginning. When this option is used in combination with log rotation, it's possible that the first log entries in a new file might be skipped. The default setting is false.
 
+This option applies to files that Filebeat has not already processed. If you ran Filebeat previously and the state of the file was already persisted, `tail_files` will not apply. Harvesting will continue at the previous offset. To apply `tail_files` to all files, you must stop Filebeat and remove the registry file. Be aware that doing this removes ALL previous states.
+
 NOTE: You can use this setting to avoid indexing old log lines when you run Filebeat on a set of log files for the first time. After the first run, we recommend disabling this option, or you risk losing lines during file rotation.
 
 ===== backoff


filebeat/docs/reference/configuration/filebeat-options.asciidoc
```diff
@@ -383,6 +383,8 @@ lines are discarded. The default is 500.
 
 If this option is set to true, Filebeat starts reading new files at the end of each file instead of the beginning. When this option is used in combination with log rotation, it's possible that the first log entries in a new file might be skipped. The default setting is false.
 
+In case filebeat was run previously and a state was already persisted for the file, `tail_files` will not apply. Harvesting will continue at the previous offset. `tail_files` only applies to files not seen before. To apply `tail_files` to all files, filebeat has to be stopped and the `registry_file` has to be removed. Be aware that this removes ALL previous states.
```
dedemorton: Suggested edits:

This setting applies to files that Filebeat has not already processed. If you ran Filebeat previously and the state of the file was already persisted, `tail_files` will not apply. Harvesting will continue at the previous offset. To apply `tail_files` to all files, you must stop Filebeat and remove the `registry_file`. Be aware that doing this removes ALL previous states.

ruflin: perfect :-)

ruflin:
@dedemorton New version pushed with your suggested changes.

dedemorton:
LGTM, except the backticks got removed. I'd add backticks around `tail_files` but looking at this again, I would say registry file instead of `registry_file`. My concern here is that the formatting might lead someone to believe that you are talking about the `registry_file` entry in the config rather than the actual file.

ruflin:
Fixed

PR #2239, opened 2016-08-11
Introduce outputs.Data

urso:
outputs.Event wraps events of type common.MapStr by adding additional metadata
to events. Currently these metadata are for internal use in publisher workers
and load-balanced output workers to share results of intermediate computations
in order to provide better consistency on failures.

Some potential use-cases:
- computed kafka partition (might be required to be 'consistent')
- computed hashes for 'routing'
- store results of computed indices/keys for use on retry
- randomized kafka key, ES document id (for ES or logstash) guaranteeing
  non-duplicates in case of errors


libbeat/outputs/mode/mode_test.go
```diff
@@ -1,309 +0,0 @@
-// +build !integration
```
ruflin: What happened to these tests?

urso: This file did not contain any tests or any piece of code used somewhere. Missing delete from the past.

ruflin:
Some thoughts / notes:

* I like that data is encapsulated into Event as it shows that there could be more things as part of an event
* There is a potential conflict of naming here as for example in Filebeat we call `Event.Data` already `Event`. Perhaps we can come up with a better naming. Same applies to meta data. There is meta data also in `Event.Data` which could become confusing
* There are some ouptut specifics above which is perhaps better handled inside the specific output
* We will still need a general way to add additional data to the event which applies to all outputs like the unique id.

Sorry, no solutions offered in the above comments so far, but I hope it can trigger more discussions.

urso:
1. I think using `Event.Data` is ok, as `Event.Data` is what will be published. For libbeat outputs everything is just an event.

2. Maybe we want to rename `Event.Meta` to `Event.Context`, to not confuse it with potential event metadata added in the future

Idea in future PR is to have some computation steps (outputs will register on) in order to pre-compute values. Sample use-case: pre-compute document id for elasticsearch + store events with context/meta in persistent-queue to drive the final outputs from.

Having meta and some signalers + cancel support it almost becomes similar to [net/context package](https://github.com/golang/net/tree/master/context). Maybe something to look at in the future. Anyway, I like the idea of [net/context](https://github.com/golang/net/tree/master/context) providing context value-lookup + non-destructive write by wrapping the old context. Maybe we should add a similar strategy here, in order to prevent potential races if multiple outputs are configured.

ruflin:
2. I like the suggestion of replacing `Meta` with `Context`.

Instead of having `Event.Data` it could be turned around to be `Data.Event` and `Data.Context`. The output has a Data object which contains the Event and the Context of the Data object.

monicasarbu:
What about using `Event.ExportData` and `Event.Context`?

libbeat/outputs/console/console.go
```diff
@@ -47,15 +47,15 @@ func (c *console) Close() error {
 func (c *console) PublishEvent(
 	s op.Signaler,
 	opts outputs.Options,
-	event common.MapStr,
+	event outputs.Data,
```
ruflin: can we renamed the field itself also to data? Otherwise we have inception below :-)

libbeat/outputs/elasticsearch/client.go
```diff
@@ -193,8 +194,8 @@ func (client *Client) Clone() *Client {
 // events not published or confirmed to be processed by elasticsearch will be
 // returned. The input slice backing memory will be reused by return the value.
 func (client *Client) PublishEvents(
-	events []common.MapStr,
-) ([]common.MapStr, error) {
+	events []outputs.Data,
```
ruflin: Not sure what the right name here is: `datas`?

libbeat/outputs/elasticsearch/client.go
```diff
@@ -285,10 +286,11 @@ func eventBulkMeta(
 		Index bulkMetaIndex `json:"index"`
 	}
 
+	data := event.Event
```
ruflin: event

libbeat/outputs/elasticsearch/client.go
```diff
@@ -275,7 +276,7 @@ func bulkEncodePublishRequest(
 func eventBulkMeta(
 	index outil.Selector,
 	_ *outil.Selector,
-	event common.MapStr,
+	event outputs.Data,
```
ruflin: data

libbeat/outputs/elasticsearch/client.go
urso: `data` is already the plural form. singular is `datum`

libbeat/outputs/elasticsearch/client.go
```diff
@@ -251,31 +252,31 @@ func bulkEncodePublishRequest(
...
 	mkMeta = eventBulkMeta
 	if pipeline != nil {
 		mkMeta = eventIngestBulkMeta
 	}
 
-	okEvents := events[:0]
-	for _, event := range events {
-		meta := mkMeta(index, pipeline, event)
-		if err := body.Add(meta, event); err != nil {
+	okEvents := data[:0]
```
ruflin: This should probably be okData

libbeat/outputs/elasticsearch/client.go
```diff
@@ -217,23 +218,23 @@ func (client *Client) PublishEvents(
...
-		failedEvents = events
+		failedEvents = data
 	} else {
 		client.json.init(result.raw)
-		failedEvents = bulkCollectPublishFails(&client.json, events)
+		failedEvents = bulkCollectPublishFails(&client.json, data)
 	}
 
-	ackedEvents.Add(int64(len(events) - len(failedEvents)))
+	ackedEvents.Add(int64(len(data) - len(failedEvents)))
```
ruflin: ackedData

libbeat/outputs/elasticsearch/client_test.go
```diff
@@ -72,9 +73,9 @@ func TestCollectPublishFailsNone(t *testing.T) {
 	response := []byte(`{"items": [` + strings.Repeat(item, N) + `]}`)
 
 	event := common.MapStr{"field": 1}
-	events := make([]common.MapStr, N)
+	events := make([]outputs.Data, N)
```
ruflin: data / events mix

libbeat/outputs/kafka/client.go
```diff
@@ -83,31 +84,31 @@ func (c *client) Close() error {
...
-		batch: events,
+		count: int32(len(data)),
+		batch: data,
 		cb:    cb,
 	}
 
 	ch := c.producer.Input()
 
-	for _, event := range events {
+	for _, d := range data {
```
ruflin: I prefer `d` over `datum` you used above 👍

libbeat/outputs/logstash/logstash_integration_test.go
```diff
@@ -409,12 +409,12 @@ func testLogstashElasticOutputPluginCompatibleMessage(t *testing.T, name string,
 	defer es.Cleanup()
 
 	ts := time.Now()
-	event := common.MapStr{
+	event := outputs.Data{Event: common.MapStr{
```
ruflin: data

libbeat/outputs/logstash/logstash_integration_test.go
```diff
@@ -463,12 +463,14 @@ func testLogstashElasticOutputPluginBulkCompatibleMessage(t *testing.T, name str
 	defer es.Cleanup()
 
 	ts := time.Now()
-	events := []common.MapStr{
+	events := []outputs.Data{
```
ruflin: data

libbeat/outputs/mode/lb/context.go
```diff
@@ -34,8 +34,8 @@ type eventsMessage struct {
 	worker       int
 	attemptsLeft int
 	signaler     op.Signaler
-	events       []common.MapStr
-	event        common.MapStr
+	data         []outputs.Data
+	datum        outputs.Data
```
ruflin: I prefer d over datum

ruflin:
@urso There are still some naming inconsistency. I merged the PR as I think it is better to have these changes in as they affect quite a large number of files. The naming cleanups can also be done when the code is touched again.

PR #2240, opened 2016-08-11
Log errors when harvester could not be started

ruflin:
So far no errors were logged when a harvester could not be strated. This should be a very rare case in general but it is good to have log messages in case it happens.

This is a small improvement based on https://github.com/elastic/beats/issues/1913


urso:
maybe you want to include the current 'case' like `on new file`, `on old file`, `on truncated file`... ?

ruflin:
@urso Log message enhanced. I was first thinking there is an other log message before that tells where it is, but it is Debug and not error. Thanks for spotting this.

urso:
small change, big improvement 👍

Issue #2171, opened 2016-08-03
beat.sh script cannot use a custom config path


tsg:
In alpha5 we added the option to use the `-c` multiple times. This somewhat breaks the beat.sh scripts that we put in the `/usr/bin` for DEB/RPM installations (the beat binary is not in $PATH by default). 

Here is an example of a script:

```
#!/bin/bash

# Script to run {.beat_name} in foreground with the same path settings that
# the init script / systemd unit file would do.

/usr/share/filebeat/bin/filebeat -e \
  -c /etc/filebeat/filebeat.yml \
  -path.home /usr/share/filebeat \
  -path.config /etc/filebeat \
  -path.data /var/lib/filebeat \
  $@
```

One of the things that was possible in alpha4 was to specify a custom config file, e.g.:

```
filebeat.sh -e -c filebeat.yml
```

This was overriding the default config file from the script with `filebeat.yml`. With alpha5, however, the Beat combines the two configuration files, which comes as a surprise for the user.

tsg:
IMO, the best solution would be if the default value for `-c` is `filebeat.yml` which is searched in `path.config`. Then we don't need to put it explicitly in the script, and the user can override it with `-c` passed to the script.

Have to check if this is a viable solution, I know we had issues with this approach before, but I think that was before introducing the config paths.

urso:
I think by default (if no -c is given), filebeat should load config file from ${path.config}/filebeat.yml by default. Solution might be to remove `-c` from script altogether. If removing `-c` doesn’t work properly we indeed have a bug in filebeat itself.

tsg:
Fixed by #2245.

PR #2245, opened 2016-08-11
Look for the config files relative to the path.config flag

tsg:
This is a proposal fix for #2171. There's a bit of a chicken and egg
problem here, since defining the paths requires the configuration file
and the other way around. The implemented logic is to:

* if the `-path.config` flag is used, look for the config file relative to
  it
* if not, but `-path.home` flag is used, look for the config file relative
  to the home path
* else, look relative to the current working directory

The last point is a BWC break, because in 1.x the configuration file was relative to the binary location.

### DIFF
--- a/libbeat/cfgfile/cfgfile.go
+++ b/libbeat/cfgfile/cfgfile.go
@@ -14,7 +14,7 @@ var (
 	// The default config cannot include the beat name as it is not initialized
 	// when this variable is created. See ChangeDefaultCfgfileFlag which should
 	// be called prior to flags.Parse().
-	configfiles = flagArgList("c", "beat.yml", "Configuration file `path`")
+	configfiles = flagArgList("c", "beat.yml", "Configuration file, relative to path.config")
 	overwrites  = common.NewFlagConfig(nil, nil, "E", "Configuration overwrite")
 	testConfig  = flag.Bool("configtest", false, "Test configuration and exit.")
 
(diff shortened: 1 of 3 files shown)


tsg:
@urso, you should probably review this one.

libbeat/cfgfile/cfgfile.go
```diff
@@ -14,7 +14,7 @@ var (
 	// The default config cannot include the beat name as it is not initialized
 	// when this variable is created. See ChangeDefaultCfgfileFlag which should
 	// be called prior to flags.Parse().
-	configfiles = flagArgList("c", "beat.yml", "Configuration file `path`")
+	configfiles = flagArgList("c", "beat.yml", "Configuration file, relative to path.config")
```
ruflin: An absolute path also works, correct?

libbeat/cfgfile/cfgfile.go
```diff
@@ -105,9 +98,27 @@ func Load(path string) (*common.Config, error) {
...
+		cfgpath = *configPath
+	} else if *homePath != "" {
+		cfgpath = *homePath
+	}
+
 	if path == "" {
-		config, err = common.LoadFiles(configfiles.list...)
+		list := []string{}
+		for _, cfg := range configfiles.list {
+			if !filepath.IsAbs(cfg) {
```
ruflin: yep, still works :-)

ruflin:
LGTM. @urso Do you also want to have a look?

urso:
LGTM

@tsg requires rebase.

tsg:
@urso rebased.

Issue #2237, opened 2016-08-11
Unable to filter winlogbeats events by keywords field


vbohata:
Version: v5.0.0-alpha5
Operating System: Windows Server 2012R2

I am unable to exclude events by keywords field. Config option " keywords: "Audit Success" " does not work, in beats log is: 

```
2016-08-11T10:05:43+02:00 WARN unexpected type []string in contains condition as it accepts only strings. 
2016-08-11T10:05:43+02:00 WARN unexpected type []string in contains condition as it accepts only strings. 
2016-08-11T10:05:43+02:00 WARN unexpected type []string in contains condition as it accepts only strings. 
2016-08-11T10:05:43+02:00 WARN unexpected type []string in contains condition as it accepts only strings. 
```

Config option " keywords: ["Audit Success"] " also does not work but there is no error in beats log.

Processors config:

```
processors:
  - drop_event:
      when:
        and:
          - equals:
              log_name: Security
          - contains:
              keywords: ["Audit Success"]
```

spacewander:
Try `keywords: "Audit Success"`?

vbohata:
I tried it, with this, there is error in log: unexpected type []string

ruflin:
@vbohata Could you provide the full error message?

@andrewkroh Not sure if this could be related to https://github.com/elastic/beats/pull/2209 ?

vbohata:
There is no more info in log. Just repeated "WARN unexpected type []string in contains condition as it accepts only strings." line.

andrewkroh:
In the event log record, `keywords` is an array. The `contains` filter only accepts a string value as its input which is what causes the warning and the filter to not work. We should enhance it to check each element in the array.

And like @spacewander  said, it should be configured as `keywords: "Audit Success"`.

andrewkroh:
I opened PR #2248 to make `contains` work on arrays of strings.

andrewkroh:
This should be fixed in 5.0.

PR #2248, opened 2016-08-11
Enhance contains condition to work on fields that are arrays of string

andrewkroh:
Addresses #2237

### DIFF
--- a/libbeat/processors/condition.go
+++ b/libbeat/processors/condition.go
@@ -269,26 +269,36 @@ func (c *Condition) checkEquals(event common.MapStr) bool {
 }
 
 func (c *Condition) checkContains(event common.MapStr) bool {
-
+outer:
 	for field, equalValue := range c.contains {
-
 		value, err := event.GetValue(field)
 		if err != nil {
 			return false
 		}
 
-		sValue, err := extractString(value)
-		if err != nil {
-			logp.Warn("unexpected type %T in contains condition as it accepts only strings. ", value)
+		switch value.(type) {
+		case string:
+			if !strings.Contains(value.(string), equalValue) {
+				return false
+			}
+		case *string:
+			if !strings.Contains(*value.(*string), equalValue) {
+				return false
+			}
+		case []string:
+			for _, s := range value.([]string) {
+				if strings.Contains(s, equalValue) {
+					continue outer
+				}
+			}
 			return false
-		}
-		if !strings.Contains(sValue, equalValue) {
+		default:
+			logp.Warn("unexpected type %T in contains condition as it accepts only strings.", value)
 			return false
 		}
 	}
 
 	return true
-
 }
 
 func (c *Condition) checkRegexp(event common.MapStr) bool {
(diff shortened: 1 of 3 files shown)


libbeat/processors/condition.go
```diff
@@ -269,26 +269,36 @@ func (c *Condition) checkEquals(event common.MapStr) bool {
 }
 
 func (c *Condition) checkContains(event common.MapStr) bool {
-
+outer:
```
ruflin: argh :-(

libbeat/processors/condition.go
```diff
@@ -269,26 +269,36 @@ func (c *Condition) checkEquals(event common.MapStr) bool {
...
+				return false
+			}
+		case *string:
+			if !strings.Contains(*value.(*string), equalValue) {
+				return false
+			}
+		case []string:
+			for _, s := range value.([]string) {
+				if strings.Contains(s, equalValue) {
+					continue outer
```
ruflin: why not just `return true` here?

andrewkroh: That would be changing the behavior of the method. This method returns true iff all the `contains` conditions are satisfied. Returning true would cause the method to not evaluate any remaining conditions.

When I was adding this, I initially did `return true` only to realize I broken a test.

libbeat/processors/condition.go
andrewkroh: Why no love for the label?

libbeat/processors/condition.go
ruflin: Got it. Other option would be to package the second for loop into a function/method so "standard" continue could be used. I just think it makes it somehow harder to read.

ruflin:
LGTM

libbeat/processors/condition.go
ruflin: It's not supported in my brain :-D

libbeat/processors/condition.go
```diff
@@ -269,26 +269,36 @@ func (c *Condition) checkEquals(event common.MapStr) bool {
...
+		switch value.(type) {
+		case string:
+			if !strings.Contains(value.(string), equalValue) {
+				return false
+			}
+		case *string:
+			if !strings.Contains(*value.(*string), equalValue) {
+				return false
+			}
+		case []string:
```
monicasarbu: @andrewkroh why not ?

```
found := false
for _, s := range value.([]string) {
  if strings.Contains(s, equalValue) {
   found = true
  }
}
if !found {
  return false
}
```

monicasarbu:
@andrewkroh does it work without updating setContains() function?

monicasarbu:
@andrewkroh can you please add documentation for it?

libbeat/processors/condition.go
andrewkroh: They are equivalent if you add the required `break` statement to the snippet above. But what's there is a bit more concise.

andrewkroh:
> does it work without updating setContains() function?

Yes, `setContains` is for reading the configuration as I understand it. This PR does not make any changes to what is accepted as part of the configuration (it accepts a single string). The PR only enhances it to work properly when the field in the event is an array.

andrewkroh:
> can you please add documentation for it?

I updated to docs. See #2285

PR #2250, opened 2016-08-12
Make FAQs a single topic

dedemorton:
- Added float tags to make all FAQ topics for a beat appear on the same page
- Shortened titles (they are easier to scan, but not proper questions...an OK compromise, IMHO)

I haven't tried to reorder the topics. Some of the topics are in a shared file, `shared-faq.asciidoc`. In order to put the topics in some kind of logical order (alphabetical or some logical grouping), I would need to put all the content (from all the beats) into `shared-faq.asciidoc` and use `ifdef` statements to include content conditionally. I try to avoid conditional stuff because it can be messy for other people to figure out. Right now, the list of questions is short enough that the order doesn't seem very important to me. WDYT?

I do like that I can remove the hand-built list and rely on the build to generate the list of topics for me. 

Here's an example page so you can see what the pages look like when they are build: https://filebeatfaqchanges.firebaseapp.com/faq.html

### DIFF
--- a/filebeat/docs/faq.asciidoc
+++ b/filebeat/docs/faq.asciidoc
@@ -4,26 +4,17 @@
 This section contains frequently asked questions about Filebeat. Also check out the
 https://discuss.elastic.co/c/beats/filebeat[Filebeat discussion forum].
 
-* <<filebeat-network-volumes>>
-* <<filebeat-not-collecting-lines>>
-* <<filebeat-cpu>>
-* <<dashboard-fields-incorrect-filebeat>>
-* <<newline-character-required-eof>>
-* <<connection-problem>>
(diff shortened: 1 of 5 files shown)


urso:
maybe an alternative options might be using imports. like every beat has it's own faq.asciidoc including the generic (about connection errors and such) at the bottom.

urso:
LGTM

dedemorton:
@urso There are lots of ways to slice/dice in asciidoc. Having a separate source file for each shared FAQ topic would probably be a lot easier to maintain than conditional coding, so maybe that's the way to go at some point. Maybe not worth it now, though, since the FAQs are fairly short.

PR #2267, opened 2016-08-15
Copy beats directory over to vendor dir for generators

ruflin:
* Automatically copy over beats directory to vendor on setup. The git directory is removed.
* Make ES_BEATS depend on the vendor directory. This makes packaging possible
* Fix fmt check

Existing community beats generated with the previous version can run the following command to get the same result:

```
mkdir -p vendor/github.com/elastic/
cp -R ${GOPATH}/src/github.com/elastic/beats vendor/github.com/elastic/
```

Afterwards the line with `ES_BEATS` has to be changed to:

```
ES_BEATS?=./vendor/github.com/elastic/beats
```

These changes to not change how a beat is generated.

Vendoring / Versioning note:

It is up to the user to do decide which tool should be used for vendoring. That is the reason no glide file for example is added. Also the versioning is up to the user. We currently recommend to check in the vendor directory as it makes sure other developers can directly start working on the project. It is also a requirement to make packaging work out of the box.

### DIFF
--- a/generate/beat/{{cookiecutter.beat}}/Makefile
+++ b/generate/beat/{{cookiecutter.beat}}/Makefile
@@ -2,7 +2,7 @@ BEATNAME={{cookiecutter.beat}}
 BEAT_DIR={{cookiecutter.beat_path}}
 SYSTEM_TESTS=false
 TEST_ENVIRONMENT=false
-ES_BEATS?=${GOPATH}/src/github.com/elastic/beats
+ES_BEATS?=./vendor/github.com/elastic/beats
 GOPACKAGES=$(shell glide novendor)
 PREFIX?=.
 
@@ -11,8 +11,15 @@ PREFIX?=.
 
 # Initial beat setup
 .PHONY: setup
-setup:
+setup: copy-vendor
 	make update
+
+# Copy beats into vendor directory
+.PHONY: copy-vendor
+copy-vendor:
+	mkdir -p vendor/github.com/elastic/
+	cp -R ${GOPATH}/src/github.com/elastic/beats vendor/github.com/elastic/
+	rm -rf vendor/github.com/elastic/beats/.git
 
 .PHONY: git-init
 git-init:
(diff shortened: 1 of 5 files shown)


generate/metricbeat/metricset/{{cookiecutter.beat}}/Makefile
```diff
@@ -11,10 +11,17 @@ PREFIX?=.
...
+setup: copy-vendor
 	make create-metricset
 	make collect
 
+# Copy beats into vendor directory
+.PHONY: copy-vendor
+copy-vendor:
+	mkdir -p vendor/github.com/elastic/
+	cp -R ${GOPATH}/src/github.com/elastic/beats vendor/github.com/elastic/
+	rm -rf vendor/github.com/elastic/beats/.git
```
tsg: Rsync excluding `.git` might be faster, although probably doesn't make that much of a difference.

tsg: Consider also calling the `scripts/clean-vendor.sh` script to reduce the number of files.

tsg:
LGTM, I'm +1 for the change.

generate/metricbeat/metricset/{{cookiecutter.beat}}/Makefile
ruflin: * rsync: I was not 100% sure if rsync is installed by default on all supported platforms. That is why I stayed away from it
* clean-vendor: This would also remove some yaml files we need.

generate/beat/{{cookiecutter.beat}}/Makefile
```diff
@@ -2,7 +2,7 @@ BEATNAME={{cookiecutter.beat}}
 BEAT_DIR={{cookiecutter.beat_path}}
 SYSTEM_TESTS=false
 TEST_ENVIRONMENT=false
-ES_BEATS?=${GOPATH}/src/github.com/elastic/beats
```
yehaha9876: `ES_BEATS?=`  why have '?'

yehaha9876: OK, thanks!

PR #2268, opened 2016-08-15
Fix panic when hitting EOF on stdin input

tsg:
The stdin harverster type has no state, so do a nil check before
trying to update it.


ruflin:
LGTM. Would be nice if we would have a small test to verify the fix.

tsg:
@ruflin I managed to make a test that was failing before and is passing afterwards. I'll open a PR to  do a similar one in master.

ruflin:
LGTM. I restarted appveyor as it didn't start properly.

ruflin:
@tsg Thanks for the fix.

PR #2269, opened 2016-08-15
Make sure the etc/kibana/index-pattern exists 

monicasarbu:
This PR makes sure the etc/kibana/index-pattern directory exists before running `make update` that ganarates the index-pattern.

### DIFF
--- a/libbeat/scripts/Makefile
+++ b/libbeat/scripts/Makefile
@@ -240,6 +240,7 @@ update: python-env
 	cp ${ES_BEATS}/libbeat/docs/version.asciidoc docs/version.asciidoc
 
 	# Generate index-pattern
+	mkdir -p $(PWD)/etc/kibana/index-pattern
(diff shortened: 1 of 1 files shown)


ruflin:
LGTM

libbeat/scripts/Makefile
```diff
@@ -240,6 +240,7 @@ update: python-env
 	cp ${ES_BEATS}/libbeat/docs/version.asciidoc docs/version.asciidoc
 
 	# Generate index-pattern
+	mkdir -p $(PWD)/etc/kibana/index-pattern
```
ruflin: It would be nice to first remove all index-pattern and then only generate the new ones. This makes sure not old left overs stay in and the generation works as expected. We do the same for example for the config files.

monicasarbu: @ruflin So far we have only one index-pattern per Beat, so I added into the Makefile to delete the index template before generating a new one.

PR #2273, opened 2016-08-16
Update CONTRIBUTING.md (cherry-pick #2263 into 5.0)

dedemorton:


dedemorton:
@elastic/beats I figured it wouldn't hurt to cherry-pick this into 5.0. However, I'm not going to bother cherry-picking into earlier releases (unless you tell me to) because I figure most people will/should be looking at this file in master anyhow, right?

CONTRIBUTING.md
```diff
@@ -70,13 +67,13 @@ Packetbeat:
     $ cd beats/packetbeat
     $ make
 
-Some of the Beats might have extra development requirements, in which case a
-CONTRIBUTING.md file is find in the Beat directory.
+Some of the Beats might have extra development requirements, in which case you'll find a
```
ruflin: As packetbeat is the only one with a local file, we should probably remove it in the future and here link to a general developer guide which then links to packetbeat dev guide. This again points in the direction that all dev guide should leave in on place (@urso ;-) ).

ruflin:
@dedemorton I think we don't need to backport these kind of changes to any of the branches as it is very rare that someone looks at a branch for contribution notes I would assume.

PR #2274, opened 2016-08-16
Fix rotation issue - 1.3

ruflin:
This PR fixes the issue that sometimes on fast file rotation it could happen, that the same file was read twice. This does now no happen anymore. There is one downside to this change and that is on very low scan_frequency it can happen that some file are never read. I'm still investigating if there is an option to prevent this. The main problem here seems to be that states are stored based on the file name, means if there are 2 states with the same filename during one scan, they overwrite each other. This is fixed in 5.0 by removing the filepath as key for the state.


ruflin:
Closing this PR as it didn't solve all the race conditions and introduced other issues as it made the scanning slower.

PR #2278, opened 2016-08-16
Added deprecation notice for symlinks

tsg:
Also moved more changelog items into 1.3, as we'll be doing another
respin for sure.


ruflin:
LGTM

PR #2279, opened 2016-08-16
Add line number counter for multiline

ruflin:
This allows to indicate if an event was multiline or not. The number of lines will be put under the multiline namespace and looks as following:

```
{
  ...
  "message": "[2015] hello world\n  First Line\n  Second Line",
  "multiline": {
    "lines": 3
  },
  ...
}
```

See elastic#957

### DIFF
--- a/filebeat/harvester/reader/multiline.go
+++ b/filebeat/harvester/reader/multiline.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/common/match"
 	"github.com/elastic/beats/libbeat/logp"
 )
@@ -248,6 +249,12 @@ func (mlr *Multiline) clear() {
 
 // finalize writes the existing content into the returned message and resets all reader variables.
 func (mlr *Multiline) finalize() Message {
+
+	mlr.message.AddFields(common.MapStr{
+		"multiline": common.MapStr{
+			"lines": mlr.numLines,
+		},
+	})
 	// Copy message from existing content
 	msg := mlr.message
 	mlr.clear()
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -824,6 +824,11 @@ https://github.com/elastic/beats/compare/v5.2.0...v5.2.1[View commits]
 *Packetbeat*
 
 - Fix error in the NFS sample dashboard. {pull}3548[3548]
+- Add enabled config option to prospectors. {pull}3157[3157]
+- Add target option for decoded_json_field. {pull}3169[3169]
+- Add the `pipeline` config option at the prospector level, for configuring the Ingest Node pipeline ID. {pull}3433[3433]
+- Update regular expressions used for matching file names or lines (multiline, include/exclude functionality) to new matchers improving performance of simple string matches. {pull}3469[3469]
+- Add multiline line number to filebeat event {pull}2279[2279]
 
 *Winlogbeat*
 
--- a/filebeat/_meta/fields.common.yml
+++ b/filebeat/_meta/fields.common.yml
@@ -41,3 +41,8 @@
     - name: fileset.name
       description: >
         The Filebeat fileset that generated this event.
+
+    - name: multiline.lines
+      type: long
+      description: >
+        This is set in case of a multiline event and contains the number of lines that the multiline event consists of.
--- a/filebeat/docs/fields.asciidoc
+++ b/filebeat/docs/fields.asciidoc
@@ -959,6 +959,14 @@ The Filebeat module that generated this event.
 The Filebeat fileset that generated this event.
 
 
+[float]
+=== `multiline.lines`
+
+type: long
+
+This is set in case of a multiline event and contains the number of lines that the multiline event consists of.
+
+
 [[exported-fields-mysql]]
 == MySQL fields
 
--- a/filebeat/tests/system/test_multiline.py
+++ b/filebeat/tests/system/test_multiline.py
@@ -352,3 +352,36 @@ SetAdCodeMiddleware.default_ad_code route """
         self.wait_until(lambda: self.log_contains("missing required field accessing") == 1)
 
         proc.check_kill_and_wait(exit_code=1)
+
+    def test_multiline_lines(self):
+        """
+        Test if multiline line counter is added
+        """
+        self.render_config_template(
+            path=os.path.abspath(self.working_dir) + "/log/*",
+            multiline=True,
+            pattern="^\[",
+            negate="true",
+            match="after",
+            close_timeout="1s",
+        )
+
+        os.mkdir(self.working_dir + "/log/")
+
+        testfile = self.working_dir + "/log/test.log"
+
+        with open(testfile, 'w', 0) as file:
+            file.write("[2015] hello world\n")
+            file.write("  First Line\n")
+            file.write("  Second Line\n")
+
+        filebeat = self.start_beat()
+
+        self.wait_until(
+            lambda: self.output_has(lines=1),
+            max_timeout=10)
+
+        filebeat.check_kill_and_wait()
+
+        output = self.read_output_json()
+        output[0]["multiline"]["lines"] = 3


filebeat/harvester/reader/json.go
```diff
@@ -116,10 +116,12 @@ func transformNumbersArray(arr []interface{}) {
...
 func (r *JSON) Next() (Message, error) {
-	reader, err := r.reader.Next()
+	message, err := r.reader.Next()
 	if err != nil {
-		return reader, err
+		return message, err
 	}
-	reader.Content, reader.Fields = r.decodeJSON(reader.Content)
-	return reader, nil
+	var fields = common.MapStr{}
```
andrewkroh: I think this should be `var fields common.MapStr` since you don't need to allocate a `MapStr` because `decodeJSON` will be returning one.

filebeat/harvester/log.go
```diff
@@ -99,6 +99,13 @@ func (h *Harvester) Harvest() {
 			event.Text = &text
 			event.JSONFields = message.Fields
 			event.EventMetadata = h.config.EventMetadata
+			if message.Fields != nil {
```
andrewkroh: `if len(message.Fields) > 0` would omit an empty MapStr from the event that is being sent. `len()` also returns 0 if the map itself is nil.

andrewkroh: > Current issue with the changes in this PR is that fields is also sent if it is empty.

Does my comment above address that problem?

ruflin: Not 100% sure I can follow. The current events look as following if fields is empty:

```
{
  "@timestamp": "2016-08-17T07:25:53.294Z",
  "beat": {
    "hostname": "ruflin",
    "name": "ruflin"
  },
  "fields": {},
  "input_type": "log",
  "message": "This exceeds the buffer",
  "offset": 24,
  "source": "/Users/ruflin/Dev/gopath/src/github.com/elastic/beats/filebeat/build/system-tests/run/test_harvester.Test.test_exceed_buffer/log/test.log",
  "type": "log"
}
```
I would expect the publisher / output to remove "fields" if it len(fields) == 0. If this is not the case yet, my question is what is the best place to implement this as I don't think this is filebeat specific.

filebeat/harvester/reader/json.go
ruflin: Changed

filebeat/harvester/log.go
ruflin: The change I made now which leads to the expected result can be found here: https://github.com/elastic/beats/pull/2279/commits/b1a1bb03f2c7056f0eb0cf1c6adf8eb14a58f0d0 So far it seems it doesn't have any other side effects.

ruflin:
https://github.com/elastic/beats/pull/2290 should be reviewed and merged as it will require additional changes in this PR.

filebeat/harvester/log.go
urso: +1 for `len(message.Fields) > 0`. No need to run Fields.Update.

filebeat/harvester/log.go
```diff
@@ -99,6 +99,13 @@ func (h *Harvester) Harvest() {
...
+				if event.EventMetadata.Fields != nil {
+					event.EventMetadata.Fields.Update(message.Fields)
```
urso: is this overwriting EventMetadata.Fields or does it create a new `common.MapStr` internally? EventMetadata directly uses config.EventMetadata.

filebeat/harvester/reader/encode.go
```diff
@@ -33,5 +34,6 @@ func (p Encode) Next() (Message, error) {
 		Ts:      time.Now(),
 		Content: c,
 		Bytes:   sz,
+		Fields:  common.MapStr{},
```
urso: why not have `Fields: nil` here?

filebeat/harvester/reader/message.go
```diff
@@ -0,0 +1,22 @@
...
+type Message struct {
+	Ts      time.Time     // timestamp the content was read
+	Content []byte        // actual content read
+	Bytes   int           // total number of bytes read to generate the message
+	Fields  common.MapStr // optional fields that can be added by reader
+}
+
+func NewMessage() Message {
+	return Message{
+		Fields: common.MapStr{},
```
urso: what's the problem with `Fields` being nil by default?

Why not have a 

```
func (msg *Message) AddFields(fields common.MapStr) {
    if msg.Fields == nil {
       msg.Fields = common.Mapstr{}
    }
    msg.Fields.Update(fields)
}
```

any processor must use `AddFields` only, so some share `common.MapStr` will be updated by accident.

libbeat/common/mapstr.go
```diff
@@ -241,7 +241,7 @@ func (m MapStr) String() string {
 // An error is returned if underRoot is true and the value of ms.fields is not a
 // MapStr.
 func MergeFields(ms, fields MapStr, underRoot bool) error {
-	if ms == nil || fields == nil {
+	if ms == nil || fields == nil || len(fields) == 0 {
```
andrewkroh: This line can be just 

    if ms == nil || len(fields) == 0 {

because of how [`len`](https://golang.org/pkg/builtin/#len) works (it returns 0 when given a nil map).

filebeat/harvester/reader/multiline.go
```diff
@@ -215,12 +216,18 @@ func (mlr *Multiline) resetBuffer(m Message) {
...
+		mlr.fields = m.Fields
+	} else {
+		mlr.fields = common.MapStr{}
+	}
 }
 
 // finalize writes the existing content into the returned message and resets all reader variables.
 func (mlr *Multiline) finalize() Message {
 
+	mlr.fields["lines"] = mlr.numLines
```
urso: is numLines correctly updated? What if max lines has been reached? Predicate in `addLine` reads really weird.

filebeat/harvester/reader/multiline.go
```diff
@@ -215,12 +216,18 @@ func (mlr *Multiline) resetBuffer(m Message) {
 	mlr.addLine(m)
 	// Timestamp of first message is taken as overall timestamp
 	mlr.ts = m.Ts
-	mlr.fields = m.Fields
+	if m.Fields != nil {
+		mlr.fields = m.Fields
+	} else {
+		mlr.fields = common.MapStr{}
+	}
```
urso: with most fields in mlr being 'duplicates' of message, consider replacing those by 'Message' and provide a `func  mergeMessage(to, from *Message)`?

filebeat/harvester/reader/timeout.go
```diff
@@ -64,6 +64,6 @@ func (p *Timeout) Next() (Message, error) {
 		}
 		return msg.line, msg.err
 	case <-time.After(p.timeout):
-		return Message{}, p.signal
+		return NewMessage(), p.signal
```
urso: ouch, here we allocate a throwaway message + common.MapStr just to ignore it. Zero value of message should suffice here.

filebeat/harvester/util.go
```diff
@@ -12,11 +12,11 @@ import (
...
 // This could potentialy be improved / replaced by https://github.com/elastic/beats/libbeat/tree/master/common/streambuf
 func readLine(reader reader.Reader) (time.Time, string, int, common.MapStr, error) {
-	l, err := reader.Next()
+	message, err := reader.Next()
 
 	// Full line read to be returned
-	if l.Bytes != 0 && err == nil {
-		return l.Ts, string(l.Content), l.Bytes, l.Fields, err
+	if message.Bytes != 0 && err == nil {
+		return message.Ts, string(message.Content), message.Bytes, message.Fields, err
```
urso: maybe at some point we want readLine to return a value of Message?

ruflin: You mean `func readLine(reader reader.Reader) (Message, error)` ? Kind of like it.

libbeat/common/mapstr.go
ruflin: fixed

filebeat/harvester/reader/timeout.go
ruflin: My idea here that it is up to the caller what he does with the message, so the message should be properly initialised.

filebeat/harvester/reader/multiline.go
ruflin: Yeah, that should simplify the code. Thinking of doing this in an other PR first.

filebeat/harvester/reader/message.go
ruflin: That seems to be the better solution. We could even think about making Fields private.

filebeat/harvester/reader/encode.go
ruflin: The idea is that all reader return an initalised message, so Update fields just work. But the method AddFields below will solve this.

filebeat/harvester/log.go
ruflin: Fixed. @urso what do you mean be no need to run Fields.Update? Is this for the code below?

filebeat/harvester/log.go
ruflin: This will get more obvious when is merged: https://github.com/elastic/beats/pull/2290 Meta data addition and merging should not be a task of the harvester.

filebeat/harvester/reader/multiline.go
```diff
@@ -225,6 +227,8 @@ func (mlr *Multiline) clear() {
 // finalize writes the existing content into the returned message and resets all reader variables.
 func (mlr *Multiline) finalize() Message {
 
+	lines := common.MapStr{"lines": mlr.numLines}
+	mlr.message.AddFields(common.MapStr{"multiline": lines})
```
tsg: What will happen if someone defines `json` or `multiline` as fields in the config file, together with `fields_under_root`?

ruflin: Good point. Our current behaviour here is that is that fields will overwrite it. For json the behaviour does not change here. So in case someone has 
```
fields: 
  json: "Hello world"
```
already now, this would overwrite the fields from the json reader. The same applies `multiline`. So I think the behaviour didn't change even though I was never really thinking of the json overwriting by fields before. Seems like it wasn't an issue so far. The good thing is that we have it even mentioned in our docs: https://www.elastic.co/guide/en/beats/filebeat/1.3/configuration-filebeat-options.html#fields-under-root

monicasarbu:
@ruflin Can you please rebase this PR?

ruflin:
Closing this PR as no request for this feature has shown up recently.

PR #2280, opened 2016-08-16
clarifies use of beat.name

hilt86:
I wasted a few hours trying to set beat.name in the configuration file, all the time it being "name" under shipper.

### DIFF
--- a/filebeat/docs/fields.asciidoc
+++ b/filebeat/docs/fields.asciidoc
@@ -21,8 +21,12 @@ Contains common fields available in all event types.
 
 ==== beat.name
 
-The name of the Beat sending the log messages. If the shipper name is set in the configuration file, then that value is used. If it is not set, the hostname is used.
+The name of the Beat sending the log messages. If the shipper name is set in the configuration file, then that value is used. If it is not set, the hostname is used. This is set under shipper sub-heading, for example :
 
+....
+shipper:
+  name: "my-shipper"
+....
 
 ==== beat.hostname
 


ruflin:
@hilt86 Thanks for adding this to the docs. Could you sign the CLA?

@dedemorton As this changes in 5.0 and is potentially also not that obvious, it is perhaps worth to also document it?

hilt86:
I've signed it. Btw can you explain how this behaviour changes in 5.0? I'm writing an Ansible role for filebeat and would like to know how these changes will affect me please..

ruflin:
@hilt86 `shipper.name` will be replaced by just `name`. See: https://github.com/elastic/beats/blob/master/libbeat/_meta/config.yml#L6

hilt86:
@ruflin thank you! So "name:" be a root attribute rather than under "shipper:" - is that correct?

ruflin:
Exactly. That applies for the config. For the event it will still be under `beat.name`.

monicasarbu:
@hilt86 The asciidoc file is generated from the `fields.yml` file, so it's better to do the change in the `./libbeat/_meta/fields.yml`, and then run `make update` to generate the asciidoc. If you like, we can also take it from here, and update the `fields.yml` in a separate PR based on your input.
cc-ed @dedemorton

hilt86:
Yeah that might be easier if you take it from here so I don't have to clone
the repo and get make working on my workstation.

H

dedemorton:
This PR is replaced by #2398

PR #2283, opened 2016-08-16
Disable part of topology test on Solaris

### LINKED ISSUE #1285: Topology test failure Solaris
[`TestPublisherTypeUpdateTopologyPeriodically`](https://github.com/elastic/beats/blob/09ca0f4cf67a314366d2023525858c1b8e8efa65/libbeat/publisher/publish_test.go#L52-L52) fails on Solaris because [`common. LocalIpAddrsAsStrings`](https://github.com/elastic/beats/blob/c7566d2e6bb43da5780787392111429bf367fd6a/libbeat/common/net.go#L27-L27) always returns zero IPs. This is because `net.InterfaceAddrs()` is not implemented on Solaris; it's a [stub](https://golang.org/src/net/interface_stub.go).

I assume this means topology is broken for Solaris. 

What should we do? At a minimum I would like to fix the test case.

andrewkroh:
Go cannot return the local interfaces because the method is not implemented on Solaris. The local interfaces are used in determining the topology.

Fixes #1285

### DIFF
--- a/libbeat/publisher/publish_test.go
+++ b/libbeat/publisher/publish_test.go
@@ -3,6 +3,7 @@
 package publisher
 
 import (
+	"runtime"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


libbeat/publisher/publish_test.go
```diff
@@ -70,5 +71,10 @@ func TestPublisherTypeUpdateTopologyPeriodically(t *testing.T) {
 
 	// Validate that PublishTopology was invoked.
 	assert.Equal(t, shipperName, <-topo.publishName)
-	assert.True(t, len(<-topo.publishLocalAddrs) > 0)
+	switch runtime.GOOS {
+	default:
```
ruflin: minor detail: I'm somehow used to have default at the end.

PR #2284, opened 2016-08-16
configurable kafka partitioner 

urso:
- Store kafka message wrapper in outputs.Data.Values for re-use on retry.
- Add support for configuring the partitioner strategy
- configurable support ignoring unavailable partitions
- partitioning is mostly stable in error case (try to schedule events to same partition if returned to libbeat on error)


urso:
Requires #2295 to fix test build

urso:
Will update docs in another PR, as a number of other settings have been added/changed for kafka output.

libbeat/outputs/kafka/kafka_integration_test.go
```diff
@@ -23,7 +24,198 @@ const (
 	kafkaDefaultPort = "9092"
 )
 
-var testOptions = outputs.Options{}
+func TestKafkaPublish(t *testing.T) {
+	single := modetest.SingleEvent
+	// multi := modetest.MultiEvent
```
ruflin: What is the idea with these 2 commented out lines?

libbeat/outputs/kafka/partition.go
```diff
@@ -0,0 +1,297 @@
...
+}
+
+var partitioners = map[string]partitionBuilder{
+	"random":      cfgRandomPartitioner,
+	"round_robin": cfgRoundRobinPartitioner,
+	"hash":        cfgHashPartitioner,
+}
+
+func initPartitionStrategy(
+	partition map[string]*common.Config,
```
ruflin: Do we need a map here as 0 or 1 partition are only allowed?

libbeat/outputs/kafka/kafka_integration_test.go
urso: they can be removed as not being used. events of type []modetest.EventInfo hold a number of events + setting to bulk or not bulk publish those events. It's used to drive sending events through output plugin.

In test randMulti is used.

libbeat/outputs/kafka/partition.go
```diff
@@ -0,0 +1,297 @@
...
+			msg.hash = hasher.Sum32()
+			hash = msg.hash
+		}
+
+		// create positive hash value
+		return hash2Partition(hash, numPartitions)
+	}
+}
+
+func makeFieldsHashPartitioner(fields []string, dropFail bool) partitioner {
```
ruflin: Could this have an affect on the output performance?

libbeat/outputs/kafka/partition.go
urso: partitioners are configured by name. the name in map[string] is for lookup into `partitioners`. We use this pattern in a few places with `plugins` being given by name. Alternate solution: capture *common.Config, get list of keys and check it's length is 1, extract Child config by key.

libbeat/outputs/kafka/partition.go
urso: Yes. No docs yet.

ruflin:
LGTM. What about the updates to the config file? Will this also follow in a separate PR?

urso:
config files can be done in this PR.

libbeat/outputs/kafka/kafka_integration_test.go
urso: removed

urso:
@ruflin added configs

packetbeat/packetbeat.yml
```diff
@@ -35,6 +35,10 @@ packetbeat.protocols.amqp:
   # the AMQP protocol by commenting out the list of ports.
   ports: [5672]
 
+packetbeat.protocols.cassandra:
+  #Cassandra port for traffic monitoring.
```
tsg: It's not part of this PR, but needs a space in front of Cassandra.

libbeat/_meta/config.full.yml
```diff
@@ -261,6 +261,23 @@ output.elasticsearch:
...
+  #partition.hash:
+    # If enabled, events will only be published to partitions with reachable
+    # leaders. Default is false.
+    #reachable_only: false
+
+    # Configure alternative event field names used to compute the hash value.
+    # If empty `output.kafka.key` setting will be used.
+    # Default value is empty list.
+    #hash: []
+
```
tsg: What about the `switch` parameter? It's not clear to me what it does and was hoping to find an explanation here :)

tsg:
LGTM, left a minor comment/question.

andrewkroh:
LGTM

packetbeat/packetbeat.yml
urso: oh, this should be fixed by another PR. will have to rebase.

libbeat/_meta/config.full.yml
urso: Here I only document the default partitioner. `partition.hash` is default strategy in libbeat and sarama. The switch parameter used by `partition.random` and `partition.round_robin` is some kind of 'bulking' parameter. That is, the 2 partitioners do select another partition only every N events. This parameter does not exist for the hashing partitioner.

urso: The other parameters will be documented in another PR updating docs about all kafka output settings.

urso:
renamed `switch` to `group_events`.

PR #2288, opened 2016-08-16
Edits to content about generating a beat

dedemorton:
First pass edit of content changes made for alpha. I have quite a few questions added as comments.

@monicasarbu Can you respond to the general questions?
@ruflin I think Monica said that the naming convents are based on something you wrote. If so, can you answer my questions in that section? Adding some concrete examples might be enough to clarify what you mean there.

### DIFF
--- a/libbeat/docs/event-conventions.asciidoc
+++ b/libbeat/docs/event-conventions.asciidoc
@@ -38,9 +38,9 @@ workers.idle
 
 
 [[abbreviations]]
-==== Standard Abbreviations
+==== Standardised Names
 
-Here is a list of standardised words and units that are used across all Beats:
+Here is a list of standardised names and units that are used across all Beats:
 
 [options="header"]
 |=======================
--- a/libbeat/docs/event-conventions.asciidoc
+++ b/libbeat/docs/event-conventions.asciidoc
@@ -1,37 +1,47 @@
 [[event-conventions]]
-=== Event conventions
+=== Naming Conventions
 
-==== Naming conventions
+When creating events, use the following conventions for field names and abbreviations.
 
-For the field names, the Beats follow the naming conventions below:
+[[field-names]]
+==== Field Names
 
-- All fields must be lower case
-- Snake case for combining words
-- Group related fields in sub documents, which means using the . notation. Groups are mostly described by common prefixes.
-- Prevent namespace duplication. If connections appears in the namespace, it's not needed in the sub document
-- Do not use complex abbreviations. A list of standardised abbreviations can be found below.
-- Organise the documents from the general to the details, which allows namespacing. The type should always be last, like .pct.
-- If two fields are the same but with different units, remove the less granular one
+Use the following naming conventions for field names:
+
+- All fields must be lower case.
+- Use snake case (underscores) for combining words.
+- Group related fields into subdocuments, which means using the dot (.) notation. Groups are mostly described by common prefixes. 
+//QUESTION: I am not sure what you mean by common prefixes here. Perhaps examples would help?
+- Prevent namespace duplication. If connections appears in the namespace, it's not needed in the sub document 
+//QUESTION: I don't understand what you mean here. Can you clarify or give an example?
+- Do not use complex abbreviations. See <<abbreviations>>.
+//QUESTION: What do we mean by complex abbreviations? Can we just tell people to use standard abbreviations and avoid using abbreviations that aren't commonly known?
+- Organise the documents from general to specific to allow for namespacing. The type, such as `.pct`, should always be last. For example, `system.core.user.pct`
+- If two fields are the same, but with different units, remove the less granular one.
+//QUESTION: What if we want to  provide a less granular view, too? It would be useful to provide an example here.
 - In case the value correlates with the name of a nested document, use value inside the document
-- Do not use . in the names
-- Use singular and plural properly for the fields. Example: sec_per_request vs open_requests
-- Use singular names for metricsets. It easier to read the event created: system.process.load = 0.3
+//QUESTON: I don't understand what you mean here. Can you clarify?
+- Do not use dots (.) in names. 
+//QUESTION: Wouldn't it be clearer to say, "Do not use dots (.) in individual field names. Dots are reserved for grouping related fields into subdocuments." Is that why you have this restriction?
+- Use singular and plural names properly to reflect the field content. For example: `sec_per_request` vs `open_requests`. 
+//QUESTION: I wonder if we should remove this rule since we sort of violate it ourselves (especially with our  abbreviations).
+- Use singular names for metricsets. It's easier to read the event created as: `system.process.load = 0.3`. The goal is to have a similar experience across all metrics. 
+//QUESTION: This seems really specific to metricbeat. Does it really belong in the general dev guide? 
 
+[[abbreviations]]
+==== Standard Abbrevations
 
-The goal is to have a similar experience across all metrics.
+Here is a list of standardised words and units that are used across all Beats:
 
-
-==== Abbrevations
-
-List of standardised words and units across all Beats:
+//QUESTION: Are we saying here to use the word/abbreviation on the left in favor of the one on the right? If so, we need to make this more clear. Maybe we could have two columns Use... Instead of...
 
 * avg: average
-* connection: conn
+* connection: conn 
 * count:
 * day: days, d
-* max: maximumg
+* max: maximum
 * min: minimum
-* pct: precentage
+* pct: percentage
 * request: req
 * sec: seconds, second, s
 * ms: millisecond, millis
--- a/libbeat/docs/event-conventions.asciidoc
+++ b/libbeat/docs/event-conventions.asciidoc
@@ -10,44 +10,54 @@ Use the following naming conventions for field names:
 
 - All fields must be lower case.
 - Use snake case (underscores) for combining words.
-- Group related fields into subdocuments, which means using the dot (.) notation. Groups are mostly described by common prefixes. 
-//QUESTION: I am not sure what you mean by common prefixes here. Perhaps examples would help?
-- Prevent namespace duplication. If connections appears in the namespace, it's not needed in the sub document 
-//QUESTION: I don't understand what you mean here. Can you clarify or give an example?
-- Do not use complex abbreviations. See <<abbreviations>>.
-//QUESTION: What do we mean by complex abbreviations? Can we just tell people to use standard abbreviations and avoid using abbreviations that aren't commonly known?
-- Organise the documents from general to specific to allow for namespacing. The type, such as `.pct`, should always be last. For example, `system.core.user.pct`
-- If two fields are the same, but with different units, remove the less granular one.
-//QUESTION: What if we want to  provide a less granular view, too? It would be useful to provide an example here.
-- In case the value correlates with the name of a nested document, use value inside the document
-//QUESTON: I don't understand what you mean here. Can you clarify?
-- Do not use dots (.) in names. 
-//QUESTION: Wouldn't it be clearer to say, "Do not use dots (.) in individual field names. Dots are reserved for grouping related fields into subdocuments." Is that why you have this restriction?
-- Use singular and plural names properly to reflect the field content. For example: `sec_per_request` vs `open_requests`. 
-//QUESTION: I wonder if we should remove this rule since we sort of violate it ourselves (especially with our  abbreviations).
-- Use singular names for metricsets. It's easier to read the event created as: `system.process.load = 0.3`. The goal is to have a similar experience across all metrics. 
-//QUESTION: This seems really specific to metricbeat. Does it really belong in the general dev guide? 
+- Group related fields into subdocuments by using dot (.) notation. Groups typically have common prefixes. For example, if you have fields called `CPULoad` and `CPUSystem` in a service, you would convert
+them into `cpu.load` and `cpu.system` in the event. 
+- Avoid repeating the namespace in field names. If a word or abbreviation appears in the namespace, it's not needed in the field name. For example, instead of `cpu.cpu_load`, use `cpu.load`.
+- Use <<abbreviations,standard abbreviations>> and avoid using abbreviations that aren't commonly known.
+- Organise the documents from general to specific to allow for namespacing. The type, such as `.pct`, should always be last. For example, `system.core.user.pct`.
+- If two fields are the same, but with different units, remove the less granular one. For example, include `timeout.sec`, but don't include `timeout.min`. If a less granular value is required, you can calculate it later.
+- If a field name matches the namespace used for nested fields, add `.value` to the field name. For example, instead of:
++
+[source,yaml]
+----------
+workers
+workers.busy
+workers.idle
+----------
++
+Use:
++
+[source,yaml]
+----------
+workers.value
+workers.busy
+workers.idle
+----------
+- Do not use dots (.) in individual field names. Dots are reserved for grouping related fields into subdocuments. 
+- Use singular and plural names properly to reflect the field content. For example, use `requests_per_sec` rather than `request_per_sec`. 
+
 
 [[abbreviations]]
-==== Standard Abbrevations
+==== Standard Abbreviations
 
 Here is a list of standardised words and units that are used across all Beats:
 
-//QUESTION: Are we saying here to use the word/abbreviation on the left in favor of the one on the right? If so, we need to make this more clear. Maybe we could have two columns Use... Instead of...
-
-* avg: average
-* connection: conn 
-* count:
-* day: days, d
-* max: maximum
-* min: minimum
-* pct: percentage
-* request: req
-* sec: seconds, second, s
-* ms: millisecond, millis
-* mb: megabytes
-* msg: message
-* ns: nanoseconds
-* norm: normalized
-* us: microseconds
-
+[options="header"]
+|=======================
+|Use...     |Instead of... 
+|avg        |average
+|connection |conn 
+|count      | 
+|day        |days, d
+|max        |maximum
+|min        |minimum
+|pct        |percentage
+|request    |req
+|sec        |seconds, second, s
+|ms         |millisecond, millis
+|mb         |megabytes
+|msg        |message
+|ns         |nanoseconds
+|norm       |normalized
+|us         |microseconds
+|=======================
--- a/libbeat/docs/newbeat.asciidoc
+++ b/libbeat/docs/newbeat.asciidoc
@@ -10,6 +10,11 @@ handling, for logging, and more. By using this common framework, you can ensure
 that all Beats behave consistently and that they are easy to package and run
 with common tools.
 
+In this guide, you learn how to use the Beat generator to create source code for
+an example Beat called Testbeat. The Beat generator creates all the files required
+for a working Beat. To create your own Beat, you modify the generated
+files and implement the custom logic needed to collect the data you want to ship.
+
 [[newbeat-getting-ready]]
 === Getting Ready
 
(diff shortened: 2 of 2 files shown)


libbeat/docs/event-conventions.asciidoc
```diff
@@ -1,37 +1,47 @@
...
+
+[[field-names]]
+==== Field Names
+
+Use the following naming conventions for field names:
+
+- All fields must be lower case.
+- Use snake case (underscores) for combining words.
+- Group related fields into subdocuments, which means using the dot (.) notation. Groups are mostly described by common prefixes. 
+//QUESTION: I am not sure what you mean by common prefixes here. Perhaps examples would help?
```
ruflin: Example: The fields in the service are called:

```
CPULoad
CPUSystem
```

They share the prefix CPU. This is converted to 

```
cpu.load
cpu.system
```

libbeat/docs/event-conventions.asciidoc
```diff
@@ -1,37 +1,47 @@
...
+- Prevent namespace duplication. If connections appears in the namespace, it's not needed in the sub document 
+//QUESTION: I don't understand what you mean here. Can you clarify or give an example?
```
ruflin: Same example as above. Instead of going with
```
cpu.cpu_load
cpu.cpu_system
```

The cpu part is removed as it already shows up in the namespace

libbeat/docs/event-conventions.asciidoc
```diff
@@ -1,37 +1,47 @@
...
+- Do not use complex abbreviations. See <<abbreviations>>.
+//QUESTION: What do we mean by complex abbreviations? Can we just tell people to use standard abbreviations and avoid using abbreviations that aren't commonly known?
```
ruflin: Sounds like we mean the same, just different wording :-)

libbeat/docs/event-conventions.asciidoc
```diff
@@ -1,37 +1,47 @@
...
+- Organise the documents from general to specific to allow for namespacing. The type, such as `.pct`, should always be last. For example, `system.core.user.pct`
+- If two fields are the same, but with different units, remove the less granular one.
+//QUESTION: What if we want to  provide a less granular view, too? It would be useful to provide an example here.
```
ruflin: The calculation to the less granular one can happen on the Kibana side. Example:

```
timeout.sec
timeout.min
```

We only send `timeout.sec`

libbeat/docs/event-conventions.asciidoc
```diff
@@ -1,37 +1,47 @@
...
 - In case the value correlates with the name of a nested document, use value inside the document
-- Do not use . in the names
-- Use singular and plural properly for the fields. Example: sec_per_request vs open_requests
-- Use singular names for metricsets. It easier to read the event created: system.process.load = 0.3
-
-
-The goal is to have a similar experience across all metrics.
+//QUESTON: I don't understand what you mean here. Can you clarify?
```
ruflin: Example. Assuming we have the following three values:
```
workers
workers_busy
workers_idle
```

We can't convert this to:
```
workers
workers.busy
workers.idle
```
The problem is here, that workers is a value instead of a dict. So instead we convert it to:
```
workers.value
workers.busy
workers.idle
```

libbeat/docs/event-conventions.asciidoc
```diff
@@ -1,37 +1,47 @@
...
+- Do not use dots (.) in names. 
+//QUESTION: Wouldn't it be clearer to say, "Do not use dots (.) in individual field names. Dots are reserved for grouping related fields into subdocuments." Is that why you have this restriction?
```
ruflin: Yes

libbeat/docs/event-conventions.asciidoc
```diff
@@ -1,37 +1,47 @@
...
+- Use singular and plural names properly to reflect the field content. For example: `sec_per_request` vs `open_requests`. 
+//QUESTION: I wonder if we should remove this rule since we sort of violate it ourselves (especially with our  abbreviations).
```
ruflin: I think the rule is important and we should keep it. Can't abbreviation be expanded to singular and plural? O:-) Where do we violate it except for the abbreviations?

libbeat/docs/event-conventions.asciidoc
```diff
@@ -1,37 +1,47 @@
...
+- Use singular names for metricsets. It's easier to read the event created as: `system.process.load = 0.3`. The goal is to have a similar experience across all metrics. 
+//QUESTION: This seems really specific to metricbeat. Does it really belong in the general dev guide?
```
ruflin: All the above rules are coming from Metricbeat. It was just nice that they also fitted for all beats. But I agree, the above is a specific naming convention for Metricsets. Probably this should be put into the dev guide for creating metricsets.

libbeat/docs/event-conventions.asciidoc
```diff
@@ -1,37 +1,47 @@
...
+//QUESTION: This seems really specific to metricbeat. Does it really belong in the general dev guide? 
 
+[[abbreviations]]
+==== Standard Abbrevations
 
-==== Abbrevations
+Here is a list of standardised words and units that are used across all Beats:
 
-List of standardised words and units across all Beats:
+//QUESTION: Are we saying here to use the word/abbreviation on the left in favor of the one on the right? If so, we need to make this more clear. Maybe we could have two columns Use... Instead of...
```
ruflin: Answer: yes. Agree, table would be nice here.

libbeat/docs/newbeat.asciidoc
```diff
@@ -19,21 +20,32 @@ language, and very few people are experts in it. In fact, several
...
 [source,shell]
 ----------------------------------------------------------------------
 go get github.com/elastic/beats
 ----------------------------------------------------------------------
 
+When you run the command, all source files will be downloaded to the
+`$GOPATH/src` path. You can ignore the "no buildable Go source files" message because
+you will build the source later.
+
+//QUESTION: Wondering if adding sentence about "no buildable Go source files" is the best way to handle this. The problem is that users might be confused because the command returns the message "package github.com/elastic/beats: no buildable Go source files in /Users/dedemorton/workspace/src/github.com/elastic/beats".
```
ruflin: What could be an alternative here?

libbeat/docs/newbeat.asciidoc
```diff
@@ -19,21 +20,32 @@ language, and very few people are experts in it. In fact, several
...
+you will build the source later.
+
+//QUESTION: Wondering if adding sentence about "no buildable Go source files" is the best way to handle this. The problem is that users might be confused because the command returns the message "package github.com/elastic/beats: no buildable Go source files in /Users/dedemorton/workspace/src/github.com/elastic/beats". 
+
 In this guide, we use working examples from the Testbeat source code in
 https://github.com/monicasarbu/testbeat[GitHub]
 to demonstrate how to implement a Beat. Just copy
 the source files and modify them as necessary for your Beat.
 
+//QUESTION: Do we really want users to copy testbeat if they are using the beat generator? Don’t we want them to use the beat generator instead? If we do want them to use testbeat, then shouldn't we tell them to clone it since it's not a part of github.com/elastic/beats?
```
ruflin: Agree, that is kind of strange. @monicasarbu It seems like testbeat is exactly what the generator creates. Why do we need testbeat here?

libbeat/docs/newbeat.asciidoc
```diff
@@ -128,10 +146,11 @@ beat_path [github.com/{github id}]:
 full_name [Firstname Lastname]: {Full Name}
 ---------
 
-This should now have created a directory `testbeat` inside our folder with several files. 
+You should see a directory called `testbeat` inside of your project folder. 
 
-We now have a raw template of the Beat, but still need to fetch the dependencies and setup the git repository.
+You now have a raw template of the Beat, but you still need to fetch the dependencies and set up the git repository.
 
+//REVIEWERS: We don't describe how to fetch dependencies or set up the git repo anywhere in the docs. Have we decided to skip talking about the make command in the docs because we want to remove  dependencies like python and the virtualenv? I'm not entirely opposed to pointing off to Jongmin's blog for details about running make and setting up the git repo, but if we do that, we need to be more explicit and we need to agree to update the blog entry whenever it gets out-of-date.
```
ruflin: I think we should have it also in the dev guide. Easier to keep it up-to-date.

ruflin:
@dedemorton I responded to the questions. Let me know if some more clarifications are needed or when I should have a second look.

libbeat/docs/event-conventions.asciidoc
monicasarbu: I am not sure if we should have this rule here, as it's common sense. I would keep the list of rules as small as possible, so we don't loose the user when reading the requirements.

libbeat/docs/event-conventions.asciidoc
monicasarbu: It was initially part of the Metricbeat, but I moved it to the Beats platform dev guide. I think the rules here can be applied for all Beats, including the community Beats, not only for MetricSets.

libbeat/docs/newbeat.asciidoc
monicasarbu: I think it would be nice if we could avoid advising the user to ignore the error. @tsg what do you think?

libbeat/docs/newbeat.asciidoc
monicasarbu: You are right, Dede. Testbeat is what the Beat generator creates, and I am using it as an example. What I wanted here to say is that after the users generate the example Beat, they can adjust some source files to create their own Beat.

libbeat/docs/newbeat.asciidoc
monicasarbu: Yes, I didn't include it as it was part of the blog post, but we can get the content from the blog post and adjust it a bit to fit in the dev guide.

libbeat/docs/newbeat.asciidoc
```diff
@@ -393,11 +412,16 @@ func main() {
...
+Lsbeat periodically ships them to Elasticsearch for storage. 
+
+To help you learn how to build a Beat, we've created this http://elastic.co//blog/build-your-own-beat[blog post]
+that describes how to build Lsbeat from scratch. The blog post includes additional details about fetching dependencies
+and compiling the code. 
 
-github.com/kimjmin/lsbeat[Lsbeat] is similar to the `ls` command line tool, but instead of printing the files and subdirectories to the screen,
-Lsbeat periodically ships them to Elasticsearch for storage.
-The http://elastic.co//blog/build-your-own-beat[blog post] describes how to build from scratch `Lsbeat`.
+//REVIEWERS: I wanted to add a little clarification here because it's initially jarring to come to this section in the doc because it talks about Lsbeat instead of the example that's used in the topics leading up to this one. I hope that flagging this section as an example makes the purpose less ambiguous.
```
monicasarbu: @dedemorton My idea was to include details about how to generate a Beat in the dev guide, and use the blog post for a concrete example (Lsbeat). That's why at the end of the dev guide, I am introducing Lsbeat to users to see how they can build a real example of a Beat. Feel free to change the content to make it more clear.

libbeat/docs/event-conventions.asciidoc
ruflin: All these rules were created when I went through all the existing metricsets and tried to unify them. The above "rule" was introduced as some services use `request_per_sec` others `requests_per_sec` and a decision had to made which one to use. I don't think that common sense will apply here as the natural thing to me initially felt like keep what the service provides, but that leaded to inconsistency.

dedemorton:
@ruflin I've updated the topic about naming conventions. Please review. I've tried to simplify the descriptions to make them easier to scan. The examples help a lot. Here I'm not sure if I should say "nested fields" or "nested documents" or "subdocuments": "If a field name matches the namespace used for nested fields."

ruflin:
@dedemorton I would go with fields, as I see the whole thing as one document.

libbeat/docs/newbeat.asciidoc
```diff
@@ -92,32 +99,33 @@ The following example shows an event object in Lsbeat:
...
 
-To generate your own beat we use the `beat-generator`. First you must install
-https://github.com/audreyr/cookiecutter[cookiecutter]. Check out the
-http://cookiecutter.readthedocs.io/en/latest/installation.html[installation
-guide]. After having installed cookiecutter, we must decide on a name for our beat. 
-The name must be one word all lowercase. In our example is Testbeat.
+Before running the Beat generator, you must install https://github.com/audreyr/cookiecutter[cookiecutter], a
+command-line utility that creates projects from project templates. Check out the 
+http://cookiecutter.readthedocs.io/en/latest/installation.html[installation guide]. After installing cookiecutter,
+decide on a name for your beat. The name must be one word in all lowercase. In our example, we use `testbeat`.
```
monicasarbu: @dedemorton Oh. My mistake here. I updated only the example, but not the text. The way we write the Beat name is starting with a capital later, and then the rest lowercase, and keep all lowercases for libbeat. I am thinking that it would be nice to advice our users to do the same.

libbeat/docs/newbeat.asciidoc
```diff
@@ -92,32 +99,33 @@ The following example shows an event object in Lsbeat:
...
 
 [source,shell]
 --------------------
-$ cd $GOPATH/src/github.com/{user}
-$ cookiecutter $GOPATH/src/github.com/elastic/beats/generate/beat
+cookiecutter $GOPATH/src/github.com/elastic/beats/generate/beat
 --------------------
 
-Cookiecutter will ask you several questions. For your project_name enter testbeat, for github_user - your github id. The next two question with for beat and beat_path should already be automatically set correct. For the last one your can insert your Firstname Lastname.
+Cookiecutter will prompt you to enter information about your Beat. For the `project_name`, enter `testbeat`.
```
monicasarbu: Testbeat

monicasarbu:
@dedemorton I left only a small comment. LGTM

dedemorton:
@ruflin I've pushed changes for Monica's feedback. Feel free to squash/merge if you review this and it looks good.

libbeat/docs/event-conventions.asciidoc
```diff
@@ -1,43 +1,63 @@
...
+workers.value
+workers.busy
+workers.idle
+----------
+- Do not use dots (.) in individual field names. Dots are reserved for grouping related fields into subdocuments. 
+- Use singular and plural names properly to reflect the field content. For example, use `requests_per_sec` rather than `request_per_sec`. 
+
+
+[[abbreviations]]
+==== Standard Abbreviations
```
ruflin: I just realised perhaps we should not call it abbreviations, as often we use the full form like `request` instead of `req`. Perhaps `Naming Standards`?

libbeat/docs/newbeat.asciidoc
```diff
@@ -19,21 +25,22 @@ language, and very few people are experts in it. In fact, several
...
 go get github.com/elastic/beats
 ----------------------------------------------------------------------
 
-In this guide, we use working examples from the Testbeat source code in
-https://github.com/monicasarbu/testbeat[GitHub]
-to demonstrate how to implement a Beat. Just copy
-the source files and modify them as necessary for your Beat.
-
+When you run the command, all source files are downloaded to the
+`$GOPATH/src` path. You can ignore the "no buildable Go source files" message because
```
ruflin: The path is `$GOPATH/src/github.com/elastic/beats`

libbeat/docs/newbeat.asciidoc
```diff
@@ -92,49 +99,92 @@ The following example shows an event object in Lsbeat:
...
-To generate your own beat we use the `beat-generator`. First you must install
-https://github.com/audreyr/cookiecutter[cookiecutter]. Check out the
-http://cookiecutter.readthedocs.io/en/latest/installation.html[installation
-guide]. After having installed cookiecutter, we must decide on a name for our beat. 
-The name must be one word all lowercase. In our example is Testbeat.
+Before running the Beat generator, you must install https://github.com/audreyr/cookiecutter[cookiecutter], a
+command-line utility that creates projects from project templates. Check out the 
+http://cookiecutter.readthedocs.io/en/latest/installation.html[installation guide]. After installing cookiecutter,
+decide on a name for your beat. The name should be one word with the first letter capitalized. In our example,
+we use `Testbeat`.
```
ruflin: Perhaps a better name in the future would be Countbeat. The reasons is that the default beat which is created has a counter inside.

libbeat/docs/newbeat.asciidoc
```diff
@@ -188,50 +238,28 @@ func (bt *Testbeat) Stop() {
...
-- `done`: Channel used by the `Run()` function to stop when the `Stop()` function is called.
-- `config`: Configuration options for the Beat
-- `client`: Publisher that takes care of sending the events to the
+<1> `done`: Channel used by the `Run()` function to stop when the `Stop()` function is called.
+<2> `config`: Configuration options for the Beat
+<3> `client`: Publisher that takes care of sending the events to the
   defined output.
 
-The `Beat` parameter received by the `Run` method contains data about the
+The `Beat` parameter received by the `Run` function contains data about the
```
ruflin: it is actually a method and not a function.

libbeat/docs/newbeat.asciidoc
```diff
@@ -279,30 +315,57 @@ type Config struct {
 var DefaultConfig = Config{
 	Period: 1 * time.Second,
 }
+
```
ruflin: This empty line is probably not needed

libbeat/docs/newbeat.asciidoc
```diff
@@ -279,30 +315,57 @@ type Config struct {
...
+    Path   string        `config:"path"`
+}
 
-	ls := &Testbeat{
-		done:   make(chan struct{}),
-		config: config,
-	}
-	return ls, nil
+var DefaultConfig = Config{
+	Period: 1 * time.Second,
```
ruflin: Looks like indentation is off here

ruflin:
@dedemorton I left a few minor additional comments.

libbeat/docs/newbeat.asciidoc
dedemorton: @monicasarbu Are you OK with me changing Testbeat to Countbeat? I agree with this change because it would be better for us to provide an example that's in the spirit of how we want Beats to be named.

libbeat/docs/newbeat.asciidoc
dedemorton: @ruflin Just an FYI that I added the empty line to the code example to fix the preview in GitHub. Before I added the extra line, the preview was messed up. Instead of seeing the formatted doc, you saw the asciidoc markup after this line. The extra line looks OK in the built doc, so I'll just leave it there for now.

libbeat/docs/event-conventions.asciidoc
dedemorton: I'll use Standardised Names because Naming Standards is too much like the little of the topic (Naming Conventions).

libbeat/docs/newbeat.asciidoc
dedemorton: OK, so I'll revert the change and fix the earlier mention where we call it a function. :-)

libbeat/docs/newbeat.asciidoc
monicasarbu: Sure. Countbeat sounds better. Thanks!

libbeat/docs/newbeat.asciidoc
monicasarbu: It looks like I used spaces, instead of tabs.

dedemorton:
I've made updates to resolve the review comments. This should be ready to squash + merge.

PR #2290, opened 2016-08-17
Add prospector meta data on prospector level to event

ruflin:
This simplifies the creation of events on the harvester level and automatically provides the functionality for future types of harvesters

### DIFF
--- a/filebeat/harvester/config.go
+++ b/filebeat/harvester/config.go
@@ -6,17 +6,15 @@ import (
 
 	cfg "github.com/elastic/beats/filebeat/config"
 	"github.com/elastic/beats/filebeat/harvester/reader"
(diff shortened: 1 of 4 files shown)


urso:
This change seems to change the event to be only partially by harvester.

I see why it makes sense to fill these common setting in prospector, but if possible i'd avoid having to pass partially initialized values.

ruflin:
I think the way it will look in the long run is that there is generic event object which is valid across all prospectors, and each prospector with its harvester will have potential additional fields on top. Somehow reminds me of: https://github.com/elastic/beats/pull/2239 I think the above event is completely valid also without the additional fields from the prospector. As it simplifies the harvester this is a step forward.

ruflin:
jenkins, retest it

PR #2291, opened 2016-08-17
Add load test for open files

ruflin:
This test creates x files to be read by filebeat on startup. The close_* config options were disabled to keep the file handlers open to hit some potential limits.

`make load-tests` was introduced as load tests are not run as part of the full test suite as they could be slow and long running.

### DIFF
--- a/filebeat/tests/system/test_load.py
+++ b/filebeat/tests/system/test_load.py
@@ -3,12 +3,15 @@ import os
 import logging
 import logging.handlers
 import json
+import unittest
 from nose.plugins.skip import Skip, SkipTest
 from nose.plugins.attrib import attr
 
 """
 Test filebeat under different load scenarios
 """
+
+LOAD_TESTS = os.environ.get('LOAD_TESTS', False)
 
 
 class Test(BaseTest):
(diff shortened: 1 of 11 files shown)


ruflin:
I did some tests with 20000 files. It seems when filebeat hits the max number of open files it waits until some of the open harvesters get closed and then continues reading. No error messages are reported.

ruflin:
As these tests seem to fail on the travis mac build, it probably makes sense to exclude these tests from our default builds and only run them manually or on certain machines. An idea is to use tags here: http://nose.readthedocs.io/en/latest/plugins/attrib.html One issue with these tags is that in case no tags are added, all tests are run. We have the opposite in the Golang tests. If you just run `go test ...` only unit tests are run and no integration tests which is kind of nice.

tsg:
Yeah, it would be nice to still have `nosetest` run only the fast tests. An idea: http://stackoverflow.com/a/30310811/15685

ruflin:
@tsg Nice, skipUnless is what I was looking for. I will also adjust it accordingly for the integration tests.

ruflin:
@tsg I adjusted the test handling accordingly. By default now all load and integration tests are skipped. Both environment variable and and the `-a load` flag are used to make it possible to only run `load` tests if needed.

PR #2297, opened 2016-08-17
Filebeat metricset in beats module

ruflin:
* Add Filebeat as a metricset to use metricbeat to monitor filbeat
* Add basic tests
* Mark metricset as experimental

More libbeat metrics should be added in the future.


metricbeat/docker-compose.yml
```diff
@@ -21,6 +24,8 @@ beat:
     - MYSQL_HOST=mysql
     - MYSQL_PORT=3306
     - POSTGRESQL_DSN=postgres://postgres@postgresql:5432?sslmode=disable
+    - POSTGRESQL_HOST=postgresql
+    - POSTGRESQL_PORT=5432
```
tsg: Just curious, why is this needed?

metricbeat/module/beats/filebeat/filebeat.go
```diff
@@ -0,0 +1,85 @@
...
+	if resp.StatusCode != 200 {
+		return nil, fmt.Errorf("HTTP error %d: %s", resp.StatusCode, resp.Status)
+	}
+
+	buf, _ := ioutil.ReadAll(resp.Body)
+	data := Filebeat{}
+	json.Unmarshal(buf, &data)
+
+	event := common.MapStr{
+		"harvesters": common.MapStr{
```
tsg: I think you could have also done this by unmarshaling into a `map[string]interface{}` and then use the `schema.mapstriface` tools to do the mapping. The advantage is that in case one of the fields is not present, it will not be included in the output, instead of being set to 0.

metricbeat/docker-compose.yml
ruflin: For the docker entrypoint script below. Otherwise the beat will not wait for postgres to properly start.

metricbeat/module/beats/filebeat/filebeat.go
ruflin: I will adjust it accordingly. That's much nicer.

ruflin:
@tsg Redone with the schema.

tsg:
Nice, LGTM.

PR #2298, opened 2016-08-17
Create a beats-dashboards package

tsg:
Contains the dashboards for each beat in a sub-directory, ready
to be used by the import Go program.

Implementation wise, this is done as a pseudo-platform so it can
use features like the common version and the build hash.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,8 +64,8 @@ simplify:
 # Collects all dashboards and generates dashboard folder for https://github.com/elastic/beats-dashboards/tree/master/dashboards
 .PHONY: beats-dashboards
 beats-dashboards:
-	mkdir -p build
-	$(foreach var,$(PROJECTS),cp -r $(var)/etc/kibana/ build/dashboards  || exit 1;)
+	mkdir -p build/dashboards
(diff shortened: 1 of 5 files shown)


tsg:
jenkins, package it

monicasarbu:
It needs to be merged with master as I did myself some changes to Makefile.

Makefile
```diff
@@ -64,22 +64,27 @@ simplify:
...
 	sh libbeat/scripts/build_docs.sh ${PROJECTS}
 
 .PHONY: package
-package:
+package: beats-dashboards
 	$(MAKE) -C libbeat package-setup
 	$(foreach var,$(BEATS),SNAPSHOT=$(SNAPSHOT) $(MAKE) -C $(var) package || exit 1;)
 
+	# build the dashboards package
+	BUILD_DIR=${shell pwd}/build SNAPSHOT=$(SNAPSHOT) $(MAKE) -C dev-tools/packer package-dashboards ${shell pwd}/build/upload/build_id.txt
```
ruflin: We should make this available in libbeat. The reason for the refactoring of the packaging was to make packaging available in each beat which includes community beats. Community beats should be able to package their dashboards too. All the implementation in this Makefile should only call the Makefile commands in the betas itself and potentially gather some data.

PR #2301, opened 2016-08-18
New community Beat for monitoring Spring Boot apps

consulthys:


ruflin:
Thanks a lot for creating one more beat. We should have a conversation if it could make sense to add these beats as modules to Metricbeat.

consulthys:
Thanks Nicolas, yes definitely, anytime!

PR #2302, opened 2016-08-18
Make generate_index_pattern.py Python 3 compatible

rbaradari:
Change `print` to use parenthesis and use `string.ascii_letters` instead of just `string.letters` to be compatible with Python 3.

This is a similar fix to #1778 .

Parenthesized form of `print` and `string.ascii_letters` is also valid in Python 2.

The change to use `ascii_letters` would also fix locale dependent field names (anything non-ASCII was actually valid in case the locale was set).


ruflin:
I tested the change locally and it seems like it works as expected. @rbaradari Thanks a lot for the fix.

Issue #2299, opened 2016-08-17
Share Kibana dashboards


monicasarbu:
The purpose is to offer a way to make it easier for the users to share their own Kibana dashboards.

The following tasks need to be done:
- [x] re-write import_dashboards in Golang, instead of having two versions, one in bash and one in powershell (done in #2155)
- [x] create an archive with all the Beats dashboards #2298
- [x] compile the `import_dashboards` script on different OS, and include it in the Beat package #2303 
- [x] make it possible for community beats to package their dashboards. #2304
- [x] update `import_dashboards` script to download & import the Elastic Beats dashboards of the same version without the need to pass the full url. #2305
- [x] update docs https://github.com/elastic/beats/pull/2324

PR #2305, opened 2016-08-18
Set the default URL in the import_dashboards program

tsg:
The URL is computed based on the current libbeat version. Part of #2299.


libbeat/dashboards/import_dashboards.go
```diff
@@ -15,13 +15,14 @@ import (
 	"strings"
 	"time"
 
+	"github.com/elastic/beats/libbeat/beat"
```
ruflin: Nice trick

PR #2315, opened 2016-08-19
Make sure we run package-setup before package-dashboards

monicasarbu:
Make package-dashboards depend on package-setup, so we install all the dependencies before.

cc-ed @tsg

### DIFF
--- a/libbeat/scripts/Makefile
+++ b/libbeat/scripts/Makefile
@@ -397,7 +397,7 @@ package:
 
 	echo "Finished packages for ${BEATNAME}"
 
-package-dashboards:
+package-dashboards: package-setup
 	mkdir -p ${BUILD_DIR}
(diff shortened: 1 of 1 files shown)


libbeat/scripts/Makefile
```diff
@@ -397,7 +397,7 @@ package:
 
 	echo "Finished packages for ${BEATNAME}"
 
-package-dashboards:
+package-dashboards: package-setup
```
ruflin: The problem with `package-setup` is that it can take a lot of time and if `package-dashboards` is called multiple times, it is not required every time. That is also the reason it is not a dependency for `make package`

monicasarbu: Ok. It's not ideal, but we need to write in the docs that the users needs to run `package-setup` first.

PR #2317, opened 2016-08-19
Fix default logging path in config file

ruflin:
mybeat had to be replaced by beatname so it is correct for each beat.

Fixes some updates which were missing for the cassandra protocol.


tsg:
LGTM.

ruflin:
Should we replace it by 
```
  #path: ${path.logs}
```
instead?

PR #2318, opened 2016-08-19
Faster make target

tsg:
This change makes the build target be executed only if any of the go
files have changed. It looks for go files in the current directory +
everything under ES_BEATS (including vendor folder). This means it should
be fine for community beats as well, as long as they vendor their deps (I
think the default now).

The difference is significant, for example for metricbeat:

before:

```
> time make
go build
        7.14 real        16.77 user         3.07 sys
```

after:

```
> time make
make: `metricbeat' is up to date.
        0.89 real         0.79 user         0.09 sys
```


ruflin:
Nice. LGTM

PR #2320, opened 2016-08-19
Add verbose to import_dashboards

monicasarbu:
Print the url to the dashboards archive, that will be downloaded. This is useful in case  no url is passed to `import_dashboards` script, and the default url is used.


ruflin:
LGTM

PR #2322, opened 2016-08-19
Remove -e from the beatname.sh

tsg:
This makes it behave more similar with the init/system.d scripts, which helps troubleshoot some situations (e.g. logging). Related to #2321.

Also removes a -c flag forgotten.

### DIFF
--- a/dev-tools/packer/platforms/centos/beatname.sh.j2
+++ b/dev-tools/packer/platforms/centos/beatname.sh.j2
@@ -3,8 +3,7 @@
 # Script to run {.beat_name} in foreground with the same path settings that
 # the init script / systemd unit file would do.
 
-/usr/share/{{.beat_name}}/bin/{{.beat_name}} -e \
-  -c /etc/{{.beat_name}}/{{.beat_name}}.yml \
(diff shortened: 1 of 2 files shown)


dev-tools/packer/platforms/centos/beatname.sh.j2
```diff
@@ -3,8 +3,7 @@
 # Script to run {.beat_name} in foreground with the same path settings that
 # the init script / systemd unit file would do.
 
-/usr/share/{{.beat_name}}/bin/{{.beat_name}} -e \
-  -c /etc/{{.beat_name}}/{{.beat_name}}.yml \
```
ruflin: looks like the behaviour was already different between centos and debian?

ruflin:
LGTM

dev-tools/packer/platforms/centos/beatname.sh.j2
ruflin: a just saw the comment, that it was forgotten ...

tsg: Yes, unintentionally.

PR #2327, opened 2016-08-20
Add IP address to -devices command output

noricube:
On Windows, device description is a device driver name.
It is confusing when there are multiple same devices.

image

So, I added a IP address to -devices command output.
It is helpful to set up a sniffer interface.

image


ruflin:
Thanks for the addition. Please run `make fmt` to make sure your code adheres to the Golang coding guidelines. 

It would be nice if there would be a system test for these changes to confirm it works as expected. But as there is not test yet that you can just extend it is probably a little bit more complicated.

Could you also update the CHANGELOG.asciidoc file with this change?

noricube:
Hello.

I had to apply the golang coding guidelines.
And added changes to CHANGELOG.asciidoc.

Thanks.

ruflin:
jenkins, test it

ruflin:
@noricube Thanks a lot for the contribution.

PR #2335, opened 2016-08-22
Fix make import-dashboards command

ruflin:
* Make sure binary is built before executing command
* Update export-dashboard script: Special treatment for windows is not needed anymore as the index pattern id does not contain -* anymore.

### DIFF
--- a/dev-tools/export_dashboards.py
+++ b/dev-tools/export_dashboards.py
@@ -76,10 +76,6 @@ def ExportIndex(es, index, kibana_index, output_directory):
         doc_type="index-pattern",
         id=index)
 
-    # Fixes windows problem with files with * inside
-    # Removes it from index pattern
-    doc['_id'] = doc['_id'][:-2]
-
     # save index-pattern
     SaveJson("index-pattern", doc, output_directory)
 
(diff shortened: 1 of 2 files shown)


dev-tools/export_dashboards.py
```diff
@@ -119,7 +115,7 @@ def main():
     args = parser.parse_args()
 
     if args.index is None:
-        args.index = args.beat.lower() + "-*"
+        args.index = args.beat.lower()
```
ruflin: @monicasarbu Could you have a quick look at this one just to make sure this is as expected. It seems all the new index patterns id's don't have `-*` anymore.

monicasarbu: Yes, the index pattern is stored under `/.kibana/index-pattern/metricbeat`, and not `/.kibana/index-pattern/metricbeat-*`. I suggest to remove exporting the index pattern from the `export_dashboards.py`, as we don't want to overwrite the generated one.

monicasarbu:
I can merge this PR for now, and open another one to remove the exporting of the index pattern from the python script.

dev-tools/export_dashboards.py
ruflin: Good idea to remove it from the export script 👍

monicasarbu: Actually with this you found the bug that the index-pattern should be stored under /.kibana/index-pattern/metricbeat-* (and not metricbeat). This way, the name of the index pattern in Kibana is `metricbeat-*`, instead of `metricbeat`. This issue was available in the old script, and I copied in the new script as well. I opened a [PR](https://github.com/elastic/beats/pull/2348) to fix it in `import_dashboards.go`.

PR #2340, opened 2016-08-22
Add registrar stats to filebeat metricset

ruflin:
* Improve registrar metrics
* Add data.json files

### DIFF
--- a/filebeat/registrar/registrar.go
+++ b/filebeat/registrar/registrar.go
@@ -26,8 +26,10 @@ type Registrar struct {
 }
 
 var (
-	statesUpdated = expvar.NewInt("registrar.state_updates")
-	statesTotal   = expvar.NewInt("registar.states.total")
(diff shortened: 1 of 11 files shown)


filebeat/registrar/registrar.go
```diff
@@ -188,7 +190,9 @@ func (r *Registrar) Run() {
 
 		beforeCount := r.states.Count()
 		r.states.Cleanup()
```
andrewkroh: Consider making `Cleanup()` return the number of "states" that were removed. This way you are guaranteed to get the correct delta value.

ruflin: I initially was thinking about this but were undecided if states.Cleanup() should return the diff or the number of states that exist after the Cleanup. Will change it to return the diff.

ruflin:
New version pushed with the reviewes.

PR #2341, opened 2016-08-22
beatname.sh.j2: fix

cyrilleverrier:
Fix "flag provided but not defined: -path.log" when using the init script


ruflin:
LGTM. @tsg Can you also check this one?

tsg:
Hmm, i see quite a few `path.logs` references into config files and so on. It's possible they are all wrong, looking..

tsg:
Ah, I read it the other way around, nevermind. LGTM.

PR #2346, opened 2016-08-23
rename duplicate prospector test

spacewander:
...and then update the broken one.

### DIFF
--- a/filebeat/tests/system/test_prospector.py
+++ b/filebeat/tests/system/test_prospector.py
@@ -475,13 +475,13 @@ class Test(BaseTest):
         self.wait_until(
                 lambda: self.log_contains(
                     # Still checking for old file name as filename does not change in harvester
-                    "Closing file: {}\n".format(os.path.abspath(testfile))),
+                    "Stopping harvester, closing file: "),
                 max_timeout=10)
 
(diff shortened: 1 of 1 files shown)


ruflin:
@spacewander Thanks a lot for spotting this. I wonder why CI didn't complain about the duplicated method name?

spacewander:
@ruflin I personally recommend to integrate [pyflakes](https://pypi.python.org/pypi/pyflakes) into python development process. It can do things more than syntax check.

ruflin:
@spacewander Thanks for the recommendation. I will have a look. I would have expected python itself to complain :-)

ruflin:
@spacewander It seems like appveyor is not happy with the change. I can have a closer look if you want?

spacewander:
@ruflin Of course you can.

ruflin:
It seems the test on windows fails, because we compare paths and on windows the paths have \ instead of /. Perhaps we could simplify the test and only watch for `Stopping harvester, closing file:`?

spacewander:
@ruflin I have updated the pr and now all tests pass.

ruflin:
Merged. @spacewander Thanks a lot for the fix.

PR #2347, opened 2016-08-23
Fix Packetbeat dashboards

monicasarbu:
Fix the Navigation visualization by reverting to the older version and adding back the Cassandra dashboard.


medcl:
@monicasarbu 
can you help to remove `beats/packetbeat/etc/kibana/search/Cassandra.json` and `beats/packetbeat/etc/kibana/visualization/Cassandra.json`
Thanks, they are exported from kibana.

monicasarbu:
@medcl If you run `make export-dashboards` in the Packetbeat repository, then they should be overwritten.

monicasarbu:
Importing the dashboard, visualizations and searches work for me now, but you cannot visualize the Cassandra dashboard as there seems to be some issues with the index-pattern, that I would like to address in a different PR.

PR #2348, opened 2016-08-23
Fix index-pattern name in Kibana

monicasarbu:
Use packetbeat-* instead of packetbeat as name for the index pattern.


ruflin:
LGTM

PR #2351, opened 2016-08-23
Strip bom from message

### LINKED ISSUE #1349: Filebeat utf-8 encoding doesn’t honor BOM
Windows often prepend UTF-8 BOM to text files, which is legal - see [Can a UTF-8 data stream contain the BOM character (in UTF-8 form)?.](http://unicode.org/faq/utf_bom.html#bom5). In my case it happens in Exchange Message Tracking logs.

Filebeat should strip BOM from files with UTF-8 encoding which it doesn't do and BOM appears in message field for the first line in a file:
```
{
  "@timestamp":"2016-04-07T11:58:36.922Z",
  "beat":{
    "hostname":"XXXXX",
    "name":"XXXXX"},
  "count":1,
  "fields":null,
  "input_type":"log",
  "message":"<U+FEFF>#Software: Microsoft Exchange Server",
  "offset":0,
  "source":"exchange/MSGTRK20160405-1.LOG",
  "type":"exchange"
}
```

Filebeat config:
```
filebeat:
  prospectors:
    -
      document_type: exchange
      input_type: log
      paths:
        - exchange/MSGTRK2*.LOG
      encoding: utf-8

output:
  file:
    path: logstash/output
    name: exchange
```

Hex dump of first line in file MSGTRK20160405-1.LOG:
```
00000000  ef bb bf 23 53 6f 66 74  77 61 72 65 3a 20 4d 69  |...#Software: Mi|
00000010  63 72 6f 73 6f 66 74 20  45 78 63 68 61 6e 67 65  |crosoft Exchange|
00000020  20 53 65 72 76 65 72 0d  0a                       | Server..|
```
First three bytes EF BB BF (UTF-8 encoded BOM) are decoded to unicode character FE FF and it appears in message field.

I'm using filebeat-1.2.0-darwin.

ruflin:
Reading a file with a bom included the bom with the first event. This change removes the bom part from the first event in case it exists.

* Tests for utf-8 and utf-16 added

Closes #1349

### DIFF
--- a/filebeat/harvester/reader/strip_newline.go
+++ b/filebeat/harvester/reader/strip_newline.go
@@ -20,6 +20,7 @@ func (p *StripNewline) Next() (Message, error) {
 
 	L := message.Content
 	message.Content = L[:len(L)-lineEndingChars(L)]
+
 	return message, err
 }
 
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -95,7 +95,8 @@ https://github.com/elastic/beats/compare/v5.0.0-alpha5...master[Check the HEAD d
 *Topbeat*
 
 *Filebeat*
-- Introduce close_timeout harvester options {issue}1600[1600]
+- Introduce close_timeout harvester options {issue}1926[1926]
+- Strip BOM from first message in case of BOM files {issue}2351[2351]
 
 
 - Add harvester_limit option {pull}2417[2417]
--- a/filebeat/harvester/log.go
+++ b/filebeat/harvester/log.go
@@ -1,6 +1,7 @@
 package harvester
 
 import (
+	"bytes"
 	"errors"
 	"expvar"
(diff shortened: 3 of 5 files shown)


filebeat/harvester/reader/strip_newline.go
```diff
@@ -20,6 +22,10 @@ func (p *StripNewline) Next() (Message, error) {
 
 	L := message.Content
 	message.Content = L[:len(L)-lineEndingChars(L)]
+
+	// Strip BOM
+	message.Content = bytes.Trim(message.Content, "\xef\xbb\xbf")
```
urso: the BOM sequence can only be at start of file/stream. Should/Must not in between. What about utf-32 bom?

tsg:
Nice! Easier than I would have thought. I let @urso have a look as well.

andrewkroh:
LGTM

PR #2360, opened 2016-08-23
Remove old dev guide and point to new one (cherrypick #2353 into 1.2)

dedemorton:
@elastic/beats This PR removes the content about creating a new beat from the current doc branch (1.2) and adds a pointer to the dev guide in master. Now is the time to object if you think that we should also provide the doc in the current release, too. Note that the topic still exists; it just points to the dev guide in master for content.

A couple of related to-do items:
- [ ] Create redirects for the pages that we remove


tsg:
+1 on this change.

> Create redirects for the pages that we remove

Does that mean 302/301 redirect? That would be even better.

PR #2362, opened 2016-08-23
Remove old dev guide and point to new one (cherrypick #2353 into 5.0)

dedemorton:
Only merge this if  we decide to merge #2360


dedemorton:
@tsg If it makes merges easier, we can leave the doc guide in 5.0 for now. WDYT?

ruflin:
@dedemorton I leave this one for @tsg to merge based on the question above.

tsg:
Yeah, I think it's easier to leave the dev guide for now, because we might overwrite 5.0 with master (didn't figure out how to best do that yet), so changes to it will be lost. Closing it for now.

PR #2363, opened 2016-08-24
Use message struct to store internal state in multiline

ruflin:
* Rename clearBuffer to clear as it clears the reader
* Renamed resetBuffer to load as it loads data into the reader

### DIFF
--- a/filebeat/harvester/reader/multiline.go
+++ b/filebeat/harvester/reader/multiline.go
@@ -5,8 +5,6 @@ import (
 	"fmt"
 	"regexp"
 	"time"
-
-	"github.com/elastic/beats/libbeat/common"
 )
 
 // MultiLine reader combining multiple line events into one multi-line event.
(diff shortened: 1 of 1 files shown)


filebeat/harvester/reader/multiline.go
```diff
@@ -150,9 +146,9 @@ func (mlr *Multiline) readNext() (Message, error) {
 
 				// return collected multiline event and
 				// empty buffer for new multiline event
-				message := mlr.finalize()
+				m := mlr.finalize()
```
urso: maybe use `msg := mlr.finalize()` here. Makes it more clear `m` is of type `message`.

ruflin: Agree. I used `m` because it is used in all other place. I changed all to `msg`.

ruflin:
jenkins, retest it

PR #2365, opened 2016-08-24
Prospector only update states when harvester finished

ruflin:
This is to prevent any later overwriting of the state by the harvester again. Tests were adjusted accordingly.


tsg:
LGTM.

PR #2371, opened 2016-08-24
Change spooler timeout flush to happen after lastFlush + timeout

ruflin:
In the previous code, the flush always happened after timeout, independent if a flush just happened before. This could lead to unnecessary flushes. In case the spooler is full and data is flushed, the ticker will be reset.

Previously the ticker was set to timeout/2. Now it is timeout.

### DIFF
--- a/filebeat/spooler/spooler.go
+++ b/filebeat/spooler/spooler.go
@@ -16,13 +16,12 @@ const channelSize = 16
 
 // Spooler aggregates the events and sends the aggregated data to the publisher.
 type Spooler struct {
-	Channel       chan *input.Event // Channel is the input to the Spooler.
-	config        spoolerConfig
-	exit          chan struct{}         // Channel used to signal shutdown.
-	nextFlushTime time.Time             // Scheduled time of the next flush.
-	publisher     chan<- []*input.Event // Channel used to publish events.
-	spool         []*input.Event        // Events being held by the Spooler.
-	wg            sync.WaitGroup        // WaitGroup used to control the shutdown.
+	Channel   chan *input.Event // Channel is the input to the Spooler.
+	config    spoolerConfig
+	exit      chan struct{}         // Channel used to signal shutdown.
+	publisher chan<- []*input.Event // Channel used to publish events.
+	spool     []*input.Event        // Events being held by the Spooler.
+	wg        sync.WaitGroup        // WaitGroup used to control the shutdown.
 }
 
 type spoolerConfig struct {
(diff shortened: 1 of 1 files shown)


filebeat/spooler/spooler.go
```diff
@@ -61,24 +59,32 @@ func (s *Spooler) Start() {
...
 			if event != nil {
 				s.queue(event)
 			}
-		case <-ticker.C:
-			s.timedFlush()
+
+			// Stop timer and drain channel. See https://golang.org/pkg/time/#Timer.Reset
+			if !timer.Stop() {
+				<-timer.C
+			}
```
urso: the timer shall be stopped only when queue is flushed, not per event. e.g. `queue` can return a `bool` indicating if queue has been flushed.

filebeat/spooler/spooler.go
```diff
@@ -61,24 +59,32 @@ func (s *Spooler) Start() {
...
+
+		case <-timer.C:
+			debugf("Flushing spooler because of timeout. Events flushed: %v", len(s.spool))
+			s.flush()
 		}
+
+		timer.Reset(s.config.idleTimeout)
```
urso: not sure, but we need to reset always? Even after having received a time event?

In case flush gets blocked, it indeed makes sense to reset timer. but then the timer should be stopped at this place. Otherwise a timer signal is still in queue if flush took very very looong.

Timer should be stopped + reset here only if buffer has been flushed.

filebeat/spooler/spooler.go
ruflin: Fixed

filebeat/spooler/spooler.go
ruflin: I initially thought we don't need to reset the timer after the timer event happened, but this seems to be the case. Changed.

filebeat/spooler/spooler.go
```diff
@@ -61,23 +59,32 @@ func (s *Spooler) Start() {
...
+					}
+					timer.Reset(s.config.idleTimeout)
+				}
 			}
-		case <-ticker.C:
-			s.timedFlush()
+		case <-timer.C:
+			debugf("Flushing spooler because of timeout. Events flushed: %v", len(s.spool))
+			s.flush()
+			timer.Reset(s.config.idleTimeout)
```
urso: what if `s.flush()` took like 5 minutes, but idleTimeout is 10s? Any signal in timer.C in this case? Consider draining timer.C here as well.

ruflin: That means we would stop and drain the timer before the flush and reset it only afterwards?

ruflin: @urso Having a look again at the code I'm not sure if this needs a change. As we use a timer and not a ticker, as long as we don't reset there shouldn't be anything in the timer I would think?

urso: right, timer fires only once.

PR #2372, opened 2016-08-24
Change sha files format

tsg:
The unified build process uses another format for the
SHA files, so aligning with that for easier testing. We'll
be able to remove our generation of sha1 files completely after we switch to using the unified build for snapshots as well.


tsg:
jenkins, package it

ruflin:
@tsg Will this mean community beats package will not create any sha1 anymore?

tsg:
@ruflin Good point, we can keep them for the community beats.

PR #2380, opened 2016-08-25
Fix unzipping of any zip archive with the dashboards

monicasarbu:
This PR fixes unzipping of any zip archive that contains the dashboards. The zip archive can have either a separate directory containing subdirectories: `index-pattern`, `dashboard`, `search`, and `visualization`, or directly these directories.

If any of the directories is missing, no error is printed.


ruflin:
LGTM. What is the reason we need to support both? Shouldn't there only be one option?

monicasarbu:
@ruflin I updated the description to answer your questions. Thanks for merging.

ruflin:
@monicasarbu thanks

Issue #2247, opened 2016-08-11
Adding php-fpm module to metricbeat


hartfordfive:
I'm currently working on adding both of these modules to metricbeats although I figured i'd create an issue for this so the community can see it's currently in progress in case anyone is interested in using them.

ruflin:
@hartfordfive Great to hear you are working on these modules. Please let us know if hit some roadblocks or need some inputs.

hartfordfive:
@ruflin So i'm trying to create and run the necessary tests for the haproxy module although i'm getting this error when running the `make integration-tests-environment` command:

```
docker-compose -f /Users/al/Documents/code/github/elastic-beats/metricbeat/../testing/environments/base.yml -f /Users/al/Documents/code/github/elastic-beats/metricbeat/../testing/environments/snapshot.yml -f docker-compose.yml run beat make integration-tests RACE_DETECTOR=
Starting elasticsearch
ERROR: Cannot start service beat: mkdir /Users/al/Documents/code/github/elastic-beats: file exists
make: *** [integration-tests-environment] Error 1
```

I'm running the command in the local directory "/Users/al/Documents/code/github/elastic-beats/metricbeat"  Any ideas exactly what it is that I'm I doing wrong here?  You can find my current code here: https://github.com/hartfordfive/beats/tree/metricbeat-haproxy

ruflin:
@hartfordfive I think the issue is with your GOPATH setup. Golang requires you to have the project in the correct directory of your $GOPATH. So in this case the project must be under `$GOPATH/src/github.com/elastic/beats` otherwise it will not work.

hartfordfive:
@ruflin Sorry I forgot to mention something.  My current path `/Users/al/Documents/code/github/elastic-beats/metricbeat` is actually just a sym link to the correct project Go directory `/usr/local/go/src/github.com/elastic/beats/`.   This made things a bit more convenient for me in my IDE.

ruflin:
@hartfordfive TBH I don't know what golang does with the symlinks and especially our Makefile magic. Could you try to run the commands in the directory itself?

hartfordfive:
@ruflin I've tried what you've suggested and had some progress although now I'm getting new errors.  First, it seems like the test just waits for HAProxy to start, but it never ends up starting.

```
Waiting for HAProxy(haproxy:14567) to start..............................................
HAProxy is not available
Address: haproxy:14567
```

[...]

PR #2384, opened 2016-08-25
Adding HAProxy module

hartfordfive:
PR for issue #2247


metricbeat/logs/metricbeat
```diff
@@ -0,0 +1,10 @@
+2016-08-10T19:06:01-04:00 INFO Metrics logging every 30s
```
ruflin: The logs directory shuld not be commited. it is under .gitignore in the master branch

metricbeat/module/haproxy/haproxy.go
```diff
@@ -0,0 +1,244 @@
...
+		return nil, err
+	}
+
+	if b, err := ioutil.ReadAll(res); err == nil {
+
+		resultMap := map[string]interface{}{}
+
+		for _, ln := range strings.Split(string(b), "\n") {
+
+			ln := strings.Trim(ln, " ")
```
andrewkroh: There is the [`strings.TrimSpace`](https://golang.org/pkg/strings/#TrimSpace) function for this.

metricbeat/module/haproxy/haproxy.go
```diff
@@ -0,0 +1,244 @@
...
+			if ln == "" {
+				continue
+			}
+
+			parts := strings.Split(strings.Trim(ln, " "), ":")
```
andrewkroh: `ln` has already been trimmed at this point.

metricbeat/module/haproxy/haproxy.go
```diff
@@ -0,0 +1,244 @@
...
+				continue
+			}
+
+			resultMap[parts[0]] = strings.Trim(parts[1], " ")
+		}
+
+		var result *Info
+		err := mapstructure.Decode(resultMap, &result)
+		if err != nil {
+			panic(err)
```
andrewkroh: Return the error rather than `panic`'ing.

metricbeat/module/haproxy/haproxy.go
```diff
@@ -0,0 +1,244 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("Error parsing CSV: %s", err)
+	}
+
+	return statRes, nil
+
+}
+
+// GetInfo returns the result from the 'show stat' command
+func (c *Client) GetInfo() (infoRes *Info, err error) {
```
andrewkroh: Don't use named return parameters unless you have good reason to. This applies to other parts of the code. https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters

metricbeat/module/haproxy/info/_meta/fields.yml
```diff
@@ -0,0 +1,194 @@
+- name: info
+  type: group
+  description: >
+    General infomration collected on HAProxy process
+  fields:
+    - name: nb_proc
+      type: intger
```
andrewkroh: `s/intger/integer/`

But instead of integer use `long`.

metricbeat/module/haproxy/info/_meta/fields.yml
```diff
@@ -0,0 +1,194 @@
...
+
+    - name: tasks
+      type: intger
+      description: >
+
+    - name: run_queue
+      type: intger
+      description: >
+
+    - name: idle_pct
```
andrewkroh: Consider making this value range on [0, 1] so you can use a float with a percentage formatter. For example: https://github.com/elastic/beats/blob/master/metricbeat/module/system/memory/_meta/fields.yml#L26-L27

metricbeat/module/haproxy/info/info.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+// Fetch methods implements the data gathering and data conversion to the right format
+// It returns the event which is then forward to the output. In case of an error, a
+// descriptive error must be returned.
+func (m *MetricSet) Fetch() (common.MapStr, error) {
+
+	hapc, err := haproxy.NewHaproxyClient(m.statsAddr)
+	if err != nil {
+		return nil, fmt.Errorf(fmt.Sprintf("HAProxy Client error: %s", err))
```
andrewkroh: `fmt.Errorf()` accepts the same parameters as `fmt.Sprintf` so you can just drop the call to `Sprintf`.

metricbeat/module/haproxy/haproxy.go
```diff
@@ -0,0 +1,244 @@
...
+// Run sends a designated command to the haproxy stats socket
+func (c *Client) run(cmd string) (*bytes.Buffer, error) {
+	var conn net.Conn
+	response := bytes.NewBuffer(nil)
+
+	conn, err := net.Dial(c.ProtoScheme, c.Address)
+	if err != nil {
+		return response, err
+	}
+	c.connection = conn
```
andrewkroh: Why store the connection in the `Client` if it's not going to be reused?

andrewkroh:
@hartfordfive Looks like you are off to a great start. Thanks for the contribution.

metricbeat/module/haproxy/haproxy.go
hartfordfive: Good point.  I've changed that.

hartfordfive:
@andrewkroh I think I've fixed most of the issues you've commented on although I have a few other HAProxy stats related bugs, so I'll let you know once that's done and I'm ready for a final review.

hartfordfive:
@andrewkroh I think the PR should be good for another review when ever you get the chance.

ruflin:
@hartfordfive Any chance you could rebase the commits or merge in master so the merging is possible? Just to make we review the most recent changes on top of master. It seems that you merged in master so rebasing could become ugly. TBH not sure what the best recommendation is here as you have lots of commits. Probably best keep merging in master and we will to the squash and rebase on our side at the end.

metricbeat/.gitignore
```diff
@@ -3,3 +3,4 @@ build
 /metricbeat
 /metricbeat.test
 /docs/html_docs
+/logs/*
```
ruflin: This should be not needed anymore as soon as this is rebased on top of master or master is merged.

metricbeat/logs/metricbeat
```diff
@@ -0,0 +1,10 @@
...
+2016-08-10T19:06:01-04:00 INFO Loading template enabled. Reading template file: /Users/alain.lefebvre/Documents/code/github_personal/beats/metricbeat/metricbeat.template.json
```
ruflin: This file must be removed manually as it is already commited.

metricbeat/metricbeat.yml
```diff
@@ -36,13 +36,24 @@ metricbeat.modules:
 
     # Per process stats
     - process
-  enabled: true
+  enabled: false
```
ruflin: This change should be reverted

metricbeat/metricbeat.yml
```diff
@@ -36,13 +36,24 @@ metricbeat.modules:
...
 
   # if true, exports the CPU usage in ticks, together with the percentage values
   cpu_ticks: false
 
+#------------------------------- haproxy Module ------------------------------
+- module: haproxy
+  metricsets: 
+    - "stat"
+    - "info"
+  enabled: true
```
ruflin: haproxy should be disabled by default in the short config

metricbeat/module/haproxy/_meta/haproxy.conf
```diff
@@ -0,0 +1,39 @@
...
+  maxconn 4000
+  #daemon
+  user  root
+  group staff
+
+  # for restarts
+  pidfile /var/run/haproxy.pid
+  # Logging to syslog facility local0
+  log   127.0.0.1       local0
+  stats socket 0.0.0.0:14567
```
ruflin: Is this something that is enabled by default in most haproxy setups? If not, could be worth to mention this in the haproxy module docs.

metricbeat/module/haproxy/info/_meta/data.json
```diff
@@ -0,0 +1,59 @@
+{
+  "@timestamp": "2016-08-13T19:44:21.036Z",
+  "beat": {
+    "hostname": "beathost",
+    "name": "beathost"
+  },
+  "haproxy": {
+    "info": {
+      "compress_bps_in": 0,
```
ruflin: The event conventions should be applied here: https://www.elastic.co/guide/en/beats/libbeat/5.0/event-conventions.html

ruflin:
@hartfordfive @andrewkroh I suggest we mark the haproxy metricsets as experimental first. We do this at first for all new metricsets to make it possible to the metricsets in early and test them, but still change the data structure if needed.

@hartfordfive Making it experimental only means adding a log message to the Metricset and a flag in the docs. Have a look at filebeat metricset as an example.

hartfordfive:
@ruflin @andrewkroh I agree it would be best to flag this module as experimental for while before it's considered stable.  I'll apply those other changes you mentioned above shortly.

metricbeat/module/haproxy/_meta/docs.asciidoc
```diff
@@ -0,0 +1,9 @@
+== haproxy Module
+
+This is the haproxy Module.  To enable stats collection from HAProxy, you must enable the stats socket via TCP.  
+For example, placing the following statement under the `global` or `default` section of the haproxy config:
+
+`stats socket 0.0.0.0:14567`
```
andrewkroh: I recommend providing a more secure example. Either show an example that binds only to `127.0.0.1` or use a unix socket.

metricbeat/module/haproxy/haproxy.go
```diff
@@ -0,0 +1,243 @@
+package haproxy
+
+import (
+	"bytes"
+	"encoding/csv"
+	"errors"
+	"fmt"
+	"github.com/gocarina/gocsv"
```
andrewkroh: Please separate the stdlib imports from the third-party imports. Most of our code follows this style if you would like to see some examples. Checkout the [goimports](http://goinbigdata.com/goimports-vs-gofmt/#goimports) tool.

metricbeat/module/haproxy/haproxy.go
```diff
@@ -0,0 +1,243 @@
...
+
+// NewHaproxyClient returns a new instance of HaproxyClient
+func NewHaproxyClient(address string) (*Client, error) {
+	parts := strings.Split(address, "://")
+	if len(parts) != 2 {
+		return nil, errors.New("Must have protocol scheme and address!")
+	}
+
+	if parts[0] != "tcp" && parts[0] != "unix" {
+		return nil, errors.New("Invalid Protocol Scheme!")
```
andrewkroh: Error strings should not be capitalized. https://github.com/golang/go/wiki/CodeReviewComments#error-strings

metricbeat/module/haproxy/haproxy.go
```diff
@@ -0,0 +1,243 @@
...
+			if len(parts) != 2 {
+				continue
+			}
+
+			resultMap[parts[0]] = strings.TrimSpace(parts[1])
+		}
+
+		var result *Info
+		err := mapstructure.Decode(resultMap, &result)
+		if err != nil {
```
andrewkroh: Consider writing this which is slightly more concise. (not a mandatory change, it's just how I'd of written it)

```
if err := mapstructure.Decode(resultMap, &result); err != nil {
    return nil, err
}
```

metricbeat/module/haproxy/info/_meta/fields.yml
```diff
@@ -0,0 +1,196 @@
...
+        Process ID
+
+    - name: uptime_sec
+      type: integer
+      description: >
+        Current uptime in seconds
+
+    - name: mem_max_mb
+      type: integer
+      format: bytes
```
andrewkroh: If the value is actually in MB then you don't want to format it as `bytes`. I recommend converting the value to bytes and renaming the field so you can take advantage of the byte formatter in Kibana (it uses http://numeraljs.com/).

andrewkroh: Oh, I see you do convert the value. So then I think the field just needs a rename to reflect the correct units.

metricbeat/module/haproxy/info/data.go
```diff
@@ -0,0 +1,95 @@
...
+	source = map[string]interface{}{}
+
+	for i := 0; i < st.NumField(); i++ {
+		f := st.Field(i)
+
+		if typeOfT.Field(i).Name == "IdlePct" {
+			// Convert this value to a float between 0.0 and 1.0
+			fval, err := strconv.ParseFloat(f.Interface().(string), 64)
+			if err != nil {
+				panic(err)
```
andrewkroh: I don't recommend panic'ing. (https://github.com/golang/go/wiki/CodeReviewComments#dont-panic) Instead, either propagate the error or handle it.

metricbeat/module/haproxy/stat/data.go
```diff
@@ -0,0 +1,138 @@
...
+
+	}
+	return results
+}
+
+// Map data to MapStr
+func eventMapping(info []*haproxy.Stat) []common.MapStr {
+
+	var events []common.MapStr
+	source := map[string]interface{}{}
```
andrewkroh: Doesn't look like this allocated map is ever used. You can remove it and declare `source` within the loop.

metricbeat/module/haproxy/info/data.go
```diff
@@ -0,0 +1,95 @@
...
+
+// Map data to MapStr
+func eventMapping(info *haproxy.Info) common.MapStr {
+	// Full mapping from info
+
+	source := map[string]interface{}{}
+
+	st := reflect.ValueOf(info).Elem()
+	typeOfT := st.Type()
+	source = map[string]interface{}{}
```
andrewkroh: This is allocating a second map. I don't see a reason for it given the `source` was allocated above.

andrewkroh:
I looked over the Golang code and left a few comments. 

I didn't review the field names very closely to check if they conform to the [event conventions](https://www.elastic.co/guide/en/beats/libbeat/5.0/event-conventions.html), nor did I review the field documentation closely (maybe @dedemorton want to look at the docs, or maybe she prefers to edit them after they are merged).

dedemorton:
It's easier for me to edit the doc changes after they are merged. Please make sure you do a local doc build before the changes are merged to ensure that your contribution doesn't break the doc build. Many thanks!

metricbeat/module/haproxy/haproxy.go
hartfordfive: Yes I definitely agree it's cleaner that way, I'll definitely change that.

hartfordfive:
I've applied most of the modifications for the comments you mentioned above @andrewkroh although i still have to look into the event conventions.  @dedemorton as for the local doc build, where can I see some details regarding that procedure?

andrewkroh:
@hartfordfive  Can you fix the merge conflicts. You probably need to rebase and address a few conflicts.

Once updated I can test it on Jenkins which will also do a doc build so you don't have to. But if you want to test locally, the command is `make docs`.

hartfordfive:
@andrewkroh I actually merged in the master and fixed the merge conflicts a few days ago. Should I do it again?

hartfordfive:
@andrewkroh Ok I've done a few other changes and also updated the index templates to follow the updated event conventions.  Let me know how it looks.

metricbeat/etc/beat.yml
```diff
@@ -43,4 +43,6 @@ metricbeat.modules:
   period: 10s
   processes: ['.*']
 
+  # if true, exports the CPU usage in ticks, together with the percentage values
```
ruflin: Not sure how this change neded up in this PR?

vendor/vendor.json
```diff
@@ -0,0 +1,19 @@
+{
```
ruflin: Which tool creates this file?

vendor/github.com/gocarina/gocsv/README.md
```diff
@@ -0,0 +1,151 @@
+Go CSV
```
ruflin: So far we added a vendor directory in the module directly if the dependency was specific to the module. TBH with csv it is fine from my point of view to have it in the top dependencies as this will be also used in other modules.

ruflin:
@hartfordfive Thanks for all the work on this one. I think there are still parts we can tweak in the field naming but for all the other modules we also had here multiple iteration. I left a few very minor comments. The most important one for me is that we mark it experimental. See here https://github.com/elastic/beats/blob/master/metricbeat/module/beats/filebeat/filebeat.go#L26 as an example. This will allow us to change fields and structure with follow up PR's without someone expecting that this is already stable. From my point of view it would be ok to merge it soonish and then go through the hole module again and tweak it in a separate PR. But lets also get @andrewkroh view on this.

vendor/vendor.json
hartfordfive: Sorry, I actually missed that file.  The tool used was [govendor](https://github.com/kardianos/govendor).  If you guys have another standard way of bringing in vendor dependencies, let me know and I'll do that instead.

vendor/github.com/gocarina/gocsv/README.md
hartfordfive: I'm not exactly sure what I should change in this case?

metricbeat/etc/beat.yml
hartfordfive: I believe I got these changes when merging in the latest master the other day into my branch.

hartfordfive:
@ruflin I've added the experimental warning messages as suggested.

metricbeat/module/haproxy/stat/stat.go
```diff
@@ -65,13 +67,13 @@ func (m *MetricSet) Fetch() ([]common.MapStr, error) {
 
 	hapc, err := haproxy.NewHaproxyClient(m.statsAddr)
 	if err != nil {
-		return nil, fmt.Errorf(fmt.Sprintf("HAProxy Client error: %s", err))
+		return nil, fmt.Errorf("HAProxy Client error: %s", err)
```
ruflin: errors.New would also work here. But that is a minor detail which can be tackled later.

metricbeat/etc/beat.yml
ruflin: Does not seem to be in master: https://github.com/elastic/beats/blob/master/metricbeat/etc/beat.yml Could you run `git checkout master metricbeat/etc/beat.yml` ?

metricbeat/metricbeat.yml
ruflin: @hartfordfive Can you please also revert this change?

metricbeat/metricbeat.yml
ruflin: @hartfordfive This also should be set to false by default.

ruflin:
@hartfordfive Thanks. Please check again my most recent comments above. I think it is important that we don't change our global defaults. Could be that these change happened sometimes during merging.

ruflin:
@hartfordfive Merged. Thanks a lot for your contribution. I opened a follow up PR here to do some fixes like fixing the doc build: https://github.com/elastic/beats/pull/2486

I also opened the github issue here to track some improvements that we have to do: https://github.com/elastic/beats/issues/2485

PR #2386, opened 2016-08-25
Generic event conversion of pointers, arrays, slices

andrewkroh:
Changes
- Dereference any pointers in the event.
- Apply generic event conversion to arrays and slices in the event.
- Drop nil values from MapStr. This does not apply to arrays, structs, or map types converted via JSON marshal/unmarshal.
- Drop keys containing values of unknown types (previously it only logged a warning).
- Provide a detailed warning when the generic event conversion is unsuccessful. A message is logged contained the map key that failed, the cause, and the event (for context).
- Due to the enhancement to normalize types in arrays, a change was required in Packetbeat's memcache proto.
- Changed a test case in Filebeat to not expect a null value.

### DIFF
--- a/libbeat/common/event.go
+++ b/libbeat/common/event.go
@@ -1,92 +1,223 @@
 package common
 
 import (
+	"encoding"
 	"encoding/json"
+	"fmt"
 	"reflect"
-	"time"
+	"strconv"
+	"strings"
 
 	"github.com/elastic/beats/libbeat/logp"
+
+	"github.com/pkg/errors"
 )
 
-func MarshallUnmarshall(v interface{}) (MapStr, error) {
-	// decode and encode JSON
-	marshaled, err := json.Marshal(v)
-	if err != nil {
-		logp.Warn("marshal err: %v", err)
-		return nil, err
-	}
-	var v1 MapStr
-	err = json.Unmarshal(marshaled, &v1)
-	if err != nil {
-		logp.Warn("unmarshal err: %v", err)
-		return nil, err
-	}
+const eventDebugSelector = "event"
 
-	return v1, nil
+var eventDebugf = logp.MakeDebug(eventDebugSelector)
+
+var textMarshalerType = reflect.TypeOf((*encoding.TextMarshaler)(nil)).Elem()
+
+// ConvertToGenericEvent normalizes the types contained in the given MapStr.
+//
+// Nil values in maps are dropped during the conversion. Any unsupported types
+// that are found in the MapStr are dropped and warnings are logged.
+func ConvertToGenericEvent(m MapStr) MapStr {
+	event, errs := normalizeMap("", m)
+	if len(errs) > 0 {
+		logp.Warn("Unsuccessful conversion to generic event: %v errors: %v, "+
+			"event=%#v", len(errs), errs, m)
+	}
+	return event
 }
 
-func ConvertToGenericEvent(v MapStr) MapStr {
+// normalizeMap normalizes each element contained in the given map. If an error
+// occurs during normalization, processing of m will continue, and all errors
+// are returned at the end.
+func normalizeMap(baseKey string, m MapStr) (MapStr, []error) {
+	var errs []error
 
-	for key, value := range v {
+	out := make(MapStr, len(m))
+	for key, value := range m {
+		fullKey := joinKeys(baseKey, key)
+		v, err := normalizeValue(fullKey, value)
+		if len(err) > 0 {
+			errs = append(errs, err...)
+		}
 
-		if value == nil {
-			// leave nil values alone
+		// Drop nil values from maps.
+		if v == nil {
+			if logp.IsDebug(eventDebugSelector) {
+				eventDebugf("Dropped nil value from event where key=%v", fullKey)
+			}
 			continue
 		}
 
-		switch value.(type) {
-		case Time, *Time:
-			continue
-		case time.Location, *time.Location:
-			continue
-		case MapStr:
-			v[key] = ConvertToGenericEvent(value.(MapStr))
-			continue
-		case *MapStr:
-			v[key] = ConvertToGenericEvent(*value.(*MapStr))
-			continue
+		out[key] = v
+	}
+
+	return out, errs
+}
+
+// normalizeMapStrSlice normalizes each individual MapStr.
+func normalizeMapStrSlice(baseKey string, maps []MapStr) ([]MapStr, []error) {
+	var errs []error
+
+	out := make([]MapStr, 0, len(maps))
+	for i, m := range maps {
+		normalizedMap, err := normalizeMap(joinKeys(baseKey, strconv.Itoa(i)), m)
+		if len(err) > 0 {
+			errs = append(errs, err...)
+		}
+		out = append(out, normalizedMap)
+	}
+
+	return out, errs
+}
+
+// normalizeMapStringSlice normalizes each individual map[string]interface{} and
+// returns a []MapStr.
+func normalizeMapStringSlice(baseKey string, maps []map[string]interface{}) ([]MapStr, []error) {
+	var errs []error
+
+	out := make([]MapStr, 0, len(maps))
+	for i, m := range maps {
+		normalizedMap, err := normalizeMap(joinKeys(baseKey, strconv.Itoa(i)), m)
+		if len(err) > 0 {
+			errs = append(errs, err...)
+		}
+		out = append(out, normalizedMap)
+	}
+
+	return out, errs
+}
+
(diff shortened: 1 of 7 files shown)


libbeat/common/event.go
```diff
@@ -1,92 +1,223 @@
...
+func normalizeSlice(baseKey string, v reflect.Value) (interface{}, []error) {
+	var errs []error
+	var sliceValues []interface{}
 
-			if typ.Kind() == reflect.Ptr {
-				typ = typ.Elem()
-			}
+	n := v.Len()
+	for i := 0; i < n; i++ {
+		sliceValue, err := normalizeValue(joinKeys(baseKey, strconv.Itoa(i)), v.Index(i).Interface())
```
tsg: Should we skip null values here as well?

tsg:
@andrewkroh Consider also adding a benchmark with a few common conversions, so we have a performance baseline for future changes. I am a bit worried about the pointer following calling reflect for all values, but hopefully the cost of it is small enough.

libbeat/common/event_test.go
```diff
@@ -122,10 +150,149 @@ func TestConvertNestedStruct(t *testing.T) {
...
+		{NetString("test"), "test"}, // It honors the TextMarshaler contract.
+		{true, true},
+		{int8(8), int8(8)},
+		{uint8(8), uint8(8)},
+		{"hello", "hello"},
+		{map[string]interface{}{"foo": "bar"}, MapStr{"foo": "bar"}},
+
+		// Other map types are converted using marshalUnmarshal which will lose
+		// type information for arrays which become []interface{} and numbers
+		// which all become float64.
```
tsg: We could apply the same strategy as we do in Filebeat to auto-detect ints instead of always converting to floats. See this: https://github.com/elastic/beats/blob/master/filebeat/harvester/reader/json.go#L63

libbeat/common/event.go
```diff
@@ -1,92 +1,223 @@
...
+
+// normalizeMap normalizes each element contained in the given map. If an error
+// occurs during normalization, processing of m will continue, and all errors
+// are returned at the end.
+func normalizeMap(baseKey string, m MapStr) (MapStr, []error) {
+	var errs []error
+
+	out := make(MapStr, len(m))
+	for key, value := range m {
+		fullKey := joinKeys(baseKey, key)
```
tsg: The full key is only used for error reporting, right?

libbeat/common/event.go
```diff
@@ -1,92 +1,223 @@
...
-				if err != nil {
-					logp.Warn("fail to marshall & unmarshall struct %v", key)
-					continue
-				}
-				v[key] = anothermap
-			default:
-				logp.Warn("unknown type %v", typ)
-				continue
+		switch v.Type().Kind() {
+		case reflect.Bool:
```
monicasarbu: Is there any reason that for basic types (bool, int, uint, float, complex, string) to do reflection? We could return their value in the first switch, and leave the default of the first switch only for special types, like map, struct, array.

libbeat/common/event.go
```diff
@@ -1,92 +1,223 @@
...
+var textMarshalerType = reflect.TypeOf((*encoding.TextMarshaler)(nil)).Elem()
+
+// ConvertToGenericEvent normalizes the types contained in the given MapStr.
+//
+// Nil values in maps are dropped during the conversion. Any unsupported types
+// that are found in the MapStr are dropped and warnings are logged.
+func ConvertToGenericEvent(m MapStr) MapStr {
+	event, errs := normalizeMap("", m)
+	if len(errs) > 0 {
+		logp.Warn("Unsuccessful conversion to generic event: %v errors: %v, "+
```
ruflin: Should we log this as an Error? In general we tried to get rid of warnings.

libbeat/common/event.go
andrewkroh: In order to handle types like `type myAwesomeInt int32` reflection is used. Maybe be don't want to support `myAwesomeInt` in which case we could drop these case statements from the reflection block?

libbeat/common/event.go
andrewkroh: Good question, I was think we might want to leave arrays alone in case the index conveys meaning. For example, if some application using structured logging writes an array of values from various sensors, dropping one of those values could cause problems.

andrewkroh:
> @andrewkroh Consider also adding a benchmark with a few common conversions, so we have a performance baseline for future changes. I am a bit worried about the copying and the pointer following calling reflect for all values, but hopefully the cost of it is small enough.

I changed up the code to only attempt to deference the pointer after testing for other types first. And add some benchmarks for comparisons.

```
# Before
BenchmarkConvertToGenericEventNetString-4           	 3000000	       494 ns/op	     376 B/op	       4 allocs/op
BenchmarkConvertToGenericEventMapStringString-4     	  500000	      4104 ns/op	    1784 B/op	      23 allocs/op
BenchmarkConvertToGenericEventMapStr-4              	  300000	      4273 ns/op	    1808 B/op	      25 allocs/op
BenchmarkConvertToGenericEventStringSlice-4         	 3000000	       496 ns/op	     400 B/op	       4 allocs/op
BenchmarkConvertToGenericEventCustomStringSlice-4   	 3000000	       497 ns/op	     400 B/op	       4 allocs/op

# After
BenchmarkConvertToGenericEventNetString-4           	 2000000	       904 ns/op	     736 B/op	       8 allocs/op
BenchmarkConvertToGenericEventMapStringString-4     	  300000	      4167 ns/op	    2120 B/op	      25 allocs/op
BenchmarkConvertToGenericEventMapStr-4              	 1000000	      1443 ns/op	    1360 B/op	       9 allocs/op
BenchmarkConvertToGenericEventStringSlice-4         	 2000000	       790 ns/op	     736 B/op	       6 allocs/op
BenchmarkConvertToGenericEventCustomStringSlice-4   	 1000000	      1504 ns/op	     882 B/op	      15 allocs/op
BenchmarkConvertToGenericEventStringPointer-4       	 2000000	       845 ns/op	     688 B/op	       5 allocs/op

# Modifying the input map (not returning a copy)
# Things will get faster if we can track down
# what causes the race condition.
BenchmarkConvertToGenericEventNetString-4           	 2000000	       676 ns/op	     400 B/op	       6 allocs/op
BenchmarkConvertToGenericEventMapStringString-4     	  500000	      3857 ns/op	    1784 B/op	      23 allocs/op
BenchmarkConvertToGenericEventMapStr-4              	 1000000	      1013 ns/op	     688 B/op	       5 allocs/op
BenchmarkConvertToGenericEventStringSlice-4         	 3000000	       561 ns/op	     400 B/op	       4 allocs/op
BenchmarkConvertToGenericEventCustomStringSlice-4   	 1000000	      1315 ns/op	     546 B/op	      13 allocs/op
BenchmarkConvertToGenericEventStringPointer-4       	 2000000	       635 ns/op	     352 B/op	       3 allocs/op
```

libbeat/common/event_test.go
andrewkroh: I would prefer to disallow the types that require marshal/unmarshal.

libbeat/common/event.go
andrewkroh: Yeah, it's only for error reporting. I made some change to minimize the number of allocations associated with this feature. And now it only concatenates the keys if an error occurs.

libbeat/common/event.go
andrewkroh: I'm thinking instead of logging, it should return the error to the caller. Though, I'm on on the fence about this. Originally this method logged the issues it detected and still passed back a `MapStr` to the caller. It had weak guarantees about the resulting MapStr being "generic".

I'm thinking the method signature should be `func ConvertToGenericEvent(m MapStr) (MapStr, error)` and on error we return a `nil` MapStr. This means that if you write a bad event, it will be dropped (but logged).

andrewkroh:
This is the race condition I'm seeing if I modify `ConvertToGenericEvent()` to modify the input parameter (vs returning a copy of the map). http://pastebin.com/raw/KCL0mTkH

tsg:
@andrewkroh Thanks for doing the tests. I'd say that the performance numbers are good enough that we don't really have to worry about, most likely the bottleneck is going to be somewhere else. And if the profiling shows that it's here, people can do the conversion in the Beat before passing it to us.

andrewkroh:
Based on discussion with urso about the stack trace from the race detector, `ConvertToGenericEvent` will continue to return a copy. So this PR is read to be squashed and merged (assuming there are no new comments).

tsg:
Merged. Great work @andrewkroh, I'm really glad we'll have this in beta1.

PR #2391, opened 2016-08-26
Regenerate OpenSSL certs used by the LS tests

tsg:
They might have been generated with an openssl version that produces
private keys in a format that is not supported by Netty 4. This
should fix the integration tests.


ruflin:
LGTM

PR #2392, opened 2016-08-26
Merge defaults for publisher queue_size and bulk_queue_size

ruflin:
There was a default in packetbeat and libbeat publisher for the default queueSize. The two are now merged and the default from libbeat was taken.

* Add bulk_queue_size to config as it was missing
* Rely on constants from libbeat

### DIFF
--- a/libbeat/publisher/publish.go
+++ b/libbeat/publisher/publish.go
@@ -101,8 +101,8 @@ type Topology struct {
 }
 
 const (
-	defaultChanSize     = 1000
-	defaultBulkChanSize = 0
+	DefaultQueueSize     = 1000
+	DefaultBulkQueueSize = 0
 )
 
 func init() {
@@ -208,15 +208,7 @@ func (publisher *BeatPublisher) init(
 		logp.Info("Dry run mode. All output types except the file based one are disabled.")
 	}
 
-	hwm := defaultChanSize
-	if shipper.QueueSize != nil && *shipper.QueueSize > 0 {
-		hwm = *shipper.QueueSize
-	}
-
-	bulkHWM := defaultBulkChanSize
-	if shipper.BulkQueueSize != nil && *shipper.BulkQueueSize >= 0 {
-		bulkHWM = *shipper.BulkQueueSize
-	}
+	shipper.InitShipperConfig()
 
 	publisher.geoLite = common.LoadGeoIPData(shipper.Geoip)
 
@@ -242,8 +234,8 @@ func (publisher *BeatPublisher) init(
 					config,
 					output,
 					&publisher.wsOutput,
-					hwm,
-					bulkHWM))
+					*shipper.QueueSize,
+					*shipper.BulkQueueSize))
 
 			if ok, _ := config.Bool("save_topology", 0); !ok {
 				continue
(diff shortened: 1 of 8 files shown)


packetbeat/beater/packetbeat.go
```diff
@@ -49,11 +49,6 @@ type CmdLineArgs struct {
 
 var cmdLineArgs CmdLineArgs
 
-const (
-	defaultQueueSize     = 2048
```
ruflin: @urso do you see an issue that this now also set to 1000?

packetbeat/beater/packetbeat.go
```diff
@@ -103,15 +98,7 @@ func (pb *Packetbeat) init(b *beat.Beat) error {
...
-	queueSize := defaultQueueSize
-	if b.Config.Shipper.QueueSize != nil {
-		queueSize = *b.Config.Shipper.QueueSize
-	}
-	bulkQueueSize := defaultBulkQueueSize
-	if b.Config.Shipper.BulkQueueSize != nil {
-		bulkQueueSize = *b.Config.Shipper.BulkQueueSize
-	}
-	pb.Pub, err = publish.NewPublisher(b.Publisher, queueSize, bulkQueueSize)
+	pb.Pub, err = publish.NewPublisher(b.Publisher, *b.Config.Shipper.QueueSize, *b.Config.Shipper.BulkQueueSize)
```
ruflin: This currently does not work as the publisher configs are only loaded after the packetbeat init function.

libbeat/publisher/publish.go
```diff
@@ -215,14 +215,13 @@ func (publisher *BeatPublisher) init(
 		logp.Info("Dry run mode. All output types except the file based one are disabled.")
 	}
 
-	hwm := defaultChanSize
-	if shipper.QueueSize != nil && *shipper.QueueSize > 0 {
-		hwm = *shipper.QueueSize
+	// TODO: Replace this by ucfg
+	if shipper.QueueSize == nil || *shipper.QueueSize <= 0 {
+		*shipper.QueueSize = DefaultQueueSize
```
urso: if shipper.QueueSize == nil, this will fail. No?

libbeat/publisher/publish.go
```diff
@@ -215,14 +215,13 @@ func (publisher *BeatPublisher) init(
...
 	}
 
-	bulkHWM := defaultBulkChanSize
-	if shipper.BulkQueueSize != nil && *shipper.BulkQueueSize >= 0 {
-		bulkHWM = *shipper.BulkQueueSize
+	if shipper.BulkQueueSize == nil || *shipper.BulkQueueSize < 0 {
+		*shipper.BulkQueueSize = DefaultBulkQueueSize
```
urso: if BulkQueueSize == nil, this will also fail

packetbeat/beater/packetbeat.go
urso: what if pointers are nil? Better have some defaults in Shipper.

packetbeat/beater/packetbeat.go
urso: as queue sizes are configurable and both depend on shipper config it should be ok.

ruflin: For packetbeat this is kind of a BC break but as it is only the internal default and we don't recommend to change this value anyway, worth mentioning?

urso: no need to mention internal default value. This only reduces buffering a little.

PR #2393, opened 2016-08-26
Move ignore_outgoing to packetbeat

ruflin:
* Update documentation
* Update CHANGELOG
* Remove option from test config files as not needed


tsg:
Can you have a look if this could be added to the migrate script? I think it should at least be possible to issue a warning that asks the user to move the option manually.

ruflin:
@tsg I added a line to the migration script.

PR #2401, opened 2016-08-28
Link to article outlining CA creation

mpchadwick:


ruflin:
@dedemorton Can you have a look at this?

libbeat/docs/shared-tls-logstash-config.asciidoc
```diff
@@ -17,7 +17,7 @@ To use TLS mutual authentication:
 
 . Create a certificate authority (CA) and use it to sign the certificates that you plan to use for
 {beatname_uc} and Logstash. Creating a correct SSL/TLS infrastructure is outside the scope of this
-document. There are many online resources available that describe how to create certificates.
+document. There are many online resources available that describe how to create certificates including https://www.elastic.co/guide/en/shield/2.3/certificate-authority.html[this section of the documentation for Shield].
```
ruflin: We should use here if possible an internal reference instead of a link to make sure it always links to the same version. @dedemorton Is that possible across projects?

dedemorton: @ruflin Sure, we can use variables defined in the index.asciidoc files just like we do for other guides. If you want to merge this, I can add the variables.

dedemorton:
LGTM

PR #2410, opened 2016-08-29
Clarify how to set shipper name (cherry-pick into 5.0)

dedemorton:


filebeat/docs/fields.asciidoc
```diff
@@ -26,7 +26,7 @@ Contains common beat fields available in all event types.
 [float]
 === beat.name
 
-The name of the Beat sending the log messages. If the shipper name is set in the configuration file, then that value is used. If it is not set, the hostname is used.
+The name of the Beat sending the log messages. If the shipper name is set in the configuration file, then that value is used. If it is not set, the hostname is used. To set the shipper name, use the `name` option in the configuration file.
```
ruflin: We always refer here to shipper name but in the config and here it is only `beat.name` or ` name`. Not sure if this could be confusing?

dedemorton: @ruflin Drats, I think you're right. I meant shipper name to be descriptive, but saying Beat name is probably less confusing.

PR #2417, opened 2016-08-30
Introduce harvester_limit to limit number of harvesters

### LINKED ISSUE #2236: Limit number of open files
Having a large amount of files which need to be harvested by filebeat it can happen that the systems max number of open file handlers is exceeded. In most cases the problem can be solved by closing files faster or not harvesting some files. Still there are edge cases where this can be an issue.

This issue should help to track progress on this topic and discuss potential implementations.

ruflin:
This limits the number of harvesters that are started per prospector.

* Changelog entry added
* Docs updated
* Config file updated

Closes #2236

### DIFF
--- a/filebeat/prospector/prospector.go
+++ b/filebeat/prospector/prospector.go
@@ -5,6 +5,8 @@ import (
 	"sync"
 	"time"
 
+	"sync/atomic"
+
 	cfg "github.com/elastic/beats/filebeat/config"
 	"github.com/elastic/beats/filebeat/harvester"
 	"github.com/elastic/beats/filebeat/input"
@@ -14,14 +16,15 @@ import (
 )
 
 type Prospector struct {
-	cfg           *common.Config // Raw config
-	config        prospectorConfig
-	prospectorer  Prospectorer
-	spoolerChan   chan *input.Event
-	harvesterChan chan *input.Event
-	done          chan struct{}
-	states        *file.States
-	wg            sync.WaitGroup
+	cfg              *common.Config // Raw config
+	config           prospectorConfig
+	prospectorer     Prospectorer
+	spoolerChan      chan *input.Event
+	harvesterChan    chan *input.Event
+	done             chan struct{}
+	states           *file.States
+	wg               sync.WaitGroup
+	harvesterCounter uint64
 }
 
 type Prospectorer interface {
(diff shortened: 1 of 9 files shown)


filebeat/prospector/prospector.go
```diff
@@ -155,6 +158,13 @@ func (p *Prospector) createHarvester(state file.State) (*harvester.Harvester, er
 }
 
 func (p *Prospector) startHarvester(state file.State, offset int64) error {
+
+	if p.config.HarvesterLimit > 0 && atomic.LoadUint64(&p.harvesterCounter) >= p.config.HarvesterLimit {
+		return fmt.Errorf("Harvester limit reached.")
+	}
+
+	atomic.AddUint64(&p.harvesterCounter, 1)
```
urso: if createHarvester fails counter is off. Either move counter right before p.wg.Add(1) or decrement counter on fail (+ explain why it must be decremented and can not be moved past this place).

tsg: I think there is a race here if `startHarverster` can be called concurrently: two threads check the condition when the counter is equal to the limit-1, then they both increment it, so you get to limit+1. Not a big issue, but wouldn't things be simpler with a mutex?

ruflin: @urso @tsg Agree to both comments. Fixing it.

urso: right. Unfortunately `atomic` does not provide an api to directly get the incremented counter value. Using a mutex for incrementing would be simpler. Otherwise one would need to use CompareAndSwap to atomically update and check counter still being valid.

ruflin: @tsg Just checked the code again and startHarvester cannot be called concurrently. The atomic operation is needed because of the decrementing of the counter inside the go routine.

ruflin: @tsg but you are right that the issue with +1 harvester still persists. :-(

filebeat/prospector/prospector.go
```diff
@@ -155,6 +158,11 @@ func (p *Prospector) createHarvester(state file.State) (*harvester.Harvester, er
 }
 
 func (p *Prospector) startHarvester(state file.State, offset int64) error {
+
+	if p.config.HarvesterLimit > 0 && atomic.LoadUint64(&p.harvesterCounter) >= p.config.HarvesterLimit {
```
urso: we should add a comment about atomic ops being safe, due to startHarvester not being executed concurrently and atomic ops only required for harvesters shutting down.

tsg:
LGTM.

filebeat/prospector/prospector.go
ruflin: Comment added in separate commit. Can be squashed.

shaiknaimuddin:
Hi,
iam running elasticsearch5.3 and kibana5.3
when i trying yo run filebeats and unable to start the service

i tried with the command   PS C:\Program Files\Filebeat> Start-Service filebeat  but returning error as
Start-Service : Cannot find any service with service name 'filebeat'.

filebeat.yml
---------
filebeat.prospectors:

- input_type: log

  # Paths that should be crawled and fetched. Glob based paths.
  paths:
    - /var/log/*.log
    #- C:\Program Files\soalogs.txt

output.elasticsearch:
  # Array of hosts to connect to.
  hosts: ["myipaddress:9200"]

please help  me in starting Beats.

ruflin:
@shaiknaimuddin For questions please use discuss: https://discuss.elastic.co/c/beats/filebeat

PR #2423, opened 2016-08-31
Add harvesters.skipped stats

ruflin:
Add statistic for number of harvester skipped because of harvester_limit to filebeat and metricbeat


tsg:
LGTM.

PR #2424, opened 2016-08-31
Deprecate Topology Map

ruflin:
The current implementation of Topology map will be deprecated as it is focused on Packetbeat. It is intented to potentially readd something similar to topology map in the future which applies to all beats.

### DIFF
--- a/filebeat/filebeat.full.yml
+++ b/filebeat/filebeat.full.yml
@@ -246,11 +246,11 @@ filebeat.prospectors:
 # sub-dictionary. Default is false.
 #fields_under_root: false
 
-# How often (in seconds) shippers are publishing their IPs to the topology map.
+# DEPRECATED: How often (in seconds) shippers are publishing their IPs to the topology map.
 # The default is 10 seconds.
 #refresh_topology_freq: 10
 
-# Expiration time (in seconds) of the IPs published by a shipper to the topology map.
+# DEPRECATED: Expiration time (in seconds) of the IPs published by a shipper to the topology map.
 # All the IPs will be deleted afterwards. Note, that the value must be higher than
 # refresh_topology_freq. This setting is used only by the Redis output. The other
 # outputs don't support expiring entries.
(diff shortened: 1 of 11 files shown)


filebeat/filebeat.full.yml
```diff
@@ -246,11 +246,11 @@ filebeat.prospectors:
 # sub-dictionary. Default is false.
 #fields_under_root: false
 
-# How often (in seconds) shippers are publishing their IPs to the topology map.
+# DEPRECATED: How often (in seconds) shippers are publishing their IPs to the topology map.
```
tsg: I didn't put other deprecated options in the full configs. I think we can remove them completely.

libbeat/docs/generalconfig.asciidoc
```diff
@@ -121,13 +99,13 @@ fields:
 
 ===== refresh_topology_freq
 
-The refresh interval of the topology map in
+DEPRECATED: The refresh interval of the topology map in
```
tsg: We have special asciidoc syntax for deprecations, which makes it more likely that we'll remember to cleanup the docs after removal. See [here](https://github.com/elastic/docs#section-notifications).

ruflin: What version would you suggest to put there? deprecated[5.0.0] or it is [5.0]?

packetbeat/docs/maintaining-topology.asciidoc
```diff
@@ -1,14 +1,16 @@
 [[maintaining-topology]]
 == Maintaining the Real-Time State of the Network Topology
 
+DEPRECATED
```
tsg: Also here it should use the `deprecated[]` syntax.

filebeat/filebeat.full.yml
ruflin: Good idea.

ruflin:
@tsg Reviews should be addressed in the new commit. Commits can be squashed.

Issue #2418, opened 2016-08-30
Import the dashboards from the selected Beat


monicasarbu:
If the user uses the `import_dashboards` from the Elastic Beat package to import the sample dashboards from the url, then by default he/she doesn't need to specify the Beat, as it should be the current Beat.

For example, if the user downloads the Metricbeat package, then in order to import the index pattern together with the dashboards, he is currently running:
```
./import_dashboards -beat metricbeat 
```
where the default url depends on the Beat's version.

should become:

```
./import_dashboards 
```

/cc @tsg

PR #2429, opened 2016-09-01
Simplify the usage of import_dashboards from the Beat package

monicasarbu:
Trying to solve https://github.com/elastic/beats/issues/2418

**Running import_dashboards from the Beat package**

The `import_dashboards` script is shipped in the Beat package, so by default the script should import the dashboards for that Beat. The command to do that is:

```
./import_dashboards
```
instead of:

```
./import_dashboards -beat metricbeat
```
This requires that the zip archive with the dashboards to contain a directory for each Beat, even if the archive contains dashboards for a single Beat.

The `import_dashboards` needs to be compiled with the BEAT flag set to the Beat name:
```
$ cd libbeat/dashboards
$ BEAT=packetbeat make
```
**Running import_dashboards from sources**

There are two options on how you can import the dashboards:
- from a local directory (`etc/kibana` of the Beat repository)
- from a zip archive and then you can:
     - import dashboards of a single Beat (you need to pass the `-beat` option)
     - import dashboards of all Beats


libbeat/dashboards/import_dashboards.go
```diff
@@ -105,6 +107,9 @@ func (cl *CommandLine) Read() error {
 	if cl.opt.Url == "" && cl.opt.File == "" && cl.opt.Dir == "" {
 		return errors.New("ERROR: Missing input. Please specify one of the options -file, -url or -dir")
 	}
+	if cl.opt.Beat == "" {
+		cl.opt.Beat = beat
+	}
```
tsg: I think it would be better to set it to `beat` only if the option is not used. This way, even when using the version from a Beat package, one can force importing all dashbaords with `-beat ""`.

PR #2430, opened 2016-09-01
Update float, half_float and scaled_float mappings

### LINKED ISSUE #2238: Review mapping templates of beats
I was looking at templates after the switch to scaled/half floats and found a couple potential issues, maybe we should do a review of these templates before the 5.0 release. For instance:
 - `apache.status.total_accesses` and `apache.status.total_kbytes` are currently an integers but a high traffic instance could overflow them I think? Should it be a `long`?
 - `apache.status.requests_per_sec` is a half float, which means that the integer value will become fuzzy after 2048, and rounded to +Infinity after 65504, which are values that could be reached on a high-traffic server that serves small static files. Maybe they should be integers instead? or scalled floats if we really want to be able to have some decimal places.
 
There are also some inconsistencies, for instance most CPU usage stats seem to have moved from half floats to scaled floats, but the cpu stats of redis are still using half floats? Is it intentional? Similarly, the system load averages are using scaled floats while the apache load averages are using half floats.

ruflin:
* The mappings for float, half_float and scaled_float were reviewed and adjusted. All are now scaled_float
* All integers were converted to long as this was already happening automatically in the script
* Json examples updated

Closes #2238

### DIFF
--- a/metricbeat/docs/fields.asciidoc
+++ b/metricbeat/docs/fields.asciidoc
@@ -57,7 +57,7 @@ Apache hostname.
 [float]
 === apache.status.total_accesses
 
-type: integer
+type: long
 
 Total number of access requests.
 
@@ -65,7 +65,7 @@ Total number of access requests.
 [float]
 === apache.status.total_kbytes
 
-type: integer
+type: long
 
 Total number of kilobytes served.
 
@@ -73,7 +73,7 @@ Total number of kilobytes served.
 [float]
 === apache.status.requests_per_sec
 
-type: half_float
+type: scaled_float
 
 Requests per second.
 
@@ -81,7 +81,7 @@ Requests per second.
 [float]
 === apache.status.bytes_per_sec
 
-type: float
+type: scaled_float
 
 Bytes per second.
 
@@ -89,7 +89,7 @@ Bytes per second.
 [float]
 === apache.status.bytes_per_request
 
-type: half_float
+type: scaled_float
 
 Bytes per request.
 
@@ -97,7 +97,7 @@ Bytes per request.
 [float]
 === apache.status.workers.busy
 
-type: integer
+type: long
 
 Number of busy workers.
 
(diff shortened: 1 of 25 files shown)


metricbeat/module/apache/status/_meta/data.json
```diff
@@ -15,17 +15,17 @@
...
             "load": {
-                "1": 1.53,
-                "15": 0.53,
-                "5": 0.92
+                "1": 0.31,
+                "15": 0.19,
+                "5": 0.2
             },
-            "requests_per_sec": 0.0119197,
+            "requests_per_sec": 0.00331858,
```
ruflin: This will be rounded to 3 digits in elasticsearch which I think is fine. @tsg WDYT?

tsg: Yeah, I think that's fine.

metricbeat/docs/fields.asciidoc
```diff
@@ -1687,15 +1687,15 @@ Number of requested checkpoints that have been performed.
 [float]
 === postgresql.bgwriter.checkpoints.times.write.ms
 
-type: float
+type: scaled_float
```
tsg: I'd leave these as floats because they are the total (not the derivative) number of milliseconds, so if you have an instance running for a year these numbers can be huge. Since float is what Postgres uses to report them, I'd say it's safer for us to also use floats to avoid overflows.

metricbeat/docs/fields.asciidoc
```diff
@@ -1687,15 +1687,15 @@ Number of requested checkpoints that have been performed.
...
 
 Total amount of time that has been spent in the portion of checkpoint processing where files are written to disk, in milliseconds.
 
 
 [float]
 === postgresql.bgwriter.checkpoints.times.sync.ms
 
-type: float
+type: scaled_float
```
tsg: Same as above.

tsg:
Left two comments, otherwise LGTM.

ruflin:
@tsg new version pushed with the adjustments to postgres.

PR #2432, opened 2016-09-01
Minor naming cleanup in reader

ruflin:
No functional changes

### DIFF
--- a/filebeat/harvester/log_test.go
+++ b/filebeat/harvester/log_test.go
@@ -12,7 +12,9 @@ import (
 	"time"
 
 	"github.com/elastic/beats/filebeat/harvester/encoding"
+	"github.com/elastic/beats/filebeat/harvester/reader"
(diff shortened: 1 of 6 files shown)


filebeat/harvester/log_test.go
```diff
@@ -127,3 +129,17 @@ func TestInitRegexp(t *testing.T) {
 	_, err := InitRegexps([]string{"((((("})
 	assert.NotNil(t, err)
 }
+
+// readLine reads a full line into buffer and returns it.
+// In case of partial lines, readLine does return and error and en empty string
```
monicasarbu: @tsg I just noticed that are two mistakes in here "does return an error and an empty string".

ruflin: Will fix it as part of one of the upcoming PR's. Thanks for spotting.

PR #2434, opened 2016-09-01
Fix corrupt Winlogbeat registry occurring on power failure

andrewkroh:
File writes are cached and not immediately flushed to disk. So if there is a power loss prior to flushing it seems that some corruption occurs. This adds the FILE_FLAG_WRITE_THROUGH flag to the CreateFile call to ensure that writes go directly to the disk on Windows.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -57,6 +57,7 @@ https://github.com/elastic/beats/compare/v5.0.0-alpha5...master[Check the HEAD d
 - Fix processor failure in Filebeat when using regex, contain, or equals with the message field. {issue}2178[2178]
 
 *Winlogbeat*
+- Fix corrupt registry file that occurs on power loss by disabling file write caching. {issue}2313[2313]
 
 ==== Added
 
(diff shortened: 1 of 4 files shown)


andrewkroh:
jenkins, test it

ruflin:
@andrewkroh I assume the same could happen in filebeat? Should we apply the same patch there (or finally extract the registry writing part)?

ruflin:
Oh, my comment was fire and forget it seems :-( Perhaps a good time to "merge" the file writing? @urso @ph ?

ph:
@ruflin
 What do you actually mean by:
`"merge" the file writing`

1. Port this PR to filebeat
2. Merge @urso optimization to the registry?

ruflin:
The part changed in this PR is mainly about writing files. Short term winlogbeat and filebeat should both use this method to write files. Long term it would be nice to have the complete registry standardised.

To have a quick fix (which I think we need) we can to a PR against Filebeat with this and generalised it later.

ph:
@ruflin ++ agreed.

PR #2436, opened 2016-09-01
Fix diskio sample event in docs

andrewkroh:
Reported here: https://discuss.elastic.co/t/diskio-documentation-update/59601


ruflin:
This one is missing the json generator. Reason is that it doesn't work on Mac OS X and the generation is a manual step :-(

PR #2438, opened 2016-09-02
[WIP] Update filebeat to use matcher for multiline and include/exclude lines

urso:


ruflin:
@urso Seeing that we use `matcher` in parts of the code, I wonder if this change happened in an other PR or at least parts of it?

urso:
Hmm... matcher is already used.

PR #2442, opened 2016-09-02

chrousto:


ruflin:
@chrousto Can you share some details on the intention of this change?

chrousto:
@ruflin Yes Nicolas, actually I can :) I am not familiar with github yet and I have submitted this too soon and I am really sorry about it.
The initial purpose of my proposition was to modifiy the doc for the filebeat multitline options in order to add some little vizualization thumbnails of what group of lines where kept together for each negate and match values.
Attached are the images I wanted to upload and then insert inside the asciidocs. I did not manage to upload them and had to attend to something else. This is when I accidentaly submitted my incomplete changes.

Thanks.
Christophe.


negate_false_match_after
negate_false_match_before
negate_true_match_after
negate_true_match_before

dedemorton:
@chrousto I see what you're trying to accomplish here, but a visual in the last cell of the table will not render well in the built doc (users will probably need to scroll right to see the visuals, and the fact that they don't wrap might cause other display problems, like the table content bleeding into the TOC). 

I added the table because I thought the combination of multiline options was a bit confusing, but I agree that it takes some effort to read through the explanations. I like where you're going (giving readers a way to process the info more quickly).

Instead of adding visuals, maybe we need to add some simple examples to get the point across. WDYT?

chrousto:
@dedemorton Yup, completely agree with the simple example way of helping the reader visualizing.

dedemorton:
Closing this PR as it will be replaced with text examples rather than images. See https://github.com/elastic/beats/issues/2774.

PR #2445, opened 2016-09-02
Add beat version number to elasticsearch template

ruflin:
This can help during debugging which template is loaded in elasticsearch. Also in case a template must be overwritten in the future for upgrading, this could be done based on the version.

### DIFF
--- a/dev-tools/set_version
+++ b/dev-tools/set_version
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 import os
 import argparse
+from subprocess import check_call
 
 template_go = '''package beat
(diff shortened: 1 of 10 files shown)


andrewkroh:
LGTM

tsg:
Should we make `dev-tools/set-version` call `make update` to make sure we don't forget?

ruflin:
@tsg I like the idea, as especially if others are running it that don't know about `make update` it will directly happen. Also we must do a manual commit after updating the version anyways so this becomes handy. Changed accordingly.

PR #2447, opened 2016-09-02
Added logic to parse prepare id from Prepare response

vik5jagan:
Prepare statement response from Cassandra contains prepareId of type UUID. Currently this is parsed as string. Fix has been added to convert UUID to string.


karmi:
Hi @vik5jagan, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/beats/pull/2447.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

PR #2448, opened 2016-09-02
Parse PrepareId from Prepare statement Response

vik5jagan:
Cassandra response for prepare statement contains prepareid of type UUID. Currently this is parsed as string. Added code fix to parse UUID and convert the same to string

### DIFF
--- a/packetbeat/protos/cassandra/internal/gocql/frame.go
+++ b/packetbeat/protos/cassandra/internal/gocql/frame.go
@@ -443,7 +443,14 @@ func (f *Framer) parseResultPrepared() map[string]interface{} {
 
 	result := make(map[string]interface{})
 
-	result["prepared_id"] = string((f.decoder).ReadShortBytes())
+        uuid,err := UUIDFromBytes((f.decoder).ReadShortBytes())
+
+	if(err != nil){
+		logp.Err("Error in parsing UUID")
+	}
+	
+	result["prepared_id"] =  uuid.String()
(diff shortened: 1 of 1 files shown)


vik5jagan:
Please let me know if anything needs to be fixed for AppVeyor Build failure.

packetbeat/protos/cassandra/internal/gocql/frame.go
```diff
@@ -443,7 +443,14 @@ func (f *Framer) parseResultPrepared() map[string]interface{} {
 
 	result := make(map[string]interface{})
 
-	result["prepared_id"] = string((f.decoder).ReadShortBytes())
+	uuid, err := UUIDFromBytes((f.decoder).ReadShortBytes())
```
medcl: I am a little confused, the uuid need 16 bytes, but the prepare_Id is only two bytes here

vik5jagan: UUID is 16 bytes long. Here we are reading the short bytes. Internally ReadShortBytes() will have A [short] n, followed by n bytes if n >= 0. In this case, n will be 16 bytes.

You have mentioned, prepare_id is only two bytes. Can you please add some more details on that?

medcl: @vik5jagan sorry, you are right, it is [short bytes] not [short]
LGTM

Issue #2281, opened 2016-08-16
runoptions is only a Packetbeat option


andrewkroh:
`runoptions` is only supported by Packetbeat on Linux so it should only appear in the Packetbeat docs and there should probably be a note about it only working on linux.

https://www.elastic.co/guide/en/beats/filebeat/current/configuration-run-options.html

ruflin:
The runoptions should probably also be moved under the packetbeat namespace (if possible)

andrewkroh:
Good catch, the example should be:

```
packetbeat:
  runoptions:
    uid=501
    gid=501
```

ruflin:
Closing as #2439 is merged.

PR #2450, opened 2016-09-02
Remove runoptions from all guides except Packetbeat (cherry-pick into 1.2)

dedemorton:
Figured it might be a good idea to cherry-pick this one back into 1.2.


packetbeat/docs/reference/configuration/runconfig.asciidoc
```diff
@@ -0,0 +1,24 @@
...
+
+WARNING: On Linux, Setuid doesn't change the uid of all threads, so the Go
+         garbage collector will continue to run as root. Also note that process
+         monitoring only works when running as root.
+
+Example configuration for the `runoptions` section of the +{beatname_lc}.yml+ config file:
+
+[source,yaml]
+------------------------------------------------------------------------------
+packetbeat:
```
ruflin: Under the 1.x release, run option are not under the packbeat namespace yet. Did we push this also the the 1.3 branch?

dedemorton:
@ruflin I changed the example in the 1.2 docs to remove the `packetbeat` namespace from the `runoptions` example. Do I need to do this for 1.3, too?

I thought that the example in 1.3 needed to change because #2281 pointed to the docs for 1.3. If that's not the case, let me know, and I'll open a PR.

ruflin:
@dedemorton runoptions was moved under the packetbeat namespace in 5.0. In 1.x release it is a global option but still should only be in the packetbeat docs. So the same applies to 1.3.

PR #2455, opened 2016-09-05
Fix async publisher sending empty events

ruflin:


filebeat/publish/publish.go
```diff
@@ -172,19 +172,28 @@ func (p *asyncLogPublisher) Start() {
...
 			case events := <-p.in:
-				pubEvents := make([]common.MapStr, len(events))
-				for i, event := range events {
+
+				tmpEvents := make([]*input.Event, 0)
+				for _, event := range events {
+					if event.HasData() {
+						tmpEvents = append(tmpEvents, event)
+					}
+				}
```
urso: we need to copy here? the events array is store in `eventsBatch` for pushing acked events(batches) to registrar. Interestingly the syncPublisher does forward all events without filtering.

ruflin: No need to copy. I was thinking that eventsBatch also needs the filtered list, but if it is for the registrar this is not the case. 

I simplified the implementation. Please have a look again.

The sync publisher also does the filtering on line 114.

filebeat/publish/publish.go
```diff
@@ -172,9 +172,12 @@ func (p *asyncLogPublisher) Start() {
 			case <-p.done:
 				return
 			case events := <-p.in:
-				pubEvents := make([]common.MapStr, len(events))
-				for i, event := range events {
-					pubEvents[i] = event.ToMapStr()
+
+				pubEvents := make([]common.MapStr, 0)
```
urso: `pubEvents := make([]common.MapStr, 0, len(events))`. See line 111. Consider helper function copying events and reuse in sync and async publisher modes.

ruflin: Changed. For the cleanup: I plan a second PR after this one that cleans up the publisher. But I didn't do it in this one as this one should be backported and I want to keep the changes as minimal as possible.

Issue #880, opened 2016-01-28
Run Filebeat once for predefined files


monicasarbu:
Similar with what we have for Packetbeat, I suggest to add an option to filebeat to pass a file with logs or a `gz` archive with logs.

For example in Packetbeat we have the `-I` option where you can pass a `pcap` file and Packetbeat is replying the traffic from that particular pcap. For Filebeat, instead of passing the `pcap` file, we can pass a `gz` archive, for example. It's up to us if we decide to copy the archive in a directory from where Filebeat is fetching the logs or we can just read that file and skip the file discovery part.

I think this would be really useful for purposes like: testing, development and even trainings.

marnikvde:
+1
I could use this as well

ruflin:
I think this is indirectly already possible through the stdin. The problem is that after processing Filebeat does not stop afterwards. To make it more obvious, I will change the title of the issue.

For the gz part this is related to https://github.com/elastic/beats/issues/637

strootman:
I could use this as well. 
I cannot rely on our network, so I need to re-ship files after nightly log rotation to ensure that the daily index is complete.

ruflin:
@strootman Filebeat is able to track file rotation and with LS you can process the log line with a timestamp inside (hopefully) which then can send the log lines to the correct index.

strootman:
@ruflin Thank you. Under normal circumstances, there is nothing wrong with what you describe. 

However, when just getting started and going through the process of creating, modifying, and refining the index mappings for my logs, I have got myself into a situation where it is just easier to blow away all my indices, and re-ship all of my previous data.

There is no problem with the normal setup of filebeat shipping my logs, but I have to go through a few manual steps to perform a one-shot task of re-shipping *ALL* of my previous logs to correct an invalid mapping. 

We use ansible to automate many of our infrastructure tasks, and having the ability to detect when filebeat is done shipping a set of files (and producing a return code of `0` or `1`) makes automating the task of `reship all these logs, because I am just starting, and didn't realize how quickly I could bring my cluster to its knees because I introduced an invalid mapping(s)` a lot easier.

Without this feature, I have to baby-sit these tasks, using a combination of `marvel`, index stats, visualvm monitoring of logstash, and inspection of the filebeat registry file to _ensure_ that everything has been re-shipped.

Not impossible, just highly inconvenient. _Especially_ since I have to do this for;
* each of our dev environments, 
* each of our QA environments, 
* and (eventually) production.

[...]

PR #2456, opened 2016-09-05
Allow filebeat to only run once

ruflin:
When the `-once` flag is used, filebeat starts all configured harvesters and prospectors and runs each prospector until the harvesters are closed. It is recommended to use the flag in combination with `close_eof` so harvester directly close when the end of the file is reached. By default harvesters are closed after `close_inactive`.

Closes https://github.com/elastic/beats/issues/880

### DIFF
--- a/filebeat/beater/filebeat.go
+++ b/filebeat/beater/filebeat.go
@@ -155,6 +155,8 @@ func (fb *Filebeat) Run(b *beat.Beat) error {
 			logp.Info("Shutdown output timer started. Waiting for max %v.", timeout)
 			waitEvents.Add(withLog(waitDuration(timeout),
 				"Continue shutdown: Time out waiting for events being published."))
+		} else {
+			waitEvents.AddChan(fb.done)
 		}
 	}
 
--- a/filebeat/beater/filebeat.go
+++ b/filebeat/beater/filebeat.go
@@ -1,6 +1,7 @@
 package beater
 
 import (
+	"flag"
 	"fmt"
 	"sync"
 
@@ -15,11 +16,12 @@ import (
 	"github.com/elastic/beats/filebeat/spooler"
 )
 
+var once = flag.Bool("once", false, "Run filebeat only once until all harvesters reach EOF")
+
 // Filebeat is a beater object. Contains all objects needed to run the beat
 type Filebeat struct {
-	config  *cfg.Config
-	sigWait *signalWait
-	done    chan struct{}
+	config *cfg.Config
+	done   chan struct{}
 }
 
 // New creates a new Filebeat pointer instance.
@@ -33,9 +35,8 @@ func New(b *beat.Beat, rawConfig *common.Config) (beat.Beater, error) {
 	}
 
 	fb := &Filebeat{
-		done:    make(chan struct{}),
-		sigWait: newSignalWait(),
-		config:  &config,
+		done:   make(chan struct{}),
+		config: &config,
 	}
 	return fb, nil
 }
@@ -45,13 +46,12 @@ func (fb *Filebeat) Run(b *beat.Beat) error {
 	var err error
 	config := fb.config
 
-	var wgEvents *sync.WaitGroup // count active events for waiting on shutdown
-	var finishedLogger publisher.SuccessLogger
+	waitFinished := newSignalWait()
+	waitEvents := newSignalWait()
 
-	if fb.config.ShutdownTimeout > 0 {
-		wgEvents = &sync.WaitGroup{}
-		finishedLogger = newFinishedLogger(wgEvents)
-	}
+	// count active events for waiting on shutdown
+	wgEvents := &sync.WaitGroup{}
+	finishedLogger := newFinishedLogger(wgEvents)
 
 	// Setup registrar to persist state
 	registrar, err := registrar.New(config.RegistryFile, finishedLogger)
@@ -60,13 +60,14 @@ func (fb *Filebeat) Run(b *beat.Beat) error {
 		return err
 	}
 
-	// Channel from harvesters to spooler
-	successLogger := newRegistrarLogger(registrar)
+	// Make sure all events that were published in
+	registrarChannel := newRegistrarLogger(registrar)
+
+	// Channel from spooler to harvester
 	publisherChan := newPublisherChannel()
 
 	// Publishes event to output
-	publisher := publisher.New(config.PublishAsync,
-		publisherChan.ch, successLogger, b.Publisher)
+	publisher := publisher.New(config.PublishAsync, publisherChan.ch, registrarChannel, b.Publisher)
 
 	// Init and Start spooler: Harvesters dump events into the spooler.
 	spooler, err := spooler.New(config, publisherChan)
@@ -75,9 +76,7 @@ func (fb *Filebeat) Run(b *beat.Beat) error {
 		return err
 	}
 
-	crawler, err := crawler.New(
-		newSpoolerOutlet(fb.done, spooler, wgEvents),
-		config.Prospectors)
+	crawler, err := crawler.New(newSpoolerOutlet(fb.done, spooler, wgEvents), config.Prospectors)
 	if err != nil {
 		logp.Err("Could not init crawler: %v", err)
 		return err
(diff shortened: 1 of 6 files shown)


filebeat/beater/filebeat.go
```diff
@@ -92,15 +99,23 @@ func (fb *Filebeat) Run(b *beat.Beat) error {
...
 	}
 	// Stop crawler -> stop prospectors -> stop harvesters
 	defer crawler.Stop()
 
-	// Blocks progressing. As soon as channel is closed, all defer statements come into play
-	<-fb.done
+	if *once {
+		logp.Debug("filebeat", "Running filebeat once. Waiting for completion ...")
+		spooler.Flush()
+		publisher.Publish()
```
ruflin: @urso What I'm trying to do here is to make sure that all events which are in the queue are sent. Is there a better way to wait / check the publisher if all events were sent?

filebeat/beater/filebeat.go
```diff
@@ -14,10 +16,17 @@ import (
 	"github.com/elastic/beats/filebeat/spooler"
 )
 
+var once *bool
+
+func init() {
+	once = flag.Bool("once", false, "Run filebeat only once until all harvesters reach EOF")
```
urso: can be initialized directly at line 19.

filebeat/prospector/prospector.go
```diff
@@ -28,6 +28,7 @@ type Prospector struct {
 	done             chan struct{}
 	states           *file.States
 	wg               sync.WaitGroup
+	channelWg        sync.WaitGroup // Separate waitgroup for channels as not stopped on completion
```
urso: I don't understand the necessary for channelWg. why not reuse wg?

filebeat/spooler/spooler.go
```diff
@@ -116,17 +117,23 @@ func (s *Spooler) queue(event *input.Event) bool {
...
 // flush flushes all events to the publisher.
-func (s *Spooler) flush() {
-	if len(s.spool) > 0 {
+func (s *Spooler) Flush() int {
+
+	// Mutex need as also written from goroutine on shutdown
+	s.flushMutex.Lock()
+	defer s.flushMutex.Unlock()
+
+	count := len(s.spool)
```
urso: `
if count == 0 {
    return 0
}
`

filebeat/prospector/prospector.go
ruflin: channelWg is needed, because when only run once, the prospector does the wait() for the first run to be completed. When the prospector Run returns, channel are not shutdown yet and shouldn't because events are still processed. This only happens when prospector is closed with done.

I see that a similar pattern could apply here that we did with the other channels, but we can do this in a follow up PR.

filebeat/beater/filebeat.go
ruflin: changed

filebeat/beater/filebeat.go
```diff
@@ -96,23 +105,47 @@ func (fb *Filebeat) Run(b *beat.Beat) error {
...
-	// Blocks progressing. As soon as channel is closed, all defer statements come into play
+	// Wait for completion if running once.
+	if *once {
+		go func() {
+			logp.Info("Running filebeat once. Waiting for completion ...")
+			crawler.WaitForCompletion()
+
+			// Flushes spooler as often as needed to send all events to publisher
+			for spooler.Flush() > 0 {
+			}
```
urso: we should a a `defer Flush` to the spooler. The spooler was changed to drain the receive channel until empty, but final flush on shutdown is missing.

filebeat/spooler/spooler.go
ruflin: done, makes the code nicer.

filebeat/beater/filebeat.go
```diff
@@ -96,23 +105,47 @@ func (fb *Filebeat) Run(b *beat.Beat) error {
...
+
+			// Wait for registrar to finish writing registry
+			finishedWg.Wait()
+			logp.Info("All data collection completed. Shutting down.")
```
urso: I wonder if `finishedWG.Wait()` is better done via defer statement by modifying the spooler shutdown order in here to:

```
if *once {
    defer publisherChan.Close() // close publisher channel (optional)
    defer finishedWG.Wait() // wait for all events processed by registrar
    defer spooler.Stop() // stop + flush spooler
} else {
    defer spooler.Stop() // stop spooler
    defer publisherChan.Close() // close publisher channel potentially unblocking the spooler
}
```

filebeat/beater/filebeat.go
```diff
@@ -96,23 +105,47 @@ func (fb *Filebeat) Run(b *beat.Beat) error {
...
 	}
 	// Stop crawler -> stop prospectors -> stop harvesters
 	defer crawler.Stop()
 
-	// Blocks progressing. As soon as channel is closed, all defer statements come into play
+	// Wait for completion if running once.
+	if *once {
+		go func() {
+			logp.Info("Running filebeat once. Waiting for completion ...")
+			crawler.WaitForCompletion()
```
urso: go-routine not really required. Just wait for completion and call fb.Stop (or close fb.done). filebeat will continue with proper shutdown

filebeat/prospector/prospector.go
urso: TBH I still not fully understand why this is required. But resource management just became somewhat more complex :( 

What bothers me is adding many async/sync code-blocks(utils) trying to manage something adding more and more complexity.

filebeat/beater/filebeat.go
ruflin: If there is no go routing CTRL-C Shutdown will not work anymore when running with `once`

filebeat/prospector/prospector.go
ruflin: Yes, complexity currently increases which is not good. I still hope as soon as we have the feature working as we want that we can reduce the complexity again through refactoring.

filebeat/beater/filebeat.go
ruflin: advantage of this is that spooler.Stop is still called only once. We must make sure that CTRL-C still works also if run with once.

urso: this code is already run on shutdown after Kill signal being received (the crawler is closed before being finished). So what todo here? wait for completion of require CTRL-C to kill the process as is instead of waiting for all data being published. Wonder if introducing stop timeout would mitigate this issue.

filebeat/beater/filebeat.go
```diff
@@ -48,10 +51,10 @@ func (fb *Filebeat) Run(b *beat.Beat) error {
 	var wgEvents *sync.WaitGroup // count active events for waiting on shutdown
 	var finishedLogger publisher.SuccessLogger
 
-	if fb.config.ShutdownTimeout > 0 {
-		wgEvents = &sync.WaitGroup{}
-		finishedLogger = newFinishedLogger(wgEvents)
-	}
+	//if fb.config.ShutdownTimeout > 0 || *once {
```
ruflin: @urso do you see an issue with this? As long as we don't use wgEvents nothing will happen.

urso: it's correct. as long as wgEvent.Wait is not installed in sigWait there is no difference in initializing these always. I just didn't want to have all the wgEvent update by default, but overhead should be ok.

filebeat/beater/filebeat.go
```diff
@@ -98,8 +100,11 @@ func (fb *Filebeat) Run(b *beat.Beat) error {
 	// Start publisher
 	publisher.Start()
 	// Stopping publisher (might potentially drop items)
-	defer publisher.Stop()
-	defer successLogger.Close()
+	defer func() {
+		// Closes first the registrar logger to make sure not more events arrive at the registrar
```
urso: maybe add comment: registrarChannel must be closed first to potentially unblock (pretty unlikely) the publisher.

filebeat/beater/filebeat.go
```diff
@@ -108,22 +113,42 @@ func (fb *Filebeat) Run(b *beat.Beat) error {
 	defer func() {
 		// With harvesters being stopped, optionally wait for all enqueued events being
 		// published and written by registrar before continuing shutdown.
-		fb.sigWait.Wait()
+		if !*once {
```
urso: no need to check `*once` flag here. Wait will return immediately if no signalers have been added.

filebeat/beater/filebeat.go
```diff
@@ -108,22 +113,42 @@ func (fb *Filebeat) Run(b *beat.Beat) error {
...
+	// Wait for completion if running once.
+	// Running the shutdown in a go routing is required for CTRL-C to still work for run-once
+
+	if *once {
+		runOnce := func() {
+			logp.Info("Running filebeat once. Waiting for completion ...")
+			crawler.WaitForCompletion()
+
+			// Wait for registrar to finish writing registry
+			wgEvents.Wait()
```
urso: don't wait here. just add `wgEvents.Wait` to sigWait after waiting for shutdown:

```
...
fb.sigWait.Wait()

waitPublished := fb.config.ShutdownTimeout > 0 || *once
if waitPublished {
   fb.sigWait.Add(wgEvents.Wait)
   if fb.config.ShutdownTimeout > 0 {
    fb.sigWait.AddTimeout(fb.config.ShutdownTimeout)
  } 
}

```

filebeat/beater/filebeat.go
```diff
@@ -108,22 +113,42 @@ func (fb *Filebeat) Run(b *beat.Beat) error {
...
+
+			logp.Info("All data collection completed. Shutting down.")
+		}
+
+		fb.sigWait.Add(runOnce)
+	}
+
+	fb.sigWait.AddChan(fb.done)
+	fb.sigWait.Wait()
```
urso: let's use a separate `signalWait` waiting for `runOnce` or `fb.done` to finished. This way the waiter waiting for shutdown and the one waiting for events published will not get mixed up. The `fb.done` channel can be installed multiple times. With channel being closed, it will send a signal immediately.

filebeat/beater/filebeat.go
ruflin: Will leave it in as default for the moment and we can still make it optional in case it becomes a performance issue (I would not assume so)

filebeat/beater/filebeat.go
ruflin: Added

filebeat/beater/filebeat.go
ruflin: removed, was a leftover from previous changes.

filebeat/beater/filebeat.go
ruflin: I adjusted it accordingly. The outcome is the same.

filebeat/beater/filebeat.go
```diff
@@ -130,14 +144,18 @@ func (fb *Filebeat) Run(b *beat.Beat) error {
...
+		waitEvents.Add(withLog(wgEvents.Wait,
 			"Continue shutdown: All enqueued events being published."))
-		fb.sigWait.Add(withLog(waitDuration(timeout),
-			"Continue shutdown: Time out waiting for events being published."))
+		// Wait for either timeout or all events having been ACKed by outputs.
+		if fb.config.ShutdownTimeout > 0 {
+			logp.Info("Shutdown output timer started. Waiting for max %v.", timeout)
+			waitEvents.Add(withLog(waitDuration(timeout),
+				"Continue shutdown: Time out waiting for events being published."))
+		}
```
urso: Add:
```
} else {
    waitEvents.AddChan(fb.done)
}
```

This will ensure CTRL-C will be handled immediately/properly if run-once without shutdown_timeout is configured. If `fb.done` has been closed already, `waitEvents.Wait` will return immediately. If `fb.done` is closed while waiting for `wgEvents.Wait`, shutdown will continue.

ruflin: Good point. Added.

PR #2458, opened 2016-09-05
Change the default beat for import_dashboards

tsg:
Depending on which package we compile for. For example,
the import_dashboard program from the packetbeat packages use
`-beat packetbeat` as the default.


tsg:
jenkins, package it

tsg:
Green on the Jenkins package.

ruflin:
LGTM

PR #2463, opened 2016-09-05
Cleanup Filebeat publishers

ruflin:
* Separate async and sync publisher into own files
* Add Publish interface method which can be used to manually trigger publish (see https://github.com/elastic/beats/pull/2456)
* Add getDataEvents function
* Rename publish to publisher package

### DIFF
--- a/filebeat/publisher/async.go
+++ b/filebeat/publisher/async.go
@@ -1,30 +1,15 @@
-package publish
+package publisher
 
 import (
-	"expvar"
+	"errors"
 	"sync"
 	"sync/atomic"
 	"time"
 
 	"github.com/elastic/beats/filebeat/input"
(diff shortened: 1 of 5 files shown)


filebeat/publisher/async.go
```diff
@@ -168,35 +76,40 @@ func (p *asyncLogPublisher) Start() {
...
-			p.collect()
+			}
 		}
 	}()
 }
 
+func (p *asyncLogPublisher) Publish() error {
+	select {
+	case <-p.done:
+		return fmt.Errorf("async publisher stopped")
```
urso: `return errors.New` ?

ruflin: Changed

PR #2465, opened 2016-09-06
Improve data.json generation for system module

ruflin:
* Some data.json generation tests were missing because they were not running on OS X (diskio, load). Tests for these were added and run under Linux
* Data generation fixed for system module. data.json were not generated as tests were run with integration flag, but system module tests had !integration flag. This is fine for unit tests, but it was removed for the tests that generate the data.json.
* All system module data.json updated


tsg:
So you just executed the make command on Linux? An idea would be to run it in a Docker container so that the results are always the same (I know, after this idea we'll have two problems).

ruflin:
@tsg That is actually what is happening. It is run inside docker containers (which is Linux). This is the default.

PR #2470, opened 2016-09-06
Improve handling of messages larger than 10MB

tsg:
Old implementation dropped messages larger than 10MB (hard coded),
to protect against memory DoS. However, in some cases the transaction
was still recorded, but the parsing of the headers was incomplete.

This makes the support for large messages explicit, by adding a mode
to the parser that "sees" the segments without storing them.

It also makes the HTTP `max_message_size` configurable (no longer hard coded) and fixes an issues in master only were the body was included if the Content-Type header was missing.


tsg:
Ready for reviews. Also, cleaned up history (I prefer 3 individual commits to separate the parts that can be backported to 1.3).

monicasarbu:
LGTM

Issue #2277, opened 2016-08-16
Add support for symlinks


tsg:
Open this to discuss a proper implementation for symlinks in Filebeat. Filebeat used to have accidental support for symlinks which resulted in duplicates and was removed in #1767.

To summarize the IRL discussions: since symlinks potentially come with lots of corner cases (files change under the symlink, the symlink gets removed by the the file not, the other way around, etc.) we think a correct implementation should be done as a new harvester type.

ruflin:
Playing around with Symlinks and looking at our code again, I realised that our current harvester is not aware if it is a symlink or not. The reason is that `file.Stat()` is used instead of `file.LStat()` during opening the file. Based on the `FileInfo` retrieved it is not possible anymore two make the difference. This is the reason even though checks were in place since LSF to disable symlinks, it always worked.

Initially I was thinking it would be best if symlinks is a different harvester type. I changed my mind here as I like the idea that the harvester does not have to care if it is a symlink or not. It keeps the code simpler and the decision if a harvester should be opened for a symlink is up to the prospector. 

One of the initial ides was to follow the symlink and then open the original file. But based on the discussion in https://github.com/elastic/beats/pull/1767 that does not seem the way it should be in the kubernetes use case where the name of symlink is relevant. If the original would be harvested, also the original file name would be reported.

Based on this my conclusion is that the initial (accidental) implementation was as expected. The only side affect with this is that if the symlink or the original file is removed, no further data can be harvested. But I think this is kind of expected.

Here is a first implementation for further discussions: https://github.com/elastic/beats/pull/2478

ruflin:
Thinking again about the additional harvester type `symlink`: The implementation would be very similar on would not change much. The advantage would be that people would be very aware that they use symlinks and would apply it to specific files. The disadvantage is that symlinks and normal files could not be mixed in one prospector. In case we have an `input_type: file` or `input_type: gz` the question would arise, if these also support symlinks. That would become harder if it is it's own harvester type. Also the more accurate name would probably be `log_symlink`.

ruflin:
Closing as https://github.com/elastic/beats/pull/2478 was merged.

PR #2478, opened 2016-09-07
Add experimental symlink support

ruflin:
* Add test to verify that symlinks are disabled by default
* Add test for symlink handling
* Improve error handling in harvester. Return proper error messages instead of logging it. Prevents too many log messages.
* Remove IsRegular file call as not needed anymore and leads to additional Stat calls
* Add documentation

See https://github.com/elastic/beats/issues/2277 and https://github.com/elastic/beats/pull/1767


tsg:
LGTM, great that you got this in before beta1. Should have a changelog item.

Issue #2044, opened 2016-07-15
Duplicate events when filebeat is killed with SIGHUP/SIGINT


robin13:
Please post all questions and issues on https://discuss.elastic.co/c/beats
before opening a Github Issue. Your questions will reach a wider audience there,
and if we confirm that there is a bug, then you can open a new issue.

For confirmed bugs, please report:
- Version: filebeat 1.2.3 and 5.0.0 alpha4 with logstash 5.0.0 alpha4
- Operating System: Ubuntu 14.0.4
- Steps to Reproduce: 

Kill filebeat with SIGHUP (as is done when using the standard service restart: `service filebeat restart`) and then restart will result in duplicate events.  A SIGHUP should be considered a normal restart, and filebeat should wait for ACK from logstash before exiting.

Attached are sample configurations for logstash and filebeat and a test script to run the test multiple times in sequence to reproduce.

[test.tar.gz](https://github.com/elastic/beats/files/365955/test.tar.gz)

Related: https://github.com/elastic/beats/issues/2041

Note - for the script to work you must start logstash with the `-r` restart option.  The line in the script which adds a comment to the logstash configuration causes logstash to reload the configuration and flush the lines.  Logshash file output _should_ flush everything to file after 5 seconds, but does not seem to be honouring this now...

robin13:
Assuming that filebeat cannot wait indefinitely for an ACK (if logstash is blocked), it should wait some "sane" amount (e.g. default 5 seconds, but configurable) to fulfil the mandate of "under normal operations, no data loss or duplication should occur"

ruflin:
The current behaviour is somehow expected, as the connection is "just" closed. I think the best option would be to introduce a config option like `shutdown_timeout` or something similar where a  timeout can be configured. I would keep it set to 0 by default as the "sane" amount can be very different depending on how many messages are sent and if servers are reachable. So it is up to the user to find the right value in his case.

tsg:
We've discussed this a bit in our team meeting, here is a summary. The solution proposed in this ticket (wait for a configurable amount of time before shutting down the publisher and the registrar) would be relatively easy to implement and would help, but it has the following disadvantages:

* The amount of time to wait that would be "enough" is hard to estimate, because we depend not only on the network RTT but also on the Logstash/Elasticsearch processing time. Logstash persistence might make the timing more reliable, but it will still be dependent on the batch size and the load of the system at the given moment.
[...]

PR #2481, opened 2016-09-07
Implement shutdown_timeout

ruflin:
See https://github.com/elastic/beats/issues/2044

Todo:
* [ ] Docs
* [ ] Tests

### DIFF
--- a/filebeat/beater/filebeat.go
+++ b/filebeat/beater/filebeat.go
@@ -2,6 +2,7 @@ package beater
 
 import (
 	"fmt"
+	"time"
 
 	"github.com/elastic/beats/libbeat/beat"
 	"github.com/elastic/beats/libbeat/common"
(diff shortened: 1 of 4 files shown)


filebeat/beater/filebeat.go
```diff
@@ -102,6 +103,9 @@ func (fb *Filebeat) Run(b *beat.Beat) error {
 	// Blocks progressing. As soon as channel is closed, all defer statements come into play
 	<-fb.done
 
+	// Uses shutdown_timeout to wait for potential completion of sending events
+	<-time.After(fb.config.ShutdownTimeout)
```
urso: Original problem with shutdown timeout was: wait for enqueued events to be ACKed on shutdown. This requires the crawler/prospectors being closed before doing the timeout. In addition, we really always want to wait, or rather wait for max timeout on shutdown until all events are ACKed -> continue killing filebeat once events have been killed.

Introducing a timer here won't achieve either, as this only delays processing of the kill signal.

ruflin:
Closing in favor of https://github.com/elastic/beats/pull/2514

PR #2486, opened 2016-09-08
Update HAProxy module

ruflin:
* Add CHANGELOG entry
* Run make collect
* Disable module for short config
* Mark it experimentatl
* Fix doc build
  * Add newline at the end of each fields.yml to ensure newline if it was forgotten in file
  * Mark source with type (still needs improvement)
  * Update link naming as it could conflict with packetbeat
  * Add modules to list (needs auto generation)


metricbeat/module/haproxy/stat/_meta/docs.asciidoc
```diff
@@ -11,8 +12,8 @@ http://www.haproxy.org/download/1.6/doc/management.txt
 
 The following documentation bellow is an extract from the URL above, more specifically from section "9.1. CSV format"
 
-[source]
------------------
+[source,ruby]
```
ruflin: This needs to be fixed. Didn't work without tag. @dedemorton Some suggestion on what the best tag would be to use here?

PR #2491, opened 2016-09-08
Replace component output channels by interfaces

urso:


filebeat/crawler/crawler.go
```diff
@@ -6,26 +6,25 @@ import (
...
-	"github.com/elastic/beats/filebeat/spooler"
 	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/logp"
 )
 
 type Crawler struct {
 	prospectors       []*prospector.Prospector
 	prospectorConfigs []*common.Config
-	spooler           *spooler.Spooler
+	out               prospector.Outlet
```
ruflin: Can you add a comment on this line that "normally" this is the spooler "outlet"? Makes the code much easier to understand.

filebeat/publisher/publisher.go
```diff
@@ -15,12 +16,16 @@ var (
 type LogPublisher interface {
 	Start()
 	Stop()
-	Publish() error
+}
+
+type SuccessLogger interface {
```
ruflin: Can you add a doc note on what this one is doing?

filebeat/publisher/sync.go
```diff
@@ -53,26 +54,26 @@ func (p *syncLogPublisher) Publish() error {
...
-		return errors.New("publishing was stopped")
+		return sigPublisherStop
 	case events = <-p.in:
 	}
 
 	ok := p.client.PublishEvents(getDataEvents(events), publisher.Sync, publisher.Guaranteed)
 	if !ok {
 		// PublishEvents will only returns false, if p.client has been closed.
-		return errors.New("publisher didn't published events")
+		return sigPublisherStop
```
ruflin: Should we really have the same signal here? Could ti be for other reasons that ok wasn't ture?

filebeat/prospector/prospector.go
```diff
@@ -36,11 +36,15 @@ type Prospectorer interface {
 	Run()
 }
 
-func NewProspector(cfg *common.Config, states file.States, spoolerChan chan *input.Event) (*Prospector, error) {
+type Outlet interface {
```
ruflin: Should we move this to a separate package?

filebeat/prospector/prospector.go
```diff
@@ -36,11 +36,15 @@ type Prospectorer interface {
...
+	OnEvent(event *input.Event) bool
```
ruflin: Send instead of OnEvent?

filebeat/crawler/crawler.go
urso: isn't the purpose of abstraction to not know what exactly will be passed?

filebeat/prospector/prospector.go
urso: well, it's specific to prospector/crawler as is. I'd keep it here as full name is `prospector.Outlet` overlapping interfaces can be defined anywhere in go.

filebeat/prospector/prospector.go
urso: I used `OnEvent` to indicate this being kind of a callback. `Send` sounds too much like a channel/io-whatever too me.

filebeat/spooler/spooler.go
```diff
@@ -33,18 +37,17 @@ type spoolerConfig struct {
 // started by calling Start before it can be used.
 func New(
 	config *cfg.Config,
-	publisher chan<- []*input.Event,
+	out Output,
```
ruflin: Can you add a comment that this is normally going to the publisher?

filebeat/publisher/publisher.go
urso: yep

filebeat/publisher/sync.go
urso: we never checked/printed the error. Errors have only be used to indicate: finish publisher loop. The publisher output being closed is one of those signals. We can add some debug statements here, though.

filebeat/beater/channels.go
```diff
@@ -0,0 +1,103 @@
...
+func (c *publisherChannel) Send(events []*input.Event) bool {
+	select {
+	case <-c.done:
+		return false
+	case c.ch <- events:
+		return true
+	}
+}
+
+func newRegistrarLogger(reg *registrar.Registrar) *registrarLogger {
```
ruflin: This could also be moved to the registrar package and then `registrary.getLogger()` is called.

filebeat/crawler/crawler.go
ruflin: The problem here is, that the package name suggests its something from the prospector. Abstraction does not prevent us to add docs to make reading the code easier. In general I think abstraction is nice but often makes code harder to read. So adding docs / comments shouldn't hurt.

filebeat/beater/channels.go
urso: Idea is to hook up all the components in beater Run method. The registrarLogger is used to connect the publisher and the registrar.

One of the reasons it's not in registrar: due to all components being async workers (which becomes super aparent in provided interfaces), we have to be able to close the channel without stopping the registrar. This is achieved via registrarLogger being closed right before `publisher.Stop` is executed.

filebeat/crawler/crawler.go
urso: well, Outlet being an interface I expect it not to be something coming from prospector, but something whatever implementing the required interface.

filebeat/spooler/spooler.go
urso: I'd rather do not. Output is the interface any component must implement. It's part of spooler's interface as reusable component.

PR #2492, opened 2016-09-08
Dockerfile for Metricbeat to test cgroups

monicasarbu:
Add a Dockerfile for Metricbeat to test cgroups functionality. 

You need to run the following commands to start a container with Kibana, one with Elasticsearch and one with Metricbeat:
```
$ make up ENV=5.0.0-cgroups.yml 
```

To stop all the containers, you can use:
```
$ make stop ENV=5.0.0-cgroups.yml
```

### DIFF
--- a/testing/environments/5.0.0-cgroups.yml
+++ b/testing/environments/5.0.0-cgroups.yml
@@ -13,3 +13,13 @@ logstash:
 kibana:
   build: ./docker/kibana
   dockerfile: Dockerfile-5.0.0-alpha5
+
+metricbeat:
+  build: ./docker/metricbeat
+  dockerfile: Dockerfile-5.0.0-cgroups
+  links:
+    - elasticsearch
+  volumes:
+    - "/proc:/hostfs/proc:ro"
+    - "/sys/fs/cgroup:/hostfs/sys/fs/cgroup"
+  command: -system.hostfs=/hostfs -E output.elasticsearch.hosts=elasticsearch -E metricbeat.modules.0.cgroups=true
--- a/testing/environments/Makefile
+++ b/testing/environments/Makefile
@@ -10,6 +10,11 @@ stop:
 	${BASE_COMMAND} stop
 	${BASE_COMMAND} rm -f
 
+
+up:
+	${BASE_COMMAND} build
+	${BASE_COMMAND} up
+
 # Be careful using this command, as it will remove all containers and volumes of your docker-machine
 clean:
 	docker stop $(shell docker ps -a -q)
--- a/testing/environments/docker/metricbeat/config/metricbeat.cgroups.yml
+++ b/testing/environments/docker/metricbeat/config/metricbeat.cgroups.yml
@@ -1,0 +1,103 @@
+###################### Metricbeat Configuration Example #######################
+
+# This file is an example configuration file highlighting only the most common
+# options. The metricbeat.full.yml file from the same directory contains all the
+# supported options with more comments. You can use it as a reference.
+#
+# You can find the full configuration reference here:
+# https://www.elastic.co/guide/en/beats/metricbeat/index.html
+
+#==========================  Modules configuration ============================
+metricbeat.modules:
+
+#------------------------------- System Module -------------------------------
+- module: system
+  metricsets:
+    # CPU stats
+    - cpu
+
+    # System Load stats
+    - load
+
+    # Per CPU core stats
+    #- core
+
+    # IO stats
+    #- diskio
+
+    # Per filesystem stats
+    - filesystem
+
+    # File system summary stats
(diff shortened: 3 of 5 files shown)


testing/environments/docker/metricbeat/config/metricbeat.cgroups.yml
```diff
@@ -0,0 +1,103 @@
...
+
+#================================ Outputs =====================================
+
+# Configure what outputs to use when sending the data collected by the beat.
+# Multiple outputs may be used.
+
+#-------------------------- Elasticsearch output ------------------------------
+output.elasticsearch:
+  # Array of hosts to connect to.
+  hosts: ["localhost:9200"]
```
ruflin: That is probably not going to work as elasticsearch is not running on localhost but `elasticsearch`. The environment variables should be used ehre.

testing/environments/5.0.0-cgroups.yml
```diff
@@ -0,0 +1,25 @@
+# This should test the environment with the latest snapshots
+# This is based on base.yml
+
+elasticsearch:
```
ruflin: To load all the envrionments normally the chaining of docker-compose files is used. In this case it would be `make start ENV=5.0.0-cgroups.yml`. Like this all port are already exposed and environment variables are set. In addition, this file gets much smaller (only metricbeat part needed)

testing/environments/docker/metricbeat/config/metricbeat.cgroups.yml
monicasarbu: The elasticsearch is set in the command for running metricbeat in the Dockerfile-5.0.0-cgroups.

testing/environments/5.0.0-cgroups.yml
monicasarbu: I didn't use the make command to start it, as it creates more containers than I want, so the idea was to use the commands that are set in the description. Probably we also need to add -f ports.yml ...

testing/environments/docker/metricbeat/config/metricbeat.cgroups.yml
```diff
@@ -0,0 +1,103 @@
...
+
+    # Network stats
+    - network
+
+    # Per process stats
+    - process
+  enabled: true
+  period: 10s
+  processes: ['.*']
+  cgroups: true
```
andrewkroh: You should be able to use the default configuration file and specify this option on the CLI using `-E metricbeat.modules.0.cgroups=true`. This was recently fixed by #2521.

testing/environments/ports.yml
```diff
@@ -11,7 +11,3 @@ logstash:
 kibana:
   ports:
     - "5601:5601"
-
-beat:
```
monicasarbu: @ruflin I moved kibana link to `base.yml`. Would that be ok with you?

testing/environments/docker/metricbeat/config/metricbeat.cgroups.yml
monicasarbu: Fixed. Thanks @urso for such a fast fix.

testing/environments/ports.yml
ruflin: @monicasarbu No. It is on purpose not in the base.yml file as if it is in the base file it is fetched and built for each CI build on travis and jenkins which makes the builds a lot slower.

testing/environments/Makefile
```diff
@@ -10,6 +10,10 @@ stop:
 	${BASE_COMMAND} stop
 	${BASE_COMMAND} rm -f
 
+
+up:
```
ruflin: I recommend also to run the build command before running up, to make sure you always have the most recent environment.

testing/environments/base.yml
```diff
@@ -4,6 +4,7 @@ beat:
   links:
     - elasticsearch
     - logstash
+    - kibana
```
ruflin: This should be removed again as otherwise CI builds will be much slower.

testing/environments/ports.yml
ruflin: @monicasarbu I should have added a comment here why it is not in the base.yml as I had it there first and it is not obvious. Could you add a comment?

monicasarbu: Thanks for the explanation. I added a comment in the code.

testing/environments/base.yml
monicasarbu: done

testing/environments/Makefile
monicasarbu: Yes, you are right. I am adding build before.

monicasarbu:
I addressed all comments, and it's ready for another review.

PR #2502, opened 2016-09-08
Set docs version to alpha5 in the 5.0 branch

tsg:
This should have been done with the alpha5 release. Better late than never.


dedemorton:
LGTM

PR #2509, opened 2016-09-09
Fix map concurrent bug

amyangfei:
crawler.Registrar.state may be accessed both in function `(*Prospector).scan` and pulling the events from the `crawler.Registrar.Persist`, which leads to `fatal error: concurrent map read and map write`

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -63,6 +63,14 @@ https://github.com/elastic/beats/compare/v1.3.0...1.3[Check the HEAD diff]
 
 ////////////////////////////////////////////////////////////
 
+[[release-notes-1.3.1]]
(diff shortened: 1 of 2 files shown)


ruflin:
@amyangfei Thanks a lot for spotting this and providing a fix. I will look into this ASAP.

ruflin:
@amyangfei I had a closer look at this and I assume this error happens directly on startup especially with config files with a large number of prospectors defined?

Can you update the CHANGELOG.asciidoc und bug fixes?

amyangfei:
@ruflin I have updated the CHANGELOG.asciidoc :smile:

ruflin:
@amyangfei Merged. Thanks a lot for the fix.

PR #2510, opened 2016-09-09
filebeat Channels close

urso:
ensure closed channels can not be used for sending after having been closed.

### DIFF
--- a/filebeat/beater/channels.go
+++ b/filebeat/beater/channels.go
@@ -2,6 +2,7 @@ package beater
 
 import (
 	"sync"
+	"sync/atomic"
 
 	"github.com/elastic/beats/filebeat/input"
 	"github.com/elastic/beats/filebeat/registrar"
@@ -12,6 +13,8 @@ type spoolerOutlet struct {
 	wg      *sync.WaitGroup
 	done    <-chan struct{}
 	spooler *spooler.Spooler
+
+	isOpen int32 // atomic indicator
 }
 
 type publisherChannel struct {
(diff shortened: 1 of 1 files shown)


filebeat/beater/channels.go
```diff
@@ -67,6 +77,7 @@ func (c *publisherChannel) Close() { close(c.done) }
 func (c *publisherChannel) Send(events []*input.Event) bool {
 	select {
 	case <-c.done:
+		c.ch = nil
```
ruflin: This is definitively worth a line of comment ;-)

filebeat/beater/channels.go
```diff
@@ -37,10 +40,16 @@ func newSpoolerOutlet(
 		done:    done,
 		spooler: s,
 		wg:      wg,
+		isOpen:  1,
 	}
 }
 
 func (o *spoolerOutlet) OnEvent(event *input.Event) bool {
+	open := atomic.LoadInt32(&o.isOpen) == 1
```
ruflin: It could still happen that 2 go routines both get `open == 0`, then o.done is closed and one goet into the first case, the other one into the second? Do we need a mutex?

urso: I think this is ok and part of non-deterministic nature of spoolerOutlet. But one prospector go-routine will re-use the same channel. That is, while there is a chance some events still being published during close, the same prospector will not be able to publish events after having seen the close signal. Most important is the individual prospectors not publishing after having seen the channel being closed (directly or indirectly).

filebeat/beater/channels.go
urso: Oh, I thought it's a common known golang pattern. assign nil to channel to block on channel forever.

ruflin:
LGTM

PR #2511, opened 2016-09-09
Cleanup naming in filebeat for better readability

ruflin:
Cleanup naming of channel and loggers in filebeat to make it easier to understand what actually happens but keep the abstraction.
Channels and loggers with a specific tasks were moved to the relevant packages


filebeat/beater/filebeat.go
```diff
@@ -82,19 +81,27 @@ func (fb *Filebeat) Run(b *beat.Beat) error {
 		logp.Err("Could not start registrar: %v", err)
 	}
 	// Stopping registrar will write last state
-	defer registrar.Stop()
+	defer func() {
```
ruflin: @urso can you have a really close look at the following defer statements which I have changed? I moved the `registrarLogger.Close()`which was `sucessLogger `before, before the registrar Stopping as I think that makes more sense. Same change I did for the publisherChan below.

urso: This change here should not matter much, due to registrar normally not being blocked for long times. But `registrarLogger` using a channel + if registrar would block, publisher would be blocked as well.

filebeat/beater/filebeat.go
```diff
@@ -82,19 +81,27 @@ func (fb *Filebeat) Run(b *beat.Beat) error {
...
 	publisher.Start()
 	// Stopping publisher (might potentially drop items)
-	defer publisher.Stop()
-	defer successLogger.Close()
+	defer func() {
+		// Closing channel to publisher
+		publisherChan.Close()
+		// Stopping publisher itself
+		publisher.Stop()
+	}()
```
urso: closing `publisherChan` after spooler can make the spooler hang if:
- publisher is blocked by output
- channel being full
- spooler is trying to push via publisherChan

=> shutdown hangs

filebeat/beater/filebeat.go
ruflin: I reverted this change and add a comment.

ruflin:
Closing in favor of https://github.com/elastic/beats/pull/2584

PR #2484, opened 2016-09-07
Adding threads_cached,threads_created,threads_connected,threads_running to mysql module exports.

claygorman:
See https://discuss.elastic.co/t/mysql-exports/59899 for information.

Issue #1795, opened 2016-06-03
Remove comments from example configs in the Beats reference guides


dedemorton:
The comments in the doc get out of sync with the comments in the config files, and they are redundant, so let's remove them.

dedemorton:
Closing in favor of tracking this issue in https://github.com/elastic/beats/issues/2482

PR #2513, opened 2016-09-09
Remove comments from config example code

dedemorton:
Removes comments from config example code (as described in issue #1795 and tracked by issue #2484). 

This change makes the config examples a lot easier to scan, and makes it easier for us to keep the config examples in sync with the .yml config files.  

The description of each option should provide any info that was available in the comment. If it doesn't, let me know so I can add more info to the description. This is also a good opportunity to scan the config examples for any old settings or configs that we don't recommend anymore. :-)

Note that I didn't remove comments from the Metricbeat config examples because they are generated. @ruflin I think it's OK if we can't modify the doc generator to remove comments because having the examples generated is a big plus IMHO.

Also made some minor tweaks, like removing unnecessary quotation marks.


libbeat/docs/generalconfig.asciidoc
```diff
@@ -20,16 +20,8 @@ Here is an example configuration:
...
 ------------------------------------------------------------------------------
-# The name of the shipper that publishes the network data. It can be used to group
-# all the transactions sent by a single shipper in the web interface.
-# If this options is not defined, the hostname is used.
-#name:
-
-# The tags of the shipper are included in their own field with each
-# transaction published. Tags make it easy to group servers by different
-# logical properties.
+name: "my-shipper"
```
ruflin: Would prefer to not have `shipper` here but don't have a good alternative.

PR #2514, opened 2016-09-10
Add shutdown timeout

urso:
Introduce signalWait structure to filebeat waiting for stop signal after
fb.done was closed. If shutdown_timeout is configured, filebeat will either
stop after timeout or after all enqueued events have been successfully
published and written by registrar.

I hope the signalWait idea can be reused for implementing run-once mode as well.

### DIFF
--- a/filebeat/beater/filebeat.go
+++ b/filebeat/beater/filebeat.go
@@ -2,6 +2,7 @@ package beater
 
 import (
 	"fmt"
+	"sync"
 
 	"github.com/elastic/beats/libbeat/beat"
 	"github.com/elastic/beats/libbeat/common"
@@ -16,8 +17,9 @@ import (
 
 // Filebeat is a beater object. Contains all objects needed to run the beat
 type Filebeat struct {
-	config *cfg.Config
-	done   chan struct{}
+	config  *cfg.Config
+	sigWait *signalWait
+	done    chan struct{}
 }
 
 // New creates a new Filebeat pointer instance.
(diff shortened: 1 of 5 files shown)


filebeat/beater/signalwait.go
```diff
@@ -0,0 +1,44 @@
...
+
+func (s *signalWait) AddTimer(t *time.Timer) {
+	s.Add(func() { <-t.C })
+}
+
+func (s *signalWait) AddTimeout(d time.Duration) {
+	s.AddTimer(time.NewTimer(d))
+}
+
+func (s *signalWait) Signal() {
```
ruflin: What is this needed for?

urso: It's for sending inserting a signal. E.g. having received Ctrl-C one can use `Signal()` to insert a signal into the waiter that triggers right now, such that `Wait` will not block.

e.g.
```
<- fb.done
if whatever {
    sigWait.Signal()
}
```

ruflin: Got it

ruflin:
LGTM

filebeat/beater/filebeat.go
```diff
@@ -106,6 +125,12 @@ func (fb *Filebeat) Run(b *beat.Beat) error {
 	// Blocks progressing. As soon as channel is closed, all defer statements come into play
 	<-fb.done
 
+	if fb.config.ShutdownTimeout > 0 {
+		// Wait for either timeout or all events having been ACKed by outputs.
```
ruflin: We should add here a log message that shutdown will wait for time X to shutdown or success

PR #2515, opened 2016-09-11
Apply schema to Nginx stub status

mrkschan:


mrkschan:
Question: How do I get derived value like https://github.com/elastic/beats/pull/2515/files#diff-75c7eb6ee659f97ca5423a61e0c2a470R75 without doing `strconv.Atoi`?

ruflin:
@mrkschan There is not easy solution for the derived values I'm aware of as the schema gets only applied before sending the event. Perhaps @tsg has some suggestions?

In general we do not try to calculate values in the client side (if not necessary) as most simple calculations can be done in Kibana directly or even be done during query time of elasticsearch. The advantage here is that in case the calculation changes, it could be redone for all values.

mrkschan:
Revised, dropped the values that are derived.

metricbeat/module/nginx/stubstatus/data.go
```diff
@@ -5,15 +5,26 @@ import (
...
+	s "github.com/elastic/beats/metricbeat/schema"
+	c "github.com/elastic/beats/metricbeat/schema/mapstrstr"
 )
 
 var (
 	activeRe  = regexp.MustCompile("Active connections: (\\d+)")
 	requestRe = regexp.MustCompile("\\s(\\d+)\\s+(\\d+)\\s+(\\d+)")
 	connRe    = regexp.MustCompile("Reading: (\\d+) Writing: (\\d+) Waiting: (\\d+)")
+
+	schema = s.Schema{
```
ruflin: nice

mrkschan:
@ruflin do you mean i should update the change log as well?

tsg:
@mrkschan we've discussed this one a bit, and I'm not very happy that you had to remove the fields because the `schema` package can't support that. I'm going to try to look for an alternative in the next few days.

If we manage to avoid removing the fields, it also makes the change "not breaking", so then it's easier to just release it in the next minor after 5.0 (the freeze for 5.0 already happened).

mrkschan:
@tsg is there any update on https://github.com/elastic/beats/pull/2515#issuecomment-252595472?

tsg:
@mrkschan Oops, sorry, this fell of my radar. To be honest, I'm not sure if it's worth using the schema in this case, looks like the current code is already not too repetitive. What do you think, should we just close the PR? Or is it fixing something?

Apologies again for the super late answer.

ruflin:
I'm closing this PR as there hasn't been an update for a long time. I like in general the idea to move to a more schema like approach and we also added new features to the schema in the last month. We probably should revisit the modules in general to see what we can do.

PR #2519, opened 2016-09-12
testing shutdown timeout

urso:


filebeat/beater/filebeat.go
```diff
@@ -126,9 +126,13 @@ func (fb *Filebeat) Run(b *beat.Beat) error {
 	<-fb.done
 
 	if fb.config.ShutdownTimeout > 0 {
+		logp.Info("Shutdown timeout %v", fb.config.ShutdownTimeout)
```
ruflin: Shutdown timeout started: %v

filebeat/beater/filebeat.go
```diff
@@ -126,9 +126,13 @@ func (fb *Filebeat) Run(b *beat.Beat) error {
...
+
 		// Wait for either timeout or all events having been ACKed by outputs.
-		fb.sigWait.Add(wgEvents.Wait)
-		fb.sigWait.AddTimeout(fb.config.ShutdownTimeout)
+		fb.sigWait.Add(withLog(wgEvents.Wait,
+			"Continue shutdown: All enqueued events being published."))
+		fb.sigWait.Add(withLog(waitDuration(fb.config.ShutdownTimeout),
+			"Continue shutdown: shutdown timeout."))
```
ruflin: "Continue shutdown: Shutdown timeout reached before all events published."

filebeat/tests/system/test_shutdown.py
```diff
@@ -22,18 +20,56 @@ def test_shutdown(self):
...
+            ignore_older="1h",
+            shutdown_timeout="10s",
+        )
+        proc = self.start_beat(logging_args=["-e", "-v"])
+        time.sleep(.5)
+        proc.check_kill_and_wait()
+
+        log = self.get_log()
+        assert "Shutdown timeout 10s" in log
+        assert "Continue shutdown: All enqueued events being published." in log
```
ruflin: Could you also check the registry file that all events were presisted there (offset)?

filebeat/tests/system/test_shutdown.py
```diff
@@ -22,18 +20,56 @@ def test_shutdown(self):
...
+            ignore_older="1h",
+            shutdown_timeout="1s",
+        )
+        proc = self.start_beat(logging_args=["-e", "-v"])
+        time.sleep(.5)
+        proc.check_kill_and_wait()
+
+        log = self.get_log()
+        assert "Shutdown timeout 1s" in log
+        assert "Continue shutdown: shutdown timeout." in log
```
ruflin: Could you check that the registry does not contain all events yet (offset)?

ruflin:
It looks like the jenkins test somehow managed to end the test with a panic.

filebeat/beater/filebeat.go
urso: when using `Shutdown timeout started` I would assume `%v` to be `time.Now()`, right?

filebeat/tests/system/test_shutdown.py
urso: any helpers for testing this?

filebeat/tests/system/test_shutdown.py
urso: any helpers for testing this?

filebeat/beater/filebeat.go
ruflin: `Shutdown timeout strated. Waiting %v`?

filebeat/tests/system/test_shutdown.py
ruflin: `self.get_registry()`: https://github.com/elastic/beats/blob/master/filebeat/tests/system/test_crawler.py#L326

urso:
There has been a race on shutdown with crawlers not being stopped yet, but `wgEvents.Wait()` being  run with potentially no events currently scheduled. The effect might be a panic on shutdown.

filebeat/tests/system/config/filebeat.yml.j2
```diff
@@ -132,11 +133,16 @@ processors:
 # Multiple outputs may be used.
 
 #------------------------------- File output ----------------------------------
+{%- if logstash %}
```
ruflin: not that we have enabled: true again, we probably don't need to full block comments ;-)

ruflin:
LGTM

filebeat/beater/filebeat.go
```diff
@@ -119,16 +119,21 @@ func (fb *Filebeat) Run(b *beat.Beat) error {
 	if err != nil {
 		return err
 	}
-	// Stop crawler -> stop prospectors -> stop harvesters
-	defer crawler.Stop()
-
 	// Blocks progressing. As soon as channel is closed, all defer statements come into play
+
+	// Stop crawler -> stop prospectors -> stop harvesters
```
ruflin: that should go one line down

filebeat/beater/filebeat.go
```diff
@@ -119,16 +119,21 @@ func (fb *Filebeat) Run(b *beat.Beat) error {
...
 	<-fb.done
+	crawler.Stop()
```
ruflin: Can you please add a note here why we don't use defer for the crawler.Stop? I already see me adding a defer to it in 3 months :-)

filebeat/beater/filebeat.go
urso: done

filebeat/beater/filebeat.go
urso: done

filebeat/tests/system/config/filebeat.yml.j2
urso: as template variable `logstash.host` is used to configure the hostname, I'd keep it like this.

Issue #2482, opened 2016-09-07
Beats doc updates


dedemorton:
This is a meta issue to track Beats documentation work for 5.0.0. This issue was created towards the end of the 5.0.0 release, so it doesn't include all the changes made to the doc during the alpha  releases.

If you have trouble clicking an item in the checklist to mark it as complete, open this comment in edit mode and change the source to `[x]`. Seems to be a bug in ticking the check boxes in this issue.

**5.0 GA:**
- [x] Resolve any [doc issues flagged for 5.0.0-GA](https://github.com/elastic/beats/issues?utf8=%E2%9C%93&q=is%3Aopen%20label%3Av5.0.0-GA%20label%3Adocs). Some of the open doc issues are duplicated in this list. 
- [x] [DeDe] Document how to configure ingest node to have GeoIP info. See https://github.com/elastic/beats/issues/1671
- [x] Update Developer Guide. See https://github.com/elastic/beats/issues/1077 
- [x] Create migration guide for 2.x to 5.x. See https://github.com/elastic/beats/issues/1488 
- [x] [DeDe ] Document load-balancing modes in Filebeat. See https://github.com/elastic/beats/issues/1852
- [x] [DeDe] Submit request to get redirects set up. See https://github.com/elastic/beats/issues/2426
- [x] [DeDe] Add doc example that shows how to configure the Kafka output. See https://github.com/elastic/beats/issues/1339 
- [x] [DeDe] Update links to shield documentation for 5.0. See https://github.com/elastic/beats/issues/1308
- [x] Edit Breaking changes and Upgrading doc (see #2631)
- [x] [DeDe] Required cleanup:
  - [x] Remove any known issue notes that are specific to the alpha or beta releases. 
  - [x] Remove any coming, added, or deprecated tags that are relevant to alpha and beta only (keep deprecation tags that mark items deprecated between 2.4 and 5.0).
  - [x] Clean up config sections that were added to support changes between alpha4 and alpha5.
- [x] Make a doc page that explains all the Breaking Changes
- [x] Add label to indicate that doc about adding a new protocol is out-of date. Include a link to the protocol analyzer generator. 
- [x] [DeDe] Improve diagrams:
  - [x] Improve conceptual diagram of Filebeat (see https://github.com/elastic/beats/issues/713).
  - [x] Improve diagram in the overview to the Platform Reference (use new icons and logos).
[...]

PR #2527, opened 2016-09-12
Add jump lists to container topics

dedemorton:
This PR adds jump lists to container topics to make it clear to readers that each container topic introduces a series of related topics. I'm not crazy about manually duplicating the TOC (it's a hack and requires manual maintenance), but because of how the TOC currently works (scrolls with the page and is sometimes slow to load), it's the best solution for now. 

Also made some structural changes to move topics that seemed to be in the wrong place. 

Plus fixed some typos that I found while spell checking my changes.


dedemorton:
This PR completes the misc editorial changes for beta 1 listed in #2482

tsg:
LGTM.

PR #2529, opened 2016-09-12
Change error message from 'no host...' to 'no hosts...'

mefju:
In config file user define 'hosts' not 'host', so this error message is confusing


ruflin:
@mefju Merged. Thanks a lot for the contribution.

PR #2530, opened 2016-09-13
Misc content changes

dedemorton:
Resolves issues listed under "Misc content changes" in #2482 

@urso The changes to faq.asciidoc are the result of some comments that you made. Can you make sure I got the details right? Also can you confirm that the root CA restriction is removed? If so, let me know whether to backport the changes.

### DIFF
--- a/winlogbeat/etc/fields.yml
+++ b/winlogbeat/etc/fields.yml
@@ -42,8 +42,11 @@
       dict-type: keyword
       required: false
       description: >
-        The event specific data. This field is mutually exclusive with
-        `user_data`.
+        The event-specific data. This field is mutually exclusive with
+        `user_data`. If you are capturing event data on versions prior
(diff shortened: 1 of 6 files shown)


winlogbeat/etc/fields.yml
```diff
@@ -42,8 +42,11 @@
       dict-type: keyword
       required: false
       description: >
-        The event specific data. This field is mutually exclusive with
-        `user_data`.
+        The event-specific data. This field is mutually exclusive with
+        `user_data`. If you are capturing event data on versions prior
+        to Windows Vista, the parameters in `event_data` are named `param1`,
+        `param2`, and so on, because event log messages are unnamed in
```
andrewkroh: `s/messages/parameters/`

dedemorton: Fixed

PR #2532, opened 2016-09-13
Allow snapshot dashboards to be loaded

ruflin:
Testing the snapshot builds its a little bit painful the load the snapshot dashboards. Because of this I introduce a -snapshot flag to load these dashboards automatically. This could even be improved if the snapshot build of `import_dashboards` would by default load the snapshot dashboards. It would require that additional flags are passed to the build.

### DIFF
--- a/libbeat/dashboards/import_dashboards.go
+++ b/libbeat/dashboards/import_dashboards.go
@@ -60,6 +60,7 @@ type Options struct {
 	Pass           string
 	OnlyDashboards bool
 	OnlyIndex      bool
+	Snapshot       bool
 }
 
 type CommandLine struct {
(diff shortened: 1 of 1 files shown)


libbeat/dashboards/import_dashboards.go
```diff
@@ -559,6 +561,9 @@ func (imp Importer) ImportArchive() error {
 		archive = imp.cl.opt.File
 	} else if imp.cl.opt.Url != "" {
 		// it's an URL
+		if imp.cl.opt.Snapshot {
```
monicasarbu: I think you wanted here to be able to pass either an `-url` or the `-snapshot` option, not both in the same time.

monicasarbu:
LGTM

PR #2539, opened 2016-09-13
Prevent prospector scans from racing and leaking harvesters

viveklak:
Relevant to https://discuss.elastic.co/t/filebeat-5-0-0-alpha5-multiple-harvesters-for-same-file-sending-the-same-messages-over-and-over

### DIFF
--- a/filebeat/prospector/prospector_log.go
+++ b/filebeat/prospector/prospector_log.go
@@ -11,6 +11,7 @@ import (
 	"github.com/elastic/beats/filebeat/input"
 	"github.com/elastic/beats/filebeat/input/file"
 	"github.com/elastic/beats/libbeat/logp"
+	"sync"
 )
 
 var (
@@ -22,6 +23,7 @@ type ProspectorLog struct {
 	Prospector *Prospector
 	config     prospectorConfig
 	lastClean  time.Time
+	mutex      sync.Mutex
 }
 
 func NewProspectorLog(p *Prospector) (*ProspectorLog, error) {
@@ -54,6 +56,8 @@ func (p *ProspectorLog) Init() {
 }
 
 func (p *ProspectorLog) Run() {
+	p.mutex.Lock()
+	defer p.mutex.Unlock()
 	logp.Debug("prospector", "Start next scan")
 
 	p.scan()
@@ -179,6 +183,7 @@ func (p *ProspectorLog) scan() {
 		// Decides if previous state exists
 		if lastState.IsEmpty() {
 			logp.Debug("prospector", "Start harvester for new file: %s", newState.Source)
+			p.Prospector.states.Update(newState)
 			err := p.Prospector.startHarvester(newState, 0)
 			if err != nil {
 				logp.Err("Harvester could not be started on new file: %s, Err: %s", newState.Source, err)


ruflin:
@viveklak Thanks for taking this on. Can you share some more thoughts on your solution and what the problem is you fixed?

filebeat/prospector/prospector_log.go
```diff
@@ -179,6 +183,7 @@ func (p *ProspectorLog) scan() {
 		// Decides if previous state exists
 		if lastState.IsEmpty() {
 			logp.Debug("prospector", "Start harvester for new file: %s", newState.Source)
+			p.Prospector.states.Update(newState)
```
ruflin: New states should not be persisted before the harvester is started as during the startup of a harvester also some checks will be made and only then the state will be sent to the prospector.

filebeat/prospector/prospector_log.go
```diff
@@ -54,6 +56,8 @@ func (p *ProspectorLog) Init() {
 }
 
 func (p *ProspectorLog) Run() {
+	p.mutex.Lock()
```
ruflin: This indicates that you suspect that Run is called for the same prospector multiple times in parallel. This should not be the case as Run is not executed inside a go routine: https://github.com/elastic/beats/blob/master/filebeat/prospector/prospector.go#L145 Let me know if I miss some parts here.

ruflin:
One thing that I could see happen is, that as all harvesters are started in a go routine and only then persist the state, that in case this takes too long to persist all new states, that a scan happens again and a new harvester is started again. This would require a rather low scan_frequency. There are some system tests that try to emulate this and so far it seems this "race" didn't happen. The reason your PR above would prevent this case is not the mutex but the state update. Unfortunately this will bring some other problems.

It would be really great if we could find a way to more or less reliably reproduce this. I will try to write a test case with a very low scan_frequency and already a very large number of files on startup.

ruflin:
I think I can also see how this relates in https://discuss.elastic.co/t/filebeat-5-0-0-alpha5-multiple-harvesters-for-same-file-sending-the-same-messages-over-and-over/59866 to a blocking kafka output. As the through which the state updates are sent is blocked by the output, the new states are not sent form the harvesters which are started. As no states are sent but scans continue, it will start the same harvester for the file again and again as the state was never processed.

filebeat/prospector/prospector_log.go
viveklak: You are right, sorry I missed that. The mutex is not required - not sure why I missed that.

viveklak:
@ruflin You understood correctly. The core of the problem is that the registration of the harvester with the prospector happens asynchronously (the harvester itself is started in a go routine and sends its state on a shared channel to the prospector). If there is a significant enough delay in the harvester returning its initial state to the prospector, the prospector may start another harvester for the same source. The delay could be due to a large number of existing harvesters competing on the same channel to send state back, a blocking output, etc. 

Looks like registration of the initial state within the prospector seems to have helped in my case but as you said, might have other issues.

ruflin:
@viveklak I opened here a PR with a potential fix: https://github.com/elastic/beats/pull/2541

The main downside / upside now is, scans can be blocked if the output is blocked. On the bright side this prevents further harvesters to be started and keep them open. On the less bright side that means that in case output becomes available again, old files will be finished sending and new files were potentially rotated away in the meantime.

It would be great if you could test my PR to see if it helps in your case.

viveklak:
@ruflin awesome! Will update with the results of my test. Thanks for the quick turn around!

ruflin:
I'm closing this PR as https://github.com/elastic/beats/pull/2541 was merged.

@viveklak Keep me posted here about your tests.

viveklak:
@ruflin Ran a test overnight and didn't see the issue anymore. Thanks again for jumping on the problem!

ruflin:
@viveklak Great to hear.

PR #2541, opened 2016-09-14
Fix concurrent harvesters

ruflin:
In case newly started harvesters did not persist their first state before the next scan started, it could have happened that multiple harvesters were started for the same file. This could have been cause by a large number of files or the output blocking.

The problem is solve that the Setup step of the Harvester is now synchronus and blocking the scan. Part of this is also updating the first state of the as part of the prospector.

The side affect of this change is that now a scan is blocking in case the channel is blocked which means the output is probably not responding. If the output is not responding, scans will not continue and new files will not be discovered until output is available again.

The code can be further simplified in the future by merging create/startHarvester. This will be done in a second step to keep backport commit to a minimum.

See also #2539

### DIFF
--- a/filebeat/prospector/prospector.go
+++ b/filebeat/prospector/prospector.go
@@ -1,6 +1,7 @@
 package prospector
 
 import (
+	"errors"
 	"expvar"
 	"fmt"
 	"sync"
@@ -29,7 +30,6 @@ type Prospector struct {
 	states           *file.States
 	wg               sync.WaitGroup
 	harvesterCounter uint64
-	stateMutex       sync.Mutex
 }
 
 type Prospectorer interface {
@@ -117,7 +117,10 @@ func (p *Prospector) Run() {
 				logp.Info("Prospector channel stopped")
 				return
 			case event := <-p.harvesterChan:
-				p.updateState(event)
+				err := p.updateState(event)
+				if err != nil {
+					return
+				}
 			}
 		}
 	}()
@@ -139,9 +142,7 @@ func (p *Prospector) Run() {
 
 // updateState updates the prospector state and forwards the event to the spooler
 // All state updates done by the prospector itself are synchronous to make sure not states are overwritten
-func (p *Prospector) updateState(event *input.Event) {
-	p.stateMutex.Lock()
-	defer p.stateMutex.Unlock()
+func (p *Prospector) updateState(event *input.Event) error {
 
 	// Add ttl if cleanOlder is enabled
 	if p.config.CleanInactive > 0 {
@@ -151,10 +152,11 @@ func (p *Prospector) updateState(event *input.Event) {
 	ok := p.outlet.OnEvent(event)
 	if !ok {
 		logp.Info("Prospector outlet closed")
-		return
+		return errors.New("prospector outlet closed")
 	}
 
 	p.states.Update(event.State)
+	return nil
 }
 
 func (p *Prospector) Stop() {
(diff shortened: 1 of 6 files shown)


filebeat/prospector/prospector.go
```diff
@@ -116,18 +117,7 @@ func (p *Prospector) Run() {
...
 			case event := <-p.harvesterChan:
-				// Add ttl if cleanOlder is enabled
-				if p.config.CleanInactive > 0 {
-					event.State.TTL = p.config.CleanInactive
-				}
-
-				ok := p.outlet.OnEvent(event)
-				if !ok {
-					logp.Info("Prospector outlet closed")
-					return
```
urso: change in logic here. This return used to stop this worker. The return having moved to updateState will not stop the worker potentially publishing more incoming events. The worker instead should be dropped + report to prospector it needs to 'shutdown' + drain the harvesterChan, so no harvester will be blocked.

filebeat/prospector/prospector.go
```diff
@@ -176,12 +186,24 @@ func (p *Prospector) startHarvester(state file.State, offset int64) error {
 	}
 
 	state.Offset = offset
+	// All harvesters must start with a finished state
+	state.Finished = false
```
urso: why?

filebeat/prospector/prospector.go
```diff
@@ -191,8 +213,9 @@ func (p *Prospector) startHarvester(state file.State, offset int64) error {
 			atomic.AddUint64(&p.harvesterCounter, ^uint64(0))
 			p.wg.Done()
 		}()
+
 		// Starts harvester and picks the right type. In case type is not set, set it to defeault (log)
-		h.Harvest()
+		h.Harvest(reader)
```
urso: why have harvester create reader in setup and finally pass it to Harvest? Isn't the reader stored in harvester?

filebeat/prospector/prospector.go
```diff
@@ -147,6 +137,26 @@ func (p *Prospector) Run() {
 	}
 }
 
+// updateState updates the prospector state and forwards the event to the spooler
+// All state updates done by the prospector itself are synchronous to make sure not states are overwritten
+func (p *Prospector) updateState(event *input.Event) {
+	p.stateMutex.Lock()
+	defer p.stateMutex.Unlock()
```
urso: mutex required for call to `p.states.Update` only?

filebeat/prospector/prospector.go
```diff
@@ -147,6 +137,26 @@ func (p *Prospector) Run() {
...
+
+	// Add ttl if cleanOlder is enabled
+	if p.config.CleanInactive > 0 {
+		event.State.TTL = p.config.CleanInactive
+	}
+
+	ok := p.outlet.OnEvent(event)
```
urso: we want to filter out 'empty' events here or are empty events required for some state update somehwere else?

filebeat/prospector/prospector.go
ruflin: I added an bool to updateState to be returned so the loop can check if event was sent and return accordingly. Same is needed for other calls of updateState.

filebeat/prospector/prospector.go
ruflin: mutex removed as not needed. Already part of states.Update

filebeat/prospector/prospector.go
ruflin: state updates will also be needed for registrar.

filebeat/prospector/prospector.go
ruflin: comment improved

filebeat/prospector/prospector.go
ruflin: Currently reader is not stored in the Harvester but we could.

urso:
LGTM

Issue #1450, opened 2016-04-21
Document supported versions


monicasarbu:
#### Supported versions for the outputs:

- Redis: 
 - v
 - v
- Kafka: 
 - v
 - v

#### Supported versions for the Metricbeat modules:

- Apache: 
 - v
 - v
- Mysql: 
 - v
 - v
- Redis:
 - v
 - v

ruflin:
For apache I would suggest:
* >= 2.2.31
* >= 2.4.16

Test with 2.4.20

@radoondas Does that make sense? Will it work with the 2.2.x versions?

Logic behind this is supporting the releases from the last 6 month and the last version of the 2.2. release as I think it is still quite popular.

ruflin:
For mysql I would suggest:

* >= 5.5.46
* >= 5.6.27
* >= 5.7.10

Tested with 5.7.10

I never tested with 5.5 and 5.6. I will do some manual testing to see if behaves the same.

For MySQL the question that probably pops up is if it also works with MariaDB. I assume it does but I didn't test it so far. Should we mentioned MariaDB?

radoondas:
@ruflin Apache 2.2 is still wildly used and it will take a long time for 2.4 to be the most used. The cycle between major releases is very long.
Implementation support both major versions and I would say, that in case of Apache I would also suggest to simplify it to:

-     = 2.2
-     = 2.4

But we can also take the latest versions of apache as is now at the apache.org and define it always to the latest ones as you suggested.

I noticed that there are new fields in 2.4 and most of the is already implemented. I do review metrics time to time to see if there are changes. Main fields don't change and stay as now.

ruflin:
@radoondas Thanks for the details. There is a difference between "supported" and "should work". The version we mention above mean also that in case of a bug or crash, we fix it. For the others, it probably works, but we will not necessary fix all issues. That is why I try to keep the "official" supported list smaller the then list of actual versions it works with.

You mentioned that there are more fields in 2.4. As we have these fields implemented, could it be that it breaks the 2.2 implementation?

ruflin:
For redis Metricbeat I suggest:

* >= 3.0.5

Tested with Redis 3.0.7.

This is probably the same for the output? @urso ?

radoondas:
@ruflin No, it will not crash 2.2 implementation. Have tested it a lot. 
[...]

PR #2549, opened 2016-09-14
Document supported versions for outputs

dedemorton:
Adds version info as described in https://github.com/elastic/beats/issues/1450

Decided to add a separate section to be consistent with the Metricbeat docs. I'm not sure a compatibility statement warrants a separate section, but it might be nice to have a section where we can add more detail if we decide to support some versions conditionally. The downside is that users might need to scroll down (or click the "compatibility" link) to see supported versions. Does this concern anyone? The other option is to just put a note at the top of the topic right after the heading.

### DIFF
--- a/libbeat/docs/outputconfig.asciidoc
+++ b/libbeat/docs/outputconfig.asciidoc
@@ -54,6 +54,10 @@ output.elasticsearch:
 
 ------------------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


dedemorton:
Documents supported output versions for item tracked in #2482

tsg:
LGTM.

PR #1525, opened 2016-04-28
New syslog output

avleen:
We had a need to have Filebeat send logs to a syslog server, so wrote this output for libbeat :-)

If there are any changes/improvements that I can make to help get this accepted, I'd be happy to do them.

There is possibly one suboptimal part: we're using fields as a way to let users configure 3 syslog parameters.
This could be moved into a separate place, but it didn't seem that bad because the output is sending the original log line, as read, rather than converting it to JSON, so the fields doesn't have any meaning in this context.
And if we made the output format (plain text vs json) configurable, then it may make sense to have this in the fields anyway.

Thanks!

PR #2551, opened 2016-09-14
S3 Output

fritzhardy:
Implements S3 output.  Marries well with logstash S3 input, using S3 as a queuing/archival mechanism.

Stages lines to a local file and uploads at configurable number of bytes (based on fileout).  Uploaded object path/name somewhat resembles CloudTrail logs currently, /somebucket/YYYY/MM/DD/hostname_ISO8601DATE.gz in UTC.

Example configuration:
```
output:
  s3:
    enabled: true
    path: "/var/log/s3"
    filename: s3
    upload_every_kb: 3
    #number_of_files: 2
    bucket: somebucket
```

### DIFF
--- a/dev-tools/packer/docker/xgo-image/base/build.sh
+++ b/dev-tools/packer/docker/xgo-image/base/build.sh
@@ -20,6 +20,7 @@
 
 # Download the canonical import path (may fail, don't allow failures beyond)
 SRC_FOLDER=$SOURCE
+AWS_FOLDER=$GOPATH/src/github.com/aws
 
 BEAT_PATH=$1
 DST_FOLDER=`dirname $GOPATH/src/$BEAT_PATH`
@@ -53,6 +54,12 @@ else
         cd $GOPATH/src/${GIT_REPO}
         echo "Fetching main git repository ${GIT_REPO} in folder $GOPATH/src/${GIT_REPO}"
         git clone https://${GIT_REPO}.git
+fi
+
+if [ "$AWSSDK" != '' ]; then
+	mkdir -p ${AWS_FOLDER}
+	echo "Copying aws-sdk folder ${AWSSDK} to folder ${AWS_FOLDER}"
+	rsync --exclude ".git" -a ${AWSSDK}/ ${AWS_FOLDER}
 fi
 
 set -e
--- a/libbeat/outputs/s3/config.go
+++ b/libbeat/outputs/s3/config.go
@@ -1,0 +1,31 @@
+package s3
+
+import (
+	"fmt"
+)
+
+type config struct {
+	Path          string `config:"path"`
+	Filename      string `config:"filename"`
+	UploadEveryKb int    `config:"upload_every_kb" validate:"min=1"`
+	NumberOfFiles int    `config:"number_of_files"`
+	Region        string `config:"region"`
+	Bucket        string `config:"bucket"`
+}
+
+var (
+	defaultConfig = config{
+		NumberOfFiles: 2,
+		UploadEveryKb: 10 * 1024,
+		Region:        "us-east-1",
+	}
+)
+
+func (c *config) Validate() error {
+	if c.NumberOfFiles < 2 || c.NumberOfFiles > managerMaxFiles {
+		return fmt.Errorf("S3 number_of_files to keep should be between 2 and %v",
+			managerMaxFiles)
+	}
+
+	return nil
+}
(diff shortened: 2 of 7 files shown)


karmi:
Hi @fritzhardy, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/beats/pull/2551.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

ruflin:
@fritzhardy Thanks a lot for all your work on the S3 output and providing directly a PR. Unfortunately we currently do not plan to add further outputs as each output does add additional work on the maintenance and support side. We are a small team and currently focused on other enhancements on the beats side.

For additional outputs we recommend to keep them in a separate repository without maintaining a full fork and add them during compile time. For further details see the comment here:  https://github.com/elastic/beats/pull/1525#issuecomment-217651768 Like this outputs work almost like pPlugins. Perhaps one day Golang will support it, which will make adding external outputs even simpler.

We are always happy to help with questions or reviewing code. Feel free to ping us on discuss at any time: https://discuss.elastic.co/c/beats

ruflin:
Based on the comment above, I'm closing this PR. @fritzhardy Feel free to use this PR for further discussions.

fritzhardy:
I think it would be a boon to the project to have this as an official addition at some point.  We find this approach gives us "networkless" log forwarding at little expense, and on the logstash side, all the configurability necessary to deal with a bucket of logs.  However, I am aware that additional output plugins have been a source of discussion.  In the meantime, I will review https://github.com/elastic/beats/pull/1525#issuecomment-217651768.  I was looking at some way to do that externally in the first place, akin to the external beats.  Thanks for your time and feedback.

trixpan:
Got to this while reading #1525...

@fritzhardy the straight to persistence output is indeed a clever approach which is not so widely used.

However, from what I gather, ( @ruflin can correct me if I am wrong) Elastic community is trying to avoid the inclusion of too many additional transport feature as there end up being hard do manage the code base in the long run. To be honest I think their approach is fair call.


Having said that, projects like Apache MiNiFi allow you to do this by simply adding existing processors (PutS3 in this case) to the MiNiFi install but at the cost of running a lightweight JVM @ the edge.

Disclaimer: I am a NiFi committer.

ruflin:
@trixpan We are actually hoping that Golang Plugins become a reality in the near future and this would directly solve the above issue.

About the approach you mentioned: Does that mean in the end 2 processes are running? MiNiFi + PutS3?

trixpan:
@ruflin yes. GoLang plugins will be handy. Will be happy to have a go trying to add NiFi site 2 site support to filebeat when your feature is stable.

Regarding MiNiFi + PutS3: Just a single JVM is executed, MiNiFi (the framework). Whitin MiNiFi we run the PutS3 processor (think of a processor as a Input/Codec/Filter/Output in logstash terms). 

To certain extend, MiNiFi as a stripped version of the overall framework (NiFi) and as such is able to reutilise processors of the main code base as illustrated here:

https://github.com/apache/nifi-minifi/blob/master/minifi-nar-bundles/minifi-standard-nar/pom.xml#L42

To cater for the S3 usecase, user would have to add NiFi NARs (nifi-aws-bundle) into the minifi install package (i.e. copy into the adequate folder) and use the installed processors (PutS3) as (s)he would normally do when using the main NiFi platform. Ironically enough, if I recall correctly outputting to a central NiFi is itself would be optional. 

In ELK terms, this would be akin to running a minimalist version of Logstash at the producer level, reusing some of the jRuby code that powers Logstash, instead of developing a golang producer (i.e. lumberjack-forwarder / *beats) from scratch.  There are pros and cons on both approaches and no wonder we are also building minifi-cpp. :smiley:

ktham:
@ruflin We're looking at dumping our logs into S3 and would like to see S3 as an output for filebeats. Looks like there was a previous effort here in this PR to add this. What is your recommendation here? Is it possible to support S3 output?

fritzhardy:
We have been using this fork in production for two years.  It needs some polish, but gets the job done.

ktham:
I see, given that it's been working without issue for the past 2 years, I'd like to see if it's possible to include this output in the upstream filebeats project, or if there's a way to factor this into a plug-in so as to avoid using a fork of filebeats.

ktham:
@ruflin/@fritzhardy Can we revisit the topic of adding an S3 output? We are willing to help with adding the code if necessary.

Issue #1742, opened 2016-05-26
Do not start Filebeat if registry file not writable


ruflin:
Currently if Filebeat is started up and the registry file cannot be written it logs an error but continues working. This means the state can never be persisted to disk.

On startup, filebeat should check, if the registry file can be created (if it doesn't exist yet) or if it is writable by the user started filebeat. If not, Filebeat should exit.

PR #2552, opened 2016-09-15
Prevent filebeat from starting on error with registry file

ruflin:
Filebeat was starting also if there were problems with the registry file. The problems could have been:

* Registry file is a directory
* Registry file is not writable
* Invalid state in registry file
* Registry file is a symlink

All this cases are now checked and in case one of the checks fails, filebeat is stopped and a descriptive error is returned.

Symlinks are not allowed as filebeat must be able to write to the file and on windows also does some renaming magic which causes issues.

Closes https://github.com/elastic/beats/issues/1742

### DIFF
--- a/filebeat/registrar/registrar.go
+++ b/filebeat/registrar/registrar.go
@@ -59,8 +59,32 @@ func (r *Registrar) Init() error {
 	registryPath := filepath.Dir(r.registryFile)
 	err := os.MkdirAll(registryPath, 0755)
 	if err != nil {
-		return fmt.Errorf("Failed to created registry file dir %s: %v",
-			registryPath, err)
+		return fmt.Errorf("Failed to created registry file dir %s: %v", registryPath, err)
+	}
+
+	// Check if files exists
+	fileInfo, err := os.Lstat(r.registryFile)
+	if os.IsNotExist(err) {
+		logp.Info("No registry file found under: %s. Creating a new registry file.", r.registryFile)
+		// No registry exists yet, write empty state to check if registry can be written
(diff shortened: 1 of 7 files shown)


ruflin:
jenkins, retest it

filebeat/registrar/registrar.go
```diff
@@ -59,8 +59,32 @@ func (r *Registrar) Init() error {
...
-			registryPath, err)
+		return fmt.Errorf("Failed to created registry file dir %s: %v", registryPath, err)
+	}
+
+	// Check if files exists
+	fileInfo, err := os.Lstat(r.registryFile)
+	if os.IsNotExist(err) {
+		logp.Info("No registry file found under: %s. Creating a new registry file.", r.registryFile)
+		// No registry exists yet, write empty state to check if registry can be written
+		err = r.writeRegistry()
```
andrewkroh: Simplify to `return r.writeRegistry()`.

filebeat/registrar/registrar.go
```diff
@@ -259,9 +269,11 @@ func (r *Registrar) writeRegistry() error {
 	// Directly close file because of windows
 	f.Close()
```
andrewkroh: If an error occurs while encoding then the temp file is not closed. Add `f.Close()` to the `if` block above.

filebeat/registrar/registrar.go
ruflin: done

filebeat/registrar/registrar.go
ruflin: done

PR #2554, opened 2016-09-15
Fix the type of some blkio fields

monicasarbu:
Some fields from the system.process.cgroups.blkio group were defined as string instead of long.


tsg:
jenkins, retest it

andrewkroh:
LGTM

Issue #2556, opened 2016-09-15
Add "version" to 5.x templates


pickypg:
Templates (and pipelines) now support a `"version"` integer field that can added to them for the type of need that Logstash has: when to replace the existing template. (https://github.com/elastic/elasticsearch/issues/20171)

We should add the `"version"` to the existing 5.x Beats template, which should allow us to simplify replacing it, as needed, in the future as well as the older templates.

``` http
GET /_template/my_template?filter_path=*.version
```

This will return an empty JSON object for non-versioned templates.

``` json
{}
```

and it will return whatever version is set for templates that do set it:

``` json
{"my_template":{"version":20160915}}
```

ruflin:
We introduced a version in our template very recently as a meta field: https://github.com/elastic/beats/blob/master/metricbeat/metricbeat.template.json#L8 It is great to see that this is now supported natively. I will adjust our script to put the version now on the top level. Thanks for pinging us on this one.

PR #2567, opened 2016-09-16
Use new version for templates

ruflin:
- [ ] Test that shows up in mapping most recent elasticsearch versions
- [ ] Use _meta version for 2.x
- [ ] Test on 2.x

Closes https://github.com/elastic/beats/issues/2556

### DIFF
--- a/filebeat/filebeat.template-es2x.json
+++ b/filebeat/filebeat.template-es2x.json
@@ -6,9 +6,6 @@
           "enabled": false
         }
       },
-      "_meta": {
-        "version": "6.0.0-alpha1"
-      },
       "dynamic_templates": [
         {
           "fields": {
@@ -77,5 +74,6 @@
   "settings": {
     "index.refresh_interval": "5s"
   },
-  "template": "filebeat-*"
+  "template": "filebeat-*",
+  "version": "6.0.0-alpha1"
 }
--- a/filebeat/filebeat.template.json
+++ b/filebeat/filebeat.template.json
@@ -4,9 +4,6 @@
       "_all": {
         "norms": false
       },
-      "_meta": {
-        "version": "6.0.0-alpha1"
-      },
       "dynamic_templates": [
         {
           "fields": {
@@ -65,5 +62,6 @@
   "settings": {
     "index.refresh_interval": "5s"
   },
-  "template": "filebeat-*"
+  "template": "filebeat-*",
+  "version": "6.0.0-alpha1"
 }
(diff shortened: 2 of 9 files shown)


filebeat/filebeat.template-es2x.json
```diff
@@ -77,5 +74,6 @@
   "settings": {
     "index.refresh_interval": "5s"
   },
-  "template": "filebeat-*"
+  "template": "filebeat-*",
+  "version": "6.0.0-alpha1"
```
andrewkroh: Should this still be under `_meta.version` for 2.x?

ruflin: yes, see Todo in the PR ;-) I have to add some more additional logic.

pickypg:
The version field is solely an integer which is intended to make comparisons fast and easy. See https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/478

I suggest only versioning when it matters too; unused meta fields excluded. The 5.x and 6.x templates should all share the same version as a result (feel free to mess with the meta fields, but then you'll know when the template was originally added).

ruflin:
Oh, it is only an integer. Because of metricbeat and the metricbeat our template changes quite frequently. So far I was thinking that the version corresponds with beat version. LS seems to go the approach to have ES_VERSION + incremental value. Need to think a little bit more about this on what makes most sense in our case.

pickypg:
For the forthcoming Monitoring templates I just use a date. But that won't work if you update older templates later than the newer one(s).

ruflin:
This are the requirements I see from the beats side:

* Differentiation between all versions (bugfix, minor, major). Example 5.1.2 to 5.1.3. The reasons is that our template could also change in a bug fix release
* Non time based differentiation: It can be that 6.0.1 is shipped before 5.3.2. But if someone updates from 5.3.2 to 6.0.1 still the new template should be applied.
* Multiple templates per version: Each beat ships with a template for 5.x and 2.x currently and the templates are not identical.

For beats it seems more important that someone uses the exact template for his beat version and elasticsearch version as templates are not necessarly backward compatible. This means it would be very nice if the version would contain beat AND es version.

Nice to have for development would be that something like the `md5` some of our template generation file as version id, as it would make it very obvious when a version change. But this is only nice to have.

I need to think of a logic on how the above could be converted to an integer and still be reasonable human readable.

ruflin:
I'm closing this PR as the versioning of index templates will need more thoughts on the beats side. Potentially there will be templates for each module / metricset and only this part of a template is loaded. Each of these templates should have its own version. With the current version meta field we have what we need to detect which version a document was created with so there is no urgent need to switch to the new field especially as the current implementation works with 2.x and 5.x (and probably 1.x).

Issue #1064, opened 2016-02-29
Add README to downloaded packages


ruflin:
The beats packages downloaded from https://www.elastic.co/downloads/beats as `tar.gz`, `tgz` or `zip` should contain a README that contains the basic information on how to get started. This README must be platform specific and added to the package during package time.

If possible this information should be generated based on a template for each beat.

ruflin:
See https://github.com/elastic/beats/pull/2570

PR #2570, opened 2016-09-16
Add README.md to beats packages

ruflin:
* Add README.md to homedir to be part of each package
* Update description of each beat to make more sense in this context
* Add version.yml to package.yml file so it doesn't have to be added in each build

Closes https://github.com/elastic/beats/issues/1064

### DIFF
--- a/dev-tools/packer/platforms/binary/build.sh
+++ b/dev-tools/packer/platforms/binary/build.sh
@@ -8,7 +8,7 @@ ARCHDIR=${BASEDIR}/../../
 # executed from the top directory
 runid=binary-$BEAT-$ARCH
 
-cat ${BUILD_DIR}/package.yml ${ARCHDIR}/archs/$ARCH.yml ${ARCHDIR}/version.yml > ${BUILD_DIR}/settings-$runid.yml
+cat ${BUILD_DIR}/package.yml ${ARCHDIR}/archs/$ARCH.yml > ${BUILD_DIR}/settings-$runid.yml
 gotpl ${BASEDIR}/run.sh.j2 < ${BUILD_DIR}/settings-$runid.yml > ${BUILD_DIR}/run-$runid.sh
 chmod +x ${BUILD_DIR}/run-$runid.sh
 
--- a/dev-tools/packer/readme.md.j2
+++ b/dev-tools/packer/readme.md.j2
@@ -1,0 +1,23 @@
+# Welcome to {{.beat_name}} {{.version}}
+
+{{.beat_description}}
+
+## Getting Started
+
+To get started with {{.beat_name}}, you need to set up Elasticsearch on your localhost first. After that, start {{.beat_name}} with:
+
+     ./{{.beat_name}}  -c {{.beat_name}}.yml -e
+
+This will start the beat and send the data to your Elasticsearch instance. To load the dashboards for {{.beat_name}} into Kibana, run:
+
+    ./scripts/import_dashboards
+
+For further steps visit the [Getting started](https://www.elastic.co/guide/en/beats/{{.beat_name}}/{{.doc_branch}}/{{.beat_name}}-getting-started.html) guide.
+
+## Documentation
+
+Visit [Elastic.co Docs](https://www.elastic.co/guide/en/beats/{{.beat_name}}/{{.doc_branch}}/index.html) for the full {{.beat_name}} documentation.
+
+## Release notes
+
+https://www.elastic.co/guide/en/beats/libbeat/{{.doc_branch}}/release-notes-{{.version}}.html
--- a/packetbeat/Makefile
+++ b/packetbeat/Makefile
@@ -1,6 +1,6 @@
 
 BEATNAME?=packetbeat
-BEAT_DESCRIPTION?=A network analyzer tool that sends data to Elasticsearch.
+BEAT_DESCRIPTION?=Packetbeat analyzes network traffic and sends the data to Elasticsearch.
 SYSTEM_TESTS=true
 TEST_ENVIRONMENT=false
 TARGETS="windows/amd64 windows/386 darwin/amd64"
--- a/dev-tools/packer/platforms/centos/build.sh
+++ b/dev-tools/packer/platforms/centos/build.sh
@@ -8,7 +8,7 @@ ARCHDIR=${BASEDIR}/../../
 # executed from the top directory
 runid=centos-$BEAT-$ARCH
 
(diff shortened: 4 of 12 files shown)


dev-tools/packer/readme.md.j2
```diff
@@ -0,0 +1,23 @@
...
+
+{{.beat_description}}
+
+## Getting Started
+
+To get running with your beat, setup elasticsearch on your localhost and start it with:
+
+     ./{{.beat_name}}  -c {{.beat_name}}.yml -e
+
+This will start the beat and send to do your elasticsearch instance. To load the dashboards for {{.beat_name}} into Kibana, run:
```
andrewkroh: `s/send to do/send to/`

packetbeat/Makefile
```diff
@@ -1,6 +1,6 @@
 
 BEATNAME?=packetbeat
-BEAT_DESCRIPTION?=A network analyzer tool that sends data to Elasticsearch.
+BEAT_DESCRIPTION?=Packetbeat: A network analyzer tool that sends data to Elasticsearch.
```
andrewkroh: Since it's not consistent with the other descriptions, how about "Packetbeat analyzes network traffic and sends the data to Elasticsearch."

dev-tools/packer/readme.md.j2
ruflin: I adjusted it to: This will start the beat and send the data to your elasticsearch instance.

packetbeat/Makefile
ruflin: done

ruflin:
jenkins, package it

dev-tools/packer/readme.md.j2
```diff
@@ -0,0 +1,23 @@
+# Welcome to {{.beat_name}} {{.version}}
+
+{{.beat_description}}
+
+## Getting Started
```
tsg: We should check with @dedemorton how she feels about having this lightweight getting started guide in here, vs just a link to the docs getting started.

dev-tools/packer/readme.md.j2
```diff
@@ -0,0 +1,23 @@
...
+
+To get running with your beat, setup elasticsearch on your localhost and start it with:
```
tsg: Capital "Elasticsearch". Also the way it's written know it sounds a bit like that is the command to start ES.

dev-tools/packer/readme.md.j2
```diff
@@ -0,0 +1,23 @@
...
+
+     ./{{.beat_name}}  -c {{.beat_name}}.yml -e
+
+This will start the beat and send the data to your elasticsearch instance. To load the dashboards for {{.beat_name}} into Kibana, run:
```
tsg: capitalize "Elasticsearch" please.

dev-tools/packer/readme.md.j2
```diff
@@ -0,0 +1,23 @@
...
+
+    ./scripts/import_dashboards
+
+For further steps visit the [Getting started](https://www.elastic.co/guide/en/beats/{{.beat_name}}/current/{{.beat_name}}-getting-started.html) guide.
```
tsg: Here it's tricky, because "current" is not always correct. For example, if you are on version 5.0.0-beta1, the link will point to the 1.3 documentation.

dev-tools/packer/readme.md.j2
```diff
@@ -0,0 +1,23 @@
...
+
+## Documentation
+
+Visit [Elastic.co Docs](https://www.elastic.co/guide/en/beats/{{.beat_name}}/current/index.html) for the full {{.beat_name}} documentation.
```
tsg: Same comment as the above, current might be wrong.

dev-tools/packer/readme.md.j2
```diff
@@ -0,0 +1,23 @@
...
+
+## Release notes
+
+https://www.elastic.co/guide/en/beats/libbeat/current/release-notes-{{.version}}.html
```
tsg: Also here.

dev-tools/packer/platforms/binary/build.sh
```diff
@@ -8,7 +8,7 @@ ARCHDIR=${BASEDIR}/../../
 # executed from the top directory
 runid=binary-$BEAT-$ARCH
 
-cat ${BUILD_DIR}/package.yml ${ARCHDIR}/archs/$ARCH.yml ${ARCHDIR}/version.yml > ${BUILD_DIR}/settings-$runid.yml
+cat ${BUILD_DIR}/package.yml ${ARCHDIR}/archs/$ARCH.yml > ${BUILD_DIR}/settings-$runid.yml
```
tsg: Are these changes required by the PR or just cleanup?

ruflin: They are indirectly required as I added version.yml to package.yml as I already need it in before_build.sh. The alternative would be to create an additional .yml file in before_build but as we always need the version, would only add complexity.

ruflin:
About the generation: I would prefer if this file is not generated by make update as otherwise it would probably have to be checked in. 

I wasn't aware that we have the doc-branch in version.asciidoc. Will try to do some magic to get the info from there for the urls.

@dedemorton Before I do additional work, can you comment in general if you see such a file helpful?

dev-tools/packer/readme.md.j2
dedemorton: I'm fine with this because it points users to the doc. I'm not sure how you will deal with versioning, though.

dev-tools/packer/readme.md.j2
dedemorton: ..."set up Elasticsearch" ("setup" should be "set up" when you're using it as a verb). :-)

dedemorton:
I think it's a good idea.

dev-tools/packer/readme.md.j2
ruflin: done

dev-tools/packer/readme.md.j2
ruflin: changed

ruflin:
The script is now adjusted to fetch the branch info from the docs version. Now all links work except the one for the release notes in master (as there are none). 

@tsg Is this worth backporting to 5.0 or should we start shipping this with 5.1?

ruflin:
jenkins, package it

ruflin:
jenkins, package it

ruflin:
jenkins, retest it

ruflin:
jenkins, package it

PR #2573, opened 2016-09-16
[WIP] go http based memory profiling metric set

urso:


ruflin:
Nice :-) Should we call it just `golang` (or go) module as the metricset is then memprof etc?

urso:
I'm not planing to continue this anytime soon.

medcl:
@urso link to this issue, https://github.com/elastic/beats/issues/3466

PR #2576, opened 2016-09-17
Edit topics about creating new dashboards

dedemorton:
Summary of changes:
- Formatted CLI options and changed the order to match how they appear when you run `-h` (makes it a lot easier to see when we've left something out).
- Divided topic about importing dashboards into sections to make it easier for non-developers to figure out what is relevant to them. Why? Because point end users to this section for more info about importing dashboards, I want them to be able to find the stuff that's relevant to them quickly.
- Did some light edits for consistency. In command examples, removed `$` because we've gotten feedback from developers saying that it makes copy/paste harder when we include `$` in our examples.

### DIFF
--- a/libbeat/docs/newdashboards.asciidoc
+++ b/libbeat/docs/newdashboards.asciidoc
@@ -42,7 +42,7 @@ There are a couple of common use cases for importing dashboards:
 [[import-single-beat]]
 ==== Import Dashboards and/or the Index Pattern for a Single Beat
 
-Using the `import_dashboards` script, you can import the dashboards and the index pattern to
+Using the `import_dashboards` script from the Beat package, you can import the dashboards and the index pattern to
 Elasticsearch running on localhost for a single Beat (eg. Metricbeat):
 
 - from a local directory:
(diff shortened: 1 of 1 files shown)


dedemorton:
@monicasarbu These are the structural changes that I mentioned I would be making eventually. Please review.

dedemorton:
This PR adds edits for work tracked in https://github.com/elastic/beats/issues/2482

monicasarbu:
LGTM. I only a few comments due to the latest changes. The PR needs to be rebased.

libbeat/docs/newdashboards.asciidoc
```diff
@@ -31,182 +31,198 @@ The following topics provide more detail about importing and working with Beats
...
 
 Using the `import_dashboards` script, you can import the dashboards and the index pattern to
 Elasticsearch running on localhost for a single Beat (eg. Metricbeat):
 
 - from a local directory:
-
++
 [source,shell]
 ----------------------------------------------------------------------
-$ ./scripts/import_dashboards -dir kibana/metricbeat
```
dedemorton: @monicasarbu Yes. I removed it for a couple of reasons. We weren't using it consistently. Also, we've heard from customers (specifically, for cloud) that the $ makes it harder to copy/paste commands. Granted...we're not consistent in the doc, but we should move towards removing it. If we think that showing the $ adds value, then we should handle it as part of the formatting that happens in the doc build when we've tagged something as shell.

dedemorton:
@monicasarbu I've rebased and fixed the issues that you identified. I also noticed that the section titles were a bit inconsistent, so I fixed that, too.

Issue #2490, opened 2016-09-08
Beat panics if log directory cannot be created


ruflin:
Version: master
Steps to reproduce:
* Put log directory in `logging.files.path:` where directory cannot be created by filebeat (no permissions)
* Start beat

Panic output
```
2016/09/08 11:09:48.485527 beat.go:285: CRIT Exiting: error initializing logging: mkdir /var/log/filebeat: permission denied
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xa4108]

goroutine 1 [running]:
panic(0x4cc680, 0xc4200120b0)
       	/usr/local/Cellar/go/1.7/libexec/src/runtime/panic.go:500 +0x1a1
github.com/elastic/beats/libbeat/logp.(*FileRotator).Rotate(0xc42016e4c0, 0xc420164930, 0xc4201648c0)
       	/Users/ruflin/Dev/gopath/src/github.com/elastic/beats/libbeat/logp/file_rotator.go:118 +0x48
github.com/elastic/beats/libbeat/logp.(*FileRotator).WriteLine(0xc42016e4c0, 0xc420164930, 0x6e, 0x70, 0x6e, 0x70)
       	/Users/ruflin/Dev/gopath/src/github.com/elastic/beats/libbeat/logp/file_rotator.go:64 +0x48
github.com/elastic/beats/libbeat/logp.send(0x4, 0x2, 0x56080a, 0x5, 0x566499, 0xb, 0xc42004bea8, 0x1, 0x1)
       	/Users/ruflin/Dev/gopath/src/github.com/elastic/beats/libbeat/logp/log.go:64 +0x1c0
github.com/elastic/beats/libbeat/logp.msg(0x2, 0x56080a, 0x5, 0x566499, 0xb, 0xc42004bea8, 0x1, 0x1)
       	/Users/ruflin/Dev/gopath/src/github.com/elastic/beats/libbeat/logp/log.go:84 +0x9d
github.com/elastic/beats/libbeat/logp.Critical(0x566499, 0xb, 0xc42004bea8, 0x1, 0x1)
       	/Users/ruflin/Dev/gopath/src/github.com/elastic/beats/libbeat/logp/log.go:101 +0x71
github.com/elastic/beats/libbeat/beat.handleError(0x7c5080, 0xc420174370, 0x7c5080, 0xc420174370)
[...]

PR #2583, opened 2016-09-19
Backport: Fix panic if logging directory cannot be generated (#2571)

tsg:
Backport of #2571. Original message:

* Assigns log rotator only after successful creation. Otherwise panic can happen during shutdown if directory cannot be generated as it tries to log.

Closes https://github.com/elastic/beats/issues/2490


tsg:
The Jenkins failure seems not related.

PR #2584, opened 2016-09-19
Rename / restructure input / output setup in filebeat

ruflin:
Naming convention:
- Everything using a channel to send data is called input / output with a Send method
- If only confirmation of events happens, it is called a Logger with a log method

More cleanup is needed in future PRs.

This replaces https://github.com/elastic/beats/pull/2511

### DIFF
--- a/filebeat/beater/channels.go
+++ b/filebeat/beater/channels.go
@@ -5,46 +5,43 @@ import (
 	"sync/atomic"
 
 	"github.com/elastic/beats/filebeat/input"
-	"github.com/elastic/beats/filebeat/registrar"
-	"github.com/elastic/beats/filebeat/spooler"
 )
 
-type spoolerOutlet struct {
-	wg      *sync.WaitGroup
-	done    <-chan struct{}
-	spooler *spooler.Spooler
-
+type prospectorOutput struct {
+	wg     *sync.WaitGroup
+	done   <-chan struct{}
+	input  chan *input.Event
 	isOpen int32 // atomic indicator
 }
 
-type publisherChannel struct {
+type spoolerOutput struct {
 	done chan struct{}
 	ch   chan []*input.Event
 }
 
-type registrarLogger struct {
+type publisherOutput struct {
 	done chan struct{}
 	ch   chan<- []*input.Event
 }
 
-type finishedLogger struct {
+type logger struct {
 	wg *sync.WaitGroup
 }
 
-func newSpoolerOutlet(
+func newProspectorOutput(
 	done <-chan struct{},
-	s *spooler.Spooler,
+	input chan *input.Event,
 	wg *sync.WaitGroup,
-) *spoolerOutlet {
-	return &spoolerOutlet{
-		done:    done,
-		spooler: s,
-		wg:      wg,
-		isOpen:  1,
+) *prospectorOutput {
+	return &prospectorOutput{
+		done:   done,
+		input:  input,
+		wg:     wg,
+		isOpen: 1,
 	}
 }
 
-func (o *spoolerOutlet) OnEvent(event *input.Event) bool {
+func (o *prospectorOutput) Send(event *input.Event) bool {
 	open := atomic.LoadInt32(&o.isOpen) == 1
 	if !open {
 		return false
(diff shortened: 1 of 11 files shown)


filebeat/beater/filebeat.go
```diff
@@ -98,9 +99,9 @@ func (fb *Filebeat) Run(b *beat.Beat) error {
 	publisher.Start()
 	// Stopping publisher (might potentially drop items)
 	defer func() {
-		// Closes first the registrar logger to make sure not more events arrive at the registrar
-		// registrarChannel must be closed first to potentially unblock (pretty unlikely) the publisher
-		registrarChannel.Close()
+		// Closes first the publisher outpt to make sure not more events are sent to the registry
```
tsg: tiny typo: "outpt".

filebeat/publisher/publisher.go
```diff
@@ -19,23 +19,22 @@ type LogPublisher interface {
 	Stop()
 }
 
-// SuccessLogger is used to report successfully published events.
-type SuccessLogger interface {
-
+// Logger is used to report successfully published events.
+type Output interface {
```
tsg: Type is Output, the comment says "Logger".

tsg:
TBH these naming conventions make the code more abstract to me. But I defer to you and @urso on what's best here.

ruflin:
@tsg I actually hoped to reach the opposite especially for the variable names. They should now say much more what they are then it was before. For the naming conventions, the problem before was that there was Outlet / Output but not really a difference and the naming was not consistent. Any specific parts you stumbled over which are now too abstract?

filebeat/beater/filebeat.go
ruflin: fixed

filebeat/publisher/publisher.go
ruflin: done

filebeat/beater/channels.go
```diff
@@ -5,46 +5,43 @@ import (
...
-type finishedLogger struct {
+type logger struct {
 	wg *sync.WaitGroup
 }
 
-func newSpoolerOutlet(
+func newProspectorOutput(
 	done <-chan struct{},
-	s *spooler.Spooler,
+	input chan *input.Event,
```
urso: while this one uses the spooler channel  only, idea of spoolerOutlet was to wrap/hide the spooler and the fact we're dealing with a channel. I'd even prefer in future calling a spooler.OnEvent directly, hiding the fact we're dealing with an asynchronous worker.

e.g. have a synchronous spooler worker with 'flush' upon full + potential async wrapper:

```
type spooler struct {}
type asyncSpooler struct {}

func (s *spooler) OnEvent()
func (s *spooler) Flush()

func newAsyncSpooler(s *spooler) *asyncSpooler
func (s *asyncSpooler) OnEvent()
func (s *asyncSpooler) Stop()

```

Both types support OnEvent, so no one needs to know about channels here.

filebeat/beater/channels.go
```diff
@@ -5,46 +5,43 @@ import (
...
 
-type spoolerOutlet struct {
-	wg      *sync.WaitGroup
-	done    <-chan struct{}
-	spooler *spooler.Spooler
-
+type prospectorOutput struct {
+	wg     *sync.WaitGroup
+	done   <-chan struct{}
+	input  chan *input.Event
```
urso: see comment on `newProspectorOutput`

filebeat/beater/channels.go
```diff
@@ -5,46 +5,43 @@ import (
...
 }
 
-type registrarLogger struct {
+type publisherOutput struct {
 	done chan struct{}
 	ch   chan<- []*input.Event
 }
 
-type finishedLogger struct {
+type logger struct {
```
urso: while `finishedLogger` is a bad naming, I wouldn't call it `logger` either, as logger somehow implies this is about logging a message to file.

filebeat/beater/filebeat.go
```diff
@@ -51,32 +51,33 @@ func (fb *Filebeat) Run(b *beat.Beat) error {
...
+	publisherOutput := newPublisherOutput(registrar.GetInput())
 
-	// Channel from spooler to harvester
-	publisherChan := newPublisherChannel()
+	// Channel from spooler to publisher
+	spoolerOutput := newSpoolerOutput()
 
 	// Publishes event to output
-	publisher := publisher.New(config.PublishAsync, publisherChan.ch, registrarChannel, b.Publisher)
+	publisher := publisher.New(config.PublishAsync, spoolerOutput.ch, publisherOutput, b.Publisher)
```
urso: here I'd prefer having something like `registrarSink`... or directly have a call like:

```
publisher.New(config.PublishAsync, pubEventSource(spoolerOutput), onFinished(registrar), b.Publisher)
```

to make more clear how events flow through the system.

filebeat/beater/filebeat.go
```diff
@@ -51,32 +51,33 @@ func (fb *Filebeat) Run(b *beat.Beat) error {
...
 
 	// Init and Start spooler: Harvesters dump events into the spooler.
-	spooler, err := spooler.New(config, publisherChan)
+	spooler, err := spooler.New(config, spoolerOutput)
```
urso: idea of using `publisherChan` (while naming is bad), was to make more clear which component the spooler outputs to.

filebeat/prospector/prospector.go
```diff
@@ -38,15 +38,15 @@ type Prospectorer interface {
 	Run()
 }
 
-type Outlet interface {
-	OnEvent(event *input.Event) bool
+type Output interface {
+	Send(event *input.Event) bool
 }
```
urso: this is not necessarily `Output`, but more like an `Eventer` with `OnEvent` method?

ruflin:
After a discussion with @urso we decided to close this PR and instead open several smaller PR's that hide the channels from `filebeat.go` so no need anymore for the naming.

PR #2585, opened 2016-09-19
Make it clear that 'any' only works on Linux

tsg:
The previous wording in the default config make it look like 'any'
would work on OS X & Windows, but it doesn't.


tsg:
Test failure not related.

PR #2589, opened 2016-09-19
Recommend to share the community Beat dashboards

monicasarbu:
Add a page in the documentation to list all the Beats dashboards created by the community. This is similar with the community Beats.
cc-ed @dedemorton

### DIFF
--- a/libbeat/docs/newdashboards.asciidoc
+++ b/libbeat/docs/newdashboards.asciidoc
@@ -25,6 +25,7 @@ The following topics provide more detail about importing and working with Beats 
 * <<generate-index-pattern>>
 * <<export-dashboards>>
 * <<archive-dashboards>>
+* <<share-beat-dashboards>>
 
 [[import-dashboards]]
 === Importing Existing Beat Dashboards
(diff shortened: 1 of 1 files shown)


libbeat/docs/communitydashboards.asciidoc
```diff
@@ -0,0 +1,31 @@
+[[community-beat-dashboards]]
+== Community Beat dashboards
+
+The open source community has been hard at work developing new Beats. You can check
```
dedemorton: Hmmm...not sure this intro makes sense, even if you change it to say :"The open source community has been hard at work developing new Beats dashboards." Maybe we could say something simple like, "To help you visualize data in Kibana, here are some dashboards contributed by the Beats development community." 

Since the list is empty right now, it sounds a bit odd to say that people have been hard at work. :-) 

I understand that you want to have the link active, but it's too bad that we can't seed this list with a couple of dashboards so contributors will have an example. Or maybe that's possible?

monicasarbu: No, there is none for now. I know it's a bit strange. Maybe we can add ... be the first to contribute ...

monicasarbu: I agree with you @dedemorton, and I suggest for now to recommend users to create a topic on discuss when they want to share their own Beat dashboards.

monicasarbu: I updated this PR with the suggested changes.

PR #2594, opened 2016-09-19
Fix go routine leak which lead to memory leak

ruflin:
The internal goroutine used to stop fileReader and for close_timeout was not stoppped when a harvester was finished which lead to a memory leak and leak of go routines. The goroutine is now properly stopped and resources are freed

### DIFF
--- a/filebeat/harvester/log.go
+++ b/filebeat/harvester/log.go
@@ -258,6 +258,9 @@ func (h *Harvester) getState() file.State {
 }
 
 func (h *Harvester) close() {
+	// stops all go routines inside the harvester
+	close(h.harvesterDone)
+
 	// Mark harvester as finished
 	h.state.Finished = true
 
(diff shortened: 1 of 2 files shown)


tsg:
The Jenkins failure is not FIlebeat related. The Appveyor one is, but could be related to the clean_remove default change, rather that this one.

tsg:
I re-triggered the appveyor test to see if it's flaky or consistently broken.

filebeat/harvester/log.go
```diff
@@ -309,16 +312,18 @@ func (h *Harvester) newLogFileReader() (reader.Reader, error) {
...
 		if h.config.CloseTimeout > 0 {
 			closeTimeout = time.After(h.config.CloseTimeout)
 		}
 
 		select {
-		case <-h.done:
 		case <-closeTimeout:
 			logp.Info("Closing harvester because close_timeout was reached: %s", h.state.Source)
+		case <-h.done:
+		case <-h.harvesterDone:
```
urso: wonder if it would make more sense to have this channel local + start go-routine in worker loop itself (Harvest method). e.g. like:

```
harvesterDone := make(chan struct{})
defer close(harvesterDone)

go func() {
   ...
}
```

What's the big difference between `done` and `harvesterDone`? We still need to check `h.done` here then?

PR #2595, opened 2016-09-19
Fix go routine leak which lead to memory leak

ruflin:
The internal goroutine used to stop fileReader and for close_timeout was not stoppped when a harvester was finished which lead to a memory leak and leak of go routines. The goroutine is now properly stopped and resources are freed


ruflin:
jenkins, retest it

PR #2597, opened 2016-09-19
Update output config docs

urso:
Update output config docs to match most recent changes in all outputs removing
settings not being available, updating docs for changed fields and adding
missing fields.

Add output plugin settings for: #2482 and #2074

### DIFF
--- a/libbeat/docs/outputconfig.asciidoc
+++ b/libbeat/docs/outputconfig.asciidoc
@@ -106,10 +106,6 @@ The number of workers per configured host publishing events to Elasticsearch. Th
 is best used with load balancing mode enabled. Example: If you have 2 hosts and
 3 workers, in total 6 workers are started (3 for each host).
 
-===== port
-
-The default port of the Elasticsearch server if the port number is missing in <<hosts-option>> URL. The default port number is 9200.
-
 ===== username
 
 The basic authentication username for connecting to Elasticsearch.
@@ -148,9 +144,51 @@ for more information about the environment variables.
 
 ===== index
 
-The index root name to write events to. The default is the Beat name.
-For example "{beatname_lc}" generates "[{beatname_lc}-]YYYY.MM.DD" indexes (for example,
-"{beatname_lc}-2015.04.26").
+The index name to write events to. The default is "{beatname_lc}-%{+yyyy.MM.dd}" (for example, "{beatname_lc}-2015.04.26").
+
+===== indices
+
+Array of index selector rules supporting conditionals, *Format String*
+based field access and name mappings. The first rule matching will be used to
+set the `index` for the event to be published. If `indices` is missing or no
+rule matches, the `index` field will be used.
+
+Rule settings:
+
+*`index`*: The index *Format String* to use. If the fields used are missing, the rule fails.
+
+*`mapping`*: Dictionary mapping index names to new names
+
+*`default`*: Default string value if `mapping` does not find a match.
+
+*`when`*: Condition which must succeed in order to execute the current rule.
+
+Examples elasticsearch output with `indices`:
+
+["source","yaml"]
+------------------------------------------------------------------------------
+output.elasticsearch:
+  hosts: ["http://localhost:9200"]
+  index: "logs-%{+yyyy.MM.dd}"
+  indices:
+    - index: "critical-%{+yyyy.MM.dd}"
+      when.contains:
+        message: "CRITICAL"
+    - index: "error-%{+yyyy.MM.dd}"
+      when.contains:
+        message: "ERR"
+------------------------------------------------------------------------------
+
+===== pipeline
+
+*Format String* value configuring the ingest node pipeline to write events to.
+
+===== pipelines
+
+Array of pipeline selector configurations supporting conditionals, *Format String*
+based field access and name mappings. The first rule matching will be used to
+set the `pipeline` for the event to be published. If `pipelines` is missing or no
+rule matches, the `pipeline` field will be used.
 
 ===== template
 
@@ -527,8 +565,56 @@ The password for connecting to Kafka.
 
 ===== topic
 
-The Kafka topic used for produced events. The setting can be a format string
+The Kafka topic used for produced events. The setting can be a *Format String*
 using any event field. To set the topic from document type use `%{[type]}`.
+
+===== topics
+
+Array of topic selector rules supporting conditionals, *Format String*
+based field access and name mappings. The first rule matching will be used to
+set the `topic` for the event to be published. If `topics` is missing or no
+rule matches, the `topic` field will be used.
+
+Rule settings:
+
+*`topic`*: The topic *Format String* to use. If the fields used are missing, the
+ rule fails.
+
+*`mapping`*: Dictionary mapping index names to new names
+
+*`default`*: Default string value if `mapping` does not find a match.
+
+*`when`*: Condition which must succeed in order to execute the current rule.
+
+===== key
+
+The Kafka event key. The event key must be unique and can be extracted from the
+event using a *Format String*.
+
+===== partition
+
+Kafka output broker event partitioning strategy. Must be one of `random`,
+`round_robin`, or `hash`. By default the `hash` partitioner is used.
+
+*`random.group_events`*: Select a new partition by random every `group_events`
+ events being published. The default value is 1 meaning after each event a new
+ parition is picked randomly.
+
+*`round_robin.group_events`*: Select the next partition every `group_events`
+ events being published. The default value is 1 meaning after each event the next partition will be selected.
+
+*`hash.hash`*: List of fields used to compute the partitioning hash value from.
+ If no field is configured, the events `key` value will be used.
+
+*`hash.random`*: Randomly distribute events if no hash or key value can be computed.
+
+All partitioners will try to publish events to all partitions by default. If a
+partitions leader becomes unreachable for the beat, the output might block. All
+partitioners support setting `reachable_only` to overwrite this
+behavior. If `reachable_only` is set to `true`, events will be published to
+available partitions only. Note: publishing to the subset of available partitions
+only potentially increases resource usage due to events becoming more unevenly
+distributed.
 
 ===== client_id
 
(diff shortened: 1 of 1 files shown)


urso:
successfully build docs via: `make docs`

libbeat/docs/outputconfig.asciidoc
```diff
@@ -106,10 +106,6 @@ The number of workers per configured host publishing events to Elasticsearch. Th
 is best used with load balancing mode enabled. Example: If you have 2 hosts and
 3 workers, in total 6 workers are started (3 for each host).
 
-===== port
```
ruflin: More as a historical remark: port was removed as it would not work will with `hosts`.

libbeat/docs/outputconfig.asciidoc
```diff
@@ -148,9 +144,51 @@ for more information about the environment variables.
 
 ===== index
 
-The index root name to write events to. The default is the Beat name.
-For example "{beatname_lc}" generates "[{beatname_lc}-]YYYY.MM.DD" indexes (for example,
-"{beatname_lc}-2015.04.26").
+The index name to write events to. The default is "%\{{beatname_lc}-%\{yyyy.MM.dd}" (for example, "{beatname_lc}-2015.04.26").
```
ruflin: What is the effect of `%\` here? Do you have on purpose {{ before beatname_lc?

libbeat/docs/outputconfig.asciidoc
```diff
@@ -148,9 +144,51 @@ for more information about the environment variables.
...
+===== pipeline
+
+*Format String* value configuring the ingest node pipeline to write events to.
+
+===== pipelines
+
+Array of pipeline selector configurations supporting conditionals, *Format String*
+based field access and name mappings. The first rule matching will be used to
+set the `pipeline` for the event to be published. If `pipelines` is missing or no
+rule matches, the `pipeline` field will be used.
```
ruflin: Similar example as with indices would be nice here.

libbeat/docs/outputconfig.asciidoc
```diff
@@ -527,9 +565,56 @@ The password for connecting to Kafka.
...
+Rule settings:
+
+*`topic`*: The topic *Format String* to use. If the fields used are missing, the
+ rule fails.
+
+*`mapping`*: Dictionary mapping index names to new names
+
+*`default`*: Default string value if `mapping` does not find a match.
+
+*`when`*: Condition which must succeed in order to execute the current rule.
```
ruflin: Perhaps we can have a general example for topics, pipelines, indices we can link to?

libbeat/docs/outputconfig.asciidoc
```diff
@@ -527,9 +565,56 @@ The password for connecting to Kafka.
...
+
+===== key
+
+The Kafka event key. The event key must be unique and can be extracted from the
```
ruflin: What happens if no key is set? Is it by default auto generated by Kafka?

libbeat/docs/outputconfig.asciidoc
```diff
@@ -527,9 +565,56 @@ The password for connecting to Kafka.
...
+event using a *Format String*.
+
+===== partition
+
+Kafka output broker event partitioning strategy. Must be one of `random`,
+`round_robin`, or `hash`. By default the `hash` partitioner is used.
+
+*`random.group_events`*: Select a new partition by random every `group_events`
+ events being published. The default value is 1.
```
ruflin: `The default value is 1 which means after each event a new parition is picked randomly`

monicasarbu:
LGTM

libbeat/docs/outputconfig.asciidoc
urso: keys in kafka are optional. if key is set and reused, the old content will be overwritten. By never changing content, keys can be used for some artificial kind of deduplication.

libbeat/docs/outputconfig.asciidoc
urso: fixed

libbeat/docs/outputconfig.asciidoc
urso: too much repetition.

libbeat/docs/outputconfig.asciidoc
urso: done

libbeat/docs/outputconfig.asciidoc
```diff
@@ -527,9 +565,57 @@ The password for connecting to Kafka.
...
+
+The Kafka event key. The event key must be unique and can be extracted from the
+event using a *Format String*.
+
+===== partition
+
+Kafka output broker event partitioning strategy. Must be one of `random`,
+`round_robin`, or `hash`. By default the `hash` partitioner is used.
+
+*`random.group_events`*: Select a new partition by random every `group_events`
```
dedemorton: I can't parse this sentence. Maybe there is a missing word?

libbeat/docs/outputconfig.asciidoc
```diff
@@ -527,9 +565,57 @@ The password for connecting to Kafka.
...
+ events being published. The default value is 1 meaning after each event a new
+ parition is picked randomly.
+
+*`round_robin.group_events`*: Select the next partition every `group_events`
```
dedemorton: Same comment as line 599

libbeat/docs/outputconfig.asciidoc
```diff
@@ -527,9 +565,57 @@ The password for connecting to Kafka.
...
+ events being published. The default value is 1 meaning after each event the next partition will be selected.
+
+*`hash.hash`*: List of fields used to compute the partitioning hash value from.
+ If no field is configured, the events `key` value will be used.
+
+*`hash.random`*: Randomly distribute events if no hash or key value can be computed.
+
+All partitioners will try to publish events to all partitions by default. If a
+partitions leader becomes unreachable for the beat, the output might block. All
```
dedemorton: partition's leader

libbeat/docs/outputconfig.asciidoc
```diff
@@ -527,9 +565,57 @@ The password for connecting to Kafka.
...
+partitioners support setting `reachable_only` to overwrite this
+behavior. If `reachable_only` is set to `true`, events will be published to
+available partitions only. Note: publishing to the subset of available partitions
```
dedemorton: Make this a proper note by starting it on a separate line and beginning with "NOTE: ". Also, saying "only" makes the sentence grammatically ambiguous in English. I would say something like:

NOTE: Publishing to a subset of available partitions potentially increases resource usage because events may become unevenly distributed.

libbeat/docs/outputconfig.asciidoc
```diff
@@ -148,9 +144,51 @@ for more information about the environment variables.
...
 ===== index
 
-The index root name to write events to. The default is the Beat name.
-For example "{beatname_lc}" generates "[{beatname_lc}-]YYYY.MM.DD" indexes (for example,
-"{beatname_lc}-2015.04.26").
+The index name to write events to. The default is "{beatname_lc}-%{+yyyy.MM.dd}" (for example, "{beatname_lc}-2015.04.26").
+
+===== indices
+
+Array of index selector rules supporting conditionals, *Format String*
```
dedemorton: This description (and the other ones like it in the doc) needs work. It took me a couple of times reading it before I realized that you're describing how to set the index dynamically. I think you're also saying that the index will be set based on the first item in the array that results in a match, but I am not sure.

libbeat/docs/outputconfig.asciidoc
dedemorton: I don't like repetition either, but there's no guarantee with reference info that users will read from top to bottom, so some repetition might be necessary (or at least a pointer to the supporting info).

libbeat/docs/outputconfig.asciidoc
```diff
@@ -148,9 +144,51 @@ for more information about the environment variables.
...
+based field access and name mappings. The first rule matching will be used to
+set the `index` for the event to be published. If `indices` is missing or no
+rule matches, the `index` field will be used.
+
+Rule settings:
+
+*`index`*: The index *Format String* to use. If the fields used are missing, the rule fails.
```
dedemorton: Format String shouldn't be capitalized because it's not a product component. Do you plan to point to the reference content that you wrote about format strings? I'm thinking about relocating the content about the config file to the guides for individual Beats, but I won't do that for now, so use the `{libbeat}/filename.html[link text]` convention to add links for now. I will clean up the links when I move the content. If you want, I can add links after this PR is merged.

libbeat/docs/outputconfig.asciidoc
dedemorton: +1 We should have a usage type topic that describes, for example, how to set the pipeline dynamically. Having the reference content is good enough for Beta, though.

libbeat/docs/outputconfig.asciidoc
ruflin: Can you add a note that the key is optional. From the current description it looks like it is required.

libbeat/docs/outputconfig.asciidoc
urso: copy'n past ;)

> I think you're also saying that the index will be set based on the first item in the array that results in a match
correct.

I did try to make it 'short, as these settings are quite 'complicated' in comparison to most other settings. `indices` (and the others) is an array of rules executed one after another. The first rule matching (conditionals match and format string doesn't fail) in the array of rules will set index the event is published to. If one rule did match, no other rule in the array will be check to not generate ambiguities in case of 2 rules potentially matching. If no rule did match, the `index` field will be evaluated (which is basically another rule). Truth is `index` is just another rule appended to the list of rules defined in `indices`.

libbeat/docs/outputconfig.asciidoc
urso: changing all `*Format String*` to `format string`.

My idea with documenting all types in "config file format" includes (long-term goal) adding the type to all settings in the reference docs by name + link to the definition of said type. Having some docs in libbeat only and others reused among multiple beats I did struggle with the execution, though.

libbeat/docs/outputconfig.asciidoc
urso: done

libbeat/docs/outputconfig.asciidoc
urso: No missing word. Using `group_events` is supposed to use the name as numeric replacement. I want to express that the number configured in `group_events` sets how many consecutive published events will be send to the same partition, before the partitioner selects a new partition to use for the next set events being published.

libbeat/docs/outputconfig.asciidoc
urso: fixed

libbeat/docs/outputconfig.asciidoc
urso: done

libbeat/docs/outputconfig.asciidoc
dedemorton: That sort of substitution works better when the name of the option clearly indicates what it represents. In this case, it's not clear by the name that `group_events` will resolve to a number, so some readers will stumble over how the sentence is constructed. I realize you're trying to avoid using a lot of words, but it's better to be explicit like you've been in your explanation to me. Also, unless you're telling the user to do something, avoid using the imperative. When someone reads "select", they think they need to select something. It might take them a few seconds to realize that you're essentially saying "configure this option so that the software selects."  Does that make sense?

PR #2599, opened 2016-09-20
Cleanup harvester stopping goroutine

ruflin:
* Move close_timeout go routine to Harvest method to only be started when Harvester starts
* Remove once lock for reader as only one location to close channel now

### DIFF
--- a/filebeat/harvester/log.go
+++ b/filebeat/harvester/log.go
@@ -57,6 +57,27 @@ func (h *Harvester) Harvest(r reader.Reader) {
 	// Makes sure file is properly closed when the harvester is stopped
 	defer h.close()
 
+	// Channel to stop internal harvester routines
+	harvestDone := make(chan struct{})
+	defer close(harvestDone)
+
+	// Closes reader after timeout or when done channel is closed
+	// This routine is also responsible to properly stop the reader
+	go func() {
+		var closeTimeout <-chan time.Time
+		if h.config.CloseTimeout > 0 {
+			closeTimeout = time.After(h.config.CloseTimeout)
+		}
+
+		select {
+		case <-closeTimeout:
+			logp.Info("Closing harvester because close_timeout was reached: %s", h.state.Source)
+		case <-h.done:
+		case <-harvestDone:
+		}
+		h.fileReader.Close()
+	}()
+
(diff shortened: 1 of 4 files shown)


filebeat/harvester/log.go
```diff
@@ -57,6 +57,27 @@ func (h *Harvester) Harvest(r reader.Reader) {
...
+	go func() {
+		var closeTimeout <-chan time.Time
+		if h.config.CloseTimeout > 0 {
+			closeTimeout = time.After(h.config.CloseTimeout)
+		}
+
+		select {
+		case <-closeTimeout:
+			logp.Info("Closing harvester because close_timeout was reached: %s", h.state.Source)
+		case <-h.done:
```
urso: with harvestDone being always closed if harvester loop quits, we still need to check for h.done? Can harvester loop 'block' and not detect h.done and will h.fileReader.Close() unblock the loop?

Maybe add doc telling about the different use-cases of each channel, as they look quite redundant as is.

filebeat/harvester/log_file.go
```diff
@@ -167,9 +165,6 @@ func (r *LogFile) wait() {
 }
 
 func (r *LogFile) Close() {
-	// Make sure reader is only closed once
-	r.singleClose.Do(func() {
-		close(r.done)
-		// Note: File reader is not closed here because that leads to race conditions
-	})
+	close(r.done)
```
urso: why remove the `sync.Once` here? Can we guarantee only one go-routine calling Close just once now?

filebeat/harvester/log.go
ruflin: Both channels are still needed. I added a comment to make it more obvious. Let me know if these comments help.

filebeat/harvester/log_file.go
ruflin: As far as I can see there is now only one place left where the fileReader is closed and that is when the goroutine above stops. It seems the Once was still in for historical reason where filereader was closed in different places (initially we had 2 go routines, one for the h.done channel and one for the timeout.

ruflin:
@urso Additional commit with comments pushed.

tsg:
LGTM, but leaving this one to @urso to merge.

PR #2605, opened 2016-09-20
json_decode_fields processor 

suraj-soni:
This pull request captures a feature to enable/disable nested json parsing in filebeat. It's picked up from where the conversations were left off in #2435.

The commit includes : 
1. decode_json_fields processor - initial commit
2. pulled common json transform code into libbeat/common/json_transform
3. refactored filebeat/harvester/reader/json.go to take advantage for 2


tsg:
@suraj-soni Thanks a lot for the PR!

I'd like to ask for a system test that implements the use case that you have (e.g. log file + configuration).

Also not sure why the PR seems to contain 110 commits, when the diff is actually small, but we can figure this out later.

suraj-soni:
@tsg : For some reason git wouldn't allow me to check in without doing a rebase on my branch. The 110 commits are mostly just rolling forward to the then  commit on master branch.

I'll share the log file and configuration used for testing the fix soon.

suraj-soni:
[json_harvester_test.zip](https://github.com/elastic/beats/files/531952/json_harvester_test.zip)
@tsg : I've uploaded a sample yml config and a log file for running the test.

libbeat/common/mapstr.go
```diff
@@ -177,6 +177,25 @@ func (m MapStr) GetValue(key string) (interface{}, error) {
 	return mapp[keyParts[keyPartsLen-1]], nil
 }
 
+func (m MapStr) AddValue(value MapStr, field string) error {
```
andrewkroh: Can you use the `MapStr.Put()` method that was recently added?

libbeat/processors/actions/json_decode_fields.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/processors"
+)
+
+type jsonDecodeFields struct {
+	Fields []string
+}
+
+func init() {
+	processors.RegisterPlugin("json_decode_fields",
```
andrewkroh: Naming-wise I'd make it `decode_json_fields`. The other processors put the verb first.

libbeat/processors/actions/json_decode_fields.go
```diff
@@ -0,0 +1,91 @@
...
+	return f, nil
+}
+
+func (f jsonDecodeFields) Run(event common.MapStr) (common.MapStr, error) {
+	for _, field := range f.Fields {
+		var output map[string]interface{}
+
+		data, err := event.GetValue(field)
+		if err != nil {
+			return event, fmt.Errorf("no %s field found: %s", field, err)
```
andrewkroh: Should the processor short-circuit on error or save the error and continue processing the remaining fields? The other processors save the error, continue, then when done they return all the errors that occurred.

suraj-soni: I looked at the drop_field processor and looks like it takes the same approach as well. 
`err := event.Delete(field)
		if err != nil {
			return event, fmt.Errorf("Fail to delete key %s: %s", field, err)
		}`

mapstr.GetValue() returns an error when it doesn't find a key. I'm hesitant to use a hasKey + getValue since it will involve parsing the data twice. 

Any suggestions on how to fix this ?

libbeat/processors/actions/json_decode_fields.go
suraj-soni: Makes sense. Will make that change.

libbeat/processors/actions/json_decode_fields.go
andrewkroh: You must be looking at a stale copy of the drop_fields.go. Here's the [latest](https://github.com/elastic/beats/blob/master/libbeat/processors/actions/drop_fields.go#L50) version that puts the errors into a slice and continues.

> mapstr.GetValue() returns an error when it doesn't find a key. I'm hesitant to use a hasKey + getValue since it will involve parsing the data twice. Any suggestions on how to fix this ?

If you want to ignore errors caused by the key being absent, you can check if the cause is an `common.ErrKeyNotFound` like the [include_fields](https://github.com/elastic/beats/blob/master/libbeat/processors/actions/include_fields.go#L55-L59) action does.

suraj-soni:
@andrewkroh : Have incorporated the changes that you suggested. However the build fails with an error : 
`filebeat/filebeat.full.yml: needs update
make: *** [check] Error 1`

Not sure how to fix this since I've updated the filebeat.full.yml to include an example of the new processor.

andrewkroh:
`filebeat/filebeat.full.yml` is a generated file. Make the changes to `filebeat/filebeat.full.yml` and then run `make update` in the filebeat directory.

Also I think you need to rebase your changes. There should only be 1 commit in the PR.

suraj-soni:
@andrewkroh  : I squashed the previous commits to be only one but now the 04b905c has the changes i made + others that were pulled from the upstream master. 
And now the subsequent builds fail with 
`metricbeat/etc/beat.full.yml: needs update
metricbeat/metricbeat.full.yml: needs update
make: *** [check] Error 1`

Not sure why this would be the case. Sorry for the confusions, this is actually the first time i'm squashing commits

suraj-soni:
@andrewkroh : 
Was able to squash all but two commits, not sure how to merge those two.

tsg:
jenkins, test it

CHANGELOG.asciidoc
```diff
@@ -9,6 +9,7 @@
 
 === Beats version HEAD
 https://github.com/elastic/beats/compare/v5.0.0-rc1...master[Check the HEAD diff]
+https://github.com/elastic/beats/compare/v5.0.0-beta1...master[Check the HEAD diff]
```
ruflin: That should be removed (merge error probably)

libbeat/processors/actions/decode_json_fields.go
```diff
@@ -0,0 +1,93 @@
...
+
+func (f decodeJSONFields) Run(event common.MapStr) (common.MapStr, error) {
+	errs := []string{}
+
+	for _, field := range f.Fields {
+		var output map[string]interface{}
+
+		data, err := event.GetValue(field)
+		if err != nil && errors.Cause(err) != common.ErrKeyNotFound {
+			errs = append(errs, err.Error())
```
andrewkroh: This should `continue` the for loop after the error.

libbeat/processors/actions/decode_json_fields.go
```diff
@@ -0,0 +1,93 @@
...
+		}
+
+		text, ok := data.(string)
+		if ok {
+			err := unmarshal([]byte(text), &output)
+			if err != nil {
+				errs = append(errs, err.Error())
+				event.Put(field, text)
```
andrewkroh: Why `Put` the value into the event on error? The value is already associated with the key in the event.

libbeat/processors/actions/decode_json_fields.go
```diff
@@ -0,0 +1,93 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("fail to unpack the decode_json_fields configuration: %s", err)
+	}
+
+	f := decodeJSONFields{Fields: config.Fields}
+	return f, nil
+}
+
+func (f decodeJSONFields) Run(event common.MapStr) (common.MapStr, error) {
+	errs := []string{}
```
andrewkroh: Use `var errs []string`. https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices

libbeat/processors/actions/decode_json_fields.go
```diff
@@ -0,0 +1,93 @@
...
+
+	for _, field := range f.Fields {
+		var output map[string]interface{}
```
andrewkroh: This doesn't need to be declared until your are actually going to use it. So move it inside of the `if ok { unmarshal ...` block.

libbeat/processors/actions/decode_json_fields.go
```diff
@@ -0,0 +1,93 @@
...
+		text, ok := data.(string)
+		if ok {
+			err := unmarshal([]byte(text), &output)
+			if err != nil {
+				errs = append(errs, err.Error())
+				event.Put(field, text)
+				continue
+			}
+
+			_, addErr := event.Put(field, output)
```
andrewkroh: You can reuse `err` here. `_, err = event.Put(field, output)`.

libbeat/processors/actions/decode_json_fields.go
```diff
@@ -0,0 +1,93 @@
...
+			if addErr != nil {
+				errs = append(errs, addErr.Error())
```
andrewkroh: To be consistent and prevent future bugs this should also `continue` after the error. It doesn't really matter in this case, but if in the future some logic gets added after the location where the error occurred it could cause a bug.

suraj-soni:
@andrewkroh : Thanks for your feedback. I've made the suggested changes and added test cases for the different conditions. 

I see that the build fails with error : 
`The command "curl -L https://github.com/docker/compose/releases/download/${DOCKER_COMPOSE_VERSION}/docker-compose-`uname -s`-`uname -m` > docker-compose" failed and exited with 18 during .`

suraj-soni:
@andrewkroh ,

Thank you for your feedback. Have made the suggested changes and added test cases as well.

libbeat/processors/actions/decode_json_fields_test.go
```diff
@@ -0,0 +1,100 @@
...
+		"pipeline": "us1",
+	}
+
+	assert.Equal(t, expected.String(), actual.String())
+
+}
+
+func TestInvalidJSON(t *testing.T) {
+	input := common.MapStr{
+		"msg":      "{\"log\":\"{\\\"level\\\":\\\"info\\\"}\",\"stream\":\"stderr\",\"count\":3",
```
andrewkroh: If you create a raw string literal by surrounding the string with back ticks (`) then you don't need to use escaping on those quotes. I don't care if you fix this, it's more of a golang tip to keep you sane.

libbeat/processors/actions/decode_json_fields_test.go
```diff
@@ -0,0 +1,100 @@
...
+var config, _ = common.NewConfigFrom(map[string]interface{}{
+	"fields": fields,
+})
+
+func TestMissingKey(t *testing.T) {
+	input := common.MapStr{
+		"pipeline": "us1",
+	}
+
+	actual := getActualValue(t, config, input)
```
andrewkroh: These test cases that expect an error should also assert that an error occurred. You can use [`assert.Error()`](https://godoc.org/github.com/stretchr/testify/assert#Error).

CHANGELOG.asciidoc
```diff
@@ -8,6 +8,7 @@
 // Template, add newest changes here
 
 === Beats version HEAD
+https://github.com/elastic/beats/compare/v5.0.0-rc1...master[Check the HEAD diff]
```
andrewkroh: It looks like you made no changes to this file, so can you reset this file so that there are no changes in the diff.

andrewkroh: Or better yet, reset it and then add a line in the Added / Affecting all Beats section that contains something like:

`- Added decode_json_fields processor for decoding fields containing JSON strings. {pull}2605[2605]`

suraj-soni:
@andrewkroh  Updated the changelog.asciidoc file to incorporate your suggestions.

libbeat/common/json_transform.go
```diff
@@ -0,0 +1,46 @@
+package common
```
urso: this file moves helpers and exposes the json specific helpers with generic names. I'd rather see some json parsing utility function/module doing the right thing instead of exposing internal helpers.

monicasarbu:
@suraj-soni Can you please address the most recent comments, before we merge the PR?

libbeat/processors/actions/decode_json_fields.go
```diff
@@ -0,0 +1,101 @@
...
+	for _, field := range f.Fields {
+		data, err := event.GetValue(field)
+		if err != nil && errors.Cause(err) != common.ErrKeyNotFound {
+			debug("Error trying to GetValue for field : %s in event : %v", field, event)
+			errs = append(errs, err.Error())
+			continue
+		}
+		text, ok := data.(string)
+		if ok {
+			var output map[string]interface{}
```
urso: we want to restrict ourselves to json objects only here? Using `var output interface{}` we can decode booleans, number, objects and arrays. Any valid json.

libbeat/processors/actions/decode_json_fields.go
```diff
@@ -0,0 +1,101 @@
...
+			sErr := unmarshal([]byte(vv), &output)
+			if sErr == nil {
+				(*fields)[k] = output
+			}
+		}
+	}
+
+	jsontransform.TransformNumbers(*fields)
+	return nil
+}
```
urso: thinking a little more about `TransformNumbers` and `unmarshal`, I wonder why wen need to put `TransformNumbers` at the end. By moving `TransformNumber` before the deep-parsing stuff, the initial decoding would be exactly similar to the filebeat reader.

```
func DecodeJson(text []byte, to interface{}) {
  dec := json.NewDecoder(bytes.NewReader(text))
  dec.UseNumber()
  err := dec.Decode(to)
  if err != nil {

  if m, ok := to.(*map[string]interface{}); ok {
    jsontransform.TransformNumbers(common.MapStr(m))
  }
  return err
}
```

Then (extending unmarshal a little), unmarshal can be generalized a little:

```
func unmarshal(maxDepth int, text []byte, fields interface{}, processArray bool) error {
  if err := DecodeJson(text, fields); err != nil {
    return err
  }

  maxDepth--
  if maxDepth == 0 {
    return nil
  }

  tryUnmarshal := func(v interface{}) (interface{}, bool) {
    str, isString := v.(string)
    if !isString {
      return v, false
    }

    var tmp interface{}
    err := unmarshal(maxDepth, []byte(str), &tmp, processArray)
    if err != nil {
      return v, false
    }

    return tmp, true
  }

  // try to deep unmarshal fields
  switch O := fields.(type); {
  case *map[string]interface{}:
    for k, v := range O {
      if decoded, ok := tryUnmarshal(v); ok {
        O[k] = decoded
      }
    }

  // We want to process arrays here
  case []interface{}:
    if !processArray {
      break
    }

    for i, v := range O {
      if decoded, ok := tryUnmarshal(v); ok {
        O[i] = decoded
      }
    }
  }
}
```

As you can see, I added a maximum parsing depth. Not sure we really want to parse fully recursively here. I'd use `maxDepth=1` by default (Make it configurable). More parsing can be done by configuring a second processor continue parsing on already parsed field.
I also added optional array support, plus lifted the requirement for JSON Objects parsing only.

Having DecodeJson, there is no need for exporting jsontransform as is (but I'd keep jsontransform package for now, as other modules in beats might make use of it).

What do you think?

urso:
Sorry for having you wait, but I wanted to give this feature a little more thought. See my comments.

suraj-soni:
I like your suggestions and will work towards incorporating them in the changes and share it for further review.

libbeat/processors/actions/decode_json_fields.go
suraj-soni: I like the idea of having a maxDepth setup so that we don't recursively parse till the end. 
This does give us the ability to configure the processors for different fields with different levels of depth parsing.

suraj-soni:
Hi @urso ,

Have made the suggested changes and added a couple of test cases to verify the maxDepth feature as well. Thanks once again for your help in refining this.

urso:
@suraj-soni LGTM. Unfortunately you've got a merge conflict on `.gitignore` and PR overwrites some vendored file with an old version. Can you rebase based on master, so I can merge?

suraj-soni:
@urso : I've merged the changes from the master and also removed the conflict with .gitignore.

urso:
Merged. Thanks!

suraj-soni:
Thanks for helping to get this in @urso  and @andrewkroh !

vendor/github.com/elastic/go-ucfg/CHANGELOG.md
```diff
@@ -120,6 +120,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 [Unreleased]: https://github.com/elastic/go-ucfg/compare/v0.3.7...HEAD
 [0.3.7]: https://github.com/elastic/go-ucfg/compare/v0.3.6...v0.3.7
+[Unreleased]: https://github.com/elastic/go-ucfg/compare/v0.3.6...HEAD
```
ruflin: @urso Now sure how this sneaked in. Can you clean that up in https://github.com/elastic/beats/pull/3045

ruflin:
@dedemorton @urso There are no docs for this yet and also no entries yet. As this will probably make it into 5.1 we should definitively add docs to it with examples on how to use it. 

@suraj-soni In case you can help here, all docs are in asciidoc and can be found here for general processors: https://github.com/elastic/beats/tree/master/libbeat/docs As this is mostly filebeat related, it probably makes sense to add examples to filebeat docs: https://github.com/elastic/beats/tree/master/filebeat/docs

ruflin:
Here is a fix for the two issues mentioned: https://github.com/elastic/beats/pull/3047

suraj-soni:
@ruflin : Sure, I'd be happy to help with the documentation for this.

PR #2610, opened 2016-09-21
Added the cherrypick_pr script to the 5.0 branch

tsg:
This is useful in case of conflicts. You need to re-run the
script with --continue, from the target branch, so it's helpful
if that branch actually has the script.


ruflin:
Agree, was also missing it today first.

PR #2612, opened 2016-09-21
[metricbeat/module/apache] Fix unused constants defaultScheme and autoQueryParam

radoondas:
Defined constants were left unused in code.

### DIFF
--- a/filebeat/tests/open-file-handlers/Dockerfile-filebeat
+++ b/filebeat/tests/open-file-handlers/Dockerfile-filebeat
@@ -1,10 +1,10 @@
 FROM debian
 
 RUN apt-get update && apt-get install -y wget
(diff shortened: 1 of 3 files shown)


karmi:
Hi @radoondas, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/beats/pull/2612.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

ruflin:
@radoondas Thanks for the fix. I think I replaced the variables when I was doing some testing on the metricset and forgot to put them back in.

ruflin:
LGTM

radoondas:
I will close this PR, fix my IDE and do new PR instead. Sorry for that. :(

PR #2625, opened 2016-09-23
Docker module for Metricbeat

ruflin:
All credit for this module goes to @douaejeouit and Ingensi. See elastic#2444


ruflin:
@douaejeouit I took again the most recent version of your branch and applied to changes. I still not sure on what goes wrong on your side. Instead of spending time to figure out this issue I suggest we are going ahead with this branch here and merge it. As soon as it is in master, we can start iterate on top of it and the changes will be much smaller and easier to debug. WDYT?

douaejeouit:
@ruflin   Ok, it sounds good to me  :tada:  !!

PR #2627, opened 2016-09-23
Send float fields as float values to Elasticsearch

monicasarbu:
This PR tries to remove the necessity to define the mapping before inserting the data into Elasticsearch, by making sure that float fields are always sent as float values even if they are not flows. For example `5` will be sent as `5.00000`, so Elasticsearch defines it as `float`.

With this change, it's not needed to load the template in advance, before inserting the data into Elasticsearch.

NOTE: This also affects processors.

### DIFF
--- a/metricbeat/tests/system/test_processors.py
+++ b/metricbeat/tests/system/test_processors.py
@@ -134,7 +134,7 @@ class TestProcessors(metricbeat.BaseTest):
         output = self.read_output(
             required_fields=["@timestamp", "type"],
         )
-        assert len(output) == 1
+        assert len(output) >= 1
 
 
     def test_include_fields(self):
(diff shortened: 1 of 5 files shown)


ruflin:
I suggest we backport this to 5.0 as soon as the tests pass as I don't think this is a breaking change. As we discussed it does not prevent people from still loading the dashboard, but it is a nice fallback.

metricbeat/tests/system/test_processors.py
```diff
@@ -134,7 +134,7 @@ def test_dropevent_with_complex_condition(self):
         output = self.read_output(
             required_fields=["@timestamp", "type"],
         )
-        assert len(output) == 1
+        assert len(output) >= 1
```
ruflin: Is this change somehow related to the changes above or just a flaky test?

ruflin:
LGTM

metricbeat/tests/system/test_processors.py
monicasarbu: flaky test, as it's variable how many events are sent out.

PR #2632, opened 2016-09-23
Fix default Metricbeat config on Windows

### LINKED ISSUE #2623: metricset 'system/load' is not registered (on Windows)
- Version: 5.0.0-beta1
- Operating System: Windows 10
- Steps to Reproduce: 

1) start metricbeat.exe with the default metricbeat.yml (actually it was the .full.yml)
2) Fatal error as shown below

```
D:\5.0.0-beta1\metricbeat>metricbeat.exe
Exiting: 1 error: metricset 'system/load' is not registered, metricset not found
```

3) I commented out the `- load` line and now it starts fine
```
#------------------------------- System Module -------------------------------
- module: system
  metricsets:
    # CPU stats
    - cpu

    # System Load stats
    #- load
```

tsg:
The load metricset is not supported on Windows, and if it's enabled
in the configuration file, it makes Metricbeat exit immediately with
an error.

This change adjusted the configuration on Windows to exclude the load
metricset.

Fixes #2623.

### DIFF
--- a/metricbeat/Makefile
+++ b/metricbeat/Makefile
@@ -57,6 +57,11 @@ imports:
 # This is called by the beats packer before building starts
 .PHONY: before-build
 before-build:
+	# disable the system/load metricset on windows
+	sed -i.bk 's/- load/#- load/' $(PREFIX)/metricbeat-win.yml
+	rm $(PREFIX)/metricbeat-win.yml.bk
+	sed -i.bk 's/- load/#- load/' $(PREFIX)/metricbeat-win.full.yml
+	rm $(PREFIX)/metricbeat-win.full.yml.bk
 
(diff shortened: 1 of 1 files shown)


tsg:
@elastic/beats this solution works, but I'm not entirely happy with it, as I'd prefer to minimize the changes we do to the default configs. Should we consider a way of automatically disabling metricsets/modules based on the platform?

ruflin:
An alternative solution is that instead of failing when a metricset does not exist that we want to load, that we only print an error. An option is to only log an error for the ones we exist but are not supported on this platform (special errors).

The problem with the automatic disabling is that when a user configures it and is not aware that it is not supported under windows, he will complain that no data is sent.

PR #2637, opened 2016-09-23
Backport:Make it clear that 'any' only works on Linux (#2585)

tsg:
Backport of #2585.

The previous wording in the default config make it look like 'any'
would work on OS X & Windows, but it doesn't.


ruflin:
LGTM

PR #2641, opened 2016-09-26
Add batch_read_size config to Winlogbeat

andrewkroh:
This configuration option allows users to control the number of event log records that are read, processed, and published in its event loop.

### DIFF
--- a/winlogbeat/eventlog/wineventlog.go
+++ b/winlogbeat/eventlog/wineventlog.go
@@ -17,9 +17,6 @@ import (
 )
 
 const (
-	// defaultBatchReadSize is the maximum number of event Read will return.
-	defaultBatchReadSize = 100
-
 	// renderBufferSize is the size in bytes of the buffer used to render events.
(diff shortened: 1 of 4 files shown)


winlogbeat/eventlog/wineventlog.go
```diff
@@ -218,7 +219,7 @@ func reportDrop(reason interface{}) {
 // newWinEventLog creates and returns a new EventLog for reading event logs
 // using the Windows Event Log.
 func newWinEventLog(options map[string]interface{}) (EventLog, error) {
-	var c winEventLogConfig
+	c := winEventLogConfig{BatchReadSize: defaultBatchReadSize}
```
ruflin: The more "ucfg" like approach here would be probably to introduce a `defaultConfig` and set detaulBatchReadSize there?

andrewkroh: Fixed

ruflin:
LGTM

PR #2642, opened 2016-09-26
Backport: Add batch_read_size config to Winlogbeat

andrewkroh:
This configuration option allows users to control the number of event log records that are read, processed, and published in its event loop.

Backport of #2641


ruflin:
There is something strange going on in the appveyor test: https://ci.appveyor.com/project/elastic-beats/beats/build/5880/job/vje5yr4pbywlwjqc

andrewkroh:
I re-ran the build and it seems to be OK. That test case could be a bit brittle, but it's not one that's affected by these changes.

PR #2662, opened 2016-10-03
Fix formatting of list

dedemorton:
Adds a missing plus sign to fix list formatting. The formatting is correct in master, so the plus must have been dropped during the merge or something.


ruflin:
LGTM

Issue #2629, opened 2016-09-23
Docker module improvements


ruflin:
- [x] Add tests to generate `data.json` files https://github.com/elastic/beats/pull/2665
- [x] Check that config files are according to standard (experimental disabled by default) https://github.com/elastic/beats/pull/2665
- [x] Review all docker fields for the naming convention https://github.com/elastic/beats/pull/2665
- [x] Add CHANGELOG update https://github.com/elastic/beats/pull/2674
- [x] Do manual tests 
- [x] Complete fields.yml with all fields https://github.com/elastic/beats/pull/2674
- [x] Add system tests to verify fields.yml https://github.com/elastic/beats/pull/2971
- [x] Add more tests
- [x] Add MetricSet for [/info](https://docs.docker.com/engine/reference/api/docker_remote_api_v1.24/#/display-system-wide-information):  https://github.com/elastic/beats/pull/2672
- [ ] Run `golint ./... | grep -v vendor` and check errors for docker module
- [x] Add dashboard #2943.
- [x] Add socket info under host if used: https://github.com/elastic/beats/pull/2665#discussion_r82192489
- [x] Change and test labels to be taken with . if possible: https://github.com/elastic/beats/pull/2665#discussion_r82193492
- [ ] Write migration guide from [dockbeat](https://github.com/Ingensi/dockbeat)
- [x] How to handle ports and labels list in container metric
- [ ] Limit number of digits for percentage value. Use same logic as in system/cpu metricset
- [ ] Fetching CPU data takes > 2s because of average calculation on docker side -> API Problem
- [ ] Clean up ssl handling / naming
- [ ] Update dashboards

tsg:
I played with creating a dashboard for a demo and I have a couple of observations/suggestions. Don't take this as criticism, please, I could create some pretty cool dashboards and the demo is going to be great!

On CPU stats:

* Looks like metrics like CPU time are [derivates](https://github.com/elastic/beats/blob/master/metricbeat/module/docker/cpu/helper.go#L93).  That's quite convenient for my demo :), but goes against the principle that Metricbeat does total values by default.  We should consider indicating somehow in the name that this is a derivate.
* For the CPU time ahead, if I interpret the code correctly, it's in seconds. We should indicate that in the metrics names.
[...]

PR #2665, opened 2016-10-03
Docker module cleanup

ruflin:
* Make default period 10s, same as for all other beats
* Remove unnecessary code comments
* Make convertContainerPort local to container MetricSet
* Only add labels to container MetricSet if not empty
* Apply some code conventions
* Convert local @timestamp to timestamp to prevent confusion
* Replace rx / tx with in / out in network metricset to be consistent with system network metricset
* Add interface name as field instead of having it as a dynamic element. This makes it easier to filter.
* If not containers are running, just no events are returned
* Add data.json for all metricsets
* Introduce ToMapStr for Container data
* Move socket to Container MapStr. Is Socket info really needed?
* Remove unnecessary nesting levels as namespace already given by metricset
* Have one docker client per metricset instead of having one global client

### DIFF
--- a/metricbeat/module/docker/config.go
+++ b/metricbeat/module/docker/config.go
@@ -1,7 +1,8 @@
 package docker
 
 type TlsConfig struct {
-	Enabled  bool   `config:"enabled"`
+	Enabled bool `config:"enabled"`
+	// TODO: Naming should be standardised with output cert configs
 	CaPath   string `config:"ca_path"`
 	CertPath string `config:"cert_path"`
 	KeyPath  string `config:"key_path"`
@@ -12,8 +13,8 @@ type Config struct {
 	Tls    TlsConfig
 }
 
-func GetDefaultConf() *Config {
-	return &Config{
+func GetDefaultConf() Config {
+	return Config{
 		Socket: "unix:///var/run/docker.sock",
 		Tls: TlsConfig{
 			Enabled: false,
--- a/metricbeat/module/docker/container/data.go
+++ b/metricbeat/module/docker/container/data.go
@@ -6,7 +6,6 @@ import (
 	dc "github.com/fsouza/go-dockerclient"
 
 	"github.com/elastic/beats/libbeat/common"
-
 	"github.com/elastic/beats/metricbeat/module/docker"
 )
 
@@ -17,24 +16,40 @@ func eventsMapping(containersList []dc.APIContainers) []common.MapStr {
 	}
 	return myEvents
 }
-func eventMapping(mycontainer *dc.APIContainers) common.MapStr {
+
+func eventMapping(cont *dc.APIContainers) common.MapStr {
 
 	event := common.MapStr{
-		"@timestamp": time.Now(),
-		"container": common.MapStr{
-			"created":      common.Time(time.Unix(mycontainer.Created, 0)),
-			"id":           mycontainer.ID,
-			"name":         docker.ExtractContainerName(mycontainer.Names),
-			"labels":       docker.BuildLabelArray(mycontainer.Labels),
-			"command":      mycontainer.Command,
-			"image":        mycontainer.Image,
-			"ports":        docker.ConvertContainerPorts(&mycontainer.Ports),
-			"size_root_fs": mycontainer.SizeRootFs,
-			"size_rw":      mycontainer.SizeRw,
-			"status":       mycontainer.Status,
+		"created": common.Time(time.Unix(cont.Created, 0)),
+		"id":      cont.ID,
+		"name":    docker.ExtractContainerName(cont.Names),
+		"command": cont.Command,
+		"image":   cont.Image,
+		"ports":   convertContainerPorts(cont.Ports),
+		"labels":  docker.BuildLabelArray(cont.Labels),
+		"size": common.MapStr{
+			"root_fs": cont.SizeRootFs,
+			"rw":      cont.SizeRw,
 		},
-		"socket": docker.GetSocket(),
+		"status": cont.Status,
 	}
 
 	return event
+}
+
+func convertContainerPorts(ports []dc.APIPort) []map[string]interface{} {
+	var outputPorts = []map[string]interface{}{}
+	for _, port := range ports {
+		outputPort := common.MapStr{
+			"ip": port.IP,
+			"port": common.MapStr{
+				"private": port.PrivatePort,
+				"public":  port.PublicPort,
+			},
+			"type": port.Type,
+		}
+		outputPorts = append(outputPorts, outputPort)
+	}
+
+	return outputPorts
 }
--- a/metricbeat/module/docker/cpu/cpu_test.go
+++ b/metricbeat/module/docker/cpu/cpu_test.go
@@ -3,13 +3,11 @@ package cpu
 import (
 	"reflect"
 	"testing"
-	"time"
 
 	dc "github.com/fsouza/go-dockerclient"
 	"github.com/stretchr/testify/assert"
 
 	"github.com/elastic/beats/libbeat/common"
-	"github.com/elastic/beats/metricbeat/module/docker"
 )
 
 func TestCPUService_PerCpuUsage(t *testing.T) {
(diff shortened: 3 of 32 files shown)


ruflin:
@douaejeouit It would be nice if you could have a look at this PR and especially have a look at the // TODO I added. Would be good to get your thoughts on these.

metricbeat/module/docker/container/data.go
```diff
@@ -17,24 +16,45 @@ func eventsMapping(containersList []dc.APIContainers) []common.MapStr {
 	}
 	return myEvents
 }
-func eventMapping(mycontainer *dc.APIContainers) common.MapStr {
+
+func eventMapping(cont *dc.APIContainers) common.MapStr {
 
 	event := common.MapStr{
-		"@timestamp": time.Now(),
+		"timestamp": time.Now(),
```
tsg: Not sure I understand this, does it mean we'll have both `@timestamp` and `timestamp`? Seems unnecessary.

ruflin: @tsg Agree. Before one was overwriting the other (or at least it tried and didn't as it was not common.Time). With the above implementation each event can have its separate timestamp instead of one for all events of this batch. But as the heavy work happens when the info from the docker api is fetched and the timestamp is generated afterwards, I suggest to just remove it.

Perhaps @douaejeouit can share some insights here?

douaejeouit: The @timestamp here shouldn't be time.now() ! The idea is to get the time generate by Docker daemon while sending the events requested ( here the containers event), this is what I did for all the other metricsets.
I will be on it tomorrow morning!  There is some smalll changes to do, otherwise, we can simply keep @timestamp as it is for the containers event! What do you think guys ? @ruflin @tsg

tsg:
LGTM.

metricbeat/module/docker/container/data.go
ruflin: @douaejeouit I saw this part, but we are talking here about a milli second difference in time. I don't think it is worth to do this and have the logic because of this different from all other modules / metricsets where the timestamp represents when the event fetching started. The nice part of this is that `timestamp + rrt` is when the even creation ended. With the above implementation it is not 100% clear to me what the timestamp will exactly represent.

@douaejeouit  Perhaps you can share some insights on what the timestamp exactly is.

ruflin:
@douaejeouit We merged this PR, but lets continue this discussion about the timestamp. Happy to readd it if needed. Please also have a look at the PR to check if I changed some code that should not be changed.

metricbeat/module/docker/container/data.go
douaejeouit: @ruflin Ok!
But .. what do you mean by having one dockerclient per metricset instead of having one global client? For me the docker client instance should be created only once. Could you please clarify this point?
Concerning timestamp, as I said before, it's overwritten and it represents the time when the stats were read from the deamon. But since it's only a matter of a milli seconds difference in time, I agree with you! We can change the logic 👍

douaejeouit: Ok, 
Indeed, I made the choice to have one client for failing to 5 simultaneous connections if all the metricsets are used.. anyway, it's Ok as it is not problematic.

ruflin: Before it was not only 1 client for all 5 metricsets, but one client for all docker modules (one docker module can be used multiple times in the config) which broke the implementation.

douaejeouit: All right .. I can see now! Thanks

metricbeat/module/docker/cpu/cpu_test.go
```diff
@@ -73,6 +74,8 @@ func TestCPUService_UsageInUsermode(t *testing.T) {
 	//  THEN
 	assert.Equal(t, float64(0.50), result)
 }
+
+/* TODO: uncomment
```
douaejeouit: Should I comment it? This function needs to be updated in order to match with last changes you made. I'll do it.

metricbeat/module/docker/helper.go
```diff
@@ -16,13 +16,29 @@ type Container struct {
 	Socket *string
 }
 
-func InitCurrentContainer(container *docker.APIContainers) *Container {
+func (c *Container) ToMapStr() common.MapStr {
+	m := common.MapStr{
+		"id":   c.Id,
+		"name": c.Name,
+		// TODO: Is this really needed
+		"socket": GetSocket(),
```
douaejeouit: It's an extra information. We can ignore it if you judge it isn't needed!

metricbeat/module/docker/helper.go
```diff
@@ -35,37 +51,25 @@ func ExtractContainerName(names []string) string {
...
+	outputLabels := make([]common.MapStr, len(labels))
 	i := 0
 	var keys []string
 	for key := range labels {
 		keys = append(keys, key)
 	}
 	sort.Strings(keys)
 	for _, k := range keys {
+		// Replace all . in the labels by _
+		// TODO: WHY?
```
douaejeouit: It was the naming used in dockbeat Project, where it has been decided to format the labels output . I didn't want it to be different, that's why I left it at it is. It's not necessery. We can leave it as it's generated by docker's API.

metricbeat/module/docker/config.go
```diff
@@ -1,7 +1,8 @@
 package docker
 
 type TlsConfig struct {
-	Enabled  bool   `config:"enabled"`
+	Enabled bool `config:"enabled"`
+	// TODO: Naming should be standardised with output cert configs
```
douaejeouit: 👍

metricbeat/module/docker/cpu/cpu_test.go
ruflin: Yes, please. Thanks

metricbeat/module/docker/helper.go
ruflin: I'm trying to understand the use case behind it. It see that people will search for id and name, but will the socket information be of any use? I think this is useful in the `info` metricset, but it is required in the other metricsets. I'm hesitant to remove it as I'm sure there was a good reason to add it which I'm just not aware of yet.

metricbeat/module/docker/helper.go
ruflin: I was first thinking . were replaced by _ because in ES <2.4 . can't be used in field names. But it seems to be the names where it is replaced? Perhaps there are some disadvantages in Kibana with the . ? Would be interesting to get some more background info on this decision.

metricbeat/module/docker/helper.go
douaejeouit: The main objectif of having it as a field is to have an clear idea about the conf and to check it as well.  Espacially in case we have two docker modules running and each one listen on different socket. Since  we can bind docker to another host or port using a tcp connection, it's always good to know if it listen on TCP or a Unix socket or both, that's why it's present n all the metricsets!

ruflin: I'm thinking in case a socket is used, if we should copy over the socket information to `metricset.host` info instead?

metricbeat/module/docker/helper.go
ruflin: I wasn't aware, that the above was and issue for values. I would expect it to be an issue for keys, but also values? Were the values defined as text or keywords? That could have been the issue, that if it was not defined as keywors that the tokenizer was applied? But this can be solved by applying the right mapping.

We should test the behaviour here, as in general I think it would be nice to keep the name of the labels the same as the user set it, as this is what he will probably look for.

metricbeat/module/docker/helper.go
douaejeouit: +1, it sounds good for me!

ruflin: Added a task to the list here: https://github.com/elastic/beats/issues/2629

metricbeat/module/docker/helper.go
ruflin: Added a task to the list here: https://github.com/elastic/beats/issues/2629

rikatz:
Guys, with my previous contribution about labels and dots, is this still necessary?

ruflin:
@rikatz There were quite a lot of changes and discussions in this PR. Which part are you referring to?

rikatz:
@ruflin Sorry! about the labels cleanup, as mentioned on the helper.go here, about the replacement of dots to underlines.

Also, the task "Change and test labels to be taken with . if possible: #2665 (comment)" on #2629

ruflin:
@rikatz I checked the checkboxes in the issue as done.

PR #2668, opened 2016-10-04
Make package-setup dependency of package

ruflin:
* If `make package` is run for all beats, the setup part is run multiple times. This should not be an issue as the setup part should be fast after the first build.
* When building the dashboards, it is now also directly a dependency.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,7 +75,6 @@ docs:
 
 .PHONY: package
 package: beats-dashboards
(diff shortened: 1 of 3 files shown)


ruflin:
jenkins, package it

tsg:
Green on jenkins: http://build-eu-00.elastic.co/view/Beats/job/beats-package-PR/34/

PR #2671, opened 2016-10-04
Fix the custom BPF filter option

### LINKED ISSUE #2660: Packetbeat: BFP filter doesn't work with flows
Version: Packetbeat 5.0.0-beta1
OS: OSX

Even if bpf filter is configured, flows emits all traffic on all ports.

tsg:
Set the `sniffer.filter` before `setFromConfig` is called. Changed the
factory prototype so it doesn't pass around `filter` and instead just
done it directly via Init.

Fixes #2660.

### DIFF
--- a/packetbeat/beater/packetbeat.go
+++ b/packetbeat/beater/packetbeat.go
@@ -198,11 +198,11 @@ func (pb *Packetbeat) setupSniffer() error {
 	}
 
 	pb.Sniff = &sniffer.SnifferSetup{}
-	return pb.Sniff.Init(false, pb.makeWorkerFactory(filter), &config.Interfaces)
+	return pb.Sniff.Init(false, filter, pb.makeWorkerFactory(), &config.Interfaces)
 }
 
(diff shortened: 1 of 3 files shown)


packetbeat/beater/packetbeat.go
```diff
@@ -198,19 +198,19 @@ func (pb *Packetbeat) setupSniffer() error {
 	}
 
 	pb.Sniff = &sniffer.SnifferSetup{}
-	return pb.Sniff.Init(false, pb.makeWorkerFactory(filter), &config.Interfaces)
+	return pb.Sniff.Init(false, filter, pb.makeWorkerFactory(), &config.Interfaces)
```
urso: no need to have `makeWorkerFactory`. consider changing to:

```
return pb.Sniff.Init(false, filter, pb.createWorker, &config.Interfaces)
...

func (pb *Packetbeat) createWorker(dl layers.LinkType) (sniffer.Worker, error) {
}
```

PR #2675, opened 2016-10-04
Fix addl download links

dedemorton:
Noticed more incorrect download links while testing the GS, so I changed the URLs to match the download site.


ruflin:
LGTM

PR #2682, opened 2016-10-05
Add container meta info on module level

ruflin:
Most of the docker module metric sets have kind of meta information about the container which is identical across the MetricSets. Until now this information was nested inside each MetricSet which made it very hard to query for all information related to one container. To make querying possible this information was moved to the module level.

Each MetricSet can add information to the module level by adding information to the event with the `_module` key. This convention is based on the assumption that this key is never used for any other field. During processing of the event this information is moved up to the module level.

It must be made sure that the mapping for this data is added to the fields.yml on the module level. In case of the docker module this is not necessary as the container MetricSet already has the mappings for these values.

An example event looks as following. This makes querying based on `docker.container.id` simple.

```
{
    "@timestamp": "2016-05-23T08:05:34.853Z",
    "docker": {
        "container": {
            "id": "2d82c01e5e8a975e8a2912f5dcd7a03ec1505f2c7d2fcf54961582d8e9e34509",
            "name": "admiring_ptolemy",
            "socket": "unix:///var/run/docker.sock"
        },
        "cpu": {
            "usage": {
                "kernel_mode": 0,
                "per_cpu": {
                    "0": 0,
                    "1": 0
                },
                "total": 0,
                "user_mode": 0
            }
        }
    },
    "metricset": {
        "host": "localhost",
        "module": "docker",
        "name": "cpu",
        "rtt": 115
    },
    "type": "metricsets"
}
```

Additional Changes:

* Set default time period to 10s
* Make event generation in test based on events builder to make sure events are the same as sent
* Remove container info from fields.yml files as not needed anymore. Update mappings.


tsg:
This is great 👍

tsg:
Implementation wise, we might at some point be more explicit about it, like returning two dicts from the Fetch methods. But I think this is fine for now.

ruflin:
@tsg This is currently fully based on a convention to not break existing MetricSets. We could return two objects or follow the approach from @urso in outputs with the `Data` struct: https://github.com/elastic/beats/blob/master/libbeat/outputs/outputs.go#L19

I would also like to see that the logic for the enrichment would be on the module level. Means we would have for example in the Docker module a method `AddMetaInfo` that has the logic inside to add this meta data based on the values provided by the MetricSets. This could allow more complex cases and not every MetricSet would have to add it to the event itself. This is very similar to the processors we discussed. Perhaps a module is also a Processor? ... Just some thoughts.

metricbeat/beater/event.go
```diff
@@ -48,21 +48,32 @@ func (b eventBuilder) build() (common.MapStr, error) {
...
-			"name":   b.metricSetName,
-			"rtt":    b.fetchDuration.Nanoseconds() / int64(time.Microsecond),
+			"module": b.ModuleName,
+			"name":   b.MetricSetName,
+			"rtt":    b.FetchDuration.Nanoseconds() / int64(time.Microsecond),
 		},
 	}
 
+	// In case meta data exists, it is added on the module level
+	if ok {
```
ruflin: I renamed it to `moduleData` and `moduleDataExists` to make it sure it does not get confused with the global meta data.

ruflin:
@tsg I changed the convention to `_module` as the other one already conflicted in the system module. `_module` is also more obvious when modifying the metricset. I pushed a new version and also updated the PR docs to not forget about it.

metricbeat/beater/event.go
```diff
@@ -15,31 +15,31 @@ const (
...
+	Host          string
+	StartTime     time.Time
+	FetchDuration time.Duration
+	Event         common.MapStr
 	fetchErr      error
 	filters       *processors.Processors
 	metadata      common.EventMetadata
 }
 
 // build builds an event from MetricSet data and applies the Module-level
```
andrewkroh: Consider diff'ing a before and after `golint ./... | grep -v vendor` output to look for new warnings.

metricbeat/beater/event.go
```diff
@@ -48,21 +48,32 @@ func (b eventBuilder) build() (common.MapStr, error) {
 		}
 	}
 
-	event = common.MapStr{
-		"@timestamp": timestamp,
-		"type":       typeName,
+	// Checks if additional meta information is provided by the MetricSet under the key _module
+	// This is based on the convention that each MetricSet can provide module data under the key _module
+	moduleData, moudleDataExists := event["_module"]
```
andrewkroh: Consider making `_module` an exported constant. It might need to be in the `mb` package in order for MetricSet implementation to use it.

metricbeat/beater/event.go
```diff
@@ -48,21 +48,32 @@ func (b eventBuilder) build() (common.MapStr, error) {
...
-			"rtt":    b.fetchDuration.Nanoseconds() / int64(time.Microsecond),
+			"module": b.ModuleName,
+			"name":   b.MetricSetName,
+			"rtt":    b.FetchDuration.Nanoseconds() / int64(time.Microsecond),
 		},
 	}
 
+	// In case meta data exists, it is added on the module level
+	if moudleDataExists {
+		event[b.ModuleName].(common.MapStr).Update(moduleData.(common.MapStr))
```
andrewkroh: You should test if it's a `common.MapStr` otherwise this could cause a panic.

tsg:
I like the `_module` convention more.

metricbeat/beater/event.go
ruflin: ugh, haven't checked golint for quite some time :-( Looks like Metricbeat needs some cleanup, especially the docker module. Will do that in an other PR (after fixing the parts of this PR).

metricbeat/beater/event.go
```diff
@@ -15,31 +15,31 @@ const (
 // eventBuilder is used for building MetricSet events. MetricSets generate a
```
ruflin: done

metricbeat/beater/event.go
ruflin: done

metricbeat/beater/event.go
ruflin: Good idea, I introduce `mb.MODULE_DATA`. This will simplify changes in the future.

metricbeat/beater/event.go
andrewkroh: I was referring to testing that `moduleData` is a common.MapStr. I didn't see where that was validated.

ruflin: Ah, got it. Will check that one instead ;-)

metricbeat/beater/event.go
andrewkroh: `MODULE_DATA` isn't a conventional Go name. Even constants should be [mixed case](https://github.com/golang/go/wiki/CodeReviewComments#mixed-caps).

ruflin: renamed to ModuleData

PR #2688, opened 2016-10-05
Set strings to be keywords by default

tsg:
This adds a dynamic mapping to all our template files to set strings to be
keywords by default.

Previously we were using the default of `text` and only switching to `keyword`
when configured in `fields.yml`. This reverses the logic and sets mappings to
`text` only when requested in `fields.yml`.

The goal is to make upgrading the mapping template less painful. In the Beats
we have so far, unexpected fields are better of as keywords.

### DIFF
--- a/filebeat/filebeat.template-es2x.json
+++ b/filebeat/filebeat.template-es2x.json
@@ -11,14 +11,13 @@
       },
       "dynamic_templates": [
         {
-          "fields": {
+          "strings_as_keyword": {
             "mapping": {
               "ignore_above": 1024,
               "index": "not_analyzed",
               "type": "string"
             },
-            "match_mapping_type": "string",
-            "path_match": "fields.*"
+            "match_mapping_type": "string"
           }
         }
       ],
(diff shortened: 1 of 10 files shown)


tsg:
I tested the change with ES 5.0, I'm the process of testing with ES 2.4.

monicasarbu:
LGTM. This is a great addition to the #2627, in order to have a better upgrading support.

tsg:
Tested with ES 2.4, seems to be working OK.

filebeat/filebeat.template-es2x.json
```diff
@@ -11,14 +11,13 @@
       },
       "dynamic_templates": [
         {
-          "fields": {
```
tsg: Yeah, it used to be the name of the dict/list field name. Now if everything is a keyword by default, the template also gets shorter, because we don't have dicts or lists of type `text`.

tsg:
Added a Changelog entry. This could go in when green from my point of view.

PR #2690, opened 2016-10-05
[metricbeat] Docker module : testing

douaejeouit:
Add unit test for both network and disk io metrics test

### DIFF
--- a/metricbeat/module/docker/cpu/cpu_test.go
+++ b/metricbeat/module/docker/cpu/cpu_test.go
@@ -12,7 +12,7 @@ import (
 	"github.com/elastic/beats/metricbeat/module/docker"
 )
 
-/*func TestCPUService_PerCpuUsage(t *testing.T) {
+func TestCPUService_PerCpuUsage(t *testing.T) {
 	//GIVEN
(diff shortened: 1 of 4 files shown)


douaejeouit:
@ruflin I added some unit tests. Could you please check it? Thanks

ruflin:
Only had a quick look, but I think some of these tests could fit very well into table driven tests: https://github.com/golang/go/wiki/TableDrivenTests

For the field checking: If you look at the system tests of other modules, there are some tests that automatically validate fields in the output based on the fields.yml file. This means the test will check if fields.yml is complete and if all fields exist. Like this we don't need to update the unit tests all the time when we change a field. Here is an example for mysql: https://github.com/elastic/beats/blob/master/metricbeat/tests/system/test_mysql.py#L41 We definitively need something like this for docker module and metricsets.

douaejeouit:
@ruflin +1 for the table driven tests.  I think it's a better option, I'll try to update unit tests!
Concerning the system test, I can see that it's written in Python. Unfortunattely It's a new langage for me, I need to develop my skills in Python as quickly as I can.  So, it may take a little bit more time than I think. But you can probabely deal with that @ruflin ?

ruflin:
@douaejeouit Sure, take your time. For the python tests, its mainly copy paste and being diligent about all the fields definitions.

Issue #2694, opened 2016-10-05
bgrewrite.last_status cannot contain '.'


trevorndodds:
Elasticsearch 2.4.1
Using metricbeat 5 beta 1 with the supplied 2.x template I'm unable to use the redis metricbeat module. 

"error"=>{"type"=>"mapper_parsing_exception", "reason"=>"Field name bgrewrite.last_status] cannot contain '.'"}

ruflin:
@trevorndodds Thanks for reporting this. I will create a fix for it.

ruflin:
https://github.com/elastic/beats/pull/2695 should solve the issue. I'm kind of surprised you get this issue in 2.4 and not previous versions and that the error is a mapper parsing exception. I still need to do some further investigations here.

ruflin:
@trevorndodds The issue happened in 2.4 as dots are disabled by default: https://www.elastic.co/guide/en/elasticsearch/reference/current/dots-in-names.html#_enabling_support_for_dots_in_field_names

trevorndodds:
@ruflin thanks! I did notice that 2.4 supported dots.in.names but I was hesitant to restart every node in the cluster with the flag for only a few metrics.

PR #2695, opened 2016-10-05
Fix redis metricset dots issue with ES 2.x

ruflin:
Not all 2.x elasticsearch releases support dots on field names. In the redis metricsets some dots instead of json objects were used as the outcome in elasticsearch 5.x is the same. All dots were removed and nested objects were created to make it compatible with elasticsearch 2.x.

Closes https://github.com/elastic/beats/issues/2694

### DIFF
--- a/metricbeat/module/redis/info/data.go
+++ b/metricbeat/module/redis/info/data.go
@@ -37,25 +37,39 @@ var (
 		"persistence": s.Object{
 			"loading": c.Bool("loading"),
 			"rdb": s.Object{
-				"last_save.changes_since": c.Int("rdb_changes_since_last_save"),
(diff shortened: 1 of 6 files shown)


metricbeat/module/redis/info/data.go
```diff
@@ -37,8 +37,10 @@ var (
 		"persistence": s.Object{
 			"loading": c.Bool("loading"),
 			"rdb": s.Object{
-				"last_save.changes_since": c.Int("rdb_changes_since_last_save"),
-				"last_save.time":          c.Int("rdb_last_save_time"),
+				"last_save": common.MapStr{
```
tsg: You probably need `s.Object` here. See compile errors.

PR #2700, opened 2016-10-06
Add beat.version field to all events

ruflin:
Having beat.version will simplify migration events from one version to an other as events can be for example reindex based on the version. This should also simplify debugging as events can be related to a beat version.

* Move version.go to its own package to prevent circular dependencies
* Add dashboard compiling to travis build to make sure it always works
* Update mappings with new field
* Fix local fields from elastic#2689


libbeat/publisher/client.go
```diff
@@ -76,6 +77,7 @@ func newClient(pub *BeatPublisher) *client {
 		beatMeta: common.MapStr{
 			"name":     pub.name,
 			"hostname": pub.hostname,
+			"version":  version.GetDefaultVersion(),
```
tsg: Is there any way for a community Beat to have their own version? I think we have a way in other places where the version is involved.

monicasarbu:
@ruflin Can you please rebase here, after the #2704 was merged?

libbeat/publisher/client.go
andrewkroh: Maybe the version can be injected into the `BeatPublisher` like the name is.

ruflin:
@monicasarbu already done.

libbeat/publisher/client.go
ruflin: @tsg good point, either I go the approach suggested by @andrewkroh or introduce a SetVersion to set the final version here: https://github.com/elastic/beats/pull/2700/files#diff-4cc56dbce8e84b1bc2a025380117788eR139

Looking into it. This should be definitively in sync with the beat version itself.

ruflin: @tsg @andrewkroh I rewrote it by passing through the param. Have a look at it again.

monicasarbu:
LGTM

monicasarbu:
I would say it's safe to backport it to 5.0

ruflin:
jenkins, retest it

PR #2689, opened 2016-10-05
Lower the font size of the metric visualizations for Metricbeat dashboards

monicasarbu:
This PR is meant to fix #2615, and comes with two big changes:

1. Set the format to bytes, instead of simple number for: `system.memory.actual.free`, `system.memory.actual.used.bytes` and percentage for `system.memory.actual.used.pct`.
2. Lower the font-size to 30pt for all metric visualizations

PR #2704, opened 2016-10-06
Update fields.yml and dashboards in the system module

monicasarbu:
This fix is requested by #2689 and updates the format of the `system.memory.actual.*locally in the system module. In addition, move the system dashboards to the system module.


ruflin:
LGTM

PR #2708, opened 2016-10-06
Add check to check command to assure update was applied

ruflin:
CI builds will go red in case make update / make collect were not applied

Add sorted() to os.listdir() as sorting order on OS X and Linux was different which leaded to diffs


ruflin:
jenkins, retest it

ruflin:
@douaejeouit The above changes the different diffs you had under windows when creating docker module. It seems Linux and MacOS X do sorting of files by default differently.

douaejeouit:
Ah !! it's good to know the reason! Thank you for informing me @ruflin !

PR #2709, opened 2016-10-06
fix a potential bug : massive cpu metrics for docker module

douaejeouit:
Negative values caused by the use of unsigned integers were transformed in huge positives value

### DIFF
--- a/metricbeat/module/docker/cpu/helper.go
+++ b/metricbeat/module/docker/cpu/helper.go
@@ -83,24 +83,24 @@ func (c *CPUService) PerCpuUsage(stats *dc.Stats) common.MapStr {
 	if cap(getNewCpu(stats).PerCpuUsage) == cap(getOLdCpu(stats).PerCpuUsage) {
 		output = common.MapStr{}
 		for index := range getNewCpu(stats).PerCpuUsage {
-			output[strconv.Itoa(index)] = c.calculateLoad(getNewCpu(stats).PerCpuUsage[index] - getOLdCpu(stats).PerCpuUsage[index])
+			output[strconv.Itoa(index)] = c.calculateLoad(int64(getNewCpu(stats).PerCpuUsage[index] - getOLdCpu(stats).PerCpuUsage[index]))
 		}
 	}
 	return output
 }
 
 func (c *CPUService) TotalUsage(stats *dc.Stats) float64 {
-	return c.calculateLoad(getNewCpu(stats).TotalUsage - getOLdCpu(stats).TotalUsage)
+	return c.calculateLoad(int64(getNewCpu(stats).TotalUsage - getOLdCpu(stats).TotalUsage))
 }
 
 func (c *CPUService) UsageInKernelmode(stats *dc.Stats) float64 {
-	return c.calculateLoad(getNewCpu(stats).UsageInKernelmode - getOLdCpu(stats).UsageInKernelmode)
+	return c.calculateLoad(int64(getNewCpu(stats).UsageInKernelmode - getOLdCpu(stats).UsageInKernelmode))
 }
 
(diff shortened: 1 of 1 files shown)


ruflin:
LGTM

ruflin:
jenkins, test it

metricbeat/module/docker/cpu/helper.go
```diff
@@ -83,24 +83,24 @@ func (c *CPUService) PerCpuUsage(stats *dc.Stats) common.MapStr {
...
+	return c.calculateLoad(int64(getNewCpu(stats).UsageInKernelmode - getOLdCpu(stats).UsageInKernelmode))
 }
 
 func (c *CPUService) UsageInUsermode(stats *dc.Stats) float64 {
-	return c.calculateLoad(getNewCpu(stats).UsageInUsermode - getOLdCpu(stats).UsageInUsermode)
+	return c.calculateLoad(int64(getNewCpu(stats).UsageInUsermode - getOLdCpu(stats).UsageInUsermode))
 }
 
-func (c *CPUService) calculateLoad(value uint64) float64 {
+func (c *CPUService) calculateLoad(value int64) float64 {
```
andrewkroh: What does this function do? The name doesn't seem to match up with what I expected. It looks more like it is converting the value given in ns to seconds? If that is true, are you sure the `value` is actually in nanoseconds? 

Normally the units on these values are in jiffies, but maybe Docker is converting them from jiffies to nanoseconds based on the `_SC_CLK_TCK` value from the OS.

metricbeat/module/docker/cpu/helper.go
```diff
@@ -83,24 +83,24 @@ func (c *CPUService) PerCpuUsage(stats *dc.Stats) common.MapStr {
...
-			output[strconv.Itoa(index)] = c.calculateLoad(getNewCpu(stats).PerCpuUsage[index] - getOLdCpu(stats).PerCpuUsage[index])
+			output[strconv.Itoa(index)] = c.calculateLoad(int64(getNewCpu(stats).PerCpuUsage[index] - getOLdCpu(stats).PerCpuUsage[index]))
 		}
 	}
 	return output
 }
 
 func (c *CPUService) TotalUsage(stats *dc.Stats) float64 {
-	return c.calculateLoad(getNewCpu(stats).TotalUsage - getOLdCpu(stats).TotalUsage)
+	return c.calculateLoad(int64(getNewCpu(stats).TotalUsage - getOLdCpu(stats).TotalUsage))
```
andrewkroh: So if this calculation results in a negative then this reports that the `TotalUsage` is negative? I think that if the difference between the usages is negative then the data we have is invalid, right? These stats are counters so the only time that the difference should be negative is when the counter rolls over.

So maybe the `calculateLoad` function should check `if value < 0 { return 0 }`.

douaejeouit: +1

metricbeat/module/docker/cpu/helper.go
```diff
@@ -83,24 +83,24 @@ func (c *CPUService) PerCpuUsage(stats *dc.Stats) common.MapStr {
 	if cap(getNewCpu(stats).PerCpuUsage) == cap(getOLdCpu(stats).PerCpuUsage) {
```
douaejeouit: With pleasure !
Please check  PR [#2734] (https://github.com/elastic/beats/pull/2734), I replied your question.

metricbeat/module/docker/cpu/helper.go
douaejeouit: Indeed. The value passed in parametrs is the count of CPU nanosecond in 1sec, so we convert it in sec.

PR #2711, opened 2016-10-06
Remove confusing error debug message

ruflin:
As the state conversion was also applied on an empty state, in debug mode a confusing messages was printed out. This message is now removed by checking if the file has more then empty state inside.


ruflin:
@monicasarbu I don't think this is worth backporting as it is a debug log message and only appears when filebeat is started the first time. As most people do not run in debug mode they will never see this message.

PR #2714, opened 2016-10-06
Add searchable and aggregatable fields to the index patterns

### LINKED ISSUE #2713: Add "searchable" and "aggregatable" to the generated index patterns
For the index patterns we generate we need to add the "searchable" and "aggregatable" fields, which are new in the latest Kibana version, otherwise https://github.com/elastic/kibana/issues/8568 happens.

tsg:
For simplicity, searchable is always true, aggregatable is false
for text fields, and true otherwise. We should improve on the logic
on future PRs, but this should make it work good enough.

Closes #2713.


monicasarbu:
LGTM. Thanks for the quick fix.

spalger:
👍

PR #2726, opened 2016-10-07
Add a note to refresh the index pattern before visualizing the data

monicasarbu:
Packetbeat uses a few fields in the event of type `dict`. The problem appears as the Beat doesn't always know the keys that a dictionary field might contain, so it's not able to define them in advance.

Let's consider as an example `http.request.headers`, that has the type `dict`. After importing the index-pattern in Kibana, the Discovery page in Kibana looks like:

screen shot 2016-10-06 at 10 08 32 pm

After you refresh the `packetbeat-*` index pattern, then all the keys are updated and the Discovery page looks like:

screen shot 2016-10-07 at 3 58 06 pm

where a field is created in the `packetbeat-*` index pattern for each key available in the dictionary.

cc-ed @dedemorton

### DIFF
--- a/libbeat/docs/faq-refresh-index.asciidoc
+++ b/libbeat/docs/faq-refresh-index.asciidoc
@@ -1,0 +1,23 @@
+[float]
+[[refresh-index-pattern]]
+=== Fields show up as nested JSON in Kibana
+
+In case the Beat exports a field of type dictionary, and the keys are not known in advance, the Discovery page from Kibana will display it as a nested JSON object:
+
+[source,shell]
+----------------------------------------------------------------------
+http.response.headers = {
+        "content-length": 12,
+        "content-type": "application/json"
+}
+----------------------------------------------------------------------
+To fix this you need to https://www.elastic.co/guide/en/kibana/5.0/settings.html#reload-fields[reload the index pattern] in Kibana under the Management->Index Patterns, and the index-pattern will be
+updated with a field for each key available in the dictionary:
+
+[source,shell]
+----------------------------------------------------------------------
+http.response.headers.content-length = 12
+http.response.headers.content-type = "application/json"
+----------------------------------------------------------------------
+
+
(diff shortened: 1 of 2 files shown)


libbeat/docs/dashboards.asciidoc
```diff
@@ -102,6 +102,10 @@ predefined +{beatname_lc}-*+ index pattern and set it as the default.
 
 image:./images/kibana-created-indexes.png[Kibana configured indexes]
 
+NOTE: After importing the index-pattern in Kibana, you need to refresh each index-pattern in Kibana in order to update
```
ruflin: In most cases probably directly after the import the refresh will not help as not much data is in the index yet. Only over time these fields will show up and as far as I understand every time a new dict with fields inside is added, the refresh is needed. Perhaps we should put this under the FAQ in something like "Fields show up a nested JSON in Kibana"?

dedemorton: @monicasarbu I think this note will be confusing to new users. It would be better, as @ruflin suggests, to describe how the problem affects the user (what they see when they try to visualize the data) and how to fix the problem. I'd suggest having a short note here and then pointing off to an FAQ topic for more detail (in the FAQ topic, you could have a more detailed example of the problem).

BTW, when you say that they need to refresh the index pattern, do you mean that they need to do this: https://www.elastic.co/guide/en/kibana/current/settings.html#reload-fields? If so, a pointer to the Kibana doc about this would be useful.

monicasarbu: @ruflin The section about loading the dashboards comes after the user has data in Elasticsearch as the previous section is "Starting Packetbeat", if the user follows the Getting started guide step by step. This is tricky as anytime, new keys can appear after you refreshed already the index-pattern.
I think adding a question under FAQ is a good idea.

monicasarbu: @dedemorton yes. I am referring to the indices refresh from Kibana. As suggested I also added to FAQ.

monicasarbu:
Based on the feedback from @ruflin @dedemorton, I am adding only a section under FAQ with the solution for nested objected under Discovery, and don't add anything in the importing dashboards section as the user might get confused at this early stage.

PR #2728, opened 2016-10-07
Add EC2, GCE, or DigitalOcean metadata to events

andrewkroh:
This introduces a new processor called `add_cloud_metadata` that detects
the hosting provider, caches instance metadata, and enriches each event
with the data. There is one configuration option named `timeout` that
configures the maximum amount of time metadata fetching can run for. The 
default value is 3s. 

Sample config:

```
processors:
- add_cloud_metadata:
    #timeout: 3s
```
Sample data from the providers:

```
{
  "meta": {
    "cloud": {
      "availability_zone": "us-east-1c",
      "instance_id": "i-4e123456",
      "machine_type": "t2.medium",
      "provider": "ec2",
      "region": "us-east-1"
    }
  }
}
```

```
{
  "meta": {
    "cloud": {
      "instance_id": "1234567",
      "provider": "digitalocean",
      "region": "nyc2"
    }
  }
}
```

```
{
  "meta": {
    "cloud": {
      "availability_zone": "projects/1234567890/zones/us-east1-b",
      "instance_id": "1234556778987654321",
      "machine_type": "projects/1234567890/machineTypes/f1-micro",
      "project_id": "my-dev",
      "provider": "gce"
    }
  }
}
```


libbeat/processors/actions/cloud_metadata.go
```diff
@@ -0,0 +1,224 @@
...
+)
+
+var (
+	metaClient = &http.Client{
+		Transport: &http.Transport{
+			Dial: (&net.Dialer{
+				Timeout:   2 * time.Second,
+				KeepAlive: 0, // We are only making a single request.
+			}).Dial,
+			ResponseHeaderTimeout: 2 * time.Second,
```
monicasarbu: @andrewkroh How did you choose 2 here? Should be 2 < 5, right?

monicasarbu:
@andrewkroh This is a great PR. I have a few minor comments here. 

I would suggest to add an example in the `.full.yml` with `cloud_metadata` processor. 
As the other actions are `drop_*` or `include_*`, starting with a verb, I suggest to change the name from `cloud_metadata` to something similar, maybe `add_metadata` with an option `cloud: true` or just `add_cloud_metadata`?

libbeat/processors/actions/cloud_metadata.go
andrewkroh: I chose 5 to allow the individual requests to either complete or timeout on their own. In the worst case, timeout could take ~4 seconds (2 seconds for connect timeout + 2 seconds for response header timeout). I left a 1 second buffer since it's execute three requests in parallel and they might not all be scheduled immediately.

andrewkroh: I changed the timeout implementation to make it configuration so this code is now different and relies on the `http.Client.Timeout`.

libbeat/processors/actions/cloud_metadata.go
```diff
@@ -0,0 +1,224 @@
...
+// be returned or for a timeout to occur then returns the results that
+// completed in time.
+func fetchMetadata() []result {
+	c := make(chan result)
+	go func() { c <- fetchJSON("digitalocean", nil, digitaloceanMetadataURL, digitaloceanSchema) }()
+	go func() { c <- fetchJSON("ec2", nil, ec2MetadataURL, ec2Schema) }()
+	go func() { c <- fetchJSON("gce", gceHeaders, gceMetadataURL, gceSchema) }()
+
+	var results []result
+	timeout := time.After(5 * time.Second)
```
andrewkroh: The timeout is now configurable and defaults to 3s. In practice when running on EC2 the request is completed in 2ms.

andrewkroh:
I renamed the processor to `add_cloud_metadata`. I added it to the config.full.yml.

libbeat/_meta/fields.yml
```diff
@@ -45,3 +45,36 @@ fields:
...
+  description: >
+    Metadata from cloud providers added by the add_cloud_metadata processor.
+  fields:
+
+    - name: cloud.provider
+      example: ec2
+      description: >
+        Name of the cloud provider. Possible values are ec2, gce, or digitalocean.
+
+    - name: cloud.instance_id
```
ruflin: We should use `cloud.instance.id` and `cloud.instance.type` here to follow our naming schema.

libbeat/processors/actions/add_cloud_metadata.go
```diff
@@ -0,0 +1,246 @@
...
+	ec2Schema      = s.Schema{
+		"instance_id":       c.Str("instanceId"),
+		"instance_type":     c.Str("instanceType"),
+		"region":            c.Str("region"),
+		"availability_zone": c.Str("availabilityZone"),
+	}.Apply
+
+	digitaloceanMetadataURL = "http://" + digitaloceanMetadataHost + digitaloceanMetadataURI
+	digitaloceanSchema      = s.Schema{
+		"instance_id": c.StrFromNum("droplet_id"),
```
ruflin: `instance.id`

libbeat/processors/actions/add_cloud_metadata.go
```diff
@@ -0,0 +1,246 @@
...
+	}.Apply
+
+	gceHeaders     = map[string]string{"Metadata-Flavor": "Google"}
+	gceMetadataURL = "http://" + gceMetadataHost + gceMetadataURI
+	gceSchema      = func(m map[string]interface{}) common.MapStr {
+		out := common.MapStr{}
+
+		if instance, ok := m["instance"].(map[string]interface{}); ok {
+			s.Schema{
+				"instance_id":       c.StrFromNum("id"),
```
ruflin: See above for the naming.

metricbeat/schema/mapstriface/mapstriface_test.go
```diff
@@ -4,6 +4,8 @@ import (
 	"testing"
 	"time"
 
+	"encoding/json"
```
ruflin: remove newline on top

packetbeat/packetbeat.full.yml
```diff
@@ -515,13 +517,19 @@ packetbeat.protocols.nfs:
 #       equals:
 #           http.code: 200
 #
+# The following example enriches each event with metadata from the cloud
+# provider about the host machine. It works on EC2, GCE, and DigitalOcean.
+#
+#processors:
+#-add_cloud_metadata:
```
ruflin: space after `-`

monicasarbu:
I agree with @ruflin and have either a config option under `add_cloud_metadata` to choose the provider or do autodiscovery where the provider is detected automatically. Also, I agree that exporting all these information under `cloud` is a bit too generic, and I suggest `meta.cloud`.

urso:
This PR introduces some kind of lookup processor as [x-exec-lookup](https://github.com/elastic/beats/tree/x-exec-lookup) branch does. For this we introduced some namespacing on filters.  e.g. `x-exec-lookup` registers to processor `lookup.exec`. For consistency reasons I'd propose to register to `lookup.cloud_metadata`.

andrewkroh:
@monicasarbu @ruflin thanks for reviewing.

> Is there a specific reason for the 3s default timeout? If not I suggest we use the same as we use for other "metrics" in Metricbeat as default which is 10s.

This processor runs once at startup and auto-detects the cloud provider using three HTTP requests executed in parallel. If the Beat is running in the cloud then it can usually reach a disposition in ~2ms. If it's not running in the cloud then these requests will usually timeout because because there is no route to the metadata services which run on a special [link-local](https://tools.ietf.org/html/rfc3927) IP address. 

I don't think there is a need to increase the timeout. It should be able to reliably reach a disposition in that 3s window (probably even shorter would be fine). This would allow you to add the processor to all of your deployments whether they be on-prem or in the cloud without much of a penalty.

> I would suggest we either split this one processor up into 4 processors or make the type configurable. We could have a processor for each cloud type + 1 one which does the auto detection. Currently as far as I understand it checks every time all 3 options? We could also have a config option inside the processor which defines the type

I really don't think this is necessary. Can we put this into master without these feature, let it get used a bit, then see if these is necessary?

> Namespace: This processor reserves the namespace cloud. I'm kind of worried that we this could conflict with other things like a potential cloud module in metricbeat or some data in other beats.

This can definitely be a problem. After looking at the exec lookup feature I think I will default this to writing the data under `fields`, add a `fields_under_root` option, and provide a way to configure the key which defaults to `cloud`. I am also going to namespace the processor as `lookup.cloud_metadata` as suggested by @urso.

> Do you plan to add docs for this in an other PR?

I'll write the docs in a second PR after the code and behavior is finalized.

ruflin:
@andrewkroh Thanks for the details.

> This processor runs once at startup

That is the part I missed. I confused the timeout with how often it runs. I thought it updates the meta data every 3s. In this case `3s` or lower makes total sense.

This also answers the second part about the config options. If it is only run once, there is very low overhead of running all 3.

> I will default this to writing the data under fields

I don't think we should mix manually added data from the user and machine generated data. That is why I would prefer NOT to put it under `fields` but find an other namespace. Also I would not provide a `fields_under_root` option as this will only lead to problems with overwriting fields and will invalidate our predefined templates. I think I don't fully understand the advantage of having `fields_under_root` option. We could put it under `lookup.cloud_metadata` namespace. This makes a logical connection between the processor and the data itself. Having auto generation of templates from processors in mind (long term idea) this would make things easier ;-)

libbeat/_meta/fields.yml
andrewkroh: After further thought, I changed instance_type to machine_type (follow GCE's naming). So I don't think we need to make it `instance.id` now.

metricbeat/schema/mapstriface/mapstriface_test.go
andrewkroh: Fixed.

packetbeat/packetbeat.full.yml
andrewkroh: Added.

andrewkroh:
I pushed a change to rename the processor as `lookup.cloud_metadata`. We just need to discuss where the data shall go.

andrewkroh:
This PR has been updated based on our discussions.

- Processor name is `add_cloud_metadata`.
- The data is added to events under `meta.cloud`. See PR description (top) for full examples.

ruflin:
LGTM. I think we should also add this change to the CHANGELOG

andrewkroh:
Added this to the CHANGELOG.

urso:
I'm not sure I agree with the chosen namespaces here.

Why use `add_cloud_metadata` instead of `lookup.cloud_metadata`?

Why choose `meta.cloud`? I'd consider `meta` namespace to be quite common regarding filebeat+json or possibly some future metricset.

Considering more processors being added in future + having some more options to add custom fields in beats (e.g. filebeat prospector, packetbeat modules, processors) I'd opt for some general guidelines regarding namespacing here. Not saying one naming is better or worse then another, but mostly striving for consistency and some general agreement here.

Instead of using `fields_under_root` or some non-configurable top-level name, how about making the namespace for additional fields to be added configurable `namespace`? This option can be reused for `fields` settings as well as for lookup like processors.

tsg:
We've discussed these things on Wednesday and the discussion was about to go long (like it usually does on things like this), so we decided to go with one of the options, knowing that we still have time to change it before this sees the light of day in 5.1.

So lets continue the discussion, although this is PR is not the best place for it.

PR #2732, opened 2016-10-10
Update redis dependency to 3.2.4

ruflin:
Make sure libbeat and metricbeat use the same version


tsg:
We state in the docs that we test against `3.2.0`. You should also upgrade that to 3.2.4

ruflin:
Done

PR #2734, opened 2016-10-10
[docker] improvement : fix cpu negative value issue & update some unit tests

douaejeouit:


metricbeat/module/docker/cpu/cpu_test.go
```diff
@@ -75,7 +77,12 @@ func TestCPUService_UsageInUsermode(t *testing.T) {
 	assert.Equal(t, float64(0.50), result)
 }
 
-/* TODO: uncomment
+//TODO: uncomment
```
ruflin: This line can be removed now?

metricbeat/module/docker/cpu/cpu_test.go
```diff
@@ -75,7 +77,12 @@ func TestCPUService_UsageInUsermode(t *testing.T) {
...
+/**
```
ruflin: For method comments I recommend to use // instead of /* ... */

metricbeat/module/docker/cpu/cpu_test.go
douaejeouit: Yes, in the future commits !

metricbeat/module/docker/cpu/cpu_test.go
douaejeouit: Ok!

metricbeat/module/docker/cpu/helper.go
```diff
@@ -80,27 +80,30 @@ func getNewCpu(stats *dc.Stats) CPURaw {
 
 func (c *CPUService) PerCpuUsage(stats *dc.Stats) common.MapStr {
 	var output common.MapStr
-	if cap(getNewCpu(stats).PerCpuUsage) == cap(getOLdCpu(stats).PerCpuUsage) {
+	if cap(getNewCpu(stats).PerCpuUsage) == cap(getOldCpu(stats).PerCpuUsage) {
```
andrewkroh: Why `cap()` and not `len()`? https://github.com/elastic/beats/pull/2709#discussion_r82501857

The range loop that follows produces iteration values from 0 up to `len(x)-1`. So even though their capacities are equal that does not guarantee that their `len()`'s are equal. This could lead to an `index out of range` runtime error.

douaejeouit: @andrewkroh, Theoretically & technically you're asolutely right! 
But the types i'm comparing are both arrays ( uint64  arrays ) and the documentation of the moethod cap(v Type) says that this function returns the capacity of v according to it's type. so when it's about arrays, it returns the number of elements in v, same as len(v). So the probability to have  an index out of range runtime error tend towards zero! correct me if i'm wrong !
https://golang.org/pkg/builtin/#cap

douaejeouit: @andrewkroh, Theoretically & technically you're asolutely right! 
But the types i'm comparing are both arrays ( uint64 arrays ) & not slices! Which mean that it doesn't return the max lengh the slice can reach but it returns the number of elements in the array. According to the documentation of the moethod cap(v Type), it confirms that function returns the capacity of v according to it's type. so when it's about arrays, it returns the number of elements in v, same as len(v). So the probability to have an index out of range runtime error tend towards zero! correct me if i'm wrong !
https://golang.org/pkg/builtin/#cap

douaejeouit:
@ruflin Could you please check the new updates? thanks.

metricbeat/module/docker/cpu/helper.go
andrewkroh: > But the types i'm comparing are both arrays ( uint64 arrays ) & not slices!

No, they are slices. An array would be declared as `[n]type` where as a slice is `[]type`.

Here's a test and demonstrates a panic.

```
func TestPerCpuUsagePanic(t *testing.T) {
	stats := &dc.Stats{}
	stats.PreCPUStats.CPUUsage.PercpuUsage = make([]uint64, 1, 2)
	stats.CPUStats.CPUUsage.PercpuUsage = make([]uint64, 2)

	NewCpuService().PerCpuUsage(stats) // Panic
}
```

douaejeouit: Ok. Now i'm convinced :) Thanks for the explainations 👍 👍 
I'll change it if you don't mind !

PR #2737, opened 2016-10-10
Kafka output drop invalid messages

urso:
- Fix too big messages not being dropped, but forcing continuous resends
- If publishing subset of batches fails, only attempt to resend failed events instead of complete batch


ruflin:
LGTM

PR #2743, opened 2016-10-10
Add heartbeat (experimental)

urso:


pmoust:
Great, 👍

ruflin:
@urso Can you please open a meta issue in which we can track things that should still be done for heartbeat like docs etc?

PR #2744, opened 2016-10-10
Point to the upgrade docs from the docs for each beat

dedemorton:
I'm concerned that existing Beats users will not notice the upgrade instructions in the Platform Reference, so I've added short topics to the doc for each Beat to point to the Platform Reference. I thought about just adding links to the Getting Started, but wanted to raise visibility of the upgrade content by making the topic a main entry in the nav.

I didn't bother single sourcing the content because duplication seemed OK given that there's so little content here.


tsg:
LGTM, agreed it makes sense to link from the Beats guides.

Issue #2646, opened 2016-09-27
Filebeat 5.0.0-beta1 not always removing deleted files from registry


doingitbigdata:
- Version: filebeat version 5.0.0-beta1 (amd64), libbeat 5.0.0-beta1
- Operating System: Red Hat Enterprise Linux Server release 6.5 (Santiago)

I'm having an issue where a deleted file is not being removed from the registry. The entry in the registry persists and I can see the timestamp for the entry increasing over time.

I do not have the clean_removed parameter specified in filebeat.yml therefore it should be enabled by default. 

I do have the clean_inactive and ignore_older parameters specified and I am deleting the file before these time periods have been reached. I am able to reproduce this result repeatedly.

I do not see this issue when I don't have the clean_inactive and ignore_older parameters specified.

ruflin:
Can you share your full config file? It would also be very helpful to get the log file from filebeat if available? Can you also share the steps you take to reproduce it?

passing:
having the same issue using filebeat-5.0.0-beta1-linux-x86_64

our configuration is:
```
filebeat.prospectors:
- input_type: log
  paths:
    - /nasdevices/jobs-logs/logs/*/*/*.log
  document_type: application_log
  multiline:
    pattern: ^[0-9]{4}-[0-9]{2}-[0-9]{2} [0-9]{2}:[0-9]{2}:[0-9]{2},[0-9]{3} \[
    negate: true
    match: after
    max_lines: 500
    timeout: 5s
  ignore_older: 24h
  clean_inactive: 25h
  close_removed: true
  clean_removed: true
name: jobs-logs-shipper
tags: ["log4j"]
output.logstash:
  hosts: ["10.0.0.99:5044"]
```

using this script to check for files in the registry that are not on disk any more:
```
#!/bin/bash

total=0
existing=0
missing=0

tempfile=`mktemp`
sed -e "s/},{/\n/g" data/registry | grep -oE "/nas[^\"]*.log" > $tempfile

while read file
do
	if [ -f $file ]
	then
		((existing++))
	else
		((missing++))
	fi
	((total++))
done < $tempfile

rm $tempfile

echo "existing: $existing"
echo "missing:  $missing"
echo "total:    $total"
```
the number of 'missing' files is growing slowly over time

ruflin:
@doingitbigdata @passing Thanks for all the details. I managed to reproduce this and will hopefully manage to provide a fix soonish.

ruflin:
[...]

PR #2750, opened 2016-10-11
Fix issue when clean_removed and clean_inactive were used together

ruflin:
If clean_removed and clean_inactive were used together, states were not directly removed from the registry. The clean_inactive option was updating the TTL after clean_removed was applied. This lead to the issue that during every run clean_removed updated the state and clean_inactive reset the TTL. In the end removed files were never purged from the registry if both options were used together.

* Test added to verify fix
* Updated incorrect log message

Closes https://github.com/elastic/beats/issues/2646

### DIFF
--- a/filebeat/prospector/prospector.go
+++ b/filebeat/prospector/prospector.go
@@ -159,8 +159,8 @@ func (p *Prospector) Run(once bool) {
 // All state updates done by the prospector itself are synchronous to make sure not states are overwritten
 func (p *Prospector) updateState(event *input.Event) error {
 
-	// Add ttl if cleanOlder is enabled
-	if p.config.CleanInactive > 0 {
+	// Add ttl if cleanOlder is enabled and TTL is not already 0
+	if p.config.CleanInactive > 0 && event.State.TTL != 0 {
 		event.State.TTL = p.config.CleanInactive
 	}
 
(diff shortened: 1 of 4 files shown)


filebeat/prospector/prospector.go
```diff
@@ -159,8 +159,8 @@ func (p *Prospector) Run(once bool) {
 // All state updates done by the prospector itself are synchronous to make sure not states are overwritten
 func (p *Prospector) updateState(event *input.Event) error {
 
-	// Add ttl if cleanOlder is enabled
-	if p.config.CleanInactive > 0 {
+	// Add ttl if cleanOlder is enabled and TTL is not already 0
+	if p.config.CleanInactive > 0 && event.State.TTL != 0 {
```
andrewkroh: I'm wondering if the new check should be for `event.State.TTL > 0` since event.State.TTL starts as a `-1 * time.Second`.

ruflin: The idea here is that TTL is reset every time the state is updated. As all new states are initialies with -1 this would mean these states would never get a TTL according to `clean_inactive`. My first idea was to have `state.Remove = true` as a flag for states that can be removed but it would blow up the change. But this would clearly show the difference between TTL and states that should be removed. TTL = 0 is like an alias for the Removed option and is only set when one of the clean options is reached.

PR #2752, opened 2016-10-11
MapStr enhancements/cleanup

andrewkroh:
- Made all MapStr methods work when it contains a nested map[string]interface{}. This is useful if you want to operate directly on unmarshalled JSON. (and the reason I started with these changes)
- Added MapStr.Put() method that handles dot-notation in keys.
- For consistency with other MapStr methods, MapStr.CopyFieldsTo() now returns an error if the specified key does not exist. If you want to ignore missing key errors then you must check for ErrKeyNotFound.
- Removed unused methods MapStr.EnsureCountField() and MapStr.EnsureTimestampField().

### DIFF
--- a/libbeat/common/mapstr_test.go
+++ b/libbeat/common/mapstr_test.go
@@ -4,7 +4,6 @@ package common
 
 import (
 	"testing"
-	"time"
 
 	"github.com/stretchr/testify/assert"
 )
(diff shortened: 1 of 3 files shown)


libbeat/common/mapstr_test.go
```diff
@@ -64,7 +63,7 @@ func TestMapStrCopyFieldsTo(t *testing.T) {
 	c := MapStr{}
 
 	err := m.CopyFieldsTo(c, "dd")
-	assert.Equal(nil, err)
+	assert.Error(err)
```
ruflin: Didn't the previous check assert that Error is nil, and now it verifies that it is no error? The equivalent would be `assert.NoError(err)`?

ruflin: Ah, just saw you changed the behaviour. All good.

PR #2754, opened 2016-10-11
Fix issue when clean_removed and clean_inactive were used together

ruflin:
The clean_inactive option was updating the TTL after clean_removed was applied. This lead to the issue that during every run clean_removed updated the state and clean_inactive reset the TTL. In the end removed files were never purged from the registry if both options were used together.

* Test added to verify fix
* Updated incorrect log message

Closes https://github.com/elastic/beats/issues/2646


tsg:
Backport of #2750.

PR #2631, opened 2016-09-23
Breaking changes and Upgrading docs

tsg:
A first draft, containing general infos about upgrading but also
specific information about the upgrade from 1.x to 5.x.

PR #2767, opened 2016-10-12
Edit upgrade and breaking changes doc

dedemorton:
@tsg Please review my edits. The content is very nicely written, so there aren't many changes!

Note that we're using title case for headings in the Beats docs, so I changed the section titles to be consistent, but someday we'll standardize the library on one style, which I hope will be sentence case. Don't worry about it, though. Nobody cares about heading capitalization but editors and writers. :-)

### DIFF
--- a/libbeat/docs/breaking.asciidoc
+++ b/libbeat/docs/breaking.asciidoc
@@ -1,5 +1,5 @@
 [[breaking-changes]]
-== Breaking changes
+== Breaking Changes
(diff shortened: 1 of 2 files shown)


dedemorton:
Implements edits for upgrade and breaking changes docs added in #2631 and tracked in #2482.

tsg:
LGTM, thanks for the editing!

PR #2769, opened 2016-10-13
Fix packaging issue with old git version on debian 6

ruflin:
Git 1.7 on debian cannot handle packages which are behind a proxy. As go-yaml is fetched through a proxy this will make the build hang. As a work around gotpl is cloned manually and go-yaml is copied from the already vendored directory.


ruflin:
jenkins, package it

ruflin:
jenkins, package it

Issue #2079, opened 2016-07-21
Count each unique mountpoint once in fsstasts


zyrill:
- Version: # metricbeat version 5.0.0-alpha4 (amd64), libbeat 5.0.0-alpha4
- Operating System: CentOS Linux release 7.2.1511 (Core)
- Steps to Reproduce: activate fsstat metricset, switch logging to debug, compare with df output

Following the full info.
metricbeat Config:
```

###################### Metricbeat Configuration Example #######################
#==========================  Modules configuration ============================
metricbeat.modules:
- module: system
  metricsets:
    - cpu
    #- core
    - diskio
    #- filesystem
    - fsstat
    - memory
    - network
    #- process
  enabled: true
  period: 10s
  processes: ['.*']
  cpu_ticks: false

- module: nginx
  metricsets: ['stubstatus']
  enabled: true
  period: 10s
  hosts: ['http://127.0.0.1']
  server_status_path: '/nginx_status'

#================================ General =====================================
#name:
#tags: ["service-X", "web-tier"]
#env: staging

#================================ Outputs =====================================
#-------------------------- Elasticsearch output ------------------------------
#output.elasticsearch:
  #hosts: ["localhost:9200"]
  #template.name: "metricbeat"
  #template.path: "metricbeat.template.json"
  #template.overwrite: false

#----------------------------- Logstash output --------------------------------
output.logstash:
  hosts: ["10.255.210.5:5044"]
  tls.certificate_authorities: ["/etc/pki/tls/certs/logstash-forwarder.crt"]
  #tls.certificate: "/etc/pki/client/cert.pem"
  #tls.certificate_key: "/etc/pki/client/cert.key"

#================================ Logging =====================================
#logging.level: debug
logging.level: error
logging.to_files: true
logging.to_syslog: false
logging.files:
  path: /var/log/metricbeat/
  name: metricbeat.log
  keepfiles: 7
```

A relevant entry of the metricbeat log:

```
2016-07-21T14:33:34+02:00 DBG  Publish: {
  "@timestamp": "2016-07-21T12:33:34.783Z",
  "beat": {
    "hostname": "CCT-WEB008-C7V",
    "name": "CCT-WEB008-C7V"
  },
  "metricset": {
    "module": "system",
    "name": "fsstat",
    "rtt": 379
  },
  "system": {
    "fsstat": {
      "count": 32,
      "total_files": 1410425,
      "total_size": {
        "free": 23922294784,
        "total": 35769786368,
[...]

PR #2777, opened 2016-10-13
Count only the first filesystem with the same mounting point

monicasarbu:
Trying to solve the issue #2079, by calculating the `fsstat` values per mounting point, and not per filesystem. If multiple filesystems point to the same mounting point, then only one would be counted.


tsg:
LGTM.

tsg:
Needs a rebase.

monicasarbu:
The rebase is done, and it should be green now.

PR #2780, opened 2016-10-14
Document load balancing modes in filebeat

### LINKED ISSUE #1852: Document load-balancing modes in filebeat
Filebeat can operate load-balancing in 3 different modes:
1. send to one host after another
2. send to N hosts in lock-step (with `spool_size = N * bulk_max_size`)
3. send fully in parallel and async (`publish_async: true`)

All modes employ a difference in memory/cpu usage. With 1. requiring the least and 3. requiring the most resources.

dedemorton:
Added to resolve #1852

### DIFF
--- a/filebeat/docs/load-balancing.asciidoc
+++ b/filebeat/docs/load-balancing.asciidoc
@@ -1,0 +1,69 @@
+[[load-balancing]]
+== Configuring Load Balancing
+
+//REVIEWERS: Which outputs support load balancing? I only see the loadbalance option described under Logstash and Redis. Are any other outputs supported?
+
+Filebeat provides configuration options that you can use to fine
+tune load balancing when sending events to multiple hosts. 
+
+To enable load balancing, you specify `load_balance: true` when you configure
+the output. For example:
+
+[source,yaml]
+-------------------------------------------------------------------------------
+output.logstash:
+  hosts: ["localhost:5044", "localhost:5045"]
+  loadbalance: true
+-------------------------------------------------------------------------------
+
+Filebeat can perform load balancing in a few different modes:
+
+* **Send events to one host after another:**
++
+By default, when you configure Filebeat to send events to multiple hosts
+(`load_balance: true`), Filebeat will send the events to one host after
+another. This mode requires the least memory and CPU usage.
+
+* **Send events to `N` hosts in lock-step:**
++
+You can configure Filebeat to send events to `N` hosts in lock-step by setting
+`spool_size = N * bulk_max_size`. This mode requires more memory and CPU usage
+than the previous mode.
++
+Example:
++
+[source,yaml]
+-------------------------------------------------------------------------------
+filebeat.prospectors:
+- input_type: log
+  paths:
+    - /var/log/*.log
+filebeat.spool_size: 4096
+output.logstash:
+  hosts: ["localhost:5044", "localhost:5045"]
+  loadbalance: true
+  bulk_max_size: 2048
+-------------------------------------------------------------------------------
+
+
+//REVIEWERS: Please confirm the accuracy of the config examples here. I'm guessing at the config and haven't tested these examples. It would probably be better to use a more realistic path for /var/log/*.log (please suggest one). I'd like to show the full config because it helps users understand where the settings fit into the overall yaml file.
+
+//REVIEWERS: I'm a bit confused in the above example because bulk_max_size does not appear under output.logstash in filebeat.full.yml. Is this intentional? It does show up as an option in https://www.elastic.co/guide/en/beats/filebeat/5.0/logstash-output.html, so I assume it's valid, correct?
+
+* **Send events in parallel and asynchronously:**
++
+You can configure Filebeat to send events in parallel and asynchronously by setting `publish_async: true`. This mode can improve load-balancing throughput, but requires the most memory and CPU usage.
++
+Example:
++
+[source,yaml]
+-------------------------------------------------------------------------------
+filebeat.prospectors:
+- input_type: log
+  paths:
+    - /var/log/*.log
+filebeat.publish_async: false
+output.logstash:
+  hosts: ["localhost:5044", "localhost:5045"]
+  loadbalance: true
+-------------------------------------------------------------------------------
--- a/filebeat/docs/load-balancing.asciidoc
+++ b/filebeat/docs/load-balancing.asciidoc
@@ -1,12 +1,10 @@
 [[load-balancing]]
-== Configuring Load Balancing
-
-//REVIEWERS: Which outputs support load balancing? I only see the loadbalance option described under Logstash and Redis. Are any other outputs supported?
+== Load Balancing
 
 Filebeat provides configuration options that you can use to fine
 tune load balancing when sending events to multiple hosts. 
 
-To enable load balancing, you specify `load_balance: true` when you configure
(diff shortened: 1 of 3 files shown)


tsg:
Let's better wait on @urso to review this one.

filebeat/docs/load-balancing.asciidoc
```diff
@@ -0,0 +1,69 @@
+[[load-balancing]]
+== Configuring Load Balancing
+
+//REVIEWERS: Which outputs support load balancing? I only see the loadbalance option described under Logstash and Redis. Are any other outputs supported?
```
urso: redis, logstash and elasticsearch.

Note: Kafka internally uses the same load-balancer module, but only for asynchronous error handling. Loadbalancing for kafka is internally solved (a little different) by client library used.

filebeat/docs/load-balancing.asciidoc
```diff
@@ -0,0 +1,69 @@
...
+  hosts: ["localhost:5044", "localhost:5045"]
+  loadbalance: true
+-------------------------------------------------------------------------------
+
+Filebeat can perform load balancing in a few different modes:
+
+* **Send events to one host after another:**
++
+By default, when you configure Filebeat to send events to multiple hosts
+(`load_balance: true`), Filebeat will send the events to one host after
```
urso: s/load_balance/loadbalance/

plus: 
- `filebeat.publish_async: false`
- `filebeat.spool_size = output.X.bulk_max_size`

The host choosen for the next batch is kind of random.

This is not really load-balancing from client point of view, but can still balance out network usage somehow if multiple filebeat instances are connected to one cluster.

Better option might be to set `loadbalance: false`, though. In this mode filebeat chooses on logstash only by random + there is a chance different filebeat instances choosing different logstash instances in a logstash cluster still somehow balancing some load.

filebeat/docs/load-balancing.asciidoc
```diff
@@ -0,0 +1,69 @@
...
+another. This mode requires the least memory and CPU usage.
+
+* **Send events to `N` hosts in lock-step:**
++
+You can configure Filebeat to send events to `N` hosts in lock-step by setting
+`spool_size = N * bulk_max_size`. This mode requires more memory and CPU usage
+than the previous mode.
++
```
urso: third mode:

set `filebeat.publish_async: true` and filebeat will try to push batches and wait for ACKs asynchronously. In this mode, filebeat will try to fully load-balance batches collected by spooler. LS instances being slower might handle less batches. This mode should get the highest throughput at the cost of highest memory usage: Batches are kept sorted in memory until ACKed by output. ACK might be out of order, but cleanup is done in order, potentially keeping ACKed batches in memory until older still active batches are ACKed.

filebeat/docs/load-balancing.asciidoc
```diff
@@ -0,0 +1,69 @@
...
+filebeat.prospectors:
+- input_type: log
+  paths:
+    - /var/log/*.log
+filebeat.spool_size: 4096
+output.logstash:
+  hosts: ["localhost:5044", "localhost:5045"]
+  loadbalance: true
+  bulk_max_size: 2048
+-------------------------------------------------------------------------------
```
urso: the loadbalancer supports multiple workers per host. By default `worker: 1`, but if number of workers is increased, more network connections will be in use.

e.g. one can configure on logstash host with `worker: 2` : 

```
output.logstash:
  hosts: ["localhost:5044"]
  loadbalance: true
  worker: 2
  bulk_max_size: 2048
```

filebeat/docs/load-balancing.asciidoc
```diff
@@ -0,0 +1,69 @@
...
+
+
+//REVIEWERS: Please confirm the accuracy of the config examples here. I'm guessing at the config and haven't tested these examples. It would probably be better to use a more realistic path for /var/log/*.log (please suggest one). I'd like to show the full config because it helps users understand where the settings fit into the overall yaml file.
```
urso: `/var/log/*.log` is even our default and should be fine on linux.

filebeat/docs/load-balancing.asciidoc
```diff
@@ -0,0 +1,69 @@
...
+
+//REVIEWERS: I'm a bit confused in the above example because bulk_max_size does not appear under output.logstash in filebeat.full.yml. Is this intentional? It does show up as an option in https://www.elastic.co/guide/en/beats/filebeat/5.0/logstash-output.html, so I assume it's valid, correct?
```
urso: `bulk_max_size` should be added to libbeat/_meta/config*yml . There is another ticket for checking for potentially missing output settings. Only docs have been synced most recently.

filebeat/docs/load-balancing.asciidoc
```diff
@@ -0,0 +1,69 @@
...
++
+Example:
++
+[source,yaml]
+-------------------------------------------------------------------------------
+filebeat.prospectors:
+- input_type: log
+  paths:
+    - /var/log/*.log
+filebeat.publish_async: false
```
urso: `s/false/true`

Oh, this is the third mode, hehe.

filebeat/docs/load-balancing.asciidoc
dedemorton: @urso I'm not sure what you want me to change or add here. I don't think I need to mention `publish_async` until later because the default is `false`. Also, I'm not sure if it makes sense to recommend changing `spool_size` until later in this topic.  I've changed the intro to say, "Filebeat can send events in a few different modes" since it sounds like the first approach is not truly load balancing (even though the `loadbalance` option is enabled). I'll push my changes in a bit, and maybe you can tell me specifically what I need to add or change to make the section right.

filebeat/docs/load-balancing.asciidoc
dedemorton: @urso I'm not sure what you want me to change (vs what you are offering here as explanation). My inclindation is to not talk about how we manage batches in memory because that's an internal implementation detail that users don't really want to know (and we might want to change someday). Best not to document unless users can configure it in someway.

dedemorton:
@urso I forgot to remove the in progress label for the PR. Is this one good to go?

filebeat/docs/load-balancing.asciidoc
urso: Yes, agree it should go in the first section about enabling `loadbalance`. The total number of workers participating in load balancing is `#hosts * worker`.

filebeat/docs/load-balancing.asciidoc
urso: I got a little confused with asciidoc a little. I did think the third mode to be a separate section and didn't see it initially. My bad.

filebeat/docs/load-balancing.asciidoc
```diff
@@ -0,0 +1,85 @@
...
+The `loadbalance` option is available for Redis, Logstash, and Elasticsearch
+outputs. The Kafka output handles load balancing internally.
+
+Filebeat can send events in a few different modes:
+
+* **Send events to one host after another:**
++
+By default, when you configure Filebeat to send events to multiple hosts
+(`loadbalance: true`), Filebeat will send the events to one host after
+another. This mode requires the least memory and CPU usage.
```
urso: Maybe add a note about this not really being load balancing?

PR #2792, opened 2016-10-17
Ignore duplicate state entries

### LINKED ISSUE #2784: Re-harvesting of files and duplicate entries in filebeat registry after migration to 5.0
While migrating from filebeat 1.2.2 to 5.0.0-beta1, we have the issue that some of the files are reharvested (seems like the state is lost) and they appear duplicated in the migrated registry file, the only difference being "ttl" - both "0" and "-1e+09".

Original registry file snippet:
```
"/var/log/cronolog/celery.20161014.log": {
    "FileStateOS": {
      "device": 64512,
      "inode": 6422570
    },
    "offset": 357081,
    "source": "/var/log/cronolog/celery.20161014.log"
  },
  "/var/log/cronolog/celerybeat.20161014.log": {
    "FileStateOS": {
      "device": 64512,
      "inode": 6422569
    },
    "offset": 120,
    "source": "/var/log/cronolog/celerybeat.20161014.log"
  },
```

Registry file after migration snippet:
```
  {
    "ttl": -1e+09,
    "timestamp": "2016-10-14T10:56:52.966694574+02:00",
    "FileStateOS": {
      "device": 64512,
      "inode": 6422569
    },
    "offset": 120,
    "source": "/var/log/cronolog/celerybeat.20161014.log"
  },
  {
    "ttl": 0,
    "timestamp": "2016-10-14T10:51:37.932376943+02:00",
    "FileStateOS": {
      "device": 64512,
      "inode": 6422569
    },
    "offset": 9527,
    "source": "/var/log/cronolog/celerybeat.20161014.log"
  },
  {
    "ttl": -1e+09,
    "timestamp": "2016-10-14T10:56:47.966070587+02:00",
    "FileStateOS": {
      "device": 64512,
      "inode": 6422570
    },
    "offset": 357081,
    "source": "/var/log/cronolog/celery.20161014.log"
  },
  {
    "ttl": 0,
    "timestamp": "2016-10-14T10:51:37.932376774+02:00",
    "FileStateOS": {
      "device": 64512,
      "inode": 6422570
    },
    "offset": 357081,
    "source": "/var/log/cronolog/celery.20161014.log"
  },
```

We're also seeing a lot of:
```
2016-10-14T11:21:33+02:00 ERR State for /var/log/cronolog/celery.20161014.log should have been dropped, but couldn't as state is not finished.
2016-10-14T11:21:33+02:00 ERR State for /var/log/cronolog/celerybeat.20161014.log should have been dropped, but couldn't as state is not finished.
``` 
in the filebeat logs.

The beginning of the log file when starting 5.0 for the first time is:
```
2016-10-14T10:51:37+02:00 INFO Metrics logging every 30s
2016-10-14T10:51:37+02:00 INFO Setup Beat: filebeat; Version: 5.0.0-beta1
2016-10-14T10:51:37+02:00 INFO Max Retries set to: 3
2016-10-14T10:51:37+02:00 INFO Activated logstash as output plugin.
2016-10-14T10:51:37+02:00 INFO Publisher name: ams1
2016-10-14T10:51:37+02:00 INFO Flush Interval set to: 1s
2016-10-14T10:51:37+02:00 INFO Max Bulk Size set to: 2048
2016-10-14T10:51:37+02:00 INFO filebeat start running.
2016-10-14T10:51:37+02:00 INFO Registry file set to: /var/lib/filebeat/registry
2016-10-14T10:51:37+02:00 INFO Loading registrar data from /var/lib/filebeat/registry
2016-10-14T10:51:37+02:00 INFO Old registry states found: 14
2016-10-14T10:51:37+02:00 INFO Old states converted to new states and written to registrar: 14
2016-10-14T10:51:37+02:00 INFO Loading Prospectors: 2
2016-10-14T10:51:37+02:00 INFO Starting Registrar
2016-10-14T10:51:37+02:00 INFO Load previous states from registry into memory
2016-10-14T10:51:37+02:00 INFO Start sending events to output
2016-10-14T10:51:37+02:00 INFO Previous states loaded: 14
2016-10-14T10:51:37+02:00 INFO Starting spooler: spool_size: 2048; idle_timeout: 5s
2016-10-14T10:51:37+02:00 INFO Load previous states from registry into memory
2016-10-14T10:51:37+02:00 INFO Previous states loaded: 14
2016-10-14T10:51:37+02:00 INFO Loading Prospectors completed. Number of prospectors: 2
2016-10-14T10:51:37+02:00 INFO All prospectors are initialised and running with 14 states to persist
2016-10-14T10:51:37+02:00 INFO Starting prospector of type: log
2016-10-14T10:51:37+02:00 INFO Starting prospector of type: log
...
2016-10-14T10:51:37+02:00 INFO Harvester started for file: /var/log/cronolog/celery.20161014.log
...
2016-10-14T10:51:38+02:00 ERR State for /var/log/cronolog/celerybeat.20161014.log should have been dropped, but couldn't as state is not finished.
...
2016-10-14T10:51:38+02:00 ERR State for /var/log/cronolog/celery.20161014.log should have been dropped, but couldn't as state is not finished.
...
2016-10-14T10:51:38+02:00 ERR State for /var/log/cronolog/celerybeat.20161014.log should have been dropped, but couldn't as state is not finished.
...
2016-10-14T10:51:43+02:00 ERR State for /var/log/cronolog/celerybeat.20161014.log should have been dropped, but couldn't as state is not finished.
...
2016-10-14T10:51:47+02:00 INFO Harvester started for file: /var/log/cronolog/celerybeat.20161014.log
2016-10-14T10:51:48+02:00 ERR State for /var/log/cronolog/celery.20161014.log should have been dropped, but couldn't as state is not finished.
...
2016-10-14T10:51:48+02:00 ERR State for /var/log/cronolog/celerybeat.20161014.log should have been dropped, but couldn't as state is not finished.
```

tsg:
When upgrading the registry file from 1.x versions, make sure the new state doesn't contain any duplicate inodes, as these will cause an invalid state. Fixes #2784.

### DIFF
--- a/filebeat/registrar/registrar.go
+++ b/filebeat/registrar/registrar.go
@@ -156,16 +156,8 @@ func (r *Registrar) loadAndConvertOldState(f *os.File) bool {
 	}
 
 	// Convert old states to new states
-	states := make([]file.State, len(oldStates))
 	logp.Info("Old registry states found: %v", len(oldStates))
-	counter := 0
-	for _, state := range oldStates {
-		// Makes timestamp time of migration, as this is the best guess
-		state.Timestamp = time.Now()
-		states[counter] = state
-		counter++
-	}
-
+	states := convertOldStates(oldStates)
 	r.states.SetStates(states)
 
 	// Rewrite registry in new format
@@ -174,6 +166,32 @@ func (r *Registrar) loadAndConvertOldState(f *os.File) bool {
 	logp.Info("Old states converted to new states and written to registrar: %v", len(oldStates))
 
 	return true
+}
+
+func convertOldStates(oldStates map[string]file.State) []file.State {
+	// Convert old states to new states
+	states := []file.State{}
+	for _, state := range oldStates {
+		// Makes timestamp time of migration, as this is the best guess
+		state.Timestamp = time.Now()
+
+		// Check for duplicates
+		dupe := false
+		for i, other := range states {
+			if state.FileStateOS.IsSame(other.FileStateOS) {
+				dupe = true
+				if state.Offset > other.Offset {
+					// replace other
+					states[i] = state
+					break
+				}
+			}
+		}
+		if !dupe {
+			states = append(states, state)
+		}
+	}
+	return states
 }
 
 func (r *Registrar) Start() error {
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -44,6 +44,7 @@ https://github.com/elastic/beats/compare/v5.0.0-rc1...master[Check the HEAD diff
 
 *Filebeat*
 - Fix issue when clean_removed and clean_inactive were used together that states were not directly removed from the registry.
+- Fix issue where upgrading a 1.x registry file resulted in duplicate state entries. {pull}2792[2792]
 
 *Winlogbeat*
 
--- a/filebeat/registrar/registrar_test.go
+++ b/filebeat/registrar/registrar_test.go
@@ -1,0 +1,59 @@
+// +build !windows,!integration
+
+package registrar
+
+import (
+	"sort"
+	"testing"
+
+	"github.com/elastic/beats/filebeat/input/file"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestConvertOldStates(t *testing.T) {
+	type io struct {
+		Name   string
+		Input  map[string]file.State
+		Output []string
+	}
+	tests := []io{
+		{
+			Name: "Simple test with three files",
+			Input: map[string]file.State{
+				"test":  file.State{Source: "test", FileStateOS: file.StateOS{Inode: 5}},
+				"test1": file.State{Source: "test1", FileStateOS: file.StateOS{Inode: 3}},
+				"test2": file.State{Source: "test2", FileStateOS: file.StateOS{Inode: 2}},
+			},
+			Output: []string{"test", "test1", "test2"},
+		},
+		{
+			Name: "De-duplicate inodes. Bigger offset wins (1)",
+			Input: map[string]file.State{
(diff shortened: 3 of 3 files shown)


filebeat/registrar/registrar.go
```diff
@@ -176,6 +168,32 @@ func (r *Registrar) loadAndConvertOldState(f *os.File) bool {
 	return true
 }
 
+func convertOldStates(oldStates map[string]file.State) []file.State {
```
andrewkroh: The logic you have looks correct. What do you think about using a map with struct key to do the deduplication?

```
	states := map[file.StateOS]file.State{}
	for _, state := range oldStates {
		// If there is a duplicate, use the one with the larger offset.
		existingState, found := states[state.FileStateOS]
		if found && existingState.Offset > state.Offset {
			continue
		}

		// Makes timestamp the time of migration, as this is the best guess.
		state.Timestamp = time.Now()
		states[state.FileStateOS] = state
	}
```

tsg: Hmm, that would simpler and would work, but bypassing the `IsSame()` method feels risky for future `StateOS` implementations (note that the structures is platform specific, so we could see variations).

I'm not really worried about the O(n^2) aspect here, since this is done only once and with an N up to tens of thousands at most.

tsg:
jenkins, retest it

filebeat/registrar/registrar.go
andrewkroh: Good point, that might be brittle and break if there are any changes to `StateOS`.

ruflin:
LGTM. Thx for the fix. I also don't worry here about the complexity here because it only happens once. And using IsSame should be "future proof".

Issue #1671, opened 2016-05-19
Document how to configure ingest node to have geoip info


monicasarbu:
As geoip is deprecated starting with alpha2, we need to provide a guide about how to configure Ingest Node to export geoip information that can be used to populate the "client location" map from the Packetbeat dashboard.

cc-ed: @tsg @dedemorton

monicasarbu:
Closing it as the PR was merged.

PR #2793, opened 2016-10-18
Document how to configure ingest node to have geoip

dedemorton:
@monicasarbu Here's my first draft. I probably need to give this another good read through, but I wanted to get it out to you for a first review. I've added a step to define mappings in the index template to get the `location` field to work in the visualization (without the mappings, the latitude and longitude get indexed as floats). (I haven't really worked much with mappings, so I hope it's OK.) If there is someway to take the latitude and longitude and index it directly into the client_location field (which is already a geo_point) then that would be better, but I'm not sure how to do that using the options that are available in the ingest geoIP processor plugin.


monicasarbu:
I have created a PR https://github.com/elastic/beats/pull/2795 to make the integration smooth between Beats and the Ingest GeoIP Processor Plugin. Here are the steps that the user should follow in case, he/she needs to add GeoIP information for the `client_ip`:
1. Install Ingest GeoIP Processor Plugin: https://www.elastic.co/guide/en/elasticsearch/plugins/master/ingest-geoip.html
2. Define a GeoIP processor in a pipeline by using Console from Kibana:
```
PUT _ingest/pipeline/geoip-info
{
  "description" : "Add geoip info",
  "processors" : [
    {
      "geoip" : {
        "field" : "client_ip",
        "target_field": "client_geoip",
        "properties": ["location"],
        "ignore_failure" : true
      }
    }
  ]
}
```
This will generate `client_geoip.location` of type geo_point.
3. Start Packetbeat (feature available only in 5.0.0-GA)
4. Import Kibana dashboard for Packetbeat or create a Tile map using `client_geoip.location` as Geohash.

Things that needs to be done:
- [ ] Mark `client_location` as DEPRECATED in favor of `client_geoip`. I couldn't find this in the documentation, but I remember that we decided to deprecate this a long time ago.

packetbeat/docs/packetbeat-geoip.asciidoc
```diff
@@ -0,0 +1,165 @@
...
+
+Prior to version 5.0, Packetbeat provided a `geoip` configuration option for
+exporting geoIP information about the source IPs.  Starting with 5.0, the
+`geoip` configuration option in Beats is deprecated in favor of using the
+ingest geoIP processor plugin. This plugin adds information about the
+geographical location of IP addresses, based on data from the Maxmind GeoLite2
+City Database. Using this plugin provides you with greater flexibility and
+control over the information that's returned by the geoIP lookup. 
+
+//REVIEWERS: trying to put some words around why the plugin is better, but I'm not sure if this is completely true. ^^
```
monicasarbu: The benefit is that you don't need to install the geoip database on all the Beats, and you can install it only on your Elasticsearch.

packetbeat/docs/packetbeat-geoip.asciidoc
```diff
@@ -0,0 +1,165 @@
...
+}
+-------------------------------------------------------------------------------
++
+Where `client_ip` is the output field in Packetbeat that contains the IP
+address of the client. You set `ignore_failure` to `true` so that the pipeline
+does not fail when it processes events that don't have a `client_ip` field. See 
+{plugindoc}/using-ingest-geoip.html[Using the Geoip Processor in a Pipeline] for
+more options. 
+
+3. Add the pipeline to Elasticsearch. For example:
```
monicasarbu: to Ingest Node instead of ES

packetbeat/docs/packetbeat-geoip.asciidoc
```diff
@@ -0,0 +1,165 @@
...
++
+[source,shell]
+-------------------------------------------------------------------------------
+curl -XPUT 'http://localhost:9200/_ingest/pipeline/geoip-info' -d@geoip-pipeline.json
+-------------------------------------------------------------------------------
```
monicasarbu: or another easier solution would be to use Console from Kibana instead of writing it to a file and then loading it usign curl.

packetbeat/docs/packetbeat-geoip.asciidoc
```diff
@@ -0,0 +1,165 @@
...
+`output.elasticsearch`. For example: 
++
+[source,yaml]
+-------------------------------------------------------------------------------
+output.elasticsearch:
+  hosts: ["localhost:9200"]
+  pipeline: geoip-info
+-------------------------------------------------------------------------------
+
+5. Before you start Packetbeat, edit the Packetbeat index template,
```
monicasarbu: this step is not needed, as I changed the template to accommodate `client_geoip.location`.

monicasarbu:
@dedemorton The documentation looks good to me. Please update it with the latest changes I wrote above.

packetbeat/docs/packetbeat-geoip.asciidoc
dedemorton: I prefer your streamlined approach. I was trying to be consistent with how we describe using ingest node elsewhere, but I think it's OK for users to see a couple of different ways to define pipelines and to know that the syntax is a bit easier if you use the console in Kibana.

packetbeat/docs/packetbeat-geoip.asciidoc
dedemorton: Awesome! That will be so much easier for the users.

dedemorton:
@monicasarbu I've resolved the comments from the review. I decided that I wouldn't cover the Kibana steps in too much detail, but I kept info about refreshing the index field list because I ran into this problem when I tested the steps.

dedemorton:
This PR resolves https://github.com/elastic/beats/issues/1671, which is also tracked in https://github.com/elastic/beats/issues/2482

PR #2796, opened 2016-10-18
Document PFRING as unsupported

tsg:
We don't regularly test it, so it cannot be supported. Modified the docs
to clarify that and direct people more towards the af_packet option.


monicasarbu:
LGTM

PR #2798, opened 2016-10-18
Update sarama to master + SASL fix

urso:
- based on v1.10.1
- master:
  - fixes bug in in v0.10 protocol if compression is enabled (default)
  - introduces broker metrics
- SASL/plain included
- Add support kafka 0.10.0.1 version string to valid libbeat kafka settings
- Hook up sarama kafka metrics (go-metrics introduces endpoint debug/metrics)

PR #2804, opened 2016-10-18
Update sarama to master + SASL fix (#2798)

urso:
* Update sarama to master + SASL fix

- based on v1.10.1
- master:
  - fixes bug in in v0.10 protocol if compression is enabled (default)
  - introduces broker metrics
- SASL/plain included


tsg:
LGTM, waiting for green.

Issue #2794, opened 2016-10-18
Data race in the add_cloud_metadata processor


tsg:
Seen this on a Jenkins run:

```
command [go test -race -cover -coverprofile /tmp/gotestcover-274791306 github.com/elastic/beats/libbeat/processors/actions]: exit status 1
==================
WARNING: DATA RACE
Write at 0x000000a1dbb0 by goroutine 22:
  github.com/elastic/beats/libbeat/processors/actions.TestRetrieveDigitalOceanMetadata()
      /home/jenkins/workspace/Beats-PR/src/github.com/elastic/beats/libbeat/processors/actions/add_cloud_metadata_test.go:240 +0x1d2
  testing.tRunner()
      /home/jenkins/.gimme/versions/go1.7.1.linux.amd64/src/testing/testing.go:610 +0xc9

Previous read at 0x000000a1dbb0 by goroutine 11:
  github.com/elastic/beats/libbeat/processors/actions.fetchMetadata.func4()
      github.com/elastic/beats/libbeat/processors/actions/_test/_obj_test/add_cloud_metadata.go:222 +0x77

Goroutine 22 (running) created at:
  testing.(*T).Run()
      /home/jenkins/.gimme/versions/go1.7.1.linux.amd64/src/testing/testing.go:646 +0x52f
  testing.RunTests.func1()
      /home/jenkins/.gimme/versions/go1.7.1.linux.amd64/src/testing/testing.go:793 +0xb9
  testing.tRunner()
      /home/jenkins/.gimme/versions/go1.7.1.linux.amd64/src/testing/testing.go:610 +0xc9
  testing.RunTests()
      /home/jenkins/.gimme/versions/go1.7.1.linux.amd64/src/testing/testing.go:799 +0x4ba
  testing.(*M).Run()
      /home/jenkins/.gimme/versions/go1.7.1.linux.amd64/src/testing/testing.go:743 +0x12f
  main.main()
      github.com/elastic/beats/libbeat/processors/actions/_test/_testmain.go:112 +0x2f8

Goroutine 11 (running) created at:
  github.com/elastic/beats/libbeat/processors/actions.fetchMetadata()
      github.com/elastic/beats/libbeat/processors/actions/_test/_obj_test/add_cloud_metadata.go:223 +0x5db
  github.com/elastic/beats/libbeat/processors/actions.newCloudMetadata()
[...]

PR #2805, opened 2016-10-19
Replace filters with processors

monicasarbu:
Replace `filters` with `processors` in the documentation.


tsg:
I guess the tests fail because of missing `make collect && make update`.

monicasarbu:
@tsg Yes, it looks like I changed the wrong file. It should work now.

tsg:
Merging, the new jenkins failure seems to be due to #2794.

PR #2808, opened 2016-10-19
Use RSS memory in the processes dashboards

tsg:
Resident was used in the tables for the percentages but not for the
value, where the virtual memory was used.

Here is a screenshot with the result:

screen shot 2016-10-19 at 17 39 33


monicasarbu:
LGTM

Issue #1823, opened 2016-06-08
packetbeat unknown link type (12) for device any on OS X


urso:
device type `any` is only supported on Linux. Create readable error message if device type is not supported. If `any` is used, packetbeat might create an error reporting unknown link type (12).

lindsay-show:
thanks,I have found out it

PR #2809, opened 2016-10-19
Improve error message when user sets any on Windows and OS X

monicasarbu:
Trying to fix https://github.com/elastic/beats/issues/1823
Check if the interface is `any`, and throw an error when the system is OS X or Windows.


urso:
this PR seems to make the packetbeat testsuite fail on OS X, as the testsuite it's config file uses `any` it seems, but the sniffer type is overwritten from command line.

monicasarbu:
@urso Fixed.

tsg:
LGTM.

PR #2810, opened 2016-10-19
Add Kafka config example

dedemorton:
I've added an example provided by @urso


dedemorton:
Adds kafka example for issue tracked in #2482

PR #1120, opened 2016-03-07
Add support for include_fields and drop_fields

monicasarbu:
This PR adds two configuration options: `include_fields` and `drop_fields`. 

There is a list of mandatory fields like: `@timestamp`, `beat`, `count`, `status` that cannot be removed and they are always available in the exported fields. With the two configuration options you can configure the additional fields that will be exported or not exported. By default, all fields are exported.

Here are the options:

- `include_fields=[]` -> exports only the mandatory fields
- `include_fields is not defined` (commented) -> export all fields (the default option)
- `drop_fields=[]` -> export all fields (the default option)
 
As fields, they accept a map like `proc` or a value like `proc.cpu.total_p`. 

In case there is a field to include that is not available in the event, then just ignore the field. 

When multiple filtering rules are defined, we start with a copy of the event and apply the filtering rules one by to the
event.

<pre>
   event -> filter rule 1 -> event 1 -> filter rule 2 -> event 2 ...
</pre>

   where event is the initial event, event1 is the event resulted after applying "filter rule1" and it's considered input for the "filter rule 2" and so on.

PR #2813, opened 2016-10-20
Combined 5.0.0 release notes

monicasarbu:
This PR contains the first draft of the combined 5.0.0 release notes that includes the changes done starting with 5.0.0-alpha1 until 5.0.0-GA.

Here the result of this PR:
screen shot 2016-10-21 at 1 14 58 pm

### DIFF
--- a/libbeat/docs/release-notes/5.0.0.asciidoc
+++ b/libbeat/docs/release-notes/5.0.0.asciidoc
@@ -139,7 +139,7 @@ The list below covers changes between 1.x to 5.0.0 releases.
 - Update to Go 1.7. {pull}2306[2306]
 - Add option to Elasticsearch output to pass http parameters in index operations {issue}805[805]
 - Improve Logstash and Elasticsearch backoff behavior. {pull}927[927]
-- Add experimental Kafka output. {pull}942[942] {pull}2188[2188] {pull}2190[2190] {pull}2284[2284]
+- Add Kafka output. {pull}942[942] {pull}2188[2188] {pull}2190[2190] {pull}2284[2284]
 - Add config file option to configure GOMAXPROCS. {pull}969[969]
 - Improve shutdown handling in libbeat. {pull}1075[1075]
 - Log total non-zero internal metrics on shutdown. {pull}2349[2349]
--- a/libbeat/docs/release-notes/5.0.0.asciidoc
+++ b/libbeat/docs/release-notes/5.0.0.asciidoc
@@ -1,0 +1,240 @@
+[[release-notes-5.0.0]]
+=== Beats version 5.0.0
+
+The list below covers changes between 1.x to 5.0.0 releases.
+
+
+==== Breaking changes
+
+*Affecting all Beats*
+
+- Rename the `filters` section to `processors`. {pull}1944[1944]
+- Require braces for environment variable expansion in config files {pull}1304[1304]
+- On DEB/RPM installations, the binary files are now found under `/usr/share/{{beat_name}}/bin`, not in `/usr/bin`. {pull}1385[1385]
+- The logs are written by default to self rotating files, instead of syslog. {pull}1371[1371]
+- Remove deprecated `host` option from elasticsearch, logstash and redis outputs. {pull}1474[1474]
+- All configuration settings under `shipper:` are moved to be top level configuration settings. I.e.
+  `shipper.name:` becomes `name:` in the configuration file. {pull}1570[1570]
+- The `topology_expire` option of the Elasticserach output was removed. {pull}1907[1907]
+- The Elasticsearch template is now loaded by default. {pull}1993[1993]
+- The Redis output `index` setting is renamed to `key`. `index` still works but it's deprecated. {pull}2077[2077]
+- The undocumented file output `index` setting was removed. Use `filename` instead. {pull}2077[2077]
+- Change Elasticsearch output index configuration to be based on format strings. If index has been configured, no date will be appended anymore to the index name. {pull}2119[2119]
+- If the path specified by the `-c` flag is not absolute and `-path.config` is not specified, it
+  is considered relative to the current working directory. {pull}2245[2245]
+- Rename `tls` configurations section to `ssl`. {pull}2330[2330]
+- Rename `certificate_key` configuration to `key`. {pull}2330[2330]
+- Replace `tls.insecure` with `ssl.verification_mode` setting. {pull}2330[2330]
+- Replace `tls.min/max_version` with `ssl.supported_protocols` setting requiring full protocol name. {pull}2330[2330]
+- A dynamic mapping rule is added to the default Elasticsearch template to treat strings as keywords by default. {pull}2688[2688]
+
+
+*Filebeat*
+
+- Scalar values in used in the `fields` configuration setting are no longer automatically converted to strings. {pull}1092[1092]
+- Count field was removed from event as not used in filebeat {issue}778[778]
+- Default location for the registry file was changed to be `data/registry` from the binary directory,
+  rather than `.filebeat` in the current working directory. This affects installations for zip/tar.gz/source,
+  the location for DEB and RPM packages stays the same. {pull}1373[1373]
+- The state for files which fall under ignore_older is not stored anymore. This has the consequence, that if a file which fell under ignore_older is updated, the whole file will be crawled.
+- Ignore symlinks by default, but they can be enabled with `symlinks` config {pull}1686[1686] {pull}2478[2478]
+- Set `close_inactive` default to 5 minutes (was 1 hour before)
+- Set `clean_removed` and `close_removed` to true by default
+
+
+*Packetbeat*
+
+- Rename output fields in the dns package. Former flag `recursion_allowed` becomes `recursion_available`. {pull}803[803]
+  Former SOA field `ttl` becomes `minimum`. {pull}803[803]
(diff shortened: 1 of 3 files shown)


libbeat/docs/release-notes/5.0.0.asciidoc
```diff
@@ -0,0 +1,268 @@
...
+
+The list below covers changes between 5.0.0-alpha1 to 5.0.0-GA.
+
+
+==== Breaking changes
+
+*Affecting all Beats*
+
+- Run function to start a Beat now returns an error instead of directly exiting. {pull}771[771]
+- The method signature of `HandleFlags()` was changed to allow returning an error {pull}1249[1249]
```
tsg: I'd leave these two (which are affecting only Beats writers) out, since a Beats user wouldn't know how this change affects them.

libbeat/docs/release-notes/5.0.0.asciidoc
```diff
@@ -0,0 +1,268 @@
...
+- Require braces for environment variable expansion in config files {pull}1304[1304]
+- On DEB/RPM installations, the binary files are now found under `/usr/share/{{beat_name}}/bin`, not in `/usr/bin`. {pull}1385[1385]
+- The logs are written by default to self rotating files, instead of syslog. {pull}1371[1371]
+- Remove deprecated `host` option from elasticsearch, logstash and redis outputs. {pull}1474[1474]
+- All configuration settings under `shipper:` are moved to be top level configuration settings. I.e.
+  `shipper.name:` becomes `name:` in the configuration file. {pull}1570[1570]
+- The `topology_expire` option of the Elasticserach output was removed. {pull}1907[1907]
+- Rename the `filters` section to `processors`. {pull}1944[1944]
```
tsg: This one we don't need, because "filters" was never released.

libbeat/docs/release-notes/5.0.0.asciidoc
```diff
@@ -0,0 +1,268 @@
...
+- Introduce the condition with `when` in the processor configuration. {pull}1949[1949]
```
tsg: We don't need this one since the processors are new in 5.0

libbeat/docs/release-notes/5.0.0.asciidoc
```diff
@@ -0,0 +1,268 @@
...
+- The Elasticsearch template is now loaded by default. {pull}1993[1993]
+- The Redis output `index` setting is renamed to `key`. `index` still works but it's deprecated. {pull}2077[2077]
+- The undocumented file output `index` setting was removed. Use `filename` instead. {pull}2077[2077]
+- Change Elasticsearch output index configuration to be based on format strings. If index has been configured, no date will be appended anymore to the index name. {pull}2119[2119]
+- Replace `output.kafka.use_type` by `output.kafka.topic` accepting a format string. {pull}2188[2188]
```
tsg: We don't need this one since Kafka is new in 5.0

libbeat/docs/release-notes/5.0.0.asciidoc
```diff
@@ -0,0 +1,268 @@
...
+
+*Filebeat*
+
+- Scalar values in used in the `fields` configuration setting are no longer automatically converted to strings. {pull}1092[1092]
+- Count field was removed from event as not used in filebeat {issue}778[778]
+- Default location for the registry file was changed to be `data/registry` from the binary directory,
+  rather than `.filebeat` in the current working directory. This affects installations for zip/tar.gz/source,
+  the location for DEB and RPM packages stays the same. {pull}1373[1373]
+- The state for files which fall under ignore_older is not stored anymore. This has the consequence, that if a file which fell under ignore_older is updated, the whole file will be crawled.
+- Stop following symlink. Symlinks are now ignored: {pull}1686[1686]
```
tsg: We changed our mind on this one. Symlinks are just disabled by default, but can be enabled from config.

dedemorton:
I don't think users will know what we mean by "combined release notes" when they look at the TOC.

It would be better just to have an entry in the nav for Beats version 5.0.0 and provide the combined release notes there (all the release notes for the pre-releases consolidated and with no mention of pre-release versions). I would also suggest removing all the alpha/beta relnotes plus the relnotes for older releases. I'm not sure if that's possible since we've automated some of the work, but I think it would make more sense for 5.0 users to only see the GA-level release notes (consolidated).

libbeat/docs/release-notes/5.0.0.asciidoc
```diff
@@ -0,0 +1,263 @@
+[[release-notes-5.0.0]]
+=== 5.0.0 Combined Release Notes
+
+The list below covers changes between 1.3.1 to 5.0.0-GA.
```
dedemorton: If users want the list of changes for 1.3. I don't think these relnotes belong in the 5.0 docs.

libbeat/docs/release.asciidoc
```diff
@@ -30,3 +30,12 @@ This section summarizes the changes in each release.
 --
 
 include::../../CHANGELOG.asciidoc[]
+
+== Combined release notes
```
dedemorton: So to clarify my earlier comment, I think we need to have only the 5.0 release notes in this document. So the nav would look like:

Release Notes
   Beats version 5.0.0

Then when 5.0.1 comes out, we'll add another section. And if people want to see a more granular description of the changes, they can look at the changelog.

Maybe we keep to the automated format when we create pre-release software, but then we curate the list and create combined release notes at GA. Does that make sense?

tsg:
@dedemorton The release notes docs simply include the CHANGELOG.asciidoc. That's why we don't have too much flexibility, and we can't remove releases from the docs but keep them in the CHANGELOG file.

I agree the combined notes should be named simply `5.0.0` and be the place where we send most users. But I'm not sure we can remove the pre-releases from there.

libbeat/docs/release-notes/5.0.0.asciidoc
monicasarbu: I wrote 1.x to 5.0.0 releases.

libbeat/docs/release-notes/5.0.0.asciidoc
```diff
@@ -0,0 +1,262 @@
...
+- Fix sync publisher `PublishEvents` return value if client is closed concurrently. {pull}2046[2046]
+- Fix Logstash output handles error twice when asynchronous sending fails. {pull}2441[2441]
+- Fix Elasticsearch structured error response parsing error. {issue}2229[2229]
+- Fixed the run script to allow the overriding of the configuration file. {issue}2171[2171]
+- Fix logstash output crash if no hosts are configured. {issue}2325[2325]
+- Fix array value support in `-E` CLI flag. {pull}2521[2521]
+- Fix merging array values if `-c` CLI flag is used multiple times. {pull}2521[2521]
+- Fix beats failing to start due to invalid duplicate key error in configuration file. {pull}2521[2521]
+- Fix panic on non writable logging directory. {pull}2571[2571]
+- Make sure Beats sent always float values when they are defined as float by sending 5.00000 instead of 5. {pull}2627[2627]
```
tsg: Not sure if this one is really a bug. Fits more into Added, to me.

libbeat/docs/release-notes/5.0.0.asciidoc
monicasarbu: Generic filtering is part of Beats starting with 1.1.0: https://github.com/elastic/beats/pull/1120

monicasarbu:
@dedemorton I suggest to keep things simple for this release, and add only a link to the combined release notes on top of all the release notes. Please see the screenshot from the description of the pull request.

libbeat/docs/release-notes/5.0.0.asciidoc
```diff
@@ -0,0 +1,240 @@
...
+
+
+==== Added
+
+*Affecting all Beats*
+
+- Update to Go 1.7. {pull}2306[2306]
+- Add option to Elasticsearch output to pass http parameters in index operations {issue}805[805]
+- Improve Logstash and Elasticsearch backoff behavior. {pull}927[927]
+- Add experimental Kafka output. {pull}942[942] {pull}2188[2188] {pull}2190[2190] {pull}2284[2284]
```
tsg: @urso Is the Kafka output experimental at this point? We don't seem to mark it as such in the docs. I'd go with fully supported.

tsg: urso confirmed on chat that we can take down "experimental".

libbeat/docs/release-notes/5.0.0.asciidoc
```diff
@@ -0,0 +1,240 @@
...
+*Metricbeat*
+
+- First public release, containing the following modules: system, Apache, MySQL, PostgreSQL, Nginx, Redis, MongoDB, and Zookeeper.
+  Metricbeat is replacing Topbeat in 5.0, and its default behaviour is to export system statistics like CPU usage,
+  memory usage, Disk IO similar with what Topbeat 1.x is doing.
+- Add `pgid` field to process information. {pull} 2021[2021]
+- Use the new `scaled_float` Elasticsearch type for the percentage values. {pull}2156[2156]
+- Improve mapping by converting `half_float` to `scaled_float` and integers to long. {pull}2430[2430]
+- Add experimental `cgroup` metrics to the system/process MetricSet. {pull}2184[2184]
+- Add experimental `haproxy` module. {pull}2384[2384]
```
monicasarbu: @tsg I used a different line for haproxy as it's experimental for now.

Issue #2790, opened 2016-10-17
Testing Filebeats config as specified will fail 


nrichers:
This tip is in at least a couple of versions of the Filebeats docs (e.g. https://www.elastic.co/guide/en/beats/filebeat/5.0/filebeat-configuration.html, https://www.elastic.co/guide/en/beats/filebeat/1.3/filebeat-configuration.html):

![image](https://cloud.githubusercontent.com/assets/15148011/19423949/1ced6300-93da-11e6-9daa-e4da4025aee3.png)

There are some issues with this tip as specified, especially for new users who are setting up Filebeat for the first time. The command will almost certainly fail for several reasons:
- There is no indication  where to enter this command from. I ran this on a CentOS 7.1 box brought up in Vagrant and installed Filebeat using the RPM method from my home directory but you could be anywhere in the file path. For this command to work, I had to be in `/usr/share/filebeat/bin/` or you'd have to prefix the command with the path. But you have to find that path first. :-) 
- The Filebeat RPM install places `filebeat.yml` and templates into `/etc/filebeat/` but the config check doesn't seem to check for the files there, at least not on the 5.0 RC test installs I did. After trying to modify `filebeat.yml` for a bit and getting some of the templates read but not all, I ended up just copying the files over into `/usr/share/filebeat/bin/` which solved that part of the problem. It would be good to indicate in the install steps where to place the files or how to make Filebeat read the files from where they are placed. [EDIT: In one of the later steps, where to load the index template from is covered, but that covers only some of the files that were causing some issues during config testing and it's after the fact.]
- I think you need to run this command with sudo permissions to make it work:

```
$ ./filebeat -configtest -e
Exiting: error setting default paths: Failed to create data path /usr/share/filebeat/bin/data: mkdir /usr/share/filebeat/bin/data: permission denied

$ sudo  ./filebeat -configtest -e
2016/10/17 02:33:29.407251 beat.go:264: INFO Home path: [/usr/share/filebeat/bin] Config path: [/usr/share/filebeat/bin] Data path: [/usr/share/filebeat/bin/data] Logs path: [/usr/share/filebeat/bin/logs]
2016/10/17 02:33:29.407290 beat.go:174: INFO Setup Beat: filebeat; Version: 5.0.0-rc1
2016/10/17 02:33:29.407390 output.go:167: INFO Loading template enabled. Reading template file: /usr/share/filebeat/bin/filebeat.template.json
[...]

PR #2826, opened 2016-10-21
Clarify instructions for running configtest

dedemorton:
Fixes issue reported in #2790 (clarified description for running configtest).

Also added some conditional coding because we were showing commands for winlogbeat that weren't right (source text is in shared files, hence the need to add conditional coding).


tsg:
jenkins, retest it

PR #2827, opened 2016-10-24
Updated testing env to build candidate 4

tsg:


testing/environments/docker/kibana/Dockerfile-snapshot
```diff
@@ -2,7 +2,7 @@ FROM debian:jessie
 
 ENV VERSION 5.0.0
 ENV FILENAME_VERSION 5.0.0-linux-x86_64
-ENV URL http://staging.elastic.co/5.0.0-731e78df/downloads/kibana/kibana-${FILENAME_VERSION}.tar.gz
```
ruflin: Perhaps we could put this hash in the future in a `.env` file so it only has to be updated in one place: https://docs.docker.com/compose/env-file/ But not sure if this will work with our "nested" setup of multiple compose files.

ruflin:
LGTM

testing/environments/docker/kibana/Dockerfile-snapshot
tsg: Yeah, I didn't try it, but I can in a future PR. I imagine it should be somehow possible.

PR #2829, opened 2016-10-24
Update docs with the GA yum/apt repos

tsg:
Remote `prelease` from the YUM/APT repos, as the GA gets to repositories.

Open for reviews, but please **don't merge until the release day**.


tsg:
jenkins, test it

PR #2834, opened 2016-10-24
A better method to generate partition id.

fatelei:
RT

### DIFF
--- a/libbeat/outputs/kafka/partition.go
+++ b/libbeat/outputs/kafka/partition.go
@@ -153,12 +153,13 @@ func cfgRoundRobinPartitioner(config *common.Config) (func() partitioner, error)
 	return func() partitioner {
 		N := cfg.GroupEvents
 		count := N
-		partition := rand.Int31()
+		var partition int32
+		partition = 0
 
 		return func(_ *message, numPartitions int32) (int32, error) {
 			if N == count {
 				count = 0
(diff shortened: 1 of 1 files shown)


libbeat/outputs/kafka/partition.go
```diff
@@ -153,12 +153,13 @@ func cfgRoundRobinPartitioner(config *common.Config) (func() partitioner, error)
 	return func() partitioner {
 		N := cfg.GroupEvents
 		count := N
-		partition := rand.Int31()
+		var partition int32
+		partition = 0
```
urso: the `rand` has been on purpose, such that the first partition to publish too is chosen by random in order to have a chance to somewhat evenly distribute events if N is large and beats get's restarted.

libbeat/outputs/kafka/partition.go
```diff
@@ -153,12 +153,13 @@ func cfgRoundRobinPartitioner(config *common.Config) (func() partitioner, error)
...
 
 		return func(_ *message, numPartitions int32) (int32, error) {
 			if N == count {
 				count = 0
-				if partition++; partition >= numPartitions {
+				if partition++; partition%numPartitions == 0 {
```
urso: what's the advantage of module operation vs. a simple comarison of `>=`?

urso:
why doing these changes? What's the actual improvement over current solution?

PR #2845, opened 2016-10-25
Fix FPM docker image

tsg:
Looks like the most recent FPM needs ruby 2.0 (via the mustache dep)?
Upgraded the image to ubuntu 16.04 as a workaround.

### DIFF
--- a/dev-tools/packer/docker/fpm-image/Dockerfile
+++ b/dev-tools/packer/docker/fpm-image/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:14.04
+FROM ubuntu:16.04
 
 MAINTAINER Tudor Golubenco <tudor@elastic.co>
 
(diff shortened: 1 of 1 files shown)


tsg:
jenkins, package it

tsg:
@jordansissel The error we were getting on `gem install fpm` was:

```
ERROR:  Error installing fpm:
	mustache requires Ruby version ~> 2.0.
```

Does this fix (basically upgrading to ruby2) seem reasonable?

ruflin:
LGTM

jordansissel:
Yeah upgrading ruby should work. I need to fix fpm to not require ruby 2,
but +1 on upgrading ruby.

tsg:
jenkins, package it

tsg:
jenkins, package it

tsg:
Thanks @jordansissel!

Issue #1308, opened 2016-04-04
Update links to Shield documentation for 5.0


dedemorton:
Reminder to update doc links that currently point to Shield documentation because the links are changing for 5.0 GA. Opening this item to track the change because right now the final location is in flux.

dedemorton:
Fixed in #2850

PR #2850, opened 2016-10-26
Update links to x-pack security docs

dedemorton:
Note that we can't link to https://www.elastic.co/guide/en/shield/current/certificate-authority.html anymore because the topic has been removed from the doc.


dedemorton:
Fixes #1308

PR #2851, opened 2016-10-26
Add banner to indicate that doc about adding protocols is out-of-date

dedemorton:
@urso should probably confirm that we want to point to the protocol generator. Seems like a pointer is better than no guidance.


urso:
LGTM

Issue #2613, opened 2016-09-21
tail_files behavior


jjfalling:
Most users might expect that tail_files would  apply `ReadFromLast` only on the first scan and from then on it is not taken into account anymore. This behavior was not possible in 1.x because of filebeat worked but it should be possible to change it in the 5.x versions.

ruflin:
During startup of filbeat 5.x, a first scan is done to set all the states. During this scan `tail_files` should apply but not during all following scans. The problem in 5.x is that `tail_files` currently still applies on a harvester level as the offset is set inside the harvester. This logic could be moved to the prospector which makes it possible, to only apply it during the first scan.

ruflin:
@jjfalling One question that popped up during implementation: Should only the files be tailed for which no state exists yet or all of them? Means if a file was harvested before and there is already a state of an offset point to the middle of the file, should filebeat set the offset to the end in the first scan or because a state exist start sending the files from the middle of the file?

ruflin:
Checking the behaviour of nxlog:

> ReadFromLast
> This optional directive takes a boolean value. If it is set to TRUE, it instructs the module to only read logs which arrived after nxlog was started in case the saved position could not be read (for example on first start). When SavePos is TRUE and a previously saved position value could be read, the module will resume reading from this saved position. If this is FALSE, the module will read all logs from the file. This can result in quite a lot of messages which is usually not the expected behaviour. If this directive is not specified, it defaults to TRUE.

In summary, nxlog does not apply `tail_files` if a state already exists.

PR #2859, opened 2016-10-26
Add state for ignore_older files

ruflin:
Previously if a file was falling under ignore_older on startup and no previous state was found, also no state was persisted. This was important in 1.x releases to prevent the registry file to contain too many files as no cleanup was possible. With the new clean_\* options the registry file can now be cleaned up. For consistency all files that fall under ignore_older should have a state, even if they are not crawled before.

This change is also required to move tail_files under the prospector (https://github.com/elastic/beats/issues/2613). Otherwise files which fall under `ignore_older` the state could not be set to the end of the file.


ruflin:
jenkins, retest it

tsg:
LGTM.

PR #2868, opened 2016-10-27
Stop filebeat on error starting registrar

ruflin:
Currently when there is an error during loading the state, filebeat 5.0 doesn't stop and just continues. The registrar loading in filebeat 5.x was already improved an additional validation was added. This PR only changes that if there is an error in loading states, filebeat is not started and a proper error is returned.

This PR does not need forward porting as in the master branch already a more enhanced error management is in place.


tsg:
LGTM.

PR #2877, opened 2016-10-28
Fix gotestcover executable in make.bat for winlogbeat

maddin2016:
On windows if you call `make.bat` you get an error for tests that `gotestcover` i not a valid command. So you have define the path and then call it with `go run`. What i also noticed is that you have install many python extensions that are not installed by default.
```
pip install pyyaml
pip install jinja2
pip install pypiwin32
pip install nose
```
So maybe you should make a note in the readme for winlogbeat.

### DIFF
--- a/winlogbeat/make.bat
+++ b/winlogbeat/make.bat
@@ -14,7 +14,8 @@ if %errorlevel% neq 0 exit /b %errorlevel%
 
 echo Testing
 mkdir build\coverage
-gotestcover -race -coverprofile=build/coverage/integration.cov github.com/elastic/beats/winlogbeat/...
+set gotestcover="%GOPATH/src/github.com/elastic/beats/vendor/github.com/pierrre/gotestcover/gotestcover.go"
+go run %gotestcover% -race -coverprofile=build/coverage/integration.cov github.com/elastic/beats/winlogbeat/...
 if %errorlevel% neq 0 exit /b %errorlevel%
 
 echo System Testing


maddin2016:
Hi @karmi, CLA is signed 👍

ruflin:
@maddin2016 Thanks a lot of providing this fix. Indeed all of us don't develop on Windows so it can easily happen that the Windows make file lags behind :-( Let us know if you hit further issues.

andrewkroh:
`gotestcover` is already installed by make.bat on line 8. The changes in this PR should not be necessary.

@maddin2016 Do you have your GOPATH included in your PATH?

maddin2016:
Hi @andrewkroh, here is step by step the error
image

As you see, windows shell doesnt recognize these commands.

maddin2016:
So gotestcover has to be something like `gotestcover.exe` to recognize it as a command. But is `gotestcover.go` 
image

which has to be run from go. I don't know how such things works on linux/mac.

andrewkroh:
Can you `echo %PATH%`?

The `$GOPATH/bin` should be included in your `$PATH` so that when you do `go install xyz` the binary becomes available on in your PATH. https://golang.org/doc/code.html#GOPATH

On [line 8](https://github.com/elastic/beats/blob/master/winlogbeat/make.bat#L8) the script installs the gotestcover.exe to `$GOPATH/bin`.

maddin2016:
Ooooohhh 🙈 Sorry!! There were 2 points that i missed. First is that you create a `bin` folder under `GOPATH` and the second is that you you have put this bin folder into PATH. Now all works fine. So this PR is not necessary. I only found how to's like [this](http://www.wadewegner.com/2014/12/easy-go-programming-setup-for-windows/). So the others are mostly for linux. Sorry again!!

PR #2884, opened 2016-10-29
Metricbeat system module updates for Windows

### LINKED ISSUE #1704: Server 2003 - PID’s not recognized - Topbeat
- Version: TopBeat 1.2.3
- Operating System: Windows Server 2003
- Steps to Reproduce: Use the same input as configured in the topbeat.yml file
- Error: PID's not recognized in Server 2003, therefor nothing is sent to Logstash

Attachment:
[topbeatSettingsAndErrorlog.zip](https://github.com/elastic/beats/files/277603/topbeatSettingsAndErrorlog.zip)

Original thread:
https://discuss.elastic.co/t/server-2003-pids-not-recognized/50450

### LINKED ISSUE #1897: [TopBeat 1.2.3.] Can't access PID, even with correct rights.
Original thread here: https://discuss.elastic.co/t/server-2012-r2-some-metrics-not-send/51403
- Version: TopBeat 1.2.3
- Operating System: Windows Server 2012 R2 Standard (VMware host)
- Steps to Reproduce:
- Installed TopBeat via Powershell with Administrator Rights (right click, run as admin). 
- In processes, it is seen that TopBeat is owned by SYSTEM. Topbeat is also executed as SYSTEM.
- The error in topbeat.log with debug level:

2016-05-31T11:00:13+02:00 DBG Skip process pid=0: error getting process state for pid=0: OpenProcess fails with The parameter is incorrect.
2016-05-31T11:00:13+02:00 DBG Skip process pid=4: error getting process state for pid=4: OpenProcess fails with Access is denied.
2016-05-31T11:00:13+02:00 DBG Windows is interactive: false

For convience, the full image:
lpv5ave

### LINKED ISSUE #2860: Metricbeat system.process.ppid is 0 on Windows
On Windows, the Metricbeat `system.process.ppid` value is always `0`. See elastic/gosigar#53 for details on the cause.
- Version: Metricbeat 5.0.0
- Operating System: Windows
- Steps to Reproduce: Enable system process metricset.

### LINKED ISSUE #2885: Metricbeat diskio metricset doesn't work on Windows XP/2003
The diskio metricset always produces an error on Windows XP. Running the same query on command line works. This is the query gopsutil runs:

`PS > Get-WmiObject -query "SELECT * FROM Win32_PerfFormattedData_PerfDisk_LogicalDisk "`

- Version: master (compiled for GOOS=386)
- Operating System: Windows XP (2003 should be the same)
- Steps to Reproduce: Enable diskio metricset.

```
{
  "@timestamp": "2016-10-29T14:11:00.135Z",
  "beat": {
    "hostname": "newxp_2007",
    "name": "newxp_2007",
    "version": "6.0.0-alpha1"
  },
  "error": "disk io counters: Unknown name.",
  "metricset": {
    "module": "system",
    "name": "diskio",
    "rtt": 721036
  },
  "system": {
    "diskio": {}
  },
  "type": "metricsets"
}
```

andrewkroh:
- Added system core metricset for Windows. Per core metrics were implemented
  in elastic/gosigar.
- Added logging of process/system details on Windows to aid in debugging (user,
  arch, cores, sid, privs).
- Fixes #2860 (PPID is zero on Windows).
- Fixes #1704 (Server 2003 - PID’s not recognized). The command line arguments
  for the process will not be reported on XP and 2003.
- Fixes #1897 (OpenProcess access denied on Windows). Added code to enable the
  SeDebugPrivilege when it is available.
- Fixes #2885 (Metricbeat diskio metricset doesn't work on Windows XP/2003).
- Enable fsstats by default in Metricbeat config file.


andrewkroh:
I would like to have elastic/gosigar#58 and #2282 (Metricbeat AppVeyor) merged first. Then I will update this PR with those fixes.

ruflin:
@andrewkroh Both PR's were merged.

metricbeat/module/system/core/_meta/docs.asciidoc
```diff
@@ -8,3 +8,4 @@ This metricset is available on:
 - FreeBSD
 - Linux
 - OpenBSD
+- Windows
```
ruflin: 👍

metricbeat/etc/beat.short.yml
```diff
@@ -1,58 +0,0 @@
-########################## Metricbeat Configuration ###########################
```
ruflin: Is that change on purpose?

metricbeat/etc/beat.full.yml
```diff
@@ -17,10 +17,10 @@ metricbeat.modules:
     - cpu
 
     # System Load stats
-    - load
+    #- load
```
ruflin: Should we mention that we changed the defaults in the config file? Why is `load` disabled now by default?

metricbeat/module/system/_meta/config.full.yml
```diff
@@ -16,7 +16,7 @@
     - filesystem
 
     # File system summary stats
-    #- fsstat
+    - fsstat
```
tsg: Just making sure. We had this disabled before because it wasn't working on Windows and now it is?

metricbeat/module/system/_meta/config.full.yml
```diff
@@ -4,10 +4,10 @@
     - cpu
 
     # System Load stats
-    - load
+    #- load
```
tsg: I think we can leave this enabled. We have code to disable it automatically on WIndows: https://github.com/elastic/beats/blob/master/metricbeat/Makefile#L61

metricbeat/module/system/_meta/config.full.yml
```diff
@@ -4,10 +4,10 @@
...
 
     # Per CPU core stats
-    #- core
+    - core
```
tsg: Also not sure about this one, I think we had it disabled by default in an effort to keep disk usage down?

metricbeat/etc/beat.short.yml
andrewkroh: Yeah, this is on purpose, but I did want to ask you about it. I removed this file. Where is this file used?

metricbeat/module/system/_meta/config.full.yml
andrewkroh: Yeah, I'll revert this. For some reason I thought we removed all of the OS specific `before-build` changes.

metricbeat/module/system/_meta/config.full.yml
andrewkroh: I can revert this too. I was thinking that it was only commented out because of it not working on Windows.

metricbeat/module/system/_meta/config.full.yml
andrewkroh: It was working on Windows previously. So maybe it was commented out to reduce the amount of data Metricbeat sends? But the fsstats data is used on the Metricbeat overview dashboard which is why I think we should enabled it by default.

andrewkroh:
I updated the PR to contain the gosigar fixes and rebased it so that it gets built on AppVeyor. I reverted the changes to `load` and `cores` in the config file. I have left `fsstats` enabled by default.

Issue #2888, opened 2016-10-31
MongoDB password from connection string is indexed in metricset.host field and is visible


cwhenderson20:
Similar to #2738. When connecting to a running instance of MongoDB, the user and password of the connection string are indexed into the metricset.host field.

However, unlike #2738, I'm not aware of another way to specify the connection string such that the user/pass is separated out. 

This prevents us from using the mongodb module as even logging the credentials of a users with minimal permissions is not acceptable.

tsg:
Ah, good point :(. Only workaround for now is to use `drop_field` to remove the host information. Treating this as high prio.

ruflin:
Closing as https://github.com/elastic/beats/pull/2889 was merged.

PR #2889, opened 2016-10-31
Add explicit username and password option for Mongodb

tsg:
We previously only had the option of specifying a user/pass in the
URL string, which is problematic because it results in the password
being indexed in Elasticsearch (#2888).

This adds the option to specify a username/password at the module
configuration. To make this happen, I had to copy some unexported
functions from the mgo driver.

### DIFF
--- a/metricbeat/module/mongodb/_meta/config.yml
+++ b/metricbeat/module/mongodb/_meta/config.yml
@@ -4,6 +4,15 @@
   #period: 10s
 
   # The hosts must be passed as MongoDB URLs in the format:
-  # [mongodb://][user:pass@]host[:port]
+  # [mongodb://][user:pass@]host[:port].
+  # Warning: specifying the user/password in the hosts array is possible
+  # but it will result in the password being present in the output documents.
+  # We recommend using the username and password options instead.
   #hosts: ["localhost:27017"]
+
+  # Username to use when connecting to MongoDB. Empty by default.
+  #username: user
+
+  # Password to use when connecting to MongoDB. Empty by default.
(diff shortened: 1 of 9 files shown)


metricbeat/module/mongodb/_meta/config.yml
```diff
@@ -4,6 +4,15 @@
...
   # The hosts must be passed as MongoDB URLs in the format:
-  # [mongodb://][user:pass@]host[:port]
+  # [mongodb://][user:pass@]host[:port].
+  # Warning: specifying the user/password in the hosts array is possible
+  # but it will result in the password being present in the output documents.
+  # We recommend using the username and password options instead.
   #hosts: ["localhost:27017"]
 
+  # Username to use when connecting to MongoDB. Empty by default.
+  # username: user
```
ruflin: no space in front of `username` and `password` needed

metricbeat/module/mongodb/status/status.go
```diff
@@ -24,26 +24,36 @@ func init() {
...
 
 	if err := base.Module().UnpackConfig(&config); err != nil {
 		return nil, err
 	}
 
+	info, err := mongodb.ParseURL(base.Host(), config.Username, config.Password)
+	if err != nil {
+		return nil, err
+	}
+	info.Timeout = base.Module().Config().Timeout
```
ruflin: 👍

tsg:
Not strictly a bug fix, but I think we should backport this to the 5.0 branch as it makes it really difficult to use the MongoDB module otherwise.

PR #2890, opened 2016-10-31
Add username/pass options for PostgreSQL

tsg:
Similar to #2889 but for PostgreSQL. Also adds docs to the Postgres module,
which were missing, and adjusted the integration tests to use the username
option instead of the full URL.

### DIFF
--- a/metricbeat/etc/beat.full.yml
+++ b/metricbeat/etc/beat.full.yml
@@ -162,10 +162,20 @@ metricbeat.modules:
   #period: 10s
 
   # The host must be passed as PostgreSQL DSN. Example:
-  # postgres://pqgotest:password@localhost:5432?sslmode=disable
+  # postgres://localhost:5432?sslmode=disable
   # The available parameters are documented here:
   # https://godoc.org/github.com/lib/pq#hdr-Connection_String_Parameters
-  #hosts: ["postgres://postgres@localhost:5432"]
+  #
+  # Warning: specifying the user/password in the hosts array is possible
+  # but it will result in the password being present in the output documents.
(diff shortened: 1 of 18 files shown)


metricbeat/etc/beat.full.yml
```diff
@@ -153,10 +153,20 @@ metricbeat.modules:
   #period: 10s
 
   # The host must be passed as PostgreSQL DSN. Example:
-  # postgres://pqgotest:password@localhost:5432?sslmode=disable
+  # postgres://localhost:5432?sslmode=disable
   # The available parameters are documented here:
   # https://godoc.org/github.com/lib/pq#hdr-Connection_String_Parameters
-  #hosts: ["postgres://postgres@localhost:5432"]
+  #
+  # Warning: specifying the user/password in the hosts array is possible
```
ruflin: Not sure if I would add this to the beat.full.yml as not all parts which are in the docs also have to be in the config file. As long as we removed username / password from the example and have username / password below I don't think we need to add this here. It will make it only harder to keep it up-to-date in case of changes.

tsg: You missed it in the mongo PR then :-)

I worry that people will still add the user/pass to the connection string because that's how they are used from other software, and then be surprised to see we've indexed the password. Maybe I'm overreacting here, happy to also take it out. But I guess once we do the sanitization part we take out the warnings and we're done with it.

ruflin: Yep, I realised when I wrote that, that i missed it there but that was too late ;-)

Taking it out later SGTM.

ruflin:
LGTM

PR #2892, opened 2016-10-31
Update graphics and intro to beats platform

dedemorton:
Reworked beats platform image to align with new product images. Also added Winlogbeat. 

Had to remove mention of {Future}beat to save space. I could show {Future}beat instead of one of the other Beats, but I like showing all the Beats for now (we shouldn't try to show more than 4).

I deleted the image in the Logstash section of the getting started because I felt that it duplicated what we show in the platform overview without adding anything new. If we want to have an image there, we should show something interesting about how Beats and Logstash work together, but I feel that the current image doesn't add much. (I did actually rework beats-logstash.png, but deleted it when I saw that it was too much like the overview.) WDYT?


ruflin:
LGTM. I agree that on the overview we should not have {Future}Beat as this will only confuse people. It would be nice to have a similar diagram with {Your}Beat in the dev guide, but that is only nice to have.

PR #2894, opened 2016-10-31
Add note so users know to update mappings for geoip

dedemorton:
The index template in 5.1 will contain the correct mappings. Until then, users need to update the index template manually.


monicasarbu:
Thanks @dedemorton

monicasarbu:
@dedemorton  Starting with 5.0.1, we will have this fixed. Please have a look at https://github.com/elastic/beats/pull/2939

PR #2896, opened 2016-11-01
Remove LSF migration guide for 5.x

ruflin:
The LSF migration guide was inteded and tested for migration from lsf to 1.x versions. As there are breaking changes in filebeat 5.0 the migration does not necessarly work. As LSF reaches EOL on Nov 7th 2016 the migration guide is removed from upcoming versions. The migration guide will still exist in the 1.x docs.

Question: Should this be backported to 5.0 branch?


tsg:
Is there any particular information in it that is wrong and cannot be updated with reasonable effort? I think there still are plenty of LSF users, and I'd prefer them to upgrade to 5.x directly rather than hop through 1.x.

ruflin:
I don't think there is any content which cannot be updated. But some reasons to remove it from my perspective:

* Registry file structure changed. People must start with a new instance anyway.
* `tail_files` behaviour is up for change
* Config migration: I prefer people getting themself familiar with the filebeat config options instead of trying to "migrate" the config file.
* fields_under_root was removed / deprecated

All of this could be updated. The most important part in the migration docs I think are the logstash-input-beats comments, but these will be obsolete soon. In general I prefer people going from LSF to Filebeat 5.0 to not think of a new version, but more a replacement as it brings lots of new features. Like this I would hope people will read up on the new options available. I agree that people should go directly to 5.x and if need this migration guide for people to be comfortable with this, then we should keep it. But I think it is more work for users to go through the migration guide then just read up on the new options.

tsg:
Shortening it would be a good idea, but I think we should have a guide that LSF users should read before upgrading. It's our chance to set some expectations. For example, the section on the [registry file](https://www.elastic.co/guide/en/beats/filebeat/current/migration-registry-file.html) does a reasonable job in explaining that the format is different and it's best to run then in parallel.

Same with the beats input in Logstash, I think it's valuable and I suspect the plugin will keep supporting LSF during the whole 5.x cycle.

We could shorten the configuration migration part, although I suspect without it people are more likely to get frustrated that they have to figure out everything that changed. 

> fields_under_root was removed / deprecated

Wait, when did that happen? :)

ruflin:
Will update the PR with an "alternative" version. Sorry, I confused `force_close_files` with `fields_under_root`. All still there ;-)

ruflin:
Closing as this is easier to start with a new PR.

PR #2897, opened 2016-11-01
Add fields local to processor

ruflin:
Having the fields definition local to each processor makes it more modular and extendable.

Further changes
* Introduce `make collect` in each beat. Now only `make collect` has to be called for all beats.
* Move processor to its own namespace


andrewkroh:
jenkins, test it

PR #2898, opened 2016-11-01
Move fields under each protocol in packetbeat

ruflin:
This should simplify the extension of packetbeat with further protocols and make it more modular.


ruflin:
The diff is quite hard to read as the output is now sorted alphabetically. I compared the file size of the documents before and after and they seem to stay the same.

monicasarbu:
LGTM

PR #2904, opened 2016-11-01
Cherry-pick to 5.0: Add username/pass options for PostgreSQL

tsg:
Cherry-pick of PR #2890 to 5.0 branch. Original message: 
Similar to #2889 but for PostgreSQL. Also adds docs to the Postgres module,
which were missing, and adjusted the integration tests to use the username
option instead of the full URL.


tsg:
Sorry for the noise here, playing with the Github API.

PR #2907, opened 2016-11-01
Set offset of files under ignore_older to file.size()

ruflin:
Until now if a file was falling under ignore_older, offset 0 was set for the file. In case the file was updated again, all content of the file would be read. This change will set the offset for files falling under ignore_older to the size of the file. This applies on start / restart for files which were not seen before.

The assumption behind this change is, that files falling under ignore_older are normally not updated, and if these files are updated only the newly added lines are expected to be read no the complete file.

The setting of the offset only happens once when it falls for the first time under ignore_older and no state exists. For files which were harvested and then fall under ignore_older, the offset is already file.size().

As this change only applies to files without a state, this should also not have any side affects on Windows where it can happen that a file was updated but the timestamp wasn't and it falls under ignore_older. The reason it doesn't have an affect is that ignore_older only applies an offset if there is no previous state.

### DIFF
--- a/filebeat/prospector/prospector_log.go
+++ b/filebeat/prospector/prospector_log.go
@@ -298,6 +298,8 @@ func (p *ProspectorLog) handleIgnoreOlder(lastState, newState file.State) error 
 		return nil
 	}
 
+	newState.Offset = newState.Fileinfo.Size()
+
 	// Write state for ignore_older file as none exists yet
 	newState.Finished = true
 	err := p.Prospector.updateState(input.NewEvent(newState))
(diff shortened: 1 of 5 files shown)


filebeat/prospector/prospector_log.go
```diff
@@ -298,6 +298,8 @@ func (p *ProspectorLog) handleIgnoreOlder(lastState, newState file.State) error
 		return nil
 	}
 
+	newState.Offset = newState.Fileinfo.Size()
```
urso: Add a comment explaining why we set offset to filesize + why is ignored file added to registry. See PR description.

PR #2908, opened 2016-11-01
Add new conceptual diagram for Filebeat

dedemorton:
The original diagram was easy to read at a glance, but was missing a lot of detail and was misleading because it did not accurately show the role of the prospector. 

I didn't include the registry file in the diagram because I didn't want to make it any more complicated (plus users learn about the registry file later)


monicasarbu:
Great picture @dedemorton LGTM

PR #2911, opened 2016-11-02
Add image that shows metricsets within a module

dedemorton:
This diagram is meant to show the relationship between modules and metricsets. I used redis in the example because it was described in the surrounding text. I think one example is good enough to show how metricsets work.

@ruflin You should probably review this one. :-)


ruflin:
LGTM. One potential improvement: The arrow to the right side should end in Elasticsearch and only have `Metrics` on top of the arrow as part of what is sent to make it more obvious the data is ending up in elasticsearch.

PR #2915, opened 2016-11-02
Don't check stdout on windows

### LINKED ISSUE #2756: console output doesn't work on windows
Using the console output fails on Windows.
- Version: 1.3.1 (didn't test 5.x but it's probably affected)
- Operating System: Windows 2012 (`vagrant up win2012`)
- Steps to Reproduce: Enable console output and start a the beat on the command line (foreground).

```
2016/10/12 03:49:27.119948 outputs.go:120: ERR failed to initialize console plugin as output: console output initialization failed with: GetFileInformationByHandle /dev/stdout: Incorrect function. Error Initialising publisher: console output initialization failed with: GetFileInformationByHandle /dev/stdout: Incorrect function.
2016/10/12 03:49:27.120904 beat.go:161: CRIT console output initialization failld with: GetFileInformationByHandle /dev/stdout: Incorrect function.
```

maddin2016:
fixes #2756 

The only way we can ignore this error, is not to call `Stat()` on windows. Or we use something like this 
```
if runtime.GOOS == "windows" {
		if !isatty.IsTerminal(os.Stdout.Fd()) {
			return nil, fmt.Errorf("console output initialization failed")
		}
	} else {
		if _, err = c.out.Stat(); err != nil {
			return nil, fmt.Errorf("console output initialization failed with: %v", err)
		}
	}
```
But this depends on external library [isatty](https://github.com/mattn/go-isatty)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,6 @@
 /build
 /*/data
 /*/logs
-/**/.vscode
 
 # Files
 .DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 /build
 /*/data
 /*/logs
+/**/.vscode
 
 # Files
 .DS_Store
--- a/libbeat/outputs/console/console.go
+++ b/libbeat/outputs/console/console.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"runtime"
 
 	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/common/op"
@@ -28,8 +29,10 @@ func New(_ string, config *common.Config, _ int) (outputs.Outputer, error) {
 	}
 
 	// check stdout actually being available
-	if _, err = c.out.Stat(); err != nil {
-		return nil, fmt.Errorf("console output initialization failed with: %v", err)
+	if runtime.GOOS != "windows" {
+		if _, err = c.out.Stat(); err != nil {
+			return nil, fmt.Errorf("console output initialization failed with: %v", err)
+		}
 	}
 
 	return c, nil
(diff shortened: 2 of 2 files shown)


andrewkroh:
jenkins, test it

maddin2016:
Mmmh, maybe it's easier to check if winlogbeat from a console or not.

maddin2016:
And i think the whole problem is [here](https://github.com/golang/go/blob/master/src/os/stat_windows.go#L15). Under windows `file` has an attribute `isConsole`. I think they have to check for that attribute.
```
func (file *File) Stat() (FileInfo, error) {
	if file.isConsole {
		return nil, nil
	}
        ...
        ....
```
But i don't know if this is an issue and we should open a PR for go 😕

maddin2016:
How do you import external libraries? Or is this not supported.

andrewkroh:
Let's go with the first commit that doesn't require the extra lib. If we get a bug report about console + Windows service problems then we can reconsider adding the [`GetConsoleMode`](https://msdn.microsoft.com/en-us/library/windows/desktop/ms683167) check.

Thank you for investigating this and finding a solution. 👍

maddin2016:
Ok, then i revert second commit.

maddin2016:
What do you think of the check `isConsole` in `stat_windows.go`. Should i open a PR?

andrewkroh:
If `Stat()` were to check if `isConsole` is set, what should it return in this case?

I wonder if all Windows versions return ERROR_INVALID_PARAMETER when calling [`GetFileInformationByHandle`](https://msdn.microsoft.com/en-us/library/windows/desktop/aa364952) on the stdout handle.

> Should i open a PR?

If you do, I would open an issue first to discuss it.

andrewkroh:
jenkins, test it

Issue #1491, opened 2016-04-26
Unable to filter more than 22 eventIDs


Alreanaes:
- Version: winlogbeat 5.0.0-alpha1
- Operating System: Windows Server 2008R2 (also confirmed on 2012R2)
- Steps to Reproduce: In config file, add line:
  `event_id: 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24`

The relevant output from -e -d "*"

```
DBG  WinEventLog[ForwardedEvents] using subscription query=<QueryList>
  <Query Id="0">
    <Select Path="ForwardedEvents">*[System[(EventID=1 or EventID=2 or EventID=3 or EventID=4 or EventID=5 or EventID=6 or EventID=7 or EventID=8 or EventID=9 or EventID=10 or EventID=11 or EventID=12 or EventID=13 or EventID=14 or EventID=15 or EventID=16 or EventID=17 or EventID=18 or EventID=19 or EventID=20 or EventID=21 or EventID=22 or EventID=23 or EventID=24)]]</Select>
  </Query>
</QueryList>
WARN EventLog[ForwardedEvents] Open() error. No events will be read from this source. The specified query is invalid.
```

According to [KB9704531](https://support.microsoft.com/en-us/kb/970453) , more than 22 event sources need to be split into seperate queries.

I have successfully tested the following query in event viewer where the above query fails

```
<QueryList>
  <Query Id="0" Path="ForwardedEvents">
    <Select Path="ForwardedEvents">*[System[(EventID=1 or EventID=2 or EventID=3 or EventID=4 or EventID=5 or EventID=6 or EventID=7 or EventID=8 or EventID=9 or EventID=10 or EventID=11 or EventID=12 or EventID=13 or EventID=14 or EventID=15 or EventID=16 or EventID=17 or EventID=18 or EventID=19 or EventID=20 or EventID=21 or EventID=22 or EventID=23)]]</Select>
  </Query>
  <Query Id="1" Path="ForwardedEvents">
    <Select Path="ForwardedEvents">*[System[(EventID=24)]]</Select>
  </Query>
</QueryList>
```

andrewkroh:
When I implemented the first round of "filtering" I was thinking in the future Winlogbeat could support more advanced queries like shown below. This would be a work around if it existed.

```
  event_logs:
    # Possible today:
    - name: Application
      include_xml: true
[...]

PR #2919, opened 2016-11-02
Update winlogbeat-options.asciidoc

gmoskovicz:
Fix for #1491.

### DIFF
--- a/winlogbeat/docs/reference/configuration/winlogbeat-options.asciidoc
+++ b/winlogbeat/docs/reference/configuration/winlogbeat-options.asciidoc
@@ -149,6 +149,24 @@ winlogbeat.event_logs:
     event_id: 4624, 4625, 4700-4800, -4735
 --------------------------------------------------------------------------------
 
+[WARNING]
+=======================================
+There is a limit on the query depth to prevent the possibility of a stack overflow from deep recursive queries. This limit is actually set to 22 filters.
+
+More than 22 event id's should be handled with multiple monitors to the same name:
+
+--------------------------------------------------------------------------------
+winlogbeat.event_logs:
+  - name: Security
+    event_id: 4624, 4625, 4700-4800, -4735 ....
+  - name: Security
+    event_id: 4624, 4625, 4700-4800, -4735 ....
+--------------------------------------------------------------------------------
+
+For more information around this, please visit https://support.microsoft.com/en-us/kb/970453.
+=======================================
+
+
 ===== event_logs.level
 
 A list of event levels to include. The value is a comma-separated list of
--- a/winlogbeat/docs/reference/configuration/winlogbeat-options.asciidoc
+++ b/winlogbeat/docs/reference/configuration/winlogbeat-options.asciidoc
@@ -151,23 +151,31 @@ winlogbeat.event_logs:
 
 [WARNING]
 =======================================
-You can specify up to 22 event IDs. Specifying more than 22 event IDs may produce deep recursive queries that could result in a stack overflow. For more information, see https://support.microsoft.com/en-us/kb/970453.
+If you specify more that 22 event IDs to include or 22 event IDs to exclude,
+Windows will prevent Winlogbeat from reading the event log because it limits the
+number of conditions that can be used in an event log query. If this occurs a similar
+warning as shown below will be logged by Winlogbeat, and it will continue
+processing data from other event logs. For more information, see
+https://support.microsoft.com/en-us/kb/970453. 
 
-To workaround this issue you can use different event logs configuration, and/or drop events with processors
+`WARN EventLog[Application] Open() error. No events will be read from this
+source. The specified query is invalid.`
+
+If you have more than 22 event IDs, you can workaround this Windows limitation
+by using a drop_event[drop-event] processor to do the filtering after
+Winlogbeat has received the events from Windows. The filter shown below is
+equivalent to `event_id: 903, 1024, 4624` but can be expanded beyond 22
+event IDs.
 
 [source,yaml]
 --------------------------------------------------------------------------------
-winlogbeat.event_logs:
-  - name: Security
-    event_id: 4624, 4625, 4700-4800, -4735 ....
-  - name: Security
-    event_id: 4624, 4625, 4700-4800, -4735 ....
-    
-processors:
- - drop_event:
-     when:
-       equals:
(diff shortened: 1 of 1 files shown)


tsg:
LGTM, waiting for reviews from @andrewkroh and @dedemorton.

winlogbeat/docs/reference/configuration/winlogbeat-options.asciidoc
```diff
@@ -149,6 +149,24 @@ winlogbeat.event_logs:
     event_id: 4624, 4625, 4700-4800, -4735
 --------------------------------------------------------------------------------
 
+[WARNING]
+=======================================
+There is a limit on the query depth to prevent the possibility of a stack overflow from deep recursive queries. This limit is actually set to 22 filters.
+
+More than 22 event id's should be handled with multiple monitors to the same name:
+
+--------------------------------------------------------------------------------
```
tsg: you can add `[source,yaml]` on this to get syntax highlighting.

winlogbeat/docs/reference/configuration/winlogbeat-options.asciidoc
```diff
@@ -149,6 +149,24 @@ winlogbeat.event_logs:
     event_id: 4624, 4625, 4700-4800, -4735
 --------------------------------------------------------------------------------
 
+[WARNING]
+=======================================
+There is a limit on the query depth to prevent the possibility of a stack overflow from deep recursive queries. This limit is actually set to 22 filters.
+
+More than 22 event id's should be handled with multiple monitors to the same name:
```
andrewkroh: `s/id/ID/`

winlogbeat/docs/reference/configuration/winlogbeat-options.asciidoc
```diff
@@ -149,6 +149,24 @@ winlogbeat.event_logs:
     event_id: 4624, 4625, 4700-4800, -4735
 --------------------------------------------------------------------------------
 
+[WARNING]
+=======================================
+There is a limit on the query depth to prevent the possibility of a stack overflow from deep recursive queries. This limit is actually set to 22 filters.
```
andrewkroh: `actually` could be dropped.

gmoskovicz:
@andrewkroh 

I am happy to remove the workaround, but if we add this in the documentation, then it means that we need to suggest a workaround. Do you want me to remove the workaround? What other workaround do you see here? If we do what you propose , do we still need this to be added in the documentation ?

andrewkroh:
@gmoskovicz, I think another possible workaround would be to use a processor to do the filtering.

I can spend some time on investigating the change to the query builder. If I can implement that then we don't need this in the documentation. But since it's a problem now for 5.0.0 we should merge this PR with a workaround that uses a processor. Can you see if you can setup a processor to do the desired filtering?

gmoskovicz:
I'll work with this and get back here.

gmoskovicz:
@andrewkroh 

Works, but the configuration gets larger since you need a huge OR with equal conditions:

```
processors:
 - drop_event:
     when:
       equals:
         event_id: 50037
```

winlogbeat/docs/reference/configuration/winlogbeat-options.asciidoc
dedemorton: We need to avoid repeating content from other websites verbatim (the content is likely copyrighted). This sentence is taken directly from the Microsoft website: "There is a limit on the query depth to prevent the possibility of a stack overflow from deep recursive queries." 

It would be better anyhow to state this limitation in terms of what the user specifies in the Winlogbeat config. How about something like this: 

You can specify up to 22 event IDs. Specifying more than 22 event IDs may produce deep recursive queries that could result in a stack overflow. For more information, see https://support.microsoft.com/en-us/kb/970453. 

I didn't suggest edits to the text about the workaround because it sounds like you are still discussing the best approach to solving this issue.

winlogbeat/docs/reference/configuration/winlogbeat-options.asciidoc
```diff
@@ -149,6 +149,24 @@ winlogbeat.event_logs:
     event_id: 4624, 4625, 4700-4800, -4735
 --------------------------------------------------------------------------------
 
+[WARNING]
```
gmoskovicz: Keeping [WARNING] and added [source,yaml]

winlogbeat/docs/reference/configuration/winlogbeat-options.asciidoc
gmoskovicz: Fixed.

winlogbeat/docs/reference/configuration/winlogbeat-options.asciidoc
gmoskovicz: FIXED

gmoskovicz:
@andrewkroh @tsg Can i have final review of this? Given that there is no consensus around the workaround, i've suggested one using multiple processors and/or multiple IDs at the input level.

winlogbeat/docs/reference/configuration/winlogbeat-options.asciidoc
```diff
@@ -149,6 +149,30 @@ winlogbeat.event_logs:
...
+=======================================
+You can specify up to 22 event IDs. Specifying more than 22 event IDs may produce deep recursive queries that could result in a stack overflow. For more information, see https://support.microsoft.com/en-us/kb/970453.
+
+To workaround this issue you can use different event logs configuration, and/or drop events with processors
+
+[source,yaml]
+--------------------------------------------------------------------------------
+winlogbeat.event_logs:
```
andrewkroh: I investigated automatically splitting up the event IDs across two queries. It gets complicated due to the interactions with the suppress filters (e.g. -1, -2). I think we should wait until https://github.com/elastic/beats/issues/1491#issuecomment-215085489 is implemented and in the meantime use the `drop_event` processor as a workaround to either include or exclude events.

I don't want to recommend creating duplicate `Security` event log consumers because it can potentially result in data loss in some conditions. And in the future we may want to disallow duplicate names to prevent the aforementioned condition (but that's TBD).

gmoskovicz:
@andrewkroh sounds good for me! I've updated it to the latest version

tsg:
Belated LGTM :)

PR #2924, opened 2016-11-02
Output without the json wrapper

treff7es:
Initial implemenation of #2000 issue to be able specify format for the outputs (console, file, kafka, redis).

For example:

```
output.kafka:
  format: "%{[@timestamp]} %{[message]}"

```

I had an earlier implemetation which I had issues so I redid it to follow the common template format.

### DIFF
--- a/libbeat/outputs/console/console.go
+++ b/libbeat/outputs/console/console.go
@@ -6,10 +6,10 @@ import (
 	"os"
 
 	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/common/fmtstr"
 	"github.com/elastic/beats/libbeat/common/op"
 	"github.com/elastic/beats/libbeat/logp"
 	"github.com/elastic/beats/libbeat/outputs"
-    "github.com/elastic/beats/libbeat/common/fmtstr"
 )
 
 func init() {
@@ -53,26 +53,26 @@ func (c *console) PublishEvent(
 	var serializedEvent []byte
 	var err error
 
-    if c.config.Format != nil {
-        formattedEvent, err := c.config.Format.Run(data.Event)
-        if err != nil {
-            logp.Err("Fail to format event (%v): %#v", err, data.Event)
-            op.SigCompleted(s)
-            return err
-        }
-        serializedEvent = []byte(formattedEvent)
-    }else {
-        if c.config.Pretty {
-            serializedEvent, err = json.MarshalIndent(data.Event, "", "  ")
-        } else {
-            serializedEvent, err = json.Marshal(data.Event)
-        }
-        if err != nil {
-            logp.Err("Fail to convert the event to JSON (%v): %#v", err, data.Event)
-            op.SigCompleted(s)
-            return err
-        }
-    }
+	if c.config.Format != nil {
+		formattedEvent, err := c.config.Format.Run(data.Event)
+		if err != nil {
+			logp.Err("Fail to format event (%v): %#v", err, data.Event)
+			op.SigCompleted(s)
+			return err
+		}
+		serializedEvent = []byte(formattedEvent)
+	} else {
+		if c.config.Pretty {
+			serializedEvent, err = json.MarshalIndent(data.Event, "", "  ")
+		} else {
+			serializedEvent, err = json.Marshal(data.Event)
+		}
+		if err != nil {
+			logp.Err("Fail to convert the event to JSON (%v): %#v", err, data.Event)
+			op.SigCompleted(s)
+			return err
+		}
+	}
 
 	if err = c.writeBuffer(serializedEvent); err != nil {
 		goto fail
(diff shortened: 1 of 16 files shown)


libbeat/outputs/console/console.go
```diff
@@ -49,21 +50,31 @@ func (c *console) PublishEvent(
...
+		if c.config.Pretty {
+			serializedEvent, err = json.MarshalIndent(data.Event, "", "  ")
+		} else {
+			serializedEvent, err = json.Marshal(data.Event)
+		}
+		if err != nil {
+			logp.Err("Fail to convert the event to JSON (%v): %#v", err, data.Event)
+			op.SigCompleted(s)
+			return err
+		}
```
urso: maybe error handling can be made a little 'nicer' here. Consider `defer op.SigCompleted(s)` at top of PublishEvent to ensure it's never forgotten.

libbeat/outputs/fileout/file.go
```diff
@@ -76,16 +80,30 @@ func (out *fileOutput) PublishEvent(
...
-		return err
+	} else {
+		serializedEvent, err = json.Marshal(data.Event)
+		if err != nil {
+			// mark as success so event is not sent again.
+			op.SigCompleted(sig)
+
+			logp.Err("Fail to json encode event(%v): %#v", err, data.Event)
+			return err
+		}
```
urso: This kind of encoding/formatting strategy seems to become a common pattern. Consider introducing some abstraction via of 'encoder'/'formatter' interface/function instead of  adding if/else branches. Should get us a more clean error handling again.

libbeat/outputs/fileout/file.go
```diff
@@ -76,16 +80,30 @@ func (out *fileOutput) PublishEvent(
...
 ) error {
-	jsonEvent, err := json.Marshal(data.Event)
-	if err != nil {
-		// mark as success so event is not sent again.
-		op.SigCompleted(sig)
+	var serializedEvent []byte
+	var err error
+
+	if out.format != nil {
+		formattedEvent, err := out.format.Run(data.Event)
```
urso: use `RunBytes` method to prevent additional copy operations.

libbeat/outputs/console/console.go
```diff
@@ -49,21 +50,31 @@ func (c *console) PublishEvent(
...
 	data outputs.Data,
 ) error {
-	var jsonEvent []byte
+	var serializedEvent []byte
 	var err error
 
-	if c.config.Pretty {
-		jsonEvent, err = json.MarshalIndent(data.Event, "", "  ")
+	if c.config.Format != nil {
+		formattedEvent, err := c.config.Format.Run(data.Event)
```
urso: `RunBytes`

urso:
Thanks for the PR!!! Great to see you also added tests.

I see some code/duplication introducing some if-else branches here. I'd rather see the formatting/encoding being generalized in code as well by having some common encoding interface potentially being satisfied by different implementations. For example:

```
type Encoding func(event common.MapStr) ([]byte, error)
```

Or 

```
type Encoding interface {
  Encode(common.MapStr) ([]byte, error)
}
```

urso:
@treff7es any updates on this PR? Anything unclear about the code review?

treff7es:
@urso Terrible sorry but I was busy with my work. I would like to do the requested changes next week.

treff7es:
Finally, I could spend time with this. Sorry again for the delay. 
As you can see I created a writer interface and json/formatted writing is implemented by a writers which implements the interface.
I moved the writer specific configs into a separate writer map:
```
output.kafka:
   writer.type: "FormatStringWriter"
   writer.format: "%{[@timestamp]} %{[message]}"

```
You need to specify the type of writer otherwise it will default to _JsonWriter_. 
Currently there are two type of writer:
_JsonWriter_ -> So far this was the only behaviour. You can set _writer.pretty_ to have the pretty printing at json writing. -> You could set it in console output. Now this parameter moved to writer instead of the console output. I think it make more sense there.
```
output.kafka:
   writer.type: "JsonWriter"
   writer.pretty: true
```

_FormatStringWriter_ -> This is the new writer type which does the formatting if you specify the _format_ parameter. It will default to the "%{[message]}" if you fail to specify the format.

The only breaking change is the move of the pretty printing to writer.

Please, if you can, check this pull request if you like the current implementation and if you have any comment. I really appreciate your previous feedbacks.

treff7es:
@urso I have updated the pull request. Please, if you have time guys check if you like it in this form.

urso:
Thanks for the PR. I will clean the code up a little and documentation in a follow-up PR.

I'm planing to change the config file format to:

```
output.X:
  codec.format: '...'
```

or

```
output.X:
  codec.json:
    pretty: ...
```

that is rename `writer` to `codec` and use the config namespace to define the codec type.

treff7es:
Cool, thanks! Sorry about the writer naming I couldn't come up with a better one. :(

raboof:
Since 977b6d8ed9954472203cead990bf87ce537200aa, it seems setting `output.console.pretty = true` no longer has effect, I need to set `output.console.writer.pretty`. 

Is that intentional?

I see there's still a `pretty` field in the `console` config, but that value doesn't appear to propagate to the Writer.

treff7es:
@raboof There was a followup pull request which addressed this issue and reworked my original pull request:
https://github.com/elastic/beats/pull/3285

raboof:
Aah looks like I wasn't completely up-to-date indeed - thanks for the pointer!

PR #2930, opened 2016-11-03
Use the flows search when displaying Flows Dashboard

monicasarbu:
Use the `Packetbeat-Flows-Search` when displaying the flows data in the `Packetbeat-Flows` Dashboard.

This PR is also fixing the issue that `Packetbeat-Flows-Search` was not imported as there was no dashboard that was using it.


monicasarbu:
This change is related to https://github.com/elastic/beats/pull/2899

PR #2931, opened 2016-11-03
Remove generated fields.yml from git repositories

ruflin:
As some of the generated fields are also needed for the tests, make collect must be run in some cases before running the tests. Testsuite does this automatically. It was not added to each test as it would run multiple times and causes some issue with docker container access rights.

* Adapt script to rely on generated fields.generated.yml file if it exists
* Add fields.generate.yml to gitignore
* Make update dependent on collect call instead of the other way around as update requires now collect.
* Make system-tests use fields.generated.yml
* Remove obsolete metricbeat.json file

NOTE: More fields.yml generated fields can be removed in the future. This PR provides the basic logic for it.


metricbeat/Makefile
```diff
@@ -22,21 +22,22 @@ include ${ES_BEATS}/libbeat/scripts/Makefile
...
 	mkdir -p etc/kibana
 	-cp -r module/*/_meta/kibana etc/
 
 # Collects all module and metricset fields
 .PHONY: fields
 fields:
 	mkdir -p etc/
-	cat ${ES_BEATS}/metricbeat/etc/_meta/fields_base.yml > etc/fields.yml
-	. ${PYTHON_ENV}/bin/activate; python ${ES_BEATS}/metricbeat/scripts/fields_collector.py >> etc/fields.yml
+	cat ${ES_BEATS}/metricbeat/etc/fields.yml > etc/fields.generated.yml
```
monicasarbu: @ruflin I am not sure I understand the reason for replacing `_meta/fields_base.yml` with `fields.yml`. It supposed to contain common fields that are exported by all Metricbeat modules, right? If yes, then wouldn't make more sense to call it something more obvious  like `etc/fields_common.yml`?

ruflin: fields.yml contains field which are common for all metricbeat events. The reason I called it fields.yml that in filebeat or winlogbeat the file with all the common options is also called `fields.yml`. `fields_common.yml` would also be an option, but my idea was to call all non generated files which contain `fields.yml`, `fields.yml`.

andrewkroh: I like `fields.common.yml` for the Beats where it only contains a subset of the fields. I like this name because it makes it clear that it's not the complete fields.yml.

ruflin: Both works for me. I renamed it to fields.common.yml

Issue #2788, opened 2016-10-14
tail_files should persist initial offset as end of file


andrewkroh:
At startup Filebeat initializes the registry with info about the file that it opens (but this does not contain an offset so it’s left as 0). This is the expected behavior for `tail_files: false` as we do not want to advance the offset until some data has been read. But for `tail_files: true` my personal expectation is that it would initially write the offset as the end of the file, instead of 0, so that if Filebeat exits it can resume reading from that point.
- Version: 1.3.1
- Operating System: affects all
- Steps to Reproduce:
  - Delete .filebeat registry file.
  - Write multiple lines of text to `out.log`.
  - Start filebeat with `tail_file: true`.
  - Stop filebeat.
  - Read the .filebeat registry. It will have an offset of 0 because no events were written to the file while Filebeat was reading.

Config:

```
filebeat:
  prospectors:
    - paths:
        - out.log
      input_type: log
      tail_files: true
      close_older: 1s
```

andrewkroh:
https://github.com/elastic/beats/blob/162fb252d855819c7c93e0759c0cf6b9069e0f79/filebeat/crawler/prospector.go#L402-L405

If the `Havester.Start()` wasn't asynchronous or it initialized the state prior to starting the havester goroutine,  then this would work as expected because `h.GetState()` would return a fully initialized `input.FileState` containing the offset.

ruflin:
This should be solved together with https://github.com/elastic/beats/issues/2613 by moving `tail_files` logic to the prospector.

ruflin:
Closing as this was solved in https://github.com/elastic/beats/pull/2932 and should be available in 5.1. As 5.0 is GA and filebeat 5.0 is backward compatible, I suggest we do not try to backport this.

PR #2932, opened 2016-11-03
Move tail_files to prospector level

ruflin:
tail_files is now only applied on the first run and after that ignored. Also the state for all files falling under tail_files and not having a state, a state will directly be written.

* Implement tail_files by setting ignore_older to 1ns for the first run
* Fix typo in stats variable names

Closes https://github.com/elastic/beats/issues/2613 and https://github.com/elastic/beats/issues/2788

### DIFF
--- a/filebeat/prospector/prospector_log.go
+++ b/filebeat/prospector/prospector_log.go
@@ -15,8 +15,8 @@ import (
 )
 
 var (
-	filesRenamed  = expvar.NewInt("filebeat.prospector.log.files.renamed")
-	filesTrucated = expvar.NewInt("filebeat.prospector.log.files.truncated")
+	filesRenamed   = expvar.NewInt("filebeat.prospector.log.files.renamed")
+	filesTruncated = expvar.NewInt("filebeat.prospector.log.files.truncated")
 )
 
 type ProspectorLog struct {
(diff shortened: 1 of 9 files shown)


urso:
This PR applies tail_files on every restart to every new file discovered since last run.

I'd rather see tail_files only being effective if no registry exists (or is empty on startup).

filebeat/prospector/prospector_log.go
```diff
@@ -60,6 +61,19 @@ func (p *ProspectorLog) Init(states []file.State) error {
...
+
+		// Overwrite ignore_older for the first scan
+		p.config.IgnoreOlder = 1
+		defer func() {
+			// Reset ignore_older after first run
+			p.config.IgnoreOlder = ignoreOlder
+			// Disable tail_files after the first run
+			p.config.TailFiles = false
+		}()
+	}
```
urso: how about:

```
if p.config.TailFiles {
  defer func(old time.Duration) {
    p.config.IgnoreOlder = old
  }(p.config.IgnoreOlder)

  p.config.TailFiles = false // disable tail_files in future calls
  p.config.IgnoreOlder = 1 * time.Nanosecond
}
```

ruflin: Should also work. The advantage of the first version is, that in case TailFiles is also used somewhere inside the first scan (which is not the case), it would still work.

PR #2938, opened 2016-11-04
Remove trailing slash from metricbeat and packetbeat markdown links

jbudz:
Links in Kibana with a trailing slash will cause a server redirect and full page refresh, causing us to lose browser state.

It's hitting this route: https://github.com/elastic/kibana/blob/05c12bbdd35f0a55080906dd099a11069eb2fd0e/src/server/http/index.js#L106


monicasarbu:
@jbudz Thank you for the fix.

PR #2956, opened 2016-11-08
Introducing named prospectors

ruflin:
Named prospectors allow to define prospectors with a name instead of an array. The name of the prospector will automatically be set as `document_type` field. The config with named and unnamed prospectors looks as following:

```
filebeat.prospectors:
- paths:
    - test.log

filebeat.prospectors.apache:
  paths:
    - apache.log
```

It is not possible to mix named and unnamed directly. This change can make configurations more readable but has the downside, that named prospectors must be unique. In general it is recommended to only use named or unnamed and not mix them in one configuration file.

### DIFF
--- a/filebeat/beater/filebeat.go
+++ b/filebeat/beater/filebeat.go
@@ -76,6 +76,11 @@ func (fb *Filebeat) Run(b *beat.Beat) error {
 		return err
 	}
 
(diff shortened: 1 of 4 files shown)


ruflin:
Closing as we currently do not need this for modules.

PR #2957, opened 2016-11-08
Improve docker-composes handling

ruflin:
* Switch to official kibana docker files
* Make use of docker-compose v2 args params

x-pack is currently still disabled.

### DIFF
--- a/testing/environments/snapshot.yml
+++ b/testing/environments/snapshot.yml
@@ -6,11 +6,15 @@ services:
     build:
       context: ./docker/elasticsearch
       dockerfile: Dockerfile-snapshot
+      args:
+        ELASTIC_VERSION: 6.0.0-alpha1-SNAPSHOT
+        ES_DOWNLOAD_URL: https://snapshots.elastic.co/downloads/elasticsearch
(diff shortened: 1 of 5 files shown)


testing/environments/snapshot.yml
```diff
@@ -21,4 +25,6 @@ services:
     build:
       context: ./docker/kibana
       dockerfile: Dockerfile-snapshot
-
+      args: [ ]
+        #KIBANA_DOWNLOAD_URL: https://snapshots.elastic.co/downloads/kibana/kibana-6.0.0-alpha1-SNAPSHOT-linux-x86_64.tar.gz
```
ruflin: This is left out currently as it throws errors in the inherited compose files in the tests which don't require the variables. Variables were left in as "comment"

tsg: So we don't need them at all? Or do we have to uncomment this in some situation?

ruflin: I'm hoping to find a solution to be able to uncomment it and use the "pure" docker file from the infra team. But I didn't want to spend too much time right now on figuring out on how this is possible with version 2 of the docker compose file.

PR #2958, opened 2016-11-08
Add ParseHost() for parsing configured hosts in the Module

andrewkroh:
Add HostParser function that can be registered with MetricSetFactories

The ParseHost() function will combine configuration data with the raw host values to produce a value appropriate for use by the MetricSet to make a connection.

Currently the HostParser parameter is optional when registering a MetricSetFactory, but it will become required once all MetricSets are updated to have a HostParser.

**updated**: Monday, Nov 21


tsg:
I wonder if we should really worry about BWC at this point. I don't think there are any external Metricbeat based Beats, and even if they were, it would be trivial to provide an empty implementation for that function.

ruflin:
I like that it is BC compatible as currently the module file normally does not exist. I would not necessarely tie it to 6.0 to make it mandatory, but only make it mandatory when we got some more experience with it to make sure external developers only need to change it once. 

For the generator we should provide also an empty method which potentially logs an error, that it is not implemented.

In the future we should offer "generic" ParseHost() methods for example for Http so these can be reused.

andrewkroh:
I changed my approach on this because providing the parsing functionality at the module level wasn't sufficient. Individual MetricSets within the same module can have unique host parsing requirements. For example, nginx MetricSets for stub-status and plus-status might take the same `host` value and append different paths.

So my approach is to allow (_require_ in the future) registering a `HostParser` function with the `MetricSetFactory`. If all MetricSets in a Module can share the same `HostParser` function then you just write that inside the module package and use it in each MetricSet registration. We can provide some utilities in a helper package to cut down on the code duplication.

As for not breaking BWC, it looks like there's only one external MetricSet at the moment. So we won't break much when we require the `HostParser` function at registration time. We don't have to wait until 6.0 to do this. Once each of our MetricSets has a `HostParser` I'll make it a requirement. I'll implement `HostParser`s for the other MetricSets in a separate PR.

metricbeat/mb/mb.go
```diff
@@ -57,6 +61,7 @@ type MetricSet interface {
 	Host() string   // Host returns a hostname or other module specific value
 	// that identifies a specific host or service instance from which to collect
 	// metrics.
+	HostInfo() HostInfo // HostInfo returns the parsed host data.
```
ruflin: Should we call this HostData()?

metricbeat/mb/mb.go
```diff
@@ -78,9 +83,23 @@ type EventsFetcher interface {
...
+	name     string
+	module   Module
+	host     string
+	hostInfo HostInfo
+}
+
+// HostInfo contains values parsed from the 'host' configuration. Other
+// configuration data like protocols, usernames, and passwords may also be
+// used to construct this HostInfo data.
+type HostInfo struct {
```
ruflin: HostData ?

metricbeat/mb/helper/url.go
```diff
@@ -0,0 +1,91 @@
+package helper
```
ruflin: should we move the helper package to the top level and not under `mb`?

metricbeat/module/apache/status/status.go
```diff
@@ -31,51 +30,49 @@ var (
...
-}
-
-// New creates new instance of MetricSet.
-func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
-	// Additional configuration options
+// ParseServerStatusURL takes the configuration data and host value and
+// constructs a HostInfo object containing the URI to the Apache server status
+// endpoint.
+func ParseServerStatusURL(mod mb.Module, host string) (mb.HostInfo, error) {
 	config := struct {
```
ruflin: I kind of liked that config part was in `New`. But I see that having it here will make it easier reusable for the different modules.

metricbeat/mb/mb.go
andrewkroh: Changed to `HostData`.

metricbeat/mb/helper/url.go
andrewkroh: I left the package inside of `mb` because it contains functions that apply specifically to the things in the `mb`. I did rename it from the generic `helpers` to `parse` which is more specific and less likely to collide with a future helpers package that may exist someday.

andrewkroh:
jenkins, test it

metricbeat/_meta/beat.full.yml
```diff
@@ -121,15 +121,16 @@ metricbeat.modules:
   #enabled: true
   #period: 10s
 
-  # Host DSN should be defined as "tcp(127.0.0.1:3306)/"
-  # The username and password can either be set in the DSN or for all hosts in username and password config option
-  #hosts: ["root@tcp(127.0.0.1:3306)/"]
+  # Host DSN should be defined as "user:pass@tcp(127.0.0.1:3306)/"
+  # The username and password can either be set in the DSN or using the username
+  # and password config options. Those specified in the DSN take precedence.
+  hosts: ["root:secret@tcp(127.0.0.1:3306)/"]
```
ruflin: This should probably be commented out?

metricbeat/mb/mb.go
```diff
@@ -78,9 +83,23 @@ type EventsFetcher interface {
...
+
+// HostData contains values parsed from the 'host' configuration. Other
+// configuration data like protocols, usernames, and passwords may also be
+// used to construct this HostData data.
+type HostData struct {
+	URI          string // The full URI that should be used in connections.
+	SanitizedURI string // A sanitized version of the URI without credentials.
+
+	// Parts of the URI.
+	Host     string // The host and possibly port.
```
ruflin: In the future we can perhaps extract also the port in case this could become useful.

metricbeat/module/docker/config.go
```diff
@@ -1,23 +1,16 @@
...
+type TLSConfig struct {
+	Enabled     *bool  `config:"enabled"`
+	CA          string `config:"certificate_authority"`
+	Certificate string `config:"certificate"`
+	Key         string `config:"key"`
 }
 
-func GetDefaultConf() Config {
-	return Config{
-		Socket: "unix:///var/run/docker.sock",
```
ruflin: No need for the default config anymore?

metricbeat/module/mongodb/mongodb.go
```diff
@@ -0,0 +1,63 @@
...
+	"strings"
+
+	"github.com/elastic/beats/metricbeat/mb"
+	"github.com/elastic/beats/metricbeat/mb/parse"
+
+	mgo "gopkg.in/mgo.v2"
+)
+
+func init() {
+	// Register the ModuleFactory function for the "mysql" module.
```
ruflin: This should say mongodb not mysql

metricbeat/module/mongodb/mongodb.go
```diff
@@ -0,0 +1,63 @@
...
+	if err := mb.Registry.AddModule("mongodb", NewModule); err != nil {
+		panic(err)
+	}
+}
+
+func NewModule(base mb.BaseModule) (mb.Module, error) {
+	// Validate that at least one host has been specified.
+	config := struct {
```
ruflin: we could turn this around and make hosts required by default and only system module overwrites it to not be required.

andrewkroh: You're right, the system module is more the exception than the rule w.r.t. not requiring or accepting a host. 

While making these changes I had an idea to accept a vararg of options when registering metricsets, like `AddMetricSet("system", "process", New, OptHostsNotAllowed, OptPushesData, OptSomeOtherAttribute)`.

ruflin: +1 on making it a list of params instead of a single one, as I also think we will add more int he future.

PR #2964, opened 2016-11-08
Split release notes by version

monicasarbu:
I would like to propose the following solution for organizing the release notes for 5.x, that it's inspired from the Elasticsearch project.

- Use `beats/CHANGELOG.asciidoc` to write the changelog for master
- When releasing version 5.x.y, copy the content of`beats/CHANGELOG.asciidoc` into `libbeat/docs/release-notes/5.x.y.asciidoc`, and add a link to the 5.x.y into `libbeat/docs/release.asciidoc`.


ruflin:
In general I'm +1 on having separate release notes asciidoc for each release. I also think we should split it up by beat. But the process should be automated based on https://github.com/elastic/beats/pull/1879 as otherwise it will involve copy around lots of items with each release. 

We should use this PR to discuss what the end result should be for all branches so I can do try to do it in https://github.com/elastic/beats/pull/1879

dedemorton:
I don't like the practice of breaking down the 5.0 release notes by alpha and beta releases. Once the product is GA-level, I feel that we should no longer support alpha and beta installations or provide content about pre-releases. We should expect users to install the GA-level software. 

By including so many sections of release notes, we actually make it a bit harder for users who are simply looking for 5.0.0 release notes.  They will have to hunt and peck a bit to find what they need, and this situation will get worse when we add 5.0.1, 5.0.2, and so on. They won't immediately understand the difference between 5.0.0 and 5.0.0-GA (ES gets around this by calling them "combined" release notes, but that leaves it up to users to figure out what we mean by "combined"). It won't be clear to them which parts they need to read. And there's no telling where google searches will send them when we have so many topics that mention 5.0. IMHO, anything that confuses the user degrades the user experience with the docs.

If we need to break the relnotes down by alpha/beta versions simply because of how we pull in the changelog, then I propose that we include only 5.0 relnotes. Including links to the 1.x relnotes in the 5.0 docs simply does not make sense and is a bit overwhelming. Users might think they need to look through *all* the release notes and feel frustrated by that.

Also, in an ideal world, each Beat would have its own release notes that pertain to that Beat (but I  think we can shelf the discussion around this for now).

debadair:
Consolidating the pre-release release notes for major releases at GA is by far the better user experience. I think the tension here is between automatically generating the release notes, and having to update them manually. For major releases, I think spending the extra time to consolidate and curate the release notes is worthwhile. We did that for X-Pack and Logstash & I think it would be great if we could adopt that practice across the stack.

monicasarbu:
@debadair We have already consolidated release notes or "combined" under the 5.0.0 link. Definitely a good idea to have it.

@dedemorton We discussed today in our team meeting about including or not the 1.x release notes, and we agreed to include them for now, as there aren't too many anyway. They are useful for our users to see what has changed if they are updating from 1.x to 5.x. I think the same discussion applies also to alphas, betas and rcs, as they are useful for a user that is updating from 5.0.0-beta1 to 5.0.1.

cc-ed @ruflin

ruflin:
LGTM. I like that it is in the 5.0 branch so we can try it out for the 5.0 release and see how it goes. The downside is that all work has to be done again for 5.1.

dedemorton:
I started to respond, but then realized I would just be repeating what I said earlier. With all of these different sections, I think it's likely that users will not read the release notes carefully and will probably miss things that are important. I wonder if we can introduce some nesting in the TOC? And also tweak the name of the combined list. Not sure what's possible in our generation process for the relnotes. 
`````
Beats version 5.0.1
Beats version 5.0.0 - (this would be a container, not the combined relnotes)
     All Beats 5.0.0 Versions Combined
     Beats version 5.0.0-GA
     Beats version 5.0.0-rc1
     Beats version 5.0.0-beta1
     Beats version 5.0.0-alpha5
     Beats version 5.0.0-alpha4
     Beats version 5.0.0-alpha3
     Beats version 5.0.0-alpha2
     Beats version 5.0.0-alpha1
Beats version 1.3.1
`````

monicasarbu:
Closing the PR and not everyone agrees on it.

PR #2966, opened 2016-11-09
add description field on init.d script for chkconfig add command

tommy-sec:
I found the luck of description field of /etc/init.d/filebeat.
After this commit, we can "chkconfig --add filebeat".
The description field's sentence is copied from /etc/init.d/elasticsearch.


karmi:
Hi @tommy-sec, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/beats/pull/2966.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

tommy-sec:
Hi. Due to a personal issue, I resend the request. Thanks!

PR #2969, opened 2016-11-09
Kafka module

ruflin:


metricbeat/module/kafka/partition/partition.go
```diff
@@ -0,0 +1,97 @@
...
+
+// Fetch partition stats list from kafka
+func (m *MetricSet) Fetch() ([]common.MapStr, error) {
+
+	config := sarama.NewConfig()
+	config.Net.DialTimeout = m.Module().Config().Timeout
+	config.Net.ReadTimeout = m.Module().Config().Timeout
+	config.ClientID = "metricbeat"
+
+	client, err := sarama.NewClient([]string{m.Host()}, config)
```
urso: error check missing?

Why just one host. Often one configures a set of 'bootstrap' endpoints to query meta-data. If one fails, ask next endpoint.

metricbeat/module/kafka/partition/partition.go
```diff
@@ -0,0 +1,97 @@
...
+	topics, err := client.Topics()
+	if err != nil {
+		return nil, err
+	}
+
+	events := []common.MapStr{}
+	for _, topic := range topics {
+		partitions, err := client.Partitions(topic)
+		if err != nil {
+			logp.Err("Fetch partition info for topic %s: %s", topic, err)
```
urso: is there a chance partitions not being empty in err? We still want to process potentially incomplete data?

metricbeat/module/kafka/partition/partition.go
```diff
@@ -0,0 +1,97 @@
...
+			oldestOffset, err := client.GetOffset(topic, partition, sarama.OffsetOldest)
+			if err != nil {
+				logp.Err("Fetching oldest offset information for partition %s in topic %s: %s", partition, topic, err)
+			}
+
+			// List of replicas ids for the given partition
+			replicas, err := client.Replicas(topic, partition)
+			if err != nil {
+				logp.Err("Fetch replicas for partition %s in topic %s: %s", partition, topic, err)
+			}
```
urso: It's kinda sad we don't get the replica offsets: https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-OffsetAPI(AKAListOffset)

metricbeat/module/kafka/partition/_meta/fields.yml
```diff
@@ -0,0 +1,38 @@
+- name: partition
+  type: group
+  description: >
+    partition
+  fields:
+    - name: brocker
+      type: integer
+      description: >
+        Broker id
```
ruflin: Good idea. I also added the address

metricbeat/module/kafka/partition/_meta/fields.yml
```diff
@@ -0,0 +1,38 @@
...
+      description: >
+        Partition id.
+    - name: topic
+      type: keyword
+      description: >
+        Topic name
+    - name: replicas
+      type: list
+      description: >
+        List of replicas ids
```
ruflin: Changed it. I saw that there are some differences with notes and was thinking about adding a "status" field but not for the first version.

metricbeat/module/kafka/partition/_meta/fields.yml
```diff
@@ -0,0 +1,38 @@
...
+    - name: broker
```
ruflin: removed

metricbeat/module/kafka/partition/partition.go
ruflin: error check added.

metricbeat/module/kafka/partition/partition.go
ruflin: Looking at the Sarama code, it looks like if an error is returned, partitions is always nil

metricbeat/module/kafka/partition/partition.go
ruflin: I'm not even sure if we should provide the replicas ids here or not.

ruflin:
jenkins, retest it

metricbeat/module/kafka/partition/partition.go
urso: In kafka (in comparison to other services), you do not monitor one host, but  you monitor one cluster. That is, when configuring the `hosts`, all hosts MUST belong to the same cluster. Having multiple `bootstraping` hosts is for redundancy, to decrease the chances of errors.

ruflin: For the first version we should recommend to only use one Host as part of the docs but think long term on how we handle these kind of situations in metricbeat as this will not only apply to Kafka.

metricbeat/module/kafka/partition/partition.go
ruflin: It will return an error event. But the problem is that it will abort the execution of the event fetch for all coming topics.

urso: In this case, we might consider adding an error event to `events` here.

metricbeat/module/kafka/partition/partition.go
urso: Hmmm... is it required by module-interface to have a method doing `Host() string`? Why not have the module/metricset decide?

metricbeat/module/kafka/partition/partition.go
```diff
@@ -0,0 +1,103 @@
...
+				"offset": common.MapStr{
+					"oldest": oldestOffset,
+					"newest": newestOffset,
+				},
+				"replicas": replicas,
+				"broker": common.MapStr{
+					"id":      brocker.ID(),
+					"address": brocker.Addr(),
+				},
+			}
```
urso: Hmm... I was in hope for getting the offset per replica and create an event per replicate. The `broker` is the current leader to a partition. All other brokers handling the same partition will be the replica. Every now and then the leader is re-elected, turning the broker into an replica and another former replica into the leader.

Some event like:

```
common.MapStr {
  "topic": topic,
  "partition": partition,
  "broker": common.MapStr {
    "id": id,
    "address": brokers[id],
  }
  "offset": offset,
  "leader": id == leaderID,
}
```

The address of every broker and broker id is included in meta-data.

metricbeat/module/kafka/partition/partition.go
ruflin: That would mean that we have an "error" entry at the wrong namespace of the event. Will open a separate issue to discuss this improvement for Metricbeat.

ruflin: Here is the related issue: https://github.com/elastic/beats/issues/3004

metricbeat/module/kafka/partition/partition.go
ruflin: If the issue is only Kafka specific, we could also have a `bootstrap_hosts: [...]` config option for the Kafka module. Lets handle this as soon as it pops up as an issue.

PR #2971, opened 2016-11-09
Add system tests to check document fields for docker module

ruflin:
Part of https://github.com/elastic/beats/issues/2629


metricbeat/tests/system/test_docker.py
```diff
@@ -0,0 +1,177 @@
...
+        self.assertNotRegexpMatches(log.replace("WARN EXPERIMENTAL", ""), "ERR|WARN")
+
+        output = self.read_output_json()
+        evt = output[0]
+
+        evt = self.clean_fields(evt)
+        self.assert_fields_are_documented(evt)
+
+    def clean_fields(self, evt):
+        # TODO: How to handle lables and ports?
```
andrewkroh: We can't validate the labels so think this handling is good. Also there's typo in the comment `s/lables/labels/`, but I think the comment can be removed.

`ports` is an array of objects and I don't think we have a way to document the fields of those objects, do we? We could document the field as `docker.container.ports.ip` for example and ignore the fact that `ports` is an array. Then I think we would just need to update the python code for `flatten_object` to step into arrays then the `assert_fields_are_documented` would properly validate those fields from the objects in the array.

metricbeat/module/docker/docker.go
```diff
@@ -50,6 +50,8 @@ func FetchStats(client *docker.Client) ([]DockerStat, error) {
 
 	containersList := []DockerStat{}
 	for _, container := range containers {
+		// This is currently very inefficient as docker calculates the average for each request,
+		// means each request will take at least 1s. Getting all stats at once is implemented here: https://github.com/docker/docker/pull/25361
 		containersList = append(containersList, exportContainerStats(client, &container))
```
andrewkroh: Any idea how long each request takes? Does each request block for some time period so that it can take two samples in order to calculate a delta or percentage?

metricbeat/tests/system/test_docker.py
```diff
@@ -0,0 +1,177 @@
...
+        proc.check_kill_and_wait()
+
+        # Ensure no errors or warnings exist in the log.
+        log = self.get_log()
+        self.assertNotRegexpMatches(log.replace("WARN EXPERIMENTAL", ""), "ERR|WARN")
+
+        output = self.read_output_json()
+        evt = output[0]
+
+        # TODO: How to handle lables and ports?
```
andrewkroh: `s/lables/labels/`

metricbeat/tests/system/test_docker.py
```diff
@@ -0,0 +1,177 @@
...
+        log = self.get_log()
+        self.assertNotRegexpMatches(log.replace("WARN EXPERIMENTAL", ""), "ERR|WARN")
+
+        output = self.read_output_json()
+        evt = output[0]
+
+        evt = self.clean_fields(evt)
+        self.assert_fields_are_documented(evt)
+
+    def clean_fields(self, evt):
```
ruflin: renamed

metricbeat/tests/system/test_docker.py
ruflin: I prefer to skip this for the moment as I wanted to look into how we can perhaps change the ports structure. But I agree, having something like described above could not only be useful here.

metricbeat/module/docker/docker.go
ruflin: Each request takes at least 2 seconds: https://github.com/docker/docker/blob/master/cli/command/container/stats_helpers.go#L148 I'm planing to open an issue in Docker to discuss if there could be flag that raw flags are shipped. There is also https://github.com/docker/docker/pull/25361 which will be useful to fetch all stats for the containers at once.

PR #2972, opened 2016-11-09
More Packetbeat cleanups

urso:
- unexport private symbols in protocol analyzer and some other modules in packetbeat
- remove some unused transaction fields in protocols analyzers

### DIFF
--- a/packetbeat/beater/packetbeat.go
+++ b/packetbeat/beater/packetbeat.go
@@ -26,11 +26,11 @@ import (
 )
 
 // Beater object. Contains all objects needed to run the beat
-type Packetbeat struct {
-	Config      config.Config
-	CmdLineArgs CmdLineArgs
-	Pub         *publish.PacketbeatPublisher
-	Sniff       *sniffer.SnifferSetup
+type packetbeat struct {
+	config      config.Config
+	cmdLineArgs flags
+	pub         *publish.PacketbeatPublisher
+	sniff       *sniffer.SnifferSetup
 
 	services []interface {
 		Start()
@@ -38,36 +38,36 @@ type Packetbeat struct {
 	}
 }
 
-type CmdLineArgs struct {
-	File         *string
-	Loop         *int
-	OneAtAtime   *bool
-	TopSpeed     *bool
-	Dumpfile     *string
-	WaitShutdown *int
+type flags struct {
+	file         *string
+	loop         *int
+	oneAtAtime   *bool
+	topSpeed     *bool
+	dumpfile     *string
+	waitShutdown *int
 }
 
-var cmdLineArgs CmdLineArgs
+var cmdLineArgs flags
 
 func init() {
-	cmdLineArgs = CmdLineArgs{
-		File:         flag.String("I", "", "Read packet data from specified file"),
-		Loop:         flag.Int("l", 1, "Loop file. 0 - loop forever"),
-		OneAtAtime:   flag.Bool("O", false, "Read packets one at a time (press Enter)"),
-		TopSpeed:     flag.Bool("t", false, "Read packets as fast as possible, without sleeping"),
-		Dumpfile:     flag.String("dump", "", "Write all captured packets to this libpcap file"),
-		WaitShutdown: flag.Int("waitstop", 0, "Additional seconds to wait before shutting down"),
+	cmdLineArgs = flags{
+		file:         flag.String("I", "", "Read packet data from specified file"),
+		loop:         flag.Int("l", 1, "Loop file. 0 - loop forever"),
+		oneAtAtime:   flag.Bool("O", false, "Read packets one at a time (press Enter)"),
+		topSpeed:     flag.Bool("t", false, "Read packets as fast as possible, without sleeping"),
+		dumpfile:     flag.String("dump", "", "Write all captured packets to this libpcap file"),
+		waitShutdown: flag.Int("waitstop", 0, "Additional seconds to wait before shutting down"),
 	}
 }
 
 func New(b *beat.Beat, rawConfig *common.Config) (beat.Beater, error) {
 	config := config.Config{
 		Interfaces: config.InterfacesConfig{
-			File:       *cmdLineArgs.File,
-			Loop:       *cmdLineArgs.Loop,
-			TopSpeed:   *cmdLineArgs.TopSpeed,
-			OneAtATime: *cmdLineArgs.OneAtAtime,
-			Dumpfile:   *cmdLineArgs.Dumpfile,
+			File:       *cmdLineArgs.file,
+			Loop:       *cmdLineArgs.loop,
+			TopSpeed:   *cmdLineArgs.topSpeed,
+			OneAtATime: *cmdLineArgs.oneAtAtime,
+			Dumpfile:   *cmdLineArgs.dumpfile,
 		},
 	}
 	err := rawConfig.Unpack(&config)
@@ -76,9 +76,9 @@ func New(b *beat.Beat, rawConfig *common.Config) (beat.Beater, error) {
 		return nil, err
 	}
 
-	pb := &Packetbeat{
-		Config:      config,
-		CmdLineArgs: cmdLineArgs,
+	pb := &packetbeat{
+		config:      config,
+		cmdLineArgs: cmdLineArgs,
 	}
 	err = pb.init(b)
 	if err != nil {
@@ -89,9 +89,9 @@ func New(b *beat.Beat, rawConfig *common.Config) (beat.Beater, error) {
 }
 
 // init packetbeat components
-func (pb *Packetbeat) init(b *beat.Beat) error {
+func (pb *packetbeat) init(b *beat.Beat) error {
 
-	cfg := &pb.Config
+	cfg := &pb.config
 	err := procs.ProcWatcher.Init(cfg.Procs)
 	if err != nil {
 		logp.Critical(err.Error())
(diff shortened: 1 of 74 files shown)


ruflin:
@urso Can you check the test failure before I start reviewing?

tsg:
jenkins, test it

dliappis:
jenkins, test it

dliappis:
jenkins, test it

dliappis:
jenkins, test it

dliappis:
jenkins, test it

dliappis:
jenkins, test it

packetbeat/beater/packetbeat.go
```diff
@@ -26,48 +26,48 @@ import (
 )
 
 // Beater object. Contains all objects needed to run the beat
-type Packetbeat struct {
-	Config      config.Config
-	CmdLineArgs CmdLineArgs
-	Pub         *publish.PacketbeatPublisher
-	Sniff       *sniffer.SnifferSetup
+type packetbeat struct {
+	config      config.Config
```
ruflin: Not sure about this change as Packetbeat can also be used as Library and it would be nice to at least have access tot he Config options. I sometimes have that problem with Golang libraries that most things are private and cannot be directly access which leads to coping code.

packetbeat/protos/amqp/amqp.go
```diff
@@ -346,44 +343,42 @@ func (amqp *Amqp) expireTransaction(trans *AmqpTransaction) {
...
 
 	if trans == nil {
-		trans = &AmqpTransaction{Type: "amqp", tuple: tuple}
-		amqp.transactions.Put(client.TCPTuple.Hashable(), trans)
+		trans = &amqpTransaction{tuple: tuple}
+		amqp.transactions.Put(client.tcpTuple.Hashable(), trans)
 	}
 
-	trans.ts = client.Ts
-	trans.Ts = client.Ts.UnixNano() / 1000
```
ruflin: What happened to this and the following line? Already part of client.ts?

packetbeat/protos/amqp/amqp.go
```diff
@@ -393,54 +388,52 @@ func (amqp *Amqp) handlePublishing(client *AmqpMessage) {
...
 
 	if trans == nil {
-		trans = &AmqpTransaction{Type: "amqp", tuple: tuple}
-		amqp.transactions.Put(server.TCPTuple.Hashable(), trans)
+		trans = &amqpTransaction{tuple: tuple}
+		amqp.transactions.Put(server.tcpTuple.Hashable(), trans)
 	}
 
-	trans.ts = server.Ts
-	trans.Ts = server.Ts.UnixNano() / 1000
```
ruflin: See above

packetbeat/protos/amqp/amqp_parser.go
```diff
@@ -83,8 +83,7 @@ func readFrameHeader(data []byte) (ret *AmqpFrame, err bool) {
 		logp.Warn("Missing frame end octet in frame, discarding it")
 		return nil, true
 	}
-	frame.Type = data[0]
-	frame.channel = binary.BigEndian.Uint16(data[1:3])
```
ruflin: What happened to this line?

packetbeat/protos/amqp/amqp_structs.go
```diff
@@ -162,66 +166,60 @@ const (
...
-	Data [2]*AmqpStream
+	data [2]*amqpStream
 }
 
-type AmqpFrame struct {
-	Type    byte
-	channel uint16
+type amqpFrame struct {
+	Type frameType
+	// channel uint16  (frame channel is currently ignored)
```
ruflin: Should we remove it then?

packetbeat/protos/amqp/amqp_structs.go
```diff
@@ -162,66 +166,60 @@ const (
...
+	body     []byte
+	bodySize uint64
 
-	Notes []string
+	notes []string
 }
 
 // represent a stream of data to be parsed
-type AmqpStream struct {
-	tcptuple *common.TCPTuple
```
ruflin: Was not used?

packetbeat/protos/amqp/amqp_structs.go
```diff
@@ -162,66 +166,60 @@ const (
...
 
 // contains the result of parsing
-type AmqpTransaction struct {
-	Type  string
+type amqpTransaction struct {
 	tuple common.TCPTuple
-	Src   common.Endpoint
-	Dst   common.Endpoint
-	Ts    int64
-	JsTs  time.Time
```
ruflin: Were these 2 not used?

packetbeat/protos/mongodb/mongodb.go
```diff
@@ -251,74 +251,65 @@ func (mongodb *Mongodb) onResponse(conn *mongodbConnectionData, msg *mongodbMess
...
-		trans.tuple = requ.TCPTuple
-
-		trans.Mongodb = common.MapStr{}
+		trans.mongodb = common.MapStr{}
 		trans.event = requ.event
 		trans.method = requ.method
 
-		trans.cmdline = requ.CmdlineTuple
-		trans.ts = requ.Ts
-		trans.Ts = int64(trans.ts.UnixNano() / 1000) // transactions have microseconds resolution
```
ruflin: As this becomes a pattern, I'm quite sure it is set as a default somewhere ;-)

packetbeat/protos/mongodb/mongodb.go
```diff
@@ -251,74 +251,65 @@ func (mongodb *Mongodb) onResponse(conn *mongodbConnectionData, msg *mongodbMess
...
 		}
 		trans.params = requ.params
 		trans.resource = requ.resource
-		trans.BytesIn = requ.messageLength
+		trans.bytesIn = requ.messageLength
 	}
 
 	// fill response
 	if resp != nil {
-		if requ == nil {
```
ruflin: Not needed anymore?

packetbeat/protos/mysql/mysql.go
```diff
@@ -489,34 +484,33 @@ func (mysql *Mysql) Parse(pkt *protos.Packet, tcptuple *common.TCPTuple,
 		}
 	}
 
-	if priv.Data[dir] == nil {
-		priv.Data[dir] = &MysqlStream{
-			tcptuple: tcptuple,
```
ruflin: This also becomes a pattern :-)

packetbeat/protos/mysql/mysql.go
```diff
@@ -560,62 +554,60 @@ func (mysql *Mysql) GapInStream(tcptuple *common.TCPTuple, dir uint8,
...
 			unmatchedRequests.Add(1)
 		}
 	} else {
-		trans = &MysqlTransaction{Type: "mysql", tuple: tuple}
+		trans = &mysqlTransaction{tuple: tuple}
 		mysql.transactions.Put(tuple.Hashable(), trans)
 	}
 
-	trans.ts = msg.Ts
-	trans.Ts = int64(trans.ts.UnixNano() / 1000) // transactions have microseconds resolution
```
ruflin: pattern again

packetbeat/protos/pgsql/pgsql.go
```diff
@@ -15,78 +15,92 @@ import (
...
 }
 
-type PgsqlTransaction struct {
-	Type         string
+type pgsqlTransaction struct {
 	tuple        common.TCPTuple
-	Src          common.Endpoint
-	Dst          common.Endpoint
-	ResponseTime int32
-	Ts           int64
```
ruflin: One more pattern

packetbeat/beater/packetbeat.go
urso: Well, packetbeat can be used as framework, not as a library. The New function for creating an instances is still available. The packetbeat internals are internal now, so no one messes with them when using packetbeat as framework.

Insteadt of using type beater.Packetbeat, one uses the type beat.Beater as returned by New method.

packetbeat/protos/amqp/amqp.go
urso: The fields removed have never been read from. They are not part of the event being published or anything => save some bytes and drop unused structure fields.

packetbeat/protos/amqp/amqp_parser.go
urso: same, remove unused fields.

urso: gotta check removal of `frame.Type` though.

packetbeat/protos/amqp/amqp_structs.go
urso: I commented it out because: a) it indirectly reflects the wire format. b) maybe in future this will be exposed?

packetbeat/protos/amqp/amqp_structs.go
urso: yep, not being used.

packetbeat/protos/amqp/amqp_structs.go
urso: TS and JsTs have not been used.

packetbeat/protos/mongodb/mongodb.go
urso: these fields have been set, but never been included in the event being generated (otherwise compilation would fail).

packetbeat/protos/mongodb/mongodb.go
urso: the value being set has been remove from trans, as field trans.tuple was unused.

packetbeat/protos/amqp/amqp_parser.go
urso: double checking, frame.Type was not remove. Got confused by diff.

packetbeat/protos/mysql/mysql.go
urso: yep, unused field

packetbeat/protos/mysql/mysql.go
urso: unused fields

packetbeat/protos/pgsql/pgsql.go
urso: unused fields

urso:
No logic has been changed in this PR. PR did rename a load of structures/fields/methods unexporting mostly private symbols. Plus only unused fields in protocol analyzer plugins have been removed (saves some bytes + I don't have to wonder about 3 timestamp values anymore). Often these fields have been initialised, but once initialized never be used again. No reads have had to be fixed after removing fields. Before removing I also double-checked with [guru](https://godoc.org/golang.org/x/tools/cmd/guru) for all potential references to removed fields.

> This changes makes most of the varialbes which are not used outside the package private. I'm not 100% sure I agree with this change. The reasoning is that I see packetbeat not only as a "tool" but also as a library. For example if someone builds a new protocol but could use some methods from Http, that was possible before but now because they are private he can't. To change it a contribution is necessary which normally leads to people just copy code over. I'm ok with the changes but I think we should be careful here and think more specific which methods should be exposed.

I see packetbeat more as a framework than a library. As I already noted, packetbeat can still be used this way. Also the sniffer, decoder and process watcher are still re-usable. The public API has just been restricted to the most essential use (for good I think).

About the HTTP example, I think it's good it discourages people from importing internals from http plugin. If required, I'd rather see common parser support being extracted/made available in general using a third package instead of importing symbols from/between other plugins.

> Is there a specific reasons you renamed MongoDb to MongodbPlugin etc? Was there an issue with the previous naming? Not sure if this addition is needed in the name.

renamed it for dealing with naming clashes

Issue #2941, opened 2016-11-04
multiple mappings of the same name appearing in the dropdown for Packetbeak index pattern. 


rasroh:
- Version: 5.0 GA
- Steps to Reproduce: 
     -In Kibana and ES  5.0 GA version,  multiple mappings of the same name appear when I choose `packetbeat` index pattern in `Kibana` Example : `@timestamp`, `@dns.answers.data` Not sure why and this looks very confusing to the user. 

@Tsg thinks that this might be a bug in Template Generation. 

Take a look at the gist with the mappings here: https://gist.github.com/rasroh/ed08025456a51faa3e6ca96a1a107c35

<img width="1224" alt="screen shot 2016-11-04 at 10 08 24 am" src="https://cloud.githubusercontent.com/assets/11466284/20017243/f42cb07e-a27f-11e6-87f2-4e862d1b230d.png">
<img width="431" alt="screen shot 2016-11-04 at 10 36 02 am" src="https://cloud.githubusercontent.com/assets/11466284/20017244/f42d327e-a27f-11e6-92ca-da45cfe42a4d.png">

cc/ @LeeDr @tsg

rasroh:
FYI : Once the packetbeat index pattern is refreshed, the dupes are not seen.

PR #2973, opened 2016-11-09
Remove duplicated fields from the fields.yml

monicasarbu:
Trying to fix https://github.com/elastic/beats/issues/2941
Remove duplicated fields from the Packetbeat fields.yml

With this PR, I added in the script that generates the Kibana index pattern to print an error when there are duplicated field names in the generated `fields.yml`.


monicasarbu:
@ruflin All duplicates are successfully removed for Packetbeat. Thanks!

ruflin:
LGTM. 

@monicasarbu Thanks a lot for cleaning this up.

PR #2975, opened 2016-11-09
Allow to use the "make testsuite" in projects vendoring beats

7AC:
Allows to import `vendor/github.com/elastic/beats/libbeat/scripts/Makefile` in projects that only vendor beats.

### DIFF
--- a/libbeat/scripts/Makefile
+++ b/libbeat/scripts/Makefile
@@ -35,6 +35,8 @@ ES_HOST?="elasticsearch"
 PWD=$(shell pwd)
 BUILD_DIR?=$(shell pwd)/build
 COVERAGE_DIR=${BUILD_DIR}/coverage
+COVERAGE_TOOL=${GOPATH}/bin/gotestcover
+COVERAGE_TOOL_REPO=github.com/elastic/beats/vendor/github.com/pierrre/gotestcover
 PROCESSES?= 4
 TIMEOUT?= 90
 TEST_ENVIRONMENT?=false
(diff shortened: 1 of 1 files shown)


ruflin:
By "only vendoring beats" you refer to that all vendor directories inside the vendored beats directory are removed?

7AC:
Hi @ruflin, I mean if you're testing a project that has `beats` in its `vendor/` directory. An example would be: https://github.com/aristanetworks/openconfigbeat/blob/master/.travis.yml (driven by https://github.com/aristanetworks/openconfigbeat/blob/master/Makefile). When we test that repo we rely on the sticky version in `vendor/github.com/elastic/beats` hence don't have `$GOPATH/src/github.com/beats`.

The `Makefile` I'm proposing to change here does:

```
go install github.com/elastic/beats/vendor/github.com/pierrre/gotestcover
```

which I think works only if you have that path already checked out in your `$GOPATH`. Please let me know if I'm misusing any of this.

Thanks!

7AC:
Actually it can be further simplified to just run `go get`: if the package is there (i.e., `beats` is already in `GOPATH`) it will just install it, otherwise get and install.

ruflin:
@7AC You are right and this is unfortunately broken :-( We never had that problem on our side as we always have the beats repo around. CI did not detect this, as there is also the beats repo around :-(

The reason we moved gotestcover into beats and do not use go get is because the repo is deprecated and for CI we don't want to have on more dependency in case for example Github does not respond (which happens more often then you expect). An alternative solution could be:

```
go install ${ES_BEATS}/vendor/github.com/pierrre/gotestcover
```

ES_BEATS should always point to beats directory. Now looking at it I see that it is a relative path so this will probably not work well with `go install`. An alternative is to use `BEAT_DIR/BEAT_NAME` for community beats. Unfortunately we are not 100% consistent yet, so this will work for community beats but not beats itself.

I have to play around with it to see which one will work. I can look into it later or beginning next week, perhaps you have some more ideas?

ruflin:
@7AC Yes, as all our beats are inside `$GOPATH/src/github.com/elastic` this will always be checked out for our CI tests. You are right that `go get` will also work 👍 

Assuming that gotestcover will potentially disappear in the near future, my preferred solution would be that also for community beats it will just work out of the box (no download needed). That is the reason for my previous suggestions.

7AC:
I see. According to @pierrre the repo will stay active:

https://github.com/pierrre/gotestcover#deprecated

So I guess the options are:
- use the `go test` equivalent provided by @pierrre
- do nothing for now and temporarily do the `go get` in my Travis config
- use `go get`

I already have the workaround in Travis in place, so feel free to close this if you prefer to keep your solution. Thanks for looking into it!

ruflin:
@7AC Merging as is as it is already an improvement and does not change the current behaviour. We can still improve it at a later stage. Thanks for the contribution.

PR #2977, opened 2016-11-10
Kafka consumer metricset

ruflin:
Depends on https://github.com/elastic/beats/pull/2969


ruflin:
@urso @andrewkroh Closing this PR as in the meantime we made quite a few changes to the kafka module and this has become out of sync. There is some code already inside for consumer which could be useful to be reused for a consumer module, but I think it will not happen as part of this PR so closing it.

PR #2982, opened 2016-11-10
Remove duplicates from fields.yml 

monicasarbu:
Remove duplicates from `fields.yml` in Heartbeat.


heartbeat/docs/fields.asciidoc
```diff
@@ -128,20 +128,6 @@ None
 
 
 [float]
-=== @timestamp
-
-type: date
-
-example: 2015-01-24 14:06:05.071000
-
-format: YYYY-MM-DDTHH:MM:SS.milliZ
```
ruflin: @urso Was there a reason you specified the format here instead of just using date?

monicasarbu: I think the format is used for generating the docs.

ruflin: @monicasarbu It also has an affect on the template. See https://github.com/elastic/beats/pull/2982/files#diff-7f554d6821a921aaa8e96f5d07d08093L3

PR #2985, opened 2016-11-11
Use fields.common.yml for libbeat

ruflin:
* Rename fields.yml to fields.common.yml
* Add generated file also for libbeat


ruflin:
Switched this to in progress again as the change gets more complex because system tests depend on the generated `fields.generated.yml` but this is only generated for the libbeat tests and not the packetbeat tests for example. Tests outside libbeat should be changed to not require the fields from libbeat.

ruflin:
Using now `fields.common.yml` for testing which is sufficient. This is now ready for review.

PR #2992, opened 2016-11-14
Fix changelog for 5.0.1/5.0.0

tsg:
The #2818 fix actually got in 5.0.0, so move the changelog entry around.


tsg:
Doc change only, but to make the test work we need to rebase on top of #2993.

PR #2996, opened 2016-11-14
Replace cookiecutter to generate beats by internal script

ruflin:
Currently cookiecutter is needed to generate a beat. This adds an additional dependency to generate a beat and does not give us a lot of control over which checks should be executed during execution. As our template is quite simple it can be done with basic string replacements. A simple python script is now done for the task which leads to less inputs needed by the user and allows us to do checks like GOPATH setup etc.

The script could potentially be done in Golang but the setup of python is needed anyways for development.

Further changes:

* Rename .go files to .go.tmpl so they are not checked for correct syntax
* Update documentation

NOTES:
* Changes for the metricbeat generator will follow in a second PR.


ruflin:
Blog posts where updated to fetch changes from the 5.1 branch. This is now ready to move forward.

PR #3001, opened 2016-11-14
Add raw mysql fields to the mysql event

ruflin:
This will include all mysql stats into the event under `mysql.stats.raw`. All keys will change to lower case.

Discussions:
* What should be namespace for these fields? `raw`, `extended`
* Should we normalise the keys to lower case?
* Should we replace `_` by `.`?
  * This can cause namespace conflicts if `a_b` and `a_b_c` has a value (happens in mysql)
  * Also we could hit index.mapping.depth.limit (default 20)
* Should this be experimental and available for all metricsets?
* What are we doing about the dynamic filed limit? -> disable it? https://github.com/elastic/elasticsearch/issues/11443 (default 1000)

Potential improvements:
* Remove duplicates which are already part of the event

TODO:
* [ ] Update docs
* [ ] Update changelog

### DIFF
--- a/metricbeat/mb/mb.go
+++ b/metricbeat/mb/mb.go
@@ -103,6 +103,10 @@ func (b *BaseMetricSet) Host() string {
 // Configuration types
 
 // ModuleConfig is the base configuration data for all Modules.
+//
+// The Raw config option is used to enable raw fields in a metricset. This means
+// the metricset fetches not only the predefined fields but add alls raw data under
+// the raw namespace to the event.
 type ModuleConfig struct {
 	Hosts      []string                `config:"hosts"`
 	Period     time.Duration           `config:"period"     validate:"positive"`
@@ -111,6 +115,7 @@ type ModuleConfig struct {
 	MetricSets []string                `config:"metricsets" validate:"required"`
 	Enabled    bool                    `config:"enabled"`
 	Filters    processors.PluginConfig `config:"filters"`
+	Raw        bool                    `config:"raw"`
 
 	common.EventMetadata `config:",inline"` // Fields and tags to add to events.
 }
--- a/metricbeat/module/mysql/status/status.go
+++ b/metricbeat/module/mysql/status/status.go
@@ -63,7 +63,7 @@ func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
 }
 
 // Fetch fetches status messages from a mysql host.
-func (m *MetricSet) Fetch() (event common.MapStr, err error) {
+func (m *MetricSet) Fetch() (common.MapStr, error) {
 	if m.db == nil {
 		var err error
 		m.db, err = mysql.NewDB(m.dsn)
@@ -77,7 +77,12 @@ func (m *MetricSet) Fetch() (event common.MapStr, err error) {
 		return nil, err
 	}
 
-	return eventMapping(status), nil
+	event := eventMapping(status)
+
+	if m.Module().Config().Raw {
+		event["raw"] = rawEventMapping(status)
+	}
+	return event, nil
 }
 
 // loadStatus loads all status entries from the given database into an array.
--- a/metricbeat/module/mysql/status/status_integration_test.go
+++ b/metricbeat/module/mysql/status/status_integration_test.go
@@ -13,7 +13,7 @@ import (
 )
 
 func TestFetch(t *testing.T) {
-	f := mbtest.NewEventFetcher(t, getConfig())
+	f := mbtest.NewEventFetcher(t, getConfig(false))
 	event, err := f.Fetch()
 	if !assert.NoError(t, err) {
 		t.FailNow()
@@ -34,8 +34,32 @@ func TestFetch(t *testing.T) {
 	assert.True(t, openStreams == 0)
 }
 
+func TestFetchRaw(t *testing.T) {
+	f := mbtest.NewEventFetcher(t, getConfig(true))
+	event, err := f.Fetch()
+	if !assert.NoError(t, err) {
+		t.FailNow()
+	}
+
+	t.Logf("%s/%s event: %+v", f.Module().Name(), f.Name(), event)
+
+	// Check event fields
+	cachedThreads := event["threads"].(common.MapStr)["cached"].(int64)
+	assert.True(t, cachedThreads >= 0)
+
+	rawData := event["raw"].(common.MapStr)
+
+	// Make sure field was removed from raw fields as in schema
+	_, exists := rawData["Threads_cached"]
+	assert.False(t, exists)
+
+	// Check a raw field if it is available
+	_, exists = rawData["Slow_launch_threads"]
+	assert.True(t, exists)
+}
+
 func TestData(t *testing.T) {
-	f := mbtest.NewEventFetcher(t, getConfig())
+	f := mbtest.NewEventFetcher(t, getConfig(false))
 
 	err := mbtest.WriteEvent(f, t)
 	if err != nil {
@@ -43,10 +67,11 @@ func TestData(t *testing.T) {
 	}
 }
 
-func getConfig() map[string]interface{} {
+func getConfig(raw bool) map[string]interface{} {
 	return map[string]interface{}{
 		"module":     "mysql",
 		"metricsets": []string{"status"},
 		"hosts":      []string{mysql.GetMySQLEnvDSN()},
+		"raw":        raw,
 	}
 }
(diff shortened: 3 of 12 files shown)


tsg:
+1 for `raw` and I think we should stick to the original names as much as possible (since this is the "raw" section). That means keeping `_` as they are, which also avoids the transformation issues.

I guess it would be nicer to have this in it's own metricset.

ruflin:
@tsg I did some tests with key names in elasticsearch and also spaces and lots of special characters just work. That means also for MongoDB we can use the original keys. In case one of the services uses dots we then have to decide if these should be replaced by _ to prevent potential conflicts.

I would prefer to have the "raw" fields inside the same metricset as it is like an extension of the metricset itself and only one config option is need to enable the additional fields. Overtime I think more and more fields will be moving from the raw to the structured document.

Implementation wise I would probably make `raw` a default config option that is available in each metricset, but it is up to the MetricSet to implement the additional fields.

andrewkroh:
Adding both the curated and raw fields seems like it creates a lot of duplicated data. As a user I think I would only want to store one copy of the data. Maybe we could remove fields that are in the schema from the raw fields. This means the module can send data for fields that are unknown a priori, and we can still provide an optimized experience for the fields that are known (like normalizing storage units to bytes so they render nicely in Kibana or using scaled_floats in mappings for percentages).

+1 for lower-casing raw fields

ruflin:
Agree that we should not duplicate fields. It would be nice if we could use apply.Schema() to remove the fields which were actually found so we would only iterate over the existing fields afterwards.

tsg:
I was trying to think through a likely user scenario:

1. user tries Metricbeat with default settings. She is generally happy.
2. But one day, she discovers that a metric that is exported by the service is not present in Metricbeat
3. Looks up a solution, learns about the `raw` option. Enables it.
4. Happy about the one metric she wanted, but not happy about the other 435 metrics added 😢 

To me it sounds like we should not care about duplication but rather allow to cherry pick raw metrics. I'm not sure how to do that from the usability point of view, especially how to find out which  metrics are available.

CC @brandonmensing

brandonmensing:
I think part of the joy of Metricbeat is that it's easy so I'm with you on the sadness of that user scenario!

I think this usage pattern might not be the most common so maybe a phased approach here would be good. Maybe we do 1 today, 2 later? 

1. Just grab the whole thing. Having the options is still better than not.
2. Allow the user to specify a list of extra fields to pick up. User has to look it up in the docs or a 3rd party reference to figure out what those fields are. 
3. When we learn more about how much anyone uses this, maybe we'll migrate extra fields into the main metricset and/or come up with a better way to allow them to configure extra fields. Maybe we develop an extra metricset of 'advanced' fields.

ruflin:
@tsg I think your use case will describe it pretty well. Our "standard" metricset should cover 90-95% of the uses cases. More experience users will look for additional fields. Based on the unhappiness in point 4 I hope 5 happens: Open Github issue or PR to add the missing metric. The hard part will be where do we say "no" and not add it to the base metricset?

If someone only wants to add one metric I would hope we can solve this with filters:

`Drop event if (mysql.status.info.raw.* AND NOT mysql.status.info.raw.your_metric)`

Otherwise a whitelist approach as proposed by @brandonmensing could also be interesting. But both approaches should probably only be added in a second step when we see how / if people are going to use it.

metricbeat/mb/mb.go
```diff
@@ -57,6 +57,7 @@ type MetricSet interface {
 	Host() string   // Host returns a hostname or other module specific value
 	// that identifies a specific host or service instance from which to collect
 	// metrics.
+	Raw() bool // Raw defines if raw metrics should be sent
```
andrewkroh: I'd like to avoid adding new methods to this interface were possible. In this case, when you need this information it is available in the MetricSet by `m.Module().Config().Raw`.

metricbeat/mb/mb.go
```diff
@@ -111,6 +118,7 @@ type ModuleConfig struct {
 	MetricSets []string                `config:"metricsets" validate:"required"`
 	Enabled    bool                    `config:"enabled"`
 	Filters    processors.PluginConfig `config:"filters"`
+	Raw        bool                    `config:"raw"`
```
andrewkroh: It would be nice to also add a godoc comment explaining the meaning of this field.

metricbeat/module/mysql/status/status_integration_test.go
```diff
@@ -34,19 +34,44 @@ func TestFetch(t *testing.T) {
...
+	cachedThreads := event["threads"].(common.MapStr)["cached"].(int64)
+	assert.True(t, cachedThreads >= 0)
+
+	rawData := event["raw"].(common.MapStr)
+
+	// Make sure field was removed from raw fields as in schema
+	_, exists := rawData["Threads_cached"]
+	assert.False(t, exists)
+
+	// Check random oder raw field if available
```
andrewkroh: `oder`?

metricbeat/mb/mb.go
andrewkroh: It would be nice to also add a godoc comment explaining the meaning of this field.

metricbeat/schema/schema_test.go
```diff
@@ -37,6 +37,21 @@ func TestSchema(t *testing.T) {
...
+			"test_a": Conv{Key: "testA", Func: nop},
+			"test_b": Conv{Key: "testB", Func: nop},
+		},
+	}
+
+	assert.True(t, schema.HasKey("test"))
+	assert.True(t, schema.HasKey("test_b"))
+	assert.True(t, schema.HasKey("test_a"))
+	assert.False(t, schema.HasKey("nooo"))
+}
```
andrewkroh: I don't think `HasKey` should return true for both the "source" keys and the "output" keys. Here's the behavior I expected which returns true only for the "source" keys.

```
func TestHasKey(t *testing.T) {
    schema := Schema{
        "test": Conv{Key: "Test", Func: nop},
        "test_obj": Object{
            "test_a": Conv{Key: "TestA", Func: nop},
            "test_b": Conv{Key: "TestB", Func: nop},
        },  
    }   

    assert.True(t, schema.HasKey("Test"))
    assert.True(t, schema.HasKey("TestA"))
    assert.True(t, schema.HasKey("TestB"))
    assert.False(t, schema.HasKey("test"))
    assert.False(t, schema.HasKey("test_obj"))
    assert.False(t, schema.HasKey("test_a"))
    assert.False(t, schema.HasKey("test_b"))
    assert.False(t, schema.HasKey("other"))
}
```

which would mean that `hasKey()` would be:

```
func hasKey(key string, mappers map[string]Mapper) bool {
	for _, mapper := range mappers {
		if mapper.HasKey(key) {
			return true
		}
	}
	return false
}
```

metricbeat/mb/mb.go
ruflin: agree, changed.

metricbeat/mb/mb.go
ruflin: done

metricbeat/module/mysql/status/status_integration_test.go
ruflin: Changed to `Check a raw field if it is available`

metricbeat/schema/schema_test.go
ruflin: obviously you are right -> wrong test leads to wrong implementation 👎

ruflin:
@andrewkroh New version pushed. The current config option is generic. Even though I currently only documented it in the status metricset as this is the only place available. As soon as we add it to other metricsets, we should have a general documentation.

metricbeat/module/mysql/status/status.go
```diff
@@ -32,6 +32,7 @@ type MetricSet struct {
 	mb.BaseMetricSet
 	dsn string
 	db  *sql.DB
+	raw bool
```
andrewkroh: Is this unused?

metricbeat/mb/mb.go
```diff
@@ -81,6 +81,7 @@ type BaseMetricSet struct {
 	name   string
 	module Module
 	host   string
+	raw    bool
```
andrewkroh: Is this unused?

metricbeat/module/mysql/status/status.go
ruflin: yes

metricbeat/mb/mb.go
ruflin: yes

PR #3003, opened 2016-11-15
HAProxy improvements

ruflin:
Restructure info and stats metricset

* [x] Review all fields for naming
* [x] Update docs with new fields

I didn't add dashboards yet. When we add dashboard it might get some more insights if there are some values we need / don't need. That is the reason I kept it experimental.


ruflin:
jenkins, retest it

metricbeat/module/haproxy/info/info.go
```diff
@@ -19,40 +19,30 @@ var (
...
 	if err := mb.Registry.AddMetricSet("haproxy", "info", New); err != nil {
 		panic(err)
 	}
 }
 
-// MetricSet type defines all fields of the MetricSet
-// As a minimum it must inherit the mb.BaseMetricSet fields, but can be extended with
-// additional entries. These variables can be used to persist data or configuration between
-// multiple fetch calls.
+// MetricSet for haproxy info
```
andrewkroh: Please 🙏  end the godocs with periods. Then no one has to wonder if you forgot to write the end of the sentence. 😼

ruflin: done

PR #2999, opened 2016-11-14
Metricbeat MongoDB module improvements

tsg:
Planning the following TODOs:

* [x] mark fields that are no longer present in 3.4 as optional. #2990
* [x] add wiredTiger (main) metrics
* [x] set up integration tests with mongo 3.4

PR #3007, opened 2016-11-15
Added a MongoDB sample dashboard

tsg:
A few of the widgets assume the WiredTiger engine. They can be easily
replaced if other engines are used.

screen shot 2016-11-15 at 14 39 54

Kind of depends on #2999 for the data and the index pattern, but doesn't strictly require a rebase.

cc @ctindel


ruflin:
LGTM. As far as I understand if the old engine is used, parts of the dashboard will stay empty?

tsg:
@ruflin That's right, a couple of widgets will report no data.

tsg:
jenkins, retest it

PR #3009, opened 2016-11-15
Remove index pattern from git as these are auto generated

ruflin:


tsg:
`make update` creates them, right? Does this mean we need to make sure `make update` is called during packaging?

ruflin:
@tsg Yes. I wrongly assumed this is already the case, but I could only find it in the set_version command: https://github.com/elastic/beats/blob/master/dev-tools/set_version#L35

We should add it probably here: 
https://github.com/elastic/beats/blob/master/dev-tools/deploy#L16

ruflin:
Update was fixed in https://github.com/elastic/beats/pull/3014 Rebased now on master and running build to see if nothing breaks.

ruflin:
jenkins, build it

PR #3010, opened 2016-11-15
Improve "pipelines" documentation

Shugyousha:
Mention the similarity to the "indices" array option and add an example.

I am not that familiar with the configuration of beats so please double-check the example.

### DIFF
--- a/libbeat/docs/outputconfig.asciidoc
+++ b/libbeat/docs/outputconfig.asciidoc
@@ -196,10 +196,35 @@ For more information, see <<configuring-ingest-node>>.
 
 ===== pipelines
 
-Array of pipeline selector configurations supporting conditionals, format string
-based field access and name mappings. The first rule matching will be used to
-set the `pipeline` for the event to be published. If `pipelines` is missing or no
-rule matches, the `pipeline` field will be used.
+Similar to the `indices` array, this is an array of pipeline selector
+configurations supporting conditionals, format string based field access
+and name mappings. The first rule matching will be used to set the
+`pipeline` for the event to be published. If `pipelines` is missing or
+no rule matches, the `pipeline` field will be used.
+
+Example elasticsearch output with `pipelines`:
+
+["source","yaml"]
+------------------------------------------------------------------------------
+filebeat.prospectors:
+- paths: ["/var/log/app/normal/*.log"]
+  fields:
+    type: "normal"
+- paths: ["/var/log/app/critical/*.log"]
+  fields:
+    type: "critical"
+
+output.elasticsearch:
+  hosts: ["http://localhost:9200"]
+  index: "filebeat-%{+yyyy.MM.dd}"
+  pipelines:
+    - pipeline: critical_pipeline
+      when.equals:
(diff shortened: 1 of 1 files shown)


Shugyousha:
I have signed the CLA now. Let me know if I should open another PR or if we can still use this one.

libbeat/docs/outputconfig.asciidoc
```diff
@@ -196,10 +196,27 @@ For more information, see <<configuring-ingest-node>>.
...
+`pipeline` for the event to be published. If `pipelines` is missing or
+no rule matches, the `pipeline` field will be used.
+
+Example elasticsearch output with `pipelines`:
+
+["source","yaml"]
+------------------------------------------------------------------------------
+output.elasticsearch:
+  hosts: ["http://localhost:9200"]
+  index: "logs-%{+yyyy.MM.dd}"
```
ruflin: Could we use filebeat- here as it is the default?

libbeat/docs/outputconfig.asciidoc
```diff
@@ -196,10 +196,27 @@ For more information, see <<configuring-ingest-node>>.
...
+  pipelines:
+    - pipeline: critical_pipeline
+      when.contains:
+        type: "critical"
```
ruflin: Should we perhaps add the top part of the config file to this example to make it obvious where type critical or normal are coming from?

libbeat/docs/outputconfig.asciidoc
Shugyousha: ok, I will change it to

```index: "filebeat-%{+yyyy.MM.dd}"```

If you prefer I can remove the formatted date field as well.

File: libbeat/docs/outputconfig.asciidoc (hunk 2 above)
Comment (Shugyousha) [inline libbeat/docs/outputconfig.asciidoc]: I am actually not completely sure where this field can be declared. I assume it comes from the fields declaration in the prospector section so a complete example (which I agree we should add) should look something like:

```
filebeat.prospectors:
- paths: ["/var/log/app/normal/*.log"]
  fields:
    type: "normal"
- paths: ["/var/log/app/critical/*.log"]
  fields:
    type: "critical"

output.elasticsearch:
  hosts: ["http://localhost:9200"]
  index: "filebeat-%{+yyyy.MM.dd}"
  pipelines:
    - pipeline: critical_pipeline
      when.contains:
        type: "critical"
    - pipeline: normal_pipeline
      when.contains:
        type: "normal"
```

Instead of ```when.contains`` using ```when.equals``` would probably be better here. Does that look okay?

dedemorton:
LGTM. Thanks for adding this. I plan to add more content around specifying pipelines, but this is a good start!

Shugyousha:
Ok, then I will force update this PR as soon as I find the time.

Shugyousha:
I pushed an improved version but CI seems to have failed due to an unrelated issue.

ruflin:
@Shugyousha @dedemorton Test failure is related to a flaky test and not this PR. Feel free to merge.

PR #3016, opened 2016-11-16
Run make update for packaging in docker container

ruflin:


tsg:
jenkins, package it

tsg:
The problem is this one is that now we have some makefile commands run in docker and some outside of docker, so that is likely to cause permission issues. If we go this route, we should probably create some sort of a "build" docker image, and run as much as possible inside it.

tsg:
Btw, the virtualenv issue is solved in the stack build, so that's kind of a non-issue at this point.

ruflin:
Closing as solved otherwise.

PR #3021, opened 2016-11-16
Fix path when generating imports

tsg:
The script used to make the assumption that we are under GOPATH.
This changes it to be based on `${BEAT_DIR}` which we have already
set in the Makefile, and is less brittle.

Should also fix the unified build snapshots, which don't checkout beats under a gopath.


tsg:
jenkins, package it

tsg:
Package tests are green: http://build-eu-00.elastic.co/job/beats-package-PR/44/

andrewkroh:
Looks like I merged this prematurely, there are some [errors](https://travis-ci.org/elastic/beats/jobs/176535288) in the generator.

metricbeat/Makefile
```diff
@@ -51,7 +51,7 @@ configs: python-env
 .PHONY: imports
 imports:
 	mkdir -p include
-	python ${ES_BEATS}/metricbeat/scripts/generate_imports.py > include/list.go
+	python ${ES_BEATS}/metricbeat/scripts/generate_imports.py ${BEAT_DIR} > include/list.go
```
ruflin: The reason this doesn't work, is that BEAT_DIR has different meaning for community beats. We need an if / else to check if it is a community beat or not ...

ruflin:
@andrewkroh @tsg Here is a fix for the broken community beats: https://github.com/elastic/beats/pull/3021

PR #3024, opened 2016-11-17
Improves Docker Labels handling and output

rikatz:
This PR improves Docker Label handling, by not turning it as an array of JSONs, but instead a set of different fields on a map string.

This is useful specially if you're building dashboards with Kibana and wants to filter based on container labels.

This is one of the items of issue #2629 

Hope it helps :smile:

### DIFF
--- a/metricbeat/module/docker/helper.go
+++ b/metricbeat/module/docker/helper.go
@@ -34,7 +34,7 @@ func NewContainer(container *docker.APIContainers) *Container {
 	return &Container{
 		Id:     container.ID,
 		Name:   ExtractContainerName(container.Names),
-		Labels: ConvertContainerLabels(container.Labels),
+		Labels: DeDotLabels(container.Labels),
 	}
 }
 
@@ -51,7 +51,9 @@ func ExtractContainerName(names []string) string {
 	return strings.Trim(output, "/")
 }
 
-func ConvertContainerLabels(labels map[string]string) common.MapStr {
+// DeDotLabels returns a new common.MapStr containing a copy of the labels
+// where the dots in each label name have been changed to an underscore.
+func DeDotLabels(labels map[string]string) common.MapStr {
 
 	outputLabels := common.MapStr{}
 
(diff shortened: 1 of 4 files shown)


rikatz:
Signed the CLA

karmi:
Hi @rikatz, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/beats/pull/3024.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

rikatz:
@karmi Ops, my bad! Added my email to my profile. Do I need to change anything else?

Thanks!

ruflin:
jenkins, test it

ruflin:
@rikatz LGTM. Thanks for the contribution.

metricbeat/module/docker/helper.go
```diff
@@ -52,24 +51,15 @@ func ExtractContainerName(names []string) string {
 	return strings.Trim(output, "/")
 }
 
-func BuildLabelArray(labels map[string]string) []common.MapStr {
+func BuildLabelArray(labels map[string]string) common.MapStr {
```
andrewkroh: I think the name of this method should change now.

andrew-moldovan: I think you meant @andrewkroh :)

andrewkroh: Looking at the what the function does, it seems to de-dot the label names. So how about `DeDotLabels()`? I would prefer that all exported functions have godoc accompanying them. So how about adding (assuming that's the name you go with):

```
// DeDotLabels returns a new common.MapStr containing a copy of the labels
// where the dots in each label name have been changed to an underscore.
```

rikatz: OK! Just did a commit right now with the name ConvertContainerLabels, will rename them again to DeDotLabels (as they are directly related to Labels), and also add the comments!

rikatz: @andrewkroh Did as suggested. Now let's wait the CI tasks complete :smile:

Hope this can help!

PR #3026, opened 2016-11-17
Add ability to have a custom name for json data

f-ld:


filebeat/docs/reference/configuration/filebeat-options.asciidoc
```diff
@@ -328,6 +329,8 @@ json.add_error_key: true
 and its value must be string, otherwise it is ignored. If no text key is defined, the line
 filtering and multiline features cannot be used.
 
+*`json_key`*:: name of the JSON key in the output document where the decoded JSON has to be placed. The default is "json"
```
andrewkroh: The name stutters in the config as it is `json.json_key`. I think it should be called `target` which is more consistent with Logstash's [json filter](https://www.elastic.co/guide/en/logstash/current/plugins-filters-json.html#plugins-filters-json-target).

andrewkroh: This is missing a period at the end.

filebeat/docs/reference/configuration/filebeat-options.asciidoc
```diff
@@ -319,6 +319,7 @@ Example configuration:
 [source,yaml]
 -------------------------------------------------------------------------------------
 json.message_key: log
+json.message_key: "json"
```
andrewkroh: `s/message_key/json_key/`

To be consistent with the line above which does not use quotes around `log`, drop the quotes from `"json"`.

filebeat/input/event.go
```diff
@@ -105,6 +105,10 @@ func mergeJSONFields(e *Event, event common.MapStr) {
 			}
 		}
 	} else {
-		event["json"] = e.JSONFields
+		key := e.JSONConfig.JsonKey
+		if len(key) == 0 {
+			key = "json"
```
andrewkroh: It would be nice if this check could be done once at initialization and then set the value to the default if it's empty. The usual pattern is to initialize the `JsonKey` value to its default value prior to unpacking the user's config; maybe that can be done in this case.

filebeat/harvester/reader/json_config.go
```diff
@@ -2,6 +2,7 @@ package reader
 
 type JSONConfig struct {
 	MessageKey    string `config:"message_key"`
+	JsonKey       string `config:"json_key"`
```
andrewkroh: `s/JsonKey/JSONKey/` https://github.com/golang/go/wiki/CodeReviewComments#initialisms

filebeat/docs/reference/configuration/filebeat-options.asciidoc
andrewkroh: `s/has to/will/`. I think it will read better with this change. WDYT?

ruflin:
@f-ld Is there a specific reason you need to do this on the beats side and can't do it in logstash or eleasticsearch ingest? From my perspective that would be the more appropriate place to do it.

filebeat/docs/reference/configuration/filebeat-options.asciidoc
f-ld: Ok, makes sense, I will apply those changes

f-ld:
@ruflin Yes, there is a reason: we are using filebeat to push to kafka directly, so it would mean have that conversion in all our kafka consumers

ruflin:
@f-ld What are you using for consuming? LS?

f-ld:
@ruflin 
We have many consumers so far:
- Confluent's Kafka ES connector
- Spark Streaming
- pinterest/secor (push to Amazon S3)
- MemSQL pipelines
And we are planning to use other connectors

f-ld:
@andrewkroh I made the requested changes.
About the default value for `json.target`, it is now set before unpacking user config, which is cleaner. But it has a consequence on unit tests not using that part and instantiating the `JSONConfig struct` directly: they have to specify that default value each time. Check last commit

f-ld:
Note: no idea why the checks failed: it seems independent of my changes. Is there a way to trigger them again?

f-ld:
Giving up on this PR, will review latest filebeat version and recreate the PR if required

PR #3028, opened 2016-11-17
Update Github issue template with address for reporting security vulnerabilities

andrewkroh:


ruflin:
@andrewkroh Perhaps instead of a bullet list, we could have a checklist for the steps. I saw that in the homebrew repo and it is quite nice.

PR #3029, opened 2016-11-17
Switch partition metricset from client to broker

ruflin:


metricbeat/module/kafka/partition/partition.go
```diff
@@ -37,57 +36,60 @@ func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
 // Fetch partition stats list from kafka
 func (m *MetricSet) Fetch() ([]common.MapStr, error) {
 
-	if m.client == nil {
+	if m.broker == nil {
```
urso: consider allocating/creating m.broker at metricset initialization and  `m.broker.Connected()` to see if broker is connected and reconnect? As `Connected` only checks if a connection object exists, you can try Heartbeat.

Alternatively (to not keep connection open for too long), use `defer m.broker.Close()` and always call `Open`.

metricbeat/module/kafka/partition/partition.go
```diff
@@ -37,57 +36,60 @@ func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
...
+		err := broker.Open(config)
 		if err != nil {
 			return nil, err
 		}
-		m.client = client
+		m.broker = broker
 	}
 
-	topics, err := m.client.Topics()
+	response, err := m.broker.GetMetadata(&sarama.MetadataRequest{})
```
urso: how about optional topic names configuration? e.g. `metricset.kafka.partitions.topics=[...]` with default being empty string. Would this work here?

Question: does the call to GetMetadata return the complete cluster-state or only topics available to the broker it's ID?

metricbeat/module/kafka/partition/partition.go
```diff
@@ -37,57 +36,60 @@ func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
...
-			}
+			offsetRequest := &sarama.OffsetRequest{}
+			// Get 2 offsets, assumption is newest first in array, oldest second
+			offsetRequest.AddBlock(topic.Name, partition.ID, sarama.OffsetNewest, 2)
 
-			broker, err := m.client.Leader(topic, partition)
-			if err != nil {
-				logp.Err("Fetching brocker for partition %s in topic %s: %s", partition, topic, err)
-			}
+			offsetResponse, _ := m.broker.GetAvailableOffsets(offsetRequest)
```
urso: Does GetAvailableOffsets add the current broker ID to the API request?

Checking the ['encode' method](https://github.com/Shopify/sarama/blob/master/offset_request.go#L29), it seems ReplicaID is always -1. Does it mean, we query offsets from leader?

metricbeat/module/kafka/partition/partition.go
```diff
@@ -37,57 +36,60 @@ func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
...
-				"topic":     topic,
-				"partition": partition,
+				"topic": common.MapStr{
+					"name":  topic.Name,
+					"error": topic.Err,
+				},
+				"partition": common.MapStr{
+					"id":       partition.ID,
+					"error":    partition.Err,
+					"leader":   partition.Leader,
```
urso: is `partition.leader == broker.id` if current broker is leader? If so, we wan't to add a 'leader' flag somewhere in the document?

metricbeat/module/kafka/partition/partition.go
ruflin: I moved the broker part to New(). Undecided which approach I should follow for the open / close part :-)

metricbeat/module/kafka/partition/partition.go
ruflin: Topic filter "should" work here but I would keep this also for a version 2. It is already possible to filter topics out, but you would need `filters` which I think is ok for now.

For the question: TBH I'm not sure. Based on https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-TopicMetadataRequest I would expect that we would get an array for each broker. But we only get a list of topics back.

metricbeat/module/kafka/partition/partition.go
ruflin: Looks like broker id is not set. Do we actually query the same information as with client? :-(

metricbeat/module/kafka/partition/partition.go
ruflin: Yes, or at least that is my assumption. I don't hink we need to add a special "boolean" to the event as a query with script filter should be easy to get the above result.

glide.yaml
```diff
@@ -62,7 +62,7 @@ import:
 - package: github.com/miekg/dns
   version: 5d001d020961ae1c184f9f8152fdc73810481677
 - package: github.com/Shopify/sarama
-  version: fix/sasl-handshake
+  version: enh/offset-replica-id
```
tsg: Just checking, does this branch contain the sasl fix?

metricbeat/module/kafka/partition/partition.go
```diff
@@ -19,75 +21,196 @@ func init() {
...
+	addr := b.Addr()
+	for _, other := range meta.Brokers {
+		if other.Addr() == addr {
+			m.id = other.ID()
+			break
+		}
+	}
+
+	if m.id == noID {
+		b.Close()
```
tsg: If `b.Close()` is needed here, maybe it's need on the other error branches as well?

glide.yaml
urso: yes, the branch is directly branched off from sasl-handshake.

metricbeat/module/kafka/partition/partition.go
urso: only close `b` if we return `nil`. That is line 68 a close is missing.

metricbeat/module/kafka/partition/partition.go
```diff
@@ -19,75 +21,196 @@ func init() {
...
+	if m.id != noID {
+		return b, nil
+	}
 
-	return &MetricSet{BaseMetricSet: base}, nil
+	// current broker is bootstrap only. Get metadata to find id:
+	meta, err := b.GetMetadata(&sarama.MetadataRequest{})
+	if err != nil {
+		return nil, err
+	}
```
urso: this call might fail if leader-election is active.

metricbeat/module/kafka/partition/partition.go
```diff
@@ -19,75 +21,196 @@ func init() {
...
+		cfg:           cfg,
+		id:            noID,
+	}, nil
+}
+
+func (m *MetricSet) connect() (*sarama.Broker, error) {
+	b := m.broker
+	if err := b.Open(m.cfg); err != nil {
+		return nil, err
+	}
```
urso: will open fail if leader-election is active?

metricbeat/module/kafka/partition/partition.go
```diff
@@ -19,75 +21,196 @@ func init() {
...
-		}
-		m.client = client
+	b, err := m.connect()
+	if err != nil {
+		return nil, err
 	}
 
-	topics, err := m.client.Topics()
+	defer b.Close()
+	response, err := b.GetMetadata(&sarama.MetadataRequest{})
```
urso: might fail if leader-election is active? consider retry (max 3 times) with waits of a few milliseconds (see sarama settings for defaults)

metricbeat/module/kafka/partition/_meta/fields.yml
```diff
@@ -16,14 +16,53 @@
...
+          type: long
+          description: >
+            Replica id (broker).
+
+        - name: insync_replica
+          type: boolean
+          description: >
+            Indicates if replica is included in the in-sync replicate set (ISR).
+
+        - name: error
```
ruflin: Hm, if we switch this one to error_code, we would have to change the other 2 errors too. WDYT?

metricbeat/module/kafka/partition/partition.go
ruflin: I added the close to line 68

metricbeat/module/kafka/partition/partition.go
ruflin: We use the defaults which are:

```
c.Metadata.Retry.Max = 3
c.Metadata.Retry.Backoff = 250 * time.Millisecond
```

I think we are good with these defaults.

metricbeat/module/kafka/partition/partition.go
ruflin: See default sarama settings below. As we do the connect for each call I think it can be ok if one of them fails.

metricbeat/module/kafka/partition/partition.go
ruflin: Not sure. But my thinking here is similar to my comment below. It is ok if it fails from time to time, as we call connect() for each Fetch. I prefer that it fails instead of waiting for a long time.

urso: well, waiting is for a few ms. in case we fail, a document will be missing for one period.

metricbeat/module/kafka/partition/partition.go
urso: well, these defaults are 'internal' to sarama. If you use client/(a)sync-publisher, sarama updates metadata from time to time. With us calling b.GetMetadata directly these settings are not in effect I think.

metricbeat/module/kafka/partition/_meta/fields.yml
tsg: `error_code` seems a bit more natural to me, but I don't feel to strong about it.

urso:
Doing some tests with master + this PR:

- connect to local broker only, instead of full cluster
  - just one error message in logs if broker is not reachable, with error message including correct IO error (used to be loads of warnings from sarama)
- fixes issue with metricbeat reporting offsets at `-1` if partition is owned by another (potentially unreachable) broker. Change includes reporting topics/partitions only if current broker is leader (limitation by kafka API)
- fix oldest partition offset being off, by explicitly querying for oldest available offset

tsg:
jenkins, retest it

metricbeat/_meta/beat.full.yml
```diff
@@ -94,6 +94,27 @@ metricbeat.modules:
   #period: 10s
   #hosts: ["localhost:9092"]
 
+  #client_id: metricbeat
+
+  #metadata.retries: 3
```
ruflin: Not sure if it we should use metadata namespace here, as in the end from a user perspective it is just retrieds. Metadata is like an internal "logic" thing.

metricbeat/module/kafka/partition/_meta/data.json
```diff
@@ -11,14 +11,19 @@
...
-                "newest": 13,
+                "newest": 11,
                 "oldest": 0
             },
-            "partition": 0,
-            "replicas": [
-                0
-            ],
-            "topic": "testtopic"
+            "partition": {
```
ruflin: This needs update. I will do a PR as soon as this is merged.

metricbeat/module/kafka/partition/partition.go
```diff
@@ -19,75 +25,292 @@ func init() {
...
-			logp.Err("Fetch partition info for topic %s: %s", topic, err)
+	evtBroker := common.MapStr{
+		"id":      m.id,
+		"address": b.Addr(),
+	}
+
+	for _, topic := range response.Topics {
+		evtTopic := common.MapStr{
+			"name": topic.Name,
+			"error": common.MapStr{
```
ruflin: As long as error is empty, we should not push it.

metricbeat/module/kafka/partition/partition.go
```diff
@@ -19,75 +25,292 @@ func init() {
...
+					continue
+				}
+
+				// create event
+				event := common.MapStr{
+					"topic":  evtTopic,
+					"broker": evtBroker,
+					"partition": common.MapStr{
+						"id": partition.ID,
+						"error": common.MapStr{
```
ruflin: Same here for the error: If it is empty, we should not add it to the event.

urso: values `error.code: 0` is a valid error code in kafka. That's why I did keep it in.

metricbeat/module/kafka/partition/partition.go
urso: error is not empty, but has `error.code: 0`.

ruflin: I'm now checking for error.code != 0 and if 0 ignore it.

metricbeat/module/kafka/partition/partition.go
ruflin: oh, it is a valid error code?!? :-( What is "no" error?

PR #3039, opened 2016-11-21
Refactor fields handling with readers

ruflin:
Each reader can add fields to the message object. The reader itself should always add data under its own namespace to prevent conflicts. All these fields are then added to the Data object. This will allow each reader in the future to add its own data if needed.

The JSON reader was simplified in the way that data by default is written under the `json` namespace. Now no special fields have to be passed for JSON and the processing can still happen on the event level.

No functional changes were made in this PR.

This PR was extracted from https://github.com/elastic/beats/pull/2279

### DIFF
--- a/filebeat/harvester/reader/json.go
+++ b/filebeat/harvester/reader/json.go
@@ -121,7 +121,7 @@ func (r *JSON) Next() (Message, error) {
 		return message, err
 	}
 
-	var fields = common.MapStr{}
+	var fields common.MapStr
 	message.Content, fields = r.decodeJSON(message.Content)
 	message.AddFields(common.MapStr{"json": fields})
 	return message, nil
--- a/filebeat/harvester/reader/json.go
+++ b/filebeat/harvester/reader/json.go
@@ -120,6 +120,9 @@ func (r *JSON) Next() (Message, error) {
 	if err != nil {
 		return message, err
 	}
-	message.Content, message.Fields = r.decodeJSON(message.Content)
+
+	var fields = common.MapStr{}
+	message.Content, fields = r.decodeJSON(message.Content)
+	message.AddFields(common.MapStr{"json": fields})
 	return message, nil
 }
--- a/filebeat/input/event.go
+++ b/filebeat/input/event.go
@@ -20,7 +20,7 @@ type Event struct {
 	Text         *string
 	JSONConfig   *reader.JSONConfig
 	State        file.State
-	Data         common.MapStr
+	Data         common.MapStr // Use in readers to add data to the event
 }
 
 func NewEvent(state file.State) *Event {
--- a/filebeat/input/event.go
+++ b/filebeat/input/event.go
@@ -18,9 +18,9 @@ type Event struct {
 	DocumentType string
 	Bytes        int
 	Text         *string
-	JSONFields   common.MapStr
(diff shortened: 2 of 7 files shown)


filebeat/harvester/reader/json.go
```diff
@@ -120,6 +120,9 @@ func (r *JSON) Next() (Message, error) {
 	if err != nil {
 		return message, err
 	}
-	message.Content, message.Fields = r.decodeJSON(message.Content)
+
+	var fields = common.MapStr{}
```
urso: prefer `var fields common.MapStr` to have a `nil` map by default.

The line as is creates a new map to discard it on next line again.

filebeat/harvester/reader/json.go
```diff
@@ -120,6 +120,9 @@ func (r *JSON) Next() (Message, error) {
...
+	message.Content, fields = r.decodeJSON(message.Content)
+	message.AddFields(common.MapStr{"json": fields})
```
urso: what if fields is empty?

filebeat/input/event.go
```diff
@@ -18,9 +18,9 @@ type Event struct {
 	DocumentType string
 	Bytes        int
 	Text         *string
-	JSONFields   common.MapStr
 	JSONConfig   *reader.JSONConfig
 	State        file.State
+	Data         common.MapStr
```
urso: why rename to `Data` instead of `Fields`? `Data` sounds quite generic, we need a doc comment saying what data is for?

filebeat/input/event.go
```diff
@@ -39,8 +39,19 @@ func (e *Event) ToMapStr() common.MapStr {
...
+	// Add data fields which are added by the readers
+	for key, value := range e.Data {
+		event[key] = value
+	}
+
+	// Check if json fields exist
+	var jsonFields common.MapStr
+	if fields, ok := event["json"]; ok {
+		jsonFields = fields.(common.MapStr)
+	}
```
urso: why not have `Data` and `JsonFields` (or `JSON`) in message type? This handling of json by unpacking from event + delete key seems a little contrived.

filebeat/harvester/reader/json.go
ruflin: agree, changed.

filebeat/harvester/reader/json.go
ruflin: I think if fields is nil probably something went wrong. But we also didn't catch this previously and I don't intend to do "logic" changes in this PR. 

Probably decodeJSON should be able to return an error.

filebeat/input/event.go
ruflin: I didn't want to confuse "fields" which are set from the user with the "fields / data" added by the readers. I will add a doc comment.

filebeat/input/event.go
ruflin: I want to have the Event to be generic and not specifically having JSON inside. Remember the origin of this PR is when adding data for the multiline reader. So instead of adding more and more variable to event, data is the generic container for it. 

Should we cleanup JSON? Yes, definitively. I even think the processing should probably not happen in the event package.

urso: +1 for not doing the json in event package.

Issue #2987, opened 2016-11-11
Mysql and Postgres modules should allow addition of custom queries


ajaybhatnagar:
Features Request:

Database monitoring modules of Mysql/Postgres should allow adding additional sql commands / queries to collect additional db performance data to customize dashboards in addition to outputs from standard commands.

Would like to see Cassandra module included .

Thanks
Ajay

ruflin:
Can you add some examples for the metrics / queries you would run?

ajaybhatnagar:
For example current mysql module collects data only from show status command. If one wants to use show global status, then there are additional parameters included. Similarly if performance stats are enabled, there are many more data available for monitoring and plotting. There one can enable selective data by specifying queries of interest in the metricbeat mysql module configuration, it can become a powerful tool for monitoring and analyzing db performance.

In general, in my view, for any monitoring tool, it should be possible to define a set of commands to be used to collect data from remote node, map the output data fields to corresponding json document for ES storage. That way one need not build new beats but use modular approach to enhance / modify data collection from the remote source, and setup it's field mapping for storing the document. Thus a generic beat module can be reused and reconfigured for multiple uses for different types of monitoring data collection.

ruflin:
I would split up this enhancement request in three parts:

* Provide all metrics which were retrieved from a stats API instead of just a filtered and revised subset. An idea for this see https://github.com/elastic/beats/pull/3001
* Allow free form of queries for fetching stats 
* Allow execution of external scripts to fetch stats

For the free form of queries, [httpbeat](https://github.com/christiangalsterer/httpbeat) comes to my mind. Also something that applies here is probably httpbeat which allows to query any http endpoint. It does not necessarely apply to all module types. An other one is [mysqlbeat](https://github.com/adibendahan/mysqlbeat). Is that the direction you were thinking?

Number 3 is the most powerful but also most complex one as it goes into external script execution. We decided to not touch this topic for at least the first version of metricbeat :-)

ajaybhatnagar:
Agree that subset of stats is not sufficient and full set of metrics be exported for any application in general. Since in case of Mysql (or Mariadb) sources of monitoring data comes from different tables (like performance tables, sys tables etc), data needed for having a meaningful monitoring or analyzing issues, would have to be in one place. That is where I think these beats can be very useful . This is where free form of queries would come handy to build custom monitoring on top of standard beats. This should not required "Recompile of code" but added / subtracted thru config files.

[...]

PR #3051, opened 2016-11-22
Add a generic jmx/jolokia metricset to metricbeat

vas78:
This is a generic metricset for fetching data from jmx using jolokia.


ruflin:
@vas78 Thanks a lot for this contribution.

So far we always predefined the data structure for each metricset in the fields.yml so can guarantee the mapping in elasticsearch is correct. With jmx this becomes a little bit tricky. Few questions as I'm not too familiar with the output coming from Jolokia / JMX:

* Is there a base structure / common fields that are always the same?
* Config is currently quite tricky to get it right. Do you see an option to simplify this?

vas78:
@ruflin I'm afraid that's exacly the point: it's up to beats user to decide which metrics he/she would like to collect with jolokia. I could image adding some common fields like uptime, heap size, java process cpu usage etc, but I`m not quite sure if it makes any sense here.

To the second question: I could try converting the current solution with an embedded json in mapping into an array of values like "metric:alias", e.g.:

    -
      host: localhost:4008
      mapping:
          "java.lang:type=Runtime:::Uptime":"uptime"
          "java.lang:type=Memory:::HeapMemoryUsage":"memory.heap_usage"
          "java.lang:type=Memory:::NonHeapMemoryUsage":"memory.non_heap_usage"
      application: application1
      instance: instance1

What do you think?

ruflin:
@vas78 Sorry for the late answer. Some more thoughts form my side:

We recently introduce `raw` fields in Metricbeat. This is for all event data we do not know the type of in advance or didn't create mapping for the index template. It would be nice if we could push some generic data for all events into a specific mapping and have all the other data under `raw` for the moment.

Looking at your events above it seems like that data for a specific MBean is known in advance. So what if we would split the above up in 2 parts:

* Generic JMX connector
* Specific MBean Metricsets

In the above case we would have for example:

```
modules:
- module: jolokia
  metricsets: ["runtime", "memory"]
```

In your test cases you also have Cassandra data. So the `cassandra` module would use the jmx connection as a library but I assume for Cassandra we know in advance what the structure of the events is.

The customer mapping definition of the jmx module goes in a similar direction as the request here: https://github.com/elastic/beats/issues/2987 There are still lots of discussion ongoing on how to make this happen best in a generic way. Inputs are more then welcome.

@vas78 Does the above assumptions make sense related to JMX?

@radoondas Mentioning you here as I know you were also looking in JMX in the past.

vas78:
@ruflin Ok, makes sense. I will split up the module into some "standard" metricsets like memory, runtime, etc. and one "free-style" metricset that can be configured as desired.

ruflin:
@vas78 Great. Potentially you can split it up in 2 PR's because we will potentially have some additional discussions on the `free-style` part but that should not block the ` standard` part to get in.

ruflin:
@vas78 Here is a first idea on the example of mysql on how a "generic" metricset could work: https://github.com/elastic/beats/pull/3170

ruflin:
@vas78 Anything I could help here with? It would be really nice to get support for jolokia / jmx in metricbeat.

vas78:
@ruflin sorry, it`s taking a bit longer than I though especially due to all the deadlines just before holidays :( I will provide an update soon.

ruflin:
@vas78 No worries. Looking forward to the changes.

ruflin:
@vas78 Did you have any chance to work on this one recently?

vas78:
@ruflin I`ve made two attempts to refactor the code and split up the module into "standard" and "custom" part, but did not really like the result. Yes, there are some metrics in MBean that are always present in every JVM, but one of the most important of these "standard" metrics - GC - depends heavily on the GC strategy configured in the application. Hard coding all of these strategies makes not that much sense since we already have a dynamic metrics mapping that solves this problem. The rest of these "standard" metrics are just some CPU and threading usage stats that are not that valuable alone in this context.

Your proposal regarding the possible future use case for custom jolokia module is embedding it into other application specific modules, e.g. cassandra, kafka, etc. and hence providing a strict mapping for each application metric set. I still think it`ll be pretty hard to guess which metrics would the "common" user pick for each application and this could end up in the constant flow of requests to add this one and that one extra metric. So why not let the users decide and configure it for themselves from the very beginning? The only difficulty in that case is really just creating/adjusting the right index mapping in Elastic which is quite simple in that case (I had to do that multiple times daily in the very beginning for every application we currently monitor with beats).

Having said all that, I would like to return to the original PR and focus on the dynamic jolokia module, rather than creating new static metricsets. Maybe I could in fact simplify the config and make it more user-friendly? Please let me know what you think.

ruflin:
@vas78 Thanks for pushing forward on this one. The reason for the fixed mapping is that we try to have a good out of the box experience whenever possible. But I see your point that with jolokia this can become quite tricky. So 👍  on moving forward with the initial proposal and figure out some better options for how to configure it. 

In the meantime we also have the prometheus/collector metricset wich builds the foundation of such a "dynamic" metricset: https://github.com/elastic/beats/tree/master/metricbeat/module/prometheus/collector The most important part here is that it can set the namespace: https://github.com/elastic/beats/blob/master/metricbeat/module/prometheus/collector/collector.go#L107 Could you update your PR with making the namespace possible? I will have a look again on your PR with the above in mind.

ruflin:
I looked at the configuration again and here are some ideas on what we could do:


JSON based:

```
- host: localhost:4008
  mapping: '{"java.lang:type=Runtime:::Uptime":"uptime","java.lang:type=GarbageCollector,name=ConcurrentMarkSweep:::CollectionTime":"gc.cms_collection_time","java.lang:type=GarbageCollector,name=ConcurrentMarkSweep:::CollectionCount":"gc.cms_collection_count","java.lang:type=GarbageCollector,name=ParNew:::CollectionTime":"gc.parnew_collection_time","java.lang:type=GarbageCollector,name=ParNew:::CollectionCount":"gc.parnew_collection_count","java.lang:type=Memory:::HeapMemoryUsage":"memory.heap_usage","java.lang:type=Memory:::NonHeapMemoryUsage":"memory.non_heap_usage"}'
```


One line per metric:
```
- host: localhost:4008
  mappings:
    - "java.lang:type=Runtime:::Uptime":"uptime"
    - "java.lang:type=GarbageCollector,name=ConcurrentMarkSweep:::CollectionTime":"gc.cms_collection_time"
    - "java.lang:type=GarbageCollector,name=ConcurrentMarkSweep:::CollectionCount":"gc.cms_collection_count"
    - "java.lang:type=GarbageCollector,name=ParNew:::CollectionTime":"gc.parnew_collection_time"
    - "java.lang:type=GarbageCollector,name=ParNew:::CollectionCount":"gc.parnew_collection_count"
    - "java.lang:type=Memory:::HeapMemoryUsage":"memory.heap_usage"
    - "java.lang:type=Memory:::NonHeapMemoryUsage":"memory.non_heap_usage"
```

Abstracted to YAML:	
```
- host: localhost:4008
  mappings:
    - type: Runtime
      attr: Uptime
      field: uptime
    - type: GarbageCollector
      name: ConcurrentMarkSweep
      attr: CollectionTime
      field: gc.cms_collection_time
    - type: GarbageCollector
      name: ConcurrentMarkSweep
      attr: CollectionCount
      field: gc.cms_collection_count
    - type: GarbageCollector
      name: ParNew
      attr: CollectionTime
      field: gc.parnew_collection_time
```

The above would allow, to potentially add the field type which would allow us to create a template out of it. So we could potentially create a template on the first run and push it to elasticsearch.

```
- host: localhost:4008
  mappings:
    - type: Runtime
      attr: Uptime
      field: uptime
      field_type: integer
```

My favorite option is currently the Abstract YAML one because I think it is the most extentable and it does not require the user to get the strings correctly, like 3 `:` etc.

ruflin:
@nfrankel Would be interesting to get your input on this one as you recently published https://blog.frankel.ch/feedback-on-feeding-spring-boot-metrics-to-elasticsearch/#gsc.tab=0 Which way would work best for you?

nfrankel:
@ruflin As a (lazy) configurer, my preferred option would be to copy-paste the `ObjectName` from the jconsole _e.g._ `java.lang:type=GarbageCollector,name=PS MarkSweep`.

Also, though I understand about the consequences of what @vas78 wrote, I still think that having standard metric sets could help tremendously. Their internal implementations can still be based on the same format that the free form metric set use anyway.

Note that though I read the thread, I'm neither a Go nor a JMX expert (like @rhuss - Hallo Kumpel !), so I might have overlooked something that seem evident to more advanced people.

ruflin:
@rhuss @nfrankel Thanks for taking a look at this. 

@nfrankel That seems your preferred option for a first try would be `one line per metric`. I'm happy to start with this option. @vas78 WDYT?

About the standard metrics: We can build that definitively on top of this dynamic fetcher. Do you have some suggestion on which metrics you would like to put together in a metricset?

vas78:
@ruflin Actually I liked the idea of the abstracted yaml since it eliminates the necessity to use some custom delimiter for mbean and attribute values that the "one line per metric" approach requires. And since we will still need not just the mbean object name itself but the attribute and it`s mapping as well, what would you say to this one?
>     -
>       host: localhost:4008
>       mapping:
>         - mbean: 'java.lang:type=Runtime'
>           attributes:
>             - attr: Uptime
>               field: uptime
>         - mbean: 'java.lang:type=GarbageCollector,name=ConcurrentMarkSweep'
>           attributes:
>             - attr: CollectionTime
>               field: gc.cms_collection_time
>             - attr: CollectionCount
>               field: gc.cms_collection_count
> 

This would allow lazy configurers like @nfrankel (and myself :)) to copy-paste the mbean name from jconsole on the one hand and group all attributes belonging to the the same mbean and thus improve config readability and maintainability on the other.
Later on we could also add an optional field_type element for dynamic template generation as originally proposed by @ruflin.

ruflin:
@vas78 SGTM. You guys are here the experts :-)

ruflin:
@vas78 It seems some generate files are out of date. Try to run `make update` and push the changes again.

vas78:
@ruflin Renamed the module and ran make update, but the "check" CI Target still fails, can you please take a look at it? Thanks!

ruflin:
@vas78 For the `check` build part, `make update` should solve your problem. Can you run it again and make sure you don't get any local diffs left?

For the other test failing. This could be old flaky one. Any chance that you could rebase on master? This should also resolved the conflicts above (files were removed in a recent PR).

metricbeat/_meta/beat.full.yml
```diff
@@ -121,6 +121,14 @@ metricbeat.modules:
   #period: 10s
   #hosts: ["tcp://127.0.0.1:14567"]
 
+#------------------------------- Jolokia Module ------------------------------
+- module: jolokia
+  metricsets: ["dynamic"]
```
ruflin: should we call it `jmx`? Not sure about the name :-)

metricbeat/_meta/beat.full.yml
```diff
@@ -121,6 +121,14 @@ metricbeat.modules:
...
+  enabled: true
+  period: 10s
+  hosts: ["localhost"]
+  namespace: "jolokia_metrics"
```
ruflin: lets call this `metrics` as `jolokia` will be already part of the namespace anyways.

metricbeat/docs/fields.asciidoc
```diff
@@ -2390,6 +2391,47 @@ type: integer
...
+
+
+[float]
+== dynamic Fields
+
+dynamic
+
+
+[float]
+=== jolokia.dynamic._namespace
```
ruflin: No need to document this one. This field will be removed and placed under `metricset.namespace`.

metricbeat/docs/fields.asciidoc
```diff
@@ -2390,6 +2391,47 @@ type: integer
...
+
+type: keyword
+
+Namespace
+
+[float]
```
ruflin: Do these 2 fields really exist in all events?

metricbeat/metricbeat.yml
```diff
@@ -46,7 +46,13 @@ metricbeat.modules:
   period: 10s
   processes: ['.*']
 
-
+#------------------------------- Jolokia Module ------------------------------
```
ruflin: If you use `short_config: false`in the module fields.yml it will not show up here. See https://github.com/elastic/beats/blob/master/metricbeat/module/kafka/_meta/fields.yml#L7

metricbeat/module/jolokia/_meta/config.yml
```diff
@@ -0,0 +1,7 @@
+- module: jolokia
```
ruflin: Currently we commend out the config by default so it is not enabled. This might change in the near future: https://github.com/elastic/beats/blob/master/metricbeat/module/kafka/_meta/config.yml

metricbeat/module/jolokia/dynamic/_meta/data.json
```diff
@@ -0,0 +1,36 @@
+{
+    "@timestamp": "2016-08-27T20:26:21.837Z",
+    "beat": {
+        "hostname": "seu",
+        "name": "seu"
+    },
+    "jolokia": {
+        "dynamic": {
+            "gc": {
```
ruflin: This needs to be updated / removed to not cause confusion.

metricbeat/module/jolokia/dynamic/data.go
```diff
@@ -0,0 +1,103 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("Cannot unmarshal json response: %s", err)
+	}
+
+	debugf("Unmarshalled json: ", entries)
+
+	event := map[string]interface{}{}
+
+	if application != "" {
+		event["application"] = application
```
ruflin: I see, this is really a predefined field. Except for putting it under a fixed namespace I don't have a good idea yet where we could put these and how to document it.

In docker we have a similar problem where several metricsets also have the container info. We use some tricks there to add this info the the evend and have it then under the container metricset. Perhaps we should also introduce such a namespace here. `jolokia.?.application`. Like this we can also document it as it will not be under the dynamic namespace. Some suggestions with what we should replace ?

metricbeat/module/jolokia/_meta/fields.yml
```diff
@@ -0,0 +1,10 @@
+- key: jolokia
+  title: "Jolokia"
+  description: >
+    Jolokia Module
+  fields:
```
ruflin: Can you add an experimental flag here? https://github.com/elastic/beats/blob/master/metricbeat/module/kafka/_meta/fields.yml#L6

metricbeat/module/jolokia/dynamic/jolokia.go
```diff
@@ -0,0 +1,98 @@
...
+	mb.BaseMetricSet
+	client          *http.Client      // HTTP client that is reused across requests
+	metricSetConfig []MetricSetConfig // array containing urls, bodies and mappings
+	namespace       string
+}
+
+// New create a new instance of the MetricSet
+func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
+	// Additional configuration options
+	config := struct {
```
ruflin: Please add an experimental message here: https://github.com/elastic/beats/blob/master/metricbeat/module/kafka/partition/partition.go#L40

metricbeat/_meta/beat.full.yml
ruflin: Ignore that and lets go for dynamic at the moment (except you like jmx better).

ruflin:
@vas78 I was thinking if we can perhaps help you with this PR to get it into metricbeat? Let me know if we should take over from here or if we can help you in any way.

vas78:
@ruflin Sorry, I was totally overloaded with some other projects lately and had no time to look into this one. Would be great if you can take it over from here, many thanks!

ruflin:
@vas78 Great, will take over from here and ping you in case I have some questions :-)

ruflin:
@vas78 Here is my WIP follow up PR: https://github.com/elastic/beats/pull/3570

ruflin:
@vas78 I'm currently work on the jmx module and discovered the following. We have `mappings` with an array of `mapping` under it in the config each with a `host`. This kind of breaks the module principle as the hosts are configured on the module level already. If someone needs different mapping types for different hosts, multiple modules should be defined. So I suggest to simplify the config as following:

```
  hosts: ["localhost:8778"]
  mapping:
    - mbean: 'java.lang:type=Runtime'
      attributes:
        - attr: Uptime
          field: uptime
    - mbean: 'java.lang:type=GarbageCollector,name=ConcurrentMarkSweep'
      attributes:
        - attr: CollectionTime
          field: gc.cms_collection_time
        - attr: CollectionCount
          field: gc.cms_collection_count
    - mbean: 'java.lang:type=Memory'
      attributes:
        - attr: HeapMemoryUsage
          field: memory.heap_usage
        - attr: NonHeapMemoryUsage
          field: memory.non_heap_usage
```
Do I miss something here that requires the additional `mappings`?

vas78:
@ruflin You got it all right, there`s no other reason to have `mappings`. No objections from my side, let`s keep the config simple.

ruflin:
@vas78 Thanks. An other question: I'm trying to get my head around the application and instance config and fields. What are these exactly used for? Could these also just go under `fields`?

vas78:
@ruflin These config items were useful in the first implementation with multiple hosts under `mappings`. This way it was possible to set an appropriate application and instance for each host and thus be able to distinguish between multiple apps/instances running on the same host but under different URLs (e.g. different Jolokia ports).
Now, since we decided to proceed with one host per config on the top level these fields can be indeed extracted and moved over to the standard `fields` section.

ruflin:
Thanks for the feedback. Will remove the fields then.

ruflin:
@vas78 https://github.com/elastic/beats/pull/3570 just got merged. Thanks for all the work you have done on this one. We are very happy to have jolokia / jmx support in Metricbeat. Closing this one.

ruflin:
@vas78 @nfrankel @rhuss As discussed previously, as a next step I would also like to have some structured metricsets for Jolokia. For this I opened the issue here: https://github.com/elastic/beats/issues/3585 But I will need your help with this one as I need your insights into which metrics exists for most installation. If you could provide me with some configurations for the jmx metricset that should become a Metricset on its own I would be more then happy to implement them. Lets continue the discussion in https://github.com/elastic/beats/issues/3585

Issue #2854, opened 2016-10-26
Feature request: Filebeat prospector configuration template


f-ld:
We are heavily starting to use filebeat (to push to kafka) and we have a configuration with 100+ prospectors (100+ kafka topics)
For each prospector we repeat some parameters like encoding, ignore_older, scan_frequency, ...

We think filebeat would benefit of either:
- a way to change the default value for all prospectors (for example the default value of max_bytes is currently 10MB (10485760). If I want to set 20MB for all my prospectors, I have to copy past the parameter for each of them. This works for people having many prospectors of the same kind
- a way to define templates: when creating a propsector, we can import values from a template or another. This works for people having many prospectors of a few kinds
- a way to copy parameters from a previously defined prospector

For now, we use puppet to create our configuration file so we use a loop in our ERB file and this works fine. The only problem is that the deployed configuration file is quite long and can be reduced with one of the options above

PS: first published here -> https://discuss.elastic.co/t/suggestion-filebeat-prospector-configuration-template/62687

ruflin:
@f-ld Thanks a lot for sharing all the details and pointing out the various options. I assigned the filebeat and libbeat label as if we think about this, we probably would want to have a generic solution that works for all beats. 

One thing of interest for me would be why you have that many different prospectors? Are these 100 different files from just one machine? Assuming you could use a "dynamic" field that is generated per event for your kafka topic selection, what that already be enough?

f-ld:
@ruflin Yes, the generic solution makes perfectly sense

About the high number of propsectors: we have a microservice architecture with roughtly 50+ different services (multiple instances of each, multiple datacenter, so 200+ services running). Each service has from 2 to 6-7 different log files (first one is http access logs of the service, next is error log. Then we have some "event" logs depending on the service).
And we are going forward service discovery with hashicorp consul, allowing any application to be deployed anywhere (so no static pre-defined set of hosts for one application), meaning we want to deploy the same filebeat configuration on each single host just in case we deploy service X on that host.
So we end up with 50+ prospectors for http access logs, 50+ access logs for errors, etc. and a parameter change for one is likely required for the 49+ others (as pointed, today, puppet is handling it for us). And we are grouping in the same kafka topic logs for the various instances of a same service, but we want to avoid mixing logs of service A with logs of service B, C, D ... even if the beat metadata would allow a kafka consumer to filter them. Just to keep topics not too big and avoid useless network traffic when one consumer is only interested by logs from 1 of the 50+ services.

[...]

PR #3073, opened 2016-11-29
configurable heartbeat defaults values

urso:
Introduce configurable heartbeat default settings for monitors. Add default values to `heartbeat.defaults.monitors` section. This helps with sharing common settings like DNS-client or service discovery settings.

### DIFF
--- a/heartbeat/_meta/beat.yml
+++ b/heartbeat/_meta/beat.yml
@@ -182,6 +182,23 @@ heartbeat.monitors:
     # Required response contents.
     #body:
 
+# Configure monitors default settings shared by all monitors.
+heartbeat.defaults.monitors:
+  # Enable/Disable monitor
+  #enabled: true
+
+  # Configure default task scheduler
+  #schedule: '@every 30s'
+
+  # Configure IP protocol types to ping on if hostnames are configured.
+  # Ping all resolvable IPs if `mode` is `all`, or only one IP if `mode` is `any`.
+  #ipv4: true
+  #ipv6: true
+  #mode: all
+
+  # Total running time per ping test.
+  #timeout: 16s
+
 heartbeat.scheduler:
   # Limit number of concurrent tasks executed by heartbeat. The task limit if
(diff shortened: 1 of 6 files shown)


heartbeat/_meta/beat.yml
```diff
@@ -182,6 +182,23 @@ heartbeat.monitors:
     # Required response contents.
     #body:
 
+# Configure monitors default settings shared by all monitors.
+heartbeat.defaults.monitors:
```
ruflin: Should this be commented out by default?

heartbeat/_meta/beat.yml
```diff
@@ -182,6 +182,23 @@ heartbeat.monitors:
...
+  #schedule: '@every 30s'
+
+  # Configure IP protocol types to ping on if hostnames are configured.
+  # Ping all resolvable IPs if `mode` is `all`, or only one IP if `mode` is `any`.
+  #ipv4: true
+  #ipv6: true
+  #mode: all
+
+  # Total running time per ping test.
+  #timeout: 16s
```
ruflin: Interesting default :-)

heartbeat/heartbeat.yml
```diff
@@ -182,6 +182,23 @@ heartbeat.monitors:
     # Required response contents.
     #body:
 
+# Configure monitors default settings shared by all monitors.
+heartbeat.defaults.monitors:
```
ruflin: I don't think this should be in the short config file

ruflin:
This goes in the direction of https://github.com/elastic/beats/issues/2854 . We should have a discussion if this is also the general approach we want to take across other beats.

heartbeat/_meta/beat.yml
urso: doesn't really matter if commented out or not. In either case it will be `nil` config.

heartbeat/_meta/beat.yml
urso: just copy'n pasted from the other modules.

PR #3081, opened 2016-11-30
Adds a couchbase module for metricbeat

paschdan:
This module has 3 metricsets to collect metrics from couchbase:

* cluster overview metrics
* node metrics
* bucket metrics


paschdan:
@ruflin I have a question regarding moving the data conversion to the schema method:
if i have a json input for example this:
```json
{
  "foo": "bar",
  "nested": {
    "foo": "nested_bar"
  }
}
```

and i want my result to be
```json
{
  "foo": "bar",
  "nested_foo": "nested_bar"
}
```

is this currently anyhow possible with schema / mapstrstr / mapstriface ?

ruflin:
@paschdan I remember we had recently the same question popping up but I don't remember if we found or implemented a solution for it. Perhaps @andrewkroh can help?

ruflin:
@paschdan Thanks a lot for contributing the couchbase module directly with 3 metricsets 🎉  Let me know when it is ready for a first review. For the field names, check out our naming conventions: https://www.elastic.co/guide/en/beats/libbeat/current/event-conventions.html

paschdan:
@ruflin i guess its ready for a review

metricbeat/module/couchbase/node/_meta/fields.yml
```diff
@@ -0,0 +1,97 @@
...
+  type: group
+  description: >
+    node
+  fields:
+    - name: hostname
+      type: keyword
+      description: >
+        The hostname of the Node
+    - name: uptime
+      type: keyword
```
ruflin: Should this be a long?

metricbeat/module/couchbase/node/_meta/fields.yml
```diff
@@ -0,0 +1,97 @@
...
+        Amount of memcached memory reserved (bytes)
+    - name: cmd_get
+      type: long
+      description: >
+        Number of get commands
+    - name: cpu_utilization_rate.pct
+      type: float
+      description: >
+        The CPU utilization rate (%)
+    - name: couch_docs_actual_disk_size
```
ruflin: couch_docs.actual_disk_size.bytes

metricbeat/module/couchbase/node/_meta/fields.yml
```diff
@@ -0,0 +1,97 @@
...
+      type: long
+      description: >
+        Amount of disk space used by Couch docs.(bytes)
+    - name: couch_docs_data_size.bytes
```
ruflin: couch_docs.data_size.bytes

metricbeat/module/couchbase/node/_meta/fields.yml
```diff
@@ -0,0 +1,97 @@
...
+      type: long
+      description: >
+        Data size of couch documents associated with a node (bytes)
+    - name: couch_spatial_data_size.bytes
```
ruflin: couch_spatial.data_size...

metricbeat/module/couchbase/node/_meta/fields.yml
```diff
@@ -0,0 +1,97 @@
...
+      type: long
+      description: >
+        couchSpatialDataSize field
+    - name: couch_spatial_disk_size.bytes
```
ruflin: couch_spatial.

I'm tempted to say even `couch.spatial.*` and apply the same to the above as there are different namespaces inside couch it seems.

metricbeat/module/couchbase/node/_meta/fields.yml
```diff
@@ -0,0 +1,97 @@
...
+      type: long
+      description: >
+        couchSpatialDiskSize field
+    - name: couch_views_actual_disk_size.bytes
+      type: long
+      description: >
+        Amount of disk space occupied by Couch views (bytes)
+    - name: couch_views_data_size.bytes
```
ruflin: See above

metricbeat/module/couchbase/node/_meta/fields.yml
```diff
@@ -0,0 +1,97 @@
...
+      type: long
+      description: >
+        Size of object data for Couch views (bytes)
+    - name: curr_items
```
ruflin: current_items.value

metricbeat/module/couchbase/node/_meta/fields.yml
```diff
@@ -0,0 +1,97 @@
...
+      type: long
+      description: >
+        Number of current items
+    - name: curr_items_tot
```
ruflin: current_items.total

metricbeat/module/couchbase/_meta/configure-node.sh
```diff
@@ -0,0 +1,38 @@
...
+    done
+
+    echo
+    echo >&2 "Failed to Start Couchbase"
+}
+
+waitForCouchbase
+
+# Setup index and memory quota
+curl -v -X POST http://127.0.0.1:8091/pools/default -d memoryQuota=300 -d indexMemoryQuota=300
```
paschdan: haha yes, me too, but unfortunately it is not possible by config, i could only set specific initial params via cli instead of rest, but loading a sample database is not possible via cli, as far as i know

metricbeat/module/couchbase/bucket/_meta/fields.yml
```diff
@@ -0,0 +1,49 @@
...
+      description: >
+        type of the bucket
+    - name: quota.ram
+      type: long
+      description: >
+        Amount of RAM used by the bucket (bytes)
+    - name: quota.raw_ram
+      type: long
+      description: >
+        Amount of raw RAM used by the bucket (bytes)
```
paschdan: so this would be correct:
```
- name: quota.ram.bytes
   format: bytes
   type: long
   description >
     Amount of RAM used by the bucket
```

ruflin: Yes, except that it seems `format` etc. has one space too much. It should be directly under name (probably copy / paste problem)

metricbeat/module/couchbase/cluster/data.go
```diff
@@ -0,0 +1,71 @@
...
+
+	var d Data
+	err := json.NewDecoder(body).Decode(&d)
+	if err != nil {
+		logp.Err("Error: ", err)
+	}
+
+	logp.Info("Printing Data:")
+	event := common.MapStr{
+		"quota.index_memory.mb":          d.IndexMemoryQuota,
```
ruflin: The reason it is possible do use dots in the fields.yml is because we do some "dedotting" magic in our python script. This is mainly to make writing the fields.yml easier. We don't have that in Golang. So yes, each dotted entry should end up in a nested document for compatibility reason with elasticsearch 2.x.

ruflin:
@paschdan Please ping me when I should have an other look. One thing we should also do is updating the CHANGELOG.asciidoc.

metricbeat/module/couchbase/bucket/_meta/fields.yml
paschdan: write on the fly problem ;)

paschdan:
@ruflin i fixed the bucket metricset, can you have a look on that? 
I will now go on with node and cluster metricsets.

paschdan:
@ruflin i updated the metricsets now to return nested metrics, only thing, that i not know is, if there is a format for "sec"

e.g. https://github.com/actano/beats/blob/29f20bdf6174af4684842879185e4348c7a5a859/metricbeat/module/couchbase/node/_meta/fields.yml#L95-L95

ruflin:
@paschdan We currently don't use any format for sec. But it is an interesting idea. We could use Duration and specific what input and output we would like to have as this exist in Kibana. @monicasarbu You could be interested in this.

metricbeat/module/couchbase/bucket/data.go
```diff
@@ -0,0 +1,69 @@
...
+	for _, Bucket := range d {
+		event := common.MapStr{
+			"name": Bucket.Name,
+			"type": Bucket.BucketType,
+			"data": common.MapStr{
+				"used.bytes": Bucket.BasicStats.DataUsed,
+			},
+			"disk": common.MapStr{
+				"fetches":    Bucket.BasicStats.DiskFetches,
+				"used.bytes": Bucket.BasicStats.DiskUsed,
```
ruflin: Unfortuantely for compatibility reasons also the `.bytes` and `.pct` have to be inside their own common.MapStr. I hope we can simplify this in the future similar to what we did on the fields.yml side.

ruflin:
@paschdan This will also need a rebase.

paschdan:
@ruflin added the changes for the format, but i am not sure how my entry in the changelog should look like.

ruflin:
jenkins, test it

paschdan:
flaky?

ruflin:
jenkins, retest it

ruflin:
@paschdan Yeah, zookeeper tests are sometimes flaky. Rerun the tests and all green. MERGED.
Again: Thanks a lot for the contribution. I'm really happy to have the couchbase module as part of metricbeat.

PR #3082, opened 2016-11-30
Pass `--always-copy` to virtualenv

phenomenes:
virtualenv creates symlinks so `make setup` fails when ran on a network mounted
fs. `--always-copy` forces virtualenv to copy files to the destination dir rather than symlinking.


phenomenes:
Signed the agreement 👍

ruflin:
@phenomenes I couldn't find the CLA in our database. Under which name / email did you sign it?

phenomenes:
Jimena Cabrera Notari

ruflin:
@phenomenes CLA now worked. Strangely it took some time until the agreement showed up in our tool :-(

PR #3094, opened 2016-12-01
Fix make package for community beats

ruflin:
gopkg.in needs to be copied from the vendor directory of libbeat in the vendor directory


ruflin:
jenkins, package it

PR #3106, opened 2016-12-02
Bump versions for 5.1

dedemorton:
Since this is a new branch and it won't be set as current until right before the release date, we can merge this anytime.


tsg:
LGTM.

Issue #2923, opened 2016-11-02
Document json_decode_fields processor


andrewkroh:
#2605 adds a processor for parsing JSON called `json_decode_fields`. We'll need to add documentation and put a note into the configuration file section for processors.

dedemorton:
Closed with #3110

PR #3110, opened 2016-12-03
Add doc for json_decode_fields processor

dedemorton:
I'm eager to get this processor documented, so I created a first draft of content for the Beats docs. I think this needs more detail. It's meant to be a starting point. I don't read Go very well, so I might have gotten some of the details wrong.

@suraj-soni I thought it might be easier for you to review this (rather than create the docs yourself) because our doc structure a little eccentric due to single sourcing some of the config content. 

In addition to adding more detail to the descriptions (if you think it's needed), it would be nice to have a real example that shows the message string, the config that you would use to decode it, and the JSON object that would result. Can someone help me with this?

This PR fixes issue #2923

### DIFF
--- a/filebeat/docs/filebeat-filtering.asciidoc
+++ b/filebeat/docs/filebeat-filtering.asciidoc
@@ -2,20 +2,31 @@
 == Filtering and Enhancing the Exported Data
 
 Your use case might require only a subset of the data exported by Filebeat, or
-you might need to add metadata. Filebeat provides a couple of options for
-filtering and enhancing exported data. You can:
+you might need to enhance the exported data (for example, by adding metadata).
+Filebeat provides a couple of options for filtering and enhancing exported
+data. You can:
 
-* <<filebeat-filtering-overview,Use simple filtering based on pattern matching>>
-* <<defining-processors,Define processors for more complex processing requirements>>
+* <<filebeat-filtering-overview,Define filters at the prospector level>> to
+configure each prospector to include or exclude specific lines or files.
+* <<defining-processors,Define processors>> to configure global processing
+across all data exported by Filebeat.
 
 [float]
 [[filebeat-filtering-overview]]
-=== Filtering Based on Pattern Matching
+=== Filtering at the Prospector Level
 
-You can specify configuration options in the `filebeat` section of the config file to define regular expressions that
-match the lines you want to include and/or exclude from the output. The supported options are <<include-lines,`include_lines`>>, <<exclude-lines,`exclude_lines`>>, and <<exclude-files,`exclude_files`>>.
+You can specify filtering options at the prospector level to configure which
+lines or files are included or excluded in the output. This allows you to
+specify different filtering criteria for each prospector.
 
-For example, you can use the `include_lines` option to export any lines that start with "ERR" or "WARN":
+You configure prospector-level filtering in the `filebeat.prospectors` section
+of the config file by specifying regular expressions that match the lines you
+want to include and/or exclude from the output. The supported options are
+<<include-lines,`include_lines`>>, <<exclude-lines,`exclude_lines`>>, and
+<<exclude-files,`exclude_files`>>.
+
+For example, you can use the `include_lines` option to export any lines that
+start with "ERR" or "WARN":
 
 [source,yaml]
 -------------------------------------------------------------------------------------
@@ -26,9 +37,11 @@ filebeat.prospectors:
   include_lines: ["^ERR", "^WARN"]
 -------------------------------------------------------------------------------------
 
-The disadvantage of this approach is that you need to implement a configuration option for each filtering criteria that you need.
+The disadvantage of this approach is that you need to implement a
+configuration option for each filtering criteria that you need.
 
-See <<configuration-filebeat-options,Filebeat configuration options>> for more information about each option.
+See <<configuration-filebeat-options,Filebeat configuration options>> for more
+information about each option.
 
 [float]
 [[defining-processors]]
@@ -85,7 +98,7 @@ filebeat.prospectors:
 
 processors:
   - decode_json_fields:
-      fields: ["inner"]
+      fields: ['inner']
 
 output.console.pretty: true
 -----------------------------------------------------
--- a/filebeat/docs/filebeat-filtering.asciidoc
+++ b/filebeat/docs/filebeat-filtering.asciidoc
@@ -1,11 +1,16 @@
 [[filtering-and-enhancing-data]]
 == Filtering and Enhancing the Exported Data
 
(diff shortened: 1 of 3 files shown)


libbeat/docs/processors-config.asciidoc
```diff
@@ -233,6 +233,7 @@ The supported actions are:
  * <<drop-fields,`drop_fields`>>
  * <<drop-event,`drop_event`>>
  * <<add-cloud-metadata,`add_cloud_metadata`>>
+ * <<json-decode-fields,`json_decode_fields`>>
```
andrewkroh: The name of the processors is `decode_json_fields` instead of `json_decode_fields`.

libbeat/docs/processors-config.asciidoc
```diff
@@ -371,3 +372,23 @@ _GCE_
...
+processors:
+ - json_decode_fields:
+     fields: ["field1", "field2", ...]
+     processArray: false 
+     maxDepth: 1
+-----------------------------------------------------
+
+The `json_decode_fields` action has the following configuration settings:
+
+`fields`:: The fields containing JSON strings to decode.
```
andrewkroh: @dedemorton Can you please mark this processor as experimental for 5.1 so that we can fix these config parameters in the next release. They should be named differently in the code to be consistent (like `process_array` instead of `processArray`).

libbeat/docs/processors-config.asciidoc
dedemorton: whoops...good catch.

andrewkroh:
LGTM. @dedemorton Is this still in progress?

dedemorton:
@andrewkroh I thought it might be nice to have a real example that shows the message string, the config that you would use to decode it, and the JSON object that would result.  Do you think it's necessary, or am I over-thinking this? Can you help me come up with an example? Here's a starting point (based on the test case, though I think you didn't like the formatting of the string with all the escaping, so I've tried to fix that...haven't tested it tho).

Here's the text I would add (again, let me know if you think this belabors the obvious).

-------

In the following example, a field called `msg` contains the following string:
`````
'{"log":"{\\"level\\":\\"info\\"}\","stream":"stderr","count":3}'
`````
To decode this string as a JSON object, you would use the following config:
`````
+[source,yaml]
+-----------------------------------------------------
+processors:
+ - decode_json_fields:
+     fields: ["msg"]
+     process_array: false 
+     max_depth: 1
+-----------------------------------------------------
`````
The resulting object would look like this:
`````
[source,json]
-----------------------------------------------------
"log":    "{\"level\":\"info\"}",
"stream": "stderr",
"count":  3
-----------------------------------------------------
`````

**Comment by andrewkroh** [comment]
@dede Yes, I think that would be very helpful. Here's an example with Filebeat where the JSON decoded by Filebeat contained a field whose value is a JSON object encoded as a string. We apply the processor to decode that inner JSON object.

input.json
```json
{ "outer": "value", "inner": "{\"data\": \"value\"}" }
```

filebeat.yml
```yaml
filebeat.prospectors:
- paths:
    - input.json
  json.keys_under_root: true

processors:
  - decode_json_fields:
      fields: ['inner']

output.console.pretty: true
```

output
```
{
  "@timestamp": "2016-12-06T17:38:11.541Z",
  "beat": {
    "hostname": "macbook13.local",
    "name": "macbook13.local",
    "version": "6.0.0-alpha1"
  },
  "inner": {
    "data": "value"
  },
  "input_type": "log",
  "offset": 55,
  "outer": "value",
  "source": "input.json",
  "type": "log"
}
```

suraj-soni:
Apologies for the delayed review on this request. Was away from technology for the better part of last week. Looking into this right now.

libbeat/docs/processors-config.asciidoc
```diff
@@ -371,3 +372,23 @@ _GCE_
...
+     fields: ["field1", "field2", ...]
+     process_array: false 
+     max_depth: 1
+-----------------------------------------------------
+
+The `decode_json_fields` action has the following configuration settings:
+
+`fields`:: The fields containing JSON strings to decode.
+`process_array`:: (Optional) A boolean that specifies whether to process arrays. The default is false.
+`max_dept`:: (Optional) The maximum parsing depth. The default is 1.
```
suraj-soni: Typo : "max_depth"

dedemorton: thanks. good catch!

suraj-soni:
In @andrewkroh's  example above, the fields configuration needs to be an array of strings 

`processors:
    decode_json_fields :
        fields : ["inner"]`

dedemorton:
@andrewkroh I've added your example (and changed the single quotes to double quotes as suggested, though I would think that single quotes is a valid string, no?). Anyhow, please review. Thanks!

dedemorton:
@andrewkroh Still working on this one to get the wording right...so don't merge yet.

dedemorton:
@andrewkroh Ok, it the PR is ready for real now. I've added your example. I also reworked the topic where we introduce options for enhancing the data because I felt that the existing description was kind of confusing. Sorry to clutter up this PR...perhaps I should have made those changes in a different PR, but I had to strike while the iron was hot. :-)

@monicasarbu might want to take a look at the changes that I made to filebeat-filtering.asciidoc and processors.asciidoc to verify that I've captured what needs to be said there? If this is better, I'f apply the changes to the other beats, too.

filebeat/docs/filebeat-filtering.asciidoc
```diff
@@ -53,4 +75,54 @@ processors:
...
+-----------------------------------------------------
+
+The resulting output looks something like this:
+
+["source","json",subs="attributes"]
+-----------------------------------------------------
+{
+  "@timestamp": "2016-12-06T17:38:11.541Z",
+  "beat": {
+    "hostname": "macbook13.local",
```
andrewkroh: Please change the hostname and name fields to `host.example.com`. We use that value in other examples (like in Metricbeat).

filebeat/docs/filebeat-filtering.asciidoc
```diff
@@ -1,16 +1,32 @@
 [[filtering-and-enhancing-data]]
 == Filtering and Enhancing the Exported Data
 
-When your use case requires only a subset of the data exported by Filebeat or you need to add metadata, you can <<filebeat-filtering-overview,use Filebeat config options to filter the data>>, or you can  <<defining-processors,define processors>>.
+Your use case might require only a subset of the data exported by Filebeat, or
```
andrewkroh: I like the ~80 char line wrap. 👍

filebeat/docs/filebeat-filtering.asciidoc
dedemorton: done

PR #3111, opened 2016-12-04
Use snake_case in the settings for the JSON processor

tsg:
This is to align them with the rest of our configuration files.


tsg:
This will (likely) not make it in 5.1.1, so I added a changelog entry for it.

PR #3112, opened 2016-12-04
Fix for errno 1734 when calling EvtNext

### LINKED ISSUE #3076: Winlogbeat - Reading large events causes error "The array bounds are invalid."
- Version: 5.0.2 (probably exists in all versions)
- Operating System: Windows 2012 Server R2 (probably affects Vista and newer)
- Steps to Reproduce: 

When reading a batch of large event log records the Windows function [`EvtNext`](https://msdn.microsoft.com/en-us/library/windows/desktop/aa385405%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396) returns errno 1734 (0x6C6) which is [RPC_S_INVALID_BOUND](https://msdn.microsoft.com/en-us/library/windows/desktop/ms681386(v=vs.85).aspx#RPC_S_INVALID_BOUND) ("The array bounds are invalid."). This seems to be a bug in Windows because there is no documentation about this behavior.

It seems that when calling `EvtNext` that internally it is allocating space for the events whose handles it will return. So if the combined size of the events being returned by `EvtNext` reaches some limit the RPC_S_INVALID_BOUND error is returned. In Winlogbeat, if you increase the `batch_read_size` you are more likely to encounter the problem, but even with a `batch_read_size` of 64 (default is 100) I was able to hit the error when reading large events.

Recovering from this problem is not as simple as calling `EvtNext` again with a smaller batch size. The internal read position associated with the event log handle is being updated even when `EvtNext` returns the RPC_S_INVALID_BOUND error. So if we just called it again we would lose events. We will have to close the event log and re-open it from the last read event and decrease the `batch_read_size` until success.

- Workaround: Set a lower value for `batch_read_size`.

andrewkroh:
When reading a batch of large event log records the Windows function
EvtNext returns errno 1734 (0x6C6) which is RPC_S_INVALID_BOUND ("The
array bounds are invalid."). This seems to be a bug in Windows because
there is no documentation about this behavior.

This fix handles the error by resetting the event log subscription
handle (so events are not lost) and then retries the EvtNext call
with maxHandles/2.

Fixes #3076

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -48,6 +48,7 @@ https://github.com/elastic/beats/compare/v5.0.1...master[Check the HEAD diff]
 - Fix registry cleanup issue when files falling under ignore_older after restart. {issue}2818[2818]
 
 *Winlogbeat*
+- Fix for "The array bounds are invalid" error when reading large events. {issue}3076[3076]
 
 ==== Added
 
--- a/winlogbeat/eventlog/eventlog.go
+++ b/winlogbeat/eventlog/eventlog.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"reflect"
 	"strconv"
+	"syscall"
 
 	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/logp"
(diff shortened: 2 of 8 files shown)


maddin2016:
@andrewkroh, nice work 👍. Tested on my machine with ~50k events decreased to 25 handles without loosing a event.

andrewkroh:
@maddin2016  Thanks for testing. I'm curious what your test setup was like. Did you use anything to create the 50k test events? Or did you use data that already existed?

maddin2016:
Hi @andrewkroh 
>Did you use anything to create the 50k test events?

Nothing special. I just created 500 events with 25000 `;` as message per event with powershell 😸

PR #3115, opened 2016-12-05
Fix ES 2.x integration test

tsg:
There was a test that was loading a mock template, and this template
was assuming 5.x.


ruflin:
LGTM

PR #3122, opened 2016-12-05
Remove fixed container names as not needed

ruflin:
Add beat name to project namespace

* This makes sure different beats environment do not affect each other for example when Kafka is used
* It also allows to run the testsuites of all the beats in parallel

Introduce `stop-environment` command to stop all containers

DOCKER_COMPOSE_PROJECT_NAME can be overloaded so in case lots of tests should be done in parallel, each environment is unique.


ruflin:
As soon as this PR is merged, the configuration of Beats-PR jenkins build should be adjusted to use `make stop-environments` for cleanup after build.

PR #3124, opened 2016-12-05
Backport: Fix for errno 1734 when calling EvtNext (#3112)

### LINKED ISSUE #3076: Winlogbeat - Reading large events causes error "The array bounds are invalid."

andrewkroh:
Backport #3112 to 5.1 branch

> When reading a batch of large event log records the Windows function
EvtNext returns errno 1734 (0x6C6) which is RPC_S_INVALID_BOUND ("The
array bounds are invalid."). This seems to be a bug in Windows because
there is no documentation about this behavior.
>
>This fix handles the error by resetting the event log subscription
handle (so events are not lost) and then retries the EvtNext call
with maxHandles/2.
>
>Fixes #3076

(cherry picked from commit 226eb10ce03042c3259e8055b18ddc8193fde519)

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -60,6 +60,7 @@ https://github.com/elastic/beats/compare/v5.1.1...5.1[Check the HEAD diff]
 *Filebeat*
 
 *Winlogbeat*
+- Fix for "The array bounds are invalid" error when reading large events. {issue}3076[3076]
 
 ////////////////////////////////////////////////////////////
 
--- a/winlogbeat/eventlog/bench_test.go
+++ b/winlogbeat/eventlog/bench_test.go
@@ -1,0 +1,137 @@
+// +build windows
+
+package eventlog
+
+import (
+	"flag"
+	"math/rand"
+	"os/exec"
+	"strconv"
+	"testing"
+	"time"
+
+	elog "github.com/andrewkroh/sys/windows/svc/eventlog"
+	"github.com/dustin/go-humanize"
+)
+
+// Benchmark tests with customized output. (`go test -v -benchtime 10s -benchtest .`)
+
+var (
+	benchTest    = flag.Bool("benchtest", false, "Run benchmarks for the eventlog package")
+	injectAmount = flag.Int("inject", 50000, "Number of events to inject before running benchmarks")
+)
+
+// TestBenchmarkBatchReadSize tests the performance of different
+// batch_read_size values.
+func TestBenchmarkBatchReadSize(t *testing.T) {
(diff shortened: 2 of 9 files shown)


ruflin:
@andrewkroh This seems to break the the winlogbeat tests

tsg:
Still fails, I'll wait for @andrewkroh then and don't include this one in 5.1.1 since I think it's too late now.

maddin2016:
@andrewkroh, @tsg 
```
const letterBytes = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"
const (
	letterIdxBits = 6                    // 6 bits to represent a letter index
	letterIdxMask = 1<<letterIdxBits - 1 // All 1-bits, as many as letterIdxBits
	letterIdxMax  = 63 / letterIdxBits   // # of letter indices fitting in 63 bits
)
```
import "math/rand"

and instead `rng.Int63` `rand.Int63` works for me.

andrewkroh:
I added d75667bc28fd44b685d86e6db5f81f5df71ce0c5 which has only the `bench_test.go` file containing the missing helper function.

andrewkroh:
jenkins, test it

PR #3125, opened 2016-12-06
Adding AmazonBeat to community beats

awormuth:
I created a basic version of amazonbeat, which reads data from an amazon product periodically. This beat does not yet publish to elasticsearch.

### DIFF
--- a/libbeat/docs/communitybeats.asciidoc
+++ b/libbeat/docs/communitybeats.asciidoc
@@ -7,6 +7,7 @@ out some of them here.
 NOTE: Elastic provides no warranty or support for community-sourced Beats.
 
 [horizontal]
+https://github.com/awormuth/amazonbeat[amazonbeat]:: Reads data from a specified Amazon product.
 https://github.com/radoondas/apachebeat[apachebeat]:: Reads status from Apache HTTPD server-status.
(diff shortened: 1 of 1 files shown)


ruflin:
@awormuth Thanks for creating the amazonbeat 🎉  Could you sing the CLA? https://www.elastic.co/contributor-agreement

tsg:
@awormuth regarding the [issue](https://github.com/awormuth/amazonbeat/blob/master/beater/amazonbeat.go#L114) you hit, you probably need to parse the number in Golang and put it as an integer in the structure.  Let me know if that doesn't help.

awormuth:
@ruflin absolutely! 

@tsg i'll give that a try. Will let you know!

awormuth:
@tsg your suggestion worked! Although, I'm not 100% sure why. Maybe I don't fully understand [fields.yml](https://github.com/awormuth/amazonbeat/blob/master/_meta/fields.yml). I initially had every field as "text", so why is ElasticSearch trying to convert to an integer? 

Also, I initially had a "url" field with type "text" and it threw the same conversion error with the following input: "http://www.amazon.com/dp/B00ZV9PXP2". I then switched "url" to "asin" and left the field as "text", then switched the value to "B00ZV9PXP2" and it was fine. Is "url" a reserved keyword or are there certain characters in the url that would make ElasticSearch complain?

PR #3127, opened 2016-12-06
Fetch container stats in parallel

ruflin:
Currently fetching container stats is very slow as each request takes up to 2 seconds. To improve the fetching time if lots of containers are around, this creates the rrequests in parallel. The main downside is that this opens lots of connections. This fix should only temporary until the bulk api is available: https://github.com/docker/docker/pull/25361


ruflin:
We will have to work with this fix for some longer as https://github.com/docker/docker/pull/25361#event-939699656 didn't make it into master.

PR #3129, opened 2016-12-06
Enhance kafka broker matching

urso:
Enhance bootstrap broker matching:
- configurable `advertised` address for matching
- compare broker addresses to hostname
- compare machines interfaces ips to all known broker ips

### DIFF
--- a/metricbeat/module/kafka/partition/partition.go
+++ b/metricbeat/module/kafka/partition/partition.go
@@ -1,9 +1,13 @@
 package partition
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"io"
+	"net"
+	"os"
+	"strings"
 	"time"
 
 	"github.com/elastic/beats/libbeat/common"
@@ -32,9 +36,11 @@ type MetricSet struct {
 	topics []string
 }
 
-var noID int32 = -1
+const noID int32 = -1
 
 var errFailQueryOffset = errors.New("operation failed")
+
+var debugf = logp.MakeDebug("kafka")
 
 // New create a new instance of the partition MetricSet
 func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
@@ -91,20 +97,15 @@ func (m *MetricSet) connect() (*sarama.Broker, error) {
 		return nil, err
 	}
 
-	addr := b.Addr()
-	for _, other := range meta.Brokers {
-		if other.Addr() == addr {
-			m.id = other.ID()
-			break
-		}
-	}
-
-	if m.id == noID {
+	other := findMatchingBroker(b.Addr(), meta.Brokers)
+	if other == nil { // no broker found
 		closeBroker(b)
-		err = fmt.Errorf("No advertised broker with address %v found", addr)
-		return nil, err
+		return nil, fmt.Errorf("No advertised broker with address %v found", b.Addr())
 	}
 
+	debugf("found matching broker %v with id %v", other.Addr(), other.ID())
+
+	m.id = other.ID()
 	return b, nil
 }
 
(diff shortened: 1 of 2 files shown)


metricbeat/module/kafka/partition/partition.go
```diff
@@ -322,3 +320,111 @@ func checkRetryQuery(err error) (retry, reconnect bool) {
...
+	advertised string,
+	brokers []*sarama.Broker,
+) *sarama.Broker {
+	// compare connection address to list of broker addresses
+	for _, b := range brokers {
+		if advertised != "" && b.Addr() == advertised {
+			return b
+		}
+
+		if b.Addr() == addr {
```
ruflin: Isn't this one always going to match as one of the b.Addr() is passed as one of the addr?

metricbeat/module/kafka/partition/partition.go
```diff
@@ -69,7 +71,8 @@ func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
 		BaseMetricSet: base,
 		broker:        broker,
 		cfg:           cfg,
-		id:            noID,
+		id:            config.ID,
```
ruflin: How is the id from the configuration exactly going to be used?

metricbeat/module/kafka/partition/partition.go
urso: what?

metricbeat/module/kafka/partition/partition.go
urso: `ID` sets the broker id. With broker id being configured, the broker address matching will not be executed. `id` should only be configured if number of hosts being monitored is 1 (e.g. `localhost:9092` only).

metricbeat/module/kafka/partition/partition.go
urso: The address returned by `b.Addr()` is the advertised hostname by current broker in metadata (b is not the broker we connect to, but a broker in meta.Brokers).

The advertised hostname must not match the hostname configured. e.g. kafka broker advertises hostname `kafkaX.mydomain.com:9092` and addr could be `localhost:9092`

metricbeat/module/kafka/partition/partition.go
ruflin: Do we really need both config options? Implementation wise for matching id is probably much easier. And the id of a broker should never change over the lifetime of the broker I assume?

metricbeat/module/kafka/partition/partition.go
ruflin: We pass `b.Addr()` of our broker we selected to `findMatchingBroker`. So is the `b.Addr()` from our broker the one we configured or the one fetched from the meta data? I assume it is the one we configured, as otherwise it would always match with at least on the brokers or otherwise the broker connection would not work.

metricbeat/module/kafka/partition/partition.go
urso: I'm actually considering to remove `id` and `advertised` settings. As these settings require `hosts` to be exactly one host.

Right, you can lookup the id. But I think it's much more convenient to not have to configure `id` or `advertised`. Problem with these options is, if they are wrongly configured (by accident), the kafka module will not work correctly.

metricbeat/module/kafka/partition/partition.go
urso: `b.Addr()` will be the address we configured (defaults to `localhost:9092`). the `meta.Brokers` will contains the addressed as advertised by kafka broker itself. Might be different.

metricbeat/module/kafka/partition/partition.go
ruflin: +1 in removing again id and advertised if this prevents `hosts` to be used.

metricbeat/module/kafka/partition/partition.go
ruflin: Perhaps we should have an error message in case there is no broker with the same advertised address as configured?

metricbeat/module/kafka/partition/partition.go
```diff
@@ -91,21 +97,17 @@ func (m *MetricSet) connect() (*sarama.Broker, error) {
...
+	other := findMatchingBroker(b.Addr(), meta.Brokers)
+	if other != nil {
+		debugf("found matching broker %v with id %v", other.Addr(), other.ID())
+		m.id = other.ID()
+		return b, nil
 	}
 
-	return b, nil
+	// no broker if found
+	closeBroker(b)
```
ruflin: Could we turn that around, that we have if other == nil we return the error and close the broker and move the non error case tot he bottom? Would make it easier readable.

urso: Done

ruflin:
@urso LGTM. Could you squash, update the commit message and add a CHANGELOG entry?

PR #3136, opened 2016-12-07
Remove --always-copy from virtualenv and make it a param

ruflin:
In https://github.com/elastic/beats/pull/3082 --always-copy was introduced. This caused issue on build on some operating system. This PR reverts the change but makes VIRTUALENV_PARAMS a variable which can be passed to the Makefile. This allows anyone to set --always-copy if needed.

@phenomenes Please be aware of this change.


phenomenes:
👍

PR #3137, opened 2016-12-07
Adjustments to full config file

ruflin:
Currently the full config has mostly the defaults in but not in all cases. The intention is to have clear rules on the structure and content of the full config files.

Rules:

* Starting filebeat with the default `filebeat.yml` or `filebeat.full.yml` must behave exactly the same
* All config options which can be uncommented, must be uncommented. This is for easy copy / pasting of options
* In case no defaults should be set like for `fields`, an example should be provided in the comment above it or below commented out
* In case not default is used, special comments must be left. Only option which are not default should be `enabled / default`.
* A second reason for no values set is if the value is dynamic like `max_procs`. Special comment must be left.
* In case not all options can be presented like for filters, link to documentation must be added. 2-3 examples should be provided.

This is currently only a started example for Filebeat to show the new structure

Question: What do we do about the special `stdin` case in Filebeat. Should we add `enabled` to the prospector for consistency. This would also make it easier for testing to just disable one prospector without having it to comment out.

### DIFF
--- a/filebeat/_meta/beat.full.yml
+++ b/filebeat/_meta/beat.full.yml
@@ -39,18 +39,19 @@ filebeat.prospectors:
   # Some sample encodings:
   #   plain, utf-8, utf-16be-bom, utf-16be, utf-16le, big5, gb18030, gbk,
   #    hz-gb-2312, euc-kr, euc-jp, iso-2022-jp, shift-jis, ...
-  #encoding: plain
-
+  encoding: plain
 
   # Exclude lines. A list of regular expressions to match. It drops the lines that are
   # matching any regular expression from the list. The include_lines is called before
   # exclude_lines. By default, no lines are dropped.
-  #exclude_lines: ["^DBG"]
+  # Example: ["^DBG"]
+  exclude_lines: []
 
   # Include lines. A list of regular expressions to match. It exports the lines that are
   # matching any regular expression from the list. The include_lines is called before
   # exclude_lines. By default, all the lines are exported.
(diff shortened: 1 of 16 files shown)


tsg:
> All config options which can be uncommented, must be uncommented. This is for easy copy / pasting of options

Hmm, this will make the file easier to scan (thanks to syntax highlighting in the editors) but it's also a bit dangerous. Errors in the configuration files will be more unforgiven. So we'd need a testing strategy for the full configs.

ruflin:
@tsg Good point about the testing strategy. We should probably also test the short config and then do a comparison of output / logs to see if the behaviour is identical.

filebeat/_meta/beat.full.yml
```diff
@@ -212,25 +213,25 @@ filebeat.prospectors:
...
 # data path.
-#filebeat.registry_file: ${path.data}/registry
+filebeat.registry_file: ${path.data}/registry
 
 #
 # These config files must have the full filebeat config part inside, but only
 # the prospector part is processed. All global options like spool_size are ignored.
 # The config_dir MUST point to a different directory then where the main filebeat config file is in.
-#filebeat.config_dir:
+filebeat.config_dir: ${path.config}
```
urso: not sure this is safe.

ruflin: Agree. I need to test the changes on all paths.

ruflin:
Closing this in favor of https://github.com/elastic/beats/issues/3179 and multiple small PR's.

PR #3139, opened 2016-12-07
Pass the certificate as a parameter to the import_dashboards script

monicasarbu:
Add `-ca` and `-key` option to the `import_dashboards` script in order to be able to pass the certificate. 
This PR is a workaround for now, until `import_dashboards` will be part of the Beat executable.

### DIFF
--- a/libbeat/dashboards/import_dashboards.go
+++ b/libbeat/dashboards/import_dashboards.go
@@ -96,9 +96,9 @@ func DefineCommandLine() (*CommandLine, error) {
 	cl.flagSet.BoolVar(&cl.opt.OnlyDashboards, "only-dashboards", false, "Import only dashboards together with visualizations and searches. By default import both, dashboards and the index-pattern.")
 	cl.flagSet.BoolVar(&cl.opt.OnlyIndex, "only-index", false, "Import only the index-pattern. By default imports both, dashboards and the index pattern.")
 	cl.flagSet.BoolVar(&cl.opt.Snapshot, "snapshot", false, "Import dashboards from snapshot builds.")
-	cl.flagSet.StringVar(&cl.opt.CertificateAuthority, "ca", "", "Certificate Authority for server verification")
-	cl.flagSet.StringVar(&cl.opt.Certificate, "cert", "", "Certificate for SSL client authentication.")
-	cl.flagSet.StringVar(&cl.opt.CertificateKey, "certkey", "", "Client Certificate Key")
+	cl.flagSet.StringVar(&cl.opt.CertificateAuthority, "cacert", "", "Certificate Authority for server verification")
+	cl.flagSet.StringVar(&cl.opt.Certificate, "cert", "", "Certificate for SSL client authentication in PEM format.")
+	cl.flagSet.StringVar(&cl.opt.CertificateKey, "key", "", "Client Certificate Key in PEM format.")
 
 	return &cl, nil
 }
@@ -116,7 +116,7 @@ func (cl *CommandLine) ParseCommandLine() error {
 	}
 
 	if cl.opt.Certificate != "" && cl.opt.CertificateKey == "" {
-		return errors.New("ERROR: A certificate key needs to be passed as well by using the -certkey option.")
+		return errors.New("ERROR: A certificate key needs to be passed as well by using the -key option.")
 	}
 
 	if cl.opt.CertificateKey != "" && cl.opt.Certificate == "" {
--- a/libbeat/dashboards/import_dashboards.go
+++ b/libbeat/dashboards/import_dashboards.go
@@ -18,8 +18,10 @@ import (
 	lbeat "github.com/elastic/beats/libbeat/beat"
 	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/common/fmtstr"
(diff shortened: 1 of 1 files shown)


andrewkroh:
I was thinking we should have an option for a client cert and key too. This would cover the case where shield is setup to require TLS mutual auth.

monicasarbu:
@andrewkroh Yes, you are right. I added another option `-key` to be able to pass the the certificate key. I have updated also the description of the PR.

libbeat/dashboards/import_dashboards.go
```diff
@@ -91,6 +95,8 @@ func DefineCommandLine() (*CommandLine, error) {
 	cl.flagSet.BoolVar(&cl.opt.OnlyDashboards, "only-dashboards", false, "Import only dashboards together with visualizations and searches. By default import both, dashboards and the index-pattern.")
 	cl.flagSet.BoolVar(&cl.opt.OnlyIndex, "only-index", false, "Import only the index-pattern. By default imports both, dashboards and the index pattern.")
 	cl.flagSet.BoolVar(&cl.opt.Snapshot, "snapshot", false, "Import dashboards from snapshot builds.")
+	cl.flagSet.StringVar(&cl.opt.Certificate, "ca", "", "Certificate for SSL client authentication.")
```
andrewkroh: There should be three options like `cacert`, `cert`, and `key`. I chose those names because they are the same as `curl`. And curl also allows a key passphrase to be specified with the `pass` CLI flag.

libbeat/dashboards/import_dashboards.go
```diff
@@ -133,11 +147,27 @@ func New() (*Importer, error) {
 	}
 	indexSel := outil.MakeSelector(outil.FmtSelectorExpr(fmtstr, ""))
 
+	var tls *transport.TLSConfig
+
+	if len(cl.opt.Certificate) > 0 && len(cl.opt.Key) > 0 {
+		tls, err = outputs.LoadTLSConfig(&outputs.TLSConfig{
+			Certificate: outputs.CertificateConfig{
+				Certificate: cl.opt.Certificate,
+				Key:         cl.opt.Key,
```
andrewkroh: You could also set `Passphrase: cl.opt.Pass` to offer a way to specify a password for an encrypted key file.

libbeat/dashboards/import_dashboards.go
```diff
@@ -107,6 +113,14 @@ func (cl *CommandLine) ParseCommandLine() error {
 		return errors.New("ERROR: Missing input. Please specify one of the options -file, -url or -dir")
 	}
 
+	if cl.opt.Certificate != "" && cl.opt.Key == "" {
+		return errors.New("ERROR: A certificate key needs to be passed as well by using the -key option.")
+	}
+
+	if cl.opt.Key != "" && cl.opt.Certificate == "" {
+		return errors.New("ERROR: A certificate needs to be passed as well by using the -ca option.")
```
andrewkroh: `s/-ca/-cert/`

libbeat/dashboards/import_dashboards.go
andrewkroh: The description of the `cacert` param should say that the file may contain multiple PEM encoded CA certificates.

libbeat/dashboards/import_dashboards.go
```diff
@@ -91,6 +95,8 @@ func DefineCommandLine() (*CommandLine, error) {
 	cl.flagSet.BoolVar(&cl.opt.OnlyDashboards, "only-dashboards", false, "Import only dashboards together with visualizations and searches. By default import both, dashboards and the index-pattern.")
 	cl.flagSet.BoolVar(&cl.opt.OnlyIndex, "only-index", false, "Import only the index-pattern. By default imports both, dashboards and the index pattern.")
 	cl.flagSet.BoolVar(&cl.opt.Snapshot, "snapshot", false, "Import dashboards from snapshot builds.")
+	cl.flagSet.StringVar(&cl.opt.Certificate, "ca", "", "Certificate for SSL client authentication.")
+	cl.flagSet.StringVar(&cl.opt.Key, "key", "", "Client Certificate Key")
```
andrewkroh: The `cert` and `key` descriptions should mention PEM. (We don't support other formats.)

libbeat/dashboards/import_dashboards.go
monicasarbu: @andrewkroh I was working in the meantime on a solution where you can pass the `-cert` the client certificate, `-certkey` certificate key, and `-ca` certificate authority. I will change the code to use the proposed names as `curl` does.

libbeat/dashboards/import_dashboards.go
monicasarbu: @andrewkroh Yes true, but as it's a temporary solution, I think it is better to minimize the changes.

libbeat/dashboards/import_dashboards.go
monicasarbu: done

libbeat/dashboards/import_dashboards.go
monicasarbu: done

monicasarbu:
@andrewkroh I addressed your comments, so please have another look.

Issue #3140, opened 2016-12-07
Packetbeat-Cassandra.json has colons in the visualization names


LeeDr:
`beats-dashboards-5.1.1/packetbeat/dashboard/Packetbeat-Cassandra.json` has colons in the visualization names

- Version: 5.1.1 build candidate from 2016-12-06
- Operating System: Ubuntu 32-bit
- Steps to Reproduce: 
1. ./import_dashboards -user $ELASTICUSER -pass $ELASTICPWD -url http://${BASEURL}/downloads/beats/beats-dashboards/beats-dashboards-5.1.1.zip

End of Output:
```
Import search  /tmp/tmp885745225/beats-dashboards-5.1.1/packetbeat/search/Thrift-transactions.json
Import vizualization  /tmp/tmp885745225/beats-dashboards-5.1.1/packetbeat/visualization/Top-Thrift-RPC-calls-with-errors.json
Import search  /tmp/tmp885745225/beats-dashboards-5.1.1/packetbeat/search/Thrift-errors.json
fail to import dashboards: fail to load directory /tmp/tmp885745225/beats-dashboards-5.1.1/packetbeat/dashboard: error loading /tmp/tmp885745225/beats-dashboards-5.1.1/packetbeat/dashboard/Packetbeat-Cassandra.json: fail to read /tmp/tmp885745225/beats-dashboards-5.1.1/packetbeat/visualization/Cassandra:-ResponseKeyspace.json. Error: open /tmp/tmp885745225/beats-dashboards-5.1.1/packetbeat/visualization/Cassandra:-ResponseKeyspace.json: no such file or directory
```

ruflin:
@LeeDr Thanks for reporting. A fix is provided here: https://github.com/elastic/beats/pull/3146 It seems we forgotten to correct the links in the Cassandra dashboards when the visualisations were renamed in https://github.com/elastic/beats/pull/2946 :-(

PR #3146, opened 2016-12-08
Fix for broken dashboard dependency in Cassandra Dashboard

ruflin:
The Cassandra Dashboard was linking to the wrong Cassandra visualisation. Some left over with : in the names were still inside

Closes https://github.com/elastic/beats/issues/3140


monicasarbu:
I have tested as well locally and it seems to work. Thank you @ruflin for the fix.

PR #3147, opened 2016-12-08
Adjust script to generate fields of type geo_point

monicasarbu:
Add functionality to generate fields of type `geo_point` in the `generate_index_pattern` script. 
NOTE: Without this fix, the user needs to reload the `packetbeat-*` index pattern in Kibana.


monicasarbu:
I have tested manually with the latest version of ES and ingest-geoip.

PR #3160, opened 2016-12-09
Restructure and edit processors content

dedemorton:
Summary of changes:
- Reorganized processor content so that each processor is a separate topic. Due to asciidoc limitations, this meant that I needed to move processors up a level in the right-hand navigation (see screen capture below). This raises the visibility of processors, which I think is a good idea anyhow.
- Made some edits to streamline the description of processor syntax. Also changed the terminology around processors so that we use "processor" instead of "action" where we are talking about a specific processor by name. This is consistent with how we describe processors for ingest node.
- Rearranged content so that processors are sorted alphabetically (I'm assuming that we will add more processors eventually and want them alphabetized).
- Wrapped lines a 80 characters (trying to do a general cleanup over time because 80 chars works better for viewing diffs). My apologies if this makes reviewing harder. 

processorsnav


metricbeat/docs/metricbeat-filtering.asciidoc
```diff
@@ -1,18 +1,23 @@
...
 
-When your use case requires only a subset of the data exported by Metricbeat or you need to add metadata,
-you can <<metricbeat-filtering-overview,use Metricbeat config options to filter
-the data>>, or you can  <<defining-processors,define processors>>.
+Your use case might require only a subset of the data exported by Metricbeat, or
+you might need to enhance the exported data (for example, by adding metadata).
+Metricbeat provides a couple of options for filtering and enhancing exported
+data. You can:
 
+* <<metricbeat-filtering-overview,Define filters at the module level>> to
```
ruflin: Should we now remove that line?

ruflin:
@dedemorton I move this PR as this PR is only for restructuring and potential removal can be done in https://github.com/elastic/beats/pull/3173

Issue #3159, opened 2016-12-09
Meta ticket: Filebeat modules


tsg:
This is the meta ticket for the Filebeat modules implementation.

## TODOs and progress:

* [x] #3158 Add a sample module (NGINX) 
* [x] #3158 Prototype module loading
* [x] #3195 Add support for multiple paths on the same OS in the Nginx module
* [x] #3171 Add sample module for Mysql
* [x] #3191 Add sample module for Syslog
* [x] #3214 Add system tests for the modules
* [x] #3221 Move Kibana dashboards at the module level
* [x] #3248 Add a module generator
* [x] #3256 Apache2 module
* [x] #3333 Phase 1 of the go implementation: create prospector from the modules
* [x] #3394 Load the pipeline automatically
* [x] #3433 Make the pipeline configurable from the prospector config
* [x] #3472 Pass the pipeline ID from modules automatically
* [x] #3436 Include the module files in the filebeat packages
* [x] #3506 Replace the `import_dashboard` program with `-setup`
* [x] #3516 Add the Beat version in the pipeline ID
* [x] #3522 Only insert the pipeline if it's not already loaded
* [x] #3524 Test the windows versions of the Filebeat modules
* [x] #3540 Replace `fields.source_type` with `fileset.module` and `fileset.name`
* [x] #3592 Docs: overview & tutorial
* [x] #3598 Docs: create & gather per module docs and configuration samples
* [x] #3616 Docs: Add a guide for creating modules
* Docs: Add Logstash equivalent configurations. Tracked in: https://github.com/elastic/logstash/issues/6542

## Overview

Filebeat modules are "packages" of the required configurations and logic to ship and analyze log files from common services. A typical module (e.g nginx) is composed of several filesets, one for each type of logs (e.g. access and error for nginx). A [typical fileset](https://github.com/elastic/beats/tree/master/filebeat/module/nginx/access) contains the following:

* filebeat prospector configuration
* Elasticsearch Ingest Node pipeline definition
* Fields definitions and docs
* Kibana dashboards (at the module level)
* Test log files
* A [manifest.yml](https://github.com/elastic/beats/blob/master/filebeat/module/nginx/access/manifest.yml) file with overwritable variables and logic to select the right files

At the moment, Filebeat modules are strictly configuration files and templates, no actual Go code. This makes it easy to create new module. Eventually we might have some of the modules include Go code/plugins for more specific needs.

[...]

PR #3171, opened 2016-12-12
Filebeat mysql module

tsg:
Contains filesets for the slow logs and the error logs. 

* The slow logs use the multiline features in filebeat.
* The error logs have two format types, using two grok patterns for them, and a catch all rule.

This also adds the concept of built in variables, for example "hostname". This is needed because the path of the mysql logs depend on the hostname.

Contains a dashboard as well:

screen shot 2016-12-13 at 13 01 00

Part of #3159.

### DIFF
--- a/filebeat/module/nginx/error/ingest/pipeline.json
+++ b/filebeat/module/nginx/error/ingest/pipeline.json
@@ -5,7 +5,12 @@
       "field": "message",
       "patterns": [
         "%{DATA:nginx.error.time} \\[%{DATA:nginx.error.level}\\] %{NUMBER:nginx.error.pid}#%{NUMBER:nginx.error.tid}: (\\*%{NUMBER:nginx.error.connection_id} )?%{GREEDYDATA:nginx.error.message}"
-      ]
+      ],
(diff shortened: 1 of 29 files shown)


filebeat/module/nginx/error/ingest/pipeline.json
```diff
@@ -5,7 +5,12 @@
       "field": "message",
       "patterns": [
         "%{DATA:nginx.error.time} \\[%{DATA:nginx.error.level}\\] %{NUMBER:nginx.error.pid}#%{NUMBER:nginx.error.tid}: (\\*%{NUMBER:nginx.error.connection_id} )?%{GREEDYDATA:nginx.error.message}"
-      ]
+      ],
+      "on_failure" : [{
```
tsg: You are commenting on an nginx file :)

tsg:
Should be ready for reviews, now.

tsg:
Added back "in progress", as this doesn't work for older mysql versions.

tsg:
Should be mergeable again.

PR #3176, opened 2016-12-13
Add edits for docker module docs

dedemorton:
Edits for docker module content.

@ruflin I made an editorial pass through the docker module content. I'll need to apply similar edits to a couple of the other newer modules so that we're consistent across each module. The content seems "light" in some topics, but I think most users will look at the sample JSON rather than reading texts, so I only want to add more descriptive text if it adds some value. :-) Let me know if you think there are some areas that need fleshing out.


ruflin:
@dedemorton LGTM. Thanks for bringing some consistency to the docs here. I agree some other parts also need cleanup to be consistent.

Issue #3179, opened 2016-12-13
Adjustments to full config file


ruflin:
Currently the full config has mostly the defaults in but not in all cases. The intention is to have clear rules on the structure and content of the full config files.

Rules:

* Starting filebeat with the default `filebeat.yml` or `filebeat.full.yml` must behave exactly the same
* All config options which can be uncommented, must be uncommented. This is for easy copy / pasting of options
* In case no defaults should be set like for `fields`, an example should be provided in the comment above it or below commented out
* In case not default is used, special comments must be left. Only option which are not default should be `enabled / default`.
* A second reason for no values set is if the value is dynamic like `max_procs`. Special comment must be left.
* In case not all options can be presented like for filters, link to documentation must be added. 2-3 examples should be provided.

This is currently only a started example for Filebeat to show the new structure

Tasks
* [x] Add enabled / disable to filebeat https://github.com/elastic/beats/pull/3157
* [ ] Update metricbeat full config https://github.com/elastic/beats/pull/3180
* [ ] Update filebeat full config https://github.com/elastic/beats/pull/3183
* [ ] Update packetbeat full config https://github.com/elastic/beats/pull/3185
* [ ] Update heartbeat full config
* [ ] Update winlogbeat full config https://github.com/elastic/beats/pull/3184
* [ ] Update libbeat full config
* [ ] Add tests to verify compact and full config file work the same

ruflin:
Working more on this, I would suggest to make the following changes to the above "rules" and change it from "rules" to "guideline" as often there is also some gray area:

Guidelines:
* Starting a beat with the default beat.yml or beat.full.yml must behave exactly the same
* The common config options with defaults should be uncommented
* In case no defaults can be set like for fields or processors, a link to the docs must be provided
* For config options which are reused across muliple modules, link to docs should be provided. Example: Authentication configs in MB modules based on http

The main change is that example configs should be in the docs and not in the full config, so the full config should make it very easy to jump to the correct docs.

ruflin:
I'm closing the already open PR's as I think the migration should not happen all at once but in small steps. There is also the still open discussion if we should potentially use a template language to generate the config files. This would allow to just have 1 file with short and full inside. Also links could be dynamically generated potentially.

ruflin:
Closing this as we renamed by now the full config to reference and since then didn't have much asks for any changes.

PR #3180, opened 2016-12-13
Update metricbeat full config files

ruflin:
* Apply https://github.com/elastic/beats/issues/3179 to config files
* Set short configs which are not enabled to enabled:false

There should not be any behavioural change.


ruflin:
@tsg As the libbeat part is still missing and in other PR's there are some open questions, I suggest we keep this for the next minor / major release, so we don't merge it now. This also gives us the opportunity to add more tests. And it will also allow us to merge one PR after the other going step by step.

ruflin:
Closing in favor of https://github.com/elastic/beats/pull/4049

PR #3184, opened 2016-12-13
Update full winlogbeat config

ruflin:
Apply https://github.com/elastic/beats/issues/3179

### DIFF
--- a/winlogbeat/_meta/beat.full.yml
+++ b/winlogbeat/_meta/beat.full.yml
@@ -12,13 +12,13 @@
 # The registry file is where Winlogbeat persists its state so that the beat
 # can resume after shutdown or an outage. The default is .winlogbeat.yml
 # in the directory in which it was started.
-#winlogbeat.registry_file: .winlogbeat.yml
+winlogbeat.registry_file: .winlogbeat.yml
 
 # Diagnostic metrics that can retrieved through a web interface if a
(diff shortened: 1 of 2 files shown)


ruflin:
@andrewkroh Is this full config file complete? If not, should we complete it?

winlogbeat/_meta/beat.full.yml
```diff
@@ -12,13 +12,13 @@
...
 # in the directory in which it was started.
-#winlogbeat.registry_file: .winlogbeat.yml
+winlogbeat.registry_file: .winlogbeat.yml
 
 # Diagnostic metrics that can retrieved through a web interface if a
 # bindaddress value (host:port) is specified. The web address will be
 # http://<bindaddress>/debug/vars
-#winlogbeat.metrics:
-#  bindaddress: 'localhost:8123'
+winlogbeat.metrics:
```
andrewkroh: Is this correct for the rules you are applying? This option does not have an `enabled: true/false` setting and it is _not_ enabled by default. So should it still be commented out?

ruflin: If it is not enabled by default and does not have an enabled flag, it should be commented out. Otherwise short and full config have different behaviour.

ruflin:
Closing because of https://github.com/elastic/beats/issues/3179#issuecomment-295606194

PR #3185, opened 2016-12-13
Update packetbeat full config

ruflin:
Apply https://github.com/elastic/beats/issues/3179

### DIFF
--- a/packetbeat/_meta/beat.full.yml
+++ b/packetbeat/_meta/beat.full.yml
@@ -21,33 +21,33 @@ packetbeat.interfaces.device: any
 # * pf_ring, which makes use of an ntop.org project. This setting provides the
 # best sniffing speed, but it requires a kernel module, and it's Linux-specific.
 # The default sniffer type is pcap.
-#packetbeat.interfaces.type: pcap
+packetbeat.interfaces.type: pcap
 
 # The maximum size of the packets to capture. The default is 65535, which is
 # large enough for almost all networks and interface types. If you sniff on a
 # physical network interface, the optimal setting is the MTU size. On virtual
 # interfaces, however, it's safer to accept the default value.
-#packetbeat.interfaces.snaplen: 65535
+packetbeat.interfaces.snaplen: 65535
 
 # The maximum size of the shared memory buffer to use between the kernel and
 # user space. A bigger buffer usually results in lower CPU usage, but consumes
 # more memory. This setting is only available for the af_packet sniffer type.
 # The default is 30 MB.
-#packetbeat.interfaces.buffer_size_mb: 30
+packetbeat.interfaces.buffer_size_mb: 30
 
 # Packetbeat automatically generates a BPF for capturing only the traffic on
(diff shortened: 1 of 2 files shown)


ruflin:
@monicasarbu Could you check if I chose the correct default values?

packetbeat/_meta/beat.full.yml
```diff
@@ -432,23 +438,22 @@ packetbeat.protocols.nfs:
...
+      cmdline_grep: postgres
+
+    - process: nginx
+      cmdline_grep: nginx
+
+    - process: app
+      cmdline_grep: gunicorn
 
-# Uncomment the following if you want to ignore transactions created
+# Set the following to true if you want to ignore transactions created
```
ruflin: @tsg The title kind of indicates that this belongs to Monitored processes, but I think it doesn't? Perhaps we should just have a title for "General" packetbeat config and then it would also belong on the top with interfaces or interfaces at the bottom. WDYT?

packetbeat/_meta/beat.full.yml
```diff
@@ -432,23 +438,22 @@ packetbeat.protocols.nfs:
...
-#    - process: nginx
-#      cmdline_grep: nginx
-#
-#    - process: app
-#      cmdline_grep: gunicorn
+packetbeat.procs:
+  enabled: false
+  monitored:
+    - process: mysqld
+      cmdline_grep: mysqld
```
tsg: Hmm, these are examples, kind of like we have for processors. I'd say it doesn't make sense to have them commented out. I know it's `enabled: false`, but still looks weird to me.

packetbeat/_meta/beat.full.yml
tsg: Ah, yes, that should have been under "General", I probably copied it there by mistake. There is already a "General" section, btw.

packetbeat/_meta/beat.full.yml
ruflin: Ok, will comment it out.

packetbeat/packetbeat.full.yml
```diff
@@ -21,33 +21,33 @@ packetbeat.interfaces.device: any
...
 
 # Packetbeat automatically generates a BPF for capturing only the traffic on
 # ports where it expects to find known protocols. Use this settings to tell
 # Packetbeat to generate a BPF filter that accepts VLAN tags.
-#packetbeat.interfaces.with_vlans: true
+packetbeat.interfaces.with_vlans: true
 
 # Use this setting to override the automatically generated BPF filter.
-#packetbeat.interfaces.bpf_filter:
+packetbeat.interfaces.bpf_filter:
```
ruflin: Yes, that would be an empty string. Isn't that also the default if it is not set? https://github.com/elastic/beats/blob/master/packetbeat/beater/packetbeat.go#L196

ruflin:
Closing because of https://github.com/elastic/beats/issues/3179#issuecomment-295606194

PR #3195, opened 2016-12-14
Add support for multiple paths per fileset

tsg:
We generally need more than one path per OS, because the logs location
is not always the same. For example, depending on the linux distribution
and how you installed it, MySQL can have it's error logs in a number of
default "paths". The solution is to configure them all, which means that
Filebeat might try to access unexisting folders.

This also improves the python prototype to accept multiple modules and
to accept namespaced parameters. E.g.:

./filebeat.py --modules=nginx,syslog -M nginx.access.paths=...

Part of #3159.


ruflin:
I think the affect of this is very minimal. If there are no files, glob will just return an empty result and nothing will happen.

PR #3197, opened 2016-12-15
Update generator docs to use specific version

ruflin:
This change will prevent people creating a new beat to rely on master and have a stable branch for development. This also allows us to modify the generator without having all docs out of sync directly.

The blog post "Build your own beat" should be updated with an "update note" about checking out branch 5.1: https://www.elastic.co/blog/build-your-own-beat

This is in preparation for https://github.com/elastic/beats/pull/2996


libbeat/docs/newbeat.asciidoc
```diff
@@ -52,7 +52,14 @@ go get github.com/elastic/beats
...
 When you run the command, all source files are downloaded to the
 `$GOPATH/src/github.com/elastic/beats` path. You can ignore the "no buildable Go source files" message because
-you will build the source later.
+you will build the source later. By default `go get`  fetches the master branch. To build your beat
+on a specific version of libbeat, check out the specific branch (5.1 in the example below):
+
+[source,bash]
+----
+cd $GOPATH/src/github.com/elastic/beats
+git checkout 5.1
```
ruflin: we could you {doc_branch} here but that will not work in the docs above which are also shown on Github. We have to find some simplification here in the future.

ruflin:
This change needs backport to 5.0 and 5.1

metricbeat/docs/developer-guide/creating-beat-from-metricbeat.asciidoc
```diff
@@ -27,7 +27,17 @@ The first step is to get the metricbeat source code:
...
 ----
 
-This will clone the beats repository into `GOPATH`. Now change to the directory where you want to create the beat.
+This will clone the beats repository into `GOPATH`. By default `go get`  fetches the master branch. To build your beat
+on a specific version of libbeat, check out the specific branch ({doc-branch} in the example below):
+
+[source,bash,,subs="attributes"]
+----
+cd $GOPATH/src/github.com/elastic/beats
+git checkout {doc-branch}
```
ruflin: @dedemorton I tried `sub="attributes"` here so the variable gets replaced. This does not seem to work. If I leave out one , I get a different error. Any idea on how this is best done?

dedemorton: Try this instead: `["source","sh",subs="attributes"]`

dedemorton:
other than the tagging you asked about, LGTM

metricbeat/docs/developer-guide/creating-beat-from-metricbeat.asciidoc
ruflin: That seems to work. Thanks.

PR #3198, opened 2016-12-15
fix inode reuse bug

liuzheng712:
really unhappy about this bug, find a lot of solutions didn't work well.

just add 3 lines fix this problem.

Why not analyze the file's size, if inode reuse, the size must smaller than before.

Just checkout the size will fix it!

### DIFF
--- a/filebeat/input/file/file_other.go
+++ b/filebeat/input/file/file_other.go
@@ -12,6 +12,7 @@ import (
 type StateOS struct {
 	Inode  uint64 `json:"inode,"`
 	Device uint64 `json:"device,"`
+        Size   uint64 `json:"size,"`
 }
 
 // GetOSState returns the FileStateOS for non windows systems
@@ -23,6 +24,7 @@ func GetOSState(info os.FileInfo) StateOS {
 	fileState := StateOS{
 		Inode:  uint64(stat.Ino),
 		Device: uint64(stat.Dev),
+		Size:   uint64(stat.Size),
 	}
 
 	return fileState
@@ -30,7 +32,7 @@ func GetOSState(info os.FileInfo) StateOS {
 
 // IsSame file checks if the files are identical
 func (fs StateOS) IsSame(state StateOS) bool {
-	return fs.Inode == state.Inode && fs.Device == state.Device
+	return fs.Inode == state.Inode && fs.Device == state.Device && fs.Size <= state.Size
 }
 
 // SafeFileRotate safely rotates an existing file under path and replaces it with the tempfile


liuzheng712:
Never see those robot before~ Really don't know what happen ( with_scare_face.png )

ruflin:
I'm not sure this solves the issue. Filebeat detects already know if the file size is smaller then previous. But that normally means a file was truncated which quite a few of the log rotating mechanism support. You you will in the tests that the truncation tests fail.

With the 5.0 release there are the clean_* options to prevent this issue. Did you find some limitations with these?

liuzheng712:
@ruflin yes, I was used the clean_* options, but it works still not very well. so I have no chose to judge the file size to do this. It is force but it works well.

The log file's definition is the size must increase.

In my company, the log rotating maybe 1min or 24h, it is hard to decide the clean_* options.

I read the tail's source code, it use eternal loop with no period time to scan a file.

But filebeat use the period time 1s, I was thought, if in that second the file happen a lot of things?

Still can't find a perfect solution.

Improve  the log transmission ability will cause a lot internal network bandwidth.
When the system print a lot log that means a lot user's request, the internal bandwidth is so import!

Still find a better solution...

ruflin:
Lots of log rotation tools also support log truncation. I would like to agree with you that the definition of a log is that it constantly increases, unfortunately the reality is different. 

Can you elaborate on the issues you hit with the clean_* options? What is the log rotation based on in your company? File size? 

Not sure I can follow how the tail file code is related to this one?

liuzheng712:
In my company, the log will rotate in daily and when size reach 200M.
Because I need support all languages' log, most languages have different log file use case.
1. always use one inode, and gzip when size 200M or daily, and the log maybe in per 1~5min gzip. 
2. use 3 inode to polling, but only one inode active, and gzip when size 200M or daily, and the log maybe in per 1~5min gzip. 


Now I am trying to make my own filebeat with metricbeat, but the metricbeat was changed, different with 5.0.0-alpha4. I was very mad with this.....


http://git.savannah.gnu.org/cgit/coreutils.git/tree/src/tail.c#n592

You can see tail use while(1), because those inode reuse problem, I try to find what tail to achieve.

liuzheng712:
I think it can be use period with 0s?
Or use `tail` for input , not read the file, or read the file the same way like `tail`?

liuzheng712:
@ruflin can you tell me which case that the log size will decrease?

ruflin:
For example RollingFileAppender in log4j can use file truncation: http://logging.apache.org/log4j/1.2/apidocs/org/apache/log4j/RollingFileAppender.html

* How is metricbeat releated to this? Are you trying to execute tail? If you want to do this, you could pipe tail to filebeat and use stdin as input probably. Still I would hope your problems could be solved with the clean options
* If you have different types of logs, you can use multiple prospectors with different configurations

I'm not sure I understand how the `while` in the tail source code solves the inode reuse case. The config option you are looking for is probably `backoff` and set reconfigure these.

What do you mean by using 3 indoes for polling?

liuzheng712:
@ruflin 

  * I was made filebeat into metricbeat module 2 times, first time was based on beats tag v5.0.0-alpha4,  second try is the tag v5.1.1 . all can be work just in input:logfile and output:reids, other case I did not test and I think it also works.
  * I wrote a tail go version demo in here, https://github.com/liuzheng712/tailgo , it can work like tail, but still have a lot problem, and I am writing the perfect version in my company, still fix some bug.
  * if use `tail -f *.log` as stdin, when log file rotated or change name which means inode did not change or filename did not change. When comes a new log file, the tail will not monitor this file.
  * I was used all the options filebeat give and tried to add some function to improve it, but did not work well as I wish.
  * `tail` use `while true` to check file status and print the log, which means no period time.
  * the logback will use 3 inodes to avoid the inode exhaustion problem, and save io open cost, I think

ruflin:
* Can you share the code you built for metricbeat? I'm curious how this works in combination with the pull model.
* logback: Is that what you are referring to? http://logback.qos.ch/ ? As far as I know log4j is a logging framework, so logback is also a logging framework? How is this exactly related to the issue here? Are you using logback to write log files?

Can you share more details on what you tried to solve the inode reuse issue and why it didn't work? I'm very curious to learn more here as so far with 5.1 the existing config options were sufficient to cover the edge cases I'm aware of.

ruflin:
@liuzheng712 Closing this PR as I do not think it fixes the issue. But I'm more then happy to discuss this further and find potential solution. Also I would really like to hear what are the options that you tried and didn't work.

Issue #3138, opened 2016-12-07
filebeat conditional checks fail on array fields when it's been parsed by json parser.


JalehD:
- Version: 5.0.2
- Operating System: any
- Steps to Reproduce: Using below sample events and config it is easy to produce below error and the events are not dropped as expected.

`{"@timestamp":"2016-12-06T11:24:13.731+00:00","@version":1,"message":"free text for reference","logger":"test","thread_name":"http-test","level":"INFO","level_value":20000,"HOSTNAME":"Juli-lab","duration":"985","depth":"1","action":"test Request","appPath":"/tmp/test","order":"1","tags":["ELK_TRACE"]}
{"@timestamp":"2016-12-06T11:24:13.788+00:00","@version":1,"message":"free text for reference","logger":"test","thread_name":"http-test","level":"INFO","level_value":20000,"HOSTNAME":"juli-lab","duration":"1060","depth":"0","action":"test response","appPath":"/tmp/test","order":"1","tags":["ELK_TRACE"]}`

With config 

```
filebeat.prospectors:
 - input_type: log
   paths: 
     - /Users/Jaleh/ALL_MY_TESTS/test.log
   json: 
     message_key: message 
     keys_under_root: true 
     fields: 
       debug_app: Y 
     fields_under_root: true
processors: 
 - drop_event: 
    when: 
     contains: 
       tags: "ELK_TRACE"
output.file:
  path: /Users/Jaleh/ALL_MY_TESTS/
  filename: filebeat.test.json
```

And the warnings looks like before the event is published instead of being dropped. 

`2016/12/07 12:38:44.965301 condition.go:292: WARN unexpected type []interface {} in contains condition as it accepts only strings.
2016/12/07 12:38:44.965481 client.go:184: DBG  Publish: 
`

monai:
This ticket was opened two years ago, and the issue is still present in v6.5.3. Do you have any plans to address it?

PR #3199, opened 2016-12-15
Rework processors checkContains condition (#3138)

roylee17:
Handle potential string array (only at root level though).

If any entry contains the target string, return true.
Otherwise, continue the search.

When a invalid entry is found, emit a debugging message and
conitinue the search.

### DIFF
--- a/libbeat/common/event.go
+++ b/libbeat/common/event.go
@@ -109,6 +109,15 @@ func normalizeSlice(v reflect.Value, keys ...string) (interface{}, []error) {
 
 		sliceValues = append(sliceValues, sliceValue)
 	}
+	if sliceValues != nil && errs != nil {
+		if _, ok := sliceValues[0].(string); ok {
+			strSlices := make([]string, n)
+			for i := 0; i < n; i++ {
+				strSlices[i] = sliceValues[i].(string)
+			}
+			return strSlices, errs
+		}
+	}
 
 	return sliceValues, errs
 }


roylee17:
Okay, confirmed the email address of CLA.

roylee17:
Previous revision changes the logics and breaks the test cases.
This one only handles the issue reported by #3138 

Discussion:
Not sure this is the right place to "fix" it.
The Conditions assumes the input are valid (strings, []string), which makes perfect sense. And it's covered by the test as well.

It is the Json parser / processor's responsibility to handle the array.
In that case, either we need another pass to check the unmarshalled map for the array and convert/reject the input, or a modified Json decoder, which convert the array in the first place during the parsing.

Either way seems to be over kill for the corner case. Any thought?

andrewkroh:
@roylee17 Thanks for looking into the issue.

I'm thinking this conversion should take place in the normalization phase. Events that come into the publisher pipeline first go through "normalization" where by the `[]interface{}` passes through [here](https://github.com/elastic/beats/blob/master/libbeat/common/event.go#L168). Probably this code should catch this and fix the type. WDYT?

Another corner case I see is with the new `decode_json_fields` process that was added. The processors run after the normalization (they're the reason for the normalization). And because the `decode_json_fields` processor can add new data to the event, this new data could contain `[]interface{}` (I think, I didn't check the [code](https://github.com/elastic/beats/blob/master/libbeat/processors/actions/decode_json_fields.go)).

roylee17:
@andrewkroh Thanks for the tip. That makes much more sense. I'll get back with another attempt.

andrewkroh:
@roylee17 Could you please add some tests to accompany this change?

PR #3201, opened 2016-12-15
Community beat can set/get their own version

cyrilleverrier:
By default, community beats have the same version as libbeat.

Prerequisite
============

A community beat can set its own version by setting the following 2 variables in their Makefile:

Makefile
```
VERSION_GOFILEPATH=version.go
VERSION_YAMLFILEPATH=version.yml
```

main.go
```

package main

func main() {
    err := beat.Run("mybeat", appVersion, beater.New)
    if err != nil {
        os.Exit(1)
    }
}
```

Usage
=====

set version
-----------
the version number can be set with the following command:
     ```make set_version VERSION=0.1.0-alpha1```

1. if version.go and version.yml don't exist, the files will be created from a template

1. the script substitute the version numbers using a regular expression.
The files are updated only when there are no detected errors

1. when versions have been rewritten, make update is called

get version
-----------
the version number can be retrieved with the following command:
      ```make get_version```

Test
===
```
examplebeat $ make set_version VERSION=0.1.0-alpha1
./vendor/github.com/elastic/beats/dev-tools/set_version  --gofile version.go  --yamlfile version.yaml 0.1.0-alpha1
Create /home/cyrille/code/go/workspace/src/github.com/cyrille/examplebeat/version.go from template
Create /home/cyrille/code/go/workspace/src/github.com/cyrille/examplebeat/version.yaml from template
Write version 0.1.0-alpha1 in file /home/cyrille/code/go/workspace/src/github.com/cyrille/examplebeat/version.go
Write version 0.1.0-alpha1 in file /home/cyrille/code/go/workspace/src/github.com/cyrille/examplebeat/version.yaml

examplebeat $ make package
....

examplebeat $ ./build/examplebeat-linux-amd64 -version
examplebeat version 0.1.0-alpha1 (amd64), libbeat 1.1

examplebeat $ rpm -qip build/upload/examplebeat-0.1.0-alpha1-SNAPSHOT-i686.rpm
Name        : examplebeat
Version     : 0.1.0_alpha1_SNAPSHOT
```

### DIFF
--- a/dev-tools/set_version
+++ b/dev-tools/set_version
@@ -1,38 +1,136 @@
 #!/usr/bin/env python
-import os
 import argparse
+import fileinput
+import os
+import re
+import sys
 from subprocess import check_call
 
-template_go = '''package beat
+pattern = re.compile(r'(const\s|)(\w*)(v|V)(ersion)(\s*)(:|=)(\s)(")(.*)(")(.*)')
+repl    = r'\1\2\3\4\5\6\7"{}"\11'
 
-const defaultBeatVersion = "{}"
+
+goversion_template = '''package main
+
+const appVersion = "{version}"
 '''
 
-template_packer = '''version: "{version}"
+goversion_template_libbeat = '''package beat
+
+const defaultBeatVersion = "{version}"
 '''
 
+
+yamlversion_template = '''version: "{version}"
+'''
+
+def split_path(path):
+    allparts = []
+    while 1:
+        parts = os.path.split(path)
+        if parts[0] == path:  # sentinel for absolute paths
+            allparts.insert(0, parts[0])
+            break
+        elif parts[1] == path: # sentinel for relative paths
+            allparts.insert(0, parts[1])
+            break
+        else:
+            path = parts[0]
+            allparts.insert(0, parts[1])
+    return allparts
+
+def get_substituted_content(filename, pattern, replacement, errormsg):
+    nbr_changes = 0
+    new_content = ""
+    with open(filename, "r") as file:
+        for line in file.readlines():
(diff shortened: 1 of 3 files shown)


dev-tools/set_version
```diff
@@ -1,38 +1,136 @@
...
+            break
+        elif parts[1] == path: # sentinel for relative paths
+            allparts.insert(0, parts[1])
+            break
+        else:
+            path = parts[0]
+            allparts.insert(0, parts[1])
+    return allparts
+
+def get_substituted_content(filename, pattern, replacement, errormsg):
```
cyrilleverrier: The templates are only used to create the files if they don't exist.

When the version file exists, the script just replace the old version number by the new one using a regular expression in `get_substituted_content`. Thus, the community beat can have the version number in the `main.go` instead of `version.go`:

Also, the script is longer due to:

1. detection of libbeat/community beat
2. error handling
3. error messages

```
const myVersion = "0.1.0-alpha2"
func main() {
	err := beat.Run("examplebeat", myVersion, beater.New)
}
```

The regular expression detects for:
`const <prefix>Version = "xxxx"` (go file) or `version: "xxx"` (yml file)

tsg: But it would be simpler if we would mandate that the community beats must have a `version.go` file that we always overwrite, right? Is there any reason that wouldn't be a good idea?

cyrilleverrier: yes that's fine to me. It is less flexible but it works for me.

cyrilleverrier:
Simpler implementation: https://github.com/elastic/beats/pull/3203

PR #3202, opened 2016-12-15
Add Prometheus module to Metricbeat

ruflin:
This first version implements very basic stats from the prometheus service itself.


tsg:
LGTM.

PR #3206, opened 2016-12-15
Provide broker implementation in kafka module for reuse

urso:
In preparation for other metricset using a broker connection only, I moved the broker from partitions metricset to the kafka module.

### DIFF
--- a/metricbeat/module/kafka/broker.go
+++ b/metricbeat/module/kafka/broker.go
@@ -1,0 +1,374 @@
+package kafka
+
+import (
+	"bytes"
+	"crypto/tls"
+	"fmt"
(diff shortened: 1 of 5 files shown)


urso:
I wonder if/how we want to move the broker config from partitions/config.go to kafka module as well.

ruflin:
As far as I understand this change, there are no change the logic, it is only moving code to the module level, correct? For the config: Lets first get at least one more metricset in kafka before we apply the abstraction / extraction of config.go

urso:
yes, just moving code and wrapping `*sarama.Broker` into `kafka.Broker`.

PR #3213, opened 2016-12-16
Prometheus data collector

ruflin:
Generic data collector metricset for prometheus

Further improvements should be made to the data model, for example group same keys and handlers into one event.


tsg:
This is a good start.

On the data model, I think it would be better to have the field name as the metric name (without the labels), so aggregations can be applied across metrics that differ only by labels. In other words, transform the labels into Metricbeat fields.

I know that's more complicated because it means we need to split them in different documents, but should be doable.

otisg:
@ruflin @tsg What's the motivation for "monitoring the monitor" (Prometheus being the monitoring tool itself)?

tsg:
@otisg The Prometheus module contains two metricsets at the moment. One is the `status` which is indeed monitoring the Prometheus server (so monitoring the monitor). The other is the `collector` which can query any Prometheus exporter or any app that offers a Prometheus endpoint. In this case, you can use Metricbeat + Elasticsearch as an alternative to Prometheus.

metricbeat/_meta/fields.common.yml
```diff
@@ -23,6 +23,11 @@
       description: >
         Event round trip time in microseconds.
 
+    - name: metricset.namespace
```
tsg: Is this entry meant to be an example? Does it make sense to include it? Seems to me it will cause more confusion.

tsg: Ah, nevermind, it's the actual `namespace`.

PR #3214, opened 2016-12-16
Filebeat module system/integration tests

tsg:
The tests are taking all the logs from the modules `test` folders,
run filebeat on them (loading the template and the pipeline), then
do a search against Elasticsearch. On the result from Elasticsearch,
the test checks that all fields are documented.

In addition, the module writer can proved `-expected` files for each
log file, in which case we test the contents as well.

To run filebeat, the test is making use of the `-once` option, so each run is reasonably fast (usually under a second). The duration to test all the existing modules is 15 seconds at the moment.

This PR moves the filebeat system tests under Docker (if available) so that they have access to the testing environment. This broke one crawler test that was dealing with file permissions, because we're root under Docker which ignores permission. The "fix" is to skip the test if we detect we are root.

Part of #3159.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ env:
   matrix:
     - TARGETS="check"
     - TARGETS="-C filebeat testsuite"
+    - TARGETS="TEST_ENVIRONMENT=0 -C filebeat testsuite"
     - TARGETS="-C heartbeat testsuite"
     - TARGETS="-C libbeat testsuite"
     - TARGETS="-C metricbeat testsuite"
@@ -55,6 +56,10 @@ matrix:
       env: TARGETS="-C metricbeat testsuite"
     - os: linux
       env: TARGETS="TEST_ENVIRONMENT=0 -C metricbeat testsuite"
+    - os: osx
+      env: TARGETS="-C filebeat testsuite"
+    - os: linux
+      env: TARGETS="TEST_ENVIRONMENT=0 -C filebeat testsuite"
     - os: osx
       env: TARGETS="-C libbeat/dashboards"
     - os: osx
--- a/filebeat/_meta/fields.common.yml
+++ b/filebeat/_meta/fields.common.yml
@@ -38,7 +38,7 @@
         Ingestion pipeline error message, added in case there are errors reported by
         the Ingest Node in Elasticsearch.
 
-    - name: beat.read_timestamp
+    - name: read_timestamp
       description: >
         In case the ingest pipeline parses the timestamp from the log contents, it stores
         the original `@timestamp` (representing the time when the log line was read) in this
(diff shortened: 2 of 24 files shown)


filebeat/_meta/fields.common.yml
```diff
@@ -38,3 +38,9 @@
         Ingestion pipeline error message, added in case there are errors reported by
         the Ingest Node in Elasticsearch.
 
+    - name: beat.read_timestamp
```
ruflin: Not 100% sure about the location of this field as all the other fields in here are "beat" specific and rater constant. This one is event specific. But I also don't have a good alternative at the moment.

filebeat/filebeat.template-es2x.json
```diff
@@ -27,17 +27,7 @@
         },
         "beat": {
           "properties": {
-            "hostname": {
```
ruflin: Why did the hostname and name field disappear?

filebeat/_meta/fields.common.yml
tsg: I could put it top level, like the `error` field. WDYT?

filebeat/filebeat.template-es2x.json
tsg: Good catch. I suspect because of the definition of `read_timestamp`.

filebeat/_meta/fields.common.yml
tsg: Moved to top-level.

tsg:
Cleaned up history, so this can go in as a merge commit.

.travis.yml
```diff
@@ -56,6 +57,10 @@ matrix:
     - os: linux
       env: TARGETS="TEST_ENVIRONMENT=0 -C metricbeat testsuite"
     - os: osx
+      env: TARGETS="-C filebeat testsuite"
```
ruflin: Perhaps we could switch to `include` instead of exclude in the future as the exclude becomes larger and larger: https://docs.travis-ci.com/user/customizing-the-build#Explicitly-Including-Jobs This should also make it easier to read.

We can try this in a future PR.

filebeat/module/mysql/error/ingest/pipeline.json
```diff
@@ -5,7 +5,7 @@
         "field": "message",
         "patterns": [
           "%{LOCALDATETIME:mysql.error.timestamp} (\\[%{DATA:mysql.error.level}\\] )?%{GREEDYDATA:mysql.error.message}",
-          "%{DATA:mysql.error.timestamp} %{NUMBER:mysql.error.id} \\[%{DATA:mysql.error.level}\\] %{GREEDYDATA:mysql.error.message1}",
+          "%{DATA:mysql.error.timestamp} %{NUMBER:mysql.error.thread_id} \\[%{DATA:mysql.error.level}\\] %{GREEDYDATA:mysql.error.message1}",
```
ruflin: Does this need an update to the fields.yml file?

tsg: It was already called that in the fields.yml. Bug caught by the new tests.

PR #3216, opened 2016-12-17
 Edit haproxy module docs

dedemorton:
Edits for HAProxy docs (what's there...there are still a bunch of field descriptions that are missing).

The field descriptions in the original doc are ambiguous, so please check my edits carefully to make sure I haven't inadvertently changed the meaning. I couldn't figure out what to do with a couple of the descriptions:

- connection.errors:  The description of this field is: "Number of requests that encountered an error trying to connect to a backend server. The backend stat is the sum of the stat for all servers of that backend, plus any connection errors not associated with a particular server (such as the backend having no active servers)." Does this edit work? "Number of requests that encountered an error trying to connect to a  server. For backends, this field reports the number of connection errors for all backend servers, plus any connection errors not associated with a particular server (such as the backend having no active servers)."

- server.active: The description of this field is: "Number of active servers (backend), server is active (server)." So, if the component type is a server, what value does this field contain to indicate it's an active server?

- server.backup: The description of this field is: "Number of backup servers (backend), server is backup (server)." So, if the component type is server, what value does this field contain to indicate that it's a backup server? For this and the previous field, I'm making some assumptions about what the descriptions mean. It's pretty ambiguous in the HAProxy docs.

- I've added some extra line breaks to get the bulleted lists to format correctly, but for some reason, the fix doesn't work for the that's generated under haproxy.stat.response.errors. What the heck? How can I fix this? Is it OK for me to add the blank lines?


monicasarbu:
@dedemorton The description is exactly the one available under the [9.1 section](http://www.haproxy.org/download/1.6/doc/management.txt), and I will leave it as it is. I think what makes it unclear is that the description is not complete. For example, `server.active`:
```
19. act [..BS]: number of active servers (backend), server is active (server)
```
is available for backend and active servers and the value reported depends on the server type.

dedemorton:
@monicasarbu I understand that the value of `server.active` depends on whether the server is a backup server or active server. What's not clear to me is what the field contains when the server is an active server. Does the field contain  the text `server is active`, or does it contain a number?  I have the same question for `server.backup`. If the server is a backup server, does the field contain `server is backup` or some other value? 

Can you also comment on the edit that I proposed for `connection.errors`? The problem with the original text is that it makes it seem like the description only applies to backend servers, but I don't think that's true. : 

`````
13. econ [..BS]: number of requests that encountered an error trying to
     connect to a backend server. The backend stat is the sum of the stat
     for all servers of that backend, plus any connection errors not
     associated with a particular server (such as the backend having no
     active servers).
`````

My proposed change tries to clarify this ambiguity: "Number of requests that encountered an error trying to connect to a server. For backends, this field reports the number of connection errors for all backend servers, plus any connection errors not associated with a particular server (such as the backend having no active servers)."

monicasarbu:
@dedemorton Yes, the suggestion is good, just a small change:

Number of requests that encountered an error trying to connect to a server. For backends, this field reports `the sum of the stat for all backend servers`, plus any connection errors not associated with a particular server (such as the backend having no active servers)."

dedemorton:
@monicasarbu Thanks. I'll fix the description of `connection.errors`. I still need input on this question, tho, so that I can wrap this up.

What does the `server.active` field contain when the server is an active server? Does the field contain  the text `server is active`, or does it contain a number?  I have the same question for `server.backup`. If the server is a backup server, does the field contain `server is backup` or some other value?

PR #3217, opened 2016-12-19
Experimental go 1.8 plugin support

urso:
This PR introduces support for loadable plugins to beats. Requires linux, go1.8 beta2+, cgo. If any is missing, the API is still valid, but plugin loading is disabled (CLI flag is missing as well).

Shared libraries export a bundle of plugins in global `Bundle` variable of type `map[string][]interface{}`. The `string` key is the plugin module type, individual plugins are loadable for.

Beats modules supporting plugin support have to register a module type before startup via `plugin.RegisterLoader`. In addition modules must provide a `Plugin` function for building a valid type-correct bundle with correct module type key. See `outputs.Plugin` and `processors.Plugin` as example.
Alternatively a solution with `Bundle` of type `[]interface{}` or `interface{}` might be possible too (no potential naming collisions). Loading can use reflection and/or type switches. This might work, by passing all loadable plugins to all module loaders. With public API as is, this implementation detail can be easily applied later without enforcing any changes on plugin writers.

Every plugin is a valid bundle itself. Use `plugin.Bundle` to flatten and combine multiple bundles/plugins into one bundle.

Bundles can be loaded from command line only. Use `-plugin <path>` to load a plugin.

Versioning: the plugin loader does no explicit version checks. A possible solution is to have `plugin.Bundle` include a version and be strict (e.g. on major version). On the other hand, the compiler already hashes all packages. The runtime already ensures no plugin with failing hash-check can be used.

thisnPR adds plugin support for:
- libbout outputs
- libbeat processors
- packetbeat protocol analyzers
- heartbeat monitors
- metricbeat modules

### DIFF
--- a/libbeat/beat/beat.go
+++ b/libbeat/beat/beat.go
@@ -46,6 +46,7 @@ import (
 	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/logp"
 	"github.com/elastic/beats/libbeat/paths"
+	"github.com/elastic/beats/libbeat/plugin"
 	"github.com/elastic/beats/libbeat/processors"
 	"github.com/elastic/beats/libbeat/publisher"
 	svc "github.com/elastic/beats/libbeat/service"
@@ -152,6 +153,10 @@ func newBeat(name, version string) *Beat {
 func (b *Beat) launch(bt Creator) error {
 	err := b.handleFlags()
 	if err != nil {
+		return err
+	}
+
+	if err := plugin.Initialize(); err != nil {
 		return err
 	}
 
--- a/libbeat/processors/registry.go
+++ b/libbeat/processors/registry.go
@@ -1,9 +1,34 @@
 package processors
 
 import (
+	"errors"
+
 	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/logp"
+	p "github.com/elastic/beats/libbeat/plugin"
 )
+
+type processorPlugin struct {
+	name   string
+	constr Constructor
+}
+
+var pluginKey = "libbeat.processor"
+
+func Plugin(name string, c Constructor) map[string][]interface{} {
+	return p.MakePlugin(pluginKey, processorPlugin{name, c})
+}
+
+func init() {
+	p.MustRegisterLoader(pluginKey, func(ifc interface{}) error {
+		p, ok := ifc.(processorPlugin)
+		if !ok {
+			return errors.New("plugin does not match processor plugin type")
+		}
+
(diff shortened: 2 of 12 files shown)


tsg:
> On the other hand, the compiler already hashes all packages. The runtime already ensures no plugin with failing hash-check can be used.

That means that the exact same libbeat version needs to be used when compiling both the Beat and the plugin?

urso:
Right. Plus compiler version, as runtime is also included in plugin.

libbeat/beat/beat.go
```diff
@@ -155,6 +156,10 @@ func (b *Beat) launch(bt Creator) error {
 		return err
 	}
 
+	if err := plugin.Initialize(); err != nil {
```
ruflin: Can we "hide" this behind a config option something like `plugins.enabled: false`? This make sure nothing happens by default and will make it easier to mark this experimential. Especially as it is not supported on all platforms and only go 1.8.

libbeat/processors/registry.go
```diff
@@ -1,10 +1,39 @@
...
 	"github.com/elastic/beats/libbeat/logp"
+	p "github.com/elastic/beats/libbeat/plugin"
 )
 
+type processorPlugin struct {
+	name   string
+	constr Constructor
+}
+
+var pluginKey = "libbeat-processor"
```
ruflin: should we use the dot notation for the keys as we do in most other places? Means `libbeat.processor`

libbeat/beat/beat.go
urso: Plugins are loadable from CLI only using the `-plugin` setting. No config file support yet, as feature is very experimental.

Only if plugin support is available, the flag will be available. The Initialize function is implemented in [cli.go](https://github.com/elastic/beats/pull/3217/files#diff-9ba51f06e4b345b7121a0f8cd521c46bR33). In case no plugin support is available [cli_stub.go](https://github.com/elastic/beats/pull/3217/files#diff-87fbcf05836a787e64b638f81e89f252R5) will be used (empty function).

This behavior close enough?

libbeat/processors/registry.go
urso: It doesn't really matter, as the pluginKey is private to the package providing a plugin type. It's only required for them to be 'unique'. We could even use a generated UUID.

libbeat/beat/beat.go
ruflin: The behaviour is the same. It's just that with a undocument config option, not even the initialize code would be execute. I would still prefer to have an additional config option :-)

andrewkroh:
jenkins, test it

PR #3222, opened 2016-12-20
Remove global Kibana directory from metricbeat

ruflin:
These are copied fields from the modules. No need to check in this directory as it is generated.


ruflin:
jenkins, retest it

ruflin:
jenkins, retest it

PR #3223, opened 2016-12-20
Improve travis targets

ruflin:
* Explicitly define all targets
* Use environment variable for Golang version
* Update docker-compose version to 1.9.0
* Add os x targets for heartbeat
* Remove docker update as not needed anymore
* Add heartbeat coverage report

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,71 +5,80 @@ services:
 
 language: go
 
-go:
-  - 1.7.1
-
 # Make sure project can also be built on travis for clones of the repo
 go_import_path: github.com/elastic/beats
 
-os:
-  - linux
-  - osx
-
 env:
-  matrix:
-    - TARGETS="check"
-    - TARGETS="-C filebeat testsuite"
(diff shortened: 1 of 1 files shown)


ruflin:
@andrewkroh The jenkins builds fail now, as they rely on the entry in the travis file. I suggest we move the golang version the version.asciidoc file and read it from there.

ruflin:
jenkins, retest it

.travis.yml
```diff
@@ -5,71 +5,80 @@ services:
...
+    - &go_version 1.7.1
 
 matrix:
-  exclude:
-    - os: osx
+  include:
+    # General checks
+    - os: linux
       env: TARGETS="check"
+      go: *go_version
```
andrewkroh: If `go_version` is global how come it must be set everywhere in the matrix?

ruflin: Unfortunately this mix of global go version and include doesn't work :-( It will not read the global golang version.

PR #3228, opened 2016-12-20
Mongo dbstats

scottcrespo:
@ruflin 

Opening the PR early so we have a place to discuss implementation. At this point, I've copied the patterns used by the `status` metricset, but I have some questions and recommendations. Keep in mind, I'm still learning about the platform, so I might be making some incorrect assumptions!


metricbeat/module/mongodb/dbstats/dbstats.go
```diff
@@ -0,0 +1,90 @@
...
+	}, nil
+}
+
+// Fetch methods implements the data gathering and data conversion to the right format
+// It returns the event which is then forward to the output. In case of an error, a
+// descriptive error must be returned.
+func (m *MetricSet) Fetch() ([]common.MapStr, error) {
+
+	// establish connection to mongo
+	session, err := mgo.DialWithInfo(m.dialInfo)
```
scottcrespo: @ruflin 

This follows the pattern used by the `status` metricset [here](https://github.com/elastic/beats/blob/master/metricbeat/module/mongodb/status/status.go#L47), but I'm not sure it makes sense to open and close the mongo connection each time fetch() is called.

According to the [docs](https://www.elastic.co/guide/en/beats/metricbeat/current/metricset-details.html#_timeout_connections_to_services) connections should be established in the `New()` method and persist between fetches. 

I have two proposals:

1. Each metricset establishes its own set of connections for fetching data from mongo, and is implemented in New(). The advantage is that each metricset has its own set of connections which can be tailored for collecting the specific metric. The disadvantages are duplicative code and increased resource utilization (although minimal). 

2. In the mongodb package, implement a NewConnection() singleton that is available to both metricsets. The advantage here is reducing the number of connections and resource utilization. The disadvantage is that both metricsets will share the same connection parameters. Therefore, connection parameters cannot be tailored for each metricset individually. Nonetheless, I'm in favor of this method. 

Lastly, based on this implementation I'm more confident that the mongodb module is not connecting directly to the individual instances, but would instead connect to the cluster. This means that any cluster member may answer the query. This issue is a bit more involved, and I don't know enough about the platform to be sure, so I may need to jump on a Zoom call or something to talk further.


P.S. I'm going through such a proposal process as I'm just getting started in Beats. I'm interested in your thoughts and recommendations so I can better understand the philosophy and guiding principals of the platform. If you're preference is for me to dive right into implementation, I'm happy to do so!

ruflin: +1 on persisting the connection. It is probably best to extract the connection setup logic to the module level and just use it as a function.

About connecting to single instances or the cluster can be come tricky. We had a similar issue with Kafka. Our recommendation is to install metricbeat on each edge node (some exceptions exist here), as it gives you also additional metrics about the system etc. I'm don't know the details of the mongodb clustering but for Kafka we solved it the following way (generic description):

- Individual node stats are reported by each mongo instance
- Global stats about cluster / indices are only fetched from master

This requires that a metricbeat instance can detect if it is a master or not. Also I assume this must be dynamic, as master can change over time.

I assume dbStats response is the same for each node, so only the data from master should be fetched.

Happy to jump on a zoom call to go into more details.

scottcrespo: @ruflin 

## Connecting to single instances

Tricky indeed. Fortunately, mgo (the mongo client) has a `dialInfo.Direct` parameter. In mongobeat, I initiate a master connection to the cluster, and then create individual node connections to each node. You can see it [here](https://github.com/scottcrespo/mongobeat/blob/master/mongo/mongo.go)

I create the individual connections in the "controller" [here](https://github.com/scottcrespo/mongobeat/blob/master/beater/mongobeat.go#L64)

I've begun implementing something similar for the mongodb module, minus node discovery. This is implemented at the module level to share logic and prevent redundant connections.

## Detecting master/primary

Easily implemented with mgo. You can specify read preference to always read from primary. If the primary changes, your reads will automatically be directed to the new primary. 

## dbStats response the same

Technically dbStats response (and serverStatus) are not the same for each node. Sometimes the differences are trivial, other times they're not. I'll cover the cases where they're *not* the same:

**[replica sets](https://docs.mongodb.com/manual/replication/)**

In the case of a replica set, the state of the secondaries lag behind the primary. Thus, seeing the stats for each node helps monitor replication lag and success/failure. This is especially helpful in the case of multi-datacenter replication, where larger discrepancies are usually present.

**[sharded clusters](https://docs.mongodb.com/manual/core/sharded-cluster-shards/)**

In the case of a sharded cluster, the complexity is compounded because nodes contain different databases and collections, and a shard can also be a replica set! =o

## Conclusion

Given these considerations, I'm still having a hard time understanding what the best implementation for metricbeat would be. Given the current implementation - the node that is read from is implicit and can change from period to period. For example, you might connect to mongo on `localhost` and expect that you're reading from that node. But under the hood, mgo is actually discovering the other nodes in the cluster and directing reads to any one of the member nodes. Thus, I think it's possible to receive unexpected and inconsistent results. 

If we enforce that metricbeat only establishes a direct connection to a single mongo instance on localhost, then all of this complexity goes away and output is very predictable. Yay! But, it limits the flexibility of the system. 

If we allow the agent to connect to mongo as a cluster, and permit configuration of multiple hosts, then I think there's a lot of different directions we could take. 

Definitely a Zoom call will be helpful. I'm sure you can teach me a lot more about how metricbeat works under the hood, so I can understand the best direction.  

I don't have your email, but can you contact me at sccrespo@gmail.com and we'll set up a time? Perhaps we could talk sometime Thursday afternoon (your time).

ruflin: @scottcrespo Sorry, I was mostly offline the last two days, so only saw the update now. 

I always try to go with the simplest solution first, which I agree is the one only connecting to localhost. Because of docker I would probably rephrase this to: The mongodb module gets only the stats from the node it connects to, as it does not necessarly always have to be localhost on docker. 

Could you update the PR so that it always only connects to the defined host (and not the cluster)?

About zoom call: Lets see how far we can take it here in the PR as I'm rarely only in the next days because of holidays. But I'm happy to answer any questions about metricbeat in more detail here if needed.

scottcrespo: @ruflin 

No problem! I've been on vacation as well, but I have a few days to work on beats before taking some additional time off.

I'm in the process of updating the mongodb module to establish direct connections to the hosts listed in the config, and to persist those connections between calls. I should be able to update this PR within a few days!

scottcrespo:
@ruflin 

I'm having trouble getting the integration test suite to pass on Travis and Appveyor. 

I think I've narrowed down the problem to the fact that the mongodb module is not finding the environment variables `MONGODB_HOST` and `MONGODB_PORT`. Thus, the module is attempting to connect to Mongo on `127.0.0.1:27017`. 

[Relevant Travis Output](https://travis-ci.org/elastic/beats/jobs/187737758#L396)
[Relevant Appveyor Output](https://ci.appveyor.com/project/elastic-beats/beats/build/7327/job/ei1rmdf1n6quqsfw#L190)

I've verified in the [docker-compose file](https://github.com/scottcrespo/beats/blob/mongo-dbstats/metricbeat/docker-compose.yml#L24) that the environment variables are set. The suite works on my local machine and connects via `mongodb:27017`

I noticed that previously there was a [comment](https://github.com/elastic/beats/blob/master/metricbeat/module/mongodb/status/status_integration_test.go#L1) in the status_integration_test.go module that prevented the tests from running. What is the significance of that comment? I see other modules have it as well. 

Obviously, I wouldn't want to disable the integration tests if we can make them work. Do you have any insight?

scottcrespo:
@ruflin 

Ready for your review. Please let me know what changes I might need to make!

metricbeat/_meta/beat.full.yml
```diff
@@ -117,7 +117,7 @@ metricbeat.modules:
 
 #------------------------------- MongoDB Module ------------------------------
 #- module: mongodb
-  #metricsets: ["status"]
+  #metricsets: ["status", "dbstats"]
```
ruflin: We should probably sort this alphabetically. But not sure if we are consistent in other modules with that ...

metricbeat/module/mongodb/dbstats/_meta/data.json
```diff
@@ -0,0 +1,19 @@
...
+    },
+    "metricset":{
+        "host":"localhost",
+        "module":"mysql",
+        "name":"status",
+        "rtt":44269
+    },
+    "mongodb":{
+        "dbstats":{
+            "example": "dbstats"
```
ruflin: An example doc should be generated here with `make generate-json`. But that can still be done after merging the PR.

metricbeat/module/mongodb/dbstats/_meta/fields.yml
```diff
@@ -0,0 +1,56 @@
...
+    - name: indexes
+      type: long
+
+    - name: num_extents
+      type: long
+
+    - name: objects
+      type: long
+
+    - name: storage_size
```
ruflin: what is the unit for all the `_size` fields?

metricbeat/module/mongodb/dbstats/_meta/fields.yml
```diff
@@ -0,0 +1,56 @@
...
+    - name: data_size
+      type: long
+
+    - name: db
+      type: keyword
+
+    - name: file_size
+      type: long
+
+    - name: index_size
```
ruflin: this is probably `index_size.bytes` and then you can add `format: bytes` to each block, so Kibana will automatically now that it is in bytes and can do some conversion.

metricbeat/module/mongodb/dbstats/data.go
```diff
@@ -0,0 +1,34 @@
...
+	"collections":  c.Int("collections"),
+	"objects":      c.Int("objects"),
+	"avg_obj_size": c.Int("avgObjSize"),
+	"data_size":    c.Int("dataSize"),
+	"storage_size": c.Int("storageSize"),
+	"num_extents":  c.Int("numExtents"),
+	"indexes":      c.Int("indexes"),
+	"index_size":   c.Int("indexSize"),
+	// mmapv1 only
+	"ns_size_mb": c.Int("nsSizeMB"),
```
ruflin: Should this field be optional as based on the comment, I assume it does not always exist?

metricbeat/module/mongodb/dbstats/dbstats.go
```diff
@@ -0,0 +1,89 @@
+package dbstats
+
+import (
+	"errors"
+
+	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/logp"
+	"github.com/elastic/beats/metricbeat/mb"
+	"github.com/elastic/beats/metricbeat/module/mongodb"
+	"gopkg.in/mgo.v2"
```
ruflin: Minor detail: We normally put this in between the general and the beats imports, so it is:

```
	"errors"

	"gopkg.in/mgo.v2"
	
	"github.com/elastic/beats/metricbeat/mb"
	"github.com/elastic/beats/metricbeat/module/mongodb"
```

metricbeat/module/mongodb/dbstats/dbstats.go
```diff
@@ -0,0 +1,89 @@
...
+	mb.BaseMetricSet
+	dialInfo     *mgo.DialInfo
+	mongoSession *mgo.Session
+}
+
+// New creates a new instance of the MetricSet
+// Part of new is also setting up the configuration by processing additional
+// configuration entries if needed.
+func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
+	dialInfo, err := mgo.ParseURL(base.HostData().URI)
```
ruflin: Could you add an experimental flag here? We normally introduce new metricsets as experimental. This allows us to first get some real word feedback for it and still change the data structure if needed without having to wait for a major release. Have a look here: https://github.com/elastic/beats/blob/master/metricbeat/module/haproxy/stat/stat.go#L34

metricbeat/module/mongodb/mongodb.go
```diff
@@ -61,3 +65,24 @@ func ParseURL(module mb.Module, host string) (mb.HostData, error) {
...
+// dialInfo parameter as a template for establishing more direct connections
+func NewDirectSession(dialInfo *mgo.DialInfo) (*mgo.Session, error) {
+
+	// make a copy
+	nodeDialInfo := *dialInfo
+	nodeDialInfo.Direct = true
+	nodeDialInfo.FailFast = true
+
+	logp.Info("Connecting to MongoDB node at %v", nodeDialInfo.Addrs)
+	fmt.Printf("Connecting to MongoDB node at %v", nodeDialInfo.Addrs)
```
ruflin: The fmt should be removed here as we already log it.

metricbeat/module/mongodb/status/status.go
```diff
@@ -17,43 +15,52 @@ TODOs:
...
 
+	// instantiate direct connections to Mongo host
+	mongoSession, err := mongodb.NewDirectSession(dialInfo)
+	if err != nil {
+		return nil, err
+	}
+
 	return &MetricSet{
 		BaseMetricSet: base,
 		dialInfo:      dialInfo,
```
ruflin: Do we still need to store the dialInfo as we use now mongoSession?

metricbeat/module/mongodb/status/status.go
```diff
@@ -17,43 +15,52 @@ TODOs:
...
+// It returns the event which is then forward to the output. In case of an error, a
+// descriptive error must be returned.
 func (m *MetricSet) Fetch() (common.MapStr, error) {
-	session, err := mgo.DialWithInfo(m.dialInfo)
-	if err != nil {
-		return nil, err
-	}
-	defer session.Close()
-
-	session.SetMode(mgo.Monotonic, true)
```
ruflin: the monotonic mode was removed. Is this intentional?

metricbeat/module/mongodb/status/status.go
```diff
@@ -17,43 +15,52 @@ TODOs:
 	* add a metricset for "metrics" data
 */
 
-var debugf = logp.MakeDebug("mongodb.status")
```
ruflin: I assume you removed this, because it was not used. We normally keep this in to make it easy to add debug messages.

scottcrespo:
@ruflin

Thanks for all the feedback! I'm offline this week, but I'll implement your suggestions startkng next week!

metricbeat/module/mongodb/status/status.go
scottcrespo: @ruflin 

Yes, this was intentional. [Monotonic Mode](https://godoc.org/gopkg.in/mgo.v2#Mode) connects to the cluster as a whole and reads from the nearest secondary (not clear how nearest is defined). So, if your desired target is busy or goes down, mgo will read from a  different node in the cluster. This failover strategy is what I was referring to earlier (which I think is undesirable) - where you may read from a node other than the target specified in the module's configuration, and obfuscate node failures. 

Thus, the change I've implemented establishes [direct connections](https://godoc.org/gopkg.in/mgo.v2#DialInfo.Direct) to each of the nodes specified in the module's settings. You can see how I did it [here](https://github.com/scottcrespo/beats/blob/mongo-dbstats/metricbeat/module/mongodb/mongodb.go#L71). This ensures that we are reading from the intended node. And, if it's slow or fails, we will know about it. 

Also, note that I've also set [FailFast](https://godoc.org/gopkg.in/mgo.v2#DialInfo.FailFast) to `True`, which helps identity latent and non-responsive servers sooner rather than later.

scottcrespo:
@ruflin 

Implemented all of your suggestions. Ready for you to take another look =)

ruflin:
jenkins, test it

ruflin:
@scottcrespo LGTM. Could you add an update the CHANGELOG.asciidoc and squash the commits into 1?
@tsg Could you also have a quick look?

ruflin:
jenkins, test it

ruflin:
@scottcrespo Merged. Thanks a lot for this contribution.

scottcrespo:
@ruflin 

You're welcome! It's been great working with you on this. Thanks for all of the help and feedback.

PR #3234, opened 2016-12-21
Simplify xgo copying of artifacts

tsg:
Instead of trying to copy them after the fact, create them
directly under /build.


tsg:
jenkins, package it

tsg:
jenkins packaging tests are green.

PR #3235, opened 2016-12-21
Fix environment variable passing to system environment tests

ruflin:
Currently system tests with environment were still skipped.

* Fix flaky filebeat tests
* Fix flaky module tests
* Always rebuild dashboards. Reason is that during tests it is possible that it was previously built for a different architecture.


andrewkroh:
jenkins, retest it

ruflin:
@tsg Can you have a look at the failing test? Could be that this one slipped through because integration tests were not run properly.

ruflin:
The problem was, that filebeat was not part of `stop-environments` so the old elasticsearch instances without the plugin were still running.

PR #3239, opened 2016-12-22
Update Sarama (kafka client) to v1.11

urso:
- we used to use a fork of 1.10.1 + patches for fixing bugs in 1.10.1 release
- Bugs patched in our fork have all been fixed in 1.11
- still requires fork for setting replicaID in OffsetRequest

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -107,3 +107,7 @@ import:
   version: v0.9.0
 - package: github.com/gorhill/cronexpr
   version: f0984319b44273e83de132089ae42b1810f4933b
+- package: github.com/pierrec/lz4
+  version: 5c9560bfa9ace2bf86080bf40d46b34ae44604df
+- package: github.com/pierrec/xxHash
+  version: 5a004441f897722c627870a981d02b29924215fa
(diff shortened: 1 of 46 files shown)


glide.yaml
```diff
@@ -107,3 +107,7 @@ import:
   version: v0.9.0
 - package: github.com/gorhill/cronexpr
   version: f0984319b44273e83de132089ae42b1810f4933b
+- package: github.com/pierrec/lz4
```
ruflin: I assume these 2 new dependencies come from sarama 1.11? They have tons of binaries inside. Should we enhance our filter to filter out some?

ruflin:
Not sure if the API has changed. CI error:
```
--- FAIL: TestData (0.02s)
	partition_integration_test.go:100: Send Kafka Event to topic: metricbeat-generate-data
	partition_integration_test.go:110: kafka: invalid configuration (Producer.Return.Successes must be true to be used in a SyncProducer)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x484ea3]
```

glide.yaml
urso: yeah, new sarama dependencies (again...)

PR #3240, opened 2016-12-22
Metricset/kafka consumer groups

urso:
Requires #3239


metricbeat/module/kafka/broker.go
```diff
@@ -140,6 +161,82 @@ func (b *Broker) PartitionOffset(
...
+	}
+	return groups, nil
+}
+
+// DescribeGroups fetches group details from broker.
+func (b *Broker) DescribeGroups(
+	queryGroups []string,
+) (map[string]GroupDescription, error) {
+	requ := &sarama.DescribeGroupsRequest{Groups: queryGroups}
+	resp, err := b.b.DescribeGroups(requ)
```
ruflin: b.b. is not very readable ;-)

metricbeat/module/kafka/consumergroup/_meta/data.json
```diff
@@ -0,0 +1,19 @@
...
+    },
+    "metricset":{
+        "host":"localhost",
+        "module":"mysql",
+        "name":"status",
+        "rtt":44269
+    },
+    "kafka":{
+        "consumergroup":{
+            "example": "consumergroup"
```
ruflin: We should populate this for the docs.

metricbeat/module/kafka/consumergroup/_meta/fields.yml
```diff
@@ -0,0 +1,62 @@
...
+          type: long
+          description: >
+            Broker id
+
+        - name: address
+          type: keyword
+          description: >
+            Broker address
+
+    - name: group
```
ruflin: This should be `id`, as it will be `kafka.consumegroup.id`. Alternative is `kafka.conumser.group`. Is the partition, topic, offset specific to a consumergroup or a consumer?

metricbeat/module/kafka/consumergroup/config.go
```diff
@@ -0,0 +1,37 @@
...
+	TLS      *outputs.TLSConfig `config:"ssl"`
+	Username string             `config:"username"`
+	Password string             `config:"password"`
+	ClientID string             `config:"client_id"`
+
+	Groups []string `config:"groups"`
+	Topics []string `config:"topics"`
+}
+
+var defaultConfig = metricsetConfig{
```
ruflin: We should move this to the kafka module later as I think it is the same for both metricsets.

metricbeat/module/kafka/consumergroup/consumergroup.go
```diff
@@ -0,0 +1,100 @@
+package consumergroup
+
+import (
+	"crypto/tls"
+
+	"github.com/Shopify/sarama"
```
ruflin: missing newline

metricbeat/module/kafka/consumergroup/consumergroup.go
```diff
@@ -0,0 +1,100 @@
...
+	groups nameSet
+}
+
+type groupAssignment struct {
+	clientID   string
+	memberID   string
+	clientHost string
+}
+
+var debugf = logp.MakeDebug("kafka")
```
ruflin: kafka.consumergroup?

metricbeat/module/kafka/consumergroup/mock_test.go
```diff
@@ -0,0 +1,129 @@
+package consumergroup
+
+import (
+	"fmt"
+	"math/rand"
+
+	"github.com/Shopify/sarama"
```
ruflin: newline

metricbeat/module/kafka/consumergroup/query.go
```diff
@@ -0,0 +1,210 @@
...
+	if len(groups) == 0 {
+		return nil
+	}
+
+	debugf("known consumer groups: ", groups)
+
+	wg := sync.WaitGroup{}
+	results := make(chan result, len(groups))
+	for _, group := range groups {
+		group := group
```
ruflin: copy needed?

metricbeat/module/kafka/consumergroup/query.go
```diff
@@ -0,0 +1,210 @@
...
+	resp, err := b.DescribeGroups(groupIDs)
+	if err != nil {
+		return nil, err
+	}
+
+	groups := map[string]map[string]map[int32]groupAssignment{}
+
+groupLoop:
+	for groupID, info := range resp {
+		G := groups[groupID]
```
ruflin: upper case?

metricbeat/module/kafka/consumergroup/query.go
```diff
@@ -0,0 +1,210 @@
...
+			}
+
+			meta := groupAssignment{
+				memberID:   memberID,
+				clientID:   clientID,
+				clientHost: clientHost,
+			}
+
+			for topic, partitions := range memberDescr.Topics {
+				T := G[topic]
```
ruflin: upper case?

metricbeat/module/kafka/consumergroup/_meta/fields.yml
urso: will rename it to `id`

metricbeat/module/kafka/consumergroup/consumergroup.go
urso: partitions is using `kafka` as well. we should clean this in a followup PR. Do we follow a debug selector naming standard in metricbeat?

metricbeat/module/kafka/consumergroup/query.go
urso: Yes, as `group` is used inside the go-routine. Copy is used to capture the `group` in the local scope.

Alternatively one can write:

```
for group := range groups {
  go func(group string) {
   ...
  }(group)
}
```

which is similar to writing:

```
for group := range groups {
  group := group
  go func() {
   ...
  }()
}
```

metricbeat/module/kafka/consumergroup/query.go
urso: I tend to use uppercase for local sets by convention (e.g. like N,M,L for length). Here G for group and T for topic. So not to shadow (overuse) `topic` and `group` used for correlating information. The 'sets' G and T hold results, so using a single upper-case makes them kinda stick out in this 'heavy' (potentially hard to follow) function.

PR #3242, opened 2016-12-23
Added cloudflarebeat to list of communitybeats

hartfordfive:
I think this beat can now be included in the list although keep in mind it's still in beta.  I'd appreciate any feedback on it.


ruflin:
Thanks for adding the beat. I had a quick look at the beat. In general it looks good. One thing I'm curious about is how you generate the template? It seems like your fields.yml is empty but you still have fields in the template. So I assume you are not using the auto generation with `make update`?

hartfordfive:
@ruflin  Thanks for pointing that out. I actualy forgot to complete the `fields.yml` file and built the templates manually.  I've created an [issue](https://github.com/hartfordfive/cloudflarebeat/issues/12) to have that fixed in my next release.

ruflin:
Cool. Let me know in case you hit some issues with `make update` (hopefully not).

PR #3246, opened 2016-12-23
Add system socket MetricSet

andrewkroh:
The system.socket metricset reports an event for each new TCP socket that it
sees. It does this by polling the kernel to get a dump of all sockets.
So using a short polling interval with this metricset is important to
not miss short lived connections.

The metricset reports the process that has the socket open. It does this
by associating the socket's inode to the process that has a file
descriptor open pointing to the socket's inode. It reads /proc and
/proc/<pid>/fd just prior to polling the kernel to get all sockets.

A reverse lookup can be performed by the metricset on the remote IP and the
returned hostname will be added to the event and cached. The is disabled by
default and can be enabled through the configuration.  If a hostname is found
then the eTLD+1 (effective top-level domain plus one level) value will also be
added to the event.

For the IP address fields the index template for Elasticsearch 5.x uses
the ip field type. But for Elasticsearch 2.x it uses string because the
ip field type in 2.x does not support IPv6 addresses.


ruflin:
Don't forget the CHANGELOG ;-)

andrewkroh:
I added a changelog entry and modified the mapping for ES 5.x to use `ip` instead of `keyword` for the IP address fields.

andrewkroh:
Fixes #3257

metricbeat/module/system/socket/socket.go
```diff
@@ -0,0 +1,294 @@
...
+	if !ok {
+		return nil, errors.New("unexpected module type")
+	}
+
+	ptable, err := NewProcTable(filepath.Join(systemModule.HostFS, "/proc"))
+	if err != nil {
+		return nil, err
+	}
+	if os.Geteuid() != 0 {
+		debugf("socket process info will only be available for metricbeat " +
```
tsg: This would make sense as a warning, maybe?

metricbeat/module/system/socket/config.go
```diff
@@ -0,0 +1,28 @@
...
+// DNS lookup behavior.
+type ReverseLookupConfig struct {
+	Enabled    *bool         `config:"enabled"`
+	SuccessTTL time.Duration `config:"success_ttl"`
+	FailureTTL time.Duration `config:"failure_ttl"`
+}
+
+// IsEnabled returns true if enabled is not defined or is set to true.
+func (c ReverseLookupConfig) IsEnabled() bool {
+	return c.Enabled == nil || *c.Enabled == true
```
tsg: Does this mean that by default reverse lookup is enabled, right? I'm not sure about enabling that by default since it can cause performance issues, especially since there's no configurable timeout on the lookup, right?

metricbeat/module/system/socket/socket.go
andrewkroh: How about INFO?

metricbeat/module/system/socket/config.go
andrewkroh: Correct, there is no configurable lookup timeout and the requests are executed serially. It's not an optimal implementation and it can slow down fetching. I'll change it to disabled by default.

metricbeat/module/system/socket/socket.go
tsg: Deal :)

andrewkroh:
Rebased. I recommend using the "Rebase and merge" option which will add my 3 separate commits, but will not create a merge commit.

PR #3247, opened 2016-12-24
Add docs for heartbeat

dedemorton:
This is the first cut of the Heartbeat documentation. I'm marking it as 'in progress' because it needs work, but it wouldn't hurt to merge this into master and get it building  if you think it's good enough.

A few things:
- Right now, all the Heartbeat config options in one topic. We should decide if we want to set up scripts like @ruflin did for modules in Metricbeat to generate docs for the monitor-related options. I'm not sure if that's necessary.
- The examples are just placeholders (my best guess). Please suggest better examples when it makes sense.
- For config options, I've tried to mention defaults, but I'm not sure I've gotten it completely right. Also, it's not always clear which options are required. 
- I included all the boiler plate stuff that I thought might be relevant to Heartbeat. You'll have to tell me which topics, if any, to remove.
- Questions and to-do items are buried in the source as comments.
- I've included placeholder files for some stuff that we want to add later. Those are commented out in the index.asciidoc.


heartbeat/docs/overview.asciidoc
```diff
@@ -0,0 +1,32 @@
+[[heartbeat-overview]]
+== Overview
+
+//TODO: Add Beta tag, when available, to generate the labels indicating that this is a Beta feature.
+
+Heartbeat is a lightweight daemon that you install on a remote server
+to ping your services periodically and determine whether they are available.
+Unlike {metricbeat}/index.html[Metricbeat], which only tells you if your servers
+are up or down, Heartbeat tells you whether your services are reachable. If
```
dedemorton: 2nd paragraph should probably  appear here and then be followed by the bit about load balancing.

heartbeat/docs/getting-started.asciidoc
```diff
@@ -0,0 +1,261 @@
...
+options, see <<heartbeat-configuration-details>>.
+
+[[heartbeat-configuration]]
+=== Step 2: Configuring Heartbeat
+
+To configure Heartbeat, you edit the configuration file. For rpm and deb,
+you'll find the configuration file at +/etc/heartbeat/heartbeat.yml+.
+For mac and win, look in the archive that you just extracted.
+
+Heartbeat uses monitors to configure ping behavior. You configure each
```
dedemorton: Should probably say, "Heartbeat uses modules called monitors to configure..."

heartbeat/docs/reference/configuration/heartbeat-options.asciidoc
```diff
@@ -0,0 +1,448 @@
...
+
+A Boolean value that specifies whether to resolve hostnames locally instead of
+resolving them on the SOCKS5 server. The default is false.
+
+[[monitor-tcp-tls-ssl]]
+===== ssl
+
+The TLS/SSL connection settings. 
+
+//Steffen: Is this option so that you can check to see if the servers certificates are valid?
```
urso: ssl is this one: https://github.com/elastic/beats/blob/master/libbeat/docs/outputconfig.asciidoc#ssl-configuration

if SSL/TLS is enabled the monitor will attempt an SSL handshake in addition. If `check` is not configured, the monitor will check only wether an SSL/TLS connection can be establied. This can fail either at TCP level or during certificate validation.

heartbeat/docs/reference/configuration/heartbeat-options.asciidoc
```diff
@@ -0,0 +1,448 @@
...
+  ipv6: true
+  mode: any
+  timeout: 16s
+  ssl:
+    certificate_authorities: ['/etc/ca.crt']
+    supported_protocols: ["TLSv1.0", "TLSv1.1", "TLSv1.2"]
+-------------------------------------------------------------------------------
+
+
+//Steffen: How are the values for supported_protocols used?
```
urso: it's this one: https://github.com/elastic/beats/blob/master/libbeat/docs/outputconfig.asciidoc#versions

Makes me wonder which one is correct. Docs or sample config? Found it, doc says `version, but correct is [supported_protocols](https://github.com/elastic/beats/blob/master/libbeat/outputs/tls.go#L33)

heartbeat/docs/reference/configuration/heartbeat-options.asciidoc
```diff
@@ -0,0 +1,448 @@
...
+===== username
+
+The username for authenticating with the server. This setting is optional. 
+
+[[monitor-http-password]]
+===== password
+
+The password for authenticating with the server. This setting is optional.
+
+//Steffen: When are authentication credential required? If you are sending a request and expecting a response?
```
urso: It depends on `check.response...` settings. I can for example not set any credentials and use `check.response.status: 403`.

The credentials are passed with the request.

heartbeat/docs/reference/configuration/heartbeat-options.asciidoc
```diff
@@ -0,0 +1,448 @@
...
+specify settings, the system defaults are used. 
+
+
+Example configuration:
+
+[source,yaml]
+-------------------------------------------------------------------------------
+- type: http 
+  schedule: '@every 5s'
+  urls: ["http://myhost:80"]
```
urso: using `http` here makes no use of the ssl settings.

heartbeat/docs/reference/configuration/heartbeat-options.asciidoc
```diff
@@ -0,0 +1,448 @@
...
+    headers:
+    body:
+  check.response:
+    status: 0
+    headers:
+    body:
+-------------------------------------------------------------------------------
+
+
+//Steffen: I need help filling out the above example with headers and body
```
urso: how about something very simple:

```
check.request.method: HEAD
check.response.status: 200
```

These checks are very specific to a service.

e.g. some check with JSON response:

```
check.request:
  method: GET
  headers:
    'X-API-Key': '12345-mykey-67890'
check.response:
  status: 200
  body: '{"status": "ok"}'
```

heartbeat/docs/reference/configuration/heartbeat-options.asciidoc
```diff
@@ -0,0 +1,448 @@
...
+
+[source,yaml]
+-------------------------------------------------------------------------------
+heartbeat.scheduler:
+  limit: 0
+  location: 'UTC-08:00'
+-------------------------------------------------------------------------------
+
+
+//Steffen: What is a reasonable limit to set in the example? Also, I just sort of guessed at what would be valid for location in my example.
```
urso: location is mostly helpful if a monitors schedule includes hours, in order to sync up multiple hearbeat instances in different locations (this feature is so not tested :/ ).

`limit` is for limiting the number of concurrent I/O tasks. e.g. most OSes/distros set an file descriptor limit of 1024. So for heartbeat to operate correctly and not accidentaly block libbeat output, `limit` should be below the configured ulimit.

heartbeat/docs/reference/configuration/heartbeat-options.asciidoc
```diff
@@ -0,0 +1,448 @@
...
+
+
+[[heartbeat-scheduler-limit]]
+===== limit
+
+The number of concurrent tasks that Heartbeat is allowed to execute. If set
```
urso: mostly I/O tasks. An tcp/http I/O tasks normally requires a file descriptor (limited OS resource).

heartbeat/docs/reference/configuration/heartbeat-options.asciidoc
```diff
@@ -0,0 +1,448 @@
...
+to 0, there is no limit. The default is 0. 
+
+[[heartbeat-scheduler-location]]
+===== location
+
+The timezone for the scheduler.
+
+//Steffen: Is the location a required setting? If not, what timezone does the scheduler default to?
```
urso: it's optional. By default it uses localtime. It's mostly useful if heartbeat runs on a machine in another timezone, but should respect the of the service running in `scheduler.location` it's timezone. This is only effective if the monitors schedule uses the `hours` setting. With most monitors having configured an periodic interval in seconds, minutes, this options is of no use.

heartbeat/docs/securing-heartbeat.asciidoc
```diff
@@ -0,0 +1,23 @@
...
+* <<securing-communication-elasticsearch>>
+* <<configuring-ssl-logstash>>
+
+--
+
+[[securing-communication-elasticsearch]]
+== Securing Communication With Elasticsearch
+include::../../libbeat/docs/https.asciidoc[]
+
+//ASK STEFFEN IF LOGSTASH EVEN MAKES SENSE WITH THIS BEAT
```
urso: one might use logstash with any beat for different reasons. Like event-routing, some other output, hide elasticsearch, do geo-ip in logstash (to figure out in which data-center service X is currently running for example)?

heartbeat/docs/getting-started.asciidoc
```diff
@@ -0,0 +1,261 @@
...
+
+[[heartbeat-installation]]
+=== Step 1: Installing Heartbeat
+
+Unlike most Beats, which you install on edge nodes, you typically install
+Heartbeat as part of monitoring service that runs on a separate machine
+and possibly even outside of the network where the services that you want to
+monitor are running. 
+
+//Steffen: I wonder if we need a topic that explores deployment scenarios in more detail (like installing on a sub-network where there's a filewall etc. I don't want to complicate the getting started with this level of details. WDYT?
```
dedemorton: Ok, I've added it to my list of future Beats enhancements just so I don't forget to add it later.

heartbeat/docs/getting-started.asciidoc
dedemorton: @urso Will we also have "passive monitors" at some point (or maybe there's a different word)? At this point, I'm wondering if it makes sense to mention that we only support active monitors. WDYT?

heartbeat/docs/getting-started.asciidoc
```diff
@@ -0,0 +1,261 @@
...
+  ipv4: true
+  ipv6: true
+  mode: any
+  timeout: 16s
+  wait: 1s
+heartbeat.scheduler:
+  location: 'UTC-08:00'
+output.elasticsearch:
+  hosts: ["myhost:9200"]
+----------------------------------------------------------------------
```
dedemorton: +++ fixed the example

heartbeat/docs/getting-started.asciidoc
```diff
@@ -0,0 +1,261 @@
...
+  hosts: ["myhost:12345"]
+  ipv4: true
+  ipv6: true
+  mode: any <3>
+  timeout: 16s
+heartbeat.scheduler:
+  location: 'UTC-08:00' <4>
+output.elasticsearch:
+  hosts: ["myhost:9200"]
+----------------------------------------------------------------------
```
dedemorton: @urso Are you suggesting that we should show the `limit` setting in the example here? Instead of location? In addition to location? (keep in mind that this is the getting started content so we don't have to show all interesting options)

heartbeat/docs/getting-started.asciidoc
```diff
@@ -0,0 +1,261 @@
...
+<1> The `icmp` monitor is scheduled to run exactly every 5 seconds (10:00:00,
```
dedemorton: Added a bit of info to the notes in the example to clarify this.

heartbeat/docs/getting-started.asciidoc
```diff
@@ -0,0 +1,261 @@
...
+----------------------------------------------------------------------
+
+By default, Windows log files are stored in +C:\ProgramData\heartbeat\Logs+.
+
+Heartbeat is now ready to send log files to your defined output.
+
+[[heartbeat-sample-dashboards]]
+=== Step 5: Loading Sample Kibana Dashboards
+
+//Steffen: Will we have Heartbeat dashboards for 5.2?
```
dedemorton: Ok, I'll remove the section about dashboards for now and add a note to remind us to add it back in later.

heartbeat/docs/heartbeat-devguide.asciidoc
```diff
@@ -0,0 +1,27 @@
...
+
+Heartbeat pings your remote services periodically and determines whether they
+are available. As a developer, you can use Heartbeat in two different ways:
+
+* Extend Heartbeat directly
+* Create your own Beat and use Heartbeat as a library
+
+We recommend that you start by creating your own Beat to keep the development of
+your own monintors independent of Heartbeat. At a later stage, if you decide to
+add a monitor to Metricbeat, you can reuse the code without making additional changes.
```
dedemorton: Fixed.

heartbeat/docs/heartbeat-devguide.asciidoc
```diff
@@ -0,0 +1,27 @@
...
+
+The following topics describe how to contribute to Heartbeat by adding new monitors
+and new Beats based on Heartbeat:
+
+--
+
+[[metricbeat-dev-overview]]
```
dedemorton: fixed

heartbeat/docs/overview.asciidoc
dedemorton: fixed

heartbeat/docs/overview.asciidoc
```diff
@@ -0,0 +1,32 @@
...
+you run your services using a DNS-based load balancer or proxy, you can configure
```
dedemorton: fixed. still says ping but I think that's OK in the context of this particular sentence.

heartbeat/docs/reference/configuration/heartbeat-options.asciidoc
```diff
@@ -0,0 +1,448 @@
+[[configuration-heartbeat-options]]
+=== Heartbeat Configuration
+
+The `heartbeat` section of the +heartbeat.yml+ config file specifies the
```
dedemorton: Yah maybe, but I should do this across the docs, not just for Heartbeat. I'll create a separate issue to track this.

heartbeat/docs/reference/configuration/heartbeat-options.asciidoc
```diff
@@ -0,0 +1,448 @@
...
+list of `monitors` that Heartbeat uses to ping your remote hosts to
```
dedemorton: changed to check

heartbeat/docs/reference/configuration/heartbeat-options.asciidoc
```diff
@@ -0,0 +1,448 @@
...
+  hosts: ["myhost:12345"]
+  ipv4: true
+  ipv6: true
+  mode: any
+  timeout: 16s
+heartbeat.scheduler:
+  location: 'UTC-08:00'
+output.elasticsearch:
+  hosts: ["myhost:9200"]
+----------------------------------------------------------------------
```
dedemorton: Fixed, but I'd like to leave the `heartbeat.scheduler` section because I want users to see it in full example so they don't try to nest it or anything. You'd think the dot notation would make that clear, but I've seen things....

heartbeat/docs/reference/configuration/heartbeat-options.asciidoc
```diff
@@ -0,0 +1,448 @@
...
+===== schedule
+
+A cron-like expression that specifies the task schedule. For example:
+
+* `*/5 * * * * * *` runs the task every 5 seconds (for example, at 10:00:00,
+10:00:05, and so on).
+* `@every 5s` runs the task every 5 seconds from the time when Heartbeat was
+started.
+
+//Steffen: Should we document the syntax of the cron expression somewhere, or point to a resource if one is available that describes specifically what the scheduler supports
```
dedemorton: added...let me know if we start using an different implementation. :-)

heartbeat/docs/reference/configuration/heartbeat-options.asciidoc
```diff
@@ -0,0 +1,448 @@
...
+`mode: all` setting is useful if you are using a DNS-load balancer and want to
+ping every IP address for the specified hostname. The default is `any`.
+
+[[monitor-timieout]]
+===== timeout
+
+The total running time for each ping test. This is the total time allowed for
+testing the connection and exchanging data. The default is 16 seconds (16s).
+
+//Steffen: What happens if this timeout is exceeded?
```
dedemorton: added

heartbeat/docs/reference/configuration/heartbeat-options.asciidoc
```diff
@@ -0,0 +1,448 @@
...
+Heartbeat checks this file periodically and starts a new monitor for any new
+objects that are added to the file. For example, imagine that you add 10
+new entries to the JSON file, each for a different hostname. When Heartbeat
+picks up the changes in the file, it starts a monitor for each host that you've
+configured. In this way, the configuration that you specify for the monitor in
+the main Heartbeat config file acts like a default config that you can
+live-reconfigure by specifying additional configurations in the external JSON
+file. 
+
+//Steffen: Not sure I got that quite right, but there you go! BTW, if you delete an object from the JSON file, I guess it stops the monitor that it has running for that object, right?
```
dedemorton: @urso So what if the JSON file contains a config for a host that you're already monitoring in the main config file? Do the settings in the JSON file override what you've specified in the Heartbeat config? In other words, can you use the JSON file to override settings in the main config file, or just to add configs? I'm assuming the latter (because you probably want to allow users to set up multiple monitors for the same host), but I'm not sure.

heartbeat/docs/reference/configuration/heartbeat-options.asciidoc
```diff
@@ -0,0 +1,448 @@
...
+
+//TODO: We have to document the structure of the JSON somewhere.
```
dedemorton: @urso I'm not following what you are saying here. Maybe it would help for me to see the format of this file. Is there one available somewhere that I can look at?

heartbeat/docs/reference/configuration/heartbeat-options.asciidoc
```diff
@@ -0,0 +1,448 @@
...
+  mode: any
+  timeout: 16s
+  wait: 1s
+  watch.poll_file:
+    path: /monitors/myfile.json
+    interval: 5s
+-------------------------------------------------------------------------------
+
+
+//Steffen: Can you suggest a realistic path? I have no idea where users would keep these files.  I'm not sure if users specif a glob pattern here or the path to a single file.
```
dedemorton: fixed

heartbeat/docs/reference/configuration/heartbeat-options.asciidoc
```diff
@@ -0,0 +1,448 @@
...
+
+[[monitor-tcp-hosts]]
+===== hosts
+
+A list of hosts to ping. The entries in the list can be:
+
+* A plain host name, such as `localhost`, or an IP address. If you specify this
+option, you must also specify a value for <<monitor-tcp-ports,`ports`>>.  If the
+monitor is configured to use SSL, Heartbeat establishes an SSL/TLS-based
+connection. Otherwise, it establishes a plain TCP connection.
```
dedemorton: done

heartbeat/docs/reference/configuration/heartbeat-options.asciidoc
```diff
@@ -0,0 +1,448 @@
...
+* A hostname and port, such as `localhost:12345`. Heartbeat connects
+to the port on the specified host. If the monitor is configured to use SSL,
+Heartbeat establishes an SSL/TLS-based connection. Otherwise, it establishes
+a TCP connection.
+* A full URL using the syntax `scheme://<host>:[port]`, where:
+** `scheme` is one of `tcp`, `plain`, `ssl` or `tls`. If `tcp` or `plain` is
+specified, Heartbeat establishes a TCP connection even if the monitor is
+configured to use SSL. If `tls` or `ssl` is specified, Heartbeat establishes
+an SSL connection. However, if the monitor is not configured to use SSL, the
+system defaults are used (not supported on Windows).
```
dedemorton: Ok, I'll add that to my list of things to track.

heartbeat/docs/reference/configuration/heartbeat-options.asciidoc
```diff
@@ -0,0 +1,448 @@
...
+** `host` is the hostname.
+** `port` is the port number. If `port` is missing in the URL, the
+<<monitor-tcp-ports,`ports`>> setting is required.
+
+// Questions: Can port above be a list of ports, or is it only a single port.
```
dedemorton: thanks for the clarification

heartbeat/docs/reference/configuration/heartbeat-options.asciidoc
```diff
@@ -0,0 +1,448 @@
...
+-------------------------------------------------------------------------------
+
+[[monitor-tcp-check]]
+===== check
+
+An optional payload string to send to the remote host and the expected answer.
+If no payload is specified, the endpoint is assumed to be available if the
+connection attempt was successful. If `send` is specified without `receive`,
+any response is accepted as OK. If `receive` is specified without `send`, no
+payload is sent, but the client expects to receive a payload on connect.
```
dedemorton: added

dedemorton: changed this to say (let me know if this isn't what you meant):

....but the client expects to receive a payload in the form of a "hello message" or "banner" on connect.

dedemorton: @urso Or did you mean to say here that the optional payload that's sent is a "hello message" or "banner"?

heartbeat/docs/reference/configuration/heartbeat-options.asciidoc
```diff
@@ -0,0 +1,448 @@
...
+  ipv6: true
+  mode: any
+  timeout: 16s
+  check:
+    send: ''
+    receive: ''
+-------------------------------------------------------------------------------
+
+
+//Steffen: Can you provide a good example of the values to set for send and receive in the above example?
```
dedemorton: fixed

heartbeat/docs/reference/configuration/heartbeat-options.asciidoc
dedemorton: added

heartbeat/docs/reference/configuration/heartbeat-options.asciidoc
dedemorton: @urso Do you want the example changed? I show the ssl options in a later example.

heartbeat/docs/reference/configuration/heartbeat-options.asciidoc
dedemorton: added

heartbeat/docs/reference/configuration/heartbeat-options.asciidoc
dedemorton: added

heartbeat/docs/getting-started.asciidoc
urso: Well, my idea is to introduce passive monitors... at some point :)

Just using 'monitors' should be ok. I wonder if people would expect to have passive/active monitors if we just use monitors.

heartbeat/docs/getting-started.asciidoc
urso: for getting started only I wouldn't show `heartbeat.schedule` section at all. It's a more advanced setting.

heartbeat/docs/reference/configuration/heartbeat-options.asciidoc
urso: Yeah, feel free to include

```
heartbeat.scheduler:
  limit: 10
```

setting limit to 10 guarantees only 10 concurrent IO-tasks being active. An IO-task can be the actual ping or resolving an address via DNS.

heartbeat/docs/reference/configuration/heartbeat-options.asciidoc
urso: the JSON object can overwrite all settings in the monitor configured. All settings but `watch`, `type`, `enabled`.

The JSON file can contain multiple JSON objects. For every JSON object one monitor instance is created.

All monitors have an unique ID based on some parameters (like protocol, host, port). If 2 monitors have the same ID (same host being configured by accident), only the last monitor (JSON object) defined will be instantiated.

heartbeat/docs/reference/configuration/heartbeat-options.asciidoc
urso: No example. It's related to 'add' vs. 'overwrite'. (see last comment).

What happens in code is, we merge the original configuration + the JSON object into a new configuration. The JSON object can overwrite any setting in the original configuration. The merged configuration is used to instantiate a monitor.

heartbeat/docs/reference/configuration/heartbeat-options.asciidoc
urso: "hello message" or "banner" implies the server sending a message right when the client connects. That is, it's related `send` not configured, but `received` setting being set.

heartbeat/docs/reference/configuration/heartbeat-options.asciidoc
urso: you can simply change `http` to `https` and the example is all fine.

heartbeat/docs/getting-started.asciidoc
dedemorton: removed

heartbeat/docs/reference/configuration/heartbeat-options.asciidoc
dedemorton: changed.

heartbeat/docs/reference/configuration/heartbeat-options.asciidoc
dedemorton: clarified the text

heartbeat/docs/reference/configuration/heartbeat-options.asciidoc
dedemorton: OK, added a bit of clarification around this, but didn't spend much time on it. :-)

heartbeat/docs/reference/configuration/heartbeat-options.asciidoc
dedemorton: fixed (not 100% sure, but you can check when I push the updates).

heartbeat/docs/reference/configuration/heartbeat-options.asciidoc
```diff
@@ -0,0 +1,448 @@
...
+
+[[monitor-tcp-proxy-url]]
+===== proxy_url
+
+The SOCKS5 proxy URL.
+
+//Steffen: Can we say more about this option?
```
dedemorton: added more info adapted from the libbeat description.

heartbeat/docs/reference/configuration/heartbeat-options.asciidoc
```diff
@@ -0,0 +1,448 @@
...
+
+[[monitor-tcp-proxy-use-local-resolver]]
+===== proxy_use_local_resolver
```
dedemorton: added more info adapted from the libbeat description

heartbeat/docs/reference/configuration/heartbeat-options.asciidoc
dedemorton: Opened PR https://github.com/elastic/beats/pull/3336

heartbeat/docs/reference/configuration/heartbeat-options.asciidoc
dedemorton: ah, of course!

heartbeat/docs/reference/configuration/heartbeat-options.asciidoc
dedemorton: I won't say more about location until we've tested the feature. For limit, I've added a bit more guidance based on what you've written about (because it seems useful). I'll let you tell me if it's TMI.

heartbeat/docs/reference/configuration/heartbeat-options.asciidoc
dedemorton: thanks for the clarification.

heartbeat/docs/reference/configuration/heartbeat-options.asciidoc
dedemorton: I'm inclined to not say much about this since we haven't tested the feature. We can add more detail after we've tested and know that we want people to use this.

heartbeat/docs/securing-heartbeat.asciidoc
dedemorton: Yes, of course. Not sure what I was thinking when I added that comment. :-P

heartbeat/docs/getting-started.asciidoc
dedemorton: TBH, right now, don't think it's worth a distinction. At some point, we should add a topic called "How Heartbeat Works" and maybe talk about the monitor behavior there (especially if we think users are going to expect behavior that we don't deliver on).

heartbeat/docs/getting-started.asciidoc
```diff
@@ -0,0 +1,261 @@
...
+
+To configure Heartbeat, you edit the configuration file. For rpm and deb,
+you'll find the configuration file at +/etc/heartbeat/heartbeat.yml+.
+For mac and win, look in the archive that you just extracted.
+
+Heartbeat uses monitors to configure ping behavior. You configure each
+monitor individually. Heartbeat currently provides monitors for ICMP ping,
+TCP ping, and HTTP-based ping (see <<heartbeat-overview>> for more about these
+monitors). Here is an example that configures Heartbeat to use an `icmp`
+monitor:
```
dedemorton: I've tried to work this extra info into the overview topic.

urso:
LGTM

PR #3251, opened 2016-12-28
Add query count to metricbeat's mysql plugin output

st1t:
Hello everyone.
Fix to issue the number of Query when using metricbeat 's MySQL module.
Since Com_select and Com_insert etc. are integrated values, we can not grasp the number of queries per second or every minute.
I was wondering whether this difference value should be implemented with metricbeat.
However, since the time interval you want to display is different, I'd like to adjust on the timelion side.

### DIFF
--- a/metricbeat/docs/fields.asciidoc
+++ b/metricbeat/docs/fields.asciidoc
@@ -3464,34 +3464,34 @@ type: long
 
 
 [float]
-== com Fields
+== command Fields
 
 
 
 
 [float]
-=== mysql.status.com.delete
+=== mysql.status.command.delete
 
 type: long
 
 
 
 [float]
-=== mysql.status.com.insert
+=== mysql.status.command.insert
 
 type: long
 
 
 
 [float]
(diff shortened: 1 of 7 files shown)


tsg:
jenkins, test it

metricbeat/docs/fields.asciidoc
```diff
@@ -3463,6 +3463,40 @@ type: long
 
 
 
+[float]
+== com Fields
```
tsg: Not sure about "com". Perhaps we could name it "commands" or "queries"? @ruflin, what do you think?

ruflin: I like `command`, but would keep it singular.

ruflin:
Could you update the CHANGELOG?

st1t:
Thank you for appending advice and code!😀
Now I do not have a pc at hand, so I will fix it later.

ruflin:
jenkins, test it

PR #3252, opened 2016-12-28
Make import_dashboards always exit 0 on errors

### LINKED ISSUE #3027: import_dashboards script return code and logging
- Version: 5.0.1
- Operating System: Debian 8
- Steps to Reproduce: Install Filebeat, Stop Elasticsearch, import dashboards

if the script fails (for example, ES is not running), it exits with a `0` code, instead of anything `> 0`, and it prints messages to `stdout` instead of `stderr`.

```
root@default:/usr/share/filebeat/scripts# ./import_dashboards
fail to set the mapping. Error: Put http://127.0.0.1:9200/.kibana/_mapping/search: dial tcp 127.0.0.1:9200: getsockopt: connection refused
Create temporary directory /tmp/tmp053241378
Downloading https://artifacts.elastic.co/downloads/beats/beats-dashboards/beats-dashboards-5.0.1.zip
Unzip archive  /tmp/tmp053241378
/tmp/tmp053241378/beats-dashboards-5.0.1/filebeat
Import directory  /tmp/tmp053241378/beats-dashboards-5.0.1/filebeat/index-pattern
Import index to /.kibana/index-pattern/filebeat-* from /tmp/tmp053241378/beats-dashboards-5.0.1/filebeat/index-pattern/filebeat.json
ERROR:  couldn't load json. Error: Put http://127.0.0.1:9200/.kibana/index-pattern/filebeat-*: dial tcp 127.0.0.1:9200: getsockopt: connection refused
fail to import index-pattern: fail to load directory /tmp/tmp053241378/beats-dashboards-5.0.1/filebeat/index-pattern: error loading /tmp/tmp053241378/beats-dashboards-5.0.1/filebeat/index-pattern/filebeat.json: couldn't load json. Error: Put http://127.0.0.1:9200/.kibana/index-pattern/filebeat-*: dial tcp 127.0.0.1:9200: getsockopt: connection refused

root@default:/usr/share/filebeat/scripts# echo $?
0
```

autarch:
I also did a little bit of grammar fixing on the error messages while I was in there.

Fixes #3027

### DIFF
--- a/libbeat/dashboards/import_dashboards.go
+++ b/libbeat/dashboards/import_dashboards.go
@@ -147,7 +147,7 @@ func New() (*Importer, error) {
 	/* prepare the Elasticsearch index pattern */
 	fmtstr, err := fmtstr.CompileEvent(cl.opt.Index)
 	if err != nil {
-		return nil, fmt.Errorf("fail to build the Elasticsearch index pattern: %s", err)
+		return nil, fmt.Errorf("failed to build the Elasticsearch index pattern: %s", err)
 	}
 	indexSel := outil.MakeSelector(outil.FmtSelectorExpr(fmtstr, ""))
 
@@ -171,7 +171,7 @@ func New() (*Importer, error) {
 
 	tls, err = outputs.LoadTLSConfig(&tlsConfig)
 	if err != nil {
-		return nil, fmt.Errorf("fail to load the SSL certificate: %s", err)
+		return nil, fmt.Errorf("failed to load the SSL certificate: %s", err)
 	}
 
 	/* connect to Elasticsearch */
@@ -187,7 +187,7 @@ func New() (*Importer, error) {
 		nil,
 	)
 	if err != nil {
-		return nil, fmt.Errorf("fail to connect to Elasticsearch: %s", err)
+		return nil, fmt.Errorf("failed to connect to Elasticsearch: %s", err)
 	}
 	importer.client = client
 
@@ -211,7 +211,7 @@ func (imp Importer) CreateIndex() error {
 			},
 		})
 	if err != nil {
-		fmt.Printf("fail to set the mapping. Error: %s\n", err)
+		fmt.Printf("failed to set the mapping. Error: %s\n", err)
 	}
 	return nil
 }
@@ -222,7 +222,7 @@ func (imp Importer) ImportJSONFile(fileType string, file string) error {
 
 	reader, err := ioutil.ReadFile(file)
 	if err != nil {
-		return fmt.Errorf("fail to read %s. Error: %s", file, err)
+		return fmt.Errorf("failed to read %s. Error: %s", file, err)
 	}
 	var jsonContent map[string]interface{}
 	json.Unmarshal(reader, &jsonContent)
@@ -230,7 +230,7 @@ func (imp Importer) ImportJSONFile(fileType string, file string) error {
 
 	err = imp.client.LoadJSON(path+"/"+fileBase, jsonContent)
 	if err != nil {
-		return fmt.Errorf("fail to load %s under %s/%s: %s", file, path, fileBase, err)
+		return fmt.Errorf("failed to load %s under %s/%s: %s", file, path, fileBase, err)
 	}
 
 	return nil
@@ -363,7 +363,7 @@ func (imp Importer) ImportSearch(file string) error {
 	var searchContent common.MapStr
 	err = json.Unmarshal(reader, &searchContent)
 	if err != nil {
-		return fmt.Errorf("fail to unmarshal search content %s: %v", searchName, err)
+		return fmt.Errorf("failed to unmarshal search content %s: %v", searchName, err)
 	}
 
 	if imp.cl.opt.Index != "" {
@@ -374,7 +374,7 @@ func (imp Importer) ImportSearch(file string) error {
 				var record common.MapStr
 				err = json.Unmarshal([]byte(source), &record)
 				if err != nil {
-					return fmt.Errorf("fail to unmarshal searchSourceJSON from search %s: %v", searchName, err)
+					return fmt.Errorf("failed to unmarshal searchSourceJSON from search %s: %v", searchName, err)
 				}
 
 				if _, ok := record["index"]; ok {
@@ -382,7 +382,7 @@ func (imp Importer) ImportSearch(file string) error {
 				}
 				searchSourceJSON, err := json.Marshal(record)
 				if err != nil {
-					return fmt.Errorf("fail to marshal searchSourceJSON: %v", err)
+					return fmt.Errorf("failed to marshal searchSourceJSON: %v", err)
 				}
 
 				savedObject["searchSourceJSON"] = string(searchSourceJSON)
@@ -456,7 +456,7 @@ func (imp Importer) ImportDir(dirType string, dir string) error {
 
 	files, err := filepath.Glob(path.Join(dir, "*.json"))
 	if err != nil {
-		return fmt.Errorf("fail to read directory %s. Error: %s", dir, err)
+		return fmt.Errorf("failed to read directory %s. Error: %s", dir, err)
 	}
 	if len(files) == 0 {
 		return fmt.Errorf("empty directory %s", dir)
@@ -470,7 +470,7 @@ func (imp Importer) ImportDir(dirType string, dir string) error {
 		}
 	}
 	if len(errors) > 0 {
-		return fmt.Errorf("fail to load directory %s: %s", dir, strings.Join(errors, ", "))
+		return fmt.Errorf("failed to load directory %s: %s", dir, strings.Join(errors, ", "))
 	}
 	return nil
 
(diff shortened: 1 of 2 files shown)


autarch:
Actually, this only partially fixes #3027. The errors are still going to `stdout` instead of `stderr`. Fixing that is a little more work, and I think would benefit from having someone review _all_ the output from this script to determine where it should go. It might also make sense to have a `--quiet` option to suppress info messages going to `stdout` entirely.

libbeat/dashboards/import_dashboards.go
```diff
@@ -677,11 +676,13 @@ func main() {
 	if importer.cl.opt.Dir != "" {
 		if err = importer.ImportKibana(importer.cl.opt.Dir); err != nil {
 			fmt.Println(err)
+			os.Exit(1)
```
andrewkroh: Throughout the main method where error logging is performed I think this should go to stderr (IIRC `fmt.Fprintln(os.stderr, ...)`).

libbeat/dashboards/import_dashboards.go
```diff
@@ -211,7 +211,7 @@ func (imp Importer) CreateIndex() error {
 			},
 		})
 	if err != nil {
-		fmt.Printf("fail to set the mapping. Error: %s\n", err)
+		fmt.Printf("failed to set the mapping. Error: %s\n", err)
```
andrewkroh: I wonder why this is only logging an error and not returning it. @monicasarbu do you know?

We should write this to stderr if it's not going to be returned.

autarch: Three are a few other places where it prints directly to `stdout` for error besides this. The script overall is not terribly consistent about this sort of thing.

libbeat/dashboards/import_dashboards.go
autarch: I agree that errors should be going to stderr. That said, I think this PR is an improvement as-is, since it at least standardizes the exit codes on error.

andrewkroh:
@autarch Thanks for the PR!

autarch:
The test failure looks to me like it might be revealing a previous bug. Is is possible that this load test never actually worked but the failure was being masked by `import_dashboards` returning 0 even when it failed in some cases? I pushed a change so we can at least see the script output when it fails.

autarch:
Alright, my latest commit definitely suggests that problem already existed and just was going unreported. The relevant bit of logs is ...

```
FAIL: Test loading dashboards for all beats
  File "/go/src/github.com/elastic/beats/libbeat/tests/system/test_dashboard.py", line 33, in test_load_dashboard
    self.assertEqual(p.returncode, 0, "stdout:\n{}\n\nstderr:{}\n".format(content, err))
AssertionError: stdout:
failed to import index-pattern: No directory ../../../metricbeat/etc/kibana/index-pattern
stderr:exit status 1
```

I have no idea how to fix this though. Should the test harness be creating the relevant files or is something else going wrong?

andrewkroh:
Looks like that test was bad since it was testing the return code.

Anyways I think the fix is to change the `etc/kibana` to `_meta/kibana` in the `test_load_dashboard` code.

autarch:
Switching to `_meta/kibana` didn't help. The files that `import_dashboard` looks for don't appear to exist in the repo at all. I think the test code needs to run `libbeat/scripts/generate_index_pattern.py` first.

autarch:
But then that python script wants a file called `fields.yml`. However, that file doesn't exist for `filebeat` or `metricbeat` and I'm not sure exactly what generates it.

andrewkroh:
Yeah, I think there was a recent change to not have the generated files checked into the repo. And now there is a dependency problem since a test case in libbeat depends on the `collect` target being run in each of the beat directories first.

@ruflin How do you want to handle this?

libbeat/dashboards/import_dashboards.go
monicasarbu: Ignoring the error here is by intention as `CreateIndex` might return an error if the index already has those settings. It might make sense to prepend "Info:" or "Warning:" to it to avoid confusion.

libbeat/dashboards/import_dashboards.go
```diff
@@ -448,16 +448,15 @@ func (imp Importer) ImportDir(dirType string, dir string) error {
 	// check if the directory exists
 	if _, err := os.Stat(dir); err != nil {
 		// nothing to import
-		fmt.Println("No directory", dir)
-		return nil
+		return fmt.Errorf("No directory %s", dir)
```
monicasarbu: Ignoring the error here was also intentional because some beats (e.g. Filebeat) only have the index-pattern but not the dashboard folder. If we return an error here, we make the script always return non-zero for Filebeat.

autarch: The directory not existing should be an error, though. All the beats have a `_meta` dir. Seems like this should give an error on a missing directory and maybe just an info notice on a missing file.

autarch: Oh, I see, this isn't the `_meta` dir, it's another directory further down the tree. I'll update this PR to account for this and the other non-error error.

autarch: Actually, shouldn't passing a non-existent dir to this script be an error? I think the fix here is to make the test harness a bit smarter. It shouldn't blindly construct a path that doesn't exist and pass it to the `import_dashboards` code.

autarch: Talking to myself is probably a sign of impending mental collapse but ... ok, I see the problem. The `ImportDir` method is called for different types of things which may or may not exist for a given beat.

To make things more complicated, AFAICT _none_ of the beats have a checked-in `index-pattern` dir. However, several of them have `dashboard` dirs and in fact some of them have per-module `dashboard` dirs. These latter cases are going to be missed by the test harness entirely, since it just passes `{beat}/_meta` to `import_dashboards`. I think what it _should_ be doing is finding all of the relevant subdirs and passing those one at a time to test them all.

So here's how I think this should work ...

* The `ImportDir` method should not return an error if the directory doesn't exist, but ...
  * The `ImportKibana` sub should check that the dir it is given exists and return an error if it doesn't.
* The test harness needs to be much smarter, finding all the relevant dirs for each beat and passing them one at a time.
* The test harness needs to call `libbeat/scripts/generate_index_pattern.py` for each beat so those can be imported as well.

monicasarbu: We decided not to commit the index-pattern and generate it every time is needed. To generate the index-pattern for a Beat, we just need to call `make update`. So, in order to fix the test, we just need to find a way to call this before running test. @ruflin knows more here.

As not all the Beats have a dashboard in the latest [beats-dashboards 5.1.1](https://artifacts.elastic.co/downloads/beats/beats-dashboards/beats-dashboards-5.1.1.zip) (see missing `filebeat/dashboard`), but all of them must have an `index-pattern` directory, we can raise an error only if the `index-pattern` directory is not available, and ignore or use `INFO` level in case the`dashboard` directory is not available.

Thank you @autarch for raising all these issues. I agree that there are a few points that we need to reconsider. We are planning to migrate the `import_dashboards` script to the Beat, so the user doesn't have to run an extra step to import the dashboards. This would be a good opportunity to improve the way we import the dashboards. 

Let's wait for @ruflin to find a way to fix the test, and then we see how we can proceed with the PR.

autarch: I did some work on this in the meantime to clean up the error handling and outputting of status messages, so I pushed that. Even if this PR can't be merged as-is, it might be useful for someone else doing the change you talked about to move `import_dashboards`.

autarch:
Note that 9fded29 fixes the rest of #3027, sending all error output to `os.stderr`.

libbeat/dashboards/import_dashboards.go
```diff
@@ -639,49 +645,77 @@ func (imp Importer) ImportArchive() error {
...
+	}
+
+	types := []string{}
+	for _, c := range check {
+		if imp.subdirExists(dir, c) {
+			types = append(types, c)
+		}
+	}
+
+	if len(types) == 0 {
```
monicasarbu: In case the script is run for Filebeat with the `-only-dashboards` argument, it shouldn't return an error as there aren't dashboards to import.

ruflin:
@andrewkroh @autarch This is kind of related to https://github.com/elastic/beats/pull/2980 These tests need some cleanup. It is kind of strange that libbeat needs to know about the other beats when running the tests. I think it shouldn't and we should probably have a generic test in libbeat itself and one that can be extended for each beat with loading the dashboards.

Unfortunately such a cleanup blocks this PR. Alternative is to skip this test for the moment.

@maddin2016 As you spent quite some time with this, perhaps you have some other thoughts how we could handle this?

maddin2016:
Hi @ruflin, actually the problem is that `import_dashboards.go` always return `0`. I think we should skip this tests for the moment until this PR is merged and then we can go on to investigate further on the tests. I also agree with you that every beat should have its own test scenario because not all of them work with the same config (see filebeat for example).

maddin2016:
One question i have. In the script dashboards and index-patterns are loaded from a folder `etc\kibana`. Are these folders created during build?

andrewkroh:
So it sounds like to move forward here we should disable/skip the dashboard test that is in libbeat and merge this PR. And open a new issue to track the need to move the dashboard test from libbeat to the individual beats.

libbeat/dashboards/import_dashboards.go
autarch: It seems odd to me to not return an error. If you tell the script you want it to do X and it can't do X, shouldn't it return an error?

tsg:
@maddin2016 can you skip the test for now so we can merge this PR, please? We're planning to move the `imports_dashboard` code in the beat themselves for the next release, and we'll redesign the test with that occasion. You can skip by adding the `@unittest.skip` decorator.

tsg:
Oops, I meant @autarch in the above ^. Sorry.

autarch:
@tsg - done.

andrewkroh:
Created #3370 to track the refactoring work.

PR #3253, opened 2016-12-28
Add script to run Jenkins builds

andrewkroh:
The jenkins_ci script configures the Go environment based on the version found in libbeat/docs/version.asciidoc. Then it executes the build (check, testsuite, coverage-report, docs).

### DIFF
--- a/dev-tools/common.bash
+++ b/dev-tools/common.bash
@@ -1,0 +1,79 @@
+#
+# File: common.bash
+#
+# Common bash routines.
+#
+
+# Script directory:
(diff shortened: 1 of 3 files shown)


tsg:
Nice, does it mean that on Go version updates we only need to update `libbeat/docs/version.asciidoc` and Jenkins will automatically pick up the right one? One small warning, I think we only have the go version in there in the master branch.

ruflin:
@andrewkroh Nice. Did you already change the Jenkins config? Should we backport this to at least 5.1 so the builds are still working?

@tsg I backported go version to 5.0 and 5.1 branch exactly for this reason ;-)

PR #3248, opened 2016-12-26
Generator for a Filebeat module/fileset

tsg:
Following the model of the Metricbeat module generator, this creates
templates for most of the required files of a fileset, making it
easier to get started.

To use, run as:

    MODULE=apache FILESET=access make create-fileset

The PR includes some minor cleanups to the Metricbeat version.

Part of #3159.

PR #3256, opened 2016-12-29
Filebeat: Apache2 module

tsg:
Contains filesets for access logs and error logs.

Dashboards are still todo. Will need a rebase on top #3248.


filebeat/module/apache2/_meta/fields.yml
```diff
@@ -0,0 +1,9 @@
+- key: apache2
+  title: "apache2"
```
ruflin: We should probably capitalise all titles. We could do this also in the script. @dedemorton did for this some cleanup in Metricbeat.

filebeat/module/apache2/access/_meta/fields.yml
```diff
@@ -0,0 +1,107 @@
...
+        - name: patch
+          type: long
+          description: >
+            The patch version of the user agent.
+        - name: name
+          type: keyword
+          example: Chrome
+          description: >
+            The name of the user agent.
+        - name: os
```
ruflin: os.name, os.major, os.minor?

filebeat/module/apache2/access/_meta/fields.yml
```diff
@@ -0,0 +1,107 @@
...
+          description: >
+            The major version of the operating system.
+        - name: os_minor
+          type: long
+          description: >
+            The minor version of the operating system.
+        - name: os_name
+          type: keyword
+          description: >
+            The name of the operating system.
```
ruflin: How is this different from `os`?

filebeat/module/apache2/access/test/test.log-expected.json
```diff
@@ -0,0 +1,140 @@
...
+        "user_name": "-",
+        "http_version": "1.1",
+        "body_sent": {
+          "bytes": "499"
+        },
+        "url": "/hello",
+        "user_agent": {
+          "major": "50",
+          "minor": "0",
+          "os": "Mac OS X 10.12",
```
ruflin: I see, `os` contains all info and the other 3 are a split up version. Do we need the split up version?

ruflin:
@tsg #3248 is merged. Can you rebase on top?

filebeat/module/apache2/_meta/fields.yml
tsg: This is created by the generator, which is somehow limited in that regard. Do you mean fixing it up by hand?

filebeat/module/apache2/access/_meta/fields.yml
tsg: These are created by the user agent processor, which is not that flexible when it comes to field names: https://www.elastic.co/guide/en/elasticsearch/plugins/master/using-ingest-user-agent.html

We could maybe do renamings, but I'd say the complexity is not worth it.

filebeat/module/apache2/access/test/test.log-expected.json
tsg: It's what the processor does by default. I'd stick with that, since if we don't, someone will ask why it's different from LS.

filebeat/module/apache2/_meta/fields.yml
tsg: I fixed it by hand.

tsg:
@ruflin rebased and squashed.

tsg:
jenkins, package it

tsg:
jenkins, test it

PR #3258, opened 2016-12-29
libbeat: generate Makefile documentation

cyrilleverrier:
Parse Makefile comments to generate the Makefile documentation

Usage:
    make help
    make list_makefile_variables

Features:
    generate documentation for each commented Makefile rules and Makefile variables
    group Makefile rules/variables by category

Examples:

```
	filebeat$ make help

	Usage: make [target] [VARIABLE=value]

	Build:
                filebeat                        Build the beat application.
		crosscompile                 	Cross-compile beat for the OS'es specified in GOX_OS variable. The binaries are placed in the build/bin directory..
		check                        	Checks project and source code if everything is according to standard.
		fmt                          	Runs gofmt -w on the project's source code, modifying any files that do not match its style.
		simplify                     	Runs gofmt -s -w on the project's source code, modifying any files that do not match its style.
		clean                        	Cleans up all files generated by the build steps.
		ci                           	Shortcut for continuous integration. This should always run before merging.
		update                       	Update expects the most recent version of libbeat in the GOPATH.
		docs                         	Builds the documents for the beat.
		docs-preview                 	Preview the documents for the beat in the browser.

	Testing:
		unit-tests                   	Runs the unit tests with coverage.  Race is not enabled for unit tests because tests run much slower.
		unit                         	Runs the unit tests without coverage reports.
		integration-tests            	Run integration tests. Unit tests are run as part of the integration tests.
		integration-tests-environment	Runs the integration inside a virtual environment. This can be run on any docker-machine (local, remote).
		system-tests                 	Runs the system tests.
		system-tests-environment     	Runs the system tests inside a virtual environment. This can be run on any docker-machine (local, remote).
		fast-system-tests            	Runs system tests without coverage reports and in parallel.
		benchmark-tests              	Runs bechmarks (NOT YET IMPLEMENTED).
		load-tests                   	Runs load tests.
		test                         	Runs unit and system tests without coverage reports.
		testsuite                    	Runs all tests and generates the coverage reports.

	Packaging:
		package                      	Create binary packages for the beat.
		set_version                  	VERSION=x.y.z set the version of the beat to x.y.z.
		get_version                  	Get the version of the beat.

	Help:
		help                         	Show this help.
		list_variables               	Show customization variables.
```

```
	filebeat$ make list_makefile_variables

	Packaging:
		BEATNAME                	Name of the application. Default: libbeat
		BEAT_DESCRIPTION        	Description of the application. Default: Sends events to Elasticsearch or Logstash
		BEAT_VENDOR             	Name of the vendor of the application. Default: Elastic
		BEAT_LICENSE            	Software license of the application. Default: ASL 2.0
		BEAT_DOC_URL            	Link to the user documentation of the application. Default: https://www.elastic.co/guide/en/beats/${BEATNAME}/current/index.html

	Community beat:
		ES_BEATS                	Must be set to ./vendor/github.com/elastic/beats. Default: ..
		GOPACKAGES              	Must be set to $(shell glide novendor). Default: ${BEAT_DIR}/...

	Building:
		PACKER_TEMPLATES_DIR    	Directory of templates that are used by "make package". Default: ${ES_BEATS}/dev-tools/packer
		GOX_OS                  	List of all OS to be supported by "make crosscompile". Default: linux darwin windows solaris freebsd netbsd openbsd
		BUILDID                 	The build ID. Default: $(shell git rev-parse HEAD)
		CGO                     	If true, Build with C Go support. Default: false
		TARGETS                 	List of platforms/architecture to be built by "make package". Default: "linux/amd64 linux/386 windows/amd64 windows/386 darwin/amd64"
		TARGETS_OLD             	List of Debian6 architecture to be built by "make package" when CGO is true. Default: ""
		PACKAGES                	List of OS to be supported by "make package". Default: ${BEATNAME}/deb ${BEATNAME}/rpm ${BEATNAME}/darwin ${BEATNAME}/win ${BEATNAME}/bin
		SNAPSHOT                	If yes, builds a snapshot version. Default: yes
		BEATS_BUILDER_IMAGE     	Name of the docker image to use when packaging the application. Default: tudorg/beats-builder
		BEATS_BUILDER_DEB6_IMAGE	Name of the docker image to use when packaging the application for Debian 6. Default: tudorg/beats-builder-deb6

	Testing:
		TEST_ENVIRONMENT        	If true, "make testsuite" runs integration tests and system tests in a dockerized test environment. Default: false
		SYSTEM_TESTS            	If true, "make test" and "make testsuite" run unit tests and system tests. Default: false

	Miscellaneous:
		DOCKER_CACHE            	If set to 0, all docker images are created without cache. Default: 1

```

### DIFF
--- a/libbeat/scripts/Makefile
+++ b/libbeat/scripts/Makefile
@@ -1,14 +1,14 @@
 ### VARIABLE SETUP ###
 ### Application using libbeat may override the following variables in their Makefile
-BEATNAME?=libbeat
-BEAT_DESCRIPTION?=Sends events to Elasticsearch or Logstash
-BEAT_VENDOR?=Elastic
-BEAT_LICENSE?=ASL 2.0
-BEAT_DOC_URL?=https://www.elastic.co/guide/en/beats/${BEATNAME}/current/index.html
+BEATNAME?=libbeat ## @packaging Name of the application
+BEAT_DESCRIPTION?=Sends events to Elasticsearch or Logstash ## @packaging Description of the application
+BEAT_VENDOR?=Elastic ## @packaging Name of the vendor of the application
+BEAT_LICENSE?=ASL 2.0 ## @packaging Software license of the application
+BEAT_DOC_URL?=https://www.elastic.co/guide/en/beats/${BEATNAME}/current/index.html ## @packaging Link to the user documentation of the application
 BEAT_DIR?=github.com/elastic/beats/${BEATNAME}
-ES_BEATS?=..
-GOPACKAGES?=${BEAT_DIR}/...
-PACKER_TEMPLATES_DIR?=${ES_BEATS}/dev-tools/packer
+ES_BEATS?=..## @community_beat Must be set to ./vendor/github.com/elastic/beats
+GOPACKAGES?=${BEAT_DIR}/...## @community_beat Must be set to $(shell glide novendor)
+PACKER_TEMPLATES_DIR?=${ES_BEATS}/dev-tools/packer ## @Building Directory of templates that are used by "make package"
 
 # Makefile for a custom beat that includes this libbeat/scripts/Makefile:
 # if glide is used to manage vendor dependencies,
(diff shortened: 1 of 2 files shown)


cyrilleverrier:
Because I was confused with all the rules and variables to override, it is easier to have an inplace documentation in the Makefile :)

tsg:
jenkins, test it

tsg:
Neat! LGTM.

libbeat/scripts/Makefile
```diff
@@ -436,8 +428,14 @@ fix-permissions:
...
 	${ES_BEATS}/dev-tools/set_version ${VERSION}
 
-get_version:
+get_version: ## @packaging get the version of the beat.
 	@${ES_BEATS}/dev-tools/get_version
+
+help: ## @help Show this help.
+	@python ${ES_BEATS}/libbeat/scripts/generate_makefile_doc.py $(MAKEFILE_LIST)
+
+list_makefile_variables: ## @help Show Makefile customizable variables.
```
ruflin: Perhaps we could call this `help_variables`?

ruflin:
@cyrilleverrier I was thinking about this change again. The only thing that bugs me, is that the previous version of the comments seemed to be better human readable when not using `make help`. Two questions:

* Is what you do with the script some Makefile standard or did you come up with this?
* Would it be possible to have the convention, that every line that starts with `##` is a comment for the following variable / target? It would make the regexp to fetch comments simpler, but would make it more complex to figure out to which target a comment belongs, as it needs to read ahead and skip things like PHONY. 

This is also an improvement we can discuss / make later. This change is already a big improvement 🎉

ruflin:
jenkins, test it

cyrilleverrier:
> @cyrilleverrier I was thinking about this change again. The only thing that bugs me, is that the previous version of the comments seemed to be better human readable when not using make help

I agree with you that the comments in the same line as the target/variable makes it harder to read.

> Is what you do with the script some Makefile standard or did you come up with this?
It is not a standard Makefile comment syntax to document rule.
I took inspiration from https://gist.github.com/prwhite/8168133#gistcomment-1394598 with the following changes:

1. write the script in python instead of perl
2. Parse both Makefile rules and Makefile variable assignments
3. Get the default value of the Makefile variable from the right part of the assignment.

> Would it be possible to have the convention, that every line that starts with ## is a comment for the following variable / target? It would make the regexp to fetch comments simpler, but would make it more complex to figure out to which target a comment belongs, as it needs to read ahead and skip things like PHONY

Yes it is possible, as you have described it.
It is just a matter of defining the right convention on how to write the comments in the Makefile.
Because we could end up with having a comment above the target/variable line AND on the same line.
Which one to use ?


Also, my script handles the substitution of `${BEATNAME}` target but not `${BEATNAME}.test` (this target does not need to be documented anyway as it is a dependency target).
The implementation only supports the substitution when the full name of the target is the name of a variable. But it is easy to fix the python script to handle cases like  `${BEATNAME}.test`

PS: This tool looks also very nice to parse the output of make:
https://www.npmjs.com/package/make2tap

> This is also an improvement we can discuss / make later. This change is already a big improvement :tada:
Hopefully this script can be improved significantly and can have its own snipped on github :)

ruflin:
I merged the PR for now so we have a first working version. As we are not the first ones to hit that issue, I'm curious if there are some conventions / standards out there already for Makefiles so we would not need our own script. 

For the comments: We would have to define a convention probably having generic comments with # and ## comments for processing. But didn't think this through ;-)

ruflin:
Side note: BTW this output is so nice, I'm very tempted to use the generated output as part of our beats developer guide :-)

PR #3270, opened 2017-01-02
Support multiple directories in GOPATH

### LINKED ISSUE #3269: Support multiple directories in GOPATH
I use Go with GOPATH set to a list of directories. This does not play well with the current `Makefile`, `cookiecutter` and the provided commands in the documentation, as they expect GOPATH to be a single location/directory.

breml:
As of `go help gopath` the GOPATH environment variable supports a list of places to look for Go code.
Change Makefiles, scripts in documentation and python helpers to support GOPATH with multiple directories.

Fixes #3269


tsg:
What does the magic syntax mean? I assume take the first value if the env var is colon separated? Maybe it would be better to define a new var (`BEATS_GOPATH=${GOPATH%%:*}`) and use that one throughout the Makefile?

breml:
`${string%%substring}`: Deletes longest match of $substring from back of $string. [reference](http://www.tldp.org/LDP/abs/html/string-manipulation.html)

So `${GOPATH%%:*}` removes everything from back that matches `:*`, which is all but the first path element (delimited by `:`) in the `GOPATH`.

Works (at least) with bash and zsh.

`$(firstword $(subst :, , ${GOPATH}))` in Makefiles does the same. Define a new var like @tsg suggested is fine for me as well.

ruflin:
I think we should definitively have support for multiple GOPATH in the scripts. What I would not do is update the docs with the commands as it makes them more complex. But we should add a note somewhere in the docs, that it the command can be adjusted with the implemented option in case multiple GOPATH are used. This is actually the first time I see multiple GOPATH so I assume it is not very broadly used. Most people new to already struggle with only setting up 1 correctly :-D

The part I'm trying to get my head around is, what is the expected behaviour if someone has multiple GOPATH. With the above we "magically" select the first entry. Is that the expected location?

I like the suggestion from @tsg about introducing `BEATS_GOPATH` but would define it as following: `BEATS_GOPATH?=${GOPATH}`. That allows us to still have the same behaviour as we have now but for expert users as @breml the GOPATH can be overloaded with the correct value, also if it is for example the second entry.

As some of the python script depend on the GOPATH directly, the question is raised if we should now always pass `BEATS_GOPATH` to the scripts, means making this a requirement to be set.

@breml Can you share some more details on the multiple GOPATH use cases? As I didn't see this in the past I'm very interested to learn how it is used as I potentially miss an important argument here to make the above detection the default.

breml:
@ruflin: Thanks for your feedback.

In the documentation files I only changed `$GOPATH` to `${GOPATH%%:*}` if it was in a "complete" script block, which is likely to be copy'n'pasted to a shell. All the other locations I left untouched.

If a tool is using `GOPATH` only to read data (src, pkg), then it simply iterates over all directories listed in `GOPATH`. `go get` is one of the tools, which do also write to locations within `GOPATH` and for `go get` this case is actually defined and documented (see `go get --help`):

> When checking out a new package, get creates the target directory GOPATH/src/<import-path>. If the GOPATH contains multiple entries, get uses the first one. See 'go help gopath'.

In my opinion, the cases affected by this change (`cookiecutter`, `make`, etc.) should handle this the same way `go get` does, because the use case is quite similar. So I guess, to "magically" take the first entry would be the expected behavior.

As I already said, I also like the suggestion from @tsg to introduce `BEATS_GOPATH`.

So last but not least: why use multiple GOPATH. I used it to separate multiple workspaces (e.g. work, private) to keep a better overview while still having some general go libraries in a central directory. I also heard of people separating different projects this way (mainly due to the lack of vendoring prior to version 1.5). One use case I also heard of is to have kind of a ephemeral GOPATH as the first one, where all the "quick" `go get` downloads are stored. If a package renders use full, this is moved (manually) to a kind of "persistent" GOPATH directory, while the ephemeral gets wiped from time to time.

I guess it is mostly a question of personal preference.

ruflin:
@breml Thanks for sharing the insights. If go get picks the first one it makes a lot of sense to follow this convention.

I'm still not too happy about the docs updates. I agree that most of the time it is copy paste, but if I copy / paste something I normally want to understand what it does. And that is not too easy with the line we have here. But I can figure out a solution to this on our side. Perhaps adding a note or something.

ruflin:
jenkins, test it

breml:
PR is merged, close this issue.

PR #3272, opened 2017-01-02
fix libbeat conversion panic

paschdan:
on 64_86 systems there is a interface conversion panic where a float64 is expected.


tsg:
jenkins, test it

PR #3275, opened 2017-01-03
Increase the fields limit

tsg:
Elasticsearch has a default limit of 1000 fields starting with 5.x.
Metricbeat already defines a bit over that in it's current mapping. In most
deployments, these fields won't be actually used all at the same time, so
the effects of increasing the limit shouldn't be that large (can still affect
the cluster state processing).

Longer term we'll need a better solution for this.


ruflin:
We should add a CHANGELOG entry for this to make people that use their own templates or send data through LS aware of this.

tsg:
Changelog updated.

PR #3279, opened 2017-01-03
Disable Filebeat modules collection in 5.2

tsg:
As we're not shipping the Filebeat modules in 5.2, this is disabling
the `collect` target so we avoid leaking the new dashboards and fields.
It also explicitly removes the `read_timestamp` and `error` fields, which
were also added for the filebeat modules.


tsg:
jenkins, package it

tsg:
jenkins, test it

tsg:
I tested the packaging on my computer, looks good.

PR #3281, opened 2017-01-04
Dynamic configuration reloading for modules in Metricbeat

ruflin:
Currently if new module configs have to be added or changed, Metricbeat needs to be restarted. This change allows to define a configuration directory where new files can be added, removed or modified and updates will automatically processed by Metricbeat. New modules will started / stopped accordingly. This is especially useful in container environments where 1 container is used to monitor all services in other containers on the same host. New containers appear and disappear dynamically which requires changes to which modules are needed and which hosts must be monitored.

**Configuration**

The configuration in the main metricbeat.yml config file looks as following:
```
metricbeat.reload.modules:
  enabled: true
  path: configs/*.yml
  period: 10s
```

A path with a glob most be defined on which files should be checked for changes. A period is set on how often the files are checked for changes.

The configuration inside the files which are found by the glob look as following:
```
- module: system
  metricsets: ["cpu"]
  enabled: false
  perdiod: 1s

- module: system
  metricsets: ["network"]
  enabled: true
  period: 10s
```

Each file directly contains a list of modules. Each file can contain one or multiple module definitions.

**How does it work**

When Metricbeat is started a process is started that monitors the defined glob files for changes. If a change is detected, all config files are reloaded and module wrappers are created. A hash is created for the module wrapper to identify if it is a module that is already running or a new module. After loading all modules, all modules which are not in the list anymore will be stopped and removed. All new ones will be started.

The effect of the above is that if 2 identical module configuration exists, only 1 is started. It is not expected to have 2 modules with the exact same configuration.

The dynamic modules have no effect on the modules defined in the global configuration file. These modules will never change and keep running. Also if there is a dynamic module with the exact same config as a global module, these two will not be correlated and both will be run.

**Changes**

* Introduce configuration variables for config reloading
* Change metricbeat behaviour that if `metricbeat.config.enabled: true`, on startup also 0 modules can be defined. Also it is possible that at some point in time no modules are running.
* Change shutdown to also shut down dynamic modules
* Expvar stats for reloading added
* Add common configuration files to metricbeat to have additional configs except from modules
* Add basic system tests


urso:
> The dynamic modules have no effect on the modules defined in the global configuration file. These modules will never change and keep running.

What if a dynamic module resembles a global module?

> Is it an issue to call Connect() for each new module?

No. This is what Connect is for, connecting multiple producers to the publisher pipeline. Just close the client once you're done.

> The current hash only takes module level configs into account. Metricset specific options are no added to the hash. Any idea how we could fix this? The problem is that the config for the metricset is only unloaded when the metricset is created (which is good). It would be nice to be able to calculate a "correct" hash based on common.Config or similar.

Hmmm... One solution might be to unpack the full module config into a `map[string]interface{}` and compute the hash from the map.

metricbeat/beater/metricbeat.go
```diff
@@ -51,8 +59,18 @@ func (bt *Metricbeat) Run(b *beat.Beat) error {
 
 	bt.client = b.Publisher.Connect()
 
+	var configReloader ConfigReloader
+	if bt.config.Config.Enabled {
+		configReloader = NewConfigReloader(bt.config.Config, b.Publisher)
+		go func() {
+			// Start Config monitoring
+			configReloader.Start()
+		}()
```
urso: `== go configReloader.Start()`

metricbeat/beater/reload.go
```diff
@@ -0,0 +1,252 @@
...
+	"github.com/pkg/errors"
+
+	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/logp"
+	"github.com/elastic/beats/libbeat/publisher"
+	"github.com/elastic/beats/metricbeat/mb"
+)
+
+var (
+	debugr        = logp.MakeDebug("reloader")
```
urso: `s/debugr/debugf`

Newline between debugf and expvars?

metricbeat/beater/reload.go
```diff
@@ -0,0 +1,252 @@
...
+	done     chan struct{}
+	wg       sync.WaitGroup
+}
+
+type RunningRegistry struct {
+	List  map[uint64]ModuleRunner
+	mutex sync.Mutex
+}
+
+func NewConfigReloader(config ReloaderConfig, p publisher.Publisher) ConfigReloader {
```
urso: why no pointer? Using NewXXX implies a pointer being returned.

metricbeat/beater/reload.go
```diff
@@ -0,0 +1,252 @@
...
+
+	return ConfigReloader{
+		registry: NewRunningRegistry(),
+		config:   config,
+		client:   client,
+		done:     make(chan struct{}),
+	}
+}
+
+func (r *ConfigReloader) Start() {
```
urso: `s/Start/Run`? This method implements the run-loop itself. Using `Start` implies a go-routine being started, but returns immediately, `Run` implies the method blocks.

metricbeat/beater/reload.go
```diff
@@ -0,0 +1,252 @@
...
+			r.StartModules(startList)
+		}
+	}
+}
+
+func (r *ConfigReloader) Stop() {
+	close(r.done)
+	r.wg.Wait()
+	// Stop all running modules
+	r.StopModules(r.registry.CopyList())
```
urso: alternatively `StopModules` can be done via `defer` in run loop.

metricbeat/beater/reload.go
```diff
@@ -0,0 +1,252 @@
...
+			newFiles, err := FetchFiles(r.config.Path, lastScan)
+			if err != nil {
+				if err != ErrNoNewFiles {
+					logp.Err("Error fetching files: %v", err)
+				}
+				// Continue on error or no new files
+				continue
+			}
+
+			lastScan = lastScanTemp
```
urso: why set lastScan only if any files have been changed?

ruflin:
@urso Thanks for feedback:

* dynamic modules resemblance. Both will be run. No correlation is done. I will update the PR description accordingly
* Connect: Great
* Hash: Will try if the above will work for comparison.

metricbeat/beater/reload.go
```diff
@@ -0,0 +1,252 @@
+package beater
```
tsg: I wonder if the code in this file could be generalized into a libbeat package. It seems to me that it contains relatively little Metricbeat specific logic, except for the `StartModules` and `StopModules` functions which can be easily injected.

I'm thinking that we'll want to make it easy to extend this type of reloading functionality to other beats.

ruflin: Fully agree, that we probably should extend this to other beats, especially filebeat. Means this functionality will move to libbeat. I prefer to keep it currently in metricbeat to keep the changes as minimal as possible and do the abstraction in a second step when we add it to other beats.

metricbeat/beater/reload.go
ruflin: agree. moved.

metricbeat/beater/reload.go
ruflin: Even thought that was lost during refactoring and initially the purpose was to only update lastScan if no error, I think it is still correct like this. If there are no file updates, it does not matter if the comparison of the files is to the most recent lastScan or one of the previous where no files were found. The resulting outcome should be the same and it keeps the code simpler. I will add a comment to the code.

metricbeat/beater/reload.go
ruflin: 👍

metricbeat/beater/reload.go
ruflin: changed

metricbeat/beater/reload.go
ruflin: it is debugr on purpose, as debugf is already taken from metricbeat.go. But I wanted the debugging in a different namespace. Indirectly implies that later this should be move to its own namespace but this requires some more refactoring which should be done in a later step.

metricbeat/beater/metricbeat.go
ruflin: changed

metricbeat/metricbeat.yml
```diff
@@ -10,41 +10,13 @@
 #==========================  Modules configuration ============================
 metricbeat.modules:
```
monicasarbu: I think `metricbeat.modules` needs to be removed here.

ruflin: @monicasarbu This one is WIP. I will restore the old metricbeat.yml as soon as I added automated tests. This file should not change.

metricbeat/_meta/beat.full.yml
```diff
@@ -7,6 +7,21 @@
...
 # https://www.elastic.co/guide/en/beats/metricbeat/index.html
 
+#============================  Config Reloading ===============================
+
+# Config reloading allows to dynamically load modules. Each file which is
+# monitored must contain one or multiple modules as a list.
+metricbeat.reload:
+
+  # Glob pattern for configuration reloading
+  path: config/*.yml
```
andrewkroh: I'd go with `conf.d/*.yml`. I think that name might provide a hint about the purpose of the directory.

metricbeat/beater/metricbeat.go
```diff
@@ -68,6 +86,12 @@ func (bt *Metricbeat) Run(b *beat.Beat) error {
 		PublishChannels(bt.client, cs...)
 	}()
 
+	// Blocks metricbeat from shutting down if no module is set. Required for config reloading
+	if bt.config.ReloadModules.Enabled {
```
andrewkroh: I think the initialization and starting of the NewConfigReloader should happen here to so that it's all in one place. You could make everything block on the same WaitGroup too...

```
if bt.config.ReloadModules.Enabled {
  configReloader := NewConfigReloader..

  // Start
  wg.Add(1)
  go func() {
    defer wg.Done()
    configReloader.Run()
  }()

  // Stop
  go func() {
    <-bt.done
    configReloader.Stop()
  }()
}

File: metricbeat/beater/reload.go [hunk 11]
```diff
@@ -0,0 +1,221 @@
...
+
+type ConfigReloader struct {
+	registry RunningRegistry
+	config   ModulesReloadConfig
+	client   func() publisher.Client
+	done     chan struct{}
+	wg       sync.WaitGroup
+}
+
+type RunningRegistry struct {
```
andrewkroh: It looks like this is used internally only by the ConfigReloader. Can this be unexported?

metricbeat/beater/reload.go
```diff
@@ -0,0 +1,221 @@
...
+	List  map[uint64]ModuleRunner
+	mutex sync.Mutex
```
andrewkroh: If the mutex is required to access `List` then they should have the same visibility. But I like embedding the mutex in the struct then you can call `runningReg.Lock()`.

metricbeat/beater/reload.go
```diff
@@ -0,0 +1,221 @@
...
+	var c []*common.Config
+	err = rawConfig.Unpack(&c)
+	if err != nil {
+		return nil, fmt.Errorf("Error reading configuration from file %s: %s", file, err)
+	}
+
+	return c, nil
+}
+
+func NewRunningRegistry() RunningRegistry {
```
andrewkroh: Given that this struct contains a `sync.Mutex`, returning a `*RunningRegistry` would better indicate to the caller that this object needs to be passed by reference than by value (as you should not copy a `sync.Mutex`).

metricbeat/beater/reload.go
```diff
@@ -0,0 +1,221 @@
...
+			if err != nil {
+				if err != mb.ErrAllModulesDisabled && err != mb.ErrEmptyConfig {
+					// Continuing as only some modules could have an error
+					logp.Err("Error creating modules: %s", err)
+				}
+			}
+
+			debugr("Number of module wrappers created: %v", len(s))
+
+			startList := []*ModuleWrapper{}
```
andrewkroh: Use `var startList []*ModuleWrapper` to avoid allocating an empty map if it doesn't get used. [ref](https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices)

metricbeat/_meta/beat.full.yml
andrewkroh: I wonder if we could create this directory automatically in `/etc/metricbeat/conf.d` when we install the rpm/deb. And do the same for Filebeat.

ruflin: Changed to `conf.d`. I like the idea of creating this automatically. Reminds me to check to what the path is relative. Should the default be `${path.conf}/conf.d/*.yml`?

metricbeat/beater/config.go
```diff
@@ -1,9 +1,29 @@
...
+
+	"github.com/elastic/beats/libbeat/common"
+)
 
 // Config is the root of the Metricbeat configuration hierarchy.
 type Config struct {
 	// Modules is a list of module specific configuration data.
-	Modules []*common.Config `config:"modules" validate:"required"`
+	Modules       []*common.Config    `config:"modules"`
+	ReloadModules ModulesReloadConfig `config:"reload.modules"`
```
ruflin: Changed it to the same logic as the output. +1 on consistency here.

metricbeat/beater/glob_change_detector.go
```diff
@@ -0,0 +1,79 @@
...
+
+		// Directories are symlinks are skipped
+		if !info.Mode().IsRegular() {
+			continue
+		}
+
+		// Check if one of the files was changed recently
+		// File modification time can be in seconds. -1 is to cover for files which
+		// were created during this second.
+		if info.ModTime().Sub(gcd.lastScan) >= -1*time.Second {
```
ruflin: Changed to your suggestion. OS X seems to store mtime in seconds. At least I only see seconds when I use `stat`.

metricbeat/beater/glob_change_detector.go
```diff
@@ -0,0 +1,79 @@
...
+			updatedFiles = true
+		}
+		files = append(files, f)
+	}
+
+	hash, err := hashstructure.Hash(files, nil)
+	if err != nil {
+		logp.Err("Error hashing files list: %s", err)
```
ruflin: I initially ignored the error as I didn't see an option how we ever get here an error. But to be sure we catch it in case it happens, I added the log line. 

If there is an error, the hash is 0. So in case hashing fails one time because of reason x, scan would still return changed which is the save way to go. I prefer to return changed once too many. If it continously fails, it will in the future depend on `updatedFiles` only. The not so nice part about this is that it does not detect removed files with updatedFiles.

Thinking about this again, probably the most consistent behaviour would be directly return `files, true, nil` in case of an error, to make sure in an error case `changed` is true. Will change it to this.

metricbeat/beater/reload.go
ruflin: yes, changed

metricbeat/beater/reload.go
ruflin: Changed

metricbeat/beater/reload.go
ruflin: changed

metricbeat/beater/reload.go
ruflin: changed

metricbeat/_meta/beat.full.yml
ruflin: BTW: Creation of the directory I would suggest not to do in this PR and look into later.

metricbeat/beater/metricbeat.go
ruflin: Changed accordingly. It is much nicer to have only 1 place that blocks.

ruflin:
@andrewkroh I pushed 2 new commits. One for the general refactoring and one for the config changes to make it relative. Could you have a look?

metricbeat/docs/reference/configuration/reload-configuration.asciidoc
```diff
@@ -0,0 +1,39 @@
...
+
+[source,yaml]
+------------------------------------------------------------------------------
+metricbeat.reload.modules:
+  enabled: true
+  path: configs/*.yml
+  period: 10s
+------------------------------------------------------------------------------
+
+A path with a glob most be defined on which files should be checked for changes. A period is set on how often
```
monicasarbu: /s/most/must

metricbeat/metricbeat.full.yml
```diff
@@ -7,6 +7,21 @@
 # You can find the full configuration reference here:
 # https://www.elastic.co/guide/en/beats/metricbeat/index.html
 
+#============================  Config Reloading ===============================
+
+# Config reloading allows to dynamically load modules. Each file which is
+# monitored must contain one or multiple modules as a list.
+metricbeat.reload:
```
monicasarbu: I think it should be `metricbeat.reload.modules` here.

metricbeat/beater/config.go
ruflin: Hm, that didn't work as I expected. Because I have a default config, which I already load on the metricbeat level, the config is never nil. I plan to extract the reload part anyways. Ok if I do it in a second step?

metricbeat/metricbeat.full.yml
ruflin: Definitively, thanks for spotting. Fixed.

metricbeat/docs/reference/configuration/reload-configuration.asciidoc
ruflin: changed

metricbeat/_meta/beat.full.yml
```diff
@@ -7,6 +7,21 @@
...
 # https://www.elastic.co/guide/en/beats/metricbeat/index.html
 
+#============================  Config Reloading ===============================
+
+# Config reloading allows to dynamically load modules. Each file which is
+# monitored must contain one or multiple modules as a list.
+metricbeat.reload:
+
+  # Glob pattern for configuration reloading
+  path: conf.d/*.yml
```
urso: how about setting this to: 

```
path: '${path.config}/conf.d/*.yml'
```

metricbeat/beater/glob_change_detector.go
```diff
@@ -0,0 +1,79 @@
...
+import (
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/elastic/beats/libbeat/logp"
+	"github.com/mitchellh/hashstructure"
+)
+
+type GlobChangeDetector struct {
```
urso: `GlobWatcher`, `Wacher`, or `FileWatcher` ?

metricbeat/beater/glob_change_detector.go
```diff
@@ -0,0 +1,79 @@
...
+		info, err := os.Lstat(f)
+		if err != nil {
+			logp.Err("Error getting stats for file: %s", f)
+			continue
+		}
+
+		// Directories and symlinks are skipped
+		if !info.Mode().IsRegular() {
+			continue
+		}
```
urso: hm... having symlink support would be nice to have. This way op can enable modules selectively via `ln -s`... (kinda common practice in init systems, even systemd).

metricbeat/beater/config.go
andrewkroh: Yeah, that's fine. With using the pointer you'll need handle defaults when reading the individual values (haven't found a better way).

metricbeat/beater/glob_change_detector.go
ruflin: I like GlobWatcher. Changing ...

metricbeat/beater/config.go
andrewkroh: You'll need to make `ReloadModules` a pointer too. But with the default present it won't be nil.

metricbeat/_meta/beat.full.yml
ruflin: changed

metricbeat/beater/glob_change_detector.go
ruflin: Ok, enabled symlinks

ruflin:
@andrewkroh @monicasarbu @urso I applied the changes. If ok for all of you I will squash and rebase to update the commit message.

metricbeat/beater/reload.go
```diff
@@ -0,0 +1,223 @@
+package beater
+
+import (
+	"expvar"
+	"fmt"
+	"sync"
+	"time"
+
```
andrewkroh: Extra line here

ruflin: About Reload: Yeah, tried different versions, but didn't work. Extracting it and having a separate unpack should work (in a later PR). Not sure why I can't comment above -> Github fail.

Will remove the extra line when rebasing.

andrewkroh:
jenkins, test it

Issue #3261, opened 2016-12-30
scripts/generate_index_pattern.py doesn't seem to be adding the default `_feilds`


blacktop:
For confirmed bugs, please report:
- Version: master
- Operating System: OSX
- Steps to Reproduce: 
`make update`

It is not making these fields that if you don't push the beat-generated index-pattern and just let kibana figure it out it DOES create them:

```json
 {\"name\":\"_id\",\"type\":\"string\",\"count\":0,\"scripted\":false,\"indexed\":false,\"analyzed\":false,\"doc_values\":false,\"searchable\":false,\"aggregatable\":false},

 {\"name\":\"_type\",\"type\":\"string\",\"count\":0,\"scripted\":false,\"indexed\":false,\"analyzed\":false,\"doc_values\":false,\"searchable\":true,\"aggregatable\":true},

 {\"name\":\"_index\",\"type\":\"string\",\"count\":0,\"scripted\":false,\"indexed\":false,\"analyzed\":false,\"doc_values\":false,\"searchable\":false,\"aggregatable\":false}

 {\"name\":\"_score\",\"type\":\"number\",\"count\":0,\"scripted\":false,\"indexed\":false,\"analyzed\":false,\"doc_values\":false,\"searchable\":false,\"aggregatable\":false}
```

PR #3282, opened 2017-01-04
Add special fields in the index pattern

monicasarbu:
Trying to fix https://github.com/elastic/beats/issues/3261 by adding `_id`, `_type`, `_index` and `_score` fields to the generated index pattern.


blacktop:
@monicasarbu I noticed that when kibana generates those fields that it makes `_type` searchable and aggregatable.  I am not sure that is important or not.

monicasarbu:
@blacktop Yes, you are right. Thank you. I have created another PR to fix the issue: https://github.com/elastic/beats/pull/3284

PR #3284, opened 2017-01-04
Enable aggregatable for special fields in index pattern

monicasarbu:
Use keyword for special fields(`_index`, `_id`, `_type` and `_score`). This depends on https://github.com/elastic/beats/pull/3282.

### DIFF
--- a/libbeat/scripts/generate_index_pattern.py
+++ b/libbeat/scripts/generate_index_pattern.py
@@ -31,7 +31,9 @@ def fields_to_json(section, path, output):
             field_to_json(field, newpath, output)
 
 
-def field_to_json(desc, path, output):
+def field_to_json(desc, path, output,
+                  indexed=True, analyzed=False, doc_values=True,
+                  searchable=True, aggregatable=True):
 
     global unique_fields
 
@@ -45,11 +47,11 @@ def field_to_json(desc, path, output):
         "name": path,
         "count": 0,
         "scripted": False,
-        "indexed": True,
-        "analyzed": False,
-        "doc_values": True,
-        "searchable": True,
-        "aggregatable": True,
+        "indexed": indexed,
+        "analyzed": analyzed,
+        "doc_values": doc_values,
+        "searchable": searchable,
+        "aggregatable": aggregatable,
     }
     # find the kibana types based on the field type
     if "type" in desc:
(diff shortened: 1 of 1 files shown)


blacktop:
So the way I fixed it was to add this: https://github.com/blacktop/brobeat/blob/master/scripts/generate_index_pattern.py#L99

Which is over-kill as your solution is much simpler, however, the only thing that is different is that the `_type` field has the following set.

```python
    meta_fields = [
        {
            "name": "_id",
            "type": "string",
            "count": 0,
            "scripted": False,
            "indexed": False,
            "analyzed": False,
            "doc_values": False,
            "searchable": False,
            "aggregatable": False,
        },
        {
            "name": "_type",
            "type": "string",
            "count": 0,
            "scripted": False,
            "indexed": False,
            "analyzed": False,
            "doc_values": False,
            "searchable": True, <===========================
            "aggregatable": True, <===========================
        },
        {
            "name": "_index",
            "type": "string",
            "count": 0,
            "scripted": False,
            "indexed": False,
            "analyzed": False,
            "doc_values": False,
            "searchable": False,
            "aggregatable": False,
        },
        {
            "name": "_score",
            "type": "number", <=======================
            "count": 0,
            "scripted": False,
            "indexed": False,
            "analyzed": False,
            "doc_values": False,
            "searchable": False,
            "aggregatable": False,
        },
    ]
```

> NOTE: I set all the default fields in my example (which I know the python will set automatically in your solution)

**Also**, I think the defaults will make the meta-fields, `indexed`, `doc_values`, `searchable` and `aggregatable` all `True`, but **Kibana** makes them all `False` by default as can be seen above.
 
This can been seen in your code here - https://github.com/monicasarbu/beats/blob/53e873325a51bfaa85b78b48b2e3bf3f32538bb6/libbeat/scripts/generate_index_pattern.py#L44

```python
    field = {
        "name": path,
        "count": 0,
        "scripted": False,
        "indexed": True, <=======================
        "analyzed": False,
        "doc_values": True, <=======================
        "searchable": True, <=======================
        "aggregatable": True, <=======================
    }
```

**Also** kibana makes `_score` a `number` and not a `integer`, but I am not sure that matters.

monicasarbu:
@blacktop I have updated the PR to set the fields (`_index`, `_id`, `_score` and `_type`) with the same options as Kibana. Thanks for the detailed feedback. I am setting `_score` of type `integer` as the script changes an `integer` to number.

PR #3289, opened 2017-01-05
fix typo

breml:
I guess, this should be _registrar_ (and not _registar_)


tsg:
I think so too, although registar sounds kind of cool :)

PR #3304, opened 2017-01-07
Support to debug print configuration files

urso:
This PR adds support for printing loaded configurations as debug message. The config is JSON encoded before printing.

This is helpful in debugging configuration issues, as the JSON can be directly copied and tested locally + indentation errors become more visible.

Changes:
- apply CLI logging flags before loading config file + configure stderr
- add (common.*Config).DebugString to create formatted string from config
  object
- print config after loading
- print used sub-config when loading output plugins and configuring
  processors

### DIFF
--- a/libbeat/cfgfile/cfgfile.go
+++ b/libbeat/cfgfile/cfgfile.go
@@ -74,6 +74,11 @@ func HandleFlags() error {
 	}
 
 	defaults.SetString("path.home", -1, home)
+
+	if len(overwrites.GetFields()) > 0 {
+		overwrites.PrintDebugf("CLI setting overwrites (-E flag):")
+	}
+
 	return nil
 }
 
@@ -120,11 +125,17 @@ func Load(path string) (*common.Config, error) {
 		return nil, err
 	}
 
-	return common.MergeConfigs(
+	config, err = common.MergeConfigs(
 		defaults,
 		config,
 		overwrites,
 	)
+	if err != nil {
+		return nil, err
+	}
+
+	config.PrintDebugf("Complete configuration loaded:")
+	return config, nil
 }
 
 // GetPathConfig returns ${path.config}. If ${path.config} is not set, ${path.home} is returned.
--- a/libbeat/common/config.go
+++ b/libbeat/common/config.go
@@ -1,8 +1,12 @@
 package common
 
 import (
+	"encoding/json"
 	"flag"
+	"fmt"
+	"strings"
 
+	"github.com/elastic/beats/libbeat/logp"
 	"github.com/elastic/go-ucfg"
 	"github.com/elastic/go-ucfg/cfgutil"
 	cfgflag "github.com/elastic/go-ucfg/flag"
@@ -10,6 +14,27 @@ import (
 )
 
 type Config ucfg.Config
+
+const (
+	selectorConfig             = "config"
+	selectorConfigWithPassword = "config-with-passwords"
+)
+
+var debugBlacklist = MakeStringSet(
+	"password",
+	"passphrase",
+	"key_passphrase",
+	"pass",
+	"proxy_url",
+	"url",
+	"urls",
+	"host",
+	"hosts",
+)
+
+// make hasSelector and configDebugf available for unit testing
+var hasSelector = logp.HasSelector
+var configDebugf = logp.Debug
 
 type flagOverwrite struct {
 	config *ucfg.Config
@@ -111,14 +136,20 @@ func NewFlagOverwrite(
 
 func LoadFile(path string) (*Config, error) {
 	c, err := yaml.NewConfigWithFile(path, configOpts...)
-	return fromConfig(c), err
+	if err != nil {
+		return nil, err
+	}
+
+	cfg := fromConfig(c)
+	cfg.PrintDebugf("load config file '%v' =>", path)
+	return cfg, err
 }
 
 func LoadFiles(paths ...string) (*Config, error) {
 	merger := cfgutil.NewCollector(nil, configOpts...)
 	for _, path := range paths {
-		err := merger.Add(yaml.NewConfigWithFile(path, configOpts...))
-		if err != nil {
+		cfg, err := LoadFile(path)
+		if err := merger.Add(cfg.access(), err); err != nil {
 			return nil, err
 		}
 	}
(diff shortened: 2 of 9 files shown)


andrewkroh:
Is there any thing we can do to prevent the `password` fields from being logged when dumping the config?

urso:
No. This PR only converts the config as is to JSON. No normalization or whatsoever. Removing passwords would require to iterate the `map[string]interface{}` and remove every `password` field for example. But then, how to handle e.g. URLs or other parameters holding credentials.

The DebugString is for debugging purposes only. No logging at info level.

libbeat/cfgfile/cfgfile.go
```diff
@@ -120,11 +128,17 @@ func Load(path string) (*common.Config, error) {
...
+	config, err = common.MergeConfigs(
 		defaults,
 		config,
 		overwrites,
 	)
+	if err != nil {
+		return nil, err
+	}
+
+	debugf("return config =>\n%v", config.DebugString())
```
ruflin: We should change this to something like: "Complete merged configuration:" as `return config` is not really explaining on what it is.

libbeat/cfgfile/cfgfile.go
```diff
@@ -74,6 +77,11 @@ func HandleFlags() error {
 	}
 
 	defaults.SetString("path.home", -1, home)
+
+	if len(overwrites.GetFields()) > 0 {
+		debugf("setting overwrites:\n%v", overwrites.DebugString())
```
ruflin: Config options set by -E:

libbeat/common/config.go
```diff
@@ -190,6 +202,39 @@ func (c *Config) SetChild(name string, idx int, value *Config) error {
...
 
+func (c *Config) IsDict() bool {
+	return c.access().IsDict()
+}
+
+func (c *Config) IsArray() bool {
+	return c.access().IsArray()
+}
+
+func (c *Config) DebugString() string {
```
ruflin: This could be called just "String()" as it is the String output of loaded config. Should this be part of ucfg directly?

libbeat/common/config.go
```diff
@@ -190,6 +202,39 @@ func (c *Config) SetChild(name string, idx int, value *Config) error {
...
+	var bufs []string
+	if c.IsDict() {
+		var content map[string]interface{}
+		if err := c.Unpack(&content); err != nil {
+			return fmt.Sprintf("<config error> %v", err)
+		}
+		j, _ := json.MarshalIndent(content, "", "  ")
```
ruflin: I assume here we could also choose yaml if needed?

ruflin:
That is a really helpful change which will make it even easier for us to not have to ask for config file if someone provides debug output. I agree with @andrewkroh that in some cases this will lead to problem as users will post username / passwords inside.

As the method can be used at any level, I'm thinking of introducing this at some places (without passwords) to also log the config options on Info level. We already do this but it is a manual print out. 

The above seems to break some metricbeat config tests. I assume this is related to the full config tests. Perhaps we should skip these again and figure out something new with this change. The tricky part here again is that two configs can be identical (because of the defaults) but the printout is not the same.

libbeat/cfgfile/cfgfile.go
urso: how about `CLI setting overwrite (-E flag):\n%v`?

libbeat/cfgfile/cfgfile.go
urso: will change to `Complete configuration loaded:\n%v`

libbeat/common/config.go
urso: `String` is already used as Accessor. Plus I like `DebugString()`, so it's more clear this should be used for debugging only (I hope) :) .

Not sure I like this in `go-ucfg`. Plus `go-ucfg` is meant to be 'wrapped/extended' on purpose (e.g. for dealing with vendoring/version conflicts).

libbeat/common/config.go
urso: tl;dr debug output is already a subset of YAML :P 

I opted for json in debug output, so it's more easy to 'see' the config as is loaded. e.g. in json it gets more apparent if users got the alignment wrong. Plus one can simply copy the json (even if debug log is not properly formatted on disucss) and use the debug config in place of the yaml (as JSON is a subset of YAML).

urso:
This PR also changes how/when the logger output is (re-)configured, so to generate some actual output (logger used to be setup only after config file is loaded => chicken egg problem). I introduced HandleFlags to 'bootstrap' the logger to print to stderr before config file loading to work around this issue. => Need to investigate why tests fail.

> As the method can be used at any level, I'm thinking of introducing this at some places (without passwords) to also log the config options on Info level. We already do this but it is a manual print out.

:(
I think I will rename/change `(*common.Config).DebugString` to `(*common.Config).PrintDebug` :P 
This feature is really only supposed to be used for debugging purposes. Especially considering `DebuString` not interpreting the config in any way + potential credentials in whatever config fields. (e.g. user has type in `pasword`).

libbeat/common/config.go
ruflin: We can still decide it later. It just looks like something other go-ucfg user could also have a need for.

urso:
Update PR to explicitely require `config` or `config-with-passwords` debug selectors. If `config` debug selector is used, some blacklisted fields will be replaced with `'xxxxx'` (see `debugBlacklist` in config.go).

libbeat/logp/log.go
```diff
@@ -28,21 +28,53 @@ type Logger struct {
...
+	}
+	return set, all
+}
 
 func debugMessage(calldepth int, selector, format string, v ...interface{}) {
 	if _log.level >= LOG_DEBUG {
 		if !_log.debugAllSelectors {
-			selected := _log.selectors[selector]
-			if !selected {
+			if _, selected := _log.selectors[selector]; !selected {
```
ruflin: Here you could use `!HasSelector(selector)`

ruflin:
I did some manual testing and it seems to work as expected. LGTM.

libbeat/logp/log.go
urso: right :)

urso:
Squashed commits and updated commit message.

PR #3308, opened 2017-01-09
Use ModuleRunner for all modules to unify handling

ruflin:
* Use moduleRunner also for non reloadable modules
* Remove go routing in reload module as not needed. It's already in the module runner

### DIFF
--- a/metricbeat/beater/metricbeat.go
+++ b/metricbeat/beater/metricbeat.go
@@ -57,35 +57,31 @@ func New(b *beat.Beat, rawConfig *common.Config) (beat.Beater, error) {
 // within the same Module and MetricSet from collection.
 func (bt *Metricbeat) Run(b *beat.Beat) error {
 
-	var runners []ModuleRunner
+	var wg sync.WaitGroup
+
 	for _, m := range bt.modules {
 		r := NewModuleRunner(b.Publisher.Connect, m)
 		r.Start()
-		runners = append(runners, r)
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			<-bt.done
+			r.Stop()
+		}()
 	}
 
 	if bt.config.ReloadModules.IsEnabled() {
 		logp.Warn("EXPERIMENTAL feature dynamic configuration reloading is enabled.")
-
 		configReloader := NewConfigReloader(bt.config.ReloadModules, b.Publisher)
-		// Start
-		go func() {
-			configReloader.Run()
-		}()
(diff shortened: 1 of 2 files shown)


ruflin:
Agree, that is also thing we should have for the dynamic modules on shutdown. Will look into it how I can parallelise this.

ruflin:
@andrewkroh Updated

metricbeat/beater/metricbeat.go
```diff
@@ -57,45 +57,35 @@ func New(b *beat.Beat, rawConfig *common.Config) (beat.Beater, error) {
...
-		configReloader := NewConfigReloader(bt.config.ReloadModules, b.Publisher)
 
+		configReloader := NewConfigReloader(bt.config.ReloadModules, b.Publisher)
 		// Start
-		wg.Add(1)
 		go func() {
-			defer wg.Done()
 			configReloader.Run()
 		}()
+		defer configReloader.Stop()
```
andrewkroh: `ConfigReloader` should be stopped in parallel with the other modules.

metricbeat/beater/metricbeat.go
```diff
@@ -57,45 +57,35 @@ func New(b *beat.Beat, rawConfig *common.Config) (beat.Beater, error) {
...
-		go func() {
-			<-bt.done
-			configReloader.Stop()
+	<-bt.done
+
+	wg := sync.WaitGroup{}
+	// Stop runners in parallel
+	for _, r := range runners {
+		wg.Add(1)
+		func() {
```
andrewkroh: You want to run this in a separate go routine?

metricbeat/beater/metricbeat.go
andrewkroh: Here's what I was thinking: https://gist.github.com/andrewkroh/d864b205a508a2246f67137dcdddc8fb

ruflin: SGTM. Long term probably all modules should be in the registry and we have only 1 place to stop, but currently this would stop the fixed modules after the first reload.

metricbeat/beater/metricbeat.go
ruflin: With your suggestion above, this becomes obsolete, right?

andrewkroh:
Still "in progress"?

ruflin:
@andrewkroh Ready to be merged. Forgot to switch the label again.

PR #3309, opened 2017-01-09
Change prospectors to only handle copy of states

ruflin:
* Only pass copy of states to prospectors. Each prospector gets a fresh copy of states.
* Pass registrar to crawler, as needed to fetch follow up states
* Simplify code for `-once` handling
* Move once variable into prospector to not have it in Run method
* Add unique identifier to each prospector
* Move hash structure lib to top level directory
* Simplify states mutex


filebeat/crawler/crawler.go
```diff
@@ -6,62 +6,70 @@ import (
...
-func (c *Crawler) Start(states file.States, once bool) error {
+func (c *Crawler) Start(r *registrar.Registrar) error {
 
 	logp.Info("Loading Prospectors: %v", len(c.prospectorConfigs))
 
 	// Prospect the globs/paths given on the command line and launch harvesters
 	for _, prospectorConfig := range c.prospectorConfigs {
-
-		prospector, err := prospector.NewProspector(prospectorConfig, states, c.out)
+		err := c.startProspector(prospectorConfig, r.GetStates())
```
urso: is there a chance r.GetStates() being different for n-th prospector due to state update triggered by first prospector publishing events?

filebeat/crawler/crawler.go
```diff
@@ -6,62 +6,70 @@ import (
...
 
-	for i, p := range c.prospectors {
+func (c *Crawler) startProspector(config *common.Config, states []file.State) error {
+	prospector, err := prospector.NewProspector(config, states, c.out)
+	if err != nil {
+		return fmt.Errorf("Error in initing prospector: %s", err)
+	}
+	prospector.Once = c.once
+
+	if prospector.IsEnabled() {
```
urso: how about:

```
func (c *Crawler) startProspector(config *common.Config, states []file.State) error {
  if ! config.Enabled() {
    return nil
  }

  prospector, err := prospector.New(config, states, c.out)
  if err != nil {
   ...
  }
  prospector.Once = c.once
  ...
}
```

filebeat/crawler/crawler.go
```diff
@@ -6,62 +6,70 @@ import (
...
+		c.prospectors[prospector.Hash] = prospector
```
urso: maybe check no prospector with same hash exists?

Naming: `prospector.ID` ?

filebeat/prospector/prospector.go
```diff
@@ -52,16 +56,25 @@ func NewProspector(cfg *common.Config, states file.States, outlet Outlet) (*Pros
...
+		Once:          false,
 	}
 
-	if err := cfg.Unpack(&prospector.config); err != nil {
+	var err error
+	if err = cfg.Unpack(&prospector.config); err != nil {
 		return nil, err
 	}
-	if err := prospector.config.Validate(); err != nil {
+	if err = prospector.config.Validate(); err != nil {
```
urso: `Validate` is executed by go-ucfg upon Unpack (if not, it's a bug in go-ucfg).

filebeat/prospector/prospector.go
```diff
@@ -52,16 +56,25 @@ func NewProspector(cfg *common.Config, states file.States, outlet Outlet) (*Pros
...
 		return nil, err
 	}
 
-	err := prospector.Init(states)
+	var h map[string]interface{}
+	cfg.Unpack(&h)
+	prospector.Hash, err = hashstructure.Hash(h, nil)
+	if err != nil {
+		return nil, err
+	}
```
urso: makes me wonder if we want to provide a `(*common.Config).Hash() ...` method.

Also consider `(*common.Config).ToMap() (map[string]interface{}, error)` and a `(*common.Config).ToArray() ([]interface{}, error)`.

As `*common.Config` can by a dictionary and a list, the hash values from array/dict conversion should be combined (See DebugString in #3304).

filebeat/prospector/prospector.go
```diff
@@ -31,18 +33,20 @@ type Prospector struct {
 	wg               sync.WaitGroup
 	channelWg        sync.WaitGroup // Separate waitgroup for channels as not stopped on completion
 	harvesterCounter uint64
+	Hash             uint64
```
urso: how about `s/Hash uint64/ID ProspectorID` with `ProspectorID string`? e.g. encode the prospector type as well: `<prospector-type>-<config-hash>`. if we ever introduce named prospectors the string ca be easily adapted to `<prospector-type>-<name>-<config-hash>`.

filebeat/crawler/crawler.go
ruflin: Yes, and this is intentionally. As long as there are no overlaps in inode/device between the harvesters started of a prospector, this will have no effect. Getting the newest states from the registrar is needed when dynamically starting prospectors. Currently we expect from the user that he makes sure there is no overlap of harvesters. In the future we should introduce a registry for harvesters to detect such cases and warn a user or not start a harvester, if there is already running one with the same state in an other prospector. 

So there is no change in behaviour here (or there shouldn't be). Perhaps I missed something :-(

filebeat/crawler/crawler.go
ruflin: Oh, `config.Enabled()` is very convenient. Forgot about this one.

filebeat/crawler/crawler.go
ruflin: I added a simple check and renamed it to prospector.ID.

filebeat/prospector/prospector.go
ruflin: As the type has to be configured, it is also hashed and and part of the ID. To better understand what an ID stands for I prefer that we log the config of a prospector with its identifier instead. I think that brings more usability to the user.

filebeat/prospector/prospector.go
ruflin: yeah, that can be removed now. We had this in because of a previous bug.

filebeat/prospector/prospector.go
ruflin: I would really like to see a `.Hash()` method so we have one place to change the hashing logic if needed. I can get this done in a later PR when I plan to unify reloading in filebeat and metricbeat.

filebeat/crawler/crawler.go
urso: I worry a little about potential races between state updates in registrar being very late while old (outdated) state being available to new dynamically started prospector. Can this generate duplicate events?

filebeat/crawler/crawler.go
```diff
@@ -6,61 +6,75 @@ import (
...
+	c.prospectors[prospector.ID] = prospector
+	c.wg.Add(1)
+
+	go func() {
+		defer func() {
+			c.wg.Done()
+			logp.Debug("crawler", "Prospector stopped: %v", prospector.ID)
+		}()
+		logp.Debug("crawler", "Starting prospector: %v", prospector.ID)
+		prospector.Run()
```
urso: how about:

```
go func() {
  logp.Debug("crawler", "Starting prospector: %v", prospector.ID)
  defer logp.Debug("crawler", "Prospector stopped: %v", prospector.ID)

  defer c.wg.Done()
  prospector.Run()
}
```

filebeat/crawler/crawler.go
ruflin: Is your concern, that a prospector is stopped which handled the file `a.log` and 2 seconds later an other prospector is started that handles `a.log`? You are right that this could become an issue, even with a central registry for all harvesters, as the harvester could already be finished but the registry is not updated yet. I think a potential solution here is, that as soon as we have a registry, the harvester add itself to the registry and in the registry harvesters are removed then `Finished` is set to true. So as long as the data for the completion of the harvester hasn't arrived at the registry, no new harvester can be started.

Are we on the same page that for this PR this has not effect yet, right?

filebeat/crawler/crawler.go
ruflin: sounds also good to me.

PR #3310, opened 2017-01-09
Skip the modules test on the 2.x environment

tsg:
ES 2.x doesn't have Ingest Node, so there's no chance to work. In the
non-prototype phase, we should find a way to detect this situation and
inform the user properly.


tsg:
I tested the 2x testing env locally.

ruflin:
@tsg Probably worth to backport at least to 5.2 so we can run the 2x builds there. I added the backport label. Please remove again if you think it is not needed.

Issue #197, opened 2015-08-11
New protocol support: Ceph API & OSD


portante:
Place holder for now, but to start would like to see metrics on:
- top operations
- top clients
- error rates
- data rates

Looking to do the contributions myself if somebody does not get to it first.

amandahla:
I'm working on a module for MetricBeat here:
https://github.com/amandahla/beats/tree/master/metricbeat/module/ceph

I do accept suggestions and help. :-)

I'm using Telegraf as reference:
https://github.com/influxdata/telegraf/blob/master/plugins/inputs/ceph/ceph.go

monicasarbu:
@amandahla I am happy to see your contribution to Metricbeat. I think the best would be to open a PR with the `ceph` module, so we can easily add suggestions there.

amandahla:
Thank you @monicasarbu !I'm still working but I opened a PR now.

PR #3311, opened 2017-01-09
Support to CEPH input.

amandahla:
Using as reference Telegraf (https://github.com/influxdata/telegraf/blob/master/plugins/inputs/ceph/ceph.go). Working in progress.
#197


metricbeat/module/ceph/_meta/config.yml
```diff
@@ -0,0 +1,30 @@
+- module: ceph
+  metricsets: ["perf"]
+  enabled: true
+  period: 1s
+
+ceph:
```
ruflin: Is this working? This looks like invalid yaml. All config options related to a metricset should be in the module itself.

metricbeat/metricbeat.yml
```diff
@@ -1,105 +0,0 @@
-###################### Metricbeat Configuration Example #######################
```
ruflin: This file should not be removed. Running `make update` should generate it again.

metricbeat/module/ceph/_meta/config.yml
amandahla: Actually no. I commited a new version with this file corrected now.

metricbeat/metricbeat.yml
amandahla: Thanks, new commit with this corrected.

amandahla:
Hi Ruflin, thanks for the review. I prefer a detailed look because everything is working but it can really be better. I had difficult to deal with json output from ceph and only tested with a container (https://hub.docker.com/r/ceph/demo/).
Any help / review is appreciated. :-)

ruflin:
I will have a closer look. Please make sure to run `make fmt` so all code adheres to the coding standards.

metricbeat/metricbeat.full.yml
```diff
@@ -90,6 +90,38 @@ metricbeat.modules:
   # Password of hosts. Empty by default
   #password: test123
 
+#-------------------------------- ceph Module --------------------------------
+- module: ceph
+  metricsets: ["perf"]
+  enabled: true
+  period: 1s
+
+ceph:
```
ruflin: This part of the config must be moved under the module. I assume it doesn't work the way it is at the moment.

metricbeat/metricbeat.full.yml
```diff
@@ -90,6 +90,38 @@ metricbeat.modules:
...
+  #
+  #####
+  ### MetricSets status,df and poolio gather statistics via ceph commands, requires ceph_user and ceph_config
+  #####
+  # Ceph user to authenticate as, ceph will search for the corresponding keyring
+  # e.g. client.admin.keyring in /etc/ceph
+  # Consult the ceph documentation for more detail on keyring generation.
+  user = "client.admin"
+  # Ceph configuration to use to locate the cluster
+  config_path = "/etc/ceph/ceph.conf"
```
ruflin: In general we recommend metricbeat to connect to each node and get the specific metrics for each node instead from the cluster. What is inside this file? You should use `hosts: ["..."]` for that if possible.

metricbeat/module/ceph/_meta/config.yml
```diff
@@ -0,0 +1,25 @@
+- module: ceph
+  metricsets: ["perf","status","df","osdpoolstats"]
+  enabled: true
+  period: 5s
+
+  ceph:
```
ruflin: what is this line for? Forgot to remove?

metricbeat/module/ceph/_meta/config.yml
```diff
@@ -0,0 +1,25 @@
+- module: ceph
+  metricsets: ["perf","status","df","osdpoolstats"]
+  enabled: true
+  period: 5s
+
+  ceph:  
+  # location of ceph binary
```
ruflin: I suggest to split up the config into config.full.yml and config.yml and only have the extensive options in the full file. See other metricsets for examples.

metricbeat/module/ceph/_meta/config.yml
```diff
@@ -0,0 +1,25 @@
+- module: ceph
+  metricsets: ["perf","status","df","osdpoolstats"]
+  enabled: true
+  period: 5s
```
ruflin: The default we use is 10s

metricbeat/module/ceph/_meta/config.yml
```diff
@@ -0,0 +1,25 @@
+- module: ceph
+  metricsets: ["perf","status","df","osdpoolstats"]
```
ruflin: Perhaps we can split up the PR in multiple parts to have one metricset first and the others to follow. This will make reviews easier and should help to get a first working PR in faster.

metricbeat/module/ceph/config.go
```diff
@@ -0,0 +1,15 @@
+package ceph
+
+type Config struct {
+	CEPH *CephConfig `config:"ceph"`
+}
+
+type CephConfig struct {
+        BinaryPath             string `config:"binary_path"	validate:"required"`
```
ruflin: I don't think all these are required if we have defaults for them? If not configured we can fall back to the defaults, right?

metricbeat/module/ceph/config.go
```diff
@@ -0,0 +1,15 @@
+package ceph
+
+type Config struct {
```
ruflin: This should not be needed

metricbeat/module/ceph/df/data.go
```diff
@@ -0,0 +1,78 @@
...
+
+func eventsMapping(input string) []common.MapStr {
+
+	data := make(map[string]interface{}) 
+        err := json.Unmarshal([]byte(input), &data)
+        if err != nil {
+                logp.Err("An error occurred while parsing data for getting ceph df: %v", err)
+        }
+
+	stats_fields, ok := data["stats"].(map[string]interface{})
```
ruflin: we use `statsFields` etc as naming ocnvention.

metricbeat/module/ceph/_meta/docs.asciidoc
```diff
@@ -0,0 +1,9 @@
+== ceph Module
+
+This is the ceph Module.
+
+We used code/reference from:
+https://github.com/influxdata/telegraf/blob/master/plugins/inputs/ceph/ceph.go
```
ruflin: We should make sure to not just copy / paste code from here. Or was this also created by you?

metricbeat/module/ceph/df/_meta/fields.yml
```diff
@@ -0,0 +1,9 @@
+- name: df
+  type: group
+  description: >
+    df
+  fields:
+    - name: example
```
ruflin: Fields docs have to be added. This will also make it easier for us to understand what kind of data is provided.

metricbeat/module/ceph/df/df.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+type MetricSet struct {
+	mb.BaseMetricSet
+	cfg *ceph.CephConfig
+}
+
+func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
+
+	config := ceph.Config{}
```
ruflin: All the Ceph metricsets should be marked experimental first. We do this for all metricsets for the first release. See other metricsets for examples.

metricbeat/module/ceph/osdpoolstats/_meta/data.json
```diff
@@ -0,0 +1,19 @@
+{
+    "@timestamp":"2016-05-23T08:05:34.853Z",
+    "beat":{
+        "hostname":"beathost",
+        "name":"beathost"
+    },
+    "metricset":{
+        "host":"localhost",
+        "module":"mysql",
```
ruflin: you can generate this files by running `make generate-json` if you have created the `integration` test files which are responsible to create the data (see other metricsets). This requires a working docker environment.

metricbeat/metricbeat.full.yml
amandahla: Moved.

metricbeat/metricbeat.full.yml
amandahla: I used telegraf as reference and that only collects performance metrics from the MON and OSD nodes in a Ceph storage cluster. I can change the description, perhaps.

metricbeat/module/ceph/_meta/config.yml
amandahla: Good ideia. But I must warn that all them work pretty similar. :-)

metricbeat/module/ceph/_meta/config.yml
amandahla: Changed.

metricbeat/module/ceph/_meta/config.yml
amandahla: Changed. Useless line, sorry.

metricbeat/module/ceph/config.go
amandahla: Changed.

metricbeat/module/ceph/config.go
amandahla: Yes, you're right. Changed also.

metricbeat/module/ceph/df/data.go
amandahla: Changed.

metricbeat/module/ceph/_meta/docs.asciidoc
amandahla: Actually I copy/paste code and then adapted to collect metrics. It's really similar but not exactly equal. That's why I wrote this on docs.asciidoc. Is it a problem?

amandahla: Ah: This referente on telegraf was not created by me.

metricbeat/module/ceph/_meta/config.yml
amandahla: Working.

metricbeat/module/ceph/df/_meta/fields.yml
amandahla: Working.

metricbeat/module/ceph/df/df.go
amandahla: Working.

metricbeat/module/ceph/osdpoolstats/_meta/data.json
amandahla: Working.

metricbeat/module/ceph/_meta/docs.asciidoc
ruflin: I think we need to be careful about the licenses here. Getting inspired by some code is one thing, copy / pasting it an other. If we have the same code, we should use it as a library ;-)

metricbeat/metricbeat.full.yml
ruflin: I'm not aware of the inner workings of CEPH. Can you share some details on what MON and OSD nodes are?

metricbeat/module/ceph/_meta/config.yml
ruflin: Do it the way it better works for you.

ruflin:
@amandahla Ping me when it is ready for an other round of reviews.

metricbeat/metricbeat.full.yml
amandahla: A Ceph Storage Cluster requires at least one MON and at least 2 OSD.

A OSD (Object Storage Daemon) handles data (storage, replication, recovery...) and provide some monitoring information to MON by checking others OSD for a heartbeat.

A MON (Monitor) handles the cluster state (maps, history...).

You can have this architecture on one server/node but just for tests. Real productions environments works in a distributed way. 

If you want metrics just for the daemons where metricbeat its collecting, you can use Admin Socket Stats. In our case here, it's the "perf" metricset.

If you want metrics for the whole cluster, you can use ceph commands. In our case here, it's the "status", "df" and "osdpoolstats" metricsets.

That's why I wrote those comments on config.yml. Depending on choosen metricset, you need just some parameters.

ruflin: So `perf` is a local stat that should be collect by all metricbeat instances and the other 3 are cluster stats, means only one of the instances connecting needs to collect them?

amandahla: "...means only one of the instances connecting needs to collect them?" Yes.
"If there are multiple MON, will they provide different data?" No, at least it's not expected this behavior.

metricbeat/module/ceph/_meta/docs.asciidoc
amandahla: I saw that Telegraf uses[ MIT License](https://github.com/influxdata/telegraf/blob/master/LICENSE). So, I understand that there is no problem to copy parts of the Software. Unfortunately, I don't think that it's possible to use as a library. 
To get a better view what kind copy I did, here it's a example. I change the code to generate events.
[Mine](https://github.com/amandahla/beats/blob/master/metricbeat/module/ceph/osdpoolstats/data.go) (line 40)
[Telegraf](https://github.com/influxdata/telegraf/blob/master/plugins/inputs/ceph/ceph.go) (line 462)

metricbeat/module/ceph/df/_meta/fields.yml
amandahla: Please, just to be sure: is there a way to generate fields automatically? Or just manually?

amandahla:
Hi @ruflin 
I made some changes here:
- Removed "perf" metricset because our CEPH Admin told me that isn't so useful. So we have now just the ones from cluster.
- Wrote fields.yml files.
- Wrote integration tests.
- Working on unity tests.
Please, can you review?

metricbeat/docker-compose.yml
```diff
@@ -46,13 +47,26 @@ services:
     volumes:
       - ${PWD}/..:/go/src/github.com/elastic/beats/
       - /var/run/docker.sock:/var/run/docker.sock
+      - datavolume:/etc/ceph
```
ruflin: Do we need these data volumes on the client side?

metricbeat/docker-compose.yml
```diff
@@ -46,13 +47,26 @@ services:
...
     command: make
     entrypoint: /go/src/github.com/elastic/beats/metricbeat/docker-entrypoint.sh
 
   # Modules
   apache:
     build: ${PWD}/module/apache/_meta
 
+  ceph:
+    image: ceph/demo
+    container_name: metricbeatceph
```
ruflin: Please do not give the container a special name. This should not be needed.

metricbeat/docker-compose.yml
```diff
@@ -91,3 +105,8 @@ services:
 
   zookeeper:
     image: jplock/zookeeper:3.4.8
+
+volumes:
```
ruflin: If these volumes are ceph specific, we should have better names. If possible, I would prefer not to have "global" volumes.

metricbeat/docs/modules/ceph.asciidoc
```diff
@@ -0,0 +1,57 @@
+////
+This file is generated! See scripts/docs_collector.py
+////
+
+[[metricbeat-module-ceph]]
+== ceph Module
+
+This is the ceph Module.
+
+We used code/reference from:
```
ruflin: If the reference is only used to get inspiration, these should be removed.

metricbeat/module/ceph/command.go
```diff
@@ -0,0 +1,45 @@
...
+
+	cmdArgs := []string{}
+	cmdArgsDefault := []string{"--conf", c.ConfigPath, "--name", c.User, "--format", "json"}
+	cmdArgs = append(cmdArgs, cmdArgsDefault...)
+	cmdArgs = append(cmdArgs, strings.Split(command, " ")...)
+
+	var output string
+
+	if strings.Contains(c.BinaryPath, "docker") {
+		cmdCeph := []string{"/usr/bin/ceph"}
```
ruflin: We should not execute binaries whenever possible. As far as I can see ceph also has a REST API to fetch monitoring data: http://ceph.com/planet/experimenting-with-the-ceph-rest-api/ This should be used to fetch the data. Calling a binary requires CEPH to be installed on the machine. This works in most cases but breaks for example in docker environments, where on container is used to monitor the other containers.

metricbeat/module/ceph/_meta/output-examples/status.json
```diff
@@ -0,0 +1,101 @@
+{
+   "health":{
+      "health":{
+         "health_services":[
+            {
+               "mons":[
+                  {
+                     "name":"prcdsrvv1619",
+                     "kb_total":6281216,
```
ruflin: Please check our naming conventions: https://www.elastic.co/guide/en/beats/libbeat/5.1/event-conventions.html

Whenever possible we should no use nested arrays. I think we need to have a closer look at the data set and potentially resturcture it to fit into metricbeat.

metricbeat/module/ceph/config.go
```diff
@@ -0,0 +1,32 @@
...
+}
+
+func CheckConfig() CephConfig {
+
+	var userConfig = CephConfig{}
+
+	var defaultConfig = CephConfig{
+		BinaryPath: "/usr/bin/ceph",
+		User:       "client.admin",
+		ConfigPath: "/etc/ceph/ceph.conf",
```
ruflin: If we use the REST API using config etc. becomes obsolete.

metricbeat/module/ceph/command.go
ruflin: BTW: Why is the execution different if it is in a container or not?

metricbeat/docker-compose.yml
amandahla: I'll study the rest api but I think that we can remove all this and don't use command any more.

metricbeat/docker-compose.yml
amandahla: Same as above :)
"I'll study the rest api but I think that we can remove all this and don't use command any more."

metricbeat/docker-compose.yml
amandahla: Same as above :)
I'll study the rest api but I think that we can remove all this and don't use command any more.

metricbeat/docs/modules/ceph.asciidoc
amandahla: Since I'll change to use Rest API, this reference will be no longer necessary indeed. I'll remove.

metricbeat/module/ceph/command.go
amandahla: If is a container, I needed to call the ceph command inside him so that's why was different. In container, command will be something like "docker exec -i container_name /usr/bin/ceph ...". 
With REST api, I can remove this too.

metricbeat/module/ceph/_meta/output-examples/status.json
amandahla: So far, I see that with REST API the output it's more simple. But I'll have a closer look at this.

metricbeat/module/ceph/config.go
amandahla: I agree. :)

ruflin:
@amandahla Keep me posted about the REST API

amandahla:
@ruflin Thanks for showing me the CEPH Rest API. I was so focused on use the same way that Telegraf did, that didn't even think about searching for another way.
Now I'll take a bit more time to change everything, but I hope to finish soon. I'll keep this PR open but when I finish, I'll open another, ok?

ruflin:
You can also just push the updated version to this branch. No worries if you overwrite the current commits. I think the title stays the same anyways ;-)

ruflin:
@amandahla Your last commit looks very promising. This heavily simplifies the implementation 👍  Let me know if you should have another detailed look.

amandahla:
Hi @ruflin
Thanks!
I think that I'll let this PR only with one MetricSet ("health") like you suggested before. Please, can you review?  
It's really much better to use the Ceph Rest API!

ruflin:
Quick note on the failing travis build: Run `make fmt` to clean up the source code  and `make update` to update all generated files. It should go green afterwards.

metricbeat/module/ceph/_meta/docs.asciidoc
```diff
@@ -0,0 +1,6 @@
+== ceph Module
+
+This is the ceph Module. Metrics are collected submitting HTTP GET requests to ceph-rest-api.
+Reference: http://docs.ceph.com/docs/master/man/8/ceph-rest-api/
+
+Thanks!
```
ruflin: Remove this line.

metricbeat/module/ceph/health/_meta/fields.yml
```diff
@@ -0,0 +1,69 @@
...
+        Map version
+    - name: cluster.timechecks.round
+      type: long
+      description: >
+        timecheck round
+    - name: cluster.timechecks.round_status
+      type: keyword
+      description: >
+        Status of the round
+    - name: mon.avail_percent
```
ruflin: I would change this to `mon.available.pct`. See https://www.elastic.co/guide/en/beats/libbeat/5.1/event-conventions.html

metricbeat/module/ceph/health/_meta/fields.yml
```diff
@@ -0,0 +1,69 @@
...
+      type: long
+      description: >
+        Available percent of the MON
+    - name: mon.health
+      type: keyword
+      description: >
+        Health of the MON
+    - name: mon.kb_avail
```
ruflin: `mon.available.kb`

metricbeat/module/ceph/health/_meta/fields.yml
```diff
@@ -0,0 +1,69 @@
...
+      type: long
+      description: >
+        Available KB of the MON
+    - name: mon.kb_total
```
ruflin: `mon.total.kb`

metricbeat/module/ceph/health/_meta/fields.yml
```diff
@@ -0,0 +1,69 @@
...
+      type: long
+      description: >
+        Total KB of the MON
+    - name: mon.kb_used
```
ruflin: `mon.used.kb`

metricbeat/module/ceph/health/_meta/fields.yml
```diff
@@ -0,0 +1,69 @@
...
+        Used KB of the MON
+    - name: mon.last_updated
+      type: date
+      description: >
+        Time when was updated
+    - name: mon.name
+      type: keyword
+      description: >
+        Name of the MON
+    - name: mon.store_stats.bytes_log
```
ruflin: mon.store_stats.log.bytes

metricbeat/module/ceph/health/_meta/fields.yml
```diff
@@ -0,0 +1,69 @@
...
+      type: long
+      description: >
+        Log bytes of MON
+    - name: mon.store_stats.bytes_misc
```
ruflin: mon.store_stats.misc.bytes

You can also add `format: bytes` to all bytes value, so Kibana will automatically know how to convert it to MB etc.

```
    - name: mon.store_stats.sst.bytes
      type: long
      description: >
        SST bytes of MON
      format: bytes
```

metricbeat/module/ceph/health/_meta/fields.yml
```diff
@@ -0,0 +1,69 @@
...
+      type: long
+      description: >
+        Misc bytes of MON
+    - name: mon.store_stats.bytes_sst
```
ruflin: same with the bytes for the following two

metricbeat/module/ceph/health/data.go
```diff
@@ -0,0 +1,119 @@
+package health
+
+import (
+	"encoding/json"
+	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/logp"
+	"io"
```
ruflin: we normally but the external imports on the two following and empty line before the internal once

metricbeat/module/ceph/health/data.go
```diff
@@ -0,0 +1,119 @@
...
+
+	var d HealthRequest
+	err := json.NewDecoder(body).Decode(&d)
+	if err != nil {
+		logp.Err("Error: ", err)
+	}
+
+	events := []common.MapStr{}
+
+	event := common.MapStr{
```
ruflin: You could probably use the schema package here to do the event conversion, but we can do this cleanup in a second step.

metricbeat/module/ceph/health/data.go
```diff
@@ -0,0 +1,119 @@
...
+		"cluster.overall_status": d.Output.OverallStatus,
+		"cluster.timechecks": common.MapStr{
```
ruflin: You must use here `"cluster" : common.MapStr{ "timechecks" : common.MapStr{ ...` as otherwise the event will not be compatible with elasticsearch 2.x version which do not support dots in fields.

amandahla:
Thanks @ruflin !
I made the changes that you asked and created a unity test. I hope that will be ok now (I always forget the make fmt/update, sorry!).

metricbeat/module/ceph/health/health_test.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/elastic/beats/libbeat/common"
+	mbtest "github.com/elastic/beats/metricbeat/mb/testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestFetchEventContents(t *testing.T) {
+	absPath, err := filepath.Abs("./testdata/")
+
+	response, err := ioutil.ReadFile(absPath + "/sample_response.json")
```
ruflin: This is a great way to test the fetcher with predefined data. We should apply the same logic to other metricsets.

ruflin:
jenkins, test it

amandahla:
Hello @ruflin . I'm thinking about start on a new metricset now. Please, is this one ok so I can use as reference?

ruflin:
@amandahla will review just now and let you know.

ruflin:
@amandahla Just merged it. I will do a follow up PR with some cleanup and will ping you on it. So you can take this directly into account in your next metricset. Thanks for pushing this forward.

metricbeat/module/ceph/health/data.go
```diff
@@ -0,0 +1,138 @@
...
+					"value":  d.Output.Timechecks.Round,
+					"status": d.Output.Timechecks.RoundStatus,
+				},
+			},
+		},
+	}
+
+	events = append(events, event)
+
+	for _, HealthService := range d.Output.Health.HealthServices {
```
ruflin: @amandahla I just realised now that we have a problem here. We send two different event types as part of the same metricset. We need to extract this part here into its own metricset. Not should war we should call it. `monitor`?

amandahla: @ruflin Sorry, I didn't notice that would be a problem because all data is related and is obtained with only one command.
Answering your question: Yes, I think that "monitor" its good.

ruflin: @amandahla As we send it here in separate events anyways, it doesn't make a difference if it is a different metricset. Can you open a PR with an additional metricset? Sorry that I missed that one during the review.

PR #3321, opened 2017-01-10
Fix broken Edit links

dedemorton:
This PR fixes the broken Edit links that pointed to shared (single-sourced) content in libbeat. 

Because the macro that we use to set the Edit url doesn't accept variables in the path (couldn't use `{doc-branch}`), I decided to point edits to `master` rather than a specific branch name. This in keeping with our preference for having PRs submitted to the `master` branch and then back/forward ported. It's also consistent with how Logstash handles doc edits.

All the Edit links in the book will now point to the `master` branch. The one main drawback of this approach is that the user might see a problem that exists in the current doc, but has been fixed in `master`. Also, the content in `master` won't always match what the user sees in the published doc (especially as the doc ages), and it's possible that the link will no longer work if we delete the file in `master`. 

It's the best solution we have for now, though, unless we want to update all the URL paths in all these links whenever we cut a new branch.

NOTE: You'll probably notice that there area  couple of instances where I set the Edit url to the location of the shared logstash config doc even though there is a corresponding local file. The local file is just a heading that wraps the shared content, so I figured it was better just to point users to the shared content.


packetbeat/docs/index.asciidoc
```diff
@@ -19,49 +19,71 @@ include::../../libbeat/docs/version.asciidoc[]
...
+pass::[<?edit_url https://github.com/elastic/beats/edit/master/packetbeat/docs/packetbeat-filtering.asciidoc ?>]
 include::./packetbeat-filtering.asciidoc[]
 
+pass::[<?edit_url https://github.com/elastic/beats/edit/master/libbeat/docs/shared-config-ingest.asciidoc ?>]
 include::../../libbeat/docs/shared-config-ingest.asciidoc[]
 
+pass::[<?edit_url https://github.com/elastic/beats/edit/master/packetbeat/docs/packetbeat-geoip.asciidoc ?>]
 include::./packetbeat-geoip.asciidoc[]
 
+pass::[<?edit_url https://github.com/elastic/beats/edit/master/libbeat/docs/shared-logstash-config.asciidoc ?>]
```
monicasarbu: I think here it should point to: https://github.com/elastic/beats/edit/master/packetbeat/docs/shared-logstash-config.asciidoc

monicasarbu: Sorry, it looks like your link is working: https://github.com/elastic/beats/edit/master/libbeat/docs/shared-logstash-config.asciidoc

PR #3322, opened 2017-01-10
Fix bug when migrating old states, 2 restarts are required

ruflin:
The state of migrated states was not properly updated in the registry file. This lead to the issue that after the first restart, the states were migrated but the prospector assumed the states were not finished and didn't start harvesting. A second restart resolved the problem.


ruflin:
This bug does not affect 5.0.0 but all follow up version as the issue was introduced here: https://github.com/elastic/beats/commit/76e8f99f45ab225bcc493f31669d461acce8c7fb

CHANGELOG.asciidoc
```diff
@@ -43,6 +43,7 @@ https://github.com/elastic/beats/compare/v5.1.1...master[Check the HEAD diff]
 
 *Filebeat*
 - Fix registry cleanup issue when files falling under ignore_older after restart. {issue}2818[2818]
+- Fix registry migration issue from old states were files were only harvested after second restart.
```
tsg: Link to the issue/PR, please.

ruflin: will do a second commit. Linking to PR is always tricky before it is created :-)

ruflin:
jenkins, retest it

PR #3326, opened 2017-01-10
Fix bug when migrating old states, 2 restarts are required (#3322)

ruflin:
The state of migrated states was not properly updated in the registry file. This lead to the issue that after the first restart, the states were migrated but the prospector assumed the states were not finished and didn't start harvesting. A second restart resolved the problem.

(cherry picked from commit fb47507f73e59a8e693c1014d2225a0c95f19fa1)


ruflin:
jenkins, retest it

PR #3328, opened 2017-01-10
Always use absolute path for event and registry

ruflin:
During some testing with filebeat I realised that when a relative path glob is put into the filebeat config, the event will contain the relative path and also the state. In most cases this should not be an issue and so far no issues were reported.

For the state itself it is not an issue as they are compared based on inode/device. It could become an issue on restart in case a config was changed from a relative to an absolute path and the prospector does not detect, that the state would belong to the same prospector. This could also have an affect when migrating to this solutions. Old states could be left over in the registry file. But this requires, that someone was using relative paths before.

### DIFF
--- a/filebeat/_meta/fields.common.yml
+++ b/filebeat/_meta/fields.common.yml
@@ -7,7 +7,7 @@
       type: keyword
       required: true
       description: >
-        The file from which the line was read. This field contains the full path to the file.
+        The file from which the line was read. This field contains the absolute path to the file.
         For example: `/var/log/system.log`.
 
     - name: offset
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -17,6 +17,7 @@ https://github.com/elastic/beats/compare/v5.1.1...master[Check the HEAD diff]
 - Change beat generator. Use `$GOPATH/src/github.com/elastic/beats/script/generate.py` to generate a beat. {pull}3452[3452]
 
 *Filebeat*
+- Always use absolute path for event and registry. This can lead to issues when relative paths were used before. {pull}3328[3328]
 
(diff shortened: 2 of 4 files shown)


filebeat/harvester/log.go
```diff
@@ -137,6 +137,7 @@ func (h *Harvester) Harvest(r reader.Reader) {
 			event.InputType = h.config.InputType
 			event.DocumentType = h.config.DocumentType
 			event.JSONConfig = h.config.JSON
+			event.Name = h.state.Fileinfo.Name()
```
ruflin: I added this for discussion as it was requested several times in the past.

filebeat/_meta/fields.common.yml
```diff
@@ -10,6 +10,11 @@
         The file from which the line was read. This field contains the full path to the file.
```
andrewkroh: Can you drop the "absolute" keyword somewhere in the `source` description.

ruflin: I replaced `full` with `absolute`.

filebeat/_meta/fields.common.yml
```diff
@@ -7,9 +7,14 @@
       type: keyword
       required: true
       description: >
-        The file from which the line was read. This field contains the full path to the file.
+        The file from which the line was read. This field contains the absolute path to the file.
         For example: `/var/log/system.log`.
 
+    - name: name
```
tsg: Note sure about calling it "name" since it might ambiguous. How about "filename"?

ruflin: agree

ruflin:
I removed the `Name` part from this PR again as it is not directly related to this change.

PR #3330, opened 2017-01-10
Fixes for dev-tool/jenkins_ci

andrewkroh:
Fix issues with the script for executing CI builds.

- Use absolute path to gimme inside of scripts to avoid requiring $HOME/bin be on the PATH.
- export the RACE_DETECTOR variable from the jenkins_ci script so that it is picked up by make.
- Disable cross-compile for darwin/arm by default.

### DIFF
--- a/dev-tools/common.bash
+++ b/dev-tools/common.bash
@@ -44,7 +44,8 @@ install_gimme() {
     chmod +x ${HOME}/bin/gimme
   fi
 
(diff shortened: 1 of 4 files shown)


ruflin:
LGTM. Not sure how the metricbeat appveyor failing test can be related.

andrewkroh:
Yeah, I don't think the Appveyor failure is related. It doesn't use any of these files.

PR #3333, opened 2017-01-10
Filebeat modules: first phase of the Go implementation

tsg:
This rewrites the following parts of the prototype in Go:

* Reading of the filesets
* Evaluation of variables, including the builtin variables
* Creation of the prospectors configurations and starting them
* Overriding of the variables from the config file and from the command line
* The -module and -M CLI flags
* The ability to override variables from multiple filesets using wildcards, for example: `-M "nginx.*.var.paths=[/var/log/syslog*]"`

At this point, the system tests are still using the filebeat.py wrapper, but
only for loading the ingest node pipeline. The prospector configuration part
and the overriding of settings is done by Filebeat itself.

Part of #3159.

### DIFF
--- a/filebeat/fileset/config.go
+++ b/filebeat/fileset/config.go
@@ -1,0 +1,19 @@
+package fileset
+
+// ModuleConfig contains the configuration file options for a module
+type ModuleConfig struct {
+	Module  string `config:"module"     validate:"required"`
+	Enabled *bool  `config:"enabled"`
+
+	// Filesets is inlined by code, see mcfgFromConfig
+	Filesets map[string]*FilesetConfig
+}
+
+// FilesetConfig contains the configuration file options for a fileset
+type FilesetConfig struct {
+	Enabled    *bool                  `config:"enabled"`
+	Var        map[string]interface{} `config:"var"`
+	Prospector map[string]interface{} `config:"prospector"`
+}
+
+var defaultFilesetConfig = FilesetConfig{}
--- a/filebeat/fileset/fileset.go
+++ b/filebeat/fileset/fileset.go
@@ -1,0 +1,257 @@
+/*
+Package fileset contains the code that loads Filebeat modules (which are
+composed of filesets).
+*/
+
+package fileset
+
+import (
+	"bytes"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+	"text/template"
+
+	"github.com/elastic/beats/libbeat/common"
+)
+
+// Fileset struct is the representation of a fileset.
+type Fileset struct {
+	name       string
+	mcfg       *ModuleConfig
+	fcfg       *FilesetConfig
+	modulePath string
+	manifest   *manifest
+	vars       map[string]interface{}
+}
+
+// New allocates a new Fileset object with the given configuration.
+func New(
+	modulesPath string,
+	name string,
+	mcfg *ModuleConfig,
+	fcfg *FilesetConfig) (*Fileset, error) {
+
+	modulePath := filepath.Join(modulesPath, mcfg.Module)
+	if _, err := os.Stat(modulePath); os.IsNotExist(err) {
+		return nil, fmt.Errorf("Module %s (%s) doesn't exist.", mcfg.Module, modulePath)
+	}
+
+	return &Fileset{
+		name:       name,
+		mcfg:       mcfg,
+		fcfg:       fcfg,
+		modulePath: modulePath,
+	}, nil
+}
+
+// Read reads the manifest file and evaluates the variables.
+func (fs *Fileset) Read() error {
+	var err error
+	fs.manifest, err = fs.readManifest()
+	if err != nil {
+		return err
+	}
+
+	fs.vars, err = fs.evaluateVars()
+	if err != nil {
+		return err
+	}
+
+	pipeline_id, err := fs.getPipelineID()
+	if err != nil {
+		return err
+	}
+	fs.vars["beat"] = map[string]interface{}{
+		"pipeline_id": pipeline_id,
+	}
+
+	return nil
+}
+
+// manifest structure is the representation of the manifest.yml file from the
+// fileset.
+type manifest struct {
+	ModuleVersion  string                   `config:"module_version"`
+	Vars           []map[string]interface{} `config:"var"`
+	IngestPipeline string                   `config:"ingest_pipeline"`
+	Prospector     string                   `config:"prospector"`
+}
+
+// readManifest reads the manifest file of the fileset.
+func (fs *Fileset) readManifest() (*manifest, error) {
+	cfg, err := common.LoadFile(filepath.Join(fs.modulePath, fs.name, "manifest.yml"))
+	if err != nil {
+		return nil, fmt.Errorf("Error reading manifest file: %v", err)
+	}
+	var manifest manifest
+	err = cfg.Unpack(&manifest)
+	if err != nil {
+		return nil, fmt.Errorf("Error unpacking manifest: %v", err)
+	}
+	return &manifest, nil
+}
+
+// evaluateVars resolves the fileset variables.
+func (fs *Fileset) evaluateVars() (map[string]interface{}, error) {
+	var err error
+	vars := map[string]interface{}{}
+	vars["builtin"], err = fs.getBuiltinVars()
+	if err != nil {
+		return nil, err
+	}
+
+	for _, vals := range fs.manifest.Vars {
+		var exists bool
+		name, exists := vals["name"].(string)
+		if !exists {
+			return nil, fmt.Errorf("Variable doesn't have a string 'name' key")
+		}
+
+		value, exists := vals["default"]
+		if !exists {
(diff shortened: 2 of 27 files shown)


filebeat/fileset/config.go
```diff
@@ -0,0 +1,17 @@
+package fileset
+
+// ModuleConfig contains the configuration file options for a module
+type ModuleConfig struct {
+	Module   string                    `config:"module"     validate:"required"`
+	Enabled  *bool                     `config:"enabled"`
+	Filesets map[string]*FilesetConfig `config:"filesets"`
```
ruflin: Is this also required?

filebeat/fileset/fileset.go
```diff
@@ -0,0 +1,255 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("Error creating config from prospector overrides: %v", err)
+		}
+		cfg, err = common.MergeConfigs(cfg, overrides)
+		if err != nil {
+			return nil, fmt.Errorf("Error applying config overrides: %v", err)
+		}
+	}
+
+	return cfg, nil
```
ruflin: a debug log entry here outputting the final merged config would be nice to have.

filebeat/fileset/flags.go
```diff
@@ -0,0 +1,59 @@
...
+	"flag"
+	"fmt"
+	"strings"
+
+	"github.com/elastic/beats/libbeat/common"
+)
+
+// Modules related command line flags.
+var (
+	modulesFlag     = flag.String("module", "", "List of enabled modules (comma separated)")
```
ruflin: should this be `modules`?

filebeat/fileset/modules.go
```diff
@@ -0,0 +1,190 @@
...
+
+	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/paths"
+)
+
+type ModuleRegistry struct {
+	registry map[string]map[string]*Fileset // module -> fileset -> Fileset
+}
+
+// newModuleRegistry readsy and loads the configured module into the registry.
```
ruflin: s/readsy/reads

filebeat/module/apache2/access/config/access.yml
```diff
@@ -1,9 +1,9 @@
-- input_type: log
-  paths:
-  {%- for path in paths %}
-   - {{path}}
-  {%- endfor %}
-  exclude_files: [".gz$"]
-  fields:
-    source_type: apache2-access
-    pipeline_id: {{beat.pipeline_id}}
+input_type: log
```
ruflin: This change means each fileset can max have 1 prospector (which I think is a good limitation)

tsg: Yeah, multiple prospectors didn't really work out with the requirement of being able to override some of the prospector settings when needed.

filebeat/fileset/flags.go
tsg: I had "modules" in the first prototype, then thought that "module" would be nicer for the most common case. But you're right, I think "modules" makes more sense overall.

filebeat/fileset/config.go
tsg: Not sure I understand the question. Whether we need `Filesets` part of the `ModuleConfig`?

ruflin: Isn't at least 1 fileset required in the config? Talking about the `validate` part.

tsg: Ah, gotcha :). It's actually that all filesets are enabled by default and you need to explicitly disable them. So a simple valid config would be:

```
modules:
- name: nginx
```

If you want to disable the error fileset, you need to write:

```
modules:
- name: nginx
  filesets.error.enabled: false
```

Note that I went for having the filesets a dictionary with all settings, unlike in Metricbeat were it's an array with the names. The reason is that I think it's going to be rare to disable filesets, and there lots of fileset specific settings.

ruflin: Interesting. Is there a reason you prefix the fileset config options with `fileset`. Namespace conflicts? The convention in metricbeat is to just use the metricset name, so above it would be `error.enabled`.

An other option about would be to use `filesets: ["access"]` if you want only one. If `filesets` is not around, all sets are loaded. I agree this makes sense in the filebeat module case. But this change would require the previous change, as otherwise `filesets` would have 2 meanings.

ruflin: Side node: I like that the fileset can be disabled with the `enabled: false` flag. Makes it command line friendly.

tsg: The reason for prefixing with `filesets` is just to keep the implementation simple, in the sense that the ModuleConfig can be just:

```
type ModuleConfig struct {
	Module   string                    `config:"module"     validate:"required"`
	Enabled  *bool                     `config:"enabled"`
	Filesets map[string]*FilesetConfig `config:"filesets"`
}
```

And future module level settings can be added easily. In order to get rid of `filesets`, I'd need something like:

```
type ModuleConfig map[string]*common.Config
```

and then have code to extract the "enabled" and "name". I can look into doing that, although I'm not sure the extra code complexity is justified.

Btw, I wrote an overview on the user interaction here: https://github.com/elastic/beats/issues/3159

ruflin: I would vote for the additional complexity to replace filesets from the config and be consistent with metricbeat. This also make the command line options look more natural as there it is already without `filesets`. The nice part about having *common.Config that it provides already an `Enabled()` function, so not even unpacking is needed.

I didn't check the code details on this yet, but one solution is to work in general with *common.Config objects which are passed around and unpacking happens locally. So also each fileset will unpack its own config options

ruflin: @tsg Looking at https://github.com/elastic/beats/issues/3159

Shouldn't it be:

```
modules:
- name: nginx
  access:
    close_eof: true
```
I removed the prospector path above. So it is actually `map[string]ProspectorConfig`. Each module in the end is 1 prospector. Or are there other configs inside `access` that could occur?

tsg: The command lines don't include `filesets`, just the configuration. For the cli, the options are prefixed with `module.fileset`, so for example `nginx.access`. From there on, the CLI flags match the structure in the configuration file. So to set `close_eof` you would type `nginx.access.prospector.close_eof`. But anyway, I'll look into getting rid of `filesets`, that would be cleaner.

There are more things to a fileset config than the prospector. Each fileset has configurable variables, which can be used in the prospector template and for deciding the pipeline. For example, you can set `format: json` and that variable is used in all the places we need it. I'd expect the module variables to be what 90% users need, and changing something directly in the prospector is "expert" mode. The pipeline itself is also a configuration at the fileset level.

tsg:
@ruflin I added another [commit](https://github.com/elastic/beats/pull/3333/commits/38ce7dcffd8b4f0bcd9a5de1564f38eb03db5d90) to move all filesets one level up in the configuration (so get rid of the `filesets` key word). The added code complexity is acceptable, and the other main drawback is that we can't have `module` or `enabled` as a fileset name. That's not a problem now, but it might become one if we'll add more config options at the module level.

Overall, I think this is a good change.

tsg:
jenkins, test it

filebeat/module/nginx/_meta/config.full.yml
```diff
@@ -0,0 +1,33 @@
+- module: nginx
+  # Access logs
+  filesets.access:
+    enabled: true
+
+    # Format. Options are with_plugins and no_plugins
+    #var.pipeline: with_plugins
```
ruflin: I'm thinking if we should change this from `var` to `option`. I somehow dislike `var` as it is not very descriptive. It is a configuration option for the metricset. As these are "metricset" specific perhaps we can even remove the `var` completely and have it under `fileset.access.pipeline`. As most fileset have `pipeline` and `paths` these two could be basic options shared across all filesets like `ingest_pipeline`?

filebeat/fileset/fileset.go
```diff
@@ -0,0 +1,257 @@
...
+			}
+		}
+		return transformed, nil
+	}
+	return value, nil
+}
+
+// applyTemplate applies a Golang text/template
+func applyTemplate(vars map[string]interface{}, templateString string) (string, error) {
+	tpl, err := template.New("test").Parse(templateString)
```
tsg: I think that doesn't really matter, but also "test" is not really good since this is not a test :). I'll make it text, thanks.

filebeat/module/nginx/_meta/config.full.yml
tsg: I was considering inlining `var` as well,  like we did for `fileset`. That would be the nicest CLI experience but the chances of collisions are higher in there.

The reason I called them "vars" is that they can reference each other. But I think I'm good with `option` as well, have to think some more about it.

ruflin: My first thought was also to namespace it based on the conflicts. But I think the number of options will be quite small here and should not correlate with the existing once. As these are the variables for the template and (probably) rarely used, it's also ok if it is perhaps sometimes a little bit a longer name that prevents a conflict. It's great now to have prospector in it's own namespace :-)

PR #3334, opened 2017-01-10
Add NOTICE file for Beats

monicasarbu:
This PR contains a python script that generates the NOTICE file. There is a single NOTICE file for all the official Beats, and it's placed in the [beats repository](https://github.com/elastic/beats).

The NOTICE file includes the copyright information for all the 3rd party libraries used in the Beats project. It follows the rules:
- if the library has apache license, and if NOTICE file is present, include that text in the NOTICE file. 
- if the library has MIT, ISC or BSD license, include the whole copyright block in NOTICE file (meaning copyright + license)

You can pass a list of vendor directories where the NOTICE* and LICENSE* files of all 3rd party libraries are available:
```
python dev-tools/generate_notice.py vendor metricbeat/vendor
```

For the official Beats, you can run the following in the [beats repository](https://github.com/elastic/beats).

```
make notice
```

The NOTICE file can be generated by the community Beats by using the script. The NOTICE file is not added by the Beat generator.

UPDATE: The script accepts also the name of the Beat and the copyright owner, to customize the NOTICE file for the community Beats.

```
$ python dev-tools/generate_notice.py -h
usage: generate_notice.py [-h] [-b BEAT] [-c COPYRIGHT] vendors [vendors ...]

Generate the NOTICE file from a list of vendor directories

positional arguments:
  vendors               List of vendor directories

optional arguments:
  -h, --help            show this help message and exit
  -b BEAT, --beat BEAT  Beat name
  -c COPYRIGHT, --copyright COPYRIGHT
                        Copyright owner
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -113,3 +113,9 @@ upload-package:
 .PHONY: release-upload
 upload-release:
 	aws s3 cp --recursive --acl public-read build/upload s3://download.elasticsearch.org/beats/
+
+.PHONY: notice
+notice: 
+	python dev-tools/generate_notice.py .
+
+
--- a/dev-tools/generate_notice.py
+++ b/dev-tools/generate_notice.py
@@ -25,7 +25,7 @@ def get_library_name(license):
         path, x = os.path.split(path)
         if len(lib) == 0:
             lib = x
-        else:
+        elif len(x) > 0:
             lib = x + "/" + lib
 
     return lib
--- a/dev-tools/generate_notice.py
+++ b/dev-tools/generate_notice.py
@@ -1,0 +1,98 @@
+import glob
+import sys
+import os
+import datetime
+import argparse
+
+
+def read_file(filename):
+
+    if not os.path.isfile(filename):
+        print("File not found {}".format(filename))
+        return ""
+
+    with open(filename, 'r') as f:
+        file_content = f.read()
+        return file_content
+
+
+def get_library_name(license):
+
+    lib = ""
+    path = os.path.dirname(license)
+    # get the last three directories
+    for i in range(0, 3):
+        path, x = os.path.split(path)
+        if len(lib) == 0:
+            lib = x
+        else:
+            lib = x + "/" + lib
+
+    return lib
+
+
+def add_licenses(f, licenses):
+
+    for license in licenses:
+        for license_file in glob.glob(license):
+            f.write("\n--------------------------------------------------------------------\n")
+            f.write("{}\n".format(get_library_name(license_file)))
+            f.write("--------------------------------------------------------------------\n")
+            copyright = read_file(license_file)
+            if "Apache License" not in copyright:
+                f.write(copyright)
+            else:
+                # it's an Apache License, so include only the NOTICE file
+                f.write("Apache License\n\n")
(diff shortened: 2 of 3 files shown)


metricbeat/NOTICE
```diff
@@ -0,0 +1,1692 @@
+License /Users/monica/src/github.com/elastic/beats/metricbeat/vendor/github.com/mitchellh/hashstructure/LICENSE
```
ruflin: We should not include here the full path as it contains your private username.

dev-tools/generate_notice.py
```diff
@@ -0,0 +1,47 @@
...
+
+if __name__ == "__main__":
+
+	beat_dir = os.getcwd()
+	libbeat_dir = os.path.join(beat_dir, "../")
+	notice = os.path.join(beat_dir, "NOTICE")
+	vendors = [beat_dir, libbeat_dir]
+	licenses = []
+
+	for vendor in vendors:
```
ruflin: Will this also include vendor directories inside metricbeat modules?

metricbeat/NOTICE
monicasarbu: Yes, sure. The plan is to include only `github.com/mitchellh/hashstructure`.

ruflin:
This could also be interesting to see which repos do not contain a license file: https://github.com/pmezard/licenses

cyrilleverrier:
This PR is going to be useful for community beats as well :)

Could you please make sure that we (developers) can configure the output by using variables instead of `Elastic Beat` and `Copyright 2014-{} Elasticsearch BV` ?

monicasarbu:
@cyrilleverrier Thank you for such a quick feedback. I added two options (`-b` and `-c`) to configure the Beat name and the copyright owner. I have updated the description of the PR with the script usage.

Makefile
```diff
@@ -113,3 +113,9 @@ upload-package:
 .PHONY: release-upload
 upload-release:
 	aws s3 cp --recursive --acl public-read build/upload s3://download.elasticsearch.org/beats/
+
+.PHONY: notice
+notice: 
+	python dev-tools/generate_notice.py metricbeat/vendor vendor
```
ruflin: This does not include the vendor directories in the metricbeat modules? How about somthing like `find . -name vendor`?

dev-tools/generate_notice.py
```diff
@@ -0,0 +1,96 @@
...
+        file_content = f.read()
+        return file_content
+
+
+def get_library_name(license):
+
+    lib = ""
+    path = os.path.dirname(license)
+    # get the last three directories
+    for i in range(0, 3):
```
ruflin: Does this also work for dependencies not on Github?

Makefile
monicasarbu: Ah, good point. Then I will search in a given director for all the vendor directories.

dev-tools/generate_notice.py
monicasarbu: It gets only the last 3rd directories from the path, but the function should be able to get less than 3 directories as well. I adjusted the code a bit to accept also less than 3 directories (example windows/project)

PR #3335, opened 2017-01-10
Added support for types short and byte in templates

### LINKED ISSUE #3303: beats/libbeat/scripts/generate_template.py does not support numeric types integer, short and bytes
Followup for #3257, as suggested by @ruflin. 

I would like to use the numeric types `integer`, `short` and `byte` which are supported by Elasticsearch ([Numeric datatypes](https://www.elastic.co/guide/en/elasticsearch/reference/current/number.html)) in the templates for [Netsamplebeat](https://github.com/breml/netsamplebeat/). 
With this beat I collect samples of network packets (mainly layer 2, 3 and 4 information). This also includes fields like IPv4 TTL (8-bit, unsigned, could be saved to a field `short`), IPv6 hop limit (8-bit, unsigned, could be saved to a field `short`) or IPv6 payload length (16-bit, unsigned, could be saved to a field `integer`).

Currently `beats/libbeat/scripts/generate_template.py` only supports `long` (and `integer`, which is [converted](https://github.com/elastic/beats/blob/master/libbeat/scripts/generate_template.py#L228-L230) to `long`).

breml:
Fixes #3303

### DIFF
--- a/libbeat/scripts/generate_index_pattern.py
+++ b/libbeat/scripts/generate_index_pattern.py
@@ -55,7 +55,7 @@ def field_to_json(desc, path, output,
     }
     # find the kibana types based on the field type
     if "type" in desc:
-        if desc["type"] in ["half_float", "scaled_float", "float", "integer", "long"]:
+        if desc["type"] in ["half_float", "scaled_float", "float", "integer", "long", "short", "byte"]:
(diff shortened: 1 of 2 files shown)


ruflin:
jenkins, test it

PR #2433, opened 2016-09-01
Introduce simple string matcher

urso:
Match further optimizes special regular expressions, trying to replace
special regular patterns with more performance string searches.

benchmark results:

run 1:

```
BenchmarkBeginning/Matcher=Regex-4       1000000              1521 ns/op
BenchmarkBeginning/Matcher=Match-4      10000000               132 ns/op
BenchmarkBeginningSpace/Matcher=Regex-4                  1000000              1368 ns/op
BenchmarkBeginningSpace/Matcher=Match-4                 10000000               135 ns/op
BenchmarkBeginningDate/Matcher=Regex-4                   1000000              1757 ns/op
BenchmarkBeginningDate/Matcher=Match-4                   1000000              1889 ns/op
BenchmarkStringPatternRegex/Matcher=Regex-4              1000000              1314 ns/op
BenchmarkStringPatternRegex/Matcher=Match-4              5000000               266 ns/op
BenchmarkStringPatternDotStarRegex/Matcher=Regex-4                 30000             37084 ns/op
BenchmarkStringPatternDotStarRegex/Matcher=Match-4               5000000               257 ns/op
```

run 2:

```
BenchmarkBeginning/Matcher=Regex-4       1000000              1413 ns/op
BenchmarkBeginning/Matcher=Match-4      10000000               135 ns/op
BenchmarkBeginningSpace/Matcher=Regex-4                  1000000              1300 ns/op
BenchmarkBeginningSpace/Matcher=Match-4                 10000000               144 ns/op
BenchmarkBeginningDate/Matcher=Regex-4                   1000000              1918 ns/op
BenchmarkBeginningDate/Matcher=Match-4                   1000000              1727 ns/op
BenchmarkStringPatternRegex/Matcher=Regex-4              1000000              1300 ns/op
BenchmarkStringPatternRegex/Matcher=Match-4              5000000               262 ns/op
BenchmarkStringPatternDotStarRegex/Matcher=Regex-4                 50000             36425 ns/op
BenchmarkStringPatternDotStarRegex/Matcher=Match-4               5000000               254 ns/op
```

In `BenchmarkBeginningDate` benchmark the regular expression can not be optimized (yet?), resulting in `Regex` and `Match` tests, both executing full regular expression.

PR #3337, opened 2017-01-11
Add environment variables to the system process metricset

andrewkroh:
This PR adds the environment variables that were used to start the process to the data reported in the system process metricset. The data is added as a dictionary under the `system.process.env` key. Environment variables must be whitelisted using an array of regular expressions specified using `process.env.whitelist: []` in the module config.

This feature implemented for FreeBSD, Linux, and OS X.

Sample config:

```
metricbeat.modules:
- module: system
  metricsets: [process]
  process.env.whitelist: ['USER', 'PATH']
```

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -25,7 +25,7 @@ import:
   subpackages:
   - /difflib
 - package: github.com/elastic/gosigar
-  version: 171a3c9e31dde9688c154ba94be6cd5d8a78bf64
+  version: b49e01eb1e5c68c469392a63feaccae9352ceb12
 - package: github.com/elastic/procfs
   version: abf152e5f3e97f2fafac028d2cc06c1feb87ffa5
 - package: github.com/samuel/go-parser
--- a/metricbeat/_meta/beat.full.yml
+++ b/metricbeat/_meta/beat.full.yml
@@ -67,6 +67,10 @@ metricbeat.modules:
   # EXPERIMENTAL: cgroups can be enabled for the process metricset.
   #cgroups: false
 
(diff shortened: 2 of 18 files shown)


andrewkroh:
What kind of configuration options if any (like filtering, whitelist, blacklist, enabled) should we provide with this feature?

ruflin:
I would only add `process.env.enabled: true` as a config option. Filtering can be done with the processors if needed. I think with the processors we should also be able to only have the variables for certain processes and remove it in all other cases etc.

I'm unsure if we should enabled it by default. On the one hand the info can be quite useful, but it can also contain quite a list of defaults which are the same for all processes.

tsg:
@joegallo @alexbrasetvik FYI, we're going this direction for Metricbeat.

tsg:
I'd say we should have a whitelist and the feature should be disabled by default due to privacy concerns. Secrets are often stored in env, so we have to make it hard to leak them accidentally.

tsg:
To be clear, I mean a whitelist on the variable names, so something like `process.env.variables: ["PATH"]`.

andrewkroh:
I added `process.env.whitelist: []` to control what environment variables are added to the system process events. It accepts a list of regular expressions. When empty or unset, no environment variables are added to the process event.

This should be updated to the the ExactMatcher from #2433 when it merges.

This PR depends on elastic/gosigar#62 and needs to be updated after that merges to put the commit hash into the glide.yaml.

andrewkroh:
The PR is now updated with the final gosigar commit hash.

Issue #3273, opened 2017-01-02
Filebeat - 64Bit alignment issue on ARM


ikofler:
I tried to compile the latest code of filebeat on my Raspbarry Pi 3 (Arch ARM Linux, armv7h). The compilation was successful, however during runtime it crashed as follows

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x4 pc=0xaaf58]

goroutine 25 [running]:
panic(0x4f8a08, 0x10b22008)
/usr/lib/go/src/runtime/panic.go:500 +0x33c
sync/atomic.addUint64(0x10d501cc, 0x1, 0x0, 0x1, 0x0)
/usr/lib/go/src/sync/atomic/64bit_arm.go:31 +0x68
github.com/elastic/beats/filebeat/prospector.(*Prospector).startHarvester(0x10d50140, 0x10d26c60, 0x23, 0x0, 0x0, 0x0, 0x751cd8, 0x10d23170, 0x169c6, 0x0, ...)
/home/kofi/beats/src/github.com/elastic/beats/filebeat/prospector/prospector.go:239 +0x30c
github.com/elastic/beats/filebeat/prospector.(*ProspectorLog).scan(0x10bca0f0)
/home/kofi/beats/src/github.com/elastic/beats/filebeat/prospector/prospector_log.go:232 +0x3f4
github.com/elastic/beats/filebeat/prospector.(*ProspectorLog).Run(0x10bca0f0)
/home/kofi/beats/src/github.com/elastic/beats/filebeat/prospector/prospector_log.go:77 +0xd8
github.com/elastic/beats/filebeat/prospector.(*Prospector).Run(0x10d50140, 0x0)
/home/kofi/beats/src/github.com/elastic/beats/filebeat/prospector/prospector.go:156 +0x338
github.com/elastic/beats/filebeat/crawler.(*Crawler).Start.func1(0x10d26840, 0x0, 0x0, 0x10d50140)
/home/kofi/beats/src/github.com/elastic/beats/filebeat/crawler/crawler.go:59 +0xf4
created by github.com/elastic/beats/filebeat/crawler.(*Crawler).Start
/home/kofi/beats/src/github.com/elastic/beats/filebeat/crawler/crawler.go:60 +0x498

[...]

PR #3338, opened 2017-01-11
Ensure harvesterCounter 8-byte alignment (#3273)

spacewrangler:
harvesterCounter is accessed atomically and will fault on x86-32 or ARM
if not 8-byte aligned. See golang/go#599 for more details on why it
fails and https://golang.org/pkg/sync/atomic/#pkg-note-BUG for how
putting the field first in the struct fixes it.


ruflin:
jenkins, test it

filebeat/prospector/prospector.go
```diff
@@ -21,6 +21,7 @@ var (
 )
 
 type Prospector struct {
+	harvesterCounter uint64         // Must be 8-byte aligned. Ensured if first field in struct
```
ruflin: As it is quite likely that this is changed by accident in the future, I suggest to add a comment on top of the variable with the following content:

// harvesterCount MUST be first field in struct. See https://github.com/golang/go/issues/599

spacewrangler: Thanks for the feedback @ruflin. Should I make the changelog entry under `Added` or `Bugfixes`? I know you wrote `Added` above but thought I'd double check since this is a bug fix.

ruflin: I'm fine putting it under Bugfixes. 

Do you have some ideas on how to add a test? Would be great to have one.

tsg:
Thanks @spacewander for the fix! As we have a little bit of urgency around this, I merged it as is and  will add the changelog in a follow up PR.

PR #3339, opened 2017-01-11
Unify BEAT_DIR and BEAT_PATH, rename BEATNAME to BEAT_NAME

ruflin:
BEAT_PATH is not consistent between community beats and official beats. Both include the beatname in the path. BEAT_PATH and BEAT_DIR were both used with the same meaning. All BEAT_DIR were not changed to BEAT_PATH. This PR breaks with existing community beats if they depend on master.

* Rename all BEAT_DIR to BEAT_PATH
* Remove notes about community beats paths as part of help now
* Fix beat URL for official beats. For filebeat it was github.com/elastic/beats/filebeat which does not work. Now links point to product page and community beats will point to Github page if not changed.
* Rename BEATNAME to BEAT_NAME for consistency with other variables


ruflin:
@cyrilleverrier You could be interested in this change.

ruflin:
jenkins, package it

ruflin:
jenkins, retest it

PR #3344, opened 2017-01-11
Swarm node and service feature for metricbeat docker module (healthcheck)

sebastienmusso:
In order to monitor, collect metrics and autoscale app for our docker cluster farm, I develop node and service module for metricbeat docker module, furthermore I add healthcheck information for container module.

All theses changes have been done the same way as others docker component (a folder per component with metadata folder). 
Needed docker api type struct definition have been added in "fsouza/go-dockerclient/external"

These metrics could be usefull for application monitoring/scalling in clustered docker environment.


ruflin:
@sebastienmusso Thanks a lot for this contribution.

I'm not too familiar with docker swarm yet but that is what I understand:
 
* A node is a docker host machine inside swarm and can contain containers
* A service is a number of the same docker containers running in swarm

Based on the above I asked myself it these two metricsets belong into same module or in a new one. For the existing metricsets each metricset is additional information about a specific container. The above is a more global view and not necessarely specific to a container. So I'm thinking if this would belong into `docker_swarm` or `swarm` module? Or would you run one module with all metricsets enabled? Let me know what you think here.

For the healthchecks: Any chance to split these into a separate PR? Because it seems these definitively belong to the docker module :-)

sebastienmusso:
After reflexion it make more cense to split Node and Service in new metricbeat module, for healthcheck part I'll do a new PR.

ruflin:
@sebastienmusso Any suggestion for how we should name the module?

sebastienmusso:
Hi,

Yeah I think the name should simple be "healthcheck" it make sense in a
docker module.
Only container have healthcheck in docker environment.

2017-01-16 14:34 GMT+01:00 Nicolas Ruflin <notifications@github.com>:

ruflin:
@sebastienmusso I meant for the parts except healthcheck :-) You mentioned above `Node and Service in new metricbeat module`. What name should we pick for it?

sebastienmusso:
Yes indeed,

I Think theses modules should be named swarmnodes and swarmservices because
some other docker orchestrator are available.


Le 16 janv. 2017 5:52 PM, "Nicolas Ruflin" <notifications@github.com> a
écrit :

ruflin:
So the module would be `swarm` and the metricsets `node` and `service`, correct?

sebastienmusso:
yes it make sense

2017-01-17 10:46 GMT+01:00 Nicolas Ruflin <notifications@github.com>:

ruflin:
Lets do it then :-) Are you going to open a new PR for it?

sebastienmusso:
Yes I will do a new PR for this but later, I'll add service, nodes and
tasks metricset for swarm module.

regards.

2017-01-18 13:31 GMT+01:00 Nicolas Ruflin <notifications@github.com>:

PR #3356, opened 2017-01-12
Fix mulitple json elements in one log entry

jeffw-wherethebitsroam:
We had a problem where non-JSON log entries were being decoded incorrectly as json, e.g.:

```
11:38:04,323 |-INFO testing
```

This is because the json.Decode would only take the first element (in this case '11') and use it as a json decoded field with no errors. This doesn't seem like the expected behaviour. It will now return an error if there are elements remaining after the first is taken.

### DIFF
--- a/libbeat/processors/actions/decode_json_fields.go
+++ b/libbeat/processors/actions/decode_json_fields.go
@@ -165,6 +165,10 @@ func DecodeJSON(text []byte, to *interface{}) error {
 		return err
 	}
 
+	if dec.More() {
+		return errors.New("Multiple json elements found")
+	}
+
 	switch O := interface{}(*to).(type) {
 	case map[string]interface{}:
 		jsontransform.TransformNumbers(O)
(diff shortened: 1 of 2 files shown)


libbeat/processors/actions/decode_json_fields.go
```diff
@@ -165,6 +165,10 @@ func DecodeJSON(text []byte, to *interface{}) error {
 		return err
 	}
 
+	if dec.More() {
+		return errors.New("Multiple json elements found")
```
andrewkroh: error strings should not be capitalized. https://github.com/golang/go/wiki/CodeReviewComments#error-strings

monicasarbu:
LGTM. Waiting for green.

jeffw-wherethebitsroam:
Hi guys, I see this is in master now, but any idea when it is likely to make it into a release? We are waiting for this feature in production and are wondering if we should wait for a release or build our own version.

Use case: reading docker logs. We use the json options in the prospector to extract the log value from the Docker json logs. However, some applications use structured logging (in JSON) and some don't. So we want to use decode_json_fields for those who use structured logging and have it fail and preserve the original log line for those who don't.

The problem this pull request fixed was that any log line with a leading number will have that number decoded as json and the rest of the log line will be lost, making it unusable.

ruflin:
@jeffreyeliotabbott This fix is part of the upcoming 5.3 release.

PR #3359, opened 2017-01-12
Introduce MetaData and BulkMetadata PublishEvent options

urso:
Introduce MetaData and BulkMetadata options to PublishEvent.
The Metadata will be combines with the events to be published in order
to pass additional meta-data to the outputs.

This comment adds support for the "pipeline" meta-data element to the
elasticsearch output.

### DIFF
--- a/libbeat/publisher/client.go
+++ b/libbeat/publisher/client.go
@@ -103,27 +103,57 @@ func (c *client) PublishEvent(event common.MapStr, opts ...ClientOption) bool {
 		return false
 	}
 
-	ctx, pipeline := c.getPipeline(opts)
+	var values *outputs.Values
+	meta, ctx, pipeline := c.getPipeline(opts)
+	if len(meta) != 0 {
+		if len(meta) != 1 {
+			logp.Debug("publish", "too many metadata, pick first")
+			meta = meta[:1]
+		}
+		values = outputs.ValuesWithMetadata(nil, meta[0])
+	}
+
 	publishedEvents.Add(1)
 	return pipeline.publish(message{
 		client:  c,
 		context: ctx,
-		datum:   outputs.Data{Event: *publishEvent},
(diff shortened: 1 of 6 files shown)


libbeat/publisher/client.go
```diff
@@ -103,27 +103,57 @@ func (c *client) PublishEvent(event common.MapStr, opts ...ClientOption) bool {
 		return false
 	}
 
-	ctx, pipeline := c.getPipeline(opts)
+	var values *outputs.Values
+	meta, ctx, pipeline := c.getPipeline(opts)
+	if len(meta) != 0 {
+		if len(meta) != 1 {
+			logp.Debug("publish", "too many metadata, pick first")
```
ruflin: Should we log that as an error? This should not happen, right?

libbeat/publisher/client.go
```diff
@@ -103,27 +103,57 @@ func (c *client) PublishEvent(event common.MapStr, opts ...ClientOption) bool {
...
 	})
 }
 
 func (c *client) PublishEvents(events []common.MapStr, opts ...ClientOption) bool {
+	var valuesAll *outputs.Values
+
+	meta, ctx, pipeline := c.getPipeline(opts)
+	if len(meta) != 0 && len(events) != len(meta) {
+		if len(meta) != 1 {
+			logp.Debug("publish",
```
ruflin: Should this also be an error?

ruflin:
This currently breaks the unit tests:
```
publisher/client_test.go:39: assignment count mismatch: 2 = 3
```

libbeat/publisher/client.go
urso: Well it can happen. e.g. if 2 prospectors are configured and only one has `fields.pipeline` and the other not. So it might fail on purpose.

PR #3360, opened 2017-01-12
Support type "nested" with empty property list

### LINKED ISSUE #3307: beats/libbeat/scripts/generate_template.py to support empty "nested" or "object" type definitions
Currently it is not possible to configure a type "nested" or "object" for a field but don't have any child fields defined. Currently this leads to an error.
This would be useful when the the child fields are not known in advance and/or are that many that you don't want to explicitly map them but would like to fall back to dynamic templates for the unmapped fields.

breml:
Fixes #3307

### DIFF
--- a/libbeat/scripts/generate_template.py
+++ b/libbeat/scripts/generate_template.py
@@ -162,10 +162,11 @@ def fill_section_properties(args, section, defaults, path):
     properties = {}
     dynamic_templates = []
 
-    for field in section["fields"]:
-        prop, dynamic = fill_field_properties(args, field, defaults, path)
-        properties.update(prop)
-        dynamic_templates.extend(dynamic)
+    if "fields" in section:
+        for field in section["fields"]:
+            prop, dynamic = fill_field_properties(args, field, defaults, path)
+            properties.update(prop)
+            dynamic_templates.extend(dynamic)
 
     return properties, dynamic_templates
 
@@ -292,12 +293,12 @@ def fill_field_properties(args, field, defaults, path):
             path = field["name"]
         prop, dynamic = fill_section_properties(args, field, defaults, path)
 
+        properties[field.get("name")] = {
+            "type": "nested",
+            "properties": {}
+        }
         # Only add properties if they have a content
         if len(prop) is not 0:
-            properties[field.get("name")] = {
-                "type": "nested",
-                "properties": {}
-            }
             properties[field.get("name")]["properties"] = prop
 
         dynamic_templates.extend(dynamic)


christiangalsterer:
Thanks very much for providing a fix for #3307. Do you see a change to backboard this also to 5.1.x?

ruflin:
jenkins, test it

libbeat/scripts/generate_template.py
```diff
@@ -292,12 +293,12 @@ def fill_field_properties(args, field, defaults, path):
             path = field["name"]
         prop, dynamic = fill_section_properties(args, field, defaults, path)
 
+        properties[field.get("name")] = {
+            "type": "nested",
```
monicasarbu: I think it's not needed `type: nested`, and only `properties: {}`

ruflin:
@christiangalsterer Would be nice if you could test this change with your beat. I would suggest not to backport this as we try to backport only bugfixes to already released minor branches.

christiangalsterer:
Applied the change to 5.1.2 locally and get the following error:
`
  File "./vendor/github.com/elastic/beats/libbeat/scripts/generate_template.py", line 332, in <module>
    fields_to_es_template(args, fields, output, args.beatname + "-*", version_data['version'])
  File "./vendor/github.com/elastic/beats/libbeat/scripts/generate_template.py", line 98, in fields_to_es_template
    defaults, "")
  File "./vendor/github.com/elastic/beats/libbeat/scripts/generate_template.py", line 162, in fill_section_properties
    prop, dynamic = fill_field_properties(args, field, defaults, path)
  File "./vendor/github.com/elastic/beats/libbeat/scripts/generate_template.py", line 264, in fill_field_properties
    prop, dynamic = fill_section_properties(args, field, defaults, path)
  File "./vendor/github.com/elastic/beats/libbeat/scripts/generate_template.py", line 162, in fill_section_properties
    prop, dynamic = fill_field_properties(args, field, defaults, path)
  File "./vendor/github.com/elastic/beats/libbeat/scripts/generate_template.py", line 280, in fill_field_properties
    properties: {},
TypeError: unhashable type: 'dict'
make: *** [update] Error 1
`

christiangalsterer:
Please ignore my previous comment, this was my bad, it works like a charm.

libbeat/scripts/generate_template.py
christiangalsterer: Strictly speaking it is not needed, but it was there before, so not preferences from my side to keep or remove.

breml: I am not sure about this one. My understanding of the elasticsearch documentation is, without the definition of the `type` it would become type [object](https://www.elastic.co/guide/en/elasticsearch/reference/5.1/object.html).
Because the user explicitly defined the field type to be `nested`, setting the type implicitly to object would be wrong and therefore we are required to explicitly define the type as [nested](https://www.elastic.co/guide/en/elasticsearch/reference/5.1/nested.html).
Either way I think we are on the safe side, if we keep the `type: nested`, because it was there and therefore we do not break backwards compatibility.

christiangalsterer: Actually you are right. The default is 'object' so that 'nested' is required here.

PR #3361, opened 2017-01-12
Fix for older style params to always be strings

dforste:
Update to the index template so that params that are older are forced to be strings no matter what es thinks they are. Fixes an issue we saw with event_data.param2 was created as type date and should be string to support other values like "stopped".

My issue is similar to what was discussed on the forum here: https://discuss.elastic.co/t/winlogbeat-5-0-mapper-parsing-exception-from-elasticsearch/57223

### DIFF
--- a/winlogbeat/winlogbeat.template.json
+++ b/winlogbeat/winlogbeat.template.json
@@ -16,6 +16,18 @@
             },
             "match_mapping_type": "string"
           }
+        },
+        {
+          "params_as_strings": {
+            "match_pattern": "regex",
+            "mapping": {
+              "ignore_above": 1024,
+              "index": "not_analyzed",
+              "type": "string"
+            },
+            "match_mapping_type": "*",
+            "match": "^param\\d+$"
+          }
         }
       ],
       "properties": {


andrewkroh:
Did you have the default provided index template installed prior to indexing data from Winlogbeat?

The reason I ask is because there is already a [component](https://github.com/elastic/beats/blob/f9279de951d7d81f790d43a1be218d1416a3ac7a/winlogbeat/winlogbeat.template.json#L12-L17) in the index template that sets any dynamic fields that are strings to type `keyword` in ES 5.x and non-analyzed string in ES 2.x. Therefore it should not be possible for any `event_data.*` field to be a `date` type in ES.

dforste:
@andrewkroh Yes I had the index template in there as well. I also had some other templates in there that got merged in at index creation time. Here is the content of the dynamic template that had the issue indexing. 
```
    "dynamic_templates": [
      {
        "strings_as_keyword": {
          "mapping": {
            "ignore_above": 1024,
            "index": "not_analyzed",
            "type": "string"
          },
          "match_mapping_type": "string"
        }
      },
      {
        "string_fields": {
          "mapping": {
            "index": "analyzed",
            "omit_norms": true,
            "type": "string",
            "fields": {
              "raw": {
                "ignore_above": 256,
                "index": "not_analyzed",
                "type": "string",
                "doc_values": true
              }
            }
          },
          "match_mapping_type": "string",
          "match": "*"
        }
      }
    ],
```

Here is after I added the template above and haven't had any issues indexing content. 
```
    "dynamic_templates": [
      {
        "strings_as_keyword": {
          "mapping": {
            "ignore_above": 1024,
            "index": "not_analyzed",
            "type": "string"
          },
          "match_mapping_type": "string"
        }
      },
      {
        "params_as_strings": {
          "match_pattern": "regex",
          "mapping": {
            "ignore_above": 1024,
            "index": "not_analyzed",
            "type": "string"
          },
          "match_mapping_type": "*",
          "match": "^param\d+$"
        }
      },
      {
        "string_fields": {
          "mapping": {
            "index": "analyzed",
            "omit_norms": true,
            "type": "string",
            "fields": {
              "raw": {
                "ignore_above": 256,
                "index": "not_analyzed",
                "type": "string",
                "doc_values": true
              }
            }
          },
          "match_mapping_type": "string",
          "match": "*"
        }
      }
    ],
```

dforste:
@andrewkroh I have been able to re-index the data from the broken index mapping and it has been indexed as string with the adjusted template.

andrewkroh:
This doesn't seem like a general problem with Beats (correct me if I'm wrong), but more of a problem with conflicting settings inside of your index templates. The `strings_as_keyword` provided by Beats and `string_fields` coming from one of your templates are trying to do different things with the same fields.

If you want multi-fields to be created for the dynamic string fields then edit our template to remove `strings_as_keyword`. Or make some changes like you have done or remove `string_fields` if you just want non-analyzed fields.

dforste:
Even with the `strings_as_keyword` template when the value param2: "2015/06/17" came in it was indexed as a date. 

I still think my template is needed to prevent that from happening. 

Can you explain why the default template from logstash breaks this? There were never any .raw fields created just indexed as type date instead of string. I would understand this as evidence that the `string_fields` template was not active and the `strings_as_keyword` failed to apply as elasticsearch determined that  "2015/06/17" was a date not a string.

andrewkroh:
> Even with the strings_as_keyword template when the value param2: "2015/06/17" came in it was indexed as a date.

My mistake I thought that this disabled date detection, but that is a separate parameter in the index template. So if you are writing the Winlogbeat data to `winlogbeat-*` then you should be able to install the following template (in addition to the provided template) to disable date detection.

```
PUT _template/winlogbeat1
{
  "mappings": {
    "_default_": {
      "date_detection": false
    }
  },
  "order": 1,
  "template": "winlogbeat-*"
}
```

I will open an issue to see about getting that parameter into the beat templates. Thanks for reporting the issue.

dforste:
Will that break the @timestamp fields I just want to ensure the fields get indexed as they should be. 

I also notice some date fields under event_data. These are not named the arbitrary fieldx. I just think those legacy fields for windows event logging should be kept as strings, and ES allowed to do what it does best for the others. I am Just trying to limit mapping changes to those fields to prevent other problems that might arise.

PR #3362, opened 2017-01-13
Implement prospector reloading

ruflin:
This PR allows to dynamically reload prospectors. It works the same way as module reloading in metricbeat.

**Refactoring**

* LoadStates was separated from NewProspector. The reason is that after New only the ID is needed and setting up states requires more calculations. So this can be done in a second step when all the validations are done.
* Only allow to start a prospector when all states are set to Finished. If not, LoadStates returns an error. This is to prevent a prospector starting before a harvester finished with a file. The prospector will be picked up again during the next reloading phase.
* Extract ReloadConfig to libbeat

**Limitations**

This implementation currently has the some limitations. This are not new in filebeat but require more care as configurations change more often.

* Two prospectors on one file: It is possible, that two prospectors pick up one file because they defined overlapping patterns. This can have the consequence that two harvesters on the same file are running which can lead to duplicates and unpredictable behaviour. The risk is minimized in that a prospector does not start as long as a state it takes care of is not finished. But it can still happen that a Finished state is picked up but it also managed by an other prospector. The user must ensure no prospector paths overlap. This problem can potentially be solved in the future with a global harvester registry.

**Notes**

* In a later PR, more refactoring and unification of the reloading should happen.


filebeat/docs/reference/configuration/reload-configuration.asciidoc
```diff
@@ -0,0 +1,41 @@
...
+
+Reload configuration allows to dynamically reload prospector configuration files. A glob can be defined which should be watched
+ for prospector configuration changes. New prospectors will started / stopped accordingly. This is especially useful in
+ container environments where 1 container is used to tail logs from services in other containers on the same host.
+
+The configuration in the main filebeat.yml config file looks as following:
+
+[source,yaml]
+------------------------------------------------------------------------------
+filebeat.reload.modules:
```
tsg: Hmm, this might be confusing. It's called `reload.modules` but reloads prospectors config, not module configs.

filebeat/prospector/reloader.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+func (r *ProspectorReloader) stopProspectors(prospectors map[uint64]*Prospector) {
+	wg := sync.WaitGroup{}
+	for _, p := range prospectors {
+		wg.Add(1)
+		go func(pr *Prospector) {
+			defer wg.Done()
+			logp.Debug("reload", "stopping prospector: %v", pr.ID)
+			// TODO: Stop prospectors in parallel
```
tsg: Isn't this already the case since a goroutine is started for each prospector?

filebeat/prospector/reloader.go
```diff
@@ -0,0 +1,167 @@
...
+	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/logp"
+)
+
+var (
+	debugr            = logp.MakeDebug("filebeat.reloader")
+	configReloads     = expvar.NewInt("filebeat.config.reloads")
+	prospectorStarts  = expvar.NewInt("filebeat.config.prospector.starts")
+	prospectorStops   = expvar.NewInt("filebeat.config.prospector.stops")
+	prospectorRunning = expvar.NewInt("filebeat.config.prospector.running")
```
tsg: This seems to only count dynamically configured prospectors, right? We should reflect that in the name.

filebeat/prospector/reloader.go
```diff
@@ -0,0 +1,167 @@
...
+
+	for {
+		select {
+		case <-r.done:
+			logp.Info("Dynamic config reloader stopped")
+			return
+		case <-time.After(r.config.Period):
+
+			debugr("Scan for new config files")
+			configReloads.Add(1)
```
tsg: If this is incremented at each period regardless of whether any file has changed, the expvar name might be a little confusing.

tsg: It would be easy to just move it after `!updated` and then it's more useful, right?

filebeat/prospector/reloader.go
```diff
@@ -0,0 +1,167 @@
...
+
+	r.wg.Add(1)
+	defer r.wg.Done()
+
+	// Stop all running modules when method finishes
+	defer r.stopProspectors(r.registry.CopyList())
+
+	path := r.config.Path
+	if !filepath.IsAbs(path) {
+		path = filepath.Join(cfgfile.GetPathConfig(), path)
```
tsg: Better use `paths.Resolve(paths.Config, path)`.

tsg: Ah, I see that `GetPathConfig`, which was recently introduced, duplicates the logic of `paths.Resolve`. Was there a reason for that? I'd prefer using `paths.Resolve` everywhere.

filebeat/docs/reference/configuration/reload-configuration.asciidoc
ruflin: It is actually `reload.prospectors` not modules. Copy / paste error from metricbeat ...

filebeat/prospector/reloader.go
ruflin: yes, the comment was a left over.

filebeat/prospector/reloader.go
ruflin: I changed it to `filebeat.config.prospector.dyamic.*`

filebeat/prospector/reloader.go
ruflin: I moved it after !updated

filebeat/prospector/reloader.go
ruflin: `GetPathConfig` was around for quite some time I think, or at least the logic around it. I was not aware of `paths.Resolve` that is why I used `GetPathConfig`. I changed it now in the reloaders.

ruflin:
@tsg New version pushed with reviews implemented.

filebeat/beater/filebeat.go
```diff
@@ -55,10 +55,14 @@ func New(b *beat.Beat, rawConfig *common.Config) (beat.Beater, error) {
 	// Add prospectors created by the modules
 	config.Prospectors = append(config.Prospectors, moduleProspectors...)
 
-	if len(config.Prospectors) == 0 {
+	if !config.ProspectorReload.Enabled() && len(config.Prospectors) == 0 {
 		return nil, errors.New("No prospectors defined. What files do you want me to watch?")
 	}
 
+	if *once && config.ProspectorReload.Enabled() {
+		return nil, errors.New("Once and prospector reloading cannot be used together.")
```
andrewkroh: Consider changing `s/Once/-once/` and/or swapping it around like "prospector reloading and -once cannot be used together".

filebeat/crawler/crawler.go
```diff
@@ -45,6 +41,15 @@ func (c *Crawler) Start(r *registrar.Registrar) error {
 		}
 	}
 
+	if reloaderConfig.Enabled() {
+		logp.Warn("EXPERIMENTAL feature dynamic configuration reloading is enabled.")
+
+		go func() {
```
andrewkroh: How does this goroutine stop?

filebeat/docs/reference/configuration/reload-configuration.asciidoc
```diff
@@ -0,0 +1,41 @@
+[[filebeat-configuration-reloading]]
+=== Reload Configuration
+
+experimental[]
+
+Reload configuration allows to dynamically reload prospector configuration files. A glob can be defined which should be watched
+ for prospector configuration changes. New prospectors will started / stopped accordingly. This is especially useful in
```
andrewkroh: "will _be_ started"

filebeat/docs/reference/configuration/reload-configuration.asciidoc
```diff
@@ -0,0 +1,41 @@
...
+------------------------------------------------------------------------------
+filebeat.reload.prospectors:
+  enabled: true
+  path: configs/*.yml
+  period: 10s
+------------------------------------------------------------------------------
+
+A path with a glob must be defined on which files should be checked for changes. A period is set on how often
+the files are checked for changes. Do not set period below 1s as the modification time of files is often stored in seconds.
+Setting it below 1s will have an unnecessary overhead.
```
andrewkroh: `s/will have/will cause/`

filebeat/beater/filebeat.go
ruflin: I swapped the two.

filebeat/crawler/crawler.go
ruflin: Good you spotted that. It's missing :-( Will add it.

filebeat/docs/reference/configuration/reload-configuration.asciidoc
ruflin: changed

filebeat/docs/reference/configuration/reload-configuration.asciidoc
ruflin: changed

ruflin:
@andrewkroh I pushed a new commit with the changes. About deprecating `config_dir`. I think we should definitively deprecate it in filebeat and reintroduce it in all beats, especially as `config_dir` requires different config files. The part I'm not sure about is if we should mix reload and config dir. From a technical point of view it is almost the same, but configuration wise I'm not sure if we should mix the two. Someone using `config_dir` should not require to enable reloading. 

I will open a seperate PR to deprecate `config_dir` with a note that it will be replace with something "better" :-)

filebeat/crawler/crawler.go
```diff
@@ -45,6 +42,15 @@ func (c *Crawler) Start(r *registrar.Registrar) error {
...
+			c.reloader = prospector.NewProspectorReloader(reloaderConfig, c.out, r)
```
andrewkroh: I don't think anything bad will happen here, but doing the `c.reloader` assignment outside of the goroutine would be less prone to race conditions.

urso: agreed, then code can become:

```
c.reloader = ...
go c.reloader.Run()
```

filebeat/crawler/crawler.go
```diff
@@ -91,7 +102,14 @@ func (c *Crawler) Stop() {
 		c.wg.Add(1)
 		go stopProspector(p)
 	}
+
+	if c.reloader != nil {
+		c.wg.Add(1)
+		go c.reloader.Stop()
```
urso: hmmm... where do you call `c.wg.Done`

```
asyncWaitStop := func(stop func()) {
    c.wg.Add(1)
    go func() {
      defer c.wg.Done()
      stop()
  }
}

...
for _, p := range c.prospectors {
    asyncWaitStop(p.Stop)
}
if c.reloader != nil {
    asyncWaitStop(c.reloader.Stop)
}
```

ruflin: c.wg.Done is called on line 50. Current implementation works as expected.

urso: oh... I'm not really ok with calling it on line 50. When using waitGroups one must/should ensure `Add` is always run before `Done`, so the internal semaphore(counter) can not become negative at any time. What if `Run` returns early (even if it doesn't). With `Done` being called on line 50, let's move the add right before that line.

ruflin: Agree. Changed.

superwhykz:
Thanks so much guys for pushing this PR!  Quick question we heavily use "config_dir" option. Does it mean if we have that option enabled reload will not work?

ruflin:
@superwhykz No, currently config_dir and the reloading can be enabled together, you should just ensure that the directories they are fetching config files from are not overlapping. Let us know if you hit any issues with the reloading.

superwhykz:
@ruflin awesome! this feature should be in night build by now right?

ruflin:
@superwhykz yes

jasonkylemorse:
In what version are you guys targeting this feature to be available? Any idea on a timeframe for that? Thanks!

ruflin:
@jasonkylemorse It is available in 5.3 as an experimental feature. 5.3 should be coming out in the next weeks. If you want to test it already, you can also use the nightly builds: https://beats-nightlies.s3.amazonaws.com/index.html?prefix=filebeat/

PR #3367, opened 2017-01-13
Add heartbeat 'beta' log message

urso:


heartbeat/beater/heartbeat.go
```diff
@@ -23,6 +23,8 @@ type Heartbeat struct {
 }
 
 func New(b *beat.Beat, cfg *common.Config) (beat.Beater, error) {
+	logp.Warn("Beta: Heartbeat is beta software")
```
ruflin: We should also add this to the docs (if not already the case)

urso: it's already tagged as beta in docs

PR #3379, opened 2017-01-16
Add NOTICE file to the packages

tsg:
Added to the home path. Also modifies the generators to include a simple
NOTICE file, which is anyway good practice for Apache licensed projects.

Tested the beat and the metricbeat generators by hand.

Follow up for #3334.

### DIFF
--- a/generate/beat/{beat}/NOTICE
+++ b/generate/beat/{beat}/NOTICE
@@ -1,0 +1,5 @@
+{beat}
+Copyright 2017 {full_name}
+
+This product includes software developed by The Apache Software
+Foundation (http://www.apache.org/).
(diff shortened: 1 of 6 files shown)


tsg:
jenkins, package it

generate/beat/{beat}/NOTICE
```diff
@@ -0,0 +1,5 @@
+{beat}
+Copyright 2017 {full_name}
```
ruflin: We should introduce {year} here so we have it always up-to-date ;-) Not needed in this PR.

ruflin:
Packaging looks good: http://build-eu-00.elastic.co/job/beats-package-PR/63/

generate/beat/{beat}/NOTICE
monicasarbu: The year is generated by the script, so you would need to re-run the script every year, at least :)

ruflin: That is even nicer. My change would only set the correct date on the first creation.

PR #3382, opened 2017-01-16
Set config file mode to 0600 in packages

andrewkroh:
This PR changes the main config file (<beatname>.yml) from 0644 to 0600. It adds a test that checks all RPM, deb, tar.gz, and zip files to ensure that the config file has the correct file mode.


andrewkroh:
jenkins, package it

dev-tools/package_test.go
```diff
@@ -0,0 +1,289 @@
...
+	"compress/gzip"
+	"flag"
+	"io"
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
+	"testing"
+
+	"github.com/blakesmith/ar"
```
ruflin: Could we add a vendor directory to the dev-tools directory? I would prefer if we don't mix vendor packages for tools with actual libbeat / beats dependencies. This makes it easier to track where a dependency belongs to. 

Long term I'm thinking also to move dependencies which only belong to one beat into the beat if possible.

glide.yaml
```diff
@@ -105,7 +105,7 @@ import:
   - zlib
 - package: github.com/klauspost/cpuid
   version: v1.0
-- package: https://github.com/go-inf/inf
+- package: github.com/go-inf/inf
```
ruflin: see comment above

dev-tools/package_test.go
andrewkroh: Yeah, I can move these to dev-tools. I was thinking about that when I added them because `glide` was giving lots of errors and also because I wanted it to be clear that these deps were not used in any Beat that we release.

andrewkroh: I update the PR to move the dependencies to `./dev-tools/vendor`.

andrewkroh:
jenkins, package it

dev-tools/glide.yaml
```diff
@@ -0,0 +1,18 @@
+package: github.com/elastic/beats/dev-tools
+import: []
+testImports:
```
ruflin: wasn't aware of this glide feature.

ruflin:
@monicasarbu @andrewkroh One thing I just realised now is that the next time we build the NOTICE file, it will also include the LICENSES from these dependencies. But I think we should leave this out, means excluding `dev-tools`?

PR #3386, opened 2017-01-16
Cleanup before_build script

ruflin:
* Pass ES_BEATS variable to before_build.sh scripts. This removes the logic to figure out which is the correct path. ES_BEAT path should be passed relative to the BEAT_PATH.
* Move gotpl and go-yaml installation to the Dockerfile itself. This removes the requirement to install go dependencies in the scripts and moves the special handling for debian6 to the debian6 image.
* Improve naming of variables
* Pass BEAT_NAME, BEAT_PATH, ES_BEATS as environment variables. This makes it possible to use the same naming/logic in all scripts.

The above allows more flexibility on how community beats can use package. As before-build is now executed as the first step, a beat can use this step to fetch beat dependencies with `go get github.com/elastic/beats`. This would fetch the master branch for building every time and does not require to vendor it. A disadvantage of the above is that `before-build` is called twice, once for each debian version.

### DIFF
--- a/dev-tools/packer/docker/xgo-image-deb6/beats-builder/Dockerfile
+++ b/dev-tools/packer/docker/xgo-image-deb6/beats-builder/Dockerfile
@@ -14,5 +14,12 @@ RUN \
 	apt-get -o Acquire::Check-Valid-Until=false update && \
 	apt-get install -y libpcap0.8-dev
 
+
+# Old git version which does not support proxy with go get requires to fetch go-yaml directly
+RUN git clone https://github.com/go-yaml/yaml.git /go/src/gopkg.in/yaml.v2
+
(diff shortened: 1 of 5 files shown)


ruflin:
jenkins, package it

tsg:
It used to be that gotpl was needed outside of the docker images. From a brief inspection of the Makefiles it seems that that's still the case.

There was a failure on packer jenkins, but it doesn't seem related to my point above.

ruflin:
jenkins, package it

ruflin:
jenkins, package it

ruflin:
@tsg You are right, it is still required. I readded that part.

ruflin:
Packaging build: http://build-eu-00.elastic.co/job/beats-package-PR/68/

PR #3387, opened 2017-01-16
Use 0600 for files created by Beats

andrewkroh:
This change applies to registry, log, and output files. Previously files were created with 0644.


filebeat/registrar/registrar.go
```diff
@@ -298,7 +298,7 @@ func (r *Registrar) writeRegistry() error {
 	logp.Debug("registrar", "Write registry file: %s", r.registryFile)
 
 	tempfile := r.registryFile + ".new"
-	f, err := os.Create(tempfile)
+	f, err := os.OpenFile(tempfile, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0600)
```
ruflin: We could put these flags into a constant as we reuse them in several places or have a function in libbeat to open files which automatically applies flags and correct access rights.

PR #3388, opened 2017-01-16
beat contrib: added apexbeat link

jenssaade:
Added link for community beat contrib. 
Additional docs can be found here: http://toolkits.verticle.io/docs/index.html#elasticsearch-apexbeats

### DIFF
--- a/libbeat/docs/communitybeats.asciidoc
+++ b/libbeat/docs/communitybeats.asciidoc
@@ -9,6 +9,7 @@ NOTE: Elastic provides no warranty or support for community-sourced Beats.
 [horizontal]
 https://github.com/awormuth/amazonbeat[amazonbeat]:: Reads data from a specified Amazon product.
 https://github.com/radoondas/apachebeat[apachebeat]:: Reads status from Apache HTTPD server-status.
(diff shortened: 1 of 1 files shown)


ruflin:
Thanks for creating apexbeats. One thing I'm curious about: Is the data structure of apex metrics known in advance or different in each case?

jenssaade:
The apex message structure is more or less generic and consisting of key-value pairs. The values types are currently not restricted but in general serializable objects are supported. There is some special handling for Java List and Map types though. The apex collector will transform those to stringified JSON objects so they don't clash with existing ES types.

ruflin:
Thanks for the details. I was thinking if this could make sense as a metricbeat module / metricset: https://www.elastic.co/guide/en/beats/metricbeat/current/creating-metricsets.html

jenssaade:
Depends on the data flow from the collector to the beat. Since metricbeat fetches the data there would be some changes to the current flow. Let's discuss on gitter.

ruflin:
Gitter? We don't have a channel there, but we have IRC: https://kiwiirc.com/client/irc.freenode.net/beats

PR #3393, opened 2017-01-17
Add fetching of testify to prepare-tests

ruflin:
This is needed for community beats which use testify and dont have it fetched in advance.


libbeat/scripts/Makefile
```diff
@@ -116,6 +117,7 @@ prepare-tests:
 	mkdir -p ${COVERAGE_DIR}
 	# gotestcover is needed to fetch coverage for multiple packages
 	go get ${COVERAGE_TOOL_REPO}
+	go get ${TESTIFY_TOOL_REPO}
```
tsg: i suggest a comment above it like we have for gotestcover

ruflin: done

PR #3394, opened 2017-01-17
Load Filebeat modules pipelines on -setup

tsg:
This adds the `-setup` CLI flag, which, for now, makes Filebeat load the pipelines at startup. In case Elasticsearch is not available when Filebeat is started with `-setup`, Filebeat will exit with an error.

This is the first version that passes the modules tests without the help of `filebeat.py` 🎉 .

This also exposes an Elasticsearch client from the output.

A few more things to note/discuss:

* For now I added `-setup` to Filebeat only. As we add support to loading the dashboards as well, we'll probably want to move the flag in libbeat and make it available for all Beats.
* The current behavior is that `-setup` exits on errors but continues with a normal Filebeat run in case of success. The idea is that you would use `-setup` to make the initial setup but also to test it. After that is successful, you would typically stop Filebeat and start it in the background via the init scripts (so without `-setup`).
* The above means that if Elastisearch is not available when calling `filebeat -setup`, the user gets an error directly, it doesn't "wait" for ES.
* I tried separating the client but resulted in a too heavy change so I bailed out. I added a function to get a client with the same configuration. This works fine but it's a bit hacky.
* the elasticsearch output `Connection.request` is made public as `Connection.Request`. This goes in the direction of separating the client. I only needed it in tests, though, so I can take a different approach if needed.
* If more ES hosts are defined, `-setup` will iterate through them and use the first one that responds to an HTTP ping.

Part of #3159.


tsg:
See also the discussion here: https://github.com/elastic/beats/issues/3159#issuecomment-273164870

filebeat/beater/filebeat.go
```diff
@@ -19,6 +20,7 @@ import (
 )
 
 var once = flag.Bool("once", false, "Run filebeat only once until all harvesters reach EOF")
```
ruflin: group together under `var(` ?

filebeat/beater/filebeat.go
```diff
@@ -63,11 +67,33 @@ func New(b *beat.Beat, rawConfig *common.Config) (beat.Beater, error) {
 	return fb, nil
 }
 
+// Setup is called on user request (the -setup flag) to do the initial Beat setup.
+func (fb *Filebeat) Setup(b *beat.Beat) error {
```
ruflin: Didn't we have a Setup step in the 1.0 beats interface? Old days :-)

filebeat/tests/system/test_modules.py
```diff
@@ -47,27 +57,27 @@ def test_modules(self):
...
+            "-c", cfgfile,
+            "-modules={}".format(module),
             "-M", "{module}.{fileset}.var.paths=[{test_file}]".format(
                 module=module, fileset=fileset, test_file=test_file),
-            "--es", self.elasticsearch_url,
-            "--index", index_name,
-            "--registry", self.working_dir + "/registry"
+            "-M", "*.*.prospector.close_eof=true",
         ]
+        # print(" ".join(cmd) + "\n")
```
ruflin: Should be removed

filebeat/beater/filebeat.go
tsg: Yeah :). I didn't add it yet into the interface but I have a feeling that's where it will eventually get there.

filebeat/tests/system/test_modules.py
tsg: I replaced the print with outputting to the log. Having the command already built in the logs is fantastic for troubleshooting :)

monicasarbu:
The main reason to add the `-setup` option was to be able to reduce the number of configuration steps to one, in order to avoid running the `import_dashboards` in addition. For now, I would say it's ok if we need to run a Beat with the `-setup` option to import the dashboards, pipelines before starting to harvest log files, but later it would be nice if any of the Beat could do the "setup" if the dashboards & pipelines are not imported. +1 of `-setup` option to do the setup and run Filebeat.

filebeat/fileset/modules.go
```diff
@@ -240,3 +241,29 @@ func (reg *ModuleRegistry) GetProspectorConfigs() ([]*common.Config, error) {
 	}
 	return result, nil
 }
+
+func (reg *ModuleRegistry) Setup(esClient *elasticsearch.Client) error {
```
urso: do we need the full `elasticsearch.Client` here? If we define an interface with methods required, it get's us an idea, well, what's really required for ModuleRegistry + we can easily mock the client in tests.

PR #3398, opened 2017-01-18
Update docker client to most recent version

ruflin:
The docker client now depends on the docker repository which means we have to add the docker repo. This adds quite a few dependencies. To reduce the number of dependencies, the docker repository was modified manually and all not required parts were removed. On the client side some code parts were removed which are not needed for our read only use case.

Long term goal is to either use the internal docker client or use our own implementation of a HTTP client that fetches data from the socket or rest API and uses our internal schema.


andrewkroh:
There looks to be some stuff under a "testfiles" dir that can be removed.

PR #3401, opened 2017-01-18
Update heartbeat sample configuration files

urso:
- add banner to configuration files
- use a small sample config in heartbeat.yml only.

### DIFF
--- a/heartbeat/_meta/beat.yml
+++ b/heartbeat/_meta/beat.yml
@@ -1,191 +1,23 @@
 ################### Heartbeat Configuration Example #########################
 
+# This file is an example configuration file highlighting only some common options.
+# The heartbeat.full.yml file in the same directory contains all the supported options
+# with detailed comments. You can use it for reference.
+#
+# You can find the full configuration reference here:
+# https://www.elastic.co/guide/en/beats/heartbeat/index.html
+
 ############################# Heartbeat ######################################
 
 # Configure monitors
 heartbeat.monitors:
-- type: icmp # monitor type `icmp` (requires root) uses ICMP Echo Request to ping
-             # configured hosts
+- type: http
 
-  # Monitor name used for job name and document type.
-  #name: icmp
-
-  # Enable/Disable monitor
-  #enabled: true
-
-  # Configure task schedule using cron-like syntax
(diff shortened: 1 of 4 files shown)


heartbeat/_meta/beat.yml
```diff
@@ -1,191 +1,20 @@
 ################### Heartbeat Configuration Example #########################
 
+# This file is an example configuration file highlighting only some common options.
+# The heartbeat.full.file file in the same directory contains all the supported options
```
ruflin: heartbeat.full.yml (instead of file)

urso:
> Left a minor comments. Besides that looks good to me. I think what is most important is that if some just runs the binary, it works out of the box and this seems to be the case with icmp only.

hmm... icmp might not work on windows though

heartbeat/_meta/beat.yml
urso: fixed

heartbeat/_meta/beat.yml
```diff
@@ -1,191 +1,20 @@
...
-
-  # Monitor name used for job name and document type.
-  #name: icmp
-
-  # Enable/Disable monitor
-  #enabled: true
-
-  # Configure task schedule using cron-like syntax
-  schedule: '*/5 * * * * * *' # exactly every 5 seconds like 10:00:00, 10:00:05, ...
+- type: icmp
```
tsg: I think you were saying the `icmp` is not available on Windows, right? I'd like the default configuration to work by default on all platforms. Plus I find the case of HTTP monitoring more natural for Heartbeat.

tsg:
jenkins, package it

heartbeat/_meta/beat.yml
ruflin: If we switch to http as default, the discussion comes up which port should be monitored by default. In the default setup we expect 9200 to be available (ES). Current default is 80.

monicasarbu: +1 for 9200 port by default.

urso: ok, will go with http + 9200 for minimal config

PR #3407, opened 2017-01-19
Return empty registry if module path doesn't exist

tsg:
This is a follow up for #3405, to avoid having a nil moduleRegistry. Upgraded the Info to Warn since that indicates a faulty installation.

Also adds a unit test for that case.

### DIFF
--- a/filebeat/fileset/modules.go
+++ b/filebeat/fileset/modules.go
@@ -86,8 +86,8 @@ func NewModuleRegistry(moduleConfigs []*common.Config) (*ModuleRegistry, error) 
 
 	stat, err := os.Stat(modulesPath)
 	if err != nil || !stat.IsDir() {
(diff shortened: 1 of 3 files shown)


filebeat/fileset/modules.go
```diff
@@ -86,8 +86,8 @@ func NewModuleRegistry(moduleConfigs []*common.Config) (*ModuleRegistry, error)
 
 	stat, err := os.Stat(modulesPath)
 	if err != nil || !stat.IsDir() {
-		logp.Info("Not loading modules. Module directory not found: %s")
-		return nil, nil
+		logp.Warn("Not loading modules. Module directory not found: %s", modulesPath)
```
ruflin: Not a big fan of the Warn message. Should we make it Error?

PR #3408, opened 2017-01-19
Make python scripts python 3 compatible

ruflin:
Run `2to3 -w -n ./` to automatically make python scripts compatible with version 3.

### DIFF
--- a/dev-tools/export_dashboards.py
+++ b/dev-tools/export_dashboards.py
@@ -14,13 +14,13 @@ def ExportDashboards(es, regex, kibana_index, output_directory):
     try:
         reg_exp = re.compile(regex, re.IGNORECASE)
     except:
-        print("Wrong regex {}".format(regex))
+        print(("Wrong regex {}".format(regex)))
         return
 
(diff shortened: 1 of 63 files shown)


dev-tools/export_dashboards.py
```diff
@@ -14,13 +14,13 @@ def ExportDashboards(es, regex, kibana_index, output_directory):
     try:
         reg_exp = re.compile(regex, re.IGNORECASE)
     except:
-        print("Wrong regex {}".format(regex))
+        print(("Wrong regex {}".format(regex)))
```
tsg: hmm, this seems unnecessary :(

ruflin: Yeah, the script didn't work as well as I hoped. It seems to just apply the same rules again and again. So it doesn't work if there is already code which is python 3 compatible. Need to check the options of the script in more detail.

ruflin:
Closing PR as this din't work as expected. Looking for less intrusive ways.

PR #3411, opened 2017-01-19
Export libbeat.outputs.acked_events via expvar

monicasarbu:
Part of https://github.com/elastic/beats/issues/3422

With this PR, each Beat exports `libbeat.redis.published_and_acked_events` and `libbeat.redis.published_but_not_acked_events` via expvar for Redis. 

In addition, `libbeat.outputs.acked_events` is exported in order to send one metric for all outputs instead of one metric per output. In the case two outputs are enabled, the number of published events will be double. 
 For example in case Redis and Elasticsearch outputs are enabled, then `libbeat.redis.published_and_acked_events: 48`, `libbeat.elasticsearch.published_and_acked_events: 48` and `libbeat.outputs.acked_events: 96`.

### DIFF
--- a/libbeat/outputs/redis/client.go
+++ b/libbeat/outputs/redis/client.go
@@ -2,6 +2,7 @@ package redis
 
 import (
 	"errors"
+	"expvar"
 	"regexp"
 	"strconv"
 	"time"
@@ -17,6 +18,9 @@ import (
 
 var (
 	versionRegex = regexp.MustCompile(`redis_version:(\d+).(\d+)`)
+
+	ackedEvents    = expvar.NewInt("libbeat.redis.published_and_acked_events")
+	eventsNotAcked = expvar.NewInt("libbeat.redis.published_but_not_acked_events")
 )
 
 type publishFn func(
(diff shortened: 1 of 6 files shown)


libbeat/outputs/redis/client.go
```diff
@@ -17,6 +18,9 @@ import (
 
 var (
 	versionRegex = regexp.MustCompile(`redis_version:(\d+).(\d+)`)
+
+	ackedEvents    = expvar.NewInt("libbeat.redis.published_and_acked_events")
```
ruflin: I suggest to name it 

```
libbeat.redis.published.events.acked
libbeat.redis. published.events.not_acked
```
This corresponds better with our metricbeat conventions.

monicasarbu: I used this name to be similar with `libbeat.elasticsearch.published_and_acked_events`, `libbeat.logstash.published_and_acked_events` and `libbeat.kafka.published_and_acked_events`. I agree that the name is not ideal, so maybe we can change all of them in a separate PR.

ruflin: Yes, I kind of assumed you copied it from there. +1 on changing / reviewing all our naming for the metrics. We started this before we had our naming conventions, so most don't fit too well ...

monicasarbu:
Closing in favor of https://github.com/elastic/beats/pull/3423

PR #3413, opened 2017-01-19
Add HTTP helper for Metricsets

ruflin:
This should simplify the implementation of MetricSets based on HTTP.

### DIFF
--- a/metricbeat/helper/http.go
+++ b/metricbeat/helper/http.go
@@ -1,0 +1,85 @@
+package helper
+
+import (
+	"bufio"
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+
+	"github.com/elastic/beats/metricbeat/mb"
+)
+
+type HTTP struct {
+	base   mb.BaseMetricSet
+	client *http.Client // HTTP client that is reused across requests.
+}
+
+// NewHTTP creates new http helper
+func NewHTTP(base mb.BaseMetricSet) *HTTP {
+	return &HTTP{
+		base:   base,
+		client: &http.Client{Timeout: base.Module().Config().Timeout},
+	}
+}
+
+// FetchResponse fetches a response for the http metricset.
+// It's important that resp.Body has to be closed if this method is used. Before using this method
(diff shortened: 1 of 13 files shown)


metricbeat/mb/metricset/http.go
```diff
@@ -0,0 +1,71 @@
...
+
+	"github.com/elastic/beats/metricbeat/mb"
+)
+
+type Http struct {
+	mb.BaseMetricSet
+	client *http.Client // HTTP client that is reused across requests.
+}
+
+// NewHttpMetricset creates new instance of HttpMetriSset
```
tsg: Type in HttpMetricSet.

tsg: Looks more like the comment is out of date, it should be "creates a new instance of Http". Btw, HTTP is an acronym so usually it shows up as HTTP in Go code, right? That might look weird though.

ruflin: Yeah, in the first version it was call HttpMetricset because it was not in the metricset package. 

You are right based on our standard it should be called `HTTP`, but it will look wiered :-( I would prefer to leave it. @urso Thoughts.

metricbeat/helper/http.go
```diff
@@ -0,0 +1,85 @@
...
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+
+	"github.com/elastic/beats/metricbeat/mb"
+)
+
+type Http struct {
```
ruflin: :-) @tsg and me had the same discussion, but came to the conclusion HTTP will look ugly :-) What would you call the private variable in this case? http?

andrewkroh: What private variable?

ruflin: Example here: https://github.com/elastic/beats/pull/3413/files#diff-78222465fa3f09f1d018ca076ab47bbbR46

andrewkroh: I don't see anything wrong with `http *helper.HTTP`. It's not much different from `http *helper.Http`. Or am I missing something?

PR #3415, opened 2017-01-19
New PHP-FPM metricbeat module

tsouza:
This adds [PHP-FPM](https://php-fpm.org) support for metricbeat (https://github.com/elastic/beats/issues/2247)

It is my first attempt writing a metricbeat module. One minor configuration that is not clear to me is how to externalize configuration of `DefaultPath` of the `HostParser`. Currently it is set to `/status` (which is php-fpm's default) but I just don't know how to make it user-configurable. 

Also, I believe metricbeat creates sample dashboards, I would like to build a php-fpm sample dashboard and add it. Where to start?


ruflin:
@tsouza I recommend you to run `make fmt` and `make update` before the next commit. This will clean up the source code and update all generated files. This should get you closer to a green build :-)

ruflin:
@tsouza Travis failures are unrelated to your PR. They should be fixed as soon as https://github.com/elastic/beats/pull/3425 is merged.

metricbeat/docs/modules/php_fpm.asciidoc
```diff
@@ -0,0 +1,65 @@
...
+in <<configuration-metricbeat>>. Here is an example configuration:
+
+[source,yaml]
+----
+metricbeat.modules:
+- module: php_fpm
+  metricsets: ["pool"]
+  enabled: true
+  period: 10s
+  hosts: ["localhost"]
```
ruflin: What is the default port the metrics are exposed? 80 or is it 9000 ?

metricbeat/module/php_fpm/_meta/config.yml
```diff
@@ -0,0 +1,6 @@
+- module: php_fpm
```
ruflin: We currently have all module configs commented out which are not enabled by default. This might change in the near future, but for consistency it should be commented out here

metricbeat/module/php_fpm/_meta/fields.yml
```diff
@@ -0,0 +1,11 @@
+- key: php_fpm
+  title: "php_fpm"
+  description: >
+    PHP-FPM server status metrics collected from PHP-FPM.
+  fields:
```
ruflin: Can you add here the `experimental[]` flag? We do that for all new modules so we can still break compatibility after the first release. This might be a beta flag soonish.

Also please add `short_config: false` so it does not show up in the short config. See https://github.com/elastic/beats/blob/master/metricbeat/module/prometheus/_meta/fields.yml#L7 as example.

metricbeat/module/php_fpm/php_fpm.go
```diff
@@ -0,0 +1,103 @@
...
+
+const (
+	defaultScheme = "http"
+	defaultPath   = "/status"
+)
+
+// HostParser is used for parsing the configured php-fpm hosts.
+var HostParser = parse.URLHostParserBuilder{
+	DefaultScheme: defaultScheme,
+	DefaultPath:   defaultPath,
```
ruflin: To define a config option to change the `defaultPath` you can use `PathConfigKey`. See https://github.com/elastic/beats/blob/master/metricbeat/module/prometheus/collector/collector.go#L25 as an example. I would suggest to use `path` or `status_path` as the config option. All other parts are done automatically by the HostParser.

metricbeat/module/php_fpm/pool/pool.go
```diff
@@ -0,0 +1,83 @@
...
+	mb.BaseMetricSet
+	client *php_fpm.StatsClient // StatsClient that is reused across requests.
+}
+
+// New create a new instance of the MetricSet
+// Part of new is also setting up the configuration by processing additional
+// configuration entries if needed.
+func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
+
+	config := struct{}{}
```
ruflin: You can remove the config part as it is empty in your case.

metricbeat/module/php_fpm/pool/pool.go
```diff
@@ -0,0 +1,83 @@
...
+func (m *MetricSet) Fetch() (common.MapStr, error) {
+	body, err := m.client.Fetch()
+
+	if err != nil {
+		return nil, err
+	}
+
+	defer body.Close()
+
+	stats := &php_fpm.PoolStats{}
```
ruflin: I would recommend to put this part into a `data.go`file under the `pool` package including the PoolStats struct. Or does it have to be on the module level?

metricbeat/module/php_fpm/proc/proc.go
```diff
@@ -0,0 +1,88 @@
...
+	mb.BaseMetricSet
+	client *php_fpm.StatsClient // StatsClient that is reused across requests.
+}
+
+// New create a new instance of the MetricSet
+// Part of new is also setting up the configuration by processing additional
+// configuration entries if needed.
+func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
+
+	config := struct{}{}
```
ruflin: Can be removed

metricbeat/module/php_fpm/proc/proc.go
```diff
@@ -0,0 +1,88 @@
...
+func (m *MetricSet) Fetch() ([]common.MapStr, error) {
+	body, err := m.client.Fetch()
+
+	if err != nil {
+		return nil, err
+	}
+
+	defer body.Close()
+
+	stats := &php_fpm.FullStats{}
```
ruflin: This could be moved into `data.go` including the FullStats struct.

metricbeat/module/php_fpm/proc/proc.go
```diff
@@ -0,0 +1,88 @@
...
+	for _, proc := range stats.Processes {
+		events = append(events, common.MapStr{
+			"hostname": m.Host(),
+
+			"pid":                 strconv.Itoa(proc.Pid),
+			"state":               proc.State,
+			"start_time":          proc.StartTime,
+			"start_since":         proc.StartSince,
+			"requests":            proc.Requests,
+			"request_duration":    proc.RequestDuration,
```
ruflin: I would change the fields below as following:
```
request.duration
request.method
request.uri
```

metricbeat/module/php_fpm/proc/_meta/fields.yml
```diff
@@ -0,0 +1,74 @@
...
+      type: long
+      description: >
+        Number of seconds since the process has started.
+
+    - name: requests
+      type: long
+      description: >
+        The number of requests the process has served.
+
+    - name: request_duration
```
ruflin: request.duration.us if it is micro seconds. See https://www.elastic.co/guide/en/beats/libbeat/5.1/event-conventions.html

metricbeat/module/php_fpm/proc/_meta/fields.yml
```diff
@@ -0,0 +1,74 @@
...
+      type: keyword
+      description: >
+        The user (`PHP_AUTH_USER`) (or `-` if not set).
+
+    - name: script
+      type: keyword
+      description: >
+        The main script called (or `-` if not set).
+
+    - name: last_request_cpu
```
ruflin: last_request.cpu

In case it is a percent value, I would suggest `last_request.cpu.pct` and make it a `scaled_float`.

metricbeat/module/php_fpm/proc/_meta/fields.yml
```diff
@@ -0,0 +1,74 @@
...
+      type: float
+      description: >
+        The %cpu the last request consumed it's always 0 if the process is not
+        in `Idle` state because CPU calculation is done when the request
+        processing has terminated.
+
+    - name: last_request_memory
```
ruflin: last_request.memory. Is this in bytes? If yes, it should be:

```
- name: last_request.memory.bytes
  type: long
  format: bytes
```

This will make it possible that kibana can do the conversions automatically.

metricbeat/module/php_fpm/pool/_meta/fields.yml
```diff
@@ -0,0 +1,78 @@
...
+      type: long
+      description: >
+        The date and time FPM has started.
+
+    - name: start_since
+      type: long
+      description: >
+        Number of seconds since FPM has started.
+
+    - name: accepted_conn
```
ruflin: connections.accepted

metricbeat/module/php_fpm/pool/_meta/fields.yml
```diff
@@ -0,0 +1,78 @@
...
+      type: long
+      description: >
+        The number of request accepted by the pool.
+
+    - name: listen_queue
```
ruflin: listen_queue.pending

metricbeat/module/php_fpm/pool/_meta/fields.yml
```diff
@@ -0,0 +1,78 @@
...
+      type: long
+      description: >
+        The number of request in the queue of pending connections.
+
+    - name: max_listen_queue
```
ruflin: listen_queue.max

metricbeat/module/php_fpm/pool/_meta/fields.yml
```diff
@@ -0,0 +1,78 @@
...
+      type: long
+      description: >
+        The maximum number of requests in the queue of pending connections
+        since FPM has started.
+
+    - name: listen_queue_len
```
ruflin: listen_queue.length

metricbeat/module/php_fpm/pool/_meta/fields.yml
```diff
@@ -0,0 +1,78 @@
...
+      type: long
+      description: >
+        The size of the socket queue of pending connections.
+
+    - name: idle_processes
```
ruflin: processes.idle

metricbeat/module/php_fpm/pool/_meta/fields.yml
```diff
@@ -0,0 +1,78 @@
...
+      type: long
+      description: >
+        The number of idle processes.
+
+    - name: active_processes
```
ruflin: processes.active

metricbeat/module/php_fpm/pool/_meta/fields.yml
```diff
@@ -0,0 +1,78 @@
...
+      type: long
+      description: >
+        The number of active processes.
+
+    - name: total_processes
```
ruflin: processes.total

metricbeat/module/php_fpm/pool/_meta/fields.yml
```diff
@@ -0,0 +1,78 @@
...
+      type: long
+      description: >
+        The number of idle + active processes.
+
+    - name: max_active_processes
```
ruflin: processes.max_active

metricbeat/module/php_fpm/php_fpm.go
```diff
@@ -0,0 +1,103 @@
...
+
+	if resp.StatusCode != 200 {
+		return nil, fmt.Errorf("HTTP error %d: %s", resp.StatusCode, resp.Status)
+	}
+
+	return resp.Body, nil
+}
+
+// PoolStats defines all stats fields from a php-fpm pool
+type PoolStats struct {
```
ruflin: If the response is in JSON, you could use the `schema` package to do the conversion. But we can also do that in a follow up PR.

metricbeat/docs/modules/php_fpm.asciidoc
tsouza: php-fpm default is `9000` and default status endpoint is `/status`. But these defaults does matter much since it's `fastcgi` protocol and can't be accessed directly. So there is a need to configure a proxy (with `fastcgi` support) and that can be any part and path (no defaults)

metricbeat/module/php_fpm/_meta/config.yml
tsouza: The module I started looking at to learn is `haproxy`, which is no commented out. But I'll do it then!

metricbeat/module/php_fpm/php_fpm.go
```diff
@@ -0,0 +1,103 @@
...
+	address  string
+	user     string
+	password string
+	http     *http.Client
+}
+
+// NewStatsClient creates a new StatsClient
+func NewStatsClient(m mb.BaseMetricSet, isFullStats bool) *StatsClient {
+	var address string
+	address = m.HostData().SanitizedURI + "?json"
```
andrewkroh: You can move the query param to the `URLHostParserBuilder` by setting `QueryParams: "json"`.

andrewkroh: You can just write `address := m.HostData().SanitizedURI` and drop the `var address string`. (https://golang.org/ref/spec#Short_variable_declarations)

metricbeat/module/php_fpm/php_fpm.go
```diff
@@ -0,0 +1,103 @@
...
+		address:  address,
+		user:     m.HostData().User,
+		password: m.HostData().Password,
+		http:     &http.Client{Timeout: m.Module().Config().Timeout},
+	}
+}
+
+// Fetch php-fpm stats
+func (c *StatsClient) Fetch() (io.ReadCloser, error) {
+	req, err := http.NewRequest("GET", c.address, nil)
```
andrewkroh: Once #3413 merges we should update this to use it.

metricbeat/module/php_fpm/_meta/fields.yml
tsouza: done

metricbeat/module/php_fpm/php_fpm.go
tsouza: done

metricbeat/module/php_fpm/php_fpm.go
tsouza: Indeed the response is in JSON and also it came to me that there would be a better way of handling this situation. But I wonder what will happen to the field names itself, since php-fpm replies a JSON with whitespace in field names (e.g. `accepted conn`)

Mind pointing out an implementation that uses it?

metricbeat/docs/modules/php_fpm.asciidoc
ruflin: Hm, problem is with the above port 80 is the default. Not sure what we should set here best.

metricbeat/module/php_fpm/php_fpm.go
ruflin: Not directly JSON, but similar: https://github.com/elastic/beats/blob/master/metricbeat/module/mongodb/status/data.go

metricbeat/module/php_fpm/_meta/fields.yml
ruflin: I think I forgot to mention it in the review, but also the metricsets should have a log line in the New(...) part that has a warning that it is experimental. Also see prometheus.

metricbeat/module/php_fpm/_meta/config.yml
ruflin: Ah, haproxy has a config.yml and a config.full.yml, that is why it also works there.

metricbeat/docs/modules/php_fpm.asciidoc
tsouza: Maybe we can use the "default non-default" 8080 (which is at least nginx's default, iirc)

ruflin: SGTM

metricbeat/module/php_fpm/pool/pool.go
tsouza: I believe it can go to `data.go`. There _might_ be a minor issue, if it happens I'll reach out for help :)

metricbeat/module/php_fpm/php_fpm.go
tsouza: Actually, I was thinking of skipping the http proxy and directly access php-fpm using a fastcgi client. I could find this one [go-fastcgi-client](https://github.com/flashmob/go-fastcgi-client) (since golang's fcgi package is for server only). What do you think?

tsouza:
The current metric names and it's description was copied directly from [php-fpm configuration file](https://github.com/php/php-src/blob/master/sapi/fpm/www.conf.in#L135). As they do not follow beats convention much (as pointed out by @ruflin), I decided to take a closer look to find proper names.

It seems that the process metrics regarding `request` IMO are useless. The metrics `request duration`, `request method` and `request uri` are from the **current** request at the time the status endpoint was invoked (which might temporarily become the last, if no request is being served by the process). Whereas the `last request cpu/memory` are calculated  only _after_  the request is done. Thus, in a running system serving several requests per second, there won't be any real correlation between `last request cpu/memory` and those `request *` metrics.

So I am worried about that users might start to correlate between `request uri` and `last request cpu`, which might be misleading. This makes me wonder how useful those metrics might actually be.

ruflin:
@tsouza If possible I would suggest to leave out calculated values as this can also be potentially done on the elasticsearch side. The way you describe it above it seems even to be temporary so not too usefule as you said.

To move this PR forward I suggest to remove the values you are not sure about are useful. The hard part of doing a Metricset is getting all the data and do the conversion. Adding a field is fairly easy. So in case someone requests it, we get the opportunity to ask why it is needed and how it is used and we can easily add it. WDYT?

tsouza:
@ruflin Agreed. I'll just keep the minimum set of metrics that I judge useful.

metricbeat/module/php_fpm/pool/pool.go
```diff
@@ -0,0 +1,70 @@
...
+	err = json.NewDecoder(body).Decode(stats)
+	if err != nil {
+		return nil, fmt.Errorf("error parsing json: %v", err)
+	}
+
+	return common.MapStr{
+		"hostname": m.Host(),
+
+		"pool":                 stats.Pool,
+		"connections.queued":   stats.ListenQueue,
```
ruflin: You have to nest this into a common.MapStr as otherwise elasticsearch 2.x will complain here. So this becomes:

```
"connections": common.MapStr{
  "queued": stats.ListenQueue,
  "accepted" : stats.AcceptdConn,
}
```

tsouza: Ok! The docs was not clear in that sense to me! So I just took a shot hah

ruflin: I think we actually detected this problem with 2.x only after we wrote the docs ;-)

PR #3416, opened 2017-01-19
Fixes for dev-tool/jenkins_ci

andrewkroh:
- Use absolute path to gimme inside of scripts to avoid requiring $HOME/bin be on the PATH.
- export the RACE_DETECTOR variable from the jenkins_ci script so that it is picked up by make
- Add execute permissions


andrewkroh:
jenkins, package it

andrewkroh:
Closing in favor of #3417

PR #3424, opened 2017-01-20
Fix testing index template

ruflin:
`index` now only supports true or false.


ruflin:
Closing in favour of https://github.com/elastic/beats/pull/3425

PR #3427, opened 2017-01-20
introduce libbeat.monitoring to unify collection from internal metrics

urso:
expose some simple metrics collection via `libbeat/monitoring` package. 

Changes/Features:
- 30s metrics snapshot is now based on `libbeat/monitoring`, adding:
  - support for bool/float/string variables
  - keys are sorted before being printed
-  The package manages a hierarchical registry of known KPIs (names are split on `.`) with optional support for registering created KPIs to expvar package (metrics registered via expvar can not be removed).
- All metrics registered must support the `monitoring.Var` interface providing a `Visit` method for reporting metric values. The `monitoring.Visitor` explicitly limits the type of values being reportable, so no `interface{}` will be used, simplifying/unifying reporting/collecting metrics.
- Registry provides `Do` method to iterate all variables, with names being 'flattened' and thanks to limitations imposed by `monitoring.Var/Visitor` the values reported can only be int64, float64, string, bool or untyped nil. 
- Having a registry lowers chances of typos. e.g.

```
var (
    metrics := monitoring.Default.NewRegistry("libbeat.outputs.logstash")
    bytesSend := metrics.NewInt("bytes_send")
    ...
)
```
- package adds support for dynamic removal of metrics
- provides adapter for go-metrics (e.g. collect stats from kafka output) with selective whitelisting/renaming... for collecting stats we otherwise would have no access to.

### DIFF
--- a/libbeat/logp/metrics.go
+++ b/libbeat/logp/metrics.go
@@ -1,0 +1,186 @@
+package logp
+
+import (
+	"bytes"
+	"fmt"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/elastic/beats/libbeat/monitoring"
+)
+
+type snapshotVisitor struct {
+	snapshot snapshot
+	level    []string
+}
+
+type snapshot struct {
+	bools   map[string]bool
+	ints    map[string]int64
+	floats  map[string]float64
+	strings map[string]string
+}
+
+// logExpvars logs at Info level the integer expvars that have changed in the
+// last interval. For each expvar, the delta from the beginning of the interval
+// is logged.
+func logMetrics(metricsCfg *LoggingMetricsConfig) {
+	if metricsCfg.Enabled != nil && *metricsCfg.Enabled == false {
+		Info("Metrics logging disabled")
+		return
+	}
+	if metricsCfg.Period == nil {
+		metricsCfg.Period = &defaultMetricsPeriod
+	}
+	Info("Metrics logging every %s", metricsCfg.Period)
+
+	ticker := time.NewTicker(*metricsCfg.Period)
+
+	prevVals := makeSnapshot()
+	for range ticker.C {
+		snapshot := snapshotMetrics()
+		delta := snapshotDelta(prevVals, snapshot)
+		prevVals = snapshot
+
+		if len(delta) == 0 {
+			Info("No non-zero metrics in the last %s", metricsCfg.Period)
+			continue
+		}
+
+		metrics := formatMetrics(delta)
+		Info("Non-zero metrics in the last %s:%s", metricsCfg.Period, metrics)
+	}
+}
+
+func LogTotalExpvars(cfg *Logging) {
+	if cfg.Metrics.Enabled != nil && *cfg.Metrics.Enabled == false {
+		return
+	}
+
+	metrics := formatMetrics(snapshotDelta(makeSnapshot(), snapshotMetrics()))
+	Info("Total non-zero values: %s", metrics)
+	Info("Uptime: %s", time.Now().Sub(startTime))
+}
+
+func snapshotMetrics() snapshot {
+	vs := newSnapshotVisitor()
+	monitoring.Default.Visit(vs)
+	monitoring.VisitExpvars(vs)
+	return vs.snapshot
+}
+
+func newSnapshotVisitor() *snapshotVisitor {
+	return &snapshotVisitor{snapshot: makeSnapshot()}
+}
+
+func makeSnapshot() snapshot {
+	return snapshot{
+		bools:   map[string]bool{},
+		ints:    map[string]int64{},
+		floats:  map[string]float64{},
+		strings: map[string]string{},
+	}
+}
+
+func (vs *snapshotVisitor) OnRegistryStart() error {
+	return nil
+}
+
+func (vs *snapshotVisitor) OnRegistryFinished() error {
+	if len(vs.level) > 0 {
+		vs.dropName()
+	}
+	return nil
+}
+
+func (vs *snapshotVisitor) OnKey(name string) error {
+	vs.level = append(vs.level, name)
+	return nil
+}
+
+func (vs *snapshotVisitor) OnKeyNext() error { return nil }
+
+func (vs *snapshotVisitor) getName() string {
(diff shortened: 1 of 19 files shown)


libbeat/outputs/kafka/kafka.go
```diff
@@ -270,6 +272,9 @@ func (k *kafka) newKafkaConfig() (*sarama.Config, error) {
 	}
 
 	cfg.Producer.Partitioner = k.partitioner
+
+	// TODO: figure out which metrics we want to collect
+	cfg.MetricRegistry = adapter.GetGoMetrics(monitoring.Default, "libbeat.output.kafka")
```
ruflin: I would probably go with `output.kafka` and not have the libbeat part. I'm aware that we have libbeat in front for other outputs part so far. But I think outputs are so independent (and potentially not necessarly in libbeat itself (plugins) in the future), that we can remove the `libbeat` part.

urso: yeah, this is some cleaning for a later phase. All outputs should use a similar convention in the end.

libbeat/monitoring/config.go
```diff
@@ -0,0 +1,25 @@
...
+	"time"
+
+	"github.com/elastic/beats/libbeat/outputs"
+)
+
+type Config struct {
+	Elasticsearch *elasticsearchConfig `config:"elasticsearch"`
+}
+
+type elasticsearchConfig struct {
```
ruflin: Lets move the output part ot an other PR

ruflin:
👍  on this.

libbeat/monitoring/adapter/go-metrics.go
```diff
@@ -0,0 +1,165 @@
...
+		w, ok := goMetricsWrap(st.metric)
+		if ok {
+			r.reg.Add(st.name, w)
+		}
+	}
+
+	return st.metric
+}
+
+func (r *GoMetricsRegistry) RunHealthchecks() {}
```
andrewkroh: What's the purpose of this method? Just to satisfy the sarama interface?

andrewkroh: Oh I see it's part of the go-metrics [Registry](https://godoc.org/github.com/rcrowley/go-metrics#RunHealthchecks).

urso: yes, purpose of `adapter` package is to integrate other metrics collection into `libbeat/monitoring`.

libbeat/logp/metrics.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+type snapshot struct {
+	bools   map[string]bool
+	ints    map[string]int64
+	floats  map[string]float64
+	strings map[string]string
+}
+
+// logExpvars logs at Info level the integer expvars that have changed in the
```
ruflin: s/logExpvars/logMetrics/ and probably some more updates to the docs needed

libbeat/monitoring/adapter/filters.go
```diff
@@ -0,0 +1,144 @@
...
+func ModifyName(f func(string) string) MetricFilter {
+	return withVarFilter(func(st state) state {
+		st.name = f(st.name)
+		return st
+	})
+}
+
+// Rename renames a metric to `to`, if the names matches `from`
+// If the name matches, it will be automatically white-listed.
+func Rename(from, to string) MetricFilter {
```
ruflin: Can you share some background here? What is the reasons this is needed? Converting metrics coming from `go-metrics`?

libbeat/monitoring/opts.go
```diff
@@ -0,0 +1,46 @@
...
+	publishExpvar bool
+}
+
+var defaultOptions = options{
+	publishExpvar: false,
+}
+
+// PublishExpvar enables publishing all registered variables via expvar interface.
+// Note: expvar does not allow removal of any stats.
+func PublishExpvar(o options) options {
```
ruflin: I'm thinking if we should move this to `libbeat/monitoring/publisher` into its own namespace. Just in case we have more publishers for metrics in the future ;-)

libbeat/outputs/kafka/kafka.go
```diff
@@ -270,6 +272,15 @@ func (k *kafka) newKafkaConfig() (*sarama.Config, error) {
 	}
 
 	cfg.Producer.Partitioner = k.partitioner
+
+	// TODO: figure out which metrics we want to collect
+	cfg.MetricRegistry = adapter.GetGoMetrics(
+		monitoring.Default,
+		"libbeat.output.kafka",
+		adapter.Rename("incoming-byte-rate", "bytes_read"),
```
ruflin: Here we go 👍

libbeat/logp/metrics.go
```diff
@@ -0,0 +1,186 @@
...
+// last interval. For each expvar, the delta from the beginning of the interval
+// is logged.
+func logMetrics(metricsCfg *LoggingMetricsConfig) {
+	if metricsCfg.Enabled != nil && *metricsCfg.Enabled == false {
```
urso: yeah, I just moved the old code over. Publishing will be addressed in another PR.

urso:
> Worth a CHANGELOG entry?

It's just changing/adding some internals and adds no real functionality yet. I'd say no (as we've got no developers changelog)

libbeat/monitoring/adapter/filters.go
urso: Filters is for selecting and manipulating metrics from other metrics interfaces (so far go-metrics). In go-metrics case, the metrics not selected will be owned by a regular shadow-go-metrics registry, so libs do not rebuild metrics over and over again.

libbeat/monitoring/opts.go
urso: yes, it's exactly what I want todo. This `PublishExpvar` is an option one can pass to `monitoring.Registry`. If this is enabled, the metrics created via `monitoring` will be made available in the `expvar` interface as well. To make a custom metric available to `expvar` one has to call `expvar.Publish`, that's why it's named `PublishExpvar`. I'm considering to remove `PublishExpvar` though.

PR #3429, opened 2017-01-20
Deprecate filebeat.config_dir config option

ruflin:
The plan is replace this with a better implementation in the future similar to prospector reloading. See https://github.com/elastic/beats/pull/3362.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -111,6 +111,7 @@ https://github.com/elastic/beats/compare/v5.1.1...master[Check the HEAD diff]
 - Fix error on importing dashboards due to colons in the Caassandra dashboard. {issue}3140[3140]
 
 *Filebeat*
+- Config option filebeat.config_dir is deprecated. It will be replaced in version 6.0 with a new implementation that also supports configuration reloading. {issue}3430[3430]
 
 *Winlogbeat*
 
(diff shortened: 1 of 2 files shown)


tsg:
Is it really helpful to deprecate it without an alternative?

ruflin:
We should definitively deprecate it int 5.3 to make sure we can remove / change it in 6.0. Deprecating does not necessarily mean we will remove it. I would also not remove it before we have a replacement. Here is a follow up Github issue to discuss the new configs: https://github.com/elastic/beats/issues/3430

CHANGELOG.asciidoc
```diff
@@ -111,6 +111,7 @@ https://github.com/elastic/beats/compare/v5.1.1...master[Check the HEAD diff]
 - Fix error on importing dashboards due to colons in the Caassandra dashboard. {issue}3140[3140]
 
 *Filebeat*
+- Config option `filebeat.config_dir` is deprecated. It is planned to replace it with a new implementation.
```
tsg: I suggest: "Config option `filebeat.config_dir` is deprecated. It will be replaced in version 6.0 with a new implementation that also supports configuration reloading"

ruflin: SGTM. I'm confident enough in https://github.com/elastic/beats/issues/3430 to get it in for 6.0

monicasarbu:
@ruflin Can you please update this PR? Thanks!

ruflin:
@tsg @monicasarbu I suggest to open this PR against 5.x to deprecate it there. And for master a PR is needed that does remove the old implementation and switches it over the new reload configuration.

PR #3433, opened 2017-01-20
Per prospector configurable pipeline

tsg:
This adds a new "pipeline" configuration option to the prospector, which
can be used to set the Elasticsearch Ingest Node pipeline from the prospector
config.

While this was already possible by using format strings in the `pipeline` config
from the output, this makes the configuration simpler in many cases and the
mechanism is needed for the Filebeat modules.

Part of #3159.


filebeat/input/event.go
```diff
@@ -58,6 +59,16 @@ func (e *Event) ToMapStr() common.MapStr {
...
 
+// Metadata creates a common.MapStr containing the metadata to
+// be associated with the event.
+func (e *Event) Metadata() common.MapStr {
+	meta := common.MapStr{}
+	if len(e.Pipeline) > 0 {
+		meta["pipeline"] = e.Pipeline
+	}
+	return meta
+}
```
urso: output metadata should also work if `meta == nil` =>

```
func (e *Event) Metadata() common.MapStr {
  if e.Pipeline != "" {
    return common.MapStr{
      "pipeline": e.Pipeline,
    }
  }
  return nil

}
```

This way you don't need to create a map if no pipeline is configured.

tsg: Nice, I didn't realize that would work.

tsg: I pushed a version with that.

PR #3434, opened 2017-01-20
Fix spelling mistakes

bengadbois:


dev-tools/vendor/github.com/cavaliercoder/go-rpm/packagefile.go
```diff
@@ -13,7 +13,7 @@ import (
 	"time"
 )
 
-// A PackageFile is an RPM package definition loaded directly from the pacakge
+// A PackageFile is an RPM package definition loaded directly from the package
```
tsg: Can you remove this one please? It's in a dependency so it doesn't help to much to fix the typo in our vendored copy.

dev-tools/vendor/github.com/cavaliercoder/go-rpm/packagefile.go
```diff
@@ -167,7 +167,7 @@ func (c *PackageFile) FileSize() uint64 {
 	return c.fileSize
 }
 
-// Checksum computes and returns the SHA256 checksum (encoded in hexidecimal) of
+// Checksum computes and returns the SHA256 checksum (encoded in hexadecimal) of
```
tsg: Same on this one.

packetbeat/flows/util.go
```diff
@@ -66,7 +66,7 @@ func (w *worker) tick(t *time.Ticker) bool {
 	}
 }
 
-func (w *worker) periodicaly(tick time.Duration, fn func() error) {
+func (w *worker) periodically(tick time.Duration, fn func() error) {
```
tsg: There's some unit [test failures](https://travis-ci.org/elastic/beats/jobs/193893438#L405) which I suspect are caused by this change.

tsg:
Thanks for the fixes, I left some comments. Out of curiosity, did you use a tool for this? Perhaps we can integrate it in our processes.

bengadbois:
I ran [Gometalinter](https://github.com/alecthomas/gometalinter) which includes a number of opinionated suggestions, but the specific linter that caught these was [misspell](https://github.com/client9/misspell).

[This](https://github.com/mholt/caddy/commit/c7c34266dab44ed222becc4a8eef982cd1707b2a) is how I added it to the Caddy's build.

I've personally not had false positives, but might not want to make that a build passing requirement if you choose to adopt it.

packetbeat/flows/util.go
bengadbois: Should be fixed now

tsg:
jenkins, test it

tsg:
Thanks @bengadbois!

PR #3435, opened 2017-01-22
Fix the cookie cutter template for beat

johnthedev97:
Added cookie cutter.json
Convert all template variables into cookiecutter namespace and inside double braces


johnthedev97:
According to documentation here (https://www.elastic.co/guide/en/beats/libbeat/current/newbeat-generate.html) the command to generate beat is ''cookiecutter $GOPATH/src/github.com/elastic/beats/generate/beat". But the travis build seems to be generating the beat by running 'beat.py' instead of the documented method, which is the expected method?

ruflin:
@johnthedev97 It depends on the version. In master I started to remove the dependency on cookiecutter as our templates are so basic it is overkill to require people to install an extra tool. And not using cookiecutter allows us to do some additional checks. Here are the docs based on master: https://www.elastic.co/guide/en/beats/libbeat/master/newbeat-generate.html

Until branch 5.2 cookiecutter is used.

ruflin:
I'm closing this as we are moving away from cookiecutter.

@johnthedev97 Keep an eye on https://github.com/elastic/beats/pull/3452 as this PR contains the next major changes to the generator.

PR #3436, opened 2017-01-22
Deploy Filebeat modules in the packages

tsg:
Uses an intermediary `_meta/module.generated/` folder which is created on
`make collect`. This gives us the opportunity to select which files are
needed to be deployed in the package (currently everything but the `_meta` and
`test` folders).

The intermediary folder also allows us to avoid providing a custom `install-home`
target in Filebeat.

Part of #3159.

### DIFF
--- a/filebeat/.gitignore
+++ b/filebeat/.gitignore
@@ -6,4 +6,5 @@ filebeat
 
 build
 _meta/kibana
+_meta/module.generated
 /tests/load/logs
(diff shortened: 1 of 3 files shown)


tsg:
jenkins, package it

ruflin:
Could we move this into the build directory where it would be ignored automatically? Not 100% sure if we should use a `.` (dot) directory. I remember something that we opted for not using `.` for some files because it could problems on Windows. What about `module.generate` without the leading dot? That would be consistent with the files which are generated.

tsg:
@ruflin ah, I didn't think of the build directory, I can easily do that. Then the dot is not needed, I added it to avoid the confusion.

tsg:
Setting in progress as this seems to cause issues on local package building.

tsg:
Removed 'in progress' since the issue seems to exists only on my computer.

ruflin:
jenkins, package it

tsg:
@ruflin I suspect the `build/` directory is cleaned during packaging, because with the new version the files are not added.

tsg:
jenkins, package it

tsg:
I changed the temp folder to `_meta/module.generated`.

PR #3438, opened 2017-01-23
Change Metricbeat generator to be based on python

ruflin:
This removes the dependency on cookiecutter for generating a beat based on Metricbeat.

Further changes:

* Unify handling of beat_path in the generator scripts
* Merge generator script for both beats types
* Move metricbeat generator out of metricset directory as it generates a beat and metricset is only a follow up call.
* Enabled metricbeat modules by default to have something to get started.


ruflin:
yep, I did. Any there are also the automated tests ;-)

PR #3439, opened 2017-01-23
Unify reloading

ruflin:


libbeat/cfgfile/reload.go
```diff
@@ -15,3 +31,158 @@ type ReloadConfig struct {
 	Period  time.Duration `config:"period"`
 	Enabled bool          `config:"enabled"`
 }
+
+type Wrapper interface {
```
andrewkroh: This looks like the interface to a factory. Maybe you could call it `RunnerFactory` with a `Create` method. 

I think it would be better to integrate the `uint64` value into the `Runner` that is returned. I'm assuming this is some sort of identifier? So probably an `ID()` method in `Runner` would make sense.

libbeat/cfgfile/reload.go
```diff
@@ -15,3 +31,158 @@ type ReloadConfig struct {
...
+		wg.Add(1)
+		// Stop modules in parallel
+		func() {
+			defer wg.Done()
+			w.Stop()
+			r.registry.Remove(hash)
+			moduleStops.Add(1)
+			moduleRunning.Add(-1)
+			debugf("Module stopped: %v", hash)
+		}()
```
andrewkroh: Is there a test for this? `hash` and `w` should be passed into the function through arguments.

libbeat/cfgfile/reload.go
```diff
@@ -15,3 +31,158 @@ type ReloadConfig struct {
...
+}
+
+func (r *Reloader) stopModules(list map[uint64]Runner) {
+	logp.Info("Stopping %v modules ...", len(list))
+
+	wg := sync.WaitGroup{}
+	for hash, w := range list {
+		wg.Add(1)
+		// Stop modules in parallel
+		func() {
```
andrewkroh: Missing the `go` as in `go func(){...}()`?

libbeat/cfgfile/reload.go
ruflin: I like `RunnerFactory` and `Create`. For the `ID()` agree, but still need to do some cleanup to get there ...

libbeat/cfgfile/reload.go
ruflin: definitively :-(

ruflin:
@andrewkroh Thanks for the early input / feedback. PR still needs quite a bit of work.

libbeat/cfgfile/reload.go
```diff
@@ -15,3 +31,163 @@ type ReloadConfig struct {
 	Period  time.Duration `config:"period"`
 	Enabled bool          `config:"enabled"`
 }
+
+type RunnerFactory interface {
+	Create(*common.Config) (Runner, error)
+}
```
urso: having just one method how about:

```
type RunnerFactory func(*common.Config) (Runner, error)
```

This is already a very common pattern in beats.

filebeat/prospector/factory.go
```diff
@@ -0,0 +1,38 @@
...
+	}
+
+	err = p.LoadStates(r.registrar.GetStates())
+	if err != nil {
+		logp.Err("Error loading states for prospector %v: %v", p.ID, err)
+		return nil, 0, err
+	}
+
+	return p, nil
+}
```
urso: when turning `RunnerFactory` you either pass `NewFactory.Create` to the registry or turn this into:

```
func MakeFactory(outlet Outlet, registrar *registrar.Registrar) cfgfile.RunnerFactory {
  return func(c *common.Config) (cfgfile.Runner, uint64, error) {
    p, err := NewProspector(c, r.outlet)
    if err != nil {
	    logp.Err("Error creating prospector: %s", err)
	    return nil, 0, err
    }
    
    err = p.LoadStates(r.registrar.GetStates())
    if err != nil {
	    logp.Err("Error loading states for prospector %v: %v", p.ID, err)
	    return nil, 0, err
    }
    
    return p, nil 
  }
}
```

filebeat/prospector/factory.go
```diff
@@ -0,0 +1,38 @@
...
+		return nil, 0, err
+	}
+
+	err = p.LoadStates(r.registrar.GetStates())
+	if err != nil {
+		logp.Err("Error loading states for prospector %v: %v", p.ID, err)
+		return nil, 0, err
+	}
+
+	return p, nil
```
urso: you missing the ID here?

filebeat/prospector/factory.go
```diff
@@ -0,0 +1,38 @@
...
+
+func NewFactory(outlet Outlet, registrar *registrar.Registrar) *Factory {
+
+	return &Factory{
+		outlet:    outlet,
+		registrar: registrar,
+	}
+}
+
+func (r *Factory) Create(c *common.Config) (cfgfile.Runner, uint64, error) {
```
urso: with `Runner` implementing `ID()`, we need to return an ID here? Or put alternatively: do we need `Runner` to implement `ID()` ?

libbeat/cfgfile/registry.go
```diff
@@ -1,39 +1,39 @@
-package module
+package cfgfile
 
 import "sync"
 
-type registry struct {
+type Registry struct {
 	sync.Mutex
 	List map[uint64]Runner
```
urso: registry already stores the id, we need `Runner` to have method `ID()`?

filebeat/crawler/crawler.go
```diff
@@ -16,7 +17,7 @@ type Crawler struct {
 	prospectorConfigs []*common.Config
 	out               prospector.Outlet
 	wg                sync.WaitGroup
-	reloader          *prospector.ProspectorReloader
+	reloader          *cfgfile.Reloader
```
ruflin: This is local to the crawler, so I think the naming fits as it reloads crawler specific configs (prospectors).

libbeat/cfgfile/registry.go
ruflin: Yes, as we have to initially fetch it from the Runner. In the initial PR it was returned but I made it now part of the interface. See discussion here: https://github.com/elastic/beats/pull/3439#discussion_r98434613

libbeat/cfgfile/reload.go
ruflin: I wonder who created this common pattern ;-) I personally prefer the interface one.

filebeat/prospector/factory.go
ruflin: argh, that is a left over from removing returning id.

libbeat/cfgfile/reload.go
```diff
@@ -15,3 +31,163 @@ type ReloadConfig struct {
...
+type Runner interface {
+	Start()
+	Stop()
+	ID() uint64
```
ruflin: I was hoping yes, but heartbeat will be the next proving ground which could affect this one again. I would expect heartbeat to use the same hashing on the config options to generate the hash.

libbeat/cfgfile/reload.go
```diff
@@ -15,3 +31,163 @@ type ReloadConfig struct {
 	Period  time.Duration `config:"period"`
 	Enabled bool          `config:"enabled"`
 }
+
+type RunnerFactory interface {
+	Create(*common.Config) (Runner, error)
+}
+type Runner interface {
```
ruflin: interesting that gofmt didn't do it.

PR #3441, opened 2017-01-23
Fix processor_test.go build failure on 32-bit systems

spacewrangler:
Int size is platform/implementation-specific (defined to be "[at least 32-bits"](https://golang.org/src/builtin/builtin.go?#L73)). On 32-bit systems it's generally 32-bits. When an empty interface is assigned an integer value greater than the max value a 32-bit integer can hold, build will fail when the compiler converts that value to an int.


urso:
thanks for fixing the test. I think the beat internally uses int64 or uint64 (looks like metricbeat system module stats). I wonder if using `int64(...)` would have fixed this as well.

spacewrangler:
Yes, that would work as well. I tend to prefer the explicit declaration also. I'll create a new PR for it.

PR #3446, opened 2017-01-24
Fix processor_test.go build failure on 32-bit systems

spacewrangler:
Int size is platform/implementation-specific (defined to be "at least 32-bits"). On 32-bit systems it's generally 32-bits. When an empty interface is assigned an integer value greater than the max value a 32-bit integer can hold, build will fail when the compiler converts that value to an int.


spacewrangler:
I haven't yet. I'm working through a few issues getting tests to pass on Raspbian running on a Rasberry Pi 3. (I think IoT scenarios are interesting for beats and I'd like to use beats to ship some sensor data to Elasticsearch.) Anyhow, a number of the issues seem to be environmental at this point but there may be other 32-bit ones lurking. I'll continue to bring issues I run into to the forums.

ruflin:
@spacewrangler Thanks for doing the testing around this and contributing back. Keep us posted.

PR #3449, opened 2017-01-24
Simplify building of docker containers test environment

ruflin:
This change will allow to change download path and stack version in a single place (`args.yml`) instead of having it to apply multiple times. This should simplify updating the stack to a new version.

* Set default env to snapshot as this is the most used one for manual testing
* Change also allows to use env to set different environments
* Update latest build to 5.1.2


ruflin:
Currently blocked by https://forums.docker.com/t/environment-variables-in-args/27522

PR #3452, opened 2017-01-24
Unify generator

ruflin:
* Use one script create generic beat or metricbeat
* Rename `generate` directory to `template`
* Share most parts of the testing makefile
* Update docs to new logic
* Update travis build file


monicasarbu:
@ruflin What about moving the generate script to the `template` directory to be together with the Beat templates, and rename the `template` directory to `generator`.?

ruflin:
@monicasarbu I actually just did the opposite step with this PR :-) The reason I put it under script (besides that it is a script) is that I think it is unlikely that we will change the name of the folder or move the folder in the near future. This means we will not break the generator again. All the other directories (be it template or generator or xyz) do not matter to the user as it we only have to adjust it inside the script. So I think we should keep `script/generate.py` but I have no strong opinion about the naming of the directory. Let me know what you prefer there.

monicasarbu:
@ruflin That's a good point. I would say to keep the script in the `script` directory, and rename the directory with the templates to `generator` even if it contains only the templates as you will use the script anyway to generate the templates.

ruflin:
@monicasarbu renamed

monicasarbu:
Waiting for green

PR #3454, opened 2017-01-24
Add TTL option to the logstash output

### LINKED ISSUE #661: Specify a TTL for connection from Filebeat to Logstash
A great enhancement would be to provide a time to live (TTL) config value to specify how long a TCP connection from Filebeat to Logstash should live.  After that TTL, a new connection would be created.  This would allow better distribution to Logstash instances behind a loadbalancer.

Since connections from filebeat to logstash are sticky, when an instance joins the load balancer, it does not get an equal distribution.  For example, if there are 4 instances behind a load balancer and 3 of them are rebooted, then all filebeat connections will go to the single instance that was not rebooted.  I am not aware of any other events, aside from failure, that would cause the connection to be reestablished to a potentially different server.

By specifying a TTL on the connection, there is an opportunity for the load balancer to distribute connections equally between the instances.  Using a load balancer tends to be more convenient than updating the configuration of all filebeat clients in an environment.

nicorevin:
Fixes #661

### DIFF
--- a/libbeat/outputs/logstash/async.go
+++ b/libbeat/outputs/logstash/async.go
@@ -1,6 +1,7 @@
 package logstash
 
 import (
+	"sync"
 	"sync/atomic"
 	"time"
 
@@ -15,7 +16,8 @@ type asyncClient struct {
 	*transport.Client
 	client *v2.AsyncClient
 	win    window
-	ttl    <-chan time.Time
+	ticker *time.Ticker
+	mutex  sync.Mutex
 
 	connect func() error
 }
@@ -27,6 +29,8 @@ type msgRef struct {
 	cb        func([]outputs.Data, error)
 	win       *window
 	batchSize int
+	client    *v2.AsyncClient
+	mu        *sync.Mutex
 }
 
 func newAsyncLumberjackClient(
@@ -41,7 +45,9 @@ func newAsyncLumberjackClient(
 	c := &asyncClient{}
 	c.Client = conn
 	c.win.init(defaultStartMaxWindowSize, maxWindowSize)
-	c.ttl = time.Tick(ttl)
+	if ttl > 0 {
+		c.ticker = time.NewTicker(ttl)
+	}
 
 	enc, err := makeLogstashEventEncoder(beat)
 	if err != nil {
@@ -68,13 +74,22 @@ func (c *asyncClient) Connect(timeout time.Duration) error {
 }
 
 func (c *asyncClient) Close() error {
+	if c.ticker != nil {
+		c.ticker.Stop()
+	}
 	logp.Debug("logstash", "close connection")
 	if c.client != nil {
-		err := c.client.Close()
+		err := closeClient(c.client, &c.mutex)
 		c.client = nil
 		return err
 	}
 	return c.Client.Close()
+}
+
+func closeClient(c *v2.AsyncClient, mutex *sync.Mutex) error {
+	mutex.Lock()
+	defer mutex.Unlock()
+	return c.Close()
 }
 
 func (c *asyncClient) AsyncPublishEvent(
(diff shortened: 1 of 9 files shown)


libbeat/outputs/logstash/async.go
```diff
@@ -128,6 +131,18 @@ func (c *asyncClient) publishWindowed(
...
+		case <-c.ttl:
+			if err := c.Close(); err != nil {
+				return 0, err
+			}
+			if err := c.Connect(0); err != nil {
+				return 0, err
+			}
+		default:
+		}
+	}
```
urso: I'm not sure I like to have the reconnect here. reason is the windowing. The output is supposed todo a slow start, increasing the window size to deal with potential timing problems if batch gets too big (e.g. connection lost, as processingt takes too long for very big events/batches). As this PR is supposed to help with load-balancers, after a reconnect the next batch might take another route. Here we should consider resetting the windowing parameters. Maybe move the timeout check before the call to publishWindowed and reset `win` to it's initial parameters.

libbeat/outputs/logstash/config.go
```diff
@@ -13,6 +13,7 @@ type logstashConfig struct {
 	LoadBalance      bool                  `config:"loadbalance"`
 	BulkMaxSize      int                   `config:"bulk_max_size"`
 	Timeout          time.Duration         `config:"timeout"`
+	TTL              time.Duration         `config:"ttl"`
```
urso: add `validate:"min=0"`

libbeat/outputs/logstash/sync.go
```diff
@@ -19,18 +19,21 @@ type client struct {
 	*transport.Client
 	client *v2.SyncClient
 	win    window
+	ttl    <-chan time.Time
```
urso: same here, use `time.Ticker` and `ttl.Stop` on close. Otherwise we introduce leaks in case we ever support reloading/reconfiguring outputs.

libbeat/outputs/logstash/sync.go
```diff
@@ -97,6 +100,18 @@ func (c *client) PublishEvents(
...
+		case <-c.ttl:
+			if err := c.Close(); err != nil {
+				return 0, err
+			}
+			if err := c.Connect(0); err != nil {
+				return 0, err
+			}
+		default:
+		}
+	}
```
urso: introduce a `reconnect` method + consider window sizing.

nicorevin:
@urso I think I've addressed all of your concerns though I'm still in doubt about `async.go`

nicorevin:
@urso Hi! Did you get a chance to review the changes?

urso:
@nicorevin sorry, haven't had a chance to look at it yet. Put it on top of my TODO list for tomorrow... sometimes it's hard to wrap ones head around async code ;)

urso:
Thinking more about the async code, I think this should be partially solved in [go-lumber](https://github.com/elastic/go-lumber) (to keep some sanity).

The idea is to signal the 'ackLoop' to close the connection once it has processed all message scheduled. After scheduling the 'shutdown', no more events must be send. This can be done by closing the channel used in this loop: https://github.com/elastic/go-lumber/blob/master/client/v2/async.go#L149.
The async client in go-lumber must ensure the channel is closed only once (e.g. use sync.Once), in order to deal with channel close on `Close`. Same for the underlying connection.

Having this change in place, the handover in libbeat can be done like this:
- timeout send 'CloseSignal' to the client. All message enqueued will still be processed or might error, like normal (no work besides signaling close required in libbeat).
- in libbeat replace the current client with a new go-lumber client instance.

=> During handover, both connections will be active. Once the old connection has finished, processing will be done by active connection only.
=> to not exhaust resources (to many parallel connections), the 'CloseSignal' to the go-lumber should provide some  callback parameter to signal the connection has been actually closed. If two connections are open and the second is to be signaled to close as well, the client must block:
   - quick solution: create channel with 2 slots pre-allocated and pre-filled (max 2 active clients) to act like a semaphore. Before connecting a new client, we read a signal from the channel and once 'CloseSignal' signals done via callback, we write a signal into the channel.

This way, the async client does not really have deal with any go-routine handling and such... it's 'just' plain close and reconnect in same worker.

In order to make some progress here I would propose to:
1. do the sync client only in this PR
2. do not handle TTL in async client (this PR), but print warning in async client if TTL is configured.
3. add required changes to go-lumber
4. create follow up PR introducing TTL support for async client.

Thanks for looking into TTL support. As you can see, there is a reason TTL is not yet in :)

If you're adventurous you can try 3. and 4., but I can help out as well.

nicorevin:
@urso thanks for a detailed explanation! I will think it all over.

urso:
closing in favor of #4343

PR #3455, opened 2017-01-24
Ceph cleanup module

ruflin:
* Add system tests to check for doc fields
* Fix docs / fields which do not correspond
* Use shared client


ruflin:
@amandahla FYI. Still WIP.

PR #3458, opened 2017-01-24
Add beta header to every page

dedemorton:
This PR puts the beta disclaimer on every page of the Heartbeat doc rather than just the overview page. I used a page header instead of the beta[ ] tag because it will be easier to remove when the feature is no longer GA level. Otherwise, I would have had to add the tag manually.

Here's what the page header looks like when rendered in HTML:

disclaimer


monicasarbu:
Great idea @dedemorton

PR #3462, opened 2017-01-25
Add images metricset : Basic informations about the images available …

douaejeouit:
…in the server.


karmi:
Hi @douaejeouit, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/beats/pull/3462.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

ruflin:
@douaejeouit Thanks for the contribution. Please ping me when it is ready for a review.

ruflin:
@douaejeouit General recommendation from my side: If possible you should rebase on top of merging master. It will keep the history clean. But now that you already started merging, probably you will run into problems if you switch now.

douaejeouit:
@ruflin I was wondering of the level of images informations we provide.
Do you think it's better to get low-level informations about each image on the server or to only provide general information about tall the images ( initially, this is what I mplemented in this PR )?

douaejeouit:
@ruflin I was wondering about  the level of images informations we provide.
Do you think it's better to get low-level informations about each image on the server or to only provide general information about tall the images ( initially, this is what I implemented here )?
Okay, Thank you for your recommandations ! ( At worst, i'll create another branch ..)

ruflin:
@douaejeouit I suggest we start with some high level metrics. The we see how it will be used and it should be fairly easy to add additional metrics in the future.

PR #3467, opened 2017-01-25
Metricbeat : new metricset image for docker module 

douaejeouit:
Basic informations about the images available on the server


karmi:
Hi @douaejeouit, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/beats/pull/3467.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

douaejeouit:
@karmi it's done!
Thanks

douaejeouit:
@ruflin your review is welcome! 
thanks

karmi:
@douaejeouit thanks!, but it keeps failing, have you added your Gmail to the Github profile? That should normally suffice to make that green, even when the commit has a different e-mail.

douaejeouit:
@karmi that's weird 😢 I added both of my e-mails ..

karmi:
Hi @douaejeouit, I've reached out via e-mail, ping me if you need assistance!

metricbeat/tests/system/test_docker.py
```diff
@@ -191,6 +191,32 @@ def test_health_fields(self):
...
+        proc.check_kill_and_wait()
+
+        # Ensure no errors or warnings exist in the log.
+        log = self.get_log()
+        self.assertNotRegexpMatches(log.replace("WARN EXPERIMENTAL", ""), "ERR|WARN")
+
+        output = self.read_output_json()
+        evt = output[0]
+
+        evt = self.remove_labels(evt)
```
ruflin: As the image data set probably doesn't have fields, this line is not needed

metricbeat/module/docker/image/data.go
```diff
@@ -0,0 +1,36 @@
...
+			"parent":  image.ParentID,
+		},
+		"created": image.Created,
+		"size": common.MapStr{
+			"regular": image.Size,
+			"virtual": image.VirtualSize,
+		},
+		"repoTags": image.RepoTags,
+	}
+	labels := docker.DeDotLabels(image.Labels)
```
ruflin: Oh, I see it has labels. Ignore my comment below. But it seems the line with the labels somehow breaks the tests. See Travis.

metricbeat/module/docker/image/_meta/fields.yml
```diff
@@ -0,0 +1,9 @@
+- name: image
+  type: group
+  description: >
+    image
+  fields:
+    - name: example
+      type: keyword
+      description: >
+        Example field
```
ruflin: This must be completed with the fields, otherwise the system test will not pass.

douaejeouit: Ok, I'll complete it!

metricbeat/module/docker/image/data.go
douaejeouit: Yes, I'll check it!

douaejeouit:
@ruflin Have you thought about the way we should document lists?

ruflin:
@douaejeouit Which lists are you referring to? labels? My solution currently is to not document it (which is not great). Perhaps we could at least define an empty object: https://www.elastic.co/guide/en/elasticsearch/reference/current/object.html

ruflin:
@douaejeouit I found that we use `dict-type: keyword` here which could become useful: https://github.com/elastic/beats/blob/master/winlogbeat/_meta/fields.yml#L40 But currently I think it does not work with our generation script.

douaejeouit:
@ruflin , yes I am referring to labels as well as tags.
I did the same, I have not documented neither labels nor tags.
Defining an empty object sounds better!

ruflin:
See https://github.com/elastic/beats/pull/3515

metricbeat/module/docker/image/_meta/data.json
```diff
@@ -0,0 +1,19 @@
+{
```
ruflin: Needs update but I can do that after merging.

ruflin:
@douaejeouit Thanks for the contribution.

douaejeouit:
With the greatest of pleasure @ruflin

PR #3468, opened 2017-01-25
Auto generate imports for protocols in Packetbeat

ruflin:
This will allow to use the protocol generator and directly create the correct imports.


urso:
2 notes for future PRs:

- I'd like to have it renamed to `packetbeat/defaults/defaults.go`, instead of `packetbeat/include/list.go`. And do the same for all beats in a consistent manner (especially libbeat)
- I'd prefer to have this code generation being build using `go generate`. As `go generate` can take any command, one can use `go run path/to/tool/main.go`. This way it better integrates with go developer tools and will also work on windows.

ruflin:
For the defaults: So you are thinkg there is a `default/protocols.go` and `default/output.go`etc?

PR #3476, opened 2017-01-27
Update outputconfig.asciidoc

markwalkom:
Clarified how we handle things when multiple hosts are listed.


ruflin:
@markwalkom thx

PR #3478, opened 2017-01-27
Monitoring enhancements

urso:
- remove NewInt/Float/String from registry:
  replace NewInt/Float/String and (*Registry).NewX by one NewX accepting the
  target registry by default. if nil is choosen, we default to the Default
  registry
- Introduce 'Report' flag to metrics to mark some basic metrics
- Add new adapter filters: ReportIf and ReportNames
- apply variable options on registry or variable node. A variable 'inherits'
  the options from it's registry on Add => using `monitoring.Report` on registry
  will set `report` on all variables added to the sub-registry
- move flattened snapshot support to monitoring/snapshot.go
- Replace global http handler for httpprof with multiplexer configuring a
  only selected set of HTTP handlers (do not expose 'unknown' handlers from
  third party packages by accident)
- custom /debug/vars http handler support expvar and libbeat/monitoring (output
  fully compatible to old expvar output)

### DIFF
--- a/libbeat/monitoring/adapter/filters.go
+++ b/libbeat/monitoring/adapter/filters.go
@@ -3,6 +3,7 @@ package adapter
 import (
 	"strings"
 
+	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/monitoring"
 )
 
@@ -26,6 +27,7 @@ type varFilter func(state) state
 type state struct {
 	kind   kind
 	action action
+	mode   monitoring.Mode
 	reg    *monitoring.Registry
 	name   string
 	metric interface{}
@@ -90,16 +92,27 @@ func WhitelistIf(pred func(string) bool) MetricFilter {
 	return ApplyIf(pred, Accept)
 }
 
+// NameIn checks a metrics name matching any of the set names
+func NameIn(names ...string) func(string) bool {
+	return common.MakeStringSet(names...).Has
+}
+
 // Whitelist sets a list of metric names to be accepted.
 func Whitelist(names ...string) MetricFilter {
-	return WhitelistIf(func(name string) bool {
-		for _, n := range names {
-			if name == n {
-				return true
-			}
-		}
-		return false
-	})
+	return WhitelistIf(NameIn(names...))
+}
+
+// ReportIf sets variable report mode for all metrics satisfying the predicate.
+func ReportIf(pred func(string) bool) MetricFilter {
+	return ApplyIf(pred, withVarFilter(func(st state) state {
+		st.mode = monitoring.Reported
+		return st
+	}))
+}
+
+// ReportNames enables reporting for all metrics matching any of the given names.
+func ReportNames(names ...string) MetricFilter {
+	return ReportIf(NameIn(names...))
 }
 
 // ModifyName changes a metric its name using the provided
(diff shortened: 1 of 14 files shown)


libbeat/monitoring/adapter/filters.go
```diff
@@ -90,18 +91,43 @@ func WhitelistIf(pred func(string) bool) MetricFilter {
...
 		return false
+	}
+}
+
+// Whitelist sets a list of metric names to be accepted.
+func Whitelist(names ...string) MetricFilter {
+	return WhitelistIf(NameIn(names...))
+}
+
+// Mode sets the monitoring mode (Basic, Enhanced, Full) for
```
ruflin: We had a similar concept in the beginning of metricbeat on which metrics should be sent to elasticsearch for a metricset. We dropped the idea rather soon because the disussion what should be in basic, enhanced and full got too complex as the answer was different for every single person. I would suggest we follow here more the approach of going with whitelist / backlist filtering instead.

libbeat/monitoring/metrics.go
```diff
@@ -12,30 +14,54 @@ type makeExpvar func() string
...
 
-// NewFloat registers a new global floating point metric.
-func NewFloat(name string) *Float {
-	return Default.NewFloat(name)
+// NewFloat creates and registers a new float variable.
+//
+// Note: If the registry is configured to publish variables to expvar, the
+// variable will be available via expvars package as well, but can not be removed
+// anymore.
+func NewFloat(r *Registry, name string, mode Mode) *Float {
```
ruflin: In addition to my above comment, not having `mode` removes the complexity during implementation to decide where it should belong. All metrics have the same importance, it is up to the consumer to decide which ones he wants.

libbeat/service/service.go
```diff
@@ -72,11 +74,49 @@ func BeforeRun() {
 	if *httpprof != "" {
 		go func() {
 			logp.Info("start pprof endpoint")
-			logp.Info("finished pprof endpoint: %v", http.ListenAndServe(*httpprof, nil))
+			mux := http.NewServeMux()
+
+			// register pprof
+			mux.HandleFunc("/debug/pprof/", pprofNet.Index)
```
ruflin: Do we need to take over all the handlers here? Is there a way to only replace the one we modify? The part I worry is if new handlers are added in golang, we have to keep track of it manually.

libbeat/monitoring/adapter/filters.go
urso: as discussed, only 2 modes left: Report and Full.

libbeat/monitoring/metrics.go
urso: the `mode` parameter has been removed. One can pass optional `Option` parameters to enable/disable expvar reporting and/or setting the mode (report or full).

libbeat/service/service.go
urso: It's kind of a trade of here. By default httpprof just uses the global default handler. Any package you import might decide to register a handler. E.g. pprof, expvar and go-metrics all registers into the default handler (not sure about other packages). Idea is to take control on what's getting exposed via httpprof (plus I need to overwrite `/debug/vars` to not use `expvar`, but `libbeat/monitoring`).

Unfortunately the `net/http/pprof` package does not make the handlers available via a `http.ServerMux`, but low-level handlers callbacks only :(

urso: Updated by proxying `/debug/pprof` to the `http.DefaultServeMux`.

PR #3487, opened 2017-01-28
Add SSL flags to import_dashboards CLI help

dedemorton:
This PR updates the doc with the flags that were added in https://github.com/elastic/beats/pull/3163 and https://github.com/elastic/beats/pull/3139.


monicasarbu:
Thank you @dedemorton

PR #3488, opened 2017-01-28
Gather module configuration fragments

tsg:
Added configuration samples for the modules and added rules to gather
them in the short and full versions of the configs. This leaves all
modules commented out, so there isn't any change in the default
behaviour, just the modules added to the configuration files.

Part of #3159.


tsg:
@ruflin Added a commit here to remove the `_meta/beat.yml` and `_meta/beat.full.yml` files from git in both Filebeat and Metricbeat. These are intermediary files, and we already stored the generated result.

filebeat/_meta/common.full.p1.yml
```diff
@@ -0,0 +1,9 @@
+##################$$$###### Filebeat Configuration ############################
```
ruflin: Are the $$$ in the line here accidential?

metricbeat/.gitignore
```diff
@@ -1,5 +1,7 @@
 build
 _meta/kibana
+_meta/beat.yml
```
ruflin: +1 on this change.

filebeat/Makefile
```diff
@@ -34,9 +34,19 @@ modules:
...
 
+# Collects all module configs
+.PHONY: configs
+configs: python-env
+	cat ${ES_BEATS}/filebeat/_meta/common.p1.yml > _meta/beat.yml
+	. ${PYTHON_ENV}/bin/activate; python ${ES_BEATS}/metricbeat/scripts/config_collector.py --beat ${BEAT_NAME} $(PWD) >> _meta/beat.yml
+	cat ${ES_BEATS}/filebeat/_meta/common.p2.yml >> _meta/beat.yml
+	cat ${ES_BEATS}/filebeat/_meta/common.full.p1.yml > _meta/beat.full.yml
+	. ${PYTHON_ENV}/bin/activate; python ${ES_BEATS}/metricbeat/scripts/config_collector.py --beat ${BEAT_NAME} --full $(PWD) >> _meta/beat.full.yml
+	cat ${ES_BEATS}/filebeat/_meta/common.full.p2.yml >> _meta/beat.full.yml
```
ruflin: p1 = part 1, p2 = part 2?

tsg: Yeah, but I'm open to naming suggestions  :)

filebeat/_meta/common.full.p1.yml
tsg: Lol, yes. It's the $$$ price tag for FBM.

filebeat/Makefile
ruflin: 👍  on the naming. Was just curious as I think I will potentially use the same in metricbeat ...

PR #3490, opened 2017-01-30
Update generator to reference to the docs

ruflin:
* All docs are now in one place which simplifies handling
* Specific branch to start your beat on was defined to prevent people using master

This change will prevent people creating a new beat to rely on master and have a stable branch for development. This also allows us to modify the generator without having all docs out of sync directly.

The blog post "Build your own beat" should be updated with an "update note" about checking out branch 5.1: https://www.elastic.co/blog/build-your-own-beat

Further changes:

* Add golang version to version.asciidoc file
* Remove copied version.asciidoc files and link global file

(cherry picked from commit 668b0d7ee86fba163932c382416234185e102e48)


dedemorton:
LGTM

Issue #3489, opened 2017-01-30
Memory leak when activating docker metrics


tobiaskraft:
When activating the docker metrics in metricbeat the Memory and also the CPU usage is continuously increasing. After two days the metricbeat process has consumed 20 % of the available memory (12 GB).
After deactivating the docker metrics the problem disappeared.

For more information and a diagram on the memory usage see also:
https://discuss.elastic.co/t/cpu-memory-consumption-after-upgrading-to-version-5-1-2/72735

- Version: Metricbeat 5.1.2
- Operating System: Ubuntu 16.04.1 LTS
- Steps to Reproduce: Activate docker metrics in the file metricbeat.yml

aanc:
Same issue here on a Fedora 24 server (Metricbeat 5.1.2 too). Memory consumption after 3 days is 70% (of 16GB), and CPU usage 10%.
Docker module is activated for 10 containers, every 10 sec.

![image](https://cloud.githubusercontent.com/assets/5970344/22416957/d5a04cf6-e6d1-11e6-87cd-d2480b02d9a7.png)
![image](https://cloud.githubusercontent.com/assets/5970344/22416981/ff6486f6-e6d1-11e6-87b1-39f89a22ac3e.png)



I'm deactivating docker module.

ruflin:
@aanc @tobiaskraft Thanks for the report. We will investigate this further.

ruflin:
I potentially found the problem here: https://github.com/elastic/beats/pull/3492

ruflin:
@aanc @tobiaskraft Could one of you test a nightly build to see if this fixes the bug? https://beats-nightlies.s3.amazonaws.com/index.html?prefix=metricbeat/ You probably need to wait 1-2 more hours until a build is available with it as the PR just got merged.

tobiaskraft:
Hi Nicolas,

the build is now installed on our system. I will provide feedback in 2 or 3 days on the memory and CPU consumption.

ruflin:
@tobiaskraft Thanks for testing. I'm hoping this gets it fixed.

aanc:
It's better now on my side.

![image](https://cloud.githubusercontent.com/assets/5970344/22545559/83dd8ca2-e939-11e6-8496-47e75e9cf323.png)

Maybe we should wait a few more days to be sure, but seems promising. 
Thanks !

ruflin:
[...]

PR #3492, opened 2017-01-30
Fix leaking go routine in docker stats fetching

ruflin:
See https://github.com/elastic/beats/issues/3489

### DIFF
--- a/metricbeat/module/docker/docker.go
+++ b/metricbeat/module/docker/docker.go
@@ -66,24 +66,26 @@ func FetchStats(client *docker.Client) ([]Stat, error) {
 
 	var wg sync.WaitGroup
 
-	containersList := []Stat{}
+	containersList := make([]Stat, 0, len(containers))
 	queue := make(chan Stat, 1)
 	wg.Add(len(containers))
 
 	for _, container := range containers {
 		go func(container docker.APIContainers) {
+			defer wg.Done()
 			queue <- exportContainerStats(client, &container)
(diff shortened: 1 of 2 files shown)


metricbeat/module/docker/docker.go
```diff
@@ -84,6 +84,7 @@ func FetchStats(client *docker.Client) ([]Stat, error) {
 	}()
 
 	wg.Wait()
+	close(queue)
```
andrewkroh: Consider swapping things around so that you can call `wg.Done()` from the worker goroutines. This more closely aligns with the [Fan-in, Fan-out](https://blog.golang.org/pipelines#TOC_4.) examples here.

```
 	for _, container := range containers {
 		go func(container docker.APIContainers) {
			defer wg.Done()
 			queue <- exportContainerStats(client, &container)
 		}(container)
 	}

 	go func() {
  	  	wg.Wait()
  	  	close(queue)
  	}()
  
  	// This will break after the queue and been drained and queue is closed.
 	for container := range queue {
 		containersList = append(containersList, container)
 	}

```

andrewkroh: And up on line 69, consider changing it to `containersList = make([]Stat, 0, len(containers))` since we know ahead of time how big the slice needs to be. This will reduce the number of allocations.

ruflin: This definitively makes the code much cleaner.

Issue #3329, opened 2017-01-10
Beats Doc Updates for 5.2


dedemorton:
This is a meta issue to track Beats documentation work for 5.2. 

If you have trouble clicking an item in the checklist to mark it as complete, open this comment in edit mode and change the source to `[x]`. Seems to be a bug in ticking the check boxes in this issue.

**5.2:**
- [x] [DeDe] Add Heartbeat docs. See https://github.com/elastic/beats/pull/3247
- [x] [DeDe] Test Getting Started section of Heartbeat docs.
- [x]  Add/edit docs for `system.socket` in Metricbeat. Experimental feature only on Linux.
- [x]  Add/edit docs for Prometheus module in Metricbeat. See https://github.com/elastic/beats/pull/3520
- [x] Edit HAProxy docs. See https://github.com/elastic/beats/pull/3216.
- [x] Document new config options for processors. See https://github.com/elastic/beats/issues/3443

NOTE: Moved wish list and future items to 5.3 meta issue here: https://github.com/elastic/beats/issues/3539

PR #3495, opened 2017-01-31
Add edits for load and socket metricset docs

dedemorton:
Adds a few edits for consistency and clarity. 

These edits are tracked in the doc meta issue #3329.


andrewkroh:
The changes LGTM, but the build failed with:

```
metricbeat/docs/fields.asciidoc: needs update
make: *** [check] Error 1
```

I think you need to run `make collect`.

ruflin:
Best directly run `make update` that will also cover `make collect`.

dedemorton:
@andrewkroh or @ruflin Should be ready now. I did run `make collect` but it sounds like `make update` is the secret sauce. :-P

PR #3496, opened 2017-01-31
Fix environments tests in 5.0 by updating dependency

ruflin:
Use stable versions which will not disappear.


ruflin:
Even though this build will not go green on Jenkins as jenkins_ci is missing, I would suggest to still merge it as at least it does check the builds with travis in the future.

PR #3506, opened 2017-01-31
Make -setup load the Beat dashboards

tsg:
With this change, `filebeat -setup` loads the sample Kibana dashboards,
being the rough equivalent of running: `scripts/import_dashboards && filebeat`.
This works for all Beats, not only Filebeat. The `import_dashboards` program
is kept for now, to avoid braking compatibility. Code wise, the two share
most of the code, which was separated in a sub-package.

All flags available for the `import_dashboard` are available as go-ucfg options.
For example, to load the dashboards from a directory you can do:

    filebeat -e -setup -E "dashboards.dir=_meta/kibana"

Or do load the snapshot version of the dashboards:

    filebeat -e -setup -E "dashboards.snapshot=true"

This also changes the pipeline loading to happen automatically regardless of
whether `-setup` was used or not.

Part of #3159.

TODOs:

* [x] Add tests for the dashboard sub-package
* [x] Document the -setup flag and dashboard settings
* [x] Changelog + deprecation notice for the import_dashboard program (?)


tsg:
@dedemorton I'd appreciate your review on the docs part.

filebeat/filebeat.full.yml
```diff
@@ -897,6 +897,51 @@ output.elasticsearch:
...
+# The directory from where to read the dashboards. It is used instead of the URL
+# when it has a value.
+#dashboards.directory:
+
+# The file archive (zip file) from where to read the dashboards. It is used instead
+# of the URL when it has a value.
+#dashboards.file:
+
+# If this option is enabled, the snapshot URL is used instead of the default URL.
+#dashboard.snapshot: false
```
ruflin: Nice.

libbeat/beat/beat.go
```diff
@@ -100,10 +102,12 @@ type BeatConfig struct {
 	Logging    logp.Logging              `config:"logging"`
 	Processors processors.PluginConfig   `config:"processors"`
 	Path       paths.Path                `config:"path"`
+	Dashboards *common.Config            `config:"dashboards"`
 }
 
 var (
 	printVersion = flag.Bool("version", false, "Print the version and exit")
+	setup        = flag.Bool("setup", false, "Load the sample Kibana dashboards")
```
ruflin: Load sampel Kibana dashboards, index pattern and index template?

libbeat/dashboards/dashboards/importer.go
```diff
@@ -0,0 +1,535 @@
+package dashboards
```
ruflin: Why is it `dashboards/dashboards/...`?

metricbeat/_meta/beat.full.yml
```diff
@@ -0,0 +1,294 @@
+########################## Metricbeat Configuration ###########################
```
ruflin: Wasn't this under .gitignore? Do we need it again?

tsg:
> Question for the -setup flag: If I run now ./filebeat -setup will filebeat stop after the setup phase or will it continue running? And will it load only dashboards or also pipelines, index patterns etc?

It will continue in case of success and stop in case of errors. Since we still have the `import_dashboards` script that does only that, I thought it would be interesting to experiment with letting it run in case of success. So far I like it like this, but it's debatable.

At the moment, `-setup` loads strictly the dashboards. Everything else (templates, pipelines, etc.) is happening automatically always, since they are lightweight operations and the Beats mallfunction if they are not loaded. The dashboards, on the other hand, are relatively heavy to load and can also be loaded after starting the Beat, without any implication regarding the correctness of the data.

libbeat/dashboards/dashboards/importer.go
tsg: First `dashboards` is the current `import_dashboards` main package. The second is where I gathered most of the code. After we remove `import_dashboards` we can move the package one level up.

metricbeat/_meta/beat.full.yml
tsg: Hmm, weird, i'll check.

libbeat/beat/beat.go
tsg: At the moment it only loads the dashboards, see the other answer.

filebeat/filebeat.full.yml
tsg: That's a feature we already have in `import_dashboards" :)

ruflin:
@tsg Thanks for the details. My preferred option would be as following:

* -setup loads everything (dashboards, pipelines, templates, ...) and stops afterwards as this is run only on one machine / separate machine. This would mostly replace the dashboard script.
* Without setup, template, pipelines are loaded by default, but can be disabled.

But if -setup only loads the dashboards, I can see that you're behaviour is preferrable as we also have "only" the script.

tsg:
We could have both `-setup` and `-setup-only` as a compromise? The `-setup` is great for demos and getting started, since it removes one step. I didn't add `-setup-only` yet because we still have `import_dashboards`.

ruflin:
For demos what can be done now is `./filebeat -E dashboards.enabled:true` which loads everything so not even `-setup` flag has to be used. That is the great thing about having it in the config file. Having this, `-setup-only` could be `-setup`.

tsg:
We're splitting hairs now, but I think `-setup` would work better in demos & getting started than `-E dashboards.enabled=true`. We could have `-E dashboards.stop_after_loading=true` or something like that instead of `-setup-only`. In other words, flip when we need the -E. 

Or maybe I'm wrong and we could skip both `-setup` and `-setup-only` and only rely on `-E` flags, this way we teach the user early on how to do that.

heartbeat/heartbeat.full.yml
```diff
@@ -745,6 +745,51 @@ output.elasticsearch:
...
+
+# The name of the Kibana index to use for setting the configuration. Default is ".kibana"
+#dashboards.kibana_index: .kibana
+
+# The Elasticsearch index name. This overwrites the index name defined in the
+# dashboards and index pattern. Example: testbeat-*
+#dashboards.index:
+
+# If enabled, load only the dashboards and not the index pattern.
+#dashboards.only_dashboards: false
```
monicasarbu: I think it doesn't make sense to migrate these two options (`only_dashboards` and `only_index`) from `import_dashboards`, and leave `-setup` to import both dashboards and the index pattern.

metricbeat/_meta/beat.full.yml
tsg: There was a typo in .gitignore, I deleted it again.

monicasarbu:
I think what we should improve here on longer term is to remove the necessity of `-setup` to load the dashboards, and have each Beat check if the dashboards version X are loaded and load them only when needed. And maybe have a "dashboards.overwrite" boolean option. This will also solve the first experience of the user as only one Beat will be able to load the dashboard, when starting more Beats at the same time.

Another improvement that we can do is to load only the Kibana dashboards for which the modules are enabled, but this requires us to split the dashboards per module. The index pattern will be loaded for all modules.

For now, I would say to keep `-setup` import only the dashboards, and try to remove it with the time.

heartbeat/heartbeat.full.yml
tsg: I removed the two options from the configuration files and docs, but they still exist in the code so that they keep working for `import_dashboards`. We can remove them from the code as well when we remove `import_dashboards`.

dedemorton:
@tsg sorry I didn't get to the review today (didn't see your request until late in the day). I'll look in the morning. But I have a couple of initial questions:
1) Since this is Beta level, I'm assuming we'll leave the steps for loading the dashboards as they are in the getting started guides for now, right?  At what point will the new way become the preferred way to install dashboards? 
2) What about the use case where community Beats developers need to import dashboards (documented [here](https://www.elastic.co/guide/en/beats/libbeat/current/import-dashboards.html)). Will we continue to provide the import_dashboards command for that specific use case?

tsg:
> Since this is Beta level, I'm assuming we'll leave the steps for loading the dashboards as they are in the getting started guides for now, right? At what point will the new way become the preferred way to install dashboards?

Correct, we don't have to change the getting started guide yet. I think the new way will become preferred in 6.0, but until then more changes might happen.

> What about the use case where community Beats developers need to import dashboards (documented here). Will we continue to provide the import_dashboards command for that specific use case?

For the moment we keep the `import_dashboards` script, so we don't have to adjust the documentation. It'll also probably change with 6.0.

PR #3507, opened 2017-02-01
Remove step to set default index pattern

dedemorton:
Removed step about setting the default index pattern (no longer required) and changed screens to reflect the new flow.


ruflin:
jenkins, retest it

PR #3509, opened 2017-02-01
CEPH module cleanup, phase 2

ruflin:
* Split up health metricset into cluster and monitor metricset
* Fix http json request header which I broke with my last PR
* Add header support to http helper
* Switch default host to local host
* Add tests for the new metricsets


ruflin:
@amandahla Could you have a look at this one and let me know if this makes sense from your perspective? I split up health into `cluster` and `monitor` metricset.

amandahla:
I think the problem it's similar with 'df' where we can two metricsets called 'dfcluster' and 'dfpool'. Maybe here we could have 'healthcluster' and 'healthmon'.
Or we can create a pattern using the CEPH basic components: cluster, mon, osd(pool and pg) and mds. Then we have: healthcluster, healthmon, dfcluster and dfpool.
What do you think?

ruflin:
I'm tempted to say it should be `health.cluster` and `health.pool` but I don't see how this will work with golang package names.

In general I prefer not to use abbreviations. What about:

* cluster_health
* pool_health
* monitor_disk
* pool_disk

The reason I turned around the names is that assuming we would have auto complete for fields in Kibana and someone starts to type `ceph.poo*` he would get the suggestion for health and disk. 

Side note: I need to check what happens if we call the package `cluster_health` but use `cluster.health` as name. We would introduce an additional namespace through the back door ...

ruflin:
@amandahla I updated the PR to use `monitor_health` and `cluster_health`. WDYT?

amandahla:
@ruflin Great!

PR #3511, opened 2017-02-01
Apply close_timeout also when output is blocked

ruflin:
Currently `close_timeout` does not apply in case the output is blocked. This PR changes the behavior of `close_timeout` to also close a file handler when the output is blocked.

It is important to note, that this closes the file handler but NOT the harvester. This is important as the closing of the harvester requires a state update to set `state.Finished=true`. If this would not happen and the harvester is closed, processing would not continue when the output becomes available again.

Previously the internal state of a harvester was updated when the event was created. This could lead to the issue that in case an event was not sent but the state update went through, that an event would be missing. This is now prevent by overwriting the internal state only when the event was successfully sent.

The done channels from prospector and harvester are renamed to be more obvious which one belongs to what: h.done -> h.prospectorDone, h.harvestDone -> h.done. As the harvester channel is close with the `stop` method in all cases `h.done` is sufficient in most places.

This PR does not solve the problem related to reloading and stopping a harvester mentioned in elastic#3511 (comment) related to reloading. This will be done in a follow up PR.

### DIFF
--- a/filebeat/harvester/harvester.go
+++ b/filebeat/harvester/harvester.go
@@ -14,6 +14,7 @@ package harvester
 import (
 	"errors"
 	"fmt"
+	"sync"
 
 	"github.com/elastic/beats/filebeat/config"
 	"github.com/elastic/beats/filebeat/harvester/encoding"
@@ -39,6 +40,8 @@ type Harvester struct {
 	fileReader      *LogFile
 	encodingFactory encoding.EncodingFactory
 	encoding        encoding.Encoding
+	prospectorDone  chan struct{}
+	once            sync.Once
 	done            chan struct{}
 }
 
@@ -53,7 +56,8 @@ func NewHarvester(
 		config:         defaultConfig,
 		state:          state,
 		prospectorChan: prospectorChan,
-		done:           done,
+		prospectorDone: done,
+		done:           make(chan struct{}),
 	}
 
 	if err := cfg.Unpack(&h.config); err != nil {
--- a/filebeat/harvester/log.go
+++ b/filebeat/harvester/log.go
@@ -58,13 +58,14 @@ func (h *Harvester) Harvest(r reader.Reader) {
 	defer h.close()
 
 	// Channel to stop internal harvester routines
-	harvestDone := make(chan struct{})
-	defer close(harvestDone)
+	defer h.stop()
 
 	// Closes reader after timeout or when done channel is closed
 	// This routine is also responsible to properly stop the reader
 	go func() {
-		var closeTimeout <-chan time.Time
+
+		closeTimeout := make(<-chan time.Time)
+		// starts close_timeout timer
 		if h.config.CloseTimeout > 0 {
 			closeTimeout = time.After(h.config.CloseTimeout)
 		}
@@ -72,12 +73,14 @@ func (h *Harvester) Harvest(r reader.Reader) {
 		select {
 		// Applies when timeout is reached
 		case <-closeTimeout:
-			logp.Info("Closing harvester because close_timeout was reached: %s", h.state.Source)
+			logp.Info("Closing harvester because close_timeout was reached.")
 		// Required for shutdown when hanging inside reader
-		case <-h.done:
+		case <-h.prospectorDone:
 		// Required when reader loop returns and reader finished
-		case <-harvestDone:
+		case <-h.done:
 		}
+
+		h.stop()
 		h.fileReader.Close()
 	}()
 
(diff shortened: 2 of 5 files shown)


ruflin:
In a call with Steffen we came to the following conclusions:

* We will use `close_timeout` to also be used for closing the channel. We don't want to introduce an additional config options
* `close_inactive` could also be used but would have the consequence that we have a default config which can potentially lead to data loss.
* In general we hope to have in the long term a solution which is based on the output / pipeline part to not have to handle it in the harvester
* It is important that the harvesters only close the file but stay open, as otherwise we would have unfinished states.
* In the code already now is kind of a bug that `h.done` could be close and then checked again in the last state update. This had no affect so far because we were always shutting down filebeat. But now with reloading this could mean that some files are not continued reading. This must be fixed.
* If filebeat is stopped when the output is blocked and the harvester is in the closing phase, it can lead to duplicated data on restart.

ruflin:
This cause a race condition on windows. Not sure yet where it is exactly happening:

```
==================
WARNING: DATA RACE
Read at 0x00c0421102b8 by goroutine 40:
  runtime.convT2E()
      c:/go1.7.4/src/runtime/iface.go:155 +0x0
  github.com/elastic/beats/filebeat/harvester.(*Harvester).Harvest.func1()
      github.com/elastic/beats/filebeat/harvester/_obj/log.go:83 +0x1a9

Previous write at 0x00c0421102b8 by goroutine 39:
  github.com/elastic/beats/filebeat/harvester.(*Harvester).Harvest()
      github.com/elastic/beats/filebeat/harvester/_obj/log.go:177 +0x6b0
  github.com/elastic/beats/filebeat/prospector.(*Prospector).startHarvester.func1()
      github.com/elastic/beats/filebeat/prospector/_obj/prospector.go:300 +0xad

Goroutine 40 (running) created at:
  github.com/elastic/beats/filebeat/harvester.(*Harvester).Harvest()
      github.com/elastic/beats/filebeat/harvester/_obj/log.go:95 +0x1eb
  github.com/elastic/beats/filebeat/prospector.(*Prospector).startHarvester.func1()
      github.com/elastic/beats/filebeat/prospector/_obj/prospector.go:300 +0xad

Goroutine 39 (running) created at:
  github.com/elastic/beats/filebeat/prospector.(*Prospector).startHarvester()
      github.com/elastic/beats/filebeat/prospector/_obj/prospector.go:301 +0x485
  github.com/elastic/beats/filebeat/prospector.(*ProspectorLog).scan()
      github.com/elastic/beats/filebeat/prospector/_obj/prospector_log.go:281 +0x83d
  github.com/elastic/beats/filebeat/prospector.(*ProspectorLog).Run()
      github.com/elastic/beats/filebeat/prospector/_obj/prospector_log.go:93 +0x11e
  github.com/elastic/beats/filebeat/prospector.(*Prospector).Run()
      github.com/elastic/beats/filebeat/prospector/_obj/prospector.go:177 +0xfe
  github.com/elastic/beats/filebeat/prospector.(*Prospector).Start.func1()
      github.com/elastic/beats/filebeat/prospector/_obj/prospector.go:145 +0x8a
==================
```

filebeat/harvester/harvester.go
```diff
@@ -39,6 +40,8 @@ type Harvester struct {
 	fileReader      *LogFile
 	encodingFactory encoding.EncodingFactory
 	encoding        encoding.Encoding
+	once            sync.Once
```
urso: it should be clear `once` is used to protected `prospectorDone` from multiple go-routines potentially doing `close(prospectorDone)`. Uhm.. is it used for `prospectorDone`?

Just seeing it protects the `done` channel. sync tools and items being protected should always stand together.

filebeat/harvester/log.go
```diff
@@ -58,26 +58,29 @@ func (h *Harvester) Harvest(r reader.Reader) {
...
+		// starts close_timeout timer
 		if h.config.CloseTimeout > 0 {
 			closeTimeout = time.After(h.config.CloseTimeout)
 		}
 
 		select {
 		// Applies when timeout is reached
 		case <-closeTimeout:
-			logp.Info("Closing harvester because close_timeout was reached: %s", h.state.Source)
+			logp.Info("Closing harvester because close_timeout was reached.")
```
urso: I see how access to h.state.Source can race. Consider printing the last state in defer statement when Harvest returns (but after workers have been finished).

filebeat/harvester/harvester.go
ruflin: yes, once protects done from only closing done once. Moved it one line down.

filebeat/harvester/log.go
ruflin: There is already a looking message in the closing part where the state.Source is inside. So I don't think it is necessary here.

ruflin:
@andrewkroh prospectorDone is the `p.done` channel in the prospector and is closed on `prospector.go:200`. I'm actually hoping to get rid of the prospectorDone channel in the long run.

ruflin:
@andrewkroh @urso new version pushed.

PR #3515, opened 2017-02-02
Add support for long and keyword in dict-type

ruflin:
As keyword is the default currently not mapping was added. For the human reader now a mapping is predefined.

For long dict-types a dynamic mapping is added as is done for text

### DIFF
--- a/heartbeat/heartbeat.template.json
+++ b/heartbeat/heartbeat.template.json
@@ -54,6 +54,9 @@
             }
           }
         },
+        "fields": {
+          "properties": {}
+        },
         "host": {
           "ignore_above": 1024,
           "type": "keyword"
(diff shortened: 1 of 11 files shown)


heartbeat/heartbeat.template.json
```diff
@@ -54,6 +54,7 @@
             }
           }
         },
+        "fields": {},
```
tsg: Maybe:

```
"fields": {
     "properties": {}
}
```

looks a bit more clear.

libbeat/scripts/generate_template.py
```diff
@@ -279,6 +279,27 @@ def fill_field_properties(args, field, defaults, path):
...
+                    "mapping": {
+                        "type": "long",
+                    },
+                    "match_mapping_type": "long",
+                    "path_match": name + ".*"
+                }
+            })
+
+        if field.get("dict-type") == "keyword":
+            properties[field["name"]] = {}
```
tsg: This should also be done for "long" and "text", then.

ruflin: Ah, yes. It applies to all dict-type entries.

PR #3516, opened 2017-02-02
Use the Beat version in the Ingest Node pipeline

tsg:
This adds the Beat version to the pipeline ID, which means that if
we change the pipeline between versions, the new version will be used
automatically. It also means that one can run different versions of the
same Beat and the pipelines won't override each other. The pipelines
are loaded automatically on the Beat start.

Part of #3159.

### DIFF
--- a/filebeat/fileset/fileset_test.go
+++ b/filebeat/fileset/fileset_test.go
@@ -149,7 +149,7 @@ func TestResolveVariable(t *testing.T) {
 
 func TestGetProspectorConfigNginx(t *testing.T) {
 	fs := getModuleForTesting(t, "nginx", "access")
-	assert.NoError(t, fs.Read())
+	assert.NoError(t, fs.Read("5.2.0"))
 
 	cfg, err := fs.getProspectorConfig()
 	assert.NoError(t, err)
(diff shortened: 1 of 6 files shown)


filebeat/fileset/fileset_test.go
```diff
@@ -183,17 +183,17 @@ func TestGetProspectorConfigNginxOverrides(t *testing.T) {
 	assert.True(t, cfg.HasField("pipeline"))
 	pipelineID, err := cfg.String("pipeline", -1)
 	assert.NoError(t, err)
-	assert.Equal(t, "nginx-access-with_plugins", pipelineID)
+	assert.Equal(t, "nginx-access-with_plugins-5.2.0", pipelineID)
```
ruflin: I would call the pipeline `filebeat-5.2.0-nginx-access-with_plugins`. This brings the following benefits:

* No conflict with potential pipelines in the future from other beats
* Assuming some can query for pipelines, to get / remove all filebeat pipelines `filebeat-*` can be used, to remove all `5.2` pipelines because everyting is using 5.3, `filebeat-5.2.*` can be used.

PR #3517, opened 2017-02-02
Add support to configure monitoring output

urso:
- Add support to configure `monitoring.elasticsearch`. If `monitoring.elasticsearch` is missing, configuration from `output.elasticsearch` will be used. If both are present, `monitoring.elasticsearch` re-uses missing options from `outputs.elasticsearch`. E.g. this way timeouts and user/password can be configured in `monitoring.elasticsearch`, while re-using `hosts` and other connection settings.
- send and publisher loop disconnected from event publisher to publish metric snapshots collected via `libbeat/monitoring`
- Introduce common.BeatInfo structure being initialized on startup with common beat meta-data to be used by outputs and monitoring reporter:
    - beat (e.g. filebeat, metricbeat, ...)
    - version
    - name (configure name, otherwise hostname)
    - hostname
    - uuid

will do doc update in followup PR, to not block review for this PR for too long.

Note:
would be nice to have `beat.Info` or `beat.Meta` instead of
`publisher.BeatInfo`, but due to potential circular imports I have had to put
the common structure into the publisher package (already contains shipper
config/meta).


libbeat/beat/beat.go
```diff
@@ -87,9 +91,7 @@ type Creator func(*Beat, *common.Config) (Beater, error)
 // Beat contains the basic beat data and the publisher client used to publish
 // events.
 type Beat struct {
-	Name      string              // Beat name.
-	Version   string              // Beat version number. Defaults to the libbeat version when an implementation does not set a version.
-	UUID      uuid.UUID           // ID assigned to a Beat instance.
+	Info      publisher.BeatInfo  // beat metadata.
```
ruflin: I can also make use of this in https://github.com/elastic/beats/pull/3527 Perhaps we should put it into a separate package to prevent circular imports.

libbeat/monitoring/monitoring.go
```diff
@@ -4,6 +4,7 @@ import "errors"
 
 type Mode uint8
 
+//go:generate stringer -type=Mode
```
ruflin: We should make sure that `make update` also executes the generate commands.

libbeat/monitoring/report/report.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/publisher"
+)
+
+type config struct {
+	// allow for maximum one reporter being configured
+	Reporter common.ConfigNamespace `config:",inline"`
+}
+
+type Reporter interface {
```
ruflin: Seems like we start to have this `Runner` interface in lots of different places. We should probably start to use a shared interface.

libbeat/publisher/publish.go
```diff
@@ -87,6 +87,14 @@ type BeatPublisher struct {
 	numClients uint32
 }
 
+type BeatInfo struct {
```
ruflin: Should we put it under `common` ?

libbeat/monitoring/monitoring.go
urso: oh, it's a leftover from debugging. There are 2 problems with stringer though:
1. you either need to have it installed or we have to vendor it and use `go run`
2. package loading in stringer requires you to have done `go install` on dependent packages (unless you have build a binary already). Otherwise `go generate` will fail (See [this ticket](https://github.com/golang/go/issues/10249)).

libbeat/monitoring/report/report.go
urso: nice thing about interfaces is, if they are assignable (same definition or sub-type relation), they can be used interchangeable. I'm aware of us having a mix of this interface.

Besides similar definition, giving it a name fitting the packages usage is sometimes nice to have. Plus, the Reporter interface might change. I'm not too much a fan of the 'Start(), Stop()` interfaces, as this clearly mandates/leaks usage of go-routines and very often channels for communication... I think in 'Reporter' we can get rid of 'Start' :)

libbeat/publisher/publish.go
urso: I'd like it to be `beat.Info`, not `common.BeatInfo`. With beats being kind of a framework, it would be nice if the `beat` (or `libbeat`) namespace would be the only packages libbeat and the actual beats would meat. Like: `beat.Client` being the instance used to publish a `beat.Event`. Only problem is initialization and circular dependencies due to having the main in `beat` package as well.

urso: Will move it to `common` for now. Seems to be the place to prevent potential circular dependencies for now.

urso: done

libbeat/monitoring/report/report.go
urso: just realized I don't need `Start()` ;)

libbeat/beat/beat.go
urso: moved to `common.BeatInfo`

ruflin:
@urso Seems like you broke the beat generator. Also there you have to pass BeatInfo.

libbeat/monitoring/monitoring.go
ruflin: Conclusion on this is that we should not add it to `make update`, correct?

libbeat/publisher/publish.go
ruflin: `beat.Info` would be nice, but lets keep such a more "radical" move for later.

libbeat/monitoring/monitoring.go
urso: yeah, I wouldn't add it to `make update`. I have had the build broken when just doing unit tests without having build the full project before `go generate`. Maybe introduce separate `make go-generate`, but this one might fail and developer has to 'fix' it locally. generated files should be committed.

libbeat/beat/beat.go
```diff
@@ -138,20 +141,35 @@ func init() {
...
+	hostname, err := os.Hostname()
+	if err != nil {
+		return nil, err
 	}
+
+	return &Beat{
+		Info: common.BeatInfo{
+			Beat:     name,
+			Version:  version,
+			Name:     hostname,
```
monicasarbu: I think here you want `Name: name`, that it's equal to the `name` configuration option.

urso: name is the name configured via config file and defaults to hostname. beat.Beat is the beat name like `filebeat`, `packetbeat`. We set `Name` later after having parsed the config file.

libbeat/monitoring/report/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,309 @@
...
+
+		snapshot := makeSnapshot(monitoring.Default)
+		if snapshot == nil {
+			debugf("Empty snapshot.")
+			continue
+		}
+
+		event := outputs.Data{Event: common.MapStr{
+			"timestamp": common.Time(ts),
+			"beat":      r.beatMeta,
```
monicasarbu: Instead of exporting `beat.beat`, wouldn't be nicer `beat.type`?

libbeat/monitoring/report/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,309 @@
...
+		}, nil)
+		if err != nil {
+			return nil, err
+		}
+
+		return newPublishClient(esClient, params, config.BulkMaxSize), nil
+	}, nil
+}
+
+func parseProxyURL(raw string) (*url.URL, error) {
```
monicasarbu: maybe we should have this in `common` to be used later by others?

libbeat/outputs/elasticsearch/url.go
```diff
@@ -114,3 +114,18 @@ func makePath(index string, docType string, id string) (string, error) {
 	}
 	return path, nil
 }
+
+func parseProxyURL(raw string) (*url.URL, error) {
```
monicasarbu: multiple times defined?

urso: added todo item. The HTTP connection/configurations should be split and generalized for all beats. e.g. heartbeat and metricbeat also have http clients with similar settings.

libbeat/monitoring/report/elasticsearch/elasticsearch.go
urso: done

libbeat/monitoring/report/elasticsearch/elasticsearch.go
urso: added todo item

PR #3518, opened 2017-02-02
Packetbeat protocol analyzer enhancements

urso:
- Optionally configure protocol analyzers using dictionary and/or list
- Add 'fields', 'fields_under_root' and 'tags' settings to every
  protocol analyzer

This change allows for configuring packetbeat protocols in 2 different styles... both styles can be used at the same time.

1.) dictionary style:

```
packetbeat.protocols.http:
  ...
packetbeat.protocols.dns:
  ...
```

2.) array style:

```
packetbeat.protocols:
- type: http
  ...
- type: dns
  ...
```

Examples (1) and (2) are equivalent. But array style allows to configure a protocol analyzer multiple times: e.g.

(3) array style with multiple instances of http protocol analyzer:

```
packetbeat.protocols:
- type: http
  ports: [80]
  fields.service: nginx
- type: http
  ports: [9200]
  fields.service: elasticsearch
```

4) mixed style:

```
packetbeat.protocols.http:
  ...

packetbeat.protocols:
- type: dns
  ...
```

Limitations:

a) due to limitations in yaml parser, only capturing the last 'name' in a dictionary the key name `packetbeat.protocols` must not be used multiple times. e.g. this will result in an incompletely processed config (only DNS will be configured):

```
packetbeat.protocols:
  http:
    ...

packetbeat.protocols:
- type: dns
```

b) Reusing port numbers (overlapping) might result in one module not seeing any packets (this is already the case if any 2 protocols shall listen on same port number). e.g.:

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -92,6 +92,8 @@ https://github.com/elastic/beats/compare/v5.1.1...master[Check the HEAD diff]
 - The HAProxy module is now GA, instead of experimental. {pull}3525[3525]
 
 *Packetbeat*
+- Add `fields` and `fields_under_root` to packetbeat protocols configurations. {pull}3518[3518]
+- Add list style packetbeat protocols configurations. This change supports specifying multiple configurations of the same protocol analyzer. {pull]3518[3518]
 
 *Winlogbeat*
 
@@ -110,6 +112,7 @@ https://github.com/elastic/beats/compare/v5.1.1...master[Check the HEAD diff]
 *Metricbeat*
 
 *Packetbeat*
+- Deprecate dictionary style protocols configuration. {pull}3518[3518]
 
 *Winlogbeat*
 
--- a/packetbeat/_meta/beat.full.yml
+++ b/packetbeat/_meta/beat.full.yml
@@ -58,11 +58,12 @@ packetbeat.flows:
 
 #========================== Transaction protocols =============================
 
-packetbeat.protocols.icmp:
+packetbeat.protocols:
+- type: icmp
   # Enable ICMPv4 and ICMPv6 monitoring. Default: true
   #enabled: true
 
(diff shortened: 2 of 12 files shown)


ruflin:
* Can you provide an example on how to config could look afterwards?
* As usual, -1 on `fields_under_root` ;-)

urso:
> As usual, -1 on fields_under_root ;-)

I just reused already available functionality...

ruflin:
@urso I suggest we change our current default config to the array style and deprecate the old style. This will mean we have the same structure across all beats.

CHANGELOG.asciidoc
```diff
@@ -92,6 +92,8 @@ https://github.com/elastic/beats/compare/v5.1.1...master[Check the HEAD diff]
 - The HAProxy module is now GA, instead of experimental. {pull}3525[3525]
 
 *Packetbeat*
+- Add `fields` and `fields_under_root` to packetbeat protocols configurations. {pull}3518[3518]
+- Add list style packetbeat protocols configurations. This change support different configurations of same protocol analyzers to be active. {pull]3518[3518]
```
dedemorton: Would change this to say, "Add list style Packetbeat protocol configurations. This change supports different configurations of the same protocol analyzer to be active." The second sentence is a bit confusing even with the edit. How about this instead? "This change supports specifying multiple configurations of the same protocol analyzer." [saying "to be active" makes the sentence harder to parse and isn't IMO necessary.]

urso: done

ruflin:
One step closer to have beats configs unified 🎉 

About defining a port multiple times. This is kind of similar to defining a file multiple times in filebeat, just easier to detect. If there would be a registry of ports, we could return an error in case this happens.

Having protocols as list will also make it easier to implement config reloading.

PR #3519, opened 2017-02-02
Enable cgroup metric collection by default

andrewkroh:
On Linux cgroup metric collection will be enabled by default. The feature is no longer labeled as experimental. The configuration option has been renamed from `cgroups` to `process.cgroups.enabled`.

### DIFF
--- a/metricbeat/metricbeat.full.yml
+++ b/metricbeat/metricbeat.full.yml
@@ -64,8 +64,8 @@ metricbeat.modules:
   # if true, exports the CPU usage in ticks, together with the percentage values
   #cpu_ticks: false
 
-  # EXPERIMENTAL: cgroups can be enabled for the process metricset.
-  #cgroups: false
+  # Enable collection of cgroup metrics from processes on Linux.
+  #process.cgroups.enabled: true
 
   # A list of regular expressions used to whitelist environment variables
   # reported with the process metricset's events. Defaults to empty.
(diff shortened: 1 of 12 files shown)


metricbeat/metricbeat.full.yml
```diff
@@ -64,8 +64,8 @@ metricbeat.modules:
   # if true, exports the CPU usage in ticks, together with the percentage values
   #cpu_ticks: false
 
-  # EXPERIMENTAL: cgroups can be enabled for the process metricset.
-  #cgroups: false
+  # Enable collection of cgroup metrics from processes on Linux.
+  #process.cgroups.enabled: true
```
ruflin: Should we add it to the short config as well as it is now a "default" feature?

andrewkroh: I'd rather leave it out of the short config. I'd actually be in favor of removing more stuff from the short config. I feel like there's a lot of "noise" in the files and we could have short comment at the top about "see xbeat.full.yml for a commented example with all options" and then remove everything that not necessary. I like the readability of something like [this](https://github.com/elastic/beats-docker/blob/master/build/metricbeat/config/metricbeat.yml).

ruflin: @andrewkroh Sounds like something we should discuss in general as it would also apply to all other beats. I kind of like the docs as it helps new users (I hope) but it is much more noisy, agree.

PR #3520, opened 2017-02-02
Add edits for prometheus module docs

dedemorton:
A few minor editorial fixes.

@ruflin The fields.yml file for the `collector` metricset is empty: https://github.com/elastic/beats/blob/master/metricbeat/module/prometheus/collector/_meta/fields.yml. Can you update the fields.yml file (or let me know the fields/descriptions, and I can add them)?

### DIFF
--- a/metricbeat/module/prometheus/collector/_meta/docs.asciidoc
+++ b/metricbeat/module/prometheus/collector/_meta/docs.asciidoc
@@ -2,6 +2,6 @@
 
 The Prometheus `collector` metricset fetches data from https://prometheus.io/docs/instrumenting/exporters/[prometheus exporters].
 
-All events with the same labels are grouped together as one event.
-
-//REVIEWERS: Should we say something about which Prometheus exporters are supported? Users might follow the link and think that all the exporters listed on the page are supported, but I'm not sure if that's true.
+All events with the same labels are grouped together as one event. The fields
+exported by this metricset vary depending on the Prometheus exporter that you're
+using.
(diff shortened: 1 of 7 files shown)


metricbeat/module/prometheus/collector/_meta/docs.asciidoc
```diff
@@ -1,5 +1,7 @@
-=== prometheus collector MetricSet
+=== Prometheus Collector Metricset
 
-This is the collector metricset of the module prometheus. It allows to fetch data from https://prometheus.io/docs/instrumenting/exporters/[prometheus exporters].
+The Prometheus `collector` metricset fetches data from https://prometheus.io/docs/instrumenting/exporters/[prometheus exporters].
 
 All events with the same labels are grouped together as one event.
+
+//REVIEWERS: Should we say something about which Prometheus exporters are supported? Users might follow the link and think that all the exporters listed on the page are supported, but I'm not sure if that's true.
```
ruflin: All the exporters are supported as it is all the same format.

dedemorton:
@ruflin I think users might be confused if they try look at the "Exported Fields" topic to figure out more about the fields returned by the collector metricset. Could we add some basic content in the fields.yml file so that we end up with a section that says something like the following? At least they'll understand why the field they're looking for is missing from the doc.

collector Fields
Stats about the Prometheus exporter. The fields exported by this metricset vary depending on the exporter that you're using.

ruflin:
@dedemorton We should not use the `fields.yml` here as it is also use to generate the template and other things. But we should add this to the docs of the metricset itself.

dedemorton:
@ruflin Updated based on your feedback.

dedemorton:
@ruflin Rebased. Should be ready to merge as soon as the checks have run.

PR #3523, opened 2017-02-03
Update docker image with data.json

ruflin:
* Add docs for labels and tags

### DIFF
--- a/metricbeat/module/docker/container/_meta/fields.yml
+++ b/metricbeat/module/docker/container/_meta/fields.yml
@@ -40,3 +40,13 @@
           type: long
           description: >
              Size of the files that have been created or changed since creation.
+    - name: labels
+      type: dict
+      dict-type: keyword
+      description: >
+        Image labels.
+    - name: tags
+      type: list
+      dict-type: keyword
+      description: >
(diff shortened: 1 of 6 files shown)


metricbeat/module/docker/container/_meta/fields.yml
```diff
@@ -40,3 +40,13 @@
...
+    - name: labels
+      type: dict
+      dict-type: keyword
+      description: >
+        Image labels.
+    - name: tags
+      type: list
+      dict-type: keyword
+      description: >
+        Image labels.
```
douaejeouit: @ruflin Don't you think that the tags description need to be updated?

ruflin: Definitively. Fixed. Thanks for spotting.

andrewkroh: Does this file need re-generated? I think it's supposed to say "Image tags"?

ruflin: No, my mistake. I updated the wrong file and then run `make update` :-( New version pushed.

PR #3528, opened 2017-02-03
Disable date detection in index templates

### LINKED ISSUE #3389: Disable date detection in index templates
We use dynamic fields in some places and date detection is enabled by default so if a Beat sends a dynamic string field that looks like a date, the field mapping will be `date`. This means that all follow on data using the same field name must also be a date or else a field mapping exception will occur.

One example where this can happen is in Winlogbeat's `event_data.*` fields. Another example is with string fields parsed from JSON.

I propose that we disable date detection in our index templates to prevent this. And require explicit mappings for date fields in our templates.

Below is a self contained example showing usage of `date_detection: true/false` where two events are indexed. The dynamic template is based on our own from 5.1.2. One event goes into an index where date detection is enabled and another event goes into an index where date detection has been disabled.

```
PUT _template/tester
{
  "mappings": {
    "_default_": {
      "_all": {
        "norms": false
      },
      "_meta": {
        "version": "5.1.2"
      },
      "dynamic_templates": [
        {
          "strings_as_keyword": {
            "mapping": {
              "ignore_above": 1024,
              "type": "keyword"
            },
            "match_mapping_type": "string"
          }
        }
      ],
      "properties": {
        "@timestamp": {
          "type": "date"
        }
      }
    }
  },
  "order": 0,
  "settings": {
    "index.refresh_interval": "5s"
  },
  "template": "tester*"
}

PUT _template/tester-nodates
{
  "mappings": {
    "_default_": {
      "date_detection": false
    }
  },
  "order": 1,
  "template": "tester-nodates"
}

PUT tester/mytype/1
{
  "data": {
    "param1": "dynamic dates enabled",
    "param2": "2016/06/17"
  }
}

PUT tester-nodates/mytype/1
{
  "data": {
    "param1": "dynamic dates disabled",
    "param2": "2016/06/17"
  }
}

GET tester*/mytype/_mapping/field/data.param*

DELETE tester*
DELETE _template/*
```

This is the field mapping:
```
{
  "tester": {
    "mappings": {
      "mytype": {
        "data.param2": {
          "full_name": "data.param2",
          "mapping": {
            "param2": {
              "type": "date",
              "format": "yyyy/MM/dd HH:mm:ss||yyyy/MM/dd||epoch_millis"
            }
          }
        },
        "data.param1": {
          "full_name": "data.param1",
          "mapping": {
            "param1": {
              "type": "keyword",
              "ignore_above": 1024
            }
          }
        }
      }
    }
  },
  "tester-nodates": {
    "mappings": {
      "mytype": {
        "data.param2": {
          "full_name": "data.param2",
          "mapping": {
            "param2": {
              "type": "keyword",
              "ignore_above": 1024
            }
          }
        },
        "data.param1": {
          "full_name": "data.param1",
          "mapping": {
            "param1": {
              "type": "keyword",
              "ignore_above": 1024
            }
          }
        }
      }
    }
  }
}
```

andrewkroh:
We use dynamic fields in some places and date detection is enabled by default so if a Beat sends a dynamic string field that looks like a date, the field mapping will be date. This means that all follow on data using the same field name must also be a date or else a field mapping exception will occur. Disabling date detection will make the generated index mappings more predicable.

Closes #3389

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -69,6 +69,7 @@ https://github.com/elastic/beats/compare/v5.1.1...master[Check the HEAD diff]
 - Add the option to pass custom HTTP headers to the Elasticsearch output. {pull}3400[3400]
 - Unify `regexp` and `contains` conditionals, for both to support array of strings and convert numbers to strings if required. {pull}3469[3469]
 - Add the option to load the sample dashboards during the Beat startup phase. {pull}3506[3506]
+- Disabled date detection in Elasticsearch index templates. Date fields must be explicitly defined in index templates. {pull}3528[3528]
 
 *Metricbeat*
 
--- a/filebeat/filebeat.template-es2x.json
+++ b/filebeat/filebeat.template-es2x.json
@@ -9,6 +9,7 @@
       "_meta": {
         "version": "6.0.0-alpha1"
       },
+      "date_detection": false,
       "dynamic_templates": [
         {
           "strings_as_keyword": {
--- a/filebeat/filebeat.template.json
+++ b/filebeat/filebeat.template.json
@@ -4,6 +4,7 @@
       "_meta": {
         "version": "6.0.0-alpha1"
       },
+      "date_detection": false,
       "dynamic_templates": [
         {
           "strings_as_keyword": {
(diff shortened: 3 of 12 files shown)


ruflin:
This could mainly have an effect on "dynamic" fields which are coming from prometheus/collector etc. But if that is becoming an issue, we should fix it there somehow. It seems `date_detection` can only be set on the root object.

PR #3531, opened 2017-02-04
Heartbeat HTTP sample dashboards

tsg:
I'm opening this early, but we shouldn't merge it until we figure out if this can/should work with older versions of Kibana.

2017-02-04-22-28-localhost-5601


tsg:
This is how it looks in Kibana 4.2:

2017-02-07-13-09-localhost-5601

The heatmap viz at the end doesn't work, as expected, but apart from that everything looks fine. I'd say we include it with 5.3 as is? I can also remove the heatmap viz and add it without.

Issue #3104, opened 2016-12-02
The -once CLI flag is not documented


tsg:
Added in #2456, the `-once` CLI flag is not showing up in our docs, only in the `-h` output. It should be documented, especially since its usage requires some explanations (like the recommendation to enable `close_eof`).

PR #3535, opened 2017-02-06
Add docs for -once command line option

ruflin:
Closes https://github.com/elastic/beats/issues/3104

### DIFF
--- a/filebeat/docs/command-line.asciidoc
+++ b/filebeat/docs/command-line.asciidoc
@@ -1,8 +1,13 @@
 [[filebeat-command-line]]
 === Command Line Options
 
-Filebeat does not have any Filebeat-specific command line options. Instead, you
-configure the behaviour of Filebeat by specifying options in the configuration file.
+The following command line option is specific to Filebeat.
+
+*`-once`*::
+When the `-once` flag is used, Filebeat starts all configured harvesters and prospectors, and runs
+each prospector until the harvesters are closed. If you set the `-once` flag, you should also set
+`close_eof` so the harvester is closed when the end of the file is reached.
+By default harvesters are closed after `close_inactive` is reached.
 
 The following command line options from libbeat are also available for Filebeat. To
 use these options, you need to start Filebeat in the foreground.


filebeat/docs/command-line.asciidoc
```diff
@@ -1,9 +1,15 @@
 [[filebeat-command-line]]
 === Command Line Options
 
-Filebeat does not have any Filebeat-specific command line options. Instead, you
+Filebeat does only have one Filebeat-specific command line option. For all other options you
```
dedemorton: I would say, "The following command line option is specific to Filebeat." I'd just remove "For all other options you configure the behaviour of Filebeat by specifying options in the configuration file." 

BTW, I'm thinking about reworking the CLI topics to use conditional tagging instead of includes (that way we could just list all the flags and not explain where they are coming from).

filebeat/docs/command-line.asciidoc
```diff
@@ -1,9 +1,15 @@
...
 configure the behaviour of Filebeat by specifying options in the configuration file.
 
+*`-once`*::
+When the `-once` flag is used, filebeat starts all configured harvesters and prospectors and runs
```
dedemorton: When the `-once` flag is used, Filebeat starts all configured harvesters and prospectors, and runs... (capitalize Filebeat and add comma).

filebeat/docs/command-line.asciidoc
```diff
@@ -1,9 +1,15 @@
...
+each prospector until the harvesters are closed. It is recommended to use the flag in combination
+with `close_eof` so harvester directly close when the end of the file is reached.
```
dedemorton: Maybe a little less formal would be good here: "If you set the `once` flag, you should also set `close_eof` so the harvester is closed when the end of the file is reached."

filebeat/docs/command-line.asciidoc
```diff
@@ -1,9 +1,15 @@
...
+By default harvesters are closed after `close_inactive`.
```
dedemorton: ...after `close_inactive` is reached?

ruflin:
@dedemorton Adjustments made as suggested.

dedemorton:
LGTM

Issue #3466, opened 2017-01-25
Support pprof metricset for golang  programs


medcl:
As we discussed today, golang’s performance tool `pprof` is very powerful tool for you to monitoring and export GO program's runtime performance, maybe we can add a `pprof` metricset module in Metricbeat, that said we can monitoring golang's runtime performance and do a long-term performance analysis, performance issue such as memory leak can be found quickly.

The `pprof` can provide too much information such as call graph or goroutine stacktrace, in Metricbeat, we may just focus the high level runtime metrics, such as:
- [ ] <del>CPU: utilization </del>
- [x] Heap: HeapAlloc, HeapSys, HeapObjects, HeapReleased, Sys, Sys-HeapReleased, StackSy
- [ ] Goroutine: num of goroutines
- [x] GC: GC time, GC count
- [ ] Some interesting calls: cgo calls, syscall.Read/Write, runtime.mallocgc

Once we found some potential performance issue from these metrics, we can then use some specify tools to investigate the problem. 

Relevant links:
[https://golang.org/pkg/runtime/#MemStats](https://golang.org/pkg/runtime/#MemStats)
[https://golang.org/pkg/net/http/pprof/](https://golang.org/pkg/net/http/pprof/)
[https://blog.golang.org/profiling-go-programs](https://blog.golang.org/profiling-go-programs)
[https://software.intel.com/en-us/blogs/2014/05/10/debugging-performance-issues-in-go-programs](https://software.intel.com/en-us/blogs/2014/05/10/debugging-performance-issues-in-go-programs)

urso:
all but CPU utilization should be available. All the GC/Heap info are available via runtime.MemStats (already exposed via expvar "memstats").

expvar also exposes the command line used to start the process. This should be added as well, so it's more clear, which process is currently being monitored.

medcl:
@urso i was thinking, in metricbeat, if  the pprof address user specified is a not local one, such as `http://another_remote_ip:6060/debug/pprof/`, there maybe a issue with CPU utilization, and for other info like heap or GC  i think should get from the `pprof` API, what's you opinion?

ruflin:
I would also not get CPU for the moment as our recommendation is to install beats as local as possible. Assuming in the case of containers you need remote IP's we support docker and cgroups.

Can we start with memory related data which helps to find memory leaks? I think that could be quite useful.

pbennett:
[...]

PR #3536, opened 2017-02-06
Golang module for metricbeat

medcl:
This PR is for https://github.com/elastic/beats/issues/3466

- [x] Expvar: memstats
- [x] Expvar: specify custom metric keys
- [x]  test
- [x]  Kibana dashboards


metricbeat/module/golang/expvar/expvar.go
```diff
@@ -0,0 +1,127 @@
...
+
+	if err != nil {
+		logp.Err(err.Error())
+	}
+
+	event := common.MapStr{
+		"cmdline": json["cmdline"],
+	}
+
+	if m.memstatsEnabled {
```
ruflin: This should be its own metricset. I even thing that is the one we should start with.

metricbeat/module/golang/_meta/config.yml
```diff
@@ -0,0 +1,8 @@
+- module: golang
+  metricsets: ["expvar","heap"]
+  enabled: true
+  period: 10s
+  hosts: ["localhost:6060"]
+  metrics_path: "/debug/vars"
```
ruflin: We should probably have this under `expvar.path` and `heap.path` as these option do not necessarly exist or are different in other metricsets we add in the future to golang.

metricbeat/module/golang/heap/_meta/data.json
```diff
@@ -0,0 +1,47 @@
...
+        "heap":{
+            "cmdline": [
+                "./gopa"
+            ] ,
+            "alloc": 1759408,
+            "buck_hash_sys": 1466577,
+            "frees": 718931,
+            "gc_cpu_fraction": 0.000002,
+            "gc_sys": 440320,
+            "heap_alloc": 1759408,
```
ruflin: Make sure to follow the naming conventions: https://www.elastic.co/guide/en/beats/libbeat/5.2/event-conventions.html

metricbeat/module/golang/expvar/_meta/docs.asciidoc
```diff
@@ -0,0 +1,6 @@
+=== golang expvar MetricSet
+
+This is the expvar metricset of the module golang.
+Golang can exposes it's variables by expvar api, with this metricset, user can collect these metrics,
+By use parameter `expvar_keys` to specify which metrics to collect, by default or this parameter was not specified will
```
ruflin: I suggest to leave out this config and let the users use filters / processors for that?

metricbeat/module/golang/_meta/config.yml
medcl: make sense

metricbeat/module/golang/expvar/_meta/docs.asciidoc
medcl: @ruflin if the user didn't specify this config, the metrics will collect all the keys, are we still need to remove this config?

medcl: @ruflin  make sense to unify filtering logic, this config removed

medcl:
snip20170220_3
screenshot.

metricbeat/module/golang/expvar/_meta/data.json
```diff
@@ -0,0 +1,60 @@
...
+                    "failures": 0,
+                    "success": 1
+                },
+                "golang-heap": {
+                    "events": 1,
+                    "failures": 0,
+                    "success": 1
+                }
+            },
+            "libbeat.config.module.running": 0,
```
ruflin: This kind of keys will not work with 2.x releases. But that is something we should fix in the beats itself to return proper json ( @urso I remember you were doing something in this direction?). 

For services that provide this kind of keys, we can either replace all `. ` with `_` or from my perspective nicer would be to created the nested documents (see `common.MapStr.Put(...)`). This brings the risk in case a value and a nested object collide. Example `a.b = c; a.b.d = f` This is not possible.

Unrelated to the above, as this is a dynamic metricset, there should be an example json file.

metricbeat/module/golang/expvar/_meta/docs.asciidoc
```diff
@@ -0,0 +1,5 @@
+=== golang expvar MetricSet
+
+This is the expvar metricset of the module golang.
+Golang can exposes it's variables by expvar api, with this metricset, user can collect these metrics,
+collect all the expvar exposed variables except memstats(if you need memstats, check out another `heap` metricset in this module.
```
ruflin: Can you link here to the expvar docs from Golang perhaps? So in case someone has a golang app and wants to expose expvars, he gets some intructions on how to do it.

metricbeat/module/golang/expvar/expvar.go
```diff
@@ -0,0 +1,82 @@
...
+	json, err := m.http.FetchJSON()
+
+	if err != nil {
+		return nil, err
+	}
+
+	event := common.MapStr{}
+
+	//export all expvar keys, except memstats
+	json["memstats"] = nil
```
ruflin: I don't think we should exclude these here. That feels like too much magic. If someone does not want the values, he can use filters / processors to exclude them.

metricbeat/module/golang/expvar/expvar.go
```diff
@@ -0,0 +1,82 @@
...
+	event = json
+
+	//flatten cmdline
+	event["cmdline"] = golang.GetCmdStr(json["cmdline"])
+
```
ruflin: We need to set `_namespace` here to make it possible to have its own namespace for expvar. See https://github.com/elastic/beats/blob/master/metricbeat/module/prometheus/collector/collector.go#L91 as an example. Namespace config will also be needed.

metricbeat/module/golang/heap/_meta/data.json
```diff
@@ -0,0 +1,46 @@
...
+                "allocated": 2165152,
+                "total": 3393472,
+                "frees": 7975,
+                "idle": 1163264,
+                "mallocs": 27041,
+                "objects": 19066
+            },
+            "cmdline": "./metricbeat -httpprof=127.0.0.1:6060 -e -v",
+            "gc": {
+                "total_count": 1,
```
ruflin: `total.count`

metricbeat/module/golang/heap/_meta/data.json
```diff
@@ -0,0 +1,46 @@
...
+                "total_pause_ns": 152192,
+                "next_gc_limit": 4194304,
+                "pause_ns": {
```
ruflin:
```
pause.total.ns
pause.avg.ns
pause.count.ns
pause.max.ns
pause. sum.ns
```

Unit should always be last. Is there a reason that `total_pause_ns` is outside?

metricbeat/module/golang/heap/heap.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+// New create a new instance of the MetricSet
+// Part of new is also setting up the configuration by processing additional
+// configuration entries if needed.
+func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
+
+	logp.Warn("EXPERIMENTAL: The golang heap metricset is experimental")
+
+	config := struct {
```
ruflin: You can leave out the config and unpack part if there are no specific configs.

metricbeat/module/golang/heap/heap.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	// garbage collector summary
+	var duration, maxDuration, avgDuration, count uint64
+	// collect last gc run stats
+	if m.lastNumGC < ms.NumGC {
+		delta := ms.NumGC - m.lastNumGC
+		start := m.lastNumGC
+		if delta > 256 {
+			logp.Info("golang", "Missing %v gc cycles", delta-256)
```
ruflin: Info log message does not have a first param. Probably this was Debug first.

metricbeat/tests/system/test_golang.py
```diff
@@ -0,0 +1,55 @@
...
+                "-httpprof",
+                os.getenv(
+                    'GOLANG_HOST',
+                    'localhost') +
+                ":" +
+                os.getenv(
+                    'GOLANG_PORT',
+                    '6060')])
+
+        time.sleep(3)
```
ruflin: Why is this needed? This mostly leads to flaky tests. Best use some "log_contains" to block here.

metricbeat/module/golang/expvar/_meta/data.json
medcl: @ruflin i think the latest es 2.x already support dot in fields, here should be fine

metricbeat/docs/modules/golang.asciidoc
```diff
@@ -0,0 +1,41 @@
+////
+This file is generated! See scripts/docs_collector.py
+////
+
+[[metricbeat-module-golang]]
+== golang Module
+
+This is the golang Module. Metrics are collected submitting HTTP GET requests to golang-expvar-api.
+Reference: https://golang.org/pkg/expvar/
```
medcl: @ruflin the expvar doc already linked  in the module section.

metricbeat/module/golang/heap/_meta/data.json
medcl: here is not a group, it is only part of the name, it is not necessary to use dot, i think

metricbeat/module/golang/heap/_meta/data.json
medcl: `total_pause_ns` is gc pause time during lifetime of the golang program,
and the `pause_ns` is durations during the metricbeat collect period, it is a group, during the collect period,
we calculate sum, max and avg.

metricbeat/module/golang/expvar/_meta/docs.asciidoc
medcl: already did in module section.

metricbeat/module/golang/expvar/_meta/data.json
ruflin: It is supported in 2.4, but only if enabled. But I agree we should not worry about this currently for dynamic metricsets as it will add quite a bit complexity. You would have to scan the full json doc.

metricbeat/module/golang/heap/_meta/data.json
ruflin: We always do a dot for things which are in our naming convention: https://www.elastic.co/guide/en/beats/libbeat/5.2/event-conventions.html count is an indication on what field type it is. So even if it is not belonging ti a group, I recommend to add the dot.

metricbeat/module/golang/heap/_meta/data.json
ruflin: I'm trying to find a way to make this more obvious. Best option I can come with at the moment is:
```
total_pause.ns
pause.avg.ns
pause.count.ns
pause.max.ns
pause.sum.ns
```
Could we add something to the name that makes the difference more obvious?

medcl: @ruflin 
is that you want?
```
event["gc"] = common.MapStr{
		"next_gc_limit": ms.NextGC,
		"total_count":   ms.NumGC,
		"cpu_fraction":  ms.GCCPUFraction,
		"total_pause": common.MapStr{
			"ns": ms.PauseTotalNs,
		},
		"pause": common.MapStr{
			"count": common.MapStr{
				"ns": count,
			},
			"sum": common.MapStr{
				"ns": duration,
			},
			"avg": common.MapStr{
				"ns": avgDuration,
			},
			"max": common.MapStr{
				"ns": maxDuration,
			},
		},
	}
```
if it is only for name purpose to prepare a group, i think we should avoid this

ruflin: Yes, that is what I mean. It is for naming purpose. As mentioned earlier, `max.ns` will break compatibility with 2.x releases. We initially used `max.ns` but had to change it because it started to break things. You are right that the code is not the nicest. Perhaps we find a nicer solution in the long run, but that is the current way we handle it.

metricbeat/docs/modules/golang.asciidoc
```diff
@@ -0,0 +1,42 @@
...
+----
+metricbeat.modules:
+#- module: golang
+#  metricsets: ["expvar","heap"]
+#  enabled: true
+#  period: 10s
+#  hosts: ["localhost:6060"]
+#  expvar.path: "/debug/vars"
+#  heap.path: "/debug/vars"
+#  expvar.namespace: "example"
```
ruflin: would be good to group `expvar` config options together. Can be done in a follow up PR if you like.

metricbeat/module/golang/expvar/_meta/data.json
```diff
@@ -0,0 +1,922 @@
+{
```
ruflin: This should be removed, as expvar can have any content. Otherwise people think it is only for betas.

PR #3538, opened 2017-02-06
[WIP] Restructure Harvester and Prospector Shutdown

ruflin:
There are two options for stopping a harvester or a prospector. Either the harvester and prospector finish sending all events and stop themself or they are killed because the output is blocking.

In case of shutting down filebeat without using `shutdown_timeout` filebeat is expected to shut down as fast as possible. This means channels are directly closed and the events are not passed through to the registry.

In case of dynamic prospector reloading, prospectors and harvesters must be stopped properly as otherwise no new harvester for the same file can be started.

So far the following changes were made:

* Introduce harvester tracking in prospector to better control / manage the harvesters
* Use `beatDone` channel coming from the beat itself for direct shutdown as this channel is closed when the beat is stopped.
* Introduce more done channels in prospector to make shutdown more fine grained
* Add system tests to verify new behaviour

TODO
* Make sure all events are drained from channel to not guarantee that state updates will be persisted.

### DIFF
--- a/filebeat/beater/filebeat.go
+++ b/filebeat/beater/filebeat.go
@@ -133,7 +133,7 @@ func (fb *Filebeat) Run(b *beat.Beat) error {
 		return err
 	}
 
-	crawler, err := crawler.New(newSpoolerOutlet(fb.done, spooler, wgEvents), config.Prospectors, *once)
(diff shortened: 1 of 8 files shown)


ruflin:
Closing in favor of https://github.com/elastic/beats/pull/3563

PR #3540, opened 2017-02-06
Add fileset.name and fileset.module fields

tsg:
This replaces the `fields.source_type` hack with a `fileset` object that
resembles the Metricbeat `metricset` object.

The implementation still uses a hack: adds two hidden options to the prospector
config, but it's the smaller evil IMHO.

Part of #3159.

### DIFF
--- a/filebeat/fileset/fileset.go
+++ b/filebeat/fileset/fileset.go
@@ -235,6 +235,16 @@ func (fs *Fileset) getProspectorConfig() (*common.Config, error) {
 		return nil, fmt.Errorf("Error setting the pipeline ID in the prospector config: %v", err)
 	}
 
+	// force our the module/fileset name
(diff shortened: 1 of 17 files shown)


filebeat/fileset/fileset.go
```diff
@@ -235,6 +235,16 @@ func (fs *Fileset) getProspectorConfig() (*common.Config, error) {
 		return nil, fmt.Errorf("Error setting the pipeline ID in the prospector config: %v", err)
 	}
 
+	// force our the module/fileset name
```
ruflin: force out?

ruflin:
@tsg I already merged it. Typo (if it is one) can still be fixed later.

filebeat/fileset/fileset.go
tsg: Ah, the "the" is extra. Thanks!

PR #3544, opened 2017-02-07
Configuration files must not be writeable by other users

andrewkroh:
This PR adds enforcement of ownership and file permissions on configuration files. Any configuration file must be owned by the same user that the Beat is running as and the file must not be writable by anyone other than the owner.

This strict permission checking is limited to platforms with POSIX file permissions. The DACLs used by Windows are not checked at this time.

The check can be disabled on the CLI with `-strict.perms=false`.

### DIFF
--- a/libbeat/common/config.go
+++ b/libbeat/common/config.go
@@ -5,14 +5,28 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"os"
+	"runtime"
 	"strings"
 
+	"github.com/elastic/beats/libbeat/common/file"
 	"github.com/elastic/beats/libbeat/logp"
 	"github.com/elastic/go-ucfg"
 	"github.com/elastic/go-ucfg/cfgutil"
 	cfgflag "github.com/elastic/go-ucfg/flag"
 	"github.com/elastic/go-ucfg/yaml"
 )
+
+var flagStrictPerms = flag.Bool("strict.perms", true, "Strict permission checking on config files")
+
+// IsStrictPerms returns true if strict permission checking on config files is
+// enabled.
+func IsStrictPerms() bool {
+	if !*flagStrictPerms || os.Getenv("BEAT_STRICT_PERMS") == "false" {
+		return false
+	}
+	return true
+}
 
 // Config object to store hierarchical configurations into.
 // See https://godoc.org/github.com/elastic/go-ucfg#Config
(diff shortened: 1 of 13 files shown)


andrewkroh:
jenkins, test it

libbeat/common/config.go
```diff
@@ -5,15 +5,29 @@ import (
...
 	cfgflag "github.com/elastic/go-ucfg/flag"
 	"github.com/elastic/go-ucfg/yaml"
 )
 
+var flagStrictPerms = flag.Bool("strict.perms", true, "Strict permission checking on config files")
+
+// IsStrictPerms returns true if strict permission checking on config files is
+// enabled.
+func IsStrictPerms() bool {
+	if !*flagStrictPerms || os.Getenv("BEAT_STRICT_PERMS") == "false" {
```
ruflin: We should probably document that there is also an environment variable available for this setting.

libbeat/common/config.go
```diff
@@ -390,3 +410,35 @@ func filterDebugObject(c interface{}) {
...
+		return nil
+	}
+
+	info, err := file.Stat(name)
+	if err != nil {
+		return err
+	}
+
+	euid := os.Geteuid()
+	fileUID, _ := info.UID()
```
tsg: Is ignoring the error here ok? I'm mainly interested in the behaviour on windows. If I understand correctly, fileUID will be -1 in that case, right? Is euid also -1 on Windows?

andrewkroh: It should never get this far because of the first check to see if this is running on Windows.

On Windows `FileInfo.UID()` will always return an error. And `syscall.Geteuid()` always returns -1.

libbeat/common/config.go
andrewkroh: Yeah, we will need to update the documentation to explain the ownership and permission requirements. This info can go in there.

SlavaValAl:
Guys, do you have plans to make this option configurable via configuration file?

andrewkroh:
@SlavaValAl No. What are you doing that you need this?

SlavaValAl:
@andrewkroh Hi, I'm working on application tool what generates prospectors configurations. I'm using this https://github.com/elastic/beats/pull/3362 feature to get it reloaded automatically. But, my application and filebeat are acting under different users. So, I'm using feature from current pull request, and, I'm prefer to configure this option in filebeat.yml instead of environment variable or CLI option. It's not so critical, but, it would be nice to have it.

andrewkroh:
Well we can't put the option to disable the check in the config file because that would be self-defeating. A bad actor could drop a config file in disabling the check and do bad things.

I think for your case I would disable the check and lock down the config permissions. The check is there to prevent people from accidentally making themselves vulnerable. I think I would make the `conf.d` dir writable only by some group that is common to both applications. That should prevent other users from being able to drop an arbitrary config file in and have it be executed.

Issue #3533, opened 2017-02-06
winlogbeat / packetbeat: Sample Dashboards broken with ES 5.2 


jochen-st:
After upgradeing KIbana and Elasticsearch 5.2 from 5.1.2, the sample winlogbeat dashboards / visualizations provided from https://artifacts.elastic.co/downloads/beats/beats-dashboards/beats-dashboards-5.2.0.zip do not work anymore.

Kibana displays the following error:

    Error: [illegal_argument_exception] [terms] exclude doesn't support values of type: START_OBJECT

The issue is likely caused by https://github.com/elastic/elasticsearch/issues/22933

Fix is to replace

    "exclude": {"pattern": "\"\""},

with

    "exclude": "\"\"",

in the affected visualisations:
winlogbeat: Event-Levels.json and Evolution-of-the-CPU-times-per-process.json
packetbeat: NFS-errors.json

jochen-st:
see https://github.com/jochen-st/beats/commit/eedc7c1834d8c5d9bb0d2c2e5a0a963be64ee215 for possible fix

tsg:
Closing as fixed.

PR #3548, opened 2017-02-07
Remove exclude pattern from all visualizations

monicasarbu:
As of the GitHub [issue](https://github.com/elastic/elasticsearch/issues/22933), `exclude pattern` doesn't work anymore in Kibana. This PR is removing the `exclude pattern` from the visualizations to fix the [issue](https://github.com/elastic/beats/issues/3533).

For the NFS errors visualization, a new search is needed after removing of the exclude pattern. The search selects all the NFS errors. We consider a NFS error if the `nfs.status` is different than `NFS_OK` and `NFSERR_NOENT`.

cc-ed @andrewkroh

### DIFF
--- a/packetbeat/_meta/kibana/search/NFS-errors-search.json
+++ b/packetbeat/_meta/kibana/search/NFS-errors-search.json
@@ -1,0 +1,16 @@
+{
+  "sort": [
+    "@timestamp", 
+    "desc"
+  ], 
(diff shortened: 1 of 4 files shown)


tsg:
There were failures on this one, but I'm merging it because the failures are not related to the dashboards, and we need this for 5.2.1

PR #3549, opened 2017-02-07
Update go-ucfg to version 0.4.6

urso:


ruflin:
Merging as Jenkins is green and travis is broken because of php-fpm.

PR #3563, opened 2017-02-09
Fix harvester shutdown for prospector reloading 

### LINKED ISSUE #3546: Known Issue: Prospector reloading unfinished files
Prospector reloading was introduce in 5.3. This issue is intended to describe a known issue with the implementation.

This bug affects all reloading which reloads a prospector with a file that was harvested before. If a prospector is started with new files this should not have any effect. In general the recommendation is to use reloading not to update settings like `fields` or `paths` in a prospector but to add new prospectors with new paths and to remove old ones.

## Example Working

subconfig.yml before
```
- input_type: log
  paths:
    - /var/log/test.log
  scan_frequency: 1s
```

subconfig.yml after
```
- input_type: log
  paths:
    - /var/log/test.log
  scan_frequency: 1s
- input_type: log
  paths:
    - /var/log/newfile.log
  scan_frequency: 1s
```

This works because only a new prospector for `newfile.log` has to be started and the older prospector keeps running.

## Example NOT Working

subconfig.yml before
```
- input_type: log
  paths:
    - /var/log/test.log
  scan_frequency: 1s
```

subconfig.yml after
```
- input_type: log
  paths:
    - /var/log/test.log
    - /var/log/newfile.log
  scan_frequency: 1s
```

The above is not working because the prospector with `test.log` has to be stopped and a new prospector with `test.log` has to be started.

## Technical Details

On shutdown filebeat tries to shut down as fast as possible. This means on shutdown filebeat is not waiting to complete sending all events and persisting all states to disk as it is unknown how long this will take. This can become an issue with reloading.

When a new prospector configuration option is loaded, filebeat ensures that all states which are loaded by the new prospector have the state set to finish. This is to verify that no file is harvested by 2 different prospectors at the same time as this could lead to duplicated events and unexpected behaviour.

In Filebeat in `harvester/log.go` the following code exists which is used to send events and state.

```
func (h *Harvester) sendEvent(event *input.Event) bool {
	select {
	case <-h.done:
		return false
	case h.prospectorChan <- event: // ship the new event downstream
		return true
	}
}
```

This code is called the last time before Stopping a harvester with `state.Finished: true` to set the state to finished, so a new harvester can pick it up. The problem now is that it can happen, that this state is not sent because the select statement goes into `h.done`. On normal shutdown this is not an issue it is not persisted to disk if a state is Finished or not. For prospector reloading it matters because the memory states contain the info if the state is Finished. So in case `h.done` is selected instead of `h.propsecotrChan` a harvester state is never finished.

The reason `h.done` is required here is that when stopping filebeat and the output is blocking, it still shuts down directly. This implies that a prospector and harvester need 2 different stop methods: One that waits for completion of sending which is used for reloading and one that shuts down immediately.

Some experiments happen in https://github.com/elastic/beats/pull/3538

ruflin:
There are two options for stopping a harvester or a prospector. Either the harvester and prospector finish sending all events and stop them self or they are killed because the output is blocking.

In case of shutting down filebeat without using `shutdown_timeout` filebeat is expected to shut down as fast as possible. This means channels are directly closed and the events are not passed through to the registry.

In case of dynamic prospector reloading, prospectors and harvesters must be stopped properly as otherwise no new harvester for the same file can be started. To make this possible the following changes were made:

* Introduce harvester tracking in prospector to better control / manage the harvesters. The implementation is based on a harvester registry which starts and stops the harvesters
* Use an outlet to send events from harvester to prospector. This outlet has an additional signal to have two options on when the outlet should be finished. Like this the outlet can be stopped by the harvester itself or globally through closing beatDone.
* Introduce more done channels in prospector to make shutdown more fine grained
* Add system tests to verify new behaviour

Closes #3546

### DIFF
--- a/filebeat/channel/outlet.go
+++ b/filebeat/channel/outlet.go
@@ -1,0 +1,80 @@
+package channel
+
+import (
+	"sync"
+	"sync/atomic"
+
+	"github.com/elastic/beats/filebeat/input"
+)
+
+type Outlet struct {
+	wg      *sync.WaitGroup
+	done    <-chan struct{}
+	signal  <-chan struct{}
+	channel chan *input.Event
+	isOpen  int32 // atomic indicator
+}
+
+func NewOutlet(
+	done <-chan struct{},
+	c chan *input.Event,
+	wg *sync.WaitGroup,
+) *Outlet {
+	return &Outlet{
+		done:    done,
+		channel: c,
+		wg:      wg,
+		isOpen:  1,
+	}
+}
+
+// SetSignal sets the signal channel for OnEventSignal
+func (o *Outlet) SetSignal(signal <-chan struct{}) {
+	o.signal = signal
+}
+
+func (o *Outlet) OnEvent(event *input.Event) bool {
+	open := atomic.LoadInt32(&o.isOpen) == 1
+	if !open {
+		return false
+	}
+
+	if o.wg != nil {
+		o.wg.Add(1)
+	}
+
+	select {
+	case <-o.done:
+		if o.wg != nil {
+			o.wg.Done()
+		}
+		atomic.StoreInt32(&o.isOpen, 0)
+		return false
+	case o.channel <- event:
+		return true
+	}
+}
+
+// OnEventSignal can be stopped by the signal that is set with SetSignal
+// This does not close the outlet. Only OnEvent does close the outlet.
+func (o *Outlet) OnEventSignal(event *input.Event) bool {
+	open := atomic.LoadInt32(&o.isOpen) == 1
+	if !open {
+		return false
+	}
+
+	if o.wg != nil {
+		o.wg.Add(1)
+	}
+
+	select {
+	case <-o.signal:
+		if o.wg != nil {
+			o.wg.Done()
+		}
+		o.signal = nil
+		return false
+	case o.channel <- event:
+		return true
+	}
+}
--- a/filebeat/channel/outlet.go
+++ b/filebeat/channel/outlet.go
@@ -7,8 +7,16 @@ import (
 	"github.com/elastic/beats/filebeat/input"
 )
 
+// Outlet struct is used to be passed to an object which needs an outlet
+//
+// The difference between signal and done channel is as following:
+// - signal channel can be added through SetSignal and is used to
+//   interrupt events sent through OnEventSignal-
+// - done channel is used to close and stop the outlet
+//
+// If SetSignal is used, it must be ensure that there is only one event producer.
 type Outlet struct {
-	wg      *sync.WaitGroup
+	wg      *sync.WaitGroup // Use for counting active events
 	done    <-chan struct{}
 	signal  <-chan struct{}
 	channel chan *input.Event
@@ -29,6 +37,7 @@ func NewOutlet(
 }
 
 // SetSignal sets the signal channel for OnEventSignal
+// If SetSignal is used, it must be ensure that only one producer exists.
 func (o *Outlet) SetSignal(signal <-chan struct{}) {
 	o.signal = signal
 }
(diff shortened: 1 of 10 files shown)


filebeat/channel/outlet.go
```diff
@@ -0,0 +1,80 @@
...
+		channel: c,
+		wg:      wg,
+		isOpen:  1,
+	}
+}
+
+// SetSignal sets the signal channel for OnEventSignal
+func (o *Outlet) SetSignal(signal <-chan struct{}) {
+	o.signal = signal
+}
```
urso: why we need `SetSignal`? This way the outlet can not be reused by multiple harvesters? In this case please add Note(godoc) to Outlet that it requires single-producers.

filebeat/channel/outlet.go
```diff
@@ -0,0 +1,80 @@
...
+	"sync"
+	"sync/atomic"
+
+	"github.com/elastic/beats/filebeat/input"
+)
+
+type Outlet struct {
+	wg      *sync.WaitGroup
+	done    <-chan struct{}
+	signal  <-chan struct{}
```
urso: Make difference between signal and done clear as godoc here?

filebeat/channel/outlet.go
```diff
@@ -0,0 +1,80 @@
...
+
+import (
+	"sync"
+	"sync/atomic"
+
+	"github.com/elastic/beats/filebeat/input"
+)
+
+type Outlet struct {
+	wg      *sync.WaitGroup
```
urso: wg used for counting active events?

filebeat/prospector/prospector.go
```diff
@@ -195,21 +210,71 @@ func (p *Prospector) updateState(event *input.Event) error {
...
+	}
+	// Waits until channel go-routine properly stopped
 	p.channelWg.Wait()
-	p.wg.Wait()
 }
 
 // createHarvester creates a new harvester instance from the given state
 func (p *Prospector) createHarvester(state file.State) (*harvester.Harvester, error) {
 
+	outlet := channel.NewOutlet(p.beatDone, p.harvesterChan, p.eventCounter)
```
urso: ah, I see... one outlet passed to harvester. If NewHarvester would get an interface instead of channel.Outlet, the wrapping could be done in the harvester. That is, the logic `OnEvent` and `OnEventSignal` will become more local to the harvester (no need for public definition).

filebeat/prospector/prospector.go
```diff
@@ -195,21 +210,71 @@ func (p *Prospector) updateState(event *input.Event) error {
...
+	// Waits on stopping all harvesters to make sure all events made it into the channel
+	p.waitEvents()
+}
+
+// Wait for completion of sending events
+func (p *Prospector) waitEvents() {
+
+	done := make(chan struct{})
+	go func() {
+		p.eventCounter.Wait()
```
urso: when calling wait here, are we sure no harvesters are running for sure? One MUST not use 'Add' after Wait has been invoked.

filebeat/channel/outlet.go
```diff
@@ -0,0 +1,80 @@
...
+	select {
+	case <-o.done:
+		if o.wg != nil {
+			o.wg.Done()
+		}
+		atomic.StoreInt32(&o.isOpen, 0)
+		return false
+	case o.channel <- event:
+		return true
+	}
```
urso: if outlet is used by multiple producers this can race I think. If it's used by one single producer only, there is no need for the atomics. Also, setting a channel to `nil` has a similar effect as a `nil` channel blocks forever and a closed channel always receives. e.g.:

```
type Outlet struct {
  stateUpdate, event cancelChannel
}

type cancelChannel struct {
  done chan struct{}
  ch chan *input.Event
}

func (ch *cancelChannel) OnEvent(evt *input.Event) bool {
  select {
  case <-ch.done:
    ch.ch = nil
    return false
  case ch.ch <- event:
    return true
  }    
}

func (o *Outlet) OnEvent(event *input.Event) bool {
  return o.event(event)
}

func (o *Outlet) OnStateUpdate(event *input.Event) bool {
  return o.stateUpdate.event(event)
}

```

filebeat/channel/outlet.go
ruflin: yes, added a comment.

filebeat/channel/outlet.go
urso: it's time to get rid of the spooler...

PR #3541, opened 2017-02-07
Add edits for couchbase module

dedemorton:

PR #3567, opened 2017-02-10
Cherrypick doc changes into 5.2 (cherrypicks #3520 and #3541)

dedemorton:
Cherrypicks edits for Prometheus (#3520) and Couchbase module #3541) plus reverts 5.3 name change that's not in 5.2.

### DIFF
--- a/metricbeat/module/prometheus/_meta/fields.yml
+++ b/metricbeat/module/prometheus/_meta/fields.yml
@@ -1,9 +1,9 @@
 - key: prometheus
   title: "Prometheus"
   description: >
-    Prometheus Module
-
-    experimental[]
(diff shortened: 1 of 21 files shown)


metricbeat/module/prometheus/_meta/fields.yml
```diff
@@ -1,9 +1,9 @@
 - key: prometheus
   title: "Prometheus"
   description: >
-    Prometheus Module
-
-    experimental[]
+    beta[]
```
ruflin: We should be careful not to change "history" as in the code it is still "experimental". I think it is fine here as we kind of "meant" beta back then but it didn't exist yet.

PR #3573, opened 2017-02-10
Filebeat: Add err check for config file read

### LINKED ISSUE #3572: Filebeat ignores additional config files on syntax error
If additional config files are defined with `filebeat.config_dir` and thous config files do have yaml syntax errors, these errors are silently ignored and filebeat is started normally. This is also true, if `-configtest` is used.

Version: current master

Steps to reproduce:

Have a filebeat.yml with `filebeat.config_dir: /tmp/filebeat` set.
Have an additional config file `/tmp/filebeat/syntax_error.yml` with content:

```
filebeat.prospectors:
- input_type: log
  paths:
    - /var/log/*.log
  exclude_files: [".gz\$"]
```

(the syntax error is in the last line, `\$` is not a valid escape character.)

breml:
Fixes #3572

### DIFF
--- a/filebeat/config/config.go
+++ b/filebeat/config/config.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"fmt"
 	"log"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


ruflin:
@breml Thanks for the fix.

@monicasarbu I think we should also have a changelog entry for this one?

monicasarbu:
I have opened a PR with the changelog changes: https://github.com/elastic/beats/pull/3577

PR #3575, opened 2017-02-10
Implement HTTP redirect caching for Elasticsearch output

rayharris-ibm:
Hi. I opened a topic on discuss.elastic.co about a week ago asking for feedback on interest in adding the ability to cache HTTP redirects in the elasticsearch output about a week ago. As I got no feedback, I'm going to go ahead and submit the PR. I've completed a CLA as rayharris-ibm/harrisra@us.ibm.com. And here is the link to the discuss topic:

https://discuss.elastic.co/t/add-ability-to-cache-http-redirects-in-elasticsearch-output/73901


libbeat/outputs/elasticsearch/client.go
```diff
@@ -185,6 +186,36 @@ func NewClient(
 		proxyURL:         s.Proxy,
 	}
 
+	if s.CacheRedirect {
+		client.Connection.http.CheckRedirect = func(req *http.Request, via []*http.Request) error {
+			if len(via) != 1 {
+				return nil
+			}
```
urso: why return `nil` on multiple redirects? we don't want to continue caching?

libbeat/outputs/elasticsearch/client.go
```diff
@@ -185,6 +186,36 @@ func NewClient(
...
+
+			if via[0].Method != "GET" {
+				return http.ErrUseLastResponse
```
urso: Why stop processing if method != "GET"? So output client tries to reconnect? In this case I wouldn't use  `ErrUseLastResponse` (which makes the client think no error occured, but check response code only). Returning an response with status 30x might actually lead to data loss if `PublishEvent` is executed.

libbeat/outputs/elasticsearch/client.go
```diff
@@ -185,6 +186,36 @@ func NewClient(
...
+
+			newClient, err := NewClient(s, onConnectCallback)
+
+			if err != nil {
+				logp.Err("Failed to Cache ES HTTP API Redirect to %s: %s", newURL, err)
+				return http.ErrUseLastResponse
+			}
+
+			client.bulkRequ = newClient.bulkRequ
+			client.Connection = newClient.Connection
```
urso: Why life patching the client instance with a new instance... looks kinda sketchy. 

The URL is stored in `client.Connection.URL`. I think you can savely overwrite the URL and continue with the redirect by returning `nil`...

libbeat/outputs/elasticsearch/client.go
```diff
@@ -185,6 +186,36 @@ func NewClient(
...
+
+			return http.ErrUseLastResponse
```
urso: same, using `ErrUseLastResponse` can lead to data-loss (which should be fixed, yeah)... I wonder if we want to enforce an error here (enforce client to ping endpoint before trying to send data) or just continue with redirect...

libbeat/outputs/elasticsearch/client.go
rayharris-ibm: I agree this is sketchy, but in my testing, I wasn't able to get the the redirects to be cached by just updating client.Connection.URL. I'll try again and provide results of the testing.

rayharris-ibm:
@urso Thanks for looking over the suggestion and commenting. Let me look over what you said and I'll respond shortly.

rayharris-ibm:
I rewrote the CheckRedirect function to be more clear about when it will cache the redirect. I also added a configuration setting to let the user control when they want sensitive headers forwarded with a redirect to a non-matching subdomain (see link in code comment). Golang doesn't do this by default.

libbeat/outputs/elasticsearch/config.go
```diff
@@ -15,6 +15,8 @@ type elasticsearchConfig struct {
 	Password         string             `config:"password"`
 	ProxyURL         string             `config:"proxy_url"`
 	LoadBalance      bool               `config:"loadbalance"`
+	CacheRedirect    bool               `config:"cache_redirect"`
```
ruflin: We should probably change this to:

```
redirect.cache.enabled
redirect.trust
```

libbeat/outputs/elasticsearch/client.go
```diff
@@ -128,7 +130,7 @@ func NewClient(
 		s.URL = u.String()
 	}
 
-	logp.Info("Elasticsearch url: %s", s.URL)
+	logp.Info("Elasticsearch UUUUurl: %s", s.URL)
```
ruflin: typo?

rayharris-ibm: 🤦‍♂️  yup

libbeat/outputs/elasticsearch/config.go
rayharris-ibm: will do

blakerouse:
@rayharris-ibm Sorry that we dropped the ball on getting this reviewed and landed. We are actively trying to clean up our pull requests, this seems like its very close and could possible be landed if we get it rebased with master. Would you be interested in picking this back up?

urso:
Unfortunately this PR is stalled for some time. Closing for now. We can re-open it, or create a new one if you decide you want to continue on it.

PR #3576, opened 2017-02-10
Pin sources.list to a single debian archive server

andrewkroh:
We are getting back “403  Forbidden [IP: 193.62.202.28 80]” when building the docker images. The server at 130.89.148.13 seems to be responding normally for the moment.


andrewkroh:
jenkins, package it

andrewkroh:
jenkins, package it

andrewkroh:
Beats-Package-PR passed: http://build-eu-00.elastic.co/view/Beats/job/beats-package-PR/77/

tsg:
@andrewkroh Seems that archive.debian.org is back up. Should we revert this PR?

andrewkroh:
Yeah, if it's working well now. Maybe we should push the images to the elastic docker registry so we don't have to rebuild them all the time?

PR #3581, opened 2017-02-13
Correct Libbeat docs directory layout

strootman:
The `/bin` was missing from the docs.


strootman:
Holy moly! 
I think this is the fastest one of my PRs has been approved!

Thanks @andrewkroh ;)

andrewkroh:
Thanks for the PR @strootman.

tsg:
Sorry I missed this one. I'm afraid it's not entirely correct. The `/bin/` part was indeed missing for deb and rpm, but the zip/tar.gz have the binary top-level, so there's no `/bin/` part.

dedemorton:
I'm adding the backport label to this PR, but it needs a forward port to master because the change was made in all branches but master.

Note that I'll also fix the issue where /bin was incorrectly added to the paths for zip/tar.gz.

PR #3583, opened 2017-02-14
Revert "Pin sources.list to a single debian archive server"

andrewkroh:
Reverts elastic/beats#3576. The hosts behind archive.debian.org are responding normally again.


andrewkroh:
I used the "Revert" button so the "revert-3576-bugfix/sources-list" branch should be deleted after merging.

jenkins, package it

PR #3589, opened 2017-02-14
Add check-extended command

ruflin:
This command checks the formatting of python files and if there are some misspellings which can be detected automatically.

Currently this command has to be run manually and is not part of CI. For autopep8 it should be discussed if it should be added. Problem with misspell can be, that it brings false positives. But there currently don't seem to be any such cases.

Note: misspell command seems to be quite slow when running on lots of files.

Simplify command was removed as it was forgotten in an older PR where simplify was removed from the other makefiles because it is obsolete.


Makefile
```diff
@@ -55,20 +57,26 @@ clean-vendor:
...
 	# Validate that all updates were committed
 	$(MAKE) update
 	git update-index --refresh
 	git diff-index --exit-code HEAD --
 
+# Corrects spelling errors
+.PHONY: misspell
+misspell:
+	go get github.com/client9/misspell
+	find . -type f -name '*' -not -path "*/vendor/*" -not -path "*/build/*" -not -path "*/.git/*" -exec misspell -w {} \;
```
tsg: since this `find` command is used in quite a few places, it might make sense to alias parts of it. Something like `FIND=find -not -path "*/vendor/*" -not -path "*/build/*" -not -path "*/.git/*"`

tsg:
+1 to have a `misspell` target but not run it automatically in the CI. We can review the list of misspellings also in [goreport](https://goreportcard.com/report/github.com/elastic/beats#misspell) (it might check only go files, though).

I think the autopep8 could be part of the CI check target. That way we avoid cleaning up things in unrelated PRs.

ruflin:
@tsg In the above PR, autopep8 is already part of the CI check (and fmt).

Makefile
ruflin: done

PR #3592, opened 2017-02-14
Added Filebeat Module overview and tutorial

tsg:
Went for documenting FBM in a tutorial fashion. I'll follow up with a PR to
create per-module docs.

Part of #3159.


tsg:
@dedemorton fyi, I started on the docs for the FBM. Happy to either get your comments here or as edits in a future PR.

filebeat/docs/modules.asciidoc
```diff
@@ -0,0 +1,182 @@
...
+  of the fields.
+
+* Sample Kibana dashboards, which can be used to visualize the log files.
+
+Filebeat automatically adjusts these configurations based on your environment
+and loads them to the respective Elastic stack components.
+
+NOTE: At the moment, Filebeat modules require using the Elasticsearch
+{elasticsearch}/ingest.html[Ingest Node]. In the future, Filebeat Modules will
+be able to also configure Logstash as a more powerful alternative to Ingest
```
ruflin: I would remove the Logstash part.

filebeat/docs/modules.asciidoc
```diff
@@ -0,0 +1,182 @@
...
+$ sudo bin/elasticsearch-plugin install ingest-geoip
+$ sudo bin/elasticsearch-plugin install ingest-user-agent
+----------------------------------------------------------------------
+
+If you are using an https://cloud.elastic.co/[Elastic Cloud] instance, you can
+enable the two plugins from the configuration page.
+
+This also assumes you have Nginx installed and writing logs in the default
+location and format. If you want to monitor another service for which a module
+exists, feel free to adjust the tutorial.
```
ruflin: for which a module exists, adjust the commands in the tutorial accordingly.

Otherwise it kind of sounds like someone should thanks this tutorial

filebeat/docs/modules.asciidoc
tsg: I hope that by the time we release this, we'll be able to say that we have sample Logstash configs which we can mention instead: https://github.com/elastic/logstash/issues/6542

Until then, and in case that doesn't happen on time, I think it's important to let people know that we do plan to allow for migrating from Ingest Node to LS.

filebeat/docs/modules.asciidoc
tsg: Good point, although I'll do the fix in https://github.com/elastic/beats/pull/3598 to simplify my rebases.

PR #3598, opened 2017-02-15
Per module docs in Filebeat

tsg:
* Each module has to provide a docs.asciidoc file to be included
  in the docs
* Following the MB model, these are collected in the `docs/` folder
  on `make update`
* Structure wise, I added a "Modules" part which has an Overview section
  and then a section for each module

Part of #3159.


filebeat/docs/modules-overview.asciidoc
```diff
@@ -0,0 +1,182 @@
...
+beta[]
+
+Filebeat modules simplify the collection, parsing, and visualization of common
+log formats.
+
+A typical module (say, for the Nginx logs) is composed of one ore
+more filesets (in the case of Nginx, `access` and `error`). A fileset contains
+the following:
+
+* Filebeat prospector configurations, which contain the default paths were to
```
ruflin: s/were/where

filebeat/docs/modules-overview.asciidoc
```diff
@@ -0,0 +1,182 @@
...
+accessible from Filebeat (see the <<filebeat-getting-started,getting started>>
+section). It also assumes that the Ingest Node GeoIP and User Agent plugins are
+installed, which you can do with the following two commands executed in the
+Elasticsearch home path:
+
+[source,shell]
+----------------------------------------------------------------------
+$ sudo bin/elasticsearch-plugin install ingest-geoip
+$ sudo bin/elasticsearch-plugin install ingest-user-agent
+----------------------------------------------------------------------
```
ruflin: should we mention that ES must be restarted afterwards? (or is this not the case?)

filebeat/module/nginx/_meta/docs.asciidoc
```diff
@@ -0,0 +1,45 @@
...
+
+This module comes with a sample dashboard.
+
+image::./images/kibana-nginx.png[]
+
+[float]
+=== Access logs fileset settings
+
+[float]
+==== var.paths
```
ruflin: As this exists for all modules / filesets, should we describe it somewhere global?

tsg: It's technically not a requirement to have it, although I suspect all filesets will have it. I'm not sure, I think this creates duplication, on the other hands is more convenient for the user to find all options in one place.

filebeat/docs/modules-overview.asciidoc
tsg: Added a note.

filebeat/module/nginx/_meta/docs.asciidoc
ruflin: Agree, probably better for the user. It will be trickier to keep it up-to-date.

Issue #2425, opened 2016-08-31
topBeat install from repositories links to filebeat install


chrisribe:
This is simply put confusing...

The documentation for topbeat install:
https://www.elastic.co/guide/en/beats/topbeat/1.3/topbeat-installation.html

Links to the libbeat main setup document:
https://www.elastic.co/guide/en/beats/libbeat/current/setup-repositories.html

But those instructions contain apt-get install **filebeat** NOT apt-get install **topbeat**

Great product, but the documentation is not simple to follow. It's like bits of docs but not all linked together properly.

dedemorton:
Thanks for taking time to give us feedback about the documentation. The topic about installing from our repositories is meant to describe how to install any Beat. We show Filebeat as an example, but I can see why the current structure might be confusing. Over time, I've been moving a lot of the shared libbeat content into the reference guides for each Beat. It sounds like I need to do the same with this topic. 

I'm also working on some changes for the upcoming 5.0 release that will hopefully address your concerns and make it easier for you to find content. 

Please let me know if you have specific feedback regarding how the docs are linked together (and where links are missing).  

Thanks for helping us improve the documentation.

dedemorton:
Sorry it took me awhile to get around to this. Topbeat is legacy now, but I've updated the other Beats docs in #3601.

PR #3601, opened 2017-02-16
Make topic about installing from repositories shared

dedemorton:
This PR makes the content about installing from repositories shared so that the content appears in the all the Beats guides (except Winlogbeat).

This PR is meant to clear up the confusion for users who are reading the getting started content and follow the link that takes them to the platform reference where the doc mentions Filebeat. See https://github.com/elastic/beats/issues/2425.

I've kept the topic in the platform reference (but set the beatname variables so that the content is more generic) because IMO there is some value in keeping that content in the platform reference. 

Right now the product pages point to the topic in the platform reference. We need to decide if we'd rather have the product pages link to the topics in the individual guides. WDYT?

### DIFF
--- a/filebeat/docs/getting-started.asciidoc
+++ b/filebeat/docs/getting-started.asciidoc
@@ -30,7 +30,7 @@ mac>> for OS X, and <<win, win>> for Windows).
 
 [NOTE]
 ==================================================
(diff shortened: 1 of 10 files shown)


ruflin:
@dedemorton Merged as is as I think this is already an improvement. I see it as beneficial for the user experience, if we have it in the future the product page links linking to the individual pages.

PR #3603, opened 2017-02-16
Implement index template generation in Golang

ruflin:
The index template generation is moved to golang.

Changes
* Improve defaults as they were (accidentally?) not used before as the properties were always overwritten by fresh properties
* Use `go run` to execute command
* Template generation is part of libbeat, only execution command is outside as it as has some additional logic on how to read the files.
* Add assumption that $ES_BEATS is always a relative path
* Add support for more fine grained versioning in template generation
* Remove python script as not needed anymore

Blocked by https://github.com/elastic/beats/pull/3655. Needs update after it is merged.


filebeat/filebeat.template-es2x.json
```diff
@@ -40,95 +43,123 @@
                 "body_sent": {
                   "properties": {
                     "bytes": {
+                      "doc_values": true,
+                      "index": true,
```
tsg: Hmm, I think `"index": true` is not valid for 2x. It should be `"index": "analyzed"`, or simply leave the default.

filebeat/filebeat.template-es2x.json
```diff
@@ -24,13 +24,16 @@
       ],
       "properties": {
         "@timestamp": {
+          "doc_values": true,
+          "index": true,
```
tsg: I'm not sure if this indexing the `@timestamp` makes sense.

packetbeat/packetbeat.template-es2x.json
```diff
@@ -37,59 +40,79 @@
...
               "type": "string"
             },
             "content-encoding": {
+              "doc_values": true,
               "ignore_above": 1024,
               "index": "not_analyzed",
               "type": "string"
             },
             "content-type": {
+              "doc_values": true,
```
tsg: Since we almost never want to disable "doc_values", I wonder if it's worth adding it explicitly to the templates.

filebeat/filebeat.template-es2x.json
ruflin: Ok, lets remove it from the defaults then.

packetbeat/packetbeat.template-es2x.json
ruflin: As long as the defaults stay the same, there is no need to add it. Lets remove this too for the moment to keep it the same as it was previously and introduce defaults in case we need it. Defaults "should have been there" in the python script but were ignored ...

filebeat/filebeat.template-es2x.json
```diff
@@ -535,6 +535,7 @@
   },
   "order": 0,
   "settings": {
+    "index.mapping.total_fields.limit": 10000,
```
tsg: This is probably ignored by ES 2.x, but I guess it's harmless.

libbeat/scripts/Makefile
```diff
@@ -249,15 +249,19 @@ update: python-env collect
...
+
+	go run ${ES_BEATS}/dev-tools/cmd/index_template/index_template.go -files ${PWD}/${ES_BEATS}/libbeat/_meta/fields.generated.yml,${BEAT_GOPATH}/src/${BEAT_PATH}/_meta/fields.yml,${BEAT_GOPATH}/src/${BEAT_PATH}/_meta/fields.generated.yml -es-version 2.0.0 -beatname ${BEAT_NAME} -output ${BEAT_GOPATH}/src/${BEAT_PATH}/${BEAT_NAME}.template-es2x.json
 
 	# Generate index-pattern
 	echo "Generate index pattern"
 	-rm -f $(PWD)/_meta/kibana/index-pattern/${BEAT_NAME}.json
 	mkdir -p $(PWD)/_meta/kibana/index-pattern
 	. ${PYTHON_ENV}/bin/activate && python ${ES_BEATS}/libbeat/scripts/generate_index_pattern.py --index '${BEAT_NAME}-*' --libbeat ${ES_BEATS}/libbeat --beat $(PWD)
 
+cc:
```
tsg: What's this target for?

ruflin: argh, testing target left over. will remove it

tsg:
There seem to be some unit test failures, atm.

filebeat/filebeat.template-es2x.json
ruflin: Will only apply it if not 2x as we have the logic for that. We should also make use of it :-)

ruflin:
I rebased and squashed now. Test failures were do to changes in the defaults.

monicasarbu:
LGTM

dev-tools/cmd/index_template/index_template.go
```diff
@@ -0,0 +1,56 @@
...
+// Generates index templates for the beats
+//
+// The command which can be used in the Makefile looks as following
+//
+//   go run ${ES_BEATS}/dev-tools/cmd/index_template/index_template.go -files ${PWD}/${ES_BEATS}/libbeat/_meta/fields.generated.yml,${BEAT_GOPATH}/src/${BEAT_PATH}/_meta/fields.yml,${BEAT_GOPATH}/src/${BEAT_PATH}/_meta/fields.generated.yml -es-version 5.0.0 -beatname ${BEAT_NAME} -output ${BEAT_GOPATH}/src/${BEAT_PATH}/${BEAT_NAME}.template.json
+
+func main() {
+
+	version := flag.String("es-version", "", "Elasticsearch version")
+	inputFiles := flag.String("files", "", "List of files, comma seperated. This files must be passed with the full path.")
```
andrewkroh: I think it would be more conventional to "args" instead of a flag for specifying files. https://golang.org/pkg/flag/#Args

dev-tools/cmd/index_template/index_template.go
```diff
@@ -0,0 +1,56 @@
...
+
+// Generates index templates for the beats
+//
+// The command which can be used in the Makefile looks as following
+//
+//   go run ${ES_BEATS}/dev-tools/cmd/index_template/index_template.go -files ${PWD}/${ES_BEATS}/libbeat/_meta/fields.generated.yml,${BEAT_GOPATH}/src/${BEAT_PATH}/_meta/fields.yml,${BEAT_GOPATH}/src/${BEAT_PATH}/_meta/fields.generated.yml -es-version 5.0.0 -beatname ${BEAT_NAME} -output ${BEAT_GOPATH}/src/${BEAT_PATH}/${BEAT_NAME}.template.json
+
+func main() {
+
+	version := flag.String("es-version", "", "Elasticsearch version")
```
andrewkroh: It would be nice to say what the default value is here. Could you call `beat.GetDefaultVersion()` here to populate the default value so that it shows up in the help.

dev-tools/cmd/index_template/index_template.go
```diff
@@ -0,0 +1,56 @@
...
+	inputFiles := flag.String("files", "", "List of files, comma seperated. This files must be passed with the full path.")
+	beatName := flag.String("beatname", "", "Base index name. Normally {beatname}")
```
andrewkroh: Is a value required? Probably it is so I would mention that it's required in the description and validate that the value is non-empty.

dev-tools/cmd/index_template/index_template.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	// Make it compatible with the sem versioning
+	if *version == "2x" {
+		*version = "2.0.0"
+	}
+
+	templateString, err := template.GetTemplate(*version, *beatName, existingFiles)
+	if err != nil {
+		fmt.Printf("Error generating template: %+v", err)
```
andrewkroh: For any prints or logs I would route them to stderr. `fmt.Fprinf(os.Stderr, ...)`

libbeat/template/fields.go
```diff
@@ -0,0 +1,60 @@
...
+		case "text":
+			mapping = field.text()
+		case "keyword":
+			mapping = field.keyword()
+		case "object":
+			mapping = field.object()
+		case "array":
+			mapping = field.array()
+		case "group":
+			var newPath = ""
```
andrewkroh: Use `var newPath string` instead.

libbeat/template/template.go
```diff
@@ -0,0 +1,119 @@
...
+		// Metricbeat exceeds the default of 1000 fields
+		basicStructure.Put("settings.index.mapping.total_fields.limit", defaultTotalFieldsLimit)
+	}
+
+	return basicStructure
+}
+
+func loadYaml(path string) (Fields, error) {
+	keys := []Field{}
+	cfg, err := common.LoadFile(path)
```
andrewkroh: You probably should use this method since it expects to be loading config files and checks the permissions on the files it reads.

Can you use `yaml.Unmarshal` more directly?

dev-tools/cmd/index_template/index_template.go
ruflin: added

dev-tools/cmd/index_template/index_template.go
ruflin: done

dev-tools/cmd/index_template/index_template.go
ruflin: done, same for beatname

dev-tools/cmd/index_template/index_template.go
ruflin: done

libbeat/template/fields.go
ruflin: done

libbeat/template/template.go
ruflin: I changed it to 

```
cfg, err := yaml.NewConfigWithFile(path)
	if err != nil {
		return nil, err
	}
	cfg.Unpack(&keys)
```

I really like to use go-ucfg here in case we add some config magic to this in the future.

andrewkroh: Yeah, that's better since it bypasses the file mode check.

PR #3606, opened 2017-02-16
Add link to iobeat.

pysysops:
Add link to iobeat which reads and publishes IO statistics from /proc/diskstats on Linux.

### DIFF
--- a/libbeat/docs/communitybeats.asciidoc
+++ b/libbeat/docs/communitybeats.asciidoc
@@ -27,6 +27,7 @@ https://github.com/YaSuenag/hsbeat[hsbeat]:: Reads all performance counters in J
 https://github.com/christiangalsterer/httpbeat[httpbeat]:: Polls multiple HTTP(S) endpoints and sends the data to
 Logstash or Elasticsearch. Supports all HTTP methods and proxies.
 https://github.com/jasperla/hwsensorsbeat[hwsensorsbeat]:: Reads sensors information from OpenBSD.
+https://github.com/devopsmakers/iobeat[iobeat]:: Reads IO stats from /proc/diskstats on Linux.
 https://github.com/radoondas/jmxproxybeat[jmxproxybeat]:: Reads Tomcat JMX metrics exposed over 'JMX Proxy Servlet' to HTTP.
 https://github.com/mheese/journalbeat[journalbeat]:: Used for log shipping from systemd/journald based Linux systems.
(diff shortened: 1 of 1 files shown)


ruflin:
@pysysops Thanks for creating iobeat. This could be also an interesting metricset in the system module for metricbeat. Let us know if you would be interested to contribute it.

pysysops:
Hi Nicolas, It'd be great to contribute to metricbeat.I'd like to get my current effort up to scratch and get some Windows, FreeBSD and Mac (Darwin) functionality in there.I'll take a look over the existing metricsets and get some unit tests etc.. in place.
On 16 February 2017 at 22:11 Nicolas Ruflin <notifications@github.com> wrote:@pysysops Thanks for creating iobeat. This could be also an interesting metricset in the system module for metricbeat. Let us know if you would be interested to contribute it.

ruflin:
@pysysops Awesome. Just ping me if you need some help.

andrewkroh:
We do have the [system diskio metricset](https://www.elastic.co/guide/en/beats/metricbeat/current/metricbeat-metricset-system-diskio.html) in Metricbeat that is collecting some IO counters. It's not implemented for Darwin; that would be nice to have. It would also be nice to see if there are any differences in what Metricbeat provides compared to iobeat.

pysysops:
iobeat is was something that I tried to put together a long time ago. Beats was very young at the time and I knew next to nothing about golang. Now I understand a little more I thought I'd get it working in some way.

I'll take a look at the code and need to become familiar with the available metric sets in general. If I can squeeze metrics from MacOS I'll open an appropriate PR or issue to start a discussion. 

Feel free to revert the addition of iobeat as it's unnecessary. Thanks.

ruflin:
I think there is some overlap between the diskio metricset and iobeat, but it seems iobeat is more `diskstats` metricset as it provides more details stats about the disk itself.

pysysops:
From what I can see, https://github.com/shirou/gopsutil (which is an awesome attempt at bringing psutil to Go) does the same thing and reads in /proc/diskstats on Linux. They also appear to have Darwin / MacOS support which is interesting...

ruflin:
We already use gopsutil for the diskio module: https://github.com/elastic/beats/blob/master/metricbeat/module/system/diskio/diskio.go#L32 But it seems to be missing some stats like ios in progress or write merges. Perhaps it even makes sense extend gopsutil with these and then get them into metricbeat.

PR #3611, opened 2017-02-17
Add `make notice` to `make check`

tsg:
This way it's checked automatically by the CI and we keep the NOTICE
file always up to date.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,7 @@ coverage-report:
 	go tool cover -html=./${COVERAGE_DIR}/full.cov -o ${COVERAGE_DIR}/full.html
 
 .PHONY: update
-update:
+update: notice
 	$(foreach var,$(PROJECTS),$(MAKE) -C $(var) update || exit 1;)
 
 .PHONY: clean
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -64,6 +64,7 @@ check: python-env
 	${FIND} -name *.py -exec autopep8 -d --max-line-length 120  {} \; | (! grep . -q) || (echo "Code differs from autopep8's style" && false)
 	# Validate that all updates were committed
 	$(MAKE) update
+	$(MAKE) notice
```
ruflin: Should we also add `$(MAKE) notice` to the beats `$(MAKE) update`? So it will also be run with the update command.

tsg: I was thinking the same. The advantage would be that we don't waste time finding the issue when CI runs. The disadvantage is that it's executed on every `make update`. But it seems that the script is very quick, so I'm going to change it.

ruflin: Yes, and as it only runs on the global `make update` which already does a lot of stuff, it I think it will not make a difference in speed.

tsg: I did the change.

ruflin:
@tsg The new check breaks the build. So we know it is working? 😆

PR #3612, opened 2017-02-17
Fix docker hanging when container killed

### LINKED ISSUE #3610: Metricbeat stops reporting data if container is killed
Steps to reproduce:

* Config
```
- module: docker
  metricsets: ["cpu"]
  period: 1s
```

* Start a container like `docker run -it debian bash` and wait until the container is picked up by metricbeat. 
* Use `docker kill container-id` to kill the container.

Metricbeat will stop reporting data and shutdown will be blocked.

This was first reported in https://github.com/elastic/beats/issues/3580. I opened a separate issue as in the other issue, 2 issues are mentioned.

ruflin:
No timeout was passed to the docker client. It seems in case of a killed container it can happen that the connection is hanging. To interrupt this connection, the timeout from the metricset is passed to the client. That means in case info for a container cannot be fetched, it will timeout.

This change requires that the docker module is not run with a timeout of 3s seconds, which indirectly means a period of 3s. The reason is that already the http request waits ~2s for the response. So if 1s is set as timeout, all requests will timeout.

Further changes:

* Containers without names will be ignored, as these are containers for which the data could not be fetched.
* Period was set to 1s by default instead of the period as document. This was changed.
* Add documentation node about minimal period.

Closes #3610

The issue with this PR was introduce in 5.2.1 by fixing the memory leak. Before go routines just piled up, but now they caused filebeat to hang.

This needs also backport to 5.2.2

### DIFF
--- a/metricbeat/docs/modules/system.asciidoc
+++ b/metricbeat/docs/modules/system.asciidoc
@@ -49,6 +49,9 @@ metricbeat.modules:
   cpu_ticks: true
 ----
 
+It is strongly recommended to not run docker metricsets with a period smaller then 3 seconds. The request to the docker
+API already takes up to 2s seconds. Otherwise all the requests would timeout and no data is reported.
+
 [float]
 === Dashboard
 
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -49,6 +49,8 @@ https://github.com/elastic/beats/compare/v5.1.1...master[Check the HEAD diff]
 *Metricbeat*
 
 - Fix go routine leak in docker module. {pull}3492[3492]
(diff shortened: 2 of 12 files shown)


metricbeat/docs/modules/system.asciidoc
```diff
@@ -49,6 +49,9 @@ metricbeat.modules:
   cpu_ticks: true
 ----
 
+It is strongly recommend to not run docker metricsets with a period smaller then 3 seconds. The request to the docker
```
tsg: "It is strongly recommended"

ruflin: fixed

PR #3615, opened 2017-02-17
Persist beat UUID

urso:
read/write file ${path.data}/meta.json including UUID (json document) upon
startup to persist a beats generated UUID among multiple restarts.


urso:
Added system test

PR #3616, opened 2017-02-17
Filebeat modules dev guide

tsg:
Basic guide for creating a new module.

Part of #3159.

### DIFF
--- a/filebeat/docs/modules-dev-guide.asciidoc
+++ b/filebeat/docs/modules-dev-guide.asciidoc
@@ -1,0 +1,304 @@
+[[filebeat-modules-devguide]]
+== Developer guide: creating a new module
+
+This guide will walk you through creating a new Filebeat module.
+
+All Filebeat modules currently live in the main
+https://github.com/elastic/beats[Beats] repository. To clone the repository and
+build Filebeat (which you will need for testing), please follow the general
+Beats https://github.com/elastic/beats/blob/master/CONTRIBUTING.md[CONTRIBUTING]
+guide.
+
+=== Overview
+
+Each Filebeat module is composed by one or more "filesets". We usually create a
+module for each service that we support (`nginx` for Nginx, `mysql` for Mysql,
+etc.) and a fileset for each type of logs that that service creates. For
+example, the Nginx module has `access` and `error` filesets. You can contribute
+a new module (with at least one fileset) or only a new fileset into an existing
+module.
+
+=== Creating a new fileset
+
+Regardless of whether you are creating a fielset in a new module or in an
+existing module, the procedure is similar. Run the following command in the
+`filebeat` folder:
+
+[source,bash]
+----
+make create-metricset
+----
+
+You'll be prompted to enter a module and metricset name. Only use characters
+`[a-z]` and, if required, underscores (`_`). No other characters are allowed.
+For the module name, you can us either a new module name or an existing module
+name. If the module doesn't exist, it will be created.
+
+In this guide we use `{fileset}` and `{module}` as placeholders for the fileset
+and module names. You need to replace these with the actual names you entered in
+the previous command.
+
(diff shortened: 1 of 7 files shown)


filebeat/docs/modules-dev-guide.asciidoc
```diff
@@ -0,0 +1,304 @@
+[[filebeat-modules-devguide]]
+== Developer guide: creating a new module
+
+This guide will walk you through creating a new Filebeat module.
+
+All Filebeat modules currently live in the main
+https://github.com/elastic/beats[Beats] repository. To clone the repository and
+build Filebeat (which you will need for testing), please follow the general
+Beats https://github.com/elastic/beats/blob/master/CONTRIBUTING.md[CONTRIBUTING]
```
ruflin: Not only related to this guide, but we should probably have contributing as part of the docs to also have it specific to a version.

filebeat/docs/modules-dev-guide.asciidoc
```diff
@@ -0,0 +1,304 @@
...
+      - /usr/local/example/test.log*
+      - /example/test.log*
+    os.windows:
+      - c:/programdata/example/logs/test.log*
+----
+
+There's quite a lot going on in the above, so let's break it down:
+
+* The name of the variable is `paths` and the default value is
+  `["/example/test.log*"]`.
```
ruflin: It could be confusing that here the compact array form is used, but above it the example not. For people familiar with Yaml, it is obvious.

filebeat/docs/modules-dev-guide.asciidoc
```diff
@@ -0,0 +1,304 @@
...
+  Filebeat is executed on the respective OS.
+
+Besides the variable definition, the `manifest.yml` file also contains
+references to the ingest pipeline and prospector configuration to use (see next
+sections):
+
+[source,yaml]
+----
+ingest_pipeline: ingest/pipeline.json
+prospectors: config/testfileset.yml
```
ruflin: Just realising it now that this should probably be singular `prospector` as now only one prospector can be defined.

filebeat/docs/modules-dev-guide.asciidoc
```diff
@@ -0,0 +1,304 @@
...
+paths:
+{{ range $i, $path := .paths }}
+ - {{$path}}
+{{ end }}
+exclude_files: [".gz$"]
+multiline:
+  pattern: "^# User@Host: "
+  negate: true
+  match: after
+input_type: log
```
ruflin: input_type: log exists twice

filebeat/docs/modules-dev-guide.asciidoc
```diff
@@ -0,0 +1,304 @@
...
+==== _meta/kibana
+
+This folder contains the sample Kibana dashboards for this module. To create
+them, you can build them visually in Kibana and then run the following command:
+
+
+[source,shell]
+----
+$ cd filebeat/module/{module}/
+python ../dev-tools/export_dashboards.py --regex {module} --dir _meta/kibana
```
ruflin: I think the path to the dev-tools needs a few more `../`

filebeat/docs/modules-dev-guide.asciidoc
tsg: Good catch, that's wrong here in the docs.

filebeat/docs/modules-dev-guide.asciidoc
tsg: Yeah, we should probably move that to the docs.

filebeat/docs/modules-dev-guide.asciidoc
tsg: I rephrased it to avoid having to use the compact form.

PR #3619, opened 2017-02-20
logbeat init

needkane:


ruflin:
@needkane Working on logbeat? Interested to hear more :-)

PR #3620, opened 2017-02-20
Use a reliable image for integration test

tsouza:


metricbeat/module/php_fpm/_meta/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM richarvey/nginx-php-fpm:1.1.2
+FROM tsouza/nginx-php-fpm
```
ruflin: Can you also tag the image? So in case we change it in the future, it will not break older releases.

tsouza: Ok, I'll do it, sorry!

tsouza: I'll tag based on php version. Sounds good?

ruflin: SGTM. Interestingly @richarvey did that also previously and then switched to an other versioning system. Would be interesting to know why.

tsouza: Ok! I'll ping him, maybe he answers

ruflin:
@tsouza Thanks for creating the container and the PR.

PR #3621, opened 2017-02-20
Filebeat modules: Fix generator files

tsg:
There were some changes in the structure of some files, and this is
needed to update the generator as well.

Part of #3159.


ruflin:
@tsg I assume that means we don't have any tests for this generator type. Will see if I can add one.

tsg:
@ruflin yeah, that's correct.

Issue #2887, opened 2016-10-30
Add timezone metadata


joshuaspence:
There are a few log formats that I am parsing with Filebeat + Logstash in which the logs contain a timestamp but no timezone. For example, here is a log event from `nginx` error logs:

```
2016/10/30 19:40:06 [error] 10522#10522: *42075309 broken header: "" while reading PROXY protocol, client: x.x.x.x, server: 0.0.0.0:443
```

In all of the cases I have seen, the timestamp is logged in the local timezone (that is, the timezone of the host on which Filebeat is running). The problem, however, is that Logstash doesn't have enough information from which the host's timezone can be inferred. Consequently, Logstash ingests the log event using the local timezone for the host that is running Logstash (which may or may not be the same as the timezone of the host which is running Filebeat).

Obviously, it would be great if all logs could contain enough information such that the timestamp is unambigious, but this is overly ambitious. Instead, it would be great if Filebeat could add a `[beat][timezone]` field to all log events processed. This field could then be used in Logstash with the `date` filter.

ruflin:
A potential solution here could be that you write the timezone of your server to an environment variable and add it as one of the `fields` variable to the config file. 

```
fields: 
- timezone: ${TIMEZONE}
```

Like this you will have the timezone information in each event.

def324:
I'd love to have this feature as well. The problem with the environment variable is that it would break if the server timezone gets updated. On the other hand if beats handles it natively it could always send the current up to date timezone.

ruflin:
@def324 Can you share the use case where the TIMEZONE of a server would change of its lifetime?

def324:
I had a situation where we consolidated servers that were previously managed by different departments. Some preferred to have their servers in EST, some in UTC. We ended up changing them all to UTC to bring them to one standard.

I guess one way would be to write the TIMEZONE variable on startup or on start of the beat, but I'm thinking having it integrated would be a cleaner approach.

monicasarbu:
We [added](https://github.com/elastic/beats/pull/3902) support for timezone in all Beats to export `beat.timezone` in case the `add_locale` processor is configured:

```
processors:
 - add_locale:
```

PR #3622, opened 2017-02-20
Add local timestamp to event

ohadravid:
Hi!
I want to add a tiny change to the wineventlog. 
Currently only the UTC time (which is saved in the event xml) is included in the event. 
It is very useful to also get the event time in the local timezone of the computer.
This is the same behavior as the Windows Event Viewer (shows the event local time).

image

### DIFF
--- a/winlogbeat/eventlog/eventlog.go
+++ b/winlogbeat/eventlog/eventlog.go
@@ -10,6 +10,7 @@ import (
 	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/logp"
 	"github.com/elastic/beats/winlogbeat/sys"
+	"time"
 )
 
 // Debug selectors used in this package.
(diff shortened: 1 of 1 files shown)


andrewkroh:
> It is very useful to also get the event time in the local timezone of the computer.

How are you using the local timestamp? In what ways is it more useful than UTC timestamp?

ohadravid:
For example, you can build a query to check if a user tried to login into the machine in a strange time (say 02:00AM). For this, you have to know what the local time of the machine is (and the network might span a number of timezones, so we can't assume to know the timezone of a specific machine).

Another example is monitoring heavy background processes which are configured to run only at night, and we want to make sure they run at the configured time correctly.

andrewkroh:
Those are similar to the reasons I was thinking this would be useful information.

We had another request for timezone info in #2887. I think we should consider a more general approach (in libbeat) that allows this to be added in any Beat where a user wants it (as determined via configuration).

@elastic/beats WDYT about an optional local timestamp?

ruflin:
@andrewkroh +1 on having a "processor" or something similar for this. Then we can nest it under its own namespace and have it extendable with additional info.

monicasarbu:
@ohadravid Thank you for your PR. It seems that there is an interest in finding the local timestamp. We had a short discussion between us and decided that a better solution would be to export the `timezone` together with the `@timestamp` in UTC, in order to find the local timestamp.

PR #3625, opened 2017-02-20
Docs should refer to correct version being used for testing

tsouza:


tsg:
@tsouza You need to run `make update` in the Metricbeat folder to fix the failing test.

PR #3627, opened 2017-02-20
Optionally use the GH API to open PRs from the cherrypick_pr script

tsg:
If you add the `--create_pr` parameter, and have a GH token in
`~/.github_token`, the script will also create a PR for you, with the right
title and tags already set.

### DIFF
--- a/dev-tools/cherrypick_pr
+++ b/dev-tools/cherrypick_pr
@@ -1,17 +1,20 @@
 #!/usr/bin/env python
 import sys
+import json
 import argparse
+from os.path import expanduser
 from subprocess import check_call, call, check_output
+import requests
 
 """
 Example usage:
 
-./dev-tools/cherrypick_pr 5.0 2565 6490604aa0cf7fa61932a90700e6ca988fc8a527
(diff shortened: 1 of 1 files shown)


dev-tools/cherrypick_pr
```diff
@@ -91,9 +98,39 @@ def main():
...
+            print("Creating PR failed: {}".format(r.json()))
+            sys.exit(1)
+        new_pr = r.json()
+
+        # add labels
+        s.post(base+"/issues/{}/labels".format(new_pr["number"]),
+               json=["backport", "review"])
+
+        print("\nDone. PR created: {}".format(new_pr["html_url"]))
+        print("Please go and check it and add the review tags")
```
ruflin: We could later probably even add tags `review` and `backport`.

tsg: We do! Hmm, damn, I forgot to adjust the print line.

ruflin: Lol, I missed it 3 lines above :-(

PR #3635, opened 2017-02-21
Make packetbeat and filebeat better compatible with community beats

ruflin:
Packetbeat and filebeat can be extended through protocols / modules. This requires both to be used as library if they are extended in their own beat. This is a first step to make it better compatible with this by adjusting import paths to depend on ES_BEATS.


tsg:
jenkins, package it

PR #3639, opened 2017-02-21
Report cwd

mcenirm:
I need metricbeat's system process metricset to report current working directory (cwd).

It was straightforward to use gosigar's ProcExe, at least for linux. I had to skip the test on darwin since gosigar doesn't get cwd, and libproc.h is not a public interface anyway. I have not checked freebsd and windows.

### DIFF
--- a/metricbeat/module/system/process/helper.go
+++ b/metricbeat/module/system/process/helper.go
@@ -49,6 +49,11 @@ func newProcess(pid int) (*Process, error) {
 		return nil, fmt.Errorf("error getting process state for pid=%d: %v", pid, err)
 	}
 
+	exe := sigar.ProcExe{}
+	if err := exe.Get(pid); err != nil {
+		return nil, fmt.Errorf("error getting process exe for pid=%d: %v", pid, err)
+	}
+
 	proc := Process{
 		Pid:      pid,
 		Ppid:     state.Ppid,
@@ -57,6 +62,7 @@ func newProcess(pid int) (*Process, error) {
 		State:    getProcState(byte(state.State)),
 		Username: state.Username,
 		Ctime:    time.Now(),
+		Cwd:      exe.Cwd,
 	}
 
 	return &proc, nil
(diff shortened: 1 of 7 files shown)


mcenirm:
I don't understand the failures under AppVeyor, especially the [one for filebeats](https://ci.appveyor.com/project/elastic-beats/beats/build/8543/job/yd8p2yxue85w5nwq#L173), which has no changes from this pull request.

I originally used tag v5.2.1 (elastic/beats@e669c2080d02af8dce9e9ae5918c320fef6b8540) as the start for my branch, since that was the latest release at the time. Later, when created the pull request, GitHub did not let me choose a tag for the comparison.

Which branch is a good starting point for retrying the pull request?

metricbeat/module/system/process/helper.go
```diff
@@ -48,6 +49,11 @@ func newProcess(pid int) (*Process, error) {
 		return nil, fmt.Errorf("error getting process state for pid=%d: %v", pid, err)
 	}
 
+	exe := sigar.ProcExe{}
+	if err := exe.Get(pid); err != nil {
```
andrewkroh: This will require more error handling. Some platforms do not implement `sigar.ProcExe` and they should return `ErrNotImplemented`. We don't want to break all process metric collection if this error occurs. It would be good to double check the implementations in gosigar to see if there are other errors that can occur in normal operation (like a permission error if metricbeat is run by a non-root user).

metricbeat/module/system/process/helper.go
```diff
@@ -172,6 +179,7 @@ func (procStats *ProcStats) GetProcessEvent(process *Process, last *Process) com
 		"name":     process.Name,
 		"state":    process.State,
 		"username": process.Username,
+		"cwd":      process.Cwd,
```
andrewkroh: I think `cwd` should only be added if it's non-empty.

ruflin:
@andrewkroh @mcenirm I suggest to open this PR again against master. New features should not be added in existing branches, only bugfixes if possible.

PR #3646, opened 2017-02-22
Cherry-pick to 5.3: Fix permissions on module files

### LINKED ISSUE #3644: Modules permission error on default DEB installation
Found in the 5.3 build candidate:

```
2017/02/22 10:51:40.953051 beat.go:339: CRIT Exiting: Error reading fileset system/syslog: Error reading manifest file: config file ("/usr/share/filebeat/module/system/syslog/manifest.yml") can only be writable by the owner but the permissions are "-rw-rw-r--"
Exiting: Error reading fileset system/syslog: Error reading manifest file: config file ("/usr/share/filebeat/module/system/syslog/manifest.yml") can only be writable by the owner but the permissions are "-rw-rw-r--"
2017/02/22 10:51:40.949540 metrics.go:23: INFO Metrics logging every 30s
```

This seems to only be an issue in the DEB files.

tsg:
Cherry-pick of PR #3645 to 5.3 branch. Original message: 

They contain configuration files, so they shouldn't be group writable,
which was the case before this patch.

Fixes #3644.

### DIFF
--- a/libbeat/scripts/Makefile
+++ b/libbeat/scripts/Makefile
@@ -327,6 +327,7 @@ install-home:
 	if [ -d _meta/module.generated ]; then \
 		install -d -m 755 ${HOME_PREFIX}/module; \
 		rsync -av _meta/module.generated/ ${HOME_PREFIX}/module/; \
+		chmod -R go-w _meta/module.generated; \
 	fi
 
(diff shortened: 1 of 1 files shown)


tsg:
jenkins, package it

PR #3652, opened 2017-02-23
Use Docker Healthcheck to wait for services to start

andrewkroh:
Let Docker manager the start order of containers based on the dependencies declared
in the docker-compose configuration files. The removes the need to have custom
Docker entrypoint scripts that test services and wait for them to start before proceeding.

- Added 127.0.0.1 to the ports in local.yml so that the containers cannot be connected to remotely.

- Updated to Elastic stack 5.2.1 for latest.yml environment.

- Use docker-compose 1.11 on Travis CI.

- Remove docker-entrypoint.sh scripts.


metricbeat/module/apache/_meta/Dockerfile
```diff
@@ -1,2 +1,4 @@
 FROM httpd:2.4.20
+RUN apt-get update && apt-get install -y curl
```
ruflin: That will slow down the builds quite a bit if we have it several times. But I don't have a better solution.

testing/environments/docker/kafka/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM spotify/kafka
```
ruflin: I'm undecided if these files belong into libbeat or here in the "general" environment area. Reason for keeping it in libbeat is as they are only used there. Reason for having it here is that our "manual testing" setup should also have kafka available. Same for redis.

testing/environments/local.yml
```diff
@@ -2,27 +2,31 @@
...
   kibana:
     ports:
-      - "5601:5601"
-    environment:
-      - ELASTICSEARCH_URL=http://elasticsearch:9200/
+      - "127.0.0.1:5601:5601"
 
   elasticsearch:
     ports:
-      - "9200:9200"
```
ruflin: Is the localhost addition needed?

ruflin:
jenkins, retest it

ruflin:
jenkins, retest it

ruflin:
jenkins, retest it

testing/environments/local.yml
andrewkroh: It prevents docker from binding to 0.0.0.0 for the forwarded ports. And this prevents someone from connecting remotely to your Dockerized services. This will be good for security 🔒 .

testing/environments/docker/kafka/Dockerfile
andrewkroh: Yeah, probably we could add a testing/environment compose file for kafka and redis that uses these for when you want to do manual testing.

andrewkroh:
@ruflin Thanks for getting Jenkins updated.

I updated the docker and docker-compose version requirements in the CONTRIBUTING.md.

testing/environments/docker/kafka/Dockerfile
ruflin: Lets do that in a follow up PR. We can do that now easily as the builds do not automatically start all images from the testing environment anymore, but each beat picks the once he needs. Like this we could also share kafka setup between metricbeat and libbeat.

Issue #3537, opened 2017-02-06
Filebeat Fails After Power Failure


andynz2017:
Issue: Filebeat running on CentOS Linux release 7.3.1611 (Core), filebeat 5.1. Server suffers power failure due to host failing. Ungraceful shutdown, VM running off remote QNAP via ISCSI. VM recovers fine but filebeat doesn't start. 

Host OS: Windows Server 2016 Datacenter Evaluation.
HyperVisor: Hyper-V 10.0.14393.0.
VM OS Affected: CentOS Linux release 7.3.1611 (Core), x64 Server Edition.
Filebeat: Version 5.1.1
Filebeat Log: Default logging verbosity.

File before failure:
- input_type: log
    - /var/log/audit/audit.log

This is the only file that was in use. /var/log/filebeat/* yields just the standard EOF error, no details around time stamp or what action the filebeat agent was taking at the time of power failure. Is there an alternate location that might tell me this information?

2017-02-05T19:21:25+13:00 INFO Home path: [/usr/share/filebeat] Config path: [/etc/filebeat] Data path: [/var/lib/filebeat] Logs path: [/var/log/filebeat]
2017-02-05T19:21:25+13:00 INFO Setup Beat: filebeat; Version: 5.1.1
2017-02-05T19:21:25+13:00 INFO Max Retries set to: 3
2017-02-05T19:21:25+13:00 INFO Activated logstash as output plugin.
2017-02-05T19:21:25+13:00 INFO Publisher name: mon.testdomain.com
2017-02-05T19:21:25+13:00 INFO Flush Interval set to: 1s
2017-02-05T19:21:25+13:00 INFO Max Bulk Size set to: 2048
2017-02-05T19:21:25+13:00 INFO filebeat start running.
2017-02-05T19:21:25+13:00 INFO Registry file set to: /var/lib/filebeat/registry
2017-02-05T19:21:25+13:00 INFO Loading registrar data from /var/lib/filebeat/registry
2017-02-05T19:21:25+13:00 INFO Total non-zero values:
2017-02-05T19:21:25+13:00 INFO Uptime: 9.9703ms
2017-02-05T19:21:25+13:00 INFO filebeat stopped.
2017-02-05T19:21:25+13:00 CRIT Exiting: Could not start registrar: Error loading state: Error decoding states: EOF

*** This issue has been moved to Github from https://discuss.elastic.co/t/filebeat-fails-after-power-failure/73967 ***

[...]

PR #3668, opened 2017-02-24
Fix empty registry file on machine crash

ruflin:
Closes https://github.com/elastic/beats/issues/3537

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -44,6 +44,7 @@ https://github.com/elastic/beats/compare/v5.1.1...master[Check the HEAD diff]
 - Always use absolute path for event and registry. {pull}3328[3328]
 - Raise an exception in case there is a syntax error in one of the configuration files available under
   filebeat.config_dir. {pull}3573[3573]
+- Fix empty registry file on machine crash. {issue}3537[3537]
 
 *Heartbeat*
 
--- a/filebeat/registrar/registrar.go
+++ b/filebeat/registrar/registrar.go
@@ -298,7 +298,7 @@ func (r *Registrar) writeRegistry() error {
 	logp.Debug("registrar", "Write registry file: %s", r.registryFile)
 
 	tempfile := r.registryFile + ".new"
-	f, err := os.OpenFile(tempfile, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0600)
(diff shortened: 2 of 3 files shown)


winlogbeat/checkpoint/file_windows.go
```diff
@@ -29,8 +30,8 @@ func createWriteThroughFile(path string) (*os.File, error) {
 		syscall.GENERIC_READ|syscall.GENERIC_WRITE,               // Access Mode
 		uint32(syscall.FILE_SHARE_READ|syscall.FILE_SHARE_WRITE), // Share Mode
 		nil, // Security Attributes
-		syscall.CREATE_ALWAYS,                                          // Create Mode
-		uint32(syscall.FILE_ATTRIBUTE_NORMAL|_FILE_FLAG_WRITE_THROUGH), // Flags and Attributes
+		syscall.CREATE_ALWAYS, // Create Mode
+		uint32(syscall.FILE_ATTRIBUTE_NORMAL|_FILE_FLAG_WRITE_THROUGH|_FILE_FLAG_NO_BUFFERING), // Flags and Attributes
```
ruflin: @andrewkroh I tried to find O_SYNC for windows and it seems to be the NO_BUFFERING. Looking into filebeat, we do not treat creating the registry for different OS differently. Should we?

In any case, we should still unify registrar handling in libbeat.

andrewkroh: We could move this file creation code up to libbeat so that we could share it. That would be simpler than trying to unify the registrar (which is still a good idea, but more time consuming).

andrewkroh: From what I understand on Windows you need to use `FILE_FLAG_WRITE_THROUGH` to get writes to bypass the cache. FILE_FLAG_NO_BUFFERING implies FILE_FLAG_WRITE_THROUGH too, but it affects reads and adds the requirement that your reads must be aligned to disk sector boundaries. So I don't think we need FILE_FLAG_NO_BUFFERING here.

andrewkroh: This is what originally pointed me to FILE_FLAG_WRITE_THROUGH. https://stackoverflow.com/questions/6525943/cause-of-corrupted-file-contents/24302389#24302389

ruflin:
I removed `FILE_FLAG_NO_BUFFERING` again. As I plan to backport this PR to older branches, I keep the changes to a minimum and will to the moving to libbeat in a later PR.

PR #3671, opened 2017-02-24
Fix elasticsearch url parsing

urso:
use regex to check if host given contains the protocol scheme. If scheme is
missing, prepend before attempting to parse the complete URL


ruflin:
I would not backport this as this fixes an issue we have in Golang 1.8. As I don't see a need to get the 5.x releases on 1.8, there is no need to backport it.

PR #3672, opened 2017-02-24
Add module info to getting started

dedemorton:
I've added some "quick start" content about modules to the getting started guide because I want to highlight modules as the _easiest_ way to get started with common log formats. 

This is  meant to complement, not replace the info in the tutorial about modules. The tutorial explores lots of ways to do things, which I think is great, but doesn't showcase how simple it is to get started if you just want to get up and running to see how things work.

More changes coming to the actual modules doc in another PR....

### DIFF
--- a/filebeat/docs/getting-started.asciidoc
+++ b/filebeat/docs/getting-started.asciidoc
@@ -33,8 +33,11 @@ formats, such as Nginx, Apache2, and MySQL, and can be run by issuing a simple
 command.  
 
 This topic shows you how to run the basic modules out of the box without extra
-configuration. For detailed documentation and the full list of supported
+configuration. For detailed documentation and the full list of available
 modules, see <<filebeat-modules>>.
+
+Skip this topic and go to <<filebeat-installation>> if you are using a log file
+type that isn't supported by one of the available Filebeat modules.
 
 ==== Prerequisites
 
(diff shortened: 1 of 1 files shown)


dedemorton:
Wanted to note that another reason for adding module info to the GS is that our translation plan will likely involve translating the Getting Started guides before the other docs.

filebeat/docs/getting-started.asciidoc
```diff
@@ -20,10 +21,120 @@ After installing the Elastic Stack, read the following topics to learn how to in
...
+[[filebeat-modules-quickstart]]
+=== Quick Start for Common Log Formats
+
+beta[]
+
+Filebeat provides a set of pre-built modules that you can use to rapidly
+implement and deploy a log monitoring solution, complete with sample dashboards
+and data visualizations, in about 5 minutes. These modules support common log
+formats, such as Nginx, Apache2, and MySQL, and can be run by issuing a simple
+command.
```
tsg: It might make sense to add a paragraph somewhere here to explain that people should follow this Quick Start if they want to monitor log files from one of the supported modules. Otherwise, they can skip to "Installing Filebeat".

dedemorton:
@tsg I liked your suggestion, so I added a short sentence for users who can't use modules.

PR #3674, opened 2017-02-26
Report cwd

mcenirm:
(A restart of #3639)

I need metricbeat's system process metricset to report current working directory (cwd).

It was straightforward to use gosigar's ProcExe, but Linux seems to be the only platform for which gosigar can provide cwd in any feasible manner. On Darwin, gosigar uses libproc.h which is not a public interface. On FreeBSD, gosigar relies on procfs and Linux-compatible linprocfs which do not provide cwd in an obvious way. On Windows, gosigar's `(*ProcExe) Get(int)` returns ErrNotImplemented.

### DIFF
--- a/metricbeat/docs/fields.asciidoc
+++ b/metricbeat/docs/fields.asciidoc
@@ -6585,7 +6585,7 @@ The username of the user that created the process. If the username cannot be det
 
 type: keyword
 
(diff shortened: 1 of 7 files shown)


ruflin:
jenkins, test it

PR #3676, opened 2017-02-27
Improve error when ES Ingest node plugins are not loaded

tsg:
We're parsing the Elasticsearch JSON error and try to produce an
error message that is as helpful as possible. The following cases
are detected:

* A plugin providing a processor is missing. In case the plugin is one of
  `ingest-geoip` or `ingest-user-agent`, we can also suggest the command that
  installs them.
* Elasticsearch < 5.0. We now detect this and tell the user that ES 5.0 is
  required by FBM.

A drawback of this approach is that if both the GeoIP and User-Agent plugins
are missing, only one will be reported. This might get solved by including the
user-agent one in ES, or by improving the error we get from ES, or by us querying
the node stats API

Note: this contains a change in the ES client, which makes it return the body
in case of errors. I think we need that part anyway, otherwise we often show
errors like `400 Bad request` without any other details. I tried to do a minimal
change there, I hope I didn't introduce any changes in behaviour.

### DIFF
--- a/filebeat/fileset/modules.go
+++ b/filebeat/fileset/modules.go
@@ -302,13 +302,13 @@ func interpretError(initialErr error, body []byte) error {
 		var response1x struct {
 			Error string `json:"error"`
 		}
-		json.Unmarshal(body, &response1x)
-		if response1x.Error != "" {
+		err1x := json.Unmarshal(body, &response1x)
+		if err1x == nil && response1x.Error != "" {
 			return fmt.Errorf("The Filebeat modules require Elasticsearch >= 5.0. "+
 				"This is the response I got from Elasticsearch: %s", body)
 		}
 
-		return fmt.Errorf("couldn't load pipeline: %v. Additionally, error decoding response body: %s.",
+		return fmt.Errorf("couldn't load pipeline: %v. Additionally, error decoding response body: %s",
 			initialErr, body)
 	}
 
--- a/filebeat/fileset/modules.go
+++ b/filebeat/fileset/modules.go
@@ -247,7 +247,7 @@ func (reg *ModuleRegistry) GetProspectorConfigs() ([]*common.Config, error) {
 // PipelineLoader is a subset of the Elasticsearch client API capable of loading
 // the pipelines.
 type PipelineLoader interface {
-	LoadJSON(path string, json map[string]interface{}) (error, []byte)
+	LoadJSON(path string, json map[string]interface{}) ([]byte, error)
 	Request(method, path string, pipeline string, params map[string]string, body interface{}) (int, []byte, error)
 }
 
@@ -275,7 +275,7 @@ func loadPipeline(esClient PipelineLoader, pipelineID string, content map[string
 		logp.Debug("modules", "Pipeline %s already loaded", pipelineID)
 		return nil
 	}
-	err, body := esClient.LoadJSON(path, content)
+	body, err := esClient.LoadJSON(path, content)
 	if err != nil {
 		return interpretError(err, body)
 	}
--- a/filebeat/fileset/modules.go
+++ b/filebeat/fileset/modules.go
@@ -1,10 +1,12 @@
 package fileset
 
 import (
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/logp"
@@ -245,7 +247,7 @@ func (reg *ModuleRegistry) GetProspectorConfigs() ([]*common.Config, error) {
 // PipelineLoader is a subset of the Elasticsearch client API capable of loading
 // the pipelines.
 type PipelineLoader interface {
-	LoadJSON(path string, json map[string]interface{}) error
+	LoadJSON(path string, json map[string]interface{}) (error, []byte)
 	Request(method, path string, pipeline string, params map[string]string, body interface{}) (int, []byte, error)
 }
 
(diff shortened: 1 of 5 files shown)


filebeat/fileset/modules.go
```diff
@@ -273,14 +275,78 @@ func loadPipeline(esClient PipelineLoader, pipelineID string, content map[string
...
+				Reason string `json:"reason"`
+				Header struct {
+					ProcessorType string `json:"processor_type"`
+				} `json:"header"`
+				Index string `json:"index"`
+			} `json:"root_cause"`
+		} `json:"error"`
+	}
+	err := json.Unmarshal(body, &response)
+	if err != nil {
```
ruflin: `client.Connection.version` could be used to check the ES version. Unfortunately the variable is currently not public.

filebeat/fileset/modules.go
```diff
@@ -273,14 +275,78 @@ func loadPipeline(esClient PipelineLoader, pipelineID string, content map[string
...
+				"This is the response I got from Elasticsearch: %s", body)
+		}
+
+		return fmt.Errorf("couldn't load pipeline: %v. Additionally, error decoding response body: %s.",
+			initialErr, body)
+	}
+
+	// missing plugins?
+	if len(response.Error.RootCause) > 0 &&
+		response.Error.RootCause[0].Type == "parse_exception" &&
```
ruflin: ES should (in the future) expose here a special root cause type so no check of the text is needed.

filebeat/fileset/modules.go
tsg: Hmm, yeah, that would be an option. But we also don't have the client here and adding it would complicate unit testing, and we still need to do the error checking anyway, so I think it doesn't win us that much to add version checks.

filebeat/fileset/modules.go
tsg: Yeah, I think mvg is working on improving that.

filebeat/fileset/modules.go
```diff
@@ -245,7 +247,7 @@ func (reg *ModuleRegistry) GetProspectorConfigs() ([]*common.Config, error) {
 // PipelineLoader is a subset of the Elasticsearch client API capable of loading
 // the pipelines.
 type PipelineLoader interface {
-	LoadJSON(path string, json map[string]interface{}) error
```
tsg: I updated this to have the error as the last returned arg.

libbeat/outputs/elasticsearch/client.go
```diff
@@ -718,15 +718,16 @@ func (conn *Connection) execHTTPRequest(req *http.Request) (int, []byte, error)
...
 	status := resp.StatusCode
+	var retErr error
 	if status >= 300 {
-		return status, nil, fmt.Errorf("%v", resp.Status)
+		retErr = fmt.Errorf("%v", resp.Status)
 	}
 
 	obj, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
 		return status, nil, err
```
urso: we want to return retErr or err here?

filebeat/fileset/modules.go
```diff
@@ -273,14 +275,78 @@ func loadPipeline(esClient PipelineLoader, pipelineID string, content map[string
...
+			} `json:"root_cause"`
+		} `json:"error"`
+	}
+	err := json.Unmarshal(body, &response)
+	if err != nil {
+		// this might be ES < 2.0. Do a best effort to check for ES 1.x
+		var response1x struct {
+			Error string `json:"error"`
+		}
+		json.Unmarshal(body, &response1x)
```
urso: json.Unmarshal might also fails, cause body is not json at all. the client was changed to always return the raw body. e.g. if error is in nginx proxying to ES, content might be a plain message.

filebeat/fileset/modules.go
```diff
@@ -273,14 +275,78 @@ func loadPipeline(esClient PipelineLoader, pipelineID string, content map[string
...
+	// missing plugins?
+	if len(response.Error.RootCause) > 0 &&
+		response.Error.RootCause[0].Type == "parse_exception" &&
+		strings.HasPrefix(response.Error.RootCause[0].Reason, "No processor type exists with name") &&
+		response.Error.RootCause[0].Header.ProcessorType != "" {
+
+		plugins := map[string]string{
+			"geoip":      "ingest-geoip",
+			"user_agent": "ingest-user-agent",
+		}
```
urso: generalize map with:

```
func ingestName(name string) name {
  return fmt.Sprintf("ingest-%v", strings.Replace(name, "_", "-"))
}
```

filebeat/fileset/modules.go
```diff
@@ -273,14 +275,78 @@ func loadPipeline(esClient PipelineLoader, pipelineID string, content map[string
...
+
+		return fmt.Errorf("couldn't load pipeline: %v. Additionally, error decoding response body: %s.",
+			initialErr, body)
+	}
+
+	// missing plugins?
+	if len(response.Error.RootCause) > 0 &&
+		response.Error.RootCause[0].Type == "parse_exception" &&
+		strings.HasPrefix(response.Error.RootCause[0].Reason, "No processor type exists with name") &&
```
urso: instead of checking error message prefix, is there an error type report we can check?

filebeat/fileset/modules.go
```diff
@@ -273,14 +275,78 @@ func loadPipeline(esClient PipelineLoader, pipelineID string, content map[string
...
+
+	// older ES version?
+	if len(response.Error.RootCause) > 0 &&
+		response.Error.RootCause[0].Type == "invalid_index_name_exception" &&
+		response.Error.RootCause[0].Index == "_ingest" {
+
+		return fmt.Errorf("The Ingest Node functionality seems to be missing from Elasticsearch. "+
+			"The Filebeat modules require Elasticsearch >= 5.0. "+
+			"This is the response I got from Elasticsearch: %s", body)
+	}
```
urso: which error message will be returned in ingest node is disabled?

filebeat/fileset/modules.go
tsg: Hmm, i don't know if that future proof. There's no guarantees that other plugins will follow this pattern.

filebeat/fileset/modules.go
tsg: Just tried it out. There's no real way of disabling the ingest functionality, but what one can do is set `node.ingest: false` on all the ES nodes. In that case, the pipeline loading works as possible but the `_bulk` insert fails with:

```
path: /_bulk, params: {}
java.lang.IllegalStateException: There are no ingest nodes in this cluster, unable to forward request to an ingest node.
```

I'd say improving the error handling in that case is beyond the scope of this PR.

filebeat/fileset/modules.go
tsg: The error type is `parse_exception`, so that's to generic.

@martijnvg are you adding a new error type as part of your PRs?

libbeat/outputs/elasticsearch/client.go
tsg: Hmm, debatable I guess. I'm changing it to `retErr` because I think that's closer to the previous behavior. Thanks.

filebeat/fileset/modules.go
tsg: I know, but in that case we fall back to the "Could not load pipeline. Additionally, error decoding body:" message, which I think is what we want. Looking into making the code clearer..

filebeat/fileset/modules.go
martijnvg: The PR doesn't change that. It failed during parsing hence the error type is `parse_exception`.

I'm open in changing this, but not sure what other existing error type to use. ES is very defensive in introducing new error types. The only existing general error type that comes up to me is `resource_not_found`. It is generic too, but maybe in this context better? (indicating that a processor type doesn't exist).

tsg: I think I'd prefer not changing it in that case, because we'd still have to leave this branch in the code for ES < 5.4, so it's probably not worth it.

filebeat/fileset/modules.go
urso: ouch. Maybe we can probe ingest node availability via simulate API?

tsg: Yeah, maybe, some other time :)

Issue #3654, opened 2017-02-23
Dynamically generate and load index templates


ruflin:
Currently index templates are generated with python and then packaged with each beat. This has the disadvantage that with more ES versions more files have to be generated and packaged. In addition it does not allow to only provide a template for one metricbeat module for example. Generating the templates as part of the beat will allow to generate the needed templates on request and does not required anymore prepackage templates.

Implementation steps:

* [x] Implement generation in Golang https://github.com/elastic/beats/pull/3603
  * [x] Remove defaults from fields.common.yml for more flexibility https://github.com/elastic/beats/pull/3602
  * [x] Unify naming of fields.yml entries https://github.com/elastic/beats/pull/3655
* [x] Unify fields generated file across all beats #3677 https://github.com/elastic/beats/pull/3683
* [x] Set template order for seamless migrations: https://github.com/elastic/beats/issues/3657
* [x] Implement dynamic generation of template when starting the beat https://github.com/elastic/beats/pull/3681
* [x] Remove generated templates
* [ ] Reorganise template config options https://github.com/elastic/beats/issues/3921
* [ ] Add config option to generate json on disk #4323
* [x] Add support to check elasticsearch version and generate templates accordingly
* [x] ~~Allow template generation per Metricbeat module~~ (multiple template not supported by ES 6)
* [ ] Move manual template generation to `beats-tools` or something similar
* [x] Identify how variables like `number_of_shards` could be easily set https://github.com/elastic/beats/pull/4284
* [ ] Review the guides in the docs around template and dashboard loading.

ruflin:
Elasticsearch will potentially remove the support for template inheritance in 6.0 and will always pick the one if the highest order. This will impact the last point "template generation per module".

tsg:
@dedemorton Pinging you here for awareness, for 6.0 we'll be significantly changing the way templates and dashboards are loaded. This will likely require some updates in the existing guides, but lets wait for things to stabilize a bit before updating them.

tsg:
Closing, looks mostly done.

PR #3681, opened 2017-02-27
Dynamically generate template on startup

ruflin:
So far the template was loaded from the `beatname.template.json` files. This now uses the `fields.yml` file to dynamically generate the template based on the ES version connected to. The name of the template is set dynamically based on `template.name`. By default it has the value {beatname}.

* [x] Remove generated templates
* [x] Package fields.yml
* [x] Clean up config options as 2.x option is not needed anymore
* [x] Update docs
* [x] Update and test packaging

Part of #3654


tsg:
My 2c on this.

> Config option is called template.path which points to fields.yml. Is the name still ok?

I'd say it's better to use a different name, since this is quite different conceptually. Also, I'd suggest it is a good idea to let the user provide a fully custom template file (which they can generate/edit whatever name they like).

I wouldn't worry too much about backwards compatibility in this case, since I'd guess (and hope) most people didn't use a custom template path.

> Should template config be moved to the top level (like dashboard)

Yeah, configuration is tricky here, perhaps we have a global `fields` section where the `fields.yml` is configured, and the rest of the things pull from that?

> What is our recommendation on loading the template manually? Generate first with index_template binary and then load manually?

I agree there should be a way to write the template to a file. But I guess the primary way of adjusting things will be by editing the `fields.yml` file, which will be considered configuration?

> Do we ship in index_template binary similar to import_dashboards? Can we integrate it into beats directly? -> setup

A binary has the disadvantage of adding to the package size. We could consider creating a `beats-tools` package that contains only these tools. This would be for things that you don't need for getting started, but are helpful in some situations.

tsg:
Another, kind of random, thought: I'd like to make it fairly easy to change some index settings (that typically go in the template), like `_source: on/off`, `number_of_shards`, `number_of_replicas`, etc. We could put these in `fields.yml` or in the main Beat configuration file, but in both cases we should be careful about how much we want to duplicate the ES settings in the Beats config.

heartbeat/heartbeat.full.yml
```diff
@@ -346,20 +346,12 @@ output.elasticsearch:
   # so the final name is heartbeat-%{[beat.version]}.
   #template.name: "heartbeat"
 
-  # Path to template file
-  #template.path: "${path.config}/heartbeat.template.json"
+  # Path to fields.yml file to generate the template
+  #template.path: "${path.config}/fields.yml"
```
ruflin: Agree on rethinking the config options. I added it to https://github.com/elastic/beats/issues/3654

ruflin:
@tsg Agreement on the points above. I kind of start to like the idea of a beats-tools package which is mainly for sysadmins etc for the setup phase but could also be used in the future to do some "checks".

I added all the points above to the meta issue to be solved https://github.com/elastic/beats/issues/3654 I suggest moving forward with this PR as is and to further changes in follow up PR's.

ruflin:
jenkins, retest it

PR #3682, opened 2017-02-27
Bug: Fix massive values for cpu metricset for docker module

douaejeouit:
Due to the use of unsigned integers, negative values were transformed into huge positive numbers.


ruflin:
@douaejeouit How can it happen in the first place that these values become negative?

douaejeouit:
@ruflin,  Actually the changes made  prevent the module from any potential bug that may be caused by negative values.As you know, the CPU's usage values are accumulated by the process of a container. So in order to get the current cpu usage value,  we need to calculate it by using the Pre-cpu  (current = current - old). The negative value may occurs if the current usage value is larger than the older value. 
My suggestion to this  is to multiply  the value by a '-1' , if it's negative, instead of setting it to 0. Thus, keeping the resulting value of the current cpu usage. 
I hope I was clear in my ideas

douaejeouit:
@ruflin,  Actually the changes made are only to  prevent the module from any potential bug that may be caused by negative values. As you know, the CPU's usage values are accumulated by the processes of the container. So in order to get the current cpu usage values,  we need to calculate them by using the Pre-cpu  (current = current - old). The negative value may occurs if the current usage value is larger than the older value. 
My suggestion to this is to multiply  the value by a '-1' , if it's negative, instead of setting it to 0. we can keep the resulting value of the current cpu usage. 
I hope I was clear in my ideas

douaejeouit:
@ruflin, Actually the changes made are only to prevent the module from any potential bug that may be caused by negative values. As you know, the CPU's usage values are accumulated by the processes of the container. So in order to get the current cpu usage value, we need to calculate it by using both cpu and  Pre-cpu usage (current = new- old). The negative value might be occurring due to the fact that the new usage value is larger than the older value.
My suggestion to this is to multiply the value by a '-1' , if it's negative, instead of setting it to 0. we can keep the resulting value of the current cpu usage. what is you opinion about it ?

ruflin:
What if we just change it from uint to int and report the actual negative values that were reported?

douaejeouit:
@ruflin,  sorry for the delay of replying. Yes, that sounds good to me! But (for the calcularion) I think it would be better to cast it to float rather than int, don't you think so?

douaejeouit:
I 've just  realised that I misspoke in my previous comment ><" ! The idea was to keep the resulting value and  not the 'resulting value of the current cup usage..  Sorry for that !
 I think it's judicious to report the value as it was reported  (negative )

ruflin:
My thought process here: 

* We need to fix the massive values
* Either docker values reported are wrong or our calculation is wrong
* Do negative values make any sense? Probably not.
* What is the reason we do any calculations on the usage? Could we just report the total values and let ES do the calculations on query time?

Do not want to hold back this PR, just want to make sure I fully understand what it does and if we are not trying to fix something, that potentially should not exist :-)

ruflin:
@douaejeouit Any thoughts on the above?

douaejeouit:
@ruflin sorry for the delay. 
Yes, we need to fix it. But we also need to understand the reason of the occurrence of the massive values.

- Why do we have negative values? 
Having a negative value doesn't make any sense. I'm not sure if it's due to the calculation or to the returned values from docker API itself.

- Why do we have massive values? 
The calculation returns a delta of old CPU values and new ones. Those values are stored in a uint64 variable type ( because we're not expecting negative values). So using this type transform negative values into huge values instead of the real calculated value.  That's why we get massive values

- What should we do? 
I see two options, we can either change the variable type used to store the delta to keep and report the actual negative value. It's true that it represents the calculation result,  but I'm afraid it doesn't make sense?  Or we can set this last into 0 et report 0 instead of the negative value! 
I did open this PR to prevent the module to report huge numbers ( it's a rare occurrence) which don't make sense.

metricbeat/module/docker/cpu/helper.go
```diff
@@ -78,21 +78,25 @@ func perCpuUsage(stats *dc.Stats) common.MapStr {
...
 
 func systemUsage(stats *dc.Stats) float64 {
-	return calculateLoad(stats.CPUStats.SystemCPUUsage - stats.PreCPUStats.SystemCPUUsage)
+	return calculateLoad(stats.CPUStats.SystemCPUUsage, stats.PreCPUStats.SystemCPUUsage)
 }
 
-func calculateLoad(value uint64) float64 {
-	return float64(value) / float64(1000000000)
+func calculateLoad(newValue uint64, oldValue uint64) float64 {
+	value := float64(newValue) - float64(oldValue)
```
ruflin: Lets add a comment here on why we do this.

metricbeat/module/docker/cpu/helper.go
```diff
@@ -78,21 +78,25 @@ func perCpuUsage(stats *dc.Stats) common.MapStr {
...
+	if value < 0 {
+		value = 0
```
ruflin: Should we log an error here? Perhaps we should even set it to -1 to indicate something is up?

Also we can directly return here the value as 0 / x is 0 again.

douaejeouit:
Ok. I fully agree with you! We need to figure out what's happening. I'll be on it this weekend. I'll ping you ASAP.

metricbeat/module/docker/cpu/helper.go
douaejeouit: This function is meant to calculate the % CPU time change between two successive readings. The "oldVlue" represents the preCpu which refers to the  CPU statistics of the last read. 
Time here is expressed by second and not by milliseconds. ( The main goal is to expose the %, in the same way, it's displayed by docker Client)

douaejeouit:
Here are my thoughts: 
- The calculation isn't wrong because it concerns the time the CPU has been in use since **boot** ( user, system or total use). Therefore, those values can't decrease ( newValue can't be < oldValue). Thus, if we get the expected values from the API, everything should be ok! 
 _I couldn't reproduce this [scenario](https://github.com/Ingensi/dockbeat/issues/112)._

- I think it's better to report the calculated percentages rather than letting ES do the calculations in query time. The reason is simple:  take advantage of the preCPU field provided by the API, ( precpu field is not the exact copy of cpu_stats according to the [documentation ](https://docs.docker.com/engine/api/v1.21/#containers) ) & Since we don't expose any PreCPU field, the user can't have the right values.

- Finally, I think that we still have to handle this just in case of! Setting the value to -1 LGTM.
What do you think about it?

metricbeat/module/docker/cpu/helper.go
ruflin: Thanks. Can you add it as a comment to the source code?

metricbeat/module/docker/cpu/helper.go
```diff
@@ -78,21 +79,30 @@ func perCpuUsage(stats *dc.Stats) common.MapStr {
...
-func calculateLoad(value uint64) float64 {
-	return float64(value) / float64(1000000000)
+// This function is meant to calculate the % CPU time change between two successive readings.
+// The "oldValue" refers to the CPU statistics of the last read.
+// Time here is expressed by second and not by nanoseconde.
+// The main goal is to expose the %, in the same way, it's displayed by docker Client.
+func calculateLoad(newValue uint64, oldValue uint64) float64 {
+	value := float64(newValue) - float64(oldValue)
+	if value < 0 {
+		logp.Err("time change calculation failed")
```
ruflin: as these errors appear on a global level, some more details would be good like that it is a the docker module and perhaps the old and new value or something that is helpful for debugging.

douaejeouit: Ok. what about:  "error calculating CPU time change for docker module: new stats value is lower than the old one". Is it fine for you?

ruflin: SGTM: I would add the two values at the end or as part of the log message something like `new stats value (%v)  is lower then the old one (%v)`

ruflin:
@douaejeouit Thanks for the changes. Can you rebase on master. There seem to be some conflicts :-(

ruflin:
@douaejeouit Merged. Thanks for going through all the back and forth with me :-)

douaejeouit:
Thank you, with pleasure!

PR #3683, opened 2017-02-27
Simplify template generation and handling

ruflin:
* Only one file is passed instead of an array as now only file is needed. Introduce `-file` param.
* Local loading of beats version was removed as it can lean to cyclic imports
* Move `_meta/fields.full.generated.yml` to `fields.yml` in preparation for packaging `fields.yml` and ignore it in git
* Cleanup Template generation package
* Rename -beat.name param to -index as this describes it better

Part of https://github.com/elastic/beats/issues/3654

### DIFF
--- a/dev-tools/cmd/index_template/index_template.go
+++ b/dev-tools/cmd/index_template/index_template.go
@@ -13,32 +13,27 @@ import (
 // main generates index templates for the beats
 func main() {
 
-	beatName := flag.String("beat.name", "", ": Base index name. Normally {beat_name} (required)")
+	beatVersion := beat.GetDefaultVersion()
+	index := flag.String("index", "", "Base index name. Normally {beat_name} (required)")
 	output := flag.String("output", "", "Required: Full path to the output file (required)")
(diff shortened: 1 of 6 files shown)


dev-tools/cmd/index_template/index_template.go
```diff
@@ -13,40 +13,41 @@ import (
...
 
-	// Skip some of the passed files, as not all beats have the same files
-	for _, f := range files {
-		if _, err := os.Stat(f); err != nil {
-			fmt.Printf("Skipping file because it does not exist: %s", f)
-			continue
-		}
-		existingFiles = append(existingFiles, f)
+	if _, err := os.Stat(*file); err != nil {
+		fmt.Fprintf(os.Stderr, "Input -file does not exist: %s", *file)
```
andrewkroh: `os.Stat` could fail for reasons other than the file does not exists. You can test for that specific case with [os.IsNotExist](https://golang.org/pkg/os/#IsNotExist) or maybe you could just include the error message from stat in the output.

ruflin: I will just include the error, because i think all errors should be covered.

PR #3689, opened 2017-02-28
Exempt root owned config files from ownership checker

andrewkroh:
If a config file is owned by root, but the process is running as a non-privileged user the Beat should run. This change exempts root from the ownership test. This makes it possible to drop privileges before executing the beat.

Needs back port to 5.3.


andrewkroh:
I updated the existing changelog entry associated with this unreleased feature.

PR #3694, opened 2017-02-28
Document process.env.whitelist config option

andrewkroh:
Document `process.env.whitelist` which is used by the Metricbeat system process metricset to specify what environment variables should be captured.

Adds documentation for #3337.

Needs back port to 5.3.

### DIFF
--- a/metricbeat/module/system/process/_meta/docs.asciidoc
+++ b/metricbeat/module/system/process/_meta/docs.asciidoc
@@ -17,3 +17,21 @@ On Linux this metricset will collect metrics from any cgroups that the process
 is a member of. This feature is enabled by default and can be disabled by adding
 `process.cgroup.enabled: false` to the system module configuration.
 
+[float]
+=== Process Environment Variables
+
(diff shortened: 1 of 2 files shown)


metricbeat/module/system/process/_meta/docs.asciidoc
```diff
@@ -17,3 +17,21 @@ On Linux this metricset will collect metrics from any cgroups that the process
 is a member of. This feature is enabled by default and can be disabled by adding
 `process.cgroup.enabled: false` to the system module configuration.
 
+[float]
+=== Process Environment Variables
+
+This metricset can collect the environment variables that were used to start the
+process. This feature is available on Linux, Darwin, and FreeBSD. By default
+no environment variables are collected. You must configure the environment
```
ruflin: We could mention here, that for security reasons no env variables are fetched by default.

andrewkroh: I added the reasoning behind the behavior.

PR #3710, opened 2017-03-02
Removed the no_plugins option from nginx and apache2

tsg:
We've decided not to support the degraded "no_plugins" mode, in case
the Ingest Node plugins are missing. So this is removing the `pipeline`
option variable from the Nginx and Apache2 module, along with the
associated tests.

Opened against 5.3 again, needs cherry pick to master.


tsg:
I think Jenkins was green, just the status is not updated: http://build-eu-00.elastic.co/job/beats-PR/5310/

hauthorn:
I'm sorry to disturb here. Does this mean that I can't use the modules if I haven't installed the required plugins?
In that case, the tutorial in the documentation is out of date: https://www.elastic.co/guide/en/beats/filebeat/current/_tutorial.html#module-varialbe-overrides

PR #3715, opened 2017-03-03
Fix close_timeout for new started harvesters

ruflin:
If close_timeout is set, a file handler inside the harvester is closed after `close_timeout` is reached even if the output is blocked. For new which were found for harvesting during the output was blocked, it was possible that the Setup of the file happened but then the initial state could not be sent. So the harvester was never started, file handler was open but `close_timeout` this not apply. Setup and state update are now turned around. This has the affect, that in case of an error during the setup phase, the state must be revert again.

See https://github.com/elastic/beats/pull/3091#issuecomment-283519009

### DIFF
--- a/filebeat/prospector/prospector.go
+++ b/filebeat/prospector/prospector.go
@@ -308,16 +308,24 @@ func (p *Prospector) startHarvester(state file.State, offset int64) error {
 		return err
 	}
 
-	reader, err := h.Setup()
-	if err != nil {
-		return fmt.Errorf("Error setting up harvester: %s", err)
-	}
-
-	// State is directly updated and not through channel to make state update immediate
-	// State is only updated after setup is completed successfully
+	// State is directly updated and not through channel to make state update synchronous
 	err = p.updateState(input.NewEvent(state))
 	if err != nil {
 		return err
+	}
+
+	reader, err := h.Setup()
+	if err != nil {
+		// Set state to finished True again in case of setup failure to make sure
+		// file can be picked up again by a future harvester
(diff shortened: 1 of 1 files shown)


urso:
+1 for change, as state represents if an harvester is `active or already scheduled` for a file to be read. That is, before calling harvester.setup it MUST be clear a harvester is scheduled. One could also protect harvester setup/start via `recover` and mark state on panic.

From code-comments (at least those I've seen in PR), the importance/role of the state is not fully clear.

currently harvester sends final state update when it goes down. Makes me wonder if we want to change the harvester/prospector interaction a little to:
- prospector.startHarvester (update state to indicate harvester is 'active/scheduled')
- harvester.Harvest returns final/last state on shutdown
- prospector.runHarvester updates the state + captures potential panic (e.g. custom input codec or some other code panicing):

```
func (p *prospector) runHarvester(h *Harvester) {
  go func() {
    defer func() {
      if fail := recover(); ... {
        ...
      }
    }

    finalState := harvester.Harvest(...)
    // TODO: update final state
  }
}
```

That is, all required calls to update the prospector state will be with the prospector package.

ruflin:
I was actually thinking going the exact opposite in the long run (not as part of this PR) and to do everything in the harvester. Assuming for the unique identifier of a harvester is not a uuid used but the combination of inode /device the registry could be instrumented to guarantee the same harvester is started twice. If we have this guarantee, the harvester can again send the state update again in its own go routine, as the registry will prevent any other harvester from being started.

PR #3717, opened 2017-03-03
Expose metrics through http endpoint

ruflin:
The following PR exposes beats metrics through a configurable http endpoint. This allows when enabled to get an insight into a running beat. For security reasons the endpoint is off by default.

**Configuration**

The configuration options are in the `http` namespace. This config naming is borrowed from Logstash. By default the http endpoint is disabled. If enabled the metrics are only exposed on localhost on port 5066.

```
http.enabled: false
http.host: localhost
http.port: 5066
```

**-httpprof ?**

The http endpoint can be enabled also in production if needed. The additional endpoint httpprof endpoint which can be enabled through `-httpprof` still exists but is only recommended for debugging purpose. The httpprof endpoint exposes many more metrics and runtime data then the metrics endpoint.

**Endpoints**

The current implementation has two endpoints:

* `/`: The standard endpoint exposes info about the beat
* `/stats`: Stats exposes all metrics collected by monitoring

The output of the data is in json. The flag `?pretty` can be used to have formatted json as output to make it more human readable. Below is an example of each endpoint.

**/**

```
{
  "beat": "metricbeat",
  "hostname": "ruflin",
  "name": "ruflin",
  "uuid": "9d6e0c3a-1677-424c-aead-097e597e09f9",
  "version": "6.0.0-alpha1"
}
```

**/stat**

```
{
  "beat": {
    "memstats": {
      "gc_next": 6262080,
      "memory_alloc": 3879968,
      "memory_total": 479284520
    }
  },
  "libbeat": {
    "config": {
      "module": {
        "running": 0,
        "starts": 0,
        "stops": 0
      },
      "reloads": 0
    }
  },
  "metricbeat": {
    "system": {
      "cpu": {
        "events": 7,
        "failures": 0,
        "success": 7
      },
      "filesystem": {
        "events": 28,
        "failures": 0,
        "success": 7
      },
      "fsstat": {
        "events": 7,
        "failures": 0,
        "success": 7
      },
      "load": {
        "events": 7,
        "failures": 0,
        "success": 7
      },
      "memory": {
        "events": 7,
        "failures": 0,
        "success": 7
      },
      "network": {
        "events": 70,
        "failures": 0,
        "success": 7
      },
      "process": {
        "events": 1324,
        "failures": 0,
        "success": 7
      }
    }
  },
  "output": {
    "elasticsearch": {
      "events": {
        "acked": 1445,
        "not_acked": 0
      },
      "publishEvents": {
        "call": {
          "count": 34
        }
      },
      "read": {
        "bytes": 17213,
        "errors": 0
      },
      "write": {
        "bytes": 1185991,
        "errors": 0
      }
    },
    "events": {
      "acked": 1445
    },
    "kafka": {
      "events": {
        "acked": 0,
        "not_acked": 0
      },
      "publishEvents": {
        "call": {
          "count": 0
        }
      }
    },
    "logstash": {
      "events": {
        "acked": 0,
        "not_acked": 0
      },
      "publishEvents": {
        "call": {
          "count": 0
        }
      },
      "read": {
        "bytes": 0,
        "errors": 0
      },
      "write": {
        "bytes": 0,
        "errors": 0
      }
    },
    "messages": {
      "dropped": 0
    },
    "redis": {
      "events": {
        "acked": 0,
        "not_acked": 0
      },
      "read": {
        "bytes": 0,
        "errors": 0
      },
      "write": {
        "bytes": 0,
        "errors": 0
      }
    },
    "write": {
      "bytes": 1185991,
      "errors": 0
    }
  },
  "publisher": {
    "events": {
      "count": 1450
    },
    "queue": {
      "messages": {
        "count": 1450
      }
    }
  }
}
```

**Questions**

* Are these good endpoints paths?
* Which should be our default port?

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -6,7 +6,7 @@ coverage:
       default:
         # basic
         target: auto
-        threshold: null
+        threshold: 0.1
         base: auto
         # advanced
         branches: null
--- a/filebeat/filebeat.full.yml
+++ b/filebeat/filebeat.full.yml
@@ -938,6 +938,19 @@ output.elasticsearch:
 # dashboards and index pattern. Example: testbeat-*
 #dashboards.index:
 
+#================================ HTTP Endpoint ======================================
+# Each beat can expose internal data points through a http endpoint. For security
+# reason the endpoint is disabled by default. This feature is currently in beta.
+
+# Defines if http endpoint is enabled
+#http.enabled: false
+
+# Host to expose the http endpoint to. It is recommended to use only localhost.
+#http.host: localhost
+
+# Port on which the http endpoint is exposed. Default is 5066
+#http.port: 5066
+
 #================================ Logging ======================================
 # There are three options for the log output: syslog, file, stderr.
 # Under Windows systems, the log files are per default sent to the file output,
--- a/libbeat/api/server.go
+++ b/libbeat/api/server.go
@@ -1,0 +1,73 @@
+package api
+
+import (
+	"fmt"
+	"net/http"
+	"net/url"
+	"strconv"
+
+	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/logp"
+	"github.com/elastic/beats/libbeat/monitoring"
+)
+
+// Start starts the metrics api endpoint on the configured host and port
+func Start(cfg *common.Config, info common.BeatInfo) {
+	logp.Beta("Metrics endpoint is enabled.")
+	config := DefaultConfig
+	cfg.Unpack(&config)
+
+	logp.Info("Starting stats endpoint")
+	go func() {
+		mux := http.NewServeMux()
+
+		// register handlers
+		mux.HandleFunc("/", rootHandler(info))
+		mux.HandleFunc("/stats", statsHandler)
+
+		url := config.Host + ":" + strconv.Itoa(config.Port)
+		logp.Info("URL: %s", url)
+		endpoint := http.ListenAndServe(url, mux)
+		logp.Info("finished starting stats endpoint: %v", endpoint)
+	}()
+}
+
+func rootHandler(info common.BeatInfo) func(http.ResponseWriter, *http.Request) {
+	return func(w http.ResponseWriter, r *http.Request) {
+		// Return error page
+		if r.URL.Path != "/" {
+			http.NotFound(w, r)
+			return
+		}
+
+		w.Header().Set("Content-Type", "application/json; charset=utf-8")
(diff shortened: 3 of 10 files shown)


urso:
- we should not start a second http server but see how we can unify this with the server started via `-httpprof`
- don't use the `Do` interface, but checkout the `CollectX` functions in monitoring package. See #3739 for flat and nested snapshot support.

ruflin:
@urso For the second http server: That is only the case if `-httpprof` is enabled and the stats endpoint is enabled. In all other cases there is only 0 or 1 http endpoint I think. As the 2 server different purposes from my point of view, I'm ok with that. Enabling `-httprof` is only for debugging purpose.

codecov.yml
```diff
@@ -6,7 +6,7 @@ coverage:
       default:
         # basic
         target: auto
-        threshold: null
+        threshold: 0.1
```
urso: uhm... correct PR?

filebeat/filebeat.full.yml
```diff
@@ -938,6 +938,19 @@ output.elasticsearch:
 # dashboards and index pattern. Example: testbeat-*
 #dashboards.index:
 
+#================================ HTTP Endpoint ======================================
+# Each beat can expose internally collected metrics through a http endpoint. For security
```
urso: can we get rid of mentioning 'metrics' here?

libbeat/api/server.go
```diff
@@ -0,0 +1,68 @@
...
+		mux.HandleFunc("/stats", metricsHandler)
+
+		url := config.Host + ":" + strconv.Itoa(config.Port)
+		logp.Info("URL: %s", url)
+		endpoint := http.ListenAndServe(url, mux)
+		logp.Info("finished starting stats endpoint: %v", endpoint)
+	}()
+}
+
+func rootHandler(w http.ResponseWriter, r *http.Request, info common.BeatInfo) {
```
urso: alternatively turn `rootHandler` into a httpHandler function via:

```
func rootHandler(info common.BeatInfo) func(http.ResponseWriter, *http.Request) {
   ...
}
```

codecov.yml
ruflin: kind off, because otherwise the PR would fail :-)

libbeat/api/server.go
ruflin: done

filebeat/filebeat.full.yml
ruflin: done

PR #3718, opened 2017-03-03
Cherry-pick to 5.3: Fix close_timeout for new started harvesters

ruflin:
Cherry-pick of PR #3715 to 5.3 branch. Original message: 

If close_timeout is set, a file handler inside the harvester is closed after `close_timeout` is reached even if the output is blocked. For new which were found for harvesting during the output was blocked, it was possible that the Setup of the file happened but then the initial state could not be sent. So the harvester was never started, file handler was open but `close_timeout` this not apply. Setup and state update are now turned around. This has the affect, that in case of an error during the setup phase, the state must be revert again.

See https://github.com/elastic/beats/pull/3091#issuecomment-283519009

### DIFF
--- a/filebeat/prospector/prospector.go
+++ b/filebeat/prospector/prospector.go
@@ -299,16 +299,24 @@ func (p *Prospector) startHarvester(state file.State, offset int64) error {
 		return err
 	}
 
-	reader, err := h.Setup()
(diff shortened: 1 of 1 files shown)


ruflin:
jenkins, retest it

PR #3721, opened 2017-03-03
@timestamp doesnt get printed when specfied in message codec

vjsamuel:
For the following output config:

```
output.console:
  codec.format:
    string: '%{[@timestamp]} %{[message]}'
```

the output is always empty because there is no switch case in the formatevents to handle `common.Time`

This PR fixes that.

### DIFF
--- a/libbeat/common/datetime.go
+++ b/libbeat/common/datetime.go
@@ -43,6 +43,10 @@ func ParseTime(timespec string) (Time, error) {
 	return Time(t), err
 }
 
+func (t Time) String() string {
+	return time.Time(t).Format(TsLayout)
+}
+
 // MustParseTime is a convenience equivalent of the ParseTime function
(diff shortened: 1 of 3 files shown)


vjsamuel:
@urso can you please provide review comments?

urso:
Some notes about `@timestamp`, not to be solved in this PR:
- MarshalJSON normalizes the `@timestamp` to UTC, but `String()` and the EventFormatter does not. That is, time zones are not used consistently in beats.
- unfortunately dtfmt does not support timezone yet, but I'd prefer formatting (in MarshalJSON and String()) done via dtfmt package. Difference is, dtfmt pre-builds the formatter, but stdlib formatting needs to parse the format string and builds the formatter every single time (well, minor detail)

Current PR seems to be most consistent solution as of now.

PR #3725, opened 2017-03-06
Fix a couple of doc nits

dedemorton:
Fixes a couple of minor doc problems:
* Processors are no longer "future" functionality.
* Users were a little confused about -E option because they thought it might actually change the setting in the config file.


tsg:
I'm ignoring the autopep8 errors on this one.

PR #3726, opened 2017-03-06
Move removal of files to clean

ruflin:
This makes sure all generated files are removed on clean and not before they are regenerated.


metricbeat/Makefile
```diff
@@ -19,11 +19,16 @@ GOX_FLAGS=-arch="amd64 386 arm ppc64 ppc64le"
...
+.PHONY: clean
+clean::
+	# Clean generated dashbord directories
+	rm -rf _meta/kibana/dashboard _meta/kibana/search _meta/kibana/visualization
+	# Clean module docs
+	rm -rf docs/modules
+
 # Collects all module dashboards
 .PHONY: kibana
 kibana:
```
tsg: Should then `kibana` depend on `clean`? Otherwise I'm afraid we might get old dashboards mixed with the new ones.

ruflin: I'm kind of undecided on this PR now looking at it again. It's nice that all clean is one place but at the same time it then requires kibana to depend on clean which I don't like. We should not get old dashboards because of CI always starts on a fresh build and would detect the old files.

tsg: Hmm, yeah, still feels risky to me. Maybe separate a `kibana-clean` target on which both clean and kibana depend?

ruflin:
@tsg I went for duplicating the code instead of introducing an additional target. The important part in this PR is that `clean:` removes all the temp files. It's ok if some of the targest to also some partial cleanup itself.

PR #3727, opened 2017-03-06
Fix get_version command

ruflin:
Command was broken when version file was moved to its own package.


ruflin:
@andrewkroh Yeah, see https://github.com/elastic/beats/pull/3729

PR #3735, opened 2017-03-09
Dockerfile: remove deprecated file

7AC:
The entrypoint script for it was removed in
651012c


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.7.4
+FROM golang:1.8.0
```
tsg: This diff doesn't seem to match the description of the commit?

7AC: My bad, fixed now

PR #3736, opened 2017-03-09
Added gpfsbeat line

itkovian:


tsg:
Nice! Would you be interested in also contributing a Metricbeat module for this one?

itkovian:
Well, this started out as a patch to metric beat, but then I reread the docs, and it seemed more prudent to have this as a separate beat. What would be the advantage/disadvantage having it in metric beat?

ruflin:
Having it in metricbeat or based on metricbeat would mean you don't have to implement the scheduling etc. yourself. Also if it would be in metricbeat itself it would mean users can enable the module with just a config change instead of having to download an other beat.

Curious to hear which part in the docs made you think it is better to go with a completely new beat?

PR #3739, opened 2017-03-12
Monitoring improvements

urso:
- move CollectStructSnapshot to monitoring package for reuse
- have XPack endpoint report all metrics for now
- convert some more missing metrics from expvar to monitoring package


filebeat/registrar/registrar.go
```diff
@@ -28,10 +28,10 @@ type Registrar struct {
 }
 
 var (
-	statesUpdate   = expvar.NewInt("registrar.states.update")
-	statesCleanup  = expvar.NewInt("registrar.states.cleanup")
-	statesCurrent  = expvar.NewInt("registrar.states.current")
-	registryWrites = expvar.NewInt("registrar.writes")
+	statesUpdate   = monitoring.NewInt(nil, "registrar.states.update")
```
ruflin: Looking at all the implementations now it seems like the name of the entry should probably be the first param (nothing to fix in this PR).

urso:
Yeah, naming schema is not perfect. I just exposed all variables, but did not attempt to rename any variable. In some follow-up PR we should have a look at all variables exposed in ES (haven't had time yet) and clean up the variables namings.

PR #3745, opened 2017-03-14
Metricbeat module stats based on libbeat/monitoring

urso:
- replace metricset expvar stats with libbeat.monitoring
- add support to delete stats for unloaded modules/metricsets (with reference counting)

### DIFF
--- a/metricbeat/mb/module/wrapper.go
+++ b/metricbeat/mb/module/wrapper.go
@@ -1,13 +1,13 @@
 package module
 
 import (
-	"expvar"
 	"fmt"
 	"sync"
 	"time"
 
 	"github.com/elastic/beats/libbeat/common"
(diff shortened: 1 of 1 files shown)


metricbeat/mb/module/wrapper.go
```diff
@@ -311,23 +317,39 @@ func writeEvent(done <-chan struct{}, out chan<- common.MapStr, event common.Map
 	}
 }
 
-func getMetricSetExpvarMap(module, name string) (*expvar.Map, error) {
-	key := fmt.Sprintf("%s-%s", module, name)
+func getMetricSetStats(module, name string) *stats {
+	key := fmt.Sprintf("metricbeat.%s.%s", module, name)
```
ruflin: That means that if a module is configured multiple times with the same metricset but different periods, the stats will be combined?

PR #3748, opened 2017-03-14
Always traverse sub-registries

urso:
- always traverse sub-registries to catch all variables at lower leaf-nodes
  being marked as 'Report'


ruflin:
@urso Merged as I don't see an issue with the change. But could you share some insights on what problem this solves?

urso:
Issue was some counters not being reported besides marked as reported, as the traversal did   filter out complete namespaces.

PR #3749, opened 2017-03-15
adding s390x support

kadamvandita:


andrewkroh:
Is this change included in the upstream `golang.org/x/net project`? If not, then this change should go into the upstream project.

Then once the change is in the upstream project we would update the [glide.yaml](https://github.com/elastic/beats/blob/master/glide.yaml#L46-L53) file that tracks the versions of dependencies that we are using and update the entire vendored copy of `golang.org/x/net` to match that commit version.

Also can you please sign the [CLA](http://www.elasticsearch.org/contributor-agreement/)?

kadamvandita:
The change is included in the upstream golang.org/x/net project. But the vendored copy of golang.org/x/net mentioned in  glide.yaml [ https://github.com/elastic/beats/blob/master/glide.yaml#L46-L53 ] does not have the above changes. The vendored copy of golang.org/x/net should be updated to the latest one in the source code of beats and accordingly glide.yaml should also be updated.

tsg:
@kadamvandita That's great news. Do you want us to take over updating the vendored copy? The process to do that is somewhat annoying because glide doesn't really help much. Happy to take over, just give us a sign.

If you want to do it, that's also great, but we'd need you to sign the CLA.

kadamvandita:
I would like to do the changes. Regarding CLA, may I know the list of companies which have signed the CLA. Is IBM a part of the list? Is it fine if I only do " s390x " related changes in the vendored copy?

ruflin:
@kadamvandita For the IBM CLA, best check internally on your side as this is tricky to check on our side. If we update the vendor parts to a new commit, we must update the full package to make sure commit and content correlate. Please be aware that "glide" is a bit broken in our repo, so you would have to update glide.yml manually and copy over the specific commit to our vendor directory.

ruflin:
@kadamvandita I recently updated the dependencies which also included the golang.org/x/net/ package. It seems now the files you are looking for are included. For example:  https://github.com/elastic/beats/blob/master/vendor/golang.org/x/net/ipv4/zsys_linux_s390x.go 

I'm closing this PR as I hope this solves your issue.

PR #3754, opened 2017-03-15
elasticsearch: fix lint issues

7AC:


libbeat/outputs/elasticsearch/api.go
```diff
@@ -160,7 +165,7 @@ func (es *Connection) DeletePipeline(
 	return withQueryResult(es.apiCall("DELETE", "_ingest", "pipeline", id, "", params, nil))
 }
 
-// A search request can be executed purely using a URI by providing request parameters.
+// SearchURI lets execute a search request using a URI by providing request parameters.
```
andrewkroh: `s/lets execute/executes/`?

libbeat/outputs/elasticsearch/api_integration_test.go
```diff
@@ -81,7 +81,7 @@ func TestIngest(t *testing.T) {
 		t.Skip("Skipping tests as pipeline not available in 2.x releases")
 	}
 
-	_, _, err := client.DeletePipeline(pipeline, nil)
+	client.DeletePipeline(pipeline, nil)
```
andrewkroh: Instead of ignoring the `err`, how about checking the error? I think the only acceptable status code would be 404 if the pipeline did not exist.

libbeat/outputs/elasticsearch/client.go
```diff
@@ -472,7 +477,7 @@ func itemStatus(reader *jsonReader) (int, []byte, error) {
 	}
 
 	// parse actual item response code and error message
-	status, msg, err := itemStatusInner(reader)
```
andrewkroh: It was probably an oversight that it didn't check the `err`. Instead of ignoring it, an error check should be added.

libbeat/outputs/elasticsearch/client.go
```diff
@@ -662,10 +670,13 @@ func (conn *Connection) Ping(timeout time.Duration) (string, error) {
 	return response.Version.Number, nil
 }
 
+// Close closes a connection.
+// TODO: it's a noop now
```
andrewkroh: Is this really a "TODO"? If so then let's say what needs to be done to make it complete.

libbeat/outputs/elasticsearch/api.go
7AC: sounds good

libbeat/outputs/elasticsearch/api_integration_test.go
7AC: ok

libbeat/outputs/elasticsearch/client.go
7AC: ok

libbeat/outputs/elasticsearch/api_integration_test.go
```diff
@@ -82,6 +86,9 @@ func TestIngest(t *testing.T) {
 	}
 
 	_, _, err := client.DeletePipeline(pipeline, nil)
+	if err != nil && err.Error() != statusNotFound {
```
andrewkroh: Comparing error stings is brittle, I would use it only when there are no other options. In this case we already have the http status code as an `int` so I think you could use that in combination with the fact that an error occurred.

7AC: Ah cool, didn't realize `DeletePipeline()` returned a status code as well. Thanks!

PR #3756, opened 2017-03-15
metricbeat: add password for redis module in doc

7AC:


metricbeat/metricbeat.full.yml
```diff
@@ -284,6 +284,9 @@ metricbeat.modules:
   # Redis hosts
   #hosts: ["127.0.0.1:6379"]
 
+  # Password to authenticate to the hosts
```
ruflin: The password is already available on line 308. I think it needs to be added to the docs instead.

7AC:
I ran `make update` and `make docs` assuming it would update the `asciidoc` files, I guess I was 👎

ruflin:
@7AC Thanks for the fix. We should backport this to 5.3.

PR #3757, opened 2017-03-15
elasticsearch: set _type=doc

7AC:
Use special `_type=doc` instead of custom ones. The `_type` field is deprecated per https://github.com/elastic/elasticsearch/issues/15613.


ruflin:
Can you link to the elasticsearch / pr of the change to provide some more feedback on why this change is needed?

This must be added to the CHANGELOG.asciidoc as this is a breaking change in case someone was using _type before. This becomes especially obvious for people using filebeat where `document_type` was mapped to `_type`.

ruflin:
Can you rebase and squash and update the commit and PR message? This makes it easier for people that stumble over the PR later to find the right info.

7AC:
I updated the first comment to the PR, I have a single commit I don't think I need squashing. Thanks!

ruflin:
@7AC I see 2 commits in this PR? One is a merge commit.

PR #3762, opened 2017-03-16
filebeat: add some debugs in test_registrar

7AC:
Prompted by two failures that seem unrelated to my changes:
https://travis-ci.org/elastic/beats/jobs/211879514
https://travis-ci.org/elastic/beats/jobs/211881826


ruflin:
Great idea. I was not aware that error messages can be added to the assert command.

PR #3767, opened 2017-03-17
Remove convert old states code

ruflin:
Filebeat 1.x had a different registry format from 5.x. So a conversion of the format happened. This code is now removed for 6.0 as the upgrade path to 6.0 is from 5.0.

The code caused some issues in the past so it is better to not have it in anymore.


CHANGELOG.asciidoc
```diff
@@ -20,6 +20,7 @@ https://github.com/elastic/beats/compare/v5.1.1...master[Check the HEAD diff]
 
 *Filebeat*
 - Always use absolute path for event and registry. This can lead to issues when relative paths were used before. {pull}3328[3328]
+- Remove code to convert states from 1.x.
```
tsg: Link to PR please.

tsg: Ah, still has in `in progress`, nevermind :)

ruflin: good you mentioned it, would have forgotten otherwise ;-)

PR #3770, opened 2017-03-17
dev-tools: fix pylint in cherrypick_pr

7AC:


dev-tools/cherrypick_pr
```diff
@@ -77,10 +79,10 @@ def main():
         check_call("git checkout -b {}".format(tmp_branch), shell=True)
         if call("git cherry-pick -x {}".format(" ".join(args.commit_hashes)),
                 shell=True) != 0:
-            print("Looks like you have cherry-pick errors.")
```
ruflin: This change will break compatibilty with python 3.

dev-tools/cherrypick_pr
```diff
@@ -106,36 +108,35 @@ def main():
     else:
         token = open(expanduser("~/.github_token"), "r").read().strip()
         base = "https://api.github.com/repos/elastic/beats"
-        s = requests.Session()
-        s.headers.update({"Authorization": "token " + token})
+        session = requests.Session()
```
ruflin: +1 on having more descriptive variable names

dev-tools/cherrypick_pr
7AC: Good catch! Do you prefer using `__future__` or ignoring the warning? https://docs.python.org/3.5/library/__future__.html

7AC: I just switched to python3 and restore the parentheses, also renamed `raw_input()` to `input()`

7AC:
Removed the `pylint: ignore` statement and added it to `.pylintrc` in https://github.com/elastic/beats/pull/3763 to avoid conflicts

dev-tools/cherrypick_pr
```diff
@@ -94,7 +96,7 @@ def main():
         return 1
 
     print("Ready to push branch.")
-    remote = raw_input("To which remote should I push? (your fork): ")
+    remote = input("To which remote should I push? (your fork): ")
```
ruflin: Will `input` instead of `raw_input` work in python 2.7?

dev-tools/cherrypick_pr
```diff
@@ -106,33 +108,32 @@ def main():
...
+            print("Creating PR failed: {}".format(request.json()))
             sys.exit(1)
-        new_pr = r.json()
+        new_pr = request.json()
 
         # add labels
-        s.post(base+"/issues/{}/labels".format(new_pr["number"]),
-               json=["backport", "review"])
+        session.post(
+            base + "/issues/{}/labels".format(new_pr["number"]), json=["backport", "review"])
```
tsg: We probably have different settings in the editor, because this version yields an pep error here for me (`line too long (93 > 79 characters)`). I _think_ I am just using the default pep settings.

We didn't yet impose pep/flake in the CI so far because they can be annoying sometimes and the python code that we have is generally not mission critical. Perhaps it's worth discussing this again.

dev-tools/cherrypick_pr
7AC: You're right, that's a risky one I'll revert it.

dev-tools/cherrypick_pr
7AC: We pass it `--max-line-length 120` in the `check` target, I'll adjust my settings to that. Thanks

PR #3776, opened 2017-03-20
Make system process metricset honor cpu_ticks config option

### LINKED ISSUE #3590: Impossible to get "system.process.cpu.total.ticks" field
It seems like there is no way to activate the "system.process.cpu.total.ticks" field with Metricbeat. Tested with version 5.1.2 and 5.2.0, the "cpu_ticks" option doesn't work. It only adds ticks information in "cpu" and "core" modules events.
Either the option should also activate the fields for "process" module, either we need an additional option. 

- Version:  5.1.2 and 5.2.0 
- Operating System: Linux
- Steps to Reproduce: Start Metricbeat with "process" module activated and "cpu_ticks: true" option. Check in Elasticsearch/Kibana the content for "process" events. The "system.process.cpu.total.ticks" field is not present.

andrewkroh:
When `cpu_ticks` is true in the config file, the system process metricset will report the number of “ticks” (actually jiffies on linux) used in user space (`system.process.cpu.user`), system (`system.process.cpu.system`), and total (`system.process.cpu.total.ticks`).

Fixes #3590

### DIFF
--- a/metricbeat/module/system/process/helper.go
+++ b/metricbeat/module/system/process/helper.go
@@ -245,23 +245,17 @@ func (procStats *ProcStats) GetProcessEvent(process *Process, last *Process) com
 		proc["env"] = process.Env
 	}
 
+	proc["cpu"] = common.MapStr{
+		"total": common.MapStr{
+			"pct": GetProcCpuPercentage(last, process),
+		},
+		"start_time": unixTimeMsToTime(process.Cpu.StartTime),
+	}
+
 	if procStats.CpuTicks {
-		proc["cpu"] = common.MapStr{
-			"user":   process.Cpu.User,
-			"system": process.Cpu.Sys,
-			"total": common.MapStr{
-				"ticks": process.Cpu.Total,
-				"pct":   GetProcCpuPercentage(last, process),
-			},
-			"start_time": unixTimeMsToTime(process.Cpu.StartTime),
-		}
-	} else {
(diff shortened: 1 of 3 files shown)


metricbeat/module/system/process/helper.go
```diff
@@ -245,23 +245,17 @@ func (procStats *ProcStats) GetProcessEvent(process *Process, last *Process) com
...
-			"start_time": unixTimeMsToTime(process.Cpu.StartTime),
-		}
-	} else {
-		proc["cpu"] = common.MapStr{
-			"total": common.MapStr{
-				"pct": GetProcCpuPercentage(last, process),
-			},
-			"start_time": unixTimeMsToTime(process.Cpu.StartTime),
-		}
+		proc.Put("cpu.user", process.Cpu.User)
```
ruflin: nice, that also cleans up the code. It bugs me that we only use `.ticks` for the total but that would be a BC break.

ruflin:
@andrewkroh I added `needs backport` label as I think we should backport it to at least 5.3.

metricbeat/module/system/process/helper.go
andrewkroh: I felt the same way, but didn't want to break BC.

ruflin: Lets keep it in mind to "potentially" break for 6.0 ;-)

PR #3778, opened 2017-03-21
Remove `_all` field from template for ES >= 6.0

exekias:
Elasticsearch 6 deprecates `_all` field, this PR adds new templates for ES6 (removing `_all`), and moves 5.X ones back to `*.template-es5x.json`, following the same logic from es2x.

This change is only for 5.x branch, as master has template generation in place and I will have to tackle this from the code. Ideally I can wait for https://github.com/elastic/beats/pull/3681 to be merged so we forget about shipping one file per ES version.

### DIFF
--- a/filebeat/filebeat.full.yml
+++ b/filebeat/filebeat.full.yml
@@ -510,6 +510,14 @@ output.elasticsearch:
   # Path to the Elasticsearch 2.x version of the template file.
   #template.versions.2x.path: "${path.config}/filebeat.template-es2x.json"
 
+  # If set to true, filebeat checks the Elasticsearch version at connect time, and if it
+  # is 6.x, it loads the file specified by the template.versions.6x.path setting. The
+  # default is true.
+  #template.versions.6x.enabled: true
+
+  # Path to the Elasticsearch 6.x version of the template file.
+  #template.versions.6x.path: "${path.config}/filebeat.template-es6x.json"
+
   # Use SSL settings for HTTPS. Default is true.
   #ssl.enabled: true
 
--- a/libbeat/outputs/elasticsearch/client_integration_test.go
+++ b/libbeat/outputs/elasticsearch/client_integration_test.go
@@ -48,6 +48,8 @@ func TestLoadTemplate(t *testing.T) {
 	templatePath := absPath + "/libbeat.template.json"
 	if strings.HasPrefix(client.Connection.version, "2.") {
 		templatePath = absPath + "/libbeat.template-es2x.json"
+	} else if strings.HasPrefix(client.Connection.version, "6.") {
+		templatePath = absPath + "/libbeat.template-es6x.json"
 	}
 	content, err := readTemplate(templatePath)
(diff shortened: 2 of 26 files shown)


7AC:
Is there a PR for the ES change?

exekias:
This is the change in ES: https://github.com/elastic/elasticsearch/pull/22144

filebeat/filebeat.template.json
```diff
@@ -1,9 +1,6 @@
 {
   "mappings": {
     "_default_": {
-      "_all": {
```
ruflin: We should probably now call this template `fielbeat.template.6x.json`. Otherwise I think it is confusing that the "default" template is not for the stable version.

filebeat/filebeat.full.yml
```diff
@@ -510,6 +510,14 @@ output.elasticsearch:
   # Path to the Elasticsearch 2.x version of the template file.
   #template.versions.2x.path: "${path.config}/filebeat.template-es2x.json"
 
+  # If set to true, filebeat checks the Elasticsearch version at connect time, and if it
+  # is 5.x, it loads the file specified by the template.versions.5x.path setting. The
+  # default is true.
+  #template.versions.5x.enabled: true
+
+  # Path to the Elasticsearch 5.x version of the template file.
+  #template.versions.5x.path: "${path.config}/filebeat.template-es5x.json"
```
ruflin: See my comment below about the naming. For the next months until 6 is GA, loading the 5.x is going to be the expected default. I'm thinking if we should have path config now for all versions? Problem is that we add even more config options (which will be remove afterwards) :-( WDYT?

exekias: Yep, I see the issue, had my doubts when setting it, perhaps the best thing we can do is switching to template-es6x and keep template.json as the default. Then deprecate this in 6.0 and move to the path based solution? Doing it now would break config compatibility

exekias:
jenkins, package it please

exekias:
Ready for a second review :)

libbeat/outputs/elasticsearch/client_integration_test.go
```diff
@@ -161,6 +165,8 @@ func TestOutputLoadTemplate(t *testing.T) {
 
 	if strings.HasPrefix(client.Connection.version, "2.") {
 		templatePath = "../../libbeat.template-es2x.json"
+	} else if strings.HasPrefix(client.Connection.version, "5.") {
```
tsg: Typo here? Should it be `6.`?

tsg:
LGTM, apart from the minor comment above. Also, please add a CHANGELOG line.

tsg:
Nice work!

ruflin:
@exekias Nice work. Did you manually test this one with elasticsearch master? (manually)

exekias:
Yes I did, worked for me :) I also checked how it was failing before my change

ruflin:
Awesome 🎉

Issue #3779, opened 2017-03-21
Filebeat race condition start / stop harvester


ruflin:
CI on AppVeyor recently run into a race condition during the tests. The race condition seems to be cause by Stopping a harvester before it was fully started, means `Wait()` is called before all `Add(1)` are called on the waitgroup. See full log messages below.

Running `test_shutdown.py:Test.test_shutdown_wait_ok` can lead to the following log output / race condition:

```
2017/03/20 16:05:50.682951 beat.go:319: INFO Home path: [C:\gopath\src\github.com\elastic\beats\filebeat] Config path: [C:\gopath\src\github.com\elastic\beats\filebeat] Data path: [C:\gopath\src\github.com\elastic\beats\filebeat\data] Logs path: [C:\gopath\src\github.com\elastic\beats\filebeat\logs]
2017/03/20 16:05:50.682951 metrics.go:23: INFO Metrics logging every 30s
2017/03/20 16:05:50.682951 beat.go:344: INFO Beat metadata path: C:\gopath\src\github.com\elastic\beats\filebeat\data\meta.json
2017/03/20 16:05:50.683952 beat.go:326: INFO Beat UUID: a9a026d7-5da0-4904-aee5-dfb1bce68f32
2017/03/20 16:05:50.683952 beat.go:208: INFO Setup Beat: filebeat; Version: 6.0.0-alpha1
2017/03/20 16:05:50.683952 processor.go:44: DBG  Processors: 
2017/03/20 16:05:50.683952 beat.go:214: DBG  Initializing output plugins
2017/03/20 16:05:50.683952 file.go:54: INFO File output path set to: C:\gopath\src\github.com\elastic\beats\filebeat\build\system-tests\run\test_shutdown.Test.test_shutdown_wait_ok/output
2017/03/20 16:05:50.683952 file.go:55: INFO File output base filename set to: filebeat
2017/03/20 16:05:50.683952 file.go:58: INFO Rotate every bytes set to: 1024000
2017/03/20 16:05:50.683952 file.go:62: INFO Number of files set to: 7
2017/03/20 16:05:50.683952 outputs.go:117: INFO Activated file as output plugin.
2017/03/20 16:05:50.683952 publish.go:236: DBG  Create output worker
2017/03/20 16:05:50.684952 publish.go:278: DBG  No output is defined to store the topology. The server fields might not be filled.
2017/03/20 16:05:50.684952 publish.go:293: INFO Publisher name: APPVYR-WIN
[...]

PR #3780, opened 2017-03-21
Fix race condition for harvester Start / Stop in registry

ruflin:
There is still a case where a race condition could happen (see TODO). Registry must be improved to also handle this case.

This should not be backported as the issue can only happen on shutdown in very rare cases and should not have any side effect.

Closes https://github.com/elastic/beats/issues/3779

### DIFF
--- a/filebeat/harvester/harvester.go
+++ b/filebeat/harvester/harvester.go
@@ -53,7 +53,7 @@ type Harvester struct {
 	encoding        encoding.Encoding
 	done            chan struct{}
 	stopOnce        sync.Once
-	stopWg          *sync.WaitGroup
+	StopWg          *sync.WaitGroup
 	outlet          Outlet
 	ID              uuid.UUID
 	processors      *processors.Processors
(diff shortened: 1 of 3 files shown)


7AC:
CI is failing with an error I also got in another review so I think it's unrelated to your changes

```
ERROR: Test that Filebeat works when stdin is closed.
  File "C:\gopath\src\github.com\elastic\beats\filebeat\tests\system\test_prospector.py", line 139, in test_stdin_eof
    self.wait_until(lambda: self.output_has(lines=2))
  File "C:\gopath\src\github.com\elastic\beats\filebeat\tests\system\../../../libbeat/tests/system\beat\beat.py", line 283, in wait_until
    "Waited {} seconds.".format(max_timeout))
Exception: Timeout waiting for 'cond' to be true. Waited 10 seconds.
```

ruflin:
jenkins, retest it

ruflin:
Closing in favor of https://github.com/elastic/beats/pull/4314 Too many things have changed in the meantime.

PR #3781, opened 2017-03-21
Improve handling of different path variants on Windows

ruflin:
If forward slashes were used on Windows the glob was not matching on startup which lead to the issue that data was resent. This is solved in 5.3 by using path.Abs for all paths which also includes Clean. To make sure Path and Glob are always clean, Cleanup was added to the MatchFile part. This makes sure in case old data / incorrect data is loaded, the comparison will still work. More tests were added for windows to verify change.

Before this change, option 1 below did not work.

```
"F:/wwwLogs/flights-wsapi/WebServicesWebApi.log"
"F:\\wwwLogs\\flights-wsapi\\WebServicesWebApi.log"
'F:\wwwLogs\flights-wsapi\WebServicesWebApi.log'
```

Based on https://discuss.elastic.co/t/duplicate-events-with-filebeat-on-windows-on-service-restart/78743/10

### DIFF
--- a/filebeat/prospector/prospector_log.go
+++ b/filebeat/prospector/prospector_log.go
@@ -4,8 +4,6 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"runtime"
-	"strings"
 	"time"
 
 	"github.com/elastic/beats/filebeat/harvester"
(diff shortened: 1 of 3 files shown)


ruflin:
This should be backported to 5.3.

filebeat/prospector/prospector_log.go
```diff
@@ -182,11 +180,9 @@ func (p *ProspectorLog) getFiles() map[string]os.FileInfo {
...
 	for _, glob := range p.config.Paths {
 
-		if runtime.GOOS == "windows" {
-			// Windows allows / slashes which makes glob patterns with / work
-			// But for match we need paths with \ as only file names are compared and no lookup happens
-			glob = strings.Replace(glob, "/", "\\", -1)
-		}
+		// Path and glob are cleaned to ensure we always compare clean paths
+		glob = filepath.Clean(glob)
+		filePath = filepath.Clean(filePath)
```
andrewkroh: It looks like cleaning `filePath` could be done only once if it were outside of the `for` loop.

PR #3788, opened 2017-03-22
Make http fields in HAProxy optional to improve compatibility with 1.5

ruflin:
It seems the http fields do not always exists in HAProxy 1.5. We currently only support 1.6 but as this will not break things, I made the fields optional for better compatibility.

### DIFF
--- a/metricbeat/module/haproxy/stat/data.go
+++ b/metricbeat/module/haproxy/stat/data.go
@@ -49,12 +49,12 @@ var (
 			"time.avg": c.Int("Rtime", s.Optional),
 			"denied":   c.Int("Dresp"),
 			"http": s.Object{
-				"1xx":   c.Int("Hrsp1xx"),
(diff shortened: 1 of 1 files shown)


exekias:
LGTM, I launched a manual test that [passed](http://build-eu-00.elastic.co/job/Beats-PR/5427/), is there any quick cheat to test it again for the PR?

tsg:
jenkins, test it

tsg:
@exekias ^^- like that

exekias:
ready to go, can you add a CHANGELOG entry?

ruflin:
@exekias I didn't add a Changelog on purpose as I don't see this as a bug fixes as it is an improvement for a non supported version. If we add a changelog under "Added" I worry that we give the impression that we support 1.5. So I tried to keep this under the radar ;-)

PR #3790, opened 2017-03-22
Introduce error reporting in schema

ruflin:
Currently in schema errors are only reported through logging which makes testing difficult. This PR introduces the collection of errors during applying the schema.

Now Apply returns a list of errors. Currently these are ignored to keep the same implementation as before. Over time this should be adjusted to handle errors correctly. Logging in the schema should be remove and moved up to the Metricsets itself.

This change should also simplify testing of different versions in unit tests, as `schema.Apply` will return errors. Currently this check is done in the system tests checking the log itself which is not very efficient.

ApplyNoError was introduced as a temporary workaround for the processors implementation.

This is part of https://github.com/elastic/beats/issues/3807

### DIFF
--- a/metricbeat/schema/error.go
+++ b/metricbeat/schema/error.go
@@ -1,0 +1,36 @@
+package schema
+
+import "fmt"
+
+const (
+	RequiredType ErrorType = iota
+	OptionalType ErrorType = iota
+)
+
+type ErrorType int
+
+type Error struct {
+	key       string
+	message   string
+	errorType ErrorType
+}
+
+func NewError(key string, message string) *Error {
+	return &Error{
+		key:       key,
+		message:   message,
+		errorType: RequiredType,
+	}
+}
+
+func (err *Error) SetType(errorType ErrorType) {
+	err.errorType = errorType
+}
+
+func (err *Error) IsType(errorType ErrorType) bool {
+	return err.errorType == errorType
+}
+
+func (err *Error) Error() string {
+	return fmt.Sprintf("Missing field: %s, Error: %s", err.key, err.message)
+}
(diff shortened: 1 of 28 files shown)


metricbeat/schema/mapstriface/mapstriface.go
```diff
@@ -71,18 +72,22 @@ type ConvMap struct {
...
+func (convMap ConvMap) Map(key string, event common.MapStr, data map[string]interface{}) multierror.Errors {
 	subData, ok := data[convMap.Key].(map[string]interface{})
 	if !ok {
 		if !convMap.Optional {
-			logp.Err("Error accessing sub-dictionary `%s`", convMap.Key)
+			err := fmt.Errorf("Error accessing sub-dictionary `%s`", convMap.Key)
+			logp.Err("%s", err)
+			errs := multierror.Errors{}
+			errs = append(errs, err)
+			return errs
```
tsg: Could these lines be written as `multierror.Errors{err}`? Would be nicer if it works.

tsg:
I like the idea. Generally I'd suggest making this be the `Apply` method to keep the interface clean and just ignore the errors for now in all the modules where we don't yet care about the answer. In that case, each module needs an update, but the change is minimal.

metricbeat/schema/error.go
```diff
@@ -0,0 +1,34 @@
...
+
+const (
+	RequiredType = iota
+	OptionalType = iota
+)
+
+type Error struct {
+	key       string
+	message   string
+	errorType int
```
tsg: Would it make sense to make ErrorType a type so that it's restricted to the types above?

ruflin: done

PR #3793, opened 2017-03-22
filebeat: fix golint warnings in prospector

7AC:
The somewhat bigger change was renaming some of the public structs so the names would make sense with the `prospector.` package qualifier.

### DIFF
--- a/filebeat/prospector/prospector_log.go
+++ b/filebeat/prospector/prospector_log.go
@@ -18,14 +18,16 @@ var (
 	filesTruncated = monitoring.NewInt(nil, "filebeat.prospector.log.files.truncated")
 )
 
-type ProspectorLog struct {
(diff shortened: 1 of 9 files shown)


filebeat/prospector/prospector_log.go
```diff
@@ -18,14 +18,16 @@ var (
 	filesTruncated = monitoring.NewInt(nil, "filebeat.prospector.log.files.truncated")
 )
 
-type ProspectorLog struct {
+// Log contains the prospector and its config
+type Log struct {
```
tsg: Curious, what's the golint message for this change?

7AC: "type name will be used as prospector.ProspectorLog by other packages, and that stutters; consider calling this Log (golint)"

PR #3795, opened 2017-03-22
filebeat: use zglob for wildcards

7AC:


glide.yaml
```diff
@@ -32,7 +32,6 @@ import:
   version: ca8abbf65d0e61dedf061f98bd3850f250e27539
 - package: github.com/samuel/go-thrift
   version: 2187045faa54fce7f5028706ffeb2f2fc342aa7e
-  # This is the same as labix.org/v2/mgo. See https://github.com/Masterminds/glide/issues/221#issuecomment-179817757
```
tsg: This seems to have gotten lost. Is there a way to keep it besides manually editing the file?

tsg:
What worries me about the zglob library is that it uses fewer tests than the one from the standard library. Compare [zglob](https://github.com/mattn/go-zglob/blob/master/zglob_test.go#L24) with [stdlib](https://golang.org/src/path/filepath/match_test.go). Is there any indication that the functionality is a super-set of what we had before?

glide.yaml
7AC: Sounds like there isn't: https://github.com/Masterminds/glide/issues/691
I'll re-add it

7AC:
Ouch 29/54 of those patterns fail with `zlog`, I need to investigate more.

filebeat/prospector/prospector_log.go
```diff
@@ -11,6 +11,7 @@ import (
 	"github.com/elastic/beats/filebeat/input/file"
 	"github.com/elastic/beats/libbeat/logp"
 	"github.com/elastic/beats/libbeat/monitoring"
+	"github.com/mattn/go-zglob"
```
ruflin: We normally put a new line between external and internal imports.

glide.yaml
```diff
@@ -115,3 +115,4 @@ import:
   version: 5a004441f897722c627870a981d02b29924215fa
 - package: github.com/mitchellh/hashstructure
   version: b098c52ef6beab8cd82bc4a32422cf54b890e8fa
+- package: github.com/mattn/go-zglob
```
ruflin: In case we are going to use an external package, the exact commit id or version should be added.

glide.yaml
```diff
@@ -93,12 +93,12 @@ import:
 - package: github.com/elastic/go-lumber
   version: v0.1.0
   subpackages:
-  - /lj
-  - /log
-  - /client/v2
-  - /server/internal
-  - /server/v2
-  - /protocol/v2
+  - lj
```
ruflin: Did this syntax change in glide?

glide.yaml
7AC: Yep not sure how that was missed. I'm going to revisit this anyways since `zlog` is clearly not a drop in replacement

filebeat/prospector/prospector_log.go
7AC: Thanks

glide.yaml
7AC: I guess so

7AC:
We're coming up with a new solution, closing this

PR #3796, opened 2017-03-23
Added new response validator to check body contains a string

nick-purdy:
Needed a validator to check if a string matches within the body of the response. Useful in the case of large responses or inconsistent white spacing.

### DIFF
--- a/heartbeat/heartbeat.full.yml
+++ b/heartbeat/heartbeat.full.yml
@@ -189,6 +189,9 @@ heartbeat.monitors:
     # Required response contents.
     #body:
 
+    # Required response contents must contain value.
+    #body_contains:
+
 heartbeat.scheduler:
   # Limit number of concurrent tasks executed by heartbeat. The task limit if
   # disabled if set to 0. The default is 0.
--- a/heartbeat/_meta/beat.full.yml
+++ b/heartbeat/_meta/beat.full.yml
@@ -189,6 +189,9 @@ heartbeat.monitors:
     # Required response contents.
     #body:
 
+    # Required response contents must contain value.
+    #body_contains:
+
 heartbeat.scheduler:
   # Limit number of concurrent tasks executed by heartbeat. The task limit if
   # disabled if set to 0. The default is 0.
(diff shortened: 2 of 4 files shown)


nick-purdy:
I've just signed the CLA. Not sure why it hasn't updated, perhaps it just takes time!

karmi:
Hi @nick-purdy, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/beats/pull/3796.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

nick-purdy:
I thought it was the same, but I just added my second email just in case.

tsg:
@nick-purdy Thanks for the PR! Looks good to me, but I'd like to wait for @urso's review as well.

In the meantime, can you run `go fmt` on your changes to make the tests pass?

heartbeat/heartbeat.full.yml
```diff
@@ -189,6 +189,9 @@ heartbeat.monitors:
     # Required response contents.
     #body:
 
+    # Required response contents must contain value.
+    #body_contains:
```
tsg: This file is generated. You need to modify in `heartbeat/_meta/beat.full.yml` and then run `make update`.

karmi:
@nick-purdy, thanks for adding the e-mail, it has fixed the failing CLA check!

heartbeat/heartbeat.full.yml
nick-purdy: Thanks! I've updated the correct file and make update'd now.

urso:
I'm a little on the fence about HTTP validation as is. I'm flirting with the idea of reusing conditionals support from beats processors. Anyway, `contains` should capture a many more use-cases then the way `body` is currently implemented.

The PR as is will require `body` and `body_contains` to be used exclusively (not enforced when unpacking the configuration), due to the use of `ioutil.ReadAll`. But I'd prefer you to change the `body` setting to always use `contains` and drop` body_contains`. This should capture a many more use-case (e.g. 'large body content', dynamic banner with timestamp/version).

andrewvc:
Thanks for the contribution. Sorry we didn't merge this. It looks like we added this functionality some time ago. See the docs [here](https://www.elastic.co/guide/en/beats/heartbeat/current/configuration-heartbeat-options.html#monitor-http-check) then scroll down a bit to body.

PR #3797, opened 2017-03-23
Use single quotes in example fields that accept regexps

exekias:
We recommend single quotes in the documentation already, this
should help avoiding mistakes from the user when adding complex
patterns that do escaping

### DIFF
--- a/filebeat/_meta/common.full.p2.yml
+++ b/filebeat/_meta/common.full.p2.yml
@@ -36,16 +36,16 @@ filebeat.prospectors:
   # Exclude lines. A list of regular expressions to match. It drops the lines that are
   # matching any regular expression from the list. The include_lines is called before
   # exclude_lines. By default, no lines are dropped.
-  #exclude_lines: ["^DBG"]
+  #exclude_lines: ['^DBG']
 
   # Include lines. A list of regular expressions to match. It exports the lines that are
   # matching any regular expression from the list. The include_lines is called before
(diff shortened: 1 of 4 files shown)


exekias:
I also took things for granted there :P https://discuss.elastic.co/t/beginner-question-searching-for-string-containing-chars/79696

tsg:
Nice, I didn't single vs double quotes are different in YAML. There's another place where we accept regexp, in the `multiline.pattern` option. For consistency, can you also update that one to use the same syntax?

exekias:
I saw it, but currently we don't quote it at all: https://github.com/exekias/beats/blob/f755701d430033177f076bde1f5c15a83051cefe/filebeat/_meta/common.full.p2.yml#L113

you want me to add quotes?

tsg:
Probably safer to leave it as is, as per YAML recommendation to don't use unless they are needed. I merged the PR as is.

urso:
regexp can contain 'special' symbols which can trigger some 'parsing' in YAML. That's why it's always recommended to use single-quotes, even when using multiline. e.g. using `[]` or `:` in a regexp can trigger parsing failures. Using single-quotes ensures YAML is passing the 'literal' string to the regexp compiler. We even have this in our filebeat FAQ.

Did we check docs consistently using single-quotes? I'd vote for single-quotes in multiline as well.

In general I'd prefer using single for a number of strings (e.g. file-path, so windows users having no idea about YAML get the correct behavior when copy-pasting a path).

Issue #3783, opened 2017-03-21
Bad hyperlink syntax in Format Strings docs


andrewkroh:
See https://www.elastic.co/guide/en/beats/libbeat/current/config-file-format-type.html#_format_string_sprintf

<img width="785" alt="screen shot 2017-03-21 at 11 16 22 am" src="https://cloud.githubusercontent.com/assets/4565752/24154549/da64e634-0e27-11e7-85bf-7ab864459c9c.png">

dedemorton:
Fixed in #3803

Issue #3649, opened 2017-02-22
Add conditional coding to install instructions so pre-release versions don't show broken links


dedemorton:
Right now, the docs in `master` give a download path to software that's unavailable. Some users end up looking at the doc in `master` without reading the text at the top of the page that clearly states that the doc is for an unreleased version. So they try to download the software, and find that the package doesn't exist.

We can address this confusion by adding conditional coding to the asciidoc source that clearly states the software isn't released.

Here's an example of how the ES docs handle this issue:

`````
[[install-zip]]
==== Download and install the `.zip` package

ifeval::["{release-state}"=="unreleased"]

Version {version} of Elasticsearch has not yet been released.

endif::[]

ifeval::["{release-state}"!="unreleased"]

The `.zip` archive for Elasticsearch v{version} can be downloaded and installed as follows:


["source","sh",subs="attributes"]
--------------------------------------------
wget https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-{version}.zip
sha1sum elasticsearch-{version}.zip <1>
unzip elasticsearch-{version}.zip
cd elasticsearch-{version}/ <2>
--------------------------------------------
<1> Compare the SHA produced by `sha1sum` or `shasum` with the
    https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-{version}.zip.sha1[published SHA].
<2> This directory is known as `$ES_HOME`.

endif::[]
`````

dedemorton:
Fixed in #3803

PR #3803, opened 2017-03-24
Misc doc changes

dedemorton:
This PR closes issue #3783 and #3649. Also fixes a heading level that incorrect.

@tsg Note that after we merge this PR, we'll probably want to add something to the release checklist to verify that `:release-state: released` is set in version.asciidoc. I added the `release-state` variable to version.asciidoc so that we can change the state in one place. It's less visible there, tho, so we'll just need to make a note to check it when we're publishing the docs for a release.


tsg:
Sounds good, I updated the checklist.

Issue #3539, opened 2017-02-06
Beats Doc Updates for 5.3


dedemorton:
**5.3:**

- [x] Make sure these sections get edited:
  - [x] Config reload (backlog. See #3281)
  - [x] Docs about -setup for loading dashboards (see https://github.com/elastic/beats/pull/3506).
- [x] Heartbeat docs: Add content about loading sample dashboards when the dashboards are available. 
- [x] Check cURL statements used in docs to make sure they include content-type when passing a body.
- [x] Make these minor fixes to the getting started guides:
  - [x] Mention that command line options aren’t available when using init.d script. See https://github.com/elastic/beats/issues/812 (also contains some info relevant to issue #1047)
  - [x] Filebeat GS:  move step about importing the index pattern in Filebeat into the doc for Filebeat rather than making users jump to the platform ref for a simple command.
- [x] Fix bugs flagged for 5.3 plus: 
  - [x] Clarify Filebeat docs wrt the index setting (see https://github.com/elastic/beats/issues/3558).
- [x] Make sure our Beats developer guide is up-to-date and that the Metricbeat docs point to the docs about the single generator (we now have a single generator for generating either a Beat or a Beat based on Metricbeat--see https://github.com/elastic/beats/pull/3491). Make sure the blog post mentioned in the PR has been updated.

**Follow up:**
- [x] See if we can remove close this issue: https://github.com/elastic/beats/issues/652

dedemorton:
Closing this and moving the backlog https://github.com/elastic/beats/issues/4021

PR #3804, opened 2017-03-24
Add docs about loading heartbeat dashboard

dedemorton:
@urso You might want to add a more interesting screen capture of Heartbeat data. 

BTW, I use "dashboards" plural in case we add more later.


dedemorton:
Adds dashboard content for item tracked in https://github.com/elastic/beats/issues/3539 meta issue.

tsg:
You can find a somehow more colorful dashboard screenshot here: https://github.com/elastic/beats/pull/3531

Also, the new screenshot for `kibana-created-indexes.png` seems wrong to me. Perhaps you replaced by mistake?

dedemorton:
@tsg I've replaced the Heartbeat dashboard. There were a few minor changes to the Kibana UI after you added the dashboard, but I think that having realistic looking data is more important than having the UI match up exactly (it's not too far off). 

The other screen capture, kibana-created-indexes.png, probably looks wrong because I replaced an old screen shot (which showed how to set an index as the default) with a screenshot of the Discover page. The old screen shot was leftover from me copying files to use as a template. We don't tell users to set an index as the default anymore because that happens automatically. So the new screenshot is the right one...it just looks weird you compare it to the old one.

Issue #3074, opened 2016-11-29
Metricbeat: Apache 2.4.23 field not found errors


profilernz:
I'm getting:

metricbeat[6664]: schema.go:36: Error on field 'total': Key ConnsTotal not found
metricbeat[6664]: schema.go:36: Error on field 'closing': Key ConnsAsyncClosing not found
metricbeat[6664]: schema.go:36: Error on field 'writing': Key ConnsAsyncWriting not found
metricbeat[6664]: schema.go:36: Error on field 'keep_alive': Key ConnsAsyncKeepAlive not found

Running Metricbeat 5.0.1 with Apache 2.4.23 (from https://ius.io/) on Centos 7:

httpd -version
Server version: Apache/2.4.23 (CentOS)
Server built: Jul 22 2016 09:11:01

metricbeat -version
metricbeat version 5.0.1 (amd64), libbeat 5.0.1

curl http://127.0.0.1:81/server-status?auto
127.0.0.1
ServerVersion: Apache/2.4.23 (CentOS) OpenSSL/1.0.1e-fips
ServerMPM: prefork
Server Built: Jul 22 2016 09:11:01
CurrentTime: Wednesday, 30-Nov-2016 05:32:43 NZDT
RestartTime: Tuesday, 29-Nov-2016 20:41:30 NZDT
ParentServerConfigGeneration: 1
ParentServerMPMGeneration: 0
ServerUptimeSeconds: 31873
ServerUptime: 8 hours 51 minutes 13 seconds
Load1: 0.00
Load5: 0.01
Load15: 0.05
Total Accesses: 2181
Total kBytes: 53079
CPUUser: 8.4
CPUSystem: 1.22
CPUChildrenUser: 0
CPUChildrenSystem: 0
CPULoad: .0301823
Uptime: 31873
ReqPerSec: .0684278
BytesPerSec: 1705.3
BytesPerReq: 24921.1
BusyWorkers: 1
IdleWorkers: 9
Scoreboard: __.W._____....................................................................................................................................................................................................................................................
TLSSessionCacheStatus
CacheType: SHMCB
CacheSharedMemory: 512000
CacheCurrentEntries: 0
CacheSubcaches: 32
CacheIndexesPerSubcaches: 88
CacheIndexUsage: 0%
CacheUsage: 0%
CacheStoreCount: 35
CacheReplaceCount: 0
CacheExpireCount: 35
CacheDiscardCount: 0
CacheRetrieveHitCount: 0
[...]

PR #3815, opened 2017-03-27
Cherry-pick to 5.3: Make some Apache server status fields optional

exekias:
As they are not reported by some systems

Backport from #3808

### DIFF
--- a/metricbeat/module/apache/status/data.go
+++ b/metricbeat/module/apache/status/data.go
@@ -38,17 +38,17 @@ var (
 			"children_system": c.Float("CPUChildrenSystem"),
 		},
 		"connections": s.Object{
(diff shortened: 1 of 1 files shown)


ruflin:
@exekias Missed in the previous PR but we should add a CHANGELOG entry to this one.

exekias:
I followed the same logic behind https://github.com/elastic/beats/pull/3788#issuecomment-288701686, but I agree with adding it anyway :)

ruflin:
I see. Problem with that is that we marked is a bug here: https://github.com/elastic/beats/issues/3074 I think this also fixes some problems with a version we list as "supported"?

exekias:
Reopened with changelog (and correct branch) https://github.com/elastic/beats/pull/3816

PR #3819, opened 2017-03-27
Add tests for different Apache status outputs

exekias:
This unit tests output mapping from different Apache2 version (and underlying systems). Related to #3808


ruflin:
Very nice. This makes it very easy to test for other outputs in the future.

PR #3821, opened 2017-03-28
Cherrypick doc changes into 5.3

dedemorton:
Cherry-picks #3803 into 5.3

@tsg Since we are so close, I set `release-state `to `released`, so you don't have to worry about that for 5.3.


tsg:
Thanks!

Issue #3785, opened 2017-03-22
Module/Prospector level processor support


vjsamuel:
I understand that today we only support `include_lines`, `exclude_lines`, and `exclude_files` as processors at a prospector level and support `filters` at metricbeat module level where a user can provide any processor as a filter for the given module. This is a great feature to have in filebeat as well. Can the same be extended to filebeat where a user can say:
```
filters:
  - processor1
  - processor2
```

This is especially useful when custom processing needs to be done for specific log files. Putting them as global processors only slows down the pipeline as each and every log line needs to be verified.

ruflin:
We will also need this to better support the reloading feature. Currently it is not possible to add a new prospector and have a specific processor for it.

I remember we wanted to add this in the past but there was an issue that the event format on the processor level is not in the format the processor expects it. So it could be that some previous refactoring is needed.

vjsamuel:
not valid anymore :)

PR #3823, opened 2017-03-28
Refactor input.Event similar to outputs.Data

vjsamuel:
Kindly review this change as this is a feature that we're looking for when we use filebeat :) If the code looks good, I can also fix the test cases.

https://github.com/elastic/beats/issues/3785

cc: @ruflin

### DIFF
--- a/filebeat/input/event.go
+++ b/filebeat/input/event.go
@@ -11,24 +11,38 @@ import (
 
 // Event is sent to the output and must contain all relevant information
 type Event struct {
-	common.EventMetadata
-	ReadTime     time.Time
-	InputType    string
-	DocumentType string
-	Bytes        int
+	*EventMeta
 	Text         *string
 	JSONConfig   *reader.JSONConfig
-	State        file.State
 	Data         common.MapStr // Use in readers to add data to the event
+
+}
+
+type EventMeta struct {
 	Pipeline     string
 	Fileset      string
 	Module       string
+	InputType    string
+	DocumentType string
+	common.EventMetadata
+	ReadTime     time.Time
+	Bytes        int
+	State        file.State
+}
+
+type EventHolder struct {
+	Event common.MapStr
+	Metadata EventMeta
+
 }
 
 func NewEvent(state file.State) *Event {
 	return &Event{
-		State: state,
+		EventMeta: &EventMeta{
+			State: state,
+		},
 	}
+
 }
 
 func (e *Event) ToMapStr() common.MapStr {
@@ -40,6 +54,7 @@ func (e *Event) ToMapStr() common.MapStr {
 		"type":                  e.DocumentType,
 		"input_type":            e.InputType,
 	}
+
 
 	if e.Fileset != "" && e.Module != "" {
 		event["fileset"] = common.MapStr{
@@ -65,19 +80,31 @@ func (e *Event) ToMapStr() common.MapStr {
 		event["message"] = *e.Text
 	}
 
-	meta := e.Metadata(); if meta != nil {
-		event["meta"] = meta
-	}
-
 	return event
 }
 
+func (e *Event) GetEventHolder() EventHolder {
+	return EventHolder{
+		Event: e.ToMapStr(),
+		Metadata: EventMeta{
+			Pipeline: e.Pipeline,
+			Bytes: e.Bytes,
+			State: e.State,
+			Fileset: e.Fileset,
+			Module: e.Module,
+			ReadTime: e.ReadTime,
+			EventMetadata: e.EventMetadata,
+		},
+	}
+}
+
+
 // Metadata creates a common.MapStr containing the metadata to
 // be associated with the event.
-func (e *Event) Metadata() common.MapStr {
-	if e.Pipeline != "" {
+func (eh *EventHolder) GetMetadata() common.MapStr {
+	if eh.Metadata.Pipeline != "" {
 		return common.MapStr{
-			"pipeline": e.Pipeline,
+			"pipeline": eh.Metadata.Pipeline,
 		}
 	}
 	return nil
@@ -85,8 +112,8 @@ func (e *Event) Metadata() common.MapStr {
 
 // HasData returns true if the event itself contains data
 // Events without data are only state updates
-func (e *Event) HasData() bool {
-	return e.Bytes > 0
+func (eh *EventHolder) HasData() bool {
+	return eh.Metadata.Bytes > 0
 }
 
 // mergeJSONFields writes the JSON fields in the event map,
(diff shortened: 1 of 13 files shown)


filebeat/f.yml
```diff
@@ -0,0 +1,143 @@
+###################### Filebeat Configuration Example #########################
```
ruflin: that was probably commited by accident. You can use filebeat.dev.yml and it will be ignored automatically by gitignore.

filebeat/prospector/config.go
```diff
@@ -43,6 +44,7 @@ type prospectorConfig struct {
 	Pipeline             string             `config:"pipeline"`
 	Module               string             `config:"_module_name"`  // hidden option to set the module name
 	Fileset              string             `config:"_fileset_name"` // hidden option to set the fileset name
+	Filters              processors.PluginConfig `config:"filters"`
```
ruflin: Probably we should call it processors, as all of them will be available. It's already causing confusion on the metricbeat side ...

filebeat/prospector/prospector.go
```diff
@@ -214,12 +223,20 @@ func (p *Prospector) updateState(event *input.Event) error {
...
 
-	ok := p.outlet.OnEvent(event)
-	if !ok {
-		logp.Info("Prospector outlet closed")
-		return errors.New("prospector outlet closed")
+	eventMap := event.ToMapStr()
+
+	//run the filters before sending to
+	eventMap = p.filters.Run(eventMap)
+	if eventMap != nil {
```
ruflin: Assuming the event was "removed" by the filter, this would probably mean no state update is sent anymore? Or when exactly is `eventMap` empty?

filebeat/f.yml
vjsamuel: removed it. thanks for the heads up. i ll use that.

filebeat/prospector/prospector.go
vjsamuel: eventMap would be empty in case of `drop_event`. But i did see that the file was traversed till the end:

```
2017/03/28 08:17:17.442705 log.go:114: INFO Reader was closed: /var/log/alf.log. Closing.
2017/03/28 08:17:17.443151 log.go:306: DBG  Stopping harvester for file: /var/log/alf.log
2017/03/28 08:17:17.443163 log.go:315: DBG  Closing file: /var/log/alf.log
2017/03/28 08:17:17.443169 log.go:183: DBG  Update state: /var/log/alf.log, offset: 0
2017/03/28 08:17:17.443130 log.go:183: DBG  Update state: /var/log/displaypolicyd.stdout.log, offset: 1155
2017/03/28 08:17:17.443030 log.go:315: DBG  Closing file: /var/log/jamf.log
2017/03/28 08:17:17.443198 log.go:183: DBG  Update state: /var/log/jamf.log, offset: 1190762
```

But since these events do goto registrar, the registry file wouldn't be updated. But this is a scenario where I configure my prospector with:
```
filters:
  drop_event:
```

In this case, no event would pass upto registrar.

filebeat/prospector/config.go
vjsamuel: ha ha! let me do that.

ruflin:
jenkins, test it

filebeat/input/event.go
```diff
@@ -11,24 +11,37 @@ import (
...
 	Module       string
+	InputType    string
+	DocumentType string
+	common.EventMetadata
+	ReadTime time.Time
+	Bytes    int
+	State    file.State
+}
+
+type EventHolder struct {
```
ruflin: Can we rename `EventHolder` to `Data` to be consistent with https://github.com/elastic/beats/blob/master/libbeat/outputs/outputs.go#L20 ? I'm ok with keeping Metadata as here it is at the moment only about Metadata

filebeat/input/event.go
```diff
@@ -11,24 +11,37 @@ import (
...
+
+}
+
+type EventMeta struct {
 	Pipeline     string
 	Fileset      string
 	Module       string
+	InputType    string
+	DocumentType string
+	common.EventMetadata
```
ruflin: lets move the common.EventMetadata to the the first line of EventMeta

filebeat/input/event.go
```diff
@@ -11,24 +11,37 @@ import (
...
-	common.EventMetadata
-	ReadTime     time.Time
-	InputType    string
-	DocumentType string
-	Bytes        int
-	Text         *string
-	JSONConfig   *reader.JSONConfig
-	State        file.State
-	Data         common.MapStr // Use in readers to add data to the event
+	EventMeta
```
ruflin: So currently we have EventMeta twice, means two copies of the data? What would be the affect if we not have it in the Event? I think we will need in the future processors to have access to the meta data too, but do we need this in the first step?

filebeat/prospector/prospector.go
```diff
@@ -214,7 +223,22 @@ func (p *Prospector) updateState(event *input.Event) error {
...
+	eventHolder := event.GetEventHolder()
+	//run the filters before sending to
+	if event.Bytes > 0 {
+		eventHolder.Event = p.processors.Run(eventHolder.Event)
+	}
+
+	var ok bool
+	if eventHolder.Event != nil {
+		//processor might decide to drop the event
+		ok = p.outlet.OnEvent(&eventHolder)
```
ruflin: Lets move the `ok = p.outlet.OnEvent(&eventHolder)` after the if / else and only have `if eventHolder.Event == nil { Bytes = 0} `. This should be the same logic. Then we also don't need the definition of ok.

filebeat/input/event.go
vjsamuel: Event will be used in harvesters and when it reaches the processor, we do a event.GetEventHolder(), which would (you are right), copy the metadata from Event to EventHolder and generate the MapStr. But after that since Event object goes out of scope, it would be garbage collected if I am not wrong. Would that be an issue? Today when we generate the common.MapStr, we copy all the log related fields over. Now with this change, we additionally copy over the meta fields as well. Please correct me if I'm wrong.

vjsamuel:
@ruflin, i have removed the processor code and renamed the object from `EventHolder` to `Data`.

Also do I submit the PR for processor code immediately or do I wait for this to be approved and merged?

filebeat/input/event.go
vjsamuel: done.

filebeat/prospector/prospector.go
vjsamuel: will add this back in the next PR. removed it for now.

ruflin:
jenkins, test it

vjsamuel:
thank you! :)

PR #3825, opened 2017-03-28
Fix packaging which broke because of asciidocs comments

ruflin:
The version.asciidoc is used during packaging to read out the version number. Recently a comment was introduce in the file which broke the packaging. In before_build.sh the content of the file is copied over and converted to yaml, see https://github.com/elastic/beats/blob/master/dev-tools/packer/xgo-scripts/before_build.sh#L58. As these lines to not seem to be required, I removed them. Alternatively the sed script could be adapted which would be more complex.

### DIFF
--- a/libbeat/docs/version.asciidoc
+++ b/libbeat/docs/version.asciidoc
@@ -1,8 +1,4 @@
 :stack-version: 6.0.0-alpha1
 :doc-branch: master
 :go-version: 1.7.4
-
-//////////
-release-state can be: released | prerelease | unreleased
-//////////
 :release-state: unreleased


ruflin:
jenkins, package it

ruflin:
Packaging looks good: http://build-eu-00.elastic.co/job/beats-package-PR/86/

tsg:
@dedemorton I'm merging this one to fix the packaging, We can discuss other options if you think it's important to keep that comment in.

dedemorton:
@tsg It's fine to remove the comments here. The main thing is having the release-state defined in a central place. :-)

PR #3827, opened 2017-03-28
Modify clean_removed handling

ruflin:
Previously if a file could be found under the same name as a state, the state was not removed. But this could have been also an other file. With this change also the file itself is compared and if it is not the same file, the state will be removed. This has the affect that in case a file is renamed after monitoring the file finished, the state will be removed. In most cases this should not have any side affect.

The positive effect of this change is that there will be less left over states in the registry.

Closes elastic#3789

### DIFF
--- a/filebeat/prospector/prospector_log.go
+++ b/filebeat/prospector/prospector_log.go
@@ -94,21 +94,32 @@ func (l *Log) Run() {
 	if l.config.CleanRemoved {
 		for _, state := range l.Prospector.states.GetStates() {
 			// os.Stat will return an error in case the file does not exist
-			_, err := os.Stat(state.Source)
+			stat, err := os.Stat(state.Source)
 			if err != nil {
-				// Only clean up files where state is Finished
-				if state.Finished {
-					state.TTL = 0
-					err := l.Prospector.updateState(input.NewEvent(state))
-					if err != nil {
-						logp.Err("File cleanup state update error: %s", err)
-					}
-					logp.Debug("prospector", "Remove state for file as file removed: %s", state.Source)
-				} else {
-					logp.Debug("prospector", "State for file not removed because not finished: %s", state.Source)
+				l.removeState(state)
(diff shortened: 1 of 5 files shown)


ruflin:
@elastic/beats Would be good to hear your thoughts on this change.

filebeat/prospector/prospector_log.go
```diff
@@ -94,24 +94,35 @@ func (l *Log) Run() {
 	if l.config.CleanRemoved {
 		for _, state := range l.Prospector.states.GetStates() {
 			// os.Stat will return an error in case the file does not exist
-			_, err := os.Stat(state.Source)
+			stat, err := os.Stat(state.Source)
 			if err != nil {
```
7AC: Should we be more strict here and check vs `os.PathError`?

filebeat/prospector/prospector_log.go
```diff
@@ -94,24 +94,35 @@ func (l *Log) Run() {
...
 			}
 		}
 	}
 }
 
+func (l *Log) removeState(state file.State) {
+	// Only clean up files where state is Finished
+	if state.Finished {
+		state.TTL = 0
+		err := l.Prospector.updateState(input.NewEvent(state))
```
7AC: We could return here and check the error in the caller, and also remove the `else`

filebeat/prospector/prospector_log.go
ruflin: Good idea. What should we do in all other cases? Log an error and not remove the state?

7AC: Yea that's probably the best we can do. `os.PathError` is all that can return right now but you never now..

filebeat/prospector/prospector_log.go
ruflin: I changed it to return directly if state is not finished but I kept the error handling local.

filebeat/prospector/prospector_log.go

PR #3831, opened 2017-03-28
libbeat/scripts: parameterize all usages of go get

7AC:


libbeat/scripts/Makefile
```diff
@@ -30,6 +30,8 @@ COVERAGE_DIR=${BUILD_DIR}/coverage
 COVERAGE_TOOL=${BEAT_GOPATH}/bin/gotestcover
 COVERAGE_TOOL_REPO=github.com/elastic/beats/vendor/github.com/pierrre/gotestcover
 TESTIFY_TOOL_REPO=github.com/elastic/beats/vendor/github.com/stretchr/testify
+GOX=${BEAT_GOPATH}/bin/gox
+GOX_REPO=github.com/mitchellh/gox
```
ruflin: Do we need this as variable? I recommend to only use variables if we have multiple usages as our Makefile is already quite messy / long.

7AC: In this cases it was mainly for consistency with the other `go get`s.

ruflin:
@7AC How should we move forward here? I prefer not to have variables when we don't need them but both works for me.

7AC:
I'll drop this one

PR #3839, opened 2017-03-29
[MetricBeat] Add possibility to add a minCpu and minMemory filter

ybonnel:
We had the need to monitor 5000 servers with a small cluster.
In order to limit the volume of data sent, we have set up a more dynamic filter on the process.
Process info is sent only if the cpu or rss exceeds the configured minimums.

### DIFF
--- a/metricbeat/metricbeat.full.yml
+++ b/metricbeat/metricbeat.full.yml
@@ -61,6 +61,12 @@ metricbeat.modules:
   period: 10s
   processes: ['.*']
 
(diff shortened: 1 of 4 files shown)


ruflin:
Couldn't you use the range processor for that in combination with drop_event? 
* https://www.elastic.co/guide/en/beats/metricbeat/current/configuration-processors.html#condition-range 
* https://www.elastic.co/guide/en/beats/metricbeat/current/drop-event.html

ybonnel:
Sure I could if I RTFM :)

Thx.

PR #3840, opened 2017-03-29
Add Beta and Experimental logging

ruflin:
These two logging methods can be used for logging beta and experimental features. This makes sure all of the features are logged the same way.

Reload Configuration was changed to beta from experimental.

### DIFF
--- a/libbeat/logp/log.go
+++ b/libbeat/logp/log.go
@@ -132,11 +132,21 @@ func Critical(format string, v ...interface{}) {
 	msg(LOG_CRIT, "CRIT ", format, v...)
 }
 
-// Deprecate logs a deprecation message
-// The version string contains the version when the future willb e removed
+// Deprecate logs a deprecation message.
+// The version string contains the version when the future will be removed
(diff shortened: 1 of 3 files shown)


libbeat/logp/log.go
```diff
@@ -136,7 +136,16 @@ func Critical(format string, v ...interface{}) {
 // The version string contains the version when the future willb e removed
 func Deprecate(version string, format string, v ...interface{}) {
 	postfix := fmt.Sprintf(" Will be removed in version: %s", version)
-	msg(LOG_WARNING, "WARN DEPRECATED: ", format+postfix, v...)
+	Warn("DEPRECATED: "+format+postfix, v...)
+}
+
+// Experimental is used to log experimetan
```
andrewkroh: Seems like this comment is incomplete. And it has a type. Suggest - "Experimental logs the usage of an experimental feature."

andrewkroh: Can you add a similar godoc for `Beta()`.

ruflin: Done

PR #3843, opened 2017-03-29
Replace glide vendoring with govendor

ruflin:
Glide is currently broken in the beats repo and adding dependencies is painful. I initially tried out `dep` (https://github.com/golang/dep) but it is not working with our nested vendor directories as it always tries to do auto discovery. It is not possible to add a single package at the moment. I would like to stick with local vendor directories as it allows to build modules outside of metricbeat for example and add a module by just copy/pasting the directory.

govendor allows to exactly do that but also allows to use autodiscovery if needed. One nice part about govendor is that it also strips testing files etc. from the repo and allows only import a package and not necessarily the full repository.

In a first step I'm using govendor for the mysql metricbeat module. In a second step, all dependencies have to be added with govendor.

The following command is used to add a dependency:
```
govendor fetch github.com/go-sql-driver/mysql@v1.3
```

Further changes:
* A `vendor.json` file was already in the top vendor directory. It seems it was commited in the past by accident. The file is removed.
* Mysql dependency is fixed to version 1.3. It was not fixed to a version before.
* glide.yaml in mysql directory removed


ruflin:
Closing in favor of https://github.com/elastic/beats/pull/3851

PR #3848, opened 2017-03-29
Enhancements of the cherrypick PRs script

monicasarbu:
This PR includes the following changes:
- add the original PR in the title
- delete the `needs_backport` label from the original PR once the backport PR was opened

### DIFF
--- a/dev-tools/cherrypick_pr
+++ b/dev-tools/cherrypick_pr
@@ -120,7 +120,7 @@ def main():
 
         # create PR
         request = session.post(base + "/pulls", json=dict(
-            title="Cherry-pick to {}: {}".format(args.to_branch, original_pr["title"]),
+            title="Cherry-pick of PR #{}: {}".format(args.pr_number, original_pr["title"]),
             head=remote_user + ":" + tmp_branch,
             base=args.to_branch,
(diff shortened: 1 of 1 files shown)


dev-tools/cherrypick_pr
```diff
@@ -135,6 +135,9 @@ def main():
         session.post(
             base + "/issues/{}/labels".format(new_pr["number"]), json=["backport", "review"])
 
+        # remove needs backport label from the original PR
+        session.delete(base + "/issues/{}/labels/needs_backport".format(args.pr_number))
```
ruflin: My only two concern with this is that the label should only be removed, when the PR is merged. And it also does not work if their are backports needed to 5.3 and 5.4

ruflin:
LGTM. See my comment above. I think it's more an edge case and more often it works then it doesn't.

dev-tools/cherrypick_pr
monicasarbu: I agree it's not ideal. In case we need to backport a PR to multiple branches, then it would be nice to enhance the script to be able to pass a list of branches instead of one, and then remove the `needs_backport`, but this is for another PR.

PR #3851, opened 2017-03-30
Moving from glide to govendor

vjsamuel:
Attempting to change go dependency management from glide to [govendor](https://github.com/kardianos/govendor).

This PR contains the following changes:

* Set a fixed version for mysql and postgresql modules in metricbeat
* Use govendor as dependency management instead of glide
* Support for vendor directory in modules by exluding modules from auto discovery
* Use auto discovery for global vendor directory
* Add dependencies for testing

This PR didn't touch the dependencies in the docker module as they have some internal changes.


vjsamuel:
cc: @ruflin

ruflin:
I updated the PR text to add more details on the changes. Some things we should do / check:

* Run `make clean-vendor` to remove some additional test files from sarama
* It removes dependencies for `github.com/andrewkroh/sys/plan9/env_plan9.go` etc. I assume we still need these, potentially for packaging? @andrewkroh If yes, should be added manually.
* Same as above for github.com/andrewkroh/sys/unix/* and github.com/andrewkroh/sys/windows/*
* Some dependencies were added we didn't have in before:
  * github.com/blakesmith/ar
  * github.com/valiercorder/go-rpm
  * github.com/go-ole/go-ole
  * github.com/tsg/gopacket
* Some dependencies changed. We should check for the same commit id
  * github.com/dustin/go-humanize
  * github.com/davecgh/go-spew
  * github.com/eapache/go-resiliency
  * ...
* Lots of files like `.golden` were removed.
* Lots of files were removed as we only need certain packages and not the full library which govendor makes possible.

ruflin:
jenkins, test it

ruflin:
jenkins, package it

ruflin:
Packaging PR: http://build-eu-00.elastic.co/view/Beats/job/beats-package-PR/87/console

ruflin:
Just found out where the "new packages" are coming from. `dev-tools` also has its own vendor directory. We should add this also to the "exclude" list and clean it up seperately.

ruflin:
I found all of the above missing / added files and all looks good. From my perspective we need to have the following changes before merging this PR:

* Add `dev-tools/vendor` to the ignore list and update the files again to not load these packages
* Run `make clean-vendor` to remove some more files
* Run `make update`

I'm ok with leaving in the the updated dependencies for some of the packages as we are in master and we need to update these anyways.

@elastic/beats WDYT?

andrewkroh:
> - It removes dependencies for github.com/andrewkroh/sys/plan9/env_plan9.go etc. I assume we still need these, potentially for packaging? @andrewkroh If yes, should be added manually.
> - Same as above for github.com/andrewkroh/sys/unix/* and github.com/andrewkroh/sys/windows/*

Of the `github.com/andrewkroh/sys` package, only `github.com/andrewkroh/sys/windows/svc/eventlog` is modified from the upstream. And it contains a small change to help with writing test cases for Winlogbeat. It's only used in test code.

vjsamuel:
@andrewkroh, if tests pass then does it mean that the change is good? I had a passing build until the previous commit. I've pushed a fix again. Build should pass now.

ruflin:
@andrewkroh @vjsamuel I merged it. The great thing about govendor is that it only include `ithub.com/andrewkroh/sys/windows/svc/eventlog` directory from the complete fork which is great. Lets do some follow up PR's in case there are some more corrections needed.

vjsamuel:
👍

PR #3853, opened 2017-03-30
Prospector level processors

vjsamuel:
Adding support to have prospector level filters. Kindly review the base code. If all is well, I can go ahead and add the system test cases and change logs.

### DIFF
--- a/filebeat/tests/system/test_registrar.py
+++ b/filebeat/tests/system/test_registrar.py
@@ -1312,3 +1312,79 @@ class Test(BaseTest):
 
         data = self.get_registry()
         assert len(data) == 0
+
+    def test_registrar_files_with_prospector_level_processors(self):
+        """
+        Check that multiple files are put into registrar file with drop event processor
+        """
+
+        self.render_config_template(
+            path=os.path.abspath(self.working_dir) + "/log/*",
+            prospector_processors=[{
+                "drop_event": {},
+            }]
+        )
+        os.mkdir(self.working_dir + "/log/")
+
+        testfile_path1 = self.working_dir + "/log/test1.log"
+        testfile_path2 = self.working_dir + "/log/test2.log"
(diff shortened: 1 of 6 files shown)


ruflin:
LGTM. Ok for me to add the system tests. @urso ok?

vjsamuel:
@ruflin I have added the system tests. I'll let the CI to complete. Local tests indicate that thing seem fine.

```
[success] 0.21% test_prospector.Test.test_prospector_filter_includefields: 0.2457s
[success] 0.21% test_prospector.Test.test_prospector_filter_dropfields: 0.2479s

Ran 116 tests in 118.419s

OK (SKIP=4)
```

urso:
I'm a little suspicious about processor dropping complete events? e.g. What happens to the registry if all events are dropped? I think we need some more tests here.

ruflin:
Agree with steffen, we should be exta careful here not to break things. Lets have a system test for each one. We can inspect the registry files for that in the tests.

filebeat/tests/system/test_registrar.py
```diff
@@ -1312,3 +1312,79 @@ def test_ignore_older_state_clean_inactive(self):
...
+            "offset": iterations * (len("hello world") + len(os.linesep)),
+        }, record)
+        self.assertTrue("FileStateOS" in record)
+        file_state_os = record["FileStateOS"]
+
+        if os.name == "nt":
+            # Windows checks
+            # TODO: Check for IdxHi, IdxLo, Vol in FileStateOS on Windows.
+            self.assertEqual(len(file_state_os), 3)
+        elif platform.system() == "SunOS":
```
ruflin: Did you run the tests on such a platform? I would assume this could also break some other tests ...

vjsamuel: i copy pasted this from another test case as i assumed that it was there for a reason.

Issue #3833, opened 2017-03-28
Filebeat Apache module grok pattern should handle absent byte count


cwurm:
In the Apache Combined Log Format, an absent byte count is signified by `-`. This has to be included in the grok pattern.

Should be something like: `"%{IPORHOST:clientip} %{USER:ident} %{USER:auth} \\[%{HTTPDATE:timestamp}\\] \"%{WORD:verb} %{DATA:request} HTTP/%{NUMBER:httpversion}\" %{NUMBER:response:int} (?:-|%{NUMBER:bytes:int}) %{QS:referrer} %{QS:agent}"`

Currently, the option `-` is missing from the default pattern in https://github.com/elastic/beats/blob/master/filebeat/module/apache2/access/ingest/default.json

cwurm:
Another thing: The grok pattern for access logs has spaces after opening parenthesis and not before at the end: `%{NUMBER:apache2.access.body_sent.bytes}( \"%{DATA:apache2.access.referrer}\")?( \"%{DATA:apache2.access.agent}\")?"`

Should be `%{NUMBER:apache2.access.body_sent.bytes} (\"%{DATA:apache2.access.referrer}\")? (\"%{DATA:apache2.access.agent}\")?"`

cwurm:
Hang on, that the whitespace didn't work was my mistake when I edited the grok pattern to fix the bytes. Apologies.

ruflin:
@cwurm Could you open a PR with this changes? Then we directly see how it affects our tests.

cwurm:
@ruflin done: https://github.com/elastic/beats/pull/3863

ruflin:
@cwurm Thanks a lot for the fix. Closing this one.

PR #3863, opened 2017-03-30
Allow `-` in Apache access log byte count

cwurm:
Allows a `-` instead of a number for the byte sent part of the Apache access log, as described in http://httpd.apache.org/docs/current/mod/mod_log_config.html

In contrast, Nginx will always output `0` - so no need to change anything there.

There are a few more pieces in the Grok pattern that are not quite in line with what Apache can produce. The Elasticsearch grok processor inherited [a list of patterns](https://github.com/elastic/elasticsearch/blob/master/modules/ingest-common/src/main/resources/patterns/grok-patterns#L95) from Logstash. One of them being `COMBINEDAPACHELOG`, we might want to mirror that in Filebeat. I guess that might be a more intrusive change, while this PR as-is can do no harm and should be fully backward compatible.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -49,6 +49,7 @@ https://github.com/elastic/beats/compare/v5.1.1...master[Check the HEAD diff]
 - Raise an exception in case there is a syntax error in one of the configuration files available under
   filebeat.config_dir. {pull}3573[3573]
 - Fix empty registry file on machine crash. {issue}3537[3537]
+- Allow `-` in Apache access log byte count. {pull}3863[3863]
 
 *Heartbeat*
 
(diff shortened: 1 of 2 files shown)


cwurm:
This addresses https://github.com/elastic/beats/issues/3833.

exekias:
jenkins, retest it please

exekias:
Could you please add  CHANGELOG entry?

cwurm:
@exekias I guess, though what should it look like? Which version(s)? Most recent release is 5.3.0, and the CHANGELOG here on master doesn't even cover that.

exekias:
Just add it in HEAD section, we will manage that in the backport

cwurm:
@exekias done

tsg:
Thanks for the fix @cwurm. I didn't use the `COMBINEDAPACHELOG` because I want control over the names of the resulting fields. Perhaps another option would have been to do renames afterwards, not sure.

cwurm:
@tsg We could take the grok pattern behind `COMBINEDAPACHELOG` and replace the variable names. Except for some for which we don't have a field Beats - but that's probably fine, as almost nobody uses `ident` anyway for example.

PR #3862, opened 2017-03-30
Update Metricbeat perfmon go:generate commands

andrewkroh:
There were a few issues with the “go generate” commands used in the perfmon metricset.

- Removed a space from `// go:generate`. There cannot be spaces.
- Fixed variable spelling in defs_windows.go. The spelling error was already fixed by hand in the generated files.
- Update the input file name in one command from pdh.go to pdg_windows.go.
- Added a wrapper script to ease running commands on Windows, setting GOARCH env var, and redirecting the output to a file.

Issue #3828, opened 2017-03-28
Enhance windows/perfmon metricset


ruflin:
This is a meta issue to track enhancements to the windows/perfmon metricset created in https://github.com/elastic/beats/pull/3758.

* [x] Cleanup fields.yml to only contain the minimal fields #3896
* [x] Refactoring config handling to use one list instead of groups https://github.com/elastic/beats/pull/3854
* [x] Improve error handling https://github.com/elastic/beats/pull/3871
* [x] Improve first polling cycle to not depend on sleep: https://github.com/elastic/beats/pull/3758/files#diff-90bec44359d2b80d45705c5b909c3643R74 https://github.com/elastic/beats/pull/3871
* [x] Add two `// go:generate <cmd>` lines to doc.go for generating `defs_windows_<goarch>.go` from `defs_windows.go`. https://github.com/elastic/beats/pull/3856 https://github.com/elastic/beats/pull/3862
* [x] Update `data.json` with a real example generated from a `_test.go` file. #4287
* [ ] Add many config examples to docs so people know how to collect common counters.
* [ ] Create example dashboards
* [x] Use wildcard queries to collect data from multiple instances from a counter
* [ ] Add config option to ignore failures for non-existent counters [[discuss]](https://discuss.elastic.co/t/windows-module-the-specified-object-was-not-found/86983)

maddin2016:
Hi @ruflin, how do you want to go on with this issue. Should we create a PR for every point or just one big PR?

ruflin:
@maddin2016 I'm a fan of lots of small PR's as it makes it easier to review and get the changes in quickly but one large would also work.

maddin2016:
I would like to start with `Improve first polling cycle`. Would that be an alternative? 
```
if firstFetch {
		// Most counters require two sample values in order to compute a displayable value. So wait and then collect the second value
		select {
		case <-time.After(time.Second * 2):
			err = _PdhCollectQueryData(q.query)
		}
	}
```

ruflin:
@maddin2016 The above will still block for 2 seconds. I was more thinking of finding an option to only wait until the all the data is available or even skip it in the first run. Not sure what the best behaviour is. That is probably one of the trickiest once in the list above.

I would suggest to first focus on the cleanup part, for example change to the config options.

maddin2016:
Ok. Here is an description of the problem
[...]

PR #3871, opened 2017-03-31
Refactor Windows perf data helper code

andrewkroh:
Refactoring of the Windows performance data code. There's more cleanup to come. See #3828.

- Improve generated PDH code to return usable error codes.
- Use the standard error interface in return values.
- Use more restrictive param types in generated functions.
- Create PdhErrno that interprets the PDH error codes and returns their human readable error string.

This PR is based on #3862, so it should be merged first to avoid duplicate code review.


metricbeat/module/windows/perfmon/pdh_windows.go
```diff
@@ -1,133 +1,237 @@
...
+	// Write the values into the map.
+	result := common.MapStr{}
+	var errs multierror.Errors
+
+	for counterPath, value := range values {
+		if value.Err != nil {
+			// XXX: Sometimes counters return negative values. We can ignore
+			// this error. For an explanation see:
+			// https://www.netiq.com/support/kb/doc.php?id=7010545
+			if err == PDH_CALC_NEGATIVE_DENOMINATOR {
```
maddin2016: This has to be `if value.Err == PDH_CALC_NEGATIVE_DENOMINATOR {`

metricbeat/module/windows/perfmon/pdh_windows.go
```diff
@@ -1,133 +1,237 @@
...
 
-func CloseQuery(q uintptr) PdhError {
-	err := _PdhCloseQuery(q)
-	if err != ERROR_SUCCESS {
-		return PdhError(err)
+func (r *PerfmonReader) Read() (common.MapStr, error) {
+	if !r.executed {
+		// Most counters require two sample values in order to compute a
+		// displayable value. So initial perform two samples.
+		r.query.Execute()
```
maddin2016: Hi @andrewkroh. You wrote 
>but the first value will always be 0

This is because the time slot is to small between these two calls. I would prefer the following solution
```golang
if err := r.query.Execute(); err != nil {
		return nil, err
	}

	// Get the values.
	values, err := r.query.Values()
	if err != nil {
		return nil, errors.Wrap(err, "failed formatting counter values")
	}

	// Write the values into the map.
	result := common.MapStr{}
	var errs multierror.Errors

	for counterPath, value := range values {
		if value.Err != nil {
			// XXX: Sometimes counters return negative values. We can ignore
			// this error. For an explanation see:
			// https://www.netiq.com/support/kb/doc.php?id=7010545
			if value.Err == PDH_CALC_NEGATIVE_DENOMINATOR {
				result.Put(counterPath, 0)
				continue
			}
			if value.Err == PDH_INVALID_DATA {
				if !r.executed {
					// Most counters require two sample values in order to compute a
					// displayable value. So the first fetch for this counters return PDH_INVALID_DATA.
					// So check if this is the first fetch. If so just move on and dont't put that 0 value
					// into our result. If not this is maybe a real error
					r.executed = true
					continue
				}
			}
			errs = append(errs, value.Err)
			continue
```

maddin2016:
Hi @andrewkroh. These changes are really great!! I have left 2 comments about the first query. One thing i'm very interested in is how you get all these PDH error codes with that `gcc` command.

maddin2016:
Ok, got it. These are `define` preprocessors coming from `pdhmsg.h`. And that `-D` command make them visible.

metricbeat/module/windows/perfmon/pdh_windows.go
andrewkroh: Good catch. It would be nice to have test for this error in the future.

metricbeat/module/windows/perfmon/pdh_windows.go
andrewkroh: I changed up the error handling a bit to sort of match ^. If an error occurs it will report the metric as zero. But it uses some conditions to determine whether or not the `error` is reported in the event. There are two special cases, one for `PDH_CALC_NEGATIVE_DENOMINATOR` and one for `PDH_INVALID_DATA` on the first fetch.

andrewkroh: Feel free to experiment and change it up. I primarily wanted to remove the `sleep`.

maddin2016:
Hi @andrewkroh. For error `PDH_CALC_NEGATIVE_DENOMINATOR` i currently write a test. But i think we should first merge your changes to not mix up things.

Issue #3807, opened 2017-03-24
Improve error handling in schema / metricsets


ruflin:
The current error handling in schema and metricsets is not optimal. The following should be done to improve it / clean it up:

* [ ] `schema.Apply` should return and error list and not only log errors. This improves unit testing. https://github.com/elastic/beats/pull/3790
* [ ] Remove temoprary `ApplyNoError` and modify processors accordingly
* [ ] New `schema.Apply` also stores Optional fields errors. Identify if this could be a performance issue in production
* [ ] Improve logging of errors by logging a missing field only once and potentially all in one line
* [ ] Check currently ignored errors in each metricset and handle errors

tsg:
Closing as old.

PR #3872, opened 2017-03-31
Remove ApplyNoError from Schema

andrewkroh:
The ApplyNoError function is not needed because users can just call Apply and ignore the error.

See #3807

### DIFF
--- a/metricbeat/schema/schema.go
+++ b/metricbeat/schema/schema.go
@@ -76,12 +76,6 @@ func (s Schema) ApplyTo(event common.MapStr, data map[string]interface{}) (commo
 	return event, errors
 }
 
-// Apply converts the fields extracted from data, using the schema, into a new map.
-func (s Schema) ApplyNoError(data map[string]interface{}) common.MapStr {
-	event, _ := s.ApplyTo(common.MapStr{}, data)
-	return event
-}
-
 // Apply converts the fields extracted from data, using the schema, into a new map and reports back the errors.
 func (s Schema) Apply(data map[string]interface{}) (common.MapStr, *Errors) {
 	return s.ApplyTo(common.MapStr{}, data)
--- a/libbeat/processors/add_cloud_metadata/add_cloud_metadata.go
+++ b/libbeat/processors/add_cloud_metadata/add_cloud_metadata.go
@@ -36,17 +36,23 @@ const (
 var debugf = logp.MakeDebug("filters")
 
 var (
-	ec2Schema = s.Schema{
-		"instance_id":       c.Str("instanceId"),
-		"machine_type":      c.Str("instanceType"),
-		"region":            c.Str("region"),
-		"availability_zone": c.Str("availabilityZone"),
-	}.ApplyNoError
+	ec2Schema = func(m map[string]interface{}) common.MapStr {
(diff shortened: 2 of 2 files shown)


andrewkroh:
jenkins, test it

metricbeat/schema/schema.go
```diff
@@ -71,19 +71,13 @@ func (o Object) HasKey(key string) bool {
 
 // ApplyTo adds the fields extracted from data, converted using the schema, to the
 // event map.
-func (s Schema) ApplyTo(event common.MapStr, data map[string]interface{}) (common.MapStr, *Errors) {
```
ruflin: Why did you change *Errors to error? The idea is that after `Apply` the `Errors` is available and checks can be made if required or invalid fields were missing to create a more detailed error report.

andrewkroh: I would strongly prefer to use `error` in public interfaces. Using an specialized error type forces a strong coupling with the caller and makes the API more brittle. Callers that care about the precise error details can use a type switch or a type assertion to look for specific errors and extract details. Or for really common checks, the package can expose helper fucntions to inspect the error (e.g. `os.IsNotExists`, `os.IsPermission`).

This how most of the stdlib works and the methods that have specialized error types document what they return for callers that need the details (e.g. `os.Open` says "If there is an error, it will be of type *PathError."). 

This is a great [blog post](https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully) about the different error techniques.

ruflin: Type switches are mostly going to solve the problem or helper functions. One problem of the implementation could be that it is called error. Perhaps it should be renamed. In case there are some "optional" errors, error is not nil but should be to support the default error handling `if err != nil`. The information returned is more some meta information about the event. 

I understand the reasoning behind the public interface should return error but one advantage of having `*Errors` returned is that it forces the person implementing the code, to deal with it and understand the logic of it. The person implementing it should make a decision on how to deal with the different types of "Errors". As I can see us adding more logic to it in the future we could rename it to "Meta" or something like that and this object has a method `GetError`. So it would not be directly and error. Just some alternative ideas on how we could deal with it.

Thanks for the blog post. One interesting method I was not aware of is `errors.Wrap` that we should probably use more often in our code.

andrewkroh: > advantage of having *Errors returned is that it forces the person implementing the code, to deal with it and understand the logic of it

I agree that the caller needs to deal with the error, but I don't think we should force them to understand the error. Returning a non-nil `error` should indicate failure only (so let's separate errors from warnings). If the caller wishes to inspect the error then there is a way to look at the individual errors caused by absent required fields.

Having a meta object to indicate the missing optional fields sounds like a good option to me. Basically we would separate errors from warnings. And warnings could be a separate return argument and distinct type. Outside of tests, I would expect most callers to be ignoring this meta return value with the list of absent optional fields (maybe with the exception of logging?).

If you agree, let's put this in the queue for #3807 and work it in a separate PR.

ruflin: Agree on the above. But that requires changes to the code as right now also missing optional fields return an error (and not nil)

PR #3874, opened 2017-03-31
Update docker module to govendor and most recent version

ruflin:
Our docker version depends now on our own fork to not break cross compiling. The fork can be found here: https://github.com/ruflin/docker/tree/17.03.1-beats-branch And here is the commit with the changes: ruflin/docker@f02993b

The changes are quite minimal which should make it rather easy to keep it up-to-date. Also this removes a list of dependency we don't need.

In the long term this should be replaced through our own client not depending on docker and the external client library.

The commands used to create this were:

```
govendor fetch +out
govendor fetch github.com/docker/docker/pkg/idtools::github.com/ruflin/docker/pkg/idtools@17.03.1-beats-branch +out
govendor fetch github.com/docker/docker/pkg/jsonmessage::github.com/ruflin/docker/pkg/jsonmessage@17.03.1-beats-branch +out
```

My intention was to make all the paths dependent on the fork with the command:

```
govendor fetch github.com/docker/docker::github.com/ruflin/docker/...@17.03.1-beats-branch +out
```

But this didn't work as expected.


metricbeat/module/docker/vendor/vendor.json
```diff
@@ -0,0 +1,274 @@
...
+		},
+		{
+			"checksumSHA1": "sjdNnoPZipse7blbn/DzfxVKpe4=",
+			"path": "github.com/docker/docker/pkg/system",
+			"revision": "f02993b2050240bbe7f2bff51f9be5b91e57b978",
+			"revisionTime": "2017-03-31T09:16:51Z",
+			"version": "17.03.1-beats-branch",
+			"versionExact": "v1.13.1"
+		},
+		{
```
vjsamuel: why are there two `github.com/ruflin/docker` in this vendor.json? The first one seemed right, where we are trying to use a fork of docker. This one doesnt seem to be right

ruflin:
@vjsamuel I played around with this some more. Unfortunately the only working option I found is that I update the two package paths I want to update by defining them specifically. This is not very nice as not all parts are not dependent on the fork :-(

PR #3875, opened 2017-03-31
Fix doc build for conf-file-permissions

ruflin:


ruflin:
@dedemorton FYI in case this applies also to other branches.

PR #3888, opened 2017-04-03
Kubernetes annotator

vjsamuel:
This PR adds support for Kubernetes as a processor for Beats. The Kubernetes processor allows Beats to enrich events with metadata coming from the Kubernetes Pod from which the event originated. Metadata might include:
* pod name
* pod namespace
* container name

The Kubernetes processor relies on two constructs:
* Indexers - used to generate metadata from Pods and store them with unique keys. Example of keys might be container IDs, IP:Port combinations, PodName.
* Matchers - used to generate a lookup key from an event. Example of a matcher might be a `log_path` matcher that looks at the `source` field of an event, extract the container ID from the log path and use it as a lookup key to retrieve metadata about the Pod from which the log message originated.


libbeat/processors/annotate/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,177 @@
...
+		endpoint := value.(string)
+		logp.Debug("kubernetes", "using endpoint: ", endpoint)
+
+		if endpoint != "" {
+			metadata = k.podWatcher.GetMetaData(endpoint)
+		} else {
+			return event, nil
+		}
+
+	} else if value, ok := event["pod"]; ok {
```
exekias: I guess this comes from a custom beat or module?

libbeat/processors/annotate/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,177 @@
...
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	"os"
+	"strings"
+	"time"
+)
+
+const (
+	timeout = time.Second * 5
```
ruflin: This should probably be configurable.

libbeat/processors/annotate/kubernetes/podwatcher.go
```diff
@@ -0,0 +1,298 @@
+package kubernetes
+
+import (
+	"github.com/elastic/beats/libbeat/logp"
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/runtime"
```
exekias: I think a custom client is not an option here, as this one is more advanced than a stateless REST consumer. Check SharedIndexInformer and Watch mechanism, the client code creates a copy of the state by watching k8s API

libbeat/processors/annotate/kubernetes/kubernetes.go
vjsamuel: we leave this check to make sure that any beat that writes an event with a key `pod`, we can still annotate it with the metadata.

libbeat/processors/annotate/kubernetes/podwatcher.go
```diff
@@ -0,0 +1,298 @@
...
+		pod, ok := podObj.(*corev1.Pod)
+		if ok {
+			logp.Info("Processing %s:%s", pod.Namespace, pod.Name)
+		}
+
+		name, endpoints, containers := p.getMetaKeys(pod)
+
+		//Pod is in terminating state, remove the metadata
+		if pod.DeletionTimestamp != nil {
+			p.removePodMapEntry(name, endpoints, containers)
```
vjsamuel: typically, log flow/metric scraping stops after the pod is marked for terminated. we could also move this to another method so that we delete it after the container is completely destroyed.

libbeat/processors/annotate/kubernetes/podwatcher.go
```diff
@@ -0,0 +1,298 @@
...
+		} else {
+			commonMeta := common.MapStr{
+				"pod":       name,
+				"namespace": pod.Namespace,
+				"labels":    pod.Labels,
```
vjsamuel: will do.

libbeat/processors/annotate/kubernetes/kubernetes.go
vjsamuel: yes. agreed.

libbeat/processors/annotate/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,177 @@
...
+
+	}
+
+	return nil, fatalError
+}
+
+func (k kubernetesAnnotator) Run(event common.MapStr) (common.MapStr, error) {
+
+	metadata := common.MapStr{}
+	if value, ok := event["source"]; ok {
```
vjsamuel: i like the idea of format strings. didnt think about it.

libbeat/processors/annotate/kubernetes/podwatcher.go
vjsamuel: @ruflin , im in agreement with @exekias, the kube client is a quite advanced client implementation. its not worth the effort to write our own client. we can have these vendored local to the kubernetes plugin.

libbeat/processors/annotate/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,177 @@
...
+		}
+
+	} else if value, ok := event["pod"]; ok {
+
+		logp.Debug("kubernetes", "using host pod name: ", value.(string))
+		metadata = k.podWatcher.GetMetaData(value.(string))
+
+		delete(event, "pod")
+	} else {
+		return event, nil
```
vjsamuel: i could have a filebeat configuration, that tails `/var/log/*` which is not managed by kubernetes and `/var/lib/docker/containers/*/*log` which is managed by kubernetes. the processor can be global and it would not pass through this check when a message from `/var/log` come in.

ruflin: That speaks for having local processors. 

If you don't want to see the error with the global config, the processor checks could be used to exclude some files from processing.

libbeat/processors/annotate/kubernetes/kubernetes.go
vjsamuel: my use case is that I have a source file that has a container id and i need to extract it out. This is not supported by format strings today.

vjsamuel:
One issue that I am facing is that `logp` and `glog` both use `v` as the verbosity flag. `logp` uses it at `-v` and `glog` uses it as `--v=<verbosity_level (1-10)>`. this causes:

```
panic: ./filebeat flag redefined: v

goroutine 1 [running]:
panic(0x5295c20, 0xc4202ddad0)
	/usr/local/Cellar/go/1.7.4_2/libexec/src/runtime/panic.go:500 +0x1a1
flag.(*FlagSet).Var(0xc42009e000, 0x5f10ea0, 0x5f7f3e8, 0x552b627, 0x1, 0x55465c0, 0x14)
	/usr/local/Cellar/go/1.7.4_2/libexec/src/flag/flag.go:791 +0x43e
flag.Var(0x5f10ea0, 0x5f7f3e8, 0x552b627, 0x1, 0x55465c0, 0x14)
	/usr/local/Cellar/go/1.7.4_2/libexec/src/flag/flag.go:806 +0x69
github.com/golang/glog.init.1()
	/Users/vija/workspace/go/src/github.com/golang/glog/glog.go:401 +0xe5
github.com/golang/glog.init()
	/Users/vija/workspace/go/src/github.com/golang/glog/glog_file.go:125 +0x1c1
k8s.io/apimachinery/pkg/labels.init()
	/Users/vija/workspace/go/src/k8s.io/apimachinery/pkg/labels/selector.go:837 +0x5d
k8s.io/apimachinery/pkg/apis/meta/v1.init()
```

ruflin:
I remember we had in the past our own FlagSet to prevent overlapping but I think we removed it as it was no issue anymore and made things simpler for having flags in different packages. Perhaps time to reintroduce it? I assume there isn't a trick in golang to disable some of the flags from library on compile time?

exekias:
for the record, after changing to a different k8s client the flags issue is gone. We changed to a more lightweight client to avoid increasing filebeat binary size

NOTICE
```diff
@@ -1050,6 +1050,202 @@ THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
...
+OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+--------------------------------------------------------------------
+golang.org/x/text
```
ruflin: We have this already in the global vendor directory. Not sure if we can tell govendor somehow about this that it only adds the packages which are not there. Or add the missing packages there? @vjsamuel any idea?

vjsamuel: @ruflin, if there is a version mismatch between `$BEATS_ROOT/vendor` and `$MODULE/vendor` then there would be two copies. if we can upgrade the root vendor to the more recent version this would ideally go away

exekias: Ah, that's cool, I personally miss some more verbosity from govendor

ruflin: @vjsamuel @exekias Could we "sync" the two to have the same commit? You can use `@hash` in govendor to pick the exact same package. For now I would just pick the same as in the top directory as we just updated to master very recently.

libbeat/processors/annotate/kubernetes/indexing.go
```diff
@@ -0,0 +1,256 @@
...
+
+// Implement default indexer:
+
+// GenMetadata generates default metadata for the given pod
+func GenMetadata(pod *corev1.Pod) common.MapStr {
+	labelMap := common.MapStr{}
+	for k, v := range pod.Metadata.Labels {
+		labelMap[k] = v
+	}
+	return common.MapStr{
```
ruflin: Under which namespace will this end up?

libbeat/processors/annotate/kubernetes/indexing.go
```diff
@@ -0,0 +1,256 @@
...
+}
+
+//TODO: Move this into filebeat
+type LogPathMatcher struct {
+	LogsPath string
+}
+
+func newLogsPathMatcher(cfg common.Config) Matcher {
+	config := struct {
+		LogsPath string `config:"logs_path"`
```
ruflin: Assuming this is in filebeat, that would mean the filebeat processors have an additional config option?

libbeat/processors/annotate/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,163 @@
...
+func init() {
+	processors.RegisterPlugin("kubernetes", newKubernetesAnnotator)
+
+	// Register default indexers
+	Indexing.AddIndexer(&PodNameIndexer{})
+	Indexing.AddIndexer(&ContainerIndexer{})
+}
+
+func newKubernetesAnnotator(cfg common.Config) (processors.Processor, error) {
+	config := kubeAnnotatorConfig{
```
ruflin: we normally have the defaults as a var external to the method. Something called `defaultKubeAnnotatorConfig` in a `config.go` file. Check some other `config.go` files for an example. This makes it easy to find the config options.

libbeat/processors/annotate/kubernetes/podwatcher.go
```diff
@@ -0,0 +1,184 @@
...
+
+	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/logp"
+
+	"github.com/ericchiang/k8s"
+	corev1 "github.com/ericchiang/k8s/api/v1"
+)
+
+const (
+	podCheckPeriod = 20 * time.Second
```
ruflin: Couldn't find where this is used?

libbeat/processors/annotate/kubernetes/podwatcher.go
```diff
@@ -0,0 +1,184 @@
...
+func (p *PodWatcher) updatePod(pod *corev1.Pod) {
+	oldPod := p.GetPod(pod.Metadata.GetUid())
+	if oldPod.Metadata.GetResourceVersion() != pod.Metadata.GetResourceVersion() {
+		//Process the new pod changes
+		p.deletePod(oldPod)
+		p.addPod(pod)
+	}
+}
+
+func (p *PodWatcher) deletePod(pod *corev1.Pod) {
```
ruflin: I would call this `remove` as the Pod will still exist (hopefully) and it's only removed from the list.

libbeat/processors/annotate/kubernetes/indexing.go
exekias: We still have the namespacing discussion open, I think we should put everything under the same `kubernetes` namespace but I was waiting to see the rest of modules to know if it makes sense

libbeat/processors/annotate/kubernetes/podwatcher.go
vjsamuel: its unused. i will remove this one.

libbeat/processors/annotate/kubernetes/podwatcher.go
vjsamuel: this is the naming convention that kube controllers have for various events that are processed which are:
```
onAdd
onUpdate
onDelete
```

we could rename it to `onPodAdd`, `onPodUpdate` and `onPodDelete`. does that work @ruflin ?

ruflin: I like the `onPodDelete` but as it is a private method naming is not too important. I just stumbled over it when looking at the code.

NOTICE
vjsamuel: i have moved the dependency to root vendor so now they share the same commit.

vjsamuel:
@ruflin, i have moved the config to a dedicated `config.go` file so that the global configs are easier to keep track of.

vjsamuel:
@ruflin, basic test cases and processor documentation have been added as well.

libbeat/processors/annotate/kubernetes/indexing.go
```diff
@@ -0,0 +1,262 @@
...
+		matchers: make(map[string]MatcherConstructor, 0),
+
+		defaultIndexers: make([]Indexer, 0),
+		defaultMatchers: make([]Matcher, 0),
+	}
+}
+
+// AddIndexer to the register
+func (r *Register) AddIndexer(name string, indexer IndexConstructor) {
+	r.indexers[name] = indexer
```
exekias: I'm missing all RWMutex locking logic in getters & setters, could you please add it? Like in https://github.com/elastic/beats/pull/3888/files#diff-2afe85a18185c27752efda431319dac3R120

filebeat/processors/annotate/kubernetes/indexing.go
```diff
@@ -0,0 +1,75 @@
...
+	if err == nil {
+		kubernetes.Indexing.AddDefaultIndexer(container)
+	} else {
+		logp.Err("Unable to load indexer plugin due to error: %v", err)
+	}
+
+	//Add a log path matcher which can extract container ID from the "source" field.
+	logsPathMatcher, err := newLogsPathMatcher(*cfg)
+	if err == nil {
+		kubernetes.Indexing.AddDefaultMatcher(logsPathMatcher)
```
exekias: Why not just adding the registered matcher/indexer name here, instead of registering an instance? like `AddDefaultIndexer("logs_path")`

This would force us to register it before marking it as default

libbeat/processors/annotate/kubernetes/config.go
```diff
@@ -0,0 +1,30 @@
...
+	EnableDefaultIndexers bool          `config:"enable_default_indexers"`
+}
+
+type PluginConfig []map[string]common.Config
+
+func defaultKuberentesAnnotatorConfig() kubeAnnotatorConfig {
+	return kubeAnnotatorConfig{
+		InCluster:             true,
+		SyncPeriod:            1 * time.Second,
+		Namespace:             "kube-system",
```
vjsamuel: i wouldnt recommend that. default namespace is only for control plane components. `kube-system` or `monitoring` are recommended namespaces.

libbeat/docs/processors-config.asciidoc
```diff
@@ -453,3 +454,67 @@ section.
...
+ The configuration below enables the processor when the beat as a pod in kubernetes.
+
+[source,yaml]
+-------------------------------------------------------------------------------
+processors:
+- kuberenes:
+    in_cluster: true
+-------------------------------------------------------------------------------
+
+The configuration below enables the processor on a beat running as a process on the kubernetes node
```
vjsamuel: outside the node is useless for logs as the filebeat needs to sit inside the same node.

exekias: I was thinking also in the metricbeat case, but works for me, it's a minor thing anyway

filebeat/beater/filebeat.go
```diff
@@ -17,6 +17,9 @@ import (
 	"github.com/elastic/beats/filebeat/publisher"
 	"github.com/elastic/beats/filebeat/registrar"
 	"github.com/elastic/beats/filebeat/spooler"
+
+	//Add filebeat level processors
+	_ "github.com/elastic/beats/filebeat/processors/annotate/kubernetes"
```
ruflin: I like that approach

filebeat/processors/annotate/kubernetes/indexing.go
```diff
@@ -0,0 +1,91 @@
+package kubernetes
```
ruflin: We have a little bit a mess with package names singular/plural. We started all new packages to be singular (like module) when it contains multiple packages with each package one module. `processors` initially contained all processors in one package I think, that is why it was called `processors` and not `processor`. It should be also changed in libbeat later but in filebeat we should directly start with `processor`.

libbeat/docs/processors-config.asciidoc
```diff
@@ -453,3 +454,67 @@ section.
...
+
+Matchers are used to build lookup keys with which indices can be queried. For example, the `fields`
+matcher can take a list of fields based on whose values, indices can be queried to retrieve metadata.
+
+Each beat can define its own default indexers and matchers which are enabled by default. For example,
+filebeat enables the `container` indexer which indexes pod metadata based on all container IDs and a
+ `logs_path` matcher that takes the `source` field, extracts the container ID and uses it to retrieve
+ metadata.
+
+ The configuration below enables the processor when the beat as a pod in kubernetes.
```
ruflin: "when the beat as a pod" ? Is a word missing here?

libbeat/docs/processors-config.asciidoc
```diff
@@ -453,3 +454,67 @@ section.
...
+The configuration below allows custom indexers and matchers to be enabled.
+
+[source,yaml]
+-------------------------------------------------------------------------------
+processors:
+- kuberenes:
+    in_cluster: false
+    host: <hostname>
+    kube_config: ~/.kube/config
+    enable_default_indexers: false
```
ruflin: I would turn this around to `default_indexers.enabled` to be consistend with `enabled` in other places. With `ucfg` the `enabled` comes for "free" as it is already part of the config option.

Same below for `default_matches.enabled`.

libbeat/docs/processors-config.asciidoc
```diff
@@ -453,3 +454,67 @@ section.
...
+    enable_default_matchers: false
+    indexers:
+      - ip_port:
+    matchers:
+      - fields:
+          lookup_fields: ["metricset.host"]
+-------------------------------------------------------------------------------
```
ruflin: I miss a note on how where the enriched data for the event will be put. How will the end event look like?

metricbeat/beater/metricbeat.go
```diff
@@ -12,6 +12,9 @@ import (
 
 	"github.com/elastic/beats/libbeat/cfgfile"
 	"github.com/pkg/errors"
+
+	//Add metricbeat specific processors
+	_ "github.com/elastic/beats/metricbeat/processors/annotate/kubernetes"
```
ruflin: We should auto generate this in the future for all beats based on the convention for the namespace and make it part of the `include` package.

metricbeat/processors/annotate/kubernetes/indexing.go
```diff
@@ -0,0 +1,126 @@
+package kubernetes
```
ruflin: For the package name see comment in filebeat.

NOTICE
ruflin: @vjsamuel That means now all dependencies are on the top level? Not sure I like that the kubernetes dependencies are on the top level as they are module specific. Any chance to have a mix? Means the common ones on the top level and all the other dependencies in here?

libbeat/docs/processors-config.asciidoc
```diff
@@ -453,3 +454,67 @@ section.
 NOTE: If you define an empty list of fields under `include_fields`, then only
 the required fields, `@timestamp` and `type`, are exported.
 
+[[kubernetes]]
+=== kubernetes
+
+The `kubernetes` processor annotates each event with relevant metadata based on
+which kubernetes pod the event originated from. Each event is annotated with:
```
dedemorton: Capitalize Kubernetes whenever it's used as a product name.

libbeat/docs/processors-config.asciidoc
```diff
@@ -453,3 +454,67 @@ section.
...
+The `kuberentes` process has two basic building blocks which are:
+
+* Indexers
+* Matchers
+
+Indexers take in a pod's metadata and builds indices based on the pod metadata. For example,
+the `ip_port` indexer can take a kubernetes pod and index the pod metadata based on all
+`pod_ip:container_port` combinations.
+
+Matchers are used to build lookup keys with which indices can be queried. For example, the `fields`
```
dedemorton: A little hard to follow because of the passive voice. Maybe say, "...build lookup keys for querying indices."

libbeat/docs/processors-config.asciidoc
```diff
@@ -453,3 +454,67 @@ section.
...
+matcher can take a list of fields based on whose values, indices can be queried to retrieve metadata.
```
dedemorton: I can't follow your meaning here, so you'll need to reword. Might also help to have a specific example. In fact, the concepts here are a bit hard to follow (not sure if it's a problem with the doc or my lack of knowledge regarding Kubernetes). Specific examples somewhere showing how the events are enriched with Kubernetes data would be very helpful.

libbeat/docs/processors-config.asciidoc
```diff
@@ -453,3 +454,67 @@ section.
...
+
+Each beat can define its own default indexers and matchers which are enabled by default. For example,
```
dedemorton: Capitalize Beat and Filebeat.

libbeat/docs/processors-config.asciidoc
dedemorton: A few minor changes to capitalization, word choice, and commas: "For example, Filebeat defines the `container` indexer, which indexes pod metadata based on all container IDs, and a `logs_path` matcher, which takes the `source` field, extracts the container ID, and uses it to retrieve metadata.

libbeat/docs/processors-config.asciidoc
```diff
@@ -453,3 +454,67 @@ section.
...
+filebeat enables the `container` indexer which indexes pod metadata based on all container IDs and a
+ `logs_path` matcher that takes the `source` field, extracts the container ID and uses it to retrieve
+ metadata.
+
+ The configuration below enables the processor when the beat as a pod in kubernetes.
+
+[source,yaml]
+-------------------------------------------------------------------------------
+processors:
+- kuberenes:
```
dedemorton: Should this say `kubernetes`? This question applies to other places where you use kuberenes instead of kubernetes.

libbeat/docs/processors-config.asciidoc
```diff
@@ -453,3 +454,67 @@ section.
...
+=== kubernetes
+
+The `kubernetes` processor annotates each event with relevant metadata based on
+which kubernetes pod the event originated from. Each event is annotated with:
+
+* Pod Name
+* Namespace
+* Labels
+
+The `kuberentes` process has two basic building blocks which are:
```
dedemorton: Should this say `kubernetes` processor instead of process? Note that there is also a typo: `kuberentes`

libbeat/docs/processors-config.asciidoc
dedemorton: How do users know which indexers and matchers are enabled for each Beat?

libbeat/docs/processors-config.asciidoc
dedemorton: Again, capitalize Beat and Kubernetes.

libbeat/docs/processors-config.asciidoc
```diff
@@ -453,3 +454,67 @@ section.
...
+[source,yaml]
+-------------------------------------------------------------------------------
+processors:
+- kuberenes:
+    in_cluster: false
+    host: <hostname>
+    kube_config: ~/.kube/config
+-------------------------------------------------------------------------------
+
+The configuration below allows custom indexers and matchers to be enabled.
```
dedemorton: Does this configuration enable the indexers and matchers, or does it allow you to enable them somewhere else? The passive voice used here makes it sound like the latter.

libbeat/docs/processors-config.asciidoc
vjsamuel: we would need to document individual beats on what indexers/matchers each of them enable by default. having them in libbeat documentation wouldnt be ideal.

metricbeat/processors/annotate/kubernetes/indexing.go
vjsamuel: done.

libbeat/docs/processors-config.asciidoc
vjsamuel: done

libbeat/docs/processors-config.asciidoc
vjsamuel: done.

filebeat/processors/annotate/kubernetes/indexing.go
vjsamuel: done

libbeat/docs/processors-config.asciidoc
vjsamuel: done

libbeat/processors/annotate/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,203 @@
...
+	logp.Debug("kubernetes", "Using host ", config.Host)
+	logp.Debug("kubernetes", "Initializing watcher")
+	if client != nil {
+		watcher := NewPodWatcher(client, &indexers, config.SyncPeriod, config.Host)
+
+		if watcher.Run() {
+			return kubernetesAnnotator{podWatcher: watcher, matchers: &matchers}, nil
+		}
+
+		return nil, fatalError
```
ruflin: this line could be removed because there is the same return in the end.

metricbeat/processor/annotate/kubernetes/indexing.go
```diff
@@ -0,0 +1,126 @@
...
+		if err == nil {
+			kubernetes.Indexing.AddDefaultIndexer(ipPort)
+		} else {
+			logp.Err("Unable to load indexer plugin due to error: %v", err)
+		}
+	} else {
+		logp.Err("Unable to get indexer plugin %s", IpPortIndexerName)
+	}
+
+	matcher := kubernetes.Indexing.GetMatcher(kubernetes.FieldMatcherName)
```
ruflin: Two init functions would simplify the logic here (see comment in filebeat)

PR #3892, opened 2017-04-03
set defaultDateDetection to true

metabr:
Elasticsearch documentation states that date_detection dynamic mapping is enabled by default. However, generated filebeat mapping have this disabled by default.

I assume that there is possibly a reason to not have that enabled by default and documentation is inconsistent, so this PR should be declined and documentation should be updated instead.

Otherwise, it would be nice to not have sane defaults changed by filebeat mapping.

### DIFF
--- a/libbeat/template/template.go
+++ b/libbeat/template/template.go
@@ -9,7 +9,7 @@ import (
 
 var (
 	// Defaults used in the template
-	defaultDateDetection    = false
+	defaultDateDetection    = true
 	defaultTotalFieldsLimit = 10000
 
 	// Array to store dynamicTemplate parts in


andrewkroh:
See https://github.com/elastic/beats/issues/3389 which issue explains why date detection is disabled. Could you help point me to the beats documentation that need updated?

metabr:
Not beats documentation, but Elasticsearch reference section on
[dynamic mappings](https://www.elastic.co/guide/en/elasticsearch/reference/current/dynamic-field-mapping.html#date-detection).

I believe it is worthwhile to mention possible issues with date_detection there.

PR #3895, opened 2017-04-03
Add experimental Kibana module to Metricbeat

ruflin:
The Kibana module is marked experimental as it might change or removed at any time.


metricbeat/module/kibana/status/data.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+var (
+	schema = s.Schema{
+		"uuid": c.Str("uuid"),
+		"name": c.Str("name"),
+		"version": c.Dict("version", s.Schema{
+			"number": c.Str("number"),
+		}),
+		"status": c.Dict("status", s.Schema{
```
ruflin: @tsg I wanted to do here `state.overall = status.overal.state` but couldn't find a way in schema. Any hints?

joegallo:
I did a quick compare and contrast between the output of `api/status` for a recent Kibana (it happens to be 5.2.2) and for an older Kibana (4.6.4). Generally the shape of things is pretty similar, but Kibana 4.6.4 reports "id" for statuses, not "name" and it lacks many of the top level elements (uuid, version, name, buildNum, etc).

I read the code briefly, but not being an expert I'll just ask: does this support 4.6.4? 

Unrelated to the above, it wasn't clear to me whether you support custom headers like [filebeat does](https://github.com/elastic/beats/blob/master/filebeat/filebeat.full.yml#L465-L467). That's a really important thing for me.

joegallo:
It looked like I needed to sign in to Kibana in order to hit `api/status` manually (otherwise I just got redirected to the login page) -- is there a configuration option for username/password for this? (there's a reference in the documentation to `configuration-metricbeat` and I've tried to follow it through but I'm a bit lost in the configuration documentation, so again, just asking.)

PR #3908, opened 2017-04-04
Add new MetricSet interfaces for Module Developers

andrewkroh:
This PR adds new interfaces for MetricSet developers. These interfaces provided additional flexibility for MetricSet implementations. The new interfaces are:

- `Closer` - With Metricbeat gaining the ability to dynamically load and unload modules there became a need to properly release resources opened by a MetricSet. If a MetricSet implements the `Closer` interface then the `Close() error` method will be invoked when the MetricSet is unloaded.

- `ReportingFetcher` - Some MetricSets needed to report multiple errors, but the existing `Fetch` methods only allowed a single error to be returned. `ReportingFetcher` passes a callback interface to the MetricSet that allows it to report any combination of events, errors, or errors with metadata.

- `PushMetricSet` - This is for push event sources. All the current MetricSet implementations are driven by periodic `Fetch` callbacks triggered by timer. A `PushMetricSet` does not receive periodic callbacks, but instead has a `Run` method. Implementations can forward events as they are received from the underlying source.

### DIFF
--- a/metricbeat/mb/example_metricset_test.go
+++ b/metricbeat/mb/example_metricset_test.go
@@ -8,7 +8,9 @@ import (
 	"github.com/elastic/beats/metricbeat/mb/parse"
 )
 
-var hostParser = parse.URLHostParserBuilder{DefaultScheme: "http"}.Build()
+var hostParser = parse.URLHostParserBuilder{
+	DefaultScheme: "http",
+}.Build()
 
 func init() {
 	// Register the MetricSetFactory function for the "status" MetricSet.
@@ -26,14 +28,23 @@ func NewMetricSet(base mb.BaseMetricSet) (mb.MetricSet, error) {
 	return &MetricSet{BaseMetricSet: base}, nil
 }
 
-func (ms *MetricSet) Fetch() (common.MapStr, error) {
-	// Fetch data from the host (using ms.HostData().URI) and return the data.
-	return common.MapStr{
-		"someParam":  "value",
-		"otherParam": 42,
-	}, nil
+// Fetch will be called periodically by the framework.
+func (ms *MetricSet) Fetch(report mb.Reporter) {
+	// Fetch data from the host at ms.HostData().URI and return the data.
+	data, err := common.MapStr{
+		"some_metric":          18.0,
+		"answer_to_everything": 42,
+	}, error(nil)
+	if err != nil {
+		// Report an error if it occurs.
+		report.Error(err)
+		return
+	}
+
+	// Otherwise report the collected data.
+	report.Event(data)
 }
 
-// ExampleMetricSetFactory demonstrates how to register a MetricSetFactory
-// and unpack additional configuration data.
-func ExampleMetricSetFactory() {}
+// ExampleReportingMetricSet demonstrates how to register a MetricSetFactory
+// and implement a ReportingMetricSet.
+func ExampleReportingMetricSet() {}
--- a/metricbeat/mb/mb.go
+++ b/metricbeat/mb/mb.go
@@ -71,6 +71,12 @@ type MetricSet interface {
 	HostData() HostData // HostData returns the parsed host data.
 }
 
+// Closer is an optional interface that a MetricSet can implement in order to
+// cleanup any resources it has open at shutdown.
+type Closer interface {
+	Close() error
+}
+
 // EventFetcher is a MetricSet that returns a single event when collecting data.
 type EventFetcher interface {
 	MetricSet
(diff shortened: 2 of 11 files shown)


metricbeat/mb/mb.go
```diff
@@ -84,6 +90,42 @@ type EventsFetcher interface {
 	Fetch() ([]common.MapStr, error)
 }
 
+// Reporter is used by a MetricSet to report events, errors, or errors with
+// metadata. The methods return false iff publishing failed because the
+// MetricSet is being closed.
+type Reporter interface {
+	Event(event common.MapStr) bool
+	ErrorWith(err error, meta common.MapStr) bool
+	Error(error) bool
```
ruflin: Do we need both interfaces or could we just have `Error(err error, meta common.MapStr)` and in case the meta does not exist it is set to nil? Thinking if we could reduce the number of interface methods.

metricbeat/mb/mb.go
```diff
@@ -84,6 +90,42 @@ type EventsFetcher interface {
 	Fetch() ([]common.MapStr, error)
 }
 
+// Reporter is used by a MetricSet to report events, errors, or errors with
+// metadata. The methods return false iff publishing failed because the
```
andrewkroh: [iff](https://en.wikipedia.org/wiki/If_and_only_if) is shorthand for "if and only if". I thought it was more common, maybe only common in math. I'll expand it to "if and only if".

metricbeat/mb/mb.go
andrewkroh: I think having separate methods for each of the cases (will) improve the readability of MetricSets that implement `ReportingFetcher`.

Maybe it's not clear that MetricSets don't implement this interface, they only use it. So reducing the number of methods doesn't change the amount of methods a MetricSet needs to implement (if that was your concern).

metricbeat/mb/module/wrapper.go
```diff
@@ -182,128 +184,155 @@ func (mw *Wrapper) Hash() uint64 {
...
-				break
-			}
-		}
+		msw.multiEventFetch(fetcher, reporter)
+	case mb.ReportingFetcher:
+		msw.reportingFetch(fetcher, reporter)
 	default:
-		return fmt.Errorf("MetricSet '%s/%s' does not implement a Fetcher "+
-			"interface", msw.Module().Name(), msw.Name())
+		panic(fmt.Sprintf("unexpected fetcher type for %v", msw))
```
andrewkroh: It's an unreachable code path and can never happen due to an earlier check in the caller. So if it does happen I want it to blow up loudly and early so we can catch it in development/unit tests.

andrewkroh:
> Not all usage of the interfaces is 100% clear to me yet, but I assume they will get answered as soon as we have the first implementation.

I would like to make it clear from the godocs. What would improve your understanding? More docs on the MetricSet types? An example for for each MetricSet type (`ReportingFetcher`, `PushMetricSet`) like we have for [`EventFetcher`](https://godoc.org/github.com/elastic/beats/metricbeat/mb#example-MetricSetFactory)?

metricbeat/mb/mb.go
ruflin: lol, makes sense now. Did not make the connection when reading it and just thought: typo :-)

metricbeat/mb/module/wrapper.go
```diff
@@ -182,128 +184,155 @@ func (mw *Wrapper) Hash() uint64 {
...
+
+func (r *eventReporter) Error(err error) bool {
+	return r.ErrorWith(err, nil)
+}
+
+func (r *eventReporter) ErrorWith(err error, meta common.MapStr) bool {
+	timestamp := r.start
+	elapsed := time.Duration(0)
+
+	if !timestamp.IsZero() {
```
ruflin: Minor comment / thought: In `Fetch() []common.MapStr` all events have the exact same time which elapsed. In this new interface the first event will have a lower `rtt` then the last event. Not sure yet if this is a good or bad thing.

metricbeat/mb/example_metricset_test.go
```diff
@@ -26,14 +28,23 @@ func NewMetricSet(base mb.BaseMetricSet) (mb.MetricSet, error) {
...
+func (ms *MetricSet) Fetch(report mb.Reporter) {
+	// Fetch data from the host at ms.HostData().URI and return the data.
+	data, err := common.MapStr{
+		"some_metric":          18.0,
+		"answer_to_everything": 42,
+	}, error(nil)
+	if err != nil {
+		// Report an error if it occurs.
+		report.Error(err)
+		return
```
ruflin: That pattern we will probably see quite often. If Error would not return anything, we could use `return report.Error(err)` probably.

metricbeat/mb/mb.go
```diff
@@ -71,19 +71,62 @@ type MetricSet interface {
...
 // EventFetcher is a MetricSet that returns a single event when collecting data.
+// Use ReportingMetricSet for new MetricSet implementations.
 type EventFetcher interface {
 	MetricSet
 	Fetch() (common.MapStr, error)
 }
 
 // EventsFetcher is a MetricSet that returns a multiple events when collecting
-// data.
+// data. Use ReportingMetricSet for new MetricSet implementations.
```
ruflin: In case we decided to keep both, we need to rename it later to something like `EventsMetricSet` to be consistent. I still like the old `Fetch` interfaces so not sure yet if Reporter will replace it or just be an other option.

PR #3916, opened 2017-04-05
Kubernetes module for metricbeat

vjsamuel:
First pass on integrating with kubelet.


metricbeat/include/list.go
```diff
@@ -28,6 +28,8 @@ import (
 	_ "github.com/elastic/beats/metricbeat/module/docker/info"
 	_ "github.com/elastic/beats/metricbeat/module/docker/memory"
 	_ "github.com/elastic/beats/metricbeat/module/docker/network"
+	_ "github.com/elastic/beats/metricbeat/module/dropwizard"
+	_ "github.com/elastic/beats/metricbeat/module/dropwizard/collector"
```
exekias: probably these are spurious :)

metricbeat/metricbeat.full.yml
```diff
@@ -179,6 +179,14 @@ metricbeat.modules:
   #username: ""
   #password: ""
 
+#------------------------------- kubelet Module ------------------------------
+- module: kubelet
+  metricsets: ["node","container","volume","pod","system"]
+  enabled: true
+  period: 10s
+  hosts: ["localhost"]
```
ruflin: What is the default port?

metricbeat/metricbeat.full.yml
```diff
@@ -179,6 +179,14 @@ metricbeat.modules:
   #username: ""
   #password: ""
 
+#------------------------------- kubelet Module ------------------------------
+- module: kubelet
```
ruflin: Should be commented out so it is not enabled by default.

metricbeat/metricbeat.yml
```diff
@@ -46,6 +46,14 @@ metricbeat.modules:
   period: 10s
   processes: ['.*']
 
+#------------------------------- kubelet Module ------------------------------
```
ruflin: You can use `short_config: false` in the module fields.yml so it is not added to the short config.

metricbeat/module/dropwizard/collector/collector.go
```diff
@@ -0,0 +1,96 @@
+package collector
```
ruflin: dropwizard? Is that here intentional? Sounds like an other potential module for the future :-)

metricbeat/module/kubelet/_meta/fields.yml
```diff
@@ -0,0 +1,9 @@
+- key: kubelet
+  title: "kubelet"
+  description: >
+    kubelet Module
```
ruflin: Can you add `[]beta` flag and `short_config: false` here? See https://github.com/elastic/beats/blob/master/metricbeat/module/docker/_meta/fields.yml#L7 for an example.

metricbeat/module/kubelet/container/data.go
```diff
@@ -0,0 +1,72 @@
...
+						"namespace": pod.PodRef.Namespace,
+						"node":      node.NodeName,
+					},
+				},
+
+				"name":       container.Name,
+				"start_time": container.StartTime,
+
+				"cpu": common.MapStr{
+					"usage.nanocores":       container.CPU.UsageNanoCores,
```
ruflin: That should be `"usage": common.MapStr{ "nanocores" }`. The reason is otherwise it will break with elasticsearch 2.x. We often use `common.MapStr.Put(...) in these cases as it does it automatically, but not sure how we could use this here ...

metricbeat/module/kubelet/container/data.go
```diff
@@ -0,0 +1,72 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("Cannot unmarshal json response: %s", err)
+	}
+
+	node := summary.Node
+	for _, pod := range summary.Pods {
+		for _, container := range pod.Containers {
+			containerEvent := common.MapStr{
+				mb.ModuleData: common.MapStr{
+					"pod": common.MapStr{
```
ruflin: Seems like the `schema` package could be used here to do the mapping. Can also be done later.

metricbeat/module/kubelet/node/data.go
```diff
@@ -0,0 +1,62 @@
...
+		"name":       node.NodeName,
+		"start_time": node.StartTime,
+
+		"cpu": common.MapStr{
+			"usage.nanocores":       node.CPU.UsageNanoCores,
+			"usage.corenanoseconds": node.CPU.UsageCoreNanoSeconds,
+		},
+
+		"memory": common.MapStr{
+			"available.bytes":  node.Memory.AvailableBytes,
```
ruflin: See comments above about schema and 2.x

metricbeat/module/kubelet/container/data.go
vjsamuel: schema wouldnt work because the input interface coming from the HTTP has arrays.

ruflin: I see. What I did in other metricsets is using multiple schemas. Means doing the array "manually" and then under it having schemas. But we can look into it after the first version.

vjsamuel: i see what your saying.

metricbeat/module/kubelet/container/data.go
vjsamuel: done

metricbeat/module/kubelet/node/data.go
vjsamuel: done

metricbeat/include/list.go
vjsamuel: removed

metricbeat/metricbeat.full.yml
vjsamuel: done

metricbeat/metricbeat.yml
vjsamuel: done

metricbeat/module/dropwizard/collector/collector.go
vjsamuel: removed. added it by mistake to this PR. it is the next one ;)

metricbeat/module/kubelet/_meta/fields.yml
vjsamuel: done.

metricbeat/module/kubelet/container/_meta/data.json
```diff
@@ -0,0 +1,78 @@
...
+    "beat": {
+        "hostname": "beathost",
+        "name": "beathost",
+        "version": "6.0.0-alpha1"
+    },
+    "kubelet": {
+        "container": {
+            "cpu": {
+                "usage": {
+                    "corenanoseconds": 3305756719,
```
ruflin: Is this nano seconds? If yes, we should use `.ns`: https://www.elastic.co/guide/en/beats/libbeat/5.1/event-conventions.html

metricbeat/module/kubelet/container/_meta/data.json
```diff
@@ -0,0 +1,78 @@
...
+                    "nanocores": 5992
```
ruflin: What is that exactly?

vjsamuel: @ruflin 
```
// CPUStats contains data about CPU usage.
type CPUStats struct {
	// The time at which these stats were updated.
	Time metav1.Time `json:"time"`
	// Total CPU usage (sum of all cores) averaged over the sample window.
	// The "core" unit can be interpreted as CPU core-nanoseconds per second.
	// +optional
	UsageNanoCores *uint64 `json:"usageNanoCores,omitempty"`
	// Cumulative CPU usage (sum of all cores) since object creation.
	// +optional
	UsageCoreNanoSeconds *uint64 `json:"usageCoreNanoSeconds,omitempty"`
}
```

andrewkroh:
jenkins, test it

PR #3919, opened 2017-04-05
Update latest in the testing env to 5.3

tsg:


tsg:
Added in progress because I want to add the geoip and user agent plugins to this one.

PR #3923, opened 2017-04-05
Ignore audit fileset on macOS

### LINKED ISSUE #3918: Auditd fileset breaks the system module on macOS
The auditd fileset has no paths defined for macOS. Unfortunately, because all filesets are enabled by default, it means that the system module doesn't work on macOS.

This is the relevant portion of the manifest file:

```
var:
  - name: paths
    default:
      - /var/log/audit/audit.log*
    os.darwin: [""]
    os.windows: []
```

An easy solution would be to replace `os.darwin: []` with `os.darwin: [""]`, which would mean that the fileset is silently ignored on macOS, but I see that and `auditd` daemon does seem to exist on macOS as well. 

@andrewkroh Would it make sense to support the macOS' `auditd`?

andrewkroh:
Silently ignore audit fileset on macOS.

Fixes #3918

### DIFF
--- a/filebeat/module/system/audit/manifest.yml
+++ b/filebeat/module/system/audit/manifest.yml
@@ -4,7 +4,7 @@ var:
   - name: paths
     default:
       - /var/log/audit/audit.log*
-    os.darwin: []
+    os.darwin: [""]
     os.windows: []
 
 ingest_pipeline: ingest/pipeline.json


filebeat/module/system/audit/manifest.yml
```diff
@@ -4,7 +4,7 @@ var:
   - name: paths
     default:
       - /var/log/audit/audit.log*
-    os.darwin: []
+    os.darwin: [""]
     os.windows: []
```
ruflin: Is the same change for winodws needed?

andrewkroh: The whole system metricset is documented as not supported on Windows.

tsg: Yeah, I think it's better to leave at is, so that users get an error on windows when trying this module (otherwise, the problem would be silently ignored).

Issue #3915, opened 2017-04-05
Heartbeat HTTP monitor should default to port 80


urso:
If port number is missing HTTP monitor loading fails with: `manager.go:125: ERR failed to load monitor tasks: missing port in address`. The monitor should default to port 80, if no port is configured.

andrewkroh:
And it should default to 443 for HTTPS URLs.

PR #3924, opened 2017-04-05
heartbeat: setup default ports in http monitors

7AC:
Default http to 80 and https to 443

Resolves https://github.com/elastic/beats/issues/3915


ruflin:
Could you update the changelog with an entry and potentially also the docs? Should we backport this?

7AC:
I updated the changelog, I'm not sure the doc deserves an entry for this. About backporting I'm open to suggestions..

tsg:
+1 to including this one in 5.4

PR #3925, opened 2017-04-05
libbeat/scripts: make testsuite target re-runnable

7AC:
Right now when you run `make testsuite` multiple times, it sometimes fails if `${BEAT_NAME}.template*.json` is not there. I added a `-f` to the `rm` command.


libbeat/scripts/Makefile
```diff
@@ -118,7 +118,7 @@ clean:: ## @build  Cleans up all files generated by the build steps
 	# Clean index pattern
 	rm -f $(PWD)/_meta/kibana/index-pattern/${BEAT_NAME}.json
 	# Remove temp templates
-	-rm ${BEAT_NAME}.template*.json
```
ruflin: This should not be needed anymore on master as no templates are generated anymore. All are loaded dynamically from fields.yml.

libbeat/scripts/Makefile
```diff
@@ -209,7 +209,8 @@ test: unit
 
 .PHONY: testsuite
 testsuite: ## @testing Runs all tests and generates the coverage reports
-testsuite: clean update
```
ruflin: what is the difference here?

ruflin:
It might be we try to fix a similar issue? https://github.com/elastic/beats/pull/3922

libbeat/scripts/Makefile
7AC: Ok I'll remove it completely

libbeat/scripts/Makefile
7AC: `clean` and `update` won't run in parallel, right now from what I can tell (on my laptop) they end up running in parallel and `update` tries to invoke `activate` from the `build` dir right after `clean` deleted all of it

libbeat/scripts/Makefile
ruflin: I should have been more accurate here. It is still possible to generate the template through `make index-template` so removing can be still needed. But it (should) have no effect on the build as it is not used.

7AC: 👍 I'll keep it

ruflin:
@7AC It would make sense to add somewhere a note that we don't support -j.

7AC:
@ruflin 👍 https://github.com/elastic/beats/pull/3937

PR #3927, opened 2017-04-05
Remove reference to old `import_dashboards`

7AC:
It was rewritten in Go.


tsg:
There's another related issue in this section: https://www.elastic.co/guide/en/beats/libbeat/master/import-dashboards.html#import-dashboards-for-development

While most of the code is still in `beats/libbeat/dashboards`, the command itself moved to `beats/dev-tools/cmd/import_dashboards` and that's also where the Makefile is.

We should also consider using "command" or "program" instead of "script".

7AC:
Superseded by https://github.com/elastic/beats/pull/3938

PR #3929, opened 2017-04-05
Don't stop Filebeat when modules + logstash are used together

tsg:
To simplify the migration from Ingest Node to Logstash, we replace the
hard error ("Filebeat modules require an Elasticsearch output defined") with
a warning.

### DIFF
--- a/filebeat/beater/filebeat.go
+++ b/filebeat/beater/filebeat.go
@@ -75,7 +75,11 @@ func New(b *beat.Beat, rawConfig *common.Config) (beat.Beater, error) {
 func (fb *Filebeat) modulesSetup(b *beat.Beat) error {
 	esConfig := b.Config.Output["elasticsearch"]
 	if esConfig == nil || !esConfig.Enabled() {
-		return fmt.Errorf("Filebeat modules configured but the Elasticsearch output is not configured/enabled")
+		logp.Warn("Filebeat is unable to load the Ingest Node pipelines for the configured" +
+			" modules because the Elasticsearch output is not configured/enabled. If you have" +
+			" already loaded the Ingest Node pipelines or are using Logstash pipelines, you" +
+			" can ignore this warning.")
+		return nil
 	}
 	esClient, err := elasticsearch.NewConnectedClient(esConfig)
 	if err != nil {


tsg:
@dedemorton This is the PR to make FBM continue if only the Logstash output is configured, with a warning. Can you review the warn message? I was thinking we could even link to a longer explanation in the docs.

filebeat/beater/filebeat.go
```diff
@@ -75,7 +75,10 @@ func New(b *beat.Beat, rawConfig *common.Config) (beat.Beater, error) {
 func (fb *Filebeat) modulesSetup(b *beat.Beat) error {
 	esConfig := b.Config.Output["elasticsearch"]
 	if esConfig == nil || !esConfig.Enabled() {
-		return fmt.Errorf("Filebeat modules configured but the Elasticsearch output is not configured/enabled")
+		logp.Warn("Filebeat modules enabled but the Elasticsearch output is not configured/enabled." +
+			" This means I can't load the Ingest Node pipelines." +
```
ruflin: I stumbled over the `I` and would replace it by `Filebeat`

ruflin:
jenkins, retest it

filebeat/beater/filebeat.go
```diff
@@ -75,7 +75,10 @@ func New(b *beat.Beat, rawConfig *common.Config) (beat.Beater, error) {
 func (fb *Filebeat) modulesSetup(b *beat.Beat) error {
 	esConfig := b.Config.Output["elasticsearch"]
 	if esConfig == nil || !esConfig.Enabled() {
-		return fmt.Errorf("Filebeat modules configured but the Elasticsearch output is not configured/enabled")
```
dedemorton: LGTM (couple of minor changes)

Filebeat modules are enabled, but the Elasticsearch output is not configured/enabled. This means Filebeat can't load the Ingest Node pipelines. If you have already loaded the pipelines, you can ignore this warning.

Not sure if this covers LS users, tho, because they really don't need to load the pipeline at all, right?

dedemorton: An alternate version (in case you want to shift the focus away from the idea that ES is required): 

Filebeat is unable to load the Ingest Node pipelines for the configured modules because the Elasticsearch output is not configured/enabled. If you have already loaded the Ingest Node pipelines or are using Logstash pipelines, you can ignore this warning.

tsg:
@dedemorton thanks, i used your second version.

PR #3938, opened 2017-04-06
Remove reference to old import_dashboards script

7AC:
Replace it with the new Go implementation in the doc and make it clear that the corresponding `make` target can only run from the root of a Beat.

### DIFF
--- a/libbeat/docs/newdashboards.asciidoc
+++ b/libbeat/docs/newdashboards.asciidoc
@@ -32,7 +32,7 @@ The following topics provide more detail about importing and working with Beats 
 
 You can use the `import_dashboards` script to import all the dashboards and the index pattern for a Beat, including the dependencies such as visualizations and searches.
 The `import_dashboards` script is available under
-https://github.com/elastic/beats/tree/master/libbeat/dashboards[beats/libbeat/dashboards]. It's also available in each Beat package under the `scripts` directory.
+https://github.com/elastic/beats/tree/master/dev-tools/cmd/import_dashboards[beats/dev-tools/cmd/import_dashboards]. It's also available in each Beat package under the `scripts` directory.
 
 There are a couple of common use cases for importing dashboards:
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -96,6 +96,11 @@ beats-dashboards:
 	mkdir -p build/dashboards
 	$(foreach var,$(BEATS),cp -r $(var)/_meta/kibana/ build/dashboards/$(var)  || exit 1;)
 
+# Collects all dashboards and generates dashboard folder for https://github.com/elastic/beats-dashboards/tree/master/dashboards
```
ruflin: As far as I know I think the collection is not for beats-dashboards any more, only for packaging.

Makefile
```diff
@@ -96,6 +96,11 @@ beats-dashboards:
...
+.PHONY: import-dashboards
+import-dashboards: beats-dashboards
```
ruflin: What is the use case of this? The main reason of import-dashboards is when working on a beat and quickly want to import the dashboards. When do you want to import the dashboards for all beats at the same time?

The other question would be why the `beats-dashboards` dependency is needed? If we need this I would expect it to depend on `update`

packetbeat/docs/fields.asciidoc
```diff
@@ -1115,6 +1115,16 @@ Unless this field is disabled, it always has a value, and it matches the `client
 
 
 [float]
+=== client_location
```
ruflin: I think this somehow sneaked in. Try to run `make clean` and then `make update`. I also had this once in the past but not sure what happened. Probably going back and forth from other branches and have some not checked in files which are outdated.

Makefile
7AC: This was copy paste by accident

packetbeat/docs/fields.asciidoc
7AC: Thanks I'll remove it

Makefile
7AC: I was going after what the doc said about `make import-dashboards`, maybe I should fix that instead then:

> If Elasticsearch is running on localhost, then you can run the following command from the Beat repository

7AC:
@ruflin done 👍

PR #3941, opened 2017-04-06
Rename Filebeat module from system.audit to auditd.log

andrewkroh:
This moves the `audit` fileset from the `system` module into its own module named `auditd`. The new fileset name is `log`.


andrewkroh:
The Jenkins failure doesn't look related to the changes.

tsg:
It was green before, so I'm merging it.

PR #3945, opened 2017-04-06
Adding support for custom http headers and TLS for metricbeat modules

vjsamuel:
This PR adds support to add TLS options when a user wants to hit HTTPS endpoints and also provide custom headers in case an endpoint has authentication enabled.

Example:

```
- module: prometheus
  namespace: "kube-apiserver"
  metricsets: ["collector"]
  enabled: true
  period: 10s
  headers:
    Authorization: "Bearer test123"
  hosts: ["https://apiserver"]
  ssl.verification_mode: "none"
```


metricbeat/helper/http.go
```diff
@@ -22,10 +25,42 @@ type HTTP struct {
 
 // NewHTTP creates new http helper
 func NewHTTP(base mb.BaseMetricSet) *HTTP {
+	config := struct {
+		TLS     *outputs.TLSConfig `config:"ssl"`
```
ruflin: @urso as we start to using this in other places, we should probably move it out from the output package to ?

ruflin:
jenkins, test it

ruflin:
@vjsamuel I create this follow up issue: https://github.com/elastic/beats/issues/3948

PR #3960, opened 2017-04-07
New location for the github token

tsg:
ES is using the `~/.elastic/github.token` location, so let's use the
same to avoid copying the token in too many places.

@elastic/beats FYI, after we merge this one, you will need to move the token in the new location before using the script next time.


tsg:
This is what you probably need:

```
mkdir -p ~/.elastic
mv ~/.github_token ~/.elastic/github.token
```

PR #3977, opened 2017-04-10
Run vet only against local packages

exekias:
Without this change we are running vet for all packages, including those under vendor folders. I got vet errors for an external dependency I was trying to include


exekias:
```
$ go list ./...  | grep -v /vendor/ > golist_dump
$ govendor list -no-status +local > govendor_dump
$ diff golist_dump govendor_dump
```

I tested both approaches, I prefer list + grep so I don't need to install govendor, but this is not a strong opinion

PR #3979, opened 2017-04-10
Add url format in the index pattern

monicasarbu:
Add the Url format in the `fields.yml`, in order to generate an index pattern with an Url field. The Url link is needed in order to link the `beat.name` to a host-specific dashboard.

Here is an example on how to make `beat.name` a link to a dashboard:
```
name: beat.name
   metricbeat.format: url
   metricbeat.params: 
       urlTemplate: "../app/kibana#/dashboard/79ffd6e0-faa0-11e6-947f-177f697178b8?_a=(query:    (query_string:(analyze_wildcard:!t,query:'beat.hostname:{{value}}'"
       labelTemplate: "{{value}}"
```
As a result, the host name appears as a link to the dashboard in all the tables.
screen shot 2017-04-10 at 7 35 00 pm

### DIFF
--- a/libbeat/_meta/fields.common.yml
+++ b/libbeat/_meta/fields.common.yml
@@ -11,15 +11,15 @@
         set in the configuration file, then that value is used. If it is not
         set, the hostname is used. To set the Beat name, use the `name`
         option in the configuration file.
-      format: url
-      params: 
-        urlTemplate: "../app/kibana#/dashboard/79ffd6e0-faa0-11e6-947f-177f697178b8?_a=(query:(query_string:(analyze_wildcard:!t,query:'beat.hostname:{{value}}')))"
-        labelTemplate: "{{value}}"
-
     - name: beat.hostname
       description: >
         The hostname as returned by the operating system on which the Beat is
         running.
+      metricbeat.format: url
+      metricbeat.params:
+        urlTemplate: "../app/kibana#/dashboard/79ffd6e0-faa0-11e6-947f-177f697178b8?_a=(query:(query_string:(analyze_wildcard:!t,query:'beat.hostname:{{value}}')))"
+        labelTemplate: "{{value}}"
+
     - name: beat.timezone
       description: >
         The timezone as returned by the operating system on which the Beat is
--- a/libbeat/_meta/fields.common.yml
+++ b/libbeat/_meta/fields.common.yml
@@ -11,6 +11,11 @@
         set in the configuration file, then that value is used. If it is not
         set, the hostname is used. To set the Beat name, use the `name`
         option in the configuration file.
+      format: url
+      params: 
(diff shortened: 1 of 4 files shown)


libbeat/_meta/fields.common.yml
```diff
@@ -11,6 +11,11 @@
         set in the configuration file, then that value is used. If it is not
         set, the hostname is used. To set the Beat name, use the `name`
         option in the configuration file.
+      format: url
+      params: 
+        urlTemplate: "../app/kibana#/dashboard/79ffd6e0-faa0-11e6-947f-177f697178b8?_a=(query:(query_string:(analyze_wildcard:!t,query:'beat.hostname:{{value}}')))"
```
andrewkroh: Is is possible to give the dashboard a name that we can reference here that is not a UUID? I think that would be a little better to maintainability.

andrewkroh: Would it be possible to only apply this to the `metricbeat-*` index pattern instead of all beats?

monicasarbu: That's a good point, Andrew. Let me see if I can do that with the current functionality.

libbeat/_meta/fields.common.yml
```diff
@@ -15,6 +15,11 @@
       description: >
         The hostname as returned by the operating system on which the Beat is
         running.
+      metricbeat.format: url
```
ruflin: If possible we should try not to mix beat specific things into libbeat.

monicasarbu: This is the only solution that I found to set a different format for `beat.hostname` for Metricbeat than for the other beats. Other option would be to remove it from common libbeat fields, but this is a bigger change.

ruflin: Ok, lets keep it then and figure out a way in the long run how we can make this nicer. Perhaps we can somehow have it also in metricbeat in the future and merge the fields values somehow. 

Do you think this will also happen to other fields then only `beat.hostname` or is this hostname specific?

libbeat/template/template.go
```diff
@@ -117,7 +117,9 @@ func (t *Template) generate(properties common.MapStr, dynamicTemplates []common.
 
 	if t.esVersion.IsMajor(2) || t.esVersion.IsMajor(5) {
 		basicStructure.Put("mappings._default_._all.norms.enabled", false)
-	} else {
+	}
+
+	if t.esVersion.major >= 5 {
```
ruflin: This change is also part of https://github.com/elastic/beats/pull/4079

monicasarbu:
After discussing internally, we decided to close this PR in favor of a custom widget where the user can configure the Dashboard where a field should point to.

Issue #2084, opened 2016-07-22
filebeat wildcard for directories


runningman84:
It looks like filebeat does not support wildcards in directories like this:

```
/opt/codedeploy-agent/deployment-root/*/*/logs/scripts.log
```

ruflin:
This is related to https://github.com/elastic/filebeat/issues/68 Wildcards for directories should already be supported. What is the behaviour you were expecting? Are you looking for `**` which crawls all subdirectories (and is not supported at the moment)?

runningman84:
My folders look like this
```
/opt/codedeploy-agent/deployment-root/4b74562c-cee3-40f5-8d36-6f588eeed802/d-KYVV5AYKF/logs/scripts.log
/opt/codedeploy-agent/deployment-root/b41f722c-80e8-41e9-866e-1f11228f5ab3/d-UIIXR9K9F/logs/scripts.log
```
This config does not seem to work:
```
'/opt/codedeploy-agent/deployment-root/*/*/logs/scripts.log'
```
Do I need to change the config to this?
```
'/opt/codedeploy-agent/deployment-root/**/logs/scripts.log'
```

ruflin:
I would expect it to work, but TBH so far I only tested `/opt/*/scripts.log` examples, means only one directory with a `*` pattern. Could you briefly tests if `/opt/codedeploy-agent/deployment-root/4b74562c-cee3-40f5-8d36-6f588eeed802/*/logs/scripts.log` works for you? Which filebeat version are you using?

runningman84:
yes 
/opt/codedeploy-agent/deployment-root/e8be8394-bb4f-403e-abd8-03045480217d/*/logs/scripts.log
works

ruflin:
Ok, so it seems to work with one directory but not 2 nested directories. We use directly `Glob` from Golang for the pattern: https://golang.org/pkg/path/filepath/#Glob Currently all pattern supported by `Glob` are supported by filebeat.

runningman84:
do you know a pattern which would work here?

andrewkroh:
@ruflin I'm thinking we should list the patterns supported (https://golang.org/pkg/path/filepath/#Match) in our documentation or link to it. Also we should put some info in there to clarify that only one wildcard is supported (assuming I understand correctly)?

If you agree I can open a new issue to add this to the docs.

ruflin:
[...]

PR #3980, opened 2017-04-10
filebeat: expand double wildcards in prospector

7AC:
Expand double wildcards into standard glob patterns, up to a maximum
depth of 16 levels after the wildcard.

Resolves https://github.com/elastic/beats/issues/2084


filebeat/prospector/prospector_log.go
```diff
@@ -112,15 +116,37 @@ func (l *Log) Run() {
 	}
 }
 
+// glob detects the use of "**" and expands it to standard glob patterns up to a max depth
+func glob(g string) ([]string, error) {
+	var globs []string
+	if filepath.Base(g) == "**" {
```
tsg: Does this mean it only detects `**` at the end of the path? Does `/foo/**/bar` work?

7AC: Yeah, I didn't realize that was also a use case. Let me fix it, thanks

tsg:
I'm 👍  with this approach, since I think it's the simplest implementation that gets us what most people want. I'm on the fence about making the 16 configurable or not, but I guess it would be good to be able to set it to 0 in which case the feature is disabled (for BWC).

filebeat/prospector/prospector_log.go
```diff
@@ -112,17 +114,69 @@ func (l *Log) Run() {
 	}
 }
 
+// globPattern detects the use of "**" and expands it to standard glob patterns up to a max depth
+func (l *Log) globPatterns(pattern string) ([]string, error) {
```
ruflin: Perhaps we could move this to a separate package and pass the max depth to the function. Reason is like this we could also use it in other places and it does not further "bloat" the prospector code.

filebeat/prospector/config.go
```diff
@@ -27,24 +27,25 @@ var (
...
+	CleanRemoved           bool                    `config:"clean_removed"`
+	HarvesterLimit         uint64                  `config:"harvester_limit" validate:"min=0"`
+	Symlinks               bool                    `config:"symlinks"`
+	TailFiles              bool                    `config:"tail_files"`
+	JSON                   *reader.JSONConfig      `config:"json"`
+	Pipeline               string                  `config:"pipeline"`
+	Module                 string                  `config:"_module_name"`  // hidden option to set the module name
+	Fileset                string                  `config:"_fileset_name"` // hidden option to set the fileset name
+	Processors             processors.PluginConfig `config:"processors"`
+	DoubleWildcardMaxDepth uint8                   `config:"double_wildcard_max_depth"`
```
ruflin: * Where is the default set for this one? 
* The config name seems to be a bit long

I would like to have this feature disabled by default and people who need it enable it specifically. This also makes it easier for us to debug it in the future in case of errors, because we know someone enabled it. 

I like the idea of defining the max depth as in case someone just needs one or two levels, this would have positive performance implications.

filebeat/prospector/prospector_log.go
```diff
@@ -112,17 +114,69 @@ func (l *Log) Run() {
...
+	if l.config.DoubleWildcardMaxDepth == 0 {
+		return []string{pattern}, nil
+	}
+	var patterns []string
+	isAbs := filepath.IsAbs(pattern)
+	patternList := strings.Split(pattern, "/")
+	for i, dir := range patternList {
+		if len(patterns) > 0 {
+			if dir == "**" {
```
ruflin: That means `test/**/hello/**/abc.log` is not going to work?

filebeat/prospector/prospector_log_test.go
```diff
@@ -62,3 +64,96 @@ func (t TestFileInfo) Mode() os.FileMode  { return 0 }
 func (t TestFileInfo) ModTime() time.Time { return t.time }
 func (t TestFileInfo) IsDir() bool        { return false }
 func (t TestFileInfo) Sys() interface{}   { return nil }
+
+func TestGlob(t *testing.T) {
+	var tests = []struct {
+		pattern         string
+		expectedMatches []string
+	}{
+		{
```
ruflin: Could you add tests which have a deeper depth like 4 or 5?

filebeat/prospector/prospector_log.go
```diff
@@ -112,17 +114,69 @@ func (l *Log) Run() {
...
 }
 
+// globPattern detects the use of "**" and expands it to standard glob patterns up to a max depth
+func (l *Log) globPatterns(pattern string) ([]string, error) {
+	if l.config.DoubleWildcardMaxDepth == 0 {
+		return []string{pattern}, nil
+	}
+	var patterns []string
+	isAbs := filepath.IsAbs(pattern)
+	patternList := strings.Split(pattern, "/")
```
urso: what about windows users having `\` in their path?

filebeat/prospector/prospector_log.go
```diff
@@ -112,17 +114,69 @@ func (l *Log) Run() {
...
+			}
+			wildcards := ""
+			for j := uint8(0); j < l.config.DoubleWildcardMaxDepth; j++ {
+				wildcards = filepath.Join(wildcards, "*")
+				patterns = append(patterns, filepath.Join(prefix, wildcards))
+			}
+		}
+	}
+	if len(patterns) == 0 {
+		patterns = []string{pattern}
```
urso: we have any kind of error if pattern is empty?

filebeat/prospector/prospector_log_test.go
```diff
@@ -62,3 +64,96 @@ func (t TestFileInfo) Mode() os.FileMode  { return 0 }
...
+			"foo/**/baz",
+			[]string{
+				"foo/bar/baz",
+			},
+		},
+		{
+			"foo/**/bazz",
+			[]string{},
+		},
```
urso: some more tests to add: 
- double `/` like `foo//bar`
- `foo\bar`
- `foo\**\bar`
- `c:\foo\**\bar`

filebeat/prospector/prospector_log.go
7AC: Yeah I like that

filebeat/prospector/prospector_log.go
7AC: Yes, we were discussing with @tsg and if we allow that the complexity will escalate quickly

filebeat/prospector/config.go
7AC: I just assumed the golang type default (0) is used as fallback? Totally agree on the config name: how about `double_star_pattern_depth`? Maybe not shorter but simpler..

filebeat/prospector/prospector_log_test.go
7AC: Sure

filebeat/prospector/prospector_log.go
7AC: Good catch, I should use `os.PathSeparator` instead

filebeat/prospector/prospector_log_test.go
7AC: 👍 👍

filebeat/prospector/prospector_log.go
7AC: Ok so right now we're not doing any validation on the pattern, with this change that wouldn't change wrt an empty pattern. Do we want to start validating those?

filebeat/input/file/glob.go
```diff
@@ -0,0 +1,65 @@
...
+		return []string{pattern}, nil
+	}
+	var patterns []string
+	isAbs := filepath.IsAbs(pattern)
+	patternList := strings.Split(pattern, string(os.PathSeparator))
+	for i, dir := range patternList {
+		if len(patterns) > 0 {
+			if dir == "**" {
+				err := fmt.Sprintf("glob(%s) failed: cannot specify multiple ** within a pattern", pattern)
+				logp.Err(err)
```
andrewkroh: I don't recommend logging errors and returning them. Do one or the other. It's probably best to let the caller make the decision on how to handle the error (which might be to log it).

filebeat/docs/reference/configuration/filebeat-options.asciidoc
```diff
@@ -41,12 +41,19 @@ supported by https://golang.org/pkg/path/filepath/#Glob[Golang Glob] are also
...
+It is possible to recursively fetch all files in all subdirectories of a directory
+using the optional <<double_star_pattern_depth>> setting.
 
 Filebeat starts a harvester for each file that it finds under the specified
 paths. You can specify one path per line. Each line begins with a dash (-).
 
+===== double_star_pattern_depth
+
+The maximum depth for `**` patterns. This setting is disabled by default, i.e.,
+`**` is not expanded. When it's set to a positive value, a single `**` per
```
andrewkroh: Why the limit in single `**` per path? With a bit of recursion in the expansion logic I think this could be handled elegantly.

filebeat/input/file/glob.go
7AC: 👍

filebeat/docs/reference/configuration/filebeat-options.asciidoc
7AC: What worried me and @tsg was how the complexity could easily escalate here

filebeat/prospector/prospector_log.go
tsg: I'd say no, since for example, I use `""` in the filebeat modules to disable a fileset on a particular version.

filebeat/docs/reference/configuration/filebeat-options.asciidoc
tsg: I was also worried about the number of generated globs. With a max depth of 16, and using `**` twice, you need to generate 16*16 patterns, right? If you use `**` three times that's already 4096 globs to evaluate, which might be problematic.

filebeat/input/file/glob.go
```diff
@@ -0,0 +1,60 @@
...
+)
+
+// globPattern detects the use of "**" and expands it to standard glob patterns up to a max depth
+func globPatterns(pattern string, doubleStarPatternDepth uint8) ([]string, error) {
+	if doubleStarPatternDepth == 0 {
+		return []string{pattern}, nil
+	}
+	var patterns []string
+	isAbs := filepath.IsAbs(pattern)
+	patternList := strings.Split(pattern, string(os.PathSeparator))
```
urso: `os.PathSeparator` might not be safe. windows users sometimes use `/` for paths (if they didn't get `\` working due to not using single-quote), which is working with go as well.

filebeat/docs/reference/configuration/filebeat-options.asciidoc
7AC: That is a bit mitigated by the fact that we default to zero (disabled) now, but I'd still wait for an actual request/use case for it before opening up to multiple **.

filebeat/input/file/glob.go
7AC: Ah interesting, didn't know that. So the windows case is basically a superset, let me handle that too.

filebeat/prospector/prospector_log_test.go
7AC: I couldn't add the last test since I end up joining all the paths to a root that has a volume in it on windows, via `ioutil.TempDir()`

7AC: I created a new test just for cases like this

filebeat/_meta/common.full.p2.yml
```diff
@@ -82,6 +82,11 @@ filebeat.prospectors:
   # This is especially useful for multiline log messages which can get large.
   #max_bytes: 10485760
 
+  # Maximum depth for "**" patterns. This setting is disabled by default, i.e.,
```
ruflin: I somehow expect we will add more config options in the future. So I would suggest to make this a namespace with different config options below. Suggestion:

```
recursive_glob.enabled: false # set this to true to get the defaults for the below options
recursive_glob.depth: 16 # max depth of expanding **
recursive_glob.max_count: 1 # max number of ** in a single path
```
This could also solve the discussion about how many patterns we allow per path.

I named it `recursive_glob` because that what its doing from my perspective. Not sure what the user would look for when searching for such a config option. I had also `recursive_path` because we call it `paths` in the config above.

filebeat/docs/reference/configuration/filebeat-options.asciidoc
ruflin: @7AC See my comment above. I'm ok with having it fixed at first but that we take into account now, it could appear in the future.

filebeat/prospector/config.go
```diff
@@ -13,38 +13,34 @@ import (
...
-		CleanInactive:  0,
-		CleanRemoved:   true,
-		HarvesterLimit: 0,
-		Symlinks:       false,
-		TailFiles:      false,
+		Enabled:       true,
+		DocumentType:  "log",
+		ScanFrequency: 10 * time.Second,
+		InputType:     cfg.DefaultInputType,
+		CleanRemoved:  true,
```
ruflin: What happened to all the other default configs?

7AC: I removed all the ones that have default values for the golang type (including the new one)

ruflin: I strongly prefer to have all values in here, even if they are the golang defaults. This makes it very easy (and future proof) to lookup what our defaults are without having to think about what the default in Golang is.

7AC: Done

filebeat/_meta/common.full.p2.yml
7AC: I renamed to `recursive_glob_depth` for now, I don't think we should encourage users to use more than one per path so the namespace seems overkill to me.

filebeat/filebeat.full.yml
```diff
@@ -211,6 +211,10 @@ filebeat.prospectors:
   # This is especially useful for multiline log messages which can get large.
   #max_bytes: 10485760
 
+  # Maximum depth for "**" patterns. When it's set to a positive value, "**"
+  # patterns are expanded up to the specified depth.
+  #recursive_glob_depth: 16
```
ruflin: Assuming we don't want to make it configurable at first, could we go with `recursive_glob.enabled` as config option instead? This will allow us to extend it in the future if needed. I'm aware that just `recursive_glob: true` would feel more natural at first, but this will prevent us from using it as a namespace in the future.

7AC: Sounds good

filebeat/filebeat.full.yml
```diff
@@ -211,6 +211,11 @@ filebeat.prospectors:
   # This is especially useful for multiline log messages which can get large.
   #max_bytes: 10485760
 
+  ### Recursive glob configuration
+
+  # Expand "**" patterns into regular glob patterns.
+  #recursive_glob.enabled: true
```
ruflin: We normally have the default value in the full config, which is false.

filebeat/docs/reference/configuration/filebeat-options.asciidoc
```diff
@@ -41,12 +41,21 @@ supported by https://golang.org/pkg/path/filepath/#Glob[Golang Glob] are also
...
+fetch log files from the `/var/log` folder itself.
+It is possible to recursively fetch all files in all subdirectories of a directory
+using the optional <<recursive_glob_depth>> setting.
 
 Filebeat starts a harvester for each file that it finds under the specified
 paths. You can specify one path per line. Each line begins with a dash (-).
 
+===== recursive_glob
+
+*`enabled`*:: Enable expanding `**` into recursive glob patterns. With this feature enabled,
```
ruflin: Should we mark this beta first to get some users to test it first? At the same time it rolls out with the 6 alpha release, so we get already some testing.

filebeat/prospector/prospector_log.go
```diff
@@ -118,11 +122,14 @@ func (l *Log) getFiles() map[string]os.FileInfo {
 
 	paths := map[string]os.FileInfo{}
 
-	for _, glob := range l.config.Paths {
-		// Evaluate the path as a wildcards/shell glob
-		matches, err := filepath.Glob(glob)
+	for _, path := range l.config.Paths {
+		depth := uint8(0)
+		if l.config.recursiveGlob.enabled {
```
ruflin: I expected to see here `recursiveGlob.Enabled()` as this come with ucfg for each "object". Problem is that if we want that we would have to do unpack each time the scan runs which is not a good idea. So this looks good to me.

filebeat/prospector/config.go
```diff
@@ -45,6 +46,11 @@ type prospectorConfig struct {
 	Module               string                  `config:"_module_name"`  // hidden option to set the module name
 	Fileset              string                  `config:"_fileset_name"` // hidden option to set the fileset name
 	Processors           processors.PluginConfig `config:"processors"`
+	recursiveGlob        *recursiveGlobConfig    `config:"recursive_glob"`
```
ruflin: why not 
```
recursiveGlob        bool   `config:"recursive_glob.enabled" `
```

filebeat/filebeat.full.yml
7AC: 👍

filebeat/docs/reference/configuration/filebeat-options.asciidoc
7AC: How do I do that?

filebeat/prospector/config.go
7AC: sure

filebeat/docs/reference/configuration/filebeat-options.asciidoc
ruflin: You can add `[]beta` and it gets expanded with a message automatically in the docs. But lets leave it out. We can still add it.

dedemorton: @7AC This won't render correctly in asciidoc because it will try to evaluate the asterisks before the backticks. Wherever you have asterisks in an example, use plus signs (+) instead of backticks. For example: `+**+`

7AC: fixed it thank you!

filebeat/docs/reference/configuration/filebeat-options.asciidoc
```diff
@@ -41,12 +41,21 @@ supported by https://golang.org/pkg/path/filepath/#Glob[Golang Glob] are also
...
+fetch log files from the `/var/log` folder itself.
+It is possible to recursively fetch all files in all subdirectories of a directory
+using the optional <<recursive_glob>> settings.
 
 Filebeat starts a harvester for each file that it finds under the specified
 paths. You can specify one path per line. Each line begins with a dash (-).
 
+===== recursive_glob
+
+*`enabled`*:: Enable expanding +**+ into recursive glob patterns. With this feature enabled,
```
dedemorton: @7AC I think my explanation of the behavior of plus signs was a little off in my previous comment. The backticks are OK for the doc build, but the content doesn't render for users who preview the file in GitHub. We aren't optimizing for viewing in GitHub, though, so technically you can use the backticks. But you should be consistent in this entire description (either use backticks or plus signs for any example that contains asterisks in the path).

filebeat/docs/reference/configuration/filebeat-options.asciidoc
```diff
@@ -41,12 +41,21 @@ supported by https://golang.org/pkg/path/filepath/#Glob[Golang Glob] are also
...
-fetch log files from the `/var/log` folder itself. Currently it is not possible
-to recursively fetch all files in all subdirectories of a directory.
+fetch log files from the `/var/log` folder itself.
+It is possible to recursively fetch all files in all subdirectories of a directory
+using the optional <<recursive_glob>> settings.
 
 Filebeat starts a harvester for each file that it finds under the specified
 paths. You can specify one path per line. Each line begins with a dash (-).
 
+===== recursive_glob
```
dedemorton: @7AC You need to add the section ID above this section or the doc build will fail. So add this to line 50:

`[[recursive_glob]]`

mbhuie:
Is this available in 5.6.3 release?

PR #3986, opened 2017-04-11
fix comment of network IO metrics

st1t:
Fixed comments in network.go.
I think that it is network io instead of disk io.


exekias:
Thank you for your contributing!

PR #3994, opened 2017-04-11
testing: set env in docker entrypoint for logstash

7AC:
Also make the test that revealed the failure a bit more strict


ruflin:
@7AC Thanks a lot for fixing this. The part is not clear to me is why it broke "now"? It seems readParams was not removed recently. 

Should we forward port the changes to the tests you did so we have it also in master for better debugging?

ruflin:
Could be that this was related to https://github.com/elastic/logstash/pull/6897

PR #3999, opened 2017-04-12
Cherry-pick #3929 to 5.x: Don't stop Filebeat when modules + logstash are used together

tsg:
Cherry-pick of PR #3929 to 5.x branch. Original message: 

To simplify the migration from Ingest Node to Logstash, we replace the
hard error ("Filebeat modules require an Elasticsearch output defined") with
a warning.


ruflin:
@tsg Would this need a changelog?

PR #4006, opened 2017-04-12
Bump LICENSE year in beat templates

srenatus:
🎉 😄

### DIFF
--- a/generator/beat/{beat}/LICENSE
+++ b/generator/beat/{beat}/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) 2016 {full_name}
+Copyright (c) 2017 {full_name}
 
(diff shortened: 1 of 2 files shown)


srenatus:
Does the CLA have a "trivial fix" clause? If so, this would sure apply 😉

tsg:
No, I think we don't have such a clause, but it should only take a few minutes to sign it. If you don't want to go through that, I'm happy to "redo" this in a new PR.

srenatus:
Ok, I think I'll go through that 😉 -- It's not too terrible if this fix sits a day longer, I suppose, and it's a good chance to settle this.

srenatus:
I've done the ICLA -- LGTM? 😃

ruflin:
@srenatus Looks good. Thanks for the contribution.

PR #4014, opened 2017-04-13
Add elasticsearch node_stats metricset

ruflin:


ruflin:
jenkins, retest it

metricbeat/module/elasticsearch/node_stats/data.go
```diff
@@ -0,0 +1,60 @@
+package node_stats
```
tsg: We should ignore this lint error, right?

ruflin: yes, because we have also other packages with this kind of names. @7AC can we configure it to ignore these errors?

7AC: Hackish but should do https://github.com/elastic/beats/pull/4015

PR #4015, opened 2017-04-13
reviewdog: ignore underscores in package names

7AC:
Unfortunately this is the best we can do between golint and reviewdog:

https://github.com/golang/lint/issues/98#issuecomment-69813135

### DIFF
--- a/reviewdog.yml
+++ b/reviewdog.yml
@@ -1,3 +1,3 @@
 runner:
   golint:
-    cmd: golint $(go list ./... | grep -v /vendor/)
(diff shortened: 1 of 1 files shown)


andrewkroh:
I ran across this the other day and thought it was sort of interesting. https://github.com/kubernetes-incubator/cri-o/blob/master/.tool/lint

7AC:
`gometalinter` is actually what I run in my editor, and is a superset of `golint` and other tools IIRC. Doesn't look like `reviewdog` supports it though.

Issue #3964, opened 2017-04-09
Filebeat suggestion - allowing for something like multiline.flush_pattern?


TheoAndersen:
I'm using multiline for merging my log-lines together to form application-centric events, which can span multiple events.

I feel like it could be a useful addition, to be able to specify a specific pattern, for when a multiline should stop and flush the event. Right now it seems to be only when either; a new multiline starts, the max-lines count is reached or the timeout is reached.

_This would make it possible to flush an event right when it ends_ - whereas right now the last event i receive, will first be flushed when either one the timeout/max-length is reached, or when a new one begins.

```
multiline
- pattern: "start-line-regex"
- negate: true
- match: after
- flush-pattern: "end-line-regex"
```

What do you think?

Now i having looked at the codebase (not that familiar with go, but i might give it a try if you suggest me to), but I'm thinking that, being that theres already the other kind of flush-guards in place, it would seem conceptually like a small task to add this?

ruflin:
We discussed in the past about a "start/end" pattern but went in the first implementation only for one pattern as it was much simpler to do. I definitively think there is value in having something similar as you described above. Here you can also find the initial discussion about multiline and what options we discussed: https://github.com/elastic/filebeat/issues/301

It reminds me also of https://github.com/elastic/beats/issues/3852 Not sure if the two feature requests could be combined into one, means it is not necessarly only mutiline specific.

TheoAndersen:
The first link you posted is from a private issue, so i can't access it. 

Regarding the #3852 issue about one-line logs, I'm not sure that this flush-pattern option would solve this - dosen't the initial line-matching logic happen outside the multiline module?

I've managed to add the flush-pattern code pretty simply into the current multiline code, and will create a PR soon - Just need to test it a bit more, only TDDd it so far.

PR #4019, opened 2017-04-14
Add multiline.flush_pattern option

TheoAndersen:
This allows for specifying a regex, which will flush the current multiline, thus ending the current multiline. Useful for using multiline to capture application events with 'start' and 'end' lines.

Example configuration
  multiline.pattern: 'start'
  multiline.negate: true
  multiline.match: after
  multiline.flush_pattern: 'end'

(#3964)

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -343,6 +343,7 @@ https://github.com/elastic/beats/compare/v5.2.2...v5.3.0[View commits]
 - The `symlinks` and `harverster_limit` settings are now GA, instead of experimental. {pull}3525[3525]
 - close_timeout is also applied when the output is blocking. {pull}3511[3511]
 - Improve handling of different path variants on Windows. {pull}3781[3781]
+- Add multiline.flush_pattern option, for specifying the 'end' of a multiline pattern {pull}4019[4019]
 
 
 *Metricbeat*
--- a/filebeat/harvester/reader/multiline.go
+++ b/filebeat/harvester/reader/multiline.go
@@ -20,16 +20,17 @@ import (
 // Errors will force the multiline reader to return the currently active
 // multiline event first and finally return the actual error on next call to Next.
 type Multiline struct {
-	reader    Reader
-	pred      matcher
-	maxBytes  int // bytes stored in content
-	maxLines  int
-	separator []byte
-	last      []byte
-	numLines  int
-	err       error // last seen error
-	state     func(*Multiline) (Message, error)
-	message   Message
+	reader       Reader
+	pred         matcher
+	flushMatcher *match.Matcher
+	maxBytes     int // bytes stored in content
+	maxLines     int
+	separator    []byte
+	last         []byte
+	numLines     int
+	err          error // last seen error
+	state        func(*Multiline) (Message, error)
+	message      Message
 }
 
 const (
@@ -71,6 +72,8 @@ func NewMultiline(
 		return nil, err
 	}
 
+	flushMatcher := config.FlushPattern
+
 	if config.Negate {
 		matcher = negatedMatcher(matcher)
 	}
@@ -93,13 +96,14 @@ func NewMultiline(
 	}
 
 	mlr := &Multiline{
-		reader:    reader,
-		pred:      matcher,
-		state:     (*Multiline).readFirst,
-		maxBytes:  maxBytes,
-		maxLines:  maxLines,
-		separator: []byte(separator),
-		message:   Message{},
+		reader:       reader,
+		pred:         matcher,
+		flushMatcher: flushMatcher,
+		state:        (*Multiline).readFirst,
+		maxBytes:     maxBytes,
+		maxLines:     maxLines,
+		separator:    []byte(separator),
+		message:      Message{},
 	}
 	return mlr, nil
 }
@@ -184,6 +188,20 @@ func (mlr *Multiline) readNext() (Message, error) {
 			msg := mlr.finalize()
 			mlr.load(message)
 			return msg, nil
+		}
+
+		// handle case when endPattern is reached
+		if mlr.flushMatcher != nil {
+			endPatternReached := (mlr.flushMatcher.Match(message.Content))
+
+			if endPatternReached == true {
+				// return collected multiline event and
+				// empty buffer for new multiline event
+				mlr.addLine(message)
+				msg := mlr.finalize()
+				mlr.resetState()
+				return msg, nil
+			}
 		}
 
 		// if predicate does not match current multiline -> return multiline event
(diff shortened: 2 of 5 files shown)


TheoAndersen:
I have signed the CLA now.. can somebody trigger it to check again? :)

TheoAndersen:
ahhr... need to run it through gofmt...

ruflin:
@TheoAndersen Thanks a lot for the contribution. I' m thinking if we should have a separate config option for start and end/flush pattern so we don't mix it with the other configs. The part I worry is that "misconfiguration" of all together could lead to strange side effects. 

At the same time the implementation looks rather simple you did

```
multiline.start_pattern:
multiline.end_pattern:
```

TheoAndersen:
@ruflin Thanks for the comments. 

This PR is just the addition of the extra optional parameter ```flush_pattern```, so i don't think it will affect existing configurations, unless you actively use this configuration.

One thing i have found though, is that the way i use this multiline configuration will force everything to be gathered into events - wheres i really just wanted to merge specific lines (matched with a start and end pattern) into a single event - and leave the rest as they are.

Ex if you have a log like this
> Line 1
> Line 2
> Line 3 - Some event starts
> Line 4 - Content of event
> Line 5 - End of event
> Line 6
> Line 7

and use a configuration like this
```
multiline.pattern: 'Some event starts'
multiline.negate: true
multiline.match: after
multiline.flush_pattern: 'End of event'
```
then it will create 3 events.
- one with line 1 to 2 merged
- one with 3 to 5 (the one i really only wanted merged)
- one with line 6 to 7

Where i really only wanted line 3 to 5 merged.

I'm not sure if this is possible to do, by continuing to hack or add on the current multiline implementation? using the ```multiline.negate: true``` will kindof make multiline try to merge all lines in the log wont it?

Do you have any suggestions for this?

Also i would ideally like to be able to specify multiple start-end patterns, if there was multiple kinds of events in there (but this could be done by piping in the regular expression i guess).

ruflin:
I feel like this PR is a good kick of to discuss how we should extend multiline and which use cases it should covered. Unfortunately the initial design doc is now in the filebeat repo which we made private to not have people opening issues there all the time. But in summary there are 2 things in there which we haven't implemented yet:

* Support for multiple patterns instead of just one
* Start / End patterns

What you implemented here is kind of the end pattern, but I agree with you that I would also expect the behaviour that one event for each line is sent, if no start pattern is started.

The initial implementation was done based on some example logs from java stack traces to json logs etc. I would suggest we proceed like this and having directly some real world examples that we want to work on. Perhaps what you did above is already all we need as the other format is more a theoretical one, but not sure.

One of the suggestion from the "old" issue looks as following:

```
# Enable mutline feature
multiline:
  # See description of start_pattern. Can be used alone and in combination with end_pattern
  start_pattern: regexp
  start_pattern_negate: bool

  # See description of end_pattern. Can be used alone and in combination with start_pattern
  end_pattern: regexp
  end_pattern_negate: bool

  # See description of after_pattern. Can only be used alone
  after_pattern: regexp
  after_pattern_negate: bool

  # See description of before_pattern. Can only be used alone
  before_pattern: regexp
  before_pattern_negate: true
```
This can probably be expressed in more compact way.

filebeat/harvester/reader/multiline_test.go
```diff
@@ -60,16 +60,20 @@ func TestMultilineAfterNegateOK(t *testing.T) {
 	)
 }
 
-func TestMultilineBeforeNegateOK(t *testing.T) {
```
ruflin: We should not overwrite an existing test but add a new one.

TheoAndersen:
This should have added the missing test, and the changelog

filebeat/harvester/reader/multiline.go
```diff
@@ -186,6 +190,21 @@ func (mlr *Multiline) readNext() (Message, error) {
 			return msg, nil
 		}
 
+		// handle case when endPattern is reached
+		if mlr.flushMatcher != nil {
+			endPatternReached := (mlr.flushMatcher.Match(message.Content))
+			//		fmt.Println("Next line: " + string(message.Content))
```
ruflin: this line should be removed.

filebeat/harvester/reader/multiline_test.go
```diff
@@ -73,6 +73,23 @@ func TestMultilineBeforeNegateOK(t *testing.T) {
 	)
 }
 
+func TestMultilineAfterNegateOKFlushPattern(t *testing.T) {
+	flushMatcher := match.MustCompile(`EventEnd`)
+
+	testMultilineOK(t,
+		MultilineConfig{
```
ruflin: could you add one more test where first there are lines which do not match the pattern `EventStart`?

TheoAndersen:
Can anybody help me with this build-error?

Can't seem to figure it out ( @ruflin? )

ruflin:
@TheoAndersen Bad timing. There was one commit today which was merged into master and was failing and you exactly got that one :-( Can you rebase on top of master and run `make update` and `make fmt` on the top beats level?

TheoAndersen:
Thats how it goes. Trying again :)

ruflin:
jenkins, test it

TheoAndersen:
Was that just thinking out loud, or does jenkins actually parse these comments?

if so..

Jenkins, test faster ;o)

ruflin:
Yes, jenkins parses the comments, but there are only a few and not all users can trigger the events ;-)

CHANGELOG.asciidoc
```diff
@@ -343,6 +343,12 @@ https://github.com/elastic/beats/compare/v5.2.2...v5.3.0[View commits]
 - The `symlinks` and `harverster_limit` settings are now GA, instead of experimental. {pull}3525[3525]
 - close_timeout is also applied when the output is blocking. {pull}3511[3511]
 - Improve handling of different path variants on Windows. {pull}3781[3781]
+- Restructure input.Event to be inline with outputs.Data {pull}3823[3823]
```
ruflin: Sorry to bug you again, but this seems to have been messed up during rebasing. Unfortunately happens from time to time :-(

ruflin:
All good from my side and ready to merge as soon as changelog is updated.

TheoAndersen:
Updated changelog

ruflin:
@TheoAndersen Merged. Thanks a lot for this contribution.

ppf2:
The changelog indicates that this feature is available in 5.3.0 and above.

https://github.com/elastic/beats/blob/master/CHANGELOG.asciidoc#beats-version-530

But the actual code does not appear until 6.0.  Should we fix the changelog? @ruflin

ruflin:
@ppf2 It definitively seems like this changelog ended up in the completely wrong place. 

@monicasarbu @tsg What should we do here as I remember our CHANGELOG is also in some release etc. I assume it's not solved by just removing it in 5.x.

PR #4023, opened 2017-04-16
Support Tencent Cloud provider for add_cloud_metadata proccessor

athom:
Hey there,
Currently the `add_cloud_metadata` processor supports  3 main providers by default, it's awesome!

However, a lot of companies in China prefer to use `Tencent Cloud` and `AliCloud`.  AFAIK, the two providers did hold a large marke share. It will be nice to support these two providers by default too.

This PR amis to support Tencent Cloud. Unlike the other 3 providers, Tencent Cloud Metadata behaves a bit different. 
You can not get the full JSON format metadata in one single round trip, only one field of the metadata can by queried every request. 
For example, you wanna to get the the metadata contains `instance-id` and `region`, you have to send two requests separately.

```
curl http://metadata.tencentyun.com/meta-data/instance-id
```

```
curl http://metadata.tencentyun.com/meta-data/placement/region
```
Here is the [document](https://www.qcloud.com/document/product/213/4934) 

Summary of the changes:

- refactor and abstract a data struct named `fetcher` who did the network/text processing job for specified provider.
- generalize origin `fetchJSON` to `fetcher.fetchRaw`.
- allow multiple requests to fetch the full metadata in the provider's fetcher level.
- replace hard code of the goroutinue number in  waiting loop with dynamic length cauculation.

Ps, I tested it on my qcloud instance with os `ubuntu-16.04` , I can offer a instance for testing, feel free to ping me if you need one.

### DIFF
--- a/libbeat/docs/processors-config.asciidoc
+++ b/libbeat/docs/processors-config.asciidoc
@@ -266,12 +266,12 @@ The `add_cloud_metadata` processor enriches each event with instance metadata
 from the machine's hosting provider. At startup it will detect the hosting
 provider and cache the instance metadata.
 
-Four cloud providers are supported.
+The following cloud providers are supported:
 
 - Amazon Elastic Compute Cloud (EC2)
 - Digital Ocean
 - Google Compute Engine (GCE)
-- [Tencent Cloud](https://www.qcloud.com/?lang=en) (QCloud)
+- https://www.qcloud.com/?lang=en[Tencent Cloud] (QCloud)
 
 The simple configuration below enables the processor.
 
--- a/libbeat/docs/processors-config.asciidoc
+++ b/libbeat/docs/processors-config.asciidoc
@@ -266,11 +266,12 @@ The `add_cloud_metadata` processor enriches each event with instance metadata
 from the machine's hosting provider. At startup it will detect the hosting
 provider and cache the instance metadata.
 
-Three cloud providers are supported.
+Four cloud providers are supported.
 
 - Amazon Elastic Compute Cloud (EC2)
 - Digital Ocean
 - Google Compute Engine (GCE)
+- [Tencent Cloud](https://www.qcloud.com/?lang=en) (QCloud)
 
 The simple configuration below enables the processor.
 
@@ -336,6 +337,22 @@ _GCE_
       "machine_type": "projects/1234567890/machineTypes/f1-micro",
       "project_id": "my-dev",
       "provider": "gce"
+    }
+  }
+}
+-------------------------------------------------------------------------------
+
+_Tencent Clound_
+
+[source,json]
+-------------------------------------------------------------------------------
+{
+  "meta": {
+    "cloud": {
+      "availability_zone": "gz-azone2",
+      "instance_id": "ins-qcloudv5",
+      "provider": "qcloud",
+      "region": "china-south-gz"
     }
   }
 }
--- a/libbeat/processors/add_cloud_metadata/add_cloud_metadata.go
+++ b/libbeat/processors/add_cloud_metadata/add_cloud_metadata.go
@@ -34,14 +34,17 @@ const (
 
 	// Tencent Clound Metadata Service
 	qcloudMetadataHost          = "metadata.tencentyun.com"
-	qcloudMetadataInstanceIdURI = "/meta-data/instance-id"
+	qcloudMetadataInstanceIDURI = "/meta-data/instance-id"
 	qcloudMetadataRegionURI     = "/meta-data/placement/region"
 	qcloudMetadataZoneURI       = "/meta-data/placement/zone"
+
+	// Default config
+	defaultTimeOut = 3 * time.Second
 )
 
 var debugf = logp.MakeDebug("filters")
 
-// metadata schemas for all prividers
+// metadata schemas for all prividers.
 var (
 	ec2Schema = func(m map[string]interface{}) common.MapStr {
 		out, _ := s.Schema{
@@ -83,12 +86,7 @@ var (
 	}
 
 	qcloudSchema = func(m map[string]interface{}) common.MapStr {
-		out, _ := s.Schema{
-			"instance_id":       c.Str("instance_id"),
-			"region":            c.Str("region"),
-			"availability_zone": c.Str("zone"),
-		}.Apply(m)
-		return out
+		return common.MapStr(m)
 	}
 )
 
@@ -99,21 +97,23 @@ func init() {
 
 type schemaConv func(m map[string]interface{}) common.MapStr
 
-type pick func([]byte, *result) error
+// responseHandler is the callback function that used to write something
+// to the result according the HTTP response.
+type responseHandler func(all []byte, res *result) error
 
 type metadataFetcher struct {
-	provider string
-	headers  map[string]string
-	pickers  map[string]pick
-	conv     schemaConv
+	provider         string
+	headers          map[string]string
+	responseHandlers map[string]responseHandler
+	conv             schemaConv
 }
 
-// fetchRaw query raw metadata from a hosting provider's metadata service.
+// fetchRaw queries raw metadata from a hosting provider's metadata service.
 func (f *metadataFetcher) fetchRaw(
 	ctx context.Context,
 	client http.Client,
 	url string,
-	pick pick,
+	responseHandler responseHandler,
 	result *result,
 ) {
 	req, err := http.NewRequest("GET", url, nil)
@@ -145,7 +145,7 @@ func (f *metadataFetcher) fetchRaw(
 	}
 
 	// Decode JSON.
-	err = pick(all, result)
+	err = responseHandler(all, result)
 	if err != nil {
 		result.err = err
 		return
@@ -154,13 +154,13 @@ func (f *metadataFetcher) fetchRaw(
 	return
 }
 
-// fetchMetadata query metadata from a hosting provider's metadata service.
-// some providers require multiple HTTP request to gather the whole metadata
-// len(f.pickers)  > 1 indicates that multiple requesting is needed
+// fetchMetadata queries metadata from a hosting provider's metadata service.
+// Some providers require multiple HTTP requests to gather the whole metadata,
+// len(f.responseHandlers)  > 1 indicates that multiple requests are needed.
 func (f *metadataFetcher) fetchMetadata(ctx context.Context, client http.Client) result {
 	res := result{provider: f.provider, metadata: common.MapStr{}}
-	for url, pick := range f.pickers {
-		f.fetchRaw(ctx, client, url, pick, &res)
+	for url, responseHandler := range f.responseHandlers {
+		f.fetchRaw(ctx, client, url, responseHandler, &res)
 		if res.err != nil {
 			return res
 		}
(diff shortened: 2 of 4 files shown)


libbeat/processors/add_cloud_metadata/add_cloud_metadata.go
```diff
@@ -31,10 +31,17 @@ const (
 
 	// Google GCE Metadata Service
 	gceMetadataURI = "/computeMetadata/v1/?recursive=true&alt=json"
+
+	// Tencent Clound Metadata Service
+	qcloudMetadataHost          = "metadata.tencentyun.com"
```
ruflin: Now that we are starting to add more cloud providers it would be nice to have each provider in its own file. This makes it easier to spot which parts are specific to a provider and which is the shared code. Just a general note, could also be done in a follow up PR.

athom: Good idea, I will apply it when implement AliCloud metadata in next PR ;)

libbeat/docs/processors-config.asciidoc
```diff
@@ -266,11 +266,12 @@ The `add_cloud_metadata` processor enriches each event with instance metadata
 from the machine's hosting provider. At startup it will detect the hosting
 provider and cache the instance metadata.
 
-Three cloud providers are supported.
+Four cloud providers are supported.
```
ruflin: We can probably drop the number here. More like `The following cloud providers are supported:`

libbeat/processors/add_cloud_metadata/add_cloud_metadata.go
```diff
@@ -209,30 +248,137 @@ func fetchMetadata(doURL, ec2URL, gceURL string, timeout time.Duration) *result
...
+			result.metadata["zone"] = string(all)
+			return nil
+		},
+	}
+	fetcher := &metadataFetcher{"qcloud", nil, picker, qcloudSchema}
+	return fetcher, nil
+}
+
+func newCloudMetadata(c common.Config) (processors.Processor, error) {
+	timeout := 3 * time.Second
```
ruflin: Could we introduce a constant for that so it is easy to find? Something like `defaultTimeout`?

andrewkroh:
jenkins, test it

libbeat/processors/add_cloud_metadata/add_cloud_metadata.go
```diff
@@ -209,30 +248,137 @@ func fetchMetadata(doURL, ec2URL, gceURL string, timeout time.Duration) *result
 	return nil
 }
 
-type addCloudMetadata struct {
-	metadata common.MapStr
-}
-
-func newCloudMetadata(c common.Config) (processors.Processor, error) {
+// getMetadataURL loads config and generate metadata url
```
andrewkroh: `s/generate metadata url/generates the metadata URLs./`

libbeat/processors/add_cloud_metadata/add_cloud_metadata.go
```diff
@@ -209,30 +248,137 @@ func fetchMetadata(doURL, ec2URL, gceURL string, timeout time.Duration) *result
...
-		return nil, errors.Wrap(err, "failed to unpack add_cloud_metadata config")
+		return urls, errors.Wrap(err, "failed to unpack add_cloud_metadata config")
+	}
+	for _, uri := range metadataURIs {
+		urls = append(urls, "http://"+config.MetadataHostAndPort+uri)
 	}
+	return urls, nil
+}
+
+// makeCommomJSONPicker generate fetch function which query json metadata from a hosting provider's HTTP response
```
andrewkroh: I think you could drop `Common` from the name. May I suggest:

```
makeJSONPicker returns a pick function that unmarshals JSON from a hosting provider's
HTTP response and writes it to the result.
```

libbeat/processors/add_cloud_metadata/add_cloud_metadata.go
```diff
@@ -209,30 +248,137 @@ func fetchMetadata(doURL, ec2URL, gceURL string, timeout time.Duration) *result
...
-	)
+// newMetadataFetcher return metadataFetcher with one pass json picker
+func newMetadataFetcher(
+	c common.Config,
+	timeout time.Duration,
+	provider string,
+	headers map[string]string,
+	host string,
+	conv schemaConv,
+	uris []string,
```
andrewkroh: Why accept a list if only one is ever used?

libbeat/processors/add_cloud_metadata/add_cloud_metadata.go
```diff
@@ -209,30 +248,137 @@ func fetchMetadata(doURL, ec2URL, gceURL string, timeout time.Duration) *result
 	return nil
 }
 
-type addCloudMetadata struct {
-	metadata common.MapStr
-}
-
-func newCloudMetadata(c common.Config) (processors.Processor, error) {
+// getMetadataURL loads config and generate metadata url
+func getMetadataURL(c common.Config, defaultHost string, timeout time.Duration, metadataURIs []string) ([]string, error) {
```
andrewkroh: This returns a slice of URLs so I'm thinking it should be `getMetadataURLs`?

libbeat/processors/add_cloud_metadata/add_cloud_metadata.go
```diff
@@ -209,30 +248,137 @@ func fetchMetadata(doURL, ec2URL, gceURL string, timeout time.Duration) *result
...
+func newQcloudMetadataFetcher(c common.Config, timeout time.Duration) (*metadataFetcher, error) {
+	urls, err := getMetadataURL(c, qcloudMetadataHost, timeout, []string{
+		qcloudMetadataInstanceIdURI,
+		qcloudMetadataRegionURI,
+		qcloudMetadataZoneURI,
+	})
+	if err != nil {
+		return nil, err
+	}
+	picker := map[string]pick{
```
andrewkroh: `s/picker/pickers/`?

libbeat/processors/add_cloud_metadata/add_cloud_metadata.go
```diff
@@ -209,30 +248,137 @@ func fetchMetadata(doURL, ec2URL, gceURL string, timeout time.Duration) *result
...
+func getMetadataURL(c common.Config, defaultHost string, timeout time.Duration, metadataURIs []string) ([]string, error) {
+	var urls []string
 	config := struct {
 		MetadataHostAndPort string        `config:"host"`    // Specifies the host and port of the metadata service (for testing purposes only).
 		Timeout             time.Duration `config:"timeout"` // Amount of time to wait for responses from the metadata services.
 	}{
-		MetadataHostAndPort: metadataHost,
-		Timeout:             3 * time.Second,
+		MetadataHostAndPort: defaultHost,
+		Timeout:             timeout,
```
andrewkroh: This `Timeout` value is not used here so it needs to be relocated.

libbeat/processors/add_cloud_metadata/add_cloud_metadata.go
andrewkroh: This timeout value should be unpacked from the configuration data. Otherwise it is hardcoded here.

libbeat/processors/add_cloud_metadata/add_cloud_metadata.go
```diff
@@ -31,10 +31,17 @@ const (
...
+	// Tencent Clound Metadata Service
+	qcloudMetadataHost          = "metadata.tencentyun.com"
+	qcloudMetadataInstanceIdURI = "/meta-data/instance-id"
+	qcloudMetadataRegionURI     = "/meta-data/placement/region"
+	qcloudMetadataZoneURI       = "/meta-data/placement/zone"
 )
 
 var debugf = logp.MakeDebug("filters")
 
+// metadata schemas for all prividers
```
andrewkroh: `s/prividers/providers./`

libbeat/processors/add_cloud_metadata/add_cloud_metadata.go
```diff
@@ -74,78 +81,108 @@ var (
...
 		return out
 	}
+
+	qcloudSchema = func(m map[string]interface{}) common.MapStr {
+		out, _ := s.Schema{
+			"instance_id":       c.Str("instance_id"),
+			"region":            c.Str("region"),
+			"availability_zone": c.Str("zone"),
+		}.Apply(m)
+		return out
```
andrewkroh: It seems like this doesn't even need to use the `Schema` as you have already built up the `map[string]interface{}` with the proper key names and data types. So I think this could just become `return common.MapStr(m)`.

libbeat/processors/add_cloud_metadata/add_cloud_metadata.go
```diff
@@ -74,78 +81,108 @@ var (
...
+type metadataFetcher struct {
+	provider string
+	headers  map[string]string
+	pickers  map[string]pick
+	conv     schemaConv
 }
 
-// fetchJSON query metadata from a hosting provider's metadata service.
-func fetchJSON(
+// fetchRaw query raw metadata from a hosting provider's metadata service.
```
andrewkroh: `s/query/queries/`

libbeat/docs/processors-config.asciidoc
athom: Good suggestion

libbeat/processors/add_cloud_metadata/add_cloud_metadata.go
athom: yep, it's hardcoded here, already move timeout configuration pack out of the getMetadataURLs and put it directly in newCloudMetadata.
The pool config is packed 5 times here 🤒, but I think it is acceptable here since it only run once in the startup phase.

libbeat/docs/processors-config.asciidoc
```diff
@@ -266,11 +266,12 @@ The `add_cloud_metadata` processor enriches each event with instance metadata
...
 
 - Amazon Elastic Compute Cloud (EC2)
 - Digital Ocean
 - Google Compute Engine (GCE)
+- [Tencent Cloud](https://www.qcloud.com/?lang=en) (QCloud)
```
athom: 🤦‍♂️  a clumsy product placement 😢
sorry for really green hand in asciidoc.

libbeat/processors/add_cloud_metadata/add_cloud_metadata.go
```diff
@@ -31,10 +31,17 @@ const (
 
 	// Google GCE Metadata Service
 	gceMetadataURI = "/computeMetadata/v1/?recursive=true&alt=json"
+
+	// Tencent Clound Metadata Service
+	qcloudMetadataHost          = "metadata.tencentyun.com"
+	qcloudMetadataInstanceIdURI = "/meta-data/instance-id"
+	qcloudMetadataRegionURI     = "/meta-data/placement/region"
+	qcloudMetadataZoneURI       = "/meta-data/placement/zone"
```
athom: AFAIK, there is no plan for Tentcent Qcloud endpoints to support `/v1/xxx`,  the metadata service is maintained in my team :)
And the other provider AliCloud also has no versioned endpoints neither, at least no indications from the Chinese documents.

libbeat/processors/add_cloud_metadata/add_cloud_metadata.go
athom: done

libbeat/processors/add_cloud_metadata/add_cloud_metadata.go
```diff
@@ -74,78 +81,108 @@ var (
...
-	provider string        // Hosting provider type.
-	err      error         // Error that occurred while fetching (if any).
-	metadata common.MapStr // A specific subset of the metadata received from the hosting provider.
-}
+type schemaConv func(m map[string]interface{}) common.MapStr
 
-func (r result) String() string {
-	return fmt.Sprintf("result=[provider:%v, error=%v, metadata=%v]",
-		r.provider, r.err, r.metadata)
+type pick func([]byte, *result) error
```
athom: good point, change applied.

libbeat/processors/add_cloud_metadata/add_cloud_metadata.go
athom: relocated.

libbeat/processors/add_cloud_metadata/add_cloud_metadata.go
athom: good point, change to single string now.

andrewkroh:
jenkins, test it

ruflin:
@athom Thanks a lot for the contribution.

athom:
Thanks for the merge

PR #4033, opened 2017-04-18
Update Go version to 1.8.1

tsg:
Doing the upgrade now because older Go versions fail to produce valid binaries on up-to-date macOs systems.

Includes:

* packer docker images
* travis & appveoyor
* Dockerfiles

@andrewkroh Anything missing for the Jenkins infra?


tsg:
I used the UI to resolve the conflict, but I think it can still be squashed.

PR #4043, opened 2017-04-19
Adding support for include_labels and include_annotations in kubernetes processor

vjsamuel:
`include_labels: ["foo"]` would make sure that only foo is part of the "labels" map in the kubernetes metadata.

if user wishes to add select annotations then it can be done using `include_annotations: ["xyz"]`

### DIFF
--- a/libbeat/processors/annotate/kubernetes/indexing.go
+++ b/libbeat/processors/annotate/kubernetes/indexing.go
@@ -48,6 +48,12 @@ type Matcher interface {
 	MetadataIndex(event common.MapStr) string
 }
 
+//GenMeta takes in pods to generate metadata for them
+type GenMeta interface {
+	//GenerateMetaData generates metadata by taking in a pod as an input
+	GenerateMetaData(pod *corev1.Pod) common.MapStr
+}
+
 type Indexers struct {
 	sync.RWMutex
 	indexers []Indexer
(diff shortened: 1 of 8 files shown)


libbeat/processors/annotate/kubernetes/indexing.go
```diff
@@ -167,28 +173,90 @@ func (m *Matchers) MetadataIndex(event common.MapStr) string {
...
 	}
+
+	if len(labelMap) != 0 {
+		meta["labels"] = labelMap
+	}
+
+	return meta
+}
+
+type GenFilteredMeta struct {
```
exekias: I'm wondering, why do you need 2 different metadata generators? All the filtering logic could go into the default, isn't it?

libbeat/processors/annotate/kubernetes/kubernetes.go
```diff
@@ -52,18 +52,44 @@ func newKubernetesAnnotator(cfg common.Config) (processors.Processor, error) {
 		return nil, err
 	}
 
+	//Load default indexer configs
+	if config.DefaultIndexers.Enabled == true {
+		for key, cfg := range Indexing.defaultIndexerConfigs {
+			config.Indexers = append(config.Indexers, map[string]common.Config{key: cfg})
```
exekias: this lacks rwmutex logic

libbeat/processors/annotate/kubernetes/kubernetes.go
```diff
@@ -52,18 +52,44 @@ func newKubernetesAnnotator(cfg common.Config) (processors.Processor, error) {
...
+		}
+	}
+
+	//Load default matcher configs
+	if config.DefaultMatchers.Enabled == true {
+		for key, cfg := range Indexing.defaultMatcherConfigs {
+			config.Matchers = append(config.Matchers, map[string]common.Config{key: cfg})
```
exekias: rwmutex too

libbeat/processors/annotate/kubernetes/kubernetes.go
vjsamuel: would this need the rwmutex part? processor init is anyway something that is serial

exekias: that's as of today, some day we may add changes like processors reloading that could mess with concurrency. I prefer just to trust the functions are thread safe as other modules are calling them

Issue #3917, opened 2017-04-05
Filebeat panic when no paths are defined in one of two prospectors


tsg:
Version: master

The following commit causes Filebeat to panic:

```
filebeat.prospectors:
- input_type: log
  paths:
    - /var/log/*.log
- input_type: log
  paths: []

output.elasticsearch:
  hosts: ["localhost:9200"]
```

Logs plus panic (expand): <details>
```
 ./filebeat -e -c filebeat.bug.yml
2017/04/05 09:35:07.077266 beat.go:325: INFO Home path: [/Users/tsg/src/github.com/elastic/beats/filebeat] Config path: [/Users/tsg/src/github.com/elastic/beats/filebeat] Data path: [/Users/tsg/src/github.com/elastic/beats/filebeat/data] Logs path: [/Users/tsg/src/github.com/elastic/beats/filebeat/logs]
2017/04/05 09:35:07.077306 beat.go:350: INFO Beat metadata path: /Users/tsg/src/github.com/elastic/beats/filebeat/data/meta.json
2017/04/05 09:35:07.077323 metrics.go:23: INFO Metrics logging every 30s
2017/04/05 09:35:07.077421 beat.go:332: INFO Beat UUID: 5919a89e-fed5-4e81-86d9-da3b852d978b
2017/04/05 09:35:07.077438 beat.go:210: INFO Setup Beat: filebeat; Version: 6.0.0-alpha1
2017/04/05 09:35:07.077704 client.go:131: INFO Elasticsearch url: http://localhost:9200
2017/04/05 09:35:07.077745 outputs.go:107: INFO Activated elasticsearch as output plugin.
2017/04/05 09:35:07.077792 publish.go:191: INFO Publisher name: a-mac-with-esc-key-2.local
2017/04/05 09:35:07.077967 async.go:63: INFO Flush Interval set to: 1s
2017/04/05 09:35:07.077978 async.go:64: INFO Max Bulk Size set to: 50
2017/04/05 09:35:07.078243 beat.go:253: INFO filebeat start running.
2017/04/05 09:35:07.078278 registrar.go:83: INFO Registry file set to: /Users/tsg/src/github.com/elastic/beats/filebeat/data/registry
2017/04/05 09:35:07.078308 registrar.go:104: INFO Loading registrar data from /Users/tsg/src/github.com/elastic/beats/filebeat/data/registry
2017/04/05 09:35:07.078331 registrar.go:115: INFO States Loaded from registrar: 0
[...]

PR #4037, opened 2017-04-18
Properly shut down crawler in case one prospector is misconfigured

ruflin:
If one prospector started to already send data and a second one was misconfigured, the beat paniced during shutdown. This is no prevented by properly shutting down the crawler also on error.

Closes https://github.com/elastic/beats/issues/3917

PR #4047, opened 2017-04-19
Backport #4037: Properly shut down crawler in case one prospector is misconfigured

ruflin:
If one prospector started to already send data and a second one was misconfigured, the beat paniced during shutdown. This is no prevented by properly shutting down the crawler also on error.

Closes https://github.com/elastic/beats/issues/3917
(cherry picked from commit 95195cc855af9a345a67e160fc20c35529478081)


ruflin:
jenkins, retest it

PR #4049, opened 2017-04-19
Update metricbeat full config to have all modules disabled

ruflin:
Currently modules in the full config were disabled by commenting it out. Unfortunately we were not very consistent in the commenting style and some were missing the comments which enabled some of the modules.

Having the configs commented out also meant they are commented out in the documentation as the same config files are used to generate the docs.

Only system module has a short config as it is the one enabled by default. For all the other modules I merged removed the short config as we should come up with one as soon as we need it, means it's in the short config. So the full config files for the modules were removed and merged into the short one.

This change in the full config will also simplify copy / paste of config options for adding it to the short config. Only thing that has to be changed is the enabled flag.

Experimental log message and note are now added to each new metricset by default.


monicasarbu:
@ruflin #4038 is merged now, can you please rebase?

ruflin:
@monicasarbu Rebased

PR #4051, opened 2017-04-19
Add filebeat.config.path as replacement for filebeat.config_dir

ruflin:
The reload feature can also be used to load configuration files once. This enables the feature to load config options once through `filebeat.config.path: ...`. The difference to config_dir is that only the prospectors have to be configured in the file and not the full filebeat config tree.

`filebeat.config_dir` will be removed in a follow up PR.

As the reloading is now also used for just loading, the code could use some refactoring in the future to make this more obvious.


ruflin:
jenkins, retest it

ruflin:
jenkins, retest it

Prinzka:
This does not appear to be functional.
I've tested on both 6.1 and 6.0.0.1
Filebeat starts without errors but does not load any prospectors from files in the config path.
I simply filebeat.config.path back to filebeat.config.dir and it loads all the prospectors in the config path without issue. It throws the deprecated warning, but at least it's functional.

ruflin:
@Prinzka Please open a topic on discuss and share your config files and filebeat logs there: https://discuss.elastic.co/c/beats/filebeat

Issue #3965, opened 2017-04-10
Follow up kubelet module


exekias:
Finish all the rough edges after initial merge: https://github.com/elastic/beats/pull/3916

- [x] Complete fields.yml files with all fields
- [x] Check fields again if they follow the conventions
- [x] Add system tests to check if all fields are documented
- [x] Add tests
- [x] Add more documentation on how to use it etc.
  - [x] Add documentation on how metricbeat should be deployed inside kubernetes
- [x] Check that all beta flags exist and work.
- [x] Add dashboard/(s)

exekias:
Documentation to deploy kubernetes was added, I think we can close this

PR #4053, opened 2017-04-19
Add kubelet module fields

exekias:
Add all missing fields for `kubelet` module.

Related to #3965

### DIFF
--- a/metricbeat/module/kubelet/container/_meta/fields.yml
+++ b/metricbeat/module/kubelet/container/_meta/fields.yml
@@ -36,6 +36,7 @@
           fields:
             - name: bytes
               type: long
+              format: bytes
               description: >
                 Logs available capacity in bytes
         - name: capacity
@@ -43,6 +44,7 @@
           fields:
             - name: bytes
               type: long
+              format: bytes
               description: >
                 Logs total capacity in bytes
         - name: used
@@ -50,6 +52,7 @@
           fields:
             - name: bytes
               type: long
+              format: bytes
               description: >
                 Logs used capacity in bytes
         - name: inodes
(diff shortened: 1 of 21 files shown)


metricbeat/module/kubelet/container/_meta/fields.yml
```diff
@@ -1,9 +1,139 @@
...
+      type: group
+      description: >
+        Logs info
+      fields:
+        - name: available
+          type: group
+          fields:
+            - name: bytes
+              type: long
+              description: >
```
ruflin: you can add here `formate: bytes` so the Kibana index pattern will contain it and will directly make sense of the unit.

metricbeat/module/kubelet/container/_meta/fields.yml
```diff
@@ -1,9 +1,139 @@
...
+                Logs total capacity in bytes
+        - name: used
+          type: group
+          fields:
+            - name: bytes
```
ruflin: See commetn above. Applies to all bytes fields.

exekias:
I added unit tests, as you say, mistakes are still possible, let's see if we can improve integration tests later on

metricbeat/module/kubelet/container/_meta/fields.yml
```diff
@@ -1,9 +1,139 @@
...
+              type: bytes
+              description: >
+                Logs total capacity in bytes
+        - name: used
+          type: group
+          fields:
+            - name: bytes
+              type: bytes
```
ruflin: You have to add:

```
type: long
format: bytes
```
In ES it is still just a long, only the kibana index pattern knows about bytes. Same for all others.

exekias: Ah sorry, that makes much more sense, should be ok now :innocent:

Issue #4029, opened 2017-04-17
Fix the structure of the Filebeat modules doc


dedemorton:
The content across the different modules is not structured consistently. In some cases the content is broken into several topics; in some cases it's only a couple of topics. It might be best to have one topic for each module. For example, all the content about the apache2 module would appear here:

https://www.elastic.co/guide/en/beats/filebeat/5.3/filebeat-module-apache2.html

And so on for other modules.

I also think that we need to be more clear with users about what the fileset settings do. If they skip over the intro material, the topic about the specific modules don't make sense. I think we could do better here.

dedemorton:
Fixed in #4065. I'll fix the fileset var description when I edit the content. I'll fix the description of the fileset settings when I edit the content.

PR #4065, opened 2017-04-20
Fix structure of module topics

dedemorton:
Adds float tags to fix the structure of the topics about each module type. Also fixed a couple of typos.

Needs backport to 5.4.


dedemorton:
This fixes https://github.com/elastic/beats/issues/4029.

PR #4066, opened 2017-04-20
Change type of ip field

### LINKED ISSUE #3926: Heartbeat - Use IP type in Elasticsearch mapping
For Heartbeat it would be nice to the the `ip` type in Elasticsearch mapping for the `ip` field. This would enable users to answer questions like, "what's the average RTT for hosts in the subnet where `ip:10.10.0.0/16`?"

- Version: 5.3

glefloch:
Update type of ip field to use ip type from elasticsearch.

Close #3926

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -65,6 +65,7 @@ https://github.com/elastic/beats/compare/v5.1.1...master[Check the HEAD diff]
 
 *Heartbeat*
 - Add default ports in HTTP monitor. {pull}3924[3924]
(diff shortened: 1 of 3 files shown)


ruflin:
@glefloch Thanks for the contribution. Could you add a line to the changelog? Even though the affect of this is probably minor I think it could have an affect on some queries / dashboards people built for heartbeat.

ruflin:
@glefloch You must run `make update` in the metricbeat directory so also docs are updated with the change.

ruflin:
@glefloch Thanks for the contribution.

PR #4077, opened 2017-04-21
Enhance node_stats metricset with index stats

ruflin:
Most of the index stats were added to the node_stats metricset. This allows to fetch the metric stats from each node instead of globally. The difference between these stats is that these are node specific. That means if there are multiple shards / replicas the sum of all doc counts from all nodes is not the same as the doc count in all indices.

Remove stats metricset as not needed anymore.


exekias:
LGTM

ruflin:
jenkins, retest it

PR #4079, opened 2017-04-21
Set 10000 field limit also for ES v5

ruflin:
This was accidentially removed in https://github.com/elastic/beats/pull/4055


exekias:
WFG

monicasarbu:
It works now, thanks @ruflin for fixing it!

PR #4080, opened 2017-04-21
Change template loading to be part of the setup

ruflin:
* Change config option to setup.template.* from outputs.elasticsearch.template.*
* Move loading logic into template package
* Remove template loading logic from elasticsearch output
* Changelog updated
* Template tests were moved from output to template package
* Documentation was updated. Will need some more work for which a follow up Github issue will be created.
* Add `GetVersion()` to elasticsearch client.
* Introduce callback registration for elasticsearch output. This should be generalised later. The template loading registers only with the output client factory which means, the template is not loaded when connecting for loading dashboards, pipeline or monitoring data which is intended.

This is only migration the existing options. New options like outputting to a json file or load additional config options will be added in a follow up PR.

Part of elastic#3654 and elastic#3921

### DIFF
--- a/libbeat/beat/beat.go
+++ b/libbeat/beat/beat.go
@@ -58,6 +58,7 @@ import (
 	"github.com/elastic/beats/libbeat/processors"
 	"github.com/elastic/beats/libbeat/publisher"
 	svc "github.com/elastic/beats/libbeat/service"
+	"github.com/elastic/beats/libbeat/template"
 	"github.com/elastic/beats/libbeat/version"
 
 	// Register default processors.
@@ -113,6 +114,7 @@ type BeatConfig struct {
 	Processors processors.PluginConfig   `config:"processors"`
 	Path       paths.Path                `config:"path"`
 	Dashboards *common.Config            `config:"setup.dashboards"`
+	Template   *common.Config            `config:"setup.template"`
 	Http       *common.Config            `config:"http"`
 }
 
@@ -213,6 +215,11 @@ func (b *Beat) launch(bt Creator) error {
 	processors, err := processors.New(b.Config.Processors)
 	if err != nil {
 		return fmt.Errorf("error initializing processors: %v", err)
+	}
+
+	err = b.registerTemplateLoading()
+	if err != nil {
+		return err
 	}
 
 	debugf("Initializing output plugins")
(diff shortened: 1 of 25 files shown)


libbeat/beat/beat.go
```diff
@@ -252,6 +254,11 @@ func (b *Beat) launch(bt Creator) error {
 		return err
 	}
 
+	err = b.loadTemplate()
```
tsg: Hmm, if I understand it correctly, this means that the template loading is only attempted once at startup not on every connection made to ES? If this is the case, then this is a regression for two reasons:

* If ES is not up when the Beat start, the Beat will fail with an error. It used to "wait" for ES to come up
* If for whatever reason ES is restarted without data, but the Beat is not restarted, the template is not loaded again, resulting in bad mapping.

libbeat/beat/beat.go
```diff
@@ -447,6 +454,48 @@ func (b *Beat) loadDashboards() error {
 	return nil
 }
 
+func (b *Beat) loadTemplate() error {
+	if *setup {
```
tsg: The template should be always loaded, not only on `-setup`.

libbeat/beat/beat.go
```diff
@@ -447,6 +454,48 @@ func (b *Beat) loadDashboards() error {
...
+		err := b.Config.Template.SetBool("enabled", -1, true)
+		if err != nil {
+			return fmt.Errorf("Error setting template.enabled=true: %v", err)
+		}
+	}
+
+	esConfig := b.Config.Output["elasticsearch"]
+
+	// Loads template by default if esOutput is enabled
+	if (b.Config.Template == nil && esConfig.Enabled()) || (b.Config.Dashboards != nil && b.Config.Template.Enabled()) {
```
tsg: There's a reference to `b.Config.Dasboards` there, by mistake?

libbeat/beat/beat.go
ruflin: Good point. This just go a lot trickier to implement as it now uses its own client.

libbeat/beat/beat.go
ruflin: fixed

libbeat/beat/beat.go
ruflin: It should be always loaded. The check on line 472 is different from the Dashboard one.

libbeat/beat/beat.go
tsg: TBH this is a deal breaker for me, especially the part where Beats don't start if ES is not up is going to really annoying to users. 

Ideally we'd also have the pipelines loaded on each connection. It's currently done only on startup, but that can lead to errors similar to these. I think we need some sort of generic callback mechanism to do this.

ruflin: agree

libbeat/beat/beat.go
tsg: Ok, but then this check here is not needed?

libbeat/outputs/elasticsearch/client.go
```diff
@@ -71,6 +71,12 @@ type Connection struct {
 	version string
 }
 
+var connectCallbackRegister connectCallback
```
tsg: Having this as global means it will also be applied to non-output ES clients (e.g. the one loading the dashboards, or the one loading the pipelines or the one for monitoring). That is perhaps OK, but mentioning it in case someone else thinks of something about that.

Of course, the global is kind of last resort solution. @urso any suggestions on how we can improve this? (Not necessarily in this PR).

ruflin: That is also getting interesting in the case we would allow in the setup to define an elasticsearch host to load the template potentially with a different user. We already have something similar for the metrics.

Talking about metrics, this means also for the connection to the monitoring instance it would load the template? Not sure if this is intended.

libbeat/beat/beat.go
ruflin: if you run `-setup` but have `setup.template.enabled: false` what is the expected behaviour?

PR #4087, opened 2017-04-22
Skip empty events without error

ruflin:
In some cases, a metricset does not return an event but also not an error. This can be the case if a metricset for a distributed system should only send events for the master. All metricbeat instances are running on all nodes, but only one node should send the events and the other skip the event. This changes makes this behaviour possible.


monicasarbu:
jenkins, test it

ruflin:
jenkins, retest it

PR #4088, opened 2017-04-22
File restructure by providers for add_cloud_metadata

athom:
This PR is a follow up issue for #4023 , focus on adjusting the files structure according the cloud providers. makes it easier to spot which parts are specific to a provider and which is the shared code.


libbeat/processors/add_cloud_metadata/provider_aws_ec2.go
```diff
@@ -0,0 +1,24 @@
+package actions
```
ruflin: Not related to this change, but why is this package called `actions`? @andrewkroh

andrewkroh: At one point there was an actions folder and I think we decided in the original PR for this processor to move it into its own folder. Looks like I did the file move, but never updated the actual package name.

athom: Should I make some related changes by the way?

andrewkroh: That would be great. Thanks

athom: Change is made, also applied on `add_locale` package too.
BTW, could the reviewdog be taught to recognize such convention?

ruflin: Thanks.

For reviewdog, @7AC perhaps knows more.

ruflin:
jenkins, test  it

libbeat/processors/add_cloud_metadata/provider_aws_ec2.go
7AC: Reviewdog just does what `golint` does

PR #4089, opened 2017-04-22
Ignore debug files generated by vscode

athom:
The following files are generated by Visual Studio Code editor when debugging.
It's will be nice to ignore them too. 

```
/filebeat/debug
/heartbeat/debug
/metricbeat/debug
/packetbeat/debug
/winlogbeat/debug
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,13 @@
 /*/fields.yml
 /*/*.template*.json
 
+# Debug binary files
+/filebeat/debug
+/heartbeat/debug
+/metricbeat/debug
+/packetbeat/debug
+/winlogbeat/debug
+
 # Files
 .DS_Store
 /beats.iml


ruflin:
I don't really know vscode but why does it create a directory in each beat and not one on the top level? I worry a bit excluding it in each beat as `debug` sounds like a potential package name.

athom:
vscode did not create directory in each beat folder, it just generate binary file named debug in the directories.


image

here is the debug config:

```
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "filebeat",
            "type": "go",
            "request": "launch",
            "mode": "debug",
            "remotePath": "",
            "port": 2345,
            "host": "127.0.0.1",
            "program": "${workspaceRoot}/filebeat/",
            "env": {},
            "args": ["-e"]
        },
        {
            "name": "metricbeat",
            "type": "go",
            "request": "launch",
            "mode": "debug",
            "remotePath": "",
            "port": 2345,
            "host": "127.0.0.1",
            "program": "${workspaceRoot}/metricbeat/",
            "env": {},
            "args": []
        }
    ]
}
```

ruflin:
Is that a default config? Is `debug` like the `filebeat` binary but with some additional magic inside?

athom:
hmm, not really default, but it's quite instinctive to do that way, I only changed the location of the launch directory. 
```
"program": "${workspaceRoot}/filebeat/",
```
and the args
```
 "args": ["-e"]
```

the `debug` might get some additional magic inside, I think more build flags involved? 
image

Maybe there is a way to rename the output binary in the settings.

andrewkroh:
BTW If the excludes are not common to other uses there are two ways you can ignore them. Using a [global exclude](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) or a [.git/info/exclude](https://help.github.com/articles/ignoring-files/#explicit-repository-excludes) file.

athom:
Good point! thanks for that info, I pick up a new trick 😃 
So if you don't think it is common enough in this case, I am ok to close this issue.

ruflin:
@athom I suggest to close this one. In case this pops up as an issue again in the future we should definitively discuss it again. I really appreciate that you opened a PR for this.

athom:
it's fine, thanks for the discussing and @andrewkroh 's suggestion.

PR #4090, opened 2017-04-23
Add (*MapStr).DeepUpdate

urso:
Add DeepUpdate to MapStr for recursive-deep updating an event.


ruflin:
Awesome. I was thinking about doing the exact same thing. I assume you did it for heartbeat but we will also need it in other places.

PR #4092, opened 2017-04-23
Httpmetricbeat

christiangalsterer:
Hi @ruflin,

here a very first commit.
There are no tests yet and I haven't run manual tests.

Actually I get a strange error when trying to run metricbeat and also in some tests, that the process are immediately killed, e.g. with "Killed: 9".

I'm running on Mac with golang 1.7.4 and my other beats are running fine. Are there any special settings needed for metricbeat on a mac?

Shall I open an issue for tracking/reference for the new metricbeat?


christiangalsterer:
Forget about my comment on build issues. It is solved by updating to gaoling 1.8.1 (should have checked beforehand).

ruflin:
I had a similar issue with 1.8 which is a golang/ os x issue. 1.8.1 should solve it.

metricbeat/module/http/json/json.go
```diff
@@ -0,0 +1,108 @@
...
+	}{}
+
+	if err := base.Module().UnpackConfig(&config); err != nil {
+		return nil, err
+	}
+
+	http := helper.NewHTTP(base)
+	http.SetMethod(config.Method)
+	http.SetBody([]byte(config.Body))
+	for key, value := range config.Headers {
```
vjsamuel: headers are unpacked and added to the http request here:
https://github.com/elastic/beats/blob/master/metricbeat/helper/http.go#L63

christiangalsterer: True, saw that as well, but then forgot to change before committing.

metricbeat/module/http/json/json.go
```diff
@@ -0,0 +1,108 @@
...
+	defer response.Body.Close()
+
+	responseBody, err := ioutil.ReadAll(response.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	event := common.MapStr{}
+
+	event["request"] = common.MapStr{
```
ruflin: As these are based on the configs set, not sure if we should really add the request to each document. What would be the use case here?

metricbeat/module/http/json/json.go
```diff
@@ -0,0 +1,108 @@
...
+	headers map[string]string
+	method  string
+	body    string
+}
+
+// New create a new instance of the MetricSet
+// Part of new is also setting up the configuration by processing additional
+// configuration entries if needed.
+func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
+
```
ruflin: Could you add here a log message like `logp.Beta("The http json metricset is in beta.")`.

metricbeat/module/http/json/json.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+// MetricSet type defines all fields of the MetricSet
+// As a minimum it must inherit the mb.BaseMetricSet fields, but can be extended with
+// additional entries. These variables can be used to persist data or configuration between
+// multiple fetch calls.
+type MetricSet struct {
+	mb.BaseMetricSet
+	http    *helper.HTTP
+	headers map[string]string
```
ruflin: In case we do not add these fields (see comment below) to the event, they are probably not needed in the object.

metricbeat/module/http/json/json.go
```diff
@@ -0,0 +1,108 @@
...
+		"body":    m.body,
+	}
+
+	event["response"] = common.MapStr{
+		"status_code": response.StatusCode,
+		"headers":     m.getHeaders(response.Header),
+		"body":        responseBody,
+	}
+
+	return event, nil
```
ruflin: The `json` metricset is a dynamic metricset. That means it can be used multiple times with different fields as output. To make sure these fields to not conflict, we need a namespace which can be configure. As an example see here the `jmx` metricset: https://github.com/elastic/beats/blob/master/metricbeat/module/jolokia/jmx/jmx.go#L93 So event are put into `http.namespace.*`.

metricbeat/module/http/json/json.go
```diff
@@ -0,0 +1,108 @@
...
+	event["request"] = common.MapStr{
+		"headers": m.headers,
+		"method":  m.method,
+		"body":    m.body,
+	}
+
+	event["response"] = common.MapStr{
+		"status_code": response.StatusCode,
+		"headers":     m.getHeaders(response.Header),
+		"body":        responseBody,
```
ruflin: As this is JSON, don't we need to decode it here?

metricbeat/module/http/json/json.go
```diff
@@ -0,0 +1,108 @@
...
+
+	event["request"] = common.MapStr{
+		"headers": m.headers,
+		"method":  m.method,
+		"body":    m.body,
+	}
+
+	event["response"] = common.MapStr{
+		"status_code": response.StatusCode,
+		"headers":     m.getHeaders(response.Header),
```
ruflin: As status_code and response header will be the same for all `http` metricsets, I was thinking to add these on the module lever for example under a `response` namespace. How to do that for example see docker: https://github.com/elastic/beats/blob/master/metricbeat/module/docker/cpu/data.go#L18

That would mean these are always under `http.response.*`.

I'm not sure if we should enable the headers by default because in the case of json I assume in most use cases people are not too interested in the headers (happy to be convinced otherwise). So I would make it configurable with a config option. As we already have `headers` in the config file, we could make it `response.headers.enabled`.

metricbeat/module/http/json/json.go
christiangalsterer: Yes will do.

metricbeat/module/http/json/json.go
christiangalsterer: There is definitely room for improvement to move some of the information to the module layer. Idea was to get something working and getting familiar with Metricbeat implementation.
W.r.t if we should add the header I don't see an issue here to always add them

metricbeat/module/http/json/json.go
christiangalsterer: Yes definitely.

metricbeat/module/http/json/json.go
christiangalsterer: True, thanks for the hint. Is namespace here the correct concept, or is it to use different document_types?

metricbeat/module/http/json/json.go
ruflin: Happy to merge the PR rater soonish and do improvements on top of it.

For the headers, I probably miss the use case.

metricbeat/module/http/json/json.go
ruflin: Namespace is the one to use here. We don't use document type, especially as it will be removed in 6.0

christiangalsterer: ok, will make the changes accordingly.

metricbeat/module/http/json/json.go
ruflin: One more note on the body: I would probably put the decode json directly on the top level and not into a body namespace.

christiangalsterer:
'make collect' doesn't seem to update the generated doc files (anymore). I did already a fresh clone, but no luck. This is why the "make check" currently fails ans the docs don't match anymore.

christiangalsterer:
intellij had some refresh issues causing some bad merges/rebases. I will fix that later today, sorry for any inconveniences caused by this.

ruflin:
use make update instead of make collect

christiangalsterer:
Thanks for the quick reply. If you want to to create a new PR as the current is cluttered with merge commits I'm happy to do so.

I will add more documentation later today. In any case thanks for any review comments in advance.

ruflin:
will probably not be able to review it today. you could also squash the commits and the force push to the sam branch.

christiangalsterer:
sure no problem, enjoy the long weekend. Will also squash the changes into a single commit

christiangalsterer:
I added PR #4155 for the confusion on `make collect` vs `make update`.

ruflin:
@christiangalsterer Any chance to push the squashed changes? Right now it is kind of tricky to review as the diff is huge because of all the merges. Probably you also have to rebase on master. As you merged previously I worry that you will have quite a few conflicts :-( In general I recommend to always rebase on master and not merge master.

christiangalsterer:
@ruflin: Sure thing. Will squash the changes and I always use rebase normally, but somehow IntelliJ played some bad tricks ;-)

christiangalsterer:
Actually maybe I close this PR and create a new one, to have a cleaner state, would you be ok with this?

ruflin:
sure, feel free to open a new one.

christiangalsterer:
Replaced by #4156.

PR #4095, opened 2017-04-24
Update docs.asciidoc

lisakaymera:
fsstats doesn't work anymore because of a previous merge. 
Would be helpful to change fsstats -> fsstat to avoid confusion.

### DIFF
--- a/metricbeat/module/system/fsstat/_meta/docs.asciidoc
+++ b/metricbeat/module/system/fsstat/_meta/docs.asciidoc
@@ -1,6 +1,6 @@
 === System Fsstat Metricset
 
-The System `fsstats` metricset provides overall file system statistics.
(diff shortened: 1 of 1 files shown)


tsg:
Good catch! May I ask you to sign the CLA, please? We need that even for docs changes, but it should only take a few minutes. If you don't want to bother with that, I'd be happy to re-add this change in another PR.

lisakaymera:
I clicked on the link, signed it, and confirmed the signature. Is that enough?

ruflin:
@lisakaymera Normally yes. Did you use the same username and email you use on github to sign it or a different one?

ruflin:
@lisakaymera Seems to have worked now. Thanks for the contribution.

PR #4098, opened 2017-04-24
Fix race in go-metrics adapater

urso:
Fix race in go-metrics adapter, if registry is updated concurrently from
multiple go routines.


tsg:
Can you also backport this to the 5.4 branch, please? I added the `needs_backport` label.

PR #4100, opened 2017-04-24
Cherry-pick #4098 to 5.3: Fix race in go-metrics adapater

urso:
Cherry-pick of PR #4098 to 5.3 branch. Original message: 

Fix race in go-metrics adapter, if registry is updated concurrently from
multiple go routines.


ruflin:
Merged before appveyor green as change is not windows specific and this hopefully gets into 5.3.

PR #4102, opened 2017-04-24
Adding query APIs for metricsets and modules from metricbeat registry

vjsamuel:


metricbeat/mb/registry.go
```diff
@@ -6,6 +6,7 @@ import (
 	"strings"
 
 	"github.com/elastic/beats/libbeat/logp"
+	"sync"
```
ruflin: this should be with `strings`

metricbeat/mb/registry.go
```diff
@@ -144,9 +147,42 @@ func (r *Register) metricSetFactory(module, name string) (MetricSetFactory, Host
 	return info.factory, info.hostParser, nil
 }
 
+//Modules returns the list of module names that are registered
+func (r *Register) Modules() []string {
+	r.RLock()
+	defer r.RUnlock()
+
+	modules := []string{}
```
andrewkroh: Change this to `make([]string, 0, len(r.modules))` since it knows the capacity requirement a priori.

metricbeat/mb/registry.go
```diff
@@ -144,9 +147,42 @@ func (r *Register) metricSetFactory(module, name string) (MetricSetFactory, Host
...
+
+	return modules
+}
+
+//MetricSets returns the list of metricsets registered for a given module
+func (r *Register) MetricSets(module string) []string {
+	r.RLock()
+	defer r.RUnlock()
+
+	metricsets := []string{}
```
andrewkroh: Declare this as `var metricsets []string`. https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices

metricbeat/mb/registry.go
```diff
@@ -144,9 +147,42 @@ func (r *Register) metricSetFactory(module, name string) (MetricSetFactory, Host
...
+
+	sets, ok := r.metricSets[module]
+	if ok {
+		for name := range sets {
```
andrewkroh: Once it reaches this point, it can allocate the slice with the `metricsets = make([]string, 0, len(sets)`.

metricbeat/mb/registry.go
```diff
@@ -46,6 +47,8 @@ type metricSetFactoryInfo struct {
 // Register contains the factory functions for creating new Modules and new
 // MetricSets.
 type Register struct {
+	//Lock to control concurrent read/writes
```
andrewkroh: This comment needs a space after the `//`.

metricbeat/mb/registry.go
```diff
@@ -46,6 +47,8 @@ type metricSetFactoryInfo struct {
...
+	sync.RWMutex
```
andrewkroh: I think embedding the mutex is not a good choice here because the methods will be become part of `Register`'s public API. Users should never need to lock/unlock the `Register` because that is handled internally.

This reminds me, can you add to the godoc for `Register` that "It is safe for concurrent usage."`.

andrewkroh:
jenkins, test it

PR #4105, opened 2017-04-25
Disable default prospector and adjust short configs

### LINKED ISSUE #3442: default filebeat.prospectors.paths to be commented out
Currently in filebeat, a fresh out of the box distribution has a default path of `/var/log/*.log` enabled:

```
filebeat.prospectors:

# Each - is a prospector. Most options can be set at the prospector level, so
# you can use different prospectors for various configurations.
# Below are the prospector specific configurations.

- input_type: log

  # Paths that should be crawled and fetched. Glob based paths.
  paths:
    - /var/log/*.log
    #- c:\programdata\elasticsearch\logs\*
```

While this has the benefit/intention of helping making an out-of-the-box good experience for some logs, I would say the any benefits of this are outweighed by the negatives and would instead like to advocate for paths being all commented out by default. The rationale behind this is as follows:

* Folks new to filebeat that want to try it out and start indexing data are likely to have tons of logs from their laptop indexed into Elasticsearch (hopefully their local instance only); the affect is greater if the user starts filebeat as root.
* If a user is required to pick a location to read logs from, it helps make sure they actively think about where to read logs from

tsg:
This does two changes:

* Adds `enabled: false` in the default prospector in the short and long configs.
  (which fixes #3442)
* Updates the short config files of the modules to include the path definitions.
  I think this is better for a "module first" experience, where it gives a bit of
  context on how they work.

This means that the default configuration starts and gives no errors, but also doesn't
publish anything. IMO, this behaviour is fine considering configuration reloading.

### DIFF
--- a/filebeat/_meta/common.full.p2.yml
+++ b/filebeat/_meta/common.full.p2.yml
@@ -16,6 +16,9 @@ filebeat.prospectors:
 #------------------------------ Log prospector --------------------------------
 - input_type: log
 
+  # Change to true to enable this prospector configuration.
+  enabled: false
+
   # Paths that should be crawled and fetched. Glob based paths.
   # To fetch all ".log" files from a specific level of subdirectories
   # /var/log/*/*.log can be used.
--- a/filebeat/_meta/common.p2.yml
+++ b/filebeat/_meta/common.p2.yml
@@ -11,6 +11,9 @@ filebeat.prospectors:
 
 - input_type: log
 
+  # Change to true to enable this prospector configuration.
+  enabled: false
+
   # Paths that should be crawled and fetched. Glob based paths.
   paths:
     - /var/log/*.log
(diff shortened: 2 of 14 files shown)


tsg:
We could perhaps check if no prosectors are defined && configuration reloading is disabled, and print an error in that case. But that seems like a nice to have and we don't currently do that.

ruflin:
Good point about the reloading. We actually do this in metricbeat but not in filebeat.

tsg:
I pushed another commit that prints an error if all modules & prospectors are disable and config reloading is disabled (the code was there, was just not looking at the enabled flag).

[This](https://github.com/tsg/beats/blob/41900fd0f5a65a7114338894c2af1a998d75e2d1/filebeat/beater/filebeat.go#L69) is the error message. I think it's clear enough for now, later we can make it link to the docs or a FAQ, but for the moment I worry about adding other suggestions to it.

dedemorton:
@tsg Shouldn't the filebeat prospector reference doc also be updated to show the `enabled` option?

ruflin:
@dedemorton Yes, I probably missed that when I introduced Enabled :-(

PR #4107, opened 2017-04-25
Unify shutdown timeout

### LINKED ISSUE #3588: Unify -waitstop and shutdown_timeout
Filebeat has a `shutdown_timeout` config option and packetbeat a `-waitstop` option. These two should be unified into 1 option.

glefloch:
This pull request aims to unify shutdown timeout. 
For now, this is must be implemented in each beat `Run` method due to specific code in `filebeat`. 
 
Close #3588

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -105,6 +105,7 @@ https://github.com/elastic/beats/compare/v5.4.0...v6.0.0-alpha1[View commits]
 *Packetbeat*
 
 - Remove deprecated geoip. {pull}3766[3766]
+- Replace `waitstop` command line argument by `shutdown_timeout` in configuration file. {pull}3588[3588]
 
 *Winlogbeat*
 
--- a/packetbeat/beater/packetbeat.go
+++ b/packetbeat/beater/packetbeat.go
@@ -40,24 +40,22 @@ type packetbeat struct {
 }
 
 type flags struct {
-	file         *string
-	loop         *int
-	oneAtAtime   *bool
-	topSpeed     *bool
-	dumpfile     *string
-	waitShutdown *int
+	file       *string
+	loop       *int
+	oneAtAtime *bool
+	topSpeed   *bool
+	dumpfile   *string
 }
 
 var cmdLineArgs flags
 
 func init() {
 	cmdLineArgs = flags{
-		file:         flag.String("I", "", "Read packet data from specified file"),
-		loop:         flag.Int("l", 1, "Loop file. 0 - loop forever"),
-		oneAtAtime:   flag.Bool("O", false, "Read packets one at a time (press Enter)"),
-		topSpeed:     flag.Bool("t", false, "Read packets as fast as possible, without sleeping"),
-		dumpfile:     flag.String("dump", "", "Write all captured packets to this libpcap file"),
-		waitShutdown: flag.Int("waitstop", 0, "Additional seconds to wait before shutting down"),
+		file:       flag.String("I", "", "Read packet data from specified file"),
+		loop:       flag.Int("l", 1, "Loop file. 0 - loop forever"),
+		oneAtAtime: flag.Bool("O", false, "Read packets one at a time (press Enter)"),
+		topSpeed:   flag.Bool("t", false, "Read packets as fast as possible, without sleeping"),
+		dumpfile:   flag.String("dump", "", "Write all captured packets to this libpcap file"),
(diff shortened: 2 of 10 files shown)


glefloch:
@ruflin can you review this pull request? Sorry for all those commits, I got some trouble running tests on my computer.

packetbeat/beater/packetbeat.go
```diff
@@ -172,9 +170,9 @@ func (pb *packetbeat) Run(b *beat.Beat) error {
 		service.Stop()
 	}
 
-	waitShutdown := pb.cmdLineArgs.waitShutdown
-	if waitShutdown != nil && *waitShutdown > 0 {
-		time.Sleep(time.Duration(*waitShutdown) * time.Second)
+	timeout := b.Config.ShutdownTimeout
```
ruflin: The problem is that packetbeat does not have the same logic here as filebeat. On the filebeat side it waits for all events to be published and then stops. If all events are sent before `shutdown_timeout` is reached, it will directly shut down. Because this tracking if all events are sent does not exist in packetbeat, packetbeat just "sleeps" for the time. 

The reason I worry about this it because there are two different behaviours with 1 config option. As this problem can't be covered in this PR I think we should solve it through documentation.

Also this code should probably not be in packetbeat anymore, but on the beats side instead. Otherwise each beat has to implement it.

packetbeat/docs/reference/configuration/packetbeat-options.asciidoc
```diff
@@ -832,6 +832,22 @@ periodically afterwards, it scans the process table for
 processes that match the values specified for this option. The match is done against the
 process' command line as read from `/proc/<pid>/cmdline`.
 
+
+[[shutdown-timeout]]
+===== shutdown_timeout
```
ruflin: Seems like this should be now documented in libbeat and not packetbeat.

packetbeat/beater/packetbeat.go
glefloch: You right, I move it back to `packetbeat`. This will make it more clearer as it's not handled in `libbeat` yet.

packetbeat/docs/reference/configuration/packetbeat-options.asciidoc
```diff
@@ -832,6 +832,22 @@ periodically afterwards, it scans the process table for
...
+
+How long Packetbeat waits on shutdown. By default, this option is disabled, and Packetbeat
+does not wait for the publisher to finish sending events before shutting down
```
ruflin: We should add here that packetbeat is just going to wait for `shutdown_timeout` and the closes and will not track if all events were sent previously.

glefloch: I rephrased the documentation as you commented it :)

ruflin:
@glefloch I think you have to rebase again on master as there are now lots of changes in this PR which shouldn't be. Could you also squash the commits into 1 commit?

glefloch:
@ruflin Sorry for all that mess. I rebase master in my branch

packetbeat/docs/reference/configuration/packetbeat-options.asciidoc
```diff
@@ -832,6 +832,23 @@ periodically afterwards, it scans the process table for
 processes that match the values specified for this option. The match is done against the
 process' command line as read from `/proc/<pid>/cmdline`.
 
+
+[[shutdown-timeout]]
+===== shutdown_timeout
+
+How long Packetbeat waits on shutdown. By default, this option is disabled. Packetbeat 
+will wait for `shutdown_timeout` and then close. It will not track if all events were send
```
ruflin: s/send/sent/

CHANGELOG.asciidoc
```diff
@@ -179,6 +179,7 @@ https://github.com/elastic/beats/compare/v5.4.0...v6.0.0-alpha1[View commits]
 
 - Add `fields` and `fields_under_root` to Packetbeat protocols configurations. {pull}3518[3518]
 - Add list style Packetbeat protocols configurations. This change supports specifying multiple configurations of the same protocol analyzer. {pull}3518[3518]
+- Replace `waitstop` command line argument by `shutdown_timeout` in configuration file. {pull}3588[3588]
```
ruflin: This should go under breaking changes.

ruflin:
@glefloch Thanks a lot for the contribution.

PR #4109, opened 2017-04-25
Suggest GOPATH in ~/go

7AC:


ruflin:
@7AC I think you have to rebase that one on master as it contains lots of other commits.

7AC:
My bad, didn't switch to master first. Moved it here https://github.com/elastic/beats/pull/4110

PR #4111, opened 2017-04-25
Support Alibaba Cloud provider for add_cloud_metadata proccessor

athom:
Follow PR of #4023 
Aims to support Alibaba Cloud provider  for add_cloud_metadata processor.
Similar with Tencent Cloud, Alibaba's metadata services requires multiple requests too.
Here is the official [document](https://help.aliyun.com/knowledge_detail/49122.html).
Tested on an esc instance with os ubuntu 16.04. 
Note, Alibaba cloud's metadata services works only when the instance's network type is VPC.

### DIFF
--- a/libbeat/docs/processors-config.asciidoc
+++ b/libbeat/docs/processors-config.asciidoc
@@ -273,6 +273,7 @@ The following cloud providers are supported:
 - Digital Ocean
 - Google Compute Engine (GCE)
 - https://www.qcloud.com/?lang=en[Tencent Cloud] (QCloud)
+- Alibaba Cloud (ECS)
 
 The simple configuration below enables the processor.
 
@@ -354,6 +355,22 @@ _Tencent Cloud_
       "instance_id": "ins-qcloudv5",
       "provider": "qcloud",
       "region": "china-south-gz"
+    }
+  }
+}
+-------------------------------------------------------------------------------
+
+_Alibaba Cloud_
+
+[source,json]
+-------------------------------------------------------------------------------
+{
+  "meta": {
+    "cloud": {
+      "availability_zone": "cn-shenzhen",
+      "instance_id": "i-wz9g2hqiikg0aliyun2b",
+      "provider": "ecs",
+      "region": "cn-shenzhen-a"
     }
   }
 }
--- a/libbeat/docs/processors-config.asciidoc
+++ b/libbeat/docs/processors-config.asciidoc
@@ -362,6 +362,8 @@ _Tencent Cloud_
 
 _Alibaba Cloud_
 
+This metadata is only available when VPC is selected as the network type of the ECS instance.
+
 [source,json]
 -------------------------------------------------------------------------------
 {
--- a/libbeat/processors/add_cloud_metadata/provider_alibaba_cloud.go
+++ b/libbeat/processors/add_cloud_metadata/provider_alibaba_cloud.go
@@ -1,0 +1,41 @@
+package add_cloud_metadata
+
+import "github.com/elastic/beats/libbeat/common"
+
+// Alibaba Cloud Metadata Service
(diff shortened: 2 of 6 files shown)


libbeat/processors/add_cloud_metadata/provider_alibaba_cloud.go
```diff
@@ -0,0 +1,41 @@
+package add_cloud_metadata
+
+import "github.com/elastic/beats/libbeat/common"
+
+// Alibaba Cloud Metadata Service
+// Document https://help.aliyun.com/knowledge_detail/49122.html
+func newAlibabaCloudMetadataFetcher(c common.Config) (*metadataFetcher, error) {
+	ecsMetadataHost := "100.100.100.200"
```
ruflin: It's interesting that this is a fixed IP and not a domain name. Couldn't make much sense of the docs above unfortunately ;-)

libbeat/docs/processors-config.asciidoc
```diff
@@ -359,6 +360,22 @@ _Tencent Cloud_
 }
 -------------------------------------------------------------------------------
 
+_Alibaba Cloud_
+
```
andrewkroh: I realize that this is supposed to just be examples of the output, but it would be good to mention something like, "Only when VPC is selected as the network type of the instance, you can get the metadata."

@dedemorton Maybe you have wording or formatting suggestions.

libbeat/processors/add_cloud_metadata/provider_tencent_cloud.go
```diff
@@ -3,6 +3,7 @@ package add_cloud_metadata
 import "github.com/elastic/beats/libbeat/common"
 
 // Tenccent Cloud Metadata Service
```
andrewkroh: `s/Tenccent/Tencent/`?

libbeat/processors/add_cloud_metadata/provider_alibaba_cloud.go
athom: Yyep, maybe metadata is not a really important feature for Chinese users, the design looks not so serious in both Tencent and Alibaba provider :-)

> Couldn't make much sense of the docs above unfortunately ;-)
You mean remove the document comment?

libbeat/processors/add_cloud_metadata/provider_tencent_cloud.go
athom: 👍 good catch!

andrewkroh:
jenkins, test it

libbeat/docs/processors-config.asciidoc
dedemorton: Let's change the wording because we don't want to take the text verbatim from the Alibaba docs. How about adding this before the example:

`This metadata is only available when VPC is selected as the network type of the instance.`

Or, assuming we are talking about an ECS instance here (apologies for my ignorance), this would be even better and less confusing for users:

`This metadata is only available when VPC is selected as the network type of the ECS instance.`

libbeat/processors/add_cloud_metadata/provider_alibaba_cloud.go
andrewkroh: The comment is fine.

libbeat/docs/processors-config.asciidoc
athom: Added on the position above the example output.

ruflin:
jenkins, test it

athom:
Did my changes affect symlink rotate feature? 
Looks an unrelated test case block here twice?  
I don't have any idea to fix it for now, any thoughts?

ruflin:
@athom no, change is not related to the CI failure.

PR #3981, opened 2017-04-10
Add example dashboards using the time series visual builder

monicasarbu:
Add example dashboards that are using the visualizations built with the time series visual builder (available in Kibana 5.4).

In order for the dashboards to work properly, we need to have [this](https://github.com/elastic/beats/pull/3979) merged that adds support for url fields.

PR #4115, opened 2017-04-26
Add example dashboards using the time series visual builder

monicasarbu:
Replace https://github.com/elastic/beats/pull/3981. 

Several improvements are done to the Metricbeat system dashboards:
- [x] change navigation to be on top of the page in order to occupy less space
- [x] add two dashboards that are using the widgets built with the new time series visual builder 
- [x] re-arrange the old dashboards to accommodate the new navigation bar 

Here are a few screenshots with the result:

screen shot 2017-04-26 at 1 46 59 pm

screen shot 2017-04-26 at 1 47 13 pm

cc-ed: @simianhacker


andrewkroh:
Why is there a "Top 10 hosts by CPU" table on the host overview dashboard? Maybe this should be processes instead.

monicasarbu:
@andrewkroh That's a good question. The second screenshot is a dashboard that shows system details per host. By default, it shows the details for all the hosts, but you can select a host by clicking on one host from the "top 10 hosts" widget. This way, you can easier have a per host view. This is a temporary workaround until we have links between dashboards.

andrewkroh:
Ok, makes sense. I use a tag cloud of hostnames for a similar purpose.

andrewkroh:
With the "alpha2" tag here, do you want to hold off the merge until after alpha1 is released?

PR #4117, opened 2017-04-26
Remove duplicate code in glob_watcher_test.go

athom:
Looks like a copy-paste mistake? @ruflin


ruflin:
TBH I'm not sure anymore. I remember there are / were some edge cases for some OS that needed some special handling like the wait we have here: https://github.com/elastic/beats/pull/4117/files#diff-31168770a0fc57b03e6d6681a3c45747L32 Lets see if the CI goes green. If not, we know it's on purpose. But if it's on purpose I would hopefully have left a comment.

ruflin:
@athom Merged as the build went green.

PR #4120, opened 2017-04-26
Fix link to the MacOSX SDK tarball

tsg:
The original download was temporarily down and then it came back up
with a different sha1. Switching to what seems to be a link closer to the
source.

This will require backporting in all branches that need to be built.


tsg:
jenkins, package it

andrewkroh:
jenkins, package it

tsg:
Packaging went green: http://build-eu-00.elastic.co/job/beats-package-PR/95/

PR #4122, opened 2017-04-26
Cherry-pick #4120 to 5.x: Fix link to the MacOSX SDK tarball

tsg:
Cherry-pick of PR #4120 to 5.x branch. Original message: 

The original download was temporarily down and then it came back up
with a different sha1. Switching to what seems to be a link closer to the
source.

This will require backporting in all branches that need to be built.


tsg:
jenkins, package it

Issue #4112, opened 2017-04-25
Metricbeat: reduce disk usage in default configuration


tsg:
Meta ticket for several improvements we plan for Metricbeat in order to reduce the storage space requirements in its default configuration.

* [x] #4127 Add a feature to allow collecting only the first top N processes by CPU and/or memory, instead of all of them.
* [x] #4231 Add new metrics for the total process counts and use those in dashboards
* [x] #4284 Make it possible to set "shards" and "replicas" from the metricbeat YAML file.
* [x] #4329 Set the default number of shards to 1.
* [x] #4317 Make it easily possible to disable `_source`, but leave it on by default.
* [x] #4329 Enable best compression by default.
* [x] #4329 Reduce the collection interval for filesystem statistics and disable `fsstat` by default. Set it to 1m.
* [x] #4329 Filter out several filesystem types by mount point (e.g. proc, tmpfs, dev)
* [ ] Filter out logical devices from the `diskio` stats

We target these changes for the 6.0 release.

tsg:
This can be considered closed.

PR #4127, opened 2017-04-27
Metricbeat: store only top N processes by CPU/memory

### LINKED ISSUE #4126: Metricbeat: store only top N processes by CPU / memory
Part of #4112, we want to add a feature where only top N by CPU and/or memory are included in the reports created by the Metricbeat system module. Optionally, for the processes that drop out of top N, we could reduce their polling interval instead of just completely dropping them.

My current plan is to implement this as a feature of the `system.process` metricset. I considered doing this as a processor but a processor would need to collect all processes from an interval in order to sort & compute the "top". This seems to me like a lot of state to put in a processor, plus it's hard for a processor to know when the list is "done".

Configuration wise, I'm thinking:

```
  process.include_top:
    enabled: true
    cpu.total.pct: 5         # include top 5 by CPU
    memory.rss.bytes: 5      # include top 5 by memory
```

Explanations:

 * `include_top` because we use `include` to mean "filter everything but these" in other places in Beats
* `enabled: true` to have an easy and clear way to opt in / opt out of this feature.
* `cpu.total.pct: 5` means "record top 5 processes by the `cpu.total.pct` field. 
* `memory.rss.bytes: 5` means "record top 5 processes by `memory.rss.bytes` field.
* If any of `cpu.total` or `memory.rss` are set to 0, it means "match no processes". So `cpu.total: 5 memory.rss: 0` will only look at CPU.

Do  we want to support sorting by other fields than the two? Make it generic? That would require a more generic (and more CPU intensive) implementation.

In the above, events for processes below the threshold are dropped. If the user wants to store them, just at a reduced resolution, the following config could be added:

```
  process.include_top:
    enabled: true
    cpu.total.pct: 5
    memory.rss.bytes: 5
    period_multiplier: 3    # downsample for processes not in top
```

Here, `period_multiplier: 3` means that we will only publish one in 3 events for the processes out of the top. Since we report counters and not rates, that should work fine.

tsg:
This adds the option to only report on the top N processes by CPU and/or memory. It is useful because storing metrics about each and every process from every host can be fairly expensive from the storage point of view. Previously it was possible to filter processes by name, which was useful if one knew in advance which are the most interesting processes. This adds a new option which should be quite convenient in practice, because the number of per-process documents gets limited while still allowing to display the top processes.

Closes #4126. 

Configuration wise it looks like this:

```
  process.include_top_n:
```

Remaining TODOs:

* [x] unit tests
* [x] document the new settings

### DIFF
--- a/metricbeat/docs/modules/system.asciidoc
+++ b/metricbeat/docs/modules/system.asciidoc
@@ -26,6 +26,25 @@ metricbeat.modules:
   metricsets: ["process"]
   processes: ['.*']
 ----
+
+*`process.include_top_n`*:: These options allow you to filter out all processes
+that are not in the top N by CPU or memory, in order to reduce the number of
+documents created. If both the `by_cpu` and `by_memory` options are used, the
+reunion of the two tops is included.
+
+*`process.include_top_n.enabled`*:: Set to false to disable the top N feature and
+include all processes, regardless of the other options. The default is `true`,
+but nothing is filtered unless one of the other options (`by_cpu` or `by_memory`)
+is set to a non-zero value.
+
+*`process.include_top_n.by_cpu`*::  How many processes to include from the top
+by CPU. The processes are sorted by the `system.process.cpu.total.pct` field.
+The default is 0.
+
+*`process.include_top_n.by_memory`*:: How many processes to include from the top
+by memory. The processes are sorted by the `system.process.memory.rss.bytes`
+field. The default is 0.
+
 *`process.cgroups.enabled`*:: When the `process` metricset is enabled, you can
 use this boolean configuration option to disable cgroup metrics. By default
 cgroup metrics collection is enabled.
@@ -100,6 +119,9 @@ metricbeat.modules:
   enabled: true
   period: 10s
   processes: ['.*']
+  process.include_top_n:
+    by_cpu: 5      # include top 5 processes by CPU
+    by_memory: 5   # include top 5 processes by memory
 ----
 
 [float]
--- a/metricbeat/module/system/_meta/config.full.yml
+++ b/metricbeat/module/system/_meta/config.full.yml
@@ -36,6 +36,23 @@
   # if true, exports the CPU usage in ticks, together with the percentage values
   #cpu_ticks: false
 
+  # These options allow you to filter out all processes that are not
+  # in the top N by CPU or memory, in order to reduce the number of documents created.
+  # If both the `by_cpu` and `by_memory` options are used, the reunion of the two tops
+  # is included.
+  #process.include_top_n:
+    #
+    # Set to false to disable this feature and include all processes
+    #enabled: true
+
+    # How many processes to include from the top by CPU. The processes are sorted
+    # by the `system.process.cpu.total.pct` field.
+    #by_cpu: 0
+
+    # How many processes to include from the top by memory. The processes are sorted
+    # by the `system.process.memory.rss.bytes` field.
+    #by_memory: 0
+
   # If false, cmdline of a process is not cached.
   #process.cmdline.cache.enabled: true
 
--- a/metricbeat/module/system/process/helper.go
+++ b/metricbeat/module/system/process/helper.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"runtime"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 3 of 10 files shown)


metricbeat/module/system/process/helper.go
```diff
@@ -216,7 +221,7 @@ func getProcState(b byte) string {
 	return "unknown"
 }
 
-func (procStats *ProcStats) GetProcessEvent(process *Process, last *Process) common.MapStr {
+func (procStats *ProcStats) GetProcessEvent(process *Process) common.MapStr {
```
tsg: I unexported this one, i don't think it needs to be exported.

metricbeat/module/system/_meta/config.full.yml
```diff
@@ -36,6 +36,21 @@
   # if true, exports the CPU usage in ticks, together with the percentage values
   #cpu_ticks: false
 
+  # These options allow you to filter out all processes that are not
```
ruflin: Perhaps add a note that it is a union?

tsg:
As a follow up, I added in #4112 a point to add new metrics for the "total number of processes" and similar, and update the dashboards to use these metrics instead of aggregating them based on the row process values.

metricbeat/module/system/_meta/config.full.yml
tsg: I added a note at the section start.

tsg:
Added CHANGELOG entry, cleaned up and rebased to master.

metricbeat/module/system/process/helper_test.go
```diff
@@ -6,11 +6,13 @@ package process
 import (
 	"os"
 	"runtime"
+	"sort"
 	"testing"
 	"time"
 
 	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/gosigar"
+	sigar "github.com/elastic/gosigar"
```
ruflin: it's tests but it seems like gosigar is imported twicw?

metricbeat/docs/modules/system.asciidoc
```diff
@@ -26,6 +26,25 @@ metricbeat.modules:
   metricsets: ["process"]
   processes: ['.*']
 ----
+
+*`process.include_top_n`*:: These options allow you to filter out all processes
+that are not in the top N by CPU or memory, in order to reduce the number of
+documents created. If both the `by_cpu` and `by_memory` options are used, the
+reunion of the two tops is included.
```
andrewkroh: `s/reunion of the two tops/union of the two sets/`?

metricbeat/module/system/process/process.go
```diff
@@ -32,17 +32,31 @@ type MetricSet struct {
 	cacheCmdLine bool
 }
 
+// includeTopConfig is the configuration for the "top N processes
+// filtering" feature
+type includeTopConfig struct {
```
andrewkroh: For some reason I thought the struct needed to be exported in order for go-ucfg to work on it via reflection. Researching further, I guess [since Go 1.6](https://golang.org/doc/go1.6#reflect) this isn't necessary.

metricbeat/module/system/process/helper.go
```diff
@@ -336,7 +339,7 @@ func (procStats *ProcStats) GetProcStats() ([]common.MapStr, error) {
 		return nil, err
 	}
 
-	processes := []common.MapStr{}
+	processes := []Process{}
```
andrewkroh: It's best to declare the empty slice as `var processes []Process` because it avoids the allocation if the slice goes unused. https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices

metricbeat/module/system/process/helper.go
```diff
@@ -363,16 +366,64 @@ func (procStats *ProcStats) GetProcStats() ([]common.MapStr, error) {
...
+			process.cpuTotalPct = GetProcCpuPercentage(last, process)
+			processes = append(processes, *process)
+		}
+	}
+	procStats.ProcsMap = newProcs
+
+	processes = procStats.includeTopProcesses(processes)
+	logp.Debug("processes", "Filtered top processes down to %d processes", len(processes))
+
+	procs := []common.MapStr{}
```
andrewkroh: We know the size a priori, so this could do the full allocation at once with `procs := make([]common.MapStr, 0, len(processes))`.

metricbeat/module/system/process/helper.go
```diff
@@ -363,16 +366,64 @@ func (procStats *ProcStats) GetProcStats() ([]common.MapStr, error) {
...
-			processes = append(processes, proc)
+func (procStats *ProcStats) includeTopProcesses(processes []Process) []Process {
+
+	if !procStats.IncludeTop.Enabled ||
+		(procStats.IncludeTop.ByCPU == 0 && procStats.IncludeTop.ByMemory == 0) {
+
+		return processes
+	}
+
+	result := []Process{}
```
andrewkroh: Same issue with the empty slice declaration.

tsg:
@andrewkroh thanks, I addressed your comments in https://github.com/elastic/beats/pull/4173

PR #4128, opened 2017-04-27
Fix issue that new prospector was not reloaded on conflict

### LINKED ISSUE #4133: Filebeat: Live Reload incorrectly stops prospectors
Bug around stopping prospectors confirmed here: https://discuss.elastic.co/t/error-when-reloading-prospectors-for-filebeat/83082

- Version: 5.3
- Operating System: CentOS6.9
- Steps to Reproduce: 
1.) Start a filebeat with a prospector config in the live-reload dir. 
2.) Add another prospector in addition to the existing one in the live-reload dir (does not matter if its in the same file or not)
3.) The first prospector will stop running once the reload occurs.

ruflin:
A new prospector cannot be started if one of the states he starts with is not set to Finished. In this case the prospector should wait until the old prospector is shutdown and try again on the next reload. This did not work as intended so far because if loading failed, the running prospector was stopped and the non running was not started again on the next run. This change should fix it.

See also https://discuss.elastic.co/t/error-when-reloading-prospectors-for-filebeat/83082/5

Closes #4133

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -70,6 +70,7 @@ https://github.com/elastic/beats/compare/v5.1.1...master[Check the HEAD diff]
 - Fix panic in JSON decoding code if the input line is "null". {pull}4042[4042]
 - Properly shut down crawler in case one prospector is misconfigured. {pull}4037[4037]
 - Fix the Mysql slowlog parsing of IP addresses. {pull}4183[4183]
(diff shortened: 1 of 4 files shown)


ruflin:
See https://github.com/elastic/beats/issues/4133

tsg:
jenkins, retest it

tsg:
WFG.

PR #4134, opened 2017-04-28
Add community beats topic to devguide

dedemorton:
As Monica suggested, I've added the community beats page to the devguide. I also moved up the content that was previously at the bottom of the page and streamlined it a bit. Now that the list of beats is long (YAY!), that info was getting lost at the bottom of the page.

Needs backport to 5.4.


dedemorton:
Actually...this should not be backported because we decided that we won't add this to 5.x.

PR #4135, opened 2017-04-28
Clarify doc for Windows commands to indicate that Powershell is required

dedemorton:
Because sometimes the things that seem obvious to us are not so obvious to users....


dedemorton:
Closes issue #4135

PR #4141, opened 2017-04-28
Add Fields and Tag to monitor config

### LINKED ISSUE #3623: Heartbeat check custom fields
It would be beneficial to support custom fields for individual checks in Heartbeat. For instance, it would be good to be able to have an environment field for the check. Right now we have to stuff this data in the name field which is not convenient for aggregation purposes.

glefloch:
Add Tags and fields attribute to monitor configuration. 

Close #3623

### DIFF
--- a/heartbeat/beater/manager.go
+++ b/heartbeat/beater/manager.go
@@ -32,15 +32,22 @@ type Monitor struct {
 	factory    monitors.Factory
 	config     *common.Config
 
-	active map[string]MonitorTask
+	active map[string]monitorTask
 }
 
-type MonitorTask struct {
-	name, typ string
+type monitorTask struct {
+	job    monitors.Job
+	cancel JobCanceller
 
-	job      monitors.Job
-	schedule scheduler.Schedule
-	cancel   JobCanceller
+	config monitorTaskConfig
+}
+
+type monitorTaskConfig struct {
+	Name     string             `config:"name"`
+	Type     string             `config:"type"`
+	Schedule *schedule.Schedule `config:"schedule" validate:"required"`
+	// Fields and tags to add to monitor.
+	EventMetadata common.EventMetadata `config:",inline"`
 }
 
 type JobControl interface {
@@ -109,7 +116,7 @@ func newMonitorManager(
 			name:    info.Name,
 			factory: factory,
 			config:  config,
-			active:  map[string]MonitorTask{},
+			active:  map[string]monitorTask{},
 		})
 	}
 
@@ -146,7 +153,7 @@ func newMonitorManager(
 }
 
 func (m *Monitor) Update(configs []*common.Config) error {
-	all := map[string]MonitorTask{}
+	all := map[string]monitorTask{}
 	for i, upd := range configs {
 		config, err := common.MergeConfigs(m.config, upd)
 		if err != nil {
@@ -154,11 +161,7 @@ func (m *Monitor) Update(configs []*common.Config) error {
 			return err
 		}
 
-		shared := struct {
-			Name     string             `config:"name"`
-			Type     string             `config:"type"`
-			Schedule *schedule.Schedule `config:"schedule" validate:"required"`
-		}{}
+		shared := monitorTaskConfig{}
 		if err := config.Unpack(&shared); err != nil {
 			logp.Err("Failed parsing job schedule: ", err)
 			return err
@@ -170,17 +173,10 @@ func (m *Monitor) Update(configs []*common.Config) error {
 			return err
 		}
 
-		name := shared.Name
-		if name == "" {
-			name = shared.Type
-		}
-
 		for _, job := range jobs {
-			all[job.Name()] = MonitorTask{
-				name:     name,
-				typ:      shared.Type,
-				job:      job,
-				schedule: shared.Schedule,
+			all[job.Name()] = monitorTask{
+				job:    job,
+				config: shared,
 			}
 		}
 	}
@@ -189,12 +185,12 @@ func (m *Monitor) Update(configs []*common.Config) error {
 	for _, job := range m.active {
 		job.cancel()
 	}
-	m.active = map[string]MonitorTask{}
+	m.active = map[string]monitorTask{}
 
 	// start new and reconfigured tasks
 	for id, t := range all {
-		job := createJob(m.manager.client, t.job, t.name, t.typ)
-		t.cancel = m.manager.jobControl.Add(t.schedule, id, job)
+		job := createJob(m.manager.client, t)
+		t.cancel = m.manager.jobControl.Add(t.config.Schedule, id, job)
 		m.active[id] = t
 	}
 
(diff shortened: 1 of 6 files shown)


ruflin:
@glefloch Haven't checked the code yet, but could you add a changelog entry?

glefloch:
Oh yes for sure, I'm doing it right now

heartbeat/heartbeat.full.yml
```diff
@@ -46,6 +46,22 @@ heartbeat.monitors:
   # Waiting duration until another ICMP Echo Request is emitted.
   wait: 1s
 
+  # The tags of the monitors are included in their own field with each
```
ruflin: This must be added to `heartbeat/_meta/beat.full.yml` and then you have to run `make update`.

heartbeat/heartbeat.full.yml
```diff
@@ -119,6 +135,22 @@ heartbeat.monitors:
     # Required TLS protocols
     #supported_protocols: ["TLSv1.0", "TLSv1.1", "TLSv1.2"]
 
+  # The tags of the monitors are included in their own field with each
```
ruflin: I suggest we add it only to one monitor (first one) as an example and not every single one.

heartbeat/heartbeat.full.yml
```diff
@@ -189,6 +221,22 @@ heartbeat.monitors:
     # Required response contents.
     #body:
 
+  # The tags of the monitors are included in their own field with each
```
ruflin: See comment above.

heartbeat/monitors/active/icmp/icmp.go
```diff
@@ -71,6 +71,12 @@ func create(
 		}
 
 		settings := monitors.MakeHostJobSettings(jobName, host, config.Mode)
+		metadata, e := monitors.AddMetadata(config.EventMetadata)
```
ruflin: Can you call it `err` instead of `e` to be consistent with most of the other code?

heartbeat/heartbeat.full.yml
glefloch: Ok, I'm reverting those modifications.

heartbeat/monitors/active/icmp/icmp.go
glefloch: Yes I will update the name.

glefloch:
@ruflin, I was expecting the same thing before starting working on this issue. But there I didn't found any common configuration object between monitors, that's why I added the `metadata` object in each monitor configuration. 
Have you any ideas on how I could make this more 'common'?

ruflin:
I wast thinking perhaps somewhere around `MakeJob`? https://github.com/elastic/beats/blob/master/heartbeat/monitors/util.go#L92 Or probably even better around `createJobTask`? https://github.com/elastic/beats/blob/master/heartbeat/beater/manager.go#L241 The tricky part is how to get the config there. @urso would know more but is offline this week.

BTW: This should also be added to the docs under the common config options: https://github.com/elastic/beats/blob/master/heartbeat/docs/reference/configuration/heartbeat-options.asciidoc

urso:
The common configurations for monitors are currently 'implicit' in code (no special type), as originally I have had only one field to load from: [See here](https://github.com/elastic/beats/blob/master/heartbeat/beater/manager.go#L157)

With us adding more fields, this should become it's own type.

All settings are added to [MonitorTask](https://github.com/elastic/beats/blob/master/heartbeat/beater/manager.go#L179).

Not sure, but I think `MonitorTask` can become a package private type + we can make it act more like the final job factory by moving `createJobTask` to `(*MonitorTask).PrepareSchedulerJob(client publisher.Client) scheduler.TaskFunc`. Then, right before [PublishEvent](https://github.com/elastic/beats/blob/master/heartbeat/beater/manager.go#L259) we can apply the EventMetaData.

I didn't touch any of this code yet, as we were planning to redo the reloading functionality, to resemble filebeat/metricbeat conf.d support.

glefloch:
@urso Thanks for those explanation. I'm going to do what you said and I will push that up.

glefloch:
I finally got time to work on this. Thanks @urso, your solution is way muck cleaner.

tsg:
@glefloch For the `make check` failure, you need to run `goimports` on the source code. Thanks a lot for contributing!

glefloch:
@tsg  thanks, I ran make check and fixed code format.

heartbeat/beater/manager.go
```diff
@@ -234,28 +230,44 @@ func createWatchUpdater(monitor *Monitor) func(content []byte) {
...
 				},
 			})
 
 			if _, exists := event["type"]; !exists {
 				event["type"] = defaultEventType
 			}
+			err := common.MergeFields(event, m.config.EventMetadata.Fields, m.config.EventMetadata.FieldsUnderRoot)
+			if err != nil {
+				logp.Err("Failed to merge fields: %v", err)
+				return nil
```
urso: we really want to stop the job by returning here? How 'fatal' is an error when adding fields? e.g. we still want to try to report the original event if this op failed? `MergeFields` only errors if some field can not be spliced into, because some supposed to be intermediate object is an actual leaf/primitive value.

Normall If `len(next) > 0`, more followup tasks will be executed to finish the final job. We want to kill the job on 'naming' conflicts in MergeFields?

glefloch:
Thanks for the review @urso, I'm going to fix your comments and check the test as well.

heartbeat/beater/manager.go
```diff
@@ -234,28 +230,44 @@ func createWatchUpdater(monitor *Monitor) func(content []byte) {
...
-		event, next, err := r.Run()
+		event, next, err := m.job.Run()
 		if err != nil {
 			logp.Err("Job %v failed with: ", err)
 		}
 
+		name := m.config.Name
+		if name == "" {
+			name = m.config.Name
+		}
```
glefloch: You are right, I move it on top of the return

heartbeat/beater/manager.go
glefloch: Well, you right, should we also ignore errors on while add tags?

tsg:
@glefloch Travis complains about the latest version. Is that test passing locally for you?

glefloch:
@tsg actually, when I run `INTEGRATION_TESTS= TESTING_ENVIRONMENT=latest make -C heartbeat testsuite` everything looks good but no python tests are ran.. So I'm not able to fix the tests... I'm looking for a fix on my machine.

glefloch:
@tsg, heartbeat testsuite is ok, but filebeat testsuite got broken because of a timeout with elasticsearch... There is also a weird error with the heartbeat testsuite using `TEST_ENVIRONMENT=0` it seems that something did not answer within 10min on testsuite initialization. 
Is it possible to restart test?

glefloch:
@tsg, I rebase master and all check are ok. Can you check this PR again?

urso:
Merged. Thanks for you patience and time getting this change in.

glefloch:
You're welcome that was a pleasure to implement this feature.

PR #4142, opened 2017-04-28
Move elasticsearch node_stats metricset under node.stats namespace

ruflin:
So far the node_stats was putting all its data under `elasticsearch.node_stats.*`. This was changed to `elasticsearch.node.stats.*`.

Further changes:

* Remove support for setting index in metricset. This was never used and can now be done through using format string in the index setting.
* Remove support for setting type in metricset. Type will be removed in elasticsearch.
* Rename ModuleData constant to ModuleDataKey.
* Introduce NamespaceKey as constant to replace hardcoded `_namespace`.
* Cleanup event generation for metricbeat.
* Remove `type` from event as not needed.

### DIFF
--- a/metricbeat/mb/module/event.go
+++ b/metricbeat/mb/module/event.go
@@ -9,10 +9,6 @@ import (
 	"github.com/elastic/beats/metricbeat/mb"
 )
 
-const (
-	defaultType = "metricsets"
-)
-
 // EventBuilder is used for building MetricSet events. MetricSets generate a
 // data in the form of a common.MapStr. This builder transforms that data into
 // a complete event and applies any Module-level filtering.
@@ -38,8 +34,6 @@ func (b EventBuilder) Build() (common.MapStr, error) {
 	}
 
 	// Get and remove meta fields from the event created by the MetricSet.
-	indexName := getIndex(event, "")
-	typeName := getType(event, defaultType)
 	timestamp := getTimestamp(event, common.Time(b.StartTime))
 
 	// Apply filters.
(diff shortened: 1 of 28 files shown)


metricbeat/mb/module/event.go
```diff
@@ -68,37 +59,32 @@ func (b EventBuilder) Build() (common.MapStr, error) {
 	if n, ok := event["_namespace"]; ok {
 		delete(event, "_namespace")
 		namespace = n.(string)
```
andrewkroh: Just to be safe, this should check the boolean that is returned that indicates the success of the type assertion.

metricbeat/mb/module/event.go
```diff
@@ -68,37 +59,32 @@ func (b EventBuilder) Build() (common.MapStr, error) {
...
+		delete(event, mb.ModuleDataKey)
 	}
 
+	moduleEvent := common.MapStr{}
+	moduleEvent.Put(namespace, event)
+
 	// In case meta data exists, it is added on the module level
+	// This is mostly used for shared fields across multiple metricsets in one module
 	if moudleDataExists {
 		if _, ok := moduleData.(common.MapStr); ok {
```
andrewkroh: Instead of ignoring the result of the type assertion with an assignment to `_`, we could use the value in the following expression instead of doing a second type assertion.

metricbeat/mb/module/event.go
ruflin: added

metricbeat/mb/module/event.go
ruflin: changed

ruflin:
jenkins, retest it

Issue #4143, opened 2017-04-28
Remove docker mention in system module docs


andrewkroh:
In the [system module](https://github.com/elastic/beats/edit/5.3/metricbeat/docs/modules/system.asciidoc) there is a misplaced paragraph. I think it should be completely removed.

> It is strongly recommended to not run docker metricsets with a period smaller then 3 seconds. The request to the docker API already takes up to 2s seconds. Otherwise all the requests would timeout and no data is reported.

dedemorton:
Closed by #4147

PR #4147, opened 2017-04-28
Remove docker statement

dedemorton:
Removed as requested in #4143 

@tsg Let me know if this statement needs to go somewhere else.


ruflin:
The note should go here: https://github.com/elastic/beats/blob/master/metricbeat/module/docker/_meta/docs.asciidoc

dedemorton:
@ruflin I've added the note to the docker section. I've changed the language because it sounded like *all* requests will fail if you specify a period that's less than 3. But I'm not sure that's true. If it is true, I can change the note to say, "Specifying less than 3 seconds will cause all requests to timeout, and no data will be reported." Just let me know.

PR #4148, opened 2017-04-28
Fix config example

dedemorton:
Fixed config example that used invalid syntax


ruflin:
@dedemorton Thanks for taking care of this one.

PR #4149, opened 2017-04-29
Move content about processors to filtering section in the doc

dedemorton:
@monicasarbu I'd like for you to take a quick look at this before I change the docs for the other beats. I've streamlined the intro because I want users to know that the prospector-level options exist, but I don't think we should cover them in very much detail since the main point of the section (after the reorg) is to cover processors. Here's what the TOC looks like (with some slight changes from your suggested titles because they made the TOC entries harder to scan): 

image


monicasarbu:
@dedemorton That looks perfect, thanks.

dedemorton:
OK, I'll apply the same redesign to the docs for the other beats, too.

dedemorton:
This is ready to merge as soon as the tests pass.

dedemorton:
@monicasarbu I'm not planning to backport this to 5.4 unless you think I really need to. Resolving the file conflicts might get a little messy because there are a couple of new processors in master...plus probably some changes. Let me know if you think it's important to backport this.

monicasarbu:
I think we should introduce these changes for 6.0.0-alpha1. Thanks @dedemorton, this is a big improvement.

Issue #581, opened 2015-12-22
RabbitMQ output


Semyazz:
Any plans to implement RabbitMQ (RMQ) output? I got isolated environments and I want to send everything through RMQ. I use cert-based auth and so on and it'd be awesome to utilize the same BUS here.

monicasarbu:
Not yet, but it would be great if you take the challenge and add support for it in Packetbeat. We are always encouraging our community to help us by adding support for the protocols they know the best.

monicasarbu:
@Semyazz Ah, sorry, I just noticed that you are referring to the RabbitMQ output. The Beats are not supporting RabbitMQ as output, but you can send data to Logstash that supports RabbitMQ output plugin: https://www.elastic.co/guide/en/logstash/current/plugins-outputs-rabbitmq.html. We don't plan adding more outputs in Beats, see for example the discussion form here: https://github.com/elastic/filebeat/issues/132

Semyazz:
Yea I've seen that discussion and I kinda disagree. To me any queue, especially RabbitMQ which is, I guess the most standard solution and supports such things like (to me the most important) cert-based auth, gives you much more freedom to plan your architecture than logstash. Yea you can implement clustering, routing and so on in Logstash, but why if you already have a working solution. 

Basically I need something light and fast, that will send all collected data to my data BUS (RMQ) and then I can transfer it wherever I want using any kind of fancy topology I come up with and at some point I do have Logstash already to process that data and push it back to RMQ to send it to another place. So as you can see in this case and after reading many posts about logstash/logstash-forwarder and ELK  deployments people do, I believe many people have the very same or at least similar use-cases.

Generally I like Linux's tools philosophy where each tool can do its job reliable and doesn't try to be a huge multipurpose software. In this case RMQ to me is a great data BUS, Logstash is a perfect data processor and Beats just like logstash-forwarder, is the perfect data collector.

rlwmmw:
+1.  
There are obviously multiple ways to approach the problem of moving data around, but having the ability to introduce message queuing at every stage would greatly improve the accuracy of log collection, and alleviate back pressure on LS and ES!

geekpete:
I'm in the same boat, as I currently use rabbitmq to queue inbound messages then have them consumed by a remote logstash. So the only currently supported method involving rabbit is to have beats send to a logstash that outputs to rabbit? I have to insert an additional logstash service in between my log shipper (beats) and my queue (rabbitmq)?

[...]

PR #4150, opened 2017-04-29
Added output for AWS CloudWatch Logs

grudzien:
We have added an output that will push data taken in by beats to AWS CloudWatch Logs. This should be a useful addition to the beats output plugins and give greater flexibility on what beats can push to.

The only requirement for this to work other than a valid AWS account is to pre-create the log group name you are pushing to. We chose not to create that automatically. The stream name, as defined by the configuration file will be created automatically once the program starts.

AWS credentials can be passed in via several methods. First, you can hard code them in the config file, though we do not recommend this method. Second, the credentials will be read directly from an AWS instance, assuming it has the proper IAM role. Third, credentials will be read through various environment variables per Amazon's AWS SDK documentation. Finally, the credentials will be read from a configuration file that gets created when you install the AWS SDK (~/.aws/credentials) on Linux.

### DIFF
--- a/filebeat/docs/reference/configuration.asciidoc
+++ b/filebeat/docs/reference/configuration.asciidoc
@@ -22,6 +22,7 @@ configuration settings, you need to restart {beatname_uc} to pick up the changes
 * <<redis-output>>
 * <<file-output>>
 * <<console-output>>
+* <<awscwl-output>>
 * <<configuration-output-ssl>>
 * <<configuration-output-codec>>
 * <<configuration-path>>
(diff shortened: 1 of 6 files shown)


grudzien:
Journera LLC has signed the CLA. It's likely they did not list my personal email address on the form, which is what my GitHub account is under. I have my work email as a supplemental email address.

tsg:
@grudzien I'm sorry, for the moment we don't want to add any new outputs to Beats.  Please see the related discussion [here](https://github.com/elastic/beats/issues/581#issuecomment-247034174). If you wish, you can maintain the output in its own repo with relatively little hassle, see [here](https://github.com/elastic/beats/pull/1525#issuecomment-217651768). We do plan a way to highlight these "community output plugins" in the future like we do with the community beats, so people can share between them.

PR #4151, opened 2017-04-30
Make kubernetes indexers/matchers pluggable

vjsamuel:
This PR makes kubernetes indexers/matchers pluggable so that users can write their own plugins and plug them into stock/custom beats


ruflin:
jenkins, test it

vjsamuel:
we have internal indexers that we would want to generate `hash` values of the dimensions and store them along with the `kubernetes` metadata. having a plugin interface makes it easier to maintain this code outside the beats repository.

ruflin:
jenkins, retest it

ruflin:
Merging as build seems to be all fine except for pushing the artifacts.

PR #4152, opened 2017-04-30
Abstracting pod interface in kubernetes plugin to enable easier vendoring

vjsamuel:
When a user creates an Indexer outside of the beats repo, he/she needs to vendor in the k8s repo. This should ideally be unnecessary as the retrieval of Pods is abstracted from the user. This PR aims at adding a wrapper struct `Pod` which would be passed to Indexers to retrieve metadata.


libbeat/processors/kubernetes/podwatcher.go
```diff
@@ -8,6 +8,8 @@ import (
 	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/logp"
 
+	"encoding/json"
```
ruflin: some goimports fun ...

vendor/vendor.json
```diff
@@ -160,187 +160,6 @@
 			"revisionTime": "2016-09-16T08:04:11Z"
 		},
 		{
-			"checksumSHA1": "K0iEPnt2DZL5/YrrzAQoMnA+9Pc=",
```
ruflin: 👍

ruflin:
@exekias Could also please have a look?

ruflin:
jenkins, test it

libbeat/processors/kubernetes/podwatcher.go
vjsamuel: bleh! i hate it when it does that.

vjsamuel: updated.

filebeat/beater/filebeat.go
```diff
@@ -20,7 +20,7 @@ import (
 	"github.com/elastic/beats/filebeat/spooler"
 
 	// Add filebeat level processors
-	_ "github.com/elastic/beats/filebeat/processors/kubernetes"
+	_ "github.com/elastic/beats/filebeat/processor/kubernetes"
```
exekias: I think we should keep consistency with libbeat (which uses `processors`)

libbeat/processors/kubernetes/podwatcher.go
```diff
@@ -124,44 +125,64 @@ func (p *PodWatcher) Run() bool {
...
-	delete(p.annotationCache.pods, pod.Metadata.GetUid())
+	delete(p.annotationCache.pods, pod.Metadata.UID)
 
 	for _, index := range p.indexers.GetIndexes(pod) {
 		delete(p.annotationCache.annotations, index)
 	}
 }
 
+func (p *PodWatcher) getPodMeta(pod *corev1.Pod) *Pod {
+
```
exekias: Could you please remove the spurious empty lines at the beginning and end of this function?

filebeat/beater/filebeat.go
ruflin: @exekias I disagree on this one as we should change it for some time in libbeat to `processor`. It is `processors` for historical reasons.

exekias: Ok then, it seems this was agreed some time ago :+1:

PR #4153, opened 2017-05-01
Fixing multiEventFetch error reporting when no events are returned

vjsamuel:
When a metricbeat module implements `func (m *MetricSet) Fetch() ([]common.MapStr, error) ` and returns an error, it is not added to the output event because of:

```
events, err := fetcher.Fetch()
	reporter.ErrorWith(err, event)
}
```
when events is of size 0 this is never executed and this prevents updating of module metrics like:

```
  "metricbeat": {
    "prometheus": {
      "collector": {
        "events": 0,
        "failures": 0,
        "success": 0
      }
    }
  },
```

which can be very misleading. 

This PR fixes the same by checking size of the event list and passing a `nil` event and the error object.


ruflin:
jenkins, test it

Issue #4154, opened 2017-05-01
Wrong "make collect" statement documented instead of "make update" when documentation should be regenerated


christiangalsterer:
https://www.elastic.co/guide/en/beats/metricbeat/current/creating-metricbeat-module.html
documents that the user should run `make collect` when some of the documentation input fields are changed and the documentation should be regenerated. The correct command to run is `make update`.

christiangalsterer:
PR #4155 available

andrewkroh:
Fixed by #4155.

PR #4155, opened 2017-05-01
[#4154] Wrong "make collect" statement documented instead of "make update" when documentation should be regenerated

christiangalsterer:


ruflin:
jenkins, test it

dedemorton:
@ruflin Does this need to be backported? If so, which releases?

ruflin:
I think it applies to all releases >= 5.1. But backporting it to 5.4 and 5.x should be enough I hope?

PR #4156, opened 2017-05-01
New module http

christiangalsterer:
- Initial commit after creating a new PR


christiangalsterer:
@ruflin , here the new PR, as replacement for PR #4092

metricbeat/docs/modules/http.asciidoc
```diff
@@ -0,0 +1,47 @@
...
+This file is generated! See scripts/docs_collector.py
+////
+
+[[metricbeat-module-http]]
+== http Module
+
+Http module is a [Metricbeat](https://www.elastic.co/products/beats/metricbeat) used to call HTTP endpoints.
+Multiple endpoints can be configured which are polled in a regular interval and the result is shipped to the configured output channel.
+
+Httpbeat is inspired by the Logstash [http_poller](https://www.elastic.co/guide/en/logstash/current/plugins-inputs-http_poller.html) input filter but doesn't require that the endpoint is reachable by Logstash as the Metricbeat module pushes the data to the configured outpust channels, e.g. Logstash or Elasticsearch.
```
ruflin: s/outpust/outputs/

metricbeat/docs/modules/http.asciidoc
```diff
@@ -0,0 +1,47 @@
...
+This is often necessary in security restricted network setups, where Logstash is not able to reach all servers. Instead the server to be monitored itself has Metricbeat installed and can send the data or a collector server has Metricbeat installed which is deployed in the secured network environment and can reach all servers to be monitored.
+
+Example use cases are:
```
ruflin: I suggest to remove these examples as we cover these by its own modules (except spring-boot), but that should come as a module soon.

metricbeat/metricbeat.full.yml
```diff
@@ -160,6 +160,15 @@ metricbeat.modules:
   period: 10s
   hosts: ["tcp://127.0.0.1:14567"]
 
+#-------------------------------- http Module --------------------------------
+- module: http
+  metricsets: ["json"]
+  enabled: true
```
ruflin: This should be set to `enabled: false` by default.

metricbeat/module/http/_meta/Dockerfile
```diff
@@ -0,0 +1,11 @@
+# Tomcat is started to fetch Jolokia metrics from it
```
ruflin: I'm thinking if we could use a less "heavy" container for the http testing?

metricbeat/module/http/json/json.go
```diff
@@ -0,0 +1,138 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	err = json.Unmarshal(body, &jsonBody)
+	if err != nil {
+		return nil, err
+	}
+
+	event := common.MapStr{
```
ruflin: I would change this code as following. Note: code will not work directly probably because of some type conversion that is needed.

```
event := jsonBody
event["_namespace"] = m.namespace

if m.config.request.enabled {
	event[mb.ModuleData]["request"] = common.MapStr{
		"headers": m.getHeaders(response.Request.Header),
		"method":  response.Request.Method,
		"body":    m.body,
	}
}

if m.config.response.enabled {
	event[mb.ModuleData]["response"] = common.MapStr{
		"headers": m.getHeaders(response.Request.Header),
		"method":  response.Request.Method,
		"body":    m.body,
	}
}
```

The thinking behind this is as following:

* Httpbeat json is used either for self build http endpoints or for endpoints like spring which are not supported yet by Metricbeat. In these cases the json body contains all the relevant data. So in most cases the response header details is not too relevant if it is a 200. 
* The json document should not be nested under `body` as we already have the namespace to put it under its separate namespace.
* There are definitively use cases where request and response can become interesting, so we should have a config option to enable this. Be aware that these use cases will also overlap potentially with heartbeat and heartbeat could be the better tool for this use case.

metricbeat/docs/modules/http.asciidoc
christiangalsterer: ok, will fix

metricbeat/docs/modules/http.asciidoc
christiangalsterer: ok, will fix

metricbeat/metricbeat.full.yml
christiangalsterer: ok, will fix

metricbeat/module/http/_meta/Dockerfile
christiangalsterer: Any suggestions? I reused the container Jolokia container as it was already there, so that no additional image needs to be build and an additional to be started.

metricbeat/module/http/json/json.go
christiangalsterer: ok, will fix. W.r.t to heartbeat I definitely see also the overlap.

metricbeat/module/http/_meta/Dockerfile
ruflin: I was thinking of having our own very small web service in go that exposes and http endpoint and sends json to make it clear it is a "dynamic" metricset. Here we have something similar for dropwizard: https://github.com/elastic/beats/tree/master/metricbeat/module/dropwizard/_meta/test Except having a http service will hopefully only be a few lines of code and then in the container `go run ...` can be directly used.

metricbeat/module/http/json/json.go
christiangalsterer: one more question: Do you think we should add the custom config struct to the MetricSet (your example) or dedicated fields for each item (current approach)?

ruflin: For enable / disable it should be config options. Not sure what you mean by dedicated fields?

christiangalsterer:
Addressed most PR comments, excepts for the Docker container.

metricbeat/module/http/json/json.go
```diff
@@ -0,0 +1,150 @@
...
+
+	logp.Warn("The http json metricset is in beta.")
+
+	config := struct {
+		Namespace       string `config:"namespace" validate:"required"`
+		Method          string `config:"method"`
+		Body            string `config:"body"`
+		RequestEnabled  bool   `config:"request.enabled"`
+		ResponseEnabled bool   `config:"response.enabled"`
+	}{}
```
ruflin: We probably should define the "defaults" here in a follow up PR.

christiangalsterer: Yes I agree.

ruflin:
jenkins, test it

ruflin:
@christiangalsterer Merged. Thanks a lot for the contribution.

metricbeat/docs/modules/http.asciidoc
```diff
@@ -0,0 +1,41 @@
+////
+This file is generated! See scripts/docs_collector.py
+////
+
+[[metricbeat-module-http]]
+== http Module
+
+Http module is a [Metricbeat](https://www.elastic.co/products/beats/metricbeat) used to call arbitrary HTTP endpoints for which not a dedicated metricbeat is available.
```
tsg: Missing "module" after "a Metricbeat"?

tsg: also there's an extra "not".

tsg: Opps sorry, I intended to comment on #4170. Ignore these.

PR #4157, opened 2017-05-01
Cleanup prospector shutdown

ruflin:
As the harvesters are now sending events directly to the spooler, prospector only has to wait for the stopping of the harvester and not completion of sending events. As a harvester only shuts down if all events are passed to the spooler, it is up to the spooler to guarantee that all events are forwarded.

Further cleanup

* rename `runDone` to `done` for consistency and there is only one done channel left
* runWG is replaced by local onceWg as only needed now when `-once` is enabled. Proper harvester shutdown is taken care of by the harvester registry.
* Remove eventCounter and channelDone as not needed anymore

### DIFF
--- a/filebeat/prospector/prospector.go
+++ b/filebeat/prospector/prospector.go
@@ -25,21 +25,17 @@ var (
 
 // Prospector contains the prospector
 type Prospector struct {
-	cfg           *common.Config // Raw config
-	config        prospectorConfig
-	prospectorer  Prospectorer
-	outlet        channel.Outleter
-	harvesterChan chan *input.Event
-	channelDone   chan struct{}
-	runDone       chan struct{}
-	runWg         *sync.WaitGroup
-	states        *file.States
-	wg            *sync.WaitGroup
-	id            uint64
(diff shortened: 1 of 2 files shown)


filebeat/prospector/prospector.go
```diff
@@ -128,17 +120,19 @@ func (p *Prospector) Start() {
...
+		defer onceWg.Wait()
 	}
 
+	onceWg.Add(1)
 	// Add waitgroup to make sure prospectors finished
-	p.runWg.Add(1)
 	go func() {
 		defer func() {
-			p.runWg.Done()
+			onceWg.Done()
```
andrewkroh: Seems like this should be the last thing called inside this function? You want `Start` to return after `p.stop()` completes for `-once`?

ruflin: `Start` can return as soon as `Run()` is completed. As `Run` is only run once, it will be already stopping when Run is finished. So `onceWg.Done()` is ok to be called before `p.stop()`.

PR #4167, opened 2017-05-02
Move json_error to error.message and error.type

ruflin:
All errors are now under the error namespace. This should also be true for json errors. Currently they are put under `json_error`. This PR changes it to `error.message` and `error.type: json`. This makes it easy to search for all json errors. This is a breaking change.

* Docs updated
* Tests updated

The naming of the config option was not touched for better backward compatibility.

### DIFF
--- a/libbeat/common/jsontransform/jsonhelper.go
+++ b/libbeat/common/jsontransform/jsonhelper.go
@@ -8,14 +8,15 @@ import (
 	"github.com/elastic/beats/libbeat/logp"
 )
 
-func WriteJSONKeys(event common.MapStr, keys map[string]interface{}, overwriteKeys bool, errorKey string) {
+// WriteJSONKeys writes the json keys to the given event based on the overwriteKeys option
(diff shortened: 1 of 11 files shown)


libbeat/common/jsontransform/jsonhelper.go
```diff
@@ -49,3 +49,7 @@ func WriteJSONKeys(event common.MapStr, keys map[string]interface{}, overwriteKe
 
 	}
 }
+
+func CreateJSONError(message string) common.MapStr {
```
tsg: Does this have to be exported? Seems simple enough to replicate if we need it elsewhere, similar to the `debug` function calls.

ruflin: I exported it because I also use it in the json reader. But I can also just copy / paste the code to the json reader.

ruflin:
PR added, function duplicated and golint errors fixed.

ruflin:
jenkins, retest it

PR #4168, opened 2017-05-02
Move fields to each monitor in heartbeat

ruflin:
Move fields.yml configs for each monitor to the monitor itself to make it more pluggable.


tsg:
LGTM.

PR #4171, opened 2017-05-02
Unify event, data and state handling in filebeat

ruflin:
The data and event handling has grown over time in filebeat. One the one hand meta data and event data was mixed and different structure to handle the data existed. The structures which existed were focused in the log prospector. This change brings a unified data handling with a data object which separates event, meta data and state. This new model should allow to add new prospector types and use the same data objects.

The reason it ended up in `util` package is because `data` is taken by the data folder and `common` would lead to too many conflicts with `common` package in libbeat.


ruflin:
jenkins, retest it

PR #4172, opened 2017-05-02
Make nosetests more verbose

ruflin:
This can simplify finding the failing tests when running the system tests locally or on CI.

Also it should prevent Jenkins from triggering a timeout too early because otherwise all dots are written on the same line and Jenkins listens for new log lines to check if the build is stuck.


ruflin:
jenkins, retest it

PR #4173, opened 2017-05-02
Follow up addressing comments in #4127

tsg:
Follow up for #4127 to address comments.


andrewkroh:
LGTM. Sorry for the late review on the other PR.

PR #4178, opened 2017-05-03
Add audit module with kernel metricset to Metricbeat

andrewkroh:
The `kernel` metricset receives audit events from the Linux Audit Framework that
is a part of the Linux kernel.

This metricset establishes a subscription to the `kernel` to receive the events
as they occur.

The Linux Audit Framework can send multiple messages for a single auditable
event. This metricset will combine all of the data from each of the messages
into a single event.

Messages for one event can be interleaved with messages from another event.
This metricset will buffer the messages in order to combine related messages
into a single event even if they arrive interleaved or out of order.

It can resolve numeric UIDs and GIDs in the event to their names.

Syscalls, architectures, and socket addresses are translated from numeric
representations to a human readable form.

Based on the audit message type, a `category` and descriptive `action` are
added to the event.

**NOTE**: The feature is not complete. There will be additional PRs (primarily to make the module be able to install audit rules to the kernel).


andrewkroh:
jenkins, test it

metricbeat/module/audit/kernel/audit_linux.go
```diff
@@ -0,0 +1,303 @@
...
+	mb.BaseMetricSet
+	config Config
+	client *libaudit.AuditClient
+}
+
+// New constructs a new MetricSet.
+func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
+	logBetaOnce.Do(func() {
+		logp.Beta("The %v metricset is a beta feature", metricsetName)
+	})
```
exekias: Why are you doing this? I understand it's to avoid multiple messages on module reload, would it make sense to put this logic inside logp.Beta?

andrewkroh: Exactly, I only wanted it to be logged once. If `logp.Beta` did that for me it would be great. I'm not sure how to implement it in logp without storing every unique message (or hash of the message) ever logged, which isn't necessarily a finite set.

metricbeat/module/audit/kernel/_meta/docs.asciidoc
```diff
@@ -0,0 +1,120 @@
...
+*`kernel.rate_limit`*:: This sets a rate limit on the number of messages/sec
+delivered by the kernel. The default is 0 which means to disable rate limiting.
+Changing this value to anything other than zero can cause messages to be lost.
+The preferred approach to reduce the messaging rate is be more selective in the
+audit ruleset.
+
+*`kernel.include_raw_message`*:: This boolean setting causes Metricbeat to
+include each of the raw messages that contributed to the event in the document.
+The default value is false. This setting is primarily used for development and
+debugging purposes.
```
andrewkroh: Added

metricbeat/metricbeat.full.yml
```diff
@@ -114,6 +114,16 @@ metricbeat.modules:
...
   #password: password
 
+#-------------------------------- Audit Module -------------------------------
+- module: audit
+  enabled: false
+  metricsets: ["kernel"]
+  kernel.resolve_ids: true
+  kernel.backlog_limit: 8196
+  kernel.rate_limit: 0
+  kernel.include_raw_message: false
```
ruflin: Should we make it a namespace? `include.raw_message`, `include.warnings`?

metricbeat/module/audit/kernel/_meta/docs.asciidoc
```diff
@@ -0,0 +1,120 @@
+=== Audit Kernel Metricset
+
+beta[]
```
ruflin: change to experimental

metricbeat/module/audit/kernel/_meta/docs.asciidoc
```diff
@@ -0,0 +1,120 @@
...
+The `kernel` metricset receives audit events from the Linux Audit Framework that
+is a part of the Linux kernel.
+
+This metricset is available only for Linux.
+
+[floag]
+=== How it Works
+
+This metricset establishes a subscription to the `kernel` to receive the events
+as they occur. So unlink most other metricsets, the `period` configuration
```
ruflin: s/unlink/unlike ?

metricbeat/module/audit/kernel/_meta/docs.asciidoc
```diff
@@ -0,0 +1,120 @@
...
+option is unused because it is not implemented using polling.
+
+The Linux kernel only supports a single subscriber to the audit events so this
+metricset cannot be used simultaneously with a service like `auditd`. `auditd`
+should be disabled if this metricset is being used. If you wish to continue to
+use `auditd` instead of this metricset to receive audit messages from the kernel
+then consider using Filebeat to collect the daemon's log files.
```
ruflin: link here to the audit filebeat module

metricbeat/module/audit/kernel/_meta/docs.asciidoc
```diff
@@ -0,0 +1,120 @@
...
+  kernel.rate_limit: 0
+  kernel.include_raw_message: false
+  kernel.include_warnings: false
+----
+
+*`kernel.resolve_ids`*:: This boolean setting enables the resolution of UIDs and
+GIDs to their associated names. The default value is true.
+
+*`kernel.backlog_limit`*:: This controls the maximum number of audit messages
+that will be buffered by kernel.
```
ruflin: buffered by the kernel metricset?

metricbeat/module/audit/kernel/audit_linux.go
ruflin: Is this really needed? I assume modules are not reloaded that often, so having this once in the log for each reload does not seem to bad? This "problem" applies to all metricsets with the beta log, rigth?

Can you change it to experimental?

metricbeat/module/audit/kernel/audit_linux.go
```diff
@@ -0,0 +1,312 @@
...
+
+	config := defaultConfig
+	if err := base.Module().UnpackConfig(&config); err != nil {
+		return nil, errors.Wrap(err, "failed to unpack the audit.kernel config")
+	}
+
+	if logp.IsDebug(metricsetName) {
```
ruflin: Why do you need this `if`? Shouldn't the selector in debugf take care of that?

vendor/github.com/elastic/go-libaudit/CHANGELOG.md
```diff
@@ -0,0 +1,64 @@
+# Change Log
```
ruflin: Can we put this inside the kernel module vendor directory as we do for other modules?

metricbeat/module/audit/kernel/audit_linux.go
andrewkroh: Removed. And changed to experimental.

metricbeat/module/audit/kernel/_meta/docs.asciidoc
andrewkroh: Changed.

metricbeat/module/audit/kernel/_meta/docs.asciidoc
andrewkroh: Fixed. Good catch.

metricbeat/module/audit/kernel/_meta/docs.asciidoc
andrewkroh: Added link.

metricbeat/module/audit/kernel/_meta/docs.asciidoc
andrewkroh: This actually does configure a buffer in the kernel.

metricbeat/module/audit/kernel/audit_linux.go
andrewkroh: I removed it, but this pattern is used to prevent the debug statement from executing if debug is disabled. Even with debug off the statement still gets called and the arguments are evaluated.

vendor/github.com/elastic/go-libaudit/CHANGELOG.md
andrewkroh: I can, but how do I accomplish this with `govendor`? I end up with shadow deps in module/audit/vendor for `github.com/pkg/errors` and `gopkg.in/yaml.v2` because they are used by `github.com/elastic/go-libaudit/...`.

```
cd modules/audit
govendor init
govendor fetch github.com/elastic/go-libaudit/...@v0.0.3
tree -L 2 vendor/
vendor/
├── github.com
│   ├── elastic
│   └── pkg
├── gopkg.in
│   └── yaml.v2
└── vendor.json
```

ruflin: I assume you could put these under `ignore`: https://github.com/elastic/beats/blob/master/metricbeat/module/docker/vendor/vendor.json#L3

@vjsamuel I remember this is something we discussed before and will have that problem more and more often. How do we deal with dependencies that are already available on the top level vendor dir?

andrewkroh:
Squash 'em 👍

vendor/github.com/elastic/go-libaudit/CHANGELOG.md
vjsamuel: @ruflin, the way that I did it for the `libbeat/processors/kubernetes/vendor` is to remove common dependencies from the vendor.json.

there is no other way currently as this is not a first class govendor feature. we are just hacking it :)

PR #4179, opened 2017-05-03
Add info about official Beats docker images

dedemorton:
Adds info about official docker images to the metricbeat docs. This note will only show up if the release state is  set to "released" in versions.asciidoc.

### DIFF
--- a/metricbeat/docs/metricbeat-in-a-container.asciidoc
+++ b/metricbeat/docs/metricbeat-in-a-container.asciidoc
@@ -1,8 +1,23 @@
 [[running-in-container]]
 == Running Metricbeat in a Container
 
-Elastic does not provide any official container images for Metricbeat. The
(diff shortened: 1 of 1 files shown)


andrewkroh:
I think it would be good to have a link to the beats-docker repo. People will be interested to see the contents of the Dockerfile.

And yes, I think we need a similar topic for all Beats that have Docker images. Each of Beats have some specific things we could document.
  - Packetbeat - how to monitor the host's network traffic, what capabilities to add to the container so that it can run as non-privileged (cap_net_raw cap_net_admin)
  - Filebeat - how to monitor the host's logs, how to monitor docker logs

dedemorton:
@andrewkroh I've added a link that points to the beats-docker repo.

I've also created #4192 to get the other beats doc updated with info about running in a container.

andrewkroh:
jenkins, test it

PR #4180, opened 2017-05-03
Filebeat Redis prospector type

ruflin:
This PR adds a new prospector type which reads the slowlog from redis. This slowlog is not in a file but in memory in redis. Because of this filebeat connects to redis and reads out the slowlog. It is important to note that the slow log size is limited in redis that is why after fetching the events from the slowlog filebeat resets the slow log.

Example event looks as following:

```
{
  "@timestamp": "2017-05-16T06:27:17.000Z",
  "beat": {
    "hostname": "ruflin",
    "name": "ruflin",
    "read_timestamp": "2017-05-16T06:27:19.275Z",
    "version": "6.0.0-alpha2"
  },
  "message": "SET hello world",
  "redis": {
    "slowlog": {
      "args": [
        "world"
      ],
      "cmd": "SET",
      "duration": {
        "us": 11
      },
      "id": 38,
      "key": "hello"
    }
  }
}
```

All args are combined in the "message" field output for easy retrieval.


filebeat/_meta/common.full.p2.yml
```diff
@@ -215,6 +215,14 @@ filebeat.prospectors:
 # Configuration to use stdin input
 #- input_type: stdin
 
+#----------------------------- Redis prospector -------------------------------
+#- input_type: redis
```
tsg: I suggest somehow making it more clear that this is strictly about slow logs. People might assume that it's a general Redis input.

filebeat/prospector/redis/doc.go
```diff
@@ -0,0 +1,14 @@
+// Redis package contains prospector and harvester to read the redis slow log
+//
+// The redis slow log is stored in memory. The slow log can be activate on the redis command line as following:
+//
+// 	CONFIG SET slowlog-log-slower-than 100
+//
+// This sets the value of the slow log to 100 micro seconds. All queries taking longer will be reported.
```
tsg: I'd use 2s or 5s as an example. 100 micros might affect the Redis performance, so we should be careful.

filebeat/prospector/redis/doc.go
```diff
@@ -0,0 +1,14 @@
...
+//
+// As the slow log is in memory, in can be configured how many items it consists:
```
tsg: small typo: "it can be configured"

filebeat/prospector/redis/harvester.go
```diff
@@ -0,0 +1,108 @@
...
+			log.key = args[1]
+		}
+		// TODO: Could args contain confidential data? Should it be optional?
+		if argsLen >= 3 {
+			log.args = args[2:]
+		}
+
+		data := util.NewData()
+		data.Event = common.MapStr{
+			// TODO: Should we still send the "fetching" timestamp or overwrite it by timestamp?
```
tsg: Hmm, good question. In the end it needs to be the event timestamp, but I see two options:

* live it like that here, but replace it in the FB redis module
* already set it hear and also set `beat.read_timestamp`

filebeat/prospector/redis/harvester.go
```diff
@@ -0,0 +1,108 @@
...
+		// This splits up the args into cmd, key, args. Should there also be an entry with
+		// all in one line.
+		argsLen := len(args)
+		if argsLen >= 1 {
+			log.cmd = args[0]
+		}
+		if argsLen >= 2 {
+			log.key = args[1]
+		}
+		// TODO: Could args contain confidential data? Should it be optional?
```
tsg: I guess people can use `drop_fields` if they don't want it.

filebeat/prospector/redis/harvester.go
ruflin: I decided for option 2 for the moment. The reasons is that I think it most cases that is what people expect in `@timestamp` here. So also without postprocessing in ingest, the data is already correct.

filebeat/prospector/redis/doc.go
ruflin: changed

filebeat/prospector/redis/doc.go
ruflin: fixed

filebeat/_meta/common.full.p2.yml
ruflin: That is a tricky one. Thinking more about the prospectors I get the impression we will have prospector with multiple harvester types potentially (similar to what we have in metricbeat). For example in file there could be `file.csv` or `file.xml`. What if we not only have slowlog for redis? But TBH I'm not sure yet what we should do here. This could also well be overengineered and we will not need it.

For the moment I think I fix the above with call it `Redis slowlog prospector` in the docs but still `redis` as the prospector type.

ruflin:
@tsg Ready for an other look. I think we are at the point now where we have to decide on what to do with it :-)

ruflin:
@dedemorton So far I only added the minimial options of docs for this = almost nothing. The reason is that this kind of challenges our existing doc structure we have for prospectors. So far the 2 prospectors shared 95% of the config options. With this change, they can be very different. I'm thinking if we should start to organise docs similar to what we do with the modules and have a block for each prospector type. I will need your help here and suggest we will only tackle this after merging the PR.

dedemorton:
@ruflin I remember you mentioned this problem awhile back ago in an email, but I haven't had a chance to follow up with you. What release are we targeting for this PR?

ruflin:
@dedemorton Not sure if I mentioned that before (but could be). It's not urgent. If it happens, somewhere around 6.0

filebeat/filebeat.full.yml
```diff
@@ -379,6 +379,27 @@ filebeat.prospectors:
 # Configuration to use stdin input
 #- type: stdin
 
+#------------------------- Redis slowlog prospector ---------------------------
+# Experimental: Config options ofr the redis slow log prospector
```
tsg: small typo on "ofr".

ruflin:
@tsg New version pushed
@dedemorton I created https://github.com/elastic/beats/issues/4398 as a follow up Github issue for the docs.

PR #4184, opened 2017-05-03
Remove duplicate error key in fields.yml

tsg:
`error` is now an object, but the old `error` key definition was still in `fields.yml`. This
led to some weird behaviours on objects having an error key.


ruflin:
jenkins, retest it

PR #4185, opened 2017-05-03
Add type to state info

ruflin:
With the introduction of new prospector types each prospector can have different state information. To differentiate between the different states for the prospector the `type` was added to each registry state.

### DIFF
--- a/filebeat/input/file/state.go
+++ b/filebeat/input/file/state.go
@@ -10,17 +10,18 @@ import (
 
 // State is used to communicate the reading state of a file
 type State struct {
-	Source      string      `json:"source"`
-	Offset      int64       `json:"offset"`
-	Finished    bool        `json:"-"` // harvester state
-	Fileinfo    os.FileInfo `json:"-"` // the file info
-	FileStateOS StateOS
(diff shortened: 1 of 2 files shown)


filebeat/input/file/state.go
```diff
@@ -10,24 +10,26 @@ import (
...
-	Finished    bool        `json:"-"` // harvester state
-	Fileinfo    os.FileInfo `json:"-"` // the file info
-	FileStateOS StateOS
+	Finished    bool          `json:"-"` // harvester state
+	Fileinfo    os.FileInfo   `json:"-"` // the file info
+	Source      string        `json:"source"`
+	Offset      int64         `json:"offset"`
 	Timestamp   time.Time     `json:"timestamp"`
 	TTL         time.Duration `json:"ttl"`
+	Type        string        `json:"type"`
```
tsg: Name it "input_type" to differentiate it from the "document_type"?

ruflin: I'm hoping to have only one type in the future. More accurate here would be `ProspectorType` but as I try to keep the state as "independent" as possible with potentially have it in libbeat one day, I would just call it `Type`.

PR #4186, opened 2017-05-03
Assert on errors in the filebeat modules

tsg:
We used to whitelist the mysql/slowlog and system/auth fileset when
checking if there were any errors while parsing. This removes all whitelisted
entries and always fails if there are error keys generated.

The mysql module was whitelisted because Mysql writes a header that doesn't match
the patterns. I solved this by excluding the header via the Filebeat config.

The auth fileset was already fixed to not generate errors, at least for the current
test set.

### DIFF
--- a/filebeat/module/mysql/slowlog/config/slowlog.yml
+++ b/filebeat/module/mysql/slowlog/config/slowlog.yml
@@ -3,8 +3,9 @@ paths:
 {{ range $i, $path := .paths }}
  - {{$path}}
 {{ end }}
-exclude_files: [".gz$"]
+exclude_files: ['.gz$']
 multiline:
-  pattern: "^# User@Host: "
+  pattern: '^# User@Host: '
   negate: true
   match: after
+exclude_lines: ['^[\/\w\.]+, Version: .* started with:.*']   # Exclude the header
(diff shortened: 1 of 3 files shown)


filebeat/module/mysql/slowlog/config/slowlog.yml
```diff
@@ -3,8 +3,9 @@ paths:
...
  - {{$path}}
 {{ end }}
-exclude_files: [".gz$"]
+exclude_files: ['.gz$']
 multiline:
-  pattern: "^# User@Host: "
+  pattern: '^# User@Host: '
   negate: true
   match: after
+exclude_lines: ['^[\/\w\.]+, Version: .* started with:.*']   # Exclude the header
```
ruflin: Should we add a note about this to the docs?

tsg: I don't know, I worry it would cause more confusion than it solves. Something like "The slowlog fileset ignores the header written by MySQL on startup."?

ruflin: I see your point. Lets ignore it for now.

tsg:
jenkins, retest it

ruflin:
jenkins, retest it

Issue #4164, opened 2017-05-01
have to save Packetbeat HTTP dashboard before you can print PDF in Kibana


LeeDr:
For confirmed bugs, please report:
- Version: 5.4.0 (I also went back and tested on 5.2.0 and see the same issue)
- Operating System: Ubuntu 14.04
- Steps to Reproduce: 
1. load packetbeat dashboards, something like this;
`./import_dashboards --insecure -user $ELASTICUSER -pass $ELASTICPWD -es ${ESURL} -url https://${BASEURL}/downloads/beats/beats-dashboards/beats-dashboards-${VERSION}${SNAPSHOT}.zip`
2. install and start packetbeat (might not really matter if you have data or not)
3. in Kibana go to Dashboards and open `Packetbeat HTTP`
4. click Reporting
5. Expect to see the Generate Printable PDF button
  Actually see a message "Please save your work before generating a report."

Also `Packetbeat MySQL performance`
And `Packetbeat MongoDB performance`
And `Packetbeat Thrift performance`
And `Packetbeat PgSQL performance`

A Kibana developer looked at it and said Kibana thinks it hasn't been saved because it's missing this;
```
"uiStateJSON": {
        "P-7": {
            "vis": {
                "params": {
                    "sort": {
                        "columnIndex": null,
                        "direction": null
                    }
                }
            }
        }
    },
```

If you save the dashboard that code above is added to the dashboard and it can be printed.

I may or may not have found all the dashboards with this issue.  I used this hack to try to find the problematic ones;

```
$ curl -XGET -k -u elastic:changeit "https://localhost:9200/.kibana/dashboard/_search" -H 'Content-Type: application/json' -d'
{
  "query": {
    "match_all": {}
  },
  "size": 50
}' | sed 's|,|,\n|g' | grep -E -w 'uiStateJSON|title'
```

which gives an output like this;
```
"title":"Metricbeat MySQL",
"uiStateJSON":"{}",
"title":"CPU/Memory per container",
"uiStateJSON":"{\"P-2\":{\"vis\":{\"params\":{\"sort\":{\"columnIndex\":null,
"title":"Metricbeat - Apache HTTPD server status",
"uiStateJSON":"{\"P-2\":{\"vis\":{\"params\":{\"sort\":{\"columnIndex\":null,
"title":"Metricbeat Docker",
"uiStateJSON":"{\"P-1\":{\"vis\":{\"params\":{\"sort\":{\"columnIndex\":1,
[...]

PR #4188, opened 2017-05-03
Re-export Packetbeat dashboards with a newer version of Kibana

monicasarbu:
Import and export the Packetbeat dashboards again, in order to have the `uiStateJSON` set. This should fix the issue https://github.com/elastic/beats/issues/4164.


monicasarbu:
I have tested and it seems to work for me.

PR #4190, opened 2017-05-03
Make state comparison more generic

ruflin:
For the state comparison FileStateOS was accessed directly. By having new prospector types there might be different state objects. This is a first step to remove the dependency on FileStateOS and do state comparison string based.

### DIFF
--- a/filebeat/input/file/state.go
+++ b/filebeat/input/file/state.go
@@ -1,7 +1,6 @@
 package file
 
 import (
-	"fmt"
 	"os"
 	"sync"
 	"time"
@@ -11,6 +10,7 @@ import (
 
 // State is used to communicate the reading state of a file
 type State struct {
+	Id          string        `json:"-"` // local unique id to make comparison more efficient
 	Finished    bool          `json:"-"` // harvester state
 	Fileinfo    os.FileInfo   `json:"-"` // the file info
 	Source      string        `json:"source"`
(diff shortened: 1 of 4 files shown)


filebeat/input/file/state.go
```diff
@@ -31,6 +32,16 @@ func NewState(fileInfo os.FileInfo, path string) State {
...
 }
 
+// String returns a unique id for the state as a string
+func (s *State) String() string {
+	return s.FileStateOS.String()
+}
+
+// IsEqual compares the state to an other state supporing stringer based on the unique string
+func (s *State) IsEqual(c fmt.Stringer) bool {
+	return s.String() == c.String()
```
tsg: Somehow relying on `String()` functions seems dangerous to me. Perhaps you could have the same logic, but based on a string `id` that is only computed once and stored in the state?

ruflin: I initially had a `ID()` function for comparison. I switched it to `String()` because the `Stringer` interface already exist. But perhaps have a local interface with `ID() string` is a better choice? +1 on storing it locally and not generating it every time, as it should not change over time.

ruflin:
New version pushed

filebeat/input/file/state.go
```diff
@@ -10,6 +10,7 @@ import (
 
 // State is used to communicate the reading state of a file
 type State struct {
+	Id          string        `json:"-"` // local unique id to make comparison more efficient
```
ruflin: this can't be ID as it otherwise conflicts with the ID() method.

PR #4195, opened 2017-05-03
Update overview.asciidoc

markwalkom:
Added a few "local" mentions in to make it clearer that filebeat reads local files only.


ruflin:
@dedemorton This one will also need forward porting.

Issue #3604, opened 2017-02-16
new protocol: SSL envelope


urso:
Add protocol analyzer for parsing and analyzing the SSL envelope. Information like protocol used und cipher suites can still be gathered. Some SSL/TLS implmentations do close the connection if handshake fails. Report this as a special event. 

Some information gathered can be added to flows.

This is not about decrypting the payload!

RFC overview: https://tools.ietf.org/wg/tls/
[TLS 1.2](https://tools.ietf.org/html/rfc5246), [TLS 1.1](https://tools.ietf.org/html/rfc4346), [TLS 1.0](https://tools.ietf.org/html/rfc2246), [SSL 3.0](https://tools.ietf.org/html/rfc6101)

- [x] add support for storing raw certificates
- [ ] update CHANGELOG
- [ ] refactor as per [tcp-protocol](https://github.com/elastic/beats/blob/master/packetbeat/scripts/tcp-protocol/README.md)
- [x] update docs
- [x] dashboard

tsg:
I think we can consider this one closed by https://github.com/elastic/beats/pull/5476.

adriansr:
I had pending to do some small changes (store binary certificates) and a little cleanup before closing this, but up to you

tsg:
@adriansr Ah, that sounds good. Feel free to edit the original description and add checkboxes with what's left to do. Then we can transform it in a meta issue (like, for example, this one: https://github.com/elastic/beats/issues/5432)

lctrcl:
any chance to add support for [ja3](https://github.com/salesforce/ja3) as well?

adriansr:
@lctrcl looks easy to implement. Feel free to open a feature request

tsg:
This can be considered done.

PR #4197, opened 2017-05-03
Add more info about flows to the doc

dedemorton:
First cut of the content on flows.

I also moved up a couple of items in the TOC because they seemed important.

Please respond to the comments that I've added inline.


packetbeat/docs/flows.asciidoc
```diff
@@ -0,0 +1,104 @@
...
+* UDP
+* SSL
+
+//REVEIWER: Not sure this list is complete. should it also say Application? Anything else?
+
+Here's an example of flow events visualized in the Flows dashbaord:
+
+//REVIEWERS: Is it OK for me to show all these IP addresses (not sure what might be sensitive)? I don't worry about that as much as I should, but now that I know Mr. Robot is watching...perhaps I should be more careful. Do we have a better image to use? This one is actually from a v5.3.
+
+image:./images/flows.png[]
```
monicasarbu: @dedemorton I am not able to find flows.png. Did you forget to add it to the PR?

packetbeat/docs/flows.asciidoc
```diff
@@ -0,0 +1,104 @@
...
+your network.
+
+For each flow, Packetbeat reports the number of packets and the total number of
+bytes sent from the source to the destination. Each flow event also contains
+information about the source and destination hosts, such as their IP address and
+location. For bi-directional flows, Packetbeat reports statistics for the reverse
+flow.
+
+Packetbeat collects and reports statistics per flow on the following network
+layers:
```
andrewkroh: When I hear network layers I think of the [OSI model](https://en.wikipedia.org/wiki/OSI_model). Flows collect the mac address and vlan ID from layer 2 (data-link), the ip address from layer 3, and udp/tcp port info from layer 4 (transport).

It doesn't do anything higher level layers yet (like SSL).

packetbeat/docs/flows.asciidoc
```diff
@@ -0,0 +1,104 @@
...
+  timeout: 30s
+  period: 10s
+--------------------------------------------------------------------------------
+
+See <<configuration-flows>> for more info about these configuration options.
+
+Here’s an example of a flow event sent by Packetbeat. See
+<<exported-fields-flows_event>> for a description of each field.
+
+//REVIEWERS: As always, I sit on the fence about whether to show the output event because it may change.
```
monicasarbu: Maybe we can include the content of the [https://github.com/elastic/beats/blob/master/metricbeat/module/system/diskio/_meta/data.json](data.json). Unfortunately this exists only for the Metricbeat metricsets, and not for all Beats.

andrewkroh: I like seeing the output in docs.

But I would only show the contents of the `_source` field (the other stuff is specific to ES). To be consistent with some other example docs, change the hostname and name to `host.example.com`. You could swap out the IP addresses with ones reserved for documentation purposes (see [ IPv4 Address Blocks Reserved for Documentation](https://tools.ietf.org/html/rfc5737)).

packetbeat/docs/flows.asciidoc
```diff
@@ -0,0 +1,104 @@
...
+You can configure Packetbeat to collect and report statistics on network flows.
+A _flow_ is a group of packets sent over the same time period that share
+common properties, such as the same source and destination address and protocol.
+You can use this feature to analyze network traffic over specific protocols on
+your network.
+
+For each flow, Packetbeat reports the number of packets and the total number of
+bytes sent from the source to the destination. Each flow event also contains
+information about the source and destination hosts, such as their IP address and
+location. For bi-directional flows, Packetbeat reports statistics for the reverse
```
andrewkroh: Location is not provided by Beat, but we do have the field in our index template. We should explain how to get these fields populated with data using geoip lookups.

packetbeat/docs/flows.asciidoc
```diff
@@ -0,0 +1,104 @@
...
+flow.
+
+Packetbeat collects and reports statistics per flow on the following network
+layers:
+
+* IP/TCP
+* UDP
+* SSL
```
monicasarbu: I think SSL is not done yet. There is an open issue for it: https://github.com/elastic/beats/issues/3604

packetbeat/docs/flows.asciidoc
```diff
@@ -0,0 +1,104 @@
...
+
+//REVEIWER: Not sure this list is complete. should it also say Application? Anything else?
```
monicasarbu: No for Application.

packetbeat/docs/flows.asciidoc
```diff
@@ -0,0 +1,104 @@
...
+  "_version": 1,
+  "_score": null,
+  "_source": {
+    "@timestamp": "2017-05-03T19:42:40.003Z",
+    "beat": {
+      "hostname": "My-MacBook-Pro.local",
+      "name": "My-MacBook-Pro.local",
+      "version": "{stack-version}"
+    },
+    "connection_id": "AQAAAAAAAAA=",
```
andrewkroh: I think the field documentation needs more details too, but we probably need urso to help us out. Some questions I had about the data:

- What is `connection_id` and `icmp_id` and where to they come from? How can I use them?
- Clarify (and confirm) that `net_bytes_total` includes the total amount of bytes across all layers. Does this count retransmits?
- Clarify if `net_packets_total` includes retransmits.
- `transport` does not appear in the field docs. What are the possible values of the field?

packetbeat/docs/flows.asciidoc
```diff
@@ -0,0 +1,104 @@
...
+location. For bi-directional flows, Packetbeat reports statistics for the reverse
+flow.
+
+Packetbeat collects and reports statistics per flow on the following network
+layers:
+
+* IP/TCP
+* UDP
+* SSL
+
```
andrewkroh: Somewhere in here we should explain the `final` flag and how the packetbeat sends intermediate reports about each flow that it is tracking with the `final = false`, then when the flow completes it sends one last event with `final = true`. Users that wish to aggregate sums of traffic will want to know this because they should be filtering on `final:true` (or using some other technique to get only the latest update from each flow).

packetbeat/docs/flows.asciidoc
dedemorton: @monicasarbu whoops! just added it.

packetbeat/docs/flows.asciidoc
dedemorton: @andrewkroh Do you think users will want to know this level of detail? If not, then I can remove lines 16-21.

andrewkroh: I think it's enough to say it collects info up to the transport layer. The details aren't that important since they can jump over to the exported field docs.

packetbeat/docs/flows.asciidoc
dedemorton: I've created a separate issue to track this list of issues: https://github.com/elastic/beats/issues/4236

packetbeat/docs/flows.asciidoc
dedemorton: @andrewkroh I've created a PR to track this work because I don't have time to do it right now. https://github.com/elastic/beats/issues/4238

dedemorton:
This PR is ready for a final review.

packetbeat/docs/flows.asciidoc
robcowart: Just wanted to point out the is not `period: -1` the correct setting is `period: -1s`

It is also incorrect in the existing docs. I ran into this a couple weeks ago.

andrewkroh: @robcowart Thanks and could you please open a new issue for this. Otherwise I'm afraid it might get lost in the noise. Plus DeDe is on vacation this week so even more change it might not be seen right away by her.

PR #4199, opened 2017-05-04
Update how-filebeat-works.asciidoc

CaptainFeng:


ruflin:
@CaptainFeng Thanks for the fix.

@dedemorton One more that needs forward porting.

Issue #3686, opened 2017-02-28
Document complex objects support in env/cli vars


urso:
See discussion: https://discuss.elastic.co/t/array-support-for-yml-environment-values/76657/2

One can insert lists/dictionaries in configurations via environment variables/cli flags using a superset of JSON ([godoc comment](https://github.com/elastic/go-ucfg/blob/master/internal/parse/parse.go#L23)). This functionality is currently not documented.

dedemorton:
Resolved with https://github.com/elastic/beats/pull/4200

PR #4200, opened 2017-05-04
Document complex object support in env vars

dedemorton:
Adds info requested in issue #3686 . Also did some reorganizing so that we no longer have duplicated content in two places.  Had to add a little conditional coding to embed the file.

Please respond to my questions inline.

### DIFF
--- a/libbeat/docs/config-file-format.asciidoc
+++ b/libbeat/docs/config-file-format.asciidoc
@@ -110,7 +110,7 @@ Simple filebeat example with partially collapsed setting names and use of compac
 
 filebeat.prospectors:
 - type: log
-  paths: ["/var/log/*.log"] 
+  paths: ["/var/log/*.log"]
   multiline.pattern: '^['
   multiline.match: after
 
@@ -213,44 +213,7 @@ format-string-with-date: '%{[fieldname]}-%{+yyyy.MM.dd}'
 [[config-file-format-env-vars]]
 === Environment Variables
 
-Beats support use of environment variables in config files to set values that
-need to be configurable during deployment. Environment variable expansion is
-introduced using `${VAR}`, where `VAR` is the name of the environment variable.
-
-Note: Only values can be set using environment variables. Environment variables
-usage in namespace and setting names are not supported.
-
-Variable references are replaced when settings are read by beats. The
-replacement is case-sensitive and occurs after the YAML file itself has been
-parsed. References to undefined variables will lead to errors when dereferenced
-and no default value is specified. To specify a default value, use:
-
-`${VAR:default_value}`
-
-Where `default_value` is the value to use if the environment variable is
-undefined.
-
-If you need to use a literal `${` in your configuration file then you can write
-`$${` to escape the expansion. The `$` symbol can be used to escape other
-characters in the default_value like using `$}` in order to generate a `}`
-character without closing the variable expansion.
-
-After changing the value of an environment variable, the beat needs to be
-restarted to pick up the new value.
-
-[float]
-==== Examples
-
-Here are some examples of configurations that use environment variables
-and what each configuration looks like after replacement:
-
-[options="header"]
-|==================================
-|Config source	       |Environment setting   |Config after replacement
-|`name: ${NAME}`       |`export NAME=elastic` |`name: elastic`
-|`name: ${NAME:beats}` |no setting            |`name: beats`
-|`name: ${NAME:beats}` |`export NAME=elastic` |`name: elastic`
-|==================================
+include::shared-env-vars.asciidoc[]
 
 [[config-gile-format-refs]]
 === Reference Variables
@@ -333,7 +296,7 @@ When installed via an RPM or DEB package, the config file at
 `/etc/{beatname}/{beatname}.yml` will have the proper owner and permissions. The
 file is owned by `root` and has file permissions of `0644` (`-rw-r--r--`).
 
-You may encounter the following errors if your config file fails these checks: 
+You may encounter the following errors if your config file fails these checks:
 
 ["source","sh"]
 --------------------------------------------------------------------------------
@@ -341,7 +304,7 @@ Exiting: error loading config file: config file ("{beatname}.yml") must be
 owned by the beat user (uid=501) or root
 --------------------------------------------------------------------------------
 
-To correct this problem you can use either `chown root {beatname}.yml` or 
+To correct this problem you can use either `chown root {beatname}.yml` or
 `chown 501 {beatname}.yml` to change the owner of the configuration file.
 
 ["source","sh"]
--- a/libbeat/docs/config-file-format.asciidoc
+++ b/libbeat/docs/config-file-format.asciidoc
@@ -335,13 +335,41 @@ overwrite settings via local configuration files.
 In addition to overwriting settings using multiple configuration files,
 individual settings can be overwritten using `-E <setting>=<value>`. The
 `<value>` can be either a single value or a complex object, such as a list or
-dictionary. For example, you can disable the Elasticsearch output and write all
-events to the console by setting:
+dictionary.
+
+For example, given the following configuration:
+
+["source","yaml"]
+--------------------------------------------------------------------------------
+output.elasticsearch:
+  hosts: ["http://localhost:9200"]
+  username: username
+  password: password
+--------------------------------------------------------------------------------
+
+You can disable the Elasticsearch output and write all events to the console by
+setting:
 
 ["source","sh"]
 --------------------------------------------------------------------------------
 -E output='{elasticsearch.enabled: false, console.pretty: true}'
 --------------------------------------------------------------------------------
+
+Any complex objects that you specify at the command line are merged with the
+original configuration, and the following configuration is passed to the Beat:
+
+["source","yaml"]
+--------------------------------------------------------------------------------
+output.elasticsearch:
+  enabled: false
+  hosts: ["http://localhost:9200"]
+  username: username
+  password: password
+
+output.console:
+  pretty: true
+--------------------------------------------------------------------------------
+
 
 [[config-file-format-tips]]
 === YAML Tips and Gotchas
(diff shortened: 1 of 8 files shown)


libbeat/docs/shared-env-vars.asciidoc
```diff
@@ -49,3 +67,40 @@ and what each configuration looks like after replacement:
...
 |`name: ${NAME:beats}` |`export NAME=elastic` |`name: elastic`
 |==================================
+
+[float]
+=== Specifying Complex Objects in Environment Variables
+
+You can specify complex objects, such as lists or dictionaries, in environment
+variables by using a superset of JSON.
+
+//REVIEWERS: Not sure I like describing this as a "superset of JSON." Would it make sense to say "by using a JSON-like syntax"?
```
tsg: +1 for "JSON-like syntax"

libbeat/docs/shared-env-vars.asciidoc
```diff
@@ -49,3 +67,40 @@ and what each configuration looks like after replacement:
...
+
+You can reference this variable in the config file:
+
+[source,yaml]
+-------------------------------------------------------------------------------
+exclude_lines: ${LOG_EXCLUDE_LINES}
+-------------------------------------------------------------------------------
+
+
+//REVIEWERS: Does anyone have a better example that is less Filebeat-specific?
```
tsg: The typical example I've seen is adding multiple Elasticsearch hosts.

libbeat/docs/shared-env-vars.asciidoc
```diff
@@ -32,8 +39,19 @@ undefined.
 If you need to use a literal `${` in your configuration file then you can write
 `$${` to escape the expansion.
 
+// REVIEWERS: Is the following statement still true? "The `$` symbol can be used to escape othercharacters in the default_value like using `$}` in order to generate a `}` character without closing the variable expansion."
```
urso: Yes, still should be true.

libbeat/docs/shared-env-vars.asciidoc
```diff
@@ -49,3 +67,36 @@ and what each configuration looks like after replacement:
 |`name: ${NAME:beats}` |no setting            |`name: beats`
 |`name: ${NAME:beats}` |`export NAME=elastic` |`name: elastic`
 |==================================
+
+[float]
+=== Specifying Complex Objects in Environment Variables
```
urso: Note: the same syntax for complex objects can be used via `-E` setting as well.

e.g. I can do from command line: `-E output='{elasticsearch.enabled: false, console.pretty: true}'`, to disable the ES output and have all events written to console.

libbeat/docs/shared-env-vars.asciidoc
```diff
@@ -49,3 +67,36 @@ and what each configuration looks like after replacement:
...
+  hosts: '${ES_HOSTS}'
+-------------------------------------------------------------------------------
+
+When the Beat loads the config file, it resolves the environment variable and
+replaces it with the specified list before reading the `hosts` setting.
+
+When you specify complex objects in environment variables, keep the following
+in mind:
+
+* Strings can be unquoted, single-quoted, or double-quoted
```
urso: Note: the quotation rules are for a) convenience in simple settings b) different quotation rules for somewhat safely mixing quotation in shell itself.

libbeat/docs/shared-env-vars.asciidoc
```diff
@@ -49,3 +67,36 @@ and what each configuration looks like after replacement:
...
+* Arrays at the top-level do not require brackets (`[]`). Instead, use commas (`,`) to
+separate the elements.
```
urso: Sounds to me like we discourage (don't allow/want) users to use `[]` for top-level arrays. Especially for more complex objects I'd encourage `[]` even on top-level. It's just `[]` is optional for top-level arrays for user convience only.

libbeat/docs/shared-env-vars.asciidoc
dedemorton: Will add to the topic that discusses the -E option

libbeat/docs/shared-env-vars.asciidoc
dedemorton: Fixed.

libbeat/docs/shared-env-vars.asciidoc
dedemorton: added clarification.

dedemorton:
@urso Added the changes you suggested. I'm not sure if I got the wording quite right on the bit about quotations (let me know). Let me know if you think it's not clear enough. I think it's probably good enough just to explain that the flexibility is there...I don't want to go into too much detail.

libbeat/docs/shared-env-vars.asciidoc
```diff
@@ -49,3 +65,38 @@ and what each configuration looks like after replacement:
 |`name: ${NAME:beats}` |no setting            |`name: beats`
 |`name: ${NAME:beats}` |`export NAME=elastic` |`name: elastic`
 |==================================
+
+[float]
+=== Specifying Complex Objects in Environment Variables
+
+You can specify complex objects, such as lists or dictionaries, in environment
+variables by using a JSON-like syntax.
```
urso: maybe we want to be more explicit here, as `JSON-like syntax` is pretty unspecific on syntax limitations + move the pointers from line 93pp up here.

e.g.:

Like with JSON, dictionaries and lists are constructed using `{}` and `[]`. But in comparison to JSON, the syntax allows for trailing commas and slightly different string quotation rules. Strings can be unquoted, single-quoted, or double-quoted, as a convenience for simple settings and to make it easier for you to mix quotation usage in the shell. Arrays at the top-level do not require brackets (`[]`).

libbeat/docs/config-file-format.asciidoc
```diff
@@ -370,7 +333,15 @@ use this, for example, for setting defaults in a base configuration file, and
...
-individual settings can be overwritten using `-E <setting>=<value>`.
+individual settings can be overwritten using `-E <setting>=<value>`. The
+`<value>` can be either a single value or a complex object, such as a list or
+dictionary. For example, you can disable the Elasticsearch output and write all
+events to the console by setting:
+
+["source","sh"]
+--------------------------------------------------------------------------------
+-E output='{elasticsearch.enabled: false, console.pretty: true}'
+--------------------------------------------------------------------------------
```
urso: maybe include sample config and final config here?

e.g. with `-E output='{elasticsearch.enabled: false, console.pretty: true}'`

and a config file like:

```
output.elasticsearch:
  hosts: ["http://localhost:9200"]
  username: username
  password: password
```

then the final configuration presented to the beat becomes:

```
output.elasticsearch:
  enabled: false
  hosts: ["http://localhost:9200"]
  username: username
  password: password

output.console:
  pretty: true
```

That is, overwriting fields with complex objects will merge the object with the original configuration, not changing/overwriting any other settings.

libbeat/docs/shared-env-vars.asciidoc
```diff
@@ -49,3 +65,38 @@ and what each configuration looks like after replacement:
 |`name: ${NAME:beats}` |no setting            |`name: beats`
 |`name: ${NAME:beats}` |`export NAME=elastic` |`name: elastic`
 |==================================
```
urso: regarding this table, 2 syntax types are missing:

According [to this](https://github.com/elastic/go-ucfg/blob/master/variables.go#L68) we've got 3 kind of operators:

`:`  is default value operator
`:?` is the error message operator. If the environment variable can not be expanded, the error message will contain the custom error message... e.g. `name: ${NAME:?You need to set the NAME environment variable}` will have the error message contain the string `You need to set the NAME environment variable`, if `NAME` is not set.
`:+` is alternate value operator (if variable is set, expand with alternative value)
        e.g. name: ${NAME:+beats} with export NAME=elastic becomes `name: beats`. By nesting variable expansions with default cases and alternative rules one can build some kind of negation and checks. This is quite some advanced use-case almost no one will use in practice (it's somewhat tricky to use this one right). e.g. use case `include other string if variable is set`:

```
cmd: a b ${param:+-x ${param}}
```

will expand to `cmd: a b -x y` if `export param=y` and `cmd: a b` only if `param` is not set.

libbeat/docs/shared-env-vars.asciidoc
dedemorton: good idea. fixed.

libbeat/docs/config-file-format.asciidoc
dedemorton: added

libbeat/docs/shared-env-vars.asciidoc
dedemorton: @urso I'm looking at this in 6.0.0-alpha2, and the error message with the custom text isn't what you would expect.  When `name: ${NAME:?You need to set the NAME environment variable}` is set in the config and the NAME env var can't be resolved, the Beat returns the error `You need to set the NAME environment variable accessing 'name' (source:'filebeat.yml')`. (Whatever you specify for custom text gets prepended to `accessing 'name' (source:'filebeat.yml')`.  I'll push the changes I made, but I'm not sure we should document the error syntax since it results in weird messages.

Also, since the alternate value operator is an advanced use case that almost no one will use in practice AND it's tricky to use, I don't think we should document it. We might cause more problems than we solve by documenting it. WDYT?

urso: oh, I did forget `go-ucfg` always appending the full settings name and source when reporting an error.

I'm ok with not documenting the `:+` operator, as it's somewhat complex. About the error operator `:?` I'm somewhat on the fence. I'm ok with just leaving it out or maybe adjusting the error message in go-ucfg itself.

urso:
LGTM

libbeat/docs/shared-env-vars.asciidoc
dedemorton: @urso I think we should document the `:?` operator and adjust the go-ucfg message. It seems likely that some users will want to return an error that provides some context when a variable cannot be resolved.

Issue #4202, opened 2017-05-04
Use MemAvailable from /proc/meminfo in the Linux 3.14 kernel


monicasarbu:
A great new metric has been introduced in “/proc/meminfo” in the Linux 3.14 kernel — MemAvailable:
```
An estimate of how much memory is available for starting new applications, without 
swapping. 
Calculated from MemFree, SReclaimable, the size of the file LRU lists, and the 
low watermarks in each zone.

The estimate takes into account that the system needs some page cache to function well, 
and that not all reclaimable slab will be reclaimable, due to items being in use. 
The impact of those factors will vary from system to system.
```
We need to adjust the gosigar library to use MemAvailable when it's available in "/proc/meminfo".

cytim:
I just wanted to create the same issue and found this. 👍 

Related discussion on the forum:

- [Metricbeat system.memory.actual.free - is it accurate?](https://discuss.elastic.co/t/metricbeat-system-memory-actual-free-is-it-accurate/80650)
- [Incorrect “system.memory.actual.free” metrics](https://discuss.elastic.co/t/incorrect-system-memory-actual-free-metrics/83783)

tony2k84:
Is this fixed and which release?

PR #4206, opened 2017-05-04
Use gopsutil library for getting memory and swap stats

monicasarbu:
Use gopsutil library for getting memory statistics. This should fix the issue https://github.com/elastic/beats/issues/4202. 

Note: This PR keeps the old fields. In a later PR, I would like to reduce the number of exported fields. For example, we should be able to delete `system.actual.used` as it can be with the Time Series Visual Bilder in Kibana.


ruflin:
+1 on reducing the number of fields if they can be easily calculated.

andrewkroh:
Can you confirm that we are not losing any platform support with the switch. We currently have:

`// +build darwin freebsd linux openbsd windows`

Otherwise, LGTM.

monicasarbu:
@andrewkroh Yes, it looks like gopsutil has support for all these platforms, but sometimes a bit different implementation.

andrewkroh:
The fact that the gopsutil implementation uses `os/exec` to collect certain metrics concerns me.

monicasarbu:
@andrewkroh Yes, I agree. I will close this PR in favor of https://github.com/elastic/gosigar/pull/71.

PR #4220, opened 2017-05-05
Fix non-deterministic vsphere datastore tests

andrewkroh:
The data used in the test appears to be simulated using `df -k`. So the assertions needed to be less strict.

And on windows, `df -k` doesn’t exist so the values are 0 which caused a panic due to a `panic: runtime error: integer divide by zero`.

Failures: https://beats-ci.elastic.co/job/elastic+beats+master+tests/beat=metricbeat,label=linux/251/console

### DIFF
--- a/metricbeat/module/vsphere/datastore/datastore.go
+++ b/metricbeat/module/vsphere/datastore/datastore.go
@@ -2,13 +2,13 @@ package datastore
 
 import (
 	"context"
-	"errors"
 	"net/url"
 
 	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/logp"
 	"github.com/elastic/beats/metricbeat/mb"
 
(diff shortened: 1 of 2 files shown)


metricbeat/module/vsphere/datastore/datastore.go
```diff
@@ -61,50 +60,46 @@ func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
 }
 
 func (m *MetricSet) Fetch() ([]common.MapStr, error) {
-
 	f := find.NewFinder(m.Client, true)
-	if f == nil {
```
ruflin: is this not needed anymore?

andrewkroh: I looked at the implementation in `find.NewFinder()` and it's not possible for it to return `nil`.

PR #4222, opened 2017-05-05
Fix freebsd tests by setting hostname

ruflin:
On freebsd the filebeat tests fail because no hostname is set. This change sets the hostname to `beats-tester`.


exekias:
LGTM if it works, in theory freebsd.vm.hostname  = "beats-tester" should do the job too

ruflin:
Will change it to your solution as its much nicer. Makes me think if we should configure it for all hosts to be the same?

exekias:
It would make sense to me, yes

ruflin:
@andrewkroh WDYT?
@exekias Could I just configure it at the top once with `config.vm.hostname` and it will be applied to all hosts?

exekias:
Yes, that should work in theory

PR #4223, opened 2017-05-05
Fix openbsd build by ignoring GID test

ruflin:


libbeat/common/file/fileinfo_test.go
```diff
@@ -1,4 +1,8 @@
-// +build !windows
+// +build !windows !openbsd
+
+// Test for openbsd are excluded here as info.GID() returns 0 instead of the actual value
+// As the code does not seem to be used in any of the beats, this should be ok
+// Still it would be interesting to know what it returns 0.
```
exekias: it would be interesting to know **why** it returns 0.

PR #4225, opened 2017-05-05
Deprecate document_type in filebeat 5.5

ruflin:
`_type` is removed in elasticsearch 6.0 and `document_type` is removed in filebeat 6.0. Use `fields` instead.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -73,6 +73,7 @@ https://github.com/elastic/beats/compare/v5.3.0...master[Check the HEAD diff]
 *Affecting all Beats*
 
 *Filebeat*
+- Deprecate `document_type` prospector config option as _type is removed in elasticsearch 6.0. Use fields instead. {pull}4225[4225]
 
 *Heartbeat*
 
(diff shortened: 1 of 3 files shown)


tsg:
Can we also add a `log.Deprecate` if the config has a value?

ruflin:
I left it out because in case of 10 prospectors, the message would be printed out 10 times and if someone is using reloading even more. But we could use sync.Once perhaps to eliminate the problem.

tsg:
@ruflin i'd prefer to do that, because for the migration guide we're going to recommend people to check the Deprecation logs before upgrading, so it would be good to catch this. Even printing it multiple times wouldn't be bad, but I think sync.Once should be pretty easy.

ruflin:
Ok, will add it.

ruflin:
New version pushed with deprecation message. It is now only happening once as this is part of the harvester and otherwise we would see this quite often.

oldcodeoberyn:
when document_type is deprecated, in the kafka output, how to use %{[type]} for conditional topics?

PR #4224, opened 2017-05-05
Fix grok pattern in filebeat module system/auth without hostname

ruflin:
Some log lines like `Feb  9 21:20:08  sshd[8317]: last message repeated 2 times` do not contain a hostname. This change in the grok pattern makes the hostname optional.

* Make system module tests more verbose on error

PR #4228, opened 2017-05-05
Fix grok pattern in filebeat module system/auth without hostname (#4224)

ruflin:
Some log lines like `Feb  9 21:20:08  sshd[8317]: last message repeated 2 times` do not contain a hostname. This change in the grok pattern makes the hostname optional.

* Make system module tests more verbose on error
(cherry picked from commit 598672b0303ca96f56e5a2b5bd95e5544818c369)


tsg:
This was a bckport of #4224.

PR #4231, opened 2017-05-05
Add process summary metrics

tsg:
As we're enabling `include_top_n` by default, a few visualizations from the
Metricbeat-processes list were no longer correct (as they aggregate only a sample of the
data).

This PR adds a new `process_summary` metricset that adds these metrics. The fields are namespaced under `process.summary`.

Remaining TODOs:

* [x] system test for the new metricset.
* [x] changelog
* [x] update Kibana dashboards

This is required for #4112.


metricbeat/module/system/process/_meta/fields.yml
```diff
@@ -487,3 +487,33 @@
               description: >
                 Total number of I/O operations performed on all devices
                 by processes in the cgroup as seen by the throttling policy.
+
+    - name: summaries
```
ruflin: I would make this singular as it is only 1 summary and not multiple.

metricbeat/module/system/process/process.go
```diff
@@ -49,6 +49,7 @@ func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
 		CPUTicks     bool             `config:"cpu_ticks"`
 		CacheCmdLine bool             `config:"process.cmdline.cache.enabled"`
 		IncludeTop   includeTopConfig `config:"process.include_top_n"`
+		Summaries    bool             `config:"process.summaries"`
```
ruflin: `process.summary.enabled`? In case we add more options to it in the future

metricbeat/module/system/process/helper.go
```diff
@@ -373,6 +383,11 @@ func (procStats *ProcStats) GetProcStats() ([]common.MapStr, error) {
 	}
 	procStats.ProcsMap = newProcs
 
+	var summaries *processSummaries
+	if procStats.Summaries {
+		summaries = procStats.getSummaries(processes)
```
andrewkroh: I'm a little concerned that users might expect the summary to unfiltered. As written it looks like the summary only includes processes that matched the given regular expressions.

If we keep it this way, I feel like we should include info about what we matched against in the summary. This would be needed to distinguish between two separate summary events if the `system/process` metricset is used more than once.

tsg: Ah, you are right, it should be unfiltered, I didn't realize that the regexps are already applied at that point.

tsg:
@ruflin I agree that a separate metricset would be somehow cleaner, although I don't like it when two metricsets have very similar names (process and processes) because it's unclear for the user which one does what.  We already have that situation with `fststat` and `filesystem`, and I find it confusing.

Another concern is that with a separate metricset we'd need to poll the process data twice, from both metricsets. Or is there a way for two metricsets to work on the same data?

tsg:
Ok, I have [tried](https://github.com/elastic/beats/pull/4231/commits/f3c38672505e1094f75493c928c19486dddb2311) the approach of having its own metricset and I have to admit it feels a lot simpler, so I'm now also leaning towards that option. I named the metricset `process_summary`, but I'm open to suggestions :).

metricbeat/module/system/process_summary/process_summary.go
```diff
@@ -0,0 +1,102 @@
...
+		idle     int
+		stopped  int
+		zombie   int
+	}
+
+	for _, pid := range pids {
+		state := sigar.ProcState{}
+		err = state.Get(pid)
+		if err != nil {
+			return nil, fmt.Errorf("Error getting process state for pid=%d: %v", pid, err)
```
andrewkroh: One bad apple will spoil the bunch. This will cause problems on Windows, because pid=0 and `csrss.exe` processes cannot be accessed. But this could also affect other OSes if a process exits in between the `Pids()` and the `state.Get(pid)` calls.

One solution would be to treat these processes as `unknown` in the state summary.

metricbeat/module/system/process_summary/process_summary.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+// Fetch methods implements the data gathering and data conversion to the right format
+// It returns the event which is then forward to the output. In case of an error, a
+// descriptive error must be returned.
+func (m *MetricSet) Fetch() (common.MapStr, error) {
+
+	pids, err := process.Pids()
+	if err != nil {
+		logp.Warn("Getting the list of pids: %v", err)
```
andrewkroh: I think the error should only be handled once. I'm advocating against logging the error also returning it.

metricbeat/module/system/process_summary/process_summary.go
```diff
@@ -0,0 +1,102 @@
...
+		case 'R':
+			summary.running += 1
+		case 'D':
+			summary.idle += 1
+		case 'T':
+			summary.stopped += 1
+		case 'Z':
+			summary.zombie += 1
+		default:
+			logp.Err("Unknown state <%v> for process with pid %d", state.State, pid)
```
andrewkroh: This could be another reason for having an `unknown` state in the summary. Without it, the sum of the counts of each state will not equal the `total` in this case.

metricbeat/module/system/process_summary/process_summary.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+// New create a new instance of the MetricSet
+// Part of new is also setting up the configuration by processing additional
+// configuration entries if needed.
+func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
+
+	config := struct{}{}
+
+	if err := base.Module().UnpackConfig(&config); err != nil {
```
andrewkroh: These can be removed since there is no config.

metricbeat/module/system/process_summary/process_summary.go
```diff
@@ -0,0 +1,102 @@
...
+
+// Fetch methods implements the data gathering and data conversion to the right format
+// It returns the event which is then forward to the output. In case of an error, a
+// descriptive error must be returned.
+func (m *MetricSet) Fetch() (common.MapStr, error) {
+
+	pids, err := process.Pids()
+	if err != nil {
+		logp.Warn("Getting the list of pids: %v", err)
+		return nil, err
```
andrewkroh: I recommend wrapping the returned error to provide some addition context, like `errors.Wrap(err, "failed to fetch list of PIDs")`.

ruflin:
Agree on the naming part, it's somehow ugly. I like the suggestion of `process_summary` and we could still put it under `process.summary` as the docs namespace. Will we allow the process_summary to also be filtered or exclude processes?

ruflin:
Talking about naming: In case we agree on `process_summary`, we should probably rename `fsstats` to `filesystem_summary` for consitency. Agree that it is suboptimal / confusing at the moment.

metricbeat/module/system/process_summary/process_summary.go
tsg: Oh, that copy pasting from the `process` metricset :) I'll fix it in both places.

metricbeat/module/system/process_summary/process_summary.go
tsg: Initially I thought that we don't need `unknown` because it can be computed on the UI side, but I think we can be explicit about it.

tsg:
> Will we allow the process_summary to also be filtered or exclude processes?

I'd say no, for that people should use the `process` metricset and do the analytics in Kibana.

tsg:
jenkins, retest it

metricbeat/module/system/process_summary/_meta/data.json
```diff
@@ -0,0 +1,24 @@
...
+        "rtt": 115
+    },
+    "system": {
+        "process_summary": {
+            "idle": 0,
+            "running": 225,
+            "sleeping": 0,
+            "stopped": 0,
+            "total": 355,
+            "unknown": 130,
```
andrewkroh: That's a lot of unknowns. Is that because of permissions on `/proc/pid/*`?

metricbeat/module/system/process_summary/_meta/docs.asciidoc
```diff
@@ -0,0 +1,3 @@
+=== system process_summary MetricSet
```
andrewkroh: `system` should be capitalized and I think we should change `process_summary` to "Process Summary". I think it would fit in better on the [metricset listing](https://www.elastic.co/guide/en/beats/metricbeat/current/metricbeat-metricset-system-diskio.html).

metricbeat/module/system/process_summary/_meta/data.json
tsg: I think so, it was generated on my mac without sudo.

metricbeat/module/system/process_summary/process_summary.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/elastic/beats/metricbeat/module/system/process"
+	"github.com/pkg/errors"
+
+	sigar "github.com/elastic/gosigar"
+)
+
+// init registers the MetricSet with the central registry.
+// The New method will be called after the setup of the module and before starting to fetch data
+func init() {
+	if err := mb.Registry.AddMetricSet("system", "process_summary", New, parse.EmptyHostParser); err != nil {
```
andrewkroh: What if the metricset name were `process.summary` instead of `process_summary`, would that alleviate some of the concerns about the name differing from the data model?

ruflin: @andrewkroh I like the idea. Would this cause any issues somewhere else?

And interesting part of this could be that one could also enable `process.*` metricsets in the future. This becomes especially interesting in the kubernetes case.

tsg: Interesting idea, but that currently doesn't work:

```
2017/05/11 10:25:00.264381 metricbeat.go:31: INFO Register [ModuleFactory:[docker, mongodb, mysql, postgresql, system], MetricSetFactory:[apache/status, audit/kernel, ceph/cluster_disk, ceph/cluster_health, ceph/monitor_health, ceph/pool_disk, couchbase/bucket, couchbase/cluster, couchbase/node, docker/container, docker/cpu, docker/diskio, docker/healthcheck, docker/image, docker/info, docker/memory, docker/network, dropwizard/collector, elasticsearch/node, elasticsearch/node_stats, golang/expvar, golang/heap, haproxy/info, haproxy/stat, http/json, jolokia/jmx, kafka/consumergroup, kafka/partition, kibana/status, kubernetes/container, kubernetes/node, kubernetes/pod, kubernetes/system, kubernetes/volume, memcached/stats, mongodb/dbstats, mongodb/status, mysql/status, nginx/stubstatus, php_fpm/pool, postgresql/activity, postgresql/bgwriter, postgresql/database, prometheus/collector, prometheus/stats, redis/info, redis/keyspace, system/core, system/cpu, system/diskio, system/filesystem, system/fsstat, system/load, system/memory, system/network, system/process, system/process.summary, vsphere/datastore, vsphere/host, vsphere/virtualmachine, zookeeper/mntr]]
panic: name process already used
```

I think this brings up the question (again), if we want to support metricset hierarchies.

ruflin: I would separate the two things. I think we should support dots in metricset names which does not necessarly mean there is a hierachy in the code structure. That it happens on the data side is nice of course.

andrewkroh: So the panic is caused by a naming collision the happens when registering the metrics key for `metricbeat.system.process.summary` with libbeat monitoring. Since the `process` metricset is already registered we end up with a collision. We could work around the issue with a change to this [line](https://github.com/andrewkroh/beats/blob/d3d14072be53041835210ebba4d65bfc54bed204/metricbeat/mb/module/wrapper.go#L362-L362) to remove dots from the key name. This would "flatten" the metrics reported about the metricsets and simplify consuming those metrics in any sort of UI.

    key := fmt.Sprintf("metricbeat.%s.%s", module, strings.Replace(name, ".", "_", -1))

But @tsg's concern that the usage of dots is overloaded was spot on.

tsg:
Updated to use `process.summary` in fields, but keep `process_summary` as the metricset name. Rebased and ready for new reviews/merging.

ruflin:
@tsg I think you have to use ["..."] in the tests there:
```
  File "/go/src/github.com/elastic/beats/metricbeat/tests/system/test_system.py", line 318, in test_process_summary
    summary = evt["system"]["process.summary"]
KeyError: 'process.summary'
```

ruflin:
jenkins, retest it

metricbeat/module/system/process_summary/process_summary_test.go
```diff
@@ -0,0 +1,37 @@
...
+func TestFetch(t *testing.T) {
+	f := mbtest.NewEventFetcher(t, getConfig())
+	event, err := f.Fetch()
+	assert.NoError(t, err)
+	assert.Contains(t, event, "sleeping")
+	assert.Contains(t, event, "running")
+	assert.Contains(t, event, "idle")
+	assert.Contains(t, event, "stopped")
+	assert.Contains(t, event, "zombie")
+	assert.Contains(t, event, "unknown")
```
andrewkroh: Missing `total`. 

And how about a test to ensure the sum of the various states equals `total`?

metricbeat/tests/system/test_system.py
```diff
@@ -292,6 +292,38 @@ def test_network(self):
...
+        for evt in output:
+            self.assert_fields_are_documented(evt)
+
+            summary = evt["system"]["process"]["summary"]
+            assert isinstance(summary["sleeping"], int)
+            assert isinstance(summary["running"], int)
+            assert isinstance(summary["idle"], int)
+            assert isinstance(summary["stopped"], int)
+            assert isinstance(summary["zombie"], int)
+            assert isinstance(summary["unknown"], int)
```
andrewkroh: Missing `total`.

PR #4233, opened 2017-05-05
Use kibana names when exporting visualization jsons

exekias:
Instead of current internal id, this should make resulting files more
human readable, ie:

 _meta/kibana/dashboard/Kubernetes overview.json
 _meta/kibana/visualization/Kubernetes - Available pods per deployment.json
 _meta/kibana/visualization/Kubernetes - Available pods.json
 _meta/kibana/visualization/Kubernetes - CPU usage by node.json
 _meta/kibana/visualization/Kubernetes - Deployments.json
 ...

### DIFF
--- a/dev-tools/export_dashboards.py
+++ b/dev-tools/export_dashboards.py
@@ -80,7 +80,8 @@ def SaveJson(doc_type, doc, output_directory):
     if not os.path.exists(dir):
         os.makedirs(dir)
     # replace unsupported characters
-    filepath = os.path.join(dir, re.sub(r'[\>\<:"/\\\|\?\*]', '', doc['_id']) + '.json')
+    filename = '-'.join(re.sub(r'[\>\<:"/\\\|\?\*]', '', doc['_source']['title']).split(' '))
+    filepath = os.path.join(dir, filename + '.json')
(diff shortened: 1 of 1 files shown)


andrewkroh:
WDYT about removing the spaces (or replacing them)?

monicasarbu:
LGTM. I think it's good to display the title of the dashboard as it is, without removing the spaces, especially as you ca pass a regular expression to the export_dashboards script.

monicasarbu:
This is the name of the json file where the dashboard/visualization/search is written locally. In the previous solution, I am removing all strange characters (`'[\>\<:"/\\\|\?\*]'`) from the name of the json file. What do you think about replacing the space with "-" in the name? I find it easier to read.

exekias:
:+1: to both, I added strange characters stripping and replaced spaces with `-`

exekias:
:hand: Do not merge, I just saw that importer script expects the files to be named with ID, this change would require to rename all current visualizations: WDYT? I can do it if we agree

https://github.com/elastic/beats/blob/master/libbeat/dashboards/importer.go#L170

exekias:
Closing this one as it would require to change all existing dashboards, we have to think more about how we want to deal with them

PR #4235, opened 2017-05-05
More test fixes

andrewkroh:
A follow up to #4216 to fix some more tests on Windows.

These are the failures being addressed (except the Filebeat ones, those need specialized attention): https://beats-ci.elastic.co/job/elastic+beats+master+tests-windows/4/testReport/

- Using single quotes around Windows paths 	e86d38b
  
  The thrift test config used double quotes around Windows path separators and this was interpreted incorrectly in YAML parsing.

- Use logp.Beta or logp.Experimental in metricsets 	077d508

  And in system tests, centralize the logic for asserting that there are no ERR or WARN in logs.

  Filter out errors about “The service process could not connect to the service controller” that occur when testing on Jenkins where Jenkins itself is running as a service. This confuses the Beat because it thinks that it is running as service, but it’s not.

- Rename TestBadCondition to TestConditions  ab84099

  This test doesn’t actually test any bad conditions. Plus there is another test in the same directory with the name TestBadCondition.

### DIFF
--- a/metricbeat/tests/system/metricbeat.py
+++ b/metricbeat/tests/system/metricbeat.py
@@ -52,3 +52,15 @@ class BaseTest(TestCase):
                 fields[key] = self.de_dot(fields[key])
 
         return fields
+
+    def assert_no_logged_warnings(self):
+        """
+        Assert that the log file contains no ERR or WARN lines.
+        """
+        log = self.get_log()
+        log = log.replace("WARN EXPERIMENTAL", "")
+        log = log.replace("WARN BETA", "")
+        # Jenkins runs as a Windows service and when Jenkins executes theses
+        # tests the Beat is confused since it thinks it is running as a service.
+        log = log.replace("ERR Error: The service process could not connect to the service controller.", "")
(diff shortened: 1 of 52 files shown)


metricbeat/tests/system/metricbeat.py
```diff
@@ -52,3 +52,15 @@ def de_dot(self, existing_fields):
                 fields[key] = self.de_dot(fields[key])
 
         return fields
+
+    def assert_no_logged_warnings(self):
+        """
+        Assert that the log file contains no ERR or WARN lines.
+        """
+        log = self.get_log()
+        log = log.replace("WARN EXPERIMENTAL", "")
```
ruflin: nice

andrewkroh: Let me fix it up to get it back to the three separate commits. Then you can use "rebase and merge".

andrewkroh:
Let me fix it up to get it back to the three separate commits. Then you can use "rebase and merge".

PR #4253, opened 2017-05-08
Add `kube-state-metrics` based metricsets

exekias:
Add k8s metrics extracted from kube-state-metrics services. This adds info specially useful about:

- Status of deployments, pods and replicasets
- Resource limits, requests and usage by pods
- Node available resources

I have updated existing dashboard with this kind of info

image


metricbeat/module/kubernetes/state_container/_meta/data.json
```diff
@@ -0,0 +1,45 @@
...
+        "requests": {
+          "cpu": {
+            "nanocores": 5000000
+          },
+          "memory": {
+            "bytes": 52428800
+          }
+        }
+      },
+      "status": {
```
ruflin: The name implies `state` and here it is `status`?

metricbeat/module/kubernetes/state_container/_meta/data.json
```diff
@@ -0,0 +1,45 @@
...
+  },
+  "kubernetes": {
+    "container": {
+      "id": "docker://91fdd43f6b1b4c3dd133cfca53e0b1210bc557c2ae56006026b5ccdb5f52826f",
+      "image": "gcr.io/google-containers/kube-addon-manager:v6.3",
+      "name": "kube-addon-manager",
+      "resource": {
+        "requests": {
+          "cpu": {
+            "nanocores": 5000000
```
ruflin: Is this something that will change over time? If we call it `state_container` all these fields which are not shared should be under one namespace.

metricbeat/module/kubernetes/state_container/data.go
```diff
@@ -0,0 +1,89 @@
...
+
+const (
+	// Nanocores conversion 10^9
+	nanocores = 1000000000
+)
+
+func eventMapping(families []*dto.MetricFamily) ([]common.MapStr, error) {
+	eventsMap := map[string]common.MapStr{}
+	for _, family := range families {
+		//fmt.Println(family)
```
ruflin: Should be removed.

metricbeat/module/kubernetes/state_container/data.go
```diff
@@ -0,0 +1,89 @@
...
+
+	dto "github.com/prometheus/client_model/go"
+)
+
+const (
+	// Nanocores conversion 10^9
+	nanocores = 1000000000
+)
+
+func eventMapping(families []*dto.MetricFamily) ([]common.MapStr, error) {
```
ruflin: I would kind of hope instead of adding dependency on prometheus here, that this can "depend" on the prometheus module / metricset.

metricbeat/module/kubernetes/state_container/data.go
```diff
@@ -0,0 +1,89 @@
...
+					event.Put("container.status.phase", "terminate")
+				}
+
+			case "kube_pod_container_status_waiting":
+				if metric.GetGauge().GetValue() == 1 {
+					event.Put("container.status.phase", "waiting")
+				}
+
+			default:
+				// Ignore unknown metric
```
ruflin: Sounds like an error or at least worth logging as debug?

metricbeat/module/kubernetes/state_container/state_container.go
```diff
@@ -0,0 +1,72 @@
+package state_container
```
ruflin: I'm starting to think this should be `container_state` as the namespace is probably `container.state`?

metricbeat/module/kubernetes/state_container/data.go
andrewkroh: This should be removed.

metricbeat/module/kubernetes/state_deployment/_meta/data.json
```diff
@@ -0,0 +1,29 @@
...
+      "paused": false,
+      "replicas": {
+        "available": 0,
+        "desired": 1,
+        "unavailable": 1,
+        "updated": 1
+      }
+    },
+    "namespace": "default",
+    "state_deployment": {}
```
ruflin: This should not exist. Probably you should change the namespace of the metricset instead of putting it under the module namespace.

metricbeat/module/kubernetes/state_deployment/data.go
```diff
@@ -0,0 +1,64 @@
+package state_deployment
```
ruflin: Why not call the metricset `deployment`if you put it under the `deployment` namespace?

metricbeat/module/kubernetes/state_deployment/data.go
```diff
@@ -0,0 +1,64 @@
...
+			if deployment == "" {
+				continue
+			}
+			event, ok := eventsMap[deployment]
+			if !ok {
+				event = common.MapStr{}
+				eventsMap[deployment] = event
+			}
+			switch family.GetName() {
+			case "kube_deployment_metadata_generation":
```
ruflin: I assume if you use prometheus module and get a mapStr back, you could use `schema` to do this mappings.

metricbeat/module/kubernetes/state_deployment/state_deployment.go
```diff
@@ -0,0 +1,72 @@
...
+
+	logp.Beta("The kubernetes state_deployment metricset is experimental")
+
+	if err := base.Module().UnpackConfig(&config); err != nil {
+		return nil, err
+	}
+
+	return &MetricSet{
+		BaseMetricSet: base,
+		counter:       1,
```
ruflin: is the counter needed?

metricbeat/module/kubernetes/state_node/_meta/data.json
```diff
@@ -0,0 +1,43 @@
+{
+  "@timestamp": "2017-05-04T17:47:20.829Z",
+  "beat": {
+    "hostname": "X1",
+    "name": "X1",
+    "version": "6.0.0-alpha1"
+  },
+  "kubernetes": {
+    "node": {
+      "allocatable": {
```
ruflin: I would put all these under `node.state`.

metricbeat/module/kubernetes/state_node/state_container.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+// MetricSet type defines all fields of the MetricSet
+// As a minimum it must inherit the mb.BaseMetricSet fields, but can be extended with
+// additional entries. These variables can be used to persist data or configuration between
+// multiple fetch calls.
+type MetricSet struct {
+	mb.BaseMetricSet
+	http    *helper.HTTP
+	counter int
```
ruflin: counter needed?

metricbeat/module/kubernetes/state_pod/_meta/data.json
```diff
@@ -0,0 +1,32 @@
...
+  "kubernetes": {
+    "namespace": "kube-system",
+    "node": {
+      "name": "minikube"
+    },
+    "pod": {
+      "host_ip": "192.168.99.100",
+      "ip": "172.17.0.4",
+      "name": "kube-state-metrics-1303537707-7ncd1",
+      "status": {
```
ruflin: state vs status?

metricbeat/module/kubernetes/state_pod/_meta/data.json
```diff
@@ -0,0 +1,32 @@
...
+        "phase": "Running",
+        "ready": "true",
+        "scheduled": "true"
+      }
+    },
+    "state_pod": {}
```
ruflin: empty namespace

metricbeat/module/kubernetes/state_replicaset/data.go
```diff
@@ -0,0 +1,59 @@
+package state_replicaset
```
ruflin: could we call this `replicaset`

metricbeat/module/kubernetes/state_replicaset/state_replicaset.go
```diff
@@ -0,0 +1,72 @@
...
+
+	logp.Beta("The kubernetes state_replicaset metricset is experimental")
+
+	if err := base.Module().UnpackConfig(&config); err != nil {
+		return nil, err
+	}
+
+	return &MetricSet{
+		BaseMetricSet: base,
+		counter:       1,
```
ruflin: remove counter

metricbeat/module/kubernetes/state_container/state_container.go
```diff
@@ -0,0 +1,72 @@
...
+
+// New create a new instance of the MetricSet
+// Part of new is also setting up the configuration by processing additional
+// configuration entries if needed.
+func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
+	config := struct{}{}
+
+	logp.Beta("The kubernetes state_container metricset is experimental")
+
+	if err := base.Module().UnpackConfig(&config); err != nil {
```
andrewkroh: This can be dropped since no config is used.

metricbeat/module/kubernetes/util/prometheus.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	"fmt"
+
+	"github.com/elastic/beats/metricbeat/helper"
+
+	dto "github.com/prometheus/client_model/go"
+	"github.com/prometheus/common/expfmt"
+)
+
+func GetFamilies(http *helper.HTTP) ([]*dto.MetricFamily, error) {
```
ruflin: Should this be a `prometheus` helper under the `helper` namespace? Or can you use the `prometheus.collector` code here?

metricbeat/module/kubernetes/state_pod/state_pod.go
```diff
@@ -0,0 +1,72 @@
...
+	counter int
+}
+
+// New create a new instance of the MetricSet
+// Part of new is also setting up the configuration by processing additional
+// configuration entries if needed.
+func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
+	config := struct{}{}
+
+	logp.Beta("The kubernetes state_pod metricset is experimental")
```
andrewkroh: This is a `logp.Experimental()`.

metricbeat/module/kubernetes/vendor/vendor.json
```diff
@@ -0,0 +1,43 @@
+{
+	"comment": "",
+	"ignore": "test",
```
ruflin: you need to add `github.com/elastic/beats` here as otherwise when you remove all the beats dependencies will be added to when doing an update.

metricbeat/module/kubernetes/state_container/_meta/fields.yml
```diff
@@ -0,0 +1,49 @@
...
+          fields:
+            - name: memory.bytes
+              type: long
+              format: bytes
+              description: >
+                Container requested memory in bytes
+            - name: cpu.nanocores
+              type: long
+              description: >
+                Container CPU requested nanocores
```
andrewkroh: So what is a nanocore? Nanoseconds of CPU time per core?

exekias: It's the other way around, in this case it's number of requested (reserved) full CPU cores. So 1 Core would mean the container has 1 full core reserved for itself, as long as it's alive.

metricbeat/module/kubernetes/state_container/data.go
exekias: That was my initial intention, but I have 2 issues with that:

- Prometheus module mangles original data on it's way, and it's not really aware of the format of these metrics, it does something generic. So I need to go over the resulting MapStr again
- If we do a change in prometheus module it may break this, having that I'm not using its format

Something I thought about was to move all prometheus code to a common helper (like HTTP), WDYT about that? At the end it looked like overkill but I'm ok with doing it

metricbeat/module/kubernetes/util/prometheus.go
exekias: ok to the helper, prometheus collector issue explained in a previous comment, but I can update it to share the helper

metricbeat/module/kubernetes/state_container/_meta/data.json
exekias: I'm using `kube-state-metrics` nomenclature here, check https://github.com/elastic/beats/pull/4253/files/0f5f98da5eb5acf11055a8d8cf551f7b4eb1ed0e#diff-c70b508c24452e40e23c98c944027e25R54 to see the metric family names


This is how I would explain it:

State (from kube-state-metrics) means that we get access to the global state of the cluster (think of it as an snapshot in time). While this `status` metrics are one of the specific metrics of a container.

metricbeat/module/kubernetes/state_container/_meta/fields.yml
exekias: BTW this metric is also present as a usage metric: https://github.com/elastic/beats/blob/master/metricbeat/module/kubernetes/container/_meta/fields.yml#L21

In that case it's the accumulated value over time (I use derivative in graphs to extract the used nancores per bucket)

exekias: A common scenario is to check requested vs limits vs real usage. Ideally you would put some sane limits based on real usage of a given container

metricbeat/module/kubernetes/state_container/data.go
ruflin: +1 on the helper. Can also be done in a follow up PR. But I think it's good to have it around as this will not be the last metricset using prometheus.

metricbeat/module/kubernetes/state_container/_meta/data.json
ruflin: got it. thanks.

ruflin: If we namespace this under `state` it probably becomes more obvious (see data structure comments).

metricbeat/module/kubernetes/state_replicaset/data.go
exekias: I wanted to keep the same format for all `kube-state-metrics` metricsets for one reason, the `hosts` setting is incompatible between `kubelet` and `kube-state-metrics`, you have to choose, and it's nice to have that clear from the metricset name, to avoid confusion.

Check https://github.com/exekias/beats/blob/d091ce76089c44e5fd872661d529932aa92633df/metricbeat/module/kubernetes/_meta/config.yml for an example of possible configurations

metricbeat/module/kubernetes/state_node/_meta/data.json
exekias: This sounds good to me, would it be possible to use `node.state` as metricset name?

ruflin: yes. see: https://github.com/elastic/beats/pull/4142

metricbeat/module/kubernetes/state_replicaset/data.go
ruflin: I see the point. Leads me to think it should be it's own module but we concluded this discussion :-) Don't have a better solution at the moment :-(

exekias: Yeah, I'm not 100% happy with this either, we can see how it goes and perhaps improve it in the future

metricbeat/module/kubernetes/state_node/_meta/data.json
exekias: I've adapted the code to use mb.Namespace (really cool!) but I still don't see the need of adding `state` to the metric key. Let's talk about this tomorrow

metricbeat/helper/prometheus.go
```diff
@@ -0,0 +1,50 @@
+package helper
+
+import (
+	"fmt"
+
+	"github.com/elastic/beats/metricbeat/mb"
+	dto "github.com/prometheus/client_model/go"
```
ruflin: Here a newline should be added.

metricbeat/helper/prometheus.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/prometheus/common/expfmt"
+)
+
+// Prometheus helper retrieves prometheus formatted metrics
+type Prometheus struct {
+	HTTP
```
ruflin: I like that, @urso will not like it too much ;-)

metricbeat/metricbeat.full.yml
```diff
@@ -238,12 +238,29 @@ metricbeat.modules:
...
+    - container
+    - volume
   period: 10s
   hosts: ["localhost:10255"]
 
+# State metrics from kube-state-metrics service:
+- module: kubernetes
+  enabled: false
+  metricsets:
+    - state_node
```
ruflin: Based on a comment from @andrewkroh in an other PR I'm hoping we can make this state.node etc. in the future (not in this PR).

metricbeat/module/kubernetes/state_container/data.go
```diff
@@ -0,0 +1,88 @@
...
+)
+
+const (
+	// Nanocores conversion 10^9
+	nanocores = 1000000000
+)
+
+func eventMapping(families []*dto.MetricFamily) ([]common.MapStr, error) {
+	eventsMap := map[string]common.MapStr{}
+	for _, family := range families {
```
ruflin: Is this still needed or can it be moved now to the helper?

exekias: I think it's pretty custom for this case, so we need to keep it.

I want to see what future metricsets using the helper have in common, perhaps we can still extract some helpers there

exekias:
jenkins, retest this please

andrewkroh:
^ You're so polite to the robots 🤖 . I guess you are hoping for favorable treatment when they take over.

exekias:
image

exekias:
jenkins, test it

libbeat/tests/system/beat/beat.py
```diff
@@ -426,7 +426,7 @@ def extract_fields(doc_list, name):
 
         # TODO: Make fields_doc path more generic to work with beat-generator
         with open(fields_doc, "r") as f:
-            path = os.path.abspath(os.path.dirname(__file__) + "../../../../_meta/fields.common.yml")
+            path = os.path.abspath(os.path.dirname(__file__) + "../../../../_meta/fields.generated.yml")
```
ruflin: Does the generated exist in all beats?

metricbeat/docker-compose.yml
```diff
@@ -51,6 +52,8 @@ services:
       jolokia:        { condition: service_healthy }
       kafka:          { condition: service_healthy }
       kibana:         { condition: service_healthy }
+      kubernetes:     { condition: service_healthy }
+      kube-state:     { condition: service_started }
```
ruflin: Interesting that this one is set to started and not healthy

metricbeat/module/kubernetes/_meta/config.yml
```diff
@@ -1,6 +1,23 @@
...
+- module: kubernetes
+  enabled: false
+  metricsets:
+    - state_node
+    - state_deployment
+    - state_replicaset
+    - state_pod
+    - state_container
+  period: 10s
+  hosts: ["kube-state-metrics:8080"]
```
ruflin: Does this have a "common" ip or dns name in a kubernetes environment?

metricbeat/tests/system/test_kubernetes.py
```diff
@@ -0,0 +1,84 @@
...
+        log = self.get_log()
+        self.assertNotRegexpMatches(log.replace("WARN BETA", ""), "ERR|WARN")
+
+        output = self.read_output_json()
+        self.assertEqual(len(output), expected_events)
+        evt = output[0]
+
+        self.assertItemsEqual(self.de_dot(KUBERNETES_FIELDS), evt.keys(), evt)
+
+        self.assert_fields_are_documented(evt)
```
ruflin: nice

libbeat/tests/system/beat/beat.py
ruflin: I think this could be the reason that the appveyor builds fail.

metricbeat/module/kubernetes/_meta/config.yml
exekias: Yes, normally it's that: https://github.com/kubernetes/kube-state-metrics/blob/master/kubernetes/kube-state-metrics-service.yaml#L4

metricbeat/docker-compose.yml
exekias: kube-state-metrics doesn't provide health service, but as soon as it starts it should work so we can live with this it seems

libbeat/tests/system/beat/beat.py
exekias: I've added a fallback to common.yml, let's see if appveyor likes that

exekias: it worked, thanks!

PR #4257, opened 2017-05-09
Fallback on LevelRaw If the Level is not in the RenderingInfo section of the event

LucasArona:
Applies to Windows Vista and above only.

https://discuss.elastic.co/t/event-fields-missing-if-renderinginfo-is-empty/84709

### DIFF
--- a/winlogbeat/eventlog/wineventlog.go
+++ b/winlogbeat/eventlog/wineventlog.go
@@ -11,7 +11,6 @@ import (
 	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/logp"
 	"github.com/elastic/beats/winlogbeat/sys"
-	"github.com/elastic/beats/winlogbeat/sys/wineventlog"
 	win "github.com/elastic/beats/winlogbeat/sys/wineventlog"
 	"github.com/joeshaw/multierror"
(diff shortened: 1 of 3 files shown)


winlogbeat/eventlog/wineventlog.go
```diff
@@ -216,6 +217,11 @@ func (l *winEventLog) buildRecordFromXML(x []byte, recoveredErr error) (Record,
 		e.RenderErr = recoveredErr.Error()
 	}
 
+	if e.Level == "" {
+		//Let's fallback on LevelRaw if the level is not set in the RenderingInfo
```
andrewkroh: Could you please change this to 

`// Fallback on LevelRaw if the Level is not set in the RenderingInfo.`

winlogbeat/eventlog/wineventlog.go
```diff
@@ -11,6 +11,7 @@ import (
 	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/logp"
 	"github.com/elastic/beats/winlogbeat/sys"
+	"github.com/elastic/beats/winlogbeat/sys/wineventlog"
```
andrewkroh: For consistency, could you use the "aliased" import below for this package. It's just aliased to `win`.

PR #4259, opened 2017-05-09
Add 5x build for cross version testing

ruflin:
We have a 2x build to test against 2.x ES and LS. This 5x test environment is to test 6.x against 5.x.

A new jenkins job should be created to run these tests.


ruflin:
Here is the 5x jenkins job: http://build-eu-00.elastic.co/view/Beats/job/Beats-5x/

ruflin:
jenkins, retest it

PR #4262, opened 2017-05-09
Fail build if one of the docs builds fail

ruflin:
On jenkins some of the builds went green even though the doc builds failed. This should fix it.

### DIFF
--- a/libbeat/scripts/build_docs.sh
+++ b/libbeat/scripts/build_docs.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -e
+
 # Checks if docs clone already exists
 if [ ! -d "build/docs" ]; then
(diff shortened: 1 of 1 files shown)


exekias:
Would it make sense to put `set -e` on top of the file? so it breaks on **any** error

ruflin:
Yes, that is even better. Means I would not need the exit part, right?

exekias:
Correct, `set -e` would be enough, I can break docs again to test it ;P

ruflin:
Please do no merge. I introduce a doc build error to see if the jenkins build fails.

ruflin:
Jenkins successfully failed: http://build-eu-00.elastic.co/job/Beats-PR/6102/

exekias:
jenkins, retest this please

PR #4266, opened 2017-05-09
Fix repositories in docs

tsg:
To use 6.x-prerelease. This should be updated to either use the `:release-status:`
variable or changed to 6.x once we reach GA.


tsg:
CC @dedemorton for awareness.

PR #4269, opened 2017-05-09
prospector: cache resolved glob patterns during init

### LINKED ISSUE #4182: Filebeat - Error in load states when enable the recursive glob
Please help to check the problem.
Version: master
Operating System: SUSE11
Steps to Reproduce: 
1. enable the recursive glob.
2. use ** in the path as /home/cmp/**/*.log
3. start the main_test, and it will collect the log
4. restart the filebeat, when it load the exist states, it will drop the actually marched source.
<pre><code>		match, err = filepath.Match(glob, filePath)
		if err != nil {
			logp.Debug("prospector", "Error matching glob: %s", err)
			continue
		}</code></pre>
The glob is /home/cmp/**/*.log, if the filePath is /home/cmp/a/b/c.log, the match will be *false*. Then the source will be dropped, which is an unexpected behavior.
Then the filebeat will collect the log from beginning, which is duplicately.

7AC:
Before this change recursive glob patterns were resolved only at runtime, so everytime filebeat restarted it would try to recover its previous state without expanding the patterns first. With this change we expand the patterns once and for all during init, fixing both the start and the restart case.

Resolves #4182

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -34,6 +34,7 @@ https://github.com/elastic/beats/compare/v6.0.0-alpha1...master[Check the HEAD d
 
 *Filebeat*
 - Fix race condition on harvester stopping with reloading enabled. {issue}3779[3779]
+- Fix recursive glob config parsing and resolution across restarts. {pull}4269[4269]
 
 *Heartbeat*
 
--- a/filebeat/tests/system/test_prospector.py
+++ b/filebeat/tests/system/test_prospector.py
@@ -1,3 +1,5 @@
+#!/usr/bin/env python
+
 from filebeat import BaseTest
 import os
 import time
@@ -678,3 +680,47 @@ class Test(BaseTest):
         )[0]
         assert "message" not in output
         assert "offset" in output
+
+    def test_restart_recursive_glob(self):
+        """
+        Check that file reading via recursive glob patterns continues after restart
+        """
+        self.render_config_template(
+            path=os.path.abspath(self.working_dir) + "/log/**",
+            scan_frequency="1s",
+            recursive_glob=True,
+        )
+
+        testfile_dir = os.path.join(self.working_dir, "log", "some", "other", "subdir")
+        os.makedirs(testfile_dir)
+        testfile_path = os.path.join(testfile_dir, "input")
+
+        filebeat = self.start_beat()
+
+        with open(testfile_path, 'w') as testfile:
+            testfile.write("entry1\n")
+
+        self.wait_until(
+            lambda: self.output_has_message("entry1"),
+            max_timeout=10,
+            name="output contains 'entry1'")
+
+        filebeat.check_kill_and_wait()
+
+        # Append to file
+        with open(testfile_path, 'a') as testfile:
+            testfile.write("entry2\n")
+
+        filebeat = self.start_beat(output="filebeat2.log")
+
+        self.wait_until(
+            lambda: self.output_has_message("entry2"),
+            max_timeout=10,
+            name="output contains 'entry2'")
+
+        filebeat.check_kill_and_wait()
+
+
+if __name__ == '__main__':
+    import unittest
+    unittest.main()
--- a/filebeat/tests/system/test_registrar.py
+++ b/filebeat/tests/system/test_registrar.py
@@ -1,9 +1,11 @@
+#!/usr/bin/env python
 """Test the registrar"""
 
 import os
 import platform
 import time
 import shutil
+
 from filebeat import BaseTest
 from nose.plugins.skip import SkipTest
 
@@ -323,7 +325,7 @@ class Test(BaseTest):
 
     def test_restart_continue(self):
         """
-        Check that file readining continues after restart
+        Check that file reading continues after restart
         """
         self.render_config_template(
             path=os.path.abspath(self.working_dir) + "/log/input*",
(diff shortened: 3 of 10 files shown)


filebeat/prospector/prospector_log.go
```diff
@@ -42,12 +43,33 @@ func NewLog(p *Prospector) (*Log, error) {
 	return prospectorer, nil
 }
 
+func (l *Log) resolvePaths() error {
+	if l.paths != nil {
```
andrewkroh: I would avoid making a distinction between non-nil and zero length slices. This can lead to subtle programming mistakes. I would check for `len(l.paths) > 0`

filebeat/prospector/prospector_log.go
```diff
@@ -42,12 +43,33 @@ func NewLog(p *Prospector) (*Log, error) {
...
+		return nil
+	}
+	if !l.config.recursiveGlob {
+		l.paths = l.config.Paths
+		return nil
+	}
+	l.paths = []string{}
```
andrewkroh: This is unnecessary if you check the `len()` of the slice above, rather than for non-nil.

filebeat/prospector/prospector_log.go
7AC: thanks, yeah I prefer that too

filebeat/prospector/prospector_log.go
7AC: 👍

ruflin:
@7AC I didn't check the PR yet but I strongly recommend to add some system test to it to check if everything goes well during restart. I remember doing the first implementation without the "extra glob" was a bit tricky, especially on windows. We must make sure the right paths are picked up again.

filebeat/prospector/config.go
```diff
@@ -53,3 +55,25 @@ func (config *prospectorConfig) Validate() error {
 
 	return nil
 }
+
+func (config *prospectorConfig) resolvePaths() error {
+	var paths []string
+	if !config.recursiveGlob {
```
ruflin: The reason this is not working is that `recursiveGlob` is a private variable and golang can only extract values in global variables. You must set it to `RecursiveGlob` in prospectorConfig.

7AC: 👍

filebeat/tests/system/test_prospector.py
```diff
@@ -678,3 +680,48 @@ def test_prospector_filter_includefields(self):
...
+        with open(testfile_path, 'w') as testfile:
+            testfile.write("entry1\n")
+
+        self.wait_until(
+            lambda: self.output_has_message("entry1"),
+            max_timeout=10,
+            name="output contains 'entry1'")
+
+        # Wait a momemt to make sure registry is completely written
+        time.sleep(1)
```
ruflin: Sleep is very flaky with these kind of things. We normally rely on the log_contains messages here. Check some other tests for it.

filebeat/tests/system/test_registrar.py
```diff
@@ -1396,3 +1398,8 @@ def test_registrar_files_with_prospector_level_processors(self):
                 "inode": stat.st_ino,
                 "device": stat.st_dev,
             }, file_state_os)
+
+
+if __name__ == '__main__':
```
ruflin: With this addition, can you directly "run" the file without prepending it with nosetests?

libbeat/tests/system/beat/beat.py
```diff
@@ -350,6 +353,16 @@ def output_has(self, lines, output_file=None):
         except IOError:
             return False
 
+    def output_has_message(self, message, output_file=None):
```
ruflin: That will become handy in other tests too.

filebeat/tests/system/test_prospector.py
7AC: It's copy/paste from `test_registrar.py` (had the same observation btw :) I don't think I need it here anyways I'll remove it

filebeat/tests/system/test_registrar.py
7AC: yep, just need to `pip install` a couple of things

libbeat/tests/system/beat/beat.py
7AC: Yeah, I tried working with the existing APIs but checking the length wasn't enough since when I restart I want to check for the different message to show up (length still 1)

filebeat/tests/system/test_prospector.py
```diff
@@ -678,3 +680,45 @@ def test_prospector_filter_includefields(self):
...
+        with open(testfile_path, 'a') as testfile:
+            testfile.write("entry2\n")
+
+        filebeat = self.start_beat(output="filebeat2.log")
+
+        self.wait_until(
+            lambda: self.output_has_message("entry2"),
+            max_timeout=10,
+            name="output contains 'entry2'")
+
```
ruflin: filebeat must be stopped again here.

CHANGELOG.asciidoc
```diff
@@ -31,6 +31,7 @@ https://github.com/elastic/beats/compare/v6.0.0-alpha1...master[Check the HEAD d
 - Fix importing the dashboards when the limit for max open files is too low. {issue}4244[4244]
 
 *Filebeat*
+- Make recursive glob resolution work across restarts. {pull}4269[4269]
```
ruflin: Should this be "make recursive glob working"? Not sure it was working before?

filebeat/tests/system/test_prospector.py
7AC: done

CHANGELOG.asciidoc
7AC: I rephrased it

PR #4273, opened 2017-05-10
Move prospector log to its own package

ruflin:
This is the last step in reorganising the packages related to prospector and harvester. Follow up PR's will mainly focusing on abstracting out common functionality, standardise naming and have proper interfaces.

* Merge log harvester and log prospector config into one config
* Rename Log to prospector as part of the new package structure
* Move log harvester logic to log prospector package
* Keep common harvester logic in its own package
* stdin harvester still heavily depends on log harvester, needs to be split up and simplified at a later stage
* Further cleanup of Prospector interface. `Wait()` only exists as a temporary solution.

### DIFF
--- a/filebeat/prospector/log/harvester.go
+++ b/filebeat/prospector/log/harvester.go
@@ -9,7 +9,7 @@
 //  line. As soon as the line is completed, it is read and returned.
 //
 //  The stdin harvesters reads data from stdin.
-package harvester
+package log
 
 import (
 	"errors"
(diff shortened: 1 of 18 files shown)


ruflin:
jenkins, retest it

filebeat/prospector/log/harvester.go
```diff
@@ -18,7 +18,7 @@ import (
 
 	"github.com/satori/go.uuid"
 
-	"github.com/elastic/beats/filebeat/config"
+	cfg "github.com/elastic/beats/filebeat/config"
```
7AC: filebeatConfig?

PR #4279, opened 2017-05-10
Use docker-compose down to remove networks

ruflin:
Docker-compose creates a unique network for each build now. docker-compose rm -v -f did not remove the created networks. The command `docker-compose down` does also remove the networks.

This is important as docker-compose has a default limit of 31 networks.


ruflin:
jenkins, retest it

Issue #4230, opened 2017-05-05
Reduce the exported fields for memory stats


monicasarbu:
In Metricbeat, the `memory` metricset of the `system` module  exports the following fields:
- system.memory.total
- system.memory.used.bytes
- system.memory.used.pct
- system.memory.free
- system.memory.actual.used.bytes
- system.memory.actual.used.pct
- system.memory.actual.free

Starting with Kibana 5.4 (that includes Time Series Visual Builder), you are able to do calculations using the Painless script based on the exported fields, so you don't need to store everything in Elasticsearch, and only the most used fields. I propose to delete `system.memory.actual.used` as it can be calculated as `system.memory.total` - `system.memory.actual.free`.

This change affects our dashboards.

cc-ed @simianhacker

ruflin:
+1 on this change. Probably best do it for 6.0 as it is a breaking change.

monicasarbu:
Before removing `system.memory.actual.used`, we need to export `cached` and/or `buffers` as they are  used in the dashboards.

ruflin:
There has been an initial approach to fix this in https://github.com/elastic/beats/pull/4280 but it didn't make it into 6.0 so we can't remove the fields at the moment. We should discuss again on how we move forward here.

ruflin:
As this is quite an old issue and hasn't been an issue recently, going to close this one.

PR #4280, opened 2017-05-10
Enhance the system.memory fields

monicasarbu:
Trying to fix https://github.com/elastic/beats/issues/4230 and it includes the following changes:

- [ ] removes the fields `system.memory.actual.used*` 
- [ ] add `system.memory.caches.bytes`
- [ ] add `system.memory.buffers.bytes`

This PR is a breaking change, and the dashboards for the system module need to be updated.

### DIFF
--- a/metricbeat/docs/fields.asciidoc
+++ b/metricbeat/docs/fields.asciidoc
@@ -9620,23 +9620,6 @@ The percentage of used memory.
 
 
 [float]
-== actual Fields
-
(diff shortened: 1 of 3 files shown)


ruflin:
Unfortunately this change did not make it for 6.0 so we can't apply this change until 7.0. I'm closing this PR for now as we still have the open issue and continue the discussion there.

PR #4282, opened 2017-05-10
Cherry-pick various test fixes to 5.x

andrewkroh:
- Ignore permission errors in Metricbeat’s TestFileSystemList (#3562)
  
  The test can fail if some calls to statfs fail due to permission errors. For example:
  `stat("/var/lib/docker/aufs/mnt/50d0d5f599f0f19450e7649f73a0e23da1f172048e555df2b1cb78b3fefa355b", 0x7ffd2e5b8ed0) = -1 EACCES (Permission denied)`

- Less strict error matching in Winlogbeat config_test 
  Error string testing is brittle. The PR makes the test less stringent by not checking the full error message that includes the Golang stdlib error.

- Miscellaneous test fixes (#4216)
  - Fix and enable the python smoke test for heartbeat
  - Remove fmt.Printf from metricbeat ceph module
  - Fix Windows path issues in libbeat/paths tests
  - Fix ioutil.TempDir usage in Packetbeat tests (it broke windows)

- Using single quotes around Windows paths (#4235)
  The thrift test config used double quotes around Windows path separators and this was interpreted incorrectly in YAML parsing.

- Rename TestBadCondition to TestConditions  (#4235)
  This test doesn’t actually test any bad conditions. Plus there is another test in the same directory with the name TestBadCondition.

- Use logp.Beta or logp.Experimental in metricsets  (#4235)
  And in system tests, centralize the logic for asserting that there are no ERR or WARN in logs.

  Filter out errors about “The service process could not connect to the service controller” that occur when testing on Jenkins where Jenkins itself is running as a service. This confuses the Beat because it thinks that it is running as service, but it’s not.

- Remove OS specific error message check from mockbeat  (#4267)
  The error message “no such file or directory” is an OS specific error message. There is a different error message on Windows. Simply checking for “error loading config file” should be sufficient.

- Use shorter filename in Filebeat test for Windows (#4271)
  The test was failing on Windows when `os.rename` failed with `[Error 3] The system cannot find the path specified`. The root cause of the failure was that the path was ~260 characters on Jenkins which is greater than the `MAX_PATH` value in Windows. This PR shortens the test log’s name to resolve the issue.

  The other changes to normalize the filepath are nice to have for Windows, but not strictly required.

- Add filesystem name to test error message (#4272)
  Errors that are logged by the system/filesystem test case don’t have enough context to debug them. This adds the filesystem that caused the error to the message.

### DIFF
--- a/filebeat/tests/system/test_prospector.py
+++ b/filebeat/tests/system/test_prospector.py
@@ -490,16 +490,17 @@ class Test(BaseTest):
         Test that close_inactive still applies also if file was rotated,
         new file created, and rotated file removed.
         """
+        log_path = os.path.abspath(os.path.join(self.working_dir, "log"))
+        os.mkdir(log_path)
+        testfile = os.path.join(log_path, "a.log")
+        renamed_file = os.path.join(log_path, "b.log")
+
         self.render_config_template(
-            path=os.path.abspath(self.working_dir) + "/log/test.log",
+            path=testfile,
             ignore_older="1h",
             close_inactive="3s",
             scan_frequency="0.1s",
         )
-
-        os.mkdir(self.working_dir + "/log/")
-        testfile = self.working_dir + "/log/test.log"
-        renamed_file = self.working_dir + "/log/test_renamed.log"
 
         filebeat = self.start_beat()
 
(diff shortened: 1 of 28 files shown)


andrewkroh:
jenkins, test it

ruflin:
jenkins, retest it

PR #4290, opened 2017-05-11
Ensure correct file permissions on system test conf files

exekias:
Most linux systems use 002 umask by default, this change ensures
we write conf files with the right permissions


andrewkroh:
I resolved the conflict using the Github web UI so that we can see if CI is passing.

exekias:
This never worked, closing it as it's not a big deal

PR #4294, opened 2017-05-12
Rename input_type to type in config and input_type to prospector.type in event

ruflin:
Document type was removed recently because _type does not exist anymore in Elasticsearch. As input_type is actually the type of the prospector it makes more sense to have it as type config options instead of type. Renaming it in the config means it should also be renamed in the event. In the event the field was renamed to `prospector.type`.

This change is on the config side backward compatible as input_type was only deprecated. On the event side the old field does not exist anymore.

* Cleanup expected test json files. Indentation was standardised and `input_type` replaced by `prospector.type`.
* Update changelog
* Add system tests for deprecated message


ruflin:
jenkins, retest it

filebeat/prospector/log/config.go
```diff
@@ -84,8 +85,13 @@ type config struct {
 
 func (c *config) Validate() error {
 
+	// DEPRECATED 6.0.0: warning is already outputted on propsector level
+	if c.InputType != "" {
+		c.Type = c.InputType
```
tsg: 👍 for backwards compatibility.

filebeat/module/system/syslog/test/darwin-syslog-sample.log-expected.json
```diff
@@ -1,90 +1,93 @@
...
-      "system" : {
-        "syslog" : {
-          "hostname" : "a-mac-with-esc-key",
-          "pid" : "21412",
-          "program" : "GoogleSoftwareUpdateAgent",
-          "message" : "2016-12-13 11:35:28.420 GoogleSoftwareUpdateAgent[21412/0x700007399000] [lvl=2] -[KSAgentApp updateProductWithProductID:usingEngine:] Checking for updates for \"All Products\" using engine <KSUpdateEngine:0x100341a00\n\t\tticketStore=<KSPersistentTicketStore:0x100204520 store=<KSKeyedPersistentStore:0x100213290\n\t\t\tpath=\"/Users/tsg/Library/Google/GoogleSoftwareUpdate/TicketStore/Keystone.ticketstore\"\n\t\t\tlockFile=<KSLockFile:0x1002160e0\n\t\t\t\tpath=\"/Users/tsg/Library/Google/GoogleSoftwareUpdate/TicketStore/Keystone.ticketstore.lock\"\n\t\t\t\tlocked=NO\n\t\t\t>\n\t\t>>\n\t\tprocessor=<KSActionProcessor:0x1003bb5f0\n\t\t\tdelegate=<KSUpdateEngine:0x100341a00>\n\t\t\tisProcessing=NO actionsCompleted=0 progress=0.00\n\t\t\terrors=0 currentActionErrors=0\n\t\t\tevents=0 currentActionEvents=0\n\t\t\tactionQueue=( )\n\t\t>\n\t\tdelegate=(null)\n\t\tserverInfoStore=(null)\n\t\terrors=0\n\t>",
-          "timestamp" : "Dec 13 11:35:28"
+    {
```
tsg: Did you have to update these files in order for the tests to pass? From what I remember we're only comparing the fields under the module. Not that it's bad to have them updated :)

Also, how come the diffs are so big, whitespace?

tsg:
@ruflin Appveyor might have caught something.

filebeat/module/system/syslog/test/darwin-syslog-sample.log-expected.json
ruflin: I actually didn't have to update them. It just passed without changes with the old `type` inside. But it bugged me to have "outdated" docs in the tests so I updated them manually. Then my editor automatically did the formatting. Changes are only in whitespaces and `input_type` to `prospector.type`. I can also revert it to master if you prefer.

ruflin:
@tsg Appveyor hit the race condition I just fixed in https://github.com/elastic/beats/pull/4314

PR #4298, opened 2017-05-12
Fix for flaky docker tests

ruflin:
Docker tests were flaky because sometimes a container of an other test was monitored which was stopped at this time. Te log entries for containers which were removed during investigation are removed.

In addition a processor was introduced to only look at the events for docker containers which are from the same environment.

### DIFF
--- a/metricbeat/tests/system/test_docker.py
+++ b/metricbeat/tests/system/test_docker.py
@@ -1,6 +1,7 @@
 import metricbeat
 
 import unittest
+import os
 from nose.plugins.attrib import attr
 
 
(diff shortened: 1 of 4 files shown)


metricbeat/tests/system/test_docker.py
```diff
@@ -11,17 +12,33 @@ def test_container_fields(self):
...
+        )
 
         proc = self.start_beat()
         self.wait_until(lambda: self.output_lines() > 0, max_timeout=20)
         proc.check_kill_and_wait()
-        self.assert_no_logged_warnings()
+        self.assert_no_logged_warnings(["WARN Container stopped when recovering stats",
+                                        "ERR An error occurred while getting docker stats"])
+
+        print(os.environ.get('TESTING_ENVIRONMENT'))
```
andrewkroh: Does this need to stay or was it just for debugging?

ruflin: I wanted to keep it in for the moment to see what went wrong in case the tests fail again.

PR #4303, opened 2017-05-12
Use .go-version to specify the Go version for all CI builds

andrewkroh:
Having a simple file that requires no parsing to retrieve the Go version
provides us a standard portable way to know what Go version to use for builds.
It's basically the least common denominator for builds accross CI systems
(Jenkins, AppVeyor, Travis) and operating systems.

Also changed AppVeyor to invalidate the cached Go version only when the
.go-version file changes instead of when the .appveyor.yml changes.


ruflin:
WFG

PR #4304, opened 2017-05-12
Cherry-pick #3942 to 5.x: Fix Winlogbeat test by checking full hostname

andrewkroh:
Cherry-pick of PR #3942 to 5.x branch. Original message: 

The `computer_name` field in events is the full hostname, but the `win32api.GetComputerName` was returning the shortened netbios name. So the test fail on machines with longer hostnames.


andrewkroh:
jenkins, retest it please

PR #4308, opened 2017-05-13
Fix process summary test case

andrewkroh:
Change test_process_summary to use assert_no_logged_warnings() instead
of a looking for "ERR|WARN". assert_no_logged_warnings() filters out
some expected messages.


tsg:
Thanks @andrewkroh.

PR #4313, opened 2017-05-15
Fixing configuration documentation for kubernetes processor

vjsamuel:
Fixing configuration documentation for kubernetes processor

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -29,6 +29,7 @@ https://github.com/elastic/beats/compare/v6.0.0-alpha1...master[Check the HEAD d
 *Affecting all Beats*
 
 - Fix importing the dashboards when the limit for max open files is too low. {issue}4244[4244]
+- Fix configuration documentation for kubernetes processor {pull}4313[4313]
 
 *Filebeat*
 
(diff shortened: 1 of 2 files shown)


vjsamuel:
@exekias, im still on the fence with this change. this would make sure that we always start at the most recent resource version. but this would also mean that we always miss events that were written by the API server when the watch was broken at our end.

vjsamuel:
@exekias Updated the PR to just fix an issue in the configuration documentation. I dont think its a good idea to do the check based on last resource version. would result in losing metadata update events.

exekias:
@vjsamuel podwatcher stores last seen resource version, you could use that :) https://github.com/elastic/beats/blob/master/libbeat/processors/kubernetes/podwatcher.go#L22

CHANGELOG.asciidoc
```diff
@@ -215,6 +215,7 @@ https://github.com/elastic/beats/compare/v5.3.2...v5.4.0[View commits]
 - Fix potential Elasticsearch output URL parsing error if protocol scheme is missing. {pull}3671[3671]
 - Downgrade Elasticsearch per batch item failure log to debug level. {issue}3953[3953]
 - Make `@timestamp` accessible from format strings. {pull}3721[3721]
+- Fixing configuration documentation for kubernetes processor {pull}4313[4313]
```
exekias: Sorry, I just noticed this line went to an already released version, could you move it to the HEAD? also it would be great if you rename `Fixing` -> `Fix`

PR #4317, opened 2017-05-15
Add option to adjust _source options + improvements

tsg:
This PR adds/fixes:

* Ability to disable _source, or set other _source related options
* Fixes the overwrite logic (was using the wrong template name on the check)
* Fixes error handling
* Integration tests for overwritting

The source can be disabled like this:

```
setup.template:
  settings._source.enabled: false
```

And _source can be also partially enabled via "includes" and "excludes", for example:

```
setup.template:
  settings._source.excludes: ["metricset", "system.process.cmdline"]
```

Adding the discuss label for naming. I hesitated between using `_source` or `source_settings` for these options, but I like the look of `_source.enabled: false`.

Part of #3654 and #4112.

Remaining TODOs:

* [x] Update docs
* [x] Changelog

### DIFF
--- a/libbeat/docs/template-config.asciidoc
+++ b/libbeat/docs/template-config.asciidoc
@@ -20,11 +20,11 @@ section for details.
 *`overwrite`*:: A boolean that specifies whether to overwrite the existing template. The default
 is false.
 
-*`settings`*:: A dictionary of settings to place into the `settings` dictionary of the Elasticsearch
-template. For more details about the available Elasticsearch mapping options, please see the
-Elasticsearch {elasticsearch}/mapping.html[mapping reference].
+*`settings.index`*:: A dictionary of settings to place into the `settings.index` dictionary of the
+Elasticsearch template. For more details about the available Elasticsearch mapping options, please
+see the Elasticsearch {elasticsearch}/mapping.html[mapping reference].
 
-For example:
+Example:
 
 ["source","yaml",subs="attributes,callouts"]
 ----------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


libbeat/template/config.go
```diff
@@ -1,12 +1,13 @@
...
-	Overwrite    bool                   `config:"overwrite"`
-	OutputToFile string                 `config:"output_to_file"`
-	Settings     map[string]interface{} `config:"settings"`
+	Enabled        bool                   `config:"enabled"`
+	Name           string                 `config:"name"`
+	Fields         string                 `config:"fields"`
+	Overwrite      bool                   `config:"overwrite"`
+	OutputToFile   string                 `config:"output_to_file"`
+	Settings       map[string]interface{} `config:"settings"`
+	SourceSettings map[string]interface{} `config:"_source"`
```
ruflin: At first I was also sceptical about using _source but I think for ES users it must be quite intuitive to use it like this.

libbeat/template/template.go
```diff
@@ -17,14 +17,15 @@ var (
...
 
 type Template struct {
-	index       string
-	beatVersion Version
-	esVersion   Version
-	settings    common.MapStr
+	index          string
+	beatVersion    Version
+	esVersion      Version
+	settings       common.MapStr
```
ruflin: Seeing both names I was thinking if it could be `settings._source` and `settings.x` but I would not know how to call `x` :-(

Is there are stuff we potentially going to add here?

tsg: I can imagine there could be more yes (e.g. _all), although I don't plan more at the moment. We could maybe use `index` for `x`, so you'd have: `settings.index.number_of_shards` and `settings._source.enabled: false`. Trying to figure out if there can be any settings that do not fall under `index`...

tsg:
I've updated the code to have `settings.index` for the global index settings and `settings._source` for the `_source` settings. Elasticsearch is a bit lenient when it comes to the index settings in the template. Both of these work:

```
"settings": {
  "number_of_shards": 1
}
```

and 

```
"settings"; {
  "index": {
    "number_of_shards": 1
  }
}
```

With this change, we are can only use the second version. I think that's good anyway, as we don't want two ways of accomplishing the same thing.

libbeat/docs/template-config.asciidoc
```diff
@@ -35,3 +35,17 @@ setup.template.settings:
   index.number_of_shards: 1
```
ruflin: Should this be updated to put `index` on the line above?

tsg: I kind of like it is now because it matches the Elasticsearch docs where they usually talk about `index.*` as the name of the options: https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping.html

PR #4323, opened 2017-05-16
Add the option to generate the template into a file

tsg:
Part of #3654.

This adds two settings: `setup.template.output_to_file.path` and `setup.template.output_to_file.version`. The version refers to the ES version and is optional, we'll use the Beats version if not specified. I put it under `output_to_file` to make it clear that it only applies when outputting to a file:

To generate a config, one can do:

```
./metricbeat -e -E setup.template.output_to_file.path=template.json
```

In the current version, the Beat automatically stops after generating the template, but the output might be slightly confusing:

```
2017/05/16 09:55:02.043671 load.go:116: INFO Template for Elasticsearch 6.0.0-alpha2 written to: template.json
2017/05/16 09:55:02.043717 beat.go:538: CRIT Exiting: Stopping after successfully writing the template to the file.
Exiting: Stopping after successfully writing the template to the file.
```

IMO this is better than the alternative of leaving it running.

To generate the template for the 2.x version, one can do:

```
./metricbeat -e -E setup.template.output_to_file.path=template.json -E setup.template.output_to_file.version=2.4.0
```

Remaining TODOs:

* [x] System test
* [x] Docs
* [x] Changelog

### DIFF
--- a/libbeat/beat/beat.go
+++ b/libbeat/beat/beat.go
@@ -469,8 +469,30 @@ func (b *Beat) registerTemplateLoading() error {
 		}
 	}
 
-	esConfig := b.Config.Output["elasticsearch"]
+	// Check if outputting to file is enabled, and output to file if it is
+	if b.Config.Template != nil && b.Config.Template.Enabled() {
+		var cfg template.TemplateConfig
+		err := b.Config.Template.Unpack(&cfg)
+		if err != nil {
+			return fmt.Errorf("unpacking template config fails: %v", err)
+		}
+		if len(cfg.OutputToFile.Path) > 0 {
+			// output to file is enabled
+			loader, err := template.NewLoader(b.Config.Template, nil, b.Info)
+			if err != nil {
+				return fmt.Errorf("Error creating Elasticsearch template loader: %v", err)
+			}
+			err = loader.Generate()
+			if err != nil {
+				return fmt.Errorf("Error generating template: %v", err)
+			}
 
+			// XXX: Should we kill the Beat here or just continue?
+			return fmt.Errorf("Stopping after successfully writing the template to the file.")
+		}
+	}
+
+	esConfig := b.Config.Output["elasticsearch"]
 	// Loads template by default if esOutput is enabled
 	if (b.Config.Template == nil && esConfig.Enabled()) || (b.Config.Template != nil && b.Config.Template.Enabled()) {
 		if esConfig == nil || !esConfig.Enabled() {
@@ -487,7 +509,7 @@ func (b *Beat) registerTemplateLoading() error {
 
 			loader, err := template.NewLoader(b.Config.Template, esClient, b.Info)
 			if err != nil {
-				return fmt.Errorf("Error creating Elasticsearch template: %v", err)
+				return fmt.Errorf("Error creating Elasticsearch template loader: %v", err)
 			}
 
 			err = loader.Load()
--- a/libbeat/template/config.go
+++ b/libbeat/template/config.go
@@ -5,8 +5,15 @@ type TemplateConfig struct {
 	Name         string           `config:"name"`
 	Fields       string           `config:"fields"`
 	Overwrite    bool             `config:"overwrite"`
-	OutputToFile string           `config:"output_to_file"`
 	Settings     templateSettings `config:"settings"`
+	OutputToFile OutputToFile     `config:"output_to_file"`
+}
+
+// OutputToFile contains the configuration options for generating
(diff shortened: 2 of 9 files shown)


ruflin:
Perhaps we should discuss again with after `--setup` the beat should stop? Because then one could do:

```
./metricbeat --setup -e -E setup.template.output_to_file.path=template.json
```
Or assuming -S is implies `--setup` and prefixes -E with the setup namespace. Then the following could be used:

```
./metricbeat -S template.output_to_file.path=template.json
```

libbeat/beat/beat.go
```diff
@@ -469,8 +469,30 @@ func (b *Beat) registerTemplateLoading() error {
...
+			if err != nil {
+				return fmt.Errorf("Error creating Elasticsearch template loader: %v", err)
+			}
+			err = loader.Generate()
+			if err != nil {
+				return fmt.Errorf("Error generating template: %v", err)
+			}
 
+			// XXX: Should we kill the Beat here or just continue?
+			return fmt.Errorf("Stopping after successfully writing the template to the file.")
```
tsg: In this case, dear reviewdog, we know that the error is final, so the punctuation looks better to the user.

tsg:
@ruflin Yeah, I guess we should review the whole `-setup` mechanism, maybe we can unify the handling in such a way that it makes sense. The `beats-tools` is tempting, but we'll need to build it for all supported platforms, which makes me a bit worried about the packaging overhead (especially since it's not actually a beat).

I have the feeling we can't get a nice solution as part of this PR, so the question is what do we want for the time being? IMO stopping is much nicer in this case, but it's inconsistent with `-setup`, that's true :(. I'm still leaning towards stopping.

ruflin:
Agree that we can't find a good solution as part of this PR. +1 on stopping as if you want to manually load the template, you don't want to already send data.

libbeat/beat/beat.go
```diff
@@ -469,8 +469,30 @@ func (b *Beat) registerTemplateLoading() error {
 		}
 	}
 
-	esConfig := b.Config.Output["elasticsearch"]
+	// Check if outputting to file is enabled, and output to file if it is
```
ruflin: This comment confused me at first because the check for the file only happens later, but I see you meant it for the "complete" block

libbeat/template/config.go
```diff
@@ -5,8 +5,15 @@ type TemplateConfig struct {
 	Name         string           `config:"name"`
 	Fields       string           `config:"fields"`
 	Overwrite    bool             `config:"overwrite"`
-	OutputToFile string           `config:"output_to_file"`
 	Settings     templateSettings `config:"settings"`
+	OutputToFile OutputToFile     `config:"output_to_file"`
```
ruflin: Not too much of a fan of the config name, but well :-)

libbeat/template/load.go
```diff
@@ -80,6 +82,43 @@ func (l *Loader) Load() error {
 	return nil
 }
 
+// Generate generates the template and writes it to a file based on the configuration
+// from `output_to_file`.
+func (l *Loader) Generate() error {
```
ruflin: We could probably now use this function also in https://github.com/elastic/beats/blob/master/dev-tools/cmd/index_template/index_template.go

libbeat/tests/system/base.py
```diff
@@ -9,4 +9,11 @@ def setUpClass(self):
         self.beat_name = "mockbeat"
         self.beat_path = os.path.abspath(os.path.join(os.path.dirname(__file__), "../../"))
         self.test_binary = self.beat_path + "/libbeat.test"
+        self.beats = [
+            "filebeat",
```
ruflin: Looks like a beats specifics dependency in libbeat.

libbeat/tests/system/test_dashboard.py
```diff
@@ -20,9 +20,7 @@ def test_load_dashboard(self):
         """
         Test loading dashboards for all beats
         """
-        beats = ["metricbeat", "packetbeat", "filebeat", "winlogbeat"]
-
-        for beat in beats:
+        for beat in self.beats:
```
ruflin: be aware that these tests are currently skipped.

libbeat/tests/system/base.py
tsg: Yeah, only needed for those skipped tets for now, I can delete them.

libbeat/template/load.go
tsg: Oh, I didn't realize we have that utility already.

PR #4329, opened 2017-05-16
Change Metricbeat default config to reduce disk space

tsg:
This implements several changes from #4112:

* reduce the frequency for the fsstats and filesystem
* filter out some filesystems by mount point
* set one shard by default
* enable best_compression by default

It also removes some of the redundant comments.

Question:
* This currently modifies the template only for metricbeat. Do we want some of these changes for other beats as well?


metricbeat/docs/modules/system.asciidoc
```diff
@@ -86,45 +86,30 @@ in <<configuration-metricbeat>>. Here is an example configuration:
...
   processes: ['.*']
   process.include_top_n:
     by_cpu: 5      # include top 5 processes by CPU
     by_memory: 5   # include top 5 processes by memory
+- module: system
+  enabled: true
+  period: 1m
+  metricsets:
+    - filesystem
+    - fsstat
```
ruflin: We should really rename this to `filbeatsystem_summary` ;-)

metricbeat/_meta/setup.yml
```diff
@@ -0,0 +1,5 @@
+#==================== Elasticsearch template setting ==========================
+setup.template.settings:
+  index.number_of_shards: 1
```
s1monw: this should be the damned default in ES to begin with...

s1monw:
I am just giving a LGTM for the sake of the number of shards setting!

metricbeat/docs/modules/system.asciidoc
monicasarbu: I guess you mean `filesystem_summary`. I agree with you.

ruflin: lol, yes :-)

PR #4330, opened 2017-05-16
Update processor docs

andrewkroh:
- Fix misspelling in add_local config option for abbreviation
- Update config.full.yml with the new cloud provider types.
- Add a tilde to the add_cloud_metadata example config to avoid having no value at all.
- Add more info to the add_locale docs.

### DIFF
--- a/libbeat/_meta/config.full.yml
+++ b/libbeat/_meta/config.full.yml
@@ -64,15 +64,18 @@
 #           http.code: 200
 #
 # The following example enriches each event with metadata from the cloud
-# provider about the host machine. It works on EC2, GCE, and DigitalOcean.
+# provider about the host machine. It works on EC2, GCE, DigitalOcean,
+# Tencent Cloud, and Alibaba Cloud.
 #
 #processors:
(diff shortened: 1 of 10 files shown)


libbeat/_meta/config.full.yml
```diff
@@ -64,15 +64,18 @@
 #           http.code: 200
 #
 # The following example enriches each event with metadata from the cloud
-# provider about the host machine. It works on EC2, GCE, and DigitalOcean.
+# provider about the host machine. It works on EC2, GCE, DigitalOcean,
+# Tencent Cloud, and Alibaba Cloud.
 #
 #processors:
-#- add_cloud_metadata:
+#- add_cloud_metadata: ~
```
exekias: What's this `~` for?

andrewkroh: It's a [null](http://yaml.org/spec/history/2002-04-07.html#trans-null) in YAML. It's not required but we wanted to have some kind of value on the RHS.

PR #4331, opened 2017-05-17
Fix panic due to no type

tsg:
Some Beats no longer add a "type", but the LS output was still requiring it.
This hardcodes the type in `@metadata` to `doc`, to keep BWC with older Logstashes/configs.

It also removes `"type"` from a few kafka integration tests, to make sure we don't depend
on it.

### DIFF
--- a/libbeat/outputs/logstash/json.go
+++ b/libbeat/outputs/logstash/json.go
@@ -34,14 +34,11 @@ func makeLogstashEventEncoder(beat string) (func(interface{}) ([]byte, error), e
 
 		buf.WriteRune('{')
 		if _, hasMeta := event["@metadata"]; !hasMeta {
-			typ := event["type"].(string)
-			buf.WriteString(`"@metadata":{"type":`)
-			encodeString(buf, typ)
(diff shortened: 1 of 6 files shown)


tsg:
jenkins, test it again

libbeat/outputs/logstash/json.go
```diff
@@ -34,14 +34,11 @@ func makeLogstashEventEncoder(beat string) (func(interface{}) ([]byte, error), e
 
 		buf.WriteRune('{')
 		if _, hasMeta := event["@metadata"]; !hasMeta {
-			typ := event["type"].(string)
-			buf.WriteString(`"@metadata":{"type":`)
-			encodeString(buf, typ)
+			buf.WriteString(`"@metadata":{"type":"doc","beat":`)
```
ruflin: What if there is a `type` like we have in packetbeat? I think it is ok to just have it outside the meta info.

tsg: Yeah, i think it's safer to hardcode the metadata one to `doc`, as otherwise existing Logstash configs will put it in `_type`.

urso:
Can we have a BC-breaking changelog entry?

Issue #4332, opened 2017-05-17
Race condition harvester shutdown on reloading


ruflin:
I hope this was fixed by https://github.com/elastic/beats/pull/4314 There is still a chance the panic can happen as the waitgroup on the harvester side is only incremented as part of the go routine and if outside the go routine the shutdown starts before, it could be closed before.

One solution would be to pass increment the waitgroup in the harvester from the outside the go routine. But this would mean adding more Interface methods to the harvester interface which I prefer not to do.

Panic from appveyor
```
2017/05/16 13:36:09.698462 crawler.go:132: INFO Crawler stopped
2017/05/16 13:36:09.698462 spooler.go:104: INFO Stopping spooler
==================
WARNING: DATA RACE
Write at 0x00c042012fc0 by goroutine 13:
  runtime.closechan()
      C:/Users/appveyor/.gvm/versions/go1.8.1.windows.amd64/src/runtime/chan.go:320 +0x0
  github.com/elastic/beats/filebeat/spooler.(*Spooler).Stop()
      github.com/elastic/beats/filebeat/spooler/_obj/spooler.go:106 +0xa3
  github.com/elastic/beats/filebeat/beater.(*Filebeat).Run.func2()
      github.com/elastic/beats/filebeat/beater/_obj/filebeat.go:208 +0x73
  github.com/elastic/beats/filebeat/beater.(*Filebeat).Run()
      github.com/elastic/beats/filebeat/beater/_obj/filebeat.go:259 +0xbff
  github.com/elastic/beats/libbeat/beat.(*Beat).launch()
      c:/gopath/src/github.com/elastic/beats/libbeat/beat/beat.go:270 +0xb17
  github.com/elastic/beats/libbeat/beat.Run.func1()
      c:/gopath/src/github.com/elastic/beats/libbeat/beat/beat.go:159 +0xc1
  github.com/elastic/beats/libbeat/beat.Run()
      c:/gopath/src/github.com/elastic/beats/libbeat/beat/beat.go:160 +0x6b
  github.com/elastic/beats/filebeat.main()
      github.com/elastic/beats/filebeat/_test/_obj_test/main.go:16 +0x97
  github.com/elastic/beats/filebeat.TestSystem()
      c:/gopath/src/github.com/elastic/beats/filebeat/main_test.go:20 +0x73
  testing.tRunner()
      C:/Users/appveyor/.gvm/versions/go1.8.1.windows.amd64/src/testing/testing.go:657 +0x10e

[...]

PR #4336, opened 2017-05-17
Remove waitgroup handling from Stopping of harvesters

ruflin:
During trying to reproduce https://github.com/elastic/beats/issues/4332 it seems the waitgroup in the Stop part is not needed. Even more it could cause issues.

The initial approach to fix the race was to pass the waitgroup to the harvester. But this would change the harvester interface which I would like to keep as simple as possible.

Before one of the problems was, that the harvester was added to the registry early which means Stop was potentially called on a harvester that was not started yet. This should now be prevented by checking on Start if the harvester was already stopped before any callback methods are registered.

### DIFF
--- a/filebeat/harvester/registry.go
+++ b/filebeat/harvester/registry.go
@@ -29,6 +29,12 @@ func (r *Registry) remove(h Harvester) {
 	delete(r.harvesters, h.ID())
 }
 
+func (r *Registry) add(h Harvester) {
+	r.Lock()
+	defer r.Unlock()
+	r.harvesters[h.ID()] = h
+}
+
 // Stop stops all harvesters in the registry
 func (r *Registry) Stop() {
 	r.Lock()
@@ -40,9 +46,7 @@ func (r *Registry) Stop() {
 	close(r.done)
 
 	for _, hv := range r.harvesters {
-		r.wg.Add(1)
 		go func(h Harvester) {
-			r.wg.Done()
 			h.Stop()
 		}(hv)
 	}
@@ -62,20 +66,17 @@ func (r *Registry) Start(h Harvester) {
 	defer r.Unlock()
 
 	// Make sure no new harvesters are started after stop was called
-	select {
-	case <-r.done:
+	if !r.active() {
 		return
-	default:
 	}
 
 	r.wg.Add(1)
-	r.harvesters[h.ID()] = h
-
 	go func() {
 		defer func() {
 			r.remove(h)
 			r.wg.Done()
 		}()
+		r.add(h)
 		// Starts harvester and picks the right type. In case type is not set, set it to default (log)
 		err := h.Run()
 		if err != nil {
@@ -89,4 +90,13 @@ func (r *Registry) Len() uint64 {
 	r.RLock()
 	defer r.RUnlock()
 	return uint64(len(r.harvesters))
+}
+
+func (r *Registry) active() bool {
+	select {
+	case <-r.done:
+		return false
+	default:
+		return true
+	}
 }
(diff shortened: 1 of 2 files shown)


filebeat/harvester/registry.go
```diff
@@ -68,13 +72,12 @@ func (r *Registry) Start(h Harvester) {
...
 
 	r.wg.Add(1)
-	r.harvesters[h.ID()] = h
-
 	go func() {
 		defer func() {
 			r.remove(h)
 			r.wg.Done()
 		}()
+		r.add(h)
```
urso: this has a very high chance to insert the harvester into the registry, after Start did return. Can we run into a race with the prospector requiring an up-to-date registry?

filebeat/harvester/registry.go
```diff
@@ -39,9 +45,7 @@ func (r *Registry) Stop() {
 	close(r.done)
 
 	for _, hv := range r.harvesters {
-		r.wg.Add(1)
 		go func(h Harvester) {
-			r.wg.Done()
 			h.Stop()
 		}(hv)
 	}
```
urso: where is `r.wg.Wait()` executed? Having `r.wg.Wait()` after this for loop, we can wait for all harvester to finish. This will allow us to make h.Stop() non-blocking ->

e.g.
```
func (r *Registry) Stop() {
  r.Lock()                // <- the lock on Start and Stop guarantees stop can only be called after Start finishes
  defer r.Unlock() 

  close(r.done)

  for _, h := range r.harvesters {
    h.Stop()
  }

  r.wg.Wait()   // we wait here before releasing the lock to ensure start can not be executed before Stop finishes 
}
```


with 

```
 func (r *Registry) Start(h Harvester) {
  r.Lock()
  defer r.Unlock()

  // Make sure no new harvesters are started after stop was called
  if !r.active() {
    return
  }

  r.wg.Add(1)
  r.harvesters[h.ID()] = h
  go func() {
    // <- note: due to stop holding the mutex and waiting on `r.wg` the harvester is removed from 
    // the registry after r.wg.Done is called. This potentially races with the prospector not being able to re-start
    // the harvester. With the prospector trying to periodically start the harvester, the fact the race exists can be "ignored".
    defer r.remove(h) 

    defer r.wg.Done()
    h.Start()
  }
}

func (r *Registry) active() bool {
  select {
    case <-r.done:
      return false
    default:
      return true
  }  
}

func (h *Harvester) Stop() {
  // return immediately -> shutdown will be complete asynchronously.
  // Due to `r.wg` being decremented, only after the harvester `Start/Run` returned, the `registry.Stop` will block anyways.
  h.stopOnce.Do(func() {
 		close(h.done)
   })
}
```

ruflin: It is waiting for Completion in the defer statement on line 36. I quite like that Stop is actually blocking because someone from outside the harvester should not require knowlege about its inner working. I need to check what we do in other cases with `Stop()`

filebeat/harvester/registry.go
ruflin: The only time the prospector interacts with the registry is when adding harvesters or on stopping it. The reason I moved it inside is that previously a harvester was added to the registry and it could happen it was started later and stopped before. With this change it is less likely but probably still possible. That is why I also added the to the harvester so in case it was already stopped, it will not start and directly return. That means in this case the harvester would also not have to be stopped, as it is already finished.

filebeat/harvester/registry.go
urso: The difference is when it's blocking and when the mutex is released. My proposal releases the mutex only after all harvesters are finished -> start can only be called after Stop did return.

ruflin: That's a good point. The stopping has now issue (at least in my implementation) is that as long as there is a lock, harvesters can't be removed from the list as they also require a lock.

ruflin: Thinking again about the above: In my implementation, because of the lock Start can only be called again after it is guaranteed that the channel is closed, means no new harvesters will be started.

andrewkroh:
I'm going to merge this so that we can get some testing time on this code.

PR #4339, opened 2017-05-17
Enhance config.Namespace

urso:
when unpacking into config.Namespace, namespaces can be disabled
via `enabled: false` now. For settings allowing exactly one configuration, this
can be used by users to enabled/disable namespace without having to comment
them out in the configuration file.

example usage:

```
type Config struct {
    Output common.ConfigNamespace
}
```

user can have at most one enabled:

```
output.namespace1:
  enabled: false

output.namespace2:
  enabled: true
```

### DIFF
--- a/libbeat/common/config.go
+++ b/libbeat/common/config.go
@@ -35,7 +35,8 @@ type Config ucfg.Config
 
 // ConfigNamespace storing at most one configuration section by name and sub-section.
 type ConfigNamespace struct {
-	C map[string]*Config `config:",inline"`
+	name   string
+	config *Config
 }
 
 type flagOverwrite struct {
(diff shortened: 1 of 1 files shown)


libbeat/common/config.go
```diff
@@ -327,33 +328,59 @@ func (f *flagOverwrite) Get() interface{} {
...
+
+/*
 // Validate checks at most one sub-namespace being set.
 func (ns *ConfigNamespace) Validate() error {
 	if len(ns.C) > 1 {
 		return errors.New("more then one namespace configured")
 	}
 	return nil
 }
+*/
```
ruflin: Should this code be removed?

ruflin:
I'm not 100% sure I fully understand what this feature is doing compare to before. What wasn't possible and what is possible now (or the other way around)?

libbeat/common/config.go
urso: ups, yes.

urso:
this was not possible before:

```
output.namespace1:
  enabled: false

output.namespace2:
  enabled: true
```

when unpacking this config, ConfigNamespace will raise an error, because it found `namespace1` and `namespace2` stored. This PR enables this use-case, as the new `Unpack` on `ConfigNamespace` ignores namespaces with `enabled: false`.

libbeat/common/config.go
```diff
@@ -327,33 +328,49 @@ func (f *flagOverwrite) Get() interface{} {
 	return f.value
 }
 
-// Validate checks at most one sub-namespace being set.
-func (ns *ConfigNamespace) Validate() error {
-	if len(ns.C) > 1 {
-		return errors.New("more then one namespace configured")
+func (ns *ConfigNamespace) Unpack(cfg *Config) error {
```
ruflin: Could you add here some docs on what is happening if more then 1 namespace is enabled?

urso: done

ruflin:
@urso Build seems to be broken :-(

Issue #4333, opened 2017-05-17
Start Filebeat Service failed on argument test when it executed by non-root user


dreampuf:
From the code, the argument test will be failed on `test` function. Cause the [checker](https://github.com/elastic/beats/blob/c10d47af73f51b3db2a46ffb606e771bb8d397b3/libbeat/common/config.go#L432). 
https://github.com/elastic/beats/blob/master/dev-tools/packer/platforms/centos/init.j2#L45

```
+ /usr/share/filebeat/bin/filebeat -c /etc/filebeat/filebeat.yml -path.home /usr/share/filebeat -path.config /etc/filebeat -path.data /var/lib/filebeat -path.logs /var/log/filebeat -e -configtest
filebeat2017/05/17 11:02:41.103321 beat.go:339: CRIT Exiting: error loading config file: config file ("/etc/filebeat/filebeat.yml") must be owned by the beat user (uid=0) or root
Exiting: error loading config file: config file ("/etc/filebeat/filebeat.yml") must be owned by the beat user (uid=0) or root
```

For confirmed bugs, please report:
- Version: 5.3.2
- Operating System: CentOS5
- Steps to Reproduce: 
  - Install the latest filebeat `yum install filebeat`
  - config the file permission of binary and configs `chown monitor:monitor /etc/filebeat/filebeat.yml /usr/share/filebeat/bin/filebeat`
  - Lauch the filebeat service. `/etc/init.d/filebeat start`

andrewkroh:
This looks like the expected behavior to me. See the documentation [here](https://www.elastic.co/guide/en/beats/libbeat/5.4/config-file-permissions.html). Why are you changing the owner of the binary and config file?

dreampuf:
@andrewkroh I want to execute the filebeat with a non-root user.

andrewkroh:
Changing the owner of the file won't cause it to be executed as a different user. You also need to modify the init.d script to change the user that the process executes as. The script will probably require multiple changes to execute the beat as a user other than root.

andrewkroh:
I think you would need to add `-u user -g group` to the `wrapperopts` at https://github.com/elastic/beats/blob/master/dev-tools/packer/platforms/centos/init.j2#L31.

The arguments are described here for daemon process: https://github.com/tsg/go-daemon/blob/master/god.c#L35-L36

In addition, the `test()` func will need modified to execute as the correct user. Or you could just have the test not check the file permissions by using `-strict.perms=false`.

[...]

PR #4340, opened 2017-05-17
Add non-root support in sysvinit script

dreampuf:
As #4333 mentioned, I found that the sysvinit script of beats will fail on non-root user launch the beats. Cause [the arguments checker](https://github.com/elastic/beats/blob/c10d47af73f51b3db2a46ffb606e771bb8d397b3/libbeat/common/config.go#L432) will be executed as root by default. 
I add an optional `BEAT_USER`, which may come from `/etc/default/$NAME`.


andrewkroh:
jenkins, package it

andrewkroh:
What provides `runuser`? And it is safe to assume that it is present?

dreampuf:
@andrewkroh `runuser`([man page](http://man7.org/linux/man-pages/man1/runuser.1.html)) is a tool released by `util-linux-2.23`[link](https://www.kernel.org/pub/linux/utils/util-linux/v2.23/v2.23-ReleaseNotes) . It has been provided in most popular distributions. CentOS 6+, Debian 8+, Ubuntu 16.04+.
`su` is another option, there are some comments of these two. Details: http://danwalsh.livejournal.com/55588.html

Basically, if Filebeat plan to support some older distribution. CentOS5/Debian7/Ubuntu14.04. `su USER -c` will be the option. Otherwise `runuser` is a better choice.

andrewkroh:
Thanks for the details on `runuser` vs `su`. We do need to support ubuntu 14.04 and debian 7. Could it test if `runuser` is available and fallback to `su` if not?

andrewkroh:
Can you please sign the CLA (contributor license agreement)? http://www.elasticsearch.org/contributor-agreement/

dreampuf:
@andrewkroh `su` added. Also signed CLA.

dev-tools/packer/platforms/centos/init.j2
```diff
@@ -42,8 +44,12 @@ if status | grep -q -- '-p' 2>/dev/null; then
     pidopts="-p $pidfile"
 fi
 
+if [ -f /sbin/runuser ]; then
```
andrewkroh: Rather than explicitly stating the path, I think it would be better to use the check described [here](http://stackoverflow.com/a/4785518/503798). WDYT?

dev-tools/packer/platforms/centos/init.j2
```diff
@@ -42,8 +44,12 @@ if status | grep -q -- '-p' 2>/dev/null; then
...
+    user_wrapper="runuser"
```
andrewkroh: Do you also need to update to `user_wrapperopts="-u $beat_user"`?

dreampuf: It's same argument and order. So I can reuse `user_wrapperopts` here.

dev-tools/packer/platforms/centos/init.j2
dreampuf: Good point.

dreampuf:
@andrewkroh It seems like only a CLA need to be done from this checklist. But I had signed a CLA yesterday(with my github username and email address). Anything I can do?

tsg:
jenkins, test it

tsg:
We should have a CHANGELOG entry on this one. @dreampuf could you add a line under the `Added` section, please? I can also easily add it in a follow up PR if you prefer to merge it quickly.

tsg:
Jenkins test failure is not related.

PR #4341, opened 2017-05-17
Add debug to Jolokia JMX metricset

andrewkroh:
Add debug logging of the JSON request and responses.

Log the JSON body in errors if unmarshalling fails.


andrewkroh:
jenkins, test it

PR #4343, opened 2017-05-18
Add TTL option to the logstash output

andrew-d:
This is a (rebased) version of #3454 that addresses the changes requested in [this comment][com].  We remove TTL support from the async client, and instead just print a warning if the value is set.  I've also updated the documentation and sample config file to mention this.

cc @urso and @nicorevin 

[com]: https://github.com/elastic/beats/pull/3454#issuecomment-280408844

### DIFF
--- a/filebeat/filebeat.full.yml
+++ b/filebeat/filebeat.full.yml
@@ -628,6 +628,13 @@ output.elasticsearch:
   # Optional load balance the events between the Logstash hosts
   #loadbalance: true
 
+  # Optional maximum time to live for a connection to Logstash, after which the
+  # connection will be re-established.  A value of 0 (the default) will disable
+  # this feature.
+  #
+  # Not yet supported for async connections (i.e. with the "pipelining" option set)
+  #ttl: 30s
+
   # Number of batches to be send asynchronously to logstash while processing
   # new batches.
   #pipelining: 0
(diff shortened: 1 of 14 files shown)


andrew-d:
(I just signed the CLA)

andrew-d:
@ruflin Anything I can do to help this get moved forward?  It'd be nice if we can get this into the Beats 6.0 release, if possible 😄

andrew-d:
Looks like we have one test failure due to a filename being too long on Windows, and an error in `TestFileSystemList` that doesn't look like it's caused by me?

ruflin:
@andrew-d I think we definitively need to wait here for @urso and @jordansissel for this as there were lots of discussion if that is actually needed.

andrew-d:
Well, let me provide a bit of background on why we'd want this, then 😄 

I'm currently running 3 logstash instances behind an Amazon ELB (in TCP mode), and running into the problem that we can't easily do a rolling restart.  Ideally, we'd like to take one instance out of service, wait for connections to drain, apply new config, and then bring it back up.  However, we can't do this for two current reasons:

1) Connections are persistent, which means that just taking an instance out of the load balancer doesn't necessarily do anything.
2) When we bring the instance back up, it doesn't start servicing new connections.  While it was down, clients have reconnected to the LB and have started talking to the other instances in service.  As such, we end up in the state where ~every client is talking to the last instance to get restarted.

In addition, point #2 means that we don't really have a way to ensure consistent load; if services come up at a different rate, the first ones to start will have the majority of clients.  We can solve this by dropping connections across the entire ELB, but that's a bit of a large hammer and is more work than just scaling the backing ASG.

Having a `ttl` to occasionally reconnect would solve all the above problems.

PR #4348, opened 2017-05-18
Including protologbeat in list of community beats

hartfordfive:


hartfordfive:
@andrewkroh @tsg @ruflin I've added this latest beat i've created.  As a sample use case, I'll be using it for diskless logging for application containers within a Kubernetes pod.  I appreciate any feedback you guys might have.

libbeat/docs/communitybeats.asciidoc
```diff
@@ -61,6 +61,7 @@ https://github.com/kozlice/phpfpmbeat[phpfpmbeat]:: Reads status from PHP-FPM.
 https://github.com/joshuar/pingbeat[pingbeat]:: Sends ICMP pings to a list
 of targets and stores the round trip time (RTT) in Elasticsearch.
 https://github.com/carlpett/prombeat[prombeat]:: Indexes https://prometheus.io[Prometheus] metrics.
+https://github.com/hartfordfive/[protologbeat][protologbeat]:: Successor to udplogbeat. Accpets structure and unstructured logs via UDP or TCP.  Can also be used to receive syslog messages or GELF formated messages.
```
ruflin: I think the first [ ] brackets are too much?

ruflin:
Very interesting. In the meantime I did quite a bit of refactoring of the filebeat prospector structure which now makes adding new prospector types easier and thing like this possible: https://github.com/elastic/beats/pull/4180

It could be discussed if this should also be a prospector type? Or if it fits in filebeat or not?

hartfordfive:
Yes I'm definitely open for that discussion.  If you see it as beneficial to merge some of this functionality into the beats library, I'm all for it.  Considering syslog messages have a predefined format, maybe that would be a good first candidate to integrate as a new prospector?  I also like the idea of adding GELF as a potential prospector considering it does have a message [standard](http://docs.graylog.org/en/2.2/pages/gelf.html#gelf-format-specification) although it can have some additional field names.  

Considering I've only briefly looked through the code in #4180, would you see the potential of unknown fields as a problem?  Could this be something the user specifies in the YAML config?  Also, keep in mind that the logs for this beat are all accepted via TCP or UDP instead of files like filebeat does.  Could prospectors also be applied to non-file originating events?

libbeat/docs/communitybeats.asciidoc
hartfordfive: Yup, missed that one. I'll fix that.

hartfordfive:
@ruflin On a total separate note (just as a heads up), I plan on also developing a custom beat to process logs from [Sendgrid Webhook events](https://sendgrid.com/docs/API_Reference/Webhooks/event.html).  I may have some questions for you guys regarding that in the IRC chat sometime in the near future.

hartfordfive:
@ruflin By the way, I was thinking about a few things regarding filebeat/libbeat.  These ideas of mine might be totally off or unreasonable so don't hesitate to let me know if out of line!

Considering now that filebeat has an `input_type` of _log_, _stdin_, and now _redis_, would it also be reasonable to potentially implement an `input_type` of _tcp_ and _udp_?  I haven't gone through that part of the code yet to see how feasible it is, but if it is and it's not overly complex to implement, maybe it could make sense to add those?   If you do think that makes sense, the would it also be reasonable to rename the filebeat project to something more suitable like "**_logbeat_**" or something similar?

Again, I know I might be way off on this, but I was just curious to know what you and the other team members feedback was on this?

libbeat/docs/communitybeats.asciidoc
```diff
@@ -61,6 +61,7 @@ https://github.com/kozlice/phpfpmbeat[phpfpmbeat]:: Reads status from PHP-FPM.
 https://github.com/joshuar/pingbeat[pingbeat]:: Sends ICMP pings to a list
 of targets and stores the round trip time (RTT) in Elasticsearch.
 https://github.com/carlpett/prombeat[prombeat]:: Indexes https://prometheus.io[Prometheus] metrics.
+https://github.com/hartfordfive/protologbeat[protologbeat]:: Successor to udplogbeat. Accpets structure and unstructured logs via UDP or TCP.  Can also be used to receive syslog messages or GELF formated messages.
```
ruflin: s/Accpets/Accepts/

I would put the successor part at the end.

hartfordfive: Sorry, seems like spelling skills are lacking today...

hartfordfive:
Yes, we can definitely do a Zoom call sometime soon to go over those discussion points in more detail.

PR #4351, opened 2017-05-18
Change grok pattern to fetch correct IP from X-Forwarded-For list

sepal:
This PR changes the grok pattern for the nginx access logs ingest file in a way, that it retrieves the correct IP if the X-Forwarded-For header was logged into instead of the remote_addr variable.

The X-Forwarded-For header is a non standard header which creates a lists of IPs through which proxies the request has passed as well as the original clients IP and looks basically like this:
```
X-Forwarded-For: client1, proxy1, proxy2
```
which results in a log line like this:
```
```

The new pattern retrieves the first IP, which is the important one, and matches non or all succeeding IPs that are concatinated with a comma and a space.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -172,6 +172,7 @@ https://github.com/elastic/beats/compare/v5.4.0...v6.0.0-alpha1[View commits]
 - Add filebeat.config.path as replacement for config_dir. {pull}4051[4051]
 - Add a `recursive_glob.enabled` setting to expand `**` in patterns. {pull}3980[3980]
 - Add Icinga module. {pull}3904[3904]
+- Add ability to parse nginx logs exposing the X-Forwarded-For header instead of the remote address.
 
 *Heartbeat*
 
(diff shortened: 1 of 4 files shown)


exekias:
jenkins, test it

exekias:
LGTM

ruflin:
LGTM. Could you add a changelog entry? I would put it under "Added" as this is not really a bugfix but a new feature.

sepal:
I added the change to the changelog as requested.

tsg:
@sepal Can you add a test case (`test.log` and `test.log-expected.json`) by the model we have in apache2: [test.log](https://github.com/elastic/beats/blob/master/filebeat/module/apache2/access/test/test.log) and [test.log-expected.json](https://github.com/elastic/beats/blob/master/filebeat/module/apache2/access/test/test.log-expected.json). The tests will automatically pick it up and check that it all works fine. Happy to also do that myself in a follow up PR if you don't have the time.

sepal:
Yes I can, I'll look into that tomorrow or on thursday.

ruflin:
jenkins, test it

PR #4352, opened 2017-05-18
Add `add_docker_metadata` processor

exekias:
This processor will enrich events matching the desired configuration with Docker info:

 - docker.container.id
 - docker.container.image
 - docker.container.name
 - docker.container.labels

This PR adds `docker/client` dependency and cleans up common vendors with current `go-dockerclient` from `metricbeat/module/docker`.

I'm also using a patch for `docker/client` (https://github.com/exekias/moby/commit/83d94aa2b98f10c4534d77077645d5e328533ce4) to ensure the client works in netbsd, will check with upstream if I can push that to master.

Important files to check here:
vendor.json:
- https://github.com/elastic/beats/pull/4352/files#diff-297657274c7cf7eeb37404116eee1e9e
- https://github.com/elastic/beats/pull/4352/files#diff-bd290170e2912d3e8694db1a151066e5

Everything under `add_docker_metadata` (https://github.com/elastic/beats/pull/4352/commits/78f9943aa259fd6432cddf1a4a25768de0cd131b)


andrewkroh:
I think we should consider using the top level vendor dir and removing the `metricbeat/module/docker/vendor` contents. We would have to check that nothing breaks in the docker module, but I'm guessing this would lead to smaller binaries. I don't know if `cgo` is used anywhere in these deps, but I have run into problems duplicate symbols when the same lib is included twice due to it being vendored under multiple paths.

libbeat/processors/add_docker_metadata/add_docker_metadata.go
```diff
@@ -0,0 +1,93 @@
...
+	config := defaultConfig()
+
+	err := cfg.Unpack(&config)
+	if err != nil {
+		return nil, fmt.Errorf("fail to unpack the add_docker_metadata configuration: %s", err)
+	}
+
+	watcher, err := watcherConstructor(config.Host, config.TLS)
+	if err != nil {
+		panic(err)
```
andrewkroh: Why not return an error?

libbeat/processors/add_docker_metadata/add_docker_metadata.go
```diff
@@ -0,0 +1,93 @@
...
+		event["docker"] = meta
+	} else {
+		logp.Debug("docker", "Container not found: %s", cid)
+	}
+
+	return event, nil
+}
+
+func (d *addDockerMetadata) String() string {
+	return "add_docker_metadata"
```
andrewkroh: I think it is customary to include config info in the string representation (assuming no secrets are included). I don't believe there is a standard format.

andrewkroh:
jenkins, test it

andrewkroh:
jenkins, test it

libbeat/processors/add_docker_metadata/watcher.go
```diff
@@ -0,0 +1,171 @@
...
+
+	containers, err := w.client.ContainerList(w.ctx, types.ContainerListOptions{})
+	if err != nil {
+		return err
+	}
+
+	for _, c := range containers {
+		w.containers[c.ID] = &Container{
+			ID:     c.ID,
+			Name:   c.Names[0][1:], // Strip '/' from container names
```
ruflin: Are we doing that also in the container metricset?

libbeat/processors/add_docker_metadata/watcher.go
```diff
@@ -0,0 +1,171 @@
...
+		Filters: filters,
+	}
+
+	for {
+		events, errors := w.client.Events(w.ctx, options)
+
+	WATCH:
+		for {
+			select {
+			case event := <-events:
```
ruflin: So this is actually a docker event listener instead of polling the data. Very similar to what we have now in kubernetes? Should we also have this in the docker module as a metricset?

filebeat/filebeat.full.yml
```diff
@@ -490,6 +490,19 @@ filebeat.prospectors:
 #- add_locale:
 #    format: offset
 #
+# The following example enriches each event with docker metadata, it matches
+# given fields to an existing container id and adds info from that container:
+#
+#processors:
+#- add_docker_metadata:
+#    match_fields: ["system.process.cgroup.id"]
```
ruflin: I was first confused by this and had to read the code to understand what it does. It takes this field and checks if the value of this field matches to any container id. If yes, it will add the container meta data to the event.

In the code, it can currently only match 1 field I think, but here it is plural and you use an array. How is this going to work exactly?

ruflin: BTW I'm not too sure about the example used here. In case someone uses cgroups, doesn't that kind of mean he doesn't want to use docker?

libbeat/processors/add_docker_metadata/add_docker_metadata.go
exekias: wow, I can believe I left that there, sorry

filebeat/filebeat.full.yml
exekias: It checks all the fields until one of them matches

exekias: As for the example, you may want to have docker info on top of the cgroup id, so you can filter by, for instance, docker image

libbeat/processors/add_docker_metadata/add_docker_metadata.go
```diff
@@ -0,0 +1,93 @@
...
+	}, nil
+}
+
+func (d *addDockerMetadata) Run(event common.MapStr) (common.MapStr, error) {
+	var cid string
+	for _, field := range d.fields {
+		value, err := event.GetValue(field)
+		if err == nil {
+			if strValue, ok := value.(string); ok {
+				cid = strValue
```
exekias: there is no way to match several fields into several containers, that should not happen. Fields is a list cause you may have different fields you want to match, not all of them present in all events

libbeat/processors/add_docker_metadata/add_docker_metadata.go
```diff
@@ -0,0 +1,93 @@
...
+
+		if len(container.Labels) > 0 {
+			labels := common.MapStr{}
+			for k, v := range container.Labels {
+				labels.Put(k, v)
+			}
+			meta.Put("container.labels", labels)
+		}
+
+		meta.Put("container.id", container.ID)
```
exekias: yes, they do

libbeat/processors/add_docker_metadata/watcher.go
exekias: This is an spurious char given by this API call & version, doesn't seem to be the case for our other calls

exekias:
Ok I'll play with vendors and open a PR, then rebase this once that is merged

filebeat/filebeat.full.yml
ruflin: So the config option is kind of `field_that_matches_container_id: ...`. What is the use case of having multiple fields defined? 

For the cgroup docker part: I see that this is what it's doing, but why would someone then not just use the docker module instead as he will rely on the docker api.

libbeat/processors/add_docker_metadata/add_docker_metadata.go
ruflin: I kind of assume, the different events will come from different modules or prospectors. Means in this case each prospector should define it's own processor. So there would not be a need to have an array of fields I think.

filebeat/filebeat.full.yml
exekias: We have several metrics matching cgroup id in system (maybe something we have to look into): system.process.cgroup.cpu.id, system.process.cgroup.memory.id, system.process.cgroup.blkio.id

libbeat/processors/add_docker_metadata/watcher.go
exekias: It could make sense yes

filebeat/filebeat.full.yml
ruflin: But I assume per config you only want to match to one?

exekias: I would like to match all of them, perhaps in the future it would be better to merge those fields into a global `system.cgroup.id`, with different meanings depending on the metricset, but as for now I would need to match any possible field there.

what do you think @andrewkroh, does this make any sense?

exekias:
Updated vendor folders, all details in PR description

tsg:
Merging as the tests passed before the rebase, which only fixed the changelog.

PR #4353, opened 2017-05-18
Add varnishlogbeat and varnishstatbeat

phenomenes:


karmi:
Hi @phenomenes, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/beats/pull/4353.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

phenomenes:
@karmi email address updated

ruflin:
@phenomenes Thanks a lot for creating these two beats. 

For varnishbeat it sounds to me like it would fit perfectly into a varnish metricbeat module (which we don't have yet). 

For the varnishlogbeat it sounds like it could fit into filebeat. I was recently working on filebeat an restructuring it to allow also other types of prospectors. One first example here is the redis slowlog prospector: https://github.com/elastic/beats/pull/4180 Perhaps it could also be part of filebeat in the future? Does the varnishlog have a state?

phenomenes:
@ruflin what do you mean by state?

ruflin:
By state I mean if it keeps track somehow on until where it was reading. Like the offset in filebeat with the registry. Or how do you decide where to start reading the log after restarting the beat?

phenomenes:
It doesn't maintain state, is event based.

About varnishstats I haven't fully read metricbeat code yet but I think it could fit.

ruflin:
Means if you restart, you just start to fetch the events that are flowing from this point in time and not from the past.

About the metricbeat module: It would be really great to have a contribution for a varnish module. Here is the dev guide: https://www.elastic.co/guide/en/beats/metricbeat/current/creating-metricsets.html In case you are interested, just ping me if I can help somehow ;-)

phenomenes:
that's exactly what it does, so yes, it maintains state.

cool, I don't have much time right now but I'll have a look as soon as I'm free.

ruflin:
@phenomenes About the state: I need to have a closer look into varnishlogbeat to see what the state structure is.

phenomenes:
To clarify, varnishlogbeat does not maintain state. It will always read starting from the last record, this is handled automatically by Varnish.

ruflin:
Interesting. So if you reconnect it knows where you left off.

PR #4355, opened 2017-05-18
Remove unnecessary print statement in schema apis

vjsamuel:


exekias:
jenkins test it

metricbeat/schema/mapstriface/mapstriface.go
```diff
@@ -128,7 +128,6 @@ func StrFromNum(key string, opts ...schema.SchemaOption) schema.Conv {
 func toStr(key string, data map[string]interface{}) (interface{}, error) {
 	emptyIface, err := common.MapStr(data).GetValue(key)
 	if err != nil {
-		fmt.Println(err)
 		return "", fmt.Errorf("Key %s not found", key)
```
ruflin: The fmt must definitively be removed. But we should add it to the error that is returned:

`return "", fmt.Errorf("Key %s not found: %s", key, err)`

CHANGELOG.asciidoc
```diff
@@ -43,7 +43,7 @@ https://github.com/elastic/beats/compare/v6.0.0-alpha1...master[Check the HEAD d
 - Fix a debug statement that said a module wrapper had stopped when it hadn't. {pull}4264[4264]
 - Use MemAvailable value from /proc/meminfo on Linux 3.14. {pull}4316[4316]
 - Fix panic when events were dropped by filters. {issue}4327[4327]
-
+- Remove unnecessary print statement in schema apis {pull}4355[4355]
```
ruflin: Can you add a newline afterwards?

metricbeat/schema/mapstriface/mapstriface.go
vjsamuel: @ruflin, we do return an error here. it can be left upto the consumer to decide if this needs to be logged or not. for instance I got this error when I had set the `Conv` to optionally extract a field out. It should be logging errors for such cases. that would just fill up my log file because there may be many fields that are set to optional. thought?

CHANGELOG.asciidoc
vjsamuel: done

andrewkroh:
jenkins, test it

ruflin:
jenkins, test it

metricbeat/schema/mapstriface/mapstriface.go
ruflin: I only mean to include the error in the error message.

ruflin:
jenkins, test it

PR #4361, opened 2017-05-19
Bring back prospector.type which went missing

ruflin:
It seems in one of the recent refactorings prospector.type went missing. Test added to make sure it will not go missing again.

### DIFF
--- a/filebeat/tests/system/test_base.py
+++ b/filebeat/tests/system/test_base.py
@@ -1,0 +1,28 @@
+from filebeat import BaseTest
+import os
+
+"""
+Test for basic object
+"""
+
+
+class Test(BaseTest):
+
+    def test_base(self):
+        """
+        Test if the basic fields exist.
+        """
+        self.render_config_template(
+            path=os.path.abspath(self.working_dir) + "/test.log",
+        )
+
+        with open(self.working_dir + "/test.log", "w") as f:
+            f.write("test message\n")
+
+        filebeat = self.start_beat()
+        self.wait_until(lambda: self.output_has(lines=1))
+        filebeat.check_kill_and_wait()
+
+        output = self.read_output()[0]
+        assert "@timestamp" in output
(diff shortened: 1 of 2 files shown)


ruflin:
@vjsamuel As you are working off master, this could interest you.

vjsamuel:
@ruflin, i did notice that `prospector.type` was missing. i thought that i was not looking at it properly :)

filebeat/tests/system/test_base.py
```diff
@@ -0,0 +1,27 @@
...
+
+        with open(self.working_dir + "/test.log", "w") as f:
+            f.write("test message\n")
+
+        filebeat = self.start_beat()
+        self.wait_until(lambda: self.output_has(lines=1))
+        filebeat.check_kill_and_wait()
+
+        output = self.read_output()[0]
+        assert "@timestamp" in output
```
glefloch: Shouldn't we assert that `prospector.type` is in the `output` ?

ruflin: YES. I'm not really confused to not see it here, because that is the reason I added this test. Not sure what I was doing. Thanks for spotting this.

ruflin: Fixed

filebeat/tests/system/test_base.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class Test(BaseTest):
+
+    def test_base(self):
+        """
+        Test if the basic fields exist.
+        """
+        self.render_config_template(
+            path=os.path.abspath(self.working_dir) + "/test.log",
```
andrewkroh: After dealing with weird path issues lately, I recommend using `os.path.join` over string concat to join paths. Or even just putting the `+ "/test.log"` inside the `abspath()` call would fix the problem because `abspath` normalizes the path that it returns.

PR #4363, opened 2017-05-19
Cleanup vendor folders

exekias:
I ran this in all of them:

   $ govendor sync
   $ govendor remove +unused


exekias:
Ok, it seems this removed to many dependencies, closing until further research

PR #4367, opened 2017-05-22
Sort NOTICE file by library path

ruflin:
Currently the NOTICE file was sorted based on the licenses in each glob which overall made it hard to find specific licenses. Now the overall license file is sorted based on the path of each license library. This had the affect that in case a dependency was moved, the NOTICE file changed which should not be the case.

This has also the affect that duplicated license files are removed in case a dependency exists in more then one vendor directory which can be the case. A notice is logged in case a license is already known which helps to identify potential duplicates.

The sorting happens alphabetically ignoring upper / lower case.

### DIFF
--- a/dev-tools/generate_notice.py
+++ b/dev-tools/generate_notice.py
@@ -15,7 +15,7 @@ def read_file(filename):
         return file_content
 
 
-def get_library_name(license):
+def get_library_path(license):
 
     lib = ""
     path = os.path.dirname(license)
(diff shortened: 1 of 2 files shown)


dev-tools/generate_notice.py
```diff
@@ -82,16 +91,16 @@ def create_notice(filename, beat, copyright, licenses):
 
     cwd = os.getcwd()
     notice = os.path.join(cwd, "NOTICE")
-    licenses = []
+    license_globs = []
 
     for root, dirs, files in os.walk(args.vendor):
```
andrewkroh: Can you filter out the `.git` dir? I see it searching inside there when I run it and it really shouldn't be looking in there.

dev-tools/generate_notice.py
```diff
@@ -30,25 +30,34 @@ def get_library_name(license):
...
+    licenses = {}
+    for license_glob in license_globs:
+        for license_file in glob.glob(license_glob):
+            lib_path =  get_library_path(license_file)
+            if lib_path in licenses:
+                print("WARNING: Dependency appears multiple times: {}".format(lib_path))
+            licenses[lib_path] = license_file
 
+    # Sort licenses by package path, ignore upper / lower case
+    for key in sorted(licenses, key=str.lower):
```
andrewkroh: 👍

dev-tools/generate_notice.py
ruflin: I couldn't ignore dot directories for the walk but did ignore all dirs starting with a dot afterwards.

PR #4370, opened 2017-05-22
Filter relative mounts in system filesystem metricset

andrewkroh:
Add filtering to system filesystem metricset to remove relative mountpoints like those from Linux network namespaces.

### DIFF
--- a/metricbeat/module/system/filesystem/helper.go
+++ b/metricbeat/module/system/filesystem/helper.go
@@ -3,10 +3,10 @@
 package filesystem
 
 import (
+	"path/filepath"
 	"time"
 
 	"github.com/elastic/beats/libbeat/common"
-	"github.com/elastic/beats/libbeat/logp"
 	"github.com/elastic/beats/metricbeat/module/system"
 	sigar "github.com/elastic/gosigar"
 )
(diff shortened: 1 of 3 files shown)


metricbeat/module/system/filesystem/helper.go
```diff
@@ -20,13 +20,21 @@ type FileSystemStat struct {
...
 		return nil, err
 	}
 
+	// Ignore relative mount points, which are present for example
+	// in /proc/mounts on Linux with network namespaces.
+	filtered := fss.List[:0]
+	for _, fs := range fss.List {
+		if filepath.IsAbs(fs.DirName) {
+			filtered = append(filtered, fs)
+		}
```
ruflin: Should we add a debug log message to list the once that were ignored?

metricbeat/module/system/filesystem/helper.go
```diff
@@ -89,13 +77,3 @@ func GetFilesystemEvent(fsStat *FileSystemStat) common.MapStr {
 		},
 	}
 }
-
-func GetFileSystemStats() ([]common.MapStr, error) {
```
ruflin: Was not used anymore?

andrewkroh: Yep, not used anywhere.

metricbeat/module/system/filesystem/helper.go
andrewkroh: Added

PR #4372, opened 2017-05-22
Sync vendor directory

ruflin:
Remove vendor directories and run `govendor sync` to see if all files are up-to-date as they should.

Also run `make clean-vendor` afterwards.


vendor/vendor.json
```diff
@@ -154,7 +154,7 @@
 			"revisionTime": "2017-02-07T06:38:51Z"
 		},
 		{
-			"checksumSHA1": "0Yo0jEfGWE40M4U9Z7JNxd/RA/w=",
+			"checksumSHA1": "uUBlYUff9goenMwEP2tUIm9qObs=",
```
andrewkroh: These changes are the result of running `make clean-vendor`. The checksums changed because there is a different set of files in the vendor dir after cleaning.

PR #4374, opened 2017-05-22
scan older files to be harvested based on mod time

waynemz:
config option to scan older files first.

Have a look at issue: https://discuss.elastic.co/t/filebeat-harvester-picks-up-files-in-random-order-when-scanning/86542/4


karmi:
Hi @waynemz, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/beats/pull/4374.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

ruflin:
@waynemz Thanks for sharing your implementation. This is definitively something we should discuss. Some thoughts:

* In case of a large amount of files, this could have quite a performance impact as all files need to be touched to fetch the ModTime().
* This feature should be off by default
* If we introduce this feature, we should allow desc and ascending order as there were also dicussions where people were interested in the most recent files first.
* In general this mainly becomes relevant if `harvester_limit` is used. Assuming `harvester_limit` is 10, it will still not guarantee which of the 10 oldest files is harvested at which speed.
* Some ideas for the config naming if we allow both sort orders?

waynemz:
@ruflin I agree the feature should be off by default given the additional performance impact. Instead of a boolean the config option could be a string with values off/ascending/descending.
I believe some documentation might be needed to explain the performance impact and the fact this would only be most useful in combination with the harvester_limit option, especially when harvester_limit=1
As for the config name I would suggest something like prospector_scan_order  but I'm open to any suggestions you have.

@karmi I added my other email to my github account.

ruflin:
I like the suggestion of `scan_order` but would leave out the prospector part. It goes in line with `scan_frequency`. It's not really the scan that happens in this order but the starting of harvester, but I think that is more a technical thing. `scan_order` should be pretty simple to understand.

I have in the back of my head that if we add this file, pretty soon people will ask for it to be sorted by filename instead of date. Could be perhaps take this already into account now when we design the config. Or is it overkill?

waynemz:
@ruflin Thats a great suggestion. I'll incorporate the filename as well. That will bring the possible values for scan_order to five: off(default), oldest_file_first, newest_file_first, ascending_by_name, descending_by_name
Can you suggest better value strings?

waynemz:
I'll correct the go imports style issue but I could not figure out what went wrong with the test:
[fail] 1.09% test_base.Test.test_invalid_config_cli_param: 0.0296s
I didn't introduce a new cli param here

ruflin:
For the config options I was more thinking of 
```
scan.order: "desc", "asc"
scan.sort: "modtime", "filename", "none"
```

This kind of implies we should switch to `scan.frequency` one day ;-)

ruflin:
For the failing check: Run `make fmt` in the filebeat directory. This should do the cleanup for you.

filebeat/prospector/log/prospector.go
```diff
@@ -7,6 +7,9 @@ import (
 	"path/filepath"
 	"time"
```
ruflin: can you remove newline?

filebeat/prospector/log/prospector.go
```diff
@@ -264,10 +267,55 @@ func (p *Prospector) matchesFile(filePath string) bool {
...
+	for path, info := range paths {
+		sortInfo := FileSortInfo{info: info, path: path}
+		sortInfos = append(sortInfos, sortInfo)
+	}
+
+	return sortInfos
+}
+
+func getSortedFiles(scanOrder string, scanSort string, sortInfos []FileSortInfo) []FileSortInfo {
+	switch scanOrder + "_" + scanSort {
```
ruflin: Could you perhaps to that in two switch / case? So first check the scanSort and then the order inside each? It feels a bit strange to create a string for comparison. Perhaps there is an even "shorter" option.

You could move the `sort.Slice` part to after `switch` and only define the function which will be used inside the switch cases. This will make the cod cleaner I think.

filebeat/prospector/log/prospector.go
```diff
@@ -264,10 +267,55 @@ func (p *Prospector) matchesFile(filePath string) bool {
...
+
+	return sortInfos
+}
+
 // Scan starts a scanGlob for each provided path/glob
 func (p *Prospector) scan() {
 
-	for path, info := range p.getFiles() {
+	paths := p.getFiles()
+	files := getSortInfos(paths)
```
ruflin: This should not happen in every call, only if sorting is enabled as this could have a performance impact. I would expect something like:

`if config.sort.Enabled() { paths := paths.Sort() `

This is heavily simplified but you get what I'm trying to explain? Do we even need to expose the fileinfo outside of the sort?

waynemz:
@ruflin I made the suggested changes. Can you resolve the conflict with changelog?

waynemz:
I fixed the goimports style but there is some build failure with the full yml file. Not sure how to resolve that.

tsg:
@waynemz You can fix that by running `make update`. Now there is also a conflict in the CHANGELOG (sorry, we've merged a large changelog modification) so you need to sync with master.

waynemz:
@ruflin @tsg Fixed all conflicts. Please approve.

tsg:
jenkins, test it

filebeat/prospector/log/prospector.go
```diff
@@ -264,10 +266,107 @@ func (p *Prospector) matchesFile(filePath string) bool {
...
+		switch scanOrder {
+		case "asc":
+			sortFunc = func(i, j int) bool {
+				return sortInfos[i].info.ModTime().Before(sortInfos[j].info.ModTime())
+			}
+		case "desc":
+			sortFunc = func(i, j int) bool {
+				return sortInfos[i].info.ModTime().After(sortInfos[j].info.ModTime())
+			}
+		default:
```
tsg: Would it be possible to return an explicit error on these default branches?

filebeat/prospector/log/prospector.go
```diff
@@ -264,10 +266,107 @@ func (p *Prospector) matchesFile(filePath string) bool {
...
 func (p *Prospector) scan() {
 
-	for path, info := range p.getFiles() {
+	var sortInfos []FileSortInfo
+	var files []string
+
+	paths := p.getFiles()
+	if strings.ToLower(p.config.ScanSort) != "none" {
+		sortInfos = getSortedFiles(strings.ToLower(p.config.ScanOrder),
+			strings.ToLower(p.config.ScanSort),
```
tsg: I'm not a fan of calling `ToLower` here, as this creates some leniency in the config parsing which we don't have in other places.

filebeat/prospector/log/prospector.go
```diff
@@ -264,10 +266,107 @@ func (p *Prospector) matchesFile(filePath string) bool {
...
+	} else {
+		files = getKeys(paths)
+	}
+
+	for i := 0; i < len(paths); i++ {
+
+		var path string
+		var info os.FileInfo
+
+		if strings.ToLower(p.config.ScanSort) != "none" {
```
tsg: Removing `ToLower` here should also mean less work on the default case.

filebeat/docs/reference/configuration/filebeat-options.asciidoc
```diff
@@ -302,6 +302,25 @@ If you require log lines to be sent in near real time do not use a very low `sca
 
 The default setting is 10s.
 
+
+[[scan-sort]]
+===== scan.sort
+
+Specifies if files should be harvested in order and how to determine the order. Possible values are modtime, filename and none. To sort by file modification time use modtime otherwise use filename.
+
+If you specify a value other than none for this setting you can determine whether to use ascending or descending order using `scan.order`
```
ruflin: Could you add `.` dots at the end of the sentences? Also check below.

CHANGELOG.asciidoc
```diff
@@ -132,6 +132,8 @@ https://github.com/elastic/beats/compare/v6.0.0-alpha1...v6.0.0-alpha2[View comm
 - Add `logging.files` `permissions` option. {pull}4295[4295]
 
 *Filebeat*
+- Added ability to sort harvested files. {pull}4374[4374]
+- Add experimental Redis slow log prospector type. {pull}4180[4180]
```
ruflin: This line probably sneaked in during rebasing?

ruflin:
@waynemz I really appreciate all the work you put into this. Thanks a lot for adding this feature.

filebeat/prospector/log/prospector.go
waynemz: @tsg what do you want to happen if the default case is reached? we can abort with an error message (panic), or log an error and use the default behavior.

waynemz:
Thanks @ruflin I appreciate the approval.
@tsg please take a look

waynemz:
Someone please take a look, there are no conflicts as of now.

filebeat/prospector/log/prospector.go
```diff
@@ -264,10 +266,119 @@ func (p *Prospector) matchesFile(filePath string) bool {
...
+	sortInfos := make([]FileSortInfo, 0, len(paths))
+	for path, info := range paths {
+		sortInfo := FileSortInfo{info: info, path: path}
+		sortInfos = append(sortInfos, sortInfo)
+	}
+
+	return sortInfos
+}
+
+func getSortedFiles(scanOrder string, scanSort string, sortInfos []FileSortInfo) ([]FileSortInfo, string) {
```
ruflin: The return argument should be `error` and not `string`. This is quite standard in Golang. Like this you can check 
```
if err != nil {
 ...
}
```

filebeat/prospector/log/prospector.go
```diff
@@ -264,10 +266,119 @@ func (p *Prospector) matchesFile(filePath string) bool {
...
+	var sortFunc func(i, j int) bool = nil
```
ruflin: no need to set it to nil here

filebeat/prospector/log/prospector.go
```diff
@@ -264,10 +266,119 @@ func (p *Prospector) matchesFile(filePath string) bool {
...
+		case "asc":
+			sortFunc = func(i, j int) bool {
+				return sortInfos[i].info.ModTime().Before(sortInfos[j].info.ModTime())
+			}
+		case "desc":
+			sortFunc = func(i, j int) bool {
+				return sortInfos[i].info.ModTime().After(sortInfos[j].info.ModTime())
+			}
+		default:
+			return nil, "Unexpected value for scan.order: " + scanOrder
```
ruflin: you can use `fmt.Errorf("Unexpected value for scan.order: %v", scanOrder)` here

filebeat/prospector/log/prospector.go
```diff
@@ -264,10 +266,119 @@ func (p *Prospector) matchesFile(filePath string) bool {
...
+		case "asc":
+			sortFunc = func(i, j int) bool {
+				return strings.Compare(sortInfos[i].info.Name(), sortInfos[j].info.Name()) < 0
+			}
+		case "desc":
+			sortFunc = func(i, j int) bool {
+				return strings.Compare(sortInfos[i].info.Name(), sortInfos[j].info.Name()) > 0
+			}
+		default:
+			return nil, "Unexpected value for scan.order: " + scanOrder
```
ruflin: see comment above

filebeat/prospector/log/prospector.go
```diff
@@ -264,10 +266,119 @@ func (p *Prospector) matchesFile(filePath string) bool {
...
+		}
+	default:
+		return nil, "Unexpected value for scan.sort: " + scanSort
```
ruflin: see comment above

filebeat/prospector/log/prospector.go
```diff
@@ -264,10 +266,119 @@ func (p *Prospector) matchesFile(filePath string) bool {
...
+	}
+
+	if sortFunc != nil {
+		sort.Slice(sortInfos, sortFunc)
+	}
+
+	return sortInfos, ""
```
ruflin: `return sortInfos, nil`

filebeat/prospector/log/prospector.go
```diff
@@ -264,10 +266,119 @@ func (p *Prospector) matchesFile(filePath string) bool {
...
+}
+
+func getFileState(path string, info os.FileInfo, p *Prospector) file.State {
+	var err error
+	var absolutePath string
+	absolutePath, err = filepath.Abs(path)
+	if err != nil {
+		logp.Err("could not fetch abs path for file %s: %s", absolutePath, err)
```
ruflin: Should we return here on error? Means this also can return an error?

filebeat/prospector/log/prospector.go
```diff
@@ -264,10 +266,119 @@ func (p *Prospector) matchesFile(filePath string) bool {
...
 // Scan starts a scanGlob for each provided path/glob
 func (p *Prospector) scan() {
 
-	for path, info := range p.getFiles() {
+	var sortInfos []FileSortInfo = nil
+	var files []string = nil
+
+	paths := p.getFiles()
+
+	var err string = ""
```
ruflin: `var err error`

filebeat/prospector/log/prospector.go
```diff
@@ -264,10 +266,119 @@ func (p *Prospector) matchesFile(filePath string) bool {
...
+
+	if p.config.ScanSort != "none" {
+		sortInfos, err = getSortedFiles(strings.ToLower(p.config.ScanOrder),
+			strings.ToLower(p.config.ScanSort),
+			getSortInfos(paths))
```
ruflin: We should handle the error directly inside the if clause. In case of `error` should we return? what are we doing?

filebeat/prospector/log/prospector.go
```diff
@@ -264,10 +266,119 @@ func (p *Prospector) matchesFile(filePath string) bool {
...
+		files = append(files, file)
+	}
+	return files
+}
+
 // Scan starts a scanGlob for each provided path/glob
 func (p *Prospector) scan() {
 
-	for path, info := range p.getFiles() {
+	var sortInfos []FileSortInfo = nil
```
ruflin: No need to set it to nil, also follow up below

filebeat/prospector/log/prospector.go
waynemz: In this case we are going to print an error saying why sorting failed and continue the regular unsorted route.

waynemz:
@ruflin @tsg added error handling

filebeat/prospector/log/prospector.go
ruflin: @waynemz Did you see this comment?

filebeat/prospector/log/prospector.go
ruflin: SGTM

waynemz:
@ruflin @tsg fixed it.

ruflin:
@waynemz Thanks for your work on this. LGTM.

Two things that came to my mind which can be done a follow up PR: 

* We should use Validate on the config to validate only valid scan options are set
* We should mark the feature in this Validate part experimental at first so we get some feedback early on and can still make heavy changes on it if needed. 

@waynemz Do you want to add this to this PR or better as a follow up?

ruflin:
@waynemz Merged. Thanks a lot for the work on this one. I will open a follow up PR with the above suggestions.

ruflin:
@waynemz Here is the follow up PR. I made a few minor changes. It would be great to also add some system tests for this feature to confirm it works as expected: https://github.com/elastic/beats/pull/4564

waynemz:
Thanks @ruflin

Issue #4334, opened 2017-05-17
packetbeat 5.x bpf_filter config option not working


clennpillo:
For confirmed bugs, please report:
- Version: 
packetbeat 5.x
- Operating System: 
macosx
- Steps to Reproduce: 
1)add configuration for bpf_filter in packetbeat.yml
2)run packetbeat with option -d "sniffer"
3)check log and found bpf_filter config did not take effect

ruflin:
Could you please share your full config file and the exact packetbeat version you are using?

clennpillo:
I tested packetbeat 5.4.0 and 5.2.1 with following command
./packetbeat -N -d "sniffer"

here is packbeat.yml
#################### Packetbeat Configuration Example #########################

# This file is an example configuration file highlighting only the most common
# options. The packetbeat.full.yml file from the same directory contains all the
# supported options with more comments. You can use it as a reference.
#
# You can find the full configuration reference here:
# https://www.elastic.co/guide/en/beats/packetbeat/index.html

#============================== Network device ================================

# Select the network interface to sniff the data. On Linux, you can use the
# "any" keyword to sniff on all connected interfaces.
packetbeat.interfaces.device: eth0
packetbeat.interfaces.bpf_filter: "net 9.12.203.36"

#================================== Flows =====================================

# Set `enabled: false` or comment out all options to disable flows reporting.
packetbeat.flows:
  # Set network flow timeout. Flow is killed if no packet is received before being
  # timed out.
  timeout: 30s

  # Configure reporting period. If set to -1, only killed flows will be reported
  period: 10s
  enabled: false

#========================== Transaction protocols =============================

packetbeat.protocols.icmp:
  # Enable ICMPv4 and ICMPv6 monitoring. Default: false
  enabled: false

packetbeat.protocols.amqp:
  # Configure the ports where to listen for AMQP traffic. You can disable
  # the AMQP protocol by commenting out the list of ports.
  #ports: [5672]

packetbeat.protocols.cassandra:
  #Cassandra port for traffic monitoring.
  #ports: [9042]

packetbeat.protocols.dns:
  # Configure the ports where to listen for DNS traffic. You can disable
  # the DNS protocol by commenting out the list of ports.
  #ports: [53]

  # include_authorities controls whether or not the dns.authorities field
  # (authority resource records) is added to messages.
  #include_authorities: true

  # include_additionals controls whether or not the dns.additionals field
  # (additional resource records) is added to messages.
  #include_additionals: true

[...]

PR #4378, opened 2017-05-23
Fix parsing of interface options with _ (#4334)

zecke:
In commit 55470602d1459ae6312fdec26b85c1c4d8da8d6c linting issues
were addresses and variables containing _ were renamed. This broke
the config parsing.

In packetbeat this seems to effect the with_vlans, bpf_filter and
the buffer_size_mb options. Correct it by using a tag in the structure.

### DIFF
--- a/packetbeat/config/config.go
+++ b/packetbeat/config/config.go
@@ -20,13 +20,13 @@ type Config struct {
 }
 
 type InterfacesConfig struct {
-	Device       string
-	Type         string
-	File         string
-	WithVlans    bool
-	BpfFilter    string
-	Snaplen      int
-	BufferSizeMb int
+	Device       string `config:"device"`
+	Type         string `config:"type"`
+	File         string `config:"file"`
+	WithVlans    bool   `config:"with_vlans"`
+	BpfFilter    string `config:"bpf_filter"`
+	Snaplen      int    `config:"snaplen"`
+	BufferSizeMb int    `config:"buffer_size_mb"`
 	TopSpeed     bool
 	Dumpfile     string
 	OneAtATime   bool


zecke:
Sorry heading out for tonight but wanted to share my findings in regard to issue #4334.

* There are more variables that might be used in the config file (Src_ip, Dst_ip, proc_prefix)
* To test one could probably easily have a round-trip test? Parse the config and write what was parsed and check there is no difference?

ruflin:
@zecke Thanks a lot for bringing this up. If you are right, it seems that this could affect quite a few more variables.

tsg:
Hmm, yeah, looks like this broke with https://github.com/elastic/beats/pull/2928. :(

@zecke Thanks a lot for finding & fixing this. Can you add `config` tags to the rest of the `InterfaceConfig` struct? All the structs in `config.go` are used in configurations, but only that struct is lacking tags.

Our system tests didn't catch these because they always work on reading from files, rather than opening an interface.

zecke:
@tsg Shouldn't bpf_filter work for reading from files too? I have not looked at the test setup but setting a filter that should only match a subset of the packets in a file and verifying the number of imported packets should be doable?

I annotated the struct (and aligned the tags to match the longest one). I have left the last options out as they are not documented in the reference file and I didn't know what is being used (top_speed, topspeed, topSpeed)

clennpillo:
Thanks for fixing this, this was preventing us from monitoring a busy interface, I guess this PR will be merged soon ?

tsg:
@zecke Currently it's not applied on reading from file. We could implement it, but then it would probably be a different setting, because this one applies to "interfaces"?

I'm going to merge it for now like this and improve on the system tests if something like this ever surfaces again. I'll add the Changelog while backporting, thanks a lot for the fix!

zecke:
@tsg "interfaces": True, let's not do that then.

In https://github.com/zecke/beats/commit/73e275db4280aca127b496d9cf6ece178b367dd8 I created a new interface (to allow easy porting of the existing beats?!) and played a bit with printing a yaml representation of what has been parsed. It prints something like this...

I think in general it is a nice addition:

1. It might help users to find typos in their config file (in case something is not picked up?)
2. We could somehow integrate it into the system tests and just compare text output?

What do you think?

```
interfaces:
  device: en0
  type: pcap
  file: ""
  withvlans: false
  bpffilter: ""
  snaplen: 65535
  buffersizemb: 0
  topspeed: false
  dumpfile: ""
  oneatatime: false
  loop: 1
flows:
  enabled: false
  timeout: 30s
  period: 10s
protocols:
  amqp: {}
  cassandra: {}
  dns: {}
  http: {}
  icmp: {}
...
runoptions:
  uid: null
  gid: null
```

Issue #2581, opened 2016-09-19
Disable useless HTML escaping when marshaling JSON


magnusbaeck:
Go's HTML marshaler escapes ampersand and angle brackets by default, i.e.

```
{"message": "<b>hello world</b>"}
```

becomes this:

```
{"message": "\u003cb\u003ehello world\u003c/b\u003e"}
```

This is done to keep browsers from interpreting the JSON text as HTML, but that's hardly useful in the Beats case. Go 1.7 introduced an option to disable this kind of escaping (see [json.Encoder.SetEscapeHTML](https://golang.org/pkg/encoding/json/#Encoder.SetEscapeHTML)), and now that Beats has switched to Go 1.7 we might as well disable this behavior. This should have a marginally positive performance impact and could decrease the payload size at least when compression is disabled.

See also https://discuss.elastic.co/t/beats-encodes-angle-brackets-and-as-u003c-and-u003e-in-json-output/60667.

christiangalsterer:
+1

ruflin:
Anyone wants to open a PR for it? :-D

lerencao:
+1

andrewkroh:
Here's another Discuss issue where this led to confusion. https://discuss.elastic.co/t/filebeat-character-encoding-problem/136486

monkkitarp:
+1 (data integrity)

PR #4379, opened 2017-05-23
Disable HTML escaping during JSON Marshaling

ruflin:
By default the golang JSON Marshaller escapes HTML entries. As this is not required for beats this is disabling this option.

* Unification of json marshalling

Closes https://github.com/elastic/beats/issues/2581

* Add changelog

### DIFF
--- a/libbeat/common/json.go
+++ b/libbeat/common/json.go
@@ -1,0 +1,28 @@
+package common
+
+import (
+	"bytes"
+	"encoding/json"
+
+	"fmt"
+)
+
+// JSONEncode encodes the given interface to JSON
+func JSONEncode(data interface{}, pretty bool) ([]byte, error) {
+
+	buffer := &bytes.Buffer{}
+	enc := json.NewEncoder(buffer)
+	enc.SetEscapeHTML(true)
+
+	enc.SetIndent("", "")
+	if pretty {
+		enc.SetIndent("", "  ")
+	}
+
+	err := enc.Encode(data)
+	if err != nil {
+		return nil, fmt.Errorf("failed to convert the data to JSON (%v): %#v", err, data)
+	}
+
+	return buffer.Bytes(), nil
+}
--- a/libbeat/logp/file_rotator.go
+++ b/libbeat/logp/file_rotator.go
@@ -67,7 +67,6 @@ func (rotator *FileRotator) WriteLine(line []byte) error {
 		}
 	}
 
-	line = append(line, '\n')
 	_, err := rotator.current.Write(line)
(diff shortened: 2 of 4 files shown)


libbeat/logp/file_rotator.go
```diff
@@ -67,7 +67,6 @@ func (rotator *FileRotator) WriteLine(line []byte) error {
 		}
 	}
 
-	line = append(line, '\n')
```
ruflin: It is not 100% clear to me what this change was needed. Seems like before there was no newline at the end of each event and now there is.

libbeat/common/json.go
```diff
@@ -0,0 +1,28 @@
+package common
+
+import (
+	"bytes"
+	"encoding/json"
+
```
andrewkroh: Remove newline.

libbeat/common/json.go
```diff
@@ -0,0 +1,28 @@
...
+)
+
+// JSONEncode encodes the given interface to JSON
+func JSONEncode(data interface{}, pretty bool) ([]byte, error) {
+
+	buffer := &bytes.Buffer{}
+	enc := json.NewEncoder(buffer)
+	enc.SetEscapeHTML(true)
+
+	enc.SetIndent("", "")
```
andrewkroh: Is this the default which would make this line unnecessary?

libbeat/common/json.go
```diff
@@ -0,0 +1,28 @@
...
+	if pretty {
+		enc.SetIndent("", "  ")
+	}
+
+	err := enc.Encode(data)
+	if err != nil {
+		return nil, fmt.Errorf("failed to convert the data to JSON (%v): %#v", err, data)
```
andrewkroh: You could use errors.Wrapf here.

libbeat/logp/file_rotator.go
andrewkroh: A quick test comparing json.Marshal() to common.JSONEncode() would be interesting to see if they produce the same output or a if one has a newline.

Is this code also used by the logp to write files? Does this affect it?

ruflin:
@andrewkroh I also don't think we should merge this without having figured out where the newline change is coming from. Unfortunately will probably not have time for it this week.

ruflin:
I'm going to close this PR as I never found the time to further investigate and the code base has changed quite a bit since then so a restart makes more sense.

@urso I think disabling the escaping would still be the way to go. Also having one method to Marshal json would be nice in libbeat so potential optimisations could be made in a central place. I leave it up to you if you want to pursue this further.

PR #4380, opened 2017-05-23
Improve `make notice` to work on libs not 3 dirs long

tsg:
Did this refactoring in preparation for a change that adds the version in,
but opening it as its own PR so we can verify the diff. There were a few
libs that were not exactly "3 dirs long".

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -1962,6 +1962,37 @@ Apache License
 
 
 --------------------------------------------------------------------
+github.com/vmware/govmomi/vim25/xml
+--------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


ruflin:
@tsg Interesting, it seems `make check` gets a different notice file ...

tsg:
Ah, it has something to do with libraries that show up multiple times. Investigating.

tsg:
This last commit sort of fixes the effect of a bigger issue: if there are multiple LICENSE files in the same library (for example, mgo has `LICENSE` and `LICENSE.libyaml`), we only include one of them. Correct would be to add them all, but that's for another PR.

PR #4381, opened 2017-05-23
Harvester setup cleanup

ruflin:


ruflin:
Closing for now as not a priority. Cleanup of the harvester setup should still happen to make a cleaner separation between stdin and log possible.

PR #4383, opened 2017-05-23
Add versions to the NOTICE file

tsg:
This does a few changes to the NOTICE file:

* Include version (when available) and git commit hash (revision) in the NOTICE file. This is currently taken from the `vendor.json` produced by the govendor tool, but other vendoring tools should be easily supported with small code modifications.
* Include duplicate libraries (e.g. same lib in top level vendor and in metricbeat vendor). This is needed because these could potentially have different versions.
* Include all LICENSE files if a library has more than one
* Include the local path to the LICENSE file, which makes auditing the NOTICE file easier
* Exclude the `dev-tools` folder since the dependencies there are not shipped.

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -10,7 +10,9 @@ Third party libraries used by the Beats project:
 
 
 --------------------------------------------------------------------
-github.com/andrewkroh/sys
+Dependency: github.com/andrewkroh/sys
+Revision: 287798fe3e430efeb9318b95ff52353aaa2b59b1
+./vendor/github.com/andrewkroh/sys/LICENSE:
 --------------------------------------------------------------------
 Copyright (c) 2009 The Go Authors. All rights reserved.
 
@@ -41,7 +43,9 @@ THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 --------------------------------------------------------------------
-github.com/armon/go-socks5
+Dependency: github.com/armon/go-socks5
+Revision: e75332964ef517daa070d7c38a9466a0d687e0a5
+./vendor/github.com/armon/go-socks5/LICENSE:
 --------------------------------------------------------------------
 The MIT License (MIT)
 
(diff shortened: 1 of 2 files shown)


NOTICE
```diff
@@ -65,29 +69,9 @@ IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
 CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 --------------------------------------------------------------------
-github.com/cavaliercoder/badio
```
ruflin: I wonder why this LICENSES disappeared?

dev-tools/generate_notice.py
```diff
@@ -102,6 +134,11 @@ def create_notice(filename, beat, copyright, vendor_dirs):
 
         if 'vendor' in dirs:
             vendor_dirs.append(os.path.join(root, 'vendor'))
+            dirs.remove('vendor')   # don't walk down into sub-vendors
```
ruflin: Is there a reason for not doing this?

NOTICE
tsg: I excluded the `dev-tools` folder, since those are not run-time dependecies. That one is needed for the test we have on packages.

dev-tools/generate_notice.py
tsg: It results in duplicates, because when we look for LICENSE files, we walk down at any level, thus catching nested vendored deps.

ruflin: Ok, I was thinking of vendor directories inside the vendor dirs. I remember seeing that somewhere.

ruflin:
@tsg There seems to be a conflict now.

dev-tools/generate_notice.py
tsg: Right, "vendor in vendor" is what it's being excluded here, but those deps are still found on the walk from line 42.

tsg:
@ruflin I rebased, and had to do some changes for the Python 3 support. I tested in both py2 and 3.

dev-tools/generate_notice.py
```diff
@@ -30,55 +34,86 @@ def get_library_path(license):
...
     for vendor in vendor_dirs:
+        libs = read_versions(vendor)
+
         # walk looking for LICENSE files
         for root, dirs, filenames in os.walk(vendor):
             for filename in sorted(filenames):
                 if filename.startswith("LICENSE"):
                     lib_path = get_library_path(root)
-                    if lib_path in licenses:
-                        print("WARNING: Dependency appears multiple times: {}".format(lib_path))
```
ruflin: I will miss that feature to figure out which dependencies exist more then once.

PR #4389, opened 2017-05-24
Do not activate memcache by default (#4335)

zecke:
The documentation states that the memcache proto can be disabled
by commenting out the list of ports. In practice one had to comment
out the entire line.

Fix it by not initializing the Ports variable inside the
defaultConfig of the memcache proto. No other proto has Ports
initialized in the defaultConfig.


tsg:
@zecke Can you add a changelog line, please? Thanks for the fix!

zecke:
Does the test failure look unrelated to you? I don't seem to be able to restart the job, can you?

ruflin:
jenkins, test it

PR #4395, opened 2017-05-24
Add a CSV summary for the dependencies

tsg:
* Adds auto-detection for the common license types
* Adds option to create a CSV file summary. This is convenient for creating
  spreadsheets and the like

### DIFF
--- a/dev-tools/generate_notice.py
+++ b/dev-tools/generate_notice.py
@@ -5,6 +5,8 @@ import os
 import datetime
 import argparse
 import json
+import csv
+import re
 
 
 def read_file(filename):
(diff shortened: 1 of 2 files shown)


dev-tools/generate_notice.py
```diff
@@ -33,7 +35,16 @@ def get_library_path(license):
     return "/".join(split)
 
 
-def add_licenses(f, vendor_dirs):
+def read_versions(vendor):
+    libs = []
+    with open(os.path.join(vendor, "vendor.json")) as f:
```
ruflin: It adds dependencies on `govendor` but I think that is fine for now as I'm still quite happy with govendor.

dev-tools/generate_notice.py
```diff
@@ -69,47 +103,89 @@ def add_licenses(f, vendor_dirs):
...
             f.write("--------------------------------------------------------------------\n")
-            copyright = read_file(lib["license_file"])
-            if "Apache License" not in copyright:
-                f.write(copyright)
+            if lib["license_summary"] != "Apache License 2.0":
+                f.write(lib["license_contents"])
             else:
                 # it's an Apache License, so include only the NOTICE file
-                f.write("Apache License\n\n")
+                f.write("Apache License 2.0\n\n")
```
ruflin: Probably would make sense to use a constant here

ruflin:
@exekias kubelete tests were flaky in this one.

ruflin:
@tsg Looks like we could start to push some stats about the beats like dependencies etc. to ES / Kibana so we see how we change it over time.

Issue #861, opened 2016-01-27
Provide more detailed documentation on log levels


ppf2:
Filing based on user experience in the field.

https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-configuration-details.html#_level

> level
> Minimum log level. One of debug, info, warning, error or critical. If debug is used, but no selectors are configured, the \* selector will be used. The default log level is error.

It looks like if everything is running fine, and no errors are logged, we do not even create the log file (at the default error level).  This can be misleading so that end users may think that their logging configuration is actually incorrect.  It will be nice to document this behavior.

At the debug level, it will print out all the events that are being flushed by filebeat, which is quite a useful debugging tool.  And if someone wants just the # of events published (to see that it is actually doing something), level: info is the one to use.  Perhaps we can also mention this as well. Thx!

ruflin:
@ppf2 Thanks a lot for bringing this up and also directly making the suggestions for improvements. We should definitively improve this part of the docs as we also put quite a bit of work in improving the logging part recently.

@dedemorton FYI

dedemorton:
Closed by #4405. Sorry it took me so long to get around to this. I think some of the changes to default behavior (such as having `info` instead of `error` as the default level) has probably helped clear up some of the confusion expressed in this issue.

PR #4405, opened 2017-05-26
Provide more detail on log levels

dedemorton:
This issue adds more detail about log levels as requested in https://github.com/elastic/beats/issues/861.

I'm still being a bit vague in some of the descriptions (especially for debug) because I don't want to expose details about beat processing that might change over time.

Other changes:
- Had to do a bit of refactoring of anchor IDs to add a link to the topic about command line options.
- Removed defaults listed in to_file and to_syslog because they seemed to contradict what we say in the intro about the default logging output being platform-specific. 
- Didn't really understand what we meant by the following sentence, so I rewrote it to say what I think we mean: "In addition to the logging system, the logging output configuration can be
modified from the command line."

### DIFF
--- a/filebeat/docs/command-line.asciidoc
+++ b/filebeat/docs/command-line.asciidoc
@@ -1,4 +1,4 @@
-[[filebeat-command-line]]
+[[command-line-options]]
 === Command Line Options
 
 The following command line option is specific to Filebeat.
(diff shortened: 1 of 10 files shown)


dedemorton:
This should be backported to 5.4.

PR #4406, opened 2017-05-26
Change haproxy.stat.check.health.last to string from long

dariusj18:
My metricbeat sends values like "HTTP status check returned code <200>" in that field

### DIFF
--- a/metricbeat/docs/fields.asciidoc
+++ b/metricbeat/docs/fields.asciidoc
@@ -5188,7 +5188,7 @@ Time in ms that it took to finish the last health check.
 [float]
 === haproxy.stat.check.health.last
 
-type: long
+type: string
 
 
 


andrewkroh:
That file is generated based on the contents of `https://github.com/elastic/beats/blob/master/metricbeat/module/haproxy/stat/_meta/fields.yml#L301`. After changing that file, you can run `make update` in the metricbeat dir and it will rebuild the all the derived files.

dariusj18:
Ahh, I see, I'm going to submit this as an issue, because there are things about it I don't understand well enough to make the commit myself. I didn't realize that until a second ago.

https://github.com/elastic/beats/issues/4407

tsg:
I'll take over this one.

Issue #4409, opened 2017-05-29
Import/export dashboards via the Kibana API


monicasarbu:
Thanks to https://github.com/elastic/kibana/pull/10858, a new API for importing and exporting Beats dashboards is added. 

The export API receives a list of dashboards ids to be exported into a single JSON file. The JSON file contains not only the dashboards, but also the visualizations, searches, and the index pattern (by default). This allows us to group all the dashboards (including visualizations and searches) of a module into a single JSON file, that makes it easier to import using the import API. The Beats are using the export API only to export the dashboards, including the dependencies, without the index pattern that is separately generated by each Beat.

The import API receives the JSON file to be imported.

Thinks that still need to be done (after the #4413 is merged):
- [x] TLS config
- [x] Configure username/password for Kibana
- [x] Add make import-dashboards 
- [ ] Add make export-dashboards
- [ ] fix test `es_loader_test` to test with an older of version of Elasticsearch
- [x] add system test in libbeat to load the dashboards in Kibana
- [x] export dashboards in the new format for the other Beats
- [ ] redo mysql and golang dashboards to use time series visual builder
- [ ] Filebeat auditd dashboard fails with an error (https://github.com/elastic/beats/issues/4595)
- [ ] Update docs 
- [ ] Add export dashboards option to the Beat executable https://github.com/elastic/beats/pull/4725
- [x] Create a `module.yml` for all the modules, not only for those that contain more than one dashboard

kvch:
This issue is for an old version of beats, I suggest to try out the latest version of beats. If this is still a problem in the newly released version feel free to reopen this issue. Thanks.

PR #4413, opened 2017-05-29
Use import/export API for loading dashboards and index pattern

monicasarbu:
This is trying to implement https://github.com/elastic/beats/issues/4409. 

The Beats 6.x dashboards are different than the Beats 5.x dashboards and are not compatible. Thanks to the Kibana API, we can have a json file for each module instead of multiple json files (for dashboards, visualizations, searches, etc). Thus, the way to import/export dashboards differs from the Kibana version that should be equal to the Elasticsearch version.

The json file containing the index-pattern contain as `version` the Beat version, and the dashboards contain as `version` the Kibana version that exported the dashboards.

If Kibana version is 5.x and older: 
- The dashboards should be imported/exported directly from Elasticsearch, and it works only if Elasticsearch is set as output
- All the Beats dashboards are moved to  the `_meta/kibana/5.x` directory 

If Kibana version is 6.x and greater:
- The dashboards should be imported/exported via the Kibana API, so the user needs to configure the kibana url under the `setup.kibana*` settings.
- All the Beats 6.x dashboards should be added to the `_meta/kibana/default` directory for each Beat module.

In order to make use of the import/export API the following changes are needed:
- [x] move all the old dashboards, search, visualizations, index-pattern to `_meta/kibana/5.x` directory
and create a new directory under `_meta/kibana/default` for the new dashboards and index-pattern that are importable via the Kibana API
- [x] update generate_index_pattern to generate the index-pattern for 5.x and >= 6.x versions
- [x] use the Kibana API to load the dashboards and the index pattern for Kibana >= 6
- [x] update import_dashboards script
- [x] update `make update`
- [x] update integration tests

Note: The older dashboards are moved to the `_meta/kibana/5.x` directory only for Metricbeat for now.


ruflin:
@monicasarbu Should this also go under `dev-tools/cmd/*` like all the other commands?

monicasarbu:
@ruflin This PR is at a very early stage. My plan is to move the import and export functionality to libbeat and then create a script to use it. For now it's just a script (that I will remove) only for testing the import/export API.

tsg:
I think we need to add `setup.kibana` to the short config as well, using `localhost` by default.

monicasarbu:
@tsg Fixed, thanks!

libbeat/dashboards/dashboards.go
```diff
@@ -26,19 +19,104 @@ func ImportDashboards(beatName, beatVersion string, esClient DashboardLoader, cf
...
+}
+
+func ImportDashboardsViaKibana(config *common.Config, dashConfig *DashboardsConfig) error {
+
+	kibanaLoader, err := NewKibanaLoader(config, dashConfig, nil)
+	if err != nil {
+		return fmt.Errorf("fail to create the Kibana loader: %v", err)
+	}
+
+	if kibanaLoader == nil {
```
tsg: Seems that etiher `err == nil` or `kibanaLoader != nil`, so I think the second check here is not needed and it doesn't follow our usual error checking pattern.

libbeat/dashboards/kibana_loader.go
```diff
@@ -0,0 +1,97 @@
...
+	content, err := ioutil.ReadFile(file)
+	if err != nil {
+		return fmt.Errorf("fail to read index-pattern: %v", err)
+	}
+
+	return loader.client.ImportJSON(importAPI, params, bytes.NewBuffer(content))
+}
+
+func (loader KibanaLoader) Close() error {
+	//return loader.client.Close()
```
tsg: Seems that there's nothing to do on Close. Then you can just delete the comment, not to confuse future readers.

libbeat/setup/kibana/client.go
```diff
@@ -0,0 +1,141 @@
...
+	kibanaURL, err := common.MakeURL(config.Protocol, config.Path, config.Host, 5601)
+	if err != nil {
+		logp.Err("Invalid host param set: %s, Error: %v", config.Host, err)
+		return nil, fmt.Errorf("invalid Kibana host: %v", err)
+	}
+
+	logp.Debug("kibana", "Kibana url: %s", kibanaURL)
+
+	client := &Client{
+
```
tsg: Minor issue, but this extra empty line could be deleted.

libbeat/setup/kibana/client.go
```diff
@@ -0,0 +1,141 @@
...
+		Connection: Connection{
+			URL: kibanaURL,
+			http: &http.Client{
+				Transport: &http.Transport{
+					TLSClientConfig: &tls.Config{InsecureSkipVerify: true}, // ignore expired SSL certificates
```
tsg: Hmm, does this mean that HTTPS is always enabled? The comment looks like "insecure by default", we have to be careful with this.

libbeat/dashboards/kibana_loader.go
```diff
@@ -0,0 +1,97 @@
...
+	config       *DashboardsConfig
+	version      string
+	msgOutputter *MessageOutputter
+}
+
+func NewKibanaLoader(cfg *common.Config, dashboardsConfig *DashboardsConfig, msgOutputter *MessageOutputter) (*KibanaLoader, error) {
+
+	if cfg == nil {
+		cfg = common.NewConfig()
+		err := cfg.SetBool("enabled", -1, true)
```
tsg: "enabled" is forced to true here, but I didn't see where it's actually checked. We should make sure that disabling with `enabled: false` works.

libbeat/setup/kibana/client.go
```diff
@@ -0,0 +1,141 @@
...
+func NewKibanaClient(cfg *common.Config) (*Client, error) {
+
+	config := defaultKibanaConfig
+	if err := cfg.Unpack(&config); err != nil {
+		return nil, err
+	}
+
+	kibanaURL, err := common.MakeURL(config.Protocol, config.Path, config.Host, 5601)
+	if err != nil {
+		logp.Err("Invalid host param set: %s, Error: %v", config.Host, err)
```
tsg: We generally avoid `logp.Err` immediately followed by returning an error.

libbeat/setup/kibana/client.go
```diff
@@ -0,0 +1,141 @@
...
+		},
+	}
+	return client, nil
+}
+
+func (conn *Connection) Request(method, extraPath string, params url.Values, body io.Reader) (int, []byte, error) {
+
+	reqURL, err := addPath(conn.URL, extraPath)
+	if err != nil {
+		return 0, nil, fmt.Errorf("fail to parse URL %s: %v", conn.URL, err)
```
tsg: Would be perhaps better to write the "fail to parse URL" error in the `addPath` method?

libbeat/setup/kibana/client.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	logp.Debug("kibana", "HTTP request URL: %s", reqURL)
+
+	req, err := http.NewRequest(method, reqURL, body)
+	if err != nil {
+		logp.Err("Fail to create a HTTP %s request: %v", method, err)
```
tsg: Return error instead of logging.

libbeat/setup/kibana/client.go
```diff
@@ -0,0 +1,141 @@
...
+		return 0, nil, err
+	}
+
+	req.Header.Set("Content-Type", "application/json")
+
+	resp, err := conn.http.Do(req)
+	if err != nil {
+		logp.Err("Fail to execute a HTTP %s request: %v", method, err)
```
tsg: same here.

libbeat/setup/kibana/client.go
```diff
@@ -0,0 +1,141 @@
...
+	type kibanaVersionResponse struct {
+		Name    string `config:"name"`
+		Version struct {
+			Number string `config:"number"`
+		} `config:"version"`
+	}
+
+	_, result, err := client.Connection.Request("GET", "/api/status", nil, nil)
+	if err != nil {
+		return "", fmt.Errorf("HTTP GET request to /api/status fails: %v", err)
```
tsg: Please add the `result` to the error message.

libbeat/setup/kibana/client.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	var kibanaVersion kibanaVersionResponse
+	err = json.Unmarshal(result, &kibanaVersion)
+	if err != nil {
+		return "", fmt.Errorf("fail to unmarshall the HTTP response from Kibana %s: %v", client.Connection.URL, err)
```
tsg: `unmarshal` with one `l`.

libbeat/setup/kibana/client.go
```diff
@@ -0,0 +1,141 @@
...
+	return kibanaVersion.Version.Number, nil
+}
+
+func (client *Client) ImportJSON(url string, params url.Values, body io.Reader) error {
+
+	statusCode, response, err := client.Connection.Request("POST", url, params, body)
+	if err != nil {
+		return fmt.Errorf("%v. Response: %s", err, response)
+	}
+	if statusCode != 200 {
```
tsg: 200 might be too strict here. Maybe `> 300`?

libbeat/dashboards/kibana_loader.go
```diff
@@ -0,0 +1,97 @@
...
+	"io/ioutil"
+	"net/url"
+
+	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/logp"
+	"github.com/elastic/beats/libbeat/setup/kibana"
+)
+
+var importAPI = "/api/kibana/dashboards/import"
+var exportAPI = "/api/kibana/dashboards/export"
```
tsg: `exportAPI` doesn't seem to be used.

libbeat/scripts/Makefile
```diff
@@ -305,16 +306,8 @@ index-template: ## @build Generate index templates for the given $VERSION. This
 ### KIBANA FILES HANDLING ###
 ES_URL?=http://localhost:9200
 
-.PHONY: export-dashboards
-export-dashboards: python-env update
-	. ${PYTHON_ENV}/bin/activate && python ${ES_BEATS}/dev-tools/export_dashboards.py --url ${ES_URL} --dir $(shell pwd)/_meta/kibana --regex ${BEAT_NAME}-*
```
tsg: Could we get `export-dashboards` and `import-dashboards` targets based on the go code? I'm good with follow up PR for them.

libbeat/setup/kibana/client.go
monicasarbu: This is something that we need to address in the next PR.

libbeat/setup/kibana/client.go
monicasarbu: fixed

libbeat/setup/kibana/client.go
monicasarbu: fixed

libbeat/dashboards/kibana_loader.go
monicasarbu: fixed

libbeat/setup/kibana/client.go
monicasarbu: fixed

libbeat/dashboards/kibana_loader.go
monicasarbu: fixed

libbeat/setup/kibana/client.go
monicasarbu: fixed

libbeat/setup/kibana/client.go
monicasarbu: fixed

libbeat/setup/kibana/client.go
monicasarbu: fixed

libbeat/setup/kibana/client.go
monicasarbu: fixed

libbeat/setup/kibana/client.go
monicasarbu: true, fixed

libbeat/dashboards/kibana_loader.go
monicasarbu: fixed

libbeat/scripts/Makefile
monicasarbu: It's a bit tricky for `export-dashboards` as we need to pass the list of ids that we want to export in one json file. Usually you would run it to export the dashboards for a module, not for a Beat. Added to the TODO list in the meta issue: https://github.com/elastic/beats/issues/4409

libbeat/setup/kibana/client.go
monicasarbu: Added to the TODO list: https://github.com/elastic/beats/issues/4409

exekias:
I think it would make sense to add `dashboards` to the existing `export` command, I think users could use that to share easily their own dashboards

monicasarbu:
@ruflin yes, that's the plan to freeze the 5.x dashboards and improve only the 6.x dashboards.

.gitignore
```diff
@@ -22,6 +22,9 @@ coverage.out
 *.swo
 *.swn
 
+# Executable
```
monicasarbu: Good idea, done.

auditbeat/auditbeat.reference.yml
```diff
@@ -664,6 +664,25 @@ output.elasticsearch:
...
+
+# Starting with Beats version 6.0.0, the dashboards are loaded via the Kibana API. 
+# This requires a Kibana endpoint configuration.
+setup.kibana:
+
+  # Kibana Host
+  # Scheme and port can be left out and will be set to the default (http and 5601)
+  # In case you specify and additional path, the scheme is required: http://localhost:5601/path
+  # IPv6 addresses should always be defined as: https://[2001:db8::1]:5601
+  # host: "localhost:5601"
```
monicasarbu: done!

libbeat/dashboards/dashboards.go
```diff
@@ -26,19 +19,107 @@ func ImportDashboards(beatName, beatVersion string, esClient DashboardLoader, cf
...
+	if err := importer.Import(); err != nil {
+		return false, fmt.Errorf("fail to import the dashboards in Elasticsearch: %v", err)
+	}
+
+	return true, nil
+
+}
+func getMajorVersion(version string) (int, error) {
+
+	majorVersion := string(version[0])
```
monicasarbu: done

tsg:
jenkins, test it

tsg:
jenkins, test it

tsg:
jenkins, test it

PR #4415, opened 2017-05-29
Allow string characters in user agent patch version

### LINKED ISSUE #3836: Filebeat Apache module cannot assume browser patch version to be a number
Version: Filebeat 5.3.0

In https://github.com/elastic/beats/blob/master/filebeat/module/apache2/access/_meta/fields.yml, we assume the browser patch version is a number:

```
        - name: patch
          type: long
          description: >
            The patch version of the user agent.
```

That is not a safe assumption, e.g. Firefox beta versions had a patch versions such as "b12". A full user agent string then looks like this:

`Mozilla/5.0 (X11; Linux x86_64; rv:2.0b12) Gecko/20110222 Firefox/4.0b12`

There are a bunch of them, see for example http://www.useragentstring.com/pages/useragentstring.php?name=Firefox

I'm not 100% sure if the other components of the user agent version string  - major, minor and build - are never not numbers. However, I've just ingested 300K log lines successfully and no error came up.

exekias:
Both for NGINX and Apache logs

Fixes #3836 

This changes the mapping, what's our policy for these cases?

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -37,6 +37,7 @@ https://github.com/elastic/beats/compare/v6.0.0-alpha1...master[Check the HEAD d
 *Filebeat*
 - Fix race condition on harvester stopping with reloading enabled. {issue}3779[3779]
 - Fix recursive glob config parsing and resolution across restarts. {pull}4269[4269]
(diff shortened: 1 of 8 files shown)


exekias:
kubelet metricset failed again here, tests are flaky, I have to look into that

PR #4417, opened 2017-05-29
Nginx module: use first not private IP address as remote_ip

### LINKED ISSUE #4322: Support multiple IPs in nginx module
Currently the nginx module only allows to fetch one IP, but if you use a proxy, you might want to output the `X-Forwarded-For` header into the logs, which would result in lines containing multiple IPs. Here is an example from our log file (with random client IPs).
```
68.75.44.178, 172.68.146.54, 127.0.0.1 - - [15/May/2017:12:16:27 +0200] "GET /sites/default/files/styles/company_profile_cover_crop/public/1500x500_1_10.jpg?itok=RUgim2UQ&sc=297009042628d7de3f0eb50e807d29e4 HTTP/1.1" 200 92763 "https://www.startus.cc/company/finleap" "Mozilla/5.0 (Windows NT 6.1; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/58.0.3029.110 Safari/537.36"
221.247.242.171, 162.158.166.51, 127.0.0.1 - - [15/May/2017:12:16:27 +0200] "GET /sites/default/files/styles/company_profile_logo/public/company_logos/aaeaaqaaaaaaaawvaaaajdk3n2vkzme0lte0zjctngy3ms1inmm4lta4ntnhzwqymzvmoq.png?itok=H2B05xX0 HTTP/1.1" 200 9296 "https://www.startus.cc/company/finleap" "Mozilla/5.0 (Windows NT 6.1; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/58.0.3029.110 Safari/537.36"
137.56.184.63, 162.158.165.50, 127.0.0.1 - - [15/May/2017:12:16:27 +0200] "GET /sites/default/files/styles/company_profile_cover/public/1500x500_1_10.jpg?itok=1cNqdGYK HTTP/1.1" 200 102268 "https://www.startus.cc/company/finleap" "Mozilla/5.0 (Windows NT 6.1; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/58.0.3029.110 Safari/537.36"
92.222.165.172, 162.158.167.202, 127.0.0.1 - - [15/May/2017:12:16:27 +0200] "POST /jstats.php HTTP/1.0" 200 13 "https://www.startus.cc/company/finleap" "Mozilla/5.0 (Windows NT 6.1; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/58.0.3029.110 Safari/537.36"
```

In our case the first IP is the actual client IP, the second is the one from cloudflare, and the localhost address comes from varnish, which runs on the same host as nginx. 

Filebeat only takes logs the localhost address from varnish i.e. `127.0.0.1`, which is kind of useless. It would be cool to have all IPs to store all IPs into elastic search, but I guess it raises some questions regarding how to geocode them.
As a quick hack I tried to create a custom pattern, which fetches the first IP instead of the last, but didn't succeed. I would really appreciate it if someone could point me in the right direction.
My Idea was to fetch the first IPORHOST by also matching by a comma, but I don't get how to exclude it with the "not captured group", at least on https://grokconstructor.appspot.com it doesn't seem to work:
```
FIRSTIPORHOST (%{IPORHOST}(?:,))
```  

Here are my specs:
Filebeat 5.4 (Running in docker)
OS: Debian 8

Steps to Reproduce:
- Enable cloudflare infront of a site with nginx, cloudflare will start sending the X-Forwarded-Forheader.
- Change the nginx.conf log_format to log the `$http_x_forwarded_for` variable. Here is our config:
```
log_format  proxy '$http_x_forwarded_for - $remote_user [$time_local] "$request" '
                      '$status $body_bytes_sent "$http_referer" '
                      '"$http_user_agent"';
```

tsg:
A common customization to the nginx logs is to add the contents
of the X-Forwarded-For header in front of the remote IPs. This
typically results in a list of remote IPs.

This adds a new field `remote_ip_list` which is an array, and uses
a Painless script to automatically select the first non-private
IP for the `remote_ip` field, which is the field on which GeoIP is
applied.

Fixes #4322.

### DIFF
--- a/filebeat/docs/fields.asciidoc
+++ b/filebeat/docs/fields.asciidoc
@@ -234,6 +234,22 @@ The longitude and latitude.
 
 
 [float]
+=== apache2.access.geoip.region_name
+
+type: keyword
+
+The region name.
+
+
+[float]
+=== apache2.access.geoip.city_name
+
+type: keyword
+
+The city name.
+
+
+[float]
 == error Fields
 
 Fields from the Apache error logs.
@@ -954,11 +970,19 @@ Contains fields for the Nginx access logs.
 
 
 [float]
+=== nginx.access.remote_ip_list
+
+type: array
+
+An array of remote IP addresses. It is a list because it is common to include, besides the client IP address, IP addresses from headers like `X-Forwarded-For`. See also the `remote_ip` field.
+
+
+[float]
 === nginx.access.remote_ip
 
 type: keyword
 
-Client IP address.
+Client IP address. The first public IP address from the `remote_ip_list` array. If no public IP addresses are present, this field contains the first private IP address from the `remote_ip_list` array.
 
 
 [float]
@@ -1139,6 +1163,22 @@ Country ISO code.
 type: geo_point
 
 The longitude and latitude.
+
+
+[float]
+=== nginx.access.geoip.region_name
+
+type: keyword
+
+The region name.
+
+
+[float]
+=== nginx.access.geoip.city_name
+
+type: keyword
+
+The city name.
 
 
 [float]
--- a/filebeat/module/nginx/access/test/test.log-expected.json
+++ b/filebeat/module/nginx/access/test/test.log-expected.json
@@ -1,109 +1,344 @@
 [
-    {
-        "_index": "filebeat-2016.12.27",
-        "_type": "log",
-        "_id": "AVlBCaYsqYg9cc5KQfcT",
-        "_score": null,
-        "_source": {
-            "@timestamp": "2016-12-07T10:05:07.000Z",
-            "offset": 191,
-            "nginx": {
-                "access": {
-                    "referrer": "-",
-                    "response_code": "200",
-                    "remote_ip": "10.0.0.2",
-                    "method": "GET",
-                    "user_name": "-",
-                    "http_version": "1.1",
-                    "body_sent": {
-                        "bytes": "571"
-                    },
-                    "url": "/ocelot",
-                    "user_agent": {
-                        "major": "49",
-                        "minor": "0",
-                        "os": "Mac OS X 10.12",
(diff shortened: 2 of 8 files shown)


filebeat/module/nginx/access/test/test.log-expected.json
```diff
@@ -1,50 +1,25 @@
...
+          "response_code" : "200",
+          "remote_ip" : "85.181.35.98",
+          "geoip" : {
+            "continent_name" : "Europe",
+            "country_iso_code" : "DE",
+            "location" : {
+              "lon" : 9.0,
+              "lat" : 51.0
             }
+<<<<<<< HEAD
```
exekias: please check this, there is a bad merge and tests are failing

filebeat/docs/fields.asciidoc
```diff
@@ -954,11 +970,19 @@ Contains fields for the Nginx access logs.
...
+An array of remote IP addresses. It is a list because it is common to include, besides the client IP address, IP addresses from headers like `X-Forwarded-For`. See also the `remote_ip` field.
+
+
+[float]
 === nginx.access.remote_ip
 
 type: keyword
 
-Client IP address.
+Client IP address. The first non-private IP address from the `remote_ip_list` is preferred. If no public IP addresses are present, the first from the `remote_ip_list` array.
```
dedemorton: So if there are no public IP addresses in `remote_ip_list`, this field contains a private IP address, right? If that's true, we should be a bit more direct here and say something like: 

The first public IP address from the `remote_ip_list`. If no public IP addresses are present, this field contains the first private IP address from the `remote_ip_list` array.

tsg:
Comments addressed.

PR #4420, opened 2017-05-30
Subcommands

exekias:
This change introduces subcommand interface to all beats.

This give us several new features:

 * Isolated subcommands (version, setup, run...)
 * Command line help for all of them
 * Autocompletion and typo hinting
 * Freedom to add beat specific subcommands where needed

Usage examples:

```
$ metricbeat help
$ metricbeat version
$ metricbeat run -h
$ metricbeat run -e -v
```

I've kept root command as an alias for `run`, so old style calls should be still working.

About flags:

Cobra supports persistent (global) and local flags for each command.
I've added config related flags as persistent, common to all
subcommands. While keeping runtime related flags under `run`, see the
full list here:

```
Flags:
  -N, --N                   Disable actual publishing for testing
      --configtest          Test configuration and exit.
      --cpuprofile string   Write cpu profile to file
  -e, --e                   Log to stderr and disable syslog/file output
  -h, --help                help for run
      --httpprof string     Start pprof http server
      --memprofile string   Write memory profile to this file
      --setup               Load the sample Kibana dashboards
  -v, --v                   Log at INFO level
      --version             Print the version and exit

Global Flags:
  -E, --E Flag                      Configuration overwrite (default null)
  -c, --c argList                   Configuration file, relative to path.config (default beat.yml)
  -d, --d string                    Enable certain debug selectors
      --path.config flagOverwrite   Configuration path
      --path.data flagOverwrite     Data path
      --path.home flagOverwrite     Home path
      --path.logs flagOverwrite     Logs path
      --strict.perms                Strict permission checking on config files (default true)
```

In the long term we should aim to deprecate a few of them (setup, version, configtest...) in
favor of subcommands. I'm keeping them for now to maintain backwards compatibilty with 5.X.

In progress:

- [x] Implement missing subcommand: setup, configtest
- [x] Revisit beat generator?


filebeat/cmd/run.go
```diff
@@ -0,0 +1,21 @@
...
+	"os"
+
+	"github.com/spf13/cobra"
+
+	"github.com/elastic/beats/filebeat/beater"
+	"github.com/elastic/beats/libbeat/beat"
+)
+
+// RunCmd runs the beat
+var RunCmd = &cobra.Command{
```
tsg: Any reason this is not also part of libbeat? Is there anything beat specific about `RunCmd`?

exekias: beater is the specific part, but now you ask, I could add that as a parameter to `GenRootCmd`

filebeat/main.go
```diff
@@ -3,12 +3,9 @@ package main
 import (
 	"os"
 
-	"github.com/elastic/beats/filebeat/beater"
-	"github.com/elastic/beats/libbeat/beat"
+	"github.com/elastic/beats/filebeat/cmd"
```
ruflin: This change brings up the question if we should also change the generator to support this?

libbeat/beat/beat.go
```diff
@@ -152,16 +152,16 @@ func init() {
...
 		if err != nil {
 			return err
 		}
 		return b.launch(bt)
 	}())
 }
 
-// newBeat creates a new beat instance
-func newBeat(name, v string) (*Beat, error) {
+// NewBeat creates a new beat instance
```
ruflin: I wonder why lint did not complain :-)

Do we have to make this global?

libbeat/cmd/root.go
```diff
@@ -0,0 +1,63 @@
...
+	rootCmd.Flags().AddFlagSet(runCmd.Flags())
+
+	// Register subcommands common to all beats
+	rootCmd.AddCommand(runCmd)
+	rootCmd.AddCommand(genVersionCmd(name))
+
+	return rootCmd
+}
+
+func init() {
```
ruflin: I would recommend to put the `init` on top, at least that is what we did so far.

libbeat/cmd/version.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/elastic/beats/libbeat/version"
+	"github.com/spf13/cobra"
+)
+
+func genVersionCmd(name string) *cobra.Command {
+	return &cobra.Command{
+		Use:   "version",
+		Short: "Show current version info",
+		Run: func(cmd *cobra.Command, args []string) {
+			beat, err := beat.New(name, "")
```
ruflin: Ok, I see why we need it global :-)

filebeat/main.go
exekias: I promise this was already there: 

image

:smile:

exekias:
jenkins retest this please

exekias:
jenkins retest this please

exekias:
jenkins retest this please

exekias:
jenkins retest this please

exekias:
branch ready from this side

tsg:
I wonder how useful is the `configtest` command. I see we also have the `--configtest` flag. The command has the drawback that it can't take the `-modules` flag, for example.

Here is an example session to show the issue:

```
$ ./filebeat configtest
Exiting: No modules or prospectors enabled and configuration reloading disabled. What files do you want me to watch?
```

```
$ ./filebeat configtest --modules=nginx                                       
Error: unknown flag: --modules
```

```
$ ./filebeat --modules=nginx --configtest                                    
Config OK
```

tsg:
Interesting that both `-modules` and `--modules` work, btw. I kind of like supporting both options. Did you add some compatibility code for that, or is it simply supported by Corba?

exekias:
@tsg:

Yes, with this change we would have duplicated functionality between subcommands and flags, but purely for backwards compatibility, I'm more in favor of subcommands for all flags that modify run behavior to do a "one shoot" thing. I can include `modules` flag in configtest if it's missing

I did some black magic for backwards compatibility, Cobra flags use UNIX style, so it would require `--` for long names: https://github.com/elastic/beats/pull/4420/files#diff-4111a17b440e3fa32c5c0b8255bd512bR15. In the long run I would aim to use Cobra style in all our documentation and scripts (init scripts, dev tooling...), but I'm ok with keeping that conversion code

tsg:
> I did some black magic for backwards compatibility, Cobra flags use UNIX style, so it would require -- for long names: https://github.com/elastic/beats/pull/4420/files#diff-4111a17b440e3fa32c5c0b8255bd512bR15. In the long run I would aim to use Cobra style in all our documentation and scripts (init scripts, dev tooling...), but I'm ok with keeping that conversion code

Ok, I think that's worth a discussion, we could also cut the bandaid with the 6.0 release.

> Yes, with this change we would have duplicated functionality between subcommands and flags, but purely for backwards compatibility, I'm more in favor of subcommands for all flags that modify run behavior to do a "one shoot" thing. I can include modules flag in configtest if it's missing

Generally I agree that things that make the Beat stop are better as commands. But the `configtest` is somehow special as you might want to check that config + CLI flags result in a working config. So either `configtest` accepts all flags that `run` does, or we keep the config test as a flag in `run`.

tsg:
I played with it and it feels great to me, nice work. Dumping some thoughts / wishlist, but for future PRs:

* The `setup` command can export more flags, making it similar in experience with the `import_dashboards` program that we have now. Then we can drop the `import_dashboards` from the packages, reducing the byte size significantly.
* The `setup.template.output_to_file` option, to generate the template into a file, could become it's own command.
* A `modules` command, that just lists the existing modules, would already be useful for filebeat & metricbeat. Enabling/disabling them is probably more work, we should discuss if we can do it for 6.0.

CHANGELOG.asciidoc
```diff
@@ -44,6 +44,8 @@ https://github.com/elastic/beats/compare/v6.0.0-alpha1...master[Check the HEAD d
 
 *Affecting all Beats*
 
+- New cli subcommands interface. {pull}4420[4420]
```
tsg: We'll likely need to deprecate some things, but that can go in a meta ticket.

exekias:
Thanks for the input, great new ideas!, I didn't think about `import_dashboards` and `output_to_file`, definitely something cool to have. I'm ok with doing it in follow up PRs

exekias:
I'll review flags for `configtest` and push a change so we can move forward

dedemorton:
@exekias There will be some doc updates required for this, too (the topic about CLI options and the places where we show commands). Feel free to create a separate issue track this work and assign it to me, if you'd like. TBH, when I look at your description above, the syntax is not entirely clear to me. I'm sure it will make more sense when I can play around with the commands in a build (or have time to look more closely at the code). 

Also, I’m wondering if we’ve taken consistency with other Elastic products (and their command line syntax) into consideration here. Do customers get frustrated when we aren't consistent across products, or am I overthinking this?

exekias:
great @dedemorton I will open a new ticket, also input is more than welcomed if you find better naming for any of the added commands.

I didn't though about consistency, but it's a good point, I'm going to check how the rest of projects handle this. While I think it's not that important (different projects ==> different ways of using them) perhaps we can achieve something more consistent, which is always good

tsg:
@exekias Can you rebase this one, I think it will fix some of the test failures.

PR #4424, opened 2017-05-31
Remove duplicated content about contributing to beats

dedemorton:
Background: After some discussion about whether to keep CONTRIBUTING.md, we decided that we wanted to keep CONTRIBUTING.md (because it's standard for other projects) and link to the documentation for details. So, I've kept the intro text in CONTRIBUTING.md to encourage users to contribute, and I've added a link that sends them to the Beats Developer Guide for details. The advantage is that all developers who read the contributing guide will know right away that there is a developer guide available. 

Changes:
*  Removed most of the content from CONTRIBUTING.md, but kept the introduction that encourages users to contribute.
* Updated contributing.asciidoc to include any changes that were made to CONTRIBUTING.md while both files existed. Also removed the introduction in favor of keeping it in CONTRIBUTING.md.

Note that most of the content from CONTRIBUTING.md was moved to contributing.asciidoc last month. This PR just follows up by removing the duplicated text from CONTRIBUTING.md.


monicasarbu:
Perfect, thanks @dedemorton

Issue #4295, opened 2017-05-12
Make permisson on logfile configureable


florian-asche:
Please see this change: https://github.com/elastic/beats/pull/3387

Can we make the permissions configureable?
(We use other admin user than user the software run under)...

batrako:
in our case, beat  user is only for run filebeat (with no login). We use other shell users to view filebeat logs and enter to system. We need specify another default mask  to filebeat log files (not valid 600 for us).

atoulme:
Possibly as a new setting under logging.files? https://www.elastic.co/guide/en/beats/filebeat/current/configuration-logging.html

florian-asche:
yes

atoulme:
Tye team merged my changes just now. @florian-asche if you agree we can close this issue.

florian-asche:
mhmh. im missing a parameter for user/group. that whould be more secure instead of readable to all users.

atoulme:
I did not understand what you meant, sorry. Would you please clarify?

atoulme:
I think I understand. If you'd like to set a different group, you should probably use the setgid flag on the directory where the logs will be created.

PR #4428, opened 2017-05-31
Add a permissions option to logging.files for all beats

atoulme:
This is a submission for a fix to issue #4295.

### DIFF
--- a/libbeat/docs/loggingconfig.asciidoc
+++ b/libbeat/docs/loggingconfig.asciidoc
@@ -27,6 +27,7 @@ logging.files:
   path: /var/log/mybeat
   name: mybeat.log
   keepfiles: 7
+  permissions: 0644
 ------------------------------------------------------------------------------
 
 TIP: In addition to setting logging options in the config file, you can modify
@@ -123,6 +124,18 @@ The default size limit is 10485760 (10 MB).
 The number of most recent rotated log files to keep on disk. Older files are
 deleted during log rotation. The default value is 7. The `keepfiles` options has to be
 in the range of 2 to 1024 files.
+
+===== files.permissions
+
+The permission mask to apply when rotating log files. The default value is 0600. The
+ `permissions` options must be a valid Unix-style file permissions mask expressed 
+ in octal notation. In Go, numbers in octal notation must start with '0'.
+ 
+Examples:
+* 0644: give read and write access to the file owner, and read access to all others.
+* 0600: give read and write access to the file owner, and no access to all others.
+* 0664: give read and write access to the file owner and members of the group 
+associated with the file, as well as read access to all other users.
 
 ==== Logging Format
 
--- a/libbeat/logp/file_rotator.go
+++ b/libbeat/logp/file_rotator.go
@@ -17,6 +17,7 @@ type FileRotator struct {
 	Name             string
 	RotateEveryBytes *uint64
 	KeepFiles        *int
+	Permissions      *uint32
 
 	current     *os.File
 	currentSize uint64
@@ -42,7 +43,7 @@ func (rotator *FileRotator) CreateDirectory() error {
 
 func (rotator *FileRotator) CheckIfConfigSane() error {
 	if len(rotator.Name) == 0 {
-		return fmt.Errorf("File logging requires a name for the file names")
+		return fmt.Errorf("file logging requires a name for the file names")
 	}
 	if rotator.KeepFiles == nil {
 		rotator.KeepFiles = new(int)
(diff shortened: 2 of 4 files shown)


libbeat/docs/loggingconfig.asciidoc
```diff
@@ -124,6 +125,11 @@ The number of most recent rotated log files to keep on disk. Older files are
 deleted during log rotation. The default value is 7. The `keepfiles` options has to be
 in the range of 2 to 1024 files.
 
+===== files.permissions
+
+The permission mask to apply when rotating log files. The default value is 0600. The 
+`permissions` option must be a valid file permissions mask.
```
dedemorton: Can we add a link to a resource that describes what constitutes a valid file permissions mask? An example is always good, too.

atoulme: Would a link to this do? https://golang.org/pkg/os/#FileMode
Or a link to https://en.wikipedia.org/wiki/File_system_permissions#Numeric_notation ?

atoulme:
Current status:
* the travis build fails to run for 3 configurations. It looks like a problem with the test script unrelated to the changes.
* the appveyor build fails on one test case. I unpacked the system test output and looked for error reports.
I found this error:
`testing.go:610: race detected during execution of test`
This seems unrelated to the changes as well.
* the documentation was reviewed by @dedemorton. I am waiting on her feedback to add a link to the doc as I'm not familiar with the format used here.

Please let me know if I can help push this any further.

libbeat/docs/loggingconfig.asciidoc
dedemorton: Hmm...I wouldn't link to wikipedia even though it's more descriptive. Linking to the golang doc might actually confuse people. 

So maybe instead of a link, you could be a bit more descriptive. Maybe say something like, "The `permissions` options must be a valid Unix-style file permissions mask expressed in octal notation (for example, `0754`). 

Or you could link to the following website, which is super useful, but I hesitate to point to it as a resource because I don't know who maintains it or how long it will be around: https://www.filepermissions.com/

atoulme: You got it. I have amended my commit according to your comments. Any further feedback is very welcome! Thanks!

tsg:
@atoulme Thanks for the PR! There's now a failure on Travis that looks related to this change: https://travis-ci.org/elastic/beats/jobs/238535618#L423

libbeat/docs/loggingconfig.asciidoc
```diff
@@ -27,6 +27,7 @@ logging.files:
   path: /var/log/mybeat
   name: mybeat.log
   keepfiles: 7
+  permissions: 0644
```
tsg: I worry slightly about how YAML interprets this number, is it sure to be octal?

atoulme: From the point of view of go, it's supposed to be a uint32. So it should be ok?
https://golang.org/pkg/os/#FileMode
Note I added validation to make sure the number is less than 7777, but now I realize we could do better.

libbeat/logp/file_rotator.go
```diff
@@ -56,6 +57,10 @@ func (rotator *FileRotator) CheckIfConfigSane() error {
 	if *rotator.KeepFiles < 2 || *rotator.KeepFiles >= RotatorMaxFiles {
 		return fmt.Errorf("The number of files to keep should be between 2 and %d", RotatorMaxFiles-1)
 	}
+
+	if rotator.Permissions != nil && (*rotator.Permissions > 7777) {
```
andrewkroh: I think `7777` should be `os.ModePerm` instead. Only the permission bits should be settable in the mode.

libbeat/docs/loggingconfig.asciidoc
andrewkroh: The users will be required to start the value with a leading 0 in order for it to be interpreted as octal in YAML.

libbeat/logp/file_rotator.go
```diff
@@ -56,6 +57,10 @@ func (rotator *FileRotator) CheckIfConfigSane() error {
...
+		return fmt.Errorf("The permissions mask %d is invalid", *rotator.Permissions)
```
andrewkroh: Error strings should not be capitalized. https://github.com/golang/go/wiki/CodeReviewComments#error-strings

andrewkroh: WDYT about providing a hint here about starting the configured value with a leading 0?

libbeat/docs/loggingconfig.asciidoc
atoulme: @andrewkroh I have added some information in the doc to that effect.

libbeat/logp/file_rotator.go
```diff
@@ -54,7 +55,11 @@ func (rotator *FileRotator) CheckIfConfigSane() error {
 	}
 
 	if *rotator.KeepFiles < 2 || *rotator.KeepFiles >= RotatorMaxFiles {
-		return fmt.Errorf("The number of files to keep should be between 2 and %d", RotatorMaxFiles-1)
+		return fmt.Errorf("the number of files to keep should be between 2 and %d", RotatorMaxFiles-1)
+	}
+
+	if rotator.Permissions != nil && (*rotator.Permissions > 07777) {
```
andrewkroh: Can you replace the `07777` with the constant `os.ModePerm` (which is `0777`). That should be the max value because users should be setting mode type.

tsg:
jenkins, test it

PR #4432, opened 2017-05-31
Fix init script

### LINKED ISSUE #4429: New init script doesn't work on ubuntu
The change introduced in #4340 breaks the init script on older ubuntu systems (12.04 and 14.04 reported).

The check for the `runuser` command doesn't work. This version seems to be correct:

```
if command -v runuser >/dev/null 2>&1; then
    USER_WRAPPER="runuser"
fi
```

But even after fixing the above, `go-daemon` seems to fail when getting `root` as an option:

```
failed to switch to user root: No such file or directory
```

Found by @LeeDr.

tsg:
* Correct the way runuser is called
* Only use `-u $beat_user` if `beat_user` is not `root`. That's because the `-u` command seems to have issues on ubuntu systems.

Fixes #4429.


tsg:
jenkins, package it

PR #4434, opened 2017-06-01
Make IP port indexer constructor public

vjsamuel:
`NewIpPortIndexer()` has scope to be used and extended by other beats. We could make the method public like all the other indexers


tsg:
jenkins, test it

Issue #4426, opened 2017-05-31
Add packetbeat support to TUN devices / RAW DataLink


giacomo83m:
VPN softwares like OpenVPN create on linux "tun devices", a network device that represents the network tunnel and presents directly IP frames (without datalink) to the operative system.
Tun devices are preferred than TAP in all context you don't require / don't want to bridge the client network connection, it's a simple and default configuration.

If you want to analyse traffic of vpn users the suggested configuration is sniff directly from this interface and not from output interface (eg. ethernet) because you want to have metadata (eg. source ip) before NAT.

Libpcap (eg. tcpdump) based solution support this configuration because the data link is detected as:
DLT_RAW; LINKTYPE_RAW=101 # raw IP; the packet begins with an IP header.

This type seems not supported yet by packetbeat (5.4.0) because if you select a sniffer.config.Type "af_packet" the following code:
sniffer.go 
```
func (sniffer *SnifferSetup) Datalink() layers.LinkType {
  if sniffer.config.Type == "pcap" {
    return sniffer.pcapHandle.LinkType()
  }
  return layers.LinkTypeEthernet
}
```

will silently select a LinkTypeEthernet dissector while in this case the datalink is not present; The software will start but isn't able to decode anything.

if you choose instead sniffer.config.Type "pcap" packetbeat will fail on startup because decoder.go doesn't support gopacket/layers/enum.go:LinkTypeRaw / layers.LayerTypeRaw
decoder.go 
```
switch datalink {
  case layers.LinkTypeLinuxSLL:
     d.linkLayerDecoder = &d.sll
     d.linkLayerType = layers.LayerTypeLinuxSLL
  case layers.LinkTypeEthernet:
     d.linkLayerDecoder = &d.eth
     d.linkLayerType = layers.LayerTypeEthernet
  case layers.LinkTypeNull: // loopback on OSx
     d.linkLayerDecoder = &d.lo
     d.linkLayerType = layers.LayerTypeLoopback
  default:
     return nil, fmt.Errorf("Unsupported link type: %s", datalink.String())
}
```

Packetbeat also crashes when you try to enumerate device if this type of device is present:
```
root@ XXXXXX # packetbeat.sh -devices
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x7c6bc6]

goroutine 1 [running]:
panic(0xad3d00, 0xc420010080)
[...]

PR #4445, opened 2017-06-02
Add handling of LayerTypeRaw used by tun devices for pcap (#4426)

zecke:
Be able to capture from a tun device. It has been manually tested by
invoking packetbeat with:

 $ ./packetbeat -c packetbeat.yml  -N -e -I tuncap.pcap

Loglines with "DBG  Layer type: Raw" and "flowid: add ipv4" can be
seen. Support for af_packet is not added here. It is not clear if raw
should be added to the defaultLayerTypes.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -83,6 +83,8 @@ https://github.com/elastic/beats/compare/v6.0.0-alpha1...master[Check the HEAD d
 
 *Packetbeat*
 
+- Add support to capture from a tun device. {issue}4426{4426}
+
 *Winlogbeat*
 
 ==== Deprecated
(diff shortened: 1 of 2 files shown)


zecke:
I should add a test for this one and after my changes to tsg/gopacket are merged someone needs to update the vendor tree.

tsg:
Thanks @zecke, I assume this depends on the gopacket additions. I'm going to review those, and afterwards we'll need to update the vendor folder.

zecke:
@tsg Right. It depends on the new Raw layertype and might have to change depending on your review. At least this patch provides a way to test the part that this is not tested by the unittest

191801737:
@tsg how is it going now

PR #4452, opened 2017-06-02
UDP prospector type with plain harvester

ruflin:
Opening this prospector mainly to show this is WIP and so I don't forget about it :-)

TODO

* [x] Changelog
* [ ] Docs
* [x] Tests
* [x] Code cleanup

And even looks as following

```
{
  "@timestamp": "2017-05-26T11:57:17.68711727+02:00",
  "beat": {
    "hostname": "ruflin",
    "name": "ruflin",
    "version": "6.0.0-alpha2"
  },
  "message": "Hello, World!99",
  "prospector": {
    "harvester": "plain",
    "type": "udp"
  }
}
```

### DIFF
--- a/filebeat/_meta/common.full.p2.yml
+++ b/filebeat/_meta/common.full.p2.yml
@@ -231,6 +231,10 @@ filebeat.prospectors:
   # Redis AUTH password. Empty by default.
   #password: foobared
 
+#- type: udp
+
+  # Maximum size of the message received over UDP
+  #max_message_size: 10240
 #========================= Filebeat global options ============================
 
 # Event count spool threshold - forces network flush if exceeded
--- a/filebeat/prospector/udp/config.go
+++ b/filebeat/prospector/udp/config.go
@@ -1,0 +1,20 @@
+package udp
+
+import (
+	"github.com/elastic/beats/filebeat/harvester"
+)
+
+var defaultConfig = config{
+	ForwarderConfig: harvester.ForwarderConfig{
+		Type: "udp",
+	},
+	MaxMessageSize: 10240,
+	// TODO: What should be default port?
+	Host: "localhost:8080",
+}
+
+type config struct {
+	harvester.ForwarderConfig `config:",inline"`
+	Host                      string `config:"host"`
+	MaxMessageSize            int    `config:"max_message_size"`
+}
--- a/filebeat/prospector/udp/harvester.go
+++ b/filebeat/prospector/udp/harvester.go
@@ -1,0 +1,72 @@
+package udp
+
+import (
+	"net"
+	"time"
+
+	"github.com/elastic/beats/filebeat/harvester"
+	"github.com/elastic/beats/filebeat/util"
+	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/logp"
+)
+
+type Harvester struct {
+	forwarder *harvester.Forwarder
+	done      chan struct{}
+	cfg       *common.Config
+	listener  net.PacketConn
+}
+
+func NewHarvester(forwarder *harvester.Forwarder, cfg *common.Config) *Harvester {
+	return &Harvester{
+		done:      make(chan struct{}),
+		cfg:       cfg,
+		forwarder: forwarder,
+	}
+}
+
+func (h *Harvester) Run() error {
+
+	config := defaultConfig
+	err := h.cfg.Unpack(&config)
+	if err != nil {
+		return err
+	}
+
+	h.listener, err = net.ListenPacket("udp", config.Host)
+	if err != nil {
+		return err
+	}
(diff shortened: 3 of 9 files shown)


filebeat/prospector/udp/harvester.go
```diff
@@ -0,0 +1,52 @@
...
+	listener, err := net.ListenPacket("udp", ":8080")
+
+	if err != nil {
+		logp.Err("error happened: %s", err)
+		return
+	}
+	defer listener.Close()
+
+	//logp.Info("Now listening for logs via %s on %s", ll.config.Protocol, address)
+	buffer := make([]byte, 1000)
```
ruflin: @hartfordfive Any recommendation on the value here?

hartfordfive: @ruflin I realize in theory you could have up to 65507 bytes of data in IPv4 (removing UDP and IP header overhead of 20 and 8 bytes respectively) although I know that you risk having fragmentation of this because of path MTU (which I guess is generally between 1400 and 1500 - not taking into account jumbo frames).  Considering this should only be done locally on the server, i figured this shouldn't happen, so thus why I chose a slightly larger 2048 or 4096 bytes buffer size.  

I know many applications  tend to limit to <= 1024 bytes for a UDP buffer (especially for cross-host communication), although from what I've tested so far, I haven't had messages that were cut off, but I'm aware this could happen, especially if you're going form one host to another even on the same rack.  You could potentially stick with the 1024 byte buffer as a default but allow users to specify the port and the buffer size via the prospector config parameters.

I think it would maybe be nice to have someone else give some feedback on this, especially from the networking stack side to confirm if this is a reasonable assumption or not.  What do you think?

ruflin: Thanks for sharing the thoughts. I set it 1024 for the moment and made it configurable. Time will tell and we can still change the default.

filebeat/prospector/udp/harvester.go
```diff
@@ -0,0 +1,77 @@
...
+		length, _, err := h.listener.ReadFrom(buffer)
+		if err != nil {
+			logp.Err("Error reading from buffer: %v", err.Error())
+			continue
+		}
+		data := util.NewData()
+		event := common.MapStr{
+			"message": string(buffer[:length]),
+			"prospector": common.MapStr{
+				"harvester": "plain",
```
ruflin: @tsg I'm asking myself if this should be code? Other options here could be `json`.

tsg: Do we need a "harverster" key? Maybe we add it when we have more of them.

I'm thinking that, for example, JSON can be accomplished with the processor, so I'm not sure harvester support for it is needed.

filebeat/_meta/common.full.p2.yml
```diff
@@ -231,6 +231,10 @@ filebeat.prospectors:
   # Redis AUTH password. Empty by default.
   #password: foobared
 
+#- type: udp
+
+  # Maximum size of each message sent over udp
+  #max_message_size: 1024
```
tsg: Two things here:

* I think it would be more clear if it would say "Maximum size of the message received over UDP".
* 1024 seems low, UDP can go up to 65K. How about 10K, which should be large enough for most usecases but still small enough to not be noticeable in the mem usage?

filebeat/prospector/udp/harvester.go
ruflin: We don't need it at the moment an can still add it later.

filebeat/_meta/common.full.p2.yml
ruflin: I set it to `10240` now. 1024 was based on the discussion here.

filebeat/prospector/udp/config.go
```diff
@@ -0,0 +1,20 @@
...
+
+import (
+	"github.com/elastic/beats/filebeat/harvester"
+)
+
+var defaultConfig = config{
+	ForwarderConfig: harvester.ForwarderConfig{
+		Type: "udp",
+	},
+	MaxMessageSize: 1024,
```
exekias: Should this be 10240?

ruflin: changed, thanks for spotting.

exekias:
WFG

dedemorton:
Can I remove the needs_docs tag? Is #4940 sufficient, or is additional doc required?

ruflin:
@dedemorton I was hoping for more docs that also shows how to configure the host for example. But I don't know how this fits into our current doc model of prospectors. Perhaps worth to discuss quickly over zoom?

dedemorton:
I'm going to remove the needs_docs label here in favor of tracking the work in https://github.com/elastic/beats/issues/5632

PR #4453, opened 2017-06-04
Separate travis jobs in stages

exekias:
This change runs `check` job first and proceeds to launch all tests
in parallel after success. No further testing is done if check fails.

See https://docs.travis-ci.com/user/build-stages

This change should save same cycles in travis so we get a cleaner queue


exekias:
Results here: https://travis-ci.org/elastic/beats/builds/239341583

ruflin:
Nice, wasn't aware Travis added this feature.

Issue #4461, opened 2017-06-06
Metricbeat - Vsphere Module - Host metricset gives negative used memory


amandahla:
Even with same unit of measure, the field memory.used.bytes is negative.

andrewkroh:
Fixed by #4462.

PR #4462, opened 2017-06-06
Used memory field corrected (#4461)

amandahla:
Change for correct the used memory field.

### DIFF
--- a/metricbeat/module/vsphere/host/data_test.go
+++ b/metricbeat/module/vsphere/host/data_test.go
@@ -1,0 +1,54 @@
+package host
+
+import (
+	"math"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/vmware/govmomi/vim25/mo"
+	"github.com/vmware/govmomi/vim25/types"
+)
+
+func TestEventMapping(t *testing.T) {
+
+	var HostSystemTest = mo.HostSystem{
+		Summary: types.HostListSummary{
+			Host: &types.ManagedObjectReference{Type: "HostSystem", Value: "ha-host"},
+			Hardware: &types.HostHardwareSummary{
+				MemorySize:  2251799812636672,
+				CpuMhz:      2294,
+				NumCpuCores: 2,
+			},
+			Config: types.HostConfigSummary{
+				Name: "localhost.localdomain",
+			},
+			QuickStats: types.HostListSummaryQuickStats{
+				OverallCpuUsage:    67,
+				OverallMemoryUsage: math.MaxInt32,
+			},
+		},
+	}
+
+	event := eventMapping(HostSystemTest, "test")
+
+	datacenter, _ := event.GetValue("datacenter")
+	assert.EqualValues(t, "test", datacenter)
+
+	cpuUsed, _ := event.GetValue("cpu.used.mhz")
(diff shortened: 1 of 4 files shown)


tsg:
jenkins, test it

andrewkroh:
Hi @amandahla, is there a test case you can add for this? It would be nice to have something in place that demonstrates this fixes the issue and prevents future regression.

And BTW if you put "Closes #4461" in the commit message then the issue will [automatically be closed](https://help.github.com/articles/closing-issues-via-commit-messages/) when this merges.

amandahla:
@andrewkroh hs.Summary.QuickStats.OverallMemoryUsage is a int32 value for memory usage in MB.
When I multiply twice to get bytes numbers without the conversion, it exceeds the int32 maximum and gives a wrong number. We just get this when we saw weird values on dashboard. So I saw that I must convert to int64 to fit the right value.
I don't know how can I test when it's necessary a bigger variable type. Any suggestions?

Thanks for the tip about close issues!

andrewkroh:
> Any suggestions?

I recommend splitting up the logic from `Fetch` into smaller functions that can be easily unit tested (without requiring the simulator).

amandahla:
@andrewkroh I'm not sure if I get the idea. Is something like this?

andrewkroh:
Yeah, sort of. I guess I was thinking more like this (or some other break down of the existing logic to make it unit testable without the use of the data simulator).

```
    ...
    for _, hs := range hst {
        events = append(events, buildEvent(hs))
    }
    ...
}
func buildEvent(hs mo.HostSystem) common.MapStr { ... }
```

```
func TestBuildEvent(t *testing.T) {
   var hs mo.HostSystem
   hs.Summary.Hardware.MemorySize = math.MaxInt32
   // set other values

   event := buildEvent(hs)

   // assert correct output w/o overflow
}
```

amandahla:
@andrewkroh Another try :)

metricbeat/module/vsphere/host/data_host_system.go
```diff
@@ -0,0 +1,129 @@
+package host
```
andrewkroh: It looks like this file is only used for testing. If so it should be named with the `_test.go` suffix so that it is only compiled and used in tests. But I would probably just move the `HostSystemTest` definition into the `data_test.go`.

andrewkroh: It looks like you could have got away with a much smaller set of stub data for your testing purposes because `eventMapping` only needs a few of the fields.

metricbeat/module/vsphere/host/data_test.go
```diff
@@ -0,0 +1,38 @@
...
+func TestEventMapping(t *testing.T) {
+
+	event := eventMapping(HostSystemTest, "test")
+
+	assert.EqualValues(t, "test", event["datacenter"])
+
+	cpu := event["cpu"].(common.MapStr)
+
+	cpuUsed := cpu["used"].(common.MapStr)
+	assert.EqualValues(t, 67, cpuUsed["mhz"])
```
andrewkroh: FYI It should be possible to write this as

```
value, _ := event.Get("cpu.used.mhz")
assert.EqualValues(t, 67, value)
```

Personally, I find using the dotted field notational easier to work with.

PR #4314, opened 2017-05-15
Fix race condition for harvester Start / Stop in registry

ruflin:
It was possible that with reloading enabled that during the shutdown of filebeat, a new harvester was started. This is now prevent by having a lock on the starting of the harvester so no new harvesters can be started when shutdown started.

This is not backported to 5.x because it can only happen on shutdown and does not have any side affects.

Closes #3779

PR #4465, opened 2017-06-06
Move libbeat filebeat dependency to libbeat

ruflin:
Libbeat had a dependency on filebeat because of file rotation. This PR removes the dependency by moving the file rotation code to


ruflin:
Failure is not related to this change. It seems like the old race condition hunting me is back: https://github.com/elastic/beats/pull/4314

PR #4467, opened 2017-06-06
Automatic merge from 5.4 to the 5.x branch

tsg:
Another try for the 5.4 -> 5.x merge.


tsg:
There's a failure starting the LS container, which according to the Logstash team should be fixed with the next snapshot build of 5.x.

PR #4475, opened 2017-06-08
fix template for output filename and path

jalvz:
This fixes the generation of the yml file for system tests. After generating an "example" beat, the resulting `tests/system/config/example.yml.j2` would look like:

```
   # Path to the directory where to save the generated files. The option is mandatory.
    path: {{ output_file_path|default(beat.working_dir + "/output") }}


    # Name of the generated files. The default is `example` and it generates
    # files: `example`, `example.1`, `example.2`, etc.
    filename: {{ output_file_filename|default("example") }}
```


tsg:
Thanks for the fix, @jalvz!

PR #4442, opened 2017-06-02
Fixed socket scanning for tcp6

tomaszrichert:
Fixed issue with /proc/net/tcp6 not being scanned due to simple bug

PR #4477, opened 2017-06-08
Add tests for latest ipv6 related fixes

exekias:
Add tests for #4442


andrewkroh:
Can you do something with this log statement. It's not true when the error occurs on the tcp6 file. https://github.com/exekias/beats/blob/33a1e5d8138ca8640b4ddbe60ca13d34fde7e302/packetbeat/procs/procs.go#L345

PR #4479, opened 2017-06-08
Load Filebeat modules pipelines on connect

### LINKED ISSUE #4444: Filebeat Module: Start problem with Elasticsearch on the same machine
- Version: Filebeat 5.4.0+
- Operating System: Ubuntu 16.04
- Steps to Reproduce: 
  * Have Elasticsearch and Filebeat on the same system
  * Enable Filebeat modules (I tested with system and nginx)
  * Elasticsearch is not yet up and running before Filebeat tries to add its ingest pipeline and dashboard and then stops because of that.

Probably https://github.com/elastic/beats/blob/master/filebeat/beater/filebeat.go#L116 and https://github.com/elastic/beats/blob/master/filebeat/beater/filebeat.go#L102 are relevant here.

Either Filebeat can retry that initial setup step a few times or Filebeat (with modules enabled) should only start up after Elasticsearch is up and running.

tsg:
This is done by registering another callback to the elasticsearch output. We
used to support only one callback, this extends to support multiple callbacks
and also protects the list with a mutex.

There is plan to refactor the package global, but after the publisher refactoring.

Fixes #4444.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -48,6 +48,7 @@ https://github.com/elastic/beats/compare/v6.0.0-alpha1...master[Check the HEAD d
 
 - Add experimental Redis module. {pull}4441[4441]
 - Nginx module: use the first not-private IP address as the remote_ip. {pull}4417[4417]
+- Load Ingest Node pipelines when the Elasticsearch connection is established, instead of only once at startup. {pull}4479[4479]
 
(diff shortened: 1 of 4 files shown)


ruflin:
@tsg Can the branch be deleted?

tsg:
@ruflin Ooops, must have pushed to the old origin, thanks for noticing.

PR #4482, opened 2017-06-09
Add kernel.audit_rules config option to set audit rules

andrewkroh:
This PR adds the ability to have Metricbeat install audit rules to the kernel.

Metricbeat supports adding both file watch rules (`-w`) and syscall rules (`-a` or `-A`). The [format](https://linux.die.net/man/8/auditctl) for specifying rules is the same as with auditd. For example, both of these are supported.

```
kernel.audit_rules: |
  -w /etc/passwd -p wa -k identity
  -a always,exit -F arch=b64 -S open -F exit=-EACCES -F key=access
```

This PR depends on https://github.com/elastic/go-libaudit/pull/7.


NOTICE
```diff
@@ -270,12 +270,188 @@ SOFTWARE.
...
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+
+
+--------------------------------------------------------------------
+github.com/stretchr/testify
```
tsg: I'm sure we had at least this one already, I wonder if duplicates are somehow created here in the NOTICE file.

andrewkroh: I accidentally added an uncommitted NOTICE file when updating the go-libaudit vendor copy. Then the `vendor/github.com/elastic/go-libaudit/NOTICE` file was include in the beat NOTICE file

metricbeat/metricbeat.full.yml
```diff
@@ -126,6 +126,11 @@ metricbeat.modules:
   kernel.rate_limit: 0
   kernel.include_raw_message: false
   kernel.include_warnings: false
+  kernel.audit_rules: |
+    # Define audit rules here.
+    # Create file watches (-a) or syscall audits (-a or -A). For example:
```
exekias: Sorry I'm not familiar with audit rules, this comment refers to '-w /etc/passwd` or the next line?

andrewkroh: It was suppose to refer to them both, but has a typo. It should say "Create file watches (-w)...". Will fix.

PR #4484, opened 2017-06-09
Add libbeat/common/atomic package

urso:
The libbeat/common/atomic package wraps sync/atomic package, providing types
atomic.T with atomic accessors only.

### DIFF
--- a/libbeat/common/atomic/atomic.go
+++ b/libbeat/common/atomic/atomic.go
@@ -1,0 +1,77 @@
+// Package atomic provides common primitive types with atomic accessors.
+package atomic
+
+import a "sync/atomic"
+
+// Bool provides an atomic boolean type.
+type Bool struct{ u Uint32 }
+
+// Int32 provides an atomic int32 type.
+type Int32 struct{ value int32 }
+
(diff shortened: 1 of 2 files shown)


libbeat/common/atomic/atomic.go
```diff
@@ -0,0 +1,77 @@
...
+type Int64 struct{ value int64 }
+
+// Uint32 provides an atomic uint32 type.
+type Uint32 struct{ value uint32 }
+
+// Uint64 provides an atomic uint64 type.
+type Uint64 struct{ value uint64 }
+
+func MakeBool(v bool) Bool             { return Bool{MakeUint32(encBool(v))} }
+func NewBool(v bool) *Bool             { return &Bool{MakeUint32(encBool(v))} }
```
andrewkroh: It will be convenient having an atomic bool.

tsg:
@urso can you rebase to master to fix the tests, please?

urso:
Added some unit tests. As common/atomic is basically a wrapper for sync/atomic, these tests are very minimal, checking operations are actually consistent (so we hopefully don't introduce subtle/stupid errors in future).

PR #4486, opened 2017-06-09
Add file integrity metricset to Metricbeat audit module [WIP]

andrewkroh:
This metricset monitors files or directories for changes. When a file is changed it will calculate a MD5, SHA1, and SHA256 hash.

```
metricbeat.modules:
- module: audit
  metricsets: [file]
  file.paths:
    binaries:
    - /usr/bin
    - /usr/sbin
```

Tested on darwin, linux, and windows.

### DIFF
--- a/metricbeat/module/audit/file/event.go
+++ b/metricbeat/module/audit/file/event.go
@@ -1,0 +1,120 @@
+package file
+
+import (
+	"crypto/md5"
+	"crypto/sha1"
+	"crypto/sha256"
+	"encoding/hex"
+	"io"
+	"os"
+	"time"
+
+	"github.com/pkg/errors"
+
+	"github.com/elastic/beats/libbeat/common"
+)
+
+type Action uint8
+
+func (a Action) String() string {
+	if name, found := actionNames[a]; found {
+		return name
+	}
+	return "unknown"
+}
+
+const (
+	Unknown = iota << 1
+	AttributesModified
+	Created
+	Deleted
+	Updated
+	MovedTo
+	CollisionWithin
+	Unmounted
+	RootChanged
+)
+
+var actionNames = map[Action]string{
+	AttributesModified: "attributes_modified",
+	Created:            "created",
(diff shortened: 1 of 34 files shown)


metricbeat/module/audit/file/event.go
```diff
@@ -0,0 +1,120 @@
...
+func hashFile(name string) (md5sum, sha1sum, sha256sum string, err error) {
+	f, err := os.Open(name)
+	if err != nil {
+		return "", "", "", errors.Wrap(err, "failed to open file for hashing")
+	}
+	defer f.Close()
+
+	m5 := md5.New()
+	s1 := sha1.New()
+	s256 := sha256.New()
```
tsg: Are all three, at the same time, useful/needed?

metricbeat/module/audit/file/eventreader_other.go
```diff
@@ -0,0 +1,82 @@
+// +build linux freebsd openbsd netbsd windows
+
+package file
+
+import (
+	"time"
+
+	"github.com/fsnotify/fsnotify"
```
tsg: I like that we get experience with this 👍

tsg:
Cool, just to make sure, this is not based on the Audit framework, but uses fsnotify to know when a file changes, right?

metricbeat/module/audit/file/event.go
andrewkroh: Maybe it depends on how you are using this info. The threat intel that I see is usually inconsistent in the hashing algorithms used. So you might get a hashes reported from one source that always provides SHA1 and another source that always provides SHA256. So for this hash matching use case it is useful to have all three.

andrewkroh:
That's correct, this is based on inotify (linux), fsevents (macos), and ReadDirectoryChangesW (windows).

I plan to re-open this PR against auditbeat after it's merged, but I'll leave this open until then in case anyone has comments.

andrewkroh:
Closed in favor of #4562

Issue #562, opened 2015-12-18
Packetbeat:  Feature request - add the pid to the exported fields


hamelg:
In the environmental packetbeat fields, we have the proc field.
proc >> the name of the process that served the transaction.

It would be very useful to add the pid.
We collect http protocol and we have hundreds apache/php processes on each server. Each process does outbound connections to mysql databases and others remote HTTP webservices. Having the pid, would allow us to identify and follow connections by apache process.
should it be possible to add this field ?

joedissmeyer:
This is a feature that we would very much like as well!

andrewkroh:
`process.pid`, `client.process.pid`, and `server.process.pid` were added as part of the ECS migration effort in https://github.com/elastic/beats/pull/9303.

PR #4487, opened 2017-06-09
Add packetbeat pid

tsg:
We had the process names, and this adds also the PID, closing #562.

Diff is quite long due to existing code repetition between modules. This also gets rid of some unnecessary type casts. 

Todo:

* [ ] figure out a way to test this
* [ ] changelog


andrewkroh:
Packetbeat now has `client.process.pid` and `server.process.pid` as of https://github.com/elastic/beats/pull/9303.

PR #4492, opened 2017-06-12
Update libbeat publisher pipeline

urso:
This PR marks the beginning of the libbeat event publisher pipeline refactoring.

- central to the publisher pipeline is the broker:
  - broker implementation can be configured when constructing the pipeline
  - common broker implementation tests in `brokertest` package
- broker features:
  - Fully in control of all published events. In comparison to old publisher pipeline with many batches in flight, the broker now configures/controls the total number of events stored in the publisher pipeline. Only after ACKs from outputs, will new space become available.
  - broker returns ACKS in correct order to publisher
  - broker batches up multiple ACKs
  - producer can only send one event at a time to the broker (push)
  - consumer can only receive batches of events from broker (pull)
  - producer can cancel(remove) active events not yet pulled by a consumer
- broker/output related interfaces defined in `publisher` package
- pipeline/client interfaces for use by beats currently defined in `publisher/beat` package
  - event structure has been changed to be more compatible with Logstash (See beat.Event): Beats can send metadata to libbeat outputs (e.g. pipeline) and logstash by using the `Event.Meta` field. Event fields will be stored on `Event.Fields`. Event fields are normalized (for use with processors) and serialized using.
- The old publishers publish API is moved to libbeat/publisher/bc/publisher for now:
  - move to new sub-package to fight of circular imports
  - package implements old pipeline API on top of new pipeline
- Filters/Processors are still executed before pushing events to the new pipeline
- New API:
  - beats client requirements are configured via `beat.ClientConfig`:
    - register async ACK callbacks (currently callbacks will not be triggered after `Client.Close`)
    - configurable sending guarantees (must match ACK support)
    - "wait on close", for beats clients to wait for pending events to be ACKed (only if ACK is configured)
    - pipeline also supports "wait on close", waiting for pending events (independent of ACK configurations). Can be used by any beat, to wait on shutdown for published events to be actually send

## Event Structure:
The event structure has been changed a little:

```
type Event struct {
	Timestamp time.Time
	Meta      common.MapStr
	Fields    common.MapStr
}
```

- We always require the timestamps.
- Meta contains additional meta data (hints?) a beat can forward to the outputs. For example `pipeline` or `index` settings for the Elasticsearch output.
- If output is not Elasticsearch, a `@metadata` field will always be written to the json document. This way Logstash can take advantage of all `@metadata`, even if the event has been send via kafka or redis.

## Output changes

The new output plugin factory is defined as:

```
type Factory func(beat common.BeatInfo, cfg *common.Config) (Group, error)
```

The package libbeat/output/mode is being removed + all it's functionality is moved into a single implementation in the publisher pipeline supporting sync/async clients with failover and load-balancing. In the future dynamic output discovery might be added as well. This change requires output.Group to return some common settings for an active output, to configure the pipeline:

```
// Group configures and combines multiple clients into load-balanced group of clients
// being managed by the publisher pipeline.
type Group struct {
	Clients   []Client
	BatchSize int
	Retry     int
}
```

Moving functionality from the outputs to the publisher pipeline restricts beats from having one output type configured only.

All client instances configured will participate in load-balancing being driven by the publisher pipeline. This removes some intermediate workers used for forwarding batches. Future changes to groups include:

- load-balance to N of M client nodes only + use left-over nodes for fail-over
- overlapping fail-over: send event to other client (very short timeout) while waiting for ACK from first client -> ACK/cancel batch once one output did ACK the event (might lead to duplicates, but improves thoughput if one output is stalled).
- provide functionality in groups to add/remove clients
- dynamic output update/reloading by instantiating a new group and dynamically switching over to new output group.
- move encoding to earlier phase in the pipeline, to limit events/memory usage and batch sizes by number of events and/or byte size

outputs always operate in batches by implement only the `Publish` method:

```
	// Publish sends events to the clients sink. A client must synchronously or
	// asynchronously ACK the given batch, once all events have been processed.
	// Using Retry/Cancelled a client can return a batch of unprocessed events to
	// the publisher pipeline. The publisher pipeline (if configured by the output
	// factory) will take care of retrying/dropping events.
	Publish(publisher.Batch) error
```

With:

```
// Batch is used to pass a batch of events to the outputs and asynchronously listening
// for signals from these outpts. After a batch is processed (completed or
// errors), one of the signal methods must be called.
type Batch interface {
	Events() []Event

	// signals
	ACK()
	Drop()
	Retry()
	RetryEvents(events []Event)
	Cancelled()
	CancelledEvents(events []Event)
}
```

The batch interface combines `events + signaling` into one common interface.
The main difference between sync/async clients is, when `batch.ACK` is called. Batches/Events can be processed out of order. The publisher pipelining doing the batching and load-balancing guarantees ACKs being returned to the beat in order + implements upper bound. Once publisher pipeline is 'full', it will block, waiting for ACKs from outputs.

The logic for dropping events on retry and guaranteed sending is moved to the publisher pipeline as well. Outputs are concerned with publishing and signaling ACK or Retry only.

### Upcoming
- improve unit tests in new pipeline
- integrate pipeline stress test tool with tests
- update beats configuration (remove `queue_size` and `bulk_queue_size` in favor of configurable broker)
- update processors to operate on `beat.Event` instead of `common.MapStr`
- make `@metadata` accessible from processors?
- update all beats to new publisher pipeline:
  1. make connect-to-new-pipeline available from `bc/publisher` package
  2. update beats one by one to new pipeline
  3. remove `bc/publisher` package
- move the outputs codec as a setting to `outputs.Group`, so codecs can be applied earlier in pipeline (=> limit events per batch/queue by memory usage)


libbeat/outputs/codec/json/event.go
```diff
@@ -0,0 +1,35 @@
+package json
+
+import (
+	"time"
+
+	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/publisher/beat"
+)
+
+// Event describes the event strucutre for events
```
andrewkroh: `s/strucutre/structure/`

filebeat/tests/system/test_shutdown.py
```diff
@@ -63,9 +63,10 @@ def test_shutdown_wait_ok(self):
         assert len(registry) == 1
         assert registry[0]["offset"] == output["offset"]
 
+    @unittest.skip("Skipping unreliable test")
```
urso: Have had to skip the test, as it has been very unreliable and I didn't manage to reproduce it yet for investigation. Plus, this feature as is will be replaced with wait shutdown support directly provided by the new publisher pipeline.

libbeat/publisher/beat/bc.go
```diff
@@ -0,0 +1 @@
+package beat
```
tsg: Empty file, is it needed?

urso:
For Reviewer notes:
- the libbeat/publisher package has been reviewed in the past. Only the logging has been adapted to use logp
- new publisher pipeline is not exposed to beats yet. `libbeat/publisher/bc/publisher` implements a compatibility layer using the old API on top of new pipeline
- have a look for recent ES/LS changes potentially missing. I have had a many merge conflicts and some auto-merge potentially breaking things (have had to re-implement multiple onConnect callbacks in ES outputs)
- `libbeat/outputs/codec` has been changed
- new event type `libbeat/publisher/event.go`
- updated `/libbeat/common/fmtstr` package to support new event type

libbeat/publisher/beat/bc.go
urso: Some left-over. Will remove it.

tsg:
@urso I hit an error when using an output config section like this:

```
output.elasticsearch:
  enabled: false
  hosts: ["localhost:9200"]
output.logstash:
  hosts: ["localhost:5044"]
```

The error is: 

```
metricbeat2017/06/21 22:35:31.167563 beat.go:632: CRIT Exiting: error unpacking config data: more then one namespace configured accessing 'output' (source:'metricbeat.dev.yml')
```

Expected?

urso:
@tsg this is to be excepted. From notes: 

> Moving functionality from the outputs to the publisher pipeline restricts beats from having one output type configured only.

With this PR one can have only one output configured. The error message is generated by beats config using `Output common.ConfigNamespace`, which allows for max one setting.

libbeat/logp/logger.go
```diff
@@ -0,0 +1,31 @@
+package logp
```
ruflin: why did you touch the logger in this PR?

libbeat/monitoring/report/elasticsearch/elasticsearch.go
```diff
@@ -52,66 +48,93 @@ var defaultParams = map[string]string{
...
 	}
+	for _, host := range hosts {
+		client, err := makeClient(host, params, proxyURL, tlsConfig, &config)
+		if err != nil {
+			return nil, err
+		}
+		out.Clients = append(out.Clients, client)
+	}
 
-	// backoff parameters
```
ruflin: I assume I will find all the defaults below somewhere else :-)

libbeat/monitoring/report/report.go
```diff
@@ -91,25 +91,14 @@ func getReporterConfig(
...
-	var found string
-	for name := range outputs {
-		if reportFactories[name] == nil {
-			continue
-		}
-
-		if found != "" {
-			err := fmt.Errorf("multiple potential monitoring reporters found (for example %v and %v)", found, name)
-			return "", nil, err
+	if outputs.IsSet() {
```
ruflin: Nice. I see the possibility coming of not having an output initially ;-)

libbeat/outputs/backoff.go
```diff
@@ -0,0 +1,51 @@
...
+
+type backoffClient struct {
+	client NetworkClient
+
+	done    chan struct{}
+	backoff *common.Backoff
+}
+
+// WithBackoff wraps a NetworkClient, adding exponential backoff support to a network client if connection/publishing failed.
+func WithBackoff(client NetworkClient, init, max time.Duration) NetworkClient {
```
ruflin: Reminds me of the filebeat readers implementation ;-)

libbeat/outputs/elasticsearch/api.go
```diff
@@ -42,15 +41,6 @@ type CountResults struct {
 	Shards json.RawMessage `json:"_shards"`
 }
 
-func (r QueryResult) String() string {
```
ruflin: I wonder where this was used and what it's going to break.

libbeat/outputs/elasticsearch/api_test.go
```diff
@@ -134,3 +136,12 @@ func TestReadSearchResult_invalid(t *testing.T) {
 func newTestClient(url string) *Client {
 	return newTestClientAuth(url, "", "")
 }
+
+func (r QueryResult) String() string {
```
ruflin: ah, was only used for testing?

libbeat/outputs/elasticsearch/client.go
```diff
@@ -53,6 +55,14 @@ type ClientSettings struct {
 	Pipeline           *outil.Selector
 	Timeout            time.Duration
 	CompressionLevel   int
+	Stats              *ClientStats
+}
+
+type ClientStats struct {
```
ruflin: Should this be generic to all clients?

libbeat/outputs/elasticsearch/client.go
```diff
@@ -302,36 +314,35 @@ func bulkEncodePublishRequest(
...
+	pipeline, err := getPipeline(event, pipelineSel)
 	if err != nil {
 		logp.Err("Failed to select pipeline: %v", err)
 	}
 
 	if pipeline == "" {
 		type bulkMetaIndex struct {
-			Index   string `json:"_index"`
-			DocType string `json:"_type"`
+			Index   string `json:"_index" struct:"_index"`
```
ruflin: why is the `_struct` needed? Does this come from go-structform?

libbeat/outputs/elasticsearch/client.go
```diff
@@ -374,29 +385,23 @@ func getPipeline(data outputs.Data, pipelineSel *outil.Selector) (string, error)
...
-				if dynamicIndexValue, ok := dynamicIndex.(string); ok {
-					return fmt.Sprintf("%s-%d.%02d.%02d",
-						dynamicIndexValue, ts.Year(), ts.Month(), ts.Day())
-				}
+func getIndex(event *beat.Event, index outil.Selector) string {
+
+	if event.Meta != nil {
+		if str, exists := event.Meta["index"]; exists {
+			idx, ok := str.(string)
+			if ok {
```
ruflin: is setting your own index still working with this change?

libbeat/outputs/elasticsearch/config.go
```diff
@@ -38,6 +45,10 @@ var (
 		CompressionLevel: 0,
 		TLS:              nil,
 		LoadBalance:      true,
+		Backoff: Backoff{
```
ruflin: here we go with the missing defaults :-)

libbeat/outputs/elasticsearch/elasticsearch.go
```diff
@@ -91,7 +183,7 @@ func NewConnectedClient(cfg *common.Config) (*Client, error) {
 	}
 
 	for _, client := range clients {
-		err = client.Connect(client.timeout)
+		err = client.Connect()
```
ruflin: Still trying to figure out where all the timeouts were moved to ...

libbeat/outputs/kafka/config.go
```diff
@@ -45,10 +46,10 @@ type metaRetryConfig struct {
...
 var (
 	defaultConfig = kafkaConfig{
-		Hosts:   nil,
-		TLS:     nil,
-		Timeout: 30 * time.Second,
-		Worker:  1,
+		Hosts:       nil,
+		TLS:         nil,
+		Timeout:     30 * time.Second,
+		BulkMaxSize: 2048,
```
ruflin: The worker feature moved to an other place or does not directly exist anymore?

libbeat/outputs/kafka/kafka.go
```diff
@@ -99,27 +76,30 @@ var (
...
-// New instantiates a new kafka output instance.
-func New(_ common.BeatInfo, cfg *common.Config) (outputs.Outputer, error) {
-	output := &kafka{}
-	err := output.init(cfg)
-	if err != nil {
-		return nil, err
-	}
-	return output, nil
+func init() {
+	sarama.Logger = kafkaLogger{}
```
ruflin: is that why you touched the Logger part?

libbeat/outputs/logstash/async.go
```diff
@@ -166,32 +179,37 @@ func (r *msgRef) callback(seq uint32, err error) {
...
 	r.dec()
 }
 
 func (r *msgRef) fail(n uint32, err error) {
 	ackedEvents.Add(int64(n))
 	outputs.AckedEvents.Add(int64(n))
-	r.err = err
-	r.batch = r.batch[n:]
+
+	if r.err == nil {
```
ruflin: looks suspicious

libbeat/outputs/logstash/ca_invalid_test.pem
```diff
@@ -1,31 +1,31 @@
 -----BEGIN CERTIFICATE-----
-MIIFXTCCA0WgAwIBAgIQCYZiCOjXqDaqWrOaCNv+JDANBgkqhkiG9w0BAQ0FADAv
+MIIFSzCCAzOgAwIBAgIQf2sHjgMXj7KsrArLn7j8wTANBgkqhkiG9w0BAQ0FADAv
```
ruflin: reason to update certificates?

libbeat/outputs/logstash/client_test.go
```diff
@@ -10,9 +10,10 @@ import (
 	"github.com/elastic/go-lumber/server/v2"
 
 	"github.com/elastic/beats/libbeat/common"
-	"github.com/elastic/beats/libbeat/outputs"
+	"github.com/elastic/beats/libbeat/outputs/outest"
```
ruflin: outest? :-)

libbeat/outputs/logstash/event.go
```diff
@@ -0,0 +1,30 @@
+package logstash
+
+/*
+// Event describes the event strucutre for events
```
ruflin: Commented out?

libbeat/outputs/logstash/logstash.go
```diff
@@ -89,113 +66,27 @@ func (lj *logstash) init(cfg *common.Config) error {
...
-) error {
-	return lj.mode.PublishEvent(signaler, opts, data)
-}
+		client = outputs.WithBackoff(client, config.Backoff.Init, config.Backoff.Max)
+		clients[i] = client
+	}
 
-// BulkPublish implements the BulkOutputer interface pushing a bulk of events
-// via lumberjack.
-func (lj *logstash) BulkPublish(
```
ruflin: Did all this stuff get replaced by new publisher?

libbeat/publisher/bc/publisher/client.go
```diff
@@ -240,3 +274,22 @@ func MakeContext(opts []ClientOption) ([]common.MapStr, Context) {
 	}
 	return meta, ctx
 }
+
+func makeEvent(fields common.MapStr, meta common.MapStr) beat.Event {
+	var ts time.Time
+	switch value := fields["@timestamp"].(type) {
+	case time.Time:
```
ruflin: seems we have this code in different places

libbeat/publisher/broker/brokertest/log.go
```diff
@@ -0,0 +1,136 @@
...
+type testLogWriter struct {
+	t *testing.T
+}
+
+func (w *testLogWriter) Write(p []byte) (int, error) {
+	w.t.Log(string(p))
+	return len(p), nil
+}
+
+func withLogOutput(fn func(*testing.T)) func(*testing.T) {
```
ruflin: own logger?

libbeat/publisher/includes/includes.go
```diff
@@ -0,0 +1,15 @@
+package includes
+
+import (
+	// load supported output plugins
```
ruflin: we could generate this in the future ... but as we are not going to add outputs ...

libbeat/tests/system/beat/beat.py
```diff
@@ -239,7 +239,9 @@ def read_output_json(self, output_file=None):
                     # hit EOF
                     break
 
-                jsons.append(json.loads(line))
+                event = json.loads(line)
+                del event['@metadata']
```
ruflin: id @metadata in the output event?

filebeat/tests/system/test_shutdown.py
urso: It's the only test I have had to skip. But will create meta-ticket for further pipeline work.

libbeat/logp/logger.go
urso: some parts of the pipeline allow for a logger to be passed by interface/type. This allows the more 'expensive' tests to capture the log output in the test context -> correctly group log messages with test output.

urso: The old internal `Logger` type is renamed to `logger`. We now export a `*Logger` type with configurable selector.

libbeat/outputs/elasticsearch/elasticsearch.go
urso: The way connection timeout works has changed a little. One sets/configures the timeouts when creating the client instance by now. The old implementation os passing the timeout to `Connect` did duplicate effort of storing/remembering these timeouts, besides always passing the same value already used to initialize the client. Plus, for applying the timeout on Connect, clients have had to adapt and reset config or actual client instances which have already been configured with the same values.
=> removed timeout parameter

libbeat/monitoring/report/elasticsearch/elasticsearch.go
urso: This is the monitoring reporter, not the actual Elasticsearch event output.
The backoff parameters have become configurable, as one has(will have to) configure the backoff strategy when creating the `outputs.Group`. This is due this PR shifting some responsibilities from the outputs to the publisher pipeline itself (e.g. error handling, backoff, retry). Some of the shifting will be necessary, when introducing dynamic output reloading, as the pipeline must transfer active events from old to new outputs.

libbeat/outputs/backoff.go
urso: yeah. It's kind of a wrapper for NetworkClient. This will be moved closer to the libbeat pipeline itself. Backoff strategy + parameters will become config parameters in `outputs.Group`.

libbeat/outputs/elasticsearch/api.go
urso: yup, only was used in unit/intgeration tests to compare the serialized json with some test string.

libbeat/outputs/elasticsearch/client.go
urso: I'd prefer (also for metrics), using structs/interfaces to be passed to output clients, instead of globals. I introduced the `ClientStats` for the ES client, as I wanted to ensure, these stats not being accidentally used by the ES monitoring reporter.
By passing stats by struct, we can register/unregister metrics in the future, when reporting stats + reloading outputs.

libbeat/outputs/elasticsearch/client.go
urso: yes, go-structform being encoding-agnostic uses the `struct` struct-tag by default. For now I preferred the types to be compatible with both, stdlib json and go-structform.

libbeat/outputs/elasticsearch/client.go
urso: Which method for setting your own index?
We used to have some logic to first look into the event, then look into the event metadata and then use the actual outputs index setting.

Was the `first` method ever used? I did keep the logic of allowing a beat to set index and ingest pipeline via event metadata fields (used/configured by filebeat modules).

libbeat/outputs/kafka/config.go
urso: worker feature?

The `worker` setting was (and still is) indirectly implemented by ReadHostList. If worker > 0, the function just duplicates the hostnames. But for the kafka output it doesn't really makes sense to use worker.

libbeat/outputs/kafka/kafka.go
urso: no. kafkaLogger always used to be here.

libbeat/outputs/logstash/async.go
urso: ouch, nicely spotted.

libbeat/outputs/logstash/ca_invalid_test.pem
urso: No, was mostly by accident I guess.

libbeat/outputs/logstash/client_test.go
urso: :)

libbeat/outputs/logstash/event.go
urso: Will remove file. Thanks.

libbeat/outputs/logstash/logstash.go
urso: yes. I modified the pipeline and output interfaces to be of one kind only. Outputs always get batches and on input it's of pipeline it's always by one event being pushed. This simplifies/removes some extra logic in outputs and pipeline itself for dealing with bulk and non-bulk event handling.

libbeat/publisher/bc/publisher/client.go
urso: Different places? The bc client is adapting the event, before pushing to the new pipeline. Maybe in some unit tests you mean?

libbeat/publisher/broker/brokertest/log.go
urso: originally I have had my own logger interface, to correctly capture log output within a test context (so log output and test name do make some more sense if something goes wrong). But when adopting pipeline to libbeat I opted for using logp. This enforced the introduction of withLogOutput, capturing stderr into `t.Log`, so I can still make sense of error logs mixed with pipeline processing logs.

libbeat/tests/system/beat/beat.py
urso: ?
I'm removing `@metadata` from events. All outputs but Elasticsearch print the events in JSON-format, as expected by logstash. including `@metadata`. So our beats->LS->ES configs are valid, even if kafka or redis is used. With `@metadata` being somewhat private, we never documented them. That is, I have to remove the fields, as a number of system tests do check all fields in an event are documented.

libbeat/monitoring/report/report.go
urso: No promisses ;)

libbeat/outputs/elasticsearch/api_test.go
urso: yup. Somewhat ugly to add a method for testing though. But it's some old code I didn't want to clean up as well.

libbeat/outputs/logstash/async.go
urso: code is correct. It's saying: set `r.err = err` if and only if `r.err` is not set yet. We report the first error only, as upcoming errors are most likely report about connection being closed, but root-cause is first error.

tsg:
@urso `test_invalid_config_cli_param` is failing on both travis and jenkins.

urso:
@tsg looking into it. Interestingly test is completely unrelated to any changes in this PR + passes for me. Did restart travis job.

urso:
@tsg did fix test, let's wait for travis.

PR #4495, opened 2017-06-13
Add source path matching to `add_docker_metadata` processor

exekias:
This change adds `match_source` setting to automatically match `source` field from filebeat events. Default settings take care of most common scenario. It will match `/var/lib/docker/containers/<container_id>/*.log`. The `match_source_index` setting allows changing this to match other position in the path. Just use:

```
processors:
   - add_docker_metadata: ~
```

### DIFF
--- a/libbeat/processors/actions/extract_field.go
+++ b/libbeat/processors/actions/extract_field.go
@@ -57,18 +57,18 @@ func NewExtractField(c common.Config) (processors.Processor, error) {
 func (f extract_field) Run(event common.MapStr) (common.MapStr, error) {
 	fieldValue, err := event.GetValue(f.Field)
 	if err != nil {
-		return nil, fmt.Errorf("Error getting field '%s' from event", f.Field)
+		return nil, fmt.Errorf("error getting field '%s' from event", f.Field)
 	}
 
 	value, ok := fieldValue.(string)
 	if !ok {
-		return nil, fmt.Errorf("Could not get a string from field '%s'", f.Field)
+		return nil, fmt.Errorf("could not get a string from field '%s'", f.Field)
 	}
 
 	parts := strings.Split(value, f.Separator)
 	parts = deleteEmpty(parts)
 	if len(parts) < f.Index+1 {
-		return nil, fmt.Errorf("Index is out of range for field '%s'", f.Field)
+		return nil, fmt.Errorf("index is out of range for field '%s'", f.Field)
 	}
 
 	event.Put(f.Target, parts[f.Index])
--- a/libbeat/processors/actions/extract_field.go
+++ b/libbeat/processors/actions/extract_field.go
@@ -1,0 +1,91 @@
+package actions
+
+import (
+	"fmt"
+	"strings"
+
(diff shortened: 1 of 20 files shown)


exekias:
jenkins retest this please

exekias:
jenkins retest this please

exekias:
jenkins retest this please

exekias:
jenkins retest this please

exekias:
jenkins test this please

exekias:
jenkins test this

libbeat/processors/actions/extract_field.go
```diff
@@ -0,0 +1,91 @@
...
+		Index:     config.Index,
+		Target:    config.Target,
+	}
+	return f, nil
+}
+
+func (f extract_field) Run(event common.MapStr) (common.MapStr, error) {
+	fieldValue, err := event.GetValue(f.Field)
+	if err != nil {
+		return nil, fmt.Errorf("Error getting field '%s' from event", f.Field)
```
andrewkroh: Use lower case error strings. https://github.com/golang/go/wiki/CodeReviewComments#error-strings

libbeat/processors/actions/extract_field.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	value, ok := fieldValue.(string)
+	if !ok {
+		return nil, fmt.Errorf("Could not get a string from field '%s'", f.Field)
+	}
+
+	parts := strings.Split(value, f.Separator)
+	parts = deleteEmpty(parts)
```
andrewkroh: Doesn't this create is disconnect between the configured index and the index into parts?

libbeat/processors/actions/extract_field_test.go
```diff
@@ -0,0 +1,87 @@
...
+		// Configure input to
+		input := common.MapStr{
+			test.Field: test.Value,
+		}
+
+		actual := runExtractField(t, testConfig, input)
+
+		result, err := actual.GetValue(test.Target)
+		assert.NoError(t, err)
+		assert.Equal(t, result.(string), test.Result)
```
andrewkroh: Does this type assertion panic if there was an err? I'm not a fan of assert.NoError due to it continuing on error. Maybe use "if assert.NoError { assert.Equal }". Check other uses and possibly replace with an "if err then t.Fatal".

libbeat/processors/actions/extract_field.go
exekias: I don't fully get this question, but will explain why I had to delete empty parts. When you split a string like `/var/lib/docker/containers/foo` you get `[ '', 'var', 'docker', 'containers', 'foo' ]`. It's counter intuitive to ask for index 4 and get `containers` (I would expect `foo`) because of the empty string at the beginning. Things get worse if you have things like `/var//lib/...`

exekias:
jenkins retest this

libbeat/processors/actions/extract_field.go
andrewkroh: That answers my question (I was thinking of the `/var//lib/...` case).

PR #4496, opened 2017-06-13
Modify add_docker_metadata processor to lookup ContainerId from source path in case of filebeat

shyamradhakrishnan:
The add_docker_metadata processor cannot work with Filebeat because in typical docker json logs, the container id is not present. This change is to lookup the ContainerId from the source path,similar to Kubernetes processor. Being a newbie in go, please excuse if some of the usages are not per go coding standards. The solution is to basically make the "matching" for ContainerId pluggable and use  a different implementation for filebeat.

### DIFF
--- a/filebeat/beater/filebeat.go
+++ b/filebeat/beater/filebeat.go
@@ -95,16 +95,13 @@ func (fb *Filebeat) modulesSetup(b *beat.Beat) error {
 			" can ignore this warning.")
 		return nil
 	}
-	esClient, err := elasticsearch.NewConnectedClient(esConfig)
-	if err != nil {
-		return fmt.Errorf("Error creating ES client: %v", err)
-	}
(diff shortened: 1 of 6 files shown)


exekias:
Hi @shyamradhakrishnan, how unfortunate, just today, some hours ago, I opened this one https://github.com/elastic/beats/pull/4495

andrewkroh:
How does this differ from https://github.com/elastic/beats/pull/4352?

shyamradhakrishnan:
ooh, @exekias did not notice it on Monday and hence started working. 
Please ignore, I should close this request then right?

exekias:
If you feel that #4495 fills your needs we can close this one, anyway I'm happy to discuss the implementation there :)

Also thank you for contributing, I would like to encourage you to keep doing it :+1:

PR #4503, opened 2017-06-14
Add random startup delay to each metricset

### LINKED ISSUE #4010: Thundering herd when starting many Beats at the same time.
The timing of Beats is so accurate that they create bursty traffic when simultaneously restarted. In a large fleet with good configuration management, it's feasible that many hundreds of Beats could be restarted within one second of each other. They then proceed to stay in perfect sync.

thundering_beats

Perhaps an intial startup delay of `rand(period)` would be nice here?

andrewkroh:
Add random startup delay to each metricset to avoid the thundering herd problem. Fixes #4010.

Running with debug, at startup you will see:

```
2017/06/14 14:24:03.652815 wrapper.go:197: DBG  system/memory will start after 6.287460518s
2017/06/14 14:24:03.652818 wrapper.go:197: DBG  system/cpu will start after 1.102366376s
2017/06/14 14:24:03.652827 wrapper.go:197: DBG  system/load will start after 4.933446568s
2017/06/14 14:24:03.652838 wrapper.go:197: DBG  system/filesystem will start after 4.47437156s
2017/06/14 14:24:03.652847 wrapper.go:197: DBG  system/fsstat will start after 5.52497095s
2017/06/14 14:24:03.652823 wrapper.go:197: DBG  system/process_summary will start after 8.93791103s
2017/06/14 14:24:03.652837 wrapper.go:197: DBG  system/network will start after 1.274884509s
2017/06/14 14:24:03.652821 wrapper.go:197: DBG  system/process will start after 6.207285781s
```

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -59,6 +59,7 @@ https://github.com/elastic/beats/compare/v6.0.0-alpha1...master[Check the HEAD d
 *Heartbeat*
 
 *Metricbeat*
+- Add random startup delay to each metricset to avoid the thundering herd problem. {issue}4010[4010]
 
 *Packetbeat*
 
--- a/metricbeat/_meta/common.full.yml
+++ b/metricbeat/_meta/common.full.yml
@@ -21,3 +21,7 @@ metricbeat.config.modules:
 
   # Set to true to enable config reloading
   reload.enabled: false
+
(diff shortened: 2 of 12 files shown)


andrewkroh:
jenkins, test it

andrewkroh:
The test failure in Filebeat's `test_reload.Test.test_load_configs` seems unrelated to the change.

PR #4506, opened 2017-06-15
Filebeat modules: Machine learning jobs

tsg:
This adds support for loading ML configurations (job + datafeed) from the filebeat modules.  An example ML configuration is added to the Nginx Filebeat module. This sample applies ML anomaly detection on the response codes.

The loading is implemented as part of the `setup` command and part of the `--setup` flag.

If a job configuration with the same ID exists, it is not overwritten, because deleting jobs could potentially delete user data. The user should manually delete the jobs in the UI if they want to upgrade.

ToDOs for the overall implementation:

* [x] Implement the jobs loading support in libbeat
* [x] Read the jobs configurations from the filesets
* [x] Add xpack to default testing env (`snaphost.yml`)
* [x] system tests
* [x] Docs & changelog

### DIFF
--- a/filebeat/beater/filebeat.go
+++ b/filebeat/beater/filebeat.go
@@ -1,10 +1,11 @@
 package beater
 
 import (
-	"errors"
 	"flag"
 	"fmt"
 	"sync"
+
+	"github.com/pkg/errors"
 
 	"github.com/elastic/beats/libbeat/beat"
 	"github.com/elastic/beats/libbeat/common"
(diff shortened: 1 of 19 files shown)


tsg:
I pushed everything, but I expect the test to be failing until the Kibana snapshots are fixed (there's an issue with the current ones). But should be otherwise ready for reviews.

tsg:
Also, I'm leaving docs for later, since this will mainly mean document the `setup` command.

filebeat/beater/filebeat.go
```diff
@@ -66,8 +67,14 @@ func New(b *beat.Beat, rawConfig *common.Config) (beat.Beater, error) {
 		}
 	}
 
+	// In the setup command,
```
exekias: spurious comment?

tsg:
Squashed and rebased. Remaining test failures seem unrelated, but I'm hoping for some more green on this run.

exekias:
:100: tested by hand, it adds the job both with `filebeat setup` and `filebeat run -setup`. Just one note, the job is not started after adding it, so the user still has to start the job by herself

filebeat/module/nginx/access/machine_learning/datafeed_response_code.json
```diff
@@ -0,0 +1,44 @@
+{
+    "job_id": "JOB_ID",
```
ruflin: Only skimmed through the PR but this kind of looked fishy?

tsg: Yeah, it gets replaced in code with a generated job id. Didn't know how to reflect this otherwise.

tsg: To clarify, I don't mean a textual replacement. The json gets decoded and the key replaced.

PR #4519, opened 2017-06-16
Fix race condition in logging

exekias:
`logp.FileRotator` had an issue when `WriteLines` is called concurrently. Added a RWLock to protect both `rotator.current` and `rotator.currentSize`

### DIFF
--- a/libbeat/logp/file_rotator.go
+++ b/libbeat/logp/file_rotator.go
@@ -6,6 +6,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"sync"
 )
 
 const RotatorMaxFiles = 1024
@@ -21,6 +22,7 @@ type FileRotator struct {
 
 	current     *os.File
 	currentSize uint64
+	currentLock sync.RWMutex
 }
 
 func (rotator *FileRotator) CreateDirectory() error {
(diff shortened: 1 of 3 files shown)


libbeat/logp/file_rotator.go
```diff
@@ -73,16 +75,26 @@ func (rotator *FileRotator) WriteLine(line []byte) error {
 	}
 
 	line = append(line, '\n')
+
+	rotator.currentLock.RLock()
 	_, err := rotator.current.Write(line)
```
andrewkroh: Shouldn't the `Write` and the increment of the `currentSize` occur within the same lock block? Without that it seems that the `currentSize` may not be accurate when it is checked by other goroutines.

exekias: Yeah you are right, I was trying to keep `Write` under a `RLock` as it may block for a while, and sacrifice some consistency in `currentSize` in favor of performance. Putting everything under a write lock would make goroutines logging wait on others, but yeah, this is probably overkill. What do you think?

tsg: I'd say having `currentSize` slightly out of date is fine, we just use it for the rotator so it's ok if that starts a bit late.

exekias: what do you think @andrewkroh? I can also add some comment explaining this

andrewkroh: I'm OK with the behavior as is.

andrewkroh:
@exekias Does this need back-ported to 5.6? https://github.com/elastic/beats/issues/5131#issuecomment-335922370

exekias:
Didn't think about it but it may make sense?

PR #4523, opened 2017-06-19
Add support for json logging to file

ruflin:
The current support is very minimal and only logs the existing log structure in json instead of plain text. In the future further support for being able to log structs and more complex events could be added.

JSON logging is disabled by default and only available for file logging.

Changelog and docs are still missing. I want to discuss first if we should move forward with this.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -60,6 +60,7 @@ https://github.com/elastic/beats/compare/v6.0.0-alpha2...master[Check the HEAD d
 - New cli subcommands interface. {pull}4420[4420]
 - Allow source path matching in `add_docker_metadata` processor. {pull}4495[4495]
 - Add support for analyzers and multifields in fields.yml. {pull}4574[4574]
+- Add support for JSON logging. {pull}4523[4523]
 
 *Filebeat*
 
--- a/libbeat/logp/log.go
+++ b/libbeat/logp/log.go
@@ -1,10 +1,12 @@
 package logp
 
 import (
+	"encoding/json"
 	"fmt"
 	"log"
 	"os"
 	"runtime/debug"
+	"strings"
 	"time"
 )
 
@@ -30,6 +32,7 @@ type logger struct {
 	level             Priority
 	selectors         map[string]struct{}
 	debugAllSelectors bool
+	JSON              bool
 
 	logger  *log.Logger
 	syslog  [LOG_DEBUG + 1]*log.Logger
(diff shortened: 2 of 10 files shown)


tsg:
I like the approach, but note that we default to logging to syslog. If we decide to recommend people to use the file logger, we'll need some improvements to it (e.g. reopen files on signal, to support external rotating strategies).

ruflin:
Should we enable json logging also for syslog? Should we make it apply to all logging methods?

tsg:
You'll need a rebase on this, but you did it to yourself :)

libbeat/logp/log.go
```diff
@@ -73,21 +76,50 @@ func parseSelectors(selectors []string) (map[string]struct{}, bool) {
...
+		// Creates the log message and formats it
+		bytes = []byte(fmt.Sprintf("%s %s %s", timestamp, prefix, message))
+	}
+
 	if _log.toSyslog {
-		_log.syslog[level].Output(calldepth, fmt.Sprintf(format, v...))
+		if _log.JSON {
+			_log.syslog[level].Output(calldepth, string(bytes))
+		} else {
+			_log.syslog[level].Output(calldepth, fmt.Sprintf("%s", message))
```
tsg: The Sprintf here seems superfluous. `string(bytes)`?

libbeat/logp/log.go
```diff
@@ -73,21 +76,50 @@ func parseSelectors(selectors []string) (map[string]struct{}, bool) {
...
 	if _log.toFile {
-		// Creates a timestamp for the file log message and formats it
-		prefix = time.Now().Format(time.RFC3339) + " " + prefix
-		_log.rotator.WriteLine([]byte(fmt.Sprintf(prefix+format, v...)))
+		if _log.JSON {
+			_log.rotator.WriteLine(bytes)
+		} else {
+			// Makes sure all prefixes have the same length
+			prefix = prefix + strings.Repeat(" ", 4-len(prefix))
+			bytes = []byte(fmt.Sprintf("%s %s %s", timestamp, prefix, message))
```
tsg: I wonder if the Sprintf, which is basically a Join, could be implemented more efficient. A benchmark between this and the `strings.Join` version would be interesting.

If it can be optimized, it's probably worth it so that we don't affect the performance in the debug logging case.

libbeat/logp/log.go
ruflin: agree, was probably there because there was more before the refactoring ...

libbeat/logp/log.go
ruflin: @urso probably knows from the top of his head? :-D

CHANGELOG.asciidoc
```diff
@@ -60,6 +60,8 @@ https://github.com/elastic/beats/compare/v6.0.0-alpha2...master[Check the HEAD d
 - New cli subcommands interface. {pull}4420[4420]
 - Allow source path matching in `add_docker_metadata` processor. {pull}4495[4495]
 - Add support for analyzers and multifields in fields.yml. {pull}4574[4574]
+- Allow source path matching in `add_docker_metadata` processor {pull}4495[4495]
```
andrewkroh: A changlog entry for add_docker_metadata, is that a merge problem?

ruflin: yes :-(

tsg:
Merged in the current form.

dedemorton:
Removing "needs_docs" label because the docs are added in #4931

PR #4471, opened 2017-06-08
Fixes to import_dashboards

cwurm:
Fixes issues with `import_dashboards`:

 1. Do not try to re-create the `.kibana` index. First check if it already exists.
 2. Stop and report HTTP error code when dashboards archive cannot be downloaded.

PR #4525, opened 2017-06-19
Fix loading dashboards in ES 5.4 + tests connection timeout increase

tsg:
I started with investigating a flaky test and increasing the
connection timeout, but went down the rabbit hole a bit:

* The dashboard loader didn't work against 5.4, i think since #4471. This
 fixes it by checking the version and only setting single_type for > 6.0.
* GetTestingElasticsearch now asserts on errors and increases the connection
  timeout
* Several tests were calling Connect twice
* withQueryResult in the ES output, now enhances its error with the full response from ES

### DIFF
--- a/libbeat/outputs/elasticsearch/api.go
+++ b/libbeat/outputs/elasticsearch/api.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 
 	"github.com/elastic/beats/libbeat/logp"
+	"github.com/pkg/errors"
 )
 
 // QueryResult contains the result of a query.
@@ -52,7 +53,7 @@ func (r QueryResult) String() string {
 
 func withQueryResult(status int, resp []byte, err error) (int, *QueryResult, error) {
 	if err != nil {
-		return status, nil, err
+		return status, nil, errors.Errorf("%v. Response: %s", err, resp)
 	}
 	result, err := readQueryResult(resp)
 	return status, result, err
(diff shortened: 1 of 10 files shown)


libbeat/outputs/elasticsearch/api.go
```diff
@@ -52,7 +53,7 @@ func (r QueryResult) String() string {
 
 func withQueryResult(status int, resp []byte, err error) (int, *QueryResult, error) {
 	if err != nil {
-		return status, nil, err
+		return status, nil, errors.Errorf("%v. Response: %s", err, resp)
```
andrewkroh: This looks like a good candidate for `errors.Wrapf` as opposed to `Errorf`.

libbeat/outputs/elasticsearch/testing.go
```diff
@@ -38,14 +40,15 @@ func GetEsPort() string {
...
+func GetTestingElasticsearch(t *testing.T) *Client {
 	var address = "http://" + GetEsHost() + ":" + GetEsPort()
 	username := os.Getenv("ES_USER")
 	pass := os.Getenv("ES_PASS")
 	client := newTestClientAuth(address, username, pass)
 
 	// Load version number
-	client.Connect(3 * time.Second)
+	err := client.Connect(60 * time.Second)
+	assert.NoError(t, err)
```
andrewkroh: It would better to use `if err != nil { t.Fatal(e) }` so that the test doesn't run. `assert.NoError` is like `t.Error` in that the test keeps executing.

tsg:
jenkins, test it

PR #4526, opened 2017-06-19
Improve `add_docker_metadata` examples for their common cases

exekias:
Fields are already documented in
https://github.com/elastic/beats/blob/master/libbeat/docs/processors-using.asciidoc#adding-docker-metadata
so I left them out of examples

### DIFF
--- a/filebeat/filebeat.full.yml
+++ b/filebeat/filebeat.full.yml
@@ -517,15 +517,19 @@ filebeat.prospectors:
 #processors:
 #- add_docker_metadata:
 #    host: "unix:///var/run/docker.sock"
-#    match_source: true
-#    match_source_index: 4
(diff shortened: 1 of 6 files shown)


exekias:
jenkins retest this please

exekias:
jenkins restest this please

filebeat/filebeat.full.yml
```diff
@@ -516,16 +516,15 @@ filebeat.prospectors:
 #
 #processors:
 #- add_docker_metadata:
-#    host: "unix:///var/run/docker.sock"
```
tsg: IMO we can leave the `host` here, and maybe the `ssl` section as well, since i think it's common to have to change those, right?

PR #4531, opened 2017-06-20
Add umask to generator travis to fix system tests

ruflin:
System tests require on travis the umask to be changed as otherwise the config files generated for the system tests have the wrong access rights.

### DIFF
--- a/generator/beat/{beat}/.editorconfig
+++ b/generator/beat/{beat}/.editorconfig
@@ -1,0 +1,27 @@
+# See: http://editorconfig.org
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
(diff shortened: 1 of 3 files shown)


ruflin:
@exekias TBH not 100% sure. @andrewkroh probably knows more, but I assumed it's base on the travis environment. But now it seems we see this also with users developing on Ubuntu.

exekias:
Yep, that's my case

andrewkroh:
I think a solid fix to the problem (ammuming it's a strict.perm issue) would be to modify the helper method in the python code that writes the config files such that it uses 0640.

exekias:
Merging this, we can deal with the better solution later

PR #4532, opened 2017-06-20
Allow enabled to be used in fields.yml

ruflin:


ruflin:
It allows for example the following:

```
"a": {
    "type": "object",
    "enabled": false
}
```

Based on https://www.elastic.co/guide/en/elasticsearch/reference/current/enabled.html this will not index the object but still store in in _source.

PR #4534, opened 2017-06-20
Enable new travis image

tsg:
As per https://blog.travis-ci.com/2017-06-19-trusty-updates-2017-Q2

Not sure if we should merge this, since it's going to be made the default tomorrow, but opening it to see if the tests are passing on the new infra.


exekias:
:+1: I guess it's better not to merge this, good to know it'll work

PR #4536, opened 2017-06-20
Fix generation of object with enabled: false 

ruflin:
Each `object` type now contains `"type": "object"`. Previously it was `"properties": {}`. Having `object` defined instead of properties allows to add more entries and the usage of the default properties object.

This is a follow up PR for #4532

### DIFF
--- a/libbeat/template/field.go
+++ b/libbeat/template/field.go
@@ -111,30 +111,31 @@ func (f *Field) text() common.MapStr {
 }
 
 func (f *Field) array() common.MapStr {
-	return common.MapStr{}
+	return f.getDefaultProperties()
 }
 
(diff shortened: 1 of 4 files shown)


tsg:
The `properties: {}` ended up causing issues for arrays at index time, which made it pass through our automated tests, so now I'm extra careful with these changes :).  Does this change anything for the `array` type?  It would be nice to have a test that indexes an object with an `array`, `object`, etc. to test that the generated template is valid.

ruflin:
This change "should" only affect arrays if `enabled: false` is set. Do we already have somewhere such "index" tests? Otherwise will create some integration tests for the template.

ruflin:
I added integration tests which we can also use for all future "problems".

PR #4543, opened 2017-06-22
Normalize system.process.cpu.total.pct based on the number of cores

andrewkroh:
This fixes an inconsistency between the CPU usage values reported by the system.process metricset and the values reported by the system.cpu and system.core metricsets. The `system.process.cpu.total.pct` was scaled by the number of cores so it ranged on `[0, 100% * number_of_cores]`. The CPU usage percentages reported elsewhere range on `[0, 100%]`.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -8,7 +8,7 @@
 // Template, add newest changes here
 
 === Beats version HEAD
-https://github.com/elastic/beats/compare/v5.4.1...master[Check the HEAD diff]
+https://github.com/elastic/beats/compare/v5.5.0...master[Check the HEAD diff]
 
 ==== Breaking changes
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.asciidoc
```diff
@@ -19,6 +19,9 @@ https://github.com/elastic/beats/compare/v5.4.1...master[Check the HEAD diff]
 *Heartbeat*
 
 *Metricbeat*
+- Change `system.process.cpu.total.pct` to be normalized by the number of CPU cores
+  such that the value will now range between 0 and 100% which is consistent
+  with CPU usage percentages provided by the system cpu metricset.
```
tsg: Can you add the PR number? Also, I would optimistically add it under the 5.5.0 section, the final BC is not yet started, I think.

PR #4544, opened 2017-06-22
Scale system.cpu.*.pct metrics by the number of cores

andrewkroh:
Change all `system.cpu.*.pct` metrics to be scaled by the number of CPU cores such that the values range on `[0, 100% * number_of_cores]`. This will make the CPU usage percentages from the system cpu metricset consistent with the system process metricset. The documentation for these metrics already stated that on multi-core systems the percentages could be greater than 100%. This makes the code match the docs, **but does cause a change in behavior to the user**.

### DIFF
--- a/metricbeat/module/system/cpu/_meta/fields.yml
+++ b/metricbeat/module/system/cpu/_meta/fields.yml
@@ -6,14 +6,14 @@
     - name: cores
       type: long
       description: >
-        The number of CPU cores.
+        The number of CPU cores. The CPU percentages can range from `[0, 100% * cores]`.
 
     - name: user.pct
       type: scaled_float
       format: percent
       description: >
         The percentage of CPU time spent in user space. On multi-core systems, you can have percentages that are greater than 100%.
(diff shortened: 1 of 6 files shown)


metricbeat/module/system/cpu/_meta/fields.yml
```diff
@@ -6,14 +6,14 @@
     - name: cores
       type: long
       description: >
-        The number of CPU cores.
+        The number of CPU cores. The CPU percentages can range from `[0, 100% * cores]`.
```
tsg: Did you intend to add this here? Is it to explain how is the number of cores useful?

tsg:
Taking it in, to make sure it makes the cut for 5.5.0.

metricbeat/module/system/cpu/_meta/fields.yml
andrewkroh: Yeah, it was intentional (for the reason you mentioned above).

Nodens2k:
Hi there,
This change prevents Grafana from correctly rendering the total CPU usage as (1 - idle%). Is there any alternative way to calculate this value?

PR #4546, opened 2017-06-22
Enable Cgo for Filebeat & Heartbeat

tsg:
This is to enable plugin support, and to enable future support (during the
6.x cycle) for importing other C libraries.

Metricbeat & Packetbeat already had Cgo enabled. We're leaving Winlogbeat
without Cgo because we don't foresee a reason for it to need it.

This change also affects the community Beats, but they can revert to pure-go
    by adding `CGO=false` in their Makefile.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -62,8 +62,11 @@ https://github.com/elastic/beats/compare/v6.0.0-alpha1...master[Check the HEAD d
 - Load Ingest Node pipelines when the Elasticsearch connection is established, instead of only once at startup. {pull}4479[4479]
 - Add support for loading Xpack Machine Learning configurations from the modules, and added sample configurations for the Nginx module. {pull}4506[4506]
 - Add udp prospector type. {pull}4452[4452]
+- Enabled Cgo which means libc is dynamically compiled. {pull}4546[4546]
 
 *Heartbeat*
+
+- Enabled Cgo which means libc is dynamically compiled. {pull}4546[4546]
 
 *Metricbeat*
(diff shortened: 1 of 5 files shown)


tsg:
jenkins, package it

filebeat/Makefile
```diff
@@ -7,6 +7,10 @@ TEST_ENVIRONMENT?=true
 GOX_FLAGS=-arch="amd64 386 arm ppc64 ppc64le"
 ES_BEATS?=..
 
+TARGETS?="windows/amd64 windows/386 darwin/amd64"
+TARGETS_OLD?="linux/amd64 linux/386"
+CGO=true
```
ruflin: Should we change this in the libbeat Makefile to make it default for all beats?

tsg: I worry it would complicate the lives of the community beats. Also, it's still not enabled for Winlogbeat, but we could overwrite it in that one.

Overall, it seemed like a less risky move to do it per Beat, but it's true that we'll have to be careful to do this whenever we add a new Beat. Any strong arguments to make it the default?

ruflin: My only argument would be that if it breaks somewhere we know earlier. I see the community beats complication but would hope they use our packaging scripts, so they get it out of the box?

tsg: Ok, I was on the fence about it, so I changed the PR to make `CGO=true` the default in all Beats, and overwrite it to `false` in Winlogbeat.

tsg:
jenkins, package it

tsg:
Packaging tests are green: http://build-eu-00.elastic.co/job/beats-package-PR/107/

tsg:
It was all green, but I rebased it to solve a CHANGELOG conflict.

PR #4552, opened 2017-06-23
Reduce verbosity of some make commands

andrewkroh:
This stops some large commands from being echo'ed to the console by make. It reduces
the amount of noise in the output. These commands are affected:

make check
make fmt
make beat.test

### DIFF
--- a/libbeat/scripts/Makefile
+++ b/libbeat/scripts/Makefile
@@ -85,7 +85,7 @@ ${BEAT_NAME}: $(GOFILES_ALL) ## @build build the beat application
 
 # Create test coverage binary
 ${BEAT_NAME}.test: $(GOFILES_ALL)
-	go test $(RACE) -c -coverpkg ${GOPACKAGES_COMMA_SEP}
+	@go test $(RACE) -c -coverpkg ${GOPACKAGES_COMMA_SEP}
 
 .PHONY: crosscompile
 crosscompile: ## @build Cross-compile beat for the OS'es specified in GOX_OS variable. The binaries are placed in the build/bin directory.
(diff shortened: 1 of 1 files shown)


libbeat/scripts/Makefile
```diff
@@ -96,15 +96,15 @@ crosscompile: $(GOFILES)
...
 	@goimports -l ${GOFILES_NOVENDOR} | (! grep . -q) || (echo "Code differs from goimports' style" && false)
-	${FIND} -name *.py -exec autopep8 -d --max-line-length 120  {} \; | (! grep . -q) || (echo "Code differs from autopep8's style" && false)
+	@${FIND} -name *.py -exec autopep8 -d --max-line-length 120  {} \; | (! grep . -q) || (echo "Code differs from autopep8's style" && false)
 
 .PHONY: fmt
 fmt: python-env ## @build Runs `goimports -l -w` and `autopep8`on the project's source code, modifying any files that do not match its style.
-	goimports -l -w ${GOFILES_NOVENDOR}
-	. ${PYTHON_ENV}/bin/activate && ${FIND} -name *.py -exec autopep8 --in-place --max-line-length 120  {} \;
+	@goimports -l -w ${GOFILES_NOVENDOR}
+	@${FIND} -name *.py -exec ${PYTHON_ENV}/bin/autopep8 --in-place --max-line-length 120  {} \;
```
ruflin: Will this also silence the error output? Because that is sometimes convenient to directly see where the problem is.

andrewkroh: No, the `@` only stops `make` from echoing the command itself. https://www.gnu.org/software/make/manual/html_node/Echoing.html

PR #4553, opened 2017-06-23
Add normalized CPU values and number of cores

andrewkroh:
This include the contents of #4550 (so merge that first and the amount to review will be smaller).

- Added normalized CPU metrics for core, cpu, and process metricsets that are divided by the number of CPU cores.
- Refactored helper code used by core, cpu, and load metricsets.
- Deprecated use of `cpu_ticks` in favor of per metricset settings.


metricbeat/metricbeat.full.yml
```diff
@@ -68,8 +68,9 @@ metricbeat.modules:
   period: 10s
   processes: ['.*']
 
-  # if true, exports the CPU usage in ticks, together with the percentage values
-  #cpu_ticks: false
+  # Configure the metric types that are included by these metricsets.
+  cpu.metrics:  [percentages]  # The other available options are normalized_percentages and ticks.
```
ruflin: Should we add "quotes" around percentages? Seems we do this everywhere else with strings.

metricbeat/module/system/core/_meta/data.json
```diff
@@ -11,14 +11,14 @@
     },
     "system": {
         "core": {
-            "id": 0,
+            "id": 1,
             "idle": {
-                "pct": 0.9063,
-                "ticks": 22204290
+                "pct": 0,
```
ruflin: If you run it 2-3 more times, you should also get some values for the `pct`.

metricbeat/module/system/core/config.go
```diff
@@ -0,0 +1,46 @@
...
+// Core metric types.
+const (
+	percentages = "percentages"
+	ticks       = "ticks"
+)
+
+// Config for the system core metricset.
+type Config struct {
+	Metrics  []string `config:"core.metrics"`
+	CPUTicks *bool    `config:"cpu_ticks"` // Deprecated.
```
ruflin: Should we deprecate it in 5.6 and remove it here?

metricbeat/module/system/cpu/config.go
```diff
@@ -0,0 +1,48 @@
...
+// Config for the system cpu metricset.
+type Config struct {
+	Metrics  []string `config:"cpu.metrics"`
+	CPUTicks *bool    `config:"cpu_ticks"` // Deprecated.
+}
+
+// Validate validates the cpu config.
+func (c Config) Validate() error {
+	if c.CPUTicks != nil {
+		logp.Deprecate("6.0.0", "cpu_ticks is deprecated. Add 'ticks' to the cpu.metrics list.")
```
ruflin: See previous comment

metricbeat/module/system/core/config.go
monicasarbu: @ruflin Any reason to deprecate it in 5.6 instead of 6.0.0?

ruflin: I don't like deprecated features in the code base. If don't do it now, we have to wait until 7.0 to remove it. Seems like a good time to do it now. Any disadvantage doing it now?

metricbeat/metricbeat.full.yml
andrewkroh: I added the quotes for consistency. 3b23220

metricbeat/module/system/core/_meta/data.json
andrewkroh: Fixed in 96a5b91

tsg:
Merged #4550, can you rebase this one, please? The flaky container failures in Metricbeat/travis should also be fixed after rebase.

metricbeat/module/system/core/config.go
andrewkroh: Does it make sense to add deprecation warnings in 5.6 when there is no replacement for cpu_ticks in 5.6? I thought there should be some period of overlap where the two config options are available to aid in migration.

ruflin: you mean no replacement in `6.0`?

andrewkroh: I meant 5.6. But possibly I misunderstood the original suggestion. You were suggesting adding the deprecation warnings in 5.6 and not having a migration path in 5.6 (e.g. we are telling the user not to use `cpu_ticks`, but not offering them an alternative in 5.6)?

andrewkroh:
I rebased the PR to resolve the conflicts, and squashed the commits where I addressed action items because I think it would be best to "Rebase and merge" this PR.

PR #4554, opened 2017-06-24
Move processors into new pipeline

urso:
- Update all processors to operate in *beat.Event
- Update processors factory to use `*common.Config`, not `common.Config`. `common.Config` works, but go-ucfg interfaces/API assume to operate on pointers (passing by value might break functionality in future?)
- Update most processors interface to pass the processors type/context by pointer instead of by value (due to use of interfaces, the processors have been allocated on heap anyways).
- Install list of processors in new publisher pipeline. This allows publisher pipeline to account for events being dropped by filters, when reporting ACKs. From beats point of view, there should be no difference between dropped events (by filters) and ACKed events (by outputs). Interface mandates all events being properly ACKed by the publisher pipeline.
- Tags, Fields and beat meta data settings are now implicitely converted to Processors as well
- -> change ensures, all metadata (local and global) being applied before the
actual processors (installed with the pipeline) are executed. Ensures all
processors will see the very same events.
- introduce `(*processors.Processors).RunBC(common.MapStr) common.MapStr`. Will be removed, once the beats themselves configure publisher pipeline to run processors (filebeat/metricbeat are executing processors on their own).
- add `Private` field to `beat.Event` for event-based ACK callbacks. Use `Private` field to store additional event-metadata required for post-processing in the ACK callback (as processor pipeline can change event at will)
- Prepare moving beats to new pipeline, but introducing `ConnectX` to BC API, for beats to connect to shared global pipeline (Note: will be removed when BC layer is removed).
- Processors are still executed by the go-routine calling `(beat.Client).Publish`.
- New Processor Execution Order (C=client based processor, P=pipeline based processor)
  1. (P) extract EventMetadataKey fields + tags (to be removed in favor of 4)
  2. (P) generalize/normalize event
  3. (P) add beats metadata (name, hostname, version)
  4. (P) add pipeline fields + tags
  5. (C) add client fields + tags
  6. (P/C) apply EventMetadataKey fields + tags (to be removed in favor of 4)
  7. (C) client processors list
  8. (P) pipeline processors list
  9. (P) (if publish/debug enabled) log event
  10. (P) (if output disabled) dropEvent

For docs: event processing execution order would be nice to have.

### DIFF
--- a/libbeat/processors/actions/decode_json_fields_test.go
+++ b/libbeat/processors/actions/decode_json_fields_test.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/logp"
+	"github.com/elastic/beats/libbeat/publisher/beat"
 	"github.com/stretchr/testify/assert"
 )
 
@@ -190,13 +191,12 @@ func getActualValue(t *testing.T, config *common.Config, input common.MapStr) co
 		logp.LogInit(logp.LOG_DEBUG, "", false, true, []string{"*"})
 	}
 
-	p, err := newDecodeJSONFields(*config)
+	p, err := newDecodeJSONFields(config)
 	if err != nil {
 		logp.Err("Error initializing decode_json_fields")
 		t.Fatal(err)
 	}
 
-	actual, err := p.Run(input)
-
-	return actual
+	actual, _ := p.Run(&beat.Event{Fields: input})
+	return actual.Fields
 }
--- a/libbeat/processors/actions/drop_event.go
+++ b/libbeat/processors/actions/drop_event.go
@@ -3,6 +3,7 @@ package actions
 import (
 	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/processors"
+	"github.com/elastic/beats/libbeat/publisher/beat"
 )
 
 type dropEvent struct{}
@@ -12,13 +13,15 @@ func init() {
 		configChecked(newDropEvent, allowedFields("when")))
 }
 
-func newDropEvent(c common.Config) (processors.Processor, error) {
-	return dropEvent{}, nil
+var dropEventsSingleton = (*dropEvent)(nil)
+
+func newDropEvent(c *common.Config) (processors.Processor, error) {
+	return dropEventsSingleton, nil
 }
 
-func (f dropEvent) Run(event common.MapStr) (common.MapStr, error) {
+func (*dropEvent) Run(_ *beat.Event) (*beat.Event, error) {
 	// return event=nil to delete the entire event
 	return nil, nil
 }
 
-func (f dropEvent) String() string { return "drop_event" }
+func (*dropEvent) String() string { return "drop_event" }
--- a/libbeat/processors/processor.go
+++ b/libbeat/processors/processor.go
@@ -6,19 +6,19 @@ import (
 
 	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/logp"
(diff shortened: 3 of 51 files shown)


libbeat/processors/actions/drop_event.go
```diff
@@ -12,13 +13,15 @@ func init() {
...
 }
 
-func (f dropEvent) Run(event common.MapStr) (common.MapStr, error) {
+func (*dropEvent) Run(_ *beat.Event) (*beat.Event, error) {
 	// return event=nil to delete the entire event
 	return nil, nil
 }
 
-func (f dropEvent) String() string { return "drop_event" }
+func (*dropEvent) String() string { return "drop_event" }
```
ruflin: Didn't know this works, thought `_` is required.

libbeat/processors/actions/decode_json_fields_test.go
```diff
@@ -190,13 +191,12 @@ func getActualValue(t *testing.T, config *common.Config, input common.MapStr) co
...
+	p, err := newDecodeJSONFields(config)
 	if err != nil {
 		logp.Err("Error initializing decode_json_fields")
 		t.Fatal(err)
 	}
 
-	actual, err := p.Run(input)
-
-	return actual
+	actual, _ := p.Run(&beat.Event{Fields: input})
```
ruflin: No error check?

libbeat/processors/processor.go
```diff
@@ -51,38 +51,54 @@ func New(config PluginConfig) (*Processors, error) {
...
-	filtered := event.Clone()
-	var err error
-
-	for _, p := range procs.list {
-		filtered, err = p.Run(filtered)
+	for _, p := range procs.List {
+		var err error
+		event, err = p.Run(event)
 		if err != nil {
+			// XXX: We don't drop the event, but continue filtering here iff the most
```
ruflin: Could we add an "error" to the event in case a processor did not work so it can be seen in the event, that something went wrong?

libbeat/publisher/pipeline/processor.go
```diff
@@ -0,0 +1,220 @@
...
+}
+
+// newProcessorPipeline prepares the processor pipeline, merging
+// post processing, event annotations and actual configured processors.
+// The pipeline generated ensure the client and pipeline processors
+// will see the complete events with all meta data applied.
+//
+// Pipeline (C=client, P=pipeline)
+//
+// 1. (P) add EventMetadataKey fields + tags (to be removed in favor of 4)
```
ruflin: We probably should have that somewhere in the docs

libbeat/processors/actions/decode_json_fields_test.go
urso: the error has never been checked for. In this PR I'm trying to adjust interfaces, not fix/improve/change behavior in tests or other pieces.

libbeat/processors/processor.go
urso: Same here. Old/Current logic does ignore errors and at most prints a debug message. I don't intend to change any processing logic.

libbeat/publisher/pipeline/processor.go
urso: Yeah. Adding tag.

urso:
> There are quite a few changes on the JSON part. I wasn't sure how this is related to the other changes?

There's some shared json parsing logic treating `@timestamp` and such a little different. In processors we're using beat.Event, but the reader in filebeat is returning `common.MapStr`. As filebeat is still using the old publisher API, I had to convert to temporary `beat.Event` type. There is no change in logic.

> If I understand it right, config wise nothing changes for fields / tags, it's just the internal handling as processor. Also these cannot be configured by a user a processor as they are not registered processors.

Yep. The pipeline/processor.go defines some internal processors, not globally registered/configurable.

PR #4556, opened 2017-06-26
Increment kibana healthcheck retries

exekias:
timeout was being reached in some travis tests, this should avoid that


exekias:
still failing.. do not merge yet

PR #4562, opened 2017-06-27
Add file integrity metricset to Auditbeat audit module

andrewkroh:
This metricset monitors files or directories for changes. When a file is changed it will calculate a MD5, SHA1, and SHA256 hash.

```
auditbeat.modules:
- module: audit
  metricsets: [file]
  file.paths:
    binaries:
    - /usr/bin
    - /usr/sbin
```

This metricset is not driven by the Linux Audit Framework (that is a future possibility). It uses inotify (linux), fsevents (macos), and ReadDirectoryChangesW (windows) to watch for changes.

### DIFF
--- a/auditbeat/module/audit/file/event.go
+++ b/auditbeat/module/audit/file/event.go
@@ -1,0 +1,120 @@
+package file
+
+import (
+	"crypto/md5"
+	"crypto/sha1"
+	"crypto/sha256"
+	"encoding/hex"
+	"io"
+	"os"
+	"time"
+
+	"github.com/pkg/errors"
+
+	"github.com/elastic/beats/libbeat/common"
+)
+
+type Action uint8
+
+func (a Action) String() string {
+	if name, found := actionNames[a]; found {
+		return name
+	}
+	return "unknown"
+}
+
+const (
+	Unknown = iota << 1
+	AttributesModified
+	Created
+	Deleted
+	Updated
+	MovedTo
+	CollisionWithin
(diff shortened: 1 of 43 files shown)


auditbeat/module/audit/file/event.go
```diff
@@ -0,0 +1,120 @@
...
+func hashFile(name string) (md5sum, sha1sum, sha256sum string, err error) {
+	f, err := os.Open(name)
+	if err != nil {
+		return "", "", "", errors.Wrap(err, "failed to open file for hashing")
+	}
+	defer f.Close()
+
+	m5 := md5.New()
+	s1 := sha1.New()
+	s256 := sha256.New()
```
exekias: I'm worried about performance, have you tested it somehow to check CPU usages? I guess RAM should be fine

tsg: I asked something similar on the old PR: https://github.com/elastic/beats/pull/4486#discussion_r121345711

andrewkroh: I should setup benchmark to see what kind of CPU cost this has on some arbitrary file sizes. With the current implementation only being event driven I am not too concerned because I think it would be rare for the metricset to be hashing a lot of files at once.

But I think users are going to want the metricset report an initial event for every file the first time it runs. When we implement this feature then I think we will need to worry about throttling the work it's doing to prevent it from consuming lots of CPU and disk IO at startup.

PR #4563, opened 2017-06-27
Rename .full.yml to .reference.yml

ruflin:
This renaming should help to better explain what the `full` config file is about.


exekias:
:100:  great idea!

PR #4566, opened 2017-06-27
Add filebeat modules config reloading

exekias:
This PR adds support for module reloading, mimicking current prospectors functionality. The mechanism is the same, just define:

```
filebeat.config.modules:
  enabled: true
  path: configs/*.yml
  reload.enabled: true
  reload.period: 10s
```


CHANGELOG.asciidoc
```diff
@@ -68,6 +68,7 @@ https://github.com/elastic/beats/compare/v6.0.0-alpha2...master[Check the HEAD d
 - Add support for loading Xpack Machine Learning configurations from the modules, and added sample configurations for the Nginx module. {pull}4506[4506]
 - Add udp prospector type. {pull}4452[4452]
 - Enabled Cgo which means libc is dynamically compiled. {pull}4546[4546]
+- Add module config reloading mechanism {pull}4566[4566]
```
tsg: We should probably mark this as `beta`, since the config reloading in metricbeat is also beta.

filebeat/crawler/crawler.go
```diff
@@ -57,6 +60,17 @@ func (c *Crawler) Start(r *registrar.Registrar, configProspectors *common.Config
 		}()
 	}
 
+	if configModules.Enabled() {
+		logp.Beta("Loading separate prospectors is enabled.")
+
+		c.reloader = cfgfile.NewReloader(configModules)
+		// TODO add beatVersion here
```
tsg: I see you pass it above, so you could add it now?

tsg:
What I didn't see in this PR is where are the Ingest Nodes pipelines loaded. They need to be registered with the Elasticsearch output so they are loaded on each re-connect. See [here](https://github.com/elastic/beats/blob/master/filebeat/beater/filebeat.go#L100).

filebeat/fileset/factory.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/logp"
+	"github.com/mitchellh/hashstructure"
+
+	"github.com/elastic/beats/filebeat/channel"
+	"github.com/elastic/beats/filebeat/prospector"
+	"github.com/elastic/beats/filebeat/registrar"
+)
+
+// Factory is a factory for registrars
```
tsg: Is this really a factory of registrars? Or did you mean a factory of modules or prospectors?

exekias: Uhm, this was copied from here: https://github.com/elastic/beats/blob/master/filebeat/prospector/factory.go#L11, I'll better put factory of modules

exekias:
Thanks for the review, I'm adding on demand pipeline load for dynamic modules, what do you think about ML jobs? I'm inclined to leave them out of this

tsg:
@exekias Yeah, ML jobs and dashboards should be on `setup` only, so not on reload.

dedemorton:
Removing needs_docs tag. Docs were added in https://github.com/elastic/beats/pull/4930

PR #4578, opened 2017-06-29
Fix make tests command

ruflin:
`make fast-system-tests` was not updated with all params from `make system-tests` which could have the affect that some tests failed.


andrewkroh:
Do you want to add the same flags to [appveyor](https://github.com/elastic/beats/blob/master/.appveyor.yml#L85) too?

ruflin:
Hm, that would mean the variables also have to be set somewhere? I think for now it is ok to not have them as the defaults are fine.

PR #4580, opened 2017-06-29
Update to latest k8s version

exekias:
We no longer require our own fork, changes accepted by upstream

cc // @ruflin


exekias:
jenkins retest this

Issue #4512, opened 2017-06-15
Add file name reference to logging when closing harvester on close_timeout being reached


ppf2:
```
2017-06-15T11:08:04-07:00 INFO Closing harvester because close_timeout was reached. 
2017-06-15T11:17:56-07:00 INFO Closing harvester because close_timeout was reached. 
2017-06-15T11:26:47-07:00 INFO Closing harvester because close_timeout was reached. 
```

Can we add a file path reference so that it's easier to debug?

https://github.com/elastic/beats/blob/4aea2947ac817d849470df185f4207e47cde7285/filebeat/prospector/log/harvester.go#L190

ruflin:
https://github.com/elastic/beats/pull/4584 opened

PR #4584, opened 2017-06-30
Improve close_timeout logging

ruflin:
Closes https://github.com/elastic/beats/issues/4512

### DIFF
--- a/filebeat/input/log/harvester.go
+++ b/filebeat/input/log/harvester.go
@@ -193,7 +193,7 @@ func (h *Harvester) Run() error {
 
 	// Closes reader after timeout or when done channel is closed
 	// This routine is also responsible to properly stop the reader
-	go func() {
+	go func(source string) {
 
 		closeTimeout := make(<-chan time.Time)
 		// starts close_timeout timer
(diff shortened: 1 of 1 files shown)


andrewkroh:
There's a data race in two tests ([test report)](https://beats-ci.elastic.co/job/elastic+beats+pull-request+multijob-windows/518/testReport/). Here's the [log file](https://beats-ci.elastic.co/job/elastic+beats+pull-request+multijob-windows/518/beat=filebeat,label=windows/artifact/src/github.com/elastic/beats/filebeat/build/system-tests/run/test_harvester.Test.test_close_timeout/filebeat.log/*view*/).

ruflin:
Argh, it's probably the reason I didn't put in the logging in the first place :-( 

@ppf2 Seems like this could be more complex then initially thought.

tsg:
@ruflin what should we do with this PR?

ruflin:
@tsg I still think it would be nice to have this "logging" in but it probably requires some refactoring on a delicate part of filebeat to make it happen without a race condition. I would keep the PR open as a reminder but not set priority on it to get it in.

ruflin:
@tsg @ppf2 Seems like I found a rather simple solution. The only trade off is that the "old" name is reported in case of file rotation but I think that is ok.

Mac CI somehow had a hick up not related to this change.

PR #4591, opened 2017-07-02
Convert heartbeat to new publisher pipeline

urso:
Requires: #4554 

- install EventMetadata processing with beat.Client on connect
- now Task and Job have different return type:
  - a Task can optionally return event fields of type common.MapStr
  - a Job returns/reports full events by wrapping the tasks events into a beat.Event type with additional annotations and start timestamp
  - unexport some 'local' types
- slight enhancements to fields/tags testing in heartbeat system tests
- add 'processors' setting to heartbeat monitors (requires doc update)

### DIFF
--- a/heartbeat/beater/heartbeat.go
+++ b/heartbeat/beater/heartbeat.go
@@ -7,7 +7,6 @@ import (
 	"github.com/elastic/beats/libbeat/beat"
 	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/logp"
-	"github.com/elastic/beats/libbeat/publisher/bc/publisher"
 
 	"github.com/elastic/beats/heartbeat/config"
 	"github.com/elastic/beats/heartbeat/monitors"
@@ -17,9 +16,8 @@ import (
 type Heartbeat struct {
 	done chan struct{}
 
-	client    publisher.Client
 	scheduler *scheduler.Scheduler
-	manager   *MonitorManager
+	manager   *monitorManager
 }
 
 func New(b *beat.Beat, cfg *common.Config) (beat.Beater, error) {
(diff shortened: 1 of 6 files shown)


heartbeat/beater/heartbeat.go
```diff
@@ -66,11 +62,12 @@ func (bt *Heartbeat) Run(b *beat.Beat) error {
 
 	<-bt.done
 
+	bt.manager.Stop()
```
ruflin: could also be done as a defer statement on line 46 I think? Is there a reason you use `bt.manager instead of `manager`

heartbeat/beater/manager.go
```diff
@@ -189,15 +200,42 @@ func (m *Monitor) Update(configs []*common.Config) error {
 
 	// start new and reconfigured tasks
 	for id, t := range all {
-		job := createJob(m.manager.client, t)
-		t.cancel = m.manager.jobControl.Add(t.config.Schedule, id, job)
+		processors, err := processors.New(t.config.Processors)
+		if err != nil {
+			logp.Critical("Fail to load monitor processors: %v", err)
```
ruflin: Not for this PR but thinking what we should do in this case? Alternatively we could return an error and move the creating of the jobs into a second for loop. Means all jobs only get created if all tasks work. Not sure what the user expects.

heartbeat/beater/heartbeat.go
urso: line 46 is another function

heartbeat/beater/manager.go
urso: This loop is about creating tasks, after merging the tasks settings (via json file) with the actual monitor settings. That is, this loop also drives monitors config reloading and is to be replaced is metricbeat/filebeat config reloading. In the filebeat/metricbeat case, the reloading is less flexible in handline settings, allowing us to load the configs once and not per task.

heartbeat/monitors/util.go
```diff
@@ -362,11 +402,15 @@ func withStart(field string, start time.Time, r TaskRunner) TaskRunner {
 
 func (f *funcJob) Name() string { return f.settings.Name }
 
+func (f *funcJob) Run() (beat.Event, []JobRunner, error) { return f.run() }
+
 func (f funcTask) Run() (common.MapStr, []TaskRunner, error) { return f.run() }
 
+/*
```
ruflin: Leftover?

PR #4592, opened 2017-07-03
Non-xpack snaphosts for testing

tsg:
We're now testing with x-pack by default, but this means that we lack on testing without x-pack. This adds another version of the testing env: `make start ENV=snaphsot-noxpack.yml` which starts the stack without x-pack installed.


tsg:
jenkins, test it

PR #4593, opened 2017-07-03
Assume no xpack installed on 400 answer

tsg:
`GET /_xpack` returns 400 when xpack is not installed, so accept that as an
acceptable error. It would be nice to first check the `_nodes` API, but I think
we need this check anyway.


tsg:
jenkins, test it

Issue #4598, opened 2017-07-03
Publisher Pipeline


urso:
- [x] Introduce new publisher pipeline with BC layer #4492 
- [x] Move processors, tags and fields into publisher pipeline #4554 
- [x] Remove compat layer:
  - [x] Update packetbeat to new pipeline #4599
  - [x] Update filebeat to new pipeline #4641 #4644 
  - [x] Update metricbeat to new pipeline #4699 
  - [x] Update winlogbeat to new pipeline #4690 
  - [x] Update heartbeat to new pipeline #4591 
  - [x] Remove bc/publisher package
    - [x] Remove bc layer #4776 
    - [x] merge libbeat/beat and publisher/beat packages #4821 
- [x] Improve ACK handler:
  - [x] Add support for pipeline-global ACK handler #4641 
- [x] Make pipeline parameters configurable (remove old configurations) #4650 
- [x] Add checks to libbeat/filebeat/metricbeat to reject 5.x queue/spooler/filters setting on startup #4726 
- [x] improve pipeline/output metrics #4663 
- [x] Update packetbeat code generator #4956 
- [x] Documentation updates
  - [x] removed unused settings from docs: #4753
    - filebeat: `filebeat.spool_size`, `filebeat.publish_async`, `filebeat.idle_timeout`,
    - libbeat: `queue_size`, `bulk_queue_size`,
    - metricbeat`filters` in modules
  - [x] Document libbeat queue settings: #4884 
    - `queue.<queue_type>` with default (and only available) `mem`:
    - mem queue settings: `events`, `flush.min_events`, `flush.timeout`.
  - [x] Document local `tags`, `fields` and `processors` settings in heartbeat monitors, filebeat prospectors, winlogbeat eventlogs, metricbeat modules, packetbeat flows/plugins. #4753 
  - [x] Update developer docs to use new publisher pipeline

PR #4599, opened 2017-07-03
Use packetbeat protocols to publish beat.Event to new pipeline

urso:
- Use packetbeat protocols to publish beat.Event to new pipeline
- Move flows to new publisher pipeline
- Do not drop events on publish, if input is via pcap file (stabilize system
  tests with potential large number of transactions)
- add fields, tags and processor settings to flows and analyzer plugins (requires docs updates)

Followup: I noticed quite some duplication in protocols unit tests, for capturing/handling transactions. Some additional cleanup providing helpers common functionality for protocols unit-testing will be provided.

### DIFF
--- a/packetbeat/beater/packetbeat.go
+++ b/packetbeat/beater/packetbeat.go
@@ -11,6 +11,9 @@ import (
 	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/common/droppriv"
 	"github.com/elastic/beats/libbeat/logp"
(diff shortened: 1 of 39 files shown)


dedemorton:
The flows documentation now has the fields, tags, and processor settings, so I'm assuming that all the end-user docs are updated for this PR. I'm removing the needs_docs label. There's still a doc item open to update the dev docs for https://github.com/elastic/beats/issues/4598, so that'll cover any remaining work required for the dev docs.

PR #4605, opened 2017-07-04
Use the noxpack version of the env for Metricbeat

tsg:
This is done only in an attempt to speed up the tests (by avoiding the
Kibana optimizing step). Once we use the snapshot docker images, we should switch
back to xpack.

Not tested yet, opening PR to see if it improves  the test times.


tsg:
This reduced the time to 38m on this run. Is still a lot longer than it used to be (under 20m), but at least we seem to now have a buffer until the limit of 50m, which hit occasionally on past PRs.

PR #4606, opened 2017-07-04
Add support for keyword in dynamic fields

ruflin:
This PR allows fields under a namespace to be set to keyword instead of auto detection from elasticsearch.

* Fixes issue where dynamic fields needed a path in addition to a name, means it did not work properly on the top level
* Add tests for dynamic fields.

### DIFF
--- a/libbeat/template/field.go
+++ b/libbeat/template/field.go
@@ -137,8 +137,9 @@ func (f *Field) array() common.MapStr {
 
 func (f *Field) object() common.MapStr {
 
+	dynProperties := f.getDefaultProperties()
+
 	if f.ObjectType == "text" {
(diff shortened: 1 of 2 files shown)


libbeat/template/field.go
```diff
@@ -137,8 +137,9 @@ func (f *Field) array() common.MapStr {
 
 func (f *Field) object() common.MapStr {
 
+	dynProperties := f.getDefaultProperties()
+
 	if f.ObjectType == "text" {
```
andrewkroh: Each of these cases are mutually exclusive so I think a switch (or if/elseif) would be better.

ruflin: Change do switch / case statement. Have a look.

PR #4611, opened 2017-07-05
Update running-on-docker.asciidoc

wangxf1987:
Add full command for metricbeat in the monitoring-host section.
In the metricbeat DockerFile, that use CMD, and not ENTRYPOINT.
It's must be input full command until the ENTRYPOINT instead of CMD.

### DIFF
--- a/metricbeat/docs/running-on-docker.asciidoc
+++ b/metricbeat/docs/running-on-docker.asciidoc
@@ -19,7 +19,7 @@ docker run \
   --volume=/sys/fs/cgroup:/hostfs/sys/fs/cgroup:ro \ <2>
   --volume=/:/hostfs:ro \ <3>
   --net=host <4>
-  {dockerimage} -system.hostfs=/hostfs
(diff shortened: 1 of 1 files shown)


exekias:
LGTM but let me try to push https://github.com/elastic/beats-docker/issues/6 before going forward with this :)

exekias:
I've just opened https://github.com/elastic/beats-docker/pull/14, let see how it goes

exekias:
It seems entrypoint changes will only be available since 6.0, so this PR can go in, could you sign the CLA?

ruflin:
@dedemorton @exekias  I wonder if we should take over this fix for our docs?

@wangxf1987 Would be great if you could sign the CLA.

ruflin:
@exekias @dedemorton In the meantime 5.6 and 6.x are out. But the line we have in is still the same. I wonder what we should do here?

exekias:
I have opened https://github.com/elastic/beats/pull/6483

PR #4614, opened 2017-07-05
Save a single dashboard per json file

monicasarbu:
Currently, in case the dashboards are linked together, all the dashboards are saved in a single json file. If one changes, then all the dashboards are affected and need to be exported. To make it easier to update/change a single dashboard, this PR saves a dashboard in a single JSON file.


filebeat/module/system/module.yml
```diff
@@ -0,0 +1,12 @@
+dashboards:
```
ruflin: @tsg @monicasarbu Should we call this file `dashboards.yml`? I was at first confused what this file is about.

monicasarbu: @ruflin My idea was that we might want to include other configurations here as well in the future.  I agree to rename it to `dashboards.yml` and we can change the name to something else when we have other configurations to add to it.

ruflin: ah, I see. I kind of like the idea. We should make sure that these "local" config options are always overwritable in a global config file like we have the the module config options. Probably not that relevant for dashboards though.

PR #4617, opened 2017-07-05
Add prometheusbeat to community beats docs

boernd:


tsg:
Very cool, this means one can use Elasticsearch as a long term storage for Prometheus, right? Do you have any thoughts about submitting this to Prometheus itself, following the model of: https://www.youtube.com/watch?v=MuHkckZg5L0

boernd:
Yes exactly, we use it for forwarding (some) Prometheus metrics to Elasticsearch for long term storage and visualization. Thx for the link, will have a look...

PR #4635, opened 2017-07-10
Add consolebeat to community beats docs

qingyunha:


exekias:
Hi @qingyunha, thanks for sending the link! Could you please explain a bit of what consolebeat does? If I understand it correctly it sends standard input stream to elasticsearch, is that correct?

Also you may be interested to take a look to `stdin` prospector in filebreat: https://www.elastic.co/guide/en/beats/filebeat/master/configuration-filebeat-options.html#_type

Cheers

qingyunha:
@exekias Yes, you're right.  It turned out that filebeat had already done that. Feel free to close it.

Issue #4569, opened 2017-06-27
Normalize timestamp time zones


andrewkroh:
Index names that contain dates may be incorrect when the system timezone is not in UTC. The index name (`<beatname>-%{+yyyy.MM.dd}`) is generated based on the internal `@timestamp` value. The date formatter that converts the `%{+yyyy.MM.dd}` into a date doesn't do any conversion to UTC.

Some Beats may normalize their timestamps to UTC before publishing events and they will not be affected by this bug.

For confirmed bugs, please report:
- Version: 5.x, 6.0
- Operating System: all
- Steps to Reproduce: Use Beats on a system where the timezone is not configured to UTC.

andrewkroh:
Branches that have been patched:

- master (6.x) - #4658 
- 5.5 - #4683
- 5.x - #4684

PR #4636, opened 2017-07-10
Auditbeat file module improvements

andrewkroh:
Unify the behavior across operating systems.

- Add tests.
- Resolve UID/GID/SID into owner/group.
- Log warning if a file watch cannot be added. Previously it would fail at startup.
- Remove usage of fsevents on macOS and use kqueues instead. The data provided was inconsistent with the fsnotify. It was valuable data but the signal needs some refinement to make it consistent. The biggest issue is that multiple events are coalesced into one. Another issue is that the watches are recursive (this is great, but it's inconsistent) and we will need to allow a glob to be specified then match the incoming events against the glob (future enhancement).
- Document hash fields
- Initialize watches on Start() rather than at construction time.
- Update data.json to include file owner and group names.
- Add file type and symlink target path to the events.
- Filter fsnotify events with empty paths (saw this once on Windows).

And add a simple dashboard:

image


auditbeat/module/audit/file/event.go
```diff
@@ -23,71 +26,97 @@ func (a Action) String() string {
...
+		}
+
+		if runtime.GOOS == "windows" {
+			if info.SID != "" {
+				m["sid"] = info.SID
+			}
+		} else {
+			m["uid"] = info.UID
+			m["gid"] = info.GID
+			m["mode"] = fmt.Sprintf("%#o", uint32(info.Mode))
```
ruflin: Why do you use `#` here?

auditbeat/module/audit/file/eventreader_fsnotify.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+func (r *reader) Start(done <-chan struct{}) (<-chan Event, error) {
+	var paths []string
+	for _, filePaths := range r.config.Paths {
+		paths = append(paths, filePaths...)
+	}
+
+	for _, p := range paths {
+		if err := r.watcher.Add(p); err != nil {
```
ruflin: Every time I see this I'm thinking if we should add a new prospector type in filebeat based on inotify which only works with "most" OS. Could make to code a lot simpler.

auditbeat/module/audit/file/eventreader_fsnotify.go
```diff
@@ -0,0 +1,119 @@
...
+			}
+		}
+	}()
+
+	return r.outC, nil
+}
+
+func convertToFileEvent(e fsnotify.Event, maxFileSize int64) Event {
+	event := Event{
+		Timestamp: time.Now().UTC(),
```
ruflin: We should introduce a helper in `common` to get `time.Now()` that always includes `UTC`. I think there are several places where it's missing.

auditbeat/module/audit/file/fileinfo_bsd.go
```diff
@@ -3,43 +3,62 @@
...
 	"github.com/pkg/errors"
 )
 
-func addFileAttributes(e *Event, maxFileSize int64) error {
-	info, err := os.Lstat(e.Path)
+func Stat(path string) (*Metadata, error) {
+	info, err := os.Lstat(path)
 	if err != nil {
-		return errors.Wrap(err, "failed to stat file")
+		return nil, errors.Wrap(err, "failed to stat path")
```
ruflin: Would be nice to include the "failed" path in the error

auditbeat/module/audit/file/fileinfo_bsd.go
```diff
@@ -3,43 +3,62 @@
...
 	}
 
 	stat, ok := info.Sys().(*syscall.Stat_t)
 	if !ok {
-		return errors.Errorf("unexpected fileinfo sys type %T", info.Sys())
+		return nil, errors.Errorf("unexpected fileinfo sys type %T", info.Sys())
```
ruflin: perhaps also include the path?

auditbeat/module/audit/file/fileinfo_bsd.go
```diff
@@ -3,43 +3,62 @@
...
-			e.SHA256 = sha256sum
-			e.Hashed = true
-		}
+	fileInfo := &Metadata{
+		Inode: stat.Ino,
+		UID:   stat.Uid,
+		GID:   stat.Gid,
+		Mode:  info.Mode(),
+		Size:  info.Size(),
+		ATime: time.Unix(0, stat.Atimespec.Nano()).UTC(),
```
ruflin: I know it was in before, but I would assume Micro should be enough.

auditbeat/module/audit/file/fileinfo_linux.go
```diff
@@ -1,43 +1,62 @@
...
 	"github.com/pkg/errors"
 )
 
-func addFileAttributes(e *Event, maxFileSize int64) error {
-	info, err := os.Lstat(e.Path)
+func Stat(path string) (*Metadata, error) {
+	info, err := os.Lstat(path)
 	if err != nil {
-		return errors.Wrap(err, "failed to stat file")
+		return nil, errors.Wrap(err, "failed to stat path")
```
ruflin: See comments above for logging, also line 22.

auditbeat/module/audit/file/fileinfo_linux.go
```diff
@@ -1,43 +1,62 @@
...
+		GID:   stat.Gid,
+		Mode:  info.Mode(),
+		Size:  info.Size(),
+		ATime: time.Unix(0, stat.Atim.Nano()).UTC(),
+		MTime: time.Unix(0, stat.Mtim.Nano()).UTC(),
+		CTime: time.Unix(0, stat.Ctim.Nano()).UTC(),
 	}
 
-	return nil
+	switch {
```
ruflin: Looks like at least some of the code could be shared between linux and bsd (did not do the detailed comparison, devil is probably in the detail.

auditbeat/module/audit/file/metricset_test.go
```diff
@@ -45,3 +49,258 @@ func getConfig(path string) map[string]interface{} {
...
+	}
+}
+
+// assertSameFile asserts that two files are the same.
+func assertSameFile(t testing.TB, f1, f2 string) {
+	if f1 == f2 {
+		return
+	}
+
+	info1, err := os.Stat(f1)
```
ruflin: Should you use LStat here? if one of the params is a symlink to the file, do you expect this check to return true or false?

auditbeat/module/audit/file/event.go
andrewkroh: This will "add leading 0 for octal (%#o)". https://golang.org/pkg/fmt/

I was wanting something like `0644`, but maybe a different format string would be more appropriate because what I really want is a minimum of 4 octal digits.

auditbeat/module/audit/file/eventreader_fsnotify.go
andrewkroh: Yeah, this fsnotify package seems to work pretty well. The only downside is that you can't do recursive monitoring. Recursive monitoring is supported by Windows but this isn't exposed by fsnotify (the other OSes don't support it).

auditbeat/module/audit/file/eventreader_fsnotify.go
andrewkroh: That [reminds me](https://github.com/elastic/beats/issues/4569), I think I will open a PR to automatically convert all time.Time objects to UTC during event normalization. Then we won't need a helper because it won't matter.

auditbeat/module/audit/file/fileinfo_bsd.go
andrewkroh: Stat/Lstat return a `*os.PathError` that includes the path already so it would be redundant.

auditbeat/module/audit/file/fileinfo_bsd.go
andrewkroh: Not sure I follow... the `time.Unix()` function accepts nanoseconds which is why I'm using `Nano()`.

auditbeat/module/audit/file/fileinfo_linux.go
andrewkroh: It really could be, and I was debating this. The only difference between the BSD and Linux variants is that the `stat_t` structs use different member names for the time values. I could create a fileinfo_posix.go that's common to BSD + Linux. Then have a fileinfo_linux.go and fileinfo_bsd.go that provide a helper method for reading the time values from the `stat_t` objects. I'll try it out, but I'm kind of thinking that the complexity out weighs the value.

auditbeat/module/audit/file/metricset_test.go
andrewkroh: Yes, will fix.

auditbeat/module/audit/file/fileinfo_linux.go
andrewkroh: I changed this to reduce duplication.

auditbeat/module/audit/file/eventreader_fsnotify.go
andrewkroh: I opened #4658.

andrewkroh:
@ruflin I addressed the review feedback.

PR #4638, opened 2017-07-10
Add permissions configuration for file output

plassa-b:
Add the ability to change file permissions on file output for Metricbeat.


karmi:
Hi @plassa-b, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/beats/pull/4638.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

plassa-b:
Hi @karmi, I added the email as you asked. Tell me if you need anything else !

plassa-b:
Hey there, any news on this ? Should I make any change ? :)

karmi:
Thanks for addint the e-mail, @plassa-b, the CLA check is now green! I imagine the Beats team will get back to you shortly.

libbeat/docs/outputconfig.asciidoc
```diff
@@ -958,6 +958,7 @@ output.file:
   filename: {beatname_lc}
   #rotate_every_kb: 10000
   #number_of_files: 7
+  #permissions: 0666
```
andrewkroh: The commented out value should reflect the default value.

metricbeat/metricbeat.full.yml
```diff
@@ -953,6 +953,9 @@ output.elasticsearch:
   # default is 7 files.
   #number_of_files: 7
 
+  # Permissions to use for file creation. The default is 0600.
+  #permissions: 0666
```
andrewkroh: This file is generated based on [libbeat/_meta/config.reference.yml](https://github.com/elastic/beats/blob/d07f82d6dff6a7a8ee0ebcb166672ba1daf2a1cb/libbeat/_meta/config.reference.yml#L516-L536). After updating the file in libbeat you should run `make update` from the `beats` dir to re-generate the config files.

plassa-b:
Hi @andrewkroh, I just made the changes you asked for. I also merged with a more recent version of beats (i realised the `config.full.yml` was renamed as `config.reference.yml`, so I figured it would be simpler for you guys)

Also, I realised that I left a segfault when no value was given for permissions in the config file (when logging the value, I was dereferencing the value without checking it wasn't nil), so I fixed that and added the default properly where needed.

Hope it does the trick. Feel free to ask me if you need anything else.

plassa-b:
Hey guys, any news about this PR ?

ejsears:
Just checking into this as we have no other way to import data. Thanks!

andrewkroh:
LGTM. Could you add this configuration option to the documentation for the file output. https://github.com/elastic/beats/blob/master/libbeat/docs/outputconfig.asciidoc#configuration-options-4

plassa-b:
Sorry for the delay.

I'm not sure I understand, if I'm correct, the documentation was already added (https://github.com/plassa-b/beats/blob/master/libbeat/docs/outputconfig.asciidoc#file-output)

Here's the diff https://github.com/elastic/beats/pull/4638/files#diff-f1a89e6c3a1ce6f296f47ad83b989ff7R961

andrewkroh:
jenkins, test it

andrewkroh:
> I'm not sure I understand, if I'm correct, the documentation was already added 

Sorry, I missed that in the diff.

Can you please check the tests. It looks like this change is causing a panic in a test. https://beats-ci.elastic.co/job/elastic+beats+pull-request+multijob-windows/1451/beat=libbeat,label=windows/testReport/junit/(root)/logp/Test_Rotator/

plassa-b:
@andrewkroh Ok so, I took a look at this test and it seems to fail because `Permissions` were not properly initialized: they are not set when initializing the `Rotator`, and the test does not call the `CheckIfConfigSane` method.

So I'm wondering wether the test should initialize the `Permissions` (to avoid calling `CheckIfConfigSane`) or if they should be defaultly initialized somewhere else.

Any preference ?

andrewkroh:
Sorry for such a late response. I just realized that this never merged and it's a great enhancement.

> So I'm wondering wether the test should initialize the Permissions (to avoid calling CheckIfConfigSane) or if they should be defaultly initialized somewhere else.

For simplicity I would just fix the test to initialize `Permissions`.

Thanks!

andrewkroh:
jenkins, test it

andrewkroh:
Could you please rebase the PR on master. We added some new scripts in master that are used by Jenkins for testing.

plassa-b:
So I just actually did a `git rebase` from master, but giving that the PR is filled with commits that aren't from me, I'm wondering if that was actually what you wanted

ruflin:
@plassa-b You probably have to fetch the most recent version of based and rebase on that.

andrewkroh:
Merged in 5c51448f5fc0c9f458273c452e0ce855fc749a1f

I used the manual instructions to rebase and squash the PR. Thanks @plassa-b

PR #4641, opened 2017-07-10
Publisher pipeline enhancements

urso:
Changes/Improvements to the publisher pipeline (required for filebeat integration).

- Clear the broker event buffer when events got ACKed by outputs. This helps
  the garbage collector to pick up already ACKed events
- Introduce BufferConfig, for the pipeline to ask some of the brokers settings
  on event buffering.
- have broker report a max number of events. Have an upper bound on events
  (published/dropped) in the pipeline.
  Note:
  - The counter is currently only used for pipeline clients potentially
    dropping events. Due to the new pipeline setup normalizing events in the
    processors, this condition is always true.
  - still, the broker blocks if max event count is reached
    -> even in worst case the total number of events in the publisher pipeline
       should only be 2*max
- add ClientConfig.DropOnCancel:
  producer.Cancel disconnects the producer from the publisher pipeline.
  Only if DropOnCancel is true, an additional signal will be send to the
  publisher pipeline, to drop any active event.
  This allows the pipeline client to determine if dropping events on close is
  feasible or not
- Producer.Publish returns a boolean to indicate publish potentially
  failed by the producer being closed
- properly disconnect blocked producer on close of producer
- add assert/panic if ACKed events succeeds total events in membroker
- fix index update summing the end-pointer eventually leading to
  slice-out-of-bounds panic
- add `SetACKHandler(PipelineACKHandler) error` to pipeline interface
  => global listener for events being ACKed by publisher pipeline (in order)
- add more settings to ClientConfig:
  - Meta: provide `@metadata` to be added to each single event published
          (e.g. pipeline)
  - Events: registers callbacks for pipeline client internal events like:
      - Closing, Closed, Published, FilteredOut, DroppedOnPublish
- redo acker and creating acker:
  - ackers operate mostly on pipeline now, so PipelineACKHandler can be served
  - client ACKer hooks into pipeline ACKing when building the ACKer.
    When client is closed, the client ACK will see no more events from the
    still active pipeline ACKer
- add asserts/panics to pipeline ACKers in case of too many events (acked >
  published) being ACKed
- make pipeline.Client thread-safe (big mutex on Publish), so multiple
  go-routines can share a beat client for publishing events (requirement for
  filebeat)
- filter out empty events (event.Fields == nil), but ensure events are
- update pipeline processors order to:
  1. (P) extract EventMetadataKey fields + tags (to be removed in favor of 4)
  2. (P) generalize/normalize event
  3. (P) add beats metadata (name, hostname, version)
  4. (C) add Meta from client Config to event.Meta
  5. (P) add pipeline fields + tags
  6. (C) add client fields + tags
  7. (P/C) apply EventMetadataKey fields + tags (to be removed in favor of 4)
  8. (C) client processors list
  9. (P) pipeline processors list
  10. (P) (if publish/debug enabled) log event
  11. (P) (if output disabled) dropEvent

### DIFF
--- a/libbeat/monitoring/report/elasticsearch/elasticsearch.go
+++ b/libbeat/monitoring/report/elasticsearch/elasticsearch.go
@@ -17,6 +17,7 @@ import (
 	"github.com/elastic/beats/libbeat/outputs/outil"
 	"github.com/elastic/beats/libbeat/outputs/transport"
 	"github.com/elastic/beats/libbeat/publisher/beat"
+	"github.com/elastic/beats/libbeat/publisher/broker"
 	"github.com/elastic/beats/libbeat/publisher/broker/membroker"
 	"github.com/elastic/beats/libbeat/publisher/pipeline"
 )
@@ -103,14 +104,14 @@ func makeReporter(beat common.BeatInfo, cfg *common.Config) (report.Reporter, er
 		out.Clients = append(out.Clients, client)
 	}
 
-	broker := membroker.NewBroker(20, false)
-	settings := pipeline.Settings{
+	brokerFactory := func(e broker.Eventer) (broker.Broker, error) {
+		return membroker.NewBroker(e, 20, false), nil
+	}
+	pipeline, err := pipeline.New(brokerFactory, out, pipeline.Settings{
 		WaitClose:     0,
 		WaitCloseMode: pipeline.NoWaitOnClose,
-	}
-	pipeline, err := pipeline.New(broker, out, settings)
+	})
 	if err != nil {
-		broker.Close()
 		return nil, err
 	}
 
--- a/libbeat/publisher/beat/pipeline.go
+++ b/libbeat/publisher/beat/pipeline.go
@@ -7,9 +7,10 @@ import (
 )
 
 type Pipeline interface {
+	Close() error
 	Connect() (Client, error)
 	ConnectWith(ClientConfig) (Client, error)
-	Close() error
+	SetACKHandler(PipelineACKHandler) error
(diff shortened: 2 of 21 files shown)


libbeat/monitoring/report/elasticsearch/elasticsearch.go
```diff
@@ -103,14 +104,14 @@ func makeReporter(beat common.BeatInfo, cfg *common.Config) (report.Reporter, er
 		out.Clients = append(out.Clients, client)
 	}
 
-	broker := membroker.NewBroker(20, false)
-	settings := pipeline.Settings{
+	brokerFactory := func(e broker.Eventer) (broker.Broker, error) {
+		return membroker.NewBroker(e, 20, false), nil
```
ruflin: I know it was already in previously but curious how you came up with 20?

libbeat/publisher/beat/pipeline.go
```diff
@@ -57,6 +65,29 @@ type ClientConfig struct {
 	ACKLastEvent func(Event)
 }
 
+// ClientEventer provides access to internal client events.
+type ClientEventer interface {
+	Closing() // Closing indicates the client is being shutdown next
```
ruflin: Kind of expected these methods to return bool. But will probably see later how they are used.

urso: why a bool? These are notifications a beat can register too (e.g. adapt wgEvents in filebeat). The listener can not influence the publisher pipeline. It's more a Listener/Observer pattern, no Delegate or such.

PR #4643, opened 2017-07-10
Bump version to 6.0.0-beta1

andrewkroh:


andrewkroh:
The test failure on Jenkins is unrelated to this change (but I did open a PR to fix it #4645).

PR #4644, opened 2017-07-10
Move filebeat to new publisher pipeline

urso:
- remove filebeat/spooler and filebeat/publisher package -> all spooling and
  reporting published events is moved to the publisher pipeline.
  Difference between spooler/publisher to new pipeline is:
  The new publisher pipeline operates fully asynchronous
- have filebeat register an eventACKer with the publisher pipeline.
  The eventACKer will forward state updates of ACKed events to the registrar
- filebeat uses beat.Event for events
- update util.Data to use beat.Event:
  - store state in event.Private field for consumption by registry
- changes to filebeat/channels package:
  - introduce OutletFactory: connect to publisher pipeline, applying common
    prospector settings
  - remove Outleter.SetSignal and Outleter.OnEventSignal
  - add Outleter.Close
  - introduce SubOutlet (separate closing):
    - when a suboutlet is closed, the original outlet is still active
    - if underlying outlet is closed, the suboutlet becomes closed as well (can
      not forward anymore)
  - introduce CloseOnSignal: close Outlet once a 'done' channel is closed
  - most functionality from harvester.Forwarder is moved into the
    outlet/publisher pipeline client

Note:
Outlet shutdown with prospectors and harvesters is somewhat delicate. There are
3 shutdown signals to take into account:
- filebeat.done
- harvester.done
- outDone (signal used to unblock prospectors from registrar on shutdown).

An outlet is shared between all harvesters of a prospector and the prospector
itself. If outDone is closed, all outlets will be closed, unblocking
potentially waiting harvesters and prosepectors on filebeat shutdown.
The prospector uses a sub-outlet for sending state updates (being closed on
filebeat.done). The harvesters sub-outlet is closed when harveser.done is
closed.
The signals are only required to unblock an harvester/prospector on exit. On
normal shutdown, the outlets are closed after all workers have been finished.


filebeat/beater/filebeat.go
```diff
@@ -194,32 +193,20 @@ func (fb *Filebeat) Run(b *beat.Beat) error {
...
 
-	// Start publisher
-	publisher.Start()
 	// Stopping publisher (might potentially drop items)
 	defer func() {
 		// Closes first the registrar logger to make sure not more events arrive at the registrar
 		// registrarChannel must be closed first to potentially unblock (pretty unlikely) the publisher
 		registrarChannel.Close()
-		publisher.Stop()
+		defer close(outDone) // finally close all active connections to publisher pipeline
```
ruflin: Why is this a `defer` inside the defer?

filebeat/harvester/forwarder.go
```diff
@@ -21,59 +19,38 @@ type Forwarder struct {
...
 
 // ForwarderConfig contains all config options shared by all harvesters
 type ForwarderConfig struct {
-	common.EventMetadata `config:",inline"`      // Fields and tags to add to events.
-	Type                 string                  `config:"type"`
-	Pipeline             string                  `config:"pipeline"`
-	Module               string                  `config:"_module_name"`  // hidden option to set the module name
-	Fileset              string                  `config:"_fileset_name"` // hidden option to set the fileset name
-	Processors           processors.PluginConfig `config:"processors"`
+	Type string `config:"type"`
```
ruflin: Trying to understand why `type` is still here and all the others were moved.

filebeat/harvester/forwarder.go
```diff
@@ -21,59 +19,38 @@ type Forwarder struct {
...
-	data.Meta.Module = f.Config.Module
-	data.Meta.Fileset = f.Config.Fileset
-
 	if data.HasEvent() {
-		data.Event[common.EventMetadataKey] = f.Config.EventMetadata
-		data.Event.Put("prospector.type", f.Config.Type)
-
-		// run the filters before sending to spooler
-		data.Event = f.Processors.RunBC(data.Event)
+		data.Event.PutValue("prospector.type", f.Config.Type)
```
ruflin: Could we also add the type in the outlet? Probably we need the type in other places too?

filebeat/harvester/forwarder.go
```diff
@@ -21,59 +19,38 @@ type Forwarder struct {
...
 	}
 
-	ok := f.Outlet.OnEventSignal(data)
+	ok := f.Outlet.OnEvent(data)
```
ruflin: For my own note: I must check that there are still 2 different methods here, one blocking and one non blocking to make sure harvester is only closed when data acked (or in queue ...)

filebeat/prospector/log/harvester.go
```diff
@@ -98,9 +111,9 @@ func NewHarvester(
 	}
 
 	// Add outlet signal so harvester can also stop itself
-	outlet.SetSignal(h.done)
-
 	var err error
+
+	outlet = channel.CloseOnSignal(outlet, h.done)
```
ruflin: Comment on line 113 is now in the wrong place. I put this one as early as possible because of our still unsovled race condition.

filebeat/prospector/log/harvester.go
```diff
@@ -326,7 +344,7 @@ func (h *Harvester) SendStateUpdate() {
 
 	d := util.NewData()
 	d.SetState(h.state)
-	h.forwarder.Outlet.OnEvent(d)
+	h.publishState(d)
```
ruflin: Is this one blocking until the events are acked?

filebeat/prospector/log/prospector.go
```diff
@@ -590,15 +614,14 @@ func (p *Prospector) startHarvester(state file.State, offset int64) error {
 	// This is synchronous state update as part of the scan
 	h.SendStateUpdate()
 
-	p.registry.Start(h)
+	p.harvesters.Start(h)
```
ruflin: Seems you didn't like the `registry` naming :-)

filebeat/prospector/log/prospector.go
```diff
@@ -632,5 +653,11 @@ func (p *Prospector) Stop() {
 	// Stop all harvesters
 	// In case the beatDone channel is closed, this will not wait for completion
 	// Otherwise Stop will wait until output is complete
-	p.registry.Stop()
+	p.harvesters.Stop()
+
+	// close state updater
+	p.stateOutlet.Close()
```
ruflin: Is the comment on line 654 still valid. It is important for reloading, that `Stop` by default waits for completion of sending all events.

filebeat/prospector/prospector.go
```diff
@@ -77,7 +82,7 @@ func (p *Prospector) initProspectorer(outlet channel.Outleter, states []file.Sta
 	case harvester.RedisType:
 		prospectorer, err = redis.NewProspector(config, outlet)
 	case harvester.LogType:
-		prospectorer, err = log.NewProspector(config, states, outlet, p.done)
+		prospectorer, err = log.NewProspector(config, states, outlet, p.done, p.beatDone)
```
ruflin: In the past the p.beatDone was already part of the outlet. I assume this changes now as the prospector itselfs creates different outlets?

filebeat/prospector/redis/harvester.go
```diff
@@ -125,13 +126,15 @@ func (h *Harvester) Run() error {
...
-			"redis": common.MapStr{
-				"slowlog": subEvent,
+		data.Event = beat.Event{
+			Timestamp: time.Unix(log.timestamp, 0).UTC(),
+			Fields: common.MapStr{
+				"message": strings.Join(args, " "),
+				"redis": common.MapStr{
+					"slowlog": subEvent,
+				},
+				"read_timestamp": common.Time(time.Now()),
```
ruflin: Not directly related, but should this also be `UTC`?

filebeat/prospector/redis/prospector.go
```diff
@@ -52,17 +57,22 @@ func (p *Prospector) LoadStates(states []file.State) error {
 func (p *Prospector) Run() {
 	logp.Debug("redis", "Run redis prospector with hosts: %+v", p.config.Hosts)
 
+	if len(p.config.Hosts) == 0 {
+		return
```
ruflin: We should probably log an error here?

filebeat/prospector/redis/prospector.go
```diff
@@ -52,17 +57,22 @@ func (p *Prospector) LoadStates(states []file.State) error {
...
 			p.config.MaxConn, p.config.IdleTimeout, p.config.IdleTimeout)
 
-		var err error
-
 		h := NewHarvester(pool.Get())
-		h.forwarder, err = harvester.NewForwarder(p.cfg, p.outlet)
-		if err != nil {
-			logp.Err("Error: %s", err)
-		}
+		h.forwarder = forwarder
```
ruflin: It is ok that all harvesters share the same forwarder / outlet?

filebeat/tests/system/test_shutdown.py
```diff
@@ -43,27 +42,31 @@ def test_shutdown_wait_ok(self):
 
         # Wait until first flush
         self.wait_until(
-            lambda: self.log_contains_count("Flushing spooler") > 1,
+            lambda: self.log_contains_count("Publish event") > 200,
```
ruflin: The problem with setting here 200 instead of one is that the likelyhood of all log lines being already read increases. I want to try to shut down as early as possible. What is the reason you increased this one?

filebeat/harvester/forwarder.go
```diff
@@ -21,59 +19,38 @@ type Forwarder struct {
...
-	Type                 string                  `config:"type"`
-	Pipeline             string                  `config:"pipeline"`
-	Module               string                  `config:"_module_name"`  // hidden option to set the module name
-	Fileset              string                  `config:"_fileset_name"` // hidden option to set the fileset name
-	Processors           processors.PluginConfig `config:"processors"`
+	Type string `config:"type"`
 }
 
 // NewForwarder creates a new forwarder instances and initialises processors if configured
 func NewForwarder(cfg *common.Config, outlet Outlet) (*Forwarder, error) {
```
urso: Note: the forwarder is going to be removed in future iterations.

filebeat/tests/system/test_shutdown.py
```diff
@@ -43,27 +42,31 @@ def test_shutdown_wait_ok(self):
...
         # validate registry entry offset matches last published event
         registry = self.get_registry()
-        output = self.read_output()[-1]
+        outputs = self.read_output()
+        offset = registry[0]["offset"]
         assert len(registry) == 1
-        assert registry[0]["offset"] == output["offset"]
 
-    @unittest.skip("Skipping unreliable test")
+        # we allow for a potential race in the harvester shutdown here.
```
ruflin: Can you elaborate on this one?

libbeat/common/mapstr.go
```diff
@@ -101,12 +101,10 @@ func (m MapStr) Clone() MapStr {
 	result := MapStr{}
 
 	for k, v := range m {
-		innerMap, err := toMapStr(v)
-		if err == nil {
-			result[k] = innerMap.Clone()
-		} else {
-			result[k] = v
+		if innerMap, ok := tryToMapStr(v); ok {
```
ruflin: Kind of surprised to see a mapstr change in this PR. Why was that needed?

libbeat/publisher/broker/membroker/broker.go
```diff
@@ -219,9 +219,9 @@ func (b *Broker) eventLoop() {
 			events = b.events
 		}
 
-		b.logger.Debug("active events: ", activeEvents)
+		// b.logger.Debug("active events: ", activeEvents)
```
ruflin: on purpose commented out? also below

libbeat/publisher/broker/membroker/buf.go
```diff
@@ -264,7 +264,7 @@ func (b *eventBuffer) Len() int {
 }
 
 func (b *eventBuffer) Set(idx int, event publisher.Event, st clientState) {
-	b.logger.Debugf("insert event: idx=%v, seq=%v\n", idx, st.seq)
+	// b.logger.Debugf("insert event: idx=%v, seq=%v\n", idx, st.seq)
```
ruflin: comment?

filebeat/beater/filebeat.go
urso: ups, copy'n paste

filebeat/harvester/forwarder.go
urso: the forwarder is subject to be removed in future iterations. It's for the PutValue when publishing. In the meantime I added `ClientConfig.Fields` to register fields to be added to every event on publish.
Basically the Forwarder will be remove in favor of the `channel.Outleter`, which will be removed in the future, in favor if `beat.Client` (once we get a more sane registry handling).

filebeat/prospector/log/harvester.go
urso: fixing.

filebeat/harvester/forwarder.go
urso: the harvester get's a `publishState`, which uses the stateOutlet. The stateOutlet (owned by the prospector) and the Outlet of the harvesters are closed on different signals, each... so many hoops...

filebeat/prospector/log/harvester.go
urso: the harvester shouldn't care, but yes. publishState calls `stateOutlet` owned by the prospector. The `stateOutlet` is only closed on `fb.done` or when all harvesters + prospector main loop is done.

filebeat/prospector/log/prospector.go
urso: yeah... I got confused by too many registries...

filebeat/prospector/log/harvester.go
urso: The events ACKs are directly forwarded to the registrar. But as the harvester and the stateOutlet share the same beat.Client instance, the events will be forwarded in correct order to the registrar.

filebeat/prospector/log/prospector.go
urso: I `p.harvesters.Stop()` should wait for completion on shutdown.

filebeat/prospector/log/harvester.go
ruflin: It's important that a harvester only gets closed and releases the file handler if the state is persisted to disk, means this event is acked.

filebeat/prospector/prospector.go
urso: well, the matter is a little more complicated. We do have 3 channels for signaling shutdown. In the past we used to have 4 or more signalers (I got rid of spooler and publisher) :)

The outDone channel (filebeat.go) replaces the publisher closing, and is therefore closed much later. The beatDone channel is require when creating the stateOutlet, as this one needs to be unblocked when `fb.done` is closed (otherwise the crawler will not stop).

filebeat/prospector/redis/harvester.go
urso: no idea. Will change this to UTC...

filebeat/prospector/redis/prospector.go
urso: fixing.

filebeat/prospector/redis/prospector.go
urso: yes, requirements on ordering in registrar did force me to make the beat.Client thread-safe (big mutex)

filebeat/tests/system/test_shutdown.py
urso: before it was checking for 'Flushing spooler' message. which can be any number of events (up to spooler size). As spooler and any kind of flushing is gone, I have had to use a replacement on number of events being published. The test file contains like 50k events. 200 should be ok, by we can decrease it.

filebeat/tests/system/test_shutdown.py
urso: see `outlet.OnEvent` in channel/outlet.go.

```
	// Note: race condition on shutdown:
	//  The underlying beat.Client is asynchronous. Without proper ACK
	//  handler we can not tell if the event made it 'through' or the client
	//  close has been completed before sending. In either case,
	//  we report 'false' here, indicating the event eventually being dropped.
	//  Returning false here, prevents the harvester from updating the state
	//  to the most recently published events. Therefore, on shutdown the harvester
	//  might report an old/outdated state update to the registry, overwriting the
	//  most recently published offset in the registry on shutdown.
```

libbeat/common/mapstr.go
urso: ups, kind of leaked into this PR :)
I found the old processors ordering potentially overwriting a globally shared MapStr and was forced to introduce some `Clone()`, so the processors always operate on copies of globally shared MapStr instances. The amount of `fmt.Errorf` was so expensive it did totally kill throughput.

libbeat/publisher/broker/membroker/broker.go
urso: yeah, loads of debugs commented out in the fast path. I did still keep them, so I can re-enabled them If I really need them.

PR #4646, opened 2017-07-10
Remove clean target from Metricbeat

andrewkroh:
The clean target was removing files that are checked into the repo.
The tasks that generate the files are responsible for cleaning the old
files before generating the new ones so the custom clean target is
not required.


ruflin:
I remember I wanted to go the other direction in the past so that all the update tasks depend on the clean, but that cause other issues. If I remember correctly I didn't want to require `update` which depends on python to get a clean repo. But I see the potential confusion with missing files.

PR #4648, opened 2017-07-11
metricbeat: enhance the system/diskio

PChou:
metricbeat: enhance the system/diskio, add more metrics in output, includes:

- iostat.rrqmCps
- iostat.wrqmCps
- iostat.rrqCps
- iostat.wrqCps
- iostat.rBps
- iostat.wBps
- iostat.avgrqSz
- iostat.avgquSz
- iostat.await
- iostat.svctm
- iostat.busy

these metrics has same meaninig of `iostat`. docs and data.json are both updated.
this feature is only avaliable on linux.

the associated topic in elastic discuss is [Why system.diskio doesn’t produce I/O statistics fields like “iostat”](https://discuss.elastic.co/t/why-system-diskio-doesnt-produce-i-o-statistics-fields-like-iostat/91354)

### DIFF
--- a/metricbeat/module/system/diskio/_meta/fields.yml
+++ b/metricbeat/module/system/diskio/_meta/fields.yml
@@ -54,3 +54,58 @@
       type: long
       description: >
         The total number of of milliseconds spent doing I/Os.
+
+    - name: iostat.read_request_merged_per_sec
+      type: float
+      description: >
+          The number of read requests merged per second that were queued to the device.
+
+    - name: iostat.write_request_merged_per_sec
+      type: float
+      description: >
+          The number of write requests merged per second that were queued to the device.
+
+    - name: iostat.read_request_per_sec
+      type: float
+      description: >
+          The number of read requests that were issued to the device per second
+
+    - name: iostat.write_request_per_sec
+      type: float
+      description: >
+          The number of write requests that were issued to the device per second
+
+    - name: iostat.read_byte_per_sec
+      type: float
+      description: >
+          The number of Bytes read from the device per second.
+
+    - name: iostat.write_byte_per_sec
+      type: float
+      description: >
+          The number of Bytes write from the device per second.
+
+    - name: iostat.avg_request_size
+      type: float
+      description: >
+          The average size (in sectors) of the requests that were issued to the device.
+
+    - name: iostat.avg_queue_size
+      type: float
+      description: >
+          The average queue length of the requests that were issued to the device.
+
+    - name: iostat.await
+      type: float
+      description: >
+          The average queue length of the requests that were issued to the device.
+
+    - name: iostat.service_time
+      type: float
+      description: >
+          The average service time (in milliseconds) for I/O requests that were issued to the device.
+
+    - name: iostat.busy
+      type: float
+      description: >
+          Percentage of CPU time during which I/O requests were issued to the device (bandwidth utilization for the device). Device saturation occurs when this value is close to 100%.
--- a/metricbeat/module/system/diskio/diskstat_linux.go
+++ b/metricbeat/module/system/diskio/diskstat_linux.go
@@ -1,0 +1,97 @@
+// +build linux
+
+package diskio
+
+import (
+	"github.com/pkg/errors"
+	"github.com/shirou/gopsutil/disk"
+
+	"github.com/elastic/beats/metricbeat/module/system"
+)
+
+func Get_CLK_TCK() uint32 {
+	//return uint32(C.sysconf(C._SC_CLK_TCK))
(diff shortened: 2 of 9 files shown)


karmi:
Hi @PChou, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/beats/pull/4648.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

PChou:
sorry, @karmi , the email used in the [commit](https://github.com/elastic/beats/pull/4648.patch) is my company email, and is used to another github account, anyway to change the email in the commit ?

metricbeat/module/system/diskio/_meta/fields.yml
```diff
@@ -54,3 +54,58 @@
       type: long
       description: >
         The total number of of milliseconds spent doing I/Os.
+
+    - name: iostat.rrqmCps
```
exekias: I'm divided here, as these names don't follow the [convention](https://www.elastic.co/guide/en/beats/libbeat/current/event-conventions.html), while they are pretty common for sysadmins. Would like to know @ruflin opinion on these.

metricbeat/module/system/diskio/diskstat_linux.go
```diff
@@ -0,0 +1,96 @@
...
+	if last, ok = stat.lastDiskIOCounters[counter.Name]; !ok {
+		stat.lastDiskIOCounters[counter.Name] = counter
+		return result, nil
+	}
+
+	// calculate the delta ms between the CloseSampling and OpenSampling
+	deltams := 1000.0 * float64(stat.curCpu.Total()-stat.lastCpu.Total()) / float64(system.NumCPU) / float64(Get_CLK_TCK())
+	if deltams <= 0 {
+		return result, errors.New("The delta cpu time between close sampling and open sampling is less or equal to 0")
+	}
```
exekias: We don't calculate deltas, but send absolute values and calculate them on the fly, check https://www.elastic.co/guide/en/beats/metricbeat/current/metricbeat-metricset-system-network.html for instance

metricbeat/module/system/diskio/diskstat_linux_test.go
```diff
@@ -0,0 +1,14 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_Get_CLK_TCK(t *testing.T) {
+	//usually the tick is 100
+	assert.Equal(t, uint32(100), Get_CLK_TCK())
```
exekias: Was this intended or some tests are missing?

metricbeat/module/system/diskio/_meta/fields.yml
ruflin: I would vote for the long version. Reasons is that this will allow us to group some of the events into namespaces and use `units` like `.sec` which are in line with the other events.

I tried my google skills on the abbreviations below but mostly ended up on this discuss issue: https://discuss.elastic.co/t/why-system-diskio-doesnt-produce-i-o-statistics-fields-like-iostat/91354/3

karmi:
@PChou, the CLA check is now green, thanks for fixing the e-mail!

metricbeat/module/system/diskio/_meta/fields.yml
PChou: Thanks, i had no idea about the name convention, i would like to change them

metricbeat/module/system/diskio/diskstat_linux.go
PChou: i think these metrics calculation is not easy on the fly, so calculation and report to backend directly is necessary here

metricbeat/module/system/diskio/diskstat_linux_test.go
PChou: er, in theory, the CLK_TCK should be got using `sysconf(C._SC_CLK_TCK)`, need cgo enabled. and this should included in cpu module. but in order to calculate disk metrics, i have to know the cpu clock ticks, and hard code the value to `100`.

exekias:
hi @PChou, do you still plan to rename fields? I can give you a hand with that if needed :)

PChou:
@exekias Hi, i try to rename these fields, please review, thanks

exekias:
Thank you for your contribution! :tada:

metricbeat/module/system/diskio/diskstat_linux.go
```diff
@@ -0,0 +1,97 @@
...
+
+	rd_ios := counter.ReadCount - last.ReadCount
```
ruflin: @exekias We should probably make this variable more "go-like" during some refactoring.

PR #4649, opened 2017-07-11
Remove yacc source file from vendor

andrewkroh:
The yacc source file isn't needed because the generated parser.go file is checked in.
Something was causing yacc to sometimes be invoked which caused errors like [this](https://travis-ci.org/elastic/beats/jobs/252183513#L3848). I'm not 100% this sure will fix things because still don't understand what was invoking yacc.


ruflin:
@andrewkroh Did you remove these files manually? The part I worry is that the next time we make a govendor update these are just added again.

exekias:
LGTM, same question as @ruflin although I'm ok with this, don't think we need a fork to keep those files out

andrewkroh:
> Did you remove these files manually?

Yes, I did this manually. Another side effect is that `govendor status` says that the package has been modified (this must invalidate the checksum it stores in the vendor.json).

ruflin:
Merging to see if it solves our problem. It still means we need to find a way to properly version it so it doesn't get added again with the next update.

PR #4650, opened 2017-07-11
Update broker/pipeline setup

urso:
- remove settings: `filebeat.spooler`, `filebeat.publish_async`, `filebeat.idle_timeout`, `queue_size`, `bulk_queue_size`
- introduce new setting: `queue.<queue_type>`
- expose memory queue (type = 'mem') settings: `events`, `flush.min_events`, `flush.timeout`

Note: docs will be updated in a separate PR.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -19,6 +19,7 @@ https://github.com/elastic/beats/compare/v6.0.0-alpha2...master[Check the HEAD d
 - The `scripts/import_dashboards` is removed from packages. Use the `setup` command instead. {pull}4586[4586]
 - Change format of the saved kibana dashboards to have a single JSON file for each dashboard {pull}4413[4413]
 - Rename `configtest` command to `test config`. {pull}4590[4590]
+- Remove setting `queue_size` and `bulk_queue_size`. {pull}4650[4650]
 
 *Filebeat*
 
@@ -70,6 +71,8 @@ https://github.com/elastic/beats/compare/v6.0.0-alpha2...master[Check the HEAD d
 - Add support for analyzers and multifields in fields.yml. {pull}4574[4574]
 - Add support for JSON logging. {pull}4523[4523]
 - Add `test output` command, to test Elasticsearch and Logstash output settings. {pull}4590[4590]
+- Introduce configurable event queue settings: queue.mem.events, queue.mem.flush.min_events and queue.mem.flush.timeout. {pull}4650[4650]
+- Enable pipelining in Logstash output by default. {pull}4650[4650]
 
 *Filebeat*
 
@@ -81,6 +84,7 @@ https://github.com/elastic/beats/compare/v6.0.0-alpha2...master[Check the HEAD d
 - Add udp prospector type. {pull}4452[4452]
 - Enabled Cgo which means libc is dynamically compiled. {pull}4546[4546]
 - Add Beta module config reloading mechanism {pull}4566[4566]
+- Remove spooler and publisher components and settings. {pull}4644[4644]
 
 *Heartbeat*
 
--- a/auditbeat/auditbeat.reference.yml
+++ b/auditbeat/auditbeat.reference.yml
@@ -86,12 +86,25 @@ auditbeat.modules:
 # sub-dictionary. Default is false.
 #fields_under_root: false
 
-# Internal queue size for single events in processing pipeline
-#queue_size: 1000
+# Internal queue configuration for buffering events to be published.
+#queue:
+  # Queue type by name (default 'mem')
+  # The memory queue will present all available events (up to the outputs
+  # bulk_max_size) to the output, the moment the output is ready to server
+  # another batch of events.
+  #mem:
+    # Max number of events the queue can buffer.
+    #events: 4096
 
-# The internal queue size for bulk events in the processing pipeline.
-# Do not modify this value.
-#bulk_queue_size: 0
+    # Hints the minimum number of events stored in the queue,
+    # before providing a batch of events to the outputs.
+    # A value of 0 (the default) ensures events are immediately available
+    # to be sent to the outputs.
+    #flush.min_events: 0
+
+    # Maximum duration after which events are available to the outputs,
+    # if the number of events stored in the queue is < min_flush_events.
+    #flush.timeout: 0s
 
 # Sets the maximum number of CPUs that can be executing simultaneously. The
 # default is the number of logical CPUs available in the system.
@@ -284,9 +297,9 @@ output.elasticsearch:
   # Optional load balance the events between the Logstash hosts
   #loadbalance: true
 
-  # Number of batches to be send asynchronously to logstash while processing
+  # Number of batches to be sent asynchronously to logstash while processing
   # new batches.
-  #pipelining: 0
+  #pipelining: 5
 
   # Optional index name. The default index name is set to name of the beat
   # in all lowercase.
--- a/filebeat/config/config.go
+++ b/filebeat/config/config.go
@@ -20,9 +20,6 @@ const (
 
 type Config struct {
 	Prospectors      []*common.Config `config:"prospectors"`
-	SpoolSize        uint64           `config:"spool_size" validate:"min=1"`
-	PublishAsync     bool             `config:"publish_async"`
-	IdleTimeout      time.Duration    `config:"idle_timeout" validate:"nonzero,min=0s"`
 	RegistryFile     string           `config:"registry_file"`
 	ConfigDir        string           `config:"config_dir"`
 	ShutdownTimeout  time.Duration    `config:"shutdown_timeout"`
@@ -34,8 +31,6 @@ type Config struct {
 var (
 	DefaultConfig = Config{
 		RegistryFile:    "registry",
-		SpoolSize:       2048,
-		IdleTimeout:     5 * time.Second,
 		ShutdownTimeout: 0,
 	}
 )
--- a/libbeat/publisher/bc/publisher/pipeline.go
+++ b/libbeat/publisher/bc/publisher/pipeline.go
@@ -2,17 +2,15 @@ package publisher
 
 import (
 	"errors"
+	"fmt"
 
 	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/logp"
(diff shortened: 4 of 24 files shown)


metricbeat/metricbeat.reference.yml
```diff
@@ -431,12 +431,12 @@ metricbeat.modules:
...
 #fields_under_root: false
 
-# Internal queue size for single events in processing pipeline
-#queue_size: 1000
-
-# The internal queue size for bulk events in the processing pipeline.
-# Do not modify this value.
-#bulk_queue_size: 0
+# Internal queue configuration for buffering events to be published.
+#queue:
```
ruflin: It's a bit strange that we call it here queue and in the code broker. This will need some docs and changelog entry. But can happen in a follow up PR. Does this have to be in this PR?

auditbeat/auditbeat.reference.yml
```diff
@@ -86,12 +86,22 @@ auditbeat.modules:
...
+#queue:
+  # Queue type by name (default 'mem')
+  #mem:
+    # Max number of events to store in the queue.
+    #events: 4096
+
+    # Hints the minimum number of events stored in the queue,
+    # before providing a batch of events to the outputs.
+    # A value of 0 (the default) ensures events are immediately available
+    # to be send to the outputs.
```
ruflin: `s/send/sent/`

auditbeat/auditbeat.reference.yml
```diff
@@ -86,12 +86,22 @@ auditbeat.modules:
...
-
-# The internal queue size for bulk events in the processing pipeline.
-# Do not modify this value.
-#bulk_queue_size: 0
+# Internal queue configuration for buffering events to be published.
+#queue:
+  # Queue type by name (default 'mem')
+  #mem:
+    # Max number of events to store in the queue.
+    #events: 4096
```
ruflin: What happens when this is reached? Flush happens? Thinking if there should be `flush.min`, `flush.max` and `flush.timeout` or similar?

libbeat/publisher/bc/publisher/pipeline.go
```diff
@@ -63,11 +54,27 @@ func createPipeline(
...
 
-	brokerFactory := func(e broker.Eventer) (broker.Broker, error) {
-		return membroker.NewBroker(e, queueSize, false), nil
+	brokerType := "mem"
+	if b := shipper.Queue.Name(); b != "" {
+		brokerType = b
 	}
 
-	p, err := pipeline.New(brokerFactory, out, settings)
+	brokerFactory := broker.FindFactory(brokerType)
```
ruflin: As quickly discussed, later we should probably rename `broker` package to `queue` to make the relation to the config more obvious.

libbeat/publisher/broker/membroker/broker.go
```diff
@@ -57,18 +59,34 @@ func create(eventer broker.Eventer, cfg *common.Config) (broker.Broker, error) {
...
 	return b, nil
 }
 
 // NewBroker creates a new in-memory broker holding up to sz number of events.
 // If waitOnClose is set to true, the broker will block on Close, until all internal
 // workers handling incoming messages and ACKs have been shut down.
-func NewBroker(eventer broker.Eventer, sz int, waitOnClose bool) *Broker {
+func NewBroker(
+	eventer broker.Eventer,
+	sz int,
```
ruflin: Seems like the 3 params could be a config struct for borkers.

auditbeat/auditbeat.reference.yml
urso: Semantics is a little different from old spooler here. The queue/broker does not flush its buffers. By default it basically stream the events right through to the outputs (if output takes somewhat longer it will be presented with all events available - up to bulk_max_size).

By configuring min_events, the queue will block the output until:
- at least min events is reached
- timeout is trigger

The `min events` is more of a hint, as the output might still receive a smaller buffer.

metricbeat/metricbeat.reference.yml
urso: These are the user-visible settings. I decided to rename the broker package to queue, but first I want to get the changes to the settings in and documented, so we won't introduce BC-breaking changes in case package updates PR gets merged late.

auditbeat/auditbeat.reference.yml
urso: The events setting is the maximum number of events the queue can hold. It's independent of min_events and flush timer.

urso: Naming is a little weird here... as it's more like 'block-processing' settings...

ruflin: @urso what happens when `events` is reached? I assume it will block from adding more events? In case of metricbeat assuming sending events is slower then generating events but still all events are acked. What will happen?

winlogbeat/winlogbeat.reference.yml
```diff
@@ -51,12 +51,22 @@ winlogbeat.event_logs:
...
 #fields_under_root: false
 
-# Internal queue size for single events in processing pipeline
-#queue_size: 1000
-
-# The internal queue size for bulk events in the processing pipeline.
-# Do not modify this value.
-#bulk_queue_size: 0
+# Internal queue configuration for buffering events to be published.
+#queue:
```
andrewkroh: I'll open a PR to delete this [block of code](https://github.com/elastic/beats/blob/09e90c66c1a1aa7e5301e54ff61219af21e705f1/winlogbeat/config/config.go#L43-L62) from Winlogbeat so that `queue` isn't rejected.

auditbeat/auditbeat.reference.yml
urso: once `events` is reached, the queue will block (no more space). Upon ACK of N events, another N events can be stored in the queue (N producers will be unblocked).

winlogbeat/winlogbeat.reference.yml
urso: hehe, thanks.

auditbeat/auditbeat.reference.yml
dedemorton: "to be sent" is correct

urso: Thanks!

CHANGELOG.asciidoc
```diff
@@ -19,6 +19,7 @@ https://github.com/elastic/beats/compare/v6.0.0-alpha2...master[Check the HEAD d
 - The `scripts/import_dashboards` is removed from packages. Use the `setup` command instead. {pull}4586[4586]
 - Change format of the saved kibana dashboards to have a single JSON file for each dashboard {pull}4413[4413]
 - Rename `configtest` command to `test config`. {pull}4590[4590]
+- Remove setting `queue_size` and `bulk_queue_size`. {pull}4650[4650]
```
ruflin: Should we add a message to the 5.6 release that these values will change?

CHANGELOG.asciidoc
```diff
@@ -81,6 +83,7 @@ https://github.com/elastic/beats/compare/v6.0.0-alpha2...master[Check the HEAD d
 - Add udp prospector type. {pull}4452[4452]
 - Enabled Cgo which means libc is dynamically compiled. {pull}4546[4546]
 - Add Beta module config reloading mechanism {pull}4566[4566]
+- Remove spooler and publisher components and settings. {pull}4644[4644]
```
ruflin: Same here, we should probably have a message in 5.6 or even better, not let the beat start in 6.0 if these values are set?

auditbeat/auditbeat.reference.yml
```diff
@@ -284,9 +297,9 @@ output.elasticsearch:
   # Optional load balance the events between the Logstash hosts
   #loadbalance: true
 
-  # Number of batches to be send asynchronously to logstash while processing
+  # Number of batches to be sent asynchronously to logstash while processing
   # new batches.
-  #pipelining: 0
+  #pipelining: 5
```
ruflin: Is there a reason we changed the default to 5? Should probably be mentioned in the changelog?

filebeat/config/config.go
```diff
@@ -20,9 +20,6 @@ const (
 
 type Config struct {
 	Prospectors      []*common.Config `config:"prospectors"`
-	SpoolSize        uint64           `config:"spool_size" validate:"min=1"`
-	PublishAsync     bool             `config:"publish_async"`
```
ruflin: Glad to see this one removed ;-)

libbeat/publisher/broker/membroker/config.go
```diff
@@ -1,9 +1,26 @@
 package membroker
 
+import (
+	"errors"
+	"time"
+)
+
 type config struct {
-	Events int `config:"events" validate:"min=32"`
+	Events         int           `config:"events" validate:"min=32"`
```
ruflin: Kind of surprised to see a minimum here.

urso: well, we need to buffer some data for pushing to the outputs. Without queue the pipeline won't be able to operate -> enforce minimum so users don't use 1 or 0.

auditbeat/auditbeat.reference.yml
urso: - pipeline can complete operate async -> take advantage of this fact
- with default settings we have a high chance to publish very small batches (no flush timeout -> reduced latency) if producers don't have a high enough rate. In Logstash case we can easily compensate for this by pushing more batches on the link

urso: Updated changelog.

winlogbeat/winlogbeat.reference.yml
urso: this PR is adjusting winlogbeat.

urso:
Translating old configurations:

- rename `queue_size` -> `queue.mem.events * 2`
- remove `bulk_queue_size`
- rename or remove `filebeat.spool_size` -> `queue.mem.flush.min_events`
- rename or remove `filebeat.idle_timeout` -> `queue.mem.flush.timeout`

dedemorton:
I'm removing the needs_docs label here because I believe we've covered the doc requirements with the topic https://www.elastic.co/guide/en/beats/filebeat/current/configuring-internal-queue.html and the removed settings that are documented in the breaking changes for the release.

PR #4654, opened 2017-07-12
Add support back for pfring (Go 1.6+)

deepanshumehndiratta:


karmi:
Hi @deepanshumehndiratta, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/beats/pull/4654.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

deepanshumehndiratta:
@karmi done

deepanshumehndiratta:
@karmi can you look into this?

tsg:
@deepanshumehndiratta Thanks for the PR! However, I'm not sure we want to bring pfring back at the moment, because we don't have a good way of testing it and because it doesn't seem to have been largely used.

I think the reason for which it is not used too much is that the CPU performance bottleneck lies for the moment in Packetbeat itself, not in the sniffing technology. From example, we have only one decoding thread. That is partially because Packetbeat is meant to be a lightweight sniffer that sits on the application server, and monitors the traffic of a single application, not large data links. 

We do hope to improve the Packetbeat architecture and increase performance, but until we do that, I'd say the benefit of pfring over af_packet doesn't justify the complexity cost of pfring. Pinging also @urso for his view on this.

karmi:
@deepanshumehndiratta, it looks like the CLA check succeeds now, thanks for fixing the e-mail! However, see @tsg's comment about the larger issues above.

tsg:
Closing for the moment.

PR #4670, opened 2017-07-14
Remove newlines at start and end of methods

andrewkroh:
This removes blank lines from the start and end of functions. For example, this function

    func foo() {

        return bar

    }

would become

    func foo() {
        return bar
    }

This tightens up the code and allows you to see more context when viewing
the source code.

### DIFF
--- a/auditbeat/main_test.go
+++ b/auditbeat/main_test.go
@@ -20,7 +20,6 @@ func init() {
 
 // Test started when the test binary is started. Only calls main.
 func TestSystem(t *testing.T) {
-
 	if *systemTest {
 		main()
 	}
--- a/dev-tools/cmd/dashboards/export_dashboards.go
+++ b/dev-tools/cmd/dashboards/export_dashboards.go
@@ -22,7 +22,6 @@ type manifest struct {
 }
 
 func makeURL(url, path string, params url.Values) string {
-
 	if len(params) == 0 {
 		return url + path
 	}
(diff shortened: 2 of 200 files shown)


ruflin:
Did you use a script for that? I kind of liked the have it at the beginning but good to unify.

filebeat/fileset/fileset.go
```diff
@@ -37,7 +37,6 @@ func New(
 	name string,
 	mcfg *ModuleConfig,
 	fcfg *FilesetConfig) (*Fileset, error) {
-
 	modulePath := filepath.Join(modulesPath, mcfg.Module)
```
tsg: I think it's cases like this one when a blank line helps readability? Otherwise, there's no indentation hint between the function header and the block, making it hard to read, IMO.

ruflin: yes, was thinking the same.

andrewkroh: I agree, I think in cases where the function declaration spans multiple lines I will leave those blank lines alone. Will update.

andrewkroh: I updated the PR and it with from 210 files modified (485 deletions) to 198 files modified (438 deletions).

tsg:
Argh, all green, but now there's a conflict.

monicasarbu:
I am adding an empty line after the function definition, especially when the function has many arguments. If the others find it better to have no empty line, I can adjust my habits.

PR #4671, opened 2017-07-14
Order imports stdlib, third party, then elastic

andrewkroh:
Apply a consistent order to imports (stdlib, third party, then elastic).
This uses the `-local` flag of goimports to group our imports separate from
third-party imports. For example:

    import (
        "fmt"

        "github.com/pkg/errors"

    )

### DIFF
--- a/auditbeat/module/audit/file/event.go
+++ b/auditbeat/module/audit/file/event.go
@@ -12,8 +12,9 @@ import (
 	"strconv"
 	"time"
 
-	"github.com/elastic/beats/libbeat/common"
 	"github.com/pkg/errors"
+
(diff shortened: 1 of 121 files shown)


ruflin:
Nice, I wasn't aware of the flag.

tsg:
This also has a conflict now.

andrewkroh:
Conflict resolved. A filebeat test failed on windows (doubt it's related to reordering imports).

andrewkroh:
For vim users:

```
" vim-go options
let g:go_fmt_command = "goimports"
let g:go_fmt_options = "-local=github.com/elastic"
```

PR #4672, opened 2017-07-14
Fix flaky harvester_limit test

ruflin:
The test sometimes failed on slow machines. The reason was that the registry was written before the first event was published. So the waiting for the registry was triggered before an event ended up on disk and the test crashed because no output file was there. Now it waits until it contains at least one line.


ruflin:
Failing metricbeat build should not be related but appveyor will be interesting to see.

PR #4675, opened 2017-07-14
Include Kibana dashboards in the individual Beats packages

tsg:
This is a PoC to:

* include the kibana directory in the home path of the packages
* make the `setup.dashboards.directory` the default instead of the URL
* remove the `snapshot.url` and `snapshot` options
* keep the `setup.dashboards.url`, but without a default value
* keep the beats-dashboards package for the moment, but it won't be easy to use (you
  need to know the URL).

TODOs:

* [x] docs update
* [x] changelog

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,6 @@
 /build
 /*/data
 /*/logs
-/*/_meta/kibana/index-pattern
 /*/fields.yml
 /*/*.template*.json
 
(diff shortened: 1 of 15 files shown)


tsg:
jenkins, package it

exekias:
@tsg, do you have any plan to update packaging to add kibana folder? I think it's not included in current packages by default

PR #4677, opened 2017-07-15
Fix file integrity test case

andrewkroh:
The test was unreliable because occasionally an "updated" event is received when moving a file. This makes the test more robust.

### DIFF
--- a/auditbeat/module/audit/file/metricset_test.go
+++ b/auditbeat/module/audit/file/metricset_test.go
@@ -98,25 +98,20 @@ func TestEventReader(t *testing.T) {
 			t.Fatal(err)
 		}
 
-		if runtime.GOOS == "windows" {
-			updated := readTimeout(t, events)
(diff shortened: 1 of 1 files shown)


auditbeat/module/audit/file/metricset_test.go
```diff
@@ -98,25 +98,20 @@ func TestEventReader(t *testing.T) {
 			t.Fatal(err)
 		}
 
-		if runtime.GOOS == "windows" {
-			updated := readTimeout(t, events)
-			assert.Equal(t, Updated.String(), updated.Action)
+		received := readMax(t, 3, events)
+		if len(received) == 0 {
```
exekias: what if you only get 1 or 2 events? something would be missing then

andrewkroh: Correct. That's the flexibility this test case needs because sometimes an "updated" event isn't sent (and idk why).

PR #4687, opened 2017-07-17
Use libbeat/common/dtfmt to encode event timestamps

urso:


urso:
@andrewkroh yes, please.

PR #4688, opened 2017-07-17
Update go-structform

urso:
- includes minor improvements on inlining map[string]interface{} (common.MapStr).


ruflin:
Seems like you have to run `make update` for the notice file

PR #4691, opened 2017-07-18
Add Flatten method to MapStr

ruflin:
This allows to flatten a MapStr to a flat map[string]interface{}. Nested keys are transformed to keys seperated by dots for each namespace. This can be useful for testing and logging.

### DIFF
--- a/libbeat/common/mapstr.go
+++ b/libbeat/common/mapstr.go
@@ -156,6 +156,39 @@ func (m MapStr) String() string {
 	return string(bytes)
 }
 
+// Flatten flattens the given MapStr and returns a flat MapStr.
+//
+// Example:
+//   "hello": MapStr{"world": "test" }
+//
+// This is converted to:
+//   "hello.world": "test"
+//
+// This can be useful for testing or logging.
+func (m MapStr) Flatten() MapStr {
+	return flatten("", m, MapStr{})
+}
+
+// flatten is a helper for Flatten. See docs for Flatten. For convenience the
+// out parameter is returned.
+func flatten(prefix string, in, out MapStr) MapStr {
+	for k, v := range in {
+		var fullKey string
+		if prefix == "" {
+			fullKey = k
+		} else {
+			fullKey = fmt.Sprintf("%s.%s", prefix, k)
+		}
+
+		if m, ok := tryToMapStr(v); ok {
+			flatten(fullKey, m, out)
+		} else {
+			out[fullKey] = v
+		}
+	}
+	return out
+}
+
 // MapStrUnion creates a new MapStr containing the union of the
 // key-value pairs of the two maps. If the same key is present in
 // both, the key-value pairs from dict2 overwrite the ones from dict1.
--- a/libbeat/common/mapstr_test.go
+++ b/libbeat/common/mapstr_test.go
@@ -423,3 +423,71 @@ func TestAddTag(t *testing.T) {
 		}
 	}
 }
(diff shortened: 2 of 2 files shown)


libbeat/common/mapstr.go
```diff
@@ -156,6 +156,38 @@ func (m MapStr) String() string {
 	return string(bytes)
 }
 
+// Flatten flattens the given MapStr and returns a flat MapStr
```
andrewkroh: Add period.

libbeat/common/mapstr.go
```diff
@@ -156,6 +156,38 @@ func (m MapStr) String() string {
...
+//
+// Example:
+//   "hello": MapStr{"world": "test" }
+//
+// This is converted to:
+//   "hello.world": "test"
+//
+// This can be useful for testing or logging
```
andrewkroh: Add period.

libbeat/common/mapstr.go
```diff
@@ -156,6 +156,38 @@ func (m MapStr) String() string {
...
+func (m MapStr) Flatten() MapStr {
+	return m.flatten("")
+}
+
+// flatten is a helper for Flatten. See docs Flatten
+func (m MapStr) flatten(prefix string) MapStr {
+	fields := MapStr{}
+	for k, v := range m {
+		fullKey := fmt.Sprintf("%s.%s", prefix, k)
+		if prefix == "" {
```
andrewkroh: Use an if/else to avoid unnecessary string allocations.

libbeat/common/mapstr.go
```diff
@@ -156,6 +156,38 @@ func (m MapStr) String() string {
...
+//   "hello.world": "test"
+//
+// This can be useful for testing or logging
+func (m MapStr) Flatten() MapStr {
+	return m.flatten("")
+}
+
+// flatten is a helper for Flatten. See docs Flatten
+func (m MapStr) flatten(prefix string) MapStr {
+	fields := MapStr{}
```
andrewkroh: Instead of creating a new MapStr every time this steps into an inner object, you could pass the output MapStr as an argument and have the function write directly to the final MapStr. This would avoid the call to `Update()`.

libbeat/common/mapstr.go
ruflin: added

libbeat/common/mapstr.go
ruflin: added

libbeat/common/mapstr.go
ruflin: changed

libbeat/common/mapstr.go
ruflin: changed

libbeat/common/mapstr.go
```diff
@@ -156,6 +156,38 @@ func (m MapStr) String() string {
...
+	for k, v := range m {
+		fullKey := fmt.Sprintf("%s.%s", prefix, k)
+		if prefix == "" {
+			fullKey = k
+		}
+
+		if m, ok := v.(MapStr); ok {
```
ruflin: changed, wasn't even aware of the method.

ruflin:
@andrewkroh Thanks a lot for the detailed review and the benchmarks. I made the changes accordingly. I didn't worry about performance too much as I expect this mainly to be used for testing, but you never now. If code is there, it will be used.

PR #4692, opened 2017-07-18
Add support for interface and object to schema

ruflin:
Currently schema was only able to process values where the full tree was visible. For cases where the sub tree should be take as is, now `Ifc` can be used. It takes whaever type it is and adds it to an event. This works for Objects or Arrays.

### DIFF
--- a/libbeat/common/schema/mapstriface/mapstriface.go
+++ b/libbeat/common/schema/mapstriface/mapstriface.go
@@ -142,6 +142,19 @@ func Str(key string, opts ...schema.SchemaOption) schema.Conv {
 	return schema.SetOptions(schema.Conv{Key: key, Func: toStr}, opts)
 }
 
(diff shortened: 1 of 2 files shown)


libbeat/common/schema/mapstriface/mapstriface.go
```diff
@@ -142,6 +142,19 @@ func Str(key string, opts ...schema.SchemaOption) schema.Conv {
 	return schema.SetOptions(schema.Conv{Key: key, Func: toStr}, opts)
 }
 
+func toIntf(key string, data map[string]interface{}) (interface{}, error) {
```
andrewkroh: IMO, `Ifc` is a more common abbreviation than `Intf` (e.g. `toIfc`).

ruflin: SGTM. I was really undecided on this one.

ruflin:
jenkins, retest it

Issue #4540, opened 2017-06-21
Beats doc updates in 6.0


dedemorton:
**GA**

- [x] [DeDe] Document index setting change in the breaking changes. See https://github.com/elastic/beats/issues/5614
- [x] [DeDe] Add link that points from release notes to breaking changes doc.
- [x] [DeDe] Change getting started guides to show how to pass credentials by updating the config file - we do not want to recommend passing credentials at the command line.
- [x] [DeDe] Add docs about manually loading the template (LS users need to do this). See https://github.com/elastic/beats/issues/5615 and https://github.com/elastic/beats/issues/5080
- [x] [DeDe] Update docs to reflect https://github.com/elastic/beats/pull/4867 (by default, the user doesn't need to configure the credentials for Kibana as they are taken from the Elasticsearch).
- [x] [DeDe] Deprecate document_type in LS examples. 
- [x] [DeDe] Verify that the latest version of the breaking changes documentation reflects Pius's comments below.
- [x] For auditbeat kernel metricset:
  - [x] Document how to check if auditd is running, how to stop it, and how to disable it from starting at boot. `service auditd status`, `service auditd stop`, `chkconfig auditd off`
  - [x] Document how to stop journald from listening to audit messages to save CPU and disk. `systemctl mask systemd-journald-audit.socket`
- [x] Add info about setting a custom index name to the GS doc. See https://github.com/elastic/beats/issues/4254
- [x] Set up redirects for topics that have moved in 6.0.
- [x] For metricbeat:
  - [x] We removed `cpu_ticks` and replaced it with `cpu.metrics` and `core.metrics`. [diff](https://github.com/elastic/beats/pull/4553/files#diff-a641598f5625c7a21eb70c1662e5d693) https://github.com/elastic/beats/pull/4909
- [x] Docs around how to import/export dashboards are changing significantly to use Kibana API for importing dashboards. 
- [x] Change doc structure to accommodate prospector types. See https://github.com/elastic/beats/issues/4398
- [x] Add/update compatibility statements for all modules in all Filebeat and Metricbeat (and others?); make compatibility sections consistent, including for 3rd party Outputs (Kafka/Redit).
- [x] Add Beats usage with Cloud
- [x] Add a link to the ES doc regarding removal of _types

**RC1 and RC2**

- [x] [DeDe] Document modules.d layout for Metricbeat. See https://github.com/elastic/beats/issues/5207
[...]

PR #4699, opened 2017-07-18
Update metricbeat to use new publisher API

urso:
- update metricbeat to use new publisher API
- remove NewModules, NewWrappers... in favor of single instance constructors NewModule, NewWrapper, ...
- rename setting `filters` to `processors` (consistent with other beats)
- move fields, tags and processors to publisher pipeline (-> processors will handle complete event)
- adjust unit + integration tests


metricbeat/mb/builders.go
```diff
@@ -22,39 +22,32 @@ var (
...
 // will be unpacked into ModuleConfig structs). r is the Register where the
 // ModuleFactory's and MetricSetFactory's will be obtained from. This method
-// returns a mapping of Modules to MetricSets or an error.
-func NewModules(config []*common.Config, r *Register) (map[Module][]MetricSet, error) {
-	if config == nil || len(config) == 0 {
-		return nil, ErrEmptyConfig
+// returns a the Modules and its configured MetricSets or an error.
+func NewModule(config *common.Config, r *Register) (Module, []MetricSet, error) {
+	if !config.Enabled() {
+		return nil, nil, nil
```
andrewkroh: WDYT about returning an error in this case? Without an error callers always need to check if Module is non-nil.

metricbeat/mb/mb.go
```diff
@@ -206,24 +205,20 @@ func (b *BaseMetricSet) HostData() HostData {
 // the metricset fetches not only the predefined fields but add alls raw data under
 // the raw namespace to the event.
 type ModuleConfig struct {
-	Hosts      []string                `config:"hosts"`
-	Period     time.Duration           `config:"period"     validate:"positive"`
-	Timeout    time.Duration           `config:"timeout"    validate:"positive"`
-	Module     string                  `config:"module"     validate:"required"`
-	MetricSets []string                `config:"metricsets" validate:"required"`
-	Enabled    bool                    `config:"enabled"`
-	Filters    processors.PluginConfig `config:"filters"`
```
andrewkroh: So `filters` is now `processors`? I think it would be beneficial to users to `Validate()` that `filters` is not set. If it is set then include in the error that they should be using `processors`.

metricbeat/mb/module/connector.go
```diff
@@ -0,0 +1,46 @@
+package module
+
+import (
+	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/processors"
+	"github.com/elastic/beats/libbeat/publisher/bc/publisher"
+	"github.com/elastic/beats/libbeat/publisher/beat"
+)
+
+// Connector configures an establishes a beat.Client for publishing events
```
andrewkroh: `s/an/and/`?

metricbeat/mb/module/event.go
```diff
@@ -20,28 +20,19 @@ type EventBuilder struct {
...
 // filters.
-func (b EventBuilder) Build() (common.MapStr, error) {
+func (b EventBuilder) Build() (beat.Event, error) {
 	// event may be nil when there was an error fetching.
-	event := b.Event
-	if event == nil {
-		event = common.MapStr{} // TODO (akroh): do we want to send an empty event field?
+	fields := b.Event
+	if fields == nil {
+		fields = common.MapStr{} // TODO (akroh): do we want to send an empty event field?
```
andrewkroh: Could you please remove this comment. This seems to be well established behavior at this point.

metricbeat/mb/mb.go
andrewkroh: We'll need documentation updates too for this since `filters` is gone? If so please drop a note into https://github.com/elastic/beats/issues/4540.

metricbeat/mb/builders.go
```diff
@@ -22,39 +22,32 @@ var (
...
-// provided configuration data. config is a list module config data (the data
+// NewModule builds a new Module and its associated MetricSets based on the
+// provided configuration data. config contains config data (the data
 // will be unpacked into ModuleConfig structs). r is the Register where the
 // ModuleFactory's and MetricSetFactory's will be obtained from. This method
-// returns a mapping of Modules to MetricSets or an error.
-func NewModules(config []*common.Config, r *Register) (map[Module][]MetricSet, error) {
-	if config == nil || len(config) == 0 {
-		return nil, ErrEmptyConfig
+// returns a the Modules and its configured MetricSets or an error.
```
urso: argh... copy'n paste

metricbeat/mb/builders.go
urso: Yep, that's on purpose, but an value would make sense.

metricbeat/mb/mb.go
urso: See: #4598 . All this is already on the todo list.

metricbeat/mb/module/connector.go
urso: yep

metricbeat/mb/module/event.go
urso: k

metricbeat/mb/mb.go
andrewkroh: Great, I didn't see that.

PR #4701, opened 2017-07-19
Pre-allocate buffers for Line.decode and Line.advance

### LINKED ISSUE #4700: Unnecessary memory allocation in harvester/reader/line.go
I'm seeing issues with filebeat 5.5.0 spending a significant amount of time garbage collecting. After investigating, it appears that the `harvester/reader.Line` struct unnecessarily allocates buffers in two places instead of re-using a byte buffer for each Line struct. 

- in `decode`, it allocates a 1KB buffer each time it decodes a line: https://github.com/elastic/beats/blob/master/filebeat/harvester/reader/line.go#L154
- in `advance`, it allocates a `bufferSize` buffer each time instead of reusing a single byte slice: https://github.com/elastic/beats/blob/master/filebeat/harvester/reader/line.go#L109

For large volume / high velocity logs, this seems like an unnecessary cost in memory allocation and GC when both could be allocated a single time in `reader.NewLine(...)`

gdearment:
- fixes #4700
- pre-allocates byte slices for buffers that were allocated on each pass
  causing unnecessary memory allocation and GC

### DIFF
--- a/filebeat/harvester/reader/line.go
+++ b/filebeat/harvester/reader/line.go
@@ -39,15 +39,15 @@ func NewLine(input io.Reader, codec encoding.Encoding, bufferSize int) (*Line, e
 	}
 
 	return &Line{
-		reader:     input,
-		codec:      codec,
-		bufferSize: bufferSize,
-		nl:         nl,
-		decoder:    codec.NewDecoder(),
-		inBuffer:   streambuf.New(nil),
-		outBuffer:  streambuf.New(nil),
+		reader:        input,
+		codec:         codec,
+		bufferSize:    bufferSize,
+		nl:            nl,
+		decoder:       codec.NewDecoder(),
+		inBuffer:      streambuf.New(nil),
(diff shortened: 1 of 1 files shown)


tsg:
@gdearment I'm not too familiar to that code, but the test failure [here](https://travis-ci.org/elastic/beats/jobs/255120268#L2021) looks relevant.

tsg:
@gdearment Let us know if you want us to take over this PR. The optimization looks worthy, but it currently seems to break one of our tests.

gdearment:
That would be great. I tried to reproduce this locally and wasn't able to, so I'm likely doing something wrong locally. I'll be able to take a look at this early next week if you aren't able to take this over.

urso:
This PR hasn't seen any update for quite some long time. Related code has also change since this PR has been opened. We are closing the PR for now. We can reopen the PR, in case you think the improvement as still required and you want to continue working on it.

PR #4706, opened 2017-07-19
Cherry-pick #4351 to 5.x: Change grok pattern to fetch correct IP from X-Forwarded-For list

tsg:
Cherry-pick of PR #4351 to 5.x branch. Original message: 

This PR changes the grok pattern for the nginx access logs ingest file in a way, that it retrieves the correct IP if the X-Forwarded-For header was logged into instead of the remote_addr variable.

The X-Forwarded-For header is a non standard header which creates a lists of IPs through which proxies the request has passed as well as the original clients IP and looks basically like this:
```
X-Forwarded-For: client1, proxy1, proxy2
```
which results in a log line like this:
```
```

The new pattern retrieves the first IP, which is the important one, and matches non or all succeeding IPs that are concatinated with a comma and a space.


tsg:
The code change in this one was already done, but this copies the test files from master to make sure everything is ok.

PR #4710, opened 2017-07-19
Update filebeat template

andrewkroh:
The template was not up-to-date on 5.x filebeat.


tsg:
@andrewkroh I fixed the same as a commit in #4706

PR #4711, opened 2017-07-19
Use the beta1-snapshot in the testing env

tsg:
🤞🤞

This also adds a `CACHE_BUST` parameter to the dockfiles, making it easier to make sure all our jenkins slaves are using a new image.


andrewkroh:
You might need to do some magic (like adding a dummy query param to the final URL) in order to invalidate the docker cache on Jenkins.

tsg:
@andrewkroh I added a generic `CACHE_BUST` parameter to the docker files, I imagine we'll need this more than one time. Let's see how it goes.

PR #4713, opened 2017-07-19
Cherry-pick #4257 to master: Fallback on LevelRaw If the Level is not in the RenderingInfo section of the event

andrewkroh:
Cherry-pick of PR #4257 to master branch. Original message: 

Applies to Windows Vista and above only.

https://discuss.elastic.co/t/event-fields-missing-if-renderinginfo-is-empty/84709


tsg:
There are a few test failures, but none is in Winlogbeat, so I think it's safe to merge.

Issue #4422, opened 2017-05-30
Simplify the section about Configuring Filebeat


dedemorton:
Right now, the section about configuring Filebeat contains a nested section called Configuration Options (Reference). 

![image](https://cloud.githubusercontent.com/assets/14206422/26602308/f12a4df4-4537-11e7-86f2-9d7f69938aad.png)

There are a few problems with this section:
- The section is meant to provide reference-style info, but because the options aren't alphabetized and the search is not always very, um, robust, users can't always find what they're looking for in the reference info. Plus the reference info contains more guidance info than you would typically find in pure ref content.
 - As we've enhanced the book with more how-to type info, we keep encountering situations where we have content about a specific topic in two different places because we don't want to duplicate content. Expecting users to look in two places for configuration info is not ideal.
- We also have a full config file that provides short descriptions of every option, so it's possible that some users don't even use the reference section to understand the configuration options.
- Now that we want to add info about different types of prospectors, the current structure doesn't scale well because we can't add sub-levels in the navigation under Prospectors. 

I'd like to propose that we remove the Configuration Options (Reference) container, pop all the configuration topics up one level in the nav, rename them using task/goal-oriented headings, and merge duplicated and repeated content so that we have one place to go for info about subjects like managing multiline messages. So the TOC would look something like this:
````
Configuring Filebeat
  Setting up prosectors
    Logs
    Redis slowlogs
    Multiple prospectors
  Managing multiline messages
  Specifying global and general options
  Configuring the output
    Elasticsearch
    Logstash
    Kafka
    Redis
    File
    Console
  Filtering and enhancing the exported data
    Defining processors
    Adding cloud metadata
    Adding the local timezone
    Decoding JSON fields
    Dropping events
    Dropping fields from events
    Adding additional fields to events
    Adding Kubernetes metadata
  Using ingest node
  Load balancing
  Reloading the config file
  Changing the output codec
  Setting paths
  Loading dashboards
  Loading the index template
  Logging
  YAML tips and gotchas
  Regular expression support
````
If we want to provide pure reference info, I suggest that we take the content that we currently use to generate the beat.full.yml file and generate a Quick Reference document in asciidoc that we can include in the docs. The benefit of doing so is that the content would be discoverable through google and provide a formatted version of the content that's in the full yml file. @ruflin I think you proposed something similar last year.

dedemorton:
Of course, whatever we decide for Filebeat will need to be implemented across the docs for Beats.

[...]

PR #4716, opened 2017-07-19
Restructure beats config options

dedemorton:
This resolves https://github.com/elastic/beats/issues/4422 plus applies the same structural changes to all the Beats documentation. Here's a link to the [docs for preview](https://logstashplugins.firebaseapp.com/) (a bit outdated because I've pushed changes to the files since I published the example). Please ignore the fact that the URL says logstash. I kept kicking Firebase, but it wouldn't let me create and deploy to a new app.

This PR mostly covers structural and style changes. I also added some missing content, moved some stuff around, and changed some text to improve the flow. The only new content (other than some filler text) is what I've added to filebeat-modules-options.asciidoc. Other stuff that appears new was just moved from another file. I needed to move things to restructure the TOC.

Summary of changes:

Changes for all beats:

- Removed the Configuration Options (Reference) container from the TOC
- Got rid of the `reference` folder in the source (it was lame)
- Moved all topics under config up one level in the hierarchy
- Reorganized config topics and merged redundant info
- Made some content changes to improve the flow of ideas
- Changed headings from title case to sentence case (yay) to align new consistency guidelines that will come out soonish
- Changed all level 2 headings and below to use verbs instead of gerunds to align with our new guidelines (verbs are a lot easier to scan).
- Added code formatting (backticks) to config option names. This will make it easier for translators to identify text that shouldn't be translated.

Additional changes for Filebeat:
- Added a new topic "Specify which modules to run" (filebeat-modules-options.asciidoc) but I didn't hook it up because I'm on the fence about whether we should discuss the modules config under "Configuring Filebeat"

Additional changes for Packetbeat:
- Chunked section about protocols into separate topics (makes it easier to scan the TOC and see what we offer...plus that section was getting very long). The section is under Configuring Packetbeat > Specify which transaction protocols to monitor 
- Added a few simplistic examples to the protocol topics in case a user accesses the individual pages before seeing the overview. These examples should be fleshed out with more config options at some point maybe.

I lost track of the changes for the other beats, so I can't summarize them here.


packetbeat/docs/packetbeat-reference-yml.asciidoc
```diff
@@ -0,0 +1,11 @@
+[[packetbeat-reference-yml]]
```
dedemorton: Trying an experiment here. Is it worthwhile to provide this in the docs? This is sort of placeholder for a real quick reference that lists all config options.

monicasarbu: As discussed, let's add the reference file to all Beats docs.

dedemorton:
@monicasarbu This is ready for a review. I have a few issues I need to address (either before or after the PR is merged):

- [ ] Decide how to handle the include directive that pulls in the reference.yml file. Do you think it's a problem if we point directly to the file under beats/beatname? Do I need to worry about new config options being added in a patch release, or is that so unusual that I needn't worry? Making a copy of the reference file is not ideal IMO because the copy will need to be maintained.

- [ ] If we point to the files under beats/beatname, I need to update the `sources:` entry in the `conf.yml` after this PR is merged so that doc build can find the referenced 

- [ ] I should update the doc collector script for each beat to make sure the capitalization matches our new standard.

monicasarbu:
@dedemorton I don't see any issue if you point directly to the reference file.

metricbeat/docs/fields.asciidoc
```diff
@@ -6728,7 +6728,7 @@ The current number of idle client connections waiting for a request.
 
 
 [[exported-fields-php_fpm]]
-== php_fpm Fields
+== PHP-FPM Fields
```
dedemorton: @monicasarbu In my edits, I'm being consistent with our general approach of using the name of the product/service that's being monitored (for example, PHP-FPM) when we describe the module. Otherwise, especially in headings, it sort of looks like we are misspelling product names. But I'm wondering...do you think this might cause people to use the wrong module names in their configs (like php-fpm instead of php_fpm)? We do show a config example, so I might be overthinking this.

monicasarbu: I would say to leave the name of the module as it shows in the configuration (with `_` instead of `-`) as otherwise might be confusing for the user.

PR #4717, opened 2017-07-20
Adds a "type" field to the filesystem beat

### LINKED ISSUE #3459: Feature: Alllow filtering out some filesystem types
Feature Request
============

Original Thread: 
https://discuss.elastic.co/t/disk-space-metrics-wrong-when-working-with-overlay-mounts/72442

There is currently a problem when using overlay mounts. In such case, the mount point has the same disk size as the original disk. In Kibana, these sizes are summed what is not the reality.

It should be possible, to allow filtering out all such types to only transmit the real size of the disk.

cv:
Its value is copied from `sigar.FileSystem.SysTypeName` and passed through so that it becomes possible to filter out `overlay`, `tmpfs` or otherwise uninteresting filesystems.

Fixes #3459.


tsg:
@cv Thanks for working on this. Run `make update` to solve the check failure.

cv:
@tsg you're most welcome, and I'm sorry I can't seem to fix the build breakage through `make update`. I'm still pretty unfamiliar with the build/doc system, so do you have any pointers?

tsg:
When I run `make update` top level in you branch, it modifies two files:

```
	modified:   libbeat/_meta/kibana/5.x/index-pattern/libbeat.json
	modified:   libbeat/_meta/kibana/default/index-pattern/libbeat.json
```

Do you get the same? Or some error during `make update`?

tsg:
`make update` is our one command to re-create all generated files (includes the notice file, the docs, etc.)

cv:
Oh, this could be it. I get no changes when I run it. Could it be an env difference?

Output: https://gist.github.com/cv/4b596841382fb6d4193a86d95ba05d38

tsg:
Can you try a rebase to master (make sure master is up to date, then `git rebase master`) and them `make update` again?

cv:
Done… still, no files get updated. I'm going to set up a Linux build environment to see if there are any differences.

cv:
Ok, so when running on Linux (a little docker image did the job), the files do get updated and I get the same result as you, @tsg. I'm wondering why the builds failed even so, as nothing in the output immediately jumps at me.

tsg:
jenkins, test it

tsg:
@cv Travis is green now, we had some issues lately with Travis timing out on some of our stuff, so it's flaky at the moment. We're looking for solutions.

tsg:
@cv can you add a line in the `CHANGELOG.asciidoc` file, please?

metricbeat/docs/fields.asciidoc
```diff
@@ -8762,6 +8762,14 @@ The disk name. For example: `/dev/disk1`
 
 
 [float]
+=== system.filesystem.device_type
```
tsg: I'm thinking if we shouldn't name this simpler as `system.filesystem.type`. I usually think of `ext4` (for example) as a filesystem type, so that seems to fit. `device_type`, on the other had, makes me think of hardware.

cv: Good rationale. I'll change it in a few minutes.

cv:
This last build looks like the least fail-y so far :)

@tsg, I'd say work here is done, and unless you have any other points you'd like me to work on, can I clean up / rebase the commit history? Or do you typically do that when squashing into master?

tsg:
@cv Yes, please rebase & squash. I'd do it on merging, but now there's a conflict anyway. Thanks a lot for your work!

cv:
All squashed (except for the merge conflict fix, which I left as its own commit so it can be reviewed).

PR #4726, opened 2017-07-20
Check for removed settings on startup

urso:
- Introduce cfgwarn package
- move Beta/Deprecated/Experimental to from `logp` to `cfgwarn`
- introduce CheckRemoved5xSettings looking for removed settings in a config object + constructs an error message with completely expanded setting name
- use CheckRemoved5xSettigns in libbeat/filebeat/metricbeat to fail on startup if any settings removed by publisher refactoring is used
- Add system tests for libbeat/filebeat/metricbeat checking removed settings are correctly identified.

### DIFF
--- a/libbeat/common/cfgwarn/cfgwarn.go
+++ b/libbeat/common/cfgwarn/cfgwarn.go
@@ -1,0 +1,24 @@
+package cfgwarn
+
+import (
+	"fmt"
+
+	"github.com/elastic/beats/libbeat/logp"
+)
+
+// Beta logs the usage of an beta feature.
+func Beta(format string, v ...interface{}) {
+	logp.Warn("BETA: "+format, v...)
+}
+
+// Deprecate logs a deprecation message.
+// The version string contains the version when the future will be removed
+func Deprecate(version string, format string, v ...interface{}) {
+	postfix := fmt.Sprintf(" Will be removed in version: %s", version)
+	logp.Warn("DEPRECATED: "+format+postfix, v...)
+}
+
+// Experimental logs the usage of an experimental feature.
+func Experimental(format string, v ...interface{}) {
+	logp.Warn("EXPERIMENTAL: "+format, v...)
+}
--- a/libbeat/common/cfgwarn/removed.go
+++ b/libbeat/common/cfgwarn/removed.go
@@ -1,0 +1,49 @@
+package cfgwarn
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/joeshaw/multierror"
+
+	"github.com/elastic/beats/libbeat/common"
+)
+
+func CheckRemoved5xSettings(cfg *common.Config, settings ...string) error {
+	var errs multierror.Errors
+	for _, setting := range settings {
+		if err := CheckRemoved5xSetting(cfg, setting); err != nil {
+			errs = append(errs, err)
+		}
+	}
+
+	return errs.Err()
+}
+
+// CheckRemoved5xSetting prints a warning if the obsolete setting is used.
+func CheckRemoved5xSetting(cfg *common.Config, setting string) error {
+	segments := strings.Split(setting, ".")
+
+	L := len(segments)
+	name := segments[L-1]
+	path := segments[:L-1]
+
+	current := cfg
+	for _, p := range path {
+		current, _ := current.Child(p, -1)
+		if current == nil {
+			break
+		}
+	}
+
+	// full path to setting not available -> setting not found
+	if current == nil {
+		return nil
+	}
+
+	if !current.HasField(name) {
+		return nil
+	}
+
+	return fmt.Errorf("setting '%v' has been removed", current.PathOf(name))
+}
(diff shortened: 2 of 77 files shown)


libbeat/common/cfgwarn/cfgwarn.go
```diff
@@ -0,0 +1,24 @@
+package cfgwarn
+
+import (
+	"fmt"
+
+	"github.com/elastic/beats/libbeat/logp"
+)
+
+// Beta logs the usage of an beta feature.
+func Beta(format string, v ...interface{}) {
```
ruflin: Seems to me this part would be more belong into logp?

libbeat/common/cfgwarn/removed.go
```diff
@@ -0,0 +1,48 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/elastic/beats/libbeat/common"
+	"github.com/joeshaw/multierror"
+)
+
+func CheckRemoved5xSettings(cfg *common.Config, settings ...string) error {
```
ruflin: Do we need the `5x` in the name? looks liek this is a generic function.

metricbeat/tests/system/test_config.py
```diff
@@ -65,5 +65,23 @@ def test_compare_config(self):
...
+            "metricsets": ["cpu"],
+            "period": "5s",
+        }])
+
+        exit_code = self.run_beat(extra_args=[
+            "-E",
+            "metricbeat.modules.0.filters.0.include_fields='field1,field2'"
+        ])
+        assert exit_code == 1
+        assert self.log_contains("setting 'metricbeat.modules.0.filters'"
```
tsg: Since `filters` is now renamed, can you update the default config files [here](https://github.com/elastic/beats/blob/master/metricbeat/modules.d/system.yml#L23) and [here](https://github.com/elastic/beats/blob/master/metricbeat/modules.d/redis.yml.disabled#L24).

libbeat/common/cfgwarn/cfgwarn.go
tsg: I'm fine with both locations, but I'm also curious why it got moved.

libbeat/common/cfgwarn/removed.go
urso: I named it 5x so we can remove all references to 5.x at once (Just delete the function and compiler will tell you all places you have to modify).

libbeat/common/cfgwarn/cfgwarn.go
urso: I introduced package cfgwarn to combine all configuration related warning messages in. Plus I find logp kind of messy right now.

metricbeat/docs/modules/redis.asciidoc
```diff
@@ -63,9 +63,9 @@ metricbeat.modules:
   #maxconn: 10
 
   # Filters can be used to reduce the number of fields sent.
-  #filters:
+  #processors:
   #  - include_fields:
-  #      fields: ["stats"]
+  #      fields: ["beat", "metricset", "redis.info.stats"]
```
urso: The publisher pipeline right now applies processors after having constructed the complete event. The disadvantage is, when using `include_fields`, one must configure all event fields. Having to configure the full field name to keep is quite an advantage I think, as we also document the full field names in our reference.

Considering a change in the processors order, by applying the `processors` before adding beats fields like `beat` or `metricset`. For metricbeat this would require some refactoring such that each metricbeat has it's own beat.Client instance (right now all metricsets share one beat.Client by forwarding events to another go-routine). Then constant fields like `metricset` can be added by the publisher pipeline by configuring the `ClientConfig.Fields = common.MapStr { "metricset": ... }`.

libbeat/common/cfgwarn/removed.go
ruflin: Works for me for the moment. i think we can use such a method in a more generic way as this is no the only time we remove settings. Something like `CheckRemoveSettings()` and the one here with 5x is just a proxy for it.

metricbeat/docs/modules/redis.asciidoc
ruflin: This is quite a breaking change i think. Meaning if someone use `fields: ["abc"]` before and keeps the config, it will now remove many more fields.

libbeat/common/cfgwarn/removed.go
urso: Yeah, it's not the only time we remove settings. But at some point in time I want to remove the messages (e.g. on 7.x release remove the 5.x checks). Being able to remove messages/warnings  is the motivation for naming this 5x

metricbeat/docs/modules/redis.asciidoc
urso: Right. This is what I'm saying. That's why I'm considering the changes to metricbeat and processors order in the publisher. Maybe we also want to introduce some whitelisting (fields that can not be removed).

PR #4727, opened 2017-07-20
Mark audit metricsets as beta

andrewkroh:
- Change log messages from experimental to beta.

- Make reassembler tuning constants controllable via config and increase the default max_in_flight from 5 to 50. Someone might need to tune these so I'm making them configurable. Will add docs in a separate PR.

- When `include_warnings` is enabled and a warning message is present, the raw messages will now always be added to the event regardless of the `include_raw_message` setting. These raw message will be critical to understanding the root cause of the warning and debugging it after the fact.


andrewkroh:
Fixed it and it passing now.

PR #4731, opened 2017-07-21
Add pattern option in the fields.yml

monicasarbu:
This PR adds the `pattern` option to be used in `fields.yml`. The `pattern` option is used to generate the index pattern with the right pattern for a number field.

cc-ed @ruflin


ruflin:
WFG

PR #4732, opened 2017-07-21
Changes to enable building GS "mini-book"

debadair:
Changed the xrefs from the topics being translated to external links where needed to enable the mini-books to build. Also added directory-layout and setup-repositories attributes for the external links in the shared files.

I'm going to have to make some custom changes in the translated files to address the duplicate IDs, but I don't see a better way around that.


dedemorton:
LGTM

PR #4757, opened 2017-07-26
Adding support to exclude labels from kubernetes pod metadata

vjsamuel:
We have seen instances where there are labels that have too high of a cardinality for which support to exclude/filter out would be good to have on the `add_kubernetes_metadata` processor.


exekias:
jenkins test it please

Issue #4755, opened 2017-07-25
Improve beats build and test time


ruflin:
Building a beat currently takes quite a bit of time because of all the dependencies. The reasons is that Golang rebuilds all packages every time. The `-i` flag can be used to save the packages under `$GOPATH/pkg` so they can be later reused. 

Here the output of a few example commands I used:

```
metricbeat$ rm -rf $GOPATH/pkg

$ time go build

real	0m51.015s
user	1m33.506s
sys	0m17.141s

$ time go build

real	0m47.963s
user	1m32.396s
sys	0m15.702s

$ time go build -i

real	0m49.321s
user	1m32.670s
sys	0m15.620s

$ time go build

real	0m5.614s
user	0m4.511s
sys	0m1.426s

$ time go build

real	0m5.526s
user	0m4.437s
sys	0m1.385s
```

First the `pkg` directory is removed and then `go build` is run twice as an example, then once with `-i` and then again twice with the prebuilt packages. The difference is ~45s which is quite a lot. The `-i` flag also has quite a bit of an effect when running unit tests.

I was thinking we could add this to our Makefile but the part I worry about is how Golang will invalidate existing package files (or not).

andrewkroh:
According to the last post [here](https://groups.google.com/forum/#!topic/golang-nuts/NtZtHx_dr0o) there is logic to determine if the packages need to be rebuilt.

One test would be to modify something in the vendor dir and see if the change is picked up.

tsg:
That's awesome, on my laptop it reduced the time to compile metricbeat from 70s to 5s. In my test it also correctly detected a change in the `vendor/` folder.

I'm +1 to adding it to the makefile and see if we hit any weird conditions from it. 

Kind of surprised we didn't see this before, but the docs are kind of misleading on it:

```
The -i flag installs the packages that are dependencies of the target.
``

ruflin:
I was also confused by the docs. I stumbled over it because I wanted to know why `go test .` sometimes takes very long and started to play with the `-x` flag. Then I discovered the `-i` for tests and so it also for builds. The problem with tests is that you need to run `go test -i .` and then `go test .` as if `-i` is used, tests are not run. Looking at the combined time of the two commands without having `pkg` it is almost identical to only run `go test .`. In case the packages are already build, `go test -i` is very quick. So I'm thinking if also here we should add the flag.

Will open a PR for further tests / discussions.

ruflin:
[...]

PR #4758, opened 2017-07-26
Improve test and build time by using -i flag

ruflin:
See https://github.com/elastic/beats/issues/4755 for details.

### DIFF
--- a/libbeat/scripts/Makefile
+++ b/libbeat/scripts/Makefile
@@ -83,10 +83,11 @@ endif
 
 
 ${BEAT_NAME}: $(GOFILES_ALL) ## @build build the beat application
-	go build
+	go build -i
 
 # Create test coverage binary
 ${BEAT_NAME}.test: $(GOFILES_ALL)
+	@go build -i -o /dev/null
 	@go test $(RACE) -c -coverpkg ${GOPACKAGES_COMMA_SEP}
 
 .PHONY: crosscompile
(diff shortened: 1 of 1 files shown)


ruflin:
Curious to see if this also speeds up tests on Jenkins as libbeat packages are only built once for all beats. On Travis it probably doesn't make a difference as it is a new environment each time.

ruflin:
Strangely the Jenkins build took longer them previous builds.

ruflin:
I just reliased this does not apply when running the `testsuite` because there the gotestcover package is used. I will check if I can use the -i flag there too.

libbeat/scripts/Makefile
```diff
@@ -142,15 +143,18 @@ prepare-tests:
 .PHONY: unit-tests
 unit-tests: ## @testing Runs the unit tests with coverage.  Race is not enabled for unit tests because tests run much slower.
 unit-tests: prepare-tests
+	go test -i ${GOPACKAGES}
```
tsg: From the help of `go test`:

```
	-i
	    Install packages that are dependencies of the test.
	    Do not run the test.
```

And indeed `go test -i` doesn't actually run the tests. However, in master `make unit-tests` seems to work fine, so I'm not quite sure what's going on. But this makes me think adding `-i` to the test commands might not do what we expect.

ruflin: We actually always run `go test -i` followed by `go test`. Just in the case here it is in the line below. Also see line 151 as an example. My tests showed that running `go test -i` followed by `go test` takes the same time together as just running `go test` alone. So why do it? If you run it more then once, `go test -i` is very quick and the run time will be much shorter.

tsg: Gotcha, I missed that we run the `go test` without `-i` afterwards. Thanks!

cyrilleverrier:
Will you merge this improvement to the 5.x branches? it is a very nice improvement.

ruflin:
@cyrilleverrier No plan to backport this as it's not directly a feature. I would also give this some more time to be tested. I kind of still worry about side effects as there must be a reason this is not enabled in Golang by default.

Issue #4737, opened 2017-07-24
Document breaking changes in 6.0


tsg:
From the CHANGELOG:

*Affecting all Beats*

- [x] Rename `kubernetes` processor to `add_kubernetes_metadata`. #4473 <-- `kubernetes` is new anyway, and in beta
- [x] Rename `*.full.yml` config files to `*.reference.yml`. #4563 <-- mention it in the upgrade guide
- [x] The `scripts/import_dashboards` is removed from packages. Use the `setup` command instead. #4586 <-- done
- [x] Change format of the saved kibana dashboards to have a single JSON file for each dashboard #4413 <-- not really important from the user POV
- [x] Rename `configtest` command to `test config`. #4590 <-- `configtest` still works so not really a breaking change? Perhaps mention it in the upgrade guide?
- [x] Remove setting `queue_size` and `bulk_queue_size`. #4650 <-- done
- [x] Remove setting `dashboard.snapshot` and `dashboard.snapshot_url`. They are no longer needed because the
  dashboards are included in the packages by default. #4675 <-- minor
- [x] Introduce beat version in the Elasticsearch index and mapping template #3527 <-- is this breaking in any way?
- [x] Usage of field `_type` is now ignored and hardcoded to `doc`. #3757 <-- done in 5.x already
- [x] Change vendor manager from glide to govendor. #3851 <-- not user facing
- [x] Rename `error` field to `error.message`. #3987 <-- minor
- [x] Change `dashboards.*` config options to `setup.dashboards.*`. #3921 <-- minor, since the main way of importing the dashboards in 5.x was `import_dashboards`
- [x] Change `outputs.elasticsearch.template.* to `setup.template.*` #4080

*Filebeat*

- [x] Rename `input_type` field to `prospector.type` #4294 <-- done
- [x] The `@metadata.type` field, added by the Logstash output, is now hardcoded to `doc` and will be removed in future versions. #4331. <-- minor
- [x] Remove code to convert states from 1.x. #3767 <-- i think only relevant in the sense that we should mention that upgrades have to be done major by major
- [x] Remove deprecated config options `force_close_files` and `close_older`. #3768 <-- minor
- [x] Change `clean_removed` behaviour to also remove states for files which cannot be found anymore under the same name. #3827 <-- minor
- [x] Remove `document_type` config option. Use `fields` instead. #4204 <-- done
- [x] Move `json_error` under `error.message` and `error.key`. #4167 <-- minor

*Heartbeat*

[...]

PR #4760, opened 2017-07-26
Added breaking changes doc for 6.0

tsg:
Part of #4737.

Note that I didn't repeat everything from the release notes, but I just highlighted the things that are more likely to cause issues. See #4737 for short motivations on each item.

@urso Can you review the Spooler removal part, please?

@dedemorton Let me know please if the overall structure makes sense, happy to consider alternatives.

### DIFF
--- a/libbeat/docs/breaking.asciidoc
+++ b/libbeat/docs/breaking.asciidoc
@@ -4,13 +4,8 @@
 As a general rule, we strive to keep backwards compatibility between minor
 versions (e.g.  6.x to 6.y) so you can upgrade without any configuration file
 changes, but there are breaking changes between major versions (e.g. 5.x to
-6.y).
-
-* <<breaking-changes-6.0>>
-* <<breaking-changes-5.0>>
-
-We also list the breaking changes at the top of the <<release-notes,release notes>>
-for each version.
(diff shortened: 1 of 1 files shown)


dedemorton:
The overall structure makes sense. It seems a little odd to have the breaking changes for 5.0 in this document, tho, since we're duplicating what we have in the 5.0 docs.

tsg:
Yeah, it's probably better to remove them. Will do.

libbeat/docs/breaking.asciidoc
```diff
@@ -2,15 +2,148 @@
...
+libbeat (the code shared by all Beats), and the presence of multiple queues
+made the performance tuning of Filebeat more complex than it needed to be.
+
+As a result, the following options are removed:
+
+- `spool_size`
+- `publish_async`
+- `idle_timeout`
+- `queue_size`
+- `bulk_queue_size`
```
urso: naming:
- these settings live in filebeat namespace and are used like this on reference configs: `filebeat.spool_size`, `filebeat.publish_async` and `filebeat.idle_timeout`
- these settings are 'global': `queue_size` and `bulk_max_size`.

PR #4769, opened 2017-07-27
Allow template pattern to be overwritten

ruflin:
With the current template loading is was not possible to overload the pattern used as name and pattern were mixed. For example in the case where someone wanted to define pattern name `a` that applies to the indice `a` this was not possible because the pattern generated was `a-*` so it didn't apply to the index `a`. This change now allows to set pattern and name separately.

To make sure index pattern and templates are in sync, the beat will not start in case the index pattern was modified but the template patterns not. We should figure out later some automated mechanisms here.

We should find a better solution for 6.x to require less config options and also include dashboard generation in this.

### DIFF
--- a/libbeat/_meta/config.reference.yml
+++ b/libbeat/_meta/config.reference.yml
@@ -144,6 +144,7 @@ output.elasticsearch:
 
   # Optional index name. The default is "beatname" plus date
   # and generates [beatname-]YYYY.MM.DD keys.
+  # In case you modify this pattern you must update setup.template.name and setup.template.pattern accordingly.
   #index: "beatname-%{[beat.version]}-%{+yyyy.MM.dd}"
 
   # Optional ingest node pipeline. By default no pipeline will be used.
@@ -601,10 +602,14 @@ output.elasticsearch:
 # Set to false to disable template loading.
 #setup.template.enabled: true
 
-# Template name. By default the template name is beatname.
-# The version of the beat will always be appended to the given name
-# so the final name is beatname-%{[beat.version]}.
-#setup.template.name: "beatname"
+# Template name. By default the template name is "beatname-%{[beat.version]}"
+# The template name and pattern has to be set in case the elasticsearch index pattern is modified.
+#setup.template.name: "beatname-%{[beat.version]}"
+
+# Template patttern. By default the template patter is "-%{[beat.version]}-*" to apply to the default index settings.
+# The first part is the version of the beat and then -* is used to match all daily indicies.
+# The template name and pattern has to be set in case the elasticsearch index pattern is modified.
+#setup.template.pattern: "beatname-%{[beat.version]}-*"
 
 # Path to fields.yml file to generate the template
 #setup.template.fields: "${path.config}/fields.yml"
--- a/libbeat/template/template.go
+++ b/libbeat/template/template.go
@@ -2,8 +2,11 @@ package template
 
 import (
 	"fmt"
+	"time"
 
+	"github.com/elastic/beats/libbeat/beat"
 	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/common/fmtstr"
 	"github.com/elastic/go-ucfg/yaml"
 )
 
@@ -17,15 +20,54 @@ var (
 )
 
 type Template struct {
-	index       string
+	name        string
+	pattern     string
 	beatVersion common.Version
 	esVersion   common.Version
 	settings    TemplateSettings
 }
 
 // New creates a new template instance
-func New(beatVersion string, esVersion string, index string, settings TemplateSettings) (*Template, error) {
+func New(beatVersion string, beatName string, esVersion string, config TemplateConfig) (*Template, error) {
 	bV, err := common.NewVersion(beatVersion)
+	if err != nil {
+		return nil, err
+	}
+
+	name := config.Name
+	if name == "" {
+		name = fmt.Sprintf("%s-%s", beatName, bV.String())
+	}
+
+	pattern := config.Pattern
+	if pattern == "" {
+		pattern = name + "-*"
+	}
+
+	event := &beat.Event{
+		Fields: common.MapStr{
+			"beat": common.MapStr{
+				"name":    beatName,
+				"version": bV.String(),
+			},
+		},
+		Timestamp: time.Now(),
+	}
+
+	nameFormatter, err := fmtstr.CompileEvent(name)
+	if err != nil {
+		return nil, err
+	}
+	name, err = nameFormatter.Run(event)
+	if err != nil {
+		return nil, err
+	}
+
+	patternFormatter, err := fmtstr.CompileEvent(pattern)
+	if err != nil {
+		return nil, err
+	}
+	pattern, err = patternFormatter.Run(event)
 	if err != nil {
 		return nil, err
 	}
@@ -41,10 +83,11 @@ func New(beatVersion string, esVersion string, index string, settings TemplateSe
 	}
 
 	return &Template{
-		index:       index,
+		pattern:     pattern,
+		name:        name,
 		beatVersion: *bV,
 		esVersion:   *esV,
-		settings:    settings,
+		settings:    config.Settings,
 	}, nil
 }
 
@@ -62,9 +105,14 @@ func (t *Template) Load(file string) (common.MapStr, error) {
 	return output, nil
 }
 
-// GetName returns the name of the template which is {index}-{version}
+// GetName returns the name of the template
 func (t *Template) GetName() string {
-	return fmt.Sprintf("%s-%s", t.index, t.beatVersion.String())
+	return t.name
+}
+
+// GetPattern returns the pattern of the template
+func (t *Template) GetPattern() string {
+	return t.pattern
 }
 
 // generate generates the full template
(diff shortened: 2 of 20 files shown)


libbeat/_meta/config.reference.yml
```diff
@@ -610,9 +610,12 @@ output.elasticsearch:
 #setup.template.enabled: true
 
 # Template name. By default the template name is beatname.
-# The version of the beat will always be appended to the given name
-# so the final name is beatname-%{[beat.version]}.
-#setup.template.name: "beatname"
+# The f
```
ruflin: Ignore this doc and below for the moment. Take a look at the PR description on the behaviour.

ruflin:
It would be nice if the template name would be based on the elasticsearch index name. the problem with that for the moment is that the index name includes the daily pattern: `  index: "beatname-%{[beat.version]}-%{+yyyy.MM.dd}"`

We could change this to:

```
elasticsearch
  index: "beatname-%{[beat.version]}"
  pattern: "-%{+yyyy.MM.dd}"
```

This would allow the template loading to take by default the index name as the name for the template. Problem with this it's an additional break in case someone configured a different index-name.

libbeat/_meta/config.reference.yml
```diff
@@ -610,9 +610,12 @@ output.elasticsearch:
...
+#setup.template.name: "beatname-%{[beat.version]}"
+
+# Template patttern. By default the template patter is "-%{[beat.version]}-*" to apply to the default index settings.
+# The first part is the version of the beat and then -* is used to match all daily indicies.
+#setup.template.pattern: "beatname-%{[beat.version]}-*"
```
ruflin: Do we even need the `-*` for the pattern or could we just use `*` so it would also apply do indices which are exactly like the index name?

tsg: Not sure I understand this suggestion. You mean something lie `beatname-%{[beat.version]}*`?

ruflin: Yes

libbeat/template/template.go
```diff
@@ -17,19 +17,30 @@ var (
...
 	}
 
+	name := config.Name
+	if name == "" {
+		name = fmt.Sprintf("%s-%s", beatName, bV.String())
+	}
+
+	pattern := config.Pattern
+	if pattern == "" {
+		pattern = name + "-*"
```
ruflin: Should we remove here the `-` and make this just `*`?

libbeat/beat/beat.go
```diff
@@ -607,7 +608,27 @@ func (b *Beat) registerTemplateLoading() error {
...
 	// Loads template by default if esOutput is enabled
 	if b.Config.Output.Name() == "elasticsearch" {
+		var cfg template.TemplateConfig
+		err := b.Config.Template.Unpack(&cfg)
+		if err != nil {
+			return fmt.Errorf("unpacking template config fails: %v", err)
+		}
+
+		// Get ES Index name for comparison
+		esCfg := struct {
```
ruflin: @urso Should I also check for `indicies` here and `len==0`?

metricbeat/docs/fields.asciidoc
```diff
@@ -2790,28 +2790,28 @@ Bytes in non-idle span.
 
 
 [[exported-fields-graphite]]
-== graphite Fields
+== graphite fields
```
ruflin: @dedemorton Seems like some of your recent changes are revert? I wonder why the previous builds went green?

libbeat/template/template.go
tsg: This is just the default, right? I'd say let's keep the `-*`, so that it's less likely to apply accidentally to other indices.

libbeat/template/template.go
```diff
@@ -64,7 +76,12 @@ func (t *Template) Load(file string) (common.MapStr, error) {
 
 // GetName returns the name of the template which is {index}-{version}
 func (t *Template) GetName() string {
-	return fmt.Sprintf("%s-%s", t.index, t.beatVersion.String())
+	return t.name
+}
+
+// GetPattern returns the name of the template which is {index}-{version}
```
tsg: Comment seems out of date.

libbeat/beat/beat.go
```diff
@@ -607,7 +608,27 @@ func (b *Beat) registerTemplateLoading() error {
...
+			Index string `config:"index"`
+		}{}
+		err = b.Config.Output.Config().Unpack(&esCfg)
+		if err != nil {
+			return err
+		}
+
+		if esCfg.Index != "" && (cfg.Name == "" || cfg.Pattern == "") {
+			return fmt.Errorf("setup.template.name and setup.template.pattern have to be set if index name is modified.")
```
tsg: Does the name really have to be changed? I'm thinking we could only check on the `setup.template.pattern`?

ruflin: It will be kind of strange if the name is not changed. Assuming someone changes the index to `test-{daily}` and the pattern to `test-*` he then sees `beatname-{version}-{daily}` as the template name with `test-*` as pattern. In case that template is already there, it would not be overwritten and he would still not have the expected template.

libbeat/template/template.go
ruflin: Yes 👍

libbeat/template/template.go
ruflin: updated

tsg:
@ruflin I tested this branch, but when using a configuration like this (which is what's suggested in the docs):

```
setup.template:
  name: test
  pattern: "metricbeat-%{[beat.version]}-*"
```

I get this in the resulting template:

```
{
  "test": {
    "order": 1,
    "index_patterns": [
      "metricbeat-%{[beat.version]}-*"
    ],
```

So it seems like the variables are not resolved as I was expecting.

urso:
I don't see any use of format strings being applied. It's all strings only. When/how does `beat.version` get resolved? How do you handle event fields being used, that are specific to a module/prospector? e.g. `%{[fields.index]}-%{+yyyy.MM.dd}`? In this case you have to log an error and ask the user to explicitly configure the template pattern (if possible).

ruflin:
Currently the default works when not set because it is set internal. But as @urso said there is currently no handling of the special strings. My suggestion is to support in this version everything under `beat.*` as that is static but not support dynamic fields from event (I would not know how we could do this). Like this the defaults also work when they are written out in the config and it should cover most of the case. We should open later a separate issue on how and if we want to support the more complex cases with `fields.index` for example.

libbeat/template/template.go
```diff
@@ -41,6 +45,39 @@ func New(beatVersion string, beatName string, esVersion string, config TemplateC
...
 	}
 
+	compiler := func(key string, ops []fmtstr.VariableOp) (fmtstr.FormatEvaler, error) {
+		if key == "[beat.name]" {
+			return Formatter{s: beatName}, nil
+		}
+
+		if key == "[beat.version]" {
+			return Formatter{s: bV.String()}, nil
+		}
```
urso: Date patterns start with `+` character. See: https://github.com/elastic/beats/blob/master/libbeat/common/fmtstr/formatevents.go#L252

That is, we can replace `%{+<pattern>}` with `*` in here.

libbeat/template/template.go
```diff
@@ -167,3 +204,19 @@ func loadYaml(path string) (Fields, error) {
 	}
 	return fields, nil
 }
+
+type Formatter struct {
```
urso: don't leak/export this type if you don't need it. Also `template.Formatter` is kind of a bad choice here.

libbeat/template/template.go
```diff
@@ -41,6 +45,39 @@ func New(beatVersion string, beatName string, esVersion string, config TemplateC
 		pattern = name + "-*"
 	}
 
+	compiler := func(key string, ops []fmtstr.VariableOp) (fmtstr.FormatEvaler, error) {
```
urso: Any plan to handle `ops`? Event format strings can have defaults if the key is missing in the event. See: https://github.com/elastic/beats/blob/master/libbeat/common/fmtstr/formatevents.go#L280

One can use `%{[my.field]:defaultValue}` in event format strings.

ruflin: For now to keep it as simple as possible I suggest to return an error in case any ops are used.

libbeat/template/template.go
ruflin: Is there an easy way to reuse this pattern compiling?

libbeat/template/template.go
```diff
@@ -41,6 +45,39 @@ func New(beatVersion string, beatName string, esVersion string, config TemplateC
...
+		if key == "[beat.name]" {
+			return Formatter{s: beatName}, nil
+		}
+
+		if key == "[beat.version]" {
+			return Formatter{s: bV.String()}, nil
+		}
+
+		logp.Debug("template", "Lookup key for pattern not available: %s", key)
+		return nil, nil
```
ruflin: yes, will return an error if there is a pattern but doesn't match either of the two.

libbeat/template/template.go
ruflin: agree, need to come up with a better name

libbeat/template/template.go
```diff
@@ -17,19 +20,58 @@ var (
...
+		Fields: common.MapStr{
+			"beat": common.MapStr{
+				"name":    beatName,
+				"version": bV.String(),
+			},
+		},
+		Timestamp: time.Now(),
+	}
+
+	nameFormatter, err := fmtstr.CompileEvent(name)
```
ruflin: @urso I changed by approach here. Instead of building my own compiler and formatter, I just create an event object and use `CompileEvent`. This makes things much easier.

ruflin:
@tsg PR is updated know and now `dates`, `beat.version` and `beat.name` should work for the patterns. As I'm now using the EventCompiler, we could also later add things like `fields` which are constant to it.

ruflin:
@tsg It seems this breaks the module tests. I must investigate why but let me know if you have a quick idea.

ruflin:
jenkins, test it

ruflin:
Rebased and squashed again.

ruflin:
jenkins, test it

libbeat/template/template.go
ruflin: It will work. I think this is a good thing or do you see an issue? 

This implementation allows us to add more fields later on and still use the same implementation.

ruflin:
jenkins, retest it

dedemorton:
Removing needs_docs labels because template.name and template.pattern options are now documented. If additional doc work is required, please open an issue against the docs.

PR #4770, opened 2017-07-27
Add http server metricset to support push metrics via http

vjsamuel:
This PR allows users to configure a http server in order to send data. The restriction right now it only supports JSON.

Path parameters can be parsed into tags. Ex:
/foo/bar would result in 
```
tag:  {
  foo: "bar",
}
```


metricbeat/module/http/_meta/config.yml
```diff
@@ -8,3 +8,14 @@
...
+
+- module: http
+  metricsets: ["server"]
+  host: "localhost"
+  port: "8080"
+  enabled: false
+#  paths:
+#    - path: "/foo"
+#      namespace: "foo"
+#      tags:
```
exekias: I think it would make sense to change this behavior, use `fields` instead of tags, and put given fields in the final event, overriding any existing one

CHANGELOG.asciidoc
```diff
@@ -52,7 +52,7 @@ https://github.com/elastic/beats/compare/v6.0.0-beta1...master[Check the HEAD di
 
 - Add support to exclude labels from kubernetes pod metadata. {pull}4757[4757]
 - Add graphite protocol metricbeat module. {pull}4734[4734]
-
+- Add http server metricset to support push metrics via http. {pull}4770[4770]
```
exekias: Missing newline after this

metricbeat/docs/fields.asciidoc
```diff
@@ -3874,6 +3874,21 @@ The HTTP payload received
...
 
+[float]
+== server Fields
+
+server
+
+
+
+[float]
+=== http.server.example
```
ruflin: We probably should not have "example" fields in the docs.

exekias:
ok to test

exekias:
You may want to take a look to travis tests, the fail looks legit

metricbeat/docs/fields.asciidoc
vjsamuel: i have removed this field from the asciidoc. during restart/multiple definitions we would error out with a critical message saying port is already in use.

exekias:
jenkins retest this

PR #4772, opened 2017-07-27
Fix mixed up modules configuration

### LINKED ISSUE #4761: Filebeat: Mixed up fileset.module
- Version: 
`Filebeat 5.6.0-SNAPSHOT` from `Jul 24 13:17`
 
- Operating System:  
`Linux ip-10-0-1-124 4.4.35-33.55.amzn1.x86_64 #1 SMP Tue Dec 6 20:30:04 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux`
2 node cluster.

- Steps to Reproduce: 
I'd tell you if I knew ;)

Whilst trying several different commands using filebeat modules, I noticed that apache2 data had been uploaded with an nginx label.

image

Commands used were probably a combination of:
`./filebeat -v -e -modules=apache2,nginx`
`./filebeat -v -e -modules=apache2`
`./filebeat -v -e -modules=nginx`

Ingest pipelines were reviewed and appeared to be ok.

Will wipe down kit, and attempt to replicate.

tsg:
Fixes #4761. Due to combining pointer and reference passing we ended up with
passing the same module pointer to multiple filesets. The pointer was correct
during initialization, but wrong during run time.

Also added an Info with the enabled modules / filesets.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -30,6 +30,8 @@ https://github.com/elastic/beats/compare/v6.0.0-beta1...master[Check the HEAD di
 
 *Filebeat*
 
+- Fix issue where the `fileset.module` could have the wrong value. {issue}4761[4761]
+
 *Heartbeat*
 
 *Metricbeat*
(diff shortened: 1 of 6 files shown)


tsg:
jenkins, test this again

PR #4773, opened 2017-07-27
Fix batch being ACKed on failure in ES monitoring reporter

urso:


monicasarbu:
Thanks @urso for the quick fix.

PR #4775, opened 2017-07-27
Fixing pod name indexer to use both namespace, pod name to frame index key

vjsamuel:
Today we use only pod name in the index key for pod name indexer. With the advent of stateful sets, users can now spin up pods with consistent names like `zookeeper-1` on multiple namespaces which can cause collision very easily. Hence it is mandatory for using `namespace/podname` as the key for pod name indexer.

### DIFF
--- a/libbeat/processors/add_kubernetes_metadata/indexing_test.go
+++ b/libbeat/processors/add_kubernetes_metadata/indexing_test.go
@@ -256,3 +256,49 @@ func TestFilteredGenMetaExclusion(t *testing.T) {
 	ok, _ = labelMap.HasKey("x")
 	assert.Equal(t, ok, false)
 }
+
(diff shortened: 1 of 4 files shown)


libbeat/processors/add_kubernetes_metadata/indexing_test.go
```diff
@@ -5,6 +5,8 @@ import (
 
 	"github.com/stretchr/testify/assert"
 
+	"fmt"
```
exekias: could you please move this one next to "testing"?

vjsamuel: sorry i keep missing this always. done.

exekias:
jenkins test it please

exekias:
jenkins test it please

Issue #4567, opened 2017-06-27
Filebeat system module startup instructions and dashboard name


tbragin:
I just tried to start a Filebeat system module on the Mac the first time (5.5 BC6 build) using 5.5 docs: https://www.elastic.co/guide/en/beats/filebeat/5.5/filebeat-module-system.html

A couple of snags:
- We refer to the "syslog" module, whereas the module name is now "syslog"
- More worryingly, in the product itself the Filebeat system dashboard is still called the "syslog" dashboard (cc @monicasarbu ), which is confusing IMO
- Docs don't mention that `ingest-geoip` plugin in ES is required (the way they do in Apache2), but on startup I got an error when that wasn't installed
- When I tried to run `sudo ./filebeat -e -modules=system -setup`, I got file ownership errors around -- not sure if this was because I was using the BC... running without `-setup` worked fine:
`{filebeat}/module/system/auth/manifest.yml`
`{filebeat}/module/system/syslog/manifest.yml`

tbragin:
cc @brandonmensing @dedemorton

dedemorton:
@tbragin For the file permission issue, see this topic: 

https://www.elastic.co/guide/en/beats/libbeat/5.5/config-file-permissions.html

I'll need to add a link to that topic from the modules documentation.

tbragin:
@dedemorton i don't see anything in that page regarding permissions for manifest.yml files.

dedemorton:
@tbragin The same issue applies to permissions on the manifest.yml. We just need to update the docs to reflect that.

tbragin:
@dedemorton Yes, agreed

tsg:
> When I tried to run sudo ./filebeat -e -modules=system -setup, I got file ownership errors around -- not sure if this was because I was using the BC or because i'm starting up the module using "sudo":

This worries me, we keep getting into issues with this. I'm thinking to opt out of the permissions check for the modules `manifest.yml` file, since they don't contain credentials and we don't reload them on changes. @andrewkroh, WDYT, any risk in that?

> I didn't see any data in the Last 15 minutes, so I thought things were broken. Brandon informed me that I may need to change range to Last 24 Hours to see system data, and indeed I see data from now (Tue Jun 27 10:10:27 PDT 2017) recorded as 2017/06/27 17:10:26.884147. This behaviour is different from Metricbdeat, is that expected? If so, should that time range just be saved with the dashboard by default?

[...]

PR #4778, opened 2017-07-27
Update info about loading dashboards and index templates

dedemorton:
I plan to make additional changes to the getting started flow about setting up the dashboards because the current flow is not ideal, but I'll make that in a separate PR.

UPDATE: Changed my mind. Didn't want to have two separate PRs with a mix of changes, so I'll put everything in this PR.


auditbeat/docs/faq-ulimit.asciidoc
```diff
@@ -0,0 +1,29 @@
+[float]
+[[ulimit]]
+=== {beatname_uc} fails to watch folders because too many files are open?
+
+Because of the way file monitoring is implemented on Mac OS, you may see a
```
andrewkroh: Apple refers to it as "macOS". https://www.apple.com/macos/sierra/

dedemorton: good catch

dedemorton:
@andrewkroh I'm working on another set of changes related to dasbhboard loading. There's quite a bit of overlap with this PR, so I'm going to take the changes in my other branch and cherry-pick them into this branch. I'll mark this PR as "in progress" and continue adding commits to it. I'm going to have to make changes to multiple beats to avoid breaking things, so I'll change the name of this PR, too.

libbeat/docs/shared-template-load.asciidoc
```diff
@@ -37,42 +37,60 @@ For more information, see:
...
+ifeval::["{requires-sudo}"=="yes"]
+
+If you changed ownership of the config file to root, you'll need preface this
+command with `sudo`.
+
+endif::[]
+
+//REVIEWERS: I'm keeping separate steps for the loading the index template and the dashboards since it seems like the index template loading happens automatically, but the dashboard loading must be specified. It might be confusing to describe this in one step. Plus having separate steps helps users understand that two very different operations are taking place (the loading of the index template into ES and the loading of dashboards into Kibana)
+
+//TODO: I think this needs to be deleted. Commenting it out for now just in case. Also need to remove conditional flag
```
dedemorton: Can someone confirm that there is no reason to keep the content commented out here?

libbeat/docs/step-configure-kibana-endpoint.asciidoc
```diff
@@ -0,0 +1,20 @@
...
+--
+Where `host` is the hostname and port of the machine where Kibana is running,
+for example, `localhost:5601`.
+
+NOTE: If you specify a path after the port number, you need to include
+the scheme and port: `http://localhost:5601/path`.
+
+Also see the security-related options in <<setup-kibana-endpoint>> if you are
+connecting to a secured instance of Kibana.
+
```
monicasarbu: There are a few more options for configuring Kibana, the same options as for configuring Elasticsearch output.

```
# Optional protocol and basic auth crendentials.
  #protocol: "https"
  #username: "elastic"
  #password: "changeme"

  # Optional HTTP Path
  #path: ""

  # Use SSL settings for HTTPS. Default is true.
  #ssl.enabled: true

  # Configure SSL verification mode. If `none` is configured, all server hosts
  # and certificates will be accepted. In this mode, SSL based connections are
  # susceptible to man-in-the-middle attacks. Use only for testing. Default is
  # `full`.
  #ssl.verification_mode: full

  # List of supported/valid TLS versions. By default all TLS versions 1.0 up to
  # 1.2 are enabled.
  #ssl.supported_protocols: [TLSv1.0, TLSv1.1, TLSv1.2]

  # SSL configuration. By default is off.
  # List of root certificates for HTTPS server verifications
  #ssl.certificate_authorities: ["/etc/pki/root/ca.pem"]

  # Certificate for SSL client authentication
  #ssl.certificate: "/etc/pki/client/cert.pem"

  # Client Certificate Key
  #ssl.key: "/etc/pki/client/cert.key"

  # Optional passphrase for decrypting the Certificate Key.
  #ssl.key_passphrase: ''

  # Configure cipher suites to be used for SSL connections
  #ssl.cipher_suites: []

  # Configure curve types for ECDHE based cipher suites
  #ssl.curve_types: []
```

libbeat/docs/step-configure-kibana-endpoint.asciidoc
```diff
@@ -0,0 +1,20 @@
...
+configure the Kibana endpoint:
++
+[source,yaml]
+----------------------------------------------------------------------
+setup.kibana:
+  host: "localhost:5601"
+----------------------------------------------------------------------
++
+--
+Where `host` is the hostname and port of the machine where Kibana is running,
```
monicasarbu: The Kibana configuration is the same as the Elasticsearch output configuration (https://www.elastic.co/guide/en/beats/filebeat/6.x/elasticsearch-output.html#hosts-option).
The `host` can be an `URL` or `IP:PORT`.  For example: http://192.15.3.2, 192:15.3.2:5601, http://192.15.3.2:5601/path. If no port is specified, then 5601 is used.

Note: 
When a node is defined as an IP:PORT, the scheme and path are taken from the `protocol` and `path` config options.

monicasarbu: **protocol**
The default for `protocol` is `http`. However, if you specify a URL for `host`, the value of `protocol` is overridden by whatever scheme you specify in the URL

libbeat/docs/shared-template-load.asciidoc
monicasarbu: It's perfect the way it's described.

libbeat/docs/template-config.asciidoc
```diff
@@ -8,19 +8,22 @@ Elasticsearch.
...
+*`setup.template.fields`*:: The path to the YAML file describing the fields. The default is +fields.yml+. If a
 relative path is set, it is considered relative to the config path. See the <<directory-layout>>
 section for details.
 
-*`overwrite`*:: A boolean that specifies whether to overwrite the existing template. The default
+*`setup.template.overwrite`*:: A boolean that specifies whether to overwrite the existing template. The default
 is false.
 
-*`settings.index`*:: A dictionary of settings to place into the `settings.index` dictionary of the
+//REVIEWERS: I don't see the following options in the reference.yml file. Are they still available/valid?
```
monicasarbu: The `setup.template.settings` shows up only in the metricbeat.yml, not in the reference.yml. I will fix it.

monicasarbu: Good catch @dedemorton. Open a PR: https://github.com/elastic/beats/pull/4791

metricbeat/docs/gettingstarted.asciidoc
```diff
@@ -267,13 +280,17 @@ docker run {dockerimage}
...
 ----------------------------------------------------------------------
 <1> You'll be running Metricbeat as root, so you need to change ownership
-of the configuration file (see
+of the configuration file and any configurations enabled in the `modules.d`
+directory (see
 {libbeat}/config-file-permissions.html[Config File Ownership and Permissions]
 in the _Beats Platform Reference_).
 
+//REVIEWERS: It seems really inconvenient to users to expect them to change ownership on all the files in `modules.d`. This is something I had to do, but not 100% sure if there is a better way. Please advise.
+
```
monicasarbu: @exekias can have a better input here.

packetbeat/docs/gettingstarted.asciidoc
```diff
@@ -289,7 +297,7 @@ docker run {dockerimage}
 [source,shell]
 ----------------------------------------------------------------------
 sudo chown root packetbeat.yml <1>
-sudo ./packetbeat -e -c packetbeat.yml -d "publish"
+sudo ./packetbeat run -e -c packetbeat.yml -d "publish"
```
monicasarbu: `run` is the default option, in case it's missing. So, it still works:
```
sudo ./packetbeat -e -c packetbeat.yml -d "publish"
```

libbeat/docs/shared-kibana-config.asciidoc
```diff
@@ -0,0 +1,213 @@
...
+setup.kibana.host: "localhost:5601"
+----
+
+[float]
+=== Configuration options
+
+You can specify the following options in the `setup.kibana` section of the
++{beatname_lc}.yml+ config file:
+
+//REVIEWERS: I've decided to use full paths when referring to options rather than using indentation in examples. WDYT? Makes the examples a tad harder to read but will help immensely with users who don't know where to put stuff in the yaml hierarchy.
```
monicasarbu: Good idea. This is easier for the users to copy & paste.

libbeat/docs/shared-kibana-config.asciidoc
```diff
@@ -0,0 +1,213 @@
...
+
+[float]
+==== `setup.kibana.host`
+
+The Kibana host where the dashboards will be loaded. This setting is required.
```
monicasarbu: By default is 127.0.0.1:5601.

libbeat/docs/shared-kibana-config.asciidoc
```diff
@@ -0,0 +1,213 @@
...
+The scheme and port default to `http` and `5601` if you don't set them
+explicitly. If you use a path after the port number, you must specify
+the scheme and port. For example: `http://localhost:5601/path`.
```
monicasarbu: The Kibana configuration is the same as the Elasticsearch output configuration (https://www.elastic.co/guide/en/beats/filebeat/6.x/elasticsearch-output.html#hosts-option).
The `host` can be an `URL` or `IP:PORT`.  For example: http://192.15.3.2, 192:15.3.2:5601, http://192.15.3.2:6701/path. If no port is specified, then 5601 is used.

Note: When a node is defined as an IP:PORT, the scheme and path are taken from the `protocol` and `path` config options.

libbeat/docs/shared-kibana-config.asciidoc
```diff
@@ -0,0 +1,213 @@
...
+`https://[2001:db8::1]:5601`.
+
+[float]
+==== `setup.kibana.protocol`
+
+The name of the protocol Kibana is reachable on. The options are: `http` or
+`https`. The default is `http`. However, if you specify a URL for host, the
+value of `protocol` is overridden by whatever scheme you specify in the URL.
+
+//REVIEWERS: Please check this example. I have not tested it.
```
monicasarbu: That's right.

libbeat/docs/shared-kibana-config.asciidoc
```diff
@@ -0,0 +1,213 @@
...
+An HTTP path prefix that is prepended to the HTTP API calls. This is useful for
+the cases where Kibana listens behind an HTTP reverse proxy that exports the API
+under a custom prefix.
+
+[float]
+==== `setup.kibana.ssl.enabled`
+
+Enables {beatname_uc} to use SSL settings when connecting to Kibana via HTTPS.
+
+//REVIEWERS: The reference config file said that the ssl.enabled: true is the default, but that doesn't seem quite right. Do you mean that this setting defaults to true when https is configured?
```
monicasarbu: If the user configures `https` then the default ssl values are used.

libbeat/docs/shared-kibana-config.asciidoc
```diff
@@ -0,0 +1,213 @@
...
+
+[source,yaml]
+----
+setup.kibana.host: "192.0.2.255:5601"
+setup.kibana.protocol: "https"
+setup.kibana.ssl.enabled: true
+setup.kibana.ssl.certificate_authorities: ["/etc/pki/root/ca.pem"]
+setup.kibana.ssl.certificate: "/etc/pki/client/cert.pem"
+setup.kibana.ssl.key: "/etc/pki/client/cert.key
+----
```
monicasarbu: seems ok to me

libbeat/docs/shared-kibana-config.asciidoc
```diff
@@ -0,0 +1,213 @@
...
+
+[float]
+==== `setup.kibana.path`
+
+An HTTP path prefix that is prepended to the HTTP API calls. This is useful for
+the cases where Kibana listens behind an HTTP reverse proxy that exports the API
+under a custom prefix.
+
+[float]
+==== `setup.kibana.ssl.enabled`
```
monicasarbu: I am not sure here, if you copied the content of the `ssl` description. I would say to link it to: https://www.elastic.co/guide/en/beats/filebeat/6.x/configuration-output-ssl.html. In case it changes, we can change it in a single place.

libbeat/docs/shared-kibana-config.asciidoc
```diff
@@ -0,0 +1,213 @@
...
+setup.kibana.ssl.enabled: true
+setup.kibana.ssl.certificate_authorities: ["/etc/pki/root/ca.pem"]
+setup.kibana.ssl.certificate: "/etc/pki/client/cert.pem"
+setup.kibana.ssl.key: "/etc/pki/client/cert.key
+----
+
+[float]
+==== `setup.kibana.ssl.verification_mode`
+
+//REVEIWERS: Most of this content is copied from the section about configuring outputs. We should probably centralize our descriptions of these options at some point, but copy/paste is faster, so I'm doing that for now
```
monicasarbu: Yes, ideally would be not to copy it from here: https://www.elastic.co/guide/en/beats/filebeat/6.x/configuration-output-ssl.html. In case something changes, we can change it in a single place.

dedemorton: @monicasarbu If the descriptions truly are identical, then I should just point to the section. I need to:
1) bump the topic about setting SSL options up one level in the nav so it's not nested under Configuring the output
2) Add an explanation of when the SSL options are valid (ES output config and Kibana endpoint config). 
3) Rename the topic ID so it's not specific to configuring the output

monicasarbu: Actually, the same ssl configuration is used also by logstash output, kafka output, so it's worth having a single copy and only point to it when needed.

libbeat/docs/shared-kibana-config.asciidoc
dedemorton: fixed

metricbeat/docs/gettingstarted.asciidoc
dedemorton: I noticed that Tudor also created an issue about this, so I think it is required.

packetbeat/docs/gettingstarted.asciidoc
dedemorton: @monicasarbu Right, but I figured we added run for a reason and might want to show it. Do you think it's better to show the command without including `run`?

metricbeat/docs/gettingstarted.asciidoc
exekias: I agree we may need to think about something here, we have `strict.perms` flag to avoid this (with the cost of losing some security). deb/rpm packages also ave these permissions right by default. Not sure we can help on the tar.gz case

monicasarbu:
Excellent job @dedemorton!

metricbeat/docs/gettingstarted.asciidoc
andrewkroh: I think using the `strict.perms` is fine here. I'd explain why it's used and link over to docs for the option.

I made this [comment](https://github.com/elastic/beats/issues/4567#issuecomment-311525217) previously, and I still think we should consider it.

> Another option would be to make `-strict.perms=false` the new default. Then modify the scripts that start Beats as services to enable the checks. I liken this to how ES disables bootstrap checks for non-production mode.

filebeat/docs/getting-started.asciidoc
```diff
@@ -218,15 +218,23 @@ include::../../libbeat/docs/shared-logstash-config.asciidoc[]
...
-=== Step 5: Start Filebeat
+=== Step 6: Start Filebeat
 
 Start Filebeat by issuing the appropriate command for your platform.
 
 NOTE: If you use an init.d script to start Filebeat on deb or rpm, you can't
 specify command line flags (see <<command-line-options>>). To specify flags,
 start Filebeat in the foreground.
 
+//REVIEWERS: With the new command syntax, should the deb and rpm instructions here say run instead of start?
```
andrewkroh: No, for deb/rpm these commands remain the same. However for different reason we should change the commands to `sudo service filebeat start`. This ensures that service is started using a clean environment.

dedemorton:
@andrewkroh I think I accidentally blew away your comment about using `-strict.perms=false`. So are you suggesting that I add this flag to all the commands

dedemorton:
Sorry didn't mean to close the PR. Meant to cancel a comment that was in progress, but managed to close the PR and submit the comment instead. :-)

andrewkroh:
@dedemorton I saw this in a GH email but can't find the comment here.

> @andrewkroh I think I accidentally blew away your comment about using -strict.perms=false. So are you suggesting that I add this flag to all the commands

If the packaging was tar.gz and the command uses `sudo` then I would add that flag rather than doing a `chown`.

dedemorton:
> @dedemorton I saw this in a GH email but can't find the comment here.

Yah, sorry to waste your time. I meant to delete the comment, not submit it. I was going to make the change and then ask you to review it. I thought about the flag a little more. Since we recommend that users not use the flag, I think it's actually better to show the file ownership change and then modify the note to indicate that the users can run the beat with `-strict.perms=false`. That way someone won't say "but the example in the Getting Started said to use the flag." I just made the changes and will push them so you can review.

metricbeat/docs/gettingstarted.asciidoc
dedemorton: @andrewkroh Ok, I've added the note about `-strict.perms`. I know we're talking about the getting started experience here, but I sort of worry if we start recommending something in the GS that could get people into trouble later on. See how this works. If the developers on the team decide that we want to tell users straight out to run the the beats with `-strict.perms=false` then I can change it, but the way it's written now users at least understand that there's a choice to be made...and might do more reading before they run the command blindly.

dedemorton:
@monicasarbu I think this should be close to being ready to merge, no?

monicasarbu:
Yes, it's ready to be merged.

metricbeat/docs/gettingstarted.asciidoc
tsg: I'd be happy to make `-strict.perms=false` the default in code when we detect "dev" mode or "getting started". The question is, what's a good sign of production? Is the fact you are running from init script / service file enough? Is deb/rpm production and the tar.gz getting started?

PR #4781, opened 2017-07-28
Set docs version to beta1

tsg:
This one should be merged a bit before the release.


exekias:
I think you will need to update libbeat/_meta/kibana/default/index-pattern/libbeat.json

tsg:
Done, although I'm not sure why that would depend on the asciidoc, that seems wrong. But I think it's only used in tests anyway.

PR #4793, opened 2017-07-31
Increase api version for Beats monitoring

monicasarbu:
This PR contains two changes:
- increase the version of the API for Beats monitoring
- add the response body with the error message returned by Elasticsearch, in case of an error

### DIFF
--- a/libbeat/outputs/elasticsearch/client.go
+++ b/libbeat/outputs/elasticsearch/client.go
@@ -719,19 +719,18 @@ func (conn *Connection) execHTTPRequest(req *http.Request) (int, []byte, error) 
 	}
 	defer closing(resp.Body)
 
-	var retErr error
 	status := resp.StatusCode
 	obj, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
-		return status, nil, retErr
+		return status, nil, err
 	}
 
 	if status >= 300 {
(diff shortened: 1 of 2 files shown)


tsg:
@monicasarbu, can you rebase this one, please? That should fix the intake job on Jenkins.

monicasarbu:
done @tsg

libbeat/outputs/elasticsearch/client.go
```diff
@@ -719,16 +719,18 @@ func (conn *Connection) execHTTPRequest(req *http.Request) (int, []byte, error)
...
 	if err != nil {
 		return status, nil, retErr
 	}
+
+	if status >= 300 {
+		// add the response body with the error returned by Elasticsearch
+		retErr = fmt.Errorf("%v: %s", resp.Status, obj)
+	}
+
 	return status, obj, retErr
```
urso: how about:

```
status := resp.StatusCode
obj, err := ioutil.ReadAll(resp.Body)
if err != nil {
  return status, nil, err
}

if status >= 300 {
  err = fmt.Errorf("%v: %s", resp.Status, obj)
}

return status, obj, err
```

monicasarbu: @urso Done, thanks for the review.

PR #4794, opened 2017-07-31
Add breaking change message about using Beats from Windows GUI

exekias:
Shell is enforced by Cobra: https://github.com/golang-mirrors/cobra/blob/master/command_win.go


exekias:
Removing needs_backport tag as it's done

PR #4796, opened 2017-07-31
Enhancements to auditbeat file metricset

andrewkroh:
- Modify `file.paths` to be a list instead of a dictionary.
- Added `file.hash_types` config option for controlling the hash types.
- Added new hash types.
- Fixed `file.max_file_size` config option. It was missing the "file." prefix when reading the option.
- Added the ability to specify byte unit suffixes to `file.max_file_size`.
- Added documentation for the audit file metricset.


andrewkroh:
I rebased it.

PR #4797, opened 2017-07-31
Update init scripts to use "test config" subcommand

### LINKED ISSUE #4600: Change init scripts to use the configtest command
Because the `-configtest` flag is deprecated.

andrewkroh:
Using "-configtest" is deprecated and the "[beat] test config" subcommand should be used instead.

Fixes #4600


tsg:
jenkins, package it

PR #4798, opened 2017-07-31
Temporary workaround Travis failing Metricbeat builds

tsg:
The Metricbeat tests are sometimes failing because no output is generated for 10m. This is trying a [workaround](https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received) that sort of disables the 10m check, so it's hopefully temporary :).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -115,7 +115,7 @@ before_install:
 install: true
 
 script:
-  - make $TARGETS
+  - travis_wait 60 make $TARGETS
 
 notifications:
   slack:


tsg:
Looks like the `travis_wait` program is working: `Still running (11 of 60): make -C metricbeat testsuite` but got stuck at minute 11. This probably means that the load created by starting the containers all of the same time is simply too big for Travis to handle.

tsg:
Closing as unsuccessful.

PR #4801, opened 2017-08-01
Metricbeat tests: split proxy_dep in 3

tsg:
This is just a test, my hope is that by splitting the dependences in three
groups, docker-compose will start the three groups in sequence, reducing the maximum
load.

The idea is to split the required containers in three groups (`proxy_dep1`, `proxy_dep2`, and `proxy_dep3`). We make all services in `proxy_dep2` depend on `proxy_dep1`, and all services in `proxy_dep3` depend on `proxy_dep2`. The effect is that the containers are started in 3 batches, rather than one big one. This seems to help on my computer.

### DIFF
--- a/metricbeat/docker-compose.yml
+++ b/metricbeat/docker-compose.yml
@@ -55,6 +55,7 @@ services:
       haproxy:        { condition: service_healthy }
       http:           { condition: service_healthy }
 
+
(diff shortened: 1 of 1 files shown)


tsg:
jenkins, test it

PR #4821, opened 2017-08-02
Combine libbeat/beat and libbeat/publisher/beat package + move all beat instance handling to libbeat/cmd package

urso:
This PR moves some types and functionality to `libbeat/beat` and `libbeat/cmd`. The `libbeat/beat` package only defines some common/important types and interfaces to be used by actual beats, without providing real implementations. With these changes, the `publisher/libbeat/beat` finally could be fully merged into `libbeat/beat` For example `beat.Client`, `beat.Pipeline`, `beat.Info`, .... Setting up a beats instance is moved to `libbeat/cmd/instance` package (due to lack of a better name), as the different commands implemented by `libbeat/cmd` do rely on a shared view of an actual configured  instance. The move was also required to resolve quite a number of circular imports, as the `instance` package basically needs to import all other libbeat packages (directly/indirectly) in order to setup a working beat.
Setting up and running a beat instance is still somewhat complicated. Did keep me quite busy in hunting chicken and eggs. Let's see if we can refine/cleanup this somewhat more in later PRs. 

- Move actual beat instance with run loop and configurations to libbeat/cmd/instance 
- Move all definitions from `libbeat/publisher/beat` to `libbeat/beat`
- As `libbeat/beat` defines shared type/interfaces between libbeat and the actual beat => reduce public interface (exports) in `libbeat/beat` to fields/types really being used/shared 
- Remove `packetbeat -device` in favor of `packetbeat device`. Removes some clunky hack we used to have for supporting the `-device` flag.
- Move `common.BeatInfo` to `beat.Info`


filebeat/prospector/udp/harvester.go
```diff
@@ -4,11 +4,12 @@ import (
...
 	"time"
 
-	"github.com/elastic/beats/filebeat/harvester"
-	"github.com/elastic/beats/filebeat/util"
+	"github.com/elastic/beats/libbeat/beat"
 	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/logp"
-	"github.com/elastic/beats/libbeat/publisher/beat"
+
+	"github.com/elastic/beats/filebeat/harvester"
```
ruflin: do you started to separate `filebeat` from `libbeat` imports on purpose?

libbeat/cmd/instance/beat_test.go
```diff
@@ -1,6 +1,6 @@
 // +build !integration
 
-package beat
+package instance
```
ruflin: If we have an `instance` package, it would kind of make sense to have this under the `beat` package?

packetbeat/beater/devices.go
```diff
@@ -1,39 +0,0 @@
-package beater
```
ruflin: I like this change but not sure how it is related to the rest of the changes?

ruflin:
In general I like the changes you made. I'm not a big fan of the `cmd` package but I don't have a better suggestion at the moment.

It seems some other cleanups / improvements sneaked into this PR like Winlogbeat Settings or packetbeat device removal. As this is a major change in how beats are setup and will break all community beats I would like to see the unrelated changes in other PR's if possible. This will make it much easier later to follow which things belonged together. For example for the device removal the changelog will link to the PR that only did this.

libbeat/beat/beat.go
```diff
@@ -1,81 +1,13 @@
...
-	_ "github.com/elastic/beats/libbeat/processors/add_docker_metadata"
-	_ "github.com/elastic/beats/libbeat/processors/add_kubernetes_metadata"
-	_ "github.com/elastic/beats/libbeat/processors/add_locale"
-
-	// Register default monitoring reporting
-	_ "github.com/elastic/beats/libbeat/monitoring/report/elasticsearch"
 )
 
+// Creator initializes and configures a new Beater instance used to execute
+// the beat its run-loop.
```
andrewkroh: `s/beat its/beat's/`

libbeat/beat/beat.go
```diff
@@ -97,564 +29,27 @@ type Beater interface {
...
-	RawConfig *common.Config  // Raw config that can be unpacked to get Beat specific config data.
-	Config    BeatConfig      // Common Beat configuration data.
-	Publisher beat.Pipeline   // Publisher pipeline
+	Info      Info     // beat metadata.
+	Publisher Pipeline // Publisher pipeline
 
 	SetupMLCallback SetupMLCallback // setup callback for ML job configs
 	InSetupCmd      bool            // this is set to true when the `setup` command is called
+
+	// XXX: remove Config from public interface
```
andrewkroh: It would be nice to include in the comment what's blocking this from happening or when it can be done.

libbeat/beat/beat.go
```diff
@@ -97,564 +29,27 @@ type Beater interface {
...
+	Config *BeatConfig // Common Beat configuration data.
+
+	BeatConfig *common.Config // The beats it's own configuration section
```
andrewkroh: `s/beats it's/beat's/`

libbeat/beat/info.go
```diff
@@ -1,9 +1,9 @@
-package common
+package beat
 
 import "github.com/satori/go.uuid"
 
 // BeatInfo stores a beats instance meta data.
-type BeatInfo struct {
+type Info struct {
 	Beat     string    // The actual beat its name
```
andrewkroh: `s/beat its/beat's/`

winlogbeat/beater/winlogbeat.go
```diff
@@ -49,21 +48,19 @@ func New(b *beat.Beat, _ *common.Config) (beat.Beater, error) {
 	// XXX: winlogbeat validates top-level config -> ignore beater config and
```
andrewkroh: This comment can be removed now that I remove the validation code.

winlogbeat/config/config.go
```diff
@@ -23,26 +23,12 @@ type Validator interface {
 	Validate() error
```
andrewkroh: Can this interface be removed too? I think it predates ucfg.

filebeat/prospector/udp/harvester.go
urso: yes :)

I did find it weird for filebeat having import order filebeat>libbeat and for packetbeat it being the other way around.

libbeat/cmd/instance/beat_test.go
urso: I put it in libbeat/cmd, as the `instance` package is used by almost all commands defined in libbeat/cmd. This way all code required or requiring a beat instance is combined in libbeat/cmd.

urso: Some more cleanup on instance or cmd package would be nice. But this was nothing I wanted to touch right now.

packetbeat/beater/devices.go
urso: it's mostly for sanity, as I wanted the `libbeat/cmd/instance` module to be used only internally. The `-devices` flag always has been a pretty ugly hack, generating some dependency between the beat and some weird global access to libbeat internals.

filebeat/prospector/udp/harvester.go
ruflin: @andrewkroh WDYT? Can we automate that? :-D

libbeat/cmd/instance/beat_test.go
ruflin: Yeah, also see some other cleanups coming but this is big enough already :-)

winlogbeat/eventlog/eventlogging.go
```diff
@@ -26,10 +26,9 @@ var eventLoggingConfigKeys = append(commonConfigKeys, "ignore_older",
 
 type eventLoggingConfig struct {
 	ConfigCommon     `config:",inline"`
-	IgnoreOlder      time.Duration          `config:"ignore_older"`
-	ReadBufferSize   uint                   `config:"read_buffer_size"   validate:"min=1"`
-	FormatBufferSize uint                   `config:"format_buffer_size" validate:"min=1"`
-	Raw              map[string]interface{} `config:",inline"`
```
ruflin: Not sure what `raw` was doing but want to make sure it didn't disappear by accident.

urso: I guess it used to be for additional 'validation' on allowed settings. But it's not used anymore in any place, so I removed it.

filebeat/prospector/udp/harvester.go
andrewkroh: Yes, but not with the standard `goimports`. It would take a customization to `goimports`, which isn't a bad idea, it would just take a little time.

PR #4832, opened 2017-08-07
Add docs for new subcommands

dedemorton:
Changes:
* moved all the content about the command-line interface for all the beats to a single file that uses conditional coding to build the correct content for each beat. This makes it easier to organize the content alphabetically but also share all the descriptions for commands that come from libbeat.
* created a container topic called "Setting up and running xyz" because I don't think the topics about the CLI and directory structure really belong in the Getting Started. 
* moved the topic about running on docker to appear below the "Setting up and running" container. 

You might want to build the content so you can see how it renders for each beat.

Please respond to the comments flagged for REVIEWERS.


libbeat/docs/command-reference.asciidoc
```diff
@@ -0,0 +1,583 @@
...
+Tests the configuration settings.
+
+//REVIEWERS: For each of the tests described here, should we say more about what we test?
+
+ifeval::["{beatname_lc}"=="metricbeat"]
+
+*`modules [MODULE_NAME] [METRICSET_NAME]`*::
+Tests module settings for all configured modules. To test the settings for a
+specific module, specify `MODULE_NAME`. To test the settings for a specific
+metricset in the module, also specify `METRICSET_NAME`.
```
exekias: For modules we do a test run, retrieving metrics with the current settings, and showing them as output

libbeat/docs/command-reference.asciidoc
```diff
@@ -0,0 +1,583 @@
...
+`--index` flags to futher define what gets exported.
+
+*FLAGS*
+
+*`--es.version VERSION`*::
+When specified along with <<template-subcommand,`template`>>, exports an index
+template that is compatible with the specified version.
+
+
+//REVIEWERS: Is the above statement true? The command-line help was too vague (said "Elastic version"). When I run export template with --es.version specified, nothing seems to happen.
```
dedemorton: That's AWESOME!

libbeat/docs/command-reference.asciidoc
```diff
@@ -0,0 +1,583 @@
...
+{beatname_lc} modules SUBCOMMAND [FLAGS]
+----
+
+
+*SUBCOMMANDS*
+
+*`disable MODULE_LIST`*::
+Disables the modules specified in the space-separated list.
+
+//REVIEWERS: It seems odd to me that we usually expect comma-separated lists, but here we expect a space-separated list.
```
dedemorton: Ah OK. I didn't understand that distinction (and couldn't really take the extra time to test all permutations).

libbeat/docs/command-reference.asciidoc
```diff
@@ -0,0 +1,583 @@
...
+
+endif::[]
+
+*`--memprofile FILE`*::
+Writes memory profile data to the specified output file. This option is useful
+for troubleshooting {beatname_uc}.
+
+ifeval::["{beatname_lc}"=="filebeat"]
+
+//REVIEWERS: I've written this topic to be consitent to what I see when I run -help, but I'm wondering why we aren't also showing the options that begin with a dash. For example, in the modules doc we show: -modules=nginx,mysql,system. Is --modules preferred?
```
dedemorton: For now, I'll keep the docs consistent with what we show in -h and change them when we update the help. I think it's probably good to keep the -options around for now because it might help existing users to see familiar commands. I do think, though, that we should consider using the --options in the getting started if those are preferred.

libbeat/docs/command-reference.asciidoc
```diff
@@ -0,0 +1,583 @@
...
+you set the `--once` flag, you should also set `close_eof` so the harvester is
+closed when the end of the file is reached. By default harvesters are closed
+after `close_inactive` is reached.
+
+
+//REVIEWERS: Is the behavior of the harvesters/prospectors described above still true?
+
+endif::[]
+
+//REVIEWERS: I'm not documenting --setup because it will be removed in 6.0. Are there any other flags that I need to remove? Or mention that they're being maintained for backwards compability?
```
dedemorton: @exekias I think it might have been a warning message when I ran the command? I'll add it back to the docs.

dedemorton:
@exekias 
>I saw the command list is following alphabetical order, would it make sense to promote some commands over others less important?

This is always a question that comes up. The problem with organizing commands by order of importance is that over time people who maintain the content end up adding new content without understanding the organizing principle. So the organization falls apart over time, and you end up with a randomly organized list. Also, as the list of commands grows, it becomes harder to find what you're looking for (maybe this will never be a problem for Beats, but it's worth considering).

With an alphabetized list, everyone understands why things are ordered the way they are, so I tend to gravitate towards organizing content alphabetically. I've had to maintain a lot  projects over the years where we always ended up having to alphabetize in the end...so I just sort of design for that in mind.

I don't have a strong preference myself, tbh, but since this list is so short, I don't think we need to worry about users missing stuff.

libbeat/docs/command-reference.asciidoc
dedemorton: @exekias Here's where setup is mentioned: 2017/08/11 21:11:27.693225 cfgwarn.go:18: WARN DEPRECATED: -setup flag has been deprectad, use setup subcommand Will be removed in version: 6.0

exekias: Fixed this here: https://github.com/elastic/beats/pull/4889

libbeat/docs/command-reference.asciidoc
```diff
@@ -0,0 +1,583 @@
...
+
+{setup-command-short-desc}.
+
+* The index template ensures that fields are mapped correctly in Elasticsearch.
+* The Kibana dashboards make it easier for you to visualize {beatname_uc} data
+in Kibana.
+* The machine learning jobs contain the configuration information and metadata
+necessary to analyze data for anomalies.
+
+//REVIEWERS: What are the requirements/permissions for users who want to load machine learning jobs? We need to document that here.
```
exekias: Do you know this @tsg?

libbeat/docs/command-reference.asciidoc
```diff
@@ -0,0 +1,583 @@
...
+*`--machine-learning`*::
+Sets up machine learning job configurations only.
+
+ifeval::["{beatname_lc}"=="filebeat"]
+
+*`--modules MODULE_LIST`*::
+Sets up the environment (index template, dashboards, etc) for the modules
+specified in the comma-separated list.
+
+//REVIEWERS: Is the above description correct? The help is unclear on what this sets up. Makes it sound like the command enables the modules. (Ran out of time to test this on my own...my apologies.)
```
exekias: `modules` flag works as in run, it's a shortcut to enable certain modules, and they will be added to the ones existing in yml conf. Then the setup process is done

libbeat/docs/command-reference.asciidoc
```diff
@@ -0,0 +1,583 @@
...
+----
+{beatname_lc} test SUBCOMMAND [FLAGS]
+----
+
+*SUBCOMMANDS*
+
+*`config`*::
+Tests the configuration settings.
+
+//REVIEWERS: For each of the tests described here, should we say more about what we test?
```
exekias: I think it's ok

libbeat/docs/command-reference.asciidoc
```diff
@@ -0,0 +1,583 @@
...
+
+*`--once`*::
+When the `--once` flag is used, {beatname_uc} starts all configured harvesters
+and prospectors, and runs each prospector until the harvesters are closed. If
+you set the `--once` flag, you should also set `close_eof` so the harvester is
+closed when the end of the file is reached. By default harvesters are closed
+after `close_inactive` is reached.
+
+
+//REVIEWERS: Is the behavior of the harvesters/prospectors described above still true?
```
urso: I don't think we touched the `--once` flag. Should still be true.

libbeat/docs/command-reference.asciidoc
dedemorton: @tsg I'm going to remove this comment from my review because it sounds like you'll be working on more docs related to ML and can update this topic if required.

libbeat/docs/command-reference.asciidoc
dedemorton: Not clear what you're saying. Are you saying that this flag is not meant to be specified with setup?

When I run the flag with the setup command, I see:

```
Rhodas-MacBook-Pro:filebeat-6.0.0-beta1-darwin-x86_64 dedemorton$ ./filebeat setup --modules nginx,mysql,system
Loaded index template
Loaded dashboards
Loaded machine learning job configurations
```
I don't see where/how the modules are being enabled by this command. It also seems like all the dashboards get loaded, not just the stuff for the modules that are specified in the list.

dedemorton: @exekias ^^

exekias: Yep, setup configures all dashboards, and I think it's the same for ML jobs. We only allow this flag to avoid errors in case no module is defined in filebeat.yml. To be consistent with `run` command. But it doesn't make a difference on what is configured.

dedemorton:
Build errors do not appear to be related to docs, so I'm merging this.

PR #4834, opened 2017-08-07
Make fields.yml part of the binary

ruflin:
Currently the fields.yml file is shipped as its own file and has to be place in the right directory to be used. As fields have become a central part of beats I suggest to make them part of the binary which will allow us to also split up the fields.yml inside the binary into multiple part (for example one per module) and use it for testing like fields checks.

The behavior from the user perspective is as following:

* By default internal `fields.yml` data is used
* If `fields.yml` path is set in the config file, it is loaded from disk

The golang files are generated when running `make update`. The content of fields.yml is stored as a compressed string.

Each asset is currently prefixed by the Beat name. The current fields registry is a basic implementation that likely needs to be extended when we add fields per module or processors.


tsg:
I think we need to move into the direction where we no longer concatenate all the module's `fields.yml` into one large file, but have them per module. This allows the compiling of "partial" mapping templates and index patterns, with only the fields that are in use.

Note sure if this is conflicting or is just orthogonal with this PR, but I wanted us to keep that in mind during the discussion.

ruflin:
I like the idea of having the "partial" fields.yml locally. The main thing that I see as a blocker to load partial mappings is that with ES 6.0 inheritance of templates was removed. This means if someone starts with the mysql module and then adds the redis module one large template with a new index would have to be used. One way out of this would be to have one index per module.

ruflin:
@andrewkroh go-bindata causes an issue here as it does not generate with which goimports is happy. I remember you had your own implementation of loading file data. Can you point me to it again?

andrewkroh:
It's over in elastic/go-libaudit.

- tool - https://github.com/elastic/go-libaudit/blob/master/aucoalesce/mknormalize_data.go
- sample usage - https://github.com/elastic/go-libaudit/blob/master/aucoalesce/normalize.go#L24
- output - https://github.com/elastic/go-libaudit/blob/master/aucoalesce/znormalize_data.go

ruflin:
@andrewkroh Thanks. I took your implementation and modified it a bit to be used here.

tsg:
@ruflin There's a probably related test failure here: https://travis-ci.org/elastic/beats/jobs/285439037#L1872

ruflin:
I was thinking of removing `fields.yml` from the package only when the export of `fields.yml` is possible. I added it as a task to this issue here: https://github.com/elastic/beats/issues/5347

ruflin:
@tsg Have a look again, tests should be fixed. Travis failure is related to https://github.com/elastic/logstash/pull/8456

filebeat/include/fields.go
```diff
@@ -0,0 +1,20 @@
...
+package include
+
+import (
+	"encoding/base64"
+
+	"github.com/elastic/beats/libbeat/assets"
+)
+
+func init() {
+	if err := assets.AddFields("fields.yml", Asset); err != nil {
```
andrewkroh: If two Beats were combined into the same process would this still work?

libbeat/cmd/instance/beat.go
```diff
@@ -147,6 +148,7 @@ func NewBeat(name, indexPrefix, v string) (*Beat, error) {
 			Hostname:    hostname,
 			UUID:        uuid.NewV4(),
 		},
+		Fields: assets.GetFields(),
```
andrewkroh: Maybe the beat name could be a param? (thinking of two beats in the same process)

libbeat/template/load_integration_test.go
```diff
@@ -237,7 +237,7 @@ func TestOverwrite(t *testing.T) {
 		Enabled: true,
 		Fields:  absPath + "/fields.yml",
 	})
-	loader, err := NewLoader(config, client, beatInfo)
+	loader, err := NewLoader(config, client, beatInfo, []byte{})
```
andrewkroh: Can you pass `nil` instead of allocating an empty slice?

libbeat/template/template.go
```diff
@@ -192,3 +211,34 @@ func (t *Template) generate(properties common.MapStr, dynamicTemplates []common.
...
+	if err != nil {
+		return nil, err
+	}
+	return load(cfg)
+}
+
+func load(cfg *ucfg.Config) (common.Fields, error) {
+	keys := []common.Field{}
+
+	cfg.Unpack(&keys)
```
andrewkroh: Missing error check?

libbeat/template/template.go
```diff
@@ -192,3 +211,34 @@ func (t *Template) generate(properties common.MapStr, dynamicTemplates []common.
...
+
+	cfg, err := yaml.NewConfig(data)
+	if err != nil {
+		return nil, err
+	}
+	return load(cfg)
+}
+
+func load(cfg *ucfg.Config) (common.Fields, error) {
+	keys := []common.Field{}
```
andrewkroh: If you use `var keys common.Field` will `Unpack` allocate this for you?

filebeat/include/fields.go
ruflin: The `AddFields` would work as currently the name is not used at all internally. What would cause problems is that the fields from libbeat would be duplicated. For now I kept the fields.yml exactly as it is and a 1-1 copy is added to each file. For some long terms ideas like splitting up for libbeat or modules, see PR description.

libbeat/cmd/instance/beat.go
ruflin: Are you thinking it cause some issues with the collectbeat from @vjsamuel ? Did not think about that TBH.

libbeat/template/load_integration_test.go
ruflin: changed

libbeat/template/template.go
ruflin: added

libbeat/template/template.go
ruflin: tested and seems to work.

filebeat/include/fields.go
andrewkroh: I was thinking that you would want the fields from each Beat to be stored independently such that each template (from `asset.GetFields`) contains only the fields for that Beat. But maybe that is moot if all beats in one process share a single output (not sure how that would work).

ruflin: It's definitively a valid point but I would ignore it for now.

andrewkroh:
Just a few notes from discussion:

You could auto-detect if `{path.home}/fields.yml` exists and use it.

ruflin:
This PR is ready for an other round of reviews.

ruflin:
jenkins, test this

ruflin:
One temporary issue of this change is that it will increase the PR conflicts every time fields are changed. But as soon as each module loads the fields locally, this should disappear again.

dev-tools/cmd/asset/asset.go
```diff
@@ -0,0 +1,89 @@
...
+
+var pkg *string
+
+func init() {
+	pkg = flag.String("pkg", "", "Package name")
+}
+
+var tpl = template.Must(template.New("normalizations").Parse(`
+// ${ES_BEATS}/dev-tools/cmd/asset/asset.go
+// MACHINE GENERATED BY THE ABOVE COMMAND; DO NOT EDIT
```
andrewkroh: This should be changed to `Code generated by beats/dev-tools/cmd/asset/asset.go - DO NOT EDIT.` Or something that matches this pattern https://github.com/golang/go/issues/13560#issuecomment-288457920. Then tools like IntelliJ and golint will recognize it as being generated.

generator/beat/{beat}/main.go.tmpl
```diff
@@ -4,6 +4,7 @@ import (
 	"os"
 
 	"{beat_path}/cmd"
+	_ "{beat_path}/include"
```
andrewkroh: I think that unused imports should be separated into their own group and have a comment explaining it's existence. Similar to the [golint warning](https://github.com/golang/lint/blob/470b6b0bb3005eda157f0275e2e4895055396a81/lint.go#L458), but golint exempts `main` packages.

ruflin:
@andrewkroh Will do a follow up PR with your requested changes.

PR #4835, opened 2017-08-07
Add support for dynamic param in fields

ruflin:
In fields.yml `dynamic: false` or `dynamic: strict` can now be used. For more details see https://www.elastic.co/guide/en/elasticsearch/reference/current/dynamic.html

### DIFF
--- a/libbeat/template/field.go
+++ b/libbeat/template/field.go
@@ -1,6 +1,7 @@
 package template
 
 import (
+	"fmt"
 	"strings"
 
 	"github.com/elastic/beats/libbeat/common"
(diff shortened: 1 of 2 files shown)


libbeat/template/field.go
```diff
@@ -23,6 +23,7 @@ type Field struct {
 	Analyzer       string `config:"analyzer"`
 	SearchAnalyzer string `config:"search_analyzer"`
 	Norms          bool   `config:"norms"`
+	Dynamic        string `config:"dynamic"`
```
urso: I'm missing some error handling for invalid strings.

e.g.

```

type Field struct {
  ...
  Dynamic dynamicType
  ...
}

type dynamicType interface{}

func (d *dynamicType) Unpack(s string) error {
  switch s {
    case "true":
      *d = true
    case "false":
      *d = false
    case "strict":
      *d = s
    default:
      return fmt.Errorf("'%v' is invalid dynamic setting", s)
  }
  return nil
}
```

you can also support `bool` in general via `interface{}`:

```
func (d *dynamicType) Unpack(in interface{}) error {
  if b, ok := in.(bool); ok {
    *d = b
    return nil
  }

  s, ok := in.(string)
  if !ok {
    return fmt.Errorf(...)
  }

  switch s {
    case "true":
      *d = true
    case "false":
      *d = false
    case "strict":
      *d = s
    default:
      return fmt.Errorf("'%v' is invalid dynamic setting", s)
  }
  return nil
}
```

libbeat/template/field.go
```diff
@@ -11,18 +12,19 @@ var (
...
+	Format         string       `config:"format"`
+	ScalingFactor  int          `config:"scaling_factor"`
+	Fields         Fields       `config:"fields"`
+	MultiFields    Fields       `config:"multi_fields"`
+	ObjectType     string       `config:"object_type"`
+	Enabled        *bool        `config:"enabled"`
+	Analyzer       string       `config:"analyzer"`
+	SearchAnalyzer string       `config:"search_analyzer"`
+	Norms          bool         `config:"norms"`
+	Dynamic        *dynamicType `config:"dynamic"`
```
urso: you don't really need a pointer here, as `Dynamic.value` will be `nil` if not set via `Unpack`.

ruflin: Fixed, new version pushed without the pointer.

ruflin:
jenkins, retest it

PR #4838, opened 2017-08-07
Add SetValue method to libbeat common.MapStr

simitt:
SetValue dereferences given values for *int, *string and *boolean,
otherwise it just sets the value for the given key as long as the value
is not nil or empty.

### DIFF
--- a/libbeat/common/mapstr.go
+++ b/libbeat/common/mapstr.go
@@ -138,13 +138,13 @@ func (m MapStr) Put(key string, value interface{}) (interface{}, error) {
 	return walkMap(key, m, mapStrOperation{putOperation{value}, true})
 }
 
-// Dereferences a value if it is a pointer. Sets the dereferenced value if not nil.
+// SetValue dereferences a value if it is a pointer. Sets the dereferenced value if not nil.
 // If value is a MapStr or an array, the value is set if not empty.
 // Supported data types so far: *string, *int, *boolean, common.MapStr, array[]int, array[]string.
 // Values of unsupported types are set if they are not nil.
 // Sets a value for a given key if the (dereferenced) value is not nil or empty.
 func (m MapStr) SetValue(key string, value interface{}) {
-	walkMap(key, m, mapStrOperation{setOperation{value}, true})
+	walkMap(key, m, mapStrOperation{setValueOperation{value}, true})
 }
 
 // StringToPrint returns the MapStr as pretty JSON.
@@ -390,11 +390,11 @@ func (op putOperation) Do(key string, data MapStr) (interface{}, error) {
 	return existingValue, nil
 }
 
-type setOperation struct {
+type setValueOperation struct {
 	Value interface{}
 }
 
-func (op setOperation) Do(key string, data MapStr) (interface{}, error) {
+func (op setValueOperation) Do(key string, data MapStr) (interface{}, error) {
 	if op.Value == nil {
 		return data, nil
 	}
--- a/libbeat/common/mapstr.go
+++ b/libbeat/common/mapstr.go
@@ -138,6 +138,15 @@ func (m MapStr) Put(key string, value interface{}) (interface{}, error) {
 	return walkMap(key, m, mapStrOperation{putOperation{value}, true})
 }
 
+// Dereferences a value if it is a pointer. Sets the dereferenced value if not nil.
+// If value is a MapStr or an array, the value is set if not empty.
+// Supported data types so far: *string, *int, *boolean, common.MapStr, array[]int, array[]string.
+// Values of unsupported types are set if they are not nil.
+// Sets a value for a given key if the (dereferenced) value is not nil or empty.
+func (m MapStr) SetValue(key string, value interface{}) {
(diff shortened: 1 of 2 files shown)


libbeat/common/mapstr.go
```diff
@@ -138,6 +138,15 @@ func (m MapStr) Put(key string, value interface{}) (interface{}, error) {
 	return walkMap(key, m, mapStrOperation{putOperation{value}, true})
 }
 
+// Dereferences a value if it is a pointer. Sets the dereferenced value if not nil.
```
ruflin: Comments in Golang must always start with the method name.

libbeat/common/mapstr.go
```diff
@@ -380,3 +389,43 @@ func (op putOperation) Do(key string, data MapStr) (interface{}, error) {
 	data[key] = op.Value
 	return existingValue, nil
 }
+
+type setOperation struct {
+	Value interface{}
+}
+
+func (op setOperation) Do(key string, data MapStr) (interface{}, error) {
```
ruflin: Should we make this private?

simitt: Then it would break with the [mapStrOperator interface](https://github.com/simitt/beats/blob/68fbee325f55615dcb402601adddb2290713e651/libbeat/common/mapstr.go#L351) which is used by all other methods. I just aligned using this interface to make it consistent.

ruflin:
jenkins, retest it

ruflin:
Jenkins is somehow hanging with auditbeat, but this is unrelated to the PR. Merging.

libbeat/common/mapstr.go
```diff
@@ -380,3 +389,43 @@ func (op putOperation) Do(key string, data MapStr) (interface{}, error) {
...
+}
+
+func (op setValueOperation) Do(key string, data MapStr) (interface{}, error) {
+	if op.Value == nil {
+		return data, nil
+	}
+
+	switch op.Value.(type) {
+	case *bool:
+		if newVal := op.Value.(*bool); newVal != nil {
```
andrewkroh: These secondary type assertions are not necessary if the switch statement is modified to be `switch v := op.Value.(type)` (see [TypeSwitchGuard](https://golang.org/ref/spec#Type_switches)). Within each case statement `v` takes on the type declared in the `case`.

libbeat/common/mapstr.go
```diff
@@ -380,3 +389,43 @@ func (op putOperation) Do(key string, data MapStr) (interface{}, error) {
...
+		}
+	case *int:
+		if newVal := op.Value.(*int); newVal != nil {
+			data[key] = *newVal
+		}
+	case *string:
+		if newVal := op.Value.(*string); newVal != nil {
+			data[key] = *newVal
+		}
+	case MapStr:
```
andrewkroh: These empty checks can be added to the appropriate `normalizeX` function in [libbeat/common/event.go](https://github.com/elastic/beats/blob/25fdff884e53a89b18207dd50e8c548361bdb851/libbeat/common/event.go).

libbeat/common/mapstr.go
```diff
@@ -138,6 +138,15 @@ func (m MapStr) Put(key string, value interface{}) (interface{}, error) {
 	return walkMap(key, m, mapStrOperation{putOperation{value}, true})
 }
 
+// SetValue dereferences a value if it is a pointer. Sets the dereferenced value if not nil.
```
ruflin: @andrewkroh I like the suggestion a lot. TBH I was not really aware / forgot that we have this. This removes the hassle of having to think about which method to use when creating the event.

What happens in the case that `Put("a.b.c", nil)` is used and the event is then normalised. Will the full tree be removed?

@simitt In case you agree with the change, do you want to work on this?

andrewkroh: > What happens in the case that Put("a.b.c", nil) is used

```
func TestEmptyMap(t *testing.T) {
	m := MapStr{}
	m.Put("a.b.c", nil)
	t.Log(ConvertToGenericEvent(m).StringToPrint())
}
```

```
{
  "a": {
    "b": {}
  }
}
```

So we need some additional checks in the normalization code to make the whole tree disappear.

simitt: Of course, I was not aware of having this normalization method in an event. Thank you for pointing this out, it will save ourselves some trouble. 
I will have a closer look and create a PR to remove empty lists then. 
@andrewkroh , @ruflin could one of you revert the commit in the beats project, I don't want to mess with your master.

libbeat/common/mapstr.go
simitt: Nice to know, will have a closer look then..

libbeat/common/mapstr.go
ruflin: @simitt No worries about reverting the commit. Just remove the code again in the follow up commit. I think so far we never reverted a commit, we only move forward :-D

simitt: just thought it would be cleaner to revert, but fine with me this way.

PR #4843, opened 2017-08-08
Fix link to image in FBM Postgres docs

tsg:


ruflin:
Merged as docs build was green.

PR #4844, opened 2017-08-08
filebeat: remove unreferenced image from documentation

kvch:
An unwanted image was pushed to the PostgreSQL documentation which is not referenced anywhere. 

Two images are included in the doc:
- filebeat-postgresql-overview.png
- filebeat-postgresql-slowlog-overview.png

This PR removes the unnecessary filebeat-postgresql.png from the docs.


ruflin:
WFG

@kvch Lets keep an eye on the doc build after merging to make sure we don't break it.

tsg:
Docs are green.

PR #4853, opened 2017-08-09
Add support for index to fields.yml

ruflin:
Setting `index: false` for a field was not possible so far. This adds this possibility. For more details in `index` see: https://www.elastic.co/guide/en/elasticsearch/reference/5.5/mapping-index.html

Additional changes:

* `enable` and `dynamic` are only available for objects. The functionality was move to object. https://www.elastic.co/guide/en/elasticsearch/reference/5.5/enabled.html
* Add support for `enable` and `index` to the docs script so it is visible in the docs which fields are indexed and which ones not

### DIFF
--- a/libbeat/template/field.go
+++ b/libbeat/template/field.go
@@ -25,6 +25,7 @@ type Field struct {
 	SearchAnalyzer string      `config:"search_analyzer"`
 	Norms          bool        `config:"norms"`
 	Dynamic        dynamicType `config:"dynamic"`
+	Index          *bool       `config:"index"`
 
 	path      string
 	esVersion common.Version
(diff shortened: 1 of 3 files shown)


libbeat/template/field.go
```diff
@@ -183,16 +192,13 @@ func (f *Field) addDynamicTemplate(properties common.MapStr, matchType string) {
...
 	// Currently no defaults exist
-	property := common.MapStr{}
-	if f.Enabled != nil {
-		property["enabled"] = *f.Enabled
-	}
+	properties := common.MapStr{}
 
-	if f.Dynamic.value != nil {
-		property["dynamic"] = f.Dynamic.value
+	if f.Index != nil {
```
simitt: Why do you treat `f.Index` as default property but not `f.Enabled`?

ruflin: As enable is only available for objects, so I moved it to objects.

ruflin:
jenkins, test it

ruflin:
jenkins, test it

Issue #4851, opened 2017-08-09
Heartbeat @timestamp value is wrong


pmusa:
- Version: 6.0.0-beta1-linux-x86_64
- Operating System: Ubuntu 16.04
- Steps to Reproduce: 
Start heartbeat with the default configuration and check that the `@timestamp` value generated is always the same for every new event that is generated. (I imagine it is using the wrong timestamp variable to set it.)

michaelgoconnor:
This issue is also reproducible on Windows

PR #4855, opened 2017-08-09
Fix wrong timestamp in heartbeat events

urso:
Resolves: #4851


andrewkroh:
LGTM. But I merged the other PR and it caused a conflict.

PR #4856, opened 2017-08-09
Fix monitor.name being empty by default

### LINKED ISSUE #4852: Heartbeat Kibana saved search has the wrong field  (dashboard does not work with the default configuration)
- Version: 6.0.0-beta1-linux-x86_64
- Operating System: Ubuntu
- Steps to Reproduce: 
Start Heartbeat with the `-setup` flag and then, in Kibana, access the Heartbeat dashboard to see `No results found` message. Digging a bit, I found that every visualization in the auto-loaded dashboard is configured with the Heartbeat "saved search", which is currently using `monitor.name` instead of `monitor.type`.

I am not sure if `monitor.name` was deprecated in favor of `monitor.type` or if there is another problem. Also, in the same "saved search" there is a meta field that also accesses `monitor.name`. Maybe it needs to be changed as well.

urso:
Resolve #4852 

- Fixes the `monitor.name` field default to `monitor.type` if not configured
- Update HTTP dashboards to filter on `monitor.type`, which will always be `http` for all HTTP related monitors.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -45,6 +45,8 @@ https://github.com/elastic/beats/compare/v6.0.0-beta1...master[Check the HEAD di
 
 *Heartbeat*
 
+- Fix monitor.name being empty by default. {issue}4852[4852]
+
 *Metricbeat*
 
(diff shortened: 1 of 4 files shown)


monicasarbu:
@urso Great, waiting for green.

PR #4860, opened 2017-08-09
[WIP] Test runner changes

urso:


andrewkroh:
```
OSError: [Errno 13] Permission denied: '/tmp/python-env/lib/python2.7/site-packages/urllib3-1.21.1.dist-info/DESCRIPTION.rst'
scripts/Makefile:203: recipe for target 'python-env' failed
```

It took me a while to figure out the cause, but I believe the cause is that the [Dockerfile](https://github.com/elastic/beats/blob/master/libbeat/Dockerfile#L13-L18) pre-builds the virtual environment as root.

ruflin:
It would be great if we get this working and can get rid of the cleanup.

ruflin:
@urso @andrewkroh I wonder how we should move forward here. We made some improvements on the docker and setup side which "mostly" solved the problem but not completely. I'm thinking of also getting this PR in as an "additional" fix? It would need a rebase. WDYT?

heartbeat/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM golang:1.8.3
```
ruflin: @urso I assume the heartbeat docker and compose file are not really needed for this PR?

urso: Depending on test environment the tests themselves run in a docker environment. I added the docker container, such that heartbeat test/build environment will be more similar to the other beats.

urso:
+1 for moving forward. I still think these changes are required for a clean test run.

ruflin:
@urso @andrewkroh Going to close this PR for now as CI is working now with fixes on the infra side. We can still come back to this later if we see issues again.

PR #4864, opened 2017-08-10
Change deprecated _default_ mapping to doc in index templates

### LINKED ISSUE #4840: Eliminate deprecated _default_ mapping in 6.x
The following warnings show up in the logs of Elasticsearch 6.x:
```
[2017-08-07T14:22:51,002][WARN ][o.e.d.a.a.i.t.p.PutIndexTemplateRequest] Deprecated field [template] used, replaced by [index_patterns]
[2017-08-07T14:22:51,019][WARN ][o.e.d.i.m.MapperService  ] [_default_] mapping is deprecated since it is not useful anymore now that indexes cannot have more than one type
```
As stated in the documentation of [default mapping](https://www.elastic.co/guide/en/elasticsearch/reference/6.x/default-mapping.html) it is deprecated since 6.0. The format of index templates must change in order to eliminated deprecated fields.

kvch:
`_default_` has been changed to `doc` as seen in this example: https://www.elastic.co/guide/en/elasticsearch/reference/master/mapping.html#_example_mapping

Closes #4840

### DIFF
--- a/libbeat/template/template.go
+++ b/libbeat/template/template.go
@@ -98,10 +98,17 @@ func (t *Template) generate(properties common.MapStr, dynamicTemplates []common.
 	}
 	indexSettings.DeepUpdate(t.settings.Index)
 
+	var mappingName string
+	if t.esVersion.Major >= 6 {
+		mappingName = "doc"
+	} else {
+		mappingName = "_default_"
+	}
+
 	// Load basic structure
 	basicStructure := common.MapStr{
 		"mappings": common.MapStr{
-			"_default_": common.MapStr{
+			mappingName: common.MapStr{
 				"_meta": common.MapStr{
 					"version": t.beatVersion.String(),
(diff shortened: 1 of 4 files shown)


libbeat/template/template.go
```diff
@@ -128,7 +128,7 @@ func (t *Template) generate(properties common.MapStr, dynamicTemplates []common.
 	}
 
 	if t.esVersion.IsMajor(2) || t.esVersion.IsMajor(5) {
-		basicStructure.Put("mappings._default_._all.norms.enabled", false)
```
ruflin: @urso We also use `doc` for all types if we write to indices from 5.x with the ES output? Just want to make sure this also works with older versions of ES.

tsg:
I merged this one, I think it's safe. @ruflin, please speak up if you have any concerns.

PR #4867, opened 2017-08-10
Set default credentials for Kibana

monicasarbu:
When loading the dashboards from Beats 6.0, the user needs to configure:
- Elasticsearch URL
- Kibana URL
- Elasticsearch credentials 
- Kibana credentials

In Could, the credentials for connecting to Elasticsearch are the same as for connecting to Kibana, so the last step can be removed if we take (by default) the Kibana credentials from the Elasticsearch credentials.

This PR gets by default the credentials for Kibana from Elasticsearch configuration.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -60,6 +60,7 @@ https://github.com/elastic/beats/compare/v6.0.0-beta1...master[Check the HEAD di
 *Affecting all Beats*
 
 - Update init scripts to use the `test config` subcommand instead of the deprecated `-configtest` flag. {issue}4600[4600]
+- Get by default the credentials for connecting to Kibana from the Elasticsearch output configuration. {pull}4867[4867]
 
 *Auditbeat*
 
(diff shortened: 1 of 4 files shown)


LeeDr:
I'm -1 on this change.  If you're using the same credentials (in Cloud or elsewhere) to connect to Elasticsearch and Kibana I'm guessing you're using the elastic superuser?  I don't think this should be suggested or the default.  You should generally use users with the minimal privileges they need to get the job done.  Even in Cloud, we should create roles and users with the correct privileges for the task.

In my integration test I create a specific role and user `beats_internal` that has just the privileges needed for any of the beats to write their data to Elasticsearch (output.elasticsearch:).

And to access the Kibana API to create the index-patterns, dashboards, etc I'm currently using the kibana user (kibana_system role).  But I think even that is more privileged than it needs to be.  Now I'm thinking the kibana_user role should be enough since any kibana_user can create all those things.

monicasarbu:
I agree that having different users for ES and Kibana is what we should recommend for production, but this simplifies the getting started with xpack security and Cloud cases. It's a tradeoff, but in this case, I think adding convenience for getting started is worth it.

Note that the old implementation (writing directly in .kibana index) was worse in the sense that one couldn't have different users. So this is not a security regression since 5.x, on the contrary.

PR #4873, opened 2017-08-11
Update build_docs.sh script to support dev-guide

ruflin:
Currently the dev-guide is not build as part of `make docs`. This adjustments to the scripts and Makefile add the dev-guide and offer more flexibility for docs build also for community beats.

The script was also moved out of libbeat to the global script directory as it does not directly belong to libbeat.


ruflin:
jenkins, test it

PR #4874, opened 2017-08-11
Experiment: How to run metricbeat module tests differently

ruflin:
This introduces a script that starts the environment for each module and tears it down again after the tests.

Currently missing is setting the right environment parameters like the password for the tests. I'm sure there is a way to easily load and set ENV variables from a file.

Other open issues:

* How would `integration-tests` execute this command?
* This requires that the ports which are needed for testing are available on the machine. This could be especially problematic for port 80 for example. We could generate a temp env file to have different ports or always export to a specific port instead.


ruflin:
Closing as @exekias did a different implementation which is now in master.

PR #4753, opened 2017-07-25
Remove settings from docs

urso:
- remove removed settings in libbeat/filebeat docs
- add tags/fields/fields_under_root/processor settings to every beat

PR #4878, opened 2017-08-11
Cherry-pick #4753 to 6.0: Remove settings from docs

urso:
Cherry-pick of PR #4753 to 6.0 branch. Original message: 

- remove removed settings in libbeat/filebeat docs
- add tags/fields/fields_under_root/processor settings to every beat


heartbeat/docs/heartbeat-options.asciidoc
```diff
@@ -123,6 +123,44 @@ value specified for `timeout` is greater than `schedule`, intermediate checks
...
+grouped under a `fields` sub-dictionary in the output document. To store the
+custom fields as top-level fields, set the `fields_under_root` option to true.
+If a duplicate field is declared in the general configuration, then its value
+will be overwritten by the value declared here.
+
+[float]
+[[monitor-fields-under-root]]
+==== `fields_under_root`
+
+If this option is set to true, the custom <<monitor-fields>>
```
dedemorton: Change this to say, 
```
the custom <<monitor-fields,`fields`>>
```

I despise hard-coding links, but unfortunately our doc build thinks it's smart and adds text to the link. You end up with this monstrosity when you don't specify link text:

image

You might not have noticed this before because the doc build only adds the link text if you're pointing to a sub-section. (And the logic for deciding whether something is a sub-section depends not just on the heading level that you use in the asciidoc source, but also the chunk level.)

dedemorton:
LGTM. Just one minor comment about the link. I've already fixed the problem in the output config topic. I guess it probably needs to be fixed in master, too.

heartbeat/docs/heartbeat-options.asciidoc
urso: How come this works correctly in master: https://www.elastic.co/guide/en/beats/heartbeat/master/configuration-heartbeat-options.html#monitor-fields-under-root

urso:
Not too happy to modify cherry picks. I'd rather update master and cherry pick the change in master into 6.0 branch, after this one has been merged.

urso:
@dedemorton see #4919

heartbeat/docs/heartbeat-options.asciidoc
dedemorton: @urso It works because Clint is amazing. I complained about the behavior again yesterday, and he fixed it! https://github.com/elastic/docs/commit/e427db2aa82fb34d0d3dae9a4f24748472dc98a7

PR #4884, opened 2017-08-11
Add queue settings to docs

urso:


libbeat/docs/generalconfig.asciidoc
```diff
@@ -103,3 +103,70 @@ processors in your config.
...
+
+Example configuration:
+
+[source,yaml]
+------------------------------------------------------------------------------
+queue.mem:
+  events: 4096
+------------------------------------------------------------------------------
+
+==== Configure the Memory Qeueue
```
dedemorton: Misspelling. This should say "Configure the memory queue". However, I don't think you need a separate section here (at least, not until we add support for more queue types--but even then, I'm not sure your need a separate section). It sounds like the memory queue is the only one that is supported right now? If that's true, we should state that.

libbeat/docs/generalconfig.asciidoc
```diff
@@ -103,3 +103,70 @@ processors in your config.
...
+[[configuration-internal-queue-memory]]
+
+The memory queue keeps all events in memory. By default no flush interval is
+configured. All events published to this queue will be directly consumed by the
+outputs. An output will consume up to the outputs `bulk_max_size` events at once.
```
dedemorton: This sentence is a bit hard to parse: "An output will consume up to the outputs `bulk_max_size` events at once."

How about this? "An output will consume events in batches based on the output's `bulk_max_size` setting."

libbeat/docs/generalconfig.asciidoc
```diff
@@ -103,3 +103,70 @@ processors in your config.
...
+
+Only after an event is dropped or acknowledged by the output, a new slot will
```
dedemorton: If the output is reading in batches, what do we mean by a slot? Does the slot open up before the batch is fully read? Maybe I'm not fully understanding the behavior here.

libbeat/docs/generalconfig.asciidoc
```diff
@@ -103,3 +103,70 @@ processors in your config.
...
+be available in the queue.
+
+By setting `flush.min_events` and `flush.timeout`, spooling in the queue is
```
dedemorton: Would be more direct and say, "To enforce spooling in the queue, set the `flush.min_events` and `flush.timeout` options.

libbeat/docs/generalconfig.asciidoc
```diff
@@ -103,3 +103,70 @@ processors in your config.
...
 Sets the maximum number of CPUs that can be executing simultaneously. The
 default is the number of logical CPUs available in the system.
+
+[float]
+[[configuration-internal-queue]]
+=== Configure the internal queue
+
+The `queue` namespace configures the queue type to be used.
+
+The Elastic Beats employ an internal queue for events to be published. The
```
dedemorton: Crap weasel. Looks like some of my comments didn't get saved when I submitted the review. I'd change this to say:

```
{beatname_uc} uses an internal queue to store events before publishing them. 
```

libbeat/docs/generalconfig.asciidoc
```diff
@@ -103,3 +103,70 @@ processors in your config.
...
+queue is responsible for buffering and combining events into batches, which can
```
dedemorton: No, it's the other way around. You use "that" for restrictive clauses. By restrictive, we mean a clause that gives information essential to the meaning of the sentence. In this case, the clause "that can be consumed by the outputs" is restrictive.  Sometimes it's a judgment call. TBH, the only people who really care about this distinction are writers and editors. :-) :-)

libbeat/docs/generalconfig.asciidoc
dedemorton: I like when we can be concise, but not if we're obfuscating meaning. In this case, I think the meaning is not clear and the grammatical structure is questionable because you are leaving words out. What you're really saying here is that "An output will consume up to the number of events specified by the output's `bulk_max_size` setting." You're using shorthand that's common for developers to use, but will probably seem off to some readers. They won't make the same substitution that you make in your head. 

Anyhow, for now, at least make "output" possessive so that you have "the output's `bulk_max_size`.

libbeat/docs/queueconfig.asciidoc
```diff
@@ -22,14 +22,14 @@ queue.mem:
 
 The memory queue keeps all events in memory. It is the only queue type
 supported right now.  By default no flush interval is configured. All events
-published to this queue will be directly consumed by the outputs. An output
-will consume up to the outputs `bulk_max_size` events at once.
+published to this queue will be directly consumed by the outputs.
+The output's `bulk_max_size` setting limits the number of events being processed at once.
```
dedemorton: this works!

libbeat/docs/queueconfig.asciidoc
```diff
@@ -0,0 +1,69 @@
+[[configuring-internal-queue]]
+== Configure the internal queue
+
+The {beatname_uc} uses an internal queue to store events before publishing them. The
```
dedemorton: Minor: remove "The".

libbeat/docs/queueconfig.asciidoc
```diff
@@ -0,0 +1,69 @@
...
+
+
+Example configuration:
+
+[source,yaml]
+------------------------------------------------------------------------------
+queue.mem:
+  events: 4096
+------------------------------------------------------------------------------
+
```
dedemorton: I'd add a [float] tag here to keep the info about the memory queue from becoming a separate topic. If you don't use the [float] tag, you get this:

image

libbeat/docs/queueconfig.asciidoc
urso: done

libbeat/docs/queueconfig.asciidoc
urso: Done. Change did require me to add `[float]` about everywhere for the doc build not to fail.

dedemorton: Yah, I really dislike [float] tags. We use them all over the library. They seem like such a hack, but are impossible to avoid without re-architecting all the content...and even so, you would still end up having to use them in some places.

ruflin:
@dedemorton Could you have a look again at this one to see if we can get it in?

libbeat/docs/queueconfig.asciidoc
```diff
@@ -0,0 +1,74 @@
...
+[float]
+[[configuration-internal-queue-memory]]
+=== Configure the memory qeueue
+
+The memory queue keeps all events in memory. It is the only queue type
+supported right now.  By default no flush interval is configured. All events
+published to this queue will be directly consumed by the outputs.
+The output's `bulk_max_size` setting limits the number of events being processed at once.
+
+The memory queue is waiting for the output to acknowledge or drop an events. If
```
dedemorton: This sentence has a typo "an events". The verb tense also seems a little off. Should probably say:

The memory queue waits for the output to acknowledge or drop events.

dedemorton:
One minor comment. Otherwise, LGTM. Sorry about the delay. Didn't realize you were waiting for me to confirm your changes.

ruflin:
@dedemorton Thanks for the review again. I fixed the sentence above and will squash merge it as soon as the build goes green.

ruflin:
@urso This needs a rebase on master because of some changes in jenkins I made :-( Could you rebase it and squash the commits?

PR #4885, opened 2017-08-11
New Filebeat module: Kafka logs

kvch:
This module is based on https://www.elastic.co/blog/monitoring-kafka-with-elastic-stack-1-filebeat written by @urso.

This PR only includes the log fileset.


filebeat/module/kafka/log/manifest.yml
```diff
@@ -0,0 +1,14 @@
+module_version: 1.0
+
+kafka.home: /opt/kafka[_-\.0-9]*
```
tsg: I think this could be made also a `var`, which can be referred from the `paths` array. This way, the user can only set that variable. Unless I missremember, the variables can refer to variables defined before them using the syntax `{{.home}}`.

tsg:
Nice work so far, could you also add a screenshot of the Kibana dashboard?

kvch:
I added the Kibana dashboard along with some missing documentation. I also made `kafka_home` configurable. Please, check if I did the right thing. :)

And for the record the GC logs will come in a separate PR. So it can be merged, after it passes the review and the tests.

filebeat/module/kafka/log/manifest.yml
```diff
@@ -0,0 +1,15 @@
+module_version: 1.0
+
+var:
+  - name: kafka_home
+    default:
+      - /opt/kafka[_-\.0-9]*a
+  - name: paths
+    default:
+      - ${kafka_home}/logs/controller.log*
```
tsg: Does this work? I would have expected this to use go template syntax, so something like `{{.kafka_home}}`.

tsg:
jenkins, retest it pls

filebeat/module/kafka/_meta/config.yml
```diff
@@ -0,0 +1,13 @@
+- module: kafka
+  # All logs
+  log:
+    enabled: true
+
+    # Set custom paths for Kafka. If left empty,
+    # Filebeat will look under /opt.
+    #var.kafka_home:
```
tsg: I think we can put the default in the commented line here, so it's clear for the user how to edit. I.e.:

```
      #var.kafka_home: /opt/kafka*
```

Btw, it's `/opt/kafka*`, not just `/opt`, right?

filebeat/module/kafka/_meta/config.yml
```diff
@@ -0,0 +1,13 @@
...
+
+    # Set custom paths for the log files. If left empty,
+    # Filebeat will choose the paths depending on your OS.
```
tsg: This comment is out-of-date now, I think. It should say it chooses the paths based on the `kibana_path` value.

tsg:
There seems to be some issue with the Kibana dashboard:

screen shot 2017-08-15 at 12 16 31

I'm not sure if something changed on the Kibana side that broke our saving/loading of dashboards..

tsg:
I see now that this also affects the Postgres dashboards, but not other dashboards.

kvch:
jenkins, test it

kvch:
I updated the dashboards and now it is working. I will open a separate PR for the PostgreSQL dashboards.

PR #4887, opened 2017-08-11
Add system uptime metricset

### LINKED ISSUE #4848: Add system uptime metricset to Metricbeat
To monitor system uptime and detect machine reboots we should add a new metricset to the Metricbeat system module. We would add this metricset to our default system module config and use a longer reporting period than we typically use since this data doesn't need to be reported very often.

andrewkroh:
This adds a new metricset that reports the system uptime. The metricset
is enabled by default with a period of 15m.

The metricset reports one field `system.uptime.duration.ms` that is the system's
uptime reported in milliseconds. In Kibana the field is formatted as a `duration`
and shown in human-readable format (e.g. "8 days").

uptime-kibana

Closes #4848


metricbeat/modules.d/system.yml
```diff
@@ -23,3 +23,7 @@
   processors:
   - drop_event.when.regexp:
       system.filesystem.mount_point: '^/(sys|cgroup|proc|dev|etc|host|lib)($|/)'
+
+- module: system
+  period: 15m
+  metricsets: [uptime]
```
exekias: I'm wondering if this format (instead of a list with `-`) may confuse some users, as we don't follow the same pattern in the whole file. WDYT?

libbeat/scripts/generate_index_pattern.py
```diff
@@ -79,6 +79,11 @@ def field_to_json(fields, desc, path, output,
         if "format" in desc:
             fieldFormat["id"] = desc["format"]
 
+            if "input_format" in desc:
```
ruflin: Nice

CHANGELOG.asciidoc
```diff
@@ -83,6 +83,8 @@ https://github.com/elastic/beats/compare/v6.0.0-beta1...master[Check the HEAD di
 - Add http server metricset to support push metrics via http. {pull}4770[4770]
 - Make config object public for graphite and http server {pull}4820[4820]
 
+- Add system uptime metricset. {issue}[4848[4848]
```
ruflin: Empty line above on purpose?

libbeat/scripts/generate_index_pattern.py
monicasarbu: Great idea!

metricbeat/modules.d/system.yml
andrewkroh: I agree. Changed to block style.

CHANGELOG.asciidoc
andrewkroh: Fixed

metricbeat/module/system/uptime/metricset_test.go
```diff
@@ -0,0 +1,28 @@
+// +build darwin linux openbsd windows
```
jsoriano: @andrewkroh was there any reason to exclude freebsd here? sigar has the [implementation for this](https://github.com/elastic/gosigar/blob/master/sigar_freebsd.go#L38). // cc @kaiyan-sheng @ruflin

andrewkroh: I can’t think of any reason. It probably should be tagged as `freebsd,cgo`.

PR #4888, opened 2017-08-12
Register kubernetes field_format matcher and remove logger in Encode API

vjsamuel:
`field_format` matcher was not registered. Fixing that.
There was a `continue` missing when an unknown matcherFunc was invoked, causing to get nil pointer exception.

Removing logger from the Encode API as the error is returned to the caller, lets leave it to the caller to decide if its an error or a graceful failure.

### DIFF
--- a/libbeat/outputs/codec/format/format.go
+++ b/libbeat/outputs/codec/format/format.go
@@ -6,7 +6,6 @@ import (
 	"github.com/elastic/beats/libbeat/beat"
 	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/common/fmtstr"
-	"github.com/elastic/beats/libbeat/logp"
 	"github.com/elastic/beats/libbeat/outputs/codec"
 )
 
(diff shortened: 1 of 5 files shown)


libbeat/outputs/codec/format/format.go
```diff
@@ -38,9 +37,5 @@ func New(fmt *fmtstr.EventFormatString) *Encoder {
 }
 
 func (e *Encoder) Encode(_ string, event *beat.Event) ([]byte, error) {
-	serializedEvent, err := e.Format.RunBytes(event)
-	if err != nil {
-		logp.Err("Fail to format event (%v): %#v", err, event)
-	}
-	return serializedEvent, err
+	return e.Format.RunBytes(event)
```
exekias: @urso, any opinion on this?

vjsamuel: what will happen when we use `field_format` matcher, is that whenever there is no such fields in the event, it would keep logging this error and the number of such errors can be massive. @urso @exekias

urso: The formatter first tries to extract the fields used in the format string from the event. And only then it applies the event to the format string. One can use default values in the format string.

On error the event will be dropped (even for filebeat). That is, we have to trade excessive logging vs. 'silently' dropping events. The error can be investigated easily. Just dropping without error might confuse users potentially thinking filebeat is super unstable, just cause the configuration is not matching the event.

I checked the outputs and all but redis do log the error as err/warning/critical (depends on output and sending guarantees). I'm ok with removing the log.Err here, but then the error must be logged in the redis output. At some point we need to collect and log potentially excessive logs in a more 'sane' manner.

vjsamuel: @urso, i have logged it as per your request.

PR #4891, opened 2017-08-14
Update Kibana index pattern with index and enable from fields

ruflin:
With https://github.com/elastic/beats/pull/4853 in fields.yml fields can be marked as `enable: false` or `index: false`. This should also have an affect on the index pattern. Objects which are not enabled will be skipped in the index pattern and fields which are not indexed all settings are set to false.

### DIFF
--- a/libbeat/scripts/generate_index_pattern.py
+++ b/libbeat/scripts/generate_index_pattern.py
@@ -71,6 +71,18 @@ def field_to_json(fields, desc, path, output,
     else:
         field["type"] = "string"
 
+    if "enable" in desc and not desc["enabled"]:
+        # Skip objects which are not enabled
+        return
+
+    if "index" in desc and not desc["index"]:
+        # If fields is not indexed, mark it in Kibana
+        field["indexed"] = False
+        field["analyzed"] = False
+        field["doc_values"] = False
+        field["searchable"] = False
+        field["aggregatable"] = False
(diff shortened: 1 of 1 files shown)


ruflin:
Set it to `in progress` as I want to do some manual tests.

libbeat/scripts/generate_index_pattern.py
```diff
@@ -71,6 +71,18 @@ def field_to_json(fields, desc, path, output,
...
+        # Skip objects which are not enabled
+        return
+
+    if "index" in desc and not desc["index"]:
+        # If fields is not indexed, mark it in Kibana
+        field["indexed"] = False
+        field["analyzed"] = False
+        field["doc_values"] = False
+        field["searchable"] = False
+        field["aggregatable"] = False
```
simitt: great you are fixing this, I already invested some time today in checking why the it still shows up in the index pattern.

You mentioned manual testing, could you also check that those fields are not shown any more in the auto-completion feature when writing a query in kibana. It is weird to get the suggestion, but then nothing is found when using it in the query.

ruflin: Will do 👍  Need to figure out how Kibana does generate this auto-completion list.

ruflin:
Closing in favor of a follow up PR from @simitt

PR #4892, opened 2017-08-14
Cherry-pick #4867 to 6.0: Set default credentials for Kibana

monicasarbu:
Cherry-pick of PR #4867 to 6.0 branch. Original message: 

When loading the dashboards from Beats 6.0, the user needs to configure:
- Elasticsearch URL
- Kibana URL
- Elasticsearch credentials 
- Kibana credentials

In Could, the credentials for connecting to Elasticsearch are the same as for connecting to Kibana, so the last step can be removed if we take (by default) the Kibana credentials from the Elasticsearch credentials.

This PR gets by default the credentials for Kibana from Elasticsearch configuration.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -49,6 +49,9 @@ https://github.com/elastic/beats/compare/v6.0.0-beta1...master[Check the HEAD di
 
 *Affecting all Beats*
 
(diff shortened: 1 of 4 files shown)


tsg:
jenkins, test it again

PR #4896, opened 2017-08-14
Set array type to ObjectType of field if available.

simitt:
Type of array items needs to be defined when creating a mapping for arrays. For dynamic mapping ES uses the type of the first array item. For pre-defined mapping use ObjectType information set in the fields.yml if available.


ruflin:
jenkins, test it

ruflin:
Merging as build failures are unrelated.

PR #4903, opened 2017-08-15
Move TCP UDP start up into `server.Start()`

vjsamuel:
Making this change since `NewTcpServer` and `NewUdpServer` open the port which is not the ideal scenario.

### DIFF
--- a/metricbeat/helper/server/udp/udp.go
+++ b/metricbeat/helper/server/udp/udp.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"net"
 
+	"github.com/pkg/errors"
+
 	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/logp"
 	"github.com/elastic/beats/metricbeat/helper/server"
@@ -11,6 +13,7 @@ import (
 )
 
 type UdpServer struct {
+	udpaddr           *net.UDPAddr
 	listener          *net.UDPConn
 	receiveBufferSize int
 	done              chan struct{}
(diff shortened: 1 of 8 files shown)


urso:
@vjsamuel please check the test output. It seems there is a panic in netpoll after this change.

metricbeat/module/graphite/server/server.go
```diff
@@ -61,7 +62,11 @@ func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
 // Run method provides the Graphite server with a reporter with which events can be reported.
 func (m *MetricSet) Run(reporter mb.PushReporter) {
 	// Start event watcher
-	m.server.Start()
+	err := m.server.Start()
+	if err != nil {
```
andrewkroh: You could tighten this up `if err := m.server.Start(); err != nil {...`

metricbeat/module/graphite/server/server.go
```diff
@@ -61,7 +62,11 @@ func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
...
+		logp.Err("%v", err)
```
andrewkroh: You might find it useful to report the error to the output (e.g. `reporter.Error(err)`). Then you will the info that one of the graphite listeners failed sent to your output.
```
err = errors.Wrap(err, "failed to start graphite server")
logp.Err("%v, err)
reporter.Error(err)
return
```

metricbeat/helper/server/udp/udp.go
```diff
@@ -43,22 +45,25 @@ func NewUdpServer(base mb.BaseMetricSet) (server.Server, error) {
...
 		done:              make(chan struct{}),
 		eventQueue:        make(chan server.Event),
 	}, nil
 }
 
-func (g *UdpServer) Start() {
+func (g *UdpServer) Start() error {
+	listener, err := net.ListenUDP("udp", g.udpaddr)
+	if err != nil {
+		return errors.Wrap(err, "Failed to start UDP server")
```
andrewkroh: Error strings should not be capitalized. https://github.com/golang/go/wiki/CodeReviewComments#error-strings

andrewkroh:
@vjsamuel According to [travis-ci](https://travis-ci.org/elastic/beats/jobs/265492775#L574-L577) this still needs to have `make fmt` run on it.

PR #4905, opened 2017-08-15
Fix PostgreSQL dashboards

kvch:
Previously broken dashboards are fixed by using names as ids.

### DIFF
--- a/filebeat/module/postgresql/_meta/kibana/default/dashboard/Filebeat-Postgresql-overview.json
+++ b/filebeat/module/postgresql/_meta/kibana/default/dashboard/Filebeat-Postgresql-overview.json
@@ -6,13 +6,13 @@
         "kibanaSavedObjectMeta": {
           "searchSourceJSON": "{\"filter\":[]}"
         },
-        "savedSearchId": "ef9da9b0-776c-11e7-9e03-393c0d1aa325",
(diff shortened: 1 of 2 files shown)


filebeat/module/postgresql/_meta/kibana/default/dashboard/Filebeat-Postgresql-overview.json
```diff
@@ -6,13 +6,13 @@
         "kibanaSavedObjectMeta": {
           "searchSourceJSON": "{\"filter\":[]}"
         },
-        "savedSearchId": "ef9da9b0-776c-11e7-9e03-393c0d1aa325",
+        "savedSearchId": "PostgreSQL All Logs",
```
tsg: Just for my understanding, were the changes in the vis and searches IDs also necessary? Or would have the index pattern change be enough, and this is only for consistency?

kvch: Index pattern change was not enough. It was necessary to change the ids otherwise searches and visualizations couldn't be located. Also, I checked the working dashboards and those JSONs contain similar IDs. I wanted to be consistent with the existing correct dashboard JSONs.

PR #4909, opened 2017-08-15
Update and reorganize system module config docs

andrewkroh:
Move configuration options from module page to their respective metricset pages.

- Add `process.include_cpu_ticks` to reference config file (it was missing).
- Rename Darwin to macOS in list of supported operating systems.
- Add docs for `core.metrics` and `cpu.metrics`.

Needs backport to 6.0 branch.

### DIFF
--- a/metricbeat/docs/modules/system.asciidoc
+++ b/metricbeat/docs/modules/system.asciidoc
@@ -5,67 +5,8 @@ This file is generated! See scripts/docs_collector.py
 [[metricbeat-module-system]]
 == System module
 
-The System module allows you to monitor your servers. Because the System module always applies to the local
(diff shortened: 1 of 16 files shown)


dedemorton:
There are a couple of other metricsets (filesystem and fsstat) where this statement might be needed:

Configuration
There are no configuration options for this metricset.

Otherwise, LGTM.

andrewkroh:
@dedemorton  Good eye, but actually I omitted this from filesystem and fsstats to avoid conflicts with https://github.com/elastic/beats/pull/4906 which adds the "Configuration" section for those two metricsets. After that one merges we will be consistent.

PR #4911, opened 2017-08-16
Update latest testing environment to 6.0.0-beta1

ruflin:


tsg:
Isn't "latest" currently 5.5? :)

ruflin:
See my small "trick" I did here: https://github.com/elastic/beats/pull/4911/files#diff-a3357c31ae0657b2202b2025c3aeb425L1 

When I wanted to test 6.0.0-beta1 I realised we don't have a file for that. So my conclusion was that `latest.yml` is the latest build we released and we should have a file `stable.yml` which would be 5.5. Problem is most of the time `latest` and `stable` are the same. The problem is kind of solved as we also have for the moment a `5.x.yml` which is `stable`. My conclusion was switching `latest` here is ok :-)

tsg:
jenkins, test it

Issue #4901, opened 2017-08-15
Remove _all settings in the 5.x version of the template


tsg:
In the 5.x version of the template we currently have:

```
"_all": { "norms": false}
```

This breaks after upgrading ES to 6.0, without upgrading the template.

~~We should replace it with:

```
"_all": { "enabled": false }
```

Disabling `_all` is fine, because ES 5.X supports searching all fields without the need of an `_all` field.~~

We should do this in both 6.0 and 5.6 branches. This will help with the upgrade process.

* [x] 5.6 branch #4902
* [x] master + 6.0 branch #4928 
* [x] Backport #4928 to 6.0 #4948
* [x] Include notice about the breaking change regarding the _all field in 6.0 #4947

tsg:
There's an issue with disabling `_all`:

With `_all` enabled, searching by one word of a `keyword` field was working, because all fields were indexed as text in `_all`. With `_all` disabled, one would have to search by the full field value, individual words don't match any document.

This leads to two issues that we should discuss:

* If we disable the `_all` fields in 5.6, it's a breaking change

* The `_all` option doesn't exist in 6.0, we'd need to explicitly simulate it using `copy_to` to a single text field. 

At the very minimum, we'll have to include a line in Breaking changes about this change in behavior.

tsg:
The new plan, due to the issue above, is to remove the `_all` settings from the 5.6 version of the template. This costs a byte per document, but it's the safest approach.

I'm updating the PRs to reflect this decision.

PR #4928, opened 2017-08-16
Remove the _all field settings from the 5.x template (master/6.0 version)

tsg:
Part of #4901.

This also updates the `5.x.yml` testing env to use 5.5.1.


tsg:
Updated PR to remove the _all settings, rather than explicitly disable the field.

tsg:
Should be read for reviews again.

PR #4930, opened 2017-08-17
[Docs] Add module config reload docs

dedemorton:


filebeat/docs/reload-configuration.asciidoc
```diff
@@ -47,6 +62,50 @@ definitions. For example:
...
+[source,yaml]
+------------------------------------------------------------------------------
+filebeat.config.modules:
+  enabled: true
+  path: configs/*.yml
+  reload.enabled: true
+  reload.period: 10s
+------------------------------------------------------------------------------
+
+TIP: If you are using the <<modules-command,`modules`>> command to enable and
```
dedemorton: @exekias I'm not sure if I'm going far enough here. I think it might be better to show the modules.d path here. WDYT? (still need to write those docs...tomorrow)

exekias: Yeah I think so, we show a common path from doc, while it's clear you can change it if you want

filebeat/docs/reload-configuration.asciidoc
```diff
@@ -47,6 +62,50 @@ definitions. For example:
...
+disable configurations, you can enable config reloading for the `modules.d`
+directory by specifying `path: ${path.config}/modules.d/*.yml`.
+
+Each file found by the Glob must contain a list of one or more module
+definitions. For example:
+
+//REVIEWERS: I got an error when I tried to have an empty line before the -module: mysql entry below. Shouldn't whitespace be allowed there?
```
exekias: Uhm I think it should be allowed so we may file a bug for this?

dedemorton: Weird. I was not able to reproduce this issue. Maybe I had a tab or something on that line? Anyhow...if I can't reproduce the problem, it doesn't exist, right? :-)

PR #4931, opened 2017-08-17
[Docs] Add option to log messages in JSON

dedemorton:
Also added the `logging` namespace to (mostly) match what users see in the config file. There are a few places where we use indentation, but I think it's better to show the full path so there is no ambiguity.

This adds docs for PR #4523

### DIFF
--- a/auditbeat/auditbeat.reference.yml
+++ b/auditbeat/auditbeat.reference.yml
@@ -790,5 +790,9 @@ logging.files:
   # Number of rotated log files to keep. Oldest files will be deleted first.
   #keepfiles: 7
 
+  # The permissions mask to apply when rotating log files. The default value is 0600.
+  # Must be a valid Unix-style file permissions mask expressed in octal notation.
+  #permissions: 0600
+
 # Set to true to log messages in json format.
 #logging.json: false
--- a/libbeat/docs/loggingconfig.asciidoc
+++ b/libbeat/docs/loggingconfig.asciidoc
@@ -139,13 +139,12 @@ in the range of 2 to 1024 files.
 [float]
 ==== `logging.files.permissions`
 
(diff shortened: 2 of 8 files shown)


libbeat/docs/loggingconfig.asciidoc
```diff
@@ -107,37 +107,39 @@ metrics and for this reason they are also not documented.
...
 
 The number of most recent rotated log files to keep on disk. Older files are
 deleted during log rotation. The default value is 7. The `keepfiles` options has to be
 in the range of 2 to 1024 files.
 
 [float]
-==== `files.permissions`
+==== `logging.files.permissions`
+
+//REVIEWERS: Noticed that his is not in the beat.reference.yml file. Should it be? Or is the option old?
```
andrewkroh: Yeah, it should be in there too. Can you please add it to [config.reference.yml](https://github.com/elastic/beats/blob/e1060ff722e8f68691ed646599dda575945e5efa/libbeat/_meta/config.reference.yml#L736-L749) and run `make update` from the top-level?

dedemorton: @andrewkroh Added. Can you review the changes? Thanks!

auditbeat/auditbeat.reference.yml
```diff
@@ -790,5 +790,9 @@ logging.files:
   # Number of rotated log files to keep. Oldest files will be deleted first.
   #keepfiles: 7
 
+  # The permissions mask to apply when rotating log files. The default value is 0600.
```
andrewkroh: This one is different than the others. Perhaps you didn't `git add` the final version of this file before committing?

dedemorton: @andrewkroh Good catch! Weirdness...I committed by using `git commit -a -m ...` so I don't know how it's possible that the files weren't updated/committed correctly. Ah well...it's fixed now.

PR #4933, opened 2017-08-17
Rename the beatname.sh script to beatname

tsg:
This makes it a bit less weird when using the commands on Debian/Centos systems.


tsg:
jenkins, package it

tsg:
jenkins, package it

PR #4935, opened 2017-08-17
Add kafkabeat to community beats doc

makkes:
It seems the dearcode/kafkabeat code is not actively maintained. I contacted the author but got no response. So we implemented our own kafkabeat which is actively maintained and properly documented.


monicasarbu:
Thank you @makkes for the Kafkabeat.

makkes:
Should I open another PR for the 5.5/5.6/6 branches or how is this handled?

Issue #3404, opened 2017-01-19
packetbeat: packetbeat should quit only after all events have been published when reading from pcap file


moooofly:
When I used `packetbeat` to analyse pcap files for redis protocol, I found that the output value of "*libbeat.publisher.published_events*" changing every time. After some discussion with *steffens* (https://discuss.elastic.co/t/why-packetbeat-generates-result-of-libbeat-publisher-published-events-changing-every-time/71699), the conclusion comes to "The pcap reading mode is mostly used for debugging/testing purposes, that's why there is no special 'blocking' support in the queue.". But what I need is exactly that making sure packetbeat quit only after all events have been published. 

Would like to see this feature added. Thanks!

Some information :
- Version: packetbeat; Version: 6.0.0-alpha1
- Operating System: 3.19.0-15-generic 15-Ubuntu SMP x86_64 GNU/Linux
- Steps to Reproduce: see this [link](https://discuss.elastic.co/t/why-packetbeat-generates-result-of-libbeat-publisher-published-events-changing-every-time/71699/6).

urso:
Agreed, if we read from pcap, there is no need to drop events.

Plus I'd like to get rid of `-waitstop` and replace it with `shutdown_timeout` as we have it in filebeat (e.g. stop if all events have been published, but wait up to N seconds to publish queue - value of 0 means wait till stopped). This should help speeding up the tests somewhat + it's easier to write new protocol tests from the beginning.

McStork:
I need this feature so much too!

PR #4939, opened 2017-08-17
Delay device opening

urso:
Reimplement the sniffer to delay:
- opening the sniffing device until sniffer.Run is executed.
- do some preliminary config validation on New only
- split file handling into separate sniffer handle type, to clean up the
  run loop a little

This change should help with packetbeat commands - other than run - complaining about not having permissions to open a device.

### DIFF
--- a/packetbeat/sniffer/file.go
+++ b/packetbeat/sniffer/file.go
@@ -53,6 +53,8 @@ func (h *fileHandler) ReadPacketData() ([]byte, gopacket.CaptureInfo, error) {
 		}
 
 		h.pcapHandle.Close()
+		h.pcapHandle = nil
+
 		h.loopCount++
 		if h.loopCount >= h.maxLoopCount {
 			return data, ci, err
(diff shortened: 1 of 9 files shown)


packetbeat/sniffer/file.go
```diff
@@ -0,0 +1,106 @@
...
+	data, ci, err := h.pcapHandle.ReadPacketData()
+	if err != nil {
+		if err != io.EOF {
+			return data, ci, err
+		}
+
+		h.pcapHandle.Close()
+		h.pcapHandle = nil
+
+		// TODO: Remove me.  still required to keep system tests working.
```
andrewkroh: Would adding `-waitstop Xms` to the system test achieve a similar result?

urso: waitstop is in seconds + I have had to add it to every test. The new publisher pipeline supports some wait stop behavior internally, but it's not used/exposed by any beat yet.

The `-waitstop` setting seems to be removed, in favor of a config file setting. I can try setting it to 300ms by default. But I didn't find this doesn't help much with test stability either. Ultimately we want #3404 (which should be possible with new publisher pipeline).

urso:
I manually tested a few sub comments with and without root on OS X and linux. Also tested af_packet and pcap on linux still working.

This change guarantees `packetbeat test config`, `packetbeat devices`, `packetbeat export config` working without requiring rights on the device to be used by `run`. Drawback is, the `test config` does not check/validate the device actually exists.

tsg:
jenkins, test it

PR #4945, opened 2017-08-18
Let docs build fail if not index.asciidoc found

ruflin:
Currently if the index.asciidoc is not found, the doc build just carries on without an error. With this change in case the index file is missing, the doc build will fail. This should not be a problem as all beats have the index file.


ruflin:
The CI server has been changed to have GOPATH as environment variable. It seems this existed in the past but is missing on the new slaves.

tsg:
The docs part of the tests are green.

PR #4956, opened 2017-08-20
Adapt packetbeat tcp protocol generator to new publisher pipeline

urso:
- Update to the tcp protocol generator to work with changes to the publisher pipeline in beats
- minor fixes in readme


urso:
Will finally add some code generator integration test in another PR.

PR #4961, opened 2017-08-21
Add make update to make system tests

ruflin:
More and more system tests rely on fields updated and generated by make update. This adds `make update` as dependency to `make system-tests`.

### DIFF
--- a/filebeat/Dockerfile
+++ b/filebeat/Dockerfile
@@ -8,6 +8,7 @@ RUN set -x && \
     apt-get clean
 
 RUN pip install --upgrade setuptools
+RUN pip install PyYAML datetime
 
 # Setup work environment
 ENV FILEBEAT_PATH /go/src/github.com/elastic/beats/filebeat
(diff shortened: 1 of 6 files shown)


exekias:
Why did you remove `python-env`?

ruflin:
Because it's already part of `make update`.

exekias:
There is an error in travis I never saw before: https://travis-ci.org/elastic/beats/jobs/266817290#L1340, do you have any idea of what's going on?

ruflin:
Also never saw an `rsync` error before. I assume that was a travis internal issue.

ruflin:
@exekias Taking that back. Same error happens on `jenkins` with filebeat: `bash: rsync: command not found`

ruflin:
The problem seems to be that we use `rsync` in the filebeat `make modules` which is called by `collect -> update`. @tsg Is this actually needed as part of `make update` or is it only needed for packaging? If yes, could it go under `before-build`?

https://github.com/elastic/beats/blob/master/filebeat/Makefile#L33

tsg:
@ruflin I *think* that can be moved to the packaging setup, but how come rsync is suddenly no longer available? We do call `make update` from all CI systems, right?

ruflin:
I think the issue is that we run it for the first time inside a container. I updated the filebeat Dockerfile to see if this makes it working again. This would also make the change more obvious.

ruflin:
This was initially intended as a quick fix but has lead to strange behaviour on Travis. Closing this for now.

PR #4962, opened 2017-08-21
Run only the required containers when testing

exekias:
This PR allows tests to define what services from `docker-compose.yml` they need. Instead of launching all at the beginning, the new framework takes care of start/stop the right services on demand.

Python tests can use `COMPOSE_SERVICES` field to define the list of needed services. Go tests can use `compose.EnsureUp` to do the same.

Changes were applied only to Metricbeat, as it's the one with most docker services, we can progressively update the rest once this one goes in.


ruflin:
Great idea. Do you have already in mind how this could work for the Golang tests?

exekias:
I'm implementing something on Go for this

metricbeat/docker-compose.yml
```diff
@@ -1,81 +1,84 @@
...
+      - ./module/postgresql/_meta/env
+      - ./module/prometheus/_meta/env
+      - ./module/rabbitmq/_meta/env
+      - ./module/redis/_meta/env
+      - ./module/zookeeper/_meta/env
 
   # Modules
   aerospike:
-    build: ${PWD}/module/aerospike/_meta
+    build: ./module/aerospike/_meta
```
ruflin: Do we still need this global compose file or could we just have a local one in the `_meta` directory? There are also the `ports` in the env file available.

exekias: what I like from the global compose file is that it's handy to keep your scenario clean. In case of error or cancellation, I can do `docker-compose stop` and kill all containers. If we split this across modules I may need to go over all subfolders

libbeat/scripts/Makefile
```diff
@@ -154,27 +162,28 @@ unit: ## @testing Runs the unit tests without coverage reports.
 .PHONY: integration-tests
 integration-tests: ## @testing Run integration tests. Unit tests are run as part of the integration tests.
 integration-tests: prepare-tests
+	rm docker-compose.yml.lock || true
```
ruflin: should we add this also to `clean`?

libbeat/tests/compose/compose.go
```diff
@@ -0,0 +1,238 @@
...
+
+	for _, service := range services {
+		err := compose.Wait(service)
+		if err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	// Return down function
+	return func() {}
```
ruflin: Is this empty on purpose? Should we at least log something here?

libbeat/tests/compose/compose.go
```diff
@@ -0,0 +1,238 @@
...
+		Create: options.Create{
+			ForceBuild: true,
+		},
+	}, service)
+}
+
+// Ensure all wanted services are healthy. Wait loop (60s timeout)
+func (c *composeProject) Wait(service string) error {
+	healthy := false
+	seconds := 60
```
ruflin: Can this be overwritten? I can see that this could be too slow for Kibana on some slow machines.

libbeat/tests/compose/compose.go
```diff
@@ -0,0 +1,238 @@
...
+
+		if s.Old {
+			c.Kill(s.Name)
+		}
+	}
+
+	return nil
+}
+
+// Adquire the lock (30s) timeout
```
ruflin: s/Adquire/Aquire

Probably comment should start with `Lock`to make code style happy.

metricbeat/docker-compose.yml
```diff
@@ -1,132 +1,83 @@
...
 
   dropwizard:
-    build: ${PWD}/module/dropwizard/_meta
+    build: ./module/dropwizard/_meta
 
   elasticsearch:
-    extends:
-      file: ${ES_BEATS}/testing/environments/${TESTING_ENVIRONMENT}.yml
-      service: elasticsearch
+    build: ./module/elasticsearch/_meta
```
ruflin: Not critical but the nice thing about having only one place where ES was defined is that there was only 1 place to have it update and all the special configs below.

metricbeat/module/aerospike/_meta/Dockerfile
```diff
@@ -1,4 +1,4 @@
 FROM aerospike:3.9.0
 
 RUN apt-get update && apt-get install -y netcat
-HEALTHCHECK CMD nc -z localhost 3000
+HEALTHCHECK --interval=3s --retries=30 CMD nc -z localhost 3000
```
ruflin: any reason for setting this to 3s instead of one? Every second would probably make the tests quicker.

metricbeat/module/apache/_meta/Dockerfile
```diff
@@ -1,4 +1,4 @@
 FROM httpd:2.4.20
 RUN apt-get update && apt-get install -y curl
-HEALTHCHECK CMD curl -f http://localhost
+HEALTHCHECK --interval=2s --retries=30 CMD curl -f http://localhost
```
ruflin: any specific reason for 2s? See also above.

metricbeat/module/aerospike/namespace/namespace_integration_test.go
```diff
@@ -5,11 +5,14 @@ package namespace
 import (
 	"testing"
 
+	"github.com/elastic/beats/libbeat/tests/compose"
 	mbtest "github.com/elastic/beats/metricbeat/mb/testing"
 	"github.com/elastic/beats/metricbeat/module/aerospike"
 )
 
 func TestData(t *testing.T) {
+	defer compose.EnsureUp(t, "aerospike")()
```
ruflin: I was first confused by this as I was "should up not be run first"? But now I see that `EnsureUp` returns a function which is executed at the end and `EnsureUp` already here, right?

metricbeat/module/ceph/_meta/Dockerfile
```diff
@@ -3,6 +3,5 @@ FROM ceph/demo:tag-build-master-jewel-centos-7
 ENV MON_IP 0.0.0.0
 ENV CEPH_PUBLIC_NETWORK 0.0.0.0/0
 
-RUN yum install -y nc && yum clean all
-HEALTHCHECK CMD nc -w 1 -v 127.0.0.1 5000 </dev/null
+HEALTHCHECK --interval=2s --retries=30 CMD ceph --status | grep HEALTH_OK
```
ruflin: `2s` -> now that I see this more often, I assume there is a reason :-)

metricbeat/tests/system/test_docker.py
```diff
@@ -12,24 +12,13 @@ def test_container_fields(self):
         """
         test container fields
         """
-
-        test_env = os.environ.get('DOCKER_COMPOSE_PROJECT_NAME')
```
ruflin: remove on purpose?

libbeat/tests/compose/compose.go
exekias: In previous iterations I did a kill after every test, long story short: `go test` forkss its process by package to run tests, so it was a concurrency mess. Current solution uses the docker-compose lock file and delayed teardown, I will get rid of the `defer` hack.

libbeat/tests/compose/compose.go
exekias: Sounds good to me, will add a parameter to override that, ES + Kibana may take longer indeed

metricbeat/docker-compose.yml
exekias: I know :(, basically `libcompose` doesn't support that syntax, it seems it has been deprecated for version 3 format too

metricbeat/module/aerospike/_meta/Dockerfile
exekias: I can unify this to something like `--interval=1s --retries=60`, WDYT?

metricbeat/docker-compose.yml
ruflin: yes, which is the reason we are still stuck with 2.1 👎

metricbeat/module/aerospike/_meta/Dockerfile
ruflin: SGTM

metricbeat/tests/system/test_docker.py
exekias: Yes, locally I get: `DOCKER_COMPOSE_PROJECT_NAME=metricbeatsnapshot-noxpack7.0.0-alpha111d177efe455c1f207da557ad4e2f2f3f1981c53`. But docker-compose strips down some chars and it ends up in: `metricbeatsnapshotnoxpack700alpha111d177efe455c1f207da557ad4e2f2f3f1981c53`.

Of course, I could strip them down, but this test doesn't really care about what containers it is retrieving, as long as it gets some

exekias:
jenkins, retest this

exekias:
jenkins, retest this, please

exekias:
Looks like metricbeat testing is mostly stable, I have to reivew some other failing tests:

Fix libbeat testsuite in travis: https://travis-ci.org/elastic/beats/jobs/269351149
Also in jenkins-win: https://beats-ci.elastic.co/job/elastic+beats+pull-request+multijob-windows/1346/beat=libbeat,label=windows/console
And jenkins linux: https://beats-ci.elastic.co/job/elastic+beats+pull-request+multijob-linux/1347/beat=libbeat,label=ubuntu/console
(all the same)

exekias:
Full green on travis: https://travis-ci.org/elastic/beats/builds/269546243 and jenkins linux: https://beats-ci.elastic.co/job/elastic+beats+pull-request+multijob-linux/1357/ :tada: 

Only pending issue is on windows: https://beats-ci.elastic.co/job/elastic+beats+pull-request+multijob-windows/1356/beat=libbeat,label=windows/console

Here: https://gist.github.com/andrewkroh/6f30b6f299878c1d1730a2e9087f73e8#file-jenkins_ci-ps1-L57, @andrewkroh do you think we can add `go get github.com/docker/libcompose` to that script? I think I cannot open PRs to a gist

exekias:
@ruflin this should be ready for a second review (the only pending issue is the windows one mentioned on previous comment). I've also created https://github.com/elastic/beats/issues/5047 as a follow up for the rest of beats (which has less issues & less priority)

libbeat/scripts/Makefile
```diff
@@ -137,8 +139,10 @@ prepare-tests:
 	mkdir -p ${COVERAGE_DIR}
 	# gotestcover is needed to fetch coverage for multiple packages
 	go get ${COVERAGE_TOOL_REPO}
-	# testify is needed to for unit and integration tests
+	# testify is needed for unit and integration tests
 	go get ${TESTIFY_TOOL_REPO}
+	# libcompose is needed for integration tests
+	go get ${LIBCOMPOSE_TOOL_REPO}
```
ruflin: Could we add to the follow up issue to research methods on how this could be vendored or at least versioned. Because I worry that over time some builds will stop working because of an incompatible version. I was hoping `govendor get ...@1.7.2` would do something like this, but this will put it into the vendor folder instead of the global one.

libbeat/tests/compose/compose.go
```diff
@@ -0,0 +1,249 @@
...
+	containers, err := c.p.Containers(context.Background(), project.Filter{State: project.Running}, filter...)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, c := range services {
+		name := strings.Split(c["Name"], "_")[1]
+		// In case of several (stopped) containers, always prefer info about running ones
+		if result[name] != nil {
+			if result[name].Running {
```
ruflin: You comment above that you are looking for the running containers but here you seem to skip the running once?

testing/environments/docker/kafka/Dockerfile
```diff
@@ -1,3 +1,3 @@
 FROM spotify/kafka
-RUN apt-get update && apt-get install -y netcat
-HEALTHCHECK CMD nc -z localhost 9092
+
+HEALTHCHECK --interval=2s --retries=30 CMD /opt/kafka_2.11-0.10.1.0/bin/kafka-topics.sh --zookeeper 127.0.0.1:2181 --list
```
ruflin: 2s interval? ;-)

libbeat/tests/compose/compose.go
exekias: I'm checking in the result array, if it already has a Running container we ignore the current one (`c`)

libbeat/scripts/Makefile
exekias: https://github.com/elastic/beats/issues/5050 feel free to update the TODO there

andrewkroh:
I modified the jenkins.ps1 script to `go get github.com/docker/libcompose`.

https://gist.github.com/andrewkroh/6f30b6f299878c1d1730a2e9087f73e8#file-jenkins_ci-ps1-L57-L58

exekias:
jenkins, retest this please

ruflin:
@exekias Awesome. Really great to have it in.

PR #4964, opened 2017-08-22
Added cloud.id and cloud.auth config settings

### LINKED ISSUE #4959: Support CLI flag shortcuts for Cloud configuration
To simplify the usage of Beats with Elastic Cloud, this proposes the addition of two new CLI flags:

```
   -E cloud.id=<cloud-id>
   -E cloud.auth"<user:pass>
```

These options are CLI only, but they overwrite settings in the configuration file.

Starting with `cloud.auth` because it's simpler, the username and password are used to overwrite `output.elasticsearch.username` and `output.elasticsearch.password`. Using the new flag is equivalent with doing `-E output.elasticsearch.username=<user> -E output.elasticsearch.password=<pass>`, but shorter.

The cloud-id, to be passed to the `cloud.id` setting, looks something like this:

`staging:dXMtZWFzdC0xLmF3cy5mb3VuZC5pbyRjZWM2ZjI2MWE3NGJmMjRjZTMzYmI4ODExYjg0Mjk0ZiRjNmMyY2E2ZDA0MjI0OWFmMGNjN2Q3YTllOTYyNTc0Mw==`

This can be decomposed using the following algorithm:

1. Everything up to the first `:` is ignored. This is an optional user friendly token that can be used to more easily identify the cloud ID. If no `:` is present, then the IS is passed as is to the next step.

2. The rest of the ID is base64 decoded:

     `us-east-1.aws.found.io$cec6f261a74bf24ce33bb8811b84294f$c6c2ca6d042249af0cc7d7a9e9625743`

3. There are three parts, separated by `$`, in order:

     host: `us-east-1.aws.found.io`
     es-id: `cec6f261a74bf24ce33bb8811b84294f`
     kibana-id: `c6c2ca6d042249af0cc7d7a9e9625743`

4. The `es-id` and `kibana-id` are parsed as URLs, and `host` is appended to their host parts. The `https` scheme and port `443` are added:

      es-url: `https://cec6f261a74bf24ce33bb8811b84294f.us-east-1.aws.found.io:443`
      kibana-url: `https://c6c2ca6d042249af0cc7d7a9e9625743.us-east-1.aws.found.io:443`

The resulting `es-url` and `kibana-url` are used to overwrite the settings `output.elasticsearch.hosts` and `setup.kibana.hosts`.

TODOs:

* [ ] Add new config settings #4964 
* [ ] Create a "Using Beats with Elastic Cloud" guide

tsg:
This adds two new configuration settings: `cloud.id` and `cloud.auth`.
They can be used in the configuration file, or from the CLI like this:

```
./metricbeat -e -E cloud.id=<cloud-id> -E cloud.auth=<cloud-auth>
```

The ES/KB settings are changed via overwritting, but we display an INFO
level message that the settings are changed by the cloud-id.

Closes #4959.

Left TODOs:

* [x] more tests
* [x] docs for the new settings
* [x] changelog

### DIFF
--- a/filebeat/filebeat.reference.yml
+++ b/filebeat/filebeat.reference.yml
@@ -597,7 +597,7 @@ filebeat.prospectors:
 
 #============================= Elastic Cloud ==================================
 
-# These settings simplify using Beatname with the Elastic Cloud (https://cloud.elastic.co/).
+# These settings simplify using filebeat with the Elastic Cloud (https://cloud.elastic.co/).
 
 # The cloud.id setting overwrites the `output.elasticsearch.hosts` and
 # `setup.kibana.host` options.
--- a/filebeat/filebeat.reference.yml
+++ b/filebeat/filebeat.reference.yml
@@ -595,6 +595,19 @@ filebeat.prospectors:
 #processors:
 #- add_docker_metadata: ~
 
+#============================= Elastic Cloud ==================================
+
+# These settings simplify using Beatname with the Elastic Cloud (https://cloud.elastic.co/).
+
+# The cloud.id setting overwrites the `output.elasticsearch.hosts` and
+# `setup.kibana.host` options.
+# You can find the `cloud.id` in the Elastic Cloud web UI.
+#cloud.id:
+
+# The cloud.auth setting overwrites the `output.elasticsearch.username` and
+# `output.elasticsearch.password` settings. The format is `<user>:<pass>`.
+#cloud.auth:
+
 #================================ Outputs ======================================
 
 # Configure what outputs to use when sending the data collected by the beat.
--- a/libbeat/cloudid/cloudid.go
+++ b/libbeat/cloudid/cloudid.go
@@ -8,9 +8,10 @@ import (
 	"net/url"
 	"strings"
 
+	"github.com/pkg/errors"
+
 	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/logp"
-	"github.com/pkg/errors"
 )
 
 // OverwriteSettings modifies the received config object by overwriting the
--- a/libbeat/cloudid/cloudid.go
+++ b/libbeat/cloudid/cloudid.go
@@ -47,13 +47,14 @@ func OverwriteSettings(cfg *common.Config) error {
 	}
 
 	// Before enabling the ES output, check that no other output is enabled
-	outputCfg, _ := cfg.Child("output", -1)
-	if outputCfg != nil {
-		var namespace common.ConfigNamespace
-		outputCfg.Unpack(&namespace)
-		if namespace.Name() != "" && namespace.Name() != "elasticsearch" {
-			return errors.Errorf("The cloud.id setting enables the Elasticsearch output, but you already have the %s output enabled in the config", namespace.Name())
-		}
+	tmp := struct {
+		Output common.ConfigNamespace `config:"output"`
+	}{}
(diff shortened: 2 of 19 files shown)


tsg:
@dedemorton I added short docs descriptions for the new settings in the Output section, together with an example, but it might make sense (for another PR) to have a dedicated "Getting started with Elastic Cloud" guide.

filebeat/filebeat.reference.yml
```diff
@@ -595,6 +595,19 @@ filebeat.prospectors:
 #processors:
 #- add_docker_metadata: ~
 
+#============================= Elastic Cloud ==================================
+
+# These settings simplify using Beatname with the Elastic Cloud (https://cloud.elastic.co/).
```
exekias: It says Beatname in all files

libbeat/cloudid/cloudid_test.go
```diff
@@ -0,0 +1,189 @@
...
+}
+
+func TestOverwriteErrors(t *testing.T) {
+	tests := []struct {
+		name   string
+		inCfg  map[string]interface{}
+		errMsg string
+	}{
+		{
+			name: "cloud.auth specified by cloud.id not",
```
exekias: I think there is a typo here: s/by/but/

libbeat/cloudid/cloudid.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	logp.Info("Setting Elasticsearch and Kibana URLs based on the cloud id: output.elasticsearch.hosts=%s and setup.kibana.host=%s", esURL, kibanaURL)
+
+	esURLConfig, err := common.NewConfigFrom([]string{esURL})
+	if err != nil {
+		return err
+	}
+
+	err = cfg.SetChild("output.elasticsearch.hosts", -1, esURLConfig)
```
urso: before setting anything in `output.elasticsearch.X`, check the config is actually enabled/available via:

```
esConfig, _ := cfg.GetChild("output.elasticsearch")
esEnabled = esConfig.Enabled()

kibConfig, _ := cfg.GetChild("setup.kibana")
kibEnabled = kibConfig.Enabled()
```

By setting values on a non-present section, you will automatically enable the section (`enabled` always defaults to true).

tsg: As discussed, I added a check that by auto-enabling the ES output we don't enable two outputs.

libbeat/cloudid/cloudid.go
```diff
@@ -46,6 +46,16 @@ func OverwriteSettings(cfg *common.Config) error {
...
+	// Before enabling the ES output, check that no other output is enabled
+	outputCfg, _ := cfg.Child("output", -1)
+	if outputCfg != nil {
+		var namespace common.ConfigNamespace
+		err = outputCfg.Unpack(&namespace)
+		if namespace.Name() != "" && namespace.Name() != "elasticsearch" {
+			return errors.Errorf("The cloud.id setting enables the Elasticsearch output, but you already have the %s output enabled in the config", namespace.Name())
+		}
+	}
+
```
urso: this can be simplified to:

```
tmp := struct {
  Output common.ConfigNamespace `config:"output"`
}{}
if err := cfg.Unpack(&tmp); err != nil {
  return err
}
if out := tmp.Output; if out.IsSet() && out.Name() != "elasticsearch" {
  return errors.Errorf("...")
}
```

You can also capture the Kibana settings, in case we want to do some validations.

tsg: Thanks, I took your version.

PR #4965, opened 2017-08-22
Cherry-pick #4769 to 6.0: Allow template pattern to be overwritten

ruflin:
Cherry-pick of PR #4769 to 6.0 branch. Original message: 

With the current template loading is was not possible to overload the pattern used as name and pattern were mixed. For example in the case where someone wanted to define pattern name `a` that applies to the indice `a` this was not possible because the pattern generated was `a-*` so it didn't apply to the index `a`. This change now allows to set pattern and name separately.

To make sure index pattern and templates are in sync, the beat will not start in case the index pattern was modified but the template patterns not. We should figure out later some automated mechanisms here.

We should find a better solution for 6.x to require less config options and also include dashboard generation in this.


ruflin:
jenkins, test it

PR #4966, opened 2017-08-22
Move template and codec message to debug level

ruflin:
Both messages show up on startup of the Beat. I do not think they are required on the Info level, so they are moved to the Debug level.

### DIFF
--- a/libbeat/outputs/codec/json/json.go
+++ b/libbeat/outputs/codec/json/json.go
@@ -9,7 +9,6 @@ import (
 
 	"github.com/elastic/beats/libbeat/beat"
 	"github.com/elastic/beats/libbeat/common"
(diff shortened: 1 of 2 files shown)


libbeat/outputs/codec/json/json.go
```diff
@@ -41,7 +41,7 @@ func init() {
 }
 
 func New(pretty bool) *Encoder {
-	logp.Info("load json codec")
+	logp.Debug("codec", "Load json codec")
```
ruflin: @urso objections?

urso: You can remove the message.

ruflin: removed

PR #4973, opened 2017-08-23
[Docs] Document how to use modules.d directory

dedemorton:
Adds docs about `modules.d` for https://github.com/elastic/beats/pull/4537

I didn't go into a lot of detail about how the modules enable/disable commands change the name of the config files because that seems like an implementation detail. But let me think if the layout needs more description. I also reworked some stuff in the other docs because it made sense to do it now.

### DIFF
--- a/filebeat/docs/filebeat-modules-options.asciidoc
+++ b/filebeat/docs/filebeat-modules-options.asciidoc
@@ -17,12 +17,12 @@ Filebeat provides a few different ways to enable modules. You can:
 * <<enable-modules-config-file>>
 
 When you enable modules, you can also
-<<override-variables,set variable overrides>> to change default settings, and
-you can specify <<advanced-settings,advanced settings>> to override prospector
-settings.
+<<specify-variable-settings,specify variable settings>> to change the default
+behavior of the modules, and you can specify
+<<advanced-settings,advanced settings>> to override prospector settings.
 
 Before running Filebeat with modules enabled, make sure you also set up the
-enviroment to use Kibana dashboards. See <<filebeat-modules-quickstart>> for
+environment to use Kibana dashboards. See <<filebeat-modules-quickstart>> for
 more information.
 
 [float]
@@ -54,9 +54,9 @@ To see a list of enabled and disabled modules, run:
 ----
 
 The default module configurations assume that the logs you’re harvesting are
-in the location expected for your OS and that the default behavior of Filebeat
-is appropriate for your environment. To change the default settings, you need
-to specify variable overrides. See <<override-variables>>.
+in the location expected for your OS and that the behavior of the module is
+appropriate for your environment. To change the default configurations, you need
+to specify variable settings. See <<specify-variable-settings>>.
 
 [float]
 [[enable-modules-cli]]
@@ -70,8 +70,6 @@ along with any modules that are enabled in the configuration file or `modules.d`
 directory. If there's a conflict, the configuration specified at the command
 line is used.
 
-//REVIEWERS: Is the above statement (about how Filebeat uses all module configs) true? I tested some combinations, but not all. Please verify.
-
 The following example shows how to enable and run the `nginx`,`mysql`, and
 `system` modules.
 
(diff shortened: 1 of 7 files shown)


filebeat/docs/modules-overview.asciidoc
```diff
@@ -34,9 +34,20 @@ Node.
...
+To learn how to configure and run Filebeat modules:
+
+* Get started by reading <<filebeat-modules-quickstart>>.
+* Try out the extended <<modules-tutorial>> for a more in-depth introduction.
+* Learn about the different ways to enable modules in <<configuration-filebeat-modules>>.
+
 [[modules-tutorial]]
 === Tutorial
 
+//REVIEWERS: Now that we have a section about specifying which modules to run (where I describe how to override settings) plus a Quick start for common log formats, the info in this tutorial is a bit redundant. I do think a more in-depth tutorial that shows a realistic use case would be pretty useful here. It could even show path overrides and prospector settings, but it wouldn't have to describe all the permutations because we do that in the reference section. IMO a streamlined version of this tutorial focused on specific user goals and built around a compelling story would be more useful. If someone can provide me with the "story" for this tutorial, I can write it up. Any volunteers?
```
dedemorton: @tsg Adding you as a reviewer because I'd specifically like your input on this question. I think you wrote the original version of the tutorial.

filebeat/docs/modules-overview.asciidoc
```diff
@@ -148,6 +160,9 @@ cannot install the plugins, you can use the following:
 ./filebeat -e -modules=nginx -M "nginx.access.var.pipeline=no_plugins"
 ----------------------------------------------------------------------
 
+
+//REVIEWERS: Is no_plugins still a thing?
```
tsg: No, that's not an option anymore, we should remove it.

filebeat/docs/modules-getting-started.asciidoc
```diff
@@ -81,20 +84,28 @@ example:
 ----------------------------------------------------------------------
 ./filebeat -e -modules=system,nginx,mysql
```
exekias: Should we update `-modules` to `--modules`? there are several occurrences on this file

filebeat/docs/modules-overview.asciidoc
tsg: Yeah, I agree this is likely not needed anymore. I'm not sure what the overlap between the getting started and the FBM tutorial is at the moment. If the Getting Started focuses on the "add your own prospector" experience, then we could keep this one, maybe simplified.

I'm not sure what sort of story do you have in mind, maybe something like you have a LAMP stack and you monitor linux logs (system module), Apache logs (apache2 module) and Mysql logs (mysql module)?

filebeat/docs/modules-getting-started.asciidoc
dedemorton: Yah, I haven't changed much in the getting started because it sounds like our plans for the getting started experience are still in flux. But since you mentioned it, beta 2 users might find the discrepancy a little confusing, so maybe I should go ahead and change them now.

filebeat/docs/filebeat-modules-options.asciidoc
```diff
@@ -1,28 +1,107 @@
...
+
+[source,shell]
+----
+./filebeat modules list
+----
+
+The default module configurations assume that the logs you’re harvesting are
+in the location expected for your OS and that the default behavior of Filebeat
+is appropriate for your environment. To change the default settings, you need
+to specify variable overrides. See <<override-variables>>.
```
tsg: I would rather recommend here to edit the `modules.d/*.yml` files, instead of using CLI to overwrite settings. The main point of splitting the huge configuration file into smaller ones is that they are easier to edit.

filebeat/docs/filebeat-modules-options.asciidoc
```diff
@@ -1,28 +1,107 @@
...
+
+Enabling <<filebeat-modules,modules>> directly in the config file is a practical
+approach if you have upgraded from a previous version of {beatname_uc} and don't
+want to move your module configs to the `modules.d` directory. You can continue
+to configure modules in the +{beatname_lc}.yml+ file, but you won't be able to
+use the `modules` command to enable and disable configurations because the
+command requires the `modules.d` layout.
+
+To enable specific modules in the +{beatname_lc}.yml+ config file, you can add
+entries to the `filebeat.modules` list. Each entry in the list begins with a
```
tsg: Most of this file uses the `{beatname_lc}` variables, but here `filebeat` is used by the name.

dedemorton: yah, I'm trying to use the variable in places where I think the name could change (like in the name of the config file). Doesn't hurt to use it here, tho, so I've changed it.

filebeat/docs/filebeat-modules-options.asciidoc
dedemorton: I thought the main purpose of the modules.d dir was to make it easier to enable/disable specific configs. But anyhow, the topic this points to does discuss changing the config before it mentions the command-line overrides. I will change the language in this sentence to make this a bit more clear, tho.

PR #4978, opened 2017-08-23
Event ACK callbacks only report the events private field

urso:
- only `beat.Event.Private` field is reported in ACK handlers -> events actual fields can be GC'ed
more early
- update filebeat/winlogbeat to only store and process state updates in the registrar


tsg:
jenkins, test it

tsg:
jenkins, test it

PR #4980, opened 2017-08-23
Add registry for prospectors

kvch:
From now on it is possible to register prospectors in `init` functions, just like in Metricbeat.

Example:
```
func init() {
	err := prospectors.Register("log", NewProspector)
	if err != nil {
		panic(err)
	}
}
```
General `prospector` functions were moved to `prospectors` package (for a lack of a better name). In `prospector.go` new prospectors should be imported to register.

### DIFF
--- a/filebeat/crawler/crawler.go
+++ b/filebeat/crawler/crawler.go
@@ -13,6 +13,8 @@ import (
 	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/common/cfgwarn"
 	"github.com/elastic/beats/libbeat/logp"
+
+	_ "github.com/elastic/beats/filebeat/include"
 )
 
 type Crawler struct {
@@ -56,9 +58,9 @@ func (c *Crawler) Start(r *registrar.Registrar, configProspectors *common.Config
 		cfgwarn.Beta("Loading separate prospectors is enabled.")
 
 		c.prospectorsReloader = cfgfile.NewReloader(configProspectors)
-		prospectorsFactory := prospector.NewFactory(c.out, r, c.beatDone)
+		registrarContext := prospector.NewRegistrarContext(c.out, r, c.beatDone)
 		go func() {
-			c.prospectorsReloader.Run(prospectorsFactory)
+			c.prospectorsReloader.Run(registrarContext)
 		}()
 	}
 
--- a/filebeat/crawler/crawler.go
+++ b/filebeat/crawler/crawler.go
@@ -20,7 +20,7 @@ import (
 type Crawler struct {
 	prospectors         map[uint64]*prospector.Prospector
 	prospectorConfigs   []*common.Config
-	out                 channel.OutleterFactory
+	out                 channel.Factory
 	wg                  sync.WaitGroup
 	modulesReloader     *cfgfile.Reloader
 	prospectorsReloader *cfgfile.Reloader
@@ -29,7 +29,7 @@ type Crawler struct {
 	beatDone            chan struct{}
 }
 
-func New(out channel.OutleterFactory, prospectorConfigs []*common.Config, beatVersion string, beatDone chan struct{}, once bool) (*Crawler, error) {
+func New(out channel.Factory, prospectorConfigs []*common.Config, beatVersion string, beatDone chan struct{}, once bool) (*Crawler, error) {
 	return &Crawler{
 		out:               out,
 		prospectors:       map[uint64]*prospector.Prospector{},
--- a/filebeat/prospector/redis/prospector.go
+++ b/filebeat/prospector/redis/prospector.go
@@ -31,7 +31,7 @@ type Prospector struct {
 }
 
 // NewProspector creates a new redis prospector
-func NewProspector(cfg *common.Config, outletFactory channel.OutleterFactory, context prospector.Context) (prospector.Prospectorer, error) {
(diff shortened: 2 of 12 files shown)


filebeat/crawler/crawler.go
```diff
@@ -7,16 +7,18 @@ import (
...
 	"github.com/elastic/beats/filebeat/input/file"
-	"github.com/elastic/beats/filebeat/prospector"
+	pros "github.com/elastic/beats/filebeat/prospector/prospectors"
 	"github.com/elastic/beats/filebeat/registrar"
 	"github.com/elastic/beats/libbeat/cfgfile"
 	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/common/cfgwarn"
 	"github.com/elastic/beats/libbeat/logp"
+
+	_ "github.com/elastic/beats/filebeat/prospector"
```
urso: in other places we use something like `github.om/elastic/beats/<beatname>/include/list.go`, to build/provide an import list of 'plugins'.

filebeat/prospector/prospectors/registry.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"fmt"
+
+	"github.com/elastic/beats/filebeat/channel"
+	"github.com/elastic/beats/filebeat/input/file"
+	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/logp"
+)
+
+type ProspectorFactoryData struct {
```
urso: how about naming it 'Context' or 'Environment'? 

Keep in mind, the package names is part of a type its full. Try to reduce stuttering if possible. e.g.

`prospectors.ProspectorFactory` can be minimized to `prospectors.Factory`. And `prospectors.ProspectorFactoryData` to `prospectors.Context`.

filebeat/prospector/prospectors/registry.go
```diff
@@ -0,0 +1,45 @@
...
+	States   []file.State
+	Done     chan struct{}
+	BeatDone chan struct{}
+}
+
+type ProspectorFactory func(config *common.Config, outletFactory channel.OutleterFactory, data ProspectorFactoryData) (Prospectorer, error)
```
urso: appropos naming, I will not complain if you rename `channel.OutleterFactory` to `channel.Factory` ;)

urso:
Just a few minor comments. All in all LGTM

filebeat/crawler/crawler.go
ruflin: Here is for example the script that generates the imports for metricbeat: https://github.com/elastic/beats/blob/master/metricbeat/scripts/generate_imports.py

filebeat/prospector/redis/prospector.go
```diff
@@ -49,6 +50,13 @@ func NewProspector(cfg *common.Config, outletFactory channel.OutleterFactory) (*
 	return p, nil
 }
 
+func init() {
```
ruflin: Could we always put the init functions on the top after the imports? I kind of like them at the top as it becomes also obvious they are executed first.

kvch:
I addressed the notes. However, I am not satisfied with the package structure and names, but so far I don't have better idea.

I also added renaming `channel.OutleterFactory` to `channel.Factory`.

filebeat/prospector/include/list.go
```diff
@@ -0,0 +1,8 @@
+package include
```
ruflin: I would suggest to move this directly under `filebeat` instead of `prospectr` as we ave it in other projects.

ruflin:
@kvch There seem to be some issues with the tests. Ping me when you got the build green.

urso:
We still need the prospector and prospectors package? Given the imports moved to `filebeat/include/list.go`, there should be no more circular dependencies.

kvch:
I will remove it \o/

kvch:
I had to rename `Factory` in `factory.go` which is a struct that contains data of the `Registrar`. Its new name is `RegistrarContext`. So outside the package its name is `prospector.RegistrarContext` and the concrete factories of prospectors type can be `prospector.Factory`.

kvch:
jenkins, retest it

kvch:
jenkins, test it

ruflin:
@urso Never saw this Jenkins error before:

```
.....F........S.SS..SS........
FAIL: Checks CLI overwrite actually overwrites some config variable by
  File "C:\Users\jenkins\workspace\elastic+beats+pull-request+multijob-windows\beat\libbeat\label\windows\src\github.com\elastic\beats\libbeat\tests\system\test_base.py", line 57, in test_invalid_config_cli_param
    proc.check_kill_and_wait()
  File "C:\Users\jenkins\workspace\elastic+beats+pull-request+multijob-windows\beat\libbeat\label\windows\src\github.com\elastic\beats\libbeat\tests\system\beat\beat.py", line 89, in check_kill_and_wait
    return self.check_wait(exit_code=exit_code)
  File "C:\Users\jenkins\workspace\elastic+beats+pull-request+multijob-windows\beat\libbeat\label\windows\src\github.com\elastic\beats\libbeat\tests\system\beat\beat.py", line 78, in check_wait
    exit_code, actual_exit_code)
AssertionError: Expected exit code to be 0, but it was 1
```

As the error is not related to this change, I think we can move forward here.

ruflin:
@kvch Great job on this one. Lets now do small interations on top of it to potentially improve handling and even simplify things further.

PR #4981, opened 2017-08-23
add_kubernetes_metadata processor: add support for "/var/log/containers/" log path 

SvenWoltmann:
Following up on this topic:

# The Issue

instead of

# About the Pull Request

The pull requests contains a simple solution, in which the container ID is extracted in a different way when the logs_path matcher's `logs_path` setting is "/var/log/containers/".

So the processor needs to be configured like this:

```
processors:
- add_kubernetes_metadata:
    in_cluster: true
    default_matchers.enabled: false
    matchers:
    - logs_path:
        logs_path: /var/log/containers/
```

However, it seems that the `logs_path` setting is used only in the code that extracts the container ID from the source. Wouldn't it make more sense to extract the log path from the `source` variable? Then the `logs_path` setting can be removed and the processor can be used with it's default configuration:

```
processors:
- add_kubernetes_metadata:
    in_cluster: true
```

### DIFF
--- a/filebeat/processor/add_kubernetes_metadata/indexing.go
+++ b/filebeat/processor/add_kubernetes_metadata/indexing.go
@@ -48,21 +48,26 @@ func newLogsPathMatcher(cfg common.Config) (add_kubernetes_metadata.Matcher, err
 	return &LogPathMatcher{LogsPath: logPath}, nil
 }
 
+// Docker container ID is a 64-character-long hexadecimal string
+const containerIdLen = 64
+
 func (f *LogPathMatcher) MetadataIndex(event common.MapStr) string {
 	if value, ok := event["source"]; ok {
 		source := value.(string)
 		logp.Debug("kubernetes", "Incoming source value: %s", source)
 		cid := ""
 		if strings.Contains(source, f.LogsPath) {
-			if f.LogsPath == "/var/log/containers/" && strings.HasSuffix(source, ".log") {
+			sourceLen := len(source)
+			logsPathLen := len(f.LogsPath)
+
+			if f.LogsPath == "/var/log/containers/" && strings.HasSuffix(source, ".log") && sourceLen >= containerIdLen + 4 {
 				// In case of the Kubernetes log path "/var/log/containers/",
 				// the container ID will be located right before the ".log" ending.
-				sourceLen := len(source)
-				cid = source[sourceLen-68 : sourceLen-4]
-			} else {
+				containerIdEnd := sourceLen - 4
+				cid = source[containerIdEnd - containerIdLen : containerIdEnd]
+			} else if sourceLen >= logsPathLen + containerIdLen {
 				// In any other case, we assume the container ID will follow right after the log path.
-				//Docker container is 64 chars in length
-				cid = source[len(f.LogsPath) : len(f.LogsPath)+64]
+				cid = source[logsPathLen : logsPathLen + containerIdLen]
 			}
 			logp.Debug("kubernetes", "Using container id: %s", cid)
 		} else {
--- a/filebeat/processor/add_kubernetes_metadata/indexing.go
+++ b/filebeat/processor/add_kubernetes_metadata/indexing.go
@@ -66,19 +66,17 @@ func (f *LogPathMatcher) MetadataIndex(event common.MapStr) string {
 
 		// In case of the Kubernetes log path "/var/log/containers/",
 		// the container ID will be located right before the ".log" extension.
-		if strings.HasPrefix(f.LogsPath, "/var/log/containers/") &&
-		   strings.HasSuffix(source, ".log") &&
-		   sourceLen >= containerIdLen + 4 {
+		if strings.HasPrefix(f.LogsPath, "/var/log/containers/") && strings.HasSuffix(source, ".log") && sourceLen >= containerIdLen+4 {
 			containerIdEnd := sourceLen - 4
-			cid := source[containerIdEnd - containerIdLen : containerIdEnd]
+			cid := source[containerIdEnd-containerIdLen : containerIdEnd]
 			logp.Debug("kubernetes", "Using container id: %s", cid)
 			return cid
 		}
 
 		// In any other case, we assume the container ID will follow right after the log path.
 		// However we need to check the length to prevent "slice bound out of range" runtime errors.
-		if sourceLen >= logsPathLen + containerIdLen {
-			cid := source[logsPathLen : logsPathLen + containerIdLen]
+		if sourceLen >= logsPathLen+containerIdLen {
+			cid := source[logsPathLen : logsPathLen+containerIdLen]
 			logp.Debug("kubernetes", "Using container id: %s", cid)
 			return cid
 		}
--- a/filebeat/processor/add_kubernetes_metadata/indexing.go
+++ b/filebeat/processor/add_kubernetes_metadata/indexing.go
@@ -43,19 +43,31 @@ func newLogsPathMatcher(cfg common.Config) (add_kubernetes_metadata.Matcher, err
 		logPath = logPath + "/"
 	}
 
+	logp.Debug("kubernetes", "logs_path matcher log path: %s", logPath)
+
 	return &LogPathMatcher{LogsPath: logPath}, nil
 }
 
 func (f *LogPathMatcher) MetadataIndex(event common.MapStr) string {
 	if value, ok := event["source"]; ok {
 		source := value.(string)
-		logp.Debug("kubernetes", "Incoming source value: ", source)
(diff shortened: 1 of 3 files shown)


exekias:
ok to test

filebeat/processor/add_kubernetes_metadata/indexing.go
```diff
@@ -43,19 +43,31 @@ func newLogsPathMatcher(cfg common.Config) (add_kubernetes_metadata.Matcher, err
...
 func (f *LogPathMatcher) MetadataIndex(event common.MapStr) string {
 	if value, ok := event["source"]; ok {
 		source := value.(string)
-		logp.Debug("kubernetes", "Incoming source value: ", source)
+		logp.Debug("kubernetes", "Incoming source value: %s", source)
 		cid := ""
 		if strings.Contains(source, f.LogsPath) {
-			//Docker container is 64 chars in length
-			cid = source[len(f.LogsPath) : len(f.LogsPath)+64]
+			if f.LogsPath == "/var/log/containers/" && strings.HasSuffix(source, ".log") {
```
exekias: I'm not a fan of hardcoding this kind of things in the code, but I think I can live with this one if anyone else is ok with it

exekias:
Thank you very much for your contribution!! Could you please add a `CHANGELOG.ascii` entry?

filebeat/processor/add_kubernetes_metadata/indexing.go
SvenWoltmann: Me neither, but the existing method already looked like a "quick-and-dirty" solution, so I just continued this way. 

I'm now working on an alternative, more generic solution that ignores the "logs_path" parameter (it's nowhere documented and it doesn't seem to be used anywhere else) and instead parses the source to extract the container ID. Will have that ready in two hours or so. Maybe that's better than this solution.

SvenWoltmann:
I've implemented an alternative cleaner and more generic solution in https://github.com/elastic/beats/pull/4995.

exekias:
I still prefer this one, it covers all default cases and still allows to change the default in case Docker is not under `/var/lib/docker`, do you I mind if I merge this one and close the rest?

SvenWoltmann:
Sure, that's OK. But then it's also necessary to update the documentation at https://www.elastic.co/guide/en/beats/filebeat/master/add-kubernetes-metadata.html with something like this:

> The default matcher detects container IDs only of files in `/var/lib/docker/containers/`. If you read files from `/var/log/containers/` instead, you need to replace the default matcher by a `logs_path` matcher with the `logs_path` set to the appropriate path:
> 
> ```
> processors:
> - add_kubernetes_metadata:
>     in_cluster: true
>     default_matchers.enabled: false
>     matchers:
>     - logs_path:
>         logs_path: /var/log/containers/
> ```

Also, the matcher might still throw a "slice bound out of range" runtime error if `source` is shorter than expected. Would you like me to add an appropriate test and fix to this PR?

exekias:
That would be awesome yes :), also please add an entry to CHANGELOG.asciidoc

SvenWoltmann:
Done :)

Now checking for the `source` length, refactored the tests and added several new tests.

filebeat/processor/add_kubernetes_metadata/indexing.go
```diff
@@ -43,19 +43,36 @@ func newLogsPathMatcher(cfg common.Config) (add_kubernetes_metadata.Matcher, err
...
+				// the container ID will be located right before the ".log" ending.
+				containerIdEnd := sourceLen - 4
+				cid = source[containerIdEnd - containerIdLen : containerIdEnd]
+			} else if sourceLen >= logsPathLen + containerIdLen {
+				// In any other case, we assume the container ID will follow right after the log path.
+				cid = source[logsPathLen : logsPathLen + containerIdLen]
+			}
+			logp.Debug("kubernetes", "Using container id: %s", cid)
+		} else {
+			logp.Debug("kubernetes", "Error extracting container id - source value does not contain log path.")
```
exekias: You will need to refactor this logic after adding length checks, in case of wrong length the error won't be logged at the moment

filebeat/processor/add_kubernetes_metadata/indexing.go
```diff
@@ -43,19 +43,36 @@ func newLogsPathMatcher(cfg common.Config) (add_kubernetes_metadata.Matcher, err
...
-		logp.Debug("kubernetes", "Incoming source value: ", source)
+		logp.Debug("kubernetes", "Incoming source value: %s", source)
 		cid := ""
 		if strings.Contains(source, f.LogsPath) {
-			//Docker container is 64 chars in length
-			cid = source[len(f.LogsPath) : len(f.LogsPath)+64]
+			sourceLen := len(source)
+			logsPathLen := len(f.LogsPath)
+
+			if f.LogsPath == "/var/log/containers/" && strings.HasSuffix(source, ".log") && sourceLen >= containerIdLen + 4 {
```
exekias: `strings.StartsWith(f.LogsPath, "/var/log/containers") would be better here, as this is user input, it may not be exactly `/var/log/containers/`

SvenWoltmann:
Thanks for your comments. I've added another commit to address both issues.

I check if the path starts with `/var/log/containers/` (with a tailing slash) instead of `/var/log/containers`, because if the user omits that tailing slash, it will automatically be added in line 43 (`logPath = logPath + "/"`).

Also I duplicated these two lines, as it makes the code flow much easier to read:
```
logp.Debug("kubernetes", "Using container id: %s", cid)
return cid
```

exekias:
Nice, please review code format (you can run `make fmt`), check is failing here: https://travis-ci.org/elastic/beats/jobs/268351019 and I think we are ready to go

SvenWoltmann:
Sorry... I didn't know about the `fmt` command, it's my first Go project.

One question though:

When formatted, the multi-line `if` statement looks like this:
```
if strings.HasPrefix(f.LogsPath, "/var/log/containers/") &&
	strings.HasSuffix(source, ".log") &&
	sourceLen >= containerIdLen+4 {
	containerIdEnd := sourceLen - 4
	...
```
So you can't easily recognize anymore where the `if` statement ends and the code block starts.

Would you prefer to write the `if` statement into a single (long) line like this?

```
if strings.HasPrefix(f.LogsPath, "/var/log/containers/") && strings.HasSuffix(source, ".log") && sourceLen >= containerIdLen+4 {
	containerIdEnd := sourceLen - 4
	...
```

exekias:
I'm ok with the single line version too

karmi:
Hi @SvenWoltmann, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/beats/pull/4981.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

SvenWoltmann:
@exekias I've committed the formatted code.

@karmi I've added my other e-mail address to my profile. I guess I've used that in an earlier commit, but I just checked, and all my commits now use the e-mail, which I used for the signature.

exekias:
Thank you for adding this feature @SvenWoltmann!

Julion:
Hi Guys, 

I hit this issue in 6.0.0-rc1. It seems that this change has not been applied to the 6.0 branch which I assume is the branch used for the release candidates (please correct me if I'm wrong). Although I understand why it would be considered a new feature, I would argue that since this the processor is beta and the change quite small it might be worth including. This is very confusing that /var/log/containers can't be a source in the current version (fluentd would be fine with it)

exekias:
Hi @Julion, the change went in after the freeze deadline for 6.0, it wasn't backported as it's not a bugfix, should make it to the next release (6.1)

jeremievallee:
Hi Guys,
@SvenWoltmann , thanks for this! We have the same use case and this look exactly like what we need. I'm having trouble around configuration though, and can't find what I need in the documentation. 
Let's say you want filebeat to get the containers logs from Kubernetes, but you would like to exclude some files. I thought this prospector config would be right, but no luck so far:
```
  - type: docker
    containers:
      ids:
        - "*"
      path: "/var/log/containers"
    exclude_files:
      - "/var/log/containers/filebeat*.log"
      - "/var/log/containers/logstash*.log"
    processors:
      - add_kubernetes_metadata:
          in_cluster: true
          default_matchers.enabled: false
          matchers:
          - logs_path:
              logs_path: /var/log/containers/
```
Am I doing something wrong, or is it just not possible at the moment ?
Using `filebeat:6.1.3`, btw.

Thanks!

exekias:
Hi @jeremievallee,

Please for questions use discuss: https://discuss.elastic.co/c/beats

jeremievallee:
Hi @exekias, thanks for your message, I will do that now. 👍

PR #4989, opened 2017-08-24
fixed bug when space in url

mazhan465:
fixed


tsg:
Hi @mazhan465, thanks for the bug fix. Can you sign the [CLA](http://www.elasticsearch.org/contributor-agreement/), please? It should only take a couple of minutes.

We'll review the PR, what would help is a test case that was failing before and is passing now.

ruflin:
Closing as I think this is supersed by https://github.com/elastic/beats/pull/5058.

PR #4990, opened 2017-08-24
Add ceph cluster status to metricbeat

elaron:
https://discuss.elastic.co/t/add-ceph-cluster-status-to-metricbeat/98196

The cluster_status will collect pgmap, osdmap, throughput and iops information of ceph cluster.


metricbeat/module/ceph/cluster_status/data.go
```diff
@@ -0,0 +1,132 @@
...
+type HealthRequest struct {
+	Status string `json:"status"`
+	Output Output `json:"output"`
+}
+
+func eventsMapping(content []byte) []common.MapStr {
+	var d HealthRequest
+	err := json.Unmarshal(content, &d)
+	if err != nil {
+		logp.Err("Error: ", err)
```
andrewkroh: I'm thinking it would be better to return the error from this function. The caller should then check if an error is returned and take the appropriate action (which in this case should be return the error from `Fetch`).

andrewkroh:
jenkins, test it

metricbeat/module/ceph/cluster_status/_meta/fields.yml
```diff
@@ -0,0 +1,105 @@
...
+  fields:
+    - name: version
+      type: long
+      description: >
+        Ceph Status version
+    - name: traffic.read_bytes_sec
+      type: long
+      description: >
+        Cluster read throughput per second
+    - name: traffic.write_bytes_sec
```
ruflin: In general we have the unit which is `bytes` here at the end. For Kibana to make to some "magic" on it you can add `format: bytes` to the yaml file. Same applies for all other bytes values.

metricbeat/module/ceph/cluster_status/data.go
elaron: Hi,  I return the error in the new commit a65dd21, please review it , thanks.

metricbeat/module/ceph/cluster_status/_meta/fields.yml
elaron: write_bytes_sec is a field name used by ceph-rest-api, so I guess it will be better to remain it, since ceph users will be familiar with it. And, in the yaml file, I have add format:bytes and format:percent for corresponding values.

ruflin: That is actually a very interesting topic as we discussed this quite frequently: Should we keep the naming of the services or go with the our convention. We decided to go with our convention as we expect that often people monitor more then one system and having a similar naming scheme for all your metrics is quite nice: https://www.elastic.co/guide/en/beats/libbeat/5.5/event-conventions.html Agree that for people experience with a specific system it could be confusing at first but I think the long term benefit is bigger with a unified structure, which potentially also allows to correlate metrics between metricsets.

elaron: I renamed some fields in commit 2a4e7b0, please check if they are appropriate, thanks :)

andrewkroh:
jenkins, test it

andrewkroh:
Thanks @elaron!

elaron:
My pleasure.

PR #4991, opened 2017-08-24
Cherry-pick #4717 to 6.0: Adds a "type" field to the filesystem beat

### LINKED ISSUE #3459: Feature: Alllow filtering out some filesystem types

tsg:
Cherry-pick of PR #4717 to 6.0 branch. Original message: 

Its value is copied from `sigar.FileSystem.SysTypeName` and passed through so that it becomes possible to filter out `overlay`, `tmpfs` or otherwise uninteresting filesystems.

Fixes #3459.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -140,6 +140,7 @@ https://github.com/elastic/beats/compare/v6.0.0-alpha2...v6.0.0-beta1[View commi
   with the system process metricset. The documentation for these metrics already
   stated that on multi-core systems the percentages could be greater than 100%. {pull}4544[4544]
 - Remove filters setting from metricbeat modules. {pull}4699[4699]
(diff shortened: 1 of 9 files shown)


tsg:
@andrewkroh Just checking, you wanted this one in 6.0, right? Becasue it seems to be somehow related to the `ingore_types` option, which is already backported.

andrewkroh:
It would be nice to have this in 6.0. This change complements the `filesystem.ignore_types` feature nicely, but they are completely independent.

PR #4992, opened 2017-08-24
Cherry-pick #4770 to 6.0: Add http server metricset to support push metrics via http

exekias:
Cherry-pick of PR #4770 to 6.0 branch. Original message: 

This PR allows users to configure a http server in order to send data. The restriction right now it only supports JSON.

Path parameters can be parsed into tags. Ex:
/foo/bar would result in 
```
tag:  {
  foo: "bar",
}
```


exekias:
wrong pr

PR #4995, opened 2017-08-24
add_kubernetes_metadata processor: add support for "/var/log/containers/" log path (generic version)

SvenWoltmann:
Following up on this topic:
and on my previous pull request https://github.com/elastic/beats/pull/4981.

This is a cleaner and more generic solution for the issue.

# The Issue

instead of

# About the Pull Request

The container ID is extracted from the `source` variable via regular expressions in the following order:

1. A regular expression that matches a log file in `/var/log/containers` and returns the container ID
2. A regular expression that matches a log file in `/var/lib/docker/containers` and returns the container ID
3. A generic fallback regular expression that matches any file containing a 64-character-long hexadecimal ID in its name and returns that as container ID

In contrast to my previous pull request https://github.com/elastic/beats/pull/4981, the processor configuration does not need to be changed.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -85,6 +85,7 @@ https://github.com/elastic/beats/compare/v6.0.0-beta1...master[Check the HEAD di
 
 - Add PostgreSQL module with slowlog support. {pull}4763[4763]
 - Add Kafka log module. {pull}4885[4885]
+- Add support for `/var/log/containers/` log path in `add_kubernetes_metadata` processor. {pull}4995[4995]
 
 *Heartbeat*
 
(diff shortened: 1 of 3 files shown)


exekias:
Interesting, did you do any kind of benchmark? I'm inclined to think that the other PR will perform better, and probably most cases are covered by it. What do you think?

SvenWoltmann:
No, I didn't. But when I had the debug output active, I saw that the `MetadataIndex` method was not called often, so I didn't think about performance issues.

I'll dig a bit deeper into the code to check when exactly this method is called. 

What I don't like about the other PR is the additional processor configuration (which seems to be unnecessary as it's used nowhere else).

I could also create a third solution that checks which log path is used with `string.HasPrefix` for example instead of a regexp and then extracts the container ID from its expected position.

SvenWoltmann:
I added a third PR that doens't use regular expressions: https://github.com/elastic/beats/pull/5011

exekias:
Closed in favor of https://github.com/elastic/beats/pull/4981

PR #4999, opened 2017-08-24
fix bug when space in url bug

mazhan465:
fixed bug in packetbeat


karmi:
Hi @mazhan465, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/beats/pull/4999.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

andrewkroh:
Looks like a duplicate of #4989.

PR #5007, opened 2017-08-24
Add Cloudfrontbeat to the community beats

jarl-tornroos:


karmi:
Hi @jarl-tornroos, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/beats/pull/5007.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

jarl-tornroos:
Hi @karmi, I have now added the email address to the Github profile.

tsg:
@jarl-tornroos Thanks for the new Beat.

PR #5011, opened 2017-08-25
add_kubernetes_metadata processor: add support for "/var/log/containers/" log path (version 3)

SvenWoltmann:
Following up on this topic:
and on my previous pull requests https://github.com/elastic/beats/pull/4981 and https://github.com/elastic/beats/pull/4995.

This is the third implementation, more details below in "About the Pull Request".

# The Issue

instead of

# About the Pull Request

This solution
* solves the issue with basic string operations;
* doesn't require the redundant `logs_path` configuration in the processor as in https://github.com/elastic/beats/pull/4981;
* performs better than https://github.com/elastic/beats/pull/4995 as it doesn't use regular expressions;
* also covers most cases,
* and prevents "slice bound out of range" runtime errors that could happen in the original implementation if `source` was too short.

The unit tests are exactly the same as in https://github.com/elastic/beats/pull/4995, except that the generic test case was removed as this solution doesn't have a generic fallback anymore.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -85,6 +85,7 @@ https://github.com/elastic/beats/compare/v6.0.0-beta1...master[Check the HEAD di
 
 - Add PostgreSQL module with slowlog support. {pull}4763[4763]
 - Add Kafka log module. {pull}4885[4885]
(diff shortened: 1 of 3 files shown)


exekias:
As discussed, closed in favor of https://github.com/elastic/beats/pull/4981

PR #5013, opened 2017-08-25
Cherry-pick #4978 to 6.0: Event ACK callbacks only report the events private field

tsg:
Cherry-pick of PR #4978 to 6.0 branch. Original message: 

- only `beat.Event.Private` field is reported in ACK handlers -> events actual fields can be GC'ed
more early
- update filebeat/winlogbeat to only store and process state updates in the registrar

### DIFF
--- a/filebeat/beater/acker.go
+++ b/filebeat/beater/acker.go
@@ -1,9 +1,7 @@
 package beater
 
 import (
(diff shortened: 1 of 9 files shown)


tsg:
@urso I also opened this backport PR, there were no conflicts.

PR #5025, opened 2017-08-25
Add Cloudfrontbeat to the community beats

jarl-tornroos:
Adding the beat also to this branch


tsg:
Thanks for adding it to all branches. Generally it's not needed, though, we can assign the `needs_backport` label and then we will make sure that it gets to the relevant branches.

jarl-tornroos:
Thanks! Didn't know about the `needs_backport`.

PR #5026, opened 2017-08-25
Add prospector plugin && make imports

kvch:
This is the continuation of PR #4980.

This includes two things so far:
1. Prospector plugin, so Filebeat can be extended by external plugins.
2. `make imports` using `generate_imports.py` script
    I am not sure yet, but as multiple beats has a `generate_imports.py`, it might worth it to create one generic import generator script to take care of all beats.

TODO
- [x] Find `RegistrarContext` a better name
- [x] Clean up around `RegistrarContext`
- [x] Figure out if script generalization is worth it


filebeat/Makefile
```diff
@@ -51,6 +51,12 @@ collect-docs: python-env
 	@mkdir -p docs/modules
 	@${PYTHON_ENV}/bin/python ${ES_BEATS}/filebeat/scripts/docs_collector.py --beat ${BEAT_NAME}
 
+# Generate imports for prospectors
+.PHONY: imports
+imports: python-env
```
ruflin: This should be added to a dependency in collect.

ruflin:
I tried to unify the import scripts in the past but I think what made it tricky is the metricbeat import script as it has some specific logic. The thing I wanted to then but never got to it is creating some library functions like that can be reused. One example would be "pass a directory and return list of all go packages", an other one "create valid golang output based a list of package names". Like this I think we could reuse 90% and only have a local script that has the specific logic where the packages are. Perhaps it could go even so far that only "params" have to be passed from the Makefile.

filebeat/Makefile
kvch: donz

kvch:
I generalized `generate_imports.py` and unified all of them in a single script. Now instead of three, only one script gets the job done.

ruflin:
Side note: I'm ok with moving forward as is as this is already a benefit, but I think we should make sure to clean it up in the future to not mix beat specific knowledge and framework. Spent quite a bit of time in the past to clean it up :-)

kvch:
I moved beat specific variables and functions back to its own beat and left the common parts in `scripts/generate_imports.py`.

filebeat/prospector/plugin.go
```diff
@@ -0,0 +1,38 @@
+package prospector
+
+import (
+	"errors"
+
+	"github.com/elastic/beats/libbeat/plugin"
+)
+
+type prospectorPlugin struct {
```
ruflin: Thinking if we should call this just `plugin`  as it's already in the prospector package.

metricbeat/scripts/generate_imports_helper.py
```diff
@@ -0,0 +1,29 @@
...
+from os import listdir
+
+comment = """Package include imports all Module and MetricSet packages so that they register
+their factories with the global registry. This package can be imported in the
+main package to automatically register all of the standard supported Metricbeat
+modules."""
+
+
+def get_importable_lines(go_beat_path):
+    import_line_module = "\t_ \"{beat_path}/module/{module}\""
```
ruflin: Would be nice of the tap and escaping magic would also be done by the global script, so only a list of paths / strings has to be passed.

ruflin:
@kvch Looks like `fmt` is not happy. Best run `make fmt` in the beats directory.

filebeat/prospector/plugin.go
kvch: I was thinking about it to. However, to stay consistent with other with existing plugin.go files I decided to name it `prospectorPlugin`. As other files these `struct`s are called `outputPlugin`, `modulePlugin`, etc.

ruflin: SGTM. Lets keep it in line for now.

ruflin:
@kvch Seems like it breaks the generator tests :-(

ruflin:
@kvch Merged. For the python files I think this is a great step forward in unifying the code. We could potentially move it even to Golang in the future to not depend on it on python. This is now quite a bit easier, as the code is simpler.

PR #5027, opened 2017-08-26
[Docs] Add dashboard upgrade info

dedemorton:
@monicasarbu I'm pretty sure that some of these details are wrong (especially around what to do with the 5.5 dashboards to get them into 5.6). I assume this info is for end-users, so we wouldn't expect them to use devtools.

Anyhow, this is my understanding of the upgrade process based on our discussion, but I think there's some missing content here.

### DIFF
--- a/libbeat/docs/upgrading.asciidoc
+++ b/libbeat/docs/upgrading.asciidoc
@@ -6,6 +6,7 @@ This section gives general recommendations for upgrading the Beats:
 * <<upgrading-minor-versions>>
 * <<upgrading-1-to-5>>
 * <<upgrade-mapping-template>>
+* <<upgrade-5-5-dashboards>>
 
 [[upgrading-minor-versions]]
 === Upgrade between minor versions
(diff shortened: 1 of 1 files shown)


libbeat/docs/upgrading.asciidoc
```diff
@@ -215,3 +216,39 @@ order of operations should be:
...
+provide dashboards that were compatible with each new version of Kibana.
+
+In version 5.6, the Beats dashboards are still loaded by default into the
+`.kibana` index in Elasticsearch, but support has been added for loading
+dashboards via the Kibana API. Regardless of how you import the dashboards
+in version 5.6, they will work when you upgrade to Kibana 6.0. 
+
+However, if you have Beats 5.5 dashboards and plan to use them with Kibana 6.0,
+you need to import them into 5.6 now by using the Kibana API, or they will no
+longer work in Kibana 6.0.
```
tsg: I don't think this is true. Kibana automatically upgrades any dashboards that work in 5.6, so as long as the user follows the normal upgrade sequence, there shouldn't be any need for extra instructions.

dedemorton:
@tsg So should I close this PR?

monicasarbu:
yes, let's close this one in favor of https://github.com/elastic/beats/pull/5068

PR #5028, opened 2017-08-28
[WIP] Add command `filebeat test multiline`

urso:
Add new command `filebeat test multiline` to filebeat. The multiline configuration can be read from one prospector, or completely set via command line (TODO: allow command line to overwrite setting in config).

The multiline tester only writes the contents read to stdout. The single events are separated by newlines.

Note: I started to add more parameters to the tester and realized the logs prospector already has a quite rich set of input processing settings. Moving the tester code to the prospector types themselves might give us support for more settings, without having to duplicate that much code. I only worry about the command line usage. E.g. `filebeat test prospector <prospector type> ...`. Plus how extensive is a test supposed to be? E.g. the log prospector might check files are available + permissions are correct, or just test the reader pipeline (user has to specify files). This might result in commands like `filebeat test prospector log permissions`, `filebeat test prospector log multiline ...`, ...

### DIFF
--- a/filebeat/cmd/multiline/reader.go
+++ b/filebeat/cmd/multiline/reader.go
@@ -1,0 +1,108 @@
+package multiline
+
+import (
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+
+	"github.com/elastic/beats/filebeat/harvester/encoding"
+	"github.com/elastic/beats/filebeat/harvester/reader"
+)
+
+type concatReader struct {
+	current io.ReadCloser
+	idx     int
+	paths   []string
+}
+
+func createReader(paths []string) (io.ReadCloser, error) {
+	switch len(paths) {
+	case 0:
+		return ioutil.NopCloser(os.Stdin), nil
+	case 1:
+		return os.Open(paths[0])
+	}
+	return newConcatReader(paths)
+}
+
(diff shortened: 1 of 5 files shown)


filebeat/cmd/multiline/reader.go
```diff
@@ -0,0 +1,108 @@
...
+	switch len(paths) {
+	case 0:
+		return ioutil.NopCloser(os.Stdin), nil
+	case 1:
+		return os.Open(paths[0])
+	}
+	return newConcatReader(paths)
+}
+
+func createPipeline(config *prospectorConfig, in io.Reader) (reader.Reader, error) {
```
ruflin: I'm hoping in the future that we can just "reuse" this code from filebeat without having to reimplement that part.

ruflin:
Great idea. Perhaps we can brainstorm a bit more together on how to make this as easy as possible for the user to use.

exekias:
:100: just a comment: in general I would prefer to put commands logic in some other package (not `cmd`), and keep `cmd` as an interface layer, this way we can easily reuse code from other parts. For this case, think of someone that could want to implement a standalone app for this, or a web service, she could easier reuse the code if it's given as a lib, then the command uses it in the same way.

PR #5034, opened 2017-08-28
Fix make index-template command

ruflin:
Implementation of the index template generation changed slightly in a recent PR, but the generation command was not updated accordingly.


ruflin:
This raises the question if we should have part of our tests also these commands to make sure they keep working when we change the code.

PR #5036, opened 2017-08-28
Reduce perfmon memory allocations

### LINKED ISSUE #5035: Metricbeat perfmon allocates more memory than necessary
While doing heap profiling and code review of the Metricbeat windows-perfmon metricset I observed that more memory was being allocated than was necessary to receive the performance counter values from Windows.

In `PdhGetFormattedCounterArray` the code was allocating memory for `bufferSize` x `sizeof(PdhCounterValueItem)`, but it only needed to allocated `bufferSize`. The buffer size is provided by Windows and already accounts for the sizeof PdhCounterValueItem.

For example if `bufferSize=2048` it would end up allocating ~100 kB instead of 2 kB. This would then be cleaned up by GC later.

- Version: 6.0.0-beta1
- Operating System: Windows 2012 R2
- Steps to Reproduce: Use a perfmon query containing a wildcard.

andrewkroh:
In `PdhGetFormattedCounterArray` the code was allocating memory for `bufferSize` x `sizeof(PdhCounterValueItem)`, but it only needed to allocate `bufferSize`. The buffer size is provided by Windows and already accounts for the sizeof PdhCounterValueItem.

Made an improvement to create regexp objects once at startup instead of on every use.

Moved config validation to startup for non-wildcard counter paths. This allows the check to only be performed one time and for any errors to be returned during the config test.

Fixes #5035

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -60,6 +60,7 @@ https://github.com/elastic/beats/compare/v6.0.0-beta1...master[Check the HEAD di
 - Added missing mongodb configuration file to the `modules.d` folder. {pull}4870[4870]
 - Fix wrong MySQL CRUD queries timelion visualization {pull}4857[4857]
 - Add new metrics to CPU metricsset {pull}4969[4969]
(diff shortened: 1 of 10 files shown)


ruflin:
Nice fix. Merged as Windows build was green. Will this need a backport?

PR #5044, opened 2017-08-29
Modify events dropping on Close

urso:
Update the publisher pipeline to only drop events from the
queue on beat.Client close when: guaranteed mode is set and no global
ACK handler is registered.


urso:
It's some very internal setting not really affecting any beats right now.

ruflin:
Merging as Jenkins is green.

PR #5051, opened 2017-08-29
Improvements to beatname.sh

andrewkroh:
Use `exec` with the command to start the beat so that it replaces the shell process leaving only the beat process. This is nicer because you don't have the "beatname.sh" process showing up in `ps` output.

Change the hashbang to `#!/usr/bin/env bash` instead of `#!/bin/bash`. This is generally more portable.

### DIFF
--- a/dev-tools/packer/platforms/centos/beatname.sh.j2
+++ b/dev-tools/packer/platforms/centos/beatname.sh.j2
@@ -1,9 +1,9 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 # Script to run {.beat_name} in foreground with the same path settings that
 # the init script / systemd unit file would do.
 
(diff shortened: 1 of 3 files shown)


andrewkroh:
I did a quick manual test on Centos 7.

dev-tools/packer/platforms/centos/beatname.sh.j2
```diff
@@ -1,9 +1,9 @@
-#!/bin/bash
+#!/usr/bin/env
```
exekias: I think bash is missing here

andrewkroh: Probably should have copied the file instead of typing it twice.

andrewkroh: Fixed.

Issue #4810, opened 2017-08-01
Module initialization errors should stop filebeat/metricbeat when reload is disabled


exekias:
After moving to modules.d layout we have lost a previous behavior. We need stop the beat if reload is false and any of this errors happens:

- [ ] No modules are configured
- [x] initialization fails for any of the modules

PR #5053, opened 2017-08-29
Check modules and prospectors settings when reload is off

exekias:
This change ensures that we check configs when loading them through
`filebeat.config.modules`, `filebeat.config.prospectors` or `metricbeat.config.modules`.

It will error and exit if settings are wrong and **reload is disabled**

Part of #4810


kvch:
Please note that it conflicts with my PR #5026, so one of us will have to rebase their branch after merging. :)

libbeat/cfgfile/reload.go
```diff
@@ -63,13 +66,60 @@ func NewReloader(cfg *common.Config) *Reloader {
...
 	return &Reloader{
 		registry: NewRegistry(),
 		config:   config,
+		path:     path,
 		done:     make(chan struct{}),
 	}
 }
 
+// Check configs are valid (only if reload is disabled)
+func (rl *Reloader) Check(runnerFactory RunnerFactory) error {
```
kvch: It is quite similar to the existing config reload code in `Run` function. I would extract anything that could be extracted into smaller functions. So those could be reused it in both `Check` and `Run`. Thus, code duplication could be eliminated.

kvch:
I have one concerns regarding the code. I have already accepted your PR, so decide if you would address it or not.

libbeat/cfgfile/reload.go
exekias: I'll do some refactoring, thanks!

exekias:
This should be ready for a second look

kvch:
After it turns green, I would merge it.

PR #5055, opened 2017-08-30
Update gotestcover tool path

kaarolch:
If the $GOPATH was set correctly during 
`go get github.com/pierrre/gotestcover` the gotestcover binary will go to ${GOPATH}/bin/gotestcover. So using the `/bin/gotestcover` don't make sens. if the system ${PATH} are correctly set we should use 
`gotestcover`.


ruflin:
This change requires that the `bin` path in the GOPATH was added to `PATH` or similar. If that is not the case it would probably not work. What is the advantage of not using the full path?

kaarolch:
By default when I pull your repo and run the `make` one of the step: 
`go get github.com/elastic/beats/vendor/github.com/pierrre/gotestcover`
download and put the binary to $HOME/go/bin  and running the make failed because you use ${BEAT_GOPATH}/bin/ to force point script to look staff in this folder. If the system or script correctly set the path env we can run the gotestcover as other system binary. If the path was not correctly set the make file also throw error because ${BEAT_GOPATH} was not correctly set and the go get by default download binnary to other repository.

kaarolch:
Ok, After deeper analysis I found that on my system the ${BEAT_GOPATH} was wrongly extracted because I miss some shell specific.

PR #5070, opened 2017-08-31
Fix ceph cluster_status field mapping

### LINKED ISSUE #5069: metricbeat template command line command fails and throws error
- Version: snapshot built from master
- Operating System: macos 10.12.6
- Steps to Reproduce: manually add the template like `./metricbeat setup --template`

The command outputs the following error:

```
Exiting: Error loading Elasticsearch template: could not load template: couldn't load template: couldn't load json. Error: 400 Bad Request: {"error":{"root_cause":[{"type":"mapper_parsing_exception","reason":"No handler for type [bool] declared on field [nearfull]"}],"type":"mapper_parsing_exception","reason":"Failed to parse mapping [doc]: No handler for type [bool] declared on field [nearfull]","caused_by":{"type":"mapper_parsing_exception","reason":"No handler for type [bool] declared on field [nearfull]"}},"status":400}. Response body: {"error":{"root_cause":[{"type":"mapper_parsing_exception","reason":"No handler for type [bool] declared on field [nearfull]"}],"type":"mapper_parsing_exception","reason":"Failed to parse mapping [doc]: No handler for type [bool] declared on field [nearfull]","caused_by":{"type":"mapper_parsing_exception","reason":"No handler for type [bool] declared on field [nearfull]"}},"status":400}
```
It seems this is related to this file: https://github.com/elaron/beats/blob/master/metricbeat/module/ceph/cluster_status/_meta/fields.yml

In line 87 and 91 there is `type: bool` — if you change this to `type: boolean` the command works.

andrewkroh:
The fields.yml had `bool` instead of `boolean` causing a `mapper_parsing_exception` due to `No handler for type [bool]`.

Fixes #5069

### DIFF
--- a/metricbeat/docs/fields.asciidoc
+++ b/metricbeat/docs/fields.asciidoc
@@ -1027,7 +1027,7 @@ Cluster status version
 [float]
 === `ceph.cluster_status.osd.full`
 
-type: bool
+type: boolean
 
 Is osd full
 
@@ -1035,7 +1035,7 @@ Is osd full
 [float]
 === `ceph.cluster_status.osd.nearfull`
 
-type: bool
+type: boolean
 
 Is osd near full
 
--- a/metricbeat/module/ceph/cluster_status/_meta/fields.yml
+++ b/metricbeat/module/ceph/cluster_status/_meta/fields.yml
@@ -84,11 +84,11 @@
       description: >
         Cluster status version
     - name: osd.full
-      type: bool
+      type: boolean
       description: >
         Is osd full
     - name: osd.nearfull
-      type: bool
+      type: boolean
       description: >
         Is osd near full
     - name: osd.num_osds


exekias:
You will need to do a `make update`

exekias:
I'm wondering, don't we test we can actually load the template? :thinking:

andrewkroh:
Updated with the `make update`. I was wondering the same thing about testing templates... @ruflin do you know?

ruflin:
I think we do test the template loading for libbeat and filebeat, but not metricbeat :-( There were some issues in the past including loading dashboards so I think the tests got disabled. I'm thinking of adding these tests with https://github.com/elastic/beats/pull/4834
But if someone get time before ... 

I'm aware we don't run `beats-tester` with elasticsearch. Now that we have a wide variety of setup commands, this could start to make sense to test the setup commands work as expected. @exekias Do we have tests for the setup commands already?

@andrewkroh Does this PR need a backport?

exekias:
We have it, but in libbeat too, wouldn't be difficult to replicate it to metricbeat and others

andrewkroh:
> @andrewkroh Does this PR need a backport?

No, the affected fields only exist in master.

PR #5071, opened 2017-08-31
Optional monitoring registry

urso:
- update pipeline to optionally accept a monitoring registry
  -> if no registry is given, don't collect any metrics
- turn observer type into interface + sub-interfaces for the specific
  pipeline components metrics are collected on
- introduce a nilObserver used if no metrics shall be collected


ruflin:
@roncohen FYI

roncohen:
thanks!

PR #5078, opened 2017-09-01
Add generator tests to global testsuite

ruflin:
The generator tests were not part anymore of the testsuite. That means they were only run on Traivs. This readds the generator tests for Jenkins.


ruflin:
@andrewkroh Can you take a look at this one?

andrewkroh:
I don't believe that this will run on Jenkins either. In Jenkins we don't run the top-level `make testsuite` target. I think to run these tests a separate job needs to be added through the JJB.

ruflin:
Got it. Will check about a PR to JJB. 

Would be nice to have something like this in the future so we would have the stages also in our repo: https://github.com/elastic/apm-agent-python/blob/master/Jenkinsfile

ruflin:
Closing in favor of a JJB PR.

PR #5081, opened 2017-09-01
Support receiving events from audit multicast group

### LINKED ISSUE #4850: Use netlink multicast group to receive audit logs
We can make Auditbeat more user friendly by enabling it to receive audit messages as a multicast subscriber rather than the sole unicast client of the audit framework. This would allow Auditbeat to become a passive listener to audit messages and it could run alongside auditd.

This would be good for people getting started that already have auditd deployed and rules configured. It would give them a quick way to explore their audit data through Auditbeat + ES + Kibana. Later they can migrate their audit rules over to the auditbeat config file and disable auditd (because you probably don't want to waste CPU by running to audit clients).

If multicast is supported by the kernel (3.16+) and no rules are defined then Auditbeat can automatically use multicast. This will make it work without having to consider whether or not auditd is running. I will add a config option to explicitly configure the `connection_type` to use (`unicast` or `multicast`).

audit-multicast-decision-chart

This relates to https://github.com/elastic/go-libaudit/issues/9.

andrewkroh:
Closes #4850 

~This needs to be updated after https://github.com/elastic/go-libaudit/pull/12 is merged and go-libaudit is released.~
It has been updated with the v0.0.6 go-libaudit release.


andrewkroh:
This PR has been updated and is ready to go.

PR #5084, opened 2017-09-01
Keep docker & k8s pod annotations while they are needed

### LINKED ISSUE #4986: Filebeat lose metadata of docker container when container stops
Environment:
filebeat version: 6.0.0-beta1(containerized)
Operating System: CentOS 7.3, Docker-ce 17.06

As it is described at [https://discuss.elastic.co/t/filebeat-miss-metadata-of-docker-container-when-container-stops/98037](url),  filebeat failed to resolve the container's metadata information when the container stops.  Ping @exekias 
Here are the steps to reproduce:
1. Create a swarm service to the swarm:
```bash
docker service create --name smallredspot \ 
  --replicas 5 \
  --network dev \
  --detach=false \
  myregistry.sample.com:5000/smallredspot:latest
```
The smallredspot is a simple java springboot container which will print multiline logs to stdout.
2. Create a filebeat service to collect the logs:
```
docker service create --name filebeat \
  --mode global \
  --mount type=bind,src=/var/run/docker.sock,dst=/var/run/docker.sock \
  --mount type=bind,src=/var/lib/docker/containers/,dst=/var/lib/docker/containers/ \
  --user root \
  --config src=filebeat,target=/usr/share/filebeat/filebeat.yml
docker.elastic.co/beats/filebeat:6.0.0-beta1
```
And the filebeat.yml:
```yml
filebeat.prospectors:
- type: log
  paths:
   - '/var/lib/docker/containers/*/*.log'
  json.message_key: log
  json.keys_under_root: true
  multiline.pattern: '^[[:space:]]+|^Caused by:'
  multiline.negate: false
  multiline.match: after
  tail_files: true
  processors:
  - add_docker_metadata: ~
output.logstash:
  hosts: ["192.168.1.120:5044"]
```
Also logstash.conf
```ruby
input {
  beats {
    port => 5044
    type => "docker_logs"
  }
}
output {
  if [type] == "docker_logs" {
    elasticsearch {
      hosts => "http://192.168.1.120:9200"
      index => "%{[docker][container][labels][com][docker][swarm][service][name]}-%{+YYYY.MM}"
    }
}
```
Label `com.docker.swarm.service.name`  was generated automatically by swarm when Creating the service. In this case, it will be the name of the service: smallredspot. So all the 5 containers' logs will goes into the `smallredspot-*` index following the logic. It works as expected till now.
3. Update the smallredspot service
```
docker service update --force smallredspot
```
This instruction will first stop and kill the container, then remove it and run a new one with new metadata but the service name will not change. Problem comes : a strange index is created on elasticsearch:

And the message is 
```json
{
    "_index": "%{[docker][container][labels][com][docker][swarm][service][name]}-2017.08",
    "_type": "docker_logs",
    "_id": "AV4N5o1hw7w152Pm1x8z",
    "_version": 1,
    "_score": 1,
    "_source": {
        "@timestamp": "2017-08-23T07:03:56.459Z",
        "docker": {
            "container": {
                "id": "43c3759ce800b61f5e1b1b7ff3b4431e7dc1d2cd3c70ff62a5b1e94c07420629"
            }
        },
        "source": "/var/lib/docker/containers/43c3759ce800b61f5e1b1b7ff3b4431e7dc1d2cd3c70ff62a5b1e94c07420629/43c3759ce800b61f5e1b1b7ff3b4431e7dc1d2cd3c70ff62a5b1e94c07420629-json.log",
        "offset": 8570,
        "stream": "stdout",
        "time": "2017-08-23T07:03:49.054752452Z",
        "log": "2017-08-23 15:03:49.054 INFO 1 --- [ Thread-2] o.s.j.e.a.AnnotationMBeanExporter : Unregistering JMX-exposed beans on shutdown",
        "beat": {
            "name": "60e1969b6f44",
            "hostname": "60e1969b6f44",
            "version": "6.0.0-beta1"
        },
        "prospector": {
            "type": "log"
        },
        "@version": "1",
        "host": "60e1969b6f44",
        "type": "docker_logs",
        "tags": [
            "beats_input_raw_event"
        ]
    }
}
```
It seems the last log line while the container is shutting down. But the container's metadata is missing. Due to this the log goes into wrong place and will not be shown in smallredspot-* index.

exekias:
In some cases pod annotations are neede after the container/pod is deleted, for
instance when filebeat is reading the log behind the container.

This change makes sure we keep metadata after a pod is gone. By storing
access times we ensure that it's available as long as it's being used

fixes #4986

### DIFF
--- a/libbeat/processors/add_docker_metadata/watcher.go
+++ b/libbeat/processors/add_docker_metadata/watcher.go
@@ -1,15 +1,17 @@
 package add_docker_metadata
 
 import (
-	"context"
 	"fmt"
 	"net/http"
+	"sync"
 	"time"
 
 	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/events"
 	"github.com/docker/docker/api/types/filters"
 	"github.com/docker/docker/client"
 	"github.com/docker/go-connections/tlsconfig"
+	"golang.org/x/net/context"
 
 	"github.com/elastic/beats/libbeat/logp"
 )
@@ -22,6 +24,9 @@ type Watcher interface {
 	// Start watching docker API for new containers
 	Start() error
 
+	// Stop watching docker API for new containers
+	Stop()
+
 	// Container returns the running container with the given ID or nil if unknown
 	Container(ID string) *Container
 
@@ -30,11 +35,15 @@ type Watcher interface {
 }
 
 type watcher struct {
-	client             *client.Client
+	sync.RWMutex
+	client             Client
 	ctx                context.Context
 	stop               context.CancelFunc
 	containers         map[string]*Container
+	deleted            map[string]time.Time // deleted annotations key -> last access time
+	cleanupTimeout     time.Duration
 	lastValidTimestamp int64
+	stopped            sync.WaitGroup
 }
 
 // Container info retrieved by the watcher
@@ -43,6 +52,12 @@ type Container struct {
 	Name   string
 	Image  string
 	Labels map[string]string
+}
+
+// Client for docker interface
+type Client interface {
+	ContainerList(ctx context.Context, options types.ContainerListOptions) ([]types.Container, error)
+	Events(ctx context.Context, options types.EventsOptions) (<-chan events.Message, <-chan error)
 }
 
 type WatcherConstructor func(host string, tls *TLSConfig) (Watcher, error)
@@ -69,28 +84,51 @@ func NewWatcher(host string, tls *TLSConfig) (Watcher, error) {
 		}
 	}
 
-	cli, err := client.NewClient(host, dockerAPIVersion, httpClient, nil)
+	client, err := client.NewClient(host, dockerAPIVersion, httpClient, nil)
 	if err != nil {
 		return nil, err
 	}
 
+	return NewWatcherWithClient(client, 60*time.Second)
+}
+
+func NewWatcherWithClient(client Client, cleanupTimeout time.Duration) (*watcher, error) {
 	ctx, cancel := context.WithCancel(context.Background())
 	return &watcher{
-		client:     cli,
-		ctx:        ctx,
-		stop:       cancel,
-		containers: make(map[string]*Container),
+		client:         client,
+		ctx:            ctx,
+		stop:           cancel,
+		containers:     make(map[string]*Container),
+		deleted:        make(map[string]time.Time),
+		cleanupTimeout: cleanupTimeout,
 	}, nil
 }
 
 // Container returns the running container with the given ID or nil if unknown
 func (w *watcher) Container(ID string) *Container {
-	return w.containers[ID]
+	w.RLock()
+	container := w.containers[ID]
+	w.RUnlock()
+
+	// Update last access time if it's deleted
+	if _, ok := w.deleted[ID]; ok {
+		w.Lock()
+		w.deleted[ID] = time.Now()
+		w.Unlock()
+	}
+
+	return container
 }
 
 // Containers returns the list of known containers
 func (w *watcher) Containers() map[string]*Container {
-	return w.containers
+	w.RLock()
+	defer w.RUnlock()
+	res := make(map[string]*Container)
+	for k, v := range w.containers {
+		res[k] = v
+	}
+	return res
 }
 
 // Start watching docker API for new containers
@@ -99,6 +137,8 @@ func (w *watcher) Start() error {
 	logp.Debug("docker", "Start docker containers scanner")
 	w.lastValidTimestamp = time.Now().Unix()
 
+	w.Lock()
+	defer w.Unlock()
 	containers, err := w.client.ContainerList(w.ctx, types.ContainerListOptions{})
 	if err != nil {
 		return err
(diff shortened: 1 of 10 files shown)


exekias:
@dengqingpei1990 this PR should fix #4986, just in case you want to give this a try before we merge it. From my local tests, it seems to work nice

libbeat/processors/add_kubernetes_metadata/podwatcher.go
```diff
@@ -189,18 +194,65 @@ func (p *PodWatcher) worker() {
...
+}
+
 func (p *PodWatcher) GetMetaData(arg string) common.MapStr {
-	p.annotationCache.Lock()
-	defer p.annotationCache.Unlock()
-	if meta, ok := p.annotationCache.annotations[arg]; ok {
+	p.annotationCache.RLock()
+	meta, ok := p.annotationCache.annotations[arg]
+	var deleted bool
+	if ok {
```
vjsamuel: For logs, we can update the pod labels using `kubectl label po`. The container name remains the same and the update will call `onPodDelete && onPodAdd`. So now the living container's metadata is in the deleted map. This portion of code will end up returning old metadata.

vjsamuel: Ideally we should first check metadata from live pods and return data for that.

exekias: I've added a commit addressing this, basically, if an add method is called it will remove any deleted flag for all its indexes

exekias:
good point, I've addressed the issue :+1:

libbeat/processors/add_docker_metadata/watcher.go
```diff
@@ -143,17 +163,22 @@ func (w *watcher) watch() {
...
 						ID:     event.Actor.ID,
 						Name:   name,
 						Image:  image,
 						Labels: event.Actor.Attributes,
 					}
+					w.Unlock()
 				}
 
 				// Delete
 				if event.Action == "die" || event.Action == "kill" {
```
hofler: If a container dies and gets started again it will still time out eventually and lose the metadata.

exekias: I think that case is treated, as `onPodAdd` (also called in case of update) cleans any deleted flag: https://github.com/elastic/beats/pull/5084/files#diff-c211c9481fb346b82d8713e779e0be9fR141

exekias: Sorry you are right, I was talking about the Kubernetes watcher, not Docker, will apply the same fix, thanks!

I'm also working on tests for both watchers

exekias: pushed https://github.com/elastic/beats/pull/5084/commits/b54187e6d907895f1bff3678338bb1e91fc5945a

hofler: When I restart a container, the events are: kill -> die -> stop -> start -> restart. So the container will not be un-deleted in this case.

Looking at it, adding the metadata on "created" and removing on "die" does seem inconsistent to me.

exekias: Let me do some tests, looks like it would be better to add it on start, isn't it?

hofler: I think adding on start is the better approach.

exekias:
jenkins retest this

libbeat/processors/add_docker_metadata/watcher.go
```diff
@@ -69,28 +84,51 @@ func NewWatcher(host string, tls *TLSConfig) (Watcher, error) {
...
+		deleted:        make(map[string]time.Time),
+		cleanupTimeout: cleanupTimeout,
 	}, nil
 }
 
 // Container returns the running container with the given ID or nil if unknown
 func (w *watcher) Container(ID string) *Container {
-	return w.containers[ID]
+	w.RLock()
+	container := w.containers[ID]
```
ruflin: I'm aware the code was here before, but should we check here if `ok` if the `ID` actually exists?

libbeat/processors/add_docker_metadata/watcher.go
```diff
@@ -164,8 +218,42 @@ func (w *watcher) watch() {
...
+		select {
+		case <-w.ctx.Done():
+			w.stopped.Done()
+			return
+		default:
+			// Check entries for timeout
+			var toDelete []string
+			timeout := time.Now().Add(-w.cleanupTimeout)
+			w.RLock()
+			for key, lastSeen := range w.deleted {
```
ruflin: should we add some debug logging to this? Otherwise very hard to figure out what is still in memory and what was removed.

libbeat/processors/add_kubernetes_metadata/podwatcher.go
```diff
@@ -189,18 +197,65 @@ func (p *PodWatcher) worker() {
...
+		time.Sleep(p.cleanupTimeout)
+
+		select {
+		case <-p.ctx.Done():
+			return
+		default:
+			// Check entries for timeout
+			var toDelete []string
+			timeout := time.Now().Add(-p.cleanupTimeout)
+			p.annotationCache.RLock()
```
ruflin: Not for this PR, but the code looks so similar to the other cleanupWorker, I'm thinking if parts could be shared.

libbeat/processors/add_docker_metadata/watcher.go
exekias: We get `nil` if the container is not there, and we treat that case

exekias:
done, thanks for the review everyone!

PR #5085, opened 2017-09-03
Remove error log from runnerfactory as error is returned by API

vjsamuel:
The same error that is being logged is also being returned by the API. it should be the caller's choice to decide if its a debug/info/error level log.

### DIFF
--- a/filebeat/prospector/runnerfactory.go
+++ b/filebeat/prospector/runnerfactory.go
@@ -5,7 +5,6 @@ import (
 	"github.com/elastic/beats/filebeat/registrar"
 	"github.com/elastic/beats/libbeat/cfgfile"
 	"github.com/elastic/beats/libbeat/common"
-	"github.com/elastic/beats/libbeat/logp"
 )
 
 // RunnerFactory is a factory for registrars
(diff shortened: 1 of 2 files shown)


exekias:
ok to test

filebeat/prospector/runnerfactory.go
```diff
@@ -28,7 +27,6 @@ func NewRunnerFactory(outlet channel.Factory, registrar *registrar.Registrar, be
 func (r *RunnerFactory) Create(c *common.Config) (cfgfile.Runner, error) {
 	p, err := New(c, r.outlet, r.beatDone, r.registrar.GetStates())
 	if err != nil {
-		logp.Err("Error creating prospector: %s", err)
```
ruflin: LGTM. The only problem with the error logged by the runner factory is probably that it's more generic and does not talk about the prospector itself?

exekias: One option would be to use `errors.Wrap` in `prospector.New`https://github.com/vjsamuel/beats/blob/e3e3770af3dd7d3f4563955274aff976f415d9cd/filebeat/prospector/prospector.go#L34

exekias:
jenkins retest this please

filebeat/prospector/runnerfactory.go
ruflin: Merged it as is as I kind of like producing errors in as few places and just pass them on. We can still improve it in case it becomes an issue.

PR #5086, opened 2017-09-03
Use fpm --deb-systemd option for systemd unit file

praseodym:
This allows fpm to do the heavy lifting regarding systemd. Unfortunately [this option does not yet exist for rpm](https://github.com/jordansissel/fpm/issues/1163).

### DIFF
--- a/dev-tools/packer/platforms/debian/run.sh.j2
+++ b/dev-tools/packer/platforms/debian/run.sh.j2
@@ -7,10 +7,6 @@ cd /build
 # the init scripts needs to have the right name
 cp ${RUNID}.init /tmp/{{.beat_pkg_name}}.init
 
-# create script to reload systemd config
-echo "#!/bin/bash" > /tmp/systemd-daemon-reload.sh
-echo "systemctl daemon-reload 2> /dev/null || true" >> /tmp/systemd-daemon-reload.sh
-
 # add SNAPSHOT if it was requested
 VERSION="{{.version}}"
 if [ "$SNAPSHOT" = "yes" ]; then
@@ -27,7 +23,7 @@ FPM_ARGS=(
         --description "{{.beat_description}}"
         --url {{.beat_url}}
         --deb-init /tmp/{{.beat_pkg_name}}.init
-        --after-install /tmp/systemd-daemon-reload.sh
+        --deb-systemd ${RUNID}.service
         --config-files /etc/{{.beat_name}}/{{.beat_name}}.yml
         homedir/=/usr/share/{{.beat_name}}
         beatname-${RUNID}.sh=/usr/bin/{{.beat_name}}
@@ -35,7 +31,6 @@ FPM_ARGS=(
         {{.beat_name}}-linux.yml=/etc/{{.beat_name}}/{{.beat_name}}.yml
         {{.beat_name}}-linux.reference.yml=/etc/{{.beat_name}}/{{.beat_name}}.reference.yml
         fields.yml=/etc/{{.beat_name}}/fields.yml
-        ${RUNID}.service=/lib/systemd/system/{{.beat_pkg_name}}.service
         god-linux-{{.arch}}=/usr/share/{{.beat_name}}/bin/{{.beat_name}}-god
   )
 


tsg:
jenkins, package it

tsg:
Nice, thanks @praseodym.

tsg:
Merged in master to have the packaging test run. I'll test the snapshots before backporting this to 6.0.

tsg:
beats-package is OK with this change: https://beats-ci.elastic.co/job/elastic+beats+master+package/98/ beats-tester currently can't run because of another error.

tsg:
@praseodym I tested this manually, but there seem to be an issue. When starting the daemon on Ubuntu 16.04, it appears to have started twice:

```
$ ps -ef | grep metricbeat
root      2738     1  0 11:19 ?        00:00:00 /usr/share/metricbeat/bin/metricbeat -c /etc/metricbeat/metricbeat.yml -path.home /usr/share/metricbeat -path.config /etc/metricbeat -path.data /var/lib/metricbeat -path.logs /var/log/metricbeat
root      2795     1  0 11:19 ?        00:00:00 /usr/share/metricbeat/bin/metricbeat-god -r / -n -p /var/run/metricbeat.pid -- /usr/share/metricbeat/bin/metricbeat -c /etc/metricbeat/metricbeat.yml -path.home /usr/share/metricbeat -path.config /etc/metricbeat -path.data /var/lib/metricbeat -path.logs /var/log/metricbeat
root      2796  2795  0 11:19 ?        00:00:00 /usr/share/metricbeat/bin/metricbeat -c /etc/metricbeat/metricbeat.yml -path.home /usr/share/metricbeat -path.config /etc/metricbeat -path.data /var/lib/metricbeat -path.logs /var/log/metricbeat
```

From the looks of it, I think Metricbeat is started once via the unit file and once via the init script. Interesting, though, before this patch the unit file was automatically used *instead* of the init script. 

@jordansissel Perhaps you have seen this before when using the `--deb-systemd` flag in FPM?

jordansissel:
@tsg I haven't much experience with the --deb-systemd flag. I'll spend some time this week fiddling with it.

(Tangent: I don't use this flag because I use `pleaserun` to do the heavy lifting for service management, and use fpm+pleaserun integeration to build a package to install the service)

praseodym:
@tsg Are you seeing this on a system that had a previous installation of Metricbeat? In that case there might be some stray files somewhere.

Also, is there a way to download 'official' snapshot packages from https://beats-ci.elastic.co? The builds jobs don't seem to have any artifacts attached to them, but maybe I'm looking in the wrong place.

tsg:
Thanks @jordansissel 

@praseodym I did made sure try a couple of times and I purged metricbeat between runs. 

The snapshots are created by another internal CI. Here is a direct link to the metricbeat [deb file](https://snapshots.elastic.co/downloads/beats/metricbeat/metricbeat-7.0.0-alpha1-SNAPSHOT-amd64.deb).

praseodym:
The problem is that the init.d script gets installed as `/etc/init.d/metricbeat` whereas the systemd unit file gets installed as `/lib/systemd/system/debian-metricbeat-amd64.service`. systemd will not load init.d scripts when the name matches a service name, which is not the case here.

I've created #5174 to address this. Sorry for the oversight!

tsg:
@praseodym Fantastic, thanks a lot!

jordansissel:
@tsg following up, I checked the Metricbeat 5.6.1 deb and I noticed something possibly interesting (and possibly a bug in fpm):

* /etc/init.d/metricbeat is marked as a config file
* /lib/systemd/system/metricbeat.service is not.

The impact is that if someone modifies `/etc/init.d/metricbeat` then it will remain untouched (keeping the operator's modified version) on a package upgrade.

Is this desired or is it a bug? I can open a separate issue to discuss if you like.

tsg:
> Is this desired or is it a bug?

Hmm, I could argue both ways :) Are all files under `/etc/` marked as config automatically? I'm not sure what the Debian policies recommend here.

jordansissel:
@tsg Debian policies and user expectations are going to be different. It's all a tricky thing, unfortunately.

PR #5099, opened 2017-09-05
Add script to test against apm-server

ruflin:
apm-server depends on libbeat. To make sure we are aware when changes in libbeat break apm-server, this script allows to run the current code base against a specific branch / commit of apm-server. The idea is to add this also to our tests in a later PR but allow it to fail. Like this we always know when we break the code from apm-server.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -140,3 +140,8 @@ notice: python-env
 python-env:
 	@test -d $(PYTHON_ENV) || virtualenv $(VIRTUALENV_PARAMS) $(PYTHON_ENV)
 	@$(PYTHON_ENV)/bin/pip install -q --upgrade pip autopep8 six
+
+# Tests if apm works with the current code
+.PHONY: python-env
+test-apm:
+	sh ./script/test_apm.sh
--- a/script/test_apm.sh
+++ b/script/test_apm.sh
@@ -1,0 +1,46 @@
+#!/usr/bin/env bash
+
+# This script allows to test against the code from apm-server
+# By default it is checked against master. The env variable BRANCH
+# can be set to any value accepted by git checkout.
+#
+# go build is executed on the apm-server code to verify that it can
+# still be built and runs the go tests with make unit.
+
+set -e
+
+# Check if a special branch env variable is set
+if [ -z "${BRANCH}" ]; then
+    BRANCH=master
+fi
+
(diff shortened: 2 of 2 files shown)


kvch:
What if someone breaks the build of APM? It it our responsibility to stay compatible or do they follow up our changes in their code?

ruflin:
It's not the beats team responsibility to fix it but the team should be pinged about it, best directly in the PR.

tsg:
@ruflin How should we present the result of these tests in the Github or CI UI? For travis we could have it under "allowed failures", but that means we might miss it if we don't look into the details.

For Jenkins, ideal would be to have another separate entry from `beats-ci`, so it shows up top level in the UI.

ruflin:
@tsg Agree on both points. I need to check if I can trigger on travis some emails for allowed failures to still be notified about it.

simitt:
@ruflin I appreciate the attempt for detecting changes in beats that'd break the APM Server a lot.

Putting apm specific tests into the beats framework does feel a bit like creating dependencies on the wrong side though. 

What do you think about creating a webhook for every beats PR that triggers a Jenkins build for the apm server instead? Given the assumption that we manage to make github webhooks properly work with Jenkinsfiles/pipelines, we could create a `DependenciesJenkinsFile` in the apm server that takes care of running the `make update-beats` cmd, considering the branch that triggers the build, and running the apm server tests afterwards.

I am aware of this suggestion involving more work and some assumption. However, it would avoid having tests run from something built on top of beats within beats, and might be worth a discussion.

ruflin:
I wanted to have it for now in the beats to raise awareness in the beats team that there is now an other project directly depending on beats and not adding the overhead to the apm-server team to understand every change we do in detail. Over time I fully agree that this should in the apm-server team itself and also part of the apm-server code base. 

Having what you suggest above would be pretty nice especially as it would work with the different branches. +1 if we can do that "pretty easy" :-)

simitt:
Looks like we are heading into the same direction. As mentioned before I highly appreciate the effort and the attempt to put more awareness for the tight connection between apm-server and beats. 

What do you think about going forward with your PR for now and I'm going to put this on my TODO list once we figured out the current Jenkins Webhook trigger issues?

ruflin:
Sounds like a good plan to me :-)

Issue #5093, opened 2017-09-05
Metricbeats Kubernetes event module unable to parse empty date field 


mikelorant:
*Issue*
Getting the following error when trying to parse an event.
```
2017-09-04T02:11:28Z WARN Can not index event (status=400): {"type":"mapper_parsing_exception","reason":"failed to parse [kubernetes.event.metadata.timestamp.deleted]","caused_by":{"type":"illegal_argument_exception","reason":"Invalid format: \"\""}}
```

Enabling debug mode showed the following event message.
```
2017-09-04T02:27:52Z DBG  Publish event: {
  "@timestamp": "2017-09-04T02:27:52.271Z",
  "@metadata": {
    "beat": "\u003cnot set\u003e",
    "type": "doc"
  },
  "kubernetes": {
    "event": {
      "message": "Created pod: ywqvd-kube-state-metrics-483864468-x39ft",
      "reason": "SuccessfulCreate",
      "type": "Normal",
      "count": 1,
      "involved_object": {
        "kind": "ReplicaSet",
        "uid": "86ada97c-9107-11e7-81a9-024fe15315e7",
        "api_version": "extensions",
        "resource_version": "19155721",
        "name": "ywqvd-kube-state-metrics-483864468"
      },
      "metadata": {
        "generate_name": "",
        "uid": "a6da2702-9118-11e7-81a9-024fe15315e7",
        "resource_version": "367834",
        "timestamp": {
          "created": "2017-09-04T02:27:52Z",
          "deleted": ""
        },
        "name": "ywqvd-kube-state-metrics-483864468.14e107918ea3c588",
        "namespace": "infrastructure-metrics-shared-v1",
        "self_link": "/api/v1/namespaces/infrastructure-metrics-shared-v1/events/ywqvd-kube-state-metrics-483864468.14e107918ea3c588"
      },
      "timestamp": {
        "first_occurrence": "2017-09-04T02:27:52.000Z",
        "last_occurrence": "2017-09-04T02:27:52.000Z"
      }
    }
  },
  "metricset": {
    "module": "kubernetes",
    "name": "event"
  },
  "beat": {
    "name": "ywqvd-metricbeat-4231345040-qks11",
    "hostname": "ywqvd-metricbeat-4231345040-qks11",
    "version": "6.0.0-beta2"
  },
  "meta": {
    "cloud": {
      "instance_id": "i-0f919437f290ba006",
      "machine_type": "r4.2xlarge",
      "region": "ap-southeast-2",
[...]

PR #5105, opened 2017-09-06
Fix kubernetes events module to be able to index time fields properly

vjsamuel:
Fixes https://github.com/elastic/beats/issues/5093

### DIFF
--- a/metricbeat/module/kubernetes/event/event.go
+++ b/metricbeat/module/kubernetes/event/event.go
@@ -84,10 +84,8 @@ func (m *MetricSet) Run(reporter mb.PushReporter) {
 			return
 		case msg := <-m.watcher.eventQueue:
 			// Ignore events that are deleted
-			if msg.Metadata.DeletionTimestamp == "" {
-				if msg.Metadata.DeletionTimestamp == "" {
-					reporter.Event(generateMapStrFromEvent(msg))
-				}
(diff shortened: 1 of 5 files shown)


metricbeat/module/kubernetes/event/event.go
```diff
@@ -97,7 +95,6 @@ func generateMapStrFromEvent(eve *Event) common.MapStr {
 	eventMeta := common.MapStr{
 		"timestamp": common.MapStr{
 			"created": eve.Metadata.CreationTimestamp,
-			"deleted": eve.Metadata.DeletionTimestamp,
```
exekias: Does this mean we no longer store `deleted` timestamp? it should be removed from fields.yml if that's the case

exekias:
Thank you @vjsamuel! :tada:, I left a comment

metricbeat/module/kubernetes/event/event.go
vjsamuel: based on https://github.com/vjsamuel/beats/blob/a22d999d3f076d7e9fda6cf93e7a883acbf02913/metricbeat/module/kubernetes/event/event.go#L87 we never write into the destination backend events that are deleted as it would skew event counts. hence it doesn't make sense to even have a deleted timestamp field in both the event and the fields.yml. So i have gone ahead and removed it.

exekias:
ok to test

PR #5106, opened 2017-09-06
Introduce BEAT_BINARY_NAME

roncohen:
This introduces BEAT_BINARY_NAME as a way to differentiate between what the actual binary is called and what we'd like to call the application.

For example:
```
BEAT_NAME=APM Server
BEAT_BINARY_NAME=apm-server
```
BEAT_BINARY_NAME defaults to BEAT_NAME so no change is needed in the existing beats.


libbeat/scripts/generate_fields_docs.py
```diff
@@ -133,6 +133,6 @@ def fields_to_asciidoc(input, output, beat):
     output = open(beat_path + "/docs/fields.asciidoc", 'w')
 
     try:
-        fields_to_asciidoc(fields, output, beat_name.title())
+        fields_to_asciidoc(fields, output, beat_name)
```
roncohen: Subtle change here: The name of the beat will no longer be the auto-capitalized binary name in autogenerated fields doc. Instead, beats should set BEAT_NAME to whatever they want to show up in the autogenerated document.

In other words, without changes to the beats, this PR will change the "Metricbeat" to "metricbeat" in the first paragraph here: https://www.elastic.co/guide/en/beats/metricbeat/master/exported-fields.html

I didn't find a clean way to implement this in a completely transparent way for the beats, but suggestions are welcome.

ruflin:
jenkins, package it

PR #5110, opened 2017-09-06
Introduce BEAT_INDEX_PREFIX

roncohen:
This allows applications relying on libbeat to override the index pattern generated.

BEAT_INDEX_PREFIX defaults to BEAT_NAME.

### DIFF
--- a/libbeat/_meta/config.reference.yml
+++ b/libbeat/_meta/config.reference.yml
@@ -155,10 +155,10 @@ output.elasticsearch:
   # Number of workers per Elasticsearch host.
   #worker: 1
 
-  # Optional index name. The default is "beatname" plus date
-  # and generates [beatname-]YYYY.MM.DD keys.
+  # Optional index name. The default is "index-prefix" plus date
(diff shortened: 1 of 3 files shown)


libbeat/_meta/config.reference.yml
```diff
@@ -155,10 +155,10 @@ output.elasticsearch:
   # Number of workers per Elasticsearch host.
   #worker: 1
 
-  # Optional index name. The default is "beatname" plus date
-  # and generates [beatname-]YYYY.MM.DD keys.
+  # Optional index name. The default is "index-prefix" plus date
```
ruflin: should we call it `beat-index-prefix` to make sure `index-prefix` is not used anywhere else and be in sync with the variable name in the Makefile?

simitt:
I work on a follow up PR that includes also change the default inside beats, see https://github.com/elastic/beats/pull/5161

ruflin:
Closing as https://github.com/elastic/beats/pull/5161 covered this work I think.

PR #5128, opened 2017-09-07
Cherry-pick #5120 to 6.0: MB mongodb module: connect on fetch, not on init

### LINKED ISSUE #5090: Beats Won't Start if Modules Can't Connect
If I use a filebeats module, like 'system' or 'nginx', filebeats will fail to start if elasticsearch isn't up and running. There's no retry logic in the modules, like there is when you run beats without them. Same is true with metricbeats and the mongodb module. If mongod isn't up, then the metric module won't even start.

This effectively means that modules cannot be reliably used in production, despite the great benefit they provide on perfect, problem free systems.

tsg:
Cherry-pick of PR #5120 to 6.0 branch. Original message: 

Moving the Dial call to the Fetch, so that in case Mongodb is not (yet)
available, Metricbeat doesn't exit with an error, but just reports the
service being down. This is consistent with the way most of the other
modules are working.

Fixes #5090


exekias:
you will need to rebase

Issue #5139, opened 2017-09-08
Memory leak seen when enabling reload on filebeat.config.prospectors


vjsamuel:
Version: HEAD on master
OS: golang:1.8 docker container

On using the following config:

```
filebeat.config.prospectors:
  enabled: true
  path: prospectors.d/*.yml
  reload: 
    enabled: true
    period: 5s
```
It is seen that filebeat memory usage on docker keeps growing until a point that it is OOM killed.

On doing more analysis with pprof, it is seen that there are many unused prospector objects that are still lingering around.

Short term work around can be to just create runner objects only when a new cfg file comes in or a config is removed. This avoids creating Runners which causes the leak in filebeat.

tsg:
Closing this one as the short term solution was implemented.

PR #5141, opened 2017-09-09
Remove runner creation from every reload check

vjsamuel:
This PR will partly mitigate https://github.com/elastic/beats/issues/5139 as it removes runner creation during every reload interval.

### DIFF
--- a/libbeat/cfgfile/reload.go
+++ b/libbeat/cfgfile/reload.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/joeshaw/multierror"
+	"github.com/mitchellh/hashstructure"
 	"github.com/pkg/errors"
 
 	"github.com/elastic/beats/libbeat/common"
@@ -134,7 +135,7 @@ func (rl *Reloader) Run(runnerFactory RunnerFactory) {
 		rl.config.Reload.Period = 0
 	}
 
-	overwriteUpate := true
+	overwriteUpdate := true
 
 	for {
 		select {
@@ -154,8 +155,8 @@ func (rl *Reloader) Run(runnerFactory RunnerFactory) {
 			}
 
 			// no file changes
-			if !updated && !overwriteUpate {
-				overwriteUpate = false
+			if !updated && !overwriteUpdate {
+				overwriteUpdate = false
 				continue
 			}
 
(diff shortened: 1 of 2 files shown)


ruflin:
Travis build failure should be solved by: https://github.com/elastic/beats/pull/5142

libbeat/cfgfile/reload.go
```diff
@@ -174,29 +175,34 @@ func (rl *Reloader) Run(runnerFactory RunnerFactory) {
 					continue
 				}
 
-				runner, err := runnerFactory.Create(c)
+				rawCfg := map[string]interface{}{}
```
ruflin: Not 100% sure anymore, but I think there is a difference between this and `ID()`. In the case of `ID()` the following to configs are the same:

```
a: b
c: d
```

and

```
c: d
a: b
```

That means the order does not matter, but with the above it does. But not 100% sure anymore if that was the reason I ended up with `ID()` but there was an issue.

ruflin:
BTW: If we can remove the `ID` method 👍  from my side. I think also @urso did not like it when I introduced it but we didn't have a better way. That is why I worry the above breaks things.

libbeat/cfgfile/reload.go
vjsamuel:
```
func main () {
	a := map[string]interface{}{
		"a": "b",
		"c": "d",
	}


	b := map[string]interface{}{
		"c": "d",
		"a": "b",
	}

	fmt.Println(hashstructure.Hash(a, nil))
	fmt.Println(hashstructure.Hash(b, nil))
}

```

output:
```
8710980024335696864 <nil>
8710980024335696864 <nil>

```

vjsamuel:
@ruflin should `ID()` be removed in the same PR or a subsequent one?

ruflin:
In case `ID()` is not used anywhere else it would be nice to directly remove it here to make sure we don't forget it. Could you rebase your PR on master because there was a build issue earlier today which is fixed now.

vjsamuel:
@ruflin i will file a subsequent PR immediately later in the day to remove ID. i have to fix collectbeat as well. if you dont mind, i will create an issue and track it against me so that it is definitely taken care.

exekias:
ok to test

andrewkroh:
@ruflin @exekias Do we need to backport this to 5.6? Asking b/c of https://discuss.elastic.co/t/metricbeat-mongodb-leaks-connections-when-metricbeat-config-modules-reload-enabled-set-to-true/108723?u=andrewkroh

ruflin:
@andrewkroh I think we could backport it, but as reloading was beta in 5.6 and is that 6.0 is out with the fix, I would rather encourage people to upgrade.

PR #5143, opened 2017-09-11
Fix the bullets from the redis docs in Filebeat

monicasarbu:


ruflin:
I think this change needs to happen directly in the module and then `make update` needs to be run.

monicasarbu:
Sorry, I have updated now the docs in the redis module.

PR #5146, opened 2017-09-11
Add flush timeout setting to filebeat registrar

urso:


filebeat/config/config.go
```diff
@@ -21,6 +21,7 @@ const (
 type Config struct {
 	Prospectors      []*common.Config `config:"prospectors"`
 	RegistryFile     string           `config:"registry_file"`
+	RegistryFlush    time.Duration    `config:"registry_flush"`
```
ruflin: Would prefer `registry.flush` and `registry.file` but too late to change I think.

filebeat/registrar/registrar.go
```diff
@@ -34,12 +38,13 @@ var (
 	registryWrites = monitoring.NewInt(nil, "registrar.writes")
 )
 
-func New(registryFile string, out successLogger) (*Registrar, error) {
+func New(registryFile string, flushTimeout time.Duration, out successLogger) (*Registrar, error) {
```
ruflin: Should we call the config also flush_timeout? Would make it more explicit.

filebeat/registrar/registrar.go
```diff
@@ -198,6 +212,17 @@ func (r *Registrar) Stop() {
 	r.wg.Wait()
 }
 
+func (r *Registrar) flushRegistry() {
+	if err := r.writeRegistry(); err != nil {
```
ruflin: Should we add a check here that `if bufferedStateUpdates==0` it just returns? Like this flushTimeout could be set very low and still not have frequent writes in case there are no updates.

urso: The timer is only started after an event has been added.

filebeat/registrar/registrar.go
urso: you mean `filebeat.registry_flush_timeout`?

PR #5148, opened 2017-09-11
Add specialized buffers to memqueue

urso:
Use specialized buffer types and eventloops depending on the memqueue
configuration.

If flushing is disabled, a region based ring buffer is used, as loads of
small batches can be generated and we want to minimize additional
allocations of small objects.

If flushing is enabled a list of active and flushed buffers is managed
by the queue. If a buffer is flushed by timeout, but not yet processed
(and not full), additional events being published are added to the
already flushed buffer.

This PR fixes a bug in beta2 regarding the flush timeout not working properly. Having distinct state machines and buffer types for flush based and non-flush based queue configurations helps solving the issue, due to having more simple use-case specific state-machines.

- buffer handling is moved from 'broker' to actual eventloop implementation
- 'broker' only provides common types/channels/interfaces
- renamed brokerBuffer to ringBuffer
- introduce batchBuffer used to hold flushable batches of events
- introduce separate event loop types


libbeat/publisher/queue/memqueue/broker.go
```diff
@@ -16,9 +13,10 @@ type Broker struct {
 
 	logger logger
 
-	buf         brokerBuffer
-	minEvents   int
-	idleTimeout time.Duration
+	bufSize int
+	// buf         brokerBuffer
```
ruflin: Leftover?

libbeat/publisher/queue/memqueue/eventloop.go
```diff
@@ -112,61 +157,309 @@ func (l *eventLoop) handleCancel(req *producerCancelRequest) {
...
-	start, buf := l.broker.get(req.sz)
+	start, buf := l.buf.reserve(req.sz)
 	count := len(buf)
 	if count == 0 {
 		panic("empty batch returned")
 	}
 
 	// log.Debug("newACKChan: ", b.ackSeq, count)
-	ackCH := newACKChan(l.ackSeq, start, count)
+	ackCH := newACKChan(l.ackSeq, start, count, l.buf.buf.clients)
```
ruflin: Naming smells a bit here ;-)

libbeat/publisher/queue/memqueue/ringbuf.go
```diff
@@ -0,0 +1,293 @@
...
+
+import (
+	"fmt"
+
+	"github.com/elastic/beats/libbeat/publisher"
+)
+
+// Internal event ring buffer.
+// The ring is split into 2 regions.
+// Region A contains active events to be send to consumers, while region B can
```
ruflin: s/send/sent/

libbeat/publisher/queue/memqueue/ringbuf.go
```diff
@@ -0,0 +1,293 @@
...
+}
+
+func (b *ringBuffer) init(log logger, size int) {
+	*b = ringBuffer{}
+	b.buf.init(size)
+	b.buf.logger = log
+}
+
+func (b *ringBuffer) insert(event publisher.Event, client clientState) (bool, int) {
+	// log := b.buf.logger
```
ruflin: Leftover?

libbeat/publisher/queue/memqueue/ringbuf.go
```diff
@@ -0,0 +1,293 @@
...
+
+		b.buf.Set(idx, event, client)
+		b.regB.size++
+
+		return true, avail - 1
+	}
+
+	// region B does not exist yet, check if region A is available for use
+	idx := b.regA.index + b.regA.size
+	// log.Debug("  - index: ", idx)
```
ruflin: Quite a few logging leftsovers in the code here and below I think.

libbeat/publisher/queue/memqueue/broker.go
urso: yep

libbeat/publisher/queue/memqueue/eventloop.go
urso: it's due to the old code. Some more cleanups will come at some point.

urso:
Debug Logging 'leftovers' are on purpose. It's some 'ExtraDebug' :)
The buffer handling is quite tricky and in case I need to debug super helpful. Just commented out by default as these create too much noise when enabling debug logging in beats.

libbeat/publisher/queue/memqueue/ringbuf.go
```diff
@@ -0,0 +1,293 @@
...
+// The ring is split into 2 regions.
+// Region A contains active events to be send to consumers, while region B can
+// only be filled by producers, if there is no more space in region A. Splitting
+// the ring buffer into regions enables the broker to send batches of type
+// []publisher.Event to the consumer without having to copy and/or grow/shrink the
+// buffers.
+type ringBuffer struct {
+	buf eventBuffer
+
+	regA, regB region
```
kvch: Just for the record: I think `regA` and `regB` could be renamed to `regConsumer` and `regProducer`, so the could can make more sense without the comments. But as it is just moved from a different a file I can live with it. It is already an improvement over the previous state.

urso: tl;dr: Only purpose of regB is to handle the filling the ring buffers when the indices in the ring buffer 'wrap around'.

regA is regConsumer and regProducer as long as there is space left in regA :)
regB only comes into existence if the ring buffer wraps around (no more space in regA).
Once regB exists, regA is indeed consumer only and regB is producer only.
After wrapping around regB becomes regA, once regA becomes empty. From then on regA is again shared by producers and consumers until another 'wrap around' on the ring buffer.

kvch: Yes, I have seen that. I was just trying to come up with names which could help the understanding more. But ofc, the proposed names were not descriptive either. :(

PR #5149, opened 2017-09-11
Reorder processors in publisher pipeline

urso:
- reorder processors, such that the `beat` namespace can not be removed from client processors
- remove support for EventsMetadataKey
- combine client internal field updates and configured field updates into one dictionary
- combine proessors for adding client and global fields+tags if no client processors are configured
- Do not Clone (deep copy) fields being added in the pipeline if no processors are configured.

### DIFF
--- a/libbeat/common/mapstr.go
+++ b/libbeat/common/mapstr.go
@@ -11,9 +11,8 @@ import (
 // Event metadata constants. These keys are used within libbeat to identify
 // metadata stored in an event.
 const (
-	EventMetadataKey = "_event_metadata"
-	FieldsKey        = "fields"
-	TagsKey          = "tags"
+	FieldsKey = "fields"
+	TagsKey   = "tags"
 )
 
 var (
--- a/libbeat/publisher/pipeline/pipeline.go
+++ b/libbeat/publisher/pipeline/pipeline.go
@@ -60,10 +60,14 @@ type pipelineProcessors struct {
 	// The pipeline its processor settings for
 	// constructing the clients complete processor
 	// pipeline on connect.
-	beatMetaProcessor  beat.Processor
-	eventMetaProcessor beat.Processor
-	processors         beat.Processor
-	disabled           bool // disabled is set if outputs have been disabled via CLI
+	fieldsProcessor beat.Processor
+	tagsProcessor   beat.Processor
+	fields          common.MapStr
+	tags            []string
+
+	processors beat.Processor
+
+	disabled bool // disabled is set if outputs have been disabled via CLI
 }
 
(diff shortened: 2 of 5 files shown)


libbeat/publisher/pipeline/pipeline.go
```diff
@@ -386,3 +367,36 @@ func (e *waitCloser) dec(n int) {
...
+	hasProcessors := processors != nil && len(processors.List) > 0
+	if hasProcessors {
+		tmp := &program{title: "global"}
+		for _, p := range processors.List {
+			tmp.add(p)
+		}
+		p.processors = tmp
+	}
+
+	fields := common.MapStr{}
```
ruflin: I suggest to move this to line 391 as it is only used there.

libbeat/common/mapstr.go
```diff
@@ -11,9 +11,8 @@ import (
 // Event metadata constants. These keys are used within libbeat to identify
 // metadata stored in an event.
 const (
-	EventMetadataKey = "_event_metadata"
```
ruflin: I assume this already has become obsolete in previous refactoring?

libbeat/publisher/pipeline/pipeline.go
```diff
@@ -386,3 +367,36 @@ func (e *waitCloser) dec(n int) {
 func (e *waitCloser) wait() {
 	e.events.Wait()
 }
+
+func makePipelineProcessors(
+	annotations Annotations,
+	processors *processors.Processors,
+	disabled bool,
```
kvch: Parameter `disabled` is not used in this function.

libbeat/common/mapstr.go
urso: The symbol still has been used in the processors pipeline. This PR removes support for `_event_metadata` in the pipeline -> can finally remove EventMetadataKey

libbeat/publisher/pipeline/pipeline.go
urso: Uh, thanks. It should be passed to `pipelineProcessors`.

libbeat/publisher/pipeline/pipeline.go
urso: Yeah. Will simplify fields initialization.

urso:
> What is the affect of this order change of processors?

In the end it moves the processor adding the `beat` namespace to an event to a very late step. This way only global processors can remove fields from the `beat` namespace (e.g. `beat.name`, `beat.hostname`) and so on.

libbeat/publisher/pipeline/pipeline.go
urso: done

PR #5154, opened 2017-09-11
Add support for post-update command

ruflin:
This command can be used in the beats to be run after `make update` is executed. This allows for example to modify the libbeat specific parts of the config file if needed.

The command is optional in a beat, meaning if it does not exist it won't return an error. Also if it exist and it returns an error, make update will still skip the error.


ruflin:
Jenkins failure seems to be unrelated.

PR #5161, opened 2017-09-12
Allow IndexPrefix to be set.

simitt:
This is a breaking change in how a new Beat is created, and adds the possibility to specify a different Index Pattern.
fyi @roncohen

### DIFF
--- a/filebeat/filebeat.reference.yml
+++ b/filebeat/filebeat.reference.yml
@@ -736,7 +736,7 @@ output.elasticsearch:
   # if no error is encountered.
   #slow_start: false
 
-  # Optional index name. The default index name is set to name of the beat
+  # Optional index name. The default index name is set to filebeat
   # in all lowercase.
   #index: 'filebeat'
 
--- a/heartbeat/heartbeat.reference.yml
+++ b/heartbeat/heartbeat.reference.yml
@@ -465,7 +465,7 @@ output.elasticsearch:
   # if no error is encountered.
   #slow_start: false
 
-  # Optional index name. The default index name is set to name of the beat
+  # Optional index name. The default index name is set to heartbeat
   # in all lowercase.
   #index: 'heartbeat'
 
--- a/libbeat/_meta/config.reference.yml
+++ b/libbeat/_meta/config.reference.yml
@@ -155,10 +155,10 @@ output.elasticsearch:
   # Number of workers per Elasticsearch host.
   #worker: 1
 
-  # Optional index name. The default is "beatname" plus date
-  # and generates [beatname-]YYYY.MM.DD keys.
+  # Optional index name. The default is "beat-index-prefix" plus date
+  # and generates [beat-index-prefix-]YYYY.MM.DD keys.
   # In case you modify this pattern you must update setup.template.name and setup.template.pattern accordingly.
-  #index: "beatname-%{[beat.version]}-%{+yyyy.MM.dd}"
+  #index: "beat-index-prefix-%{[beat.version]}-%{+yyyy.MM.dd}"
 
   # Optional ingest node pipeline. By default no pipeline will be used.
   #pipeline: ""
@@ -251,9 +251,9 @@ output.elasticsearch:
   # if no error is encountered.
   #slow_start: false
 
-  # Optional index name. The default index name is set to name of the beat
+  # Optional index name. The default index name is set to beat-index-prefix
   # in all lowercase.
-  #index: 'beatname'
+  #index: 'beat-index-prefix'
 
   # SOCKS5 proxy server URL
   #proxy_url: socks5://user:password@socks5-server:2233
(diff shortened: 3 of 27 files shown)


libbeat/cmd/modules.go
```diff
@@ -42,7 +42,7 @@ func GenModulesCmd(name, version string, modulesFactory modulesManagerFactory) *
 
 // Instantiate a modules manager or die trying
 func getModules(name, version string, modulesFactory modulesManagerFactory) ModulesManager {
-	b, err := instance.NewBeat(name, version)
+	b, err := instance.NewBeat(name, name, version)
```
ruflin: Could we leave it empty by default, means if it is `instance.NewBeat(name, "", version)` it will use the name automtically?

libbeat/scripts/Makefile
```diff
@@ -265,11 +266,11 @@ update: python-env collect
 	@echo "Updating generated files for ${BEAT_NAME}"
 
 	@# Update config files.
-	@cat _meta/beat.yml ${ES_BEATS}/libbeat/_meta/config.yml | sed -e "s/beatname/${BEAT_NAME}/g" > ${BEAT_NAME}.yml
+	@cat _meta/beat.yml ${ES_BEATS}/libbeat/_meta/config.yml | sed -e "s/beatname/${BEAT_NAME}/g;s/index-prefix/${BEAT_INDEX_PREFIX}/g" > ${BEAT_NAME}.yml
```
ruflin: I would prefer to call it `beat-index-prefix` to have it in sync with the ENV variable. Also I think the name is less likely to show up by accident somewhere else in the file.

libbeat/cmd/instance/beat.go
```diff
@@ -125,7 +125,7 @@ func Run(name, version string, bt beat.Creator) error {
 }
 
 // NewBeat creates a new beat instance
-func NewBeat(name, v string) (*Beat, error) {
+func NewBeat(name, indexPrefix, v string) (*Beat, error) {
```
ruflin: I suggest if `indexPrefix` is empty, we set it to `name`. Like this the param can be set to `""`.

simitt: will change

simitt:
Looking at the failing tests on Travis and Jenkins I'm not sure if they are connected to my changes. Could someone of the beats team please have a look and give me some feedback on this?

heartbeat/heartbeat.reference.yml
```diff
@@ -465,7 +465,7 @@ output.elasticsearch:
   # if no error is encountered.
   #slow_start: false
 
-  # Optional index name. The default index name is set to name of the beat
+  # Optional index name. The default index name is set to the index-prefix of the beat
```
ruflin: I'm think what we should name this? Probably best would be `beat-index-prefix` as before it was generic but having the the `beat-index-prefix` replacement inside should be valid.

ruflin:
For the failing tests. The failure on the generator tests are related: https://travis-ci.org/elastic/beats/jobs/275475569

The others don't look related.

filebeat/filebeat.reference.yml
```diff
@@ -736,7 +736,7 @@ output.elasticsearch:
   # if no error is encountered.
   #slow_start: false
 
-  # Optional index name. The default index name is set to name of the beat
+  # Optional index name. The default index name is set to the filebeat
```
ruflin: Probably the `the` should now be removed.

libbeat/_meta/config.reference.yml
```diff
@@ -620,14 +620,14 @@ output.elasticsearch:
 # Set to false to disable template loading.
 #setup.template.enabled: true
 
-# Template name. By default the template name is "beatname-%{[beat.version]}"
+# Template name. By default the template name is "beat-index-prefix-%{[beat.version]}"
 # The template name and pattern has to be set in case the elasticsearch index pattern is modified.
-#setup.template.name: "beatname-%{[beat.version]}"
+#setup.template.name: "beat-index-prefix-%{[beat.version]}"
```
ruflin: Can you have a look at line `612` and `605`. I assume we also need a change there. Sorry that I only spotted that now.

simitt: Regarding the `beat name` at 605, could you briefly explain why you think this needs to be chnaged? From the description this still should be the name of the beat.

Regarding line 612, this basically tells you to set an index name if you want to overwrite what you have in the dashboards. There is no concrete pattern in the template, just the naming of the setting itself with `setup.dashboards.index:`.  I think this is still valid.

ruflin: For 605: You are right. I think for apm the case of multiple dashboards in one zip is even highly unlikely.

For 612: I need to check the code here if by default it does nothing or if it overwrites it with `beatname-*`. We must make sure that for apm it takes the index pattern by default or does nothing which also works.

ruflin: Here is the line that concerns me: https://github.com/elastic/beats/blob/master/libbeat/dashboards/kibana_loader.go#L62

If the dashboards are loaded through ES, empty Config.Index always seems to be ignored, but not for the Kibana API. @monicasarbu What is the default for `Config.Index` here and should it be only "overwritten" if set not empty?

simitt: What I am saying is that 612 was empty alreay before my changes. So if there is a bug it should have nothing to do with this PR should it?

ruflin: I'm not worried about it having any affect on our beats as only there the default is changed from `apm-server-*` to `apm-*` and not consitent with the beat name.

PR #5174, opened 2017-09-13
Fix systemd unit file name in Debian package

praseodym:
Fixes duplicate daemon start after #5086.


tsg:
Thanks! Will test on the snapshots.

PR #5191, opened 2017-09-18
Updated testing env 5.6.1

tsg:


tsg:
beats-ci failed in the expected places (windows), and travis is green, so this is good to go.

PR #5199, opened 2017-09-19
Fix kibana tests

exekias:
Some jobs fail due to an error resolving `kibana` host on Docker. This PR tries to solve that


exekias:
I ran it 3 times and kibana test worked in all of them (some other things failed but will tackle them in different PRs), I would merge it and see if it happens again

ruflin:
I think the change is fine as Kibana waits for ES to be available anyway I think to be healthy.

PR #5200, opened 2017-09-19
Fix packetbeat flaky tests

tsg:
Moved the shutdown timeout after stopping the flows, but before stopping
the transaction publishing. It needs to be after stopping the flows because
the tests rely on the shutdown sequence to create flow reports.

### DIFF
--- a/packetbeat/beater/packetbeat.go
+++ b/packetbeat/beater/packetbeat.go
@@ -183,6 +183,12 @@ func (pb *packetbeat) Run(b *beat.Beat) error {
 	}
 
 	defer pb.transPub.Stop()
+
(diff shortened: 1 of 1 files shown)


packetbeat/beater/packetbeat.go
```diff
@@ -183,6 +183,14 @@ func (pb *packetbeat) Run(b *beat.Beat) error {
 	}
 
 	defer pb.transPub.Stop()
+
+	timeout := pb.config.ShutdownTimeout
+	if timeout > 0 {
+		defer func() {
```
andrewkroh: The closure around the Sleep doesn't look to be necessary.

PR #5201, opened 2017-09-19
Ensure docs are build locally just as they are in prod

watson:
By default all headlines with level 1 (`==`) will get its own page.

When setting `--chunk=1`, all headlines with level 2 (`===`) will get its own page as well and will be shown in the ToC like so:

screen shot 2017-09-19 at 08 58 49


ruflin:
@watson Thanks for the contribution.

PR #5214, opened 2017-09-20
Remove changeme password from the docs

dedemorton:
The following PR needs to be merged before #5214: https://github.com/elastic/docs/pull/239

I'm including the shared attributes to use the pwd attribute in our docs. However, the paths in our docs are still set in the individual book indexes. We will move to the shared attributes eventually.


ruflin:
LGTM. I merged the other PR first.

PR #5215, opened 2017-09-21
Add LICENSE.txt into the packages

tsg:
We had it in the repository, but wasn't deployed in the packages. This
corrects this. It also renamed LICENSE and NOTICE to LICENSE.txt/NOTICE.txt
to align with Elasticsearch and Kibana.


tsg:
Tested with `make package` locally for filebeat. Looks good from what I see.

PR #5218, opened 2017-09-21
Remove pkg folder before running vet

exekias:
:hand: do not merge, this is a test PR to check if we get rid of vet errors


exekias:
Confirmed that this: https://github.com/elastic/beats/pull/5219 was the issue

PR #5219, opened 2017-09-21
Clean pkg before vet

exekias:
This ensure we ignore cached compiled packages when doing vet, some
changes break type checking from the previous version.

In go 1.9 vet has a `-source` flag. From vet doc at https://golang.org/cmd/vet/:

  By default vet uses the object files generated by 'go install
  some/pkg' to typecheck the code. If the -source flag is provided, vet
  uses only source code.`

### DIFF
--- a/libbeat/scripts/Makefile
+++ b/libbeat/scripts/Makefile
@@ -103,7 +103,7 @@ crosscompile: $(GOFILES)
 
 .PHONY: check
 check: python-env ## @build Checks project and source code if everything is according to standard
-	@rm -rf ${GOPATH}/pkg/*/github.com/elastic/beats # TODO go 1.9 supports `-source` flag in go vet, remove this in favor of the flag
+	find ${GOPATH}
 	@go vet ${GOPACKAGES}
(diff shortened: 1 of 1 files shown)


libbeat/scripts/Makefile
```diff
@@ -103,6 +103,7 @@ crosscompile: $(GOFILES)
 
 .PHONY: check
 check: python-env ## @build Checks project and source code if everything is according to standard
+	@rm -rf ${GOPATH}/pkg # TODO go 1.9 supports `-source` flag in go vet, remove this in favor of the flag
```
ruflin: It's quiet intrusive that we remove the complete `pkg` folder. I wonder if this could have some side affects.

exekias:
jenkins retest this please

libbeat/scripts/Makefile
exekias: I'll change this to go/pkg/linux_amd64/github.com/elastic/beats/ only, does it work for you?

exekias:
we finally decided to put this in jenkins, where it belongs

PR #5220, opened 2017-09-21
Allow for more than one index file in docs folders.

simitt:
In the APM project we have multiple entrypoints for documentation within subdirectories of the docs folder. When running `make docs` locally currently only one of two entry point documents is built. 
This change allows to have multiple `index.asciidoc` documents within the `docs` folder and subdirectories.

### DIFF
--- a/script/build_docs.sh
+++ b/script/build_docs.sh
@@ -17,17 +17,19 @@ else
     echo "$docs_dir already exists. Not cloning."
 fi
 
+index_list="$(find ${GOPATH%%:*}/src/$path -name 'index.asciidoc')"
+for index in $index_list
+do
+  echo "Building docs for ${name}..."
+  echo "Index document: ${index}"
+  index_path=$(basename $(dirname $index))
+  echo "Index path: $index_path"
 
-index="${GOPATH%%:*}/src/${path}/index.asciidoc"
-
-echo "Building docs for ${name}..."
-echo "Index document: ${index}"
-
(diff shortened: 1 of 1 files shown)


script/build_docs.sh
```diff
@@ -17,17 +17,19 @@ else
...
+  echo "Building docs for ${name}..."
+  echo "Index document: ${index}"
+  index_path=$(basename $(dirname $index))
+  echo "Index path: $index_path"
+
+  dest_dir="$html_dir/${name}/${index_path}"
+  mkdir -p "$dest_dir"
+  params="--chunk=1"
+  if [ "$PREVIEW" = "1" ]; then
+    params="--chunk=1 -open chunk=1 -open"
```
ruflin: Not sure what will happen here as `-open` actually opens a browser window.

simitt: good point! 
I tried it out with two `index.asciidoc` files. The script builds the docs for the first one and opens it in the browser. Once this is interrupted the second one is built and opened in the browser. I thought this workflow is proper, but please let me know if you have concerns.

simitt:
As far as I can say the broken build should have nothing to do with this PR. Would be great though if someone of the beats team could verify this.

script/build_docs.sh
ruflin: Thanks for testing. SGTM.

ruflin:
Merged. Flaky tests are related to 2 other we are working on.

PR #5221, opened 2017-09-21
Fix flaky reload tests

ruflin:
Because of the new publisher, sometimes events were still written to disk after the reloading stopped message. This is fixed by setting the flush size to 1 so events are immidiately flushed.


exekias:
Thank you for taking this one!

PR #5222, opened 2017-09-21
Added the 5.x version of the Auditbeat dashboards

tsg:
Exported the current dashboards from master using Kibana 5.6 and stored
them in the 5.x folder of the module.

Also upgrades the 5x.yml testing env to use 5.6.1


tsg:
@andrewkroh FYI

PR #5223, opened 2017-09-21
Optimize kubernetes logs_path matcher

vjsamuel:


exekias:
I'm not very fan of adding a new config key for this, did you see any change in your performance or did a benchmark to see what's the gain?

ruflin:
@vjsamuel Could you share some background on why this change is needed?

vjsamuel:
this is not needed when autodiscovery comes in.

PR #5228, opened 2017-09-22
Add comment for default value of load balancing in filebeat.reference.yml

DdPerna:
- Add comment for default value of load balancing in filebeat.reference.yml
- Fix typo in outputconfig.asciidoc

### DIFF
--- a/libbeat/docs/outputconfig.asciidoc
+++ b/libbeat/docs/outputconfig.asciidoc
@@ -444,7 +444,7 @@ output.logstash:
 
 ===== `pipelining`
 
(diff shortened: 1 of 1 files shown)


tsg:
Thanks for opening the PR. The file that you modified, however, is a generated file. You should edit `libbeat/_meta/config.reference.yml` instead.

Also, may I ask you to sign the [CLA](http://www.elasticsearch.org/contributor-agreement/)? You can do it online and should only take a couple of minutes.

If you prefer, I can also take this changes in a new PR and close this one.

Issue #5209, opened 2017-09-20
Monitor filesystem usage in metricbeat


henrikno:
It would be nice if the metricbeat elasticsearch module could report file system usage as reported by Elasticsearch. It'd allow to alert when going low on disk and tracking usage over time.

```
      "fs" : {
        "timestamp" : 1491553781646,
        "total" : {
          "total_in_bytes" : 67371577344,
          "free_in_bytes" : 51743817728,
          "available_in_bytes" : 48291119104
        },
```

It is reported by ES stats here:
https://github.com/elastic/beats/blob/c2bd3efb2994606731b25e1a761b877f3b70dd25/metricbeat/module/elasticsearch/node_stats/_meta/test/node_stats.522.json#L355-L357

The `store.size.bytes` is not accurate enough when indices are closed or not loaded.

ruflin:
@henrikno Please have a look at https://github.com/elastic/beats/pull/5230

PR #5230, opened 2017-09-25
Add filesystem stats to ES node_stats metricset

ruflin:
Closes https://github.com/elastic/beats/issues/5209

### DIFF
--- a/metricbeat/module/elasticsearch/node_stats/_meta/data.json
+++ b/metricbeat/module/elasticsearch/node_stats/_meta/data.json
@@ -1,105 +1,118 @@
 {
     "@timestamp": "2016-05-23T08:05:34.853Z",
-    "beat": {
-        "hostname": "host.example.com",
-        "name": "host.example.com"
-    },
     "elasticsearch": {
-        "cluster": {
-            "name": "elasticsearch"
-        },
         "node": {
-            "name": "8zgOwJ24TMamDDH9amWINQ",
             "stats": {
-                "indices": {
-                    "docs": {
-                        "count": 0,
-                        "deleted": 0
-                    },
-                    "segments": {
-                        "count": 0,
-                        "memory": {
-                            "bytes": 0
-                        }
-                    },
-                    "store": {
-                        "size": {
-                            "bytes": 0
-                        }
-                    }
-                },
                 "jvm": {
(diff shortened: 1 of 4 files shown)


metricbeat/mb/testing/data_generator.go
```diff
@@ -62,7 +62,7 @@ func WriteEvents(f mb.EventsFetcher, t testing.TB) error {
 // This simulates the output of Metricbeat as if it were
 // 2016-05-23T08:05:34.853Z and the hostname is host.example.com.
```
kvch: I would be nicer if the change of timestamp is followed up.

metricbeat/module/elasticsearch/node_stats/_meta/data.json
```diff
@@ -1,105 +1,118 @@
 {
     "@timestamp": "2016-05-23T08:05:34.853Z",
```
kvch: Is this file is a test output? Why isn't the timestamp changed here? I am asking because I would like to understand this PR fully.

ruflin: Good question. I would also expect it to change. Let me check.

There is an additional issue here. @urso recently change the generation of the to our internal json library where an beat.Event can be passed. Unfortunately this has the side affect, that it also adds @metadata which I think is confusing and I remove it manually below. Also it changes indentation from 4 to 2 which changes all the files.

I would prefer to decouple the above issues from this PR that is why I did some manual tweaking.

metricbeat/mb/testing/data_generator.go
```diff
@@ -62,7 +62,7 @@ func WriteEvents(f mb.EventsFetcher, t testing.TB) error {
...
 func CreateFullEvent(ms mb.MetricSet, metricSetData common.MapStr) beat.Event {
-	startTime, err := time.Parse(time.RFC3339Nano, "2016-05-23T08:05:34.853Z")
+	startTime, err := time.Parse(time.RFC3339Nano, "2017-09-25T08:05:34.853Z")
```
ruflin: Let me revert this part of the change, so we can investigate in a follow up PR on why this change does not apply and how to cleanup / improve the generation.

PR #5232, opened 2017-09-25
Cherry-pick #5157 to master: Update outputconfig.asciidoc

tsg:
Cherry-pick of PR #5157 to master branch. Original message: 

Made it clear that we only support a single output config directive.

Only applies from 6.0 onwards.


tsg:
@ruflin green :)

PR #5235, opened 2017-09-25
Add jenkins scripts to repo

andrewkroh:
This adds the scripts that are executed by Jenkins to this repo. They
need to be added to other branches and kept in sync. As a temporary measure for the initial Jenkins setup these were stored in a gist and pulled by Jenkins on each execution. This made it simple to apply a change across all branches.

Once these are added to our active branches we can update the Jenkins config to execute the script from the repo rather than from the gist.


ruflin:
Merging as failures are not related. @andrewkroh Will you backport it to all the branches?

PR #5236, opened 2017-09-25
Add .go files style to editorconfig

pgilad:


ruflin:
We didn't have it in this in the .editorconfig so far as we format all the `.go` files with `goimports`. I'm not sure how goimports and editorconfig will compete on the formatting even though I think 4 tabs should be correct normally.

pgilad:
I use vim with editorconfig plugin for editing, so currently it had no idea that the project uses tabs for go files (there is no default). I guess this could be the same for IntelliJ and the rest of those bunch. Anyway if the project defines a clear style for `.go` files, it is a standard to add it to `editorconfig` as well

ruflin:
We use the Golang standard: https://blog.golang.org/go-fmt-your-code I don't know if this could be reproduced with `.editorconfig`?

pgilad:
I'm not a `golang` pro, but I don't think that declaring `.go` files with tabs will ruin it ;)

ruflin:
Best option to figure out is merging and see what happens.

PR #5245, opened 2017-09-26
Autodiscover

exekias:
This PR adds support for Autodiscover, and implements a first provider for Docker.

### Autodiscover

Autodiscover allows the user to define different providers to watch for system changes and instantiate new modules as these happen.

Providers watch for changes and emit events to a common bus, then the autodiscovery process detect situations when there is something we can monitor and instantiate a modules for it.

Bus messages contain useful information about what providers see, with some common well-known optional fields (provider specific fields may also be present):

- host - service host
- port - service port
- config - recommended config for this instance (from the user or environment)

#### Docker provider

Docker provider watches for docker container events, it supports config mapping from container metadata -> config templates, so new modules are created when a container starts. The available data for a container (that you can use in the config template) is:

- host
- port
- docker.container.id
- docker.container.name
- docker.container.image
- docker.container.labels

#### Testing

This should be enough to monitor any present and future redis instances that run on docker:
```
metricbeat.autodiscover:
  providers:
    - type: docker
      templates:
        - condition:
            equals.docker.container.image: redis
          config:
            - module: redis
              metricsets: ["info", "keyspace"]
              hosts: "${data.host}:${data.port}"
```

Or reading redis logs only:
```
filebeat.autodiscover:
  providers:
    - type: docker
      templates:
        - condition:
            equals.docker.container.image: redis
          config:
            - type: log
              paths:
                - /var/lib/docker/containers/${data.docker.container.id}/*.log
```

You can launch it like this:
```
docker run -p 6379:6379 redis
```
This PR is in progress, with some missing parts:
- [x] Revisit Docker IP/port mapping, do we really need exposed ports? probably not
- [x] Add support in filebeat
- [x] Revisit bus implementation, use interfaces (instead of MapStr)?
- [x] Add docker metadata to modules discovered by docker provider
- [x] System tests
- [x] Update metricbeat and filebeat YAML files
- [ ] Docs

### DIFF
--- a/filebeat/crawler/crawler.go
+++ b/filebeat/crawler/crawler.go
@@ -21,6 +21,8 @@ type Crawler struct {
 	prospectorConfigs   []*common.Config
 	out                 channel.Factory
 	wg                  sync.WaitGroup
+	ProspectorsFactory  cfgfile.RunnerFactory
+	ModulesFactory      cfgfile.RunnerFactory
 	modulesReloader     *cfgfile.Reloader
 	prospectorsReloader *cfgfile.Reloader
 	once                bool
@@ -53,27 +55,27 @@ func (c *Crawler) Start(r *registrar.Registrar, configProspectors *common.Config
 		}
 	}
 
+	c.ProspectorsFactory = prospector.NewRunnerFactory(c.out, r, c.beatDone)
 	if configProspectors.Enabled() {
 		c.prospectorsReloader = cfgfile.NewReloader(configProspectors)
-		runnerFactory := prospector.NewRunnerFactory(c.out, r, c.beatDone)
-		if err := c.prospectorsReloader.Check(runnerFactory); err != nil {
+		if err := c.prospectorsReloader.Check(c.ProspectorsFactory); err != nil {
 			return err
 		}
 
 		go func() {
-			c.prospectorsReloader.Run(runnerFactory)
+			c.prospectorsReloader.Run(c.ProspectorsFactory)
 		}()
 	}
 
+	c.ModulesFactory = fileset.NewFactory(c.out, r, c.beatVersion, pipelineLoaderFactory, c.beatDone)
 	if configModules.Enabled() {
 		c.modulesReloader = cfgfile.NewReloader(configModules)
-		modulesFactory := fileset.NewFactory(c.out, r, c.beatVersion, pipelineLoaderFactory, c.beatDone)
-		if err := c.modulesReloader.Check(modulesFactory); err != nil {
+		if err := c.modulesReloader.Check(c.ModulesFactory); err != nil {
 			return err
 		}
 
 		go func() {
-			c.modulesReloader.Run(modulesFactory)
+			c.modulesReloader.Run(c.ModulesFactory)
 		}()
 	}
 
@@ -86,7 +88,7 @@ func (c *Crawler) startProspector(config *common.Config, states []file.State) er
 	if !config.Enabled() {
 		return nil
 	}
-	p, err := prospector.New(config, c.out, c.beatDone, states)
+	p, err := prospector.New(config, c.out, c.beatDone, states, nil)
 	if err != nil {
 		return fmt.Errorf("Error in initing prospector: %s", err)
 	}
--- a/filebeat/fileset/factory.go
+++ b/filebeat/fileset/factory.go
@@ -42,7 +42,7 @@ func NewFactory(outlet channel.Factory, registrar *registrar.Registrar, beatVers
 }
 
 // Create creates a module based on a config
-func (f *Factory) Create(c *common.Config) (cfgfile.Runner, error) {
+func (f *Factory) Create(c *common.Config, meta *common.MapStrPointer) (cfgfile.Runner, error) {
 	// Start a registry of one module:
 	m, err := NewModuleRegistry([]*common.Config{c}, f.beatVersion, false)
 	if err != nil {
@@ -64,7 +64,7 @@ func (f *Factory) Create(c *common.Config) (cfgfile.Runner, error) {
 
 	prospectors := make([]*prospector.Prospector, len(pConfigs))
 	for i, pConfig := range pConfigs {
-		prospectors[i], err = prospector.New(pConfig, f.outlet, f.beatDone, f.registrar.GetStates())
+		prospectors[i], err = prospector.New(pConfig, f.outlet, f.beatDone, f.registrar.GetStates(), meta)
 		if err != nil {
 			logp.Err("Error creating prospector: %s", err)
 			return nil, err
@@ -109,4 +109,8 @@ func (p *prospectorsRunner) Stop() {
 	for _, prospector := range p.prospectors {
 		prospector.Stop()
 	}
+}
+
+func (p *prospectorsRunner) String() string {
+	return p.moduleRegistry.InfoString()
 }
--- a/filebeat/prospector/prospector.go
+++ b/filebeat/prospector/prospector.go
@@ -1,6 +1,7 @@
 package prospector
 
 import (
+	"fmt"
 	"sync"
 	"time"
 
@@ -36,6 +37,7 @@ func New(
 	outlet channel.Factory,
 	beatDone chan struct{},
 	states []file.State,
+	dynFields *common.MapStrPointer,
 ) (*Prospector, error) {
 	prospector := &Prospector{
 		config:   defaultConfig,
@@ -64,9 +66,10 @@ func New(
 	}
 
 	context := Context{
-		States:   states,
-		Done:     prospector.done,
-		BeatDone: prospector.beatDone,
+		States:        states,
+		Done:          prospector.done,
+		BeatDone:      prospector.beatDone,
+		DynamicFields: dynFields,
 	}
 	var prospectorer Prospectorer
 	prospectorer, err = f(conf, outlet, context)
@@ -140,4 +143,8 @@ func (p *Prospector) stop() {
 	} else {
 		p.prospectorer.Stop()
 	}
+}
+
+func (p *Prospector) String() string {
+	return fmt.Sprintf("prospector [type=%s, ID=%d]", p.config.Type, p.ID)
 }
--- a/libbeat/autodiscover/autodiscover.go
+++ b/libbeat/autodiscover/autodiscover.go
@@ -1,0 +1,215 @@
+package autodiscover
+
+import (
+	"github.com/elastic/beats/libbeat/autodiscover/meta"
+	"github.com/elastic/beats/libbeat/cfgfile"
+	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/common/bus"
+	"github.com/elastic/beats/libbeat/logp"
+
+	"github.com/mitchellh/hashstructure"
+)
+
+const debugK = "autodiscover"
+
+// TODO autodiscover providers config reload
+
+// Adapter must be implemented by the beat in order to provide Autodiscover
+type Adapter interface {
+	// TODO Hints
+
+	// CreateConfig generates a valid list of configs from the given event, the received event will have all keys defined by `StartFilter`
+	CreateConfig(bus.Event) ([]*common.Config, error)
+
+	// CheckConfig tests given config to check if it will work or not, returns errors in case it won't work
+	CheckConfig(*common.Config) error
+
+	// RunnerFactory provides runner creation by feeding valid configs
+	cfgfile.RunnerFactory
+
+	// StartFilter returns the bus filter to retrieve runner start triggering events
+	StartFilter() []string
+
+	// StopFilter returns the bus filter to retrieve runner stop triggering events
+	StopFilter() []string
+}
+
+// Autodiscover process, it takes a beat adapter and user config and runs autodiscover process, spawning
+// new modules when any configured providers does a match
+type Autodiscover struct {
+	bus       bus.Bus
+	adapter   Adapter
+	providers []Provider
+	runners   *cfgfile.Registry
+	meta      *meta.Map
+
+	startListener bus.Listener
+	stopListener  bus.Listener
+}
+
+// NewAutodiscover instantiates and returns a new Autodiscover manager
+func NewAutodiscover(name string, adapter Adapter, config *Config) (*Autodiscover, error) {
+	// Init Event bus
+	bus := bus.New(name)
+
+	// Init providers
+	var providers []Provider
+	for _, providerCfg := range config.Providers {
+		provider, err := ProviderRegistry.BuildProvider(bus, providerCfg)
+		if err != nil {
+			return nil, err
+		}
+		logp.Debug(debugK, "Configured autodiscover provider: %s", provider)
+		providers = append(providers, provider)
+	}
+
+	return &Autodiscover{
+		bus:       bus,
+		adapter:   adapter,
+		runners:   cfgfile.NewRegistry(),
+		providers: providers,
+		meta:      meta.NewMap(),
+	}, nil
+}
+
+// Start autodiscover process
+func (a *Autodiscover) Start() {
+	if a == nil {
+		return
+	}
+
+	logp.Info("Starting autodiscover manager")
+	a.startListener = a.bus.Subscribe(a.adapter.StartFilter()...)
+	a.stopListener = a.bus.Subscribe(a.adapter.StopFilter()...)
+
+	for _, provider := range a.providers {
+		provider.Start()
+	}
+
+	go a.startWorker()
+	go a.stopWorker()
+}
+
+func (a *Autodiscover) startWorker() {
+	for event := range a.startListener.Events() {
+		// This will happen on Stop:
+		if event == nil {
+			return
+		}
+
+		configs, err := a.adapter.CreateConfig(event)
+		if err != nil {
+			logp.Debug(debugK, "Could not generate config from event %v: %v", event, err)
+			continue
+		}
+		logp.Debug(debugK, "Got a start event: %v, generated configs: %+v", event, configs)
+
+		meta := getMeta(event)
+		for _, config := range configs {
+			hash, err := hashstructure.Hash(config, nil)
+			if err != nil {
+				logp.Debug(debugK, "Could not hash config %v: %v", config, err)
+				continue
+			}
+
+			err = a.adapter.CheckConfig(config)
+			if err != nil {
+				logp.Debug(debugK, "Check failed for config %v: %v, won't start runner", config, err)
+				continue
+			}
+
+			// Update meta no matter what
+			dynFields := a.meta.Store(hash, meta)
+
+			if a.runners.Has(hash) {
+				logp.Debug(debugK, "Config %v is already running", config)
+				continue
+			}
+
+			runner, err := a.adapter.Create(config, &dynFields)
+			if err != nil {
+				logp.Debug(debugK, "Failed to create runner with config %v: %v", config, err)
+				continue
+			}
+
+			logp.Info("Autodiscover starting runner: %s", runner)
+			a.runners.Add(hash, runner)
+			runner.Start()
+		}
+	}
+}
+
+func (a *Autodiscover) stopWorker() {
+	for event := range a.stopListener.Events() {
+		// This will happen on Stop:
+		if event == nil {
+			return
+		}
+
+		configs, err := a.adapter.CreateConfig(event)
+		if err != nil {
+			logp.Debug(debugK, "Could not generate config from event %v: %v", event, err)
+			continue
+		}
+		logp.Debug(debugK, "Got a stop event: %v, generated configs: %+v", event, configs)
+
+		for _, config := range configs {
+			hash, err := hashstructure.Hash(config, nil)
+			if err != nil {
+				logp.Debug(debugK, "Could not hash config %v: %v", config, err)
+				continue
+			}
+
+			if !a.runners.Has(hash) {
+				logp.Debug(debugK, "Config %v is not running", config)
+				continue
+			}
+
+			if runner := a.runners.Get(hash); runner != nil {
+				logp.Info("Autodiscover stopping runner: %s", runner)
+				runner.Stop()
+				a.runners.Remove(hash)
+			} else {
+				logp.Debug(debugK, "Runner not found for stopping: %s", hash)
+			}
+		}
+	}
+}
+
+func getMeta(event bus.Event) common.MapStr {
+	m := event["meta"]
+	if m == nil {
+		return nil
+	}
+
+	logp.Debug(debugK, "Got a meta field in the event")
+	meta, ok := m.(common.MapStr)
+	if !ok {
+		logp.Err("Got a wrong meta field for event %v", event)
+		return nil
+	}
+	return meta
+}
+
+// Stop autodiscover process
+func (a *Autodiscover) Stop() {
+	if a == nil {
+		return
+	}
+
+	// Stop listening for events
+	a.startListener.Stop()
+	a.stopListener.Stop()
+
+	// Stop providers
+	for _, provider := range a.providers {
+		provider.Stop()
+	}
+
+	// Stop runners
+	for hash, runner := range a.runners.CopyList() {
+		runner.Stop()
+		a.meta.Remove(hash)
+	}
+	logp.Info("Stopped autodiscover manager")
+}
--- a/libbeat/autodiscover/provider.go
+++ b/libbeat/autodiscover/provider.go
@@ -1,0 +1,86 @@
+package autodiscover
+
+import (
+	"fmt"
+	"strings"
+	"sync"
+
+	"github.com/elastic/beats/libbeat/cfgfile"
+	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/common/bus"
+	"github.com/elastic/beats/libbeat/logp"
+)
+
+// Provider for autodiscover
+type Provider interface {
+	cfgfile.Runner
+}
+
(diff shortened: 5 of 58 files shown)


libbeat/autodiscover/provider.go
```diff
@@ -0,0 +1,88 @@
...
+// NewRegistry creates and returns a new Registry
+func NewRegistry() *registry {
+	return &registry{
+		providers: make(map[string]ProviderBuilder, 0),
+	}
+}
+
+// AddProvider registers a new ProviderBuilder
+func (r *registry) AddProvider(name string, provider ProviderBuilder) error {
+	r.lock.Lock()
```
kvch: Why do you lock here? Are `init` functions called concurrently?

libbeat/common/bus/bus.go
```diff
@@ -0,0 +1,102 @@
...
+	"github.com/elastic/beats/libbeat/logp"
+)
+
+// Event sent to the bus
+type Event common.MapStr
+
+// Bus provides a common channel to emit and listen for Events
+type Bus interface {
+	// Emit an event to the bus
+	Emit(Event)
```
kvch: I think `Publish` would be a more expressive name than `Emit`, as it is a kind of Publish-Subscribe model.

libbeat/autodiscover/common/config.go
```diff
@@ -0,0 +1,63 @@
+package common
```
ruflin: I worry a bit that this will conflict with our libbeat common package in some places.

libbeat/autodiscover/providers/docker/config.go
```diff
@@ -0,0 +1,18 @@
+package docker
+
+import (
+	"github.com/elastic/beats/libbeat/autodiscover/common"
+	"github.com/elastic/beats/libbeat/processors/add_docker_metadata"
```
ruflin: Feels strange to include here a processor. Perhaps we should extract the TLS config part to a docker package.

libbeat/autodiscover/providers/docker/docker.go
```diff
@@ -0,0 +1,117 @@
+package docker
+
+import (
+	"github.com/elastic/beats/libbeat/autodiscover"
+	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/common/bus"
+	"github.com/elastic/beats/libbeat/logp"
+
+	// TODO move add_docker_metadata.Watcher to a common place
```
ruflin: :-D

libbeat/processors/add_docker_metadata/watcher.go
```diff
@@ -32,6 +33,12 @@ type Watcher interface {
 
 	// Containers returns the list of known containers
 	Containers() map[string]*Container
+
+	// ListenStart returns a bus listener to receive container started events, with a `container` key holding it
+	ListenStart() bus.Listener
```
ruflin: I wonder if we should put these in the same interface or one more specific to autodiscovery which is composeo f this Watcher interface.

ruflin:
@exekias I stumbled over the config a bit because of all the indentation it has. Here an alternative suggestion:

```
metricbeat.autodiscover:
  providers:
    - type: docker
      filter:
        docker-image: redis
        port: 6309
      config:
        - module: redis
          metricsets: ["info"]
          hosts: "${data.host}:${data.port}"
        - module: redis
          metricsets: ["info", "keyspace"]
          hosts: "${data.host}:${data.port}"
          period: 60s
    - type: docker
      conditions:
        - when.equals.docker-image: mysql
        - when.range.port.gte: 6000
      config:
        - module: mysql
          metricsets: ["stats"]
          hosts: "${data.host}:${data.port}"
```

The basic idea is to make it more obvious out of what the config will be generated and allow multiple `and` criterias / filters in a simple way. But at the same time provide an option to use our more powerful conditions coming from the processors for matching. 

I don't think we should provide the conditions in the first implementation but wanted to add it here to take into account where this could go. I'm not happy with `filter` naming in the first config part. Perhaps `criteria` or `match` could be an alternative?

I'm pretty sure the above config brings up other issues as I only focused on the config and not potential implications on the implementation itself.

libbeat/autodiscover/common/config.go
```diff
@@ -0,0 +1,63 @@
...
+					logp.Err("Error building config: %v", err)
+				}
+				opts := []ucfg.Option{
+					ucfg.PathSep("."),
+					ucfg.Env(vars),
+					ucfg.ResolveEnv,
+					ucfg.VarExp,
+				}
+				for _, config := range configs {
+					c, err := ucfg.NewFrom(config, opts...)
```
urso: Why do you need ucfg here?

Minor problem with this approach is, you are loosing the full setting paths. If some config unpacking fails, the setting it's full path is reported with the error message.

libbeat/autodiscover/provider.go
exekias: Nope, or not at the moment :innocent:, but some defensive code there won't hurt I think, this may be a thing in the future?

libbeat/autodiscover/common/config.go
exekias: I pass provider event data for variable expansion, so config templates can use data from the event. Agree to the minor issue (would affect logs mainly?)

vjsamuel:
@exekias on general thought. with autodiscover we should avoid using `add_docker_metadata` and `add_kubernetes_metadata`, we should provide all metadata upfront as part of the module configuration. This will improve performance tremendously as we would not need to do append of metadata at each event.

we should also have some plugability in place for folks who want to run something custom on top of auto discovery. this is similar to collectbeat appenders where we have our own internal appenders more custom tailored for our needs.

libbeat/autodiscover/common/config.go
urso: Yes, it's for logs only. As we can't store file positions (by line) yet, not having the correct names can make it more difficult to track-down misconfigurations.

vjsamuel: discovery can be more automatic than providing `docker-image: redis`. if we use docker labels, we should be able to pass `annotation` like information same like collectbeat to be able to discover workloads automatically. i could run 10 containers in a host. it would be a huge config if i have to say for each via config how it has to be discovered.

exekias: I agree @vjsamuel, this is a first PR, intended to bring the basic functionality and building blocks to grow from there. Definitely that's something we want to have

fin09pcap:
subscribe

exekias:
@ruflin I updated config format to take advantage of processor conditions, let me know what you think about it:
```
metricbeat.autodiscover:
  providers:
    - type: docker
      templates:
        - when.equals.docker-image: redis
          use:
            - module: redis
              metricsets: ["info", "keyspace"]
              hosts: "${data.host}:${data.port}"
```

christiangalsterer:
This is a really great feature as this really great for mid to large scale Docker/Kubernetes installations to auto-configure Beats when container come and go.
Do you have by chance any idea in which version and when this will be available? A first indication would already help if this is something we can expect in the next month, next quarter, etc.

exekias:
Hi @christiangalsterer, I'm glad you like it! this is something we want to have in the 6.X timeline, probably in a few months

ruflin:
We had a discussion with @exekias and @urso about the config and that is what we came up with. I added some comments to explain why the decision was made this way:

```
metricbeat.autodiscover:
  # Each Provider can be defined multiple times.
  # Example: Assuming we have a port scanner provider, different scan intervals are applied for different port ranges
  providers:
    - type: docker
    
      # Each provider can have it's specific config options
      host: docker/socket/path
      
      # Multiple templates can be used per provider with different condtions
      templates:
      
          # One condition is used per template. Sub conditions can be added with and to nest conditions
        - condition:
            and:
              - equals.docker.container.image: redis
        
          # The configuration fields that will be created
          config:
            - module: redis
              metricsets: ["info"]
              hosts: "${data.host}:${data.port}"
              frequency: 1m
            - module: redis
              metricsets: ["keyspace"]
              hosts: "${data.host}:${data.port}"
              frequency: 10s
              
          # Same config as above just starting with config instead of condition
        - config:
            - module: redis
              metricsets: ["info", "keyspace"]
              hosts: "${data.host}:${data.port}"
          condition:
            when.equals.docker.container.image: redis
```

We couldn't decide if it should be called `config` or `configs` but sticked with `config` for now.

filebeat/crawler/crawler.go
```diff
@@ -21,6 +21,8 @@ type Crawler struct {
 	prospectorConfigs   []*common.Config
 	out                 channel.Factory
 	wg                  sync.WaitGroup
+	ProspectorsFactory  cfgfile.RunnerFactory
```
ruflin: Note for myself: Need to understand why we need to store the factory now in the Crawler struct.

filebeat/fileset/factory.go
```diff
@@ -110,6 +110,7 @@ func (p *prospectorsRunner) Stop() {
 		prospector.Stop()
 	}
 }
-func (p *prospectorsRunner) ID() uint64 {
```
ruflin: Was the ID still used somewhere? If not, perhaps remove it in a different PR?

filebeat/prospector/prospector.go
```diff
@@ -141,3 +144,7 @@ func (p *Prospector) stop() {
 		p.prospectorer.Stop()
 	}
 }
+
+func (p *Prospector) String() string {
```
ruflin: Is that a required method or just for convenience / logging?

libbeat/autodiscover/autodiscover.go
```diff
@@ -0,0 +1,243 @@
...
+	// CreateConfig generates a valid list of configs from the given event, the received event will have all keys defined by `StartFilter`
+	CreateConfig(bus.Event) ([]*common.Config, error)
+
+	// CheckConfig tests given config to check if it will work or not, returns errors in case it won't work
+	CheckConfig(*common.Config) error
+
+	// RunnerFactory provides runner creation by feeding valid configs
+	cfgfile.RunnerFactory
+
+	// StartFilter returns the bus filter to retrieve runner start triggering events
```
ruflin: So far it seems these are all the same for the current implementation. How will these look for other adapters?

libbeat/autodiscover/autodiscover.go
```diff
@@ -0,0 +1,243 @@
...
+		bus:       bus,
+		adapter:   adapter,
+		runners:   cfgfile.NewRegistry(),
+		providers: providers,
+		meta:      make(map[uint64]*common.MapStrPointer),
+	}, nil
+}
+
+// Start autodiscovery process
+func (a *Autodiscover) Start() {
```
ruflin: Some idea that popped up in my head when reading this: Will it be needed to enabled / disable autodiscovery dynamically? For example on startup you want the beat to autodiscover all the services that are out there and then stop the discovery for example? Or only use autodiscovery for some time enabled through centralised config? We definitively not need that now, just thinking out loud.

libbeat/autodiscover/autodiscover.go
```diff
@@ -0,0 +1,243 @@
...
+				runner.Stop()
+				a.runners.Remove(hash)
+			} else {
+				logp.Debug(debugK, "Runner not found for stopping: %s", hash)
+			}
+		}
+	}
+}
+
+func getMeta(event bus.Event) *common.MapStr {
```
ruflin: Should this be a method on `bus.Event`?

libbeat/autodiscover/autodiscover.go
```diff
@@ -0,0 +1,243 @@
...
+
+func (a *Autodiscover) storeMeta(hash uint64, meta *common.MapStr) *common.MapStrPointer {
+	a.metaMutex.Lock()
+	defer a.metaMutex.Unlock()
+
+	if meta == nil {
+		return nil
+	}
+
+	if a.meta[hash] == nil {
```
ruflin: I wonder if the code would become simpler if `MetaList` or similar would be it's own struct so all these methods and locks would be on metaList instead of the Autodiscovery object.

libbeat/autodiscover/provider.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/elastic/beats/libbeat/logp"
+)
+
+// Provider for autodiscovery
+type Provider interface {
+	cfgfile.Runner
+}
+
+// Registry holds all known autodiscover providers, they must be added to it to enable them for use
+var Registry = NewRegistry()
```
ruflin: Should we call it ProviderRegistry?

libbeat/autodiscover/provider.go
```diff
@@ -0,0 +1,88 @@
...
+	logp.Debug(debugK, "Provider registered: %s", name)
+	return nil
+}
+
+// GetProvider returns the provider with the giving name, nil if it doesn't exist
+func (r *registry) GetProvider(name string) ProviderBuilder {
+	r.lock.RLock()
+	defer r.lock.RUnlock()
+
+	name = strings.ToLower(name)
```
ruflin: What is the reason you do the `ToLower`?

libbeat/autodiscover/providers/docker/config.go
```diff
@@ -0,0 +1,18 @@
+package docker
```
ruflin: I would call the package path `autodiscovery/provider/docker/*` instead of `providers`. It's messy across the code base but I think recently we tried to call it plural if it has multiple implementations directly in the package itself and singular if it has multiple sub packages with one implementation each.

libbeat/autodiscover/providers/docker/docker.go
```diff
@@ -0,0 +1,136 @@
...
+			"name":   container.Name,
+			"image":  container.Image,
+			"labels": container.Labels,
+		},
+	}
+
+	// Emit container info
+	d.emit(bus.Event{
+		flag:     true,
+		"host":   container.IPAddresses[0],
```
ruflin: can it happen that there is no IPAdress -> panic? Just double checking.

libbeat/autodiscover/providers/docker/docker.go
```diff
@@ -0,0 +1,136 @@
...
+		"meta": common.MapStr{
+			"docker": meta,
+		},
+	})
+
+	// Emit container private ports
+	for _, port := range container.Ports {
+		event := bus.Event{
+			flag:     true,
+			"host":   container.IPAddresses[0],
```
ruflin: see above

libbeat/cfgfile/registry.go
```diff
@@ -3,7 +3,7 @@ package cfgfile
 import "sync"
 
 type Registry struct {
-	sync.Mutex
+	sync.RWMutex
```
ruflin: I like that change. Could we move it to a separate PR to get it in quickly? Trying to reduce the size of this PR :-)

libbeat/common/bus/bus.go
```diff
@@ -0,0 +1,102 @@
...
+
+	// Subscribe to all events, filter them to the ones containing *all* the keys in filter
+	Subscribe(filter ...string) Listener
+}
+
+// Listener retrieves Events from a Bus subscription until Close is called
```
ruflin: until Stop is called?

libbeat/common/bus/bus.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+type listener struct {
+	filter  []string
+	channel chan Event
+	bus     *bus
+}
+
+// New initializes a new bus with the given name and returns it
+func New(name string) Bus {
```
ruflin: What are the names of the buses you have in mind?

libbeat/common/docker/watcher_test.go
```diff
@@ -12,15 +12,17 @@ import (
...
+	containers [][]types.Container
 	// event list to send on Events call
 	events []interface{}
 
 	done chan interface{}
 }
 
 func (m *MockClient) ContainerList(ctx context.Context, options types.ContainerListOptions) ([]types.Container, error) {
-	return m.containers, nil
+	res := m.containers[0]
```
ruflin: Can you elaborate on this change?

libbeat/common/mapstr_pointer.go
```diff
@@ -0,0 +1,29 @@
+package common
+
+import (
+	"sync/atomic"
+	"unsafe"
+)
+
+// MapStrPointer stores a pointer to atomically get/set a MapStr object
```
ruflin: I would have prefered to have "optimizations" like this in a follow up PR to have the code we put in as simple as possible to start with and then when we do optimizations we also add some benchmark tests for it. Makes it easier to later detect in case these optimizations causes some issues, if it happened with the optimization or the original code.

libbeat/processors/condition.go
```diff
@@ -42,6 +42,10 @@ type WhenProcessor struct {
 	p         Processor
 }
 
+type ValuesMap interface {
```
ruflin: Note to myself: Figure out why this change is needed and if it could happen in a separate PR.

filebeat/crawler/crawler.go
exekias: Short answer: we use this factory from Autodiscovery. In the long term, we should unify all prospector/module initialization in a common place

filebeat/fileset/factory.go
exekias: It wasn't used, it's a spurious implementation, probably forgot from a previous removal

filebeat/prospector/prospector.go
exekias: It's for logging purposes only, autodiscover will tell about new spawned prospectors/modules in the log output

libbeat/autodiscover/autodiscover.go
exekias: I don't know yet :innocent:, I think this code can be used in other parts of beats, not only to detect new modules to lauch, so wanted to keep this open

libbeat/autodiscover/autodiscover.go
exekias: this is a good point, something like modules reload could make sense for autodiscover providers

libbeat/autodiscover/autodiscover.go
exekias: bus.Event is quite generic and can be used in other parts, `meta` field is just a convention for autodiscover. I would prefer to keep bus slim until we see how it gets used

libbeat/autodiscover/autodiscover.go
exekias: Totally, will do it, thanks!

libbeat/autodiscover/provider.go
exekias: User input sanitizing, `type: Docker` would still work, I think I copied this from somewhere else but I'm not sure :P

libbeat/autodiscover/providers/docker/docker.go
exekias: good point, It should not happen, but will add some defensive code here

libbeat/common/bus/bus.go
exekias: I don't know, this name is used just for logging

libbeat/common/docker/watcher_test.go
exekias: I added support for several requests to this mock, each request returns the next response and pops it out the list

libbeat/common/mapstr_pointer.go
exekias: Agreed, but as I try to get a stable API to introduce new providers as soon as possible this somehow made sense while I waited for reviews. Once this is merged, autodiscover API should have everything we need

libbeat/autodiscover/autodiscover.go
urso: Do we even need the filters methods on the interface? Can the filters be passed when registering the adapter, so to keep the interface somewhat 'smaller'?

libbeat/beat/pipeline.go
```diff
@@ -34,6 +34,9 @@ type ClientConfig struct {
 	// Fields provides additional 'global' fields to be added to every event
 	Fields common.MapStr
 
+	// DynamicMeta provides additional fields to be added to every event, supporting live updates
+	DynamicMeta *common.MapStrPointer
```
urso: To reduce confusion, rather use `DynamicFields`. The beat event has `Meta` and `Fields`. The `Meta` are somewhat internal (like LS `@metadata`). But `DynamicMeta` is updating the event `Fields`, not `Meta`.

Btw. instead of adding the functionality to the publisher pipeline, you can just pass a processor doing the magic.

Instead of passing DynamicMeta, we could add atomic values support to the normalizer of `common.MapStr`. This would allow you to pass updatable fields via `Fields` (given you don't need them on top-level).

libbeat/common/mapstr_pointer.go
```diff
@@ -0,0 +1,29 @@
...
+// This should give faster access for use cases with lots of reads and a few
+// changes.
+// It's imortant to note that modifying the map is not thread safe, only fully
+// replacing it.
+type MapStrPointer struct {
+	p unsafe.Pointer
+}
```
urso: How about changing the type to:

```
type VolatileMapStr struct {
  p *unsafe.Pointer
}

func MakeVolatileMapStr(m MapStr) VolatileMapStr {
  return VolatileMapStr{ p: &unsafe.Pointer(&m) }
}

func (m VolatileMapStr) Get() MapStr {
  return *((*MapStr)(atomic.LoadPointer(m.p)))
}

func (m VolatileMapStr) Set(new MapStr) {
  atomic.StorePointer(m.p, unsafe.Pointer(&new))
}

```

Then you can use `VolatileMapStr` and `MapStr` by value, just like `MapStr` is already passed by value.

libbeat/beat/pipeline.go
exekias: Thanks for the input, I agree DynamicFields is the right name.

I thought about the processor but that would cause Copy to be called for all cases, I guess that's not too bad?

libbeat/common/mapstr_pointer.go
exekias: LGTM, but don't really like the Volatile name, somehow it hides the real thing?

libbeat/autodiscover/providers/docker/config.go
exekias: There is something I don't like from that approach. If I'm the top folder of these layout and I find `autodiscover/provider` I don't realize I have several different providers inside. What's the benefit of using singular for this case?

ruflin: If you read the full package name it reads that docker is only 1 provider. But this is definitively personal preference and something @urso and I came up with over time. Not something that would hold back this PR :-)

libbeat/autodiscover/autodiscover.go
exekias: The adapter is not registered but passed to Autodiscover Manager. If you look at is implementation, it's very small generally, I think it makes sense to keep these filters together with the functions that use it. If in the future we use something different from config we need to update the filters too https://github.com/exekias/beats/blob/411d13a878d64dbe62b6008057bbaa1c207f9e4e/metricbeat/beater/autodiscover.go

libbeat/autodiscover/provider.go
ruflin: I would suggest not to sanatise it and force users to do "correct" config files instead. Reducing magic :-)

exekias: Just found where I copied this from: https://github.com/elastic/beats/blob/354fdd6792b77ed5f099e017caab7c8ec4fa2ead/metricbeat/mb/registry.go#L77, I'm ok with removing it anyway

libbeat/cfgfile/registry.go
kvch: huge :+1: on reducing the size of this PR.

metricbeat/metricbeat.yml
```diff
@@ -19,6 +19,24 @@ metricbeat.config.modules:
   # Period on which files under path should be checked for changes
   #reload.period: 10s
 
+#============================== Autodiscover ===================================
+
+# Autodiscover allows you to detect changes in the system and spawn new modules
+# as they happen.
+
+metricbeat.autodiscover:
```
tsg: I think Autodiscovery doesn't need to be in the short config (which we try to keep minimal), but only in the reference one. Especially since Autodiscovery is going to be in the K8s deployment manifersts, right?

exekias: :+1: that works for me

exekias:
jenkins retest it

exekias:
This PR is ready for a (final?) review, I'm currently opening side PRs to shrink this one

1. https://github.com/elastic/beats/pull/5581

ruflin:
@exekias just merged the other two. Could you rebase on top? If it reduces the hassle of rebasing, feel free to squash it into 1 commit already (or as many as you want in the end).

tsg:
@exekias Merged, but please keep in mind to add docs before the 6.1 GA.

libbeat/autodiscover/provider.go
ruflin: Undecided on this one. Probably easier on our users. @andrewkroh WDYT?

libbeat/autodiscover/providers/docker/config.go
ruflin: @exekias you sneaked it in ;-)

ruflin:
@exekias Awesome stuff. Thanks for making this happen.

libbeat/autodiscover/provider.go
andrewkroh: As the person who made the registry case-insensitive I like that it accepts any case as input.

dedemorton:
Autodiscover docs were added with https://github.com/elastic/beats/pull/5868, so I'm removing the needs_docs label.

PR #5246, opened 2017-09-26
Migrate Sirupsen/logrus to sirupsen/logrus (#4642)

### LINKED ISSUE #4642: Fix logrus dependency
github.com/Sirupsen/logrus has moved to github.com/sirupsen/logrus, update dependencies accordingly.

liketic:
Resolves #4642 . I'm not sure if I need to update the 3rd party libraries which referenced Sirupsen/logrus for example docker.


andrewkroh:
I would think that any vendored code referencing the old name should be updated. Maybe I don't fully understand all the subtleties at work here, but I'm surprised things compile properly without updating the imports in the docker code.

liketic:
@andrewkroh Thanks. Yes, I didn't update the docker related package because I can compile it and run all tests. Seems like some packages are redundant. I can try to update the docker packages too.

ruflin:
I wonder if we should just change the import in all places where it's used and then let govendor to the magic?

liketic:
@ruflin Agreed and updated. Could you please review this? Thanks in advance.

vendor/github.com/docker/go-connections/tlsconfig/certpool_go17.go
```diff
@@ -6,7 +6,7 @@ import (
 	"crypto/x509"
 	"runtime"
 
-	"github.com/Sirupsen/logrus"
```
ruflin: Did you change this line manually? The problem now is that the version linked in the version.json file is not in sync with this version anymore. This seems to become a bit trickier.

liketic:
@ruflin Thanks for reviewing. The latest `github.com/docker/go-connections/tlsconfig` already removed `Sirupsen `, and only the `docker/docker/..` related library still use `Sirupsen `, we cannot migrate `sirupsen` to `Sirupsen` until `docker` migrated theirs dependency. Closing this without merging.

ruflin:
@liketic Thanks for the investigations. I wonder if docker would accept a PR :-)

PR #5247, opened 2017-09-26
Cherry-pick #5235 to 6.0: Add jenkins scripts to repo

andrewkroh:
Cherry-pick of PR #5235 to 6.0 branch. Original message: 

This adds the scripts that are executed by Jenkins to this repo. They
need to be added to other branches and kept in sync. As a temporary measure for the initial Jenkins setup these were stored in a gist and pulled by Jenkins on each execution. This made it simple to apply a change across all branches.

Once these are added to our active branches we can update the Jenkins config to execute the script from the repo rather than from the gist.


dev-tools/jenkins_ci.sh
```diff
@@ -0,0 +1,42 @@
...
+  exit
+fi
+
+# Setup Go.
+export GOPATH=${WORKSPACE}
+export PATH=${GOPATH}/bin:${PATH}
+if [ -f ".go-version" ]; then
+  eval "$(gvm $(cat .go-version))"
+else
+  eval "$(gvm 1.7.5)"
```
ruflin: We should probably update this to 1.8

dev-tools/jenkins_intake.sh
```diff
@@ -0,0 +1,27 @@
...
+: "${HOME:?Need to set HOME to a non-empty value.}"
+: "${WORKSPACE:?Need to set WORKSPACE to a non-empty value.}"
+
+# Setup Go.
+export GOPATH=${WORKSPACE}
+export PATH=${GOPATH}/bin:${PATH}
+if [ -f ".go-version" ]; then
+  eval "$(gvm $(cat .go-version))"
+else
+  eval "$(gvm 1.7.5)"
```
ruflin: Same here

dev-tools/jenkins_ci.sh
andrewkroh: We could probably remove that these scripts. I had it there for the time when we didn't have .go-version in all branches. But _I think_ we do now.

PR #5250, opened 2017-09-26
Switch to using SHA512 for checking the downloads

tsg:


tsg:
We're probably need to get this in 6.0 and master branches as well, but at the moment the snapshot builds are broken there, so there are still sha1 published.

PR #5251, opened 2017-09-26
Bump version to 5.6.3 in the 5.6 branch

tsg:


ruflin:
It seems the 5.6.2 snapshot is not around anymore. Not related to this change but we should fix it in the 5.6 branch to get the tests back to green.

ruflin:
The issue is actually not the snapshot is not available anymore, but that the files is now called sha512 and not sha1 anymore.

ruflin:
@tsg Just saw that you already fixed it. Thanks :-)

PR #5258, opened 2017-09-27
Change field type of http header from nested to object

ruflin:
The field type of http headers was set to nested instead of object. In metricbeat we normally do not used nested fields. Also nested fields are not compatible with the sorting on index time feature coming in 6.0.

The problem with indexing the headers is that it could lead to field explosion if there are many different headers. An alternative would be to not index the headers. For now my recommendation is if someone has too many headers, filters should be used to remove most of the entries before it is sent to Elasticsearch.


ruflin:
@exekias I'm struggling where to add this in the Changelog. I personally would probably treat this as a bug as I think `nested` was not intentional.

Will add it to the docs.

ruflin:
@exekias Pushed the docs change and added it to bugs for now.

exekias:
but LGTM too :+1:

PR #5260, opened 2017-09-27
improve scaffold for common/mapstr

PChou:
the following functions try to walk the MapStr according to the key path, and return converted type instead of interface{}. easier to deal with json

- GetValueNoErr: return nil instead of err
- GetStringValue,GetStringValueWithDefault: return converted string or ""
- GetInt64Value,GetInt64ValueWithDefault,GetFloat64Value,GetFloat64ValueWithDefault: return converted numeric or 0
- GetObject,GetObjectWithDefault: return nested MapStr or nil
- GetObjectArray,GetStringArray: return []MapStr or []string or nil
- MapStrUnmarshal: unmarshal from []byte to MapStr
- MarshalBinary: encoding.BinaryMarshaler implementation

### DIFF
--- a/libbeat/common/mapstr.go
+++ b/libbeat/common/mapstr.go
@@ -3,6 +3,8 @@ package common
 import (
 	"encoding/json"
 	"fmt"
+	"reflect"
+	"strconv"
 	"strings"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


karmi:
Hi @PChou, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/beats/pull/5260.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

PChou:
ci tells

```
docker-compose.yml is locked, waiting
--- FAIL: TestFetch (125.84s)
	compose.go:64: Timeout waiting for services to be healthy
time="2017-09-28T03:41:34Z" level=info msg="[0/24] [kibana]: Starting " 
time="2017-09-28T03:41:34Z" level=warning msg="Error while reading .dockerignore (/go/src/github.com/elastic/beats/metricbeat/module/kibana/_meta/.dockerignore) : open /go/src/github.com/elastic/beats/metricbeat/module/kibana/_meta/.dockerignore: no such file or directory" 
```
please help!

exekias:
You can ignore that particular error, it's a temporary fail when bringing up the test scenario

exekias:
Hi @PChou, could you please share a bit of context on the purpose of this helper functions? I guess you are you using these for a custom beat?

PChou:
@exekias thanks for your reply. Yes, when i build some custom beat(private for my company), i have to get some json data by http, and then parse the json, construct the beat's event according to the parsed result. 

Well, the `common/schema` implements a common `json to json` mapping, but not suitable for me. In my requirement, i have to store the last two parsed results, and calculate some float fields of them. so, i have to write huge codes like:

```
total, err := raw.GetValue("total")
if err != nil {
	return
}

if totalInt64, ok := total.(float64); ok {
	rs.total = int64(totalInt64)
}
```

with these helper functions, i can improve these code to:

```
rs.total = raw.GetInt64ValueWithDefault("total", 0)
```

The reason i don't use "json unmarshal to struct" is that the `raw` json is not stable, i means most fields are always dynamic, for example:

```
{
    "stats": {
        "74824783290368": 65466,
        "89923445067776": 12912,
        "89968267759616": 113480,
        "89993945055232": 28194,
        "90009396273152": 99810
    }
}
```

"schema mapping" and "unmarshal to struct" are both not suitable for me, so i want to add some helper functions to the `common.MapStr`. I believe that `common.MapStr` can be improved to annother [go-simplejson](https://github.com/bitly/go-simplejson)

exekias:
Awesome! thanks for the input, could you please add some more tests to improve coverage? some of the signatures are not tested

urso:
-1 from me on this PR. I totally see value in some of these kind of helper functions. But as `common.MapStr` is this 'universal' in beats we will have to maintain all code/helper on `common.MapStr` I'm not happy with the way these helper are added to `common.MapStr` ('fat object antipattern'). My problem with more and more MapStr utilities is, it becomes harder and harder to use `common.MapStr`. We already have a number of functions/methods with some non-subtle overlap I rather want to see getting cleaned up as well. Plus I want to keep the opportunity to replace `common.MapStr` with some other types for event handling in the future. The more direct methods/function we add on top of `common.MapStr`, the harder it will be.

I’d rather have common.MapStr to get a very minimal interface and have these (and other) utility methods turned into function. Maybe even in another package.

I can think of a few possible solutions to the 'interface' problem, while keeping the functionality provided by this PR. The main observation with your PR is, all wrap `GetValue`. First of all, let's turn all methods into hepler functions. Depending on the solution you choose, the difference is in the arguments you pass to your helpers:
1. have first parameter `common.MapStr`
2. replace `common.MapStr` with an interface like `type ValueGetter struct { GetValue(string) (interface{}, error) }`. Now code becomes a little more reusable
3. have `(interface{}, error)` as arguments + create a small DSL on top of it. Btw. the 'schema' library in metricbeat is some kind of descriptive DSL to extract and convert JSON data from external systems.

Personally I prefer option 3. The 'DSL' relies on the fact that one can always call `f(g)` in golang, as long as the number of arguments of `f` and return values of `g` are exactly the same.

e.g.

```
package get // alternative package names: as, to, conv

var (
  errNoString = errors.New("value is no string")
)

func DefaultNil(value interface{}, err error) (interface{}) {
  if err != nil {
    return nl
  }
  return value
}

// converters

func String(value interface{}, err error) (string, error) {
  if err != nil {
    return "", err
  }
  if s, ok := value.(string); ok {
    return s, nil
  }
  return errNoString
}

func Int(value interface{}, err error) (int64, error) {
  ...
}

...

// default value support

func StringDefault(default string) func(interface{}, error) string {
  return func(value interface{}, err error) string {
    s, err := String(value, err)
    if err != nil {
      return default
    }
    return s
  }
}

func IntDefault(default int64) func(interface{}, error) int64 {
  ...
}

```

Also replace Marshal/Unmarshal helpers with functions.

e.g. your sample code will become

```
total, _ := get.Int(raw.GetValue("total"))
```

alternative one can implement the package like this:

```
package conv

var (
  errNoString = errors.New("value is no string")
)

type Val struct { 
  value interface{}
  err error
}

func Value(v interface{}, err error) Val {
  return Val{v, err}
}

func (v Val) ToString() (string, error) {
  if v.err != nil {
    return "", v.err
  }
  if s, ok := v.value.(string); ok {
    return s, nil
  }
  return "", errNoString
}

func (v Val) ToStringDefault(df string) string {
  if s, err := v.String(); err == nil {
    return s
  }
  return df
}

func (v Val) ToInt() (int64, error) {
  ...
}

func (v Val) ToIntDefault(df int64) int64 {
  ...
}

```

Now your sample becomes:

```
total, _ := conv.Value(raw.GetValue("total")).ToInt()
// or 
total := conv.Value(raw.GetValue("total")).ToIntDefault(0)
```

Btw. as this approach just relies on `(interface{}, error)` tuple, this must not be necessarily added to beats. You can have your own project/repo providing these helpers. Plus you can use these helpers with other libraries/projects as well.

I can't stress enough the importance of errors and zero-values. In a number of cases your helpers 'swallow' the error message. That is, one can not really tell the reason for the value returned. E.g. is the value actually missing or do we have a type mismatch or is the value present but 'nil' in the incoming JSON document. If one needs to check for type mismatch, one can not use the helpers provided (which might result in more boilerplate). Plus it seems in a number of cases one wants the golang 'Zero'-value (e.g. int 0 or empty string). This later case easily done (and becomes apparent in code), by just ignoring the error as I did in my sample.

ph:
Sorry for the long delay, I am going to close this for now feel free to reopen with @urso suggestion.

PR #5270, opened 2017-09-29
Add modules.d info to metricbeat docs

### LINKED ISSUE #5207: Document modules.d layout for Metricbeat
I can probably reuse the content that I've already added for Filebeat here: 

https://www.elastic.co/guide/en/beats/filebeat/master/configuration-filebeat-modules.html#enable-modules-d-configs

dedemorton:
Closes #5207

### DIFF
--- a/metricbeat/docs/metricbeat-options.asciidoc
+++ b/metricbeat/docs/metricbeat-options.asciidoc
@@ -1,30 +1,85 @@
 [[configuration-metricbeat]]
 == Specify which modules to run
 
-To enable specific modules and metricsets, you add entries to the `modules` list in of the +{beatname_lc}.yml+ config file. Each entry in the list begins with a dash (-) and is followed by settings for that <<metricbeat-modules,module>>.
+Metricbeat provides a couple different ways to enable modules and metricsets:
 
-//REVIWERS: Using "list" here instead of "array" to match the terminology we use in the Platform Ref to describe the config file
+* <<enable-modules-d-configs>>
+* <<enable-modules-config-file>>
+
+[float]
+[[enable-modules-d-configs]]
+=== Enable module configs in the `modules.d` directory
+
+The `modules.d` directory contains default configurations for all the modules
+available in Metricbeat. You can enable or disable specific module
+configurations  under `modules.d` by running the <<modules-command,`modules
+enable` or `modules disable`>> commands.
+
+For example, to enable the `apache` and `mysql` configs in the `modules.d`
+directory, you use:
+
+[source,shell]
+----
+./metricbeat modules enable apache mysql
+----
+
+Then when you run Metricbeat, it loads the corresponding module configurations
+specified in the `modules.d` directory (for example, `modules.d/apache.yml` and
+`modules.d/mysql.yml`).
+
+To see a list of enabled and disabled modules, run:
+
+[source,shell]
+----
+./metricbeat modules list
+----
+
+You can change the default module configurations by modifying the `.yml` files
+in the `modules.d` directory. 
 
 The following example shows a basic configuration for the Apache module:
 
 [source,yaml]
 ----
-metricbeat:
-  modules:
-    - module: apache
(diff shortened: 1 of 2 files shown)


metricbeat/docs/metricbeat-options.asciidoc
```diff
@@ -1,30 +1,85 @@
...
+  fields:
+    dc: west
+  tags: ["tag"]
+  processors:
+  ....
 ----
 
+// REVIEWERS: We aren't very consistent about when to show enabled: true. I'm
+// assuming this option is available for all modules, but wondering if we should
+// show it in the config.
```
exekias: We had some discussion about `enabled: true`, the idea is that having that it's the default and we provide an alternative to manage enabled/disabled status (renaming the yml file), it looks like a good idea to hide that so we don't provide to ways of doing the same thing

metricbeat/docs/metricbeat-options.asciidoc
```diff
@@ -45,90 +100,107 @@ metricbeat.modules:
...
-      metricsets: ["info"]
-      hosts: ["host1"]
-      period: 1s
-      enabled: true
-    - module: apache
-      metricsets: ["info"]
-      hosts: ["host1"]
-      period: 30s
-      enabled: true
+metricbeat.modules:
```
exekias: is this line still wanted? I'm not sure from the context

metricbeat/docs/metricbeat-options.asciidoc
```diff
@@ -45,90 +100,107 @@ metricbeat.modules:
...
+You can specify a module configuration that uses different combinations of
+metricsets, periods, and hosts. 
+
+For a module with multiple metricsets defined, it's possible to define the
+module twice and specify a different period to use for each metricset. For the
+following example, the `set1` metricset will be fetched every 10 seconds, while
+the `set2` metricset will be fetched every 2 minutes:
+
+// REVIEWERS: I'm assuming the following config is still supported if you are
+// using the modules.d layout. Is that correct?
```
exekias: it's correct :+1:

metricbeat/docs/metricbeat-options.asciidoc
```diff
@@ -1,30 +1,85 @@
...
+// show it in the config.
 
-The following example shows a configuration where the Apache and MySQL modules are enabled:
+See <<config-combos>> for additional configuration examples.
+
+[float]
+[[enable-modules-config-file]]
+=== Enable module configs in the +{beatname_lc}.yml+ file
+
+// REVIEWERS: I think the following statement is true.
```
exekias: I would say yes, but want @tsg opinion here

metricbeat/docs/metricbeat-options.asciidoc
dedemorton: yah, because this sections shows how to configure multiple modules in the metricbeat.yml file. I don't want users to get confused and add this to one of the config files under modules.d.

metricbeat/docs/metricbeat-options.asciidoc
dedemorton: Ok, I'll remove it from the examples then.

metricbeat/docs/metricbeat-options.asciidoc
dedemorton: @tsg Is the following statement true? Sounds like we would really prefer for users to use modules.d, right?

dedemorton:
@exekias I've removed `enabled: true` from the examples (and deleted the questions flagged for reviewers). Do you have any other comments? Just waiting to hear back from tsg and get your LGTM. Thanks!

exekias:
:+1: on my side

metricbeat/docs/metricbeat-options.asciidoc
tsg: Yes, it's true, we should guide people to use the `modules.d` folder.

Issue #4899, opened 2017-08-15
Add support for script fields in fields.yml


ruflin:
Some dashboards in Kibana require script fields. As `fields.yml` currently does not support script fields and the index pattern is generated from `fields.yml`, these dashboards do not work. Either the `fields.yml` should support to define script fields or we should find an other automated way on how to load them.

The advantage of using `fields.yml` is that then these fields could potentially also shown in the docs or used in other places.

An example script fields looks as following after exporting:

In the `fields` string in the index pattern:
```
{
	"name": "error id icon",
	"type": "string",
	"count": 0,
	"scripted": true,
	"script": "doc['error.checksum'].value",
	"lang": "painless",
	"searchable": true,
	"aggregatable": true,
	"readFromDocValues": false
}
```

Under `fieldFormatMap` in the index pattern:

```
"error id icon":{  
   "id":"url",
   "params":{  
      "labelTemplate":"-"
   }
}
```

monicasarbu:
That's a great addition to the `fields.yml`. I suggest to use the field name as in Kibana, so `language` instead of `lang`. This should support to add a `type` and a `format` for the scripted fields.

ruflin:
I suggest in the config options to `skip` the language part and only support `painless`. It seems I even missed part for the data that has to be generated as in Kibana there are also options like `Url Template` which should be taken into account.

simitt:
Scripted fields are generally supported now. Additional attributes can be added as they are needed, so I suggest to close this issue.

PR #5271, opened 2017-09-29
Move Field(s) to common.

simitt:
Make type Field(s) reusable and seperate from transformation logic.

This is the first part for implementing #4899.

This PR moves the type `Field` and `Fields` to package `common` and seperates the types from the ES transformation logic. This is a first step to reuse the data structure for transforming information from the `fields.yml` also for Kibana index pattern creation.

Changes:
- `template.Field` -> `common.Field`
- `template.Fields` -> `common.Field`
- combined `template.processing` logic from `template.Fields` and `template.Field` in one file `template.Processor`
- adopted existing tests to work with the new structure but NOT adding or removing any tests.


libbeat/template/field.go
```diff
@@ -1,238 +0,0 @@
...
-	DocValues      *bool       `config:"doc_values"`
-
-	path      string
-	esVersion common.Version
-}
-
-// This includes all entries without special handling for different versions.
-// Currently this is:
-// long, geo_point, date, short, byte, float, double, boolean
-func (f *Field) other() common.MapStr {
```
simitt: This logic and the other methods regarding transformation moved to `template.Processor`.

libbeat/template/field.go
```diff
@@ -1,238 +0,0 @@
...
-// The mapping requires "properties" between each layer. This is added here.
-func generateKey(key string) string {
-	if strings.Contains(key, ".") {
-		keys := strings.SplitN(key, ".", 2)
-		key = keys[0] + ".properties." + generateKey(keys[1])
-	}
-	return key
-}
-
-type dynamicType struct{ value interface{} }
```
simitt: moved to `common.Field`

libbeat/template/field_test.go
```diff
@@ -1,325 +0,0 @@
-package template
```
simitt: the tests are mostly moved over to `template/processor_test`, tests regarding unpacking are moved to `common/field_test.go`

libbeat/template/fields.go
```diff
@@ -1,120 +0,0 @@
...
-		}
-	}
-
-	return nil
-}
-
-// HasKey checks if inside fields the given key exists
-// The key can be in the form of a.b.c and it will check if the nested field exist
-// In case the key is `a` and there is a value `a.b` false is return as it only
-// returns true if it's a leave node
```
simitt: moved to `common.Field`

libbeat/template/processor.go
```diff
@@ -0,0 +1,259 @@
...
+		}
+
+		if len(mapping) > 0 {
+			output.Put(common.GenerateKey(field.Name), mapping)
+		}
+	}
+	return nil
+}
+
+func (p *Processor) other(f *common.Field) common.MapStr {
```
ruflin: I personally preferred that these were methods of Field as all the parts in this methods are applied to on the passed field anyway. What is the reason you changed this logic? Also this increases the number of params to be passed.

simitt: I think it makes sense to make the `fields` structure the same and therefore reusable between different processors. 
This is the reason I would not bind processor specific logic to the data structure itself. The logic that is generally applicable for transformation to a ES template and also to a Kibana index pattern could be moved to the `common.fields`. However, I might do this in a seperate step if we encounter processor logic that is generally applicable. 

Another advantage is that this makes it easier to test when you pass a data structure into a method instead having a large interface with lots of methods.

ruflin: Got it. I think we could perhaps somehow compose the overall fields out of `TemplateFields` and `KibanaFields` so the TemplateFields would still keep it's method and not have to be passed each time. But as they overlap this could become trickier.

I think I miss the part on how it simplifies testing.

I suggest we move forward as is right now as functionality will be identical. We can still do refactoring when we have all the changes in.

simitt: E.g some of the methods use ` p.EsVersion` for decisions on how to transform. I would not consider an ES Version to be a general attribute of a `Field`. 
I fully agree though that we should share logic between `ES` and `Kibana` specific transformations when it overlaps or depends on each other.

PR #5272, opened 2017-09-29
Move index gen to go

simitt:
This PR is the second part of refactoring the logic for creating the Kibana Index Pattern, and re-implements the generation logic from python to go and adds tests.

I added the former index pattern json files as `old` files and the new ones and created a  [test](https://github.com/elastic/beats/compare/master...simitt:move-index-gen-to-go?expand=1#diff-fadb89edc74f7ee4f4f893c6709fa9e0R93 that compares them and ensures the attributes and values are exactly the same. 
It is added in a seperate commit, that can be removed again when the overall refactoring is done. 

@monicasarbu cc


libbeat/common/field.go
```diff
@@ -31,12 +33,30 @@ type Field struct {
 	Dynamic        DynamicType `config:"dynamic"`
 	Index          *bool       `config:"index"`
 	DocValues      *bool       `config:"doc_values"`
+	Analyzed       *bool       `config:"analyzed"`
```
ruflin: Could you add a line here `// Kibana specific fields` and also move Pattern down here?

libbeat/kibana/index.go
```diff
@@ -0,0 +1,120 @@
...
+	BeatDir   string
+	BeatName  string
+
+	fieldsYaml       string
+	targetDirDefault string
+	targetDir5x      string
+	targetFilename   string
+}
+
+func (i *Index) Create() ([]string, error) {
```
ruflin: I assume CI is going to complain that there are no comments here.

libbeat/kibana/index_test.go
```diff
@@ -0,0 +1,174 @@
...
+		{Index: Index{Version: "6.0", IndexName: "beat-index", BeatName: "mybeat."}},
+		{Index: Index{Version: "6.0", IndexName: "beat-index", BeatDir: beatDir}},
+	}
+	for idx, test := range tests {
+		err := test.Index.init()
+		msg := fmt.Sprintf("(%v): Should have raised error", idx)
+		assert.Error(t, err, msg)
+	}
+
+	//checks for fields.yml
```
ruflin: Should we move these tests to a separate test to not mix with the table tests above?

libbeat/kibana/testdata/5.x-libbeat.json
```diff
@@ -0,0 +1,6 @@
+{
```
ruflin: Nice. Like this we see in the future when we make changes on what happens.

libbeat/kibana/transform.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/elastic/beats/libbeat/common"
+)
+
+func TransformFields(timeFieldName string, title string, commonFields common.Fields) common.MapStr {
+	fields := []common.MapStr{}
+	fieldFormatMap := common.MapStr{}
+
+	transformFields(commonFields, &fields, fieldFormatMap, "")
+
+	//add some meta fields
```
ruflin: nit picking: I'm used to  a space after `//`

ruflin:
CI error is not related. It fails because the stack has moved to sha256 check files. I assume you don't have the most recent version of beats master.

I suggest to merge this PR and adress comments in one of the follow up PR's.

libbeat/_meta/kibana/5.x/index-pattern/libbeat.old.json
```diff
@@ -0,0 +1,6 @@
+{
```
monicasarbu: Is the `libbeat.old.json` file committed by mistake?

libbeat/kibana/testdata/5.x-libbeat.json
```diff
@@ -0,0 +1,6 @@
...
+  "fieldFormatMap": "{\"group_enabled.message\":{\"id\":\"url\",\"params\":{\"inputFormat\":\"string\"}}}",
```
monicasarbu: I suggest instead of keeping two files `libbeat.json` and `libbeat.old.json` under `libbeat/_meta/kibana/5.x/index-pattern/`, to have the two files under the `kibana/testdata` directory together with a `fields.yml` file.  For a better test, I suggest to use the JSON file and the `fields.yml` from a Beat, instead of libbeat, maybe Metricbeat as it's the most complex one.

libbeat/kibana/testdata/fields.yml
```diff
@@ -0,0 +1,21 @@
+- key: test
```
monicasarbu: See my comment above. I suggest to use one of the `fields.yml` from a Beat, maybe Metricbeat?

ruflin: I like to have a simple test file for the tests. 

If we use a field.yml also from an external beats for more extensive tests, we should copy it here and not link it from metricbeat as there should not be dependencies from libbeat to a beat.

libbeat/kibana/index.go
ruflin: hm, interestingly I haven't seen it recently, so it seems it's not executed anymore.

simitt:
Thank you @monicasarbu for this great suggestion of using `metricbeat` files in the the test folder. I changed the [test](https://github.com/simitt/beats/blob/b33602a109fd24032bb4ee2f130efe31248bfe8a/libbeat/kibana/index_test.go#L81) accordingly.

libbeat/kibana/testdata/fields.yml
simitt: I copied the file from metricbeat and checked that the new generator creates the exact same output. 
Since in the next PR I want to add bugfixes and logic enhancement I'd remove this complex test file again and make a smaller more explicit test. But for this PR I think it is a great idea to have a complex test file ensuring I didn't bring in any unintended behavior.

libbeat/kibana/index.go
```diff
@@ -0,0 +1,121 @@
...
+	targetDirDefault string
+	targetDir5x      string
+	targetFilename   string
+}
+
+// Create the Index-Pattern for Kibana for 5.x and default.
+func (i *Index) Create() ([]string, error) {
+	indices := []string{}
+
+	err := i.init()
```
ruflin: The pattern we often follow in Beats is to have `NewIndex` function which takes 4 params in this case and returns an `Index` instance. Then `Create` would be run on the object as you have here. So this would probably move the `init` part to the `NewIndex` function.

But that is refactoring we can still do later.

libbeat/kibana/index.go
```diff
@@ -0,0 +1,121 @@
...
+
+	commonFields, err := common.LoadFieldsYaml(i.fieldsYaml)
+	if err != nil {
+		return nil, err
+	}
+	transformed := TransformFields("@timestamp", i.IndexName, commonFields)
+
+	if fieldsBytes, err := json.Marshal(transformed["fields"]); err != nil {
+		return nil, err
+	} else {
```
ruflin: I assume the else here is needed because you only assign `fieldsBytes` above. My personal preference would be not have the else but make the call outside the if and then only check `if err != nil { ` 

Same below.

libbeat/kibana/index.go
simitt: thanks for pointing that out, I will change it in the follow up PR then.

libbeat/kibana/index.go
```diff
@@ -0,0 +1,121 @@
...
+
+	fieldsYaml       string
+	targetDirDefault string
+	targetDir5x      string
+	targetFilename   string
+}
+
+// Create the Index-Pattern for Kibana for 5.x and default.
+func (i *Index) Create() ([]string, error) {
+	indices := []string{}
```
ruflin: this could be moved to line 57 so no allocation happens in case of errors or just have `var indicies []string` here.

libbeat/kibana/index.go
```diff
@@ -0,0 +1,121 @@
...
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"regexp"
+
+	"github.com/elastic/beats/libbeat/common"
+)
+
+type Index struct {
+	Version   string
```
ruflin: Reminds me a lot of `beat.Info` (https://github.com/elastic/beats/blob/master/libbeat/beat/info.go#L6) but Dir is missing there.

PR #5273, opened 2017-09-29
Fix mapping for ceph num_remapped_pgs field

ruflin:
The field num_remapped_pgs was mapped as num_in_osds which lead to a field defined twice and the mapping of the field did not exist.

@simitt Thanks for finding this


simitt:
This is a breaking change. Since it is a bugfix I'd be fine with merging it though.

PR #5283, opened 2017-10-02
Fix empty dashboard due to empty query

pmusa:
Currently the dashboard is not showing anything by default because the dashboard configuration has an empty query in it. One can easily go around it by adding a * to the query bar.


tsg:
Pinging also @simianhacker. This might have been intentional so that one is forced to select a host. But since we don't yet have a very easy way to select a host, it's probably a good idea to accept this change.

I'm adding it to our team agenda for discussion.

simianhacker:
Argh... I'm not sure how to answer this one. If I had the option to make that a "hidden" dashboard I totally would.

tsg:
This one was superseded by #5340. Thanks @pmusa!

PR #5290, opened 2017-10-02
Align kibana index generation to beats standard.

simitt:
* Move generator script to dev-tools.
* Add NewGenerator method which returns struct instead of init method.


libbeat/kibana/index_pattern_generator.go
```diff
@@ -11,43 +10,56 @@ import (
...
 	targetDirDefault string
 	targetDir5x      string
 	targetFilename   string
 }
 
-// Create the Index-Pattern for Kibana for 5.x and default.
-func (i *Index) Create() ([]string, error) {
-	indices := []string{}
+// Create an instance of the Kibana Index Pattern Generator
+func NewGenerator(indexName, beatName, beatDir, version string) (IndexPatternGenerator, error) {
```
ruflin: We normally return a reference here, this also make it possible to return `nil, err` on line 28.

libbeat/kibana/transform.go
```diff
@@ -7,46 +7,62 @@ import (
...
+type Transformer struct {
+	fields                    common.Fields
+	transformedFields         []common.MapStr
+	transformedFieldFormatMap common.MapStr
+	timeFieldName             string
+	title                     string
+	keys                      common.MapStr
+}
+
+func NewTransformer(timeFieldName, title string, fields common.Fields) Transformer {
```
ruflin: Same comment as above that we could return a reference. Didn't check if it is required in this case. It seems to work just fine :-)

PR #5295, opened 2017-10-03
Implement logic for disabled groups

simitt:
If a group is set to `enable: false` to not add it or fields within its namespace to the kibana index pattern.

### DIFF
--- a/libbeat/kibana/transform_test.go
+++ b/libbeat/kibana/transform_test.go
@@ -283,3 +283,71 @@ func TestTransformFieldFormatMap(t *testing.T) {
 		assert.Equal(t, test.expected, out, fmt.Sprintf("Failed for idx %v", idx))
 	}
 }
+func TestTransformGroupAndEnabled(t *testing.T) {
+	tests := []struct {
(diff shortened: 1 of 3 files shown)


libbeat/kibana/transform_test.go
```diff
@@ -283,3 +283,71 @@ func TestTransformFieldFormatMap(t *testing.T) {
...
+						},
+					},
+				},
+			},
+			expected: []string{"context.type", "context.metric.object"},
+		},
+		{
+			commonFields: common.Fields{
+				common.Field{Name: "enabledField"},
+				common.Field{Name: "disabledField", Enabled: &falsy}, //enabled is ignored for Type!=group
```
ruflin: Makes sense. I wonder if for this case we should even return an error as we already know it's false? Or is this only for Kibana?

simitt:
failing error on travis looks unrelated

libbeat/kibana/transform_test.go
simitt: This is only for kibana

PR #5296, opened 2017-10-03
Add multifields support.

simitt:
- Allow adding multifield values from `field.yml` with different `name` and `type`. 
- Adding a simple test file for better overview and moving metricbeat test file to `exhaustive` testing.


ruflin:
WFG

PR #5297, opened 2017-10-03
Allow setting values

simitt:
Allow setting values from outside. If no value given apply the default values. (based on PR #5296 ).


ruflin:
Merged as libbeat green on Travis. Jenkins has still some issues.

PR #5298, opened 2017-10-03
Add scripts support.

simitt:
Support scripted fields.

### DIFF
--- a/libbeat/common/field.go
+++ b/libbeat/common/field.go
@@ -34,10 +34,11 @@ type Field struct {
 	DocValues      *bool       `config:"doc_values"`
 
 	// Kibana specific
-	Analyzed     *bool `config:"analyzed"`
-	Count        int   `config:"count"`
-	Searchable   *bool `config:"searchable"`
-	Aggregatable *bool `config:"aggregatable"`
+	Analyzed     *bool  `config:"analyzed"`
+	Count        int    `config:"count"`
+	Searchable   *bool  `config:"searchable"`
+	Aggregatable *bool  `config:"aggregatable"`
+	Script       string `config:"script"`
 	// Kibana field format specific
 	Pattern         string `config:"pattern"`
 	InputFormat     string `config:"input_format"`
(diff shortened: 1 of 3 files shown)


libbeat/kibana/transform_test.go
```diff
@@ -189,9 +189,20 @@ func TestTransformMisc(t *testing.T) {
...
 		{commonField: common.Field{}, expected: false, attr: "scripted"},
+		{commonField: common.Field{}, expected: nil, attr: "script"},
+		{commonField: common.Field{Script: "doc[]"}, expected: true, attr: "scripted"},
+		{commonField: common.Field{Script: "doc[]"}, expected: "doc[]", attr: "script"},
+
+		// language
+		{commonField: common.Field{}, expected: nil, attr: "lang"},
+		{commonField: common.Field{Language: "lucene"}, expected: nil, attr: "lang"},
+		{commonField: common.Field{Language: "lucene", Script: "doc[]"}, expected: "lucene", attr: "lang"},
+		{commonField: common.Field{Script: "doc[]"}, expected: "painless", attr: "lang"},
```
simitt: @simianhacker I saw that in the index-pattern you used in `apm-server` the `readFromDocValues`  was set to `false` for scripted fields. Thus I added the logic here to set the field to false. Does this make sense?

You used `readFromDocValues` rather than `doc_values` - which one should we use for the Kibana index pattern?

libbeat/common/field.go
```diff
@@ -34,12 +34,15 @@ type Field struct {
...
 
 	// Kibana specific
-	Pattern      string `config:"pattern"`
 	Analyzed     *bool  `config:"analyzed"`
 	Count        int    `config:"count"`
 	Searchable   *bool  `config:"searchable"`
 	Aggregatable *bool  `config:"aggregatable"`
-	InputFormat  string `config:"input_format"`
+	Script       string `config:"script"`
+	Language     string `config:"lang"`
```
ruflin: I would suggest we only support painless script for now so we could hard code this.

simitt: changed

ruflin: Do we still need this entry if it is always `painless`

ruflin:
@simitt This needs a rebase.

PR #5300, opened 2017-10-03
Run make misspell

ruflin:
* Exclude .json files from misspell as Kibana files contain `seperate_axis` which is misspelled but part of Kibana and should not be fixed.
* Fix Transfered to Transferred manually in Kibana files

PR #5305, opened 2017-10-03
Add Pre-Commit hook.

simitt:
Since @tsg was mentioning a `lint bot` in PR #5300 I added this pre-commit handling. This script is not nearly as advanced as a lint bot, but I generally like pre-commit hooks and maybe someone else wants to also make use of it.

The script basically fetches the unstaged changes, then runs several commands and compares the unstaged changes again. If one of the commands fails or files were changed, it does not run the commit. This ofc does not work if someone changes a file while the commit hook is running. 

Running `make init` copies the script `script/pre_commit.sh` to the git pre-commit dir and makes it executable. Afterwards for every commit this hook is run automatically. If someone wants to get rid of the hook again, just remove the hook, `rm .git/hooks/pre-commit`. If the hook should be changed, just change the `script/pre-commit.sh` and run `make init` again.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,12 @@ REVIEWDOG_REPO=github.com/haya14busa/reviewdog/cmd/reviewdog
 
 # Runs complete testsuites (unit, system, integration) for all beats with coverage and race detection.
 # Also it builds the docs and the generators
+
+.PHONY: setup-commit-hook 
+setup-commit-hook:
+	@cp script/pre_commit.sh .git/hooks/pre-commit
+	@chmod 751 .git/hooks/pre-commit
+
 .PHONY: testsuite
 testsuite:
 	@$(foreach var,$(PROJECTS),$(MAKE) -C $(var) testsuite || exit 1;)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -15,6 +15,12 @@ REVIEWDOG_REPO=github.com/haya14busa/reviewdog/cmd/reviewdog
 
 # Runs complete testsuites (unit, system, integration) for all beats with coverage and race detection.
 # Also it builds the docs and the generators
+
+.PHONY: init
+init:
```
ruflin: can we call this `setup-commit-hook` or something like this? Otherwise I think this gets confused with `init` for the `beats` project which I don't think it is.

simitt: sure! 
I also don't mean to enforce this on the beats team, just a suggestion.

ruflin: I think it's great to have this in they way it is as it makes it optional. I see it especially useful for people new to the project. Instead of having to tell them all the commands to run, they only need to run one to setup the pre-commit hook.

tsg:
@elastic/beats FYI, see this commit, you can run `make setup-commit-hook` to install a common pre-commit hook. Will also mention it in the team meeting today.

PR #5306, opened 2017-10-03
Remove post-update command

ruflin:
The command was initially designed for the apm-server to modify the index pattern. But now a nicer solution was found and this is not needed anymore.


ruflin:
jenkins, test it

ruflin:
Failing tests are not related.

PR #5315, opened 2017-10-04
Introduce jenkins_setup command to be reused

ruflin:
jenkins_setup in common.bash will also be used for the generator tests.

### DIFF
--- a/dev-tools/common.bash
+++ b/dev-tools/common.bash
@@ -78,3 +78,21 @@ setup_go_path() {
 
   debug "GOPATH=${GOPATH}"
 }
+
+jenkins_setup() {
+  : "${HOME:?Need to set HOME to a non-empty value.}"
+  : "${WORKSPACE:?Need to set WORKSPACE to a non-empty value.}"
+
+  # Setup Go.
+  export GOPATH=${WORKSPACE}
+  export PATH=${GOPATH}/bin:${PATH}
+  if [ -f ".go-version" ]; then
+    eval "$(gvm $(cat .go-version))"
+  else
+    eval "$(gvm 1.7.5)"
+  fi
+
+  # Workaround for Python virtualenv path being too long.
+  export TEMP_PYTHON_ENV=$(mktemp -d)
+  export PYTHON_ENV="${TEMP_PYTHON_ENV}/python-env"
+}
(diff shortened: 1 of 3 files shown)


ruflin:
jenkins, retest it

dev-tools/jenkins_setup.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+
+#!/usr/bin/env bash
+set -euox pipefail
+
+: "${HOME:?Need to set HOME to a non-empty value.}"
```
andrewkroh: I think this could go into [common.bash](https://github.com/elastic/beats/blob/ce9e47b83ab5a7dd09c99832cdcf0ebdbc485cb0/dev-tools/common.bash) as a function like `jenkins_setup`.

dev-tools/jenkins_setup.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+
+#!/usr/bin/env bash
```
andrewkroh: You have two hashbangs.

dev-tools/jenkins_setup.sh
```diff
@@ -0,0 +1,20 @@
...
+set -euox pipefail
```
andrewkroh: If this is going to be used a library by other scripts then I would omit this and the hashbangs.

dev-tools/jenkins_setup.sh
ruflin: removed

dev-tools/jenkins_setup.sh
ruflin: removed

ruflin:
@andrewkroh I updated the PR accordingly

dev-tools/common.bash
```diff
@@ -78,3 +78,21 @@ setup_go_path() {
...
+  : "${HOME:?Need to set HOME to a non-empty value.}"
+  : "${WORKSPACE:?Need to set WORKSPACE to a non-empty value.}"
+
+  # Setup Go.
+  export GOPATH=${WORKSPACE}
+  export PATH=${GOPATH}/bin:${PATH}
+  if [ -f ".go-version" ]; then
+    eval "$(gvm $(cat .go-version))"
+  else
+    eval "$(gvm 1.7.5)"
```
andrewkroh: We could remove this now that `.go-version` should be on this branch.

ruflin: Will do that in a follow up PR.

PR #5318, opened 2017-10-04
Fix OutputPrecision in kibana-index-generation script.

simitt:
The current implementation treats the `OutputPrecision` as a `String`, but it actually should be an `Integer`. This is a bugfix for that.


simitt:
@ruflin failing test does not seem to be related.

Issue #5308, opened 2017-10-03
Beats should not enforce `setup.template.name` and `pattern` if templates are disabled.


pmusa:
When you set `output.elasticsearch.index: "my_index"` Beats force you to also set `setup.template.name` and `setup.template.pattern` (`Exiting: setup.template.name and setup.template.pattern have to be set if index name is modified.`). However, if you don't want Beats to manage templates and set `setup.template.enabled: false` it still complains about the settings above, but it should not.

kurtado:
Maybe these could be added to the default config and just left commented out? Would be much easier for users to figure out and fix.

ruflin:
I opened a PR here to fix the issue: https://github.com/elastic/beats/pull/5319

@kurtado If this will be a common issue that people change the index pattern, we should add it to the default config options to make it easily modifiable. I'm kind of hoping the default is not changed that often as also the setting to change the ES index is not in the default config.

PR #5319, opened 2017-10-04
Do not require template if index change and template disabled

ruflin:
If the elasticsearch index name is changed, its required to set a template. The error message was also shown when template itself are disabled. This is now changed that if templates are disabled, no error is returned.

Closes https://github.com/elastic/beats/issues/5308


tsg:
Changelog entry? Or too small of a change?

ruflin:
@tsg Good point, definitively should have one as it's a bug fix. Added.

PR #5320, opened 2017-10-04
metricbeat: add more node metrics in elasticseach module

wangdisdu:
metricbeat: add more node metrics in elastics each module

node:

- jvm.memory.heap.init
- jvm.memory.heap.max
- jvm.memory.nonheap.init
- jvm.memory.nonheap.max
- process.mlockall

### DIFF
--- a/metricbeat/module/elasticsearch/node/data.go
+++ b/metricbeat/module/elasticsearch/node/data.go
@@ -16,10 +16,26 @@ var (
 		"jvm": c.Dict("jvm", s.Schema{
 			"version": c.Str("version"),
 			"memory": c.Dict("mem", s.Schema{
-				"heap_init": s.Object{
-					"bytes": c.Int("heap_init_in_bytes"),
+				"heap": s.Object{
+					"init": s.Object{
+						"bytes": c.Int("heap_init_in_bytes"),
+					},
+					"max": s.Object{
+						"bytes": c.Int("heap_max_in_bytes"),
+					},
+				},
+				"nonheap": s.Object{
+					"init": s.Object{
+						"bytes": c.Int("non_heap_init_in_bytes"),
+					},
+					"max": s.Object{
+						"bytes": c.Int("non_heap_max_in_bytes"),
+					},
 				},
 			}),
+		}),
+		"process": c.Dict("process", s.Schema{
+			"mlockall": c.Bool("mlockall"),
 		}),
 	}
 )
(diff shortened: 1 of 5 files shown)


ruflin:
@wangdisdu Could you share some more details on how you use these metrics in your environment?

wangdisdu:
@ruflin Well, i build many charts with [echart](http://echarts.baidu.com/) in use of node_stats metrics, inculde the charts show jvm memory used by `heap` and `nonheap`. i will add two base lines into them: `init line`, and `max line`. So, i can know it maybe in unhealthy state when the used line is approaching the `max line`, and it is healthy in `init line` nearby. 

And the process.mlockall is very important for me. if it is `false`, i need to fix that by modify system evironment and restart elasticsearch.

ruflin:
@wangdisdu Thanks for sharing the details. How does the data from beats end up in echarts? Is that on top of elasticsearch?

We are looking into which values we should add to elasticsearch module without adding too many values. We will keep you posted.

metricbeat/module/elasticsearch/node/data.go
```diff
@@ -16,11 +16,27 @@ var (
...
+				"heap": s.Object{
+					"init": s.Object{
+						"bytes": c.Int("heap_init_in_bytes"),
+					},
+					"max": s.Object{
+						"bytes": c.Int("heap_max_in_bytes"),
+					},
+				},
+				"nonheap": s.Object{
+					"init": s.Object{
```
ruflin: Note for myself: This is a breaking change.

metricbeat/module/elasticsearch/node/data.go
```diff
@@ -16,11 +16,27 @@ var (
 		"jvm": c.Dict("jvm", s.Schema{
 			"version": c.Str("version"),
 			"memory": c.Dict("mem", s.Schema{
-				"heap_init": s.Object{
-					"bytes": c.Int("heap_init_in_bytes"),
+				"heap": s.Object{
+					"init": s.Object{
```
ruflin: Note for myself: this is a breaking change.

ruflin:
@wangdisdu I think we should move forward with this PR. Could you add a line to the CHANGELOG under breaking changes as the `heap.init` which was `heap_init` is a breaking change?

wangdisdu:
@ruflin Thanks for your reviewed. I has added the breaking change log into CHANGELOG: Rename `heap_init` field to `heap.init`.

By the way, the [echarts](http://echarts.baidu.com/) is just a javascript libaray.  The echarts can be used to build beautiful chart with the metric data that beats collect and store into elasticsearch.

ruflin:
jenkins, test it

ruflin:
@wangdisdu Thanks for your contribution.

Issue #4581, opened 2017-06-29
Beats features status


monicasarbu:
- Beats:
    - [ ] Auditbeat (Experimental - in 6.0 is build on top of Metricbeat)
    - [ ] Heartbeat (in 6.0 is Beta) https://github.com/elastic/beats/issues/3952
    - [ ] Packetbeat, Filebeat, Metricbeat, Winlogbeat (GA)
- [ ] Filebeat modules (Beta) -> GA (in 6.0)
- [ ] Filebeat/Metricbeat config reloading (in 6.0 is Beta) -> GA
- Processors:
     - [ ] add_docker_metadata (in 6.0 is Beta)
     - [ ] add_kubernetes_metadata (in 6.0 is Beta)
- [ ] Expose go metrics (Doesn't exist)

cc-ed @brandonmensing @ruflin

PR #5321, opened 2017-10-04
Deprecate config_dir (Filebeat) and refactor docs

tsg:
This deprecates the config_dir option in Filebeat, as it's functionality is
covered by `filebeat.config.prospectors`. There's a bigger refactoring in the
docs, to reflect that the `filebeat.config.*` and `metricbeat.config.*` options
can be used for both including external files and live reloading. Only the
live reloading part stays in Beta, including external config files is GA.

Part of #4581.

### DIFF
--- a/filebeat/docs/reload-configuration.asciidoc
+++ b/filebeat/docs/reload-configuration.asciidoc
@@ -2,13 +2,12 @@
 == Load external configuration files
 
 Filebeat can load external configuration files for prospectors and modules,
-which allows you to separate your configuration in multiple smaller
-configuration files. To use this functionality, use the `path` option from the
-`filebeat.config.prospectors` (for prospector configs) or
-`filebeat.config.modules` (for module configs) settings.
+which allows you to separate your configuration into multiple smaller
+configuration files. See the <<load-prospector-config>> and the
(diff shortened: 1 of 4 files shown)


filebeat/docs/reload-configuration.asciidoc
```diff
@@ -1,39 +1,17 @@
...
-changes. When the files found by the Glob change, new prospectors and/or
-modules are started and stopped according to changes in the configuration files.
-
-This feature is especially useful in container environments where one container
-is used to tail logs for services running in other containers on the same host.
-
-To enable dynamic config reloading, you specify the following options under
+Filebeat can load external configuration files for prospectors and modules,
+which allows you to separate your configuration in multiple smaller
+configuration files. To use this functionality, use the `path` option from the
```
dedemorton: I wonder if we really need to mention the options here since we describe them more completely later. Might actually be a bit confusing. 

Maybe just add links to the sections that follow and say:

To use this functionality, specify the `path` option when you configure the <<reload-prospector-config,prospector>> or <<reload-module-config,module>>.

BTW, I would probably rename the anchors to load-prospector-config and load-module-config.

Issue #5278, opened 2017-09-29
Metricbeat 6.0 Dashboards are not compatible with Kibana Master (7.x)


simianhacker:
Kibana changed the format of the query argument for SavedObjects. This is causing the 6.0 dashboards to not work in Kibana 7.x (master). We will need to install separate dashboards for 5.x, 6.x and 7.x

tsg:
@simianhacker Wouldn't be possible to do the translation at import time, in the Kibana plugin? I imagine that Kibana will need anyway a plan to upgrade existing dashboards, like they have for the 5 to 6 transition, right?

urso:
@ruflin @simianhacker Can this be closed?

ruflin:
@urso Closing as only MB 6.7 will work with 7.0.

PR #5328, opened 2017-10-05
Rename Kibana dashboard directories

ruflin:
Currently the dashboard directories are called 5.x and default. The naming default is a bit confusing as it is not clear what it means. In the current context it means the dashboards for all 6.x and 7.x releases. Unfortunately already now there are breaking changes in 7.x which means default does not work for both: https://github.com/elastic/beats/issues/5278

To allow us more flexibility the directory 5.x is renamed to 5 and the directory default is renamed to 6. This allows as to add a directory for 7 dashboards in the future. The reason I switched away from the `.x` naming is as this naming scheme also allows us to introduce for example 6.2 dashboards, meaning dashboards which only work with 6.2 and upwards in parallel to the 6 dashboards which work with all 6 releases.

TODO:
* [x] Rename all kibana directories in modules
* [x] Rename all kibana directories in all beats

@monica As a test I applied the change only to filebeat. To fully work all directories in our beats would have to be renamed. If we agree to move on with this, will update the PR with these changes too.


ruflin:
The implementation could be changed, that `default` directory is a fallback. Meaning if the `6` or `6.2` directory is not found, it will try to find `default` directory.  This would make sure assuming we ship this with `6.1` that if someone keeps his `6.0`  directory with the `default` folder structure, it still works. A similar thing would have to be done for the `5.x` directory.

But as I we ship the dashboards with each release, I would expect people that load dashboards to use the one shipped with the binary so I do not see this as a breaking change.

monicasarbu:
Yes, agree. This doesn't seem to be a breaking change, so we can add it to the `6.1` release.

libbeat/kibana/index_pattern_generator.go
```diff
@@ -40,6 +40,7 @@ func NewGenerator(indexName, beatName, beatDir, beatVersion string, version comm
 
 // Create the Index-Pattern for Kibana for 5.x and default.
 func (i *IndexPatternGenerator) Generate() (string, error) {
+
```
ruflin: I hope @andrewkroh doesn't see this 🙈

andrewkroh: See what? 😉

ruflin:
@monicasarbu @simitt This should be ready to go now.

@simitt Please also have a look as it will have an affect on apm and the naming of the directories there.

ruflin:
@monicasarbu Could you please have a look at this one? I need to keep rebasing this as it affects a lot of files. Would be good to get this in rather soonish to make sure everything keeps working.

ruflin:
@monicasarbu Rebase again on master and hopefully should go green now. Please merge if ok.

ruflin:
jenkins, test it

PR #5329, opened 2017-10-05
Support versioned urlTemplates in Kibana index pattern.

simitt:
UrlTemplates can differ between versions. This adds support for defining
urlTemplates with a version in fields.yml.


libbeat/common/field.go
```diff
@@ -39,17 +39,22 @@ type Field struct {
...
 	Aggregatable *bool  `config:"aggregatable"`
 	Script       string `config:"script"`
-	// Kibana field format specific
-	Pattern         string `config:"pattern"`
-	InputFormat     string `config:"input_format"`
-	OutputFormat    string `config:"output_format"`
-	OutputPrecision *int   `config:"output_precision"`
-	LabelTemplate   string `config:"label_template"`
-	UrlTemplate     string `config:"url_template"`
+	//Kibana params
```
ruflin: You didn't like the space? :-D

libbeat/common/version.go
```diff
@@ -56,6 +56,15 @@ func (v *Version) IsMajor(major int) bool {
 	return major == v.Major
 }
 
+// Equal returns true if v is exactly the same as v1.
+// The meta part is not taken into account.
+func (v *Version) Equal(v1 *Version) bool {
+	if v.Major == v1.Major && v.Minor == v1.Minor && v.Bugfix == v1.Bugfix {
```
ruflin: Would `v.version == v1.version` also work? Or compare `.String()`?

libbeat/common/version_test.go
```diff
@@ -45,6 +45,61 @@ func TestVersion(t *testing.T) {
...
+			result:   false,
+		},
+		{
+			name:     "5.4.3 = 5.4.3",
+			version:  "5.4.3",
+			version1: "5.4.3",
+			result:   true,
+		},
+		{
+			name:     "1.2.3 = 1.2.3-alpha1",
```
ruflin: Ok, I see why comment above would not work. I would not treat these as equal (at least in our stack as there are breaking changes between each alpha release.

libbeat/kibana/testdata/fields.yml
```diff
@@ -10,9 +10,20 @@
 
     - name: long
       type: long 
+      format: url
+      input_format: string
+      output_format: float 
+      output_precision: 5
+      label_template: "long template"
+      url_template:
+        - min_version: 5.0.0
```
ruflin: I like the specification of the min version here.

libbeat/kibana/transform.go
```diff
@@ -120,63 +125,74 @@ func transformField(f common.Field) (common.MapStr, common.MapStr) {
...
-	if (*f)["params"] == nil {
-		(*f)["params"] = common.MapStr{}
+	paramVer, _ := common.NewVersion("0.0.0")
+	var paramVal string
+	for _, v := range val {
+		minVer, err := common.NewVersion(v.MinVersion)
+		if err != nil {
+			continue
+		}
+		if minVer.Equal(version) {
```
ruflin: I can see now why you want equal here as we should not have in our fields.yml something like `beta1` specified. Perhaps we should name the method differently?

libbeat/kibana/transform.go
```diff
@@ -120,63 +125,74 @@ func transformField(f common.Field) (common.MapStr, common.MapStr) {
...
+func addVersionedFormatParam(f *common.MapStr, version *common.Version, key string, val []common.VersionizedString) {
+	if len(val) == 0 {
 		return
 	}
-	if (*f)["params"] == nil {
-		(*f)["params"] = common.MapStr{}
+	paramVer, _ := common.NewVersion("0.0.0")
+	var paramVal string
+	for _, v := range val {
+		minVer, err := common.NewVersion(v.MinVersion)
```
ruflin: This case would mean an invalid fields file? Should we at least log an error?

libbeat/kibana/transform_test.go
```diff
@@ -13,10 +13,11 @@ var (
 	truthy     = true
 	falsy      = false
 	ctMetaData = 4
+	version, _ = common.NewVersion("6.0.0")
 )
 
 func TestEmpty(t *testing.T) {
-	trans := NewTransformer("name", "title", common.Fields{})
+	trans, _ := NewTransformer("name", "title", version, common.Fields{})
```
ruflin: Should we check that err is nil in the tests?

libbeat/common/field.go
simitt: need to change my editor config for comments ;)

libbeat/common/version.go
simitt: I sticked to leaving out the meta information for `rc1` ,`alpha`, `beta` etc as this is also left out in the `LessThan` method. If treeting e.g. `6.0.0.alpha` to not be equal to `6.0.0` then this would lead to the assumption that `6.0.0.alpha` is **less than** `6.0.0` but this returns `false`. 

Will think about how we can change that without breaking existing behavior.

libbeat/kibana/transform.go
simitt: I have given it a second thought and I think it is even better to stop the processing here and return an error, since the kibana pattern can't be created as intended.

libbeat/common/version.go
```diff
@@ -76,3 +105,16 @@ func (v *Version) LessThan(v1 *Version) bool {
...
+	if v.Meta == "" && v1.Meta == "" {
+		return true
+	}
+	if v.Meta == "" {
+		return false
+	}
+	if v1.Meta == "" {
+		return true
+	}
+	return v.Meta <= v1.Meta
```
simitt: This ofc is not generically valid, but I think with our current version handling (`alpha, beta, rc1, ..`) it should be good enough.

ruflin: nice trick :-) I assume then <= will check the alphabetical order?

libbeat/kibana/transform.go
```diff
@@ -7,28 +7,42 @@ import (
...
 		transformedFieldFormatMap: common.MapStr{},
 		keys: common.MapStr{},
-	}
+	}, nil
 }
 
-func (t *Transformer) TransformFields() common.MapStr {
-	t.transformFields(t.fields, "")
+func (t *transformer) transformFields() (transformed common.MapStr, err error) {
+	defer func() {
```
ruflin: What is this needed?

libbeat/kibana/transform.go
```diff
@@ -120,63 +135,71 @@ func transformField(f common.Field) (common.MapStr, common.MapStr) {
...
 	}
-	if (*f)["params"] == nil {
-		(*f)["params"] = common.MapStr{}
+	paramVer, _ := common.NewVersion("0.0.0")
+	var paramVal string
+	for _, v := range val {
+		minVer, err := common.NewVersion(v.MinVersion)
+		if err != nil {
+			msg := fmt.Sprintf("ERROR: Parameter Version <%s> for <%s> is invalid. Please update and try again.", v.MinVersion, key)
+			panic(errors.New(msg))
```
ruflin: Why do you panic here instead of returning an error?

libbeat/common/version.go
simitt: yes it seems to use alphabetical order

simitt:
I used the `panic` and `recover` trick since I don't think that this is especially performance critical since it is only run with the `make` cmds.
Additionally I found the official [go wiki entry](https://github.com/golang/go/wiki/PanicAndRecover#usage-in-a-package) about panic and recovery stating:

> Within a package, however, especially if there are deeply nested calls to non-exported functions, it can be useful (and improve readability) to use panic to indicate error conditions which should be translated into error for the calling function. 

Now it panics as soon as something within this transformation goes wrong and then returns an error to the caller. It seems more convenient than having additional nil checks through all the nsted functioncalls.

ruflin:
I definitively see the benefit of readability in this case. Also as all the methods are private it's not expose external and could hit someone by surprise. 

I personally started to like (took me some time ;-) ) that errors are handled at all level and have to be dealt with. But I can hardly argue with the Golang docs :-)

Will move forward with the PR. WFG.

PR #5334, opened 2017-10-05
Cherry-pick #5319 to 6.0: Do not require template if index change and template disabled

ruflin:
Cherry-pick of PR #5319 to 6.0 branch. Original message: 

If the elasticsearch index name is changed, its required to set a template. The error message was also shown when template itself are disabled. This is now changed that if templates are disabled, no error is returned.

Closes https://github.com/elastic/beats/issues/5308


ruflin:
Failing tests should not be related.

PR #5335, opened 2017-10-05
Add support for copy_to to fields.yml

ruflin:
This allows to use the `copy_to` feature in the Elasticsearch index template: https://www.elastic.co/guide/en/elasticsearch/reference/current/copy-to.html Like this one field can be copied to an other name if needed.


ruflin:
@exekias I left it out as this should not have any user impact and is mainly for developers (for now).

PR #5339, opened 2017-10-05
Support stored scripts

kvch:
Previously, scripts of pipelines were inlined in JSON files. By moving scripts into separate files it makes them readable and testable. Now scripts are placed under `ingest/script` in every fileset which has scripts in its pipeline. Of course, it is still possible to use inline scripts in pipelines.

Now only painless scripts are supported.

TODO
- [x] fail gracefully if non-general purpose language is used


ruflin:
@kvch I think it's enough if we support painless for now. Is there a reason we would need others?

filebeat/fileset/modules.go
```diff
@@ -285,6 +294,40 @@ func (reg *ModuleRegistry) LoadPipelines(esClient PipelineLoader) error {
 	return nil
 }
 
+func scriptPayload(source, lang string) map[string]interface{} {
+	s := make(map[string]string)
+	s["lang"] = lang
+	s["source"] = source
+	payload := make(map[string]interface{})
```
ruflin: I suggest we use `common.MapStr` here mainly because we use it everywhere else where we have to create json. It has the same base interface.

kvch:
It is true that right now we only have painless scripts. But to be more future-proof and leave more room to play with FB modules it would be ideal to have all languages which is supported by ES.

I tried to come up with names that would not conflict with other scripts in users' ES. However, in the end I think it is the users' responsibility to make sure their script ids are not conflicting with FB's.

ruflin:
In 6.0 the only general purpose language left is painless: https://www.elastic.co/guide/en/elasticsearch/reference/6.0/modules-scripting.html

For the script naming if we go with approach 2 we should have at least a prefix with the beat name, something like `filebeat-is-private-ip`. I think it's our responsibility to heavily reduce the chance of name conflicts. I don't expect a user that load filebeat modules to know what the name of our scripts are and to check if they don't conflict with his.

kvch:
Great! How would you handle situations if a user wants to have a special-purpose language? Would you return an error or just ignore it and tell the user to rewrite it in painless?

Yes, you are right. First I named the files "filebeat-stored-is-private-ip", etc. But I seemed weird, so I deleted it. But I will reintroduce it.

ruflin:
@kvch What do you think about shipping the scripts with the processor instead of uploading them as separate script? Are there any downside of uploading it as part of the processor?

kvch:
@ruflin I understand that you would like to eliminate the possibility of collusion. However, if we add scripts to pipelines, we take away the possibility from users to view, edit or debug their scripts after feeding it into ES.

Furthermore, there is a Filebeat module for Kafka GC logs based on @urso's Kafka blogpost, but right now ES can't process its pipeline, because it includes too many scripts. But I rebased the branch on top of this one, and now it works smoothly.

I thinks prefixing names with "filebeat-stored-" should be enough to avoid collisions with users' stored scripts.

I am okay with documenting this behaviour. But I still need to implement graceful degradation in case users' would like to upload special-purpose languages.

ruflin:
One more question for the script loading: In case X-Pack is enabled, are the same permission required to load a pipeline like are needed to load a script?

For the user modification of scripts: In general I like that idea to offer this flexibility to the user. But do we expect the user to do this modifications in ES or on the Beats side in the script file? In case someone changes the script, I assume he would also have to modify other things from the module?

ruflin:
@kvch Can you elaborate on `ES can't process its pipeline, because it includes too many scripts.`? Means if a pipeline contains scripts that are too long directly, it does not work?

kvch:
@ruflin There is one privilege for pipeline handling called `manage_pipeline`. This includes all pipeline operations. As both inline and stored scripts are pipeline processors, it means that if someone has permission to load a pipeline, he/she can load scripts, too.

It would be nice if both scenarios were supported. But currently, if someone changes the script in ES, it should be changed in the FB module, too. It would be a nice addition to have a script/tool which exports scripts from ES and puts it under the right FB module. For now changing script in FB seems to be more comfortable.

The ingesting the pipeline of Kafka GC logs are not valid reason anymore, because the pipeline can be loaded using ES 6.0.0-rc1. The way script compilation is limited was changes since I last tested in August.

filebeat/fileset/modules.go
```diff
@@ -296,6 +296,11 @@ func (reg *ModuleRegistry) LoadPipelines(esClient PipelineLoader) error {
 
 func loadScript(name, source string, esClient PipelineLoader) error {
 	parts := strings.Split(name, ".")
+
+	if parts[1] != "painless" {
```
ruflin: you should probably check that `len(parts) > 1 `

tsg:
@kvch I'm thinking that in order to reduce confusion, it would be good for the scripts to follow the same practice as the pipelines, which is to encode the version in the pipeline, e.g.  `filebeat-7.0.0-alpha1-nginx-error-pipeline`. Otherwise, if you have two versions of Filebeat running, and the script was updated between them, you get the older version use the newer version of the script.

tsg:
I would consider explicitly listing the scripts in the `manifest.yml` file, instead of loading them all from the filesystem. This is what we do for the other Filebeat modules artifacts (e.g. ML jobs). I don't have a reason for which necessarily need this today, but I'd be more comfortable if we follow the same pattern for every such thing.

filebeat/fileset/fileset.go
```diff
@@ -350,13 +388,34 @@ func (fs *Fileset) GetPipeline(esVersion string) (pipelineID string, content map
...
 	if err != nil {
 		return "", nil, fmt.Errorf("Error JSON decoding the pipeline file: %s: %v", path, err)
 	}
+
 	return fs.pipelineID, content, nil
 }
 
+// formatScriptIDTemplate generates the ID to be used for the pipeline ID in Elasticsearch
+func formatScriptIDTemplate(module, fileset, beatVersion string) string {
+	return fmt.Sprintf("filebeat-%s-%s-%s-{{.}}", beatVersion, module, fileset)
```
ruflin: What is `{{.}}` doing?

filebeat/module/redis/log/manifest.yml
```diff
@@ -12,3 +12,7 @@ var:
 
 ingest_pipeline: ingest/pipeline.json
 prospector: config/log.yml
+
+pipeline_script:
```
ruflin: I wonder if we should rename this to `scripts` instead of `pipeline_script(s)` as in ES these are just scripts and not related to a pipeline as far as I know?

filebeat/fileset/fileset.go
ruflin: It seems this is replaced by the file / script name?

filebeat/module/redis/log/manifest.yml
kvch: AFAIK no. I added `pipeline_` prefix so it is unambiguous that is a pipeline setting, because manifest includes other types of settings, too. If it's not needed I am happy to remove the prefix.

ruflin: I think we can remove it. @tsg please object if not the case.

tsg: I agree we can do without the prefix.

filebeat/fileset/fileset.go
```diff
@@ -350,13 +388,49 @@ func (fs *Fileset) GetPipeline(esVersion string) (pipelineID string, content map
...
+	}
+
+	scriptElemFull := fmt.Sprintf(scriptPipelinePattern, scriptID.String())
+	jsonString = strings.Replace(jsonString, scriptElem, scriptElemFull, -1)
+
+	return jsonString, nil
+
+}
+
+// formatScriptIDTemplate generates the ID to be used for the pipeline ID in Elasticsearch
```
tsg: The comment should probably say "used for the pipline script IDs".

filebeat/fileset/fileset.go
```diff
@@ -318,6 +325,37 @@ func (fs *Fileset) getProspectorConfig() (*common.Config, error) {
...
+		if err != nil {
+			return nil, fmt.Errorf("Error while reading script %s for fileset %s: %v", name, fs.name, err)
+		}
+		scripts[name] = string(source[:])
+	}
+
+	return scripts, nil
+}
+
+// getScriptIDTemplate returns the Ingest Node pipeline ID
```
tsg: Comment seems wrong, I guess it is about the script IDs.

filebeat/fileset/fileset.go
```diff
@@ -350,13 +388,49 @@ func (fs *Fileset) GetPipeline(esVersion string) (pipelineID string, content map
...
+	scriptElem := fmt.Sprintf(scriptPipelinePattern, p[0])
+
+	scriptID := bytes.NewBufferString("")
+	err := t.Execute(scriptID, p[0])
+	if err != nil {
+		return "", err
+	}
+
+	scriptElemFull := fmt.Sprintf(scriptPipelinePattern, scriptID.String())
+	jsonString = strings.Replace(jsonString, scriptElem, scriptElemFull, -1)
```
tsg: Using string replacement for this seems brittle. Alternatively we could go through the JSON keys and look for the script tags. I think we do something like that for the ML job ID replacement. This gets more complicated if we look into anything but the top level processors, but I think that's all we need for the moment? What do you think?

ruflin:
@kvch I would really like to reactive this PR as I think it would make creating scripts much nicer. 

@jsoriano @sayden FYI

kvch:
o7

urso:
@kvch I think we can close it.

PR #5341, opened 2017-10-06
Add more context and examples to Filebeat module docs 

dedemorton:
This effort is to provide more contextual information within the topics about each module so that users who go from the solution pages directly to the modules doc have the info they need to work with the modules without jumping around a lot.

I plan to take all the shared content and pull it into shared files, but I wanted to make the content easier to read for the initial draft.

If this looks like the right approach, I'll apply it to the docs across the Filebeat modules.

I removed the tutorial because it's now redundant (we have the quick start and the config reference info that covers the content that was previously in the tutorial).


tbragin:
cc @brandonmensing

dedemorton:
@elastic/beats My updates to the modules doc are ready to review. A few outstanding items (some of which you can help with):

- [ ] update screenshots pf dashbpards
- [ ] update compatibility statements (def need your help here)
- [ ] update other modules that were added after RC1 

I wanted to get this out tonight in case you can look tomorrow, so my apologies for not putting up a build on firebase. I'm using a lot of includes because there is a lot of shared content. The best way to review is to pull down the files and build the doc.

dedemorton:
@brandonmensing ^^

brandonmensing:
I reviewed one module and the included reused sections, very nice! The what-happens section I think should state that we're using ingest node. Right now it sounds like Filebeat does the parsing which is maybe a little misleading. "Parses and processes the log lines..."

tsg:
One small nitpick that I found is that at the end of each module there's this phrase: "For a description of each field in the metricset, see the exported fields section.". The nitpick is that metricset is Metricbeat specific terminology, it should just say "module" instead, I think.

PR #5344, opened 2017-10-06
Scan files at startup for Auditbeat file integrity module

### LINKED ISSUE #5167: Add scanning to Auditbeat file integrity module
Currently the events generated by the file integrity module are triggered only on changes to the watched files. Adding a scan feature will enable the module to read through the watched files at startup and send events for each file thus establishing a baseline for these files that is stored in Elasticsearch.

In order to not hog CPU or IO resources at module startup the feature should have max limit on the average rate at which files are hashed.

The module should also be able to persist file state information to disk such that if the module restarts it will only send events for files that have changed.

andrewkroh:
At startup the file integrity module will iterate over directories and their contents and send events if any changes have occured since the last time Auditbeat was running. State information is persisted to {path.data}. Scanning at startup is enabled by default.

Closes #5167

For reference these are the benchmark results:
```
$ go test -v -bench . -benchmem -run non
BenchmarkHashFile/md5-8               10         175624594 ns/op           33464 B/op         12 allocs/op
BenchmarkHashFile/sha1-8              10         129382837 ns/op           33496 B/op         12 allocs/op
BenchmarkHashFile/sha224-8             5         294997162 ns/op           33512 B/op         12 allocs/op
BenchmarkHashFile/sha256-8             5         294042531 ns/op           33512 B/op         12 allocs/op
BenchmarkHashFile/sha384-8             5         323455129 ns/op           33624 B/op         12 allocs/op
BenchmarkHashFile/sha512-8             5         322871823 ns/op           33640 B/op         12 allocs/op
BenchmarkHashFile/sha512_224-8         5         322361720 ns/op           33608 B/op         12 allocs/op
BenchmarkHashFile/sha512_256-8         5         325741477 ns/op           33608 B/op         12 allocs/op
--- BENCH: BenchmarkHashFile
        event_test.go:206: file size: 104857600 bytes
BenchmarkFBEncodeEvent-8         2000000               946 ns/op               0 B/op          0 allocs/op
BenchmarkFBEventDecode-8         2000000               883 ns/op             708 B/op          8 allocs/op
BenchmarkJSONEventEncoding-8      200000              5899 ns/op            2176 B/op         26 allocs/op
BenchmarkJSONEventDecode-8        200000             11024 ns/op            1272 B/op         30 allocs/op
PASS
ok      github.com/elastic/beats/auditbeat/module/audit/file    32.903s
```


.gitignore
```diff
@@ -15,6 +15,7 @@
 *.generated.yml
 coverage.out
 .python-version
+beat.db
```
ruflin: I wonder if this file should not be in `data`?

NOTICE.txt
```diff
@@ -116,6 +116,34 @@ IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
 CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 --------------------------------------------------------------------
+Dependency: github.com/boltdb/bolt
```
ruflin: We had recently some discussions about alternatives for the FB registry file. As this is also kind of a registry, I wonder how this fits in. @urso

auditbeat/docs/fields.asciidoc
```diff
@@ -55,13 +55,21 @@ type: keyword
 The target path for symlinks.
 
 [float]
+=== `audit.file.source`
```
ruflin: When I read the name I assume it contains the filepath. When reading the text I see that source is used here differently. My wrong assumption mainly comes from filebeat as we use it there differently. I wonder if perhaps `audit.file.trigger` could be an alternative?

auditbeat/module/audit/file/scanner.go
```diff
@@ -0,0 +1,156 @@
...
+func (s *scanner) scan() {
+	if logp.IsDebug(metricsetName) {
+		debugf("%v File system scanner is starting for paths [%v].",
+			s.logPrefix, strings.Join(s.config.Paths, ", "))
+		defer debugf("%v File system scanner is stopping.", s.logPrefix)
+	}
+	defer close(s.eventC)
+	startTime := time.Now()
+
+	for _, path := range s.config.Paths {
```
ruflin: Would be create to share some code in the future with FB on this one.

auditbeat/docs/fields.asciidoc
andrewkroh: I removed the `source` field (bad name + it wasn't useful for anything).

.gitignore
andrewkroh: Auditbeat writes this file to `${path.data}`.

ruflin: But that would mean the ignore path should be `data/beat.db`?

andrewkroh: Having a gitignore of `beat.db` covers `data/beat.db`.

andrewkroh:
The travis-ci failure is not related to the changes (it's the libbeat -> Logstash integration test).

PR #5348, opened 2017-10-06

nilscant:
This document states that setting "document_type" is deprecated, yet it proposes to use it in the example at the top of the page. I've changed it so that it adds an "apache" tag instead.


ruflin:
Thanks for bringing this up. I suggest we completely remove the line and do not add `tags`.

Could you sign the CLA? http://www.elasticsearch.org/contributor-agreement/

nilscant:
Sure thing. I'll just make a new one.

Issue #5292, opened 2017-10-02
Document how to run Beats in Kubernetes


exekias:
Import existing manifests from https://github.com/elastic/beats-kubernetes and document how to deploy them to Kubernetes

- [x] Add kubernetes folder
- [x] Document how to deploy using the manifests

exekias:
We are finally using github URLs instead of packer, so this can be closed with https://github.com/elastic/beats/pull/5407

exekias:
#5407 was merged

PR #5349, opened 2017-10-06
Add kubernetes manifests

exekias:
Import https://github.com/elastic/beats-kubernetes, these manifests
deploy filebeat & metricbeat to Kubernetes

Also add documentation about them

Related to #5292


exekias:
jenkins package it please

Makefile
```diff
@@ -149,6 +155,13 @@ python-env:
 	@$(PYTHON_ENV)/bin/pip install -q --upgrade pip autopep8 six
 
 # Tests if apm works with the current code
-.PHONY: python-env
+.PHONY: test-apm
```
ruflin: thanks for the fix

deploy/kubernetes/.travis/setup.sh
```diff
@@ -0,0 +1,66 @@
...
+       hack/local-up-cluster.sh &
+popd
+
+# Wait until kube is up and running
+TIMEOUT=0
+TIMEOUT_COUNT=40
+until $( curl --output /dev/null --silent http://localhost:8080 ) || [ $TIMEOUT -eq $TIMEOUT_COUNT ]; do
+  echo "Kube is not up yet"
+  let TIMEOUT=TIMEOUT+1
+  sleep 20
```
ruflin: That would mean in total it could wait up to 800s? Seems like a bit too much. I would probably increase the TIMEOUT_COUNT to a larger number like 300 but just sleep for 1 second each time.

deploy/kubernetes/Makefile
```diff
@@ -0,0 +1,22 @@
+ALL = filebeat metricbeat
```
ruflin: Is the space after `ALL` on purpose?

deploy/kubernetes/.travis/setup.sh
exekias: It normally goes up to ~28 iteration of 40, that leaves around 3 extra minutes for the worst case, is not that much, I could make it more granular if you want, but total timeout time looks good to me as it is, wdyt?

deploy/kubernetes/Makefile
exekias: No special reason, can change it

exekias:
pinging @dedemorton on the documentation, let me know what do you think

deploy/kubernetes/.travis/setup.sh
ruflin: oh, didn't expect it to take several minutes to setup. Then the total is definitively good :-)

I would probably modify the checks as checking every second does not seem to be much overhead, it makes the tests quicker (in best case 19s ;-) ) and I think it's easier to read for a dev changing the code (personal preference).

PR #5350, opened 2017-10-06

nilscant:
Removing  document_type configuration option in prospector example, as the same page lists it as being deprecated.


ruflin:
jenkins, test it

ruflin:
@nilscant Thanks for the contribution.

PR #5340, opened 2017-10-05
Improvements and fixes to the Host Overview dashboard

tsg:
* Use `beat.name` instead of `beat.hostname` in visualizations. See #5276 for
  the motivation
* Add a "tip" widget that tells user how they can select another host.
* Automatically set a search pattern of the form `beat.name: "HOSTNAME"` where
  HOSTNAME is the Beat name (hostname) that uploads the dashboards.

The way the last point works is that I saved the dashbaord using this filter:
`beat.name:"CHANGEME_HOSTNAME"`. The kibana loader code does a string replacement
and replaces `CHANGEME_HOSTNAME` with the actual hostname. The disadvantage of this
approach (vs doing JSON parsing) is that we must remember to always save that
dashboard with the `CHANGEME_HOSTNAME` wildcard in place. I have added a unit test
to the system module that checks for that, so it should be relatively hard for
us to forget.

Closes #5276.

PR #5351, opened 2017-10-06
Cherry-pick #5340 to 6.0: Improvements and fixes to the Host Overview dashboard

### LINKED ISSUE #5276: 6.0 Metricbeat dashboards switched name key?
I am trying out the 6.0 metricbeat and kibana dashboards, in preparation for getting some screenshots for the new scheme, and I ran in to an issue.

In the 5.x versions of the metricbeat dashboards, the host stuff was keyed on `beat.name` (which defaults to the hostname, but can be overridden by setting the name in the config, https://www.elastic.co/guide/en/beats/metricbeat/5.6/configuration-general.html#_name)

With the 6.0 the visualizations are keyed on beat.*hostname* - which cannot be overridden.  Was this a choice that was made?  Or perhaps someone simply picked the wrong field for all of the visualizations?

I was told the change was intentional, but I am looking for a discussion before GA.

I think the `beat.name` makes a lot more sense - customers can either accept the hostname, or override the beat name and use that.  I see a lot of value in overriding - host names might be something generated, like `us-east-dc-1-rack-3`, but when they were keyed off of the overridable `name`, it could show up as `webserver1`.  they can always go back to look at the source data to find out the hostname webserver1 - because that is in the data.  

@monicasarbu @tsg @simianhacker

tsg:
Cherry-pick of PR #5340 to 6.0 branch. Original message: 

* Use `beat.name` instead of `beat.hostname` in visualizations. See #5276 for
  the motivation
* Add a "tip" widget that tells user how they can select another host.
* Automatically set a search pattern of the form `beat.name: "HOSTNAME"` where
  HOSTNAME is the Beat name (hostname) that uploads the dashboards.

The way the last point works is that I saved the dashbaord using this filter:
`beat.name:"CHANGEME_HOSTNAME"`. The kibana loader code does a string replacement
and replaces `CHANGEME_HOSTNAME` with the actual hostname. The disadvantage of this
approach (vs doing JSON parsing) is that we must remember to always save that
dashboard with the `CHANGEME_HOSTNAME` wildcard in place. I have added a unit test
to the system module that checks for that, so it should be relatively hard for
us to forget.

Closes #5276.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -45,6 +45,8 @@ https://github.com/elastic/beats/compare/v6.0.0-rc1...master[Check the HEAD diff
 
 *Metricbeat*
 
+- Use `beat.name` instead of `beat.hostname` in the Host Overview dashboard. {pull}5340[5340]
+
 *Packetbeat*
 
 *Winlogbeat*
@@ -60,6 +62,8 @@ https://github.com/elastic/beats/compare/v6.0.0-rc1...master[Check the HEAD diff
 *Heartbeat*
 
 *Metricbeat*
+
+- Auto-select a hostname (based on the host on which the Beat is running) in the Host Overview dashboard. {pull}5340[5340]
 
 *Packetbeat*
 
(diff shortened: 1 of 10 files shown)


CHANGELOG.asciidoc
```diff
@@ -61,6 +64,14 @@ https://github.com/elastic/beats/compare/v6.0.0-rc1...master[Check the HEAD diff
 
 *Metricbeat*
 
+- Add graphite protocol metricbeat module. {pull}4734[4734]
```
ruflin: Changelog mess?

PR #5352, opened 2017-10-06
Modify generator to get version to generate passed

ruflin:
Having the version passed from external makes the generator more reusable and cleans up the internal code.

* Modify internal dashboard version to common.Version

This is a prerequisit for #5328 and will make the change less complex.

### DIFF
--- a/dev-tools/cmd/kibana_index_pattern/kibana_index_pattern.go
+++ b/dev-tools/cmd/kibana_index_pattern/kibana_index_pattern.go
@@ -5,16 +5,16 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/kibana"
 	"github.com/elastic/beats/libbeat/version"
 )
 
 func main() {
-	beatVersion := version.GetDefaultVersion()
 	index := flag.String("index", "", "The name of the index pattern. (required)")
 	beatName := flag.String("beat-name", "", "The name of the beat. (required)")
 	beatDir := flag.String("beat-dir", "", "The local beat directory. (required)")
-	version := flag.String("version", beatVersion, "The beat version.")
+	beatVersion := flag.String("version", version.GetDefaultVersion(), "The beat version.")
 	flag.Parse()
 
 	if *index == "" {
@@ -32,18 +32,21 @@ func main() {
 		os.Exit(1)
 	}
 
-	indexPatternGenerator, err := kibana.NewGenerator(*index, *beatName, *beatDir, *version)
-	if err != nil {
-		fmt.Fprintf(os.Stderr, err.Error())
-		os.Exit(1)
-	}
+	version5, _ := common.NewVersion("5.0.0")
+	version6, _ := common.NewVersion("6.0.0")
+	versions := []*common.Version{version5, version6}
+	for _, version := range versions {
 
-	pattern, err := indexPatternGenerator.Generate()
-	if err != nil {
-		fmt.Fprintf(os.Stderr, err.Error())
-		os.Exit(1)
-	}
-	for _, p := range pattern {
-		fmt.Fprintf(os.Stdout, "-- The index pattern was created under %v\n", p)
+		indexPatternGenerator, err := kibana.NewGenerator(*index, *beatName, *beatDir, *beatVersion, *version)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, err.Error())
+			os.Exit(1)
+		}
+		pattern, err := indexPatternGenerator.Generate()
+		if err != nil {
+			fmt.Fprintf(os.Stderr, err.Error())
+			os.Exit(1)
+		}
+		fmt.Fprintf(os.Stdout, "-- The index pattern was created under %v\n", pattern)
 	}
 }
--- a/libbeat/dashboards/importer.go
+++ b/libbeat/dashboards/importer.go
@@ -11,6 +11,8 @@ import (
 	"path"
 	"path/filepath"
 	"strings"
+
+	"github.com/elastic/beats/libbeat/common"
 )
 
 // MessageOutputter is a function type for injecting status logging
@@ -19,7 +21,7 @@ type MessageOutputter func(msg string, a ...interface{})
 
 type Importer struct {
 	cfg     *Config
-	version string
+	version common.Version
 
 	loader Loader
 }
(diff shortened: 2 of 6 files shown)


ruflin:
@simitt I think we should merge https://github.com/elastic/beats/pull/5329 before this one as I want to add some more things to this PR which depends on methods you introduced in your PR.

libbeat/kibana/index_pattern_generator.go
```diff
@@ -29,33 +29,35 @@ func NewGenerator(indexName, beatName, beatDir, version string) (*IndexPatternGe
...
 
-	index5xPath, err := i.generate5x(commonFields)
-	if err != nil {
-		return nil, err
+	var path string
+
+	if i.version.Major == 5 {
+		path, err = i.generate5x(commonFields)
+	} else {
+		path, err = i.generate6x(commonFields)
```
simitt: Since the version is now passed in I think we should also rename `generate5x` and `generate6x` in something more generic, like 
```
if i.version.Major >= 6 {
  path, err = i.generateMinVersion6(commonFields)
} else if i.version.Major >= 5 {
  path, err = i.generateMinVersion5(commonFields)
} else {
  return errors.New("version unsupported..")
}
```

libbeat/kibana/index_pattern_generator.go
```diff
@@ -29,33 +29,35 @@ func NewGenerator(indexName, beatName, beatDir, version string) (*IndexPatternGe
...
 	}
 
-	index6xPath, err := i.generate6x(commonFields)
 	if err != nil {
-		return nil, err
+		return "", err
```
simitt: Since you already define the path veriable upfront you could `return path, err` without an explicit check here.

libbeat/kibana/index_pattern_generator.go
```diff
@@ -65,7 +67,7 @@ func (i *IndexPatternGenerator) generate5x(fields common.Fields) (string, error)
 		return "", err
 	}
 
-	file5x := filepath.Join(i.targetDir5x, i.targetFilename)
+	file5x := filepath.Join(i.targetDir, i.targetFilename)
```
simitt: We should remove line 64:
```
version, _ := common.NewVersion("5.0.0")
```

Rather use the given version, otherwise you don't really use the provided version from outside for fetching versionized information from the fields.

When generalizing this, the `generate5x` and `generate6x` methods have more overlap and might be a bit restructured.

dev-tools/cmd/kibana_index_pattern/kibana_index_pattern.go
```diff
@@ -32,18 +32,21 @@ func main() {
...
-
-	pattern, err := indexPatternGenerator.Generate()
-	if err != nil {
-		fmt.Fprintf(os.Stderr, err.Error())
-		os.Exit(1)
-	}
-	for _, p := range pattern {
-		fmt.Fprintf(os.Stdout, "-- The index pattern was created under %v\n", p)
+	version5, _ := common.NewVersion("5.0.0")
+	version6, _ := common.NewVersion("6.0.0")
```
simitt: should this be `6.0.0` already or rather be `6.0.0-rc1`?
As you mentioned in the former PR that we should take the `meta` information of versions into account, and there are breaking changes between pre-releases, the generator now treats `6.0.0.rc1 < 6.0.0`.

libbeat/kibana/index_pattern_generator.go
ruflin: changed

libbeat/kibana/index_pattern_generator.go
ruflin: good point, changed

libbeat/kibana/index_pattern_generator.go
ruflin: I definitively agree on this one and think the methods should probably be merge in the long run. But I suggest to do this in a follow up PR to not have too many changes in here.

dev-tools/cmd/kibana_index_pattern/kibana_index_pattern.go
ruflin: As we don't pass the version in I would stick with 6.0.0 as I don't think this is a field we should change over time with each release. Hopefully we can get rid of it completely in the long run as index pattern could be generated on request now that we have it in Golang so we would not have to pregenerate them.

ruflin:
@simitt Thanks for the review. There are quite a few things I want further clean up in this code but I prefer to do it in multiple smaller PR's which makes spotting problems easier. This PR was mainly an extraction of some code from https://github.com/elastic/beats/pull/5328 to make that PR smaller.

libbeat/kibana/index_pattern_generator.go
```diff
@@ -29,55 +29,53 @@ func NewGenerator(indexName, beatName, beatDir, version string) (*IndexPatternGe
...
-	}
+	var path string
 
-	index6xPath, err := i.generate6x(commonFields)
-	if err != nil {
-		return nil, err
+	if i.version.Major >= 6 {
+		path, err = i.genereteMinVersion6(commonFields)
+	} else {
+		path, err = i.genereteMinVersion5(commonFields)
```
simitt: There is a typo: `generate` instead of `generete`.

libbeat/kibana/index_pattern_generator.go
simitt: Agree on doing the method cleanup in a next version. 
Not passing in the version here but hardcoding it, I consider to be a bug though, e.g. one could pass in `7.0.0` which might have differences in the fields.yml to `6.0.0` but would still get the values for `6.0.0` with this handling.

ruflin: I changed it to `i.version`. I agree with 7.0 it would not be the expected behaviour but this is not a "bug" that was introduced here but already in the code.

simitt: The difference is that so far no version was given from the outside, but it was hardcoded inside the generator. Thus it was clear that it gets generated for those two defined versions.

This PR adds the possibility to pass in a version. The "bug" I was referring to would be to not use this version then in all places.

libbeat/dashboards/importer.go
```diff
@@ -245,7 +247,12 @@ func (imp Importer) downloadFile(url string, target string) (string, error) {
 func (imp Importer) ImportKibanaDir(dir string) error {
 	var err error
 
-	dir = path.Join(dir, imp.version)
+	versionPath := "default"
+	if imp.version.Major == 5 {
+		versionPath = "5x"
```
tsg: @ruflin For when you are back from vacation, I think this should be `5.x` instead of `5x`. I'm going to make it `5.x` in a PR, but let me know if it was `5x` on purpose.

ruflin: That was definitively a typo.

PR #5360, opened 2017-10-09
Cherry-pick #5353 to 5.6: Add support TLS renegotiation

### LINKED ISSUE #4386: Expose RenegotiateFreelyAsClient for SSL/TLS
By default, Apache will occasionally trigger a renegotiation for the SSL session. When Heartbeat gets the reset, there is an error in the logs: local error: tls: no renegotiation.

This feature in Go would allow Heartbeat to deal with this:
https://golang.org/pkg/crypto/tls/#RenegotiationSupport

andrewkroh:
Cherry-pick of PR #5353 to 5.6 branch. Original message: 

This PR adds support for enabling TLS renegotiation. The setting is `ssl.renegotiation` and the options are `never` (default), `once`, and `freely`. This exposes the three options from https://golang.org/pkg/crypto/tls/#RenegotiationSupport.

Fixes #4386

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -42,6 +42,8 @@ https://github.com/elastic/beats/compare/v5.6.2...5.6[Check the HEAD diff]
 
 *Affecting all Beats*
 
+- Add support for enabling TLS renegotiation. {issue}4386[4386]
+
 *Filebeat*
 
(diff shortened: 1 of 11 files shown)


andrewkroh:
The beats-ci failure is unrelated to the changes (it's the packetbeat 5x python tests on windows).

PR #5368, opened 2017-10-09
[WIP] Update docker to most recent version

ruflin:
This is updating the docker dependencies to v17.03.2-ce. This is a test to see if it would pass CI. Problem is that this adds a lot of files which could increase our binaries.


ruflin:
Closing for now, too many files added and breaks too many things.

PR #5369, opened 2017-10-09
Close changelog for 5.6.3

tsg:


tsg:
I checked with the LS team, the LS error is in the current snapshots but doesn't affect the release. So this can be merged.

Issue #4842, opened 2017-08-08
FileBeat elasticsearch output, unable to add "host" to http headers


zpoint:
configure file:

        headers:
            host: "datasearch"
            abc: "def"

I sniffer the network and find that "Abc" added to http header in every request, 
but "Host" remains default value (someting like "120.76.205.241:8000")

I need to use specific value for "Host", or my elasticsearch server will return 404.

Already asked in 
https://discuss.elastic.co/t/filebeat-elasticsearch-output-unable-to-add-host-to-http-headers/96073

PR #5373, opened 2017-10-10
Support defining a `host` key in the headers

ph:
Its common to be able to define the host header in an http request,
sadly golang's standard library will clear any `Host` defined key in the
header maps with the value of the `req.Host`, which will default to host
of the current machine.

This commits allow users to set the `Host` header key and the code will
pass this values to the `req.Host` method, this will make sure that on the
other hand of the connection the value of the header will be the
expected.

If not explicit value is set we will default to the standard library
behavior.

Fixes: #4842


ph:
I am looking at the failing tests, seems unrelated.

ruflin:
Change LGTM. Could you add a CHANGELOG entry?

ph:
@ruflin updated with changelog.

ph:
jenkins test this please

PR #5384, opened 2017-10-12
Update generation of data.json files

ruflin:
In a recent change the generation of the example data.json events was changed to use the internal event json encoder. The effect of this was that the meta data was also added to the event and indentation changed which cause large diffs. I prefer to leave out the metadata out of the example events as the data should be the same as it ends up in elasticsearch and is documented in the fields.yml.

As a first step, `data.json` for the system module were updated.


metricbeat/module/system/diskio/_meta/data.json
```diff
@@ -1,45 +1,30 @@
...
+                "count": 4045827,
+                "time": 695534
             },
             "write": {
-                "bytes": 2281472,
-                "count": 224,
-                "time": 29700
-            },
-            "iostat": {
-                "avg_queue_size": 0.020000,
```
andrewkroh: It would be nice to advertise these metrics in the example. Is there a config option we can set in `TestData` or do you want only the default fields in the example?

ruflin: I would definitively like to have these in too. I think the problem here is that I did run this on my Mac where these values are not available. Let me do a run for this data set in a linux container.

ruflin:
@andrewkroh Example from linux for diskio added. Can you have a look again?

PR #5393, opened 2017-10-13
Added KrakenBeat

PPACI:
https://github.com/PPACI/krakenbeat[krakenbeat]:: Collect information on each transaction on the Kraken crypto platform.


karmi:
Hi @PPACI, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/beats/pull/5393.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

PPACI:
@karmi : Done !

kvch:
Thank you for your contribution! Could you please put your beat after kafkabeat? The list is sorted alphabetically.

PPACI:
I'm sorry, i miss clicked when pasted my change

kvch:
thanks!

Issue #5301, opened 2017-10-03
Filebeat module for Elasticsearch 


tsg:
This is the meta ticket for implementing a Filebeat module for collecting and parsing the logs of Elasticsearch.

* [x] Bring up topic of more structured logging with Elasticsearch team https://github.com/elastic/elasticsearch/issues/31046
* [x] Server Fileset
  * [x] Add basic fileset: https://github.com/elastic/beats/pull/7124
  * [x] Add support for log lines from ML.
  * [x] Add enhanced structure: https://github.com/elastic/beats/pull/5394
* [x] Slowlog Fileset https://github.com/elastic/beats/pull/7473
  * [x] Add basic Fileset: Ingest data with timestamp extraction + test file
  * [x] Enhanced processing of log lines
* [x] Deprecation Fileset https://github.com/elastic/beats/pull/7474
  * [x] Add basic Fileset: Ingest data with timestamp extraction + test file
  * [x] Enhanced processing of log lines
* [x] GC log Fileset
  * [x] Add basic Fileset: Ingest data with timestamp extraction + test file https://github.com/elastic/beats/pull/7305
  * [x] Enhanced processing of log lines
* [x] GC log jvm9 Filset
  * [x] Add basic Fileset: Ingest data with timestamp extraction + test file https://github.com/elastic/beats/pull/7390
  * [x] Enhanced processing of log lines
* [x] Audit log fileset https://github.com/elastic/beats/pull/7365
  * [x] Add basic Fileset: Ingest data with timestamp extraction + test file
  * [x] Enhanced processing of log lines

joshuar:
I've submitted a PR for the beginnings of an Elasticsearch module in #5394 .

ruflin:
I updated this issue to track all filesets in more detail.

immon:
Doesn't `Cluster Log` cover the same as `Server Fileset`? I was about to work on it but either I don't understand a difference or this is the same.

joshuar:
@immon I believe so, see the discussion on https://github.com/elastic/beats/pull/7124.  I've removed the Cluster section.

ruflin:
++ I assume we can also merge gc for jvm and jvm9 as it's one fileset, just 2 grok patterns.

ruflin:
[...]

PR #5394, opened 2017-10-14
Initial commit of Elasticsearch Filebeat module.

joshuar:
This adds an initial rudimentary Filebeat module for Elasticsearch.  At the moment, it will parse the base Elasticsearch log but in the future will hopefully also handle other Elasticsearch log types like deprecation, slowlogs and audit log.


filebeat/docs/modules/elasticsearch.asciidoc
```diff
@@ -0,0 +1,37 @@
+////
+This file is generated! See scripts/docs_collector.py
+////
+
+[[filebeat-module-elasticsearch]]
+== Apache2 module
```
kvch: Could you please fix the title?

tsg:
@joshuar A high level question that I have: I see that Elasticsearch uses separate files for deperecation logs, slowlogs, audit, etc. Would it perhaps make sense to create indiviudual "filesets" for each of these? The advantage would be that the pipeline for each would be simpler, since it would have to deal with only one type of message. The disadvantage is that it wouldn't work *if* the log files would be merged together. I don't know if Elasticsearch allows that via configuration.

filebeat/module/elasticsearch/_meta/kibana/default/dashboard/elasticsearch-logs.json
```diff
@@ -0,0 +1,19 @@
+[
```
tsg: @joshuar These dashboard files don't seem to have been exported correctly. How did you do it? Best would be to do it like this: https://www.elastic.co/guide/en/beats/devguide/current/export-dashboards.html#_export_all_module_beat_dashboards

Or were they created on a 5.x instance?

jakommo:
I'm +1 on splitting up different log types. IMO you want to have a different dashboard for deprecation/audit logs anyway.

filebeat/module/elasticsearch/_meta/kibana/default/dashboard/elasticsearch-logs.json
tsg: I think I can recover them from these files if needed, but if you can export them by following that guide it's going to be easier.

tsg:
I captured some follow ups in #5301. We can merge this one and do the follow ups in new PRs, but I'd like to see if we can convert the dashboards in a better format first.

filebeat/module/elasticsearch/_meta/kibana/default/dashboard/elasticsearch-logs.json
joshuar: @tsg ah yep, lemme export them the proper way.  I'll update the PR next week.

joshuar:
I've regenerated the Kibana objects using the python script (made in 5.x).  Will these need to be updated for 6.x?  

Agree with different filesets for each of standard, slow, audit and deprecation logs.  That was my intention originally too, but I've only managed to work on the standard logs, so the remaining log types (and appropriate filesets) need to be added.  I don't think we have to worry about log merging,  the default configuration of Elasticsearch (which I assume we target with this module) has all these log types as separate files?

tsg:
Thanks @joshuar, I can take it from here and migrate the dashboards to 6.0.

tsg:
@joshuar Can you run `make update` in this branch, though. The `make check` test is failing and we can't merge without that.

filebeat/module/elasticsearch/log/_meta/fields.yml
```diff
@@ -0,0 +1,301 @@
...
+            - name: name
+              description: >
+                The cluster setting name.
+            - name: old_value
+              description: >
+                The previous value of the setting.
+            - name: new_value
+              description: >
+                The new value of the setting.
+    - name: node
```
ruflin: In a second step we should unify this with the data from metricbeat: https://github.com/elastic/beats/blob/master/metricbeat/module/elasticsearch/node/_meta/fields.yml#L1 So if someone searches for `elasticsearch.node.name` for example he gets information for logs and metrics.

joshuar: This is a good idea! I will work on adopting the established names in Metricbeat into this module.

ruflin:
@joshuar Thanks for pushing this forward. Can you ping us when you think we should have a closer look / review again?

joshuar:
@ruflin can do.  Are you guys considering this for any particular version?  Or just when its ready? :)  There is probably still quite a bit of work to do (add filesets for all log types, add tests, document).

ruflin:
@joshuar It's read when it's ready ;-) What I would suggest is to split it up in multiple parts and only do one log time at the time. Like this we get the "general ES logic" in with the first log type and follow up PR's will "only" have to add what is need for this log type. Like this we keep the PR's smaller, easier to review and get them in quicker. What we sometimes do is open a meta issue with all the tasks inside that need to be done. Like this even dashboards could be in a follow up PR or docs.

kvch:
jenkins test it

dilchenko:
Pardon my ignorance, but does this looks like an uphill loosing battle only to me? Spent past couple days writing Grok regexps to parse ES logs @ Swiftype, and there is absolutely no way this won't break with pretty much every Elasticsearch release. 
In order to reliably index elasticsearch logs w/ filebeat, Elasticsearch needs an ability to write exactly the same logs in a JSON format. These JSON logs *do not* need to replace the existing plaintext log, in fact that might be a bad idea.

ruflin:
I would like to get this PR again. Instead of supporting all options I would like to start very basic. For now I would ignore all logs except the `elasticsearch.log`.

What is the lowest common denominator across all lines? Here is an example line:

```
[2018-05-07T07:43:00,387][INFO ][o.e.c.m.MetaDataCreateIndexService] [vWNJsZ3] [metricbeat-7.0.0-alpha1-2018.05.07] creating index, cause [auto(bulk api)], templates [metricbeat-7.0.0-alpha1], shards [1]/[1], mappings [doc]
```

We start with the timestamp and then the log level? Are these two in all single and multiline events? Or do we also have the component everywhere? Everything which is not shared for now can go into `message` and we can start splitting it up later.

To make the PR simpler, I would also remove the dashboards from the PR for now. We can add these in a follow up PR.

@joshuar I'm happy to help on this PR to move it forward. Let me know how your preferred way would be to collaborate on this one. One thing we definitively need is some test log files. This also makes sure we discuss the exact same logs.

inqueue:
We have split this PR across a number of people in support to also include GC, deprecation, audit, and slow logs. We can go simple with all of these.

I found while working on GC logs that even though I had all of the sample logs parsed, getting the schema right is difficult. I was not happy with it and tossed it out to get back to basics. The schema needs careful consideration to ensure the fields are going to be useful for visualizations. 

Should we consider narrowing our scope to just the Elasticsearch cluster log?

ruflin:
I think the most important part is to have the module create with a first fileset. Each fileset should be it's own PR and we can then iterate on top of the existing filesets to extract more structure, but lets get them in simple.

For me the scope of this PR is only the basic Elasticsearch log not the other logs. The overall scope of the Elasticsearch module in Filebeat should be all logs. So I suggest everyone which works on a fileset should open a separate PR. @inqueue Does that make sense?

ruflin:
@inqueue I now opened https://github.com/elastic/beats/pull/7124 with the very basic fileset I had in mind. It hast a lot less functionality then we have in this PR. My intention with this PR is get a very basic version in Filebeat that allows us to ship raw logs, having the Elasticsearch module with the basic testing setup and infrastructure to make small iterative changes on top of it.

For the log fileset what I'm hoping is that now several smaller contributions follow that add grok patterns for the most important features and also add test logs for these cases. @joshuar @inqueue Let me know if you want to sync up over zoom here.

joshuar:
Hey @ruflin thanks for the ping here.  Apologies, I missed this in my email (GitHub notifications suck).  

As it stands with this PR, there is a basic parsing of the standard log file, with just a few additional lines parsed which are useful.  There is also dashboards available.  The only thing missing is a test log file.  The conflicts here can probably be easily resolved.

How do we consolidate this work with #7124?  Should we somehow merge these?

ruflin:
@joshuar I would like to get #7124 in quickly as it allows to create PR's for the other fileset without closing conflicts. This PR should be rebased on top of it as soon as it's merged. I hoping we can move forward in small steps to get the most relevant parts from the ES logs structured.

joshuar:
@ruflin okay sounds like a plan, I'll probably need to pick your brain on how to rebase in this way.  I added some comments to #7124

ruflin:
@joshuar Yes, there will be quite a few conflicts afterwards. But I hope this will remove lots of conflicts in the future. Ping me on zoom if I can help.

joshuar:
With #7124, this PR is largely obsolete.  Any missing functionality in the latter which is found in this PR can be put into a smaller, newer PR against master.  Closing this one off.

PR #5395, opened 2017-10-14
Fix tags type to []interface{} in event (#5389)

### LINKED ISSUE #5389: Filebeat json input with tags not supported
- Version: Any filebeat version supporting JSON input
- Steps to Reproduce: 
  - json input with events having a tags field as array of string
  - add additional tags in filebeat configuration to the events
  - events from configuration file are not applied

Problem is the json parser converts the strings on input to `[]interface{}`, but when adding any new tags, a type check testing for `[]string` is in place only.

liketic:
Closes #5389 @urso Could you please review this? 😄

### DIFF
--- a/filebeat/tests/system/test_json.py
+++ b/filebeat/tests/system/test_json.py
@@ -130,6 +130,29 @@ class Test(BaseTest):
         assert output["source"] == "hello"
         assert output["message"] == "test source"
 
+    def test_json_add_tags(self):
+        self.render_config_template(
+            path=os.path.abspath(self.working_dir) + "/log/*",
+            json=dict(
+                keys_under_root=True,
+            ),
+            agent_tags=["tag3", "tag4"]
+        )
+        os.mkdir(self.working_dir + "/log/")
+        self.copy_files(["logs/json_tag.log"],
+                        source_dir="../files",
+                        target_dir="log")
+
+        proc = self.start_beat()
+        self.wait_until(
+            lambda: self.output_has(lines=1),
+            max_timeout=10)
+
+        proc.check_kill_and_wait()
+
+        output = self.read_output()[0]
+        assert sorted(output["tags"]) == ["tag1", "tag2", "tag3", "tag4"]
+
     def test_config_no_msg_key_filtering(self):
         """
         Should raise an error if line filtering and JSON are defined,
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -46,6 +46,7 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
 
 - Fix default paths for redis 4.0.1 logs on macOS {pull}5173[5173]
 - Fix Filebeat not starting if command line and modules configs are used together. {issue}5376[5376]
+- Add support for adding string tags {pull}5395{5395}
 
 *Heartbeat*
 
--- a/filebeat/tests/files/logs/json_tag.log
+++ b/filebeat/tests/files/logs/json_tag.log
@@ -1,0 +1,1 @@
+{"http_user_agent": "ELB-HealthChecker/1.0", "tags": ["tag1", "tag2"]}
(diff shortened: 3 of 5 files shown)


liketic:
Maybe we should support both `[]string` and `[]interface{}` like:
`
	if tagsIfc, ok := ms[TagsKey]; ok {
		switch tagsIfc.(type) {
		case []interface{}:
			existingTags := tagsIfc.([]interface{})
			for _, tag := range tags {
				existingTags = append(existingTags, tag)
			}
			ms[TagsKey] = existingTags
		case []string:
			existingTags := tagsIfc.([]string)
			existingTags = append(existingTags, tags...)
			ms[TagsKey] = existingTags
		default:
			return errors.Errorf("expected string array by type is %T", tagsIfc)
		}
	} else {
		ms[TagsKey] = tags
	}
`

ruflin:
@liketic Thanks for the fix. This could also use a CHANGELOG entry. The code review I let to @urso .

urso:
The type-switch use-case from your comment is definitely better. With some cleanup it becomes:

```
  T, exists := ms[TagsKey]
  if !exists {
    ms[TagsKey] = tags
    return nil
  }

  switch arr := T.(type) {
    case []string:
      ms[TagsKey] = append(arr, tags...)
    case []interface:
      for _, tag := range tags {
        arr = append(arr, tag)
      }
      ms[TagsKey] = arr
    default:
      return errors....
  }

  return nil
```

In addition one can change [this event normalization/parsing code](https://github.com/elastic/beats/blob/master/libbeat/common/jsontransform/jsonhelper.go#L23) to also handle the `tags` keyword. But once AddTags supports `[]string` and `[]interface{}` it should be fine.

Please add a [Changlog entry](https://github.com/elastic/beats/blob/master/CHANGELOG.asciidoc) when you are done. 

As this fixes an issue with filebeat, having a filebeat system test would be nice to have. Check for [tests/system/test_json.py](https://github.com/elastic/beats/blob/master/filebeat/tests/system/test_json.py).

liketic:
@urso I updated the pr, please review again. Thanks!

liketic:
Sorry I didn't notice your comments, will update again.

liketic:
Hi @urso , I didn't find a way to run the system test `system/test_json.py`, could you please provide some help? 😃

urso:
system tests are run via nosetest. You can setup the python environment from within the filebeat directory via 

```
make python-env
source build/python-env/bin/active
```

head over to `tests/system` and run the tests via `nosetest test_json.py`.

filebeat/tests/system/test_json.py
```diff
@@ -367,3 +390,4 @@ def test_integer_condition(self):
         output = self.read_output()
         assert len(output) == 1
         assert output[0]["status"] == 404
+
```
urso: Not sure, but the linter might complain about a newline at the end of the file.

filebeat/tests/system/test_json.py
```diff
@@ -130,6 +130,29 @@ def test_simple_json_overwrite(self):
...
+        proc = self.start_beat()
+        self.wait_until(
+            lambda: self.output_has(lines=1),
+            max_timeout=10)
+
+        proc.check_kill_and_wait()
+
+        output = self.read_output()[0]
+        assert sorted(output["tags"]) == ["tag1", "tag2", "tag3", "tag4"]
```
urso: 👍

urso:
PR looks good to me. Thanks for taking the time!!! Also the filebeat system test are much appreciated!

Unfortunately there is a conflict in the changelog file (github sometimes has problem with the changelog). Please rebase and check git not introducing any more errors (e.g. wrong changelog entries). This should also restart the tests, which have not been executed at all :(

liketic:
@urso I have to say I didn't get it passed on my mac... Because I always got a error `file not found` seems like the root cause is the file `filebeat.test` not generated. It's too hard to do the test without a guide.  And also, seems like the permission of a lot of files is root user access only. Is that expected behaviour? Is the testing failure caused by my changes? Thanks very much!

urso:
I see. you can build `filebeat.test` via `make filebeat.test`.

No idea why files are owned by root user. Did you run via docker? Try `sudo make clean`.

urso:
I checked CI and your tests did pass. It's some flaky kafka tests in metricbeat that have failed.

Will merge PR. Thanks for taking to time fix the issue.

PR #5402, opened 2017-10-18
Add experimental Docker json-file prospector

exekias:
This pull request adds an experimental `docker` prospector. Docker supports different ways of outputting logs, this prospector implements [json-file](https://docs.docker.com/engine/admin/logging/json-file/).

By offering the prospector we abstract the format from the user, so there is no need to manually configure JSON decoding, also we can get timestamp from the JSON file, something that's not possible as of now from existing settings.

```
prospectors:
   - type: docker
     containers.ids:
       - c3ec7a0bd9640151a768663b7e78c115d5b1a7f87fba572666bacd8065893d41
```

This also abstract specifics of the chosen provider, for this case this would create a file prospector for `/var/lib/docker/containers/c3ec7a0bd9640151a768663b7e78c115d5b1a7f87fba572666bacd8065893d41/*.log`.

In the future, we can extend this to support other logging drivers/formats, like CRI format, or a plugin implementation. I would add a conf key like `mode: json-file` once that happens.

TODO:
- [x] Abstract file paths by adding a `container_id` parameter.
- [x] Extract @timestamp from JSON
- [x] Add tests


filebeat/harvester/util.go
```diff
@@ -4,18 +4,20 @@ import "github.com/elastic/beats/libbeat/common/match"
...
+	DockerType = "docker"
 )
 
 // ValidType of valid input types
 var ValidType = map[string]struct{}{
-	StdinType: {},
-	LogType:   {},
-	RedisType: {},
-	UdpType:   {},
+	StdinType:  {},
```
ruflin: @kvch Not related to this PR, but as we register the prospectors now I'm curious if we still need this?

kvch: `ValidType` is still used in `log` prospector for config validation and opening the right input.
But there is no other check/use of it. The check now happens when a factory is retrieved from registry using `GetFactory`. So it is not required, but this part needs further refactoring to make `ValidType` `log` specific. For the sake of consistency and to make refactoring easier, it is acceptable to add docker to the list.

ruflin:
Lets move this forward to play around with it but mark it experimental for now. Can you add an experimental log message and add it to the CHANGELOG?

kwojcicki:
I would like to have filebeat only harvest from currently active docker containers. Was wondering if this would be an appropriate thing to add to this docker prospector or maybe it should be added somewhere else?

exekias:
This should be ready for a second look @ruflin, I implemented timestamp parsing, put message in the right place and added some tests

exekias:
jenkins, retest this please

filebeat/harvester/reader/docker_json.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	// Parse timestamp
+	ts, err := time.Parse(time.RFC3339, line.Timestamp)
+	if err != nil {
+		return message, errors.Wrap(err, "parsing docker timestamp")
+	}
+
+	message.AddFields(common.MapStr{
+		"stream": line.Stream,
```
ruflin: Could you give some details on what values `stream` will have? Should we add this to a `fields.yml`?

filebeat/harvester/reader/docker_json.go
```diff
@@ -0,0 +1,54 @@
...
+	var line dockerLog
+	dec := json.NewDecoder(bytes.NewReader(message.Content))
+	if err = dec.Decode(&line); err != nil {
+		return message, errors.Wrap(err, "decoding docker JSON")
+	}
+
+	// Parse timestamp
+	ts, err := time.Parse(time.RFC3339, line.Timestamp)
+	if err != nil {
+		return message, errors.Wrap(err, "parsing docker timestamp")
```
ruflin: Perhaps worth adding the given timestamp to the error. Probably already part of `err`?

filebeat/harvester/reader/docker_json.go
ruflin: I see below this is `stdin` for example. If we know this field will exist, probably worth adding as keyword to fields.yml

filebeat/harvester/util.go
ruflin: I pushed https://github.com/elastic/beats/pull/5537 to remove this part.

filebeat/prospector/docker/config.go
```diff
@@ -0,0 +1,11 @@
+package docker
+
+var defaultConfig = config{
+	ContainersIDs:  []string{},
+	ContainersPath: "/var/lib/docker/containers",
+}
+
+type config struct {
+	ContainersIDs  []string `config:"containers_ids"`
```
ruflin: Could we make this `container.ids` and `container.path` or `containers.ids` and `containers.path`?

filebeat/prospector/log/config.go
```diff
@@ -83,6 +83,9 @@ type config struct {
 	MaxBytes     int                     `config:"max_bytes" validate:"min=0,nonzero"`
 	Multiline    *reader.MultilineConfig `config:"multiline"`
 	JSON         *reader.JSONConfig      `config:"json"`
+
+	// Hidden on purpose, used by the docker prospector:
+	DockerJSON bool `config:"docker-json"`
```
ruflin: Kind of hacky but don't have a better solution.

filebeat/prospector/log/harvester.go
```diff
@@ -515,6 +517,11 @@ func (h *Harvester) newLogFileReader() (reader.Reader, error) {
 		return nil, err
 	}
 
+	if h.config.DockerJSON {
```
ruflin: It's interesting that form an implementation point of view it just comes down to adding a reader in the harvester (simplified). Seems like the log harvester is pretty generic and perhaps should be extracted. Prospectors then could configure "readers" on the harvester. (just thinking out loud for later and @urso was also talking about similar stuff in the past).

filebeat/prospector/log/config.go
exekias: @urso recommended something like this to avoid repeating code or a mayor refactor, works for me :innocent:

filebeat/harvester/reader/docker_json.go
exekias: parse error looks pretty verbose: https://golang.org/src/time/format.go?s=23626:23672#L669

filebeat/harvester/reader/docker_json.go
exekias: Just pushed it, thanks!

exekias:
rebased after master changes

filebeat/_meta/fields.common.yml
```diff
@@ -23,6 +23,12 @@
       description: >
         The content of the line read from the log file.
 
+    - name: stream
```
ruflin: Could we call this `log.stream` or something similar? `log.source`? Source is already overloaded.

@weltenwort has some ideas for a better name?

ruflin:
Merged this one as the field naming discussion we can still do in a follow up PR.

tsg:
@exekias Please remember to add docs for this one before the 6.1 GA date.

dedemorton:
Removing needs_docs label because it looks like docs were added in https://github.com/elastic/beats/pull/5752

PR #5408, opened 2017-10-18
New protocol support: SMTP

### LINKED ISSUE #467: New protocol support: SMTP
Add SMTP protocol support.
### HELO/Envelope
- HELO/EHELO
- MAIL FROM
- RCPT TO
### DATA
- From
- To
- CC
- Date
- Subject
- Headers
- Message content
### Responses & Other commands
- 2xx/3xx/4xx/5xx response codes
- RSET, VRFY, NOOP, QUIT, etc

geokat:
This commit closes #467

* No explicit handling of AUTH LOGIN (handles related requests as
  ad-hoc commands)

* No STARTTLS detection


packetbeat/_meta/beat.reference.yml
```diff
@@ -316,6 +316,28 @@ packetbeat.protocols:
...
+  ports: [25]
+
+  # If this option is enabled, the SMTP requests are included in published
+  # events. The default is true.
+  #send_request: true
+
+  # If this option is enabled, the SMTP DATA request headers are included
+  # in published events. The default is false. Enable `send_request`
+  # before enabling this option.
+  #send_data_headers: true
```
andrewkroh: The commented out values should match the default for consistency reasons.

packetbeat/protos/smtp/smtp_test.go
```diff
@@ -0,0 +1,196 @@
+// +build !integration
+
+package smtp
+
+import (
+	"testing"
+
+	"github.com/elastic/beats/libbeat/common"
+	"github.com/stretchr/testify/assert"
```
andrewkroh: Please run `make fmt` to automatically reorder the imports using `goimports` (we group by stdlib, third-party, then elastic).

geokat:
@andrewkroh Thanks, I addressed your comments and added a couple of other changes.

andrewkroh:
jenkins, test it

geokat:
Windows test error for the system tests with gaps in the pcap files:
> CRIT Exiting: Sniffer main loop failed: Error starting sniffer: bad dump file format

Those test pcap files were produced by deleting a packet from a complete pcap using `editcap`. Apparently the windows version of packetbeat does not like this. I'll look into it.

geokat:
jenkins, test it

andrewkroh:
jenkins, test it

I think it only trust's people in the Elastic GH org.

packetbeat/_meta/beat.reference.yml
```diff
@@ -316,6 +316,26 @@ packetbeat.protocols:
...
+  # body is included in published events. The default is false.
+  #send_data_body: false
+
+  # If this option is enabled, the raw message of the request (`request` field)
+  # is sent to Elasticsearch. The default is false.
+  #send_request: false
+
+  # If this option is enabled, the raw message of the response (`response`
+  # field) is sent to Elasticsearch. The default is false.
+  #send_response: false
```
urso: From description it's somewhat unclear to me how these settings are related to each other. If `send_request` is enabled, does this mean `send_data_headers` and `send_data_body` are automatically enabled as well?

packetbeat/protos/smtp/smtp_test.go
```diff
@@ -0,0 +1,205 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/elastic/beats/libbeat/common"
+)
+
+func TestSmtpParser_startResponse(t *testing.T) {
```
urso: please use camel case for test names only. Without underscore. I know other modules may still use `_` in names, but this is some minor technical debt not being cleaned up yet.

e.g `TestSmtpParser_startResponse -> TestParserStartResponse` 

If possible consider more tabular tests for testing some common functionality (only use tabular testing if it's more convenient and reduces common boilerplate). E.g.:

```
func TestParser(t *testing.T) {
  var tests := []struct{
    name string
    pub *transPub
    inputs []string
    validate func(*testing.T, pub *transPub, *message)
  } {
  }
 
  for _, test := range tests {
    test := test // so name is unique within scope
    t.Run(test.name, func(t *testing.T) {
      ...
    })
  }
}
```

The `validate` is a function + you can compose multiple function (using closures) to build more enhance validators. If you want to validate per input consider replacing `inputs` and `validate` with a struct like

```
type testInput struct {
  content string
  validate func(...)
}
```

packetbeat/protos/smtp/pub.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+		dst := &common.Endpoint{
+			IP:   requ.Tuple.DstIP.String(),
+			Port: requ.Tuple.DstPort,
+			Proc: string(requ.CmdlineTuple.Dst),
+		}
+
+		fields["responsetime"] = responseTime
+		fields["src"] = src
+		fields["dst"] = dst
```
urso: I think we can still tell src/dst from the response message in this case.

packetbeat/protos/smtp/pub.go
```diff
@@ -0,0 +1,112 @@
...
+
+		if pub.sendRequest {
+			fields["request"] = common.NetString(requ.raw)
+		}
+
+		requNotes = requ.Notes
+		ts = requ.Ts
+
+		details["request"] = common.MapStr{}
+		dr := details["request"].(common.MapStr)
```
urso: no need to case, as MapStr is shared by reference:

```
dr := common.MapStr{}
details["request"] = dr
```

packetbeat/protos/smtp/pub.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+	}
+
+	// add processing notes/errors to event
+	if len(requNotes)+len(resp.Notes) > 0 {
+		fields["notes"] = append(requNotes, resp.Notes...)
+	}
+
+	details["response"] = common.MapStr{}
+	dr := details["response"].(common.MapStr)
```
urso: same:

```
dr := common.MapStr{}
details["response"] = dr
```

packetbeat/protos/smtp/smtp.go
```diff
@@ -0,0 +1,212 @@
...
+	defer logp.Recover("Parse SMTP exception")
+
+	conn := smtp.ensureConnection(private)
+
+	// There's no way to recover synchronization from packet loss
+	// during SMTP session. However packetbeat doesn't let us simply
+	// drop the session. The best we can do is suppress any message
+	// generation for the rest of the session.
+	if conn.isPacketLoss {
+		return conn
```
urso: Problem exists even on startup, as packetbeat might be started in the middle of an SMTP session. The parser should be hardened enough to return errors + ensure state is dropped if it expects to parse an SMTP header. At some point packetbeat needs to sync with an already active connection (by luck). This is true for the other protocols as well. We just try to parse and drop state, until we're finally/eventually in sync with an already existing connection.

I see how SMTP can be somewhat ambigious when not being in sync. One option trying to sync is looking for specific message header/type at beginning of line or look out for final `<CRLF>.<CRLF>` sequence to start syncing after current message has been transferred in case a session is used to send multiple mails.

packetbeat/protos/smtp/trans.go
```diff
@@ -0,0 +1,128 @@
...
+	requests := &trans.requests
+	responses := &trans.responses
+
+	// drop responses with missing requests, unless they are "prompt"
+	// responses (220, 421, etc.)
+	if requests.empty() {
+		for !responses.empty() {
+			resp := responses.pop()
+			switch resp.statusCode {
+			case 220, 421:
```
urso: what about 554? Any more codes that are commonly used when prompting?

packetbeat/protos/smtp/parser.go
```diff
@@ -0,0 +1,261 @@
...
+		var code int
+
+		byts, err := p.buf.CollectUntil(constCRLF)
+		if err != nil {
+			// Get more data
+			return nil, nil
+		}
+
+		nbytes := len(byts)
+		str := string(byts[:nbytes-constCRLFSize])
```
urso: this allocates a new buffer and copies all contents. Consider using the `bytes` package or use stringbuf.Buffer capabilities for parsing + advancing pointers. e.g. `(*Buffer). UintASCII` to parse the status code.

packetbeat/protos/smtp/parser.go
```diff
@@ -0,0 +1,261 @@
...
+}
+
+func (p *parser) parse() (*message, error) {
+	m := p.message
+
+	for {
+		var err error
+		var code int
+
+		byts, err := p.buf.CollectUntil(constCRLF)
```
urso: minor one, but `byts` is somewhat weird. How about one of `buf`, `raw`, `data`, `content`?

packetbeat/protos/smtp/parser.go
```diff
@@ -0,0 +1,261 @@
...
+	param   common.NetString
+	// Request data payload
+	headers map[string]common.NetString
+	body    common.NetString
+
+	// Response
+	statusCode    int
+	statusPhrases []common.NetString
+
+	raw []byte
```
urso: consider `streambuf.Buffer` for the raw data. Supports append and implements reader interface, tracking amount of bytes processed.

packetbeat/_meta/beat.reference.yml
geokat: > If send_request is enabled, does this mean send_data_headers and send_data_body are automatically enabled as well?

No, the 2 pairs are not related. As with HTTP, the user enables `send_request` and `send_response` if they want the raw data included under the event's top-level `request` and `response` fields.

`send_data_headers` and `send_data_body`, on the other hand, add relevant fields to `smtp -> request` for `EOD`-type requests.

I'll try to make it clear in the description.

packetbeat/protos/smtp/smtp.go
geokat: Great idea, `<CRLF>.<CRLF>` can only happen inside of a `DATA` request, so it's a good way to re-sync the parser.

packetbeat/protos/smtp/trans.go
geokat: I'll check again to make sure, but 554 can only happen as a response to `DATA`.
I'm not aware of any other prompt response codes.

geokat:
@urso
> Thusly all commands between (including) mail and data will be reported as one single transaction.

Should the individual requests/responses still be included in such a transaction? Something like this (an array of objects):
```json
{
  "smtp": {
    "session_id": "56b7209a-3bad-4d68-85ea-e948e818b645",
    "mail": [
      {
        "request": {
          "raw": "...",
          "command": "MAIL",
          "param": "FROM:bar@example.org"
        },
        "response": {
          "code": 250,
          "phrases": [
            "OK"
          ]
        }
      },
      {
        "request": {
          "raw": "...",
          "command": "RCPT",
          "param": "TO:foo@example.org"
        },
        "response": {
          "raw": "...",
          "code": 250,
          "phrases": [
            "Accepted"
          ]
        }
      },
      {
        "request": {
          "raw": "...",
          "command": "RCPT",
          "param": "TO:baz@example.org"
        },
        "response": {
          "raw": "...",
          "code": 250,
          "phrases": [
            "Accepted"
          ]
        }
      },
      {
        "request": {
          "raw": "...",
          "command": "DATA"
        },
        "response": {
          "raw": "...",
          "code": 354,
          "phrases": [
            "Enter message, ending with \".\" on a line by itself"
          ]
        }
      },
      {
        "request": {
          "raw": "...",
          "command": "EOD",
          "headers": {
            "From": "bar@example.org",
            "Subject": "Test",
            "To": "foo@example.com",
            "Date": "Thu, 20 Dec 2012 12:00:00 +0000"
          },
          "body": "Testing\r\n"
        },
        "response": {
          "raw": "...",
          "code": 250,
          "phrases": [
            "OK id=1e6kRR-0002cD-Mw"
          ]
        }
      }
    ]
  }
}
```
I guess other commands (other than `MAIL`/`RCPT`/`DATA`) can then be included under a `command` header:
```
{
  "smtp": {
    "session_id": "56b7209a-3bad-4d68-85ea-e948e818b645",
    "command": {
      "request": {
        "raw": "...",
        "command": "EHLO",
        "param": "localhost"
      },
      "response": {
        "raw": "...",
        "phrases": [
          "localhost Hello localhost [::1]",
          "SIZE 52428800",
          "8BITMIME",
          "PIPELINING",
          "PRDR",
          "HELP"
        ],
        "code": 250
      }
    }
  }
}
```

urso:
As long as the intermedia messages are 'okayed' (250 status code), I would not include them in the transaction. For intermediate errors we can put them into the events 'Notes' field. It's a while I've been working with SMTP, but I wonder if any of these error codes can actually abort the initial Mail transaction.

Plus, objects in arrays do ont play well with kibana. That's why we try to not use complex array, but only arrays of strings if possible.

How about storing the Mail transaction commands as  special fields like `receipient` being an array of strings.

For commands I like your proposal. As command can be transactions a field naming the transaction/type would be helpful:

```
{
  "smtp": {
    "transaction": "MAIL", // or name field 'type' or 'command'?
    "request": { ... }
    "response": { ... }
  }
}
```

urso:
@geokat Thanks for all the effort you are putting into this module!

I think you need to rebase to master, to fix the merge conflict. Otherwise the tests are not run.

Ping me when you need another review.

geokat:
@urso Thanks for the heads up, will do! I'll rebase once all the changes are in.

ruflin:
@geokat Did you have any chance to tackle the changes?

eloyekunle:
I'll like to continue from where @geokat stopped.

eBulldog:
+1 we'd see this as a major reason to use beats over anything else if we can get SMTP transaction monitoring in there.

PR #5416, opened 2017-10-21
Fix VSphere config yml

### LINKED ISSUE #5403: metricbeat - vsphere module config example is wrong
The default file for the vsphere module (modules.d/vsphere.yml.disabled) contains an invalid configuration example:
`metricsets: ["datastore, host, virtualmachine"]`

but this will cause metricbeat to report an invalid metricsets error at startup.
The correct configuration is:

`metricsets: ["datastore", "host", "virtualmachine"]`

Please also note the same error can be found in the official docs at https://www.elastic.co/guide/en/beats/metricbeat/6.0/metricbeat-module-vsphere.html

Tested with metricbeat 6.0.0-rc1 on Windows Server 2008R2.

sitaramshelke:
The metricsets should be array of strings. Fixes #5403

### DIFF
--- a/metricbeat/metricbeat.reference.yml
+++ b/metricbeat/metricbeat.reference.yml
@@ -425,7 +425,7 @@ metricbeat.modules:
 
 #------------------------------- vSphere Module ------------------------------
 - module: vsphere
-  metricsets: ["datastore", host", "virtualmachine"]
+  metricsets: ["datastore", "host", "virtualmachine"]
   period: 10s
   hosts: ["https://localhost/sdk"]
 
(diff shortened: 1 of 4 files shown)


exekias:
Thanks for taking this one! could you please run `make update` to make sure everything gets updated? That includes docs, reference files and so

metricbeat/metricbeat.reference.yml
```diff
@@ -425,7 +425,7 @@ metricbeat.modules:
 
 #------------------------------- vSphere Module ------------------------------
 - module: vsphere
-  metricsets: ["datastore, host, virtualmachine"]
+  metricsets: ["datastore", host", "virtualmachine"]
```
andrewkroh: Looks like it's missing a quote at the start of `host`.

andrewkroh:
jenkins, test it

PR #5426, opened 2017-10-24
Fix data race in mockbeat (#5037)

### LINKED ISSUE #5037: Data race in mockbeat unit test
A test failed (`test_base.Test.test_invalid_config_cli_param`) in the 6.0 branch due to a data race. It needs investigated to see if the race extends to all Beats or is limited to mockbeat (which would be a lower priority).

[mockbeat.log.txt](https://github.com/elastic/beats/files/1258877/mockbeat.log.txt)

https://beats-ci.elastic.co/job/elastic+beats+6.0+multijob-windows/beat=libbeat,label=windows/lastCompletedBuild/artifact/src/github.com/elastic/beats/libbeat/build/system-tests/run/test_base.Test.test_invalid_config_cli_param/mockbeat.log/*view*/

adriansr:
Mockbeat expects `Beater.Stop` to be called after `Beater.Run` has
initialized some shared state. This is a race as `Stop` can be called at
any moment, even before `Run` is ever called.

This caused race warnings and segfaults when running unit tests.

Closes #5037

### DIFF
--- a/libbeat/mock/mockbeat.go
+++ b/libbeat/mock/mockbeat.go
@@ -14,8 +14,7 @@ var Version = "9.9.9"
 var Name = "mockbeat"
 
 type Mockbeat struct {
-	done   chan struct{}
-	client beat.Client
+	done chan struct{}
 }
 
 // Creates beater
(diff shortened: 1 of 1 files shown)


urso:
This is a general problem all beats currently have. The `Stop` method might need to cleanup/close state/variables that are only created during the run method. Reason `Stop` closes the client is, if the client is blocking the beat will never stop.

+100 If this PR really solves some issues when using `-race`.

But longterm I'd like to have some support to signal shutdown in the run method, potentially removing the Stop method. e.g. think `context.Context`. Or something more tailored like:

```
type Signal struct { ... }

func (s *Signal) C() <-chan struct{} { ... }

func (s *Signal) CloseOnSig(c io.Closer) { ... }
func (s *Signal) CBOnSig(fn func()) { ... }
```

PR #5431, opened 2017-10-24
Handle certificate generation with password for testing purpose

sterchelen:
Change made on GenCertsForIPIfMIssing function:
- Rename the function to GenCertForTestingPurpose
- Replace ip parameter by host, can be an ip or a dns name
- Add password parameter to generate encrypted private key


sterchelen:
Is there anyone who can handle this PR ?

PR #5434, opened 2017-10-25
Refactor add_kubernetes_metadata to support autodiscovery

vjsamuel:
Autodiscovery when it is available must be able to use indexers. Indexers are currently exclusive to `add_kubernetes_metadata`. This also prevents collectbeat from using them. This PR moves code around to reduce file sizes and exposes some of the needed functionality in `add_kubernetes_metadata`.


exekias:
jenkins, test it please

exekias:
LGTM, as a note, we will also need to move podwatcher to some common place, as I did for the docker one here: https://github.com/elastic/beats/pull/5245

PR #5436, opened 2017-10-25
Fix double @timestamp key when using JSON decoding

tsg:
The MergeJSONFields was adding the parsed @timestamp key to the fields,
instead of modifying it into the Event structure. This change makes it
return the new timestamp (or the empty Timestamp if no change required),
and the caller sets it into the event.


friesoft:
@tsg thank you for the fast fix! 👍

Issue #5404, opened 2017-10-18
metricbeat - metricbeat stops publishing events to Redis


alextxm:
Metricbeat 6.0.0-rc1 stops publishing events to Redis (using the redis output module) after some time, depending on the events volume, but usually in the 4000-5000 events range.

Enabling debugging does not give any futher info; after reaching 4000-5000 events metricbeat continue running but no events are published to redis while the log is filled with INFO level messages like this one:
`INFO Non-zero metrics in the last 30s: beat.memstats.memory_alloc=7872 beat.memstats.memory_total=7872`

It should be noted that metricbeat successfully works when configured to use Logstash publishing instead (using the logstash output module).
The Redis output module configuration uses default values except for the following customized values: hosts, port, password, key. i've also tried raising the bulk_max_size value but nothing changed.

Tests were performed enablding only the system module, only the vsphere module and finally both of them; all the configurations triggered the problem; please also note the same configurations worked fine in 6.0.0-alpha2.

Tested with:
- metricbeat 6.0.0-rc1 on Windows Server 2008R2
- redis 3.0.2 on Ubuntu 16.04
- redis 4.0.2 on Ubuntu 16.04
- logstash 5.4.0 on Ubuntu 16.04

alextxm:
I've done some more testing and it happens on both 6.0-beta1/beta2/rc1 but not alpha2 (alpha2 works fine). Publishing stops just after 4096 items as you can see per this screenshots taken from redis-cli.
![mb-redis-60b1](https://user-images.githubusercontent.com/6169392/31890659-2ca7d664-b804-11e7-8184-c6d63c3fb2fe.PNG)
![mb-redis-60b2](https://user-images.githubusercontent.com/6169392/31890660-2cd3b34c-b804-11e7-9b6a-b18970d63466.PNG)

Could please someone check this issue?
Thanks in advance

urso:
Thanks for testing. Can you share your redis config? Are you using a list or a channel?

urso:
I was able to reproduce the problem. It should be resolved with #5441. Thanks for testing.

PR #5441, opened 2017-10-25
Fix missing ACK in redis output

urso:
Resolves: #5404

### DIFF
--- a/libbeat/outputs/redis/client.go
+++ b/libbeat/outputs/redis/client.go
@@ -128,7 +128,10 @@ func (c *client) Publish(batch publisher.Batch) error {
 	if rest != nil {
 		c.stats.Failed(len(rest))
 		batch.RetryEvents(rest)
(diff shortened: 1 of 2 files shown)


libbeat/outputs/redis/client.go
```diff
@@ -128,7 +128,10 @@ func (c *client) Publish(batch publisher.Batch) error {
 	if rest != nil {
```
kvch: Question so I would undestand: what does publish only return event if an error occured?

kvch: Now I see what is happening here.

urso: The publish functions do return the events not being published yet. Normally that's true if an error occured. We always return the error value, just in case.

kvch: Thanks.

PR #5446, opened 2017-10-26
corrected format of used metricsets

rsdrakh:
when using 
`metricsets: ["datastore, host, virtualmachine"]`
metricbeat considers that one metricset (and fails)
`metricsets:["datastore", "host", "virtualmachine"]`
resolves correctly to 3 metricsets


exekias:
Thank you for taking the time @rsdrakh, in this case I'm afraid this was done already: https://github.com/elastic/beats/pull/5416 :/

PR #5443, opened 2017-10-25
Fix relative paths in prospectors definitions

tsg:
If a relative path was used in a prospector definition, it could happen
that when loading the state from the registry, the path didn't match the
pattern, which made it ignore the saved state. Fixes #5442.

Additionally, this PR cleans up some code related to the `recursive_glob`
setting. The docs used to claim that the feature is disabled by default, but
that wasn't the case and it was impossible to disable it. This change leaves
it enabled by default, but makes it possible to disable it.

PR #5449, opened 2017-10-26
Make python tests not executables

tsg:
Turns out that if a `test_*.py` file is executable, nosetests
doesn't pick it up, so the tests are not executed on Linux/Mac.
This was the case with two files in Filebeat.

This adds a `make check` check to look for such files, so this
doesn't happen in the future.

### DIFF
--- a/filebeat/tests/system/test_prospector.py
+++ b/filebeat/tests/system/test_prospector.py
@@ -3,6 +3,7 @@
 from filebeat import BaseTest
 import os
 import time
+import unittest
 
 from beat.beat import Proc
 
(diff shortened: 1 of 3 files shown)


tsg:
I opened this separately from #5443 so we can backport it immediately to the 6.0 branch.

andrewkroh:
`find: invalid mode ‘+111’` coming from Jenkins

andrewkroh:
man find
> -perm +mode
> This is no longer supported (and has been deprecated since 2005).  Use -perm /mode instead.

These work on Linux but not on macOS. hmm

`-perm /a+x`
`-type f  -executable`

andrewkroh:
`test_prospector.Test.test_harvester_limit` is failing on Travis for both Linux and macOS, but it's passing on Jenkins.

tsg:
That test seems stable on my computer, so I'm pretty clueless on why it might fail on Travis only.

tsg:
No, I spoke to soon. The test is flaky and I could get it to fail on my laptop when running it in a loop.

tsg:
I opened #5458 for the flaky test.

tsg:
I skipped the flaky test, we have #5458 to track that.

tsg:
jenkins, test it

Issue #5440, opened 2017-10-25
Filebeat 6.0.0-rc1: json.overwrite not working with @timestamp


friesoft:
- Version: 6.0.0-rc1
- Operating System: Red Hat 7
- Steps to Reproduce: 

Forwarding json messages from a log file to logstash using Filebeat does not set the @timestamp attribute correcty. The timestamps are already off on the Filebeat side (console output). 
With 5.0.1 and 5.6.3 (latest GA) it is working, with 6.0.0-rc1 we get duplicate @timestamp fields.

Sample JSON to reproduce the issue:
`{ "@timestamp": "2017-10-17T10:03:14.301Z", "request": "/" }`

Sample Filebeat Config 5.0.1/5.6.3 (working):
```
filebeat.prospectors:
- input_type: log
  paths:
    - serverlogs/apache.json
  json.keys_under_root: true
  json.add_error_key: true
  json.overwrite_keys: true
  fields_under_root: true

output.console:
  pretty: true
```

Sample Filebeat Config 6.0.0-rc1 (not working):
```
filebeat.prospectors:
- prospector_type: log
  paths:
    - serverlogs/apache.json
  json.keys_under_root: true
  json.add_error_key: true
  json.overwrite_keys: true
  fields_under_root: true

output.console:
  pretty: true
```

Running with a 5.0.1 installation (download, untar, add serverlogs/apache.json file with sample provided above, add filebeat.json.yml with sample provided above, run):
```
[friedreb@pc64901 filebeat-5.0.1-linux-x86_64]$ rm -rf data/ && ./filebeat -c filebeat.json.yml
{
  "@timestamp": "2017-10-17T10:03:14.301Z",
  "beat": {
    "hostname": "pc64901",
    "name": "pc64901",
    "version": "5.0.1"
  },
  "input_type": "log",
  "offset": 61,
  "request": "/",
  "source": "serverlogs/apache.json",
  "type": "log"
}
```

Running with a 6.0.0-rc1 installation (download, untar, add serverlogs/apache.json file with sample provided above, add filebeat.json.yml with sample provided above, run):
```
[friedreb@pc64901 filebeat-6.0.0-rc1-linux-x86_64]$ ./filebeat -c filebeat.json.yml
{
  "@timestamp": "2017-10-25T07:54:57.673Z",
  "@metadata": {
  "beat": "filebeat",
  "type": "doc",
  "version": "6.0.0-rc1"
  },
[...]

Issue #5437, opened 2017-10-25
Double `@timestamp` key when using the JSON decoder


tsg:
Initially reported on [discuss](https://discuss.elastic.co/t/6-0-0-rc1-json-overwrite-keys-not-working-with-timestamp/105189), the bug is about adding two `@timestamp` keys when using JSON decoding on a doc that has a valid `@timestamp` key in it.

* [x] Hot fix in the 6.0 branch #5436 
* [x] Forward port to master #5444
* [x] Add integration tests on JSON decoding that check for double `@timestamp` #5455
* [ ] Refactor the `MergeJSONFields` function to work with `beat.Event` instead of `common.MapStr`

kvch:
An issue for the initial bug is opened: https://github.com/elastic/beats/issues/5440

kvch:
Resolved in later versions.

PR #5455, opened 2017-10-26
Catch duplicate keys in integration tests

tsg:
This enabled duplicate key detection when we read the results in
the system tests. This makes sure that if we ever have a duplicate key
like in #5440 we will notice.

Part of #5437.

### DIFF
--- a/libbeat/tests/system/beat/beat.py
+++ b/libbeat/tests/system/beat/beat.py
@@ -227,7 +227,8 @@ class TestCase(unittest.TestCase, ComposeMixin):
                     break
 
                 try:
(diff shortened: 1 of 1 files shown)


libbeat/tests/system/beat/beat.py
```diff
@@ -249,11 +249,22 @@ def read_output_json(self, output_file=None):
                     # hit EOF
                     break
 
-                event = json.loads(line)
+                event = json.loads(line, object_pairs_hook=self.json_raise_on_duplicates)
                 del event['@metadata']
                 jsons.append(event)
         return jsons
 
+    def json_raise_on_duplicates(self, ordered_pairs):
```
andrewkroh: 🥇

PR #5457, opened 2017-10-27
Fix missing length check in PgSQL

tsg:
There was a length check missing.


exekias:
CLA checker is down, feel free to merge wen if it comes back

PR #5469, opened 2017-10-28
i5354: Harmonize field names between Packetbeat HTTP and Metricbeat Module

christiangalsterer:
Should also get merged to master. Shall I create a separate PR for this?


ruflin:
LGTM. Could you open this PR against master instead of 6.0 so we can decide afterwards to which branches we backport it?

You need to run `make fmt` to make CI happy.

christiangalsterer:
ok, will do that

christiangalsterer:
New PR available at #5521

PR #5476, opened 2017-10-30
TLS plugin for packetbeat

adriansr:
This patch adds TLS protocol support to Packetbeat.

It will parse the initial handshake and send an event to Elasticsearch for every TLS connection.

Events contain:
- ciphers supported by client
- cipher selected by server
- extensions
- client and server certificate chain (if any)
  - key algorithm
  - signing algorithm
  - subject alternative name (SAN)
  - validity dates
- whether the session is resumed using session IDs or tickets
- alert (if any)

There is support for Application Layer Protocol Negotiation (ALPN) extension to TLS. It parses extension data found in hello messages. This involves a list of protocols in the client hello:

    "application_layer_protocol_negotiation": "h2,http/1.1"

And the selected protocol in the server hello.

    "application_layer_protocol_negotiation": "h2"

References

ALPN: https://en.wikipedia.org/wiki/Application-Layer_Protocol_Negotiation


adriansr:
Please review when you have time. It is a work in progress (working on the tests now)

packetbeat/_meta/fields.yml
```diff
@@ -1748,3 +1748,23 @@
...
 
+- key: tls
+  title: "TLS"
+  description: >
+    TLS-specific event fields.
+  fields:
+    - name: tls
+      type: group
+      fields:
+        - name: negotiated cypher
```
andrewkroh: This field name has a space in it. Probably you want an underscore here. But it looks like this file is progress because I don't see this field in the code so I'm sure you be updating all of this.

packetbeat/protos/tls/parse.go
```diff
@@ -0,0 +1,506 @@
...
+
+func (s serverHelloMessage) toMap() common.MapStr {
+	m := s.commonHelloFields.toMap()
+	m["cipher"] = s.cipher.String()
+	m["compression_method"] = s.compMethod.String()
+	m["extensions"] = s.extensions
+	return m
+}
+
+func (parser *parser) parse(buf *streambuf.Buffer, conn *tlsConnectionData, plugin *tlsPlugin) (bool, bool) {
```
urso: Try not to introduce circular dependencies in any way. At optimium the parser must be testable without requiring an instance of `tlsPlugin`. This simplifies potential unit and/or fuzzy testing of the parser only.

The TCP code generate creates different types with different responsibilities on purpose. One type per task like: manage connections, parse, correlate messages, create events, publish events

packetbeat/protos/tls/parse.go
```diff
@@ -0,0 +1,506 @@
...
+
+	if conn.handshakeCompleted {
```
urso: 1. Abstraction leak. The parser should be about parsing messages from receive buffer only. The parser should not have to care about the connection state if possible.

2. if connection is 'completed' and there is nothing new to learn during the lifetime of the connect, clear the buffers more early and consider not appending to the buffer again. This will make the module much more lightweight.

packetbeat/protos/tls/parse.go
```diff
@@ -0,0 +1,506 @@
...
+		limit := recordHeaderSize + int(header.length)
+		if !buf.Avail(limit) {
+			// wait for complete record
+			return true, false
+		}
+
+		switch header.recordType {
+		case recordTypeChangeCipherSpec: // single message of size 1 (byte 1)
+			conn.handshakeCompleted = true
+			plugin.sendEvent(conn)
```
urso: Abstraction leak. Parser should not directly trigger an event, depending on potential connection state. Instead parser should return/report messages only.

packetbeat/protos/tls/parse.go
```diff
@@ -0,0 +1,506 @@
...
+			bufferView{&parser.handshakeBuf, handshakeHeaderSize, limit})
+		parser.handshakeBuf.Advance(limit)
+	}
+	if parser.handshakeBuf.Len() == 0 {
+		parser.handshakeBuf.Reset()
+	}
+	return true
+}
+
+func (parser *parser) processHandshake(conn *tlsConnectionData, handshakeType handshakeType, reader bufferView) bool {
```
urso: This looks like a task for the message 'correlator'. Note, depending on deployment (e.g. port forwarding), packetbeat might potentially see packets out of order.

urso:
Nice progress. Can't wait for finally having TLS envelope support 🎉

adriansr:
Will update with correct fields.yml

packetbeat/_meta/fields.yml
```diff
@@ -1748,3 +1748,293 @@
...
+            - name: version
+              type: keyword
+              description: >
+                The version of the TLS protocol by which the client wishes to
+                communicate during this session.
+
+            - name: timestamp
+              type: long
+              description: >
+                The current time and date in standard UNIX 32-bit format
```
andrewkroh: Is this seconds since Unix epoch? Either way I suggest converting it to a date rather than a long so that’s easier to use in ES.

packetbeat/_meta/fields.yml
```diff
@@ -1748,3 +1748,293 @@
...
+                The list of compression methods the client supports.
+                See https://www.iana.org/assignments/comp-meth-ids/comp-meth-ids.xhtml
+
+            - name: extensions
+              type: group
+              description: The hello extensions provided by the client.
+              fields:
+                - name: server_name_indication
+                  type: string
+                  description: Comma-separated list of server names
```
andrewkroh: Instead of comma separated, I think a list would be better for searching and aggregating.

packetbeat/_meta/fields.yml
```diff
@@ -1748,3 +1748,293 @@
...
+
+                - name: application_layer_protocol_negotiation
+                  type: string
+                  description: Comma-separated list of protocols
+
+                - name: session_ticket
+                  type: string
+                  description: >
+                    Length of the session ticket, if provided, or empty
```
andrewkroh: “or empty” - does this mean the field is not sent when a value is not present?

packetbeat/_meta/fields.yml
```diff
@@ -1748,3 +1748,293 @@
...
+              type: group
+              description: The hello extensions provided by the server.
+              fields:
+                - name: application_layer_protocol_negotiation
+                  type: string
+                  description: Negotiated application layer protocol
+
+                - name: session_ticket
+                  type: string
+                  description: Empty
```
andrewkroh: Could you please clarify here.

packetbeat/_meta/fields.yml
```diff
@@ -1748,3 +1748,293 @@
...
+          type: group
+          description: Certificate provided by the client for authentication.
+          fields:
+
+            - name: version
+              type: long
+              description: X509 format version.
+
+            - name: serial_number
+              type: keyword
```
andrewkroh: Are the some benefits to sending this as a keyword rather than a long? Will users be expecting to see hex? Maybe we should have a field formater in Kibana for hex (possibly there is one; I didn’t check).

packetbeat/_meta/fields.yml
adriansr: Right. Will do

packetbeat/_meta/fields.yml
adriansr: The serial number can be up to 160 bits long so it doesn't fit on a numeric type.

I have changed the format to decimal. The most common  way of representation seems to be decimal up to 64 bit in length and then hexadecimal number or hex dump of individual bytes. I will stick with decimal in a keyword field.

monicasarbu:
@adriansr I just realized that this PR doesn't have a changelog. I think the easiest would be if @tsg adds it to the https://github.com/elastic/beats/pull/5853.

Issue #5378, opened 2017-10-11
filebeat add_kubernetes_metadata requires namespace param


tarasglek:
If I deploy filebeat in kubernetes under namespace that is not kube-system, it fails to add kubernetes metadata.

```
    processors:
    - add_kubernetes_metadata:
        namespace: default
```
This should  be set automatically from `/var/run/secrets/kubernetes.io/serviceaccount/namespace`

exekias:
This is what we have been using so far: https://github.com/elastic/beats/blob/master/deploy/kubernetes/filebeat/filebeat-daemonset.yaml#L34-L37, but your idea looks good to me, do you have any link on the docs for this?

PR #5482, opened 2017-10-31
Remove hard-coded namespace from k8s config

jpeeler:
See commit text for more details. I haven't actually tested this yet, but looking at the client code I can't see how this would fail in any previously working scenario. @tarasglek care to verify all is well now?

(#5378)

### DIFF
--- a/libbeat/processors/add_kubernetes_metadata/config.go
+++ b/libbeat/processors/add_kubernetes_metadata/config.go
@@ -30,12 +30,11 @@ type Enabled struct {
 
 type PluginConfig []map[string]common.Config
 
-func defaultKuberentesAnnotatorConfig() kubeAnnotatorConfig {
+func defaultKubernetesAnnotatorConfig() kubeAnnotatorConfig {
 	return kubeAnnotatorConfig{
 		InCluster:       true,
 		SyncPeriod:      1 * time.Second,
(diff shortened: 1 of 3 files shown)


libbeat/processors/add_kubernetes_metadata/config.go
```diff
@@ -30,12 +30,11 @@ type Enabled struct {
 
 type PluginConfig []map[string]common.Config
 
-func defaultKuberentesAnnotatorConfig() kubeAnnotatorConfig {
+func defaultKubernetesAnnotatorConfig() kubeAnnotatorConfig {
 	return kubeAnnotatorConfig{
 		InCluster:       true,
 		SyncPeriod:      1 * time.Second,
 		CleanupTimeout:  60 * time.Second,
-		Namespace:       "kube-system",
```
exekias: I think it's best if we remove `Namespace` field from the kubeAnnotatorConfig, and move all references to `client.Namespace`

exekias:
jenkins, test it please

libbeat/processors/add_kubernetes_metadata/config.go
jpeeler: Yeah, I considered doing that initially as well. Will adjust now!

exekias:
Just gave this a try, works for me, thank you :tada:

PR #5487, opened 2017-10-31
Add `cloud.id` support to Kubernetes manifests

exekias:
This change adds `ELASTIC_CLOUD_ID` and `ELASTIC_CLOUD_AUTH` env variables to Kubernetes manifests, so user can use those variables with cloud instead of the common `ELASTICSEARCH_*`.

When they are not used common ones will prevail

I also updated Makefile to always update YAML manifests

### DIFF
--- a/deploy/kubernetes/Makefile
+++ b/deploy/kubernetes/Makefile
@@ -1,7 +1,9 @@
 ALL=filebeat metricbeat
 BEAT_VERSION=$(shell head -n 1 ../../libbeat/docs/version.asciidoc | cut -c 17- )
 
-all: ${ALL:=-kubernetes.yaml}
+.PHONY: all $(ALL)
+
+all: $(ALL)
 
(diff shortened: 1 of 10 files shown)


deploy/kubernetes/Makefile
```diff
@@ -1,7 +1,9 @@
 ALL=filebeat metricbeat
 BEAT_VERSION=$(shell head -n 1 ../../libbeat/docs/version.asciidoc | cut -c 17- )
 
-all: ${ALL:=-kubernetes.yaml}
+.PHONY: $(ALL)
```
tsg: Shoudn't `all` also be a phony target? In our other makefiles, we just make almost all targets `.PHONY`.

exekias: It worked without it on my machine, but I just pushed it in, just in case it's needed for other systems

PR #5488, opened 2017-10-31
New metric: uptime

kvch:
Uptime is exposed as a metric called `beat.info.uptime`. Its value is the number of seconds passed since start. @tsullivan 

I have changed the architecture a bit. `startTime` is initialized in `instance/beat.go`. Then later this time is passed to `logp` thorugh `logp.Init`. Thus, `logp` can display uptime when the Beat stops.

I am still having a problem with the value of the metric. It the event it shows 30 always.
```
INFO Non-zero metrics in the last 30s:
beat.info.uptime=30
beat.memstats.memory_alloc=11440
beat.memstats.memory_total=11440
```
But when I check the value it seems to be correct.

I opened the PR so the architecture could be validated until I fix the problem.

### DIFF
--- a/libbeat/cmd/instance/metrics.go
+++ b/libbeat/cmd/instance/metrics.go
@@ -7,23 +7,14 @@ import (
 	"github.com/elastic/beats/libbeat/monitoring"
 )
 
-type memstatsVar struct{}
-
-type fixedstatsVar struct{}
-
-var (
-	metrics = monitoring.Default.NewRegistry("beat")
-)
-
 func init() {
-	var ms memstatsVar
-	metrics.Add("memstats", ms, monitoring.Reported)
+	metrics := monitoring.Default.NewRegistry("beat")
 
-	var fs fixedstatsVar
-	metrics.Add("info", fs, monitoring.Reported)
+	monitoring.NewFunc(metrics, "memstats", reportMemStats, monitoring.Report)
+	monitoring.NewFunc(metrics, "info", reportInfo, monitoring.Report)
 }
 
-func (memstatsVar) Visit(m monitoring.Mode, V monitoring.Visitor) {
+func reportMemStats(m monitoring.Mode, V monitoring.Visitor) {
 	var stats runtime.MemStats
 	runtime.ReadMemStats(&stats)
 
@@ -37,11 +28,10 @@ func (memstatsVar) Visit(m monitoring.Mode, V monitoring.Visitor) {
 	}
 }
 
-func (fixedstatsVar) Visit(m monitoring.Mode, V monitoring.Visitor) {
-	uptime := int64(time.Now().Sub(startTime).Seconds())
-
+func reportInfo(_ monitoring.Mode, V monitoring.Visitor) {
 	V.OnRegistryStart()
 	defer V.OnRegistryFinished()
 
-	monitoring.ReportInt(V, "uptime", uptime)
+	uptime := int64(time.Since(startTime).Seconds()) * 1000
+	monitoring.ReportInt(V, "uptime.ms", uptime)
 }
--- a/libbeat/cmd/instance/metrics.go
+++ b/libbeat/cmd/instance/metrics.go
@@ -2,11 +2,14 @@ package instance
 
 import (
 	"runtime"
+	"time"
 
 	"github.com/elastic/beats/libbeat/monitoring"
 )
 
 type memstatsVar struct{}
+
+type fixedstatsVar struct{}
 
 var (
 	metrics = monitoring.Default.NewRegistry("beat")
@@ -15,6 +18,9 @@ var (
 func init() {
 	var ms memstatsVar
 	metrics.Add("memstats", ms, monitoring.Reported)
+
+	var fs fixedstatsVar
+	metrics.Add("info", fs, monitoring.Reported)
 }
 
 func (memstatsVar) Visit(m monitoring.Mode, V monitoring.Visitor) {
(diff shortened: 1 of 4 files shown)


libbeat/cmd/instance/metrics.go
```diff
@@ -30,3 +36,12 @@ func (memstatsVar) Visit(m monitoring.Mode, V monitoring.Visitor) {
...
 	}
 }
+
+func (fixedstatsVar) Visit(m monitoring.Mode, V monitoring.Visitor) {
+	uptime := int64(time.Now().Sub(startTime).Seconds())
+
+	V.OnRegistryStart()
+	defer V.OnRegistryFinished()
+
+	monitoring.ReportInt(V, "uptime", uptime)
```
andrewkroh: Can we send `uptime.ms` (number of milliseconds since unix epoch)? This is how we send uptime in the [system-uptime metricset](https://github.com/elastic/beats/blob/bbc28b08c891d7aab1c6a13c64570a8536a40cdd/metricbeat/module/system/uptime/_meta/fields.yml#L6-L11) and in the new windows-service metricset.

libbeat/cmd/instance/metrics.go
```diff
@@ -30,3 +36,12 @@ func (memstatsVar) Visit(m monitoring.Mode, V monitoring.Visitor) {
 		monitoring.ReportInt(V, "gc_next", int64(stats.NextGC))
 	}
 }
+
+func (fixedstatsVar) Visit(m monitoring.Mode, V monitoring.Visitor) {
+	uptime := int64(time.Now().Sub(startTime).Seconds())
```
andrewkroh: Use `time.Since(startTime)` as it is more clear.

libbeat/cmd/instance/metrics.go
urso: @tsullivan what's the expected format for uptime?

libbeat/cmd/instance/metrics.go
```diff
@@ -2,19 +2,25 @@ package instance
...
 	metrics = monitoring.Default.NewRegistry("beat")
 )
 
 func init() {
 	var ms memstatsVar
 	metrics.Add("memstats", ms, monitoring.Reported)
+
+	var fs fixedstatsVar
+	metrics.Add("info", fs, monitoring.Reported)
 }
```
urso: one can use `monitoring.NewFunc` here:

```
func init() {
...
monitoring.NewFunc(metrics, "info", reportInfo, monitoring.Report)
...
}
```

with

```
func reportInfo(_ monitoring.Mode, V monitoring.Visitor) {
  V.OnRegistryStart()
  defer V.OnRegistryFinished()
  monitoring.ReportInt(V, "uptime", time.Since(startTime).Seconds())
}
```

tsg:
> I am still having a problem with the value of the metric. It the event it shows 30 always.

Just in case you didn't figure it out in the mean time: that's more the effect of the logging that we have in logp. It always subtracts the old value, see here: https://github.com/elastic/beats/blame/master/libbeat/logp/metrics.go#L30

So I think the metric is correct, just the logging of it is wrong.

kvch:
I have addressed the notes except for seconds vs. milliseconds question as we are still waiting for input.

kvch:
We discussed with @tsullivan that seconds are sufficient for them.

kvch:
To stay consistent with existing uptime handling I changed it to milliseconds.
@tsullivan Please, let us know if it is problem.

tsullivan:
I don't foresee any problem with the metric being milliseconds. Thanks for checking!

libbeat/cmd/instance/metrics.go
```diff
@@ -30,3 +27,11 @@ func (memstatsVar) Visit(m monitoring.Mode, V monitoring.Visitor) {
 		monitoring.ReportInt(V, "gc_next", int64(stats.NextGC))
 	}
 }
+
+func reportInfo(_ monitoring.Mode, V monitoring.Visitor) {
+	V.OnRegistryStart()
+	defer V.OnRegistryFinished()
+
+	uptime := int64(time.Since(startTime).Seconds()) * 1000
```
urso: why use `Seconds()`, do a floor operation and multiply by 1000? 

shouldn't it be:

```
now := time.Since(startTime)
int64(now.Seconds()) * 1000 + int64(now.Nanoseconds() / 1000000)
```

libbeat/cmd/instance/metrics.go
```diff
@@ -30,3 +27,12 @@ func (memstatsVar) Visit(m monitoring.Mode, V monitoring.Visitor) {
 		monitoring.ReportInt(V, "gc_next", int64(stats.NextGC))
 	}
 }
+
+func reportInfo(_ monitoring.Mode, V monitoring.Visitor) {
+	V.OnRegistryStart()
+	defer V.OnRegistryFinished()
+
+	delta := time.Since(startTime)
+	uptime := int64(delta.Seconds())*1000 + int64(delta.Nanoseconds()/1000000)
```
andrewkroh: A `time.Duration` is an `int64` giving the number of nanoseconds so you should just need to divide by `time.Millisecond` to convert to milliseconds.

`uptimeMillis := int64(time.Since(startTime)/time.Millisecond)`

tsullivan:
I collected some samples of metrics and looked at the new uptime metric, and I think that there is a bug in it:

```
"timestamp": "2017-11-09T22:02:58.532Z"                  "uptime.ms": 5220012
"timestamp": "2017-11-09T22:02:48.532Z"                  "uptime.ms": 5200012
"timestamp": "2017-11-09T22:02:38.531Z"                  "uptime.ms": 5180011
"timestamp": "2017-11-09T22:02:28.532Z"                  "uptime.ms": 5160012
"timestamp": "2017-11-09T22:02:18.533Z"                  "uptime.ms": 5140013
"timestamp": "2017-11-09T22:02:08.531Z"                  "uptime.ms": 5120011
"timestamp": "2017-11-09T22:01:58.530Z"                  "uptime.ms": 5100010
"timestamp": "2017-11-09T22:01:48.532Z"                  "uptime.ms": 5080012
"timestamp": "2017-11-09T22:01:38.531Z"                  "uptime.ms": 5060011
"timestamp": "2017-11-09T22:01:28.532Z"                  "uptime.ms": 5040012
```

If you'll notice, my timestamps are going up by 10 seconds, but the uptime values are going up by 20000 milliseconds. Is the uptime value getting doubled somehow?

kvch:
Yes, you are right. I opened a PR to fix it: https://github.com/elastic/beats/pull/5552

Issue #5458, opened 2017-10-27
Race condition on harvester limit


tsg:
The `test_prospector.py:Test.test_harvester_limit` test is flaky. The fail can be reproduced by running it in a while loop like this:

```
 while nosetests test_prospector.py:Test.test_harvester_limit; do echo .; done
```

On my computer, this fails after 10 or 20 iterations.

Looking at the logs and [code](https://github.com/elastic/beats/blob/master/filebeat/prospector/log/prospector.go#L587-L618), it seems to be a race condition between checking `p.harvesters.Len()` and `p.harvesters.Start()`. We could serialise the whole `startHarvester()` method with a new lock, but I'm slightly worried about dead locks, because the harvester registry has it's own lock. So I'm opening a ticket for now to separate this from #5449.

PR #5490, opened 2017-10-31
Race condition on harvester limit

adriansr:


tsg:
Cross-linking #5458.

filebeat/harvester/registry.go
```diff
@@ -5,6 +5,8 @@ import (
 
 	uuid "github.com/satori/go.uuid"
 
+	"fmt"
```
andrewkroh: Please move this up to the top with the other stdlib imports.

filebeat/tests/system/test_prospector.py
```diff
@@ -583,7 +583,7 @@ def test_skip_symlinks(self):
         # Make sure there is only one entry, means it didn't follow the symlink
         assert len(data) == 1
 
-    @unittest.skip("Flaky due to race. Opened https://github.com/elastic/beats/issues/5458")
+    #@unittest.skip("Flaky due to race. Opened https://github.com/elastic/beats/issues/5458")
```
andrewkroh: I think this can be removed since it should be fixed.

filebeat/harvester/registry.go
```diff
@@ -59,14 +60,14 @@ func (r *Registry) WaitForCompletion() {
...
-func (r *Registry) Start(h Harvester) {
+func (r *Registry) Start(h Harvester) error {
 	// Make sure stop is not called during starting a harvester
 	r.Lock()
 	defer r.Unlock()
 
 	// Make sure no new harvesters are started after stop was called
 	if !r.active() {
-		return
+		return fmt.Errorf("registry already stopped")
```
urso: No formatting, use `errors.New` + (optional here) predefined variable.

Issue #4974, opened 2017-08-23
bug when space in url


mazhan465:
When I get a request with space in url like this:
"http://a.example.com/index?id=1 asd"
The packetbeat can't work normally.
This is a Bug When parse http protocol.

while-loop:
I'm not sure if this is actually a bug. Currently the code parses the request-line by whitespace which is the correct way to do it (as described in the HTTP RFC 7230 [section 3.1.1](https://tools.ietf.org/html/rfc7230#section-3.1.1)).

In the [code](https://github.com/elastic/beats/blob/master/packetbeat/protos/http/http_parser.go#L164), if it is unable to parse the request-line, then it stops parsing the request.

This may be a bug in the client you are using that is sending the HTTP request and not correctly encoding the URL.

> Recipients typically parse the request-line into its component parts
   by splitting on whitespace (see Section 3.5), since no whitespace is
   allowed in the three components.  Unfortunately, some user agents
   fail to properly encode or exclude whitespace found in hypertext
   references, resulting in those disallowed characters being sent in a
   request-target.

> ... A recipient SHOULD NOT attempt
   to autocorrect and then process the request without a redirect, since
   the invalid request-line might be deliberately crafted to bypass
   security filters along the request chain.


Cheers!

mazhan465:
The packetbeat is used to monitoring flow .When a hacker attemp to attack us using a deformed payload which maybe inculde a space.And in this case,packetbeat will dorp the packet ,so the hacker will Bypass monitoring。

ph:
Fixing the code to parse the URI with space doesn't look too complicated, what worries me is how the parameters parsing will behave in this case since we don't have the space url encoded.

PR #5495, opened 2017-11-01
Add support for space in get request

ph:
This fix an issue when the http request contains a space instead of
breaking the line with `bytes.fields` we are finding the start and the end
of the URI using the `METHOD` verb and the `HTTP/{VERSION}`. This will
allow packet beat to record theses request instead of ignoring them.

Fixes: #4974

### DIFF
--- a/packetbeat/protos/http/http_parser.go
+++ b/packetbeat/protos/http/http_parser.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"time"
+	"unicode"
 
 	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/common/streambuf"
@@ -74,8 +75,9 @@ var (
 
 	constCRLF = []byte("\r\n")
 
-	constClose     = []byte("close")
-	constKeepAlive = []byte("keep-alive")
+	constClose       = []byte("close")
+	constKeepAlive   = []byte("keep-alive")
+	constHTTPVersion = []byte("HTTP/")
 
 	nameContentLength    = []byte("content-length")
 	nameContentType      = []byte("content-type")
@@ -145,7 +147,7 @@ func (*parser) parseHTTPLine(s *stream, m *message) (cont, ok, complete bool) {
 		}
 		return false, false, false
 	}
-	if bytes.Equal(fline[0:5], []byte("HTTP/")) {
+	if bytes.Equal(fline[0:5], constHTTPVersion) {
 		//RESPONSE
 		m.isRequest = false
 		version = fline[5:8]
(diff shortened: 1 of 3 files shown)


ph:
@urso This is the PR I want to discuss with you, this change for accuracy and supporting badly form URI, wring some slowness to that part of the code. It add bit more allocation, but It should be faster if the URI has more space in it than using ` bytes.Fields(fline)`. WDYT?


### This PR
```
BenchmarkHTTPSimpleRequest-4       	 1000000	      1790 ns/op
```

### master
```
BenchmarkHTTPSimpleRequest-4       	 1000000	      2136 ns/op
```

packetbeat/protos/http/http_parser.go
```diff
@@ -160,20 +162,23 @@ func (*parser) parseHTTPLine(s *stream, m *message) (cont, ok, complete bool) {
...
 			if isDebug {
 				debugf("Couldn't understand HTTP request: %s", fline)
 			}
 			return false, false, false
 		}
 
-		m.method = common.NetString(slices[0])
-		m.requestURI = common.NetString(slices[1])
+		m.method = common.NetString(fline[:afterMethodIdx])
+		m.requestURI = common.NetString(fline[afterMethodIdx+1 : afterRequestURIIdx])
```
adriansr: if there is more than one space between URI and HTTP_VERSION it will be stored into requestURI. I think you should check for that (bytes.Fields did skip consecutive separators)

packetbeat/protos/http/http_parser.go
```diff
@@ -160,20 +162,23 @@ func (*parser) parseHTTPLine(s *stream, m *message) (cont, ok, complete bool) {
...
 
-		if bytes.Equal(slices[2][:5], []byte("HTTP/")) {
+		if bytes.Equal(fline[afterRequestURIIdx+1:afterRequestURIIdx+6], constHTTPVersion) {
```
adriansr: Suggestion
+len(constHTTPVersion)+1 instead of +6
and :len(constHTTPVersion) in the line above

ph:
@adriansr thanks, I will do another check.

packetbeat/protos/http/http_parser.go
ph: I gave some thoughts about it, I don't think we should check for that, if we do it it would mean that we could _potentially normalize_ the query string parameters. 

If we look at the following case:
```
GET /e/b?movie=a brand new world HTTP/1.1 => '/e/b?movie=a brand new world'
GET /e/b?movie=a brand new world  HTTP/1.1 =>  
/elastic/beats?movie=a brand new world '
```

When we extract the parameters that should gives us 2 distinct values for the `movie` field: `a brand new world` and `a brand new world ` (not the space at the end of world)

WDYT?

ph:
@adriansr I have updated the code with the change with `len()`

packetbeat/protos/http/http_parser.go
adriansr: I ran a quick check and the server did skip those extra spaces.

Now I was testing again and see that it's more common for servers to return a 400 Bad Request.

So I guess you can just ignore the comment. It's not common at all and spaces are escaped to %20 or + anyway

ph:
This can safely go into 6.0.1, I have created the label for it.

PR #5498, opened 2017-11-02
Add ceph osd tree infomation to metricbeat

elaron:
This PR is trying to collect the osd tree infomation of a ceph cluster. The osd tree infomation is important for understanding and monitoring the structure of a ceph cluster.

In this PR, the code will collect all the osd nodes' weight, status, primary_affinity and other usefull info.


metricbeat/module/ceph/osd_tree/_meta/fields.yml
```diff
@@ -0,0 +1,57 @@
...
+    - name: crush_weight
+      type: float
+      description: >
+        osd node crush weight
+    - name: depth
+      type: long
+      description: >
+        node depth
+    - name: exists
+      type: long
```
exekias: this could be a boolean

metricbeat/module/ceph/osd_tree/_meta/fields.yml
```diff
@@ -0,0 +1,57 @@
...
+  type: group
+  description: >
+    ceph osd tree info
+  fields:
+    - name: id
+      type: long
+      description: >
+        osd or bucket node id
+    - name: name
+      type: string
```
exekias: This can either be `keyword` or `text`. You want to use `keyword` on fields like this, with a limited number of possible values, and `text` for free form strings

metricbeat/module/ceph/osd_tree/_meta/fields.yml
elaron: hi, I have update the fields.yml and the changelog.

elaron:
Is there any way for me to change the modules.d/ceph.yaml.disabled file, and add "osd_tree" as one of the monitor options like this:
metricsets: ["cluster_disk", "cluster_health", "monitor_health", "pool_disk", "**osd_tree**"]

andrewkroh:
Yes, modify [metricbeat/module/ceph/_meta/config.yml](https://github.com/elaron/beats/blob/1f1aa1f177e7d64b919957a1ca864ce116c1c90a/metricbeat/module/ceph/_meta/config.yml#L2) then run `make update`.

elaron:
@exekias fileds.yml has been update, please check~

metricbeat/docs/fields.asciidoc
```diff
@@ -1193,6 +1193,117 @@ Last updated
...
+
+type: long
+
+osd or bucket node typeID
+
+
+[float]
+=== `ceph.osd_tree.children`
+
+type: string
```
exekias: I think this should be keyword?

metricbeat/docs/fields.asciidoc
```diff
@@ -1193,6 +1193,117 @@ Last updated
...
+
+type: keyword
+
+the device class of osd, like hdd, ssd etc.
+
+
+[float]
+=== `ceph.osd_tree.father`
+
+type: string
```
exekias: looks like this should be keyword

metricbeat/docs/fields.asciidoc
```diff
@@ -1193,6 +1193,117 @@ Last updated
...
+
+type: long
+
+the reweight of osd
+
+
+[float]
+=== `ceph.osd_tree.status`
+
+type: string
```
exekias: this should be keyword

metricbeat/docs/fields.asciidoc
elaron: Actually children is a list of children nodes, I don't think it is keyword type.

metricbeat/docs/fields.asciidoc
elaron: father is the parent node's name, it could be keyword, I will fix it.

metricbeat/docs/fields.asciidoc
elaron: yes, it is, I will fix it.

exekias:
We just enabled hound, please don't take everything it says as mandatory, we are still testing the platform

elaron:
ok, I will just update "Id" to "ID", seems it looks much more unified.

metricbeat/docs/fields.asciidoc
exekias: How are children named? It should be either keyword or text. Have a look to ES documentation:https://www.elastic.co/guide/en/elasticsearch/reference/current/array.html: `In Elasticsearch, there is no dedicated array type. Any field can contain zero or more values by default, however, all values in the array must be of the same datatype`

elaron: children list is a ID list seperated by comma, looks like:
"children": "4,3,1,0"
I think it's much more like text. I will change its type from string to text.

exekias: ah thank you, I understand it now, would it make sense to split that string into an array of keywords?

elaron: Actually I display this children list in Grafana as a text, seems like there's no need to make them into an array of keyword now. If there's a need to make them into array in the future, I will pull a patch for it.

exekias:
Thank you again! you rock :tada:

Issue #5433, opened 2017-10-24
Integer overflow in Beats metrics


kvch:
Version: 6.0.0-rc2
Operating System: Debian GNU/Linux 9
Beat: Filebeat

### Configuration
```yml
#=========================== Filebeat prospectors =============================

filebeat.prospectors:
- type: stdin
  enabled: true

#------------------------------- Kafka output ----------------------------------
output.kafka:
  # Boolean flag to enable or disable the output module.
  enabled: true

  # The list of Kafka broker addresses from where to fetch the cluster metadata.
  # The cluster metadata contain the actual Kafka brokers events are published
  # to.
  hosts: ["localhost:9092"]

  # The Kafka topic used for produced events. The setting can be a format string
  # using any event field. To set the topic from document type use `%{[type]}`.
  topic: beats

  # The Kafka event partitioning strategy. Default hashing strategy is `hash`
  # using the `output.kafka.key` setting or randomly distributes events if
  # `output.kafka.key` is not configured.
  partition.hash:
    # If enabled, events will only be published to partitions with reachable
    # leaders. Default is false.
    reachable_only: true

```
### Failure

After sending all events to Kafka:
```
2017/10/24 20:42:46.204889 log.go:99: DBG Source is not continuable: /dev/stdin
2017/10/24 20:42:46.204899 harvester.go:230: INFO End of file reached: . Closing because close_eof is enabled.
2017/10/24 20:42:46.204910 harvester.go:453: DBG Stopping harvester for file:
2017/10/24 20:42:46.204928 harvester.go:463: DBG Closing file:
2017/10/24 20:42:46.204934 harvester.go:474: DBG harvester cleanup finished for file:
2017/10/24 20:42:46.204952 logger.go:18: DBG client: closing acker
2017/10/24 20:42:46.204960 logger.go:18: DBG client: done closing acker
2017/10/24 20:42:46.204966 logger.go:29: DBG client: cancelled 0 events
2017/10/24 20:42:46.876412 client.go:220: DBG finished kafka batch
2017/10/24 20:42:46.876487 logger.go:29: DBG broker ACK events: count=615, start-seq=280240, end-seq=280854
2017/10/24 20:42:46.876507 logger.go:18: DBG ackloop: return ack to broker loop:615
2017/10/24 20:42:46.876518 logger.go:18: DBG ackloop:  done send ack
```

[...]

PR #5505, opened 2017-11-02
Log absolute values for gauges internal metrics

tsg:
We currently don't have a way for the internal Beats metrics to register
themselves as counters or gauges, so the logging code that prints the
values every 30s assumes they are all counters and prints the difference
since the last update.

This issue has became more pressing in 6.0 because we now have several
common gauges. This PR implements a temporary hack that lists these gauges
in logging code. A follow up PR will create a proper solution for the next
release.

Part of #5433.

### DIFF
--- a/libbeat/logp/metrics.go
+++ b/libbeat/logp/metrics.go
@@ -56,6 +56,23 @@ func snapshotMetrics() monitoring.FlatSnapshot {
 	return monitoring.CollectFlatSnapshot(monitoring.Default, monitoring.Full, true)
 }
 
+// List of metrics that are gauges, so that we know for which to
+// _not_ subtract the previous value in the output.
+// TODO: Replace this with a proper solution that uses the metric
+// type from where it is defined. See:
+// https://github.com/elastic/beats/issues/5433
+var gauges = map[string]bool{
+	"libbeat.pipeline.events.active": true,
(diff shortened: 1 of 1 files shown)


libbeat/logp/metrics.go
```diff
@@ -56,6 +56,23 @@ func snapshotMetrics() monitoring.FlatSnapshot {
...
+	"libbeat.pipeline.events.active": true,
+	"libbeat.pipeline.clients":       true,
+	"libbeat.config.module.running":  true,
+	"registrar.states.current":       true,
+	"filebeat.harvester.running":     true,
+	"filebeat.harvester.open_files":  true,
+	"beat.memstats.memory_total":     true,
+	"beat.memstats.memory_alloc":     true,
+	"beat.memstats.gc_next":          true,
+}
```
andrewkroh: [This one](https://github.com/elastic/beats/blob/33270f1305dec9f286bcd3a6f91daac252d9ab21/winlogbeat/eventlog/cache.go#L54) presents a problem because the name is not known ahead of time. Perhaps a pattern could be used to handle it or we just ignore it (I doubt anyone will be using this metric before we have a final solution).

tsg: Yeah, I saw that one. I'd say we just ignore it in this phase and it can be handled with the real solution.

PR #5507, opened 2017-11-03
Cherry-pick #5505 to master: Log absolute values for gauges internal metrics

tsg:
Cherry-pick of PR #5505 to master branch. Original message: 

We currently don't have a way for the internal Beats metrics to register
themselves as counters or gauges, so the logging code that prints the
values every 30s assumes they are all counters and prints the difference
since the last update.

This issue has became more pressing in 6.0 because we now have several
common gauges. This PR implements a temporary hack that lists these gauges
in logging code. A follow up PR will create a proper solution for the next
release.

Part of #5433.


ruflin:
Merged as it is just a cherry pick. But I think we should have a Changelog entry in a follow up PR as this changes the log output which I think quite a few people look at.

PR #5508, opened 2017-11-03
Change `@timestamp` to `timestamp` in monitoring events

kvch:
New event type is introduced to report metrics, because `@timestamp` has to be changed to `timestamp`.
It is tested. However, I hasn't seen the whole pipeline working together, because the other part does not support it yet.

### DIFF
--- a/libbeat/monitoring/report/elasticsearch/client.go
+++ b/libbeat/monitoring/report/elasticsearch/client.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/monitoring/report"
 	esout "github.com/elastic/beats/libbeat/outputs/elasticsearch"
 	"github.com/elastic/beats/libbeat/publisher"
 	"github.com/elastic/beats/libbeat/testing"
@@ -83,8 +84,12 @@ func (c *publishClient) Publish(batch publisher.Batch) error {
 	events := batch.Events()
 	bulk := make([]interface{}, 0, 2*len(events))
 	for _, event := range events {
+		mEvent := report.Event{
+			Timestamp: event.Content.Timestamp,
+			Fields:    event.Content.Fields,
+		}
 		bulk = append(bulk,
-			actMonitoringBeats, event.Content,
+			actMonitoringBeats, mEvent,
 		)
 	}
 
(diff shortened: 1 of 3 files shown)


kvch:
The message format was tested using WireShark and its format is correct.

libbeat/outputs/elasticsearch/enc.go
```diff
@@ -94,6 +95,8 @@ func (b *jsonEncoder) AddRaw(obj interface{}) error {
 		err = b.folder.Fold(event{Timestamp: v.Timestamp, Fields: v.Fields})
 	case *beat.Event:
 		err = b.folder.Fold(event{Timestamp: v.Timestamp, Fields: v.Fields})
+	case report.Event:
+		err = b.folder.Fold(report.Event{Timestamp: v.Timestamp, Fields: v.Fields})
```
urso: this should not be necessary. You unpack and pack the event... the default clause should have done the trick

andrewkroh:
LGTM, but is there some kind of test case we can add for encoding the reporting events?

libbeat/outputs/elasticsearch/enc.go
kvch: I wish there was a unisex facepalm emoji in github. I would use it now. :D

kvch:
@urso i have deleted the unnecessary part
@andrewkroh i have added a few tests to check the encoding. let me know if i should add more.

libbeat/monitoring/report/elasticsearch/client.go
```diff
@@ -83,8 +84,12 @@ func (c *publishClient) Publish(batch publisher.Batch) error {
 	events := batch.Events()
 	bulk := make([]interface{}, 0, 2*len(events))
 	for _, event := range events {
+		mEvent := report.Event{
+			Timestamp: event.Content.Timestamp,
+			Fields:    event.Content.Fields,
+		}
 		bulk = append(bulk,
-			actMonitoringBeats, event.Content,
+			actMonitoringBeats, mEvent,
```
urso: (optional) in order to save some lines you can turn this into:

```
bulk = append(bulk, actMonitoringBeats, report.Event{
  Timestamp: event.Content.Timestamp,
  Fields: event.Content.Fields,
})
```

libbeat/outputs/elasticsearch/enc_test.go
```diff
@@ -0,0 +1,46 @@
...
+	"time"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/elastic/beats/libbeat/beat"
+	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/monitoring/report"
+)
+
+func TestJSONEncoderMarshalBeatEvent(t *testing.T) {
```
andrewkroh: Thanks

PR #5509, opened 2017-11-03
Fix ML jobs setup for dynamic modules

### LINKED ISSUE #5504: Filebeat module doesn't load ML jobs without --modules
When using --modules= the Nginx module's Machine Learning jobs for X-Pack will load. However, users who prefer to use the enable/disable CLI commands should be able to take advantage of this functionality as well. Similarly, users who have enabled the module via the metricbeat.yml should be able to use this also? (I didn't test this method)

- Version: 6.0.0 rc2
- Operating System: MacOS
- Steps to Reproduce: 
./filebeat modules enable nginx
./filebeat setup
### Note that ML jobs don't load.
./filebeat setup --modules=nginx

exekias:
Modules from `modules.d` were ignored by both `setup` command and
`--setup` flag.

Fixes #5504

### DIFF
--- a/filebeat/beater/filebeat.go
+++ b/filebeat/beater/filebeat.go
@@ -4,9 +4,11 @@ import (
 	"flag"
 	"fmt"
 
+	"github.com/joeshaw/multierror"
 	"github.com/pkg/errors"
 
 	"github.com/elastic/beats/libbeat/beat"
+	"github.com/elastic/beats/libbeat/cfgfile"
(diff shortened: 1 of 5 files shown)


tsg:
Thanks for finding a solution! I think the approach is good, but one of the tests is failing.

exekias:
Thanks for noticing, kibana is needed to load with `--setup` flag, I added it to docker-compose and applied some clean up to tests

exekias:
jenkins, retest it please

PR #5520, opened 2017-11-06
Fix duplicate batches in retry queue

urso:
On error in Logstash output sender, failed batches might get enqueued two
times. This can lead to multiple resends and ACKs for the same events.

In filebeat/winlogbeat, waiting for ACK from output, at most one ACK is
required. With potentially multiple ACKs (especially with multiple
consecutive IO errors) a deadlock in the outputs ACK handler can occur.

This PR ensures batches can not be returned to the retry queue via 2 code
paths (remove race between competing workers):
- async output worker does not return events back into retry queue
- async clients are required to always report retrieable errors via
callbacks
- add some more detailed debug logs to the LS output, that can help in
  identifiying ACKed batches still being retried

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -28,7 +28,6 @@ https://github.com/elastic/beats/compare/v5.6.4...5.6[Check the HEAD diff]
 
 *Affecting all Beats*
 
-- Fix race condition in internal logging rotator. {pull}4519[4519]
 - Fix duplicate batches of events in retry queue. {pull}5520[5520]
 
 *Filebeat*
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -28,6 +28,9 @@ https://github.com/elastic/beats/compare/v5.6.4...5.6[Check the HEAD diff]
 
 *Affecting all Beats*
 
+- Fix race condition in internal logging rotator. {pull}4519[4519]
+- Fix duplicate batches of events in retry queue. {pull}5520[5520]
+
 *Filebeat*
 
 *Heartbeat*
--- a/libbeat/outputs/logstash/async.go
+++ b/libbeat/outputs/logstash/async.go
@@ -107,15 +107,7 @@ func (c *asyncClient) AsyncPublishEvents(
 		return nil
 	}
 
-	ref := &msgRef{
-		client:    c,
-		count:     1,
-		batch:     data,
-		batchSize: len(data),
-		win:       c.win,
-		cb:        cb,
-		err:       nil,
-	}
+	ref := newMsgRef(c, data, cb)
 	defer ref.dec()
 
 	for len(data) > 0 {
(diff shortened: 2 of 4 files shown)


libbeat/outputs/logstash/async.go
```diff
@@ -183,7 +175,33 @@ func (r *msgRef) callback(seq uint32, err error) {
...
+		client:    client,
+		count:     1,
+		batch:     data,
+		batchSize: len(data),
+		win:       client.win,
+		cb:        cb,
+		err:       nil,
+	}
+
+	debug("msgref(%p) new: batch=%p, cb=%p", r, &r.batch[0], cb)
```
ruflin: This will create a lot of debug messages. I know that this info is for debugging, but I start to get the feeling we need something in between `info` and `debug`, something like `info --verbose` :-)

ruflin:
This change seems to cause some issues: https://travis-ci.org/elastic/beats/jobs/298049838#L642

libbeat/outputs/logstash/async.go
urso: Yeah, some extra-debug would be nice to have. Normally the debug messages create about 4 new debug messages per batch (2048) events. Not too bad, but super helpful to identify potential issues, should we still face some problems.

libbeat/outputs/mode/lb/async_worker.go
```diff
@@ -111,25 +111,10 @@ func (w *asyncWorker) sendLoop() (done bool) {
...
-		}
-
-		// asynchronously retry to insert message (if attempts left), so worker can not
-		// deadlock on retries channel if client puts multiple failed outstanding
-		// events into the pipeline
-		w.onFail(msg, err)
-	}
-
-	return err
+	return w.client.AsyncPublishEvents(w.handleResults(msg), msg.data)
```
urso: Change in semantics is here. We always require the `w.client` instance to use the callback build via `handleResult` to report success/failure within a batch. This allows the output client to decide on sync or async error reporting. 
The async client as provided by `go-lumber` does require full async reporting, but the Logstash output did some sync reporting as well, leading to duplicates. PR changes to have `go-lumber` only trigger the async reporting, indirectly via `msgRef`.

ph:
@urso I am aware of the possible deadlock, but it could cause also just pure duplicates on retry right? I remember seen a few cases recently concerning duplicates. Do we also have an idea since when this problem occur, It might help debugging future past cases.

urso:
The bug could produce duplicates on retry if `pipelining` is configured (by default it's disabled), but will always result in a deadlock in filebeat -> filebeat needs to be restarted. Number of total batches transfered before the deadlock depends on the `pipelining` setting. Adding load-balancing with multiple endpoints can potentially prolong/hide the deadlock, as the other output workers might still publish events.

CHANGELOG.asciidoc
```diff
@@ -28,6 +28,8 @@ https://github.com/elastic/beats/compare/v5.6.4...5.6[Check the HEAD diff]
 
 *Affecting all Beats*
 
+- Fix duplicate batches of events in retry queue. {pull}5520[5520]
```
ph: I would probably mention deadlock, duplicates can be seen in other scenario.

ruflin:
@urso I merged this one based on all the 👍  The comments by @ph can still be adressed in follow up PR.

PR #5526, opened 2017-11-06
Guarantee ACK list is in order

urso:
Upon ACK from outputs, guarantee the ACK list is always collected in
order. If the first to be scheduled batch is not yet ACKed, return an
empty ACK list and wait for more ACK signals before continue processing.


libbeat/publisher/queue/memqueue/ackloop.go
```diff
@@ -109,9 +114,6 @@ func (l *ackLoop) handleBatchSig() int {
 func (l *ackLoop) collectAcked() chanList {
 	lst := chanList{}
 
-	acks := l.lst.pop()
```
ruflin: Can you quickly comment for my understand why this is not needed anymore?

urso: The `pop` is already done in the for loop. But the for loop also checks the list entry has actually been ACKed. The `pop` operation removed did assume the entry has been ACKed, without checking if that's the case.

urso: Argh... the pop has been here, cause the entry has already been ACKed. The channel is stored in `l.sig` in the run loop.

tsg:
There seems to be a panic in the tests: https://travis-ci.org/elastic/beats/jobs/298272126#L643

PR #5531, opened 2017-11-07
Fix description

robin13:
Description indicated that metricbeat sends only to elasticsearch, but output can be to elasticsearch, logstash, kafka, ... and in future many more.


ruflin:
So far we "sticked" to just naming Elasticsearch (except for Filebeat) as it's our default and "recommended" output. In general I think the descriptions could be improved. Perhaps we could sync the description with the one we have on the website? https://www.elastic.co/products/beats/metricbeat If we update it, we should update all beats.

@dedemorton Any thoughts?

dedemorton:
@ruflin I'm curious...where does the description actually surface for users? IMO, I wouldn't try too hard to sync our descriptions with the website because they change the website frequently...they're always looking for something snappy that will grab someone's attention.

If you want to keep it simple but still alighted with marketing, maybe say: "Metricbeat is a lightweight shipper for metrics." 

Really depends on where the content is displayed because a short description often raises more questions than it answers. :-P

robin13:
> @ruflin I'm curious...where does the description actually surface for users?

In the init script for debian for one.  I was just setting up metricbeat to send to logstash and saw the inappropriate description here.

I'm also happy with `Metricbeat is a lightweight shipper for metrics` - shall I change the PR accordingly @ruflin ?

ruflin:
It also shows up in the README in the beats package (as far as I remember). 

@robin13 @dedemorton The above proposal SGTM.

metricbeat/Makefile
```diff
@@ -3,7 +3,7 @@
 # Name can be overwritten, as Metricbeat is also a library
 BEAT_NAME?=metricbeat
 BEAT_TITLE?=Metricbeat
-BEAT_DESCRIPTION?=Metricbeat sends metrics to Elasticsearch.
+BEAT_DESCRIPTION?=Metricbeat is a lightweight shipper for metrics
```
ruflin: Sorry to be picky, but can you add a dot at it end?

ruflin:
@robin13 Thanks for the fix.

Issue #5524, opened 2017-11-06
Packetbeat panic: Too many events acked


jamiesmith:
A google search found a similar issue on discuss https://discuss.elastic.co/t/packetbeat-panic-too-many-events-acked/102055/2

- Version: packetbeat-6.0.0-rc2-darwin-x86_64
- Operating System: osx 10.13.1
- Steps to Reproduce: does it pretty much every time, some times right away, while other times a few minutes.  I have tried both logstash and elasticsearch outputs.  

`sudo ./packetbeat -e -c packetbeat.yml`

Logs:
```
2017/11/06 19:14:32.960838 beat.go:426: INFO Home path: [/Users/jamie/ELK/packetbeat-6.0.0-rc2-darwin-x86_64] Config path: [/Users/jamie/ELK/packetbeat-6.0.0-rc2-darwin-x86_64] Data path: [/Users/jamie/ELK/packetbeat-6.0.0-rc2-darwin-x86_64/data] Logs path: [/Users/jamie/ELK/packetbeat-6.0.0-rc2-darwin-x86_64/logs]
2017/11/06 19:14:32.960849 metrics.go:23: INFO Metrics logging every 30s
2017/11/06 19:14:32.961000 beat.go:433: INFO Beat UUID: ffd57af7-5cdb-499d-8e29-819beb3fc8b4
2017/11/06 19:14:32.961015 beat.go:192: INFO Setup Beat: packetbeat; Version: 6.0.0-rc2
2017/11/06 19:14:32.964648 client.go:123: INFO Elasticsearch url: http://orion:9200
2017/11/06 19:14:32.965746 module.go:80: INFO Beat name: orion.jamieandamy.com
2017/11/06 19:14:32.965887 procs.go:78: INFO Process matching disabled
2017/11/06 19:14:32.966985 beat.go:260: INFO packetbeat start running.


2017/11/06 19:14:34.520111 client.go:651: INFO Connected to Elasticsearch version 6.0.0-rc2
2017/11/06 19:14:34.522433 load.go:73: INFO Template already exists and will not be overwritten.
[...]

PR #5536, opened 2017-11-07
Cherry-pick #5532 to master: Fix panic if event is dropped by queue

tsg:
Cherry-pick of PR #5532 to master branch. Original message: 

Resolves: #5524 

- Only increase event seq-no if event has been pushed
- Fix invalid ACK list if empty list is given to concat
- Only execute ack handlers if number of events being ACKed > 0
- Add missing debug log when collecting first ACK in ACK list

### DIFF
--- a/libbeat/publisher/queue/memqueue/ackloop.go
+++ b/libbeat/publisher/queue/memqueue/ackloop.go
@@ -53,13 +53,20 @@ func (l *ackLoop) run() {
 			count, events := lst.count()
 			l.lst.concat(&lst)
 
-			// log.Debugf("ackloop: scheduledACKs count=%v events=%v\n", count, events)
+			// log.Debug("ACK List:")
+			// for current := l.lst.head; current != nil; current = current.next {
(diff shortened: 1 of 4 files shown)


libbeat/publisher/queue/memqueue/ackloop.go
```diff
@@ -53,13 +53,20 @@ func (l *ackLoop) run() {
 			count, events := lst.count()
 			l.lst.concat(&lst)
 
-			// log.Debugf("ackloop: scheduledACKs count=%v events=%v\n", count, events)
+			// log.Debug("ACK List:")
```
ruflin: Is this commented out on purpose?

urso: Yes. debug creates quite some noise if not required. There are a many debug statements in this package being commented out and only selectively uncommented when required (when debugging locally).

urso:
Jenkins, test it.

PR #5537, opened 2017-11-07
Remove prospector validation as not needed anymore

ruflin:
Since the prospector are now registered through init, the validation is not needed anymore as in case a prospector type does not exist, and error is returned anyway. The harvester types are still needed because several harvesters use the log harvester.

### DIFF
--- a/filebeat/harvester/util.go
+++ b/filebeat/harvester/util.go
@@ -10,14 +10,6 @@ const (
 	UdpType   = "udp"
 )
 
-// ValidType of valid input types
-var ValidType = map[string]struct{}{
-	StdinType: {},
-	LogType:   {},
-	RedisType: {},
-	UdpType:   {},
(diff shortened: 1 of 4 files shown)


kvch:
Right now the following error is presented if the name of the prospector is invalid:
```
2017/11/08 19:33:26.012726 beat.go:632: CRIT Exiting: Error in initing prospector: Error retrieving factory for prospector 'cica'
Exiting: Error in initing prospector: Error retrieving factory for prospector 'cica'
```

I wonder if a clearer error message should be shown. For example: `"Unknown prospector type"`.

ruflin:
@kvch I pushed https://github.com/elastic/beats/pull/5537/commits/c25a8ddb4d77786d0c408264ce9a40f4f05e2cb2 with a better error message.

PR #5539, opened 2017-11-08
Remove redundant log prefix for auditbeat module (#5533)

### LINKED ISSUE #5533: Debug selector logged twice in Auditbeat modules
Debug log statements from Auditbeat include the selector name twice. The debug statements (and only the debug statements) should no longer include `logPrefix` because it is automatically added (see #5380) by `logp` for all `logp.Debug()` calls.

`2017-11-07T18:46:08Z DBG  [audit.file] [audit.file] Received fsnotify event: path=/etc/ld.so.cache action=CREATE`

https://github.com/elastic/beats/blob/81dfb731024cf34c9e112aa2e1b0c311036fb605/auditbeat/module/audit/file/metricset.go#L205-L206

liketic:
Closes #5533


andrewkroh:
Thanks for picking this issue up. Two things,

1. Only the debugf and logp.Debug calls are affected. The Info/Warn/Error don’t need to be changed because logp does not include the selector.
1. There are some debug statements in the audit/kernel metricset that also need updated.

liketic:
@andrewkroh Thanks! I updated the PR. Please review again.

andrewkroh:
Jenkins, test it please!

andrewkroh:
jenkins, test it

PR #5338, opened 2017-10-05
Generate fields.yml of Filebeat modules

kvch:
I have reimplemented the Python PoC in Golang. Most of the output problems are solved. However, it requires gopkg.in/yaml.v2 to write YAML files.

Example command to generate log fileset of postgresql module:
```
$ go run generate_fields_yml.go -module postgresql -fileset log
```
Output:
```
- name: log
  description: Please add description
  example: Please add example
  type: group
  fields:
  - name: timestamp
    description: Please add description
    example: Please add example
  - name: timezone
    description: Please add description
    example: Please add example
    type: keyword
  - name: thread_id
    description: Please add description
    example: Please add example
    type: long
  - name: user
    description: Please add description
    example: Please add example
    type: keyword
  - name: database
    description: Please add description
    example: Please add example
    type: keyword
  - name: level
    description: Please add description
    example: Please add example
    type: keyword
  - name: duration
    description: Please add description
    example: Please add example
    type: long
  - name: query
    description: Please add description
    example: Please add example
    type: text
  - name: message
    description: Please add description
    example: Please add example
    type: text
```

TODO
- [x] implement remove, rename processors
- [x] make adding description, example fields optional
- [ ] add it to FB module dev guide

Questions
- Should it be a subcommand of `filebeat`?

PR #5544, opened 2017-11-08
Module generator tools for Filebeat

kvch:
Previously, creating filesets required calling `make` and used Python scripts.
Now a `create_fileset.py` is ported to Golang and `create-fields` is added,  too.

Both scripts are available as Make targets:
make create-fileset MODULE={module} FILESET={fileset}
make create-fields MODULE={module} FILESET={fileset}

## Process of creating a new FB module: postgresql/log
1. run `make create-fileset MODULE=postgresql FILESET=log`
2. edit pipeline.json
3. run `make create-fields MODULE=postgresql FILESET=log`
4. edit documentation fields
5. enjoy

I am closing my previous PR #5338 which includes only partial patches.


filebeat/cmd/root.go
```diff
@@ -25,5 +25,8 @@ func init() {
 	RootCmd.PersistentFlags().AddGoFlag(flag.CommandLine.Lookup("M"))
 	RootCmd.TestCmd.Flags().AddGoFlag(flag.CommandLine.Lookup("modules"))
 	RootCmd.SetupCmd.Flags().AddGoFlag(flag.CommandLine.Lookup("modules"))
-	RootCmd.AddCommand(cmd.GenModulesCmd(Name, "", buildModulesManager))
+
+	modulesCmd := cmd.GenModulesCmd(Name, "", buildModulesManager)
+	modulesCmd.AddCommand(genGenerateCmd())
```
ruflin: I would probably move this out of the `module` command do directly have `filebeat generate ...`. Like this we can use it to generate things which are not necessarly attached to module in the future.

filebeat/fileset/generate_fields_yml.go
```diff
@@ -0,0 +1,323 @@
+package fileset
```
ruflin: I would not mix the fileset package with the generator code. WDYT about `filebeat/generate/` package instead?

filebeat/fileset/generate_fields_yml.go
```diff
@@ -0,0 +1,323 @@
...
+	}
+
+	return &processors{
+		patterns: patterns,
+		remove:   rmFields,
+		rename:   mvFields,
+	}, nil
+}
+
+func getFieldByName(f []*fieldYml, name string) *fieldYml {
```
ruflin: I wonder if we could extract this code later in `libbeat` itself because I think generating the `fields.yml` file is something we should have in the core.

kvch: This generator generates the file based on the processors in pipelines. So the `readPipeline` and `toFieldsYml` partially has to be generalized to utilize it e.g in metricbeat. It is exactly my plan to have similar generators for both beats. However, it would be a too big change to implement everything in one step.

filebeat/cmd/root.go
kvch: I think the current functionality is good under `modules`. But as we move away from `make` e.g. eliminating `make update`, I can totally see all generators under `generate` where it also includes generating of index templates, index patterns, etc.

From users point of view you are right about moving it to `generate`. So as we move more functionality here, users would not be confused if `modules generate` disappears and `generate` shows up instead.

ruflin: We should decide now in which package it belongs in the long run as moving it later would be a breaking change.

filebeat/fileset/generate_fields_yml.go
ruflin: Definitively. Just thought I mention it as it popped up in my head when I saw it. +1 on doing it later.

filebeat/cmd/root.go
kvch: Agreed

kvch:
I moved these two generators under `scripts`. The parametrization hasn't changed from the last commits.

Usage of scripts:
```
$ go run scripts/generator/fileset/main.go -module kafka -fileset log
New module was generated. After setting up Grok pattern in pipeline.json,
please generate fields.yml
```
```
$ go run scripts/generator/fields/main.go -module kafka -fileset log
Fields.yml generated for kafka/log
```

kvch:
* updated `make create-fileset`
* added `make create-fields`
* updated module developer documentation
* added Changelog entry
* squashed and rebased commits, so #6037 is included

CHANGELOG.asciidoc
```diff
@@ -72,6 +72,7 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
 - Fix `recursive_globe.enabled` option. {pull}5443[5443]
 - Change pipeline delimiter to `{<` and `>}`. {pull}5702[5702]
 - Fix variable name for `convert_timezone` in the system module. {pull}5936[5936]
+- Add module generator tools for Filebeat. {pull}5544[5544]
```
ruflin: Probably we can skip the changelog here as we normally don't to changelog entry for dev changes.

filebeat/Makefile
```diff
@@ -64,6 +64,10 @@ collect: fields kibana modules configs collect-docs imports
 
 # Creates a new fileset. Requires the params MODULE and FILESET
 .PHONY: create-fileset
-create-fileset: python-env
-	@${PYTHON_ENV}/bin/python ${ES_BEATS}/filebeat/scripts/create_fileset.py --path=$(PWD) --es_beats=$(ES_BEATS) --module=$(MODULE) --fileset=$(FILESET)
+create-fileset:
+	@go run ${ES_BEATS}/filebeat/scripts/generator/fileset/main.go --path=$(PWD) --beats_path=$(BEAT_GOPATH)/src/$(BEAT_PATH) --module=$(MODULE) --fileset=$(FILESET)
 
+# Creates a fields.yml based on a pipeline.json file. Requires the params MODULE and FILESET
+.PHONY: generate-fields
```
ruflin: This should be `create-fields`

kvch:
@ruflin done. thanks for your review.

filebeat/scripts/generator/fields/main.go
```diff
@@ -0,0 +1,340 @@
...
+
+	if *fileset == "" {
+		fmt.Println("Missing parameter: fileset")
+		os.Exit(1)
+	}
+
+	p, err := readPipeline(*beatsPath, *module, *fileset)
+	if err != nil {
+		fmt.Printf("Cannot read pipeline.yml of fileset: %v\n", err)
+		os.Exit(2)
```
ruflin: I stumbled a bit over the part that you keep increasing the Exit codes, but as long as it's not 0, should work.

ruflin:
Merged as only having this in our code based will make sure we start using it. I didn't test it locally yet but code looks good.

ruflin:
@kvch Thanks for tackling this. Having this code in is a huge step forward to make it possible one day that also users and not only devs can create FB modules.

PR #5545, opened 2017-11-08
New metric: CPU usage

kvch:
* Add both percentage and normalized percentage of CPU.
* Move Round function to common.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,10 +63,11 @@ jobs:
       env: TARGETS="-C libbeat testsuite"
       go: $GO_VERSION
       stage: test
-    - os: linux
-      env: TARGETS="-C libbeat crosscompile"
-      go: $GO_VERSION
-      stage: test
+      # Skip tests: see https://github.com/elastic/beats/issues/5984
+      #    - os: linux
+      #      env: TARGETS="-C libbeat crosscompile"
+      #      go: $GO_VERSION
+      #      stage: test
     - os: linux
       env: TARGETS="-C libbeat stress-tests"
       go: $GO_VERSION
@@ -81,10 +82,11 @@ jobs:
       env: TARGETS="TEST_ENVIRONMENT=0 -C metricbeat testsuite"
       go: $GO_VERSION
       stage: test
-    - os: linux
-      env: TARGETS="-C metricbeat crosscompile"
-      go: $GO_VERSION
-      stage: test
+      # Skip tests: see https://github.com/elastic/beats/issues/5984
+      #    - os: linux
+      #      env: TARGETS="-C metricbeat crosscompile"
+      #      go: $GO_VERSION
+      #      stage: test
 
     # Packetbeat
     - os: linux
--- a/libbeat/cmd/instance/metrics.go
+++ b/libbeat/cmd/instance/metrics.go
@@ -1,19 +1,46 @@
 package instance
 
 import (
+	"fmt"
 	"runtime"
 	"time"
 
+	"github.com/elastic/beats/libbeat/logp"
+	"github.com/elastic/beats/libbeat/metric/system/cpu"
+	"github.com/elastic/beats/libbeat/metric/system/process"
 	"github.com/elastic/beats/libbeat/monitoring"
+	"github.com/elastic/beats/libbeat/monitoring/report/log"
 )
 
-var startTime = time.Now()
+var (
+	cpuMonitor       *cpu.Monitor
+	beatProcessStats *process.Stats
+)
 
 func init() {
-	metrics := monitoring.Default.NewRegistry("beat")
+	beatMetrics := monitoring.Default.NewRegistry("beat")
+	monitoring.NewFunc(beatMetrics, "memstats", reportMemStats, monitoring.Report)
+	monitoring.NewFunc(beatMetrics, "cpu", reportBeatCPU, monitoring.Report)
+	monitoring.NewFunc(beatMetrics, "info", reportInfo, monitoring.Report)
 
-	monitoring.NewFunc(metrics, "memstats", reportMemStats, monitoring.Report)
-	monitoring.NewFunc(metrics, "info", reportInfo, monitoring.Report)
+	systemMetrics := monitoring.Default.NewRegistry("system")
+	monitoring.NewFunc(systemMetrics, "load", reportSystemLoadAverage, monitoring.Report)
+	monitoring.NewFunc(systemMetrics, "cpu", reportSystemCPUUsage, monitoring.Report)
+}
+
+func setupMetrics(name string) error {
+	cpuMonitor = new(cpu.Monitor)
+
+	beatProcessStats = &process.Stats{
+		Procs:        []string{name},
+		EnvWhitelist: nil,
+		CpuTicks:     false,
+		CacheCmdLine: true,
+		IncludeTop:   process.IncludeTopConfig{},
+	}
+	err := beatProcessStats.Init()
+
+	return err
 }
 
 func reportMemStats(m monitoring.Mode, V monitoring.Visitor) {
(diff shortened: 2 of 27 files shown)


libbeat/cmd/instance/metrics.go
```diff
@@ -36,3 +55,38 @@ func reportInfo(_ monitoring.Mode, V monitoring.Visitor) {
...
+func getCPUUsage() (float64, float64) {
+	pid := os.Getpid()
+
+	sample := struct {
+		time      time.Time
+		procTimes sigar.ProcTime
+	}{
+		time.Now(),
+		sigar.ProcTime{},
+	}
```
ph: We are using the same anonymized struct in [two different places in the code](https://github.com/elastic/beats/pull/5545/files#diff-b248fa29eddb03765933dc1227905d4bR13) We can make a concrete struct maybe called `CpuSample` and reuse it or we could also encapsulate the creation of the sample into his own method and return the sample or error.

libbeat/cmd/instance/metrics.go
```diff
@@ -36,3 +55,38 @@ func reportInfo(_ monitoring.Mode, V monitoring.Visitor) {
...
+	if err := sample.procTimes.Get(pid); err != nil {
+		return 0, 0
+	}
+
+	dTime := sample.time.Sub(lastSample.time)
+	dMilli := dTime / time.Millisecond
+	dCPU := int64(sample.procTimes.Total - lastSample.procTimes.Total)
+
+	usage := float64(dCPU) / float64(dMilli)
+	normalized := usage / float64(numCores)
```
ph: I think I might be missing some context here concerning the nature of the recording.
This will give us a rate of cpu usage?

libbeat/common/math.go
```diff
@@ -0,0 +1,23 @@
...
+	digit := pow * val
+	_, div := math.Modf(digit)
+	if div >= 0.5 {
+		round = math.Ceil(digit)
+	} else {
+		round = math.Floor(digit)
+	}
+	newVal = round / pow
+	return
+}
```
ph: 🐼  I wish it was in the [standard library](https://github.com/golang/go/issues/20100), I think we should make this method takes an argument for the precision and create a method in the metrics that use the `maxDecimalPlaces`, since it's in the common file I don't think we should make arbitrary decision on the precision. 

But It can be up to discussion

libbeat/cmd/instance/metrics.go
kvch: Yes, it gives the total CPU usage of a beat between the last two samples.

I used the method from `metricbeat/modules/system/process/helper`. I thought about extracting the whole `Process` and `ProcStats` and moving it `libbeat`. However, it seemed unnecessary to move such a big part of `metricbeat` into `libbeat` for only this minor calculation. But I am open to reconsider moving it.

libbeat/common/math.go
kvch: I implemented it

metricbeat/module/system/util.go
```diff
@@ -3,16 +3,12 @@
 package system
 
 import (
-	"math"
 	"runtime"
 
 	sigar "github.com/elastic/gosigar"
+	"github.com/elastic/libbeat/common"
```
tsg: I assume this should be `github.com/elastic/beats/libbeat/common`

metricbeat/module/system/util.go
```diff
@@ -197,36 +193,18 @@ type LoadAverages struct {
 // 0 to NumCPU.
 func (m *LoadMetrics) Averages() LoadAverages {
 	return LoadAverages{
-		OneMinute:     Round(m.sample.One),
-		FiveMinute:    Round(m.sample.Five),
-		FifteenMinute: Round(m.sample.Fifteen),
+		OneMinute:     common.Round(m.sample.One),
```
tsg: These need updating to the new `Round()` signature.

tsg:
The new metrics (`beat.cpu.usage`, `beat.cpu.usage.normalized`) should be added to the list [here](https://github.com/elastic/beats/blob/master/libbeat/logp/metrics.go#L64-L74).

Btw, I think needs to be done also for the `beat.info.uptime.ms` metric from the other PR.

tsg:
@ruflin I wonder if we should replace `beat` with `agent` as a namespace for these before they start being used. I'm thinking in the context of the common schema, ofc.

kvch:
I added `beat.cpu.*` to gauges list. I also added `beat.info.uptime.ms` to the list in a separate commit. Should I cherry-pick it to a new branch and open a new PR?

libbeat/cmd/instance/metrics.go
ph: All fine then! thanks @kvch !

libbeat/cmd/instance/metrics.go
```diff
@@ -36,3 +59,35 @@ func reportInfo(_ monitoring.Mode, V monitoring.Visitor) {
...
+		procTimes: sigar.ProcTime{},
+	}
+
+	if err := sample.procTimes.Get(pid); err != nil {
+		return 0, 0
+	}
+
+	dTime := sample.time.Sub(lastSample.time)
+	dMilli := dTime / time.Millisecond
+	dCPU := int64(sample.procTimes.Total - lastSample.procTimes.Total)
```
ruflin: If we would only report `sample.procTimes.Total` with a timestamp + # of cores, we could leave the calculation of 1, 5 and 15 minutes average to Kibana? Like this it would be up to KB to decide which averages it wants to calculate. Perhaps I miss something?

libbeat/logp/metrics.go
```diff
@@ -71,6 +71,9 @@ var gauges = map[string]bool{
 	"beat.memstats.memory_total":     true,
 	"beat.memstats.memory_alloc":     true,
 	"beat.memstats.gc_next":          true,
+	"beat.cpu.usage":                 true,
```
ruflin: @tsg At one point we should review these metrics to align them with `metricbeat` so it would be `system.cpu.*` .

ruflin:
@tsg Yes, we should review the naming of all metrics. So my comment inline.

libbeat/cmd/instance/metrics.go
kvch: I like your idea. @tsullivan WDYT?

tsullivan: Not sure I'm following... is this for system load average?

cc @pickypg

kvch: Right now it returns the number of CPU usage between the last two sampling times. Let's say that the sampling time is 1 minute. It would return the following metrics in case of constant usage:
```
beat.cpu.usage=0.1   // at 12:00
beat.cpu.usage=0.1   // at 12:01
beat.cpu.usage=0.1   // at 12:02
```

@ruflin's idea is to return a timestamp and the total CPU usage since the Beat has been started. It would look like this:
```
timestamp=12:00 beat.cpu.usage=0.1
timestamp=12:01 beat.cpu.usage=0.2
timestamp=12:02 beat.cpu.usage=0.3 
```
@ruflin please, correct me if I misunderstood your idea.

ruflin: Exactly. So on the UI side there would be fully flexibility if it should be 1,5,15 or 1,10,30 minute average and we report only one value each time.

libbeat/cmd/instance/metrics.go
```diff
@@ -1,16 +1,39 @@
...
+	numCores   = runtime.NumCPU()
+	lastSample = cpuSample{
+		time:      time.Now(),
+		procTimes: sigar.ProcTime{},
+	}
 )
 
 func init() {
+	pid := os.Getpid()
+	err := lastSample.procTimes.Get(pid)
```
andrewkroh: `Get()` is not implemented for all platforms so by adding this you will be limiting the platforms for which Beats can be compiled. This should probably be stubbed in gosigar, but that is currently not the case so you'll need to only use it on the platforms where it's available (linux, windows, openbsd, darwin).

libbeat/cmd/instance/metrics.go
```diff
@@ -1,16 +1,39 @@
...
+	if err != nil {
+		panic(err)
```
andrewkroh: I think you should handle the error rather than `panic`'ing. I would reserve the use of panics to cases that are completely unrecoverable or are programmer errors that are covered by unit tests (like hitting the `default` case in a switch statement that is supposed to cover all enum values). Panics on init can be really hard to debug especially on Windows.

libbeat/cmd/instance/metrics.go
```diff
@@ -36,3 +59,35 @@ func reportInfo(_ monitoring.Mode, V monitoring.Visitor) {
...
+func reportCPU(_ monitoring.Mode, V monitoring.Visitor) {
+	V.OnRegistryStart()
+	defer V.OnRegistryFinished()
+
+	cpuUsage, normalizedCPU := getCPUUsage()
+	monitoring.ReportFloat(V, "usage", cpuUsage)
+	monitoring.ReportFloat(V, "usage.normalized", normalizedCPU)
+}
+
+func getCPUUsage() (float64, float64) {
```
andrewkroh: Suggest `getProcessCPUUsage` to make it clear that it's CPU usage for this process only vs overall CPU usage. It would be nice to have a godoc statement explaining what the return values are since there are two of them (and what their ranges are like [0, 1] or [0, 100]).

libbeat/cmd/instance/metrics.go
andrewkroh: Can this value be negative? (like if the counter rolls over)

libbeat/cmd/instance/metrics.go
```diff
@@ -36,3 +59,35 @@ func reportInfo(_ monitoring.Mode, V monitoring.Visitor) {
...
+	pid := os.Getpid()
+
+	sample := cpuSample{
+		time:      time.Now(),
+		procTimes: sigar.ProcTime{},
+	}
+
+	if err := sample.procTimes.Get(pid); err != nil {
```
andrewkroh: I'd prefer to return the error and have the caller handle it.

libbeat/cmd/instance/metrics.go
```diff
@@ -1,16 +1,39 @@
...
+	pid := os.Getpid()
+	err := lastSample.procTimes.Get(pid)
+	if err != nil {
+		panic(err)
+	}
+
 	metrics := monitoring.Default.NewRegistry("beat")
 
 	monitoring.NewFunc(metrics, "memstats", reportMemStats, monitoring.Report)
+	monitoring.NewFunc(metrics, "cpu", reportCPU, monitoring.Report)
```
andrewkroh: If the Beat is running on a platform where we have no support for getting process CPU data maybe we should not register this metric?

@tsullivan I assume in the UI it would be best to tell the user that we don't have CPU usage data rather than showing 0. Would not sending the metric be the way to go?

libbeat/cmd/instance/metrics.go
kvch: A potential problem came to my mind. It might be confusing to users to see that the internal metrics are reported differently from metricbeat's format. For the sake of consistency, I would stick with the current solution. (Even if it's not as flexible as your proposal is.)

pickypg: So my only concern with this approach is that it's different from what we're currently doing.

In the short term, can we have both? The other parts of the stack also report it in the not-rolled-up format, so we it may be confusing to present a different, albeit better approach to showing the utilization. However, longer term, I would love to see us move toward the more flexible approach, which is only possible if we already have the data (and it seems cheap to have both).

ruflin: Having both SGTM

kvch:
I opened a PR in gosigar, so `crosscompile` will be fixed: https://github.com/elastic/gosigar/pull/83

kvch:
Existing metricbeat code which collected info on the processes and CPU usage of the host machine was moved to `libbeat/metrics/system` and `libbeat/metrics/process`. The name `metrics` is not good enough, but right now I don't have a better name. Please, propose something. :)

New counters are added:
* `beat.cpu.usage`
* `beat.cpu.usage.normalized`
* `beat.cpu.usage.since_start`
* `host.load_averages.1m` 
* `host.load_averages.5m` 
* `host.load_averages.15m` 
* `host.load_averages.1m.norm` 
* `host.load_averages.5m.norm` 
* `host.load_averages.15m.norm`
* `host.cpu.usage`
* `host.cpu.usage.normalized`

@ruflin I used `host` instead of `system` so it does not collide with metricbeat's metrics. However, `host.cpu.usage` is the same as `system.cpu.total.pct`. WDYT?

This PR requires a new gosigar release. As long as it is not available in this repo, tests will fail.

metricbeat/module/system/util.go
```diff
@@ -3,16 +3,12 @@
...
-	"math"
 	"runtime"
 
+	"github.com/elastic/beats/libbeat/common"
 	sigar "github.com/elastic/gosigar"
 )
 
-// maxDecimalPlaces is the maximum number of decimal places that the Round
-// function return.
-const maxDecimalPlaces = 4
```
ruflin: Why did you remove this constant and hardcoded it instead?

ruflin:
For the naming of the fields: I would suggest to keep it identical to what we have in metricbeat. I think having the exact same name is a feature. Or would this cause any issues? See my old comment here on this: https://github.com/elastic/beats/pull/5545/files#r150119927

For the package naming I would follow the scheme we have from metricbeat too. So it would be `libbeat/metric/system/process/process.go`. I like to have the top package `metric` but inside I think these metrics are about the system (there could be others) so I would group them together.

libbeat/metrics/process/process.go
```diff
@@ -258,6 +263,7 @@ func (procStats *ProcStats) getProcessEvent(process *Process) common.MapStr {
 			"norm": common.MapStr{
 				"pct": process.cpuTotalPctNorm,
 			},
+			"total_since_start": process.cpuSinceStart,
```
ruflin: * Is this a change that would also show up in Metricbeat?
* Cam we call this `total.value`? As far as I can see this is the main value we need to do all the calculations for the other values.
* Do we report NumCPU somewhere? If not, I think we should add it as it is required to make all calculations on the ES / KB side if needed.

libbeat/logp/metrics.go
```diff
@@ -71,6 +71,18 @@ var gauges = map[string]bool{
...
+	"beat.cpu.usage.normalized":      true,
+	"beat.cpu.usage.since_start":     true,
+	"beat.info.uptime.ms":            true,
+	"host.cpu.usage":                 true,
+	"host.cpu.usage.normalized":      true,
+	"host.load_average.1m":           true,
```
ruflin: Let's follow this scheme here: https://github.com/elastic/beats/blob/master/metricbeat/module/system/load/_meta/data.json#L12

metricbeat/module/system/util.go
kvch: It was requested by @ph to make the number of decimal places configurable. So it was not needed anymore. But I can see having a new exported constant named `DefaultDecimalPlaces` or something similar. It would be cleaner.

libbeat/metrics/process/process.go
kvch: * Yes, it shows up in metricbeat. I need to add documentation for that new field.
* Yes, I will rename it.
* I will check it. If not, I will add it to the event.

ruflin:
Changes LGTM. I think the tests now fail because `total.value` must be documented in metricbeat as this field now also shows up there.

metricbeat/module/system/cpu/_meta/fields.yml
```diff
@@ -129,6 +129,13 @@
         The percentage of CPU time spent in non-idle state.
 
 
+    # Total
```
ruflin: This is also needed in process as it is added for each process too.

ruflin:
@kvch Please ping me when an other round of review is needed here. Let's get this in :-)

kvch:
@ruflin This PR is ready for a new review.
I removed the `crosscompile` jobs from Travis and opened a meta issue: https://github.com/elastic/beats/issues/5984

.travis.yml
```diff
@@ -63,10 +63,10 @@ jobs:
       env: TARGETS="-C libbeat testsuite"
       go: $GO_VERSION
       stage: test
-    - os: linux
-      env: TARGETS="-C libbeat crosscompile"
-      go: $GO_VERSION
-      stage: test
+      #    - os: linux
```
tsg: Maybe leave a comment here that links to https://github.com/elastic/beats/issues/5984

PR #5549, opened 2017-11-10
Test new Kafka setup for more stable builds

ruflin:
* Use Docker image from Logstash team. This allows us to specify the exact version on Kafka which is tested and be in control of the image.
* Use healthcheck which creates a random topic to make sure broker is available
* Change config options in golang and python tests to make sure sending events is retried
* Fix version in python to a specific kafka version which is in sync with the docker image

### DIFF
--- a/metricbeat/module/kafka/_meta/Dockerfile
+++ b/metricbeat/module/kafka/_meta/Dockerfile
@@ -1,9 +1,25 @@
-# TODO: No tags currently exist for this image. Tags should be used whever possible
-# as otherwise builds are not identical over time.
-FROM spotify/kafka
+FROM debian:stretch
 
-HEALTHCHECK --interval=1s --retries=90 CMD /opt/kafka_*/bin/kafka-topics.sh --zookeeper localhost:2181 --list
+ENV KAFKA_HOME /kafka
+# The advertised host is kafka. This means it will not work if container is started locally and connected from localhost to it
+ENV KAFKA_ADVERTISED_HOST kafka
+ENV KAFKA_LOGS_DIR="/kafka-logs"
+ENV KAFKA_VERSION 0.10.2.1
+ENV _JAVA_OPTIONS "-Djava.net.preferIPv4Stack=true"
+ENV TERM=linux
 
-EXPOSE 2181 9092
+RUN apt-get update && apt-get install -y curl openjdk-8-jre-headless netcat
 
-ENV ADVERTISED_HOST kafka
+RUN mkdir -p ${KAFKA_LOGS_DIR} && mkdir -p ${KAFKA_HOME} && curl -s -o $INSTALL_DIR/kafka.tgz \
+    "http://ftp.wayne.edu/apache/kafka/${KAFKA_VERSION}/kafka_2.11-${KAFKA_VERSION}.tgz" && \
+    tar xzf ${INSTALL_DIR}/kafka.tgz -C ${KAFKA_HOME} --strip-components 1
+
+ADD run.sh /run.sh
+
+EXPOSE 9092
+EXPOSE 2181
+
+# Healtcheck creates an empty topic foo. As soon as a topic is created, it assumes broke is available
+HEALTHCHECK --interval=1s --retries=90 CMD ${KAFKA_HOME}/bin/kafka-topics.sh --zookeeper=127.0.0.1:2181 --create --partitions 1 --topic "foo" --replication-factor 1
+
+ENTRYPOINT ["/run.sh"]
--- a/metricbeat/module/kafka/_meta/run.sh
+++ b/metricbeat/module/kafka/_meta/run.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+
+wait_for_port() {
+    count=20
+    port=$1
+    while ! nc -z localhost $port && [[ $count -ne 0 ]]; do
+        count=$(( $count - 1 ))
+        [[ $count -eq 0 ]] && return 1
+        sleep 0.5
+    done
+    # just in case, one more time
+    nc -z localhost $port
+}
+
+echo "Starting ZooKeeper"
+${KAFKA_HOME}/bin/zookeeper-server-start.sh ${KAFKA_HOME}/config/zookeeper.properties &
+wait_for_port 2181
+
+echo "Starting Kafka broker"
+mkdir -p ${KAFKA_LOGS_DIR}
+${KAFKA_HOME}/bin/kafka-server-start.sh ${KAFKA_HOME}/config/server.properties \
+    --override delete.topic.enable=true --override advertised.host.name=${KAFKA_ADVERTISED_HOST} \
+    --override listeners=PLAINTEXT://0.0.0.0:9092 \
+    --override logs.dir=${KAFKA_LOGS_DIR} --override log.flush.interval.ms=200 &
+
+wait_for_port 9092
+
+echo "Kafka load status code $?"
+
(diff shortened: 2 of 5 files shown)


metricbeat/module/kafka/_meta/run.sh
```diff
@@ -0,0 +1,35 @@
...
+}
+
+echo "Starting ZooKeeper"
+${KAFKA_HOME}/bin/zookeeper-server-start.sh ${KAFKA_HOME}/config/zookeeper.properties &
+wait_for_port 2181
+
+echo "Starting Kafka broker"
+mkdir -p ${KAFKA_LOGS_DIR}
+${KAFKA_HOME}/bin/kafka-server-start.sh ${KAFKA_HOME}/config/server.properties \
+    --override delete.topic.enable=true --override advertised.host.name=kafka \
```
ruflin: @urso @original-brownbear The problem here with setting `avertised.host.name` to `kafka` is that this "should" work when also the kafka test is running inside an other docker container. But if the kafka container is only started locally it should be `localhost`. I tried to figure out if there is an option ot listen on both but was not successful. Do you know if this is possible.

ruflin:
jenkins, retest it

ruflin:
Interestingly now, running only docker-compose for Kafka makes the tests pass. As with the new implementation from @exekias also with all environments built, I would assume only kafka is run so no other containers should impact it?

metricbeat/module/kafka/_meta/run.sh
original-brownbear: @ruflin I don't think you can advertise two hosts. 
I assume you want this to work with Docker on Mac too if you're talking about `localhost`? This will be tricky to without binding to `localhost` and exposing the ports on the host network.

Maybe you should simply add an environment variable and `if` to this script when setting advertised hosts.

Then for local runs you can advertise on `localhost` and for the docker network you can use the `eth0` IP (this will also work if the Docker daemon runs locally on a Linux host, but not with Docker on Mac) via:

```sh
root@48b797207577:/# HOSTIP=`ip -4 addr show scope global dev eth0 | grep inet | awk '{print \$2}' | cut -d / -f 1`
root@48b797207577:/# echo $HOSTIP
172.17.0.2
```

ruflin:
jenkins, retest it

metricbeat/module/kafka/_meta/run.sh
ruflin: @original-brownbear Thanks for the input. Yes, that was my idea. But I'm already more then happy if it works without being flaky on CI. My hope was that the new Docker image would help but the tests are still flaky and we get errors like `NoBroker available`. It could be that our system is just overloaded. Did you see similar issues on the LS side?

original-brownbear: @ruflin hmm no we were pretty stable with Docker on Travis. Maybe you need to wait longer  for the Broker to come up (or have a more stable way of detecting it's up)?
Did you check if the Broker maybe dies for some reason?

ruflin: The problem I have at the moment is that locally it works and on CI it doesn't (most of the times). So I assume it is one of the above problems you mentioned, unfortunately I don't know which one. 

As I would think that the health checks are good enough to wait for the broker to be available, I'm thinking the second could be the case. Do you have any idea on why the broker could die? Or ideas on how to better detect if the broker is started. Unfortunately my experience with running Kafka is pretty limited, so any help is appreciated :-)

metricbeat/module/kafka/_meta/Dockerfile
```diff
@@ -1,9 +1,24 @@
...
+    "http://ftp.wayne.edu/apache/kafka/${KAFKA_VERSION}/kafka_2.11-${KAFKA_VERSION}.tgz" && \
+    tar xzf ${INSTALL_DIR}/kafka.tgz -C ${KAFKA_HOME} --strip-components 1
+
+ADD run.sh /run.sh
+
+EXPOSE 9092
+EXPOSE 2181
+
+#HEALTHCHECK --interval=1s --retries=90 CMD ${KAFKA_HOME}/bin/kafka-topics.sh --create --partitions 1 --replication-factor 1 --topic ${KAFKA_TOPIC} --zookeeper 127.0.0.1:2181
+HEALTHCHECK --interval=1s --retries=90 CMD ${KAFKA_HOME}/bin/kafka-topics.sh --zookeeper 127.0.0.1:2181 --list
```
original-brownbear: @ruflin this is the problem imo, it's not a valid `HEALTHCHECK`:

If I only have a running (and empty) Zookeeper, it still doesn't error, just tried it locally:

```sh
➜  kafka_2.12-0.11.0.0 bin/kafka-topics.sh --zookeeper=127.0.0.1:2181 --list 
[2017-11-14 11:35:07,659] INFO Accepted socket connection from /127.0.0.1:58444 (org.apache.zookeeper.server.NIOServerCnxnFactory)
[2017-11-14 11:35:07,664] INFO Client attempting to establish new session at /127.0.0.1:58444 (org.apache.zookeeper.server.ZooKeeperServer)
[2017-11-14 11:35:07,671] INFO Established session 0x15fba1822250002 with negotiated timeout 30000 for client /127.0.0.1:58444 (org.apache.zookeeper.server.ZooKeeperServer)
[2017-11-14 11:35:07,687] INFO Processed session termination for sessionid: 0x15fba1822250002 (org.apache.zookeeper.server.PrepRequestProcessor)
[2017-11-14 11:35:07,688] INFO Closed socket connection for client /127.0.0.1:58444 which had sessionid 0x15fba1822250002 (org.apache.zookeeper.server.NIOServerCnxn)
➜  kafka_2.12-0.11.0.0 echo $?                                               
0
```

Creating some random topic on the other hand is a valid check, it errors out when the broker isn't up:

```sh
➜  kafka_2.12-0.11.0.0 bin/kafka-topics.sh --zookeeper=127.0.0.1:2181 --create --partitions 1 --topic "foo" --replication-factor 1
[2017-11-14 11:39:20,379] INFO Accepted socket connection from /127.0.0.1:58468 (org.apache.zookeeper.server.NIOServerCnxnFactory)
[2017-11-14 11:39:20,384] INFO Client attempting to establish new session at /127.0.0.1:58468 (org.apache.zookeeper.server.ZooKeeperServer)
[2017-11-14 11:39:20,391] INFO Established session 0x15fba1822250004 with negotiated timeout 30000 for client /127.0.0.1:58468 (org.apache.zookeeper.server.ZooKeeperServer)
Error while executing topic command : replication factor: 1 larger than available brokers: 0
[2017-11-14 11:39:20,443] ERROR org.apache.kafka.common.errors.InvalidReplicationFactorException: replication factor: 1 larger than available brokers: 0
 (kafka.admin.TopicCommand$)
[2017-11-14 11:39:20,444] INFO Processed session termination for sessionid: 0x15fba1822250004 (org.apache.zookeeper.server.PrepRequestProcessor)
[2017-11-14 11:39:20,445] INFO Closed socket connection for client /127.0.0.1:58468 which had sessionid 0x15fba1822250004 (org.apache.zookeeper.server.NIOServerCnxn)
➜  kafka_2.12-0.11.0.0 echo $?
1
```

ruflin: Thanks for checking. Will try that out.

ruflin:
Seems like finally tests are green.

@original-brownbear Thanks a lot for your help.
@urso We should probably switch our output test image to this one too?

PR #5552, opened 2017-11-10
Fix uptime value

kvch:
Previously, uptime was duplicated due to a wrong formula.
`int64(delta.Seconds())*1000 + int64(delta.Nanoseconds()/1000000)` leads to duplicated value as it adds both converted value from seconds and nanoseconds.

Now simply the time in seconds is converted into milliseconds.
@urso WDYT?

### DIFF
--- a/libbeat/cmd/instance/metrics.go
+++ b/libbeat/cmd/instance/metrics.go
@@ -33,6 +33,6 @@ func reportInfo(_ monitoring.Mode, V monitoring.Visitor) {
 	defer V.OnRegistryFinished()
 
 	delta := time.Since(startTime)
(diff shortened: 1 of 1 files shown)


andrewkroh:
See my comment: https://github.com/elastic/beats/pull/5488#discussion_r149460397

urso:
@andrewkroh @kvch reading the docs on time.Duration I agree with `uptimeMillis := int64(time.Since(startTime)/time.Millisecond)` being the most straight forward idiomatic version. I confused behavior of Nanosecond() method on `time.Time` and `time.Duration` when proposing this `int64(delta.Seconds())*1000 + int64(delta.Nanoseconds()/1000000)`.

kvch:
I've just updated the PR to the formula of @andrewkroh . Also, sorry for missing your comment. :(

PR #5558, opened 2017-11-12
Move pre-commit hook to second command in Makefile

ruflin:
Because the pre-commit hook was the first command in the beats Makefile, this was the standard command that was executed when running make. Before the default command was the teestsuite. I move the pre-commit hook now after the testsuite to make the default working as before.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,14 +16,14 @@ REVIEWDOG_REPO=github.com/haya14busa/reviewdog/cmd/reviewdog
 # Runs complete testsuites (unit, system, integration) for all beats with coverage and race detection.
 # Also it builds the docs and the generators
 
+.PHONY: testsuite
+testsuite:
+	@$(foreach var,$(PROJECTS),$(MAKE) -C $(var) testsuite || exit 1;)
+
 .PHONY: setup-commit-hook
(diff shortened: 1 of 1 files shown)


karmi:
Hi @ruflin, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/beats/pull/5558.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

karmi:
Hi @ruflin, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/beats/pull/5558.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

karmi:
Hi @ruflin, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/beats/pull/5558.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

exekias:
This explains some things :smile_cat:

PR #5570, opened 2017-11-14
Add number_of_routing_shards config set to 30

ruflin:
With elastic/elasticsearch#26931 the possibility for splitting shards was introduced. To make use of this feature for indices created with ES >=6.1 the config option `index.number_of_routing_shards` is required. This adds this config option currently set to 30 as it's a multiple of 1, 3 and 5, our current number of default shards in Beats and ES. This allows users with default configs to scale their split their shards.

The `number_of_routing_shards` can also be overwritten in the config file.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -94,6 +94,7 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
 - Refactor add_kubernetes_metadata to support autodiscovery {pull}5434[5434]
 - Improve custom flag handling and CLI flags usage message. {pull}5543[5543]
 - Update to Golang 1.9.2
+- Add number_of_routing_shards config set to 30 {pull}5570[5570]
 
 *Auditbeat*
 
--- a/auditbeat/auditbeat.reference.yml
+++ b/auditbeat/auditbeat.reference.yml
@@ -755,6 +755,7 @@ setup.template.settings:
   #index:
     #number_of_shards: 1
     #codec: best_compression
+    #number_of_routing_shards: 30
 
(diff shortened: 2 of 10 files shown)


tsg:
The tests complain that:

```
could not load template: couldn't load template: couldn't load json. Error: 400 Bad Request: {\"error\":{\"root_cause\":[{\"type\":\"illegal_argument_exception\",\"reason\":\"the number of source shards [5] must be a must be a factor of [24]\"}],\"type\":\"illegal_argument_exception\",\"reason\":\"the number of source shards [5] must be a must be a factor of [24]\"},\"status\":400}.
```

I think that's currently only an issue in tests, but it does worry me that if anyone is changing `number_of_shards` to 5 in their Beat configuration, they will hit that error.

I think we should:

* Make the `number_of_routing_shards` also a configuration option in the Beat config file, if it's not already there indirectly.
* Change the number to include 5? Maybe 2 * 3 * 5 = 30? Or 4 * 3 * 5 = 60?

ruflin:
Setting it to 30, would mean for the once with 1 shard could do `1,2,3,5,6,10,15,30, the once with 3 shards could do `6,15,30, and 5 shards could do `10,15,30`. Going with 60 would only add `12` to the 3 shard option in the lower band, so not a big benefit. So I think 30 would be a pretty good pick also with the ones that have the default shards set to 5. I think the majority of people will use shards in a single digit area.

ruflin:
@tsg Can you take a look again. I update it to be 30 by default and also updated PR title and description accordingly. I added it to the reference config and added a test to checked if it can be overwritten.

Feder1co5oave:
And I just lost 2 hours trying to setup metricbeat templates because I set the default number of shards to 4 (my number of cluster nodes)...
Silly me!

Issue #5421, opened 2017-10-23
Add recursive watches to file integrity module


andrewkroh:
The file integrity module currently supports watching individual directories and does not recursively watch sub-directories. To add support for recursive watches while still maintaining the ability to detect changes in near real-time we need notifications from the OS that work recursively (see [how it works](https://www.elastic.co/guide/en/beats/auditbeat/master/auditbeat-metricset-audit-file.html) for background info).

So for the various operating systems this means

- Windows - `ReadDirectoryChangesW` already natively supports recursive monitoring. The feature is not exposed through [fsnotify](https://github.com/fsnotify/fsnotify/blob/4da3e2cfbabc9f751898f250b49f2439785783a1/windows.go#L350-L351) (`watchSubtree` is hard-coded to false). Hopefully a small change can be made to fsnotify to expose this feature.
- macOS (Darwin) - We use `kqueue` now which doesn't have support for recursive watches. But we can switch to using [FSEvents](https://github.com/fsnotify/fsevents) which does support recursive watches. Plus it doesn't use as many file descriptors. One important thing to note is that multiple changes can be coalesced into a single notification (more details in this [blog](http://nicoleibrahim.com/apple-fsevents-forensics/)) and changes will be needed to account for this.
- Linux - `inotify` doesn't support recursive watches so we will need to implement this ourselves by adding and removing watches manually for sub-dirs.

### Configuration

We can add a new configuration option like `file.recursive` to enable recursive watches.

```
- module: audit
  metricsets: [file]
  file.paths:
  - /etc
  file.recursive: true
```

### Implementation

I would prefer to have these changes made incrementally for each operating system (e.g. one or more PR's per OS) in order to keep the changes managible from a review and testing perspective.

andrewkroh:
This implements for Linux, macOS, and Windows now.

We can possibly optimize Windows more if we use `ReadDirectoryChangesW` with the `watchSubtree` flag. But we would need to do some measurements to see how the current version compares the the proposed version.

jakommo:
Awesome, which version will this be in. 6.2?

andrewkroh:
Yes, expect a 6.2 release for this feature.

hmnguyen1201:
Thank you for adding this feature. It is great. I am also curious to know if there is an option to allow recursive on selected directories instead of everything?

andrewkroh:
[...]

PR #5575, opened 2017-11-14
macOS recursive file monitoring

adriansr:
This patch adds a new file monitoring backend to the integrity module, using the FSEvents facility from macOS.

**Coalesced events**

FSEvents uses a bitmask to notify for more than one event on the same path. For example created|modified|deleted. Ordering information is lost, so it's impossible to tell from
the event itself if the file was modified, then deleted and then re-created, or first created,
then modified and then deleted. The ordering in which the events are stored in Elasticsearch
seems of little importance as they usually get assigned the same timestamp (granularity of one millisecond).

This patch orders the set of actions in a single event to be meaningful depending if the file existed in the beat database and if it exists anymore at the moment of processing the event.

#5421

### DIFF
--- a/auditbeat/module/audit/file/action.go
+++ b/auditbeat/module/audit/file/action.go
@@ -1,0 +1,152 @@
+package file
+
+import "strings"
+
+// Action is a description of the changes described by an event.
+type Action uint8
+
+// actionArray is just syntactic sugar to invoke methods on []Action receiver
+type actionArray []Action
+
+// List of possible Actions.
+const (
+	None               Action = 0
+	AttributesModified        = 1 << (iota - 1)
+	Created
+	Deleted
+	Updated
+	Moved
+	ConfigChange
+)
+
+var actionNames = map[Action]string{
+	None:               "none",
+	AttributesModified: "attributes_modified",
+	Created:            "created",
+	Deleted:            "deleted",
+	Updated:            "updated",
+	Moved:              "moved",
+	ConfigChange:       "config_change",
+}
+
+type actionOrderKey struct {
+	ExistsBefore, ExistsNow bool
+	Action                  Action
+}
+
+// Given the previous and current state of the file, and an action mask
+// returns a meaningful ordering for the actions in the mask
+var actionOrderMap = map[actionOrderKey]actionArray{
+	{false, false, Created | Deleted}:                  {Created, Deleted},
+	{true, true, Created | Deleted}:                    {Deleted, Created},
+	{false, false, Moved | Created}:                    {Created, Moved},
+	{true, true, Moved | Created}:                      {Moved, Created},
+	{true, true, Moved | Deleted}:                      {Deleted, Moved},
+	{false, false, Moved | Deleted}:                    {Moved, Deleted},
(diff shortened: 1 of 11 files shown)


auditbeat/module/audit/file/action.go
```diff
@@ -0,0 +1,152 @@
...
+	{true, true, Updated | Moved | Deleted}:            {Deleted, Moved, Updated},
+	{false, false, Updated | Created | Deleted}:        {Created, Updated, Deleted},
+	{true, true, Updated | Created | Deleted}:          {Deleted, Created, Updated},
+	{false, true, Updated | Moved | Created | Deleted}: {Created, Deleted, Moved, Updated},
+	{true, false, Updated | Moved | Created | Deleted}: {Deleted, Created, Updated, Moved},
+}
+
+func (action Action) isMultiple() bool {
+	// checks if there is more than one bit set
+	return action != 0 && (action&(action-1)) != 0
```
andrewkroh: There is a bits.OnesCount8 function that is supposed to be optimized for this.

auditbeat/module/audit/file/eventreader_fsevents.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+	return strings.Join(list, "|")
+}
+
+func getInode(path string) (uint64, error) {
+	info, err := os.Lstat(path)
+	if err != nil {
+		return 0, errors.Wrap(err, fmt.Sprintf("failed to stat"))
+	}
+	meta, err := NewMetadata(path, info)
```
andrewkroh: Given that you need only the inode and file type I would use the os.FileInfo directly to avoid some of the extra work that NewMetadata does (like lookup username and group).

auditbeat/module/audit/file/eventreader_fsnotify.go
```diff
@@ -19,6 +20,10 @@ type reader struct {
 
 // NewEventReader creates a new EventProducer backed by fsnotify.
 func NewEventReader(c Config) (EventProducer, error) {
+
```
andrewkroh: Please drop this newline.

auditbeat/module/audit/file/event.go
```diff
@@ -211,7 +180,7 @@ func (e *Event) String() string {
 	return string(data)
 }
 
-func buildMapStr(e *Event) common.MapStr {
+func buildMapStr(e *Event, existsBefore, existsNow bool) common.MapStr {
```
andrewkroh: Could existsNow be determined by looking at e? Then you wouldn’t need the third param.

andrewkroh:
jenkins, test it

andrewkroh:
This looks pretty complete. @adriansr, should the "in progress" be removed?

andrewkroh:
The way `Action` is persisted is going to need to be updated. Or perhaps there is no use in persisting this value?

https://github.com/elastic/beats/blob/d92aa2e327ac4df712578b2b464cd52bc7ead079/auditbeat/module/audit/file/flatbuffers.go#L151-L164

adriansr:
Ok I updated the flatbuffers to persist the action as a bitmask, although the field is not used.

PR #5577, opened 2017-11-14
Added a FAQ topic on disk full conditions

tsg:
It recommends using `close_timeout`. Also updated the `close_timeout` docs
to remove the notice about it not working when the output is blocked.

### DIFF
--- a/filebeat/docs/faq.asciidoc
+++ b/filebeat/docs/faq.asciidoc
@@ -111,5 +111,22 @@ Filebeat uses a newline character to detect the end of an event. If lines are ad
 harvested, a newline character is required after the last line, or Filebeat will not read the last line of
 the file.
 
+[float]
+[[faq-deleted-files-are-not-freed]]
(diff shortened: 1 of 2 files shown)


filebeat/docs/faq.asciidoc
```diff
@@ -111,5 +111,22 @@ Filebeat uses a newline character to detect the end of an event. If lines are ad
...
+
+In the default behaviour, Filebeat opens the files and keeps them open until it
+reaches the end of them.  In situations when the configured output is blocked
+(e.g. Elasticsearch or Logstash is unavailable) for a long time, this can cause
+Filebeat to keep file handlers to files that were deleted from the file system
+in the mean time. As long as Filebeat keeps the deleted files open, the
+operating system doesn't free up the space on disk, which can lead to increase
+disk utilisation or even out of disk situations.
+
+To mitigate this issue, you can set the <<close-timeout>> setting to `5m`. This
```
ruflin: Should we add a note that this leads to data loss?

PR #5578, opened 2017-11-14
Metricbeat docs fixes

exekias:
Corrected docs for new metricbeat / diskio / await section (there was a copy/paste error)


tsg:
@exekias I added the `needs_backport` because I think we need this in the 6.0 branch at least.

exekias:
@tsg If I'm not wrong this wasn't released yet as the metric is from https://github.com/elastic/beats/pull/4648, that wasn't backported to 6.0?

dedemorton:
Confirmed that the docs for this are not in 6.0. Removing the `needs_backport` label.

Issue #4579, opened 2017-06-29
Metricbeat modules status


monicasarbu:
Metricbeat modules:

- **apache** 
- **ceph** (Beta / Experimental)
     - cluster_disk (Experimental)
     - pool_disk (Experimental)
     - cluster_health (Beta)
     - monitor_health (Beta)
- **couchbase** (Beta)
- **docker** (Beta)
- **dropwizard** (Beta)
- **elasticsearch** 
- **golang** (Experimental)
- **haproxy**
- **http** (Beta)
- **jolokia** (Beta)
- **kafka** (Beta)
- **kibana** (Experimental in docs)
- **Kubernetes** (Beta/Experimental)
     - event (Experimental)
     - container, node, pod, state_container, state_deployment, state_node, state_pod, state_replicaset, system, volume (Beta)
- **memcached**
- **mongodb** (Experimental)
- **mysql**
- **nginx**
- **php_fpm** (Beta)
- **postgresql**
- **prometheus** (Beta)
- **rabbitmq** (Experimental)
- **redis**
- **system**
- **vsphere** (Experimental)
- **windows** (Beta)
- **zookeeper**

Questions: ???
- Shouldn't be Elasticsearch  experimental?

cc-ed @ruflin @brandonmensing

ruflin:
Closing this as we have now an identical auto generated list here: https://www.elastic.co/guide/en/beats/metricbeat/master/metricbeat-modules.html

PR #5590, opened 2017-11-14
Flag experimental and beta modules in the docs

dedemorton:
@monicasarbu I've updated docs to reflect the status of the modules shown here.

https://github.com/elastic/beats/issues/4579

A couple of questions: 
* should the ES module be flagged as experimental? beta?
* the list doesn't mention the aerospike or graphite modules. How should those be flagged?

### DIFF
--- a/metricbeat/module/ceph/cluster_status/_meta/docs.asciidoc
+++ b/metricbeat/module/ceph/cluster_status/_meta/docs.asciidoc
@@ -1,3 +1,3 @@
-=== Ceph cluster_health metricset
+=== Ceph cluster_status metricset
 
 This is the `cluster_status` metricset of the Ceph module.
--- a/metricbeat/module/dropwizard/_meta/docs.asciidoc
+++ b/metricbeat/module/dropwizard/_meta/docs.asciidoc
@@ -1,4 +1,6 @@
 == Dropwizard module
 
+beta[]
+
 This is the http://dropwizard.io[Dropwizard] module.
 
--- a/metricbeat/docs/fields.asciidoc
+++ b/metricbeat/docs/fields.asciidoc
@@ -2946,7 +2946,7 @@ format: bytes
 [[exported-fields-etcd]]
 == Etcd fields
 
-[]experimental
+experimental[]
 etcd Module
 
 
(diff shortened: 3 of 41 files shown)


metricbeat/module/ceph/cluster_status/_meta/docs.asciidoc
```diff
@@ -1,3 +1,3 @@
-=== Ceph cluster_health metricset
+=== Ceph cluster_status metricset
```
ruflin: Good catch

ruflin:
Thanks for adding docs for this. 

* Aerospike: Experimental
* Graphite: Experimental

For ES we don't have a beta or experimental flag, so I think it's fine as is.

dedemorton:
@ruflin Oh wait, I see that someone added the experimental flag here, but the syntax is wrong, so it's not resolving correctly: https://www.elastic.co/guide/en/beats/metricbeat/6.0/exported-fields-elasticsearch.html. Should I mark the module as experimental?

Also, I found an option under the mysql module status metricset marked as experimental. Should I keep or remove the flag?

https://www.elastic.co/guide/en/beats/metricbeat/6.0/metricbeat-metricset-mysql-status.html#_raw_config_option

I'm done combing the docs for stuff, so that should be all. :-)

ruflin:
@dedemorton Yes, please format it correctly. And I should add a log line about being experimental to the metricsets, because it's missing. For 6.1 we should go again through the list to see what we make beta and what we release.

For the mysql status metricsets, it's only the raw option which is experimental. This should stay in.

dedemorton:
@ruflin This is ready for a final review.

metricbeat/module/dropwizard/_meta/docs.asciidoc
```diff
@@ -1,4 +1,6 @@
 == Dropwizard module
 
+beta[]
```
ruflin: This should also be experimental I think.

ruflin:
@dedemorton I'm good with merging as is. I think it's good that we now have it "properly" in the docs. I think we should do a second round later on to decide which ones we make beta or release.

PR #5598, opened 2017-11-15
Update kafka container for output tests

urso:
- based on container used by LS (as done in #5549)
- Use healthcheck which creates a random topic to make sure broker is available

metricbeat module tests and testing/environment (only used by libbeat?) each have their own docker setup for common services. This PR copies the container setup from metricbeat. At some point we should see if we can reuse some of the complex container setups like kafka.


urso:
Jenkins, test it.

PR #5602, opened 2017-11-15
Fix repo steps in the Beats platform ref

dedemorton:
This PR fixes an issue where the steps for installing from the repositories were not showing up in the Beats platform ref.

Why did this happen? The `beatname_pkg` attribute was introduced, but never included in libbeat. When asciidoc finds an attribute it cannot resolve, it omits the entire line on which the attribute appears.

I plan to move all the Beats attributes to the [shared file](https://github.com/elastic/docs/tree/master/shared) in the docs repo but have not done that yet because it requires some rework to make our attribute names consistent with other books in the library. The shared file is relatively new, so there are still other projects that are not using it yet.


ruflin:
@dedemorton Does this PR need a `needs_backport` label?

dedemorton:
oy...probably. Thanks for catching this...

PR #5604, opened 2017-11-16
Remove ingest node directions from metricbeat docs

dedemorton:
@monicasarbu  pointed out that it doesn't make sense to use ingest node with Metricbeat, so this PR removes the shared topic from the Metricbeat docs.

### DIFF
--- a/libbeat/docs/outputconfig.asciidoc
+++ b/libbeat/docs/outputconfig.asciidoc
@@ -223,7 +223,11 @@ output.elasticsearch:
   pipeline: my_pipeline_id
 ------------------------------------------------------------------------------
 
+ifeval::["{beatname_lc}"!="metricbeat"]
+
 For more information, see <<configuring-ingest-node>>.
+
+endif::[]
 
 ===== `pipelines`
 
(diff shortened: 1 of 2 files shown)


monicasarbu:
True, enriching your events with geoip info can be useful for all the Beats. Maybe we should have a separated section only about geoip, and include it in the user guide of each Beat.

dedemorton:
@monicasarbu There's a topic in the Packetbeat docs that describes how to use the geoip processor: https://www.elastic.co/guide/en/beats/packetbeat/current/packetbeat-geoip.html#packetbeat-geoip

I can pull that content into a shared file and include it in the docs for other Beats. Are there any Beats that shouldn't have this content? (Of course, I'll make sure the shared content works for all Beats and is not so Packetbeat-specific).

ruflin:
I think it's a good idea to have that geoip for all beats, because it's also useful in all other beats. For this PR I would probably not change our existing content and keep ingest in for now.

dedemorton:
Closing because it sounds like we decided to keep this topic for now

PR #5605, opened 2017-11-16
Pipeline stress tests

urso:
Add support for running go based stress tests for internal components that make use of concurrency. Right now only the publisher pipeline is tested. The tests are run for a configurable duration and should look for races, deadlocks, and panics (invariants should be protected with panics). 

- Package libbeat/publisher/pipeline/stress to have blackbox testing +
  don't mix testing code with actual pipeline implementation.
- Additional sub-package cmd/stress to run custom configurations with
  setting overwrites (for manual testing).
- Test configures a generator (configs/gen) + queue (configs/pipeline) + test output (configs/out)
  - All predefined config files are mixed so to run all combinations of
    tests
  - Test output currently supports:
    - Randomized batch latency (min and max configurable)
    - Fail every n-th batch
- Test errors:
  - Generator has a watchdog timer (configurable) checking the generator
    not being blocked by outputs (e.g. deadlock).
  - Test relies on panics/asserts in critical places in the publisher
    pipeline (these should not be triggered, or test fails).
- By default all tests are run for 10 seconds. Tests are marked as
  parallel, so if multiple CPUs are available, tests will run concurrently
  (still in isolation with separate pipeline instance)


.travis.yml
```diff
@@ -59,6 +59,10 @@ jobs:
       env: TARGETS="-C libbeat crosscompile"
       go: $GO_VERSION
       stage: test
+    - os: linux
```
ruflin: We must make sure to add these tests also to Jenkins

libbeat/publisher/pipeline/stress/gen.go
```diff
@@ -0,0 +1,140 @@
...
+		})
+	}
+
+	logp.Info("start (%v) generator: %v", id, time.Now())
+	defer logp.Info("stop (%v) generator: %v", id, time.Now())
+
+	for cs.Active() {
+		event := beat.Event{
+			Timestamp: time.Now(),
+			Fields: common.MapStr{
```
ruflin: Would be interesting here to see what happens with some large events like very large multiline or some events of metricbeat. "Could" have an affect.

urso: The tests don't do any encoding (yet). So event size has no real impact.

urso:
> I assume the longer we run these tests the more likely we are to detect some issues. My suggestion would be to do a quick Run with x seconds for each PR to keep the CI times low but have for master something similar to packaging that is run after merging and runs for x minutes (if that makes sense).

The tests add a `-duration` flag. The default is 10 seconds (which already takes 200s on travis). As more pipeline tests will be added in the future we decided to run the tests in parallel (hence separate job) during CI.

ruflin:
@urso What you suggest as next step here? Should we get this in an take it from there?

urso:
Too many tests are failing. Even restarting tests a few times still produces much too many errors. I definitely want these new tests in, but will need to take some to to investigate the other failing tests.

urso:
@ruflin fixed failing tests by introducing `libbeat/scripts/cmd` package (which is ignored when building the test binary).

.travis.yml
ruflin: @urso Can you open a follow up issue to add it to Jenkins?

urso: See #5821

ruflin:
@urso thanks for making this happen. Would be great to have a follow up issue to track jenkins and follow up steps.

PR #5625, opened 2017-11-17
Make file integrity path field analyzed

andrewkroh:
- Use a multi-field for the `audit.file.path` field in order to make the path searchable and aggregatable.
- Update the file integrity dashboard to use `audit.file.path.raw` for aggs.
- Modify the file integrity dashboard
  - Replace most active agent viz with number of world-writable files.
  - Add a table summarizing the number of events from each host.
  - Add a saved search viz showing the raw events.
- Fix an issue where `audit.file.type` was not being sent for `file` (it was only send for dir and symlink).

auditbeat-file-integrity-overview-dashboard


auditbeat/module/audit/file/schema/Type.go
```diff
@@ -3,12 +3,14 @@
 package schema
 
 const (
-	TypeFile    = 0
-	TypeDir     = 1
-	TypeSymlink = 2
+	TypeUnknown = 0
```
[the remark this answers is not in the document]
andrewkroh: Dear houndci-bot, this is generated code. Take it up with the flatbuffer compiler.

auditbeat/docs/fields.asciidoc
```diff
@@ -43,11 +43,19 @@ The file metricset generates events when a file changes on disk.
 [float]
 === `audit.file.path`
 
-type: keyword
+type: text
 
 The path to the file.
 
 [float]
+=== `audit.file.path.raw`
```
andrewkroh: Do we want these to show up in the docs? I added code here to do it, but I'm on the fence.

andrewkroh:
There's a test failure that I'm investigating.

auditbeat/docs/fields.asciidoc
ph: I am OK to have it in the doc since its useful for aggregation.
Even if we don't display it in our doc it will still be accessible in the UI :)

libbeat/scripts/generate_fields_docs.py
```diff
@@ -62,6 +62,10 @@ def document_field(output, field, path):
         if not field["enable"]:
             output.write("{}\n\n".format("Object is not enabled."))
 
+    if "multi_fields" in field:
+        for subfield in field["multi_fields"]:
+            document_field(output, subfield, path + "." + subfield["name"])
+
```
ph: good catch.

CHANGELOG.asciidoc
```diff
@@ -18,6 +18,8 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
 
 *Auditbeat*
 
+- Changed `audit.file.path` to be a multi-field so that path is searchable. {pull}nnnn[nnnn]
+
```
ph: update with PR number

andrewkroh:
@ph It should be fixed now. Attempts to debug after midnight are futile. It was much easier to debug after resting 🛌 .

libbeat/scripts/generate_fields_docs.py
ruflin: @simitt Mentioning you here as apm-server uses multi fields as far as I remember. Not really a bug but just in case you are surprised why the docs are update with the most recent beats version.

PR #5631, opened 2017-11-17
Backout deb fpm changes

### LINKED ISSUE #5477: 6.x has different post-install script from 6.0
It looks like in the 6.x branch we no longer have this [post-install script](https://github.com/elastic/beats/blob/6.0/dev-tools/packer/platforms/debian/run.sh.j2#L10-L12). Instead, I think FPM generates default post-install/pre-rm scripts that cause the apm-server 6.0 to behave slightly different from MB/FB 6.0. In particular, apm-server starts automatically when installed, while the Beats don't.

Opening this to further investigate the issue. cc @ruflin

tsg:
This backouts changes in #5086 and #5174. From testing the resulting packages it turns out that after that change the post-rm and pre-install scripts are significantly different. This results in at least one change in behaviour: the service is started automatically at install time, which is not the case in 6.0. There could be more changes in behaviour as well, so I'd like to play it safe and backout the changes for now.

Closes #5477.

### DIFF
--- a/dev-tools/packer/platforms/debian/run.sh.j2
+++ b/dev-tools/packer/platforms/debian/run.sh.j2
@@ -7,6 +7,10 @@ cd /build
 # the init scripts needs to have the right name
 cp ${RUNID}.init /tmp/{{.beat_pkg_name}}.init
 
+# create script to reload systemd config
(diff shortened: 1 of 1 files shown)


tsg:
Yes, I gave it some quick local testing, looks good.

PR #5633, opened 2017-11-17
Deprecate the LS document_type config option

dedemorton:
Mentions that document_type is deprecated in LS 6.0.

(This resolves an issue being tracked in https://github.com/elastic/beats/issues/4540)


ruflin:
@dedemorton Should that have been a `needs_backport` label instead of backport?

PR #5635, opened 2017-11-18
Clarify docs about manually loading the template

### LINKED ISSUE #5615: Docs to manually loading template in 6.0
The documentation for manually loading index templates only works if you are using the ES output. The users that most commonly need to manually load the index templates are users of non-ES outputs.

To accomplish this task you need to disable the Logstash output and manually enable ES on the CLI.

```
filebeat setup --template -E output.logstash.enabled=false -E 'output.elasticsearch.hosts=["localhost:9200"]'
```

Alternatively you can export the template to a file and use `curl` or `Invoke-RestMethod`. This is useful if the host running Beats does not have direct connectivity to ES.

Windows:

```
# Write the template to disk.
PS> .\winlogbeat.exe export template --es.version 6.0.0 | Out-File -Encoding UTF8 winlogbeat.template.json

# Install it to ES.
PS > Invoke-RestMethod -Method Put -ContentType "application/json" -InFile winlogbeat.template.json -Uri http://elasticsearch:9200/_template/winlogbeat-6.0.0
```

Unix:

```
filebeat export template > filebeat.template.json

curl -XPUT -H 'Content-Type: application/json' http://elasticsearch:9200/_template/filebeat-6.0.0 -d@filebeat.template.json
```

dedemorton:
Resolves #5615 

I also did a bit of editing to pare away some unnecessary words, reflow sections, and add context where I felt it was missing.

The conditional coding on this particular topic is a bit eccentric due to the existence of Winlogbeat, so I'm going to paste a screen capture of the compiled webpage here. For Winlogbeat (not shown here), all the unix-specific commands are removed. 

**These screens are old. See the diff for changes**

load-template

(I'm going to fix the "pass credentials" section in a separate PR because it's shared content that lives in a separate file.)

### DIFF
--- a/libbeat/docs/shared-template-load.asciidoc
+++ b/libbeat/docs/shared-template-load.asciidoc
@@ -17,16 +17,21 @@
 
 
 In Elasticsearch, {elasticsearch}/indices-templates.html[index
-templates] are used to define settings and mappings that determine how fields should be analyzed.
+templates] are used to define settings and mappings that determine how fields
+should be analyzed.
 
-The recommended index template file for {beatname_uc} is installed by the {beatname_uc} packages. If you accept
-the default configuration for template loading in the +{beatname_lc}.yml+ config file,
-{beatname_uc} loads the template automatically after successfully connecting to Elasticsearch. If the template
-already exists, it's not overwritten unless you configure {beatname_uc} to do so.
+The recommended index template file for {beatname_uc} is installed by the
+{beatname_uc} packages. If you accept the default configuration in the
++{beatname_lc}.yml+ config file, {beatname_uc} loads the template automatically
+after successfully connecting to Elasticsearch. If the template already exists,
+it's not overwritten unless you configure {beatname_uc} to do so.
 
-If you want to disable automatic template loading, or you want to load your own template,
-you can change the settings for template loading in the {beatname_uc} configuration file. If you
-choose to disable automatic template loading, you need to load the template manually.
+You can disable automatic template loading, or load your own template, by
+configuring template loading optons in the {beatname_uc} configuration file. 
+
+NOTE: A connection to Elasticsearch is required to load the index template. If
+the output is Logstash, you must load the template manually. 
+
 For more information, see:
 
 * <<load-template-auto>> - supported for Elasticsearch output only
@@ -35,10 +40,11 @@ For more information, see:
 [[load-template-auto]]
 ==== Configure template loading
 
-By default, {beatname_uc} automatically loads the recommended template file, +fields.yml+,
-if Elasticsearch output is enabled. You can configure {beatname_lc} to load a different template
-by adjusting the `setup.template.name` and `setup.template.fields` options in
-+{beatname_lc}.yml+ file:
+By default, {beatname_uc} automatically loads the recommended template file,
++fields.yml+, if Elasticsearch output is enabled. You can configure
+{beatname_lc} to load a different template by adjusting the
+`setup.template.name` and `setup.template.fields` options in +{beatname_lc}.yml+
+file:
 
 ["source","yaml",subs="attributes,callouts"]
 ----------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


libbeat/docs/shared-template-load.asciidoc
```diff
@@ -47,27 +53,42 @@ setup.template.fields: "path/to/fields.yml"
...
+Elasticsearch, see <<load-template-manually-alternate>>.
 
 ifdef::allplatforms[]
 
 *deb, rpm, and mac:*
 
 ["source","sh",subs="attributes"]
 ----
-./{beatname_lc} setup --template
+./{beatname_lc} setup --template -E output.logstash.enabled=false -E 'output.elasticsearch.hosts=["localhost:9200"]'
```
dedemorton: decided to show this in the example since it sounds like the most common use case for manual loading is when Logstash output is configured.

libbeat/docs/shared-template-load.asciidoc
```diff
@@ -123,6 +143,37 @@ match the pattern.
...
+Elasticsearch, you can export the index template to a file and then install the
+template manually.
+
+ifdef::allplatforms[]
+
+*deb, rpm, and mac:*
+
+["source","sh",subs="attributes"]
+----
+./{beatname_lc} export template > {beatname_lc}.template.json
```
andrewkroh: For Winlogbeat I think it makes sense to show the `curl` command because you might export it on Windows and then import it on a *nix machine. This does mess up the conditional logic because the export command needs to be `.\{beatname_lc}.exe`.

To solve this I recommend breaking apart the export and import steps. Keep the conditional logic around the export commands. And always show how to import it using both Windows (`Invoke-RestMethod` ) and nix (`curl`).

andrewkroh:
In the section where we show how to delete the index I think we should show the equivalent Windows command.

`Invoke-RestMethod -Method Delete "http://localhost:9200/filebeat-*"`

libbeat/docs/shared-template-load.asciidoc
```diff
@@ -107,6 +110,7 @@ ifeval::["{beatname_lc}"!="auditbeat"]
 docker run {dockerimage} setup --template -E output.logstash.enabled=false -E 'output.elasticsearch.hosts=["localhost:9200"]'
 ----------------------------------------------------------------------
 
+
```
dedemorton: Adding double line breaks after examples because it fixes rendering issues when these files are viewed on GitHub.

dedemorton:
@andrewkroh I've made the changes that you requested. I decided to make the Note a separate section because it was getting too long to be a note. 

I have not updated the original screen shots in the PR to show these changes. If you want me to, just let me know.

libbeat/docs/shared-template-load.asciidoc
```diff
@@ -101,27 +126,78 @@ and run:
...
+This command deletes all indexes that match the pattern +{beatname_lc}-*+.
+Before running this command, make sure you want to delete all indexes that match
+the pattern.
+
+[[load-template-manually-alternate]]
+==== Load the template manually (alternate method)
+
+If the host running {beatname_uc} does not have direct connectivity to
+Elasticsearch, you can export the index template to a file and then install the
```
andrewkroh: Maybe it won't be obvious to users that they need to move or copy it to a different machine with ES connectivity. Should we add this into this sentence (e.g. export, transfer, and then install)?

libbeat/docs/shared-template-load.asciidoc
```diff
@@ -101,27 +126,78 @@ and run:
...
+PS> .{backslash}{beatname_lc}.exe export template --es.version {stack-version} | Out-File -Encoding UTF8 {beatname_lc}.template.json
+----
+
+. Install the template:
++
+*deb, rpm, and mac:*
++
+["source","sh",subs="attributes"]
+----
+curl -XPUT -H 'Content-Type: application/json' http://elasticsearch:9200/_template/{beatname_lc}-{stack-version} -d@{beatname_lc}.template.json
```
andrewkroh: In other places we just use `localhost:9200`. To be consistent I think this (and the one below) should also be `localhost:9200`.

dedemorton:
@andrewkroh I've made the additional changes that you requested.

Issue #5638, opened 2017-11-20
Docs pointing to prerelease for 6.0


markwalkom:
As seen here - https://www.elastic.co/guide/en/beats/filebeat/current/setup-repositories.html. Also looks like metricbeat has the same issue.

DEB;
> echo "deb https://artifacts.elastic.co/packages/6.x-prerelease/apt stable main" | sudo tee -a /etc/apt/sources.list.d/elastic-6.x-prerelease.list

RPM;
> [elastic-6.x-prerelease]

I can see that we dynamically figure this out, but I am not sure how that works so didn't want to make a PR and break things :)

markwalkom:
Reported by @prasket here https://twitter.com/_prasket/status/932406690200162304

ruflin:
@markwalkom Thanks for bringing this up. I opened a PR here to fix this for 6.0: https://github.com/elastic/beats/pull/5641

@dedemorton @tsg Should we introduce here alson an if/else clause to make sure this is set currectly based on the prerelease variable? For 6.x we need to make sure to set this correctly. But not sure at which stage we set it to "release" and is the reason I didn't forward port this PR yet.

dedemorton:
@ruflin I'll add conditional coding to the source, so don't worry about forward porting this. Here's how we manage the 3 conditions:

unreleased - prior to alpha/beta
prerelese - alpha/beta docs
released - GA docs

We set these states when we prepare the docs for a release. Unfortunately the apt/yum content didn't have the conditional coding.

dedemorton:
Fixed for 6.0 in #5641 

Also added conditional coding in https://github.com/elastic/beats/issues/5652 so the correct into will be set based on the release-state.

PR #5641, opened 2017-11-20
Fix prerelease links in apt/yum package docs

ruflin:
Closes https://github.com/elastic/beats/issues/5638


tsg:
Thanks for fixing it @ruflin. I think this doesn't need a backport, right? Because it's the only recently released branch.

@dedemorton I was sure these are hooked to the `:release-state:` variable. Am I miss-remembering or did we lost that at some point?

dedemorton:
@tsg I thought so too! I must have overlooked this bit of content when I added the prerelease tag. I will add some conditional coding to this topic so we don't have to keep changing this manually.

Issue #5600, opened 2017-11-15
Module/metricset for Metricbeat: RAID Metrics 


plinde:
Enhancement to Metricbeat for collecting RAID-related metrics; specifically for the equivalents of these commands:
- `cat /proc/mdstat`
- `mdadm`

ruflin:
Here an example content of `/proc/mdstat` from a machine with 2 disks and raid1:

```
cat mdstat
Personalities : [linear] [raid0] [raid1] [raid10] [raid6] [raid5] [raid4]
md2 : active raid1 sda5[0] sdb5[1]
      2925531648 blocks super 1.2 [2/2] [UU]

md1 : active raid1 sda2[0] sdb2[1]
      2097088 blocks [2/2] [UU]

md0 : active raid1 sda1[0] sdb1[1]
      2490176 blocks [2/2] [UU]

unused devices: <none>
```

@andrewkroh We should probably add support for this in gosigar? https://github.com/elastic/gosigar

andrewkroh:
@ruflin There's support for this in procfs so we should use that. https://github.com/elastic/procfs/blob/master/mdstat.go

The question I'm pondering is where do we add this. linux/mdstat, system/mdstat, or a maybe more general system/raid?

ruflin:
My thought process here:

* Raid is something generic that exists on all (?) OS
* Information on Windows should be similar (never created a Raid on Windows)

This kind of leads to `system/raid`. One issue I have with `raid` is that `raid` could be perhaps `mdstats + ?`, meaning is there more information under `/proc` we should add to raid?

`raid` also seems to be the most user friendly one. I would not have known from top of my head that the information is in `mdstats`.

ruflin:
@plinde I put a PR together with the data in. https://github.com/elastic/beats/pull/5642

@andrewkroh The metricset is not tested yet as I'm not sure how test this best with "actual" raid data. Some ideas here?

ruflin:
@plinde Could you let me know which are the exact metrics you are looking for from the mdstats file? I have a few in #5642 but not sure if these are the right ones and if you need some more.

plinde:
@ruflin Looks great! I think it would be beneficial to include the following additional metrics per RAID device.
- Working Devices: int
- Failed Devices: int
- Spare Devices: int

If possible, it would also be good to compare the blocks.synced/total and perhaps include a Boolean for "synced: true". The scenarios for this being false would include during the rebuild (syncing) of a disk.

[...]

PR #5642, opened 2017-11-20
Add system/raid metricset

ruflin:
Implements #5600.


metricbeat/module/system/raid/raid.go
```diff
@@ -0,0 +1,82 @@
...
+	debugf        = logp.MakeDebug(debugSelector)
+)
+
+func init() {
+	if err := mb.Registry.AddMetricSet("system", "raid", New, parse.EmptyHostParser); err != nil {
+		panic(err)
+	}
+}
+
+type MetricSet struct {
```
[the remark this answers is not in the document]
tsg: I guess it would be worth modifying the generator to add comments for all these.

tsg:
LGTM, but it would be good to have _some_ sort of testing on it.

metricbeat/module/system/raid/raid.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	events := make([]common.MapStr, 0, len(stats))
+	for _, stat := range stats {
+		event := common.MapStr{
+			"name":           stat.Name,
+			"activity_state": stat.ActivityState,
+			"disks": common.MapStr{
+				"active": stat.DisksActive,
+				"total":  stat.DisksTotal,
```
andrewkroh: Usage question: How would I figure out how many available spares there are? Is `spares = total - active`? Or does `total` include the number of failed disks still in the array?

metricbeat/module/system/raid/raid.go
```diff
@@ -0,0 +1,76 @@
...
+func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
+	cfgwarn.Experimental("The system raid metricset is experimental")
+
+	systemModule, ok := base.Module().(*system.Module)
+	if !ok {
+		return nil, errors.New("unexpected module type")
+	}
+
+	mountPoint := filepath.Join(systemModule.HostFS, procfs.DefaultMountPoint)
+	fs, err := procfs.NewFS(mountPoint)
```
andrewkroh: For testing purposes I'd make it possible to fully define the proc dir location so that you can create `testdata/proc/mdstat` and read it.

metricbeat/module/system/raid/raid_test.go
```diff
@@ -0,0 +1,25 @@
...
+package raid
+
+import (
+	"testing"
+
+	mbtest "github.com/elastic/beats/metricbeat/mb/testing"
+)
+
+func TestData(t *testing.T) {
+
```
andrewkroh: Consider removing this newline for me 😄 . (BTW I have a tool for this at [nonewlines](https://github.com/andrewkroh/go-nonewlines))

metricbeat/module/system/raid/raid.go
andrewkroh: Here's some sample data for testing: https://github.com/elastic/procfs/blob/master/fixtures/mdstat

ruflin:
@andrewkroh I added some basic tests (still WIP). It works with the file from the procfs repo but not the one I got from our servers. The error I get:

```
        Error:          Received unexpected error "error parsing _meta/test/proc/mdstat: unexpected statusline:       497754112 blocks super 1.2 512k chunks"
```

The content of the file looks as following:

```
Personalities : [raid0] [raid1] [linear] [multipath] [raid6] [raid5] [raid4] [raid10]
md0 : active raid1 sdb1[1] sda1[0]
      1047552 blocks super 1.2 [2/2] [UU]

md1 : active raid0 sdb2[1] sda2[0]
      497754112 blocks super 1.2 512k chunks

unused devices: <none>
```

Probably an issue we have to solve in procfs (didn't look into it yet).

metricbeat/module/system/raid/_meta/test/mdstat1
```diff
@@ -0,0 +1,8 @@
+Personalities : [raid0] [raid1] [linear] [multipath] [raid6] [raid5] [raid4] [raid10]
```
andrewkroh: I recommend storing this under `system/raid/testdata/[testname]/proc/mdstat`. Using `testdata` seems to be conventional to Go; we have it in a few places. And keeping the `proc/mdstat` ensures that we don't have to mess with temporary files which eliminates code and potential problems. To switch between test cases you just change the `raid.mount_point` between `testdata/scenarioA` and `testdata/scenarioB`.

The "[testname]" could be anything "hostA", "scenarioA", "ubuntu16".

metricbeat/module/system/raid/raid.go
ruflin: @andrewkroh Based on https://raid.wiki.kernel.org/index.php/Mdstat I think what you state is correct.

metricbeat/module/system/raid/_meta/test/mdstat1
ruflin: I changed it to just `/proc/mdstats` inside testdata as I realised one mdstat file should be enough and we can just add more lines. I kept the `testdata` directory inside _meta as we do for other metricsets too to not conflict with package and and "automatic logic to gather files" we have around metricbeat. Like this only the `_meta` directory has to be ignored from all these logics.

ruflin:
@andrewkroh I for now skipped the special raid data I have pasted above as this has to be fixed in procfs. I wanted to open an issue there for it but it seems there is no issue tab? https://github.com/elastic/procfs Is that intentional?

metricbeat/module/system/raid/raid_test.go
ruflin: @andrewkroh You should add this tool to our `make fmt` part so CI can bug us in the future instead of requiring the krohcompiler. Like this you can also go on vacation :-D

metricbeat/module/system/raid/raid.go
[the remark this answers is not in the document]
ruflin: @tsg We actually have pretty extensive comments in the generator which tell you what to do: https://github.com/elastic/beats/blob/master/metricbeat/scripts/module/metricset/metricset.go.tmpl#L17 I removed all of them manually to not have the boiler plate ...

andrewkroh:
I think issues are turned off by default for forks to encourage using the parent's issue tracker. But you can turn issues on in the project's settings IIRC.

metricbeat/module/system/raid/raid_test.go
andrewkroh: It will be part of gofmt in Go 1.10. I'm just trying to prepare you in advance.

ruflin:
@andrewkroh How close are we still to the "parent project". I actually don't have access rights to turn on the issue tracker :-)

metricbeat/module/system/raid/raid_test.go
ruflin: I need to check if it was you that contributed it to Go 1.10 :-)

ruflin: BTW: Probably worth adding the script it will take some time until we are on 1.10 (lots of newlines to come until then)

ruflin:
@plinde @andrewkroh I suggest we get this in as is and start to iterate on top of it?

ruflin:
jenkins, retest it

Issue #5608, opened 2017-11-16
Packetbeat:mysql readLinteger funciton get wrong offset (bug)


ggg-geyi:
I'm using packetbeat to monitor MySQL database, and found that packetbeat do not support mysql_stmt_prepare and mysql_stmt_execute command, and too many SQL querys can not be extracted because of using prepare statement.Is there a plan to support these command type?

Also i found a little bug in mysql.go when call readLinteger function.See the picture below:
![image](https://user-images.githubusercontent.com/24643999/32882573-769e236c-ca7a-11e7-92ed-93a4ab247505.png)

ggg-geyi:
By the way，i am doing the job to support prepare statement command

urso:
This issues seems to be fixed with #8084. Closing. Please reopen if issue is still present in newer versions.

PR #5643, opened 2017-11-20
Support mysql prepare statement command

ggg-geyi:
#5608


tsg:
@ggg-geyi Thanks for the PR! Can you add a sample PCAP file that contains these commands?

Issue #5644, opened 2017-11-20
Build after clone source from github fails


dlasrado:
If the local source is empty, the build script clones the code from Github. But the source code is copied to a wrong folder which causes build_before to fail during docker run.
The code should be copied to github.com/elastic rather than github.com/elastic/beats
(Line 57 in dev-tools/packer/docker/xgo_image/base/build.sh)

ruflin:
Linking the PR here: https://github.com/elastic/beats/pull/5645

andrewkroh:
We have completely refactored how packaging is done since this was opened. I suggest giving the latest a try.

PR #5645, opened 2017-11-20
Fix for #5644 - Clone from github caused build_before to fail

dlasrado:
The clone from Githup repo (when source == '') goes into $GOPATH/src/${GIT_REPO} which is beats folder. It should be rather at github.com/elastic folder.
This bug causes the build_before.sh to fail during Make.

### DIFF
--- a/dev-tools/packer/docker/xgo-image/base/build.sh
+++ b/dev-tools/packer/docker/xgo-image/base/build.sh
@@ -50,6 +50,7 @@ if [ "$SOURCE" != "" ]; then
 else
         mkdir -p $GOPATH/src/${GIT_REPO}
         cd $GOPATH/src/${GIT_REPO}
+	cd ..
         echo "Fetching main git repository ${GIT_REPO} in folder $GOPATH/src/${GIT_REPO}"
         git clone https://${GIT_REPO}.git
 fi


ruflin:
@dlasrado Can you share some details in which directory did you run this exactly and which command? Also was this for one of our beats or a community beat?

dlasrado:
@ruflin This issue occurred when I was trying to build my own output plugin for filebeat (within libbeats/outputs). I dont need to fork or modify the actual code at elastic/beats. I only need the Makefile to build a rpm.
The option to clone the beats code directly from Github is used during docker build. This is achieved by sending SOURCE variable as empty (-e SOURCE=\).
xgo-image/base/build.sh has a condition to get the source from Github directly into docker rather than use from a volume (line 48).
The folder in which code is checkout (Cloned) is github.com/elastic/beats rather than github.com/elastic. Its a small bug, but it stalls the build process.

ruflin:
Can you share what the exact commands are that you execute?

The reason I'm asking is because the changed script is used for our automated packaging and it seems to work.

dlasrado:
@ruflin , Thanks for being patient with me.

Below is a way to replicate.
In libbeat/scripts/Makefile, modify the docker run command to look like below. Or just run the docker run to build the rpms.
# Prepares for packaging. Builds binaries with cgo
.PHONY: prepare-package-cgo
prepare-package-cgo:

	# cross compile on ubuntu
	docker run --rm \
		-v $(abspath ${ES_BEATS}/dev-tools/packer/xgo-scripts):/scripts \
		-v $(abspath ${PACKER_TEMPLATES_DIR}):/templates \
		-v $(BUILD_DIR):/build \
		-e PACK=${BEAT_NAME} \
		-e BEFORE_BUILD=before_build.sh \
		-e SOURCE= \
		-e TARGETS=${TARGETS} \
		-e BUILDID=${BUILDID} \
		-e ES_BEATS=${ES_BEATS} \
		-e BEAT_PATH=${BEAT_PATH} \
		-e BEAT_NAME=${BEAT_NAME} \
		${BEATS_BUILDER_IMAGE}
As you can see, the SOURCE is empty. Hence the build will try to clone from github, which is what I want.

But due to the bug, we get below error:
Cloning into 'beats'...
Fetching dependencies...
Execute /scripts/before_build.sh github.com/elastic/beats/filebeat ..
cp: cannot stat 'fields.yml': No such file or directory
make[1]: *** [prepare-package-cgo] Error 1
make: *** [package] Error 2

ruflin:
So you are not trying to build filebeat, but only the output plugin? For the above: As `/source` seems to be hardcoded, it means you modified the Makefile on your side?

@urso As this is about plugins, perhaps you can chime in?

dlasrado:
@ruflin,  yes your are right.

ruflin:
Ok, I see. So the change you pushed so far would break our build or the one for the community beats I think. Could you push all the changes you made?

ruflin:
@dlasrado Any update on this one?

ruflin:
Closing as no activity recently.

@dlasrado We can still reopen the PR if needed. Let's continue the discussion.

PR #5647, opened 2017-11-20
Use local timezone for TZ conversion in the FB system module

### LINKED ISSUE #3898: Filebeat 'system' module assumes UTC in ingest pipeline
My initial question on ES discuss:

> I'm using filebeat to import syslog messages. When looking at the ES document it appears filebeat incorrectly assumes UTC:
> 
> ES document:
> 
> "@timestamp": "2017-04-01T15:26:51.000Z"
> Syslog message
> 
> "timestamp": "Apr 1 15:26:51"
> OS:
> 
> Sat Apr 1 15:26:51 CEST 2017
>
> Should filebeat not store the @timestamp as 13:26:51?

Response from @tsg: 

> Thanks for the feedback, I think a variable for the timezone makes sense in this module. Could you open a Github ticket for it, please? It's not trivial to implement because currently we insert the pipelines "as is" without expanding variables into them, but I think that's something we can add.

- Version: 5.3.0
- Operating System: Debian 8 (using ES repo)
- Steps to Reproduce: add syslog messages in localtime using filebeat, messages are stored in UTC while the time is still localtime and not converted to UTC

tsg:
This adds a `convert_timezone` fileset parameter that, when enabled,
does two things:

* Uses the `add_locale` processor in the FB proespector config
* Uses `{{ beat.timezone }}` as the `timezone` parameter for the
  date processor in the Ingest Node pipeline. This parameter accepts
  templates starting with ES 6.1.

For the moment the `convert_timezone` flag is off by default, to keep
backwards compatibility.

If ES is < 6.1, the flag is automatically considered off when creating the pipeline
(but the `add_locale` processor is still used).

Closes #3898.

For now this is only applied to the system module, but likely more
modules would benefit from this feature.

Todos:

* [x] Automatically disable the flag in case of ES < 6.1
* [x] Add the flag to the auth fileset as well
* [x] Tests
* [x] Document the new flag + Changelog

### DIFF
--- a/filebeat/fileset/fileset.go
+++ b/filebeat/fileset/fileset.go
@@ -17,6 +17,7 @@ import (
 	"text/template"
 
 	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/logp"
 	mlimporter "github.com/elastic/beats/libbeat/ml-importer"
 )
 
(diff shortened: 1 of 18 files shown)


filebeat/fileset/fileset.go
```diff
@@ -155,18 +161,58 @@ func (fs *Fileset) evaluateVars() (map[string]interface{}, error) {
...
+	haveVersion, err := common.NewVersion(esVersion)
+	if err != nil {
+		return vars, fmt.Errorf("Error parsing version %s: %v", esVersion, err)
+	}
+
+	for _, vals := range fs.manifest.Vars {
+		var exists bool
+		name, exists := vals["name"].(string)
+		if !exists {
+			return nil, fmt.Errorf("Variable doesn't have a string 'name' key")
```
andrewkroh: Not sure it matters, but `!exists` in this case is telling you that either "name" does not exist or `name` is not a string. And maybe calling the variable `ok` would be better since it is more vague.

Same comment for the other locations where the type assertion and map lookup signals are coalesced.

filebeat/fileset/fileset.go
```diff
@@ -155,18 +161,58 @@ func (fs *Fileset) evaluateVars() (map[string]interface{}, error) {
...
+		}
+
+		minESVersion, exists := vals["min_elasticsearch_version"].(map[string]interface{})
+		if exists {
+			minVersion, err := common.NewVersion(minESVersion["version"].(string))
+			if err != nil {
+				return vars, fmt.Errorf("Error parsing version %s: %v", minESVersion["version"].(string), err)
+			}
+
+			logp.Debug("fileset", "Comparing ES version %s with %s", haveVersion, minVersion)
```
andrewkroh: I suggest "Comparing ES version %s with requirement  of %s" to make the meaning of the two versions more clear.

filebeat/fileset/fileset.go
```diff
@@ -155,18 +161,58 @@ func (fs *Fileset) evaluateVars() (map[string]interface{}, error) {
 	return vars, nil
 }
 
+// turnOffElasticsearchVars re-evaluates the variables that have `min_elasticsearch_version`
+// set.
+func (fs *Fileset) turnOffElasticsearchVars(vars map[string]interface{}, esVersion string) (map[string]interface{}, error) {
```
andrewkroh: Nice feature.

PR #5653, opened 2017-11-20
Support dashboard loading without Elasticsearch

kvch:
Make it possible to import dashboards without Elasticsearch using Kibana API.

Usage
```
$ ./metricbeat setup --dashboards --force-kibana
```
The flag `force-kibana` can be specified only when dashboards are loaded.
```
$ ./metricbeat --template --force-kibana
Error setting up beat: --force-kibana is only used in case of dashboard loading
```

The name of the flag is not the best. I am open to proposals.

Based on a discussion over IRC.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -14,7 +14,6 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
 
 *Affecting all Beats*
 
-- Determine log level for kafka output. {pull}5397[5397]
 - Support dashboard loading without Elasticseach {pull}5653[5653]
 
 *Auditbeat*
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -14,6 +14,9 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
 
 *Affecting all Beats*
 
+- Determine log level for kafka output. {pull}5397[5397]
+- Support dashboard loading without Elasticseach {pull}5653[5653]
+
 *Auditbeat*
 
 - Changed `audit.file.path` to be a multi-field so that path is searchable. {pull}5625[5625]
--- a/auditbeat/auditbeat.reference.yml
+++ b/auditbeat/auditbeat.reference.yml
@@ -722,6 +722,9 @@ output.elasticsearch:
 # dashboards and index pattern. Example: testbeat-*
 #setup.dashboards.index:
 
+# Force loading of dashboards using the Kibana API without querying Elasticsearch for the version
+#always_kibana: false
+
 #============================== Template =====================================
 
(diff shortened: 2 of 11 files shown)


ruflin:
@kvch Can you share some more details on why we need this? Instead of requiring a `force` param couldn't we change the current load dashboard command so it works with ES in 6.x? Why is ES required at all here?

tsg:
@ruflin The current algorithm is something like this:

* Check the ES version
 * if it's < 6.0, load the dashboards directly in ES 5.x style
 * if it's >= 6.0, load the dashboards via the Kibana API

What this PR changes is a way to skip the ES version check for when the Beat has access to the Kibana API but not the ES API.

@kvch We generally prefer configuration file options more than CLI flags (because any config can be set from the CLI anyway), so maybe a configuration option in the `setup.kibana` namespace. Since this is adding a flag only to the `setup` command, it's not too bad, though. But it could be that the config option results in cleaner code.

kvch:
I added a new config option called `load_dashboards` under `setup.kibana` as @tsg requested. By default it is false.

kvch:
The new config option is under `setup.dashboards` called `always_kibana`. By default it is false.

```
# Force loading of dashboards using Kibana instead of Elasticsearch
#always_kibana: false
```

tsg:
@kvch LGTM. Can you also add documentation for the new option and a CHANGELOG entry, please?

libbeat/_meta/config.reference.yml
```diff
@@ -628,6 +628,9 @@ output.elasticsearch:
 # dashboards and index pattern. Example: testbeat-*
 #setup.dashboards.index:
 
+# Force loading of dashboards using Kibana instead of Elasticsearch
```
tsg: Suggestion: "Force loading of dashboards using the Kibana API without querying Elasticsearch for the version"

kvch:
I added the entry and changed the documentation.

CHANGELOG.asciidoc
```diff
@@ -14,6 +14,9 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
 
 *Affecting all Beats*
 
+- Determine log level for kafka output. {pull}5397[5397]
```
tsg: This is probably a rebase artifact. Can you remove the line, please?

tsg:
Please also add the config option to `libbeat/docs/dashboardsconfig.asciidoc`.

auditbeat/auditbeat.reference.yml
```diff
@@ -722,6 +722,9 @@ output.elasticsearch:
 # dashboards and index pattern. Example: testbeat-*
 #setup.dashboards.index:
 
+# Force loading of dashboards using the Kibana API without querying Elasticsearch for the version
+#always_kibana: false
```
ruflin: @kvch Should this be prefixed by `setup.dashboards.`?

Issue #5652, opened 2017-11-20
Conditionally code apt/yum install instuctions for pre-releases


dedemorton:
Right now the instructions for installing from apt/yum do not make use of the prerelease tag. They should.

See related issue: https://github.com/elastic/beats/pull/5641/

dedemorton:
Fixed in #5654

PR #5654, opened 2017-11-20
Add prerelease conditional coding to apt/yum install instructions

dedemorton:
I'll include screen captures here to show how the page renders based on the release-state setting:

**When release-state == prerelease**

apt-yum-prerelease

**When release-state == released**

apt-yum-released


dedemorton:
Fixes #5652

PR #5655, opened 2017-11-20
TLS: Config flag to include raw certificates

adriansr:
Use `tls.include_raw_certificates` to enable addition of the raw certificate
in PEM format to events. Default: false.

 #3604

### DIFF
--- a/packetbeat/_meta/fields.yml
+++ b/packetbeat/_meta/fields.yml
@@ -1898,6 +1898,14 @@
               description: >
                 The algorithm used for the certificate's signature.
 
+            - name: alternative_names
+              type: array
+              description: Subject Alternative Names for this certificate.
+
+            - name: raw
+              type: keyword
+              description: The raw certificate in PEM format.
+
             - name: subject
               type: group
               description: Subject represented by this certificate.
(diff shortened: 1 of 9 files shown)


packetbeat/packetbeat.reference.yml
```diff
@@ -433,6 +433,10 @@ packetbeat.protocols:
   # certificate chains are sent to Elasticsearch. The default is true.
   #send_certificates: true
 
+  # If this option is enabled, the raw certificates will be stored
+  # in PEM format under the `raw` key. The default is false.
+  #include_raw_certificates: false
```
tsg: This one is generated, you need to modify `packetbeat/_meta/beat.reference.yml` instead and run `make update`. That's why you get a check failure.

packetbeat/_meta/fields.yml
```diff
@@ -1898,6 +1898,14 @@
               description: >
                 The algorithm used for the certificate's signature.
 
+            - name: alternative_names
+              type: array
+              description: Subject Alternative Names for this certificate.
+
+            - name: raw
+              type: keyword
```
andrewkroh: I think our index templates use an ignore_above value of 1024 bytes. Will the PEM values hit that limit?

https://www.elastic.co/guide/en/elasticsearch/reference/current/ignore-above.html
https://github.com/elastic/beats/blob/master/libbeat/template/processor.go#L143

tsg: I'd say that's ok, they won't be indexed but are still returned in `source`. We could consider making them `text`, but not sure how "searchable" the content is anyway.

PR #5660, opened 2017-11-21
Fix autodiscovery config hashing

ruflin:
The config hash was created based on a cfg object instead of the unpacked object. This has the affect that all hashs are equal and new objects are not loaded. We already hit the same issue in reloading and have the same fix there.

@exekias FYI

### DIFF
--- a/libbeat/autodiscover/autodiscover.go
+++ b/libbeat/autodiscover/autodiscover.go
@@ -106,7 +106,11 @@ func (a *Autodiscover) startWorker() {
 
 		meta := getMeta(event)
 		for _, config := range configs {
-			hash, err := hashstructure.Hash(config, nil)
+
+			rawCfg := map[string]interface{}{}
+			err := config.Unpack(rawCfg)
+
+			hash, err := hashstructure.Hash(rawCfg, nil)
 			if err != nil {
 				logp.Debug(debugK, "Could not hash config %v: %v", config, err)
 				continue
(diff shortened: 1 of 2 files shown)


libbeat/autodiscover/autodiscover.go
```diff
@@ -106,7 +106,11 @@ func (a *Autodiscover) startWorker() {
 
 		meta := getMeta(event)
 		for _, config := range configs {
-			hash, err := hashstructure.Hash(config, nil)
+
```
andrewkroh: Is this newline intentional?

libbeat/autodiscover/autodiscover_test.go
```diff
@@ -186,3 +187,60 @@ func TestAutodiscover(t *testing.T) {
...
+	// Test start event
+	eventBus.Publish(bus.Event{
+		"start": true,
+		"meta": common.MapStr{
+			"foo": "bar",
+		},
+	})
+	time.Sleep(10 * time.Millisecond)
+	runners := adapter.Runners()
+	assert.Equal(t, len(runners), 2)
```
andrewkroh: I don't fully understand what's being tested here so I'll just ask if it it needs to be asserting anything on the second runner.

libbeat/autodiscover/autodiscover.go
ruflin: no. Will remove it for you ;-)

libbeat/autodiscover/autodiscover_test.go
ruflin: The main thing tested here is that 2 runner are running. Without the previous changes, only 1 runner was running. I will add the checks below also for runner 2 to be sure.

PR #5663, opened 2017-11-21
Support mysql prepare statement command 

ggg-geyi:
#5608


tsg:
@ggg-geyi It seems you have issues with the python virtualenv. Please double check that you meet the requirements from [here](https://www.elastic.co/guide/en/beats/devguide/current/beats-contributing.html#update-scripts). Note that you need python2, I'm not sure if python3 works.

ggg-geyi:
@tsg I run "make testsuite" in packetbeat directory, it seems run ok.
 I did not change anything in metricbeat, but I run "make testsuite" in metricbeat directory, it failed.

tsg:
@ggg-geyi weird, but we can ignore it for the purpose of this PR, the Metricbeat test failures are likely unrelated.

tsg:
@ggg-geyi Could you add a system test like this one, please: https://github.com/elastic/beats/blob/master/packetbeat/tests/system/test_0001_mysql_spaces.py#L6

ggg-geyi:
@tsg sure

packetbeat/protos/mysql/mysql.go
```diff
@@ -667,8 +785,161 @@ func (mysql *mysqlPlugin) receivedMysqlResponse(msg *mysqlMessage) {
...
+	// flags
+	offset++
+	// iterations
+	offset += 4
+	// null-bitmap
+	if nparam > 0 {
+		offset += (nparam + 7) / 8
+	}
+	// stmt bound
+	stmtBound := data[offset]
```
tsg: Is `data` guaranteed to have at least `offset` bytes here? If not, a check should be made before accessing to avoid "index out of bound" panics.

packetbeat/protos/mysql/mysql.go
```diff
@@ -667,8 +785,161 @@ func (mysql *mysqlPlugin) receivedMysqlResponse(msg *mysqlMessage) {
...
+	offset++
+	paramOffset := offset
+	if stmtBound == 1 {
+		paramOffset += nparam * 2
```
tsg: Probably you can add a check that `len(data) >= paramOffset` here.

packetbeat/protos/mysql/mysql.go
```diff
@@ -667,8 +785,161 @@ func (mysql *mysqlPlugin) receivedMysqlResponse(msg *mysqlMessage) {
...
+	}
+
+	for stmtPos := 0; stmtPos < nparam; stmtPos++ {
+		paramType = nparamType[stmtPos]
+		// dissect parameter on paramType
+		switch paramType {
+		// FIELD_TYPE_TINY
+		case 0x01:
+			paramOffset++
+			valueString := strconv.Itoa(int(data[paramOffset]))
```
tsg: Also before these accesses there should be checks that data has at least `paramOffset` bytes.

packetbeat/protos/mysql/mysql.go
ggg-geyi: You are right, a check should be made before accessing the data because some illegal packet will conflict with the protocol specification

packetbeat/protos/mysql/mysql.go
ggg-geyi: protocol fuzzing test or malicious attack will cause the exception

packetbeat/protos/mysql/mysql.go
```diff
@@ -18,7 +19,10 @@ import (
 
 // Packet types
 const (
-	mysqlCmdQuery = 3
+	mysqlCmdQuery       = 3
+	mysqlCmdStmtPrepare = 22
+	mysqlCmdStmtExecute = 23
+	mysqlCmdStmtClose   = 25
```
urso: Some more packet types to consider:

StmtBulkExecute: https://mariadb.com/kb/en/library/com_stmt_bulk_execute/
StmtFetch: https://mariadb.com/kb/en/library/com_stmt_fetch/

packetbeat/protos/mysql/mysql.go
```diff
@@ -58,6 +62,10 @@ type mysqlMessage struct {
 	cmdlineTuple *common.CmdlineTuple
 	raw          []byte
 	notes        []string
+
+	statementID       int
+	numberOfParameter int
+	param             string
```
urso: why not `param []string`? The execute statement contains an array of parameters.

packetbeat/protos/mysql/mysql.go
```diff
@@ -160,6 +175,13 @@ func (mysql *mysqlPlugin) init(results protos.Reporter, config *mysqlConfig) err
 		mysql.transactionTimeout,
 		protos.DefaultTransactionHashSize)
 	mysql.transactions.StartJanitor(mysql.transactionTimeout)
+
+	// preparement cache
+	mysql.preparestatements = common.NewCache(
+		mysql.preparestatementTimeout,
+		protos.DefaultTransactionHashSize)
+	mysql.preparestatements.StartJanitor(mysql.preparestatementTimeout)
```
urso: Do we really want to timeout prepared statements? Maybe have timeout only active on prepared statements after packet loss only (and only for N seconds). If a prepared statement is lost in the cache, we can not emit the original query in the event.

MySql/MariaDB itself enforces a limit on total statements one can use. So application using large number of prepared statements either create/close statements quite often or use an internal cache of active prepared statements.

packetbeat/protos/mysql/mysql.go
```diff
@@ -217,7 +256,8 @@ func mysqlMessageParser(s *mysqlStream) (bool, bool) {
 			if m.seq == 0 {
 				// starts Command Phase
 
-				if m.typ == mysqlCmdQuery {
+				if m.typ == mysqlCmdQuery || m.typ == mysqlCmdStmtPrepare ||
+					m.typ == mysqlCmdStmtExecute || m.typ == mysqlCmdStmtClose {
```
urso: in case list of supported packet types keeps increasing we should introduce a helper function like:

```
func isRequest(type) (request bool, supported bool) {}
```

packetbeat/protos/mysql/mysql.go
```diff
@@ -311,8 +357,25 @@ func mysqlMessageParser(s *mysqlStream) (bool, bool) {
...
 				len(s.data[s.parseOffset:]))
-			return true, true
 
+			// PREPARE_OK packet for Prepared Statement
+			// a trick for classify special OK packet
+			if m.isOK && m.packetLength == 12 {
+				idPtr := s.data[m.start+5 : m.start+13]
+				m.statementID = int(idPtr[0]) | int(idPtr[1])<<8 | int(idPtr[2])<<16 | int(idPtr[3])<<24
+				m.numberOfFields = int(idPtr[4]) | int(idPtr[5])<<8
+				m.numberOfParameter = int(idPtr[6]) | int(idPtr[7])<<8
```
urso: one can use `binary.BigEndian.UintX` here.

packetbeat/protos/mysql/mysql.go
```diff
@@ -603,9 +666,38 @@ func (mysql *mysqlPlugin) receivedMysqlRequest(msg *mysqlMessage) {
...
+			if value, ok := stmts[trans.statementID]; ok {
+				trans.query = value.query
+				// parse parameters
+				parameters := mysql.parseMysqlExecuteStatement(msg.raw, value)
+				trans.param = strings.Join(parameters, "#")
+				logp.Debug("mysqldetailed", "parameters: %s", trans.param)
+			} else {
+				logp.Debug("mysqldetailed", "Request execute statement from unknown prepare statement ID. Ignoring.")
+				mysql.transactions.Delete(tuple.Hashable())
+				return
```
urso: As packetbeat can be started at any time, there is quite a chance of packetbeat not knowing about a prepared statement. Plus, the statement cache might already have deleted a still active statement. The prepare command should become a document being published as well. Then we can at least create publish an execute statement (maybe optional) with statement ID + Note warning about the actual query being used is unknown. This information is still helpful as we can still build metrics/visualizations on number of execute statements and actual statement prepare/close (statement churn).

Some application use prepared statements for security purposes, basically creating and closing statements like crazy, all the time. Others might cache statements. It's a great source of information to learn about the applications behavior.

packetbeat/protos/mysql/mysql.go
```diff
@@ -650,6 +742,32 @@ func (mysql *mysqlPlugin) receivedMysqlResponse(msg *mysqlMessage) {
...
+				stmtData := &mysqlStmtData{
+					query:           trans.query,
+					numOfParameters: msg.numberOfParameter,
+				}
+				stmts[msg.statementID] = stmtData
+			}
+			mysql.preparestatements.Put(msg.tcpTuple.Hashable(), stmts)
+		}
+		// not publish prepare statement
+		mysql.transactions.Delete(trans.tuple.Hashable())
```
urso: I'd argue we want prepare/close statement publishing to be optional :)

packetbeat/protos/mysql/mysql.go
urso: +1 on offset check.

Packetbeat being a passive sniffer also has to deal with packet loss and being started late. That is, protocol analyzers in Packetbeat need  to sync and re-sync every now and so often to the actual TCP stream. Not being in sync leads to parsing errors. The other protocols drop state on parsing error and retry. Until we eventually find a packet starting with a query. Depending on query rates (lower rates are more likely to sync on), this can take a while. In order to not generate panics for the time the analyzer is not in sync yet, better have the analyzer somewhat robust.

packetbeat/protos/mysql/mysql.go
```diff
@@ -667,8 +785,205 @@ func (mysql *mysqlPlugin) receivedMysqlResponse(msg *mysqlMessage) {
...
+			valueString := strconv.Itoa(int(data[paramOffset]) | int(data[paramOffset+1])<<8)
+			paramString = append(paramString, valueString)
+			paramOffset += 2
+		// FIELD_TYPE_LONG
+		case 0x03:
+			if dataLen < paramOffset+4 {
+				logp.Debug("mysql", "Data too small")
+				return []string{}
+			}
+			valueString := strconv.Itoa(int(data[paramOffset]) | int(data[paramOffset+1])<<8 | int(data[paramOffset+2])<<16 | int(data[paramOffset+3])<<24)
```
urso: use helpers from binary package.

packetbeat/protos/mysql/mysql.go
ggg-geyi: yes, the execute statement contains an array of parameters.
actually I joint the array of parameters with “#” for easy in my circumstance

packetbeat/protos/mysql/mysql.go
ggg-geyi: I create a statement map for each flow in CACHE.
All the prepare statements of a flow will be cached in a  same map.
So there will be more and more maps in CACHE. I can delete the map when receive the Fin packet of a flow, but sometimes Fin packet lost or not be sniffed.
So I add the timeout method to control the size of the CACHE.
The cache expiration time will be overwritten after each put or get method.

packetbeat/protos/mysql/mysql.go
ggg-geyi: introduce a helper function will be better

packetbeat/protos/mysql/mysql.go
ggg-geyi: It'a big problem because packetbeat can be started at anytime and the mysql server to be audited is on for a long time. Plus, packetbeat should be updated and restarted occasionally.
So there is a quite a  chance of not knowing about a prepared statement.
I also cache the prepared statements in a remote redis server and will read back when restart packetbeat. 
You are right, it's a good choice to publish it instead of ignoring the event.

packetbeat/protos/mysql/mysql.go
ggg-geyi: Can be optional:)

packetbeat/protos/mysql/mysql.go
ggg-geyi: +1

ggg-geyi: Actually when calling this function, it has guaranteed this is a full length mysql packet according to the mysql header.But offset check will better have the analyzer robust.

packetbeat/protos/mysql/mysql.go
ggg-geyi: I will consider these types
I did not capture any packet of these types so I just let it go：）

packetbeat/protos/mysql/mysql.go
ggg-geyi: it's little-endian format

ruflin:
I think this change will also need a CHANGELOG entry.

@urso / @tsg Could you have a look again?

packetbeat/protos/mysql/mysql.go
urso: What's the reason to separate paramters by '#'? Can the '#' symbol be part of a string parameter, getting you a false-positive separator char?

hm... I'd rather see it being `[]string`. Or introduce a setting to allow for both styles.

packetbeat/protos/mysql/mysql.go
urso: makes sense.

swift1911:
is there any news about the feature?

packetbeat/protos/mysql/mysql.go
ggg-geyi: `[]string` better

ggg-geyi:
@tsg Could you have a look again?

swift1911:
is there any news about the feature?

packetbeat/protos/mysql/mysql.go
```diff
@@ -667,8 +799,207 @@ func (mysql *mysqlPlugin) receivedMysqlResponse(msg *mysqlMessage) {
...
+		return []string{}
+	}
+	stmtBound := data[offset]
+	offset++
+	paramOffset := offset
+	if stmtBound == 1 {
+		paramOffset += nparam * 2
+		if dataLen <= paramOffset {
+			logp.Debug("mysql", "Data too small to contain parameters")
+			return []string{}
```
adriansr: Nit: is simpler to `return nil`. Nil is always a valid empty slice.

adriansr:
@ggg-geyi do you mind if I take your changes and clean them up so we can finally merge this feature?

ggg-geyi:
@adriansr OK

ggg-geyi:
I am trying to fix the conflict and compare across forks.But these commits confusing me.

adriansr:
@ggg-geyi I submitted a new PR with your changes correctly rebased into the current master: https://github.com/elastic/beats/pull/8084

Can you check to see if it's OK?

ggg-geyi:
@adriansr I will check right now. Thanks a lot. I have learned something about git rebase.>-<

andrewkroh:
I'm closing this PR since there is a follow-up in #8084 that incorporates this work.

Thanks @ggg-geyi .

PR #5671, opened 2017-11-21
[FileBeat] Adding SSL log fileset to Apache2 and new module for Win IIS

BongoEADGC6:
- Default ssl_request.log Is not processed by the access fileset in the apache2 module and is different enough to require one. I've created a new fileset in the module to allow that.

- Windows did not have an IIS log module as well. I have setup one that will process default w3svc logs. 

Todo:
[ ] - IIS: Will also include a grok for checkboxes 
[ ] - IIS: Needs a kibana dashboard


ruflin:
@BongoEADGC6 Thanks a lot for your contribution. Any chance you could split this PR up in two parts. One for the additional apache fileset and one for the new module? Like this the PR's are getting smaller, easier to review and one does not block the other.

BongoEADGC6:
No problem at all, sir. Give me a little bit.

PR #5673, opened 2017-11-21
Show config for passing credentials rather than command line

dedemorton:
In previous releases, we showed how to set credentials from the command line because it simplified the getting started experience, but this introduced other problems (credentials were less secure), so this PR changes the getting started docs to show how to configure credentials in the config file. This PR also fixes https://github.com/elastic/beats/pull/4867

These changes make the asciidoc source *a lot* simpler because we can remove the conditional coding now that we tell everyone to use the config file.

Note that I have another PR open that affects the topic about loading dashboards: https://github.com/elastic/beats/pull/5635


ruflin:
@dedemorton Needs a rebase

dedemorton:
@ruflin I've rebased

Issue #5422, opened 2017-10-23
Refactoring Auditbeat Modules


andrewkroh:
We want to do some refactoring to change the config and event schema before the Auditbeat GA release.  Auditbeat reuses Metricbeat's framework, and the framework's orientation around metrics is visible to the Auditbeat user, but it shouldn't be.

### Goals

- Simplify configuration for the end-user
  - Modules should map clearly to events the users want to monitor.
- Divorce from the Metricbeat event data model
  - The `metricset` namespace isn't necessary here.
- Abstract the data source from module
  - Users shouldn't need to know how the data is collected
  - If needed an advanced user should be able to control what data sources are used.
  - For example have a module for monitoring what processes are running or what sockets are active rather than a module oriented around a data source like the kernel's audit framework or eBPF.
- Provide specialized dashboards for each module.

### Steps

1. Modify the Metricbeat framework to allow for customization of the event builder (the builder is responsible for adding the `metricset` field to all events).
1. Reorganize the config to allow for modules without "metricsets".
1. Rename the existing audit/kernel metricset to the `auditd` module. (As we add modules I expect this to be used less.)
1. Rename the existing audit/file metricset to `file_integrity`.

### Sample Configs

Note the removal of `metricsets: []` and the removal of the module name prefixes from the config options.

```
auditbeat.modules:
- module: file_integrity
  paths:
  - /bin
  - /usr/bin
  - /sbin
  - /usr/sbin
  - /etc
  scan_at_start: true
  scan_rate_per_sec: 50 MiB
  max_file_size: 100 MiB
  hash_types: [sha1]
```

```
auditbeat.modules:
- module: auditd
  resolve_ids: true
  failure_mode: silent
  backlog_limit: 8196
  rate_limit: 0
  include_raw_message: false
  include_warnings: false
```

ruflin:
Separation of data source from modules reminds me a lot of prospectors types and modules in Filebeat. Would be great to sync up on this one for Filebeat to make sure the configuration options are similar.

PR #5675, opened 2017-11-21
Allow specifying default metricsets

andrewkroh:
This is useful for modules containing a single metricset where you don't need the granularity of multiple metricsets. If `metricsets` isn't found in the config then the default MetricSets will be instantiated. If no defaults are defined for a module then an error is shown to the user (as was the case before this change when no `metricsets` were configured).

```
func init() {
	mb.Registry.MustAddMetricSet(moduleName, metricsetName, New,
		mb.DefaultMetricSet(),
		mb.WithHostParser(parse.EmptyHostParser),
	)
}
```

### DIFF
--- a/metricbeat/mb/registry.go
+++ b/metricbeat/mb/registry.go
@@ -42,6 +42,27 @@ type metricSetFactoryInfo struct {
 	name       string
 	factory    MetricSetFactory
 	hostParser HostParser
+	isDefault  bool
+}
+
+// MetricSetOption sets an option for a MetricSetFactory that is being
+// registered.
+type MetricSetOption func(info *metricSetFactoryInfo)
+
+// WithHostParser specifies the HostParser that should be used with the
+// MetricSet.
+func WithHostParser(p HostParser) MetricSetOption {
+	return func(i *metricSetFactoryInfo) {
+		i.hostParser = p
+	}
+}
+
+// DefaultMetricSet specifies that the MetricSetFactory will be the default
+// when no MetricSet names are specified in the configuration.
+func DefaultMetricSet() MetricSetOption {
+	return func(i *metricSetFactoryInfo) {
+		i.isDefault = true
+	}
 }
 
 // Register contains the factory functions for creating new Modules and new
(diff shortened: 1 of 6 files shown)


andrewkroh:
This is be used by Auditbeat to simplify the configuration. It's part of the work for #5422.

metricbeat/mb/registry.go
```diff
@@ -94,7 +115,28 @@ func (r *Register) AddModule(name string, factory ModuleFactory) error {
 // HostParser function for parsing the 'host' configuration data. An error is
 // returned if any parameter is empty or nil or if a factory has already been
 // registered under the name.
+//
+// Use MustAddMetricSet for new code.
```
ruflin: Does that mean we need to update all existing metricsets?

ruflin: UPDATE: Just read your comment again in the PR and does not seem to be the case.

ruflin:
@andrewkroh Does this need a review label?

andrewkroh:
@ruflin Yeah, I had forgotten to add it when I opened the PR.

andrewkroh:
The test failures on jenkins-windows are for filebeat and libbeat and appear to be unrelated.

metricbeat/mb/registry.go
```diff
@@ -94,7 +115,28 @@ func (r *Register) AddModule(name string, factory ModuleFactory) error {
...
+func (r *Register) MustAddMetricSet(module, name string, factory MetricSetFactory, options ...MetricSetOption) {
+	if err := r.addMetricSet(module, name, factory, options...); err != nil {
+		panic(err)
+	}
+}
+
+// addMetricSet registers a new MetricSetFactory. An error is returned if any
+// parameter is empty or nil or if a factory has already been registered under
+// the name.
+func (r *Register) addMetricSet(module, name string, factory MetricSetFactory, options ...MetricSetOption) error {
```
ruflin: I like that we now have `options` instead of the more specific hostParser 👍

metricbeat/mb/registry.go
```diff
@@ -146,20 +190,47 @@ func (r *Register) metricSetFactory(module, name string) (MetricSetFactory, Host
 	module = strings.ToLower(module)
 	name = strings.ToLower(name)
 
-	modules, exists := r.metricSets[module]
+	metricSets, exists := r.metricSets[module]
```
ruflin: Good catch

metricbeat/mb/registry.go
```diff
@@ -172,14 +243,14 @@ func (r *Register) Modules() []string {
...
-//MetricSets returns the list of metricsets registered for a given module
+// MetricSets returns the list of MetricSets registered for a given module
 func (r *Register) MetricSets(module string) []string {
 	r.lock.RLock()
 	defer r.lock.RUnlock()
 
 	var metricsets []string
 
-	sets, ok := r.metricSets[module]
+	sets, ok := r.metricSets[strings.ToLower(module)]
```
ruflin: I wonder if we should add a test somewhere to make sure this is not missing in other parts of the code.

andrewkroh:
> I assume if two metricsets are defined as default in a module, both are loaded?

That's correct, there can be more than one metricset marked as default.

PR #5678, opened 2017-11-21
Add auth fileset options

dedemorton:
The first commit adds the missing description for 6.0

This has been fixed in master with https://github.com/elastic/beats/commit/e125cf1daa98c7696ac42a2adf0f150b97a26dbf#diff-e429a0e021a0f063055c4128b317741b

The second commit updates the compatibility statement. This needs to a forward port.

### DIFF
--- a/filebeat/module/system/_meta/docs.asciidoc
+++ b/filebeat/module/system/_meta/docs.asciidoc
@@ -63,3 +63,7 @@ include::../include/config-option-intro.asciidoc[]
 
 include::../include/var-paths.asciidoc[]
 
+[float]
+==== `auth` fileset settings
+
+include::../include/var-paths.asciidoc[]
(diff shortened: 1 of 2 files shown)


dedemorton:
@tsg I've also updated the compatibility statement. Is this correct? I always install those plugins right away when I set up a new ES install, so I'm not sure. This is related to an old issue that Tanya created: https://github.com/elastic/beats/issues/4567

I'd like to get that one closed out.

I plan to make a full pass at all the docs to clear up the use of sudo in a separate PR. Other than that, I think we've addressed all of her concerns piecemeal over the past several months, no?

filebeat/module/system/_meta/docs.asciidoc
```diff
@@ -13,6 +13,10 @@ include::../include/what-happens.asciidoc[]
 This module was tested with logs from OSes like Ubuntu 12.04, Centos 7, and
 macOS Sierra.
 
+This module requires the
```
ruflin: Looking at the code it seems this one only requires the geoip and no the user agent. But TBH I think we should make the two a general requirement for all modules to make it simpler for the user to always install just both.

ruflin:
@dedemorton I think this PR needs a `make update`. Besides that LGTM.

dedemorton:
@ruflin Ran make update as requested.

PR #5680, opened 2017-11-22
[WIP] Add testing of template

ruflin:
@exekias I wanted to use `metricbeat setup --template` to make sure we directly know in the future when we break a template for ES. Unfortunately the below code does not work because it seems `setup` is not part of the testing binary (or is swallowed somehow). Any idea here?


exekias:
It was just metricbeat refuses to start when no module is configured, I've opened a new PR, will add tests for the rest of beats: https://github.com/elastic/beats/pull/5682

PR #5682, opened 2017-11-22
Test all beats can setup their template

exekias:
This will ensure we don't introduce errors in our fields definitions


auditbeat/tests/system/test_base.py
```diff
@@ -23,3 +25,24 @@ def test_start_stop(self):
         assert self.log_contains("Setup Beat: auditbeat")
         assert self.log_contains("auditbeat start running")
         assert self.log_contains("auditbeat stopped")
+
+    def test_template(self):
```
andrewkroh: I think this should have a `@unittest.skipUnless(INTEGRATION_TESTS, "integration test")`?

ruflin:
Seems this failed for metricbeat on travis:

```
FAIL: Test that the template can be loaded with `setup --template`
  File "/go/src/github.com/elastic/beats/metricbeat/tests/system/test_base.py", line 48, in test_template
    assert exit_code == 0
AssertionError
```

@exekias Thanks for taking this one over.

exekias:
it seems `TEST_ENVIRONMENT` is always `false` for packetbeat, will need to add a `docker-compose.yml` for this

ruflin:
@exekias Should we skip packetbeat for now in this PR and add it in a follow up PR?

exekias:
Looks good to me, I'm afraid I have little bandwidth at the moment, could get something by the end of the week, if everything goes well, but merging now sounds safer

ruflin:
@exekias To get this in now (which I think we should do) you need to remove the packetbeat commit again to get it to green.

exekias:
Applied your comment and rebased with master to get latest tests fixes

metricbeat/module/elasticsearch/_meta/env
```diff
@@ -1,2 +1,2 @@
-ELASTICSEARCH_HOST=elasticsearch
-ELASTICSEARCH_PORT=9200
+ES_HOST=elasticsearch
```
ruflin: Is there a specific reason you changed this? I kept it identical to the module name as I initially intended to do some auto generation here. Will merge it as is but curious if there were some issues with the longer version.

exekias: yes, it seems we use ES_HOST across all beats here: https://github.com/elastic/beats/blob/master/libbeat/tests/system/beat/beat.py#L546

PR #5686, opened 2017-11-22
Dashboard loading improvements and fixes

urso:
  install via kibana only
  enabled

### DIFF
--- a/auditbeat/auditbeat.reference.yml
+++ b/auditbeat/auditbeat.reference.yml
@@ -722,7 +722,8 @@ output.elasticsearch:
 # dashboards and index pattern. Example: testbeat-*
 #setup.dashboards.index:
 
-# Force loading of dashboards using the Kibana API without querying Elasticsearch for the version
+# Always use the Kibana API for loading the dashboards instead of autodetecting
+# how to install the dashboards by first querying Elasticsearch.
 #setup.dashboards.always_kibana: false
 
 #============================== Template =====================================
--- a/auditbeat/auditbeat.reference.yml
+++ b/auditbeat/auditbeat.reference.yml
@@ -723,7 +723,7 @@ output.elasticsearch:
 #setup.dashboards.index:
 
 # Force loading of dashboards using the Kibana API without querying Elasticsearch for the version
-#always_kibana: false
+#setup.dashboards.always_kibana: false
 
 #============================== Template =====================================
 
--- a/libbeat/dashboards/dashboards.go
+++ b/libbeat/dashboards/dashboards.go
@@ -1,72 +1,115 @@
 package dashboards
 
 import (
+	"errors"
 	"fmt"
 	"path/filepath"
 	"strconv"
 	"strings"
 
 	"github.com/elastic/beats/libbeat/common"
-	"github.com/elastic/beats/libbeat/logp"
 )
 
-func ImportDashboards(beatName, hostname, homePath string,
-	kibanaConfig *common.Config, esConfig *common.Config,
-	dashboardsConfig *common.Config, msgOutputter MessageOutputter) error {
-
+// ImportDashboards tries to import the kibana dashboards.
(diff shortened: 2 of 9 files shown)


auditbeat/auditbeat.reference.yml
```diff
@@ -723,7 +723,7 @@ output.elasticsearch:
 #setup.dashboards.index:
 
 # Force loading of dashboards using the Kibana API without querying Elasticsearch for the version
```
andrewkroh: WDYT about rewording this to not include "force"? @kvch 

```
# Always use the Kibana API for loading the dashboards instead of autodetecting
# how to install the dashboards by first querying Elasticsearch.
```

libbeat/dashboards/dashboards.go
```diff
@@ -1,72 +1,115 @@
...
+		if !kibanaConfig.HasField("username") && username != "" {
+			kibanaConfig.SetString("username", -1, username)
+		}
+		if !kibanaConfig.HasField("password") && password != "" {
+			kibanaConfig.SetString("password", -1, password)
+		}
+	}
+
+	// check import route
+	const (
```
kvch: I would move these constants out of the functions. It seems weird to have enum-like constants when the whole file is only about importing dashboards.

libbeat/dashboards/dashboards.go
```diff
@@ -1,72 +1,115 @@
...
 )
 
-func ImportDashboards(beatName, hostname, homePath string,
-	kibanaConfig *common.Config, esConfig *common.Config,
-	dashboardsConfig *common.Config, msgOutputter MessageOutputter) error {
-
+// ImportDashboards tries to import the kibana dashboards.
+// If the Elastic Stack is at version 6.0+, the dashboards should be installed
+// via the kibana dashboard loader plugin. For older versions of the Elastic Stack
+// we write the dashboards right into the .kibana index.
```
andrewkroh: I suggest `s/right/directly/` to avoid having two homophones in the sentence. Thank you for adding docs.

libbeat/dashboards/dashboards.go
```diff
@@ -1,72 +1,115 @@
...
+	}
+
+	// check import route
+	const (
+		importNone = iota
+		importViaKibana
+		importViaES
+	)
+
+	var (
```
kvch: These unnecessarily indented lines feel quite odd during reading the code.

andrewkroh: I tend to like this style when there is a group of var declarations.

auditbeat/auditbeat.reference.yml
kvch: Yes, much better. The name of the option and its functionality changed slightly since I added those line. :)

libbeat/dashboards/dashboards.go
kvch: Yes, it is strongly a personal preference from my side, thus the wording "feel". I am ok with leaving these lines as is.

libbeat/dashboards/dashboards.go
```diff
@@ -1,72 +1,115 @@
...
+
+	requiresKibana := dashConfig.AlwaysKibana || !esConfig.Enabled()
+	if requiresKibana {
+		importVia = useKibana
+	} else {
+		// Check import route via elasticsearch version. If Elasticsearch major
+		// version is >6, we assume Kibana also being at versions >6.0. In this
+		// case dashboards will be imported using the new kibana dashboard loader
+		// plugin.
+		// XXX(urso): Why do we test the Elasticsearch version? If kibana is
```
ruflin: @urso agree, we should be able to get the version out of the Kibana API probably even before 6.

ruflin:
I added the backport label for it for now as I think we need this in 6.1. For further improvements like the TODO from @urso in the code I think we can do as follow ups?

urso:
Jenkins, test it!

PR #5693, opened 2017-11-22
Enable specifying pip install options

jasontedor:
If a binary wheel is available, `pip` will prefer that to building from source. Yet some packagers prefer to build everything from source. For this, the ability to specify the `--no-binary` flag is needed for `pip`. This commit enables this possibility by enabling the `PIP_INSTALL_PARAMS` environment variable to be specified so that packagers can specify `PIP_INSTALL_PARAMS="--no-binary :all:"`.


jasontedor:
This is desired by Homebrew where it is desired that all dependencies, even build-time dependencies, be built from source.

ruflin:
Merged as failures do not look related. @jasontedor Thanks.

jasontedor:
Thanks @ruflin. Can this be backported to at least 6.x?

ruflin:
@jasontedor Sure. We plan to merge master into 6.x anyways again in the near future, so it will make it in all upcoming 6.x releases.

PR #5695, opened 2017-11-23
Remove full kibana directory when collecting kibana files

ruflin:
Previously the index pattern directory was not removed because it was created manually. Now that this is generated out of the fields.yml the complete fields.yml directory can be removed for cleanup.

This PR comes out of https://github.com/elastic/beats/pull/5328 to make sure this change has not side effects. Needs to be merged before https://github.com/elastic/beats/pull/5328.


metricbeat/Makefile
```diff
@@ -21,7 +21,7 @@ include ${ES_BEATS}/libbeat/scripts/Makefile
 .PHONY: kibana
 kibana:
 	@# To not remove index-pattern as generated by update
```
ph: I think we can remove that comment?

ruflin: removed

PR #5697, opened 2017-11-23
Cherry-pick: Dashboard loading improvements and fixes (#5686)

kvch:
* Dashboard loading improvements and fixes

  install via kibana only
  enabled

* Review updates

(cherry picked from commit 554701f0a04fe25229be4afd6cccd745f2844860)


ruflin:
@kvch Should we also add this to 6.1 (if not already the case)?

Issue #5662, opened 2017-11-21
Steps in "Creating a Beat based on Metricbeat" [6.0] fails


shino:
For confirmed bugs, please report:
- Version: 6.0
- Operating System: Mac (darwin)
- Steps to Reproduce: 

```
% go get github.com/elastic/beats/metricbeat
% cd $GOPATH/src/github.com/elastic/beats
% git checkout 6.0
% mkdir -p $GOPATH/src/github.com/shino
% cd $GOPATH/src/github.com/shino
% python ${GOPATH}/src/github.com/elastic/beats/script/generate.py --type=metricbeat
Beat Name [Examplebeat]: Testbeat
Your Github Name [your-github-name]: shino
Beat Path [github.com/shino/testbeat]:
Firstname Lastname: shino
% cd testbeat

(py27) % make setup
mkdir -p vendor/github.com/elastic/
cp -R /Users/shino/go2/src/github.com/elastic/beats vendor/github.com/elastic/
rm -rf vendor/github.com/elastic/beats/.git
make create-metricset
make[1]: warning: jobserver unavailable: using -j1.  Add '+' to parent make rule.
make[1]: Entering directory '/Users/shino/go2/src/github.com/shino/testbeat'
New python executable in /Users/shino/go2/src/github.com/shino/testbeat/build/python-env/bin/python
Installing setuptools, pip, wheel...done.
Module name: test1
Metricset name: set1
Module test1 created.
Metricset set1 created.
make[1]: Leaving directory '/Users/shino/go2/src/github.com/shino/testbeat'
make collect
make[1]: warning: jobserver unavailable: using -j1.  Add '+' to parent make rule.
make[1]: Entering directory '/Users/shino/go2/src/github.com/shino/testbeat'
find: warning: you have specified the -maxdepth option after a non-option argument -type, but options are not positional (-maxdepth affects tests specified before it as well as those specified after it).  Please specify options before other arguments.

find: warning: you have specified the -mindepth option after a non-option argument -type, but options are not positional (-mindepth affects tests specified before it as well as those specified after it).  Please specify options before other arguments.

cp: cannot stat 'module/*/_meta/kibana': No such file or directory
make[1]: [/Users/shino/go2/src/github.com/elastic/beats/metricbeat/Makefile:25: kibana] Error 1 (ignored)
make[1]: Leaving directory '/Users/shino/go2/src/github.com/shino/testbeat'
```

[...]

PR #5699, opened 2017-11-24
Bugfix/use make macro

shino:
Arose from the conversation at [1]

This PR replaces direct `make` call in Makefiles by `$(MAKE)` macro.
It suppress warning message and it is good way to use make (cf [2].)

Note:

- The 2nd commit is logically independent of `$(MAKE)` fix, but it is needed to execute
  `make test` in `generator/` because I use GNU toolset on darwin. If it should be
  separate PR or abandoned, it's OK.
- The issue [1] originally found in 6.0 branch. I'm not sure what branching strategy this
  repository uses, please let me know if this PR should go to 6.0 branch.

[1] https://github.com/elastic/beats/issues/5662#issuecomment-346215133
[2] https://www.gnu.org/software/make/manual/html_node/MAKE-Variable.html#MAKE-Variable

Thanks.

### DIFF
--- a/generator/beat/{beat}/Makefile
+++ b/generator/beat/{beat}/Makefile
@@ -15,7 +15,7 @@ NOTICE_FILE=NOTICE
 # Initial beat setup
 .PHONY: setup
 setup: copy-vendor
-	make update
+	$(MAKE) update
 
 # Copy beats into vendor directory
 .PHONY: copy-vendor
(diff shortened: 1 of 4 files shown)


ruflin:
jenkins, test it

ruflin:
@shino About the branching. I suggest to not backport this to 6.0 as we mostly only backport bugs.

shino:
@ruflin Thanks for review and comment to my question 😄

PR #5702, opened 2017-11-24
Change pipeline delimiter to {< and >}

### LINKED ISSUE #5701: Filebeat modules: ingest pipeline parsing issue
It looks like there is a recent code change that is causing some issues with parsing certain patterns in ingest pipeline configs in Filebeat.  After a bit of debugging, the following ingest pipeline config in a custom module will fail to be loaded into Elasticsearch:

```
{
  "description": "whatevs",
  "processors": [
    {
      "grok": {
        "field": "message",
        "patterns": [
          "\\{%{INT:whatever}"
        ]
      }
    }
  ]
}
```

The issue appears to be due to the combination of a pattern having a literal left curly brace `{`, followed by the start of a grok pattern (i.e., `%`) as evidenced by the code here:

https://github.com/elastic/beats/blob/master/filebeat/fileset/fileset.go#L234

This was working in 5.x but appears broken in master (6.1+ as of this issue).  The above pipeline is valid and can be PUT directly into Elasticsearch without error. 

With the above ingest pipeline definition in a Filebeat module, the following error is produced when attempting to PUT the pipeline into Elasticsearch from Filebeat:

```
2017/11/24 06:23:43.235683 output.go:74: ERR Failed to connect: Connection marked as failed because the onConnect callback failed: Error getting pipeline for fileset elasticsearch/log: Error interpreting the template of the ingest pipeline: Error parsing template {
  "description": "whatevs",
  "processors": [
    {
      "grok": {
        "field": "message",
        "patterns": [
          "\\{%{INT:whatever}"
        ]
      }
    }
  ]
}
: template: text:8: unexpected "{" in command
```

kvch:
The new delimiters are `{<` and `>}`. These are not likely to turn up regexes or Grok pattern. Regexes between `{` and `}` metacharacters include numbers and comma. Grok patterns include letters inside `%{` and `}`. Also, it is not likely that someone would compare anything which ends with `{` or starts with `}`.

Previously used `{%` and `%}` were more likely to cause problems as Grok patterns start with `%{`.

Fixes #5701

### DIFF
--- a/filebeat/fileset/fileset.go
+++ b/filebeat/fileset/fileset.go
@@ -226,12 +226,12 @@ func resolveVariable(vars map[string]interface{}, value interface{}) (interface{
 }
 
 // applyTemplate applies a Golang text/template. If specialDelims is set to true,
-// the delimiters are set to `{%` and `%}` instead of `{{` and `}}`. These are easier to use
+// the delimiters are set to `{<` and `>}` instead of `{{` and `}}`. These are easier to use
 // in pipeline definitions.
 func applyTemplate(vars map[string]interface{}, templateString string, specialDelims bool) (string, error) {
 	tpl := template.New("text")
 	if specialDelims {
-		tpl = tpl.Delims("{%", "%}")
+		tpl = tpl.Delims("{<", ">}")
 	}
 	tpl, err := tpl.Parse(templateString)
 	if err != nil {
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -51,6 +51,7 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
 - Fix race condition when limiting the number of harvesters running in parallel {issue}5458[5458]
 - Fix relative paths in the prospector definitions. {pull}5443[5433]
 - Fix `recursive_globe.enabled` option. {pull}5443[5443]
+- Change pipeline delimiter to `{<` and `>}`. {pull}5702[5702]
 
 *Heartbeat*
 
--- a/filebeat/module/system/auth/ingest/pipeline.json
+++ b/filebeat/module/system/auth/ingest/pipeline.json
@@ -32,7 +32,7 @@
 					"MMM  d HH:mm:ss",
 					"MMM dd HH:mm:ss"
         ],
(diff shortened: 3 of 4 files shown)


ph:
@kvch Oops, forgot 3 things:

- Can you add a changelog entry for that?
- If there any doc reference about that syntax?

kvch:
I added the PR to the Changelog.
I haven't found any documentation on this delimiter, thus I haven't changed anything in docs.

ph:
Merged @kvch

filebeat/fileset/fileset.go
```diff
@@ -226,12 +226,12 @@ func resolveVariable(vars map[string]interface{}, value interface{}) (interface{
 }
 
 // applyTemplate applies a Golang text/template. If specialDelims is set to true,
-// the delimiters are set to `{%` and `%}` instead of `{{` and `}}`. These are easier to use
+// the delimiters are set to `{<` and `>}` instead of `{{` and `}}`. These are easier to use
```
ruflin: I wonder if we should have just gone back to {{ ?

kvch:
It is not a breaking change, because it hasn't been released.

filebeat/fileset/fileset.go
kvch: I am not sure if we can use `{{` and `}}`, because there are two substitutions happening at two different times. See in https://github.com/elastic/beats/pull/5702/files#diff-1b64e79bf3f0b28f0a29965eb2bf0cecR35

ph: @ruflin like @kvch said, we will run into the double substitution issue.

ruflin:
Ok, I assume the `{%` logic is already in 6.0 Didn't know we just added that.

filebeat/fileset/fileset.go
ruflin: 👍

tsg:
Thanks for the fix @kvch!

vjsaisha:
@ruflin Can you teach me why we have  // in this patterns line for sudo(?:\\[%{POSINT:system.auth.pid}\\])  ? What is the use of // in grok pattern?

ruflin:
@vjsaisha Could you open a topic on discuss for this? https://discuss.elastic.co/c/beats/filebeat

PR #5707, opened 2017-11-24
Use `add_kubernetes_metadata` IP & port matching in Packetbeat

exekias:
The code was there for Metricbeat already, this PR moves it to libbeat
and adapts Packetbeat to use it, with good defaults.


libbeat/processors/add_kubernetes_metadata/indexers.go
```diff
@@ -248,3 +249,81 @@ func containerID(status PodContainerStatus) string {
...
+	return &IPPortIndexer{genMeta: genMeta}, nil
+}
+
+// GetMetadata returns metadata for the given pod, if it matches the index
+func (h *IPPortIndexer) GetMetadata(pod *Pod) []MetadataIndex {
+	commonMeta := h.genMeta.GenerateMetaData(pod)
+	hostPorts := h.GetIndexes(pod)
+	var metadata []MetadataIndex
+
+	if pod.Status.PodIP == "" {
```
ruflin: If a pod does not have an IP address, it means no metadata is availabe? Should this return an error?

CHANGELOG.asciidoc
```diff
@@ -122,6 +122,8 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
 
 *Packetbeat*
 
+- Configure good defaults for `add_kubernetes_metadata`. {pull}5707[5707]
```
ruflin: I would add 2 change entries. One to "applies to all beats" that the IpProcessor is now available for all beats (also filebeat for example) and the other one as above that IpProcessor is set as default for PB.

libbeat/processors/add_kubernetes_metadata/indexers.go
exekias: We wouldn't be dealing with this error anyway, it could make sense to log this, but I think it's just defensive code

CHANGELOG.asciidoc
exekias: good point! I added a new entry

exekias:
Just opened https://github.com/elastic/beats/pull/5721 as a follow-up, with more changes needed after doing som real life usage with Packetbeat and APM

PR #5714, opened 2017-11-26
Fix flaky node_stats Logstash tests

### LINKED ISSUE #5712: Flaky Logstash node_stats tests
Currently the Logstash node_stats test is flaky. It faily with the following error:

```
--- FAIL: TestFetch (2.74s)
	Error Trace:	node_stats_integration_test.go:21
	Error:		Received unexpected error "HTTP error 500 in node_stats: 500 Internal Server Error"
		
FAIL
```

Interestingly the `node` integration tests pass consistently without issue, meaning LS is running fine.  I wonder under which circumstances node_stats returns a 500. As it can't be reproduced manually I assume it's kind of a timing issue that `node_stats` is not ready or similar. Unfortunately there are no LS logs here for now.

This is running against Logstash master.

@elastic/logstash Any hints are appreciated here.

ruflin:
Update healthcheck to use _node/stats instead. As soon as this returns a 200 all endpoints should work as expected. See elastic/logstash#8004.

* Remove skipped tests
* Update LS image to latest stable version

Closes #5712

### DIFF
--- a/metricbeat/module/logstash/_meta/Dockerfile
+++ b/metricbeat/module/logstash/_meta/Dockerfile
@@ -1,4 +1,4 @@
-FROM docker.elastic.co/logstash/logstash:6.0.0-rc2
+FROM docker.elastic.co/logstash/logstash:6.0.0
 
 ENV XPACK_MONITORING_ENABLED=FALSE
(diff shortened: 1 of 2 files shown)


ruflin:
jenkins, test it

PR #5715, opened 2017-11-27
Fix README.md link in packages

ruflin:
The links in the README.md file are broken for the 6.0 release. The reason is that gotpl interprets 6.0 and makes 6 out of it. Add "..." around fixes the issue.


ruflin:
I opened this one against 6.0 as this is where it's needed most. We should forward port this to master as we will hit this issue again with 7.0.

ruflin:
Tested it locally and seems to fix the issue. Ready for review.

PR #5717, opened 2017-11-27
Create module overview page for Metricbeat

ruflin:
This creates a module overview page which lists each Module with its metricsets and the release stats of the Module and Metricsets. The goal is to give an simple overview over all modules and metricsets and what release stage they are in.

An Screenshot on how it looks can be found below:

screen shot 2017-11-27 at 15 58 45

@dedemorton Any input on how you would like to have this table or other format you would suggest?

### DIFF
--- a/metricbeat/docs/modules_list.asciidoc
+++ b/metricbeat/docs/modules_list.asciidoc
@@ -2,37 +2,122 @@
 This file is generated! See scripts/docs_collector.py
 ////
 
-  * <<metricbeat-module-aerospike,Aerospike>>
-  * <<metricbeat-module-apache,Apache>>
-  * <<metricbeat-module-ceph,Ceph>>
-  * <<metricbeat-module-couchbase,Couchbase>>
-  * <<metricbeat-module-docker,Docker>>
-  * <<metricbeat-module-dropwizard,Dropwizard>>
-  * <<metricbeat-module-elasticsearch,Elasticsearch>>
-  * <<metricbeat-module-etcd,Etcd>>
-  * <<metricbeat-module-golang,Golang>>
-  * <<metricbeat-module-graphite,Graphite>>
-  * <<metricbeat-module-haproxy,HAProxy>>
-  * <<metricbeat-module-http,HTTP>>
-  * <<metricbeat-module-jolokia,Jolokia>>
-  * <<metricbeat-module-kafka,Kafka>>
-  * <<metricbeat-module-kibana,Kibana>>
-  * <<metricbeat-module-kubernetes,Kubernetes>>
-  * <<metricbeat-module-logstash,Logstash>>
-  * <<metricbeat-module-memcached,Memcached>>
-  * <<metricbeat-module-mongodb,MongoDB>>
-  * <<metricbeat-module-mysql,MySQL>>
-  * <<metricbeat-module-nginx,Nginx>>
-  * <<metricbeat-module-php_fpm,PHP_FPM>>
-  * <<metricbeat-module-postgresql,PostgreSQL>>
-  * <<metricbeat-module-prometheus,Prometheus>>
-  * <<metricbeat-module-rabbitmq,RabbitMQ>>
-  * <<metricbeat-module-redis,Redis>>
-  * <<metricbeat-module-system,System>>
-  * <<metricbeat-module-vsphere,vSphere>>
-  * <<metricbeat-module-windows,Windows>>
-  * <<metricbeat-module-zookeeper,ZooKeeper>>
-
+[options="header"]
+|========================
+|Modules   |Metricsets   
+|<<metricbeat-module-aerospike,Aerospike>>  beta[]   |    
+.1+|   |<<metricbeat-metricset-aerospike-namespace,namespace>> beta[]  
+|<<metricbeat-module-apache,Apache>>     |    
+.1+|   |<<metricbeat-metricset-apache-status,status>>   
+|<<metricbeat-module-ceph,Ceph>>  beta[]   |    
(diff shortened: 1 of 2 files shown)


dedemorton:
@ruflin TBH, I struggle myself with how to format content like this because as a general rule we try not to use tables in the docs. Tables don't display well on mobile devices plus they sometimes cause text overruns in the TOC (but only when we have wide cells). It's a limitation of our doc tool chain.

But the cells in your table here are fairly narrow, so I think it's acceptable.

You can make the table a bit easier to read by spanning the rows where you have mutliple metricsets (see my example below) but this will complicate your logic for generating the table. 

Other feedback:

- I would change the name of the Release column. When I see the word "release," I expect to see a version number that corresponds to a product release. Maybe say "Release level" or "Release status" instead.

- Do we describe our release levels somewhere? If so, we should link to our descriptions. If not, I think we need to provide the descriptions on this page, or readers might now know the difference between beta and experimental.

- Minor: I would capitalize GA 'cause it looks weird in lowercase.

Here's an example of what the table with spanned rows would look like (scroll down past the screen shot to see the asciidoc code used to create the table):

image

In the following example, `.6+` tells asciidoc to span over 6 rows. In this case, I'm spanning the empty row. You could span the row that shows the module name instead of the empty row, but that looks weird to me because the "Release level" column also applies to the module.

````
[options="header"]
|================================
|Module   |Metricset   |Release level
|<<metricbeat-module-aerospike,Aerospike>>    |   |beta
|   |<<metricbeat-metricset-aerospike-namespace,namespace>>   |beta
|<<metricbeat-module-apache,Apache>>    |   |GA
|   |<<metricbeat-metricset-apache-status,status>>   |GA
|<<metricbeat-module-ceph,Ceph>>    |   |beta
.6+|   |<<metricbeat-metricset-ceph-cluster_disk,cluster_disk>>   |beta
|<<metricbeat-metricset-ceph-cluster_health,cluster_health>>   |beta
|<<metricbeat-metricset-ceph-cluster_status,cluster_status>>   |beta
|<<metricbeat-metricset-ceph-monitor_health,monitor_health>>   |beta
|<<metricbeat-metricset-ceph-osd_tree,osd_tree>>   |beta
|<<metricbeat-metricset-ceph-pool_disk,pool_disk>>   |beta
|<<metricbeat-module-couchbase,Couchbase>>    |   |beta
.3+|   |<<metricbeat-metricset-couchbase-bucket,bucket>>   |beta
|<<metricbeat-metricset-couchbase-cluster,cluster>>   |beta
|<<metricbeat-metricset-couchbase-node,node>>   |beta
|<<metricbeat-module-docker,Docker>>    |   |beta
.8+|   |<<metricbeat-metricset-docker-container,container>>   |beta
|<<metricbeat-metricset-docker-cpu,cpu>>   |beta
|<<metricbeat-metricset-docker-diskio,diskio>>   |beta
|<<metricbeat-metricset-docker-healthcheck,healthcheck>>   |beta
|<<metricbeat-metricset-docker-image,image>>   |beta
|<<metricbeat-metricset-docker-info,info>>   |beta
|<<metricbeat-metricset-docker-memory,memory>>   |beta
|<<metricbeat-metricset-docker-network,network>>   |beta
.
.
.
````

ruflin:
@dedemorton I'm pretty sure I can update the logic in a way to allow the above table formatting. Will do that and adjust the titles.

For the release status: It follows the same as the rest of the Elastic stack. Not sure if we have that described somewhere?

dedemorton:
@ruflin Actually...we should use the beta[] and experimental[] tags _inline_ just like we do for parameter descriptions. When used inline, the tag resolves to a link that displays a pop-up description when the user hovers over the link. We can have the tag in a separate column like you have things now (first screenshot below), or we can remove the last column and use the tag after the module/metricset name (second screenshot). The separate column makes the table less messy (IMO), but then someday we might end up with a column that's empty when all the metricsets and modules are GA.

**Screenshot 1:**

image

**Screenshot 2:**

image

The code is simple: Just include `beta[]` where you want the text/link to appear.

ruflin:
I really like the idea of the `beta` labels where people can hover. Also +1 on removing the column as definitively the goal is having all GA in the end. 

With this change I wonder if we still need a table. I kind of like the way the table gives more of a structured overview and makes it clear what a module and a metricset is, but a list could also work. One reason I'm thinking to also stick to a table is because one day we can perhaps add version compatibility here too (or other data).

ruflin:
@dedemorton One more thing I have in mind for the table: A checkmark if dashboards exist.

dedemorton:
Ok then let's keep the table, but use the 3rd column to indicate whether a dashboard is available.

we def need to limit the table to 3 cols or fewer

ruflin:
I pushed now a version which looks like this:

screen shot 2017-11-29 at 09 23 59

dedemorton:
Looks good! Normally I would suggest condensing the table further by moving the first metricset  up to the same row as the module name. However, the current design will work better when you add a third column containing info about dashboards. I think you're imagining something like this (faked the table, but should be easy in asciidoc):

mockup

ruflin:
I don't have the Dashboard data yet, only plan to add that later. Not sure what you meant by tweaking the title? Couldn't find the difference :-)

dedemorton:
Sorry, I wasn't clear. The screen shot doesn't show the change I suggested. Just make all the column titles plural:

**Modules Metricsets** (and someday) **Dashboards**

ruflin:
Ah sorry, got it. I just updated the PR for `Modules` and `Metricsets`.

ruflin:
@dedemorton If you are good with the change, can you merge / squash it?

PR #5721, opened 2017-11-27
Index pod IP too in `ip_port` kubernetes indexer

exekias:
Before this change, it only indexed ip:port combination, but in some cases, we want to have IP only, as source port may not be exported

### DIFF
--- a/libbeat/processors/add_kubernetes_metadata/indexers.go
+++ b/libbeat/processors/add_kubernetes_metadata/indexers.go
@@ -269,7 +269,14 @@ func (h *IPPortIndexer) GetMetadata(pod *Pod) []MetadataIndex {
 	if pod.Status.PodIP == "" {
 		return metadata
 	}
-	for i := 0; i < len(hostPorts); i++ {
+
+	// Add pod IP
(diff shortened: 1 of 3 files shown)


ruflin:
@exekias Can you rebase this one?

exekias:
done!

libbeat/processors/add_kubernetes_metadata/indexers.go
```diff
@@ -269,7 +269,14 @@ func (h *IPPortIndexer) GetMetadata(pod *Pod) []MetadataIndex {
...
 	}
-	for i := 0; i < len(hostPorts); i++ {
+
+	// Add pod IP
+	metadata = append(metadata, MetadataIndex{
+		Index: pod.Status.PodIP,
+		Data:  commonMeta,
+	})
+
+	for i := 1; i < len(hostPorts); i++ {
```
ruflin: Should this be `i := 0`?

exekias: no, as the first item in the list will be the ip, this filters it to range on ip:port pairs

PR #5725, opened 2017-11-27
Backport: Dashboard loading improvements and fixes (#5686)

kvch:
* Dashboard loading improvements and fixes

  install via kibana only
  enabled

* Review updates

(cherry picked from commit 554701f0a04fe25229be4afd6cccd745f2844860)

### DIFF
--- a/auditbeat/auditbeat.reference.yml
+++ b/auditbeat/auditbeat.reference.yml
@@ -722,8 +722,9 @@ output.elasticsearch:
 # dashboards and index pattern. Example: testbeat-*
 #setup.dashboards.index:
 
-# Force loading of dashboards using the Kibana API without querying Elasticsearch for the version
-#always_kibana: false
+# Always use the Kibana API for loading the dashboards instead of autodetecting
+# how to install the dashboards by first querying Elasticsearch.
+#setup.dashboards.always_kibana: false
 
 #============================== Template =====================================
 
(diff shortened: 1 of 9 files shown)


ruflin:
@kvch There are some failures on Jenkins which worry me a bit as it's related to loading templates. Not directly related to the change but better check. There are errors related to the shard splitting which is introduced in 6.1. We should check if the above changes are really tested against 6.1 of ES?

kvch:
Thanks. I am checking it.

kvch:
jenkins, test it

kvch:
@ruflin Are you worrying about `number_of_routing_shards` and `number_of_shards` compatibility? It don't think that it is related at all. This PR only changes the way dashboards are loaded. Template loading remains the same.

ruflin:
Yes, that are the ones I'm worried about. The part I wonder what it pops up in this PR but not other 6.1 PR's like https://github.com/elastic/beats/pull/5750 ? It seems like it uses an older ES version for testing? On the other hand it seems to pass without issues on travis?!?

@kvch I will merge this one as I think it would be great to have this clean in 6.1. But could you manually test this change with an ES/KB 6.1 build to make sure it works?

kvch:
@ruflin I tested template loading manually both with version 6.1 and 7.0.0-alpha1-snapshot. It works with both versions.

ruflin:
@kvch Awesome, thanks.

PR #5729, opened 2017-11-28
Fix the include top N processes feature for cases where there are fewer processes than N

ruflin:
In case include top N specified more processes then there were processes available an index out of bound panic was thrown. This is a use case that can especially happen in docker environments with few processes.

This needs backport to 6.0 and 6.1

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -68,6 +68,7 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
 - Fix error `datastore '*' not found` in Vsphere module. {issue}4879[4879]
 - Fix error `NotAuthenticated` in Vsphere module. {issue}4673[4673]
 - Fix connection leak in mongodb module. {issue}5688[5688]
+- Fix the include top N processes feature for cases where there are fewer processes than N. {pull}5729[5729]
 
 *Packetbeat*
 
--- a/metricbeat/module/system/process/helper.go
+++ b/metricbeat/module/system/process/helper.go
@@ -408,17 +408,27 @@ func (procStats *ProcStats) includeTopProcesses(processes []Process) []Process {
 
 	var result []Process
 	if procStats.IncludeTop.ByCPU > 0 {
+		numProcs := procStats.IncludeTop.ByCPU
+		if len(processes) < procStats.IncludeTop.ByCPU {
+			numProcs = len(processes)
+		}
+
 		sort.Slice(processes, func(i, j int) bool {
 			return processes[i].cpuTotalPct > processes[j].cpuTotalPct
 		})
-		result = append(result, processes[:procStats.IncludeTop.ByCPU]...)
+		result = append(result, processes[:numProcs]...)
 	}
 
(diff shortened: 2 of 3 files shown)


metricbeat/module/system/process/helper_test.go
```diff
@@ -287,3 +287,30 @@ func TestIncludeTopProcesses(t *testing.T) {
 		assert.Equal(t, resPids, test.ExpectedPids, test.Name)
 	}
 }
+
+// TestIncludeTopProcessesWithOnlyOne makes sure that alogrithm also works if only 1 process exists
```
andrewkroh: `s/alogrithm/algorithm/`

metricbeat/module/system/process/helper.go
```diff
@@ -401,7 +401,8 @@ func (procStats *ProcStats) GetProcStats() ([]common.MapStr, error) {
 
 func (procStats *ProcStats) includeTopProcesses(processes []Process) []Process {
 	if !procStats.IncludeTop.Enabled ||
-		(procStats.IncludeTop.ByCPU == 0 && procStats.IncludeTop.ByMemory == 0) {
+		(procStats.IncludeTop.ByCPU == 0 && procStats.IncludeTop.ByMemory == 0) ||
+		len(processes) == 1 {
```
andrewkroh: Isn't this a general problem that will cause a panic anytime that `len(processes) < procStats.IncludeTop.ByCPU` or `len(processes) < procStats.IncludeTop.ByMemory`? ((Try adding a test with two processes.)

I think logic like this for ByCPU and ByMemory would fix it. 

```
		numProcs := procStats.IncludeTop.ByCPU
		if len(processes) < procStats.IncludeTop.ByCPU {
			numProcs = len(processes)
		}
		result = append(result, processes[:numProcs]...)
```

ruflin: 👍  Added a more generic fix and more tests to cover this. Remove the special 1 tests as not needed.

CHANGELOG.asciidoc
```diff
@@ -65,6 +65,7 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
 - Change `MySQL active connections` visualization title to `MySQL total connections`. {issue}4812[4812]
 - Fix map overwrite in docker diskio module. {issue}5582[5582]
 - Fix connection leak in mongodb module. {issue}5688[5688]
+- Fix include top processes in case more processes included then existed. {pull}5729[5729]
```
andrewkroh: I think it's the other way around.

"Fix the include top N processes feature for cases where there are fewer processes than N."

ruflin: Isn't both the same just formulating it the other way around? Changing it to your option as it's easier to understand.

andrewkroh: Oh, I see what you were saying. The "more processes included" was misinterpreted by me. My bad.

ruflin:
@andrewkroh Updated changelog and rebased / squashed.

tsg:
Can't believe I didn't test that case. Thanks for the fix @ruflin.

PR #5741, opened 2017-11-28
Fixes #5739 - Removing index patterns with '*' from TSVB based visualizations

### LINKED ISSUE #5739: Change TSVB based visualizations index patterns to use metricbeat-*
Currently the TSVB visualizations use `*` for the index pattern, this is a problem for large clusters because it spreads the queries across all the indices. We should change this to use `metricbeat-*` so the queries are isolated to just the Metricbeat indices.

simianhacker:
This PR changes all the `\"index_pattern\":\"*\"` keys to a more appropriate index pattern per Beats module. For example in Metricbeat it changes from `\"index_pattern\":\"*\"` to `\"index_pattern\":\"metricbeat-*\"`. This will be a performance improvement for user with larger clusters.


ruflin:
Thanks for the fix. It seems that this is something we could also automatically overwrite as we already have the feature `setup.dashboards.index:` which can be used to overwrite this: https://github.com/elastic/beats/blob/master/libbeat/_meta/config.reference.yml#L629

@simianhacker Did you do a find/replace action to make this happen? If yes, were there some exceptions? More thinking if we should add a check to our scripts to make sure all index-pattern have `beat-*` in the future to not stumble over this issue again.

simianhacker:
@ruflin Yes... I searched for `\"index_pattern\":\"*\"` and then replaced the wildcard with the corresponding pattern `metricbeat-*` or `auditbeat-*` or `filebeat-*`.

ruflin:
Merged this one as I think we must have a fix now but we should figure out a way to automate this.

ruflin:
There is an issue in some of the dashboards from metricbeat as they contain `":\"metricbeat-*\".gc | update,`. Somehow my previous comment got swallowed by Github :-( @simianhacker Can you either revert this or create a follow up PR?

PR #5748, opened 2017-11-29
Make dialer based job creation of TCP monitor reusable

urso:
By making the job creation more reusable, new UDP/TCP based service can
be added, without having to introduce too much boilerplate code to
required to generate the correct job types.

- introduce dialchain.Builder and dialchain.MakeDialerJobs to create
  jobs based one common connection settings:
  - SOCKS5 proxy
  - ssl/tls settings
  - host + list of ports
  - ip resolve settings (one job per host vs. job per IP host resolves to)
- ensure TCP dialer itself adds the `tcp.port` field to an event

### DIFF
--- a/heartbeat/monitors/active/dialchain/net.go
+++ b/heartbeat/monitors/active/dialchain/net.go
@@ -3,6 +3,7 @@ package dialchain
 import (
 	"fmt"
 	"net"
+	"strconv"
 	"time"
 
 	"github.com/elastic/beats/heartbeat/look"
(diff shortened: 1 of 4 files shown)


heartbeat/monitors/active/dialchain/net.go
```diff
@@ -64,6 +65,16 @@ func netDialer(timeout time.Duration) NetDialer {
 				return nil, err
 			}
 
+			portNum, err := strconv.Atoi(port)
```
ruflin: I wonder if this should be more a config validation already?

urso: unfortunately stdlib net uses string for ports in quite a lot of places. Kind of forces me to switch between string and uint16 every now and then :(

ruflin: Sorry, comment was on the wrong line. I was wondering about the valid port range.

PR #5751, opened 2017-11-29
Cherry-pick #5714 to 6.1: Fix flaky node_stats Logstash tests

### LINKED ISSUE #5712: Flaky Logstash node_stats tests

ruflin:
Cherry-pick of PR #5714 to 6.1 branch. Original message: 

Update healthcheck to use _node/stats instead. As soon as this returns a 200 all endpoints should work as expected. See elastic/logstash#8004.

* Remove skipped tests
* Update LS image to latest stable version

Closes #5712

### DIFF
--- a/metricbeat/module/logstash/_meta/Dockerfile
+++ b/metricbeat/module/logstash/_meta/Dockerfile
@@ -1,4 +1,4 @@
-FROM docker.elastic.co/logstash/logstash:6.0.0-rc2
+FROM docker.elastic.co/logstash/logstash:6.0.0
 
 ENV XPACK_MONITORING_ENABLED=FALSE
-HEALTHCHECK --interval=1s --retries=90 CMD curl -f http://localhost:9600
+HEALTHCHECK --interval=1s --retries=90 CMD curl -f http://localhost:9600/_node/stats
--- a/metricbeat/module/logstash/node_stats/node_stats_integration_test.go
+++ b/metricbeat/module/logstash/node_stats/node_stats_integration_test.go
@@ -13,7 +13,6 @@ import (
 )
 
 func TestFetch(t *testing.T) {
-
 	compose.EnsureUpWithTimeout(t, 120, "logstash")
 
 	f := mbtest.NewEventFetcher(t, logstash.GetConfig("node_stats"))


ruflin:
Backported to get rid of the flaky tests in 6.1

ruflin:
@original-brownbear This is interesting. Even though this waits for a 200 from _node/stats the `in` part in the _node response is still missing: https://travis-ci.org/elastic/beats/jobs/308772888

```
ERR Error on field \'in\': Missing field: in, Error: Expected integer, found <nil>
```

I assume if we wait for `_node/stats` to return a 200, the `in` issue would also be solved. Any other checks we can add to solve the `in` issue?

original-brownbear:
@ruflin not really I'm afraid. All our tests on this thing basically work by retrying in a loop :(

ruflin:
@original-brownbear An other work around i can do, is make the field `in` optional on our side so I would not see an error. But that seems to defeat the purpose of why we have required fields. 

Is there an other method then "looping" we could use? We wait for the 200 and then something like send 1 request and it will be available on the second? Or we just wait 1s? What is causing the issue? Is it "just" a delay?

original-brownbear:
@ruflin yea it's just a delay resulting form the strange way metrics are set up. Waiting `1s` from the first time you see a `200` seems plenty here in theory, but in practice it's probably still gonna be flaky every now and then. => I'd go with a loop with things the way they are with that API right now, I don't think there is a smarter way than that with what LS provides as of today.

ruflin:
We could introduce a loop probably but I would prefer to keep the testing the same as we have for the other 30+ services. For now I introduced a 1s delay.

PR #5759, opened 2017-11-29
Fix WatcherDie test timings

exekias:
This change increases times in WatcherDie test, so it doesn't fail on slow machines


exekias:
jenkins retest this, please

exekias:
jenkins, retest it please

andrewkroh:
It's green on Jenkins. We'll keep an on this test and see if we need more adjustments.

ruflin:
@exekias I suggest we backport this to 6.1 as otherwise the 6.1 build is really flaky.

PR #5762, opened 2017-11-29
Update documentation for newbeat.asciidoc

tehmoon:
There was a missing package that I had trouble finding upgrading my beats from `5.x` to `6.x`

### DIFF
--- a/docs/devguide/newbeat.asciidoc
+++ b/docs/devguide/newbeat.asciidoc
@@ -474,9 +474,9 @@ package main
 
 import (
 	"os"
-	"github.com/spf13/cobra"
 
 	"github.com/elastic/beats/libbeat/beat"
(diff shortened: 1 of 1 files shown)


ruflin:
@tehmoon Thanks for fixing this.

@dedemorton What is our take on the dev guide? Which branches do we get these changes in?

dedemorton:
@ruflin Originally we said we would use master for building the dev guide, but since the content is getting automatically merged into the 6.x branches, it's showing up there, too. We probably need to revisit the question of whether to keep master as `current`, but for now, I think we should forward/backport the changes to master and 6.1.

ruflin:
Or alternatively remove it from these branches to have it only in one place. Perhaps worth discussing in the sync up?

PR #5764, opened 2017-11-29
Collect hearbeat monitors via make update

urso:
- Add scripts and Makefile target to update the monitor import list via
  `make update`, `make collect` or `make import`

- Update the generic script to support custom output file + adapt
  package name to actual output path


ruflin:
@urso What is this file `heartbeat/scripts/generate_imports_helper.yp` ?

script/generate_imports.py
```diff
@@ -1,18 +1,17 @@
...
 from generate_imports_helper import comment, get_importable_lines
 
 
 import_line_format = "\t_ \"{beat_path}/{module}/{name}\""
-include_list_path = "include/list.go"
 import_template = """/*
 {comment}
 */
-package include
+package {package}
```
ruflin: I wonder if this is going to break other beats.

BTW: you need to run `make fmt` to make CI happy, then will see ;-)

urso: package name is computed from file path. It shouldn't break the other beats.

PR #5766, opened 2017-11-30
TLS: report handshake latency

adriansr:
- Time spent between first packet and completion of handshake is reported in the common field `responsetime`.

- Histogram visualization added to Kibana dashboard

- New screen capture for docs


packetbeat/protos/tls/tls.go
```diff
@@ -343,6 +348,12 @@ func (plugin *tlsPlugin) createEvent(conn *tlsConnectionData) beat.Event {
 		}
 	}
 
+	// set "responsetime" if handshake completed
+	responseTime := int32(conn.endTime.Sub(conn.startTime).Nanoseconds() / 1e6)
```
andrewkroh: I'd prefer `time.Millisecond` instead of `1e6`.

PR #5770, opened 2017-11-30
Adds system test to Metricbeat to load dashboards

ruflin:
This adds a test to the system tests to load all dashboards into Kibana throught the Kibana API.

In a follow up PR this should be added to all Beats.

@simianhacker This currently goes green even thought some of the dashboards actually contain invalid json (see https://github.com/elastic/beats/pull/5741#issuecomment-348081687). I wonder if we could check in the API if the content that is received is at least valid json?


exekias:
Test failures seem unrelated, merging

PR #5773, opened 2017-11-30
Update port for SSL/TLS on http monitor

ugosan:
the port is usually 443 for SSL


ruflin:
@ugosan Thanks for the fix.

PR #5775, opened 2017-11-30
Update elastic/gosigar to v0.6.0

kvch:
Updated gosigar to v0.6.0
This PR is required by #5545


andrewkroh:
This is also fixing some issues in Metricbeat so we need to have some entries in the CHANGELOG for those issues.

andrewkroh:
LGTM

PR #5776, opened 2017-11-30
Updated docs for TLS module

adriansr:
Updated packetbeat docs regarding the TLS module.

I didn't see anything else to update but I might have skipped something. There is a `tls-dashboard.png` image but I don't see where to put it.

### DIFF
--- a/packetbeat/docs/packetbeat-options.asciidoc
+++ b/packetbeat/docs/packetbeat-options.asciidoc
@@ -1119,10 +1119,6 @@ An example of indexed event:
 
 [source,json]
 ------------------------------------------------------------------------------
-{
-  "@timestamp": "2017-11-30T23:25:50.330Z",
-  "type": "tls",
-  "client_port": 63530,
   "tls": {
     "handshake_completed": true,
     "server_certificate": {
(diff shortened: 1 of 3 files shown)


packetbeat/docs/packetbeat-options.asciidoc
```diff
@@ -1090,6 +1092,48 @@ at the end of the document.
...
+==== Configuration options
+
+The `send_certificates` and `include_raw_certificates` settings are useful
+for limiting the amount of data Packetbeat indexes in its certificate fields.
+
+Also see <<common-protocol-options>>.
+
+===== `send_certificates`
+
+If the certificates presented by the client and server must be included in the
```
dedemorton: You have a sentence fragment here. Did something get deleted by mistake? Did you mean to say, "Specify this setting if the certificates....?

packetbeat/docs/packetbeat-options.asciidoc
```diff
@@ -1090,6 +1092,48 @@ at the end of the document.
...
+event. The server's certificate is indexed in the `server_certificate` field and
+its certification chain in the `server_certificate_chain`. For the client,
```
dedemorton: Change to: "the `server_certificate_chain` field.

packetbeat/docs/packetbeat-options.asciidoc
```diff
@@ -1090,6 +1092,48 @@ at the end of the document.
...
+------------------------------------------------------------------------------
+packetbeat.protocols:
+- type: tls
+  send_certificates: true
+  include_raw_certificates: true
+------------------------------------------------------------------------------
+
+==== Configuration options
+
+The `send_certificates` and `include_raw_certificates` settings are useful
```
dedemorton: Might help to add another sentence or two here because it's not immediately clear to me how these fields limit the amount of data indexed. Maybe it would help to explain the default behavior when users don't specify these fields. If you think this detail is too much to include in the description, maybe consider including a summary in the intro to this topic.

packetbeat/docs/packetbeat-options.asciidoc
```diff
@@ -1090,6 +1092,48 @@ at the end of the document.
...
+If the certificates presented by the client and server must be included in the
+event. The server's certificate is indexed in the `server_certificate` field and
+its certification chain in the `server_certificate_chain`. For the client,
+`client_certificate` and `client_certificate_chain` are used.
+
+The default is `true`.
+
+===== `include_raw_certificates`
+
+In addition to a fields representation of each certificate, include also the
```
dedemorton: Maybe switch this around a bit because it's a little hard to parse at first:

When `send_certificates` is `true`, you can set `include_raw_certificates` to include the raw certificate encoded in PEM format as a `raw` field under each certificate field.

If `send_certificates` is `false`, this setting is ignored. The default is `false`.

packetbeat/docs/packetbeat-options.asciidoc
```diff
@@ -1090,6 +1092,48 @@ at the end of the document.
 Note that limiting documents in this way means that they are no longer correctly
 formatted JSON objects.
 
+[[configuration-tls]]
+=== Capture TLS traffic
+
+++++
+<titleabbrev>TLS</titleabbrev>
+++++
+
```
andrewkroh: I think we need an overview of the feature that explains what it does and how it does. Some points to possibly cover:

- It captures TLS handshake data.
- Make it clear that this does not decrypt any traffic.
- It provides information about the TLS versions, ciphers, and certificates being used. It measures the time it takes for the handshake to complete.
- It computes a JA3 fingerprint of the client's TLS parameters. The fingerprint can be used to identify an application associated with the traffic. The fingerprint is independent of network addresses and certificates.

packetbeat/docs/packetbeat-options.asciidoc
adriansr: No, I'm just a lousy writer.

packetbeat/docs/packetbeat-options.asciidoc
adriansr: I will leave the JA3 and handshake time for another patch so this one can be backported to 6.1.

adriansr:
@dedemorton I condensed the event a little bit, hope that is enough.

Please feel free to comment on all those nit picks or better, just rewrite yourself anything you don't like.

Issue #5746, opened 2017-11-29
Jenkins tests fail after previous test was aborted without cleanup


ruflin:
If a new commit to an existing PR is pushed or the PR is closed, the currently running CI jobs are aborted. Unfortunately this aborting starts the cleanup but does not let it finish. Here is an example of such a job: https://beats-ci.elastic.co/job/elastic+beats+pull-request+multijob-linux/beat=filebeat,label=ubuntu/2291/console

The output looks as following:

```
16:59:21 Test ML are installed in all possible ways ... ok (24.3232s)
17:00:31 Build was aborted
17:00:31 Archiving artifacts
17:00:31 Tests all filebeat modules ... make: *** [testsuite] Terminated
17:00:31 Running cleanup...
17:00:31 + cleanup
17:00:31 + echo 'Running cleanup...'
17:00:31 + rm -rf /tmp/tmp.I13AHtRbbx
17:00:31 /tmp/jenkins8075086993931634532.sh: line 2:  9910 Terminated              ./dev-tools/jenkins_ci.sh
17:00:31 Aborting.
17:00:31 make[1]: *** [system-tests-environment] Terminated
17:00:31 ERROR: Failed to archive artifacts: **/build/system-tests/run/**/*
17:00:31 java.io.EOFException: Unexpected end of ZLIB input stream
17:00:31 	at com.jcraft.jzlib.InflaterInputStream.fill(InflaterInputStream.java:186)
17:00:31 	at com.jcraft.jzlib.InflaterInputStream.read(InflaterInputStream.java:106)
17:00:31 	at org.apache.commons.compress.utils.IOUtils.readFully(IOUtils.java:160)
17:00:31 	at org.apache.commons.compress.utils.IOUtils.readFully(IOUtils.java:134)
17:00:31 	at org.apache.commons.compress.archivers.tar.TarArchiveInputStream.readRecord(TarArchiveInputStream.java:419)
17:00:31 	at org.apache.commons.compress.archivers.tar.TarArchiveInputStream.getRecord(TarArchiveInputStream.java:388)
17:00:31 	at org.apache.commons.compress.archivers.tar.TarArchiveInputStream.getNextTarEntry(TarArchiveInputStream.java:269)
17:00:31 	at hudson.FilePath.readFromTar(FilePath.java:2295)
```

It seems the cleanup starts but then is also aborted.

Here some potential leads for solutions:

* https://wiki.jenkins.io/display/JENKINS/Aborting+a+build

[...]

PR #5780, opened 2017-11-30
Do cleanup on Jenkins before build starts

ruflin:
This should solve the issue that in case a build is canceled, the cleanup is interrupted and not fully finished and leaves files behind which break the following build. See https://github.com/elastic/beats/issues/5746

* Move cleanup function to common.bash


andrewkroh:
IIUC this will not fix the issue with aborted builds. The git plugin attempts to wipe the workspace before proceeding with its checkout. It is during this wipe that permission errors are encountered. And this is before any of our scripts are executed.

ruflin:
Closing as this didn't solve the problem and we try now to use the jenkins cleanup plugin.

Issue #5783, opened 2017-12-01
Packaging broken because of uint64 overflow in auditbeat


ruflin:
Packaging for auditbeat currently fails with the following error:

```
23:56:44 /usr/local/osx-ndk-x86/bin/../SDK/MacOSX10.11.sdk/usr/include/dispatch/module.map:1:27: warning: unknown attribute 'extern_c' [-Wignored-attributes]
23:56:44 # github.com/elastic/beats/vendor/github.com/fsnotify/fsevents
23:56:44 vendor/github.com/fsnotify/fsevents/wrap.go:38: constant 36893488147419103231 overflows uint64
23:56:47 make[2]: *** [prepare-package-cgo] Error 2
23:56:47 make[2]: Leaving directory `/var/lib/jenkins/workspace/elastic+beats+master+package/src/github.com/elastic/beats/auditbeat'
23:56:47 make[1]: *** [package] Error 2
23:56:47 make[1]: Leaving directory `/var/lib/jenkins/workspace/elastic+beats+master+package/src/github.com/elastic/beats/auditbeat'
```

The culprit seems to be the following line: https://github.com/elastic/beats/blob/master/vendor/github.com/fsnotify/fsevents/wrap.go#L38

Strangely the dependency did not change in the last 5 months. I will test if I can reproduce it locally.

ruflin:
@adriansr @andrewkroh The problem was introduced with this PR: https://github.com/elastic/beats/pull/5575 We had the package in beats before but it was not used. It started to be used in the above PR. 

Also locally on my machine `go build` returns the overflow issue. I think our CI did not catch this issue because currently it's not complied for OS X. I will open a PR for that.

andrewkroh:
https://github.com/fsnotify/fsevents/issues/31

ruflin:
Probably related to https://github.com/golang/go/issues/21708

ruflin:
@andrewkroh @adriansr I'm reopening this issue as this was auto closed by https://github.com/elastic/fsevents/pull/2 I think merging the PR does not fix it yet as it first needs to make it into the vendor directory of beats.

andrewkroh:
Fixed by #5866

PR #5785, opened 2017-12-01
Add targets to crosscompile auditbeat and test on OS X

ruflin:
For travis tests were added to run the testsuite also on OS X on travis and a crosscompile target was added for auditbeat. Having auditbeat built on OS X should prevent issues like https://github.com/elastic/beats/issues/5783.

In addition a global crosscompile target was added which should be used in Jenkins in the future. It does not use the BEATS array because packetbeat cannot be just crosscompiled on each machine for the different platforms because of the pcap dependencies. Currently the list of beats is hardcoded.

### DIFF
--- a/auditbeat/Makefile
+++ b/auditbeat/Makefile
@@ -3,6 +3,7 @@ BEAT_TITLE=Auditbeat
 BEAT_DESCRIPTION=Audit the activities of users and processes on your system.
 SYSTEM_TESTS=false
 TEST_ENVIRONMENT=false
+GOX_OS?=linux darwin windows ## @Building List of all OS to be supported by "make crosscompile".
 
 # Path to the libbeat Makefile
 -include ../libbeat/scripts/Makefile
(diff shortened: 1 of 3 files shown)


auditbeat/Makefile
```diff
@@ -3,6 +3,7 @@ BEAT_TITLE=Auditbeat
 BEAT_DESCRIPTION=Audit the activities of users and processes on your system.
 SYSTEM_TESTS=false
 TEST_ENVIRONMENT=false
+GOX_OS?=linux darwin ## @Building List of all OS to be supported by "make crosscompile".
```
ruflin: @andrewkroh @adriansr Please advice if these are the correct platforms to crosscompile to.

andrewkroh: + windows

ruflin: added

andrewkroh:
@ruflin Can you rebase this one now that fsevents has been patched.

ruflin:
@andrewkroh I think failures on CI are not related to this PR.

PR #5787, opened 2017-12-01
Ignore docker kill events in `add_docker_metadata`

exekias:
Docker kill events are caused by any signal sent to the container, not only a SIGKILL, we can safely ignore all of them, as `die` will always happen for the case we are really interested on.

Context on the issue: https://discuss.elastic.co/t/filebeats-docker-metadata-watcher-losing-container-from-its-list-somehow/109897


exekias:
wrong branch

PR #5792, opened 2017-12-01
Heartbeat monitor code generator

urso:
- add support for `make create-monitor`
- code that needs to be adapted by user is marked with `IMPLEMENT_ME`

Requires: #5764


ruflin:
@urso Can you fix the Makefile conflict?

heartbeat/Makefile
```diff
@@ -22,3 +22,8 @@ fields:
 	@mkdir -p _meta/
 	@cp ${ES_BEATS}/heartbeat/_meta/fields.common.yml _meta/fields.generated.yml
 	@cat ${ES_BEATS}/heartbeat/monitors/active/*/_meta/fields.yml >> _meta/fields.generated.yml
+
+# Create a new monitor. Requires the parameter MONITOR
+.PHONY: create-monitor
+create-monitor:
+	go run scripts/generate_monitor/main.go -monitor=${MONITOR}
```
ruflin: Nice, it's in golang. I wonder if we should put this under the `heartbeat/cmd/generate_monitor` package. It would follow better what we do with `dev-tools` but would conflict with the logic we have now that "all" commands under `cmd` can be accessed through `cobra`. @exekias Any thoughts?

heartbeat/scripts/generator/Readme.md
```diff
@@ -0,0 +1,33 @@
+## Readme
```
ruflin: Nit: We normally call it `README.md`.

heartbeat/scripts/generator/Readme.md
```diff
@@ -0,0 +1,33 @@
...
+- `job.go`: Implements the ping function for connecting and validating an
+  endpoint. This file generates the monitor specific fields to an event.
+- `<name>.go`: The monitor entrypoint registering the factory for setting up
+  monitoring jobs.
+- `_meta/fields.yml`: Document the monitors event field.
+- `_meta/config.yml`: Minimal sample configuration file 
+- `_meta/config.reference.yml`: Reference configuration file. All availalbe
+  settings are documented in the reference file.
+
+Code comments tagged with `IMPLEMENT_ME` in the go and meta files give details
```
ruflin: I like that convention

heartbeat/scripts/generator/{{monitor}}/_meta/config.reference.yml.tmpl
```diff
@@ -0,0 +1,62 @@
...
+  schedule: '@every 30s'
+
+  # list of hosts to monitor
+  hosts: ["localhost"]
+
+  # list of ports to ping (used if host is given without portname)
+  ports: [12345]
+
+  # SOCKS5 proxy url
+  # proxy_url: ''
```
ruflin: remove leading space

heartbeat/Makefile
urso: wouldn't put it in `heartbeat/cmd/generate_monitor`, as this would imply the code generator being available in the heartbeat binary itself. As the generator requires a full dev setup, I'd rather keep it where it is (well, I find having the generators in `scripts` somewhat confusing though).

Most of the go tool is driving the template engine. We can make it re-usable, but introducing some of the common functionalities in `beats/dev-tools/...`

Normally `cmd` indicates a multitude of compilable tools, but we use it for structuring sub-commands. Adding dev tools to `cmd`, I'd also move the current `cmd` package by one level to:

```
beats/heartbeat/cmd:
  generator/ ...  # the code generator
  main/ ...          # all contents currently in cmd package
```

or 

```
beats/heartbeat/cmd:
  generator/ ...  # the code generator
  heartbeat/ ...          # all contents currently in cmd package
```

Alternatively (given we take beat as framework more serious):

```
beats/cmd/
  filebeat/...
  packetbeat/...
  heartbeat/...
  generate.../...
beats/libbeat/...
beats/filebeat/...
beats/heartbeat/...
beats/packetbeat/...
```

Oh my

ruflin: Lets ignore that for now but I think we have similar ideas in mind moving forward.

ruflin:
I haven't tested it locally on my end. We should add some tests similar to the other generators at one stage to make sure the generator code stays up-to-date if we change our internal code.

urso:
Followup #5823 . It's not the the only generator which requires testing.

PR #5606, opened 2017-11-16
Add ceph osd_df to metricbeat

elaron:
osd_df will record each osd node's disk usage for ceph cluster.

PR #5794, opened 2017-12-01
Update modules list for Metricbeat

andrewkroh:
metricbeat/docs/modules_list.asciidoc needs updated in master.


ruflin:
Seems this was broken in https://github.com/elastic/beats/pull/5606 CI did not fail as the PR that changed the generation was merged after CI was run.

PR #5796, opened 2017-12-03
[Metricbeat] Step down from strong consistency mongodb sessions

### LINKED ISSUE #4689: Mongobeat secondary node "no reachable servers" error
- Version:  5.4.1
- Operating System: linux 
- Steps to Reproduce: 

metricbeat config:

- module: mongodb
  metricsets: ["status"]
  enabled: true
  period: 10s

  hosts: ["mongodb://user:password@host-secondary-1:27017"]

Mongo CLuster Config
Mongo is made up of a replica of 3 nodes. host-primary is master. host-secondary-[1/2] are two slave

How to reporduce?

- set up mongo cluster as above
- start metricbeat with above mongobeat config
- you will get "no reachable servers" error

MerlinDMC:
Allow for low consistency session so connections to secondary nodes are allowed to read data.

Fixes #4689

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -73,6 +73,7 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
 - Fix the include top N processes feature for cases where there are fewer processes than N. {pull}5729[5729]
 - Fix `ProcState` on Linux and FreeBSD when process names contain parentheses. {pull}5775[5775]
 - Fix incorrect `Mem.Used` calculation under linux. {pull}5775[5775]
+- Fix mongodb session consistency mode to allow command execution on secondary nodes. {issue}4689[4689]
 
 *Packetbeat*
 
--- a/metricbeat/module/mongodb/mongodb.go
+++ b/metricbeat/module/mongodb/mongodb.go
@@ -82,5 +82,8 @@ func NewDirectSession(dialInfo *mgo.DialInfo) (*mgo.Session, error) {
 		return nil, err
 	}
 
+	// Relax consistency mode so reading from a secondary is allowed
+	session.SetMode(mgo.Monotonic, true)
+
 	return session, nil
 }


ruflin:
@MerlinDMC Thanks for taking this one. This PR triggered some thoughts on my end.

Our general recommendation for Metricbeat is to have it installed on the edge node with each service. For distributed system like Elasticsearch, this means having a metricbeat instance for each node. This gives insights into each local machine and gets data from each node locally. Like this node information (in the terms of ES) can also be collected when a node is disconnected to a cluster for example. 

Now reading up on the Mongo change here it implies that there are primary and secondary server that we gather data from. For the `dbstats` metricset I don't think it should matter from which instance we gather the data. For the `status` metricset I'm not sure. The status of the primary server could be different from one of the secondary (that is the reason we have multiple server). I would expect metricbeat to be installed for each mongo server and deliver data for each. So assuming we have 1 primary and 2 secondary, 3 metricbeat instance would be running connecting to each instance locally.

So far it seems we connected to master only which is also not expected from my point of view. I would expect that the data can only come from the "server" we connected to for status info. So if I connected to secondary 1, I don't want to have the status info from secondary to. In elasticsearch there is a flag `_local` I can use on a connection to only get the status info from the local node. I don't know if there is something similar on MongoDB?

@MerlinDMC @MrGarry2016 I would like to hear your thoughts on the above and how you think it should work.

MerlinDMC:
@ruflin As far as I can tell status and dbstats are different between primary and secondary nodes. But it will highly depend on your overall usage.
It gets worse when you go to collstats (I've got code ready for a PR but wanted to get the simple things in first).

Overall I think you can't get around connecting to each node individually to get a full picture as you otherwise can't correlate mongodb stats with the actual system stats. It also doesn't look like it would be expected to only connect to the primary nodes `NewDirectSession` wants to create a direct session it just missed the part about those in the actual mgo docs.
https://github.com/globalsign/mgo/blob/5be15cc6a44937b6937a9487e1633f8a5fb25e4a/session.go#L224-L229

ruflin:
My goal is that it always gets information only from the node it connects to. So if I understand you and your PR correctly, we already do that, but right now it was not possible to connect to a secondary node and this PR allows it? +1 from my side then :-D

Could you change the CHANGELOG entry a bit that it states something like the above?

MerlinDMC:
@ruflin It will connect and authenticate no matter what consistency your session is running in. But if you are in a session asking for strong consistency and you try to read data from a secondary node that is asynchronously updated the mongodb server will fail all commands.
And in this case here because you connected in direct mode the error coming back isn't very helpful `no reachable servers`.

I've updated the changelog entry.

ruflin:
jenkins, test it

ruflin:
@MerlinDMC Merged. Thanks for the contribution. Looking forward to the mentioned PR's 🎉

PR #5797, opened 2017-12-03
Make logstash node system test more stable

ruflin:
Introduce a 1 sec timeout to help the service to get available and return all fields.

### DIFF
--- a/metricbeat/module/logstash/node/node_integration_test.go
+++ b/metricbeat/module/logstash/node/node_integration_test.go
@@ -13,7 +13,6 @@ import (
 )
 
 func TestFetch(t *testing.T) {
-
 	compose.EnsureUpWithTimeout(t, 120, "logstash")
 
 	f := mbtest.NewEventFetcher(t, logstash.GetConfig("node"))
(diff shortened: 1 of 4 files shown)


exekias:
In the past I used grep on the healthcheck, to ensure the missing field is already populated, wdyt?

ruflin:
@exekias Interesting idea. We could definitively do that. Not a fan of regexp foo but probably the most stable solution. Which service did you do that for some copy / paste? Seems like we need in the end a script for the healtcheck :-(

exekias:
```
module/ceph/_meta/Dockerfile:HEALTHCHECK --interval=1s --retries=90 CMD ceph --status | grep HEALTH_OK
module/kibana/_meta/Dockerfile:HEALTHCHECK --interval=1s --retries=120 CMD curl -f http://localhost:5601/api/status | grep '"disconnects"'
module/kubernetes/_meta/Dockerfile:HEALTHCHECK --interval=1s --retries=90 CMD curl -f localhost:10255/stats/summary | grep kube-addon-manager
```

ruflin:
jenkins, retest it

metricbeat/module/logstash/node/node_integration_test.go
```diff
@@ -13,7 +13,6 @@ import (
 )
 
 func TestFetch(t *testing.T) {
-
```
ruflin: @andrewkroh for you 😆

PR #5803, opened 2017-12-04
Fix crash in TLS event generation

### LINKED ISSUE #5801: TLS module panics when server hello is missing
The following panic has been reported while using the TLS module:

`2017-12-04T21:20:53Z ERR  Process tcp exception. Recovering, but please report this: runtime error: invalid memory address or nil pointer dereference.
2017-12-04T21:20:53Z ERR  Stacktrace: goroutine 65 [running]:
runtime/debug.Stack(0x1131ec6, 0x2b, 0x13dab988)
        /home/akroh/.gvm/versions/go1.9.2.linux.armv6l/src/runtime/debug/stack.go:24 +0x80
github.com/elastic/beats/libbeat/logp.Recover(0x111905b, 0x15)
        /home/akroh/go/src/github.com/elastic/beats/libbeat/logp/log.go:176 +0xb4
panic(0x10172c8, 0x170dac8)
        /home/akroh/.gvm/versions/go1.9.2.linux.armv6l/src/runtime/panic.go:491 +0x204
github.com/elastic/beats/packetbeat/protos/tls.(*tlsPlugin).createEvent(0x13ce9c60, 0x13a3da00, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
        /home/akroh/go/src/github.com/elastic/beats/packetbeat/protos/tls/tls.go:283 +0x248
github.com/elastic/beats/packetbeat/protos/tls.(*tlsPlugin).sendEvent(0x13ce9c60, 0x13a3da00)
        /home/akroh/go/src/github.com/elastic/beats/packetbeat/protos/tls/tls.go:216 +0x74
github.com/elastic/beats/packetbeat/protos/tls.(*tlsPlugin).GapInStream(0x13ce9c60, 0x13e5277c, 0x75463400, 0x1c, 0xff6d28, 0x13a3da00, 0x113acd8, 0x31, 0x144894c0)`

It can be reproduced by removing the server hello from a captured TLS handshake and the client hello includes a non-empty session ID.

adriansr:
A panic was raised when the server hello was missing
and the client hello contained a session ID.

Fixes #5801

### DIFF
--- a/packetbeat/protos/tls/tls.go
+++ b/packetbeat/protos/tls/tls.go
@@ -260,6 +260,8 @@ func (plugin *tlsPlugin) createEvent(conn *tlsConnectionData) beat.Event {
 	if server.parser.hello != nil {
 		serverHello = server.parser.hello
 		tls["server_hello"] = serverHello.toMap()
+	} else {
+		serverHello = emptyHello
 	}
 	if cert, chain := getCerts(client.parser.certificates, plugin.includeRawCertificates); cert != nil {
 		tls["client_certificate"] = cert


andrewkroh:
LGTM. Long term I think we should get some more test cases added. Like playing back some interesting pcaps and asserting on the events that are output.

packetbeat/protos/tls/tls.go
```diff
@@ -260,6 +260,8 @@ func (plugin *tlsPlugin) createEvent(conn *tlsConnectionData) beat.Event {
 	if server.parser.hello != nil {
 		serverHello = server.parser.hello
 		tls["server_hello"] = serverHello.toMap()
+	} else {
+		serverHello = emptyHello
```
ruflin: We could init serverHello and clientHello with emptyHello on line 247. Would probably make the a bit more readable (for me).

PR #5805, opened 2017-12-04
Fix crash in TLS event generation [6.1]

adriansr:
#5803 backported to 6.1


andrewkroh:
Fixes #5801 for 6.1.0.

ruflin:
Merging as packetbeat tests are green.

PR #5807, opened 2017-12-05
Use default frequency and query_delay in ML datafeeds

dimitris-athanasiou:
There is no need to set explicitly `frequency` and `query_delay`
parameters for ML datafeeds. The current values are the default
values but a change in the way datafeeds with aggregations work
makes those frequencies invalid settings. Also, in the future the
smart defaults might evolve, so using them will leverage such
improvements.

### DIFF
--- a/filebeat/module/nginx/access/machine_learning/datafeed_low_request_rate.json
+++ b/filebeat/module/nginx/access/machine_learning/datafeed_low_request_rate.json
@@ -1,7 +1,5 @@
 {
     "job_id": "JOB_ID",
-    "query_delay": "60s",
-    "frequency": "450s",
     "indexes": [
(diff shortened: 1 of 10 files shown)


ruflin:
@dimitris-athanasiou Thanks for the change. In the future please use "Squash and Merge" if you merge it yourself. Does this need a backport? I assume that is what you meant with the labels?

dimitris-athanasiou:
@ruflin It all went wrong. Github's default when you merge a PR for the first time in a repo is "Merge" but out of habit I just pressed it. I also backported by cherry-picking. I discussed this already with Tudor. I'll know better in the future :-)

ruflin:
@dimitris-athanasiou no worries :-D

Issue #5795, opened 2017-12-01
Fields merging in 6.0 differs from 5.x


tehmoon:
Following the bug report I filed on discuss: https://discuss.elastic.co/t/fields-merging-in-6-0-differs-from-5-x/109838

I was in the process of upgrade my filebeats from `5.6` to `6.0` on `linux`. But it broke my `logstash` pipeline. The reason is -- I think -- a bug in the merging fields process of the `fields` option in the yaml config file.

My configuration is as follow:

```
filebeat.prospectors:

- input_type: log
  paths:
    - /xxx/ansible_events/events.log
  fields:
    type: ansible_events

- input_type: log
  paths:
    - /xxx/xxx/log/main.log
  include_lines: ['^\[info\]'] # include info log
  exclude_lines: ['\)$', 'GET /users ', 'GET /ping '] # exclude lines that doesn't have the Sent information and also /users and /ping
  fields:
    type: cs_main

- input_type: log
  paths:
    - /var/log/syslog
    - /var/log/auth.log
    - /var/log/kern.log
  fields:
    type: syslog

name: filebeat

fields:
  env: production
  datacenter: yyyy
  group: zzzz

output.logstash:
  hosts: ["x.x.x.x:5044"]

path.data: /xxx/elastic_beats/filebeat/data
path.logs: /xxxx/elastic_beats/filebeat/log
```

I do all the sorting on the `fields.type` field. Everything worked fine in `5.6.4`, the files `syslog`,`auth.log` and `kern.log` would have `fields.type` set to `syslog`. `events.log` would have it set to `ansible_events` and `main.log` to `cs_main`.

After upgrading to `6.0.0` all my events were set to `syslog`. A quick debugging showed that `filebeat` was taking *the last* `input_type` declared. To prove it, I just moved some section on top to the bottom and all of the events got that `fields.type`.

I'm not too sure how to debug further, but I'll happy to answer quickly if need help.

ruflin:
I did a quick test here to reproduce it. The following still works as expected:

```
- type: log
  paths:
    - .../test1.log
  fields:
    type: test1

- type: log
  paths:
    - .../test2.log
  fields:
    type: test2
```

But if I add global fields like below, I have `fields.type: test2` also in the events from `test1.log`.

```
fields:
  group: zzzz
```

@urso The problem must happen when merging global with local fields. Perhaps some state?

@tehmoon Thanks for reporting. A workaround for now would be to specify all fields local for each prospector.

[...]

PR #5808, opened 2017-12-05
Fix isolation of modules when merging local and global field settings.

urso:
Resolves: #5795 

The user global and local fields configurations must be merged
on pipeline setup. The DeepUpdate method did only keep references to the
globally shared fields, such that local fields settings do overwrite
the shared global state. If more then one beat.Client instance did set
local fields, the global state (and all fields added to an event) will
be given by the last client being instantiated.

The fix Clones the global fields before merging, such that the
beat.Client fields operate fully in isolation.

### DIFF
--- a/libbeat/publisher/pipeline/pipeline.go
+++ b/libbeat/publisher/pipeline/pipeline.go
@@ -68,7 +68,8 @@ type pipelineProcessors struct {
 
 	processors beat.Processor
 
-	disabled bool // disabled is set if outputs have been disabled via CLI
+	disabled   bool // disabled is set if outputs have been disabled via CLI
+	alwaysCopy bool
 }
 
 // Settings is used to pass additional settings to a newly created pipeline instance.
@@ -304,8 +305,7 @@ func (p *Pipeline) ConnectWith(cfg beat.ClientConfig) (beat.Client, error) {
 		}
 	}
 
-	processors := p.newProcessorPipeline(cfg)
-
+	processors := newProcessorPipeline(p.beatInfo, p.processors, cfg)
 	acker := p.makeACKer(processors != nil, &cfg, waitClose)
 	producerCfg := queue.ProducerConfig{
 		// Cancel events from queue if acker is configured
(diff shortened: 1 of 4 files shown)


libbeat/publisher/pipeline/pipeline.go
```diff
@@ -68,7 +68,8 @@ type pipelineProcessors struct {
 
 	processors beat.Processor
 
-	disabled bool // disabled is set if outputs have been disabled via CLI
+	disabled   bool // disabled is set if outputs have been disabled via CLI
+	alwaysCopy bool
```
kvch: Where does this variable get its value?

urso:
> WDYT about adding a system test to verify the change?

I reproduced + verified the issue via unit tests. Really need an additional system test here?

libbeat/publisher/pipeline/pipeline.go
urso: I introduced it for the unit tests enforce the copy of fields. This way, when adding more tests we can see if copy vs. non-copy of fields makes a difference.

ruflin:
@urso Not required, only nice to have. Glad we have the unit tests.

libbeat/publisher/pipeline/processor.go
```diff
@@ -39,7 +39,9 @@ type processorFn struct {
 //  8. (P) pipeline processors list
 //  9. (P) (if publish/debug enabled) log event
 // 10. (P) (if output disabled) dropEvent
-func (p *Pipeline) newProcessorPipeline(
+func newProcessorPipeline(
+	info beat.Info,
```
ruflin: Why did you decide to pass info instead of using it as variable as before?

urso: The info is used for the debug print. It used to be part of the pipeline variable. I removed the method and created a constructor passing only those information required to build the processor pipeline. The decoupling simplified the construction of the processor pipeline in the unit tests.

ruflin:
@urso Can you take care of the backports?

ruflin:
@urso I think we also need to backport this to 6.0?

PR #5811, opened 2017-12-05
 Elasticsearch bulk request enhancements

urso:
Requires: #5810 

- Log error and drop event if no index name can be computed (ES API
  returns error if index is empty)
- Set `_id` if event.Meta["id"] is set
- Use `index` action if no `id` is set and `create` if `id` is set.

### DIFF
--- a/libbeat/outputs/elasticsearch/client.go
+++ b/libbeat/outputs/elasticsearch/client.go
@@ -83,10 +83,10 @@ type bulkCreateAction struct {
 }
 
 type bulkEventMeta struct {
-	Index    string      `json:"_index" struct:"_index"`
-	DocType  string      `json:"_type" struct:"_type"`
-	Pipeline string      `json:"pipeline,omitempty" struct:"pipeline,omitempty"`
-	ID       interface{} `json:"_id,omitempty" struct:"_id,omitempty"`
+	Index    string `json:"_index" struct:"_index"`
+	DocType  string `json:"_type" struct:"_type"`
+	Pipeline string `json:"pipeline,omitempty" struct:"pipeline,omitempty"`
+	ID       string `json:"_id,omitempty" struct:"_id,omitempty"`
 }
 
 type bulkResultStats struct {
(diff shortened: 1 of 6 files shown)


libbeat/outputs/elasticsearch/client.go
```diff
@@ -318,7 +346,11 @@ func bulkEncodePublishRequest(
 	okEvents := data[:0]
 	for i := range data {
 		event := &data[i].Content
-		meta := createEventBulkMeta(index, pipeline, event)
+		meta, err := createEventBulkMeta(index, pipeline, event)
+		if err != nil {
+			logp.Err("Failed to encode event meta dat: %s", err)
```
ruflin: Nit `s/dat/data`

libbeat/outputs/elasticsearch/client.go
```diff
@@ -329,48 +361,38 @@ func bulkEncodePublishRequest(
...
+		return nil, err
 	}
 
-	type bulkMetaIndex struct {
-		Index    string `json:"_index" struct:"_index"`
-		DocType  string `json:"_type" struct:"_type"`
-		Pipeline string `json:"pipeline" struct:"pipeline"`
+	var id interface{}
+	if m := event.Meta; m != nil {
+		id = m["id"]
```
ruflin: Similar comment to other PR: Should we really allow here interface or make sure this is converted to a string? Do we know the key `"id"` exists in `m` ?

ruflin:
@urso Needs a rebase.

libbeat/outputs/elasticsearch/client.go
urso: can `_id` be a number instead of a string?

ruflin: I'm assuming that `_id` on the ES side is a keyword. So if `12` is put in it ends up also as `"12"`. Even if that is not the case, I think we should enforce string values on beats side.

urso: +1

urso:
@ruflin rebased and ensure `id` is a string.

PR #5818, opened 2017-12-06
Remove dependency on testing package in all beats

urso:
Due to shared code used to connect to the shared elasticsearch test
instance, all beats do include the golang testing framework. This PR
breaks the dependency by moving common code into the
elasticsearch/internal package (which itself does not depend on testing
by using interfaces) + introduces estest package for use by other
packages.


libbeat/outputs/elasticsearch/estest/estest.go
```diff
@@ -0,0 +1,23 @@
+package estest
```
ruflin: We could also have called `testing` as we do for other such packages.

ruflin:
Seems like this breaks some filebeat tests: `fileset/modules_integration_test.go:17:12: undefined: elasticsearch.GetTestingElasticsearch`

libbeat/outputs/elasticsearch/estest/estest.go
urso: no big fan of 'testing' package, as `estest` some a little more vocal about helpers being about elasticsearch related tests. Fortunately we don't have many tests importing multiple testing packages ;)

PR #5819, opened 2017-12-06
haproxy http stats and some fixes in tests

### LINKED ISSUE #5178: HAProxy module authentication support for metrics
HAProxy metrics can be configured with authentication. The MetricBeat HAProxy module does not seem to be able to pull this data when authentication is supported. My client is adamant he needs this feature to monitor sensitive production devices.

jsoriano:
Adds support for haproxy http stats with support with basic authentication (what fixes #5178).

It also tries to fix a couple of problems seen when running tests:
* Configuration file created with umask 0177 so it has proper permissions for metricbeat
* `test-module` make target depends on `fields` and `metricbeat.test`, without them the target doesn't work, and execution is error-prone as `metricbeat.test` is not regenerated if changes are done to go code (or maybe I shouldn't be using this target directly?)


ruflin:
@jsoriano Thanks a lot for you contribution. Any chance we could separate the PR into 2 PR's. One with the changes to the testing / Makefile and the other one with the Stat change. So we can discuss these changes seperately and one does not hold back the other.

jsoriano:
@ruflin sure

metricbeat/module/haproxy/haproxy.go
```diff
@@ -243,3 +220,84 @@ func (c *Client) GetInfo() (*Info, error) {
...
+
+	d, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("couldn't read response body: %v", err)
+	}
+	return bytes.NewBuffer(d), nil
+}
+
+func (p *httpProto) Info() (*bytes.Buffer, error) {
+	return nil, errors.New("not supported")
```
ruflin: That is interesting. So http does not support the `info` request? We should probably state that in the docs somewhere.

metricbeat/tests/system/test_haproxy.py
```diff
@@ -10,17 +10,7 @@ class Test(metricbeat.BaseTest):
...
-        """
-        haproxy info metricset test
-        """
-        self.render_config_template(modules=[{
-            "name": "haproxy",
-            "metricsets": ["info"],
-            "hosts": self.get_hosts(),
-            "period": "5s"
-        }])
+    def _test_info(self):
```
ruflin: Why did you add an `_` in front and remove the intergration part line?

jsoriano:
Changes on tests moved to #5820

ruflin:
Thanks a lot for taking this. Only had a quick look so far but that looks really promising. Will have a deeper look later.

metricbeat/module/haproxy/haproxy.go
jsoriano: It seems so :/ I couldn't find a way to do it (apart of parsing the human stats html), and according to the code the dump info function is only called when executing the `show info` command https://github.com/haproxy/haproxy/blob/master/src/stats.c#L3130.

Ok, I'll review docs in the repo.

metricbeat/tests/system/test_haproxy.py
jsoriano: I moved common code to "internal" functions so it can be reused for multiple cases. At the end this test has only the tcp socket case (it could maybe have also another case for a unix socket file), but the stat test has multiple cases.

metricbeat/module/haproxy/haproxy.go
```diff
@@ -243,3 +220,84 @@ func (c *Client) GetInfo() (*Info, error) {
...
+}
+
+type httpProto struct {
+	URL *url.URL
+}
+
+func (p *httpProto) Stat() (*bytes.Buffer, error) {
+	url := p.URL.String()
+	// Force csv format
+	if !strings.HasSuffix(url, ";csv") {
```
ruflin: What other formats do we have available?

metricbeat/module/haproxy/haproxy.go
```diff
@@ -243,3 +220,84 @@ func (c *Client) GetInfo() (*Info, error) {
...
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("couldn't connect: %v", err)
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("invalid response: %s", resp.Status)
+	}
+
+	d, err := ioutil.ReadAll(resp.Body)
```
ruflin: I wonder if we need here afterwards `resp.Body.Close()` to not leak the resource?

metricbeat/tests/system/test_haproxy.py
ruflin: Got it. The diff tricked me here :-)

metricbeat/module/haproxy/haproxy.go
jsoriano: Only human-oriented html, so csv is the only reasonable one to parse :)

metricbeat/module/haproxy/haproxy.go
jsoriano: Ouch, you are right, good catch

jsoriano: Close added

metricbeat/module/haproxy/haproxy.go
jsoriano: Doc and line to changelog added.

metricbeat/module/haproxy/haproxy.go
```diff
@@ -137,60 +139,37 @@ type Info struct {
...
+		return nil, errors.New("invalid url")
 	}
 
-	if strings.HasPrefix(response.String(), "Unknown command") {
-		return response, fmt.Errorf("unknown command: %s", cmd)
+	switch u.Scheme {
+	case "tcp":
+		return &Client{&unixProto{Network: u.Scheme, Address: u.Host}}, nil
+	case "unix":
+		return &Client{&unixProto{Network: u.Scheme, Address: u.Path}}, nil
```
jsoriano: I have modified this, before the url was parsed just splitting protocol and address, but now after the change to use url package for url parsing the address is `Host` for tcp, but `Path` for unix.

In any case, even if supported, there was no test coverage for unix socket, this will have to be revisited, as a volume in the docker would be needed to test this.

metricbeat/module/haproxy/haproxy.go
```diff
@@ -243,3 +222,86 @@ func (c *Client) GetInfo() (*Info, error) {
...
+		url += ";csv"
+	}
+
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	if p.URL.User != nil {
+		password, _ := p.URL.User.Password()
```
ruflin: I wonder if we should check this error here? Is no password = error?

metricbeat/module/haproxy/haproxy.go
```diff
@@ -137,60 +139,37 @@ type Info struct {
...
-		return response, fmt.Errorf("unknown command: %s", cmd)
+	switch u.Scheme {
+	case "tcp":
+		return &Client{&unixProto{Network: u.Scheme, Address: u.Host}}, nil
+	case "unix":
+		return &Client{&unixProto{Network: u.Scheme, Address: u.Path}}, nil
+	case "http", "https":
+		return &Client{&httpProto{URL: u}}, nil
+	default:
+		return nil, errors.New("invalid protocol scheme")
```
andrewkroh: Include the invalid value in the error message with errors.Errorf.

metricbeat/module/haproxy/haproxy.go
jsoriano: Exactly :D

metricbeat/module/haproxy/_meta/docs.asciidoc
```diff
@@ -1,13 +1,30 @@
...
+ stats socket 127.0.0.1:14567
 
 NOTE: You should use an internal private IP, or secure this with a firewall
 rule, so that only designated hosts can access this data.
 
+To configure the HTTP stats frontend, a frontend with stats enabled has to
+be added. For example, to open this frontend to any IP on port 14567 with
+required authentication add this to the haproxy config:
+
+ listen stats
```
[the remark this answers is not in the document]
jsoriano: It is rendered by github as code just leaving spaces before, the same with the previous snippet. Should I better use code blocks in any case?

metricbeat/module/haproxy/haproxy.go
jsoriano: The ignored value is a boolean telling if the password was set, I don't know if there are cases where basic authentication would work with an empty password... In any case I think it can be fine as is, if the credentials are wrong, the request will fail and this error is checked.

metricbeat/module/haproxy/haproxy.go
```diff
@@ -137,60 +139,37 @@ type Info struct {
...
-	_, err = conn.Write([]byte(cmd + "\n"))
-	if err != nil {
-		return response, err
-	}
-
-	_, err = io.Copy(response, conn)
+	u, err := url.Parse(address)
 	if err != nil {
-		return response, err
+		return nil, errors.New("invalid url")
```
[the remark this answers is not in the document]
jsoriano: Ok.

metricbeat/module/haproxy/haproxy.go
jsoriano: Ok, using `errors.Errorf` in the rest of the file too.

andrewkroh:
jenkins, test it

metricbeat/module/haproxy/_meta/docs.asciidoc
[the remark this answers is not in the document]
ruflin: Yes, please use code blocks as for our docs build Github is not really the one which counts in the end. We use code block in hopefully most other places.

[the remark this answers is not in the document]
jsoriano: Ok :)

ruflin:
jenkins, test it

ruflin:
@jsoriano Can you run `make fmt` and push again? That should make CI happy.

jsoriano:
Pushed changes after `make fmt` and `make collect-docs` and CI seems happier :)

ruflin:
jenkins, test it

metricbeat/module/haproxy/_meta/docs.asciidoc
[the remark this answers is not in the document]
jsoriano: I have set `haproxy` as "language", if empty, `make docs` fails.

ruflin:
jenkins, test it

ruflin:
@jsoriano Merged. Thanks for all your work on this one.

metricbeat/module/haproxy/haproxy.go
jsoriano: JSON has been added to haproxy 1.8, but we may continue using CSV for backwards compatibility.

ruflin: Sounds like something we should probably add more details about to the docs.

PR #5820, opened 2017-12-06
Some fixes on tests

jsoriano:
It tries to fix a couple of problems seen when running tests:
* Configuration file created with umask 0177 so it has proper permissions for metricbeat
* `test-module` make target depends on `fields` and `metricbeat.test`, without them the target doesn't work, and execution is error-prone as `metricbeat.test` is not regenerated if changes are done to go code (or maybe I shouldn't be using this target directly?)


libbeat/tests/system/beat/beat.py
```diff
@@ -207,8 +207,11 @@ def render_config_template(self, template_name=None,
 
         kargs["beat"] = self
         output_str = template.render(**kargs)
+
+        prev_umask = os.umask(0177)
```
ruflin: @andrewkroh Could you have a look at this change? WDYT?

metricbeat/Makefile
```diff
@@ -90,6 +90,6 @@ run-module:
 
 .PHONY: test-module
 test-module: ## @testing Tests the given module. Needs $MODULE as param an run-module must be started first.
-test-module: python-env
+test-module: python-env fields metricbeat.test
```
ruflin: Reminds me of https://github.com/elastic/beats/pull/4961 which strangely fails on travis. I wonder if it should be more then just `fields` but `update`.

jsoriano: Ok, using `update` instead.

libbeat/tests/system/beat/beat.py
andrewkroh: Rather than setting a temporary umask, how about we write the file with the permissions that we want. Can we do something like this: https://stackoverflow.com/a/5624691/503798 ?

metricbeat/Makefile
```diff
@@ -90,6 +90,6 @@ run-module:
 
 .PHONY: test-module
 test-module: ## @testing Tests the given module. Needs $MODULE as param an run-module must be started first.
-test-module: python-env
+test-module: python-env update metricbeat.test
```
ruflin: The part I worry now is that `test-module` was thought of a quick way to run the tests again and again during development and now each time update and the build process is run. But I definitively see your point that in case the code is changed, the above is needed anyways. LGTM for now :-)

libbeat/tests/system/beat/beat.py
jsoriano: Yes, better doing it explicitely, doing it with a chmod, I think it's more clear this way.

metricbeat/Makefile
jsoriano: Maybe a new target could be added later with the specific things needed by `test-module`, for example is it needed to update the docs to this quick tests on modules?

ruflin:
jenkins, test it

metricbeat/Makefile
ruflin: Yes, that is a good idea. The other thing we are working on is converting more and more things of `update` which are in python to Golang, means the python-env is not needed anymore which makes things much quicker.

libbeat/tests/system/beat/beat.py
```diff
@@ -207,7 +207,10 @@ def render_config_template(self, template_name=None,
 
         kargs["beat"] = self
         output_str = template.render(**kargs)
-        with open(os.path.join(self.working_dir, output), "wb") as f:
+
+        output_path = os.path.join(self.working_dir, output)
+        with open(output_path, "wb") as f:
+            os.chmod(output_path, 0600)
```
ruflin: @andrewkroh Looks good to you?

libbeat/tests/system/beat/beat.py
```diff
@@ -198,7 +198,7 @@ def render_config_template(self, template_name=None,
         if template_name is None:
             template_name = self.beat_name
 
-        template_path = "./tests/system/config/" + template_name + ".yml.j2"
+        template_path = os.path.join("./tests/system/config", template_name + ".yml.j2")
```
ruflin: Thanks fo rthe cleanup

ruflin:
@jsoriano Thanks for the contribution.

libbeat/tests/system/beat/beat.py
ruflin: Interestingly this was working before with "only" forward slashes but now that we start mixing forward and backward slashes, it breaks on windows. Will open a follow up PR.

PR #5822, opened 2017-12-06
Make WatcherDie tests less flaky

ruflin:
Now the test checks up to 10s if the containers were actually updated.


exekias:
Thank you for tackling this

PR #5825, opened 2017-12-06
Cherry-pick #5808 to 6.1: Fix isolation of modules when merging local and global field settings.

urso:
Cherry-pick of PR #5808 to 6.1 branch. Original message: 

Resolves: #5795 

The user global and local fields configurations must be merged
on pipeline setup. The DeepUpdate method did only keep references to the
globally shared fields, such that local fields settings do overwrite
the shared global state. If more then one beat.Client instance did set
local fields, the global state (and all fields added to an event) will
be given by the last client being instantiated.

The fix Clones the global fields before merging, such that the
beat.Client fields operate fully in isolation.


ruflin:
@urso Needs rebase because of the CHANGELOG.

urso:
rebased

PR #5828, opened 2017-12-06
Show separate commands for deb/rpm and mac

dedemorton:
The documentation was slightly incorrect wrt running the setup commands on rpm and deb because it showed a mac-specific syntax. This PR fixes the issue by adding a separate section for rpm and deb.

I've also done some refactoring of the asciidoc attributes (to move them to a shared file) because I was originally planning to add a new attribute. I decided that a new attribute was not required, but left the changes in this PR because it makes sense to centralize shared attributes.

I left the name and version-related attributes in the index files because they are either specific to each beat or I thought we might want more granular control over setting those attributes at some point in the future. Please let me know if I've put anything in the shared file that is not likely to be the same across all Beats in the future.

### DIFF
--- a/libbeat/docs/dashboards.asciidoc
+++ b/libbeat/docs/dashboards.asciidoc
@@ -28,25 +28,32 @@ Make sure Kibana is running before you perform this step. If you are accessing a
 secured Kibana instance, make sure you've configured credentials as described in
 <<{beatname_lc}-configuration>>.
 
-To set up the Kibana dashboards for {beatname_uc}:
+To set up the Kibana dashboards for {beatname_uc}, use the appropriate command
+for your system.
 
 ifdef::allplatforms[]
 
-*deb, rpm, and mac:*
+ifeval::["{requires-sudo}"=="yes"]
 
-From the directory where you installed {beatname_uc}, run:
+include::../../libbeat/docs/shared-note-sudo.asciidoc[]
 
-["source","sh",subs="attributes,callouts"]
+endif::[]
+
+*deb and rpm:*
+
+["source","sh",subs="attributes"]
+----------------------------------------------------------------------
+{beatname_lc} setup --dashboards
+----------------------------------------------------------------------
+
+
+*mac:*
+
+["source","sh",subs="attributes"]
 ----------------------------------------------------------------------
 ./{beatname_lc} setup --dashboards
 ----------------------------------------------------------------------
 
-ifeval::["{requires-sudo}"=="yes"]
-
-If you changed ownership of the config file to root, you'll need preface this
-command with `sudo`.
-
-endif::[]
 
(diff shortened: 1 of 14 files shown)


dedemorton:
Marked as in-progress because I still need to fix the examples for running metricbeat modules.

dedemorton:
Actually, I'm going to fix the metricbeat command examples in a separate PR because I need to figure out the best way to show commands that work for all users.

ruflin:
CI does not seem happy. It fails with the following error:

```
20:09:22 asciidoc: WARNING: shared-template-load.asciidoc: line 92: include file not found: /var/lib/jenkins/workspace/elastic+beats+pull-request+multijob-docs/src/github.com/elastic/beats/libbeat/docs/shared-note-sudo.asciidoc
20:09:22 asciidoc: WARNING: dashboards.asciidoc: line 38: include file not found: /var/lib/jenkins/workspace/elastic+beats+pull-request+multijob-docs/src/github.com/elastic/beats/libbeat/docs/shared-note-sudo.asciidoc
20:09:22 ../libbeat/scripts/Makefile:321: recipe for target 'docs' failed
```

dedemorton:
@ruflin Whoops...forgot to check in the file. Added it.  _hangs head in shame_...

libbeat/docs/dashboards.asciidoc
```diff
@@ -28,25 +28,32 @@ Make sure Kibana is running before you perform this step. If you are accessing a
...
 ----------------------------------------------------------------------
-./{beatname_lc} setup --dashboards
+{beatname_lc} setup --dashboards
 ----------------------------------------------------------------------
 
-ifeval::["{requires-sudo}"=="yes"]
 
-If you changed ownership of the config file to root, you'll need preface this
-command with `sudo`.
+*mac:*
```
exekias: This is also valid for Linux on `tar.gz` scenarios, maybe we should say `tar.gz` instead of `mac`?

dedemorton: I'll just say mac and linux to match what we say on the product download pages.

dedemorton: @exekias Actually, if I add linux to the topics here, I should also add Linux to the installation steps in the getting started. Then I'll have to decide whether I want to cover 32 or 64 bit. The docs start to get more complicated when we cover more platforms. I'm not saying we shouldn't...just that there are implications. WDYT?

ruflin: I think if we specify `mac` we will need linux as this is still the most common platform I think. I wonder if we could build this in a more modular way as it will be pretty tricky to keep it all up-to-date.

dedemorton:
@ruflin I've rebased.

ruflin:
@dedemorton Merged. Not sure what our follow up for the discussion we had about the linux builds is ...

PR #5835, opened 2017-12-07
Make Metricbeat framework more configurable

andrewkroh:
This makes the Metricbeat framework more flexible by allowing developers to specify options when instantiating Metricbeat. This adds an option for controlling the event format and whether or not the traditional `metricset` info object is added to each event that contains the metricset name, module name, host, and round-trip-time (rtt).

Additionally this adds an Event struct that MetricSets can use to deliver their data. The Event struct has fields for all data that was previously sent in one `MapStr` by using dedicated key names like `_namespace`, `_rtt`, `_module`, and `@timestamp`. This makes the MetricSet code a bit more clear.

Because of the introduction of the `mb.Event` struct I'm deprecating the `mb.Reporter` interface that contained methods for reporting data using a `MapStr`. The `mb.Reporter` interface defines the methods `Event(MapStr)`, `Error(error)`, and `ErrorWith(error, MapStr)`. This was added recently to enable push metricsets so it isn't widely used yet. After a short deprecation period this interface will be removed.  This interface is replaced with `mb.ReporterV2` that contains `Event(mb.Event)` and `Error(error)`.


ruflin:
@andrewkroh There are some tests failing which I think are related to these changes.

metricbeat/scripts/module/metricset/metricset.go.tmpl
```diff
@@ -42,15 +39,14 @@ func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
...
-// descriptive error must be returned.
-func (m *MetricSet) Fetch() (common.MapStr, error) {
-
-	event := common.MapStr{
-		"counter": m.counter,
-	}
+// Fetch methods implements the data gathering and data conversion to the right
+// format. It publishes the event which is then forwarded to the output. In case
+// of an error set the Error field of mb.Event or simply call report.Error().
+func (m *MetricSet) Fetch(report mb.ReporterV2) {
```
ruflin: As far as I understand, this would not have to be changed even after the deprecation? The old `Fetch()` which is used in most modules stays? As `reporterV2` can do everything that `Fetch` can do (I think) I wonder if we should even get rid of the `Fetch()` interface and only use reporter? It makes the implementation a bit less strict to only return one event at the end, but perhaps having less options to implement things is good?

andrewkroh: Have less options is a good thing from a code maintainability POV.

I didn't want to go as far as labelling the old `Fetch() MapStr` and `Fetch() []MapStr` methods as deprecated now, but that's an option. It would mean we have *lot's* of code to update.

I think we should use this new `mb.Event` with the ReporterV2 interface for a little bit and see if we like the code flow. If it's good then we can seriously consider deprecating the original Fetch methods.

ruflin: Was not thinking of deprecating it in this PR but more as a discussion we should perhaps have.

andrewkroh:
jenkins, retest it

andrewkroh:
jenkins, retest it  (there's an unrelated Filebeat failure)

metricbeat/mb/module/wrapper.go
```diff
@@ -243,27 +250,17 @@ func (msw *metricSetWrapper) String() string {
...
-			return
-		}
-
-		reporter := &testingReporter{
-			driver: d,
-			done:   done,
-		}
-		msw.fetch(reporter)
+		events := make(chan beat.Event, 1)
+		done := receiveOneEvent(d, events, msw.module.maxStartDelay+5*time.Second)
```
alexmo16: Hardcoding a timeout is not a best practise. Please make it configurable in a config file perhaps.

andrewkroh: There is a configurable timeout associated with each module. We should ensure that value instead and ensure the value is initialized with a default. Do you want to open a PR to change this?

https://github.com/elastic/beats/blob/bc4b07c922d57bdafd3bc12e9e23f190c2592a09/metricbeat/mb/mb.go#L287

PR #5840, opened 2017-12-08
Fix template path for system tests

ruflin:
This broke by the change in #5820 for Windows. This reverts it to the old format as it seems `self.template_env.get_template(...)` expects a path without forward slashes under Windows too.

### DIFF
--- a/libbeat/tests/system/beat/beat.py
+++ b/libbeat/tests/system/beat/beat.py
@@ -198,7 +198,7 @@ class TestCase(unittest.TestCase, ComposeMixin):
         if template_name is None:
             template_name = self.beat_name
 
-        template_path = os.path.join("./tests/system/config", template_name + ".yml.j2")
+        template_path = "./tests/system/config/" + template_name + ".yml.j2"
 
         if output is None:
             output = self.beat_name + ".yml"


libbeat/tests/system/beat/beat.py
```diff
@@ -198,7 +198,7 @@ def render_config_template(self, template_name=None,
         if template_name is None:
             template_name = self.beat_name
 
-        template_path = os.path.join("./tests/system/config", template_name + ".yml.j2")
+        template_path = "./tests/system/config/" + template_name + ".yml.j2"
```
jsoriano: Maybe `normpath` or `normcase` can be used to cover Windows? Something like this:
```
template_path = os.path.join(os.path.normpath("./tests/system/config"), template_name + ".yml.j2")
```

ph: [os.path.normpath](https://docs.python.org/2/library/os.path.html#os.path.normpath) should work or :

```python
template_path  = os.path.join(".", "tests", "system", "config", template_name + ".yml.j2")
```

ph: Oh I see that you tested the second suggestion and it failed @ruflin, so normpath should work in this case.

ruflin: @ph  @jsoriano It seems `self.template_env.get_template(` expects a path without forward slashes. I revert it to what we had initially to bring our tests back to green. I tried `normpath` etc. and it did not work.

Issue #4254, opened 2017-05-08
[Filebeat][Doc] Elasticsearch output - loading index template missing index name option


sherry-ger:
- Version:  Filebeat 5.4.0

The option to set custom index name while configuring Elasticseach template is not included in the documentation - https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-template.html#load-template-auto

As a part of configuring custom index template, it is common to add ones' own index name as well.  The option is present and documented in the default filebeat.full.yml file.

```
  # Optional index name. The default is "filebeat" plus date
  # and generates [filebeat-]YYYY.MM.DD keys.
  #index: "filebeat-%{+yyyy.MM.dd}"
```

Please add this as a part of the filebeat get started guide as well.

PR #5841, opened 2017-12-08
Fix docs for index naming and loading

dedemorton:
Fixes https://github.com/elastic/beats/issues/4254 plus some examples that we missed when we updated the index name to include version info.

I've added info about setting the index name to the getting started, but I wonder if it really belongs there (as opposed to the reference topic). I think we need to make larger changes to the structure of the GS topics, so I'll leave this as-is for now. 

IMO, changing the index name is a bit more work than it should be (not ideal for the user because it's easy to miss something).

### DIFF
--- a/libbeat/docs/config-file-format.asciidoc
+++ b/libbeat/docs/config-file-format.asciidoc
@@ -77,11 +77,11 @@ For example this setting:
 
 output:
   elasticsearch:
-    index: 'beat-%{+yyyy.MM.dd}'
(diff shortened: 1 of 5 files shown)


ruflin:
@dedemorton Agree that it's kind of tricky to change the index name at the moment. The initial approach contained some magic to make it easier but had too many unexpected side effects. We are aware of it but don't have a full solution for it yet.

PR #5842, opened 2017-12-08
Allow ES search queries with bodies.

simitt:


libbeat/outputs/elasticsearch/api.go
```diff
@@ -168,7 +168,18 @@ func (es *Connection) DeletePipeline(
...
+
+// SearchURIWithBody executes a search request using a URI by providing request
+// parameters and a request body.
+func (es *Connection) SearchURIWithBody(
+	index string,
+	docType string,
+	params map[string]string,
+	body interface{},
+) (int, *SearchResults, error) {
+	status, resp, err := es.apiCall("GET", index, docType, "_search", "", params, body)
```
ph: @simitt Code looks good nice refactoring, even if _SearchURIWithBody_ is tested indirectly with  _SearchURI_ tests, could you add a direct call to it in _api_interation_test.go_, I am worried that it gets deleted somehow since we don't depend on it directly but APM does.

simitt:
@ph I added a test and rebased to master.

ph:
LGTM, WFSG

PR #5844, opened 2017-12-08
Add support to set the document ID in the filebeat json reader

urso:
Requires: #5811 

This PR adds support to configure a key for setting the document ID in the harvester JSON settings. The ID will be store in the events Meta["id"] for the output to pick up. With #5811 will the elasticsearch output is the Meta["id"] field to set the document its ID (uses op_type="create" to count duplicate inserts of same ID). For other output type, the document ID will be forwarded via `@metadata.id`.


filebeat/docs/filebeat-options.asciidoc
```diff
@@ -534,4 +538,4 @@ When using `docker` prospector type you must define `containers.ids`, these are
 
 *`ids`*:: Required, the list of Docker container IDs to read logs from, `'*'` can be used as ID to read from all containers.
 
-*`path`*:: Base path where Docker logs are located. The default is `/var/lib/docker/containers`.
\ No newline at end of file
+*`path`*:: Base path where Docker logs are located. The default is `/var/lib/docker/containers`.
```
ruflin: What did change on this line? I think I'm blind.

filebeat/harvester/reader/json.go
```diff
@@ -93,12 +93,20 @@ func createJSONError(message string) common.MapStr {
 // respecting the KeysUnderRoot and OverwriteKeys configuration options.
 // If MessageKey is defined, the Text value from the event always
 // takes precedence.
-func MergeJSONFields(data common.MapStr, jsonFields common.MapStr, text *string, config JSONConfig) time.Time {
+func MergeJSONFields(data common.MapStr, jsonFields common.MapStr, text *string, config JSONConfig) (interface{}, time.Time) {
```
ruflin: Should we require the `id` to be a string?

filebeat/harvester/reader/json_test.go
```diff
@@ -304,6 +305,13 @@ func TestAddJSONFields(t *testing.T) {
 			},
 			ExpectedTimestamp: time.Time{},
 		},
+		{
+			// if document_id is set, extract the ID from the event
+			Name:       "extract event id",
+			Data:       common.MapStr{"@timestamp": common.Time(now), "json": common.MapStr{"id": "test_id"}},
```
ruflin: What happen if `id` is not a string?

ruflin:
@urso After all the renaming, this needs a rebase.

urso:
@ruflin rebased. Also update PR to enforce 'id' to be a string.

ruflin:
This seems to break quite a few tests. 

Other note: We will also need to add a doc entry for this. Part of this PR or follow up issue?

filebeat/harvester/reader/json_test.go
urso: Adde a test for "id" not being a string.
Reading ID is similar to reading "MessageKey". If it's missing we return an empty string.

urso:
> Other note: We will also need to add a doc entry for this. Part of this PR or follow up issue?

The PR already modifies the filebeat json config docs

urso:
There is some minimal doc update in this PR. But there are some other places throughout the docs where we want to document this feature. Labeling as needs_doc and creating doc issue to follow up.

PR #5846, opened 2017-12-08
Add faq about disabling ssl to capture mysql data

dedemorton:


packetbeat/docs/faq-mysql-ssl.asciidoc
```diff
@@ -0,0 +1,22 @@
...
+-----
+packetbeat.interfaces.device: lo0
+-----
+
+* Make sure the client programs you are monitoring run `mysql` with SSL
+disabled. For example:
++
+[source,shell]
+-----
+mysql --protocol tcp --host=127.0.0.1 --port=3306 --ssl-mode=DISABLED
```
ruflin: Should we add a note that this is a security risk?

dedemorton: @ruflin Hmmm. I'm not what we should say here because it seems odd to be telling users to do something that might expose them to a security risk. Any suggestions for the best wording around this? Should we tell users that they should not set ssl-mode to DISABLED if they are sending sensitive data?

ruflin: @andrewkroh WDYT?

andrewkroh: I think I'd make it clear that Packetbeat can only monitor the MySQL traffic if it is unencrypted. If you want to monitor the data with Packetbeat then you must disable encryption. I like the wording in their docs.

> "With an unencrypted connection between the MySQL client and the server, someone with access to the network could watch all your traffic and inspect the data being sent or received between client and server." - https://dev.mysql.com/doc/mysql-security-excerpt/5.7/en/encrypted-connections.html

So if the network is untrusted then it isn't advisable to disable encryption.

dedemorton: @andrewkroh I like the wording, too, but of course I can't reuse it since it's copyrighted. :-) So I've written a note in my own words. Please review.

packetbeat/docs/faq-mysql-ssl.asciidoc
```diff
@@ -0,0 +1,27 @@
...
+-----
+
+IMPORTANT: When SSL is disabled, the connection between the MySQL client and
+server is unencrypted, which means that anyone with access to your network may
+be able to inspect data sent between the client and server. If MySQL is running
+in an untrusted network, it's not advisable to disable encryption.
```
dedemorton: Or if we ant to be even more direct, we can say:

If MySQL is running in an untrusted network, disabling encryption poses a security risk.

packetbeat/docs/faq-mysql-ssl.asciidoc
dedemorton: Or if we ant to be even more direct, we can say:

If MySQL is running in an untrusted network, disabling encryption poses a security risk.

dedemorton: @andrewkroh What do you like better? ^^

andrewkroh: I like what you have in the PR.

PR #5847, opened 2017-12-08
Make the close_timeout test bit more reliable

ph:
The current implementation was failing on fairly new hardware, the test
is changed a bit and should be more reliable concerning the timing
issue.

Before starting filebeat, we add a few initial messages into the logs, so
the logs get pickup up in the first scan, then we sleep a little longer
than a frequency scan. After we write a few more lines to the logs.

To make sure that logs are available we force a flush and a sync on
disk.

### DIFF
--- a/filebeat/tests/system/test_harvester.py
+++ b/filebeat/tests/system/test_harvester.py
@@ -405,6 +405,9 @@ class Test(BaseTest):
         for n in range(0, iterations1):
             file.write("example data")
             file.write("\n")
+            # Make sure some contents are written to disk so the harvested is able to read it.
+            file.flush()
+            os.fsync(file)
             time.sleep(0.001)
 
         file.close()


ph:
jenkins test this please

filebeat/tests/system/test_harvester.py
```diff
@@ -421,7 +435,7 @@ def test_close_timeout(self):
         assert len(data) == 1
 
         # Check that not all but some lines were read. It can happen sometimes that filebeat finishes reading ...
-        assert self.output_lines() <= 1000
+        assert self.output_lines() <= 50
```
ruflin: This was set to 1000 to make sure in most cases 1000 is not reached. With a value of only 50 I think the likelyhood that all lines are read is very high which is not the intention of this test.

ph: Lets sync on that, I think the test was not exactly what you want before.

ph: I've talked to @ruflin I will rework that.

ph:
@ruflin This test pass every time now.

ruflin:
@ph Thanks for taking care of this.

PR #5856, opened 2017-12-11
Document modules SSL parameters

### LINKED ISSUE #4268: Document authentication/SSL support for http, jolokia and other metricbeats
Metricbeat modules like http and jolokia actually already support authentication in the product today.   This means that we have some stale documentation to update :)

* https://www.elastic.co/guide/en/beats/metricbeat/master/metricbeat-metricset-jolokia-jmx.html#_limitations (For Jolokia, it still talks about not having authentication support as a limitation)
* Documentation for both modules do not provide info on how users will pass in authentication information.  If this is done via passing in authorization information in the request headers, it will be nice to provide an example
  * https://www.elastic.co/guide/en/beats/metricbeat/master/metricbeat-module-jolokia.html
  * https://www.elastic.co/guide/en/beats/metricbeat/master/metricbeat-module-http.html
* While we are at it, if these modules also support SSL configuration, it will be helpful to document the available configuration settings as well.

### LINKED ISSUE #5827: [Metricbeat] Elasticsearch Module settings should document TLS / SSL support
The Elasticsearch Metricbeat module supports TLS / SSL, you can apparently specifiy the same settings as the Elasticsearch output (but under the module's configuration).

From @andrewkroh:

> Pretty much any of the modules that are HTTP based all share the same code and all support the same SSL options.

We should probably make a common set of TLS / SSL options that documentation can inherit / insert into their pages so that we don't have to repeat this everywhere.

exekias:
This PR adds a pointer to SSL settings doc to all modules that support it, it's defined in module `fields.yml` file. I kept space for future common settings parameters, just in case we add something after SSL.

Closes #5827 
Closes #4268

### DIFF
--- a/metricbeat/scripts/docs_collector.py
+++ b/metricbeat/scripts/docs_collector.py
@@ -83,6 +83,10 @@ in <<configuration-metricbeat>>. Here is an example configuration:
 
             module_file += "----\n\n"
 
+        # HTTP helper
+        if 'ssl' in get_settings(module_fields):
+            module_file += "This module supports TLS connection when using `ssl` config field, as described in <<configuration-ssl>>.\n\n"
+
         # Add metricsets title as below each metricset adds its link
         module_file += "[float]\n"
         module_file += "=== Metricsets\n\n"
@@ -208,7 +212,6 @@ For a description of each field in the metricset, see the
 
 
 def get_release(fields):
-
     # Fetch release flag from fields. Default if not set is experimental
     release = "experimental"
     if "release" in fields:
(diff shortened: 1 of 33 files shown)


metricbeat/scripts/docs_collector.py
```diff
@@ -83,6 +83,10 @@ def collect(beat_name):
 
             module_file += "----\n\n"
 
+        # HTTP helper
```
ruflin: @exekias Awesome. Was thinking of exactly that when I saw the notification for the PR.

dedemorton:
Thanks for adding this!

PR #5857, opened 2017-12-11
Exclude stress test on Windows as it currently breaks the build

ruflin:
As on Windows currently all packages are included for testing, the -E flag is discovered twice. This excludes the stress_pipeline file to bring the builds / tests back to green. Windows tests should also use specific packages https://github.com/elastic/beats/blob/master/dev-tools/jenkins_ci.ps1#L55 This is mainly a temporary fix.


ruflin:
This needs rebase after https://github.com/elastic/beats/pull/5840 is merged to see if tests go green actually.

PR #5866, opened 2017-12-12
Update vendored fsevents to use elastic fork

adriansr:
This PR makes vendored fsvents to use the changes in our fork while keeping the same import path.


andrewkroh:
According to jenkins, this needs a `make notice` to update the NOTICE.txt.

PR #5870, opened 2017-12-13
TLS certificates include public key size

### LINKED ISSUE #5804: Packetbeat TLS: Report the key size used for certificates
This is a request to enhance the Packetbeat TLS module to report key sizes used in certificates.

As a user I want to be able to alert when I find a service on my network using weak key sizes.

adriansr:
This adds a `public_key_size` entry to certificates that contains
the number of bits in the public key. RSA, DSA and ECDSA public keys
are supported. The value of this field is zero if the key size
cannot be determined.

Closes  #5804

### DIFF
--- a/packetbeat/protos/tls/parse.go
+++ b/packetbeat/protos/tls/parse.go
@@ -1,6 +1,9 @@
 package tls
 
 import (
+	"crypto/dsa"
+	"crypto/ecdsa"
+	"crypto/rsa"
 	"crypto/x509"
 	"crypto/x509/pkix"
 	"encoding/hex"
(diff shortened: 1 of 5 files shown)


packetbeat/protos/tls/parse.go
```diff
@@ -488,10 +491,40 @@ func (version tlsVersion) String() string {
...
+		}
+	}
+	return 0
+}
+
 func certToMap(cert *x509.Certificate, includeRaw bool) common.MapStr {
 	certMap := common.MapStr{
 		"signature_algorithm":  cert.SignatureAlgorithm.String(),
 		"public_key_algorithm": toString(cert.PublicKeyAlgorithm),
+		"public_key_size":      getKeySize(cert.PublicKey),
```
andrewkroh: WDYT about not adding the field if we can't determine the value? I'm wondering if that would make it easier to use.

adriansr: Yes that makes more sense.

PR #5878, opened 2017-12-14
elasticsearch: show event being indexed in case of failure

7AC:
If there are multiple failures it is useful to keep track of them.


exekias:
ok to test

7AC:
Thanks @exekias! ;)

Issue #4285, opened 2017-05-10
Metricbeat Kafka module does not publish consumergroup metrics (Cluster mode)


rahuljoshitaulia:
**Test setup:**

1. Setup a kafka and zookeeper cluster for HA using 3 kafka nodes and 3 zookeeper nodes. Apache open source documentation is available on their website on how to set this up. Use latest kafka version. Make sure kafka is configured to store offsets on kafka broker and not on zookeeper. 
2. Install metricbeat 5.3.2 on each kafka node
3. Configure each metricbeat installation to collect`partition` and `consumergroup` metricsets. Send metrics to elastic search. turn on debug logging in the metricbeats configuration for *
4. Setup Logstash server to produce and consume from the kafka server. Produce to 1 one topic with 10 partitions and and startup 4 consumer groups consuming from that topic. You will have to change the logstash configuration provided below to use the parameter `bootstrap_servers => "server1,server2,server3"` so that logstash connects to the cluster.  

Sample logstash configuration for producer:
```
input {
  generator {}
}

filter {
  sleep {
    time => "1"
    every => 1000
  }
}

output {
  kafka {
    client_id => "lsp"
    topic_id => "test1"
  }
}
```
Sample logstash configuration for consumer:
```
input {
  kafka {
    group_id => "lsc1"
    topics => "test1"
  }
  kafka {
    group_id => "lsc2"
    topics => "test1"
  }
  kafka {
    group_id => "lsc3"
    topics => "test1"
  }
  kafka {
    group_id => "lsc4"
    topics => "test1"
  }
  kafka {
    group_id => "lsc5"
    topics => "test1"
  }
}

output {
  null {}
}
```

**Test Results:**
You will notice in the metricbeats logs that you are receiving events for the `partition` metricset but not for the `consumergroup` metricset. 

For a more detailed discussion of this issue you can check the following thread on the forum:
https://discuss.elastic.co/t/metricbeat-kafka-module-consumergroup-metric-set-does-not-report-metrics/83623

ruflin:
Thanks a lot for the detailed steps on how to reproduce it.

luckb008:
@ruflin Hi Nicolas. Do you have any ideas when this bug will be fixed?

huangchaosuper:
we get the same problem

oleksandrtub:
Got the same problem when updated Kafka cluster from 0.10.0.1 to 0.10.2.1 and reporting of consumer metrics stopped

[...]

PR #5880, opened 2017-12-14
metricbeat: fix the `consumergroup` of kafka module incorrectly work

wangdisdu:
The `consumergroup` fetch offset info by sending `OffsetFetch` request to Kafka.
But the `topic` and `partition` parameters missed in the sent `OffsetFetch` request.
These parameters is required, and kafka will do not response anything if they are missed.

This is Kafka protocol guide: http://kafka.apache.org/protocol.html#The_Messages_OffsetFetch

### DIFF
--- a/metricbeat/module/kafka/consumergroup/query.go
+++ b/metricbeat/module/kafka/consumergroup/query.go
@@ -1,8 +1,6 @@
 package consumergroup
 
 import (
-	"sync"
-
 	"github.com/Shopify/sarama"
 
 	"github.com/elastic/beats/libbeat/common"
@@ -13,7 +11,7 @@ import (
 type client interface {
 	ListGroups() ([]string, error)
 	DescribeGroups(group []string) (map[string]kafka.GroupDescription, error)
-	FetchGroupOffsets(group string) (*sarama.OffsetFetchResponse, error)
+	FetchGroupOffsets(group string, partitions map[string][]int32) (*sarama.OffsetFetchResponse, error)
 }
 
 func fetchGroupInfo(
@@ -22,9 +20,10 @@ func fetchGroupInfo(
 	groupsFilter, topicsFilter func(string) bool,
 ) error {
 	type result struct {
-		err   error
-		group string
-		off   *sarama.OffsetFetchResponse
+		err    error
+		group  string
+		assign map[string]map[int32]groupAssignment
+		off    *sarama.OffsetFetchResponse
 	}
 
 	groups, err := listGroups(b, groupsFilter)
@@ -38,51 +37,64 @@ func fetchGroupInfo(
 
 	debugf("known consumer groups: ", groups)
 
-	wg := sync.WaitGroup{}
-	results := make(chan result, len(groups))
-	for _, group := range groups {
-		group := group
+	assignments, err := fetchGroupAssignments(b, groups)
+	if err != nil {
+		logp.Err("failed to fetch kafka group assignments: %v", err)
+		return err
+	}
+	if len(assignments) == 0 {
+		return nil
+	}
 
-		wg.Add(1)
-		go func() {
-			defer wg.Done()
-			resp, err := fetchGroupOffset(b, group, topicsFilter)
+	results := make(chan result)
+	waiting := 0
+	for group, topics := range assignments {
+		// generate the map topic to partitions
+		queryTopics := make(map[string][]int32)
+		for topic, partitions := range topics {
+			if topicsFilter != nil && !topicsFilter(topic) {
+				continue
+			}
+
+			// copy partition ids
+			count := len(partitions)
+			if count == 0 {
+				continue
+			}
+
+			ids, i := make([]int32, count), 0
+			for partition := range partitions {
+				ids[i], i = partition, i+1
+			}
+			queryTopics[topic] = ids
+		}
+
+		if len(queryTopics) == 0 {
+			continue
+		}
+
+		// fetch group offset
+		waiting++
+		go func(group string, partitions map[string][]int32, assign map[string]map[int32]groupAssignment) {
+			resp, err := fetchGroupOffset(b, group, partitions)
 			if err != nil {
 				logp.Err("failed to fetch '%v' group offset: %v", group, err)
 			}
-			results <- result{err, group, resp}
-		}()
+			results <- result{err, group, assign, resp}
+		}(group, queryTopics, topics)
 	}
 
-	go func() {
-		wg.Wait()
-		close(results)
-	}()
-
-	assignments, err := fetchGroupAssignments(b, groups)
-	if err != nil {
-		// wait for workers to stop and drop results
-		for range results {
+	for waiting > 0 {
+		ret := <-results
+		waiting--
+		if ret.err != nil && err == nil {
+			err = ret.err
+		}
+		if err != nil {
+			continue
 		}
 
-		return err
-	}
-
-	for ret := range results {
-		if err := ret.err; err != nil {
-			// wait for workers to stop and drop results
-			for range results {
-			}
-			return err
-		}
-
-		asgnGroup := assignments[ret.group]
 		for topic, partitions := range ret.off.Blocks {
-			var asgnTopic map[int32]groupAssignment
-			if asgnGroup != nil {
-				asgnTopic = asgnGroup[topic]
-			}
-
 			for partition, info := range partitions {
 				event := common.MapStr{
 					"id":        ret.group,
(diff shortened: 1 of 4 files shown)


ruflin:
@wangdisdu Just wanted to let you know that we saw the PR. As @urso was thinking already a bit about how to fix Kafka I'm hoping he can have a look at it.

metricbeat/module/kafka/consumergroup/query.go
```diff
@@ -38,51 +39,55 @@ func fetchGroupInfo(
...
+
 	wg := sync.WaitGroup{}
-	results := make(chan result, len(groups))
-	for _, group := range groups {
-		group := group
+	results := make(chan result, len(assignments))
+
+	waiting := 0
+	for group, topics := range assignments {
+		_partitions := make(map[string][]int32)
```
urso: no underscore please.

in beats codebase we prefer to use make/new only if there is not other syntax pattern. Use

```
partitions = map[string][]int32{}
```

How about (reduce key lookups and appends/alloc to slice):

```
  queryTopics := map[string][]int32{}
  for topic, partitions := range topics {
    if topicFilter != nil && !topicFilter(topic) {
      continue
    }

    // copy partition ids
    ids, i := make([]int32, len(partitions)), 0
    for partition := range partitions {
      ids[i], i = partition, i+1
    }
    queryTopics[topic] = ids
  }

  if len(queryTopics) == 0 {
    continue
  }
 
  ...
```

metricbeat/module/kafka/consumergroup/query.go
```diff
@@ -38,51 +39,55 @@ func fetchGroupInfo(
...
+		for topic, partitions := range topics {
+			if topicsFilter == nil || topicsFilter(topic) {
+				for partitionID := range partitions {
+					if _, ok := _partitions[topic]; !ok {
+						_partitions[topic] = make([]int32, 0)
+					}
```
urso: This if-clause is not required. `_partitions[topic]` returns `nil`, if topic is not known. `append(([]int32)(nil), value)` returns a new `[]int32{value}`.

metricbeat/module/kafka/consumergroup/query.go
```diff
@@ -38,51 +39,55 @@ func fetchGroupInfo(
...
+	}
+	if len(assignments) == 0 {
+		return nil
+	}
+
 	wg := sync.WaitGroup{}
-	results := make(chan result, len(groups))
-	for _, group := range groups {
-		group := group
+	results := make(chan result, len(assignments))
```
urso: As the loop collecting data does not wait for wg.Wait anymore, but through all results, the wg is not really required anymore. We can even introduce an unbuffered channel `make(chan result)` now.

metricbeat/module/kafka/consumergroup/query.go
```diff
@@ -38,51 +39,55 @@ func fetchGroupInfo(
...
-		if err := ret.err; err != nil {
-			// wait for workers to stop and drop results
-			for range results {
-			}
-			return err
+	for waiting > 0 {
+		ret := <-results
+		waiting--
+		if ret.err != nil {
+			continue
```
urso: originally we did return an error -> we still want to return the first error we encounter
```
for waiting > 0 {
  ret := <- results
  waiting--
  if ret.err != nil {
    if err == nil {
      err = ret.err
    }
    continue
  }

  ...
}
if err != nil {
  return err
}

```

wangdisdu:
Hi @ruflin @urso Thanks for your reviewed. I hava adjusted the code:
1.  removed unecessary waitgroup
2. change underscore variable name
3. optimalize the

metricbeat/module/kafka/consumergroup/query.go
wangdisdu: fixed

metricbeat/module/kafka/consumergroup/query.go
wangdisdu: fixed

metricbeat/module/kafka/consumergroup/query.go
wangdisdu: fixed: optimalized the code generate partition array

wangdisdu: fixed: return the error if has

metricbeat/module/kafka/consumergroup/query.go
wangdisdu: fixed: return the error if has

metricbeat/module/kafka/consumergroup/query.go
```diff
@@ -38,51 +37,64 @@ func fetchGroupInfo(
...
-			for range results {
+	for waiting > 0 {
+		ret := <-results
+		waiting--
+		if ret.err != nil {
+			if err == nil {
+				err = ret.err
 			}
-			return err
+			continue
```
urso: we can stop processing data after having seen an error by doing this:

```
for waiting > 0 {
  ret := <-results
  waiting--
  if ret.err != nil && err == nil {
    err = ret.err
  }
  if err != nil {
    continue
  }
}
```

This will keep the waiting loop alive, assign the first `ret.err` only, but also stop processing any further results (results are discarded anyways).

urso:
> And For the issue: some consumergroup state being held by another kafka broker. I do not think it is an issue. Because the metricbeat will collect consumergroup info only from configed kafka host. If someone want collect all info, he need to config all kafka host in config file.

This might still be a problem. The consumergroup needs to correlate data, which are potentially distributed between different kafka nodes. Problem is, the metricsets in metricbeat are fully isolated. That is, even if you configure all brokers, the metricsets are running in isolation per broker -> can not correlated meta-data distributed on different brokers.

See related ticket: #4285 
The discuss forums discussion linked in the ticket contains quite some more details.

metricbeat/module/kafka/consumergroup/query.go
wangdisdu: @urso I fix it.

wangdisdu:
@urso Yes, you are right. It might be a problem. I'll puzzle over the problem, and create a new PR base on this PR in the future if it is really a problem.

wangdisdu:
@urso Hi, I found out that it is not a problem. 
> The offsets for a given consumer group are maintained by a specific broker called the group coordinator.  Link [kafka protocol guide](https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-GroupCoordinatorRequest)

We get the group list from a specific coordinator broker first, and can fetch all offsets info from it.
And I test and verify it on kafka 0.10.2.

urso:
@wangdisdu Cool. how many kafka brokers have you used? Did you use one central metricbeat, or one per host? Normally we recommend installing metricbeat on each host and configure metricbeat kafka module to connect to `localhost` broker only (metricbeat matches the meta-data with the hostname to figure out the actual brokers name from the metadata response). With having one metricbeat instance per host, we can also collect CPU/memory/disk usage on this host as well. Even though state is distributed between kafka nodes, each metricbeat will collect a subset of the overall kafka cluster state.

PR look good to me. Can you add a changelog about fixing the OffsetFetch request?

wangdisdu:
@urso For test kafka metricset, I build a kafka cluster with 3 brokers. 
For produce test data, I use standard java kafka client to produce and consume message.
I test metricbeat on per broker by configing local host ip (same as `localhost`).

wangdisdu:
@ruflin Will this PR be merged? So I can do more thing base on it, like batching requests together for efficiency.

urso:
@wangdisdu how many partitions and consumer groups did you use. The ticket I mentioned uses 10 partitions and 3 consumer groups, as sometimes for less consumer groups we could see all data. Only adding more groups/partitions to the picture to show us some missing consumer groups data.

urso:
Merged the PR. Thank  you for taking the time to work on this long outstanding issue. Your help is highly appreciated ;)

wangdisdu:
@urso It working well even if add more groups. I test and verify it:

-  I create topic `wd-test-mb-cg` with config 10 partitions and 2 peplication factors
-  I use 3 consumer groups to fetch message from  topic `wd-test-mb-cg`
-  I checked the data info is not missing
-  I increase count of consumer groups to 25 to fetch message from  topic `wd-test-mb-cg`
-  I checked the data info is not missing

ruflin:
@wangdisdu I can only double down on what @urso said. We really appreciate your work here and being patient with us to get the changes in. ❤️

npredey:
Which version of metricbeat is this available on and will it work with Elasticsearch 5.3? Thanks.

ruflin:
This is part of the 6.3 release.

Issue #5342, opened 2017-10-06
Add exclude_files option to file integrity module


andrewkroh:
An `exclude_files` configuration option should be added to the file integrity module in Auditbeat to ignore files that match any of the regular expressions specified in the list.

For example:

```
auditbeat.modules:
- module: audit
  metricsets: [file]
  file.paths:
  - /bin
  - /usr/bin
  - /usr/local/bin
  file.exclude_files: ['\.DS_Store$', '\.swp$']
```

Naming wise I'm debating between `file.exclude` and `file.exclude_files`. The latter is consistent with [Filebeat](https://www.elastic.co/guide/en/beats/filebeat/current/configuration-filebeat-options.html#exclude-files).

liketic:
Seem like `fsnotify` does not expose an exclude API. We can only do filter after received events for now.

andrewkroh:
Right, for fsnotify we can filter the incoming events based on their `Name`. For scanning it will be similar. During the `filepath.Walk` we can skip anything that matches one of the regexes.

But the scanning feature (#5344) hasn't merged yet so I would hold off a second to avoid any merge conflicts (I did a lot of refactoring).

adriansr:
I'm wondering about case-sensitivity for this regular expressions.
Can't decide what makes more sense. Ask the user to include (?i) at the start of every expression if case-insensitive matching is desired, or force case-insensitive matching in our side?

andrewkroh:
For consistency with other parts of Beats I think the user should have to add `(?i)`. AFAIK this is how other areas work. [Looks](https://github.com/exekias/beats/blob/edde8912793f3055f08937fb88fe432e47a8baaa/filebeat/harvester/reader/multiline_config.go#L16) like we use `match.Matcher` in the config and go-ucfg will validate the expression.

andrewkroh:
Implemented in #5888

PR #5888, opened 2017-12-15
Auditbeat: Add exclude_files option to file integrity module

adriansr:
Added a new config field, `exclude_files` consisting of regular expressions used to exclude files from monitoring by the file integrity module.

See #5342

### DIFF
--- a/auditbeat/docs/modules/file_integrity.asciidoc
+++ b/auditbeat/docs/modules/file_integrity.asciidoc
@@ -54,6 +54,10 @@ Linux.
   - /sbin
   - /usr/sbin
   - /etc
+  exclude_files:
+  - '(?i)\.sw[nop]$'
+  - '~$'
+  - '/\.git($|/)'
   scan_at_start: true
   scan_rate_per_sec: 50 MiB
   max_file_size: 100 MiB
@@ -63,6 +67,12 @@ Linux.
 
 *`paths`*:: A list of paths (directories or files) to watch. Globs are
 not supported. The specified paths should exist when the metricset is started.
+
+*`exclude_files`:: A list of regular expressions used to filter out events
+for unwanted files. By default, no files are excluded. See <<regexp-support>>
+for a list of supported regexp patterns. It is recommended to wrap regular
+expressions in single quotation marks to avoid issues with YAML escaping
+rules.
 
 *`scan_at_start`*:: A boolean value that controls if {beatname_uc} scans
 over the configured file paths at startup and send events for the files
@@ -94,8 +104,9 @@ value is sha1.
 
 *`recursive`*:: By default, the watches set to the paths specified in
 `paths` are not recursive. This means that only changes to the contents
-of this directories are watched. If `recursive` is set to `true`, the file
-metric will watch for changes on this directories and all their subdirectories.
+of this directories are watched. If `recursive` is set to `true`, the
+`file_integrity` module will watch for changes on this directories and all
+their subdirectories.
 
 
 [float]
@@ -115,6 +126,6 @@ auditbeat.modules:
   - /sbin
   - /usr/sbin
   - /etc
-  
+
 ----
 
--- a/auditbeat/docs/modules/file_integrity.asciidoc
+++ b/auditbeat/docs/modules/file_integrity.asciidoc
@@ -68,8 +68,9 @@ Linux.
 *`paths`*:: A list of paths (directories or files) to watch. Globs are
 not supported. The specified paths should exist when the metricset is started.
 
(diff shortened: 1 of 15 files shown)


adriansr:
Adding a unit test

auditbeat/module/file_integrity/config.go
```diff
@@ -40,14 +42,16 @@ const (
...
+	Paths               []string        `config:"paths" validate:"required"`
+	HashTypes           []HashType      `config:"hash_types"`
+	MaxFileSize         string          `config:"max_file_size"`
+	MaxFileSizeBytes    uint64          `config:",ignore"`
+	ScanAtStart         bool            `config:"scan_at_start"`
+	ScanRatePerSec      string          `config:"scan_rate_per_sec"`
+	ScanRateBytesPerSec uint64          `config:",ignore"`
+	Recursive           bool            `config:"recursive"` // Recursive enables recursive monitoring of directories.
+	Exclude             []string        `config:"exclude_files"`
+	ExcludeMatchers     []match.Matcher `config:",ignore"`
```
andrewkroh: `match.Matcher` implements `ucfg.StringUpacker` so it will automatically be populated with the compiled regexes. If any of them fail to compile ucfg should emit an error. So I don't think you need the `[]string` version at all.

auditbeat/module/file_integrity/metricset.go
```diff
@@ -167,6 +167,13 @@ func (ms *MetricSet) init(reporter mb.PushReporterV2) bool {
 }
 
 func (ms *MetricSet) reportEvent(reporter mb.PushReporterV2, event *Event) bool {
+	// ignore excluded paths
+	for _, matcher := range ms.config.ExcludeMatchers {
```
andrewkroh: While it is most convenient to apply the filtering here, this wouldn't be much different that using a drop_event processor with a regexp condition. In my mind the benefit of having a dedicated `exclude_files` option is that the filter can be applied earlier in order to save the CPU and I/O cycles associated with reading the file and hashing it. Would it be feasible to apply this filter earlier?

It would need to be applied to the scanner and fsnotify/fsevents readers. I think the events that are triggered by `purgeDeleted` would need filtered as well.

auditbeat/module/file_integrity/metricset_test.go
```diff
@@ -105,6 +105,68 @@ func TestDetectDeletedFiles(t *testing.T) {
...
+
+	ms := mbtest.NewPushMetricSetV2(t, getConfig(dir))
+	capture := &mbtest.CapturingReporterV2{DoneC: make(chan struct{})}
+
+	go func() {
+		for _, f := range []string{"FILE.TXT", "FILE.TXT.SWP", "file.txt.swo", ".git/HEAD", ".gitignore"} {
+			file := filepath.Join(dir, f)
+			ioutil.WriteFile(file, []byte("hello world"), 0600)
+		}
+		for i := 0; i < 100 && len(capture.Events) < 3; i++ {
```
andrewkroh: Waiting for a certain number of events is a pretty common task in tests. We should make this easier to do with the `mb/testing` package.

metricbeat/mb/testing/modules.go
```diff
@@ -217,29 +217,34 @@ func NewPushMetricSetV2(t testing.TB, config interface{}) mb.PushMetricSetV2 {
...
-func (r *capturingReporterV2) Done() <-chan struct{} {
-	return r.done
+// Done returns the Done channel for this reporter.
+func (r *CapturingReporterV2) Done() <-chan struct{} {
+	return r.DoneC
 }
 
 // RunPushMetricSetV2 run the given push metricset for the specific amount of
 // time and returns all of the events and errors that occur during that period.
 func RunPushMetricSetV2(duration time.Duration, metricSet mb.PushMetricSetV2) []mb.Event {
```
andrewkroh: I think it would make sense to add another parameter (`maxEvents`) to RunPushMetricSetV2 for specifying the maximum number of events to wait for. So then the behavior would be run until you reach `duration` (probably should rename to timeout) or until the maximum number of events is received (if `maxEvents > 0`).

auditbeat/docs/modules/file_integrity.asciidoc
```diff
@@ -64,6 +68,12 @@ Linux.
 *`paths`*:: A list of paths (directories or files) to watch. Globs are
 not supported. The specified paths should exist when the metricset is started.
 
+*`exclude_files`:: A list of regular expressions used to filter out events
```
andrewkroh: Based on questions I've seen for Filebeat, I think it would be good to explicitly state that this regular expression is matched against the file's absolute path.

PR #5891, opened 2017-12-15
support multiple passwords for redis output

aboullaite:
This PR add ability to specify multiple passwords for redis. Users now should add a list of passwords instead of one.
- if no password is specified then we suppose a no auth
- if one password is specified we consider it as a default password
- we map each password with host in case of multiple passwords. if the number of hosts and password doesn't match, we fail

### DIFF
--- a/libbeat/outputs/hosts.go
+++ b/libbeat/outputs/hosts.go
@@ -33,3 +33,17 @@ func ReadHostList(cfg *common.Config) ([]string, error) {
 
 	return hosts, nil
 }
+
+// OriginalHostList returns the list  of hosts to connect to as specified by
+func OriginalHostList(cfg *common.Config) ([]string, error) {
+	config := struct {
+		Hosts []string `config:"hosts"  validate:"required"`
+	}{}
+
+	err := cfg.Unpack(&config)
+	if err != nil {
+		return nil, err
+	}
+
+	return config.Hosts, nil
+}
--- a/libbeat/outputs/redis/config.go
+++ b/libbeat/outputs/redis/config.go
@@ -12,7 +12,7 @@ import (
 )
 
 type redisConfig struct {
-	Password    string                `config:"password"`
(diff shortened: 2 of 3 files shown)


ruflin:
@aboullaite Thanks for your contribution, a few thoughts.

* If we allow to set multiple passwords, we should make this change for all outputs
* I would prefer if we find an other way to do this like for example supporting to set the password as part of the host part as we do with elasticsearch. 
* If we make such a change, we need tests.

The part I worry with the above that there will be too many variants on how users want to use it. Could we provide the password as part of the hosts `hosts: ["localhost:6379"]` array?

aboullaite:
Hi @ruflin,

Thanks for your answer. I agree that this change should affect for all outputs, but I don't think adding it to the host is the best option, since will have another separator on the hosts list. That would be too much inputs for one var.

andrewkroh:
> Could we provide the password as part of the hosts `hosts: ["localhost:6379"]` array?

This is how I would expect it to work. This is how the ES output works. And I believe this is how most Metricbeat modules handle this (but not all).

So for a config like

```
hosts:
- user:pass@remote:6379
- localhost:6379
username: default
password: secret
```

it would use `user:pass` for connecting to `remote:6379` and would fallback to `default:secret` for `localhost:6379` because it did not contain a credentials.

aboullaite:
After a deep thinking, I understand your point of you and I agreed on! and the format you @andrewkroh suggested seems to be the best one!

ruflin:
+1 on @andrewkroh proposal. @aboullaite Want to modify your PR?

aboullaite:
@ruflin Yes, I'm actually working!

einali:
hi everybody 
is there any update about this improvement?
i need it !!

urso:
Hi, I'm closing this PR, because the PR has been stalled for quite some time and the proposed changes haven't been implemented yet. I created the follow up issue #16058

PR #5901, opened 2017-12-18
Refactor logp and add structured logging support

andrewkroh:
This refactors logp and adds support for structured logging. logp uses github.com/uber-go/zap in its implementation. There are no changes to the user facing logging configuration. The logger output will have some format differences, but in general will have a more consistent format across outputs.

Here is some sample output taken from the `TestLogger` test case.

```
=== RUN   TestLogger
2017-12-17T19:48:16.374-0500    INFO    logp/core_test.go:13    unnamed global logger
2017-12-17T19:48:16.374-0500    INFO    [example]       logp/core_test.go:16    some message
2017-12-17T19:48:16.374-0500    INFO    [example]       logp/core_test.go:17    some message with parameter x=1, y=2
2017-12-17T19:48:16.374-0500    INFO    [example]       logp/core_test.go:18    some message    {"x": 1, "y": 2}
2017-12-17T19:48:16.374-0500    INFO    [example]       logp/core_test.go:19    some message    {"x": 1}
2017-12-17T19:48:16.374-0500    INFO    [example]       logp/core_test.go:20    some message with namespaced args       {"metrics": {"x": 1, "y": 1}}
2017-12-17T19:48:16.374-0500    INFO    [example]       logp/core_test.go:21            {"empty_message": true}
2017-12-17T19:48:16.374-0500    WARN    [example]       logp/core_test.go:24    logger with context     {"x": 1, "y": 2}
2017-12-17T19:48:16.374-0500    INFO    [example]       logp/core_test.go:30    some message with struct value  {"metrics": {"x":1,"y":2}}
{"level":"info","timestamp":"2017-12-17T19:48:16.374-0500","caller":"logp/core_test.go:13","message":"unnamed global logger"}
{"level":"info","timestamp":"2017-12-17T19:48:16.374-0500","logger":"example","caller":"logp/core_test.go:16","message":"some message"}
{"level":"info","timestamp":"2017-12-17T19:48:16.374-0500","logger":"example","caller":"logp/core_test.go:17","message":"some message with parameter x=1, y=2"}
{"level":"info","timestamp":"2017-12-17T19:48:16.374-0500","logger":"example","caller":"logp/core_test.go:18","message":"some message","x":1,"y":2}
{"level":"info","timestamp":"2017-12-17T19:48:16.374-0500","logger":"example","caller":"logp/core_test.go:19","message":"some message","x":1}
{"level":"info","timestamp":"2017-12-17T19:48:16.374-0500","logger":"example","caller":"logp/core_test.go:20","message":"some message with namespaced args","metrics":{"x":1,"y":1}}
{"level":"info","timestamp":"2017-12-17T19:48:16.374-0500","logger":"example","caller":"logp/core_test.go:21","message":"","empty_message":true}
{"level":"warn","timestamp":"2017-12-17T19:48:16.374-0500","logger":"example","caller":"logp/core_test.go:24","message":"logger with context","x":1,"y":2}
{"level":"info","timestamp":"2017-12-17T19:48:16.374-0500","logger":"example","caller":"logp/core_test.go:30","message":"some message with struct value","metrics":{"x":1,"y":2}}
```

Any test code that calls `logp.LogInit()` in libbeat needs to be updated to use `logp.TestingSetup()`.


filebeat/fileset/fileset_test.go
```diff
@@ -204,9 +204,7 @@ func TestGetPipelineNginx(t *testing.T) {
 }
 
 func TestGetPipelineConvertTS(t *testing.T) {
-	if testing.Verbose() {
-		logp.LogInit(logp.LOG_DEBUG, "", false, true, []string{"fileset", "modules"})
-	}
+	logp.TestingSetup(logp.WithSelectors("fileset", "modules"))
```
tsg: Nice!

libbeat/logp/logger.go
```diff
@@ -1,31 +1,147 @@
...
-import "fmt"
+import (
+	"go.uber.org/zap"
+)
 
-// Logger provides a logging type using the global logp functionality.
-// The Logger should be used to use with libraries havng a configurable logging
-// functionality.
+// LogOption configures a Logger.
+type LogOption = zap.Option
```
tsg: I think this only works in Go 1.9, right? I think that's fine, just flagging in case this blocks some use case (like gccgo).

ph: Yes, type aliasing is only 1.9.

Should be OK in GCC8, not sure when is the release date.

__Due to the alignment of Go's semiannual release schedule with GCC's annual release schedule, GCC release 7 contains the Go 1.8.3 version of gccgo. We expect that the next release, GCC 8, will contain the Go 1.10 version of gccgo.__

tsg: ++, I think we're good to make use of it, as go 1.10 is in beta now so I imagine it's not that long until there's a GCC version that can work with it.

ruflin:
@elastic/apm This will also affect logging of the apm-server.

libbeat/logp/core_test.go
```diff
@@ -0,0 +1,188 @@
...
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"go.uber.org/zap"
+)
+
+func TestLogger(t *testing.T) {
```
ruflin: Thanks for adding tests 🎉

libbeat/logp/encoding.go
```diff
@@ -0,0 +1,66 @@
...
+		LineEnding:     zapcore.DefaultLineEnding,
+		EncodeLevel:    zapcore.LowercaseLevelEncoder,
+		EncodeTime:     zapcore.ISO8601TimeEncoder,
+		EncodeDuration: millisecondsDurationEncoder,
+		EncodeCaller:   zapcore.ShortCallerEncoder,
+		EncodeName:     zapcore.FullNameEncoder,
+	}
+}
+
+func consoleEncoderConfig() zapcore.EncoderConfig {
```
ruflin: I wonder if we should have a base config like the `consoleEncoderConfig` and to the changes on top of it the same way you did for the syslogEncoder but also the jsonEncoder? It seems the jsonEncoder has also only 2 differences? Would make it easier to figure out what we do as a "base level" and which changes are specific to a logger.

libbeat/logp/level.go
```diff
@@ -0,0 +1,74 @@
...
+// Level is a logging priority. Higher levels are more important.
+type Level int8
+
+// Logging levels.
+const (
+	DebugLevel Level = iota - 1
+	InfoLevel
+	WarnLevel
+	ErrorLevel
+	CriticalLevel // Critical exists only for config backward compatibility.
```
ruflin: Do we use `Critical` somewhere in our beats code?

vendor/vendor.json
```diff
@@ -1098,6 +1098,74 @@
...
 		},
+		{
+			"checksumSHA1": "HedK9m8E8iyib4bIBtIX7xprOgo=",
+			"path": "go.uber.org/atomic",
+			"revision": "8474b86a5a6f79c443ce4b2992817ff32cf208b8",
+			"revisionTime": "2017-11-14T20:44:01Z"
+		},
+		{
+			"checksumSHA1": "JDGx7hehaQunZySwPs7yvdUs2m8=",
+			"path": "go.uber.org/multierr",
```
ruflin: Seems like we have multiple multierr packages? Probably no way around this.

libbeat/docs/loggingconfig.asciidoc
```diff
@@ -15,20 +15,19 @@
 
 The `logging` section of the +{beatname_lc}.yml+ config file contains options
 for configuring the Beats logging output. The logging system can write logs to
-the syslog or rotate log files. If logging is not explicitly configured, file
-output is used on Windows systems, and syslog output is used on Linux and OS X.
+the syslog or rotate log files. If logging is not explicitly configured the file
```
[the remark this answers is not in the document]
andrewkroh: It didn't make this change. I just fixed the docs to match the behavior 😄 . This was changed in 6.0 I believe.

libbeat/docs/loggingconfig.asciidoc
```diff
@@ -158,14 +157,23 @@ When true, logs messages in JSON format. The default is false.
...
 === Logging format
 
-The logging format is different for each logging type:
+The logging format is generally the same for each logging output. The one
+exception is with the syslog output where the timestamp is not included in the
+message because syslog adds its own timestamp.
+
+Each log message consists of the following parts:
+
+* Timestamp in ISO8601 format
```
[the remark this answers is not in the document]
andrewkroh: They are almost the same, but RFC3339 (`RFC3339     = "2006-01-02T15:04:05Z07:00"`) does not have sub-second precision so it's not ideal for a logger timestamp IMHO.

libbeat/logp/encoding.go
andrewkroh: I agree and noticed this myself that there was some duplication. I can clean this up in another PR.

libbeat/logp/level.go
andrewkroh: We previously supported `critical` as a config option so I had to carry it forward (`logging.level: critical`). I removed it from the docs.

vendor/vendor.json
andrewkroh: Yeah, no way other than modifying the third-party software.

libbeat/docs/loggingconfig.asciidoc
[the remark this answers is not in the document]
ruflin: Ups :-) We should probably backport this part of the doc changes to 6.1/6.0 then. @dedemorton FYI

andrewkroh:
> The part I'm more worried about is changing the default "log output". Does this change in this PR?

No, we made the change to output to file by default a while ago. See https://github.com/elastic/beats/pull/1437

> For example if someone uses filebeat to fetch logs for metricbeat and has a filter for ERR things would not work anymore.

This is a valid concern (though hopefully ERR will still match ERROR). But I really hope the format change doesn't cause too much trouble. 🤞

roncohen:
Fantastic to have structured logging in logp!

AFAIK, Kibana will use an easily readable logging format depending on whether there's a TTY or not. 

I haven't had time to look into the implementation here, but it seems like it would be nice to output a human readable format when starting with `-e`, but allowing it to be overridden in case STDOUT is being piped through logstash for parsing or similar.

roncohen:
@andrewkroh is this going in 6.2? We'd like to use it

andrewkroh:
> it would be nice to output a human readable format when starting with -e, but allowing it to be overridden in case STDOUT is being piped

Interesting idea. That's possibly something we could add for 7.0 (since it's kind of breaking).

> is this going in 6.2? We'd like to use it

I believe so. I think it's just waiting on the next mass merge from master to the 6.x branch.

PR #5910, opened 2017-12-18
Allow TEST_ENVIRONMENT to be externally disabled

andrewkroh:
Builds that set TEST_ENVIRONMENT=0 should be able to override this
value in order to avoid Docker usage.


ruflin:
@andrewkroh Can you have a look at the auditbeat failure? It might be related.

andrewkroh:
It's not related to this change, but have asked @adriansr to investigate.

PR #5912, opened 2017-12-19
Use shared mode for opening files on Windows

adriansr:
This PR updates a couple of places to use full shared mode when opening files on Windows, allowing other processes to access the files at the same time:
- When hashing the files
- When looking up the file owner

To facilitate reuse, this functionality which already existed in filebeat has been moved to libbeat


ruflin:
Thanks for moving this to libbeat.

PR #5913, opened 2017-12-19
Add log output to Windows Event Log

andrewkroh:
By setting `logging.to_eventlog: true` all log output will be written
to the Application log. The source name will be name of the Beat.

### DIFF
--- a/auditbeat/auditbeat.reference.yml
+++ b/auditbeat/auditbeat.reference.yml
@@ -831,12 +831,11 @@ setup.kibana:
 
 
 #================================ Logging ======================================
-# There are three options for the log output: syslog, file, stderr.
-# Under Windows systems, the log files are per default sent to the file output,
-# under all other system per default to syslog.
+# There are four options for the log output: file, stderr, syslog, eventlog
+# The file output is the default.
 
 # Sets log level. The default log level is info.
-# Available log levels are: critical, error, warning, info, debug
+# Available log levels are: error, warning, info, debug
 #logging.level: info
 
 # Enable debug output for selected components. To enable all selectors use ["*"]
@@ -845,7 +844,10 @@ setup.kibana:
 #logging.selectors: [ ]
 
 # Send all logging output to syslog. The default is false.
-#logging.to_syslog: true
+#logging.to_syslog: false
+
+# Send all logging output to Windows Event Logs. The default is false.
+#logging.to_eventlog: false
 
 # If enabled, auditbeat periodically logs its internal metrics that have changed
 # in the last period. For each metric that changed, the delta from the value at
--- a/libbeat/logp/encoding.go
+++ b/libbeat/logp/encoding.go
@@ -6,6 +6,21 @@ import (
 	"go.uber.org/zap/zapcore"
 )
 
+var baseEncodingConfig = zapcore.EncoderConfig{
+	TimeKey:        "timestamp",
+	LevelKey:       "level",
+	NameKey:        "logger",
+	CallerKey:      "caller",
+	MessageKey:     "message",
+	StacktraceKey:  "stacktrace",
+	LineEnding:     zapcore.DefaultLineEnding,
+	EncodeLevel:    zapcore.LowercaseLevelEncoder,
+	EncodeTime:     zapcore.ISO8601TimeEncoder,
+	EncodeDuration: millisecondsDurationEncoder,
+	EncodeCaller:   zapcore.ShortCallerEncoder,
+	EncodeName:     zapcore.FullNameEncoder,
+}
+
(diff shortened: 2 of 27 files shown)


andrewkroh:
I rebased this to fix the CHANGELOG merge conflict.

auditbeat/auditbeat.reference.yml
```diff
@@ -845,7 +844,10 @@ setup.kibana:
 #logging.selectors: [ ]
 
 # Send all logging output to syslog. The default is false.
-#logging.to_syslog: true
+#logging.to_syslog: false
+
+# Send all logging output to Windows Event Logs. The default is false.
+#logging.to_eventlog: false
```
ruflin: Not a big fan of naming we have for historical reason and I'm thinking we should change it for 7.0. Perhaps we could already follow here a new scheme similar to what we have in other places:

`logging.eventlog.enabled`. Like this we can have specific logging options under the namespace.

libbeat/logp/encoding.go
```diff
@@ -6,6 +6,21 @@ import (
 	"go.uber.org/zap/zapcore"
 )
 
+var baseEncodingConfig = zapcore.EncoderConfig{
```
ruflin: +1 on the cleanup

libbeat/logp/eventlog_windows.go
```diff
@@ -0,0 +1,91 @@
...
+import (
+	"strings"
+
+	"github.com/pkg/errors"
+	"go.uber.org/zap/zapcore"
+	"golang.org/x/sys/windows/svc/eventlog"
+)
+
+const (
+	eventID  = 100
```
ruflin: Does `100` have a special meaning in windows event logs?

libbeat/logp/eventlog_windows.go
```diff
@@ -0,0 +1,91 @@
...
+	supports = eventlog.Warning | eventlog.Info | eventlog.Info
```
ruflin: Why does `eventlog.Info` show up twice? Intentional or should it be Debug?

libbeat/logp/syslog_unix.go
```diff
@@ -53,7 +53,7 @@ func (c *syslogCore) Write(entry zapcore.Entry, fields []zapcore.Field) error {
 	}
 
 	// Console encoder writes tabs which don't render nicely with syslog.
-	replaceTabsWithSpaces(buffer.Bytes(), 2)
+	replaceTabsWithSpaces(buffer.Bytes(), 4)
```
ruflin: Is that change related to this PR?

auditbeat/auditbeat.reference.yml
andrewkroh: I am 100% on board with changing how logging is configured for 7.0. I did it this way to be consistent. I will open an issue to track things that I'd like to "break" for 7.0 (I have a list 😄 ).

libbeat/logp/eventlog_windows.go
andrewkroh: No, it's completely arbitrary. The value needs to be between 1 and 1000. Most of the original log events have three digit IDs so I just chose 100.

libbeat/logp/eventlog_windows.go
andrewkroh: That should be `Error`. Will push a fix.

libbeat/logp/syslog_unix.go
andrewkroh: It's not related. I just realized that there can be up to 4 tabs.

auditbeat/auditbeat.reference.yml
ruflin: It should be possible to already do the "new" scheme here as so far it's just a name. More thinking then we have to break one option less for 7 :-) But not sure if we are already settle on if that should be new scheme.

+1 on the issue.

andrewkroh: Well the scheme I had in mind kind of conflicts. I was thinking of something like.

```
logging.output:
- type: file
   level: info
   json: true
- type: eventlog
  level: warn
```

ruflin: got it, then lets keep it as is in your PR and break it later so we don't need to come to a conclusion now.

I like your format as it gives max flexibility. I assume you have in mind that more then 1 log output can be enabled?

andrewkroh: Yeah, more than one output and possibly more than one instance of an output type (not 100% sure yet; requires more thought).

andrewkroh:
I just did a manual squash to keep the vendor changes separate from the code. Now you can do a rebase and merge. Thanks

PR #5914, opened 2017-12-19
Fix race-condition on RunPushMetricSetV2

adriansr:
This patch fixes a race condition in the metricset test helper RunPushMetricSetV2.

A previous patch already fixed some issues but there was another race-condition left where a returned slice of bytes could be appended to by a different thread.

### DIFF
--- a/metricbeat/mb/testing/modules.go
+++ b/metricbeat/mb/testing/modules.go
@@ -38,7 +38,6 @@ package testing
 
 import (
 	"sync"
-	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


metricbeat/mb/testing/modules.go
```diff
@@ -221,38 +220,51 @@ func NewPushMetricSetV2(t testing.TB, config interface{}) mb.PushMetricSetV2 {
...
+	doneC  chan struct{}
+	mutex  sync.Mutex
 }
 
 // Event stores the passed-in event into the events array
 func (r *capturingReporterV2) Event(event mb.Event) bool {
-	r.Events = append(r.Events, event)
-	atomic.AddInt32(&r.count, 1)
+	r.mutex.Lock()
+	r.events = append(r.events, event)
```
ruflin: Should we use `defer r.mutex.Unlock()` here to have the unlock close to the `mutex`? We have similar pattern in most of the code.

metricbeat/mb/testing/modules.go
```diff
@@ -221,38 +220,51 @@ func NewPushMetricSetV2(t testing.TB, config interface{}) mb.PushMetricSetV2 {
...
+	r.mutex.Unlock()
 	return true
 }
 
 // Error stores the given error into the errors array.
 func (r *capturingReporterV2) Error(err error) bool {
-	r.Events = append(r.Events, mb.Event{Error: err})
+	r.mutex.Lock()
+	r.events = append(r.events, mb.Event{Error: err})
```
ruflin: See comment above

metricbeat/mb/testing/modules.go
```diff
@@ -221,38 +220,51 @@ func NewPushMetricSetV2(t testing.TB, config interface{}) mb.PushMetricSetV2 {
...
 // Done returns the Done channel for this reporter.
 func (r *capturingReporterV2) Done() <-chan struct{} {
-	return r.DoneC
+	return r.doneC
 }
 
 func (r *capturingReporterV2) eventCount() int {
-	return int(atomic.LoadInt32(&r.count))
+	r.mutex.Lock()
+	defer r.mutex.Unlock()
```
ruflin: here we go :-)

metricbeat/mb/testing/modules.go
```diff
@@ -262,17 +274,16 @@ func RunPushMetricSetV2(timeout time.Duration, waitEvents int, metricSet mb.Push
...
-			return r.Events
+			done = true
 		case <-timeoutC:
-			return r.Events
+			done = true
 		default:
-			if waitEvents != 0 && waitEvents <= r.eventCount() {
-				return r.Events
-			}
 			time.Sleep(100 * time.Millisecond)
```
ruflin: @andrewkroh What was the reason we wait here 100ms? Perhaps worth a comment (not related to this PR).

metricbeat/mb/testing/modules.go
adriansr: I was a little inconsistent with the mutex unlock, doing it manually in some places and using defer when the return statement required the mutex to be kept in locked state. I was trying to avoid the `defer` overhead but that is stupid as this is just a testing helper.

metricbeat/mb/testing/modules.go
adriansr: I added this sleep in a previous commit, to avoid constant checking of the exit condition until the `termC` or `timeoutC` channels fire. Again, this is just a testing helper, but I can add an explanatory comment if you think it's necessary.

metricbeat/mb/testing/modules.go
```diff
@@ -262,17 +274,16 @@ func RunPushMetricSetV2(timeout time.Duration, waitEvents int, metricSet mb.Push
...
+			done = waitEvents != 0 && waitEvents <= r.eventCount()
 		}
 	}
+	return r.Events()
 }
```
andrewkroh: May I suggest an alternative approach that uses channels instead of locks. How does this look:

```
// capturingReporterV2 stores all the events and errors from a metricset's
// Run method.
type capturingReporterV2 struct {
	done    chan struct{}
	eventsC chan mb.Event
}

// Event stores the passed-in event into the events array
func (r *capturingReporterV2) Event(event mb.Event) bool {
	r.eventsC <- event
	return true
}

// Error stores the given error into the errors array.
func (r *capturingReporterV2) Error(err error) bool {
	r.eventsC <- mb.Event{Error: err}
	return true
}

// Done returns the Done channel for this reporter.
func (r *capturingReporterV2) Done() <-chan struct{} {
	return r.done
}

// RunPushMetricSetV2 run the given push metricset for the specific amount of
// time and returns all of the events and errors that occur during that period.
func RunPushMetricSetV2(timeout time.Duration, waitEvents int, metricSet mb.PushMetricSetV2) []mb.Event {
	var (
		r      = &capturingReporterV2{done: make(chan struct{}), eventsC: make(chan mb.Event, 1)}
		wg     sync.WaitGroup
		events []mb.Event
	)

	// Producer
	wg.Add(1)
	go func() {
		defer wg.Done()
		defer close(r.eventsC)
		metricSet.Run(r)
	}()

	// Consumer
	wg.Add(1)
	go func() {
		defer wg.Done()
		defer close(r.done)

		timer := time.NewTimer(timeout)
		defer timer.Stop()
		for {
			select {
			case <-timer.C:
				return
			case e := <-r.eventsC:
				events = append(events, e)
				if len(events) >= waitEvents {
					return
				}
			}
		}
	}()

	wg.Wait()
	return events
}
```

adriansr: One thing I liked about the original code is that it allows the metricSet to generate more events than expected and this can be detected by tests. But only one test was relying on this

ruflin: @adriansr A bit late to the game, sorry. But can you elaborate on this how we changed the behaviour?

andrewkroh: I think what he was referring to is that in my version I forgot to allow for `waitEvents == 0` which indicates that the only bound on the running time is the timeout (a user doesn't have to specify the exact number of events to receive).

In the final version @adriansr changed the `if len(events) >= waitEvents {` to `if waitEvents > 0 && len(events) >= waitEvents {`.

PR #5919, opened 2017-12-19
metricbeat: add cluster_health metricset into elasticsearch module

wangdisdu:
`cluster_health` metricset include follow metrics:
- `cluster.status`
- `cluster.nodes.active.count`
- `cluster.shards.active.count`


ruflin:
@wangdisdu Give us some more time to discuss internally on how many metrics for Elasticsearch we want in Metricbeat and which one. Thanks for your contributions.

PR #5933, opened 2017-12-20
Auditbeat: Add setuid/setgid fields to file_integrity

### LINKED ISSUE #5527: Add setuid and setgid fields to file integrity module
It would be convenient for alerting and Kibana usage to send an explicit field for files that have their setuid or setgid bits set. I would only send the field when those bits are set. Something like `audit.file.setuid: true` and `audit.file.setgid: true`.

Since we only need this on POSIX platforms we can update the `fileinfo_posix.go` to set two new fields in the `Metadata` object.

https://github.com/elastic/beats/blob/1ec7437036852f6ab234c8b5d25ca8ce45bcb3f5/auditbeat/module/audit/file/fileinfo_posix.go#L24

adriansr:
Added two new fields: `setuid` and `setgid` which are only present
if the given bit is set in the file. The only possible value is `true`.
This fields are only used in POSIX platforms.

Closes #5527

### DIFF
--- a/auditbeat/module/file_integrity/_meta/fields.yml
+++ b/auditbeat/module/file_integrity/_meta/fields.yml
@@ -60,12 +60,12 @@
       - name: setuid
         type: boolean
         example: true
-        description: Set if the file has the `setuid` bit set. Missing otherwise.
+        description: Set if the file has the `setuid` bit set. Omitted otherwise.
 
       - name: setgid
         type: boolean
         example: true
-        description: Set if the file has the `setgid` bit set. Missing otherwise.
+        description: Set if the file has the `setgid` bit set. Omitted otherwise.
 
       - name: size
         type: long
--- a/auditbeat/module/file_integrity/_meta/fields.yml
+++ b/auditbeat/module/file_integrity/_meta/fields.yml
@@ -57,6 +57,16 @@
         example: 0640
         description: The mode of the file in octal representation.
 
+      - name: setuid
+        type: boolean
+        example: true
+        description: Set if the file has the `setuid` bit set. Missing otherwise.
+
+      - name: setgid
+        type: boolean
+        example: true
+        description: Set if the file has the `setgid` bit set. Missing otherwise.
+
       - name: size
         type: long
         description: The file size in bytes (field is only added when `type` is `file`).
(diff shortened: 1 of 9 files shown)


auditbeat/module/file_integrity/schema/Metadata.go
```diff
@@ -130,8 +130,32 @@ func (rcv *Metadata) MutateType(n byte) bool {
 	return rcv._tab.MutateByteSlot(20, n)
 }
 
+func (rcv *Metadata) Setuid() byte {
```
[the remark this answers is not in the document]
adriansr: shut up, dog. This is autogenerated

auditbeat/module/file_integrity/schema.fbs
```diff
@@ -31,6 +31,8 @@ table Metadata {
   mtime_ns:long;
   ctime_ns:long;
   type:Type = 1;
+  setuid:bool;
```
andrewkroh: Could we persist these bits with the `mode` to save a little space? I think currently we are at ~152 bytes per record.

auditbeat/module/file_integrity/_meta/fields.yml
```diff
@@ -57,6 +57,16 @@
         example: 0640
         description: The mode of the file in octal representation.
 
+      - name: setuid
+        type: boolean
+        example: true
+        description: Set if the file has the `setuid` bit set. Missing otherwise.
```
andrewkroh: `s/Missing/Omitted/g` would sound better IMO.

auditbeat/module/file_integrity/fileinfo_test.go
```diff
@@ -76,3 +77,59 @@ func TestNewMetadata(t *testing.T) {
...
+	meta, err := NewMetadata(f.Name(), info)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	assert.False(t, meta.SetUID)
+	assert.False(t, meta.SetGID)
+
+	if runtime.GOOS == "windows" {
+		t.Skip("Not setuid/setgid bits on Windows")
```
andrewkroh: `s/Not/No/` ?

adriansr: good catch

auditbeat/module/file_integrity/schema.fbs
adriansr: sure thing

PR #5936, opened 2017-12-20
Fix variable name for convert_timezone

tsg:


exekias:
You will need to run `make update`

PR #5938, opened 2017-12-20
Add dashboard test loading for auditbeat

ruflin:
In combination with adding the dashboard loading a libbeat.yml.j2 was created so all beats can share the common config options and we do not have to add each option for each beat.

So far this is applied to metricbeat, auditbeat and filebeat. Other beats will follow in a follow up PR.


ruflin:
@elastic/apm-server Could be useful for apm-server too.

auditbeat/tests/system/test_base.py
```diff
@@ -47,3 +47,25 @@ def test_template(self):
...
+        Test that the dashboards can be loaded with `setup --dashboards`
+        """
+
+        kibana_dir = os.path.join(self.beat_path, "_meta", "kibana")
+        shutil.copytree(kibana_dir, os.path.join(self.working_dir, "kibana"))
+
+        es = Elasticsearch([self.get_elasticsearch_url()])
+        self.render_config_template(
+            modules=[{
+                "name": "auditd",
```
andrewkroh: I don't think these test get executed because the config for the one above is out-of-date.

Let's use the `file_integrity` module instead of `auditd` because it can run on more platforms. 

```
             modules=[{
                 "name": "file_integrity",
                 "extras": {
                     "paths": ["file.example"],
                 },
```

libbeat/tests/system/config/libbeat.yml.j2
```diff
@@ -0,0 +1,98 @@
...
+  {%- if agent_tags -%}
+      {%- for tag in agent_tags -%}
+          "{{ tag }}"
+          {%- if not loop.last %}, {% endif -%}
+      {%- endfor -%}
+  {%- endif -%}
+]
+
+{% if geoip_paths is not none %}
+geoip:
```
andrewkroh: What's this for? AFAIK we don't have any builtin geoip enrichment.

ruflin: That was copy paste from filebeat. Removed now as we don't use it anymore.

auditbeat/tests/system/test_base.py
ruflin: Changed to your suggested config. I assume the first paragraph was related to the second part with the new config to enable it?

ruflin: @exekias In general I wonder if we should require a module to run the `setup` command?

auditbeat/tests/system/test_base.py
```diff
@@ -60,8 +60,10 @@ def test_dashboards(self):
         es = Elasticsearch([self.get_elasticsearch_url()])
         self.render_config_template(
             modules=[{
-                "name": "auditd",
-            }],
+                "name": "file_integrity",
```
andrewkroh: Can you fix the config for the one above too, please.

In auditbeat/Makefile I think `SYSTEM_TESTS` needs to be set to true for these to execute.

ruflin:
@andrewkroh I only realised now that the system tests we had were not running at all for auditbeat. So I enabled system tests + integration tests. I'm pushing a change now to see if everything works as expected but will squash and rebase again in case it goes green.

I also enabled the file_integrity module for all platforms, not only linux.

ruflin:
Opened https://github.com/elastic/beats/pull/5945 This needs rebase as soon as the other PR is merged.

libbeat/tests/system/config/libbeat.yml.j2
```diff
@@ -0,0 +1,93 @@
...
+  {%- if agent_tags -%}
+      {%- for tag in agent_tags -%}
+          "{{ tag }}"
+          {%- if not loop.last %}, {% endif -%}
+      {%- endfor -%}
+  {%- endif -%}
+]
+
+{% if setup_template_name %}
+setup.template.name: setup_template_name
```
andrewkroh: Should these be a reference to the `setup_template_name ` variable?

ruflin: I would say yes. It got copied from here: https://github.com/elastic/beats/pull/5938/files#diff-bfed2fbdbe1d791a97c6ca3a63e6a8deL152

Will fix it for both variables.

ruflin:
The failing test is a flaky test in k8s and should not be related.

@exekias ^

Issue #5632, opened 2017-11-17
Beats doc updates in 6.x


dedemorton:
FYI: I'm planning to move the roadmap for docs to a project instead of a meta issue because the meta issue format isn't working well. I end up doing a lot of work that never makes it onto this list.

**6.x**
- MOVED TO ROADMAP Document the fields.yml and backport to 6.0, if possible. See https://github.com/elastic/beats/issues/6049
- [x] [DeDe] Make sure GS guides accurately reflect requirement for using sudo (Most of the Beats require to sudo commands. Exception is Filebeat and Heartbeat (only if icmp is not configured)). Look at the soln docs as a source of truth for recommendations.
- MOVED TO ROADMAP Add topic [about exporting GeoIP info ](https://www.elastic.co/guide/en/beats/packetbeat/current/packetbeat-geoip.html) to the documentation for all the Beats
- MOVED TO ROADMAP Add more about machine learning jobs. The setup command is documented, but we probably need to add info about required privileges plus maybe have some conceptual info. _Update 8/14: We've documented the privileges required. Reached out to dev to see if we need to say more in the Beats docs._

**Old Defects (open too long)**

REMOVED THE LIST HERE BECAUSE IT'S DOUBLE-TRACKING

**Defects related to developer docs**

REMOVED THE LIST HERE BECAUSE IT'S DOUBLE-TRACKING

**Future:**

- MOVED TO ROADMAP Add docs that are more task-based and focused on solutions, such as "Monitoring your infrastructure"
- REMOVED, COVERED BY https://github.com/elastic/beats/issues/7320. Refactor files in libbeat so that they can be more easily used by projects like apm. This includes moving the files that are not part of the Platform Reference into a directory called `include` under `libbeat\docs.

**Backlog:**

REMOVED ANYTHING HERE THAT IS ALREADY BEING TRACKED IN AN ISSUE

- [ ] Improve docs about running Beats on containers. Highlight support for getting logs and metrics from apps running on containers and bring all the topics into a single section. 
- [ ] Edit this content as time allows:
  - [ ] Metricbeat docs about the Kafka module.
  - [ ] Options added to output config doc since last edit (make sure all additions in #2597 have been edited)
- [ ] Packetbeat docs: Update New Protocol guide to describe the generator for Packetbeat. Include info about protocol analyzer generator. See https://github.com/elastic/beats/issues/1111.  
- [ ] Character encoding of source files
- [ ] Add topic about using Beats with Kafka?
[...]

PR #5944, opened 2017-12-21
[WIP] Rename prospector to input

ph:
Rename usage of prospector to input inside the code base to align with the wording or other product like Logstash and also its better name for UDP and TCP.

This PR use _type aliasing_ to make the old types works with the new names, this mean it should be backward compatible with the community beats without requiring them to changes their code. They will need golang 1.9 to leverage that feature, we have guards in place to fails the build in other version.

Also the usage of `prospectors` in the configuration and in the module definition will work until 7.0.

## TODO
- [ ] Rename prospector usage in python scripts.
- [ ] Real world testing
- [ ] Create a PR for the documentation. related to https://github.com/elastic/beats/issues/5632
- [ ] squash better commit message
- [ ] split into multiple PR.


ph:
@ruflin After changing the name from prospector to input, I believe we should also change of harvester, It sound weird in the context of TCP/UDP and also redis. I don't have a proposal yet for that.


```
harvester | ˈhärvəstər |
noun
a person or machine that gathers crops as a harvest: back-breaking labor as hired sugar beet harvesters | the world's first mechanical harvester of grapes.
• a person who catches or kills animals for human consumption: coastal fish harvesters are always the first to notice the decline of fish stocks.
• a person or organization that collects or obtains a resource for future use: one of the world's biggest harvesters of the sun's energy.
```

ruflin:
Agree. I think `harvesters` together with `prospectors` made kind of sense but removing one the other one has to follow too. But I would do this in separate steps.

I had in the past discussions with @urso about this. There are at least 2 cases here. In same cases what we have as harvester is not needed anymore. For example for `stdin` there is only 1 prospector, 1 harvester and we  have a 1-1 relation ship. Potentially harvester can become obsolete here. For the log case we have 1-n. In the log harvester big parts are abtracted as `readers`. The readers can be reused across different prospector/input types like multiline etc. They are chained together. We discussed in the past to rename `harvester` to `reader` but I think the way we have reader at the moment is pretty nice and it would conflict here.

An option could be to call it `listeners`. It only fits partially with the file case. The reason I liked `harvester` is it describes filebeat going "actively" to the file and check for new data.

One additional note for the prospector / input. For the file case (and probably some others) there is a specific `scanner` phase which we discussed to extract in the past which is responsible to scan the file system. It potentially could be exchanged with a different scanner (inotify). Just to add some more names to the equation.

Let's keep it "simple" for now and do only the prospector -> input renaming and I'm pretty sure if we add more inputs some of the names become more obvious.

ph:
Agree lets do it in multiple steps, naming things are hard :)

urso:
about harvester: From my point of view the prospectors are the inputs. If an input/prospector can handle the input directly, it should not be required to create another go-routine doing the actual work. The harvesters to date are basically just the `worker`s of the log-input type -> therefore I'd just rename them to `worker` and even hide the fact we have workers from the user (it's an implementation detail, one could use select on FDs as well). If you really need a name `collector` or `reader` might do.

filebeat/_meta/common.reference.p2.yml
```diff
@@ -1,22 +1,22 @@
-#=========================== Filebeat prospectors =============================
+#=========================== Filebeat inputs =============================
 
-# List of prospectors to fetch data.
-filebeat.prospectors:
-# Each - is a prospector. Most options can be set at the prospector level, so
-# you can use different prospectors for various configurations.
-# Below are the prospector specific configurations.
+# List of input to fetch data.
```
dedemorton: I'd say inputs (plural)

filebeat/_meta/common.reference.p2.yml
```diff
@@ -1,22 +1,22 @@
...
 # Type of the files. Based on this the way the file is read is decided.
-# The different types cannot be mixed in one prospector
+# The different types cannot be mixed in one input
 #
 # Possible options are:
 # * log: Reads every line of the log file (default)
 # * stdin: Reads the standard in
 
-#------------------------------ Log prospector --------------------------------
+#------------------------------ Log input --------------------------------
```
dedemorton: Should we change this to say "Input" instead of "Log input" given the various types we now support?

filebeat/tests/system/test_registrar.py
```diff
@@ -980,13 +980,13 @@ def test_restart_state(self):
         # Make sure all 4 states are persisted
         self.wait_until(
             lambda: self.log_contains(
-                "Prospector states cleaned up. Before: 4, After: 4", logfile="filebeat2.log"),
+                "input states cleaned up. Before: 4, After: 4", logfile="filebeat2.log"),
```
dedemorton: Shouldn't this be capitalized?

filebeat/tests/system/test_registrar.py
```diff
@@ -980,13 +980,13 @@ def test_restart_state(self):
...
             max_timeout=10)
 
         # Wait until registry file is cleaned
         self.wait_until(
             lambda: self.log_contains(
-                "Prospector states cleaned up. Before: 0, After: 0", logfile="filebeat2.log"),
+                "input states cleaned up. Before: 0, After: 0", logfile="filebeat2.log"),
```
dedemorton: Same as previous comment

ph:
@dedemorton Thanks for the check :)

ruflin:
@ph Ping me when you think this needs a review.

ph:
closing opening followup PRs.

Issue #5932, opened 2017-12-20
add_kubernetes_metadata processors supports more k8s resources


walktall:
Now add_kubernetes_metadata only supports pods, it is fine for enriching logs with pod meta.

But when using metricbeats to persist k8s events, I also want to enrich the event item with involved object's label and annotations, so resources meta caches for other resources are needed too.

I think it can be achieved by adding some resource watchers in add_kubernetes_metadata processors.

Is there any plan about it?
If it is ok to add it, I can try to make a PR for it.

exekias:
Thank you for your feedback! This is something we have discussed before, it could make sense. Could explain a bit about what's missing for you? Some examples would help here

walktall:
I want something like this:

```
            "involved_object" : {
                "kind" : "Pod",
                "name" : "mongodb-app-v2-0-1543739302-kf642",
                "uid" : "23bdc75a-e55c-11e7-a29a-52540096096d",
                "namespace": "default",
                "api_version" : "v1",
                "resource_version" : "117819",
                "fieldPath" : "spec.containers{c0}",
                "labels" : {
                    "app" : "mongodb"
                }
            },
```

And then I can search all the events related to `"app" : "mongodb"`. Unlike the logs, event items in k8s are not only from pods, they could be triggered by other resources too... So if I want to get all events related to `"app" : "mongodb"`, add_kubernetes_metadata need to create caches for these resources like hpa, nodes, deployment so that involved_object can be enriched with k8s labels and 
annotations which then can be used for searching.

exekias:
This idea sounds good in general, could you give an example of how would you use this with other resources, including the expected result in the final document?

walktall:
@exekias I want to write DSL like this to get all my k8s resource has `helm_sh/release": "nginxdemo"` in annotations. I use this annotation as a filter for different application.

```
{
  "size": 0,
  "query": {
    "term": {
      "kubernetes.event.involved_object.annotations.helm_sh/release": "nginxdemo"
    }
  },
  "aggs": {
    "by_UID": {
      "terms": {
        "field": "kubernetes.event.metadata.uid",
        "size": 10,
        "order": {
          "show_recent_happened": "desc"
        }
      },
      "aggs": {
        "show_recent_happened": {
          "max": {
            "field": "kubernetes.event.timestamp.last_occurrence"
          }
        },
        "top_hit": {
          "top_hits": {
            "size": 1,
            "sort": [
              {
[...]

PR #5949, opened 2017-12-22
libbeat: add_kubernetes_metadata supports more k8s resources

walktall:
Ref: #5932 

This PR is kind of proposal about how to make add_kubernetes_metadata processor support other k8s resources than only pod.

And I also added a use case that in metricbeat, k8s events module can enrich involved_object with resource meta.

Currently, only podWatcher was added. It should be easy to add other resource watchers by implement watcher interface.


walktall:
Close since part of the feature is going to be implemented in https://github.com/elastic/beats/pull/5972.

exekias:
Thanks for opening this PR @walktall! It's a little unfortunate that the two of us were working on the same paths at the same time. In any case I would say this PR is still valid and makes a point, let's rework this on top of #5972 :+1: I hope you are still able to help here :smiley:

Issue #5659, opened 2017-11-21
add support for condition on bool type


TomonoriSoejima:
Currently

[Doc](https://www.elastic.co/guide/en/beats/packetbeat/current/defining-processors.html#condition-equals) does not support `bool` type.
```
With the equals condition, you can compare if a field has a certain value. The condition accepts only an integer or a string value.
```

When I did so, I got this warning packetbeat.

```
2017/11/21 01:10:50.019224 condition.go:235: WARN unexpected type bool in equals condition as it accepts only integers and strings. 
```

ewgRa:
This problem looks like duplicate of #4637, probably one of them can be closed in favor of another.

Merged PR #5954 looks like solution for both issues.

PR #5954, opened 2017-12-23
Add support for condition on bool type

ewgRa:
Pull request that add ability to use bool in processor conditions, related to #5659


libbeat/processors/condition.go
```diff
@@ -120,10 +121,17 @@ func (c *Condition) setEquals(cfg *ConditionFields) error {
 			c.equals[field] = EqualsValue{Int: uintValue}
```
ruflin: Perhaps we should use `continue` in the code to have less nested if / else? Same on line 125.

libbeat/processors/condition.go
```diff
@@ -259,12 +267,22 @@ func (c *Condition) checkEquals(event ValuesMap) bool {
...
 		} else {
 			sValue, err := extractString(value)
-			if err != nil {
-				logp.Warn("unexpected type %T in equals condition as it accepts only integers and strings. ", value)
-				return false
-			}
-			if sValue != equalValue.Str {
-				return false
+
+			if err == nil {
```
ruflin: Same comment as above, use continue to make the code more readable and less nested.

libbeat/processors/condition_test.go
```diff
@@ -93,6 +93,18 @@ func TestEqualsCondition(t *testing.T) {
 				"proc.cpu.total_p.gt": 0.5,
 			}},
 		},
+
+		{
+			Equals: &ConditionFields{fields: map[string]interface{}{
```
ruflin: I opened https://github.com/elastic/beats/pull/5960 to make adding tests cleaner. I realised when reading this that the way tests are written at the moment is not optimal for diffs.

ruflin:
jenkins, test it

ewgRa:
@ruflin done

About "continue", for me both ways looks not so good, since "continue" also like "goto" here, but it is definitely better than X levels of if-else. I also thinking about extract this part to "extractValue" function, "NewEqualsValue(unk...)", or introduce "BoolCondition", but hard to say for me will be this better, or worse, so, I think current implementation also will work.

Please review

ruflin:
jenkins, test it

libbeat/processors/condition.go
```diff
@@ -257,16 +267,32 @@ func (c *Condition) checkEquals(event ValuesMap) bool {
...
+			if sValue != equalValue.Str {
 				return false
 			}
-			if sValue != equalValue.Str {
+
+			continue
+		}
+
+		bValue, err := extractBool(value)
+
```
ruflin: Remove newline

libbeat/processors/condition.go
```diff
@@ -257,16 +267,32 @@ func (c *Condition) checkEquals(event ValuesMap) bool {
...
-		} else {
-			sValue, err := extractString(value)
-			if err != nil {
-				logp.Warn("unexpected type %T in equals condition as it accepts only integers and strings. ", value)
+
+			continue
+		}
+
+		sValue, err := extractString(value)
+
```
ruflin: nit: remove new line

libbeat/processors/condition.go
```diff
@@ -118,13 +119,22 @@ func (c *Condition) setEquals(cfg *ConditionFields) error {
...
+			continue
 		}
+
+		bValue, err := extractBool(value)
+		if err == nil {
+			c.equals[field] = EqualsValue{Bool: bValue}
+			continue
+		}
+
+		return err
```
ruflin: I wonder if we should create our "own" error here to describe better what happened. Otherwise just return the `err` from extractBool.

libbeat/processors/condition.go
```diff
@@ -257,16 +267,32 @@ func (c *Condition) checkEquals(event ValuesMap) bool {
...
+
+		if err == nil {
+			if bValue != equalValue.Bool {
 				return false
 			}
+
+			continue
 		}
+
+		logp.Warn("unexpected type %T in equals condition as it accepts only integers, strings or bools. ", value)
```
ruflin: Can we change this to `logp.Err`? In general we prefer not to use `Warn` and make things either `Info` or `Err`.

libbeat/processors/condition.go
ewgRa: Done

Have the same thought, just don't know how far changes can be.

libbeat/processors/condition.go
ewgRa: Done in both cases.

Can you give explanations? Is there coding style agreements/convention and how it looks like?

I think would be nice to add this to "make fmt" for example, anyway, would be nice to give this work to some script. If you give me explanations, I will try to prepare PR that made such checks automatically as part of CI process.

libbeat/processors/condition.go
ewgRa: Done

libbeat/processors/condition.go
ruflin: This is more a beats team convention that we normally check an error directly on the next line. Not sure if this can be covered with a script as it's pretty specific to the error case. Don't think too much about it.

ruflin:
jenkins, test it

libbeat/processors/condition.go
ewgRa: Is it possible to publish beats team convention in Dev Guide? 
There is only one mention of coding style related to EditorConfig, but looks like there is another. I think it save a lot of time to you and contributors.

And perhaps it is possible to automate, with parser, or ast, we just need to check that it is "if" and in expression there is value with type "error", but I'm not sure, haven't so much experience with ast and parser.

ewgRa: Ok, looks like for fixing coding style we can start looking for something similar as https://github.com/golang/go/tree/master/src/cmd/fix or gofix. I pretty sure there is a possibility to do desired. I think I will try later.

ruflin: @ewgRa All the code rules we require are covered by `make fmt`. The other ones like no new line to check the error and I'm sure there are a few others, are not very strict. We should add them to the Dev Guide but TBH we as a team would also have to think about which these actually are. It would be great if all of them could be checked by a tool

Issue #5953, opened 2017-12-23
Kibana docker hangs our test on travis and jenkins


ph:
We have been noticing a few hangs in our test suite which seems to come from the merge of the Kibana docker changes in https://github.com/elastic/beats/pull/5928

```
---> 9efd55ef328a
Removing intermediate container 69e5eaebda55
Step 18/20 : USER kibana
 ---> Running in 1251f3fcec91
 ---> 5f05ea710ee4
Removing intermediate container 1251f3fcec91
Step 19/20 : RUN if [ ${XPACK} = "1" ]; then bin/kibana-plugin install ${DOWNLOAD_URL}/kibana-plugins/x-pack/x-pack-${ELASTIC_VERSION}.zip?c=${CACHE_BUST}; fi
 ---> Running in 4170c8755aa5
Attempting to transfer from https://snapshots.elastic.co/downloads/kibana-plugins/x-pack/x-pack-7.0.0-alpha1-SNAPSHOT.zip?c=20171201
Transferring 282481096 bytes....................
Transfer complete
Retrieving metadata from plugin archive
Extracting plugin archive
Extraction complete
Optimizing and caching browser bundles...
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated
```

ph:
@ruflin could you take a look?

ruflin:
@ph I think this one is out of our control. It's potentially only indirectly related to #5928 The reasons this times out (especially on travis)  is because the optimize process takes longer then 10 minutes on a machine with not too much processing power. I also see this on my local laptop from time to time. The problem could be circumvented if there would be an option that during the optimize some logs are outputed but there is no option for that in Kibana.

I would expect it to work on Jenkins except if Kibana snapshot is broken (which can happen from time to time) but is a good way to figure out issues early. I tried the most recent version locally and it seemed to build Kibana as expected (but it takes a long time). I will flush all images again and retry.

The way I'm hoping to solve this problem is that we will have the docker images for snapshots also available somewhere in a registry, meaning we don't have to build it ourself each time. This is potentially going to happen soon ;-)

ruflin:
On one of the most recent jenkins builds I got an output I haven't seen so far:

```
04:42:33 Optimizing and caching browser bundles...
04:54:06 
[...]

PR #5955, opened 2017-12-24
add allow_failures for travis

ewgRa:
As I see, for now there is a lot of PR do not pass CI, since there is timeouts for some tests.

There is several ways to fix it:

1. Increase timeout

2. Reorganize tests to avoid job fail because of timeout

3. Setup allow_failures in travis

All this three ways better than how it work now (fail by timeout). First way impossible to do, since looks like timeouts hardcoded inside travis.

Second way the best option, but needs time.

So, this PR implement third way as a trade of. After merge this PR, there is can be issue created to reorganize tests and drop "allow_failures".

ps. Also test build take so much time, I think there is must be "fast" tests. Or think about speedup tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -127,6 +127,9 @@ jobs:
         - TARGETS="-C deploy/kubernetes test"
         - TRAVIS_KUBE_VERSION=v1.8.0
       stage: test
+  allow_failures:
+    - env: TARGETS="-C filebeat testsuite"
+    - env: TARGETS="-C libbeat testsuite"
 
 addons:
   apt:


ruflin:
@ewgRa Thanks for taking this one. Some notes here on why the tests fail: https://github.com/elastic/beats/issues/5953

For now I don't think allow the tests to fail is the way forward. If a build is green, I assume tests passed which is not the case here. If I see it being red, I will look it up to see what failed and will potentially restart it. For this reason I would like to keep it as is for now.

About the time the tests need: Agree that they started to take too long, especially in the Metricbeat case. We are working on ideas how to speed it up. For local development most tests can be run very quickly with just `go test` but the integration tests take much longer and more resources.

ewgRa:
Ok, than better to solve it in #5953, this is right way, agree, but question only when it will be solved, since solution not obvious.

ruflin:
The good news is that on Jenkins it seems to be fine as there are more performant machines. And that is what we trust in in the end. Some fixes will also rely on the Kibana team probably. 

Here one stab to get the filebeat builds greener: https://github.com/elastic/beats/pull/5959

PR #5956, opened 2017-12-25
add plugin event processor

ewgRa:
This processor add ability to easy implement your own processor and do what you want with event.

What do you think?

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -14,6 +14,7 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
 
 *Affecting all Beats*
 
+- Add new `plugin` processor that allow to use go plugins to process events. {pull}5956[5956]
 - Support dashboard loading without Elasticseach {pull}5653[5653]
 - Addition and update of the following libraries: pbkdf2, terminal, windows, unix {pull}5735[5735]
 - Update the command line library cobra and add support for zsh completion {pull}5761[5761]
--- a/libbeat/cmd/instance/beat.go
+++ b/libbeat/cmd/instance/beat.go
@@ -47,6 +47,7 @@ import (
 	_ "github.com/elastic/beats/libbeat/processors/add_docker_metadata"
 	_ "github.com/elastic/beats/libbeat/processors/add_kubernetes_metadata"
 	_ "github.com/elastic/beats/libbeat/processors/add_locale"
+	_ "github.com/elastic/beats/libbeat/processors/plugin"
 
 	// Register autodiscover providers
 	_ "github.com/elastic/beats/libbeat/autodiscover/providers/docker"
--- a/libbeat/processors/plugin/foo_plugin.go.source
+++ b/libbeat/processors/plugin/foo_plugin.go.source
@@ -1,0 +1,10 @@
+package main
+
+import (
+	"github.com/elastic/beats/libbeat/beat"
+)
+
+func Run(event *beat.Event) (*beat.Event, error) {
+	event.PutValue("foo", "bar")
+	return event, nil
+}
--- a/libbeat/processors/plugin/plugin.go
+++ b/libbeat/processors/plugin/plugin.go
@@ -1,0 +1,55 @@
+package plugin
+
+import (
+	"plugin"
+
+	"github.com/pkg/errors"
+
+	"github.com/elastic/beats/libbeat/beat"
+	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/processors"
+)
+
+type pluginProc struct {
+	File string
+}
+
+func init() {
+	processors.RegisterPlugin("plugin", newPlugin)
+}
+
+func newPlugin(c *common.Config) (processors.Processor, error) {
+	config := struct {
+		File string `config:"file"`
+	}{}
+
+	err := c.Unpack(&config)
+	if err != nil {
+		return nil, errors.Wrap(err, "fail to unpack the plugin processor configuration")
+	}
+
+	var plg pluginProc
+
+	plg.File = config.File
+
+	return plg, nil
+}
+
+func (p pluginProc) Run(event *beat.Event) (*beat.Event, error) {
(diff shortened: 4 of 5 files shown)


ewgRa:
Ok, looks like "plugin.Open: plugin was built with a different version of package" need to be solved by build plugin during tests. It can be done if pull request idea will be accepted in general.

ruflin:
I wonder how this is different from what we have already in https://github.com/elastic/beats/tree/master/libbeat/plugin ? Here are some examples for processors plugin which should already work now https://github.com/urso/beats-sample-plugins/tree/master/plugins

But I think @urso is best to comment here.

ewgRa:
Indeed, looks quite similar, only difference that I consider custom "plugin" just a one function, that called inside "plugin" processor, and in @urso version plugin considered as a whole processor implementation.

Is there documentation for @urso version? It kind of expected to be here https://www.elastic.co/guide/en/beats/packetbeat/current/defining-processors.html, but there is nothing. Because of experimental?

andrewkroh:
There are not any official docs AFAIK. There’s another example of using the plugin mechanism at https://github.com/andrewkroh/beats-processor-fingerprint.

ewgRa:
Ok, than up to you just to decline this PR, since functionality already implemented.

Looks like functionality the same, but in my approach it is easy to implement and use, just a standard one processor with one function in plugin. I even think not to import of beats Event inside plugin, and just work with "map" directly, it can help to avoid problems with different versions of packages. Also in my implementation plugin know less about internals of beats, that help support backward compatibility in future.

ruflin:
As far as I understand, this change would require us to package beats with `-buildmode=plugin`? As plugins are still a bit shaky and getting to work needs quite some knowledge we don't do this so far.

ewgRa:
> this change would require us to package beats with -buildmode=plugin?

No, this option used only to build *.so plugin files, that later can be opened with "plugin" package in beats. In my approach instructions looks like this:
1. Write plugin source code where is only Run function necessary as entry point
2. Build plugin with -buildmode=plugin, you will have foo.so file
3. Setup processor like:
processors:
- plugin:
   file: foo.so

For exists one implementation it looks like:
1. Write plugin source code with full implementation of processor, like in beats
2. Build plugin with -buildmode=plugin, you will have foo.so file
3. Run beats with -plugin command line arguments.

urso:
The plugin version we have in beats is indeed a little more complicated.

Some Reasons:
- We want to be able to bundle multiple plugins into one `.so` file. The `.so` file contains complete runtime + dependencies. That is, these files can become hughe for simple plugins.
- A processor plugin should be instantiable multiple times. This would work by using the `plugin` processor multiple time
- As processor plugins are instantiable multiple types, they should be configurable. As you only implement a `Run` method, there is no way of passing additional settings. That is plugins need to be super specific/custom + you might end up with a few very big `.so` files.

Your foo example can be implemented like this:

```
package main

var Bundle = processors.Plugin("foo", createFoo)

type foo struct{}

func createFoo(cfg *common.Config) (processors.Processor, error) {
  return &foo{}, nil
}

func (f *foo) Run(event *beat.Event) (*beat.Event, error) {
  event.PutValue("foo", "bar")
  return event, nil
}
```

In case the `foo` pattern with very specialised plugins is a quite common pattern for you, this can be simplified a little by introducing some kind of ProcessorFn (in a separate package):

```
package processorfn

type Fn struct {
  fn func(*beat.Event) (*beat.Event, error)
}


func New(fn func(*beat.Event) (*beat.Event, error)) *Fn {
  return &Fn{fn}
}

func Create(fn func(*beat.Event) (*beat.Event, error)) func(*common.Config) (processors.Processor, error) {
  f := New(fn)
  func(*common.Config) (processors.Processor, error) {
    return f, nil
  }
}

func (f *Fn) Run(event *beat.Event) (*beat.Event, error) {
  return f(event)
}
```


Then `foo` becomes:

```
package main

import (...)

var Bundle = processors.Plugin("foo", processorfn.Create(run))

func run(event *beat.Event) (*beat.Event, error) {
  event.PutValue("foo", "bar")
  return event, nil
}
```

This still allows you to bundle multiple plugins into one `.so` files.

e.g.

myplugin.go:

```
package main

import (
  ...
  
  "path/to/foo"
  "path/to/foo2"
)

var Bundle = plugins.Bundle(
  processors.Plugin("foo", processorfn.Create(foo.Run)),
  processors.Plugin("foo2", processorfn.Create(foo2.Run)),
)
```

with

```
package foo

import (...)

func Run(event *beat.Event) (*beat.Event, error) {
  event.PutValue("foo", "bar")
  return event, nil
}
```

Using this approach one can provide plugin building packages of different "granularity". e.g.:

```
type eventFn func(map[string]interface{}) (map[string]interface{}, error)

Create(fn eventFn)  func(*common.Config) (processors.Processor, error) 
```

or

```
type eventFn func(*beat.Event) (*beat.Event, error)

Create (fn func(*common.Config) (eventFn, error)) func(*common.Config) (processors.Processor, error)
```

These support packages can very well be provided by libbeat (including tests, so plugin support get's properly updated maintained).

Given go plugins themselves are quite experimental (not only in beats), that's a quite an ok compromise I think.

ewgRa:
Thanks for explanation.

Looks complicated just because usage case is super complex.
But even with this case I think it can be even simple.

```
processors:
- plugin:
   file: foo.so
   entry_function: FooFunction
   instance: FooInstance
   config: ...
```
or similar to that. entry_function, config and instance will be optional, if you have super simple case. If you want multiple plugin instances, different configs and so on - expand configuration.

Only one good argument is: "Given go plugins themselves are quite experimental (not only in beats), that's a quite an ok compromise I think.", but as far as I understand current implementation use the same approach, there is same go plugin .so files.

Anyway, I will close this for now, feel free to continue discussion, if you saw something promising here that can reduce complexity of plugins.

ruflin:
@ewgRa I still keep coming back to this. I think you triggered an interesting discussion here if we should spend more time to figure out ways to make plugins even simpler for Beats or if this would only cause more issues. No conclusion on this yet and I don't think we have one very soon but it's good you triggered this discussion. Thanks for your contribution.

Issue #5942, opened 2017-12-21
Allow for modifying json field names with dots in them in beats


dl1ely:
Following the suggestion in https://discuss.elastic.co/t/allow-for-renaming-of-json-fields-in-http-module-output/112461 i open this issue to discuss the challenges regarding json field names with dots in them when these get collected/created by beats, especially metricbeat.

The problem is also described in https://discuss.elastic.co/t/v6-0-0-beta2-http-module-mapping-trouble/99703?source_topic_id=112461 and was already addressed in httpbeat, the predecessor of the metricbeat http module (https://github.com/christiangalsterer/httpbeat/blob/master/config/config.go#L34). Unfortunately, the http module does not allow a replacement of dots in json keys collected from the source.

The issue also surfaced in Spring Boot Actuator https://github.com/spring-projects/spring-boot/issues/10449 but it affects other producers of metrics data too, and will not be fixed in Spring Boot 1.5

christiangalsterer:
@ruflin: If you have an idea/preference/suggestion how to address this I can have a look into it.

ruflin:
I think this is actually problem we have to solve in different places but probably have a method internally to do it. In general we should have  a method replacing dot separated keys with `_`  as we already do for example for docker labels.

Here the different fixes we should do:

* Jolokia: We should do it here by default as part of the metricset as it happens too often
* http json metricset: Have a config option there
* JSON processor: We should allow to do this as an option there too, which probably internally can be used by the http/ json metricset.

@christiangalsterer Would be great to have a PR for the `http/json` metricset, I will have a look at the jolokia one.

ruflin:
BTW here is the discussion from the Jolokia side: https://discuss.elastic.co/t/include-fields-problem/109809

ruflin:
@christiangalsterer You can use the dedot function introduced here: https://github.com/elastic/beats/pull/5957

christiangalsterer:
@ruflin: Thanks for the info. Will have a look into the `http/json` metricset.

ruflin:
@christiangalsterer Awesome. Ping me if I can be of any help.

walktall:
Just like docker labels, add_kubernetes_metadata processors also need to handle k8s labels and annotations.

[...]

PR #5957, opened 2017-12-26
Dedot keys for jolokia metricsets

ruflin:
This should prevent name collisions and allow filtering on these fields. The feature is enabled by default and cannot be disable as otherwise ingestion could stop as Elasticsearch is returning an error. Currently not config option is provided as this should not be disabled.

* Introduces DeDot function in common package
* Adds tests to jmx for new functionality
* Use new function also for docker labels

Related to https://github.com/elastic/beats/issues/5942
Closes https://github.com/elastic/beats/pull/5916

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -38,6 +38,7 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
 - Rename `http.response.status_code` field to `http.response.code`. {pull}5521[5521]
 - Rename `golang.heap.system.optained` field to `golang.heap.system.obtained`. {issue}5703[5703]
 - Support haproxy stats gathering using http (additionaly to tcp socket). {pull}5819[5819]
+- De dot keys in jolokia/jmx metricset to prevent collisions. {pull}5957[5957]
 
 *Packetbeat*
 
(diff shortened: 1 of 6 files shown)


walktall:
@ruflin [This method](https://github.com/elastic/beats/blob/master/libbeat/processors/add_kubernetes_metadata/indexers.go#L123) should also apply `common.DeDot` for keys of labelMap and annotationsMap.

ruflin:
@walktall Good point. Can you reference that here on this issue so we don't loose track of it? https://github.com/elastic/beats/issues/5942

ruflin:
@andrewkroh Could you have a look at this one. It should be ready to go.

CHANGELOG.asciidoc
```diff
@@ -37,6 +37,7 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
 - Rename `http.response.status_code` field to `http.response.code`. {pull}5521[5521]
 - Rename `golang.heap.system.optained` field to `golang.heap.system.obtained`. {issue}5703[5703]
 - Support haproxy stats gathering using http (additionaly to tcp socket). {pull}5819[5819]
+- De dot keys in jolokia/jmx metricset to prevent collisions.
```
andrewkroh: Can you add the PR number.

christiangalsterer: done.

ruflin: Added and rebased PR.

PR #5959, opened 2017-12-27
Disable x-pack for filebeat tests

ruflin:
For filebeat integration tests x-pack is not required.

* Introduce separate build command in environment Makefile


ruflin:
Closing as x-pack is needed for the ml part. I wonder if we could / should disable x-pack only for Kibana as we don't test the UI part.

PR #5960, opened 2017-12-27
Start cleaning up condition tests

ruflin:
This simplifies part of the condition tests to use table tests. Like this the expected result and the condition itself are close together in the code which makes it easier to read.

Cleanup was triggered when looking at https://github.com/elastic/beats/pull/5954 and realised it hard to follow in a diff.


exekias:
Yes! sorry, I re-spined some tests yesterday

PR #5966, opened 2017-12-28
Improve Kibana docker image build

### LINKED ISSUE #5953: Kibana docker hangs our test on travis and jenkins

ruflin:
* Add `NODE_OPTIONS="--max-old-space-size=4096"` to circumvent elastic/kibana#15683
* Make it more obvious in the Docker file which parts are copied over and which are from the official repo
* Remove xpack.monitoring config option as this seems to cause a rebuild.
* Increase LS timeout for health check
* Remove duplicated kafka health check
* Increase health check timeouts to 10min
* Set defaults for `ARGS` in kibana container. This makes it easy to also build the container directly.
* Adjusts docker image paths to all use `ES_BEATS` variable.
* Remove `detector_rules` for ML as not supported / required anymore in master.
* Remove unused `SHIELD=false` variable in compose file
* Fix geoip looksup that changed and improve error logging

It would be great if we could disable most of the Kibana x-pack plugins as they are not needed for our testing. But the problem is 1) we cannot install only 1 plugin AFAIK 2) disabling one of the plugins causes a rebuild which makes the build even longer.

Closes #5953

### DIFF
--- a/testing/environments/docker/kibana/Dockerfile-snapshot
+++ b/testing/environments/docker/kibana/Dockerfile-snapshot
@@ -1,19 +1,22 @@
 # Copy of https://github.com/elastic/kibana-docker/blob/master/build/kibana/Dockerfile
 FROM centos:7
 LABEL maintainer "Elastic Docker Team <docker@elastic.co>"
-
 EXPOSE 5601
 
-ARG DOWNLOAD_URL
-ARG ELASTIC_VERSION
+
+### Beats specific args ####
+ARG DOWNLOAD_URL=https://snapshots.elastic.co/downloads
+ARG ELASTIC_VERSION=7.0.0-alpha1-SNAPSHOT
 ARG CACHE_BUST=1
-ARG XPACK
+ARG XPACK=1
+
+# Healthcheck create by beats team
+RUN yum install update -y epel-release && yum install -y jq
+HEALTHCHECK --interval=1s --retries=600 CMD curl -f http://localhost:5601/api/status | jq '. | .status.overall.state' | grep -q green
+### Beats ###
 
 # Add Reporting dependencies + healthcheck tool
-RUN yum update -y && yum install -y epel-release
-RUN yum install -y fontconfig freetype jq && yum clean all
-
-HEALTHCHECK --interval=1s --retries=300 CMD curl -f http://localhost:5601/api/status | jq '. | .status.overall.state' | grep -q green
+RUN yum update -y && yum install -y fontconfig freetype && yum clean all
 
 WORKDIR /usr/share/kibana
 RUN curl -Ls ${DOWNLOAD_URL}/kibana/kibana-${ELASTIC_VERSION}-linux-x86_64.tar.gz?c=${CACHE_BUST} | tar --strip-components=1 -zxf - && \
@@ -39,6 +42,8 @@ RUN groupadd --gid 1000 kibana && \
       kibana
 USER kibana
 
-RUN if [ ${XPACK} = "1" ]; then bin/kibana-plugin install ${DOWNLOAD_URL}/kibana-plugins/x-pack/x-pack-${ELASTIC_VERSION}.zip?c=${CACHE_BUST}; fi
+# Beats specific check for XPACK to have both variables in one
+RUN if [ ${XPACK} = "1" ]; then NODE_OPTIONS="--max-old-space-size=4096" bin/kibana-plugin install ${DOWNLOAD_URL}/kibana-plugins/x-pack/x-pack-${ELASTIC_VERSION}.zip?c=${CACHE_BUST}; fi
 
 CMD ["/bin/bash", "/usr/local/bin/kibana-docker"]
+
--- a/testing/environments/docker/kibana/config/kibana-x-pack.yml
+++ b/testing/environments/docker/kibana/config/kibana-x-pack.yml
@@ -6,4 +6,15 @@ server.host: "0"
 elasticsearch.url: http://elasticsearch:9200
 elasticsearch.username: elastic
 elasticsearch.password: changeme
-xpack.monitoring.ui.container.elasticsearch.enabled: true
+
+# The following config options are currently commented out as they would cause a rebuild of kibana which takes a long time.
+
+#xpack.monitoring.ui.container.elasticsearch.enabled: true
+
+# Potentially disabled for beats testing when rebuilding is not needed
+#xpack.apm.enabled: false
+#xpack.graph.enabled: false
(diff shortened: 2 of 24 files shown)


testing/environments/docker/kibana/Dockerfile-snapshot
```diff
@@ -1,19 +1,16 @@
 # Copy of https://github.com/elastic/kibana-docker/blob/master/build/kibana/Dockerfile
 FROM centos:7
 LABEL maintainer "Elastic Docker Team <docker@elastic.co>"
-
 EXPOSE 5601
 
+# Beats specific args
```
ruflin: @jarpy Below is one the main difference between the original image (https://github.com/elastic/kibana-docker/blob/master/templates/Dockerfile.j2) and what we use for beats. The reason we use `ARG` as this allows us to use one Dockerfile and one docker-compose file an only change the args depending on the environment. Any idea how we could rely even more on the official images?

testing/environments/docker/kibana/Dockerfile-snapshot
```diff
@@ -42,3 +39,9 @@ USER kibana
 RUN if [ ${XPACK} = "1" ]; then bin/kibana-plugin install ${DOWNLOAD_URL}/kibana-plugins/x-pack/x-pack-${ELASTIC_VERSION}.zip?c=${CACHE_BUST}; fi
 
 CMD ["/bin/bash", "/usr/local/bin/kibana-docker"]
+
+# Healthcheck create by beats team
+USER root
+RUN yum install -y epel-release && yum clean all
+HEALTHCHECK --interval=1s --retries=300 CMD curl -f http://localhost:5601/api/status | jq '. | .status.overall.state' | grep -q green
```
ruflin: @dliappis  That is how we do the healhcheck for kibana (rather complicated as we need additional packages). Based on https://github.com/elastic/elasticsearch-docker/issues/60#issuecomment-340920511 I assume healthcheck will not show up in our official images. The problem with the above is that we can't do it just in docker-compose as some packages are missing. Are there alternative solutions here?

dliappis: @ruflin Best would be to add all the missing packages to be able to define health checks. For example we include `wget` and `curl` in the elasticsearch image, feel free to specify which additional packages you need in the image to properly implement health checks (e.g. `jq`?)

ruflin: @dliappis Yes, here it is `jq` as we need to parse the json response from Kibana status. It would be nice if there would be an API from Kibana to get the status which is easier to parse with the default Linux tools, something like `_cat` in ES.

testing/environments/docker/kibana/config/kibana-x-pack.yml
```diff
@@ -6,4 +6,12 @@ server.host: "0"
 elasticsearch.url: http://elasticsearch:9200
 elasticsearch.username: elastic
 elasticsearch.password: changeme
-xpack.monitoring.ui.container.elasticsearch.enabled: true
+#xpack.monitoring.ui.container.elasticsearch.enabled: true
```
ruflin: @jbudz This config option is in the new default docker file from Infra. It seems when I have this line in on startup it rebuilds kibana again. I read this about the plugins enabling below. Is this also the case for this var?

ruflin:
@andrewkroh @ph This seems to now finally work on Jenkins. Let's get this in an rebase the other PR's on top of it.

ruflin:
Unfortunately for libbeat I still see the following in the logs:

```
{"type":"log","@timestamp":"2017-12-29T05:00:00Z","tags":["info","optimize"],"pid":1,"message":"Optimizing and caching bundles for ml, stateSessionStorageRedirect, status_page, timelion, graph, monitoring, login, logout, dashboardViewer, apm and kibana. This may take a few minutes"}
```

It seems the Kibana plugins are optimized and cached once again after the build?

testing/environments/docker/kibana/config/kibana-x-pack.yml
jbudz: I wouldn't expect this to, the rough heuristic is anything that changes the client side code will cause a rebundle.

I tried toggling this on the docker.elastic.co/kibana/kibana:6.1.1 images and it didn't cause any rebuilds - I'll give it a try with your dockerfile.

ruflin: The image used here is almost identical to our official image with x-pack based on 7.0.0-alpha1-SNAPSHOT. I assumed we the exact same build is used for our filebeat and libbeat build, but I didn't see it on both builds (also 1 passed). So either it could be that I played around with too many different things and things got messed up when I tried to reproduce it or it's not always triggered. 

Are there any environment variables that can trigger a rebuild?

andrewkroh:
Looks like you fix the ML failure in libbeat. The only failures I see now are for Filebeat modules. It appears that on Travis and Jenkins the same test failed, but in different places.

ruflin:
@andrewkroh Ready to go

andrewkroh:
Awesome!

ruflin:
@jbudz Even though the PR is merged, I think our issue still exists. It is just that now Jenkins waits long enough to do stuff. Interestingly if I run just the kibana image alone, I don't see the rebuild (yet).

Issue #3480, opened 2017-01-27
Feature: add add_host_metadata similar to add_cloud_metadata but for self hosted environments


jakommo:
In some use cases it could be beneficial to gather more information's on the sending side. 
E.g. get the source IP of the client sending the event.
While we already have `beat.hostname`, there are environments where this is not unique enough.
It's already possible to work around this by adding a custom field and using a environment variable with the IP as a value, but it's complicated when there are multiple interfaces and it needs to be deployed on a lot of (differently configured) hosts. 

Proposing to have something like [add_cloud_metadata](https://www.elastic.co/guide/en/beats/filebeat/master/add-cloud-metadata.html) but for self hosted environments, that hold additional information on the host sending the event.

andrewkroh:
This seems related to this request https://github.com/elastic/beats/issues/4356 (system.info metricset request).

jakommo:
@andrewkroh I think it's related but not the same.  https://github.com/elastic/beats/issues/4356 sounds more like reading those stats from the OS, while this issue was intended to get the IP of the interface sending the event. 

I could imagine that this would be none static, i.e. if one interface / network goes down and another interface / route is used, it should reflect that by tagging the events with that new interface IP. 

I saw this request a few times already, but mostly related to Filebeat. So maybe we should change the Label?

ruflin:
I started some work here: https://github.com/elastic/beats/pull/5968

I changed the label to `libbeat` as this would be a processor in libbeat.

urso:
Feature has been merged last year, plus 'beta' label has been removed recently.

Issue #5946, opened 2017-12-22
Proposal: Log system details to aid in support and debugging


andrewkroh:
At startup we should log details about the system to aid in providing support. This information could help us resolve issues faster because there would be fewer back-and-forths. When logs are provided by the user there wouldn't be a need for questions like "how did you run it", "what OS", "what kind of hardware", "what user were you running as", "are you running in Docker", "what is the machine's local timezone", etc.

For example, at startup we could add a single log line with the system info. We've had [similar logging](https://github.com/elastic/gosigar/blob/36e30f86fab34ac5e56a59cfe80929d65babb00d/sys/windows/privileges.go#L95-L102) limited to Metricbeat-Windows and it has proven useful when debugging permissions issues.

```
2017-12-22T01:27:46.710-0500	INFO	instance/beat.go:443	Home path: [/Users/akroh/go/src/github.com/elastic/beats/filebeat] Config path: [/Users/akroh/go/src/github.com/elastic/beats/filebeat] Data path: [/Users/akroh/go/src/github.com/elastic/beats/filebeat/data] Logs path: [/Users/akroh/go/src/github.com/elastic/beats/filebeat/logs]
2017-12-22T01:27:46.710-0500	INFO	instance/beat.go:450	Beat UUID: 06925eef-4707-40f9-82aa-acfe7903dde9
[...]

PR #5968, opened 2017-12-29
Add host metadata processor

ruflin:
This adds a processor to add some information about the host machine to each event very similar to what we have for the cloud metadata processor. The idea is to make it possible to allow better filtering based on additional information about the host if needed. This information partially overlaps with what we have in `beat.hostname` but adds additional fields. Other fields like the IP address(es) could be also added here.


libbeat/processors/add_host_metadata/add_host_metadata.go
```diff
@@ -0,0 +1,46 @@
...
+
+type addHostMetadata struct{}
+
+func newHostMetadataProcessor(_ *common.Config) (processors.Processor, error) {
+	return &addHostMetadata{}, nil
+}
+
+func (p addHostMetadata) Run(event *beat.Event) (*beat.Event, error) {
+
+	info, err := host.Info()
```
exekias: Do you have any benchmark on this? it would make sense to move it to the constructor and cache the result, we just need to document that a hostname change won't get reflected here

andrewkroh:
I was kind of working on something related over the break. It can (or will be able to) fetch [some of the information](https://github.com/andrewkroh/go-sysinfo/blob/222d280c6e0e6eee0b79f057765505031e3707e0/types/host.go#L9-L21) that you want to add with this processor. I am working on this because I wanted to log (https://github.com/elastic/beats/issues/5946) some of this information and also make this data available for xpack monitoring.

https://github.com/andrewkroh/go-sysinfo

tsg:
Big +1 to this. I think a processor makes sense to mirror `add_cloud_metadata` and co, but we can put it into the default configuration file.

ruflin:
@andrewkroh Very nice. Did not look into your code yet. Did you use a third party lib or implement it yourself?

monicasarbu:
+1 on adding `add_host_metadata`, that will also add cloud metadata in the future.

ruflin:
This will close https://github.com/elastic/beats/issues/3480

libbeat/processors/add_host_metadata/add_host_metadata.go
ruflin: I changed it to be instantiated in the constructor. I'm pretty sure later a request for expiring the info is coming up :-)

libbeat/processors/add_host_metadata/add_host_metadata.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+func (p addHostMetadata) Run(event *beat.Event) (*beat.Event, error) {
+
+	data := common.MapStr{
+		"host": common.MapStr{
+			"hostname": p.info.Hostname,
+			"id":       p.info.UniqueID,
+			"os": common.MapStr{
+				"architecture": p.info.Architecture,
```
ruflin: @andrewkroh Seeing that you have `Architecture` outside of `info` I have second thoughts on puttig it under `os` as I agree it more belongs to the `host` then the os itself.

Also thinking about putting `os` info on the top level.

libbeat/processors/add_host_metadata/add_host_metadata.go
```diff
@@ -0,0 +1,50 @@
+package add_host_metadata
+
+import (
+	sysinfo "github.com/elastic/go-sysinfo"
```
ruflin: @andrewkroh Did you have to add a `-` to the lib name? :-D

andrewkroh: With Go you don't need to add a package alias, just import plain old "github.com/elastic/go-sysinfo" and use `sysinfo.XYZ` in the code.

andrewkroh:
I'm glad you are trying out go-sysinfo. I'll take a closer look at this next week. I've been sitting on a bunch of updates to go-sysinfo, that I haven't had time to cleanup. I plan on getting the lib into a better place so that this can make 6.3.

libbeat/processors/add_host_metadata/add_host_metadata.go
ruflin: It was mainly my IDE that complained :-)

ruflin:
@andrewkroh Good timing. I plan to put some work into this next week again. But also plan to start with a very basic version first which we then can expand on. The current fields are probably already enough for a first version.

andrewkroh:
I think we no longer need to keep `github.com/elastic/procfs` in vendor/. I'm pretty sure any changes got pushed upstream. Can you try removing references to `github.com/elastic/procfs` and replace them with `github.com/prometheus/procfs` so we don't have a duplicate dep under two names.

ruflin:
`github.com/elastic/procfs` is still used in the `system/socket` and `system/raid` metricset. I suggest to make this change in a follow up PR and not mix it with this one.

ruflin:
Here is a separate PR that changes the dependency. We can merge the other one first and rebase on top of it: https://github.com/elastic/beats/pull/6548

ruflin:
This PR currently fails because of a crosscompile issue with darwin 32 bit: https://github.com/elastic/go-sysinfo/pull/3

andrewkroh:
I believe https://github.com/elastic/go-sysinfo/pull/4 should fix the cross-compile issue seen [here](https://travis-ci.org/elastic/beats/jobs/353341250#L564). The sysinfo code was lacking cgo tags to prevent the darwin provider from being included when cross-compiling on Linux (without a C x-compiler).

TimWardOrigami:
Is the host's FQDN included in the output (eg "hostname -f" output)? If not should/can it be?

TimWardOrigami:
@ruflin How do I "add a feature request" please?

ruflin:
@TimWardOrigami Just open a new Github issue in this repo, explain quickly on what it is a about and what it is needed for. I will label it then accordingly.

PR #5970, opened 2018-01-01
i5942: http/json module: Allow for modifying json field names with do…

christiangalsterer:
- Introduced new config parameter `dedot.enabled`, with default value `false`.
- Made all necessary changes to implementation, added test case and documentation.
- Uses a local implementation for actual replacement of file and not yet the implementation in common.DeDot introduced in https://github.com/elastic/beats/pull/5957. Waiting for this PR to get merged and will change to this implementation (already prepared, just needs to be enabled).

### DIFF
--- a/metricbeat/module/http/json/json.go
+++ b/metricbeat/module/http/json/json.go
@@ -50,6 +50,7 @@ type MetricSet struct {
 	body            string
 	requestEnabled  bool
 	responseEnabled bool
+	deDotEnabled    bool
 }
 
 // New create a new instance of the MetricSet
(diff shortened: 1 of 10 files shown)


christiangalsterer:
Included changes from #5957.

metricbeat/module/http/json/json.go
```diff
@@ -153,3 +162,16 @@ func (m *MetricSet) getHeaders(header http.Header) map[string]string {
 	}
 	return headers
 }
+
+func replaceDots(data interface{}) interface{} {
```
ruflin: I wonder if we should have this recursive function also in libbeat later on.

ruflin:
jenkins, test it

metricbeat/module/http/json/json.go
christiangalsterer: I actually had the very same idea as it is quite likely that also beats have json as payload where in all keys "dots" shall be replaced.

I can open a new PR for this if you OK.

ruflin: Sounds like a good idea.

PR #5971, opened 2018-01-02
Filebeat: osquery module

tsg:
This adds a Filebeat module for centralising osquery logs. 

screencapture-localhost-5601-vjo-app-kibana-1514472334644

osqueryd writes the results in JSON, which makes it fairly easy to ingest to the Elastic stack. The module uses the JSON decoding support in Filebeat, and then renames the fields to match the Beats naming conventions (most fields prefixed with `osquery.result`). There is an option (`use_namespace: false`) to leave the fields as they were in the original JSON, but changing that settings makes the sample dashboards unusable.

Another issue is that the osquery JSON format represents all data as strings (the numbers are quoted). This is both bad and good. It's bad because number aggregations don't work on what should be numbers, but it's good because it means there can't be any type conflicts. We could, potentially, do the type conversions ourselves based on the osquery schema, but that seems risky with regards to the schema changes. The osquery JSON output should also be fixed once osquery switches to RapidJSON, which seems to be [in progress](https://github.com/facebook/osquery/issues/3164).


filebeat/module/osquery/result/_meta/fields.yml
```diff
@@ -0,0 +1,27 @@
...
+        The identifier for the host on which the osquery agent is running.
+        Normally the hostname.
+    - name: unix_time
+      type: long
+      description: >
+        Unix timestamp of the event. Used for computing the `@timestamp` column.
+    - name: calendar_time
+      tupe: keyword
+      description: >
+        String representation of the collection time, as formatted by osquery.
```
ruflin: Interesting that this is called `calendar_time`. I like the description with `collection time`. I wonder if we should have a common field in the future for the `collection_time` (or a different name).

tsg: In osquery, they have `unixTime` and `calendarTime` which are different representations of the same moment. These get translated to `unix_time` and `calendar_time` by the module, and `@timestamp` is computed from `unix_time`. So there's quite a bit of redundancy, we could consider dropping the `calendar_time` by default.

filebeat/module/osquery/result/_meta/fields.yml
```diff
@@ -0,0 +1,27 @@
...
+        or removed. It can be one of "added", "removed", or "snapshot".
+    - name: host_identifier
+      type: keyword
+      description: >
+        The identifier for the host on which the osquery agent is running.
+        Normally the hostname.
+    - name: unix_time
+      type: long
+      description: >
+        Unix timestamp of the event. Used for computing the `@timestamp` column.
```
andrewkroh: Is this in seconds since epoch?

tsg: Yes, I'll update the docs for clarity.

tsg:
Added another dashboard, addressed comments, and rebased to master. If green, this is good for merging from my POV.

filebeat/module/osquery/result/_meta/fields.yml
tsg: I added `read_time` (btw, it's not prefixed by `beat`) as being the read time when filebeat reads the log lines.

filebeat/module/kafka/log/ingest/pipeline.json
```diff
@@ -45,7 +45,7 @@
     {
       "rename": {
         "field": "@timestamp",
-        "target_field": "beat.read_time"
+        "target_field": "read_timestamp"
```
ruflin: @tsg This seems to be a breaking change? Also saw there is no CHANGELOG for this PR.

etursunbaev:
Hi, all!
Cannot find information how to add osquery module to filebeat?
In /usr/share/filebeat there is no osquery module
```
/usr/share/filebeat/module# ls -l
total 20
drwxr-xr-x 4 root root 4096 Feb 19 17:51 apache2
drwxr-xr-x 3 root root 4096 Feb 19 17:51 auditd
drwxr-xr-x 4 root root 4096 Feb 19 17:51 mysql
drwxr-xr-x 4 root root 4096 Feb 19 17:51 nginx
drwxr-xr-x 4 root root 4096 Feb 19 17:51 system
```

jjqq2013:
@etursunbaev your filebeat version seems old.

To use osquery module in older version, you can reference this:
https://github.com/elastic/beats/commit/94ad82c19dd748cbc7a87f3eee419c5d4c699a54#commitcomment-27831592

https://github.com/jjqq2013/misc/tree/master/elasticsearch5.4.0

jjqq2013:
Hi, all, do you know what will be the document type of osquery module?
When I use output to kafka, I need set the topic

```
topic: '%{[fields.log_topic]}'
```

What is the value when use osquery module?

jjqq2013:
Oh, no problem. I found it: I need define fields. log_topic in  prospector.

PR #5972, opened 2018-01-02
Move Kubernetes Pod watcher to libbeat/common 

exekias:
This PR decouples Pod watcher from `add_kubernetes_metata` and enables us to use it for autodiscover.

The code mimics Docker containers watcher, offering buses to get notified of pod creation/update/deletion


libbeat/processors/add_kubernetes_metadata/indexers.go
```diff
@@ -39,15 +40,11 @@ type Indexers struct {
...
-	GenerateMetaData(pod *Pod) common.MapStr
-}
-
-type IndexerConstructor func(config common.Config, genMeta GenMeta) (Indexer, error)
+// IndexerConstructor builds a new indexer from its settings
+type IndexerConstructor func(config common.Config, metaGen kubernetes.MetaGenerator) (Indexer, error)
 
-func NewIndexers(configs PluginConfig, metaGen *GenDefaultMeta) *Indexers {
+// NewIndexers  builds indexers object
+func NewIndexers(configs PluginConfig, metaGen kubernetes.MetaGenerator) *Indexers {
```
vjsamuel: should indexers and matchers be part of `add_kubernetes_metadata` or should they go into `kubernetes`?

libbeat/common/kubernetes/watcher.go
```diff
@@ -0,0 +1,263 @@
...
+		return errors.New("Timeout while doing initial Kubernetes pods sync")
+	case <-synced:
+		// Watch for new changes
+		go p.watch()
+		go p.cleanupWorker()
+		return nil
+	}
+}
+
+func (p *podWatcher) watch() {
```
vjsamuel: do you think this is generic enough that it can be used by any implementations other than autodiscover and add_kubernetes_metadata? Would we need a scenario where we would want to write a controller which behaves a bit differently that we might want to initialize it with a different func that handles add/update/delete a bit differently?

libbeat/processors/add_kubernetes_metadata/kubernetes.go
```diff
@@ -166,7 +192,54 @@ func (k *kubernetesAnnotator) Run(event *beat.Event) (*beat.Event, error) {
 	return event, nil
 }
 
-func (*kubernetesAnnotator) String() string { return "add_kubernetes_metadata" }
+// worker watches pod events and keeps a map of metadata
+func (k *kubernetesAnnotator) worker() {
```
vjsamuel: i like this a lot. lets change the metricbeat event module to use this as well and rely on the same watcher implementation as in this PR.

libbeat/common/kubernetes/watcher.go
```diff
@@ -0,0 +1,263 @@
...
+// Watcher reads Kubernetes events and keeps a list of known pods
+type Watcher interface {
+	// Start watching Kubernetes API for new containers
+	Start() error
+
+	// Stop watching Kubernetes API for new containers
+	Stop()
+
+	// Pod returns the running container with the given ID or nil if unknown
+	Pod(ID string) *Pod
```
walktall: Could you make this `Pod` method more generic? If I want to watch k8s events, it is not a Pod here. I think there should be a more generic interface which should be implemented by resource you want to watch. I have made a PR #5932 and using an `type Resource interface {
	GetMetadata() *ObjectMeta
}
` for this purpose. So in this case, it might be `Resource(id string) Resource`.

libbeat/processors/add_kubernetes_metadata/indexers.go
exekias: Metadata generation must be common, as we use the same for processor and autodiscover, but indexers/matchers are something used only from the processor, I would keep them there until we see the need to make them common

libbeat/common/kubernetes/watcher.go
exekias: I like this approach! What do you think about rebasing/reworking this change once this PR is merged?

walktall: Sure, I will make a PR about this once yours got merged. 😃

libbeat/processors/add_kubernetes_metadata/kubernetes.go
exekias: SGTM, let's wait for the outcome of #5949 as I think it may help here

vjsamuel:
LGTM

ruflin:
@exekias is this in progress or ready to go. Labels and PR description kind of conflict.

libbeat/common/kubernetes/watcher.go
```diff
@@ -0,0 +1,263 @@
...
+		go p.watch()
+		go p.cleanupWorker()
+		return nil
+	}
+}
+
+func (p *podWatcher) watch() {
+	for {
+		logp.Info("kubernetes: %s", "Watching API for pod events")
+		watcher, err := p.client.WatchPods(p.ctx, "", p.nodeFilter)
```
walktall: Should we use lastResourceVersion here since we saved it when doing list.

libbeat/common/kubernetes/watcher.go
```diff
@@ -0,0 +1,263 @@
...
+			//watch pod failures should be logged and gracefully failed over as metadata retrieval
+			//should never stop.
+			logp.Err("kubernetes: Watching API error %v", err)
+			time.Sleep(time.Second)
+			continue
+		}
+
+		for {
+			_, apiPod, err := watcher.Next()
+			if err != nil {
```
walktall: There is a problem I met before while testing.
This [errors](https://github.com/ericchiang/k8s/issues/68) makes it doing WatchPods forever and never stop. So we might need to check the err type returned by Next.
If it is an ioErr, we should retry watch, but if it is an protocol error, maybe skip?

libbeat/common/kubernetes/watcher.go
exekias: I'm +1 to do this in a subsequent PR, it will need more testing, I don't fully remember why we are not using it :innocent:

libbeat/common/kubernetes/watcher.go
exekias: Skip may be risky here? At the end, the break will respawn a new watcher and restart the loop, sounds like a safe bet. If the error is permanent, the watcher creation will sleep for a while.

exekias:
let's merge now @ruflin, as this only moves code, we can update tests in subsequent PRs

libbeat/common/kubernetes/watcher.go
walktall: Watcher creation will not sleep, since `p.client.WatchPods` will not return error, the error was returned by `watcher.Next`, and then it break the inner for, and `p.client.WatchPods` immediately since watcher creation will sleep only when `p.client.WatchPods` got error.

libbeat/common/kubernetes/watcher.go
walktall: I found this method may not needed by interface, since pod entity can be observed by bus.Listener?

libbeat/common/kubernetes/watcher.go
exekias: You are right sorry, WatchPods will sleep after error

walktall: Also, it needs to Close the watcher before break, or it will be leaked here.

exekias: Good point, I've pushed the Close call

exekias:
rebased to bring master fixes

walktall:
Could this PR be merged? @exekias

libbeat/common/kubernetes/types.go
```diff
@@ -98,3 +106,21 @@ type Pod struct {
...
+	bytes, err := json.Marshal(pod)
+	if err != nil {
+		logp.Warn("Unable to marshal %v", pod.String())
+		return nil
+	}
+
+	po := &Pod{}
+	err = json.Unmarshal(bytes, po)
+	if err != nil {
+		logp.Warn("Unable to marshal %v", pod.String())
```
ruflin: This should stay "Unable to unmarshal". I would also suggest to make this and the above error message a bit longer to add some context from where this log message is coming from.

libbeat/common/kubernetes/watcher.go
```diff
@@ -0,0 +1,261 @@
...
+		ctx:                 ctx,
+		stop:                cancel,
+		pods:                make(map[string]*Pod),
+		deleted:             make(map[string]time.Time),
+		bus:                 bus.New("kubernetes"),
+	}
+}
+
+func (p *podWatcher) syncPods() error {
+	logp.Info("kubernetes: %s", "Performing a pod sync")
```
ruflin: Perhaps we could use a kubernetes prefix for the above log message (see comment).

libbeat/common/kubernetes/watcher.go
```diff
@@ -0,0 +1,261 @@
...
+
+func (p *podWatcher) watch() {
+	for {
+		logp.Info("kubernetes: %s", "Watching API for pod events")
+		watcher, err := p.client.WatchPods(p.ctx, "", p.nodeFilter)
+		if err != nil {
+			//watch pod failures should be logged and gracefully failed over as metadata retrieval
+			//should never stop.
+			logp.Err("kubernetes: Watching API error %v", err)
+			time.Sleep(time.Second)
```
ruflin: A note on why this sleep here is needed in the source code would be nice.

ruflin:
@walktall Merged. Would be great to see https://github.com/elastic/beats/pull/5949 reworked on top of this.

walktall:
@ruflin Thanks! I will make another PR based on this one after redesign the code.

PR #5973, opened 2018-01-03
include build time and revision in version information

graphaelli:
I couldn't find a clever way to hook up tests for this, here's one way to verify the changes:
```
$ cd $GOPATH/src/github.com/elastic/beats/auditbeat
$ make -B
./auditbego build -i -ldflags "-X github.com/elastic/beats/libbeat/version.buildTime=2018-01-03T01:07:04 -X github.com/elastic/beats/libbeat/version.commit=65a82579ddc7fa34a82cf554e6ef4adcbf24bdfa"
at version
$ ./auditbeat version
auditbeat version 7.0.0-alpha1 (amd64), libbeat 7.0.0-alpha1 [65a82579ddc7fa34a82cf554e6ef4adcbf24bdfa built 2018-01-03T01:07:04]
```

I got caught up for awhile getting this to work for apps that vendor libbeat, like apm-server.  This worked there, note the `github.com/elastic/apm-server/vendor/` prefix:
```
$ cd $GOPATH/src/github.com/elastic/apm-serve
$ go build -i -ldflags "-X github.com/elastic/apm-server/vendor/github.com/elastic/beats/libbeat/version.buildTime=2018-01-03T01:00:47 -X github.com/elastic/apm-server/vendor/github.com/elastic/beats/libbeat/version.commit=c066b5885efb57decafbee9fbbb532b4e2041529"
$ ./apm-server version
apm-server version 7.0.0-alpha1 (amd64), libbeat 7.0.0-alpha1 [c066b5885efb57decafbee9fbbb532b4e2041529 built 2018-01-03T01:00:47]
```

### DIFF
--- a/libbeat/cmd/version.go
+++ b/libbeat/cmd/version.go
@@ -22,8 +22,13 @@ func genVersionCmd(name, beatVersion string) *cobra.Command {
 					return fmt.Errorf("error initializing beat: %s", err)
 				}
 
-				fmt.Printf("%s version %s (%s), libbeat %s\n",
-					beat.Info.Beat, beat.Info.Version, runtime.GOARCH, version.GetDefaultVersion())
+				buildTime := "unknown"
+				if bt := version.BuildTime(); !bt.IsZero() {
+					buildTime = bt.String()
+				}
+				fmt.Printf("%s version %s (%s), libbeat %s [%s built %s]\n",
+					beat.Info.Beat, beat.Info.Version, runtime.GOARCH, version.GetDefaultVersion(),
+					version.Commit(), buildTime)
 				return nil
 			}),
 	}
--- a/libbeat/scripts/Makefile
+++ b/libbeat/scripts/Makefile
@@ -36,7 +36,8 @@ COVERAGE_TOOL?=${BEAT_GOPATH}/bin/gotestcover
 COVERAGE_TOOL_REPO?=github.com/elastic/beats/vendor/github.com/pierrre/gotestcover
 TESTIFY_TOOL_REPO?=github.com/elastic/beats/vendor/github.com/stretchr/testify
 LIBCOMPOSE_TOOL_REPO?=github.com/docker/libcompose
-GOBUILD_FLAGS?=-i
+NOW=$(shell date -u '+%Y-%m-%dT%H:%M:%SZ')
+GOBUILD_FLAGS?=-i -ldflags "-X github.com/elastic/beats/libbeat/version.buildTime=$(NOW) -X github.com/elastic/beats/libbeat/version.commit=$(COMMIT_ID)"
 GOIMPORTS=goimports
 GOIMPORTS_REPO?=golang.org/x/tools/cmd/goimports
 GOIMPORTS_LOCAL_PREFIX?=github.com/elastic
--- a/libbeat/version/version.go
+++ b/libbeat/version/version.go
@@ -1,3 +1,25 @@
 package version
 
+import "time"
+
 const defaultBeatVersion = "7.0.0-alpha1"
+
+var (
+	buildTime = "unknown"
+	commit    = "unknown"
+)
+
+// BuildTime exposes the compile-time build time information.
+// It will represent the zero time instant if parsing fails.
+func BuildTime() time.Time {
+	t, err := time.Parse(time.RFC3339, buildTime)
+	if err != nil {
+		return time.Time{}
+	}
+	return t
+}
+
+// Commit exposes the compile-time commit hash.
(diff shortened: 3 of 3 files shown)


libbeat/scripts/Makefile
```diff
@@ -36,7 +36,8 @@ COVERAGE_TOOL?=${BEAT_GOPATH}/bin/gotestcover
 COVERAGE_TOOL_REPO?=github.com/elastic/beats/vendor/github.com/pierrre/gotestcover
 TESTIFY_TOOL_REPO?=github.com/elastic/beats/vendor/github.com/stretchr/testify
 LIBCOMPOSE_TOOL_REPO?=github.com/docker/libcompose
-GOBUILD_FLAGS?=-i
+NOW=$(shell date -u '+%Y-%m-%dT%H:%M:%S')
```
andrewkroh: This looks to be very close to RFC3339. Can you add the Zulu timezone identifier (e.g. `date -u +"%Y-%m-%dT%H:%M:%SZ"`). This will make the value parse nicely like `time.Parse(time.RFC3339, "2018-01-03T16:03:28Z")`.

libbeat/version/version.go
```diff
@@ -1,3 +1,18 @@
...
 
 const defaultBeatVersion = "7.0.0-alpha1"
+
+var (
+	buildTime = "unknown"
+	commit    = "unknown"
+)
+
+// BuildTime exposes the compile-time build time information
+func BuildTime() string {
```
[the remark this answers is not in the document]
graphaelli: I like that but concerned about how to handle parse failure, in case of user error.  At the risk of overkill, what about a `versionTime` type with `func String() string` and `func Parse() (time.Time, error)` methods?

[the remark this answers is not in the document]
andrewkroh: I was thinking to only have `version.BuildTime() time.Time`. If the value was not set or was set incorrectly then return `time.Time{}`. For me this would be adequate (I could check for validity if my use case required it with `time.IsZero()`).

But if you want to be more explicit in stating that the build time value is bogus then I'd use `version.BuildTime() (time.Time, error)`.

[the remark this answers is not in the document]
graphaelli: updated as requested, now:

```
# valid time
$ go build -i -ldflags "-X github.com/elastic/beats/libbeat/version.buildTime=2018-01-03T17:30:56Z -X github.com/elastic/beats/libbeat/version.commit=1234"
$ ./libbeat version
mockbeat version 9.9.9 (amd64), libbeat 7.0.0-alpha1 [1234 built 2018-01-03 17:30:56 +0000 UTC]

# invalid time
$ go build -i -ldflags "-X github.com/elastic/beats/libbeat/version.buildTime=XXX -X github.com/elastic/beats/libbeat/version.commit=1234"
$ ./libbeat version
mockbeat version 9.9.9 (amd64), libbeat 7.0.0-alpha1 [1234 built unknown]
```

libbeat/version/version.go
```diff
@@ -1,3 +1,24 @@
...
+import "time"
+
 const defaultBeatVersion = "7.0.0-alpha1"
+
+var (
+	buildTime = "unknown"
+	commit    = "unknown"
+)
+
+// BuildTime exposes the compile-time build time information
```
andrewkroh: You should document the conditions under which a zero value can be returned.

libbeat/version/version.go
```diff
@@ -1,3 +1,24 @@
...
+func BuildTime() time.Time {
+	t, err := time.Parse(time.RFC3339, buildTime)
+	if err != nil {
+		return time.Time{}
+	}
+	return t
+}
+
+// Commit exposes the compile-time commit information
```
andrewkroh: I think this should mention "hash" or "commit ID".

libbeat/cmd/version.go
```diff
@@ -22,8 +22,15 @@ func genVersionCmd(name, beatVersion string) *cobra.Command {
 					return fmt.Errorf("error initializing beat: %s", err)
 				}
 
-				fmt.Printf("%s version %s (%s), libbeat %s\n",
-					beat.Info.Beat, beat.Info.Version, runtime.GOARCH, version.GetDefaultVersion())
+				var buildTime string
```
andrewkroh: Consider initializing `buildTime := "unknown"`. I think it could make this a bit more concise.

andrewkroh:
Regarding testing, we should probably add a check to [beats-tester](https://github.com/elastic/beats-tester) to verify that the .build_hash.txt file matches the `beat version` output.

PR #5974, opened 2018-01-03
index pattern support for binary fields

graphaelli:


libbeat/kibana/transform.go
```diff
@@ -118,6 +118,14 @@ func transformField(version *common.Version, f common.Field) (common.MapStr, com
 		field["type"] = t
 	}
 
+	if f.Type == "binary" {
+		field["aggregatable"] = false
+		field["analyzed"] = false
+		field["indexed"] = false
+		field["doc_values"] = false
```
simitt: I think `doc_values` should be set accordingly to the value in `fields.yml`, but defaulting to `false` for binary, as this is a valid parameter for binary types, according to https://www.elastic.co/guide/en/elasticsearch/reference/current/binary.html#binary-params.

libbeat/kibana/testdata/extensive/fields.yml
```diff
@@ -31,6 +31,11 @@
       description: >
         The timestamp when the event log record was generated.
 
+    - name: blob
+      type: binary
+      description: >
+        Some binary blob.
```
simitt: I recommend to change the `fields.yml` file rather than the `extensive/fields.yml` as it is easier to grasp. 
The extensive tests were added to ensure that no breaking change was introduced when rewriting the code in go and it reflects the complete fields.yml of `metricbeat` at that time.

libbeat/kibana/transform.go
simitt: Could you please add tests for this to ensure values are set properly https://github.com/elastic/beats/blob/master/libbeat/kibana/transform_test.go#L188

graphaelli:
@simitt thanks for the thorough review, all changes implemented

graphaelli:
did as instructed, has been sitting in ci queue for the last 2 hours

ruflin:
@graphaelli Thanks. Merged. We need more OS X machines for our builds ...

PR #5979, opened 2018-01-04
Update golang.org/x/sys/unix packaget to most recent version

ruflin:
This should solve a conflict with the golang.org/x/crypto/ssh/terminal terminal package in apm.

@graphaelli Can you check if this solves the issue?


graphaelli:
@ruflin - It turns out govendor is not automatically updating beats' deps when we `update-beats`, but rather only apms' immediate deps.  APM can resolve this on our side by forcing the update there.  Closing this but of course feel free to reopen if beats needs it, we'll pull it in after.

ruflin:
@andrewkroh WDYT, should we get this update in?

PR #5986, opened 2018-01-04

tsg:
There were no conflicts besides the versions.


ruflin:
I think the module tests are failing because ES 6.2 has a different version of the geoip lookup library then 7.0. For some of the IP addresses these return geo locations. I would suggest we merge in the changes and do a follow up PR to adjust the locations?

ruflin:
Merged, will do a follow up PR with the fix.

PR #5990, opened 2018-01-04
[DOCS] Changed certgen to certutil

lcawl:
The certgen tool is deprecated, so the Beats reference books should refer to "certutil" instead in 6.1 and later.


ruflin:
I added the `needs_backport` label to make sure we get it into 6.1 and 6.x.

tvernum:
LGTM

lcawl:
Backported to 6.x and 6.1

Issue #5984, opened 2018-01-04
Reintroduce crosscompile on tested platforms


kvch:
Cross-compilation Travis jobs is removed for a time being in #5545. Right now CGO is not enabled for these tests. Introduction of CGO, a bigger change is required in setting up test environments. To avoid blocking the merge of CPU usage, this task is moved to a different issue.

Most of the temporarily removed platforms are not supported officially. It might interfere with compiling master branch of Beats on these platforms, as a new release of `gosigar` is required and must be move to `vendor`.

The tests can be still run locally using `make crosscompile`.

ruflin:
As part of this we should add crosscompile builds also to jenkins.

It also reminds of this old issue here: https://github.com/elastic/beats/issues/2251

kvch:
Resolved in #5993

PR #5993, opened 2018-01-05
Readd crosscompile builds

ruflin:
In #5545 the crosscompile builds were disable because they didn't work with the changes. This PR reenables the crosscompiling by making the following changes:

* Introduce build flags for the metrics. Currently only build on darwin, linux, windows
* Readd build flags to cpu and memory. They were removed in the previous PR but not sure why.
* Uncomment crosscompile builds for travis

This only brings back the previous behaviour. Additional changes should be made to support monitoring also on missing platforms.

See #5984

### DIFF
--- a/libbeat/metric/system/memory/memory.go
+++ b/libbeat/metric/system/memory/memory.go
@@ -1,3 +1,5 @@
+// +build darwin freebsd linux openbsd windows
+
 package memory
 
 import (
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,11 +63,10 @@ jobs:
       env: TARGETS="-C libbeat testsuite"
       go: $GO_VERSION
       stage: test
-      # Skip tests: see https://github.com/elastic/beats/issues/5984
-      #    - os: linux
-      #      env: TARGETS="-C libbeat crosscompile"
-      #      go: $GO_VERSION
-      #      stage: test
+    - os: linux
+      env: TARGETS="-C libbeat crosscompile"
+      go: $GO_VERSION
+      stage: test
     - os: linux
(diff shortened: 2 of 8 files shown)


libbeat/metric/system/memory/memory.go
```diff
@@ -1,3 +1,5 @@
+// +build darwin freebsd linux openbsd windows
```
ruflin: @kvch I readded these build tags as I'm not sure why they were removed in the previous PR? The part I didn't test is if the flags are still all correct as there are is potentially new functionality. Was there a reason you removed these?

kvch: I think I removed it, because at some point netbsd and solaris did not compile. As I see, now it works. It might got lost in the ocean of patches I produced. Thanks for adding it again.

kvch:
For the record, my original intention was to return with `ErrNotImplemented` from gosigar and handle all platform specific codes there, because it is the library which suppose to interact with the OS. So Beats would not need to care about platforms at all.
Now it feels to me that we still divide the responsibility of interacting and "caring" about the platform between gosigar and Beats. But in an ideal world in my mind handling OS specific code is done by gosigar where possible. I do understand it is not an ideal world and our models are not perfect. :)

Your solution is acceptable to me. It is a small change and easy to understand.

ruflin:
I'm on the fence between the two option. I think beats should be able to guarantee on compile time the the monitoring for our supported platforms work. If we delegate everything to gosigar we only know at run time if it works, which means we would need tests to be run on all our platform to check for the data to be there.

@andrewkroh I wonder how this affects go-sysinfo?

PR #5994, opened 2018-01-05
[DOCS] Add X-Pack security information to Beats reference books

lcawl:
This PR adds the information that was buried in the X-Pack Reference (https://www.elastic.co/guide/en/x-pack/master/beats.html) to the appropriate Beats books. 

It will be removed from the X-Pack Reference in a separate PR.

NOTE: The addition of the "branch" attribute in the version.asciidoc is to fix links that use URLs from the elastic/docs/shared/attributes.asciidoc file.


libbeat/docs/version.asciidoc
```diff
@@ -1,7 +1,8 @@
-:stack-version: 7.0.0-alpha1
-:doc-branch: master
-:go-version: 1.9.2
-:release-state: unreleased
-:python: 2.7.9
-:docker: 1.12
-:docker-compose: 1.11
+:stack-version:   7.0.0-alpha1
```
ruflin: As we use these variables in other places (yes, we shouldn't) I wonder if this could cause some issues with spaces in front.

@exekias These are used in the k8s configs too.

exekias: Yes it seems there is an issue with the extra spaces https://travis-ci.org/elastic/beats/jobs/325272871#L717, @lcawl could you remove them?

lcawl:
Thanks for all the feedback, @dedemorton!  I've implemented those changes.

PR #5996, opened 2018-01-05
Metricbeat: Update MySQL dashboard to use TSVB

tsg:
The MySQL dashbaord was a combination of timelion and Vizualise. This updates
to use TSVB for all visualsations.


tsg:
Added `in progress` because I need to test this with Kibana 6.2 before merging.

tsg:
Tested with Kibana 6.1. Should be ready to merge now.

Issue #5275, opened 2017-09-29
Metricbeat Dashboard causes query exception in master/7.0


beniwohli:
With Elasticsearch/Kibana on master, when running metricbeat with the system module active, opening the Host Overview dashboard leads to an error

   "reason":"field expansion matches too many fields, limit: 1024, got: 1030"

The full error, as well as the metricbeat configuration I used can be found here: https://gist.github.com/beniwohli/bb1bfe63ed02480d08025a444d157376

ruflin:
@simianhacker I assume these are the same issues? https://github.com/elastic/beats/issues/5278

simianhacker:
@ruflin I don't think so... I get that error when searching in Discover as well.

simianhacker:
@elastic/beats This is related to https://github.com/elastic/elasticsearch/pull/26541 which is going to be a problem for Metricbeat.

Bargs:
I just caught up on some discussion of this in Slack from last week. I just wanted to note that this will still affect users when they switch to kuery since the limit also applies to multi_match queries, which kuery uses under the hood.

simianhacker:
So the fix we decided on last week is to force the admin to add a list of 1024 fields to the index setting `index.query.default_field`. Since we are administering the `metrcibeat-*` indices we will be responsible for changing our index template to include `index.query.default_field`.

Bargs:
That won't work for kuery since `multi_match` doesn't use `default_field`. But if I understood the conversation, what you described is meant to be a simple fix for the short term, so that's probably fine.

simianhacker:
@Bargs Yeah... that's going to be a problem for multi_match. What do you guys put in the fields array now just `*`? It would suck to ask the user to fill that in for each index pattern that has more the 1024 searchable fields.

Bargs:
Yep, just `*`. I asked if the ES guys have any advice in the [other issue](https://github.com/elastic/elasticsearch/pull/26541#issuecomment-340608785).

rasroh:
We hit the same issue as above while running the integration tests on master . 
```
org.elasticsearch.transport.RemoteTransportException: [cQvsA1-][10.0.2.15:9300][indices:data/read/search[phase/query]]
Caused by: org.elasticsearch.index.query.QueryShardException: failed to create query: {
  "bool" : {
    "must" : [
      {
        "query_string" : {
          "query" : "*",
          "default_field" : "*",
          "fields" : [ ],
          "type" : "best_fields",
[...]

PR #5997, opened 2018-01-05
Reduce mapping size by relying more on dynamic typing

tsg:
Opening this for discussion. I hit the Metricbeat query exception when running against Kibana master (https://github.com/elastic/beats/issues/5275), and as a workaround I used this patch to use dynamic mappings for keywords and logs. The code change is trivial (see diff) and solved the issues while I didn't notice any other regression in my unscientific testing. This might also solve the slowdown on the Discover page that was experienced by some users.

The patch causes a reduction of the template from 7K lines to 1k lines.

What I'd like to discuss:

* Are there objections with implementing this in 6.x, with potentially reverting it for 7.0 when we have a better solution (e.g. index per module)?
* What tests should we make to make sure this doesn't break anything?
* How could we check that the slowdown in Discovery gets improved.

### DIFF
--- a/libbeat/template/processor.go
+++ b/libbeat/template/processor.go
@@ -34,12 +34,14 @@ func (p *Processor) process(fields common.Fields, path string, output common.Map
 			mapping = p.scaledFloat(&field)
 		case "half_float":
 			mapping = p.halfFloat(&field)
-		case "integer":
-			mapping = p.integer(&field)
+		case "integer", "long":
+			// don't explicitly include the integers, ES automatically detects them
+			//mapping = p.integer(&field)
 		case "text":
 			mapping = p.text(&field)
 		case "", "keyword":
-			mapping = p.keyword(&field)
+			// don't explicitly include the keywords, let the dynamic rule handle them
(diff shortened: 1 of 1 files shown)


urso:
From diff it's not clear how numbers will be handled. Will numbers be `long` or `integer` by default? As `integer` and `long` are different in number of bits used, this seem to be a breaking change.

ruflin:
To better play around with this I suggest we introduce it as a config option like `setup.template.dynamie: true` (better name needed). So in 6.2 we could introduce this but have it set to "false" by default as we have it now. In 6.3 we could put it into our `short config` so new users automatically start using it but internally the default stays the same. As I think it does not fully count as a breaking change I think that's ok.

@urso We already store all integers as `long` in Elasticsearch. There were some tests in the past which showed there are no downsides of using long for all values.

ruflin:
@tsg Stumbled over this again. I think it's still worth investigating / testing. Let's introduce a config option to play around with it.

tsg:
Closing as we have a new plan.

PR #5999, opened 2018-01-05
Switch packer to sha512

tsg:
The unified build process doesn't use these files, but updating them
from sha1 to sha512 to simplify the life of our own Jenkins testing.


tsg:
Done.

PR #6006, opened 2018-01-06
[Metricbeat] new uWSGI module

iahmedov:
This adds uWSGI server monitoring metrics.

- [x] docs asciidoc
- [x] fields
- [x] dashboard
- [x] unit tests
- [x] integration tests
- [x] system tests

uwsgi_dashboard


metricbeat/module/uwsgi/_meta/config.yml
```diff
@@ -0,0 +1,7 @@
+- module: uwsgi
+  metricsets: ["status"]
+  enabled: false
```
exekias: Could you remove `enabled` here? we use the default now (true), as this is handled by `modules` subcommand

metricbeat/module/uwsgi/_meta/config.yml
```diff
@@ -0,0 +1,7 @@
...
+  period: 10s
+  hosts: ["tcp://127.0.0.1:9191"]
+
+
```
exekias: some spurious empty lines here

metricbeat/module/uwsgi/status/_meta/fields.yml
```diff
@@ -0,0 +1,122 @@
...
+        worker ID
+    - name: core.worker_pid
+      type: long
+      description: >
+        Parent worker PID
+    - name: core.requests
+      type: long
+      description: >
+        Number of requests served
+    - name: core.static_requests
```
exekias: Sounds like you cold group requests like: `requests.total`, `requests.static`, `requests.routed`, etc...

metricbeat/module/uwsgi/status/status.go
```diff
@@ -0,0 +1,95 @@
...
+	mb.Registry.AddMetricSet("uwsgi", "status", New, uwsgi.HostParser)
+}
+
+// MetricSet for fetching uwsgi metrics from StatServer.
+type MetricSet struct {
+	mb.BaseMetricSet
+}
+
+// New creates a new instance of the MetricSet.
+func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
```
exekias: Could you add an Experimental warning here? Something like:
```
cfgwarn.Experimental("The uWSGI status metricset is experimental")
```

metricbeat/module/uwsgi/status/_meta/fields.yml
exekias: @ruflin may have some more comments here :)

exekias:
jenkins, test it

iahmedov:
Thanks @exekias for review, made requested changes

metricbeat/module/uwsgi/status/_meta/data.json
```diff
@@ -0,0 +1,53 @@
...
+    "uwsgi": {
+      "status": {
+        "total": {
+          "requests": 102,
+          "exceptions": 1,
+          "write_errors": 0,
+          "read_errors": 0,
+          "pid": 1
+        },
+
```
ruflin: Did you generate this file or add it manually? I kind of wonder how this newline here happened.

metricbeat/module/uwsgi/status/_meta/fields.yml
```diff
@@ -0,0 +1,121 @@
+- name: status
+  type: group
+  description: >
+    uwsgi.status metricset fields
+  fields:
+    - name: total.requests
```
ruflin: I would probably turn around the namespacing as I think`requests` is the namespace and `total` only describes which values it is in the `requests` namespace. Same below like:


`errors.write.total`
`errors.read.total`

and others.

ruflin:
jenkins, test it

metricbeat/module/uwsgi/status/_meta/data.json
iahmedov: I made this file manually, actually uwsgi.status contains only one of the nested fields (total, worker or core)

metricbeat/module/uwsgi/status/_meta/fields.yml
iahmedov: `total` was kind of summary for all workers, it has `requests, exceptions, read, write errors`, if I turnaround total.requests, can you suggest naming for summary data?

current structure is as below
```
"total": {
  "requests": 102,
  "exceptions": 1,
  "write_errors": 0,
  "read_errors": 0,
  "pid": 1
}
```

turned around might look like this
```
"summary": {
  "requests": {
     "total": 102
  },
  "exceptions": 1,
  "errors": {
    "write": {
       "total" : 0
    },
    "read": {
       "total" : 0
    },
  }
  "pid": 1
}
```

feels like to much overhead, but if it is considered as more readable I welcome the changes.
I appreciate if you can suggest better structure for summary data.

exekias:
jenkins, test it please

exekias:
Merging it now, thank you so much for this awesome module! As @ruflin said, we can open follow up PRs to tweak field names

metricbeat/module/uwsgi/status/_meta/fields.yml
ruflin: In general my approach is going from the generic one to the more detailed. So in the above case I think a user might first search for an `error` and then he can decide if writes or reads. I would assume he does not look at the document and think "I need a total" and then check what is inside total. But I'm pretty sure it could be argued the other way around :-)

metricbeat/module/uwsgi/status/_meta/data.json
ruflin: Normally we generate these fields automatically with one of the tests like here for example: https://github.com/elastic/beats/blob/master/metricbeat/module/kibana/status/status_integration_test.go#L27 Like this we assure that the file stays up-to-date if we change mappings. Perhaps you could open a follow up PR with a file that is auto generated?

PR #6008, opened 2018-01-07
Check float metrics against gauges list

andrewkroh:
Float metrics were not checked against the list of gauge values so the CPU and Load metrics were logged incorrectly. By default the difference between the current and last value are reported in the logs. But if a value is whitelisted as a gauge then it reports the raw value instead. Only the int values were being checked. Now the float64 values are checked too.


andrewkroh:
This should fix the negative load values I was seeing in my logs.

```
      "system": {
        "cpu": {
          "total": {
            "norm": {
              "pct": 0.027100000000000013
            },
            "pct": 0.2168
          }
        },
        "load": {
          "1": -0.14749999999999996,
          "norm": {
            "1": -0.0184
          }
        }
      }
```

PR #6010, opened 2018-01-08
Add process.GetOne to get one process by PID

kvch:
The PR removes the unnecessary debug messages when the stats of the Beat process is queried.

@roncohen This hopefully eliminates the flood of debug messages before monitoring events.

### DIFF
--- a/libbeat/metric/system/process/process.go
+++ b/libbeat/metric/system/process/process.go
@@ -352,7 +352,7 @@ func (procStats *Stats) Init() error {
 	return nil
 }
 
-// Get fetches process data which matche the provided regexes from the host.
+// Get fetches process data which matches the provided regexes from the host.
 func (procStats *Stats) Get() ([]common.MapStr, error) {
 	if len(procStats.Procs) == 0 {
 		return nil, nil
@@ -367,33 +367,11 @@ func (procStats *Stats) Get() ([]common.MapStr, error) {
 	newProcs := make(ProcsMap, len(pids))
 
 	for _, pid := range pids {
-		var cmdline string
-		var env common.MapStr
-		if previousProc := procStats.ProcsMap[pid]; previousProc != nil {
-			if procStats.CacheCmdLine {
-				cmdline = previousProc.CmdLine
-			}
-			env = previousProc.Env
-		}
-
-		process, err := newProcess(pid, cmdline, env)
-		if err != nil {
-			logp.Debug("processes", "Skip process pid=%d: %v", pid, err)
+		process := procStats.getSingleProcess(pid, newProcs)
+		if process == nil {
 			continue
 		}
-
-		if procStats.matchProcess(process.Name) {
-			err = process.getDetails(procStats.isWhitelistedEnvVar)
-			if err != nil {
-				logp.Err("Error getting process details. pid=%d: %v", process.Pid, err)
-				continue
-			}
-
-			newProcs[process.Pid] = process
-			last := procStats.ProcsMap[process.Pid]
-			process.cpuTotalPctNorm, process.cpuTotalPct, process.cpuSinceStart = GetProcCPUPercentage(last, process)
-			processes = append(processes, *process)
-		}
+		processes = append(processes, *process)
 	}
 	procStats.ProcsMap = newProcs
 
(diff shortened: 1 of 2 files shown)


libbeat/metric/system/process/process.go
```diff
@@ -409,6 +388,53 @@ func (procStats *Stats) Get() ([]common.MapStr, error) {
...
+	if previousProc := procStats.ProcsMap[pid]; previousProc != nil {
+		if procStats.CacheCmdLine {
+			cmdline = previousProc.CmdLine
+		}
+		env = previousProc.Env
+	}
+
+	process, err := newProcess(pid, cmdline, env)
+	if err != nil {
+		return nil, fmt.Errorf("Skip process pid=%d: %v", pid, err)
```
ph: Error string should not be capitalized, https://github.com/golang/go/wiki/CodeReviewComments#error-strings

ph:
@kvch I am a bit confused by the description of this PR, I was looking at the code, when you say `This hopefully eliminates the flood of debug messages before monitoring events`, Do you mean that you change the log level from `Err` to `debug` for this line? https://github.com/elastic/beats/pull/6010/files#diff-fa9c1f1c480f569e22179691de27404dL388

kvch:
The problem was that if you start a Beat as root using `filebeat -e -d "*"` the debug logs are flooded with these messages every 30 seconds before a monitoring event is created:
```
2018-01-08T22:40:09.884+0100    DEBUG   [processes]     process/process.go:381  Skip process pid=3: error getting process exe for pid=3: readlink /proc/3/exe: no such file or directory
2018-01-08T22:40:09.884+0100    DEBUG   [processes]     process/process.go:381  Skip process pid=5: error getting process exe for pid=5: readlink /proc/5/exe: no such file or directory
2018-01-08T22:40:09.885+0100    DEBUG   [processes]     process/process.go:381  Skip process pid=7: error getting process exe for pid=7: readlink /proc/7/exe: no such file or directory
2018-01-08T22:40:09.885+0100    DEBUG   [processes]     process/process.go:381  Skip process pid=8: error getting process exe for pid=8: readlink /proc/8/exe: no such file or directory
2018-01-08T22:40:09.885+0100    DEBUG   [processes]     process/process.go:381  Skip process pid=9: error getting process exe for pid=9: readlink /proc/9/exe: no such file or directory
2018-01-08T22:40:09.885+0100    DEBUG   [processes]     process/process.go:381  Skip process pid=10: error getting process exe for pid=10: readlink /proc/10/exe: no such file or directory
2018-01-08T22:40:09.885+0100    DEBUG   [processes]     process/process.go:381  Skip process pid=11: error getting process exe for pid=11: readlink /proc/11/exe: no such file or directory
2018-01-08T22:40:09.886+0100    DEBUG   [processes]     process/process.go:381  Skip process pid=12: error getting process exe for pid=12: readlink /proc/12/exe: no such file or directory
2018-01-08T22:40:09.886+0100    DEBUG   [processes]     process/process.go:381  Skip process pid=13: error getting process exe for pid=13: readlink /proc/13/exe: no such file or directory
2018-01-08T22:40:09.886+0100    DEBUG   [processes]     process/process.go:381  Skip process pid=14: error getting process exe for pid=14: readlink /proc/14/exe: no such file or directory
```
It happens because the current `process.Get` iterates over all PIDs of the current processes, but it cannot access everything. As only the Beat process is interesting to us, there is no need to try to create a `Process` for every PID on the host. I added a new function which gets a PID and fetches process data for that PID. This way, there is no unnecessary iterating over list of PIDs.
I simply pass the PID of the running Beat process and get data for only it. So these messages never show up.
I extracted it into a smaller function so getting a data for a PID can be used from both `process.Get` and `process.GetOne`. I tried to preserve the format and the content of the error messages. But I will change it.

kvch:
It seems to me that every log stayed at the same level.

ph:
@kvch Thanks for the clarificatios, code looks good I will test tomorrow.

libbeat/metric/system/process/process.go
```diff
@@ -409,6 +388,53 @@ func (procStats *Stats) Get() ([]common.MapStr, error) {
...
+	if procStats.matchProcess(process.Name) {
+		err = process.getDetails(procStats.isWhitelistedEnvVar)
+		if err != nil {
+			return nil, fmt.Errorf("error getting process details. pid=%d: %v", process.Pid, err)
+		}
+
+		newProcs[process.Pid] = process
+		last := procStats.ProcsMap[process.Pid]
+		process.cpuTotalPctNorm, process.cpuTotalPct, process.cpuSinceStart = GetProcCPUPercentage(last, process)
+	} else {
```
ruflin: Could we change the code to `if !procStats.matchProcess(process.Name) { return nil, err... }` and not have an else so the code above would be in the normal flow?

ruflin:
@kvch Thanks for the details in https://github.com/elastic/beats/pull/6010#issuecomment-356109858 Could you update the PR / commit message with something similar. That makes it much easier to understand what changed.

kvch:
I changed the code as you requested.
I also moved the logging back to `getSingleProcess`, because there were two debug logs and one error logs, so I accidentally transformed one error log into a debug log. Now it is fixed.

PR #6018, opened 2018-01-08
Update logging from auditd module

andrewkroh:
Use `logp.Logger` for all logging from the auditd module. I also increased the logging level for two statements because they will be useful for troubleshooting (without having to ask users to re-run with debug enabled).

### DIFF
--- a/auditbeat/module/auditd/audit_linux.go
+++ b/auditbeat/module/auditd/audit_linux.go
@@ -21,15 +21,11 @@ import (
 )
 
 const (
-	logPrefix = "[" + moduleName + "]"
-
 	// Use old namespace for data until we do some field renaming for GA.
 	namespace = "audit.kernel"
 )
 
 var (
-	debugf = logp.MakeDebug(moduleName)
-
 	auditdMetrics = monitoring.Default.NewRegistry(moduleName)
 	lostMetric    = monitoring.NewInt(auditdMetrics, "lost")
 )
@@ -50,6 +46,7 @@ type MetricSet struct {
 	mb.BaseMetricSet
 	config Config
 	client *libaudit.AuditClient
+	log    *logp.Logger
 }
 
 // New constructs a new MetricSet.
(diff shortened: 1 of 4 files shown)


auditbeat/module/auditd/audit_linux.go
```diff
@@ -116,14 +115,14 @@ func (ms *MetricSet) Run(reporter mb.PushReporterV2) {
...
-		logp.Err("%v %v", logPrefix, err)
+		ms.log.Errorw("Failure adding audit rules", "error", err)
 		return
 	}
 
 	out, err := ms.receiveEvents(reporter.Done())
 	if err != nil {
 		reporter.Error(err)
-		logp.Err("%v %v", logPrefix, err)
+		ms.log.Errorw("Failure receiving audit events", "error", err)
```
ruflin: Would be nice if this would be `error.message` the same as we have in our `fields.yml`.

metricbeat/mb/testing/modules.go
```diff
@@ -269,7 +269,10 @@ func RunPushMetricSetV2(timeout time.Duration, waitEvents int, metricSet mb.Push
 			select {
 			case <-timer.C:
 				return
-			case e := <-r.eventsC:
+			case e, ok := <-r.eventsC:
```
ruflin: @andrewkroh Is this change related? LGTM but curious if that is related to an other issue.

andrewkroh: While testing some failure conditions I noticed the test case didn’t stop immediately and this was the fix.

auditbeat/module/auditd/audit_linux.go
andrewkroh: It would. The error is written as an object, but it doesn’t match our normal structure. See https://github.com/uber-go/zap/blob/master/zapcore/error.go#L28

andrewkroh: It should be possible to intercept (but with some expense) the Error types and wrap them in another type that implements https://godoc.org/go.uber.org/zap/zapcore#ObjectMarshaler to customize the marshaling.

ruflin: The part I'm worried if we don't do it is that someone is going to read filebeat json logs which will conflict with our own filebeat template.

PR #6022, opened 2018-01-09
Fix attribute order

dedemorton:
Well, I goofed. The include statement for the shared attributes needs to appear after the file that defines `beatname_l`c because the `dockerimage` attribute has a dependency on `beatname_lc`. 

Because the doc build didn't complain, I didn't realize there was an issue until I saw an extra space in some of the docker commands. The commands had a space *instead* of the docker image name. 

This issue only exists in master because the PR that cherry-picks these changes is not merged yet.


ruflin:
Merged as doc build was green.

Issue #6019, opened 2018-01-08
Check if audit config is locked


andrewkroh:
If the audit config is locked (audit_status.Enabled == 2) then the auditd metricset should either not attempt specific config changes or should continue on `EPERM` errors. The goal is to allow Auditbeat to read events even if the config has been locked in the kernel.

- https://github.com/linux-audit/audit-kernel/blob/bebc6082da0a9f5d47a1ea2edc099bf671058bd4/kernel/audit.c#L379

adriansr:
After playing a little bit more with audit and locking, I'm afraid what you're trying to do here is not possible.

Once the config has been locked, not only rules cannot be added, but also the status fields cannot be changed. This includes the PID of the active auditing process. In this case, the PID is either that of the audit process that was active at the time of locking or zero if no audit process has been set, or it terminated already. The only way to unlock the configuration and register a new PID is to restart the system. 

This rules out all auditing using the `unicast` socket, as Auditbeat cannot register its PID.

The good news is, on the multicast side, no changes are necessary to the current Auditbeat to be able to receive audit events when the config is locked. Even if the audit process has terminated (PID=0), events are still received.

So my suggestion is either transition to `multicast` when the configuration is locked or fail and suggest the operator use a multicast socket.

andrewkroh:
I think falling back to the multicast socket would be good (and logging a warning about why). I'm surprised that the kernel doesn't allow the PID to be set since this prevents the audit daemon from restarting.

PR #6023, opened 2018-01-09
auditbeat: Warn if auditd is running

### LINKED ISSUE #5845: Auditbeat: Warn if Auditbeat and auditd are running
When using a unicast netlink socket you can have only one userspace process registered to receive data from the kernel audit framework. If two are running only one will receive data and the other will get nothing and no errors.

Auditbeat could be enhanced to make it more user friendly by detecting this condition and warning the user by logging a message and sending an error event to ES.

Detection could be accomplished by:

- Periodically checking if the Auditbeat PID is registered as the recipient of Audit data in the kernel. We can check the [audit status](https://github.com/elastic/go-libaudit/blob/7fd432dba2fe716f4ead79ac3a57cae7b5c1b422/audit.go#L490) and if our PID isn't present warn and try to fix the issue.
- Dump the process table and look for "auditd".

adriansr:
Detect initialization failures for the auditd module:
- When another process is already set as the audit process (auditd already running). This error is reported.
- If rules are locked, prints a warning message and does not attempt to add its own rules (This implies auditd is running).

Closes #5845 and #6019

### DIFF
--- a/auditbeat/module/auditd/audit_linux.go
+++ b/auditbeat/module/auditd/audit_linux.go
@@ -24,6 +24,8 @@ import (
 const (
 	// Use old namespace for data until we do some field renaming for GA.
 	namespace = "audit.kernel"
+
+	auditLocked = 2
 )
 
 var (
@@ -153,6 +155,17 @@ func (ms *MetricSet) addRules(reporter mb.PushReporterV2) error {
 		return errors.Wrap(err, "failed to create audit client for adding rules")
 	}
 	defer client.Close()
+
+	// Don't attempt to change configuration if audit rules are locked (enabled == 2).
+	// Will result in EPERM.
+	status, err := client.GetStatus()
+	if err != nil {
+		return errors.Wrap(err, "failed to get audit status before adding rules")
+	}
+	if status.Enabled == auditLocked {
+		ms.log.Warn("Skipping rule configuration: Audit rules are locked")
+		return nil
+	}
 
 	// Delete existing rules.
 	n, err := client.DeleteRules()
--- a/auditbeat/module/auditd/audit_linux.go
+++ b/auditbeat/module/auditd/audit_linux.go
@@ -26,6 +26,9 @@ const (
 	namespace = "audit.kernel"
 
 	auditLocked = 2
+
+	unicast   = "unicast"
+	multicast = "multicast"
 )
 
 var (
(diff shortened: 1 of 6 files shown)


ruflin:
@adriansr Seems to have a conflict.

auditbeat/module/auditd/audit_linux.go
```diff
@@ -153,6 +156,17 @@ func (ms *MetricSet) addRules(reporter mb.PushReporterV2) error {
...
 	defer client.Close()
 
+	// Don't attempt to change configuration if audit rules are locked (enabled == 2).
+	// Will result in EPERM.
+	status, err := client.GetStatus()
+	if err != nil {
+		return errors.Wrap(err, "failed to get audit status before adding rules")
+	}
+	if status.Enabled == auditLocked {
+		ms.log.Warn("Skipping rule configuration: Audit rules are locked")
```
andrewkroh: Can you push the error to the `reporter` for this case (like we do at https://github.com/elastic/beats/pull/6023/files#diff-a3559204e3ac05aabc310b7b3dfd90e3R120). As an operator I'd like to see this type of information reported in Elasticsearch.

auditbeat/module/auditd/audit_linux.go
```diff
@@ -217,11 +231,15 @@ func (ms *MetricSet) initClient() error {
 			return errors.Wrap(err, "failed to enable auditing in the kernel")
 		}
 	}
-
-	if err := ms.client.SetPID(libaudit.NoWait); err != nil {
-		return errors.Wrap(err, "failed to set audit PID")
+	if err := ms.client.WaitForPendingACKs(); err != nil {
```
andrewkroh: Thought of one more thing. 😄
If the audit config is locked, will any of the above config changes result in a failure?

adriansr: Yes, currently this is failing either in this line or the SetPID below.

I think we need to handle configuration locking differently as it will only work with a multicast socket. See [this comment](https://github.com/elastic/beats/issues/6019#issuecomment-356524906) I left on the issue.

auditbeat/module/auditd/audit_linux.go
```diff
@@ -454,3 +464,80 @@ func kernelVersion() (major, minor int, full string, err error) {
...
+						"will be used instead, which does not support setting rules")
+					return multicast, nil
+				}
+				return unicast, nil
+			}
+			return multicast, nil
+		}
+		if isLocked {
+			log.Errorf("Cannot continue: audit configuration is locked" +
+				" in the kernel (enabled=2) which prevents using unicast " +
```
andrewkroh: Nit - Move the space to the previous line for consistency?

auditbeat/module/auditd/audit_linux.go
```diff
@@ -454,3 +464,80 @@ func kernelVersion() (major, minor int, full string, err error) {
...
+		}
+		return c.SocketType, nil
+	}
+	rules, _ := c.rules()
+
+	isLocked := status.Enabled == auditLocked
+	hasMulticast := hasMulticastSupport()
+	hasRules := len(rules) > 0
+
+	const useAutodetect = "Remove the socket_type option to have auditbeat " +
```
andrewkroh: I like the logic change here. It's deterministic now if the user specifies a `socket_type`.

auditbeat/module/auditd/audit_linux.go
```diff
@@ -153,6 +142,19 @@ func (ms *MetricSet) addRules(reporter mb.PushReporterV2) error {
...
+	// Don't attempt to change configuration if audit rules are locked (enabled == 2).
+	// Will result in EPERM.
+	status, err := client.GetStatus()
+	if err != nil {
+		err = errors.Wrap(err, "failed to get audit status before adding rules")
+		reporter.Error(err)
+		return err
+	}
+	if status.Enabled == auditLocked {
+		reporter.Error(errors.New("Skipping rule configuration: Audit rules are locked"))
```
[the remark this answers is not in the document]
adriansr: Nope. I wasn't sure which one did you want to be reported in your first comment.

PR #6024, opened 2018-01-09
Add support for setting fields in the config file

ruflin:
So far if the user wanted to modify the generated template, he either had to create his own template.json file or modify our `fields.yml` file. The problem with changing the `fields.yml` file is that with new versions it was hard to keep these up-to-date. This change allows to specify the few fields which should be set as part of the config.

Setting fields is especially useful in the case of Metricbeat for modules like Http or Prometheus where the data is user specific and we don't know the structure in advance.

This change also has affects on the generation of the index pattern in Kibana.

The configuration looks as following:

```
setup.template.append_fields:
  - name: test.name
    type: keyword
  - name: test.hostname
    type: long
```

I would have preferred to use `setup.template.fields:` but that is already taken by the path to the file.

Notes:
* For this change to happen the template and index pattern must be overwritten
* Overwriting existing fields is not allowed

### DIFF
--- a/libbeat/common/field.go
+++ b/libbeat/common/field.go
@@ -99,6 +99,44 @@ func (f Fields) HasKey(key string) bool {
 	return f.hasKey(keys)
 }
 
+// HasNode checks if inside fields the given node exists
+// In contrast to HasKey it not only compares the leaf nodes but
+// every single key it traverses.
+func (f Fields) HasNode(key string) bool {
+	keys := strings.Split(key, ".")
+	return f.hasNode(keys)
+}
+
+func (f Fields) hasNode(keys []string) bool {
+
+	// Nothing to compare, so does not contain it
+	if len(keys) == 0 {
+		return false
+	}
+
+	key := keys[0]
+	keys = keys[1:]
+
+	for _, field := range f {
+
+		if field.Name == key {
+
+			//// It's the last key to compare
+			if len(keys) == 0 {
+				return true
+			}
+
+			// It's the last field to compare
+			if len(field.Fields) == 0 {
+				return true
+			}
+
+			return field.Fields.hasNode(keys)
+		}
+	}
+	return false
+}
+
 // Recursively generates the correct key based on the dots
 // The mapping requires "properties" between each layer. This is added here.
 func GenerateKey(key string) string {
@@ -133,4 +171,28 @@ func (f Fields) hasKey(keys []string) bool {
 		}
 	}
 	return false
+}
+
+// GetKeys returns a flat list of keys this Fields contains
+func (f Fields) GetKeys() []string {
+	return f.getKeys("")
+}
+
+func (f Fields) getKeys(namespace string) []string {
+
+	var keys []string
+
+	for _, field := range f {
+		fieldName := namespace + "." + field.Name
+		if namespace == "" {
+			fieldName = field.Name
+		}
+		if len(field.Fields) == 0 {
+			keys = append(keys, fieldName)
+		} else {
+			keys = append(keys, field.Fields.getKeys(fieldName)...)
+		}
+	}
+
+	return keys
 }
--- a/libbeat/docs/template-config.asciidoc
+++ b/libbeat/docs/template-config.asciidoc
@@ -83,6 +83,6 @@ setup.template.settings:
   _source.enabled: false
 ----------------------------------------------------------------------
 
-*`setup.template.append_fields`*:: A list of of fields to be added to the template and Kibana index pattern.
+*`setup.template.append_fields`*:: A list of of fields to be added to the template and Kibana index pattern. experimental[]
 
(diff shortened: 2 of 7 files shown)


exekias:
I like the idea :taco:, it will enable users to map their own stuff, which is nice. I would call it `extra_fields`, to make it clear that it's adding fields on top of the default ones.

I'm worried about allowing overwrites, I cannot think of any case that doesn't break our events. I would say you can override fully y replacing `fields.yaml` if that's what you want. That will ensure someone adding just some fields cannot break anything.

tsg:
Two thoughts on this:

* It might be nicer to bring the configuration at the module level. That way, we can in the future support simple customisations to Filebeat modules by modifying only the configuration file. But that can come later.
* Just adding those fields won't be enough, because the template needs to be overwritten, which means that the overwrite flag needs to be set to true at least once, and a new index needs to be created.

I was working/thinking of a similar development for the Filebeat modules pipelines. In that case, as a solution to the issue above, we could allow the user to provide a custom string to add to the pipeline ID. So something like this:

```
  pipeline:
    append_processors:
      - rename: ...
    custom_token: mypipeline
```

Then the pipeline ID created would be of the form: `filebeat-<module>-<version>-mypipeline` which means it would be applied automatically for the Beats with that configuration, and not for the other.

I'm not sure if we can translate this idea to the templates as well, but just wanted to put it out there.

Naming suggestion: the config setting could be called `append_fields` to point out that the fields are appending to the existing one, not replacing them. Or would replacing be supported as well?

ruflin:
I agree we should only append fields and not even allow to overwrite them. As template inheritance might still be around in the future in ES, we potentially could load separate templates for these additional fields. Like this we would only have to overwrite the additional template.

I agree the support for seems to be most important inside modules. So far I was mainly thinking of prometheus or http module in metricbeat, but Filebeat is also a good use case. 

Long term I'm hoping the `fields.yml` is part of the binary already (https://github.com/elastic/beats/pull/4834), so the user will not even be tempted to touch it. If he wants to add some fields, he can do it in the beats config file directly either on the global level, module or prospector level. If he has many fields, he should also be able to provide a path to one or many additional `fields.yml` files. And then an option should be available to overwrite the template or not. In the best case, this is only for the parts the user modified. 

As soon as we have the `fields.yml` as part of the Golang code, we can split it up internally into one file for each module which we already have for development. For the modules this would allow then to either use the `fields.yml` in the binary or if a file exists in the module directory, to use the one in the file. This will be tricky with overwriting etc. but I think there are a few ideas how we can solve that (see below).

It would be nice if we could detect on the beat side if the template in ES is identical to the local one or not. Or if we have multiple templates, if one does not exist or one has changed. So in the default case we could at least report that the template is not identical or one of the templates does not exist or is not identical. Then we could offer the user to only add non existing templates or overwrite the ones which are not identical which would mainly be used for development I hope.

Moving forward I suggest to put the feature on a global level out there to see how people use it and disallow overwriting existing fields. I would start with `append_fields` as config option, mark is as experimental, wait for user feedback and revise it in combination with our development on modules again.

ruflin:
This is now ready for a round of reviews. I added the feature as experimental for now so we can start playing around with it.

libbeat/common/field.go
```diff
@@ -99,6 +99,44 @@ func (f Fields) HasKey(key string) bool {
 	return f.hasKey(keys)
 }
 
+// HasNode checks if inside fields the given node exists
+// In contrast to HasKey it not only compares the leave nodes but
```
andrewkroh: `s/leave/leaf/` since it's singular in this context.

libbeat/template/template_test.go
```diff
@@ -54,3 +57,107 @@ func TestNumberOfRoutingShardsOverwrite(t *testing.T) {
...
+				},
+			},
+			error: true,
+		},
+	}
+
+	for _, test := range tests {
+		_, err := appendFields(test.fields, test.appendFields)
+		if test.error {
+			fmt.Printf("%v\n\n", test.fields)
```
andrewkroh: `t.Logf` instead?

libbeat/common/field.go
ruflin: fixed

ruflin:
@andrewkroh Ready for an other look, please squash.

libbeat/docs/template-config.asciidoc
```diff
@@ -82,3 +82,7 @@ setup.template.overwrite: false
 setup.template.settings:
   _source.enabled: false
 ----------------------------------------------------------------------
+
+*`setup.template.append_fields`*:: A list of of fields to be added to the template and Kibana index pattern.
```
andrewkroh: Based on the changelog comment this should probably be marked as experimental.

ruflin: experimental flag added.

PR #6028, opened 2018-01-09
New beat info: ephemeral_id

kvch:
New "metric" is added: `beat.info.ephemeral_id`. I am open to alternative names for this.
This ID is generated every time the Beat starts/restarts. It does not change when the config is reloaded.

I also added `strConsts` list which contains strings which should be added to the monitoring events even if it is unchange. This solution is the same as the current `gauges` list.

### DIFF
--- a/libbeat/monitoring/report/log/log.go
+++ b/libbeat/monitoring/report/log/log.go
@@ -41,6 +41,11 @@ var gauges = map[string]bool{
 	"system.load.norm.15":            true,
 }
 
+// TODO: Change this when gauges are refactored, too.
+var strConsts = map[string]bool{
+	"beat.info.ephemeral_id": true,
+}
+
 var (
 	// StartTime is the time that the process was started.
 	StartTime = time.Now()
@@ -141,7 +146,9 @@ func makeDeltaSnapshot(prev, cur monitoring.FlatSnapshot) monitoring.FlatSnapsho
 	}
 
 	for k, s := range cur.Strings {
-		if p, ok := prev.Strings[k]; !ok || p != s {
+		if _, found := strConsts[k]; found {
+			delta.Strings[k] = s
+		} else if p, ok := prev.Strings[k]; !ok || p != s {
 			delta.Strings[k] = s
 		}
 	}
--- a/libbeat/cmd/instance/metrics.go
+++ b/libbeat/cmd/instance/metrics.go
@@ -8,6 +8,8 @@ import (
 	"runtime"
 	"time"
 
+	"github.com/satori/go.uuid"
+
 	"github.com/elastic/beats/libbeat/logp"
 	"github.com/elastic/beats/libbeat/metric/system/cpu"
 	"github.com/elastic/beats/libbeat/metric/system/process"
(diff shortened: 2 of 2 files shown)


libbeat/monitoring/report/log/log.go
```diff
@@ -141,7 +146,9 @@ func makeDeltaSnapshot(prev, cur monitoring.FlatSnapshot) monitoring.FlatSnapsho
 	}
 
 	for k, s := range cur.Strings {
-		if p, ok := prev.Strings[k]; !ok || p != s {
+		if _, found := strConsts[k]; found {
+			delta.Strings[k] = s
+		} else if p, ok := prev.Strings[k]; !ok || p != s {
 			delta.Strings[k] = s
```
ph: I've looked through the conditionals here, do we need a special logic for that value?
On the first run, the value of the variable **s** for the UUID won't be found in the _delta.Strings_ so it will be added and the next runs will never change that value.

ph: Actually, we don't need that change because we never do _delta on string_ we just need to return the current value we don't need to add any special logic to handle strings constants, because once they are set they will be immutable unless we manually change that value.

This is good finding @kvch! when we refactor the gauge we will also need an _Immutable gauge_ type or something similar.

ph: even if not necessary we might want to add it to the gauge list defined in this file.

exekias:
Would `instance_id` look nicer? I'm not quite sure, It may be misleading

ph:
@exekias I think we should stay with the ephemeral_id, this is what ES and Logstash are using for the lifetime UUID.

exekias:
Good point, forget about my message then :innocent:

libbeat/monitoring/report/log/log.go
kvch: My solution is temporary, until gauges are refactored. I created a separate list, because it is not really a gauge, but a string constant. I am not sure what changes you are requesting.

ph: My point was, adding this code will result in the same behavior as before.  The code present in the _if_ and the _else if_ part of the conditionals are exactly the same. I was challenging the need to add this logic and the current need to differentiate a **static gauge** and a **strconstant** before we have a _concrete type_ in our system that will take care of the inherent difference between the two.

kvch: The first if checks if the value whether the value is constant or not. If it is constant, it is added to the `delta.Strings` list. The else if part checks if it is the first time it shows up in the string metrics or it is changed since the last sample, it is also added. If it hasn't changed since the last sample it does not show up.

In other words the first if checks if the string is in `strConsts`. The second if checks if the string was in the previous sample or if it has changed since the previous sample. Adding this code definitely changes the behavior of a Beat, as before I added this code `ephemeral_id` showed up only in the first event. The events after that did not include this info.

I don't want to mix string constants into gauges. Gauges go into a different lists which include numerical values `delta.Ints` or `delta.Floats`. Different types of metrics are already handled differently here, see `delta.Ints`, `delta.Floats and `delta.Strings`.

ph: > In other words the first if checks if the string is in strConsts. The second if checks if the string was in the previous sample or if it has changed since the previous sample. Adding this code definitely changes the behavior of a Beat, as before I added this code ephemeral_id showed up only in the first event. The events after that did not include this info.

Good point

PR #6030, opened 2018-01-09
Log prospector configured paths at INFO level

tsg:
I found myself enabling `-d "prospector"` quite often to
check the configured paths, so I thought upgrading it to INFO level is
going to be beneficial for users.

As I didn't want to increase verbosity, I looked for other Info messages that
could be Debug messages, and demoded a couple that where redundant (same information
printed in another Info message) or a little to obscure to print all the time (the ID
of the prospector).

### DIFF
--- a/filebeat/prospector/prospector.go
+++ b/filebeat/prospector/prospector.go
@@ -84,7 +84,7 @@ func New(
 // Start starts the prospector
 func (p *Prospector) Start() {
 	p.wg.Add(1)
-	logp.Info("Starting prospector of type: %v; ID: %d ", p.config.Type, p.ID)
+	logp.Debug("prospector", "Starting prospector of type: %v; ID: %d ", p.config.Type, p.ID)
 
 	onceWg := sync.WaitGroup{}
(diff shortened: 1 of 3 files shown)


filebeat/prospector/prospector.go
```diff
@@ -84,7 +84,7 @@ func New(
 // Start starts the prospector
 func (p *Prospector) Start() {
 	p.wg.Add(1)
-	logp.Info("Starting prospector of type: %v; ID: %d ", p.config.Type, p.ID)
+	logp.Debug("prospector", "Starting prospector of type: %v; ID: %d ", p.config.Type, p.ID)
```
andrewkroh: I think the trailing space in `ID: %d "` could be dropped (not sure what it was there originally).

tsg: Done.

ruflin: I think I introduced this because of the `reloading` feature to have some info there on what happens. Is this showing up also in other parts?

PR #6035, opened 2018-01-10
Add QoS to Beats

vjsamuel:
This PR allows users to add `weight` as a parameter to prospectors so that the QoS mechanism that has been implemented would prioritize producers and allow them to publish to the configured queue accordingly.

### DIFF
--- a/libbeat/publisher/queue/memqueue/broker.go
+++ b/libbeat/publisher/queue/memqueue/broker.go
@@ -209,6 +209,9 @@ func (l *chanList) concat(other *chanList) {
 		return
 	}
 
+	m := sync.Mutex{}
+	s := sync.NewCond(&m)
+	s.Signal()
 	l.tail.next = other.head
 	l.tail = other.tail
 }
--- a/libbeat/publisher/queue/qos.go
+++ b/libbeat/publisher/queue/qos.go
@@ -1,0 +1,112 @@
+package queue
+
+import (
+	"container/heap"
+	"time"
+
+	"github.com/elastic/beats/libbeat/common"
+)
+
+type Qoser interface {
+	// Release the process lock and give the processing time to the next available producer
+	ReleaseToken()
+
+	// Schedule picks the next producer who gets to produce to the queue
+	Schedule()
+
+	// Clone an instance of QoSer
+	CreateClient() QoserClient
+}
+
+type QoserClient interface {
+	// Add the producer in the wait queue based
+	RequestToken(*ProducerWrapper, int, time.Time)
+
+	// Wait until the next available slot
+	Wait()
+
+	// Send a signal to the producer to wake up
+	Awaken()
+}
+
+type weightedScheduler struct {
+	heap heap.Interface
+
+	notify chan struct{}
+}
+
+func NewWeightedScheduler() Qoser {
+	return &weightedScheduler{
+		heap:   NewHeap(),
+		notify: make(chan struct{}),
+	}
+}
+
+func (w *weightedScheduler) ReleaseToken() {
+	// Pop the most high priority producer
+	iface := w.heap.Pop()
+
+	if iface == nil {
+		return
+	}
+
+	if it, ok := iface.(*item); ok {
+		producer := it.value.(*ProducerWrapper)
+		// Notify the producer to wake up.
+		producer.Awaken()
+	}
+}
+
+func (w *weightedScheduler) Schedule() {
+	go func() {
+		for {
+			select {
+			case <-w.notify:
+				w.ReleaseToken()
+			}
+		}
+	}()
+}
+
+func (w *weightedScheduler) CreateClient() QoserClient {
+	return NewWeightedClient(w.heap, w.notify)
+}
+
+type weightedClient struct {
+	heap heap.Interface
+
+	cond   chan struct{}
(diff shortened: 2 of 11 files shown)


libbeat/publisher/queue/qos.go
```diff
@@ -0,0 +1,112 @@
...
+type QoserClient interface {
+	// Add the producer in the wait queue based
+	RequestToken(*ProducerWrapper, int, time.Time)
+
+	// Wait until the next available slot
+	Wait()
+
+	// Send a signal to the producer to wake up
+	Awaken()
+}
```
urso: a beat.Client must be able to disconnect at anytime... you missing a `Close()` as well?

libbeat/publisher/queue/qos.go
```diff
@@ -0,0 +1,112 @@
...
+	c.notify <- struct{}{}
+}
+
+func (c *weightedClient) Wait() {
+	<-c.cond
+}
+
+// TODO: Add the packet delivery logic after initial review
+func (c *weightedClient) computePriority(weight int, t time.Time) common.Float {
+	return common.Float(int64(weight))
```
urso: This seems to be quite unfair queue :)

libbeat/publisher/queue/qos.go
```diff
@@ -0,0 +1,112 @@
...
+		producer.Awaken()
+	}
+}
+
+func (w *weightedScheduler) Schedule() {
+	go func() {
+		for {
+			select {
+			case <-w.notify:
+				w.ReleaseToken()
```
urso: I wonder how effective this scheduler is. I'm a little afraid the heap will be of length 1 most of the time. The scheduler shown here very much relies on contention on the heap structure.

libbeat/publisher/queue/wrapper.go
```diff
@@ -0,0 +1,93 @@
...
+		p.RequestToken(p, p.weight, event.Content.Timestamp)
+		// Wait for the producer to get a cycle to publish
+		p.Wait()
+		p.wg.Done()
+	}()
+
+	// Wait till the go-routine is done signifying that the scheduler has allocated
+	// time for the publisher to publish the event
+	p.wg.Wait()
+}
```
urso: this function is equivalent to:

```
p.RequestToken(p, p.weight, event.Content.Timestamp)
p.Wait()
```

libbeat/publisher/queue/wrapper.go
```diff
@@ -0,0 +1,93 @@
...
+
+func (p *ProducerWrapper) Publish(event publisher.Event) bool {
+	p.waitForPublish(event)
+	return p.p.Publish(event)
+}
+
+func (p *ProducerWrapper) TryPublish(event publisher.Event) bool {
+	p.waitForPublish(event)
+	return p.p.TryPublish(event)
```
urso: There is no limit on waitForPublish. `TryPublish` must return `false` immediately if it can not serve the event right now. This is required for packetbeat, so it can continue processing events if publishing is not possible right now.

libbeat/publisher/queue/wrapper.go
```diff
@@ -0,0 +1,93 @@
...
+
+import (
+	"sync"
+
+	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/publisher"
+)
+
+// Wraps Queue interface to add QoS functionality on top of it.
+type QueueWrapper struct {
```
urso: I don't think wrappers need to be exported. Keep number of exported symbols low if possible.

ruflin:
I like to have the concept of some QoS in our pipeline that can be reused not only by the prospectors but also in modules in metricbeat or protocols in packetbeat. 

@urso In your above comment you separate queue and scheduler. Will the scheduler pick up a batch of events and then priortise them or is that part of querying the queue to make sure only the events with highest priority are picked up?

I'm thinking if the scheduler potentially already should be before the queue. Reason being is for example in filebeat, having the scheduler before the queue is the scheduler could put backpressure on the harvesters which send events with lower priority. So if the queue gets full the next events that end up in the queue are the events with highest priority / weight and not just all events. Or in the case of dropping events, events with lower priority would be dropped first. This would require the scheduler to have some knowledge of what kind of events are already in the queue. For example if the queue is full with events of weight 3 and event with weight 1 comes in, it can be dropped. But I'm aware this could add too much complexity.

libbeat/publisher/queue/memqueue/broker.go
```diff
@@ -209,6 +209,9 @@ func (l *chanList) concat(other *chanList) {
 		return
 	}
 
+	m := sync.Mutex{}
+	s := sync.NewCond(&m)
+	s.Signal()
```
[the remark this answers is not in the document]
vjsamuel: apologies. this was not supposed to be here.

opsnull:
any update?

ycombinator:
Hey @vjsamuel, are you still interested / have the time to resume work on this PR? It's pretty old at this point so just wondering if you want to pick it up back up in it's current state or close it out for a fresh implementation later

urso:
Unfortunately this PR is stalled for quite some time. We are closing it for now. We can re-open it if you decide to continue working on it. Thanks.

PR #6037, opened 2018-01-10
Improve generated documentation

tsg:
The Filebeat module all docs follow the same template. This updates
the generated docs to contain this template and mark the section that
need work as TODOs.

### DIFF
--- a/filebeat/scripts/module/docs.asciidoc
+++ b/filebeat/scripts/module/docs.asciidoc
@@ -25,6 +25,15 @@ include::../include/configuring-intro.asciidoc[]
 
 TODO: provide an example configuration
 
+:fileset_ex: {fileset}
+
 include::../include/config-option-intro.asciidoc[]
 
-TODO: document: fileset variables
+TODO: document the variables from each fileset. If you're describing a variable
+that's common to other modules, you can reuse shared descriptions by including
+the relevant file. For example:
+
+[float]
+==== `{fileset}` log fileset settings
+
+include::../include/var-paths.asciidoc[]
--- a/filebeat/scripts/module/docs.asciidoc
+++ b/filebeat/scripts/module/docs.asciidoc
@@ -1,3 +1,30 @@
+:modulename: {module}
+
 == {module} module
 
 This is the {module} module.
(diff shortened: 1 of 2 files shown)


tsg:
@dedemorton You review on this one would be appreciated.

filebeat/scripts/module/docs.asciidoc
```diff
@@ -1,3 +1,30 @@
...
+=== Example dashboard
+
+This module comes with a sample dashboard. For example:
+
+TODO: include an image of a sample dashboard
+
+include::../include/configuring-intro.asciidoc[]
+
+TODO: provide an example configuration
+
```
dedemorton: There's an example in config-option-intro.asciidoc that uses the `fileset_ex` attribute to resolve a fileset name:

````
TIP: When you specify a setting at the command line, remember to prefix the
setting with the module name, for example, +{modulename}.{fileset_ex}.var.paths+
instead of +{fileset_ex}.var.paths+.
````

So...I would add a TODO item before line 54.  Something like:

```
TODO: Set the fileset_ex attribute to the name of any fileset in your module. For example:
:fileset_ex: access
```

File: filebeat/scripts/module/docs.asciidoc [hunk 2]
```diff
@@ -1,3 +1,30 @@
...
+include::../include/config-option-intro.asciidoc[]
+
+TODO: document: fileset variables
```
dedemorton: Should we expand the instructions here and provide an example? Or will that confuse users? If you think this would be confusing to module developers, we can provide feedback when the module is submitted for review.

````
TODO: document: fileset variables. If you're describing a variable that's common to other modules, you
can reuse shared descriptions by including the relevant file. For example:
[float]
==== `access` log fileset settings

include::../include/var-paths.asciidoc[]

tsg:
@dedemorton While addressing your comments I realised that I can also make use of the `{fileset}` variable to makes things simpler. Can you have another look?

ruflin:
@kvch This might affect your PR in https://github.com/elastic/beats/pull/5544

PR #6048, opened 2018-01-11
Added example for SSL

robin13:
Clarification exactly where the SSL details should be put in the configuration

### DIFF
--- a/metricbeat/docs/modules/kubernetes.asciidoc
+++ b/metricbeat/docs/modules/kubernetes.asciidoc
@@ -54,7 +54,26 @@ metricbeat.modules:
     - event
 ----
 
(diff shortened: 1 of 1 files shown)


exekias:
Thank you for opening this, I'm afraid that part is autogenerated :/

What would you think about adding SSL parameters to default config? We can put them there, commented out https://github.com/elastic/beats/blob/master/metricbeat/module/kubernetes/_meta/config.yml

robin13:
Yes - that would be a good compromise.

ruflin:
@robin13 Let me know when you find some time to copy over your changes to the file proposed by @exekias and then run `make update`. You will see that the second file has the same changes inside as it's auto generated.

ruflin:
Closing in favor of https://github.com/elastic/beats/pull/6558 . Thanks @robin13 for bringing this up.

PR #6055, opened 2018-01-11
Add autodiscover for kubernetes

vjsamuel:
This PR adds auto discovery for kubernetes. Some sample specs for using this feature:

Filebeat: (discover all workloads in default namespace)
```
filebeat.autodiscover:
  providers:
    - type: kubernetes
      in_cluster: false
      kube_config: /home/docker/config
      host: minikube
      templates:
        - condition:
            equals:
              kubernetes.namespace: default
          config:
            - type: docker
              containers.ids:
                - "${data.kubernetes.container.id}"


output.console.pretty: true
```

Metricbeat: query for pods with label `project: prometheus`
```
metricbeat.autodiscover:
  providers:
    - type: kubernetes
      in_cluster: false
      kube_config: /home/docker/config
      host: minikube
      templates:
        - condition:
            equals:
              kubernetes.labels.project: prometheus
          config:
            - module: prometheus
              metricsets: ["stats"]
              hosts: "${data.host}:9090"

output.console.pretty: true
```


libbeat/common/kubernetes/metadata.go
```diff
@@ -47,6 +47,9 @@ func (g *metaGenerator) PodMetadata(pod *Pod) common.MapStr {
 		"pod": common.MapStr{
 			"name": pod.Metadata.Name,
 		},
+		"node": common.MapStr{
```
ruflin: Will this need an update to our `fields.yml` or is `node` already in there?

libbeat/common/kubernetes/util.go
```diff
@@ -0,0 +1,62 @@
...
+	"os"
+
+	"github.com/ericchiang/k8s"
+	"github.com/ghodss/yaml"
+
+	"github.com/elastic/beats/libbeat/logp"
+)
+
+func GetKubernetesClient(in_cluster bool, kube_config string) (client *k8s.Client, err error) {
+	if in_cluster == true {
```
ruflin: We normally use `inCluster`for naming, same for `kubeConfig`

libbeat/common/kubernetes/util.go
```diff
@@ -0,0 +1,62 @@
...
+
+func DiscoverKubernetesNode(host string, client *k8s.Client) string {
+	ctx := context.Background()
+	if host == "" {
+		podName := os.Getenv("HOSTNAME")
+		logp.Info("Using pod name %s and namespace %s", podName, client.Namespace)
+		if podName == "localhost" {
+			host = "localhost"
+		} else {
+			pod, error := client.CoreV1().GetPod(ctx, podName, client.Namespace)
```
ruflin: nit: We often use `err` instead of `error`

ruflin:
jenkins, test it

libbeat/processors/add_kubernetes_metadata/kubernetes.go
```diff
@@ -93,47 +87,12 @@ func newKubernetesAnnotator(cfg *common.Config) (processors.Processor, error) {
...
-				logp.Err("Querying for pod failed with error: ", error.Error())
-				logp.Info("Unable to find pod, setting host to localhost")
-				config.Host = "localhost"
-			} else {
-				config.Host = pod.Spec.GetNodeName()
-			}
-
-		}
-	}
+	config.Host = kubernetes.DiscoverKubernetesNode(config.Host, client)
```
walktall: The behavior of this Func is odd when beat is put outside of k8s cluster. As far as I know, host is used as a node selector while making watch requests. So when beat is deployed outside the cluster, host should be ignored or it won't get any watched info from k8s. And if beat is put in cluster, host should be set by k8s downward api I think. So should we remove this host config since it is either wrong in case outside cluster or error-prone while user specify a wrong host in config.

vjsamuel: @walktall, this piece of code when provided with a host just uses the host. it is good especially when you are testing locally without having to deploy the beat onto Kubernetes. when running in cluster, this code, gets the pod name and uses it to get the pod spec and figures out what host the pod is deployed on.

walktall: Yes, it is useful while test beat locally since user can set host to k8s node name. But if it is deployed outside the cluster or as a k8s deployment, we still need to ignore host anyway. Say I put beat in another k8s cluster, the result will be wrong here.

walktall: > when running in cluster, this code, gets the pod name and uses it to get the pod spec and figures out what host the pod is deployed on.

I prefer using k8s downward api here to get pod name though.

exekias: In general, watching all nodes from outside the cluster may suffer from other issues, like scalability.

These settings work best when running from inside the cluster, or orchestrating beats as part of the deployment (one per node, with the host value set).

We can discuss this in a new issue if you find the idea worth exploring, as I want to move this PR forward, and it only moved code around to get a cleaner interface.

libbeat/autodiscover/providers/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,175 @@
...
+		metagen:        metagen,
+		watcher:        watcher,
+		stop:           make(chan interface{}),
+		startListener:  start,
+		stopListener:   stop,
+		updateListener: update,
+	}, nil
+}
+
+func (p *Provider) Start() {
```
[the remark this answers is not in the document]
exekias: @vjsamuel lately we have been trying to make hound happy for most of its messages, could you add these comments?

exekias:
ok to test

libbeat/autodiscover/providers/kubernetes/kubernetes.go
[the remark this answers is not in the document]
vjsamuel: done

exekias:
ok to test

exekias:
Awesome work @vjsamuel, as usual, thank you! 🎉

dedemorton:
I see that docs are added in https://github.com/elastic/beats/pull/6065, so I'm removing the needs_docs label from this PR. Please open an issue against docs if additional changes are required.

PR #6057, opened 2018-01-12
Add a parameter to `docker` prospector to filter on stream

exekias:
Sometimes you are only interested on stdout or stderr, this parameters
allows to filter the input and only read one of them:

```
prospectors:
  - type: docker
    containers.stream: stderr
    containers.ids:
      - '*'
```

### DIFF
--- a/filebeat/prospector/log/config.go
+++ b/filebeat/prospector/log/config.go
@@ -85,7 +85,7 @@ type config struct {
 	JSON         *reader.JSONConfig      `config:"json"`
 
 	// Hidden on purpose, used by the docker prospector:
-	DockerJSON bool `config:"docker-json"`
+	DockerJSON string `config:"docker-json"`
 }
 
 type LogConfig struct {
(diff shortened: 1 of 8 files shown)


filebeat/prospector/log/config.go
```diff
@@ -85,7 +85,7 @@ type config struct {
 	JSON         *reader.JSONConfig      `config:"json"`
 
 	// Hidden on purpose, used by the docker prospector:
-	DockerJSON bool `config:"docker-json"`
+	DockerJSON string `config:"docker-json"`
```
ruflin: What's the reason for this change?

exekias: I'm using the internal parameter to enable docker parsing **and** pass the desired stream through it, possible values:

- `""` - default, docker parsing is disabled
- `all` - default for `docker` prospector, read all streams
- `stdout`
- `stderr`

exekias:
Thank you for the review @ruflin, this should be ready for a second look :smile:

filebeat/prospector/log/config.go
ruflin: Got it, didn't realise you passed through the stream param.

PR #6069, opened 2018-01-13
Adapt kibana index-pattern cleanup to changed dir structure.

simitt:
I assume the cleanup script should be adapted to the changed structure of the Kibana Index Pattern.


ruflin:
@simitt Good you found this. Also stumbled over it in https://github.com/elastic/beats/pull/6075#discussion_r161617208

simitt:
Thanks @ruflin! Could you please also ensure this gets backported.

ruflin:
@simitt I think we do an other 6.x PR at the end of the day (@tsg ok?).

PR #6075, opened 2018-01-15
Bump docs version for 6.1.2

tsg:
Should be merged only on the release day.

### DIFF
--- a/libbeat/_meta/kibana/default/index-pattern/libbeat.json
+++ b/libbeat/_meta/kibana/default/index-pattern/libbeat.json
@@ -12,5 +12,5 @@
       "version": 1
     }
   ],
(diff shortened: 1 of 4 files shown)


libbeat/_meta/kibana/default/index-pattern/libbeat.json
```diff
@@ -12,5 +12,5 @@
       "version": 1
     }
   ],
-  "version": "6.1.1"
```
ruflin: Not related to this PR but I wonder why this index pattern is even checked in ...

ruflin: Solved in https://github.com/elastic/beats/pull/6069

PR #6076, opened 2018-01-15
Fix CPU usage report in process.GetOne

### LINKED ISSUE #6063: Beats monitoring reports 0 for its own CPU usage

```
  "monitoring": {
    "metrics": {
      "beat": {
        "cpu": {
          "total": {
            "norm": {
              "pct": 0
            },
            "pct": 0,
            "value": 0
          }
        },
        "info": {
```

kvch:
I added the missing line of updating the state.
However, it still produces 0 from time to time. Reporting CPU usage has to be done differently as @andrewkroh  and @urso pointed it out.

I am opening the PR, because at least the total CPU usage since start in `beat.cpu.total. value` is correct. So calculating CPU usage percentage can be done in different applications.

### DIFF
--- a/libbeat/metric/system/process/process.go
+++ b/libbeat/metric/system/process/process.go
@@ -402,6 +402,7 @@ func (procStats *Stats) GetOne(pid int) (common.MapStr, error) {
 	}
 
 	e := procStats.getProcessEvent(p)
+	procStats.ProcsMap = newProcs
 
 	return e, nil
 }
--- a/libbeat/cmd/instance/metrics.go
+++ b/libbeat/cmd/instance/metrics.go
@@ -19,7 +19,6 @@ import (
 )
 
 var (
-	cpuMonitor       *cpu.Monitor
 	beatProcessStats *process.Stats
 	ephemeralID      uuid.UUID
 )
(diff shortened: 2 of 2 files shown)


kvch:
I am removing the label `in progress`, so it can be merged asap. I am going to open a separate PR with the separated snapshotting and reporting changes.

kvch:
I have removed all calculated metrics including
* `beat.cpu.total.pct`
* `beat.cpu.total.norm.pct`
* `system.cpu.total.pct`
* `system.cpu.total.norm.pct`

So it would not confuse users when looking at metrics.

All metrics are going to be reimplemented.

libbeat/metric/system/process/process.go
```diff
@@ -402,6 +402,7 @@ func (procStats *Stats) GetOne(pid int) (common.MapStr, error) {
 	}
 
 	e := procStats.getProcessEvent(p)
+	procStats.ProcsMap = newProcs
```
exekias: I don't fully get why is this updated, could you explain a bit about it? sorry, :innocent:

kvch: It contains the last samples. It must be updated so based on the previous and the current samples averages and percentages can be calculated.

PR #6077, opened 2018-01-15
Refuse to store dotted keys to prevent cyclic reference in our configuration.

ph:
The following reference can make the configuration parse go into an
infinite loop and panic.
```
filebeat_name: ${filebeat_name}
```

### DIFF
--- a/libbeat/keystore/file_keystore.go
+++ b/libbeat/keystore/file_keystore.go
@@ -13,6 +13,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"strings"
 	"sync"
 
 	"golang.org/x/crypto/pbkdf2"
@@ -88,6 +89,9 @@ func (k *FileKeystore) Retrieve(key string) (*SecureString, error) {
 
 // Store add the key pair to the secret store and mark the store as dirty.
 func (k *FileKeystore) Store(key string, value []byte) error {
+	if err := k.validateKey(key); err != nil {
+		return err
+	}
 	k.Lock()
 	defer k.Unlock()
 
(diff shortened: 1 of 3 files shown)


ph:
@tsg this PR block usage of "dot" in keys for the keystore until cyclic reference from go-ucfg is correctly supported, the changes are bigger to do in the library than I have originally though.

libbeat/keystore/file_keystore.go
```diff
@@ -382,3 +386,11 @@ func (k *FileKeystore) checkPermissions(f string) error {
 func (k *FileKeystore) hashPassword(password, salt []byte) []byte {
 	return pbkdf2.Key(password, salt, iterationsCount, keyLength, sha512.New)
 }
+
+func (k *FileKeystore) validateKey(key string) error {
+	if strings.IndexAny(key, ".") != -1 {
+		return fmt.Errorf("invalid key format, key should not have '.', key: %s", key)
```
ruflin: I wonder if we should change this message to something like `is not supported yet`. I'm ok with not supporting it in the first iteration but I think moving forward to definitively have to (if possible) as it's kind of awkward having the dot nation in all our config files but not allow it for the key store. I think it's one of the first issues people will run into.

ph: I am ok with the "is not supported yet"

ph: @ruflin its possible just it require a bit more work on ucfg :)

ph: @ruflin Maybe  "Nested key is not supported yet, key: .."

ruflin: TBH I think `nested` will not make it very clear on what it's about. Already now I often have confusion around nested fields and nested documents etc in ES so I would stear away from that word. 


```
+		return fmt.Errorf("invalid key format. '.' in keys are not supported yet. key: %s", key)
 ```

ph:
@tsg travis-ci was looking stuck on something and was not running any jobs, I've rebased the PR and repush the changes. Hopefully travis will pick it up.

ruflin:
Merged as travis and appveyor are heavily lagging behind. Jenkins was green.

PR #6078, opened 2018-01-15
Refactor: Renaming prospector to input

ph:
This is the first PR to refactor the usage of the `prospector` type to the `input` type, the code uses type aliasing to allow the input and the prospector to continue to work in parallel, providing an easier update path to the community beats. We have build pragma option to make sure we cannot compile this project if you are using an older version of go.

The prospector code didn't change in this PR, they were just moved into the input directory and still use the old prospector type.

Changes that will come after:

- Refactor each prospector to use the new Input type.
- Rename prospector reference in the yaml to input.
- Update the test to use prospector instead of input.
- Update log statement to use the new wording.

*Note: for clarity I didn't merge the commits in this PR before the review.*

This is a followup of #5944


ph:
More stuff need to be extracted from #5944, but lets get this reviewed merged. I think at 500 lines is much easier.

@ruflin Also for the input, I will migrate them one by one to reduce conflicts :)

ph:
Missed the docker compose commit. . .

filebeat/beater/filebeat.go
```diff
@@ -54,6 +54,14 @@ func New(b *beat.Beat, rawConfig *common.Config) (beat.Beater, error) {
 		return nil, err
 	}
 
+	if len(config.Prospectors) > 0 {
```
ruflin: Can you add some system tests with the old config options to make sure this works as expected? We have done this for previous renamings and I think it's pretty useful to make sure it actually works and be remembered that we break something, when we remove it.

filebeat/channel/factory.go
```diff
@@ -24,15 +24,15 @@ type clientEventer struct {
...
 // for the publisher pipline.
-type prospectorOutletConfig struct {
+type inputOutletConfig struct {
 	// event processing
 	common.EventMetadata `config:",inline"`      // Fields and tags to add to events.
 	Processors           processors.PluginConfig `config:"processors"`
 
 	// implicit event fields
-	Type string `config:"type"` // prospector.type
+	Type string `config:"type"` // input.type
```
ruflin: In case the comment is correct, this is a breaking change of the data format I think.

filebeat/channel/factory.go
```diff
@@ -99,6 +99,9 @@ func (f *OutletFactory) Create(cfg *common.Config, dynFields *common.MapStrPoint
 		fields["prospector"] = common.MapStr{
 			"type": config.Type,
 		}
+		fields["input"] = common.MapStr{
```
ruflin: Could we call this `event.type` instead if we already duplicate the field? Thinking of ECS here.

filebeat/config/config.go
```diff
@@ -82,7 +83,15 @@ func mergeConfigFiles(configFiles []string, config *Config) error {
 			return fmt.Errorf("Failed to read %s: %s", file, err)
 		}
 
-		config.Prospectors = append(config.Prospectors, tmpConfig.Filebeat.Prospectors...)
+		if len(tmpConfig.Filebeat.Prospectors) > 0 {
```
ruflin: Some system tests for this would also be nice.

filebeat/config/config.go
```diff
@@ -82,7 +83,15 @@ func mergeConfigFiles(configFiles []string, config *Config) error {
...
+			cfgwarn.Deprecate("7.0.0", "prospectors are deprecated, Use `inputs` instead.")
+			if len(tmpConfig.Filebeat.Inputs) > 0 {
+				return fmt.Errorf("prospectors and inputs used in the configuration file, define only inputs not both")
+			}
+			config.Inputs = append(config.Inputs, tmpConfig.Filebeat.Prospectors...)
```
ruflin: Nit but code wise I would prefer if you would do here what you did above which is copying `tmpConfig.Filebeat.Prospectors` to `tmpConfig.Filebeat.Inputs` and then not do an else. Like this when the code is removed later, nothing changes in the "correct" code.

filebeat/docker-compose.yml
```diff
@@ -28,13 +28,13 @@ services:
 
   elasticsearch:
     extends:
-      file: ${ES_BEATS}/testing/environments/${TESTING_ENVIRONMENT}.yml
+      file: ../testing/environments/${TESTING_ENVIRONMENT}.yml
```
ruflin: What is the reason you changed this back?

filebeat/fileset/config.go
```diff
@@ -13,5 +20,24 @@ type ModuleConfig struct {
...
+
+// NewFilesetConfig creates a new FilesetConfig from a common.Config.
+func NewFilesetConfig(cfg *common.Config) (*FilesetConfig, error) {
+	var fcfg FilesetConfig
+	err := cfg.Unpack(&fcfg)
+	if err != nil {
+		return nil, fmt.Errorf("error unpacking configuration")
+	}
+
+	if len(fcfg.Prospector) > 0 {
```
ruflin: Tests ...

filebeat/input/input.go
```diff
@@ -0,0 +1,150 @@
+package input
```
ruflin: I assume that code is mostly copy/paste/rename

filebeat/prospector/prospector.go
```diff
@@ -1,150 +1,39 @@
...
-	dynFields *common.MapStrPointer,
-) (*Prospector, error) {
-	prospector := &Prospector{
-		config:   defaultConfig,
-		wg:       &sync.WaitGroup{},
-		done:     make(chan struct{}),
-		Once:     false,
-		beatDone: beatDone,
-	}
+//+build go1.9
```
ruflin: Do you really want to add this build flag as we already require Golang 1.9 for beats. Not having it would make us detect if by accident something is built previous to 1.9 and have an error.

filebeat/input/input.go
ph: Exact, this should be as is.

filebeat/prospector/prospector.go
ph: We do require golang 1.9, but after looking at it I am not sure we enforce it the correct way.


```
 go version
go version go1.8.3 darwin/amd64
```

## This is with the pragma `build go1.9`
```
 make
go build -i -ldflags "-X github.com/elastic/beats/libbeat/version.buildTime=2018-01-16T14:18:24Z -X github.com/elastic/beats/libbeat/version.commit=c2b149f7d67b082b20f8201db719ba1c01755683"
fileset/factory.go:7:2: no buildable Go source files in /Users/ph/go/src/github.com/elastic/beats/filebeat/prospector
make: *** [filebeat] Error 1
```

## This is without the pragma `build go1.9`
```
go build -i -ldflags "-X github.com/elastic/beats/libbeat/version.buildTime=2018-01-16T14:18:42Z -X github.com/elastic/beats/libbeat/version.commit=c2b149f7d67b082b20f8201db719ba1c01755683"
# github.com/elastic/beats/libbeat/logp
../libbeat/logp/logger.go:8: syntax error: unexpected = in type declaration
make: *** [filebeat] Error 2
```

filebeat/docker-compose.yml
ph: I shouldn't have changed this.

filebeat/beater/filebeat.go
ph: Of course, I was planning to do it in a dedicated PR, because right now all the tests use the old config options :)

filebeat/channel/factory.go
ph: Why its a breaking changes? The name of the private struct changed and the only the comment changed?

filebeat/channel/factory.go
ph: I can change it to `event.type`, I would prefer to not mix concerns of ECS, but since we already add this fields it should be OK.

filebeat/config/config.go
ph: I am not sure, my argument was to prevent people defining both **prospector** and inputs, but I understand your point I will just remove that error and just copy it. Once we hit 7.0 we can "die hard"
landscape-1478879153-die-hard

ph: I have the same logic there https://github.com/elastic/beats/pull/6078/files/c2b149f7d67b082b20f8201db719ba1c01755683#diff-7a00b9fb331934c8b3bb7cf5eac57576R37, So I will remove both.

filebeat/docker-compose.yml
ph: bad rebase on my side.

filebeat/config/config.go
ph: nevermind, my previous comments, I understand your proposal.

filebeat/fileset/config.go
ph: I have added unit test for this path, which are a lot simplier than integration test.

ph: There is also test there https://github.com/elastic/beats/pull/6078/files/c2b149f7d67b082b20f8201db719ba1c01755683#diff-68d15c248f81d1c5e67c59c6dab45574R59

ph:
@ruflin I have updated and fixed a few of your concerns.

Also I didn't update the json expected file, they will come in specific PR for each of the prospector. (re: this concern the event.type)

filebeat/beater/filebeat.go
ruflin: SGTM. Didn't realise it's all on the old config. Good to know.

filebeat/channel/factory.go
ruflin: Only indirectly breaking if the comment is correct. I assume when looking at it that you removed `prospector.type` from the event, but you copied it to `input.type`.

ph: I did not remove the `prospector.type` in the event. https://github.com/elastic/beats/pull/6078/files#diff-b6244e3a701e1fad9969756499d1e8fcR99

ph:
@ruflin minus the rebase, anything else you need on this PR?

filebeat/fileset/config.go
ruflin: The part which I think unit tests here can't cover is to check if it's logged correctly. +1 on the unit test.

filebeat/prospector/prospector.go
```diff
@@ -1,150 +1,37 @@
...
+// Deprecated: See input.Register
+var Register = input.Register
 
-	var err error
-	if err = conf.Unpack(&prospector.config); err != nil {
-		return nil, err
-	}
+// GetFactory wrapper for backward compatibility
+// Deprecated: See input.GetFactory
+var GetFactory = input.GetFactory
```
ruflin: Good timing that Golang 1.9 added this :-)

filebeat/tests/system/config/filebeat.yml.j2
```diff
@@ -213,7 +213,3 @@ path:
   data: {{path_data}}
 {% endif %}
 
-{% if keystore_path %}
```
ruflin: Was this a duplicate?

ph: yup

ph:
@ruflin Rebased and changelog added.

ruflin:
@ph There seem to be some issues on CI (I looked up the commit id).

ph:
@ruflin I have fixed the keystore issue, added the `event.type` fields and updated the asciidoc and pushed a clean rebase.

ruflin:
Small PR for @ph, big step for Filebeat 🎉

PR #6080, opened 2018-01-16
Renamed auditd module fields

### LINKED ISSUE #5423: Unify the Auditbeat Event Schema
In order to make it possible to correlate data produced by one module with data from another module we need to ensure that a common set of field names is used by all modules (e.g. common data like user ID should be called `uid` rather than `audit.kernel.uid` and `audit.file.uid`).

I'd like to review the field names and make changes where possible.

andrewkroh:
Updates to the auditd module

- Update to latest go-libaudit.
- Fixed an issue where the proctitle value was being truncated.
- Fixed an issue where process args were incorrectly interpretted as hex data.
- Fixed parsing of the `key` value when multiple keys are present.
- Added a warning to the examples rules about the cost of auditing all network traffic.
- Changed the event format to allow for some common fields between modules.
- Updated fields.yml.

The dashboards need to be updated due to the field changes.

Closes #5423


auditbeat/module/auditd/_meta/accept.json
```diff
@@ -0,0 +1,76 @@
...
+                "secondary": "58140",
+                "type": "socket"
+            }
+        }
+    },
+    "event": {
+        "action": "accepted-connection-from",
+        "category": "audit-rule",
+        "module": "auditd",
+        "type": "syscall"
```
ruflin: Nice usage of `type`, module and category.

auditbeat/module/auditd/_meta/accept.json
```diff
@@ -0,0 +1,76 @@
...
+        "ip": "72.83.230.100",
+        "port": "58140"
+    },
+    "tags": [
+        "net"
+    ],
+    "user": {
+        "auid": "unset",
+        "egid": "0",
+        "euid": "0",
```
ruflin: These are the the ids which would also end up in `user.id: [...]` I assume?

andrewkroh: Yes. Do you think I should add that now? I could add it later.

ruflin: No need to add it now as it's just an addition / feature.

andrewkroh:
I updated this with fields.yml changes, squashed, and give it a useful commit message.

Please use the Rebase and Merge to keep the vendor changes separate from code changes.

auditbeat/_meta/fields.common.yml
```diff
@@ -3,7 +3,121 @@
   description: >
     Contains common fields available in all event types.
   fields:
-    - name: event.module
+  - name: event.module
+    description: >
+      The name of the module that generated the event.
+
+  - name: event.action
```
ruflin: @MikePaquette That is an interesting field. I wonder if we can reuse this.

auditbeat/_meta/fields.common.yml
```diff
@@ -3,7 +3,121 @@
...
+
+  - name: file
+    type: group
+    description: File attributes.
+    fields:
+    - name: path
+      type: text
+      description: The path to the file.
+      multi_fields:
+      - name: raw
```
ruflin: @andrewkroh Potentially something we should support in ECS by default.

auditbeat/module/auditd/_meta/fields.yml
```diff
@@ -2,51 +2,158 @@
...
+      description: set group ID
+    - name: fsgid
+      type: keyword
+      description: file system group ID
+    - name: selinux
       type: group
+      description: The SELinux identity of the actor.
       fields:
-      - name: action
+      - name: user
```
ruflin: I kind strange naming as it's `user.selinux.user`. Should this be account?

auditbeat/module/auditd/_meta/fields.yml
```diff
@@ -2,51 +2,158 @@
...
+      type: keyword
+      description: The port number.
+    - name: hostname
+      type: keyword
+      description: Hostname of the source.
+    - name: path
+      type: keyword
+      description: This is the path associated with a unix socket.
+
+  - name: network.direction
```
ruflin: Fingers cross that this is not changing ;-)

auditbeat/module/auditd/_meta/fields.yml
```diff
@@ -176,35 +179,59 @@
...
-            example: s0
-            description: The object's SELinux level.
       - name: how
         type: keyword
         description: >
           This describes how the action was performed. Usually this is the exe
           or command that was being executed that triggered the event.
-      - name: key
+
+    - name: paths
```
ruflin: Should this be singular?

auditbeat/module/auditd/_meta/fields.yml
```diff
@@ -821,6 +824,25 @@
...
+        - name: socket
+          type: group
+          fields:
+          - name: port
+            type: keyword
+            description: The port number.
+          - name: saddr
+            type: keyword
+            description: The raw socket address structure.
+          - name: addr
```
ruflin: I suggest not to use abbreviation for address.

ruflin:
@andrewkroh This PR needs a make update.

auditbeat/module/auditd/_meta/fields.yml
andrewkroh: I think we need to standardize on the enum values.

auditbeat/module/auditd/_meta/fields.yml
andrewkroh: It is usually a list of raw path objects so I named with a plural. Is that the correct way to do it?

auditbeat/module/auditd/_meta/fields.yml
andrewkroh: I opted for `addr` because that's the exact name coming from the kernel. But I can rename this is `go-libaudit`.

andrewkroh:
I rebased it to fix some merge conflicts caused by me merging another PR.

auditbeat/module/auditd/_meta/fields.yml
ruflin: +1 on having a recommend standard.

auditbeat/module/auditd/_meta/fields.yml
ruflin: My argument is normally to make it singular as long as it's not an array, as the user that is accessing it only looks for a single item and the "key" for this entry reads correct. Here for example `paths.inode`, inode is only one path. But I have this conversation often with @exekias and we probably agreed to disagree 😆

ruflin:
@andrewkroh I merged this to get the major change in. We can still continue to discuss the minor details.

Issue #6083, opened 2018-01-16
RFE: System diskio positive selection


stefws:
It might be nice with an option for system.diskio like:

diskio.name_match_regexp: '^(dm-|vd)'

to make diskio only collectd and submit data for device names matching such regexp

as discussed [here](https://discuss.elastic.co/t/system-diskio-positive-selection/115536)

ruflin:
@stefws See my comment in https://github.com/elastic/beats/pull/6085

ruflin:
Closing as https://github.com/elastic/beats/pull/6085 was merged.

PR #6085, opened 2018-01-16
Add diskio and raid filtering using regular expressions

iahmedov:
Per feature request #6083 added system.diskio and system.raid regular expression filters


iahmedov:
> We should find one that is reusable as this becomes a more common pattern.

> We have processors and should use them whenever possible to have only 1 implementation.

I agree with you, we probably need some generic way of filtering. By the way, didn't know about libbeat.Processor, which can be applied here as a way of generic filtering.
Something like this could do the job

```
- module: system
  metricsets: ["diskio", "raid"]
  processors:
    - include_event:
        regexp: ^sda
        field: diskio.name
    - drop_event:
        regexp: ^md1
        field: raid.name
```

This is going to be slower, since it works after fetching events, but it can be "ok" for some metricsets like `raid` where there is not too much data. For performance critical cases, there is always a way to write specific filters inside metricsets. If this is okay we can add regexp filter to processor actions.

> As far as I understand the initial request from @stefws the important part is that the data is already filtered before it's event retrieved (if that is possible)

Yes possible, since methods we are using do not support filtering while parsing and if only 2 metricsets require this, we can open pull requests to 3rd party libs we depend on:
 * diskio (https://github.com/shirou/gopsutil) 
 * raid (https://github.com/elastic/gosigar)

Any thoughts on this @ruflin ?

stefws:
> On 17 Jan 2018, at 05.43, Ikhtiyor <notifications@github.com> wrote:
> 
> We should find one that is reusable as this becomes a more common pattern.
> 
> We have processors and should use them whenever possible to have only 1 implementation.
> 
> I agree with you, we probably need some generic way of filtering. By the way, didn't know about libbeat.Processor, which can be applied here as a way of generic filtering.
> Something like this could do the job
> 
> - module: system
>   metricsets: ["diskio", "raid"]
>   processors:
>     - include_event:
>         regexp: ^sda
>         field: diskio.name
>     - drop_event:
>         regexp: ^md1
>         field: raid.name
> This is going to be slower, since it works after fetching events,
> 
Just my point, on own hypervisor nodes, there are hundredes of guest devices, that I’ll like not to pull data from just to drop such event again in processor action. Seems waste of time if filtering can be applied before collecting data ImHO
> but it can be "ok" for some metricsets like raid where there is not too much data. For performance critical cases, there is always a way to write specific filters inside metricsets. If this is okay we can add regexp filter to processor actions.
> 
> As far as I understand the initial request from @stefws <https://github.com/stefws> the important part is that the data is already filtered before it's event retrieved (if that is possible)
> 
@ruffin: correct

> Yes possible, since methods we are using do not support filtering while parsing and if only 2 metricsets require this, we can open pull requests to 3rd party libs we depend on:
> 
> diskio (https://github.com/shirou/gopsutil <https://github.com/shirou/gopsutil>)
> raid (https://github.com/elastic/gosigar <https://github.com/elastic/gosigar>)
> Any thoughts on this @ruflin <https://github.com/ruflin> ?
> 
Maybe if architecture would allow preprocessor filter actions more generic it could be used by other metric sets going forward…

/Steffen

iahmedov:
> Just my point, on own hypervisor nodes, there are hundredes of guest devices, that I’ll like not to pull data from just to drop such event again in processor action. 

@stefws now I understand your point, if you have exact names of devices to include, then there is a quicker way, to add it inside diskio metricset, since psutil/disk.IOCounters (which diskio uses) supports filtering by exact match names

ruflin:
@stefws @iahmedov Thanks for the input. Moving forward I suggest not to add any post processing features to the metricsets if the processors can be used. If the processors can't do what we need, we should discuss if we should extend the processors or add special functionality.

For the pre processing it really depends if the client library supports it as stated. If the feature is useful we should definitively try to make a contribution. In some cases the libs are even under our control as we already forked it because of other changes that were critical for us. 

For this PR I suggest we change it to focus only on pre-processing as @stefws initially request. So if the lib already supports it, lets see if we can use it (didn't dig into it).

stefws:
@iahmedov 'add it inside diskio metricset, since psutil/disk.IOCounters (which diskio uses) supports filtering by exact match names'

Howto utilize this for diskio?

iahmedov:
@stefws I will change the PR soon (I am on mobile now), so that in config you can give an array of device names to include in events

```
module: system
  metricsets: ["diskio"]
  diskio.names: ["sda", "sda1", "sda2", "sdb1", .... ] # exact match used
...
```

@ruflin, @stefws  what do you think about this field, is this fine?

As for `raid` they do not support matching by name yet, should be opened PR there.
Let me correct myself, raid metricset uses elastic/procfs not elastic/gosigar project

ruflin:
@iahmedov I like your suggestion. Could you also a CHANGELOG?

Looking at the code I somehow missed where the actual implementation in the call to fetch the data happens. Did I miss something?

iahmedov:
> Looking at the code I somehow missed where the actual implementation in the call to fetch the data happens. Did I miss something?

Thankfully you didn't miss anything @ruflin , instead I missed to push changes

ruflin:
@iahmedov Good to know. Somehow Github did not published my comments before, so perhaps some are not valid anymore. Will review later again.

ruflin:
jenkins, test it

metricbeat/metricbeat.reference.yml
```diff
@@ -111,6 +111,9 @@ metricbeat.modules:
   #socket.reverse_lookup.success_ttl: 60s
   #socket.reverse_lookup.failure_ttl: 60s
 
+  # Diskio configurations
+  #diskio.include_devices: ["sda", "sda1", "sda2"]
```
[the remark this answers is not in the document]
iahmedov: > Should we have [ ] to disable it or ["*"] ? Alternatives?

not specifying the config or just empty array disables filtering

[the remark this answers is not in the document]
ruflin: As our config normally follow the logic to have the "default" configure I suggest we have here `[]`. It's good to have the examples in our docs as you did.

metricbeat/module/system/diskio/_meta/docs.asciidoc
```diff
@@ -11,4 +11,17 @@ This metricset is available on:
 [float]
 === Configuration
 
-There are no configuration options for this metricset.
+*`diskio.include_devices`*:: When the `diskio` metricset is enabled, you can use the
+`diskio.include_devices` option to define a list of device names to pre-filter the
+devices that are reported. Filters only exact matches
```
ruflin: Can you change this to: `This filters only exact matches.` (dot at the end).

metricbeat/module/system/diskio/_meta/docs.asciidoc
```diff
@@ -11,4 +11,17 @@ This metricset is available on:
...
+If not set, all disk devices are returned
```
ruflin: If not set or an empty array is set, all disk devices are returned. 

(also missing dot at the end).

metricbeat/module/system/diskio/diskio.go
```diff
@@ -21,20 +21,29 @@ func init() {
...
 	mb.BaseMetricSet
 	statistics *DiskIOStat
+	names      []string
 }
 
 // New is a mb.MetricSetFactory that returns a new MetricSet.
 func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
-	ms := &MetricSet{
+	config := struct {
+		Names []string `config:"diskio.include_devices"`
```
ruflin: Can we name this variable `IncludeDevices`. This makes it easier to follow the code I think.

metricbeat/modules.d/system.yml
```diff
@@ -14,6 +14,7 @@
   process.include_top_n:
     by_cpu: 5      # include top 5 processes by CPU
     by_memory: 5   # include top 5 processes by memory
+  #diskio.include_devices: ["sda", "sda1", "sda2"]
```
ruflin: We should not add this here.

metricbeat/docs/modules/system.asciidoc
```diff
@@ -41,6 +41,7 @@ metricbeat.modules:
   process.include_top_n:
     by_cpu: 5      # include top 5 processes by CPU
     by_memory: 5   # include top 5 processes by memory
+  #diskio.include_devices: ["sda", "sda1", "sda2"]
```
ruflin: I don't think it should be added here as the above is our "default" config.

metricbeat/module/system/diskio/diskio.go
stefws: +1 would make sense

ruflin: @stefws Not sure I get your comment above?

ruflin:
@iahmedov Can you ping me when you pushed an update. I don't think I get a notification on commits.

iahmedov:
@ruflin its ready, can you please review again, thanks.

metricbeat/module/system/diskio/diskio.go
stefws: @ruflin I seconded your comment on the more sensible name for the variable and since this is a code change it means that this feature isn't currently available yet (as in v.5.6.5) which I mistakenly thought was whats iahmedov ment when he said: 
'since psutil/disk.IOCounters (which diskio uses) supports filtering by exact match names'

ruflin:
jenkins, test it

metricbeat/module/system/_meta/config.yml
```diff
@@ -14,6 +14,7 @@
   process.include_top_n:
     by_cpu: 5      # include top 5 processes by CPU
     by_memory: 5   # include top 5 processes by memory
+  #diskio.include_devices: []
```
ruflin: Can you remove this line? We should not have this in our short config. Also run  `make update` afterwards.

metricbeat/modules.d/system.yml
```diff
@@ -14,6 +14,7 @@
   process.include_top_n:
     by_cpu: 5      # include top 5 processes by CPU
     by_memory: 5   # include top 5 processes by memory
+  #diskio.include_devices: []
```
ruflin: This line should also be removed here (if `make update` does not cover it).

iahmedov:
@ruflin comments are implemented

ruflin:
jenkins, test it

CHANGELOG.asciidoc
```diff
@@ -51,6 +51,8 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
 - Rename `golang.heap.system.optained` field to `golang.heap.system.obtained`. {issue}5703[5703]
 - Support haproxy stats gathering using http (additionaly to tcp socket). {pull}5819[5819]
 - De dot keys in jolokia/jmx metricset to prevent collisions. {pull}5957[5957]
+- Support to optionally 'de dot' keys in http/json metricset to prevent collisions. {pull}5957[5957]
```
ruflin: I think this sneaked in during rebasing?

iahmedov: Turns out yes, removed entry from changelog which is not related to this pull request (just tagging to inform about change @ruflin )

ruflin:
jenkins, test it

ruflin:
Thanks @iahmedov for the ping. I just pushed one more commit as there was now a conflict in the changelog. Waiting for green from CI now.

ruflin:
@iahmedov Could you run `make update` in the metricbeat directory? One of the auto generated files seem to be out of sync.

karmi:
Hi @iahmedov, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/beats/pull/6085.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

iahmedov:
@karmi In case I missed something signed again, before also signed with same email in patch

iahmedov:
@ruflin new commit

iahmedov:
@karmi fixed mistake, rebased commit

ruflin:
jenkins, test it

ruflin:
jenkins, test it

ruflin:
@iahmedov Finally merged. Took a bit longer then expected. Thanks a lot for the contribution.

iahmedov:
@ruflin Thanks for thorough and patient review 👍

PR #6086, opened 2018-01-16
Add monitoring.Collector

### LINKED ISSUE #6063: Beats monitoring reports 0 for its own CPU usage

kvch:
This PR introduces a `monitoring.Collector` which collects metrics of the Beat and its host, so stateful metrics can be calculated correctly.

It still has minor problems, like calculated CPU usage becomes 0 from time to time.


kvch:
Closing, as calculated metrics would not be supported.

Issue #6051, opened 2018-01-11
version info not in releases


graphaelli:
same as elastic/apm-server#479, introduced into beats by #5973

graphaelli:
fixed by #6360

PR #6050, opened 2018-01-11
move compile time version information

graphaelli:
since version.go is generated by dev-tools/set_version

PR #6087, opened 2018-01-16
Revert "move compile time version information (#6050)"

graphaelli:
Temporarily remove version info that is not provided by the current release process.

before:
```
$ ./libbeat version 
mockbeat version 9.9.9 (amd64), libbeat 6.2.0 [unknown built unknown]
```

after:
```
$ ./libbeat version
mockbeat version 9.9.9 (amd64), libbeat 6.2.0
```

for #6051


ruflin:
@graphaelli Did you investigate on why it's not in 6.2? Lets see if we can get this later in again for the next 6.x releases.

graphaelli:
I ran out of time to investigate.  I assume something in the `package` target does a `go build` directly rather than relying on the `${BEAT_NAME}` name.  I imagine we need that for cgo to work.  I'll follow up with you directly.

graphaelli:
@ruflin I can't find an issue for fixing this, but I have a fix (updates to `dev-tools/packer/docker/xgo-image/beats-builder/build.sh`) that works for the linux package, and I think the windows package (haven't tested directly), but doesn't work for darwin.  I'll ask for help if I can't figure it out.

graphaelli:
figured it out, can only provide one `-ldflags` arg to `go build`, last one wins

PR #6094, opened 2018-01-17
Bump version to 6.1.3 in the 6.1 branch

tsg:


andrewkroh:
This would probably be a good time to update our testing/environment versions and deploy/kubernetes/*-kubernetes.yaml files too (assuming they haven't already been updated).

simitt:
@andrewkroh would it be ok with you to merge this and then create a seperate PR for updating testing/env, so we can update the apm-server version?

PR #6097, opened 2018-01-17
New metrics: CPU ticks && CPU time

### LINKED ISSUE #6063: Beats monitoring reports 0 for its own CPU usage

kvch:
New metrics that report values since starting the Beat:
* `beat.cpu.user.ticks` user CPU usage in ticks
* `beat.cpu.user.time` user CPU usage time in milliseconds
* `beat.cpu.system.ticks` system CPU usage in ticks
* `beat.cpu.system.time` system CPU usage in milliseconds
* `beat.cpu.total.value` total CPU usage
* `beat.cpu.total.ticks` total CPU usage in ticks
* `beat.cpu.total.time` total CPU usage in milliseconds

Please, note this PR includes gosigar changes to see if tests pass. After it passes, I will open a PR in gosigar with the its changes and remove them from this PR.

Closes #6063


libbeat/cmd/instance/metrics.go
```diff
@@ -106,36 +106,81 @@ func reportBeatCPU(_ monitoring.Mode, V monitoring.Visitor) {
 	V.OnRegistryStart()
 	defer V.OnRegistryFinished()
 
-	totalCPUUsage, err := getCPUUsage()
+	totalCPUUsage, totalCPUUserTicks, totalCPUSystemTicks, totalCPUTicks, err := getCPUUsage()
```
urso: If we have `time`, I don't think we need ticks. Are the ticks used somewhere else?

libbeat/cmd/instance/metrics.go
```diff
@@ -106,36 +106,81 @@ func reportBeatCPU(_ monitoring.Mode, V monitoring.Visitor) {
...
+		monitoring.ReportInt(V, "time", userTime)
+	})
+	monitoring.ReportNamespace(V, "system", func() {
+		monitoring.ReportInt(V, "ticks", int64(totalCPUSystemTicks))
+		monitoring.ReportInt(V, "time", systemTime)
+	})
 	monitoring.ReportNamespace(V, "total", func() {
 		monitoring.ReportFloat(V, "value", totalCPUUsage)
+		monitoring.ReportInt(V, "ticks", int64(totalCPUTicks))
+		monitoring.ReportInt(V, "time", userTime+systemTime)
```
urso: I see how `monitoring.ReportX` can become very cumbersome to use. Using go-structform, we could create a `structform.Visitor` mapping to `monitoring.Visitor`. Then we could write:

```
type allCPUStats struct {
  User cpuStats
  System cpuStats
  Total cpuStats
}

type cpuStats struct {
  Time int64
}

...

monitoring.ReportStruct(V, allCPUStat{
  User: cpuStat{ ... }
  System: cpuStat{ ... }
  Total: cpuStat{ ... }
})
```

WDYT?

libbeat/cmd/instance/metrics.go
```diff
@@ -106,36 +106,81 @@ func reportBeatCPU(_ monitoring.Mode, V monitoring.Visitor) {
...
 	})
-
 }
 
-func getCPUUsage() (float64, error) {
+func getCPUUsage() (float64, uint64, uint64, uint64, error) {
```
urso: return tuple is quite big. Some naming (e.g. introduce a new type) would make the result more parseable.

libbeat/metric/system/process/process.go
```diff
@@ -227,6 +227,20 @@ func getProcState(b byte) string {
 	return "unknown"
 }
 
+// GetOwnResourceUsageTime return the user and system CPU usage time in milliseconds
+func GetOwnResourceUsageTime() (int64, int64, error) {
+	r := sigar.Rusage{}
```
urso: nice :)

btw. the time is based on ticks.

vendor/github.com/elastic/gosigar/sigar_windows.go
```diff
@@ -408,3 +408,7 @@ func getWin32OperatingSystem() (Win32_OperatingSystem, error) {
 	}
 	return dst[0], nil
 }
+
+func GetRusage() (windows.Rusage, error) {
+	return nil, ErrNotImplemented{runtime.GOOS}
+}
```
urso: for windows have a look at PgSQL getrusage compat code: https://doxygen.postgresql.org/getrusage_8c.html

libbeat/metric/system/process/process.go
urso: On linux getrusage is a syscall: https://github.com/torvalds/linux/blob/master/kernel/sys.c#L1660
 (also see: [task_cputime](https://github.com/torvalds/linux/blob/6363b3f3ac5be096d08c8c504128befa0c033529/kernel/sched/cputime.c#L865 and [vtime.h](https://github.com/torvalds/linux/blob/ead751507de86d90fa250431e9990a8b881f713c/include/linux/vtime.h#L52:13)).

andrewkroh: If this is returning "time" then I think it should use `time.Duration`.

kvch: @urso I decided to include both because I read that a CPU tick can be a different time on Linux and Windows.

libbeat/cmd/instance/metrics.go
kvch: No, it's not used, it's just reported. I thought it could be useful as the relation CPU time and CPU ticks are different on Linux and Windows.

kvch:
I implemented partially `getrusage` for Windows.
I am still working on `reportStruct`.
@urso should I delete ticks from the reported metrics?

urso:
> I implemented partially getrusage for Windows.

👍 

> I am still working on reportStruct.

Cool.

> should I delete ticks from the reported metrics?

From UI point, ticks is of limited use without knowing the cpu frequency... well... given ticks/time is about constant over time (makes me wonder if CPU throttling could be detected :) ). We will restrict the reported metrics to required fields anyways. Just keep them in for now.

kvch:
I cleaned up the PR and opened a new one in gosigar with vendor changes: https://github.com/elastic/gosigar/pull/95
This PR will fail in the CIs until that PR is merged and updated in Beats.

kvch:
I rebased the PR, so now it includes gosigar changes in vendor.

PR #6098, opened 2018-01-17
Revert the validation on dotted key in the keystore and add a cyclic reference test case

ph:
We are reverting the change in  0cdcf4f957f6acb8686b0ac5002b14fed636c4ba and we will allow dotted key in the keystore. We also add a system env test that add a cyclic reference use case.

This PR depends on https://github.com/elastic/go-ucfg/pull/97 allowing configuration to define cyclic references.


ph:
https://github.com/elastic/go-ucfg/pull/97 was merged, waiting for CI to run everything.

ph:
Thanks @exekias!

ph:
updated with a fixed version on go-ucfg (0.5.0) instead of a commit hash.

PR #6100, opened 2018-01-17
Enhance add_docker_metadata to enrich based on PID

andrewkroh:
This PR enhances `add_docker_metadata` with the ability to enrich events containing process IDs.

The processor uses cgroup membership data from `/proc/pid/cgroup` to determine if the process is running inside of a Docker container. It caches the PID -> CID mapping for 5 minutes (based on time of last access).

The default configuration sets `match_pids: [process.pid, process.ppid]`. It falls back to the PPID in case the process has exited before the processing occurs.

### DIFF
--- a/libbeat/_meta/config.reference.yml
+++ b/libbeat/_meta/config.reference.yml
@@ -97,6 +97,10 @@
 #- add_docker_metadata:
 #    host: "unix:///var/run/docker.sock"
 #    match_fields: ["system.process.cgroup.id"]
+#    match_pids: ["process.pid", "process.ppid"]
+#    match_source: true
+#    match_source_index: 4
+#    cleanup_timeout: 60
 #    # To connect to Docker over TLS you must specify a client and CA certificate.
 #    #ssl:
 #    #  certificate_authority: "/etc/pki/root/ca.pem"
--- a/libbeat/processors/add_docker_metadata/add_docker_metadata_test.go
+++ b/libbeat/processors/add_docker_metadata/add_docker_metadata_test.go
@@ -1,6 +1,8 @@
 package add_docker_metadata
 
 import (
+	"fmt"
+	"os"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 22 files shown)


libbeat/processors/add_docker_metadata/add_docker_metadata_test.go
```diff
@@ -11,6 +13,27 @@ import (
 	"github.com/elastic/beats/libbeat/common/docker"
 )
 
+func init() {
+	// Stub out the procfs.
+	processCgroupPaths = func(_ string, pid int) (map[string]string, error) {
+		switch pid {
+		case 1000:
+			return map[string]string{
+				"cpu": "/docker/FABADA",
```
exekias: Random trivia: Did you know Fabada in Spanish typical meal?, delicious:

Fabada asturiana

libbeat/processors/add_docker_metadata/config.go
```diff
@@ -6,23 +6,26 @@ import (
...
 	CleanupTimeout time.Duration `config:"cleanup_timeout"`
 }
 
 func defaultConfig() Config {
 	return Config{
 		Host:        "unix:///var/run/docker.sock",
 		MatchSource: true,
-		SourceIndex: 4,
+		SourceIndex: 4, // Use 4 to match the CID in /var/lib/docker/containers/<container_id>/*.log.
+		MatchPIDs:   []string{"process.pid", "process.ppid"},
```
exekias: Perhaps we can add a default for `hostFS` here?

andrewkroh: In the cgroup package if the param is empty it will set a default. If you want me to add one here (`/`) to override the package default (also `/`) I can add that.

andrewkroh:
The defaults field names are based on ECS. Auditbeat has been updated to follow that naming convention (`audit.kernel.data.pid` was changed to `process.pid` for 6.2).

libbeat/processors/add_docker_metadata/config.go
exekias: nah, that's ok! I didn't know `""` was a valid default, works for me :+1:

libbeat/processors/add_docker_metadata/add_docker_metadata_test.go
dedemorton: That looks amazing. I want some.

libbeat/_meta/config.reference.yml
```diff
@@ -97,6 +97,10 @@
 #- add_docker_metadata:
 #    host: "unix:///var/run/docker.sock"
 #    match_fields: ["system.process.cgroup.id"]
+#    match_pids: ["process.pid", "process.ppid"]
```
ruflin: Could we make `match.*` a namespace?

libbeat/processors/add_docker_metadata/add_docker_metadata_test.go
ruflin: I think I need to order one of these next time I get spanish food.

PR #6113, opened 2018-01-19
Use official client in Docker module

exekias:
This unifies the client we are using for all Docker matters to the official one, with the added benefit of saving some space in the Metricbeat binary (~2MB)


metricbeat/module/docker/docker.go
```diff
@@ -99,38 +118,18 @@ func FetchStats(client *docker.Client, timeout time.Duration) ([]Stat, error) {
...
-		ID:      container.ID,
-		Stats:   statsC,
-		Stream:  false,
-		Timeout: timeout,
+	containerStats, err := client.ContainerStats(ctx, container.ID, false)
+	if err != nil {
+		return event
 	}
 
-	wg.Add(2)
```
ruflin: Really happy to see all these go functions gone.

douaejeouit:
Great changes  👍 !

ruflin:
@douaejeouit Good to see you are still following / watching your baby 🎉

pstrzelczak:
This change does not seem to be released in any official version. Any plans to do so?
I am asking this because according to @cpuguy83 this would fix https://github.com/moby/moby/issues/37061
Could you please confirm this statement?

ruflin:
This didn't make it into 6.2 but will be in the upcoming 6.3.

pstrzelczak:
@ruflin Thans for this information.
Can anyone confirm or reject @cpuguy83 comment about the impact of using old docker client on dockerd?
"
Looking at the code for metricsbeat, it looks like it was recently changed to use the official docker client. But before that it was using a different client, and it was doing it incorrectly where the response body was never closed. I'm not sure what version you are on, or if there is a release of it that includes this change: elastic/beats#6113
"
"
The thing is, in dockerd the client connection is handled by the stdlib http server. These are automatically closed out except in a few specific cases (where the daemon hijacks the connection to enable 2-way communication for container attach). The old metricsbeat code was *definitely* incorrect (the docker client was waiting on a nil channel, which will always block because metricsbeat wasn't setting the channel, in order to close out the response body).
"

cpuguy83:
The go-dockerclient package has this: https://github.com/fsouza/go-dockerclient/blob/master/container.go#L1150

The docker module in metricsbeat is not setting the `Done` channel. https://github.com/elastic/beats/blob/v6.2.4/metricbeat/module/docker/docker.go#L108
Since it is a nil channel, the response body is never closed.

exekias:
I've opened https://github.com/elastic/beats/issues/7150 to track the issue in Metricbeat, from what I see, master and `6.3` are not affected.

exekias:
From what I see, we don't pass the `Done` channel, but still, https://github.com/fsouza/go-dockerclient/blob/master/container.go#L1114 should be called. `Done` channel is used to prematurely stop the call. If I'm correct, this is not a bug in Metricbeat. I couldn't reproduce the issue either.

cpuguy83:
Ah cheers, I missed that defer.
Yes, the issue in moby is not likely from this, but something I found while
investigating, seems fine in any case.

exekias:
no worries, thank you for taking the time!

PR #6114, opened 2018-01-19
Set version to 6.3.0 in the 6.x branch

tsg:


tsg:
Added `in progress` until the others teams are ready with the same version bump.

Issue #6068, opened 2018-01-13
[Discuss] port k8s client to kubernetes/client-go


walktall:
Just like what is described in https://github.com/ericchiang/k8s#should-i-use-this-or-client-go.

I think that using kubernetes/client-go is a better choice since caching in libbeat can be handled by client-go Informer directly with less code and better abstract.

@exekias Any idea? I am currently writing code for https://github.com/elastic/beats/issues/5932, and I am thinking about using client-go instead.

exekias:
Hi @walktall, thank you for your input, it's very valid.

While we would like to have all the benefits of `client-go`, it would increase our resulting binary size by more than 30MB. While one could argue it's not that much, Beats support lots of different use cases and hardware, and in some of them, size matters.

We have been trying to avoid this change, at least for now. And from our experience so far, current client is working fine. Probably we can reevaluate this in some months, and check if `client-go` got slimmer or kubernetes use case of beats grew significantly.

walktall:
@exekias Agree with you that size could be matter, I think it is worth to try though. I will try to replace it to client-go first since code change not that much and client-go may not take that much space now than 7 month ago when k8s client vendored.

ruflin:
@walktall I'm quite interested in the outcome of this experiment. I remember doing something similar about a year ago with docker and the outcome on our side was a fork of the docker repo with 1-2 methods removed. The reasons for this was that removing these 1-2 methods (which we didn't need) removed 95% of the dependencies from our side. As far as I remember the dependencies were related to command line options for windows but as we don't access it through the command line, we didn't need all of this. I'm writing this because I could see similar things happening on the k8s-client side.

walktall:
@exekias You are right, I have made a client-go version of beats using latest client-go^6.0.0, for metricbeat, it grows 32MB.

@ruflin I am going to find out if there is something similar to remove in client-go.

PR #6117, opened 2018-01-19
[Discuss&POC] refactor k8s client to k8s.io/client-go

walktall:
Ref: https://github.com/elastic/beats/issues/6068

I didn't update vendor and fully test it yet, just open this PR for discuss.

Will be very happy if you can take a look! @exekias @vjsamuel @ruflin 

Maybe it's still too big for bin to change client, but I abstract the `common/kubernetes` pkg and make it easier to use hopefully.

And also, found 2 bugs while refactoring.
 * query pod meta won't refresh deleted map timeout
 * delete event for k8s autodiscovery will be delayed since watcher delay the event push

With one another enhancement: beat can auto discovery correct k8s node now. So host in config might only be useful when you are testing...


ruflin:
@walktall Thanks for taking a stab at this. I haven't looked much at the code yet but I saw you mentioned that you discovered / solved two issue. I'm wondering if these could go into a separate PR to have them in rather soon or these depend on changing the library?

walktall:
@ruflin Yes, I will make another PR about two issues only.

libbeat/common/kubernetes/util.go
```diff
@@ -1,68 +1,123 @@
...
-				host = "localhost"
-			} else {
-				host = pod.Spec.GetNodeName()
-			}
+// inClusterNamespace gets namespace from serviceaccount when beat is in cluster.
+// code borrowed from client-go with some changes.
+func inClusterNamespace() string {
+	// This way assumes you've set the POD_NAMESPACE environment variable using the downward API.
+	// This check has to be done first for backwards compatibility with the way InClusterConfig was originally set up
+	if ns := os.Getenv("POD_NAMESPACE"); ns != "" {
```
exekias: We stopped using this, in favor of `client.Namespace`, what's the motivation for this change?

libbeat/common/kubernetes/util.go
```diff
@@ -1,68 +1,123 @@
...
+		return defaultNode
+	}
+
+	nodes, err := clientset.CoreV1().Nodes().List(v1.ListOptions{})
+	if err != nil {
+		logp.Err("kubernetes: Querying for nodes failed with error: ", err.Error())
+		return defaultNode
+	}
+	for _, n := range nodes.Items {
+		if n.Status.NodeInfo.MachineID == mid {
```
exekias: Cool addition, I guess all the fallback mechanisms make node discovery bulletproof :)

libbeat/common/kubernetes/watcher.go
```diff
@@ -21,241 +27,89 @@ type Watcher interface {
...
 
-	// ListenStart returns a bus listener to receive pod started events, with a `pod` key holding it
-	ListenStart() bus.Listener
-
-	// ListenUpdate returns a bus listener to receive pod updated events, with a `pod` key holding it
-	ListenUpdate() bus.Listener
-
-	// ListenStop returns a bus listener to receive pod stopped events, with a `pod` key holding it
-	ListenStop() bus.Listener
+	AddEventHandler(ResourceEventHandler)
```
exekias: :100: to this approach

libbeat/common/kubernetes/util.go
walktall: Since `k8s.io/client-go` doesn't have a field `Namespace` in it, so `client.Namespace` is kind like a tricky in `ericchiang/k8s` I think. This func is borrowed from client-go, and after rethink about it, I think user should not set namespace according downward API since if beat is in cluster, then it should be auto detected, if beat is outside cluster, this checking is meaningless. Will remove it from new PR.

walktall:
@exekias For switching client-go:

With go 1.9.2, build with `go build -ldflags="-s -w"`, it grows 12mb, from 53mb to 65mb.

With go 1.9.2, build with `go build`, it grows 19mb. From 104mb to 123mb.

Anyway, I will put the switching pr the last one, since it is only imported by `common/kubernetes`.

ruflin:
@walktall Did you ever investigate if we could sligthly tweak the import to massively reduce the dependency. I had a similar experience with docker.

vjsamuel:
i think that might be a good exercise to see if it works but the Kube client as is does a lot of things in addition to just making API calls.

walktall:
@ruflin I tried it before but I can not find it out...

@vjsamuel If we want to use k8s informer, then it is not just making API calls I think.

walktall:
Close due to no further plan for this right now since library currently used works well.

PR #6120, opened 2018-01-19
 Migrate: Log prospector to the input interface

ph:
require #6078

### DIFF
--- a/filebeat/input/log/input.go
+++ b/filebeat/input/log/input.go
@@ -11,8 +11,8 @@ import (
 
 	"github.com/elastic/beats/filebeat/channel"
 	"github.com/elastic/beats/filebeat/harvester"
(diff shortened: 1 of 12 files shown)


filebeat/input/log/input.go
```diff
@@ -11,8 +11,8 @@ import (
 
 	"github.com/elastic/beats/filebeat/channel"
 	"github.com/elastic/beats/filebeat/harvester"
+	input "github.com/elastic/beats/filebeat/input"
```
ruflin: Why do you prefix it here with `input`? It should already be input.

ph: let me check that, i believe my editor did this change.

ph:
@ruflin rebase and removed the input, not sure about the import, I've never managed the import manually.

Issue #6123, opened 2018-01-19
[Metricbeat] kubernetes.node.cpu.allocatable.cores should be a float not a long 


simianhacker:
The value stored for `kubernetes.node.cpu.allocatable.cores` is a float not a long. We should update the mapping accordingly.

ruflin:
I opened https://github.com/elastic/beats/pull/6130 to solve this.

PR #6130, opened 2018-01-21
Change kubernetes.node.cpu.allocatable.cores to float

ruflin:
Closes https://github.com/elastic/beats/issues/6123


exekias:
should we add a changelog entry?

ruflin:
@exekias Yes we should. Added and rebased on master which should also fix / rerun jenkins.

PR #6138, opened 2018-01-22
Auditbeat GA cleanup

andrewkroh:
Fix issues affecting the latest version of Auditbeat.

- Add filter to omit symlinks from "World Writeable File Count" viz.
- Remove cfgwarn.Beta from Auditbeat modules.
- Adding missing structured logging key name to an Auditbeat warning.
- Add breaking changes documentation for Auditbeat 6.2

### DIFF
--- a/auditbeat/docs/breaking.asciidoc
+++ b/auditbeat/docs/breaking.asciidoc
@@ -11,10 +11,12 @@ There are changes that affect both the configuration and the event schema.
 [float]
 === Configuration Changes
 
-The modules and metricsets were renamed and your configuration must be updated.
+The audit module has been renamed and is now two separate modules: the
+<<auditbeat-module-auditd,auditd module>> and the
+<<auditbeat-module-file_integrity,file_integrity module>>. You must update your
+configuration to use these modules.
 
-The `kernel` metricset has been renamed to the
-<<auditbeat-module-auditd,auditd module>>.
+The `kernel` metricset has become the <<auditbeat-module-auditd,auditd module>>.
 
 .Old Config
 [source,yaml]
@@ -45,7 +47,7 @@ The `kernel` metricset has been renamed to the
     # Rules
 ----
 
-The `file` metricset has been renamed to the
+The `file` metricset has become the
 <<auditbeat-module-file_integrity,file_integrity module>>.
 
 .Old Config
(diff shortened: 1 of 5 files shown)


auditbeat/docs/breaking.asciidoc
```diff
@@ -0,0 +1,129 @@
...
+
+If you are using the default index naming and template settings then no action
+is required because the Auditbeat version number is included in then index
+and template names. This prevents any field mapping issues between versions.
+
+In Kibana you will need to import the latest dashboards that work with the
+new event format. The new dashboards will not work with data produced by older
+versions of Auditbeat.
+
+.Summary of Field Renames
```
dedemorton: I'd just say "Renamed Fields" for the heading here.

auditbeat/docs/breaking.asciidoc
```diff
@@ -0,0 +1,129 @@
...
+Most field names were changed in 6.2. This is because of the module renames as
+well as a desire to use common field names for similar data types between
+the modules (and future modules). The table below provides a summary of the
+field changes.
+
+If you are using the default index naming and template settings then no action
+is required because the Auditbeat version number is included in then index
+and template names. This prevents any field mapping issues between versions.
+
+In Kibana you will need to import the latest dashboards that work with the
```
dedemorton: Just say, "In Kibana you need to import....

dedemorton: might be worth providing the command here for importing the dashboards? Or a link to the docs?

auditbeat/docs/breaking.asciidoc
```diff
@@ -0,0 +1,129 @@
+[[auditbeat-breaking-changes]]
```
dedemorton: Might be a good idea to add something to the [breaking changes](https://www.elastic.co/guide/en/beats/libbeat/current/breaking-changes-6.0.html) section of the platform reference that points to the Auditbeat topic.

andrewkroh: I think that it would be good to add something over there, but that page is specific to 6.0. Perhaps on the [parent page](https://www.elastic.co/guide/en/beats/libbeat/current/breaking-changes.html) to that one we could add a list of links?

auditbeat/docs/breaking.asciidoc
```diff
@@ -0,0 +1,129 @@
...
+
+[float]
+=== Event Schema Changes
+
+Most field names were changed in 6.2. This is because of the module renames as
+well as a desire to use common field names for similar data types between
+the modules (and future modules). The table below provides a summary of the
+field changes.
+
+If you are using the default index naming and template settings then no action
```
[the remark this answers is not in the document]
andrewkroh: As it turns out there are no mapping conflicts between 6.1.2 and 6.2 so I'm removing this paragraph.

andrewkroh:
@dedemorton The PR is updated. Thanks for reviewing.

auditbeat/docs/breaking.asciidoc
dedemorton: Yes, a list of links might work, or we could add a container for 6.2 that just points to the Auditbeat topic. If Auditbeat is the only Beat that has breaking changes, a container for 6.2 is overkill, IMO.

auditbeat/docs/index.asciidoc
```diff
@@ -17,6 +17,8 @@ include::../../libbeat/docs/contributing-to-beats.asciidoc[]
 
 include::./getting-started.asciidoc[]
 
+include::./breaking.asciidoc[]
```
dedemorton: Didn't notice this the first time I reviewed, but the include for breaking changes needs to come after the repositories topic (line 22), or you end up with a TOC that looks like this:

image

andrewkroh:
Fixed the TOC, but I didn't do the linking between the platform BC page and the Auditbeat BC page.

PR #6139, opened 2018-01-22
[DOCS] Upgrade branch to 6.2

lcawl:
This PR changes the "branch" attribute from 6.x to 6.2.

### DIFF
--- a/libbeat/docs/version.asciidoc
+++ b/libbeat/docs/version.asciidoc
@@ -1,5 +1,5 @@
 :stack-version: 6.2.0
-:doc-branch: master
+:doc-branch: 6.2
 :go-version: 1.9.2
 :release-state: unreleased
 :python: 2.7.9
--- a/libbeat/docs/version.asciidoc
+++ b/libbeat/docs/version.asciidoc
@@ -5,4 +5,4 @@
 :python: 2.7.9
 :docker: 1.12
 :docker-compose: 1.11
-:branch: 6.x
+:branch: 6.2


libbeat/docs/version.asciidoc
```diff
@@ -5,4 +5,4 @@
 :python: 2.7.9
 :docker: 1.12
 :docker-compose: 1.11
-:branch: 6.x
+:branch: 6.2
```
dedemorton: doc-branch also needs to be 6.2. That's the attribute used in 99% of the Beats docs.

The `branch` attribute was added recently when you integrated the X-Pack content here https://github.com/elastic/beats/commit/08f9be04058782c31a5774f5b74b91f4aa0fbc95. You reverted some of the changes. Is {branch} even used anywhere?

dedemorton: Oh I see. It's the disconnect between the shared attributes and the ones I use for Beats. I'll add a cleanup issue for this. Just make sure you bump the doc-branch version.

Incidentally, I think doc-branch is more descriptive

lcawl:
Thanks, @dedemorton, I didn't notice that other doc-branch attribute.  I've updated it now. 
With respect to which topics are using the "branch" attribute, I know at the very least the recently-added security topics use it (e.g. https://raw.githubusercontent.com/elastic/beats/master/libbeat/docs/security/securing-beats.asciidoc, which uses the shared attribute {xpack-ref}, which in turn uses {branch} per https://github.com/elastic/docs/blob/master/shared/attributes.asciidoc).

dedemorton:
@lcawl I think since I import the shared attributes and then import the Beats attributes (which override any duplicates in the shared attributes) we are probably covered without {branch}. I still have some cleanup work pending for the attributes anyhow, so I'll just add that to the list.

PR #6143, opened 2018-01-23
Removing beta label from jolokia/jmx metricset

ruflin:


metricbeat/module/jolokia/jmx/jmx.go
```diff
@@ -69,16 +62,19 @@ func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
 	http.SetMethod("POST")
 	http.SetBody(body)
 
+	log := logp.NewLogger(metricsetName)
+
 	if logp.IsDebug(metricsetName) {
-		debugf("The body for POST requests to jolokia host %v is: %v",
-			base.HostData().Host, string(body))
+		log.Debugw("Jolokia request body",
```
ruflin: @andrewkroh Is this the expected use of the new logging (first time user ;-) ).

metricbeat/module/jolokia/jmx/jmx.go
```diff
@@ -69,16 +62,19 @@ func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
 	http.SetMethod("POST")
 	http.SetBody(body)
 
+	log := logp.NewLogger(metricsetName)
```
andrewkroh: Given that there can be multiple instances of this metricset I recommend instantiating this `Logger` with context about the host. For example:

```
log := logp.NewLogger(metricsetName).With("host", base.HostData().Host)
```

metricbeat/module/jolokia/jmx/jmx.go
andrewkroh: You can omit `host` if you add it to the logger's context.

ruflin: done

metricbeat/module/jolokia/jmx/jmx.go
ruflin: done

metricbeat/module/jolokia/jmx/jmx.go
```diff
@@ -69,16 +62,19 @@ func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
...
+
 	if logp.IsDebug(metricsetName) {
```
ruflin: @andrewkroh Do we have any direct replacment method for `IsDebug`?

andrewkroh: Nope. Calling `IsDebug` is mostly unnecessary when using the structured logger because the fields are lazily marshaled to a log message. And for the print and printf style methods the string formatting is not performed until after checking if the logging level is enabled. So the only performance hit is in the allocation of the slice that holds the fields.

ruflin:
@andrewkroh Ready for an other review. Changed the logging, added a Changelog entry and rebase on master.

ruflin:
Rebased again.

PR #6146, opened 2018-01-23
libbeat: Add a new way to detect node for k8s which using machine-id

walktall:
This PR comes from https://github.com/elastic/beats/pull/6117

Focus on node auto detection when beat is deployed outside k8s cluster.

@exekias PTAL~


exekias:
jenkins, test it please

walktall:
@exekias changelog  entry added, PTAL.

Issue #3731, opened 2017-03-07
Winlogbeat misses some events with Forwarded Events


dandrestor:
I am running Winlogbeat 5.2.0 x64 on Windows 2012 R2. This windows machine (henceforth "the collector") is receiving forwarded event logs from various sources. I have isolated one remote machine, sending only one event log (Security), for testing purposes, in a separate event log on the collector.

Winlogbeat is dropping some of these events. I have noticed that the number of events written by Winlogbeat is consistently about 95% of the actual number of events received through Windows Event Forwarding. I see no errors in the Winlogbeat logs.

I understand this description is not very detailed, but I'm not sure what other information I could provide. If anybody more knowledgeable picks this up however, I'm happy to contribute with any testing that might be required.

dandrestor:
Possibly related: https://discuss.elastic.co/t/winlogbeat-ignores-registry-file-with-forwarded-events/76601/5

andrewkroh:
I believe that the problem is with the way Winlogbeat stores the "read pointer" for the `Forwarded Events`. We need to make [this](https://github.com/elastic/beats/issues/1054#issuecomment-200548820) enhancement to fix it.

paddibr:
same error here on Windows Server 2012 R2 with forwarded events

andrewkroh:
One workaround until this is fixed is to to use Logstash with fingerprint filter to create the _id field based on the event's `@timestamp` + `computer_name` + `log_name` + `record_number`. (`@timestamp` comes from the event record and not the time Winlogbeat read the event so it's safe to use in a fingerprint.)

Then elasticsearch will deduplicate the data because each unique event has only one possible `_id`.

dandrestor:
@andrewkroh I saw you changed the title, and wanted to double-check that we are speaking about the same thing here. I opened this issue not because Winlogbeat will lose events when starting up (resuming), but rather during normal operation. The issue with the resume we discussed earlier is possibly related, but not the same problem.
If these are in fact two separate problem, maybe it's worth changing the title again to something more suggestive. Also, would your workaround still apply in this case?

PeterZuge:
I am wondering on the status of this bug (Winlogbeat doesn't resume correctly with Forwarded Events #3731).  I can see it is still marked open, but I was hoping it might be picked up on a roadmap or has a date the bug would be likely fixed.

[...]

Issue #1054, opened 2016-02-27
[Winlogbeat] XML filtering


jimmyjones2:
The Windows API allows [XML filtering of events](https://blogs.technet.microsoft.com/askds/2011/09/26/advanced-xml-filtering-in-the-windows-event-viewer/) - so you can only subscribe to "useful" events rather than the general torrent, again something I've used heavily in the past with nxlog.

``` xml
<QueryList>
  <Query Id="0">
    <Select Path="Security">*[System[(EventID='4663')]]</Select>
  </Query>
</QueryList>
```

andrewkroh:
#1218 adds the ability to do simple filtering based on event_id, provider, and level.

Before adding the ability to specify arbitrary XML queries which can read from multiple logs, we need to change how "bookmarks" are persisted. Currently "bookmarks" are limited to a single event log so if a query read from multiple logs we wouldn't be able to persist all of the read positions. So a change needs to be made to allow writing the read position of all logs involved in a query (probably we will just persist the Windows XML bookmark).

Once that change is made then we can allow XML queries to be used.

prehor:
We would like to switch to Winlogbeat but without better support for event filtering unfortunately we can not. On the other hand, we do not need the full functionality of XML filtering, which requires substantial modification in Winlogbet, but we would be satisfied with the XPath Query used for individual channel events:

```
winlogbeat:
  event_logs:
    - Name: Security
      select:>
        * [System [(EventID=4624 or EventID=4625 or EventID=4648) and
        (Level=0 or Level = 4) and
        Provider [@ Name = 'Microsoft-Windows-Security-Auditing']]]
      supress:>
        * [System [(EventID = 4624)]] and (
        * [EventData [Data [@ Name = "TargetUserSid"] = "S-1-5-7"]] or
        * [EventData [Data [@ Name = "TargetUserSid"] = "S-1-5-18"]] or
        * [EventData [Data [@ Name = "TargetUserSid"] = "S-1-5-19"]] or
        * [EventData [Data [@ Name = "TargetUserSid"] = "S-1-5-20"]]
        )
```

where event_logs.select could be ORed with the XPath Query created using event_logs.eventid, event_logs.level and event_logs.provider or used mutually exclusive.

ajhstn:
Hows this coming along, is using xml filtering an option yet in winlogbeat?

SpencerLN:
Has there been any additional work on this? It is a crucial feature for most security monitoring use cases.

[...]

PR #6150, opened 2018-01-23
Winlogbeat: Use bookmarks to persist last published event

adriansr:
When using the Windows Event Log API, a bookmark is used to persist the
last position read in the stream. This is a first step to support XML
queries potentially involving multiple channels (#1054). Also fixes a problem
when using Forwarded Events (#3731).

### DIFF
--- a/winlogbeat/checkpoint/checkpoint.go
+++ b/winlogbeat/checkpoint/checkpoint.go
@@ -45,6 +45,7 @@ type EventLogState struct {
 	Name         string    `yaml:"name"`
 	RecordNumber uint64    `yaml:"record_number"`
 	Timestamp    time.Time `yaml:"timestamp"`
+	Bookmark     string    `yaml:"bookmark"`
 }
 
 // NewCheckpoint creates and returns a new Checkpoint. This method loads state
@@ -156,11 +157,12 @@ func (c *Checkpoint) States() map[string]EventLogState {
 }
 
 // Persist queues the given event log state information to be written to disk.
-func (c *Checkpoint) Persist(name string, recordNumber uint64, ts time.Time) {
+func (c *Checkpoint) Persist(name string, recordNumber uint64, ts time.Time, bookmark string) {
 	c.PersistState(EventLogState{
 		Name:         name,
 		RecordNumber: recordNumber,
 		Timestamp:    ts,
+		Bookmark:     bookmark,
 	})
 }
 
--- a/winlogbeat/checkpoint/checkpoint.go
+++ b/winlogbeat/checkpoint/checkpoint.go
@@ -45,7 +45,7 @@ type EventLogState struct {
 	Name         string    `yaml:"name"`
 	RecordNumber uint64    `yaml:"record_number"`
 	Timestamp    time.Time `yaml:"timestamp"`
-	Bookmark     string    `yaml:"bookmark"`
+	Bookmark     string    `yaml:"bookmark,omitempty"`
 }
 
 // NewCheckpoint creates and returns a new Checkpoint. This method loads state
--- a/winlogbeat/eventlog/eventlog.go
+++ b/winlogbeat/eventlog/eventlog.go
@@ -41,7 +41,7 @@ type EventLog interface {
 	// Open the event log. recordNumber is the last successfully read event log
 	// record number. Read will resume from recordNumber + 1. To start reading
 	// from the first event specify a recordNumber of 0.
-	Open(recordNumber uint64) error
+	Open(state checkpoint.EventLogState) error
 
 	// Read records from the event log.
 	Read() ([]Record, error)
@@ -56,8 +56,9 @@ type EventLog interface {
 // Record represents a single event from the log.
 type Record struct {
 	sys.Event
-	API string // The event log API type used to read the record.
-	XML string // XML representation of the event.
+	API    string                   // The event log API type used to read the record.
+	XML    string                   // XML representation of the event.
+	Offset checkpoint.EventLogState // Position of the record within its source stream.
 }
 
 // ToMapStr returns a new MapStr containing the data from this Record.
@@ -112,11 +113,7 @@ func (e Record) ToEvent() beat.Event {
 	return beat.Event{
 		Timestamp: e.TimeCreated.SystemTime,
 		Fields:    m,
-		Private: checkpoint.EventLogState{
-			Name:         e.Channel,
-			RecordNumber: e.RecordID,
-			Timestamp:    e.TimeCreated.SystemTime,
-		},
+		Private:   e.Offset,
 	}
 }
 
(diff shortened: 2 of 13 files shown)


adriansr:
jenkins, test it

adriansr:
Jenkins, test it and not fail this time please

winlogbeat/eventlog/wineventlog.go
```diff
@@ -89,10 +91,16 @@ func (l *winEventLog) Name() string {
...
 
-func (l *winEventLog) Open(recordNumber uint64) error {
-	bookmark, err := win.CreateBookmark(l.channelName, recordNumber)
-	if err != nil {
-		return err
+func (l *winEventLog) Open(state checkpoint.EventLogState) error {
+	var bookmark win.EvtHandle
+	var err error
+	if len(state.Bookmark) > 0 {
+		bookmark, err = win.CreateBookmarkFromXML(state.Bookmark)
```
andrewkroh: This is missing an error check. Looks like you intended to move the error check below outside the `if/else`.

winlogbeat/eventlog/wineventlog.go
```diff
@@ -154,8 +162,17 @@ func (l *winEventLog) Read() ([]Record, error) {
 			incrementMetric(dropReasons, err)
 			continue
 		}
+
+		r.Offset = checkpoint.EventLogState{
+			Name:         l.channelName,
+			RecordNumber: r.RecordID,
+			Timestamp:    r.TimeCreated.SystemTime,
+		}
+		if r.Offset.Bookmark, err = l.createBookmarkFromEvent(h); err != nil {
```
andrewkroh: Any idea as to the impact of calling this for every event? Could you run these [benchmarks](https://github.com/elastic/beats/blob/e53f42d5e1a12e0a0cbbdfa9837b0e50a91bc097/winlogbeat/eventlog/bench_test.go#L17) and share the before and after results.

winlogbeat/sys/wineventlog/wineventlog_windows.go
```diff
@@ -197,9 +197,9 @@ func RenderEvent(
 // include the RenderingInfo (message). If the event is not rendered then the
 // XML will not include the message, and in this case RenderEvent should be
 // used.
-func RenderEventXML(eventHandle EvtHandle, renderBuf []byte, out io.Writer) error {
+func RenderEventXML(eventHandle EvtHandle, flag EvtRenderFlag, renderBuf []byte, out io.Writer) error {
```
andrewkroh: This function name is no longer fully reflective of the usage because it can render both events and bookmarks based on the flag. I suggest renaming it to be more general such as `RenderXML` or create `RenderEventXML` and `RenderBookmarkXML` and have them call this function (with it being renamed to something like `renderXML`).

winlogbeat/eventlog/wineventlog.go
```diff
@@ -12,6 +12,8 @@ import (
 	"github.com/pkg/errors"
 	"golang.org/x/sys/windows"
 
+	"github.com/elastic/beats/winlogbeat/checkpoint"
```
andrewkroh: Group this import with the other elastic imports.

winlogbeat/checkpoint/checkpoint.go
```diff
@@ -45,6 +45,7 @@ type EventLogState struct {
 	Name         string    `yaml:"name"`
 	RecordNumber uint64    `yaml:"record_number"`
 	Timestamp    time.Time `yaml:"timestamp"`
+	Bookmark     string    `yaml:"bookmark"`
```
andrewkroh: Since this is optional for the eventlog API add `,omitempty` to the yaml tag.

```
update_time: 2018-01-24T16:05:31.2972263Z
event_logs:
- name: WinlogbeatTestPython
  record_number: 37134
  timestamp: 2018-01-24T16:05:30Z
  bookmark: ""
```

[source](https://beats-ci.elastic.co/job/elastic+beats+pull-request+multijob-windows/2929/beat=winlogbeat,label=windows/artifact/src/github.com/elastic/beats/winlogbeat/build/system-tests/run/test_eventlogging.Test.test_registry_data/data/.winlogbeat.yml/*view*/)

andrewkroh:
There doesn't appear to be any existing test that verifies Winlogbeat resumes from the position specified in the registry. I think this would be a good test to add. Something like

- write an event to the log
- run WLB
- stop WLB after the event is read
- write more events to the log
- run WLB
- assert that the first event has the correct record number

winlogbeat/eventlog/eventlog.go
```diff
@@ -41,7 +41,7 @@ type EventLog interface {
 	// Open the event log. recordNumber is the last successfully read event log
```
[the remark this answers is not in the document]
adriansr: done

winlogbeat/checkpoint/checkpoint.go
adriansr: done

winlogbeat/eventlog/eventlogging.go
```diff
@@ -9,6 +9,8 @@ import (
 
 	"github.com/joeshaw/multierror"
 
+	"github.com/elastic/beats/winlogbeat/checkpoint"
```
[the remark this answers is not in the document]
adriansr: done

winlogbeat/eventlog/wineventlog.go
adriansr: done

winlogbeat/eventlog/wineventlog.go
adriansr: good catch

winlogbeat/sys/wineventlog/wineventlog_windows.go
adriansr: done

winlogbeat/eventlog/wineventlog.go
adriansr: I did run some tests at first and method was taking less than 1ms to complete. However, I also found out that every event was persisted so it was necessary to always create a bookmark anyway. Maybe there was something wrong in my configuration, a side-effect of running with -N.

The benchmark results:

Before:
```
ok      github.com/elastic/beats/winlogbeat/eventlog    88.679s
ok      github.com/elastic/beats/winlogbeat/eventlog    89.105s
ok      github.com/elastic/beats/winlogbeat/eventlog    88.853s
ok      github.com/elastic/beats/winlogbeat/eventlog    89.293s
ok      github.com/elastic/beats/winlogbeat/eventlog    89.383s
```

After:
```
ok      github.com/elastic/beats/winlogbeat/eventlog    91.626s
ok      github.com/elastic/beats/winlogbeat/eventlog    90.079s
ok      github.com/elastic/beats/winlogbeat/eventlog    89.980s
ok      github.com/elastic/beats/winlogbeat/eventlog    90.663s
ok      github.com/elastic/beats/winlogbeat/eventlog    90.641s
```

It is a 1.7% overhead.

I am fine with having the bookmark XML calculated only when needed

winlogbeat/sys/wineventlog/wineventlog_windows.go
adriansr: done

PR #6153, opened 2018-01-23
Suppress system load error messages on Windows

kvch:
The following error message showed up in the logs of a Beat on Windows, because system load is not implemented:
```
{
 "level": "error",
 "timestamp": "2018-01-23T09:39:50.787-0800",
 "caller": "instance/metrics.go:147",
 "message": "Error retrieving load average: not implemented on windows"
}
```

As it is not supported by Windows, the error message is unnecessary. The metrics is not included in the monitoring events of a Beat on a Windows. So the expected `N/A` can show up in the charts.

I separated the `reportSystemLoadAverage` function of metrics into three parts. One includes the function that registers the reporter function built for all supported platforms. But sending the event is only implemented on Darwin and Linux, thus it's in `metrics_darwin_linux`. For Windows the function does not do anything.

### DIFF
--- a/libbeat/cmd/instance/metrics.go
+++ b/libbeat/cmd/instance/metrics.go
@@ -30,8 +30,10 @@ func init() {
 	monitoring.NewFunc(beatMetrics, "info", reportInfo, monitoring.Report)
 
 	systemMetrics := monitoring.Default.NewRegistry("system")
-	monitoring.NewFunc(systemMetrics, "load", reportSystemLoadAverage, monitoring.Report)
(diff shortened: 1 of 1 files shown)


ph:
@kvch The `make check` command fails for the metrics_darwin_linux.go file.

andrewkroh:
Did you consider a runtime check instead of build tags? Like

```
if runtime.GOOS != "windows" {
    // register load function
}
```

kvch:
@andrewkroh I did previously. But discarded the idea, because I did not want to mix build tags with runtime checks. However, it's a smaller change and hard to miss when changing the code, so I went your way.

PR #6157, opened 2018-01-23
generate module.yml for create-fileset (#5657)

ewgRa:
This PR is improvement for #5657 and some bugfixes

1. I realize that check isExists not working (at least for Ubuntu). Even if module exists, check not passed, and I haven't error: module already exists. Because os.Stat return nil error if directory exists. So, I changed it to !isNotExists that works. Would be nice if someone check it on their systems.

2. Code looks strange, it suppose to exit if module exists, and checking that fileset exists make not sense. Kind of unreachable code. Would be nice to have explanations. Actually I see there improvement, if module exists, just add another fileset, even mazbe automatically add it to module.yml.

3. Module level files doc* and fields.yml according to docs and already exists modules must be placed to _meta folder, so I change this part.

4. I change call of copyTemplatesToDest, since everywhere there is fileset param, but it is replaced to empty string, that looks strange.

Please review, if there is some thing must be reverted, feel free to write it. Would be nice with explanations.


filebeat/scripts/generator/fileset/main.go
```diff
@@ -29,7 +29,7 @@ func copyTemplatesToDest(templatesPath, name, filesetPath, module, fileset strin
 
 func generateModule(module, fileset, modulePath, beatsPath string) error {
 	p := path.Join(modulePath, "module", module)
-	if _, err := os.Stat(p); os.IsExist(err) {
+	if _, err := os.Stat(p); !os.IsNotExist(err) {
 		return fmt.Errorf("module already exists: %s at %s", module, p)
```
kvch: Could add code to return from here silently?

kvch:
1. I think the problem is rather returning with error if the module exists. If the module exists we can assume that the user wants to add a new fileset to an existing module. So we should return silently from the function, thus fileset can be generated and the existing module files are not overwritten by the script. That's why it was never caught. Could you change your code accordingly? Thanks for catching it!

2. The code should not exit if the module exists. If the module exists, it should continue with creating the fileset. See the point above.

3. The templates are residing under `scripts/module`, so you should not change the source of those files.

4. Fileset is not passed to `copyTemplatesToDest` when the destination is the module root, not the fileset.

filebeat/scripts/generator/fileset/main.go
```diff
@@ -40,9 +40,9 @@ func generateModule(module, fileset, modulePath, beatsPath string) error {
 	}
 
 	templatesPath := path.Join(beatsPath, "scripts", "module")
-	filesToCopy := []string{path.Join("fields.yml"), path.Join("docs.asciidoc")}
```
kvch: Both `fields.yml` and `docs.asciidoc` templates are under `scripts/module`. You should not change this path, otherwise templates won't be found.

filebeat/scripts/generator/fileset/main.go
```diff
@@ -40,9 +40,9 @@ func generateModule(module, fileset, modulePath, beatsPath string) error {
...
+	filesToCopy := []string{path.Join("_meta", "fields.yml"), path.Join("_meta", "docs.asciidoc"), path.Join("module.yml")}
 	for _, f := range filesToCopy {
-		err := copyTemplatesToDest(templatesPath, f, p, module, "")
```
kvch: Do not add fileset here, because the templates will be copied to an incorrect destination. If it helps with understanding the code a separate `copyModuleTemplates` and `copyFilesetTemplates` could be created.

filebeat/scripts/generator/fileset/main.go
kvch: My bad. You seem to be correct here.

ewgRa: Yes, in docs https://www.elastic.co/guide/en/beats/devguide/current/filebeat-modules-devguide.html and exists modules it is in _meta everywhere. So, I think generated also must be in _meta. Please confirm that correction right and I don't need to revert it.

filebeat/scripts/generator/fileset/main.go
ewgRa: Check please copyTemplatesToDest, this is not a path, this is fileset name, and it is used like:
> c = bytes.Replace(c, []byte("{fileset}"), []byte(fileset), -1)

So, there is nothing related to copy files, they will be copied in a right place even with this change. Only one change, that for example in docs* file {fileset} will be replaced right. Now it is replaced to "".

Can you please confirm that I need revert it? Because even with your comment looks like this is right change for me.

filebeat/scripts/generator/fileset/main.go
kvch: I tested you script. It's correct.

filebeat/scripts/generator/fileset/main.go
kvch: It's right, too.

filebeat/scripts/generator/fileset/main.go
ewgRa: Done.

ewgRa:
What do you think, if module already created and user create new fileset, does we need automatically update module.yml and add there this new fileset? Or better not touch it?
For me looks user friendly to do this, but in another case there can be unexpected changes for user, since it think that only new fileset will be created.

kvch:
How do you want to update `module.yml` when generating a fileset? By the time a fileset is generated there isn't any dashboards to put in the list. Also, the id of the bashboard is given by the user AFAIK.
If there is a reasonable way to do it, I am all for it. The more stuff is automated, the happier I am.

ewgRa:
We can help user and suggest some kind of default name for dashboard and so on, based on pattern that we use when create module: "id: Filebeat-{module}-{fileset}-Dashboard" - it is from module.yml.

How to add it - this is just a technical thing.
We can parse module.yml and add it to specific place, or keep it simple and make just "cat fileset_template | sed... >> module.yml"

Is I'm thinking right?

Also there is idea make a two commands: create-module, and create-fileset. Splitting this for to commands will add more transparency and reduce complexity.

kvch:
Yes, it's true. If it's added, I suggest to add a notice for users to make sure they set properly the name of the dashboards. Are you going to add this changes to this PR?

Also, splitting it into separate commands is a good idea. It does have the advantages you mentioned. However, I am wondering if it's making the process less comfortable by requiring the user to run two commands instead of two.

ewgRa:
@kvch I've add this modification of module.yml when new fileset is created, please review. Also I've add notice about checking such modification manually. Hope it is not increase PR complexity. Further changes I think better made as followup changes. There is already changes kind of out of scope described task.

> I am wondering if it's making the process less comfortable by requiring the user to run two commands instead of two

For me personally it is vice versa. When we will create fileset and there is no module created, we can give user good hint - please create module first. When user create module - we can give good hint - now you can create your filesets. Step-by-step process always better than all-in-one. And from understanding perspective, and from maintenance (code) perspective.

kvch:
Great! Thanks for updating the PR. Also, could you please update the docs Filebeat module development? It's in `docs/devguide/modules-dev-guide.asciidoc`. A few sentences should be sufficient to cover these changes.

I see your point. Are you open to implement those changes? It might require a refactor as module and fileset generator are tightly coupled.

ewgRa:
@kvch docs about module.yml added, please check.


> Are you open to implement those changes?

Yes, I will write it to my todo list as followup changes. I have there one thing that I want to do first, and after that I can return to create-fileset command changes. I think it is not so critical and can wait. I haven't here good estimations, but definitely would be interesting to make PR for this sooner or later.

kvch:
Thanks for your contribution. Looking forward to your next PR.

CHANGELOG.asciidoc
```diff
@@ -70,6 +70,8 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
 
 *Filebeat*
 
+- Right check of module/fileset exists for create-fileset command {issue}5657[5657]
```
ruflin: @kvch I think we should remove these lines from the changelog again as we normally don't up dev related stuff in here. Perhaps we should put it in an other category? We discussed it in the past where to put it but never came to a conclusion :-(

ewgRa: @ruflin do you have a link on discussion?

PR #6158, opened 2018-01-24
Add pct calculated fields for Pod and container CPU and memory usages

### LINKED ISSUE #6124: [Metricbeat] kubernetes.container.cpu.limit.nanocores should default to the nodes total nanocores
If admins don't set a limit for `nanocores` then Metricbeat should set `kubernetes.container.cpu.limit.nanocores` to the total nanocores for the node.

### LINKED ISSUE #6125: [Metricbeat] kubernetes.container.memory.available.bytes should default to the system allocatable memory.
When admins don't set `kubernetes.container.memory.available.bytes` it should default to  `kubernetes.node.memory.allocatable.bytes`

exekias:
This PR adds the following set of calculated metrics for Pods and containers:

- `kubernetes.container.cpu.usage.node.pct`
- `kubernetes.container.cpu.usage.limit.pct`
- `kubernetes.container.memory.usage.node.pct`
- `kubernetes.container.memory.usage.limit.pct`
- `kubernetes.pod.cpu.usage.nanocores`
- `kubernetes.pod.cpu.usage.node.pct`
- `kubernetes.pod.cpu.usage.limit.pct`
- `kubernetes.pod.memory.usage.bytes`
- `kubernetes.pod.memory.usage.node.pct`
- `kubernetes.pod.memory.usage.limit.pct`

As the source of data to calculate these values comes from different places (kubelet & kube-state-metrics) I added a mechanism to share some performance metrics through an in-memory cache. For that reason, these new metrics will only be available when `state_*` metricsets are enabled.

After talking with @simianhacker, I think this:
closes #6125
closes #6124

Please @simianhacker have a look and let me know if you need anything else

### DIFF
--- a/metricbeat/module/kubernetes/util/metrics_cache.go
+++ b/metricbeat/module/kubernetes/util/metrics_cache.go
@@ -1,0 +1,106 @@
+package util
+
+import (
+	"sync"
+	"time"
+)
+
+// PerfMetrics stores known metrics from Kubernetes nodes and containers
+var PerfMetrics *PerfMetricsCache
+
+const defaultTimeout = 120 * time.Second
+
+func init() {
+	PerfMetrics = NewPerfMetricsCache()
+}
+
+// NewPerfMetricsCache initializes and returns a new PerfMetricsCache
+func NewPerfMetricsCache() *PerfMetricsCache {
+	return &PerfMetricsCache{
+		NodeMemAllocatable:   newValueMap(defaultTimeout),
+		NodeCoresAllocatable: newValueMap(defaultTimeout),
+
+		ContainerMemLimit:   newValueMap(defaultTimeout),
+		ContainerCoresLimit: newValueMap(defaultTimeout),
+	}
+}
+
+// PerfMetricsCache stores known metrics from Kubernetes nodes and containers
+type PerfMetricsCache struct {
+	mutex                sync.RWMutex
+	NodeMemAllocatable   *valueMap
+	NodeCoresAllocatable *valueMap
+
+	ContainerMemLimit   *valueMap
+	ContainerCoresLimit *valueMap
+}
+
+func newValueMap(timeout time.Duration) *valueMap {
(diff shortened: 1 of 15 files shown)


ruflin:
@exekias Tests somehow fail on darwin. Probably need to skip it.

metricbeat/module/kubernetes/util/metrics_cache.go
```diff
@@ -0,0 +1,106 @@
...
+
+import (
+	"sync"
+	"time"
+)
+
+// PerfMetrics stores known metrics from Kubernetes nodes and containers
+var PerfMetrics *PerfMetricsCache
+
+const defaultTimeout = 120 * time.Second
```
ruflin: Should we make this configurable?

exekias: It's a little tricky cause it's a global cache. I used a somehow safe value, I don't think someone needs to worry about it, what do you think?

metricbeat/module/kubernetes/util/metrics_cache.go
```diff
@@ -0,0 +1,106 @@
+package util
+
+import (
+	"sync"
+	"time"
+)
+
+// PerfMetrics stores known metrics from Kubernetes nodes and containers
+var PerfMetrics *PerfMetricsCache
```
walktall: How about init it directly here
`var PerfMetrics = NewPerfMetricsCache()`

exekias: Sounds good to me, pushed the change, thanks!

exekias:
Thanks for the reviews! @ruflin this one should be ready

metricbeat/module/kubernetes/util/metrics_cache.go
ruflin: Let's do it until it breaks :-)

PR #6159, opened 2018-01-24
kubernetes watcher refactor

walktall:
This PR comes from https://github.com/elastic/beats/pull/6117

It does following items:

1. Update k8s vendor to get an abstract api for watching and listing.

2. Refactor watcher. The style is from client-go informer so that we can port to client-go without much effort in future. Also, it removes side effect in watcher, such as delay delete event and cache the resources, these side-effect should be implemented by caller who has that kind of requirement I think.

3. Although as mentioned by @exekias , bug fix for querying pod should be put in another pr, but I found it is hard to do that since moving cache out from watcher is part of refactor...

4. Event metricset in metricbeat now is using new refactored watcher.

@exekias PTAL~


libbeat/common/kubernetes/watcher.go
```diff
@@ -2,260 +2,189 @@ package kubernetes
...
 		return err
 	}
 
-	p.Lock()
-	for _, apiPod := range pods.Items {
-		pod := GetPod(apiPod)
-		p.pods[pod.Metadata.UID] = pod
+	// ugly reflect since ResourceList does not exposed Items
+	// can be avoid by using k8s.io/client-go
+	list := reflect.ValueOf(w.resourceList).Elem().FieldByName("Items")
```
exekias: You can avoid this if you create a new interface for resource list, and write adapters to get the Items from PodList and EventList, what do you think?

exekias:
ok to test

libbeat/common/kubernetes/watcher.go
walktall: Good idea! I will make a try. Thanks!

ruflin:
@walktall This will need a rebase.

walktall:
@ruflin rebased

@exekias made some changes to avoid reflect, PTAL.

PR #6162, opened 2018-01-24
log error message when no topic could be selected for kafka output

crazyvikas23:
When kafka is selected as output and `topic` key is absent from kafka configuration and a topic could not be selected from the list of topics provided by `topics` key, no error message is shown. This commit fixes adn

### DIFF
--- a/libbeat/outputs/kafka/client.go
+++ b/libbeat/outputs/kafka/client.go
@@ -1,6 +1,7 @@
 package kafka
 
 import (
+	"errors"
 	"fmt"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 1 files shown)


karmi:
Hi @crazyvikas23, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/beats/pull/6162.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

crazyvikas23:
@karmi  fixed the commit to contain my github email address

kvch:
jenkins test it

urso:
Thanks for improving the error handling.

PR #6165, opened 2018-01-24
Add a dev-tools/set_docs_version script

tsg:
Similar to the `dev-toosl/set_version` script. At the moment, it
edits the `version.asciidoc` file and runs `make update`.

### DIFF
--- a/dev-tools/set_docs_version
+++ b/dev-tools/set_docs_version
@@ -1,0 +1,28 @@
+#!/usr/bin/env python
+import argparse
+from subprocess import check_call
+
+def main():
+    parser = argparse.ArgumentParser(
+        description="Used to set the current docs version. Doesn't commit changes.")
+    parser.add_argument("version",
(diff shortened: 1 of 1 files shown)


dev-tools/set_docs_version
```diff
@@ -0,0 +1,28 @@
...
+def main():
+    parser = argparse.ArgumentParser(
+        description="Used to set the current docs version. Doesn't commit changes.")
+    parser.add_argument("version",
+                        help="The new docs version")
+    args = parser.parse_args()
+    version = args.version
+
+    # make sure we have no dirty files in this branch (might throw off `make update`)
+    #check_call("git clean -dfx", shell=True)
```
ph: This looks like a good idea, any reason why its commented? If we don't plan to have this feature we should just remove the code.

tsg:
Good catch, I pushed another commit.

PR #6166, opened 2018-01-24
Correctly propagate the error to the user when a `store` fails

ph:
If storing a key to the keystore fails the error was not send back to
the user and the incomplete keystore was persisted to disk.

### DIFF
--- a/libbeat/cmd/keystore.go
+++ b/libbeat/cmd/keystore.go
@@ -196,14 +196,13 @@ func addKey(store keystore.Keystore, keys []string, force, stdin bool) error {
 			return fmt.Errorf("could not read value from the input, error: %s", err)
 		}
 	}
-
-	store.Store(key, keyValue)
-	err = store.Save()
-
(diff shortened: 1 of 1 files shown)


ph:
waiting for ci, before marking it as ready for review.

libbeat/cmd/keystore.go
```diff
@@ -196,14 +196,13 @@ func addKey(store keystore.Keystore, keys []string, force, stdin bool) error {
...
-	err = store.Save()
-
-	if err == nil {
-		fmt.Println("Successfully updated the keystore")
+	if err = store.Store(key, keyValue); err != nil {
+		return fmt.Errorf("could not add the key in the keystore, error: %s", err)
+	}
+	if err = store.Save(); err != nil {
+		return fmt.Errorf("fail to save the keystore: %s", err)
 	} else {
```
ruflin: I think that else here is now obsolete.

PR #6168, opened 2018-01-24
carry over the 'worker' default value from load balancing

tedder:
This default was missing; it's [listed on the load balancing page as having a default of 1.](https://www.elastic.co/guide/en/beats/filebeat/current/load-balancing.html)


ruflin:
@tedder It would be great if you could open PR's against master and then we can do the backporting.

tedder:
done! closing this one.

PR #6170, opened 2018-01-24
Provide 32-bit Sample Config for Auditbeat

### LINKED ISSUE #6156: Provide 32-bit Sample Config for Auditbeat
The sample rules provided in the auditbeat.yml are specific to 64-bit architectures. Since we provide 32-bit packages we should customize the rules for those packages.

Because the config file is generated based on GOOS and GOARCH this should be easy to accomplish.

ripcurld0:
Generate auditbeat sample rules according to the given CPU architecture: `amd64` or `i386`.

By default, this value is set to the running machine's CPU architecture which is retrieved from `runtime.GOARCH`.

closes #6156

### DIFF
--- a/auditbeat/scripts/generate_config.go
+++ b/auditbeat/scripts/generate_config.go
@@ -18,6 +18,7 @@ const defaultGlob = "module/*/_meta/config*.yml.tpl"
 
 var (
 	goos      = flag.String("os", runtime.GOOS, "generate config specific to the specified operating system")
+	goarch    = flag.String("arch", runtime.GOARCH, "generate config specific to the specified CPU architecture")
 	reference = flag.Bool("ref", false, "generate a reference config")
 	concat    = flag.Bool("concat", false, "concatenate all configs instead writing individual files")
 )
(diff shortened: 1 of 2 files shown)


auditbeat/scripts/generate_config.go
```diff
@@ -40,7 +41,17 @@ func getConfig(file string) ([]byte, error) {
...
+	switch *goarch {
+	case "i386":
+		arch = "32"
+	case "amd64":
+		arch = "64"
+	default:
+		return nil, fmt.Errorf("supporting only i386 and amd64 architecture")
+	}
 	data := map[string]interface{}{
+		"goarch":    arch,
```
andrewkroh: Perhaps this should be called `arch_bits` or something other than `goarch` since it's not exactly the GOARCH value?

ripcurld0: Make sense!

andrewkroh:
jenkins, test it

ripcurld0:
Let me fix that...

andrewkroh:
jenkins, test it

ripcurld0:
beats-ci failures aren't related to code:
> compile: writing output: write $WORK\github.com\elastic\beats\libbeat\logp\configure.a: There is **not enough space on the disk**.
> fatal: write error: **No space left on device**
> Out-File : There is **not enough space on the disk**

andrewkroh:
Thanks again! Are you interested in updating the packaging to make use of this?

Current when the packages are constructed they look for `{{.beat_name}}-linux.yml` (no arch in the name) which is generated by the auditbeat Makefile.

https://github.com/elastic/beats/blob/df9ad146e282f716211f614478088bc0b9c8f449/dev-tools/packer/platforms/debian/run.sh.j2#L35

https://github.com/elastic/beats/blob/df9ad146e282f716211f614478088bc0b9c8f449/auditbeat/Makefile#L35-L43

So I think the run.sh.j2 files need to be updated to check if an architecture specific file exists (e.g. `{{.beat_name}}-linux-{{.arch}}.yml`) and if so use it, otherwise fallback to `{{.beat_name}}-linux.yml`. The fallback is required in order to not break packaging for the other Beats.

Then in the auditbeat Makefile update it to generate the architecture specific config files (only on Linux).

PR #6172, opened 2018-01-25
Add short ID matching to add_docker_metadata

### LINKED ISSUE #6092: [Feature Request] add_docker_metadata should accept short ids for containers
I have a case where I'm trying to write out structured data to a log file in a container on a shared tmpfs volume with the filebeat container, and I'd like to attach the docker metadata. I'm writing the logs into the path /logs/${HOSTNAME}/*.log, in order to extract the container id, since docker will set the hostname to the short id of the container, so the short id is fairly accessible without having to jump through a lot of extra hoops. Unfortunately, when I tried using the add_docker_metadata.match_source option to pull out the id, it wouldn't match, and upon further investigation it looks like it will only match the long ids.

Since the short id is easily accessible from within a container, I think it would be useful to match against it. Is this something that's possible?

ripcurld0:
closes #6092

### DIFF
--- a/libbeat/common/docker/watcher.go
+++ b/libbeat/common/docker/watcher.go
@@ -18,7 +18,10 @@ import (
 )
 
 // Select Docker API version
-const dockerAPIVersion = "1.22"
+const (
+	dockerAPIVersion = "1.22"
+	shortIDLen       = 12
+)
 
 // Watcher reads docker events and keeps a list of known containers
 type Watcher interface {
@@ -59,6 +62,7 @@ type watcher struct {
 	lastValidTimestamp int64
 	stopped            sync.WaitGroup
 	bus                bus.Bus
+	shortID            bool // whether to store short ID in "containers" too
 }
 
 // Container info retrieved by the watcher
@@ -77,10 +81,11 @@ type Client interface {
 	Events(ctx context.Context, options types.EventsOptions) (<-chan events.Message, <-chan error)
 }
 
-type WatcherConstructor func(host string, tls *TLSConfig) (Watcher, error)
+// WatcherConstructor represent a function that creates a new Watcher from giving parameters
+type WatcherConstructor func(host string, tls *TLSConfig, storeShortID bool) (Watcher, error)
 
 // NewWatcher returns a watcher running for the given settings
-func NewWatcher(host string, tls *TLSConfig) (Watcher, error) {
+func NewWatcher(host string, tls *TLSConfig, storeShortID bool) (Watcher, error) {
 	var httpClient *http.Client
 	if tls != nil {
 		options := tlsconfig.Options{
@@ -106,10 +111,11 @@ func NewWatcher(host string, tls *TLSConfig) (Watcher, error) {
 		return nil, err
 	}
 
-	return NewWatcherWithClient(client, 60*time.Second)
+	return NewWatcherWithClient(client, 60*time.Second, storeShortID)
 }
 
-func NewWatcherWithClient(client Client, cleanupTimeout time.Duration) (*watcher, error) {
+// NewWatcherWithClient creates a new Watcher from a given Docker client
+func NewWatcherWithClient(client Client, cleanupTimeout time.Duration, storeShortID bool) (Watcher, error) {
 	ctx, cancel := context.WithCancel(context.Background())
 	return &watcher{
 		client:         client,
@@ -119,6 +125,7 @@ func NewWatcherWithClient(client Client, cleanupTimeout time.Duration) (*watcher
 		deleted:        make(map[string]time.Time),
 		cleanupTimeout: cleanupTimeout,
 		bus:            bus.New("docker"),
+		shortID:        storeShortID,
 	}, nil
 }
 
@@ -126,13 +133,17 @@ func NewWatcherWithClient(client Client, cleanupTimeout time.Duration) (*watcher
 func (w *watcher) Container(ID string) *Container {
 	w.RLock()
 	container := w.containers[ID]
-	_, ok := w.deleted[ID]
+	if container == nil {
+		w.RUnlock()
+		return nil
+	}
+	_, ok := w.deleted[container.ID]
 	w.RUnlock()
 
 	// Update last access time if it's deleted
 	if ok {
 		w.Lock()
-		w.deleted[ID] = time.Now()
+		w.deleted[container.ID] = time.Now()
 		w.Unlock()
 	}
 
@@ -145,7 +156,9 @@ func (w *watcher) Containers() map[string]*Container {
 	defer w.RUnlock()
 	res := make(map[string]*Container)
 	for k, v := range w.containers {
-		res[k] = v
+		if !w.shortID || len(k) != shortIDLen {
+			res[k] = v
+		}
 	}
 	return res
 }
@@ -165,6 +178,9 @@ func (w *watcher) Start() error {
 
 	for _, c := range containers {
 		w.containers[c.ID] = c
+		if w.shortID {
+			w.containers[c.ID[:shortIDLen]] = c
+		}
 	}
 
 	// Emit all start events (avoid blocking if the bus get's blocked)
@@ -223,6 +239,9 @@ func (w *watcher) watch() {
 
 					w.Lock()
 					w.containers[event.Actor.ID] = container
+					if w.shortID {
+						w.containers[event.Actor.ID[:shortIDLen]] = container
+					}
 					// un-delete if it's flagged (in case of update or recreation)
 					delete(w.deleted, event.Actor.ID)
 					w.Unlock()
@@ -326,6 +345,9 @@ func (w *watcher) cleanupWorker() {
 			for _, key := range toDelete {
 				delete(w.deleted, key)
 				delete(w.containers, key)
+				if w.shortID {
+					delete(w.containers, key[:shortIDLen])
+				}
 			}
 			w.Unlock()
 		}
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -240,6 +240,7 @@ The list below covers the changes between 6.0.0-rc2 and 6.0.0 GA only.
 *Filebeat*
 
 - Add Kubernetes manifests to deploy Filebeat. {pull}5349[5349]
+- Add container short ID matching to add_docker_metadata. {pull}6172[6172]
 
 *Metricbeat*
 
--- a/auditbeat/auditbeat.reference.yml
+++ b/auditbeat/auditbeat.reference.yml
@@ -205,6 +205,7 @@ auditbeat.modules:
 #    match_pids: ["process.pid", "process.ppid"]
 #    match_source: true
 #    match_source_index: 4
+#    match_short_id: false
 #    cleanup_timeout: 60
 #    # To connect to Docker over TLS you must specify a client and CA certificate.
 #    #ssl:
--- a/filebeat/filebeat.reference.yml
+++ b/filebeat/filebeat.reference.yml
@@ -653,6 +653,7 @@ filebeat.prospectors:
 #    match_pids: ["process.pid", "process.ppid"]
 #    match_source: true
 #    match_source_index: 4
+#    match_short_id: false
 #    cleanup_timeout: 60
 #    # To connect to Docker over TLS you must specify a client and CA certificate.
 #    #ssl:
--- a/heartbeat/heartbeat.reference.yml
+++ b/heartbeat/heartbeat.reference.yml
@@ -314,6 +314,7 @@ heartbeat.scheduler:
 #    match_pids: ["process.pid", "process.ppid"]
 #    match_source: true
 #    match_source_index: 4
+#    match_short_id: false
 #    cleanup_timeout: 60
 #    # To connect to Docker over TLS you must specify a client and CA certificate.
 #    #ssl:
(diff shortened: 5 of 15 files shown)


ripcurld0:
@exekias sure! That's what I meant to do (hence "WIP" in the title).
No problem, I will update `CHANGELOG.asciidoc` too

exekias:
Awesome, I totally missed the WIP part :innocent:

libbeat/common/docker/watcher.go
```diff
@@ -125,7 +128,7 @@ func NewWatcherWithClient(client Client, cleanupTimeout time.Duration) (*watcher
 // Container returns the running container with the given ID or nil if unknown
 func (w *watcher) Container(ID string) *Container {
 	w.RLock()
-	container := w.containers[ID]
+	container := w.containers[ID[:truncateLen]]
```
ruflin: I wonder if we should make this a config option or check always for the truncated id. I assume also with 12 digits the likelyhood of overlap is very low.

exekias: It's an interesting point, *in theory*, chances of collision are round 0.000059%, which makes it very unlikely. That said, someone reported one there:
https://github.com/moby/moby/issues/28260#issuecomment-259820264

ripcurld0: That's really interesting. I didn't think about it, because I thought it was very-very-very-very uncommon to get a collision. I guess we should consider printing a warning message if we get more than one containers for a specific ID. Thanks for noticing this.

ruflin: My initial thought was that by default we should use the full `id` as there are probably reason the id is that long. The shorter id my assumption is more for user consumption but as it seems it's also around it some other places. 

If a short id is found / provided it should be best effort to match it to a long one with something like `contains`. The problem is that this could become pretty inefficient as it's not a direct match in the array which makes the implementation you did pretty nice.

It would be great if we could find an efficient method to internally store the full id but still match the short id when needed.

Thought: What if we store both?

ripcurld0: You can have another map `map[string]*Container` that contains short ID as a key.
When we detect that the given ID is in length 12 then we use that map otherwise use the original map.

The disadvantage here is obviously a space-time trade-off.

exekias: What about keeping current code and make truncate length configurable? That would support the short id use case, but you have to enable it

ripcurld0: fine by me. @ruflin WDYT?

andrewkroh: If the short IDs are consistently 12 characters then adding a secondary entry to the map SGTM. But if we need to do a prefix search where the prefixes are of arbitrary length then maybe consider a different data structure.

ripcurld0: @andrewkroh there are going to be only two kind of IDs:
* short ID (length between 1 and 64 theoretically)
* ID (length of 64)

In other words, the length of a key in the map is `64` or `12`. You won't find a key that its length is `23`.

ruflin: I would also just add a second entry to the map so on the lookup side it still works as before. I would not expect too much overhead as this map will probably not contain more then 1000 entries?

If it becomes an overhead, I would introduce a config option to enable short id's and have it not store in the map if it's off, but still use the same map if it's on. But we can do that if it's causes any issue.

libbeat/common/docker/watcher.go
```diff
@@ -222,7 +225,11 @@ func (w *watcher) watch() {
 					container := containers[0]
 
 					w.Lock()
-					w.containers[event.Actor.ID] = container
+					// Docker uses very frequently the truncated container ID to represent a container.
+					// For example, by default container's hostname is set to the truncated container ID.
+					// To allow matching by truncated and non-truncated container IDs the map keys
+					// are the truncated version.
+					w.containers[event.Actor.ID[:truncateLen]] = container
```
ruflin: If we store both, I assume this line would be changed to:

```
w.containers[event.Actor.ID] = container
w.containers[event.Actor.ID[:truncateLen]] = container
```

ruflin:
@ripcurld0 Do you have any time on your end to push this forward? I think it would be a great feature to have in.

ripcurld0:
@ruflin I am going to push changes today.
Sorry for taking me long. 😊

libbeat/common/docker/watcher.go
```diff
@@ -106,10 +110,10 @@ func NewWatcher(host string, tls *TLSConfig) (Watcher, error) {
 		return nil, err
 	}
 
-	return NewWatcherWithClient(client, 60*time.Second)
+	return NewWatcherWithClient(client, 60*time.Second, storeShortID)
 }
 
-func NewWatcherWithClient(client Client, cleanupTimeout time.Duration) (*watcher, error) {
+func NewWatcherWithClient(client Client, cleanupTimeout time.Duration, storeShortID bool) (*watcher, error) {
```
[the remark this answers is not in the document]
ripcurld0: This is actually true, we might want to return "Watcher" (the interface) but I am afraid there are function that are not included in the interface which other packages would like to use.

libbeat/common/docker/watcher.go
```diff
@@ -119,20 +123,24 @@ func NewWatcherWithClient(client Client, cleanupTimeout time.Duration) (*watcher
...
 	}, nil
 }
 
 // Container returns the running container with the given ID or nil if unknown
 func (w *watcher) Container(ID string) *Container {
 	w.RLock()
 	container := w.containers[ID]
-	_, ok := w.deleted[ID]
+	if container == nil {
+		return nil
```
exekias: :warning: Unlock won't be called in this case

ripcurld0: 😱 thanks!

ripcurld0:
I ran some manual tests on my computer.

## preparations 
- I ran `ElasticSearch` and `Kibana` (both version 6) inside containers:
```
$ docker run -d --name es -p 9200:9200 -p 9300:9300 -e "discovery.type=single-node" docker.elastic.co/elasticsearch/elasticsearch:6.1.2
$ docker run -d --name kb -p 5601:5601 --link es:elasticsearch docker.elastic.co/kibana/kibana:6.1.2
```

- Then, I created `logs` and `config` directories:
```
$ cd ~
$ mkdir {config,logs}
```

In `logs` all the containers log files are going to be stored and in `config` I am going to store `filebeat.yml`.

Lets start with the following `filebeat.yml`:
```
filebeat.prospectors:
    - type: log
      paths:
          - '/home/ripcurld0/logs/*/*'
      processors:
          - add_docker_metadata:
              match_source: false

output.elasticsearch:
    hosts: ['localhost:9200']
```
This should read all log files in `logs` and send them to ES.

- Build a container to write data to a file.
Since I can't just:
```
$ docker run -v ${HOME}/logs/:/tmp/logs busybox echo "data" > /tmp/logs/${HOSTNAME}/file.log
```
or
```
$ docker run -v ${HOME}/logs/:/tmp/logs busybox sh -c "echo data > /tmp/logs/${HOSTNAME}/file.log"
```
Because in the first one it will pipe it to my machine /tmp/logs/${HOSTNAME} and in the second one although it will put it in the container's /tmp/logs the hostname will be my machine hostname.

Thus, I had to build an image.
1) created `~/dockerfile`
```
$ mkdir ~/dockerfile
$ cd ~/dockerfile
```
2) create Dockerfile:
```
FROM busybox
ADD ./init.sh /init.sh
RUN chmod +x /init.sh
ENTRYPOINT ["/init.sh"]
```
3) create the `init.sh` script
```
#!/bin/sh
mkdir -p /tmp/logs/${HOSTNAME}
while [[ true ]]; do
    date >> /tmp/logs/${HOSTNAME}/file.log
    sleep 1s
done
```
4) build image
```
$ docker build -t fbgen ~/dockerfile/.
```

5) run a container
```
docker run  -d --rm -v /home/vagrant/logs:/tmp/logs fbgen
```

### testing

- Compiled my code:
```
$ cd ~/go/src/github.com/elastic/beats/filebeat
$ make clean && make filebeat
```

- Running my local filebeat with the given config
```
$ ./filebeat -e -c ~/configs/filebeat.yml -d "publish"
```

The results:
```
2018-02-12T12:57:43.138Z        DEBUG   [publish]       pipeline/processor.go:275       Publish event: {                                                                                       
  "@timestamp": "2018-02-12T12:57:43.138Z",                                                                                                                                                    
  "@metadata": {                               
    "beat": "filebeat",                        
    "type": "doc",                             
    "version": "7.0.0-alpha1"                  
  },                                           
  "beat": {                                    
    "name": "elastic",                         
    "hostname": "elastic",                     
    "version": "7.0.0-alpha1"                  
  },                                           
  "source": "/home/ripcurld0/logs/196f8a55cb7d/data.txt",                                        
  "offset": 39266,                             
  "message": "Mon Feb 12 12:57:42 UTC 2018",   
  "prospector": {                              
    "type": "log"                              
  },                                           
  "event": {                                   
    "type": "log"                              
  }                                            
}
```

- This seems right, now lets add `docker_metadata`:
```
cat ~/configs/filebeat.yml
filebeat.prospectors:
    - type: log
      paths:
          - '/home/ripcurld0/logs/*/*'
      processors:
          - add_docker_metadata:
              match_source: true
              match_source_index: 3

output.elasticsearch:
    hosts: ['localhost:9200']
```
Notice I am not enabling the `short id` options. So I expect that the published data won't have container's information.

- Restarting filebeat and the results are the same:
```
2018-02-12T13:57:19.444Z        DEBUG   [publish]       pipeline/processor.go:275       Publish event: {
  "@timestamp": "2018-02-12T13:57:19.444Z",                                                    
  "@metadata": {                                                                               
    "beat": "filebeat",                                                                        
    "type": "doc",                                                                                                                                                                             
    "version": "7.0.0-alpha1"                                                                                                                                                                  
  },                                                                                           
  "beat": {                                                                                    
    "name": "elastic",                                                                         
    "hostname": "elastic",                                                                     
    "version": "7.0.0-alpha1"                                                                                                                                                                    },                                                                                                                                                                                             "source": "/home/ripcurld0/logs/196f8a55cb7d/file.log",                                                                                                                                          "offset": 19546,                                                                                                                                                                               "message": "Mon Feb 12 13:57:18 UTC 2018",                                                                                                                                                   
  "event": {                                                                                   
    "type": "log"                                                                                                                                                                              
  },                                           
  "prospector": {
    "type": "log"
  }
}
```

But to be sure this is because the container ID `196f8a55cb7d` wasn't found I had to add "add_docker_metadata" into the `-d` option:
```
$ ./filebeat -e -c ~/configs/filebeat.yml -d "publish,add_docker_metadata"
```

and I do get the following:
```
2018-02-12T14:00:53.480Z        DEBUG   [add_docker_metadata]   add_docker_metadata/add_docker_metadata.go:169  Container not found: cid=196f8a55cb7d                                 [22/9786]
2018-02-12T14:00:53.480Z        DEBUG   [publish]       pipeline/processor.go:275       Publish event: {
....
```
that means that everything works like expected in `add_docker_metadata` as much as I can imagine.

- Enabling `short id`:
```
cat ~/configs/filebeat.yml
filebeat.prospectors:
    - type: log
      paths:
          - '/home/ripcurld0/logs/*/*'
      processors:
          - add_docker_metadata:
              match_source: true
              match_source_index: 3
              match_short_id: true

output.elasticsearch:
    hosts: ['localhost:9200'] 
```

- Running again filebeat (with `add_docker_meta` in `-d`)
The results are:
```
2018-02-12T14:05:56.854Z        DEBUG   [publish]       pipeline/processor.go:275       Publish event: {                                                                                       
  "@timestamp": "2018-02-12T14:05:56.854Z",    
  "@metadata": {                               
    "beat": "filebeat",                        
    "type": "doc",                                                                             
    "version": "7.0.0-alpha1"                                                                  
  },                                                                                           
  "event": {                                                                                   
    "type": "log"                                                                              
  },                                                                                           
  "prospector": {                                                                              
    "type": "log"                                                                              
  },                                                                                           
  "docker": {                                                                                  
    "container": {                                                                             
      "id": "196f8a55cb7dce6953a9f79bb137511dac12edc787d2be00861accc080ae949b",                
      "image": "fbgen",                                                                        
      "name": "heuristic_tesla"                                                                
    }                                                                                          
  },                                                                                           
  "beat": {                                                                                    
    "name": "elastic",                                                                         
    "hostname": "elastic",                                                                     
    "version": "7.0.0-alpha1"                                                                  
  },                                                                                           
  "message": "Mon Feb 12 14:05:56 UTC 2018",                                                   
  "source": "/home/ripcurld0/logs/196f8a55cb7d/file.log",                                        
  "offset": 34539                              
}                                              
```
As you can see here no error is shown on `add_docker_metadata` which is great.
In addition to that `docker.container` has the container information (`name`, `image` and the long ID).

### Notes
if you have other ideas, feel free to let me know.
I used similar configs to @chrisregnier [configuration](https://github.com/elastic/beats/issues/6092#issuecomment-361694206).

I think this is now more than `WIP`, although any design and implementation feedback/review is welcome!

(If we can have more eyes on this, that will be super awesome)
@ruflin @andrewkroh @exekias @chrisregnier

exekias:
ok to test

ripcurld0:
sure @exekias

chrisregnier:
This is fantastic! Thanks @ripcurld0

ripcurld0:
Thanks a lot @exekias @ruflin @chrisregnier @andrewkroh for your review.
Looking forward to contribute more to this amazing project! :tada: :rainbow: :cake:

PR #6174, opened 2018-01-25
Output template in error message

cyrilleverrier:
When there is an error in the generated template, the error message (returned by Elasticsearch server) is not enough to debug,
Having the content of the generated template displayed in the console helps to troubleshoot errors.

### DIFF
--- a/libbeat/template/load.go
+++ b/libbeat/template/load.go
@@ -67,7 +67,7 @@ func (l *Loader) Load() error {
 
 		err = l.LoadTemplate(tmpl.GetName(), output)
 		if err != nil {
-			return fmt.Errorf("could not load template: %v\nTemplate is:\n%s", err, output)
+			return fmt.Errorf("could not load template. Elasticsearh returned: %v. Template is: %s", err, output)
 		}
 	} else {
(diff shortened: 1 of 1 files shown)


ph:
jenkins test this please

libbeat/template/load.go
```diff
@@ -67,7 +67,7 @@ func (l *Loader) Load() error {
 
 		err = l.LoadTemplate(tmpl.GetName(), output)
 		if err != nil {
-			return fmt.Errorf("could not load template: %v", err)
+			return fmt.Errorf("could not load template: %v\nTemplate is:\n%s", err, output)
```
ruflin: I would prefer not to have any new lines in the error message if possible. In case this ends up in the logs somewhere it would mean it's on multiple log lines so it will end up in multiple (unrelated) events.

ruflin: @cyrilleverrier Could you change the above?

cyrilleverrier: @ruflin I changed to output the message in one single line of text.
I also verified it locally

kvch:
jenkins test this

PR #6180, opened 2018-01-25
Add missing default Kubernetes cleanup time

exekias:
Without this, Kubernetes autodiscover cleanup goroutine would have an inifnite loop
consuming CPU.

A default cleanup time is necessary to ensure we sleep between cleans.

A bigger refactoring is in the works (https://github.com/elastic/beats/pull/6159), but this should fix master and 6.2

### DIFF
--- a/libbeat/autodiscover/providers/kubernetes/config.go
+++ b/libbeat/autodiscover/providers/kubernetes/config.go
@@ -24,7 +24,8 @@ type Config struct {
 
 func defaultConfig() *Config {
 	return &Config{
-		InCluster:  true,
-		SyncPeriod: 1 * time.Second,
+		InCluster:      true,
+		SyncPeriod:     1 * time.Second,
+		CleanupTimeout: 60 * time.Second,
 	}
 }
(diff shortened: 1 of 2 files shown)


libbeat/autodiscover/providers/kubernetes/config.go
```diff
@@ -24,7 +24,8 @@ type Config struct {
 
 func defaultConfig() *Config {
 	return &Config{
-		InCluster:  true,
-		SyncPeriod: 1 * time.Second,
+		InCluster:      true,
+		SyncPeriod:     1 * time.Second,
+		CleanupTimeout: 60 * time.Second,
```
ph: I understand to provide a default values for the cleanup, but do we need to change the locking strategy to fix the issue?

ph: ah nevermind I understand now :)

ph:
WFSG, go go beats-ci

libbeat/autodiscover/providers/kubernetes/config.go
exekias: Not really, but I think that slipped my review, locking while sending to the bus is not needed so I moved it outside. I also added a debug line.

Issue #5630, opened 2017-11-17
Add support for CRI logs format


exekias:
Container Runtime Interface (CRI) initiative standardizes an API for container implementations. It defines its own format for logs, compatible with Docker `json-file`, but with an alternative implementation (no JSON).

We should update our [Docker prospector](https://github.com/elastic/beats/pull/5402) to support that format, probably we can do autodetection just by checking first chars.

Example log line: 
```
2017-09-12T22:32:21.212861448+00:00 stdout 2017-09-12 22:32:21.212 [INFO][88] table.go 710: Invalidating dataplane cache ipVersion=0x4 reason="refresh timer" table="nat"
```
CRI-o project: https://github.com/kubernetes-incubator/cri-o
CRI-o code writing logs: https://github.com/mrunalp/cri-o/blob/37e282a29b84ca12b594cc55767295f37da848ea/conmon/conmon.c#L290-L407

ruflin:
@exekias Good idea. We should have it available as autodiscovery or config option.

exekias:
Closed by #6181

PR #6181, opened 2018-01-25
Add support for CRI logs format

jreyeshdez:
Adds support for CRI logs format, also created tests for specific CRI logs.
Autodiscovery by looking at first char and it will invoke CRI or Docker log parser depending on it.
Solution for #5630

### DIFF
--- a/filebeat/harvester/reader/docker_json.go
+++ b/filebeat/harvester/reader/docker_json.go
@@ -8,6 +8,7 @@ import (
 	"github.com/elastic/beats/libbeat/common"
 
 	"github.com/pkg/errors"
+	"strings"
 )
 
 // DockerJSON processor renames a given field
@@ -23,12 +24,67 @@ type dockerLog struct {
 	Stream    string `json:"stream"`
 }
 
+type crioLog struct {
+	Timestamp time.Time
+	Stream    string
+	Log       []byte
+}
+
 // NewDockerJSON creates a new reader renaming a field
 func NewDockerJSON(r Reader, stream string) *DockerJSON {
 	return &DockerJSON{
 		stream: stream,
 		reader: r,
 	}
+}
+
+// parseCRILog parses logs in CRI log format.
+// CRI log format example :
+// 2017-09-12T22:32:21.212861448Z stdout 2017-09-12 22:32:21.212 [INFO][88] table.go 710: Invalidating dataplane cache
+func parseCRILog(b []byte, msg *crioLog, message Message) (Message, error) {
+	log := string(b)
+	ts, err := time.Parse(time.RFC3339, strings.Fields(log)[0])
+	if err != nil {
+		return message, errors.Wrap(err, "parsing CRI timestamp")
+	}
+
+	stream := strings.Fields(log)[1]
+	// Anything after stream.
+	logMessage := strings.Fields(log)[2:]
+	msg.Timestamp = ts
+	msg.Stream = stream
+	msg.Log = []byte(strings.Join(logMessage, " "))
+	message.AddFields(common.MapStr{
+		"stream": msg.Stream,
+	})
+	message.Content = msg.Log
+	message.Ts = ts
+
+	return message, nil
+}
+
+// parseDockerJSONLog parses logs in Docker JSON log format.
+// Docker JSON log format example:
+// {"log":"1:M 09 Nov 13:27:36.276 # User requested shutdown...\n","stream":"stdout"}
+func parseDockerJSONLog(b []byte, msg *dockerLog, message Message) (Message, error) {
+	dec := json.NewDecoder(bytes.NewReader(b))
+	if err := dec.Decode(&msg); err != nil {
+		return message, errors.Wrap(err, "decoding docker JSON")
+	}
+
+	// Parse timestamp
+	ts, err := time.Parse(time.RFC3339, msg.Timestamp)
+	if err != nil {
+		return message, errors.Wrap(err, "parsing docker timestamp")
+	}
+
+	message.AddFields(common.MapStr{
+		"stream": msg.Stream,
+	})
+	message.Content = []byte(msg.Log)
+	message.Ts = ts
+
+	return message, nil
 }
 
 // Next returns the next line.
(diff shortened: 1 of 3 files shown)


karmi:
Hi @jreyeshdez, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/beats/pull/6181.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

filebeat/harvester/reader/docker_json.go
```diff
@@ -31,6 +38,55 @@ func NewDockerJSON(r Reader, stream string) *DockerJSON {
...
+// CRI log format example :
+// 2017-09-12T22:32:21.212861448Z stdout 2017-09-12 22:32:21.212 [INFO][88] table.go 710: Invalidating dataplane cache
+func parseCRILog(b []byte, msg *crioLog, message Message) (Message, error) {
+	log := string(b)
+	ts, err := time.Parse(time.RFC3339, strings.Fields(log)[0])
+	if err != nil {
+		return message, errors.Wrap(err, "parsing CRI timestamp")
+	}
+
+	stream := strings.Fields(log)[1]
```
exekias: Calling strings.Fields only once should be more efficient, you can do it at the beginning of this method and use the resulting variable. It will probably be more efficient if you use `SplitN`, to as you don't want to split the whole message.

Also it would be nice to check the resulting array, to ensure it's size is correct, and avoid errors when accessing different fields

filebeat/harvester/reader/docker_json.go
```diff
@@ -39,27 +95,19 @@ func (p *DockerJSON) Next() (Message, error) {
...
-		if err = dec.Decode(&line); err != nil {
-			return message, errors.Wrap(err, "decoding docker JSON")
-		}
+		var dockerLine dockerLog
+		var crioLine crioLog
 
-		if p.stream != "all" && p.stream != line.Stream {
-			continue
+		if strings.HasPrefix(string(message.Content), "{") {
+			message, err = parseDockerJSONLog(message.Content, &dockerLine, message)
```
exekias: Having that you are passing `message` anyway, you could reduce the number of params to 2 (`&message` and `&dockerLine`). Same applies to parseCRILog

filebeat/harvester/reader/docker_json_test.go
```diff
@@ -32,12 +32,28 @@ func TestDockerJSON(t *testing.T) {
 			stream:        "all",
 			expectedError: true,
 		},
+		// Wrong CRI
+		{
+			input:         [][]byte{[]byte(`{this is not JSON nor CRI`)},
```
exekias: I'm guessing `2017-09-12T22:32:21.212861448Z stdout` as input would cause a panic, it would be nice to add it to the list of tests :)

filebeat/harvester/reader/docker_json.go
```diff
@@ -39,27 +95,19 @@ func (p *DockerJSON) Next() (Message, error) {
...
-		dec := json.NewDecoder(bytes.NewReader(message.Content))
-		if err = dec.Decode(&line); err != nil {
-			return message, errors.Wrap(err, "decoding docker JSON")
-		}
+		var dockerLine dockerLog
+		var crioLine crioLog
 
-		if p.stream != "all" && p.stream != line.Stream {
-			continue
+		if strings.HasPrefix(string(message.Content), "{") {
```
ruflin: I wonder if we should make this a config option instead of "guessing" it here? Are there potentially other formats which will be added?

filebeat/harvester/reader/docker_json.go
jreyeshdez: Thanks for your comments @exekias !
I will use `SplitN` and check array len, if less than 3 then it'll be considered an invalid CRI log.

filebeat/harvester/reader/docker_json.go
jreyeshdez: Agree! I'll reduce it to 2 params.

filebeat/harvester/reader/docker_json_test.go
jreyeshdez: Good catch! I added it to the test cases, its covered by checking array len.

filebeat/harvester/reader/docker_json.go
jreyeshdez: I'm planning to test the code on a running cluster, I tried yesterday but had some issues while setting up the environment on my Windows system.

filebeat/harvester/reader/docker_json.go
exekias: I don't think new formats will be added, CRI is the standard Kubernetes went for, the idea of this is to support both the standard and old Docker format.
I'm ok with autodetecting it as of today, we can change this in the future if we see the point

filebeat/harvester/reader/docker_json.go
```diff
@@ -41,16 +41,19 @@ func NewDockerJSON(r Reader, stream string) *DockerJSON {
...
 	}
 
-	stream := strings.Fields(log)[1]
+	stream := log[1]
 	// Anything after stream.
-	logMessage := strings.Fields(log)[2:]
+	logMessage := log[2:]
 	msg.Timestamp = ts
 	msg.Stream = stream
 	msg.Log = []byte(strings.Join(logMessage, " "))
```
exekias: You no longer need `strings.Join` here, this should be just log[2]

exekias:
Thank you for the update, I left one more comment, please address it and I think we are ready to go :smiley_cat:

filebeat/harvester/reader/docker_json.go
jreyeshdez: True, no need for `strings.Join` , but you meant `msg.Log = []byte(logMessage[0])` right ?

exekias: yep, or remove `logMessage` variable and just do `msg.Log = []byte(log[2]))`

exekias:
ok to test

PR #6183, opened 2018-01-25
Add script for promoting doc branch and version

dedemorton:
We could also have this script set the release state, but typically we bump the version and branch before we set the release state.

@tsg Just made some simple changes to modify your script to work here. Not sure if we want this script to do all the same stuff. I got some errors when I ran this, but they were the same errors I saw for your set_docs_version script, so I dunno if they are an issue.

### DIFF
--- a/dev-tools/promote_docs
+++ b/dev-tools/promote_docs
@@ -1,0 +1,35 @@
+#!/usr/bin/env python
+import argparse
+from subprocess import check_call
+
+def main():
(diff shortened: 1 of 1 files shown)


dedemorton:
Another option would be to call this set_docs_branch and have it simply set the docs branch. WDYT?

ruflin:
I merging this PR as the script does no change to our existing build. Like this we can iterate on top of the script if needed.

PR #6186, opened 2018-01-25
Set up logging -> Configure logging

roncohen:
image

### DIFF
--- a/libbeat/docs/loggingconfig.asciidoc
+++ b/libbeat/docs/loggingconfig.asciidoc
@@ -11,7 +11,7 @@
 //////////////////////////////////////////////////////////////////////////
 
 [[configuration-logging]]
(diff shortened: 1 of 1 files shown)


dedemorton:
Sure, looks fine. TBH, I want to do more cleanup on this TOC, but it seems like a low priority. I'm not sure any user will be confused, though maybe you're suggesting "configure" because you think it will be better for SEO?

dedemorton:
Since you are in the files already...I am OK with you fixing the problem everywhere.

dedemorton:
@roncohen Nevermind, I will just merge and fix the consistency later. thanks!

roncohen:
thanks @dedemorton! I agree there are more inconsistencies. I just saw this as i was working to move it into APM docs.

PR #6187, opened 2018-01-25
Cherry-pick #6180 to 6.2: Add missing default Kubernetes cleanup time

exekias:
Cherry-pick of PR #6180 to 6.2 branch. Original message: 

Without this, Kubernetes autodiscover cleanup goroutine would have an inifnite loop
consuming CPU.

A default cleanup time is necessary to ensure we sleep between cleans.

A bigger refactoring is in the works (https://github.com/elastic/beats/pull/6159), but this should fix master and 6.2


exekias:
that's correct

PR #6197, opened 2018-01-26
Add note about changing the index pattern for 5.x

roncohen:
image


libbeat/docs/dashboardsconfig.asciidoc
```diff
@@ -92,6 +92,8 @@ is `".kibana"`
 The Elasticsearch index name. This setting overwrites the index name defined
 in the dashboards and index pattern. Example: `"testbeat-*"`
 
+NOTE: Changing this setting while trying to load dashboards into Elasticsearch 5.x using {beatname_uc} 6.x will not work.
```
andrewkroh: Perhaps we should directly say:

    NOTE: This setting only works for Kibana 6.0 and newer.

I think the "using {beatname_uc} 6.x will not work" is unnecessary since this statement will only be included in the documentation for the versions it affects.

PR #6198, opened 2018-01-26
Fix metricbeat test util RunPushMetricSetV2

andrewkroh:
A test case could become deadlocked waiting to write to a channel. This would happen if the metricset under test attempted to publish events after the event consumer was stopped. Now when the metricset publishes to the stopped Reporter it will return false just as the real Reporter does in Metricbeat.

I also identified a resource leak while stress testing Auditbeat on Windows. The file_integrity module's fsnotify reader was not being closed when the done channel was closed. This would only affect Windows/Linux users that were constantly reloading their file_integrity config with config reloading.


andrewkroh:
@ruflin This should fix the `file_integrity.TestData` failures we were seeing on Windows.

ruflin:
@andrewkroh I wonder if we should backport this to 6.2 even though it probably only happens in edge cases?

PR #6203, opened 2018-01-27
Dedot kubernetes fields

walktall:
Related to https://github.com/elastic/beats/issues/5942

@ruflin @exekias PTAL~


libbeat/common/event.go
```diff
@@ -262,16 +262,22 @@ func DeDot(s string) string {
...
-	switch json.(type) {
+	switch json := json.(type) {
 	case map[string]interface{}:
 		result := map[string]interface{}{}
-		for key, value := range json.(map[string]interface{}) {
+		for key, value := range json {
+			result[DeDot(key)] = DeDotJSON(value)
+		}
+		return result
+	case MapStr:
```
ruflin: Thanks for adding this option here. Could you add a test case for this "switch" option?

libbeat/common/kubernetes/metadata.go
```diff
@@ -54,11 +54,11 @@ func (g *metaGenerator) PodMetadata(pod *Pod) common.MapStr {
 	}
 
 	if len(labelMap) != 0 {
-		meta["labels"] = labelMap
+		meta["labels"] = common.DeDotJSON(labelMap)
```
ruflin: @exekias This is a breaking change. I wonder if you see an issue in this?

ruflin:
jenkins, test it

libbeat/common/event.go
walktall: @ruflin Tests added, and also made some changes to original test code...

ruflin:
jenkins, test it

libbeat/common/event_test.go
```diff
@@ -399,49 +395,60 @@ func BenchmarkConvertToGenericEventStringPointer(b *testing.B) {
...
-
-	actualJSONBody = DeDotJSON(actualJSONBody).([]interface{})
-
-	assert.Equal(t, expectedJSONBody, actualJSONBody)
+func TestDeDotJSON(t *testing.T) {
+	var tests = []struct {
+		input  []byte
+		output []byte
+		valuer func() interface{}
+	}{
```
ruflin: Thanks for converting this to a table test. It makes it much easier to compare input with output. It makes it a bit harder to write json then before but I think the benefits outweight here.

libbeat/common/kubernetes/metadata.go
exekias: We need this as a bugfix anyway, isn't it? Didn't follow previous conversations but I guess this is needed to avoid mapping issues?

In any case it's a breaking change and we should make it clear in the release notes

exekias:
jenkins, test it please

libbeat/common/kubernetes/metadata.go
ruflin: As far as I know it didn't break yet but it would pretty soon :-) So yes having it as breaking change caused by a bug is a good way to state it.

ruflin:
@walktall Thanks for your work on this.

@exekias We should make sure this is communicate properly for 6.x in case people had queries on the label keys which will not work anymore.

vjsamuel:
There should be a better way to handle this as it invokes disparity between how objects are represented in Kubernetes and how autodiscover is configured in Beats. For example a label `prometheus.io.scrape` would now have to be configured as `prometheus_io_scrape`. if such conversions need to happen they should ideally happen either in a configurable manner where dedoting can be turned off or later in the pipeline in the form of a processor. Not all backends may need this dedoting to happen and there maybe usecases where everything appears the way it actually is on Kubernetes.

exekias:
I'm ok with making this optional, will open a PR for it :+1:

walktall:
Maybe it can be done here?https://github.com/elastic/beats/blob/master/libbeat/processors/add_kubernetes_metadata/kubernetes.go#L124

vjsamuel:
@walktall, autodiscover doesnt use this as it presets all the pod meta. so it would fail for that flow.

ruflin:
The problem with the dotted keys is that they also caused some issues in the processors. But this should be solved by https://github.com/elastic/beats/pull/5916

I agree we should make it optional as I would also prefer to have the same representation in ES as is delivered by k8s. It creates the potential issue of collisions as also discussed in https://github.com/micrometer-metrics/micrometer/issues/154#issuecomment-366096851 My thinking here is that Metricbeat should have a way prevent such collisions at least for one single event. Meaning if there is `{"a": 15, "a.b": 17}` that it makes something like `{"a.value": 15, "a.b": 17}` out of it. Would be great to have a very efficient way of doing that.

+1 on having a processor for it. The important part with a processor is that we can specify which part of an event should be dedotted as in most cases it's only the labels.

PR #6204, opened 2018-01-27
Edit autodiscover docs

dedemorton:
Noticed a typo, so made a few minor edits while I had the file open.


libbeat/docs/shared-autodiscover.asciidoc
```diff
@@ -114,7 +114,7 @@ For example:
 }
 -------------------------------------------------------------------------------------
 
-Configuration of templates and conditions are exactly similar as that of Docker provider. Configuration templates can
+The configuration of templates and conditions is similar to that of the Docker provider. Configuration templates can
```
dedemorton: this sentence was ambiguous. Should say wither "is similar to" or "is exactly the same"

The verb is singular because the noun is configuration, and it's singular.

Issue #6192, opened 2018-01-26
Stacktrace when monitoring Elasticsearch with MetricBeat


beiske:
Beats version 6.1.2. 
Elasticsearch output cluster version 6.0.1. 
Monitored Elasticsearch cluster versions: 2.4.6
Operating system Ubuntu 64 bit.

Config
```
cat /mnt/data/local-1/allocator/10.0.2.15/state/elasticsearch/admin-console/instance-0000000000/beat-modules.yaml
---
- module: "elasticsearch"
  metricsets:
  - "node"
  - "node_stats"
  period: "10s"
  user: "ec-local-beats-monitor"
  password: "**REDACTED**"
  hosts:
  - "10.0.2.15:18461"
  ssl: false
```

```
 cat metricbeat.yml
###################### Metricbeat Configuration  #######################

#==========================  Modules configuration ============================

metricbeat.config.modules:
   path: /mnt/data/local-1/allocator/10.0.2.15/state/elasticsearch/*/*/beat-modules.yaml
  reload.enabled: true
  reload.period: 10s

#==================== Elasticsearch template setting ==========================

setup.template.settings:
  index.number_of_shards: 1
  index.codec: best_compression
  #_source.enabled: false

#================================ General =====================================


cloud.id: "temperature:**REDACTED**"
cloud.auth: "elastic:**REDACTED**"
```

This results in:
```
./metricbeat -e
metricbeat2018/01/26 09:29:59.231294 cloudid.go:42: INFO Setting Elasticsearch and Kibana URLs based on the cloud id: output.elasticsearch.hosts=https://****.eu-west-1.aws.found.io:443 and setup.kibana.host=https://****.eu-west-1.aws.found.io:443
2018/01/26 09:29:59.231982 beat.go:436: INFO Home path: [/home/found/metricbeat-6.1.2-linux-x86_64] Config path: [/home/found/metricbeat-6.1.2-linux-x86_64] Data path: [/home/found/metricbeat-6.1.2-linux-x86_64/data] Logs path: [/home/found/metricbeat-6.1.2-linux-x86_64/logs]
2018/01/26 09:29:59.232129 beat.go:443: INFO Beat UUID: 67de923d-31e7-4cf3-bd17-ec6be5f761d5
2018/01/26 09:29:59.232252 beat.go:203: INFO Setup Beat: metricbeat; Version: 6.1.2
[...]

PR #6205, opened 2018-01-28
Fix issue that wrong config panics http metricsets

ruflin:
When setting an invalide http config in a metricbeat module / metricset it could happen that the metricset paniced. The reason is that the error when unpacking the config was not properly returned and handled which lead to an empty http instance.

This PR changes the behaviour to return the errors which can then be handled by the metricsets. The issue also affects all metricsets which were based on the prometheus helper.

Closes https://github.com/elastic/beats/issues/6192


ruflin:
@exekias I don't think we should backport this to 6.2 as the change is too big and the "bug" only happens if there is an invalid configuration. It should definitively be in 6.x which will happen automatically with the next bulk merge.

exekias:
WFG

PR #6213, opened 2018-01-29
Start pods watcher when autodiscover provider starts

exekias:
Pods watcher was starting too soon, so initial pods where not spawning
new autodiscover events, as autodiscover manager is not registered to
listen yet.

This only affects master, as it was recently refactored: #6159

### DIFF
--- a/libbeat/autodiscover/providers/kubernetes/kubernetes.go
+++ b/libbeat/autodiscover/providers/kubernetes/kubernetes.go
@@ -77,15 +77,14 @@ func AutodiscoverBuilder(bus bus.Bus, c *common.Config) (autodiscover.Provider, 
 		},
 	})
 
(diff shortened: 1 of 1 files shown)


walktall:
Sorry for introducing this bug 😢

exekias:
Don't be, this hasn't been released yet! These things happen when big chunks of code are updated. Also I merged it, so we can share the pain :smile:

ruflin:
@exekias On CI a test related to k8s fails. I would think it's not related but better have a look: https://beats-ci.elastic.co/job/elastic+beats+pull-request+multijob-darwin/381/beat=metricbeat,label=macosx/console

exekias:
Failing test is being addressed here: https://github.com/elastic/beats/issues/6228

PR #6217, opened 2018-01-30
Fix event_logs.processors config being rejected

andrewkroh:
The `event_logs.processors` keyword was being rejected as invalid config by Winlogbeat. This fixes the issue by adding "processors" as an allowed configuration key for `event_logs` and adds a system test case.

### DIFF
--- a/winlogbeat/eventlog/factory.go
+++ b/winlogbeat/eventlog/factory.go
@@ -10,7 +10,8 @@ import (
 	"github.com/elastic/beats/libbeat/common"
 )
 
-var commonConfigKeys = []string{"api", "name", "fields", "fields_under_root", "tags"}
+var commonConfigKeys = []string{"api", "name", "fields", "fields_under_root",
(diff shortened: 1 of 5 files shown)


andrewkroh:
This was reported in https://discuss.elastic.co/t/winlogbeat-processor-not-dropping-events/115634.

winlogbeat/eventlog/factory.go
```diff
@@ -10,7 +10,8 @@ import (
 	"github.com/elastic/beats/libbeat/common"
 )
 
-var commonConfigKeys = []string{"api", "name", "fields", "fields_under_root", "tags"}
+var commonConfigKeys = []string{"api", "name", "fields", "fields_under_root",
```
ruflin: Kind of nice to have this list for error checking but requires extendig it every time we add a global option.

ruflin:
@andrewkroh Worth backporting?

PR #6219, opened 2018-01-30
Ensure autodiscover events are processed in order

exekias:
We need to read start/stop events from a single queue to ensure they are
processed in order.

Before this change, in some edge cases, when updates spawn stop/start
events, autodiscover was processing the start event before stop, leading
to an undesired state: no running config for an existing pod.

### DIFF
--- a/libbeat/autodiscover/autodiscover.go
+++ b/libbeat/autodiscover/autodiscover.go
@@ -27,11 +27,8 @@ type Adapter interface {
 	// RunnerFactory provides runner creation by feeding valid configs
 	cfgfile.RunnerFactory
 
-	// StartFilter returns the bus filter to retrieve runner start triggering events
-	StartFilter() []string
-
-	// StopFilter returns the bus filter to retrieve runner stop triggering events
-	StopFilter() []string
+	// EventFilter returns the bus filter to retrieve runner start/stop triggering events
+	EventFilter() []string
 }
 
 // Autodiscover process, it takes a beat adapter and user config and runs autodiscover process, spawning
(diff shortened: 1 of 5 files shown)


libbeat/autodiscover/autodiscover.go
```diff
@@ -27,11 +27,8 @@ type Adapter interface {
 	// RunnerFactory provides runner creation by feeding valid configs
 	cfgfile.RunnerFactory
 
-	// StartFilter returns the bus filter to retrieve runner start triggering events
-	StartFilter() []string
-
-	// StopFilter returns the bus filter to retrieve runner stop triggering events
-	StopFilter() []string
+	// EventFilter returns the bus filter to retrieve runner start/stop triggering events
+	EventFilter() []string
```
ruflin: Makes also the interface simpler.

libbeat/autodiscover/autodiscover_test.go
```diff
@@ -173,19 +165,40 @@ func TestAutodiscover(t *testing.T) {
...
+	assert.False(t, runners[1].stopped)
+
+	// Test stop event
+	eventBus.Publish(bus.Event{
+		"stop": true,
+		"meta": common.MapStr{
+			"foo": "baz",
+		},
+	})
+	time.Sleep(10 * time.Millisecond)
```
ruflin: Let's hope it's enough on slow machines :-)

exekias:
Thank you for the review! you are right about the moved code blocks, most of it is just wrapping old code into new methods

libbeat/autodiscover/autodiscover_test.go
exekias: I don't learn the lesson :), let see if jenkins like it, I'll can add some back-off code if not

ruflin: It's more a question of how often Jenkins will like it 😆

ruflin:
@exekias I assume the commits can be squashed?

exekias:
Sure

exekias:
Failing test is being addressed here: https://github.com/elastic/beats/pull/6232

PR #6224, opened 2018-01-30
Fix system process metricset for kernel processes

### LINKED ISSUE #5700: Metricbeat6.0 system module doesn't capture kernel process in Centos7
- Version: metricbeat6.0.0
- Operating System: Centos7 x64
- Steps to Reproduce: 

metricbeat system module doesn't capture the system process there cmdline always like this ` [xxxx]  `
and I enabled debug log found nothing useful, there nothing about these process
It's all kernel process like this:

>  11 root rt 0 0 0 0 S 0.0 0.0 0:30.50 watchdog/0
>  12 root rt 0 0 0 0 S 0.0 0.0 0:31.35 watchdog/1
>  13 root rt 0 0 0 0 S 0.0 0.0 0:19.37 migration/1
>  14 root 20 0 0 0 0 S 0.0 0.0 1:25.61 ksoftirqd/1
>  ...

https://discuss.elastic.co/t/metricbeat6-0-system-module-doesnt-capture-linux-system-process/108654

andrewkroh:
On Linux, kernel processes don't have an exe on disk so the `/proc/[PID]/exe` symlink is broken. When Metricbeat tries to read the symlink it gets an error and it skips the whole process. This adds a check for ENOENT and ignores the error which allows kernel processes to be reported by Metricbeat. This is the error that it fixes.

    2018-01-30T20:31:02.512Z	DEBUG	[processes]	process/process.go:443	Skip process pid=12113: error getting process exe for pid=12113: readlink /proc/12113/exe: no such file or directory

Fixes #5700

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -111,6 +111,7 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
 - Fix process cgroup memory metrics for memsw, kmem, and kmem_tcp. {issue}6033[6033]
 - Fix kafka OffsetFetch request missing topic and partition parameters. {pull}5880[5880]
 - Change kubernetes.node.cpu.allocatable.cores to float. {pull}6130[6130]
+- Fix system process metricset for kernel processes. {issue}5700[5700]
 
 *Packetbeat*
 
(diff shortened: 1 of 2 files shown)


ruflin:
@andrewkroh Should we backport this one?

Issue #6214, opened 2018-01-29
Packetbeat:  cyclic reference detected for key: 'path.logs' 


ph:
```json
{
  "level": "error",
  "timestamp": "2018-01-29T15:57:07.020-0500",
  "caller": "instance/beat.go:667",
  "message": "Exiting: error initializing publisher: cyclic reference detected for key: 'path.logs' accessing 'output.file.path' (source:'packetbeat.yml')"
}
```

```yaml
packetbeat.interfaces.device: en0

packetbeat.protocols:
- type: http
  ports: [80]

output.file:
  path: '${path.logs}'
  filename: output.json

logging:
  level: info
  json: true
  files:
    name: packetbeat.json
```
This same config works in 6.1.2.

Reported by @andrewkroh

ph:
Also in metricbeat.

ph:
I think its related to the merge behavior and the way we do keep track of the visited node.

If you add the following line, it does out of the cyclic reference:

```
path.logs: "/var/log"
```

andrewkroh:
I'm adding a work-around to beats-tester that should be removed after a fix is available. See https://github.com/elastic/beats-tester/pull/72/files#diff-1221d011d05b696f87a414547df772dbR28

ph:
Thanks @andrewkroh, I've pushed the PR for review at https://github.com/elastic/go-ucfg/pull/100

PR #6225, opened 2018-01-30
Update go-ucfg to fix an issue with cyclic reference

ph:
When a field was referencing a field which is a reference of another and its not a
cyclic reference the algorithm was detecting it as a cyclic reference
when we were reify the values into the struct.

We have fixed the issue in https://github.com/elastic/go-ucfg/pull/100

Fixes: #6214


ph:
The two failling tests seems to be unrelated.
I've created https://github.com/elastic/beats/issues/6228

PR #6229, opened 2018-01-30
Bump version to 5.6.8 in the 5.6 branch

tsg:


ruflin:
Lots of failing tests but non are related to this change I think. For the Mac setup on Jenkins it seems so changes we made are needed so skip docker test and we don't have that in the 5.x branch.

PR #6230, opened 2018-01-30
Use last resource version when watching kubernetes events

exekias:
This will avoid duplicated events when watching for pod events. Duplicated events don't affect `add_kubernetes_metadata` but can cause issues when used by autodiscover, as procesing a start/stop event twice can lead to undesired status. For instance, without this change it's common to get a list of start-stop-start events for running pods during startup.

master is not affected as this code was refactored

### DIFF
--- a/libbeat/common/kubernetes/watcher.go
+++ b/libbeat/common/kubernetes/watcher.go
@@ -131,7 +131,7 @@ func (p *podWatcher) Start() error {
 func (p *podWatcher) watch() {
 	for {
 		logp.Info("kubernetes: %s", "Watching API for pod events")
-		watcher, err := p.client.WatchPods(p.ctx, "", p.nodeFilter)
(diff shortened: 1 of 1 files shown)


CHANGELOG.asciidoc
```diff
@@ -141,6 +141,7 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
 - Add autodiscover for kubernetes. {pull}6055[6055]
 - Add the abilility for the add_docker_metadata process to enrich based on process ID. {pull}6100[6100]
 - The `add_docker_metadata` and `add_kubernetes_metadata` processors are now GA, instead of Beta. {pull}6105[6105]
+- Use last resource version when watching for kubernetes events, to avoid duplicates. {pull}6230[6230]
```
ruflin: Is this changelog even neded as k8s watcher was not in a previous version I think?

ruflin: Is this changelog even neded as k8s watcher was not in a previous version I think?

exekias: true, I'll remove it

PR #6233, opened 2018-01-30
Fix setup.dashboards.always_kibana when using Kibana 5.6

### LINKED ISSUE #6090: setup.dashboards.always_kibana doesn't work with Kibana 5.6
Attempts to use `-E setup.dashboards.always_kibana=true` with Kibana 5.6 always result in an error. This is the code in question:

https://github.com/elastic/beats/blob/6130ef3b464008b9cdfae9476bdcf03076e642fd/libbeat/setup/kibana/client.go#L172-L182

After it successfully falls back to unmarshaling with the 5.x data type the code path always hits line 181 which is to return an error like the one below.

```
Exiting: Error importing Kibana dashboards: fail to create the Kibana loader: Error creating Kibana client: fail to get the Kibana version:fail to unmarshal the response from GET http://localhost:5601/api/status: json: cannot unmarshal string into Go struct field kibanaVersionResponse.version of type struct { Number string "json:\"number\""; Snapshot bool "json:\"build_snapshot\"" }. Response: {"name":"kibana","version":"5.6.6","buildNum":15605,"buildSha":"b67ac57661fc94df6ee476d0af6b3734211df61c","uuid":"25e5ec36-a8a6-4934-9010-5e78e2e345eb","status":{"overall":{"state":"green","title":"Green","nickname":"Looking good","icon":"success","s... (truncated)
```

andrewkroh:
In order to handle the different Kibana /api/status response formats in 5.x and 6.x the client attempts to parse it as 6.x and then falls back to the 5.x format on error. However if it was successful at parsing it was a 5.x response type is still always reported that an error had occurred.

Fixes #6090

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -66,6 +66,7 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
 - Fix logstash output debug message. {pull}5799{5799]
 - Fix isolation of modules when merging local and global field settings. {issue}5795[5795]
 - Fix panic when Events containing a float32 value are normalized. {pull}6129[6129]
+- Fix `setup.dashboards.always_kibana` when using Kibana 5.6. {issue}6090[6090]
 
 *Auditbeat*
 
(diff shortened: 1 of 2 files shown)


ruflin:
Merging as the failing tests do not seem to be related.

PR #6235, opened 2018-01-31
[DOCS] Monitoring Beats

lcawl:
This PR adds monitoring configuration information in all of the appropriate appropriate books.


urso:
The elasticsearch reporter supports quite a many more settings. See [config.go](https://github.com/elastic/beats/blob/master/libbeat/monitoring/report/elasticsearch/config.go#L11). It's mostly the same settings as supported by Elasticsearch output.

If monitoring + elasticsearch output are configured, the monitoring configuration with merged with the output configuration. If monitoring uses same host and credentials, this minimal config should work as well:

```
xpack.monitoring:
  enabled: true
  elasticsearch:
```

Please note, xpack monitoring creates a `beats_system` user for use with the monitoring feature.

monicasarbu:
@urso can you please have a look? It would be nice to merge this today or tomorrow. Thanks!

libbeat/docs/monitoring/configuring.asciidoc
```diff
@@ -0,0 +1,178 @@
...
+  elasticsearch:
+    url: ["https://example.com:9200", "https://example2.com:9200"]
+    username: elastic
+    password: changeme
+--------------------
+
+Many of the configuration options for `xpack.monitoring` match the
+`output.elasticsearch` options. If you configured {es} output and you want to
+use the same {es} cluster and credentials to store your monitoring metrics, you
+can specify the following minimal configuration options:
```
urso: Note: the monitoring data must be shipped to the production cluster. It is the x-pack production cluster forwarding the metrics to the monitoring cluster. The hosts should only be configured if another output then elasticsearch is configured (e.g. Kafka or Logstash).

libbeat/docs/monitoring/configuring.asciidoc
```diff
@@ -0,0 +1,178 @@
...
+`xpack.monitoring` settings in the {beatname_uc} configuration file. For
+example:
++
+--
+[source, yml]
+--------------------
+xpack.monitoring:
+  enabled: true
+  elasticsearch:
+    url: ["https://example.com:9200", "https://example2.com:9200"]
```
urso: setting is `hosts`, not `url`.

The reference config contains all settings with defaults + docs: https://github.com/elastic/beats/blob/master/libbeat/_meta/config.reference.yml#L787

libbeat/docs/monitoring/configuring.asciidoc
```diff
@@ -0,0 +1,178 @@
...
+[float]
+===== `ssl`
+
+Configuration options for Transport Layer Security (TLS) or Secure Sockets Layer
+(SSL) parameters like the certificate authority (CA) to use for HTTPS-based
+connections. If the `ssl` section is missing, the host CAs are used for
+HTTPS connections to {es}. For more information, see <<configuration-ssl>>.
+
+[float]
+===== `url`
```
urso: I don't think this setting exists

libbeat/docs/monitoring/configuring.asciidoc
urso: Just notices the reference config is missing period and buffer_size. Can you update these in the config.reference.yml file as well?

To some extend we try to keep the reference config file in sync with the docs.

libbeat/docs/monitoring/configuring.asciidoc
```diff
@@ -0,0 +1,178 @@
...
+
+The maximum number of metrics to bulk in a single {es} bulk API index request.
+The default is `50`. For more information, see <<elasticsearch-output>>.
+
+////
+[float]
+===== `buffer_size`
+
+TBD. The default value is `50`.
+////
```
urso: This setting is obsolete. It's not used anymore. Will remove it from code base in follow up PR.

libbeat/docs/monitoring/configuring.asciidoc
```diff
@@ -0,0 +1,178 @@
...
+===== `password`
+
+The password that {beatname_uc} uses to authenticate with the {es} instances for
+shipping monitoring data.
+////
+[float]
+===== `period`
+
+TBD. The default value is 60 * time.Second.
+////
```
urso: Monitoring requires the beat to push/publish metrics every N seconds to elasticsearch. The `period` configures the time interval a new snapshot of beats metrics is generated and scheduled to be published. If output is unresponsive for some long time and queues are full, a snapshot might get missing.

libbeat/docs/monitoring/configuring.asciidoc
```diff
@@ -0,0 +1,178 @@
...
+===== `url`
+
+The URL for the {es} instances. You can specify a single host as a string or
+specify multiple hosts as an array. Defaults to `http://localhost:9200`.
+
+[float]
+===== `username`
+
+The user ID that {beatname_uc} uses to authenticate with the {es} instances for
+shipping monitoring data.
```
urso: Note: this is by default `beats_system`. X-Pack automatically creates a `beats-system` user for use with monitoring. The password should be configured, as X-Pack creates a random password at install time.

lcawl:
Thanks @urso I've implemented your suggestions.  If I misunderstood your request related to the config.reference.yml file, just let me know and we can pursue that update in a separate PR.

tvernum:
@urso 
I think we have a small problem here (sorry, I know I'm this is only the docs PR, but it's the one that's open and mentions `beats_system`).

The beats system user doesn't exist in Elasticsearch. It was backed out 6 months ago in this PR
https://github.com/elastic/x-pack-elasticsearch/pull/2042 because beat monitoring was delayed.

The current build (BC6) of X-Pack-ES 6.2 does not have a builtin user for beat monitoring.

dedemorton:
@tvernum While we're waiting for `beats_system` user to be added in a future release, maybe we should document how to create a user that has the required permissions/roles to use Beats monitoring. WDYT?

tvernum:
@dedemorton It's possible to document the permissions the user needs, it should be fairly straight forward. I'm not up to speed on Beat monitoring, but I assume it just needs the same permissions as the `logstash_system` user.

If so, we can either document it that way :: "Just create a user + role that matches logstash_system" or copy-and-paste the logstash definitions into the docs as an example to follow.

Either way, let's not call them `beats_system` (in the docs) because that will cause issues when we actually add the builtin user with that name.

auditbeat/docs/configuring-howto.asciidoc
```diff
@@ -36,6 +36,7 @@ The following topics describe how to configure {beatname_uc}:
 * <<yaml-tips>>
 * <<regexp-support>>
 * <<{beatname_lc}-reference-yml>>
+* <<configuring-monitoring>>
```
dedemorton: I'd rather that this topic come after `<<configuration-logging>>` for all the Beats. If you are pressed for time, I can do the work after you merge this PR. I think it would be good to get this in quickly.

urso:
@pickypg @tsullivan Can you provide some input on required settings?

libbeat/_meta/config.reference.yml
```diff
@@ -188,7 +188,10 @@ output.elasticsearch:
   # Configure http request timeout before failing an request to Elasticsearch.
   #timeout: 90
 
-  # Use SSL settings for HTTPS.
+  # The time interval (in seconds) when metrics are set to Elasticsearch.
```
urso: s/set/send/ ?

libbeat/_meta/config.reference.yml
```diff
@@ -188,7 +188,10 @@ output.elasticsearch:
...
+  #period: 60 * time.Second
+
+  # Use SSL settings for HTTPS. Default is true.
```
urso: where is this doc comming from? Should be `Use SSL settings for HTTPS`.

lcawl: @urso, I'm not sure I understand your feedback, so I'm going to omit the config.reference.yml changes from this documentation-specific PR.  They can be implemented in a separate PR.

lcawl:
@urso @tvernum I've updated the instructions to say: 
"Create a user that has appropriate authority to send system-level monitoring data to Elasticsearch."

If we can be more specific (e.g. telling them grant that user the built-in "logstash_system" role), let me know and I'll add that.

libbeat/_meta/config.reference.yml
urso: Uhm... I would like you to modify config.reference.yml.

So originally this line said: `# Use SSL settings for HTTPS. Default is true.`
As this comment is not correct we changed it to: `# Use SSL settings for HTTPS.`

Somehow your PR undid the change back to the incorrect doc comment.

Feel free to ignore config.reference.yml for now. I can prepare another PR doing the config.reference.yml update. Btw. one also needs to run `make update` after updateing config.reference.yml, so to update all global config files we ship.

tsullivan:
I **thought** that the user would simply be able to create a user with the built-in `beats_system` role, which would have been intuitive, but unfortunately such a role does not exist.

I tested the alternative by creating a user with the `logstash_system` role, and that seems to work fine.

### Steps:
1. Create a user with the `logstash_system` role:
  ```
  curl -H "Content-Type: application/json" -XPUT -u elastic:changeme "${ES_HOST}/_xpack/security/user/logstash_system/_password" -d '{"password" : "changeme"}'
  ```
2. Supply those user credentials in the Beats `.yml` file under `xpack.monitoring`:
  ```
  xpack.monitoring:
    enabled: true
    elasticsearch:
      url: "https://spicy.local:9200"
      username: logstash_system
      password: changeme
  ```
3. Start the Beat, and the metrics data eventually shows in the dashboards
  - image

tsullivan:
@lcawl pointed out the `logstash_system` user is built-in, so they don't need to be **created** but they do need to have the password set in order for the user to be activated. Seems that is effectively what I did when I ran the `curl` command, but the documentation recommends an easier way: https://www.elastic.co/guide/en/x-pack/current/setting-up-authentication.html#set-built-in-user-passwords

Further note from @lcawl:
> .. Either they need to use the `logstash_system` built-in user (which has the `logstash_system` role) or else they need to create some other generic user and grant it the `logstash_system` role.

which is a correct statement ^^

libbeat/docs/monitoring/monitoring-beats.asciidoc
```diff
@@ -0,0 +1,69 @@
...
+{kibana-ref}/beats-page.html[Beats Monitoring Metrics].
+
+To configure {beatname_uc} to collect and send monitoring metrics:
+
+. Create a user that has appropriate authority to send system-level monitoring
+data to {es}. For example, you can use the built-in `logstash_system` user or
+assign the built-in `logstash_system` role to another user. For more
+information, see
+{xpack-ref}/setting-up-authentication.html[Setting Up User Authentication] and
+{xpack-ref}/built-in-roles.html[Built-in Roles].
```
urso: It should be clear the generic user is not named `beats_system`, as this name will be used in future revisions.

lcawl:
Backported to 6.x and 6.2 branchs with e19d3b59848774cf3a47a1008b0b5b9727b6757f

PR #6237, opened 2018-01-31
Fix event formatting with missing params (#6234)

### LINKED ISSUE #6234: Bad event log records can cause Winlogbeat crashes in XP / 2003
If an event is published to the Windows event log that contains fewer message parameters than are required by the message template this will cause a crash.  [`FormatMessage`](https://msdn.microsoft.com/en-us/library/windows/desktop/ms679351%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396) does not safely check the number of arguments provided against the number required by the template.

For example:
1. Application reports event 123 with parameters `["akroh", "192.168.0.1"]`.
2. Event 123 is defined as `User %1 connected from %2:%3.`.
3. Winlogbeat receives the event and attempts to populate the template using the given paremters.
4. `FormatMessage` causes a STATUS_ACCESS_VIOLATION exception that crashes Winlogbeat.

- Version: all
- Operating System: Windows XP and Windows 2003 (all architectures)
- Steps to Reproduce:
  1. An application publishes an event with fewer parameters than are required for the message template.

adriansr:
There was a crash in the eventlogging module used in legacy Windows versions (XP and 2003 server). It is necessary to account for the case where an event contains fewer parameters than required by its format string.

<strike>This patch also opportunistically adds the ability to lookup event descriptions in the system message table, as a last resort when a message can't be found in the message files specified in the registry.</strike> (feature removed for now)

Closes #6234

### DIFF
--- a/winlogbeat/eventlog/eventlogging_test.go
+++ b/winlogbeat/eventlog/eventlogging_test.go
@@ -474,6 +474,49 @@ func TestReadWhileCleared(t *testing.T) {
 	}
 }
 
+// Test event messages that include less parameters than required for message
+// formating (caused a crash in previous versions)
+func TestReadMissingParameters(t *testing.T) {
+	configureLogp()
+	log, err := initLog(providerName, sourceName, servicesMsgFile)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer func() {
+		err := uninstallLog(providerName, sourceName, log)
+		if err != nil {
+			t.Fatal(err)
+		}
+	}()
+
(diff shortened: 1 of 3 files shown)


adriansr:
Not mergeable yet: There is some improvement I would like to do in order to protect against maliciously formatted messages. Also one winlogbeat test is failing

adriansr:
Closed as now we have a better fix for this

andrewkroh:
The new PR for this is https://github.com/elastic/beats/pull/6247.

PR #6238, opened 2018-01-31
Cherry-pick #6219 to 6.2: Ensure autodiscover events are processed in order

exekias:
Cherry-pick of PR #6219 to 6.2 branch. Original message: 

We need to read start/stop events from a single queue to ensure they are
processed in order.

Before this change, in some edge cases, when updates spawn stop/start
events, autodiscover was processing the start event before stop, leading
to an undesired state: no running config for an existing pod.


exekias:
Red in jenkins is due to a Docker daemon stall, after tests passed

PR #6240, opened 2018-01-31
Logging fixes for verbosity

### LINKED ISSUE #6239: Logging level issues
@simitt found a problem with logging configuration:

* log level warning actually needs to be warn, otherwise an error is thrown and the server doesn't start (fixed in https://github.com/elastic/beats/pull/6184/commits/5f597df1d2cf402fa3bfa3e32a26e8f7904466e1)
* log level error still outputs info and warn logs.
* log level warn still outputs info logs.

tsg:
* Add a check for the `verbose` flag, before going back to Info.
* Revert to using `warning` in the configuration instead of `warn`. It seems that
  we have always used `warning`.

Fixes #6239.


simitt:
@roncohen I think you have to change back the `warn` in the docs to `warning`

PR #6254, opened 2018-02-01
Add monitoring settings to reference configuration

tsg:
Note that monitoring is disabled by default in Beats.

Related to #6235.

### DIFF
--- a/libbeat/_meta/config.reference.yml
+++ b/libbeat/_meta/config.reference.yml
@@ -784,6 +784,93 @@ logging.files:
 #logging.json: false
 
 
+#============================== Xpack Monitoring =====================================
+# beatname can export internal metrics to a central Elasticsearch monitoring cluster.
+# This requires xpack monitoring to be enabled in Elasticsearch.
+# The reporting is disabled by default.
+
+# Set to true to enable the monitoring reporter.
+#xpack.monitoring.enabled: false
+
+# Uncomment to send the metrics to Elasticsearch. Any setting that is not set is
+# automatically inherited from the output configuration, so if you want to use the
+# same cluster for monitoring data, you can simply uncomment the following line.
+#xpack.monitoring.elasticsearch:
+
+  # Array of hosts to connect to.
+  # Scheme and port can be left out and will be set to the default (http and 9200)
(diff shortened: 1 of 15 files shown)


libbeat/_meta/config.reference.yml
```diff
@@ -784,6 +784,93 @@ logging.files:
...
+
+  # The maximum number of events to bulk in a single Elasticsearch bulk API index request.
+  # The default is 50.
+  #bulk_max_size: 50
+
+  # Configure http request timeout before failing an request to Elasticsearch.
+  #timeout: 90
+
+  # Use SSL settings for HTTPS. Default is true.
+  #ssl.enabled: true
```
urso: Comment is somewhat misleading. This default config has no SSL settings enabled. The `ssl.enabled` setting is only used to 'ignore' the ssl settings if `ssl.enabled: false`. Plus, if "https" scheme is used system certificates will be used by the stdlib.

libbeat/_meta/config.yml
```diff
@@ -91,3 +91,18 @@ output.elasticsearch:
...
+# reporting is disabled by default.
+
+# Set to true to enable the monitoring reporter.
+#xpack.monitoring.enabled: false
+
+# Uncomment to send the metrics to Elasticsearch. Most settings from the
+# elasticsearch output are accepted here as well. Any setting that is not set is
+# automatically inherited from the output configuration, so if you want to use
+# the same cluster for monitoring data, you can simply uncomment the following
+# line.
```
urso: x-pack always requires to pass data through production cluster. That is hosts must match the elasticsearch output.

>  so if you want to use the same cluster for monitoring data, you can simply uncomment the following
 line.

Note: elasticsearch report configs are only inherrited from output.elasticsearch config. If logstash is configured as output, the host+username+password for ES must be configured here.

libbeat/_meta/config.reference.yml
tsg: I just copied this from the ES output SSL settings :)

libbeat/_meta/config.yml
tsg: Ah, I forgot about that. Let's see if I can clarify it.

PR #6266, opened 2018-02-02
Feature: TCP Input

ph:
Allow to receive new events via TCP, this will create a new event per
line and add some useful information about the connected client to the
evant. This input is marked as **experimental**.

This input expose the following settings:

- `line_delimiter`: The characters used to split incoming events, by
default it will split on `\n` and will also strip both `\n` and `\r`.
You can also use any characters like `;` and you can also used multiple
characters delimiter like `<END>`, the delimiter tokens will always be
removed from the string.

- `max_read_buffer`: This is a number of bytes that a client can buffer
in memory before finding a new message, if the limit is reached the
connection is killed and the event is logged. This is to prevent rogue
clients to DoS attack by consuming all the available memory. The default
values is 20971520.

- `timeout`: The server will close any client that reach this inactivity
timeout.

**TODO**
I've an almost TLS support done, but this PR was getting pretty big and I still have the tests to write so I've excluded it.

Ref: #5862

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -49,6 +49,8 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
   the system test. {pull}6121[6121]
 - Add IIS module to parse access log and error log. {pull}6127[6127]
 - Remove the undefined `username` option from the Redis input and clarify the documentation. {pull}6662[6662]
+- Refactor the usage of prospector to input in the YAML reference {pull}6121[6121]
+- Addition of the TCP input {pull}6266[6266]
 
 *Heartbeat*
 
--- a/filebeat/_meta/common.reference.p2.yml
+++ b/filebeat/_meta/common.reference.p2.yml
@@ -240,6 +240,20 @@ filebeat.inputs:
   # Maximum size of the message received over UDP
   #max_message_size: 10240
 
+#------------------------------ TCP prospector --------------------------------
+# Experimental: Config options for the TCP input
+#- type: tcp
+  #enabled: false
+
+  # The host and port to receive the new event
+  #host: "localhost:9000"
+
+  # Character used to split new message
+  #line_delimiter: "\n"
+
+  # Maximum size in bytes of the message received over TCP
+  #max_message_size: 20971520
+
 #========================== Filebeat autodiscover ==============================
 
 # Autodiscover allows you to detect changes in the system and spawn new modules
--- a/filebeat/docs/filebeat-options.asciidoc
+++ b/filebeat/docs/filebeat-options.asciidoc
@@ -14,7 +14,7 @@ and configuring modules.
 To configure {beatname_uc} manually (instead of using
 <<{beatname_lc}-modules-overview,modules>>), you specify a list of inputs in the
 +{beatname_lc}.inputs+ section of the +{beatname_lc}.yml+. Inputs specify how
-{beatname_uc} locates and processes input data.  
+{beatname_uc} locates and processes input data.
 
 The list is a http://yaml.org/[YAML] array, so each input begins with
 a dash (`-`). You can specify multiple inputs, and you can specify the same
@@ -47,6 +47,7 @@ You can configure {beatname_uc} to use the following inputs:
 * <<{beatname_lc}-input-redis>>
 * <<{beatname_lc}-input-udp>>
 * <<{beatname_lc}-input-docker>>
+* <<{beatname_lc}-input-tcp>>
 
 
 
@@ -59,3 +60,5 @@ include::inputs/input-redis.asciidoc[]
 include::inputs/input-udp.asciidoc[]
 
 include::inputs/input-docker.asciidoc[]
+
+include::inputs/input-tcp.asciidoc[]
--- a/filebeat/filebeat.reference.yml
+++ b/filebeat/filebeat.reference.yml
@@ -535,6 +535,20 @@ filebeat.inputs:
   # Maximum size of the message received over UDP
   #max_message_size: 10240
 
+#------------------------------ TCP prospector --------------------------------
+# Experimental: Config options for the TCP input
+#- type: tcp
+  #enabled: false
+
+  # The host and port to receive the new event
+  #host: "localhost:9000"
+
+  # Character used to split new message
+  #line_delimiter: "\n"
+
+  # Maximum size in bytes of the message received over TCP
+  #max_message_size: 20971520
+
 #========================== Filebeat autodiscover ==============================
 
 # Autodiscover allows you to detect changes in the system and spawn new modules
--- a/filebeat/input/tcp/client.go
+++ b/filebeat/input/tcp/client.go
@@ -22,7 +22,7 @@ type Client struct {
 	done           chan struct{}
 	metadata       common.MapStr
 	splitFunc      bufio.SplitFunc
-	maxReadMessage int64
+	maxReadMessage uint64
 	timeout        time.Duration
 }
 
@@ -31,7 +31,7 @@ func NewClient(
 	conn net.Conn,
 	forwarder *harvester.Forwarder,
 	splitFunc bufio.SplitFunc,
-	maxReadMessage int64,
+	maxReadMessage uint64,
 	timeout time.Duration,
 ) *Client {
 	client := &Client{
@@ -54,7 +54,8 @@ func (c *Client) Handle() error {
 	scanner.Split(c.splitFunc)
 
 	for scanner.Scan() {
-		if scanner.Err() != nil {
+		err := scanner.Err()
+		if err != nil {
 			// we are forcing a close on the socket, lets ignore any error that could happen.
 			select {
 			case <-c.done:
@@ -62,10 +63,10 @@ func (c *Client) Handle() error {
 			default:
 			}
 			// This is a user defined limit and we should notify the user.
-			if IsMaxReadBufferErr(scanner.Err()) {
-				logp.Err("tcp client error: %s", scanner.Err())
+			if IsMaxReadBufferErr(err) {
+				logp.Err("tcp client error: %s", err)
 			}
-			return errors.Wrap(scanner.Err(), "tcp client error")
+			return errors.Wrap(err, "tcp client error")
 		}
 		r.Reset()
 		c.forwarder.Send(c.createEvent(scanner.Text()))
--- a/filebeat/input/tcp/client.go
+++ b/filebeat/input/tcp/client.go
@@ -43,7 +43,7 @@ func NewClient(
 		timeout:        timeout,
 		metadata: common.MapStr{
 			"hostnames":  remoteHosts(conn),
-			"ip_address": c.conn.RemoteAddr().String(),
+			"ip_address": conn.RemoteAddr().String(),
 		},
 	}
 
--- a/filebeat/input/tcp/client.go
+++ b/filebeat/input/tcp/client.go
@@ -1,0 +1,115 @@
+package tcp
+
+import (
+	"bufio"
+	"net"
+	"strings"
+	"time"
+
+	"github.com/pkg/errors"
+
+	"github.com/elastic/beats/filebeat/harvester"
+	"github.com/elastic/beats/filebeat/util"
+	"github.com/elastic/beats/libbeat/beat"
+	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/logp"
+)
+
+// Client is a remote client.
+type Client struct {
+	conn           net.Conn
+	forwarder      *harvester.Forwarder
+	done           chan struct{}
+	metadata       common.MapStr
+	splitFunc      bufio.SplitFunc
+	maxReadMessage int64
+	timeout        time.Duration
+}
+
+// NewClient returns a new client instance for the remote connection.
+func NewClient(
+	conn net.Conn,
+	forwarder *harvester.Forwarder,
+	splitFunc bufio.SplitFunc,
+	maxReadMessage int64,
+	timeout time.Duration,
+) *Client {
+	client := &Client{
+		conn:           conn,
+		forwarder:      forwarder,
+		done:           make(chan struct{}),
+		splitFunc:      splitFunc,
+		maxReadMessage: maxReadMessage,
+		timeout:        timeout,
+	}
+	client.cacheMetadata()
+	return client
+}
+
+// Handle is reading message from the specified TCP socket.
+func (c *Client) Handle() error {
+	r := NewMeteredReader(NewDeadlineReader(c.conn, c.timeout), c.maxReadMessage)
+	buf := bufio.NewReader(r)
+	scanner := bufio.NewScanner(buf)
+	scanner.Split(c.splitFunc)
+
+	for scanner.Scan() {
+		if scanner.Err() != nil {
+			// we are forcing a close on the socket, lets ignore any error that could happen.
+			select {
+			case <-c.done:
+				break
+			default:
+			}
+			// This is a user defined limit and we should notify the user.
+			if IsMaxReadBufferErr(scanner.Err()) {
+				logp.Err("tcp client error: %s", scanner.Err())
+			}
+			return errors.Wrap(scanner.Err(), "tcp client error")
+		}
+		r.Reset()
+		c.forwarder.Send(c.createEvent(scanner.Text()))
+	}
+	return nil
+}
+
+// Close stops reading from the socket and close the connection.
+func (c *Client) Close() {
+	close(c.done)
+	c.conn.Close()
+}
+
+func (c *Client) createEvent(rawString string) *util.Data {
+	data := util.NewData()
+	data.Event = beat.Event{
+		Timestamp: time.Now(),
+		Meta:      c.metadata,
+		Fields: common.MapStr{
+			"message": rawString,
+		},
+	}
+	return data
+}
+
+// GetRemoteHosts take the IP address of the client and try to resolve the name, if it fails we
+// fallback to the IP, IP can resolve to multiple hostname.
+func (c *Client) getRemoteHosts() []string {
+	ip := c.conn.RemoteAddr().String()
+	idx := strings.Index(ip, ":")
+	if idx == -1 {
+		return []string{ip}
+	}
+	ip = ip[0:idx]
+	hosts, err := net.LookupAddr(ip)
+	if err != nil {
+		hosts = []string{ip}
+	}
+	return hosts
+}
+
+func (c *Client) cacheMetadata() {
+	c.metadata = common.MapStr{
+		"hostnames":  c.getRemoteHosts(),
+		"ip_address": c.conn.RemoteAddr().String(),
+	}
+}
--- a/filebeat/input/tcp/client.go
+++ b/filebeat/input/tcp/client.go
@@ -48,7 +48,7 @@ func NewClient(
 
 // Handle is reading message from the specified TCP socket.
 func (c *Client) Handle() error {
-	r := NewMeteredReader(NewDeadlineReader(c.conn, c.timeout), c.maxReadMessage)
+	r := NewResetableLimitedReader(NewDeadlineReader(c.conn, c.timeout), c.maxReadMessage)
 	buf := bufio.NewReader(r)
 	scanner := bufio.NewScanner(buf)
 	scanner.Split(c.splitFunc)
--- a/filebeat/input/tcp/client.go
+++ b/filebeat/input/tcp/client.go
@@ -41,8 +41,12 @@ func NewClient(
 		splitFunc:      splitFunc,
 		maxReadMessage: maxReadMessage,
 		timeout:        timeout,
+		metadata: common.MapStr{
+			"hostnames":  remoteHosts(conn),
+			"ip_address": c.conn.RemoteAddr().String(),
+		},
 	}
-	client.cacheMetadata()
+
 	return client
 }
 
@@ -94,8 +98,8 @@ func (c *Client) createEvent(rawString string) *util.Data {
 
 // GetRemoteHosts take the IP address of the client and try to resolve the name, if it fails we
 // fallback to the IP, IP can resolve to multiple hostname.
-func (c *Client) getRemoteHosts() []string {
-	ip := c.conn.RemoteAddr().String()
+func remoteHosts(conn net.Conn) []string {
+	ip := conn.RemoteAddr().String()
 	idx := strings.Index(ip, ":")
 	if idx == -1 {
 		return []string{ip}
@@ -106,11 +110,4 @@ func (c *Client) getRemoteHosts() []string {
 		hosts = []string{ip}
 	}
 	return hosts
-}
-
-func (c *Client) cacheMetadata() {
-	c.metadata = common.MapStr{
-		"hostnames":  c.getRemoteHosts(),
-		"ip_address": c.conn.RemoteAddr().String(),
-	}
 }
--- a/filebeat/input/tcp/config.go
+++ b/filebeat/input/tcp/config.go
@@ -1,0 +1,25 @@
+package tcp
+
+import (
+	"time"
+
+	"github.com/elastic/beats/filebeat/harvester"
+)
+
+type config struct {
+	harvester.ForwarderConfig `config:",inline"`
+	Host                      string        `config:"host"`
+	LineDelimiter             string        `config:"line_delimiter" validate:"nonzero"`
+	Timeout                   time.Duration `config:"timeout" validate:"nonzero,positive"`
+	MaxMessageSize            int64         `config:"max_message_size" validate:"nonzero,positive"`
+}
+
+var defaultConfig = config{
+	ForwarderConfig: harvester.ForwarderConfig{
+		Type: "tcp",
+	},
+	LineDelimiter:  "\n",
+	Host:           "localhost:9000",
+	Timeout:        time.Second * 5 * 60,
+	MaxMessageSize: 20 * 1024 * 1024,
+}
--- a/filebeat/input/tcp/config.go
+++ b/filebeat/input/tcp/config.go
@@ -20,6 +20,6 @@ var defaultConfig = config{
 	},
 	LineDelimiter:  "\n",
 	Host:           "localhost:9000",
-	Timeout:        time.Second * 5 * 60,
+	Timeout:        time.Minute * 5,
 	MaxMessageSize: 20 * 1024 * 1024,
 }
--- a/filebeat/input/tcp/config.go
+++ b/filebeat/input/tcp/config.go
@@ -11,7 +11,7 @@ type config struct {
 	Host                      string        `config:"host"`
 	LineDelimiter             string        `config:"line_delimiter" validate:"nonzero"`
 	Timeout                   time.Duration `config:"timeout" validate:"nonzero,positive"`
-	MaxMessageSize            int64         `config:"max_message_size" validate:"nonzero,positive"`
+	MaxMessageSize            uint64        `config:"max_message_size" validate:"nonzero,positive"`
 }
 
 var defaultConfig = config{
--- a/filebeat/input/tcp/config.go
+++ b/filebeat/input/tcp/config.go
@@ -8,7 +8,7 @@ import (
 
 type config struct {
 	harvester.ForwarderConfig `config:",inline"`
-	Host                      string        `config:"host"`
+	Host                      string        `config:"host" validate:"required"`
 	LineDelimiter             string        `config:"line_delimiter" validate:"nonzero"`
 	Timeout                   time.Duration `config:"timeout" validate:"nonzero,positive"`
 	MaxMessageSize            uint64        `config:"max_message_size" validate:"nonzero,positive"`
@@ -19,7 +19,6 @@ var defaultConfig = config{
 		Type: "tcp",
 	},
 	LineDelimiter:  "\n",
-	Host:           "localhost:9000",
 	Timeout:        time.Minute * 5,
 	MaxMessageSize: 20 * 1024 * 1024,
 }
--- a/filebeat/input/tcp/conn.go
+++ b/filebeat/input/tcp/conn.go
@@ -15,12 +15,12 @@ var ErrMaxReadBuffer = errors.New("max read buffer reached")
 // error when we reach the limit.
 type MeteredReader struct {
 	reader        io.Reader
-	maxReadBuffer int64
-	byteRead      int64
+	maxReadBuffer uint64
+	byteRead      uint64
 }
 
 // NewMeteredReader returns a new MeteredReader
-func NewMeteredReader(reader io.Reader, maxReadBuffer int64) *MeteredReader {
+func NewMeteredReader(reader io.Reader, maxReadBuffer uint64) *MeteredReader {
 	return &MeteredReader{
 		reader:        reader,
 		maxReadBuffer: maxReadBuffer,
@@ -33,7 +33,7 @@ func (m *MeteredReader) Read(p []byte) (n int, err error) {
 		return 0, ErrMaxReadBuffer
 	}
 	n, err = m.reader.Read(p)
-	m.byteRead += int64(n)
+	m.byteRead += uint64(n)
 	return
 }
 
--- a/filebeat/input/tcp/conn.go
+++ b/filebeat/input/tcp/conn.go
@@ -1,0 +1,72 @@
+package tcp
+
+import (
+	"io"
+	"net"
+	"time"
+
+	"github.com/pkg/errors"
+)
+
+// ErrMaxReadBuffer returns when too many bytes was read on the io.Reader
+var ErrMaxReadBuffer = errors.New("max read buffer reached")
+
+// MeteredReader is based on LimitedReader but allow to reset the byte read and return a specific
+// error when we reach the limit.
+type MeteredReader struct {
+	reader        io.Reader
+	maxReadBuffer int64
+	byteRead      int64
+}
+
+// NewMeteredReader returns a new MeteredReader
+func NewMeteredReader(reader io.Reader, maxReadBuffer int64) *MeteredReader {
+	return &MeteredReader{
+		reader:        reader,
+		maxReadBuffer: maxReadBuffer,
+	}
+}
+
+// Read reads the specified amount of byte
+func (m *MeteredReader) Read(p []byte) (n int, err error) {
+	if m.byteRead >= m.maxReadBuffer {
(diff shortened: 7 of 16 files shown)


ph:
Waiting for CI to switch it to `review`

filebeat/input/tcp/client.go
```diff
@@ -0,0 +1,118 @@
...
+			}
+			// This is a user defined limit and we should notify the user.
+			if IsMaxReadBufferErr(scanner.Err()) {
+				logp.Err("tcp client error: %s", scanner.Err())
+			}
+			return errors.Wrap(scanner.Err(), "tcp client error")
+		}
+		r.Reset()
+		// TODO(ph), should we keep the byte array instead of allocating a string?
+		c.forwarder.Send(c.createEvent(scanner.Text()))
```
ph: This is a legitimate question, I am not sure how the remaining part of the code will work if I keep the message as a byte array.

filebeat/input/tcp/conn.go
```diff
@@ -0,0 +1,72 @@
+package tcp
```
ph: I've keep theses methods public, because I might move it to common.

ph:
Working on fixing the test suite, simple environment difference with how I assert the tests, I will make it more robust. thanks Travis.

filebeat/_meta/common.reference.p2.yml
```diff
@@ -238,6 +238,16 @@ filebeat.prospectors:
...
 
+#------------------------------ Tcp prospector --------------------------------
+# Experimental: Config options for the TCP input
+#- type: tcp
+  #enabled: false
+  # The host and port to receive the new event
+  #host: "localhost:8080"
+
+  # Character used to split new message
+  #line_delimiter: "\n"
```
ruflin: Do we need this to be configurable? I would expect that same behaviour that supports also `\r` etc. like the log prospector / harvester.

filebeat/filebeat.reference.yml
```diff
@@ -507,6 +507,16 @@ filebeat.prospectors:
   # Maximum size of the message received over UDP
   #max_message_size: 10240
 
+#------------------------------ Tcp prospector --------------------------------
```
ruflin: s/Tcp/TCP

filebeat/input/tcp/client.go
ruflin: I would ignore the overhead for now if it adds code complexity.

filebeat/input/tcp/client.go
```diff
@@ -0,0 +1,118 @@
...
+		maxReadBuffer: maxReadBuffer,
+		timeout:       timeout,
+	}
+	client.cacheMetadata()
+	return client
+}
+
+// Handle is reading message from the specified TCP socket.
+func (c *Client) Handle() error {
+	defer c.conn.Close()
```
ruflin: I wonder if this `defer` should really be here or outside where the connection is opened or `Handle` is called.

filebeat/input/tcp/client.go
```diff
@@ -0,0 +1,118 @@
...
+		Timestamp: time.Now(),
+		Meta:      c.metadata,
+		Fields: common.MapStr{
+			"message": rawString,
+		},
+	}
+	return data
+}
+
+// GetRemoveHosts take the IP address of the client and try to resolve the name, if it fails we
```
ruflin: `s/getRemoveHosts/getRemoteHosts`

filebeat/input/tcp/client.go
```diff
@@ -0,0 +1,118 @@
...
+	hosts, err := net.LookupAddr(ip)
+	if err != nil {
+		hosts = []string{ip}
+	}
+	return hosts
+}
+
+func (c *Client) cacheMetadata() {
+	c.metadata = common.MapStr{
+		"hostnames":  c.getRemoteHosts(),
```
ruflin: Should we report all hosts? Should we exclude some?

filebeat/input/tcp/conn.go
ruflin: We should only move it to common if we also use it in `metricbeat` for example, which could happen.

filebeat/tests/system/test_tcp.py
```diff
@@ -0,0 +1,68 @@
+from filebeat import BaseTest
+import socket
+
+
+class Test(BaseTest):
```
ruflin: Nice tests. I already head in mind to ask for this, especially seeing if the shut down works 👍

libbeat/common/scan.go
```diff
@@ -0,0 +1,31 @@
+package common
+
+import (
+	"bufio"
+	"bytes"
+)
+
+// ScanDelimiter return a function to split line using a custom delimiter, the delimiter
+// is stripped from the returned value.
+func ScanDelimiter(delimiter []byte) bufio.SplitFunc {
```
ruflin: I would not put functions into common which we only use in one beat. We should have at least 2 use cases.

ph:
@ruflin You can go ahead with the full review, I was just marking it as in progress to wait for CI. Since I was testing some goroutine interaction I wanted to make it was solid on the CI. I haven't made your initial changes to this PR yet. But go a full review and I will make them in batch. 

Thanks for checking it out.

filebeat/input/tcp/client.go
ph: I am not sure what is the best behavior here:

1. taking the first one could be wrong.
2. taking a subset could also be wrong.
3. They are set in the metadata field for now, so users are feed to use them or not.

filebeat/input/tcp/client.go
ph: Make sense, I will that change when you are done with the review.

filebeat/filebeat.reference.yml
ph: +1 for that change, I've went with `Tcp` even if  golang style would complain because of existing code, I wanted consistency with our `Udp` input. I guess I can make a PR on top of `UDP` to be all caps. :D

filebeat/_meta/common.reference.p2.yml
ph: I believe it needs to be configurable; this input is more than just syslog entries.
If you look at the code by default, it will use the [ScanLines](https://golang.org/src/bufio/scan.go) function to deal with that case. Anything else will use our own implementation of the split.

The default method will detect `\n,` but it will also strip `\r`, I think this is the most common use case with plain sockets protocol. So I think we should be fine(tm)?

The custom delimiter is OK for more esoteric implementation. Maybe in a future revision, we could allow multiple patterns.

filebeat/input/tcp/client.go
ph: For me on my side there is no overhead, it just a matter of calling `scanner.Bytes()` instead of `scanner.Text()`, I am just unaware of the downstream consequence for processor If I keep the bytes.

ph: Actually, I just read the [doc](https://golang.org/src/bufio/scan.go), It's better to do an allocation because the original bytes can be overridden.

```golang
// Bytes returns the most recent token generated by a call to Scan.
  // The underlying array may point to data that will be overwritten
  // by a subsequent call to Scan. It does no allocation.
  func (s *Scanner) Bytes() []byte {
  	return s.token
  }
```

ph:
jenkins test this please

filebeat/filebeat.reference.yml
ruflin: Yep, saw the UDP one when I reviewed one of your other PR's and was thinking the same.

filebeat/_meta/common.reference.p2.yml
ruflin: The feature got requested already several times for the log harvester ...

filebeat/docs/filebeat-options.asciidoc
```diff
@@ -511,6 +512,25 @@ The `enabled` option can be used with each prospector to define if a prospector
...
 
+[float]
+[[line-delimiter]]
+==== `line_delimiter`
+
+When used with `type: tcp`, specifies the the characters used to split the incoming events. The default is '\n'.
+
+[float]
+[[max-read-buffer]]
+==== `max_read_buffer`
```
ruflin: We have `max_message_size` in udp. We should probably use the same config name. Also we can change the udp one.

filebeat/filebeat.reference.yml
```diff
@@ -507,6 +507,16 @@ filebeat.prospectors:
...
+# Experimental: Config options for the TCP input
+#- type: tcp
+  #enabled: false
+  # The host and port to receive the new event
+  #host: "localhost:8080"
```
ruflin: Did you test to have multiple syslog inputs configured?  I assume the following should work:

```
- type: tcp
   host: "localhost:8080"
- type: tcp
   host: "localhost:8081"
```

filebeat/input/tcp/client.go
ruflin: Any chance you could provide some example json output docs in the comment of the PR? I think that makes it easier to understand how the end event will look like.

filebeat/input/tcp/config.go
```diff
@@ -0,0 +1,23 @@
...
+	Timeout                   int    `config:"timeout" validate:"nonzero,positive"`
+	MaxReadBuffer             int64  `config:"max_read_buffer" validate:"nonzero,positive"`
+}
+
+var defaultConfig = config{
+	ForwarderConfig: harvester.ForwarderConfig{
+		Type: "tcp",
+	},
+	LineDelimiter: "\n",
+	Host:          "localhost:9000",
```
ruflin: Is 9000 a "common" default?

filebeat/input/tcp/config.go
```diff
@@ -0,0 +1,23 @@
...
+	Timeout:       5 * 60,
```
ruflin: I think you will have to define the unit here (time.Second)

filebeat/input/tcp/conn.go
```diff
@@ -0,0 +1,72 @@
...
+
+import (
+	"io"
+	"net"
+	"time"
+
+	"github.com/pkg/errors"
+)
+
+// ErrMaxReadBuffer returns when too much byte was read on the io.Reader
```
ruflin: too many bytes?

filebeat/input/tcp/conn.go
```diff
@@ -0,0 +1,72 @@
...
+
+// MeteredReader is based on LimitedReader but allow to reset the byte read and return a specific
+// error when we reach the limit.
+type MeteredReader struct {
+	reader        io.Reader
+	maxReadBuffer int64
+	byteRead      int64
+}
+
+// NewMeteredReader returns a Meteredread
```
ruflin: a  `MeteredReader`

filebeat/input/tcp/harvester.go
```diff
@@ -0,0 +1,158 @@
...
+		server:    server,
+		splitFunc: sf,
+	}, nil
+}
+
+// Run start and run a new TCP listener to receive new data
+func (h *Harvester) Run() error {
+	logp.Info("Started listening for incoming TCP connection on: %s:", h.config.Host)
+	for {
+		conn, err := h.server.Accept()
```
ruflin: I would expect that the prospector/input would be the server and then each harvester would be a client?

filebeat/input/tcp/config.go
ph: @ruflin Not sure about that, I think it's has good as the default we have in UDP. :)

filebeat/filebeat.reference.yml
ph: I haven't tested it, but it should work if every output have their own port.

filebeat/input/tcp/harvester.go
ph: Like we discussed in the input refactoring, I am not sure the wording harvester work in the context of UDP and TCP, I would be to rename `harvester` to `server` and use `client` for the actual remote connection. Would you be OK with that?

My reasoning is `harvester` is more active process to fetch new things (pull), which in the other hand a TCP / UDP is more passive (push).

filebeat/_meta/common.reference.p2.yml
ph: I haven't look at the log harvester how it does the splitting... maybe we can reuse it.

filebeat/docs/filebeat-options.asciidoc
ph: I will rename `max_read_buffer` to `max_message_size`, after looking back at the implementation they are a bit different but the underlying concept is the same, _making sure we don't OOM on incoming messages_

filebeat/_meta/common.reference.p2.yml
ph: I did a quick look at the log harvester, I _believe_ we could use the same scanner/split.

filebeat/filebeat.reference.yml
ph: If both port are free:

```
2018-02-07T11:47:40.997-0500    INFO    tcp/harvester.go:58     Started listening for incoming TCP connection on: localhost:8080`
2018-02-07T11:47:40.997-0500    INFO    input/input.go:87       Starting input of type: tcp; ID: 14733937925171989371
2018-02-07T11:47:40.997-0500    DEBUG   [processors]    processors/processor.go:49      Processors:
2018-02-07T11:47:40.997-0500    INFO    tcp/input.go:50 Starting TCP input
2018-02-07T11:47:40.997-0500    INFO    tcp/harvester.go:58     Started listening for incoming TCP connection on: localhost:8081
2018-02-07T11:47:40.998-0500    DEBUG   [input] log/config.go:178       recursive glob enabled
```

if one port is already taken:

```
2018-02-07T11:48:52.928-0500    ERROR   instance/beat.go:674    Exiting: Error in initing input: listen tcp 127.0.0.1:8080: bind: address already in use
Exiting: Error in initing input: listen tcp 127.0.0.1:8080: bind: address already in use
```
FB exits

filebeat/input/tcp/config.go
ruflin: 👍

filebeat/input/tcp/harvester.go
ruflin: I'm not really concerned here about not using the name harvester but why `input != server`. From my point of view 1 input = 1 server, so if I start an input that should mean the server is started. Instead of having `input`, `server`, `client` I think we only need `input=server`, `client` and can remove one abstraction layer.

I'm not too concerned about not having the name harvester anywhere as this is not something exposed to the user.

ruflin: @ph Now looking at my own UDP code I see that I used also the harvester as the server :-( Leading by bad example? :-(

ruflin:
@ph I suggest we get this PR in rather sooner then later. The internal code structure I brought up above can be discussed and refactored later especially as I realised now the changes would also apply to UDP. Can you go through the other comments above again and see what needs addressing, rebase and squash and ping me for an other review and merge?

filebeat/input/tcp/harvester.go
ph: Not, you lead by good example you see the problem :)

ph:
@ruflin Thanks for answering my comments will do the changes, I wanted to bulk them. :)

ph:
This is an example of the event format from a received events with this input.
```json
 {
  "@timestamp": "2018-02-12T19:12:29.000Z",
  "@metadata": {
    "beat": "filebeat",
    "type": "doc",
    "version": "7.0.0-alpha1",
    "hostnames": [
      "localhost"
    ],
    "ip_address": "127.0.0.1:56576"
  },
  "beat": {
    "hostname": "sashimi",
    "version": "7.0.0-alpha1",
    "name": "sashimi"
  },
  "message": "Hello world",
  "input": {
    "type": "tcp"
  },
  "prospector": {
    "type": "tcp"
  }
}

```

ph:
I am fixing the flaky tests, oh network services.

ruflin:
Thanks for the json output. We should probably also add an example to the docs. I'm a bit on the fence about the `@metadata` usage but it's something we can figure out later.

ph:
@ruflin Let's do a following PR for the JSON example in the doc, not sure where it would be the best place to insert it, inputs doesn't dedicated pages yet, Any idea @dedemorton where I could place this [information](https://github.com/elastic/beats/pull/6266#issuecomment-365030285) ?

ph:
@ruflin Go ahead with the review, there is a failure in the Filebeat test with x-pack, I think the available snapshot isn't correct right now..

```
ERROR: This plugin was built with an older plugin structure. Contact the plugin author to remove the intermediate "elasticsearch" directory within the plugin zip.
```

CHANGELOG.asciidoc
```diff
@@ -48,6 +48,8 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
   the system test. {pull}6121[6121]
 
 - Add IIS module to parse access log and error log. {pull}6127[6127]
+- Refactor the usage of prospector to input in the YAML reference {pull}6121[6121]
```
ruflin: Should this line be here?

filebeat/docs/filebeat-options.asciidoc
```diff
@@ -43,6 +43,7 @@ One of the following input types:
     * redis: Reads slow log entries from redis (experimental).
     * udp: Reads events over UDP. Also see <<max-message-size>>.
     * docker: Reads logs from Docker. Also see <<config-containers>> (experimental).
+    * tcp: Reads events over TCP. Also see <<line-delimiter>>, <<max-message-size>> and <<timeout>> (experimental).
```
ruflin: As we don't have separate pages yet for inputs I wonder if we should mark here that this feature is actually experimental.

@dedemorton Thoughts?

filebeat/docs/filebeat-options.asciidoc
```diff
@@ -509,7 +510,21 @@ The `enabled` option can be used with each prospector to define if a prospector
 [[max-message-size]]
 ==== `max_message_size`
 
-When used with `type: udp`, specifies the maximum size of the message received over UDP. The default is 10240.
+When used with `type: udp` or `type: tcp`, specifies the maximum size of the message received over UDP or TCP.
+
+The default for UDP is 10240 and TCP 20971520.
+
+[float]
+[[line-delimiter]]
```
ruflin: Good for now but we now have really the need to split up config options per input (@dedemorton I remember you were thinking about this).

filebeat/docs/filebeat-options.asciidoc
```diff
@@ -509,7 +510,21 @@ The `enabled` option can be used with each prospector to define if a prospector
...
+==== `line_delimiter`
+
+When used with `type: tcp`, specifies the the characters used to split the incoming events. The default is '\n'.
+
+[float]
+[[timeout]]
+==== `timeout`
+
+The number of seconds of inactivity before a remote connection is closed. The default is 300.
```
ruflin: We should mention that this is for tcp only.

filebeat/input/tcp/client.go
```diff
@@ -0,0 +1,115 @@
...
+)
+
+// Client is a remote client.
+type Client struct {
+	conn           net.Conn
+	forwarder      *harvester.Forwarder
+	done           chan struct{}
+	metadata       common.MapStr
+	splitFunc      bufio.SplitFunc
+	maxReadMessage int64
```
ruflin: Could be a unit64?

filebeat/input/tcp/client.go
```diff
@@ -0,0 +1,115 @@
...
+	for scanner.Scan() {
+		if scanner.Err() != nil {
+			// we are forcing a close on the socket, lets ignore any error that could happen.
+			select {
+			case <-c.done:
+				break
+			default:
+			}
+			// This is a user defined limit and we should notify the user.
+			if IsMaxReadBufferErr(scanner.Err()) {
```
ruflin: I would suggest to store `scanner.Err()` in a variable instead of calling it each time?

filebeat/input/tcp/client.go
```diff
@@ -0,0 +1,115 @@
...
+		r.Reset()
+		c.forwarder.Send(c.createEvent(scanner.Text()))
+	}
+	return nil
+}
+
+// Close stops reading from the socket and close the connection.
+func (c *Client) Close() {
+	close(c.done)
+	c.conn.Close()
```
ruflin: I was wondering if we could have `c.conn.Close()` in the same method where it's opened to make sure we don't forget about it. But not sure if this is possible here.

filebeat/input/tcp/harvester.go
```diff
@@ -0,0 +1,157 @@
...
+		if err != nil {
+			select {
+			case <-h.done:
+				return nil
+			default:
+				logp.Debug("tcp", "Can not accept the connection", err)
```
ruflin: This should be `logp.Debug("tcp", "Can not accept the connection: %s", err)` I think

filebeat/input/tcp/harvester.go
```diff
@@ -0,0 +1,157 @@
...
+		)
+		logp.Debug(
+			"tcp",
+			"New client, remote: %s (total clients: %d)",
+			conn.RemoteAddr(),
+			h.clientsCount(),
+		)
+		h.wg.Add(1)
+		go func() {
+			defer conn.Close()
```
ruflin: Seeing that you also close the connection here, I wonder what happens if `Close` is called twice?

filebeat/input/tcp/harvester.go
```diff
@@ -0,0 +1,157 @@
...
+		client.Close()
+	}
+	h.wg.Wait()
+	logp.Debug("tcp", "Remote connections closed")
+}
+
+func (h *Harvester) registerClient(client *Client) {
+	h.Lock()
+	defer h.Unlock()
+	h.clients[client] = struct{}{}
```
ruflin: Should each client have a unique id instead? Could make it interesting for stats later on.

filebeat/input/tcp/input.go
```diff
@@ -0,0 +1,72 @@
...
+		outlet:    out,
+		harvester: harvester,
+		started:   false,
+	}, nil
+}
+
+// Run start a TCP input
+func (p *Input) Run() {
+	if !p.started {
+		logp.Info("Starting TCP input")
```
ruflin: Want to add the `host` info here?

filebeat/input/tcp/input.go
```diff
@@ -0,0 +1,72 @@
...
+			if err != nil {
+				logp.Err("Error running TCP harvester, error: %s", err)
+			}
+		}()
+	}
+}
+
+// Stop stops TCP server
+func (p *Input) Stop() {
+	logp.Info("Stopping TCP input")
```
ruflin: Add the host info?

filebeat/input/tcp/config.go
```diff
@@ -0,0 +1,25 @@
...
+	MaxMessageSize            int64         `config:"max_message_size" validate:"nonzero,positive"`
+}
+
+var defaultConfig = config{
+	ForwarderConfig: harvester.ForwarderConfig{
+		Type: "tcp",
+	},
+	LineDelimiter:  "\n",
+	Host:           "localhost:9000",
+	Timeout:        time.Second * 5 * 60,
```
kvch: Why don't you write `5 * time.Minute` instead of calculating it?

filebeat/input/tcp/conn.go
```diff
@@ -0,0 +1,72 @@
...
+
+	"github.com/pkg/errors"
+)
+
+// ErrMaxReadBuffer returns when too many bytes was read on the io.Reader
+var ErrMaxReadBuffer = errors.New("max read buffer reached")
+
+// MeteredReader is based on LimitedReader but allow to reset the byte read and return a specific
+// error when we reach the limit.
+type MeteredReader struct {
```
kvch: I am not a big fan of this name. It would be nice to express that it is a special `LimitedReader`.
However, I am not able to come up with a proper name. My best alternative proposal is `ResetterLimitedReader`. I am willing to let the change go.

filebeat/input/tcp/config.go
ph: I can and will.

filebeat/input/tcp/conn.go
ph: Naming is hard but, ResetterLimitedReader or ResetableLimitedReader. I am ok with both.

filebeat/input/tcp/harvester.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+// Stop stops accepting new incoming TCP connection and close any active clients
+func (h *Harvester) Stop() {
+	logp.Info("Stopping TCP harvester")
+	close(h.done)
+	h.server.Close()
+
+	logp.Debug("tcp", "Closing remote connections")
+	for _, client := range h.allClients() {
```
kvch: I think you could eliminate calling `allClients` and simply loop over the map:
```
for client, _ := range h.clients {
    client.Close()
}
```
At this point there can't be any new connection, right? Does locking make sense here? If locking is required, I am ok with keeping the function. But the transformation seems unnecessary.

filebeat/input/tcp/harvester_test.go
```diff
@@ -0,0 +1,266 @@
...
+	o.ch <- data
+	return true
+}
+
+func newTestingOutlet(ch chan *util.Data) *testingOutlet {
+	return &testingOutlet{ch: ch}
+}
+
+// This could be extracted into testing utils and we could add some unicode chars to the charsets.
+func randomString(l int) string {
```
kvch: +1 for extracting this function

filebeat/input/tcp/harvester_test.go
```diff
@@ -0,0 +1,266 @@
+package tcp
```
kvch: Nitpicking, I am completely fine if you ignore this comment:
I have seen that you used top-down design, I am not sure if it was on purpose, it made reviewing your code much easier. Do you mind reordering the functions here, so it would follow the same pattern?

filebeat/docs/filebeat-options.asciidoc
dedemorton: @ruflin Looking at this is at the top of my list for next week.

filebeat/input/tcp/client.go
ph: I did the change, it not required since a scanner keep the current state of the read, so scanner.Err() will always return the same value during a scanner.Scan().

filebeat/input/tcp/harvester.go
ph: I think it still fine in that case, I don't think a unique ID like a **UUID** will give us more than using the ptr.

filebeat/input/tcp/harvester.go
ph: When this code is run we stop refusing new connection but, when the client is shutting down each client are removed from the hash at the end of the [goroutine](https://github.com/elastic/beats/pull/6266/files#diff-d84dc93dff4114566dc2663822cc67adR92). 

So I have two options either I use a copy of the original hash which is what *allClients* is doing or I lock before looping the element in the array this will make all the goroutine wait for the mutex before they can remove themself. I went with the former to keep consistency in the values in the hash during the loop.

filebeat/input/tcp/client.go
ph: I think its necessary to call `c.conn.Close()` here to make sure we don't accept new clients.

ph:
@ruflin @kvch I have update this PR with your comments, tests were running locally but I will keep an eye on the CI.

CHANGELOG.asciidoc
```diff
@@ -46,8 +46,8 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
 - Refactor the Stdin prospector to use the input interface {pull}6116[6116]
 - Refactor the usage of prospector to input in the YAML reference and
   the system test. {pull}6121[6121]
-
```
ph: before @ruflin you comment on it :) https://github.com/elastic/beats/pull/6427

filebeat/input/tcp/harvester_test.go
ph: I will wait when we add a bit more feature to it.

filebeat/input/tcp/client.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/elastic/beats/filebeat/util"
+	"github.com/elastic/beats/libbeat/beat"
+	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/logp"
+)
+
+// Client is a remote client.
+type Client struct {
+	conn           net.Conn
+	forwarder      *harvester.Forwarder
```
urso: can we replace `*harvester.Forwarder` by an interface or function type? Some more isolation from harvester.Forwarder (I hope to remove Forwarder and Outlet in the future).

filebeat/input/tcp/client.go
urso: Agree on async Close. Still, double check by running/compiling with `-race` the async Close does not trigger race condition warnings.

filebeat/input/tcp/client.go
ph: Knowing that it make sense to change it.

filebeat/input/tcp/client.go
```diff
@@ -0,0 +1,116 @@
...
+func (c *Client) Close() {
+	close(c.done)
+	c.conn.Close()
+}
+
+func (c *Client) createEvent(rawString string) *util.Data {
+	data := util.NewData()
+	data.Event = beat.Event{
+		Timestamp: time.Now(),
+		Meta:      c.metadata,
```
urso: The `metadata` field behaves like `@metadata` in Logstash. If one sets `index` or `id` or `pipeline`, this will be configure some paramters in the Elasticsearch output. This on purpose?

filebeat/input/tcp/client.go
urso: Also wonder if we want to opt-out from resolving the hostname.

filebeat/input/tcp/client.go
```diff
@@ -0,0 +1,116 @@
...
+		Fields: common.MapStr{
+			"message": rawString,
+		},
+	}
+	return data
+}
+
+// GetRemoteHosts take the IP address of the client and try to resolve the name, if it fails we
+// fallback to the IP, IP can resolve to multiple hostname.
+func (c *Client) getRemoteHosts() []string {
```
urso: do we need a method on Client for this one?

filebeat/input/tcp/client.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+	ip = ip[0:idx]
+	hosts, err := net.LookupAddr(ip)
+	if err != nil {
+		hosts = []string{ip}
+	}
+	return hosts
+}
+
+func (c *Client) cacheMetadata() {
```
urso: `cacheMetadata` is quite short and only executed once by the constructor. How about moving the body into the constructor and removing the function.

Alternatively have it be a function returning a `common.MapStr`?

filebeat/input/tcp/harvester.go
```diff
@@ -0,0 +1,150 @@
...
+	"net"
+	"sync"
+
+	"github.com/elastic/beats/libbeat/logp"
+
+	"github.com/elastic/beats/filebeat/harvester"
+)
+
+// Harvester represent a TCP server
+type Harvester struct {
```
urso: Let us not export the `Harvester`. It's some internal detail specific to the input. How about renaming to `server`?

Do we actually need a special `Harvester/server` type? The `Input` implementations seems to mostly wrap the server.

filebeat/input/tcp/harvester.go
```diff
@@ -0,0 +1,150 @@
...
+				logp.Debug("tcp", "Client error: %s", err)
+			}
+			h.unregisterClient(client)
+			logp.Debug(
+				"tcp",
+				"Client disconnected, remote: %s (total clients: %d)",
+				conn.RemoteAddr(),
+				h.clientsCount(),
+			)
+			h.wg.Done()
```
urso: better put `defer h.wg.Done()` at the top of the go-func. Also try to capture panics and print an error. If something goes really wrong with one client, the other clients can still be served.

Same for `h.unregisterClient()`.

e.g. 

```
go func() {
  defer logp.Recover()

  defer h.wg.Done()
  defer conn.Close()
  
  h.registerClient()
  defer h.unregisterClient()

  err := client.Handle()
  ...
}()
```

filebeat/input/tcp/harvester.go
ph: I would leave it as is and rename the harvester to server similar when doing #6361 similar to #6439, But I can do it right now if you wish.

filebeat/input/tcp/client.go
ph: I will go with the `common.MapStr`.

filebeat/input/tcp/client.go
ph: I did it on purpose, I am not sure how valuable that information it is. but I could make it public field and store in the mapping. WDYT?

filebeat/input/tcp/client.go
ph: Similar to #6439  ?:)
I was planning to do it in second part but I can do it right now.

filebeat/input/tcp/client.go
ph: Right it could be costly or slow.

filebeat/input/tcp/client.go
ph: I don't see the need to make it a simple function or if It can be reused elsewhere, it will called once and we cache the result.

filebeat/input/tcp/client.go
ruflin: let's keep the PR as small as possible and do the interface when we have a second implementation.

ruflin:
@ph Can you ping us when this needs an other look?

filebeat/input/tcp/client.go
ph: I've manually tested with a build with `-race` and run the test with `-race`, didn't see any race condition warnings.

ph:
@ruflin @urso 

I have rebased this PR and included your latest comments.

Open questions:

Should we extract the TCP logic into his own class in this PR, like I did in refactor of #6439?
I would prefer to do it into his own PR and get this one merged.

What do we do about `metadata` concerning the [remote hosts](https://github.com/elastic/beats/pull/6266#discussion_r169656151), keep it there, use an actual field from ECS?

I would like to move this forward or at least have a plan to unblock #6433.

ph:
@ruflin @urso Don't check it out yet, I will make some changes I've learned from @andrewkroh

ph:
I think it make sense to reopen a new one, since structure has changes locally to follow the udp refactor.

PR #6283, opened 2018-02-05
MongoDB module

kvch:
I added a new Filebeat module for MongoDB what I created during testing the FB module generator tooling.

This is the dashboard I created:
filebeat-mongodb-overview

It does not look good, as I hasn't been able to get real error messages from my instance.


filebeat/module/mongodb/log/_meta/fields.yml
```diff
@@ -0,0 +1,27 @@
+- name: log
+  type: group
+  description: >
+      Contains fields from MongoDB logs.
+  fields:
+  - name: timestamp
```
ruflin: Should we use `@timestamp` for this?

filebeat/module/mongodb/log/_meta/fields.yml
```diff
@@ -0,0 +1,27 @@
...
+    description: >
+        Functional categorization of message
+    example: COMMAND
+    type: keyword
+  - name: context
+    description: >
+        Context of message
+    example: initandlisten
+    type: keyword
+  - name: message
```
ruflin: We should use the common `message` field here.

filebeat/module/mongodb/log/_meta/fields.yml
kvch: What do you mean?

ruflin: Defining `timestamp` here means it will be `mongodb.log.timestamp` in the event. But normally the processed timestamp from the log line when the event happened ends up in `@timestamp`. Is this a different timestamp here?

kvch: I think a separate `timestamp` field is required, because it stores the actual time the log was generated. `@timestamp` is the time it was processed. These two times are not necessarily the same.

filebeat/module/mongodb/log/_meta/fields.yml
kvch: I dropped `type: text`.

filebeat/module/mongodb/log/_meta/fields.yml
ruflin: We have similar behaviour in other modules, for example Kafka. What we do there is that we have the timestamp of the actual log line in `@timestamp` and the one when the log line was actually read in `read_timestamp`. Name of `read_timestamp` is not great but we don't have a better one yet.

If you look at the ingest pipeline of the Kafka module you will see the the timestamps are moved around to make sure `@timestamp` is the one after processing.

filebeat/module/mongodb/log/_meta/fields.yml
```diff
@@ -0,0 +1,26 @@
...
+      Contains fields from MongoDB logs.
+  fields:
+  - name: timestamp
+    description: >
+        Timestamp of message
+  - name: severity
+    description: >
+        Severity level of message
+    example: I
+    type: keyword
```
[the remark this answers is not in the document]
kvch: I came up with field names based on MongoDB logging guide: https://docs.mongodb.com/manual/reference/log-messages/ It might provide more context on the subject.
However, I agree that it is a bit weird in MongoDB. If you look at the section "Severity", you can see the taxonomy of log levels. So I am not sure why it's named severity.
Does it cause problem in ECS?

kvch:
@ruflin I think it's ready for a new round of review. Travis fails, because of XPack installation problems.

filebeat/module/mongodb/_meta/config.yml
```diff
@@ -0,0 +1,8 @@
+- module: mongodb
+  # All logs
```
ruflin: What does all logs mean here? Does mongodb only have 1 type of logs?

filebeat/module/mongodb/log/_meta/fields.yml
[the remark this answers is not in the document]
ruflin: I think we can keep it as is as it's under the mongodb namespace. In case we map this to ECS later it would end up in `log.level` from my perspective. No need to solve this now.

filebeat/module/mongodb/log/ingest/pipeline.json
```diff
@@ -0,0 +1,41 @@
+{
+  "description": "Pipeline for parsing MongoDB log logs",
```
ruflin: Mongodb log logs -> I think I start to understand why you call the filset `log`. What exactly is in these logs?

filebeat/module/mongodb/_meta/config.yml
kvch: Yes, that's all I found in their docs: https://docs.mongodb.com/manual/reference/log-messages/

filebeat/module/mongodb/log/_meta/fields.yml
kvch: I readded `text`.

PR #6284, opened 2018-02-05
Update filebeat python tests to use the new field `input(s)` instead of `prospector(s)`

ph:
Update all python based tests to use `input` instead of prospectors,
All test were changed to use the new options in the YAML configuration.
This commit also fixes an issue when using the
`filebeat.config.inputs` field instead of the
`filebeat.config.prospectors` and add a deprecation test for it.


ph:
failures is fixed in #6318

PR #6292, opened 2018-02-06
Add rename fields processor

ruflin:
The rename processor allows to rename fields before they are indexed to standardise on names or move fields around. This becomes useful when building filebeat modules which read from json files. With the rename processor no ingest pipeline is needed to follow the naming schema. This should make some modules simpler to build. It's also useful in combination with #6024 to rename some fields according to the schema.

```
processors:
- rename:
    fields:
     - from: "a"
       to: "b"
```

Intention of rename
* Adjust fields to mapping
* Prevent conflicts like `a` and `a.b` by renaming `a` to `a.value`

Limitations
* Will not overwrite keys

### DIFF
--- a/libbeat/_meta/config.reference.yml
+++ b/libbeat/_meta/config.reference.yml
@@ -137,6 +137,14 @@
 #       equals:
 #           http.code: 200
 #
+# The following example renames the field a to b:
+#
+#processors:
+#- rename:
+#    fields:
+#       - from: "a"
+#         to: "b"
+#
 # The following example enriches each event with metadata from the cloud
 # provider about the host machine. It works on EC2, GCE, DigitalOcean,
 # Tencent Cloud, and Alibaba Cloud.
--- a/libbeat/docs/processors-using.asciidoc
+++ b/libbeat/docs/processors-using.asciidoc
@@ -43,7 +43,7 @@ The supported processors are:
  * <<drop-event,`drop_event`>>
  * <<drop-fields,`drop_fields`>>
  * <<include-fields,`include_fields`>>
- * <<rename,`rename`>>
+ * <<rename-fields,`rename`>>
  * <<add-kubernetes-metadata,`add_kubernetes_metadata`>>
  * <<add-docker-metadata,`add_docker_metadata`>>
  * <<add-host-metadata,`add_host_metadata`>>
--- a/libbeat/docs/processors-using.asciidoc
+++ b/libbeat/docs/processors-using.asciidoc
@@ -535,23 +535,20 @@ the required fields, `@timestamp` and `type`, are exported.
 [[rename-fields]]
 === Rename fields from events
 
-The `rename` processor specifies a list of fields that should be renamed. The
-list of fields to be renamed are listed under `fields`. Each entry contains a
-`from: old-key` and a `to: new-key` pair. `from` is the origin and `to` the
-target name of the field.
-
-The renaming of fields can be useful in case two fields potentially conflict.
+The `rename` processor specifies a list of fields to rename. Under the `fields`
+key each entry contains a `from: old-key` and a `to: new-key` pair. `from` is
+the origin and `to` the target name of the field.
 
 Renaming fields can be useful in cases where field names cause conflicts. For
-example if an event has two fields, `a` and `a.b`, that are both assigned a
-scalar values (e.g. `{"a": 1, "a.b": 2}`) this will result in an error at ingest
-time from Elasticsearch. This is because the value of a cannot simultaneously be
-a scalar and an object. To prevent this rename_fields can be used to rename a to
-a.value.
+example if an event has two fields, `c` and `c.b`, that are both assigned scalar
+values (e.g. `{"c": 1, "c.b": 2}`) this will result in an Elasticsearch error at
+ingest time. This is because the value of a cannot simultaneously be a scalar
+and an object. To prevent this rename_fields can be used to rename `c` to
+`c.value`.
 
 Rename fields cannot be used to overwrite fields. To overwrite fields either
-first rename the target field or use the drop_fields processor to drop the field
-and then rename the field.
+first rename the target field or use the `drop_fields` processor to drop the
+field and then rename the field.
 
 [source,yaml]
 -------
--- a/libbeat/docs/processors-using.asciidoc
+++ b/libbeat/docs/processors-using.asciidoc
@@ -43,6 +43,7 @@ The supported processors are:
  * <<drop-event,`drop_event`>>
  * <<drop-fields,`drop_fields`>>
  * <<include-fields,`include_fields`>>
+ * <<rename,`rename`>>
  * <<add-kubernetes-metadata,`add_kubernetes_metadata`>>
  * <<add-docker-metadata,`add_docker_metadata`>>
  * <<add-host-metadata,`add_host_metadata`>>
@@ -54,8 +55,8 @@ Each condition receives a field to compare. You can specify multiple fields
 under the same condition by using `AND` between the fields (for example,
 `field1 AND field2`).
 
-For each field, you can specify a simple field name or a nested map, for
-example `dns.question.name`.
+For each field, you can specify a simple field name or a nested map, for example
+`dns.question.name`.
 
 See <<exported-fields>> for a list of all the fields that are exported by
 {beatname_uc}.
@@ -530,6 +531,52 @@ section.
 
 NOTE: If you define an empty list of fields under `include_fields`, then only
 the required fields, `@timestamp` and `type`, are exported.
+
+[[rename-fields]]
+=== Rename fields from events
+
+The `rename` processor specifies a list of fields that should be renamed. The
+list of fields to be renamed are listed under `fields`. Each entry contains a
+`from: old-key` and a `to: new-key` pair. `from` is the origin and `to` the
+target name of the field.
+
+The renaming of fields can be useful in case two fields potentially conflict.
+
+Renaming fields can be useful in cases where field names cause conflicts. For
+example if an event has two fields, `a` and `a.b`, that are both assigned a
+scalar values (e.g. `{"a": 1, "a.b": 2}`) this will result in an error at ingest
+time from Elasticsearch. This is because the value of a cannot simultaneously be
+a scalar and an object. To prevent this rename_fields can be used to rename a to
+a.value.
+
+Rename fields cannot be used to overwrite fields. To overwrite fields either
+first rename the target field or use the drop_fields processor to drop the field
+and then rename the field.
+
+[source,yaml]
+-------
+processors:
+- rename:
+    fields:
+     - from: "a.g"
+       to: "e.d"
+    ignore_missing: false
+    fail_on_error: true
+-------
+
+The `rename` processor has the following configuration settings:
+
+`ignore_missing`:: (Optional) If set to true, no error is logged in case a key
+which should be renamed is missing. Default is `false`.
+
+`fail_on_error`:: (Optional) If set to true, in case of an error the renaming of
+fields is stopped and the original event is returned. If set to false, renaming
+continues also if an error happened during renaming. Default is `true`.
+
+See <<conditions>> for a list of supported conditions.
+
+You can specify multiple `ignore_missing` processors under the `processors`
+section.
 
 [[add-kubernetes-metadata]]
 === Add Kubernetes metadata
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -100,6 +100,7 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
 - Add support for spooling to disk to the beats event publishing pipeline. {pull}6581[6581]
 - Added logging of system info at Beat startup. {issue}5946[5946]
 - Do not log errors if X-Pack Monitoring is enabled but Elastisearch X-Pack is not. {pull}6627[6627]
+- Add rename processor. {pull}6292[6292]
 
 *Auditbeat*
 
--- a/auditbeat/auditbeat.reference.yml
+++ b/auditbeat/auditbeat.reference.yml
@@ -242,6 +242,14 @@ auditbeat.modules:
 #       equals:
 #           http.code: 200
 #
+# The following example renames the field a to b:
+#
+#processors:
+#- rename:
+#    fields:
+#       - from: "a"
+#         to: "b"
+#
 # The following example enriches each event with metadata from the cloud
 # provider about the host machine. It works on EC2, GCE, DigitalOcean,
 # Tencent Cloud, and Alibaba Cloud.
--- a/filebeat/filebeat.reference.yml
+++ b/filebeat/filebeat.reference.yml
@@ -741,6 +741,14 @@ filebeat.inputs:
 #       equals:
 #           http.code: 200
 #
+# The following example renames the field a to b:
+#
+#processors:
+#- rename:
+#    fields:
+#       - from: "a"
+#         to: "b"
+#
 # The following example enriches each event with metadata from the cloud
 # provider about the host machine. It works on EC2, GCE, DigitalOcean,
 # Tencent Cloud, and Alibaba Cloud.
--- a/heartbeat/heartbeat.reference.yml
+++ b/heartbeat/heartbeat.reference.yml
@@ -351,6 +351,14 @@ heartbeat.scheduler:
 #       equals:
 #           http.code: 200
 #
+# The following example renames the field a to b:
+#
+#processors:
+#- rename:
+#    fields:
+#       - from: "a"
+#         to: "b"
+#
 # The following example enriches each event with metadata from the cloud
 # provider about the host machine. It works on EC2, GCE, DigitalOcean,
 # Tencent Cloud, and Alibaba Cloud.
--- a/libbeat/processors/actions/rename.go
+++ b/libbeat/processors/actions/rename.go
@@ -1,0 +1,104 @@
+package actions
+
+import (
+	"fmt"
+
+	"github.com/pkg/errors"
+
+	"github.com/elastic/beats/libbeat/beat"
+	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/common/cfgwarn"
+	"github.com/elastic/beats/libbeat/logp"
+	"github.com/elastic/beats/libbeat/processors"
+)
+
+type renameFields struct {
+	config renameFieldsConfig
+}
+
+type renameFieldsConfig struct {
+	Fields        []fromTo `config:"fields"`
+	IgnoreMissing bool     `config:"ignore_missing"`
+	FailOnError   bool     `config:"fail_on_error"`
+}
+
+type fromTo struct {
+	From string `config:"from"`
+	To   string `config:"to"`
+}
+
+func init() {
+	processors.RegisterPlugin("rename",
+		configChecked(newRenameFields,
+			requireFields("fields")))
+}
+
+func newRenameFields(c *common.Config) (processors.Processor, error) {
+
+	cfgwarn.Beta("Beta rename processor is used.")
+	config := renameFieldsConfig{
+		IgnoreMissing: false,
+		FailOnError:   true,
+	}
+	err := c.Unpack(&config)
+	if err != nil {
+		return nil, fmt.Errorf("failed to unpack the rename configuration: %s", err)
+	}
+
+	f := &renameFields{
+		config: config,
+	}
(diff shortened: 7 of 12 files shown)


libbeat/processors/actions/rename_fields.go
```diff
@@ -0,0 +1,69 @@
...
+
+		err = event.Fields.Delete(field.From)
+		if err != nil {
+			logp.Err("Could not delete key: %s,  %+v", field.From, err)
+			continue
+		}
+	}
+
+	return event, nil
+}
```
ph: In the context of parsing multiple JSON document, I think that logging the error everytime could be quite noisy. I think we should add `ignore_missing`, this would be similar to what ingest provide if the key doesn't exist in multiple processors. See the [rename processor](https://www.elastic.co/guide/en/elasticsearch/reference/master/rename-processor.html)

libbeat/processors/actions/rename_fields.go
```diff
@@ -0,0 +1,69 @@
...
+	config := struct {
+		Fields []FieldTuple `config:"fields"`
+	}{}
+	err := c.Unpack(&config)
+	if err != nil {
+		return nil, fmt.Errorf("fail to unpack the rename_fields configuration: %s", err)
+	}
+
+	// TODO: Check if some fields are not allowed to be renamed
+
```
ph: What fields cannot be renamed?

libbeat/processors/actions/rename_fields.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/logp"
+	"github.com/elastic/beats/libbeat/processors"
+)
+
+type renameFields struct {
+	Fields []FieldTuple
+}
+
+type FieldTuple struct {
```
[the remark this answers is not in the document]
ph: +1 to hide it.

ph:
+1 for `ignore_missing`, but set it to false by default.

libbeat/processors/actions/rename_fields.go
```diff
@@ -0,0 +1,69 @@
...
+	f := &renameFields{Fields: config.Fields}
+	return f, nil
+}
+
+func (f *renameFields) Run(event *beat.Event) (*beat.Event, error) {
+
+	for _, field := range f.Fields {
+		value, err := event.Fields.GetValue(field.From)
+		if err != nil {
+			logp.Err("Could not fetch from key: %s", field.From)
```
andrewkroh: Every key that does not exists will be an error. That seems to verbose to me. Maybe you want to handle the error types separately -- log a debug message if the key does not exists and log an error for anything else.

libbeat/processors/actions/rename_fields.go
ruflin: Not sure if we should allow to rename `@timestamp` for example.

libbeat/processors/actions/rename_fields.go
ruflin: I think adding the `ignore_missing` config option should solve this? There is still the issue of what if there is an "other" error. Need to check if any other errors can happen.

libbeat/processors/actions/rename_fields.go
ph: @ruflin are you planning to block the following two options:
```yaml
     - from: "@timestamp"
       to: "another_timestamp"
     - from: "good_timestamp"
       to: "@timstamp"
```

The last one would require time parsing?

libbeat/processors/actions/rename_fields.go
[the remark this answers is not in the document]
ruflin: done

libbeat/processors/actions/rename_fields.go
ruflin: I was initially thinking about it but now thinking we should not add any artificial limitations here and lets see what happens.

ruflin:
I changed / simplified the config a bit and would like to get some feedback. It now looks like below:

```
processors:
- rename_fields:
    fields:
     - "a.b": "c.d"
    ignore_missing: false
- rename_fields:
    fields:
     - "a.g": "e.d"
    ignore_missing: true
```

Writing tests I realised the from / to was kind of too long and not needed. Using just key / value pairs is enough and makes it very easy to specify the values.

The ignore_missing is per processor instance, so if only for a few fields it should be ignored, the processor can be set twice.

libbeat/processors/actions/rename_fields_test.go
```diff
@@ -0,0 +1,96 @@
...
+				"a": common.MapStr{
+					"b": common.MapStr{
+						"c": 12,
+					},
+				},
+			},
+			error: false,
+		},
+		{
+			// This overwrites an existing field
```
ruflin: @andrewkroh @ph I wonder what our expected behaviour here is. We could later add an option "prevent overwrite" or something if it becomes an issue.

libbeat/processors/actions/rename_fields.go
```diff
@@ -0,0 +1,76 @@
...
+
+	f := &renameFields{
+		Fields:        config.Fields,
+		ignoreMissing: config.IgnoreMissing,
+	}
+	return f, nil
+}
+
+func (f *renameFields) Run(event *beat.Event) (*beat.Event, error) {
+
```
ph: space?

libbeat/processors/actions/rename_fields.go
ph: I  see we diverge from the `ignore_missing` semantic from the ingest pipeline, in their case if the flag is set to false its a fatal error. This might be OK in our usecase.

libbeat/processors/actions/rename_fields_test.go
ph: Well, I think we should aim for the **least surprise**, what I mean, is we should fails if the target field already exist. They can always use a `drop_field` before a rename to ensure it always works?

Reference:
https://github.com/elastic/elasticsearch/blob/master/modules/ingest-common/src/main/java/org/elasticsearch/ingest/common/RenameProcessor.java#L68-L75

libbeat/common/mapstr.go
```diff
@@ -322,6 +322,17 @@ func walkMap(key string, data MapStr, op mapStrOperation) (interface{}, error) {
 	var err error
 	keyParts := strings.Split(key, ".")
 
+	// Only check for full key if key contains dots
+	if len(keyParts) > 1 {
+		// Converted dotted key in object to nested object
+		// This is needed for processors to support dotted keys
```
andrewkroh: Can you add a test case to the MapStr tests for this. Is this needed for data that has not been de-dotted (the key names contain dots)?

libbeat/processors/actions/rename_fields.go
ruflin: It's going to be fine as long as you don't tell @andrewkroh :-D

libbeat/processors/actions/rename_fields.go
ruflin: What happens in the case of a fatal error? The original event is stored or the processed one until the error happened? We could have an additional config ignore_errors.

ph: @ruflin Good question, I do not have good answer, Do we have any other existing processor that implement this kind of behavior?

libbeat/processors/actions/rename_fields.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	return event, nil
+}
+
+func (f *renameFields) renameField(from string, to string, fields common.MapStr) error {
+
+	// Fields cannot be overwritten. Either the target field has to be dropped first or renamed first
+	exists, err := fields.HasKey(to)
+	// TODO: Why is it an error if HasKey is false?
```
andrewkroh: `HasKey` returns true iff (if and only if) the key exists, so you can ignore the error.

The error is just some extra information to go along with the bool. For example `HasKey` will return an error if the object has a `x.y` and you ask for `x.y.z`, because it expects `y` to be an object, not a scalar.

libbeat/processors/actions/rename_fields.go
```diff
@@ -0,0 +1,101 @@
...
+	//	return fmt.Errorf("has key error: %s", err)
+	//}
+	if exists {
+		return fmt.Errorf("Target field %s already exists, drop or rename this field first.", to)
+	}
+
+	value, err := fields.GetValue(from)
+	if err != nil {
+		// Ignore ErrKeyNotFound errors
+		if errors.Cause(err) == common.ErrKeyNotFound && f.ignoreMissing {
```
andrewkroh: I'd swap the order here to take advantage of the short-circuit. When `ignoreMissing==false` you can avoid calling `errors.Cause`.

libbeat/processors/actions/rename_fields.go
```diff
@@ -0,0 +1,101 @@
...
+func (f *renameFields) renameField(from string, to string, fields common.MapStr) error {
+
+	// Fields cannot be overwritten. Either the target field has to be dropped first or renamed first
+	exists, err := fields.HasKey(to)
+	// TODO: Why is it an error if HasKey is false?
+	//if err != nil {
+	//	return fmt.Errorf("has key error: %s", err)
+	//}
+	if exists {
+		return fmt.Errorf("Target field %s already exists, drop or rename this field first.", to)
```
andrewkroh: Error strings should not be capitalized. (there are few of these) https://github.com/golang/go/wiki/CodeReviewComments#error-strings

libbeat/processors/actions/rename_fields.go
```diff
@@ -0,0 +1,101 @@
...
+
+func newRenameFields(c *common.Config) (processors.Processor, error) {
+	config := struct {
+		Fields        map[string]string `config:"fields"`
+		IgnoreMissing bool              `config:"ignore_missing"`
+		FailOnError   bool              `config:"fail_on_error"`
+	}{}
+	err := c.Unpack(&config)
+	if err != nil {
+		return nil, fmt.Errorf("fail to unpack the rename_fields configuration: %s", err)
```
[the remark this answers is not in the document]
ruflin: changed

ruflin:
@ph @andrewkroh I updated the PR with more tests, added docs and implemented the config options `ignore_missing` and `fail_on_error`. I would appreciate some feedback on the config options to hear if these are enough, too much or others are needed.

libbeat/processors/actions/rename_fields.go
```diff
@@ -0,0 +1,99 @@
...
+			event.Fields = backup
+			return event, err
+		}
+	}
+
+	return event, nil
+}
+
+func (f *renameFields) renameField(from string, to string, fields common.MapStr) error {
+
```
ph: space?

libbeat/docs/processors-using.asciidoc
```diff
@@ -513,6 +514,41 @@ section.
 NOTE: If you define an empty list of fields under `include_fields`, then only
 the required fields, `@timestamp` and `type`, are exported.
 
+[[rename-fields]]
+=== Rename fields from events
+
+The `rename_fields` processor specifies a list of fields which should be renamed. The list of fields to be
```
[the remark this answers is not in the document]
ruflin: done

ruflin:
@andrewkroh @ph New version pushed

ruflin:
Please do not merge it yet. I did run some additional manual tests and run into issue with dotted keys in yaml. Perhaps the config needs to change again :-(

libbeat/docs/processors-using.asciidoc
```diff
@@ -513,6 +514,45 @@ section.
...
+=== Rename fields from events
+
+The `rename_fields` processor specifies a list of fields that should be renamed.
+The list of fields to be renamed are listed under `fields`. Each entry contains a `from: old-key` and a `to: new-key` pair. `from` is the origin and `to` the target name of the field.
+
+The renaming of fields can be useful in case two fields potentially conflict.
+Having a field with the name `a:2` and `a.b:4` will return an error from
+Elasticsearch during ingest time as `a` would be an integer and object at the
+same time. To prevent this the rename processor can be used to rename `a` to
+`a.value` for example.
```
andrewkroh: I was struggling on "Having a field with the name `a:2` and `a.b:4`" so I have a suggestion for the paragraph.

Renaming fields can be useful in cases where field names cause conflicts. For example if an event has two fields `a` and `a.b` that are both assigned a scalar values (e.g. `{"a": 1, "a.b": 2}`) this will result in an error at ingest time from Elasticsearch. This is because the value of `a` cannot simultaneously be a scalar and an object. To prevent this `rename_fields` can be used to rename `a` to `a.value`.

libbeat/docs/processors-using.asciidoc
```diff
@@ -513,6 +514,45 @@ section.
 NOTE: If you define an empty list of fields under `include_fields`, then only
 the required fields, `@timestamp` and `type`, are exported.
 
+[[rename-fields]]
+=== Rename fields from events
+
+The `rename_fields` processor specifies a list of fields that should be renamed.
+The list of fields to be renamed are listed under `fields`. Each entry contains a `from: old-key` and a `to: new-key` pair. `from` is the origin and `to` the target name of the field.
```
andrewkroh: Can you please fix the line wrapping again. Also below, the config options lines like `ignore_missing` are long too.

ph:
@ruflin I took another look at this PR, concerning the `from/to` combo we could use `source/target` or we could use the same naming as the [ingest processor](https://www.elastic.co/guide/en/elasticsearch/reference/master/rename-processor.html).
```
{
  "rename": {
    "field": "foo",
    "target_field": "foobar"
  }
}
```

We could also drop the `_field` and use `rename` instead of rename field.

libbeat/processors/actions/rename_fields.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+func init() {
+	processors.RegisterPlugin("rename_fields",
+		configChecked(newRenameFields,
+			requireFields("fields")))
+}
+
+func newRenameFields(c *common.Config) (processors.Processor, error) {
+	config := struct {
```
andrewkroh: There's some duplication in this anonymous struct and the `renameFields` struct. Perhaps you could elevate this anonymous part to `type renameFieldsConfig struct` and then embed it in the `renameFields` processor.

```
type renameFields struct {
    renameFieldsConfig
}

type renameFieldsConfig struct {
    Fields        []fromTo `config:"fields"`
    IgnoreMissing bool     `config:"ignore_missing"`
    FailOnError   bool     `config:"fail_on_error"`
}
```

ruflin:
@ph +1 on dropping `_field`. I got inspired by `drop_fields` but the shorter version seems to be better.

For the ingest configs I struggle a bit with `target_field`. The reason I ended up with `from/to` is that it is pretty easy to read. `rename` -> `fields` -> `from`  x -> `to` -> `y`. source/target is a bit longer but also an option :-) Naming is hard.

ph:
> Now that you suggested removing _fields I think I'm in favor again of from/to. Naming is hard.

I am OK with that go ahead :)

libbeat/docs/processors-using.asciidoc
ruflin: Done, thanks for the proposal.

libbeat/docs/processors-using.asciidoc
ruflin: Done. I hoped to put this in our editor_config to have it automatically but it's not supported yet. https://github.com/editorconfig/editorconfig/wiki/Property-research:-maximum-line-length

Also pity that IntellJ does not support it based on file types: https://youtrack.jetbrains.com/issue/IDEA-89608

libbeat/processors/actions/rename_fields.go
ruflin: done

libbeat/_meta/config.reference.yml
```diff
@@ -137,6 +137,14 @@
 #       equals:
 #           http.code: 200
 #
+# The following example renames the field a to b:
```
ruflin: @dedemorton The processor part in the reference file becomes too extensive in my opinion. We should find a way to link here to the docs instead. WDYT?

ruflin:
@ph @andrewkroh PR updated and rebased again on master.

libbeat/_meta/config.reference.yml
dedemorton: @ruflin I agree with you. We need to be careful about linking to docs, tho. We need to make sure the links get validated. I can investigate. We probably shouldn't hold up the merge of this PR for that, tho.

libbeat/docs/processors-using.asciidoc
```diff
@@ -531,6 +532,52 @@ section.
 NOTE: If you define an empty list of fields under `include_fields`, then only
 the required fields, `@timestamp` and `type`, are exported.
 
+[[rename-fields]]
+=== Rename fields from events
+
+The `rename` processor specifies a list of fields that should be renamed. The
```
dedemorton: Would be better to say, "...specifies a list of fields to rename."

libbeat/docs/processors-using.asciidoc
```diff
@@ -531,6 +532,52 @@ section.
...
+list of fields to be renamed are listed under `fields`. Each entry contains a
```
dedemorton: "list of fields to be renamed are listed under `fields`" seems a little redundant.

libbeat/docs/processors-using.asciidoc
```diff
@@ -531,6 +532,52 @@ section.
...
+`from: old-key` and a `to: new-key` pair. `from` is the origin and `to` the
+target name of the field.
+
+The renaming of fields can be useful in case two fields potentially conflict.
```
dedemorton: Delete this sentence? The sentence that follows seems to say the same thing

libbeat/docs/processors-using.asciidoc
```diff
@@ -531,6 +532,52 @@ section.
...
+
+Renaming fields can be useful in cases where field names cause conflicts. For
+example if an event has two fields, `a` and `a.b`, that are both assigned a
+scalar values (e.g. `{"a": 1, "a.b": 2}`) this will result in an error at ingest
```
dedemorton: "a scalar values is wrong". should be either "that are both assigned scaler values"

libbeat/docs/processors-using.asciidoc
```diff
@@ -531,6 +532,52 @@ section.
...
+time from Elasticsearch. This is because the value of a cannot simultaneously be
```
dedemorton: wording here is a little confusing. I think you want to say: "...this will result in an Elasticsearch error at ingest time"

dedemorton: Shouldn't `a` be wrapped in back ticks here? In general, I'd avoid using "a" as a field name in an example because it's so easy to confuse it with a part of speech (article) rather than a field name.

libbeat/docs/processors-using.asciidoc
```diff
@@ -531,6 +532,52 @@ section.
...
+a scalar and an object. To prevent this rename_fields can be used to rename a to
```
dedemorton: same comment here. Use back ticks to wrap field names, especially if they can be confused with articles.

libbeat/docs/processors-using.asciidoc
```diff
@@ -531,6 +532,52 @@ section.
...
+a.value.
+
+Rename fields cannot be used to overwrite fields. To overwrite fields either
+first rename the target field or use the drop_fields processor to drop the field
```
dedemorton: wrap processor names in back ticks: use the `drop_fields` processor

libbeat/docs/processors-using.asciidoc
ruflin: changed

libbeat/docs/processors-using.asciidoc
ruflin: changed

libbeat/docs/processors-using.asciidoc
ruflin: removed

libbeat/docs/processors-using.asciidoc
ruflin: fixed

libbeat/docs/processors-using.asciidoc
ruflin: change

ruflin: added ticks and replaced by `c`.

libbeat/docs/processors-using.asciidoc
ruflin: done

ruflin:
Updated version pushed.

PR #6236, opened 2018-01-31
Add Kubernetes StatefulSet metricset

fassisrosa:
Added support for Kubernetes StatefulSet.

PR #6294, opened 2018-02-06
Kubernetes state_* metricsets are ignoring namespace #6281

### LINKED ISSUE #6281: Kubernetes state_* metricsets are ignoring namespace
Currently Kubernetes `state_*` metricsets are ignoring the namespace while retrieving resource metrics. This can be a problem if two resources from different namespaces have the same name.

We should include the namespace in the key when mapping the different values, like here: https://github.com/elastic/beats/pull/6236/files#diff-c3696d859317f46a4736d5e6d30a5f71R20

fassisrosa:
As discussed here: https://github.com/elastic/beats/pull/6236
Requested by @exekias 

Fixes #6281


exekias:
Can you please add a changelog entry?

jenkins, test it please

exekias:
jenkins, retest it please

fassisrosa:
Added changelog entry... someday I'll remember to add it upfront!!! ;) ;) ;)
Thanks!

metricbeat/module/kubernetes/state_deployment/data.go
```diff
@@ -21,11 +21,14 @@ func eventMapping(families []*dto.MetricFamily) ([]common.MapStr, error) {
 			if deployment == "" {
 				continue
 			}
-			event, ok := eventsMap[deployment]
+			namespace := util.GetLabel(metric, "namespace")
+			deploymentKey := namespace + "::" + deployment
```
ruflin: @exekias Does this mean we have `::` in the field key?

metricbeat/module/kubernetes/state_deployment/state_deployment_test.go
```diff
@@ -44,35 +44,76 @@ func TestEventMapping(t *testing.T) {
...
+	testCases := testCases()
 	for _, event := range events {
 		name, err := event.GetValue("name")
-		if err == nil && name == "jumpy-owl-redis" {
-			for k, v := range testCases {
-				testValue(t, event, k, v)
+		if err == nil {
+			namespace, err := event.GetValue("_module.namespace")
+			if err == nil {
+				eventKey := namespace.(string) + "@" + name.(string)
```
ruflin: And I assume this one means an `@` in the field key? Not that ES would disallow this, but I wonder if it could cause issues on the query side or KB?

metricbeat/module/kubernetes/state_deployment/data.go
exekias: No, this is a key in the `eventMap`, we use it to correlate all different metrics into a single event, then we ship events, forgetting about the keys in the map

metricbeat/module/kubernetes/state_deployment/state_deployment_test.go
exekias: This @ works in the same way, it's used to fetch the correct test case from here: https://github.com/elastic/beats/pull/6294/files/b514a509da8360ab225e85eb73df79468057cbd7#diff-2b3ff3319749ebeecd269e64715d1969R80

The @ is not in the final event

metricbeat/module/kubernetes/state_deployment/data.go
ruflin: got it, thanks.

PR #6207, opened 2018-01-29
Update debian and centos packers to use i386 architecture

ripcurld0:
A follow up to #6156

PR #6299, opened 2018-02-06
Fix Auditbeat yml permissions

tsg:
A side effect of the #6207 was that the `auditbeat.yml` permissions where
changed to 644, instead of the required 600. This was caught by the package tests,
but those currently break the unified build.


tsg:
I tested it locally, so it should be fine to be merged.

Issue #5547, opened 2017-11-09
suggestion on optimization


ZhiqinYang:
func (s *State) IsEqual(c *State) bool {
     // return s.ID() == c.ID()
      return (s.FileStateOS.Inode == c.FileStateOS.Inode && s.FileStateOS.Device == c.FileStateOS.Device)
  }
s.Id not init at begin lead to every time need to excute s.FileStateOS.String()

ruflin:
Can you share some more details here?  The current code checks if the id is already set. If yet it directly returns. So `String` is only called once.

```
// ID returns a unique id for the state as a string
func (s *State) ID() string {
	// Generate id on first request. This is needed as id is not set when converting back from json
	if s.Id == "" {
		s.Id = s.FileStateOS.String()
	}
	return s.Id
}

// IsEqual compares the state to an other state supporing stringer based on the unique string
func (s *State) IsEqual(c *State) bool {
	return s.ID() == c.ID()
}
```

ZhiqinYang:
@ruflin 
327 func (h *Harvester) sendEvent(data *util.Data) bool {
  328     if h.source.HasState() {
  329         h.states.Update(data.GetState())
  330     }
  331     err := h.forwarder.Send(data)
  332     return err == nil
  333 }
data.GetState() is value copy  , s.Id no init , each time would do s.Id = s.FileStateOS.String()

ruflin:
The above code would make it work only for Linux and not Windows. Did you hit any performance issues with the current code? Happy to do such a change if needed but we should add some benchmarks to it then to compare it.

ZhiqinYang:
I used it in Linux. and you ignored that in Harvester the stats in source id is "", data.GetState() is only value copy, id always is "", so each time would execute  follow:  
     s.Id = s.FileStateOS.String()

ruflin:
@ZhiqinYang As mentioned before, you are right about the execution each time. I'm open for a PR if it also takes changes on other platforms into account. It would be good to have a quick benchmark test with it to confirm the performance improvement.

ZhiqinYang:
thank you very much

andrewkroh:
Some changes were made in #6323 so this may be moot. If we want to discuss further I suggest opening a PR with a benchmark test so we can discuss the impact quantitatively.

PR #6323, opened 2018-02-08
Filebeat registrar update optimizations

urso:
- replace fmt.Sprintf in file state ID generation with append/strconv
  operations on stack allocated buffer
- change file state registry to use map[string]State instead of []State
- use time.Now() only once if registrar updates multiple states
- gc old state in registry file on flush only:
  - allow gc input states only if any events processed by registrar are eligible to be gc'ed
  - still make sure no state gc, if no events have been processed by registrar

### DIFF
--- a/filebeat/input/file/state.go
+++ b/filebeat/input/file/state.go
@@ -107,16 +107,22 @@ func (s *States) findPrevious(newState State) (State, bool) {
 
 // Cleanup cleans up the state array. All states which are older then `older` are removed
 // The number of states that were cleaned up is returned
-func (s *States) Cleanup() int {
+func (s *States) Cleanup() (int, int) {
 	s.Lock()
 	defer s.Unlock()
 
 	statesBefore := len(s.states)
+	numCanExpire := 0
 
 	currentTime := time.Now()
 	for id, state := range s.states {
+		canExpire := state.TTL >= 0
+		if canExpire {
+			numCanExpire++
+		}
+
 		expired := (state.TTL > 0 && currentTime.Sub(state.Timestamp) > state.TTL)
-		if !(state.TTL == 0 || expired) {
+		if !canExpire && expired {
 			continue
 		}
 
@@ -127,10 +133,11 @@ func (s *States) Cleanup() int {
 
 		// drop state
 		delete(s.states, id)
+		numCanExpire-- // event removed -> reduce count of pending events again
 		logp.Debug("state", "State removed for %v because of older: %v", state.Source, state.TTL)
 	}
 
-	return statesBefore - len(s.states)
+	return statesBefore - len(s.states), numCanExpire
 }
 
 // Count returns number of states
--- a/filebeat/input/file/state.go
+++ b/filebeat/input/file/state.go
@@ -122,19 +122,16 @@ func (s *States) Cleanup() (int, int) {
 		}
 
 		expired := (state.TTL > 0 && currentTime.Sub(state.Timestamp) > state.TTL)
-		if !canExpire && !expired {
-			continue
-		}
+		if state.TTL == 0 || expired {
+			if !state.Finished {
+				logp.Err("State for %s should have been dropped, but couldn't as state is not finished.", state.Source)
+			} else {
+				logp.Debug("state", "State removed for %v because of older: %v", state.Source, state.TTL)
 
-		if !state.Finished {
-			logp.Err("State for %s should have been dropped, but couldn't as state is not finished.", state.Source)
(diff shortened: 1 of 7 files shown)


filebeat/input/file/state.go
```diff
@@ -56,83 +56,88 @@ func (s *State) IsEmpty() bool {
 
 // States handles list of FileState
 type States struct {
-	states []State
+	states map[string]State
```
ruflin: @urso Does this change the output format of the registry?

urso: No. The registry used to create a copy and print the resulting array. The GetStates() method still returns a copy of `[]State`.

Alternatively I could have introduced some 'indexing' so to have `States` to be like:

```
type States struct {
  idx map[String]int  // map state ID to index into array
  states []State
}
```

ruflin: Seeing the `map[string]State` just reminded me of our first version of the registry where the string was the path and cause lots of issue. Having the unique id here is not a problem I think especially as a unique id will also be able to be created for potentially other state types.

filebeat/input/file/state.go
```diff
@@ -56,83 +56,85 @@ func (s *State) IsEmpty() bool {
...
+	// ensure ID is set, so it won't be generated in find and on insert
+	id := newState.ID()
 
-	if index >= 0 {
-		s.states[index] = newState
-	} else {
-		// No existing state found, add new one
-		s.states = append(s.states, newState)
-		logp.Debug("input", "New state added for %s", newState.Source)
+	if logp.IsDebug("input") {
```
ruflin: Should we move this into `FindPrevious` and change the log message a bit? Seems to be out of place here a bit.

filebeat/input/file/state.go
```diff
@@ -56,83 +56,85 @@ func (s *State) IsEmpty() bool {
...
+// UpdateWithTimestamp updates a state, using the passed timestamp. If the
+// previous state didn't exist, a new entry is created.
+func (s *States) UpdateWithTimestamp(newState State, ts time.Time) {
 	s.Lock()
 	defer s.Unlock()
 
-	index, _ := s.findPrevious(newState)
-	newState.Timestamp = time.Now()
+	// ensure ID is set, so it won't be generated in find and on insert
+	id := newState.ID()
```
ruflin: This reminds me of https://github.com/elastic/beats/issues/5547

filebeat/input/file/state_test.go
```diff
@@ -10,42 +10,43 @@ import (
...
 var cleanupTests = []struct {
+	title        string
 	state        State
 	countBefore  int
 	cleanupCount int
 	countAfter   int
 }{
 	{
-		// Finished and TTL set to 0
+		"Finished and TTL set to 0",
```
ruflin: nice

libbeat/common/file/file_windows.go
```diff
@@ -43,7 +44,13 @@ func (fs StateOS) IsSame(state StateOS) bool {
 }
 
 func (fs StateOS) String() string {
-	return fmt.Sprintf("%d-%d-%d", fs.IdxHi, fs.IdxLo, fs.Vol)
+	var buf [92]byte
```
ruflin: I wonder how much this improves performance compared to more readable code. I would not expect this to make a big difference in the overall performance.

urso: Quite hughe difference if registry file is big. With hughe registry file ID generation with fmt.Sprintf is one of the worst contenders in a CPU profile (at least during startup).

Follow up PR with String builder optimization in #6343

filebeat/input/file/state.go
urso: No, as findPrevious is about searching. This message is about adding a new entry.

filebeat/input/file/state.go
urso: Somewhat related. The state is passed by value, while ID is tried to be generated lazily. Pass-by-value + lazy init does results in init multiple times...

urso:
Close in favor of #6343, #6346 and #6347

filebeat/input/file/state.go
ruflin: Do you think it's also something we should address in one of the other PR's?

urso: I'd address this in another PR. Question is: why is ID generated lazily, even though it's used right away for state lookup.

ruflin: I think there is not really a reason.

PR #6325, opened 2018-02-08
Fix http parsing of repeated headers

adriansr:
Packetbeat HTTP protocol parser was not concatenating properly repeated
headers in a request or response. This caused corruption in the header
in the form of null bytes (\u0000).


tsg:
Can you backport to the 6.2 branch, please?

adriansr:
@tsg here #6330

PR #6332, opened 2018-02-09
Refactor prometheus endpoint parsing to look similar to upstream prometheus

vjsamuel:
I have changed the prometheus helper to parse in a manner similar to how it is done upstream. This would make it easier to know if a prometheus endpoint is wrong or if the issue comes from our Beats parsing logic.

Upstream sample link: https://github.com/prometheus/common/blob/master/expfmt/bench_test.go

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -223,6 +223,7 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
 - Docker and Kubernetes modules are now GA, instead of Beta. {pull}6105[6105]
 - Add pct calculated fields for Pod and container CPU and memory usages. {pull}6158[6158]
 - Add statefulset support to Kubernetes module. {pull}6236[6236]
+- Refactor prometheus endpoint parsing to look similar to upstream prometheus {pull}6332[6332]
 
 *Packetbeat*
 
--- a/metricbeat/helper/prometheus.go
+++ b/metricbeat/helper/prometheus.go
@@ -2,11 +2,12 @@ package helper
 
 import (
 	"fmt"
-
-	"github.com/elastic/beats/metricbeat/mb"
+	"io"
 
 	dto "github.com/prometheus/client_model/go"
 	"github.com/prometheus/common/expfmt"
+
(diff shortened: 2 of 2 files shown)


metricbeat/module/prometheus/collector/collector.go
```diff
@@ -65,6 +65,7 @@ func (m *MetricSet) Fetch() ([]common.MapStr, error) {
 		return nil, fmt.Errorf("Unable to decode response from prometheus endpoint")
 	}
 
+	fmt.Println(families)
```
ruflin: leftover?

ruflin:
@vjsamuel Do updating prometheus dependencies require the change or the other way around? If not I suggest we update the dependency in a separate PR first?

vjsamuel:
@ruflin the vendor update and this patch are independent. i have removed the vendor update. i will submit another PR now.

metricbeat/module/prometheus/collector/collector.go
vjsamuel: removed.

exekias:
ok to test

vjsamuel:
@exekias please do not merge. i am still figuring something out on this.

vjsamuel:
@exekias i ended up changing only a minor error handling as per how prometheus client does it. please merge once CI approves and when you are ok with the change.

CHANGELOG.asciidoc
```diff
@@ -224,6 +224,7 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
 - Docker and Kubernetes modules are now GA, instead of Beta. {pull}6105[6105]
 - Add pct calculated fields for Pod and container CPU and memory usages. {pull}6158[6158]
 - Add statefulset support to Kubernetes module. {pull}6236[6236]
+- Refactor prometheus endpoint parsing to look similar to upstream prometheus {pull}6332[6332]
```
ruflin: When I initially read this I thought the parsing of `prometheus input -> elasticsearch output` changed. But it seems mainly the error handling changed / was improved?

vjsamuel: it was always meant to change only the way parsing happens.

PR #6344, opened 2018-02-09
macOS packaging for beats

adriansr:
Added support to create signed PKG installers, bundled in a DMG file with a rudimentary uninstaller in it.

~~Currently disconnected from the packaging flow, must be run by hand after `make package`~~

Will be run by `dev-tools/deploy` given the environment variables `KEYCHAIN` and `KEYCHAN_PASS` point to a keychain file with the code-signing certificates and its password, respectively.


dev-tools/packer/platforms/darwin/pkg/templates/launchd-daemon.plist.j2
```diff
@@ -0,0 +1,24 @@
...
+	<array>
+            <string>{{.install_path}}/{{.beat_vendor}}/{{.beat_name}}/bin/{{.beat_name}}</string>
+            <string>-c</string>
+            <string>/etc/{{.beat_name}}/{{.beat_name}}.yml</string>
+            <string>--path.home</string>
+            <string>{{.install_path}}/{{.beat_vendor}}/{{.beat_name}}</string>
+            <string>--path.config</string>
+            <string>/etc/{{.beat_name}}</string>
+            <string>--path.data</string>
+            <string>{{.install_path}}/{{.beat_vendor}}/{{.beat_name}}</string>
```
andrewkroh: Would it be OK to set path.data to `/var/lib/{{.beat_name}}` similar to Linux?

https://github.com/elastic/beats/blob/545938e899c9f085a9b3b7d1d8dd119c139c7a90/dev-tools/packer/platforms/centos/beatname.sh.j2#L9

dev-tools/packer/platforms/darwin/pkg/internal_build.sh
```diff
@@ -0,0 +1,67 @@
...
+    --timestamp \
+    "$PKG_NAME"
+
+cp "$PKG_NAME" dmg/
+hdiutil create \
+    -volname "$BEAT $VERSION" \
+    -srcfolder dmg \
+    -ov \
+    "$DMG_NAME"
+
```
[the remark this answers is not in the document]
adriansr: Sure. I didn't know signing DMGs was a thing.

dev-tools/packer/platforms/darwin/pkg/templates/launchd-daemon.plist.j2
adriansr: Done

adriansr:
Missing tasks:

- [ ] Add KEYCHAIN* parameters to jenkins jobs.
- [ ] do not add icon to DMG. It only works locally.
- [ ] Add icon to uninstaller (artwork pending).
- [ ] Include uninstaller along the beat binaries.
- [ ] preference pane.

andrewkroh:
This looks really good!

> Include uninstaller along the beat binaries.

What do you think about creating an `/Applications/Elastic` directory that contains an "Uninstall Filebeat" link to the uninstall app?

Is there anything I can help with?

I'm thinking the next step should be to get this building as part of https://beats-ci.elastic.co/job/elastic+beats+master+package/. We can get the key information stored into Vault then load it from there. This should make us a bit more prepared to integrate with the Release Manager.

adriansr:
> What do you think about creating an /Applications/Elastic directory that contains an "Uninstall Filebeat" link to the uninstall app?

My plan was to add an Uninstall button to each beat in the Preference Pane.

> I'm thinking the next step should be to get this building as part of https://beats-ci.elastic.co/job/elastic+beats+master+package/. We can get the key information stored into Vault then load it from there. This should make us a bit more prepared to integrate with the Release Manager.

Yes, that is already in progress. The certificates are deployed to the macOS images and the password is in Vault, now I'm waiting for infra to make the password and certificate path available to the build job via env vars (I added a link to the issue above).

adriansr:
jenkins, test it

adriansr:
jenkins, package it

dev-tools/packer/platforms/darwin/build.sh
```diff
@@ -17,3 +17,6 @@ docker run --rm -v ${BUILD_DIR}:/build \
     tudorg/fpm /build/run-$runid.sh
 
 rm ${BUILD_DIR}/settings-$runid.yml ${BUILD_DIR}/run-$runid.sh
+
+echo 'Running macOS packager'
+${BASEDIR}/pkg/build.sh
```
andrewkroh: I think there should be a conditional preventing this from error'ing out if we aren't on Darwin.

Or it might be better to separate this into it's own build target such that you can run `make packages` on any machine with Docker and can run `make macos-dmgs` (naming TBD) afterwards to build the the DMGs.

adriansr:
jenkins, test it

dev-tools/osx-deploy
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env python
+import os
+import argparse
+from subprocess import check_call
+
+def main():
+    parser = argparse.ArgumentParser(
+        description="Builds OSX packages for Beats")
+    args = parser.parse_args()
```
andrewkroh: Probably this should have the same option as `dev-tools/deploy`?

https://github.com/elastic/beats/blob/96b5aeeaa776dcfe2252209c3633d6339d94592e/dev-tools/deploy#L9-L10

andrewkroh:
screen shot 2018-05-31 at 12 50 31 am

adriansr:
Please do not merge until further notice

adriansr:
Ready to merge

andrewkroh:
@adriansr With KEYCHAIN (and KEYCHAIN_PASS) not being used, can they be removed from the Jenkins job configuration?

It looks like Ansible installs the keychain-db file to Jenkins, but does not add it to the keychain search list or unlock it. Was that done manually or am I missing something?

It looks like Ansible could be made do the following and this would put the worker in a ready state. WDYT?

`security list-keychains -s <keychain>`
`security -v unlock-keychain <keychain> < vault read -field=password`

adriansr:
@andrewkroh 

The problem that prompted me to get rid of KEYCHAIN is that, in macOS 10.12 and later, there is no reliable way of handling keychains from launchd services, due to issues with session isolation.

The only reliable way to perform the code-signing from Jenkins is to copy the certificates into the System keychain. So we will need Ansible to do something like this https://apple.stackexchange.com/questions/80623/import-certificates-into-the-system-keychain-via-the-command-line

andrewkroh:
Added via https://github.com/elastic/beats/pull/7481.

PR #6345, opened 2018-02-09
[WIP] Update generator

graphaelli:
version issue has been lingering since #5973.

Marked WIP as seemingly unrelated tests failed locally, would like to see what CI says.

### DIFF
--- a/generator/beat/{beat}/beater/{beat}.go.tmpl
+++ b/generator/beat/{beat}/beater/{beat}.go.tmpl
@@ -18,15 +18,15 @@ type {Beat} struct {
 }
 
 // Creates beater
-func New(b *beat.Beat, cfg *common.Config) (beat.Beater, error) {
-	config := config.DefaultConfig
-	if err := cfg.Unpack(&config); err != nil {
+func New(b *beat.Beat, c *common.Config) (beat.Beater, error) {
+	cfg := config.DefaultConfig
+	if err := c.Unpack(&cfg); err != nil {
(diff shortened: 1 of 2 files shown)


generator/beat/{beat}/beater/{beat}.go.tmpl
```diff
@@ -18,15 +18,15 @@ type {Beat} struct {
 }
 
 // Creates beater
-func New(b *beat.Beat, cfg *common.Config) (beat.Beater, error) {
-	config := config.DefaultConfig
-	if err := cfg.Unpack(&config); err != nil {
+func New(b *beat.Beat, c *common.Config) (beat.Beater, error) {
```
ruflin: What is the reason for the renaming here?

ruflin:
@graphaelli Could you update the PR title and description a bit to make it more self explanatory what the change is for?

generator/beat/{beat}/beater/{beat}.go.tmpl
graphaelli: this stops shadowing the `config` import

ruflin: I see. Seems like we have that also in our beats.

The reason the renaming is a bit unfortunate is because in most of our code base we use `cfg` to represent go-ucfg / common.Config type. WDYT think about using `c := config.DefaultConfig` instead and keeping `cfg` as it is?

graphaelli: Sounds sane, I'll make that update

PR #6346, opened 2018-02-09
Filebeat in memory states registry improvements

urso:
- make States type easier to find by moving into separate states.go file
- Ensure provides States constructor is actually used
- Add ID->array index, index for faster lookups on update and find
  operations. When updating states in a big registry, the Updates
  converged to quadratic complexity. The index helps in keeping the
  complexity about linear in number of state updates.
- Debug will print number of states subject to future cleanups (if state TTL > 0)
- Add title to states unit tests

### DIFF
--- a/filebeat/input/file/states.go
+++ b/filebeat/input/file/states.go
@@ -64,7 +64,7 @@ func (s *States) FindPrevious(newState State) State {
 	return s.states[i]
 }
 
-// findPrevious returns the previous state fo the file
+// findPrevious returns the previous state for the file.
 // In case no previous state exists, index -1 is returned
 func (s *States) findPrevious(id string) int {
 	if i, exists := s.idx[id]; exists {
@@ -76,7 +76,6 @@ func (s *States) findPrevious(id string) int {
 // Cleanup cleans up the state array. All states which are older then `older` are removed
 // The number of states that were cleaned up and number of states that can be
 // cleaned up in the future is returned.
-
 func (s *States) Cleanup() (int, int) {
 	s.Lock()
 	defer s.Unlock()
--- a/filebeat/input/file/states.go
+++ b/filebeat/input/file/states.go
@@ -76,36 +76,41 @@ func (s *States) findPrevious(id string) int {
 // Cleanup cleans up the state array. All states which are older then `older` are removed
 // The number of states that were cleaned up and number of states that can be
 // cleaned up in the future is returned.
+
 func (s *States) Cleanup() (int, int) {
 	s.Lock()
 	defer s.Unlock()
 
+	currentTime := time.Now()
 	statesBefore := len(s.states)
 	numCanExpire := 0
 
-	currentTime := time.Now()
-	states := s.states[:0]
-
-	for i := range s.states {
+	for i := 0; i < len(s.states); i++ {
 		state := &s.states[i]
 		expired := (state.TTL > 0 && currentTime.Sub(state.Timestamp) > state.TTL)
 
 		if state.TTL == 0 || expired {
-			if state.Finished {
-				logp.Debug("state", "State removed for %v because of older: %v", state.Source, state.TTL)
-				delete(s.idx, state.ID())
-				continue // drop state
-			} else {
+			if !state.Finished {
 				logp.Err("State for %s should have been dropped, but couldn't as state is not finished.", state.Source)
+				continue
 			}
-		}
 
-		if state.TTL >= 0 {
+			// remove state from index
+			delete(s.idx, state.ID())
+
+			// remove entry by relocating last entry + shrink states array
+			last := len(s.states) - 1
+			s.states[i] = s.states[last]
+			s.states = s.states[:last]
+
+			// update index if replacement state still in registry
+			if i < len(s.states) {
+				s.idx[s.states[i].ID()] = i
+			}
+		} else if state.TTL > 0 {
 			numCanExpire++
 		}
-		states = append(states, *state) // in-place copy old state
 	}
-	s.states = states
 
 	return statesBefore - len(s.states), numCanExpire
 }
--- a/filebeat/input/file/states.go
+++ b/filebeat/input/file/states.go
@@ -1,0 +1,148 @@
+package file
+
+import (
+	"sync"
+	"time"
+
+	"github.com/elastic/beats/libbeat/logp"
+)
+
+// States handles list of FileState. One must use NewStates to instantiate a
+// file states regisry. Using the zero-value is not safe.
+type States struct {
+	sync.RWMutex
+
+	// states store
+	states []State
+
+	// idx maps state IDs to state indexes for fast lookup and modifications.
+	idx map[string]int
+}
+
+// NewStates generates a new states registry.
+func NewStates() *States {
+	return &States{
+		states: nil,
+		idx:    map[string]int{},
+	}
+}
+
+// Update updates a state. If previous state didn't exist, new one is created
+func (s *States) Update(newState State) {
+	s.UpdateWithTs(newState, time.Now())
+}
+
+// UpdateWithTs updates a state, assigning the given timestamp.
+// If previous state didn't exist, new one is created
+func (s *States) UpdateWithTs(newState State, ts time.Time) {
+	s.Lock()
+	defer s.Unlock()
+
+	id := newState.ID()
+	index := s.findPrevious(id)
+	newState.Timestamp = ts
+
+	if index >= 0 {
+		s.states[index] = newState
+	} else {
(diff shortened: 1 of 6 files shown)


urso:
In-memory state registry optimizations only (original ticket: #6323).

ph:
One test is still failing for filebeat? `ERROR: Checks that states are properly removed after clean_inactive`

filebeat/input/file/states.go
```diff
@@ -0,0 +1,155 @@
...
+	s.RLock()
+	defer s.RUnlock()
+	i := s.findPrevious(newState.ID())
+	if i < 0 {
+		return State{}
+	}
+	return s.states[i]
+}
+
+// findPrevious returns the previous state fo the file
```
ph: s/fo/for

filebeat/input/file/states.go
```diff
@@ -0,0 +1,155 @@
...
+	if i, exists := s.idx[id]; exists {
+		return i
+	}
+	return -1
+}
+
+// Cleanup cleans up the state array. All states which are older then `older` are removed
+// The number of states that were cleaned up and number of states that can be
+// cleaned up in the future is returned.
+
```
ph: Space?

filebeat/input/file/states.go
```diff
@@ -0,0 +1,155 @@
...
+				continue
+			}
+
+			// remove state from index
+			delete(s.idx, state.ID())
+
+			// remove entry by relocating last entry + shrink states array
+			last := len(s.states) - 1
+			s.states[i] = s.states[last]
+			s.states = s.states[:last]
```
ph: thinking out loud here, related to the case of a lot of files.
Could this generate a lot of eviction/resizing in the real world? I guess we are just moving pointers.

urso: You mean CPU cache? No idea without testing/profiling. But we basically narrow down the array from beginning and end at the same time.

urso: Also depends on how many entries need to be removed.

filebeat/input/file/states.go
```diff
@@ -0,0 +1,157 @@
...
+// States handles list of FileState. One must use NewStates to instantiate a
+// file states regisry. Using the zero-value is not safe.
+type States struct {
+	sync.RWMutex
+
+	// states store
+	states []State
+
+	// idx maps state IDs to state indexes for fast lookup and modifications.
+	idx map[string]int
```
ruflin: What is the reason you changed your mind to the separate map string for keeping track of the ids?

filebeat/input/file/states.go
```diff
@@ -0,0 +1,157 @@
...
+// cleaned up in the future is returned.
+func (s *States) Cleanup() (int, int) {
+	s.Lock()
+	defer s.Unlock()
+
+	currentTime := time.Now()
+	statesBefore := len(s.states)
+	numCanExpire := 0
+
+	L := len(s.states)
```
ruflin: upper case on purpose?

filebeat/registrar/registrar.go
```diff
@@ -185,14 +185,14 @@ func (r *Registrar) onEvents(states []file.State) {
...
-	cleanedStates := r.states.Cleanup()
+	cleanedStates, pendingClean := r.states.Cleanup()
 	statesCleanup.Add(int64(cleanedStates))
 
 	r.bufferedStateUpdates += len(states)
 
 	logp.Debug("registrar",
-		"Registrar states cleaned up. Before: %d, After: %d",
-		beforeCount, beforeCount-cleanedStates)
+		"Registrar states cleaned up. Before: %d, After: %d, Pending: %d",
```
ruflin: I already see questions coming on what pending cleanup means :-)

filebeat/input/file/states.go
```diff
@@ -0,0 +1,157 @@
...
+			logp.Debug("state", "State removed for %v because of older: %v", state.Source, state.TTL)
+
+			L--
+			if L != i {
+				s.states[i] = s.states[L]
+				s.idx[s.states[i].ID()] = i
+			}
+		} else {
+			i++
+			if canExpire {
```
ruflin: What is the reason you added `canExpire`? If there is `clean_inactive` configured it applies to all states.

filebeat/input/file/states.go
```diff
@@ -0,0 +1,157 @@
...
+			if !state.Finished {
+				logp.Err("State for %s should have been dropped, but couldn't as state is not finished.", state.Source)
+				i++
+				continue
+			}
+
+			delete(s.idx, state.ID())
+			logp.Debug("state", "State removed for %v because of older: %v", state.Source, state.TTL)
+
+			L--
```
ruflin: Looking at the code with `i++` and `L--` I'm thinking if the option with combining index and registry would actually bring less complexity?

filebeat/input/file/states.go
urso: Too many copies :)

The states mapping has these very frequent access patterns:
- access single entry by ID
- loop/copy all states

Performance wise, both access patterns are equally important, as they are run very very often (basically per event).

Optimizing the access by ID case (using map[string]state) I found the profiling being dominated by extra copies, even if only one field is required. Introducing an index on array indexes, we still have the well working 'all-access loop' + we optimize single entry access to be about O(1) complexity.

Kind of best of both worlds with a little constant overhead when actually removing entries from the registry. As state update/add + iteration + copy are the most often used operations, the added overhead on removal is pretty much worth it.

filebeat/input/file/states.go
urso: yep. L for "length" :)

filebeat/registrar/registrar.go
urso: We can call it 'active'. The number of 'pending' states is all states with TTL > 0.

Following the state changes and TTLs, I found TTL basically has 4 meanings:
- `-2`: zombie state (no prospector/input processing the file)
- `-1`: alive (prospector/input owning the file, but no TTL)
- `0`: done (entry is marked as removable)
- `>0`: subject to timeout

Any better naming?

Currently The 'pending' count is the number of states potentially subject to being timed out. But states `-1` and `>0` indicate the state to be 'active'.

filebeat/input/file/states.go
urso: See my `TTL` comment. The `canExpire` marks states with `TTL > 0` -> states subject to timeout. If Cleanup returns and number of events with `TTL>0` is 0, no cleanup needs to be run until a new state with `TTL>=0` is added.

I also make use of this in my followup PR (#6347) by deferring/suspending states Cleanup in the registrar.

filebeat/input/file/states.go
urso: Maybe it's not that clear as some extra work is done within the loop, but the loop basically sorts the list into alive and dead states. The L is the current length. Removal/sorting is in-place. That is, we're forced to check the current index again upon removal.
The `i++` and `L--` act as delimiters of processed and alive/dead states => both need to be updated, depending if state is alive or not. Very basic in-place removal loop without extra copies.

would this be more clear for you:

```
func (s *States) Cleanup() (int, int) {
	s.Lock()
	defer s.Unlock()

	L := len(s.states)
	currentTime := time.Now()
	statesBefore := L
	numCanExpire := 0 // number of alive states with TTL > 0

	for i := 0; i < L; {
		state := &s.states[i]
		canExpire := state.TTL > 0
		expired := (canExpire && currentTime.Sub(state.Timestamp) > state.TTL)

		if (state.TTL == 0 || expired) && state.Finished {
			// remove by replacing entry at i with last entry and shrink 'alive' states length
			L--
			if L != i {
				// swap and update index
				s.states[i], s.states[L] = s.states[L], s.states[i]
				s.idx[s.states[i].ID()] = i
			}
		} else {
			i++
			if canExpire {
				numCanExpire++
			}
		}
	}

	alive := s.states[:L]
	dead := s.states[L:]
	for i := range dead {
		delete(s.idx, dead[i].ID())
	}

	s.states = alive
	return statesBefore - len(s.states), numCanExpire
}
```

urso:
Failing ci test is on auditbeat. Unrelated to this PR.

filebeat/input/file/states.go
ph: When I've looked at the code initially I had to use pseudo code here to make sure the in place was working correctly. 

I prefer the first version.

ph:
@urso for me I think changes are well worth it, lets have the green light from @ruflin too.

ruflin:
@ph Please go forward and merge if Jenkins failure is not related. The most important part is the functionality and there we agree. We can still do minor refactoring later.

PR #6347, opened 2018-02-09
Defer registry state cleanup

urso:
- Defer registrar state gc until the registry needs to be written.
- when updating regitrar states from batch, ensure all updated states will
have same timestamp.

Depends on #6346

### DIFF
--- a/filebeat/registrar/registrar.go
+++ b/filebeat/registrar/registrar.go
@@ -23,6 +23,8 @@ type Registrar struct {
 	wg           sync.WaitGroup
 
 	states               *file.States // Map with all file paths inside and the corresponding state
+	gcRequired           bool         // gcRequired is set if registry state needs to be gc'ed before the next write
+	gcEnabled            bool         // gcEnabled indictes the registry contains some state that can be gc'ed in the future
 	flushTimeout         time.Duration
 	bufferedStateUpdates int
 }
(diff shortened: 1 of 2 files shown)


filebeat/registrar/registrar.go
```diff
@@ -183,24 +185,51 @@ func (r *Registrar) Run() {
...
-	r.bufferedStateUpdates += len(states)
-
 	logp.Debug("registrar",
-		"Registrar states cleaned up. Before: %d, After: %d",
-		beforeCount, beforeCount-cleanedStates)
+		"Registrar states cleaned up. Before: %d, After: %d, Pending: %d",
+		beforeCount, beforeCount-cleanedStates, pendingClean)
+
+	r.gcRequired = false
+	r.gcEnabled = pendingClean > 0
```
ruflin: @urso Looking at this PR I now better understand why you introduced `pendingClean`. I think the confusing part for me here is the name of the variable. pending indicates for me it's already in a queue or similar for cleanup but has not finished yet. But what we have here is states which have an expiry time stamp but are not expired yet. Unexpired is the word that comes to my mind but is not nice.

It would be nice to know which states were removed because of `clean_inactive` and which ones because of `clean_removed`. So fare we don't make a difference. It would me we would log 4 stats:

* Total states after cleanup without expiry 
* Total states after cleanup with expiry 
* States removed because of expiry
* States removed because of removal

Based on this the number of previous states could be calculated.

ph: I think theses addition makes sense, it's mostly for us for debugging registry problems.

urso: Let's do the stats reporting in another PR. Here I'd like to focus on deferring the cleanup

urso:
rebased PR on master

filebeat/registrar/registrar.go
```diff
@@ -183,24 +185,51 @@ func (r *Registrar) Run() {
...
+	}
+
+	logp.Debug("registrar", "Registrar state updates processed. Count: %v", len(states))
+
+	// new set of events received -> mark state registry ready for next
+	// cleanup phase in case gc'able events are stored in the registry.
+	r.gcRequired = r.gcEnabled
+}
+
+// gcStates runs a registry Cleanup. The bool returned indicates wether more
```
ruflin: There is no bool returned here.

PR #6348, opened 2018-02-11
[Filebeat] create-module command

ewgRa:
This is followup PR from #6157, that introduce new command: create-module.

To this command was extracted functionality from create-fileset command, related to creation module.

create-fileset command now just check that module exists, if not - just exit.


kvch:
jenkins test it

PR #6360, opened 2018-02-12
set compile time version options in package builds

graphaelli:
since #5973 `make package` produces binaries that report unknown values for the `version` subcommand, like:

```
$ ./examplebeat version
examplebeat version 7.0.0-alpha1 (amd64), libbeat 7.0.0-alpha1 [unknown built unknown]
```

This pull request sets the compile time values for both official beats (those in this repo) and those that vendor beats, like apm-server.

```
examplebeat version 7.0.0-alpha1 (amd64), libbeat 7.0.0-alpha1 [e92d2d9fa4d38bee23eab80e3b8262f045825eb9 built 2018-02-12 19:28:25 +0000 UTC]
```

Two notes:
* symbol table and debug stripping is retained for darwin.
* static builds on linux seem to work (manually tested) even though it doesn't seem to be used currently

### DIFF
--- a/dev-tools/packer/docker/xgo-image/base/build.sh
+++ b/dev-tools/packer/docker/xgo-image/base/build.sh
@@ -105,7 +105,14 @@ fi
 
 if [ "$FLAG_V" == "true" ]; then V=-v; fi
 if [ "$FLAG_RACE" == "true" ]; then R=-race; fi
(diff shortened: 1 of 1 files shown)


graphaelli:
fixes #6051

tsg:
> static builds on linux seem to work (manually tested) even though it doesn't seem to be used currently

Yes, we tried them in the past but they caused issues at runtime. Since you already went through the trouble of keeping support for static, I think it's good to have it there in case we want to experiment with it in the future.

PR #6379, opened 2018-02-13
Update filebeat kubernetes autodiscover documentation

tauffredou:
# What I did
I've updated the documentation to use **kubernetes.container.id** instead of **docker.container.id**.

# Why
The current documentation may work but its still confusing to rely on docker metadata when using kubernetes provider.

In addition, using the docker metada adds an unnecessary dependency to the docker metadata generator if we don't plan to use any specific docker attributes.

By removing this reference to docker, the add_docker_metadata processor and the access to the docker daemon are no longer needed.

### DIFF
--- a/filebeat/docs/autodiscover-kubernetes-config.asciidoc
+++ b/filebeat/docs/autodiscover-kubernetes-config.asciidoc
@@ -12,7 +12,7 @@ filebeat.autodiscover:
           config:
             - type: docker
               containers.ids:
-                - "${data.docker.container.id}"
+                - "${data.kubernetes.container.id}"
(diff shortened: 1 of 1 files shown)


kvch:
Thanks for fixing the docs!
Could you please sign the CLA? https://www.elastic.co/contributor-agreement

tauffredou:
[image: Images intégrées 1][image: Images intégrées 1]Hi,

As you can see, this has already been done. Can I do something else?

Thomas

2018-02-15 11:36 GMT+01:00 Noémi Ványi <notifications@github.com>:

kvch:
No, now it seems ok. Thanks!

PR #6385, opened 2018-02-14
Update running-on-docker.asciidoc

CodingSpiderFox:
-e option was missing

See https://discuss.elastic.co/t/running-metricbeat-in-docker-and-monitoring-the-host-system-does-not-work-with-command-from-documentation/119821

### DIFF
--- a/metricbeat/docs/running-on-docker.asciidoc
+++ b/metricbeat/docs/running-on-docker.asciidoc
@@ -19,7 +19,7 @@ docker run \
   --volume=/sys/fs/cgroup:/hostfs/sys/fs/cgroup:ro \ <2>
   --volume=/:/hostfs:ro \ <3>
   --net=host <4>
-  {dockerimage} -system.hostfs=/hostfs
+  {dockerimage} -e -system.hostfs=/hostfs
 ----
 
 <1> Metricbeat's <<metricbeat-module-system,system module>> collects much of its data through the Linux proc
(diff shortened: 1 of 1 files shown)


CodingSpiderFox:
Signed the CLA

andrewkroh:
@dedemorton - A question about the asciidoc callouts and trailing spaces -- If you copy and paste the `docker run` command on [this page](https://www.elastic.co/guide/en/beats/metricbeat/6.1/running-on-docker.html#monitoring-host) that uses callouts you end up with trailing spaces on lines with a callout.

Do you know if there is anything we can change to make the trailing spaces go away? The trailing spaces break the ability to directly paste the command to the CLI.

dedemorton:
@andrewkroh AFAIK, the extra whitespace is a side effect of having a callout. I'm gong to follow up with our docs infra person because I think it would be great to have some kind of Copy Command functionality in the docs. Thanks for bringing this up.

metricbeat/docs/running-on-docker.asciidoc
```diff
@@ -19,7 +19,7 @@ docker run \
   --volume=/sys/fs/cgroup:/hostfs/sys/fs/cgroup:ro \ <2>
   --volume=/:/hostfs:ro \ <3>
   --net=host <4>
```
dedemorton: This line needs a trailing backslash for line continuation. Should be: `--net=host \ <4>`

Issue #6389, opened 2018-02-14
Beat uninstall-service-[beatname].ps1 do not stop the service before remove


palethorper:
The uninstall-service-[beatname].ps1 script do not have a section to stop the service before removing it.
The install script do already include the log to atop the service before installing new version.

- Operationing System : Windows
- Version: up to 6.2.1

- Steps to reproduce:
1. Have a beat (filebeat, winlogbeat, metricbeat etc) service installed and running on windows
2. Run the uninstall script that is included in the package. e.g. uninstall-service-filebeat.ps1

The uninstall will not succeed.

Fix:
Include the stop service log in the uninstall script
`# delete service if it already exists
if (Get-Service filebeat -ErrorAction SilentlyContinue) {
  $service = Get-WmiObject -Class Win32_Service -Filter "name='filebeat'"
  $service.StopService()
  Start-Sleep -s 1
  $service.delete()
}`

andrewkroh:
Could you please open a pull request for this change?

The file that needs the modification is at https://github.com/elastic/beats/blob/master/dev-tools/packer/platforms/windows/uninstall-service.ps1.j2. It is a template that is shared by all Beats so any time you need to reference the name use the variable `{{.beat_name}}`.

palethorper:
Yep - just forked it and doing it now.

michha:
As I look into https://github.com/elastic/beats/blob/master/dev-tools/packaging/templates/windows/uninstall-service.ps1.tmpl it has the StopService call implemented, but filebeat-6.3.2 x64 (24th July) is still missing the StopService in the uninstall PowerShell script

michha:
Version 6.4 released yesterday fixed this 👍 
This issue can be resolved as fixed in v6.4

PR #6390, opened 2018-02-14
Stop beat service before remove when uninstalling

palethorper:
Fix for issue #6389

### DIFF
--- a/dev-tools/packer/platforms/windows/uninstall-service.ps1.j2
+++ b/dev-tools/packer/platforms/windows/uninstall-service.ps1.j2
@@ -1,4 +1,4 @@
-# delete service if it already exists
+# delete service if it exists
 if (Get-Service {{.beat_name}} -ErrorAction SilentlyContinue) {
   $service = Get-WmiObject -Class Win32_Service -Filter "name='{{.beat_name}}'"
   $service.StopService()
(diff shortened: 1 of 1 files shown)


tsg:
jenkins, test it

dev-tools/packer/platforms/windows/uninstall-service.ps1.j2
```diff
@@ -1,5 +1,7 @@
 # delete service if it exists
 if (Get-Service {{.beat_name}} -ErrorAction SilentlyContinue) {
   $service = Get-WmiObject -Class Win32_Service -Filter "name='{{.beat_name}}'"
+  $service.StopService()
+  Start-Sleep -s 1
```
tsg: I assume StopService happens asynchronously and that's why we need the sleep, right? One second might not be enough for the Beat to stop, so ideally we wouldn't rely on a sleep here, but if there's no way of waiting for StopService to be done, I'm fine with this solution and the choses 1s interval.

palethorper: Will test it.
The StopService call itself should wait for the executable to stop.
The 1 second is for - on windows - it can take slightly longer for the executable to actually terminate.

Note - the call in install-service scripts is exactly the same - if there is an issue here 0 it is more likely to be uncovered there than as part of the uninstall script ...

palethorper: Question: Is there somewhere that I can get the built artifacts (ie. the beats packages) from? Or do I need to build it all locally?

urso:
Looks like the uninstaller already has some similar change. Closing.

PR #6398, opened 2018-02-16
Fix fileset generator config.yml naming

kvch:
Previously, the generator created a config file named `config.yml`. This returned the following error in case of cat/bat fileset:
```
Exiting: Error getting config for fielset cat/bat: Error reading input file config/bat.yml: open /home/n/go/src/github.com/elastic/beats/filebeat/module/cat/bat/config/bat.yml: no such file or directory
```
From now on the config file is generated as `{{filesetname}}.yml`.

I also added a missing error check to the script.


ruflin:
jenkins, test it

PR #6404, opened 2018-02-18
Log reconnect attempts

### LINKED ISSUE #5175: Logging:  Log attempts from Beats to re-establish connections
Currently, at the default logging level, we are not logging entries on when Logstash attempts to re-establish connections (eg. due to DNS for downstream servers not available).  For example, after these messages when the DNS issue is resolved:

```
2017-09-07T02:50:43-07:00 WARN DNS lookup failure "<hostname>": lookup <hostname> on <ip>: read udp <ip>:51430-><ip>: i/o timeout
2017-09-07T02:50:43-07:00 ERR Connect failed with: lookup <hostname> on <ip>: read udp <ip>:51430-><ip>: i/o timeout
```

Having this information logger at the default level can help with troubleshooting.

agathver:
Closes #5175 

Logs an info line "Attempting to reconnect" at every reconnect attempt.

### DIFF
--- a/libbeat/publisher/pipeline/output.go
+++ b/libbeat/publisher/pipeline/output.go
@@ -77,21 +77,30 @@ func (w *netClientWorker) Close() error {
 
 func (w *netClientWorker) run() {
 	for !w.closed.Load() {
+		reconnectAttempts := 0
+
 		// start initial connect loop from first batch, but return
 		// batch to pipeline for other outputs to catch up while we're trying to connect
 		for batch := range w.qu {
 			batch.Cancelled()
 
 			if w.closed.Load() {
+				logp.Info("Closed connection")
 				return
+			}
+
+			if reconnectAttempts > 0 {
+				logp.Info("Attempting to reconnect with %d reconnect attempt(s)", reconnectAttempts)
 			}
 
 			err := w.client.Connect()
 			if err != nil {
 				logp.Err("Failed to connect: %v", err)
+				reconnectAttempts++
 				continue
 			}
 
+			reconnectAttempts = 0
 			break
 		}
 
--- a/libbeat/publisher/pipeline/output.go
+++ b/libbeat/publisher/pipeline/output.go
@@ -85,21 +85,24 @@ func (w *netClientWorker) run() {
 			batch.Cancelled()
 
 			if w.closed.Load() {
-				logp.Info("Closed connection")
+				logp.Info("Closed connection to %v", w.client)
 				return
 			}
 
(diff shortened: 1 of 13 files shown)


kvch:
jenkins test it

libbeat/publisher/pipeline/output.go
```diff
@@ -69,12 +71,18 @@ func (w *netClientWorker) run() {
 				return
 			}
 
+			if attemptReconnect {
+				logp.Info("Attempting to reconnect.")
+			}
```
urso: +1 for logging connection and reconnect attempts

We have two loops in this methods. First loop attempts to connect. Second loop processes events.

The first loop becomes active after beats startup, once the first event is to be published. That is connecting the outputs is done lazily. In this case the log message should indicate this being the initial connect.

Once the send loop fails, we're about to reconnect.

Having a counter on failed connect attempts would be nice to have as well.

E.g. for the lifetime of an output it would be nice to have log messages like (assuming we would have an 'identifier' per output): 

```
Startup output to <identifier>
Connect to <identifier>
Failed to connect to <identifier>: <reason>
Connect to <identifier>
Connection to <identifier> established
Failed to publish events to <identifier>: %v
Closed connection to <identifier>
Attempting to reconnect to <identifier> with 0 reconnect attempt(s).
Failed to connect to <identifier>: <reason>
Attempting to reconnect to <identifier> with 1 reconnect attempt(s).
Connection to <identifier> established
Shutdown output to <identifier>
```

That is, we have 4 phases/states to distinguish:
1. init worker
2. create initial connection
3. reconnect
4. shutdown

Having an enum to differentiate on state and logging should help here.

agathver: I wanted to use hostname as an identifier. However, we don't have it here.
I'll be adding additional logging lines for other events.

agathver:
@urso We counts for reconnect attempts now.

agathver:
@urso I need some pointers for how to create the identifier. 

The state enum should probably go into the `Connection` type ?

ruflin:
@agathver The team is at Elastic{ON} / All hands this and next week so it will take some time to respond.

agathver:
Any updates on this?

agathver:
@urso, Added identifiers for connections.

I have used a composition-like approach for naming (borrowed from the convention in ReactJS), in tune with how most clients are being used.

Example, a redis client wrapped with a failover will be `failover(redis://10.0.0.1, redis://10.0.0.2)`
Default ES config gives `backoff(elasticsearch(http://localhost:9200))`

PR #6406, opened 2018-02-19
Rework MapStr.walkMap to support dotted keys

ruflin:
Currently if a dotted key is passed to walkmap it is split up in it's part and the map is traversed for each part. In case an event has a dotted key inside walkMap does not return the expected value. For the Event `{"a.b": 1}` the method `GetValue("a.b")` has previously returned `ErrKeyNotExists` but now returns 0.

With this change all processors which rely on Get/Put functionalities of MapStr can now also work with events containing dotted keys.

**Changes**

* The walkMap function was rewritten to a recursive function. This made adding the new functionality easier.
* Add tests for GetValue to make possible usages more clear
* Add Benchmark tests for walkMap use cases

**Limitations**

When using `walkMap` with a Key, either the full key has to be dotted or the last part. `GetValue("a.b.c")` on an event `{"a.b":{"c":17}}` does not work as only the first part of the key or the full key are checked against the map.

Benchmark old to new walkMap:

```
go test  -bench=BenchmarkWalkMap -run=BenchmarkWalkMap
```

Recursive:
```
BenchmarkWalkMap/Get-4           1000000              1274 ns/op
BenchmarkWalkMap/Put-4           1000000              1277 ns/op
BenchmarkWalkMap/HasKey-4        2000000               880 ns/op
BenchmarkWalkMap/HasKeyFirst-4  20000000               111 ns/op
BenchmarkWalkMap/Delete-4        1000000              1313 ns/op
```

Iterative:
```
BenchmarkWalkMap/Get-4            500000              2608 ns/op
BenchmarkWalkMap/Put-4           1000000              1140 ns/op
BenchmarkWalkMap/HasKey-4        3000000               411 ns/op
BenchmarkWalkMap/HasKeyFirst-4  20000000                85.9 ns/op
BenchmarkWalkMap/Delete-4        1000000              1008 ns/op
```

The benchmark show a difference in the code before and after. The `GetValue` function has become quite a bit more efficient. The `HasKey` is now less efficient. I still need to investigate in detail where the differences come from but the assumption is it comes from the additional functionality.

The old function was left in the code for comparison reason in this PR. It can be removed later.

This change was initially triggered by https://github.com/elastic/beats/pull/5916 but is split out into a separate PR because it has usage in multiple places.

### DIFF
--- a/libbeat/common/mapstr.go
+++ b/libbeat/common/mapstr.go
@@ -315,40 +315,53 @@ func tryToMapStr(v interface{}) (MapStr, bool) {
 	}
 }
 
-// walkMap walks the data MapStr to arrive at the value specified by the key.
+// walkMapRecursive walks the data MapStr to arrive at the value specified by the key.
 // The key is expressed in dot-notation (eg. x.y.z). When the key is found then
 // the given mapStrOperation is invoked.
-func walkMap(key string, data MapStr, op mapStrOperation) (interface{}, error) {
-	var err error
-	keyParts := strings.Split(key, ".")
+func walkMapRecursive(key string, data MapStr, op mapStrOperation) (interface{}, error) {
 
-	// Walk maps until reaching a leaf object.
-	m := data
-	for i, k := range keyParts[0 : len(keyParts)-1] {
-		v, exists := m[k]
-		if !exists {
-			if op.CreateMissingKeys {
-				newMap := MapStr{}
-				m[k] = newMap
-				m = newMap
-				continue
-			}
-			return nil, errors.Wrapf(ErrKeyNotFound, "key=%v", strings.Join(keyParts[0:i+1], "."))
-		}
+	// Splits up the key in two parts: full key and first part before the dot
+	_, exists := data[key]
+	// If leave node or key exists directly
+	if exists {
+		// Execute the mapStrOperator on the leaf object.
+		return op.Do(key, data)
+	}
 
-		m, err = toMapStr(v)
-		if err != nil {
-			return nil, errors.Wrapf(err, "key=%v", strings.Join(keyParts[0:i+1], "."))
+	keyParts := strings.SplitN(key, ".", 2)
+	// If leave node or key exists directly
+	if len(keyParts) == 1 {
+		// Execute the mapStrOperator on the leaf object.
+		return op.Do(key, data)
+	}
+
+	// Checks if first part of the key exists
+	k := keyParts[0]
+	d, keyExist := data[k]
+	if !keyExist {
+		if op.CreateMissingKeys {
+			d = MapStr{}
+			data[k] = d
+		} else {
+			return nil, ErrKeyNotFound
 		}
 	}
 
-	// Execute the mapStrOperator on the leaf object.
-	v, err := op.Do(keyParts[len(keyParts)-1], m)
+	v, err := toMapStr(d)
 	if err != nil {
-		return nil, errors.Wrapf(err, "key=%v", key)
+		return nil, err
 	}
 
-	return v, nil
+	return walkMapRecursive(keyParts[1], v, op)
+}
+
+func walkMap(key string, data MapStr, op mapStrOperation) (interface{}, error) {
+	v, err := walkMapRecursive(key, data, op)
+	if err != nil {
+		// Add key to error
+		err = errors.Wrapf(err, "key=%v", key)
+	}
+	return v, err
 }
 
 // mapStrOperation types
--- a/libbeat/common/mapstr_test.go
+++ b/libbeat/common/mapstr_test.go
@@ -60,6 +60,16 @@ func TestMapStrDeepUpdate(t *testing.T) {
 			MapStr{"a": 1},
 			MapStr{"a": 1},
 		},
+		{
+			MapStr{"a.b": 1},
+			MapStr{"a": 1},
+			MapStr{"a": 1, "a.b": 1},
+		},
+		{
+			MapStr{"a": 1},
+			MapStr{"a.b": 1},
+			MapStr{"a": 1, "a.b": 1},
+		},
 	}
 
 	for i, test := range tests {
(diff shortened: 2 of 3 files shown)


andrewkroh:
Could you run the benchmark with -benchmem enable.

ruflin:
Here is the output:

Recursive
```
BenchmarkWalkMap/Get-4           1000000              1263 ns/op             432 B/op          7 allocs/op
BenchmarkWalkMap/Put-4           1000000              1329 ns/op            1120 B/op         10 allocs/op
BenchmarkWalkMap/HasKey-4        2000000               846 ns/op             192 B/op          6 allocs/op
BenchmarkWalkMap/HasKeyFirst-4  10000000               110 ns/op              32 B/op          1 allocs/op
BenchmarkWalkMap/Delete-4        1000000              1324 ns/op            1120 B/op         10 allocs/op
```

Iterative
```
BenchmarkWalkMap/Get-4            500000              2677 ns/op             824 B/op         13 allocs/op
BenchmarkWalkMap/Put-4           1000000              1015 ns/op            1072 B/op          8 allocs/op
BenchmarkWalkMap/HasKey-4        5000000               375 ns/op              96 B/op          2 allocs/op
BenchmarkWalkMap/HasKeyFirst-4  20000000                83.5 ns/op            16 B/op          1 allocs/op
BenchmarkWalkMap/Delete-4        1000000              1038 ns/op            1072 B/op          8 allocs/op
```

libbeat/common/mapstr.go
```diff
@@ -315,10 +315,45 @@ func tryToMapStr(v interface{}) (MapStr, bool) {
...
 // the given mapStrOperation is invoked.
-func walkMap(key string, data MapStr, op mapStrOperation) (interface{}, error) {
+func walkMapRecursive(key string, data MapStr, op mapStrOperation) (interface{}, error) {
+
+	// Splits up the key in two parts: full key and first part before the dot
+	keyParts := strings.SplitN(key, ".", 2)
+	_, exists := data[key]
+
+	// If leave node or key exists directly
+	if len(keyParts) == 1 || exists {
```
ruflin: This if could be split up in 2 parts so in case the `key` matches directly, no SplitN is needed.

libbeat/common/mapstr.go
```diff
@@ -315,10 +315,45 @@ func tryToMapStr(v interface{}) (MapStr, bool) {
...
+		v, err := op.Do(key, data)
+		if err != nil {
+			return nil, err
+		}
+		return v, nil
+	}
+
+	// Checks if first part of the key exists
+	k := keyParts[0]
+	_, keyExist := data[k]
```
andrewkroh: If you store the value here you could avoid the lookup later on 348.

libbeat/common/mapstr.go
andrewkroh: I was thinking this too. (But I don’t have any idea how common the case is.)

ruflin: Probably depends on us how we document things :-) I did split it up now as it brings some improvements in performance.

ruflin: Unfortunately we can't compare this part to the old behaviour as it does not exist there.

ruflin:
Comparison with the latest changes

Recursive
```
BenchmarkWalkMap/Get-4           1000000              1334 ns/op             432 B/op          7 allocs/op
BenchmarkWalkMap/Put-4           1000000              1273 ns/op            1120 B/op         10 allocs/op
BenchmarkWalkMap/HasKey-4        2000000               690 ns/op             160 B/op          5 allocs/op
BenchmarkWalkMap/HasKeyFirst-4  50000000                28.8 ns/op             0 B/op          0 allocs/op
BenchmarkWalkMap/Delete-4        1000000              1431 ns/op            1120 B/op         10 allocs/op

```

Iterative
```
BenchmarkWalkMap/Get-4            500000              2577 ns/op             824 B/op         13 allocs/op
BenchmarkWalkMap/Put-4           1000000              1052 ns/op            1072 B/op          8 allocs/op
BenchmarkWalkMap/HasKey-4        5000000               381 ns/op              96 B/op          2 allocs/op
BenchmarkWalkMap/HasKeyFirst-4  20000000                89.6 ns/op            16 B/op          1 allocs/op
BenchmarkWalkMap/Delete-4        1000000              1083 ns/op            1072 B/op          8 allocs/op
```

libbeat/common/mapstr.go
```diff
@@ -315,10 +315,46 @@ func tryToMapStr(v interface{}) (MapStr, bool) {
...
+		// Execute the mapStrOperator on the leaf object.
+		return op.Do(key, data)
+	}
+
+	// Checks if first part of the key exists
+	k := keyParts[0]
+	d, keyExist := data[k]
+	if !keyExist {
+		if op.CreateMissingKeys {
+			data[k] = MapStr{}
```
andrewkroh: You’ll need to set d here as well.

ruflin: Hm, good catch. Problem is when setting using `d` here instead I first have to do the type conversion / check. Let me see if I can reorder code further.

andrewkroh: I _think_ what you had was fine. `d` is an `interface{}` so you should be able to assign any value to it. Looks like you just need this:

```
d = MapStr{}
data[k] = d
```

ruflin: done and works.

urso:
@ruflin for comparing benchmarks, have a look at [benchcmp](https://godoc.org/golang.org/x/tools/cmd/benchcmp). 

I believe most of the difference in performance comes from the `splitN(, 2)` vs `split` call. I think this can be optimised even further by using an iterative approach (tail-call optimization not done by compiler) and using `strings.Index` to work on sub-strings of the key, so to reduce allocations even further.

ruflin:
@urso Thanks for the tool. Will use it next time.

For the optimizations: Put it on my todo list as I think what you suggest is definitively worth it as we call this function a lot.

PR #6407, opened 2018-02-19
Add close_check option to frequently check close options

ruflin:
At the moment close options are only checked when reaching the end of file. This had the problem that in case the output was blocked the end of the file was not reached and close options were not applied. Also in the case of very large files the close options took very long to be applied.

The current proposed workaround was to use `close_timeout` but this had the side affect that the files the harvester was always stopped. This PR introduces `close_check` which can be set to a duration. Based on the defined period it checks the `close_removed`, `close_renamed`, `close_inactive` and if the offset got smaller. `close_eof` is not checked as this does not make sense in this context.

Todo:
* Add tests
* Add docs
* Add Changelog

Discuss
* Good idea to have this config option?
* Name of config option
* Should close_check also apply to close_inactive or would this be unexpected behaviour when the output is blocked?
* Is there still value in close_timeout?


filebeat/input/log/log.go
```diff
@@ -100,10 +102,42 @@ func (f *Log) errorChecks(err error) error {
...
+		if f.backoff > f.config.MaxBackoff {
+			f.backoff = f.config.MaxBackoff
+		}
+	}
+}
+
+// Close closes the done channel but no th the file handler
+func (f *Log) Close() {
+	close(f.done)
+	// Note: File reader is not closed here because that leads to race conditions
```
ruflin: This could actually now be possible with go 1.9

tsg:
Big +1 on the improvement, and I like that the code change looks fairly simple. Is the plan to start with `close_check` disabled by default, but we start recommending it instead of `close_timeout`, and then at some point switch to it being the default?

ruflin:
@tsg That is what I had in mind. Enabling it by default would be a breaking change so we should probably do it for 7.0. I even wonder if there is still use for `close_timeout` when we have `close_check`. We could remove it in 7.0 and have `close_check` as a deprecated alias for `close_timeout` for example.

filebeat/input/log/harvester.go
```diff
@@ -195,18 +195,37 @@ func (h *Harvester) Run() error {
...
 			closeTimeout = time.After(h.config.CloseTimeout)
 		}
 
-		select {
-		// Applies when timeout is reached
-		case <-closeTimeout:
-			logp.Info("Closing harvester because close_timeout was reached: %s", source)
-		// Required when reader loop returns and reader finished
-		case <-h.done:
+	OuterLoop:
```
urso: ugh... labels :)

Have you consider to use defer for cleanup cleanup (`h.close()` and `h.log.Close()`)?

filebeat/input/log/harvester.go
```diff
@@ -195,18 +195,37 @@ func (h *Harvester) Run() error {
...
+		for {
+			select {
+			// Applies when close check is reached
+			case <-tickerChan:
+
+				if err := h.log.closeChecks(); err != nil {
+					logp.Info("Closing harvester because close_timeout was reached: %s", source)
```
urso: Do not swallow the error value. `closeChecks` returns the reason, which should be handled like in the synchronous case.

filebeat/input/log/log.go
```diff
@@ -100,10 +102,42 @@ func (f *Log) errorChecks(err error) error {
...
+}
+
+func (f *Log) closeChecks() error {
+
+	f.closeMutex.Lock()
+	defer f.closeMutex.Unlock()
```
urso: Still smells like race conditions. File processing state like `f.offset` and `f.lastTimeRead` is accessed, but potentially modified by another thread.

filebeat/input/log/harvester.go
ruflin: Yeah, also not a big fan of labels but worked here :-)

Changed to defer, much nicer.

filebeat/input/log/harvester.go
ruflin: good point

filebeat/input/log/harvester.go
```diff
@@ -311,6 +314,27 @@ func (h *Harvester) Run() error {
...
+	case ErrRemoved:
+		logp.Info("File was removed: %s. Closing because close_removed is enabled.", h.state.Source)
+	case ErrRenamed:
+		logp.Info("File was renamed: %s. Closing because close_renamed is enabled.", h.state.Source)
+	case ErrClosed:
+		logp.Info("Reader was closed: %s. Closing.", h.state.Source)
+	case io.EOF:
+		logp.Info("End of file reached: %s. Closing because close_eof is enabled.", h.state.Source)
+	case ErrInactive:
+		logp.Info("File is inactive: %s. Closing because close_inactive of %v reached.", h.state.Source, h.config.CloseInactive)
```
ruflin: Looking at it `h.state.Source` could cause a race I think :-(

filebeat/input/log/log.go
```diff
@@ -58,8 +60,10 @@ func (f *Log) Read(buf []byte) (int, error) {
 
 		n, err := f.fs.Read(buf)
 		if n > 0 {
+			f.closeMutex.Lock()
```
ruflin: This is to prevent a race for f.offset and f.lastTimeRead. It's probably better to introduce getters which have the lock as otherwise this causes issue every time the variable is used somewhere else.

PR #6412, opened 2018-02-19
Fix conditions checking on autodiscover Docker labels

exekias:
Type was wrong (map[string]string instead of MapStr) so autodiscover conditions on Docker labels were not being processed.


m4ce:
How can this be used in autodiscovery? This doesn't seem to work:

```
filebeat.autodiscover:
  providers:
    - type: docker
      templates:
        - condition:
            equals:
              docker.container.labels:
                "io.gsr.filebeat.enable": "true"
```

m4ce:
The condition seems to be wrong? 
```
-> New condition equals: map[docker.container.labels.filebeat.enable:true]
```

ruflin:
@m4ce Could you open a topic on discuss with this? https://discuss.elastic.co/c/beats/filebeat Please provide there also the list of labels you have in docker.

PR #6415, opened 2018-02-19
Fix Kubernetes dashboard views for non default time ranges

### LINKED ISSUE #6395: Kubernetes dashboard memory calculation is time range dependent
In the Kubernetes dashboard from metricbeat 6.2.1, the `Memory usage by node [Metricbeat Kubernetes]` dashboard places a red line to show available memory. This line is time range dependent on the query, it expects a 15 minute time range. Changing the time range results in a scaled graph.

- Version:  6.2.1
- Operating System: Linux / Kubernetes 1.9.2
- Steps to Reproduce: Change time range on Kubernetes dashboard

e.g.

    Quick -> Last 15 Minutes

image

    Quick -> Last 4 hours

image

There's some strange scaling on `CPU usage by node [Metricbeat Kubernetes]` as well but I'm not sure if it is a bug or some averaging / min / max thing.

exekias:
When switching to bigger time ranges, some of the visualizations in the dashboard
broke due to some wrong calculations.

I have updated them to work with any time range.

Before this change:

before

After this change:

after

Fixes #6395

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -132,6 +132,7 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
 - Fix panic in http dependent modules when invalid config was used.
 - Fix system.filesystem.used.pct value to match what df reports. {issue}5494[5494]
 - Fix namespace disambiguation in Kubernetes state_* metricsets. {issue}6281[6281]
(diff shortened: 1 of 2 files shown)


exekias:
Went through this with @simianhacker during EAH (thanks!) and he confirmed the way to go for these views. this PR should be ready for merge now.

PR #6416, opened 2018-02-20
Temporarily switching to `latest` for integration tests

tsg:
The Elasticsearch snapshots are broken for a several days
(xpack installation fails). This temporarily switches to using
latest (currently 6.2.1) for our testing, so we don't get used
to the builds being red.


ph:
@tsg Most of the tests are green, we have on integration test with ML failing, I am not sure if its a transient or a fatal error. Since this tests really depends on the x-pack artifact.

```
ERROR: Test ML are installed in all possible ways
```

tsg:
Looks like a permanent error, at least it happened  again after restarting the test on Travis. I'll try to reproduce it locally.

tsg:
Ah, it's because the docker images come with Xpack basic, but ML is in Gold.

PR #6417, opened 2018-02-20
Update Ingest pipelines if enabled

kvch:
New flag is introduced: `--update-pipelines`.

It is quite uncomfortable to delete Ingest pipelines manually before testing a small change. So during development no one cares if pipelines are updated by Filebeat unnecessarily. Or if someone does he/she can stick with the default behaviour by not specifying `--update-pipelines`.

It is intended to be used during module development or testing pipeline changes.
It should not be used when running Beats in production, because it's unnecessary to update pipelines every time a Beat is started.

### DIFF
--- a/filebeat/fileset/factory.go
+++ b/filebeat/fileset/factory.go
@@ -18,6 +18,7 @@ type Factory struct {
 	registrar             *registrar.Registrar
 	beatVersion           string
 	pipelineLoaderFactory PipelineLoaderFactory
+	updatePipelines       bool
 	beatDone              chan struct{}
 }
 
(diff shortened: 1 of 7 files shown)


filebeat/_meta/common.reference.p2.yml
```diff
@@ -270,6 +270,11 @@ filebeat.inputs:
 # Default is 0, not waiting.
 #filebeat.shutdown_timeout: 0
 
+# Defines if Filebeat pipelines have to be updated on Ingest Node.
```
ruflin: As this is not expected to be used in production I would prefer to not have it in the config file. 

What about `filebeat pipelines update` as a command? I'm hoping like this the config option would also not have to be passed through all levels.

kvch: My problem with your proposal is the user would need run two commands, instead of one.
What if it would be a command line flag? `--update-pipelines` or something like that. So it wouldn't be in the config file and users wouldn't need to run two commands.

ruflin: I like the idea of having it as a command line flag. Normally I'm oposed to have command line flags which are not in the docs but I think this one is a good example where it makes a sense.

It also should allow you to make the code more local.

kvch:
I updated the PR. FB modules dev guide requires updating.

filebeat/_meta/common.reference.p2.yml
ruflin: I assume that change is not needed anymore?

filebeat/fileset/factory.go
```diff
@@ -18,6 +18,7 @@ type Factory struct {
 	registrar             *registrar.Registrar
 	beatVersion           string
 	pipelineLoaderFactory PipelineLoaderFactory
+	updatePipelines       bool
```
ruflin: Should you directly specify the update-pipeline flag here instead of passing through all instances? It would simplify the code I think.

kvch: I would rather keep it this way. If I moved the flag here, the code would get messier places. This way, global flags could be in the same place and can be tracked easily to the place they are used.

ruflin: Got it, I missed the part that you also introduced it in the commands list.

kvch:
Yes, it is!

Issue #6361, opened 2018-02-12
Refactor input interactions and boundaries


ph:
This issue is coming from a discussion between @urso, @kvch and @ph

---

Recently we have refactored the interface used by prospectors to input, but we should do a bit more refactoring on the usage of the inputs especially between the interaction with each other.

- Currently, we have the Docker and the Log input coupled.
- We will have the Syslog and the UDP/TCP  input coupled.

We think that **An input must not use another inputs**, inputs should be isolated and mostly deal
with configuration. Instead, the TCP/UDP/Log should be refactored to be reusable components for the
inputs. They could become `InputReader` or `InputStream`, that defined in their package.

We could have the following:

- A UDP Server
- A TCP Server
- A FileWatcher for files/log

This change is necessary with the planning of adding more inputs to Filebeat and making sure the code stay testable and isolated.

This issue is created to discuss implementation details and interface required for this change.

ruflin:
I wonder if Docker, Syslog should just become a module instead of being an input. This change would mean certain modules in FB can also contain golang code which is more similar to Metricbeat. We were discussing in the past to remove the code part from the metricbeat modules but I see also advantages of having optional code there.

exekias:
You can use Syslog and Docker as inputs in modules, I think that's why it makes sense to keep them as inputs. Apache could be running on Docker, and you want to pipe that to the module

ph:
I will move this a bit forward this is blocking me for the Syslog input, I will start with the UDP and refactor TCP once its merged.

ph:
By moving this forward I will create a proposal PR, currently the UDP part of the UDP takes care of parsing the strings and create the beats event, we need to make it work with low level type, in this case byte array.

We could create a constructor with the following signature.
```golang
func NewTCP(config *Config, callback func([]byte)) {
//...
}
```

ph:
TCP and UDP changes are now in, only thing missing if the filewatcher refactor.

ph:
I don't think this is necessary.

PR #6433, opened 2018-02-21
[WIP] Add a Syslog input listening to UDP and TCP connection

ph:
# Not ready for review

**This is just a placeholder PR, since hound is a bit of a pain to deal with I have disable it on this PR and will create a new PR when it's ready for review..**
---

This commit add a new input named Syslog which can receive syslog
formated events from either UDP or TCP, currently without any SSL.
The events are locally parsed and send downstream.

The parser will support RFC3164 formatted events, the parser uses
state machines to parses the messages really fast instead of regular
expression.

The state machine are written using Ragel and that generate compact and
optimized go code, using state machine is a trade off of speed vs size
of the binary. This generated file add 32K to the final binary which is
an OK trade of.

Limitation:

Due to the nature of syslog, it's possible that some messages are not
correctly parsed, the parser is a work in progress and will benefit from
any bug report.

The best solution for them is to tag them and keep the `message` as the
raw message.

depends on #6266 and #6439

### DIFF
--- a/filebeat/input/syslog/parser.go
+++ b/filebeat/input/syslog/parser.go
@@ -1,0 +1,1389 @@
+
+//line parser.rl:1
+package main
+
+
+//line parser.go:7
+const syslog_start int = 1
+const syslog_first_final int = 69
+const syslog_error int = 0
+
+const syslog_en_main int = 1
+
+
+//line parser.rl:8
+
+
+// syslog
+//<34>Oct 11 22:14:15 mymachine su: 'su root' failed for lonvick on /dev/pts/8
+//<13>Feb  5 17:32:18 10.0.0.99 Use the BFG!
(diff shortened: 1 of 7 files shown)


ph:
This PR contains only the parser and the initial works for the input, I need to move https://github.com/elastic/beats/issues/6361 forward for the UDP/TCP So I can reuse them as component in the Syslog input.

filebeat/input/syslog/parser.go
```diff
@@ -0,0 +1,1389 @@
+
```
ph: This is generated code.. go away hound.

ph:
Closing and reopening a new one since the UDP and TCP are now merged and it will be clearer without all the hound comments.

Issue #6421, opened 2018-02-20
Mapping error with Docker Swarm labels


exekias:
Both when using `add_docker_metadata` and Docker autodiscover provider with Docker Swarm, in some cases we see mapping issues due to this labels set:

```
  "com.docker.swarm.task": "",
  "com.docker.swarm.task.id": "xxxxx",
  "com.docker.swarm.task.name": "xxxxx"
```

We should offer a way to "dedot" label names, or have an alternative way to detect the situation and avoid mapping errors.

Reference in discuss: https://discuss.elastic.co/t/using-autodiscover-feature-for-docker-does-not-work-when-running-in-swarm-mode/120342

Current workaround is to use `drop_field` processor like this:

```
processors:
  - drop_fields:
      fields:
        - "docker.containers.labels.com.docker.swarm.task"
```

exekias:
This was fixed by #6434

PR #6434, opened 2018-02-21
Add `safemapstr.Put`

exekias:
Add `safemapstr.Put`:

This method implements a way to put dotted keys into a MapStr while
ensuring they don't override each other when they are not really prepared to be stored in a tree. For example:

```
  a := MapStr{}
  a.Put("com.docker.swarm.task", "x")
  a.Put("com.docker.swarm.task.id", 1)
  a.Put("com.docker.swarm.task.name", "foobar")
```

Will result in `{"com":{"docker":{"swarm":{"task":"x"}}}}`

Put detects this scenario and renames the common base key, by appending
`.value`:

```
 a := MapStr{}
 safemapstr.Put(a, "com.docker.swarm.task", "x")
 safemapstr.Put(a, "com.docker.swarm.task.id", 1)
 safemapstr.Put(a, "com.docker.swarm.task.name", "foobar")
```

Will result in `{"com":{"docker":{"swarm":{"task":{"id":1,"name":"foobar","value":"x"}}}}}`

The plan is to use this method to convert all unstructured external data, like labels, annotations and so

Related to #6421

### DIFF
--- a/libbeat/common/safemapstr/safemapstr.go
+++ b/libbeat/common/safemapstr/safemapstr.go
@@ -28,10 +28,10 @@ func Put(data common.MapStr, key string, value interface{}) error {
 	if len(keyParts) == 1 {
 		oldValue, exists := data[key]
 		if exists {
-			switch oldValue.(type) {
-			case common.MapStr:
+			oldMap, ok := tryToMapStr(oldValue)
+			if ok {
 				// This would replace a whole object, change its key to avoid that:
-				oldValue.(common.MapStr)[alternativeKey] = value
+				oldMap[alternativeKey] = value
 				return nil
 			}
 		}
--- a/libbeat/common/safemapstr/safemapstr.go
+++ b/libbeat/common/safemapstr/safemapstr.go
@@ -1,0 +1,66 @@
+package safemapstr
+
+import (
+	"strings"
+
+	"github.com/elastic/beats/libbeat/common"
+)
+
+// Put This method implements a way to put dotted keys into a MapStr while
+// ensuring they don't override each other. For example:
+//
+//  a := MapStr{}
+//  safemapstr.Put(a, "com.docker.swarm.task", "x")
+//  safemapstr.Put(a, "com.docker.swarm.task.id", 1)
+//  safemapstr.Put(a, "com.docker.swarm.task.name", "foobar")
+//
+// Will result in `{"com":{"docker":{"swarm":{"task":{"id":1,"name":"foobar","value":"x"}}}}}`
+//
+// Put detects this scenario and renames the common base key, by appending
+// `.value`
+func Put(data common.MapStr, key string, value interface{}) error {
+	keyParts := strings.SplitN(key, ".", 2)
+
+	// If leaf node or key exists directly
+	if len(keyParts) == 1 {
+		oldValue, exists := data[key]
+		if exists {
+			switch oldValue.(type) {
+			case common.MapStr:
+				// This would replace a whole object, change its key to avoid that:
+				oldValue.(common.MapStr)["value"] = value
+				return nil
+			}
+		}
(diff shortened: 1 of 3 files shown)


exekias:
@ruflin I came up with this while working on "dedoting", what do you think?

urso:
TBH I'm kind of -1 on this.

First of all, common.MapStr is already a 'monster'. Please let's not add more methods on top of it.

There is some magic going on, which might not be always clear when and how it is triggered.
Also wonder if/how `fields.yml` needs to be aware of this magic as well? The `Put` function allows you to use flat names, but it's already doing some magic in hiding the fact names are not flat, but we have some namespacing going on.

The docker swarm case of having these kind of labels is quite bad practice. When trying to add some hierarchy into labels using dots, the labels open up some kind of namespacing. Therefore a label string must not be a prefix of another label. This should be the case, not just in the context of beats.

Instead of adding the DeDot functionality to common.MapStr, how about having autodiscovery (or some other module) detect violations in label use and fix the label naming there. We could also introduce some label-rewriting rules (with default rules). Also consider opening an issue against docker-swarm to use sane labels (what's the difference between `task` and `task.name`?).

exekias:
Thanks for the feedback, sounds good to me! I'll create a helper function to build the safe MapStr somewhere outside MapStr code :+1:

exekias:
I moved the code to a new `safemapstr` package, this should be ready for a second look :smiley:

libbeat/common/safemapstr/safemapstr.go
```diff
@@ -0,0 +1,66 @@
...
+//  a := MapStr{}
+//  safemapstr.Put(a, "com.docker.swarm.task", "x")
+//  safemapstr.Put(a, "com.docker.swarm.task.id", 1)
+//  safemapstr.Put(a, "com.docker.swarm.task.name", "foobar")
+//
+// Will result in `{"com":{"docker":{"swarm":{"task":{"id":1,"name":"foobar","value":"x"}}}}}`
+//
+// Put detects this scenario and renames the common base key, by appending
+// `.value`
+func Put(data common.MapStr, key string, value interface{}) error {
```
urso: how about introducing a 'defaultKey' to be used instead of hard-coded 'value' ?

libbeat/common/safemapstr/safemapstr.go
```diff
@@ -0,0 +1,66 @@
...
+	keyParts := strings.SplitN(key, ".", 2)
```
urso: I see it's a copy'n paste of `MapStr.Put`.

Using `strings.Index`, sub-slices and loop might make the routine even faster.

libbeat/common/safemapstr/safemapstr.go
exekias: do you mean moving it to a `const` or passing it as a parameter to `Put`?

urso: parameter. I wonder if in some cases we can declare/define a better name then 'value'.

exekias:
jenkins, retest it please

libbeat/common/safemapstr/safemapstr.go
```diff
@@ -0,0 +1,69 @@
...
+// `.value`
+func Put(data common.MapStr, key string, value interface{}) error {
+	// XXX This implementation mimics `common.MapStr.Put`, both should be updated to have similar behavior
+	keyParts := strings.SplitN(key, ".", 2)
+
+	// If leaf node or key exists directly
+	if len(keyParts) == 1 {
+		oldValue, exists := data[key]
+		if exists {
+			switch oldValue.(type) {
```
andrewkroh: I think you should use `tryToMapStr` here to be consistent with the handling of `common.MapStr` and `map[string]interface{}`.

Issue #6271, opened 2018-02-03
Unexpectedly huge byte values returned for system.memory.swap.used.bytes


ppf2:
6.1

Metricbeat is returning ginormous byte values for swap used.  There are docker and kube processes running on this bare metal server (metricbeat is running directly on the host machine, not within containers).

```
[2018-01-31T17:07:47,804][WARN ][logstash.outputs.elasticsearch] Could not index event to Elasticsearch.
{"type"=>"mapper_parsing_exception", "reason"=>"failed to parse [system.memory.swap.used.bytes]", "caused_by"=>{"type"=>"i_o_exception", "reason"=>"Numeric value (18446744070278606848) out of range of long (-9223372036854775808 - 9223372036854775807)\n at [Source: org.elasticsearch.common.bytes.BytesReference$MarkSupportingStreamInputWrapper@45bfe490; line: 1, column: 342]"}}}}}
```

Workaround is to filter out fields using a processor with these large values.  If this is not something we can avoid because of golang, it will be nice for beats to filter these invalid values automatically without manual intervention.

CentOS Linux release 7.4.1708 (Core)

```
 free -m
              total        used        free      shared  buff/cache   available
Mem:         257675        9188      244749          27        3737      247391
Swap:          4083           0        4083
```

```
top - 10:38:10 up 3 days,  3:45,  4 users,  load average: 0.53, 0.70, 0.54
Tasks: 895 total,   1 running, 894 sleeping,   0 stopped,   0 zombie
%Cpu(s):  0.7 us,  0.9 sy,  0.0 ni, 98.4 id,  0.0 wa,  0.0 hi,  0.0 si,  0.0 st
KiB Mem : 26385961+total, 25061606+free,  9416044 used,  3827500 buff/cache
KiB Swap:  4182012 total,  4182012 free,        0 used. 25332158+avail Mem
```

exekias:
Hi @ppf2, would it be possible to get the output of `cat /proc/meminfo`?

ppf2:
cat /proc/meminfo

MemTotal:    263859608 kB

MemFree:    246927004 kB

MemAvailable:  250038484 kB

Buffers:      6256 kB

Cached:     3830644 kB

SwapCached:      0 kB

Active:     12868848 kB

Inactive:    1777640 kB

Active(anon):  10819508 kB

Inactive(anon):  42668 kB

Active(file):  2049340 kB

Inactive(file): 1734972 kB

Unevictable:    24528 kB

[...]

PR #6436, opened 2018-02-21
Add support for huge pages on Linux

### LINKED ISSUE #6351: Include hugepages metrics in metricbeat
If metricbeat could capture [hugepages metrics](https://www.kernel.org/doc/Documentation/vm/hugetlbpage.txt) such as `HugePages_Total`, `HugePages_Free`,`HugePages_Rsvd`,`HugePages_Surp`,`Hugepagesize`, and `Hugetlb` it would provide a more granular view of memory usage on the system.

`system.memory.used.pct` includes memory that the system has allocated to huge pages.
e.g.
```
[ hostname ] # cat /proc/meminfo | egrep "Total|Free|Huge"
MemTotal:       132027020 kB
MemFree:        10274084 kB
…
HugePages_Total:   56274
HugePages_Free:    23506
…
Hugepagesize:       2048 kB
```
Metricbeat reports this system as having ~92% memory used based on `100-(MemFree/(MemTotal/100))` - however there’s actually 23506*2048kB available as huge pages, which is about 58% of huge pages free.

jsoriano:
Add support for huge pages on Linux, fixes #6351, requires https://github.com/elastic/gosigar/pull/97

### DIFF
--- a/libbeat/metric/system/memory/memory.go
+++ b/libbeat/metric/system/memory/memory.go
@@ -87,3 +87,35 @@ func AddSwapPercentage(s *SwapStat) {
 	perc := float64(s.Swap.Used) / float64(s.Swap.Total)
 	s.UsedPercent = common.Round(perc, common.DefaultDecimalPlacesCount)
 }
+
+// HugeTLBPagesStat includes metrics about huge pages usage
+type HugeTLBPagesStat struct {
+	sigar.HugeTLBPages
+	UsedPercent float64 `json:"used_p"`
+}
+
+// GetHugeTLBPages returns huge pages usage metrics
+func GetHugeTLBPages() (*HugeTLBPagesStat, error) {
+	pages := sigar.HugeTLBPages{}
+	err := pages.Get()
+
+	if err == nil {
+		return &HugeTLBPagesStat{HugeTLBPages: pages}, nil
+	}
+
+	if sigar.IsNotImplemented(err) {
+		return nil, nil
+	}
+
+	return nil, err
+}
+
+// AddHugeTLBPagesPercentage calculates ratio of used huge pages
+func AddHugeTLBPagesPercentage(s *HugeTLBPagesStat) {
+	if s.Total == 0 {
+		return
+	}
+
+	perc := float64(s.Total-s.Free+s.Reserved) / float64(s.Total)
+	s.UsedPercent = common.Round(perc, common.DefaultDecimalPlacesCount)
+}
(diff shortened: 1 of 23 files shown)


libbeat/metric/system/memory/memory.go
```diff
@@ -87,3 +87,48 @@ func AddSwapPercentage(s *SwapStat) {
...
+	if s.Total == 0 {
+		return
+	}
+
+	perc := float64(s.Total-s.Free+s.Reserved) / float64(s.Total)
+	s.UsedPercent = common.Round(perc, common.DefaultDecimalPlacesCount)
+}
+
+// GetHugeTLBPagesEvent returns the event created from huge page usage metrics
+func GetHugeTLBPagesEvent(s *HugeTLBPagesStat) common.MapStr {
```
exekias: I think this function is not used?

exekias:
Also have a look to system tests, you may want to check fields are in place in Linux: https://github.com/elastic/beats/blob/master/metricbeat/tests/system/test_system.py

exekias:
Failing tests should be fixed by `make update`, that will format the code to the standard and update `NOTICE.txt`

libbeat/metric/system/memory/memory.go
jsoriano: Umm, there are other similar `GetFooEvent` functions here, indeed they are not used in the beats repo, I though they were exported to be used somewhere else.

exekias:
I saw https://github.com/elastic/gosigar/pull/97 was merged :tada:. Let's move this forward once we get a new tag version of it

jsoriano:
gosigar 0.9.0 released, I will go on with this PR using this version.

libbeat/metric/system/memory/memory.go
jsoriano: @exekias WDYT? Should we keep these `GetFooEvent` functions?

jsoriano: Removed new unused function, old ones are left there by now.

ruflin:
Can you add a changelog entry?

jsoriano:
Changelog entry added.
Please wait a bit before merging this, I'd like to try to reproduce this problem seen upgrading to gosigar 0.9.0: https://discuss.elastic.co/t/metricbeat-5-3-0-reporting-strange-memory-value/124825

jsoriano:
We can continue merging this, it seems that the problem with gosigar was not introduced in 0.9.0, it was already reported before (#6271)

Issue #6191, opened 2018-01-26
There is a problem when using MongoDB with packetbeat.


pooqadmin:
Please post all questions and issues on https://discuss.elastic.co/c/beats
before opening a Github Issue. Your questions will reach a wider audience there,
and if we confirm that there is a bug, then you can open a new issue.

For security vulnerabilities please only send reports to security@elastic.co.
See https://www.elastic.co/community/security for more information.

For confirmed bugs, please report:
- Version: packetbeat-6.1.2-linux-x86_64
- Operating System: Linux contents-mongo-dev-01 4.9.20-11.31.amzn1.x86_64 #1 SMP Thu Apr 13 01:53:57 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
- Mongodb Version: db version v3.2.13
- Steps to Reproduce: 

[root@contents-mongo-dev-01 packetbeat-6.1.2-linux-x86_64]#  /beat/packetbeat-6.1.2-linux-x86_64/packetbeat -c /beat/packetbeat-6.1.2-linux-x86_64/packetbeat.yml -e 

Q) When packetbeat is executed, the following error occurs.

2018/01/26 06:30:01.311175 mongodb_parser.go:42: ERR Unknown operation code: 
2018/01/26 06:30:01.311872 mongodb_parser.go:42: ERR Unknown operation code: 
2018/01/26 06:30:01.311902 mongodb_parser.go:42: ERR Unknown operation code: 
2018/01/26 06:30:01.311928 mongodb_parser.go:42: ERR Unknown operation code: 
2018/01/26 06:30:01.403019 mongodb_parser.go:42: ERR Unknown operation code: 
2018/01/26 06:30:01.403631 mongodb_parser.go:42: ERR Unknown operation code: 

Q) My setup is as below and I simply linked it to the logstash. Please check the cause of the above error.

[root@contents-mongo-dev-01 packetbeat-6.1.2-linux-x86_64]# cat packetbeat.yml | grep -v '#'

packetbeat.interfaces.device: any

packetbeat.flows:
  timeout: 30s
  period: 10s

packetbeat.protocols:
- type: mongodb
  ports: [27017]
  send_request: true
  send_response: true
  max_docs: 0
  max_doc_length: 0

setup.template.settings:
  index.number_of_shards: 3
setup.kibana:

output.elasticsearch:
  hosts: ["internal-bh-elasticsearch-lb-1602908268.ap-northeast-2.elb.amazonaws.com:9200"]

[...]

PR #6440, opened 2018-02-22
Packetbeat: add COMMAND, COMMANDREPLY mongodb opcodes

jaipradeesh:
Ref: https://github.com/elastic/beats/issues/6191

r? @adriansr 

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -82,6 +82,15 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
 
 *Packetbeat*
 
+- Fix http status phrase parsing not allow spaces. {pull}5312[5312]
+- Fix http parse to allow to parse get request with space in the URI. {pull}5495[5495]
+- Fix mysql SQL parser to trim `\r` from Windows Server `SELECT\r\n\t1`. {pull}5572[5572]
+- Fix corruption when parsing repeated headers in an HTTP request or response. {pull}6325[6325]
+- Fix panic when parsing partial AMQP messages. {pull}6384[6384]
+- Fix out of bounds access to slice in MongoDB parser. {pull}6256[6256]
+- Fix sniffer hanging on exit under Linux. {pull}6535[6535]
+- Adds OP_COMMAND and OP_COMMANDREPLY to MongoDB response(opcode parse/check) whitelist. {pull}6440[6440]
+
 *Winlogbeat*
 
 - Fixed a crash under Windows 2003 and XP when an event had less insert strings than required by its format string. {pull}6247[6247]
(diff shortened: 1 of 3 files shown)


ruflin:
@jaipradeesh Could you add some tests with pcap files for this if possible? Also please add a line to the changelog file.

Still needs review from probably @adriansr

jaipradeesh:
```
15:41 < dolftax> In need of pcap files to test OP_COMMAND and OP_COMMANDREPLY opcodes // How to trigger them?
15:46 <@Derick> dolftax: let me check for you
15:46 <@Derick> dolftax: what do you need the pcap files for?
15:47 <@Derick> dolftax: I believe these are used by the mongos balancer in a sharded environment
15:48 <@Derick> and only with MongoDB 3.4 (although 3.2 and 3.6 support it too for up/downgrade)
15:49 < dolftax> Just need to know it some calls to DB from the client returns such opcodes. Since what queries is made by the client is not in my control, just need to run a quick test on the response opcodes and see if it sends any invalid opcodes.
15:49 < dolftax> invalid -> not whitelisted by us.
15:49 <@Derick> I would strongly suggest you do not white or blacklist any opcodes that MongoDB uses, as this can change in the future without warning
15:51 <@Derick> But: mongos 3.4 balancer commands use it
15:52 <@Derick> and to illustrate that op codes get added all the time: https://derickrethans.nl/wireshark-mongo-36.html
15:53 < dolftax> So, https://docs.mongodb.com/manual/reference/mongodb-wire-protocol/ are not really a whitelist of opcodes?
15:54 <@Derick> no
15:54 <@Derick> it misses things
15:55 <@Derick> do not restrict what MongoDB can send/receive
```

@ruflin In such case, I don't think validating across the whitelist of opcodes makes sense. What do you think?

andrewkroh:
LGTM, but I would feel better about merging if there were a test covering these opcodes. Would you be able to add a test case? Like in either of:

- https://github.com/elastic/beats/blob/master/packetbeat/tests/system/test_0025_mongodb_basic.py
- https://github.com/elastic/beats/blob/master/packetbeat/protos/mongodb/mongodb_test.go

jaipradeesh:
@andrewkroh I was waiting for @adriansr 's input on this because (I was looking for a sample pcap file with OP_COMMAND and OP_COMMANDREPLY codes) Derick from MongoDB team suggested not to whitelist any opcodes as it might change without warning. An example scenario from Wireshark who did the same (whitelist opcodes on wire protocol) - https://derickrethans.nl/wireshark-mongo-36.html

For example: `OP_COMPRESSED` isn't documented on Wire protocol opcode docs - https://docs.mongodb.com/manual/reference/mongodb-wire-protocol/ but it is a valid one.

Ref: https://github.com/elastic/beats/issues/6191#issuecomment-368825298

adriansr:
Hi @jaipradeesh 

How do you feel about adding the test?

I think best is adding a system-test (the python ones) that reads a pcap file and checks that no warning/errors are printed to the log (`self.assert_no_logged_warnings()`).

Issue #5637, opened 2017-11-18
schema.go:41: ERR Error on field 'server_uptime': Missing field: server_uptime, Error: Key `ServerUptimeSeconds` not found


binary64:
18/11/2017 12:51:492017/11/18 12:51:49.692005 schema.go:41: ERR Error on field 'server_uptime': Missing field: server_uptime, Error: Key `ServerUptimeSeconds` not found
18/11/2017 12:51:492017/11/18 12:51:49.692028 schema.go:41: ERR Error on field 'user': Missing field: user, Error: Key `CPUUser` not found
18/11/2017 12:51:492017/11/18 12:51:49.692037 schema.go:41: ERR Error on field 'system': Missing field: system, Error: Key `CPUSystem` not found
18/11/2017 12:51:492017/11/18 12:51:49.692044 schema.go:41: ERR Error on field 'children_user': Missing field: children_user, Error: Key `CPUChildrenUser` not found
18/11/2017 12:51:492017/11/18 12:51:49.692051 schema.go:41: ERR Error on field 'children_system': Missing field: children_system, Error: Key `CPUChildrenSystem` not found

My `server-status?auto` looks like this:-

```

Total Accesses: 65
Total kBytes: 52
CPULoad: .00496689
Uptime: 604
ReqPerSec: .107616
BytesPerSec: 88.1589
BytesPerReq: 819.2
BusyWorkers: 2
IdleWorkers: 5
Scoreboard: __W___K...............................................................................................................................................

```

Apache Version | Apache/2.4.10 (Debian)

ruflin:
This is interesting as we use `2.4.20` Docker container for testing. Which version of debian is this? How did you install apache? I wonder why here for example `ServerUptimeSeconds` is missing. 

I remember we had a discussion about making some more fields optional in apache in an issue but I couldn't find it :-(

ruflin:
@radoondas Pinging you on this one, perhaps you know more?

binary64:
I used `FROM php:7.1-apache` and `RUN a2enmod status`

ruflin:
@binary64 Thanks for the info. That is pretty useful. I wonder what the difference is between the version in the `httpd` Docker container and the one shipped with php. Do you know any details here?

ruflin:
@jsoriano Closing this as https://github.com/elastic/beats/pull/6450 is merged and should resolve this issue.

PR #6450, opened 2018-02-22
Metricbeat: Support multiple apache status page versions

jsoriano:
In Apache 2.4.16 server status page changed to provide more CPU information and uptime in human format (#5637). These changes make apache module support old and new schemas.

To test it, I add a second apache with the latest version using the old format, but it doesn't look very orthodox as we are not doing it with any other module. @exekias @ruflin wdyt? do we have any other approach when needing to support multiple formats?


metricbeat/module/apache/status/data.go
```diff
@@ -141,6 +163,9 @@ func eventMapping(scanner *bufio.Scanner, hostname string) (common.MapStr, *s.Er
 		},
 	}
 	_, err := schema.ApplyTo(event, fullEvent)
+	if err != nil {
```
ruflin: Is there a way to identify on the first request which Apache version it is to then pick the schema? Otherwise we decode every event twice.

metricbeat/module/apache/status/data.go
```diff
@@ -51,6 +50,29 @@ var (
 			"15": c.Float("Load15", s.Optional),
 		},
 	}
+
+	// Schema used till apache 2.4.12
+	schemaOld = s.Schema{
```
ruflin: We must make sure all these fields are also in the `fields.yml` documented.

metricbeat/module/apache/status/data.go
jsoriano: Umm, indeed I can do it depend on the different fields seen in `fullEvent` instead of completelly decoding it twice.

metricbeat/module/apache/status/data.go
jsoriano: schemaOld contains a subset of the fields of the current schema.

jsoriano: Well, I have removed the `server_uptime` field, it looks redundant to me with the `uptime` field. Also I think `uptime` shouldn't be a group, it should be enough with a `long` for this value, but changing from group to long looks fearful for backwards compatibility.

ruflin: Let's not introduce a breaking change here as this potentially goes out in a minor.

We should start a list of fields we want to change in the next master where we can break things.

For `server_uptime` vs `uptime` I'm not sure if one is from the actual host itself and the other one apache?

jsoriano: Right, I revert these changes on `server_update`.

exekias:
I'm ok with this approach while we work on a permanent solution, please have a look to failing tests, they look legit

jsoriano:
jenkins, test it

jsoriano:
Changelog and info about tested versions in doc updated

PR #6451, opened 2018-02-22
Fix examples by removing old syntax

dedemorton:
I noticed while working on a different PR that some of our examples use the old way of specifying paths. Maybe this still works for backwards compatibility (?), but it's potentially confusing for users.

I noticed that testfilebeat/tests/system/config/filebeat_inputs.yml.j2 also uses the old syntax, but I didn't fix it because I don't know if that test is still used.

Please eyeball the examples carefully. I didn't run them to test because the change seemed small.

### DIFF
--- a/filebeat/docs/filebeat-filtering.asciidoc
+++ b/filebeat/docs/filebeat-filtering.asciidoc
@@ -68,7 +68,8 @@ The following configuration decodes the inner JSON object:
 [source,yaml]
 -----------------------------------------------------
 filebeat.prospectors:
-- paths:
+- type: log
+  paths:
     - input.json
   json.keys_under_root: true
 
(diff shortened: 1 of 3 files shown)


ruflin:
@dedemorton I'm not sure I can follow what you mean by the old way? Are you referring that we skip the default type? These three are all the same:

```
- type: log
  paths:
    - input.json
```

```
  paths:
    - input.json
```

```
- paths:
    - input.json
  type: log
```

dedemorton:
@ruflin Sorry, my confusion! I thought the type was required. 🤦‍♀️ Ironically I did not RTFM. 

I wonder if it's less confusing for new users, though, to see the type defined explicitly.

ruflin:
@dedemorton I think it's a good change to add the type to the config especially as we recently added more. I was just confused that you wrote "the old way".

PR #6456, opened 2018-02-23
Preserve runtime from container statuses in Kubernetes autodiscover

vjsamuel:
The initial implementation assumed always that containers come from docker. There are kube clusters that run `rkt` as the container runtime. Atleast we would not spin up docker prospectors/inputs by default when Beats is run in them and based on `rkt` behavior we can atleast manually set a config template on the `logs` builder config.

### DIFF
--- a/libbeat/autodiscover/providers/kubernetes/kubernetes.go
+++ b/libbeat/autodiscover/providers/kubernetes/kubernetes.go
@@ -111,7 +111,7 @@ func (p *Provider) emitEvents(pod *kubernetes.Pod, flag string, containers []kub
 	containerstatuses []kubernetes.PodContainerStatus) {
 	host := pod.Status.PodIP
 
-	// Collect all container IDs and runtimes from status information. Kubernetes has both docker and rkt
+	// Collect all container IDs and runtimes from status information.
 	containerIDs := map[string]string{}
 	runtimes := map[string]string{}
 	for _, c := range containerstatuses {
@@ -207,14 +207,10 @@ func (p *Provider) generateHints(event bus.Event) bus.Event {
 
 	if rawCont, ok := kubeMeta["container"]; ok {
 		container = rawCont.(common.MapStr)
-		// This would end up adding a docker|rkt.container entry into the event. This would make sure
+		// This would end up adding a runtime entry into the event. This would make sure
 		// that there is not an attempt to spin up a docker input for a rkt container and when a
 		// rkt input exists it would be natively supported.
-		if runtime, ok := container["runtime"]; ok {
-			e[runtime.(string)] = common.MapStr{
-				"container": container,
-			}
-		}
+		e["container"] = container
 	}
 
 	cname := builder.GetContainerName(container)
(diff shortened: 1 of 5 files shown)


ruflin:
jenkins, test it

vjsamuel:
@ruflin we do not but it is definitely possible to have it in a Kube cluster.

ruflin:
@vjsamuel Interesting, was not aware that in one cluster you can mix the two.

vjsamuel:
A single node can have only one runtime but there can be carved out nodes that do docker and some that do rkt if im not mistaken.

libbeat/autodiscover/providers/kubernetes/kubernetes.go
```diff
@@ -204,8 +207,13 @@ func (p *Provider) generateHints(event bus.Event) bus.Event {
 
 	if rawCont, ok := kubeMeta["container"]; ok {
 		container = rawCont.(common.MapStr)
-		e["docker"] = common.MapStr{
-			"container": container,
+		// This would end up adding a docker|rkt.container entry into the event. This would make sure
+		// that there is not an attempt to spin up a docker input for a rkt container and when a
+		// rkt input exists it would be natively supported.
+		if runtime, ok := container["runtime"]; ok {
+			e[runtime.(string)] = common.MapStr{
```
exekias: I've second thoughts about our current field names. I think we should rename `docker.container.*` to `container.*` for 7.0. `container.runtime` will give you `docker`, `rkt` or any other.

For the time being, I'm ok with this I think, as removing the `docker` prefix would be a breaking change

libbeat/autodiscover/providers/kubernetes/kubernetes.go
```diff
@@ -111,19 +111,22 @@ func (p *Provider) emitEvents(pod *kubernetes.Pod, flag string, containers []kub
 	containerstatuses []kubernetes.PodContainerStatus) {
 	host := pod.Status.PodIP
 
-	// Collect all container IDs
+	// Collect all container IDs and runtimes from status information. Kubernetes has both docker and rkt
```
exekias: minor thing here, but it could be others at some point (apart from docker and rkt)

vjsamuel: should i update this?

exekias: if you want :)

libbeat/autodiscover/providers/kubernetes/kubernetes.go
exekias: What do you think about this @vjsamuel? I'm thinking we could apply the change for autodiscover already, somehow duplicating `docker.container` == `container`.

Then in 7.0 we update metadata processor to use the new fields

libbeat/autodiscover/providers/kubernetes/kubernetes.go
vjsamuel: done

exekias:
jenkins, test this please

exekias:
Thanks for the changes @vjsamuel :tada:

PR #6459, opened 2018-02-23
Avoid git cloning from the deb6 image

tsg:
It seems that by now Debian 6 is so old that `git clone` from Github
repositories no longer works. We only needed git clone from it to
get gotpl running, which is a relatively small program so I included it
and its dependency (go-yaml) as a zip file.

Tested `make package` locally.

### DIFF
--- a/dev-tools/packer/docker/xgo-image-deb6/beats-builder/Dockerfile
+++ b/dev-tools/packer/docker/xgo-image-deb6/beats-builder/Dockerfile
@@ -20,11 +20,14 @@ RUN ./fetch.sh http://www.tcpdump.org/release/libpcap-1.8.1.tar.gz 32d7526dde8f8
   ./configure --enable-usb=no --enable-bluetooth=no --enable-dbus=no && \
   make
 
-# Old git version which does not support proxy with go get requires to fetch go-yaml directly
-RUN git clone https://github.com/go-yaml/yaml.git /go/src/gopkg.in/yaml.v2
+# Old git version which does not support proxy with go get directly, and git cloning
+# from github also no longer works.
+ADD yaml.v2 /go/src/gopkg.in/yaml.v2
+ADD gotpl /go/src/github.com/tsg/gotpl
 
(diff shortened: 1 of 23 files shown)


tsg:
The longer term plan is to reconsider if we still need support for Debian 6.

graphaelli:
Can we use zips from https://github.com/go-yaml/yaml/archive/v2.zip and https://github.com/tsg/gotpl/archive/master.zip instead of adding them to the repo?

dev-tools/packer/docker/xgo-image-deb6/beats-builder/Dockerfile
```diff
@@ -20,11 +20,18 @@ RUN ./fetch.sh http://www.tcpdump.org/release/libpcap-1.8.1.tar.gz 32d7526dde8f8
   ./configure --enable-usb=no --enable-bluetooth=no --enable-dbus=no && \
   make
 
-# Old git version which does not support proxy with go get requires to fetch go-yaml directly
-RUN git clone https://github.com/go-yaml/yaml.git /go/src/gopkg.in/yaml.v2
+# Old git version which does not support proxy with go get directly, and git cloning
+# from github also no longer works.
+ADD go-yaml.tar.gz /go-yaml.tar.gz
```
andrewkroh: I think we would have move visibility into this dependency if it were copied from the dev-tools/vendor directory. This will probably be better for maintainability.

dev-tools/packer/docker/xgo-image-deb6/build.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/sh
-
+cp -r ../../../vendor/gopkg.in/yaml.v2 beats-builder/yaml.v2
```
andrewkroh: Probably should add these dirs to .gitignore.

tsg:
`make package` seems to work locally, with the vendored version.

tsg:
Added `.gitignore` and rebased to cleanup the history.

PR #6461, opened 2018-02-23
Remove dependency on gopkg.in

tsg:
This removes the remaining `go get` for gotpl, by using the vendored
copy. This should get rid of our dependency on the gopkg.in service.


tsg:
`make package` seems to work locally.

Issue #6453, opened 2018-02-23
Support for wildcard in jolokia module in metricbeat


diavel:
Please post all questions and issues on https://discuss.elastic.co/c/beats
before opening a Github Issue. Your questions will reach a wider audience there,
and if we confirm that there is a bug, then you can open a new issue.

For security vulnerabilities please only send reports to security@elastic.co.
See https://www.elastic.co/community/security for more information.

**Describe the feature**:
Support wildcards for Metricbeat Jolokia module

As pointed out in this [discussion](https://discuss.elastic.co/t/metricbeat-jolokia-problem-using-wildcard/114943), there are many places where the wildcard support in the jolokia module would come in handy. I am trying to collect jmx metrics from flink through the jolokia module and lack of wildcard support makes it unusable. Jolokia does support wildcard in it's queries. 

**Steps to reproduce**:

Please include a *minimal* but *complete* recreation of the problem, including
(e.g.) index creation, mappings, settings, query etc.  The easier you make for
us to reproduce it, the more likely that somebody will take the time to look at it.

 1. Setup Jolokia metric beat module
 2. Use jmx mapping similar to what is shown below. 
```
jmx.mappings:
       mbean: 'Catalina:name=*,type=ThreadPool'
       attributes:
                  attr: maxConnections
                  field: max_connections 3. 
```

**Provide logs (if relevant)**:
Error observed when using wildcard

` errors: metric key '' not found in response; metric key '' not found in response`

ruflin:
Thanks for the feature request. I agree it would be valuable to have this feature in our module.

daniilyar:
We would very benefit from wildcards support also. Our Jolokia configs for things like Kafka, Spark and Hadoop have thousands of lines that makes them really hard to maintain.

Just try to gather Kafka JMX metrics for e.g. 50 topics to feel our pain )
In Kafka JMX for each topic there are same requestsPerSec, bytesPerSec and 20+ other metrics that without wildcards one should just hardcode for each topic. 20+ metrics multiplied by 50 topics = 1000 metrics to hardcode, or 2000 lines in config file. And that would be only topic metrics, imagine that there are also ~10 identical metrics metrics for each of 15 Kafka request types, and many others.

ruflin:
@daniilyar Thanks for sharing your use case. Perhaps you can also chime in on https://github.com/elastic/beats/pull/6462 to what your preferred way would be related to event generation (see discussion).

jsoriano:
[...]

PR #6462, opened 2018-02-23
Metricbeat: Support wildcards in jolokia (#6453)

jsoriano:
Partially support wildcards in jolokia. This would cover the case of #6453 where wildcard is used to match one of two possible names. In a more general case this implementation is incomplete because if multiple mbeans match, their field mapping will collide and only one value will be stored.

To solve this I see two options:

1. Generate an event for each matching mbean, so each event has a field with the name of the mbean. For example, current events are like this:
```
{
  "port": 8080,
  "max_connections": 200,
  "_namespace": "testnamespace"
}
```
New events would contain also the mbean, so with a wildcard these events could be generated:
```
{
  "port": 8080,
  "max_connections": 200,
  "mbean": "Catalina:name=\"http-bio-8080\",type=ThreadPool",
  "_namespace": "testnamespace"
}

{
  "port": 8009,
  "max_connections": 200,
  "mbean": "Catalina:name=\"ajp-bio-8009\",type=ThreadPool",
  "_namespace": "testnamespace"
}
```

2. Add a regular expression or similar to the attributes mapping that can be used to fill placeholders in the field name, e.g:
```
jmx.mappings:
       mbean: 'Catalina:name=*,type=ThreadPool'
       attributes:
                  attr: maxConnections
                  field: "${name}.max_connections"
                  mbeanFields: 'Catalina:name=(?P<name>[^,]),type=ThreadPool'
```
Would generate events like this one:
```
{
  "http-bio-8080": {
    "port": 8080,
    "max_connections": 200
  },
  "ajp-bio-8009": {
    "port": 8009,
    "max_connections": 200
  },
  "_namespace": "testnamespace"
}
```
I think option 1 is more intuitive, but I'm not sure if generating multiple events fits with the general design of the module, opinions?

### DIFF
--- a/metricbeat/module/jolokia/jmx/config.go
+++ b/metricbeat/module/jolokia/jmx/config.go
@@ -10,6 +10,7 @@ type JMXMapping struct {
 type Attribute struct {
 	Attr  string
 	Field string
+	Event string
 }
 
 // RequestBlock is used to build the request blocks of the following format:
@@ -37,8 +38,22 @@ type RequestBlock struct {
 	Attribute []string `json:"attribute"`
 }
 
-func buildRequestBodyAndMapping(mappings []JMXMapping) ([]byte, map[string]string, error) {
-	responseMapping := map[string]string{}
+type attributeMappingKey struct {
+	mbean, attr string
+}
+
+// AttributeMapping contains the mapping information between attributes in Jolokia
+// responses and fields in metricbeat events
+type AttributeMapping map[attributeMappingKey]Attribute
+
+// Get the mapping options for the attribute of an mbean
+func (m AttributeMapping) Get(mbean, attr string) (Attribute, bool) {
+	a, found := m[attributeMappingKey{mbean, attr}]
+	return a, found
+}
+
+func buildRequestBodyAndMapping(mappings []JMXMapping) ([]byte, AttributeMapping, error) {
+	responseMapping := make(AttributeMapping)
 	var blocks []RequestBlock
 
 	for _, mapping := range mappings {
@@ -49,7 +64,7 @@ func buildRequestBodyAndMapping(mappings []JMXMapping) ([]byte, map[string]strin
 
 		for _, attribute := range mapping.Attributes {
 			rb.Attribute = append(rb.Attribute, attribute.Attr)
-			responseMapping[mapping.MBean+"_"+attribute.Attr] = attribute.Field
+			responseMapping[attributeMappingKey{mapping.MBean, attribute.Attr}] = attribute
 		}
 		blocks = append(blocks, rb)
 	}
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -257,6 +257,7 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
 - Making the MongoDB module GA. {pull}6554[6554]
 - Allow to disable labels `dedot` in Docker module, in favor of a safe way to keep dots. {pull}6490[6490]
 - Add experimental module to collect metrics from munin nodes. {pull}6517[6517]
+- Add support for wildcards and explicit metrics grouping in jolokia/jmx. {pull}6462[6462]
 
 *Packetbeat*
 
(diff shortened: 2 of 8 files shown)


ruflin:
Could you share an example of two json events on how they would look like with option 1?

jsoriano:
@ruflin I have added an example for option 1 to the description.

jsoriano:
I guess that something like option 2 fits better with `EventFetcher` interface, so for a `Fetch()` an only event is generated.

jsoriano:
After talking offline with @exekias about this I will go for option 1 implementing the `EventsFetcher` interface that allows to generate multiple events in an only call.

ruflin:
* For 1 event per mbean I'm wondering on how many metrics we expect per mbean?
* The reason I prefer multiple mbeans in one event is that the user can always configure metricbeat to have multiple instances of the metricset and configure 1 mbean per event if he wants to (except for the wildcard). 
* Should we make it a config option if it should be 1 or multiple events?
* For the name conflicts, can saftemapstr help here?

jsoriano:
@ruflin thanks for your comments, find answers inline.

> - For 1 event per mbean I'm wondering on how many metrics we expect per mbean?

It will depend on the mappings the user configures, for example for the case mentioned in https://github.com/elastic/beats/issues/6453#issuecomment-372855357 it'd be about 20 metrics per matching mbean, but it could be only 2 as in the example I posted.

> - The reason I prefer multiple mbeans in one event is that the user can always configure metricbeat to have multiple instances of the metricset and configure 1 mbean per event if he wants to (except for the wildcard).

The case of building an only event from multiple known mbeans would still be supported, as well as the case of generating multiple events for known mbeans using multiple instances of the module. Multiple events would only be generated when using wildcards.

> - Should we make it a config option if it should be 1 or multiple events?

I think this would be more confusing, in general if a wildcard is used, multiple matches of similar things should be expected, so as I see it, multiple events should be generated to avoid conflicts with field names, and to support the case of https://github.com/elastic/beats/issues/6453#issuecomment-372855357 that I think it fits better with the use of wildcards.
If no wildcard is used, then the user is building an only event from one or multiple known mbeans (the case supported now).
There could be a case where the user wants to build an event from mixed known mbeans and a wildcard that is expected to match only once, this seems a tricky corner case, but it could be actually supported by allowing to force the generation of an only event no matter what.
What do you think, should we support this case?

> - For the name conflicts, can saftemapstr help here?

Well, conflicts here wouldn't come from having dots in the names but by having directly multiple fields with exactly the same name coming from similar mbeans.
On the other hand, names for the fields come from user configuration, so if it has conflicts by dots it could be avoided by defining proper names, we could notice that in the documentation. Or we could use `safemapstr` to prevent problems with this... Opinions?

jsoriano:
As talked offline, and in the line of #6585, I have added explicit grouping of metrics. An optional `event` parameter can be added to each attribute mapping so metrics with the same `event` are sent in the same event to Elastic.

jsoriano:
jenkins, test it again please

metricbeat/module/jolokia/jmx/config.go
```diff
@@ -37,8 +38,22 @@ type RequestBlock struct {
...
+	mbean, attr string
+}
+
+// AttributeMapping contains the mapping information between attributes in Jolokia
+// responses and fields in metricbeat events
+type AttributeMapping map[attributeMappingKey]Attribute
+
+// Get the mapping options for the attribute of an mbean
+func (m AttributeMapping) Get(mbean, attr string) (Attribute, bool) {
+	a, found := m[attributeMappingKey{mbean, attr}]
```
ruflin: Nit: Seems like you could return here directly.

ruflin:
@jsoriano Thanks for pushing this forward. Could you update the PR message above with the end solution you went with so people coming to this PR know directly what changed?

Would be great to have a blog post about jolokia/jmx monitoring.

metricbeat/module/jolokia/jmx/config.go
jsoriano: I think it doesn't work when the expression is a value obtained from a map.

jsoriano:
Updated description

NeQuissimus:
Is this going to make it into 6.3 or left for 7.0?

ruflin:
@NeQuissimus Will be in 6.3

PR #6463, opened 2018-02-23
Prevent to start filebeat when STDIN is used with other inputs

ph:
- **breaking changes:** Stdin will now requires exclusive mode to run.
- Added helpers on top of the config class to retrieve enabled plugins.
- Refactored the stdin python integration testing into their own file.
- Added tests to check the combination of STDIN with other plugins
- Loop once on the enabled inputs


ph:
Due to the nature of stdin related to EOF, it more appropriate to make sure the Stdin input is run in exclusive mode.

CHANGELOG.asciidoc
```diff
@@ -47,6 +47,8 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
 - Refactor the usage of prospector to input in the YAML reference and
   the system test. {pull}6121[6121]
 - Add IIS module to parse access log and error log. {pull}6127[6127]
+- Add validation for Stdin, when Filebeat is configured with Stdin and any other inputs, Filebeat
```
ruflin: I wonder if this is a bug fix or a breaking change :-)

ruflin:
@ph We should add a note about this also to the docs.

ph:
@ruflin I need to take a look at config reload, maybe we should do some logic to prevent it there too can be a separate PR.

CHANGELOG.asciidoc
ph: Its kind of both 🍁

ph:
I have added a note to the documentation.

@tsg could you take a look?

filebeat/config/config_test.go
```diff
@@ -95,3 +96,81 @@ func TestMergeConfigFiles(t *testing.T) {
...
+		tests := []struct {
+			name     string
+			input    string
+			expected bool
+			config   *Config
+		}{
+			{name: "input exists and enabled", input: "stdin", expected: true, config: config},
+			{name: "input exists and disabled", input: "udp", expected: false, config: config},
+			{name: "input doesn't exist", input: "redis", expected: false, config: config},
+			{name: "not inputs are enabled", input: "redis", expected: false, config: &Config{}},
```
ruflin: s/not/no/

filebeat/docs/inputs/input-stdin.asciidoc
```diff
@@ -9,6 +9,8 @@
 
 Use the `stdin` input to read events from standard in.
 
+Note: This input cannot be run at the same time of other inputs type.
```
ruflin: at the same time "with" other input types?  (plural types, singluar input?)

ph:
@ruflin typos fixed, rebased with master.

PR #6468, opened 2018-02-26
Change crosscompile target to depend on gox as default

ruflin:
With https://github.com/mitchellh/gox/pull/96 gox has been update to support all the platforms for 1.10. Instead of depending on our self defined list as default this changes to dependn on gox defaults.

The difference on platforms built by default looks as following:

Removed:
```
-->     linux/ppc64: github.com/elastic/beats/filebeat
-->   linux/ppc64le: github.com/elastic/beats/filebeat
-->     linux/arm64: github.com/elastic/beats/filebeat
```

Added:
```
-->      darwin/386: github.com/elastic/beats/filebeat
```

It's still possible as before to overwrite these defaults in each Beat.

The gox platforms depend on the go version used. Because of this updating the Go version on our side could bring new platforms. But as we version the go version as part of the repo we would detect this directly. Having said that, recently not new platforms have been added.

This PR also changes the defaults for Filebeat to depend on the gox defaults.

### DIFF
--- a/libbeat/scripts/Makefile
+++ b/libbeat/scripts/Makefile
@@ -52,8 +52,8 @@ NOSETESTS_OPTIONS?=--process-timeout=$(TIMEOUT) --with-timer -v --with-xunit --x
 TEST_ENVIRONMENT?=false ## @testing if true, "make testsuite" runs integration tests and system tests in a dockerized test environment
 SYSTEM_TESTS?=false ## @testing if true, "make test" and "make testsuite" run unit tests and system tests
 STRESS_TESTS?=false ## @testing if true, "make test" and "make testsuite" run also run the stress tests
-GOX_OS?=linux darwin windows freebsd netbsd openbsd ## @Building List of all OS to be supported by "make crosscompile".
-GOX_OSARCH?=!darwin/arm !darwin/arm64 !darwin/386 ## @building Space separated list of GOOS/GOARCH pairs to build by "make crosscompile".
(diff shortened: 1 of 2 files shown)


libbeat/scripts/Makefile
```diff
@@ -52,8 +52,8 @@ NOSETESTS_OPTIONS?=--process-timeout=$(TIMEOUT) --with-timer -v --with-xunit --x
 TEST_ENVIRONMENT?=false ## @testing if true, "make testsuite" runs integration tests and system tests in a dockerized test environment
 SYSTEM_TESTS?=false ## @testing if true, "make test" and "make testsuite" run unit tests and system tests
 STRESS_TESTS?=false ## @testing if true, "make test" and "make testsuite" run also run the stress tests
-GOX_OS?=linux darwin windows freebsd netbsd openbsd ## @Building List of all OS to be supported by "make crosscompile".
-GOX_OSARCH?=!darwin/arm !darwin/arm64 !darwin/386 ## @building Space separated list of GOOS/GOARCH pairs to build by "make crosscompile".
+GOX_OS?=## @Building List of all OS to be supported by "make crosscompile".
```
ruflin: I have seconds thoughts on dropping this part because with the previous version we had "hardcoded" on what we wanted to test as default. Other thoughts?

ruflin:
@andrewkroh I'm going to close this PR as I have second thoughts about depending on a third party lib on which environments we build / crosscompile. I rather extend the list in libbeat and each beat that we crosscompile against.

PR #6470, opened 2018-02-26
Add ARM build to our packaging

ruflin:
For my own need I played around with packaging arm in a tar.gz. The following changes to the packaging make ARM builds available for all Beats except packetbeat. The ARM build is disabled for packetbeat. Not sure if pcap is the issue: google/gopacket#100

The environment variable `PACKAGES_EXPERIMENTAL` is introduced to make sure ARM packages are at the moment only built for SNAPSHOT builds and not when `SNAPSHOT=no`. Like this is should not have any effect on the current packaging.

This change does not mean ARM is supported but it allows us to get more testing on it.

### DIFF
--- a/libbeat/scripts/Makefile
+++ b/libbeat/scripts/Makefile
@@ -79,9 +79,10 @@ PERM_EXEC?=$(shell [ `uname -s` = "Darwin" ] && echo "+111" || echo "/a+x")
 
 # Cross compiling targets
 CGO?=true ## @building if true, Build with C Go support
-TARGETS?="windows/amd64 windows/386 darwin/amd64" ## @building list of platforms/architecture to be built by "make package"
+TARGETS?="windows/amd64 windows/386 darwin/amd64 linux/arm" ## @building list of platforms/architecture to be built by "make package"
 TARGETS_OLD?="linux/amd64 linux/386" ## @building list of Debian6 architecture to be built by "make package" when CGO is true
-PACKAGES?=${BEAT_NAME}/deb ${BEAT_NAME}/rpm ${BEAT_NAME}/darwin ${BEAT_NAME}/win ${BEAT_NAME}/bin ## @Building List of OS to be supported by "make package"
+PACKAGES?=${BEAT_NAME}/deb ${BEAT_NAME}/rpm ${BEAT_NAME}/darwin ${BEAT_NAME}/win ${BEAT_NAME}/bin ${BEAT_NAME}/arm ## @Building List of OS to be supported by "make package"
+PACKAGES_EXPERIMENTAL?=${BEAT_NAME}/arm ## @Building List of experimental OS by "make package". Only build when SNAPSHOT=yes
 SNAPSHOT?=yes ## @Building If yes, builds a snapshot version
 BEATS_BUILDER_IMAGE?=tudorg/beats-builder ## @Building Name of the docker image to use when packaging the application
 BEATS_BUILDER_DEB_IMAGE?=tudorg/beats-builder-deb7 ## @Building Name of the docker image to use when packaging the application for Debian 7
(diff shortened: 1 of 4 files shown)


andrewkroh:
What kind of errors are you seeing with cgo enabled? Maybe we are missing the cross-compiler for arm from the docker image?

libbeat/scripts/Makefile
```diff
@@ -74,9 +74,9 @@ PERM_EXEC?=$(shell [ `uname -s` = "Darwin" ] && echo "+111" || echo "/a+x")
 
 # Cross compiling targets
 CGO?=true ## @building if true, Build with C Go support
-TARGETS?="windows/amd64 windows/386 darwin/amd64" ## @building list of platforms/architecture to be built by "make package"
+TARGETS?="windows/amd64 windows/386 darwin/amd64 linux/arm" ## @building list of platforms/architecture to be built by "make package"
 TARGETS_OLD?="linux/amd64 linux/386" ## @building list of Debian6 architecture to be built by "make package" when CGO is true
-PACKAGES?=${BEAT_NAME}/deb ${BEAT_NAME}/rpm ${BEAT_NAME}/darwin ${BEAT_NAME}/win ${BEAT_NAME}/bin ## @Building List of OS to be supported by "make package"
+PACKAGES?=${BEAT_NAME}/deb ${BEAT_NAME}/rpm ${BEAT_NAME}/darwin ${BEAT_NAME}/win ${BEAT_NAME}/bin ${BEAT_NAME}/arm ## @Building List of OS to be supported by "make package"
```
ruflin: @tsg Will this have any affect on which packages are published? It would be nice to publish this for our nightlies but not have it in the releases to make sure it's clear that this is not supported.

andrewkroh: I'm pretty sure that the release manager will publish them if they exists. It doesn't name each artifact but instead points at the upload directory.

andrewkroh:
> I disabled packetbeat for now.

Packetbeat is probably a little different from the rest of the Beats due to the pcap dependency. IIRC We build libpcap from source for each target platform.

Do the rest of the Beats build for ARM with cgo on?

ruflin:
When I build the binary with `GOOS=linux GOARCH=arm go build` I can run it as expected. The binary from the packages does not work and exists with the error: `/lib/ld-linux.so.3: No such file or directory`. Will pick it up later again.

@andrewkroh I changed the packaging that the experimental packages are only build when `SNAPSHOT=yes` means they are not built for the releases.

ruflin:
I tested this again with the `armhf/debian` docker container. If I build it directly on my mac machine with `GOOS=linux GOARCH=arm go build` it works without changes. If I build it with `make package` I have to run `ln -s /lib/arm-linux-gnueabihf/ld-linux.so.3 /lib/ld-linux.so.3` on the container first. 

Any hints welcome.

ruflin:
@watson Did some testing of this on a real PI instead of the docker image and it seems there the missing file exists + symlink in the right place.

For moving this forward I was thinking if we should merge it in to get more testing on the produced packages? The ARM assets should only be created for the snapshot builds based on the new introduced flag.

ruflin:
PR rebased, PR message and commit message updated.

libbeat/scripts/Makefile
```diff
@@ -95,6 +96,10 @@ ifeq ($(RACE_DETECTOR),1)
 	RACE=-race
 endif
 
+# Only build experimental targets for snapshots
+ifeq ($(SNAPSHOT),no)
```
andrewkroh: All of the other checks against this value look for `yes`. So I think it would be better to do an "ifneq" to "yes" check.

ruflin:
This only creates the binary packages, no deb or rpm.

PR #6478, opened 2018-02-27
Added hsnburrowbeat to Community Beats

hsngerami:
Features:

Updated to Beats 6.2.1
Updated to Golang 1.9.4
Burrow V1.0.0(API V3) Support
Docker Support


karmi:
Hi @hsngerami, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/beats/pull/6478.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

hsngerami:
@karmi Done

ruflin:
@hsngerami No need from your side to do the backport as @dedemorton does frequent bulk updates to the list. Also it there went something wrong during the cherry picking. If ok for you I suggest to close this PR.

hsngerami:
ok. thanks

Issue #6472, opened 2018-02-26
Support arrays in in http/json metricset


ruflin:
Currently the http/json metricset in Metricbeat only can query information from http endpoints which expose map[string]interface{}. For endpoints which expose and array on the root level the json metricset does not work.

It's a common problem to have such endpoints. Either the http/json metricset should support such endpoints or a new http/json_list metricset should be added.

If a list is retrieved from an endpoint, each entry should be sent as an event from the metricset.

We discussed previously to autodetect if it's an array or not. To get started I prefer to have either a separate metricset or a config option like `json.list: true`. Having a config option has the advantage that auto detection could be added.

ruflin:
For reference, here is the initial issue: https://github.com/elastic/beats/issues/6027

jaipradeesh:
I was thinking to do the following way ..

```go
	switch body := jsonBody.(type) {
	case []map[string]interface{}:
            for _, obj := range {
                // Prepare `event` and return
            }
	case map[string]interface{}:
		// Prepare `event` and return
	default:
		// Log error
	}
```
.. but to match the `EventFetcher` (or) `EventsFetcher` interface, it would be better to go with `json_list` metricset.

```go
// EventFetcher is a MetricSet that returns a single event when collecting data.
// Use ReportingMetricSet for new MetricSet implementations.
type EventFetcher interface {
	MetricSet
	Fetch() (common.MapStr, error)
}

// EventsFetcher is a MetricSet that returns a multiple events when collecting
// data. Use ReportingMetricSet for new MetricSet implementations.
type EventsFetcher interface {
	MetricSet
	Fetch() ([]common.MapStr, error)
}
```

ruflin:
I'm more going in the direction of not adding a metricset for this an always use the second interface even if there is just one event. If it's just a single event, it has to be convered to []common.MapStr

The implementation you suggest above looks a lot like auto detection. I would prefer not to do that and use a config option instead. So if someone configures array but a non array json object is returned, an error is logged. We can still later add a config that enables auto detection.

christiangalsterer:
As commented in #6027 I can look into this in the next days.
W.r.t. to the implementation I would change to EventsFetcher interface but have a preference for auto-detection of arrays as this is more user-friendly and requires less configuration, but we can also add it later if this is currently the preferred way.

christiangalsterer:
Just saw that there is #6480 already on the way.. so will have a look into that one.

ruflin:
[...]

PR #6480, opened 2018-02-27
Metricbeat/HTTP: Support array in http/json metricset

jaipradeesh:
Currently (before this commit) the http/json metricset in Metricbeat
only can query information from http endpoints which expose
map[string]interface{}. For endpoints which expose an array on the
root level, the json metricset does not work.

A config option is added `response.isarray | bool`. If
someone configures array but a non array json object
is returned, an error is logged.

Event response is unified to []map[string]interface{} even
if the response is map[string]interface{}.

Ref: https://github.com/elastic/beats/issues/6472

### DIFF
--- a/metricbeat/docs/modules/http.asciidoc
+++ b/metricbeat/docs/modules/http.asciidoc
@@ -35,6 +35,7 @@ metricbeat.modules:
   #method: "GET"
   #request.enabled: false
   #response.enabled: false
+  #json.is_array: false
   #dedot.enabled: false
 
 - module: http
--- a/metricbeat/module/http/_meta/test/main.go
+++ b/metricbeat/module/http/_meta/test/main.go
@@ -7,13 +7,20 @@ import (
 )
 
 func main() {
-	http.HandleFunc("/", serve)
+	http.HandleFunc("/jsonarr", serveJSONArr)
+	http.HandleFunc("/jsonobj", serveJSONObj)
+	http.HandleFunc("/", serveJSONObj)
+
 	err := http.ListenAndServe(":8080", nil)
 	if err != nil {
 		log.Fatal("ListenAndServe: ", err)
 	}
 }
 
-func serve(w http.ResponseWriter, r *http.Request) {
+func serveJSONArr(w http.ResponseWriter, r *http.Request) {
+	fmt.Fprint(w, `[{"hello1":"world1"}, {"hello2": "world2"}]`)
+}
+
+func serveJSONObj(w http.ResponseWriter, r *http.Request) {
 	fmt.Fprint(w, `{"hello":"world"}`)
 }
--- a/metricbeat/module/http/json/_meta/data.json
+++ b/metricbeat/module/http/json/_meta/data.json
@@ -5,15 +5,14 @@
         "name": "host.example.com"
     },
     "http": {
-        "testnamespace": {
+        "json": {
             "hello": "world"
         }
     },
     "metricset": {
-        "host": "http:8080",
+        "host": "127.0.0.1:8080",
         "module": "http",
         "name": "json",
-        "namespace": "testnamespace",
         "rtt": 115
     }
 }
--- a/metricbeat/module/http/json/json.go
+++ b/metricbeat/module/http/json/json.go
@@ -49,6 +49,7 @@ type MetricSet struct {
 	body            string
 	requestEnabled  bool
 	responseEnabled bool
+	jsonIsArray     bool
 	deDotEnabled    bool
 }
 
@@ -63,12 +64,14 @@ func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
 		Body            string `config:"body"`
 		RequestEnabled  bool   `config:"request.enabled"`
 		ResponseEnabled bool   `config:"response.enabled"`
+		JSONIsArray     bool   `config:"json.is_array"`
 		DeDotEnabled    bool   `config:"dedot.enabled"`
 	}{
 		Method:          "GET",
 		Body:            "",
 		RequestEnabled:  false,
 		ResponseEnabled: false,
+		JSONIsArray:     false,
 		DeDotEnabled:    false,
 	}
 
@@ -91,37 +94,18 @@ func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
 		http:            http,
 		requestEnabled:  config.RequestEnabled,
 		responseEnabled: config.ResponseEnabled,
+		jsonIsArray:     config.JSONIsArray,
 		deDotEnabled:    config.DeDotEnabled,
 	}, nil
 }
 
-// Fetch methods implements the data gathering and data conversion to the right format
-// It returns the event which is then forward to the output. In case of an error, a
-// descriptive error must be returned.
-func (m *MetricSet) Fetch() (common.MapStr, error) {
-	response, err := m.http.FetchResponse()
-	if err != nil {
-		return nil, err
-	}
-	defer response.Body.Close()
-
-	var jsonBody map[string]interface{}
-	var event map[string]interface{}
-
-	body, err := ioutil.ReadAll(response.Body)
-	if err != nil {
-		return nil, err
-	}
-
-	err = json.Unmarshal(body, &jsonBody)
-	if err != nil {
-		return nil, err
-	}
+func (m *MetricSet) processBody(response *http.Response, jsonBody interface{}) common.MapStr {
+	var event common.MapStr
 
 	if m.deDotEnabled {
-		event = common.DeDotJSON(jsonBody).(map[string]interface{})
+		event = common.DeDotJSON(jsonBody).(common.MapStr)
 	} else {
-		event = jsonBody
+		event = jsonBody.(common.MapStr)
 	}
 
 	if m.requestEnabled {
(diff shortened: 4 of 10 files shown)


jaipradeesh:
@ruflin WIP. Yet to be tested thoroughly. Thoughts on the approach?

metricbeat/docs/modules/http.asciidoc
```diff
@@ -38,6 +38,7 @@ metricbeat.modules:
   #request.enabled: false
   #response.enabled: false
   #dedot.enabled: false
+  #response.isarray: false
```
ruflin: I'm thinking if we should make this `json.is_array` or `json.response.is_array`. Now that we have multiple metricsets in the http module not all settings apply to all metricsets I think.

I'm aware that previous configs do not follow the convention :-(

metricbeat/module/http/json/_meta/data.json
```diff
@@ -1,4 +1,4 @@
-{
+[{
```
ruflin: This part is not going to change. In case of an array, each document will be sent as a single vent in the previous format.

metricbeat/module/http/json/json.go
```diff
@@ -150,6 +135,48 @@ func (m *MetricSet) Fetch() (common.MapStr, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	err = json.Unmarshal(body, &jsonBody)
+	if err != nil {
+		return nil, err
+	}
+
+	switch body := jsonBody.(type) {
```
ruflin: Let's separate the processing and not only do a "dynamic" switch to figure out if there is an error. The way I imagine is that we have jsonBody and jsonBodyArray (or similar) and directly decode in what we want in the end. Like this I would assume we don't have to do any type checks and directly get the errors.

ruflin:
It would be great if for the tests we could spin up a small mock http server in Golang as some other modules do. And then have some example json files with both variants to test against.

metricbeat/module/http/json/json.go
```diff
@@ -150,6 +135,48 @@ func (m *MetricSet) Fetch() (common.MapStr, error) {
...
+// descriptive error must be returned.
+func (m *MetricSet) Fetch() ([]common.MapStr, error) {
+	response, err := m.http.FetchResponse()
+	if err != nil {
+		return nil, err
+	}
+	defer response.Body.Close()
+
+	var jsonBody interface{}
+	var event []common.MapStr
```
christiangalsterer: `event` should be named `events` as this now an array of events

ruflin: @christiangalsterer glad you stumbled over the PR.

metricbeat/module/http/_meta/test/main.go
```diff
@@ -7,13 +7,20 @@ import (
...
+	http.HandleFunc("/", serveJsonObj)
+
 	err := http.ListenAndServe(":8080", nil)
 	if err != nil {
 		log.Fatal("ListenAndServe: ", err)
 	}
 }
 
-func serve(w http.ResponseWriter, r *http.Request) {
+func serveJsonArr(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
jaipradeesh: Ack

metricbeat/module/http/_meta/test/main.go
```diff
@@ -7,13 +7,20 @@ import (
...
+	fmt.Fprint(w, `[{"hello1":"world1"}, {"hello2": "world2"}]`)
+}
+
+func serveJsonObj(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
jaipradeesh: Ack

jaipradeesh:
There is some issue with the config picking in tests. Looking into it.

jaipradeesh:
https://github.com/jaipradeesh/beats/blob/dad33c5a941924b346e8c27dcf6d054567315a81/metricbeat/module/http/json/json_integration_test.go#L26

This test would fail if `response.is_array` is false. Is there a way to pass (or) modify metricset configuration when running the tests?

@ruflin @christiangalsterer

jaipradeesh:
```
➜  metricbeat git:(metricbeat-http-json-arrays) go test --tags=integration github.com/elastic/beats/metricbeat/module/http/json/... -v                   

=== RUN   TestFetch
--- PASS: TestFetch (1.14s)
	json_integration_test.go:24: http/json event: [{"_namespace":"testnamespace","hello":"world"}]
	json_integration_test.go:32: http/json event: [{"_namespace":"testnamespace","hello1":"world1"} {"_namespace":"testnamespace","hello2":"world2"}]
=== RUN   TestData
--- SKIP: TestData (1.18s)
	data_generator.go:43: skip data generation tests
PASS
ok  	github.com/elastic/beats/metricbeat/module/http/json	2.330s
```

@ruflin This PR LGTM. Let me know if any changes are required.

metricbeat/module/http/json/json.go
```diff
@@ -93,37 +96,18 @@ func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
...
-	}
+func (m *MetricSet) processBody(response *http.Response, jsonBody interface{}) common.MapStr {
+	var event common.MapStr
 
 	if m.deDotEnabled {
-		event = common.DeDotJSON(jsonBody).(map[string]interface{})
+		event = common.DeDotJSON(jsonBody).(common.MapStr)
 	} else {
-		event = jsonBody
+		event = jsonBody.(common.MapStr)
```
ruflin: We probably should do some type conversion checks on both lines?

metricbeat/module/http/json/json_integration_test.go
```diff
@@ -15,32 +15,53 @@ import (
...
-	f := mbtest.NewEventFetcher(t, getConfig())
+	f := mbtest.NewEventsFetcher(t, getConfig("object"))
 	event, err := f.Fetch()
 	if !assert.NoError(t, err) {
 		t.FailNow()
 	}
 
 	t.Logf("%s/%s event: %+v", f.Module().Name(), f.Name(), event)
+
+	f = mbtest.NewEventsFetcher(t, getConfig("array"))
```
ruflin: Can we split this up in 2 tests? Like TestFetchObject and TestFetchArray? This will make it eaiser to spot what failed if something fails

metricbeat/module/http/json/json_integration_test.go
```diff
@@ -15,32 +15,53 @@ import (
...
+	f := mbtest.NewEventsFetcher(t, getConfig("object"))
+	err := mbtest.WriteEvents(f, t)
 	if err != nil {
 		t.Fatal("write", err)
 	}
+
 }
 
-func getConfig() map[string]interface{} {
+func getConfig(jsonType string) map[string]interface{} {
```
ruflin: you could make this a bool called `isArray`. But that also works. Do you expect other types here in the future?

metricbeat/docs/modules/http.asciidoc
```diff
@@ -37,6 +37,7 @@ metricbeat.modules:
   #method: "GET"
   #request.enabled: false
   #response.enabled: false
+  #response.is_array: false
```
ruflin: I was thinking again about the naming and I think we should go here with `json.is_array`. It's the json metricset so a prefix is needed, and as it's always about processing of the body / response anyways, we don't need this in.

metricbeat/module/http/json/json_integration_test.go
jaipradeesh: If confirming to JSON standards, yes. A string, number, object, array, true, false and null could be valid JSON. I would say to leave it this way than adding a boolean between array or not.

metricbeat/docs/modules/http.asciidoc
jaipradeesh: `json.is_array` makes sense to me. Will amend the change.

metricbeat/module/http/json/json.go
jaipradeesh: Sorry, I don't understand what you mean by 'both lines'. Elaborate please?

jaipradeesh:
Build failed due to fetch (thus build) error.

metricbeat/module/http/json/json.go
ruflin: Sorry, I meant on line 106 and 108 as both to a type converstion to `common.MapStr`.

metricbeat/module/http/json/json.go
```diff
@@ -148,7 +132,49 @@ func (m *MetricSet) Fetch() (common.MapStr, error) {
...
+// It returns the event which is then forward to the output. In case of an error, a
+// descriptive error must be returned.
+func (m *MetricSet) Fetch() ([]common.MapStr, error) {
+	response, err := m.http.FetchResponse()
+	if err != nil {
+		return nil, err
+	}
+	defer response.Body.Close()
+
+	var jsonBody common.MapStr
```
ruflin: As `jsonBody` and `jsonBodyArray` are "local" to the if / else parts, these could be defined inside the if clause.

metricbeat/module/http/json/json_integration_test.go
ruflin: Interesting point about the other types. I wonder when a request for that will pop up :-) Lets go with your version.

metricbeat/module/http/json/json.go
jaipradeesh: Did assertion because `common.DeDotJSON` accepts and emits `interface{}`. Type conversion would fail.

If types are mismatched, line 160 and 170 should error out.

jaipradeesh:
Sorry about the merge commit. Please squash before merging into `master` :)

metricbeat/module/http/json/json.go
ruflin: One more miscommunication on my end. I meant type assertions. I was looking for:

```
event, ok = jsonBody.(common.MapStr)
if !ok { ... }
```

If for whatever reason the assertion does not work, we would have a panic otherwise. At the same time I see it should not happen here.

ruflin:
@jaipradeesh This seems to conflict with the master branch. Could you rebase on master and push again?

I'm good with the changes. @christiangalsterer could you also have a look again?

ruflin:
jenkins, test it

PR #5945, opened 2017-12-21
Enable system tests for auditbeat

ruflin:
* There were system tests and system integration tests for auditbeat but so far they were not run.
* Integration tests enabled
* Module config changed to module that works on more systems. Limitation for linux removed.

This is prerequisite for https://github.com/elastic/beats/pull/5938

PR #6481, opened 2018-02-27
Cleanup Metricbeat composer and Makefile

ruflin:
This PR is a subset of https://github.com/elastic/beats/pull/5945 to remove the Metricbeat changes from the PR.

### DIFF
--- a/metricbeat/docker-compose.yml
+++ b/metricbeat/docker-compose.yml
@@ -1,12 +1,12 @@
-version: '2'
+version: '2.1'
 services:
   beat:
-    build: .
+    build: ${PWD}/.
     environment:
       - TEST_ENVIRONMENT=false
     working_dir: /go/src/github.com/elastic/beats/metricbeat
     volumes:
-      - ./..:/go/src/github.com/elastic/beats/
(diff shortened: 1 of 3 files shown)


andrewkroh:
I’m seeing this in the logs.

> time="2018-02-27T07:33:04Z" level=error msg="Could not parse config for project metricbeatsnapshotnoxpack700alpha10ecbe845b0c8f8f3d4ee4bf1ed7fdc9f337ebd4b : yaml: unmarshal errors:\n  line 1: cannot unmarshal !!str `2.1` into config.RawService"

metricbeat/docker-compose.yml
```diff
@@ -1,12 +1,12 @@
-version: '2'
```
ruflin: @exekias It seems the docker/libcompose does not support composer files of the version 2.1. Not sure if the project is still actively maintained. I reverted to version 2 as it does not make a difference for us at the moment I think.

exekias: I'm afraid that's the case, in the future this may be an issue, but for the way we use compose today it should be OK

Issue #3744, opened 2017-03-13
Metricbeat index date timezone


Exocomp:
Per https://discuss.elastic.co/t/metricbeat-index-date-timezone/78296.

How do you control the timezone metricbeat uses for the date part of the index name:

For ex:

metricbeat-2017.03.12

Is there a way I can tell metricbeat to use a particular timezone when generating the date part above. I'd like it to use UTC (which is what Filebeat and winlogbeat uses) but is using local.

The reason why this causes a problem is that filebeat and winlogbeat indices rollover to the next day (since it uses UTC) but metricbeat doesn't (since it uses local timezone) so they are not in sync and using 3rd party tools where you set the timezone for the index doesn't work because they are not all the same.

Thanks,
E

Elufimov:
I have the same problem and I think this is a bug.

ruflin:
I'm surprised that filebeat and metricbeat do not use the same indices as the same logic is used: https://github.com/elastic/beats/blob/master/libbeat/outputs/elasticsearch/client.go#L370. As the timestamp is used for the index name, somewhere in metricbeat the timestamp must be set differently. Which metricsets are you using? Can you share your config?

Exocomp:
@ruflin 

I found out the difference, where it is working filebeat and winlogbeat are going through logstash where I'm using the date filter and setting the timezone, metricbeat is going to elasticsearch directly.

ruflin:
Thanks for the update. Does that mean all implementations do not report UTC?

Exocomp:
@ruflin 

It is just for the date part of the index, as you say they all share same logic so if you send the other beats to elasticsearch directly I bet those index dates would be local also.  But I haven't tested the others going to elasticsearch directly just metricbeat.

urso:
checking the implementation, the '@timestamp' value is reported in UTC (timestamp is normalized before serializing), but when building the index or using the time format-string, the timestamp is still local time.

My proposal would be to always normalize the timestamp to UTC (so timestamps/events are comparable between different timezones) and report the `locale` in the event `beat` namespace. If required, one can use `local` and `@timestamp` to create the local timestamp in logstash or ES ingest node.

ruflin:
This processor is the first step to get the timezone into the event: https://github.com/elastic/beats/pull/3902

andig:
[...]

PR #6485, opened 2018-02-27
Use UTC time when formatting timestamps (#3744)

fjgal:
Solves issue #3744

Now the when the timestamp is stored in the context passed to the formatter is to converted to UTC. Some unit test (formatevents_test.go and datetime_test.go) had to be adapted to this behavior.

Modified GetIndex unit tests (elasticsearch/client_test.go) to expose this bug, now they use a timezone and a time that falls in between localtime and UTC to make sure the date in the index name is based on UTC.

### DIFF
--- a/libbeat/common/datetime_test.go
+++ b/libbeat/common/datetime_test.go
@@ -67,9 +67,9 @@ func TestTimeMarshal(t *testing.T) {
 	tests := []inputOutput{
 		{
 			Input: MapStr{
-				"@timestamp": Time(time.Date(2015, time.March, 01, 11, 19, 05, 112*1e6, time.UTC)),
+				"@timestamp": Time(time.Date(2015, time.March, 01, 11, 19, 05, 112*1e6, time.FixedZone("GMT-5", -3600*5))),
 			},
-			Output: `{"@timestamp":"2015-03-01T11:19:05.112Z"}`,
+			Output: `{"@timestamp":"2015-03-01T16:19:05.112Z"}`, // expect GMT
 		},
 		{
 			Input: MapStr{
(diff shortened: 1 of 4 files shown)


karmi:
Hi @fjgal, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/beats/pull/6485.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

fjgal:
Hi @karmi I have ammended the commit using the email address I have in my github account. Hope this is good now. Thanks for pointing this out!

ruflin:
jenkins, test it

ruflin:
@fjgal Thanks for taking a stab at this. CLA works now.

I need to have a closer look at the PR and all the parts this could affect to better understand if that is our way forward.

ruflin:
@urso @tsg Could you comment here? I remember at least one of you was thinking more about timezones etc.

urso:
As we store timestamps in UTC-0, this should be reflected in the index name creation as well..

I'm a little on the fence, though. This change potentially affects every use of format strings, forcing UTC-0 on every user.

The timestamp is inserted using the time format: `%{+yyyy.MM.dd}`. Alternatively we could modify/enhance the timestamp formatter to support UTC conversion internally. e.g. add timezone support to our data time formatter and change the default pattern to: `%{+yyyy.MM.ddZ0}`.

ruflin:
We should probably have a broader discussion on where we expect to use UTC and where we allow the user to have time zone specific time stamps.

In general I would expect that the events which happen at the same time whereever they are in the world end up in the same index. At the same time I think an event must be able to indicate from which time zone it was sent and converted. I wonder if `beat.timezone` is enough here?

jsoriano:
I think this is probably solved by the use of ILM in recent versions, when index timestamp doesn't depend on the content of the event.

Please reopen if this is causing issues in recent versions.

PR #6490, opened 2018-03-01
Use safe put for k8s metadata

exekias:
This makes use of https://github.com/elastic/beats/pull/6434 as an alternative to avoid mapping issues while keeping dots in metadata.

### DIFF
--- a/metricbeat/module/kubernetes/event/event.go
+++ b/metricbeat/module/kubernetes/event/event.go
@@ -7,6 +7,7 @@ import (
 	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/common/cfgwarn"
 	"github.com/elastic/beats/libbeat/common/kubernetes"
+	"github.com/elastic/beats/libbeat/common/safemapstr"
 	"github.com/elastic/beats/metricbeat/mb"
 )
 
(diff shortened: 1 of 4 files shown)


exekias:
jenkins, retest it please

metricbeat/module/kubernetes/event/event.go
```diff
@@ -114,7 +115,7 @@ func generateMapStrFromEvent(eve *kubernetes.Event) common.MapStr {
 	if len(eve.Metadata.Annotations) != 0 {
 		annotations := make(common.MapStr, len(eve.Metadata.Annotations))
 		for k, v := range eve.Metadata.Annotations {
-			annotations[common.DeDot(k)] = v
+			safemapstr.Put(annotations, k, v)
```
ruflin: I wonder if we should document this under breaking changes? As in case someone filtered for the annotations this queries now stop working. Will still merge the PR but we can discuss this as a follow up.

ruflin: Note: I agree this is a bug fix in general.

exekias: Yep, I wonder how we should communicate this, so far without this change the only way to get some field that is renamed by this was using `drop_field` processor to ensure there are no mapping issues.

ruflin: As we link to the PR in the changelog could you add more details to the PR message? So people ending up on this PR get some background info?

exekias: good point, done!

PR #6491, opened 2018-03-01
Remove unnecessary double-slashes in Windows script

adriansr:
The install-service.ps1 script is escaping backslashes inside a string but that is unnecessary in PowerShell, resulting in double-slashes inside paths for the service configuration.

This is only a cosmetic fix.


andrewkroh:
It’s a small change but it would be nice to have a note in the change log.

Issue #6273, opened 2018-02-04
Concept of Elastic linter


ewgRa:
During #5954 review @ruflin ask about changes like: please remove new line here, please use logp.Err instead of logp.Warn.

Looks like there is some coding style rules in Elastica, that can be introduced as a CI build process and make no sense to spend human time on this things.

I prepare https://github.com/elastic/beats/pull/6272 pull request as concept of linter, that can output for example something like that:
"test.go:20: Must be no newline before check error" for code like that: \n\nif err == nil ...

Can you please make a quick look and maybe say something about concept?
How you see it? Do you need it? Just to be know that I'm in right direction and it make sense to continue working on it.

From my side, I have something to discuss.

Is it linter, or formatter? There is good start https://github.com/golang/lint, "Golint differs from gofmt".

Must it check all files in build? Or just changed in this CI build? I think with checking all files there will be a problem that a lot of them already incompatible with some coding style rules.

Hope to hear feedback. Thanks

andrewkroh:
I would rather not take on the task of maintaining a linter for Go as part of this project. Perhaps it could be in it's own repo and this project uses it just like it does goimports.

My biggest concern about adding a tool like this to the build is false positives. Personally I'm pretty happy with the results we get from running golint and goimports on PRs (automated by Hound CI and Travis/Jenkins, respectively). But if we could define a set of supplemental linter rules for Go that do not produce a lot false positives then this could be helpful in reviews. I often run [gometalinter](https://github.com/alecthomas/gometalinter) on PRs.

My guess is that writing a linter is easier because you do not need to reformat the code. This gives some leeway w.r.t to false positives since users can just ignore the warning.

> Must it check all files in build? Or just changed in this CI build? I think with checking all files there will be a problem that a lot of them already incompatible with some coding style rules.

The tool itself should check the files as directed on the CLI (e.g. `mylinter ./...`). There are other tools that can handle producing warnings only for the changed files. We do this for `golint`'s output when `make lint` is executed (see our Makefile). It uses [reviewdog](https://github.com/haya14busa/reviewdog) to do this.

ewgRa:
> Perhaps it could be in it's own repo and this project uses it just like it does goimports.

I thought about that, just fear that some checks can be super custom and depends on Beats code, since I knew only about two of them. This two can be definitely implemented in own repo. Ok, if I have the same thoughts, than probably it must be another repo.

> My biggest concern about adding a tool like this to the build is false positives

[...]

PR #6496, opened 2018-03-05
Elastic linter

ewgRa:
PR for elastic coding style checks (#6273), now support check of - "no new line before error check" as only one well known and described.

Some demo of results, how it works with comments, that was created by CI: 
https://github.com/ewgRa/beats/pull/3
https://github.com/ewgRa/beats/pull/2

Some readme (not so friendly now, but more or less give explanations): https://github.com/ewgRa/gocsfixer/blob/master/README.md

For usage, there GOCSFIXER_GITHUB_TOKEN env must be set in travis repository settings, it is "Access Key" with public_repo rights.


ruflin:
Thanks for putting all the efforts into this. Two things I really like that this is a generic project which can be applied to any other project and that it can be run locally or on CI in the same way as far as I understand.

Do you already use the project in one of your go projects?

ewgRa:
@ruflin 

>  that it can be run locally or on CI in the same way

I think you know PHP-CS-Fixer project, gocsfixer - kind of the same, but for Go. This is main idea. Yes, it can be run on CI and locally for checking and fixing. I try to make it flexible as much as possible, to fit future requirements from other projects and companies, that can be interested in this tool.

> Do you already use the project in one of your go projects?

No. I have super small go projects (mostly pet projects) and there is no coding style conventions.

I think adopt it in beats first, have feedback, make changes if needed. After that introduce it to local go community as next step. At this step maybe introduce good commenting process for specific PR line, like "hound" doing it.

Why are you asking for? Chicken-egg problem? :)

ewgRa:
Inline comments from bot added, example: https://github.com/ewgRa/beats/pull/5#pullrequestreview-126981133

ruflin:
@ewgRa Thanks for keep pushing this. It's still on my list to review and think about but there were just too many more urgent priorities recently.

ruflin:
Recently @urso started to work on https://github.com/urso/beatsfmt because of our new lincese headers. I could see this projects overlapping.

ewgRa:
Looks like there is will be another linter, "cli" one. That for lint or fix will perform some cli commands. For example @urso case can be done by several unix commands. Also as "File header" linter can be achived with combination of "cat", "diff", "head" etc.

ruflin:
I think @urso also has in mind to make it more generic why I think long term the projects could be getting closer to each other.

One thing that I think is important for any "fixer" is that it's cross platform so it also work on Windows systems.

ewgRa:
Ok, I'm finished fixers, that I face during contribution to Beats. I think it is good milestone for next step.

I will polish a little bit documentation and within one-two month will prepare presentation for local golang meetup.

I would be happy if close to this time gocsfixer will be adopted in Beats, but I don't see what else I can do from my side. I'm ready to discuss it, answer on questions, help where I can.
If it will be adopted - it will be good use case for presentation. If not - nothing critical,
I think if gocsfixer can bring profit to developers - they will be interested anyway. If not, then it will be just experience for me :)

For now on gocsfixer support five fixers, mostly oriented on Beats cases:

1. Alternative call: can cover case "Instead of logp.Warn use logp.Err"

2. File header

3. Group standard imports, like:
```
import (
    "os"
    "github.com/foo/bar"
    "testing"
)

```
will be
```
import (
    "os"
    "testing"

    "github.com/foo/bar"
)
```

4. No new line before error check:
```
err = test()

if err == nil {
}
```

will be 

```
err = test()
if err == nil {
}
```
5. Use path.Join: `os.Readlink("foo/bar")` will be `os.Readlink(path.Join("foo", "bar"))`

ruflin:
Thanks for keeping pushing the project forward. The main issue on our end is time and priority. So far this didn't make it to the top of our priority list. We will get back to you when we find the time for it.

leehinman:
@ewgRa Thank you for all the hard work on the project but getting a linter integrated hasn't been a top priority.  If we decide to add a linter in the future we will definitely look at gocsfixer.

PR #6501, opened 2018-03-06
Do not build internal metrics without CGO

kvch:
Gosigar only requires CGO for Darwin, FreeBSD and OpenBSD. If someone compiles Beats on these platforms without CGO, metrics are not supported.

Add ephemeral ID to Beats which don't report internal memory usage, etc. So it can still show up in Monitoring UI.

### DIFF
--- a/libbeat/cmd/instance/metrics.go
+++ b/libbeat/cmd/instance/metrics.go
@@ -1,5 +1,4 @@
-// +build darwin linux windows
-// +build cgo
+// +build darwin,cgo freebsd,cgo linux windows
 
 package instance
 
@@ -7,38 +6,31 @@ import (
 	"fmt"
 	"os"
 	"runtime"
-	"time"
-
-	"github.com/satori/go.uuid"
 
 	"github.com/elastic/beats/libbeat/logp"
 	"github.com/elastic/beats/libbeat/metric/system/cpu"
 	"github.com/elastic/beats/libbeat/metric/system/process"
 	"github.com/elastic/beats/libbeat/monitoring"
-	"github.com/elastic/beats/libbeat/monitoring/report/log"
 )
 
 var (
 	beatProcessStats *process.Stats
-	ephemeralID      uuid.UUID
+	systemMetrics    *monitoring.Registry
 )
 
 func init() {
-	beatMetrics := monitoring.Default.NewRegistry("beat")
+	systemMetrics = monitoring.Default.NewRegistry("system")
+}
+
+func setupMetrics(name string) error {
 	monitoring.NewFunc(beatMetrics, "memstats", reportMemStats, monitoring.Report)
 	monitoring.NewFunc(beatMetrics, "cpu", reportBeatCPU, monitoring.Report)
-	monitoring.NewFunc(beatMetrics, "info", reportInfo, monitoring.Report)
 
-	systemMetrics := monitoring.Default.NewRegistry("system")
 	monitoring.NewFunc(systemMetrics, "cpu", reportSystemCPUUsage, monitoring.Report)
 	if runtime.GOOS != "windows" {
 		monitoring.NewFunc(systemMetrics, "load", reportSystemLoadAverage, monitoring.Report)
 	}
 
-	ephemeralID = uuid.NewV4()
-}
-
-func setupMetrics(name string) error {
 	beatProcessStats = &process.Stats{
 		Procs:        []string{name},
 		EnvWhitelist: nil,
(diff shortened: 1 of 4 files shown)


libbeat/cmd/instance/metrics_other.go
```diff
@@ -1,8 +1,20 @@
...
 package instance
 
-import "github.com/elastic/beats/libbeat/logp"
+import (
+	"github.com/satori/go.uuid"
+
+	"github.com/elastic/beats/libbeat/logp"
+)
+
+var (
```
ruflin: Could we move the ephemeralID part to a third file which gets compile on all platforms? Like this we can make sure we only have to change it in one place.

andrewkroh:
Metrics reporting should work for the Windows binaries where cgo isn't used. The build tags in metrics.go are

```
// +build darwin linux windows
// +build cgo
```
which is `(darwin OR linux OR windows) AND cgo`. But we need to have `(darwin AND cgo) OR (freebsd AND cgo) OR linux OR windows`. And as a build tag this would become

```
// +build darwin,cgo freebsd,cgo linux windows
```

Then over in `metrics_other.go` you want the build tag to be the exact logical opposite such that the stub is used for all other platforms.

andrewkroh:
I think we must have a test case to verify that xpack monitoring metrics are working correctly in the released binaries. I don't get a strong confidence from evaluating the build tags in my head against how I think they are set during the builds.

A good place to add this would be beats-tester. As you can see on [Jenkins](https://beats-ci.elastic.co/job/elastic+beats-tester+master/lastSuccessfulBuild/artifact/logs/) we have the log output from each Beat on each platform. The log contains a message like "Total non-zero metrics" that has all the monitoring. We can parse this JSON log message and assert that it contains all of the correct fields as well as the `ephemeral_id`.

The assertions need to be written twice, once for posix and once for Windows. The two places where it needs added are:

- https://github.com/elastic/beats-tester/blob/master/roles/test-beat/tasks/common/assert.yml
- https://github.com/elastic/beats-tester/blob/master/roles/test-beat/tasks/win32nt/assert.yml

As an example you can use [this](https://github.com/elastic/beats-tester/blob/master/roles/test-beat/tasks/packetbeat/assert.yml) which collects the line or interested into a variable, parse the data as JSON, then asserts that the object contains particular fields.

kvch:
I updated the branch with build tags I believe to be correct. :) I am still working on adding the assertions to Beats tester. @andrewkroh thanks for the pointers

ruflin:
I'm wondering if we should merge it as is and then run beats-tester to verify the change or have the beats-tester changes first? I'm good with both options.

ruflin:
@kvch I added the backport label as I assume the change is also needed for 6.2?

kvch:
Thanks for adding the tag. It's definitely needed in 6.2.

libbeat/cmd/instance/metrics_other.go
```diff
@@ -1,8 +1,11 @@
-// +build !darwin,!linux,!windows darwin,!cgo linux,!cgo windows,!cgo
+// +build darwin,!cgo freebsd,!cgo openbsd,!cgo
```
andrewkroh: Definitely double check my work, but I think we have a different set of built tags.

From metrics.go (without openbsd):
`(darwin AND cgo) OR (freebsd AND cgo) OR linux OR windows`

Negating this and simplifying it:
```
!((darwin AND cgo) OR (freebsd AND cgo) OR linux OR windows)
!(darwin AND cgo) AND !(freebsd AND !cgo) AND !linux AND !windows
(!darwin OR !cgo) AND (!freebsd OR !cgo) AND !linux AND !windows
```

Conversion to build tags:
```
// +build !darwin !cgo
// +build !freebsd !cgo
// +build !linux,!windows
```

kvch:
I opened a PR for Beats tester: https://github.com/elastic/beats-tester/pull/74

andrewkroh:
I kicked off a beats-tester job with the updated test cases. https://beats-ci.elastic.co/job/elastic+beats-tester+master/172/

We should see it fail until this PR gets merged. 👀

andrewkroh:
We found a secondary problem with the new tests we added. See https://github.com/elastic/beats-tester/pull/74#issuecomment-372464204.

ruflin:
Looking at the errors there also seem to be an error related to this change: 

```
17:08:22 TASK [test-beat : Check winlogbeat has monitoring metrics (win)] ***************
17:08:22 fatal: [tester-win12-64]: FAILED! => {"msg": "The conditional check 'log_metrics_event.monitoring.metrics.beat.cpu.system.ticks >= 0' failed. The error was: error while evaluating conditional (log_metrics_event.monitoring.metrics.beat.cpu.system.ticks >= 0): 'dict object' has no attribute 'beat'"}
```

kvch:
@ruflin The error you pasted here is expected. Right now on master the registry named "beat" is not added in case of winlogbeat or anything crosscompiled. This PR fixes that problem, because it adds the registry even if system and beat metrics are not supported on a platform.

I would merge this PR now. The other test failures are orthogonal to this PR as @andrewkroh pointed it out. So it should not be blocked.

libbeat/cmd/instance/metrics.go
```diff
@@ -1,44 +1,39 @@
...
 var (
 	beatProcessStats *process.Stats
-	ephemeralID      uuid.UUID
+	systemMetrics    *monitoring.Registry
 )
 
 func init() {
-	beatMetrics := monitoring.Default.NewRegistry("beat")
+	systemMetrics = monitoring.Default.NewRegistry("system")
+	ephemeralID = uuid.NewV4()
```
ruflin: The ephmeralID init is not needed here anymore I think.

ruflin:
Ok, left a minor comment. Could you also add a CHANGELOG entry?

kvch:
Nica catch! Thanks for the review @ruflin 
I added a CHANGELOG entry.

andrewkroh:
This should be back-ported incase we release a 6.2.4.

kvch:
I am doing that now. :D

Issue #6503, opened 2018-03-07
Infinite loop while watching for Kubernetes events


exekias:
Both when using `add_kubernetes_metadata` or `kubernetes` autodiscover provider, some users have reported this issue (which causes an infinite loop):

```
2018-03-05T08:16:36.021Z        ERROR   kubernetes/watcher.go:188       kubernetes: Watching API error EOF
2018-03-05T08:16:36.021Z        INFO    kubernetes/watcher.go:173       kubernetes: Watching API for resource events
2018-03-05T08:16:36.030Z        ERROR   kubernetes/watcher.go:188       kubernetes: Watching API error proto: wrong wireType = 6 for field ServiceAccountName
2018-03-05T08:16:36.030Z        INFO    kubernetes/watcher.go:173       kubernetes: Watching API for resource events
2018-03-05T08:16:36.031Z        ERROR   kubernetes/watcher.go:188       kubernetes: Watching API error proto: wrong wireType = 6 for field ServiceAccountName
2018-03-05T08:16:36.031Z        INFO    kubernetes/watcher.go:173       kubernetes: Watching API for resource events
2018-03-05T08:16:36.033Z        ERROR   kubernetes/watcher.go:188       kubernetes: Watching API error proto: wrong wireType = 6 for field ServiceAccountName
```

More details can be found here: https://github.com/elastic/beats/pull/6353

exekias:
I've pushed a branch ignoring unmarshall errors: https://github.com/exekias/beats/tree/ignore-k8s-errors

Also built an image from it: `exekias/filebeat:ignore-k8s-errors`, @simioa it would be awesome if you could do another test, your help here is much appreciated

simioa:
@exekias Looks good, no problems till now.

exekias:
Good to hear, @simioa, I'm wondering, did you see the ERROR happen?

The code doesn't really avoid that message but should recover from it in all cases:
`kubernetes: Watching API error proto: wrong wireType = 6 for field ServiceAccountName`

simioa:
Not that ERROR Message but this:
```
2018-03-08T15:33:50.114Z        ERROR   kubernetes/watcher.go:198       kubernetes: Watching API error EOF
```
I just looked into an Filebeat 6.2.2 Log and saw that right after the `kubernetes: Watching API error EOF` ERROR appeared, the infinite loop started.

This is not the case anymore, the only thing logged after the EOF Error related to the Watching API is:
```
[...]

PR #6504, opened 2018-03-07
Fix infinite failure on Kubernetes watch

vjsamuel:
This PR fixes https://github.com/elastic/beats/issues/6503

How to reproduce: Run filebeat pointing to minikube. 

```
minikube ssh
sudo su

ps aux | grep localkube
kill -9 process_id
```

This will force a failure on the API server, and when the API server comes back up it will not be able to serve up the last resource version that we had requested with the failure:
```
type:"ERROR" object:<raw:"k8s\000\n\014\n\002v1\022\006Status\022C\n\004\n\000\022\000\022\007Failure\032)too old resource version: 310742 (310895)\"\004Gone0\232\003\032\000\"\000" >  typeMeta:<apiVersion:"v1" kind:"Status" > raw:"\n\004\n\000\022\000\022\007Failure\032)too old resource version: 310742 (310895)\"\004Gone0\232\003" contentEncoding:"" contentType:""  <nil>
```

In such scenarios the only mitigation would be to move the resource version to the latest. Scenarios like this would be addressed by `client-go`. The reason why the code fails with error is because we pass a `Pod` resource to do the `watcher.Next()` in this scenario the resource that is attempted to be parsed is an `Error` resource and the protobuf unmarshalling fails. This is a limitation in the client that we use as the resource needs to be passed explicitly. 

This fix is not the best in the world as it might miss few state changes.

### DIFF
--- a/libbeat/common/kubernetes/watcher.go
+++ b/libbeat/common/kubernetes/watcher.go
@@ -3,6 +3,7 @@ package kubernetes
 import (
 	"context"
 	"fmt"
+	"io"
 	"time"
 
 	"github.com/ericchiang/k8s"
@@ -10,6 +11,9 @@ import (
 
 	"github.com/elastic/beats/libbeat/logp"
 )
+
+// Max back off time for retries
+const maxBackoff = 30 * time.Second
 
 func filterByNode(node string) k8s.Option {
 	return k8s.QueryParam("fieldSelector", "spec.nodeName="+node)
@@ -161,6 +165,9 @@ func (w *watcher) Start() error {
 }
 
 func (w *watcher) watch() {
+	// Failures counter, do exponential backoff on retries
+	var failures uint
+
 	for {
 		select {
 		case <-w.ctx.Done():
@@ -176,7 +183,8 @@ func (w *watcher) watch() {
 			//watch failures should be logged and gracefully failed over as metadata retrieval
 			//should never stop.
 			logp.Err("kubernetes: Watching API error %v", err)
-			time.Sleep(time.Second)
+			backoff(failures)
+			failures++
 			continue
 		}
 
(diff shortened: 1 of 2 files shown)


vjsamuel:
@exekias i stole the binary back off method from your attempt to fix this. Apologies for that :)

libbeat/common/kubernetes/watcher.go
```diff
@@ -186,8 +194,14 @@ func (w *watcher) watch() {
 			if err != nil {
 				logp.Err("kubernetes: Watching API error %v", err)
 				watcher.Close()
+				if !(err == io.EOF || err == io.ErrUnexpectedEOF) {
+					// This is an error event which can be recovered by moving to the latest resource verison
+					logp.Info("kubernetes: Ignoring event, moving to most recent resource version")
+					w.lastResourceVersion = ""
```
jsoriano: We may need to resync the stored objects, e.g. what would happen if we lose a delete event while we reconnect?

exekias: I've been doing some tests. At least when I reproduce this last resource version is increased by several hundred while no real change happened.

I agree with you @jsoriano, a resync may be needed to guarantee no events are missed during the reconnect.

Still, taking into account the severity of this issue, I'm ok with merging + backporting this as it is, as it improves the current situation (infinite loop + crash). I can do some more research and come up with a resync mechanism afterwards.

What do you think?

exekias:
jenkins, test this please

vjsamuel:
+1 the reconciliation might be a bigger change as we need to do some state management. we can unblock customers and fix that one in a subsequent PR.

vjsamuel:
@exekias based on this comment, it seems to be normal:
https://github.com/kubernetes/kubernetes/issues/55230#issuecomment-342632961

exekias:
Failing test is not related to this change

libbeat/common/kubernetes/watcher.go
vjsamuel: i think we are good to merge this one in.

PR #6508, opened 2018-03-07
update build image from debian 6 to 7

### LINKED ISSUE #6502: xgo-image-deb6 container image build fails
`make package` produces:

```
10:21:06 E: There are problems and -y was used without --force-yes
10:21:06 The command '/bin/sh -c apt-get -o Acquire::Check-Valid-Until=false update &&   apt-get install -y automake autogen build-essential ca-certificates     gcc-multilib     clang llvm-dev  libtool libxml2-dev uuid-dev libssl-dev pkg-config     patch make xz-utils cpio wget unzip git mercurial bzr rsync --no-install-recommends' returned a non-zero code: 100
10:21:06 make[1]: *** [xgo-image] Error 100
```

This happens because of a recent key expiration:
```
Step 9/22 : RUN   apt-get -o Acquire::Check-Valid-Until=false update &&   apt-get install -y automake autogen build-essential ca-certificates     gcc-multilib     clang llvm-dev  libtool libxml2-dev uuid-dev libssl-dev pkg-config     patch make xz-utils cpio wget unzip git mercurial bzr rsync --no-install-recommends
 ---> Running in d13be47f8090
Get:1 http://archive.debian.org squeeze Release.gpg [1655 B]
Get:2 http://archive.debian.org squeeze-lts Release.gpg [819 B]
Get:3 http://archive.debian.org squeeze Release [96.0 kB]
Ign http://archive.debian.org squeeze Release
Get:4 http://archive.debian.org squeeze-lts Release [34.3 kB]
Get:5 http://archive.debian.org squeeze/main amd64 Packages [8370 kB]
Get:6 http://archive.debian.org squeeze/contrib amd64 Packages [61.7 kB]
Get:7 http://archive.debian.org squeeze-lts/main amd64 Packages [390 kB]
Fetched 8955 kB in 16s (533 kB/s)
Reading package lists...
W: GPG error: http://archive.debian.org squeeze Release: The following signatures were invalid: KEYEXPIRED 1520281423 KEYEXPIRED 1501892461
```

specifically:
```
$ docker run -it --rm debian:6 apt-key list | grep -B1 squeeze
pub   4096R/473041FA 2010-08-27 [expired: 2018-03-05]
uid                  Debian Archive Automatic Signing Key (6.0/squeeze) <ftpmaster@debian.org>
```

graphaelli:
debian 6 has reached EOL and no longer has a valid package signing key, making apt-get installs unsafe.

fixes #6502


libbeat/scripts/Makefile
```diff
@@ -440,7 +440,7 @@ prepare-package-cgo:
 		-e BEAT_NAME=${BEAT_NAME} \
 		${BEATS_BUILDER_IMAGE}
 
-	# linux builds on debian 6 for compatibility
+	# linux builds on older debian for compatibility
```
ruflin: I remember we had debian 6 here because for some specific compatibility which was not the case with newer debian versions. Does this change mean we stop supporting Debian 6? And do we still need the old debian builds? @tsg might now more here.

tsg:
To give a bit of history information:

As we dynamically link against libc, the Beats work on any distribution with a newer libc than the one we package on, but not older distributions. From the distributions we support, debian 6 had the oldest libc.

If we do this, we should make sure that 1. we agree to drop support for deb6 and 2. debian 7 libc is the oldest libc from the ones we support. 

The other thing is that the reason we have two images for packing is that debian 6 was too old to make the cross-compilation to windows/macOS work. If we switch to debian 7, it might be that we can switch back to a single packer image.

PR #6510, opened 2018-03-08
Describe accurate default val in libbeat reference.yml.

simitt:
Default val for `queue.mem.flush.min_events` is 2048 instead of 0.

### DIFF
--- a/libbeat/docs/queueconfig.asciidoc
+++ b/libbeat/docs/queueconfig.asciidoc
@@ -28,7 +28,7 @@ If no flush interval and no number of events to flush is configured,
 all events published to this queue will be directly consumed by the outputs.
 To enforce spooling in the queue, set the `flush.min_events` and `flush.timeout` options.
 
-By default `flush.min.events` is set to 1024 and `flush.timeout` is set to 1s.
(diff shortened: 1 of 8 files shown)


urso:
Can you also check the docs? Might also say default value is 0.

libbeat/docs/queueconfig.asciidoc
```diff
@@ -22,16 +22,20 @@ queue.mem:
...
 The memory queue keeps all events in memory. It is the only queue type
-supported right now.  By default no flush interval is configured. All events
-published to this queue will be directly consumed by the outputs.
+supported right now. 
+
+If no flush interval and no number of events to flush is configured, 
+all events published to this queue will be directly consumed by the outputs.
+To enforce spooling in the queue, set the `flush.min_events` and `flush.timeout` options.
+
+By default `flush.min.events` is set to 1024 and `flush.timeout` is set to 1s.
```
urso: I like the new paragraph shortly explaining implications of these settings. Only: `s/flush.min.events/flush.min_events` , `s/1024/2048/`.

ruflin:
@simitt You probably have to run `make update` in the beats directory to make CI happy.

Issue #6473, opened 2018-02-26
Support for munin module in Metricbeat


medcl:
Munin: http://munin-monitoring.org/

> Munin is a networked resource monitoring tool that can help analyze resource trends and "what just happened to kill our performance?" problems. It is designed to be very plug and play. A default installation provides a lot of graphs with almost no work.

Munin is a widely used monitoring solution, and support a lot of platforms, such as OpenWRT, Solaris/SmartOS(which beats is not able to reach, for now, limited by Golang's implementation),

Also with a lot of metric collect plugins: 

- https://github.com/munin-monitoring/munin/tree/master/plugins
- https://github.com/munin-monitoring/contrib/tree/master/plugins

Munin can be deployed as Munin-node in the edge machine, listen on tcp port 9494, 
and a Munin server pulling these metrics via that agent's ip and port and generate reports, seems works like that, please correct me if I am wrong.

But Munin metrics analysis layer is kindly of out-dated, it uses a cron style to generate static html reports, neat but not enough, some modern analysis features are not there.

If we can leverage Munin-node to collect metrics, as they are pretty mature, and then use Metricbeat to collect these metrics and send to elasticsearch, now the world will rocks. 

@elastic/beats Any thoughts?

ruflin:
I like the idea. I would imagine it would work pretty similar to our http or prometheus module. In the best case metricbeat would be installed along the munin node agent but for not supported system it could be also remote.

jsoriano:
@medcl we have merged #6517 with a new experimental module for munin, this should be enough by now to collect metrics from a munin node agent.

I was wondering if we can also setup something for visualization.

Munin nodes also provide "[config](http://guide.munin-monitoring.org/en/latest/master/network-protocol.html#config)" about how each metric they expose should be represented, this info is used by munin masters to periodically generate the static reports. We could maybe reuse this config to generate kibana dashboards but I think it doesn't fit so well with our current setup workflow: munin metrics config is though to be retrieved during runtime, and `metricbeat setup` is intended to be run as a previous step. I guess that the ideal approach (from munin point of view) would be to have the ability to setup kibana dashboards dynamically during runtime.

Other options would be to have kibana dashboards for well-known munin plugins, or to have an only dashboard that is generated depending on the information in the index or the events (maybe vega can help on that?)

@mattapperson @ruflin thoughts about this?

tsg:
+1 for dashboards for the most popular munin plugins, or for what we don't support currently natively with Metricbeat.

[...]

PR #6517, opened 2018-03-09
Metricbeat: Add munin module

jsoriano:
Add module to export metrics from munin nodes (#6473). It can obtain metrics from any agent that implements the basic [protocol](http://guide.munin-monitoring.org/en/latest/master/network-protocol.html#network-protocol) for data exchange between master and nodes. Under the terms of this protocol, metricbeat would be acting as a master.


jsoriano:
jenkins, test it

metricbeat/module/munin/munin.go
```diff
@@ -0,0 +1,109 @@
...
+			value := scanner.Text()
+
+			key := fmt.Sprintf("%s.%s", item, name)
+
+			if value == unknownValue {
+				errs = append(errs, errors.Errorf("unknown value for %s", key))
+				continue
+			}
+			if f, err := strconv.ParseFloat(value, 64); err == nil {
+				event.Put(key, f)
```
exekias: for the record, this is a candidate for `safemapstr.Put`: https://github.com/elastic/beats/pull/6434 as soon as it's in

metricbeat/module/munin/munin.go
```diff
@@ -0,0 +1,109 @@
...
+				continue
+			}
+			event.Put(key, value)
```
exekias: `safemapstr.Put` too

metricbeat/module/munin/node/node.go
```diff
@@ -0,0 +1,60 @@
...
+
+// New creates a new instance of the MetricSet. New is responsible for unpacking
+// any MetricSet specific configuration options if there are any.
+func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
+	cfgwarn.Experimental("The munin node metricset is experimental.")
+
+	config := struct{}{}
+	if err := base.Module().UnpackConfig(&config); err != nil {
+		return nil, err
+	}
```
exekias: I guess this is not needed, as there is no custom config to parse?

metricbeat/module/munin/munin.go
```diff
@@ -0,0 +1,109 @@
...
+	if err != nil {
+		return nil, err
+	}
+	n := &Node{conn: conn,
+		writer: conn,
+		reader: bufio.NewReader(conn),
+	}
+	// Cosume and ignore first line returned by munin, it is a comment
+	// about the node
+	bufio.NewScanner(n.reader).Scan()
```
ruflin: Should we check for the return value here? In case of `false` we should probably check the `Err` method for the error.

metricbeat/module/munin/munin.go
```diff
@@ -0,0 +1,109 @@
...
+
+// List of items exposed by the node
+func (n *Node) List() ([]string, error) {
+	_, err := io.WriteString(n.writer, "list\n")
+	if err != nil {
+		return nil, err
+	}
+
+	scanner := bufio.NewScanner(n.reader)
+	scanner.Scan()
```
ruflin: Should we also check here the return value?

metricbeat/module/munin/munin.go
```diff
@@ -0,0 +1,109 @@
...
+	return strings.Fields(scanner.Text()), scanner.Err()
```
ruflin: I see you run here `scanner.Err()` which makes the return value check obsolete before.

metricbeat/module/munin/node/node.go
```diff
@@ -0,0 +1,60 @@
...
+
+	return &MetricSet{
+		BaseMetricSet: base,
+		timeout:       base.Module().Config().Timeout,
+	}, nil
+}
+
+// Fetch method implements the data gathering
+func (m *MetricSet) Fetch() (common.MapStr, error) {
```
ruflin: As if this metricset is used for different host, the returned data can contain very different structures. We should allow to set the namespace for this metricset as we do for example for the http/json metricset to make sure users can choose their namespace to prevent field conflicts.

metricbeat/module/munin/node/node.go
```diff
@@ -0,0 +1,60 @@
+package node
```
ruflin: You wrote above that metricbeat is the master in this scenario so I was tempted to say we should call the default namespace `master`. But as it's about the data from the node (is that naming correct?) +1 on node.

metricbeat/module/munin/munin.go
exekias: FYI, the PR went in already, so you can update these :)

jsoriano: Good point, but I think dots are not allowed in munin metric names.
We could use `safemapstr` in any case to protect metricbeat against incorrect implementations of munin nodes of plugins, wdyt?

metricbeat/module/munin/node/node.go
jsoriano: Actually there is no custom config so far, I'll remove this.

metricbeat/module/munin/munin.go
jsoriano: Yes, returning `scanner.Error()`.

metricbeat/module/munin/node/node.go
jsoriano: Protocol-wise we are like implementing the server side, but the metrics on the metricset are about the node, so +1 too to node :)

metricbeat/module/munin/node/node.go
jsoriano: Ok, adding namespace.

metricbeat/module/munin/node/node.go
jsoriano: I have finally kept it to support namespaces.

metricbeat/module/munin/munin.go
ruflin: I'm wondering if the float parsing will make a difference in ES. If we have a value 3 and we parse it it will still put 3 into the event and on the ES side it will assume it's an integer so conflicts will happen afterwards. Sounds like a good use case for the local fields definitions.

metricbeat/module/munin/node/_meta/docs.asciidoc
```diff
@@ -0,0 +1,34 @@
...
+and sends them as events to Elastic.
+
+[source,yaml]
+---
+- module: munin
+  metricsets: ["node"]
+  hosts: ["localhost:4949"]
+---
+
+All metrics exposed by a single munin node will be sent as an only event,
```
ruflin: "All metrics exposed by a single munin node will be sent in a single event, grouped by munin items." ?

An example here would be nice.

metricbeat/modules.d/munin.yml.disabled
```diff
@@ -0,0 +1,6 @@
+- module: munin
+  metricsets: ["node"]
+  enabled: false
+  period: 10s
+  hosts: ["localhost:4949"]
```
ruflin: The namespace config should be an option here.

metricbeat/module/munin/node/node.go
```diff
@@ -0,0 +1,75 @@
...
+	timeout   time.Duration
+}
+
+// New creates a new instance of the MetricSet. New is responsible for unpacking
+// any MetricSet specific configuration options if there are any.
+func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
+	cfgwarn.Experimental("The munin node metricset is experimental.")
+
+	config := struct {
+		Namespace string `config:"namespace" validate:"required"`
```
ruflin: I know this is not the case in the other implementations but we should prefix all configs with the metricset to make sure it's obvious to which metricset they belong. So this should be `node.namespace`.

metricbeat/module/munin/_meta/config.yml
```diff
@@ -0,0 +1,6 @@
+- module: munin
+  metricsets: ["node"]
+  enabled: false
+  period: 10s
+  hosts: ["localhost:4949"]
```
ruflin: Add namespace config.

metricbeat/docs/modules/munin.asciidoc
```diff
@@ -0,0 +1,41 @@
...
+in <<configuration-metricbeat>>. Here is an example configuration:
+
+[source,yaml]
+----
+metricbeat.modules:
+- module: munin
+  metricsets: ["node"]
+  enabled: false
+  period: 10s
+  hosts: ["localhost:4949"]
```
ruflin: Add namespace config

metricbeat/module/munin/munin.go
jsoriano: Umm, so what do you recommend to do here?

jsoriano:
Added namespace config where it was missing, and renamed it to `node.namespace`.

jsoriano:
I still have to take a look to the `data.json` generator.

metricbeat/module/munin/munin.go
exekias: I've checked metrics format, sounds like we don't need `safemapstr.Put` here

ruflin: I think what I'm getting at we could even skip this code part here and it would make no difference? If someone wants to make sure the mapping is correct he has to use https://github.com/elastic/beats/pull/6024

Please double check the above statement as I don't know much about the munin output.

ruflin:
@jsoriano I think for this metricset the `data.json` does not make much sense. We also skipped it for the other "dynamic" metricsets. Could you rebase again and squash your commits into one with a nice commit message?

@exekias Can you also check the PR again?

jsoriano:
@ruflin, after changing from `namespace` to `node.namespace` integration test is not working, metricbeat complains about missing required field. Actually in the generated `build/system-tests/last_run/metricbeat.yml` I couldn't see this setting.

I have seen in the base template that it expects the namespace in `namespace`. I have had to add it in the extras to make it work:
```
            "extras": {
                "node.namespace": namespace,
            },
```

Is this ok?

ruflin:
The extras is fine to solve the CI issue.

I'm going forth and back if namespace should be prefix by the metricset or not. The reason I like to have it prefix is that we do that will all metricset specific configs. Like this it becomes obvious to which metricset it belongs and does not accidentially overlap with other metricsets configs. Probably the reason we have it globally is that "dynamic" metricsets like this one should also be configured alone, meaning that the list of metricsets enable in a module should only contain this one. We don't enforce this yet but probably should. Like this an overlap cannot happen. Having the namespace config option on the module level could allow us to make it supported in a generic way and not have to implement it in each module.

@exekias @andrewkroh Any thoughts here?

ruflin:
I merged the PR to have the munin module in. The discussion about the naming of the `namespace` option should continue. As the module is not yet release and still experimental we can still make changes.

PR #6522, opened 2018-03-09
Update go-structform dependency

urso:
- go-structform has been transfered to github.com/elastic
- No changes/fixes affecting beats


andrewkroh:
Are there any bugfixes in v0.0.3 that are worthy of a Beats changelog entry?

urso:
No changes affecting beats in one or the other way. Only updated vendor, cause the package has been transferred.

Issue #6375, opened 2018-02-13
Update docs to reflect prospector > input refactoring


dedemorton:
Doc changes for https://github.com/elastic/beats/issues/6122.

dedemorton:
Closed by https://github.com/elastic/beats/pull/6537

Issue #6122, opened 2018-01-19
Meta: Refactor prospector to input


ph:
Track the ongoing task for refactoring the prospector to input.

- [x] Rename the prospector interface to the input interface https://github.com/elastic/beats/pull/6078
- [x] UDP prospector https://github.com/elastic/beats/pull/6118
- [x] Log prospector https://github.com/elastic/beats/pull/6120
- [x] stdin prospector https://github.com/elastic/beats/pull/6116
- [x] docker prospector https://github.com/elastic/beats/pull/6278
- [x] redis prospector https://github.com/elastic/beats/pull/6119
- [x] Convert modules yaml and json file to prospector and check for `event.type` https://github.com/elastic/beats/pull/6121
- [x] Update all the python test to not use any prospector words. https://github.com/elastic/beats/pull/6284
- [x] Update filebeat configuration yaml to remove any prospector words https://github.com/elastic/beats/pull/6290
- ~~update the documentation.~~ Move to #6375
- [x] create PR for changelog entries for all input.

ph:
@ruflin I've created this high level list, the PRs link in this issue are marked as `in progress` until we merge #6078, all of them target a feature branch.

ruflin:
Thx. Lets get #6078 in and from there, there is no way back 🎉

ph:
@ruflin I've rebased all the PRs for this feature and will let the CI run them.

ph:
@dedemorton concerning the point of updating the documentation to reflect the change from `prospector` to `input`, It will be a bit more than search and replace because we will have to redirect 302 some pages?

ph:
@ruflin I've created the last few PRs. 💃

ph:
Done! 🕺

PR #6537, opened 2018-03-13
Refactor input docs

dedemorton:
Resolves doc issue: https://github.com/elastic/beats/issues/6375

I still have a quite a few questions, which I've added as comments flagged for REVIEWERS. Please read and respond to questions.

Here's what the TOC will look like when the book is built:

image

Changes: 

- adds docs about UDP input for https://github.com/elastic/beats/pull/4452
- renames prospector > inputs for https://github.com/elastic/beats/pull/5944 and https://github.com/elastic/beats/issues/6122
- Also some light cleanup to use attribute for the Filebeat name (more work is required, but it's not urgent)

To do after merging this PR:

- [x] Create issue to track: The IIS and Nginx module descriptions in filebeat.reference.yml still refer to prospector. Should I fix? Created https://github.com/elastic/beats/issues/6642 
- [x] Create issue to track: The manifest for IIS and nginx still refer to prospector. Created https://github.com/elastic/beats/issues/6642 
- [x] Create issue to track: rework the topic about how filebeat works to be more generic and perhaps move file-specific into to the topic about the log input, if relevant. Created: https://github.com/elastic/beats/issues/6572


dedemorton:
cc @ruflin Didn't want to flag you as a reviewer, but just so you know this is up.

filebeat/docs/docker-input.asciidoc
```diff
@@ -0,0 +1,64 @@
+:type: docker
+
+[[docker-input]]
+=== Docker input
```
urso: Hmmm... docker input/prospector is available since 6.1. Only found it [mentioned here](https://www.elastic.co/guide/en/beats/filebeat/current/configuration-filebeat-options.html#_literal_type_literal).

As this is a new file, maybe we want to have an extra PR for it? 

Maybe we want to have a separate doc per input type? 

How about naming the file `input-docker.asciidoc`. This way, all input related docs will be displayed together in the file browser.

With inputs/modules being mostly self-contained, I wonder if we want to add the docs to the modules/inputs `_meta` path.

filebeat/docs/docker-input.asciidoc
```diff
@@ -0,0 +1,64 @@
...
+This input searches for container logs under its path, and parse them into
+common message lines, extracting timestamps too. Everything happens before line
+filtering, multiline, and JSON decoding, so this input can be used in
+combination with those settings.
+
+Example configuration:
+
+// REVIEWERS: The filebeat.reference.yml file does not show a docker input type
+// so I'm not sure if the following config is correct. docker only shows up under
+// autodiscover.
```
urso: @exekias can you have a look?

filebeat/docs/filebeat-options.asciidoc
```diff
@@ -1,544 +1,68 @@
...
-the key must be a string, otherwise no filtering or multiline aggregation will
-occur.
-
-[float]
-==== `multiline`
-
-Options that control how Filebeat deals with log messages that span multiple lines. See <<multiline-examples>> for more information about configuring multiline options.
+//REVIEWERS: I'm using the same example here that's used in the topic about the
+//log input. I wonder if we have a better one that shows a couple of different
+//input types?
```
urso: I would say it's good enough, as you're mostly setting generic configs. Maybe @exekias @ph have some more examples available?

filebeat/docs/filebeat-options.asciidoc
```diff
@@ -1,544 +1,68 @@
...
-When used with `type: udp`, specifies the maximum size of the message received over UDP. The default is 10240.
-
-[float]
-[[config-containers]]
-==== `containers`
+You can configure {beatname_uc} to use the following inputs:
 
-experimental[]
+// REVIWERS: I would like to make the following list generated, if possible. Can
+// we reuse the logic that we use for generating the modules list?
```
urso: This would be awesome.

filebeat/docs/filebeat-options.asciidoc
```diff
@@ -1,544 +1,68 @@
...
 
-These options are only available when using `docker` prospector type. They allow to configure the list of containers to read logs from.
+* <<log-input>>
+* <<stdin-input>>
+* <<redis-input>>
+* <<udp-input>>
+* <<docker-input>>
```
urso: How about renaming the anchors to `filebeat-input-<service>`. Also name the files `input-<service>.asciidoc`. So we have some grouping and reduce chance of naming conflicts.

filebeat/docs/how-filebeat-works.asciidoc
```diff
@@ -1,16 +1,25 @@
...
 
-In this topic, you learn about the key building blocks of Filebeat and how they work together. Understanding these concepts will help you make informed decisions about configuring Filebeat for specific use cases.
+// REVIWERS: I think we need to modify this topic to be more generic, or we
+// need to explain how different inputs handle the event data differently. It
+// doesn't seem accurate to use the word "file" everywhere as we do here. I
+// tried replacing file with log, but that doesn't work in all situations, so
+// I backed out my changes, and I'm looking for alternatives. Maybe we should
+// make this description more generic, and move the details about reading files
+// to the new topic about the log input. WDYT? I might do that in a separate
+// PR so we can get the refactoring merged.
```
urso: +1 on more generic description. As of now filebeat is about collecting logs only, but given as logs can be anything + we might widen the scope of filebeat in the future, `events` or `documents` might be more appropriate.

filebeat/docs/how-filebeat-works.asciidoc
```diff
@@ -1,16 +1,25 @@
...
 
-Filebeat consists of two main components: <<prospector,prospectors>> and <<harvester,harvesters>>. These components work together to tail files and send event data to the output that you specify.
+In this topic, you learn about the key building blocks of {beatname_uc} and how they work together. Understanding these concepts will help you make informed decisions about configuring {beatname_uc} for specific use cases.
+
+{beatname_uc} consists of two main components: <<input,inputs>> and <<harvester,harvesters>>. These components work together to tail files and send event data to the output that you specify.
```
urso: TBH I don't like this documentation file. The point is, the presence of a harvester and a particular harvester implementation is 'implementation specific'. A harvester is basically just a worker, allowing for concurrent collection of logs.

I understand the need of this doc, but in the end, we might have inputs and harvesters/workers document for each input type.

Different input types might have different strategies for collecting logs. E.g. for stdin there is no real need for an harvester.

We're also about entangling the harvesters, to have per input type worker implementations. The log harvester is overkill for some use-cases, but re-use 'complicates' the logic.

filebeat/docs/how-filebeat-works.asciidoc
```diff
@@ -21,60 +30,58 @@ Closing a harvester has the following consequences:
...
+{beatname_uc} keeps the state of each file and frequently flushes the state to disk in the registry file. The state is used to remember the last offset a harvester was reading from and to ensure all log lines are sent. If the output, such as Elasticsearch or Logstash, is not reachable, {beatname_uc} keeps track of the last lines sent and will continue reading the files as soon as the output becomes available again. While {beatname_uc} is running, the state information is also kept in memory for each input. When {beatname_uc} is restarted, data from the registry file is used to rebuild the state, and {beatname_uc} continues each harvester at the last known position.
```
urso: The 'flush' is the operation. One can change this to:
filebeat keeps the file states in memory and writes a snapshot to the registry file  whenever events are acknowledged by the outputs or when the registry flush timer is trigger. The flush timer is only active if some state has been changed.

filebeat/docs/overview.asciidoc
```diff
@@ -5,14 +5,28 @@
...
-and forwards them either to https://www.elastic.co/products/elasticsearch[Elasticsearch] or https://www.elastic.co/products/logstash[Logstash] for indexing.
+{beatname_uc} is a lightweight shipper for forwarding and centralizing log data.
+Installed as an agent on your servers, {beatname_uc} monitors the log
+files or locations that you specify, tails the logs, and forwards log data
+to either to https://www.elastic.co/products/elasticsearch[Elasticsearch] or
+https://www.elastic.co/products/logstash[Logstash] for indexing.
 
-Here's how Filebeat works: When you start Filebeat, it starts one or more prospectors that look in the local paths you've specified for log files. For each log file that the prospector locates, Filebeat starts a harvester. Each harvester reads a single log file for new content and sends the new log data to libbeat, which aggregates the events and sends the aggregated data to the output that you've configured for Filebeat.
+// REVIEWERS: Is it still accurate to say "tails the logs" or is that a file
+// concept that doesn't make sense for some of the newer inputs that we've added.
```
urso: kind of depends on the input type. In a 'pull' based input, we will try to 'tail' in a sense of trying to collect new logs only. We also have 'push' based inputs (e.g. syslog or plain UDP/TCP), this would make sense for.

How about 'collect'. Instead of `logs` we might have `events` or `documents`. With introduction of UDP/TCP, we already widened the scope.

filebeat/docs/redis-input.asciidoc
```diff
@@ -0,0 +1,93 @@
...
+- type: redis
+  hosts: ["localhost:6379"]
+  username: "$\{redis_user\}"
+  password: "$\{redis_pwd\}"
+----
+
+
+// REVIEWERS: Which options should I show in the example config? Incidentally,
+// in the reference.yml, password shows up twice in the same namespace. This
+// needs to be fixed in the reference.yml file
```
urso: we have a ticket for the option being shown twice?

filebeat/docs/redis-input.asciidoc
```diff
@@ -0,0 +1,93 @@
...
+
+==== Configuration options
+
+The `redis` input supports the following configuration options plus the
+<<{type}-input-common-options>> described later.
+
+//REVIEWERS: These descriptions are new, so please review carefully.
```
urso: 👍

filebeat/docs/redis-input.asciidoc
```diff
@@ -0,0 +1,93 @@
...
+
+[float]
+[[redis-scan_frequency]]
+===== `scan_frequency`
+
+How often {beatname_uc} reads entries from Redis slowlogs. Specify `1s` to scan
+Redis as frequently as possible without causing {beatname_uc} to scan too
+frequently. Do not set this value to less than `1s`.
+
+The default is `10s`.
```
urso: Maybe we want to add a warning here. Redis slow logs are not permanent. We have to connect and query logs. If redis did collect more slow logs then can be buffered within the interval of `scan_frequency`, logs will be missing.

filebeat/docs/redis-input.asciidoc
```diff
@@ -0,0 +1,93 @@
...
+How long to wait for a response from Redis before the input returns an error.
+The default is `1s`.
+
+[float]
+[[redis-network]]
+===== `network`
+
+The network type to use for the Redis connection. The default is `tcp`.
+
+//REVIEWERS: What (if any) other network types are supported here?
```
urso: Hmm... the beats code is not doing validating the network type. And the redis client use does no validation as well.

The stdlib docs say:

```
Known networks are "tcp", "tcp4" (IPv4-only), "tcp6" (IPv6-only), "udp", "udp4" (IPv4-only), "udp6" (IPv6-only), "ip", "ip4" (IPv4-only), "ip6" (IPv6-only), "unix", "unixgram" and "unixpacket".
```

The `ipX` settings do not make sense.

Redis requires `TCP` -> drop UDP settings.

While redis (theoretical) support unix sockets, I'm not sure about `"unixgram" and "unixpacket"`.

filebeat/docs/udp-input.asciidoc
```diff
@@ -0,0 +1,40 @@
...
+
+==== Configuration options
+
+The `udp` input supports the following configuration options plus the
+<<{type}-input-common-options>> described later.
+
+//REVIEWERS: What other options need to be documented here? I don't see host
+//and port listed in the reference.yml. Are the host and port separate options?
+//What guidance should we provide to help users set these options? Config ex
+//above should include this as an eample, too.
```
urso: @ph can you have a look?

filebeat/docs/stdin-input.asciidoc
```diff
@@ -0,0 +1,24 @@
...
+Example configuration:
+
+["source","yaml",subs="attributes"]
+----
+{beatname_lc}.inputs:
+- type: stdin
+----
+
+// REVIEWERS: Are there any stdin-specific config options, or are all the config
+// options inherited from libbeat?
```
urso: right now it supports all harvester settings (mostly the same as are available for `log` input type). Some settings just don't make sense for stdin.

filebeat/docs/docker-input.asciidoc
exekias: The example configuration is correct :+1:

filebeat/docs/docker-input.asciidoc
```diff
@@ -0,0 +1,64 @@
...
+
+Reads from the specified streams only: `all`, `stdout` or `stderr`. The default
+is `all`.
+
+// REVIEWERS: I'm guessing where these settings live in the config 
+// because it's not clear from the docs or the reference.yml. Would be nice to
+// have a realistic example here.
+
+// REVIEWERS: Sounds like I definitely need to add multiline to the docs here.
+// which other config options from the harvester are used by the docker input?
```
exekias: Docker input should support all the same settings as input, `paths` would be the only one we don't want to expose here, as internally it's defined by `containers.ids`

filebeat/docs/docker-input.asciidoc
```diff
@@ -0,0 +1,64 @@
...
+is `/var/lib/docker/containers`.
+
+===== `containers.stream`
+
+Reads from the specified streams only: `all`, `stdout` or `stderr`. The default
+is `all`.
+
+// REVIEWERS: I'm guessing where these settings live in the config 
+// because it's not clear from the docs or the reference.yml. Would be nice to
+// have a realistic example here.
```
exekias: A full example could be:
```
- type: docker
  containers:
    paths: "/var/lib/docker/containers"
    stream: "stdout"
    ids:
      - "*"
```
This config will read stdout stream from all containers under the default docker containers path

filebeat/docs/docker-input.asciidoc
dedemorton: @urso 

- re docker prospector for 6.1: Getting the prospector options broken into separate topics has been on my to-do list for awhile. I probably should have done a separate PR for the structural changes to make it easier to backport them. But since I didn't, I'll just have to create a separate PR for 6.1/6.2 that has relevant changes (if we decide to backport). First I want to make sure this gets into 6.3.
- This PR does add a separate doc for each input.
- I'll rename the files as you've suggested
- If we use _meta folders, I'd like to keep it simple and add an include rather than doing a copy to the docs dir. I've asked ph for his input.

filebeat/docs/how-filebeat-works.asciidoc
dedemorton: TBH, I think this whole topic needs a rewrite. I'd like to defer this to another PR. I've created an issue to track the work here: https://github.com/elastic/beats/issues/6572

filebeat/docs/how-filebeat-works.asciidoc
dedemorton: @urso I think you're proposing that I replace the entire line with your description, right?

filebeat/docs/redis-input.asciidoc
dedemorton: Created a ticket here: https://github.com/elastic/beats/issues/6573

filebeat/docs/stdin-input.asciidoc
dedemorton: @urso Can you look through the document I created and tell me which options make sense for stdin? I put a question mark for the options *I think* are applicable. https://docs.google.com/spreadsheets/d/1ixRNLA22LiBAgXGZzDBPxz-7kflaSn-JOQkSIcU-d78/edit#gid=0

filebeat/docs/filebeat-options.asciidoc
dedemorton: Hmm...I looked at the logic that we use to create the list of modules for the docs, and I don't like that it recurses the directory where the source files are stored. I'd rather have something that looks at the hierarchy defined by the asciidoc markup (rather than relying on the source files to be in a particular directory). 

Our doc tools really should support a macro or something for generating "jump" lists or tables. This is not a new concept. I could do it in FrameMaker 15 years ago. Rather than coming up with a solution just for Beats, I'm going to open a request in the docs repo.

dedemorton: I've created this issue to track the enhancement request: https://github.com/elastic/docs/issues/324

filebeat/docs/overview.asciidoc
dedemorton: I've made this description more generic.

filebeat/docs/inputs/input-docker.asciidoc
```diff
@@ -0,0 +1,75 @@
...
+The list of Docker container IDs to read logs from. Specify
+`containers.ids: '*'` to read from all containers.
+
+===== `containers.paths`
+
+The base path where Docker logs are located. The default
+is `/var/lib/docker/containers`.
+
+// REVIEWERS: Not clear to me if this setting accepts an array or a string. If
+// it's just a string, why is it paths (plural)?
```
ph: @dedemorton I believe you are right here, it should be singular.

https://github.com/elastic/beats/blob/master/filebeat/input/docker/config.go#L15-L21

Nothing in the code currently accept multiple paths for the docker containers.

@exekias WDYT to move it to singular?

filebeat/docs/inputs/input-udp.asciidoc
```diff
@@ -0,0 +1,42 @@
...
+
+==== Configuration options
+
+The `udp` input supports the following configuration options plus the
+<<{beatname_lc}-input-{type}-common-options>> described later.
+
+//REVIEWERS: PH: What other options need to be documented here? I don't see host
+//and port listed in the reference.yml. Are the host and port separate options?
+//What guidance should we provide to help users set these options? Config ex
+//above should include this as an eample, too.
```
ph: @dedemorton Its currently a [single option](https://github.com/elastic/beats/blob/master/filebeat/input/udp/config.go#L13). 

It should be:

```yaml
filebeat.inputs:
- type: udp
  max_message_size: 10240
  host: "localhost:8080"
```

The default value is "localhost:8080", To be honest, I think we should not provide a default for it and enforce user to set it, so making sure the example provide the host will help us move forward.

FYI, I would like to make it mandatory for 7.0

filebeat/docs/migration.asciidoc
```diff
@@ -96,7 +96,7 @@ section to understand the Filebeat options.
 [float]
 === Migrate the "files" section
 
-To migrate the `files` section from the Logstash Forwarder configuration, create a  `prospectors` section in the Filebeat config file. For example, assuming that you start
+To migrate the `files` section from the Logstash Forwarder configuration, create an  `inputs` section in the Filebeat config file. For example, assuming that you start
```
ph: Double space between `an` and `input`

filebeat/docs/migration.asciidoc
```diff
@@ -361,11 +361,11 @@ Forwarder.
...
-you can use the <<fields-under-root>> configuration option:
+you can use the <<fields-under-root-log>> configuration option:
 
 [source,yaml]
 -------------------------------------------------------------------------------------
-filebeat.prospectors:
+filebeat.inputs:
 - type: log
   paths:
     - /var/log/*.log
```
ph: Should we use the asciidoc variable for all the `filebeat` reference in this file? ({beatname_uc})

filebeat/docs/multiline.asciidoc
```diff
@@ -17,7 +17,7 @@ Also read <<yaml-tips>> and <<regexp-support>> to avoid common mistakes.
 [[multiline]]
 === Configuration options
 
-You can specify the following options in the `filebeat.prospectors` section of
+You can specify the following options in the `filebeat.inputs` section of
```
ph: Should we use asciidoc variable here? {beatname_uc}

filebeat/docs/overview.asciidoc
```diff
@@ -5,14 +5,25 @@
...
 
 image:./images/filebeat.png[Beats design]
 
-For more information about prospectors and harvesters, see <<how-filebeat-works>>.
+For more information about inputs and harvesters, see
+<<how-filebeat-works>>.
 
-Filebeat is a https://www.elastic.co/products/beats[Beat], and it is based on the libbeat framework.
-General information about libbeat and setting up Elasticsearch, Logstash, and Kibana are covered in the {libbeat}/index.html[Beats Platform Reference].
+{beatname_uc} is a https://www.elastic.co/prproducts/beats[Beat], and it is based on
```
ph: typo in the URL should be `https://www.elastic.co/products/beats`

filebeat/docs/overview.asciidoc
```diff
@@ -5,14 +5,25 @@
...
 
-Here's how Filebeat works: When you start Filebeat, it starts one or more prospectors that look in the local paths you've specified for log files. For each log file that the prospector locates, Filebeat starts a harvester. Each harvester reads a single log file for new content and sends the new log data to libbeat, which aggregates the events and sends the aggregated data to the output that you've configured for Filebeat.
+Here's how {beatname_uc} works: When you start {beatname_uc}, it starts one or
+more inputs that look in the locations you've specified for log data. For
+each log that {beatname_uc} locates, {beatname_uc} starts a harvester. Each
+harvester reads a single log for new content and sends the new log data to
+libbeat, which aggregates the events and sends the aggregated data to the output
+that you've configured for {beatname_uc}.
 
 image:./images/filebeat.png[Beats design]
```
ph: We have replaced `Filebeat` with the variable should we do the same with this path?

filebeat/docs/inputs/input-docker.asciidoc
exekias: Actually, the setting key is `path`, this looks like a typo in docs: https://github.com/elastic/beats/blob/master/filebeat/input/docker/config.go#L17

ph: Good catch @exekias

filebeat/docs/overview.asciidoc
dedemorton: My cat did that.

filebeat/docs/how-filebeat-works.asciidoc
dedemorton: Yah, I agree. I'm going to track this work in a separate issue: https://github.com/elastic/beats/issues/6572

filebeat/docs/migration.asciidoc
dedemorton: Yes...not urgent, but over I'm trying to replace references to product names with asciidoc variables whenever I open up a file

filebeat/docs/multiline.asciidoc
dedemorton: yup. I've fixed the references in this file, too. I'm doing these changes in small batches with other changes because it's not always a simple text replacement. Formatting affects how the attributes are interpreted.

filebeat/docs/docker-input.asciidoc
dedemorton: I confirmed that the docker options are covered in 6.1, just not ideally organized. Backporting this work to 6.2 would be a pretty big effort. The reorg isn't straightforward because it involves moving a lot of content around. I'll revisit the decision later if it makes sense, but for now, I'd like to keep this in 6.3 and not backport.

dedemorton:
@ph I think I've addressed all of your issues.

Here are a few things left to do/decide that, IMO, can wait until after this thing is merged. I've opened issues for all the other stuff I found.

- [ ] Do we have the correct options listed for all of the input types? https://docs.google.com/spreadsheets/d/1ixRNLA22LiBAgXGZzDBPxz-7kflaSn-JOQkSIcU-d78/edit#gid=0know: meta folders
- [ ] Do we need to have the source for the input docs in _meta files under the inputs? I'd rather not because it complicates everything (including the doc build). I vote to leave as-is for the initial merge. 
- [ ]  Todo: DeDe to set up redirects. We can't redirect links to anchors, but any existing links to the config options will point to the log input type.

ph:
I believe you have all the options, from what I've read.

Concerning the location of the input doc, I agree to keep them where they are currently. We can always move the doc in the `_meta` folder later if we wish to have module more self contained.

Issue #5981, opened 2018-01-04
[heartbeat] HTTP check.body enhancements proposal


Crazybus:
As discussed with @ruflin I would like to make some enhancements to the `check.body` option in heartbeat. Happy to write the implementation but first wanted to discuss how this should look configuration wise since that is the main challenge. 

## Current situation
`check.response.body` calls the function [checkBody](https://github.com/elastic/beats/blob/e8b8fa9bb857f74e424a07295a3550a7523e2b0d/heartbeat/monitors/active/http/check.go#L87) which compares the entire reponse body to make sure it is equal. In the `body: '{"status": "ok"}'` that works perfectly however if you have an entire webpage it isn't practical to paste the entire webpage into `check.response.body`
```
check.response:
  body: '{"status": "ok"}'
```

## Proposal check.response.body with contains
Modify `func checkBody` to check if the body contains the string rather than being equal

## Proposal check.response.body with regex
Make `check.response.body` a regex pattern for matching the body
```
check.response:
  body: /ok/
```

## Bonus proposal check.resonse.bodyJSON
Instead of checking for a JSON string actually parse the JSON and compare fields.
```
check.response:
  bodyJSON:
    status: "ok"
```

ruflin:
I like both suggestion. For option 1 we keep the existing functionality but also allow regexp expression in addition. We already use someting similar for `exclude_lines` in FB an other places:

```
	ExcludeLines []match.Matcher         `config:"exclude_lines"`
```

An other more powerful option could be to introduce somehow the query language we use in processor. Something like

```
check.response:
  when.equals:
    body: 123
```
This would allow to use the full power of the query language like conditions etc: https://www.elastic.co/guide/en/beats/filebeat/5.3/configuration-processors.html#conditions I wonder if in this case we should even write above `check.when.equals: repsonse.body: ...` to directly mention the full field. This would even make it possible to mix request / response and header / body etc together.

Taking the above into account, I'm thinking for the response but it should be possible to define that the response is in json directly as part of the monitor outside the check. Then the above query language could be used to access all the fields which would be put under the response namespace.

Crazybus:
@ruflin I really really like the option of using the query language. However I can imagine that would be rather difficult to implement and keep backwards compatible with current heartbeat configurations. 

Would you be open to me implementing option 1 for now and opening a separate issue for using the query language for checks?

ruflin:
+1 on moving forward as you suggested.

[...]

PR #6539, opened 2018-03-13
Use match.Matcher for checking heartbeat response bodies #5981

Crazybus:
The current behaviour of `check.response.body` is to match the entire response body. This works great for simple examples however doesn't work so great when you want to find a small string in a large body.

As discussed in #5981 I have modified the `check.response.body` parameter to expect a regex pattern (`match.Matcher`) instead of the entire body string to match. 

The awesome thing is this remained fully backwards compatible with the old body check method without any config changes even when using the json example in the docs :)

Another nice enhancement that I added was the use of [subtests](https://blog.golang.org/subtests). I originally copied some tests from another beat and was disappointed to find out that it would stop as soon as a test case failed and not run all of them. Subtests allows you to run all tests using the table driven pattern with the output and behaviour of writing multiple tests.

```
❯ go test github.com/elastic/beats/heartbeat/monitors/active/http -v
=== RUN   TestCheckBody
=== RUN   TestCheckBody/Single_regex_that_matches
=== RUN   TestCheckBody/Regex_matching_json_example
=== RUN   TestCheckBody/Regex_matching_first_line_of_multiline_body_string
=== RUN   TestCheckBody/Regex_matching_lastline_of_multiline_body_string
=== RUN   TestCheckBody/Regex_matching_multiple_lines_of_multiline_body_string
=== RUN   TestCheckBody/Regex_not_matching_multiple_lines_of_multiline_body_string
=== RUN   TestCheckBody/Single_regex_that_doesn't_match
=== RUN   TestCheckBody/Multiple_regex_match_where_at_least_one_must_match
=== RUN   TestCheckBody/Multiple_regex_match_where_none_of_the_patterns_match
--- PASS: TestCheckBody (0.00s)
    --- PASS: TestCheckBody/Single_regex_that_matches (0.00s)
    --- PASS: TestCheckBody/Regex_matching_json_example (0.00s)
    --- PASS: TestCheckBody/Regex_matching_first_line_of_multiline_body_string (0.00s)
    --- PASS: TestCheckBody/Regex_matching_lastline_of_multiline_body_string (0.00s)
    --- PASS: TestCheckBody/Regex_matching_multiple_lines_of_multiline_body_string (0.00s)
    --- PASS: TestCheckBody/Regex_not_matching_multiple_lines_of_multiline_body_string (0.00s)
    --- PASS: TestCheckBody/Single_regex_that_doesn't_match (0.00s)
    --- PASS: TestCheckBody/Multiple_regex_match_where_at_least_one_must_match (0.00s)
    --- PASS: TestCheckBody/Multiple_regex_match_where_none_of_the_patterns_match (0.00s)
=== RUN   TestSplitHostnamePort
--- PASS: TestSplitHostnamePort (0.00s)
PASS
ok  	github.com/elastic/beats/heartbeat/monitors/active/http	0.019s
```

### DIFF
--- a/heartbeat/docs/heartbeat-options.asciidoc
+++ b/heartbeat/docs/heartbeat-options.asciidoc
@@ -469,13 +469,13 @@ contains JSON:
 - type: http
   schedule: '@every 5s'
   urls: ["https://myhost:80"]
-check.request:
-  method: GET
-  headers:
-    'X-API-Key': '12345-mykey-67890'
-check.response:
-  status: 200
-  body: '{"status": "ok"}'
+  check.request:
+    method: GET
+    headers:
+      'X-API-Key': '12345-mykey-67890'
+  check.response:
+    status: 200
+    body: '{"status": "ok"}'
 -------------------------------------------------------------------------------
 
 The following configuration shows how to check the response for multiple regex
@@ -486,15 +486,15 @@ patterns:
 - type: http
   schedule: '@every 5s'
   urls: ["https://myhost:80"]
-check.request:
-  method: GET
-  headers:
-    'X-API-Key': '12345-mykey-67890'
-check.response:
-  status: 200
-  body:
-    - hello
-    - world
+  check.request:
+    method: GET
+    headers:
+      'X-API-Key': '12345-mykey-67890'
+  check.response:
+    status: 200
+    body:
+      - hello
+      - world
 -------------------------------------------------------------------------------
 
 The following configuration shows how to check the response with a multiline
@@ -505,13 +505,13 @@ regex:
 - type: http
   schedule: '@every 5s'
   urls: ["https://myhost:80"]
-check.request:
-  method: GET
-  headers:
-    'X-API-Key': '12345-mykey-67890'
-check.response:
-  status: 200
-  body: '(?s)first.*second.*third'
+  check.request:
+    method: GET
+    headers:
+      'X-API-Key': '12345-mykey-67890'
+  check.response:
+    status: 200
+    body: '(?s)first.*second.*third'
 -------------------------------------------------------------------------------
 
 
--- a/heartbeat/docs/heartbeat-options.asciidoc
+++ b/heartbeat/docs/heartbeat-options.asciidoc
@@ -459,7 +459,7 @@ Under `check.response`, specify these options:
 *`status`*:: The expected status code. If this setting is not configured or
 it's set to 0, any status code other than 404 is accepted.
 *`headers`*:: The required response headers.
-*`body`*:: The required response body content.
+*`body`*:: A list of regular expressions to match the the body output. Only a single expression needs to match.
 
 The following configuration shows how to check the response when the body
 contains JSON:
@@ -476,6 +476,42 @@ check.request:
 check.response:
   status: 200
   body: '{"status": "ok"}'
+-------------------------------------------------------------------------------
+
+The following configuration shows how to check the response for multiple regex
+patterns:
+
+[source,yaml]
+-------------------------------------------------------------------------------
+- type: http
+  schedule: '@every 5s'
+  urls: ["https://myhost:80"]
+check.request:
+  method: GET
+  headers:
+    'X-API-Key': '12345-mykey-67890'
+check.response:
+  status: 200
+  body:
+    - hello
+    - world
+-------------------------------------------------------------------------------
+
+The following configuration shows how to check the response with a multiline
+regex:
+
+[source,yaml]
+-------------------------------------------------------------------------------
+- type: http
+  schedule: '@every 5s'
+  urls: ["https://myhost:80"]
+check.request:
+  method: GET
+  headers:
+    'X-API-Key': '12345-mykey-67890'
+check.response:
+  status: 200
+  body: '(?s)first.*second.*third'
 -------------------------------------------------------------------------------
 
 
--- a/heartbeat/monitors/active/http/check.go
+++ b/heartbeat/monitors/active/http/check.go
@@ -1,12 +1,12 @@
 package http
 
 import (
-	"bytes"
 	"errors"
 	"fmt"
-	"io"
 	"io/ioutil"
 	"net/http"
+
+	"github.com/elastic/beats/libbeat/common/match"
 )
 
 type RespCheck func(*http.Response) error
@@ -29,7 +29,7 @@ func makeValidateResponse(config *responseParameters) RespCheck {
 	}
 
 	if len(config.RecvBody) > 0 {
-		checks = append(checks, checkBody([]byte(config.RecvBody)))
+		checks = append(checks, checkBody(config.RecvBody))
 	}
 
 	return checkAll(checks...)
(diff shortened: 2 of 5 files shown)


heartbeat/docs/heartbeat-options.asciidoc
```diff
@@ -478,6 +478,42 @@ check.response:
...
 
+The following configuration shows how to check the response for multiple regex
+patterns:
+
+[source,yaml]
+-------------------------------------------------------------------------------
+- type: http
+  schedule: '@every 5s'
+  urls: ["https://myhost:80"]
+check.request:
```
ruflin: I think the indentation is off here. It should be on the same level as `urls` for example.

Same in all the examples below.

heartbeat/monitors/active/http/check.go
```diff
@@ -84,18 +84,17 @@ func checkHeaders(headers map[string]string) RespCheck {
...
-		content, err := ioutil.ReadAll(in)
+		content, err := ioutil.ReadAll(r.Body)
 		if err != nil {
 			return err
 		}
-
-		if !bytes.Equal(body, content) {
-			return errBodyMismatch
+		for _, m := range body {
+			if m.MatchString(string(content)) {
```
urso: use `m.Match(content)`, so to reduce the chance of allocating and copying all contents, just for matching.

heartbeat/monitors/active/http/check_test.go
```diff
@@ -0,0 +1,106 @@
...
+		{
+			"Multiple regex match where none of the patterns match",
+			"ok",
+			[]match.Matcher{match.MustCompile("notok"), match.MustCompile("yay")},
+			false,
+		},
+	}
+
+	for _, test := range matchTests {
+		t.Run(test.description, func(t *testing.T) {
```
urso: +1

heartbeat/monitors/active/http/check_test.go
```diff
@@ -0,0 +1,106 @@
...
+			ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
urso: Requiring an http server is somewhat overkill. Makes me wonder if we can 'simplify' the validators interface/function, so to not depend on a http response.

Especially the fact the `body` validator consuming the complete body is not useful if we want to be able to combine (and/or) more complicated filtering in the future.

Using interfaces one could define:

```
type response struct {
  Status() uint
  Body() ([]byte, error)
  ...
}
```

The adapter for use with http.Response could lazily read and cache the body then.

heartbeat/monitors/active/http/check_test.go
```diff
@@ -0,0 +1,106 @@
...
+
+	"github.com/elastic/beats/libbeat/common/match"
+)
+
+func TestCheckBody(t *testing.T) {
+
+	var matchTests = []struct {
+		description string
+		body        string
+		patterns    []match.Matcher
```
urso: as these are patterns, how about `[]string`, so to make the table a little more readable?

urso:
Switching to `match.Matcher` is not exactly backwards compatible. The original string comparison did require an exact match. The `match.Matcher` implementation checks for a matching sub-string in the body. => users will have to use `^<contents string>$`, so to have exact same behaviour and runtime complexity.

For the future I was considering more enhanced body matching, that can also support parsing json/xml, so to check a document field it's value. This would require the `body` validator to provide different validator types. Like:

full contents:
```
body.equals: <full contents string>
```

using matches:

```
body.matches: <regex>
```

json:

```
body.json:
  field: ...
  equals: ...
```

...

heartbeat/docs/heartbeat-options.asciidoc
Crazybus: Nice catch! The [current example](https://www.elastic.co/guide/en/beats/heartbeat/current/configuration-heartbeat-options.html) also has the wrong indentation too (which I copy pasted from)
image. So I'll fix up the original too.

heartbeat/monitors/active/http/check.go
Crazybus: Fixed.

heartbeat/monitors/active/http/check_test.go
Crazybus: If it isn't obvious I'm still very new to writing go :) Does this look a bit more readable now? https://github.com/elastic/beats/pull/6539/commits/f63257b1f50825a3f51a393ba56f502b2827bdc4 Or is there maybe an even cleaner way to do it?

ruflin:
@urso I would be fine with the breaking change here as we are still in beta and can document it. WDYT? In both cases the change to the above propose structure will be breaking I think.

heartbeat/monitors/active/http/check_test.go
Crazybus: I totally agree 👍 

> Makes me wonder if we can 'simplify' the validators interface/function, so to not depend on a http response.

The wording you used here makes this comment sound like "something we would like to improve in the future". Can you confirm that I interpreted this correctly? Otherwise I'd be happy to have a go refactoring in your suggestion but I would definitely need a bit of help since this goes beyond my basic go knowledge.

Crazybus:
@urso Thanks for the review! I'm still very new to writing Go so the feedback really taught me a lot!

> Switching to match.Matcher is not exactly backwards compatible. The original string comparison did require an exact match. The match.Matcher implementation checks for a matching sub-string in the body. => users will have to use ^<contents string>$, so to have exact same behaviour and runtime complexity.

Ahh yes you are right! 

It would be easy enough to modify the regex to remain backwards compatible by requiring it to be an exact match. I don't feel strongly either way but I personally would vote for the default to be "match this string anywhere in the contents". In my experience checking the body is normally done as a "site is up and has real content" sanity check. Instead of only checking for a HTTP 200 which a lot of "website is on fire" status pages return. 

However as a user I'm trying to think of a realistic scenario where this change in behaviour would actually cause issues. I can think of one example where this could bite someone but I would really hope nobody has a healthcheck endpoint like this. 

* If a user had a health check endpoint that returned `ok` and `not ok` and `check.response.body: ok` and both endpoints returned a `200`. With the new regex style checking the body would match for both outputs.

> For the future I was considering more enhanced body matching, that can also support parsing json/xml, so to check a document field it's value. This would require the body validator to provide different validator types.

I like all of these suggestions. In case you missed it I made some similar proposals in https://github.com/elastic/beats/issues/5981 . The main one I am interested in is having a proper way to check JSON output. The main use case would be monitoring the health of our Elasticsearch clusters that are running in cloud. 

```
check.response:
  bodyJSON:
     status: "green"
```

Which could of course be done with a regex too

```
check.response:
  body: '"status": "green"
```

ruflin:
To move forward on this I suggest we go with the breaking change but document it properly in the changelog. From there we can further discuss about the above proposal to make it even more powerful.

Crazybus:
> @Crazybus I see you put this PR 'On Hold'. 

Sorry for the confusion, my "On Hold" status is also used when I'm waiting to hear back from someone (in this case you). 

> Given there are no critical blockers in this PR, how about merging it as is and do further (general) improvements on validation in future PRs.

Sounds good to me 👍Is it ok for me to merge this in myself? I do have the rights (as a GitHub admin) however I'm not part of the beats team so I just wanted to check first.

urso:
Normally the PR dev doesn't merge the PR.

Can you rebase on top of master and add an changelog entry? I will merge the PR once CI is green. Thanks!

urso:
Merged. Thank you for improving heartbeat.

PR #6542, opened 2018-03-13
Send available perfmon data on error

andrewkroh:
When the windows/perfmon metricset encountered any error it would discard any valid metrics it had read and send a single error event. With this change it will send one event per counter. If an individual counter returns an error then it will send an event for that counter containing the error.message field and a zero-value measurement field. Depending on your use case you may want to update any metric aggregations you have to ignore error events.

I'm also adding more documentation for the configuration options.


metricbeat/module/windows/perfmon/pdh_windows.go
```diff
@@ -368,38 +372,37 @@ func (r *PerfmonReader) Read() ([]common.MapStr, error) {
...
-			result = append(result, ev)
-		}
-	}
+			event := mb.Event{
+				MetricSetFields: common.MapStr{},
+				Error:           errors.Wrapf(val.Err, "failed on query=%v", counterPath),
+			}
+
+			if val.Instance != "" {
+				event.MetricSetFields.Put(r.instanceLabel[counterPath], val.Instance)
```
ruflin: @exekias I wonder if this would also be a place to use safteput. Also below.

exekias: I don't have much knowledge about Windows perfom metrics, in general `safemapstr.Put` would be required when we know the field may contain dots and these don't necessarily imply nesting. Basically when field names can cause a nesting issue like this:

```
field1.total: 3
field1: "foo"
```

andrewkroh: I don't believe `safemapstr.Put` is very applicable here (I did consider it during this fix). The user has full control over the field names plus there is only one metric in an event.

andrewkroh:
Updated with a CHANGELOG entry.

andrewkroh:
> For my understanding: As long as all counters exist, nothing changes from the previous behaviour.

Correct.

And if a counter does not exist, instead of getting a single error event you will get one error event per counter PLUS an event per each valid counter.

PR #6545, opened 2018-03-13
Use safemapstr.Put for docker labels

exekias:
Docker module is "dedoting" labels, this change allows to disable that and make use of `safemapstr.Put` to store labels in a nested way. Config looks like this:

```
- module: docker
  metricsets: ["container", "cpu", "diskio", "healthcheck", "info", "memory", "network"]
  period: 10s

  # Replace dots in labels with `_`. Set to false to keep dots
  labels.dedot: true
```

### DIFF
--- a/metricbeat/docs/modules/docker.asciidoc
+++ b/metricbeat/docs/modules/docker.asciidoc
@@ -34,7 +34,7 @@ metricbeat.modules:
   period: 10s
 
   # Replace dots in labels with `_`. Set to false to keep dots
-  dedot: true
+  labels.dedot: true
 
   # To connect to Docker over TLS you must specify a client and CA certificate.
   #ssl:
--- a/metricbeat/docs/modules/docker.asciidoc
+++ b/metricbeat/docs/modules/docker.asciidoc
@@ -33,6 +33,9 @@ metricbeat.modules:
   hosts: ["unix:///var/run/docker.sock"]
   period: 10s
 
+  # Replace dots in labels with `_`. Set to false to keep dots
+  dedot: true
+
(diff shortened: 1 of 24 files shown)


ruflin:
As it's a breaking change we should delay the switch to 7.0. What we should do now is having a config option to disable dedoting for labels so people can already start using it.

exekias:
Oh, now I remember we even talked about that :innocent:, will do the changes

exekias:
Added a `dedot` module setting (`true` by default)

exekias:
This should be ready for a second look

metricbeat/docs/modules/docker.asciidoc
```diff
@@ -33,6 +33,9 @@ metricbeat.modules:
   hosts: ["unix:///var/run/docker.sock"]
   period: 10s
 
+  # Replace dots in labels with `_`. Set to false to keep dots
+  dedot: true
```
ruflin: Can we call this `labels.dedot` to make it clear it's not dedotting the complete event?

metricbeat/modules.d/docker.yml.disabled
```diff
@@ -3,6 +3,9 @@
   hosts: ["unix:///var/run/docker.sock"]
   period: 10s
 
+  # Replace dots in labels with `_`. Set to false to keep dots
+  dedot: true
```
ruflin: `labels.dedot`?

exekias:
This should be ready, failing test is unrelated, tackled here: https://github.com/elastic/beats/pull/6564

ruflin:
@exekias We should perhaps open a meta issue to add docs for all these dedot / safemapstr changes.

PR #6553, opened 2018-03-15
Update screens and instructions to match latest version of Kibana

dedemorton:
Updated the screens and fixed queries that weren't working due to data model changes. Also shuffled some text around to make it easier to follow. Also cleaned up the line wraps. Sorry if this makes the review a little harder.

Note that there are a couple of notes to reviewers that I'll need to remove before we merge this.


dedemorton:
So I went ahead and signed the CLA because I accidentally committed while logged in as a different user (a personal account I set up for a workshop). Anyhow, I hope that's OK.

andrewkroh:
That should be fine. FYI you can fix that with a `git commit --amend --author="Author Name <email@address.com>"`. Then do a force push.

PR #6554, opened 2018-03-15
Move MongoDB module to GA

exekias:


exekias:
I checked current fields, we don't always have the same opinion there so a second look would be ok if you want, I'm happy to change some if you point them out :innocent:

ruflin:
@exekias If you checked and you're happy, I'm happy ;-)

PR #6560, opened 2018-03-15
Configuration to allow users to retry loading dashboards 

jalvz:
... if Kibana not reachable, instead of exiting right away.

### DIFF
--- a/filebeat/filebeat.reference.yml
+++ b/filebeat/filebeat.reference.yml
@@ -1225,13 +1225,13 @@ output.elasticsearch:
 
 # If true and Kibana is not reachable at the time when dashboards are loaded,
 # it will retry to reconnect to Kibana instead of exiting with an error.
-#setup.dashboards.retry.enabled
+#setup.dashboards.retry.enabled: false
 
 # Duration interval between Kibana connection retries.
-#setup.dashboards.retry.interval
+#setup.dashboards.retry.interval: 1s
 
 # Maximum number of retries before exiting with an error, 0 for unlimited retrying.
-#setup.dashboards.retry.maximum
+#setup.dashboards.retry.maximum: 0
 
 
 #============================== Template =====================================
--- a/filebeat/filebeat.reference.yml
+++ b/filebeat/filebeat.reference.yml
@@ -1223,6 +1223,17 @@ output.elasticsearch:
 # how to install the dashboards by first querying Elasticsearch.
 #setup.dashboards.always_kibana: false
 
+# If true and Kibana is not reachable at the time when dashboards are loaded,
+# it will retry to reconnect to Kibana instead of exiting with an error.
+#setup.dashboards.retry.enabled
+
+# Duration interval between Kibana connection retries.
+#setup.dashboards.retry.interval
+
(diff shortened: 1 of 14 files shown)


libbeat/dashboards/config.go
```diff
@@ -11,11 +13,21 @@ type Config struct {
...
+	Enabled  bool          `config:"enabled"`
+	Interval time.Duration `config:"interval"`
+	Maximum  uint          `config:"maximum"`
 }
 
 var defaultConfig = Config{
 	KibanaIndex:  ".kibana",
 	AlwaysKibana: false,
+	Retry: &Retry{
+		Enabled: false,
```
ruflin: Can you add the defaults for interval and maximum here too?

jalvz: sure thing, what do you think is reasonable? 
1 sec/ unlimited?

ruflin: SGTM, especially as the feature is off by default.

jalvz: done 3bd3affaa9242bd7072207a911f1313ead19c30e

jalvz:
added changelog fdc0308f204f20d088b571460a509bae5b2427a5

libbeat/docs/dashboardsconfig.asciidoc
```diff
@@ -99,3 +99,19 @@ NOTE: This setting only works for Kibana 6.0 and newer.
 
 Force loading of dashboards using the Kibana API without querying Elasticsearch for the version
 The default is `false`.
+
+[float]
+==== `setup.dashboards.retry.enabled`
```
ruflin: Thanks for the docs.

jalvz:
done  68d3c51a359076eda33338ab62def80e26eb57b2
is this a concern?
https://beats-ci.elastic.co/job/elastic+beats+pull-request+multijob-windows/3524/beat=filebeat,label=windows/testReport/junit/test_crawler/Test/test_tail_files/

filebeat/filebeat.reference.yml
```diff
@@ -1223,6 +1223,17 @@ output.elasticsearch:
 # how to install the dashboards by first querying Elasticsearch.
 #setup.dashboards.always_kibana: false
 
+# If true and Kibana is not reachable at the time when dashboards are loaded,
+# it will retry to reconnect to Kibana instead of exiting with an error.
+#setup.dashboards.retry.enabled
```
ruflin: If there are default values we normally put them into the reference doc, so here it would be `#setup.dashboards.retry.enabled: false`. Similar for the values below.

libbeat/docs/dashboardsconfig.asciidoc
```diff
@@ -99,3 +99,19 @@ NOTE: This setting only works for Kibana 6.0 and newer.
...
+
+If this option is set to true, and Kibana is not reachable at the time when dashboards are loaded,
```
ruflin: My comment above made me realise that having a comment about the defaults here would also be great. Some at then of the sentence like. `Default is false.`

ruflin:
Thanks for the addition.

PR #6341, opened 2018-02-09
Update golang/x/net to release-branch.go1.9

cyrilleverrier:

PR #6563, opened 2018-03-15
Fix data race in harvester termination

adriansr:
A data race was sometimes triggered by system tests where a harvester is Run() and Stop()'ed in quick succession from different goroutines.

It is necessary to protect access to a sync.WaitGroup with a mutex to prevent wg.Add(1) and wg.Wait() to execute concurrently.

This race condition had little or no impact in real world usage.

Failed tests for #6341

### DIFF
--- a/filebeat/input/log/harvester.go
+++ b/filebeat/input/log/harvester.go
@@ -62,6 +62,7 @@ type Harvester struct {
 	done     chan struct{}
 	stopOnce sync.Once
 	stopWg   *sync.WaitGroup
+	stopLock sync.Mutex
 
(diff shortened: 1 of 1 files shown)


adriansr:
With this patch, input/log/Harvester will be using a WaitGroup, a Mutex, a Channel, and a Once for termination :) Maybe someone else can come up with a cleaner pattern for shutting it down.

ruflin:
Thanks for tracking this down and fixing it. I hope we find in general a simpler solutions for all the locks, channel etc. on shutdown.

PR #6566, opened 2018-03-15
Fix `state_container` metrics when several containers have the same name

exekias:
Most of this was fixed by https://github.com/elastic/beats/pull/6294,
but for the specific case of containers, their uniqueness is defined by
namespace + pod name. As several pods may have containers with the same
name.

Rest of metrics are not affected in the same way, and namespace should
be enough for them.

Not adding a CHANGELOG as #6294 wasn't released yet

### DIFF
--- a/metricbeat/module/kubernetes/state_container/state_container_test.go
+++ b/metricbeat/module/kubernetes/state_container/state_container_test.go
@@ -44,22 +44,27 @@ func TestEventMapping(t *testing.T) {
 	events, err := f.Fetch()
 	assert.NoError(t, err)
 
-	assert.Equal(t, 10, len(events), "Wrong number of returned events")
+	assert.Equal(t, 12, len(events), "Wrong number of returned events")
 
 	testCases := testCases()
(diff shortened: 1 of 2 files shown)


exekias:
jenkins, retest this please

metricbeat/module/kubernetes/state_container/state_container_test.go
```diff
@@ -44,22 +44,27 @@ func TestEventMapping(t *testing.T) {
...
 			namespace, err := event.GetValue("_module.namespace")
-			if err == nil {
-				eventKey := namespace.(string) + "@" + name.(string)
-				oneTestCase, oneTestCaseFound := testCases[eventKey]
-				if oneTestCaseFound {
-					for k, v := range oneTestCase {
-						testValue(eventKey, t, event, k, v)
-					}
-					delete(testCases, eventKey)
+			if err != nil {
```
ruflin: Makes also the code more readable 👍

PR #6581, opened 2018-03-16
Introduce spooling to disk

urso:
This PR implements the `queue.Queue` interface, adding spooling to disk functionality to all beats. The queue interface requires all queues to provide 'connections' for Producers and Consumers. The interface also demands all ACKs to be executed asynchronously. Once an event is ACKed, it must not be presented to a Consumer (the outputs worker queue) anymore.

The new queue type is marked as Beta-feature.

In the spool, the events are stored in a dynamic ring buffer, in one single file only. The maximum file size must be configured at startup (default 100MiB). The file layer needs to use available file space for event data and additional meta data. Events are first written into a write buffer, which is flushed once it is full or if the flush timeout is triggered. No limit on event sizes is enforced. An event being bigger then the write buffer will still be accepted by the spool, but will trigger a flush right away. A successful flush also requires 2 fsync operations (1. for data 2. update file header). The spool blocks all producers, once it can not allocate any more space within the spool file. Writing never grows the file past the configured maximum file size. All producers are handled by the `inBroker`. There is no direct communication between producers and consumers. All required signaling and synchronisation is provided by the file layer.

Once the write buffer is flushed, a signal is returned to each individual Producer, notifying the producers of events being published. This signal is used by filebeat/winlogbeat to update the registry file.

Consumers are handled by the `outBroker`. Consumers request a batch of events. The broker reads up to N message from the file and forwards these. The reading process is 'readonly' and does not update the on-disk read pointers (only in-memory read pointers are updated).
The file is memory mapped for reading. This will increases the process it's reported memory usage significantly.

The outputs asynchronously ACK batches. The ACK signals are processed by the brokers `ackLoop`. Due to load-balancing or retries, ACKs can be received out of order. The broker guarantees, ACKs are sorted in the same order events have been read from the queue. Once a continuous set of events (starting from the last on-disk read pointer) is ACKed, the on-disk read pointer is update and space occupied by ACKed events is freed. As free space is tracked by the file, the file meta-data must be updated. If no more space for file meta data updates is available, there is a chance of the file potentially growing a few pages past max-size. Growing is required to guarantee progress (otherwise the spool might be stalled forever). In case the file did grow on ACK, the file layer will try to free the space with later write/ACK operations, potentially truncating the file again.

The file layer is provided by [go-txfile](github.com/elastic/go-txfile). The file is split into pages of equal size. The layer provides transactional access to pages only. All writes (ACK, flushing the write buffer) are handled concurrently, using write transaction. The reader is isolated from concurrent writes/reads, using a read transaction. The last transaction state is never overwritten in a file. If a beat crashes during a write transaction, the most recent committed transaction is still available, so the beat can continue from the last known state upon restart. No additional repair-phase is required.

Known limitations (To be addressed in future PRs):
- File maximum size can not be changed once file is generated:
  - Add support to grow max size
  - Add support to shrink max size. Shrinking could be made dynamic, trying to return space once pages at end of file are freed
  - Add command to transfer the queue into a new queue file. In case fragmentation prevents dynamic shrinking or user doesn't want to wait. 
- Monitoring metrics do not report already available events upon beats restart (requires some more changes to the queue interface itself).
- No file related metrics available yet.
- If file is too small and write buffer to big, queue can become stalled. Potential solution requires (all):
  - Limit maximum event size
  - Generously preallocate meta-area (reserve pages for meta-data only on file creation)
  - Ensure usable data area is always > 2*write buffer -> partial write buffer flush?
  - Startup check validating combination of max_size, data vs. meta area size, event size, write buffer size

### DIFF
--- a/libbeat/publisher/pipeline/stress/stress_test.go
+++ b/libbeat/publisher/pipeline/stress/stress_test.go
@@ -43,12 +43,22 @@ func TestPipeline(t *testing.T) {
 	}
 
 	if duration == 0 {
-		duration = 10 * time.Second
+		duration = 15 * time.Second
 	}
 
 	configTest(t, "gen", genConfigs, func(t *testing.T, gen string) {
 		configTest(t, "pipeline", pipelineConfigs, func(t *testing.T, pipeline string) {
 			configTest(t, "out", outConfigs, func(t *testing.T, out string) {
+
+				if testing.Verbose() {
+					start := time.Now()
+					fmt.Printf("%v Start stress test %v\n", start.Format(time.RFC3339), t.Name())
+					defer func() {
+						end := time.Now()
+						fmt.Printf("%v Finished stress test %v. Duration=%v\n", end.Format(time.RFC3339), t.Name(), end.Sub(start))
+					}()
+				}
+
 				config, err := common.LoadFiles(gen, pipeline, out)
 				if err != nil {
 					t.Fatal(err)
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,7 +68,7 @@ jobs:
       go: $GO_VERSION
       stage: test
     - os: linux
-      env: TARGETS="-C libbeat stress-tests"
+      env: STRESS_TEST_OPTIONS="-timeout=20m -race -v -parallel 1" TARGETS="-C libbeat stress-tests"
       go: $GO_VERSION
       stage: test
 
--- a/libbeat/publisher/pipeline/stress/gen.go
+++ b/libbeat/publisher/pipeline/stress/gen.go
@@ -1,7 +1,9 @@
 package stress
 
 import (
+	"bytes"
 	"fmt"
+	"runtime/pprof"
 	"sync"
 	"time"
 
@@ -25,7 +27,7 @@ var defaultGenerateConfig = generateConfig{
 	ACK:       false,
 	MaxEvents: 0,
 	WaitClose: 0,
-	Watchdog:  1 * time.Second,
+	Watchdog:  2 * time.Second,
 }
 
 var publishModes = map[string]beat.PublishMode{
@@ -92,7 +94,7 @@ func generate(
 		// start generator watchdog
 		withWG(&wg, func() {
 			last := uint64(0)
-			ticker := time.NewTicker(config.Watchdog) // todo: make ticker interval configurable
+			ticker := time.NewTicker(config.Watchdog)
 			defer ticker.Stop()
 			for {
 				select {
@@ -105,7 +107,11 @@ func generate(
 
 				current := count.Load()
 				if last == current {
-					err := fmt.Errorf("no progress in generators (last=%v, current=%v)", last, current)
+					// collect all active go-routines stack-traces:
+					var buf bytes.Buffer
+					pprof.Lookup("goroutine").WriteTo(&buf, 2)
+
+					err := fmt.Errorf("no progress in generator %v (last=%v, current=%v):\n%s", id, last, current, buf.Bytes())
 					errors(err)
 				}
 				last = current
--- a/libbeat/scripts/Makefile
+++ b/libbeat/scripts/Makefile
@@ -52,6 +52,7 @@ NOSETESTS_OPTIONS?=--process-timeout=$(TIMEOUT) --with-timer -v --with-xunit --x
 TEST_ENVIRONMENT?=false ## @testing if true, "make testsuite" runs integration tests and system tests in a dockerized test environment
 SYSTEM_TESTS?=false ## @testing if true, "make test" and "make testsuite" run unit tests and system tests
 STRESS_TESTS?=false ## @testing if true, "make test" and "make testsuite" run also run the stress tests
+STRESS_TEST_OPTIONS?=-timeout=20m -race -v
 GOX_OS?=linux darwin windows freebsd netbsd openbsd ## @Building List of all OS to be supported by "make crosscompile".
 GOX_OSARCH?=!darwin/arm !darwin/arm64 !darwin/386 ## @building Space separated list of GOOS/GOARCH pairs to build by "make crosscompile".
 GOX_FLAGS?= ## @building Additional flags to append to the gox command used by "make crosscompile".
@@ -206,7 +207,7 @@ fast-system-tests: ${BEAT_NAME}.test python-env
 stress-tests: ## @testing Runs the stress tests with race detector enabled
 stress-tests:
 	if [ -n '${GOPACKAGES_STRESSTESTS}' ]; then \
-		go test -race --tags=stresstest -v ${GOPACKAGES_STRESSTESTS}; \
+		go test --tags=stresstest ${STRESS_TEST_OPTIONS} ${GOPACKAGES_STRESSTESTS}; \
 	fi
 
 # Run benchmark tests


libbeat/publisher/pipeline/stress/stress_test.go
```diff
@@ -68,7 +81,7 @@ func configTest(t *testing.T, typ string, configs []string, fn func(t *testing.T
 	for _, config := range configs {
 		config := config
 		t.Run(testName(typ, config), func(t *testing.T) {
-			t.Parallel()
+			// t.Parallel()
```
urso: TODO: restore capability to run tests in parallel (will need randomized file names)

auditbeat/auditbeat.reference.yml
```diff
@@ -148,6 +148,50 @@ auditbeat.modules:
...
     #flush.timeout: 1s
 
+  # The spool queue will store events in a local spool file, before
+  # forwarding the events to the outputs.
+  # The spool file is a circular buffer, which blocks once the file/buffer is full.
+  # Events are put into a write buffer and flushed once the write buffer
+  # is full or the flush_timeout is triggered.
+  # Once ACKed by the output, events are removed immediately from the queue,
+  # making space for new events to be persisted.
+  #spool:
```
ruflin: Why not call this `file` or `disk`? It would make it more obvious that is persistent.

auditbeat/auditbeat.reference.yml
```diff
@@ -148,6 +148,50 @@ auditbeat.modules:
...
+    # The file namespace configures the file path and the file creation settings.
+    # Once the file exists, the `size`, `page_size` and `prealloc` settings
+    # will have no more effect.
+    #file:
+      # Location of spool file. This value must be configured.
+      #path: ""
```
ruflin: Should this default to our `path.data` directory?

auditbeat/auditbeat.reference.yml
```diff
@@ -148,6 +148,50 @@ auditbeat.modules:
     # if the number of events stored in the queue is < min_flush_events.
     #flush.timeout: 1s
 
+  # The spool queue will store events in a local spool file, before
```
ruflin: I assume only 1 queue can be used at the same time. Should we mention that in a comment on line 130?

libbeat/publisher/pipeline/stress/data/placeholder
```diff
@@ -0,0 +1 @@
+// Do not remove
```
ruflin: Perhaps add a note on why it should not be removed.

libbeat/publisher/pipeline/stress/run.go
```diff
@@ -36,13 +37,13 @@ func RunTests(
...
 	if err := cfg.Unpack(&config); err != nil {
-		return err
+		return fmt.Errorf("unpacking config failed: %v", err)
 	}
 
 	// reg := monitoring.NewRegistry()
 	pipeline, err := pipeline.Load(info, nil, config.Pipeline, config.Output)
 	if err != nil {
-		return err
+		return fmt.Errorf("Loading pipeline failed: %v", err)
```
ruflin: start error lower case

libbeat/publisher/queue/spool/codec.go
```diff
@@ -0,0 +1,182 @@
...
+	Timestamp int64
+	Flags     uint8
+	Meta      common.MapStr
+	Fields    common.MapStr
+}
+
+const (
+	// Note: Never change order. Codec IDs must be not change in the future. Only
+	//       adding new IDs is allowed.
+	codecUnknown codecID = iota
```
ruflin: Not sure I understand why we need different codes. It seems it could be configurable through the config but it's not listed in the reference file. I assume it would also cause issues if one changes the code after the beats was running. 

It's good for testing to have the different codes but should we for now just set one?

libbeat/publisher/queue/spool/config.go
```diff
@@ -0,0 +1,123 @@
...
+
+func (c *pathConfig) Validate() error {
+	var errs multierror.Errors
+
+	if c.Path == "" {
+		errs = append(errs, errors.New("empty path"))
+	}
+
+	if c.MaxSize < humanize.MiByte {
+		errs = append(errs, errors.New("max size must be larger 1MiB"))
```
ruflin: What happens if someone has a multiline event of 2MB and the MaxSize here is 1.1MiB?

libbeat/_meta/config.reference.yml
```diff
@@ -43,6 +43,50 @@
...
     #flush.timeout: 1s
 
+  # The spool queue will store events in a local spool file, before
+  # forwarding the events to the outputs.
+  # The spool file is a circular buffer, which blocks once the file/buffer is full.
+  # Events are put into a write buffer and flushed once the write buffer
+  # is full or the flush_timeout is triggered.
+  # Once ACKed by the output, events are removed immediately from the queue,
+  # making space for new events to be persisted.
+  #spool:
```
ruflin: We should mark this beta at first, in the code as a log message, here in the config and in the docs.

libbeat/publisher/queue/spool/inbroker.go
```diff
@@ -0,0 +1,482 @@
...
+// any new events.
+// The broker will wait for an ACK signal from the outputs and retry flushing,
+// in the hope of enough memory being available to flush the buffers.
+// If flush did succeed, we try to add the pending event.
+// For the time the broker is in this state, no events from any producers will
+// be accepted. Thusly all producers will block. Closing a producer, unblocks
+// the producer. The producers event (after close) might be processed or
+// ignored in the future.
+//
+// stateBlocked transitions:
```
ruflin: In the case of Filebeat this means the harvesters are blocked. In case of metricbeat does means the last events coming in will be dropped? So if the output becomes available again, all events in the queue will be sent and the events which came in between "queue full" and "output available again" will be lost.

metricbeat/metricbeat.reference.yml
```diff
@@ -551,6 +551,50 @@ metricbeat.modules:
...
+  # Once ACKed by the output, events are removed immediately from the queue,
+  # making space for new events to be persisted.
+  #spool:
+    # The file namespace configures the file path and the file creation settings.
+    # Once the file exists, the `size`, `page_size` and `prealloc` settings
+    # will have no more effect.
+    #file:
+      # Location of spool file. This value must be configured.
+      #path: ""
+
```
kvch: My vim shows an unnecessary tab here.

metricbeat/metricbeat.reference.yml
```diff
@@ -551,6 +551,50 @@ metricbeat.modules:
...
+    # The file namespace configures the file path and the file creation settings.
+    # Once the file exists, the `size`, `page_size` and `prealloc` settings
+    # will have no more effect.
+    #file:
+      # Location of spool file. This value must be configured.
+      #path: ""
+      
+      # Configure file permissions if file is created. The default value is 0600.
+      #permissions: 0600
+
```
kvch: Also, an unncessary tab here.

libbeat/publisher/queue/spool/spool.go
```diff
@@ -0,0 +1,221 @@
...
+	file  *txfile.File
+}
+
+type spoolCtx struct {
+	logger logger
+	wg     sync.WaitGroup
+	active atomic.Bool
+	done   chan struct{}
+
+	handlers []func()
```
kvch: I think calling the list of functions set by users to run when a `spoolCtx` is too general. I think handlers can be called whenever they are needed. But this list contains only a special kind of handlers which run during closing the session.
I would rather name it `onCloseCBs`, `onCloseCallbacks` or `closingCallbacks`. This change would make more obvious when these functions are called and what their role is.

libbeat/publisher/queue/spool/consume.go
```diff
@@ -0,0 +1,118 @@
...
+		// internal API
+		resp: make(chan getResponse),
+		requ: requ,
+	}
+}
+
+func (c *consumer) Get(sz int) (queue.Batch, error) {
+	log := c.ctx.logger
+
+	if c.closed.Load() || c.ctx.Closed() {
```
kvch: It would be nice to wrap `c.closed.Load` into a function named `c.Closed`. Just like it is done in `spoolCtx`.

libbeat/publisher/queue/spool/outbroker.go
```diff
@@ -0,0 +1,489 @@
...
+	ctx   *spoolCtx
+	state func(*outBroker) bool
+
+	// internal API
+	sigFlushed chan uint
+	get        chan getRequest
+
+	// ack signaling
+	pendingACKs   chanList      // list of pending batches to be forwarded to the ackLoop
+	scheduledACKs chan chanList // shared channel for forwading batches to ackLoop
```
ph: s/forwading/forwarding/

libbeat/publisher/queue/spool/outbroker.go
```diff
@@ -0,0 +1,489 @@
...
+
+var ackChanPool = sync.Pool{
+	New: func() interface{} {
+		return &ackChan{
+			ch: make(chan batchAckMsg, 1),
+		}
+	},
+}
+
+var errRetry = errors.New("retry")
```
ph: I had to check where it was used to understand the context of this error, from what I see it is used to signal that we need to **retry to decode the event** when no **events were decoded**. Maybe `errNoEventDecoded` with a better message?

libbeat/publisher/queue/spool/outbroker.go
```diff
@@ -0,0 +1,489 @@
...
+
+func (b *outBroker) ackLoop() {
+	log := b.ctx.logger
+
+	log.Debug("start output ack loop")
+	defer log.Debug("stop output ack loop")
+
+	var (
+		activeList chanList
+	)
```
ph: nitpick, move to a single line.

libbeat/publisher/queue/spool/outbroker.go
```diff
@@ -0,0 +1,489 @@
...
+
+		case <-activeList.channel():
+			ackCh := activeList.pop()
+
+			for {
+				log.Debugf("receive ACK of %v events\n", ackCh.total)
+				err := b.queue.ACK(uint(ackCh.total))
+				if err != nil {
+					log.Debug("ack failed with:", err)
+					time.Sleep(1 * time.Second)
```
ph: My understanding of this part, is we want to give some backoff on the acking when en error occur.
Suggestion, use an exponential backoff with a fixed cap using 1s by default seems a high?

libbeat/publisher/queue/spool/outbroker.go
```diff
@@ -0,0 +1,489 @@
...
+			break
+		}
+
+		// not enough events -> start timer and try to collect more
+		b.events = events
+		b.required = required
+		b.active = req
+		b.total = total
+		b.timer.Start()
+		log.Debug("  outbroker (stateActive): switch to stateWithTimer")
```
ph: uneeded space?

libbeat/publisher/queue/spool/outbroker.go
```diff
@@ -0,0 +1,489 @@
...
+	case <-b.timer.C:
+		b.timer.Stop(true)
+		log.Debug("outbroker (stateWithTimer): flush timer")
+		b.returnEvents(b.active, b.events, b.total)
+
+		log.Debug("outbroker (stateWithTimer): switch to stateActive")
+		b.initState()
+
+	case n := <-b.sigFlushed:
+		// yay, more events \o/
```
ph: 🎉

libbeat/publisher/queue/spool/produce.go
```diff
@@ -0,0 +1,190 @@
...
+	s.clients = s.clients[:len(s.clients)-1]
+	s.mux.Unlock()
+}
+
+func (s *clientStates) Pop(n int) (states []clientState) {
+	s.mux.Lock()
+	states, s.clients = s.clients[:n], s.clients[n:]
+	s.mux.Unlock()
+	return states
+}
```
ph: nitpick, I would have used a _defer_ for the _unlock_.

libbeat/publisher/queue/spool/spool.go
ph: agreed ^

kvch:
I've finished the first round of review. I added comments with my questions/requests and agree with most of what @ph noted. I am waiting for your changes/responses.

I tested it manually and I haven't found anything problematic so far.

ph:
I did some test manually too over the weekend, was green for me, but it still high level testing.

@urso @kvch I think we need to come up with a solution for stress testing this, do we want to do some `chaos monkley` suite for that?

auditbeat/auditbeat.reference.yml
urso: @ph @ruflin 

Hmmm... introducing `queue.type` would be a bc-breaking change. As one can have only one queue type, the queue config follows the dictionary style config-pattern like:

```
queue.<type>:
```

What wrong with `spool`?

Right now this one implements a FIFO. For metricbeat use cases we might actually introduce another persistent LIFO queue type. This one should have another type. How would you name any of these now?

auditbeat/auditbeat.reference.yml
urso: Yeah, was thinking the same. Will default to `${path.data}/spool.dat`.

auditbeat/auditbeat.reference.yml
urso: We need to?

libbeat/publisher/pipeline/stress/data/placeholder
urso: will change tests to create a temporary directories. Was just nice to have files available after tests (for inspection).

When removing the file, git will remove the directory -> tests would fail.

libbeat/publisher/queue/spool/codec.go
urso: The JSON codec is nice for debugging(assuming I would have to use an hexeditor). Checking out a file in an hexeditor can be complicated enough. No need to make debugging more complicated by having binary data only.

The UBJSON can be removed.

Benchmarks are available in the go-structform tests (using beats events):

```
BenchmarkDecodeBeatsEvents/packetbeat/structform-json-4         	     200	   9210052 ns/op	  74.12 MB/s	 3251196 B/op	   62062 allocs/op
BenchmarkDecodeBeatsEvents/packetbeat/structform-ubjson-4       	     200	   7440986 ns/op	  87.19 MB/s	 3171845 B/op	   56362 allocs/op
BenchmarkDecodeBeatsEvents/packetbeat/structform-cborl-4        	     200	   7310081 ns/op	  81.46 MB/s	 3171266 B/op	   56361 allocs/op
BenchmarkDecodeBeatsEvents/metricbeat/structform-json-4         	     500	   3333191 ns/op	  62.41 MB/s	 1387710 B/op	   24852 allocs/op
BenchmarkDecodeBeatsEvents/metricbeat/structform-ubjson-4       	     500	   2955721 ns/op	  72.92 MB/s	 1387472 B/op	   24758 allocs/op
BenchmarkDecodeBeatsEvents/metricbeat/structform-cborl-4        	     500	   2816873 ns/op	  67.70 MB/s	 1387136 B/op	   24758 allocs/op
BenchmarkDecodeBeatsEvents/filebeat/structform-json-4           	      30	  46966885 ns/op	 105.49 MB/s	15055009 B/op	  390517 allocs/op
BenchmarkDecodeBeatsEvents/filebeat/structform-ubjson-4         	      30	  38593911 ns/op	 126.92 MB/s	14874314 B/op	  390499 allocs/op
BenchmarkDecodeBeatsEvents/filebeat/structform-cborl-4          	      50	  36702929 ns/op	 123.75 MB/s	14874551 B/op	  390500 allocs/op
BenchmarkEncodeBeatsEvents/packetbeat/structform-json-4         	     200	   5970464 ns/op	 114.26 MB/s	       0 B/op	       0 allocs/op
BenchmarkEncodeBeatsEvents/packetbeat/structform-ubjson-4       	     500	   3749375 ns/op	 173.04 MB/s	       0 B/op	       0 allocs/op
BenchmarkEncodeBeatsEvents/packetbeat/structform-cborl-4        	     500	   2829715 ns/op	 210.43 MB/s	       0 B/op	       0 allocs/op
BenchmarkEncodeBeatsEvents/metricbeat/structform-json-4         	    1000	   2274028 ns/op	  91.03 MB/s	       0 B/op	       0 allocs/op
BenchmarkEncodeBeatsEvents/metricbeat/structform-ubjson-4       	    1000	   1672348 ns/op	 128.88 MB/s	       0 B/op	       0 allocs/op
BenchmarkEncodeBeatsEvents/metricbeat/structform-cborl-4        	    1000	   1236829 ns/op	 154.18 MB/s	       0 B/op	       0 allocs/op
BenchmarkEncodeBeatsEvents/filebeat/structform-json-4           	      50	  32523323 ns/op	 151.97 MB/s	       0 B/op	       0 allocs/op
BenchmarkEncodeBeatsEvents/filebeat/structform-ubjson-4         	     100	  22533875 ns/op	 217.37 MB/s	       0 B/op	       0 allocs/op
BenchmarkEncodeBeatsEvents/filebeat/structform-cborl-4          	     100	  15919294 ns/op	 285.32 MB/s	       0 B/op	       0 allocs/op
```

ubjson is faster on decode, cborl is much faster on encode :)

libbeat/publisher/queue/spool/config.go
urso: Well, there are some "limits" that are not supported.

If event is > total file size, we will block forever.

As event sizes are dynamic and we don't want to employ hard limits on event sizes, we have kind of a weird effect here. There are potential parameters combinations + event sizes that can block the queue. The best we could do is: introduce a max event size and drop events if they hit the limit. This way we can precompute the worst page-usage of an event and figure if max size must be even larger. It's kinda tricky, that's why I didn't do it yet.

libbeat/publisher/queue/spool/inbroker.go
urso: It's the same behavior as for the in memory queue. If the queue is full, it is full! If queue is full, it blocks (no event is dropped).

Block vs. drop (beside having similar effects on metricbeat) do have subtle differences. Metricbeat could report/monitor long block periods by having a watchdog checking metricsets actually report events with the configured period (or multiple of the configured periods).

It's by design. Administrator pre-allocate/assign disk space for use by beats and beats guarantees (best effort) it does not use any more disk space once the queue is full. Better have beats block if queue is full then making a system go out of space.

libbeat/publisher/queue/spool/spool.go
urso: Urgh... I felt dirty when introducing the `handlers`. No I'm feeling more "dirty" giving the impression we want multiple of these callbacks. Let's see if I can get rid of `handlers` or other callbacks.

Also not, these are for internal use only. No external lib/component is ever allowed to install a custom handler! The idea of `spoolCtx` is to provide a very small required subset of `context.Context` in order to deal with queue shutdown.

urso:
@ph > So we need to clear out any unnecessary space.

The spaces in logp.Debug are on purpose. It's like a 'multiline' log. The indentation helps me a little to keep context. :/

libbeat/publisher/queue/spool/config.go
```diff
@@ -0,0 +1,123 @@
...
+		errs = append(errs, fmt.Errorf("permissions %v are not regular file permissions", c.Permissions.String()))
+	} else {
+		m := c.Permissions.Perm()
+		if (m & 0400) == 0 {
+			errs = append(errs, errors.New("file must be readable by current user"))
+		}
+		if (m & 0200) == 0 {
+			errs = append(errs, errors.New("file must be writable by current user"))
+		}
+	}
```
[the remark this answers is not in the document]
urso: Yeah, an extra check would make sense. As queue is not encrypted there is indeed the chance of it holding sensitive data.

libbeat/publisher/queue/spool/outbroker.go
urso: Question is how often this could happen. Only reason ACKing fails is due to disk/write failures (network shares).

With exponential backoff we might want to introduce more configurable parameters. I'd prefer not to introduce additional config parameters for dealing with this kind of fatal disk/io issues.

libbeat/publisher/queue/spool/codec.go
urso: btw. the codec setting is not documented for now.

libbeat/_meta/config.reference.yml
urso: Done.

libbeat/publisher/queue/spool/spool.go
urso: Haha... handlers is not used anymore for quite some time -> removing it \o/

ph:
> The spaces in logp.Debug are on purpose. It's like a 'multiline' log. The indentation helps me a little to keep context. :/
 
@urso Using a contextual logger wouldn't help in that case?

```golang
l := logp.NewLogger("spool").With("ABC", "WYZ")
lo.Debugw("something wrong", "abc", 1234)
```
:(

ph:
Some files need to be generated:
```
NOTICE.txt: needs update
auditbeat/auditbeat.reference.yml: needs update
filebeat/filebeat.reference.yml: needs update
heartbeat/heartbeat.reference.yml: needs update
metricbeat/metricbeat.reference.yml: needs update
packetbeat/docs/fields.asciidoc: needs update
packetbeat/packetbeat.reference.yml: needs update
winlogbeat/winlogbeat.reference.yml: needs update
make: *** [check] Error 1
```

libbeat/publisher/queue/spool/inbroker.go
```diff
@@ -0,0 +1,482 @@
...
+		st := &states[i]
+		if st.state == nil {
+			continue
+		}
+
+		count := (st.seq - st.state.lastACK)
+		if count == 0 || count > math.MaxUint32/2 {
+			// seq number comparison did underflow. This happens only if st.seq has
+			// already been acknowledged
+			// log.Debug("seq number already acked: ", st.seq)
```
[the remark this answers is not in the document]
urso: It's some kind of tracing (we don't have trace level in beats). For debugging locally it's super helpful, but otherwise it's much to noisy for users (every event ever published will trigger a log line).

libbeat/publisher/queue/spool/inbroker.go
```diff
@@ -0,0 +1,482 @@
...
+	select {
+	case <-b.ctx.Done():
+		return false
+
+	case req := <-b.events:
+		log.Debug("inbroker (stateEmpty): new event")
+
+		buf, st, err := b.encodeEvent(&req)
+		if err != nil {
+			log.Debug("  inbroker (stateEmpty): encode failed")
```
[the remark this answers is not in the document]
urso: I added the space on purpose. It's kind of a multiline log on the current inbroker state.

libbeat/publisher/queue/spool/inbroker.go
```diff
@@ -0,0 +1,482 @@
...
+
+func (b *inBroker) handleCancel(req *producerCancelRequest) {
+	// mark state as cancelled, so to not accept any new events
+	// from the state object.
+	if st := req.state; st != nil {
+		st.cancelled = true
+	}
+
+	if req.resp != nil {
+		req.resp <- producerCancelResponse{removed: 0}
```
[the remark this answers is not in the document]
urso: allocate? The response is by value. You mean providing a const?

libbeat/publisher/queue/spool/log.go
```diff
@@ -0,0 +1,44 @@
...
+	Error(...interface{})
+	Errorf(string, ...interface{})
+}
+
+type outLogger struct {
+}
+
+var defaultLogger = logp.NewLogger("spool")
+
+// var defaultLogger = (*outLogger)(nil)
```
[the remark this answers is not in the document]
urso: I'd keep it. Very helpful when debugging e.g. a panic. Normal logs can be swallowed by the testing framework -> never printed on panic

libbeat/publisher/queue/spool/outbroker.go
urso: It's named `errRetry`, as it's purpose is to mostly act as a signal to the consumer to retry (ignore error). Like, we didn't get an event for you, please retry. The retry signal is required to guarantee a consumer can be correctly closed asynchronously.

ph:
> Also note, the package provides it's local logger, not relying on logp, which pretty much leaks the implementation of zap (sugar) implementation. The state machine is in the fast path, being executed on every single event. I'd rather prefer not to allocate another temporary context per event.

Make sense, anyway its more for debugging so I think its _OK_ in that case.

ph:
`make check` fails with an asciidoc from packetbeat, are you missing a commit?

auditbeat/auditbeat.reference.yml
ruflin: I don't think we should introduce a type because of the reasons mentioned by @urso. There can only be one queue at the time. 

For me spooling can be to disk or memory. Historically we had a spooler in filebeat which kept the data in memory. Other options would be to call it `file`?

+1 on the proposal from @ph about the priority. FIFO or LIFO is a config option of the queue. It can mean in the background a completely different implementation but the user should not have to worry about that.

auditbeat/auditbeat.reference.yml
ruflin: I mainly brought it up as people keep trying to use 2 outputs in Beats and they will for queues. So having it documented makes it easy to point people to it. So they know it already from the docs before running the docs.

libbeat/publisher/queue/spool/codec.go
ruflin: For not documenting it: If the config is there, it should be in the reference file. We did this "trick" in the past and I think it didn't benefit us as people discovered the option and started using it. Better have it in the reference file as experimental and say what it does.

If I understand the comment from ph a user can have code A and then configure codec B on restart and everything will keep working smootly? The events are read in with codec A and the next time the file is written, it will be all in codec B.

urso:
@ph Fixed filebeat hanging. Fix with details is in [this commit](https://github.com/elastic/beats/pull/6581/commits/8f98284bf68f83c1f5b11a8d6b71a70b1b466803). Plus [safety-net commit](https://github.com/elastic/beats/pull/6581/commits/985e474280808b29b66a17e9a240e049067f75c3), in case user set flush events to -1.

auditbeat/auditbeat.reference.yml
urso: Regarding priority I was initially thinking the same. But if we have a priority setting, user must be allowed to change the setting between restarts.

But async ACK behaviour of the queue makes the on-disk structure a little more complicated. When introducing stack like functionality, we will end up with a many holes. That is, freeing space will be somewhat more complicated in the LIFO case. I'd like to solve the LIFO case separately, potentially merging both cases into a common file format, later in time. Priority based queue, using heaps might become even more complicated.

libbeat/publisher/queue/spool/codec.go
urso: Added `codec` setting to the reference file.

> If I understand the comment from ph a user can have code A and then configure codec B on restart and everything will keep working smootly? The events are read in with codec A and the next time the file is written, it will be all in codec B.

The codec can be change between restarts. The setting is only used by the write. The reader can easily deal with all codecs within a file.

auditbeat/auditbeat.reference.yml
ph: @urso Look like a good compromise.

@ruflin You have a point, Disk or File at the discretion of @urso.

libbeat/publisher/queue/spool/outbroker.go
ph: OK, good for now.

ph:
I a few tests concerning when ES was unreachable and recovering in a middle of a read, I didn't see any problem there.

ph:
Found an issue with the `Logstash output` which might be linked to issue with the registry.

## Scenario

### Spooling to disk is off.
- Use the Logstash output without SSL default settings (pipelining: 2)
- FB Read a 500 000 events files. (EOF is reached)
- Check Elasticsearch 500 000 events are present.

### Spooling to disk is off.
- Use the Logstash output without SSL default settings (pipelining: 2)
- FB Read a 500 000 events files. (EOF is reached)
- Check Elastricsearch only ~31K events are present.
- No more events are sent

ph:
Same issue with redis, I would expect kafka to have the same problem. goroutine dump at https://gist.github.com/ph/410ccdbe48cd00cd42603a0ec47cdc6a

urso:
Checking the stack-trace + filebeat offset update problems, it was a missing signal from the queue to the publisher pipeline that the file has been flushed. Signaling is done via channels. After a few flushes the signalling channels have been filled up and the producers ACK handling has deadlocked.

ph:
@urso more testing Logstash with recovering and killing instance. It worked, I had some duplicates,  which are expected,  but didn't lose events.

Can we cleanup this log statement a bit, I think we should have a space between collected and 2048.
```
2018-04-03T15:56:47.446-0400    DEBUG   [spool] spool/outbroker.go:239    outbroker (stateActive): events collected2048 2048 <nil>
```

Also, the registry is now correctly updated, and logs show correctly.

urso:
Couldn't reproduce stress tests failing. I'm still investigating. CI should be at least somewhat stable before merging.

ph:
Amazing work @urso ! waiting on Jenkins green.

ph:
One of the metricbeat testsuite failed on Travis related to a MySQL test, it just took too much time to execute and the job was killed by their watchdog. Its the first time I've seen this error, I believe it's a red herring and I've restarted the job.

PR #6587, opened 2018-03-17

tsg:
There were no conflicts.


tsg:
@andrewkroh @adriansr Both Jenkins and Travis failed in the Auditbeat tests, while downloading ES. Weirdly, though, that didn't fail in libbeat or for the other Beats.

ruflin:
I think this is related to the dashboard tests. Metricbeat and Auditbeat are the only once that use `snapshot-noxpack`. But as I realised earlier Auditbeat is actually the only beat that depends on it as Metricbeat has it's own Docker files for Kibana and Metricbeat and does not use the SNAPSHOTS.

It seems the problem is that the ES Snapshotbuilds for 6.3 are not available yet: `https://snapshots.elastic.co/downloads/elasticsearch/elasticsearch-6.3.0-SNAPSHOT.tar.gz.sha512`. We could change the environment to test against in auditbeat to make it working.

tsg:
@ruflin Thanks for looking into it. I'm thinking we should merge this and solve the testing issue in a new PR, so that it's easier to track and back out when the time comes.

ruflin:
@tsg I will open a PR against 6.x with a potential fix.

tsg:
Thank you!

PR #6589, opened 2018-03-18
Mark the system.syslog.message field as text

tsg:
It was marked as a keyword, but that was a bug, because this field is
long and generally textual. I went for a breaking change instead of
making it a multi-field, because I don't see the use case of heaving
it as a keyword.


tsg:
Issue was reported here: https://discuss.elastic.co/t/cannot-search-var-log-messages-without-wildcard-before-string/123952/6

PR #6590, opened 2018-03-19
Fix docker autodiscover hint generation autodiscover logging

vjsamuel:
In order to allow hint generation, there needs to be a MapStr of labels that isnt nested. This PR fixes that along with some logging.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -91,6 +91,7 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
 - Fix infinite failure on Kubernetes watch {pull}6504[6504]
 - Ensure Kubernetes labels/annotations don't break mapping {pull}6490[6490]
 - Report ephemeral ID and uptime in monitoring events on all platforms {pull}6501[6501]
+- Fix docker autodiscover hint generation and autodiscover logging {pull}6590[6590]
 
 *Auditbeat*
 
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -91,7 +91,6 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
 - Fix infinite failure on Kubernetes watch {pull}6504[6504]
 - Ensure Kubernetes labels/annotations don't break mapping {pull}6490[6490]
 - Report ephemeral ID and uptime in monitoring events on all platforms {pull}6501[6501]
-- Fix docker autodiscover hint generation and autodiscover logging {pull}6590[6590]
 
 *Auditbeat*
 
--- a/libbeat/autodiscover/builder/plugin.go
+++ b/libbeat/autodiscover/builder/plugin.go
@@ -23,7 +23,7 @@ func init() {
 	p.MustRegisterLoader(pluginKey, func(ifc interface{}) error {
 		b, ok := ifc.(builderPlugin)
 		if !ok {
-			return errors.New("plugin does not match processor plugin type")
+			return errors.New("plugin does not match builder plugin type")
 		}
 
 		return autodiscover.Registry.AddBuilder(b.name, b.builder)
--- a/libbeat/autodiscover/providers/docker/docker.go
+++ b/libbeat/autodiscover/providers/docker/docker.go
@@ -119,12 +119,8 @@ func (d *Provider) emitContainer(event bus.Event, flag string) {
 	}
 
 	labelMap := common.MapStr{}
-	labels := common.MapStr{}
 	for k, v := range container.Labels {
 		safemapstr.Put(labelMap, k, v)
-
-		// Needed for hint processing
-		labels[k] = v
 	}
 
 	meta := common.MapStr{
@@ -135,15 +131,12 @@ func (d *Provider) emitContainer(event bus.Event, flag string) {
 			"labels": labelMap,
 		},
 	}
-	dockerMeta := meta.Clone()
-	dockerMeta.Put("container.labels", labels)
-
 	// Without this check there would be overlapping configurations with and without ports.
 	if len(container.Ports) == 0 {
 		event := bus.Event{
 			flag:     true,
 			"host":   host,
-			"docker": dockerMeta,
+			"docker": meta,
 			"meta": common.MapStr{
 				"docker": meta,
 			},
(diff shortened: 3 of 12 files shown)


libbeat/autodiscover/builder/plugin.go
```diff
@@ -23,7 +23,7 @@ func init() {
 	p.MustRegisterLoader(pluginKey, func(ifc interface{}) error {
 		b, ok := ifc.(builderPlugin)
 		if !ok {
-			return errors.New("plugin does not match processor plugin type")
+			return errors.New("plugin does not builder processor plugin type")
```
ruflin: does not have builder ...?

libbeat/autodiscover/providers/docker/docker.go
```diff
@@ -51,16 +51,16 @@ func AutodiscoverBuilder(bus bus.Bus, c *common.Config) (autodiscover.Provider,
 	var builders autodiscover.Builders
 	for _, bcfg := range config.Builders {
 		if builder, err := autodiscover.Registry.BuildBuilder(bcfg); err != nil {
-			logp.Warn("docker", "failed to construct autodiscover builder due to error: %v", err)
+			logp.Warn("failed to construct autodiscover builder due to error: %v", err)
```
ruflin: I know it was already there but I wonder if we should make this either error or info as we try not to use `Warn`. The problem with warn is that it's unclear if this is just an "Info" or needs action in which case it would be an error.

libbeat/autodiscover/providers/docker/docker.go
```diff
@@ -51,16 +51,16 @@ func AutodiscoverBuilder(bus bus.Bus, c *common.Config) (autodiscover.Provider,
...
 			builders = append(builders, builder)
 		}
 	}
 
 	var appenders autodiscover.Appenders
-	for _, acfg := range config.Builders {
+	for _, acfg := range config.Appenders {
 		if appender, err := autodiscover.Registry.BuildAppender(acfg); err != nil {
-			logp.Warn("docker", "failed to construct autodiscover appender due to error: %v", err)
+			logp.Warn("failed to construct autodiscover appender due to error: %v", err)
```
ruflin: See note about warn above

libbeat/autodiscover/providers/docker/docker.go
```diff
@@ -51,16 +51,16 @@ func AutodiscoverBuilder(bus bus.Bus, c *common.Config) (autodiscover.Provider,
...
-			logp.Warn("docker", "failed to construct autodiscover builder due to error: %v", err)
+			logp.Warn("failed to construct autodiscover builder due to error: %v", err)
 		} else {
 			builders = append(builders, builder)
 		}
 	}
 
 	var appenders autodiscover.Appenders
-	for _, acfg := range config.Builders {
+	for _, acfg := range config.Appenders {
```
ruflin: Was that a bug?

libbeat/autodiscover/providers/docker/docker.go
```diff
@@ -119,8 +119,12 @@ func (d *Provider) emitContainer(event bus.Event, flag string) {
 	}
 
 	labelMap := common.MapStr{}
+	labels := common.MapStr{}
```
ruflin: is this a `map[string]string`? Perhaps that would make it more obvious that it is not used differently?

libbeat/autodiscover/providers/kubernetes/kubernetes.go
```diff
@@ -64,7 +64,7 @@ func AutodiscoverBuilder(bus bus.Bus, c *common.Config) (autodiscover.Provider,
 	var builders autodiscover.Builders
 	for _, bcfg := range config.Builders {
 		if builder, err := autodiscover.Registry.BuildBuilder(bcfg); err != nil {
-			logp.Warn("kubernetes", "failed to construct autodiscover builder due to error: %v", err)
+			logp.Warn("failed to construct autodiscover builder due to error: %v", err)
```
ruflin: About warn, see above. Applies also below :-)

libbeat/autodiscover/builder/plugin.go
vjsamuel: reflects sleep deprivation :) fixing. apologies for poor grammar

libbeat/autodiscover/providers/docker/docker.go
vjsamuel: yes.

libbeat/autodiscover/providers/docker/docker.go
vjsamuel: we have standardized on `MapStr` for labels/annotations/hints on all of the autodiscover related code. I kept it the same for consistency sake. @exekias thoughts?

libbeat/autodiscover/providers/docker/docker.go
vjsamuel: i would like to keep those failure messages as just `Info` as they dont impact the coming up of the provider.

ruflin:
jenkins, test it

libbeat/autodiscover/providers/docker/docker.go
exekias: I'm wondering if we should change `GenerateHints` instead, to take MapStr nesting into account. We are enforcing MapStr nesting to avoid mapping issues in Elasticsearch, so this will be a pattern. Having this in the "unsafe" way may lead to different results, what do you think?

exekias:
I'm sorry, I introduced the breaking change and missed this, we should add some integration tests around hints based autodiscover.

I'm guessing kubernetes provider is also affected, let me know what do you think about updating `GenerateHints` instead, I think it should be safer in the long run (`MapStr.Get` works well with both nested and plain representations)

vjsamuel:
@exekias done

CHANGELOG.asciidoc
```diff
@@ -91,6 +91,7 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
 - Fix infinite failure on Kubernetes watch {pull}6504[6504]
 - Ensure Kubernetes labels/annotations don't break mapping {pull}6490[6490]
 - Report ephemeral ID and uptime in monitoring events on all platforms {pull}6501[6501]
+- Fix docker autodiscover hint generation and autodiscover logging {pull}6590[6590]
```
exekias: no need for a changelog entry

exekias:
ok to test

vjsamuel:
@exekias removed change log. thanks for the quick review

exekias:
Failing test is not related

PR #6591, opened 2018-03-19
[Filebeat] - Add raw json field when unmarshaling fails

### LINKED ISSUE #6516: [Filebeat] Data loss when JSON decoding fails
When JSON decoding fails Filebeat should include the raw data in the `message` field.

```
filebeat.prospectors:
- paths:
  - input.json
  json.keys_under_root: true
  json.add_error_key: true

output.console.enabled: true
```

input.json
```
Hello world!
{bad: json}
{"hello": "world"}
```

```json
ata": {
    "beat": "filebeat",
    "type": "doc",
    "version": "6.2.2"
  },
  "source": "/Users/akroh/Downloads/filebeat-6.2.2-darwin-x86_64/input.json",
  "offset": 13,
  "error": {
    "message": "Error decoding JSON: invalid character 'H' looking for beginning of value",
    "type": "json"
  },
  "beat": {
    "name": "macbook",
    "hostname": "macbook",
    "version": "6.2.2"
  }
}
{
  "@timestamp": "2018-03-08T21:52:43.891Z",
  "@metadata": {
    "beat": "filebeat",
    "type": "doc",
    "version": "6.2.2"
  },
  "source": "/Users/akroh/Downloads/filebeat-6.2.2-darwin-x86_64/input.json",
  "offset": 25,
  "error": {
    "type": "json",
    "message": "Error decoding JSON: invalid character 'b' looking for beginning of object key string"
  },
  "beat": {
    "name": "macbook",
    "hostname": "macbook",
    "version": "6.2.2"
  }
}
{
  "@timestamp": "2018-03-08T21:52:43.891Z",
  "@metadata": {
    "beat": "filebeat",
    "type": "doc",
    "version": "6.2.2"
  },
  "offset": 44,
  "hello": "world",
  "source": "/Users/akroh/Downloads/filebeat-6.2.2-darwin-x86_64/input.json",
  "beat": {
    "name": "macbook",
    "hostname": "macbook",
    "version": "6.2.2"
  }
}

```

crazyvikas23:
Fixes #6516.


ruflin:
jenkins, test it

crazyvikas23:
@andrewkroh I think you are write, `message` key inside `error` should only contain error message. I have added raw json data in message field outside.

filebeat/harvester/reader/json.go
```diff
@@ -125,3 +127,10 @@ func MergeJSONFields(data common.MapStr, jsonFields common.MapStr, text *string,
 	}
 	return time.Time{}
 }
+
+func appendRawJSONData(jsonFields map[string]interface{}, text []byte) {
+	if jsonFields == nil {
+		jsonFields = common.MapStr{}
+	}
+	jsonFields["message"] = fmt.Sprintf("Raw json data: %s", string(text))
```
andrewkroh: I think this is going to produce `json.message` in the final event. It looks to me like the `message.Content` field is being populated with the `text` (raw json) when an error occurs.

So I think something upstream in the harvester needs to change to allow the `fields["message"] = message.Content` to be set. I think the issue lies within this code:

https://github.com/elastic/beats/blob/a93acbd9c1d922f4c9e514ec9a3bf6cfb54c0ed2/filebeat/input/log/harvester.go#L272-L307

@ruflin I think we need your Filebeat expertise on this modification.

crazyvikas23: I think what you want is to append a message key in the final event. That should be pretty straightforward to implement.

filebeat/harvester/reader/json.go
```diff
@@ -30,7 +30,8 @@ func (r *JSON) decodeJSON(text []byte) ([]byte, common.MapStr) {
 	err := unmarshal(text, &jsonFields)
 	if err != nil || jsonFields == nil {
 		if !r.cfg.IgnoreDecodingError {
-			logp.Err("Error decoding JSON: %v", err)
+
+			logp.Err("Error decoding JSON: %v. Raw data: %s", err, string(text))
```
andrewkroh: Maybe we shouldn't put this raw message into the error logs. I'm worried about sensitive information being leaked inside of the Filebeat logs.

crazyvikas23:
@andrewkroh I don't think we can write a test case for same in `json_test.go` because file it contains unit test cases for json decoding, which already works fine. We should take this as integration test elsewhere.

andrewkroh:
jenkins, test it

andrewkroh:
jenkins, test it

PR #6596, opened 2018-03-19
Fix docker metrics discovery by converting port to int64

vjsamuel:
The metrics builder requires port number for being able to spin up metricbeat modules that poll. Kubernetes uses port as int64 and docker uses uint16 for the same. There needs to be a consistent notation of ports for us to be able to use them in the metrics builder.


exekias:
I think it will be better to cast in the side reading this value, something similar to https://github.com/elastic/beats/blob/master/libbeat/processors/add_docker_metadata/add_docker_metadata.go#L248

That should avoid this error in future providers

vjsamuel:
does it make sense to extract this method into a common location? Its being used in multiple places.

exekias:
I would be ok with that, what do you think @andrewkroh?

andrewkroh:
I'm good with moving [`tryToInt`](https://github.com/elastic/beats/blob/972fb3b6ec3dd0cf2dc60a7454900822682b593a/libbeat/processors/add_docker_metadata/add_docker_metadata.go#L248:6) to a location that supports reuse.

Without a description or a link to a bug report I'm having trouble understanding what this pull request fixes or what the root cause of the problem is. Can you please add more information to the description.

vjsamuel:
@andrewkroh updated description. apologies for the lack of clarity. what would be a suitable place to place this method?

andrewkroh:
> what would be a suitable place to place this method?

Thanks for adding the description. In common is good. I'm thinking instead of `int.go` it could be `coerce.go` because we also have a function named `tryToMapStr` that could be moved there in the future. It has been duplicated a few times.

vjsamuel:
@andrewkroh @exekias i have done the renaming.

andrewkroh:
jenkins, test it

exekias:
I think there is no need for changelog, this was affecting unreleased code only

PR #6602, opened 2018-03-20
[6.x] Configuration to allow users to retry loading dashboards  (#6560)

jalvz:
Backports the following commits to 6.x:
 - Configuration to allow users to retry loading dashboards   (#6560)


jalvz:
@ruflin i would be great if you could merge this after ci run, so we can backport it also in apm-server and do some tests with 6.x even before FF

ruflin:
@jalvz Thanks for the backport. We do bulk backports so no need to manually backport. I will check that we can make one other bulk backport soonish so you can test this. Going to close this one.

Issue #6442, opened 2018-02-22
Metricbeat exchanges and queues' message rates (enhancement request)


CodingSpiderFox:
From https://discuss.elastic.co/t/metricbeat-rabbitmq-collect-message-rates/120910 

It would be great to also have the incoming/outgoing message rates fields collected by metricbeat for both queues and exchanges.

This way, I could monitor whether someone is still sending data over my RMQ instance.

CodingSpiderFox:
I try to figure out how the schema works. 

I found these lines in the metricbeat/module/rabbitmq/queue/data.go file:

```` 
var (
	schema = s.Schema{
		"name":        c.Str("name"),
		"vhost":       c.Str("vhost"),
		"durable":     c.Bool("durable"),
		"auto_delete": c.Bool("auto_delete"),
		"exclusive":   c.Bool("exclusive"),
		"node":        c.Str("node"),
		"state":       c.Str("state"),
		"arguments": c.Dict("arguments", s.Schema{
			"max_priority": c.Int("x-max-priority", s.Optional),
		}),
		"consumers": s.Object{
			"count": c.Int("consumers"),
			"utilisation": s.Object{
				"pct": c.Int("consumer_utilisation", s.Optional),
			},
		},
		"messages": s.Object{
			"total": s.Object{
				"count": c.Int("messages"),
			},
			"ready": s.Object{
				"count": c.Int("messages_ready"),
			},
			"unacknowledged": s.Object{
				"count": c.Int("messages_unacknowledged"),
			},
			"persistent": s.Object{
				"count": c.Int("messages_persistent"),
			},
		},
		"memory": s.Object{
			"bytes": c.Int("memory"),
		},
		"disk": s.Object{
			"reads": s.Object{
				"count": c.Int("disk_reads"),
			},
			"writes": s.Object{
				"count": c.Int("disk_writes"),
			},
		},
	}
)
```

I'm not sure how to get the data. In the result from the rabbitmq API, the "rate" keys do not have unique names for all of the 3 occurences:

```

[...]

PR #6607, opened 2018-03-20
Rabbitmq exchanges metricset

kvalev:
This pull requests add a new metricset for RabbitMQ exchanges.

Related to #6442


metricbeat/docs/modules/rabbitmq.asciidoc
```diff
@@ -21,7 +21,7 @@ in <<configuration-metricbeat>>. Here is an example configuration:
 ----
 metricbeat.modules:
 - module: rabbitmq
-  metricsets: ["node", "queue", "connection"]
+  metricsets: ["node", "queue", "connection", "exchange"]
```
ruflin: Reminds me we should sort metricsets alphabetical (I can do that in a follow up PR).

metricbeat/module/rabbitmq/exchange/_meta/data.json
```diff
@@ -0,0 +1,19 @@
+{
```
ruflin: Any chance you could create a data.json here? The command you need is as following:
```
go test -tags=integration github.com/elastic/beats/metricbeat/module/rabbitmq/exchange/... -data
```

It assumes you have rabbitmq running on localhost.

metricbeat/module/rabbitmq/exchange/data.go
```diff
@@ -0,0 +1,96 @@
...
+		"arguments":   c.Dict("arguments", s.Schema{
+		}),
+		"messages":    c.Dict("message_stats", s.Schema{
+			"publish": s.Object{
+				"count": c.Int("publish"),
+				"details": c.Dict("publish_details", s.Schema{
+					"rate": c.Float("rate"),
+				}),
+			},
+			"publish_in": s.Object{
```
ruflin: We could make the above `publish.total.count` and this one here `publish.in` and the following `publish.out`. Is it correct the the above is just the total?

metricbeat/module/rabbitmq/exchange/exchange_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+	t.Logf("%s/%s event: %+v", f.Module().Name(), f.Name(), event.StringToPrint())
+
+	assert.EqualValues(t, "exchange.name", event["name"])
+	assert.EqualValues(t, "/", event["vhost"])
+	assert.EqualValues(t, true, event["durable"])
+	assert.EqualValues(t, false, event["auto_delete"])
+	assert.EqualValues(t, false, event["internal"])
+
+	messages := event["messages"].(common.MapStr)
```
ruflin: I wonder if for the comparison here you could create a `common.MapStr` with all fields inside with the expected values and then compare it? Would probably make the code a bit more readable and easier to extend.

metricbeat/module/rabbitmq/exchange/data.go
jsoriano: The advantage of keeping it as `publish_in` is that all the groups of metrics under `messages` have the same structure, and they are also kept as a more direct mapping of source metrics.

metricbeat/module/rabbitmq/exchange/_meta/data.json
kvalev: Will do.

metricbeat/module/rabbitmq/exchange/data.go
ruflin: I'm always in favor or having all the related metrics under on group but it's really hard to have a clear rule here. I'm good with both.

kvalev: TBH I am not sure what "publish" is supposed to be. The [official documentation](https://rawcdn.githack.com/rabbitmq/rabbitmq-management/rabbitmq_v3_6_11/priv/www/doc/stats.html) says "Count of messages published.", but in my environment it is always 0, although publish_in and publish_out are both 0+.
Regarding the naming - it is up to you as I am not familiar with the metric set naming guidelines.

metricbeat/module/rabbitmq/exchange/_meta/data.json
kvalev: This command does not generate the full json file (the message stats are missing). Am I missing something?

```
{
    "@timestamp": "2017-10-12T08:05:34.853Z",
    "beat": {
        "hostname": "host.example.com",
        "name": "host.example.com"
    },
    "metricset": {
        "host": "localhost:15672",
        "module": "rabbitmq",
        "name": "exchange",
        "rtt": 115
    },
    "rabbitmq": {
        "exchange": {
            "arguments": {},
            "auto_delete": false,
            "durable": true,
            "internal": false,
            "name": "",
            "type": "direct",
            "vhost": "/"
        }
    }
}
```

CodingSpiderFox:
@kvalev will you find the time to finish this PR in the near future? I need this feature soon

ruflin:
@kvalev I case you can't find the time to finish this PR, lets us know so we can take it over (or perhaps @CodingSpiderFox? ) to get it merged.

metricbeat/module/rabbitmq/exchange/exchange_test.go
kvalev: I tried doing that, but the Equals comparison fails for some reason, although the maps are identical (albeit in a different order): 
```
	Error:      	Not equal:
	            	expected: common.MapStr{"return_unroutable":common.MapStr{"count":40, "details":common.MapStr{"rate":123.123}}, "redeliver":common.MapStr{"count":30, "details":common.MapStr{"rate":0.123}}, "publish":common.MapStr{"count":123, "details":common.MapStr{"rate":0.1}}, "publish_in":common.MapStr{"count":100, "details":common.MapStr{"rate":0.5}}, "publish_out":common.MapStr{"count":99, "details":common.MapStr{"rate":0.9}}, "ack":common.MapStr{"count":60, "details":common.MapStr{"rate":12.5}}, "deliver_get":common.MapStr{"details":common.MapStr{"rate":43.21}, "count":50}, "confirm":common.MapStr{"count":120, "details":common.MapStr{"rate":98.63}}}
	            	actual  : common.MapStr{"ack":common.MapStr{"count":60, "details":common.MapStr{"rate":12.5}}, "deliver_get":common.MapStr{"count":50, "details":common.MapStr{"rate":43.21}}, "confirm":common.MapStr{"count":120, "details":common.MapStr{"rate":98.63}}, "return_unroutable":common.MapStr{"count":40, "details":common.MapStr{"rate":123.123}}, "redeliver":common.MapStr{"count":30, "details":common.MapStr{"rate":0.123}}, "publish":common.MapStr{"count":123, "details":common.MapStr{"rate":0.1}}, "publish_in":common.MapStr{"count":100, "details":common.MapStr{"rate":0.5}}, "publish_out":common.MapStr{"details":common.MapStr{"rate":0.9}, "count":99}}

```

CodingSpiderFox:
I'm not good doing git stuff. Still learning. How can I do the rebase? Fork this repo, pull the branch kvalev and push it to my repo, rebase it on master and then open another PR?

kvalev:
@ruflin I verified it again against my local RabbitMQ instance and it seems that in case the exchanges were not used recently (probably since last restart), the `messages_stats` section is missing. Even if there were some messages, not all subsection will be returned by the RabbitMQ - only the ones that are not 0. This is also the reason why the data.json generation returns a very basic data skeleton. If I understand it correctly, the data generation uses the Dockerfile to start a RabbitMQ instance and fetch the data from there. However given that the instance was just started it does not report any stats back. If I publish some test messages (using `rabbitmqadmin`), parts of the stats sections are present. Should I change the Dockerfile to add some test data, so that the data.json generation returns something meaningful? I dont know whether this will break something else.
I also made all exchange stat sections optional - I hope this was the right way to do it.

jsoriano:
jenkins, test this please

ruflin:
@kvalev I like the idea of populating some data before the test. It could also be part of the test instead of the docker image, like this it also work if done with a local instance.

CodingSpiderFox:
> @kvalev I like the idea of populating some data before the test. It could also be part of the test instead of the docker image, like this it also work if done with a local instance.

So what is the actual TODO here?

jsoriano:
I have created PR #6955 to resolve pending issues

ruflin:
Closing as https://github.com/elastic/beats/pull/6955 was merged. @kvalev Thanks a lot for your work and kicking this off.

PR #6608, opened 2018-03-20
Metricbeat: Use system-wide ticks to calculate docker cpu usage

jsoriano:
This implementation looks for a better coherence with `docker stats`.

It is based on the ratio between the number of CPU ticks of the container and the number of ticks of the whole system during a fixed period of time, as `docker stats` does.

Some documentation texts have been rewritten ~~, and scaling factor of percentages has been increased, as cpu percentage can be over 100, up to 100 times the number of cores (as in `top`, `docker stats` or similar tools).~~

I think we might deprecate, or move to other part the "`system`" CPU metric. It is used to calculate percentages of metrics for containers, but I don't think it is very useful beyond that. Also, with current implementation its percentage should always constant, as it is calculated as a ratio with itself.

### DIFF
--- a/metricbeat/module/docker/cpu/_meta/fields.yml
+++ b/metricbeat/module/docker/cpu/_meta/fields.yml
@@ -8,15 +8,16 @@
       type: scaled_float
       format: percentage
       description: >
-        The system kernel consumed by the Docker server.
+        Percentage of time in kernel space.
     - name: kernel.ticks
       type: long
       description: >
-        CPU kernel ticks.
+        CPU ticks in kernel space.
     - name: system.pct
       type: scaled_float
       format: percentage
       description: >
+        Percentage of total CPU time in the system.
     - name: system.ticks
       type: long
       description: >
@@ -25,10 +26,11 @@
       type: scaled_float
       format: percentage
       description: >
+        Percentage of time in user space.
     - name: user.ticks
       type: long
       description: >
-        CPU user ticks
+        CPU ticks in user space.
     - name: total.pct
       type: scaled_float
       format: percentage
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -69,6 +69,7 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
 - Add config option for windows/perfmon metricset to ignore non existent counters. {pull}6432[6432]
 - Support apache status pages for versions older than 2.4.16. {pull}6450[6450]
 - Add support for huge pages on Linux. {pull}6436[6436]
+- Refactor docker CPU calculations to be more consistent with `docker stats`. {pull}6608[6608]
 
 *Packetbeat*
 
--- a/metricbeat/docs/fields.asciidoc
+++ b/metricbeat/docs/fields.asciidoc
@@ -2150,7 +2150,7 @@ type: scaled_float
 
 format: percentage
 
-The system kernel consumed by the Docker server.
+Percentage of time in kernel space.
 
 
 [float]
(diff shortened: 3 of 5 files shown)


exekias:
The reasoning looks good to me, have a look at tests, something is wrong there:
```
# github.com/elastic/beats/metricbeat/module/docker/cpu
module/docker/cpu/cpu_test.go:46:10: undefined: perCpuUsage
module/docker/cpu/cpu_test.go:73:10: undefined: totalUsage
module/docker/cpu/cpu_test.go:96:10: undefined: usageInKernelmode
module/docker/cpu/cpu_test.go:119:10: undefined: usageInUsermode
```

metricbeat/module/docker/cpu/_meta/fields.yml
```diff
@@ -6,31 +6,37 @@
   fields:
     - name: kernel.pct
       type: scaled_float
+      scaling_factor: 1000000
```
ruflin: I remember a scaling factor > 1000 will not have any compression advantage anymore over just using float instead of scaled_float. So I suggest we just change to float here. Also below.

ruflin:
This also cleans up the implementation a bit. A CHANGELOG should be added for this change.

Perhaps you can add a few more details in the PR message about the previous value compare to the current value which is more focused on users coming to this PR to understand the change?

metricbeat/module/docker/cpu/_meta/fields.yml
jsoriano: Umm, if I understood well, `scaled_float` is stored as an integer, wouldn't it be a breaking change to change it to float? (OTOH changing the `scaling_factor` could also be a breaking change)

jsoriano:
Added a line to changelog and more information in the commit message.

metricbeat/module/docker/cpu/_meta/fields.yml
ruflin: As each version has a new template, things should not break here. From a query perspective I think it was already like a float before so having a float afterwards should just work.

If someone ingest with the old template he would just have lower accuracy but not get an error.

@jpountz Is the above correct? The ingestion and quering of `float` and `scaled_float` looks identical?

jpountz: Assuming all values are between 0 and 1, this field will likely require only 20 bits per value, which is still better than a float. However I'm very surprised you need so much precision for a percentage, why are 100 or 1000 not enough?

ruflin:
@jsoriano Needs another rebase.

jsoriano:
Rebased, but let's wait to make a decission on the type of percentage fields before merging.

metricbeat/module/docker/cpu/_meta/fields.yml
jsoriano: @jpountz wdyt? would it be ok to change to `float` for this?

jpountz: The fact it may go over 1 is not an issue, I still think scaled_float is a better fit. However let's use a more reasonable scaling factor like 100?

jpountz: Oh now I wonder that you might have interpreted the scaling factor as an upper bound of the values, while it's not. It just means that Elasticsearch will store `x * scaling_factor` as a `long` internally.

jsoriano: Oh, ok, understood now :) Then 100/1000 will be probably enough, yes. Thanks @jpountz !!

I'll leave it empty as it was before, default seems to be 1000.

jsoriano:
Changes to `scaling_factor` in percentages removed.

ruflin:
@jsoriano Could you update the PR description with details on what the logic is before and what it is now? So users coming here to figure out how their values changed get it without reading the code.

jsoriano:
PR description updated.

PR #6609, opened 2018-03-20
Stop if autodiscover configuration is wrong

exekias:
This change stops the beat if the configuration for builders or appenders is wrong.

I also added a new integration test on label based autodiscover


metricbeat/tests/system/test_autodiscover.py
```diff
@@ -53,3 +53,42 @@ def test_docker(self):
...
+        proc = self.start_beat()
+        docker_client.images.pull('memcached:1.5.3')
+        labels = {
+            'co.elastic.metrics/module': 'memcached',
+            'co.elastic.metrics/hosts': "'${data.host}:11211'",
+        }
+        container = docker_client.containers.run('memcached:1.5.3', labels=labels, detach=True)
+
+        self.wait_until(lambda: self.log_contains('Autodiscover starting runner: memcached'))
+        sleep(2)
```
ruflin: sleeps in tests :-(

PR #6614, opened 2018-03-21
Update hosts example

artonge:
I add to format the host like this so go-sql-driver could parse it correctly.


exekias:
Thank you for fixing this @artonge :smile:

exekias:
ok to test

PR #6621, opened 2018-03-21
Rename metrics/logs builders to `hints`

exekias:
This change renames current `logs` and `metrics` builders to `hints`. These builders are beat specific and are minded to create configs based on hints (from labels or annotations). Config would then look like this:

```
 metricbeat.autodiscover:
   providers:
     - type: kubernetes
       builders:
        - type: hints
```

It also introduces a shortcut to configure just the `hints` builder, which should cover most use cases for common users:

```
metricbeat.autodiscover:
  providers:
    - type: kubernetes
      hints.enabled: true
```

No changelog is needed as this is updating hints based autodiscover, which is unreleased


exekias:
Yes, if you want advanced use cases like that one, you can just remove `hints.enabled` and define the set of builders. I've added an "Advanced usage" topic inside autodiscover docs for that

filebeat/docs/autodiscover-hints.asciidoc
```diff
@@ -38,8 +37,18 @@ include_lines
 
 [float]
 ==== Docker
-Docker allows providing hints through Docker labels. When containers are spun up the Builder understands
-labels that convey hints to convey additional configuration to the input.
+
+Docker autodiscover provider supports hints too, it's enabled in the same way:
```
ruflin: I would not use `too` here as people might jump to this section of the docs directly.

libbeat/autodiscover/providers/docker/docker.go
```diff
@@ -48,22 +48,14 @@ func AutodiscoverBuilder(bus bus.Bus, c *common.Config) (autodiscover.Provider,
...
 	}
 
-	var builders autodiscover.Builders
-	for _, bcfg := range config.Builders {
-		builder, err := autodiscover.Registry.BuildBuilder(bcfg)
-		if err != nil {
-			return nil, err
-		}
-		builders = append(builders, builder)
+	builders, err := autodiscover.NewBuilders(config.Builders, config.HintsEnabled)
```
ruflin: Nice simplification of the code.

libbeat/docs/shared-autodiscover.asciidoc
```diff
@@ -119,35 +114,52 @@ contain variables from the autodiscover event. They can be accessed under data n
...
 
+[[configuration-autodiscover-hints]]
+=== Hints based autodiscover
+
+include::../../{beatname_lc}/docs/autodiscover-hints.asciidoc[]
+
+[[configuration-autodiscover-advanced]]
+=== Advanced usage
+
+////
```
ruflin: Leftover comment?

metricbeat/tests/system/test_autodiscover.py
```diff
@@ -86,9 +86,4 @@ def test_docker_labels(self):
         container.stop()
         self.wait_until(lambda: self.log_contains('Autodiscover stopping runner: memcached'))
 
-        output = self.read_output_json()
```
ruflin: Is this check not needed anymore?

libbeat/docs/shared-autodiscover.asciidoc
exekias: I left this commented as I expect new builders coming soon, they will be documented there (including hints) once we have more than just 1.

metricbeat/tests/system/test_autodiscover.py
exekias: I removed it as redundant, but on a second thought, it makes sense to double check here, I'll add it again

exekias: I've been fighting them a little more, hopefully they work now, without sleeps

exekias: It worked :+1:

exekias:
jenkins, retest this please

Issue #96, opened 2015-03-02
Restricting the size of MySQL query 


srikanthBezawada:
Hi,

Is there any way to restrict the size of MySQL query while sending to elastic search ?

Thanks,
Srikanth.

tsg:
No, sorry, we have some flexibility for that in the HTTP code, but not yet for Mysql. We're doing a code reorganisation in preparation for 0.5.0, which will make it possible to implement these kind of things once and have them automatically for all protocols.

ibrahimsharaf:
Hi @monicasarbu, I want to start working on this one, how can I start?

monicasarbu:
@ibrahimsharaf Great to hear you would like to start working on this one. I do something similar in Packetbeat for the HTTP protocol: https://github.com/elastic/beats/blob/master/packetbeat/protos/http/http.go#L603

ibrahimsharaf:
Hi @monicasarbu, sorry for the delay, would the changes take place in this file (https://github.com/elastic/beats/blob/master/packetbeat/protos/mysql/mysql.go)? is the change about restricting the size of the query in terms of query length?

ak1t0:
Hi @monicasarbu. I would like to work on this issue.

ruflin:
@ak1t0 Great to hear. Feel free to open a PR as soon as you feel comfortable to discuss the code.

onlinejudge95:
@ruflin Any updates on this? I checked the PR associated with this. closed down due to inactivity. I am interested to take it further

jlind23:
Backlog grooming: Stalled issue, closing for now. Feel free to reopen it if needed

ak1t0:
Sorry for very late reply. 
At the time I did not understand anything about this project and issue.
I apologize for the inconvenience and its cost. 
Please close this issue.

:bowing_man:  :bowing_man:  :bowing_man:

PR #6625, opened 2018-03-22
Packetbeat: add function restrict query size

ak1t0:
Fix for https://github.com/elastic/beats/issues/96

- add function to restrict query size

### DIFF
--- a/packetbeat/protos/mysql/mysql.go
+++ b/packetbeat/protos/mysql/mysql.go
@@ -863,6 +863,16 @@ func (mysql *mysqlPlugin) publishTransaction(t *mysqlTransaction) {
 	})
 }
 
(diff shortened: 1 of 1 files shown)


jarrocha:
Hello,
I have two quick questions.

I have been taking a look at old issues around the forum and was wondering why an error like this happen? The Travis CI error is:
`apt-get install failed`
Its details are due to a directory that could not be locked.

On a similar topic, if we introduce a new function/method should we also add unit tests?

faec:
This seems to have stalled in an incomplete state -- is there any interest in reviving it? If not this will be closed soon.

ak1t0:
Sorry for very late reply.
At the time I did not understand anything about this project and issue.
I apologize for the inconvenience and its cost.
Please close this issue.

bowing_man bowing_man bowing_man

PR #6627, opened 2018-03-22
Fail gracefully if X-Pack monitoring is enabled

ruflin:
Before starting to send data to x-pack monitoring a check is made if x-pack monitoring is enabled and data collection is enabled. This makes sure no error a logged in case monitoring was enabled but x-pack does not exist or collection is disabled. Only an info log message during startup will be logged.

This check only happens once on startup. Later it could be modified that the check happens on the connection call back. This change is bigger as it requires a check if the reporter was already started and the moment when the reporter is started would have to change.

This PR has been manually tested against 6.2 but no automated tests were implemented yet.


libbeat/cmd/instance/beat.go
```diff
@@ -318,6 +321,62 @@ func (b *Beat) launch(bt beat.Creator) error {
...
+				Monitoring struct {
+					// This is a string but should be a bool
+					Enabled string `json:"enabled"`
+				} `json:"monitoring"`
+			} `json:"xpack"`
+		} `json:"defaults"`
+	}{}
+	err = json.Unmarshal(body, &defaults)
+	if err != nil {
+		logp.Debug("monitoring", "Error decoding xpack monitoring defaults from Elasticsearch: %s", err)
```
kvch: `%s/xpack monitoring/X-Pack Monitoring/`

libbeat/cmd/instance/beat.go
```diff
@@ -318,6 +321,62 @@ func (b *Beat) launch(bt beat.Creator) error {
 	return beater.Run(&b.Beat)
 }
 
+// checkXPackMonitoring checks if xpack is enabled and collection is enabled
+func (b *Beat) checkXPackMonitoring() bool {
+
+	if b.Config.Output.Name() != "elasticsearch" {
+		logp.Info("Disabling X-Pack monitoring as output is not Elasticsearch.")
```
kvch: `%s/monitoring/Monitoring

libbeat/cmd/instance/beat.go
```diff
@@ -318,6 +321,62 @@ func (b *Beat) launch(bt beat.Creator) error {
...
+		return false
+	}
+
+	esConfig := b.Config.Output.Config()
+	esClient, err := elasticsearch.NewConnectedClient(esConfig)
+	if err != nil {
+		logp.Info("Disabling X-Pack monitoring as connection to Elasticsearch could not be established.")
```
kvch: `%s/monitoring/Monitoring/`

ruflin:
I set to in progress again as I haven't managed to test it yet with X-Pack installed and ingestion disabled.

ruflin:
@kvch @andrewkroh Review comments applied. But please have a second look. The implementation was simplified a bit was decided not to check the collection part (see updated PR message).

Ready for an other round of reviews.

libbeat/cmd/instance/beat.go
```diff
@@ -318,6 +321,41 @@ func (b *Beat) launch(bt beat.Creator) error {
...
+	features := struct {
+		Features struct {
+			Monitoring struct {
+				Enabled bool `json:"enabled"`
+			} `json:"monitoring"`
+		} `json:"features"`
+	}{}
+	err = json.Unmarshal(body, &features)
+	if err != nil {
+		logp.Debug("monitoring", "Error fetching monitoring enabled feature from Elasticsearch: %s", err)
```
kvch: Shouln't this be logged at error level?

ruflin: The idea here was that in all cases where something strange comes back from ES because of Monitoring, not errors are logged. When reading this review I realised there should be an additional error check on line 339 for the error there.

libbeat/cmd/instance/beat.go
```diff
@@ -273,11 +273,14 @@ func (b *Beat) launch(bt beat.Creator) error {
 	}
 
 	if b.Config.Monitoring.Enabled() {
-		reporter, err := report.New(b.Info, b.Config.Monitoring, b.Config.Output)
-		if err != nil {
-			return err
+		reportingActive := b.checkXPackMonitoring()
```
andrewkroh: It seems like this logic is integral to the Elasticsearch monitoring output. So why not have all of the related code to check if x-pack monitoring is enabled  part of the same package?

ruflin: Looking into this I realised we already have an almost identical code here: https://github.com/elastic/beats/blob/master/libbeat/monitoring/report/elasticsearch/client.go#L41 I need to revisit the implementation to see how the two can be combined.

ruflin:
@kvch @andrewkroh The PR is completely rewritten. Please have an other look and have a look again through the PR description if that is also the behaviour you expect.

ruflin:
@andrewkroh in progress label removed. Can be merged.

PR #6631, opened 2018-03-22
http: support broken status line

### LINKED ISSUE #6176: Packetbeat HTTP parsing fails on empty status phrase
An HTTP response whose status phrase is empty causes the following error:

> 2018-01-25T10:50:49Z WARN Failed to understand HTTP response status: 302

For example :
```
HTTP/1.1 302
Location: http://172.17.0.12:8080/resources/userinfo
Content-Length: 0
Date: Thu, 25 Jan 2018 10:50:48 GMT
Connection: close

```

The response parser in https://github.com/elastic/beats/blob/master/packetbeat/protos/http/http_parser.go#L224 is expecting the status-phrase to be non-empty.

From [this question on discuss](https://discuss.elastic.co/t/failed-to-understand-http-response/117036).

adriansr:
User reports some HTTP servers may respond with a broken status line that's missing a space between the status code and the optional status phrase. HTTP clients tested already support this behavior. This patch adjusts the http parser so that this deviation from the standard is accepted.

Fixes #6176


ruflin:
Could you add an entry to the CHANGELOG. Failing tests are not related.

adriansr:
@rufflin done

PR #6634, opened 2018-03-22
Use default metricsets in hints based autodiscover

exekias:
When using hints in Metricbeat, follow this pattern:

 - If `metricsets` hint is given, use it
 - If not, use module default metricsets
 - If no default metricsets, use all module metricsets

### DIFF
--- a/metricbeat/autodiscover/builder/hints/config.go
+++ b/metricbeat/autodiscover/builder/hints/config.go
@@ -1,11 +1,15 @@
 package hints
 
+import "github.com/elastic/beats/metricbeat/mb"
+
 type config struct {
-	Key string `config:"key"`
+	Key      string `config:"key"`
+	Registry *mb.Register
 }
 
 func defaultConfig() config {
 	return config{
-		Key: "metrics",
+		Key:      "metrics",
+		Registry: mb.Registry,
 	}
 }
--- a/metricbeat/autodiscover/builder/hints/metrics.go
+++ b/metricbeat/autodiscover/builder/hints/metrics.go
@@ -33,7 +33,8 @@ const (
 )
 
 type metricHints struct {
-	Key string
+	Key      string
+	Registry *mb.Register
 }
 
 // NewMetricHints builds a new metrics builder based on hints
@@ -46,7 +47,7 @@ func NewMetricHints(cfg *common.Config) (autodiscover.Builder, error) {
 		return nil, fmt.Errorf("unable to unpack hints config due to error: %v", err)
 	}
 
-	return &metricHints{config.Key}, nil
+	return &metricHints{config.Key, config.Registry}, nil
 }
 
 // Create configs based on hints passed from providers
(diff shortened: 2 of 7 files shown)


exekias:
I've opened https://github.com/elastic/beats/pull/6636 to still support the now unhardcoded Prometheus use case

metricbeat/autodiscover/builder/metrics/metrics.go
```diff
@@ -113,17 +114,18 @@ func (m *metricAnnotations) getModule(hints common.MapStr) string {
...
-		// Prometheus stats can be explicitly configured if need be.
-		if module == "prometheus" {
-			msets = []string{"collector"}
-		} else {
-			msets = mb.Registry.MetricSets(module)
+		// If no metricset list is given, take module defaults
+		// fallback to all metricsets if module has no defaults
+		msets, err = m.Registry.DefaultMetricSets(module)
+		if err != nil || len(msets) == 0 {
+			msets = m.Registry.MetricSets(module)
```
ruflin: I worry that with that we enable experimental metricsets which is probably not our intention.

metricbeat/autodiscover/builder/metrics/metrics_test.go
```diff
@@ -40,48 +41,66 @@ func TestGenerateHints(t *testing.T) {
...
-				"namespace":  "test",
-				"metricsets": []interface{}{"collector"},
+				"module":     "mockmodule",
+				"metricsets": []interface{}{"one"},
+				"timeout":    "3s",
+				"period":     "1m",
+				"enabled":    true,
+			},
+		},
+		// Only module, it should return defaults
```
ruflin: Instead of a comment here you could add this as `message` or `description` to the test and output it when the test fails. This makes hunting down a failing test much easier. We have that already in quite a few other places.

metricbeat/autodiscover/builder/metrics/metrics.go
jsoriano: Could we exclude experimental metricsets from defaults? Is there an alternative way to explicitly enable metricsets from hints?

exekias: Do we agree to set default metricsets for all modules? If that's the case this is not a issue

ruflin: yes, we do. but if we merge this we should make sure then that we follow up with the changes until the next release.

metricbeat/autodiscover/builder/metrics/metrics_test.go
exekias: Well, having that this feature is experimental... we could probably live with experimental metricsets being spawned here :innocent:

metricbeat/autodiscover/builder/metrics/metrics.go
exekias: Well, having that this feature is experimental... we could probably live with experimental metricsets being spawned here innocent

PR #6636, opened 2018-03-22
Make `collector` default prometheus metricset

exekias:
This change makes prometheus collector the default metricset, so there is no need to declare it when using it by hand or in autodiscover:

```
metricbeat.modules:
  - module: prometheus
    period: 10s
    hosts: ["localhost:9090"]
    metrics_path: /metrics
```

### DIFF
--- a/metricbeat/module/prometheus/collector/collector.go
+++ b/metricbeat/module/prometheus/collector/collector.go
@@ -24,9 +24,10 @@ var (
 )
 
 func init() {
-	if err := mb.Registry.AddMetricSet("prometheus", "collector", New, hostParser); err != nil {
-		panic(err)
-	}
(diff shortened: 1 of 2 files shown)


metricbeat/module/prometheus/collector/collector.go
```diff
@@ -24,9 +24,10 @@ var (
 )
 
 func init() {
-	if err := mb.Registry.AddMetricSet("prometheus", "collector", New, hostParser); err != nil {
-		panic(err)
-	}
+	mb.Registry.MustAddMetricSet("prometheus", "collector", New,
```
ruflin: We should have a mention of this in the docs.

@andrewkroh Thanks for this feature. We should probably make use of it in some other modules.

ruflin:
@exekias I think the failing tests could be related (docker labels).

Issue #6342, opened 2018-02-09
Automatically lowercase the index name when custom fields is used to build it


daohodac:
As uppercase is invalid in elasticsearch indices, Beats should either provide a lowercase processor or always ensure index names are lowercased when using custom fields

Use case (filebeat):

A document:
```
{"HDG": 125.0, "fwot": "FOO", "date": "2018-01-31 23:00:03"}
```

A beat configuration:

```
.......
- type: log
  paths:
    - /var/log/myfeed/*.log
  json.keys_under_root: true
......
output.elasticsearch:
  hosts: ["localhost:9200"]
  protocol: "http"
  username: "elastic"
  password: "changeme"
  index: "custom-%{[fwot]}-%{+yyyy.MM}-v1"
.....
```

ph:
I would prefer `explicit lowercase` of the field vs magic. @daohodac are you getting bulk insertion error when uppercase are present?

daohodac:
@ph well, not an error, but an error log,. Nothing inserted, just fails. 
And beat does not have a dead letter, so it only appears on the logs

kvch:
Hits to get started:
A new processor needs to be added to lowercase field values. Example processor: https://github.com/elastic/beats/blob/master/libbeat/processors/actions/truncate_fields.go

urso:
I wonder if the processor is the right place. Maybe we should not tamper with event contents itself, but ensure that the strings we extract when creating the index name are lower-case.

Joginder22:
+1 facing same issue with ES output when index name is in uppercase.  I'ts better to have lowercase keyword in the output to make every index name in lowercase letters only.

PR #6640, opened 2018-03-22
Fix issue 6342: Automatically lowercase the index name

Hexilee:
Fix issue [6342](https://github.com/elastic/beats/issues/6342)

### DIFF
--- a/libbeat/outputs/elasticsearch/utils.go
+++ b/libbeat/outputs/elasticsearch/utils.go
@@ -1,0 +1,18 @@
+package elasticsearch
+
+import (
+	"unicode"
+	"github.com/elastic/beats/libbeat/logp"
+	"strings"
+)
+
+func TryLowercaseIndex(index string) string {
+	for _, u := range []rune(index) {
+		if unicode.IsUpper(u) {
+			newIndex := strings.ToLower(index)
+			logp.Warn("Index name %s is invalid, replace it with %s", index, newIndex)
+			return newIndex
+		}
+	}
+	return index
+}
--- a/libbeat/outputs/elasticsearch/client.go
+++ b/libbeat/outputs/elasticsearch/client.go
@@ -390,7 +390,7 @@ func createEventBulkMeta(
 	}
 
 	meta := bulkEventMeta{
-		Index:    index,
+		Index:    TryLowercaseIndex(index),
 		DocType:  eventType,
 		Pipeline: pipeline,
 		ID:       id,
--- a/libbeat/outputs/elasticsearch/utils_test.go
+++ b/libbeat/outputs/elasticsearch/utils_test.go
@@ -1,0 +1,18 @@
+package elasticsearch
+
+import (
+	"testing"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestTryLowercaseIndex(t *testing.T) {
+	testCases := map[string]string{
+		"ElasticSearch": "elasticsearch",
+		"ELASTIC":       "elastic",
+		"beats":         "beats",
+	}
+
+	for former, latter := range testCases {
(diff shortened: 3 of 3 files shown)


libbeat/outputs/elasticsearch/utils.go
```diff
@@ -0,0 +1,18 @@
+package elasticsearch
+
+import (
+	"unicode"
+	"github.com/elastic/beats/libbeat/logp"
+	"strings"
+)
+
+func TryLowercaseIndex(index string) string {
```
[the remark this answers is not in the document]
ph: We probably don't need to export it :)

libbeat/outputs/elasticsearch/utils.go
```diff
@@ -0,0 +1,18 @@
...
+	for _, u := range []rune(index) {
+		if unicode.IsUpper(u) {
+			newIndex := strings.ToLower(index)
+			logp.Warn("Index name %s is invalid, replace it with %s", index, newIndex)
+			return newIndex
+		}
+	}
+	return index
+}
```
ph: There is a few things that bug me in this:
- This could get noisy since it could be potentially be logged on every events.
- Worse case when the index is OK we iterate on all the chars, so an happy path will be hit by this.
- Due to the dynamic nature of index, we use string interpolation here, we have to do this check on every events.

I wonder if instead, we could **provide a better errors message** when this situation happen and either tell them to use a lowercase index or we could provide a lowercase processor.

Hexilee: Could we provide a better errors message and a additional flag like `--lowercase-index` to enable the processor?

ph: Personnaly I would not add a custom flag, they are easy to add, hard to remove and it make the CLI more complicated. 

But the error message could mention how to fix the situation.

"Use the lowercase processor to normalize your index"

ph: @ruflin would you mind give your opinion on my [comment](https://github.com/elastic/beats/pull/6640#discussion_r177299610)?

ruflin: First of all I don't like magic. Sending a field with upper case and then converting it to lower case automatically for the index naming feels like magic to me. I see multiple options moving forward:

* A config option in the elasticsearch output to normalies / lowercase the index which is off by default.
* A lower case processor. The downside of the lower case processor is that if it's applied on a field, I expect it to be also changed in the final event it is sent.

I'm leaning towards the output config for the reason that this is an Elasticsearch specific issue so it should be an Elasticsearch config option. Having a config option off by default will also make it clear that when turning it on, it will have some processing overhead.

+1 on providing better error messages.

ph: > A lower case processor. The downside of the lower case processor is that if it's applied on a field, I expect it to be also changed in the final event it is sent.

Correct.

Glad we got a conclusion.

@Hexilee 

I would be +1 to add a specific options to the elasticsearch output to lowercase the index name and make it off by default.

If you move forward with the changes can you sign the [CLA](https://www.elastic.co/contributor-agreement)?

Hexilee: @ph Yeah, I'd like to, and I have signed the CLA.
However, as you say, current error message is generated by ElasticSearch, so should we improve ElasticSearch only? Whatever providing better error messages or lowercase processor.

ph:
sorry for the long delay seems there still a few comment open on this PR? Are you still interested in moving it forward? I am going to mark this PR as stalled and we will autoclose it.

urso:
Closing this due to PR being stalled for quite some time. An alternative implementation is given in #16081 .

Issue #6620, opened 2018-03-21
[Metricbeat] Monitoring metrics don't work when containerised


Constantin07:
Hello,

I'm still getting these errors in metricbeat logs with version 6.2.3 when deployed as docker container:

```
2018-03-21T10:10:07.539Z	ERROR	instance/metrics.go:69	Error while getting memory usage: error retrieving process stats
2018-03-21T10:10:07.539Z	ERROR	instance/metrics.go:113	Error retrieving CPU percentages: error retrieving process stats
```

but there is no clue what could be the problem. I'm running an official metricbeat docker image and trying to pull stats from host.

The `metricbeat` container is run using this command:
```
docker run -d --restart=always --name metricbeat \
--net=host \
-u root \
-v /proc:/hostfs/proc:ro \
-v /sys/fs/cgroup:/hostfs/sys/fs/cgroup:ro \
-v /:/hostfs:ro \
-v /var/run/docker.sock:/var/run/docker.sock \
metricbeat:6.2.3 -system.hostfs=/hostfs
```

cat system.yml
```yml
- module: system
  period: 10s
  metricsets:
    - cpu
    - load
    - memory
    - network
    - process
    - process_summary
    #- core
    - diskio
    #- socket
  processes: ['.*']
  process.include_top_n:
    by_cpu: 5      # include top 5 processes by CPU
    by_memory: 5   # include top 5 processes by memory

- module: system
  period: 1m
  metricsets:
    - filesystem
    - fsstat
  processors:
  - drop_event.when.regexp:
      system.filesystem.mount_point: '^/(sys|cgroup|proc|dev|etc|host|lib)($|/)'
```

- Metricbeat version: 6.2.3
- Operating System: CoreOS Linux 1465.8.0
- Docker Client 1.10.3 (API 1.22), Docker Server 1.12.6 (API 1.24)

Any idea what could be the cause ?

andrewkroh:
My initial guess is that it cannot find its own metrics in `/hostfs/proc` because it is pid namespaced and therefore is looking for something like 2 but when on the host the PID is very different.

This problem extends from the fact that `/hostfs` is treated as a global variable. This causes all metrics collecting code to read from `/hostfs/proc`. But the self monitoring metrics should always come from `/proc`.

@kvch Can you try to reproduce this? I think we need a test case for this, and we can discuss some possible solutions.

andrewkroh:
[...]

PR #6641, opened 2018-03-22
Fix self metrix when containerised #6620

ewgRa:
This is a fix for #6620.

Probably it is a little bit dirty solution, but on other hand there is big refactoring of gosigar and there is a big question about later upstream merging.

### DIFF
--- a/libbeat/cmd/instance/metrics.go
+++ b/libbeat/cmd/instance/metrics.go
@@ -4,7 +4,6 @@ package instance
 
 import (
 	"fmt"
-	"os"
 	"runtime"
 
 	"github.com/elastic/beats/libbeat/logp"
@@ -65,8 +64,12 @@ func reportMemStats(m monitoring.Mode, V monitoring.Visitor) {
 }
 
 func getRSSSize() (uint64, error) {
-	beatPID := os.Getpid()
-	state, err := beatProcessStats.GetOne(beatPID)
+	pid, err := process.GetSelfPid()
+	if err != nil {
+		return 0, fmt.Errorf("error getting PID for self process: %v", err)
+	}
+
+	state, err := beatProcessStats.GetOne(pid)
 	if err != nil {
 		return 0, fmt.Errorf("error retrieving process stats: %v", err)
 	}
@@ -121,8 +124,12 @@ func reportBeatCPU(_ monitoring.Mode, V monitoring.Visitor) {
 }
 
 func getCPUUsage() (float64, *process.Ticks, error) {
-	beatPID := os.Getpid()
-	state, err := beatProcessStats.GetOne(beatPID)
+	pid, err := process.GetSelfPid()
+	if err != nil {
+		return 0.0, nil, fmt.Errorf("error getting PID for self process: %v", err)
+	}
+
+	state, err := beatProcessStats.GetOne(pid)
 	if err != nil {
 		return 0.0, nil, fmt.Errorf("error retrieving process stats: %v", err)
 	}
--- a/libbeat/metric/system/process/process_linux.go
+++ b/libbeat/metric/system/process/process_linux.go
@@ -1,0 +1,20 @@
+package process
+
+import (
+	"os"
+	"path"
+	"strconv"
+
+	"github.com/elastic/gosigar"
+)
+
+// GetSelfPid returns the PID for this process
+func GetSelfPid() (int, error) {
+	pid, err := os.Readlink(path.Join(gosigar.Procd, "self"))
+
+	if err != nil {
+		return 0, err
+	}
+
+	return strconv.Atoi(pid)
+}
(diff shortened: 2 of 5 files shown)


libbeat/metric/system/process/process_linux.go
```diff
@@ -0,0 +1,22 @@
...
+var InitSigarProcd string
+var curSigarProcd string
+
+func init() {
+	InitSigarProcd = sigar.Procd
+}
+
+func (procStats *LocalStats) preGetOne() {
+	curSigarProcd = sigar.Procd
+	sigar.Procd = InitSigarProcd
```
jsoriano: Solving it this way would probably require a locking mechanism in `gosigar`, as this value is global and there can be other goroutines using it at the same time.

ewgRa: If this PR will be accepted in general, I think it can be done without big changes of gosigar.

libbeat/metric/system/process/process_linux.go
```diff
@@ -0,0 +1,28 @@
...
+	"strconv"
+	"strings"
+	"syscall"
+
+	"github.com/elastic/gosigar"
+)
+
+// GetSelfPid return self pid for current running process
+func GetSelfPid() (int, error) {
+	contents, err := ioutil.ReadFile(gosigar.Procd + "/self/stat")
```
jsoriano: I was just thinking that it could be enough with `os.Readlink(gosigar.Procd + "/self")`, then no more parsing would be needed. Could you try? Thanks!

jsoriano: Btw, to join paths it'd be better to use `path.Join`.

libbeat/metric/system/process/process_linux.go
```diff
@@ -0,0 +1,28 @@
+package process
```
jsoriano: `// +build linux`

libbeat/metric/system/process/process_other.go
```diff
@@ -0,0 +1,10 @@
+// +build darwin freebsd windows
```
jsoriano: `// +build !linux`

libbeat/metric/system/process/process_linux.go
jsoriano: Oh, this won't be needed as the file has the `_linux` suffix.

ewgRa:
@jsoriano thanks for review. Agree with everything and already made changes. ReadLink works. Check please.

Only requested "// +build linux" for _linux file make no sense, as you already mention.

libbeat/cmd/instance/metrics.go
```diff
@@ -65,8 +64,12 @@ func reportMemStats(m monitoring.Mode, V monitoring.Visitor) {
 }
 
 func getRSSSize() (uint64, error) {
-	beatPID := os.Getpid()
-	state, err := beatProcessStats.GetOne(beatPID)
+	pid, err := process.GetSelfPid()
+	if err != nil {
+		return 0, fmt.Errorf("error retrieving process stats: %v", err)
```
jsoriano: Rewrite to more specific message?

libbeat/cmd/instance/metrics.go
```diff
@@ -121,8 +124,12 @@ func reportBeatCPU(_ monitoring.Mode, V monitoring.Visitor) {
 }
 
 func getCPUUsage() (float64, *process.Ticks, error) {
-	beatPID := os.Getpid()
-	state, err := beatProcessStats.GetOne(beatPID)
+	pid, err := process.GetSelfPid()
+	if err != nil {
+		return 0.0, nil, fmt.Errorf("error retrieving process stats: %v", err)
```
jsoriano: Same here.

libbeat/metric/system/process/process_linux.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	"os"
+	"strconv"
+
```
andrewkroh: This newline could be dropped to group all stdlib imports together.

libbeat/metric/system/process/process_linux.go
```diff
@@ -0,0 +1,21 @@
...
+	"path"
+
+	"github.com/elastic/gosigar"
+)
+
+// GetSelfPid return self pid for current running process
```
andrewkroh: How about "GetSelfPid returns the PID for this process."

jsoriano:
jenkins, test this please

ewgRa:
@jsoriano I squash it and change error to more specific, but a little bit surprised, since github allow to do that. Or just to simplify commit message?

@andrewkroh done group of stdlib imports and doc for function (but I did it without dot at the end, since docs everywhere without dots at the end).


Thanks for review, so small diff at the end, but so much effort.

jsoriano:
@ewgRa thanks for taking care of this!

Yes, I asked you to squash it to have a simpler message, and also a message that you chose as author of the changes :)

jsoriano:
Failing test doesn't seem related, merging this.

Issue #6285, opened 2018-02-05
Filter conditions when field is present or not present.


gwsales:
Need added filter conditions to filter on events that do or do not have specific fields present in the events. 

Logstash Example to be ported to winlogbeats: 

if ![field] {
  drop {}
}
or
if [field]{
  drop {}
}

Current workaround is to do a regex with .* or a not regex .*.

exekias:
I'm :+1: to this feature, having a condition to check if a field (or list of fields) exist would be very useful

videetssinghai:
Hey If anyone isn't working on this, can I take this up?
Please provide some guidance to start

exekias:
Awesome @videetssinghai :tada:, as far as I know, no one is working on this, so feel free to give it a try! Here you have some useful pointers:

1. Docs on existing conditions: https://www.elastic.co/guide/en/beats/filebeat/current/defining-processors.html#conditions
2. Implementation of current conditions: https://github.com/elastic/beats/blob/master/libbeat/processors/condition.go

andrewvc:
Closing, was fixed in #6653

PR #6653, opened 2018-03-24
Add Condition to check if a list of fields exists

sriranganathan:
- Create condition to check whether the give list of fields exist in the event
- Add Tests for the created condition
- Add Documentation for the created condition
- closes - https://github.com/elastic/beats/issues/6285


libbeat/processors/condition.go
```diff
@@ -399,6 +405,17 @@ func (c *Condition) checkRange(event ValuesMap) bool {
 	return true
 }
 
+func (c *Condition) checkHasFields(event ValuesMap) bool {
+	for _, field := range c.hasfields.fields {
+		_, err := event.GetValue(field)
+		if err != nil {
+			logp.Info("Field %v not present in the event", field)
```
ewgRa: I think it must be close to exists code, like in checkEquals, in case if there is no value, just return false.
There is no info logging at all, is it make sense to introduce it only to this case?

libbeat/processors/condition.go
```diff
@@ -32,6 +32,9 @@ type Condition struct {
 		name    string
 		filters map[string]match.Matcher
 	}
+	hasfields struct {
```
ewgRa: 1. Maybe better just have hasfields []string?

2. And maybe rename it to requiredFields? In this case would be more natural "requiredFields []string"

ewgRa:
@sriranganathan I leave my 5cents, just consider it as a suggestion, since I not a part of Elastica team.

libbeat/processors/condition.go
sriranganathan: `requiredFields` is indeed an elegant name, but I feel it doesn't go that well with `when` and other conditions like `not`. I think we should keep this discussion open and try to find a better name for the condition.

libbeat/processors/condition.go
sriranganathan: Thanks for the comments @ewgRa . I have made the changes.

libbeat/processors/condition_test.go
```diff
@@ -145,6 +145,18 @@ func TestCondition(t *testing.T) {
 			},
 			result: true,
 		},
+		{
+			config: ConditionConfig{
+				HasFields: []string{"proc.cmdline"},
```
exekias: Could you please add some tests with several fields here? To check all of them should match

libbeat/docs/processors-using.asciidoc
```diff
@@ -157,6 +158,23 @@ range:
 ------
 
 
+[float]
+[[condition-has_fields]]
+===== `has_fields`
+
+The `has_fields` condition checks if the given fields exist in the
```
andrewkroh: I think this would be more clear if it said "checks if _all_ the given fields exist".

kvch:
jenkins test it

kvch:
jenkins test it

exekias:
This is ready for merge, could you please add a line to the CHANGELOG.asciidoc file?

sriranganathan:
@exekias  added a changelog entry.

ruflin:
jenkins, test it

PR #6657, opened 2018-03-26
Fix differences in Jolokia config examples (#6643)

### LINKED ISSUE #6643: Example config mismatch for Metricbeat Jolokia module
In the sample config at https://github.com/elastic/beats/blob/master/metricbeat/modules.d/jolokia.yml.disabled#L7 we have `jmx.mapping`

While we expect `jmx.mappings` here: https://github.com/elastic/beats/blob/master/metricbeat/module/jolokia/jmx/jmx.go#L49

https://discuss.elastic.co/t/metricbeat-jolokia-configuration/125305

Perhaps we should even add more examples in the sample file

agathver:
Closes #6643 

I also added a few examples of MBeans config

### DIFF
--- a/metricbeat/module/jolokia/_meta/config.yml
+++ b/metricbeat/module/jolokia/_meta/config.yml
@@ -3,7 +3,21 @@
   period: 10s
   hosts: ["localhost"]
   namespace: "metrics"
-  path: "/jolokia/?ignoreErrors=true&canonicalNaming=false"
-  jmx.mapping:
-  jmx.application:
-  jmx.instance:
+  jmx.mappings:
+    - mbean: 'java.lang:type=Runtime'
+      attributes:
+        - attr: Uptime
+          field: uptime
+    - mbean: 'java.lang:type=Memory'
+      attributes:
+        - attr: HeapMemoryUsage
+          field: memory.heap_usage
+        - attr: NonHeapMemoryUsage
+          field: memory.non_heap_usage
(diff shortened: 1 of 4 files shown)


metricbeat/module/jolokia/_meta/config.yml
```diff
@@ -4,6 +4,21 @@
   hosts: ["localhost"]
   namespace: "metrics"
   path: "/jolokia/?ignoreErrors=true&canonicalNaming=false"
```
jsoriano: This is the default path, this can be removed from reference configuration.

metricbeat/module/jolokia/_meta/config.yml
```diff
@@ -4,6 +4,21 @@
...
-  jmx.mapping:
-  jmx.application:
-  jmx.instance:
+  jmx.mappings:
```
jsoriano: Thanks for these examples. Are these mbeans always available in Jolokia/JMX?

agathver: They have been part of Oracle JVM since 1.6 

Most user tools assume them to be present. Though the JMX standard doesn't explicitly state them. 

I'm not sure about IBM JDK.

agathver:
@jsoriano There is nothing much in the docs as of now.

jsoriano:
Ok, LGTM then

Issue #6573, opened 2018-03-15
[Filebeat] filebeat.reference.yml shows password option twice in the same scope


dedemorton:
The filebeat.reference.yml shows password in two places (lines [513](https://github.com/elastic/beats/blob/master/filebeat/filebeat.reference.yml#L513) and 527)

![image](https://user-images.githubusercontent.com/14206422/37496779-c456c35c-2871-11e8-8d36-f2b5d5872368.png)

PR #6662, opened 2018-03-26
Remove double `password` and `username` in the Redis input config opt…

ph:
…ions

- Remove a double password in the yml file.
- Remove mention of username in the configuration, nothing in the input
code uses thats.
- Remove space

Fixes: #6573


kvch:
Could you please add a changelog entry?

ph:
@kvch good catch, in the changelog I only mentioned the removal of the `username` option, everything else will be for the clarification of the documentation.

PR #6663, opened 2018-03-26
Add more details to system socket metricset

### LINKED ISSUE #6646: Metricbeat Socket - Per connection state and keepalive timers
Using metricbeat's `system.socket` module, I would like to see the state of the connection. Example would be whether its `ESTABLISHED` or `TIME_WAIT`. Ideally we could graph the box, and see the state of all connections. High `TIME_WAIT` can indicate an issue.

A second part of this would be to include the `keepalive` timers for individual TCP connections (`netstat -not`). 

- [ ] Add state per socket connection
- [ ] Add keepalive timers per socket connection

exekias:
This PR adds connection state info and keep-alive details (if any) to socket metricset.

Closes #6646

### DIFF
--- a/metricbeat/docs/fields.asciidoc
+++ b/metricbeat/docs/fields.asciidoc
@@ -11971,6 +11971,16 @@ Name of the user running the process.
 
 
 [float]
+=== `system.socket.state`
+
+type: keyword
+
+example: ESTABLISHED
+
+Socket connection state, possible values are: ESTABLISHED, SYN_SENT, SYN_RECV, FIN_WAIT1, FIN_WAIT2, TIME_WAIT, CLOSED, CLOSE_WAIT, LAST_ACK, LISTEN, CLOSING.
+
+
+[float]
 == uptime fields
 
 `uptime` contains the operating system uptime metric.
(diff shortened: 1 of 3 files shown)


andrewkroh:
The statefulness of the metricset is going to be an issue in the usage of this feature. As currently implemented only one event is ever sent per flow (tuple of ip_type:src_ip:src_port:dst_ip:dst_port). So there won't by any updates beyond the first event about the changing TCP state or keep-alive metrics.

https://github.com/elastic/beats/blob/47bd6a043682b7e05c1514154caeeeab35870027/metricbeat/module/system/socket/socket.go#L112-L113

exekias:
oh, that's true, I totally forgot about it, we should come up with some other options to support these cases.

Would it make sense to send all sockets on every fetch? I guess it's like it is to avoid too many documents?

andrewkroh:
There could a new config option to enable the detailed socket information. This option would change the behavior to sending an event anytime the state changes and sending periodically after the keepalive state is reached. We should try to find a solution that doesn't change the behavior too much for existing users.

ruflin:
We could introduce a `filter` config and the default is filter the know sockets. Having the filter off would resend everything every time. An other filer could be send on state change.

@andrewkroh When you wrote "anytime the state changes" were you thinking of more of subscription model to not send events based on the period but on every change directly?

exekias:
closing as this will require deeper changes

PR #6667, opened 2018-03-26
Remove legacy uptime reporting from Winlogbeat

kvch:
Remove the following metrics from Winlogbeat in favour of the new one provided by libbeat:
* `uptime.start_time`
* `uptime.uptime`
* `uptime.uptime_ms`
* `uptime.server_time`


ruflin:
I was asking myself if we should add a changelog for this, but as it was only internal +1 on not having one.

PR #6669, opened 2018-03-27
Set namespace as default metricset in Aerospike module

ruflin:
* Remove metricset config from short config
* Document default

### DIFF
--- a/metricbeat/docs/modules/aerospike.asciidoc
+++ b/metricbeat/docs/modules/aerospike.asciidoc
@@ -7,7 +7,7 @@ This file is generated! See scripts/docs_collector.py
 
 beta[]
 
-The Aerospike module uses the http://www.aerospike.com/docs/reference/info[Info command] to collect metrics.
+The Aerospike module uses the http://www.aerospike.com/docs/reference/info[Info command] to collect metrics. The default metricset is `namespace`.
 
 [float]
 === Compatibility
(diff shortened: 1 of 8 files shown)


metricbeat/docs/modules/aerospike.asciidoc
```diff
@@ -25,7 +25,6 @@ in <<configuration-metricbeat>>. Here is an example configuration:
 ----
 metricbeat.modules:
 - module: aerospike
-  metricsets: ["namespace"]
```
ruflin: I removed the metricset config from the short config to make it even shorter, but not sure yet if this is a good idea.

ruflin: Working on further modules I realised we should make a decision for each module. If there is only 1 metricset for example, we don't need it. Updated PR accordingly.

metricbeat/docs/modules/aerospike.asciidoc
```diff
@@ -25,8 +25,6 @@ in <<configuration-metricbeat>>. Here is an example configuration:
...
-  enabled: false
   period: 10s
```
andrewkroh: Personally I'd like to see `period: 10s` removed too. And only show this in the reference config.

ruflin: SGTM. @jsoriano WDYT, should we clean up the period too (in case we don't set a different default)? Will make things even shorter.

ruflin:
New version pushed with `period` removed.

Issue #6668, opened 2018-03-26
Set default metricsets for modules


ruflin:
Recently the feature to set default metricsets for a module were introduced. If a module is enabled an no metricsets are set, this list of metricsets is used. Now for each module the metricsets must be defined. In general this is all metricsets which are GA but there might be some exceptions like prometheus where the default one is only collector.

This issue is to track the progress on setting the defaults metricsets for all modules.

 * [x] aerospike https://github.com/elastic/beats/pull/6669
 * [x] apache https://github.com/elastic/beats/pull/6673
 * [x] ceph #6682
 * [x] couchbase #6683
 * [x] docker https://github.com/elastic/beats/pull/6718
 * [x] dropwizard https://github.com/elastic/beats/pull/6719
 * [x] elasticsearch https://github.com/elastic/beats/pull/6755
 * [x] etcd https://github.com/elastic/beats/pull/6756
 * [x] ~~golang~~
 * [x] graphite https://github.com/elastic/beats/pull/6757
 * [x] haproxy https://github.com/elastic/beats/pull/6758
 * [x] ~~http~~
 * [x] ~~jolokia~~
 * [x] kafka https://github.com/elastic/beats/pull/6759
 * [x] kibana https://github.com/elastic/beats/pull/6762
 * [x] kubernetes https://github.com/elastic/beats/pull/6764
 * [x] logstash https://github.com/elastic/beats/pull/6763
 * [x] memcached https://github.com/elastic/beats/pull/6765
 * [x] mongodb https://github.com/elastic/beats/pull/6766
 * [x] munin https://github.com/elastic/beats/pull/6767
 * [x] mysql https://github.com/elastic/beats/pull/6769
 * [x] nginx #6770 
 * [x] php_fpm #6768 
 * [x] postgresql #6761
 * [x] prometheus #6747 (and #6636)
 * [x] rabbitmq #6743 
 * [x] redis #6742
 * [x] system #6689 
 * [x] uwsgi #6688
 * [x] vsphere #6676
 * [x] windows #6675
 * [x] zookeeper #6674

ruflin:
@jsoriano I started working from the top of the list.

ruflin:
@jsoriano When changing each module I also directly check https://github.com/elastic/beats/issues/6672 So I add `enabled: true` to the reference config file (if there is one) and remove `enabled: *` from the short config. 

Other changes I do:

*  Add a note to the docs
* Add an entry to the changelog

[...]

PR #6671, opened 2018-03-27
Set reference config as config for docs

ruflin:
Having the reference file as example config in the docs has the advantage that all config options are shown.

### DIFF
--- a/metricbeat/docs/modules/system.asciidoc
+++ b/metricbeat/docs/modules/system.asciidoc
@@ -29,35 +29,74 @@ in <<configuration-metricbeat>>. Here is an example configuration:
 ----
 metricbeat.modules:
 - module: system
-  process.include_top_n:
-    by_cpu: 5      # include top 5 processes by CPU
-    by_memory: 5   # include top 5 processes by memory
-
-#- module: system
-#  metricsets:
-#    - core
-#    - diskio
-#    - socket
-
-- module: system
-  period: 1m
   metricsets:
-    - filesystem
-    - fsstat
-  processors:
-  - drop_event.when.regexp:
-      system.filesystem.mount_point: '^/(sys|cgroup|proc|dev|etc|host|lib)($|/)'
+    - cpu             # CPU usage
+    - filesystem      # File system usage for each mountpoint
+    - fsstat          # File system summary metrics
+    - load            # CPU load averages
+    - memory          # Memory usage
+    - network         # Network IO
+    - process         # Per process metrics
+    - process_summary # Process summary
+    - uptime          # System Uptime
+    #- core           # Per CPU core usage
+    #- diskio         # Disk IO
+    #- raid           # Raid
+    #- socket         # Sockets and connection info (linux only)
+  enabled: true
+  period: 10s
+  processes: ['.*']
 
-- module: system
-  period: 15m
-  metricsets:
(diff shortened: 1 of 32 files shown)


metricbeat/metricbeat.reference.yml
```diff
@@ -377,16 +377,15 @@ metricbeat.modules:
...
+  metricsets: ["stubstatus"]
+  period: 10s
 
   # Nginx hosts
-  #hosts: ["http://127.0.0.1"]
+  hosts: ["http://127.0.0.1"]
 
   # Path to server status. Default server-status
-  #server_status_path: "server-status"
+  server_status_path: "server-status"
```
ph: I've looked at the configuration, I see something we define `enabled: false` and sometime we don't, I think we should be explicit and it should be defined all the time. 

I am saying this because your changes is removing that option.

ruflin: Agree, see https://github.com/elastic/beats/issues/6672 We will be cleaning this up going through https://github.com/elastic/beats/issues/6668

metricbeat/docs/modules/system.asciidoc
```diff
@@ -26,35 +26,70 @@ in <<configuration-metricbeat>>. Here is an example configuration:
...
+  # collect data from filesystems matching any of the specified types, and
+  # fsstats will not include data from these filesystems in its summary stats.
+  #filesystem.ignore_types: []
+
+  # These options allow you to filter out all processes that are not
+  # in the top N by CPU or memory, in order to reduce the number of documents created.
+  # If both the `by_cpu` and `by_memory` options are used, the union of the two sets
+  # is included.
+  #process.include_top_n:
+    #
```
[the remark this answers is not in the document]
ruflin: cleaned it up in our `beat.reference.yml`.

ruflin:
Cleaned up version pushed.

ruflin:
This should be rebased and run `make update` as soon as https://github.com/elastic/beats/issues/6668 is closed. Set to in progress for now.

metricbeat/docs/modules/system.asciidoc
```diff
@@ -26,35 +26,70 @@ in <<configuration-metricbeat>>. Here is an example configuration:
...
-      system.filesystem.mount_point: '^/(sys|cgroup|proc|dev|etc|host|lib)($|/)'
+  # Configure the metric types that are included by these metricsets.
+  cpu.metrics:  ["percentages"]  # The other available options are normalized_percentages and ticks.
+  core.metrics: ["percentages"]  # The other available option is ticks.
 
-- module: system
-  period: 15m
-  metricsets:
-    - uptime
+  # A list of filesystem types to ignore. The filesystem metricset will not
```
dedemorton: We decided awhile back ago to strip out most comments from config examples because it meant updating duplicate content and making the docs longer. I think short comments are probably OK, but the comments here seem like that might get out-of-date over time. I worry that we'll forget to update the comments here.

ruflin: This is from our reference docs and are the ones that should be most up-to-date. This part is auto generated.

ph:
Will need a rebase :(

ruflin:
Ready to go.

PR #6673, opened 2018-03-27
Set `status` as default metricset in Apache module

ruflin:
The metricset config was removed from the short config as there is only 1 metricset anyway. No need for the user to configure this.

### DIFF
--- a/metricbeat/docs/modules/apache.asciidoc
+++ b/metricbeat/docs/modules/apache.asciidoc
@@ -6,7 +6,7 @@ This file is generated! See scripts/docs_collector.py
 == Apache module
 
 This module periodically fetches metrics from https://httpd.apache.org/[Apache
-HTTPD] servers.
+HTTPD] servers. The default metricset is `status`.
 
 [float]
 === Compatibility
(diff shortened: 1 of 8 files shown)


metricbeat/docs/modules/apache.asciidoc
```diff
@@ -6,7 +6,7 @@ This file is generated! See scripts/docs_collector.py
 == Apache module
 
 This module periodically fetches metrics from https://httpd.apache.org/[Apache
-HTTPD] servers.
+HTTPD] servers. The default metricset is `namespace`.
```
jsoriano: Defalt metricset is status :)

ruflin: copy/paste :-( Fixed.

metricbeat/module/apache/_meta/config.yml
```diff
@@ -1,5 +1,4 @@
 - module: apache
-  metricsets: ["status"]
   period: 10s
 
   # Apache hosts
```
andrewkroh: I think we could drop the "Apache hosts" comment. It really doesn't any new info IMO.

ruflin: host comment removed and also period removed.

PR #6674, opened 2018-03-27
Set default metricset in zookeeper module

jsoriano:
See #6668


jsoriano:
Docs and reference config updated

PR #6675, opened 2018-03-27
Set default metricset in windows module to `service`

jsoriano:
See #6668

### DIFF
--- a/metricbeat/docs/modules/windows.asciidoc
+++ b/metricbeat/docs/modules/windows.asciidoc
@@ -7,7 +7,8 @@ This file is generated! See scripts/docs_collector.py
 
 beta[]
 
-This is the Windows module.
+This is the Windows module. It collects metrics from Windows systems,
+by default metricset `service` is enabled.
 
 
 [float]
@@ -20,12 +21,6 @@ in <<configuration-metricbeat>>. Here is an example configuration:
 ----
 metricbeat.modules:
 - module: windows
-  metricsets: ["perfmon"]
-  period: 10s
-  perfmon.counters:
-
-- module: windows
-  metricsets: ["service"]
   period: 60s
 ----
 
--- a/metricbeat/metricbeat.reference.yml
+++ b/metricbeat/metricbeat.reference.yml
@@ -493,11 +493,18 @@ metricbeat.modules:
 #------------------------------- Windows Module ------------------------------
 - module: windows
   metricsets: ["perfmon"]
+  enabled: true
   period: 10s
+  perfmon.ignore_non_existent_counters: true
   perfmon.counters:
+  #  - instance_label: processor.name
+  #    instance_name: total
+  #    measurement_label: processor.time.total.pct
+  #    query: '\Processor Information(_Total)\% Processor Time'
 
 - module: windows
   metricsets: ["service"]
+  enabled: true
   period: 60s
 
 #------------------------------ ZooKeeper Module -----------------------------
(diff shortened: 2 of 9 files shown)


metricbeat/metricbeat.reference.yml
```diff
@@ -491,14 +491,9 @@ metricbeat.modules:
 
 #------------------------------- Windows Module ------------------------------
 - module: windows
-  metricsets: ["perfmon"]
   period: 10s
   perfmon.counters:
 
-- module: windows
```
ruflin: Reference file should definitiveyl contain both options.

jsoriano:
Added reference config and updated doc.

Regarding why chosing perfmon as default option I don't have a strong opinion on that, I only though that maybe by default the user may not want to monitor all services.
What do you think? Should we enable both metricsets by default?

metricbeat/docs/modules/windows.asciidoc
```diff
@@ -20,13 +21,8 @@ in <<configuration-metricbeat>>. Here is an example configuration:
 ----
 metricbeat.modules:
 - module: windows
-  metricsets: ["perfmon"]
   period: 10s
   perfmon.counters:
```
andrewkroh: `perfmon` requires configuration so maybe it shouldn't be a "default". You must configure `perfmon.counters` in order for it to be useful so I think this example needs improvement.

The `service` metricset does not require any configuration so it might be a good default. But TBH the default period of 10s is a excessive. I'd probably increase it 60s at a minimum or drop events for healthy services.

ruflin: +1 on the proposal. As we have now also a `config.reference.yml` we can set the period to 60s (non default) in the config.yml. Also we can add a processor to drop the healthy events with a processor in the `config.yml`.

jsoriano: Agree with the proposals, thanks @andrewkroh 

Would it be ok to set a different default for the period in the initialization of this module?

ruflin: It would be possible but as the default period is possible, I would not do that and use the `config.yml` for it. This keeps the code and defaults simple, but still allows us to overwrite them.

jsoriano: Oh, period is module-wide :thinking: it shouldn't be set from the metricset.

ruflin: You can define a module multiple times in the config if you want to have different periods for different metricsets.

jsoriano:
Changed default metricset to `service`, reverted changes in periods in reference config, and added commented-out example for `perfmon`.

metricbeat/metricbeat.reference.yml
```diff
@@ -492,11 +492,18 @@ metricbeat.modules:
...
   perfmon.counters:
+  #  - instance_label: processor.name
+  #    instance_name: total
+  #    measurement_label: processor.time.total.pct
+  #    query: '\Processor Information(_Total)\% Processor Time'
 
 - module: windows
   metricsets: ["service"]
+  enabled: true
   period: 60s
```
ruflin: Now it becomes interesting. In the `reference` config we should use defaults and only in `config.yml`overwrite the defaults. At least that is what we did so far.

jsoriano: This is how periods were in the reference config before this PR :)

I think that supporting different default periods can make a lot of sense (this module is a good example), but this gives for a bigger discussion out of the scope of this PR, wdyt?

ruflin: That we had different periods in the reference config was more an accident as the values were taken from the short config. Now that we have a reference file we could use the default one.

There are a lot of interesting discussion popping up with our changes here. We can take these offline. Will merge as is for now.

PR #6676, opened 2018-03-27
Set default metricsets in vsphere module

jsoriano:
See #6668

### DIFF
--- a/metricbeat/metricbeat.reference.yml
+++ b/metricbeat/metricbeat.reference.yml
@@ -477,6 +477,7 @@ metricbeat.modules:
 
 #------------------------------- vSphere Module ------------------------------
 - module: vsphere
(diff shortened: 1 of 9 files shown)


metricbeat/metricbeat.reference.yml
```diff
@@ -477,7 +477,6 @@ metricbeat.modules:
 
 #------------------------------- vSphere Module ------------------------------
 - module: vsphere
-  metricsets: ["datastore", "host", "virtualmachine"]
```
ruflin: We should keep all metricsets in the reference file and also add `enabled: true`.

I assume there is no `config.reference.yml` in vshpere so we probably have to add one.

jsoriano:
`config.reference.yml` added and docs updated

PR #6680, opened 2018-03-27
Reduce memory usage of http parser

### LINKED ISSUE #6679: http parser should not buffer the complete request and response in memory
Users have reported that the performance of packetbeat is poor when dealing with HTTP requests and responses with large bodies.

After a quick analysis of the HTTP parser it has been identified the case where the parser keeps the whole request and response (including bodies) in memory, even when it is not necessary.

adriansr:
This patch updates the http protocol parser in packetbeat to not buffer the full request/response.

Body is only buffered when neccessary, i.e. the `include_body_for` option specifies that the body should be included in the event, or a request body of type form-urlencoded is sent.

Fixes #6679


packetbeat/protos/http/http.go
```diff
@@ -514,6 +512,15 @@ func (http *httpPlugin) newTransaction(requ, resp *message) beat.Event {
 	}
 }
 
+func (http *httpPlugin) getRaw(m *message) []byte {
```
urso: maybe rename this to `buildRawMessage` or `makeRawMessage`. Quite some allocation and copying is happening here.

packetbeat/protos/http/http.go
```diff
@@ -583,82 +589,42 @@ func parseCookieValue(raw string) string {
...
 	authText := []byte("uthorization:") // [aA] case insensitive, also catches Proxy-Authorization:
 
-	authHeaderStartX := m.headerOffset
-	authHeaderEndX := m.bodyOffset
+	authHeaderStartX := bytes.Index(msg, constCRLF)
+	if authHeaderStartX < 0 {
+		return
+	}
+	authHeaderStartX += len(constCRLF)
+	authHeaderEndX := limit
```
urso: why remove the `header/bodyOffset` in favour of bytes.Index? They still seem to be usefull when postprocessing messages.

packetbeat/protos/http/http_parser.go
```diff
@@ -375,28 +379,39 @@ func (parser *parser) parseHeader(m *message, data []byte) (bool, bool, int) {
...
 		// HTTP/1.0 no content length. Add until the end of the connection
 		if isDebug {
-			debugf("http conn close, received %d", len(s.data)-s.parseOffset)
+			debugf("http conn close, received %d", len(s.data))
 		}
-		s.bodyReceived += (len(s.data) - s.parseOffset)
-		m.contentLength += (len(s.data) - s.parseOffset)
-		s.parseOffset = len(s.data)
+		if m.saveBody {
+			m.body = append(m.body, s.data...)
```
urso: Is there a chance `len(body) > http header content length` after this append? E.g. another HTTP message starting right after the current one due to HTTP keep-alive/pipelining.

urso:
It's a many small, but subtle changes. Most notable is the removal of parseOffset in favour of accessing s.data directly. Can you add a short comment to the PR description why it's removed and how this changes parsing (I guess the s.data slice is always advanced by number of consumed bytes).

packetbeat/protos/http/http.go
adriansr: Removal of headerOffset was a mistake. bodyOffset is unnecessary as the headers and body are now stored in the different slices.

packetbeat/protos/http/http_parser.go
adriansr: Not in this particular append, as it is only called in the particular case where the length is not known and connection is `close` or http/1.0 is used, which rules out keep-alive/pipelining.

packetbeat/protos/http/http.go
```diff
@@ -514,6 +512,16 @@ func (http *httpPlugin) newTransaction(requ, resp *message) beat.Event {
 	}
 }
 
+func (http *httpPlugin) makeRawMessage(m *message) string {
+	var result []byte
+	result = append(result, m.rawHeaders...)
+	if m.sendBody {
+		result = append(result, m.body...)
+	}
+	// TODO: (go1.10) Use strings.Builder to avoid allocation/copying
```
urso: We have a followup github issue?

PR #6682, opened 2018-03-28
Set all ceph metrics as default in metricbeat module

ruflin:


metricbeat/module/ceph/_meta/config.reference.yml
```diff
@@ -0,0 +1,5 @@
+- module: ceph
+  metricsets: ["cluster_disk", "cluster_health", "monitor_health", "pool_disk", "osd_tree"]
```
jsoriano: This is another example of module where we may want different default periods for different metricsets.

ruflin: good point, let me look into this. I also realised one of the metricsets is experimental so not all of them should be here (or I should set it to beta).

ruflin:
I adjusted the defaults so the disk metrics are only reported every minute. I skipped osd_df for now as this metricset needs some cleanup before going into beta.

PR #6686, opened 2018-03-28
Add dashboard check to metricbeat module overview

ruflin:
On discuss we frequently get asked if there are dashboards for a module. This adds a column with a checkbox if there are dashboards for a module.

It only checks if there is a Kibana directory.

Below is a screenshot on how it looks at the moment. Instead of yes/no I would like to have something like a checkbox. @dedemorton What would you recommend to be used here?

screen shot 2018-03-28 at 16 32 46


metricbeat/scripts/docs_collector.py
```diff
@@ -177,17 +178,21 @@ def collect(beat_name):
...
 
         release_label = ""
         if m["release"] != "ga":
             release_label = m["release"] + "[]"
 
-        module_list_output += '|{} {}   |{}    \n'.format("<<metricbeat-module-" +
-                                                          key + "," + m["title"] + ">> ", release_label, "")
+        dashboards = "no"
+        if m["dashboards"]:
+            dashboards = "yes"
```
jsoriano: `dashboards = "yes" if m["dashboards"] else "no"`?

ruflin: Works for me. I will wait with the change until we get some feedback from @dedemorton on if we could use something else then yes/no like a checkmark / emoji. 

Added the `in progress` label.

ruflin:
Thanks @dedemorton for the details. I tried it out and it now looks like this. WDYT?

screen shot 2018-04-04 at 22 26 50

dedemorton:
@ruflin IMO, the check marks look better than yes/no... I don't like that users need to scan across a whole column to see info that applies to the module. You *could* change the order of cols...but I'm not sure it's worth your time. But here's a mockup (used a graphics editor) to show you what I mean: 

image

ruflin:
@dedemorton I changed it accordingly. The table makes much more sense like this. Thanks for taking the time to build all these mockups.

ruflin:
@dedemorton @jsoriano Ready to go.

dedemorton:
@ruflin LGTM. I think it's OK as is, but wasn't expecting the row borders between the metricsets to appear only in the metricsets column. However I recall from when we first created the table that we had to compromise a bit in terms of layout because we are building the table programmatically. I think it looks fine and is completely readable as it is now.

ruflin:
@dedemorton No tradeoffs have to be made here. I followed up with https://github.com/elastic/beats/pull/6788

PR #6687, opened 2018-03-28
Add hex decoding to name field in audit path records

andrewkroh:
This updates go-libaudit to v0.1.0 which contains a fix for elastic/go-libaudit#20.


ruflin:
@andrewkroh Probably needs a make update for the NOTICE file.

PR #6688, opened 2018-03-28
Set default metricset for uwsgi module

jsoriano:
See #6668

### DIFF
--- a/metricbeat/module/uwsgi/_meta/config.yml
+++ b/metricbeat/module/uwsgi/_meta/config.yml
@@ -1,4 +1,2 @@
 - module: uwsgi
-  metricsets: ["status"]
-  period: 10s
   hosts: ["tcp://127.0.0.1:9191"]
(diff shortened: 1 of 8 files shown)


metricbeat/module/uwsgi/_meta/config.yml
```diff
@@ -1,4 +1 @@
 - module: uwsgi
-  metricsets: ["status"]
-  period: 10s
-  hosts: ["tcp://127.0.0.1:9191"]
```
ruflin: did you remove the host intentionally? Our `config.yml` gets shorter and short. There could be arguments to also remove the hosts part. @andrewkroh ?

jsoriano: I was thinking that this was the default, I was wrong :) I'll readd the `host` entry.

jsoriano: Well, in the documentation it says this is the default, but I cannot find it in the code, I'm going to take a look to this so doc and code are aligned.

jsoriano: I have readded the `host` entry, and rephrashed doc to mention `tcp://127.0.0.1:9191` as example instead of default.

PR #6689, opened 2018-03-28
Set default metricsets for system module

jsoriano:
See #6668 

Set as default metricset the ones that appear in example configuration enabled and don't require further configuration.

### DIFF
--- a/metricbeat/docs/modules/system.asciidoc
+++ b/metricbeat/docs/modules/system.asciidoc
@@ -8,6 +8,9 @@ This file is generated! See scripts/docs_collector.py
 The System module allows you to monitor your servers. Because the System module
 always applies to the local server, the `hosts` config option is not needed.
 
+The default metricsets are `cpu`, `load`, `memory`, `network`, `process` and
+`process_summary`.
+
 [float]
 === Dashboard
 
@@ -26,21 +29,15 @@ in <<configuration-metricbeat>>. Here is an example configuration:
 ----
 metricbeat.modules:
 - module: system
-  period: 10s
-  metricsets:
-    - cpu
-    - load
-    - memory
-    - network
-    - process
-    - process_summary
-    #- core
-    #- diskio
-    #- socket
-  processes: ['.*']
   process.include_top_n:
     by_cpu: 5      # include top 5 processes by CPU
     by_memory: 5   # include top 5 processes by memory
+
+#- module: system
+#  metricsets:
+#    - core
+#    - diskio
+#    - socket
 
 - module: system
   period: 1m
@@ -55,6 +52,12 @@ metricbeat.modules:
   period: 15m
   metricsets:
     - uptime
+
+#- module: system
+#  period: 5m
+#  metricsets:
+#    - raid
+#  raid.mount_point: '/'
 ----
 
 [float]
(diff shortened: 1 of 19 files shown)


metricbeat/docs/modules/system.asciidoc
```diff
@@ -55,6 +47,16 @@ metricbeat.modules:
   period: 15m
   metricsets:
     - uptime
+    #- core
```
ruflin: We probably can remove this and the following lines from the `config.yml`. Or would you keep them in the advertise the others?

metricbeat/metricbeat.reference.yml
```diff
@@ -106,6 +107,9 @@ metricbeat.modules:
   # to false.
   #process.include_cpu_ticks: false
 
+  # Raid mount point to monitor
+  #raid.mount_point: '/'
```
ruflin: Thanks for completing this.

metricbeat/docs/modules/system.asciidoc
jsoriano: Yes, I though it could be fine to leave them for visibility. I could also leave them in another commented-out module block.

jsoriano: I have finally added another commented-out module block for these non-default metricsets.

jsoriano:
Fixed a typo that broke compilation, and another one in the documentation.

metricbeat/module/system/_meta/config.yml
```diff
@@ -28,3 +23,9 @@
   period: 15m
   metricsets:
     - uptime
+
+# - module: system
```
ruflin: Nit: There is a space after # which we normally don't have except for comments.

metricbeat/module/system/_meta/config.yml
```diff
@@ -1,20 +1,15 @@
...
-    #- core
-    #- diskio
-    #- socket
-  processes: ['.*']
+  # processes: ['.*']
   process.include_top_n:
     by_cpu: 5      # include top 5 processes by CPU
     by_memory: 5   # include top 5 processes by memory
 
+#- module: system
```
ruflin: Not sure if the indentation is correct if the `#` is removed here. It should be 2 spaces before the `#`.

metricbeat/module/system/_meta/config.yml
jsoriano: Ok, fixed.

metricbeat/module/system/_meta/config.yml
jsoriano: I think identation is correct, if the `#` is removed it stays at the same level as other blocks.

ruflin: ah, you are right.

metricbeat/module/system/_meta/config.yml
```diff
@@ -1,20 +1,15 @@
...
-    - load
-    - memory
-    - network
-    - process
-    - process_summary
-    #- core
-    #- diskio
-    #- socket
-  processes: ['.*']
+  # processes: ['.*']
```
ruflin: Two thins here: 

* It seemed we changed the default here from having `processes` filter enabled to disabled
* If it is commented out, the space after the `#` should be removed.

I would keep the previous config if possible.

jsoriano: I commented it out because this value is already the default one, I could remove it as it is already in the reference config.

ruflin: Let's remove it.

jsoriano: Ok, done.

PR #6691, opened 2018-03-28
allow forcing dynamic object type mapping

graphaelli:
Currently dynamic mappings are tolerant of unexpected data types.  For example, this definition:
```
- name: test
    type: object
    object_type: long
    dynamic: true
```

produces this dynamic template:
```
"test": {
  "path_match": "test.*",
  "match_mapping_type": "long",
  "mapping": {
    "type": "long"
  }
}
```

If the first value seen is a string, say indexing this document:
```
"test": {
      "somelong": 16,
      "somestring": "foo"
}
```

the resulting mapping is actually:
```
"test": {
  "dynamic": "true",
  "properties": {
    "test": {
      "properties": {
        "somelong": {
          "type": "long"
        },
        "somestring": {
          "type": "keyword",
          "ignore_above": 1024
        }
      }
    }
  }
}
```

In apm-server, data is already validated before being sent along to elasticsearch, so barring a bug, a string will not be sent where a long is required.  This change allows forcing the data for a dynamic field to be a specific type via field config, eg for:

```
- name: test
    type: object
    object_type: long
    object_type_mapping_type: "*"
    dynamic: true
```

The resulting dynamic template would be:

```
"test": {
  "path_match": "test.*",
  "match_mapping_type": "*",
  "mapping": {
    "type": "long"
  }
}
```

Any value that can't be mapped to `long` will result in an indexing error.  Continuing this example:
```
"error": {
  "root_cause": [
    {
      "type": "mapper_parsing_exception",
       "reason": "failed to parse [test.somestring]"
    }
  ]
}
```

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -25,6 +25,7 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
 - Add logging when monitoring cannot connect to Elasticsearch. {pull}6365[6365]
 - Rename beat.cpu.*.time metrics to beat.cpu.*.time.ms. {pull}6449[6449]
 - Mark `system.syslog.message` and `system.auth.message` as `text` instead of `keyword`. {pull}6589[6589]
+- Allow override of dynamic template `match_mapping_type` for fields with object_type. {pull}6691[6691]
 
 *Auditbeat*
 
--- a/libbeat/common/field.go
+++ b/libbeat/common/field.go
@@ -17,22 +17,23 @@ import (
 type Fields []Field
 
 type Field struct {
-	Name           string      `config:"name"`
-	Type           string      `config:"type"`
-	Description    string      `config:"description"`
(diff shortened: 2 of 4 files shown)


ruflin:
Should we introduce a config `match_mapping_type` instead to be closer to the naming in Elasticsearch. Instead of setting the force config it would allow to set the type that it should actually map, also `*`.

Can you add a CHANGELOG?

For the PR description it would be nice if it would also contain the new functionality not only the way it works now which would make it easy to compare the two without having to go into the code.

graphaelli:
> Should we introduce a config match_mapping_type instead to be closer to the naming in Elasticsearch. Instead of setting the force config it would allow to set the type that it should actually map, also *.

Works for me, is that your preference then?

> Can you add a CHANGELOG?

Will do once this is takes shape.

> For the PR description it would be nice if it would also contain the new functionality not only the way it works now which would make it easy to compare the two without having to go into the code.

Hey, I thought the original description was pretty good!  I'll update for future readers.

PR #6693, opened 2018-03-29
Be more strict in LICENSE checking

tsg:
This adds checks that:

* everything that we vendor has at least a LICENSE file
* all licenses used are from the accepted list

The checks are executed on `make check` and `make notice`.


dev-tools/generate_notice.py
```diff
@@ -51,32 +51,85 @@ def gather_dependencies(vendor_dirs):
...
+    """
+    Get a list of license files from a given directory.
+    """
+    licenses = []
+    for filename in os.listdir(folder):
+        if filename.startswith("LICENSE") and "docs" not in filename:
+            licenses.append(filename)
+        if filename.startswith("APLv2"): # gorhill/cronexpr
+            licenses.append(filename)
+    return licenses
```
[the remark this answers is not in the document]
tsg: I need to call `len(get_licenses(...))` a couple of times. That would mean I'd have to call `len(list(get_licenses(...)))` which I find uglier than the current version.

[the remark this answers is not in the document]
jsoriano: Ok, I see. For these checks you could also use `if len([get_licenses(...)]) > 0:` (uglier too), or `if any(get_licenses(...)):` (a bit trickier).

ruflin:
It seems in this PR some dependencies were updated but the vendor.json file did not change?

It will also need an other `make update`.

tsg:
@ruflin I removed `goautoneg` from `vendor.json` as well and rebased the PR.

dev-tools/generate_notice.py
```diff
@@ -51,35 +51,92 @@ def gather_dependencies(vendor_dirs, overrides=None):
...
+                    print("WARNING: No version information found for: {}".format(lib_path))
+                    lib = {"path": lib_path}
+                else:
+                    lib = lib_search[0]
+                lib["license_file"] = os.path.join(root, filename)
+
+                lib["license_contents"] = read_file(lib["license_file"])
+                lib["license_summary"] = detect_license_summary(lib["license_contents"])
+                if lib["license_summary"] == "UNKNOWN":
+                    print("WARNING: Unknown license for: {}".format(lib_path))
```
ruflin: For the future: We could exit with 1 in such cases to make sure CI fails and we don't add any unknown licenese.

metricbeat/vendor/vendor.json
```diff
@@ -2,12 +2,6 @@
 	"comment": "",
 	"ignore": "test github.com/elastic/beats",
 	"package": [
-		{
```
ruflin: @tsg It is still strange that there are new LICENSE files for `go-sysinfo` for example and it does not show up as a new commit id in the vendor.json? Same for the other 4 where licene files were added.

dev-tools/generate_notice.py
tsg: There's another check for "acceptable" licenses at the end of the script, which raises an exception and therefore makes CI red.

metricbeat/vendor/vendor.json
tsg: Uhm, yes, I manually copied the license files in some cases 😇 

In some cases, govendor doesn't copy those files, because they are at a different level than the go files. That was the case of `go-ole/go-ole`. I'm thinking manually copying it is the best approach for that case.

For `go-sysinfo` and `go-windows` I only copied them that way because the upstream PRs that were adding the licenses upstream weren't merged yet. I'll do those the right way now.

tsg:
jenkins, test it

metricbeat/vendor/vendor.json
tsg: Hmm, I tried to upgrade the two but they both come with a few code changes unrelated to the license. Fine if we do the upgrade in a new PR so that the code changes are easier to review?

ruflin: @tsg yes

For the go-ole one I worry that every time we update we forget the LICENSE but not sure how to fix that.

tsg: We can't forget because the script will now return an error if any dep doesn't have the license.

PR #6700, opened 2018-03-29
Feature: TCP Input

ph:
Allow to receive new events via TCP, this will create a new event per
line and add some useful information about the connected client to the
event. This input is marked as **experimental**.

This input expose the following settings:

- `line_delimiter`: The characters used to split incoming events, by
default it will split on `\n` and will also strip both `\n` and `\r`.
You can also use any characters like `;` and you can also used multiple
characters delimiter like `<END>`, the delimiter tokens will always be
removed from the string.

- `max_message_size`: This is a number of bytes that a client can buffer
in memory before finding a new message, if the limit is reached the
connection is killed and the event is logged. This is to prevent rogue
clients to DoS attack by consuming all the available memory. The default
values is 20971520.

- `timeout`: The server will close any client that reach this inactivity
timeout.

- `Host`: Which IP and port to bind.

This is a new PR based on #6266 

A few things are differents:

- TCP is now extracted from the actual inputs.
- Visibility of struct has been reduced.
- Logging is done with a contextual logging.
- Follow the same semantic as #6439 (Thanks @andrewkroh)

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -52,6 +52,7 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
 - Remove the undefined `username` option from the Redis input and clarify the documentation. {pull}6662[6662]
 - Add validation for Stdin, when Filebeat is configured with Stdin and any other inputs, Filebeat
   will now refuses to start. {pull}6463[6463]
+- Addition of the TCP input {pull}6700[6700]
 
 *Heartbeat*
 
--- a/filebeat/_meta/common.reference.p2.yml
+++ b/filebeat/_meta/common.reference.p2.yml
@@ -240,6 +240,23 @@ filebeat.inputs:
   # Maximum size of the message received over UDP
   #max_message_size: 10240
 
+#------------------------------ TCP prospector --------------------------------
+# Experimental: Config options for the TCP input
+#- type: tcp
+  #enabled: false
+
+  # The host and port to receive the new event
+  #host: "localhost:9000"
+
+  # Character used to split new message
+  #line_delimiter: "\n"
+
+  # Maximum size in bytes of the message received over TCP
+  #max_message_size: 20MiB
+
+  # The number of seconds of inactivity before a remote connection is closed.
+  #timeout: 300s
+
 #========================== Filebeat autodiscover ==============================
 
 # Autodiscover allows you to detect changes in the system and spawn new modules
--- a/filebeat/docs/inputs/input-tcp.asciidoc
+++ b/filebeat/docs/inputs/input-tcp.asciidoc
@@ -1,0 +1,55 @@
+:type: tcp
+
+[id="{beatname_lc}-input-{type}"]
+=== TCP input
+
+++++
+<titleabbrev>TCP</titleabbrev>
+++++
+
+Use the `TCP` input to read events over TCP.
+
+Example configuration:
+
+["source","yaml",subs="attributes"]
+----
+{beatname_lc}.inputs:
+- type: tcp
+  max_message_size: 10240
+  host: "localhost:9000"
+----
+
+
+==== Configuration options
+
+The `tcp` input supports the following configuration options plus the
+<<{beatname_lc}-input-{type}-common-options>> described later.
+
+[float]
+[id="{beatname_lc}-input-{type}-max-message-size"]
+==== `max_message_size`
+
+The maximum size of the message received over TCP. The default is `20MiB`.
+
+[float]
+[id="{beatname_lc}-input-{type}-host"]
+==== `host`
+
+The host and TCP port to listen on for event streams.
+
+[float]
+[id="{beatname_lc}-input-{type}-line-delimiter"]
+==== `line_delimiter`
+
+Specify the characters used to split the incoming events. The default is '\n'.
+
+[float]
+[id="{beatname_lc}-input-{type}-timeout"]
+==== `timeout`
+
+The number of seconds of inactivity before a remote connection is closed. The default is `300s`.
+
+[id="{beatname_lc}-input-{type}-common-options"]
+include::../inputs/input-common-options.asciidoc[]
+
+:type!:
--- a/filebeat/input/tcp/input.go
+++ b/filebeat/input/tcp/input.go
@@ -1,0 +1,109 @@
+package tcp
+
+import (
+	"time"
+
+	"github.com/elastic/beats/filebeat/channel"
+	"github.com/elastic/beats/filebeat/harvester"
+	"github.com/elastic/beats/filebeat/input"
+	"github.com/elastic/beats/filebeat/inputsource/tcp"
+	"github.com/elastic/beats/filebeat/util"
+	"github.com/elastic/beats/libbeat/beat"
+	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/common/atomic"
+	"github.com/elastic/beats/libbeat/common/cfgwarn"
+	"github.com/elastic/beats/libbeat/logp"
+)
+
+func init() {
+	err := input.Register("tcp", NewInput)
+	if err != nil {
+		panic(err)
+	}
+}
+
+// Input for TCP connection
+type Input struct {
+	server  *tcp.Server
+	started atomic.Bool
+	outlet  channel.Outleter
+	config  *config
+	log     *logp.Logger
+}
+
+// NewInput creates a new TCP input
+func NewInput(
+	cfg *common.Config,
+	outlet channel.Factory,
+	context input.Context,
+) (input.Input, error) {
+	cfgwarn.Experimental("TCP input type is used")
+
+	out, err := outlet(cfg, context.DynamicFields)
+	if err != nil {
+		return nil, err
+	}
+
+	forwarder := harvester.NewForwarder(out)
+
+	config := defaultConfig
+	err = cfg.Unpack(&config)
+	if err != nil {
+		return nil, err
+	}
+
+	cb := func(data []byte, metadata tcp.Metadata) {
+		event := createEvent(data, metadata)
+		forwarder.Send(event)
+	}
+
+	server, err := tcp.New(cb, &config.Config)
+	if err != nil {
+		return nil, err
+	}
+
+	return &Input{
+		server:  server,
+		started: atomic.MakeBool(false),
+		outlet:  out,
+		config:  &config,
+		log:     logp.NewLogger("tcp input").With(config.Config.Host),
+	}, nil
+}
+
+// Run start a TCP input
+func (p *Input) Run() {
+	if !p.started.Load() {
+		p.log.Info("Starting TCP input")
+		err := p.server.Start()
+		if err != nil {
+			p.log.Errorw("Error starting the TCP server", "error", err)
+		}
+		p.started.Swap(true)
+	}
+}
+
+// Stop stops TCP server
+func (p *Input) Stop() {
+	p.log.Info("Stopping TCP input")
+	defer p.outlet.Close()
+	defer p.started.Swap(false)
+	p.server.Stop()
+}
+
+// Wait stop the current server
+func (p *Input) Wait() {
+	p.Stop()
+}
+
+func createEvent(raw []byte, metadata tcp.Metadata) *util.Data {
+	data := util.NewData()
+	data.Event = beat.Event{
+		Timestamp: time.Now(),
+		Fields: common.MapStr{
+			"message": string(raw),
+			"source":  metadata.RemoteAddr.String(),
+		},
+	}
+	return data
+}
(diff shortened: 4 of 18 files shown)


ph:
I've squashed everything into a single commit, ready to get that reviewed, there is one test failling related to metricbeat and apache 2.4.12.

```
 File "/usr/local/lib/python2.7/dist-packages/compose/service.py", line 998, in build
    'memory': parse_bytes(memory) if memory else None
  File "/usr/local/lib/python2.7/dist-packages/docker/api/build.py", line 154, in build
    with open(dockerfile, 'r') as df:
IOError: [Errno 2] No such file or directory: 'Dockerfile.2.4.12'
```

I've created #6722 to tackle this.

ph:
@ruflin @urso @kvch you were in the original PR it was just clearer to create a new one with the externalized TCP input, its pretty close to the #6439 in the behavior. I did not enforce an interface, before freezing an interface we need to extract the Filewatcher.

ph:
jenkins test this please

CHANGELOG.asciidoc
```diff
@@ -51,6 +51,8 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
 - Remove the undefined `username` option from the Redis input and clarify the documentation. {pull}6662[6662]
 - Add validation for Stdin, when Filebeat is configured with Stdin and any other inputs, Filebeat
   will now refuses to start. {pull}6463[6463]
+- Refactor the usage of prospector to input in the YAML reference {pull}6121[6121]
```
ruflin: This line is not related to this PR

CHANGELOG.asciidoc
```diff
@@ -51,6 +51,8 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
...
+- Addition of the TCP input {pull}6266[6266]
```
ruflin: PR number should be adjusted.

filebeat/_meta/common.reference.p2.yml
```diff
@@ -240,6 +240,20 @@ filebeat.inputs:
   # Maximum size of the message received over UDP
   #max_message_size: 10240
 
+#------------------------------ TCP prospector --------------------------------
+# Experimental: Config options for the TCP input
+#- type: tcp
+  #enabled: false
+
+  # The host and port to receive the new event
+  #host: "localhost:9000"
```
ruflin: I remember @andrewkroh suggested to not have a default and require the use to specify it?

filebeat/docs/inputs/input-tcp.asciidoc
```diff
@@ -0,0 +1,55 @@
...
+==== Configuration options
+
+The `tcp` input supports the following configuration options plus the
+<<{beatname_lc}-input-{type}-common-options>> described later.
+
+[float]
+[id="{beatname_lc}-input-{type}-max-message-size"]
+==== `max_message_size`
+
+The maximum size of the message received over TCP. The default is `20971520`.
```
ruflin: Perhaps mention that this is 20MB

filebeat/docs/inputs/input-tcp.asciidoc
```diff
@@ -0,0 +1,55 @@
...
+
+[float]
+[id="{beatname_lc}-input-{type}-line-delimiter"]
+==== `line_delimiter`
+
+Specify the characters used to split the incoming events. The default is '\n'.
+
+[float]
+[id="{beatname_lc}-input-{type}-timeout"]
+==== `timeout`
```
ruflin: I didn't see this config in the reference config file. Also make sure to use `300s` (with unit) and not just 300. See other timeout config options. Since we had issue in the past without units we now always try to have units.

filebeat/input/tcp/input.go
```diff
@@ -0,0 +1,111 @@
...
+		log:     logp.NewLogger("tcp input").With(config.Config.Host),
+	}, nil
+}
+
+// Run start a TCP input
+func (p *Input) Run() {
+	if !p.started.Load() {
+		p.log.Info("Starting TCP input")
+		err := p.server.Start()
+		if err != nil {
```
ruflin: Shouldn't this return the error?

filebeat/input/tcp/input.go
```diff
@@ -0,0 +1,111 @@
...
+func (p *Input) Wait() {
+	p.Stop()
+}
+
+func createEvent(raw []byte, metadata tcp.Metadata) *util.Data {
+	data := util.NewData()
+	data.Event = beat.Event{
+		Timestamp: time.Now(),
+		Meta: common.MapStr{
+			"ip_address": metadata.IPAddress,
```
ruflin: I suggest we have this information also in each event under something like `source.ip` or similar?

filebeat/inputsource/tcp/conn.go
```diff
@@ -0,0 +1,72 @@
...
+// ResetableLimitedReader is based on LimitedReader but allow to reset the byte read and return a specific
+// error when we reach the limit.
+type ResetableLimitedReader struct {
+	reader        io.Reader
+	maxReadBuffer uint64
+	byteRead      uint64
+}
+
+// NewResetableLimitedReader returns a new ResetableLimitedReader
+func NewResetableLimitedReader(reader io.Reader, maxReadBuffer uint64) *ResetableLimitedReader {
```
ruflin: We already have `harvester/reader/` package. I wonder if these have the same interface and should go there.

filebeat/inputsource/tcp/scan.go
```diff
@@ -0,0 +1,31 @@
+package tcp
+
+import (
+	"bufio"
+	"bytes"
+)
+
+// ScanDelimiter return a function to split line using a custom delimiter, the delimiter
+// is stripped from the returned value.
+func scanDelimiter(delimiter []byte) bufio.SplitFunc {
```
ruflin: Is this similar to the strip_newline reader?

filebeat/inputsource/tcp/client.go
```diff
@@ -0,0 +1,82 @@
...
+	log            *logp.Logger
+	callback       CallbackFunc
+	done           chan struct{}
+	metadata       Metadata
+	splitFunc      bufio.SplitFunc
+	maxReadMessage uint64
+	timeout        time.Duration
+}
+
+// NewClient returns a new client instance for the remote connection.
```
kvch: `s/NewClient/newClient` so it matches the name.

filebeat/inputsource/tcp/conn.go
ph: I would prefer to keep them together until it's reused by something else.

filebeat/_meta/common.reference.p2.yml
ph: We currently don't have any default on this value, I've provided an example hosts/port, but It's not a hard default.

```golang
var defaultConfig = config{
	ForwarderConfig: harvester.ForwarderConfig{
		Type: "tcp",
	},
	Config: tcp.Config{
		LineDelimiter:  "\n",
		Timeout:        time.Minute * 5,
		MaxMessageSize: 20 * 1024 * 1024,
	},
}
```

Following your comment, I did a small change and added a more explicit error for the host.

filebeat/docs/inputs/input-tcp.asciidoc
ph: switched to `go-humanize` instead and allow 20MiB

filebeat/input/tcp/input.go
ph: This would require changes in the upstream interface, I can create a followup issue on that?

```
// Input is the interface common to all input
type Input interface {
	Run()
	Stop()
	Wait()
}
```

filebeat/inputsource/tcp/scan.go
ph: Not really, I will rename this method to `factoryDelimiter()` which is what it does.

ph: It returns a custom `split func` that support a custom delimiter which can be a multibytes delimiter.

filebeat/input/tcp/input.go
ph: +1 for source.ip when following ECS, but currently this will clash with `source` which is the original read file. So using `source.ip` will need to wait for 7.0, maybe we can use `client.ip` ?

```yaml
      description: > source
        The file from which the line was read. This field contains the absolute path to the file.
        For example: `/var/log/system.log`.
```

andrewkroh: The collision is unfortunate. Under ECS `source` will be moving to `file.path`. I think we have three options

1. Put the remote IP and port into `source` (e.g. `source: 127.0.0.1:38849`)
1. Leave it under `@metadata`. The value will be there if someone really needs it (they can use the rename processor to get it stored in ES). Then in 7.0 we move this to `source.ip`. Because it's tucked away in `@metadata` the impact of renaming it in 7.0 is minimized.
1. Store it in `client.ip` and rename it to `source.ip` for 7.0. This adds more fields that need to be renamed for 7.0, but on the plus side the value is more visible if people want to use it.

I like number 1.

ph: 1. make sense, I didn't see like that.

ph: This is effectively the `source` file for it.

ph:
```json
{
  "@timestamp": "2018-04-03T20:32:55.908Z",
  "@metadata": {
    "beat": "filebeat",
    "type": "doc",
    "version": "7.0.0-alpha1"
  },
  "source": "127.0.0.1:61656",
  "prospector": {
    "type": "tcp"
  },
  "input": {
    "type": "tcp"
  },
  "beat": {
    "name": "sashimi",
    "hostname": "sashimi",
    "version": "7.0.0-alpha1"
  },
  "message": "heee"
}
```

filebeat/inputsource/tcp/client.go
```diff
@@ -0,0 +1,79 @@
...
+	client := &client{
+		conn:           conn,
+		log:            log.With("address", conn.RemoteAddr()),
+		callback:       callback,
+		done:           make(chan struct{}),
+		splitFunc:      splitFunc,
+		maxReadMessage: maxReadMessage,
+		timeout:        timeout,
+		metadata: Metadata{
+			IPAddress: conn.RemoteAddr().String(),
```
andrewkroh: The way this is populated it's more than an IP address so I think the name is confusing. I'd go with `RemoteAddr`. Or [split](https://golang.org/pkg/net/#SplitHostPort) it here and have two fields `RemoteIP` and `RemotePort`.

ruflin: If I understand this correctly, the ip address provided here is in the end the host that sent the log message. So my suggestion would be to use `host.ip` for this. So the `host.ip` would be the same if the file is read locally or sent over tcp.

For the `port` I'm not sure if this is relevant in this context and could be skipped. If we add the port, it would be the port of the service running, so `service.port`?

filebeat/input/tcp/input.go
ruflin: In general the way we use `Run()` is if it's blocking and often returns an error. In contrast we use `Start()` if it's not blocking. I would expect all three in the above interface to potentially return an error.

ph: > In general the way we use Run() is if it's blocking and often returns an error. In contrast we use Start() if it's not blocking. I would expect all three in the above interface to potentially return an error.

Agree on that point but I think changing that interface is out of scope for this PR, so do you agree the plan of action is to do a followup issue.

filebeat/inputsource/tcp/client.go
ph: Maybe you are right port in that context isn't necessary and we can drop it and I would also say the same for port of the service running. 

@ruflin For the future I still believe that `source.ip` is what we should aim, I think in using `source` as the IP until ECS is deployed is a good compromised into not creating new fields that we have to deprecated right away.

ph: agreement on zoom: 
- set remote ip + remote port to the `source` field in the event.
- on Metadata uses `RemoteAddr` instead of `IPAdress`.

Example format : `127.0.0.1:99998`

filebeat/input/tcp/input.go
ruflin: Yes, if we follow up on this ;-)

ph: created https://github.com/elastic/beats/issues/6771

filebeat/input/tcp/input.go
```diff
@@ -73,22 +74,28 @@ func NewInput(
...
 			p.log.Errorw("Error starting the TCP server", "error", err)
 		}
-		p.started.Swap(true)
+		p.started = true
 	}
 }
 
 // Stop stops TCP server
 func (p *Input) Stop() {
 	p.log.Info("Stopping TCP input")
```
andrewkroh: The logging is not congruent with the UDP input. In UDP the message is logged after acquiring the lock. IDK which order, but I would like the behavior to be the same.

ph: Correct, done in the next commit.

PR #6701, opened 2018-03-29
Metricbeat: Improvements in docker diskio metricset

jsoriano:
This PR continues with #6608 in the aim to provide docker metrics more consistent with `docker stats`. Currently we report for disk I/O the number of operations per second.

This PR adds accumulated I/O stats per number of operations and per data volume since the container started, as `docker stats` does. I think this information is more valuable, as it permits to have further aggregations when visualizing it. It also addresses some other issues in the metricset.

Fields in events have been reorganized and old ones have been marked as deprecated.

We could also add some extra functionality, as obtaining metrics also on Windows hosts (as is this module now it would only work with docker daemons running on Linux and not sure about Mac). In Linux we could also collect metrics per block device.

- [x] Add accumulated I/O stats to diskio metricset
- [x] Fix diskio metricset when docker reads from multiple devices
- [x] Some naming and documentation fixes
- [x] Fix memory leak in blkio service (last seen stats were being stored, but never removed)
- [x] More tests

To follow-up in other issues:

- ~~Group by device in Linux?~~ To be done if requested in the future
- Support for Windows storage stats? Follow-up issue #6815

### DIFF
--- a/metricbeat/module/docker/diskio/_meta/fields.yml
+++ b/metricbeat/module/docker/diskio/_meta/fields.yml
@@ -15,6 +15,7 @@
             Number of reads during the life of the container
         - name: bytes
           type: long
+          format: bytes
           description: >
             Bytes read during the life of the container
         - name: rate
@@ -37,6 +38,7 @@
             Number of writes during the life of the container
         - name: bytes
           type: long
+          format: bytes
           description: >
             Bytes written during the life of the container
         - name: rate
@@ -59,6 +61,7 @@
             Number of I/O operations during the life of the container
         - name: bytes
           type: long
+          format: bytes
           description: >
             Bytes read and written during the life of the container
         - name: rate
--- a/metricbeat/module/docker/diskio/_meta/fields.yml
+++ b/metricbeat/module/docker/diskio/_meta/fields.yml
@@ -4,15 +4,69 @@
     Disk I/O metrics.
   release: ga
   fields:
+    - name: read
+      type: group
+      description: >
+        Accumulated reads during the life of the container
+      fields:
+        - name: ops
+          type: long
+          description: >
+            Number of reads during the life of the container
+        - name: bytes
+          type: long
+          description: >
+            Bytes read during the life of the container
+        - name: rate
+          type: long
+          description: >
+            Number of current reads per second
     - name: reads
       type: scaled_float
+      deprecated: true
       description: >
-        Number of reads.
+        Number of current reads per second
+    - name: write
+      type: group
+      description: >
+        Accumulated writes during the life of the container
+      fields:
+        - name: ops
+          type: long
+          description: >
+            Number of writes during the life of the container
+        - name: bytes
+          type: long
+          description: >
+            Bytes written during the life of the container
+        - name: rate
+          type: long
+          description: >
+            Number of current writes per second
     - name: writes
       type: scaled_float
+      deprecated: true
       description: >
-        Number of writes.
+        Number of current writes per second
+    - name: summary
+      type: group
+      description: >
+        Accumulated reads and writes during the life of the container
+      fields:
+        - name: ops
+          type: long
+          description: >
+            Number of I/O operations during the life of the container
+        - name: bytes
+          type: long
+          description: >
+            Bytes read and written during the life of the container
+        - name: rate
+          type: long
+          description: >
+            Number of current operations per second
     - name: total
       type: scaled_float
+      deprecated: true
       description: >
-        Number of reads and writes combined.
+        Number of reads and writes per second
--- a/metricbeat/module/docker/diskio/data.go
+++ b/metricbeat/module/docker/diskio/data.go
@@ -21,6 +21,21 @@ func eventMapping(stats *BlkioStats) common.MapStr {
 		"reads":  stats.reads,
 		"writes": stats.writes,
 		"total":  stats.totals,
+		"read": common.MapStr{
+			"ops":   stats.serviced.reads,
+			"bytes": stats.servicedBytes.reads,
+			"rate":  stats.reads,
+		},
+		"write": common.MapStr{
+			"ops":   stats.serviced.writes,
+			"bytes": stats.servicedBytes.writes,
+			"rate":  stats.writes,
+		},
+		"summary": common.MapStr{
+			"ops":   stats.serviced.totals,
+			"bytes": stats.servicedBytes.totals,
+			"rate":  stats.totals,
+		},
 	}
 
 	return event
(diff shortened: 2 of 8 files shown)


metricbeat/module/docker/diskio/helper.go
```diff
@@ -45,56 +48,62 @@ func (io *BLkioService) getBlkioStatsList(rawStats []docker.Stat, dedot bool) []
 	return formattedStats
 }
 
-func (io *BLkioService) getBlkioStats(myRawStat *docker.Stat, dedot bool) BlkioStats {
+// TODO: User StorageStats instead of BlkioStats on Windows
```
jsoriano: According to docker stats API documentation, `BlkioStats` is only populated in Linux, for Windows there is another structure under `StorageStats`.

metricbeat/module/docker/diskio/helper.go
```diff
@@ -45,56 +48,62 @@ func (io *BLkioService) getBlkioStatsList(rawStats []docker.Stat, dedot bool) []
...
-func (io *BLkioService) getNewStats(time time.Time, blkioEntry []types.BlkioStatEntry) BlkioRaw {
+func (io *BlkioService) getNewStats(time time.Time, blkioEntry []types.BlkioStatEntry) BlkioRaw {
 	stats := BlkioRaw{
 		Time:   time,
 		reads:  0,
 		writes: 0,
 		totals: 0,
 	}
+
 	for _, myEntry := range blkioEntry {
```
jsoriano: Entries are per operation type, but also per Major/Minor (per device), so they should be summed by type. We could also decide to send metrics per device. Opinions?

metricbeat/module/docker/diskio/_meta/fields.yml
```diff
@@ -7,12 +7,56 @@
     - name: reads
       type: scaled_float
       description: >
-        Number of reads.
+        Number of current reads per second
```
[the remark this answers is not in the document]
jsoriano: +1 to the deprecated field, I think we need something like this.

The API exports it as gauges, but I'd say it doesn't do any calculation for block I/O. The problem I see with counters and this kind of metrics is that you have to be very careful about how and when you collect them.

metricbeat/module/docker/diskio/data.go
```diff
@@ -21,6 +21,20 @@ func eventMapping(stats *BlkioStats) common.MapStr {
 		"reads":  stats.reads,
 		"writes": stats.writes,
 		"total":  stats.totals,
+		"serviced": common.MapStr{
```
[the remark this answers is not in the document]
jsoriano: Umm, this could be an approach, yes, I'll update the PR with this.

metricbeat/module/docker/diskio/helper.go
jsoriano: The biggest issue about Windows at the moment is testing it, maybe better to leave the TODO here and open a follow-up issue, yes.

metricbeat/module/docker/diskio/helper.go
jsoriano: These blkio entries are defined as:
```
// BlkioStatEntry is one small entity to store a piece of Blkio stats
// Not used on Windows.
type BlkioStatEntry struct {
    Major uint64 `json:"major"`
    Minor uint64 `json:"minor"`
    Op    string `json:"op"`
    Value uint64 `json:"value"`
}
```
I mean Major/Minor numbers as the identifier of a device in a posix system, in this case they identify the block device related to these stats.

So we have information per device, and we could send that, though the mapping between Major/Minor and device files, mountpoints or something useful would be tricky, and we couldn't provide the same functionality for Windows.

metricbeat/module/docker/diskio/helper.go
```diff
@@ -45,56 +48,62 @@ func (io *BLkioService) getBlkioStatsList(rawStats []docker.Stat, dedot bool) []
...
 		if myEntry.Op == "Write" {
-			stats.writes = myEntry.Value
+			stats.writes += myEntry.Value
```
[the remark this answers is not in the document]
jsoriano: We only get the value of the last request. The sum is to sum the same operation on multiple devices.

metricbeat/module/docker/diskio/data.go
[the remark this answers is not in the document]
jsoriano: Added a new commit with this approach https://github.com/elastic/beats/pull/6701/commits/2a0306bb6c45e8cfbe6d759f2c05c6b074afa904

metricbeat/module/docker/diskio/helper.go
jsoriano: Yes, in any case I'm not sure about how to do it in a useful way with proper mapping between device ids and device files or mountpoints. We can leave it till we see someone needing it.

metricbeat/module/docker/diskio/_meta/fields.yml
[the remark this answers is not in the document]
jsoriano: Added

metricbeat/module/docker/diskio/helper.go
```diff
@@ -45,56 +50,64 @@ func (io *BLkioService) getBlkioStatsList(rawStats []docker.Stat, dedot bool) []
...
+	oldBlkioStats, exist := io.BlkioStatsPerContainer[myRawStat.Container.ID]
 	if exist {
 		myBlkioStats.reads = io.getReadPs(&oldBlkioStats, &newBlkioStats)
 		myBlkioStats.writes = io.getWritePs(&oldBlkioStats, &newBlkioStats)
 		myBlkioStats.totals = io.getTotalPs(&oldBlkioStats, &newBlkioStats)
 	}
 
-	io.BlkioSTatsPerContainer[myRawStat.Container.ID] = newBlkioStats
+	// FIXME: Memory leak
+	io.BlkioStatsPerContainer[myRawStat.Container.ID] = newBlkioStats
```
jsoriano: I think we have a memory leak here, entries are never removed, I'll take a look to this too.

jsoriano:
I have added a new test, to cover all calculations done on raw stats, there are some other tests that could be revisited, but I left them by now. Regarding Windows, I have created a follow-up issue #6815.

This would be ready for review.

metricbeat/module/docker/diskio/_meta/fields.yml
```diff
@@ -4,15 +4,69 @@
...
+      type: group
+      description: >
+        Accumulated reads during the life of the container
+      fields:
+        - name: ops
+          type: long
+          description: >
+            Number of reads during the life of the container
+        - name: bytes
+          type: long
```
ruflin: Could you add here `format: bytes` to kibana can make sense of it. Same is true for other bytes values.

jsoriano:
`data.json` updated and bytes fileds formatted as bytes

metricbeat/module/docker/diskio/_meta/fields.yml
jsoriano: Done

jsoriano:
Failing tests not related to these changes.

PR #6707, opened 2018-03-29
Make cherrypick_pr and open_pr scripts add version labels

tsg:
When cherry-picking, automatically reads the version from the target branch
and adds it as a label to the original PR.

When opening a new PR, automatically add the version label.

### DIFF
--- a/dev-tools/cherrypick_pr
+++ b/dev-tools/cherrypick_pr
@@ -2,6 +2,7 @@
 """Cherry pick and backport a PR"""
 
 import sys
+import os
 import argparse
 from os.path import expanduser
 import re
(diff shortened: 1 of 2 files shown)


ruflin:
One thing I worry with labels on each PR is what we had close to the 6.0 release. Some labels had `v6.0.0-*` label on them but the PR's did not make it in in time. Same for backports which don't get merged in time. In 6.0 the issue was the some people expected on the label that it will be in the release.

Not sure I have the full background on why we need labels with the version number it should go in on the PR.

tsg:
@ruflin I see, ideally the label would be added at merge time instead of opening time, but that doesn't quite match our workflow (we merge with the green button).  I'd say that compared to the grand time line, the difference is not to big, so there shouldn't be a ton of affected PRs, but it's something to keep in mind.

I suspect that we had issues before because the labels were completely manual.

PR #6708, opened 2018-03-30
Add support for docker autodiscover to monitor containers on host network

vjsamuel:
In today's approach autodiscover docker provider can not monitor containers running on Host network. If a container runs on host network it would not have `.NetworkSettings.IPAddress` as seen below:

```
# docker inspect --format '{{ .NetworkSettings.IPAddress }}' jobmanager #runs as --net=host

# docker inspect --format '{{ .NetworkSettings.IPAddress }}' filebeat #normal container
172.17.0.2
```

To be able to monitor these containers we detect a host address that can be used be the docker auto discover as follows:
* If filebeat/metricbeat runs as a container, try to check with the the Docker daemon if the hostname is the prefix of one of the running containers. If yes, get the `Gateway` of the network interface. That will be the host address that can be used.
* If filebeat/metricbeat are running as processes on the host, get an IP address on the host that is not a loop back address and use them to monitor.


ruflin:
As we already had this feature in 6.2, could you add a CHANLOGE entry? As far as I understand the change, it's an addition / enhancement and will not break anything existant.

For the `data.host` change: This change is only indirectly related to the previous one. So probably worth 2 changelog entries?

ruflin:
jenkins, test it

vjsamuel:
@ruflin i think it is necessary to break this PR into 2. Otherwise you wont be able to back port this fix to 6.2. Thoughts?

ruflin:
I'm always up for 2 PR's if they can be separated, especially if one of them needs backport.

vjsamuel:
@ruflin this PR has to be back ported to 6.2 and I have pulled out the hints builder changes to a separate PR[https://github.com/elastic/beats/pull/6727] to allow the same.

ruflin:
jenkins, test it

libbeat/common/docker/util.go
```diff
@@ -0,0 +1,47 @@
...
+
+func getLocalIP() string {
+	addrs, err := net.InterfaceAddrs()
+	if err != nil {
+		return ""
+	}
+	for _, address := range addrs {
+		// Check the address type and if it is not a loop back then return it
+		if ipnet, ok := address.(*net.IPNet); ok && !ipnet.IP.IsLoopback() {
+			if ipnet.IP.To4() != nil {
```
jsoriano: Any reason to limit to IPv4 only?

libbeat/common/docker/watcher.go
```diff
@@ -291,8 +293,15 @@ func (w *watcher) listContainers(options types.ContainerListOptions) ([]*Contain
...
+		var gateways []string
 		for _, net := range c.NetworkSettings.Networks {
-			ipaddresses = append(ipaddresses, net.IPAddress)
+			if net.IPAddress != "" {
+				ipaddresses = append(ipaddresses, net.IPAddress)
+			}
+
+			if net.Gateway != "" {
+				gateways = append(gateways, net.Gateway)
+			}
```
jsoriano: Here, when requesting information from the daemon we could also get the Hostname, at least if we use inspect, not sure with list.

libbeat/common/docker/util.go
```diff
@@ -0,0 +1,47 @@
...
+			}
+		}
+	}
+
+	// If the process is not running in a container then return a valid IPv4 address that can be used to access
+	// the service as it would be running on host network.
+	return getLocalIP()
+}
+
+func getLocalIP() string {
```
jsoriano: This method is a bit tricky, are we sure that `net.InterfaceAddrs()` always returns the addresses in the same order? what happens if an address is added or removed?

I think there is no safe method to guess the address of a host with multiple addresses :neutral_face: maybe for that case we should allow to configure a host address in metricbeat (e.g. sensu has the [address setting](https://docs.sensu.io/sensu-core/1.2/reference/clients/#client-attributes), and kubelet has the `--node-ip` flag), and take it from there @ruflin wdyt?

libbeat/common/docker/watcher.go
jsoriano: A container running in host network gets the hostname of the host.

ruflin:
jenkins, test it

libbeat/common/docker/watcher.go
```diff
@@ -292,7 +293,20 @@ func (w *watcher) listContainers(options types.ContainerListOptions) ([]*Contain
...
+				ipaddresses = append(ipaddresses, net.IPAddress)
+			}
+		}
+
+		// If there are no network interfaces, assume that the container is on host network
+		// Inspect the container directly and use the hostname as the IP address in order
+		if len(ipaddresses) == 0 {
+			info, err := w.client.ContainerInspect(w.ctx, c.ID)
+			if err == nil {
+				ipaddresses = append(ipaddresses, info.Config.Hostname)
```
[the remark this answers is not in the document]
vjsamuel: i checked this behavior and it seems to work.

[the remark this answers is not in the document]
vjsamuel: neither approach work (using hostname and gateway ip) on mac however

ruflin:
@jsoriano @vjsamuel Can you push this one forward together? @exekias FYI

exekias:
I'm ok with this change if we don't find something more reliable. I was planning to do some research on this and try to come up with some alternative way to retrieve host IP, not sure it's possible though :)

exekias:
I've been researching this with @vjsamuel, it seems we can get host IP by checking the default gateway:
```
root@85ae06759fea:/# netstat -nr | grep '^0\.0\.0\.0' | awk '{print $2}'
172.17.0.1
```
It should be possible to read it from proc filesystem, I guess this will support more use cases, still to be tested

vjsamuel:
@exekias i was revisiting the code today, i feel that this might be the cleanest solution code-wise. what do you think? if you are not convinced, i can make the above discussed change

CHANGELOG.asciidoc
```diff
@@ -161,6 +161,9 @@ https://github.com/elastic/beats/compare/v6.2.3...master[Check the HEAD diff]
 - Add experimental Jolokia Discovery autodiscover provider. {pull}7141[7141]
 - Add owner object info to Kubernetes metadata. {pull}7231[7231]
 - Add beat export dashboard command. {pull}7239[7239]
+- Add a default seccomp (secure computing) filter on Linux that prohibits
+  execve, execveat, fork, and vfork syscalls. A custom policy can be configured. {issue}5213[5213]
+- Add support for docker autodiscover to monitor containers on host network {pull}6708[6708]
```
exekias: some entries sneaked in here

libbeat/common/docker/watcher.go
```diff
@@ -290,7 +291,20 @@ func (w *watcher) listContainers(options types.ContainerListOptions) ([]*Contain
...
+			} else {
+				logp.Debug("docker", "unable to inspect container %s due to error %v", c.ID, err)
```
exekias: this should probably be a warning message? as it should not happen frequently, and can cause some debugging nightmares

exekias:
jenkins, test this please

PR #6709, opened 2018-03-30
Cherry-pick #6130 to 6.2: Change kubernetes.node.cpu.allocatable.cores to float

ruflin:
Cherry-pick of PR #6130 to 6.2 branch. Original message: 

Closes https://github.com/elastic/beats/issues/6123


jsoriano:
Could we have used `scaled_float` here?

ruflin:
@jsoriano Probably yes. I'm hesitant to change it in the backport but then would rather merge this one, open a second PR against master and backport this too.

I wonder if we should treat float to scaled_float as a breaking change. In general I would say no, so we go have `float` in 6.2 and do the scaled_float improvement in 6.3 to save space.

jsoriano:
Ok, merging this then

PR #6714, opened 2018-03-30
Move Logstash node_stats under node.stats

ruflin:
* This is the same as for elasticsearch
* Update data.json accordingly
* Update Logstash version to 6.2.3
* Update data.json generator to support namespace setting

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -73,6 +73,7 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
 - Support apache status pages for versions older than 2.4.16. {pull}6450[6450]
 - Add support for huge pages on Linux. {pull}6436[6436]
 - Refactor docker CPU calculations to be more consistent with `docker stats`. {pull}6608[6608]
+- Update logstash.node_stats metricset to write data under `logstash.node.stats.*`. {pull}[]
 
 *Packetbeat*
 
--- a/metricbeat/mb/testing/data_generator.go
+++ b/metricbeat/mb/testing/data_generator.go
@@ -78,9 +78,11 @@ func StandardizeEvent(ms mb.MetricSet, e mb.Event, modifiers ...mb.EventModifier
 	}
 
 	e.Timestamp = startTime
-	e.Namespace = ms.Registration().Namespace
 	e.Took = 115 * time.Microsecond
 	e.Host = ms.Host()
+	if e.Namespace == "" {
+		e.Namespace = ms.Registration().Namespace
+	}
 
 	fullEvent := e.BeatEvent(ms.Module().Name(), ms.Name(), modifiers...)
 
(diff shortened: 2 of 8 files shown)


metricbeat/mb/testing/data_generator.go
```diff
@@ -78,7 +78,6 @@ func StandardizeEvent(ms mb.MetricSet, e mb.Event, modifiers ...mb.EventModifier
 	}
 
 	e.Timestamp = startTime
-	e.Namespace = ms.Registration().Namespace
```
ruflin: @andrewkroh Can you check if this change has unintended side effects on the auditbeat side?

CHANGELOG.asciidoc
```diff
@@ -73,6 +73,7 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
 - Support apache status pages for versions older than 2.4.16. {pull}6450[6450]
 - Add support for huge pages on Linux. {pull}6436[6436]
 - Refactor docker CPU calculations to be more consistent with `docker stats`. {pull}6608[6608]
+- Update logstash.node_stats metricset to write data under `logstash.docker.stats.*`. {pull}[]
```
ph: I think you meant `logstash.node.stats.*` not `logstash.docker.stats.*`?

metricbeat/module/logstash/_meta/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM docker.elastic.co/logstash/logstash:6.0.0
+FROM docker.elastic.co/logstash/logstash:6.2.3
```
ph: We have to figure out a way to proactively tests on the snapshots release for that.

CHANGELOG.asciidoc
ruflin: ups, fixed.

metricbeat/module/logstash/_meta/Dockerfile
ruflin: Yeah, it is a general thing we have to come up with a solution, meaning testing each module against multiple versions.

metricbeat/mb/testing/data_generator.go
andrewkroh: This will affect MetricSets that provide a Namespace at registration time. I think that by adding a conditional this should be safe.

```
if e.Namespace == "" {
    e.Namespace = ms.Registration().Namespace
}
```

ruflin: @andrewkroh Add the if clause and switched node_stats to also use the registry for the namespace as it's constant. I will do the same for node_stats on the ES side.

ruflin:
Looks like the node_stats test fail. I probably have to adjust some setting, will investigate.

andrewkroh:
Looks like the failure is due to fields.yml.

`Key 'logstash.node.stats.events.in' found in event is not documented!`

ruflin:
@andrewkroh @ph Tests should be fixed now. Ready for an other round.

PR #6716, opened 2018-03-30
Introduce environment variable to skip compose starting environment

ruflin:
By default the golang integration tests in metricbeat use docker compose to start their own environment for testing. This is useful for quick and automated testing. But in some cases a local instance of the service is used for developing and testing a metricset. This PR introduces the `NO_COMPOSE` environment variable which can be used to skip the starting of the service with compose.

To generate data for logstash.node_stats metricset from a local LS instance, the following command can be used:

```
NO_COMPOSE=1 go test -tags=integration github.com/elastic/beats/metricbeat/module/logstash/node_stats/... -data
```

### DIFF
--- a/libbeat/tests/compose/compose.go
+++ b/libbeat/tests/compose/compose.go
@@ -11,6 +11,8 @@ import (
 	"testing"
 	"time"
 
+	"strconv"
+
 	"github.com/docker/libcompose/docker"
 	"github.com/docker/libcompose/docker/ctx"
 	"github.com/docker/libcompose/project"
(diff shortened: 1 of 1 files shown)


libbeat/tests/compose/compose.go
```diff
@@ -43,6 +43,12 @@ func EnsureUp(t *testing.T, services ...string) {
 // EnsureUpWithTimeout starts all the requested services (must be defined in docker-compose.yml)
 // Wait for `timeout` seconds for health
 func EnsureUpWithTimeout(t *testing.T, timeout int, services ...string) {
+	// The NO_COMPOSE env variables makes it possible to skip the starting of the environment.
+	// This is useful it the service is already running locally.
```
andrewkroh: `s/it/if/`

libbeat/tests/compose/compose.go
```diff
@@ -43,6 +43,12 @@ func EnsureUp(t *testing.T, services ...string) {
...
+	noCompose := os.Getenv("NO_COMPOSE")
```
andrewkroh: Using `ParseBool` would give a bit more flexibility in the values. https://golang.org/pkg/strconv/#ParseBool

```
	if noCompose, err := strconv.ParseBool(os.Getenv("NO_COMPOSE")); err == nil && noCompose {
		return
	}
```

libbeat/tests/compose/compose.go
ruflin: fixed

libbeat/tests/compose/compose.go
ruflin: changed

PR #6718, opened 2018-03-30
Set default metricsets in Docker module

ruflin:
The following metricests are set as default: `container`, `cpu`, `diskio`, `healthcheck`, `info`, `memory` and `network`. `image` is not set as default.

### DIFF
--- a/metricbeat/module/docker/_meta/config.yml
+++ b/metricbeat/module/docker/_meta/config.yml
@@ -1,13 +1,5 @@
 - module: docker
-  metricsets: ["container", "cpu", "diskio", "healthcheck", "info", "memory", "network"]
   hosts: ["unix:///var/run/docker.sock"]
-  period: 10s
 
   # Replace dots in labels with `_`. Set to false to keep dots
   labels.dedot: true
-
-  # To connect to Docker over TLS you must specify a client and CA certificate.
-  #ssl:
-    #certificate_authority: "/etc/pki/root/ca.pem"
-    #certificate:           "/etc/pki/client/cert.pem"
-    #key:                   "/etc/pki/client/cert.key"
--- a/metricbeat/module/docker/image/image.go
+++ b/metricbeat/module/docker/image/image.go
@@ -14,9 +14,9 @@ import (
 // init registers the MetricSet with the central registry.
 // The New method will be called after the setup of the module and before starting to fetch data
 func init() {
-	if err := mb.Registry.AddMetricSet("docker", "image", New); err != nil {
-		panic(err)
-	}
+	mb.Registry.MustAddMetricSet("docker", "image", New,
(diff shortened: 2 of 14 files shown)


ruflin:
Rebased without changelog. Changelog entry is here: https://github.com/elastic/beats/issues/6668#issuecomment-378182918 and will be added as one PR in the end.

metricbeat/module/docker/image/image.go
```diff
@@ -14,9 +14,9 @@ import (
 // init registers the MetricSet with the central registry.
 // The New method will be called after the setup of the module and before starting to fetch data
 func init() {
-	if err := mb.Registry.AddMetricSet("docker", "image", New); err != nil {
-		panic(err)
-	}
+	mb.Registry.MustAddMetricSet("docker", "image", New,
+		mb.WithHostParser(docker.HostParser),
```
jsoriano: It had no host parser before :thinking: but I guess it should have had it.

metricbeat/module/docker/_meta/config.yml
```diff
@@ -1,13 +1,5 @@
 - module: docker
-  metricsets: ["container", "cpu", "diskio", "healthcheck", "info", "memory", "network"]
   hosts: ["unix:///var/run/docker.sock"]
-  period: 10s
 
   # Replace dots in labels with `_`. Set to false to keep dots
   labels.dedot: true
```
jsoriano: This is the default, so I guess it is not needed in `config.yml`.

metricbeat/module/docker/_meta/config.yml
```diff
@@ -1,13 +1,5 @@
 - module: docker
-  metricsets: ["container", "cpu", "diskio", "healthcheck", "info", "memory", "network"]
   hosts: ["unix:///var/run/docker.sock"]
```
jsoriano: Out of the scope, but I think this should be the default.

ruflin: +1 on that.

metricbeat/module/docker/_meta/config.yml
ruflin: The reason I left this in is because we plan to change the default here in 7.0. So if it is in the config most people use, 7.0 will be less breaking for existing users.

metricbeat/module/docker/image/image.go
ruflin: Hm, good point. Yes it should have one.

ruflin:
@jsoriano See my comments. Let me know if you need changes or if it is ok to go.

metricbeat/module/docker/_meta/config.yml
jsoriano: Oh ok, right then.

jsoriano:
It is ok to go :)

PR #6720, opened 2018-03-30
Fixes autodiscover's appender configuration example

ph:
The current `YAML` representation in the configuration is invalid, this
PR fixes the alignment.

Reported in https://discuss.elastic.co/t/filebeat-autodiscovery-appenders/126009

@exekias @vjsamuel I think this should be correct format from what I see in the code?

### DIFF
--- a/libbeat/docs/shared-autodiscover.asciidoc
+++ b/libbeat/docs/shared-autodiscover.asciidoc
@@ -157,9 +157,9 @@ metricbeat.autodiscover:
         ...
   appenders:
     - type: config
-      - condition.equals:
-          kubernetes.labels.app: "prometheus"
-        config:
-          fields:
-            type: monitoring
+      condition.equals:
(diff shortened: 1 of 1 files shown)


ph:
Orignal PR with the documentation: https://github.com/elastic/beats/pull/6469

ph:
added @vjsamuel to the description :)

ruflin:
@vjsamuel Why do we have the `CoditionConfig` in the AppenderConfig struct 

https://github.com/elastic/beats/blob/fbc61e82107563d211ab23747902e0ad9b29d656/libbeat/autodiscover/config.go#L26

an also in the appender config:

https://github.com/elastic/beats/blob/fbc61e82107563d211ab23747902e0ad9b29d656/libbeat/autodiscover/appenders/config/config.go#L20

Do we need both?

PR #6721, opened 2018-03-30
support revision overrides for notice generation

graphaelli:
To support elastic/apm-server#460 - NOTICE there has revisions pointing to beats rather than the original dependencies, since they are transitive and govendor doesn't preserve that information.  This change should have no impact on beats directly.


ph:
This will need to be in 6.x too.

PR #6723, opened 2018-03-30
Use golang.org/x/time for rate limiting

andrewkroh:


auditbeat/module/file_integrity/scanner.go
```diff
@@ -155,13 +154,24 @@ func (s *scanner) throttle(fileSize uint64) {
...
+	delay := reservation.Delay()
+	if delay == 0 {
+		return
+	}
+
+	timer := time.NewTimer(delay)
+	defer timer.Stop()
+	select {
+	case <-s.done:
+	case <-timer.C:
```
ph: Didn't know about that library, TIL.

auditbeat/module/file_integrity/scanner.go
```diff
@@ -55,10 +54,10 @@ func (s *scanner) Start(done <-chan struct{}) (<-chan Event, error) {
 				s.config.ScanRatePerSec,
 				s.config.MaxFileSize)
 
-		s.tokenBucket = ratelimit.NewBucketWithRate(
-			float64(s.config.ScanRateBytesPerSec)/2., // Fill Rate
-			int64(s.config.MaxFileSizeBytes))         // Max Capacity
-		s.tokenBucket.TakeAvailable(math.MaxInt64)
+		s.tokenBucket = rate.NewLimiter(
```
ph: Not sure we were using half the `ScanRateBytesPerSec` before? Everything else seems OK for the migration.

andrewkroh: That was a fudge factor. 🤦‍♂️

ph:
LGTM failures is fixed in a newer release of docker-py

ph:
We will need to backport to 6.2

PR #6725, opened 2018-03-31
Use `--mount` over `-v` syntax

dgroomes:
Reference https://docs.docker.com/storage/bind-mounts/

The `--mount` syntax is encouraged over the legacy `-v` syntax.

Also, eliminate the world "volume", because *a bind mount is not a volume*. A bind mount is a mount. A volume mount is a mount. A tmpfs mount is a mount (see https://docs.docker.com/storage/#choose-the-right-type-of-mount). These are the three types of mounts. Using the `--mount` syntax over the `-v` helps avoid this confusion.

Also, I changed the example to `$(pwd)` like they have it at https://docs.docker.com/storage/bind-mounts/#choosing-the--v-or---mount-flag because I was having trouble with `$(~)`, it didn't work for some reason (using bash)... Bind mounts require a fully qualified path (which is actually how the original example worked too because the `~` expands in the shell but in my changes the `~` would not expand because it is not the first character in the argument (not really sure how it works, but my example works))

### DIFF
--- a/metricbeat/docs/running-on-docker.asciidoc
+++ b/metricbeat/docs/running-on-docker.asciidoc
@@ -15,9 +15,9 @@ host machine from within the container.
 ["source","sh",subs="attributes"]
 ----
 docker run \
-  --volume=/proc:/hostfs/proc:ro \ <1>
-  --volume=/sys/fs/cgroup:/hostfs/sys/fs/cgroup:ro \ <2>
-  --volume=/:/hostfs:ro \ <3>
+  --mount type=bind,source=/proc,target=/hostfs/proc,readonly \ <1>
+  --mount type=bind,source=/sys/fs/cgroup,target=/hostfs/sys/fs/cgroup,readonly \ <2>
+  --mount type=bind,source=/,target=/hostfs,readonly \ <3>
   --net=host <4>
   {dockerimage} -system.hostfs=/hostfs
 ----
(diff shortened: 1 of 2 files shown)


andrewkroh:
+1 to this change. The same issue affects `./metricbeat/docs/running-on-docker.asciidoc` too. Would you mind fixing that file too please?

dgroomes:
Sounds good, I'll check that out

dedemorton:
@dgroomes I'd like to get this issue closed out, if possible. Can you update running-on-docker.asciidoc as requested, or would you like us to take over ownership of this change? Thanks for your contribution!

dgroomes:
Thanks @dedemorton, I just pushed the requested change to `running-on-docker.asciidoc` (squashed it into my first commit). It looks right, but not able to try it.

metricbeat/docs/running-on-docker.asciidoc
```diff
@@ -15,9 +15,9 @@ host machine from within the container.
 ["source","sh",subs="attributes"]
 ----
 docker run \
-  --volume=/proc:/hostfs/proc:ro \ <1>
-  --volume=/sys/fs/cgroup:/hostfs/sys/fs/cgroup:ro \ <2>
-  --volume=/:/hostfs:ro \ <3>
+  --mount type=bind,source=/proc,target=/hostfs/proc,readonly \ <1>
+  --mount type=bind,source=/sys/fs/cgroup,target=/hostfs/sys/fs/cgroup,readonly \ <2>
+  --mount type=bind,source=/,target=/hostfs,readonly \ <3>
```
jsoriano: These bind mounts are equivalent to the previous "volumes", and they work, but not sure if they make sense, the last one includes the other two.
In any case this is not introduced in this PR, we can continue with this in a follow-up PR.

dedemorton:
I'm marking this as needs_backport to remind myself to forward port to master.

PR #6727, opened 2018-04-02
Allow autodiscover to monitor unexposed ports

vjsamuel:
The metrics hint builder needs a slight change to allow monitoring of endpoints like `${data.host}:9090`. Currently it blocks such cases. For host network containers, there wouldnt be any port on the configuration as well. We would just spin up metricbeat modules, if port is 0 and the host hint has `data.host` on it.


ruflin:
jenkins, test it

ruflin:
jenkins, test it

vjsamuel:
@ruflin added the documentation. this really only affects Kubernetes autodiscover and not docker.

ruflin:
jenkins, test it

PR #6728, opened 2018-04-02
Add cluster config parameter to kubernetes metagenerator

vjsamuel:
This PR allows `autodiscover.kubernetes` and `processors.add_kubernetes_metadata` to add a `cluster` parameter to the configuration. This config would be used by `kubernetes.MetadataGenerator` to add `kubernetes.cluster.name` as a field when using the above said two. The default is set to `default`. Allowing Beats to autoname clusters would be difficult as each node runs Beats and there would need to be communication between all the Beats. We can start with a configured default and move to a more automated way if and when Kubernetes exposes cluster names via API.

### DIFF
--- a/libbeat/common/kubernetes/metadata.go
+++ b/libbeat/common/kubernetes/metadata.go
@@ -16,14 +16,16 @@ type MetaGenerator interface {
 
 type metaGenerator struct {
 	annotations   []string
+	cluster       string
 	labels        []string
 	labelsExclude []string
 }
 
 // NewMetaGenerator initializes and returns a new kubernetes metadata generator
-func NewMetaGenerator(annotations, labels, labelsExclude []string) MetaGenerator {
+func NewMetaGenerator(cluster string, annotations, labels, labelsExclude []string) MetaGenerator {
 	return &metaGenerator{
 		annotations:   annotations,
+		cluster:       cluster,
 		labels:        labels,
 		labelsExclude: labelsExclude,
 	}
@@ -47,6 +49,9 @@ func (g *metaGenerator) PodMetadata(pod *Pod) common.MapStr {
 
 	annotationsMap := generateMapSubset(pod.Metadata.Annotations, g.annotations)
 	meta := common.MapStr{
+		"cluster": common.MapStr{
+			"name": g.cluster,
+		},
 		"pod": common.MapStr{
 			"name": pod.Metadata.Name,
 		},
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -206,6 +206,7 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
 - Add appender support to autodiscover {pull}6469[6469]
 - Add add_host_metadata processor {pull}5968[5968]
 - Retry configuration to load dashboards if Kibana is not reachable when the beat starts. {pull}6560[6560]
+- Add cluster config parameter to kubernetes metagenerator {pull}6728[6728]
 
 *Auditbeat*
 
--- a/auditbeat/docs/fields.asciidoc
+++ b/auditbeat/docs/fields.asciidoc
@@ -2593,6 +2593,14 @@ Kubernetes metadata added by the kubernetes processor
 
 
 [float]
+=== `kubernetes.cluster.name`
+
+type: keyword
+
+Kubernetes cluster name
+
+
+[float]
 === `kubernetes.pod.name`
 
 type: keyword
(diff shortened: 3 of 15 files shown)


ruflin:
LGTM. I like the idea of making it possible to add a cluster name. I wonder if we should really use `default` as default or just not have the field if it is not set?

Could you add docs for the config options to the processor and autodiscovery docs?

libbeat/common/kubernetes/metadata.go
```diff
@@ -47,6 +49,9 @@ func (g *metaGenerator) PodMetadata(pod *Pod) common.MapStr {
 
 	annotationsMap := generateMapSubset(pod.Metadata.Annotations, g.annotations)
 	meta := common.MapStr{
+		"cluster": common.MapStr{
+			"name": g.cluster,
+		},
```
jsoriano: Could we have it as a plain string like with `namespace`? I wonder why we have some names as strings and some others as `MapStr`.

ruflin: You mean `cluster.name: g.cluster` instead?  In general we always try to convert everything to already the nested object as otherwise it has to happen later or on the ES side.

jsoriano:
Some thoughts about this.

Having something like the cluster name when having multiple clusters is handy (even needed), but with this solution we'd be adding to kubernetes metagenerator something that is not specific from kubernetes.

This is something that could be equally handy in other distributed systems that also don't have something like a cluster name, but I don't think we are going to add fields like `mysql.cluster.name`, `etcd.cluster.name` or `ceph.cluster.name` in their modules.

Would it be an option to let the user add these fields in a generic way? (is it possible to add fields with processors?)
This would also allow for example to let the user add the same field to a kubernetes cluster and its etcd cluster in case they want to visualize all logs together for example.

libbeat/common/kubernetes/metadata.go
jsoriano: I mean not nesting :) `cluster: g.cluster` or `cluster_name: g.cluster`.

ruflin:
@jsoriano So far my assumption was that `etc.cluster.name` will be different from `kubernetes.cluster.name` and these will not be used for correlation across services why I was thinking we should have it in each service. The way In understand your comment correlation on cluster names from different services could be valuable?

jsoriano:
@ruflin this was only an example. My point is that kubernetes clusters don't have a name, as well as other clustered services also don't have any cluster-level identifiers. If we add some static configuration for Kubernetes then we could be adding static configurations to any other module or processor that potentially works with clusters, I don't think we want to do that.

I wonder if for these cases it wouldn't be better to check if we can provide a generic solution like adding custom fields or metadata from configuration, maybe using a processor. This would cover this case, and any other grouping of hosts.

For example, if I were a user with multiple mysql clusters, how could I filter events per cluster?

vjsamuel:
@jsoriano if you look at GKE, there are lot of parameters that define a cluster and cluster name is just one of them. other information can be what is the region it is deployed against. so cluster being a `MapStr` allows placing cluster characteristics into the metadata.

libbeat/common/kubernetes/metadata.go
vjsamuel: @jsoriano if you look at GKE, there are lot of parameters that define a cluster and cluster name is just one of them. other information can be what is the region it is deployed against. so cluster being a MapStr allows placing cluster characteristics into the metadata.

jsoriano: Ok then, agree, it is more flexible this way for the future.

jsoriano:
When I talk about adding custom fields I refer to something like [`fields`](https://www.elastic.co/guide/en/beats/metricbeat/current/configuration-general-options.html#libbeat-configuration-fields), I didn't know we already had something for this :)

libbeat/common/kubernetes/metadata.go
ruflin: what vj said :-)

ruflin:
I think technically I agree that we don't necessarly need an additional an additional config option as the possibilities are already there. But I see advantages from a usability perspective. Most users wont read the documentation and if they do, they will only read the part they are interested in, so in the above case they look at the config or docs for the k8s meta data processor or module. And they will directly see an option to specify a cluster name without having to figuring out that there are `fields` they can set. The other advantage of having a predefined field is that we can preset the type of it in the `fields.yml`.

For me it comes down to the question if it is beneficial for us and the user if he can set a k8s cluster name or in most cases it's not something that is needed.

exekias:
tbh, I would relay on existing `fields` setting to include this info, in the end, this is a custom field to identify the cluster, but not something we can retrieve from kubernetes itself.

If you are monitoring several clusters with the same Elasticsearch, probably you are an advanced user anyway, so maybe you are even using the `fields` setting for more things.

PR #6735, opened 2018-04-03
New metricset named pending_tasks for elasticsearch module

jclegras:

With this metricset I can keep an eye on these following metrics :

- Number of pending tasks ;
- Number of urgent pending tasks ;
- Number of high-priority pending tasks.


metricbeat/module/elasticsearch/pending_tasks/_meta/data.json
```diff
@@ -0,0 +1,22 @@
+{
+    "@timestamp": "2018-04-02T13:53:10.360Z",
+    "beat": {
+        "name": "host.example.com",
+        "hostname": "host.example.com"
+    },
+    "elasticsearch": {
+      "pending_tasks": {
```
jsoriano: Should this be an array?

metricbeat/module/elasticsearch/pending_tasks/data.go
```diff
@@ -0,0 +1,44 @@
...
+	var events []common.MapStr
+	errors := s.NewErrors()
+
+	for _, task := range tasksStruct.Tasks {
+		event, errs := eventMapping(task)
+		errors.AddErrors(errs)
+		events = append(events, event)
+	}
+
+	return events, errors
```
jsoriano: Maybe sending all pending tasks is too much for heavy loaded clusters, it could be useful also to a summarized event with information like number of events or max time in queue, wdyt?

jclegras: I agree with you.

We could just send an event with following metrics : 
- total pending tasks
- total pending tasks with priority urgent
- total pending tasks with priority high
- max time in queue

It's a good indicator.

Nonetheless we lose the source field. I'll have to request the cluster what's happening.

ruflin:
@jclegras Thanks for opening this PR. I'm currently reworking the Elasticsearch module in Metricbeat. I would like to hold back this PR a bit until the refactoring happened. I suggest we keep this PR open and I ping you as soon as we can move forward here.

ruflin:
@jclegras The time has come to move forward on this PR. Will leave some comments in the PR for further discussions.

metricbeat/docs/fields.asciidoc
```diff
@@ -3020,6 +3020,53 @@ format: bytes
...
 
+[float]
+== pending_tasks fields
+
+pending_tasks
+
+
+
+[float]
+=== `elasticsearch.pending_tasks.max_time_in_queue_millis`
```
ruflin: As these are the pending tasks for the cluster I'm wondering if we should put it under `elasticsearch.cluster.pending_tasks`. This does not necessaryl mean we have to rename the metricset. The namespace is configurable for each metricset.

metricbeat/module/elasticsearch/pending_tasks/_meta/data.json
ruflin: There will be a list of events sent to ES but each single event I think is correct they way it is here.

metricbeat/module/elasticsearch/pending_tasks/_meta/fields.yml
```diff
@@ -0,0 +1,25 @@
+- name: pending_tasks
+  type: group
+  description: >
+    pending_tasks
+  fields:
+    - name: max_time_in_queue_millis
+      type: long
+      description: >
+        Max time in queue (ms)
+    - name: pending_tasks_priority_high
```
ruflin: Let's not repeat pending_tasks in the names.

metricbeat/module/elasticsearch/pending_tasks/_meta/fields.yml
```diff
@@ -0,0 +1,25 @@
...
+      type: long
+      description: >
+        Number of pending tasks in priority high
+    - name: pending_tasks_priority_urgent
```
ruflin: In general check naming conventions here: https://www.elastic.co/guide/en/beats/devguide/current/event-conventions.html

metricbeat/module/elasticsearch/pending_tasks/_meta/test/empty.json
```diff
@@ -0,0 +1,3 @@
+{
```
ruflin: Love the extensive testing set you added. I assume these files all come ES 6.x? One thing I did with other datasets is to add the version to the file name for later to indicated which version it came from. This also makes it easy to test across major versions.

metricbeat/module/elasticsearch/pending_tasks/data.go
```diff
@@ -0,0 +1,75 @@
...
+		Tasks []task `json:"tasks"`
+	}{}
+
+	if err := json.Unmarshal(content, &tasksStruct); err != nil {
+		return nil, err
+	}
+
+	maxTimeInQueueMillis := 0.
+	nbTasksByPriority := make(map[string]int)
+	for _, task := range tasksStruct.Tasks {
```
ruflin: It seems what you do here is comibining two metricsets:

* Reporting every pending task
* Reporting a summary. I suggest we make two different metricsets out of it, for example similar to what we did in `system.filesystem.summary`

metricbeat/module/elasticsearch/pending_tasks/data.go
ruflin: See comment above, i think we should have two different metricsets if that works.

metricbeat/module/elasticsearch/pending_tasks/pending_tasks.go
```diff
@@ -0,0 +1,73 @@
...
+// The New method will be called after the setup of the module and before starting to fetch data
+func init() {
+	mb.Registry.AddMetricSet("elasticsearch", "pending_tasks", New, hostParser)
+}
+
+var (
+	hostParser = parse.URLHostParserBuilder{
+		DefaultScheme: "http",
+		PathConfigKey: "path",
+		DefaultPath:   "_cluster/pending_tasks",
```
ruflin: The tricky part with cluster stats compared to node stats is that they should not be fetched from every node but master only. I assume for this the `IsMaster` function I?m creating here will become useful: https://github.com/elastic/beats/pull/6807/files#diff-89821f8160e055daac8a00b1f0cf846aR38

jclegras:
Hi, thanks for the feedback.
I will proceed to the modifications.

jclegras:
@ruflin If you have some time, that's ok for me. Thanks.

metricbeat/module/elasticsearch/pending_tasks/_meta/fields.yml
```diff
@@ -0,0 +1,25 @@
+- name: cluster.pending_tasks
```
ruflin: I would prefer to call this `pending_task`instead of `pending_tasks`. Reason is that the document ending up in Elasticsearch is a single task and not multiple.

metricbeat/module/elasticsearch/pending_tasks/_meta/fields.yml
```diff
@@ -0,0 +1,25 @@
...
+  type: group
+  description: >
+    `cluster.pending_tasks` contains pending tasks description.
+  fields:
+    - name: time_in_queue
```
ruflin: What is the unit of this one? In general I would prefer that we only have 1 field, best the one with more accuracy and set the type as postfix.

metricbeat/module/elasticsearch/pending_tasks/pending_tasks.go
```diff
@@ -0,0 +1,79 @@
...
+func (m *MetricSet) Fetch() ([]common.MapStr, error) {
+	isMaster, err := elasticsearch.IsMaster(m.http, m.HostData().SanitizedURI)
+	if err != nil {
+		return nil, err
+	}
+
+	// Not master, no event sent
+	if !isMaster {
+		logp.Debug("elasticsearch", "Trying to fetch pending tasks from a none master node.")
+		return nil, errors.New("None master node")
```
ruflin: I suggest not to return an error here as otherwise this will trigger an event each time.

metricbeat/module/elasticsearch/pending_tasks_summary/_meta/fields.yml
```diff
@@ -0,0 +1,21 @@
+- name: cluster.pending_tasks_summary
```
ruflin: Could we split this second metricset in a separate PR?

jclegras:
Hi Ruflin, i will proceed to the modifications and create two PRs.

metricbeat/module/elasticsearch/pending_tasks/_meta/fields.yml
ruflin: What are the units of the two fields?

ruflin: Thanks for the link. Had a look at the document and I think we should drop `time_in_queue` is this mainly for human consumption and instead use `time_in_queue_millis` and name it `time_in_queue.ms`. Like this we should have both in one field.

metricbeat/module/elasticsearch/pending_tasks_summary/_meta/fields.yml
jclegras: Hi, I split this PR into two PRs :

- https://github.com/elastic/beats/pull/7144
- https://github.com/elastic/beats/pull/7145

ruflin:
Going to close this PR as https://github.com/elastic/beats/pull/7144 is merged. For https://github.com/elastic/beats/pull/7145 lets continue the discussion there.

Thanks for splitting the PR up so one metricset does not block the other.

PR #6743, opened 2018-04-03
Set default metricsets for rabbitmq module

jsoriano:
See #6668 

Warnings reporting modules as experimental have been changed to beta, so all metricsets are set as default.

### DIFF
--- a/metricbeat/docs/modules/rabbitmq.asciidoc
+++ b/metricbeat/docs/modules/rabbitmq.asciidoc
@@ -9,6 +9,7 @@ beta[]
 
 The RabbitMQ module uses http://www.rabbitmq.com/management.html[HTTP API] created by the management plugin to collect metrics.
 
+The default metricsets are `connection`, `node` and `queue`.
 
 
 [float]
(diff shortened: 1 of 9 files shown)


metricbeat/module/rabbitmq/_meta/config.reference.yml
```diff
@@ -0,0 +1,8 @@
+- module: rabbitmq
+  metricsets: ["node", "queue", "connection"]
+  enabled: true
+  period: 10s
+  hosts: ["localhost:15672"]
+
+  username: guest
```
ruflin: Do we have a  default username and password? If not we should comment these two out.

metricbeat/docs/modules/rabbitmq.asciidoc
```diff
@@ -21,12 +22,7 @@ in <<configuration-metricbeat>>. Here is an example configuration:
 ----
 metricbeat.modules:
 - module: rabbitmq
-  metricsets: ["node", "queue", "connection"]
```
ruflin: Should we just make the other 2 metricsets beta?

jsoriano: Done.

metricbeat/module/rabbitmq/_meta/config.reference.yml
jsoriano: Commented out, they are indeed not set by default.

PR #6746, opened 2018-04-03
[Kibana Module] Add additional metric fields 

tsullivan:
Adds a new Kibana module for `stats`. This is based on the existing `status` module, which will be deprecated in Kibana once `stats` becomes established.

Depends on https://github.com/elastic/kibana/pull/17773


tsullivan:
This is a work-in-progress change as it'll need some extra work after this first commit to implement the schema I'm going for:

- `avg_in_millis` is sometimes an integer, and sometimes a `null` value. If there were no requests in the measurement window, then the value is `null`
- `requests.status_codes` is a dictionary with keys for the status code. Only the status codes that have been measured in the window are present as keys.
   - Example:
      ```
      "requests": {
         "total": 3,
         "disconnects": 0,
         "status_codes": {
            "200": 3
         }
      }
      ```

These need to be handled in the schema to avoid error log messages. Advice is welcome! :)

ruflin:
To make certain fields optional, you can pass an optional flag. See here for an example: https://github.com/elastic/beats/blob/master/metricbeat/module/apache/status/data.go#L22

Also metricbeat tests will required you to document the new fields here (https://github.com/elastic/beats/blob/master/metricbeat/module/kibana/status/_meta/fields.yml) and then run `make update`.

tsullivan:
> To make certain fields optional, you can pass an optional flag

I'm going to ensure that all the fields are predictable by the API, so I won't really need to do this.

> Also metricbeat tests will required you to document the new fields here

Thanks for the tip. I made some changes and ran `make update`

metricbeat/module/kibana/stats/_meta/data.json
```diff
@@ -0,0 +1,69 @@
+{
+    "@timestamp": "2018-04-19T23:17:10.298Z",
+    "beat": {
+        "hostname": "host.example.com",
+        "name": "host.example.com"
+    },
+    "kibana": {
+        "stats": {
+            "cluster_uuid": "G279hqGeSDqjj_OgBq6wUw",
```
ruflin: If this is the elasticsearch cluster uuid, we should put it under `elasticsearch.cluster.id` as we do in other events. This makes it possible to correlate all events from one cluster. I can take care of that in a follow up PR.

metricbeat/module/kibana/stats/stats.go
```diff
@@ -0,0 +1,58 @@
...
+
+// MetricSet type defines all fields of the MetricSet
+type MetricSet struct {
+	mb.BaseMetricSet
+	http *helper.HTTP
+}
+
+// New create a new instance of the MetricSet
+func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
+	cfgwarn.Beta("The kibana metricset is beta")
```
ruflin: We have an interesting case here we haven't had in Beats so far, that we release a metricset for an endpoint which is not available yet.

Is the stats endpoint going into 6.4?

metricbeat/module/kibana/stats/data.go
```diff
@@ -0,0 +1,77 @@
...
+		"uuid":         c.Str("uuid"),
+		"version": c.Dict("version", s.Schema{
+			"number": c.Str("number"),
+		}),
+		"status": c.Dict("status", s.Schema{
+			"overall": c.Dict("overall", s.Schema{
+				"state": c.Str("state"),
+			}),
+		}),
+		"uptime_in_millis": c.Int("uptime_in_millis"),
```
ruflin: We should follow our metricbeat conventions here for the units: https://www.elastic.co/guide/en/beats/devguide/current/event-conventions.html

metricbeat/module/kibana/stats/_meta/fields.yml
```diff
@@ -0,0 +1,156 @@
...
+              description: >
+                Average system load over the last 15 minutes.
+        - name: mem
+          type: group
+          description: >
+            Memory usage metrics of the operating system.
+          fields:
+            - name: free_in_bytes
+              type: long
+              description: >
```
ruflin: For the bytes you can add `format: bytes` so the Kibana index pattern will automatically contain the unit.

ruflin:
@tsullivan There seems to be a panic in the tests at the moment, probably because `metrics` or something that you access doesn't exist in the event. 

I suggest we get this PR in as soon as the stats endpoint is merged in Kibana and then can do follow up PR's to clean up the naming and add integration tests with the actual endpoints.

metricbeat/module/kibana/stats/stats.go
tsullivan: Yes, it's planned for 6.4: https://github.com/elastic/kibana/pull/17773

tsullivan:
> then can do follow up PR's to clean up the naming and add integration tests with the actual endpoints.

Sounds good to me. I will remove the added integration tests for now. 

The feedback around naming looks like things I should work on in https://github.com/elastic/kibana/pull/17773 and then update here to reflect the change before anything gets merged. What do you think?

metricbeat/module/kibana/stats/_meta/data.json
tsullivan: It would affect the Kibana PR as well as this one, so wouldn't it make sense to do this before merging, to reduce the follow-up needed?

The Kibana PR still depends on another one that no one has had time to review yet, so we still have some time to adjust the data structure.

metricbeat/module/kibana/stats/data.go
tsullivan: Just for my clarity, you're suggesting that the change would just be a naming switch in this file, correct?

```
"uptime_in_millis": c.Int("uptime_ms"),
```

?

metricbeat/module/kibana/stats/_meta/data.json
tsullivan: Correct me if I'm wrong about this affecting the Kibana PR. Maybe as an alternative the solution is to map incoming `cluster_uuid` value to `elasticsearch.cluster.id` in `metricbeat/module/kibana/stats/data.go`? If yes, then I'd agree to taking care of it in a followup PR.

metricbeat/module/kibana/stats/data.go
tsullivan: a65066bb2

tsullivan: I guess this idea wouldn't work:
```
2018-04-23T17:44:15.601-0700    ERROR   schema/schema.go:41     Error on field 'uptime_in_millis': Missing field: uptime_in_millis, Error: Key uptime_ms not found
2018-04-23T17:44:15.601-0700    ERROR   schema/schema.go:41     Error on field 'avg_in_millis': Missing field: avg_in_millis, Error: key avg_ms not found
2018-04-23T17:44:15.601-0700    ERROR   schema/schema.go:41     Error on field 'max_in_millis': Missing field: max_in_millis, Error: Key max_ms not found
```

@ruflin is there a way we can pivot on the naming within the beat module? It would help us avoid some messy work in Kibana around not introducing breaking API changes.

ruflin:
@tsullivan For the naming: We could follow the same approach here as we do with the ES endpoints to have an XPack version and a MB version data structure for the next 6.x release. Like this the naming in Kibana is consistent and we can still follow the MB naming scheme to be in line with all other endpoints. 

I'm also ok to get it in with the Kibana naming so I can add it as being X-Pack in a follow up PR and add the MB naming part.

tsullivan:
https://github.com/elastic/kibana/pull/17773 is merged, making this PR testable against the master branch of Kibana!

ruflin:
@tsullivan That is great. We need to modify the testing somehow as so far all metricbeat modules are tested against stable releases. But @jsoriano introduced for other modules to test it against multiple versions which we can also do for Kibana. Main difference is that we will have to build our own image here (or reuse the one from testing/environments). CI will probably not very happy as Kibana takes forever to start and 2 instances makes it worse, but it should work. @tsullivan Let me see how we can make that happen in the repo and then rebase this PR on top of it.

tsullivan:
> @tsullivan Let me see how we can make that happen in the repo and then rebase this PR on top of it.

Sounds like a plan. Let me know when this branch should be rebased.

metricbeat/module/kibana/testing.go
```diff
@@ -0,0 +1,32 @@
+package kibana
```
jsoriano: Move common testing code to `kibana/mtest` as discussed in https://github.com/elastic/beats/pull/7106#discussion_r188708832?

metricbeat/module/kibana/testing.go
```diff
@@ -0,0 +1,32 @@
...
+	}
+	return port
+}
+
+// GetConfig returns config for kibana module
+func GetConfig(metricset string) map[string]interface{} {
+	return map[string]interface{}{
+		"module":     "kibana",
+		"metricsets": []string{metricset},
+		"hosts":      []string{GetEnvHost() + ":" + GetEnvPort()},
```
jsoriano: `net.JoinHostPort()` :)

ruflin: changed. We have this all over the place. Probably time to extract this method more generic.

metricbeat/module/kibana/testing.go
ruflin: done

PR #6750, opened 2018-04-03
Fix minor panic in http parser

adriansr:
There was a bounds check error in parsing HTTP responses. A malformed response line in the form "HTTP/1.1\r\n" would cause a panic when parsed.

Related to #6409


ruflin:
Can you add a changelog entry?

PR #6752, opened 2018-04-04
Rename host.hostname to host.name in processor

ruflin:
host.name is nicer then host.hostname. This is not a breaking change as the processor was not release yet.


ruflin:
CI failures do not seem to be related.

PR #6753, opened 2018-04-04
Fix notice script string conversion

ruflin:
When running the notice script I got the following error:

```
TypeError: replace() argument 1 must be str, not bytes
```

This fixes it by decoding the bytes to utf-8 and a string.

### DIFF
--- a/dev-tools/generate_notice.py
+++ b/dev-tools/generate_notice.py
@@ -216,7 +216,7 @@ def detect_license_summary(content):
     # replace all white spaces with a single space
     content = re.sub(r"\s+", ' ', content)
     # replace smart quotes with less intelligent ones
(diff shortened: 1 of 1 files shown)


ruflin:
I'm kind of surprised our CI did not fail with the current implementation.

@graphaelli I wonder why it worked on your end. Which python version are you using?

ruflin:
Interesting, this now fails on CI but works on my local machine. I probably need to investigate my local machine then.

ruflin:
2.7 is the requirement, 3 is good to have. The part I'm confused is that I thought I have and run python 2.7 locally but it fails. Seems there is a bit of mess up on my system. Closing this PR.

@graphaelli If you have time to make it working for 2.7 and 3, more then happy to get a PR for it.

PR #6757, opened 2018-04-04
Set server metricset in graphite as default

ruflin:


metricbeat/module/graphite/_meta/config.reference.yml
```diff
@@ -0,0 +1,10 @@
+- module: graphite
+  metricsets: ["server"]
+  enabled: true
+  #protocol: "udp"
```
jsoriano: Should we add an example of how to configure listening address and port?

ruflin: Good call. I added quite a bit more details and config parameters. Graphite will need more docs later.

PR #6761, opened 2018-04-04
Set default metricsets for postgresql module

jsoriano:
See #6668


ruflin:
Merged as Travis was green and jenkins did not show any intention to start for this PR ❓

PR #6762, opened 2018-04-04
Set status as default metricset for Kibana module

ruflin:


ruflin:
Changelog added to https://github.com/elastic/beats/issues/6668

PR #6765, opened 2018-04-04
Set stats as default metricset for memcached module

ruflin:


metricbeat/module/memcached/stats/stats.go
```diff
@@ -11,9 +11,9 @@ import (
 )
 
 func init() {
-	if err := mb.Registry.AddMetricSet("memcached", "stats", New); err != nil {
-		panic(err)
-	}
+	mb.Registry.MustAddMetricSet("memcached", "stats", New,
+		mb.DefaultMetricSet(),
```
jsoriano: No host parser here?

ruflin: memcache was one of the early modules before we had the host parser feature. We should probably add one but would not make it part of this PR.

jsoriano: Ok

PR #6767, opened 2018-04-04
Set node as default metricset for munin

ruflin:
The assumption here is that the port 4949 is the default port that a local service will expose the munin protocol. And for the namespace we manually set `node` as the default.

### DIFF
--- a/metricbeat/metricbeat.reference.yml
+++ b/metricbeat/metricbeat.reference.yml
@@ -383,7 +383,7 @@ metricbeat.modules:
 #-------------------------------- Munin Module -------------------------------
 - module: munin
   metricsets: ["node"]
-  enabled: false
+  enabled: true
   period: 10s
   hosts: ["localhost:4949"]
   node.namespace: node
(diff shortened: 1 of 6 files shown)


metricbeat/metricbeat.reference.yml
```diff
@@ -383,10 +383,10 @@ metricbeat.modules:
 #-------------------------------- Munin Module -------------------------------
 - module: munin
   metricsets: ["node"]
-  enabled: false
+  enabled: true
   period: 10s
   hosts: ["localhost:4949"]
-  node.namespace: node
+  #node.namespace: node
```
jsoriano: `node.namespace` is required

ruflin: Updated

This is one of these fields where I'm not sure if we should set it or not. In general we have the defaults in the reference file but already with the `hosts` part we break this principle. It seems more the reference file is going in the direction of "if you copy this, you have something running" which I kind of like.

PR #6768, opened 2018-04-04
Set `pool` as default metricset for php_fpm module

jsoriano:
See #6668

### DIFF
--- a/metricbeat/module/php_fpm/pool/pool.go
+++ b/metricbeat/module/php_fpm/pool/pool.go
@@ -13,9 +13,10 @@ import (
 
 // init registers the MetricSet with the central registry.
 func init() {
-	if err := mb.Registry.AddMetricSet("php_fpm", "pool", New, HostParser); err != nil {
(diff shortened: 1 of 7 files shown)


metricbeat/module/php_fpm/pool/pool.go
```diff
@@ -13,18 +13,19 @@ import (
 
 // init registers the MetricSet with the central registry.
 func init() {
-	if err := mb.Registry.AddMetricSet("php_fpm", "pool", New, HostParser); err != nil {
-		panic(err)
-	}
+	mb.Registry.MustAddMetricSet("php_fpm", "pool", New,
+		mb.WithHostParser(hostParser),
+		mb.DefaultHost(),
```
ruflin: This should be `mb.DefaultMetricSet()`

jsoriano: Oh, what was I thinking about :grimacing: fixed.

PR #6769, opened 2018-04-04
Set status as default in mysql module

ruflin:


metricbeat/module/mysql/_meta/config.yml
```diff
@@ -1,17 +1,2 @@
 - module: mysql
-  metricsets: ["status"]
-  period: 10s
-
-  # Host DSN should be defined as "user:pass@tcp(127.0.0.1:3306)/"
-  # The username and password can either be set in the DSN or using the username
-  # and password config options. Those specified in the DSN take precedence.
   hosts: ["root:secret@tcp(127.0.0.1:3306)/"]
```
jsoriano: Could we leave this without the user and password?

ruflin: yes, changed.

PR #6770, opened 2018-04-04
Set `stubstatus` as default metricset for nginx module

jsoriano:
See #6668


metricbeat/module/nginx/_meta/config.yml
```diff
@@ -1,6 +1,5 @@
 - module: nginx
   metricsets: ["stubstatus"]
```
ruflin: This line should be removed.

jsoriano: Ouch, done.

PR #6778, opened 2018-04-05
Fix flaky windows diskio tests

ruflin:
Some on windows the `disk io counters: context deadline exceeded` error showed up. It seems this happens after a 3s timeout for the query to get the diskio stats on Windows. I'm not sure by what this pretty long delay is caused but I now modified the tests on Windows to only validate the event when there isn't an error.
me.

### DIFF
--- a/metricbeat/tests/system/test_system.py
+++ b/metricbeat/tests/system/test_system.py
@@ -195,8 +195,9 @@ class Test(metricbeat.BaseTest):
 
         for evt in output:
             self.assert_fields_are_documented(evt)
-            diskio = evt["system"]["diskio"]
-            self.assertItemsEqual(self.de_dot(SYSTEM_DISKIO_FIELDS), diskio.keys())
+            if 'error' not in evt:
+                diskio = evt["system"]["diskio"]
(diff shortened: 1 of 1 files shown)


jsoriano:
jenkins, test it please

jsoriano:
Is this happening only on tests?
I'm ok with making the check conditional if it is failing as it only makes noise, but we should investigate it further if it happens when running metricbeat in production too.

ruflin:
So far I didn't see it reported on discuss which doesn't mean it will not happen in production. The good news is if it happens in production, an error event should be sent out to notify the user.

@andrewkroh Do you perhaps know what can cause a query to take longer then 3s?

andrewkroh:
I didn't look too closely, but if it's failing on the windows disk IO stats the answer is yes. The code is using a WMI query. The WMI subsystem could be under load or the machine is slow.

https://github.com/shirou/gopsutil/blob/cd915bdc31582b0a56405ede7fa2f4ab043f851b/disk/disk_windows.go#L147

ruflin:
As the 3s are hard coded as a context timeout I think it's ok if the request fails and we send an error event in case of a slow machine.

jsoriano:
Let merge it by now as flaky tests are noisy, but we should revisit it at some moment to be able to test this.

PR #6780, opened 2018-04-05
Add CHANGELOG-developer file to cover internal changes

ruflin:
The CHANGELOG.asciidoc contains all user facing changes. During the development of Beats lots of internal changes are made to also public API methods. Some of these are used by community Beats. The CHANGELOG-developer.asciidoc is intended to cover these changes and explain how community beats can be migrated. Only a small subset of all API changes will be documented here.

### DIFF
--- a/CHANGELOG-developer.asciidoc
+++ b/CHANGELOG-developer.asciidoc
@@ -3,7 +3,12 @@
 :issue: https://github.com/elastic/beats/issues/
 :pull: https://github.com/elastic/beats/pull/
 
-This changelog is intended for community Beats developers. It covers the major breaking changes to the interal API's in the official Beats and changes around developing a Beat like generators or `fields.yml`. Only the major changes will be covered in this changelog which are expected to affect community developers. Each breaking change added here should have an explanation on how other Beats should be migrated.
+This changelog is intended for community Beat developers. It covers the major
+breaking changes to the internal APIs in the official Beats and changes related
+to developing a Beat like code generators or `fields.yml`. Only the major
+changes will be covered in this changelog that are expected to affect community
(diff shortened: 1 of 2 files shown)


CHANGELOG-internal.asciidoc
```diff
@@ -0,0 +1,16 @@
+// Use these for links to issue and pulls. Note issues and pulls redirect one to
+// each other on Github, so don't worry too much on using the right prefix.
+:issue: https://github.com/elastic/beats/issues/
+:pull: https://github.com/elastic/beats/pull/
+
+This internal changelog is intended to cover the major breaking changes to the interal API's in the official Beats to make it for community Beats easier to follow the changes. Only the major changes will be covered in this changelog which are not all changes to the API's. Each breaking change added here should have an explanation on how other Beats should be migrated.
```
andrewkroh: I think "developer" (or maybe "api") rather than "internal" would be a better name. When I hear "internal" I'm thinking of an API that only a core developer would care about, not something that would affect a person developing their own beat, processor, output, codec, filebeat input, heartbeat monitor, metricbeat module, packetbeat protocol,  or autodiscovery appender/builder/provider (I did a search to see what parts of beats have pluggable interfaces).

ruflin: I started with api initially but then realised it could be more then just the API that changes. For example if something in the packaging or generation of the beat changes it should be there too.

I like the developer proposal.

ruflin:
I renamed the it `CHANGELOG-developer.md` and also added a note to the dev docs.

Additional change is adjusting lines in docs to 80 chars.

ruflin:
@andrewkroh Please have a look again.

PR #6782, opened 2018-04-05
Socket summary module

### LINKED ISSUE #4474: MetricSet - Add System Socket Summary
MetricSet - Add System Socket Summary

agathver:
Closes #4474 
#6646 

Added new socket summary metrics under `system.socket.summary.*`


agathver:
How to make hound-ci happy?

metricbeat/module/system/socket_summary/socket_summary.go
```diff
@@ -0,0 +1,87 @@
...
+
+	"github.com/shirou/gopsutil/net"
+)
+
+// init registers the MetricSet with the central registry as soon as the program
+// starts. The New function will be called later to instantiate an instance of
+// the MetricSet for each host defined in the module's configuration. After the
+// MetricSet has been created then Fetch will begin to be called periodically.
+func init() {
+	mb.Registry.MustAddMetricSet("system", "socket_summary", New)
```
ruflin: A new feature is that the namespace can now also be set during registration time: https://github.com/elastic/beats/blob/master/metricbeat/module/logstash/node_stats/node_stats.go#L22

agathver:
@ruflin  Made changes as requested

metricbeat/module/system/socket_summary/socket_summary.go
```diff
@@ -0,0 +1,88 @@
...
+
+	return common.MapStr{
+		"tcp": common.MapStr{
+			"connections": tcpConns,
+			"listening":   tcpListening,
+		},
+		"udp": common.MapStr{
+			"connections": udpConns,
+		},
+	}
```
jsoriano: Maybe it'd be interesting to have also the total number of connections, and we could also get different stats for ipv4 and ipv6.

metricbeat/module/system/socket_summary/socket_summary.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// Fetch methods implements the data gathering and data conversion to the right
+// format. It publishes the event which is then forwarded to the output. In case
+// of an error set the Error field of mb.Event or simply call report.Error().
+func (m *MetricSet) Fetch(report mb.ReporterV2) {
+
+	conns, err := net.Connections("inet")
```
jsoriano: If we allow to parametrize this we could use this module with the [possible kinds of connections](https://github.com/shirou/gopsutil/blob/master/net/net_linux.go#L281) we could use the same module to have stats also on unix sockets. This can be left for a future PR in any case.

metricbeat/module/system/socket_summary/socket_summary.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/shirou/gopsutil/net"
+)
+
+// init registers the MetricSet with the central registry as soon as the program
+// starts. The New function will be called later to instantiate an instance of
+// the MetricSet for each host defined in the module's configuration. After the
+// MetricSet has been created then Fetch will begin to be called periodically.
+func init() {
+	mb.Registry.MustAddMetricSet("system", "socket_summary", New,
+		mb.WithNamespace("system.socket.summary"),
```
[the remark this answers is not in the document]
ruflin: In case we would also use `summary` in the socket metricset it would so we shouldn't. We do the same trick in some other metricsets to make it easier for the user to access the metrics. 

Should we create a collision on day by accident, CI should detect multiple definitions of the same field.

metricbeat/module/system/socket_summary/socket_summary.go
ruflin: @jsoriano I like that idea a lot. We don't have to add these necessarly in this PR but we should make sure the data structure is ready to be extended with these.

jsoriano: Agree, this can be left for future enhancements

agathver:
I use this structure now:
````js
"system": {
    "socket": {
      "summary": {
        "tcp": {
          "all": {
            "connections": 36,
            "listening": 13
          }
        },
        "udp": {
          "all": {
            "connections": 14
          }
        },
        "all": {
          "connections": 50,
          "listening": 13
        }
      }
    }
````

Using this structure, I think we can easily extend to any other protocol we want. We can even group by `ipv4` and `ipv6` by adding respective groups under `tcp` and `udp`.

We can add new protocols like `SCTP` or `unix`, by adding more top level keys. Similarly, we can group them by `ipv4` or `ipv6` if we want.

ruflin:
jenkins, test it

agathver:
I think it errored due to one of the new fields returning an unexpected value in the integration tests. I'll take a look at them.

ruflin:
@agathver Could you also rebased on the most recent master. We had some issue on CI yesterday.

agathver:
@jsoriano I too agree that it should be count, I have changed that in my most recent commit.
@ruflin  Done!

jsoriano:
@agathver let's revive this PR :slightly_smiling_face:
Could you update the branch with master and resolve the conflicts? Thanks!

jsoriano:
Skipping the test on Windows as the required gopsutil method is indeed [not implemented](https://github.com/shirou/gopsutil/blob/4c73494c780fd9b46c1ce1953907d436516a3ce0/net/net_windows.go#L89) for Windows.

PR #6787, opened 2018-04-06
Provide access to a global logp.Logger

andrewkroh:
This provides static contexts with access to a `logp.Logger` by calling `logp.L()`.

This also fixes the two instances where `logp.NewLogger` was called from a global context.

### DIFF
--- a/libbeat/logp/logger.go
+++ b/libbeat/logp/logger.go
@@ -12,21 +12,33 @@ type Logger struct {
 	sugar *zap.SugaredLogger
 }
 
-// NewLogger returns a new Logger labeled with the name of the selector. This
-// should never be used from any global contexts (instead create "per instance"
-// loggers).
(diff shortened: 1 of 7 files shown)


libbeat/logp/logger.go
```diff
@@ -12,23 +12,35 @@ type Logger struct {
...
 // to the child don't affect the parent, and vice versa.
 func (l *Logger) With(args ...interface{}) *Logger {
 	return &Logger{l.sugar.With(args...)}
 }
 
+// Named adds a new path segment to the logger's name. Segments are joined by
+// periods.
+func (l *Logger) Named(name string) *Logger {
+	return &Logger{l.sugar.Named(name)}
+}
```
ph: TIL

ruflin:
I rebased and merged as I assumed you wanted to keep the 2 commits.

PR #6793, opened 2018-04-06
Metricbeat: Integration tests for haproxy 1.7 and 1.8

jsoriano:
Test haproxy with versions 1.7 and 1.8 apart of 1.5.

It also adds a method to base compose mixin for tests to obtain hosts for tests using docker inspect.


metricbeat/tests/system/test_haproxy.py
```diff
@@ -87,7 +87,15 @@ def test_stat_http_auth(self):
...
             "username": "admin",
             "password": "admin",
-            "hosts": ["http://%s:%d/stats" % (os.getenv('HAPROXY_HOST', 'localhost'), 14569)],
+            "hosts": ["http://%s:%d/stats" % (self.compose_hosts()[0], 14569)],
             "period": "5s"
         }])
         self._test_stat()
+
+
+class Haproxy_1_6_Test(HaproxyTest):
```
ruflin: Not for now, but what if we could autogenerate this part? Meaning we would separate the test code from the versions against we want to test. This is getting interesting :-)

metricbeat/tests/system/test_haproxy.py
```diff
@@ -32,7 +32,7 @@ def test_info_socket(self):
         self.render_config_template(modules=[{
             "name": "haproxy",
             "metricsets": ["info"],
-            "hosts": ["tcp://%s:%d" % (os.getenv('HAPROXY_HOST', 'localhost'), 14567)],
+            "hosts": ["tcp://%s:%d" % (self.compose_hosts()[0], 14567)],
```
ruflin: What happens if more then 1 service is running? I think we are fine in this case as there will be only 1 haproxy instance but there are tests which require more then 1 service to be running.

metricbeat/docs/modules/haproxy.asciidoc
```diff
@@ -43,7 +43,7 @@ The default metricsets are `info`and `stat`.
 [float]
 === Compatibility
 
-The HAProxy metricsets were tested with HAProxy 1.6 and are expected to work with all 1.6 versions.
+The HAProxy metricsets are tested with HAProxy versions from 1.6, 1.7 to 1.8.
```
ruflin: Should we find a way to define the supported versions in fields.yml so we can generate this?

metricbeat/tests/system/test_haproxy.py
```diff
@@ -6,7 +6,7 @@
 HAPROXY_FIELDS = metricbeat.COMMON_FIELDS + ["haproxy"]
 
 
-class Test(metricbeat.BaseTest):
+class HaproxyTest(metricbeat.BaseTest):
```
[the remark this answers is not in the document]
jsoriano: Yeah, indeed `haproxy` is already in the name of the file, I'll change it when I revisit this :)

metricbeat/tests/system/test_haproxy.py
jsoriano: There are ways to generate classes, I could give a try, but I'd wait to see more cases, for example in apache we had to override some parts of the test because each version provides different metrics, it was not only changing the scenario.

metricbeat/tests/system/test_haproxy.py
jsoriano: This method could return a hash with the service name as key, I'll refactor this. It could also return the hash with a list as values, in case some service has multiple replicas, e.g:
```
{
  "proxy": ["172.0.18.1"],
  "server": ["172.0.18.2", "172.0.18.3"],
}
```

I don't know if we have some case like this.

metricbeat/docs/modules/haproxy.asciidoc
jsoriano: We could have a central place where to define the versions matrix for the tests and use it also to generate doc, yes, this will require further thinking.

metricbeat/tests/system/test_haproxy.py
jsoriano: Although if we have different service names for different versions this is not going to work as I'd want, I have to think more about this...

metricbeat/tests/system/test_haproxy.py
ruflin: +1 on waiting for some more cases to see how we can generalise it.

metricbeat/docs/modules/haproxy.asciidoc
ruflin: I wonder if it would be enough if we start with an array `version: ["1.7", "2.1", "5.2-alpha"] or if this will not be enough, as we should also have the supported OS in there and for some versions only a subset of metricset works? Perhaps we should also allow to overwrite these values per metricset?

Issue #257, opened 2015-09-18
Publish an Event for an Incomplete Request/Response


andrewkroh:
Plugins that expire requests that do not receive a responses should publish an error event. Some plugins silently discard the events.

PR #6794, opened 2018-04-06
http: Report unmatched requests or responses

adriansr:
Until now, unmatched requests or responses are silently discarded by the http plugin.
 
This patch makes two changes to this behavior:
    
- When an HTTP response is received without a matching request, an event with Error status is generated.
    
- When an HTTP stream times out, any pending request or response that has received complete headers is also reported with Error status.


adriansr:
Partially addressing #257, as it only works with HTTP.

packetbeat/protos/http/http.go
```diff
@@ -406,18 +406,33 @@ func (http *httpPlugin) handleHTTP(
...
 		conn.responses.append(m)
-		http.correlate(conn)
+		http.correlate(conn, false)
 	}
 }
 
-func (http *httpPlugin) correlate(conn *httpConnectionData) {
-	// drop responses with missing requests
-	if conn.requests.empty() {
+func (http *httpPlugin) correlate(conn *httpConnectionData, flush bool) {
```
urso: Will the correlate always report all active messages? Depending on where packetbeat is running (passive sniffing), a request might be received after the corresponding resource. This make correlation quite tricky in general.

Is there a chance of flushRequests (at the end of the function) is prematurely flushing events? In which situation will the `flush` flag be set? Always on shutdown?

How about extracting the local functions into methods on the httpPlugin. The correlate can re-use them and we can remove the `flush` flag from correlate. Whenever we would set the flush flag, we would call `flushAll`, which calls `flushResponses` and `flushRequests`.

urso:
LGTM. Let's pull change the local closures to httpPlugin methods and get rid of the `flush` flag on correlate.

adriansr:
Thanks @urso, I've addressed your comment.

Also, I've included a fix for a race condition when expired streams are notified. libbeat's common.Cache uses a separate goroutine for the janitor, and removal notifications from the cache are reported from this goroutine. I've refactored it so that plugins are called for this event from the same goroutine that handles the packet processing. Give me your opinion on it, it's all on the first commit, `TCP plugins can be notified of stream expirations`

packetbeat/protos/http/http.go
```diff
@@ -415,14 +415,31 @@ func (http *httpPlugin) handleHTTP(
 	}
 }
 
+func (http *httpPlugin) flushResponses(conn *httpConnectionData) {
+	for !conn.responses.empty() {
+		debugf("Response from unknown transaction. Reporting error.")
```
andrewkroh: If there's any identifying information available about the flow it would likely be beneficial to include a few details in the log. The same in `flushRequests`.

adriansr: Done

PR #6807, opened 2018-04-09
Provide same data structure as X-Pack for ES node_stats

ruflin:
This PR adds a flag `xpack.enabled` to the Elasticsearch Metricbeat module. If enabled the data is transformed to the format as X-Pack Monitoring needs for the UI. It overwrites the index with `monitoring`.

A config to enable the feature currently looks as following:

```
metricbeat.modules:
- module: elasticsearch
  metricsets: ["node_stats"]
  hosts: ["localhost:9200"]
  period: 1s
  xpack.enabled: true
```

* Data is sent to the index `.monitoring-es-6-mb-{date}`. This is hard coded.
* To enabled the feature `xpack.enabled` has to be set in the module
* If a node is master is checked on each request as it can change over time. The master request is not atomic so this information could be inaccurate. The master detection also only works correctly, if data is fetched from only 1 node.
* The cluser_uuid is fetched on the first request and cached from then based on the assumption that the cluster_id for a node id does not change over time.
* `limit_in_bytes` and `usage_in_bytes` are provided as string by ES and not converted to int as they could overflow in ES.
* Experimental message is logged when used. The feature is not documented yet on purpose.

As the data goes into a separate index it is expected that monitoring already has loaded the template.

### DIFF
--- a/metricbeat/mb/event.go
+++ b/metricbeat/mb/event.go
@@ -18,6 +18,7 @@ type Event struct {
 	ModuleFields    common.MapStr // Fields that will be namespaced under [module].
 	MetricSetFields common.MapStr // Fields that will be namespaced under [module].[metricset].
 
+	Index     string        // Index name prefix. If set overwrites the default prefix.
 	Namespace string        // Fully qualified namespace to use for MetricSetFields.
 	Timestamp time.Time     // Timestamp when the event data was collected.
 	Error     error         // Error that occurred while collecting the event data.
@@ -58,6 +59,11 @@ func (e *Event) BeatEvent(module, metricSet string, modifiers ...EventModifier) 
 		}
 
 		e.MetricSetFields = nil
+	}
+
+	// Set index prefix to overwrite default
+	if e.Index != "" {
+		b.Meta = common.MapStr{"index": e.Index}
 	}
 
 	if e.Error != nil {
--- a/metricbeat/module/elasticsearch/node_stats/data_xpack.go
+++ b/metricbeat/module/elasticsearch/node_stats/data_xpack.go
@@ -200,11 +200,8 @@ func eventsMappingXPack(r mb.ReporterV2, m *MetricSet, content []byte) {
 		sourceNode["uuid"] = nodeID
 
 		nodeData, _ := schemaXpack.Apply(node)
-		nodeStats := common.MapStr{
-			"node_master": isMaster,
-			"node_id":     nodeID,
-		}
-		nodeStats.DeepUpdate(nodeData)
+		nodeData["node_master"] = isMaster
+		nodeData["node_id"] = nodeID
 
 		event.RootFields = common.MapStr{
 			"timestamp":    time.Now(),
@@ -212,7 +209,7 @@ func eventsMappingXPack(r mb.ReporterV2, m *MetricSet, content []byte) {
 			"interval_ms":  m.Module().Config().Period.Nanoseconds() / 1000 / 1000,
 			"type":         "node_stats",
 			"source_node":  sourceNode,
-			"node_stats":   nodeStats,
+			"node_stats":   nodeData,
 		}
 
 		// Hard coded index prefix for monitoring, no detection done for ES version at the moment
--- a/metricbeat/module/elasticsearch/node_stats/data_xpack.go
+++ b/metricbeat/module/elasticsearch/node_stats/data_xpack.go
@@ -1,0 +1,223 @@
+package node_stats
+
+import (
+	"encoding/json"
+
+	"time"
+
+	"github.com/elastic/beats/libbeat/common"
+	s "github.com/elastic/beats/libbeat/common/schema"
+	c "github.com/elastic/beats/libbeat/common/schema/mapstriface"
+	"github.com/elastic/beats/libbeat/logp"
+	"github.com/elastic/beats/metricbeat/mb"
+	"github.com/elastic/beats/metricbeat/module/elasticsearch"
+)
+
+var (
+	sourceNodeXpack = s.Schema{
+		"host":              c.Str("host"),
+		"transport_address": c.Str("transport_address"),
+		"ip":                c.Str("ip"),
+		"name":              c.Str("name"),
+	}
+
+	schemaXpack = s.Schema{
+		"indices": c.Dict("indices", s.Schema{
+			"docs": c.Dict("docs", s.Schema{
+				"count": c.Int("count"),
+			}),
+			"store": c.Dict("store", s.Schema{
+				"size_in_bytes": c.Int("size_in_bytes"),
+			}),
+			"indexing": c.Dict("indexing", s.Schema{
+				"index_total":             c.Int("index_total"),
+				"index_time_in_millis":    c.Int("index_time_in_millis"),
+				"throttle_time_in_millis": c.Int("throttle_time_in_millis"),
+			}),
+			"search": c.Dict("search", s.Schema{
+				"query_total":          c.Int("query_total"),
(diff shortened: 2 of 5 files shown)


metricbeat/module/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,127 @@
...
+		return k, nil
+	}
+	return "", fmt.Errorf("No local node found")
+}
+
+func getMasterName(http *helper.HTTP, uri string) (string, error) {
+
+	// Parses the uri to replace the path
+	u, _ := url.Parse(uri)
+	u.Path = "_cluster/state/master_node"
```
pickypg: We can probably call `GET /_cluster/state/master_node?local=true`  to avoid having to bounce to the elected master to find out what this node thinks. Since we have to asynchronously fetch this, it's a race condition regardless. If it's wrong, it'll fix itself on the next pass.

metricbeat/module/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,127 @@
...
+	json.Unmarshal(content, &clusterStruct)
+
+	return clusterStruct.MasterNode, nil
+}
+
+func getInfo(http *helper.HTTP, uri string) (*infoStruct, error) {
+
+	// Parses the uri to replace the path
+	u, _ := url.Parse(uri)
+	u.Path = ""
```
pickypg: It's a shame that we have to call this to get the `cluster_uuid` after invoking the `GET /_cluster/state/master_node`. Perhaps we can get it added over there so we can do a single request to fetch both details.

metricbeat/module/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,127 @@
...
+	if err != nil {
+		return false, err
+	}
+
+	return master == node, nil
+}
+
+func getNodeName(http *helper.HTTP, uri string) (string, error) {
+
+	// Parses the uri to replace the path
```
pickypg: Lines 64-73 are the same for the next 2 methods except for `u.Path`. Perhaps we can have a helper method that avoids repeating it all?

metricbeat/module/elasticsearch/node_stats/data_xpack.go
```diff
@@ -0,0 +1,225 @@
...
+		Nodes       map[string]map[string]interface{} `json:"nodes"`
+	}{}
+
+	json.Unmarshal(content, &nodesStruct)
+
+	// Normally the nodeStruct should only contain one node. But if _local is removed
+	// from the path and Metricbeat is not installed on the same machine as the node
+	// it will provid the data for multiple nodes. This will mean the detection of the
+	// master node will not be accurate anymore as often in these cases a proxy is in front
+	// of ES and it's not know if the request will be routed to the same node as before.
```
pickypg: In order to support proxies, we _could_ [allow users to specify the node name](https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster.html#cluster-nodes) that they want Beats to monitor and, if unspecified, use `_local` as the name.

Pitfalls:

- Misconfiguration means it's broken.
- Duplicate node names can cause confusion.
- Requires the module's configuration to be reloaded if the node is renamed.

Advantages:

- We'll work, by default, for non-proxies.
- We can work through proxies.

metricbeat/module/elasticsearch/node_stats/data_xpack.go
```diff
@@ -0,0 +1,225 @@
...
+
+		event := mb.Event{}
+		// Build source_node object
+		sourceNode, _ := sourceNodeXpack.Apply(node)
+		sourceNode["uuid"] = nodeId
+
+		nodeData, _ := schemaXpack.Apply(node)
+		nodeStats := common.MapStr{
+			// This value is currently hard coded as it's not available in node_stats endpoint
+			"mlockall":    false,
```
pickypg: We can ignore this value, which was a holdover from when it was in `_nodes/stats`. We do not use it and it should be in an information document [that does not exist] -- not a stats one.

In case you're wondering, you _could_ fetch it via `GET /_nodes/_local/process`.

metricbeat/mb/event.go
```diff
@@ -60,6 +61,12 @@ func (e *Event) BeatEvent(module, metricSet string, modifiers ...EventModifier)
 		e.MetricSetFields = nil
 	}
 
+	// Set index prefix
+	// TODO: does this break if user sets an index himself?
```
ruflin: @urso What will happen here with user defined index patterns? Does this get overwritten?

metricbeat/module/elasticsearch/elasticsearch.go
ruflin: done

metricbeat/module/elasticsearch/elasticsearch.go
ruflin: Nice, wasn't aware of this param. That means we also get an answer in case the node is disconnected I assume.

metricbeat/module/elasticsearch/elasticsearch.go
ruflin: That would be great. Can you take this up with the ES team as you probably know best where this change would have to happen?

metricbeat/module/elasticsearch/node_stats/data_xpack.go
ruflin: That is pretty interesting. I would skip it for the first iteration but I'm pretty sure that is a request that will pop up.

It will probably be a bit tricker for dynamic environments where the user does not have much control around the naming of nodes.

metricbeat/module/elasticsearch/node_stats/data_xpack.go
ruflin: Removed.

ruflin:
Review feedback addressed. We should leave this open until our last 6.x merge from master for 6.3 is done.

metricbeat/module/elasticsearch/node_stats/data_xpack.go
pickypg: In that case they can set its address, which hopefully they know.

metricbeat/module/elasticsearch/elasticsearch.go
pickypg: https://github.com/elastic/elasticsearch/pull/30143

ruflin:
This PR is ready for an other review.

As soon as https://github.com/elastic/elasticsearch/pull/30143 is merged we can do a follow up PR for further simplifying the collection.

ruflin:
One additional note on https://github.com/elastic/elasticsearch/pull/30143 As this will be only for 6.4 forward we will still need the lookup for earlier versions and it means we have to detect which version we are monitoring.

metricbeat/module/elasticsearch/node_stats/data_xpack.go
```diff
@@ -0,0 +1,223 @@
...
+		// Build source_node object
+		sourceNode, _ := sourceNodeXpack.Apply(node)
+		sourceNode["uuid"] = nodeID
+
+		nodeData, _ := schemaXpack.Apply(node)
+		nodeStats := common.MapStr{
+			"node_master": isMaster,
+			"node_id":     nodeID,
+		}
+		nodeStats.DeepUpdate(nodeData)
```
exekias: You could also use:
```
nodeData["node_master"] = isMaster
nodeData["node_id"] = nodeId
```

That would avoid DeepUpdate iterating over all fields

ruflin: Good point, changed.

PR #6809, opened 2018-04-09
Fix leaks in filebeat log harvester

### LINKED ISSUE #6797: [Filebeat] Memory leak associated with failure to setup harvesters
Filebeat has a memory leak which is exposed by repeated failure to setup harvesters (eg. `permission denied`). This ticket is a result of the [original discussion](https://discuss.elastic.co/t/filebeat-memory-leak-associated-with-failure-to-open-harvesters/123100/8) on the community support site.

From what I can tell, this comes down to [channel.SubOutlet](https://github.com/elastic/beats/blob/6.2/filebeat/prospector/log/prospector.go#L577) being created for the new harvester, but not cleaned up properly when `h.Setup()` throws an exception. I think this is because something that happens in `harvester.Run()` is what actually cleans up the outlet. Another possibility is that the new harvester is getting added to a state or registry structure somewhere and forgotten, so garbage collection isn't reaping it.

Here's some output from pprof. I've also attached the pprof file [here](https://github.com/elastic/beats/files/1885425/pprof.filebeat.alloc_objects.alloc_space.inuse_objects.inuse_space.005.pb.gz).

```(pprof) top10
33159.58kB of 33159.58kB total (  100%)
Dropped 308 nodes (cum <= 165.80kB)
Showing top 10 nodes out of 51 (cum >= 512.04kB)
      flat  flat%   sum%        cum   cum%
19975.92kB 60.24% 60.24% 19975.92kB 60.24%  runtime.malg
 7680.66kB 23.16% 83.40%  7680.66kB 23.16%  github.com/elastic/beats/filebeat/channel.SubOutlet
 2048.19kB  6.18% 89.58%  2048.19kB  6.18%  github.com/elastic/beats/filebeat/prospector/log.NewHarvester
 1357.91kB  4.10% 93.68%  1357.91kB  4.10%  runtime.allgadd
 1024.08kB  3.09% 96.76%  1024.08kB  3.09%  runtime.acquireSudog
  544.67kB  1.64% 98.41%   544.67kB  1.64%  github.com/elastic/beats/libbeat/publisher/queue/memqueue.NewBroker
  528.17kB  1.59%   100%   528.17kB  1.59%  regexp.(*bitState).reset
         0     0%   100%   528.17kB  1.59%  github.com/elastic/beats/filebeat/beater.(*Filebeat).Run
         0     0%   100%   512.04kB  1.54%  github.com/elastic/beats/filebeat/channel.CloseOnSignal.func1
         0     0%   100%   512.04kB  1.54%  github.com/elastic/beats/filebeat/channel.SubOutlet.func1
(pprof) list SubOutlet
Total: 32.38MB
ROUTINE ======================== github.com/elastic/beats/filebeat/channel.SubOutlet in /home/jeremy/src/go/src/github.com/elastic/beats/filebeat/channel/util.go
    7.50MB     7.50MB (flat, cum) 23.16% of Total
         .          .     15:// SubOutlet create a sub-outlet, which can be closed individually, without closing the
         .          .     16:// underlying outlet.
         .          .     17:func SubOutlet(out Outleter) Outleter {
         .          .     18:	s := &subOutlet{
         .          .     19:		isOpen: atomic.MakeBool(true),
       1MB        1MB     20:		done:   make(chan struct{}),
       2MB        2MB     21:		ch:     make(chan *util.Data),
    4.50MB     4.50MB     22:		res:    make(chan bool, 1),
         .          .     23:	}
         .          .     24:
         .          .     25:	go func() {
         .          .     26:		for event := range s.ch {
         .          .     27:			s.res <- out.OnEvent(event) 
```

profile001

Config (abridged):

```
filebeat.modules:
  - module: system

# yes, we know fqdn is in the `beats` field
fields_under_root: true
fields:
  source_host: {{ fqdn }}

filebeat.shutdown_timeout: 10s
filebeat.registry_flush: 30s

output.kafka:
  hosts: ["kafka-01..."]

  topic: {{ topic }}

  required_acks: 1
  compression: snappy
  client_id: '{{ client_id }}'
  keep_alive: 10m

  partition.round_robin:
    reachable_only: true

logging.level: info

```
Note, in the above scenario, `/var/log/auth.log` is the file which Filebeat can't access. We can, of course, fix the permissions to alleviate this problem, but that doesn't really fix anything.

- Version: Filebeat 6.2.2
- Operating System: Ubuntu 14.04. Kernel: 4.4.0-111-generic
- Steps to Reproduce: Run Filebeat with the above config. Make sure at least 1 file found by the prospector generates a `permission denied` error when opening the file.
- Notes: This is a slow leak, which makes testing it harder. I think lowering the default value of `scan_frequency` (eg. `1s`) will make it leak memory faster.

Please let me know if there are more details I can provide, and I'm happy to test any potential solutions when we narrow down the problem.

Thanks,
Jeremy

adriansr:
Two independent but related causes of goroutine leaks have been found in filebeat:

- channel.SubOutlet fails to terminate an internal goroutine, which keeps references to the SubOutlet itself and the Outlet being wrapped, preventing them from being freed.

- log.Harvester constructor receives an Outlet which is closed by Harvester.Stop() or internally after Harvester.Run() is invoked. When Harvester.Setup() fails, Stop() is not called, causing a leak. A CloseOnSignal() util is also leaked for the same reasons, which adds another goroutine leak.

Fixes #6797

### DIFF
--- a/filebeat/channel/util.go
+++ b/filebeat/channel/util.go
@@ -22,13 +22,20 @@ func SubOutlet(out Outleter) Outleter {
 		res:    make(chan bool, 1),
 	}
 
-	go func() {
-		for event := range s.ch {
-			s.res <- out.OnEvent(event)
-		}
-	}()
+	go s.drainLoop(out)
 
 	return s
+}
+
+func (o *subOutlet) drainLoop(out Outleter) {
+	for {
+		select {
+		case <-o.done:
+			return
+		case event := <-o.ch:
+			o.res <- out.OnEvent(event)
+		}
+	}
 }
 
 func (o *subOutlet) Close() error {
--- a/filebeat/input/log/harvester.go
+++ b/filebeat/input/log/harvester.go
@@ -144,6 +144,7 @@ func (h *Harvester) ID() uuid.UUID {
 func (h *Harvester) Setup() error {
 	err := h.open()
 	if err != nil {
+		h.stop()
 		return fmt.Errorf("Harvester setup failed. Unexpected file opening error: %s", err)
 	}
 
@@ -152,6 +153,7 @@ func (h *Harvester) Setup() error {
 		if h.source != nil {
 			h.source.Close()
 		}
+		h.stop()
 		return fmt.Errorf("Harvester setup failed. Unexpected encoding line reader error: %s", err)
 	}
 


filebeat/channel/util.go
```diff
@@ -22,15 +22,22 @@ func SubOutlet(out Outleter) Outleter {
...
-		for event := range s.ch {
-			s.res <- out.OnEvent(event)
-		}
-	}()
+	go s.drainLoop(out)
 
 	return s
 }
 
+func (o *subOutlet) drainLoop(out Outleter) {
```
urso: maybe name it workerLoop? This loop is forwarding all received events to `out.OnEvent`.

urso:
The SubOutlet is still some left-over to be removed in the future. As the underlying channel is shared with the prospector, so to guarantee order on state updates, it can not be that easily refactored. Plus shutdown is potentially riddled with race conditions.

Cleaning up prospector/harvester and state handling depends on #6810.

ruflin:
Some more refactoring will be need but we should not block this change by it as it's not a small undertaking.

ph:
> Some more refactoring will be need but we should not block this change by it as it's not a small undertaking.

Yep not for blocking this, it's just related as to keep note of what can be improved in a more long term work.

urso:
@ph @ruflin @asr problem with this change is, I have a hard time to figure out the actual implications of this change. We still pass loads of channels in filebeat and shutdown still potentially riddled with a number of race conditions. This change indirectly affects the interworkings of prospectors, harvesters, registry, the event publishing pipeline, global counters on filebeat shutdown. It took us a while to get shutdown working somewhat stable (somewhat, as in stdin is still not working correctly). That is, by fixing one issue, we might re-introduce some other issues from the past.

Trying to fix whatever issue we have with shutdown/startup will be a never ending story (it still is), without the required refactorings on registry and shutdown waiting logic (waiting is supported by the publisher pipeline since 6.0, but can't be used by filebeat yet).

tl;dr here be dragons

ruflin:
I share the concern from @urso that we are not 100% sure about the side effects of this change. But I think we should find a way forward in small steps instead of one big one which will also introduce other issues again. Best would be if we can have tests for every step. For example the new pipeline in 6.x in Beats is great and a required step we had to do but also has reintroduce some issue for example with the once shutdown in stdin. Same will be true for any small or major changes we will do here, so it is a trade off we have to choose. There will be dragons everywhere ;-)

@adriansr I wonder if we could have a test that failed before having this change but goes green with the change. BTW: As @urso predicted, it broke the shutdown test on windows: https://beats-ci.elastic.co/job/elastic+beats+pull-request+multijob-windows/3895/beat=filebeat,label=windows/testReport/junit/test_shutdown/Test/test_shutdown_wait_timeout/ If we get our tests green, I'm relatively confident in any changes we do as the test suite is rather extensive.

Issue #3384, opened 2017-01-16
[Metricbeat] Excluding bind mounts


abraxxa:
Feature request per https://discuss.elastic.co/t/excluding-bind-mounts/71509

- Version: 5.1.2
- Operating System: Debian 8 Linux 64bit
- Steps to Reproduce: 
install and configure metricbeat to collect filesystem/fsstat data
look at both installed Kibana dashboards, storage space is counted multiple times when bind mounts exist

Suggested solution:
Add a filesystem type field to the mapping, add a filter to the default config or hardcode exclude bind mounts (that would always make sense for fsstat as the total space is wrong with them).

willemdh:
The fsstat module is really useless atm, please exclude mount points and subpaths by default.

jsoriano:
After #6819, each filesystem should appear only once even if it is mounted multiple times, if it is, only the shorter mount path is reported. Bind mounts won't be reported as it is assumed that the source/parent filesystem is also mounted and reported on its own.

Also, file systems of "virtual" filesystem types won't be reported by default (these types are autodiscovered when possible and added to [`filesystem.ignore_types`](https://www.elastic.co/guide/en/beats/metricbeat/current/metricbeat-metricset-system-filesystem.html)).

I hope it helps to make these metricsets more useful by default, closing this by now.

abraxxa:
Sounds great, thanks! In which version will this be included?

ruflin:
@abraxxa 6.3

PR #4823, opened 2017-08-03
Add filesystem.ignore_types to ignore filesystem types

andrewkroh:
Add `filesystem.ignore_types` to the system module for ignoring filesystems
in the `filesystem` and `fsstat` metricsets. The new configuration option
accepts a list of filesystem types.

    metricbeat.modules:
    - module: system
      metricsets: [filesystem, fsstat]
      filesystem.ignore_types: [nfs, smbfs, proc, cgroups]

Closes #4685

PR #6819, opened 2018-04-10
Metricbeat: Remove duplicated filesystems in system module

jsoriano:
Mountpoints whose device name is the same absolute path are considered to be the same and are counted only once. The directory name used is the shorter between them. If the device of a filesystem is a directory, it is considered some kind of bind mount and is ignored. This avoids duplication of filesystems as can happen with bind mounts in Linux (#3384). This mimics the behaviour of `df` command.

There are still filesystems that can be counted twice if union filesystems are used (as aufs or overlay), but this can be controlled with the `filesystem.ignore_types` option. ~~Maybe we should blacklist them by default.~~ Update: For that, if the option is left empty, we fill the list of ignored types with all `nodev` devices in `/proc/filesystems` where this file exists.

### DIFF
--- a/metricbeat/module/system/filesystem/filesystem.go
+++ b/metricbeat/module/system/filesystem/filesystem.go
@@ -3,6 +3,8 @@
 package filesystem
 
 import (
+	"strings"
+
 	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/logp"
 	"github.com/elastic/beats/metricbeat/mb"
@@ -30,6 +32,13 @@ func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
 	var config Config
 	if err := base.Module().UnpackConfig(&config); err != nil {
 		return nil, err
+	}
+
+	if config.IgnoreTypes == nil {
+		config.IgnoreTypes = DefaultIgnoredTypes()
+	}
+	if len(config.IgnoreTypes) > 0 {
+		logp.Info("Ignoring filesystem types: %s", strings.Join(config.IgnoreTypes, ", "))
 	}
 
 	return &MetricSet{
--- a/metricbeat/module/system/filesystem/helper.go
+++ b/metricbeat/module/system/filesystem/helper.go
@@ -3,6 +3,7 @@
 package filesystem
 
 import (
+	"os"
 	"path/filepath"
 	"time"
 
@@ -53,7 +54,14 @@ func filterFileSystemList(fsList []sigar.FileSystem) []sigar.FileSystem {
 			continue
 		}
 
-		// If a block device is mounted multiple times (e.g. with bind mounts),
+		// If the device name is a directory, this is a bind mount or nullfs,
+		// don't count it as it'd be counting again its parent filesystem.
+		devFileInfo, _ := os.Stat(fs.DevName)
+		if devFileInfo != nil && devFileInfo.IsDir() {
+			continue
+		}
+
+		// If a block device is mounted multiple times (e.g. with some bind mounts),
 		// store it only once, and use the shorter mount point path.
 		if seen, found := devices[fs.DevName]; found {
 			if len(fs.DirName) < len(seen.DirName) {
--- a/metricbeat/module/system/filesystem/helper.go
+++ b/metricbeat/module/system/filesystem/helper.go
@@ -3,13 +3,17 @@
 package filesystem
 
 import (
+	"bufio"
 	"os"
+	"path"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"runtime"
 
 	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/metricbeat/module/system"
 	sigar "github.com/elastic/gosigar"
 )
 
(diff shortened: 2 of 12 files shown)


metricbeat/module/system/filesystem/helper.go
```diff
@@ -31,19 +31,45 @@ func GetFileSystemList() ([]sigar.FileSystem, error) {
...
+		// Don't do further checks in special devices
+		if !filepath.IsAbs(fs.DevName) {
 			filtered = append(filtered, fs)
 			continue
 		}
-		debugf("Filtering filesystem with relative mountpoint %+v", fs)
+
+		// If a block device is mounted multiple times (e.g. with bind mounts),
+		// store it only once, and use the shorter mount point path.
+		if seen, found := devices[fs.DevName]; found {
```
jsoriano: This is not completely accurate, there can be different device paths that are really pointing to the same block device, but I don't believe this is common at all. To control this we should check duplicates by device identifier (`Major:Minor`) but this is not supported in gosigar yet. Does it worth to implement this?

jsoriano:
Please don't merge this, I have just seen that gosigar obtains list of filesystem from mtab, what doesn't include the device name for bind mounts.

metricbeat/module/system/filesystem/helper.go
ruflin: I would leave it what you have for now but potentially follow up with a change in gosigar to make it possible if it is a case that can happen (more then once :-)). @andrewkroh WDYT?

metricbeat/module/system/filesystem/helper_test.go
```diff
@@ -47,6 +47,103 @@ func TestFileSystemList(t *testing.T) {
...
+			fss: []sigar.FileSystem{
+				{DirName: "/", DevName: "/dev/sda1"},
+				{DirName: "/", DevName: "/dev/sda1"},
+			},
+			expected: []sigar.FileSystem{
+				{DirName: "/", DevName: "/dev/sda1"},
+			},
+		},
+		{
+			description: "Don't repeat devices, sortest of dir names should be used",
```
ruflin: nit: shortest, also in all the ones below

jsoriano: :D

jsoriano:
On linux we could maybe ignore by default all types marked as `nodev` in `/proc/filesystems`.

ruflin:
@jsoriano I'm ok with excluding `nodev` type by default. Your comment reminded me that as part of this PR we should extend the docs for filesystem and explain in detail what we include and exclude. This question will pop up and it's good to have a place to link to and look it up. This should also make it easier to revisit the decision later again.

jsoriano:
Two more changes added:
* Filesystems whose device name is a directory are not collected, it is assumed to be some kind of bind mount.
* If no `filesystem.ignore_types` is set, it is set to the list of known virtual filesystems in the system, it is by now the list of types marked as `nodev` in systems where `/proc/filesystems` is available.

jsoriano:
Testing it in a linux machine it looks quite better now with the default settings.

metricbeat/module/system/filesystem/filesystem.go
```diff
@@ -32,6 +34,13 @@ func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
 		return nil, err
 	}
 
+	if len(config.IgnoreTypes) == 0 {
```
ruflin: The behaviour I would have expected that if a user sets `filesystem.ignore_types: []` that then no file systems would be ignore. But checking for 0 indicates that it also applies in this case. Could we instead check if the config is set?

If we go with the above suggestion, the reference config file should be adjusted to contain the two values.

jsoriano: Oh, right, checking nil instead of zero lenght. I have added a comment also in reference config.

metricbeat/module/system/filesystem/helper.go
```diff
@@ -31,19 +36,56 @@ func GetFileSystemList() ([]sigar.FileSystem, error) {
...
+		// Don't do further checks in special devices
+		if !filepath.IsAbs(fs.DevName) {
 			filtered = append(filtered, fs)
 			continue
 		}
-		debugf("Filtering filesystem with relative mountpoint %+v", fs)
+
+		// If the device name is a directory, this is a bind mount or nullfs,
+		// don't count it as it'd be counting again its parent filesystem.
+		devFileInfo, err := os.Stat(fs.DevName)
```
andrewkroh: One of the original reasons we added `ignore_types` was to avoid doing a `statfs` on autofs filesystems. Metricbeat was causing autofs mounts to never be unmounted because its continuous metric gathering was making autofs think the mount was being accessed.

Will this cause the same problem?

metricbeat/module/system/filesystem/helper.go
andrewkroh: If it will help make Metricbeat more accurate then I'm +1.

metricbeat/module/system/filesystem/helper.go
jsoriano: Good point.
Here the stat is done against the device name, on #4823 the mentioned issued seemed to appear when doing statfs on the mount point (`DirName`). So I don't think it can cause the same problem.

metricbeat/module/system/filesystem/helper.go
jsoriano: I will take a look to this in gosigar, meanwhile I think we are good enough with the checks based on device name.

ruflin:
@jsoriano I think the failing windows tests are related to this change.

jsoriano:
Removing filtering on Windows

ruflin:
Something went wrong in the helper:

```
darwin/beat/metricbeat/label/macosx/src/github.com/elastic/beats/metricbeat/module/system/filesystem/helper.go:41:3: not enough arguments to return
08:29:51 	have ([]gosigar.FileSystem)
08:29:51 	want ([]gosigar.FileSystem, error)
```

jsoriano:
jenkins, test again please

PR #6820, opened 2018-04-10
Add event hubs output

jseely:
Adds an output to Azure Event Hubs


jseely:
Just realized I had custom properties, will resubmit later

ruflin:
@jseely Related to new outputs best have a look at the thread here: https://github.com/elastic/beats/issues/581

PR #6823, opened 2018-04-11
Add support for Kafka 1.0.0 and 1.1.0 to kafka metricbeat module

filipecosta90:
Add support for Kafka 1.0.0 and 1.1.0 to kafka metricbeat module


ruflin:
@filipecosta90 Thanks for taking this on. I think it's good to add these new versions to the file to start testing against these version. To make it clear for others coming to this issue, having these version in there does not automatically mean we support all of them as we first need to get the testing in place to verify it works.

@urso I wonder if these versions also have to be added somewhere in the output? https://github.com/elastic/beats/blob/master/libbeat/outputs/kafka/kafka.go#L66 Interestingly the two files are already now out of sync. Should we sync them up?

urso:
The kafka output and metricbeat kafka module have some duplicated out-of-sync definitions/code here. We should unify some functionality in a `libbeat/net/kafka` package.

PR #6825, opened 2018-04-11
Allow system tests to be run from any directory

ruflin:
Currently all system tests have to be in `tests/system`. To allow more flexibility the code was modified to search in all directories for `test_*.py` files and run them. To run them from any place it's important that they don't use relative paths. This PR fixes the tests which did not have that yet.

Changes
* Change running system tests to be based on a directory to search for all test_*.py files outside vendor and build. This allows the tests files to be placed anywhere.
* Change relative paths in all beats to depend on self.beat_path. Like this the location form where `nosetests` is triggered does not matter.
* Unify directory tests/files across Beats to contain files. Moved files for packetbeat and dashboard tests accordingly.
* Generalise test for Metricset in Elasticsearch. This could probably be generalised further and used for most of the other tests.
* Fix flaky clean_inactive Filebeat test
* Prefix python fix with @ to not always have it in the logs

This PR is extracted from https://github.com/elastic/beats/pull/6534


auditbeat/tests/system/auditbeat.py
```diff
@@ -1,7 +1,8 @@
 import os
 import sys
 
-sys.path.append('../../../metricbeat/tests/system')
+sys.path.append(os.path.join(os.path.dirname(__file__), '../../../metricbeat/tests/system'))
```
jsoriano: metricbeat in an auditbeat file?

filebeat/tests/system/test_shutdown.py
```diff
@@ -143,15 +143,14 @@ def test_once(self):
     def nasa_logs(self):
 
         # Uncompress the nasa log file.
-        nasa_log = '../files/logs/nasa-50k.log'
+        nasa_log = self.beat_path + '/tests/files/logs/nasa-50k.log'
```
jsoriano: `os.path.join`?

auditbeat/tests/system/auditbeat.py
ruflin: yes, auditbeat depends on metricbeat

filebeat/tests/system/test_shutdown.py
ruflin: changed

PR #6828, opened 2018-04-11
add note about overwriting dashboards

loekvangool:


libbeat/docs/dashboardsconfig.asciidoc
```diff
@@ -41,7 +41,9 @@ You can specify the following options in the `setup.dashboards` section of the
 
 If this option is set to true, {beatname_uc} loads the sample Kibana dashboards
 automatically on startup. If no other options are set, the dashboard are loaded
-from the local `kibana` directory in the home path of the installation.
+from the local `kibana` directory in the home path of the installation. Note that
```
dedemorton: This seems fairly important, so you might want to call this out as a separate note. I wonder if it might make sense to have a `setup.dashboards.overwrite` option like we have with the index template?

I'd suggest the following changes (includes a few edits to remove ambiguity in the existing text): 

```
If this option is set to true, {beatname_uc} loads the sample Kibana dashboards
from the local `kibana` directory in the home path of the {beatname_uc} installation.

NOTE: When dashboard loading is enabled, {beatname_uc} overwrites any existing
dashboards that match the names of the dashboards you are loading. This happens
every time {beatname_uc} starts. 

```

loekvangool: @dedemorton https://github.com/elastic/beats/blob/9e2838d9ecaf4d9c3ea05ba7f97182cdd2d93574/libbeat/dashboards/kibana_loader.go#L89 suggests we indeed force overwriting the dashboards (not a developer on Beats, could be wrong about this).

I opened a Discuss on this: https://discuss.elastic.co/t/should-beats-modules-overwrite-kibana-dashboards/127984

I like the new wording. I patched the PR.

ruflin:
@kvch Could you have a look at this if it is our expected behaviour? Could you also have a look at the overwrite config comment? This could be in line with the pipeline loading config?

libbeat/docs/dashboardsconfig.asciidoc
```diff
@@ -40,8 +40,11 @@ You can specify the following options in the `setup.dashboards` section of the
 ==== `setup.dashboards.enabled`
 
 If this option is set to true, {beatname_uc} loads the sample Kibana dashboards
-automatically on startup. If no other options are set, the dashboard are loaded
```
kvch: Why the sentence which start with "If no other options are set..." is removed? AFAIK it is still possible to override it using `setup.dashboards.directory`.

kvch:
@ruflin Dashboards are indeed overwritten every time someone calls `setup`. This note is valid.
What do you mean by your last question?

libbeat/docs/dashboardsconfig.asciidoc
dedemorton: @kvch I think this was my fault! When I suggested an edit to @loekvangool, I inadvertently deleted that line when I copy/pasted the original text. I didn't notice the line was missing because the sentence read correctly without the deleted line. My edit *should* have said:

```
If this option is set to true, {beatname_uc} loads the sample Kibana dashboards
automatically on startup. If no other options are set, the dashboard are loaded
from the local `kibana` directory in the home path of the {beatname_uc} installation.
```

Sorry about the confusion.

loekvangool:
>Dashboards are indeed overwritten every time someone calls `setup`

It happens when:
- `setup.dashboards.enabled: true` and Beat is restarted
- `{beat} setup --dashboards` is called

The first one is the tricky one for me. I myself was running with `setup.dashboards.enabled: true` for the first time in a new environment because it's easy. I didn't immediate get that it would do this on every subsequent startup and overwrite dashboards.

I reinstated the sentence, thanks for noticing!

ruflin:
@kvch My last comment was referring to your PR here: https://github.com/elastic/beats/pull/6814 Here we also have an overwriting config but this one is off by default. Perhaps we should have dashboards off by default too if they already exist?

loekvangool:
Discuss topic here: https://discuss.elastic.co/t/should-beats-modules-overwrite-kibana-dashboards/127984

kvch:
The other updating option is disabled by default, because I did not want to change the default behaviour as someone might be building on that. I would go with the current behaviour of dashboard loading now.

In the future it would be nice to decide if loaded object are updated or not and introduce it to all objects regardless where/when it's loaded.

dedemorton:
jenkins test this please

dedemorton:
Build failures are unrelated to these changes, so I am merging.

dedemorton:
This needs to be backported to 6.3 and forward ported to master.

PR #6831, opened 2018-04-12
Remove flush interval from Kafka run script

ruflin:
The very short flush interval caused the container sometimes to crash. Removing it should make tests more stable.


ruflin:
jenkins, test it

PR #6833, opened 2018-04-12
Remove version from docker-compose project name

exekias:
Latest docker-compose version includes `-` in the project name, while
previous versions were striping it. This is causing tests to fail as containers
are no longer in the same network as beats. (beats is launched by local docker-compose
which is in an old version, while testing containers are launched by the new version)

I'm removing the version part of the current name to ensure we support both old and
new docker-compose behaviors, the commit hash is there already anyway.

This PR is also pinning existing python requirements to avoid new
versions breaking tests.


libbeat/tests/system/requirements.txt
```diff
@@ -1,11 +1,29 @@
...
-autopep8
-six
-docker-compose
+autopep8==1.3.5
+backports.ssl-match-hostname==3.5.0.1
+cached-property==1.4.2
+certifi==2018.1.18
+chardet==3.0.4
+docker==3.2.1
+docker-compose==1.21.0
```
jsoriano: What about updating also the version of docker-compose in `.travis.yml`?

ruflin:
@exekias Will have to backport this to 6.2 too?

exekias:
We can backport it to be sure, although I think it's not affected as the version there doesn't contain `-`

jsoriano:
We could pin versions in 6.2 in any case.

libbeat/tests/system/requirements.txt
exekias: Done, in theory it should not affect the result, but :+1: to use the same one there

exekias:
jenkins, test it again please

PR #6836, opened 2018-04-12
Add namespaces to monitoring registry

ruflin:
So far for metricsets and modules only combined metrics were collected. There was no visibility into a single metricset. This PR provides the base functionality to use have each metricset provide it's own metrics.

The BaseMetricSet was extended with an `ID()` and `Metrics()` method. The id is required to have a unique identifier for each metricset during reporting and the `Metrics()` method can be used inside each Mmetricsetetricset to access the metricset specific registry.

Currently the data collected from all metricsets are exposed under `/dataset`. This is probably going to change and is mainly to show case the first implementation. The data exposed looks as following:

```
{
  "05d4bd84-2ca2-4d9d-862d-c7822b4389f5": {
    "id": "05d4bd84-2ca2-4d9d-862d-c7822b4389f5",
    "metricset": "process_summary",
    "module": "system",
    "starttime": "2018-04-17 15:41:34.154435426 +0200 CEST m=+10.037457766"
  },
  "2f85ed18-343b-4c33-865c-44e624cdaf6d": {
    "id": "2f85ed18-343b-4c33-865c-44e624cdaf6d",
    "metricset": "load",
    "module": "system",
    "starttime": "2018-04-17 15:41:34.15164335 +0200 CEST m=+10.034665690"
  }
}
```

The initial idea was to expose an array instead of map with unique keys. But having a map with unique keys makes it easier to add and remove registries when a metricset runner is started / stopped and the functionality already existed. In case the above metrics were reported to Elasticsearch, each block would be one event. To automate this conversion, we could use a flag on a registry to potential have a different reporting for Elasticsearch.

The concept of namespaces was added to manage the existing registries instead of just having on global (`Default`). This can be used in the future for additional API endpoints.

### DIFF
--- a/metricbeat/mb/builders.go
+++ b/metricbeat/mb/builders.go
@@ -6,8 +6,10 @@ import (
 
 	"github.com/joeshaw/multierror"
 	"github.com/pkg/errors"
+	"github.com/satori/go.uuid"
 
 	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/monitoring"
 )
 
 var (
@@ -156,10 +158,21 @@ func newBaseMetricSets(r *Register, m Module) ([]BaseMetricSet, error) {
 	for _, name := range metricSetNames {
 		name = strings.ToLower(name)
 		for _, host := range hosts {
+			id := uuid.NewV4().String()
+			metrics := monitoring.NewRegistry()
+			monitoring.NewString(metrics, "module").Set(m.Name())
+			monitoring.NewString(metrics, "metricset").Set(name)
+			if host != "" {
+				monitoring.NewString(metrics, "host").Set(host)
+			}
+			monitoring.NewString(metrics, "id").Set(id)
+
 			metricsets = append(metricsets, BaseMetricSet{
-				name:   name,
-				module: m,
-				host:   host,
+				id:      id,
+				name:    name,
+				module:  m,
+				host:    host,
+				metrics: metrics,
 			})
 		}
 	}
--- a/metricbeat/mb/mb.go
+++ b/metricbeat/mb/mb.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/monitoring"
 )
 
 const (
@@ -78,6 +79,7 @@ func (m *BaseModule) UnpackConfig(to interface{}) error {
 // addition to this interface, all MetricSets must implement either
 // EventFetcher or EventsFetcher (but not both).
 type MetricSet interface {
+	ID() string     // Unique ID identifying a running MetricSet.
 	Name() string   // Name returns the name of the MetricSet.
 	Module() Module // Module returns the parent Module for the MetricSet.
 	Host() string   // Host returns a hostname or other module specific value
@@ -85,6 +87,7 @@ type MetricSet interface {
 	// metrics.
 	HostData() HostData                  // HostData returns the parsed host data.
 	Registration() MetricSetRegistration // Params used in registration.
+	Metrics() *monitoring.Registry       // MetricSet specific metrics
 }
 
 // Closer is an optional interface that a MetricSet can implement in order to
@@ -213,11 +216,13 @@ func (h HostData) GoString() string { return h.String() }
 // MetricSet interface requirements, leaving only the Fetch() method to be
 // implemented to have a complete MetricSet implementation.
 type BaseMetricSet struct {
+	id           string
 	name         string
 	module       Module
 	host         string
 	hostData     HostData
 	registration MetricSetRegistration
+	metrics      *monitoring.Registry
 }
 
 func (b *BaseMetricSet) String() string {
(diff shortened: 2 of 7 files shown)


urso:
Yep, that's basically the idea. Also missing:

- Allow for namespaces to have a sub-namespace
- Support for iterating namespaces and registries within namespaces
- have ES monitoring reporter make use of namespaces and report all registries that that are marked as 'exported'.

ruflin:
@urso I changed my initial approach with namespace quite a bit and also updated the PR message. I realised most of the Add/Remove functionality is already given by the Registry if we use unique id's. The namespace are now mainly a global "registry" for the Registry.

The naming is still super confusing. I think we need to separate monitoring and metrics into 2 packages. This should make things less confusing.

metricbeat/mb/builders.go
```diff
@@ -156,10 +159,22 @@ func newBaseMetricSets(r *Register, m Module) ([]BaseMetricSet, error) {
 	for _, name := range metricSetNames {
 		name = strings.ToLower(name)
 		for _, host := range hosts {
+
```
andrewkroh: Remove this newline?

metricbeat/mb/mb.go
```diff
@@ -78,13 +79,15 @@ func (m *BaseModule) UnpackConfig(to interface{}) error {
 // addition to this interface, all MetricSets must implement either
 // EventFetcher or EventsFetcher (but not both).
 type MetricSet interface {
+	ID() string     // Unique id identifying a running MetricSet.
```
andrewkroh: s/id/ID/

metricbeat/mb/mb.go
```diff
@@ -78,13 +79,15 @@ func (m *BaseModule) UnpackConfig(to interface{}) error {
...
 	Name() string   // Name returns the name of the MetricSet.
 	Module() Module // Module returns the parent Module for the MetricSet.
 	Host() string   // Host returns a hostname or other module specific value
 	// that identifies a specific host or service instance from which to collect
 	// metrics.
 	HostData() HostData                  // HostData returns the parsed host data.
 	Registration() MetricSetRegistration // Params used in registration.
+	Metrics() *monitoring.Registry       // Metricset specific metrics
```
andrewkroh: s/Metricset/MetricSet/ and a period.

metricbeat/mb/module/wrapper.go
```diff
@@ -106,9 +106,17 @@ func (mw *Wrapper) Start(done <-chan struct{}) <-chan beat.Event {
...
 		go func(msw *metricSetWrapper) {
+			metricsPath := fmt.Sprintf("%s", msw.ID())
+			registry := monitoring.GetNamespace("dataset").GetRegistry()
+
+			defer registry.Remove(metricsPath)
 			defer releaseStats(msw.stats)
 			defer wg.Done()
 			defer msw.close()
+
+			registry.Add(metricsPath, msw.Metrics(), monitoring.Full)
```
andrewkroh: It seems redundant to put the metrics under the ID key AND to have an id field in the object.

metricbeat/mb/module/wrapper.go
```diff
@@ -106,9 +106,17 @@ func (mw *Wrapper) Start(done <-chan struct{}) <-chan beat.Event {
 	wg.Add(len(mw.metricSets))
 	for _, msw := range mw.metricSets {
 		go func(msw *metricSetWrapper) {
+			metricsPath := fmt.Sprintf("%s", msw.ID())
```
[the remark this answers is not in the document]
ruflin: This is a left over as I used longer metric paths initially. Will clean it up.

metricbeat/mb/module/wrapper.go
```diff
@@ -106,9 +106,17 @@ func (mw *Wrapper) Start(done <-chan struct{}) <-chan beat.Event {
...
+			registry := monitoring.GetNamespace("dataset").GetRegistry()
```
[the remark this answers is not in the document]
ruflin: Long term this will the case. I didn't change it in this PR as I didn't want to change any of the existing stats outputs as it could break central monitoring.

metricbeat/mb/builders.go
ruflin: of course ;-)

metricbeat/mb/module/wrapper.go
ruflin: I'm going back and forth on this one. My initial plan was to show an array instead of having a list with `id` as key. The nice part with the `id` in the key is that it works out of the box with our current implementation. The disadvantage is that the data needs to be sent in a different format to elasticsearch (1 event per entry). And that is where the id also in the object is coming from. It makes parsing each object much easier. Otherwise additional logic is needed to get it out of the key.

metricbeat/mb/module/wrapper.go
```diff
@@ -106,9 +106,17 @@ func (mw *Wrapper) Start(done <-chan struct{}) <-chan beat.Event {
...
+
+			defer registry.Remove(metricsPath)
 			defer releaseStats(msw.stats)
 			defer wg.Done()
 			defer msw.close()
+
+			registry.Add(metricsPath, msw.Metrics(), monitoring.Full)
+			monitoring.NewString(msw.Metrics(), "starttime").Set(time.Now().String())
```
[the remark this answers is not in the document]
ruflin: There is no format for dates yet but I think it would be valuable to have. Will change it to common.Time for now

ruflin:
@andrewkroh Could you have an other look. I would like to get this in mainly for starting to use namespaces. The format on the metrics reported will still change as I have to figure out how to modify the ES reporter to report one event for each entry in the `dataset` but still use some global data. We normally to this magic in Metricsets but would like to have an automated way here.

ruflin:
I opened a meta issue here to track the changes: https://github.com/elastic/beats/issues/7027

PR #6842, opened 2018-04-12
Syslog inputs parses RFC3164 events via TCP or UDP

### LINKED ISSUE #5862: Allow filebeat to collect syslog events
Since Filebeat is installed directly on the machine, it makes sense to allow Filebeat to collect local syslog data and send it to Elasticsearch or Logstash. With the currently available filebeat prospector it is possible to collect syslog events via UDP.

## POC

```javascript
{
  "_index": "filebeat-6.0.1-2017.12.12",
  "_type": "doc",
  "_id": "mO8YTGABuTrMYTLosnUm",
  "_score": 1,
  "_source": {
    "@timestamp": "2017-12-12T18:59:34.077Z",
    "message": "<13>Dec 12 18:59:34 testing root: Hello PH <3",
    "prospector": {
      "type": "udp"
    },
    "beat": {
      "name": "testing",
      "hostname": "testing",
      "version": "6.0.1"
    }
  },
  "fields": {
    "@timestamp": [
      "2017-12-12T18:59:34.077Z"
    ]
  }
}
```
Filebeat configuration:

```                                                                                                                                                      
- type: udp                                                                                                                                                   
  enabled: true
```

Rsyslog configuration:

```
*.* @127.0.0.1:8080
```

### TODO:

- [x] Evaluate the _experimental_ status of UDP
- [ ] Decide which syslog format to support
- [x] Create an Ingest pipeline to extract the data.
- [x] Discuss if using the UDP type instead of syslog is confusing for the user, should we create a specific type for syslog.

**Related links:** 
See existing Logstash plugins concerning syslog. 
https://github.com/logstash-plugins/?utf8=%E2%9C%93&q=syslog&type=&language=

ph:
The Syslog inputs will use the UDP and TCP source lib, allowing the same socket behavior and the
same options as the two existing inputs. The parser is a state machine build with [ragel](http://www.colm.net/open-source/ragel/) and allow to parse [FC3164](https://tools.ietf.org/html/rfc3164) events with some less than perfect variants, if the received event is a complete
[FC3164](https://tools.ietf.org/html/rfc3164)  we will extract all of them, for us the minimum valid message MUST have the `date` and the `message` defined.

Anything else we will log and drop them.

Fields:
* priority
* timestamp
* program
* pid
* message
* facility: extracted from the priority
* severity: extracted from the priority
* severity_label: mapped from the [official list](https://tools.ietf.org/html/rfc3164)
* facility_label: mapped from the [official list](https://tools.ietf.org/html/rfc3164)

Sample Configuration:

```yaml
  #enabled: false

  #protocol.tcp:
    # The host and port to receive the new event
    #host: "localhost:9000"

    # Character used to split new message
    #line_delimiter: "\n"

    # Maximum size in bytes of the message received over TCP
    #max_message_size: 20MiB

    # The number of seconds of inactivity before a remote connection is closed.
    #timeout: 300s

  #protocol.udp:
    # The host and port to receive the new event
    #host: "localhost:9000"

    #max_message_size: 10240
```

Limitations:
* Doesn't support multiline events like darwin can do, we need to extract the multiline logic from the log input.
* Only support RFC3164, RFC5424 will require more work on the parser.

close #5862

## Missing in this PR
* Update the fields.yml with the new fields, validate with ECS.
* docs / asciidoc

## Note to reviewers:
* event.go is built by the state machine as we go, there is no validation in this class since all the validation
is done at the state machine level.
* Ignore the `parser.go`, this is a generated file from `parser.rl` and `syslog_rfc3164.rl`, it's a messy file using all the possible
tricks in the book to make it really fast.
* parser_test.go contains all the usecase I've defined, I might miss some, please add suggestions.
* I haven't made the build install ragel, unless we have more than one file.
* I've changed a bit how TCP/UDP interract to make it easier to use, the changes should be minimum.

## Speed
Performance looks OK to me, I don't have the number from my regular expression poc, but it's much better.
```
pkg: github.com/elastic/beats/filebeat/input/syslog
BenchmarkParser-8       10000000               176 ns/op              64 B/op          3 allocs/op
PASS
ok      github.com/elastic/beats/filebeat/input/syslog  1.977s
```


filebeat/input/syslog/input.go
```diff
@@ -0,0 +1,205 @@
...
+	"github.com/elastic/beats/filebeat/util"
+	"github.com/elastic/beats/libbeat/beat"
+	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/common/cfgwarn"
+	"github.com/elastic/beats/libbeat/logp"
+)
+
+// Parser is generated from a ragel state machine using the following command:
+//#go generate ragel -Z -G2 parser.rl -o parser.go
+var log = logp.NewLogger("syslog")
```
andrewkroh: This won't work as desired.

> This should never be used from any global contexts, otherwise you will receive a no-op Logger. This is because the logp package needs to be initialized first. Instead create new Logger instance that your object reuses.

https://godoc.org/github.com/elastic/beats/libbeat/logp#NewLogger

filebeat/input/syslog/input.go
```diff
@@ -0,0 +1,205 @@
...
+	"github.com/elastic/beats/filebeat/inputsource"
+	"github.com/elastic/beats/filebeat/util"
+	"github.com/elastic/beats/libbeat/beat"
+	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/common/cfgwarn"
+	"github.com/elastic/beats/libbeat/logp"
+)
+
+// Parser is generated from a ragel state machine using the following command:
+//#go generate ragel -Z -G2 parser.rl -o parser.go
```
andrewkroh: You can write `//go:generate ragel -Z -G2 parser.rl -o parser.go` so that `go generate` will run the commands for you.

filebeat/input/syslog/syslog_rfc3164.rl
```diff
@@ -0,0 +1,45 @@
+%%{
```
andrewkroh: Interesting!

filebeat/inputsource/tcp/config.go
```diff
@@ -7,6 +7,9 @@ import (
 	"github.com/dustin/go-humanize"
 )
 
+// NAME is the human readable name and identifier.
+const NAME = "tcp"
```
andrewkroh: CapitalCamelCase instead?

filebeat/input/syslog/input.go
ph: Oh I've commented it for testing something, I will revert it back.

filebeat/_meta/common.reference.p2.yml
```diff
@@ -258,6 +258,31 @@ filebeat.inputs:
...
+    # Character used to split new message
+    #line_delimiter: "\n"
+
+    # Maximum size in bytes of the message received over TCP
+    #max_message_size: 20MiB
+
+    # The number of seconds of inactivity before a remote connection is closed.
+    #timeout: 300s
+
+  #protocol.udp:
```
ruflin: Can one input have tcp and udp at the same time?  If not, we should probably have the syslog input defined twice here.

filebeat/_meta/common.reference.p2.yml
```diff
@@ -258,6 +258,31 @@ filebeat.inputs:
...
+    # The host and port to receive the new event
+    #host: "localhost:9000"
+
+    # Maximum size of the message received over UDP
+    #max_message_size: 10240
```
ruflin: Should we bring message size for the 2 in line? Also the way it's configured, I like the humanize one.

filebeat/input/syslog/parser_test.go
```diff
@@ -0,0 +1,316 @@
+package syslog
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestParseSyslog(t *testing.T) {
```
ruflin: Nice, that are the tests I was looking for.

filebeat/input/syslog/input.go
```diff
@@ -0,0 +1,217 @@
...
+	if ev.Hostname() != "" {
+		f["hostname"] = ev.Hostname()
+	}
+
+	if ev.Program() != "" {
+		f["program"] = ev.Program()
+	}
+
+	if ev.HasPid() {
+		f["pid"] = ev.Pid()
```
ruflin: Could we align the fields here already with ECS? This would be `process.pid`. 

To make sure we don't conflict, all these fields should be prefixed with `syslog.` if they are not already used fields.

kvch:
@ph Do you mind opening an issue for RFC5424 after this PR is merged?

ph:
@kvch 👍 for another issue for rfc5424, some state machine of the parser can be reused as is, the `structured` part of rfc5424 is the weirdest part..

filebeat/_meta/common.reference.p2.yml
ph: I will need to add support for humanize in udp, I will do it.

filebeat/_meta/common.reference.p2.yml
ph: it cannot for complexity reason :) I am OK to have it twice.

ruflin: I'm glad it can't. Perhaps we should have a check to make sure only one is used at the same time.

ph: Will take a note to check if we can change that error message.

ph:
@kvch followup issue for rfc5424 in https://github.com/elastic/beats/issues/6872

filebeat/input/syslog/config.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+var defaultTCP = tcp.Config{
+	LineDelimiter:  "\n",
+	Timeout:        time.Minute * 5,
+	MaxMessageSize: 20 * humanize.MiByte,
+}
+
+var defaultUDP = udp.Config{
+	MaxMessageSize: 10240,
```
kvch: Please humanize this option, as it's done in `defaultTCP`.

filebeat/_meta/common.reference.p2.yml
ph: @kvch @ruflin done in https://github.com/elastic/beats/pull/6886

filebeat/input/syslog/syslog_rfc3164.rl
```diff
@@ -0,0 +1,45 @@
...
+  single_digit_day = [1-9]>tok %day;
+  day = (space? single_digit_day | multiple_digits_day);
+
+  # Match: hh:mm:ss (24 hr format)
+  hour = ([01][0-9]|"2"[0-3])>tok %hour;
+  minute = ([0-5][0-9])>tok %minute;
+  second = ([0-5][0-9])>tok %second;
+  nanosecond = digit+;
+  time = hour ":" minute ":" second ("." nanosecond)?;
+  timestamp= month space day space time;
```
kvch: Please add a space before `=` :)

filebeat/input/syslog/syslog_rfc3164.rl
kvch: I really like this approach. Do you mind sharing why did you choose ragel? What other lexers did you consider? I am just curious.

kvch:
Thanks for implementing BSD syslog protocol. It's great that you added TCP to the supported transports, regardless that it's not required by the legacy protocol. I am very happy we went for using lexers instead of unmaintainable monster regexes.

Furthermore, with this PR we move forward with extending capabilities and use cases of Filebeat which is the "beatific" way. But it's clear that the current architecture is not flexible enough yet. I am a bit sad it can't receive syslog messages from arbitrary source (e.g file). Unfortunately, right now reading from sources and parsing the incoming bytes is too tightly coupled to achieve that. This coupling is the a major problem with the architecture of your PR. That's where all of my review notes stems from. I haven't shared those problems here, as it is out of scope of this PR. Also, this feature is valuable as is.

We really need a big refactor to address this architectural problem. We should decouple transport layer from creating, parsing and transforming events. Otherwise we would never be able to read syslog messages from a local file. (Unless, we send through to a local port over TCP or UDP :D)

I am glad to merge this PR as soon as review notes are resolved so we could focus on the refactor. FYI I am preparing a proposal of this. :)

ph:
>Furthermore, with this PR we move forward with extending capabilities and use cases of Filebeat which is the "beatific" way. But it's clear that the current architecture is not flexible enough yet. I am a bit sad it can't receive syslog messages from arbitrary source (e.g file). Unfortunately, right now reading from sources and parsing the incoming bytes is too tightly coupled to achieve that. This coupling is the a major problem with the architecture of your PR. That's where all of my review notes stems from. I haven't shared those problems here, as it is out of scope of this PR. Also, this feature is valuable as is.

Yes, this is why I was making baby steps, moving the UDP logic and the TCP out of the inputs so they can become composable. I wanted to do the same with the log inputs, the syslog parsing is currently coupled with theses two protocols, but the syslog parsing is just a function call so it can be another building block. 

> I am glad to merge this PR as soon as review notes are resolved so we could focus on the refactor. FYI I am preparing a proposal of this. :)

This is nice I was working on a similar proposal, we should probably sync on that, when I've created that feature, I've seen the same problems but it would require a bit more work to achieve without doing some refactoring.

My idea higher level was to be able mostly works with stream and leverage more of the processor interface.

To be able to do:
- inputs X
- Multiline
- Syslog
- Dissect

filebeat/input/syslog/syslog_rfc3164.rl
ph: I haven't look too much in other lexer other than https://github.com/timtadh/lexmachine

I've worked with ragel before, I've made a SQL like DSL on top of a lucene document. They really focus on speed and this is a mature project.

Also, I like the fact that the lexer and the target language are not linked, Meaning we can target both the beats syslog input and the logstash input syslog. But ragel isn't the only tool available that can do that.

filebeat/input/syslog/parser_test.go
```diff
@@ -0,0 +1,316 @@
...
+			log := fmt.Sprintf("<%d>Oct 11 10:15:15 mymachine postfix/smtpd[2000]: 'su root' failed for lonvick on /dev/pts/8", d)
+			l := newEvent()
+			Parse([]byte(log), l)
+			assert.Equal(t, d, l.Priority())
+		})
+		return
+	}
+}
+
+func TestInvalidSyslog(t *testing.T) {
```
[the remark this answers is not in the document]
ph: I can delete that, I've added all tests for somewhat invalid syslog in the main test https://github.com/elastic/beats/pull/6842/files#diff-6ae69c7f1d1edd97ba8ebcdc7df2e855R18

On failure the catch all is to only populate the `event.Message` and for us to be a valid syslog event we must have  timestamp and a message defined, everything else is just extra data.

filebeat/input/syslog/config.go
andrewkroh: It would be nice to make bytes a first class citizen in ucfg.

kvch:
@ph I would keep it as is, so we have a chance to release it in 6.3.

filebeat/input/syslog/config.go
ph: @andrewkroh I prefer to have it in the beats repository, this way we can more control `deprecation` if needed.

ph:
@kvch @ruflin I've updated the PR with the reviews comments, I've followed ECS as much as I can, I think it make sense. I've updated the fields too, added a wip docs for it.

The following is an example of the output if all fields are present:
```json
{
  "@timestamp": "2018-10-12T02:14:15.000Z",
  "@metadata": {
    "beat": "filebeat",
    "type": "doc",
    "version": "7.0.0-alpha1",
    "truncated": false
  },
  "message": "'su root' failed for lonvick on \/dev\/pts\/8 0",
  "source": "127.0.0.1:56949",
  "hostname": "wopr.mymachine.co",
  "event": {
    "severity": 5
  },
  "input": {
    "type": "syslog"
  },
  "syslog": {
    "priority": 13,
    "severity_label": "Notice",
    "facility": 1
  },
  "process": {
    "pid": 2000,
    "program": "postfix"
  },
  "agent": {
    "name": "user-level"
  },
  "prospector": {
    "type": "syslog"
  },
  "beat": {
    "name": "sashimi",
    "hostname": "sashimi",
    "version": "7.0.0-alpha1"
  }
}```

File: filebeat/docs/inputs/input-syslog.asciidoc [hunk 12]
```diff
@@ -0,0 +1,39 @@
...
+    host: "localhost:9000"
+----
+
+==== Configuration options
+
+The `syslog` input supports the following configuration options plus the
+<<{beatname_lc}-input-{type}-common-options>> described later.
+
+
+[id="{beatname_lc}-input-{type}-common-options"]
```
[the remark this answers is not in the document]
dedemorton: Yup, I'd do something similar to what I've done here:

https://raw.githubusercontent.com/elastic/beats/master/filebeat/docs/inputs/input-common-file-options.asciidoc

So maybe you would have inputs/input-common-tcp-options.asciidoc and input-common-udp-options.asciidoc. Make sure that you add a comment to the top of the file explaining how the options are shared (see input-common-file-options.asciidoc for an example). 

Of course, you'll need to use attributes to resolve the section IDs for each option because the IDs need to be unique across the book. Something like:

[id="{beatname_lc}-input-{type}-option-name"]

ruflin:
@ph ECS implementation looks really good 🎉 . The only thing I wonder is the content of `agent.name`. Where is this value coming from?

ph:
@ruflin `agent.name` is kind of tricky, I think it will move it in the syslog, in BSD world its the `facility` this is extracted from the priority with some bits operation and mapped to a static list of `facility` this list is the official list in RFC, see https://github.com/elastic/beats/pull/6842/files#diff-325cb617098659d9dad2180c415be116R43.

So its kind-of-agent

ph:
@kvch @ruflin I have made the changes to `agent.name`, extracts tcp/udp options and added docs.
The PR was rebased + squashed.

rhclayto:
I believe there is a bug in syslog_rfc3164.rl .

See referenced issue:
https://github.com/elastic/beats/issues/9323

PR #6844, opened 2018-04-12
Set elasticsearch discovery to single-mode in testing environments

jsoriano:
This tries to address problems seen starting logstash and elasticsearch in test builds.


jsoriano:
It does't fix the issue in travis :confused:

ruflin:
Now it happens on Jenkins too :-) Probably because your change invalidated the images somehow.

PR #6845, opened 2018-04-12
Cherry-pick #6833 to 6.2: Remove version from docker-compose project name

exekias:
Cherry-pick of PR #6833 to 6.2 branch. Original message: 

Latest docker-compose version includes `-` in the project name, while
previous versions were striping it. This is causing tests to fail as containers
are no longer in the same network as beats. (beats is launched by local docker-compose
which is in an old version, while testing containers are launched by the new version)

I'm removing the version part of the current name to ensure we support both old and
new docker-compose behaviors, the commit hash is there already anyway.

This PR is also pinning existing python requirements to avoid new
versions breaking tests.


exekias:
Wrong PR, will reopen

PR #6773, opened 2018-04-04
Fix unstable ceph tests

jsoriano:
Ceph docker healtcheck only waited to have a response to `ceph --status`, but at this moment monitor health information can be still unavailable, what makes `test_monitor_health` to fail. Healthcheck checks now that the health endpoint has information about monitors, this can take about one minute, so compose timeout has been increased for these tests.

Some additional assertions have been also added to check not only that events are received but also that they don't contain errors.

PR #6847, opened 2018-04-12
Print container logs when compose up fails in tests

jsoriano:
Print container logs when compose up fails in test, and make it fail fast if container stops while waiting it to be healthy.

Splitted PR from #6773


ruflin:
That is a nice change that hopefully should make it easier for us to find system tests issue easier in the future.

PR #6851, opened 2018-04-13
Fixed omission of github.com in git clone path

cjimti:
Should clone into ${GOPATH}/src/github.com/elastic/beats as following instructions point to files in that path.


ruflin:
jenkins, test it

ruflin:
@cjimti Thanks for providing the fix.

dedemorton:
This needs to be backported to 6.3

PR #6852, opened 2018-04-13
[6.x] Do not render irrelevant docs for apm server. (#6841)

simitt:
Backports the following commits to 6.x:
 - Do not render irrelevant docs for apm server.  (#6841)


ruflin:
@simitt Thanks for the backport, but currently not needed as we still do "bulk back ports". I can open one if you need it rather urgent.

simitt:
Ok, wasn't aware of that. It is not urgent to backport.

PR #6853, opened 2018-04-13
Fix file permission in the MongoDB module

tsg:
This is breaking the packaging TestRPM tests because they check for
proper permissions on all yaml files.


kvch:
I opened an issue to fix this problem in the generator: https://github.com/elastic/beats/issues/6856

PR #6861, opened 2018-04-13
Change format of exported fields content

ruflin:
This PR updates the generate_fields_docs.py to make the layout of the reference content easier to read. It uses definition lists instead of sections to format the options. This makes the content easier to scan, and it makes the "On this page" section more useful.

Here's an example. I'm still not happy with repeated section names, but this PR at least makes the content easier to read.

image


exekias:
This will need a rebase

ruflin:
done

PR #6863, opened 2018-04-13
Revert docker compose in travis to 1.11.1

jsoriano:
Issues appeared in libbeat testsuite after upgrading to docker compose 1.21.0 regarding communication problems with the logstash container.

Reverting by now to 1.11.1 as it seems to work on this version, we'll evaluate to upgrade to newer versions in the future.


exekias:
jenkins, test it again please

PR #6868, opened 2018-04-15
made changes to support repeating flush pattern

coder-hub:
Made changes to support cases where the flush pattern could repeat and they need to be combined as one multiline message

For example,
date some message
   repeating line 1
   repeating line 2
   repeating line n

If the need is to combine all of the above lines as 1 multiline event, it could be supported with additional flag that the flush pattern is a repeating one.

### DIFF
--- a/filebeat/harvester/reader/multiline_test.go
+++ b/filebeat/harvester/reader/multiline_test.go
@@ -131,6 +131,24 @@ func TestMultilineBeforeNegateOKWithEmptyLine(t *testing.T) {
 	)
 }
 
+func TestMultilineRepeatingFlushPattern(t *testing.T) {
+	flushMatcher := match.MustCompile(`EventEnd`)
+	pattern := match.MustCompile(`EventStart`)
+
+	testMultilineOK(t,
+		MultilineConfig{
+			Pattern:      &pattern,
+			Negate:       true,
+			Match:        "after",
+			FlushPattern: &flushMatcher,
+			LastMatch:    true,
+		},
+		2, //first two non-matching lines, will be merged to one event
+		"EventStart\nEventId: 1\nEventEnd\nEventEnd\nEventEnd\nEventEnd\n",
+		"EventStart\nEventId: 2\nEventEnd\nEventEnd\nEventEnd\n",
+	)
+}
+
 func testMultilineOK(t *testing.T, cfg MultilineConfig, events int, expected ...string) {
 	_, buf := createLineBuffer(expected...)
 	reader := createMultilineTestReader(t, buf, cfg)
--- a/filebeat/harvester/reader/multiline.go
+++ b/filebeat/harvester/reader/multiline.go
@@ -21,17 +21,19 @@ import (
 // Errors will force the multiline reader to return the currently active
 // multiline event first and finally return the actual error on next call to Next.
 type Multiline struct {
-	reader       Reader
-	pred         matcher
-	flushMatcher *match.Matcher
-	maxBytes     int // bytes stored in content
-	maxLines     int
-	separator    []byte
-	last         []byte
-	numLines     int
-	err          error // last seen error
-	state        func(*Multiline) (Message, error)
-	message      Message
+	reader              Reader
+	pred                matcher
+	flushMatcher        *match.Matcher
+	lastMatch           bool
+	flushMatchedAlready bool
+	maxBytes            int // bytes stored in content
+	maxLines            int
+	separator           []byte
+	last                []byte
+	numLines            int
+	err                 error // last seen error
+	state               func(*Multiline) (Message, error)
+	message             Message
 }
 
 const (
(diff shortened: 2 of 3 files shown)


ruflin:
@coder-hub Could you add some unit tests in Golang for this and also system test (under tests/system/*) with actual log files to verify this? 

Can you sign the CLA?

ruflin:
@urso your thoughts on this one would be good.

coder-hub:
I signed the CLA after submitting the PR. I didnt see any unit tests and that's why i didnt add any tests. I will check the integration/system tests. In the mean time, please review and let me know for any questions.

urso:
TBH, the `flush_pattern` setting is rarely used correctly, and so far I've seen only one case it was really required.

 I guess the use-cases is more like this:

```
date some message
repeating line 1
repeating line 2
repeating line n
date some other message
date yet some other message
repeating line 1
repeating line 2
date one more
```

with use wanting to create 4 events like:

```
date some message
repeating line 1
repeating line 2
repeating line n

date some other message

date yet some other message
repeating line 1
repeating line 2

date one more
```

coder-hub:
May i know the next steps, please? This enhancement will help simplify our filebeat/logstash configurations by a whole lot. Thanks.

urso:
Is this PR anyhow related to this? https://discuss.elastic.co/t/flush-pattern/127509/2

TBH I would like not to add any more changes or controversial features to the multiline support in filebeat, unless there is a very good reason to do so. Given your current description and explanations, I don't see any good reason for this change.

Can you please explain more detailed the actual use-case and why exactly this feature is required? The way you described it + my example given, I would say it's already solved by the other multiline settings. But maybe I'm missing something here.

How does this change simplify your Logstash configuration?

urso:
Why not use an `or` in your pattern expression? Like:

```
multiline.pattern: '^(DateA|DateB|DateC)'
```

if these patterns become too complicated, one can use references to build these patterns:

```

filebeat.prospectors:
- ...
  multiline.pattern: '^(${patterns.date_a}|${patterns.date_b}|${patterns.date_c})'

patterns:
  date_a: '(DateA)'
  date_b: '(DateB)'
  date_c: '(DateC)'

```

When building regexes, don't copy those from LS. These patterns are quite generic and more expensive to execute, then actual use-case tailored patterns.

Having different timestamp formats is a sign of a log file getting entries from different log sources/formatters. This is quite bad A practice. Having to rely on 'continuous flush patterns' looks quite fragile to me. By matching on date patterns, you are still safe if one of the other log sources has/introduces multiline events.

coder-hub:
No that wont work because the first line of the multiline always has a single date format. If i give multiple date formats, i will get other multiline messages that i dont need.

This PR also supports another use case where there are empty lines in the log file and we wsnt to flush based on repeating empty lines.

coder-hub:
Can you please let me know where the tests related to flush pattern are? Do i need docker setup to run the tests?

coder-hub:
I have added the tests. Thanks.

coder-hub:
@urso may i know the next steps, please?

filebeat/harvester/reader/multiline_test.go
```diff
@@ -131,6 +131,24 @@ func TestMultilineBeforeNegateOKWithEmptyLine(t *testing.T) {
...
+			Negate:       true,
+			Match:        "after",
+			FlushPattern: &flushMatcher,
+			LastMatch:    true,
+		},
+		2, //first two non-matching lines, will be merged to one event
+		"EventStart\nEventId: 1\nEventEnd\nEventEnd\nEventEnd\nEventEnd\n",
+		"EventStart\nEventId: 2\nEventEnd\nEventEnd\nEventEnd\n",
+	)
+}
```
jsoriano: This could pass with existing options, like this:
```
func TestMultilineRepeatingFlushPattern(t *testing.T) {
        pattern := match.MustCompile(`EventStart`)

        testMultilineOK(t,
                MultilineConfig{
                        Pattern: &pattern,
                        Negate:  true,
                        Match:   "after",
                },
                2, //first two non-matching lines, will be merged to one event
                "EventStart\nEventId: 1\nEventEnd\nEventEnd\nEventEnd\nEventEnd\n",
                "EventStart\nEventId: 2\nEventEnd\nEventEnd\nEventEnd\n",
        )
}
```

jsoriano:
Hey,

This has been stalled for a while, I think the problem is that it is not clear the issues that this change would solve. Multiline already has many options and is sometimes complex to configure it correctly, but adding more options will only make it even more complex.

If this is a real issue, it would be good to have some real world examples that could help to confirm  that this cannot be solved with current features.

In the meantime I am going to close this PR, it may be good to open a topic in https://discuss.elastic.co to discuss about possible multiline configurations for specific logs.

Thanks @coder-hub for your effort here!

PR #6874, opened 2018-04-16
Don't stop Metricbeat if aerospike server is down

exekias:
Ensure Aerospike module constructor doesn't return an error if it cannot
immediately connect to the server.


exekias:
failing tests are not related

PR #6877, opened 2018-04-16
Commit registry writes to stable storage to avoid corrupt registry files

### LINKED ISSUE #6792: Fix corrupt registry occurring on power failure
This PR fixes a problem with the winlogbeat registry on windows: https://github.com/elastic/beats/pull/2434

Both winglogbeat and filebeat write registry files, by first creating a new file and then replacing the old file using move(Rename).

In winlogbeat it is solved by enabling direct write through when writing the registry file (O_DIRECT + O_SYNC?).

The main issue is, a `Close` operation on a file does not necessarily imply the file is flushed to disk in full. This is no windows only issue. Instead of relying on direct write through, a better sequence for filebeat/winlogbeat registry updates would be: create tmp file -> serialize registry -> fsync tmp file -> close tmp file -> move tmp file into correct location -> run fsync on new parent directory.

Filebeat already uses O_SYNC, but I'm not sure about the current state of O_SYNC on Linux. In the past it has been known O_SYNC actually behaving like O_DSYNC (wonder if this is solved by now).

kvch:
Called `Sync` before closing the file in order to flush everything to disk before closing the registry file.

Closes #6792

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -63,6 +63,7 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
 - Fix panic when log prospector configuration fails to load. {issue}6800[6800]
 - Fix memory leak in log prospector when files cannot be read. {issue}6797[6797]
 - Add raw JSON to message field when JSON parsing fails. {issue}6516[6516]
+- Commit registry writes to stable storage to avoid corrupt registry files. {pull}6877[6877]
 
 *Heartbeat*
 
--- a/filebeat/registrar/registrar.go
+++ b/filebeat/registrar/registrar.go
@@ -280,6 +280,8 @@ func (r *Registrar) writeRegistry() error {
 		return err
 	}
 
+	// Commit the changes to storage to avoid corrupt registry files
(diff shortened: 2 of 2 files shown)


urso:
I don't think we need O_DIRECT. Alternatively to the different `Open` functions, we could call `Sync` before Close. This way we don't need the specialized windows vs. Posix code. The `(*os.File).Sync` code translates to the correct syscalls (`fsync` on Unix, `FlushFileBuffers` on Windows). 

Something went wrong. Almost all filebeat tests are failing.

ruflin:
PR will need a changelog.

kvch:
@urso @andrewkroh Thanks for the early review! I assumed that something is not right that's why I haven't added the label "review" to my PR. :)

kvch:
Apparently, the CIs are on fire...

ruflin:
The original change to windlogbeat has been quite a bit more extensive: https://github.com/elastic/beats/pull/2434 @urso I wonder if `f.Sync()` is going to cover the same cases and if yes, we should probably also simplify it in winlogbeat or even better extract it to have only 1 way of doing it.

Issue #5396, opened 2017-10-15
Add IP-addresses and MAC-addresses to event


hypp:
This is a feature request as requested [here](https://discuss.elastic.co/t/add-ip-addresses-and-mac-addresses-to-event/103465/2).

I would like to dynamically add all IP-addresses and all MAC-addresses of the sender host to each event sent by filebeat.

I propose that it is implemented as a new processor similar to add_locale.
Something like:
Loop over net.Interfaces() to get the HardwareAddr field,
skip loopback interfaces,
skip empty HardwareAddr fields,
convert to a string and store them in a list, 
and then event.Put(key, list) to every event. 
And a second inner loop over Interface.Addrs() to get the IP-addresses in a similar way. 

I think that "key" should be user configurable.

If needed I can contribute with the code for this.

ruflin:
Thanks for filing this. I think we still need to figure out where this would fit best in. I could see this as a processor like `add_host_metadata` which probably does more then just ip address and mac address, but for example also fetches the OS. @andrewkroh Other thoughts?

omairkhalid20:
Hi, @ruflin  @andrewkroh  and @hypp  , so has this feature been developed or has been parked for now.

ruflin:
@omairkhalid20 We recently added the `add_host_metadata` processor but by now it does not contain yet the mac address. But this is probably a good time to restart the discussion. 

@hypp @omairkhalid20 Interested in opening a PR against the add_host_metadata processor? That would probably fast track the discussion.

hypp:
Done!
https://github.com/elastic/beats/issues/6867

hypp:
@ruflin I can open a PR for this. Just so I understand correctly: I fork the beats repo, create a new topic branch, make my code changes, and then submit a pull request?

ruflin:
@hypp exactly. On the PR we can then directly discuss if we should have config options for it etc. Feel free to open a PR early.

hypp:
Ok, should be done now. I have never contributed before, so I hope that I got it right?

ruflin:
@hypp Looks great, left some comments on the PR. Let's continue the discussion there.

gk4:
I'm looking for this patch to apply to beats-6.2.2 and/or 6.3  Where can I find it?

ruflin:
It didn't make it into 6.3. For snapshot builds see https://github.com/elastic/beats/pull/6878#issuecomment-387323512

q2dg:
[...]

PR #6878, opened 2018-04-17
Add IP-addresses and MAC-addresses to event

hypp:
This is a pull request for issue #5396

### DIFF
--- a/libbeat/docs/processors-using.asciidoc
+++ b/libbeat/docs/processors-using.asciidoc
@@ -732,6 +732,7 @@ beta[]
 processors:
 - add_host_metadata:
     netinfo.enabled: false
+-------------------------------------------------------------------------------
 
 It has the following settings:
 
--- a/libbeat/docs/processors-using.asciidoc
+++ b/libbeat/docs/processors-using.asciidoc
@@ -727,6 +727,16 @@ forget metadata for a container, 60s by default.
 
 beta[]
 
+[source,yaml]
+-------------------------------------------------------------------------------
+processors:
+- add_host_metadata:
+    netinfo.enabled: false
+
+It has the following settings:
+
+`netinfo.enabled`:: (Optional) Default false. Include IP adresses and MAC addresses as fields host.ip and host.mac
+
 The `add_host_metadata` processor annotates each event with relevant metadata from the host machine.
 The fields added to the event are looking as following:
 
@@ -742,7 +752,9 @@ The fields added to the event are looking as following:
          "build":"16G1212",
          "platform":"darwin",
          "version":"10.12.6"
-      }
+      },
+      ip: ["192.168.0.1", "10.0.0.1"],
+      mac: ["00:25:96:12:34:56", "72:00:06:ff:79:f1"]
    }
 }
 -------------------------------------------------------------------------------
--- a/libbeat/processors/add_host_metadata/_meta/fields.yml
+++ b/libbeat/processors/add_host_metadata/_meta/fields.yml
@@ -31,11 +31,11 @@
           type: keyword
           description: >
             OS family (e.g. redhat, debian, freebsd, windows).
-        - name: net.ip
+        - name: ip
           type: ip
           description: >
             List of IP-addresses.
-        - name: net.hw
+        - name: mac
           type: keyword
           description: >
             List of hardware-addresses, usually MAC-addresses.
--- a/libbeat/processors/add_host_metadata/_meta/fields.yml
+++ b/libbeat/processors/add_host_metadata/_meta/fields.yml
@@ -31,3 +31,10 @@
           type: keyword
           description: >
             OS family (e.g. redhat, debian, freebsd, windows).
+        - name: net.ip
+          description: >
+            List of IP-addresses.
+        - name: net.hw
+          description: >
+            List of hardware-addresses, usually MAC-addresses.
+
--- a/libbeat/processors/add_host_metadata/_meta/fields.yml
+++ b/libbeat/processors/add_host_metadata/_meta/fields.yml
@@ -32,9 +32,11 @@
           description: >
             OS family (e.g. redhat, debian, freebsd, windows).
         - name: net.ip
+          type: ip
           description: >
             List of IP-addresses.
         - name: net.hw
+          type: keyword
           description: >
             List of hardware-addresses, usually MAC-addresses.
 
--- a/libbeat/processors/add_host_metadata/add_host_metadata.go
+++ b/libbeat/processors/add_host_metadata/add_host_metadata.go
@@ -5,12 +5,13 @@ import (
 	"net"
 	"time"
 
+	"github.com/pkg/errors"
+
 	"github.com/elastic/beats/libbeat/beat"
 	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/processors"
 	"github.com/elastic/go-sysinfo"
 	"github.com/elastic/go-sysinfo/types"
-	"github.com/pkg/errors"
 )
 
 func init() {
--- a/libbeat/processors/add_host_metadata/add_host_metadata.go
+++ b/libbeat/processors/add_host_metadata/add_host_metadata.go
@@ -111,7 +111,7 @@ func (p addHostMetadata) getNetInfo() ([]string, []string, error) {
 	// Get all interfaces and loop through them
 	ifaces, err := net.Interfaces()
 	if err != nil {
-		return ipList, hwList, err
+		return nil, nil, err
 	}
 	for _, i := range ifaces {
 		// Skip loopback interfaces
--- a/libbeat/processors/add_host_metadata/add_host_metadata.go
+++ b/libbeat/processors/add_host_metadata/add_host_metadata.go
@@ -75,8 +75,8 @@ func (p *addHostMetadata) loadData() {
 
 		// IP-address and MAC-address
 		var ipList, hwList = p.getNetInfo()
-		p.data.Put("host.net.ip", ipList)
-		p.data.Put("host.net.hw", hwList)
+		p.data.Put("host.ip", ipList)
+		p.data.Put("host.mac", hwList)
 
 		p.lastUpdate = time.Now()
 	}
--- a/libbeat/processors/add_host_metadata/add_host_metadata.go
+++ b/libbeat/processors/add_host_metadata/add_host_metadata.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/elastic/beats/libbeat/beat"
 	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/logp"
 	"github.com/elastic/beats/libbeat/processors"
 	"github.com/elastic/go-sysinfo"
 	"github.com/elastic/go-sysinfo/types"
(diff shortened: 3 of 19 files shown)


libbeat/processors/add_host_metadata/_meta/fields.yml
```diff
@@ -31,3 +31,10 @@
           type: keyword
           description: >
             OS family (e.g. redhat, debian, freebsd, windows).
+        - name: net.ip
+          description: >
+            List of IP-addresses.
+        - name: net.hw
+          description: >
+            List of hardware-addresses, usually MAC-addresses.
```
ruflin: I assume the type here would be `keyword`.

libbeat/processors/add_host_metadata/_meta/fields.yml
```diff
@@ -31,3 +31,10 @@
           type: keyword
           description: >
             OS family (e.g. redhat, debian, freebsd, windows).
+        - name: net.ip
+          description: >
+            List of IP-addresses.
```
ruflin: We can use the type `ip` here: https://www.elastic.co/guide/en/elasticsearch/reference/6.2/ip.html

libbeat/processors/add_host_metadata/add_host_metadata.go
```diff
@@ -71,10 +72,54 @@ func (p *addHostMetadata) loadData() {
 		if p.info.OS.Build != "" {
 			p.data.Put("host.os.build", p.info.OS.Build)
 		}
+
+		// IP-address and MAC-address
+		var ipList, hwList = p.getNetInfo()
```
ruflin: I'm thinking to make this information optional, meaning it is not sent by default but we have a config option in the processor to enable it. Not sure yet how we should call the config options, any suggestions?

libbeat/processors/add_host_metadata/add_host_metadata.go
```diff
@@ -71,10 +72,54 @@ func (p *addHostMetadata) loadData() {
...
+		p.data.Put("host.net.ip", ipList)
```
ruflin: I suggest to call these fields `host.ip` and `host.mac`.

libbeat/processors/add_host_metadata/_meta/fields.yml
hypp: Ok, I'll fix that

libbeat/processors/add_host_metadata/_meta/fields.yml
hypp: Yes, of course! I'll fix that too.

libbeat/processors/add_host_metadata/add_host_metadata.go
hypp: Perhaps
netinfo: enabled/disabled 
or 
add_netinfo: true/false?

ruflin: Looking at the code I was even wondering if we should separate mac address and ip. I think lots of people want to have the list of ip addresses but are less interested in the actual mac address. This could lead to a config like:

```
add_fields: [`ip`, `mac`] 
```
I don't like add_fields name to much but the gist is that we could add more fields in the future here.

@andrewkroh Thoughts?

hypp: Should we do that for all fields or just for ip and mac?

ruflin: The way I was thinking that the fields we have now are always there. And we only have that for additional fields like the once proposed here.

hypp: Ok, I'll fix that.

andrewkroh: > The way I was thinking that the fields we have now are always there. And we only have that for additional fields like the once proposed here.

Personally I'd like be able to fully select each field that is included. Keeping the current set as the default. The behavior would be similar to what you get with Metricbeat `metricsets` (there's a default set, and when you define any value you override the default set).

ruflin: @andrewkroh If I understand this correctly, my config example above would mean `host.name` for example would not be sent?

andrewkroh: > add_fields: [`ip`, `mac`] 

Correct. That would give you only those two fields. But I wouldn't call it `add_fields` if we went with this behavior.

hypp: I have time to work on this on friday, so it would be great to have a decision by then.

ruflin: @hypp We had a quick internal chat and decided to go with the proposal for `netinfo.enabled` for now and we can still improve it later.

libbeat/processors/add_host_metadata/add_host_metadata_test.go
```diff
@@ -31,4 +34,39 @@ func TestRun(t *testing.T) {
...
+
+	v, err = newEvent.GetValue("host.ip")
+	assert.Error(t, err)
+	assert.Nil(t, v)
+
+	v, err = newEvent.GetValue("host.mac")
+	assert.Error(t, err)
+	assert.Nil(t, v)
+
+	event = &beat.Event{
```
ruflin: Could you split this up into 2 tests? Like this if the tests fail we can already tell based on the name which part is probably not working.

hypp: @ruflin Is it really enough to `make update` in the metricbeat directory? Or should I do it for every beat?

ruflin: Sorry, my mistake. You have to run it in the top level `beats` directory. Same for `make fmt`.

libbeat/processors/add_host_metadata/add_host_metadata.go
```diff
@@ -71,10 +83,57 @@ func (p *addHostMetadata) loadData() {
...
 }
 
+func (p addHostMetadata) getNetInfo() ([]string, []string) {
+	var ipList []string
+	var hwList []string
+
+	// Get all interfaces and loop through them
+	ifaces, err := net.Interfaces()
+	if err != nil {
+		return ipList, hwList
```
ruflin: Instead or returning the empty lists could we return here the error? Meaning it would return `nil, nil, err`. On line 89 we could then check for the error and would not add the fields in case of an error and could also log the error.

libbeat/processors/add_host_metadata/add_host_metadata.go
```diff
@@ -71,10 +83,57 @@ func (p *addHostMetadata) loadData() {
...
+
+		hw := i.HardwareAddr.String()
+		// Skip empty hardware addresses
+		if hw != "" {
+			hwList = append(hwList, hw)
+		}
+
+		addrs, err := i.Addrs()
+		if err != nil {
+			return ipList, hwList
```
ruflin: Similar question as above. I wonder if in case of an error we should just log it but continue trying to collect the addresses from the interfaces instead of returning.

libbeat/processors/add_host_metadata/add_host_metadata.go
```diff
@@ -71,10 +83,57 @@ func (p *addHostMetadata) loadData() {
 		if p.info.OS.Build != "" {
 			p.data.Put("host.os.build", p.info.OS.Build)
 		}
+
+		if p.config.NetInfoEnabled {
+			// IP-address and MAC-address
+			var ipList, hwList = p.getNetInfo()
+			p.data.Put("host.ip", ipList)
```
ruflin: Reviewing the code below and the possibility that ipList and hwList can be empty, we should check for `len(x) == 0` and in case of 0 not add the field.

libbeat/processors/add_host_metadata/config.go
```diff
@@ -0,0 +1,16 @@
+package add_host_metadata
+
+//import (
```
ruflin: leftover?

libbeat/processors/add_host_metadata/add_host_metadata.go
hypp: Sure, do you have a code example of logging, that I could look at?

libbeat/processors/add_host_metadata/add_host_metadata.go
hypp: Ok, I'll fix that

libbeat/processors/add_host_metadata/add_host_metadata.go
hypp: It is a matter of taste, I prefer to fail fast. You decide.

libbeat/docs/processors-using.asciidoc
```diff
@@ -727,6 +727,16 @@ forget metadata for a container, 60s by default.
 
 beta[]
 
+[source,yaml]
+-------------------------------------------------------------------------------
+processors:
+- add_host_metadata:
+    netinfo.enabled: false
+
```
ruflin: Here the end of the source block seems to be missing.

libbeat/processors/add_host_metadata/add_host_metadata.go
```diff
@@ -71,10 +84,68 @@ func (p *addHostMetadata) loadData() {
 		if p.info.OS.Build != "" {
 			p.data.Put("host.os.build", p.info.OS.Build)
 		}
+
+		if p.config.NetInfoEnabled {
+			// IP-address and MAC-address
+			var ipList, hwList, err = p.getNetInfo()
+			if err != nil {
+				logp.Warn("Error when getting network information %v", err)
```
ruflin: We prefer not to use `Warn` as it's not clear if this is something that needs action. I suggest we set it to `Info` for now.

libbeat/processors/add_host_metadata/add_host_metadata.go
```diff
@@ -71,10 +84,68 @@ func (p *addHostMetadata) loadData() {
...
 }
 
+func (p addHostMetadata) getNetInfo() ([]string, []string, error) {
+	var ipList []string
+	var hwList []string
+
+	// Get all interfaces and loop through them
+	ifaces, err := net.Interfaces()
+	if err != nil {
+		return ipList, hwList, err
```
ruflin: Can you use `return nil, nil, err` here? This makes it clear that both ipList and hwList do not have any values yet. Code logic is exactly the same as before.

libbeat/processors/add_host_metadata/add_host_metadata.go
```diff
@@ -71,10 +84,68 @@ func (p *addHostMetadata) loadData() {
...
+		hw := i.HardwareAddr.String()
+		// Skip empty hardware addresses
+		if hw != "" {
+			hwList = append(hwList, hw)
+		}
+
+		addrs, err := i.Addrs()
+		if err != nil {
+			// If we get an error, log it and continue with the next interface
+			logp.Warn("Error when getting IP address %v", err)
```
ruflin: Lets not log each error but add it to the list. We can use `multierror` here. Outside the loop you can specify `var errs multierror.Errors` and then use `errs = append(errs, err)`. On line 145 you can then return `errs.Err()`.

libbeat/docs/processors-using.asciidoc
hypp: Should be fixed now

libbeat/processors/add_host_metadata/add_host_metadata.go
hypp: Ok!

ruflin:
jenkins, test this

ruflin:
@hypp Thanks a lot for your contribution.

hypp:
@ruflin So it is all done now? Great!

ruflin:
@hypp Yes. This should automatically end up in the 6.4 release. If you want to test it earlier, there are snapshot build available: https://beats-package-snapshots.s3.amazonaws.com/index.html?prefix=metricbeat/ (warning, just saw they are not 100% up-to-date. Need to check why).

PR #6880, opened 2018-04-17
Backport doc fixes for libbeat.

simitt:
Backports some changes to beats 6.2, mostly necessary for APM Server doc updates.

### DIFF
--- a/heartbeat/docs/index.asciidoc
+++ b/heartbeat/docs/index.asciidoc
@@ -8,6 +8,10 @@ include::{asciidoc-dir}/../../shared/attributes62.asciidoc[]
 :beatname_lc: heartbeat
 :beatname_uc: Heartbeat
 :beatname_pkg: heartbeat-elastic
+:github_repo_name: beats
+:discuss_forum: beats/{beatname_lc}
+:beat_default_index_prefix: {beatname_lc}
+:has_ml_jobs: yes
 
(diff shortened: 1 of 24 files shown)


heartbeat/docs/index.asciidoc
```diff
@@ -8,6 +8,10 @@ include::{asciidoc-dir}/../../shared/attributes62.asciidoc[]
 :beatname_lc: heartbeat
 :beatname_uc: Heartbeat
 :beatname_pkg: heartbeat-elastic
+:github_repo_name: beats
+:discuss_forum: beats/{beatname_lc}
+:beat_default_index_prefix: {beatname_lc}
+:has_ml_jobs: yes
```
ruflin: @dedemorton This is only the backport of the PR so not directly related to this PR but it seems we have `has_ml_jobs` set to yes for all beats but they only exist in Filebeat. Should we turn it off for the other beats?

dedemorton: @ruflin I think that's one of the conditions the apm team added. I'll explore some more and follow up.

simitt: Afaik before adding this the docs showed the machine learning related information for every beat. Introducing this variable enabled us to avoid having notes about machine learning in the APM docs. As we didn't want to change any semantics for beats though, the variable was set to `true` for all beats.

PR #6883, opened 2018-04-17
Change config reloading options in the short config

ruflin:
Config reloading has gone through several release cycles and no major issues have shown up. We should make reloading GA.

There are several changes in this PR:

* Config reloading is made GA but in the code it's disabled by default as otherwise this would be a breaking change
* Enabling config reloading in our short config. This will make it easier for new users to use commands like `metricbeat modules enable system` as they will work out of the box.
* Change reloading to true in K8s files. Not sure if we treat this as a breaking change?

### DIFF
--- a/deploy/kubernetes/filebeat/filebeat-configmap.yaml
+++ b/deploy/kubernetes/filebeat/filebeat-configmap.yaml
@@ -14,11 +14,11 @@ data:
         # Mounted `filebeat-prospectors` configmap:
         path: ${path.config}/prospectors.d/*.yml
         # Reload prospectors configs as they change:
-        reload.enabled: false
+        reload.enabled: true
       modules:
         path: ${path.config}/modules.d/*.yml
         # Reload module configs as they change:
-        reload.enabled: false
+        reload.enabled: true
 
     processors:
       - add_cloud_metadata:
--- a/metricbeat/metricbeat.yml
+++ b/metricbeat/metricbeat.yml
@@ -14,7 +14,7 @@ metricbeat.config.modules:
   path: ${path.config}/modules.d/*.yml
 
   # Set to true to enable config reloading
-  reload.enabled: false
+  reload.enabled: true
 
   # Period on which files under path should be checked for changes
   #reload.period: 10s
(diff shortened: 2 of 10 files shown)


deploy/kubernetes/filebeat/filebeat-configmap.yaml
```diff
@@ -14,11 +14,11 @@ data:
         # Mounted `filebeat-prospectors` configmap:
         path: ${path.config}/prospectors.d/*.yml
         # Reload prospectors configs as they change:
-        reload.enabled: false
+        reload.enabled: true
```
ruflin: @exekias Is this is a breaking change from your perspective? If yes, will remove it from the PR.

exekias: I would say it is. +1 to keep it to `false` by default, rest of the PR LGTM

exekias: Sorry, I realize now this is the configmap :man_facepalming: , I'm ok with having this enabled by default

metricbeat/metricbeat.yml
```diff
@@ -14,7 +14,7 @@ metricbeat.config.modules:
   path: ${path.config}/modules.d/*.yml
 
   # Set to true to enable config reloading
-  reload.enabled: false
+  reload.enabled: true
```
exekias: This is a breaking change too, I'm wondering if we should recommend this mode by default or not. It may make things less predictable, wdyt?

kvch: +1 for not enabling reload by default. If I were a user who is getting started and editing the configuration just to see what a Beat can do, I would not like it to restart without my explicit permission.

ruflin: @exekias Now that we have the making it GA out of the way, lets continue the discussion here. Any thoughts to the above?

exekias: Good point about the breaking change, I guess it's not. I still don't like to have it enabled by default, I don't know many (or any) projects doing live changes without the user requesting them, it's probably unexpected.

ruflin: Ok, will close the PR then :-D

ruflin:
Closing as we decided not to change the defaults in the short config as it could be an unexpected behaviour.

PR #6887, opened 2018-04-17
Metricbeat: RabbitMQ/Queue disk reads and writes count keys made optional

jaipradeesh:
disk_reads and disk_writes keys of rabbitmq management plugin
aren't available at times, throwing error.


ruflin:
jenkins, test it

ruflin:
@jaipradeesh Thanks for the PR, changes LGTM. I wonder what is the reason these metrics are sometimes not available?

jaipradeesh:
I assume its because of the persistent state of messages (or) the memory required by RabbitMQ isn't greater than 40% of machine's memory. Thus, paging to disk never happens.

https://www.rabbitmq.com/memory.html

ruflin:
@jaipradeesh Perfect, exactly the explanation I was looking for, thanks.

PR #6889, opened 2018-04-17
Bump version to 5.6.10

andrewkroh:


ruflin:
We should be extra careful with 5.6.**10** to make sure it doesn't break anything during packaging or similar in case we have somewhere a regexp which matches 1 digit version numbers.

PR #6893, opened 2018-04-18
Auditbeat recursive file watches for Windows

### LINKED ISSUE #6864: Auditbeat's file_integrity module deadlocks under Windows
When the file_integrity module is used under Windows to monitor large and/or busy filesystems, it can end up in a deadlock.

When using this configuration:
```
- module: file_integrity
  paths:
  - C:/windows
  - C:/windows/system32
  - C:/Program Files
  - C:/Users
  recursive: true
  hash_types: [sha256]
```

Auditbeat will not generate any events whatsoever.

The reason for this is that the Windows implementation of fsnotify uses a single goroutine to forward events to auditbeat and to install watches. Communication with this goroutine is done via channels. In this case, what's happening is that the queue on channel to forward events fills and the whole goroutine blocks, so its unable to install new watches.

This is because as watches are installed, events start to be received but auditbeat is not consuming them, as it first installs the watches and then starts the consumer. 

Auditbeat use of fsnotify looks like this:
```
  # setup phase
  for dir in config.paths:
     installWatch(dir)
  # run phase
  consumeEvents()
```
The recursive feature makes the problem much likely to appear for two reasons:
- it involves installing a watch for every subdirectory inside a watched directory. That's thousands of watches for `C:\windows`.
- it does not only install watches during the setup phase, but also when a subdirectory is created inside a watched directory. This means consuming events also involves installing more watches.

The simpler way to fix it requires two changes:
- Don't install any watches during setup phase. Use a goroutine inside consumeEvents() so that events are being consumed while watches are installed.
- Modify fsnotify to use OS support for recursive watches. This makes installing watches for new subdirectories unnecessary and simplifies things.

adriansr:
Added recursive file watches support to Auditbeat using the Windows API

This helps prevent a deadlock under Windows when a lot of file events are generated while watches are being installed.

Fixes #6864

### DIFF
--- a/auditbeat/module/file_integrity/eventreader_fsnotify.go
+++ b/auditbeat/module/file_integrity/eventreader_fsnotify.go
@@ -36,6 +36,14 @@ func NewEventReader(c Config) (EventProducer, error) {
 }
 
 func (r *reader) Start(done <-chan struct{}) (<-chan Event, error) {
+	if err := r.watcher.Start(); err != nil {
+		return nil, errors.Wrap(err, "unable to start watcher")
+	}
+	go r.consumeEvents(done)
+
+	// Windows implementation of fsnotify needs to have the watched paths
+	// installed after the event consumer is started, to avoid a potential
+	// deadlock. Do it on all platforms for simplicity.
 	for _, p := range r.config.Paths {
 		if err := r.watcher.Add(p); err != nil {
 			if err == syscall.EMFILE {
@@ -48,10 +56,6 @@ func (r *reader) Start(done <-chan struct{}) (<-chan Event, error) {
 		}
 	}
 
-	if err := r.watcher.Start(); err != nil {
-		return nil, errors.Wrap(err, "unable to start watcher")
-	}
-	go r.consumeEvents(done)
 	r.log.Infow("Started fsnotify watcher",
 		"file_path", r.config.Paths,
 		"recursive", r.config.Recursive)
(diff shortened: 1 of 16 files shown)


andrewkroh:
I don't see the fsnotify changes to add `SetRecursive` in the PR. And I left a comment at https://github.com/adriansr/fsnotify/commit/c9bbe1f46f1da9904baf3916a4ba4aec7f1e9000#r28653958.

Do you plan to try to contribute those changes upstream? I think I'd like to keep the import statements as `github.com/fsnotify/fsnotify` and have govendor point to your fork (e.g. `govendor fetch github.com/fsnotify/fsnotify/...::github.com/adriansr/fsnotify`).

adriansr:
@andrewkroh answered your comment in adriansr/fsnotify and changed the dependency to fsnotify/fsnotify as well

auditbeat/module/file_integrity/eventreader_fsnotify.go
```diff
@@ -36,22 +37,25 @@ func NewEventReader(c Config) (EventProducer, error) {
...
-					"open files allowed with 'ulimit -a')",
-					"file_path", p, "error", err)
-			} else {
-				r.log.Warnw("Failed to add watch", "file_path", p, "error", err)
-			}
-		}
+
+	addWatchesFirst := runtime.GOOS != "windows"
+
+	if addWatchesFirst {
```
andrewkroh: What's the benefit of configuring the paths before `Start`'ing in the non-windows case?

adriansr: I was afraid it could introduce problems in platforms other than Windows. That's why I created the system test, but at the end forgot to update this code.

I've pushed a new version were watches are always installed after Start().

andrewkroh:
It looks like Windows CI in green now. 👍 

Did you want these changes squashed or left as is?

andrewkroh:
I totally read that as "No, squash please". 🤦‍♂️

PR #6899, opened 2018-04-19
Update Filebeat hints to also support modules

exekias:
This PR adds support for modules to current Filebeat hints based autodiscover, modules can be hinted through labels like this:

```
co.elastic.logs/module: apache2
co.elastic.logs/fileset: access
```

It also allows to map the log streams from the container to different module filesets:
```
co.elastic.logs/module: apache2
co.elastic.logs/fileset.stdout: access
co.elastic.logs/fileset.stderr: error
```

This PR doesn't need changelog as hints based autodiscover hasn't been released yet

### DIFF
--- a/filebeat/autodiscover/builder/hints/logs.go
+++ b/filebeat/autodiscover/builder/hints/logs.go
@@ -2,7 +2,9 @@ package hints
 
 import (
 	"fmt"
+	"regexp"
 
+	"github.com/elastic/beats/filebeat/fileset"
 	"github.com/elastic/beats/libbeat/autodiscover"
 	"github.com/elastic/beats/libbeat/autodiscover/builder"
 	"github.com/elastic/beats/libbeat/autodiscover/template"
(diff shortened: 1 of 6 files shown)


filebeat/autodiscover/builder/hints/logs.go
```diff
@@ -22,9 +24,13 @@ const (
 	excludeLines = "exclude_lines"
 )
 
+// validModuleNames to sanitize user input
+var validModuleNames = regexp.MustCompile("[^a-zA-Z0-9]+")
```
jsoriano: We should maybe include hyphen and/or underscore, we have a case in metricbeat.

exekias: This only filters modules for filebeat, normally we wouldn't need to check this, but as this will go to the disk to retrieve module YAML we want to make sure people doesn't play with this value.

I'm ok with adding a check in a follow up PR, although it's not really needed by now

jsoriano: Ok to add it in a follow-up PR

PR #6900, opened 2018-04-19
Metricbeat: Add reference config and default metricset to kvm module

jsoriano:
Follow-up of #6265


ruflin:
Failing tests do not seem to be related.

PR #6901, opened 2018-04-19
Allow `_` and `-` in Filebeat module name hints

exekias:
Follow up of #6899, this change allows module names with `_` or `-` chars, just in case we add some in the future.

### DIFF
--- a/filebeat/autodiscover/builder/hints/logs.go
+++ b/filebeat/autodiscover/builder/hints/logs.go
@@ -25,7 +25,7 @@ const (
 )
 
 // validModuleNames to sanitize user input
-var validModuleNames = regexp.MustCompile("[^a-zA-Z0-9]+")
+var validModuleNames = regexp.MustCompile("[^a-zA-Z0-9\\_\\-]+")
 
 type logHints struct {
 	Key      string


filebeat/autodiscover/builder/hints/logs.go
```diff
@@ -25,7 +25,7 @@ const (
 )
 
 // validModuleNames to sanitize user input
-var validModuleNames = regexp.MustCompile("[^a-zA-Z0-9]+")
```
ph: change lgtm, I think we should probably make it a type and when we cast it it should fail? wdyt?

ph: I don't know too much about this part of the code, w
here is this module name configured in the yml?

exekias: It comes from autodiscover hints, so normally user would enter it through docker/k8s labels

exekias: Now I realize I'm not sure how the change would look :innocent: , let's sync tomorrow

exekias:
This one should be ready for merge

ph:
@exekias this will need a backport to 6.3, right?

exekias:
I would say there is no need, as we don't have any module with these characters yet, this change was more a preparation just in case that happens at some point.

PR #6902, opened 2018-04-19
Add file spool to queue docs

urso:


libbeat/docs/queueconfig.asciidoc
```diff
@@ -76,3 +77,147 @@ will be immediately available for consumption.
 
 The default values is 0s.
 
+[float]
+[[configuration-internal-queue-spool]]
+=== Configure the file spool queue
+
+The file spooling queue stores all events in an on disk ring buffer. The spool
```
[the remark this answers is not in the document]
dedemorton: Is a ring buffer something that end users are likely to know about? If not, it's worth adding a sentence to explain what this is.

libbeat/docs/queueconfig.asciidoc
```diff
@@ -76,3 +77,147 @@ will be immediately available for consumption.
 
 The default values is 0s.
 
+[float]
+[[configuration-internal-queue-spool]]
+=== Configure the file spool queue
```
dedemorton: Now that we have a couple of queue types, we should explain why users might choose to use a specific type of queue.

libbeat/docs/queueconfig.asciidoc
```diff
@@ -76,3 +77,147 @@ will be immediately available for consumption.
...
+has a write buffer, which new events are written to. Events written to the
+spool are forwarded to the outputs, only after the write buffer has been
+flushed successfully.
+
+The spool waits for the output to acknowledge or drop events. If the spool is
+full, no new events can be inserted. The spool will block. Space is freed only
+after a signal from the output has been received.
+
+On disk, the spool operates in pages. The `file.page_size` setting configures
+the files page size on creation time. The optimal page size depends on the
```
dedemorton: "files page size" isn't grammatically correct. Do you mean:

file's page size?
file page size?
page size for files?

search for other instances of "files page size" and replace them with whichever is correct.

libbeat/docs/queueconfig.asciidoc
```diff
@@ -76,3 +77,147 @@ will be immediately available for consumption.
...
+effective block size, used by the underlying file system.
+
+This sample configuration enables the spool with all default settings and the
+default file path:
+
+[source,yaml]
+------------------------------------------------------------------------------
+queue.spool: ~
+------------------------------------------------------------------------------
+
```
dedemorton: Might be worth pointing the user to the Configuration options section for the defaults.

libbeat/docs/queueconfig.asciidoc
```diff
@@ -76,3 +77,147 @@ will be immediately available for consumption.
...
+This sample configuration creates a spool of 512MiB, with 16KiB pages. The
+write buffer is flushed if 10MiB of contents, or 1024 events have been
+written. If oldest available event is already waiting for 5s in the write
```
dedemorton: Change to say, "If the oldest available event has been waiting for 5s..."

libbeat/docs/queueconfig.asciidoc
```diff
@@ -76,3 +77,147 @@ will be immediately available for consumption.
...
+[float]
+===== `file.path`
+
+The spool file path. The file is created on startup, if it does not exist.
+
+The default value is "${path.data}/spool.dat".
+
+===== `file.permissions`
+
+The file permissions. The permissions are applied when the file is
```
dedemorton: Change "The file" to "the file"

libbeat/docs/queueconfig.asciidoc
```diff
@@ -76,3 +77,147 @@ will be immediately available for consumption.
...
+created. In case the file already exists, The file permissions are compared
+with `file.permissions`. The spool file is not opened if the actual file
```
dedemorton: What happens when the spool file cannot be opened? What does the user do? Change the file.permissions to match the actual file, or do they simply specify a different file path? Hopeful the message they see is descriptive enough so they know what to change.

libbeat/docs/queueconfig.asciidoc
```diff
@@ -76,3 +77,147 @@ will be immediately available for consumption.
...
+The default value is 0600.
+
+
+===== `file.size`
+
+Spool file size. 
+
+The default value is 100 MiB.
+
+Note: The size should be much larger then the expected event sizes
```
dedemorton: If you use NOTE (all caps), you'll get a nicely formatted note block. otherwise, this will just be a paragraph.

libbeat/docs/queueconfig.asciidoc
```diff
@@ -76,3 +77,147 @@ will be immediately available for consumption.
...
+and write buffer size. Otherwise the queue will block, because it has not
+enough space.
+
+Note: The file size can not be changed once the file has been generated. This
```
dedemorton: can not > cannot (apply this change to all instances)

libbeat/docs/queueconfig.asciidoc
```diff
@@ -76,3 +77,147 @@ will be immediately available for consumption.
...
+limitation will be removed in the future.
+
+
+===== `file.page_size`
+
+The files page size.
```
dedemorton: see comment above

libbeat/docs/queueconfig.asciidoc
```diff
@@ -76,3 +77,147 @@ will be immediately available for consumption.
...
+
+The spool file is split into multiple pages of `page_size`. All write
+operations write a complete page.
+
+The default value is 4096 (4KiB).
+
+Note: This setting should match the file systems minimum block size. If the
```
dedemorton: same comment as above (NOTE instead of Note). I won't flag all of these.

Also it should be "file system's" (possessive) not "file systems".

libbeat/docs/queueconfig.asciidoc
```diff
@@ -76,3 +77,147 @@ will be immediately available for consumption.
...
+`page_size` is not a multiple of the file systems block size, the file system
```
dedemorton: ditto: "file system's" (possessive) not "file systems"

libbeat/docs/queueconfig.asciidoc
```diff
@@ -76,3 +77,147 @@ will be immediately available for consumption.
...
+might create additional read operations on writes.
+
+Note: The page size is only set on file creation time. It can not be changed
```
dedemorton: "only set at file creation time" sounds better to me. (comment applies to other instances)

libbeat/docs/queueconfig.asciidoc
```diff
@@ -76,3 +77,147 @@ will be immediately available for consumption.
...
+===== `write.flush.events`
+
+Number of buffered events. The write buffer is flushed once the limit is reached.
+
+The default value is 16384.
+
+===== `read.flush.timeout`
+
+If configured, the `read.flush.timeout` ensures the spool will wait for more
+events to be flushed to the spool file, If the number of available events is
```
dedemorton: A couple of typos. Should be:

...if the number of available events is less than the output's `bulk_max_size`...

ruflin:
I added the `needs_backport` label.

libbeat/docs/queueconfig.asciidoc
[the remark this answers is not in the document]
urso: I'd say yes, as ring buffers are quite common.

libbeat/docs/queueconfig.asciidoc
urso: Yeah. The interaction of a configured queue and the actual beat differs per beat. This would call  for some beat specific documentation. As this will require some more work, I'd prefer to postpone the per-beats documentation for now.

libbeat/docs/queueconfig.asciidoc
urso: `file's page size`. A file is divided into pages of `page_size`.

Note: as followup we want to add a FAQ entry on determining the optimal page size. These can differ by OS and filesystem. Tools and nomenclature is different on different OSes as well.

libbeat/docs/queueconfig.asciidoc
urso: Added a link

libbeat/docs/queueconfig.asciidoc
urso: The beat won't startup on permissions errors.

File permission error occurs if the on-disk file permissions do not match the required permissions. For example the configuration mandates 'current user only' (permissions flag 0600), but the configuration file is readable/writeable by every user in the sytem (0666).

The error message generated by the spool: 

```
<some message by the beat>: file permissions must be more strict (required permissions: 0600, actual permissions: 0666)
```

libbeat/docs/queueconfig.asciidoc
urso: I have a hard time writing docs on this setting. Did rewrite complete section. Hope it's somewhat better now.

urso:
@dedemorton I updated the docs. Thanks for the input.

I think we might need some more follow up docs. Maybe per beat, as the impact of the queue type kind of depends on the actual beat and the data source.

As FAQ entry I can think of determining the optimal 'page size' using OS tools.

libbeat/docs/queueconfig.asciidoc
```diff
@@ -11,7 +11,7 @@ options in the `queue` section of the +{beatname_lc}.yml+ config file. Only one
 queue type can be configured.
 
 
-Example configuration:
+This sample configurations configures the memory queue, buffering up to 4096 events:
```
dedemorton: The phrase "configurations configure" sounds a big odd (and should be singular not plural). How about:

This sample configuration sets the memory queue to buffer up to 4096 events.

libbeat/docs/queueconfig.asciidoc
```diff
@@ -76,3 +77,153 @@ will be immediately available for consumption.
...
++{beatname_lc}.yml+ config file:
+
+[float]
+===== `file.path`
+
+The spool file path. The file is created on startup, if it does not exist.
+
+The default value is "${path.data}/spool.dat".
+
+===== `file.permissions`
```
dedemorton: @urso The content looks good, but just noticed that some of the sections are missing `[float]` tags starting with `file.permissions`.

Issue #5419, opened 2017-10-23
Document backoff setting


urso:
Output Backoff parameters have become configurable. Add missing settings to docs:

- [ ] Logstash [backoff settings](https://github.com/elastic/beats/blob/master/libbeat/outputs/logstash/config.go#L23)

Note: check all outputs expose backoff settings

dedemorton:
@urso It looks like these settings aren't shown in the reference.yml files. Do we still want to document these? Looks like Logstash, Elasticsearch, and Kafka all have backoff settings. I can help add the content to the docs, but I don't see the options in the reference.yml.

ruflin:
We should also add them to the reference file if they are missing.

urso:
Checking the code, we have different backoff behavior/config per output.

Elasticsearch/Logstash have `backoff.init` and `backoff.max` settings. Both employ exponential backoff, with `init` (default=1s) being the initial waiting time and max (default=60s) being the max waiting time.

The kafka backoff setting is available regarding the meta-data update only. Setting is `output.kafka.metadata.retry.backoff`. Unfortunately the client doesn't always handle backoff correctly. This needs to be resolved.

Checking the redis output, we don't have any backoff settings in here.

If backoff settings are missing from reference.yml, they should be added as well. Anyways, backoff behavior needs some overhaul, so we can guarantee same settings/semantics for all outputs.

urso:
I created issue: #6791

For now let's add only Elasticsearch and Logstash to the reference configs only.

dedemorton:
I've added the backoff settings for the ES and LS outputs here: https://github.com/elastic/beats/issues/5419.

PR #6903, opened 2018-04-19
Document backoff options

dedemorton:
Adds backup options for LS and ES outputs (closes https://github.com/elastic/beats/issues/5419)

NOTE: I wasn't sure quite where to add the options in the yaml file.

Do we want to say anything about the backoff factor used here? Probably not since it's not configurable?

### DIFF
--- a/libbeat/_meta/config.reference.yml
+++ b/libbeat/_meta/config.reference.yml
@@ -270,7 +270,17 @@ output.elasticsearch:
   # The default is 50.
   #bulk_max_size: 50
 
-  # Configure http request timeout before failing an request to Elasticsearch.
+  # The number of seconds to wait before attempting to connect to Elasticsearch
+  # after an initial connection error. The default is 1s. If additional errors
+  # occur, the wait time is increased exponentially for each try until
+  # backoff.max is met.
+  #backoff.init: 1s
+
+  # The maximum number of seconds to wait before attempting to connect to
+  # Elasticsearch after a connection error. The default is 60s.
+  #backoff.max: 60s
+
+  # Configure http request timeout before failing a request to Elasticsearch.
   #timeout: 90
 
   # Use SSL settings for HTTPS.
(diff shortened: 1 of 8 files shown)


dedemorton:
@urso I hope I got the details right this time. Right now we're ordering the options by how commonly they are used (with related options appearing next to each other). Not sure that's the best approach, but for now, I've added the options to an arbitrary location because I'm not sure how common it is for users to turn these knobs.

libbeat/_meta/config.reference.yml
```diff
@@ -256,7 +256,17 @@ output.elasticsearch:
...
 
-  # Configure http request timeout before failing an request to Elasticsearch.
+  # The number of seconds to wait before attempting to connect to Elasticsearch
+  # after an initial connection error. The default is 1s. If additional errors
+  # occur, the wait time is increased exponentially for each try until
+  # backoff.max is met.
+  #backoff.init: 1s
+
+  # The maximum number of seconds to wait before attempting to connect to
+  # Elasticsearch after a connection error. The default is 60s.
```
urso: Some more clarification:

`backoff.init` is applied on an network error. If connection was active, but connection gets lost or some other protocol/network error occurs, beats will disconnect and use `backoff.init`. After `backoff.init`, beats try to reconnect. If the reconnect fails, the backoff is increased exponentially up to `backoff.max`. After successful reconnect, the backoff timer is reset. On next network error, backoff will start with again with `backoff.init`.

urso: @dedemorton SGTM

dedemorton:
@urso This PR is ready for a final review.

dedemorton:
@urso I had to rebase and then run `make update` again. I guess that's a good general best practice...that you should rebase a stale branch before running `make update` to pick up doc changes (even if GitHub isn't asking you to rebase).

PR #6907, opened 2018-04-20
Harden file integrity flaky test

adriansr:
The file integrity system test depends on auditbeat's file output to validate functionality. This output is sometimes too slow and causes random CI failures.

This patch makes sure that all the required events are written to the file output before stopping auditbeat.

---

I'm leaving this here so it can be merged if we see any flakiness in the file_integrity test. I saw the original test fail under Linux when run enough times in a loop, always with `Dir '{0}' not found`.

Related to #6893


ruflin:
This will need `make fmt` to based the intake job.

ruflin:
jenkins, test this

adriansr:
@ruflin it can be merged

PR #6908, opened 2018-04-20
Unify default module configurations

exekias:
Since we split module configurations into modules.d folder, we can keep
configurations more verbose, so users have a quick reference when
opening them for editing.

This change unifies configuration files using the followin rules:

 * Write at least the following settings, in this order: `module`,
 `metricsets`, `period`, `hosts`.
 * Default metricsets list shown as it helps to know what's the list,
 or disable just one. Inline if there is only one metricset, one line
 per metricset if there is more than one.
 * Some modules have exceptions and include more definitions, or different
 defaults.
 * Comments are allowed (and encouraged), both to explain a setting and
 to suggest a full configuration.
 * `enabled` setting is not shown at all, modules are enabled by using
 `metricbeat modules enable <module name>`. Use comments to suggest
 settings that are disabled by default.

### DIFF
--- a/metricbeat/Makefile
+++ b/metricbeat/Makefile
@@ -10,6 +10,7 @@ ES_BEATS?=..
 GOX_OS=netbsd linux windows
 GOX_FLAGS=-arch="amd64 386 arm ppc64 ppc64le"
 
+DOCS_BRANCH=$(shell grep doc-branch ../libbeat/docs/version.asciidoc | cut -c 14-)
 
 include ${ES_BEATS}/libbeat/scripts/Makefile
 
@@ -41,8 +42,8 @@ configs: python-env
 	@cat ${ES_BEATS}/metricbeat/_meta/setup.yml >> _meta/beat.yml
 	@cat ${ES_BEATS}/metricbeat/_meta/common.reference.yml > _meta/beat.reference.yml
 	@${PYTHON_ENV}/bin/python ${ES_BEATS}/script/config_collector.py --beat ${BEAT_NAME} --full $(PWD) >> _meta/beat.reference.yml
-	@rm -rf modules.d && mkdir -p modules.d
-	@for MODULE in `find module -maxdepth 1 -mindepth 1 -type d -exec basename {} \;`; do cp -a $(PWD)/module/$$MODULE/_meta/config.yml modules.d/$$MODULE.yml.disabled; done
+	@rm -rf modules.d
+	${PYTHON_ENV}/bin/python ${ES_BEATS}/metricbeat/scripts/modules_collector.py --docs_branch=$(DOCS_BRANCH)
 	@chmod go-w modules.d/*
 	@# Enable system by default:
 	@if [ -f modules.d/system.yml.disabled ]; then mv modules.d/system.yml.disabled modules.d/system.yml; fi
--- a/metricbeat/docs/modules/golang.asciidoc
+++ b/metricbeat/docs/modules/golang.asciidoc
@@ -20,9 +20,9 @@ in <<configuration-metricbeat>>. Here is an example configuration:
 ----
 metricbeat.modules:
 - module: golang
-  metricsets:
-    - expvar
-    - heap
+#  metricsets:
+#    - expvar
+#    - heap
   period: 10s
   hosts: ["localhost:6060"]
   heap.path: "/debug/vars"
--- a/metricbeat/docs/modules/golang.asciidoc
+++ b/metricbeat/docs/modules/golang.asciidoc
@@ -20,9 +20,9 @@ in <<configuration-metricbeat>>. Here is an example configuration:
 ----
 metricbeat.modules:
 - module: golang
-#  metricsets:
-#    - expvar
-#    - heap
+  #metricsets:
+  #  - expvar
+  #  - heap
   period: 10s
   hosts: ["localhost:6060"]
   heap.path: "/debug/vars"
--- a/metricbeat/docs/modules/golang.asciidoc
+++ b/metricbeat/docs/modules/golang.asciidoc
@@ -20,7 +20,9 @@ in <<configuration-metricbeat>>. Here is an example configuration:
 ----
 metricbeat.modules:
 - module: golang
-  metricsets: ["expvar","heap"]
+  metricsets:
+    - expvar
+    - heap
   period: 10s
   hosts: ["localhost:6060"]
   heap.path: "/debug/vars"
--- a/metricbeat/docs/modules/jolokia.asciidoc
+++ b/metricbeat/docs/modules/jolokia.asciidoc
@@ -19,29 +19,29 @@ in <<configuration-metricbeat>>. Here is an example configuration:
 ----
 metricbeat.modules:
 - module: jolokia
-#  metricsets: ["jmx"]
+  #metricsets: ["jmx"]
   period: 10s
   hosts: ["localhost"]
   namespace: "metrics"
-#  path: "/jolokia/?ignoreErrors=true&canonicalNaming=false"
+  #path: "/jolokia/?ignoreErrors=true&canonicalNaming=false"
   jmx.mappings:
-    # - mbean: 'java.lang:type=Runtime'
-    #   attributes:
-    #     - attr: Uptime
-    #       field: uptime
-    # - mbean: 'java.lang:type=Memory'
-    #   attributes:
-    #     - attr: HeapMemoryUsage
-    #       field: memory.heap_usage
-    #     - attr: NonHeapMemoryUsage
-    #       field: memory.non_heap_usage
+    #- mbean: 'java.lang:type=Runtime'
+    #  attributes:
+    #    - attr: Uptime
+    #      field: uptime
+    #- mbean: 'java.lang:type=Memory'
+    #  attributes:
+    #    - attr: HeapMemoryUsage
+    #      field: memory.heap_usage
+    #    - attr: NonHeapMemoryUsage
+    #      field: memory.non_heap_usage
     # GC Metrics - this depends on what is available on your JVM
-    # - mbean: 'java.lang:type=GarbageCollector,name=ConcurrentMarkSweep'
-    #   attributes:
-    #     - attr: CollectionTime
-    #       field: gc.cms_collection_time
-    #     - attr: CollectionCount
-    #       field: gc.cms_collection_count
+    #- mbean: 'java.lang:type=GarbageCollector,name=ConcurrentMarkSweep'
+    #  attributes:
+    #    - attr: CollectionTime
+    #      field: gc.cms_collection_time
+    #    - attr: CollectionCount
+    #      field: gc.cms_collection_count
 
   jmx.application:
   jmx.instance:
--- a/metricbeat/docs/modules/jolokia.asciidoc
+++ b/metricbeat/docs/modules/jolokia.asciidoc
@@ -19,11 +19,11 @@ in <<configuration-metricbeat>>. Here is an example configuration:
 ----
 metricbeat.modules:
 - module: jolokia
-  metricsets: ["jmx"]
+#  metricsets: ["jmx"]
   period: 10s
   hosts: ["localhost"]
   namespace: "metrics"
-  #path: "/jolokia/?ignoreErrors=true&canonicalNaming=false"
+#  path: "/jolokia/?ignoreErrors=true&canonicalNaming=false"
   jmx.mappings:
     # - mbean: 'java.lang:type=Runtime'
     #   attributes:
--- a/metricbeat/docs/modules/jolokia.asciidoc
+++ b/metricbeat/docs/modules/jolokia.asciidoc
@@ -23,17 +23,18 @@ metricbeat.modules:
   period: 10s
   hosts: ["localhost"]
   namespace: "metrics"
+  #path: "/jolokia/?ignoreErrors=true&canonicalNaming=false"
   jmx.mappings:
-    - mbean: 'java.lang:type=Runtime'
-      attributes:
-        - attr: Uptime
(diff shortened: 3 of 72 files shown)


exekias:
Changes are duplicated so there is only need to review 32 files under `module/*/_meta/config.yml`, `modules.d` is a generated copy from there

metricbeat/module/system/_meta/config.yml
```diff
@@ -1,14 +1,20 @@
...
+    - cpu
+    - load
+    - memory
+    - network
+    - process
+    - process_summary
+    #- core
+    #- diskio
+    #- socket
+  processes: ['.*']
```
jsoriano: This is the default, should we leave it commented out?

metricbeat/module/aerospike/_meta/config.yml
```diff
@@ -1,2 +1,4 @@
 - module: aerospike
+  metricsets: ["namespace"]
```
ruflin: The reason we left out the metricsets in the short config is to make sure we always have the default behaviour an it's not influence by the config. So if we add one more metricset to the defaults, a user will get it automatically without having to change the config.

metricbeat/module/aerospike/_meta/config.yml
```diff
@@ -1,2 +1,4 @@
...
+  period: 10s
```
ruflin: I remember @andrewkroh suggested to remove period. The part I like about it is that it makes the config even shorter.

metricbeat/module/ceph/_meta/config.yml
```diff
@@ -1,7 +1,16 @@
 - module: ceph
-  metricsets: ["cluster_health", "cluster_status", "monitor_health"]
+  metricsets:
+    - cluster_health
```
ruflin: +1 on the new structure here, makes it easier to comment one.

metricbeat/module/ceph/_meta/config.yml
```diff
@@ -1,7 +1,16 @@
...
+    - cluster_status
+    - monitor_health
   period: 10s
+  hosts: ["localhost:5000"]
```
ruflin: oh, seems like we forgot hosts here :-(

metricbeat/module/ceph/_meta/config.yml
```diff
@@ -1,7 +1,16 @@
...
 
 - module: ceph
-  metricsets: ["cluster_disk", "osd_tree", "pool_disk"]
+  metricsets:
+    - cluster_disk
+    - osd_tree
+    - osd_df
+    - pool_disk
   period: 1m
```
ruflin: this is a good example I think where having period in makes sense.

metricbeat/module/windows/_meta/config.yml
```diff
@@ -1,2 +1,8 @@
 - module: windows
-  period: 60s
+  metricsets: ["perfmon"]
```
ruflin: This is not enabled by default so it should not be here.

metricbeat/module/aerospike/_meta/config.yml
andrewkroh: One option would be to only include the period when it's not the default of 10s.

metricbeat/modules.d/system.yml
```diff
@@ -1,14 +1,20 @@
...
 
-#- module: system
-#  metricsets:
-#    - core
-#    - diskio
-#    - socket
-
 - module: system
   period: 1m
   metricsets:
```
andrewkroh: I think it might make sense to use the` ignore_types` option instead of the filter. Or maybe these mounts are filtered by default now?

metricbeat/modules.d/windows.yml.disabled
```diff
@@ -1,2 +1,8 @@
 - module: windows
-  period: 60s
+  metricsets: ["perfmon"]
+  period: 10s
+  perfmon.counters:
```
andrewkroh: While it would be a deviance from the defaults, I think that including an example would be very helpful to users. We have some in the [docs](https://www.elastic.co/guide/en/beats/metricbeat/master/metricbeat-metricset-windows-perfmon.html#_configuration_14).

metricbeat/modules.d/system.yml
```diff
@@ -1,14 +1,20 @@
 - module: system
+  period: 10s
+  metricsets:
+    - cpu
+    - load
```
andrewkroh: `load` doesn't exist for Windows. Does `load` get deleted during packaging for Windows (check `before-build` in the Makefile)?

exekias: It shoulds, but it's a good point, I must test this

metricbeat/modules.d/system.yml
exekias: That's a good question, @jsoriano do your recent changes affect this?

metricbeat/module/aerospike/_meta/config.yml
exekias: I see this point, what about having them commented?

metricbeat/module/aerospike/_meta/config.yml
exekias: In general, I would prefer to keep it as it adds visibility to the setting, any user opening the config file for the module will know what that is and how it works really quick. If we remove it they need to go back to the docs.

metricbeat/module/windows/_meta/config.yml
exekias: I commented it out

exekias:
I gave this another spin: included link to docs + commented metricsets

metricbeat/docs/modules/golang.asciidoc
```diff
@@ -20,7 +20,9 @@ in <<configuration-metricbeat>>. Here is an example configuration:
 ----
 metricbeat.modules:
 - module: golang
-  metricsets: ["expvar","heap"]
+#  metricsets:
```
ruflin: Nit: Could you put the # directly before the `metricset` without space? It's more similar to what we do everywhere else.

metricbeat/docs/modules/jolokia.asciidoc
```diff
@@ -19,28 +19,32 @@ in <<configuration-metricbeat>>. Here is an example configuration:
 ----
 metricbeat.modules:
 - module: jolokia
-  metricsets: ["jmx"]
+#  metricsets: ["jmx"]
```
ruflin: Comment above applies to all `#`

metricbeat/module/mysql/_meta/config.yml
```diff
@@ -1,2 +1,17 @@
...
+  # and password config options. Those specified in the DSN take precedence.
+  hosts: ["root:secret@tcp(127.0.0.1:3306)/"]
+
+  # Username of hosts. Empty by default.
+  #username: root
+
+  # Password of hosts. Empty by default.
+  #password: secret
+
+  # By setting raw to true, all raw fields from the status metricset will be added to the event.
```
ruflin: Let's remove this from here as I don't want to encourage people to use it.

metricbeat/module/mysql/_meta/config.yml
```diff
@@ -1,2 +1,17 @@
 - module: mysql
-  hosts: ["tcp(127.0.0.1:3306)/"]
+#  metricsets: ["status"]
```
ruflin: Could we use here also the `-` notation already?

metricbeat/module/php_fpm/_meta/config.yml
```diff
@@ -1,2 +1,5 @@
 - module: php_fpm
+#  metricsets: ["pool"]
+  period: 10s
   hosts: ["localhost:8080"]
+  status_path: "/status"
```
ruflin: Is this a default? If yes, it should be commented out.

metricbeat/module/redis/_meta/config.yml
```diff
@@ -1,2 +1,31 @@
 - module: redis
+#  metricsets:
+#    - info
+#    - keyspace
+  period: 10s
+
+  # Redis hosts
   hosts: ["127.0.0.1:6379"]
+
+  # Timeout after which time a metricset should return an error
```
ruflin: I would remove this as I don't expect most users to use it.

metricbeat/module/redis/_meta/config.yml
```diff
@@ -1,2 +1,31 @@
...
+  # Timeout is by default defined as period, as a fetch of a metricset
+  # should never take longer then period, as otherwise calls can pile up.
+  #timeout: 1s
+
+  # Optional fields to be added to each event
```
ruflin: This is not specific to redis, lets remove it.

metricbeat/module/redis/_meta/config.yml
```diff
@@ -1,2 +1,31 @@
...
+  #fields:
+  #  datacenter: west
+
+  # Network type to be used for redis connection. Default: tcp
+  #network: tcp
+
+  # Max number of concurrent connections. Default: 10
+  #maxconn: 10
+
+  # Filters can be used to reduce the number of fields sent.
```
ruflin: This is not redis specific, lets remove it.

metricbeat/scripts/modules_collector.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+def collect(docs_branch):
+
+    base_dir = "module"
+    path = os.path.abspath("module")
+
+    # TODO add module release status if beta or experimental
+    header = """# Module: {module}
+# Docs: https://www.elastic.co/guide/en/beats/metricbeat/{docs_branch}/metricbeat-module-{module}.html
```
ruflin: Nice

metricbeat/scripts/modules_collector.py
```diff
@@ -0,0 +1,48 @@
...
+
+"""
+
+    # Create directory for module confs
```
ruflin: Could we have the docs link also in the reference config file?

metricbeat/Makefile
```diff
@@ -10,6 +10,7 @@ ES_BEATS?=..
 GOX_OS=netbsd linux windows
 GOX_FLAGS=-arch="amd64 386 arm ppc64 ppc64le"
 
+DOCS_BRANCH=$(shell grep doc-branch ../libbeat/docs/version.asciidoc | cut -c 14-)
```
ruflin: LGTM for now but just want to mention that this pretty fragile as we found out other cases.

metricbeat/scripts/modules_collector.py
ruflin: As this is the module collection only for metricbeat, why did it change the other config files?

exekias: I was talking about reference file, which is being collected by a different script

metricbeat/metricbeat.reference.yml
```diff
@@ -251,7 +253,8 @@ metricbeat.modules:
 
 #-------------------------------- HTTP Module --------------------------------
 - module: http
-  metricsets: ["json"]
+  #metricsets:
```
ruflin: Should we in the reference file have the `enabled` one commented out to make it more clear? I like it in the short config to make sure the defaults keep working but here it's more about documenting it?

metricbeat/module/aerospike/_meta/config.yml
```diff
@@ -1,2 +1,4 @@
 - module: aerospike
+#  metricsets: ["namespace"]
```
ruflin: # should be before metricsets directly for consistency.

metricbeat/module/postgresql/_meta/config.yml
```diff
@@ -1,2 +1,21 @@
 - module: postgresql
+  #metricsets:
+  #  # Stats about every PostgreSQL database
```
ruflin: I would remove the comments for the metricsets. For these details we have the docs.

metricbeat/module/rabbitmq/_meta/config.yml
```diff
@@ -1,2 +1,10 @@
 - module: rabbitmq
+  #metricsets:
+  #  - node
+  #  - queue
+  #  - connection
+  period: 10s
   hosts: ["localhost:15672"]
+
+  username: guest
```
ruflin: Should this be commented out or is `guest` a common default?

jsoriano: Yes, guest exists by default.

ruflin: So it's better to connect by default with `guest/guest` the have no credentials?

jsoriano: `guest`/`guest` can be seen as some kind of "no credentials" for rabbitmq :)
By default it always exists, and I don't know if it is possible to connect with rabbitmq without credentials. I think it'd be good if metricbeat tries to connect with these credentials by default too, so it can collect some metrics out of the box.

metricbeat/modules.d/system.yml
jsoriano: Oh sorry, I forgot this question. My latest changes filter by default the filesystem types marked as `nodev` in `/proc/filesystems`, and remove duplications caused by bind mounts and similar. Probably this is enough to have meaningful events by default without needing these processors.

metricbeat/module/rabbitmq/_meta/config.yml
ruflin: SGTM. Not part of this PR but I wonder if we should in these case then also set them directly in the code as defaults.

jsoriano: Yes, good point, I'll take care of this.

PR #6913, opened 2018-04-20
Exclude common compression formats

djs:
Logrotate may be configured to use various compression formats other than
gzip, so exclude the most common.

### DIFF
--- a/filebeat/module/nginx/access/config/nginx-access.yml
+++ b/filebeat/module/nginx/access/config/nginx-access.yml
@@ -3,4 +3,4 @@ paths:
 {{ range $i, $path := .paths }}
  - {{$path}}
 {{ end }}
-exclude_files: [".gz$"]
+exclude_files: [".gz$", ".bz2$", ".xz$"]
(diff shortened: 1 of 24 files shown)


djs:
Yikes, I don't know how that happened, but I pushed a new version with commas this time.

filebeat/module/nginx/access/config/nginx-access.yml
```diff
@@ -3,4 +3,4 @@ paths:
 {{ range $i, $path := .paths }}
  - {{$path}}
 {{ end }}
-exclude_files: [".gz$"]
+exclude_files: [".gz$". ".bz2$", ".xz$"]
```
jsoriano: @djs there is a dot still trying to escape here :slightly_smiling_face:

jsoriano: I think that only this one is pending

djs:
Still working on this. I'm having trouble with my environment to test this locally. I mistakenly thought it would be an easy change.

faec:
It looks like this PR has stalled and fallen out of date. If there's interest in reviving it please let us know, otherwise it will be auto-closed in two weeks. Thanks!

PR #6916, opened 2018-04-23
Add `GetProcessedMetrics` method to Prometheus helper

exekias:
This PR adds a new `GetProcessedMetrics` method to ease the creation of metricsets based on Prometheus. It moves from code (current approach) to a mapping template for the conversion of Prometheus metrics to our format, better suited for Elasticsearch mappings.

With this new method you only need to fill the `MetricsMapping` structure to map all metrics to events. There are a couple of different use cases to support different patterns in Prometheus metrics, like a way to extract labels into `keyworud` values, or doing boolean conversion.

This PR is divided into 2 commits: 
1. Introducing the method
2. Adapt existing `state_*` metricsets to make use of it

I plan to use this for incoming modules/metricsets.

:warning: There is a breaking change affecting the `state_container` metricset (currently beta):

- Fixed a typo in `status.phase` value: `terminate` -> `terminated`

Also this deprecates calculated metric `cpu.*.nanocores` in favor of `cpu.*.cores`.


metricbeat/helper/prometheus/label.go
```diff
@@ -0,0 +1,42 @@
...
+
+// Label maps a Prometheus label to a Metricbeat field
+func Label(field string) LabelMap {
+	return &commonLabel{
+		field: field,
+		key:   false,
+	}
+}
+
+// KeyLabel maps a Prometheus label to a Metricbeat field. The lable is flagged as key.
```
jsoriano: Typo `lable`

metricbeat/module/kubernetes/state_container/_meta/fields.yml
```diff
@@ -26,14 +26,14 @@
     - name: cpu
       type: group
       fields:
-        - name: limit.nanocores
+        - name: limit.cores
```
ruflin: As `state_container` is already GA we should not do breaking changes. Could we perhaps keep both fields but mark one as deprecated? Same below.

metricbeat/helper/prometheus/prometheus.go
```diff
@@ -0,0 +1,165 @@
...
+	Metrics map[string]MetricMap
+
+	// Labels translate from prometheus label names to Metricbeat fields
+	Labels map[string]LabelMap
+
+	// ExtraFields adds the given fields to all events coming from `GetProcessedMetrics`
+	ExtraFields map[string]string
+}
+
+func (p *prometheus) GetProcessedMetrics(mapping *MetricsMapping) ([]common.MapStr, error) {
```
ruflin: It reminds me a lot of our schema package.

metricbeat/helper/prometheus/prometheus.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	eventsMap := map[string]common.MapStr{}
+	for _, family := range families {
+		for _, metric := range family.GetMetric() {
+			m, ok := mapping.Metrics[family.GetName()]
+
+			// Ignore unknown metrics
+			if !ok {
+				continue
```
ruflin: We should sync up the behaviour with the schema package on what we do on error / missing fields. I'm not happy with the current error handling in schema. Let's have a follow up chat on how we should push that forward.

CHANGELOG.asciidoc
```diff
@@ -39,6 +39,8 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
 - Add config option for windows/perfmon metricset to ignore non existent counters. {pull}6432[6432]
 - Refactor docker CPU calculations to be more consistent with `docker stats`. {pull}6608[6608]
 - Update logstash.node_stats metricset to write data under `logstash.node.stats.*`. {pull}6714[6714]
+- Changed `state_container` metric from `cpu.*.nanocores` to `cpu.*.cores`.
+- Fixed typo in values for `state_container` `status.phase`, from `terminate` to `terminated`.
```
ruflin: Here we could perhaps use `copy_to` to copy the field for now?

metricbeat/helper/prometheus/prometheus.go
exekias: Definitely inspired on it :)

metricbeat/helper/prometheus/metric.go
```diff
@@ -0,0 +1,126 @@
...
+type MetricMap interface {
+	// GetField returns the resulting field name
+	GetField() string
+
+	// GetValue returns the resulting value
+	GetValue(m *dto.Metric) interface{}
+}
+
+// Metric directly maps a Prometheus metric to a Metricbeat field
+func Metric(field string) MetricMap {
```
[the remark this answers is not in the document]
exekias: Uhm that's a good question, so far I think that mapping should take care of that?

metricbeat/module/kubernetes/state_container/_meta/fields.yml
exekias: I've pushed a change keeping them (deprecated)

CHANGELOG.asciidoc
exekias: I used code as there is a change on the scale, so needed to perform the math

metricbeat/helper/prometheus/metric.go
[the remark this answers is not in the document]
ruflin: It indeed does for the define fields. Coming from the `schema` package I kind of expected to have both.

PR #6921, opened 2018-04-23
Ensure the dashboard zip is sane

tsg:
This adds a check that all files from the dashboard zip file
are pointing to the right target, and don't override other configs.

### DIFF
--- a/libbeat/dashboards/importer.go
+++ b/libbeat/dashboards/importer.go
@@ -126,7 +126,7 @@ func (imp Importer) unzip(archive, target string) error {
 			return err
 		}
 		if strings.HasPrefix(filepath.ToSlash(relPath), "../") {
-			return fmt.Errorf("Zip file contains files outside of directory target: %s", relPath)
+			return fmt.Errorf("Zip file contains files outside of the target directory: %s", relPath)
 		}
 
 		if file.FileInfo().IsDir() {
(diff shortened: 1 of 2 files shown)


libbeat/dashboards/importer.go
```diff
@@ -119,6 +119,16 @@ func (imp Importer) unzip(archive, target string) error {
...
 		filePath := filepath.Join(target, file.Name)
 
+		// check that the resulting file path is indeed under target
+		// Note that Rel calls Clean.
+		relPath, err := filepath.Rel(target, filePath)
+		if err != nil {
+			return err
+		}
+		if strings.HasPrefix(filepath.ToSlash(relPath), "../") {
+			return fmt.Errorf("Zip file contains files outside of directory target: %s", relPath)
```
urso: s/directory target/target directory/ ?

hansmi:
If I may I'd recommend using [`cyphar/filepath-securejoin`](https://github.com/cyphar/filepath-securejoin) rather than coming up with your own path verification.

tsg:
Thank you for the suggestion @hansmi. We had some concerns about the Windows portability of the filepath-securejoin lib, because it relies on the PathSeparator. So for now we merged the PR like this to catch the release train, but we'll follow up on this. cc @urso

PR #6931, opened 2018-04-24
Fix self metrix when containerised (backport to 6.2)

jsoriano:
Backport of #6641 to 6.2


ruflin:
jenkins, test it

ruflin:
@jsoriano Can you confirm that CI failures are not related just to double check?

In general I'm not sure if perhaps the download URL changes for Kibana for example and is why the build is breaking on Travis.

PR #6934, opened 2018-04-24
Metricbeat: Dashboards and visualizations for haproxy

jsoriano:
Add some dashboards and visualizations for haproxy.

I think it is a bit tricky to have generic dashboards for haproxy, there are too many ways of configuring frontends, backends and server lists. This collection of dashboards and visualizations can serve as a good base so users can create their own dashboards for their specific use cases.

Overview that acts as an index, each entry in the three columns is a parameterized link to one of the following views.
haproxy-overview

Dashboard for an HTTP frontend
haproxy-frontend

Dashboard for an HTTP backend
haproxy-backend

Dashboard for an HTTP server
haproxy-server

A couple of dashboards are also added for TCP frontends and backends, they are similar to the HTTP ones, but with less things.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -201,6 +201,7 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
 - Ignore virtual filesystem types by default in system module. {pull}6819[6819]
 - Release config reloading feature as GA. {pull}6891[6891]
 - Add experimental Elasticsearch index metricset. {pull}6881[6881]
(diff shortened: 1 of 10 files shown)


ruflin:
Nice dashboards.

The Jenkins LS failure seems to be unrelated but seems like for whatever reasons the dashboard loading tests are failing: https://travis-ci.org/elastic/beats/jobs/370730035#L4268

jsoriano:
I had exported the dashboards from kibana directly, I have done it now with the export tool.

Issue #5512, opened 2017-11-03
Update metricbeat getting started to show how to use modules.d layout


dedemorton:
Metricbeat getting started should mention how to use the modules.d layout to enable and disable modules. Right now, it still tell users to update the metricbeat.yml file.

Issue #6196, opened 2018-01-26
Windows instructions vary


roncohen:
Sometimes there's a `.\` sometimes there isn't. In our tests in PS, it seemed that the `.\` was necessary.

![image](https://user-images.githubusercontent.com/744/35450779-507837f2-02c1-11e8-8649-1fa486c28126.png)

andrewkroh:
+1  The `.\` is needed in these cases.

We don't instruct the user to put our binaries on their PATH.

ruflin:
@dedemorton As #6941 is merged, can this one be closed?

dedemorton:
Yes. Closing....

PR #6941, opened 2018-04-25
Add modules.d layout to getting started and fix Windows command examples

dedemorton:
Fixes https://github.com/elastic/beats/issues/5512 and https://github.com/elastic/beats/issues/6196

I've also changed the metricbeat getting started to use asciidoc attributes to resolve "metricbeat" and "Metricbeat". My goal is to convert all the topics over to using attributes, but I don't want to open a big PR for those changes because it's easier for me to check for mistakes if I make the changes in smaller batches. 

Question for reviewers:

One thing that's missing here is advice on enabling/disabling metricsets. Do we want to mention that step in the getting started, or leave users to read the configuration reference docs if they want to change the defaults?

### DIFF
--- a/metricbeat/docs/gettingstarted.asciidoc
+++ b/metricbeat/docs/gettingstarted.asciidoc
@@ -176,6 +176,8 @@ You can either enable the default module configurations defined in the
 +{beatname_lc}.yml+ file. The `modules.d` directory contains default
 configurations for all available {beatname_uc} modules. 
 +
+If you are using a Docker image, see <<running-on-docker>>.
++
 The following examples enable the `apache` and `mysql` configs in the
 `modules.d` directory :
 +
@@ -191,13 +193,6 @@ The following examples enable the `apache` and `mysql` configs in the
 ["source","sh",subs="attributes"]
 ----
 ./{beatname_lc} modules enable apache mysql
-----
-+
-*docker*:
-+
-["source","sh",subs="attributes"]
-----
-docker run docker.elastic.co/beats/{beatname_lc}:{version} modules enable apache mysql
 ----
 +
 *win:*
@@ -280,6 +275,8 @@ sudo service {beatname_lc} start
 ----------------------------------------------------------------------
 docker run {dockerimage}
 ----------------------------------------------------------------------
+
+Also see <<running-on-docker>>.
 
 *mac:*
 
--- a/metricbeat/docs/gettingstarted.asciidoc
+++ b/metricbeat/docs/gettingstarted.asciidoc
@@ -1,10 +1,10 @@
-[[metricbeat-getting-started]]
-== Getting started with Metricbeat
+[id="{beatname_lc}-getting-started"]
+== Getting started with {beatname_uc}
 
-Metricbeat helps you monitor your servers and the services they host by
+{beatname_uc} helps you monitor your servers and the services they host by
 collecting metrics from the operating system and services.
 
-To get started with your own Metricbeat setup, install and configure these
+To get started with your own {beatname_uc} setup, install and configure these
 related products:
 
  * Elasticsearch for storage and indexing the data.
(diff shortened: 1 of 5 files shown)


metricbeat/docs/gettingstarted.asciidoc
```diff
@@ -122,100 +122,103 @@ endif::[]
...
+["source","sh",subs="attributes"]
+----
+./{beatname_lc} modules enable apache mysql
+----
++
+*docker*:
++
+["source","sh",subs="attributes"]
+----
+docker run docker.elastic.co/beats/{beatname_lc}:{version} modules enable apache mysql
```
exekias: I think this won't work, as docker doesn't store the container status between runs

exekias: Probably it's tricky, they would need to mount a `modules.d` folder or `metricbeat.yml` from the host,

dedemorton: @exekias Yah, that makes sense. I was trying to add platform-specific commands for all the platforms we currently show in the getting started docs, but I didn't think it through very carefully. Since Docker is a special beast, sounds like I just need to point to: 

https://www.elastic.co/guide/en/beats/metricbeat/current/running-on-docker.html

dedemorton: @exekias One question, tho...should we mention in the command reference that the modules command won't work with Docker images?

dedemorton:
@exekias I've replaced the docker example with a link to the extended topic. I think we need to do more work to improve the docs around running on containers, but that's a separate item on my to-do list. Should someone else (maybe @ruflin) review this topic? Would like to get it merged soon.

metricbeat/docs/gettingstarted.asciidoc
```diff
@@ -122,100 +122,98 @@ endif::[]
...
-  period: 10s
-  processes: ['.*']
-  cpu_ticks: false
--------------------------------------
-
-The following example shows how to configure two modules: the system module
-and the Apache HTTPD module:
-
-[source, shell]
--------------------------------------
```
ruflin: I wonder why your removed this part? I remember I added it because in the early days of Metricbeat on discuss several times the question popped up on how to configure 2 modules or if it's possible to configure the same module twice. I was hoping this example here helps.

ruflin:
Changes LGTM. Left one questions.

metricbeat/docs/gettingstarted.asciidoc
ruflin: I would skip the docker command and only link to the docs as the command alone spins up the container but will not provide the user any useful output.

metricbeat/docs/gettingstarted.asciidoc
dedemorton: @ruflin I removed it because I wanted the getting started guide to focus using the configs in the modules.d directory because using the enable modules command is a better getting started experience. I thought users would look at this topic for config metricbeat.yml config examples: https://www.elastic.co/guide/en/beats/metricbeat/master/configuration-metricbeat.html

Do you want me to add the deleted example to the ones shown in configuration-metricbeat.html, or do you think the example that are there are good enough?

ruflin: @dedemorton You are right, it's all there. Forgot that we also have there some good examples. +1 on removing it here.

metricbeat/docs/gettingstarted.asciidoc
exekias: I'm ok if we skip docker here

dedemorton: @ruflin I pushed another commit that removes the run example (and points to the full topic about running on docker) because it sounds like it's not useful to run that command. If it looks good, feel free to squash and commit.

PR #6947, opened 2018-04-25
Fix race on shared maps in global fields

urso:
On publish fields are added to an event in this order:
- local/global configured fields
- dynamic fields
- "beat" metadata

When merging the fields, shared structures must not be overwritten or
updated concurrently. This is enforced by cloning the original fields
structure before applying updates.

This adds missing Clone operations if configured fields add new
fields to the `beat` namespace or if dynamic fields are enabled.

The dynamic fields support has been changed to dynamically test if a clone is required or not.

### DIFF
--- a/libbeat/publisher/pipeline/processor.go
+++ b/libbeat/publisher/pipeline/processor.go
@@ -53,8 +53,7 @@ func newProcessorPipeline(
 		localProcessors = makeClientProcessors(config)
 	)
 
-	// needsCopy := global.alwaysCopy || localProcessors != nil || global.processors != nil
-	needsCopy := true
+	needsCopy := global.alwaysCopy || localProcessors != nil || global.processors != nil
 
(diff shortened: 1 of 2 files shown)


libbeat/publisher/pipeline/processor.go
```diff
@@ -53,7 +53,8 @@ func newProcessorPipeline(
 		localProcessors = makeClientProcessors(config)
 	)
 
-	needsCopy := global.alwaysCopy || localProcessors != nil || global.processors != nil
+	// needsCopy := global.alwaysCopy || localProcessors != nil || global.processors != nil
+	needsCopy := true
```
tsg: This looks weird. Is the intention to always copy? Or can the `needsCopy` value change to false in other parts of the code?

urso: oh, some leftover from testing. :(

PR #6948, opened 2018-04-25
Update to Golang 1.10.1

ruflin:


exekias:
Can you add `prepare-tests` after `python-env` here? I think it should fix libcompose vendor issue: https://github.com/elastic/beats/blob/master/libbeat/scripts/Makefile#L118

ruflin:
For the generator test failure, the fix happening in https://github.com/elastic/beats/pull/4834 should also help here.

Issue #6616, opened 2018-03-21
[Heartbeat] Crashing on concurrent map write


andrewkroh:
Heartbeat is crashing after ~20 minutes with fatal map errors ("concurrent map writes" or "concurrent map iteration and map write").

For confirmed bugs, please report:
- Version: 6.1.3
- Operating System: RHEL7 x86_64
- Discuss: https://discuss.elastic.co/t/heartbeat-fatal-error-concurrent-map-iteration-and-map-write/124716
- Stack Trace: See Discuss.
- Steps to Reproduce: Using the config below, the user is observing crashes periodically (like every 20 minutes).

```
heartbeat.monitors:

- type: tcp
  hosts: ["s1234", "s1234", "s1234", "s1234", "s1234", "s1234"]
  ports: [5555,9000]
  schedule: '@every 180s'
  timeout: 30s

- type: tcp
  hosts: ["s1234", "s12345"]
  ports: [12503]
  schedule: '@every 180s'
  timeout: 30s

#================================ General =====================================
# The name of the shipper that publishes the network data. It can be used to group
hostname: smonitor

#----------------------------- Logstash output --------------------------------
output.logstash:
  # The Logstash hosts
  hosts: ["localhost:5045"]
```

tsg:
@ruflin FYI, I added this one to the Heartbeat GA checklist.

JoostV:
2018-04-03T15:03:03.282+0200    DEBUG   [publish]       pipeline/processor.go:275       Publish event: {
  "@timestamp": "2018-04-03T13:03:03.269Z",
  "@metadata": {
    "beat": "heartbeat",
    "type": "doc",
    "version": "6.2.3"
  },
  "tcp": {
    "port": 7224,
    "rtt": {
      "connect": {
        "us": 641
      }
    }
  },
  "monitor": {
    "ip": "127.0.0.1",
    "duration": {
      "us": 13544
    },
    "status": "up",
    "id": "tcp-tcp@localhost:[7224 7226 7228 7230]",
    "host": "localhost",
    "scheme": "tcp",
    "name": "tcp",
    "type": "tcp"
  },
  "resolve": {
    "host": "localhost",
    "ip": "127.0.0.1",
    "rtt": {
      "us": 12174
    }
  },
  "type": "monitor",
  "beat": {
    "name": "pdccva.dc.post21.nl",
    "hostname": "pdccva.dc.post21.nl",
    "version": "6.2.3"
  }
}
[...]

PR #6950, opened 2018-04-25
Fix heartbeat races on event updates

urso:
Resolves: #6616

Heartbeat has a many components, each adding fields to an existing
event. As events are created per 'TaskRun', this is normally fine. But
in case the TaskRunner branches of into multiple Sub-tasks, we will see
races on shared event structures. This is the case with multiple ports
in the http/tcp configuration or the IPAll setting (ping all known IPs
of a given domain).

Namespaces in an event can be potentially populated via globally shared
structures. This can also lead to unwanted updates or races.

This fix clones the event structure before continuing event updates, so
to remove the chance of races.

I did run some tests before and after the change, with the race detector enabled. No more race have been found with this change.

### DIFF
--- a/heartbeat/monitors/util.go
+++ b/heartbeat/monitors/util.go
@@ -122,6 +122,8 @@ func annotated(
 		}
 
 		if fields != nil {
+			fields = fields.Clone()
+
 			status := look.Status(err)
 			fields.DeepUpdate(common.MapStr{
 				"monitor": common.MapStr{
@@ -130,7 +132,7 @@ func annotated(
 				},
 			})
 			if user := settings.Fields; user != nil {
-				fields.DeepUpdate(user)
+				fields.DeepUpdate(user.Clone())
 			}
 
 			event.Timestamp = start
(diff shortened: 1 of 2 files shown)


heartbeat/monitors/util.go
```diff
@@ -370,10 +372,13 @@ func resolveErr(host string, err error) (common.MapStr, []TaskRunner, error) {
 func WithFields(fields common.MapStr, r TaskRunner) TaskRunner {
 	return MakeCont(func() (common.MapStr, []TaskRunner, error) {
 		event, cont, err := r.Run()
-		if event == nil {
+		if event != nil {
+			event = event.Clone()
+			event.DeepUpdate(fields)
+		} else if err != nil {
 			event = common.MapStr{}
+			event.DeepUpdate(fields)
```
ruflin: Reading this with the `else if` I asked myself what the `else` case in the end which is not covered here actually means and if it can happen. So event is nil and err is nil?

ruflin:
I added the needs_backport label

ruflin:
@urso Can you rebase this on master. This should fix the LS CI failure.

heartbeat/monitors/util.go
urso: A runner can return an event, but must not. It's fully optional. We have intermediate runners (resolve IP address), that do not return an event, but create a set of continuation runners. So `event` can be nil. 

If a runner returns an error, no followup runners are created. => We need to create an error event. On error a runner can optionally return an event. The WithFields runner enforces the creation of an event on error, such that the already known fields (like IP resolve results) an be merged into the final event

The WithFields wrapper adds shared fields only if an event will be returned.

To be more verbose, the conditions is:

```
if err != nil {
  if event == nil {
    event = common.MapStr{}
  }
  event.DeepUpdate(fields)
} else { // no error
  if event != nil {
    event = event.Clone() // we need to copy the event to ensure no error are returned.
    event.DeepUpdate(fields)
  }
}
```

Due to the `else if`, the second else branch is equivalent to `event == nil && err != nil`.

The case `event == nil && err == nil` is perfectly valid -> we just don't report an event.

urso:
@ruflin rebased.

PR #6966, opened 2018-04-27
Remove `-` char from docker-compose project name

exekias:
New docker-compose version ignores these chars, but libcompose is still
using them. To avoid errors we remove it from the name.

master is currently not affected, but this remove ensures this same
issue doesn't bite us in the future


exekias:
jenkins, test this again

ruflin:
logstash node_stats timeout not related.

PR #6967, opened 2018-04-27
Join Docker log lines when they are split because of size

### LINKED ISSUE #6605: filebeats truncates log messages after 16k
```
# filebeat 6.2.2
filebeat.prospectors:
-
  paths: [ "/var/lib/docker/containers/*/*-json.log" ]
  harvester_buffer_size: 65536
  json.message_key: log
  json.keys_under_root: true
  json.add_error_key: true
  fields_under_root: true
  processors:
    - add_docker_metadata: ~

harvester_buffer_size: 65536

output.file:
  path: "/tmp/filebeat"
  filename: filebeat
```

- The docker container outputs a single line in valid json (about 18k)
- The value of the "log" message key is also a single line in valid json
- The message seems to be cut off at about 16k or a bit above (depends if you count the backslashes for escaping)
- A second message gets created with the remaining part of the message including full decoration (docker meta data, additional fields etc)
- Looks like filebeat splits the message into 2 separate ones
- harvester_buffer_size has no effect
- removing the json.* options has no effect

**How to reproduce:**

use my config, start a container and run the following:

```
#!/bin/bash
key="somerandomkey_"
value="somerandomvalue_"
echo -n '{'
for i in $(seq 420); do
  echo -n "\"${key}${i}\":\"${value}${i}\","
done
echo '"lastkey":"end"}'
```

This produces a valid json output. For me it cuts off at "somerand" and a new message continues with "omkey_396".

We also tried the following and it does not fix the issue: (as well as with and without the `harvester_buffer_size` option.)

```
filebeat.prospectors:
- type: docker
  containers.ids:
    - "*"
  processors:
    - add_docker_metadata: ~
```

@paltryeffort https://discuss.elastic.co/t/filebeat-splits-message-after-16k/123718/4

exekias:
Docker `json-file` driver splits lines longer than 16k bytes, this
change adapts the code to detect that situation and join them again to
output a single event.
    
This behavior is enabled by default, it can be disabled by using the new
`combine_partials` flag.
   
Fixes #6605

### DIFF
--- a/filebeat/docs/inputs/input-docker.asciidoc
+++ b/filebeat/docs/inputs/input-docker.asciidoc
@@ -9,7 +9,7 @@
 
 experimental[]
 
-Use the `docker` input to read logs from Docker containers. 
+Use the `docker` input to read logs from Docker containers.
 
 This input searches for container logs under its path, and parse them into
 common message lines, extracting timestamps too. Everything happens before line
@@ -49,12 +49,19 @@ is `/var/lib/docker/containers`.
 Reads from the specified streams only: `all`, `stdout` or `stderr`. The default
 is `all`.
 
+===== `combine_partial`
+
+Enable partial messages joining. Docker `json-file` driver splits log lines larger than 16k bytes,
+end of line (`\n`) is present for common lines in the resulting file, while it's not the for the lines
+that have been split. `combine_partial` joins them back together when enabled. It is enabled by default.
+
 The following input configures {beatname_uc} to read the `stdout` stream from
 all containers under the default Docker containers path:
 
 [source,yaml]
 ----
 - type: docker
+  combine_partial: true
   containers:
     path: "/var/lib/docker/containers"
     stream: "stdout"
--- a/filebeat/harvester/reader/docker_json.go
+++ b/filebeat/harvester/reader/docker_json.go
@@ -16,6 +16,9 @@ type DockerJSON struct {
 	reader Reader
 	// stream filter, `all`, `stderr` or `stdout`
 	stream string
+
+	// join partial lines
+	partial bool
 }
 
 type dockerLog struct {
@@ -31,10 +34,11 @@ type crioLog struct {
 }
 
 // NewDockerJSON creates a new reader renaming a field
-func NewDockerJSON(r Reader, stream string) *DockerJSON {
+func NewDockerJSON(r Reader, stream string, partial bool) *DockerJSON {
 	return &DockerJSON{
-		stream: stream,
-		reader: r,
+		stream:  stream,
+		partial: partial,
+		reader:  r,
 	}
 }
 
@@ -100,6 +104,21 @@ func (p *DockerJSON) Next() (Message, error) {
 
 		if strings.HasPrefix(string(message.Content), "{") {
 			message, err = parseDockerJSONLog(message, &dockerLine)
+			if err != nil {
+				return message, err
+			}
+			// Handle multiline messages, join lines that don't end with \n
+			for p.partial && message.Content[len(message.Content)-1] != byte('\n') {
+				next, err := p.reader.Next()
+				if err != nil {
+					return message, err
+				}
+				next, err = parseDockerJSONLog(next, &dockerLine)
+				if err != nil {
+					return message, err
+				}
+				message.Content = append(message.Content, next.Content...)
+			}
 		} else {
 			message, err = parseCRILog(message, &crioLine)
 		}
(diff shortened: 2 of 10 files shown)


filebeat/harvester/reader/docker_json.go
```diff
@@ -100,6 +100,21 @@ func (p *DockerJSON) Next() (Message, error) {
 
 		if strings.HasPrefix(string(message.Content), "{") {
 			message, err = parseDockerJSONLog(message, &dockerLine)
+			if err != nil {
+				return message, err
+			}
+			// Handle multiline messages, join lines that don't end with \n
+			for message.Content[len(message.Content)-1] != byte('\n') {
```
ruflin: Should we enable this first through a config option and have it as default later on? This would allow us to start playing around with it.

@exekias I'm glad we have the docker prospector ;-)

exekias: That's a great idea, I'll add it

filebeat/docs/inputs/input-docker.asciidoc
```diff
@@ -49,6 +49,11 @@ is `/var/lib/docker/containers`.
 Reads from the specified streams only: `all`, `stdout` or `stderr`. The default
 is `all`.
 
+===== `containers.partial`
+
+Enable partial messages joining. Docker splits log lines larger than 16k bytes, enable this
+to join them while reading (default: false).
```
andrewkroh: I think it would be useful to specify that it works by joining lines that don't end with \n.

ruflin:
Code LGTM. For the config I was thinking to put it outside `container` and directly on the prospector level. Otherwise it gives the impression it's a container specific setting but I don't think it's really related to containers itself.

Naming wise I was thinking if we could find a name that is more directed into what it does instead of what problem it solves. So instead of `partial` more like `combine_partial_messages` which is too long but I hope describes what i mean. Any suggestions?

exekias:
:+1: How about `combine_partials`? I'm not too worried as this will be a rather exotic flag (once we set it to true by default)

filebeat/docs/inputs/input-docker.asciidoc
exekias: uhm that's something specific to the `json-file` format, it's not the user who adds the `\n`. Do you think it's worth it? If yes, I would explain the whole story about how the format works

ruflin:
👍 on combine_partials. I just realised the the docker input is still in experimental, so we could already turn it on by default and see what happens. Still good to have a config option so in case things go wrong, we can turn it off.

exekias:
I've renamed the flag to `combine_partials`, set it to **true** by default and updated docs

exekias:
done!

bitsofinfo:
nice!

SharpEdgeMarshall:
When should be available? version 6.4?

exekias:
That's correct, it should be available by 6.4

towolf:
@exekias This was _not_ released with 6.3.0 today right?

bitsofinfo:
?! hope it was.... really need this

bitsofinfo:
Whats the eta on 6.4?

exekias:
I'm afraid this didn't make it to 6.3.0, it should be available with `6.4.0`. We can consider a backport to `6.3.1`if we consider this a bugfix, any opinion here @ruflin ?

ruflin:
As Carlos mentioned we normally don't backport any features and I don't think we can qualify it as a bug. The biggest worry with backporting is that it could break something that works now which can always happen. I personally would prefer to not backport but make sure it is as stable as possible in 6.4.

towolf:
So the cadence is roughly quarterly and we can expect 6.4 in 3 months or so?

ruflin:
We don't give an estimates on the release dates ;-) If you want to test / use the feature we have some snapshots available: https://beats-package-snapshots.s3.amazonaws.com/index.html?prefix=filebeat/

bitsofinfo:
so 3 more months?

buddyledungarees:
Is it possible to use existing multiline match parameters to replicate this functionality?

```
multiline.pattern: '\n$'
multiline.negate: true
multiline.match: after
````

**Comment by wallies** [comment]
We are running a fork of this fix, but in the for loop, we are noticing that it doesnt catch EOF of files and will sometimes exit half way through processing a long line. This means that the currently-processed partial log line will not have been completely processed. *Later, when the next part of the now-half-processed log line arrives in the tailed log file, filebeat will fail to parse it.* . Specifically, the following check will fail:
```
if strings.HasPrefix(string(message.Content), "{") {
```

Below is the parse error we see
```
ERROR`	log/harvester.go:243	Read line error: parsing CRI timestamp: parsing time ",\\\"protocol\\\":\\\" as "2006-01-02T15:04:05Z07:00": cannot parse 
```

We have added a retry if io.EOF is reached for now, but im wondering whether the backoff code, that ive seen in other parts of the code, is getting into this bit of the code, or whether it should.

exekias:
Hi @wallies, thank you for your feedback, and thank you for testing unreleased features! Could you please open a new issue for this one (or pull request :wink:)? happy to discuss it there and try to fix it for 6.4

towolf:
@ruflin it's been 3 months now. Can't you just push this out?

ruflin:
@towolf It will be released with the rest of the stack in 6.4 hopefully pretty soon.

PR #6971, opened 2018-04-28
Rabbitmq node only reports metrics of itself by default

### LINKED ISSUE #6556: [Metricbeat] RabbitMQ cluster - node metrics reported multiple times (duplicates)
For a RabbitMQ cluster, metricbeat sends 3 beats for each node instead of just one thus triplicating the number of metrics which is unnecessary.

The issue in details is described [here](https://discuss.elastic.co/t/rabbitmq-cluster-node-metrics-reported-multiple-times/123592)

- Version: 6.2.2
- Operating System: Linux Ubuntu 16.04
- Steps to Reproduce: Install metricbeat on all nodes of RabbitMQ cluster and check the records in Elasticsearch.

jsoriano:
A configuration option is added, `node.collect` that can be set to
`cluster` to have the previous behaviour. The default is `node`.

Fixes #6556

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -43,6 +43,7 @@ https://github.com/elastic/beats/compare/v6.2.3...master[Check the HEAD diff]
 - Update logstash.node_stats metricset to write data under `logstash.node.stats.*`. {pull}6714[6714]
 - Fixed typo in values for `state_container` `status.phase`, from `terminate` to `terminated`. {pull}6916[6916]
 - RabbitMQ management plugin path is now configured at the module level instead of having to do it in each of the metricsets. New `management_path_prefix` option should be used now {pull}7074[7074]
+- RabbitMQ node metricset only collects metrics of the instance it connects to, `node.collect: cluster` can be used to collect all nodes as before. {issue}6556[6556] {pull}6971[6971]
 
 *Packetbeat*
 
--- a/metricbeat/module/rabbitmq/node/_meta/docs.asciidoc
+++ b/metricbeat/module/rabbitmq/node/_meta/docs.asciidoc
@@ -1,1 +1,10 @@
-This is the `node` metricset of the RabbitMQ module.
+This is the `node` metricset of the RabbitMQ module and collects metrics
+about RabbitMQ nodes.
+
+The metricset has two modes to collect data which can be selected with the
+`node.collect` setting:
+
+* `node`: collects metrics only from the node `metricbeat` connects to. This is the
+  default, as it is recommended to deploy `metricbeat` in all nodes.
+* `cluster`: collects metrics from all the nodes in the cluster. This is recommended
+  when collecting metrics of an only endpoint for the whole cluster.
--- a/metricbeat/module/rabbitmq/node/data.go
+++ b/metricbeat/module/rabbitmq/node/data.go
@@ -3,10 +3,9 @@ package node
 import (
 	"encoding/json"
 
-	"github.com/elastic/beats/libbeat/common"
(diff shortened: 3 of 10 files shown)


metricbeat/module/rabbitmq/node/node.go
```diff
@@ -40,20 +50,61 @@ func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
...
+		BaseMetricSet: base,
+		HTTP:          http,
+		nodesURL:      nodesURL,
+		overview:      overview,
 	}, nil
 }
 
-func (m *MetricSet) Fetch() ([]common.MapStr, error) {
-	content, err := m.HTTP.FetchContent()
+type apiOverview struct {
```
ruflin: We normally keep all the mapping code in `data.json`. But happy to change this convention :-)

jsoriano:
Ok, agree to allow the user to still collect metrics from all the nodes.

jsoriano:
Added `node.mode` config flag to choose between `self` (new default) and `all` (previous behaviour), to collect metrics only from the node we are connecting to, or from the whole cluster.

CHANGELOG.asciidoc
```diff
@@ -86,6 +86,7 @@ https://github.com/elastic/beats/compare/v6.0.0-beta2...master[Check the HEAD di
 - Don't stop Metricbeat if aerospike server is down. {pull}6874[6874]
 - disk reads and write count metrics in RabbitMQ queue metricset made optional. {issue}6876[6876]
 - Add mapping for docker metrics per cpu. {pull}6843[6843]
+- RabbitMQ node metricset only collects metrics of the instance it connects to, `node.mode: all` can be used to collect all nodes as before. {issue}6556[6556] {pull}6971[6971]
```
jsoriano: I wonder if we should move it to breaking changes.

metricbeat/module/rabbitmq/node/_meta/docs.asciidoc
```diff
@@ -1 +1,10 @@
 This is the `node` metricset of the RabbitMQ module.
+
+It collects metrics of nodes, it has two modes that can be selected with the
+`node.mode` setting:
+
+* `self`: collects metrics only from the node from whose endpoint `metricbeat`
+  is collecting metrics. This is the default, and the recommended mode when
+  `metricbeat` is deployed in the same node as the RabbitMQ server.
+* `all`: collects metrics from all the nodes. This is recommended when
```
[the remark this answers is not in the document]
jsoriano: Ok, changing to `node.collect` with values `node` and `cluster`.

jsoriano:
Implemented using reporter V2 api now.

metricbeat/module/rabbitmq/node/data.go
```diff
@@ -129,27 +128,22 @@ var (
...
+		eventMapping(r, node)
 	}
-
-	return events, errors
 }
 
-func eventMapping(node map[string]interface{}) (common.MapStr, *s.Errors) {
-	return schema.Apply(node)
+func eventMapping(r mb.ReporterV2, node map[string]interface{}) {
+	event, _ := schema.Apply(node)
```
jsoriano: I tried not to ignore errors here, but `Apply` is returning an error that seems like an empty list of missing fields (Literally: "`Required fields are missing: `"). This happens also with the existing code, so this is not a regression. I leave investigating this for a future PR.

ruflin: Yes, we need a better general approach here. LGTM for now. Check out my recent changes in the elasticsearch module. There I use the errors for testing purpose only.

ruflin:
jenkins, test this

PR #6974, opened 2018-04-30
Make template processing methods public

ruflin:
As the template processing methods are private, using it from an external tool was not possible. There is some further cleanup needed to fully reuse the template generation methods externally to not have mention of Beats in the index pattern but this is a first step.


ph:
@ruflin I am a bit surprised why hound didn't yell about the comments that describe the methods, would you mind cleaning the method comments for `#Process,` `#Generate` and the type `Processor` so they are valid in go docs.

ruflin:
@ph I'm surprised too. Added.

PR #6976, opened 2018-04-30
Fix docker folder for updating to Golang 1.10.1

simitt:
Rename folder according to current Golang version.

Seems this has been overlooked in https://github.com/elastic/beats/pull/6948


ruflin:
I will keep an eye on the next run of `make package` to confirmed that it goes green this time.

PR #6983, opened 2018-05-01
Update elasticsearch/node metricset to use reporter interface

ruflin:
* Simplify Elasticsearch system tests
* Add service.name field


ruflin:
Failing logstash tests should not be related.

PR #6991, opened 2018-05-02
Detect error during requesting status info from Kibana

kvch:
Previously, if Kibana returned an error during requesting its status info using `/api/status`, it failed silently and the version info was empty. This caused problems during detecting the version info of Kibana for loading ML jobs, leading to cryptic error message: `Exiting: Passed version is not semver:`.

From now on, Kibana response errors are detected by checking the response code. So the following error message is displayed now:
```
Exiting: Error creating Kibana client: fail to get the Kibana version: HTTP GET request to /api/status fails: <nil>. Response: {"statusCode":401,"error":"Unautho
rized","message":"[security_exception] failed to authenticate user [elastic], with { header={ WWW-Authenticate=\"Basic realm=\\\"security\\\" charset=\\\"UTF-8\
\\"\" } }"}.
```


tsg:
@kvch can you backport this one to 6.3, please?

PR #6906, opened 2018-04-20
Start simplify system tests

ruflin:
The idea of this PR is to start to play around with ways to simplify our system tests are remove most of the duplicated code.

* Use check_metricset method for aerospike test
* Add fields check to check_metricset method
* Use parameterized for ceph and couchbase tests

Further fixes:

* Increase timeout for k8s tests
* Fix regexp to remove cfgwarn warnings. The removal removed all WARN / ERROR messages.
* Fix flaky auditbeat test by skipping the error message.

PR #6998, opened 2018-05-02
Prevent warning in Auditbeat's system tests

adriansr:
Cleanup the tests so there is no need to ignore a particular warning printed in the logs.

This was necessary after the logged warnings regexp was fixed in #6906

### DIFF
--- a/auditbeat/tests/system/test_base.py
+++ b/auditbeat/tests/system/test_base.py
@@ -3,7 +3,7 @@ import sys
 import os
 import shutil
 import unittest
-from auditbeat import BaseTest
+from auditbeat import *
 from elasticsearch import Elasticsearch
 from beat.beat import INTEGRATION_TESTS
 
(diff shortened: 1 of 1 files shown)


auditbeat/tests/system/test_base.py
```diff
@@ -13,66 +13,72 @@ def test_start_stop(self):
...
-        exit_code = self.run_beat(extra_args=["setup", "--template"])
+            self.render_config_template(
+                modules=[{
+                    "name": "file_integrity",
+                    "extras": {
+                        "paths": dirs,
+                    }
+                }],
+                elasticsearch={"host": self.get_elasticsearch_url()})
+            exit_code = self.run_beat(extra_args=["setup", "--template"])
```
andrewkroh: `run_beat` accepts an `exit_code` param if you'd like to have it check the value for you.

adriansr: done

PR #7005, opened 2018-05-03
Winlogbeat: Replace hardcoded delays in checkpoint tests

adriansr:
Tests in the checkpoint package were susceptible to the timing of certain writes to disk, which caused intermittent failures in CI. Replaced hardcoded delays with an adaptative timeout check.


ruflin:
This is a nice testing feature we should probably use in other places we use time.Sleep.

PR #7012, opened 2018-05-03
Update github.com/elastic/go-libaudit

andrewkroh:
- Added better error messages for when the auditd module fails due to the
  Linux kernel not supporting auditing (CONFIG_AUDIT=n). elastic/go-libaudit#32


ph:
@andrewkroh would it be risky to backport it to 6.2.x?

andrewkroh:
It wouldn't be a problem to bring this change alone into 6.2.x. But given that the change is in go-libaudit v0.2.1, I would need to update the 6.2 branch from v0.1.0 to v0.2.1 and there are some additional changes that would be included (none of them are risky but it would include several enhancements).

PR #7018, opened 2018-05-04
Reduce libbeat state being hold in beat for dynamic loading modules

urso:
Metricbeat/Filebeat modules and inputs require to create a connection to
the shared libbeat per go-routine. This requires the beat to keep some
libbeat state, in order to connect a module to the pipeline when loading
a module dynamically (config reloading/autodiscovery).

With this change, the pipeline is passed to the RunnerFactory, This way
the libbeat-internal state required for connecting a dynamically loaded
module to libbeat is managed by libbeat, not the beat itself.

### DIFF
--- a/CHANGELOG-developer.asciidoc
+++ b/CHANGELOG-developer.asciidoc
@@ -19,5 +19,7 @@ The list below covers the major changes between 6.3.0 and master only.
 
 ==== Breaking changes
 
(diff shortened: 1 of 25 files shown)


exekias:
OK I think I pressed the wrong button, as I realize now @ph review was still missing :disappointed:

@ph I think we can ask for follow up PRs if you see anything that needs to be changed, or I can open a revert PR

Sorry everyone

ph:
@exekias All good for me, I am adding a post merge LGTM

PR #7023, opened 2018-05-04
Feature: TLS support for TCP input

ph:
**This PR is composed of multiple commits that touch multiple parts of beats that were required, each commits message should be explicit.**

We can now receive events on the TCP input with a TLS connection, the
input uses existing type to make sure we have the same naming convention
and code used by outputs that support TLS communication (Elasticsearch
and Logstash).

The configuration will look like this:

```
  host: "localhost:9000"
  ssl.enabled: true
  ssl.verification_mode: full # default
  ssl.supported_protocols: [TLSv1.1]
  ssl.cipher_suites: []
  ssl.certificate_authorities: ["/etc/cacert"]
  ssl.certificate: /etc/mycert.crt
  ssl.key: /etc/mycert.key
  ssl.client_authentification: required

```

One added configuration is `client_authentification`, this option is
only used in the context of server and define how we will force the
authentification, it support the three following options:

- `required`: Assume that the client will provide a certificate and we
will verify it. (default)
- `optional`: If a certificate is given by the client.
- `none`: We don't validate client certificate.

*Note: This commit contains a script to generate certs from a self
signed CA.

### Note to reviewers

Please use rebase not squash, each commit make sure beats was stable at that point.


ph:
thanks hound I will fix them :)

libbeat/common/transport/tlscommon/config.go
```diff
@@ -0,0 +1,86 @@
+package tlscommon
```
ruflin: I assume you named it `tlscommon` instead of `tls` because of the name conflict?

winlogbeat/docs/fields.asciidoc
```diff
@@ -215,7 +215,7 @@ Contains common fields available in all event types.
 required: True
 
 The event log API type used to read the record. The possible values are "wineventlog" for the Windows Event Log API or "eventlogging" for the Event Logging API.
-The Event Logging API was designed for Windows Server 2003 or Windows 2000 operating systems. In Windows Vista, the event logging infrastructure was redesigned. On Windows Vista or later operating systems, the Windows Event Log API is used. Winlogbeat automatically detects which API to use for reading event logs.
+The Event Logging API was designed for Windows Server 2003, Windows XP, or Windows 2000 operating systems. In Windows Vista, the event logging infrastructure was redesigned. On Windows Vista or later operating systems, the Windows Event Log API is used. Winlogbeat automatically detects which API to use for reading event logs.
```
ruflin: I wonder how that got in again as I think @dedemorton removed it recently.

dedemorton: Indeed I did: https://github.com/elastic/beats/commit/33b4f9a0a339f8cbc7893ddb93b53017bef5fb2a#diff-5ef49766a1ba7710815be5733d346e9e

ph:
@dedemorton @ruflin I will rebase it, maybe I was missing a commits or something, I do not remember having a merge conflict at all.

libbeat/common/transport/tlscommon/config.go
ph: Exactly, I am open to better name.

ph:
@ruflin concerning the multiple commits I was reading your mind and I will make multiple PRs for it. It just easier to develop it as a whole thing.

ruflin:
@ph Fully agree that development is much easier tackling the full thing at once.

PR #7030, opened 2018-05-07
Disable fsstat and filesystem as not enabled by default

ruflin:
In the reference file fsstat and filesystem were enabled but are not part of the defaults for the system module. With this change the metricsets in the reference file are in sync with the defaults for the module.


ruflin:
jenkins, test this

PR #7031, opened 2018-05-07
Change title to Drop Events of processor

ruflin:
The title of the `drop_event` processor was "Drop Events" but I think it should be "Drop Event" to be in line with other processors.

### DIFF
--- a/libbeat/docs/processors-using.asciidoc
+++ b/libbeat/docs/processors-using.asciidoc
@@ -469,7 +469,7 @@ exist in the event are overwritten by keys from the decoded JSON object. The
 default value is false.
 
 [[drop-event]]
-=== Drop events
+=== Drop event
 
 The `drop_event` processor drops the entire event if the associated condition
 is fulfilled. The condition is mandatory, because without one, all the events


libbeat/docs/processors-using.asciidoc
```diff
@@ -469,7 +469,7 @@ exist in the event are overwritten by keys from the decoded JSON object. The
 default value is false.
 
 [[drop-event]]
-=== Drop events
+=== Drop event
```
dedemorton: I would either say "Drop an event" or "Drop events" (for consistency with other docs).

ruflin: Drop fields is actually plural because it can drop multiple fields. As the processor only processes one event, it can only drop one at the time and is why the processor name is singular.

Let me know if you prefer to keep as is.

dedemorton: I see where you're coming from (the processor itself drops a single event at at time, unlike `drop_fields`, which can drop multiple fields). But from the user's perspective, won't the processor drop all events that match the specified condition? In general, we should use titles that describe the user goal rather than how the software behaves.

ruflin:
Closing the PR based on the feedback above.

PR #7037, opened 2018-05-08
Add processor per prospector definition to docs

stevokk:
I lost some time because I didn't know you could add processors under prospectors.

https://discuss.elastic.co/t/nginx-harvester-finds-and-loads-log-files-but-no-publish-output/130657

Hopefully this helps somebody else.

Your documentation opened the edit onto branch "5.3" but I assume this works from other versions too.


ruflin:
@stevokk Thanks for your PR.  You add this against the 5.3 branch. Any chance you could open this against master and we can then backport it from there.

Can you also sign the CLA?

PR #7038, opened 2018-05-08
Add module specific prospectors to processors

stevokk:
I had to close the previous commit as the file location and contents had changed too much to merge over.

Slightly expanded on why you would want to scope processors to a module (You use the module terminology in the config directory, and that might be clearest for readers looking at the setup file structure)

Edit: Signed the CLA

### DIFF
--- a/libbeat/docs/processors-using.asciidoc
+++ b/libbeat/docs/processors-using.asciidoc
@@ -32,6 +32,20 @@ condition is present, then the action is executed only if the condition is
 fulfilled. If no condition is passed, then the action is always executed.
 * `<parameters>` is the list of parameters to pass to the processor.
 
+
+You can also scope processors to individual plugin prospectors by placing them under the prospector definition, this is common when wanting to process only logs from modules such as docker, but having more than one prospector enabled: 
+
+[source,yaml]
+------
+- type: <propector_type>
+  processors:
+   - <processor_name>:
+     when:
+        <condition>
+     <parameters>
+...
+------
+
 [[processors]]
 ==== Processors
 


ruflin:
@stevokk Thanks for opening this against master and signing the CLA.

ruflin:
jenkins, test it

libbeat/docs/processors-using.asciidoc
```diff
@@ -32,6 +32,20 @@ condition is present, then the action is executed only if the condition is
 fulfilled. If no condition is passed, then the action is always executed.
 * `<parameters>` is the list of parameters to pass to the processor.
 
+
+You can also scope processors to individual plugin prospectors by placing them under the prospector definition, this is common when wanting to process only logs from modules such as docker, but having more than one prospector enabled:
```
dedemorton: Thanks for your contribution!

In 6.3, we are renaming `prospectors` to `inputs`. Both names are valid in the config, but since the docs reflect the name change, we should  use `input` here. I'd also prefer to simplify the language (we're getting users from all sorts of backgrounds these days). Because this content is shared across all the Beats books, we also need to add some conditional coding so that the info renders in the Filebeat docs, but not elsewhere. 

I'd replace your changes with the following text (includes asciidoc formatting to make everything work):

```
ifeval::["{beatname_lc}"=="filebeat"]
You can define processors at the top-level in the config to filter all data
collected by {beatname_uc}. Or you can specify processors under a specific input
to filter the data collected by that input. For example:

[source,yaml]
------
- type: <input_type>
  processors:
   - <processor_name>:
     when:
        <condition>
     <parameters>
...
------

This is useful when you have more than one input defined, but want to filter
only the events collected by a specific input.
endif::[]
```

I wouldn't mention modules here because that's a specific (more advanced) use case.

dedemorton:
@stevokk I'll take care of backporting this change to earlier releases where the name is still prospector (you don't have to worry about that). Typically we don't backport changes all the way back to the 5.x branches. Is that OK with you?

stevokk:
@dedemorton Your version looks much simpler than mine but definitely explains it clearly, would you like to make the change or should I ? Also happy for this to be closed off if you want to start fresh.

dedemorton:
@stevokk I'll create a PR on top of your changes and submit it. Thanks!

dedemorton:
Closing because this is replaced by https://github.com/elastic/beats/pull/7085

PR #7042, opened 2018-05-08
Readd BEAT_VERSION to compose project

ruflin:
This was removed previously but potentially could cause issue if the same commit is in different branches and the projects were not cleaned up properly.

Further changes:

* `--pull` and `--force-rm` are added when building the image. The `--pull` flag ensures that for images which we didn't specify the exact version it is checked every time if a new version is available. The `--force-rm` removes intermediate containers after the build. This should lead to less obsolete intermediate containers on the CI workers.
* The error handling inside compose for Golang was improved to have error messages log in case of killing errors
* logstash and kibana were added to the services that are kept running. The assumption is that sometimes a race happened trying to start Logstash during shutdown


ruflin:
jenkins, test this

Issue #7041, opened 2018-05-08
Dashboards saved from master don't work in kibana 6.3


tsg:
It seems that we saved some of the dashbaord using the master version of Kibana. This can cause issues in 6.3 because there's some compatibility code in Kibana that takes the version into account in order to apply BWC code. See https://github.com/elastic/kibana/issues/18913

The fix is simple: we need to replace `7.0.0-SNAPSHOT` with `6.2.4` in the dashboard JSONs. I know that the Auditbeat dashboards are affected, but we should check all dashboards.

PR #7043, opened 2018-05-08
Fix dashboards for Kibana 6.3

adriansr:
Dashboards need to use version 6.2.4 to enable a workaround for the new [smaller panel grid dimensions](https://github.com/elastic/kibana/pull/16763).

Closes https://github.com/elastic/beats/issues/7041


adriansr:
Tested all the affected dashboards with Kibana 6.3, now they have the right size

PR #7049, opened 2018-05-09
Combine Elasticsearch module integration tests

ruflin:
This combines all Elasticsearch module integration tests on the module level. This has the advantage that it removes duplicated code and makes adding a metricset to the test very simple. Also data generators was move to the module level to have it one place. It also should speed up tests as Elasticsearch has to be started only once.

As package name `elasticsearch_test` is used to make sure tests happen inside it's own package.

For the data generation a path was introduced to still write the data files into the correct directory.


metricbeat/mb/testing/data_generator.go
```diff
@@ -75,13 +75,13 @@ func WriteEventsCond(f mb.EventsFetcher, t testing.TB, cond func(e common.MapStr
...
 	fullEvent := CreateFullEvent(f, *event)
-	WriteEventToDataJSON(t, fullEvent)
+	WriteEventToDataJSON(t, fullEvent, "")
 	return nil
 }
 
 // WriteEventsReporterV2 fetches events and writes the first event to a ./_meta/data.json
 // file.
-func WriteEventsReporterV2(f mb.ReportingMetricSetV2, t testing.TB) error {
+func WriteEventsReporterV2(f mb.ReportingMetricSetV2, t testing.TB, path string) error {
```
jsoriano: I wonder if the path is going to be most of the times based on the metricset name, because then maybe we could guess it from `f.Name()` or similar. This would require less knowlegde when writting tests. WDYT?

ruflin: It is for all tests which are written on the module level. Not sure yet where we end up here TBH. Ok to simplify this later?

PR #6911, opened 2018-04-20
Port fields.yml collector to Golang

kvch:
Existing Python and shell code is ported to Golang. The previous Beat specific field collection is is generalized and moved to the `Makefile` of `libbeat`. A new variable is added to makefiles: `FIELDS_FILE_PATH`. This specifies where `fields.yml` are.

## Misc
- Packetbeat's `fields_base.yml` is renamed to `fields.common.yml` to be the same as in other Beats

# Breaking change for community Beats
If your Beat has partial `fields.yml` files which need to be collected by `make update`, please fill in the `FIELDS_FILE_PATH` variable in your `Makefile` with the root directory of your fields files.

If you need a custom collector function you can implement it `module_fields_collector.go` with the following type signature:
```
func CollectModuleFiles(root string) ([]*fields.YmlFile, error)
```

PR #7051, opened 2018-05-09
Add host.name in the events

### LINKED ISSUE #7050: Metricbeat `host` namespace causes mapping conflict when used with Logstash
Here is what's happening:
* the `add_host_metadata` processor creates a `host` *object*. The new processor is disabled by default, but the fields are present in the Beats templates.
* The Logstash beats input adds a `host` (*string*) if host is not found in the original object. Code is [here](https://github.com/logstash-plugins/logstash-input-beats/blob/776f05ce52111d39e394798428598abfd4dbbf68/lib/logstash/inputs/beats/event_transform_common.rb#L14-L20)
* since the Beats processor is not enabled by default, the `host` sent from Logstash to Elasticsearch is a *string*. This conflicts with the object mapping that we have in the Metricbeat template.

The result of the above is that a stock MB -> LS -> ES pipeline doesn't work :(

tsg:
As a solution for #7050, we're adding a `host.name` field to
all events. This is duplicate information from `beat.name`,
but is used to avoid the mapping conflict and to slowly
introduce the "host as an object" approach.

To remove the duplication, you can remove `beat.name` like this:

    processors:
      - drop_fields.fields: ["beat.name"]

Closes #7050.

### DIFF
--- a/dev-tools/jenkins_ci.ps1
+++ b/dev-tools/jenkins_ci.ps1
@@ -36,13 +36,14 @@ exec { go get -u github.com/jstemmer/go-junit-report }
 echo "Building $env:beat"
 exec { go build } "Build FAILURE"
 
+# always build the libbeat fields
+cp ..\libbeat\_meta\fields.common.yml ..\libbeat\_meta\fields.generated.yml
+cat ..\libbeat\processors\*\_meta\fields.yml | Out-File -append -encoding UTF8 -filepath ..\libbeat\_meta\fields.generated.yml
+cp ..\libbeat\_meta\fields.generated.yml ..\libbeat\fields.yml
+
 if ($env:beat -eq "metricbeat") {
     cp .\_meta\fields.common.yml .\_meta\fields.generated.yml
     python .\scripts\fields_collector.py | out-file -append -encoding UTF8 -filepath .\_meta\fields.generated.yml
-} elseif ($env:beat -eq "libbeat") {
-    cp .\_meta\fields.common.yml .\_meta\fields.generated.yml
-    cat processors\*\_meta\fields.yml | Out-File -append -encoding UTF8 -filepath .\_meta\fields.generated.yml
-    cp .\_meta\fields.generated.yml .\fields.yml
 }
 
 echo "Unit testing $env:beat"
--- a/libbeat/publisher/pipeline/module.go
+++ b/libbeat/publisher/pipeline/module.go
@@ -50,13 +50,15 @@ func Load(
 		Processors:    processors,
 		Annotations: Annotations{
 			Event: config.EventMetadata,
-			Beat: common.MapStr{
-				"name":     name,
-				"hostname": beatInfo.Hostname,
-				"version":  beatInfo.Version,
-			},
-			Host: common.MapStr{
-				"name": name,
+			Builtin: common.MapStr{
+				"beat": common.MapStr{
+					"name":     name,
+					"hostname": beatInfo.Hostname,
+					"version":  beatInfo.Version,
+				},
+				"host": common.MapStr{
+					"name": name,
+				},
 			},
 		},
 	}
--- a/libbeat/publisher/pipeline/module.go
+++ b/libbeat/publisher/pipeline/module.go
@@ -55,6 +55,9 @@ func Load(
 				"hostname": beatInfo.Hostname,
 				"version":  beatInfo.Version,
 			},
+			Host: common.MapStr{
+				"name": name,
+			},
 		},
 	}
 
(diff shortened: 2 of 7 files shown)


libbeat/publisher/pipeline/pipeline.go
```diff
@@ -93,6 +94,7 @@ type Settings struct {
 type Annotations struct {
 	Beat  common.MapStr
 	Event common.EventMetadata
+	Host  common.MapStr
```
urso: Hmmm.... do we consider to introduce more namespaces? If so, how about changing Annotations to:

```
type Annotations struct {
  Event common.EventMetadata // additional event fields configured by user
  Builtin common.MapStr
}
```

And then init Annotations with:

```
Annotations{
  Event: config.EventMetaData,
  Builtin: common.MapStr{
    "beat": ...
    "host": ...
  },
}
```

This way the logic how/when the builtin namespaces are merged needs to be done only once.

libbeat/publisher/pipeline/processor.go
```diff
@@ -91,10 +91,13 @@ func newProcessorPipeline(
 	// setup 5: client processor list
 	processors.add(localProcessors)
 
-	// setup 6: add beats metadata
+	// setup 6: add beats and host metadata
 	if meta := global.beatsMeta; len(meta) > 0 {
 		processors.add(makeAddFieldsProcessor("beatsMeta", meta, needsCopy))
 	}
+	if meta := global.hostMeta; len(meta) > 0 {
+		processors.add(makeAddFieldsProcessor("hostMeta", meta, needsCopy))
```
urso: This is safe, as needsCopy is true if any global processors are configured.

We can also 'reduce' some copies by introducing `needsGlobalCopy := global.alwaysCopy || global.processors != nil` and `needsCopy := needsGlobalCopy || local.processors != nil`.

andrewkroh:
This new `host.name` field also needs to be added to fields.yml.

IIUC this could cause mapping conflicts for anyone using LS in their pipeline today. In particular it will cause issues for anyone not using versioned index names (e.g. `%{[@metadata][beat]}-%{[@metadata][version]}-%{+YYYY.MM.dd}`).

And it could causes issues with any LS config that expects `[host]` to be a string (e.g. LS conditionals). So I think we might want to put something in the breaking changes section about this.

tsg:
@urso @ruflin Can you have a second look? Also, @andrewkroh has a point above, this is kind of a breaking change. I hope we're not missing other cases like this. cc @jsvd as well.

tsg:
jenkins, test it

tsg:
Jenkins, test this again

dev-tools/jenkins_ci.ps1
```diff
@@ -36,13 +36,14 @@ exec { go get -u github.com/jstemmer/go-junit-report }
 echo "Building $env:beat"
 exec { go build } "Build FAILURE"
 
+# always build the libbeat fields
+cp ..\libbeat\_meta\fields.common.yml ..\libbeat\_meta\fields.generated.yml
```
ruflin: As soon as https://github.com/elastic/beats/pull/6911 is in this should become much simpler.

libbeat/publisher/pipeline/module.go
```diff
@@ -50,10 +50,15 @@ func Load(
 		Processors:    processors,
 		Annotations: Annotations{
 			Event: config.EventMetadata,
-			Beat: common.MapStr{
-				"name":     name,
-				"hostname": beatInfo.Hostname,
-				"version":  beatInfo.Version,
+			Builtin: common.MapStr{
```
ruflin: I like that change because it prepares us for more fields coming from ECS.

orarnon:
hi @tsg , We still have issues with Filebeat 6.3.1:  
```
[2018-08-01T09:19:04,028][WARN ][logstash.outputs.elasticsearch] Could not index event to Elasticsearch. {:status=>400, :action=>["index", {:_id=>nil, :_index=>"tag-mediation-2018.08.01", :_type=>"logs", :_routing=>nil}, 2018-08-01T09:19:03.185Z {"name":"tagmediation-i-0c6dadf01c99ec637.eu-west-1.production"} %{message}], :response=>{"index"=>{"_index"=>"tag-mediation-2018.08.01", "_type"=>"logs", "_id"=>"AWT0yJcsvdiggc-2mDWo", "status"=>400, "error"=>{"type"=>"mapper_parsing_exception", "reason"=>"failed to parse [host]", "caused_by"=>{"type"=>"illegal_state_exception", "reason"=>"Can't get text on a START_OBJECT at 1:656"}}}}}
```

urso:
@orarnon This error message looks very unrelated. Please use https://discuss.elastic.co to ask for some help.

Issue #7036, opened 2018-05-08
Fetch logs from the Elastic Stack with Filebeat


ruflin:
Make it possible to fetch logs from the Elastic Stack with Filebeat.

This is a meta issue to track the ongoing work.

* [x] Elasticsearch https://github.com/elastic/beats/issues/5301
* [x] Kibana: https://github.com/elastic/beats/pull/7052
* [x] Logstash https://github.com/elastic/beats/pull/5481
* [ ] APM-Server
* [ ] Check that all filesets report `service.name`

urso:
@ycombinator Any idea if we can close this or tick off some checkboxes?

ycombinator:
Thanks for the assignment and ping @urso. I ticked off some checkboxes :) and will look into the remaining items.

jlind23:
Backlog grooming: Closing it for now until further activity, can still be reopened if needed.

PR #7052, opened 2018-05-09
Filebeat Kibana module log fileset

ruflin:
This adds the Kibana module to Filebeat reading the Kibana log. The Kibana log is in JSON format which simplifies reading it but at the same time it provides many fields which are not necessarily know in advance. To solve this issue The data is put under `kibana.log.meta` and the most common known fields are picked into `kibana.log.*` or directly put into the global field like `message`. The fields under `meta` are stored as keyword. This make sure all the meta information around the event is still in the index but prevents potential type conflicts like long vs double.

Additional change:

* Module tests always overwrite the pipeline. This should make development easier.


ruflin:
@tsg The Kibana log is pretty interesting as it's all in json which is pretty nice. The downside of it is there are lots of fields and I assume in my test logs not all fields that can show up in there. At the moment I'm not sure how we should deal with all these fields as I don't think it will scale to map all of them. I was thinking perhpas we can pick the ones we think are most relevant and put all the others under a namespace and just say it's object type and mapping will be done dynamically?

ruflin:
PR updated with proposed solution for json handling (see PR message). Ready for review.

filebeat/filebeat.reference.yml
```diff
@@ -161,6 +161,16 @@ filebeat.modules:
     #var.paths:
 
 
+#------------------------------- kibana Module -------------------------------
+- module: {{ module }}
```
kvch: Hmmm... This seems to be a problem of the generator. Do you mind changing it manually now. I am opening a PR with the fix in the meantime.

filebeat/module/kibana/_meta/config.yml
```diff
@@ -0,0 +1,8 @@
+- module: {{ module }}
```
kvch: Also, this file is problematic.

filebeat/modules.d/kibana.yml.disabled
```diff
@@ -0,0 +1,8 @@
+- module: {{ module }}
```
kvch: This one, too.

ruflin:
@kvch I made 2 additional small changes: Added a beta label to the docs and adjusted the path to read logs from.

filebeat/module/kibana/log/ingest/pipeline.json
```diff
@@ -0,0 +1,63 @@
+{
+    "description": "Pipeline for parsing Kibana log logs",
```
kvch: Sorry for spotting it quite late, but could you please remove "log" from the description.

kvch:
I tested it locally, works like a charm. After that minor problem is fixed, it's ready to be merged.

tsg:
Would it be possible to add at least a simple Kibana dashboard? So far all FB modules come with at least a dashboard, and I'd like to keep it that way :)

filebeat/module/kibana/log/ingest/pipeline.json
ruflin: done

ruflin:
Added Dashboards to the meta issue here: https://github.com/elastic/beats/issues/7036

PR #7059, opened 2018-05-09
Add kubernetes apiserver metricset

exekias:
This new metricset gathers metrics from Kubernetes API servers, giving visibility on number of requests and latency.

This change adds support for `summary` and `histogram` metric type to our Prometheus helper, histogram is used to retrieve latency info.

I also simplified the process of creating a new metricset based on this helper, to the point you only need [this](https://github.com/exekias/beats/blob/5f4568208ac273d9b759343b22f783184951a477/metricbeat/module/kubernetes/apiserver/apiserver.go) to create it and [this](https://github.com/exekias/beats/blob/5f4568208ac273d9b759343b22f783184951a477/metricbeat/module/kubernetes/apiserver/apiserver_test.go) to test it.

download


metricbeat/module/kubernetes/apiserver/apiserver_test.go
```diff
@@ -0,0 +1,23 @@
...
+)
+
+const testFile = "_meta/test/metrics"
+
+func TestEventMapping(t *testing.T) {
+	prometheus.TestMetricSet(t, "kubernetes", "apiserver",
+		prometheus.TestCases{
+			{
+				MetricsFile:  "./_meta/test/metrics",
+				ExpectedFile: "./_meta/test/metrics.expected",
```
ruflin: This reminds me a lot of what @simitt was doing here with the approvals: https://github.com/elastic/apm-server/blob/master/tests/approvals.go I wanted to look into it to also do some comparison on json docs for Elasticsearch. Perhaps we can sync up on this.

simitt: Thats the `approvals` script that you can run from cmd line: https://github.com/elastic/apm-server/blob/master/tests/scripts/approvals.go

Would be great to put this in its own small library for more generic usage.

exekias: Agreed, we can discuss this and come up with a standardized way to deal with expected output in these tests, also pinging @jsoriano here, as he is working on some ideas to improve our testing.

exekias:
jenkins, retest this please

metricbeat/helper/prometheus/module.go
```diff
@@ -0,0 +1,141 @@
...
+			"metricsets": []string{metricset},
+			"hosts":      []string{server.URL},
+		}
+
+		f := mbtest.NewEventsFetcher(t, config)
+
+		events, err := f.Fetch()
+		assert.NoError(t, err)
+
+		if update {
```
jsoriano: What about using a flag for `update` as we do with `go test -data`?

metricbeat/module/kubernetes/apiserver/apiserver_test.go
jsoriano: The ideas I'm working on are more focused to improve and scale integration tests, but any abstraction like these ones are more than welcome of course :)

metricbeat/module/kubernetes/apiserver/apiserver.go
```diff
@@ -0,0 +1,28 @@
...
+			"scope":       prometheus.KeyLabel("request.scope"),
+			"subresource": prometheus.KeyLabel("request.subresource"),
+			"verb":        prometheus.KeyLabel("request.verb"),
+		},
+	}
+
+	if err := mb.Registry.AddMetricSet("kubernetes", "apiserver",
+		prometheus.MetricSetBuilder(mapping), prometheus.HostParser); err != nil {
+		panic(err)
+	}
```
jsoriano: Use `mb.Mregistry.MystAddMetricset()` here?

exekias:
@jsoriano I've incorporated your suggestions, thanks!

exekias:
jenkins, retest this please

metricbeat/helper/prometheus/module.go
```diff
@@ -0,0 +1,144 @@
...
+)
+
+var (
+	// HostParser validates Prometheus URLs
+	HostParser = parse.URLHostParserBuilder{
+		DefaultScheme: defaultScheme,
+		DefaultPath:   defaultPath,
+	}.Build()
+
+	expectedFlag = flag.Bool("update_expected", false, "Update prometheus expected files")
```
ruflin: Is this only used for testing? If yes, should it really be registered here?

metricbeat/helper/prometheus/module.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+}
+
+type prometheusMetricSet struct {
+	mb.BaseMetricSet
+	prometheus Prometheus
+	mapping    *MetricsMapping
+}
+
+func (m *prometheusMetricSet) Fetch() ([]common.MapStr, error) {
```
ruflin: Can happen in a later PR, but would be great to also use here the ReporterV2 interface.

metricbeat/helper/prometheus/module.go
```diff
@@ -0,0 +1,144 @@
...
+	MetricsFile string
+
+	// ExpectedFile containing resulting documents
+	ExpectedFile string
+}
+
+// TestMetricSet goes over the given TestCases and ensures that source Prometheus metrics gets converted into the expected
+// events when passed by the given metricset.
+// If -update_expected flag is passed, the expected JSON file will be updated with the result
+func TestMetricSet(t *testing.T, module, metricset string, cases TestCases) {
```
ruflin: I'm thinking we should separate the testing parts and the module parts.

metricbeat/module/kubernetes/apiserver/apiserver.go
```diff
@@ -0,0 +1,27 @@
+package apiserver
```
ruflin: That is probably the most compact metricset we have. I like it.

metricbeat/helper/prometheus/module.go
jsoriano: You are right, if it is defined here it will appear in the production binary. Testing code should be in a submodule to avoid that.

exekias: As we use cobra, common flags are not really "leaked" to the binary, I'm ok with moving this though

exekias:
I converted the helper to use `ReporterV2`

exekias:
jenkins, test this again

ruflin:
@exekias Will need a rebase and a `make update`.

metricbeat/helper/prometheus/ptest/ptest.go
```diff
@@ -0,0 +1,106 @@
+package ptest
```
ruflin: I wonder if we should call these packages always `testing` as we do in some other places or if this is always and issue because there is a `testing` package.

Note: Just for discussion, no change in the PR needed.

exekias: I had the same thought, but as this is always used along with the Go `testing` package clash is guaranteed.

This already happens with the `mb/testing` package, we end up renaming it on import to `mbtest`, with the annoyance of breaking automatic imports.

jsoriano: There is another example in the latest changes on RabbitMQ module, all modules end up being renamed, but I think this is not so bad (https://github.com/elastic/beats/pull/7106/files#diff-7d67ada6eac5ae4109019e586c93d3a1R8)

I'm slightly in favour of having the same name for all test modules, though not sure if `testing` is the best option as it collides with the standard library.

ruflin: For the module testing packages I think all must have the same name or structure, as we need to somehow modify the `imports` script to skip them. In general what i worry about directories in module is that they can always conflict with metricset names.

exekias:
jenkins, retest this please

PR #7047, opened 2018-05-08
Metricbeat: Ensure canonical naming for JMX beans is disabled

jsoriano:
Canonical naming for JMX beans in Jolokia is enabled by default, this
orders mbean fields, what breaks metricbeat internal mapping.

We were already setting these options in the url, but url parameters are
ignored in POST requests. This change adds these parameters as part of
the body.

PR #7061, opened 2018-05-09
Metricbeat: Ensure canonical naming for JMX beans is disabled (backport of #7047 to 6.3)

jsoriano:
Canonical naming for JMX beans in Jolokia is enabled by default, this
orders mbean fields, what breaks metricbeat internal mapping.

We were already setting these options in the url, but url parameters are
ignored in POST requests. This change adds these parameters as part of
the body.


jsoriano:

ruflin:
This should not be merged before 6.3 is released as I don't think this is a critical bug fix.

PR #7072, opened 2018-05-10
Add pod-uid support for add_kubernetes_metadata 

mariomechoulam:
Hello everyone!

I am opening this pull request to gather some feedback on the approach.
There are a few things that can be cleaned up and improved for sure, as this is my first try with Go, but the basic should be working.
I was hesitating whether to create a completely new `Matcher` and duplicate a lot of the existing logic or extend the current one. Extending did not feel entirely right, because the `LogPathMatcher` was focused on container ID - nevertheless it is still a _log_path_ matcher. Thus this middle-ground approach in which I am adding a new configuration parameter for the `Matcher`; it should be seamlessly backwards compatible.

I believe this is the configuration that should make it work...
```
processors:
  - add_kubernetes_metadata:
      in_cluster: true
      default_indexers.enabled: false
      default_matchers.enabled: false
      indexers:
        - pod_uid
      matchers:
        - logs_path:
            logs_path: /var/lib/kubelet/pods/
            resource_type: pod
```

My understanding is that, ideally, an integration test of some sort would be useful - at least to test that the above configuration will indeed produce the desired functionality. Will welcome any guidance :)

Cheers!


mariomechoulam:
I guess that the CI is failing because of the style and formatting. I have tried executing `make fmt` but I get an error. I am on Manjaro now, which ships with both Python 2 and 3.
Have tried everything step by step to make sure that I have virtualenv with Python 2 and pip2 - and I have. Even manually `source build/python-env/bin/activate`-d before running the task. However every time I get:

```
make[1]: Entering directory '/home/mario/git/go/src/github.com/elastic/beats/libbeat'
Cache entry deserialization failed, entry ignored
  Cache entry deserialization failed, entry ignored
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-install-8n1zw1h_/functools32/
make[1]: *** [scripts/Makefile:232: python-env] Error 1
make[1]: Leaving directory '/home/mario/git/go/src/github.com/elastic/beats/libbeat'
make: *** [Makefile:92: fmt] Error 1
```
Probably those cache lines are a hint that it tries to run a different pip version. I am confident that executing `sudo pip2 install --upgrade functools32` is working, not sure why it does not through `make fmt`.

Fetched `goimports` manually and after running it I am getting quite a lot of modifications, most in files I did not modify and several imports being removed altogether. Honestly, don't feel comfortable commiting that...
Can anyone shed some light and unblock me? I can send a patch with the `goimports` modifications.

filebeat/processor/add_kubernetes_metadata/matchers.go
```diff
@@ -24,13 +24,16 @@ const LogPathMatcherName = "logs_path"
 
 type LogPathMatcher struct {
 	LogsPath string
+	ResourceType string
 }
 
 func newLogsPathMatcher(cfg common.Config) (add_kubernetes_metadata.Matcher, error) {
 	config := struct {
 		LogsPath string `config:"logs_path"`
+		ResourceType string `config:"resource_type"`
```
exekias: this config starts to get complicated, I guess we can live with this how it is now, but I will try to give it a thought on how to simplify this mechanism. Not a blocker for this PR :)

libbeat/common/kubernetes/metadata.go
```diff
@@ -67,6 +70,21 @@ func (g *metaGenerator) PodMetadata(pod *Pod) common.MapStr {
 	return meta
 }
 
+func (g *metaGenerator) PodMetadataWithUID(pod *Pod) common.MapStr {
```
exekias: I think this should be a setting instead, so `pod.uid` is added if requested (both by the user or your indexer

exekias:
About the `goimports` issue, you can probably just use `goimports -local github.com/elastic -l -w <file_to_review.go`. Let me know if that works for you

libbeat/common/kubernetes/metadata.go
mariomechoulam: Hey!
Thanks for the feedback. So, just to make sure I am understand correctly, we are talking about adding the `pod.uid` from the configuration, anytime that
* The `pod_uid` Indexer is used
* `pod.uid` is specified as a field inside the `fields` element?

exekias: I was thinking more about doing something similar to the current `include_annotations` setting, for instance. Parameters are passed here: https://github.com/elastic/beats/blob/master/libbeat/common/kubernetes/metadata.go#L24

And the processor allows configuration for them here: https://github.com/elastic/beats/blob/master/libbeat/processors/add_kubernetes_metadata/config.go#L24

The indexer would pass the right parameter (enable pod.uid) without the user asking for it.

What do you think?

mariomechoulam:
Hi, just refactored the code as discussed and updated the unit-tests (also the autodiscover feature albeit a bit late, apologies).
Looks like the formatting is working now, thanks @exekias !
Just ran `make docs` and looks like it did stuff, but right after that I am not seeing any changes on the branch to be commited. Am I doing anything wrong?

exekias:
You will need to:
- rebase master
- run `make update` in the root beats folder

libbeat/processors/add_kubernetes_metadata/kubernetes.go
```diff
@@ -65,7 +66,10 @@ func newKubernetesAnnotator(cfg *common.Config) (processors.Processor, error) {
 		Indexing.RUnlock()
 	}
 
-	metaGen := kubernetes.NewMetaGenerator(config.IncludeAnnotations, config.IncludeLabels, config.ExcludeLabels)
+	//Checks whether the PodUIDIndexer has been added in order to turn on the pod uid gathering
+	_, hasPodUIDIndexer := Indexing.indexers["pod_uid"]
```
exekias: I think it's better if we don't do this automatically, the user must enable it. In any case, the indexer would still work without this, isn't it?

mariomechoulam: Hmmm the intention here was to enable the gathering of `pod.uid` if the indexer was specified in the config, even if the config bool `include_pod_uid` was not.
I thought this is what you meant with your previous comment: `The indexer would pass the right parameter (enable pod.uid) without the user asking for it.`
Or perhaps I am missing something...

exekias: Hi,

Yes, I apologize as that is what I said, then I realized the indexer doesn't need the pod uid in the metadata :innocent: 

Still, I think having `pod.uid` is a good feature for some people, so I would keep it as something optional you can enable. I would only remove this auto-enable thingy and we are ready to go.

mariomechoulam: Awesome!
I'll remove the auto-enable and add the info in the changelog first thing tomorrow.
Thanks man! And no need to apologise :slightly_smiling_face:

exekias:
jenkins, test this please

yu-yang2:
@mariomechoulam Can I use this with emptyDir like this now?
```
processors:
  - add_kubernetes_metadata:
      in_cluster: true
      default_indexers.enabled: false
      default_matchers.enabled: false
      indexers:
        - pod_uid
      matchers:
        - logs_path:
            logs_path: /var/lib/kubelet/pods/
            resource_type: pod
```

exekias:
Thank you @mariomechoulam for your contribution :tada: 

It will go out in `6.4` release

mariomechoulam:
Awesome! Many thanks @exekias for your support!
Do we know if it is working so I can update the initial post and the discuss topic? I was not able to test it beyond unit-test myself...

mariomechoulam:
Hi @yu-yang2 
I have updated the original post with the configuration, I believe, would make it work (notice I added the `include_pod_uid: true` option). Nevertheless, I would wait myself until the confirmation that this is, indeed, the working setup.

exekias:
I can confirm it works, both the indexer and adding `pod.uid`

yu-yang2:
@mariomechoulam you have missing a `:` after pod_uid beyond indexers, here is my config file, this is perfect work with emptyDir even in multi replicas deployment.
```
filebeat.prospectors:
- type: log
  paths:
    - /var/lib/kubelet/pods/*/volumes/kubernetes.io~empty-dir/*/*.log
processors:
  - add_kubernetes_metadata:
      in_cluster: true
      default_indexers.enabled: false
      default_matchers.enabled: false
      include_pod_uid: true
      indexers:
        - pod_uid:
      matchers:
        - logs_path:
            logs_path: /var/lib/kubelet/pods/
            resource_type: pod
output.elasticsearch:
  hosts: elasticsearch-logging:9200
```

mariomechoulam:
Nice spot @yu-yang2 ! And many thanks for adding the correct configuration here; I'll update the original post.
Very happy to see that you found this helpful and are using it :slightly_smiling_face: 
Cheers!

Issue #6875, opened 2018-04-16
RabbitMQ module path_prefix


xadcoh:
Hello.

This issue based on https://discuss.elastic.co/t/rabbitmq-module-path-prefix/127292

When RabbitMQ configured with option:
`management.path_prefix = /<prefix>`
Metrics can't be collected with:
`metricset: ["node", "queue"]
hosts: ["<ip>:15672"]`
Error is:
`HTTP error 406 in node: 406 Not Acceptable`
This is expected behavior because there is no option for prefix in module config. Metricbeat will try to reach RabbitMQ with:
`<ip>:15672/api/queues
<ip>:15672/api/nodes`
In this example correct paths are:
`<ip>:15672/rabbit/api/queues
<ip>:15672/rabbit/api/nodes`
Workaround is point metricbeat with full path.
Nodes:
`- module: rabbitmq
  metricsets: ["node"]
  period: 10s
  hosts: ["<ip>:15672/rabbit/api/nodes"]
  username: rabbitmetricbeatuser
  password: strongpass`
Queues:
`- module: rabbitmq
  metricsets: ["queue"]
  period: 10s
  hosts: ["<ip>:15672/rabbit/api/queues"]
  username: rabbitmetricbeatuser
  password: strongpass`

For confirmed bugs, please report:
- Version: 
  metricbeat: 6.2.3
  kibana: 6.2.2
  elasticsearch: 6.2.2
- Operating System:
  centos 7 
- Steps to Reproduce: 
1. Install RabbitMQ
2. Put `management.path_prefix = /<prefix>` in /etc/rabbitmq/rabbitmq.conf
3. Install metricbeat
4. Configure rabbitmq module:
`- module: rabbitmq
  metricsets: ["node", "queue"]
  period: 10s
  hosts: ["<ip>:15672"]
  username: rabbitmetricbeatuser
  password: strongpass`

PR #7074, opened 2018-05-11
Metricbeat: Add configuration for RabbitMQ management prefix path

jsoriano:
Refactor RabbitMQ metricsets to reuse host parser and metricset builder
so management prefix path can be configured as an only module setting.
Check in tests the paths metricbeat is trying to access to collect the
metrics.

Previously it required to configure each metricset in its own module
block with the path in the host, as described in  #6875.

### DIFF
--- a/metricbeat/module/rabbitmq/connection/connection_test.go
+++ b/metricbeat/module/rabbitmq/connection/connection_test.go
@@ -14,13 +14,21 @@ import (
 )
 
 func TestFetchEventContents(t *testing.T) {
-	absPath, err := filepath.Abs("../_meta/testdata/")
+	absPath, _ := filepath.Abs("../_meta/testdata/")
 
-	response, err := ioutil.ReadFile(absPath + "/connection_sample_response.json")
+	response, _ := ioutil.ReadFile(absPath + "/connection_sample_response.json")
+	notFound, _ := ioutil.ReadFile(absPath + "/notfound_response.json")
 	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		w.WriteHeader(200)
-		w.Header().Set("Content-Type", "application/json;")
-		w.Write([]byte(response))
+		switch r.URL.Path {
+		case "/api/connections":
+			w.WriteHeader(200)
+			w.Header().Set("Content-Type", "application/json;")
+			w.Write([]byte(response))
+		default:
+			w.WriteHeader(404)
+			w.Header().Set("Content-Type", "application/json;")
+			w.Write([]byte(notFound))
+		}
 	}))
 	defer server.Close()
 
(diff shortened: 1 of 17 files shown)


metricbeat/module/rabbitmq/connection/connection_test.go
```diff
@@ -14,13 +14,21 @@ import (
 )
 
 func TestFetchEventContents(t *testing.T) {
-	absPath, err := filepath.Abs("../_meta/testdata/")
+	absPath, _ := filepath.Abs("../_meta/testdata/")
```
ruflin: I would add a assert.NoError( for these erors. This will make debugging test much easier.

metricbeat/module/rabbitmq/exchange/exchange_test.go
```diff
@@ -14,13 +14,21 @@ import (
...
-		w.WriteHeader(200)
-		w.Header().Set("Content-Type", "application/json;")
-		w.Write([]byte(response))
+		switch r.URL.Path {
+		case "/api/exchanges":
+			w.WriteHeader(200)
+			w.Header().Set("Content-Type", "application/json;")
+			w.Write([]byte(response))
+		default:
+			w.WriteHeader(404)
```
ruflin: Nice, I should steal this for my elasticsearch http tests.

metricbeat/module/rabbitmq/exchange/exchange.go
```diff
@@ -3,57 +3,31 @@ package exchange
...
-		DefaultPath:   defaultPath,
-	}.Build()
-)
-
-// init registers the MetricSet with the central registry as soon as the program
-// starts. The New function will be called later to instantiate an instance of
-// the MetricSet for each host defined in the module's configuration. After the
-// MetricSet has been created then Fetch will begin to be called periodically.
 func init() {
-	if err := mb.Registry.AddMetricSet("rabbitmq", "exchange", New, hostParser); err != nil {
```
ruflin: Did we miss that in our PR bonanza or this came in later?

metricbeat/module/rabbitmq/url.go
```diff
@@ -0,0 +1,26 @@
+package rabbitmq
+
+import (
+	"github.com/elastic/beats/metricbeat/mb/parse"
+)
+
+// Subpaths to management plugin endpoints
+const (
+	ConnectionsPath = "/api/connections"
```
ruflin: I'm going forth an back if this should be in the module or in the metricset. Is there a benefit that the module knows about these paths?

metricbeat/module/rabbitmq/url.go
```diff
@@ -0,0 +1,26 @@
...
+)
+
+const (
+	defaultScheme = "http"
+	pathConfigKey = "management_path_prefix"
+)
+
+var (
+	// HostParser parses host urls for RabbitMQ management plugin
+	HostParser = parse.URLHostParserBuilder{
```
ruflin: I wonder if it could become an issue that all Metricsets will now share one host parser as a global object?

andrewkroh: It shouldn't be an issue.

ruflin:
Merged it already to move forward with this. I really hope we get similar simplifications for other modules.

PR #7075, opened 2018-05-11
Add Beats state reporting

ruflin:
Currently in monitoring or the API endpoint it's not visible which modules are running. This PR adds and endpoint `/state` which shows the following data:

```
{
  "module": {
    "count": 3,
    "names": [
      "system"
    ]
  }
}
```

The names are a unique list of module names that are running and the count is a total of all the modules that are running. In the above case it's 3 as the system module is started by default 3 times with different configs / periods. Currently this reporting only works for Metricbeat.

The same state metrics are reported to X-Pack Monitoring. The data is reported every minute by default.

Additional changes:

* Introduce ArraySlice support in metrics
* Improve namespace handling to prevent races
* Add monitoring reporter for state endpoint
* Add config options for reporting frequency. The old config option was removed. I don't consider this a breaking change as the old option was never document and not intended to be changed.

### DIFF
--- a/heartbeat/heartbeat.reference.yml
+++ b/heartbeat/heartbeat.reference.yml
@@ -1252,7 +1252,7 @@ logging.files:
   # never, once, and freely. Default is never.
   #ssl.renegotiation: never
 
-  #stats.period: 10s
+  #metric.period: 10s
   #state.period: 1m
 
 #================================ HTTP Endpoint ======================================
--- a/heartbeat/heartbeat.reference.yml
+++ b/heartbeat/heartbeat.reference.yml
@@ -1252,7 +1252,7 @@ logging.files:
   # never, once, and freely. Default is never.
   #ssl.renegotiation: never
 
-  #metric.period: 10s
+  #metrics.period: 10s
   #state.period: 1m
 
 #================================ HTTP Endpoint ======================================
--- a/heartbeat/heartbeat.reference.yml
+++ b/heartbeat/heartbeat.reference.yml
@@ -1252,6 +1252,9 @@ logging.files:
   # never, once, and freely. Default is never.
   #ssl.renegotiation: never
 
+  #stats.period: 10s
+  #state.period: 1m
+
 #================================ HTTP Endpoint ======================================
 # Each beat can expose internal metrics through a HTTP endpoint. For security
 # reasons the endpoint is disabled by default. This feature is currently experimental.
--- a/libbeat/_meta/config.reference.yml
+++ b/libbeat/_meta/config.reference.yml
@@ -1038,7 +1038,7 @@ logging.files:
   # never, once, and freely. Default is never.
   #ssl.renegotiation: never
 
-  #stats.period: 10s
+  #metric.period: 10s
   #state.period: 1m
 
 #================================ HTTP Endpoint ======================================
--- a/libbeat/_meta/config.reference.yml
+++ b/libbeat/_meta/config.reference.yml
@@ -1038,6 +1038,9 @@ logging.files:
   # never, once, and freely. Default is never.
   #ssl.renegotiation: never
 
+  #stats.period: 10s
+  #state.period: 1m
+
 #================================ HTTP Endpoint ======================================
 # Each beat can expose internal metrics through a HTTP endpoint. For security
 # reasons the endpoint is disabled by default. This feature is currently experimental.
--- a/libbeat/_meta/config.reference.yml
+++ b/libbeat/_meta/config.reference.yml
@@ -1038,7 +1038,7 @@ logging.files:
   # never, once, and freely. Default is never.
   #ssl.renegotiation: never
 
-  #metric.period: 10s
+  #metrics.period: 10s
   #state.period: 1m
 
 #================================ HTTP Endpoint ======================================
--- a/libbeat/api/server.go
+++ b/libbeat/api/server.go
@@ -41,6 +41,7 @@ func Start(cfg *common.Config) {
 
 		// register handlers
 		mux.HandleFunc("/", rootHandler())
+		mux.HandleFunc("/state", stateHandler)
 		mux.HandleFunc("/stats", statsHandler)
 		mux.HandleFunc("/dataset", datasetHandler)
 
@@ -61,10 +62,19 @@ func rootHandler() func(http.ResponseWriter, *http.Request) {
 
 		w.Header().Set("Content-Type", "application/json; charset=utf-8")
 
-		data := monitoring.CollectStructSnapshot(monitoring.GetNamespace("state").GetRegistry(), monitoring.Full, false)
+		data := monitoring.CollectStructSnapshot(monitoring.GetNamespace("info").GetRegistry(), monitoring.Full, false)
 
 		print(w, data, r.URL)
 	}
+}
+
+// stateHandler reports state metrics
+func stateHandler(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json; charset=utf-8")
+
+	data := monitoring.CollectStructSnapshot(monitoring.GetNamespace("state").GetRegistry(), monitoring.Full, false)
+
+	print(w, data, r.URL)
 }
 
 // statsHandler report expvar and all libbeat/monitoring metrics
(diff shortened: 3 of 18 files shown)


ruflin:
@pickypg @tsullivan I think we should have a different reporting interval for the beat state. Any recommendations on the interval?

metricbeat/mb/module/runner.go
```diff
@@ -56,9 +70,44 @@ func (mr *runner) Stop() {
...
+}
+
+func removeModule(module string) {
+	moduleListLock.Lock()
+	defer moduleListLock.Unlock()
+	moduleList[module]--
+}
+
+func reportModules(m monitoring.Mode, V monitoring.Visitor) {
+
```
exekias: empty line here

libbeat/api/server.go
```diff
@@ -24,6 +24,7 @@ func Start(cfg *common.Config) {
 
 		// register handlers
 		mux.HandleFunc("/", rootHandler())
+		mux.HandleFunc("/state", stateHandler)
```
exekias: I wonder if it would make more sense to call this `/modules`, do you plan to add anything else?

ruflin: I'm thinking of also adding information about the datasets, the output used. But not sure where it will evolve into. We probably need soon a more general discussion about api endpoint naming and structure.

ruflin: yes, API endpoints are not documented so far and we can still change them.

libbeat/monitoring/report/elasticsearch/elasticsearch.go
```diff
@@ -217,6 +216,7 @@ func (r *reporter) snapshotLoop() {
 		r.client.Publish(beat.Event{
 			Timestamp: ts,
 			Fields:    fields,
+			Meta:      common.MapStr{"type": "beats_" + namespace, "interval_ms": period * time.Millisecond},
```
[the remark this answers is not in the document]
pickypg: Unfortunately not. The parameter is required to be passed as a querystring field.

pickypg:
> I think we should have a different reporting interval for the beat state. Any recommendations on the interval?

Perhaps once every 5 minutes?

libbeat/monitoring/report/elasticsearch/elasticsearch.go
```diff
@@ -111,46 +96,84 @@ func makeReporter(beat beat.Info, cfg *common.Config) (report.Reporter, error) {
...
+		return pipeline.New(
+			beat,
+			monitoring,
+			queueFactory, out, pipeline.Settings{
+				WaitClose:     0,
+				WaitCloseMode: pipeline.NoWaitOnClose,
+			})
+	}
+
+	statsPipeline, err := getPipeline("stats", config.StatsPeriod)
```
ruflin: @urso This code works but is not very nice. The reason I have to define 2 different pipelines is because the interval params is different and the monitoring registry conflicts. I would prefer to have only 1 pipeline but be able to set the params on request. Are there other options?

ruflin:
@urso Ready for a round of reviews.

@pickypg I selected 1m as default as I felt only 5 minutes is too low in fast changing environments. But it's a tricky one. The good thing is at the moment there is not much data. Perhaps we have to reconsider later.

libbeat/monitoring/namespace.go
```diff
@@ -1,34 +1,26 @@
 package monitoring
 
-import (
-	"sync"
-)
+import "sync"
 
-var namespaces = struct {
```
ruflin: @jsoriano This code change as I also already had it change in this PR. Let me now if my implementation also looks good to you.

jsoriano: LGTM

libbeat/monitoring/report/elasticsearch/client.go
```diff
@@ -80,15 +80,42 @@ func (c *publishClient) Close() error {
...
+		// Extract potential additional params
+		p, err := event.Content.Meta.GetValue("params")
+		if err == nil {
+			p2, ok := p.(map[string]string)
+			if ok {
+				for k, v := range p2 {
+					params[k] = v
+				}
+			}
+		}
```
urso: how about us copying/allocating contents if required only:

```
var params, userParams, clientParams map[string]string
clientParams = c.params
if p, err := event.Content.Meta.GetValue("params"); err == nil {
  if tmp, ok := p.(map[string]string); ok {
    userParams = tmp
  }
}

switch {
  case len(userParams) == 0:
    params = clientParams
  case len(clientParams) == 0:
    params = userParams
  default:
    params = make(map[string]string, len(userParams) + len(clientParams))
    for k, v := range clientParams {
      params[k] = v
    }
    for k, v := range userParams {
      params[k] = v
    }
}
```

libbeat/_meta/config.reference.yml
```diff
@@ -982,6 +982,9 @@ logging.files:
   # never, once, and freely. Default is never.
   #ssl.renegotiation: never
 
+  #stats.period: 10s
+  #state.period: 1m
```
urso: `stats/state`? I wonder how many users will get confused by this (or how many typos we will see). How about `metrics.period` and `state.period`?

ruflin:
@pickypg It seems because of the interval param that we have to send each type separately and can't send it in one bulk request. I wonder if moving forward we could make the `interval` query param optional and instead also accept it as part of the event. Like this bulk requests would work as expected.

ruflin:
@urso This should be ready for an other round of reviews.

libbeat/monitoring/report/elasticsearch/client.go
```diff
@@ -79,23 +79,58 @@ func (c *publishClient) Close() error {
 
 func (c *publishClient) Publish(batch publisher.Batch) error {
 	events := batch.Events()
-	bulk := make([]interface{}, 0, 2*len(events))
+	var failed []publisher.Event
+	var reason error
+	var params map[string]string
```
urso: we still need the global 'params' variable? It's re-initialized in the loop for every bulk request.

libbeat/monitoring/report/elasticsearch/client.go
urso: Is c.params and params via 'meta' always configured? If not, we can reduce some copying here:

```
eventParams := func(meta common.MapStr) map[string]string {
  tmp, err := event.Content.Meta.GetValue("params")
  if err != nil {
    return nil
  }
  p, ok := tmp.(map[string]string)
  if !ok {
    return nil
  }
  return p
}

cpyKV := func(to map[string]string, from map[string]string) {
  for k, v := range from {
    to[k] = v
  }
}

params = c.params
if userParams := eventParams(event.Content.Meta); len(userParams) > 0 {
  if len(c.params) > 0 {
    params = make(map[string]string, len(params) + len(userParams))
    cpyKV(params, c.params)
    cpyKV(params, userParams)
  } else {
    params = userParams
  }
}
```

ruflin: As each event contains the interval param in meta and it's required at the moment, I think we have to reprocess it every time.

What do you mean by we `loose` params? As it's reprocessed every time and we don't modify c.params, all "base" params should always be there?

libbeat/monitoring/report/elasticsearch/client.go
ruflin: no, this resolved also my comment above.

heartbeat/heartbeat.reference.yml
```diff
@@ -1218,6 +1218,9 @@ logging.files:
   # never, once, and freely. Default is never.
   #ssl.renegotiation: never
 
+  #metrics.period: 10s
```
ruflin: @urso I have second thoughts if we even want to add this to the config file, meaning should we even recommend users to change it?

ruflin:
@urso Ready for an other round.

I will follow up with the ES team to potentially change the requirement on the xpack _bulk endpoint to simplify this implementation.

ruflin:
As soon as this PR is merged, the monitoring template in Elasticsearch has to be updated.

heartbeat/heartbeat.reference.yml
urso: It's the reference config. I don't really see a problem in configuring the interval. Allows users to reduce number of documents for monitoring if required.

ruflin:
PR for template can be found here: https://github.com/elastic/elasticsearch/pull/31809

PR #7078, opened 2018-05-11
Enhance testing of Metricbeat Elasticearch module

ruflin:
This PR is to have a few new ideas on how we could test across versions without running the services itself. This approach here does only work for services with an http endpoint.

* Standardise on 2 methods which are used to test the mapping
* Add http endpoint tests for `node` metricset. Could be used further.

### DIFF
--- a/metricbeat/module/elasticsearch/index/data.go
+++ b/metricbeat/module/elasticsearch/index/data.go
@@ -32,17 +32,25 @@ var (
 	}
 )
 
-func eventsMapping(r mb.ReporterV2, info elasticsearch.Info, content []byte) {
+func eventsMapping(r mb.ReporterV2, info elasticsearch.Info, content []byte) []error {
 
 	var indicesStruct struct {
 		Indices map[string]map[string]interface{} `json:"indices"`
 	}
 
-	json.Unmarshal(content, &indicesStruct)
+	err := json.Unmarshal(content, &indicesStruct)
+	if err != nil {
+		r.Error(err)
+		return []error{err}
+	}
 
+	var errs []error
 	for name, index := range indicesStruct.Indices {
 		event := mb.Event{}
-		event.MetricSetFields = eventMapping(index)
+		event.MetricSetFields, err = schema.Apply(index)
+		if err != nil {
+			errs = append(errs, err)
+		}
 		// Write name here as full name only available as key
 		event.MetricSetFields["name"] = name
 		event.RootFields = common.MapStr{}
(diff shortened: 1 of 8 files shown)


metricbeat/module/elasticsearch/index/data.go
```diff
@@ -32,17 +32,25 @@ var (
 	}
 )
 
-func eventsMapping(r mb.ReporterV2, info elasticsearch.Info, content []byte) {
+func eventsMapping(r mb.ReporterV2, info elasticsearch.Info, content []byte) []error {
```
jsoriano: This looks a bit confusing to me, do we need two ways of reporting errors? (One with the reporter api `r.Error`, and another one returning the list of errors).

metricbeat/module/elasticsearch/testing.go
```diff
@@ -0,0 +1,70 @@
...
+	// Makes sure glob matches at least 1 file
+	assert.True(t, len(files) > 0)
+
+	for _, f := range files {
+		t.Run(f, func(t *testing.T) {
+			input, err := ioutil.ReadFile(f)
+			assert.NoError(t, err)
+
+			reporter := &mbtest.CapturingReporterV2{}
+			errors := mapper(reporter, input)
```
jsoriano: I think we could also obtain the errors with `reporter.GetErrors()` and then we wouldn't need to also return them.

metricbeat/module/elasticsearch/testing.go
```diff
@@ -0,0 +1,70 @@
+// +build !integration
```
jsoriano: We could also move this common testing code to its own package (under `metricbeat/module/elasticsearch/testing/`) so it is only imported in tests and we don't need to add build tags. I was thinking about doing something like this for the RabbitMQ test server.

metricbeat/module/elasticsearch/index/data.go
ruflin: It's two different types of errors. The ones with `r.Error` will be send as error events to Elasticsearch. The `[]error` here is only used for testing. I agree, it's not optimal and we should come up with a cleaner approach in the future.

metricbeat/module/elasticsearch/testing.go
ruflin: +1 on moving it in it's own package in the future and also generalise it. There is already a `testing` package in Metricbeat. But before doing that i would check if the approach here fits other modules and have at least 2-3 more similar implementation to then take it out.

metricbeat/module/elasticsearch/testing.go
ruflin: See comment above, it's not the same errors.

metricbeat/module/elasticsearch/index/data.go
jsoriano: Ok

Issue #6642, opened 2018-03-22
Some files in the IIS and Nginx modules show "prospector" instead of "input"


dedemorton:
It looks like the renaming from prospector > input was not made in some of the files for the IIS and Nginx modules (not sure why):

* In the filebeat.reference.yml file, the IIS and Nginx module sections still use prospector instead of input. for example: https://github.com/elastic/beats/blob/master/filebeat/filebeat.reference.yml#L133

* The manifest files for IIS and Nginx still refer to prospectors. For example: https://github.com/elastic/beats/blob/master/filebeat/module/iis/access/manifest.yml#L13

dedemorton:
@ph Can we still fix this for 6.3, or is it too late? I think users are going to be confused if they see a mixture of "input" and "prospector" in our reference config files. I think both will work (right?) but the docs are going to be confusing to anyone who is looking for info about prospector settings.

dedemorton:
hmm...looks like some of the manifest files still refer to prospectors, too.

ph:
I will make a PR tomorrow for that I think we can get that in before the release.

ph:
both will work , but I prefer that we remove any confusion good catch.

ph:
@dedemorton  I also found theses mentions in the kubernetes examples files and created https://github.com/elastic/beats/pull/7080

PR #7079, opened 2018-05-11
Update modules s/prospector(s)/input(s)

ph:
Some modules and yaml reference were still referencing prospectors
instead of inputs, I have also updated the fileset manifest to make sure
it uses the new naming.

closes: #6642

### DIFF
--- a/filebeat/filebeat.reference.yml
+++ b/filebeat/filebeat.reference.yml
@@ -130,9 +130,9 @@ filebeat.modules:
     # Filebeat will choose the paths depending on your OS.
     #var.paths:
 
-    # Prospector configuration (advanced). Any prospector configuration option
+    # input configuration (advanced). Any input configuration option
     # can be added under this section.
-    #prospector:
+    #input:
 
   # Error logs
   #error:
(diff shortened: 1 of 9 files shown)


filebeat/filebeat.reference.yml
```diff
@@ -130,9 +130,9 @@ filebeat.modules:
     # Filebeat will choose the paths depending on your OS.
     #var.paths:
 
-    # Prospector configuration (advanced). Any prospector configuration option
+    # input configuration (advanced). Any input configuration option
```
ruflin: Should input here be upper case?

filebeat/module/iis/_meta/config.reference.yml
```diff
@@ -7,9 +7,9 @@
     # Filebeat will choose the paths depending on your OS.
     #var.paths:
 
-    # Prospector configuration (advanced). Any prospector configuration option
+    # input configuration (advanced). Any input configuration option
```
ruflin: Upper case?

ph:
@ruflin yes it should be uppercase, are you fine in merging that for 6.3?

ph:
@ruflin updated, let's wait for the green lantern.

ph:
actually lets wait until 6.3 is unfrozen.

ruflin:
Yes, lets get that into 6.3.1 as it's not urgent.

PR #7080, opened 2018-05-11
Update kubernetes examples configuration to use inputs

ph:
When we made the rename of the prospectors to input we overlooked the update of
the kubernetes examples files and readme.

Ref #6642

### DIFF
--- a/deploy/kubernetes/filebeat-kubernetes.yaml
+++ b/deploy/kubernetes/filebeat-kubernetes.yaml
@@ -10,10 +10,10 @@ metadata:
 data:
   filebeat.yml: |-
     filebeat.config:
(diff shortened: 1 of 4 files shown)


ph:
@exekias I am not too familiar with theses files and how to test them, but they were using `prospectors` instead of ` inputs`, can you review theses changes?

exekias:
We test the manifest can be loaded to a mock kubernetes, so it has a valid format, but we don't really do an integration test, which would require running an actual kubernetes cluster (probably?).

I'm open to ideas to improve this, it would be nice to have some e2e testing here :innocent:

ph:
@exekias we will get there eventually :)

PR #7081, opened 2018-05-11
Add seccomp whitelist policies for each Beat

### LINKED ISSUE #5213: Add syscall filtering via seccomp for security
The Beat process on Linux can drop privileges to perform specific syscalls (like execve, fork, clone) using a seccomp-bpf filter. This is one-way operation where the process drops privileges and cannot restore them. This limits the damage that can be done if there ever would be a vulnerability.

There is a kernel version requirement, but we can test if the feature is available use it when possible.

I wrote a POC that uses google/kafel, a C library that generates the BPF syscall filter from a policy string. We could statically link with the library (we already use cgo in our Beats since 6.0).

- [Proof-of-concept GIST](https://gist.github.com/andrewkroh/1c8db5a2321ba37a728c4a453f156b48)
- [google/kafel](https://github.com/google/kafel)

When there is a policy violation the kernel kills the thread rather than the whole process. When it does this an audit event is generated.

FYI you can use `/proc/[pid]/status` to see if the process has a seccomp filter applied. Look for the `Seccomp:` line and a value > 0.

We can audit the syscalls used by a Beat by using `strace -c [beat cmd]` then develop a policy based on that data.

andrewkroh:
Add seccomp whitelist policy
    
This PR adds to the seccomp enhancement in #6832 by adding a whitelist policy.
    
The policies are derived from a combination of static analysis of the object code and runtime profiling (to see what the cgo linked libraries are doing). The policies also account for the system calls used while running the system tests with coverage and race detection.
    
Closes #5213


CHANGELOG.asciidoc
```diff
@@ -125,8 +125,8 @@ https://github.com/elastic/beats/compare/v6.2.3...master[Check the HEAD diff]
 - Do not log errors if X-Pack Monitoring is enabled but Elastisearch X-Pack is not. {pull}6627[6627]
 - Add rename processor. {pull}6292[6292]
 - Add IP-addresses and MAC-addresses to add_host_metadata. {pull}6878[6878]
-- Add a default seccomp (secure computing) filter on Linux that prohibits
-  execve, execveat, fork, and vfork syscalls. A custom policy can be configured. {issue}5213[5213]
+- Added a seccomp (secure computing) filter on Linux that whitelists the
```
ruflin: Was this release before? Mainly asking because the changelog was updated instead of a new one.

libbeat/common/seccomp/README.md
```diff
@@ -0,0 +1,103 @@
+# Package seccomp
```
ruflin: Great README

libbeat/common/seccomp/seccomp.go
```diff
@@ -0,0 +1,115 @@
...
+		return
+	}
+
+	filter := seccomp.Filter{
+		NoNewPrivs: true,
+		Flag:       seccomp.FilterFlagTSync,
+		Policy:     *p,
+	}
+
+	// TODO (andrewkroh): For testing only, remove.
```
ruflin: Should be removed now or later?

CHANGELOG.asciidoc
andrewkroh: This hasn't been released yet.

andrewkroh:
I added the review label and removed in-progress.

libbeat/common/seccomp/README.md
```diff
@@ -0,0 +1,112 @@
...
+| seccomp-profiler-allow.txt     | A list of system calls that are always included in the resulting policy. System calls that are made from CGO or linked libraries (such as libpthread) are not detected by seccomp-profiler and must be manually added to the list. Tools like `strace -c` and Auditbeat (where event.action=violated-seccomp-policy) can be used to find missing system calls. System calls from any architecture may be added to the list because it is automatically filtered based on architecture. |
+| policy_linux_$arch.go          | This is a default policy provided by libbeat. It is used when a Beat does not register its own application specific policy with `seccom.MustRegisterPolicy()` (e.g. community Beats).|
```
adriansr: Nit: seccomp instead of seccom

andrewkroh: Fixed.

PR #7082, opened 2018-05-11
[WIP] Event scheduling interfaces

urso:
The changes in this PR add support for configurable event scheduling in libbeat. Event scheduling will be applied when a beat publishes an event. Users configure policies like rate limiting and others, which will be applied to event publishing, before the event processors are executed.

beats pipeline with scheduling

Policies can be local or have shared state. Thanks to shared state, policies like rate limiting can affect a group of workers/publishers.

A policy implementation is free to create back-pressure by blocking event publishing or drop/modify events.

Besides introducing interfaces and configuration support, a simple `ratelimit` policy is provided in this PR. The current implementation is only for demonstration purposes (sample code) and not recommended for production. The rate limiter supports full local rate limiting (per `beat.Client` connection), but also a shared mode. With shared mode, the total bandwidth enforced by the rate limiter will be shared between all `beat.Client` instances.

Scheduling can be defined locally per input/module instance, per `group` and even globally, per beat. A scheduling group can also have a parent group. The policies are concatenated in this order: local, group, [parent group], global.

Scheduling 'groups' are used to configured shared resources between multiple workers/data sources.

Sample config:

```
filebeat.inputs:
- ...
  scheduling.policies:
  - ratelimit.events_per_second: 800
- ...
  scheduling.group: A
  scheduling.policies:
  - ratelimit:
      events_per_second: 100
      shared: false
- ...
  scheduling.group: A
  - ratelimit:
      events_per_second: 500
      shared: false

# per group policies
scheduling.groups:
  A.policies:
  - ratelimit.events_per_second: 800

# global policies
scheduling.policies:
- ratelimit:
    events_per_second: 1000
```

With this sample configuration, the beats global bandwidth is limited to 1000eps. The first input's bandwidth is limited to 800eps. If the first input creates multiple workers/harvesters, the bandwidth of 800eps will be shared by all harvesters. Due to `shared: false`, each harvester created by input 2 will have a bandwidth limit of 100eps. Input 3 enforces a limit of 500eps per harvester. Due to input 2 and 3 joining the scheduling group `A`, the total shared bandwidth of all harvesters started by input 2 and 3 is limited to 800eps.

Tasks:
- [x] introduce event scheduling
- [x] support for defining shared scheduling groups
- [x] add support to filebeat inputs to configure local policies
- [ ] add support to filebeat modules to configure local policies
- [x] add support to metricbeat/auditbeat modules/metricsets to configure local policies
- [ ] add support to heartbeat to configure local policies
- [ ] add support to packetbeat to configure local policies
- [ ] support for passing additional settings from local policies to parent groups on `Connect`.

### DIFF
--- a/libbeat/publisher/scheduling/ratelimit/config.go
+++ b/libbeat/publisher/scheduling/ratelimit/config.go
@@ -3,7 +3,7 @@ package ratelimit
 import "errors"
 
 type config struct {
-	EventsPerSecond uint `config:"events_pre_second" validate:"min=1"`
+	EventsPerSecond uint `config:"events_per_second" validate:"min=1"`
 	Shared          bool `config:"shared"`
 }
 
--- a/libbeat/publisher/scheduling/ratelimit/config.go
+++ b/libbeat/publisher/scheduling/ratelimit/config.go
@@ -1,0 +1,22 @@
+package ratelimit
+
+import "errors"
+
+type config struct {
+	EventsPerSecond uint `config:"events_pre_second" validate:"min=1"`
+	Shared          bool `config:"shared"`
+}
+
+func defaultConfig() config {
+	return config{
+		Shared: true,
+	}
+}
+
+func (c *config) Validate() error {
+	if c.EventsPerSecond == 0 {
+		return errors.New("events_per_second must be > 0")
+	}
+
+	return nil
+}
--- a/filebeat/channel/factory.go
+++ b/filebeat/channel/factory.go
@@ -4,6 +4,7 @@ import (
 	"github.com/elastic/beats/libbeat/beat"
 	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/processors"
+	"github.com/elastic/beats/libbeat/publisher/scheduling"
 )
 
 type OutletFactory struct {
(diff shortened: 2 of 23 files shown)


ruflin:
As the rate limiting is applied before the processors, this means if a processor would drop all events which contain a field `DEBUG` and this would be 99 out of 100 fields and the rate limit would be lower, it would be throttled even though the outcome would be just one event?

libbeat/publisher/scheduling/ratelimit/config.go
```diff
@@ -0,0 +1,22 @@
+package ratelimit
+
+import "errors"
+
+type config struct {
+	EventsPerSecond uint `config:"events_per_second" validate:"min=1"`
```
ruflin: Should we make the time unit configurable instead of hardcoding it "per second"? It would mean to split up this config into 2 parts, one for the duration and one for the number of events.

urso:
Closing. No progress for some time. If there is still some interest, please contact us by commenting on this PR.

slmingol:
Curious where this ended up? See the PR was closed but was this feature enabled or is it still waiting on this PR?

nebriv:
I'd be very interested in a sampling rate function as well.

PR #7085, opened 2018-05-12
Clarify docs to indicate where processors are valid in the config

dedemorton:
Replaces https://github.com/elastic/beats/pull/7038

### DIFF
--- a/libbeat/docs/processors-using.asciidoc
+++ b/libbeat/docs/processors-using.asciidoc
@@ -32,6 +32,20 @@ condition is present, then the action is executed only if the condition is
 fulfilled. If no condition is passed, then the action is always executed.
 * `<parameters>` is the list of parameters to pass to the processor.
 
+
+You can also scope processors to individual plugin prospectors by placing them under the prospector definition, this is common when wanting to process only logs from modules such as docker, but having more than one prospector enabled: 
+
+[source,yaml]
+------
+- type: <propector_type>
+  processors:
+   - <processor_name>:
+     when:
+        <condition>
+     <parameters>
+...
+------
+
 [[processors]]
 ==== Processors
 
--- a/libbeat/docs/processors-using.asciidoc
+++ b/libbeat/docs/processors-using.asciidoc
@@ -1,25 +1,24 @@
 [[defining-processors]]
 === Define processors
 
-You define processors in the +{beatname_lc}.yml+ file to filter and enhance the
-data before sending events to the configured output.
-
-To define a processor, you specify the processor name, an optional condition,
-and a set of parameters:
+You can use processors to filter and enhance data before sending it to the
+configured output. To define a processor, you specify the processor name, an
+optional condition, and a set of parameters:
 
 [source,yaml]
 ------
 processors:
- - <processor_name>:
-     <parameters>
-     when:
-       <condition>
- - <processor_name>:
-     <parameters>
-     when:
-       <condition>
-...
+- <processor_name>:
+    when:
+      <condition>
+    <parameters>
 
+- <processor_name>:
+    when:
+      <condition>
+    <parameters>
+
+...
 ------
 
 Where:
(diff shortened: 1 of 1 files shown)


libbeat/docs/processors-using.asciidoc
```diff
@@ -32,6 +32,26 @@ condition is present, then the action is executed only if the condition is
 fulfilled. If no condition is passed, then the action is always executed.
 * `<parameters>` is the list of parameters to pass to the processor.
 
+ifeval::["{beatname_lc}"=="filebeat"]
```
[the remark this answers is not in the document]
dedemorton: @andrewkroh I'll need to add some conditional coding to this topic because we probably need examples for each beat. Incidentally, when I was looking into what to add here, I looked at the reference.yml files for each beat to figure out where  processors are valid. It's not clear that they are supported at the module level for auditbeat, metricbeat, and winlogbeat.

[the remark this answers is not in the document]
andrewkroh: I agree that the reference configs are not clear about where processors can be used. Here's where they can be used. The list is derived from this [search](https://github.com/elastic/beats/search?l=Go&q=processors.PluginConfig).

- auditbeat.modules
- filebeat.inputs
- heartbeat.monitors
- metricbeat.modules
- packetbeat.flows
- packetbeat.protocols

dedemorton:
@andrewkroh This is ready for another review. Sorry it took me so long to get back to it...kept following down my to-do list. Easiest way to review this is probably to just build the docs for each beat.

libbeat/docs/processors-using.asciidoc
```diff
@@ -1,25 +1,24 @@
 [[defining-processors]]
 === Define processors
 
-You define processors in the +{beatname_lc}.yml+ file to filter and enhance the
-data before sending events to the configured output.
-
-To define a processor, you specify the processor name, an optional condition,
-and a set of parameters:
+You can use processors to filter and enhance data before sending it to the
```
dedemorton: Now that we have some configs in the modules.d directory, telling users that they specify processors in the {beatname_lc}.yml file seems misleading. I think it's pretty clear that these are configuration settings by the context, no?

libbeat/docs/processors-using.asciidoc
```diff
@@ -32,6 +31,130 @@ condition is present, then the action is executed only if the condition is
...
+  - name: <network_shipper_name>
+  processors:
+  - <processor_name>:
+      when:
+        <condition>
+      <parameters>
+----
+endif::[]
+
+// REVIEWERS: I'm not sure if the syntax of the winlog example is correct.
```
webmat: - [ ] Let's make sure this doesn't get forgotten in the file final file 😄

dedemorton: that's the plan.

dedemorton: Note that I've tested the syntax of the other examples, but not winlogbeat, so I would appreciate it if someone would confirm.

webmat: @dedemorton Note that Andrew is on vacation right now. He'll be back next week

libbeat/docs/processors-using.asciidoc
```diff
@@ -1,25 +1,24 @@
...
+
+- <processor_name>:
+    when:
+      <condition>
+    <parameters>
 
+...
 ------
 
 Where:
```
andrewkroh: Below, in `<when: condition>`, should that be just `<condition`?

libbeat/docs/processors-using.asciidoc
```diff
@@ -32,6 +31,130 @@ condition is present, then the action is executed only if the condition is
...
+        <condition>
+      <parameters>
+----
+endif::[]
+ifeval::["{beatname_lc}"=="winlogbeat"]
+[source,yaml]
+----
+winlogbeat.event_logs:
+  - name: <network_shipper_name>
+  processors:
```
andrewkroh: The indentation is off a little for the Winlogbeat example. `processors` needs to be at the same level as `name`.

```
winlogbeat.event_logs:
- name: <network_shipper_name>
  processors:
  - <processor_name>: ...

dedemorton:
@andrewkroh I've made the changes that you requested. Thanks!

PR #7091, opened 2018-05-14
Set default shards to 1 in all our configs

tsg:
Elasticsearch is moving to having 1 shard by default: https://github.com/elastic/elasticsearch/pull/30539

We already set shards to 1 in our Metricbeat and Heartbeat configurations, but we were
explicitly setting it to 3 in the other configs. I changed to be 1 everywhere. I went
for a change that still contains the number of shards in the Beat config, rather than
just rely on the ES default, so that Beats 7.0 behave the same even when talking to ES 6.x.

### DIFF
--- a/auditbeat/_meta/common.p2.yml
+++ b/auditbeat/_meta/common.p2.yml
@@ -1,6 +1,6 @@
 
 #==================== Elasticsearch template setting ==========================
 setup.template.settings:
-  index.number_of_shards: 3
+  index.number_of_shards: 1
   #index.codec: best_compression
   #_source.enabled: false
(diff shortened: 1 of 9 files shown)


tsg:
However, we probably only want this in 7.0, in which case we should evaluate on when is the right time to merge this. I added `in progress` and added it to the team meeting agenda to discuss.

ruflin:
@urso I promised to take this over some time ago but didn't get to it. I'm suggestion to move forward with what we have now in this PR.

@jsvd did you apply the changes on the LS side already?

jsvd:
the elasticsearch output for logstash uses `number_of_shard: 1` in the template that is installed if connecting to a 7.x cluster: https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/master/lib/logstash/outputs/elasticsearch/elasticsearch-template-es7x.json#L6

urso:
LGTM

tsg:
Closing in favour of #10430.

PR #7095, opened 2018-05-14
Cherry-pick #7080 to 6.3: Update kubernetes examples configuration to use inputs

ph:
Cherry-pick of PR #7080 to 6.3 branch. Original message: 

When we made the rename of the prospectors to input we overlooked the update of
the kubernetes examples files and readme.

Ref #6642

### DIFF
--- a/deploy/kubernetes/filebeat-kubernetes.yaml
+++ b/deploy/kubernetes/filebeat-kubernetes.yaml
@@ -10,10 +10,10 @@ metadata:
 data:
   filebeat.yml: |-
     filebeat.config:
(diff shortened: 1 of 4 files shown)


ph:
Wait until 6.3 is unfrozen.

exekias:
A note about freeze period and this. These files are not packaged with the release, they work in the same way as docs, so HEAD of 6.3 is what users see. In general you can treat them as you would treat docs

ph:
@exekias you are right! so feel free to merge if they are green.

PR #7097, opened 2018-05-15
Add x-pack data for Elasticsearch shard metricset

ruflin:
* Introduces GetNodeInfo method to fetch additional info about the node. This should become obsolete in the future.
* Refactor shard metricset to use module level hostParser and metricset.

The xpack feature works but will need further testing with new builds of Elasticsearch. The plan is to test all xpack metricsets together when they are all done and do further tweaks.

### DIFF
--- a/metricbeat/module/elasticsearch/shard/data.go
+++ b/metricbeat/module/elasticsearch/shard/data.go
@@ -11,17 +11,19 @@ import (
 
 var (
 	schema = s.Schema{
-		"state":   c.Str("state"),
-		"primary": c.Bool("primary"),
-		"node":    c.Str("node"),
-		"index":   c.Str("index"),
-		"shard":   c.Int("number"),
+		"state":           c.Str("state"),
+		"primary":         c.Bool("primary"),
+		"node":            c.Str("node"),
+		"index":           c.Str("index"),
+		"shard":           c.Int("number"),
+		"relocating_node": c.Str("relocating_node"),
 	}
 )
 
 type stateStruct struct {
 	ClusterName  string `json:"cluster_name"`
 	StateID      string `json:"state_uuid"`
+	MasterNode   string `json:"master_node"`
 	RoutingTable struct {
 		Indices map[string]struct {
 			Shards map[string][]map[string]interface{} `json:"shards"`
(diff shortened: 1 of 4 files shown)


metricbeat/module/elasticsearch/shard/data.go
```diff
@@ -11,17 +11,19 @@ import (
...
-		"primary": c.Bool("primary"),
-		"node":    c.Str("node"),
-		"index":   c.Str("index"),
-		"shard":   c.Int("number"),
+		"state":           c.Str("state"),
+		"primary":         c.Bool("primary"),
+		"node":            c.Str("node"),
+		"index":           c.Str("index"),
+		"shard":           c.Int("shard"),
+		"relocating_node": c.Str("relocating_node"),
```
ruflin: @pickypg This value is most of the time `null`. In these cases metricbeat just skips the value instead of sending `null`. Is that ok?

pickypg: Yes, that should be fine.

metricbeat/module/elasticsearch/shard/data_xpack.go
```diff
@@ -0,0 +1,72 @@
...
+
+	nodeInfo, err := elasticsearch.GetNodeInfo(m.HTTP, m.HostData().SanitizedURI+statePath, stateData.MasterNode)
+	if err != nil {
+		r.Error(err)
+		return
+	}
+
+	// This is currently needed because the cluser_uuid is `na` in stateData in case not the full state is requested.
+	// This needs to be brought up with the ES team.
+	clusterID, err := elasticsearch.GetClusterID(m.HTTP, m.HostData().SanitizedURI+statePath, stateData.MasterNode)
```
ruflin: @pickypg @tlrx You might now more here? On ES master the following request `http://localhost:9200/_cluster/state/version,master_node,routing_table?pretty` returns:

```
{
  "cluster_name" : "elasticsearch",
  "compressed_size_in_bytes" : 36170,
  "cluster_uuid" : "_na_",
  "version" : 59,
```

If I fetch `http://localhost:9200/_cluster/state/?pretty` I get

```
{
  "cluster_name" : "elasticsearch",
  "compressed_size_in_bytes" : 36170,
  "cluster_uuid" : "Ir3Dhzo-TluhYWEaz4UZlA",
  "version" : 59,
```

Is this a bug?

PR #7100, opened 2018-05-15
Refactor http.server metricset 

ruflin:
This PR cleans up the http.server metricset which fixes a bug but also has breaking changes.

Breaking changes:

* Before events were put under `http.http` prefix by default which was not intended. Now it is `http.server` by default.
* Config options for the metricset are prefixed now by `server.` as we have for other metricsets.

Further changes

* Metricset now uses PushReporterV2 interface
* System test added to confirm data structure
* Add more docs

### DIFF
--- a/metricbeat/module/http/server/_meta/docs.asciidoc
+++ b/metricbeat/module/http/server/_meta/docs.asciidoc
@@ -1,1 +1,15 @@
 This is the server metricset of the module http.
+
+Events sent to the http endpoint will be put by default under the `http.server` prefix. To change this use the `server.paths`
+config options. In the example below every request to `/foo` will be put under `http.foo`.
+
+["source","yaml",subs="attributes"]
+------------------------------------------------------------------------------
+- module: http
+  metricsets: ["server"]
+  host: "localhost"
+  port: "8080"
+  server.paths:
+    - path: "/foo"
+      namespace: "foo"
+------------------------------------------------------------------------------
--- a/metricbeat/module/http/server/config.go
+++ b/metricbeat/module/http/server/config.go
@@ -7,8 +7,8 @@ import (
 )
 
 type HttpServerConfig struct {
-	Paths       []PathConfig `config:"paths"`
-	DefaultPath PathConfig   `config:"default_path"`
+	Paths       []PathConfig `config:"server.paths"`
+	DefaultPath PathConfig   `config:"server.default_path"`
 }
 
 type PathConfig struct {
@@ -21,7 +21,7 @@ func defaultHttpServerConfig() HttpServerConfig {
 	return HttpServerConfig{
 		DefaultPath: PathConfig{
 			Path:      "/",
-			Namespace: "http",
+			Namespace: "server",
 		},
 	}
 }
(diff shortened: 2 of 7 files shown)


metricbeat/module/http/server/data.go
```diff
@@ -77,6 +77,7 @@ func (p *metricProcessor) Process(event server.Event) (common.MapStr, error) {
 		return nil, errors.New(fmt.Sprintf("Unsupported Content-Type: %s", contentType))
 	}
 
+	// QUESTION: What is the exact expected behaviour here if no list of paths is set
```
ruflin: @vjsamuel Could you comment here? It seems right now if I just run it the events end up under `http/http`. Not sure if that is expected.

vjsamuel:
```
func defaultHttpServerConfig() HttpServerConfig {
	return HttpServerConfig{
		DefaultPath: PathConfig{
			Path:      "/",
			Namespace: "http",
		},
	}
}
```
it just exposes an API on `/` where you can send the data and the default namespace would be http

ruflin: Ok, at the moment the default is `http/http` so this is a bug I assume.

metricbeat/module/http/server/config.go
```diff
@@ -7,8 +7,8 @@ import (
 )
 
 type HttpServerConfig struct {
-	Paths       []PathConfig `config:"paths"`
-	DefaultPath PathConfig   `config:"default_path"`
+	Paths       []PathConfig `config:"server.paths"`
+	DefaultPath PathConfig   `config:"server.default_path"`
```
exekias: LGTM, if you want to do a softer deprecation you could still read old settings before these, but I don't think this is a hard requirement.

metricbeat/module/http/server/_meta/docs.asciidoc
```diff
@@ -1 +1,15 @@
...
+
+["source","yaml",subs="attributes"]
+------------------------------------------------------------------------------
+- module: http
+  metricsets: ["server"]
+  host: "localhost"
+  port: "8080"
+  server.paths:
+    - path: "/foo"
+      namespace: "foo"
```
jsoriano: Cool

metricbeat/tests/system/test_http.py
```diff
@@ -40,6 +42,37 @@ def test_json(self):
...
+        self.render_config_template(modules=[{
+            "name": "http",
+            "metricsets": ["server"],
+        }])
+        proc = self.start_beat()
+        time.sleep(5)
+
+        self.wait_until(lambda: self.log_contains("Starting http server on localhost:8080"))
+
+        requests.post('http://localhost:8080/', json={'hello': 'world'}, headers={'Content-Type': 'application/json'})
```
jsoriano: This request should go to the value in `$HTTP_HOST` instead of to `localhost`.

ruflin:
PR updated with review comment.

jsoriano:
Failure in travis could be related https://travis-ci.org/elastic/beats/jobs/383064425#L4420

PR #7101, opened 2018-05-15
Update export_dashboards.go

hakan-carlsson:
Added support to include index-pattern in output JSON. Also added -quiet option to make it silent.


dev-tools/cmd/dashboards/export_dashboards.go
```diff
@@ -48,7 +48,7 @@ func ExtractIndexPattern(body []byte) ([]byte, error) {
 		if !ok {
 			return nil, fmt.Errorf("type key not found or not string")
 		}
-		if _type != "index-pattern" {
+		if _type != "index-pattern" || indexPattern == true {
```
kvch: Why not write `|| indexPattern` instead of `|| indexPattern == true`?

hakan-carlsson: Yes, works too. IMHO I like to be explicit.

dev-tools/cmd/dashboards/export_dashboards.go
```diff
@@ -68,7 +68,7 @@ func Export(client *http.Client, conn string, dashboard string, out string) erro
 	params.Add("dashboard", dashboard)
 
 	fullURL := makeURL(conn, exportAPI, params)
-	fmt.Printf("Calling HTTP GET %v\n", fullURL)
+	if ! quiet { fmt.Printf("Calling HTTP GET %v\n", fullURL) }
```
kvch: Could you please add new lines after `{` and `fullURL)`?

Most of the codebase does not use this format. So in order to keep the uniformity of the code, please change it.

dev-tools/cmd/dashboards/export_dashboards.go
```diff
@@ -95,7 +95,7 @@ func Export(client *http.Client, conn string, dashboard string, out string) erro
 
 	err = ioutil.WriteFile(out, body, 0666)
 
-	fmt.Printf("The dashboard %s was exported under the %s file\n", dashboard, out)
+	if ! quiet { fmt.Printf("The dashboard %s was exported under the %s file\n", dashboard, out) }
```
kvch: Same here with the newlines.
```
if !quiet {
    fmt.Printf("The dashboard %s was exported under the %s file\n", dashboard, out)
}
```

kvch:
jenkins test this

dev-tools/cmd/dashboards/export_dashboards.go
kvch: I checked other parts of the code. Although, I find it weird, other parts of the code does include bool checking using `== true`, so I am ok with letting this go.

dev-tools/cmd/dashboards/export_dashboards.go
hakan-carlsson: ok, point taken.

hakan-carlsson: made changes to indents, removed explicit =='s and removed explict bool

kvch:
jenkins. test this, please

kvch:
Could you please add your change to `CHANGELOG.asciidoc`?

kvch:
Thanks.

PR #7103, opened 2018-05-15
Move parts of metricset code to Elasticsearch module

ruflin:
This is inspired by https://github.com/elastic/beats/pull/7074. More work will follow in other PR's.

### DIFF
--- a/metricbeat/module/elasticsearch/metricset.go
+++ b/metricbeat/module/elasticsearch/metricset.go
@@ -9,7 +9,7 @@ import (
 
 const (
 	defaultScheme = "http"
-	pathConfigKey = "management_path_prefix"
+	pathConfigKey = "path"
 )
 
 var (
--- a/metricbeat/module/elasticsearch/metricset.go
+++ b/metricbeat/module/elasticsearch/metricset.go
@@ -1,0 +1,58 @@
+package elasticsearch
+
+import (
+	"github.com/elastic/beats/libbeat/common/cfgwarn"
+	"github.com/elastic/beats/metricbeat/helper"
+	"github.com/elastic/beats/metricbeat/mb"
+	"github.com/elastic/beats/metricbeat/mb/parse"
+)
+
+const (
+	defaultScheme = "http"
+	pathConfigKey = "management_path_prefix"
+)
+
+var (
+	// HostParser parses host urls for RabbitMQ management plugin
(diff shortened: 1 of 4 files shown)


ruflin:
jenkins, test this

metricbeat/module/elasticsearch/metricset.go
```diff
@@ -0,0 +1,58 @@
...
+import (
+	"github.com/elastic/beats/libbeat/common/cfgwarn"
+	"github.com/elastic/beats/metricbeat/helper"
+	"github.com/elastic/beats/metricbeat/mb"
+	"github.com/elastic/beats/metricbeat/mb/parse"
+)
+
+const (
+	defaultScheme = "http"
+	pathConfigKey = "management_path_prefix"
```
jsoriano: The `pathConfigKey` was `path` before. I used `management_path_prefix` in #7074 because is the terminology in RabbitMQ.

metricbeat/module/elasticsearch/node_stats/node_stats.go
```diff
@@ -2,74 +2,46 @@ package node_stats
...
-		return nil, err
-	}
-
-	if config.XPack {
-		cfgwarn.Experimental("The experimental xpack.enabled flag in elasticsearch/node_stats metricset is enabled.")
-	}
-
-	http, err := helper.NewHTTP(base)
+	// Get the stats from the local node
+	ms, err := elasticsearch.NewMetricSet(base, "/_nodes/_local/stats")
```
jsoriano: Move this path also to its own constant?

metricbeat/module/elasticsearch/metricset.go
ruflin: Ups, this is a leftover. Changed to `path`.

metricbeat/module/elasticsearch/node_stats/node_stats.go
ruflin: move to a constant. I normally only use constant if it's used more then at least once.

ruflin:
For the path change you mentioned: It's not something that is documented in the Elasticsearch module and I hope / assume it's not used as it's not something that can be changed itself in ES. I wonder if we should remove it completely?

jsoriano:
Merging it now, and letting the path option removal by now.

PR #7106, opened 2018-05-15
Move common RabbitMQ testing code to its own package

jsoriano:
New testing package provides a RabbitMQ mocked management API that can be used by all tests in the module.

Some additional assertions have been added to detect missing testing files.

### DIFF
--- a/metricbeat/scripts/generate_imports_helper.py
+++ b/metricbeat/scripts/generate_imports_helper.py
@@ -17,7 +17,8 @@ def get_importable_lines(go_beat_path, import_line):
         imported_lines.append(module_import)
 
         module_path = join(path, module)
-        metricsets = [m for m in listdir(module_path) if isdir(join(module_path, m)) and m not in ["_meta", "vendor"]]
+        ignore = ["_meta", "vendor", "mtest"]
+        metricsets = [m for m in listdir(module_path) if isdir(join(module_path, m)) and m not in ignore]
         for metricset in metricsets:
             metricset_name = "{}/{}".format(module, metricset)
             metricset_import = import_line.format(beat_path=go_beat_path, module="module", name=metricset_name)
--- a/metricbeat/module/rabbitmq/_meta/testdata/nodes_sample_response.json
+++ b/metricbeat/module/rabbitmq/_meta/testdata/nodes_sample_response.json
@@ -1,0 +1,339 @@
+[
+  {
+    "partitions": [],
+    "os_pid": "133",
+    "fd_total": 1048576,
+    "sockets_total": 943626,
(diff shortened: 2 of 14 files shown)


jsoriano:
I have moved all common code for integration tests to the `rabbitmq/testing` package too.

metricbeat/scripts/generate_imports_helper.py
```diff
@@ -17,7 +17,8 @@ def get_importable_lines(go_beat_path, import_line):
         imported_lines.append(module_import)
 
         module_path = join(path, module)
-        metricsets = [m for m in listdir(module_path) if isdir(join(module_path, m)) and m not in ["_meta", "vendor"]]
+        ignore = ["_meta", "vendor", "testing"]
+        metricsets = [m for m in listdir(module_path) if isdir(join(module_path, m)) and m not in ignore]
```
jsoriano: @ruflin @exekias we are going to have to make a decision about where to place the common testing code for modules. If we place it as a subdirectory of the module then we have to add an exception here so it is not imported, and then the module has to be named always the same in all modules.

If we want to avoid collisions with `testing` package maybe we could call this `mstest` (ms from metricset) or something like this as in principle two metricsets are only going to import this package from its own module.

If we decide to place these packages in other subdirectory we should also decide where to place it, an option could be under `_meta`, but I don't like it so much. In any case I'd place them inside the module.

Opinions?

(for context, related with discussion here https://github.com/elastic/beats/pull/7059#discussion_r188599605)

jsoriano: (Btw, I love that tests have detected this issue in this PR :smile:)

jsoriano:

metricbeat/scripts/generate_imports_helper.py
ruflin: I wonder if under `_meta` would even work as I remember something golang ignores everything which starts with and underline.

What about `mtest` instead of `mstest` as it's the the things which are shared across the module? Reason I prefer `mtest` or `mstest` over `testing` is one the conflict and the other part is that I think the likelyhood of having a conflicting metricset name is lower.

ruflin: I think the reason the tests catched this is because we had some issue around it in the past ;-)

jsoriano: Ok, I'm good with `mtest` too.

ruflin: lets move forward with it and fix it when we break things :-)

jsoriano: Renamed to `mtest`.

PR #7107, opened 2018-05-15
Metricbeat: Set guest as default user in RabbitMQ module

jsoriano:
Follow-up of https://github.com/elastic/beats/pull/6908#discussion_r187912037

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -247,6 +247,7 @@ https://github.com/elastic/beats/compare/v6.2.3...master[Check the HEAD diff]
 - Add shard metricset to Elasticsearch module. {pull}7006[7006]
 - Add apiserver metricset to Kubernetes module. {pull}7059[7059]
 - Add maxmemory to redis info metricset. {pull}7127[7127]
+- Set guest as default user in RabbitMQ module. {pull}7107[7107]
 
 *Packetbeat*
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.asciidoc
```diff
@@ -236,6 +236,7 @@ https://github.com/elastic/beats/compare/v6.2.3...master[Check the HEAD diff]
 - Add Elasticsearch index_summary metricset. {pull}6918[6918]
 - Add config option `management_path_prefix` for RabbitMQ module to configure management plugin path prefix {issue}6875[6875] {pull}7074[7074]
 - Add shard metricset to Elasticsearch module. {pull}7006[7006]
+- Set guest as default user in RabbitMQ module. {pull}7107[7107]
```
ruflin: Sounds a bit like a breaking change. Or was it before a bug?

jsoriano: I think it is an enhancement for the out-of-the-box case, it can be breaking if someone has managed to setup the RabbitMQ endpoint without authentication, a very corner case if even possible.

If there are other authentication mechanisms configured in RabbitMQ, probably plain authentication headers are ignored, but not sure. I can try to test if this breaks this case.

metricbeat/mb/parse/url.go
```diff
@@ -37,13 +39,17 @@ func (b URLHostParserBuilder) Build() mb.HostParser {
 			if !ok {
 				return mb.HostData{}, errors.Errorf("'username' config for module %v is not a string", module.Name())
 			}
+		} else {
+			user = b.DefaultUsername
```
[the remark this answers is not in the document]
jsoriano: By now I'd prefer to leave it as is, but I agree that some improvements are possible here.

jsoriano:
Data generation is broken here, I'll solve it after #7106

jsoriano:
Ok, data generation works now as before.

PR #7116, opened 2018-05-16
Move filebeat module tests to parameterized tests

ruflin:
With this change, each test file inside a fileset is executed as its own test. This will make debugging issues much easier and prevent the problem on CI that there is no output for too long. Also the outputs, configs etc. for each test are in a separate directory.

### DIFF
--- a/filebeat/tests/system/test_modules.py
+++ b/filebeat/tests/system/test_modules.py
@@ -8,6 +8,45 @@ import subprocess
 from elasticsearch import Elasticsearch
 import json
 import logging
+from parameterized import parameterized
+
+
+def load_fileset_test_cases():
+    """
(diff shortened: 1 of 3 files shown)


ruflin:
jenkins, test this

filebeat/tests/system/test_modules.py
```diff
@@ -8,6 +8,45 @@
 from elasticsearch import Elasticsearch
 import json
 import logging
+from parameterized import parameterized
+
+
+def load_fileset_test_cases():
+    """
+    Creates a list of all modules, filesets and testfiles inside for testing.
+    To execute tests for only 1 module, set the env varialbe TESTING_FILEBEAT_MODULES
```
ph: s/varialble/variable

PR #7118, opened 2018-05-16
Add missing Pod UID registration

exekias:
This is a leftover from #7072, the indexer was not registered


libbeat/processors/add_kubernetes_metadata/kubernetes.go
```diff
@@ -31,6 +31,7 @@ func init() {
 	Indexing.AddIndexer(PodUIDIndexerName, NewPodUIDIndexer)
 	Indexing.AddIndexer(ContainerIndexerName, NewContainerIndexer)
 	Indexing.AddIndexer(IPPortIndexerName, NewIPPortIndexer)
+	Indexing.AddIndexer(PodUIDIndexerName, NewPodUIDIndexer)
```
mariomechoulam: Weird, I remember adding it, and it is indeed there on line 31, isn't it? :smiley:

exekias: uhm, that's correct :man_facepalming: , let me close this, I can confirm it works on master, probably I used a wrong build on my testing

PR #7120, opened 2018-05-16
Fix snapshot packaging for packetbeat and winlogbeat

ruflin:
arm builds were excluded for the builds which were run without SNAPSHOT param. But for the builds with SNAPSHOT=yes the PACKAGES_EXPERIMENTAL were not overwritten in packetbeat and winlogbeat.

* Change order of defining `TARGETS` and include Makefile as otherwise variable is overwritten


ruflin:
Wait with merging, there is still something wrong here.

PR #7123, opened 2018-05-16
Check if seccomp is supported

andrewkroh:
Rather than logging a warning when seccomp is not supported by the kernel this will log a simple info statement and bail out.

This also steps down the verbosity of the seccomp log message in the success case by not logging the filter. The filter will only be shown in debug mode.


andrewkroh:
These PRs https://github.com/elastic/go-seccomp-bpf/pulls should be merged first, then this PR can will be updated.

PR #7124, opened 2018-05-17
Add basic Elasticsearch server fileset

ruflin:
The Elasticsearch server fileset reads the standard Elasticsearch server log, handles multiline, extracts timestamp, log level and component. This is a very basic implementation to in a first step get logs into Elasticsearch. Additional work for more parsing is needed.

See https://github.com/elastic/beats/pull/5394/files for more details.


ruflin:
Should we call this `server` instead of `log` fileset?

inqueue:
>Should we call this `server` instead of `log` fileset?

I wondered about calling the base fileset `log` as well. Enough that it makes me wish the log was literally named `cluster.log`, since that is what is really is. What do you think about calling it `cluster`?

ruflin:
@inqueue Now that you mention cluster I wonder if this is node log file? Each ES node has this log file and in quite a few log lines the node id is even logged.

filebeat/module/elasticsearch/log/config/log.yml
```diff
@@ -0,0 +1,10 @@
+type: log
+paths:
+{{ range $i, $path := .paths }}
+ - {{$path}}
+{{ end }}
+exclude_files: [".gz$"]
```
joshuar: This needs to exclude the "other" log files that are likely in the same directory, see the exclude_files line used here:

https://github.com/joshuar/beats/blob/702f57baf6011dfd97482e7747743319d5915484/filebeat/module/elasticsearch/log/config/log.yml#L5

filebeat/module/elasticsearch/log/ingest/pipeline.json
```diff
@@ -0,0 +1,51 @@
...
+    ],
+    "processors": [
+        {
+            "grok": {
+                "field": "message",
+                "pattern_definitions": {
+                    "GREEDYMULTILINE": "(.|\n)*"
+                },
+                "patterns": [
+                    "\\[%{TIMESTAMP_ISO8601:timestamp}\\]\\[%{LOGLEVEL:log.level}%{SPACE}?\\]\\[%{DATA:elasticsearch.log.component}%{SPACE}*\\]%{GREEDYMULTILINE:message}"
```
joshuar: Does this single pattern parse all messages in a fully-configured ES with X-Pack? I believe that some components (notably ML) do not log with this structure (i.e., timestamp level component node as a prefix)?

filebeat/module/elasticsearch/log/config/log.yml
ruflin: Thanks, copied over. It would be great if in the future this log file would have a unique pattern like `server-*.log` prefix.

filebeat/module/elasticsearch/log/ingest/pipeline.json
ruflin: Was not aware of that unfortunately. I hoped that at least this part here is the same for all logs. I suggest we get it in as is and fix this in a follow up PR where we also add ml test logs.

ruflin:
Based on same discussions with the Elasticsearch team I will rename this metricset to server.

ruflin:
Fileset is now renamed to server, PR is rebased. I added a note about support for ML to https://github.com/elastic/beats/issues/7036. I'm planning to update https://github.com/elastic/beats/issues/5301 with more details on all the filesets and related tasks.

ruflin:
jenkins, test this

joshuar:
Do we need a test log with at least one exception trace in it so we can ensure the multiline handling is working (and catch regressions in it as new patterns are added)?

PR #7127, opened 2018-05-17
Metricbeat: Add maxmemory to redis info metricset

jsoriano:


metricbeat/module/redis/info/_meta/fields.yml
```diff
@@ -93,6 +93,17 @@
           description: >
             Used memory lua.
 
+        - name: max.value
```
ruflin: Based on the convention this would be `max.value.bytes` but I see it makes it a bit too long here.

PR #7131, opened 2018-05-18
Add support for keyword multifields

ruflin:
Before a multifield was only possible if the top level field was text. Now it can also be keyword.


ruflin:
@karmi FYI

karmi:
@ruflin, great!

PR #7134, opened 2018-05-18
Document username/password options for HTTP based modules

### LINKED ISSUE #6493: Document http authentication for metricbeat http module
https://github.com/elastic/beats/issues/4268 is closed, but authentication for the http module is not yet documented.

This can be specified via a headers map with a base64 string (example below is for elastic:changeme) for basic authorization, eg.

```
...
...
  request.enabled: true
  response.enabled: true
  headers:
    Authorization: Basic ZWxhc3RpYzpjaGFuZ2VtZQ==
```

tsg:
This adds commented out `username` and `password` settings to most
of the config places where they are supported (this part is manual).
Since those configs are included in the docs, they also show up there.

This also adds a line in the docs that link to "HTTP specific settings"
(this part is automatised during `make update`).

Completes #4268 and closes #6493.


metricbeat/docs/metricbeat-options.asciidoc
```diff
@@ -200,3 +200,19 @@ A list of processors to apply to the data generated by the metricset.
 See <<filtering-and-enhancing-data>> for information about specifying
 processors in your config.
 
+[float]
+[[module-http-config-options]]
+=== Standard HTTP config options
```
exekias: I wonder if we should just copy SSL settings here too, as both things are supported by the same helper

tsg: I'm curious what @dedemorton thinks of this. Perhaps a fragment that we include in both (or multiple) places?

dedemorton: @tsg Thank you for doing this! Right now, I agree that the user experience is not ideal (there's a lot of jumping around to find the full list of config options). Rather than pointing to another topic, I'd prefer to use an include to pull in any common (shared) options. To make this work, we'll need to add an attribute to the docs for each module (something like `:modulename: modulename`) and then use that attribute in the ID for shared sections. Do you want me to do that? This has been low down on my to-do list for awhile; I can bump the priority.

dedemorton: ...and looking at my notes, I see that the current design (pointing off to options) has resulted in problems for users in other areas (like not knowing that processors are valid in the config).

exekias:
I'm ok with merging this one, we can find a long-term solution later on

metricbeat/docs/metricbeat-options.asciidoc
dedemorton: I have this as a to-do item in a meta issue, but I've created a real issue to track the work: https://github.com/elastic/beats/issues/7160

PR #7135, opened 2018-05-18
Packetbeat process monitor enhancements

### LINKED ISSUE #541: Packetbeat - Associate a process with generic TCP/UDP flows
If Packetbeat is being used to monitor traffic on a single host (rather than consuming a data from a port mirror) it would be nice to associate each flow with a process on that host.

Possible ways to obtain that info:
Windows: [GetExtendedTcpTable](https://msdn.microsoft.com/en-us/library/windows/desktop/aa365928%28v=vs.85%29.aspx) / [GetExtendedUdpTable](https://msdn.microsoft.com/en-us/library/windows/desktop/aa365930%28v=vs.85%29.aspx)

adriansr:
This patch extends the functionality of the processes monitor in packetbeat:

- report the full command-line for all processes (not only those configured as `monitored`).
- Add support for Windows process monitoring.
- Disable the monitor when using file input.
- Get rid of polling for processes and sockets.

Closes #541

Updated functionality
----------------------

The original monitor will only track those processes that match a pattern specified in the configuration:

```
packetbeat.procs:
  enabled: true
  monitored:
    - process: mysqld
      cmdline_grep: mysqld

    - process: pgsql
      cmdline_grep: postgres

    - process: nginx
      cmdline_grep: nginx

    - process: app
      cmdline_grep: gunicorn
```

Will result in the following fields being added to generated events:

`"proc": "mysqld"` when the process is detected to be on the server-side of a stream, and
`"client_proc": "mysqld"` when the process is detected on the client-side.

While maintaining this behavior, this patch will add:

`"cmdline": "/usr/bin/mysqld --bind-address=0.0.0.0 --data-dir=/var/lib/mysql"`

and the same for `client_cmdline` when the process is on the client-side of a connection.

The `cmdline` and `client_cmdline` fields will be added to events generated by any running process. The `proc` and `client_proc` fields will be added only to those processes in the `monitored` section, so the original behavior is maintained.

Windows support
------------------

Added support for Windows using IP Helper functions (see issue #541).

Gosigar dependency
--------------------

The original feature was tailored to Linux. To make it multiplatform, gosigar is used to get process information. This also means that it is no longer required to run packetbeat as root to access the command-line from processes belonging to other users.

Get rid of processes/sockets polling
----------------

The original implementation would refresh the process list with a specified frequency (`refreshPidsFreq`). The mapping of TCP ports to processes would also be updated after a specified interval (`maxReadFreq`). This caused the feature to frequently miss process information for an event, specially with short-lived processes and connections.

In this implementation, such delays are removed. The list of processes and active sockets is updated as necessary. While this is a somewhat expensive operation, ranging from ~10ms under Linux and ~100ms in Windows, it is only performed when an event is being populated, and the source or destination of this traffic is a local port for which we don't have any information yet.

As the processes monitor is an opt-in feature, disabled by default, I find it more important to aim for more reliable behavior in the expense of some delays.

Missing features
----------------

Some limitations of the original implementation have not been addressed by this PR:
- Support UDP sockets.
- Handle processes bound to a subset of the local network addresses, and by extension two or more processes using the same port on different local addresses.
- Fetch Linux sockets using elastic/gosigar/sys/linux/inetdiag (faster than the current proc-based access).
- Move Windows socket querying to gosigar/sys/windows and use it in metricbeat's system/socket module.
- Support macOS.


packetbeat/procs/procs_windows.go
```diff
@@ -0,0 +1,127 @@
...
+type tcpRowOwnerPIDExtractor callbackFn
+type tcp6RowOwnerPIDExtractor callbackFn
+
+var tables = []struct {
+	family    uint32
+	function  GetExtendedTableFn
+	class     uint32
+	extractor extractorFactory
+}{
+	{windows.AF_INET, _GetExtendedTcpTable, TCP_TABLE_OWNER_PID_ALL, extractTCPRowOwnerPID},
```
andrewkroh: These TCP table structures look fun to work with [_read with sarcastic tone_].

adriansr:
Updated CHANGELOG

andrewkroh:
jenkins, test it

PR #7140, opened 2018-05-18
Fix permissions of generated Filebeat filesets

### LINKED ISSUE #6856: Fix file permissions of generated files by module and fileset generator
Right now thanks to permission problems snapshot builds fail every time a generated module is merged to upstream.

The current permission is 755. This has to be set to 644.

kvch:
Closes #6856

### DIFF
--- a/filebeat/scripts/generator/generator.go
+++ b/filebeat/scripts/generator/generator.go
@@ -49,7 +49,7 @@ func AppendTemplate(template, dest string, replace map[string]string) error {
 		return err
 	}
 
-	f, err := os.OpenFile(dest, os.O_WRONLY|os.O_APPEND, os.ModePerm)
+	f, err := os.OpenFile(dest, os.O_WRONLY|os.O_APPEND, 0644)
 	if err == nil {
 		_, err = f.Write(c)
 	}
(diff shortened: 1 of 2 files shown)


filebeat/scripts/generator/generator.go
```diff
@@ -70,7 +70,8 @@ func copyTemplate(template, dest string, replace map[string]string) error {
 	if err != nil {
 		return fmt.Errorf("cannot copy template: %v", err)
 	}
-	return nil
+
+	return os.Chmod(dest, 0644)
```
andrewkroh: `0644` should be passed to the `ioutil.WriteFile` rather than doing this in a separate syscall.

On a side note, it looks like this file is is incorrectly using `os.ModePerm`. Generally this value is used as a mask with mode bits to isolate the permission bits (e.g. `perm = fileinfo.Mode() & os.ModePerm`). Writing a file with `0777` should almost never be done for security reasons. In most cases if you are creating a file without any sensitive data use `0644` and if it's a dir use `0755`.

ruflin:
@kvch If it's also a problem in master, I would suggest to open it against master and then do the backport with out backport scripts.

kvch:
I meant if the developer changelog is the correct changelog to put this entry.
But you are right, this should be backported.

ruflin:
@kvch Argh, sorry. I was looking at an other PR against 6.2 and saw this comment and thought I'm still in the same PR. Ignore my comment.

+1 on where you added it.

PR #7141, opened 2018-05-18
Jolokia discovery autodiscover provider

jsoriano:
[Jolokia Discovery](https://github.com/rhuss/jolokia/issues/127) as an [autodiscover](https://www.elastic.co/guide/en/beats/metricbeat/current/configuration-autodiscover.html) provider.

Jolokia discovery works using UDP multicast, services can expose their jolokia endpoints by joining a multicast group and replying a probe when requested. In this implementation, metricbeat sends the query periodically (every 10 seconds by default) to the multicast group and keeps a list of known instances. Unknown instances are reported as started the first time they are seen, and known instances that stop responding during a time (~30 seconds by default) are considered stopped. The probe is sent using a list of configured interfaces.

I was not sure if doing it as an autodiscover provider was the proper way to do it, but at the end having templates for this is pretty handy. To support Jolokia Discovery as a plain module/metricset would have required some tricky host-less configurations.

I thought about not having generic templates and having just specific Jolokia/JMX configurations as by now it mostly makes sense to use it to configure the jmx metricset. But this way we can use the same templates here, and it can also be a starting point for more creative uses, as collecting logs apart of metrics (if paths can be derived from jolokia info), or to configure future metricsets based on Jolokia (imagine for example metricsets in the kafka or tomcat modules that collect metrics using Jolokia).

It works with a configuration like this one:
```
metricbeat.autodiscover:
  providers:
    - type: jolokia
      interfaces:
      - name: br*
        interval: 5s
        grace_period: 10s
      - name: en*
        interval: 60s
      templates:
      - condition:
          contains:
            jolokia.server.product: "tomcat"
        config:
        - module: jolokia
          metricsets: ["jmx"]
          hosts: "${data.jolokia.url}"
          namespace: test
          jmx.mappings:
          - mbean: "java.lang:type=Runtime"
            attributes:
            - attr: Uptime
              field: uptime
```
With this configuration it generates events containing these fields:
```
    "jolokia": {
      "test": {
        "uptime": 17396313
      },
      "url": "http://172.18.0.3:8778/jolokia",
      "agent": {
        "id": "172.18.0.3-6-47ef4541-servlet",
        "version": "1.5.0"
      },
      "secured": false,
      "server": {
        "product": "tomcat",
        "vendor": "Apache",
        "version": "7.0.86"
      }
    },
```


libbeat/autodiscover/providers/jolokia/discovery.go
```diff
@@ -0,0 +1,269 @@
...
+			if _, err := conn.WriteTo(queryMessage, &discoveryAddress); err != nil {
+				logp.Err(err.Error())
+				return
+			}
+
+			b := make([]byte, 1500)
+			for {
+				n, _, err := conn.ReadFrom(b)
+				if err != nil {
+					if !err.(net.Error).Timeout() {
```
exekias: probably you need a type check/switch here

metricbeat/module/jolokia/jmx/jmx_integration_test.go
```diff
@@ -29,6 +29,8 @@ func TestFetch(t *testing.T) {
 }
 
 func TestData(t *testing.T) {
+	compose.EnsureUp(t, "jolokia")
```
exekias: how was this passing tests before?

ruflin:
+1 on moving forward with this one.

libbeat/autodiscover/providers/jolokia/discovery.go
jsoriano: In principle `err` here is always going to be `nil` or of type `net.Error`, but yes, just in case it can be a good idea to properly check the type :)

jsoriano:
@exekias regarding using it only in `metricbeat`, yes, I think that in principle it has mostly sense to be used in metricbeat, and more specifically in the jolokia jmx metricset :)
But as I mention in the description of the PR, I'd leave it open to creative uses, for example an application could have its logs in `/var/log/apps/${data.jolokia.agent.id}.log` and an specific prospector could be created dynamically for it, or filebeat modules could be dynamically enabled depending on the `server.product`.

jsoriano:
Interval and grace period configuration at the interface level now.

jsoriano:
I have been playing with jolokia and the possibilities for autoconfiguration are quite amazing, basically anything that is configurable in the JVM is exposed, so for example from a Tomcat process we can obtain:
* `Catalina:type=Engine/baseDir`: `"/usr/apache-tomcat-7.0.86"`
* `Catalina:host=localhost,name=AccessLogValve,type=Valve/directory`: `"logs"`
* `Catalina:host=localhost,name=AccessLogValve,type=Valve/pattern`: `"%h %l %u %t \"%r\" %s %b"`

With that we'd know the format of the logs, and the directory where they are stored.

In any case by now I'm going to get Discovery support merged and we can think on more possibilities in the future.

filebeat/docs/autodiscover-jolokia-config.asciidoc
```diff
@@ -0,0 +1,23 @@
+Filebeat supports templates for inputs and modules:
+
+["source","yaml",subs="attributes"]
+-------------------------------------------------------------------------------
+metricbeat.autodiscover:
```
ruflin: This should be `filebeat.autodisover`

libbeat/autodiscover/providers/jolokia/config.go
```diff
@@ -0,0 +1,88 @@
...
+	// Time to wait till a response to a probe arrives
+	ProbeTimeout time.Duration `config:"probe_timeout"`
+
+	// Time since an instance is last seen and is considered removed
+	GracePeriod time.Duration `config:"grace_period"`
+}
+
+// WithDefaults returns an InterfaceConfig with default values
+func (c InterfaceConfig) WithDefaults() InterfaceConfig {
+	if c.Interval == 0 {
```
ruflin: Could we use ucfg here somehow?

libbeat/docs/shared-autodiscover.asciidoc
```diff
@@ -114,6 +114,57 @@ contain variables from the autodiscover event. They can be accessed under data n
...
+Jolokia Discovery is based on UDP multicast requests. Agents join the multicast
+group 239.192.48.84, port 24884, and discovery is done by sending queries to
+this group. You have to take into account that UDP traffic between Metricbeat
+and the Jolokia agents has to be allowed. Also notice that this multicast
+address is in the 239.0.0.0/8 range, that is reserved for private use within an
+organization, so it can only be used in private networks.
+
+These are the available fields on every event:
+
+  * jolokia.agent.id
```
ruflin: Are these fields also added to each event indirectly created through auto discovery?

filebeat/docs/autodiscover-jolokia-config.asciidoc
jsoriano: Good catch, fixed

libbeat/docs/shared-autodiscover.asciidoc
jsoriano: Yes, they are added, I think that by the Autodiscover framework.

exekias: Everything under `meta` here is added: https://github.com/elastic/beats/pull/7141/files#diff-9132fdc5edac3dbc7a2ef1540ad3e471R68

libbeat/autodiscover/providers/jolokia/config.go
jsoriano: It took a while to me to understand how I could do it, but done :)

jsoriano:

jsoriano:
I think this is ready for a final review as in any case it is going to be experimental by now.

metricbeat/tests/system/test_autodiscover_jolokia.py
```diff
@@ -0,0 +1,54 @@
...
+import unittest
+from nose.plugins.attrib import attr
+
+
+class Test(metricbeat.BaseTest):
+
+    COMPOSE_SERVICES = ['jolokia']
+
+    @unittest.skipUnless(metricbeat.INTEGRATION_TESTS, "integration test")
+    def test_jolokia_jmx(self):
```
jsoriano: This is the only test really covering this feature by now, and is under metricbeat. This is not ideal, but it should be enough to detect big issues in its main use case.
Big part of the code is dependant of the discovery protocol and for that we'd need scenarios like the ones we have in metricbeat. I'd leave for the future further tests, like mocking all the networking-dependant part, or preparing a more general testing scenario under libbeat.

libbeat/autodiscover/providers/jolokia/discovery.go
```diff
@@ -0,0 +1,296 @@
...
+// }
+//
+// Example discovery probe with socat
+//
+//   echo '{"type":"query"}' | sudo socat STDIO UDP4-DATAGRAM:239.192.48.84:24884,interface=br0 | jq .
+//
+
+// Message contains the information of a Jolokia Discovery message
+var messageSchema = s.Schema{
+	"agent": s.Object{
```
ruflin: So this is not the `agent` that ships the data but the jolokia agent that discovered the data. Just trying to get my head around this for ECS: https://github.com/elastic/ecs#agent The `agent` on the top level in these cases would still be metricbeat or filebeat, right?

ruflin:
Should we add a quick example to the autodiscover part in the reference file? At the same time I think it's time to link to the auto discovery docs from our reference and config files to get all details there.

libbeat/autodiscover/providers/jolokia/discovery.go
```diff
@@ -0,0 +1,296 @@
...
+func (d *Discovery) run() {
+	var cases []reflect.SelectCase
+	for _, i := range d.Interfaces {
+		// Send a probe on each interface when starting
+		d.sendProbe(i)
+
+		// And then one after each interval
+		ticker := time.NewTicker(i.Interval)
+		defer ticker.Stop()
+		cases = append(cases, reflect.SelectCase{
```
exekias: I'm not particularly a fan of using `reflect`, it makes the code difficult to read, perhaps this could be solved with a goroutine per interface?

libbeat/autodiscover/providers/jolokia/discovery.go
```diff
@@ -0,0 +1,296 @@
...
+
+func (d *Discovery) update(config InterfaceConfig, message common.MapStr) {
+	v, err := message.GetValue("agent.id")
+	if err != nil {
+		logp.Err("failed to update agent without id: " + err.Error())
+		return
+	}
+	agentID, ok := v.(string)
+	if len(agentID) == 0 || !ok {
+		logp.Err("empty agent?")
```
exekias: This should be more verbose

libbeat/autodiscover/providers/jolokia/discovery.go
jsoriano: Yes, this is about the jolokia agent, under the `jolokia` namespace.

jsoriano:
@ruflin, do you mean adding an example to the reference config of JMX?

libbeat/autodiscover/providers/jolokia/discovery.go
jsoriano: Better without `reflect`, yes :) I traslated directly the select of the previous implementation to a dynamic select.

ruflin:
@jsoriano Yes, similar to what we have here for the docker autodiscovery: https://github.com/elastic/beats/blob/master/metricbeat/_meta/common.reference.yml#L34 But TBH I think it's better we start to add links to our config files to the docs.

In general I wonder if autodiscovery should be documented in one place and show example for each Beat or if it should be document in each beat and there only have the examples for this Beat. (cc @dedemorton )

dedemorton:
@ruflin First I'd consider the audience. Since the person who will set up autodiscovery is also the person who will set up the config file, I'd leave the content where it is for now (single-sourced in libbeat and reused in all the beats docs). 

I already have an item on my to-do list to improve the docs about running on containers. I think we bury too much info in the "configuring" sections. I'd like to pull all the topics about running beats on containers and put them together in one section since the target audience is clear. It's on my to-do list, but maybe I need to pop the priority?

ruflin:
@dedemorton I think the problem here is that the examples for Filebeat and Metricbeat are potentially different so it can't be just reused. This here is not necessarly related to docker / K8s. For the priorities : you know best :-)

dedemorton:
@ruflin Not sure I'm following you completely. The way the documentation is set up, the include statement already pulls in the example that's relevant to the beat: `include::../../{beatname_lc}/docs/autodiscover-jolokia-config.asciidoc[]`

Maybe you're thinking about this from the perspective of how you will link to the doc examples from within the reference yml?

(I get your point about Jolokia and docker.... I guess I should understand more about Jolokia before I make broad statements.  :-P)

ruflin:
@dedemorton I think I missed the part that we already include beat specific configs.

@jsoriano You will need an other rebase.

jsoriano:
Branch updated.

@ruflin @dedemorton do you think I should change something in the docs before merging? or we can merge and revisit this later?

ruflin:
@jsoriano No change needed now.

exekias:
jenkins, retest this please (WFG)

beirtipol:
@jsoriano - thanks for turning this around. Any eta for this going live?

beirtipol:
@jsoriano thanks

PR #7144, opened 2018-05-20
Elasticsearch pending tasks metricset

jclegras:

With this metricset I can keep an eye on these pending tasks.

Render them as such : 

```
"elasticsearch": {
  "cluster": {
    "pending_task": {
      "insert_order": 47,
      "priority": "HIGH",
      "source": "put-mapping",
      "time_in_queue.ms": 34
    }
  }
}
```


ruflin:
jenkins, test it

metricbeat/module/elasticsearch/pending_tasks/data.go
```diff
@@ -0,0 +1,39 @@
...
+	s "github.com/elastic/beats/libbeat/common/schema"
+	c "github.com/elastic/beats/libbeat/common/schema/mapstriface"
+)
+
+var (
+	schema = s.Schema{
+		"insert_order":     c.Int("insert_order"),
+		"priority":         c.Str("priority"),
+		"source":           c.Str("source"),
+		"time_in_queue.ms": c.Int("time_in_queue_millis"),
```
ruflin: This should be as following to make sure internally it's already an object. But outcome on the ES side is identical. We can do this in a follow up PR.

```
"time_in_queue": s.Object{
  "ms": c.Int("time_in_queue_millis"),
}
```

metricbeat/module/elasticsearch/pending_tasks/pending_tasks.go
```diff
@@ -0,0 +1,77 @@
...
+		PathConfigKey: "path",
+		DefaultPath:   "_cluster/pending_tasks",
+	}.Build()
+)
+
+// MetricSet holds any configuration or state information. It must implement
+// the mb.MetricSet interface. And this is best achieved by embedding
+// mb.BaseMetricSet because it implements all of the required mb.MetricSet
+// interface methods except for Fetch.
+type MetricSet struct {
```
ruflin: There has been some somplifications to the Elasticsearch metricsets in the last days, see for example https://github.com/elastic/beats/blob/master/metricbeat/module/elasticsearch/index/index.go#L24 We can also address this in a follow up PR. This will work as expected.

ruflin:
jenkins, test this

ruflin:
@jclegras Thanks for splitting up the PR's. Merged.

PR #7145, opened 2018-05-20
Elasticsearch pending tasks summary metricset

jclegras:

With this metricset I can keep an eye on a summary of these pending tasks.

Render as such :

```
  "elasticsearch": {
    "cluster": {
      "pending_tasks_summary": {
        "count_total": 2,
        "count_priority_urgent": 1,
        "count_priority_high": 1,
        "max_time_in_queue_millis": 86
      }
    }
  }
```


ruflin:
You will have to rebase this one on master but don't worry about resolving the conflicts except for the one in `elasticsearch.asciidoc`. The other 3 are auto generated files and will automatically update when you run `make update` after the rebase.

metricbeat/module/elasticsearch/pending_tasks_summary/_meta/data.json
```diff
@@ -0,0 +1,24 @@
...
+  "metricset": {
+    "host": "localhost:9200",
+    "rtt": 5818,
+    "namespace": "elasticsearch.cluster.pending_tasks_summary",
+    "name": "pending_tasks_summary",
+    "module": "elasticsearch"
+  },
+  "elasticsearch": {
+    "cluster": {
+      "pending_tasks_summary": {
```
ruflin: I'm still in the progress on figuring out on what our "logic" is on how we add summary metricsets. If we would follow the logic from `system.process.summary` this would be `elasticsearch.cluster.pending_task.summary`. But not sure if this fits well here.

I think there are 2 parts here: 

* Configuration + name of metricset
* Data format it is ingested in

The configuration part should make it really easy to understand what data is added. The data that then is added should make it as simple as possible to query the data without necessarily having knowledge about metricsets etc. Basically making it possible for someone the explore the data without knowledge about Metricbeat.

An alternative proposal here would be (replace . with json objects)
```
elasticsearch.cluster.pendings_tasks.total
elasticsearch.cluster.pendings_tasks.priority.urgent
elasticsearch.cluster.pendings_tasks.priority.high
elasticsearch.cluster.pendings_tasks.time_in_queue.max.ms
```

The last one is a bit unfortunate with `max.ms`. 

The downside of the above is that there will be `pending_task` and `pending_tasks` which could be confusing. At the same time I think the full names make it obvious what it is.

Curious to hear other thoughts.

jclegras:
Have you decided @ruflin ?

ruflin:
Thanks for reviving this PR @jclegras . My current favorite is the one with `pending_tasks`. What is your preference? We didn't come up with a generic way so let's pick one for now to move forward.

@jsoriano @ycombinator Any thoughts?

metricbeat/module/elasticsearch/pending_tasks_summary/_meta/data.json
jsoriano: Currently we have two `*_summary` metricsets (one of them the `index_summary` also in the `elasticsearch` module), and they both add a `summary` object level over the summarized thing. I think it'd be a good option to follow this schema also here.
```
elasticsearch.pending_task.summary.total
elasticsearch.pending_task.summary.priority.urgent
elasticsearch.pending_task.summary.priority.high
elasticsearch.pending_task.summary.time_in_queue.ms.max
```

For the time in queue I put `...ms.max` instead of `max.ms` so we show that this is an aggregation of the `pending_task.time_in_queue.ms` metric. Also if we have additional aggregations like the average we could add it as `...ms.avg`.

ruflin: +1. Thanks for the input @jsoriano 

@jclegras Does that work for you?

jsoriano: Oh, I removed the `cluster` level in my proposal, I guess that this is another discussion, do we need it?

ruflin: It's not required. I felt like it makes it easy to see that these are pending_tasks on the cluster level but users probably know that, so we can skip it.

jclegras: I agree with you as long as it's consistent with the other modules.
I'll rewrite this PR on the following days.
Regards.

ruflin: thanks

PR #7163, opened 2018-05-22
Stop worker goroutine on auditd metricset shutdown

andrewkroh:
A worker goroutine for reading events from socket was being leaked when the auditd metricset was stopped. The goroutine will now stop when the AuditClient is closed because it will exited on EBADF (bad file descriptor).

I cleaned up a few tests in the process and added a filter for AUDIT_REPLACE messages.


exekias:
jenkins, retest this please

andrewkroh:
jenkins, test it

andrewkroh:
jenkins, retest this please

Issue #7119, opened 2018-05-16
Proxy settings in heartbeat


kvalev:
According to the [documentation](https://www.elastic.co/guide/en/beats/heartbeat/master/configuration-heartbeat-options.html#monitor-http-proxy-url), one can configure a proxy to be used for an HTTP monitory by either providing the `proxy_url` parameter or by setting the `HTTP_PROXY` environment variable. However it seems that the HTTP_PROXY environment is actually ignored (and even if it isnt, it does not take NO_PROXY into account).
From what I can tell, the [elasticsearch output](https://github.com/elastic/beats/blob/32c60c136eda2256f25cac361c7ab37444f4a260/libbeat/docs/outputconfig.asciidoc#proxy_url) that takes the environment variables into account.

ruflin:
Can you share your full config, heartbeat version and heartbeat logs?

kvalev:
I am not using a publicly accessible systems, so there is no point in sharing the config. Anyway the issue was quite easy to solve (see patch below).
However it seems to be another, a bit more complicated problem - when using an heartbeat monitor for an https endpoint that goes through an http proxy, the ssl verification mode is being ignored.

```
From e7053d8b8ebabe06ddfcc66d41b7699c0e501561 Mon Sep 17 00:00:00 2001
From: Krassimir Valev <krassimir_valev@yahoo.com>
Date: Thu, 17 May 2018 09:52:05 +0200
Subject: [PATCH] Advanced proxy settings for the http heartbeat monitor

---
 heartbeat/monitors/active/http/http.go | 14 +++++++-------
 heartbeat/monitors/active/http/task.go |  2 +-
 2 files changed, 8 insertions(+), 8 deletions(-)

diff --git a/heartbeat/monitors/active/http/http.go b/heartbeat/monitors/active/http/http.go
index 986f1b0..eceed85 100644
--- a/heartbeat/monitors/active/http/http.go
+++ b/heartbeat/monitors/active/http/http.go
@@ -57,12 +57,12 @@ func create(
 
 	jobs := make([]monitors.Job, len(config.URLs))
 
-	if config.ProxyURL != "" {
-		transport, err := newRoundTripper(&config, tls)
-		if err != nil {
-			return nil, err
-		}
+	transport, err := newRoundTripper(&config, tls)
+	if err != nil {
+		return nil, err
+	}
 
+	if transport.Proxy != nil {
 		for i, url := range config.URLs {
[...]

PR #7166, opened 2018-05-23
Advanced proxy handling in heartbeat

kvalev:
This PR tries to showcase the following two proxy-related issues when using http monitors in heartbeat:
- the proxy-related environment variables are not taken into account, even though [according to the documentation](https://www.elastic.co/guide/en/beats/heartbeat/master/configuration-heartbeat-options.html#monitor-http-proxy-url) HTTP_PROXY should be 
- when connecting to an https endpoint without a proper ssl certificate via a proxy, the `ssl.verification_mode` is being ignored (both with an http and http proxy) - see test methods TestHTTPProxyForHTTPSEndpoint & TestHTTPSProxyForHTTPSEndpoint (both of them fail)

Relates to elastic/beats#7119

### DIFF
--- a/heartbeat/monitors/active/http/http.go
+++ b/heartbeat/monitors/active/http/http.go
@@ -57,12 +57,12 @@ func create(
 
 	jobs := make([]monitors.Job, len(config.URLs))
 
-	if config.ProxyURL != "" {
-		transport, err := newRoundTripper(&config, tls)
-		if err != nil {
-			return nil, err
-		}
+	transport, err := newRoundTripper(&config, tls)
+	if err != nil {
+		return nil, err
+	}
 
+	if transport.Proxy != nil {
 		for i, url := range config.URLs {
 			jobs[i], err = newHTTPMonitorHostJob(url, &config, transport, enc, body, validator)
 			if err != nil {
@@ -71,7 +71,7 @@ func create(
 		}
 	} else {
 		for i, url := range config.URLs {
-			jobs[i], err = newHTTPMonitorIPsJob(&config, url, tls, enc, body, validator)
+			jobs[i], err = newHTTPMonitorIPsJob(url, &config, tls, enc, body, validator)
 			if err != nil {
 				return nil, err
 			}
@@ -82,7 +82,7 @@ func create(
 }
 
 func newRoundTripper(config *Config, tls *transport.TLSConfig) (*http.Transport, error) {
-	var proxy func(*http.Request) (*url.URL, error)
+	proxy := http.ProxyFromEnvironment
 	if config.ProxyURL != "" {
 		url, err := url.Parse(config.ProxyURL)
 		if err != nil {
--- a/heartbeat/monitors/active/http/http_test.go
+++ b/heartbeat/monitors/active/http/http_test.go
@@ -1,0 +1,210 @@
+package http
+
+import (
+	"github.com/stretchr/testify/assert"
+	"time"
+	"net"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"testing"
+
+	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/heartbeat/monitors"
+)
+
+type ServerHandler struct {
+	Invoked bool
+}
+
+func (h *ServerHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	h.Invoked = true
+}
+
+type ProxyHandler struct {
+	Invoked bool
+}
+
+func (h *ProxyHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	h.Invoked = true
+
+	// proxy implementation from https://medium.com/@mlowicki/6a51c2f2c38c
+	if r.Method == http.MethodConnect {
+		handleTunneling(w, r)
+	} else {
+		handleHTTP(w, r)
+	}
+}
+
+func handleTunneling(w http.ResponseWriter, r *http.Request) {
+    destConnection, err := net.DialTimeout("tcp", r.Host, 10*time.Second)
+    if err != nil {
+        http.Error(w, err.Error(), http.StatusServiceUnavailable)
+        return
+    }
+    w.WriteHeader(http.StatusOK)
+    hijacker, ok := w.(http.Hijacker)
+    if !ok {
+        http.Error(w, "Hijacking not supported", http.StatusInternalServerError)
+        return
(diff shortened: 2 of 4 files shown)


heartbeat/monitors/active/http/http.go
```diff
@@ -57,12 +57,12 @@ func create(
...
-		transport, err := newRoundTripper(&config, tls)
-		if err != nil {
-			return nil, err
-		}
+	transport, err := newRoundTripper(&config, tls)
+	if err != nil {
+		return nil, err
+	}
 
+	if transport.Proxy != nil {
```
kvalev: This significantly changes the monitor behavior, as transport.Proxy will actually never be nil... meaning `newHTTPMonitorHostJob` will be used to create the monitor job and not `newHTTPMonitorIPsJob`, although I dont quite understand why there are two methods (the implementation looks quite different).

heartbeat/monitors/active/http/http_test.go
```diff
@@ -0,0 +1,210 @@
...
+			"verification_mode": "none",
+		},
+	}
+
+	runAndAssert(t, rawCfg)
+	assert.True(t, proxyHandler.Invoked)
+}
+
+func TestHTTPProxyConfiguredViaEnv(t *testing.T) {
+	t.Skip("This test cannot be executed until https://github.com/golang/go/issues/22079 is merged, as the proxy settings are being cached and there is no way to refresh them.")
```
kvalev: Unfortunately it not possible to test/showcase the first issue (taking the environment variables into account), as the proxy settings are being cached once by the http module and there is no way (yet) to reload them.

heartbeat/monitors/active/http/http.go
```diff
@@ -71,7 +71,7 @@ func create(
 		}
 	} else {
 		for i, url := range config.URLs {
-			jobs[i], err = newHTTPMonitorIPsJob(&config, url, tls, enc, body, validator)
+			jobs[i], err = newHTTPMonitorIPsJob(url, &config, tls, enc, body, validator)
```
ruflin: Is there a reason you changed the params order?

heartbeat/monitors/active/http/http.go
```diff
@@ -82,7 +82,7 @@ func create(
 }
 
 func newRoundTripper(config *Config, tls *transport.TLSConfig) (*http.Transport, error) {
-	var proxy func(*http.Request) (*url.URL, error)
+	proxy := http.ProxyFromEnvironment
```
ruflin: To keep the current behaviour, I'm wondering if we should first check if `HTTP_PROXY` is set. Main problem is it would not support all env variables supported by `ProxyFromEnvironment`.

heartbeat/monitors/active/http/http.go
kvalev: I changed it to be consistent with the param order of `newHTTPMonitorHostJob`. Can revert it if needed.

heartbeat/monitors/active/http/http.go
kvalev: This would mean also checking `HTTPS_PROXY` and both lowercase spellings to be compatible with the `ProxyFromEnvironment`. There is also the option of creating a dummy `Request` and checking whether a proxy should be used according to `ProxyFromEnvironment`.

urso: The HTTP module does not load proxy settings from the environment on purpose. For heartbeat there are potentially two networks. One being the network/service to be monitored and the other being the one we send the results too. The elasticsearch output uses ProxyFromEnvironment in order to support proxy setting when publishing events.

In heartbeat monitors one should rely on provided settings as much as possible (one can reference environment variables in config files btw.). I'm even considering to put a DNS client with configurable resolver and caching strategy into heartbeat.

andrewvc: To add to this @kvalev, this is a great feature to add, but I think we need to handle DNS resolution in a more controlled manner. Are you interested in updating this PR to do that?

kvalev: Thanks for the feedback.

I can somewhat see how the first point that I raised (proxy support in heartbeat) can be considered a feature. And you are right that one can explicitly set `proxy_url` for a specific group of hosts, instead of having to rely on the underlying configuration. I was just thinking it might be a nice idea to have it natively in heartbeat. It is fair if you see it differently, but then the docs should be updated, as they are straight up incorrect (I am referring to the `proxy_url` documentation, which mentions that the `HTTP_PROXY` environment variable is taken into account).

However the second point that I raised (that I have provided unit tests tests for) seems to be a bug from what I can see. I was looking into fixing the issue myself when I initially stumbled upon it, but it seems to be a bit more complex for my newbie Go skills and inexperience with the provided HTTP API. I was hoping that it might be much easier to be fixed by one of the core developers, given the provided test cases.

urso: > I can somewhat see how the first point that I raised (proxy support in heartbeat) can be considered a feature. And you are right that one can explicitly set proxy_url for a specific group of hosts, instead of having to rely on the underlying configuration. I was just thinking it might be a nice idea to have it natively in heartbeat. It is fair if you see it differently, but then the docs should be updated, as they are straight up incorrect (I am referring to the proxy_url documentation, which mentions that the HTTP_PROXY environment variable is taken into account).

You are right about the doc being incorrect. Clearly a doc bug. Mostly likely some copy-n-paste error :(
I will submit a PR removing the mention of HTTP_PROXY in the docs.
Thank you for pointing this out!

> However the second point that I raised (that I have provided unit tests tests for) seems to be a bug from what I can see. I was looking into fixing the issue myself when I initially stumbled upon it, but it seems to be a bit more complex for my newbie Go skills and inexperience with the provided HTTP API. I was hoping that it might be much easier to be fixed by one of the core developers, given the provided test cases.

I didn't dig into the unit test, but yeah, sounds like a bug. Thanks for taking the time to write unit tests! This PR seems to solve two things at the same time. How about splitting it up, such that we can review and possibly merge the bug fix in a separate PR.

urso: See #7831 for documentation update.

andrewvc:
@kvalev sorry for the delay here, I'd like to review this PR soon. Mind updating the PR to resolve the latest conflicts? Thanks for the great work :)

andrewvc:
Closing this as the docs have been updated and no longer mention the ENV var. Given the issues mentioned in-line I think this is now safe to close out

PR #7167, opened 2018-05-23
libbeat: schema.Apply() returns plain error or nil

jsoriano:
schema.Apply() returned an schema.Errors object as error, with this
object is not easy to check if there was really an error, specially when
optional fields are involved. This leads to lots of cases where these
errors are being ignored.

This change makes Apply to return an error only if a non-optional field
is missing or in any case if some field is misformatted. Otherwise it
returns plain nil, so a more idiomatic go error checking can be used.


jsoriano:
WIP, I'm not completely happy with the result and probably I'm missing some cases.

ruflin:
For testing purpose it is useful to also check for non required fields if they are missing. Is this still possible?

jsoriano:
Not with the returned error. For testing we might think in other approximation for explicitly checking that optional fields are also present, but I wouldn't make it return an error by default if an optional field is missing, this break a bit the nature of a non-required thing.

ruflin:
I agree with the statement above. The part that I worry is that our current testing in some cases relies on checking for the error, converting it and then check if there are also no optional fields missing.

We can definitively decide to move forward with this suggestion and fix this as a follow up.

jsoriano:
I'm following with this with other approach (in #7335), returning a "result" object instead of just an error, this result object would make easier to check the errors, and also the list of missing fields, what could be used in tests.

jsoriano:
Closing this one, continuing in #7335

PR #7170, opened 2018-05-24
libbeat: Stop autodiscover runners in a goroutine

jsoriano:
Stopping runners can be a blocking operation, what can block the
autodiscover event loop. Run it in a goroutine so events can continue
being handled.

### DIFF
--- a/libbeat/autodiscover/autodiscover.go
+++ b/libbeat/autodiscover/autodiscover.go
@@ -177,8 +177,12 @@ func (a *Autodiscover) handleStop(event bus.Event) {
 		}
 
 		if runner := a.runners.Get(hash); runner != nil {
+			// Stop can block, we run it asyncrhonously to avoid blocking
+			// the whole events loop. The runner hash is removed in any case
+			// so an equivalent configuration can be added again while this
+			// one stops, and a duplicated event don't try to stop it twice.
 			logp.Info("Autodiscover stopping runner: %s", runner)
-			runner.Stop()
+			go runner.Stop()
 			a.runners.Remove(hash)
 		} else {
 			logp.Debug(debugK, "Runner not found for stopping: %s", hash)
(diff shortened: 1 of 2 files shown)


exekias:
@urso @ruflin I find this interesting, do you have any insight about why module stopping may block for a while?

exekias:
@jsoriano we should backport this to, at least, 6.3

jsoriano:
In case it gives any clue, `runner.Start()` doesn't seem to block, and when it is blocked on `Stop()` it also blocks graceful stop with Ctrl-C.

libbeat/autodiscover/autodiscover.go
```diff
@@ -178,7 +178,7 @@ func (a *Autodiscover) handleStop(event bus.Event) {
 
 		if runner := a.runners.Get(hash); runner != nil {
 			logp.Info("Autodiscover stopping runner: %s", runner)
-			runner.Stop()
+			go runner.Stop()
 			a.runners.Remove(hash)
```
ruflin: the hash should only be removed if stopping is finished I would think

ruflin:
Could this cause some race conditions?

libbeat/autodiscover/autodiscover.go
exekias: We actually want to remove it here, if a new event arrives to start the same module we want that to happen, even if the previous instance is still stopping

jsoriano:
@ruflin it is possible, yes, not sure what happens if a runner is stopped twice.

exekias:
I don't think it's possible, events are processed by autodiscover one bye one, it acts as a state machine. This should ensure that the event is removed from the list by the moment you process the next event. For instance a 2 stop events (this can happen) would go like this:

- First event calls runner.Stop() as a goroutine, stop beggins. The runner is removed from the list.
- Second event goes to the list and see no runner for the given config, so it does nothing; even if the runner is still stopping.

jsoriano:
Added changelog.

libbeat/autodiscover/autodiscover.go
ruflin: I see, so adding and removing is synchronous. To make it more clear in the code we should add a comment here and move the stopping command after the removal of the hash.

Any concerns about leaking go routines here if `Stop()` gets stuck?

ruflin:
Change LGTM but I think we should document what we discussed here in the code for the next person touching this code.

jsoriano:
@ruflin ok, I have added a comment explaining the decissions there.

ruflin:
jenkins, test this

Issue #7028, opened 2018-05-07
Central Management


exekias:
Beats central management will allow users to configure & manage Beats centrally from Kibana. This is a meta issue to group all efforts for the first version of this feature:

Fixes: 

- [x] Bug when unenrolling a beats https://github.com/elastic/beats/issues/9452

Tasks:

- [x] Elasticsearch license checking and built-in roles: https://github.com/elastic/elasticsearch/issues/30493
   - [x] Need to backport these two PRs to `6.x` whenever the rest of Beats management is ready:
      * [x] https://github.com/elastic/elasticsearch/pull/30520
      * [x] https://github.com/elastic/stack-docs/pull/80
- [x] Beats initial enrollment https://github.com/elastic/beats/issues/7032
- [ ] Kibana management UI - kibana repo https://github.com/elastic/kibana/issues/18858
	- [x] https://github.com/elastic/kibana/issues/20837 Create beat enrollments screen(s)
	- [x] https://github.com/elastic/kibana/issues/20849 Create ui adapters
	- [x] https://github.com/elastic/kibana/issues/20844 table search and action controls
	- [x] https://github.com/elastic/kibana/issues/20834 Beats List Screen
	- [x] https://github.com/elastic/kibana/issues/20845 Shared Beat table
	- [x] https://github.com/elastic/kibana/issues/20836 Tag list screen
	- [x] https://github.com/elastic/kibana/issues/20843 Shared Tags table view
	- [x] https://github.com/elastic/kibana/issues/20838 Tag creation screen
	- [ ] https://github.com/elastic/kibana/issues/20835 Beat activity view
	- [x] https://github.com/elastic/kibana/issues/20840 Beat details - Config
	- [ ] https://github.com/elastic/kibana/issues/20846 config creation / view slide in view
	- [ ] https://github.com/elastic/kibana/issues/20842 Beat details - logs
	#### Tasks currently awaiting efforts from other teams
	- [ ] Activity view (awaiting logging UI in EUI)
	#### Overlay work (tasks that do not impede a working demo/beta)
	- [x] License check
	- [x] user permission check in UI and associated error messaging
- [x] Kibana license checking
- [x] Beats central management manager #8263
- [x] Basic telemetry for Beats central management https://github.com/elastic/beats/issues/7979 
- [x] Add docs https://github.com/elastic/beats/issues/8470

exekias:
Pull request to merge Central Management branch to Beats master is now open: https://github.com/elastic/beats/pull/8559

[...]

PR #7172, opened 2018-05-24
Unify Runner dynamic launching and reloading under RunnerList

exekias:
This PR unifies all uses of dynamic runner handling under a new `RunnerList` structure. It can receive a list of configs as the desired state and do all the needed changes (start/stopping) to transition to it.

I plan to use this as part of #7028


libbeat/cfgfile/list.go
```diff
@@ -0,0 +1,133 @@
...
+		} else {
+			startList[hash] = config
+		}
+	}
+
+	// Stop removed runners
+	for hash, runner := range stopList {
+		debugf("Stopping runner: %s", runner)
+		delete(r.runners, hash)
+		go runner.Stop()
```
ruflin: This is a change in behaviour from previous implementation where it was synchronous. This could become a problem in Filebeat or at least lead to more errors. Before a new prospector can be started on an existing state of a file, the state must be set to Finish. If the shutdown happens in go routine, this overlap can become bigger.  But I think it should still work.

libbeat/cfgfile/list.go
```diff
@@ -0,0 +1,133 @@
...
+		delete(r.runners, h)
+		runner.Stop()
+	}
+}
+
+// Has returns true if a runner with the given hash is running
+func (r *RunnerList) Has(hash uint64) bool {
+	r.mutex.Lock()
+	defer r.mutex.Unlock()
+	_, ok := r.runners[hash]
```
ruflin: A read write mutex could be used here.

libbeat/cfgfile/registry_test.go
```diff
@@ -1,48 +0,0 @@
-package cfgfile
```
ruflin: Any replacement for these tests?

libbeat/cfgfile/reload.go
```diff
@@ -167,49 +167,7 @@ func (rl *Reloader) Run(runnerFactory RunnerFactory) {
...
-				err := c.Unpack(rawCfg)
-
-				if err != nil {
-					logp.Err("Unable to unpack config file due to error: %v", err)
-					continue
-				}
-
-				hash, err := hashstructure.Hash(rawCfg, nil)
-				if err != nil {
-					// Make sure the next run also updates because some runners were not properly loaded
```
ruflin: Is this case still handled?

libbeat/cfgfile/list.go
```diff
@@ -0,0 +1,133 @@
...
+
+// Stop all runners
+func (r *RunnerList) Stop() {
+	r.mutex.Lock()
+	defer r.mutex.Unlock()
+
+	for h, runner := range r.runners {
+		debugf("Stopping runner: %s", runner)
+		delete(r.runners, h)
+		runner.Stop()
```
ruflin: Before runners were stopped in parallel. This is important to speed up the shutdown process especially with a large amount of modules or prospectors.

libbeat/cfgfile/registry_test.go
exekias: I forgot to add the file to the commit :man_facepalming: pushed

libbeat/cfgfile/reload.go
exekias: Nope :/ I'll have a look and find a solution

libbeat/cfgfile/list.go
exekias: makes sense, will update the code to do that

libbeat/cfgfile/reload.go
exekias: See 4ea21f2

exekias:
Thanks @ruflin for the thorough review, I think this is ready for a 2nd look :)

exekias:
I've added 0d44b7e after our last discussion, that should cover cases were a Filebeat input fails to start due to unclosed states (it will keep retrying every 10s)

ruflin:
The RACE detector does not seem to be happy on Jenkins

exekias:
fixed

libbeat/autodiscover/autodiscover.go
```diff
@@ -87,38 +94,59 @@ func (a *Autodiscover) Start() {
...
+			}
+
+			configs := make([]*cfgfile.ConfigWithMeta, 0, len(a.configs))
+			for _, c := range a.configs {
+				configs = append(configs, c)
+			}
+
+			err := a.runners.Reload(configs)
+
+			// On error, make sure the next run also updates because some runners were not properly loaded
```
ruflin: Should we create a debug log entry for this err?

libbeat/autodiscover/autodiscover.go
```diff
@@ -176,18 +198,15 @@ func (a *Autodiscover) handleStop(event bus.Event) {
...
-		if runner := a.runners.Get(hash); runner != nil {
-			// Stop can block, we run it asyncrhonously to avoid blocking
-			// the whole events loop. The runner hash is removed in any case
-			// so an equivalent configuration can be added again while this
-			// one stops, and a duplicated event don't try to stop it twice.
-			logp.Info("Autodiscover stopping runner: %s", runner)
-			go runner.Stop()
-			a.runners.Remove(hash)
+		if a.runners.Has(hash) {
+			delete(a.configs, hash)
```
ruflin: Can only 1 go routine at the time access `a.configs`?

Where is the runner stopped now?

exekias: configs is only modified by the event loop (`handleStart` and `handleStop`), so there is only one goroutine using it. Start/Stop is now handled by `runners.Reload`: https://github.com/elastic/beats/pull/7172/files/63c350a7ea841fd6f2b5bea7bea4debe1c0afdf5#diff-3b0ae03a54b2e31c58f8dcf308a6a7b8R127. This is the same code path used by configuration reload mechanism

PR #7176, opened 2018-05-25
filebeat keep the deleted files

rivencxl:
Found on my centos,  filebeat takes some have been deleted file handle, resulting in there are lots of ghost files and disk space is not released.

Then I read the code, and find there may be something wrong:
  Application in log rotation, will delete the old log file, or directly to compressed archive the old file, then create a new file with the same name. And filebeat is according to the file name if there is to determine whether a file is removed.
https://github.com/elastic/beats/blob/800d95f0c8362fe4ac0dc02117fa03523d03d3df/filebeat/input/log/log.go#L136-L144

## Reproduce
Using a command simulate continuous writting a log file , if delete it, will create and write again.
`while true;do echo "I'm a test log" >> /tmp/test.log && sleep 1; done`

Running filebeat with the filebeat.yml.
filebeat.yml:
```
filebeat.prospectors:

- type: log

  enabled: true

  paths:
    - /tmp/test.log

output.file:
  path: "/tmp/"
  filename: filebeat_test.log
```
run filebeat:
```
$ ./filebeat &
[1] 2485
$ ps -ef| grep filebeat
apps      2485  5223  0 16:04 pts/19   00:00:00 ./filebeat
apps      2693  5223  0 16:04 pts/19   00:00:00 grep filebeat
$ ll /proc/2485/fd
total 0
lrwx------ 1 apps apps 64 May 24 16:05 0 -> /dev/pts/19
lrwx------ 1 apps apps 64 May 24 16:05 1 -> /dev/pts/19
lrwx------ 1 apps apps 64 May 24 16:04 2 -> /dev/pts/19
lr-x------ 1 apps apps 64 May 24 16:05 3 -> /dev/urandom
lr-x------ 1 apps apps 64 May 24 16:05 4 -> eventpoll:[3219478084]
l-wx------ 1 apps apps 64 May 24 16:05 5 -> /tmp/filebeat-6.2.4-linux-x86_64/logs/filebeat
lr-x------ 1 apps apps 64 May 24 16:05 6 -> /tmp/test.log
l-wx------ 1 apps apps 64 May 24 16:05 7 -> /tmp/filebeat_test.log
```
remove /tmp/test.log or zip test.log, simulate the rotation
```
$ rm /tmp/test.log
$ ll /proc/2485/fd
total 0
lrwx------ 1 apps apps 64 May 24 16:05 0 -> /dev/pts/19
lrwx------ 1 apps apps 64 May 24 16:05 1 -> /dev/pts/19
lrwx------ 1 apps apps 64 May 24 16:04 2 -> /dev/pts/19
lr-x------ 1 apps apps 64 May 24 16:05 3 -> /dev/urandom
lr-x------ 1 apps apps 64 May 24 16:05 4 -> eventpoll:[3219478084]
l-wx------ 1 apps apps 64 May 24 16:05 5 -> /tmp/filebeat-6.2.4-linux-x86_64/logs/filebeat
lr-x------ 1 apps apps 64 May 24 16:05 6 -> /tmp/test.log (deleted)
l-wx------ 1 apps apps 64 May 24 16:05 7 -> /tmp/filebeat_test.log
lr-x------ 1 apps apps 64 May 24 16:05 8 -> /tmp/test.log
```
rm /tmp/test.log more times
image

### DIFF
--- a/filebeat/input/log/log.go
+++ b/filebeat/input/log/log.go
@@ -7,15 +7,14 @@ import (
 
 	"github.com/elastic/beats/filebeat/harvester"
 	"github.com/elastic/beats/filebeat/input/file"
+	filehelper "github.com/elastic/beats/libbeat/common/file"
 	"github.com/elastic/beats/libbeat/logp"
-
-	file_helper "github.com/elastic/beats/libbeat/common/file"
 )
 
 // Log contains all log related data
 type Log struct {
 	fs           harvester.Source
-	stateOS      file_helper.StateOS
+	stateOS      filehelper.StateOS
 	offset       int64
 	config       LogConfig
 	lastTimeRead time.Time
@@ -26,7 +25,7 @@ type Log struct {
 // NewLog creates a new log instance to read log sources
 func NewLog(
 	fs harvester.Source,
-	stateOS file_helper.StateOS,
+	stateOS filehelper.StateOS,
 	config LogConfig,
 ) (*Log, error) {
 	var offset int64
@@ -143,7 +142,7 @@ func (f *Log) errorChecks(err error) error {
 		stat, statErr := os.Stat(f.fs.Name())
 
 		// Error means file does not exist, or if inode has changed, then file was removed.
-		if statErr != nil || !file_helper.GetOSState(stat).IsSame(f.stateOS) {
+		if statErr != nil || !filehelper.GetOSState(stat).IsSame(f.stateOS) {
 			return ErrRemoved
 		}
 	}
--- a/filebeat/input/log/log.go
+++ b/filebeat/input/log/log.go
@@ -8,11 +8,14 @@ import (
 	"github.com/elastic/beats/filebeat/harvester"
 	"github.com/elastic/beats/filebeat/input/file"
 	"github.com/elastic/beats/libbeat/logp"
+
+	file_helper "github.com/elastic/beats/libbeat/common/file"
 )
 
 // Log contains all log related data
 type Log struct {
 	fs           harvester.Source
+	stateOS      file_helper.StateOS
 	offset       int64
 	config       LogConfig
 	lastTimeRead time.Time
@@ -23,6 +26,7 @@ type Log struct {
 // NewLog creates a new log instance to read log sources
 func NewLog(
 	fs harvester.Source,
+	stateOS file_helper.StateOS,
 	config LogConfig,
 ) (*Log, error) {
 	var offset int64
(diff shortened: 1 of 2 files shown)


karmi:
Hi @rivencxl, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/beats/pull/7176.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

rivencxl:
@karmi Thanks for your reminding, I have added it.

filebeat/input/log/log.go
```diff
@@ -8,11 +8,14 @@ import (
 	"github.com/elastic/beats/filebeat/harvester"
 	"github.com/elastic/beats/filebeat/input/file"
 	"github.com/elastic/beats/libbeat/logp"
+
+	file_helper "github.com/elastic/beats/libbeat/common/file"
```
jsoriano: Please use a name without underscores, and don't leave an empty line between this import and the ones for other beats packages.

andrewkroh:
Nice find!

filebeat/input/log/log.go
```diff
@@ -135,10 +139,10 @@ func (f *Log) errorChecks(err error) error {
 
 	if f.config.CloseRemoved {
 		// Check if the file name exists. See https://github.com/elastic/filebeat/issues/93
-		_, statErr := os.Stat(f.fs.Name())
+		stat, statErr := os.Stat(f.fs.Name())
 
-		// Error means file does not exist.
-		if statErr != nil {
+		// Error means file does not exist, or if inode has changed, then file was removed.
```
ruflin: If the inode of a file path has changed, it does not mean the file has been removed. It can also mean the file was removed. If you want to close a file this file you should use `close_renamed` (see check above).

ruflin:
The case you describe above should be covered by `close_renamed`. If the same file changed it's inode, it does not mean the file has been removed it could just be renamed. Perhaps I miss something here.

filebeat/input/log/log.go
rivencxl: I tested set close_renamed to true, but the result dosen't what I want： after close the deleted  files twice，with the same name. Filebeat just stop handle the new file with the same name.(It means I can't find any '/tmp/test.log' in '/proc/filebeat_pid/fd')

kvch:
We are sorry but we cannot take that contribution in at the moment as it needs an update, I will mark this pull request as stalled and we eventually close it in 2 weeks if there are no more updates.
If you decide to continue working on it later we can reopen the PR or create a new one.

PR #7179, opened 2018-05-25
Auditd module: New metric for kernel lost events

adriansr:
The `auditd` module exports a new metric, `kernel_lost`, that accounts for the events dropped by the kernel's audit framework.

The original `lost` metric has been renamed to `reassembler_lost`.

### DIFF
--- a/auditbeat/module/auditd/audit_linux.go
+++ b/auditbeat/module/auditd/audit_linux.go
@@ -27,11 +27,14 @@ const (
 
 	unicast   = "unicast"
 	multicast = "multicast"
+
+	lostEventsUpdateInterval = time.Second * 15
 )
 
 var (
-	auditdMetrics = monitoring.Default.NewRegistry(moduleName)
-	lostMetric    = monitoring.NewInt(auditdMetrics, "lost")
+	auditdMetrics         = monitoring.Default.NewRegistry(moduleName)
+	reassemblerlostMetric = monitoring.NewInt(auditdMetrics, "reassembler_lost")
+	kernelLostMetric      = monitoring.NewInt(auditdMetrics, "kernel_lost")
 )
 
 func init() {
@@ -48,9 +51,13 @@ func init() {
 // does not rely on polling.
 type MetricSet struct {
 	mb.BaseMetricSet
-	config Config
-	client *libaudit.AuditClient
-	log    *logp.Logger
+	config     Config
+	client     *libaudit.AuditClient
+	log        *logp.Logger
+	kernelLost struct {
+		enabled bool
+		counter uint32
+	}
 }
 
 // New constructs a new MetricSet.
@@ -69,7 +76,8 @@ func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
 		return nil, errors.Wrap(err, "failed to create audit client")
 	}
 
-	lostMetric.Set(0)
+	reassemblerlostMetric.Set(0)
+	kernelLostMetric.Set(0)
 
 	return &MetricSet{
 		BaseMetricSet: base,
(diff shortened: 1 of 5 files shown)


auditbeat/module/auditd/audit_linux.go
```diff
@@ -531,7 +589,7 @@ func (s *stream) ReassemblyComplete(msgs []*auparse.AuditMessage) {
 }
 
 func (s *stream) EventsLost(count int) {
-	lostMetric.Inc()
+	lostMetric.Add(int64(count))
```
adriansr: I don't completely understand the logic behind the original "lost" metric, but I thought this to be a mistake.

wdyt @andrewkroh ?

auditbeat/module/auditd/audit_linux.go
```diff
@@ -230,15 +259,35 @@ func (ms *MetricSet) initClient() error {
...
+}
+
+func (ms *MetricSet) receiveEvents(done <-chan struct{}) (<-chan []*auparse.AuditMessage, <-chan *libaudit.AuditStatus, error) {
 	if err := ms.initClient(); err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 
 	out := make(chan []*auparse.AuditMessage, ms.config.StreamBufferQueueSize)
+	statusC := make(chan *libaudit.AuditStatus, ms.config.StreamBufferQueueSize)
```
andrewkroh: I imagine someone might want to increase the their stream buffer size for events (like to 10000+). If they did, I don't think it would make sense to increase the channel size for the status events.

auditbeat/module/auditd/audit_linux.go
```diff
@@ -104,19 +112,37 @@ func (ms *MetricSet) Run(reporter mb.PushReporterV2) {
...
+
 	for {
 		select {
 		case <-reporter.Done():
 			return
+
 		case msgs := <-out:
 			reporter.Event(buildMetricbeatEvent(msgs, ms.config))
+
+		case status := <-statusC:
```
andrewkroh: Reading from `statusC` and sending the status request should probably be done from its own goroutine because when the output is blocked due to back-pressure (e.g. `reporter.Event()` is slow) then no metrics are updated.

auditbeat/module/auditd/audit_linux.go
```diff
@@ -251,6 +300,15 @@ func (ms *MetricSet) receiveEvents(done <-chan struct{}) (<-chan []*auparse.Audi
 			if err != nil {
 				continue
 			}
+			if raw.Type == auparse.AUDIT_GET {
+				status := &libaudit.AuditStatus{}
+				if err := status.FromWireFormat([]byte(raw.Data)); err == nil {
+					statusC <- status
```
andrewkroh: If for some reason `statusC` is not being drained do we want to block the reader loop? Perhaps this should be a non-blocking channel send that drops status events if the channel is full.

auditbeat/module/auditd/audit_linux.go
andrewkroh: At one point I changed the meaning behind the metric to be the number of times sequence number gaps were detected (`lostMetric.Inc()`). 

I did this because it seemed like value of `count` might not be accurate in all cases. IIRC I was seeing some wildly large `count` values.

Let's keep your change and modify the name of this metric to be `reassembler_lost`.

adriansr: I observed that audit events from the kernel always have a sequence number of zero. The only messages with a sequence number are the replies to our commands. That can cause the wrong count to be reported, yes.

I agree with the name change.

andrewkroh: Darn, I was hoping that would explain the crazy values I saw and that all we needed was to special case the zero seq in the `libaudit.Reassembler`.

PR #7188, opened 2018-05-28
Backport filebeat registry state update improvements

urso:
Backport changes from #6346 and #6347, to improve state update
performance on large registry files.

- make States type easier to find by moving into separate states.go file
- Add ID->array index, index for faster lookups on update and find
  operations. When updating states in a big registry, the Updates
  converged to quadratic complexity. The index helps in keeping the
  complexity about linear in number of state updates.
- Do not gc state if there are no 'pending' entries


urso:
Closing, as backport is not working as stable as I did hope for.

PR #7189, opened 2018-05-28
Disable modules in the filebeat.reference.yml file

ph:
If you create a new module with only a `_meta/config.yml`, the content
of this file will be used for the filebeat reference, usually this file
is uncommented.

This means the following module: Kibana, Elasticsearch, osquery, and Kafka
were enabled by default. Since the Elasticsearch module requires a
configured paths, Filebeat was throwing an error.

I've created a specific commented config.reference.yml for each of the
module.

**Questions**

Should we make the generator create a default config.reference.yml?

### DIFF
--- a/filebeat/filebeat.reference.yml
+++ b/filebeat/filebeat.reference.yml
@@ -83,10 +83,10 @@ filebeat.modules:
     #input:
 
 #---------------------------- elasticsearch Module ---------------------------
-- module: elasticsearch
+#- module: elasticsearch
   # Server log
-  server:
-    enabled: true
+  #server:
+    #enabled: true
 
     # Set custom paths for the log files. If left empty,
     # Filebeat will choose the paths depending on your OS.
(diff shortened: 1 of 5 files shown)


ruflin:
@ph This error was thrown when you tried to use the reference file with the `-c` flag? 

I would suggest to follow the example of metricbeat where all the modules are enabled in the reference file and the options are uncommented. This makes it very copy friendly.

ph:
@ruflin Exact it was with `-c`, so the reference file is not expected to be runnable right away?

I always expected it to be a more complete version of the default one but with more detailed options. 

This is how I've used it previously.

ph:
I've took a look at the auditbeat file and it seems to be runnable right away.

ruflin:
@ph Yes, that is the way we used it in the past and mostly today. It becomes a bit trickier with the modules. The reason is that modules are enabled by default if they are listed in the reference file. In metricbeat we went with the option to have them all in there and that they are enabled. But someone running the reference file would get a lot of errors because it's unlikely that all services are running.

I think what diverged is not really the reference file but the short config file. There we started to optimise for the getting started experience and overwrite some of our internal defaults to also provide some examples. 

What does this mean for the filebeat reference file: I would suggest to follow the model of the metricbeat.reference.yml file and have all modules inside and enabled with good defaults.

PR #7192, opened 2018-05-29
Update default value for flush.timeout and flush.min_events

ismael-hasan:
Checking the code, it seems the default `flush.timeout` is `1s` - https://github.com/elastic/beats/blob/v6.2.4/libbeat/publisher/queue/memqueue/config.go#L14-L18 - instead of `0s` . Change needs some dev review to confirm!

### DIFF
--- a/libbeat/docs/queueconfig.asciidoc
+++ b/libbeat/docs/queueconfig.asciidoc
@@ -74,5 +74,5 @@ The default value is 0.
 Maximum wait time for `flush.min_events` to be fulfilled. If set to 0s, events
 will be immediately available for consumption.
 
-The default values is 0s.
+The default values is 1s.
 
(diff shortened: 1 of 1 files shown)


libbeat/docs/queueconfig.asciidoc
```diff
@@ -74,5 +74,5 @@ The default value is 0.
 Maximum wait time for `flush.min_events` to be fulfilled. If set to 0s, events
 will be immediately available for consumption.
 
-The default values is 0s.
+The default values is 1s.
```
urso: `flush.min_events` should default to 2048 as well.

ismael-hasan: roger that @urso , changed that as well and corrected a typo (and the title of the issue)

PR #7196, opened 2018-05-29
Add ml_job metricset to Elasticsearch module

ruflin:
This adds a basic ml_job metricset to the Elasticsearch module.

The testing was enhanced to enabled to also test non x-pack Basic features by enabling the trial license as part of the test run.


ruflin:
This PR depends on the release of ES 6.3.

metricbeat/module/elasticsearch/ml_job/data.go
```diff
@@ -0,0 +1,67 @@
...
+
+func eventsMapping(r mb.ReporterV2, content []byte) []error {
+
+	jobsData := &jobsStruct{}
+	err := json.Unmarshal(content, jobsData)
+	if err != nil {
+		r.Error(err)
+		return []error{err}
+	}
+
```
jsoriano: Return error also if `jobsData.Jobs` is nil?

metricbeat/module/elasticsearch/test_elasticsearch.py
```diff
@@ -27,10 +29,36 @@ def test_metricsets(self, metricset):
...
                 os.getenv('ES_PORT', '9200')]
+
+    def create_ml_job(self):
+        es = Elasticsearch(self.get_hosts())
+
+        # Enable xpack trial
+        try:
+            es.transport.perform_request('POST', "/_xpack/license/start_trial?acknowledge=true")
+        except:
+            print "Trial already enabled"
```
jsoriano: We could print the error instead, in case it is caused by other reason.

metricbeat/module/elasticsearch/test_elasticsearch.py
```diff
@@ -27,10 +29,36 @@ def test_metricsets(self, metricset):
...
+
+        # Check if an ml job already exists
+        response = es.transport.perform_request('GET', "/_xpack/ml/anomaly_detectors/_all/")
+        if response["count"] > 0:
+            return
+
+        file = os.path.join(self.beat_path, "module", "elasticsearch", "ml_job", "_meta", "test", "test_job.json")
```
jsoriano: Could we use this file also for the integration test?

metricbeat/module/elasticsearch/ml_job/_meta/test/xpack.job.stats.json
```diff
@@ -0,0 +1,62 @@
...
+            "model_size_stats": {
+                "job_id": "test",
+                "result_type": "model_size_stats",
+                "model_bytes": 0,
+                "total_by_field_count": 0,
+                "total_over_field_count": 0,
+                "total_partition_field_count": 0,
+                "bucket_allocation_failures_count": 0,
+                "memory_status": "ok",
+                "log_time": 1525847933184
```
jsoriano: Are we going to collect more metrics?

metricbeat/module/elasticsearch/ml_job/ml.go
```diff
@@ -0,0 +1,78 @@
...
+
+	isMaster, err := elasticsearch.IsMaster(m.HTTP, m.HostData().SanitizedURI+jobPath)
+	if err != nil {
+		r.Error(err)
+		return
+	}
+
+	// Not master, no event sent
+	if !isMaster {
+		logp.Debug("elasticsearch", "Trying to fetch index summary stats from a none master node.")
```
ycombinator: Spelling: "none master" should be "non-master"

metricbeat/module/elasticsearch/ml_job/ml.go
```diff
@@ -0,0 +1,78 @@
...
+		return
+	}
+
+	content, err := m.HTTP.FetchContent()
+	if err != nil {
+		r.Error(err)
```
ycombinator: When X-Pack isn't installed in ES, this if condition becomes true because a 400 error code is returned by the call to the `/_xpack/ml/anomaly_detectors/_all/_stats` API. This then results in error documents being indexed in the metricbeat index that look like this:

```json
{
  "@timestamp":"2018-07-05T16:35:40.258Z",
  "metricset":{
    "rtt":6264,
    "namespace":"elasticsearch.ml.job",
    "name":"ml_job",
    "module":"elasticsearch",
    "host":"localhost:9200"
  },
  "error":{
    "message":"HTTP error 400 in ml_job: 400 Bad Request"
  },
  "beat":{
    "version":"7.0.0-alpha1",
    "name":"Shaunaks-MBP-2",
    "hostname":"Shaunaks-MBP-2"
  },
  "host":{
    "name":"Shaunaks-MBP-2"
  }
}
```

Just want to confirm that this is the behavior we want? Another option could be to log an error (but not index anything) if this API returns 400?

metricbeat/module/elasticsearch/ml_job/ml.go
ycombinator: Also, "index summary stats" should probably be "machine learning job stats".

metricbeat/module/elasticsearch/ml_job/data.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/elastic/beats/metricbeat/mb"
+)
+
+var (
+	schema = s.Schema{
+		"id":    c.Str("job_id"),
+		"state": c.Str("state"),
+		"data_counts": c.Dict("data_counts", s.Schema{
+			"processed_record_count": c.Int("processed_record_count"),
+		}),
```
ycombinator: Might be worth including `invalid_date_count` as well. From https://www.elastic.co/guide/en/elasticsearch/reference/6.2/ml-jobstats.html#ml-datacounts:

> `invalid_date_count`
>     (long) The number of records with either a missing date field or a date that could not be parsed.

metricbeat/module/elasticsearch/ml_job/_meta/docs.asciidoc
```diff
@@ -0,0 +1,3 @@
+=== elasticsearch ml MetricSet
```
ycombinator: Should this say `ml_job` instead of `ml`?

metricbeat/module/elasticsearch/elasticsearch_integration_test.go
```diff
@@ -110,31 +121,110 @@ func getConfig(metricset string) map[string]interface{} {
...
+
+	if resp.StatusCode != 200 {
+		return fmt.Errorf("HTTP error %d: %s", resp.StatusCode, resp.Status)
+	}
+
+	return nil
+}
+
+// createIndex creates and elasticsearch index in case it does not exit yet
+func enableLicense(host string) error {
```
ycombinator: Nit: Name this `enableTrialLicense` just to be more specific about what it actually does?

metricbeat/module/elasticsearch/ml_job/data.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+)
+
+type jobsStruct struct {
+	Jobs []map[string]interface{} `json:"jobs"`
+}
+
+func eventsMapping(r mb.ReporterV2, content []byte) []error {
```
ycombinator: It seems like the return value is being ignored anyway so my following comment my be moot:

Looking at other metricsets in the elasticsearch module, it looks like their `eventsMapping` functions return `error` instead of `[]error`. Further, to build this `error` return object they use `var errs multierror.Errors` to collect multiple errors and then `return `errs.Err()` at the end of the function. For consistency maybe the `eventsMapping` function here should do the same?

metricbeat/module/elasticsearch/ml_job/ml.go
ycombinator: Alternatively, could we somehow make use of the `xpack` config setting here?

ruflin: If someone enables the ml_job metricset and X-Pack is not enabled I think an error document should be returned. What we probably should improve is the error message above and have something inside that it is not enabled.

I think our `xpack` config is not related to this.

metricbeat/module/elasticsearch/ml_job/data.go
ruflin: Yes, need to change it. This changed recently with the schema change from Jaime. Will adapt it accordingly.

metricbeat/module/elasticsearch/elasticsearch_integration_test.go
```diff
@@ -110,31 +121,110 @@ func getConfig(metricset string) map[string]interface{} {
...
+		"function": "sum",
+		"field_name": "total"
+		}
+		]
+		},
+		"data_description" : {
+		"time_field":"timestamp",
+		"time_format": "epoch_ms"
+		}
+	}`
```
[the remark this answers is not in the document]
ruflin: I will use the file that I already created for the system tests ...

metricbeat/module/elasticsearch/ml_job/_meta/test/xpack.job.stats.json
ruflin: Yes. This PR mainly creates the foundation.

metricbeat/module/elasticsearch/ml_job/data.go
ruflin: I would say in case it's empty no error should be returned.

metricbeat/module/elasticsearch/test_elasticsearch.py
ruflin: done

metricbeat/module/elasticsearch/test_elasticsearch.py
ruflin: done

metricbeat/module/elasticsearch/ml_job/ml.go
ruflin: Both fixed.

metricbeat/module/elasticsearch/ml_job/data.go
ruflin: Added. In general I think we will have to add more fields but I also would like to involve the ML team here in a follow up. I don't think we should send all the stats but only the ones that are most relevant which is a tricky question to answer.

metricbeat/module/elasticsearch/ml_job/_meta/docs.asciidoc
ruflin: changed

metricbeat/module/elasticsearch/elasticsearch_integration_test.go
```diff
@@ -191,7 +180,7 @@ func createMLJob(host string) error {
 		return nil
 	}
 
-	req, err := http.NewRequest("PUT", "http://"+host+jobURL, strings.NewReader(mlJob))
+	req, err := http.NewRequest("PUT", "http://"+host+jobURL, bytes.NewReader(mlJob))
```
ycombinator: This is more of question for my learning, not a review suggestion: why is `bytes.NewReader` more appropriate here than `strings.NewReader`?

ruflin: The main difference is that one takes a `[]byte` array and the other on a  `string` as param. ReadFile above returns a `[]byte` array and previously it was a string (mlJob var).

PR #7198, opened 2018-05-29
Comply with PostgreSQL database name format

### LINKED ISSUE #7195: [Filebeat] Postgres Module fails if db name contains underscore
- Version: 6.2.4
- Operating System:  linux, GCP k8
- Steps to Reproduce: 

Postgresql [supports underscores in the dbname](https://www.postgresql.org/docs/current/static/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS).

The db name is logged and parsed by an ingest pipeline e.g.

`2018-05-29 12:01:03.172 UTC [9851] elastic@elastic_db LOG:  duration: 0.092 ms  statement: UPDATE pgbench_branches SET bbalance = bbalance + 1155 WHERE bid = 1;`

This causes the pipeline to fail which currently doesn't allow for underscores i.e.

https://github.com/elastic/beats/blob/master/filebeat/module/postgresql/log/ingest/pipeline.json#L9-L13

which uses HOSTNAME for the dbname which doesn't support _ chars. https://github.com/elastic/elasticsearch/blob/master/libs/grok/src/main/resources/patterns/grok-patterns#L30

kvch:
Support `_` and `$` in PostgreSQL database names as all of those are valid characters in a name.

Reference:
> SQL identifiers and key words must begin with a letter (a-z, but also letters with diacritical marks and non-Latin letters) or an underscore (_). Subsequent characters in an identifier or key word can be letters, underscores, digits (0-9), or dollar signs ($). Note that dollar signs are not allowed in identifiers according to the letter of the SQL standard, so their use might render applications less portable. 

(Source: https://www.postgresql.org/docs/current/static/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS)

Closes #7195

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -81,6 +81,7 @@ https://github.com/elastic/beats/compare/v6.2.3...master[Check the HEAD diff]
 - Commit registry writes to stable storage to avoid corrupt registry files. {issue}6792[6792]
 - Fix a data race between stopping and starting of the harverters. {issue}#6879[6879]
 - Fix a parsing issue in the syslog input for RFC3339 timestamp and time with nanoseconds. {pull}7046[7046]
+- Comply with PostgreSQL database name format {pull}7198[7198]
 
 *Heartbeat*
 - Fix race due to updates of shared a map, that was not supposed to be shared between multiple go-routines. {issue}6616[6616]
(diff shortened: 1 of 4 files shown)


kvch:
jenkins test this please

ruflin:
WFG

ph:
I think we should backport it to 6.3, I will backport #7269 and I would prefer we don't diverge too much :)

PR #7204, opened 2018-05-30
MINOR: Resolve string concatenation todo in packetbeat/.../http.go

original-brownbear:
Just resolving the todo now that Beats uses 1.10 :)


original-brownbear:
@adriansr thanks :)

PR #7205, opened 2018-05-30
Add response fields to Kibana module based on ECS

ruflin:
This is based on the ECS http fields: https://github.com/elastic/ecs#http


filebeat/_meta/fields.common.yml
```diff
@@ -114,3 +114,20 @@
...
         @timestamp is the processed timestamp from the log line. If both are identical
         only @timestamp should be used.
+
+    - name: http.response.status_code
+      type: long
+      description: >
+        HTTP response status_code.
+      example: 404
+
+    - name: http.response.time
```
ruflin: I'm not happy with the naming of this field as it's more the duration of the request and does not directly belong to the response?

ruflin: This is probably what we have in heartbeat under `http.rtt.total`: https://github.com/elastic/beats/blob/master/heartbeat/monitors/active/http/_meta/fields.yml#L85

ruflin: Any suggestions?

ruflin: I like the suggestion, changed to elapsed_time

PR #7206, opened 2018-05-30
[Metricbeat] Add /snap mountpoints to blacklist per default

xoh:
By design these mounts always show 100% utilization, thus the information of them being full is negligible. These changes add /snap/ to the list of mounts to be ignored

### DIFF
--- a/deploy/kubernetes/metricbeat-kubernetes.yaml
+++ b/deploy/kubernetes/metricbeat-kubernetes.yaml
@@ -60,7 +60,7 @@ data:
         - fsstat
       processors:
       - drop_event.when.regexp:
-          system.filesystem.mount_point: '^/(sys|cgroup|proc|dev|etc|host|lib)($|/)'
+          system.filesystem.mount_point: '^/(sys|cgroup|proc|dev|etc|host|lib|snap)($|/)'
   kubernetes.yml: |-
     - module: kubernetes
       metricsets:
--- a/deploy/kubernetes/metricbeat/metricbeat-daemonset-configmap.yaml
+++ b/deploy/kubernetes/metricbeat/metricbeat-daemonset-configmap.yaml
@@ -60,7 +60,7 @@ data:
         - fsstat
       processors:
       - drop_event.when.regexp:
-          system.filesystem.mount_point: '^/(sys|cgroup|proc|dev|etc|host|lib)($|/)'
+          system.filesystem.mount_point: '^/(sys|cgroup|proc|dev|etc|host|lib|snap)($|/)'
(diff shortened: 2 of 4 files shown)


andrewkroh:
What device type is associated with these mount points (it will be shown in /etc/mtab)?

I think it would be better to use `filesystem.ignore_types` for all of these mounts where possible. We should convert the existing ones over if we can because it filters them earlier causing less work to be done and removes them from the fsstats summary.

See Filtering at https://www.elastic.co/guide/en/beats/metricbeat/current/metricbeat-metricset-system-filesystem.html.

xoh:
/snap/ is of FS type squashfs<sup>[1](https://en.wikipedia.org/wiki/SquashFS)</sup>, a read-only file system. Monitoring utilization of a read-only file systems should be stopped.

FS types of other currently excluded mount points are (on my system - `Ubuntu 16.04 4.4.0-127-generic`):
`/sys` --> `sysfs`
`/cgroup` --> not present as mount point, only as mount device (1<sup>st</sup> column), within the path (e.g. `/sys/fs/cgroup/systemd`) or as FS type `cgroup`
`/proc` --> `proc`
`/dev` --> `devtmpfs`
`/etc` --> no separate partition/mount
`/host` --> no such partition/mount/directory
`/lib` --> no such partition/mount

Other FS types I would consider to be excluded are:
`binfmt_misc` used at `/proc/sys/fs/binfmt_misc/register`. Some kind of kernel feature alternative to shebangs.
`autofs` used at `/proc/sys/fs/binfmt_misc`.
`debugfs` used at `/sys/kernel/debug`.
`mqueue` used at `/dev/mqueue`. Message queues.
`devpts` used at `/dev/pts`.

I would not filter monitoring of `tmpfs` by default, as this is frequently used for caches &c.

Filtering by FS type seems the right way, but the sheer number of types would be overwhelming. The way various kernel features work and the FS type they use to mount $stuff in `/proc` or other directories is dependend on tons of things (OS, flavor, init daemon, kernel version, ...). TBH I think filtering `/proc` & co is more future proof compared to filtering `procfs` and `autofs` and `binfmt_misc` and ...

What do you think about a hybrid approach:
* Filter well-known and somewhat widespread FS types using `filesystem.ignore_types` to reduce load:
    * `proc`
    * `devtmpfs`
    * `sysfs`
    * `squashfs`
    * ...
* Filter directories known as "hosting" various FS types we do not care about via `drop_event.when.regexp`:
    * `/proc`
    * `/sys`
    * `/run`
    * ...

ph:
This seems to be an OK enhancement? Anyone integration could take a look?

PR #7207, opened 2018-05-30
Add raw log message to log handler

ruflin:
By default filebeat adds log.message to each event which contains the raw unprocessed message. This can be disabled by setting `raw_message` to false to save disk space.

The message in `log.message` contains a full multiline message, is encoded and the new line at the end is stripped.


ruflin:
@tsg I'm not sure if we should enable this by default or only have it for example in all our modules enabled.

CHANGELOG.asciidoc
```diff
@@ -178,6 +178,8 @@ https://github.com/elastic/beats/compare/v6.2.3...master[Check the HEAD diff]
 - Correctly join partial log lines when using `docker` input. {pull}6967[6967]
 - Add support for TLS with client authentication to the TCP input {pull}7056[7056]
 - Converted part of pipeline from treafik/access metricSet to dissect to improve efficeny. {pull}7209[7209]
+- Add support for TLS with client authentication to the TCP input. {pull}7056[7056]
```
kvch: This entry does not belong here.

CHANGELOG.asciidoc
```diff
@@ -178,6 +178,8 @@ https://github.com/elastic/beats/compare/v6.2.3...master[Check the HEAD diff]
...
+- Add log.message to each log event. {pull}[]
```
kvch: Please, add the number of the PR.

kvch:
I've just looked at the issue which references it. I still believe that on input level, raw should not be added to the event. But I am fine with adding this to the modules. I am keeping my review as request changes, because currently it enables keeping the raw message for inputs, not just for modules.

filebeat/harvester/reader/limit.go
```diff
@@ -18,5 +18,9 @@ func (p *Limit) Next() (Message, error) {
 	if len(message.Content) > p.maxBytes {
 		message.Content = message.Content[:p.maxBytes]
 	}
+
+	if len(message.Raw) > p.maxBytes {
+		message.Raw = message.Raw[:p.maxBytes]
```
kvch: I am on the fence with this. Raw message should not be truncated, because it is a form of processing. But this could protect users from injecting too long messages to their pipelines.

kvch: I would go with truncating the raw message.

kvch: I mean such a big message which could disrupt the output of Filebeat.

kvch:
I think we should. In case of `log` input reprocessing can be achieved by rereading the data (if it's not yet deleted). But for other inputs it might not be straightforward to replay the incoming messages. So it might be more valuable for other inputs than for `log`.

ruflin:
Closing in favor of https://github.com/elastic/beats/pull/8448

PR #7210, opened 2018-05-30
Unify handling of root and extracted key for dissect

ruflin:
Previously for keys put under root the dots were expanded to objects but the ones with a target were not. This unifies the code that all keys are prefixed.

### DIFF
--- a/libbeat/processors/dissect/processor.go
+++ b/libbeat/processors/dissect/processor.go
@@ -3,6 +3,8 @@ package dissect
 import (
 	"fmt"
 
+	"github.com/pkg/errors"
+
 	"github.com/elastic/beats/libbeat/beat"
 	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/processors"
@@ -59,12 +61,18 @@ func (p *processor) mapper(event *beat.Event, m common.MapStr) (*beat.Event, err
 	if p.config.TargetPrefix != "" {
 		prefix = p.config.TargetPrefix + "."
 	}
+	var prefixKey string
 	for k, v := range m {
-		if _, err := event.GetValue(k); err == common.ErrKeyNotFound {
-			event.PutValue(prefix+k, v)
+		prefixKey = prefix + k
+		if _, err := event.GetValue(prefixKey); err == common.ErrKeyNotFound {
+			event.PutValue(prefixKey, v)
 		} else {
 			event.Fields = copy
-			return event, fmt.Errorf("cannot override existing key: `%s`", k)
+			// When the target key exists but is a string instead of a map.
+			if err != nil {
+				return event, errors.Wrapf(err, "cannot override existing key with `%s`", prefixKey)
+			}
+			return event, fmt.Errorf("cannot override existing key with `%s`", prefixKey)
 		}
 	}
 
--- a/libbeat/processors/dissect/processor.go
+++ b/libbeat/processors/dissect/processor.go
@@ -8,11 +8,8 @@ import (
 	"github.com/elastic/beats/libbeat/processors"
 )
 
-type mapperFunc func(e *beat.Event, m common.MapStr) (*beat.Event, error)
-
 type processor struct {
 	config config
-	mapper mapperFunc
 }
 
 func init() {
@@ -26,12 +23,6 @@ func newProcessor(c *common.Config) (processors.Processor, error) {
 		return nil, err
 	}
 	p := &processor{config: config}
-
-	if config.TargetPrefix == "" {
-		p.mapper = p.mapToRoot
-	} else {
-		p.mapper = p.mapToField
-	}
 
 	return p, nil
 }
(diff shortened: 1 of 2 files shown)


ruflin:
@ph This test fails: I wonder why it worked previously? https://github.com/elastic/beats/blob/master/libbeat/processors/dissect/processor_test.go#L52 Also now it does not fail the way I would expect it and return an error, but it just does not modify the file. Did not have time yet to fully investigate.

ph:
The test failling appear to be legit, if you look at the `Event.Fields` you will get:

```
{"extracted":"not hello","message":"hello world super"}
```
The previous code was correctly nesting these values.

ruflin:
My issue is that it should return an error. As the `extracted` field is a string it should not be "deleted" and overwritten by an object and an error should be returned?

ph:
@ruflin there was a small issue with own you deal with prefix key and `event.GetValue()`.
The following gist should correctly handle the type mistmatch and return the following errors:

```
cannot override existing key with `key`
cannot override existing key with `extracted.key`: expected map but type is string
```


```diff
diff --git a/libbeat/processors/dissect/processor.go b/libbeat/processors/dissect/processor.go
index a10737245..e40829684 100644
--- a/libbeat/processors/dissect/processor.go
+++ b/libbeat/processors/dissect/processor.go
@@ -3,6 +3,8 @@ package dissect
 import (
 	"fmt"
 
+	"github.com/pkg/errors"
+
 	"github.com/elastic/beats/libbeat/beat"
 	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/processors"
@@ -59,12 +61,18 @@ func (p *processor) mapper(event *beat.Event, m common.MapStr) (*beat.Event, err
 	if p.config.TargetPrefix != "" {
 		prefix = p.config.TargetPrefix + "."
 	}
+	var prefixKey string
 	for k, v := range m {
-		if _, err := event.GetValue(k); err == common.ErrKeyNotFound {
-			event.PutValue(prefix+k, v)
+		prefixKey = prefix + k
+		if _, err := event.GetValue(prefixKey); err == common.ErrKeyNotFound {
+			event.PutValue(prefixKey, v)
 		} else {
 			event.Fields = copy
-			return event, fmt.Errorf("cannot override existing key: `%s`", k)
+			// When the target key exists but is a string instead of a map.
+			if err != nil {
+				return event, errors.Wrapf(err, "cannot override existing key with `%s`", prefixKey)
+			}
+			return event, fmt.Errorf("cannot override existing key with `%s`", prefixKey)
 		}
 	}
 
diff --git a/libbeat/processors/dissect/processor_test.go b/libbeat/processors/dissect/processor_test.go
index d96a1a321..147139b67 100644
--- a/libbeat/processors/dissect/processor_test.go
+++ b/libbeat/processors/dissect/processor_test.go
@@ -55,7 +55,7 @@ func TestProcessor(t *testing.T) {
 				"target_prefix": "extracted",
 				"field":         "message",
 			},
-			fields: common.MapStr{"message": "hello world super", "extracted": "not hello"},
+			fields: common.MapStr{"message": "hello world super"},
 			values: map[string]string{"extracted.key": "world", "extracted.key2": "super"},
 		},
 	}
@@ -109,24 +109,47 @@ func TestFieldDoesntExist(t *testing.T) {
 }
 
 func TestFieldAlreadyExist(t *testing.T) {
-	t.Run("root prefix", func(t *testing.T) {
-		c, err := common.NewConfigFrom(map[string]interface{}{
-			"tokenizer":     "hello %{key}",
-			"target_prefix": "",
+	tests := []struct {
+		name      string
+		tokenizer string
+		prefix    string
+		fields    common.MapStr
+	}{
+		{
+			name:      "no prefix",
+			tokenizer: "hello %{key}",
+			prefix:    "",
+			fields:    common.MapStr{"message": "hello world", "key": "exists"},
+		},
+		{
+			name:      "with prefix",
+			tokenizer: "hello %{key}",
+			prefix:    "extracted",
+			fields:    common.MapStr{"message": "hello world", "extracted": "exists"},
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			c, err := common.NewConfigFrom(map[string]interface{}{
+				"tokenizer":     test.tokenizer,
+				"target_prefix": test.prefix,
+			})
+
+			if !assert.NoError(t, err) {
+				return
+			}
+
+			processor, err := newProcessor(c)
+			if !assert.NoError(t, err) {
+				return
+			}
+
+			e := beat.Event{Fields: test.fields}
+			_, err = processor.Run(&e)
+			if !assert.Error(t, err) {
+				return
+			}
 		})
-		if !assert.NoError(t, err) {
-			return
-		}
-
-		processor, err := newProcessor(c)
-		if !assert.NoError(t, err) {
-			return
-		}
-
-		e := beat.Event{Fields: common.MapStr{"message": "hello world", "key": "exist"}}
-		_, err = processor.Run(&e)
-		if !assert.Error(t, err) {
-			return
-		}
-	})
+	}
 }

```

ph:
@ruflin sorry for the snippet, I to understand it in the plane :)

PR #7211, opened 2018-05-30
Make Kibana client reusable

exekias:
This change moves Kibana client to a common place in libbeat, it also makes client config public and adds support for headers passing.

### DIFF
--- a/libbeat/kibana/client_config.go
+++ b/libbeat/kibana/client_config.go
@@ -6,7 +6,8 @@ import (
 	"github.com/elastic/beats/libbeat/common/transport/tlscommon"
 )
 
-type kibanaConfig struct {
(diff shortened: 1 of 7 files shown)


libbeat/kibana/client_config.go
```diff
@@ -6,7 +6,8 @@ import (
 	"github.com/elastic/beats/libbeat/common/transport/tlscommon"
 )
 
-type kibanaConfig struct {
+// ClientConfig to connect to Kibana
+type ClientConfig struct {
```
ruflin: Is there an other config then clientConfig for Kibana? Otherwise you could just call it `config`.

ruflin:
Merged, we can still make further adjustments later.

Issue #7202, opened 2018-05-30
Filebeat TCP input panics after 2^31 events received


adriansr:
User reports a panic ("sync: negative WaitGroup counter") after a week of running filebeat. filebeat stays running but does not accept the connections.

Stacktrace:
```
github.com/elastic/beats/libbeat/logp.Recover
	/home/jason/go/src/github.com/elastic/beats/libbeat/logp/global.go:88
runtime.call32
	/usr/lib/go-1.10/src/runtime/asm_amd64.s:573
runtime.gopanic
	/usr/lib/go-1.10/src/runtime/panic.go:502
sync.(*WaitGroup).Add
	/usr/lib/go-1.10/src/sync/waitgroup.go:73
github.com/elastic/beats/filebeat/beater.(*eventCounter).Add
	/home/jason/go/src/github.com/elastic/beats/filebeat/beater/channels.go:61
github.com/elastic/beats/filebeat/channel.(*outlet).OnEvent
	/home/jason/go/src/github.com/elastic/beats/filebeat/channel/outlet.go:43
github.com/elastic/beats/filebeat/harvester.(*Forwarder).Send
	/home/jason/go/src/github.com/elastic/beats/filebeat/harvester/forwarder.go:33
github.com/elastic/beats/filebeat/input/tcp.NewInput.func1
	/home/jason/go/src/github.com/elastic/beats/filebeat/input/tcp/input.go:59
github.com/elastic/beats/filebeat/inputsource/tcp.(*client).handle
	/home/jason/go/src/github.com/elastic/beats/filebeat/inputsource/tcp/client.go:71
github.com/elastic/beats/filebeat/inputsource/tcp.(*Server).run.func1
	/home/jason/go/src/github.com/elastic/beats/filebeat/inputsource/tcp/server.go:99
```

Log (two consecutive panics, same stracktrace):

```
[...]

PR #7214, opened 2018-05-30
Filebeat: Allow stateless and stateful ACKer on the global ack handler

ph:
This commit introduces a change in how filebeat handle ACK by default,
before the ACK was using the private field of the event to retrieve a
state. The updated state was sent to the registrar, and the registrar
was finalizing the ACK.

But with the introduction of the TCP/UDP and the Redis input, the events
don't have any state attached. So in that scenario, Filebeat was not
correctly acking these events to some wait group.

The ACKer was modified to handle both stateless (default) and stateful
events, when stateful is required, the states are sent to the registry
otherwise, the waiting groups are directly updated.

Fixes: #7202

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -82,6 +82,7 @@ https://github.com/elastic/beats/compare/v6.2.3...master[Check the HEAD diff]
 - Fix a data race between stopping and starting of the harverters. {issue}#6879[6879]
 - Fix a parsing issue in the syslog input for RFC3339 timestamp and time with nanoseconds. {pull}7046[7046]
 - Comply with PostgreSQL database name format {pull}7198[7198]
+- Fix an issue with an overflowing wait group when using the TCP input. {issue}7202[7202]
 
 *Heartbeat*
 - Fix race due to updates of shared a map, that was not supposed to be shared between multiple go-routines. {issue}6616[6616]
--- a/filebeat/beater/acker.go
+++ b/filebeat/beater/acker.go
@@ -2,31 +2,41 @@ package beater
 
 import (
 	"github.com/elastic/beats/filebeat/input/file"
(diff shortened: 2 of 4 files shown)


ph:
@kvch I see your point and it is a good comment.

I think my original goal was If the type doesn't require a registrar like the Redis, TCP or UDP we should not  delegate the ack to the specialize Acker. I see a world where a source could provide their own Acker.

So actually the acker type now acts as a multiplexer of ACKs and delegates them the right acker using the type of `event.Private`. If we ever implement another source, maybe a queue, we could correctly ack the events that we read. Another way would be to add a ack handler on the client and each input have their own client.

I think maybe the problem you see is the ACKER is not generalized enough, would you be OK if we make the ACKER more generic but Keep the two acking strategy?

Pseudocode

```
m := NewActMultiplexer(DefaultStatelessAck)
m.Add(file.State, StatefulAcker)
```

ph:
` I am a bit worried about updating the WaitGroup in multiple places. `

I think this could be removed because of some changes in the pipeline.
I wanted to keep the changes small in that PR to ship it with 6.3.

Also, I am not too worried about possible problems:
- The implementation is threadsafe
- We were actually decorating it with the registrar.
- Input not requiring registrar update should still not invoke the registrar.

urso:
Changes LGTM.

The finishedLogger and the eventACKER decreasing the WaitGroup is a filebeat solution only. It is used to implement the `wait_shutdown` setting. The publisher pipeline in libbeat supports this use-case as well, but we didn't manage to replace the finishedLogger with the libbeat implementation for various reason (more refactoring required). The finishedLogger and the WaitGroup will be removed in the future + all beats will switch to the libbeat `wait_shutdown` logic.

As is done in this PR, stateless events should not be routed via the registry, as the registry introduces another potential delay/backpressure, even if not required.

PR #7221, opened 2018-05-31
Add comments to the beat template

dadoonet:
When we don't have comments or incorrect comments,
that is producing some warnings when compiling the code.

### DIFF
--- a/generator/beat/{beat}/beater/{beat}.go.tmpl
+++ b/generator/beat/{beat}/beater/{beat}.go.tmpl
@@ -64,7 +64,7 @@ func (bt *{Beat}) Run(b *beat.Beat) error {
 	}
 }
 
-// Stop is called when the beater is started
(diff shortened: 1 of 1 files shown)


generator/beat/{beat}/beater/{beat}.go.tmpl
```diff
@@ -11,13 +11,14 @@ import (
...
 
+// {Beat} configuration
 type {Beat} struct {
 	done   chan struct{}
 	config config.Config
 	client beat.Client
 }
 
-// Creates beater
+// New is called when the beater is created
```
ph: We should use the present tense and we should add a period.

```
New creates an instance of {beat}.
```

ruflin:
@dadoonet Thanks for the changes. 

We will probably need the same changes for the metricbeat generator.

PR #7223, opened 2018-05-31
Do not spam Filebeat logs with errors if harvester_limit is reached

### LINKED ISSUE #6169: filebeat harvester_limit causes log spam
**Filebeat version**: 6.1.2
**OS**: CentOS Linux release 7.4.1708 (Core)
**Steps to Reproduce**: Have lots of log files, set harvester_limit to something very low.

`harvester_limit` is a prospector configuration that limits the amount of harvesters spun up. I think the use case is to prevent filebeat from blowing up a systems resources if there's many many logs files to ingest, which is what I'm using it for.

When enabling this on a server with many many log files to ingest, it causes filebeat to spam "error" level warnings every 10 seconds for the same files. Here's a small snippet (obfuscated some)

```Jan 24 20:35:26 prod.server.hostname /usr/share/filebeat/bin/filebeat[30447]: prospector.go:437: Harvester could not be started on new file: /var/log/apache2/domlogs/website-1/website-1-ssl_log, Err: Harvester limit reached
Jan 24 20:35:26 prod.server.hostname /usr/share/filebeat/bin/filebeat[30447]: prospector.go:437: Harvester could not be started on new file: /var/log/apache2/domlogs/website/website-2, Err: Harvester limit reached
Jan 24 20:35:26 prod.server.hostname /usr/share/filebeat/bin/filebeat[30447]: prospector.go:437: Harvester could not be started on new file: /var/log/apache2/domlogs/anotherwebsite/website-3-ssl_log, Err: Harvester limit reached```

Within about 10 minutes filebeat's log file had grown to about 1 MB in size. I don't think this is a useful warning to spam about every 10 minutes.

```find /var/log/apache2/domlogs/ -type f | wc -l
436```

### PR DESCRIPTION
**Pull request by kvch**
* add special error for reaching `harvester_limit`
* report this error on debug level
* do not capitalize error messages, so it looks better in logs

Closes #6169

### DIFF
--- a/filebeat/input/log/input.go
+++ b/filebeat/input/log/input.go
@@ -631,7 +631,7 @@ func (p *Input) startHarvester(state file.State, offset int64) error {
 	err = h.Setup()
 	if err != nil {
 		p.numHarvesters.Dec()
-		return fmt.Errorf("Error setting up harvester: %s", err)
+		return fmt.Errorf("error setting up harvester: %s", err)
 	}
 
 	// Update state before staring harvester
--- a/filebeat/input/log/input.go
+++ b/filebeat/input/log/input.go
@@ -22,12 +22,15 @@ import (
 
 const (
 	recursiveGlobDepth = 8
+	harvesterErrMsg    = "Harvester could not be started on new file: %s, Err: %s"
 )
 
 var (
 	filesRenamed     = monitoring.NewInt(nil, "filebeat.input.log.files.renamed")
 	filesTruncated   = monitoring.NewInt(nil, "filebeat.input.log.files.truncated")
 	harvesterSkipped = monitoring.NewInt(nil, "filebeat.harvester.skipped")
+
+	errHarvesterLimit = errors.New("harvester limit reached")
 )
 
 func init() {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by ruflin** [comment]
Change LGTM but I think it breaks the harvester_limit system test. Probably needs some log message adjustement there.

**Comment by kvch** [comment]
@ruflin updated the system tests. failing tests in Jenkins are unrealted.

File: filebeat/tests/system/config/filebeat.yml.j2 [hunk 1]
```diff
@@ -105,6 +105,10 @@ filebeat.config.{{ reload_type|default("inputs") }}:
   {% endif -%}
 {% endif -%}
 
+{% if logging_level -%}
```
ruflin: If we add a config option for logging level we should do it here: https://github.com/elastic/beats/blob/master/libbeat/tests/system/config/libbeat.yml.j2

But I don't think this is needed as python catches the output from `-d "*"` and does not use the logs.

filebeat/tests/system/test_input.py
```diff
@@ -520,6 +520,7 @@ def test_harvester_limit(self):
             harvester_limit=1,
             close_inactive="1s",
             scan_frequency="1s",
+            logging_level="debug"
```
ruflin: I think this can be removed and it should still work.

kvch: removed

kvch:
@ruflin i think it's ready. thanks for the review.

PR #7224, opened 2018-05-31
Extract Kibana dashboards

ruflin:
Currently when a PR is opened to change or add a dashboard the PR's are hard to review even if only a small detail was changed. The reasons is that the Kibana json objects contain json as string. The content inside these strings is valid JSON if it is decoded.

With this PR the dashboards in Metricbeat are modified that they contain the full decoded JSON objects instead of the string. Additional the JSON entries are sorted. This makes also the content of the visualisations readable, will create minimal diffs and allows to even apply small fixes on the code base. In addition we can now validate if it's valid JSON and introduce automatic scripts to remove potentially unneeded fields from the dashboards to clean them up.

All the existing dashboards were decoded with the following command:
```
python ../libbeat/scripts/unpack_dashboards.py --transform=decode --glob="/Users/ruflin/Dev/gopath/src/github.com/elastic/beats/metricbeat/module/*/_meta/kibana/6/dashboard/*.json"
```

This command has to be applied to dashboards exported from Kibana before adding them to the module directory.

To make sure the import still works as before, on collection of the dashboard they are converted back into the encoded format.

This change currently only applies to Metricbeat dashboards but could be applied to all.

### DIFF
--- a/libbeat/scripts/unpack_dashboards.py
+++ b/libbeat/scripts/unpack_dashboards.py
@@ -19,7 +19,9 @@ def transform_data(data, method):
             obj["attributes"]["visState"] = method(obj["attributes"]["visState"])
 
         if "kibanaSavedObjectMeta" in obj["attributes"] and "searchSourceJSON" in obj["attributes"]["kibanaSavedObjectMeta"]:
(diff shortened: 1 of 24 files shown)


ruflin:
Here is a follow up PR to also adjust the exporter script: https://github.com/elastic/beats/pull/7241

PR #7225, opened 2018-05-31
add basic index recovery metricset

ruflin:


metricbeat/module/elasticsearch/index_recovery/index_recovery.go
```diff
@@ -0,0 +1,99 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return &MetricSet{MetricSet: ms, recoveryPath: localRecoveryPath}, nil
+}
+
+// Fetch gathers stats for each index from the _stats API
+func (m *MetricSet) Fetch(r mb.ReporterV2) {
+
+	isMaster, err := elasticsearch.IsMaster(m.HTTP, m.HostData().SanitizedURI+m.recoveryPath)
```
jsoriano: Is `recoveryPath` used to check master state?

metricbeat/module/elasticsearch/index_recovery/data.go
```diff
@@ -0,0 +1,76 @@
...
+		r.Error(err)
+		return err
+	}
+
+	for indexName, d := range data {
+		for _, data := range d["shards"] {
+			event := mb.Event{}
+			event.ModuleFields = common.MapStr{}
+			event.ModuleFields.Put("index.name", indexName)
+			event.MetricSetFields, err = schema.Apply(data)
```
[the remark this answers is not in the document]
ruflin: I added an `if err ` check and send an the error event. I wonder if this is the right thing to do here.

metricbeat/module/elasticsearch/index_recovery/data.go
```diff
@@ -0,0 +1,76 @@
...
+	var data map[string]map[string][]map[string]interface{}
+
+	err := json.Unmarshal(content, &data)
+	if err != nil {
+		r.Error(err)
+		return err
+	}
+
+	for indexName, d := range data {
+		for _, data := range d["shards"] {
```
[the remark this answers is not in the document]
ruflin: Yes

metricbeat/tests/system/config/metricbeat.yml.j2
```diff
@@ -7,6 +7,12 @@ metricbeat.modules:
   enabled: {{ m.enabled }}
   {% endif -%}
 
+  {% for items in m.metricset_options -%}
+  {% for k, v in items.items() -%}
+  {{ k }}: {{ v }}
+  {% endfor %}
+  {% endfor %}
+
```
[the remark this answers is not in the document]
ruflin: removed, didn't reliase `extra` is there.

metricbeat/module/elasticsearch/index_recovery/data.go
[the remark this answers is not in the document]
ruflin: I now skiped the error completely. I initially reported it as an event. The good thing about it is that I detected a problem I had in the schema. The bad thing was that now instead of in the logs the user would have tons of documents in his index with errors which I think is not good.

metricbeat/module/elasticsearch/index_recovery/index_recovery.go
ruflin: The full url must be passed as this is the url that is revert to afterwards again. If not passed parts of the path will go missing. This is because of a hack to reuse the http client.

metricbeat/module/elasticsearch/index_recovery/data.go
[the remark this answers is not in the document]
jsoriano: Maybe we could log the errors instead of sending them, or collect them in a `multierror` and send them all together as an only event.

metricbeat/module/elasticsearch/index_recovery/_meta/fields.yml
```diff
@@ -0,0 +1,47 @@
+- name: index.recovery
+  type: group
+  description: >
+    index
+  fields:
+    - name: id
```
ycombinator: Should this be `shard_id`?

metricbeat/module/elasticsearch/index_recovery/_meta/fields.yml
```diff
@@ -0,0 +1,47 @@
...
+      description: >
+        Source node id.
+    - name: source.host
+      type: ip
+      description: >
+        Source node host address.
+    - name: source.name
+      type: keyword
+      description: >
+        Source node name.
```
ycombinator: What about other fields such as `translog.*`, `start_time`, `stop_time`, etc.? Are you planning on adding them later?

metricbeat/module/elasticsearch/index_recovery/data.go
```diff
@@ -0,0 +1,79 @@
...
+		}),
+		"target": c.Dict("target", s.Schema{
+			"id":   c.Str("id", s.Optional),
+			"host": c.Str("host", s.Optional),
+			"name": c.Str("name", s.Optional),
+		}),
+	}
+)
+
+func eventsMapping(r mb.ReporterV2, info elasticsearch.Info, content []byte) error {
```
ycombinator: Maybe I'm missing something but at first glance it looks like `info` isn't actually being used here?

metricbeat/module/elasticsearch/index_recovery/index_recovery.go
```diff
@@ -0,0 +1,102 @@
...
+		return
+	}
+
+	content, err := m.HTTP.FetchContent()
+	if err != nil {
+		r.Error(err)
+		return
+	}
+
+	info, err := elasticsearch.GetInfo(m.HTTP, m.HostData().SanitizedURI+m.recoveryPath)
```
ycombinator: See my [other comment](https://github.com/elastic/beats/pull/7225/files#r201889653) about `info` not being used.

metricbeat/module/elasticsearch/index_recovery/_meta/fields.yml
ruflin: Add them later.

metricbeat/module/elasticsearch/index_recovery/data.go
ruflin: Ok, removed and new version pushed.

PR #7229, opened 2018-05-31
Move log/harvester/reader to reader

kvch:
This is the first step towards separating reading and parsing from specific input types of Filebeat.
* move types from `log/harvester/reader` to `reader`
* rename types to eliminate repeating names in packages and types


ph:
@kvch Can you rename the package from the hound warning to follow go convention?

ph:
@kvch actually don't change it yet, I will take a bigger look.

kvch:
I haven't renamed the packages, because we already have package names in this format. For example `add_host_metadata`.

urso:
I like this change, as this better reflects the coming plugin structure.

Regarding the packaging, we might still consider to combine multiple readers into one package. E.g. we might end up with different JSON readers like:

```
pacakge readjson

// special json reader for parsing docker container logs
func DockerLogs(source reader.Reader, stream string, partial bool) reader.Reader {
   ...
}

// combines line splitting + json parsing (one JSON document per line)
func JSONLines(source reader.Reader) reader.Reader {
}

// streams through input, reading any JSON value:
// - top level primitive typed values are parsed into string and put into a 'message' field
// - top level objects will create one event
// - top level arrays will treat each array element as a top level value (-> arrays and nested arrays will be flattened)
func JSONStream(source reader.Reader) reader.Reader {
}

```

The `strip_newline` is a very simple primitive reader manipulating strings. We might introduce others for replacing contents based on regex and others. How about a `stringread` package providing some primitive ops inspired by `strings` package. The `strip_newline.Reader` is a special case of `TrimRight(..., "\r\n")`.

Question is, what kind of granularity we want. Introducing many more operators, the amount of packages might get out of hand.

Right now the top-level reader config still uses `json.Config` and other specialized types. Next step would be to replace them with `*common.Config` and pass the raw configuration object to the reader factories.

ph:
@urso Yeah that would be the same comment for me,  I would prefer that we go with less granularity in the first phase and add more if necessary, I think we should try to combine them in the same package. So I would be +1 for _readjson_, _readline_ and maybe _accumulator_ for multiline?

For the last one I presume we might have different way to combine or accumulate events produced by readers.

Concerning the `strip_newline`, we could combine it with another reader that allow tokenizing on an arbitrary string, I believe this is a requested feature, and we could leverage the SplitFunc of the TCP input :)

ph:
@urso but, I think we can merge this as is and evaluate later if we pushed too far on the granularity. WDYT?

PR #7230, opened 2018-05-31
Fix system process CPU ticks field mapping

andrewkroh:
Enabling `process.include_cpu_ticks: true` would cause an Elasticsearch mapping conflict because the `.ticks` fields for user and system were not defined in the fields.yml.


ctindel:
Shouldn't we add some tests that expose the original issue to make sure this doesn't happen again?

andrewkroh:
@ctindel I was thinking the same thing. I updated the test case for the process metricset. It ran the updated test case without the fix and it failed (e.g. `Exception: Key 'system.process.cpu.system.ticks' found in event is not documented!`).

ctindel:
Well done!

ctindel:
@andrewkroh do you know which point release this will ship in?

ruflin:
@andrewkroh I did a rebase merge on this as it had 2 clean commits with description. But after merging I realised it would have been better as 1 commit because of the backport :-(

PR #7231, opened 2018-05-31
Add owner object info to Kubernetes metadata

exekias:
This change adds object name for the Pod owner (if any) when enriching
with kubernetes metadata.

This information is really useful when mapping running containers to
their creator workload.

Resulting metadata will now include the name of the Deployment, ReplicaSet or ReplicaSet who created the Pod (if any), ie:

```
{
  "kubernetes":{
    "container":{
      "name":"kube-state-metrics"
    },
    "labels":{
      "k8s-app":"kube-state-metrics",
      "pod-template-hash":"2035844717"
    },
    "namespace":"kube-system",
    "node":{
      "name":"minikube"
    },
    "pod":{
      "name":"kube-state-metrics-6479d88c5c-5b6cl"
    },
    "replicaset":{
      "name":"kube-state-metrics-6479d88c5c"
    }
  }
}
```

### DIFF
--- a/libbeat/autodiscover/providers/kubernetes/config.go
+++ b/libbeat/autodiscover/providers/kubernetes/config.go
@@ -16,11 +16,6 @@ type Config struct {
 	SyncPeriod     time.Duration `config:"sync_period"`
 	CleanupTimeout time.Duration `config:"cleanup_timeout"`
 
-	IncludeLabels      []string `config:"include_labels"`
-	ExcludeLabels      []string `config:"exclude_labels"`
-	IncludeAnnotations []string `config:"include_annotations"`
-	IncludePodUID      bool     `config:"include_pod_uid"`
-
 	Prefix       string                  `config:"prefix"`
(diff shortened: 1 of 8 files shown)


ruflin:
@exekias Should we add this field to ECS?

libbeat/autodiscover/providers/kubernetes/config.go
```diff
@@ -16,11 +16,6 @@ type Config struct {
 	SyncPeriod     time.Duration `config:"sync_period"`
 	CleanupTimeout time.Duration `config:"cleanup_timeout"`
 
-	IncludeLabels      []string `config:"include_labels"`
```
ProNic-QY, line 19: Why remove config for labels and annotations?

exekias: This change only moved them around in the code (while didn't change how they worked): https://github.com/elastic/beats/pull/7231/files#diff-f7068996b1041934b4fa71b630725b55R20

PR #7238, opened 2018-06-01
 Allow to override the `ignore_above` option for field of the type  keyword.

ph:
By default all fields of the type keyword were limitted to 1024
characters, this PR keep the default but allow developers to redefine that limit using the
`ignore_above` directive in the field.yml.

Ref: https://github.com/elastic/ecs/pull/6

### DIFF
--- a/libbeat/template/processor.go
+++ b/libbeat/template/processor.go
@@ -14,6 +14,7 @@ type Processor struct {
 
 var (
 	defaultScalingFactor = 1000
+	defaultIgnoreAbove   = 1024
 )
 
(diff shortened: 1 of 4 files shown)


libbeat/template/processor.go
```diff
@@ -150,11 +150,14 @@ func (p *Processor) keyword(f *common.Field) common.MapStr {
 	defaultFields = append(defaultFields, fullName)
 
 	property["type"] = "keyword"
-	property["ignore_above"] = 1024
+	if f.IgnoreAbove != "" {
```
ruflin: Could you add a test for this?

PR #7239, opened 2018-06-01
Add support to export dashboard to Beat

ruflin:
Currently to export a Dashboard from the Kibana exporter API it has to be done manually or the Beats repository must be used. To simplify the exporting of own dashboards the command `beat export dashboard --id="dashboard-id"` is added. This should allow all users to export their own dashboards.

The output is written to stdout, so the expected usage is to pipe it into a file inside the `kibana/dashboards/6` directory.

For the Kibana connection settings it uses the setting from `setup.kibana.*`. If none are set it uses the defaults.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -147,6 +147,7 @@ https://github.com/elastic/beats/compare/v6.2.3...master[Check the HEAD diff]
 - Add support for keyword multifields in field.yml. {pull}7131[7131]
 - Add dissect processor. {pull}6925[6925]
 - Add owner object info to Kubernetes metadata. {pull}7231[7231]
+- Add beat export dashboard command. {pull}7239[7239]
 
 *Auditbeat*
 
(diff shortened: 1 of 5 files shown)


ruflin:
@dedemorton I was not sure where I should put the docs for this one. Suggestions?

dedemorton:
@ruflin There are a few places where we should talk about exporting dashboards:

- Update the content about the `export` command [here](https://www.elastic.co/guide/en/beats/filebeat/current/command-line-options.html): Make sure you also update the short description at the beginning of the topic. (I wish asciidoc had support for proper railroad track style syntax diagrams because one would be useful here. The examples help with the syntax, tho, so make sure you add one for `export dashboards`.)

- In the Beats Developer Guide, add a new section called something like "Exporting dashboards to stdout or a file" to the existing topic [here](https://www.elastic.co/guide/en/beats/devguide/current/export-dashboards.html). You'll need to rework the intro a bit.

- In a couple of places where we talk about the dashboards (for example, [here](https://www.elastic.co/guide/en/beats/filebeat/current/view-kibana-dashboards.html) and [here](https://www.elastic.co/guide/en/beats/filebeat/current/load-kibana-dashboards.html)), add a tip to tell users that they can export the dashboards after customizing them, and point to the developer guide for more info.

TBH, I think we have overloaded the getting started topics in the reference guide. We should have a separate section that tells users how to work with the assets (like dashboards) that Beats deploys.  Customizing and exporting our dashboards isn't just a developer task, I think. Someday....

dedemorton:
note that I've created a related enhancement request here: https://github.com/elastic/beats/issues/7243

ruflin:
@dedemorton Thanks for the pointers. Will open a follow up PR with the docs changes.

PR #7241, opened 2018-06-01
 Update export_dashboard script to export full dashboard json doc

ruflin:
In #7239 support to export a dashboard is a added to each Beat. This is expected to be used by users. The export_dashboard script from the Beats repository is expected by the Devs and contributors which want to add new dashboards.

In #7224 the dashboards are modified to be stored with decoded json objects for better versioning. This change modifies the export dashboard script so it generates the same decoded output so no additional conversion is needed.

Note: The export_dashboard script could need some cleanup but I didn't tackle this in this PR and leave it for later to keep the change to a minimum.


ruflin:
Rebased and ready for review.

PR #7246, opened 2018-06-01
Fix conditional coding and keystore CLI example

dedemorton:
Fixes conditional coding introduced by backport (https://github.com/elastic/beats/commit/29df70d700ee513c328ccf9ec734e7f196ea910f#diff-dbc3375d3e67b2b366ab28f712c00f9d)--`if/else` statements are not valid in asciidoc. The docs in master and 6.3 are OK.

Also fixes the keystore example that shows how to specify a key on the command line (adds backslash to prevent env var evaluation by the shell). This commit needs to be backported/forwardported to other branches.

### DIFF
--- a/libbeat/docs/command-reference.asciidoc
+++ b/libbeat/docs/command-reference.asciidoc
@@ -11,7 +11,6 @@
 
 
 // These attributes are used to resolve short descriptions
-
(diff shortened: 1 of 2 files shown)


ph:
LGTM

ph:
oops didn't see it was assigned to @andrewkroh

PR #7249, opened 2018-06-02
Forward port to master: Fix keystore example (#7246)

dedemorton:
Cherry-picks the following PR into master:

https://github.com/elastic/beats/pull/7246

### DIFF
--- a/libbeat/docs/keystore.asciidoc
+++ b/libbeat/docs/keystore.asciidoc
@@ -35,7 +35,7 @@ For example, imagine that the keystore contains a key called `ES_PWD` with the
 value `yourelasticsearchpassword`:
 
 * In the configuration file, use `output.elasticsearch.password: "${ES_PWD}"`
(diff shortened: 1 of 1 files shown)


libbeat/docs/keystore.asciidoc
```diff
@@ -35,7 +35,7 @@ For example, imagine that the keystore contains a key called `ES_PWD` with the
 value `yourelasticsearchpassword`:
 
 * In the configuration file, use `output.elasticsearch.password: "${ES_PWD}"`
-* On the command line, use: `-E "output.elasticsearch.password=${ES_PWD}"`
+* On the command line, use: `-E "output.elasticsearch.password=$\{ES_PWD}"`
```
ph: @dedemorton I just realize I've made a mistake in the original review, we need to escape the `$` not the `{`:

The following will work `-E "output.elasticsearch.password=\${ES_PWD}"` or use single quote.

dedemorton: Yah that was a typo. Thanks for catching this!

PR #7253, opened 2018-06-04
Metricbeat: Collect accumulated docker network metrics

### LINKED ISSUE #7240: Fields docker.network.out.bytes and docker.network.in.bytes doesn't report what they say
The fields `docker.network.out.bytes` and `docker.network.in.bytes` are actually reporting the average number of bytes per seconds for the duration, but the docs state that they're totals. Some kind of total is what I intuitively would expect from the field names, but docs doesn't say if the totals are supposed to be totals since container or metricbeat started or if theyre the total amount since the previous value was reported.

For monitoring in Elastic Cloud we need the total amount transferred and estimating that based on average speed is cumbersome and error prone. Ideally we want the amount since the previous value.

jsoriano:
Collect accumulated docker network metrics and mark previous metrics as deprecated.

Fixes #7240


metricbeat/module/docker/network/_meta/fields.yml
```diff
@@ -33,8 +34,55 @@
...
+            Total number of dropped outgoing packets.
+        - name: errors
+          type: long
+          description: >
+            Total errors on outgoing packets.
+        - name: packets
+          type: long
+          description: >
+            Total number of outgoing packets.
+    - name: inbound
```
ruflin: To follow ECS, should it be on the root level? Meaning having it under `network.inbound.bytes` ?

jsoriano: Umm, you are right, I was misinterpreting ECS for this case, would it be ok to have this without the `docker` preffix?

ruflin: If the content of the field matches ECS I would say yes.

jsoriano:
@ruflin I have done the changes for more proper ECS, this would probably require its own PR, but lets discuss all together before.

libbeat/_meta/fields.ecs-ext.yml
```diff
@@ -0,0 +1,26 @@
+- key: ecs-ext
+  title: ECS extensions
+  description: >
+    Contains extensions to ECS (https://github.com/elastic/ecs)
```
jsoriano: I have added this for common fields that are not (yet) in ECS, but they could be in the future. Changes here should be followed by discussions/PRs in ECS repo.
We could also use the commons file for that, but I think that it's fine to leave the common file for things needed by beats but that don't make sense in ECS.

libbeat/_meta/fields.ecs.yml
```diff
@@ -0,0 +1,55 @@
+- key: ecs
```
jsoriano: This file should probably be automatically generated from https://github.com/elastic/ecs/tree/master/schemas...

metricbeat/module/docker/network/data.go
```diff
@@ -5,32 +5,70 @@ import (
...
+		eventMapping(r, &netsStats)
 	}
-	return myEvents
 }
 
-func eventMapping(stats *NetStats) common.MapStr {
-	event := common.MapStr{
-		mb.ModuleDataKey: common.MapStr{
-			"container": stats.Container.ToMapStr(),
+func eventMapping(r mb.ReporterV2, stats *NetStats) {
```
jsoriano: Three events would be created now:
* Incoming traffic
* Outgoing traffic
* Legacy event

libbeat/_meta/fields.ecs.yml
ruflin: Agree. Let's keep it manual for now to not add fields we don't use at the moment.

metricbeat/module/docker/network/_meta/fields.yml
```diff
@@ -33,8 +34,9 @@
             Total number of incoming packets.
     - name: out
       type: group
+      deprecated: true
```
ruflin: We should start making use of this flag in the docs.

metricbeat/module/docker/network/data.go
```diff
@@ -5,32 +5,70 @@ import (
...
+	container := stats.Container.ToMapStr()
+	container.Put("runtime", "docker")
+
+	// Inbound traffic event
+	r.Event(mb.Event{
```
ruflin: I would still expect that 1 event is sent out with all the stats inside. It seems now we send out 3 events? Why not combine inbound and outbound?

metricbeat/module/docker/network/data.go
```diff
@@ -5,32 +5,70 @@ import (
...
+		RootFields: common.MapStr{
+			"container": container,
```
ruflin: Not sure I would mix the `container` to the root level into this PR.

metricbeat/module/docker/network/network.go
```diff
@@ -45,12 +44,13 @@ func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
 }
 
 // Fetch methods creates a list of network events for each container.
-func (m *MetricSet) Fetch() ([]common.MapStr, error) {
+func (m *MetricSet) Fetch(r mb.ReporterV2) {
```
ruflin: yay, v2 reporter.

metricbeat/module/docker/network/data.go
jsoriano: I did that initially, but then I saw that we have a `network.direction` field in ECS, then I though that according to ECS we'd have different events for inbound and outbound traffic. Not sure what would be the use case for this field if not used for something like this.

metricbeat/module/docker/network/data.go
jsoriano: The idea is to move the `container` fields to ECS structure too.

metricbeat/module/docker/network/network.go
jsoriano: Essential for migration to ECS :smile:

ruflin:
As discussed yesterday, lets move the ECS part to an other PR and continue at the moment with just adding the new fields. I still would like to see the change to v2 reporter ;-)

jsoriano:
I have reverted the changes for ECS, but kept the change to v2 reporter, I popened #7301 to continue discussion on ECS. I think we should backport this for 6.3.1.

metricbeat/module/docker/network/_meta/fields.yml
```diff
@@ -33,8 +34,55 @@
...
-        Outgoing network stats.
+        Outgoing network stats per second.
+      fields:
+        - name: bytes
+          type: long
+          format: bytes
+          description: >
+            Total number of outgoing bytes.
+        - name: dropped
+          type: scaled_float
```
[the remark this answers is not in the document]
jsoriano: Yes, all these fields should be long, I'll change it.

ruflin:
@jsoriano Do you want to backport it to 6.3?

jsoriano:
I have opened backport (#7363), but this is not a fix. In any case all related changes are already in 6.3, so I think it can make sense to merge this too.

PR #7258, opened 2018-06-04
Cherry-pick #7214 to 6.3: Filebeat: Allow stateless and stateful ACKer on the global ack handler

tsg:
Cherry-pick of PR #7214 to 6.3 branch. Original message: 

This commit introduces a change in how filebeat handle ACK by default,
before the ACK was using the private field of the event to retrieve a
state. The updated state was sent to the registrar, and the registrar
was finalizing the ACK.

But with the introduction of the TCP/UDP and the Redis input, the events
don't have any state attached. So in that scenario, Filebeat was not
correctly acking these events to some wait group.

The ACKer was modified to handle both stateless (default) and stateful
events, when stateful is required, the states are sent to the registry
otherwise, the waiting groups are directly updated.

Fixes: #7202


urso:
@tsg Changelog requires cleanup

PR #7265, opened 2018-06-05
Unify Kibana dashboard directory logic across Beats

ruflin:
Currently all dashboards end up under `_meta/kibana` in each Beat. The problem with this directory is that for some Beats it contains the automatically collected and generated files and for others the original. This prevents us from having decoded JSON files for all dashboards across Beats.

An other benefit of having the `kibana.generated` directory is that now `_meta/kibana` can become the place for dashboards across modules in one Beat without having to worry about overwriting / removing these dashboards.

Community Beats must convert their dashboards with the script under #7224.

Further changes:

* Decode Packetbeat and Winlogbeat dashboards
* Move encoding step to libbeat so it is applied automatically to all dashboards in the kibana.generated directory.
* Remove _meta/kibana from .gitignore as it can now be used to contain dashboards in all Beats.
* Adjust Makefiles for the new logic.
* Adjust tests to depend on kibana.generated directory

### DIFF
--- a/filebeat/Dockerfile
+++ b/filebeat/Dockerfile
@@ -4,7 +4,7 @@ MAINTAINER Nicolas Ruflin <ruflin@elastic.co>
 RUN set -x && \
     apt-get update && \
     apt-get install -y --no-install-recommends \
-         netcat python-pip virtualenv && \
+         netcat python-pip rsync virtualenv && \
     apt-get clean
 
 RUN pip install --upgrade setuptools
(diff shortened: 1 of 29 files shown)


filebeat/Dockerfile
```diff
@@ -4,7 +4,7 @@ MAINTAINER Nicolas Ruflin <ruflin@elastic.co>
 RUN set -x && \
     apt-get update && \
     apt-get install -y --no-install-recommends \
-         netcat python-pip virtualenv && \
+         netcat python-pip rsync virtualenv && \
```
jsoriano: `rsync` added but not used?

ruflin: Yes and now. It was used when `make update` was called inside the container. Now it happens outside again. The reason I would like to keep it in is that if someone wants to tests things from inside the container and run `make update` rsync will be needed for Filebeat.

PR #7274, opened 2018-06-06
Rename http.response.status to http.response.status_code to align with ECS

ruflin:
See https://github.com/elastic/ecs/pull/4


ruflin:
CI failure should not be related.

PR #7281, opened 2018-06-06
Keep different registry entry per container stream

### LINKED ISSUE #7045: filebeat docker input stores wrong offset
- Version: 6.2.4
- Operating System:  CentOS 7
- Steps to Reproduce: 

I'm running 2 docker inputs (for stdout and stderr) for the same container log file and all working good until the restart of filebeat after which offset gets wrong number of bytes saved, and docker JSON decoding puts the error:
<pre>
2018-05-08T13:50:25.423Z	ERROR	log/harvester.go:243	Read line error: decoding docker JSON: invalid character 'R' looking for beginning of value; File: %!(EXTRA string=/var/lib/docker/containers/6a394f8e4332087bce1fb95223aad238dc6bc27c7cebe2f493909a3734c7e0c0/6a394f8e4332087bce1fb95223aad238dc6bc27c7cebe2f493909a3734c7e0c0-json.log)
</pre>

<pre>

2018-05-08T13:50:25.423Z	DEBUG	[harvester]	log/harvester.go:468	Stopping harvester for file: /var/lib/docker/containers/6a394f8e4332087bce1fb95223aad238dc6bc27c7cebe2f493909a3734c7e0c0/6a394f8e4332087bce1fb95223aad238dc6bc27c7cebe2f493909a3734c7e0c0-json.log

2018-05-08T13:50:25.423Z	DEBUG	[harvester]	log/harvester.go:478	Closing file: /var/lib/docker/containers/6a394f8e4332087bce1fb95223aad238dc6bc27c7cebe2f493909a3734c7e0c0/6a394f8e4332087bce1fb95223aad238dc6bc27c7cebe2f493909a3734c7e0c0-json.log

2018-05-08T13:50:25.423Z	DEBUG	[harvester]	log/harvester.go:348	Update state: /var/lib/docker/containers/6a394f8e4332087bce1fb95223aad238dc6bc27c7cebe2f493909a3734c7e0c0/6a394f8e4332087bce1fb95223aad238dc6bc27c7cebe2f493909a3734c7e0c0-json.log, offset: 20350
</pre>

<pre>
# dd if=/var/lib/docker/containers/6a394f8e4332087bce1fb95223aad238dc6bc27c7cebe2f493909a3734c7e0c0/6a394f8e4332087bce1fb95223aad238dc6bc27c7cebe2f493909a3734c7e0c0-json.log bs=1 skip=20350 count=100
RV_USER=\"$3\"\\nSERV_PASS=\"$4\"\\nDB_FOLDER=${5-\"mysqldumps\"}\\nPREFIX_LOCAL=${6-\"/home/httpd\"
</pre>
and the error:
<pre>
2018-05-08T13:50:25.423Z	ERROR	log/harvester.go:243	Read line error: decoding docker JSON: invalid character 'R' looking for beginning of value; File: %!(EXTRA string=/var/lib/docker/containers/6a394f8e4332087bce1fb95223aad238dc6bc27c7cebe2f493909a3734c7e0c0/6a394f8e4332087bce1fb95223aad238dc6bc27c7cebe2f493909a3734c7e0c0-json.log)
</pre>
infinitely repeated.

But when switching input for the same container to "log" error appears one more time and after that all working as expected.

filebeat config:
<pre>
filebeat.autodiscover:
  providers:
  - type: docker
    templates:
    - condition:
        contains:
          docker.container.labels.log_type: test
      config:
      - type: docker
        containers:
          ids:
          - "${data.docker.container.id}"
          stream: stdout
        processors:
         - add_docker_metadata: ~
      - type: docker
        containers:
          ids:
          - "${data.docker.container.id}"
          stream: stderr
        multiline.pattern: '^\['
        multiline.negate: true
        multiline.match: after
        processors:
         - add_docker_metadata: ~
output.logstash:
  hosts:
  - "logstash:5044"

</pre>

exekias:
This PR adds a metadata map to prospectors state. Adding metadata to states allows having a different state per unique metadata for the same file.

This change is used by the `docker` prospector to have different states per stream, as some users configure a container prospector per stream (`stdout` & `stderr`).

Probably fixes #7045

### DIFF
--- a/filebeat/input/file/state.go
+++ b/filebeat/input/file/state.go
@@ -2,7 +2,6 @@ package file
 
 import (
 	"os"
-	"reflect"
 	"strconv"
 	"strings"
 	"time"
@@ -73,5 +72,8 @@ func (s *State) IsEqual(c *State) bool {
 
 // IsEmpty returns true if the state is empty
 func (s *State) IsEmpty() bool {
-	return reflect.DeepEqual(*s, State{})
+	return s.FileStateOS == file.StateOS{} &&
+		s.Source == "" &&
+		s.Meta == nil &&
+		s.Timestamp.IsZero()
 }
--- a/filebeat/input/file/state.go
+++ b/filebeat/input/file/state.go
@@ -1,27 +1,32 @@
 package file
 
 import (
+	"fmt"
 	"os"
+	"reflect"
 	"time"
+
+	"github.com/mitchellh/hashstructure"
 
 	"github.com/elastic/beats/libbeat/common/file"
 )
 
 // State is used to communicate the reading state of a file
 type State struct {
-	Id          string        `json:"-"` // local unique id to make comparison more efficient
-	Finished    bool          `json:"-"` // harvester state
-	Fileinfo    os.FileInfo   `json:"-"` // the file info
-	Source      string        `json:"source"`
-	Offset      int64         `json:"offset"`
-	Timestamp   time.Time     `json:"timestamp"`
-	TTL         time.Duration `json:"ttl"`
-	Type        string        `json:"type"`
+	Id          string            `json:"-"` // local unique id to make comparison more efficient
+	Finished    bool              `json:"-"` // harvester state
+	Fileinfo    os.FileInfo       `json:"-"` // the file info
+	Source      string            `json:"source"`
+	Offset      int64             `json:"offset"`
+	Timestamp   time.Time         `json:"timestamp"`
+	TTL         time.Duration     `json:"ttl"`
+	Type        string            `json:"type"`
+	Meta        map[string]string `json:"meta"`
 	FileStateOS file.StateOS
 }
 
 // NewState creates a new file state
-func NewState(fileInfo os.FileInfo, path string, t string) State {
+func NewState(fileInfo os.FileInfo, path string, t string, meta map[string]string) State {
 	return State{
 		Fileinfo:    fileInfo,
 		Source:      path,
@@ -30,6 +35,7 @@ func NewState(fileInfo os.FileInfo, path string, t string) State {
 		Timestamp:   time.Now(),
 		TTL:         -1, // By default, state does have an infinite ttl
 		Type:        t,
+		Meta:        meta,
 	}
 }
 
@@ -37,7 +43,8 @@ func NewState(fileInfo os.FileInfo, path string, t string) State {
 func (s *State) ID() string {
 	// Generate id on first request. This is needed as id is not set when converting back from json
 	if s.Id == "" {
-		s.Id = s.FileStateOS.String()
+		h, _ := hashstructure.Hash(s.Meta, nil)
+		s.Id = s.FileStateOS.String() + fmt.Sprintf("-%d", h)
 	}
 	return s.Id
 }
@@ -49,5 +56,5 @@ func (s *State) IsEqual(c *State) bool {
 
 // IsEmpty returns true if the state is empty
 func (s *State) IsEmpty() bool {
-	return *s == State{}
+	return reflect.DeepEqual(*s, State{})
 }
--- a/filebeat/input/file/state.go
+++ b/filebeat/input/file/state.go
@@ -1,9 +1,10 @@
 package file
 
 import (
-	"fmt"
 	"os"
 	"reflect"
+	"strconv"
+	"strings"
 	"time"
 
 	"github.com/mitchellh/hashstructure"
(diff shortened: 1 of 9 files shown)


exekias:
Hound wanted me to rename `Id` field to `ID`, but that would spawn a broader changelog, I don't think it's worth it, as this field wasn't added on this PR.

filebeat/input/file/state.go
```diff
@@ -30,14 +35,16 @@ func NewState(fileInfo os.FileInfo, path string, t string) State {
...
 	}
 }
 
 // ID returns a unique id for the state as a string
 func (s *State) ID() string {
 	// Generate id on first request. This is needed as id is not set when converting back from json
 	if s.Id == "" {
-		s.Id = s.FileStateOS.String()
+		h, _ := hashstructure.Hash(s.Meta, nil)
+		s.Id = s.FileStateOS.String() + fmt.Sprintf("-%d", h)
```
urso: ouch... we optimized ID generation by removing Sprintf and reduce allocations. The ID generation used to be one of the worst offenders in CPU profiles. This change makes it worse again. How about:

```
if s.Meta == nil {
  s.Id = s.FileStateOS.String()
} else {
  hashValue, _ := hashstructure.Hash(s.Meta, nil)
  var hashBuf [17]byte
  hash, _ := strconv.AppendUint(h[:0], hashValue, 16)
  hash = append(hash, '-')

  fileID := s.FileStateOS.String()  

  var b strings.Builder
  b.Grow(len(hash) + len(fileID))
  b.Write(hash)
  b.WriteString(fileID)

  s.Id = b.String()
}
```

This still does execute an extra alloc, but only in the docker case and should be less expensive.

filebeat/input/file/state.go
```diff
@@ -49,5 +56,5 @@ func (s *State) IsEqual(c *State) bool {
 
 // IsEmpty returns true if the state is empty
 func (s *State) IsEmpty() bool {
-	return *s == State{}
+	return reflect.DeepEqual(*s, State{})
```
urso: we need `DeepEqual`?

urso: How about `tmp := *s; tmp.Meta = nil; return tmp == State{} && len(s.Meta) == 0`

filebeat/input/file/state.go
exekias: sounds like a good idea, pushing this change

filebeat/input/file/state.go
exekias: The compiler still says it cannot compare maps, are your concerns about performance? I could do manual comparison for all fields, or use 2 structs like:

```
type State struct {
   Common
   Meta        map[string]string `json:"meta"`
}
```

then:

```
if len(s.Meta) > 0 { return false }
return *s.Common == Common{}
```

urso: a little performance, but I'd also prefer to keep out `reflect` from most packages.

I wonder if checking for FileStateOS or the timestamp only would be enough. This method is used to determine if we have a valid/usable state or not.

exekias: I don't have enough experience with all the possible situations with the registry so I'll trust you here on what to do :innocent:

exekias: I updated this to use `s.FileStateOS != file.StateOS{} && !s.Timestamp.IsZero()`, also pinging @ruflin for awareness

exekias:
This should be ready for another review, thanks for all the help @urso!

exekias:
jenkins, retest this please

exekias:
jenkins, retest this please

exekias:
jenkins, retest this please

filebeat/tests/system/test_registrar.py
```diff
@@ -1506,3 +1506,58 @@ def test_registrar_files_with_input_level_processors(self):
...
+
+    def test_registrar_meta(self):
+        """
+        Check that multiple entries for the same file are on the registry when they have
+        different meta
+        """
+
+        self.render_config_template(
+            type='docker',
+            input_raw='''
```
urso: btw. check out the `textwrap` package. This allows you to indent multiline strings within the code more properly.

urso:
LGTM and filebeat CI is green. Waiting for review from @ruflin

filebeat/input/file/state.go
```diff
@@ -49,5 +72,8 @@ func (s *State) IsEqual(c *State) bool {
 
 // IsEmpty returns true if the state is empty
 func (s *State) IsEmpty() bool {
-	return *s == State{}
+	return s.FileStateOS == file.StateOS{} &&
```
ruflin: Can you share some details here on why not only add s.Meta but also the other chnages?

exekias: As I added a map to the `State` struct, a simple comparison is no longer possible (Go doesn't implement map compare), I had to update the code to do an individual comparison on the struct fields

exekias:
Docker `json-file` format stores both `stderr` and `stdout` in the same file. They use a JSON object per line, with the following schema:

```
{"log":"example line\n","stream":"stdout","time":"2018-04-13T08:38:08.395077018Z"}
{"log":"example line\n","stream":"stderr","time":"2018-04-13T08:38:08.395077018Z"}
```

Our `docker` input can parse that and demultiplex the 2 different streams, as some times you want to do different processing on them (for instance, pass them to 2 different filesets)

PR #7287, opened 2018-06-07
Ensure node and pod info are always present in state_container

exekias:
This PR fixes info retrieval for state_container after latest Prometheus helper changes.

I've added code to create repeatable expected files (sorted), this should avoid big diffs in the future when we regenerate them.


ruflin:
@exekias Do you want to have the 2 commits merged or not? I assume we don't need a changelog as it's not release yet?

exekias:
Yes, the 2 commits are fine, a rebase would make more sense this time. As you said, this is fixing master for an unreleased change, no need for a changelog.

Thanks for having a look!

PR #7293, opened 2018-06-11
Allow user to rely on Elasticsearch default ignore above for keyword

ph:
value.

Allow users to rely on Elasticsearch ignore_above default for the keyword
data type, sometimes we want to have a large keyword to hold
certificates.

Behavior:

- Not specifying a value will use 1024.
- Will use the specified value.
- Using -1 will fallback to ES default.

### DIFF
--- a/libbeat/template/processor.go
+++ b/libbeat/template/processor.go
@@ -151,9 +151,12 @@ func (p *Processor) keyword(f *common.Field) common.MapStr {
 	defaultFields = append(defaultFields, fullName)
 
 	property["type"] = "keyword"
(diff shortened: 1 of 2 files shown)


ph:
Linked to discussion in https://github.com/elastic/ecs/pull/6#issuecomment-396328772

ruflin:
Now that we start to use fields.yml outside beats, it perhaps makes sense to soon move this code out of beats and into a library. There we could also not have any defaults but they would be set by Beats when using the lib.

ph:
@ruflin I think this would be a good idea but  it is a low hanging fruit for now.

PR #7294, opened 2018-06-12
Use KEY_WOW64_64KEY to access MachineGuid

andrewkroh:
KEY_WOW64_64KEY on a 64-bit OS means that the registry access, no matter if it's a 32 or 64 bit process, will access the 64 bit registry view.

This fixes an issue where the Metricbeat windows-service metricset would not start if the 32-bit binary was used on a 64-bit machine. The flag will have no effect on 32-bit machines.


ruflin:
Worth a backport?

PR #7296, opened 2018-06-12
Clarify docs around x-pack monitoring

ruflin:
When the Elasticsearch output is used, no additional hosts can be configured in the X-pack configs.

### DIFF
--- a/libbeat/docs/monitoring/monitoring-beats.asciidoc
+++ b/libbeat/docs/monitoring/monitoring-beats.asciidoc
@@ -46,7 +46,8 @@ xpack.monitoring:
   elasticsearch:
 --------------------
 
-Otherwise, you must specify additional configuration options. For example:
+If you are using a different output then Elasticsearch, you must specify
+additional configuration options. For example:
 
 [source, yml]
 --------------------
@@ -57,6 +58,10 @@ xpack.monitoring:
     username: beats_system
     password: beatspassword
 --------------------
+
+NOTE: If the Elasticsearch output is used, no additional hosts can be
+configured. At the moment it is required that the Monitoring data is sent to the
+same cluster as all the other events.
 
 --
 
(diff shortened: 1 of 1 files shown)


libbeat/docs/monitoring/monitoring-beats.asciidoc
```diff
@@ -46,7 +46,8 @@ xpack.monitoring:
   elasticsearch:
 --------------------
 
-Otherwise, you must specify additional configuration options. For example:
+If you are using a different output then Elasticsearch, you must specify
```
dedemorton: I thinks these details are still a little hard to follow. Are you saying that users *must* send monitoring data to the production cluster? If so, you should also modify the sentence that begins on line 37 (remove "and you want to use the same {es} production cluster and credentials"). So I would change this to say something like:

```
. Add the `xpack.monitoring` settings in the {beatname_uc} configuration file. If you configured {es} output,
specify the following minimal configuration:
.
.
.
If you configured a different output, such as {ls}, you must specify additional configuration options.
For example:
.
.
.
NOTE: Currently you must send monitoring data to the same cluster as all other events. If you configured the
{es} output, do not specify additional hosts in the monitoring configuration. 
```

ruflin: Thanks for the proposal. Adjusted accordingly.

libbeat/docs/monitoring/monitoring-beats.asciidoc
```diff
@@ -33,20 +33,17 @@ information, see
...
-options:
+. Add the `xpack.monitoring` settings in the {beatname_uc} configuration file. If you
+configured {es} output, specify the following minimal configuration:
 +
 --
 [source, yml]
 --------------------
-xpack.monitoring:
-  enabled: true
-  elasticsearch:
```
ruflin: @urso Any reason the `elasticsearch` part was in before?

PR #7298, opened 2018-06-12
Close changelog for 5.6.10

tsg:
Can be merged, we don't need to wait for the release day.


ruflin:
CI is more red then I like but not sure if it is worth fixing. I think the issue is that we added some more beats like heartbeat to test inside environments which are not in the 5.6 branch.

PR #7301, opened 2018-06-12
Add common fields files for ECS

jsoriano:
There will be three fields with common fields in beats:
* common: Beats common fields that don't make sense in ECS
* ecs: ECS fields
* ecs-ext: Fields that are not in ECS but that are candidate to be
there.

Continues from #7253, network and container fields present in ecs files left there from the previous PR.


auditbeat/docs/fields.asciidoc
```diff
@@ -3042,6 +3035,184 @@ type: object
...
+--
+
+[[exported-fields-ecs]]
+== ECS fields
+
+Contains ECS fields (https://github.com/elastic/ecs)
+
+
+
+*`container.runtime`*::
```
ruflin: Are we using the `container` fields in any beat already? If yes, I would expect them to be removed somewhere. If not, should we already add them?

PR #7302, opened 2018-06-12
Adds table header to command summary table

dedemorton:
Added a header and converted the list to table format to remove any ambiguity wrt what the list contains.

Output looks like this:

image


ruflin:
Nice, I like it.

PR #7305, opened 2018-06-12
Create jvm fileset

inqueue:
Adding fileset JVM GC logs for Elasticsearch Filebeat module. This baseline fileset does not do much yet, more to come. 

https://github.com/elastic/beats/issues/5301.


inqueue:
Inadvertently left out the multiline config.

filebeat/module/elasticsearch/jvm/test/gc.jdk8.624.log
```diff
@@ -0,0 +1,702 @@
+2018-06-11 01:51:14 GC log file created /var/log/elasticsearch/gc.log.4
```
ruflin: As this log contains `jdk` in the name I'm now thinking if this is actually a jdk log and not jvm. I'm confused :-)

ruflin:
Note: Let's discuss the above quickly but will also have a second review on the metricset itself so we can get CI green.

filebeat/module/elasticsearch/jvm/test/gc.jdk8.624.log
inqueue: This log came from a running JDK (as opposed to JRE) and should be renamed to `gc.log`.

filebeat/module/elasticsearch/gc/ingest/pipeline.json
```diff
@@ -0,0 +1,37 @@
+{
+  "description": "Pipeline for parsing Elasticsearch JVM garbage collection logs",
+  "processors": [
+    {
+      "grok": {
+        "field": "message",
+        "patterns": [
+          "%{TIMESTAMP_ISO8601:timestamp}",
+          "%{GREEDYMULTILINE:elasticsearch.gc.message}"
```
ruflin: I would suggest to put this under just `message`. It means we will overwrite what is in message at the moment. If we want to keep it, we could use `log.message`.

filebeat/module/elasticsearch/module.yml
```diff
@@ -1,2 +1,4 @@
 dashboards:
 
+- id: Filebeat-elasticsearch-gc-Dashboard
```
ruflin: This should be removed as there are no dashboards yet.

filebeat/module/elasticsearch/gc/manifest.yml
```diff
@@ -0,0 +1,11 @@
+module_version: 1.0
+
+var:
+  - name: paths
+    default:
+      - /var/log/elasticsearch/gc.log.[0-9]*
```
ruflin: What is the name of the first log? If it is called `gc.log` I think the pattern will not match it.

filebeat/module/elasticsearch/gc/ingest/pipeline.json
ruflin: I think the missing `elasticsearch.gc.message` part in the expected log file is the reason CI is failing.

filebeat/module/elasticsearch/gc/_meta/fields.yml
```diff
@@ -0,0 +1,16 @@
+- name: gc
+  type: group
+  description: >
+  fields:
+    - name: relative_process_timestamp_secs
```
ruflin: These should adhere to our naming conventions: https://www.elastic.co/guide/en/beats/devguide/current/event-conventions.html We can also tackle this in a follow up PR.

filebeat/module/elasticsearch/gc/config/gc.yml
```diff
@@ -0,0 +1,15 @@
...
+{{ range $i, $path := .paths }}
+ - {{$path}}
+{{ end }}
+exclude_files: [".gz$"]
+multiline:
+  pattern: '^[0-9]{4}-[0-9]{2}-[0-9]{2}'
+  negate: true
+  match: after
+
+  fields:
```
ruflin: Here the indentation seems to be off `fields` and `fields_under_root` is probably not applying because of this.

filebeat/module/elasticsearch/gc/ingest/pipeline.json
```diff
@@ -0,0 +1,39 @@
+{
+  "description": "Pipeline for parsing Elasticsearch JVM garbage collection logs",
+  "processors": [
+    {
+      "grok": {
+        "field": "message",
+        "patterns": [
+          "%{GCTIMESTAMP}: %{GCPROCRUNTIME}: Total time for which application threads were stopped: %{BASE10NUM:elasticsearch.gc.threads_total_stop_time_secs} seconds, Stopping threads took: %{BASE10NUM:elasticsearch.gc.stopping_threads_time_secs} seconds",
```
ruflin: @ph Looks like we could this in dissect?

filebeat/module/elasticsearch/gc/manifest.yml
ruflin: Got it, then the pattern makes sense.

filebeat/module/elasticsearch/gc/_meta/fields.yml
inqueue: Sure, we can fix on a follow up PR.

inqueue:
Fixed two indentation issues @ruflin. Should be good?

ruflin:
@inqueue Thanks a lot for the PR. Now lets start to iterate on top of it :-) Really happy to have this in.

filebeat/docs/fields.asciidoc
```diff
@@ -800,6 +800,42 @@ elasticsearch Module
...
+
+
+
+
+*`elasticsearch.gc.relative_process_timestamp_secs`*::
++
+--
+type: float
+
+Garbage collection threads total stop time seconds.
```
danielmitterdorfer: @inqueue Just one clarification on the meaning of those: The metrics that are captured here are not necessarily tied to garbage collection. These times are actually related to safepoints. At a safepoint all application threads are halted and the JVM can do any administrative work that requires that the application cannot interfere. That might include such things as deoptimization, biased locking or garbage collection. 

The first metric tells for how long the application threads have been halted (the duration of the pause due to the safepoint) and the second one tells how long it took from the point in time the JVM requested a safepoint until all application threads could been halted.

inqueue: Thanks @danielmitterdorfer, that makes sense. I did want to capture that metric in this initial commit as it is probably the most frequently logged in gc.log. Can we work with you to isolate the most meaningful metrics? We also intend to incorporate GC metrics captured in the Kafka Filebeat module- https://www.elastic.co/blog/monitoring-kafka-with-elastic-stack-1-filebeat.

danielmitterdorfer: > Can we work with you to isolate the most meaningful metrics?

Yes. The most meaningful ones in my opinion are:

* Absolute timestamp of an event
* Type of pause (young GC, old GC)
* Pause time (how long was the application stopped?), GC (phase) duration (might happen concurrently)
* Amount of memory cleaned up

Although I think it will get very tricky to parse all those variations:

* The log file format has changed with JDK 9 (so called "unified logging").
* There are different garbage collection algorithms (e.g. CMS, G1)

https://www.youtube.com/watch?v=M9AYK5PtX40 should give a good idea about this.

danielmitterdorfer: You're welcome. Glad I could help.

PR #7309, opened 2018-06-13
Improve Elasticsearch server fileset

ruflin:
* Add `service.name: elasticsearch` already on the Beats side.
* Modify ingest processor to use `event.created` for the collection timestamp instead of read_timestamp.

### DIFF
--- a/filebeat/module/elasticsearch/server/config/log.yml
+++ b/filebeat/module/elasticsearch/server/config/log.yml
@@ -8,3 +8,8 @@ multiline:
   pattern: '^\[[0-9]{4}-[0-9]{2}-[0-9]{2}'
   negate: true
   match: after
+
+fields:
+  service.name: "elasticsearch"
+
+fields_under_root: true
--- a/filebeat/module/elasticsearch/server/ingest/pipeline.json
+++ b/filebeat/module/elasticsearch/server/ingest/pipeline.json
@@ -10,6 +10,12 @@
     ],
     "processors": [
         {
+            "rename": {
+                "field": "@timestamp",
+                "target_field": "event.created"
+            }
+        },
+        {
             "grok": {
                 "field": "message",
                 "pattern_definitions": {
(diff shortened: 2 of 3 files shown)


filebeat/module/elasticsearch/server/ingest/pipeline.json
```diff
@@ -22,29 +28,8 @@
...
-        {
-            "remove": {
-                "field": "timestamp"
-            }
-        },
-        {
-            "append": {
-                "field": "service.name",
-                "value": "elasticsearch"
+                "target_field": "@timestamp"
```
ruflin: @tsg @ph This is needed because it seems inside grok I can't use `@timestamp`. Sound like the `@` symbol has it's issue.

filebeat/module/elasticsearch/server/config/log.yml
```diff
@@ -8,3 +8,8 @@ multiline:
   pattern: '^\[[0-9]{4}-[0-9]{2}-[0-9]{2}'
   negate: true
   match: after
+
+fields:
```
ruflin: This is meant as an example / suggestion on how we should add fields in the future. 

Instead of using `fields` I would prefer if we would have a processor for it. @urso WDYT?

ruflin:
Failure should not be related.

filebeat/module/elasticsearch/server/ingest/pipeline.json
ph: Yeah, I think you have a limited set of characters for grok keys.
Also, I haven't tested it personally but usually the extracted date isn't a 1:1 match, maybe this is why I never trip into it. :)

Also, did you get an error or you were not able to access the key?

filebeat/module/elasticsearch/server/config/log.yml
ph: So we would remove fields completely and have a `addCustomFieldProcessor` or `addFieldProcessor`?

If this is the case I would be +1 to do it.

filebeat/module/elasticsearch/server/ingest/pipeline.json
ruflin: I did get an error from Elasticsearch, something like key not found.

filebeat/module/elasticsearch/server/config/log.yml
ruflin: Exactly. We could add the processor in the 6.x cycle and deprecated fields. Then it's replaced in 7.0. With a processor we could also define things like if fields should be overwritten etc.

ph: +1 for flexibility, can you create an issue for it?

ruflin: Done: https://github.com/elastic/beats/issues/7350

PR #7315, opened 2018-06-13
Add warning about autodiscover and template scoping

### LINKED ISSUE #7312: Clarify free use of wildcards in Filebeat autodiscover docs
When using autodiscover with Filebeat, you  are allowed to use wildcards (`*`) in input configurations, we should clarify docs to explain the risk of doing so, if paths are not scoped to the container.

exekias:
When using autodiscover, you have to be careful when defining config templates, especially if they are
reading from places holding information for several containers. For instance, under this file structure:

```
/mnt/logs/<container_id>/*.log
```

You can define a config template like this:

:warning: Wrong settings:
```
autodiscover.providers:
  - type: docker
    templates:
      - condition.contains:
          docker.container.image: nginx
        config:
          - type: log
            paths:
              - "/mnt/logs/*/*.log"
```

That would read all the files under the given path several times (one per nginx container). What you really want is to scope your template to the container that matched the autodiscover condition:

:green_heart: Good settings:
```
autodiscover.providers:
  - type: docker
    templates:
      - condition.contains:
          docker.container.image: nginx
        config:
          - type: log
            paths:
              - "/mnt/logs/${data.docker.container.id}/*.log"
```

I'm not sure if I explained this well in the text, so any input is welcomed :)

Closes #7312

### DIFF
--- a/libbeat/docs/shared-autodiscover.asciidoc
+++ b/libbeat/docs/shared-autodiscover.asciidoc
@@ -62,6 +62,16 @@ For example, with the example event, "`${data.port}`" resolves to `6379`.
 
 include::../../{beatname_lc}/docs/autodiscover-docker-config.asciidoc[]
 
+[WARNING]
+=======================================
+Something to take into account when defining config templates is that they will be launched for every container
+matching the given condition. In practice, this means that your configuration template should be scoped to the
+container. You can do so by using `${data.*}` fields in the config definition. For instance, use `${data.host}`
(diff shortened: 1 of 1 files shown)


ruflin:
@exekias I like a lot what you put in the PR message. I wonder if we should put this into the docs?

exekias:
updated to include the PR message

ruflin:
Worth backporting?

exekias:
I opened it, just in case ^

PR #7316, opened 2018-06-13
more dynamic mappings for numeric object types

graphaelli:
This change adds support for generating dynamic templates with field mapping types for `byte, float, short`.

Given `fields.yml` with:
```
    - name: foo
      type: object
      object_type: float
      object_type_mapping_type: long
      dynamic: true
```

Current code will ignore this definition.  This update generates this dynamic template:

```
{
  "foo": {
    "mapping": {
      "type": "float"
    },
    "match_mapping_type": "long",
    "path_match": "foo.*"
  }
}
```


exekias:
jenkins, retest this please

ph:
I've rekicked the travis job, seems a network hiccups when installing some debian packages.

PR #7321, opened 2018-06-13
Metricbeat: Canonicalize MBean names in Jolokia module

jsoriano:
We were setting the `canonicalNaming` option to `false` in requests to
Jolokia, assuming that it should leave MBean names in responses as they
are. We need MBeans in request and in responses to be the same for
mappings to work.

But in some scenarios, as when using wildcards, Jolokia is responding with
canonicalized names, what breaks mappings.

This change canonicalizes names from config only for mappings and the
pre-built request. If we see that after setting `canonicalNaming` to `true`
there is some scenario in which Jolokia replies with non-canonicalized
names, then we'd have to canonicalize also MBeans for responses, or rethink
how we do the mapping.

### DIFF
--- a/metricbeat/module/jolokia/jmx/data.go
+++ b/metricbeat/module/jolokia/jmx/data.go
@@ -8,6 +8,7 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/logp"
 )
 
 const (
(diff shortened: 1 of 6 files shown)


jsoriano:

metricbeat/module/jolokia/jmx/data.go
```diff
@@ -149,7 +151,10 @@ func parseResponseEntry(
 ) error {
 	field, exists := mapping.Get(requestMbeanName, attributeName)
 	if !exists {
-		return errors.Errorf("metric key '%v' for mbean '%s' not found in mapping (%+v)", attributeName, requestMbeanName, mapping)
+		// This shouldn't ever happen, if it does it is probably that some of our
+		// assumptions when building the request and the mapping is wrong.
+		logp.Debug("jolokia.jmx", fmt.Sprintf("mapping: %+v", mapping))
```
ruflin: you don't need the `fmt.Sprintf` part here, you can write:

```
logp.Debug("jolokia.jmx", "mapping: %+v", mapping)
```

ph: Or you can use `Debugw` with a the structured logger with a "jolokia" context.

jsoriano: Much better without `Sprintf` yes, thanks :)

PR #7331, opened 2018-06-14
Allow to specify auditd rules in separate files

### LINKED ISSUE #7113: [Auditbeat] Allow audit rules to be specified in their own files
As a user I'd like to be able to specify my audit rules in dedicate file(s).

- Audit rules can be specified in their own file without having to worry about YAML string rules.
- Audit rules can be split across multiple files (e.g. `/etc/auditbeat/audit.rules.d/*.conf`) that are loaded in lexicographical order.

- Version: Auditbeat 6.2
- Operating System: Linux
- Discuss Forum URL: https://discuss.elastic.co/t/auditbeat-conf-d-files/130661/5?u=andrewkroh

adriansr:
Added a new configuration field, `audit_rule_files` to the auditd
module. It allows to specify a set of files that contain audit rules.

```
- module: auditd
  audit_rule_files: [ '${path.config}/auditd.rules.d/*.conf' ]
```

Rules will be loaded into the kernel in the following order:
1. Rules specified in the `audit_rules` field at auditbeat.yml.
2. Rules from files in the `audit_rule_files`, in the order they are
   listed.
3. If a path contains a glob, the resulting files are loaded in
   lexicographical order.

Closes #7113

### DIFF
--- a/auditbeat/module/auditd/_meta/docs.asciidoc
+++ b/auditbeat/module/auditd/_meta/docs.asciidoc
@@ -136,6 +136,11 @@ embedded in the string using `#` as a prefix. The format for rules is the same
 used by the Linux `auditctl` utility. {beatname_uc} supports adding file watches
 (`-w`) and syscall rules (`-a` or `-A`).
 
+*`audit_rule_files`*:: A list of files to load audit rules from. This files are
+loaded after the rules declared in `audit_rules` are loaded. Wildcards are
+supported and will expand in lexicographical order. The format is the same as
+that of the `audit_rules` field.
+
 [float]
 === Audit rules
 
--- a/auditbeat/module/auditd/config.go
+++ b/auditbeat/module/auditd/config.go
@@ -20,6 +20,11 @@ package auditd
 import (
 	"bufio"
 	"bytes"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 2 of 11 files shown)


exekias:
I recently unified how config reload works: https://github.com/elastic/beats/pull/7172

It would make sense to use or imitate this API, it would be especially useful to have support for file reloading, autodiscover or central management in the future. What do you think?

auditbeat/module/auditd/config.go
```diff
@@ -39,17 +46,40 @@ type Config struct {
...
+	Warnings:               false,
+	ReassemblerMaxInFlight: 50,
+	ReassemblerTimeout:     2 * time.Second,
+	StreamBufferQueueSize:  8192,
+	StreamBufferConsumers:  0,
+}
+
 // Validate validates the rules specified in the config.
 func (c *Config) Validate() error {
+
```
andrewkroh: Shall we drop this line? :smile:

auditbeat/module/auditd/_meta/docs.asciidoc
```diff
@@ -158,6 +161,10 @@ Defining any audit rules in the config causes {beatname_uc} to purge all
 existing audit rules prior to adding the rules specified in the config.
 Therefore it is unnecessary and unsupported to include a `-D` (delete all) rule.
 
+The rules declared in the `audit_rules` field are loaded first. Then the files
```
andrewkroh: I think I would move this above with the reference information under `audit_rule_files`.

andrewkroh:
> It would make sense to use or imitate this API, it would be especially useful to have support for file reloading, autodiscover or central management in the future.

@exekias I'm not 100% sure what you mean. Were you thinking that the auditd module should reuse the config reloading code? Or that it would be useful to have a generic file reloading capability (since the files here are not standard Beat YAML config files)?

I do think that we'll see requests for auto-reloading of the audit rules when they change.

exekias:
Agreed, audit rules reloading is going to be really nice!

So https://github.com/elastic/beats/pull/7172 unifies how configs are reloaded for modules and inputs. The problem is split into 2: config files handling (detect changes, load them), and config reloading (take the new configs and do the proper changes to reach the desired state).

The second part may be useful here, and somehow prepares Auditbeat for the future of central management and autodiscover.

This is the current interface implementing config reloading: https://github.com/elastic/beats/pull/7172/files#diff-ab62770abd1b3493ace2c1a6c5426f0eR44 is used both by autodiscover & file config reload mechanisms as of today.

Happy to take this one into a call if needed. Also as I said, if you see no fit for these APIs I wouldn't block this PR, we can have a look to this in the future.

adriansr:
@exekias @andrewkroh after looking at the required changes for config reloading I see that the required refactor is too big and I prefer not to block this PR, and have a separate PR for the config reloading.

ruflin:
+1 on moving forward.

@andrewvc FYI: Seems we probably have one more implementation for config reloading from file.

adriansr:
@andrewkroh I addressed your comment, please review again

PR #7335, opened 2018-06-14
libbeat: Refactor error handling in schema.Apply()

jsoriano:
`schema.Apply()` returned a `schema.Errors` object as error, with this
object is not easy to check if there was really an error, specially when
optional fields are involved. This leads to lots of cases where these
errors were being ignored.

This change introduces a series of changes to have better control on the
errors happening during the application of an schema.

The interface is kept as `event, err := schema.Apply(data)`, but now the
error returned is always a plain error in case of one or more errors
happened, or nil if no error happened.

`schema.ApplyTo()` returns the list of errors, and can be used where
multiple schemas are applied for the same event.

`Apply()` and `ApplyTo()` can now receive a list of options, these
options are functions that can be used to filter the errors returned or
to access the unfiltered list of errors.  This allows different
combinations of resulting errors in case of optional or required fields.
It also allows to add additional checks on all the errors happened, even
on the ones ignored for the final result. Three options are added by now:
* `AllRequired` to fail if any non-optional field is missing, this
mimics the current behaviour, and is set by default if no other option
is set.
* `FailOnRequired` to fail if any required field is missing. A new
`Required` option is also added to be able to declare required fields.
* `NotFoundKeys(cb func([]string))` is an option builder that receives
a function, this function is called with the list of missing keys, so
additional custom checks or tests can be done.

For lists of errors, the `multierror` library is used now, some custom
errors have been added to help on previous modifications and to improve
some error messages. Errors on structured data contain now the full
paths of the fields.


ruflin:
+100 on that change. Now that we have a result struct we can also add potentially more / other data in the future if we need it.

ruflin:
Two other options I was thinking of on what we could return:

```
event, result, err := schema.Apply()
if err != nil {
   return err
}
```
It would make error checking more go like and often people would probably skip the result.

```
result, err := schema.Apply()
if err != nil {
   return err
}
result.Event ...
```

An other option is to have event as part of the result. When I read the code I asked myself why is event not part of the result as it's kind of the result.

I'm good with all 3 option as all 3 are much better then we have now. Thanks for coming up with this.

jsoriano:
The main reason why I chose to keep the event as first returned value and the "result" as second was to keep it more similar to the current interface so it continue working the same as before for the majority of cases where the error is being ignored.

But it is true that in any case the signature is changing and maybe it is a good idea to force the review of cases where the error is being ignored. I like the `result, err := schema.Apply()` form, I'll give a try to this.

libbeat/common/schema/mapstriface/mapstriface.go
```diff
@@ -70,26 +72,30 @@ type ConvMap struct {
...
+		return multierror.Errors{err}
 	}
 
 	subEvent := common.MapStr{}
-	convMap.Schema.ApplyTo(subEvent, subData)
+	_, errors := convMap.Schema.ApplyTo(subEvent, subData)
+	for _, err := range errors {
+		if err, ok := err.(schema.KeyError); ok {
+			err.SetKey(convMap.Key + "." + err.Key())
+		}
```
jsoriano: This is to have more meaningful errors on nested fields, previously a missing key would report something like `Key "segments" not found`, now it reports `key "total.segments" not found`.

metricbeat/module/elasticsearch/index_summary/data.go
```diff
@@ -16,68 +16,67 @@ var (
 			"docs": c.Dict("docs", s.Schema{
 				"count":   c.Int("count"),
 				"deleted": c.Int("deleted"),
-			}),
+			}, c.DictOptional),
```
jsoriano: Added these optional modifiers because there are test files that don't contain these fields.

metricbeat/module/elasticsearch/pending_tasks/data_test.go
```diff
@@ -80,24 +66,18 @@ func TestInvalidJsonForRequiredFieldShouldThrowError(t *testing.T) {
 	content, err := ioutil.ReadFile(file)
 	assert.NoError(t, err)
 
-	_, errs := eventsMapping(content)
-
-	errors, ok := errs.(*s.Errors)
-	if ok {
-		assert.True(t, errors.HasRequiredErrors(), "mapping error: %s", errors)
-		assert.EqualError(t, errors, "Required fields are missing: ,source")
```
jsoriano: @ruflin this is the only place where I have seen that we'd make use of `ApplyResult.NotFound`, to check if `source` is in the list of missing fields, but to take advantage of this we'd have to complicate the `eventsMapping` function to return the result and not only the event and the errors. Also, I think that here it'd be enough with testing that it returns an error for the case of having a missing field, going further looks like if we were testing here the behaviour of `schema.Apply`, that is already being tested in the `schema` package.

Given that, I wonder again if it worths it to have the `ApplyResult` object. From the perspective of a consumer of the API, it is cleaner to just have `event, err := schema.Apply(data)`, with an event and an error that can be used directly. I think it is also more valuable to be able to give directly the semantics of the use case, as in `event.MetricSetFields, err = schema.Apply(index)` or `nodeData, _ := schemaXpack.Apply(node)` instead of obtaining the result and then assigning `result.Event` to a meaningful variable. From our perspective, if we want to test more particular cases of missing fields, I think that a better place would be the `schema` package.

I'd go back to the approach in #7167, but with the additional improvements in error handling added here too.

metricbeat/module/elasticsearch/pending_tasks/data_test.go
```diff
@@ -80,24 +66,18 @@ func TestInvalidJsonForRequiredFieldShouldThrowError(t *testing.T) {
...
-	} else {
-		t.Error(err)
-	}
+	_, err = eventsMapping(content)
+	assert.Error(t, err)
 }
 
 func TestInvalidJsonForBadFormatShouldThrowError(t *testing.T) {
+	t.Skip("the code is not really checking for uknown fields")
+
```
jsoriano: This test was passing because `schema.Apply()` returned the empty `schema.Errors`, it was not detecting the unknown field in the (valid) JSON file. This test can be probably removed if the missing field is optional, if is not optional, an explicit check should be added in `eventsMapping` to check that it exists.

metricbeat/module/kibana/stats/data.go
```diff
@@ -31,8 +31,8 @@ var (
 			},
 		}),
 		"requests": c.Dict("requests", s.Schema{
-			"total":       c.Int("total"),
-			"disconnects": c.Int("disconnects"),
+			"total":       c.Int("total", s.Optional),
+			"disconnects": c.Int("disconnects", s.Optional),
```
jsoriano: These fields don't exist in the test input:
```
	Error:      	Received unexpected error:
	            	2 errors: key `requests.total` not found; key `requests.disconnects` not found
	Test:       	TestStats
	Messages:   	_meta/test/stats.700.json
```

libbeat/common/schema/mapstriface/mapstriface_test.go
```diff
@@ -103,6 +103,122 @@ func TestConversions(t *testing.T) {
...
+		_, err := c.Schema.Apply(c.Input)
+		assert.Error(t, err, c.Description)
+		assert.Contains(t, err.Error(), c.Expected, c.Description)
+
+		_, errs := c.Schema.ApplyTo(common.MapStr{}, c.Input)
+		assert.Equal(t, 1, len(errs), c.Description)
+		keyErr, ok := errs[0].(s.KeyError)
+		if assert.True(t, ok, c.Description) {
+			assert.Equal(t, c.Expected, keyErr.Key(), c.Description)
+		}
```
jsoriano: `ApplyTo` can be used as a more advanced method for special cases, it still returns the full list of errors, that can be further analyzed if needed.

jsoriano:

metricbeat/module/elasticsearch/pending_tasks/data_test.go
ruflin: I can see that from a user perspective it is much more convenient if the event can be assigned directly. It's also the reason why I started to think about 3 arguments.

For testing for the missing fields which are optional I think this is something specific to metricsets and not necessarly internal testing of Apply. The reason it only exists in one place so far I think is more that it was too complicated to do it. When we start to add more versions and have very subtle differences between each version I think tests for missing fields become more common. 

I assume moving forward we will make optional fields the default as we discussed in the past and not required. In this scenario most fields will be optional. So I our testing will potentially more look like: max 3 missing fields and we get an error if there are more.

metricbeat/module/elasticsearch/pending_tasks/data_test.go
jsoriano: Assuming this is not optional, explicit check added and skip removed.

metricbeat/module/elasticsearch/pending_tasks/data_test.go
jsoriano: I have changed the approach again, to `event, err := schema.Apply(data, options...)`, `options` is optional, and can be used to select the mode of operation of `Apply` (so for example we can make all fields optional by default). Or to access unfiltered errors, so additional tests can be done on the missing fields for example. The test that was checking the missing fields has been refactored to this approach.

I have updated the description of the PR accordingly.

metricbeat/module/elasticsearch/index_summary/data.go
jsoriano: Instead of that, added option to `Apply` to fail only on required fields.

metricbeat/module/elasticsearch/index_summary/data.go
```diff
@@ -66,25 +66,24 @@ var (
...
 
-func eventMapping(r mb.ReporterV2, info elasticsearch.Info, content []byte) []error {
+func eventMapping(r mb.ReporterV2, info elasticsearch.Info, content []byte) error {
 	var all struct {
 		Data map[string]interface{} `json:"_all"`
 	}
 
 	err := json.Unmarshal(content, &all)
 	if err != nil {
 		r.Error(err)
```
ruflin: One nice thing that we can do now is that we can check for the error from `eventMapping` and report it in only 1 place. The current code with reporting and returning the error now is only returning the error as there is only 1 type of errors left.

metricbeat/module/elasticsearch/pending_tasks/data.go
```diff
@@ -29,28 +31,34 @@ var (
 	schema = s.Schema{
 		"insert_order":     c.Int("insert_order"),
 		"priority":         c.Str("priority"),
-		"source":           c.Str("source"),
+		"source":           c.Str("source", s.Required),
```
ruflin: How did you pick this field as required?

jsoriano: There is a test failing if this field is not present, but I'm thinking now that all fields here seem to be required, I'm going to revert this part.

jsoriano:
@exekias @andrewkroh if you have time, could you also take a look to this PR? Thanks!

PR #7342, opened 2018-06-15
Improve and separate Filebeat module testing

ruflin:
Currently the module, ml and pipeline tests were in the same file. When trying to run only 1 module test, pipeline and ml were always executed too. This makes separation easier. Also for the module tests only Elasticsearch is needed, for the ml also Kibana. So the setup can be simplified for each.

In addition for the ml testing parameterized tests are used to have a separate output directory for each option. This makes it easier to figure out which test failed if one failed.


ruflin:
Failing test should not be related.

PR #7346, opened 2018-06-15
Fix autodiscover hint names

exekias:
I misspelled hint names in my previous PR :man_facepalming:


exekias:
^ I directly opened the backport

PR #7352, opened 2018-06-18
Fix duplication of dynamic fields on reconnect

ruflin:
If `setup.template.overwrite: true` is set the template is overwritten each time the Beat reconnects. The `dynamicTemplates` are a global variable that keeps state. This had the side effect that each time the template was generated for loading, the dynamic fields were append again. The logic in the code is changed now that each time when the template is generated first the dynamicFields array is reset.

This also works with the new `append_fields` config option because it is read during the load process to also potentially add dynamic fields.

The same applies to the default fields which are global. Also this array is reset. As default_fields were only used for Elasticsearch 7.0 is does not need an entry in the changelog.

If possible in the future both variables should be part of the template object instead of being global. Unfortunately this would required major changes.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -71,6 +71,7 @@ https://github.com/elastic/beats/compare/v6.2.3...master[Check the HEAD diff]
 - Do not emit Kubernetes autodiscover events for Pods without IP address. {pull}7235[7235]
 - Allow to override the `ignore_above` option when defining new field with the type keyword. {pull}7238[7238]
 - Allow index-pattern only setup when setup.dashboards.only_index=true. {pull}7285[7285]
+- Fix duplicating dynamic_fields in template when overwriting the template. {pull}7352[7352]
 
 *Auditbeat*
 
--- a/libbeat/template/template.go
+++ b/libbeat/template/template.go
@@ -97,6 +97,9 @@ func New(beatVersion string, beatName string, esVersion string, config TemplateC
 
 func (t *Template) load(fields common.Fields) (common.MapStr, error) {
 
+	dynamicTemplates = nil
+	defaultFields = nil
+
 	var err error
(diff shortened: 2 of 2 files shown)


CHANGELOG.asciidoc
```diff
@@ -71,6 +71,7 @@ https://github.com/elastic/beats/compare/v6.2.3...master[Check the HEAD diff]
 - Do not emit Kubernetes autodiscover events for Pods without IP address. {pull}7235[7235]
 - Allow to override the `ignore_above` option when defining new field with the type keyword. {pull}7238[7238]
 - Allow index-pattern only setup when setup.dashboards.only_index=true. {pull}7285[7285]
+- Fix duplicating dynamic_fields in template when overwriting the template. {pull}[]
```
tsg: Please add the PR number.

ruflin: added

ruflin:
jenkins, test this

ruflin:
jenkins, test this

libbeat/template/template.go
```diff
@@ -97,6 +97,9 @@ func New(beatVersion string, beatName string, esVersion string, config TemplateC
 
 func (t *Template) load(fields common.Fields) (common.MapStr, error) {
 
+	dynamicTemplates = nil
```
andrewkroh: With this function being called each time that a client connects to Elasticsearch I would expect that this could lead to data races (especially if the number of works is greater than 1). So because these are globals I think a mutex is required for reads and writes.

ruflin: I added a lock directly above for the `load` method. All access to dynamicTemplates and defaultFields happens inside the processor and generator which both are triggered here. Generate could also be called from outside but in this case it's one of the Cobra commands so will not overlap with the running beat and only happening once.

PR #7357, opened 2018-06-18
Update getting started intro and links

dedemorton:
Summary of changes:

- Removes old getting started content in favor of pointing to new content in the stack overview [getting started tutorial](https://www.elastic.co/guide/en/elastic-stack-overview/current/get-started-elastic-stack.html).
- Update links to point to stack getting started tutorial
- Moves shared intro content into a shared file to avoid duplicating identical content
- Update overview topic + diagram in Beats Platform Reference to include all the Beats
- Minor cleanup to use attributes instead of full names (ongoing task)

After merging, I need to:

- [ ]  Set up redirects for the deleted topic (installing-beats.html)
- [ ]  [can happen later] Update the Beats docs to use the shared attributes instead of a separate file--right now, we use attributes from both, but we have naming conflicts that prevent us from using the shared file only. 
- [ ] Update Beats plugin for LS docs to include additional footnotes to describe metadata


libbeat/docs/gettingstarted.asciidoc
```diff
@@ -1,492 +1,17 @@
...
-https://www.elastic.co/downloads/kibana[Kibana downloads page].
-
-==== Launch the Kibana web interface
-
-To launch the Kibana web interface, point your browser to port 5601. For example,
-http://127.0.0.1:5601[http://127.0.0.1:5601].
-
-You can learn more about Kibana in the
-http://www.elastic.co/guide/en/kibana/current/index.html[Kibana User Guide].
+* {packetbeat-ref}/packetbeat-getting-started.html[Packetbeat]
```
andrewkroh: Should we alphabetize the list of Beats? The order seems arbitrary.

libbeat/docs/overview.asciidoc
```diff
@@ -4,22 +4,34 @@
...
+https://www.elastic.co/products/elasticsearch[{es}]. Elastic provides {beats}
+for capturing:
+
+[horizontal]
+Network traffic:: https://www.elastic.co/products/beats/packetbeat[Packetbeat] 
+Log files:: https://www.elastic.co/products/beats/filebeat[Filebeat] 
+Metrics:: https://www.elastic.co/products/beats/metricbeat[Metricbeat] 
+Windows event logs:: https://www.elastic.co/products/beats/winlogbeat[Winlogbeat] 
+Audit data:: https://www.elastic.co/products/beats/auditbeat[Auditbeat] 
+Uptime:: https://www.elastic.co/products/beats/heartbeat[Heartbeat]
```
andrewkroh: I think "Availability" is a more accurate term to use for Heartbeat than "Uptime". Availability is whether or not the service is reachable. It's possible to have 100% uptime but be unreachable if the network is down. And Heartbeat will tell if you if it could connect to the service (hence telling you if it was available).

libbeat/docs/gettingstarted.asciidoc
dedemorton: Sort of. I ordered the list in a way that required the fewest changes to the diagram. :-) I don't have a strong preference either way because lists under 8 items are pretty easy for people to scan (we alphabetize mainly to make it easier for people to find stuff). Since you brought it up, though, I will alphabetize the list.

libbeat/docs/overview.asciidoc
dedemorton: Good point. I wonder if we need to change this on the product pages, too, because we describe Heartbeat as providing uptime monitoring.

dedemorton:
@andrewkroh I've updated the PR with the changes you suggested. Any more comments before we merge?

PR #7359, opened 2018-06-19
Add DHCP protocol handling for packetbeat

brianwaskiewicz:
First at contributing to packetbeat; this adds support for parsing DHCP messages.


brianwaskiewicz:
@andrewkroh Added a test and fixed the formatting. I think the make check is failing because the fields.go and fields.asciidoc need to be updated. Any pointers on the best way to do that?
________________________________

andrewkroh:
If you run `make update` that should regenerate those files and include your new DHCP content.

andrewkroh:
jenkins, please test this

andrewkroh:
jenkins, please test it

andrewkroh:
jenkins test this

packetbeat/protos/dhcp/dhcp.go
```diff
@@ -0,0 +1,198 @@
...
+	rows := payloadToRows(pkt.Payload)
+	dhcpFields["transaction_id"] = hex.EncodeToString(rows[1])
+	dhcpFields["client_ip"] = bytesToIPv4(rows[3])
+	dhcpFields["assigned_ip"] = bytesToIPv4(rows[4])
+	dhcpFields["server_ip"] = bytesToIPv4(rows[5])
+	dhcpFields["gateway_ip"] = bytesToIPv4(rows[6])
+	hwaddr := combineRows(rows, 7, 4)
+	dhcpFields["client_hwaddr"] = bytesToHardwareAddress(hwaddr, int(pkt.Payload[2]))
+	serverName := trimNullBytes(combineRows(rows, 11, 16))
+	dhcpFields["server_name"] = string(serverName)
```
adriansr: Is this "server_name" actually used in DHCP or is it a BOOTP leftover? I couldn't find much information about it and in my captures is always empty. I suggest you don't set the field at all in dhcpFields if empty.

packetbeat/protos/dhcp/dhcp.go
```diff
@@ -0,0 +1,198 @@
...
+			if option != 0 {
+				result[option] = body
+				body = make([]byte, 0)
+			}
+			option = int(options[idx])
+			if option == 0 || option == 255 {
+				idx = len(options)
+			} else {
+				idx++
+				remaining = int(options[idx])
```
adriansr: Missing bounds check here, it can panic on a malformed or crafted packet.

packetbeat/protos/dhcp/dhcp.go
```diff
@@ -0,0 +1,198 @@
...
+
+func extractOptions(payload []byte) []byte {
+	result := make([]byte, 0)
+	var idx = 240
+	for idx < len(payload) {
+		result = append(result, payload[idx])
+		idx++
+	}
+	return result
+}
```
adriansr: Same as just using payload[240:], don't need a function for this unless there is a point in copying the slice, which you can do with copy(...)

packetbeat/protos/dhcp/dhcp.go
```diff
@@ -0,0 +1,198 @@
...
+			option = int(options[idx])
+			if option == 0 || option == 255 {
+				idx = len(options)
+			} else {
+				idx++
+				remaining = int(options[idx])
+			}
+		} else {
+			body = append(body, options[idx])
+			remaining--
```
adriansr: This function can be improved to read the option at once instead of looping and appending every byte:
```
remaining = int(options[idx])
if idx + remaining + 1 > len(options) {
   error...
}
result[option] = options[idx+1:idx+1+remaining]
```

As before, I'm not sure if it's necessary to copy into a new slice or not.

adriansr: I'm worried that this protocol can create a lot of noise on a busy network.

What do you think about adding a config option to only report dhcp messages based on their type. Something like:

report_types: [ 'DHCPACK', 'DHCPNACK' ]

So that only those two types get reported.

Also I'm not 100% sure which events would be interesting. Do ACK and NACK contain all the information (IP assignments) and do all transactions end in ACK or NACK?

adriansr:
I'm worried that this protocol can create a lot of noise on a busy network.

What do you think about adding a config option to only report dhcp messages based on their type. Something like:

report_types: [ 'DHCPACK', 'DHCPNACK' ]

So that only those two types get reported.

Also I'm not 100% sure which events would be interesting. Do ACK and NACK contain all the information (IP assignments) and do all transactions end in ACK or NACK?

andrewkroh:
For transacted protocols like this, where there is a request/response and transaction ID to link the two packets, Packetbeat usually combines the whole transaction into a single event. 

I think we should keep it simple for DHCP and not combine the request and responses. There are scenarios in DHCP where a client sends only a single message, like a DHCPDECLINE, and does not expect a response. It would be simpler to not have to put any protocol state knowledge into the code.

Cons of Combining Request/Response
- The src and dest IPs used in the two packets of the transaction are not the same. The client uses src 0.0.0.0 and broadcasts to 255.255.255.255. And the server responds to client's assigned address.
- A request might receive multiple responses if there is more than one DHCP server. Handling this would be messy in the schema (arrays of objects).

Pros
- With transactions we can detect error scenarios where the client never received a response and indicate this is the event. However this can be handled in other ways such as with a flag in the request event or by use Watcher.

andrewkroh:
I think it would make sense to call this `dhcpv4` because we will likely want to add a `dhcpv6` in the future. And dhcpv6 would be a separate protocol operating on separate ports.

andrewkroh:
@brianwaskiewicz  I expanded on your pull request in https://github.com/elastic/beats/pull/7647. I tried a third-party library for the parsing.

andrewkroh:
DHCP support has been merged in #7647. 

@brianwaskiewicz, you're listed as a co-author in the commit. Thanks!

PR #7363, opened 2018-06-19
Cherry-pick #7253 to 6.3: Metricbeat: Collect accumulated docker network metrics

### LINKED ISSUE #7240: Fields docker.network.out.bytes and docker.network.in.bytes doesn't report what they say

jsoriano:
Cherry-pick of PR #7253 to 6.3 branch. Original message: 

Collect accumulated docker network metrics and mark previous metrics as deprecated.

Fixes #7240


jsoriano:
This is not a fix, but a set of related issues are already in 6.3, so we should probably merge this too for completeness.

ruflin:
jenkins, test this

jsoriano:

PR #7365, opened 2018-06-19
Filebeat: Elasticsearch module: Audit log 

radoondas:
This is initial PR for Audit logs in Elasticsearch. It successfully index basic events. Following PR's will improve grok patterns and add files for testing. Dashboards will come as bonus.


filebeat/module/elasticsearch/audit/_meta/fields.yml
```diff
@@ -0,0 +1,39 @@
+- name: audit
+  type: group
+  description: >
+  fields:
+    - name: timestamp
```
ruflin: I this timestamp still needed?

filebeat/module/elasticsearch/audit/_meta/fields.yml
```diff
@@ -0,0 +1,39 @@
...
+      description: Please add description
+      example: Please add example
```
ruflin: Can you update descriptions and potential examples if needed.

filebeat/module/elasticsearch/audit/_meta/fields.yml
```diff
@@ -0,0 +1,40 @@
+- name: audit
+  type: group
+  description: >
+  fields:
+    - name: node_name
```
ruflin: Is this the Elasticsearch node name? If yes, this field should end up under `elasticsearch.node.name`. Like this we will be able to correlate it.

filebeat/module/elasticsearch/audit/_meta/fields.yml
```diff
@@ -0,0 +1,40 @@
...
+      example: "access_granted"
+      type: keyword
+    - name: origin_type
+      description: "Where the request originated: rest (request originated from a REST API request), transport (request was received on the transport channel), local_node (the local node issued the request)"
+      example: "local_node"
+      type: keyword
+    - name: origin_address
+      description: "The IP address from which the request originated"
+      example: "192.168.1.42"
+      type: keyword
```
ruflin: Should this be of type ip?

filebeat/module/elasticsearch/audit/_meta/fields.yml
```diff
@@ -0,0 +1,40 @@
...
+    - name: event_type
+      description: "The type of event that occurred: anonymous_access_denied, authentication_failed, access_denied, access_granted, connection_granted, connection_denied, tampered_request, run_as_granted, run_as_denied"
+      example: "access_granted"
+      type: keyword
+    - name: origin_type
+      description: "Where the request originated: rest (request originated from a REST API request), transport (request was received on the transport channel), local_node (the local node issued the request)"
+      example: "local_node"
+      type: keyword
+    - name: origin_address
```
ruflin: So this is the ip of the machine on which for example an Elasticsearch client is running which makes requests?

filebeat/module/elasticsearch/audit/_meta/fields.yml
```diff
@@ -0,0 +1,40 @@
...
+      type: keyword
+    - name: action
+      description: "The name of the action that was executed"
+      example: "cluster:monitor/main"
+      type: keyword
+    - name: uri
+      description: "The REST endpoint URI"
+      example: /_xpack/security/_authenticate
+      type: keyword
+    - name: request
```
ruflin: `request` could be confusing as I would expect more data to be inside. Perhaps this is `request_type`?

radoondas: I followed the documentation names for all of the fields from Auditing.

filebeat/module/elasticsearch/audit/_meta/fields.yml
radoondas: Yeah, this can be an IP.

filebeat/module/elasticsearch/audit/_meta/fields.yml
radoondas: Yes

filebeat/module/elasticsearch/audit/_meta/fields.yml
radoondas: I am investigating here a bit more, because we have several options and more fields can show up. I need to run different configurations to follow up on behaviour and best approach. I will keep in mind that this field could be moved.

skearns64:
@ruflin - which versions was this merged into? Just master? Should we update the labels to reflect?

PR #7371, opened 2018-06-19
Set doc branch to 6.x

dedemorton:
Doc branch is incorrectly set to master, which breaks some links that point to removed topics.


dedemorton:
Fixed error found by @lcawl. Merging since this has been reviewed.

PR #7372, opened 2018-06-20
Add ability to configure autodiscover with JSON hints

vjsamuel:
Allow filebeat to configure an array of inputs using `co.elastic.logs/inputs: [stringified json]` and `co.elastic.metrics/modules` to do the same for metricbeat.

### DIFF
--- a/libbeat/autodiscover/builder/helper.go
+++ b/libbeat/autodiscover/builder/helper.go
@@ -1,11 +1,13 @@
 package builder
 
 import (
+	"encoding/json"
 	"fmt"
 	"strconv"
 	"strings"
 
 	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/logp"
 )
 
 // GetContainerID returns the id of a container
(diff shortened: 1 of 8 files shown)


libbeat/autodiscover/builder/helper.go
```diff
@@ -64,6 +65,16 @@ func getStringAsList(input string) []string {
 	return list
 }
 
+func GetHintAsConfig(hints common.MapStr, key, config string) []common.MapStr {
+	if str := GetHintString(hints, key, config); str != "" {
+		cfg := []common.MapStr{}
+		if err := json.Unmarshal([]byte(str), &cfg); err == nil {
```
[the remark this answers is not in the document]
vjsamuel: @ruflin modified.

libbeat/autodiscover/builder/helper.go
```diff
@@ -64,6 +66,19 @@ func getStringAsList(input string) []string {
 	return list
 }
 
+// GetHintAsConfig can read a hint in the form of a stringified JSON and return a common.MapStr
+func GetHintAsConfig(hints common.MapStr, key, config string) []common.MapStr {
```
exekias: this should probably be `GetHintAsConfigs`, as it returns a list

libbeat/autodiscover/builder/helper.go
```diff
@@ -64,6 +66,19 @@ func getStringAsList(input string) []string {
...
+	if str := GetHintString(hints, key, config); str != "" {
+		cfg := []common.MapStr{}
```
exekias: Would it make sense to accept both a list or a single config `{...}`? I guess in most cases the user wants just a config, with some special exceptions where you may want more

PR #7375, opened 2018-06-20
Fix json doc example for add_host_metadata processor

ruflin:
ip and mac were missing `"`.


ph:
This need to be in 6.3.0

PR #7381, opened 2018-06-20
Merge `state_` and common metricsets for node, pod & container

### LINKED ISSUE #6637: Consider combining "<type>" and "state_<type>" kubernetes metricsets into a single metricset/event.
Looking at https://github.com/elastic/beats/pull/6158 it seems like metricbeat is heading in that direction. It'd be convenient for certain visualizations (e.g. tables) to have all of the metrics in a single document from both the kubelet and kube-state-metrics for nodes, pods and containers. @exekias

exekias:
This PR is a rework of #6158, closes #6637

Merging `state_pod` & `pod` metricsets is tricky, mainly because `state_*` metricsets get events for the whole cluster (a single metricbeat instance for a kubernetes cluster), and `pod` metricset get's events for the current node. This is why we need both a `DaemonSet` and a `Deployment` when using Metricbeat.

Previous PR was flawed because it didn't have that into account, so the only way to get `pct` fields was running `state_*` metricsets in all nodes. That would send same events for a container from all nodes!

This change introduces `kube-state-metrics` fetching to `pod`, `container` and `node` metricsets, so they can use both kubelet `/stats/summary/` endpoint and `kube-sate-metrics` together, merging their metrics into a single event per entity.

Config looks like this:

```
- module: kubernetes
  metricsets:
    - pod
    - container
    - node
  hosts: ["192.168.99.100:10255"]  # kubelet
  period: 10s
  state_metrics:
    host: "kube-state-metrics:8080"
```

This change narrows down the supported `pct` fields to (by the moment):
```
kubernetes.container.cpu.usage.limit.pct
kubernetes.container.memory.usage.limit.pct
```
I find `pod` ones not needed, as they are a sum of `container` pct's. `node.pct` can be added, but I leave that out for a follow up PR, as this one is big already.

I think we should deprecate `state_node`, `state_pod` and `state_container` in favor of this (7.0).


metricbeat/module/kubernetes/util/mapstr.go
```diff
@@ -0,0 +1,98 @@
...
+		id := eventID(event, match)
+		events[id] = event
+	}
+
+	for _, event := range b {
+		// Skip events that don't match the filter
+		skip := false
+		for k, v := range filter {
+			if GetString(event, k) != v {
+				skip = true
```
jsoriano: `continue` with label directly here?

metricbeat/module/kubernetes/util/mapstr.go
```diff
@@ -0,0 +1,98 @@
...
+	if !ok {
+		return 0
+	}
+	return valStr
+}
+
+// MergeEvents from b events into a. The process will ensure that:
+//   - only events matching the given filter are processed.
+//   - fields in the delete list will be removed from the event
+//   - match fields will be used to match events from a & b, if all fields are equal, they will be merged
```
jsoriano: Could we separate this into three functions? one for merging, another one for filtering and another one to delete fields?

metricbeat/module/kubernetes/container/data.go
```diff
@@ -97,29 +95,33 @@ func eventMapping(content []byte, perfMetrics *util.PerfMetricsCache) ([]common.
...
-			}
+			containers = append(containers, containerEvent)
+		}
+	}
 
-			events = append(events, containerEvent)
+	events := util.MergeEvents(containers, stateMetrics,
+		map[string]string{
+			mb.ModuleDataKey + ".node.name": node.NodeName,
+		},
```
jsoriano: I guess the `/metrics` endpoint doesn't offer any filtering... There can be a huge list here with big clusters.

metricbeat/module/kubernetes/container/container.go
```diff
@@ -61,7 +76,15 @@ func (m *MetricSet) Fetch() ([]common.MapStr, error) {
 		return nil, err
 	}
 
-	events, err := eventMapping(body, util.PerfMetrics)
+	var stateMetrics []common.MapStr
+	if m.state != nil {
```
jsoriano: How can this be nil? or it is only for testing?

metricbeat/module/kubernetes/container/data.go
exekias: Yes, I have some concerns about this approach, I'm trying something else, let's keep this PR unmerged in the meanwhile.

Thanks for the review!

mumoshu: Hi, let me chime in because I'm being affected by this just now 😄 

----

Yes, so my best bet was to "silently" run `state_*` metricsets to populate the `perfMetrics` only for consumption by the `pod` metricset. In this mode, `state_*` metricsets shouldn't publish anything:

```yaml
- module: kubernetes
  metricsets:
    - pod
    - container
    - node
  hosts: ["192.168.99.100:10255"]  # kubelet
  period: 10s
  state_metrics:
    host: "kube-state-metrics:8080"
```

And `state_*` metricsets are enabled for publishing only when you explicitly listed them under the `metricsets`:

```yaml
- module: kubernetes
  metricsets:
    - state_container
    - state_node
  hosts: ["192.168.99.100:10255"]  # kubelet
  period: 10s
  state_metrics:
    host: "kube-state-metrics:8080"
```

Then you can deploy the former one in a k8s daemonset, whereas the latter is deployed in a k8s deployment.

But obviously my suggested implementation doesn't resolve #6637.

exekias:
Closed in favor of the new approach in #7470

metricbeat/module/kubernetes/container/data.go
mumoshu: #7470 seems awesome! But not sure what you mean by the perfMetrics part yet.

Would you be enabling an enhanced version of `NodeMetadataEnricher` on the container metricset [here](https://github.com/elastic/beats/pull/7470/files#diff-96e4742da1cc1cb927b8fdd795b09e39R70), in addition to the ContainerMetadataEnricher, in order to fill `perfMetrics.NodeCoresAllocatable` that is used for calculating `container.cpu.usage.node.pct`?

Thanks anyway for your efforts as always! I'll keep eyes on #7470 👍

exekias: yes, that's the plan :smiley_cat:

PR #7388, opened 2018-06-21
Refactor Beat packaging and cross-building

andrewkroh:
This refactors Beat packaging to use a declarative YAML based packaging
specification. This makes it easier to customize the contents of packages (e.g.
adding additional X-Pack content or simply tailoring what's included for single
Beat).

The specification itself is pretty simple. It consists of package metadata and a
list of files to include. The values can be templated which allows for a single
package spec to be reused across Beats. Here's an example spec for an OSS
Windows zip package.

```
spec:
  name:         '{{.BeatName}}-oss'
  service_name: '{{.BeatServiceName}}'
  os:           '{{.GOOS}}'
  arch:         '{{.PackageArch}}'
  vendor:       '{{.BeatVendor}}'
  version:      '{{ beat_version }}'
  license:      ASL 2.0
  url:          '{{.BeatURL}}'
  description: '{{.BeatDescription}}'
  files:
    '{{.BeatName}}{{.BinaryExt}}':
      source: build/golang-crossbuild/{{.BeatName}}-{{.GOOS}}-{{.Platform.Arch}}{{.BinaryExt}}
      mode:   0755
    fields.yml:
      source: fields.yml
      mode:   0644
    LICENSE.txt:
      source: '{{ repo.RootDir }}/licenses/APACHE-LICENSE-2.0.txt'
      mode: 0644
    NOTICE.txt:
      source: '{{ repo.RootDir }}/NOTICE.txt'
      mode:   0644
    README.md:
      template: '{{ elastic_beats_dir }}/dev-tools/mage/templates/common/README.md.tmpl'
      mode:   0644
    .build_hash.txt:
      content: >
        {{ commit }}
      mode:   0644
    '{{.BeatName}}.reference.yml':
      source: '{{.BeatName}}.reference.yml'
      mode:   0644
    '{{.BeatName}}.yml':
      source: '{{.BeatName}}.yml'
      mode: 0600
      config: true
    kibana:
      source: _meta/kibana.generated
      mode:   0644
    install-service-{{.BeatName}}.ps1:
      template: '{{ elastic_beats_dir }}/dev-tools/mage/templates/windows/install-service.ps1.tmpl'
      mode: 0755
    uninstall-service-{{.BeatName}}.ps1:
      template: '{{ elastic_beats_dir }}/dev-tools/mage/templates/windows/uninstall-service.ps1.tmpl'
      mode: 0755
```

Each Beat has two build targets for packaging.

- `make snapshot`
- `make release`

The set of target platforms can be influenced by the `PLATFORMS` environment
variable which accepts an platform selection expression. For example to add
ARMv7 (for your Raspberry Pi) to the default set of platforms you would use
`PLATFORMS='+linux/armv7' make snapshot`. Or to only build for Windows set
`PLATFORMS='windows'`. Full details can be found in the godocs for
`NewPlatformList`.

For the release manager there are two new top-level targets that take care of
ensuring that the proper Go version is used. The naming here aligns with what
several of the other projects are using for their release manager targets.

- `make release-manager-snapshot`
- `make release-manager-release`

Build Process Details
------

Below is the command line output of building and packaging Packetbeat for
linux/armv7 only. I'll describe each step in the build process.

```
$ PLATFORMS='+all linux/armv7' make snapshot
Installing mage from vendor
>> golangCrossBuild: Building for linux/armv7
>> buildGoDaemon: Building for linux/armv7
>> Building using: cmd='build/mage-linux-amd64 golangCrossBuild', env=[CC=arm-linux-gnueabihf-gcc, CXX=arm-linux-gnueabihf-g++, GOARCH=arm, GOARM=7, GOOS=linux, PLATFORM_ID=linux-armv7]
>> Building using: cmd='build/mage-linux-amd64 buildGoDaemon', env=[CC=arm-linux-gnueabihf-gcc, CXX=arm-linux-gnueabihf-g++, GOARCH=arm, GOARM=7, GOOS=linux, PLATFORM_ID=linux-armv7]
grammar.y: warning: 38 shift/reduce conflicts [-Wconflicts-sr]
>> package: Building packetbeat type=tar.gz for platform=linux/armv7
>> package: Building packetbeat-oss type=deb for platform=linux/armv7
>> package: Building packetbeat type=rpm for platform=linux/armv7
>> package: Building packetbeat type=deb for platform=linux/armv7
>> package: Building packetbeat-oss type=tar.gz for platform=linux/armv7
>> package: Building packetbeat-oss type=rpm for platform=linux/armv7
>> Testing package contents
package ran for 1m6.597416206s

$ tree build/distributions/
build/distributions/
├── packetbeat-oss-7.0.0-alpha1-SNAPSHOT-armhf.deb
├── packetbeat-oss-7.0.0-alpha1-SNAPSHOT-armhf.deb.sha512
├── packetbeat-oss-7.0.0-alpha1-SNAPSHOT-armhfp.rpm
├── packetbeat-oss-7.0.0-alpha1-SNAPSHOT-armhfp.rpm.sha512
├── packetbeat-oss-7.0.0-alpha1-SNAPSHOT-linux-armv7.tar.gz
└── packetbeat-oss-7.0.0-alpha1-SNAPSHOT-linux-armv7.tar.gz.sha512
```

1. Install [mage](https://magefile.org/) from the vendor directory to `$GOPATH/bin`.

   Mage is used to provide a simple gmake like wrapper around build logic that's
   written in Go. Additionally it makes is possible to build and package without
   gmake by invoking mage directly [think Windows users]. For example:

```
   $ mage -l
    Targets:
      build                 builds the Beat binary.
      buildGoDaemon         builds the go-daemon binary (use crossBuildGoDaemon).
      clean                 cleans all generated files and build artifacts.
      crossBuild            cross-builds the beat for all target platforms.
      crossBuildGoDaemon    cross-builds the go-daemon binary using Docker.
      golangCrossBuild      build the Beat binary inside of the golang-builder.
      package               packages the Beat for distribution.
      testPackages          tests the generated packages (i.e.
      update                updates the generated files (aka make update).
```

2. Cross-build Packetbeat for linux/armv7. Cross-building requires Docker and
   uses images hosted at `docker.elastic.co/beats-dev/golang-crossbuild`. The
   repo for these images is https://github.com/elastic/golang-crossbuild. These
   images give us wider platform support for cross-building than we had.

   `mage golangCrossBuild` is invoked inside of the container. This handles
   cross-compiling libpcap and then invoking `go build` with the proper args.

3. Cross-build go-daemon for linux/armv7. This is actually done in parallel with
   the other cross-builds. GOMAXPROCS determines the number of concurrent jobs.

   `mage buildGoDaemon` is invoked inside of the container.

4. After all cross-builds complete, packaging begins. The package types are
   decided based on the package specs that are registered for each OS.

   Zip and tar.gz files are built natively with Go. RPM and deb packages are
   first generated as tar.gz where we have full control over the target file
   names, ownership, and modes regardless of the underlying filesystem [think
   Windows]. Then FPM is invoked inside of Docker to translate the tar.gz file
   into a proper RPM or deb.

5. SHA512 side-car files are generated for each package. Go is used for this so
   no special command line tools are needed.

6. The generated packages are inspected with `dev-tools/package_test.go`. This
   looks at the contents of the packages to ensure the files have the expected
   ownership and modes (e.g. the config file should have 0600).

### CI Testing

- [X] beats-tester on release - https://beats-ci.elastic.co/job/elastic+beats-tester+master/352/
- [x] beats-tester on snapshot - https://beats-ci.elastic.co/job/elastic+beats-tester+master/353/

### Follow-up Tasks

- [ ] Integrate in the macOS packaging and preference pane
- [ ] Sync to 6.4 branch
- [ ] Set `PLATFORMS` for beats-ci packaging job to enable snapshot packages for additional platforms (like `+linux/armv7 +linux/ppc64le +linux/s390x`).
- [ ] Update commands used by release manager on 6.4 and master
- [ ] Retire dev-tools/deploy
- [ ] Test on Windows
- [ ] Work with apm-server to update (I have already done some testing) - [magefile.go for apm-server](https://gist.github.com/andrewkroh/15d81ec0d04a4678fbf031af458cddf1)

[snapshot-urls.txt](https://github.com/elastic/beats/files/2122012/snapshot-urls.txt)


ruflin:
@elastic/apm-server This PR might have an affect on apm-server packaging.

Vagrantfile
```diff
@@ -200,6 +200,16 @@ Vagrant.configure(2) do |config|
     c.vm.synced_folder ".", "/vagrant", type: "virtualbox"
   end
 
+  # Windows Server 2016
```
ruflin: I assume this is not directly related to this PR?

auditbeat/magefile.go
```diff
@@ -0,0 +1,212 @@
...
+
+	"github.com/magefile/mage/mg"
+	"github.com/magefile/mage/sh"
+	"github.com/pkg/errors"
+
+	"github.com/elastic/beats/dev-tools/mage"
+)
+
+func init() {
+	mage.BeatDescription = "Audit the activities of users and processes on your system."
```
ruflin: Would be nice to have the description in the future as part of the yaml file instead of the Golang file. Would make it easier to reuse if needed (not now).

auditbeat/magefile.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+// TestPackages tests the generated packages (i.e. file modes, owners, groups).
+func TestPackages() error {
+	return mage.TestPackages()
+}
+
+// Update updates the generated files (aka make update).
+func Update() error {
+	return sh.Run("make", "update")
```
ruflin: Looking forward to have it all in Go so this would also work on Windows.

dev-tools/mage/build.go
```diff
@@ -0,0 +1,144 @@
...
+			return err
+		}
+	}
+
+	// Environment
+	env := params.Env
+	if env == nil {
+		env = map[string]string{}
+	}
+	cgoEnabled := "0"
```
ruflin: If I remember correctly we use now CGO by default. Should we also set it here to 1 by default? At the same time I like that community beats can easily built without it.

dev-tools/mage/clean.go
```diff
@@ -0,0 +1,49 @@
...
+	"build",
+	"docker-compose.yml.lock",
+	"{{.BeatName}}",
+	"{{.BeatName}}.exe",
+	"{{.BeatName}}.test",
+	"{{.BeatName}}.test.exe",
+	"fields.yml",
+	"_meta/fields.generated.yml",
+	"_meta/kibana.generated",
+	"_meta/kibana/5/index-pattern/{{.BeatName}}.json",
```
ruflin: This should not be needed anymore as the index-patterns end up in the `kibana.generated` directory.

dev-tools/mage/crossbuild.go
```diff
@@ -0,0 +1,240 @@
...
+		args = append(args,
+			"--env", "EXEC_UID="+strconv.Itoa(os.Getuid()),
+			"--env", "EXEC_GID="+strconv.Itoa(os.Getgid()),
+		)
+	}
+	args = append(args,
+		"--rm",
+		"--env", "MAGEFILE_VERBOSE="+verbose,
+		"--env", "MAGEFILE_TIMEOUT="+EnvOr("MAGEFILE_TIMEOUT", ""),
+		"--env", "CGO_ENABLED=1",
```
ruflin: Should this depend on the environment variable?

dev-tools/mage/files/packages.yml
```diff
@@ -0,0 +1,222 @@
...
+- &common
+  name:         '{{.BeatName}}'
+  service_name: '{{.BeatServiceName}}'
+  os:           '{{.GOOS}}'
+  arch:         '{{.PackageArch}}'
+  vendor:       '{{.BeatVendor}}'
+  version:      '{{ beat_version }}'
+  license:      '{{.BeatLicense}}'
+  url:          '{{.BeatURL}}'
+  description:  '{{.BeatDescription}}'
```
ruflin: For community Beats they will have to move the description to the Golang file?

dev-tools/mage/pkg.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+func (b packageBuilder) Build() error {
+	fmt.Printf(">> package: Building %v type=%v for platform=%v\n", b.Spec.Name, b.Type, b.Platform.Name)
+	log.Printf("Package spec: %+v", b.Spec)
+	return errors.Wrapf(b.Type.Build(b.Spec), "failed building %v type=%v for platform=%v",
+		b.Spec.Name, b.Type, b.Platform.Name)
+}
+
+// TestPackages executes the package tests on the produced binaries. These tests
```
ruflin: Really nice. We could also extend these with things like checking if the dashboard files are all in the correct directory etc. ¨

dev-tools/mage/settings.go
```diff
@@ -0,0 +1,573 @@
...
+	"sync"
+	"time"
+
+	"github.com/magefile/mage/sh"
+	"github.com/pkg/errors"
+	"golang.org/x/tools/go/vcs"
+)
+
+const (
+	fpmVersion = "1.10.0"
```
ruflin: Should this be in sync with the Golang version?

dev-tools/package_test.go
```diff
@@ -47,7 +47,7 @@ const (
 )
 
 var (
-	configFilePattern   = regexp.MustCompile(`.*beat\.yml`)
+	configFilePattern   = regexp.MustCompile(`.*beat\.yml|apm-server\.yml`)
```
ruflin: I see we are already prepared for apm 👍

Vagrantfile
andrewkroh: It's only related in that I was testing these build changes on win2016 and we didn't have a machine for it yet.

auditbeat/magefile.go
andrewkroh: Yes. And a few other things that would be useful to externalize in a single yaml file (because it messy to guess and locate the data in the various repos (elastic/beats, apm-server, and community beats):

- beat version
- beat doc branch
- go version

This can be done a future cleanup. I'd like to move more stuff out of the makefile too.

dev-tools/mage/clean.go
andrewkroh: I had removed those, but when working across branches it's really nice to be able to fully clean things left behind. I suggest moving all generated files that are not committed to `build/` over time.

dev-tools/mage/crossbuild.go
andrewkroh: It doesn't matter that it's set (or not set) here because the decision to use cgo for cross-builds is made by the `mage golangCrossBuild`. I'll remove this to avoid confusion.

dev-tools/mage/files/packages.yml
andrewkroh: Yes. Or `export BEAT_DESCRIPTION` from their Makefile. But I'd prefer the `magefile.go` to be the source of truth for this information (or an external YAML file as you suggested) so that `make` isn't a requirement for building.

dev-tools/mage/pkg.go
andrewkroh: Good idea. I suggest enhancing the test outside of this PR. The package_test.go is executed by the release-manager today.

dev-tools/mage/settings.go
andrewkroh: No, 1.10.0 is the latest [release of FPM](https://github.com/jordansissel/fpm/releases).

dev-tools/package_test.go
andrewkroh: Yes, I have tested this with APM. I updated the description to include a link to the `magefile.go` I used which customizes their packaging to not include the reference YAML and use a customized readme.md.

andrewkroh:
> One more general question from my side is on how exactly an update of the Golang version will look like in the future? Assuming we go to 1.11, what are all the things we have to change / do?

It's generally the same as it is today, except the docker images are stored in a different repo. Instructions for updating can be found at https://github.com/elastic/golang-crossbuild#releasing-images-for-a-new-go-version.

andrewkroh:
> readdirent: input/output error

That must be some kind of an issue with the shared filesystem between macOS and the VM. I haven't seen that error yet on my machine. I have tested this on Docker for Mac, Boot2Docker (on mac), and Docker CE on Linux.

andrewkroh:
> One strange thing I realise after trying to package is that I had a diff in one of the vendored files:

This isn't new. When packaging Packetbeat we patch `github.com/tsg/go-packet` so that we can adjust the `LDFLAGS` and `CFLAGS` that are hard-coded in the source. As a result, when packaging completes you have a modified source file.

I could "fix" this by adding a `defer sh.Run(git checkout vendor/github.com/tsg/gopacket/pcap/pcap.go)` at the end. I'll try this.

Thanks for testing on your machine!

andrewkroh:
@ruflin I pushed an update based on your review. af3cb054f8d7c43962f48d1f57f1fb6c2bf46477

- Fix naming of `make package` which is now either `make release` or `make snapshot`.
- Remove unneeded CGO_ENABLED env var from the cross build container
- Undo go-packet patch with 'git checkout' after crossBuild completes

dev-tools/mage/build.go
andrewkroh: The passed in `BuildArgs` created by `DefaultBuildArgs()` and `DefaultGolangCrossBuildArgs` will set the cgo flag appropriately. (it's a bit complex)

- `mage build` (equivalent of `make` or `go build`) - This uses `DefaultBuildArgs()` as its starting point (Beats can further override the args). The logic for enabling cgo is exactly the same as it is for Go because it reuses the stdlib logic. Basically the logic is `if env[GOOS] == runtime.GOOS` then CGO_ENABLED=1. This ensures that cross-builds (`GOOS=netbsd mage build`) have CGO turned off since cross-compilers are not typically installed.

- `mage crossBuild` - The defaults come from `DefaultGolangCrossBuildArgs()`. This is the target that builds the binaries inside of docker using [golang-crossbuild](https://github.com/elastic/golang-crossbuild). CGO is enabled here by default (generally). More precisely if the target platform has CGO support then it will be enabled (that list came from `go tool dist list -json`).

dev-tools/mage/clean.go
```diff
@@ -0,0 +1,49 @@
...
+	"_meta/kibana/6/index-pattern/{{.BeatName}}.json",
+}
+
+// Clean clean generated build artifacts.
+func Clean() error {
```
urso: There are a many defaults/magic paths here. How about making it somewhat configurable by passing the paths as arguments?

```
func Clean(pathsList ...[]string) error {
  if len(pathsList) == 0 {
    pathsList = [][]string{CleanPaths}
  }
  for _, lst := range pathsList {
    for _, f := range lst {
      ..
    }
  }
}
```

So the per beat magefile can be written like:

```
func Clean() error {
  return make.Clean(mage.DefaultCleanPaths, []string{
    "...",
    "...",
  })
}
```

This is still valid.

```
func Clean() error {
  return mage.Clean()
}
```

CHANGELOG.asciidoc
```diff
@@ -19,6 +19,7 @@ https://github.com/elastic/beats/compare/v6.2.3...master[Check the HEAD diff]
 - Allow override of dynamic template `match_mapping_type` for fields with object_type. {pull}6691[6691]
 - Set default kafka version to 1.0.0 in kafka output. Older versions are still supported by configuring the `version` setting. {pull}7025[7025]
 - Add `host.name` field to all events, to avoid mapping conflicts. This could be breaking Logstash configs if you rely on the `host` field being a string. {pull}7051[7051]
+- Packaging has been refactored and updates are required. See the PR for migration details. {pull}7388[7388]
```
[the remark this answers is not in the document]
andrewkroh: Yeah, I meant to put it there.

dev-tools/mage/clean.go
andrewkroh: Changed. 👍

auditbeat/magefile.go
```diff
@@ -0,0 +1,212 @@
...
+		return errors.Wrap(err, "failed to find config templates")
+	}
+
+	var shortIn []string
+	shortIn = append(shortIn, "_meta/common.p1.yml")
+	shortIn = append(shortIn, configFiles...)
+	shortIn = append(shortIn, "_meta/common.p2.yml")
+	shortIn = append(shortIn, "../libbeat/_meta/config.yml")
+	if !mage.IsUpToDate(shortConfigTemplate, shortIn...) {
+		fmt.Println(">> Building", shortConfigTemplate)
```
kvch: Why didn't you extracted this 4 lines into a separate function with parameters `template`, `snippets`, `permission` which passes the correct permissions to `MustFileConcat`? I think this would look cleaner:

```golang
if !mage.IsUpToDate(shortConfigTemplate, shortIn...) {
    updateConfigTemplate(shortConfigTemplate, shortIn, 0600)
}

if !mage.IsUpToDate(referenceConfigTemplate, referenceIn...) {
   updateConfigTemplate(referenceConfigTemplate, referenceIn, 0644)
}

// ...
```
The new extracted function could be:
```golang
func updateConfigTemplates(template string, snippets []string, permission int) {
    fmt.Println(">> Building", template)
    mage.MustFileConcat(template, 0644, snippets...)
    mage.MustFindReplace(template, regexp.MustCompile("beatname"), "{{.BeatName}}")
    mage.MustFindReplace(template, regexp.MustCompile("beat-index-prefix"), "{{.BeatIndexPrefix}}")
}
```

So if the way config files are updated is changed, we only need to modify it in one function.

auditbeat/magefile.go
```diff
@@ -0,0 +1,212 @@
...
+func Update() error {
+	return sh.Run("make", "update")
+}
+
+// -----------------------------------------------------------------------------
+// Customizations specific to Auditbeat.
+// - Config files are Go templates.
+
+const (
+	configTemplateGlob      = "module/*/_meta/config*.yml.tmpl"
```
kvch: Does it work on Windows? Shouldn't this be
```
configTemplateGlob = filepath.Join("module", "*", "_meta", "config*.yml.tmpl")
```
?

auditbeat/scripts/generate_config.go
```diff
@@ -52,26 +52,29 @@ func findConfigFiles(globs []string) ([]string, error) {
...
+// generate architecture specific audit rules.
+func archBits(goarch string) int {
+	switch goarch {
+	case "386", "arm":
+		return 32
+	default:
+		return 64
+	}
+}
+
```
ph: I've seen this method twice,  it should be extracted.

andrewkroh:
Right now the beat version used inside of binaries is coming from:

https://github.com/elastic/apm-server/blob/968f8621fc11c1cfd5ec8eeef2d3816869585973/magefile.go#L36

But I think there may be more places that have the Beat version (like a version.asciidoc IIRC). It would be nice to get down to a single file in each repo that is the source of truth for things like beat version, go version, doc branch, etc.

PR #7390, opened 2018-06-21
Add jvm9 logging format to GC fileset of ES module

immon:


filebeat/module/elasticsearch/gc/_meta/fields.yml
```diff
@@ -14,3 +14,11 @@
       type: float
       description: >
         Time took to stop threads seconds.
+    - name: pid
```
ruflin: I wonder if we should change this to `process.pid` in a follow up PR? https://github.com/elastic/ecs#process

immon: sounds good. `tags` field seems like a good candidate too.

immon: initially I thought to rename `elasticsearch.gc.tags` to `tags` but now I don't think it's a good idea. `Tags` are more generic while `elasticsearch.gc.tags` are GC specific fields with values of `safepoint`, `gc`, `heap`, `age`. I will rename PID field only.

ruflin: ++ on what you stated about the `tags`. We had very similar discussions in other places like `docker` and often these tags are service specific so don't belong in the global ones.

PR #7391, opened 2018-06-21
Add Apache headers to all .go source files

andrewkroh:
Headers were added by https://github.com/elastic/go-licenser.

Make Changes

Added Targets

- `check-headers` - verifies headers exist on .go files.
- `add-headers` - adds headers to all .go files.

Changed Targets

- `check` - added dependency on `check-headers`.
- `fmt` - added dependency on `add-headers`.

Variables

- `CHECK_HEADERS_DISABLED` - if set then headers are neither checked nor added.

REVIEWERS: I separated this into two commits to make reviewing simpler. You can squash them together.

- https://github.com/elastic/beats/pull/7391/commits/c5cd1f5edc9a55c72a5a8d1d664251ef4ade13a7 - This has the tooling changes.
- https://github.com/elastic/beats/pull/7391/commits/156763ffe5c012165ebf3fb02d8235f777b0edf9  - This has the updates to the .go files to add headers.


andrewkroh:
I opened some issues with go-licenser to allow for better integration. Specififically when we add Go x-pack content we will need a way to check those licenses.

- https://github.com/elastic/go-licenser/issues/10
- https://github.com/elastic/go-licenser/issues/1

ruflin:
@andrewkroh Is there an automated way to add the headers to all files? How did you add them?

andrewkroh:
The same tool that checks the headers can be used to add them.

`go-licenser` - This adds headers to all .go files recursively.
`go-licenser -d` - Checks for the existence of headers.

andrewkroh:
As discussed with @ruflin  and @urso , `make fmt` will add headers to .go files.

ruflin:
PR looks good to me and we should merge it rather soonish. I think we should have an entry at least in the developer changelog for this.

ph:
Changes LGTM, @andrewkroh won't the commit race for 2018. :)

I think we want to apply the same logic to other source code, like our python tests? (in another PR of course)

Issue #7093, opened 2018-05-14
[Discuss]  Refactor How libbeat features are activated for individual beats.


ph:
When I was working on a function agent, I found the need to have an ephemeral beat, I mean a beat that doesn't write anything to disk, or executes any disk related actions, like creating new directories.

A custom beat developer should be able to tell libbeat what features to activate; we could provide a default sets of
capabilities.

 This would mean we could break the core into these capabilities:

- Keystore
- Seccomp
- Reporting
- Persisted (or ephemeral data, write log to disk, create data folder)
- Install Dashboards
- Install Ingest Pipeline
- Install ML Job
- LogSystemInfo

By default beats author have access to all the capabilities:

##
Advantages
 - Tailored experience, use what you need
 - Reduce the complexity of cmd/instance/beat.go
 - Easier testing
 - Force isolation.
 -  Possibility of better logger
 - Faster startup time
## Possible interface
```golang
func init() {
  c := capabilities.List() # return the currently active capabilities
  if err := capabilities.Clear(); err != nil {
    // capabilities already activated
  }

  // Since capabilities might have have dependency, we let the capabilities package handle that.
  // first version could be hardcoded execution
  if err := capabilities.Enable(capabilities.Seccomp); err != nil {
    //.. panic could not enable capabilities
    // require X capabilities or not on linux.
  }
  capabilities.Enable(capabilities.Ephemeral)
}
```

This might be complicated if we merge multiple beats into one, we could solve this by adding the beat canonical name to the `Enable` method and do some validation at startup to make sure the capabilities are compatible.

Each capability would be responsable to deal with validating his configuration.

ph:
We could also add a hook for `config transformer`, which takes an existing configuration and try to set other config options, the code dealing with the `cloud_id` could be the first one.

ph:
I believe, the downside of this interface, is the capabilities would still be part of the binary, we might want to change that to make sure they are excluded at compile time.

As @ruflin pointed out, we might want to have something similar to the outputs or processors.

kvch:
What about adding a configuration file to the generators which specifies which parts of libbeat should be copied to the vendor folder of a new Beat? Also, we could still keep the command line flags and prompting for answers when we need them.

```
capabilites:
 - input/file
 - processors:
    - drop_event
 - output/kafka
```

This way everyone could generate a small Beat which satisfies their needs. However, I rather see my idea as a dream. I am not sure if it's feasible.

ph:
[...]

PR #7392, opened 2018-06-21
Refactor: Introduce a Global Registry

ph:
Motivation:

In the current libbeat implementation and also inside the actual beat we are defining registries for each
types of feature that we want to expose. This add duplication to the project, the global registry
is a way to keep the flexibility of multiple features and reduce the duplication code only to take
care of the type satefy.

Also all features now use an init function to make the plugin register with their specific registry.
This PR is a step forward to remove that pattern and use a global variable in the package to identify
the feature. This change will allow a beat author to build a beat with only a specific set of feature.

Example: Build with only ES out and not Logstash and kafka, this could reduce the size of some beats.

This PR is written in a backward compatible way, to make the init and the new feature works both at
the same time.

Instead of using an init function you will the following to expose the feature.
```golang

// Feature exposes a spooling to disk queue.
var Feature = queue.Feature("spool", create, feature.Beta)
```

Each new type of feature require to implement two things for type satefy:

- A factory method to assert the correct type at runtime.
- A sugar method like the `queue.Feature`, for type satefy at compile time.

Example:

```golang
// Feature creates a new type of queue.
func Feature(name string, factory Factory, stability feature.Stability) *feature.Feature {
	return feature.New(Namespace, name, factory, stability)
}

// FindFactory retrieves a queue types constructor. Returns nil if queue type is unknown
func FindFactory(name string) Factory {
	f, err := feature.Registry.Find(Namespace, name)
	if err != nil {
		return nil
	}
	factory, ok := f.Factory().(Factory)
	if !ok {
		return nil
	}

	return factory
}

```

How it will look like for building beats with a minimal set of plugins:

```golang
b := MustBundle(
  MustBundle(docker.Feature),
  MustBundle(dissect.Feature),
  MustBundle(elasticsearch.Feature, logstash.Feature),
)

feature.RegisterBundle(b)

```

Caveats:

we still expose the methods and the registry as global, but this is a step to to isolate a registry 
per beat.

TODO:
- [ ] Move all plugins to the feature concept.
- [ ] Move the external plugin to the feature concept.
- [ ] Update the auto generation tools.
- [x] Changelog
- [x] Developper changelog
- [ ] Meta issue for migration
  - [ ] codec
  - [ ] output
  - [ ] processor
  - [ ] autodiscover
- [ ] How to discover features, we could generate a feature list automatically.

Ref: #7093


ph:
As a side effect this gives us the opportunity to make sure a beats doesn't load anything experimental or beta.

```golang
b := MustBundle(
  MustBundle(docker.Feature),
  MustBundle(dissect.Feature),
  MustBundle(elasticsearch.Feature, logstash.Feature),
)

feature.RegisterBundle(b.Filter(feature.Stable))
feature.RegisterBundle(b.Filter(feature.Stable, Feature.Beta))

```

ph:
In my previous comment, I meant _execute_ not _load_, the features will still get compiled in.

libbeat/feature/bundle.go
```diff
@@ -0,0 +1,55 @@
...
+		case Featurable:
+			merged = append(merged, v)
+		case *Bundle:
+			merged = append(merged, v.Features...)
+		default:
+			return nil, fmt.Errorf("unknown type, expecting 'Featurable' or 'Bundle' and received '%T'", v)
+		}
+	}
+	return NewBundle(merged), nil
+}
```
ph: I will change this, I will remove the switch case.

ph: The following might be better.
```diff
diff --git a/libbeat/feature/bundle.go b/libbeat/feature/bundle.go
index 174c2f973..9886b68b5 100644
--- a/libbeat/feature/bundle.go
+++ b/libbeat/feature/bundle.go
@@ -1,22 +1,26 @@
 package feature
 
-import "fmt"
+// Groupable allows features and Bundle to be merged together.
+type Groupable interface {
+	// Features returns a slice of features.
+	Features() []Featurable
+}
 
 // Bundle defines a list of features available in the current beat.
 type Bundle struct {
-	Features []Featurable
+	features []Featurable
 }
 
 // NewBundle creates a new Bundle of feature to be registered.
 func NewBundle(features []Featurable) *Bundle {
-	return &Bundle{Features: features}
+	return &Bundle{features: features}
 }
 
 // Filter creates a new bundle with only the feature matching the requested stability.
 func (b *Bundle) Filter(stabilities ...Stability) *Bundle {
 	var filtered []Featurable
 
-	for _, feature := range b.Features {
+	for _, feature := range b.features {
 		for _, stability := range stabilities {
 			if feature.Stability() == stability {
 				filtered = append(filtered, feature)
@@ -27,29 +31,16 @@ func (b *Bundle) Filter(stabilities ...Stability) *Bundle {
 	return NewBundle(filtered)
 }
 
-// MustBundle takes existing bundle or features and create a new Bundle with all the merged Features,
-// will panic on errors.
-func MustBundle(features ...interface{}) *Bundle {
-	b, err := BundleFeature(features...)
-	if err != nil {
-		panic(err)
-	}
-	return b
+// Features returns a slice of features.
+func (b *Bundle) Features() []Featurable {
+	return b.features
 }
 
-// BundleFeature takes existing bundle or features and create a new Bundle with all the merged
-// Features,
-func BundleFeature(features ...interface{}) (*Bundle, error) {
+// MustBundle takes existing bundle or features and create a new Bundle with all the merged Features.
+func MustBundle(features ...Groupable) *Bundle {
 	var merged []Featurable
 	for _, feature := range features {
-		switch v := feature.(type) {
-		case Featurable:
-			merged = append(merged, v)
-		case *Bundle:
-			merged = append(merged, v.Features...)
-		default:
-			return nil, fmt.Errorf("unknown type, expecting 'Featurable' or 'Bundle' and received '%T'", v)
-		}
+		merged = append(merged, feature.Features()...)
 	}
-	return NewBundle(merged), nil
+	return NewBundle(merged)
 }
diff --git a/libbeat/feature/bundle_test.go b/libbeat/feature/bundle_test.go
index df293cc4b..e675a7f09 100644
--- a/libbeat/feature/bundle_test.go
+++ b/libbeat/feature/bundle_test.go
@@ -16,25 +16,25 @@ func TestBundle(t *testing.T) {
 
 	t.Run("Creates a new Bundle", func(t *testing.T) {
 		b := NewBundle(features)
-		assert.Equal(t, 3, len(b.Features))
+		assert.Equal(t, 3, len(b.Features()))
 	})
 
 	t.Run("Filters feature based on stability", func(t *testing.T) {
 		b := NewBundle(features)
 		new := b.Filter(Experimental)
-		assert.Equal(t, 1, len(new.Features))
+		assert.Equal(t, 1, len(new.Features()))
 	})
 
 	t.Run("Filters feature based on multiple different stability", func(t *testing.T) {
 		b := NewBundle(features)
 		new := b.Filter(Experimental, Stable)
-		assert.Equal(t, 2, len(new.Features))
+		assert.Equal(t, 2, len(new.Features()))
 	})
 
 	t.Run("Creates a new Bundle from specified feature", func(t *testing.T) {
 		f1 := New("libbeat.outputs", "elasticsearch", factory, Stable)
 		b := MustBundle(f1)
-		assert.Equal(t, 1, len(b.Features))
+		assert.Equal(t, 1, len(b.Features()))
 	})
 
 	t.Run("Creates a new Bundle with grouped features", func(t *testing.T) {
@@ -49,6 +49,6 @@ func TestBundle(t *testing.T) {
 			MustBundle(f3, f4),
 		)
 
-		assert.Equal(t, 4, len(b.Features))
+		assert.Equal(t, 4, len(b.Features()))
 	})
 }
diff --git a/libbeat/feature/feature.go b/libbeat/feature/feature.go
index 0931bbe53..1d2621982 100644
--- a/libbeat/feature/feature.go
+++ b/libbeat/feature/feature.go
@@ -59,6 +59,11 @@ func (f *Feature) Stability() Stability {
 	return f.stability
 }
 
+// Features allows to Bundle a Feature with a bundle of features.
+func (f *Feature) Features() []Featurable {
+	return []Featurable{f}
+}
+
 // Equal return true if both object are equals.
 func (f *Feature) Equal(other Featurable) bool {
 	// There is no safe way to compare function in go,
```

ruflin:
@ph As we currently generate the imports we guarantee that we don't forget about including an output for example or a new Metricbeat module. Would be nice if we could find a method to still verify this with this new method.

libbeat/feature/feature.go
```diff
@@ -0,0 +1,120 @@
+package feature
+
+import (
+	"fmt"
+	"reflect"
+)
+
+//go:generate stringer -type=Stability
```
urso: How about the 'generate stringer' command next to the type declaration or in a separate `0gen.go` file. The generated file already contains the type name. Having the generate command next to the type declaration makes it easier to find when navigating the code. Like:

```
// Comment...
// go:generate stringer -type=Stability
type Stability int
```

libbeat/feature/feature.go
```diff
@@ -0,0 +1,120 @@
...
+
+// Registry is the global plugin registry, this variable is mean to be temporary to move all the
```
urso: s/mean/meant/

libbeat/feature/feature.go
```diff
@@ -0,0 +1,120 @@
...
+	Factory() interface{}
+
+	// Stability returns the stability of the feature.
+	Stability() Stability
+
+	// Equal returns true if the two object are equal.
+	Equal(other Featurable) bool
+
+	String() string
+}
```
urso: 'X returns x' is kind of redundant. Plus godoc will not generate any docs for the methods. Instead the type will be shown in the docs as is in code. Rather then stating the obvious, describing how `Featurable` interacts with the Bundle and Registry would be more helpful. What's the purpose of Namespace. Why do I need to implement 'equal'?

libbeat/feature/feature.go
```diff
@@ -0,0 +1,120 @@
...
+
+// Stability returns the stability level of the feature, current: stable, beta, experimental.
+func (f *Feature) Stability() Stability {
+	return f.stability
+}
+
+// Equal return true if both object are equals.
+func (f *Feature) Equal(other Featurable) bool {
+	// There is no safe way to compare function in go,
+	// but since the method are global it should be stable.
```
urso: s/method/function pointers/

libbeat/feature/feature.go
```diff
@@ -0,0 +1,120 @@
...
+// Stability defines the stability of the feature, this value can be used to filter a bundler.
+type Stability int
+
+// List all the available stability for a feature.
+const (
+	Stable Stability = iota
+	Beta
+	Experimental
+	Undefined
+)
```
urso: Nice trick:

```
const (
```

libbeat/feature/registry.go
```diff
@@ -0,0 +1,141 @@
...
+
+	r.log.Debugw(
+		"registering new feature",
+		"namespace",
+		feature.Namespace(),
+		"name",
+		feature.Name(),
+	)
+
+	r.namespaces[feature.Namespace()][feature.Name()] = feature
```
urso: Instead of calling a function that might potentially generate a string on every call (it's all interfaces), let's introduce some local vars: `ns, name := feature.Namespace(), feature.Name()`.

libbeat/feature/registry.go
```diff
@@ -0,0 +1,141 @@
...
+
+	// Lazy create namespaces
+	_, found := r.namespaces[feature.Namespace()]
+	if !found {
+		r.namespaces[feature.Namespace()] = make(map[string]Featurable)
+	}
+
+	f, found := r.namespaces[feature.Namespace()][feature.Name()]
+	if found {
+		if feature.Equal(f) {
```
urso: We can remove `Equal` from the public interface by providing a helper function checking for equal:

```
func featuresEq(f1, f2 Featurable) bool {
```

libbeat/feature/registry.go
```diff
@@ -0,0 +1,141 @@
...
+	_, found = v[name]
+	if !found {
+		return fmt.Errorf("unknown feature '%s' in namespace '%s'", name, namespace)
+	}
+
+	delete(r.namespaces[namespace], name)
+	return nil
+}
+
+// Find returns a specific Find from a namespace or an error if not found.
```
urso: Find searches for a Feature by the namespace-name pair.

Alternative name: Lookup

libbeat/publisher/queue/queue_reg.go
```diff
@@ -4,23 +4,34 @@ import (
...
+	f, err := feature.Registry.Find(Namespace, name)
+	if err != nil {
+		return nil
+	}
+	factory, ok := f.Factory().(Factory)
+	if !ok {
+		return nil
+	}
+
+	return factory
```
urso: Using reflection, this pattern can be simplified to:

```
func FindFactory(name string) (f Factory) {
  if err := feature.Registry.Lookup(Namespace, name, &f); err != nil {
    // log error?
  }
  return
}
```

Using reflaction based function calling support, we could even implement an 'Instantiate/Invoke' method in the registry (maybe too much magic, though).

ph:
@ruflin Nothing prevents us to use code generation to generate an include with the new format, there is a point in the description for `Update the auto generation tools`

libbeat/feature/feature.go
ph: This was added automatically when you call `generate stringer -type=Stability` in the feature directory :)

libbeat/feature/feature.go
ph: Good point, I have added undefined at the end, will make the change.

libbeat/publisher/queue/queue_reg.go
ph: I think I would prefer to stay away from reflexion as much as possible in the context of plugin, this force user to make the choice.

libbeat/feature/registry.go
ph: I'm OK for using Lookup instead.

libbeat/publisher/queue/queue_reg.go
ph: I've created a private bundleable interface implement by both Feature and Bundle.

I've kept the `Feature` and `Bundle`  type public, since we will interact with it for creating new Features and filtering / merging bundle.

ph:
I've moved the TODO list to a meta issue at https://github.com/elastic/beats/issues/7423

PR #7393, opened 2018-06-21
Use type alias to avoid remarshalling Pod events in Kubernetes watcher

exekias:
This gives us freedom to use any field in the Pod object without the
need of bringing it to our own structs.

This is just a refactor that should not change any behavior. I plan to
use this in following PRs.


exekias:
jenkins, retest this please

metricbeat/module/kubernetes/event/event.go
```diff
@@ -139,28 +139,28 @@ func generateMapStrFromEvent(eve *kubernetes.Event) common.MapStr {
...
 
 	tsMap := make(common.MapStr)
 
 	if eve.FirstTimestamp != nil {
-		tsMap["first_occurrence"] = eve.FirstTimestamp.UTC()
+		tsMap["first_occurrence"] = kubernetes.Time(eve.FirstTimestamp)
 	}
 
 	if eve.LastTimestamp != nil {
-		tsMap["last_occurrence"] = eve.LastTimestamp.UTC()
```
ruflin: It seems we remove the conversion to UTC. Are all timestamps still UTC?

exekias: uhm that's a good point, I believe dates must be in UTC anyway, but I'll put the conversion back in, just in case.

exekias:
jenkins, retest this please

Issue #7368, opened 2018-06-19
[Filebeat] [Doc] Metricbeat `host` namespace causes mapping conflict when used with Logstash


jkelastic:
Need to add the following to the breaking changes part of the Filebeat / Logstash docs

Metricbeat `host` namespace causes mapping conflict when used with Logstash

For confirmed bugs, please report:
- Version: Filebeat and Logstash 6.3
- Github Link https://github.com/elastic/beats/issues/7050

```
After upgrade to 6.3.0, some logsources will cause an error in logstash and will not make it to elasticsearch. 

The error: 
[2018-06-15T22:22:53,246][WARN ][logstash.outputs.elasticsearch] Could not index event to Elasticsearch. {:status=>400, :action=>["index", {:_id=>nil, :_index=>"logstash-2018.06.15", :_type=>"doc", :_routing=>nil}, #<LogStash::Event:0x5b4f0227>], :response=>{"index"=>{"_index"=>"logstash-2018.06.15", "_type"=>"doc", "_id"=>"rg0dBWQB0H0urKNSctIj", "status"=>400, "error"=>{"type"=>"mapper_parsing_exception", "reason"=>"failed to parse [host]", "caused_by"=>{"type"=>"illegal_state_exception", "reason"=>"Can't get text on a START_OBJECT at 1:620"}}}}}
```

andrewkroh:
To clarify, I think you are proposing to have it listed on a new "Breaking Changes in 6.3" page linked from [here](https://www.elastic.co/guide/en/beats/libbeat/current/breaking-changes.html). Correct?

It is listed as a breaking change in the [release notes](https://www.elastic.co/guide/en/beats/libbeat/6.3/release-notes-6.3.0.html).

ppf2:
+1.  Users generally go to breaking changes for breaking changes and may miss them in release notes.
For this particular one, I will suggest not only highlight it in breaking changes, but also provide a list of potential workarounds to address the field conflicting type issue.

jkelastic:
@andrewkroh yes, please also include it in the breaking changes as customer don't read the release notes.

dedemorton:
Added a first draft of the breaking change docs here: https://github.com/elastic/beats/pull/7398

PR #7398, opened 2018-06-21
Add breaking change about host namespace causing mapping issues

dedemorton:
Fixes https://github.com/elastic/beats/issues/7368

I found the original document hard to follow (had to read it through multiple times), so I tried to rewrite it my own words based on what I understood after reading the document. We need to connect more of the dots for users who aren't familiar with this problem.

We don't have enough section levels in asciidoc to properly format this, so I've fudged some sections (don't tell the other writers).

Not sure it's 100% there, but here you go.

UPDATE: Please make sure you respond to the questions I've raised for reviewers.

### DIFF
--- a/libbeat/docs/breaking.asciidoc
+++ b/libbeat/docs/breaking.asciidoc
@@ -61,12 +61,11 @@ searches in Elasticsearch still work as expected.
 If you searched for the `host` field previously, you now need to search for the
 `host` and `host.name` fields.  
 
-If you have visualizations in Kibana that reference the `host` field, export the
-objects by using the Kibana API or UI, change any aggregation references to use
-`beat.hostname` instead of `host`, and then re-import the objects into Kibana.
+If you have multiple visualizations in Kibana that reference the `host` field,
+you can export the objects by using the Kibana API or UI, change any aggregation
+references to use `host.name` instead of `host`, and then re-import the
+objects into Kibana.
 
-// REVIEWERS: Are the above details correct? Got them from a comment Pius made
-// on ruflin's doc.
 
 //TODO: Add links to the Kibana docs for exporting/importing objects.
 
--- a/libbeat/docs/breaking.asciidoc
+++ b/libbeat/docs/breaking.asciidoc
@@ -23,10 +23,6 @@ Breaking Changes sections of the <<release-notes,release notes>>.
 [[breaking-changes-mapping-conflict]]
 ==== New `host` namespace may cause mapping conflicts for Logstash
 
-// REVIEWERS: I describe these as mapping conflicts throughout, but the acutal
-// error message says mapper parsing exception. I'm wondering if I should use
-// that instead?
-
 This breaking change applies only to users who send Beats events to Logstash.
 
 Starting with version 6.3, Beats provides an `add_host_metadata` processor for
@@ -44,14 +40,9 @@ conflict with the `host` object defined in the index template.)
 
 *What does this mean to you?*
 
-If you *are* using the index template and versioned indices recommended for
-Beats, you might need to modify your searches and visualizations. See:
+See the info for your particular use case:
 
 * <<beats-template-versioned-indices>>
-
-If you *are not* using the recommended index template and versioned indices,
-see the section for your use case:
-
 * <<custom-template-non-versioned-indices>>
 * <<beats-template-non-versioned-indices>>
 
@@ -68,11 +59,7 @@ searches in Elasticsearch still work as expected.
 *What do you need to change?*
 
 If you searched for the `host` field previously, you now need to search for the
-`host` and `host.name` fields. Or you can adjust your queries to use
-`beat.hostname`, instead of `host`, to work across all indices. 
-
-// REVIEWERS: Would it better just to tell users to uses to change their
-// searches to use `beat.hostname` instead of `host`?
+`host` and `host.name` fields.  
 
 If you have visualizations in Kibana that reference the `host` field, export the
 objects by using the Kibana API or UI, change any aggregation references to use
@@ -125,7 +112,9 @@ processor.
 +
 With this configuration, Beats drops the `host` fields before sending the
 event to Logstash, and Logstash adds a default `host` field, as it did with
-previous Beats versions.
+previous Beats versions. This approach resolves the mapping conflict, but you
+should plan to migrate your Logstash configurations to use `host.name` in
+future releases. 
 
 [[beats-template-non-versioned-indices]]
 ===== Use case: You use the Beats index template and your indices are not versioned
@@ -134,27 +123,42 @@ In this use case, you load the Beats index template manually into Elasticsearch,
 and send your data through Logstash, but you don’t use the versioned index
 pattern to create versioned indices.
 
-You cannot resolve the problem by dropping the `host.*` fields, as described
-earlier, because Logstash will add a default `host` field, resulting in a
-mapping conflict because the index template defines the `host` field as an
-object. 
+You cannot resolve the problem by dropping the `host.*` fields, because Logstash
+will add a default `host` field, resulting in a mapping conflict because the
+index template defines the `host` field as an object. 
 
 *What do you need to change?*
 
 To resolve the mapping issue, do *one* of the following:
 
-// TOOD: Should add an example of the LS filter here, don't you think?
+* Drop the `host.*` fields in a Logstash filter. For example:
++
+[source,yaml]
+----
+filter {
+  mutate {
+    remove_field => [ "[host]" ]
+  }
+}
+----
++
+With this configuration, there will be no `host` field in the final event at
+ingestion time, and the mapping conflict is avoided.
 
-* Drop the `host.*` fields in a Logstash filter. There will be no
-`host` field at ingestion time, and the mapping conflict is avoided.
 * Or:
 ** Modify the Beats index template by removing the `host.*` fields, and
-** Configure Beats to drop all `host.*` fields, as described earlier.
+** Configure Beats to drop all `host.*` fields, as described in
+<<custom-template-non-versioned-indices,this section>>.
 +
 This solution prevents a mapping conflict because the fields are no longer
 defined in the Elasticsearch template. Elasticsearch can use the `host`
 mapping created when Logstash added a default `host` field.
 
+The difference between these two approaches is that the first approach, using
+a Logstash filter, drops the `host` fields completely. There will be no `host`
+field in the final event. The second approach drops the `host` fields from the
+Beats event, but because Logstash adds a default `host` field, there will be a
+`host` field in the final event.
 
 [[breaking-changes-6.0]]
 === Breaking changes in 6.0
--- a/libbeat/docs/breaking.asciidoc
+++ b/libbeat/docs/breaking.asciidoc
@@ -124,8 +124,8 @@ and send your data through Logstash, but you don’t use the versioned index
 pattern to create versioned indices.
 
 You cannot resolve the problem by dropping the `host.*` fields, because Logstash
-will add a default `host` field, resulting in a mapping conflict because the
-index template defines the `host` field as an object. 
+will add a default `host` field, resulting in a mapping conflict with the
+`host` field defined as an object in the index template. 
 
 *What do you need to change?*
 
--- a/libbeat/docs/breaking.asciidoc
+++ b/libbeat/docs/breaking.asciidoc
@@ -9,8 +9,151 @@ changes, but there are breaking changes between major versions (e.g. 5.x to
 
 See the following topics for a description of breaking changes:
 
+* <<breaking-changes-6.3>>
 * <<breaking-changes-6.0>>
 * {auditbeat}/auditbeat-breaking-changes.html[Breaking changes in Auditbeat 6.2]
+
+[[breaking-changes-6.3]]
+=== Breaking changes in 6.3
+
+This section discusses the main changes that you should be aware of if you
+upgrade the Beats to version 6.3. Please also review the relevant
+Breaking Changes sections of the <<release-notes,release notes>>.
+
+[[breaking-changes-mapping-conflict]]
+==== New `host` namespace may cause mapping conflicts for Logstash
+
+// REVIEWERS: I describe these as mapping conflicts throughout, but the acutal
+// error message says mapper parsing exception. I'm wondering if I should use
+// that instead?
+
+This breaking change applies only to users who send Beats events to Logstash.
+
+Starting with version 6.3, Beats provides an `add_host_metadata` processor for
+adding fields, such as `host.name` and `host.id`, to Beats events. These fields
+are grouped under a `host` prefix and conform to the
+https://github.com/elastic/ecs[Elastic Common Schema (ECS)]. The `host` object
+is defined in the Elasticsearch index template even if the processor is not
+used.
+
+We've also added a `host.name` field to all events sent by Beats. This field
+prevents the Beats input plugin in Logstash from adding a default `host` field.
+(By default, the plugin adds a `host` field if the event doesn't already have
+one. We don't want the plugin to add this field because it causes a mapping
+conflict with the `host` object defined in the index template.)
+
+*What does this mean to you?*
+
+If you *are* using the index template and versioned indices recommended for
+Beats, you might need to modify your searches and visualizations. See:
+
+* <<beats-template-versioned-indices>>
+
+If you *are not* using the recommended index template and versioned indices,
+see the section for your use case:
+
+* <<custom-template-non-versioned-indices>>
+* <<beats-template-non-versioned-indices>>
+
+[[beats-template-versioned-indices]]
+===== Use case: You use the Beats index template and versioned indices
+
+In this use case, you load the versioned template manually and use the Beats
+versioned index pattern, `%{[@metadata][beat]}-%{[@metadata][version]}-%{+YYYY.MM.dd}`,
+as recommended in the {logstash-ref}/plugins-inputs-beats.html[Beats input
+plugin] documentation. This results in a `host` field in 6.2 indices, and a
+`host.name` field in 6.3 indices. There are no mapping conflicts. Cross index
+searches in Elasticsearch still work as expected. 
+
+*What do you need to change?*
+
+If you searched for the `host` field previously, you now need to search for the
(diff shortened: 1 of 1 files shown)


libbeat/docs/breaking.asciidoc
```diff
@@ -9,9 +9,152 @@ changes, but there are breaking changes between major versions (e.g. 5.x to
...
+is defined in the Elasticsearch index template even if the processor is not
+used.
+
+We've also added a `host.name` field to all events sent by Beats. This field
+prevents the Beats input plugin in Logstash from adding a default `host` field.
+(By default, the plugin adds a `host` field if the event doesn't already have
+one. We don't want the plugin to add this field because it causes a mapping
+conflict with the `host` object defined in the index template.)
+
+*What does this mean to you?*
```
dedemorton: Hmmm...the formatting here looks OK in the built docs, but not so good in GitHub, so I might change it.

libbeat/docs/breaking.asciidoc
```diff
@@ -9,9 +9,152 @@ changes, but there are breaking changes between major versions (e.g. 5.x to
...
+// REVIEWERS: I describe these as mapping conflicts throughout, but the acutal
+// error message says mapper parsing exception. I'm wondering if I should use
+// that instead?
+
+This breaking change applies only to users who send Beats events to Logstash.
+
+Starting with version 6.3, Beats provides an `add_host_metadata` processor for
+adding fields, such as `host.name` and `host.id`, to Beats events. These fields
+are grouped under a `host` prefix and conform to the
+https://github.com/elastic/ecs[Elastic Common Schema (ECS)]. The `host` object
```
dedemorton: I hope it's OK to call this the `host` object because saying `host` field doesn't convey the idea of the type being different.

dedemorton:
@karenzone If you have the time, can you take a look at this and offer suggestions for places where the language is confusing? Once you understand the problem, the details are easy to follow...but otherwise is a confusing situation.

libbeat/docs/breaking.asciidoc
```diff
@@ -9,9 +9,152 @@ changes, but there are breaking changes between major versions (e.g. 5.x to
...
+=== Breaking changes in 6.3
+
+This section discusses the main changes that you should be aware of if you
+upgrade the Beats to version 6.3. Please also review the relevant
+Breaking Changes sections of the <<release-notes,release notes>>.
+
+[[breaking-changes-mapping-conflict]]
+==== New `host` namespace may cause mapping conflicts for Logstash
+
+// REVIEWERS: I describe these as mapping conflicts throughout, but the acutal
```
ruflin: I like mapping conflict as it is a mapping conflict :-)

libbeat/docs/breaking.asciidoc
```diff
@@ -9,9 +9,152 @@ changes, but there are breaking changes between major versions (e.g. 5.x to
...
+`host.name` field in 6.3 indices. There are no mapping conflicts. Cross index
+searches in Elasticsearch still work as expected. 
+
+*What do you need to change?*
+
+If you searched for the `host` field previously, you now need to search for the
+`host` and `host.name` fields. Or you can adjust your queries to use
+`beat.hostname`, instead of `host`, to work across all indices. 
+
+// REVIEWERS: Would it better just to tell users to uses to change their
```
ruflin: `beat.hostname` is going to disappear in 7.0. So I would prefer to stear people in the direction of `host.name` if possible so they have to change it only once.

libbeat/docs/breaking.asciidoc
```diff
@@ -9,9 +9,152 @@ changes, but there are breaking changes between major versions (e.g. 5.x to
...
+// searches to use `beat.hostname` instead of `host`?
+
+If you have visualizations in Kibana that reference the `host` field, export the
+objects by using the Kibana API or UI, change any aggregation references to use
+`beat.hostname` instead of `host`, and then re-import the objects into Kibana.
+
+// REVIEWERS: Are the above details correct? Got them from a comment Pius made
```
ruflin: The above should work but if someone has only 1 search based on the `host` field I would rather recommend to directly edit it in Kibana instead of exporting / importing it. The above might be better if there are lots of queries dependening on it but I think it requires more knowledge as the files have to be edited manually. 

I would propose both options.

@ppf2 WDYT?

libbeat/docs/breaking.asciidoc
```diff
@@ -9,9 +9,152 @@ changes, but there are breaking changes between major versions (e.g. 5.x to
...
++
+--
+IMPORTANT: If you drop the `host` fields, you cannot use the `add_host_metadata`
+processor.
+
+--
++
+With this configuration, Beats drops the `host` fields before sending the
+event to Logstash, and Logstash adds a default `host` field, as it did with
+previous Beats versions.
```
ruflin: Long term I think the `host` field will disappear for the beats-input-plugin in Logstash. I wonder if we should make a recommendation here to start thinking about how to migrate long term to `host.name`.

libbeat/docs/breaking.asciidoc
```diff
@@ -9,9 +9,152 @@ changes, but there are breaking changes between major versions (e.g. 5.x to
...
+You cannot resolve the problem by dropping the `host.*` fields, as described
+earlier, because Logstash will add a default `host` field, resulting in a
+mapping conflict because the index template defines the `host` field as an
+object. 
+
+*What do you need to change?*
+
+To resolve the mapping issue, do *one* of the following:
+
+// TOOD: Should add an example of the LS filter here, don't you think?
```
ruflin: ++

@ph Could you provide one? 

There is a bit a difference if `host.*` is dropped by Beats or LS. If it is dropped by Beats, LS will add `host` field. If it's dropped by LS, there will be no host field at all in the final event. I wonder if we should mention that difference.

ph: @ruflin @dedemorton  Yes, the following example will bring to pre 6.3.0. 

```
mutate {
  remove_field => [ "[host]" ]
}
mutate {
  add_field => {
    "host" => "%{[beat][hostname]}"
   }
}
```

libbeat/docs/breaking.asciidoc
ph: It's a mapping conflict.

libbeat/docs/breaking.asciidoc
```diff
@@ -9,9 +9,152 @@ changes, but there are breaking changes between major versions (e.g. 5.x to
...
+Breaking Changes sections of the <<release-notes,release notes>>.
+
+[[breaking-changes-mapping-conflict]]
+==== New `host` namespace may cause mapping conflicts for Logstash
+
+// REVIEWERS: I describe these as mapping conflicts throughout, but the acutal
+// error message says mapper parsing exception. I'm wondering if I should use
+// that instead?
+
+This breaking change applies only to users who send Beats events to Logstash.
```
ph: And users using their own template.

libbeat/docs/breaking.asciidoc
ph: @ruflin but taking the content of `[beat][hostname]` for host will result in the same behavior as the input had. WDYT?

libbeat/docs/breaking.asciidoc
```diff
@@ -9,9 +9,152 @@ changes, but there are breaking changes between major versions (e.g. 5.x to
...
+https://github.com/elastic/ecs[Elastic Common Schema (ECS)]. The `host` object
+is defined in the Elasticsearch index template even if the processor is not
+used.
+
+We've also added a `host.name` field to all events sent by Beats. This field
+prevents the Beats input plugin in Logstash from adding a default `host` field.
+(By default, the plugin adds a `host` field if the event doesn't already have
+one. We don't want the plugin to add this field because it causes a mapping
+conflict with the `host` object defined in the index template.)
+
```
karenzone: The first sentence in the parentheses doesn't say anything new, so consider deleting it, and adding `host` to remaining sentence.

+(We don't want the plugin to add the `host` field because it causes a mapping
+conflict with the `host` object defined in the index template.)

libbeat/docs/breaking.asciidoc
karenzone: I see what you mean about formatting being weird in GitHub. That threw me off at first. Interesting that it builds correctly in the book.

libbeat/docs/breaking.asciidoc
```diff
@@ -9,9 +9,152 @@ changes, but there are breaking changes between major versions (e.g. 5.x to
...
+Beats, you might need to modify your searches and visualizations. See:
+
+* <<beats-template-versioned-indices>>
+
+If you *are not* using the recommended index template and versioned indices,
+see the section for your use case:
+
+* <<custom-template-non-versioned-indices>>
+* <<beats-template-non-versioned-indices>>
+
```
karenzone: +*What does this mean to you?*
See the info for your particular use case:
* <<link use case 1>>
* <<link use case 2>>
* <<link use case 3>>

You did a good job capturing the use case scenario in each title.  The use cases are easier to read and parse without the surrounding text.

libbeat/docs/breaking.asciidoc
ph: I will do a check, but the _add_field_ should replace the original value.

libbeat/docs/breaking.asciidoc
```diff
@@ -9,9 +9,152 @@ changes, but there are breaking changes between major versions (e.g. 5.x to
...
+[[beats-template-non-versioned-indices]]
+===== Use case: You use the Beats index template and your indices are not versioned
+
+In this use case, you load the Beats index template manually into Elasticsearch,
+and send your data through Logstash, but you don’t use the versioned index
+pattern to create versioned indices.
+
+You cannot resolve the problem by dropping the `host.*` fields, as described
+earlier, because Logstash will add a default `host` field, resulting in a
+mapping conflict because the index template defines the `host` field as an
```
karenzone: "earlier" might be problematic.  Linking to different use cases doesn't promote linear reading, but "earlier" seems to imply that you've read everything.  Here, I think we can delete the reference to earlier and just say, "You cannot resolve the problem by dropping the `host.*` fields,  because Logstash will add a default `host` field..."
Especially since what we're referencing isn't going to work in this situation anyway. :-)

libbeat/docs/breaking.asciidoc
```diff
@@ -9,9 +9,152 @@ changes, but there are breaking changes between major versions (e.g. 5.x to
...
+To resolve the mapping issue, do *one* of the following:
+
+// TOOD: Should add an example of the LS filter here, don't you think?
+
+* Drop the `host.*` fields in a Logstash filter. There will be no
+`host` field at ingestion time, and the mapping conflict is avoided.
+* Or:
+** Modify the Beats index template by removing the `host.*` fields, and
+** Configure Beats to drop all `host.*` fields, as described earlier.
++
```
karenzone: WRT "earlier".  Same concern applies (this document isn't designed to be read linearly). 

If we're referencing this: "In the Beats config file, configure Beats to drop all `host.*` fields: ", just say that instead of referencing it.

brandonmensing:
LGTM - nothing to add on top of the comments and todos already going on there. Thanks!

libbeat/docs/breaking.asciidoc
dedemorton: @ph You're saying that this issue applies to Beats users who use a custom template and aren't using Logstash, correct? If they have `host` field defined in their custom template, they'll run into mapping errors, no?

libbeat/docs/breaking.asciidoc
dedemorton: I had it that way originally, but thought being explicit might be useful for users who don't understand that LS adds the field if there's not one. This behavior isn't consistent across input plugins, so it might be worth calling out. But I agree...it's a bit redundant--that's why I made it parenthetical even though I know people hate parens. If no one disagrees, I will remove the statement.

libbeat/docs/breaking.asciidoc
dedemorton: Well the docs use our CSS to control the size of things. GitHub shouldn't make bolded text into a big section header, but it does.

libbeat/docs/breaking.asciidoc
dedemorton: OK, then I'll remove mention of this field.

libbeat/docs/breaking.asciidoc
ph: @dedemorton 
It applies to the following use cases:

- FB -> LS,  using the default logstash-output-elasticsearch template.
- FB -> LS,  using their own template templates with the host defined.

libbeat/docs/breaking.asciidoc
dedemorton: I'll add a link to the section because I want users to know that the topic describes how to do this.

libbeat/docs/breaking.asciidoc
dedemorton: @ph See my comment ^^

dedemorton:
I've pushed some changes based on review feedback. There are a couple of sections that still need work pending a response from reviewers:
@ppf2 - https://github.com/elastic/beats/pull/7398/files#r197348926

@ph @ruflin https://github.com/elastic/beats/pull/7398#discussion_r197524285

@tsg FYI...

libbeat/docs/breaking.asciidoc
ph: @dedemorton this is perfectly valid, thanks for  that :)
```
mutate {
  add_field => {
    "host" => "%{[host][name]}"
   }
}```

File: libbeat/docs/breaking.asciidoc (hunk 8 above)
Comment (ph) [reply libbeat/docs/breaking.asciidoc]: @dedemorton Yes this is true, so let's keep it more specific we can adjust if people are posting on discuss.

File: libbeat/docs/breaking.asciidoc [hunk 13]
```diff
@@ -9,9 +9,155 @@ changes, but there are breaking changes between major versions (e.g. 5.x to
...
+If you searched for the `host` field previously, you now need to search for the
+`host` and `host.name` fields.  
+
+If you have multiple visualizations in Kibana that reference the `host` field,
+you can export the objects by using the Kibana API or UI, change any aggregation
+references to use `host.name` instead of `host`, and then re-import the
+objects into Kibana.
+
+
+//TODO: Add links to the Kibana docs for exporting/importing objects.
```
ruflin: Want to resolve it as part of this PR?

dedemorton: @ruflin If you think it's good enough to merge, maybe we should merge and then iterate over it after it's merged. I think @tsg wanted to review this.

libbeat/docs/breaking.asciidoc
```diff
@@ -9,9 +9,155 @@ changes, but there are breaking changes between major versions (e.g. 5.x to
 
 See the following topics for a description of breaking changes:
 
+* <<breaking-changes-6.3>>
```
ppf2: For completeness, I think we will want to include headings for 6.1 and 6.2 as well since there were also breaking changes in these releases that are currently only documented in the release notes today.

https://www.elastic.co/guide/en/beats/libbeat/6.2/release-notes-6.1.0.html#_breaking_changes_3
https://www.elastic.co/guide/en/beats/libbeat/6.2/release-notes-6.2.0.html#_breaking_changes_2
https://www.elastic.co/guide/en/beats/libbeat/6.2/release-notes-6.2.3.html#_breaking_changes

libbeat/docs/breaking.asciidoc
```diff
@@ -9,9 +9,155 @@ changes, but there are breaking changes between major versions (e.g. 5.x to
...
+plugin] documentation. This results in a `host` field in 6.2 indices, and a
+`host.name` field in 6.3 indices. There are no mapping conflicts. Cross index
+searches in Elasticsearch still work as expected. 
+
+*What do you need to change?*
+
+If you searched for the `host` field previously, you now need to search for the
+`host` and `host.name` fields.  
+
+If you have multiple visualizations in Kibana that reference the `host` field,
```
ppf2: I will send a separate note on this shortly to discuss.  Somewhere in here, we will want to provide additional details on tackling the outcome of having multiple indices with different schemas for the Kibana use case given Kibana's inconsistent handling of conflicting field types today.  Stay tuned.  Just adding a comment here as something to refine.

libbeat/docs/breaking.asciidoc
ruflin: @ppf2 I think the difference with the aboves is that these we features which still were in beta.

dedemorton: I'll either add sections or make it clear that this section only covers the main breaking changes. We say that in a couple of places plus provide links to the relnotes, but maybe it's not clear enough.

ppf2: +1, let's clarify.  Are we saying that the breaking changes guide will only cover breaking changes against the GA product?  Or do we mean that it only covers the "main" breaking changes of GA product?  If "main" is the keyword here, let's also confirm that this is consistent with the rest of the Elastic stack.  Afaik, Elasticsearch documents all breaking changes against the GA product, not just the main ones.

If we are only covering breaking changes of the GA product, we can state that for breaking changes against pre-GA product releases, these are documented only in the release notes, etc.. 

thanks!

dedemorton: That's a good question, and I'm not sure about the answer. I'd like @tsg or @ruflin to weigh in here.

dedemorton:
I've created an issue to track the remaining work for this PR (just in case things get busy...I do plan to get this resolved asap): https://github.com/elastic/beats/issues/7408

libbeat/docs/breaking.asciidoc
ruflin: Moving forward we should document here all breaking changes which happen for any GA features. In general no breaking changes in minor should happen, but there are exceptions.

For beta or experimental features, the breaking changes should only be in the changelog.

ppf2:
@dedemorton @ruflin Can we merge this change to both 6.3 and 6.x branches?  I only see it in master right now :)

dedemorton:
@ppf2 I'll cherry-pick the changes today. Yesterday I was waiting for the doc build on master to be reflected in the published docs, but there was a problem with the site updating.

PR #7401, opened 2018-06-22
Metricbeat: MongoDB TLS connection support

### LINKED ISSUE #6619: Support TLS configuration in metricbeat mongodb module
Support [TLS configuration](https://docs.mongodb.com/manual/reference/program/mongo/#tls-ssl-options) in metricbeat mongodb module.

It could be supported by adding a `DialServer` to [`DialInfo`](https://godoc.org/gopkg.in/mgo.v2#DialInfo), that sets up connection configuration with TLS options when they need to be used.

jsoriano:
Add support for TLS in MongoDB metricbeat module. A common builder is also added for MongoDB metricsets.

Fixes #6619


metricbeat/docs/modules/mongodb.asciidoc
```diff
@@ -77,13 +77,30 @@ metricbeat.modules:
...
+  #ssl.key: "/etc/pki/client/cert.key"
+
   # Username to use when connecting to MongoDB. Empty by default.
   #username: user
 
   # Password to use when connecting to MongoDB. Empty by default.
   #password: pass
 ----
 
+This module supports TLS connection when using `ssl` config field, as described in <<configuration-ssl>>. It also supports the options described in <<module-http-config-options>>.
```
jsoriano: This message is autogenerated after adding the `ssl` settings to `fields.yml`. I think it shouldn't mention by default the http options, as for example this module uses TLS but doesn't use HTTP. To be fixed in a future PR.

metricbeat/module/mongodb/metricset.go
```diff
@@ -0,0 +1,70 @@
...
+	dialInfo.Timeout = base.Module().Config().Timeout
+
+	if config.TLS.IsEnabled() {
+		tlsConfig, err := tlscommon.LoadTLSConfig(config.TLS)
+		if err != nil {
+			return nil, err
+		}
+
+		dialInfo.DialServer = func(addr *mgo.ServerAddr) (net.Conn, error) {
+			return tls.Dial("tcp", addr.String(), tlsConfig.BuildModuleConfig(""))
```
ycombinator: Should we pass `addr.String()` to `tlsConfig.BuildModuleConfig` instead of `""`? Maybe it doesn't matter but reading through https://golang.org/pkg/crypto/tls/#Config it appears this value is used for hostname verification.

metricbeat/module/mongodb/status/status.go
```diff
@@ -47,24 +46,18 @@ func init() {
 // additional entries. These variables can be used to persist data or configuration between
 // multiple fetch calls.
 type MetricSet struct {
-	mb.BaseMetricSet
-	dialInfo *mgo.DialInfo
+	*mongodb.MetricSet
```
ycombinator: This may be an invalid/stupid question because I'm still new to beats + golang but why do we need to embed the `*mongodb.MetricSet` type inside this `MetricSet` struct? Couldn't we get rid of the outer/wrapper struct now and in the `New` function below do this instead:

```
    return mongodb.NewMetricSet(base)
}
```

metricbeat/module/mongodb/metricset.go
jsoriano: In other uses of this function we are passing the empty string but yes, it'd make sense to pass the hostname in the address, I'll try this.

metricbeat/module/mongodb/status/status.go
jsoriano: There are no stupid questions, this is a good one :)

Here all the metricsets share the same fields, so I use the same builder, but they need different implementations for the `Fetch()` method, so I embed the common data in structs implementing these specializations.

jsoriano:
@ycombinator thanks for your review!

I have added the server name to the TLS config, but after some tests I couldn't see any change in how it behaves. It seems that the [dialer method](https://golang.org/src/crypto/tls/tls.go?s=3255:3347#L88) already tries to infer the server name from the address if it is not set. In any case I'm fine with leaving it explicit.

metricbeat/module/mongodb/_meta/config.yml
```diff
@@ -11,6 +11,21 @@
   # password configuration options.
   hosts: ["localhost:27017"]
 
+  # Optional SSL. By default is off.
+  #ssl.enabled: true
+
+  # Mode of verification of server certificate ('none' or 'full')
+  #ssl.verification_mode: 'full'
+
+  # List of root certificates for HTTPS server verifications
```
ycombinator: Probably should say TLS instead of HTTPS (since there's no HTTP connection here)?

jsoriano: You are right, fixed.

PR #7413, opened 2018-06-26
Metricbeat module for Traefik

### LINKED ISSUE #7412: Metricbeat module for Traefik
Create a metricbeat module for the [Traefik](https://traefik.io/) reverse proxy/load balancer.

ycombinator:
Resolves #7412.

This PR adds a Metricbeat module for the [Traefik](https://traefik.io/) reverse proxy/load balancer. In this module, it adds the health metricset, whose data it retrieves from polling the Traefik instance's health HTTP API endpoint.


ycombinator:
@ruflin Looking at the `beats-ci` console output, the failure is this:

```
08:21:59 ./module/traefik/health/health.go
08:21:59 Code differs from goimports' style ^
```

How can I fix this?

@exekias Is there some VS Code extension I should use to auto-fix this on save or something? I'm currently using this in my VS Code config:

```
  // Pick 'gofmt', 'goimports', 'goreturns' or 'goformat' to run on format.
  "go.formatTool": "goreturns",
```

exekias:
You can use `make fmt` to fix these kind of issues.

As for VS Code, try adding this:

```
    "go.formatFlags": ["-local", "github.com/elastic/beats"],
```

metricbeat/module/traefik/health/data.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"github.com/elastic/beats/libbeat/common"
+	s "github.com/elastic/beats/libbeat/common/schema"
+	c "github.com/elastic/beats/libbeat/common/schema/mapstriface"
+)
+
+var (
+	schema = s.Schema{
+		"uptime": s.Object{
+			"sec": c.Float("uptime_sec"),
```
ruflin: Should this be Int?

metricbeat/module/traefik/_meta/fields.yml
```diff
@@ -0,0 +1,11 @@
+- key: traefik
+  title: "traefik"
+  description: >
+    experimental[]
```
ruflin: I think we now generate this automatically if the release setting is not set. So you can skip it here.

metricbeat/module/traefik/health/data.go
ycombinator: This value is a float in the response from Traefik's health API (http://localhost:8080/health) so I kept it the same here.

ruflin: In this case i would suggest we use milliseconds instead here, especially having https://github.com/elastic/elasticsearch/issues/31244 in mind. TBH for uptime I would also be fine to just convert it to Int. I don't think there is a benefit of having uptime in sub second detail.

ycombinator: > I don't think there is a benefit of having uptime in sub second detail.

Yeah, good point. Will change to seconds.

ycombinator:
I'm investigating the integration test failures by running them locally as you suggested. From what I can tell so far I think the failure has to do with the Docker container for Traefik not starting up correctly. I'm still playing with this to figure out what's wrong.

ruflin:
It seems that the container tries to expose port 80 instead of 8080. Is treafik by default running on 80 or 8080? Perhaps you have change the config here and add `EXPOSE 8080` to the docker file.

ycombinator:
Turned out to be a combination of issues in the Dockerfile, you can see the necessary changes to make the `HEALTHCHECK` work here: 2fcc0de27355bd68c2719850ed6e207992dcb094.

metricbeat/module/traefik/health/_meta/data.json
```diff
@@ -0,0 +1,19 @@
+{
```
ycombinator: What is the purpose of this file? How is it used?

ruflin: This is an example output document that we should in the docs. It is generated by this command here: https://github.com/elastic/beats/blob/master/metricbeat/Makefile#L73 As this updates all data.json I normally run the following command:

```
go test -tags=integration github.com/elastic/beats/metricbeat/module/traefik/... -data
```
In you case this assumes the traefik docker container is running and exposes the port on localhost. This will generate a data.json file and `make update` will link it in the docs.

Here you can see an example in the docs: https://www.elastic.co/guide/en/beats/metricbeat/master/metricbeat-metricset-elasticsearch-index.html

metricbeat/module/traefik/health/_meta/fields.yml
```diff
@@ -0,0 +1,45 @@
+- name: health
+  type: group
+  description: >
+    Metrics obtained from Traefik's health API endpoint
+  fields:
+    - name: uptime
```
ruflin: One trick you can do here not to have to specify the group and describe is to directly write `uptime.sec`. The generators will do the rest of the magic for you.

metricbeat/module/traefik/health/_meta/fields.yml
```diff
@@ -0,0 +1,45 @@
...
+          type: int
+          description: >
+            Number of responses
+        - name: avg_time
+          type: group
+          description: >
+            Average response time
+          fields:
+            - name: sec
+              type: float
```
ruflin: As discussed with the other sec type I struggle that we have a `sec` value that is float. I understand that for the average it makes more sense but in this case I would rather use a smaller unit.

metricbeat/module/traefik/health/_meta/fields.yml
```diff
@@ -0,0 +1,45 @@
...
+              description:
+                  Average response time in seconds
+        - name: status_codes
+          type: group
+          description: >
+            Respone status codes
+          fields:
+            - name: code
```
ruflin: Not sure if this is correct. If understand the code correct it will be `status_codes.404: 17`? So `code` is actually the key and not the value? For such cases have a look at the dynamic templates. Search for `object_type` usage in our fields.yml to see how it can be used.

metricbeat/module/traefik/health/data.go
```diff
@@ -0,0 +1,50 @@
...
+	schema = s.Schema{
+		"uptime": s.Object{
+			"sec": c.Int("uptime_sec"),
+		},
+		"response": s.Object{
+			"count": c.Int("total_count"),
+			"avg_time": s.Object{
+				"sec": c.Float("average_response_time_sec"),
+			},
+			"status_codes": s.Object{},
```
ruflin: I would not define it here but leave it empty. Like this if we don't have any status codes, no object shows up in Elasticsearch.

metricbeat/module/traefik/health/health_test.go
```diff
@@ -0,0 +1,91 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/elastic/beats/libbeat/common"
+	mbtest "github.com/elastic/beats/metricbeat/mb/testing"
+)
+
+// raw response copied from Traefik instance's health API endpoint
+const response = `{
```
ruflin: Instead of having it to define here, I would recommend to put this into a file which is loaded. This makes it then very easy to have a list of files to test against, for example different versions. See Elasticsearch module as an example.

metricbeat/tests/system/test_traefik.py
```diff
@@ -0,0 +1,36 @@
...
+import unittest
+import time
+
+
+class Test(metricbeat.BaseTest):
+
+    COMPOSE_SERVICES = ['traefik']
+
+    @unittest.skipUnless(metricbeat.INTEGRATION_TESTS, "integration test")
+    def test_health(self):
```
ruflin: there is a new method `check_metricset` in metricbeat.py. Using it could simplify your code here.

metricbeat/tests/system/test_traefik.py
```diff
@@ -0,0 +1,36 @@
+import os
```
ruflin: We recently started to put this file directly into the module directory (see Elasticsearch) to have all pieces in one place. But still both works.

metricbeat/module/traefik/health/_meta/fields.yml
ycombinator: Makes sense, will use milliseconds and int here.

CHANGELOG.asciidoc
```diff
@@ -119,6 +119,7 @@ https://github.com/elastic/beats/compare/v6.2.3...master[Check the HEAD diff]
 - Fix field mapping for the system process CPU ticks fields. {pull}7230[7230]
 - Fix Windows service metricset when using a 32-bit binary on a 64-bit OS. {pull}7294[7294]
 - Fix Jolokia attribute mapping when using wildcards and MBean names with multiple properties. {pull}7321[7321]
+- Added Traefik module with health metricset. {pull}7413[7413]
```
jsoriano: Move this entry to the `Added` section

metricbeat/module/traefik/health/_meta/fields.yml
jsoriano: For proxies performance monitoring it is useful to have sub-millisecond response times if available.

metricbeat/module/traefik/health/health_integration_test.go
```diff
@@ -0,0 +1,54 @@
...
+
+	ms := mbtest.NewReportingMetricSetV2(t, traefik.GetConfig("health"))
+	reporter := &mbtest.CapturingReporterV2{}
+
+	ms.Fetch(reporter)
+	assert.Nil(t, reporter.GetErrors(), "Errors while fetching metrics")
+
+	event := reporter.GetEvents()[0]
+	assert.NotNil(t, event)
+	t.Logf("%s/%s event: %+v", ms.Module().Name(), ms.Name(), event)
```
jsoriano: Could you also check something in the event fields?

CHANGELOG.asciidoc
ycombinator: Oops, good catch. Thanks!

ycombinator:
@jsoriano @ruflin Thanks for your review. I believe I've addressed all the feedback now. Ready for your 👀 again.

metricbeat/module/traefik/test_traefik.py
```diff
@@ -0,0 +1,28 @@
+import os
+import unittest
+import time
+from parameterized import parameterized
+
+sys.path.append(os.path.join(os.path.dirname(__file__), '../../tests/system'))
```
jsoriano: `import sys` :)

ycombinator: Ahahaha, thank you.

Related: Is there some way to run system tests (i.e. the Python ones) for a specific module?

ruflin: `nosetests module/traefik/` should do the trick

metricbeat/module/traefik/testing.go
```diff
@@ -0,0 +1,49 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
```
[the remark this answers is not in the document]
jsoriano: You are right, I'd prefer that, yes.

ycombinator:
@ruflin Made the changes you suggested in the latest round of review. Ready for your eyes again, when you get a chance. Thanks!

metricbeat/module/traefik/health/health.go
```diff
@@ -0,0 +1,91 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+package health
+
+import (
+	"github.com/pkg/errors"
+
+	"github.com/elastic/beats/libbeat/common"
+
```
ruflin: newline can be removed

metricbeat/module/traefik/health/health.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+var (
+	hostParser = parse.URLHostParserBuilder{
+		DefaultScheme: "http",
+		DefaultPath:   "health",
+	}.Build()
+)
+
+// MetricSet holds any configuration or state information. It must implement
```
ruflin: I often remove the generator comments and make them more specific to the metricset or just very short. The comments were intended for someone that creates the metricset. Applies also to all other comments.

metricbeat/module/traefik/health/data.go
```diff
@@ -0,0 +1,57 @@
...
+			"avg_time": s.Object{
+				"us": c.Int("average_response_time_us"),
+			},
+		},
+	}
+)
+
+func eventMapping(health map[string]interface{}) (common.MapStr, *s.Errors) {
+	if averageResponseTimeSec, ok := health["average_response_time_sec"]; ok {
+		health["average_response_time_us"] = averageResponseTimeSec.(float64) * 1000 * 1000
```
ruflin: What happens if for whatever reason `averageResponseTimeSec` cannot be typed to `(float64)`

metricbeat/modules.d/traefik.yml.disabled
```diff
@@ -0,0 +1,7 @@
+# Module: traefik
+# Docs: https://www.elastic.co/guide/en/beats/metricbeat/master/metricbeat-module-traefik.html
```
ruflin: I like the link to the docs. One issue this has is that we `hardcode` master here to every released config file will point to master. Would be nice if in the future this would be a variable that is replaced by the version during packaging.

ycombinator:
@ruflin Addressed latest round of review feedback. Ready for your 👀 again. Thanks!

ruflin:
@ycombinator Thanks for pushing this forward.

ycombinator:
Backported to:

`6.x` / `6.4.0`: 0ec0e7f

PR #7421, opened 2018-06-26
Updating docs to use new `Fetch` method signature

ycombinator:
Previously, the `Fetch` method for `MetricSet`s would accept no arguments and return `(common.MapStr, error)`, which would be used to publish the event or error.

Now the preferred way is to use the `Fetch` method which accepts `mb.ReporterV2` as its argument and returns nothing. Inside the method, the `mb.ReporterV2` object is used to publish the event or error.

This PR updates the developer guide to consistently use the preferred way in all places.


dedemorton:
@ycombinator I didn't have a chance to review this before it got merged, but it looks good. Thank you for doing this!

PR #7426, opened 2018-06-26
Enforce normalization on Namespace and Name and check for nil factory

ph:
Most of the registries implementation in libbeat were not doing
normalization on the string, except for the 'Autodiscover' feature.
Since all code will now use the same registry, I've opted to lowercase
the namespace and the feature name before registering or doing any lookup this will
make sure we don't introduce any breaking changes.


ph:
I will still add a few things from the autodiscover logic.

PR #7430, opened 2018-06-27
Add generation of generated files for Filebeat module tests

ruflin:
Filebeat module tests allow to provide an `*-expected.json` file to compare the ingested data with the expected data. So far these files had to be generated manually by exporting the data from Elasticsearch. To make the generation easier now the environment variable `GENERATE=1` can be used and the existing file will be overwritten.

As the files are now generated, the structure of the files changed a little bit:

* fields like source path, event.created, beat.hostname are removed from the events as they change between tests.
* Objects are flattened for easier comparison
* The header response part with `_id` etc. was removed as it's not needed.

All expected.json files are overwritten in this PR to follow the new schema.

An example command to overwrite the generated files for Kibana looks as following:
```
INTEGRATION_TESTS=1 TESTING_FILEBEAT_MODULES=kibana GENERATE=1 nosetests tests/system/test_modules.py
```

Others:

* Fix naming of Kibana expected file.
* Remove extended check environment variable as all modules go now through the extended check.
* Skip @timestamp comparison for filesets without timestamp in log line.


filebeat/module/apache2/access/test/test.log-expected.json
```diff
@@ -1,205 +1,75 @@
...
+        "apache2.access.method": "GET", 
+        "apache2.access.remote_ip": "::1", 
+        "apache2.access.response_code": "404", 
+        "apache2.access.url": "/favicon.ico", 
+        "apache2.access.user_name": "-", 
+        "fileset.module": "apache2", 
+        "fileset.name": "access", 
+        "input.type": "log", 
+        "offset": 0, 
+        "prospector.type": "log"
```
exekias: this is much more readable! I wonder if we should do this change for example `data.json` files, and Metricbeat tests in general

filebeat/tests/system/test_modules.py
```diff
@@ -158,26 +167,22 @@ def _test_expected_events(self, module, test_file, objects):
...
                 # Flatten objects for easier comparing
                 obj = self.flatten_object(obj, {}, "")
-                ex = self.flatten_object(ev["_source"], {}, "")
-
                 clean_keys(obj)
-                clean_keys(ex)
 
-                # All modules in EXTENDED_COMPARE are checked in more detail
-                if ex == obj:
+                # Remove timestamp for comparison where no timestamp is not part of the log line
```
exekias: there is a typo in this sentence, `where **no** timestamp is not part...`

filebeat/module/apache2/access/test/test.log-expected.json
ruflin: My intention with this was actually to make it easier for machines.

I personally like the nested approach to read documents especially in the editor as I can "fold" parts. We could "flatten" all objects. What I worry is that people will think this is the way it's stored in Elasticsearch but it's actually stored as objects.

filebeat/tests/system/test_modules.py
ruflin: fixed

PR #7434, opened 2018-06-27
Improve autodiscover tests resiliency to timing issues

exekias:


libbeat/autodiscover/autodiscover_test.go
```diff
@@ -283,3 +288,17 @@ func TestAutodiscoverHash(t *testing.T) {
 	assert.True(t, runners[1].started)
 	assert.False(t, runners[1].stopped)
 }
+
+func wait(t *testing.T, test func() bool) {
+	sleep := 20 * time.Millisecond
+	ready := test()
+	for !ready && sleep < 10*time.Second {
+		time.Sleep(sleep)
+		sleep = sleep + 1*time.Second
```
ruflin: Why do you keep incrementing sleep? Couldn't you set it to 10*time.Millisecond as you had it before and then you have max 1000 iterations to reach 10s. In the above it could be that tests wait 1s even though the delay was only 21ms?

ruflin:
Merged so we can see if it solves the problem. Comment can still be addressed but not very important.

PR #7435, opened 2018-06-27
Fix naming of fields for galera status

jsoriano:
Follow-up of #6892


jsoriano:
Failing test doesn't seem related.

PR #7441, opened 2018-06-27
Refactor asset generation

ruflin:
* Extract asset template part into libbeat so it can be reused
* Have a Beat and a Name as part an asset definition. This will be later used for the module name.


libbeat/asset/asset.go
```diff
@@ -0,0 +1,49 @@
+package asset
+
+import "html/template"
```
andrewkroh: This was using `text/template`.

ruflin: You just saved my day. I already started to smash my head against the wall on why there is a diff in the fields.go content but there shouldn't be one. I definitively owe you some beers. 🍻

ruflin:
Ready for a round of reviews.

PR #7443, opened 2018-06-27
Refactor: Move Autodiscovery to the new global registry

ph:
The autodiscovery feature was defining 3 customs registry, this PR do
the following:

- Provider, Appender and Builder now users the feature registry.
- It move the Factory type into their right package.
- Config are now inside their own package.
- Use the same naming for finding factories and building object.
- type aliasing is in place to make the transition smoother for types.
- A shim is in place that will proxy the registry of plugin to the
global registry.
- package.Plugin methods are now returning a feature instead of a
hashmap.

### DIFF
--- a/libbeat/autodiscover/appenders/feature.go
+++ b/libbeat/autodiscover/appenders/feature.go
@@ -58,7 +58,7 @@ func FindFactory(name string) (Factory, error) {
 
 	factory, ok := f.Factory().(Factory)
 	if !ok {
-		return nil, fmt.Errorf("incompatible type for appender, received: '%T'", factory)
+		return nil, fmt.Errorf("incompatible type for appender, received: '%T'", f.Factory())
 	}
 
 	return factory, nil
--- a/libbeat/autodiscover/appenders/feature.go
+++ b/libbeat/autodiscover/appenders/feature.go
@@ -1,0 +1,81 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package appenders
+
+import (
+	"fmt"
+
+	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/common/bus"
+	"github.com/elastic/beats/libbeat/feature"
+	"github.com/elastic/beats/libbeat/processors"
+)
+
+// Namespace is the registry namespace for autodiscover appenders.
+var Namespace = "libbeat.autodiscover.appender"
+
+// Appender provides an interface by which extra configuration can be added into configs
+type Appender interface {
+	// Append takes a processed event and add extra configuration
+	Append(event bus.Event)
+}
+
+// Config settings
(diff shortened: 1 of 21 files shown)


ph:
The adding/remove stats on the diff are a bit skewed because of creating new files with the license headers..

exekias:
This is looking awesome! There are some beat specific builders and appenders that you may want to convert too:

https://github.com/ph/beats/blob/d6a981a6c9819d29cf1e39b9d1f8be193a20a49a/filebeat/autodiscover/builder/hints/logs.go
https://github.com/ph/beats/blob/d6a981a6c9819d29cf1e39b9d1f8be193a20a49a/metricbeat/autodiscover/builder/hints/metrics.go
https://github.com/ph/beats/blob/d6a981a6c9819d29cf1e39b9d1f8be193a20a49a/metricbeat/autodiscover/appender/kubernetes/token/token.go

If I understand this correctly they still work after this change, by using the wrapper backwards compatibility mechanism, but it may make sense to convert them to all this goodness :)

exekias:
Also pinging @vjsamuel for awareness, as he uses the builders plugin interface.

ph:
@exekias Thanks, I've seen the filebeat autodiscover, I will move them in another PR. But yes they will still work with the backward compatibility layer.

I am currently testing go-plugins to make sure they still work as intended.

libbeat/autodiscover/appenders/feature.go
```diff
@@ -0,0 +1,81 @@
...
+var Namespace = "libbeat.autodiscover.appender"
+
+// Appender provides an interface by which extra configuration can be added into configs
+type Appender interface {
+	// Append takes a processed event and add extra configuration
+	Append(event bus.Event)
+}
+
+// Config settings
+type Config struct {
```
urso: We need to export this type?

libbeat/autodiscover/providers/jolokia/jolokia.go
```diff
@@ -19,15 +19,16 @@ package jolokia
...
 	"github.com/elastic/beats/libbeat/common/bus"
 	"github.com/elastic/beats/libbeat/common/cfgwarn"
+	"github.com/elastic/beats/libbeat/feature"
 )
 
-func init() {
-	autodiscover.Registry.AddProvider("jolokia", AutodiscoverBuilder)
-}
+// Feature exposes the Jolokia autodiscovery provider.
+var Feature = providers.Feature("jolokia", AutodiscoverBuilder, feature.Experimental)
```
urso: The autodiscover builder logs an experimental warning. If possible the warnings about beta/experimental features should be logged by the registry. 

How about adding support for 'Name' and 'doc strings' to features.  In the future we will also have to add a config schema here. We might even register a fields.yml here, for building the actual event schema of a running beat. Some of these might be optional.
This allows us to create commands to print available features. This makes me wonder if we want to introduce a generic such that a Feature is registered like this:

```
var Feature = providers.Feature("jolokia", AutodiscoverBuilder, feature.Details{
  FullName: "Jolokia Discovery",
  Doc: "<short one liner doc>",
  Stability: feature.Experimental,
  Settings: ... // <schema definition>,
  Fields: ... // <fields.yml> documenting fields being added to events
  FullDoc: ... // very long doc string (or function for producing doc strings... have developer create docs in "docs.go")
})
```

With these information we can:
- build config validation
- create a minimal fields.yml
- provide a command `filebeat features all`
- provide a command like `filebeat help autodiscovery.jolokia` to print module docs

urso:
Nice cleanup

libbeat/autodiscover/appenders/feature.go
ph: We have a type alias on the autodiscover package so we need to export it.

libbeat/autodiscover/providers/jolokia/jolokia.go
exekias: If things go that route central management could use this too, so we can reload a full feature from a new config.

ph: @dedemorton We can still keep all the docs outside the main go code and just provide helpers method that will read the files when we create the `Feature.Detail`.

@urso I will make the changes to receive a `feature.Detail` instead of a stability for future proofing it.

PR #7444, opened 2018-06-27
Update go-structform to 0.0.4

urso:
- Update go-structform to version 0.0.4. New version adds support to disable HTML escaping (used in follow up PR #7445)


ph:
@urso Can you add in the commit message what changed in 0.0.4? It just easier when you go through commit history when looking for bugs.

urso:
@ph Nothing really. But see followup PR #7445

ph:
@urso I was missing the followup PR.

Issue #6106, opened 2018-01-18
Refactors for 7.0 (Breaking changes I want to make)


andrewkroh:
This is a running list of things to consider changing for 7.0. Feel free to add your own with attribution.

- [ ] Refactor the logging configuration (remove all the `to_files`, `to_syslog` settings etc.).
- [ ] Make Windows event log the default log output for Windows.
- [x] Set escape_html to false: elastic/beats#7445 elastic/beats#9914 @urso
- [x] Remove notion of **prospector** in the filebeat config. (@ph) PR https://github.com/elastic/beats/pull/8909
- [ ] Cleanup the methods used by `go-plugins`
- [ ] Default Template Mappings for IP fields are keyword type elastic/beats#7847 elastic/beats#9303  @ruflin
- [x] Remove ports from LS and Redis https://github.com/elastic/beats/issues/7881 elastic/beats#9934  @urso
- [x] Dissect `*` and named skip field change https://github.com/elastic/beats/issues/8054, PR At https://github.com/elastic/beats/pull/8750 (@ph)
- [x] Make beats log to journald by default on systems with systemd (add `-e` by default to systemd unit) (@jsoriano) elastic/beats#8942 
- [x] Dedot docker and k8s labels by default, ~or use [new queryable object type](https://github.com/elastic/elasticsearch/issues/25312) elastic/beats#9286~ (@kaiyan-sheng) 
- [x] Remove calls to `cfgwarn.CheckRemoved5xSetting(s)` (@ph) https://github.com/elastic/beats/pull/9002
- [ ] Remove 5.x dashboard directories, add 7 directories? elastic/beats#8927 
- [x] Assert any 'deprecated' comment in beats. (@ph)
- [ ] Change types to double in metricbeat modules for big gauges (@jsoriano, see #8991, #9271)
- [x] Rename `custom.*` to `journald.custom.*` fields (@kvch, already on master, needs changelog entry)
- [x] New location of registry of Journalbeat (@kvch, already on master, needs BC and changelog)
- [x] Deprecate fields/tags and introduce add_fields/add_tags processors https://github.com/elastic/beats/issues/4518 elastic/beats#9973 (@urso)
- [x] Remove ES 1.x and 5.x support in template/dashboard loading code
- [x] Unify module names for the same services elastic/beats#9395 (@jsoriano)
- [ ] Fixed any issues with the _xpack deprecation elastic/beats#9424

ph:
[...]

PR #7445, opened 2018-06-27
Html escaping

urso:
Requires: #7444 
Closes: #2581 

Add support to codecs and outputs to enable/disable escaping of html symbols in JSON strings.

By default html escaping is enabled.

- Add new setting `escape_html` to `codec.json`
- Add new setting `output.elasticsearch.escape_html`
- Add new setting `output.logstash.escape_html`
- Add settings to reference config files and output documentation

### DIFF
--- a/libbeat/outputs/codec/json/json.go
+++ b/libbeat/outputs/codec/json/json.go
@@ -31,18 +31,21 @@ import (
 
 // Encoder for serializing a beat.Event to json.
 type Encoder struct {
-	buf     bytes.Buffer
-	folder  *gotype.Iterator
-	pretty  bool
+	buf    bytes.Buffer
+	folder *gotype.Iterator
+
 	version string
+	config  config
 }
 
 type config struct {
-	Pretty bool
+	Pretty     bool
+	EscapeHTML bool
 }
 
 var defaultConfig = config{
-	Pretty: false,
+	Pretty:     false,
+	EscapeHTML: true,
 }
 
 func init() {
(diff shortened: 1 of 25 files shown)


libbeat/outputs/codec/json/json.go
```diff
@@ -54,19 +57,23 @@ func init() {
...
-func New(pretty bool, version string) *Encoder {
-	e := &Encoder{pretty: pretty, version: version}
+func New(pretty, escapeHTML bool, version string) *Encoder {
+	e := &Encoder{version: version, config: config{
+		Pretty:     pretty,
+		EscapeHTML: escapeHTML,
+	}}
 	e.reset()
 	return e
 }
```
ph: Maybe a later refactoring would change the signature to take an object instead. No need to do it now.

ph:
Happy to merge this, I've looked at the current failure in the metricbeat job, this is not due to the change in this PR but a timeout on k8s.

ruflin:
WDYT about having it off by default for 7.0? I would expect by default that no escaping is happening.

andrewkroh:
+1 for changing this in 7.0 to disabled.

ph:
Also agree with the above.

ruflin:
I added this to the list here: https://github.com/elastic/beats/issues/6106

I wonder if we should even remove to make it configurable in 7.0. What would be the reasons to enable it?

urso:
Not sure if we should disable (or even remove) the setting by default. So far we've only heard from users not wanting to escape html. But we don't know how many users actually did rely on beats escaping the HTML symbols. Not escaping HTML puts consumers displaying raw event contents in browsers at risk. Reason escaping is still enabled by default is, I didn't want to introduce potential security issues for those users.

ruflin:
Yes, we should not disable it 6.x as it would be a breaking change.

Good point about the potential security implications for some users. If we disable it, we should mention it clearly that this could have an affect. In general I still think we can disable / remove it in 7.0 as I don't think escaping should be part of the ingesting data but should be done by the consumer if needed.

PR #7446, opened 2018-06-27
Fix doc on output.kafka.version setting

urso:


ruflin:
Does this need a backport?

PR #7449, opened 2018-06-27
 Fix a panic when dealing with remaining at the end of a delimiter

ph:
When the last delimiter is not greedy, and we have captured all the data
for the defined fields in the Tokenizer string, we should ignore the
remaining of the string.


ph:
This issue was reported by @vjsamuel

vjsamuel:
@ph thanks for this fix. @ruflin @exekias if you could fast track this, it would be great :)

ph:
Dissect is not in 6.3 so no need to backport it.

PR #7453, opened 2018-06-27
Clarify purpose of breaking changes doc

dedemorton:
During the review of [this PR](https://github.com/elastic/beats/pull/7398), we discussed clarifying the [breaking changes](https://www.elastic.co/guide/en/beats/libbeat/current/breaking-changes.html) topic so users understand that it doesn't cover changes to beta and experimental functionality. 

I've changed the language to clarify.

I've also added sections for 6.2 and 6.1. Right now, those sections just contain the text that links to the release notes. In future releases, we should say something like, "There are no breaking changes to GA-level functionality in this release. {see-relnotes}


ppf2:
Not able to test the actual links, but lgtm, thx!

PR #7457, opened 2018-06-28
Add additional types to kubernetes metadata

vjsamuel:
The lack of the below type definitions is impacting testing custom indexers that are loaded in as go-plugins. It would be nice to have them defined as well.


vjsamuel:
@exekias done.

PR #7458, opened 2018-06-28
Add kubernetes to services which are kept running

ruflin:
Our metricbeat tests on Jenkins are flaky because of Kubernetes. My current assumption is that there are some overlaps when starting / stopping kuberenetes. Reason for this is that we do not see the same issue on Travis. The different between Travis and Jenkis is probably the number of processes which run in parallel. So either we could reduce the number of processes for the tests to 1 or add kubernetes to the services that keep running. I decided to keep it running for now to see what the outcome is.

* Changing heartbeat testing ports as sometimes they conflicted with other services running.

### DIFF
--- a/heartbeat/tests/system/test_monitor.py
+++ b/heartbeat/tests/system/test_monitor.py
@@ -21,7 +21,7 @@ class Test(BaseTest):
         self.render_config_template(
             monitors=[{
                 "type": "http",
-                "urls": ["http://localhost:8181"],
+                "urls": ["http://localhost:8185"],
             }],
         )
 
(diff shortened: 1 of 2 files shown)


ruflin:
jenkins, test this

heartbeat/tests/system/test_monitor.py
```diff
@@ -21,7 +21,7 @@ def test_http(self, status_code):
         self.render_config_template(
             monitors=[{
                 "type": "http",
-                "urls": ["http://localhost:8181"],
+                "urls": ["http://localhost:8185"],
```
exekias: Is this change intended? I don't see anything about it in the description

ruflin: yes, I just added a commit a few minutes ago. Some heartbeat tests were failing because port already in use. I'm not sure if it was related to the k8s change as I saw it before. But I thought I also fix it.

I need to update the description of builds go green.

PR #7464, opened 2018-06-28
Migrate: Queue to the global registry

ph:


libbeat/publisher/queue/feature.go
```diff
@@ -27,17 +27,12 @@ import (
 // Namespace is the feature namespace for queue definition.
 var Namespace = "libbeat.queue"
 
-// Global queue type registry for configuring and loading a queue instance
-// via common.Config
-var queueReg = map[string]Factory{}
-
 // RegisterType registers a new queue type.
 func RegisterType(name string, fn Factory) {
```
urso: Looks like we can remove RegisterType

ph: We keep it for backward compatibility. :(

ph:
@urso ready to be merged in the feature branch.

PR #7467, opened 2018-06-28
Update log builder settings unpack testing

exekias:
https://github.com/elastic/beats/pull/7510 has fixed a bug when unpacking into `*common.Config`, this PR updates a test in autodiscover logs builder to make sure that's working

### DIFF
--- a/filebeat/autodiscover/builder/hints/config.go
+++ b/filebeat/autodiscover/builder/hints/config.go
@@ -39,3 +39,22 @@ func defaultConfig() config {
 		Config: cfg,
 	}
 }
+
+// Unpack is needed here as go-ucfg fails to unpack the Config object by default
+func (c *config) Unpack(from *common.Config) error {
+	tmpConfig := struct {
+		Key string `config:"key"`
+	}{
+		Key: c.Key,
+	}
+	if err := from.Unpack(&tmpConfig); err != nil {
+		return err
+	}
+
+	if config, err := from.Child("config", -1); err == nil {
+		c.Config = config
+	}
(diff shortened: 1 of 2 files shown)


vjsamuel:
thanks @exekias i am good with this change.

filebeat/autodiscover/builder/hints/config.go
```diff
@@ -39,3 +39,25 @@ func defaultConfig() config {
...
+	tmpConfig := struct {
+		Key string `config:"key"`
+	}{
+		Key: c.Key,
+	}
+	if err := from.Unpack(&tmpConfig); err != nil {
+		return err
+	}
+
+	if config, err := from.Child("config", -1); err == nil {
```
vjsamuel: we do not need to unpack the config here and can be:
```
	if config, err := from.Child("config", -1); err == nil {
		c.Config = config
	}

```
This would help builders that inherit this one like how we do.

vjsamuel:
thanks @exekias for accommodating the change

urso:
The custom Unpack should not be required. It's exactly the use-case handled be go-ucfg. Do we have an unit test which fails without the custom unpack? I'd rather like to check/fix this in go-ucfg instead.

exekias:
sounds reasonable to me! the test is updated in this PR to fail if `Unpack` is not correct, thank you for having a look!

exekias:
Once https://github.com/elastic/go-ucfg/pull/106 is merged and vendored here we can get rid of the custom Unpack, thanks @urso!

PR #7468, opened 2018-06-28
Clarify autodiscover docs in relation to startup process

exekias:
Autodiscover reacts to already existing containers first, then it
starts listening for changes of that initial state.

### DIFF
--- a/libbeat/docs/shared-autodiscover.asciidoc
+++ b/libbeat/docs/shared-autodiscover.asciidoc
@@ -15,9 +15,8 @@ Autodiscover providers work by watching for events on the system and translating
 events with a common format. When you configure the provider, you can use fields from the autodiscover event to set
 conditions that, when met, launch specific configurations.
 
-On start, {beatname_uc} will scan all the already existing containers and launch the proper configs for them. Then
-it will watch for new start/stop events. This ensures you don't need to worry about state, but only define your desired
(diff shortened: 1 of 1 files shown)


exekias:
@andrewkroh not sure this is the best explanation, I open to ideas to improve it :innocent:

andrewkroh:
Relates: https://discuss.elastic.co/t/beats-and-docker-autodiscover-only-works-with-container-start-stop/137853/

libbeat/docs/shared-autodiscover.asciidoc
```diff
@@ -15,6 +15,10 @@ Autodiscover providers work by watching for events on the system and translating
 events with a common format. When you configure the provider, you can use fields from the autodiscover event to set
 conditions that, when met, launch specific configurations.
 
+On start, {beatname_uc} will scan all the already existing containers and launch the proper configs for them. Then
```
dedemorton: I think you can probably just say "scan existing containers" (the phrase "all the already existing containers" seems redundant).

exekias:
Thanks for the reviews! @dedemorton I just pushed your requested change

dedemorton:
@exekias I think we should backport htis, no? If so, just add the label needs_backport, and I'll pick it up when I backport other PRs.

PR #7469, opened 2018-06-28

andrewkroh:
Take 2 (this time against 6.x)

If 6.x is going to the same as master, but with a different set of versions used several places, then maybe it would be safer/cleaner to just reset 6.x to the master commit and change a few versions/branch names + do a make update.


ruflin:
@andrewkroh Merged it for now but agree that we potentially should just reset 6.x to master and apply changes again to reduce the risk.

@elastic/apm-server In case you want to update your 6.x branch.

PR #7470, opened 2018-06-28
Automatically enrich Kubernetes module events

### LINKED ISSUE #7148: Include labels and annotations in all Kubernetes metricsets
Metricbeat doesn't include object labels and annotations on Kubernetes metricsets. As of today It's possible to enrich `pod`, `container`, `state_pod` and `state_container` events using `add_kubernetes_metadata`, but it would be nice to automatically add these fields to all object kinds, without the processor.

We should implement a metadata retrieval mechanism (or watcher) for all object kinds, and use a configuration similar to `add_kubernetes_metadata` to whitelist labels and annotations.

As a result, all kubernetes metricsets should include `kubernetes.labels` and `kubernetes.annotations` if the user wants them.

exekias:
This PR adds automatically enriching of `kubernetes` module metricsets. It will behave as `add_kubernetes_metadata`, but enrich all events coming out of this module by default. This will not only add labels annotations to Pods but any other Resource, like nodes, containers, deployments...

It will be on by default, some configurations are allowed:

```
- module: kubernetes
  metricsets:
    - pod
    - node
    ...
  in_cluster: false
  kube_config: /home/exekias/.kube/config
  host: minikube
  include_labels:
    - tier
  include_annotations:
    - ...
```

Closes #7148

TODO:
- [x] container
- [x] state_container
- [x] state_deployment
- [x] state_statefulset
- [x] state_replicaset
- [x] update docs
- [x] add more tests
- [x] extend metricbeat service account permissions


metricbeat/docs/fields.asciidoc
```diff
@@ -6890,12 +6890,12 @@ Kubernetes pod name
 
 --
 
-*`kubernetes.pod.uid`*::
+*`kubernetes.uid`*::
```
jsoriano: Breaking change? should we keep both fields till 7.0?

libbeat/common/kubernetes/metadata.go
```diff
@@ -26,21 +26,27 @@ import (
...
 
-type metaGenerator struct {
+// MetaGeneratorConfig settings
+type MetaGeneratorConfig struct {
 	IncludeLabels          []string `config:"include_labels"`
 	ExcludeLabels          []string `config:"exclude_labels"`
 	IncludeAnnotations     []string `config:"include_annotations"`
-	IncludePodUID          bool     `config:"include_pod_uid"`
+	IncludeUID             bool     `config:"include_uid"`
 	IncludeCreatorMetadata bool     `config:"include_creator_metadata"`
```
jsoriano: Are these config settings documented?

metricbeat/module/kubernetes/node/node.go
```diff
@@ -82,5 +89,7 @@ func (m *MetricSet) Fetch() (common.MapStr, error) {
 		return nil, err
 	}
 
+	m.enricher.Enrich([]common.MapStr{event})
+
```
jsoriano: With current abstractions metric sets have basically the same `Fetch` (and `New`) implementation for all resources, maybe we could have an only metricset for them, parametrized with the eventMapper and the enricher.

metricbeat/module/kubernetes/util/kubernetes.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+
+	if !m.watcherStarted {
+		err := m.watcher.Start()
+		if err != nil {
+			logp.Warn("Error starting Kubernetes watcher: %s", err)
+		}
+		m.watcherStarted = true
+	}
+}
```
jsoriano: And a `Stop()` to release the watcher resources?

metricbeat/module/kubernetes/util/kubernetes.go
```diff
@@ -0,0 +1,243 @@
...
+	"sync"
+	"time"
+
+	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/common/kubernetes"
+	"github.com/elastic/beats/libbeat/logp"
+	"github.com/elastic/beats/metricbeat/mb"
+)
+
+var nilEnricher = &enricher{}
```
jsoriano: We could define a nilEnricher type that implements the interface with empty methods, to avoid needing the `if e == nilEnricher` in all methods of `enricher`.

metricbeat/module/kubernetes/util/kubernetes.go
```diff
@@ -0,0 +1,243 @@
...
+			for _, container := range append(pod.GetSpec().GetContainers(), pod.GetSpec().GetInitContainers()...) {
+				id := join(r.GetMetadata().GetNamespace(), r.GetMetadata().GetName(), container.GetName())
+				m[id] = meta
+			}
+		},
+		// delete
+		func(m map[string]common.MapStr, r kubernetes.Resource) {
+			pod := r.(*kubernetes.Pod)
+			for _, container := range append(pod.GetSpec().GetContainers(), pod.GetSpec().GetInitContainers()...) {
+				id := join(r.GetMetadata().GetNamespace(), r.GetMetadata().GetName(), container.GetName())
```
jsoriano: You could also have a struct type for the ids, so the map is `map[resourceID]common.MapStr`, and `id := resourceID{r.GetMetadata().GetNamespace(), ...`.
I have had bad experiences in the past building map keys by joining strings :slightly_smiling_face:

metricbeat/docs/fields.asciidoc
exekias: `kubernetes.pod.uid` is still unreleased, so this change should be ok

libbeat/common/kubernetes/metadata.go
exekias: No, I struggle a bit with these, I wonder if that means we should just hardcode them in. They were kept just in case someone wants to disable them, but I'm not sure

metricbeat/module/kubernetes/node/node.go
exekias: Yep, I had the same feeling, I think it should be done in a follow up PR

metricbeat/docs/fields.asciidoc
jsoriano: Oh ok :)

libbeat/common/kubernetes/metadata.go
jsoriano: Is it currently disabled by default?

As this doesn't look so costly to collect, maybe we can just add them, and they can be removed with `drop_fields`.

metricbeat/module/kubernetes/util/kubernetes.go
exekias: Added, thank you for bringing this in, it was in my TODO :+1:

metricbeat/module/kubernetes/util/kubernetes.go
exekias: pushed

metricbeat/module/kubernetes/util/kubernetes.go
exekias: I would say in general I can do the same mistakes with both approaches, `join` method is syntactically similar to your `resourceID`, in both cases I can pass parameters in the wrong order.

exekias:
Thank you for the review @jsoriano!! It should be ready for a second pass. I'll wait for that before rebasing master

libbeat/common/kubernetes/metadata.go
exekias: yes, so far I have left them undocumented to see how much pushback they get once released, I can leave a note here to say we want to remove these settings

exekias:
:+1: rebased master

libbeat/common/kubernetes/metadata_test.go
```diff
@@ -70,12 +71,12 @@ func TestPodMetadataDeDot(t *testing.T) {
 				},
 			},
 			meta: common.MapStr{
-				"pod":       common.MapStr{"name": "", "uid": "005f3b90-4b9d-12f8-acf0-31020a840133"},
-				"namespace": "",
-				"node":      common.MapStr{"name": "test"},
-				"labels":    common.MapStr{"a": common.MapStr{"value": "bar", "key": "foo"}},
+				"pod":    common.MapStr{"name": ""},
+				"uid":    "005f3b90-4b9d-12f8-acf0-31020a840133",
```
ruflin: As `uid` has be already around before, we should not change anything here so this is only a note from my side.

I often use `id` instead of `uuid` or `uid` for consistency. And as long as there is only one `id` I expect it to be unique.

metricbeat/docs/fields.asciidoc
ruflin: Oh, it's not released yet, then my comment above may apply :-)

ruflin: For my education: This is unique for 1 kubernetes "cluster". Is there also such a thing as a unique pod it?

metricbeat/module/kubernetes/container/container.go
```diff
@@ -83,5 +87,13 @@ func (m *MetricSet) Fetch() ([]common.MapStr, error) {
 		return nil, err
 	}
 
+	m.enricher.Enrich(events)
+
 	return events, nil
 }
+
+// Close stops this metricset
+func (m *MetricSet) Close() error {
```
ruflin: Did you test if Close is called?

metricbeat/module/kubernetes/node/node.go
ruflin: Perhaps time to also convert it to v2 API.

metricbeat/module/kubernetes/util/kubernetes.go
```diff
@@ -0,0 +1,249 @@
...
+
+	// Watch objects in the node only
+	if nodeScope {
+		options.Node = kubernetes.DiscoverKubernetesNode(config.Host, config.InCluster, client)
+	}
+
+	return kubernetes.NewWatcher(client, resource, options)
+}
+
+// NewResourceMetadataEnricher returns a Enricher configured for kubernetes resource events
```
ruflin: Nit: `an Enricher`

metricbeat/module/kubernetes/util/kubernetes.go
```diff
@@ -0,0 +1,249 @@
...
+func NewResourceMetadataEnricher(
+	base mb.BaseMetricSet,
+	resource kubernetes.Resource,
+	nodeScope bool) Enricher {
+
+	watcher, err := GetWatcher(base, resource, nodeScope)
+	if err != nil {
+		logp.Warn("Error initializing Kubernetes metadata enricher: %s", err)
```
ruflin: I would prefer not to use `Warn` but either `Info` or `Error`. Seems more like `Info` in this case as even if the enricher fails, the metricset still starts?

metricbeat/module/kubernetes/util/kubernetes.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+
+	if watcher == nil {
+		logp.Info("Kubernetes metricset enriching is disabled")
+		return &nilEnricher{}
+	}
+
+	metaConfig := kubernetes.MetaGeneratorConfig{}
+	if err := base.Module().UnpackConfig(&metaConfig); err != nil {
+		logp.Warn("Error initializing Kubernetes metadata enricher: %s", err)
```
ruflin: See comment above.

metricbeat/module/kubernetes/util/kubernetes.go
```diff
@@ -0,0 +1,249 @@
...
+}
+
+// NewContainerMetadataEnricher returns a Enricher configured for container events
+func NewContainerMetadataEnricher(
+	base mb.BaseMetricSet,
+	nodeScope bool) Enricher {
+
+	watcher, err := GetWatcher(base, &kubernetes.Pod{}, nodeScope)
+	if err != nil {
+		logp.Warn("Error initializing Kubernetes metadata enricher: %s", err)
```
ruflin: See comment above, applies also further below.

metricbeat/modules.d/kubernetes.yml.disabled
```diff
@@ -28,6 +37,8 @@
 #    - state_container
 #  period: 10s
 #  hosts: ["kube-state-metrics:8080"]
+#  add_metadata: true
```
ruflin: Not introduced in this PR, but I think commenting with `#` should be smae as first part of config (line 16 etc).

libbeat/common/kubernetes/metadata_test.go
exekias: `UID` is a well-known field in Kubernetes, changing it to `id` could confuse some users IMO

ruflin: ok

metricbeat/module/kubernetes/container/container.go
exekias: Yes, surprisingly it's called :smiley_cat: here: https://github.com/elastic/beats/blob/master/metricbeat/mb/module/wrapper.go#L132

metricbeat/module/kubernetes/util/kubernetes.go
ruflin: We tried to get rid of all Warn messages in the past. The reason is that for Warn it's not clear if users have to take again or not. It sounds like in the above case they have to take again so probably move it to `error`.

metricbeat/docs/fields.asciidoc
ruflin: Can there be cases where more then one of the id's from 2 different resources is in the same event? For example having the `pod.uid` and the `kubernetes.uid` which describes the overall cluster in one event? If yes, I would probably put each `uid` under it's resource. Also correlation wise it probably only makes sense to correlate `pod.uid` but not correlated it with any other resources?

exekias: That case can happen, since https://github.com/elastic/beats/pull/7231

But same would apply to labels/annotations. There is always a main resource for the event, and we put the labels & annotations for that one. Would it make sense to move those under `pod.labels`?

ruflin: I so far assume `labels` is a more generic thing and similar labels apply to different resources. A query could be "give me all resources with label `a`". Based on this assumption I think the current handling of labels and tags make still sense. But if the labels don't overlap like in the case of the `uid` the above would seem to make more sense.

exekias: Pushed b6d8f7628 to move it back to `pod.uid`

exekias:
@ruflin let me know if this one is good to go, I can rebase again once reviewed

metricbeat/module/kubernetes/container/container.go
```diff
@@ -66,13 +67,16 @@ func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
...
 		http:          http,
+		enricher:      util.NewContainerMetadataEnricher(base, true),
 	}, nil
 }
 
 // Fetch methods implements the data gathering and data conversion to the right format
 // It returns the event which is then forward to the output. In case of an error, a
 // descriptive error must be returned.
 func (m *MetricSet) Fetch() ([]common.MapStr, error) {
+	m.enricher.Start()
```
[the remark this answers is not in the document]
exekias: Moving it to new is not probably a bad option, some processes instantiate a module without launching it. I will add a mutex for the variable

exekias:
Should be ready for another look

metricbeat/module/kubernetes/container/container.go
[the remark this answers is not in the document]
ruflin: What modules are these?

PR #7471, opened 2018-06-29
Recommend using beat.hostname to fix visulations that use host

dedemorton:
Adds clarification about fixing visualizations so they work with older indices that use host and newer indices that use host.name.

I wonder if this info applies to other use cases, tho. If they are removing host, I kind of assume they don't have visualizations that rely on it, but I wonder if users need to be reminded? If so, I can move this to a different section and point to it from all relevant use cases where `host` doesn't exist.

### DIFF
--- a/libbeat/docs/breaking.asciidoc
+++ b/libbeat/docs/breaking.asciidoc
@@ -56,21 +56,24 @@ In this use case, you load the versioned template manually and use the Beats
 versioned index pattern, `%{[@metadata][beat]}-%{[@metadata][version]}-%{+YYYY.MM.dd}`,
 as recommended in the {logstash-ref}/plugins-inputs-beats.html[Beats input
 plugin] documentation. This results in a `host` field in 6.2 indices, and a
-`host.name` field in 6.3 indices. There are no mapping conflicts. Cross index
-searches in Elasticsearch still work as expected. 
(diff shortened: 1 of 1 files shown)


libbeat/docs/breaking.asciidoc
```diff
@@ -56,21 +56,24 @@ In this use case, you load the versioned template manually and use the Beats
...
 
-If you have multiple visualizations in Kibana that reference the `host` field,
-you can export the objects by using the Kibana API or UI, change any aggregation
-references to use `host.name` instead of `host`, and then re-import the
-objects into Kibana.
+If you have multiple visualizations in Kibana that reference the `host` field:
 
-
-//TODO: Add links to the Kibana docs for exporting/importing objects.
+. Export the objects. You can use the Kibana API, or in the Kibana UI, go to
```
dedemorton: I added general steps here because I couldn't find any documentation in the Kibana Reference that tells users how to export/modify/import objects.

ppf2: I think we can link to these 2 places:

https://www.elastic.co/guide/en/kibana/6.3/api.html for the API
https://www.elastic.co/guide/en/kibana/6.3/managing-saved-objects.html which talks about Export of the objects

@alexfrancoeur

PR #7474, opened 2018-06-29
Deprecation fileset for Elasticsearch filebeat module

immon:


filebeat/docs/fields.asciidoc
```diff
@@ -800,6 +800,22 @@ elasticsearch Module
 
 
 
+[float]
+== deprecation fields
+
+
+
+
+*`elasticsearch.deprecation.component`*::
```
ruflin: Is this component the same used across all elasticsearch logs? I wonder if we should have a generic field for it if yes.

immon: yes, exactly. it is the same.
Deprecation fileset is very similar to server fileset. We could merge those two. We still probably want to mark one logfiles as server log and the other as deprecation log.

ruflin: Does Elasticsearch create a log file for each? If yes, we should keep 2 filesets.

For sharing code: No easy way to do this yet but I can look into it if we have more similar cases in the future.

Any suggestion on how we should call the field? Just `elasticsearch.component`. I feel like there should be something before `component`. Is it `server`?

PR #7476, opened 2018-06-29
Overwrite feature

ph:
Allow to Overwrite an existing feature in the registry and add package
level function to deal bundle that need to overwrite feature.

This is require by the feature defined in the `add_kubernetes_metadata`
processors, the core ships with a set of default options for the feature
but a specific beats can decide to provide another set of default options.

### DIFF
--- a/libbeat/feature/feature.go
+++ b/libbeat/feature/feature.go
@@ -101,7 +101,41 @@ func New(namespace, name string, factory interface{}, stability Stability) *Feat
 // RegisterBundle registers a bundle of features.
 func RegisterBundle(bundle *Bundle) error {
 	for _, f := range bundle.Features() {
-		Registry.Register(f)
+		err := Registry.Register(f)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// MustRegisterBundle register a new bundle and panic on error.
+func MustRegisterBundle(bundle *Bundle) error {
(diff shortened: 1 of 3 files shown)


libbeat/feature/feature.go
```diff
@@ -101,7 +101,41 @@ func New(namespace, name string, factory interface{}, stability Stability) *Feat
...
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// MustOverwriteBundle register a bundle of feature, replace any existing feature with a new
+// implementation and panic on error.
+func MustOverwriteBundle(bundle *Bundle) error {
```
vjsamuel: this is such a nice to have :)

libbeat/feature/feature.go
```diff
@@ -101,7 +101,41 @@ func New(namespace, name string, factory interface{}, stability Stability) *Feat
...
+		err := Registry.Register(f)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// MustRegisterBundle register a new bundle and panic on error.
+func MustRegisterBundle(bundle *Bundle) error {
```
exekias: This function should not return anything, as it either works or panics

libbeat/feature/feature.go
exekias: Same comment applies here

PR #7478, opened 2018-06-29
Add links plus minor edits

dedemorton:
Added links plus a few edits (because I can't resist). The Kibana docs are out of date, so I've opened a PR to get that fixed, too.

Normally I wouldn't document Kibana steps in the Beats docs, but I think it's OK here since this topic won't live indefinitely.

### DIFF
--- a/libbeat/docs/breaking.asciidoc
+++ b/libbeat/docs/breaking.asciidoc
@@ -67,13 +67,21 @@ If you searched for the `host` field previously, modify your searches to use the
 and contains the same information as `host.name`. Using this field ensures that
 your queries and aggregations will work as expected in earlier releases and 6.3.
 
-If you have multiple visualizations in Kibana that reference the `host` field:
+If you have multiple visualizations in Kibana that reference the `host` field,
+export the objects, modify them by changing `host` to `beat.hostname`, and then
+re-import them into Kibana. You can use the Kibana UI or API.
 
-. Export the objects. You can use the Kibana API, or in the Kibana UI, go to
-*Management > Kibana > Saved Objects* to export the objects. 
+To use the Kibana UI:
+
+. In Kibana, go to *Management > Kibana > Saved Objects* and export the objects. 
 . In the exported JSON file, change `host` to `beat.hostname`.
-. Import the modified objects into Kibana. You can use the Kibana API, or in the
-Kibana UI, go to *Saved Objects* to import the objects.
+. In Kibana, go to *Saved Objects* and import the modified objects. 
+
+For more information, see:
+
+* {kibana-ref}/saved-objects-api.html[Kibana Saved Objects API]
+* {kibana-ref}/managing-saved-objects.html[Managing Saved Searches, Visualizations, and Dashboards]
+
 
 [[custom-template-non-versioned-indices]]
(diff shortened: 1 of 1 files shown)


libbeat/docs/breaking.asciidoc
```diff
@@ -67,13 +67,21 @@ If you searched for the `host` field previously, modify your searches to use the
...
 your queries and aggregations will work as expected in earlier releases and 6.3.
 
-If you have multiple visualizations in Kibana that reference the `host` field:
+If you have multiple visualizations in Kibana that reference the `host` field,
+export the objects, modify them by changing `host` to `beat.hostname`, and then
+re-import them into Kibana. You can use the Kibana UI or API.
 
-. Export the objects. You can use the Kibana API, or in the Kibana UI, go to
-*Management > Kibana > Saved Objects* to export the objects. 
+To use the Kibana UI:
```
ppf2: I am thinking we can remove this "to use the Kibana UI" section, so after "..... You can use the Kibana UI or API." just have the "For more information, see:" section to follow it.  Reason being that more advanced users will be able to figure out "modify them by changing `host` to `beat.hostname`" and what specific pieces to replace in the JSON.  I would rather not see some users do a global search/replace which can potentially break some Kibana objects if the string host appears elsewhere that is used for something else, etc..  The Kibana links will be a good reference though. thx!

dedemorton: @ppf2 I sort of agree, but found the Kibana docs to be lacking. In this situation, I don't think we can rely on users being advanced.  As a relatively novice Kibana user myself, I had to do a bit of digging to figure out how to export and edit the visualizations. I didn't really find what I needed when I searched the docs, so the links do help, but the docs about managing saved objects are really out of date and don't provide much context as to why users might want to export and then re-import objets . I'm OK with removing this section because I've opened an issue against the Kibana docs to get the info fixed, but it won't be fixed for another week or so. For the breaking changes, I thought it might be worth spelling out the details...at least until the Kibana docs get updated. We could tweak the language a bit. WDYT?

ppf2: The thing is, if our goal is to spell out the details, then we will need to put step by step instructions including example json syntax and everything, and API calls from Kibana in the Beats documentation.  So I would rather see us addressing the lack of documentation in this area directly in the Kibana guides so that Beats can cross reference them (thanks for filing the doc issue against Kibana):

Users basically have these options:
- They can always use the Kibana UI to edit their objects to make the changes like how they build the visualizations/dashboards in the first place (novice users should be able to perform this)
- If they have a large number of objects to update and prefers to batch update/automate this process:
  - Use Kibana management's export feature to export the objects to json, make the direct json modification, and import them back in, or
  - Use the Kibana API to export the objects to json, make the direct json modification, and import them back in

I think it is fine to list these options and the Kibana reference links that we have today until the Kibana docs get updated with better information.

ruflin:
@dedemorton @ppf2 Based on your comments above, should we merge it as is and Dede will follow up as soon as the KB docs are in place?

dedemorton:
yup, I will backport it. Weird...I thought I had the label on there. arg.

PR #7483, opened 2018-07-01
Update gopsutil to 2.18.06

jsoriano:


jsoriano:

jsoriano:
@ruflin @andrewkroh I have just seen that I overlooked the use of `HOST_*` environment variables, I have opened a new PR to fix this https://github.com/shirou/gopsutil/pull/544.
In any case it wouldn't be a regression because previous implementation with `udevadmn` wouldn't also make use of these environment variables.

PR #7484, opened 2018-07-01
Add grok patterns for CMS Initial Mark and Final Remark GC phases

inqueue:
CMS Initial Mark and Final Remark are stop the world GC collection phases. This PR adds field support for these phases to the GC fileset of the Filebeat Elasticsearch module.


danielmitterdorfer:
> as such it should be used for matching in the pattern?

I meant: it should **not** be used for matching the pattern.

filebeat/module/elasticsearch/gc/_meta/fields.yml
```diff
@@ -1,20 +1,110 @@
...
+    - name: phase
+      type: group
+      description: >
+        Fields specific to GC phase.
+      fields:
+        - name: name
+          type: keyword
+          description: >
+            Name of the GC collection phase.
+        - name: duration_sec
```
ruflin: We should follow the naming convention here: https://www.elastic.co/guide/en/beats/devguide/current/event-conventions.html

You can add here also `format: duration` so Kibana will not it's a duration.

filebeat/module/elasticsearch/gc/_meta/fields.yml
```diff
@@ -1,20 +1,110 @@
...
+          type: float
+          description: >
+            Collection phase duration according to the Java virtual machine.
+        - name: scrub_symbol_table_time_sec
```
ruflin: Same comment as above and also for the following fields.

General note: I would prefer not to have sec and float but instead ms and a long. But not sure if we can make that happen here. 

What should be the accuracy of these fields?

filebeat/module/elasticsearch/gc/_meta/fields.yml
```diff
@@ -1,20 +1,110 @@
...
     - name: tags
       type: keyword
       description: >
         GC logging tags.
+    - name: heap
+      type: group
+      description: >
+        Heap allocation and total size.
+      fields:
+        - name: size_kb
```
ruflin: `format: bytes` can be added for Kibana to detect bytes values. 

Normally we ingest always bytes. Not sure if we can do some conversion on ingest time.

filebeat/module/elasticsearch/gc/ingest/pipeline.json
```diff
@@ -5,13 +5,16 @@
...
+          "(?:%{JVM8HEADER}|%{JVM9HEADER}) \\[GC \\(%{DATA:elasticsearch.gc.phase.name}\\) \\[YG occupancy: %{BASE10NUM:elasticsearch.gc.young_gen.used_kb} K \\(%{BASE10NUM:elasticsearch.gc.young_gen.size_kb} K\\)\\]%{BASE10NUM}: \\[Rescan \\(parallel\\) , %{BASE10NUM:elasticsearch.gc.phase.parallel_rescan_time_sec} secs\\]%{BASE10NUM}: \\[weak refs processing, %{BASE10NUM:elasticsearch.gc.phase.weak_refs_processing_time_sec} secs\\]%{BASE10NUM}: \\[class unloading, %{BASE10NUM:elasticsearch.gc.phase.class_unload_time_sec} secs\\]%{BASE10NUM}: \\[scrub symbol table, %{BASE10NUM:elasticsearch.gc.phase.scrub_symbol_table_time_sec} secs\\]%{BASE10NUM}: \\[scrub string table, %{BASE10NUM:elasticsearch.gc.phase.scrub_string_table_time_sec} secs\\]\\[1 CMS-remark: %{BASE10NUM:elasticsearch.gc.old_gen.used_kb}K\\(%{BASE10NUM:elasticsearch.gc.old_gen.size_kb}K\\)\\] %{BASE10NUM:elasticsearch.gc.heap.used_kb}K\\(%{BASE10NUM:elasticsearch.gc.heap.size_kb}K\\), %{BASE10NUM:elasticsearch.gc.phase.duration_sec} secs\\] %{PROCTIME}",
```
ruflin: This pattern looks pretty dissect friendly: https://www.elastic.co/guide/en/beats/filebeat/master/dissect.html This would be more efficient I think.

We could also do 90% with dissect and the missing parts with grok on the ingest side.

filebeat/module/elasticsearch/gc/test/test.log-expected.json
```diff
@@ -1,15 +1,62 @@
...
+        "elasticsearch.gc.phase.cpu_time.real_sec": "0.00", 
+        "elasticsearch.gc.phase.cpu_time.sys_sec": "0.00", 
+        "elasticsearch.gc.phase.cpu_time.user_sec": "0.01", 
+        "elasticsearch.gc.phase.duration_sec": "0.0021716", 
+        "elasticsearch.gc.phase.name": "CMS Initial Mark", 
         "fileset.module": "elasticsearch", 
         "fileset.name": "gc", 
         "input.type": "log", 
-        "message": "2018-06-11T01:53:11.382+0000: 1396138.752: Total time for which application threads were stopped: 0.0083760 seconds, Stopping threads took: 0.0000702 seconds\n\n", 
+        "message": "2018-03-03T19:37:06.157+0500: 14597.826: [GC (CMS Initial Mark) [1 CMS-initial-mark: 131804K(174784K)] 142444K(253440K), 0.0021716 secs] [Times: user=0.01 sys=0.00, real=0.00 secs]",
```
ruflin: What is the part we want to have here as part of the message?

inqueue:
@danielmitterdorfer `%{JVM9HEADER}` should have been removed; these patterns are only designed to work with JDK 8 only. JDK 9+ support must be added separately.

The use of `1 CMS-initial-mark` was an assumption based on various logs from test and production clusters. Since the level can change, we can do some numeric matching instead.

> You are assuming that the rescan phase is parallel when it can also be non-parallel
I will account for this in my next update.

danielmitterdorfer:
> @danielmitterdorfer Any chance to provide us with such log files? This would make it pretty easy for us to add them to the test suite.

I provided @inqueue with a few pointers on how this could be tackled.

inqueue:
@ruflin @danielmitterdorfer I removed the Java 9 header from the most recent iteration. I think we might be good to merge, though there is still work to do on fields, etc. We can take rescan into account in a future iteration because that too will require a field name change.

ruflin:
I'm good with moving forward here and merging. @danielmitterdorfer WDYT?

PR #7485, opened 2018-07-02
Disable labels dedotting in default docker module configuration

jsoriano:
Default docker module configuration made metricbeat to replace dots
with underscores in labels. This conflicted with `add_docker_metadata`
that also adds labels, but without replacing dots.

This makes docker module and `add_docker_metadata` processor to have a
consistent behaviour, but there can be still cases where type mapping
conflicts appear. For these cases there are some options using the
`rename` or `drop_fields` processors.

### DIFF
--- a/metricbeat/docs/modules/docker.asciidoc
+++ b/metricbeat/docs/modules/docker.asciidoc
@@ -42,8 +42,8 @@ metricbeat.modules:
   period: 10s
   enabled: true
 
-  # Replace dots in labels with `_`. Set to false to keep dots
-  labels.dedot: true
+  # If set to true, replace dots in labels with `_`.
+  #labels.dedot: false
 
   # To connect to Docker over TLS you must specify a client and CA certificate.
   #ssl:
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -47,6 +47,7 @@ https://github.com/elastic/beats/compare/v6.2.3...master[Check the HEAD diff]
 - RabbitMQ management plugin path is now configured at the module level instead of having to do it in each of the metricsets. New `management_path_prefix` option should be used now {pull}7074[7074]
 - RabbitMQ node metricset only collects metrics of the instance it connects to, `node.collect: cluster` can be used to collect all nodes as before. {issue}6556[6556] {pull}6971[6971]
 - Change http/server metricset to put events by default under http.server and prefix config options with server.. {pull}7100[7100]
+- Disable dedotting in docker module configuration. This will change the out-of-the-box behaviour, but not the one of already configured instances. {pull}7485[7485]
 
 *Packetbeat*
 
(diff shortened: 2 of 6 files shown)


metricbeat/docs/modules/docker.asciidoc
```diff
@@ -42,8 +42,8 @@ metricbeat.modules:
   period: 10s
   enabled: true
 
-  # Replace dots in labels with `_`. Set to false to keep dots
-  labels.dedot: true
+  # If set to true, replace dots in labels with `_`.
```
ruflin: We should be careful here as this ends up in 6.4 and docker module is in GA. So this change can only happen for 7.0.

jsoriano: Yeah, @exekias and I talked about this, but we still thought that this could happen in 6.4, there are some reasons:
* The behaviour between the docker module and the `add_docker_metadata` [beta] processor is inconsistent with the provided configurations, and produces mapping conflicts. This alone could be considered a bug, as it breaks the initial experience.
* We think that it is better to don't dedot by default. When dedotting, the user cannot use the labels directly to make queries, what is a bit counterintuitive. Also information in the label name is lost (different labels could generate the same field name and it is not always possible to rebuild the original label from the dedotted string).
* An alternative to fix the "inconsistency bug" would be to add dedot capabilities to the processor (that is still beta), but then we'd be adding two changes, one in 6.4, and another one in 7.0, as we'd like the default to be not-dedotting.
* This BC/"fix" changes the provided and reference configurations, it doesn't change the implementation of the module.

On the cons side, not dedotting can still produce mapping conflicts if a source uses labels with a subpath of other source (e.g. `foo.bar` and `foo.bar.baz`), so the user may still need some additional processors to handle these cases.

We can also leave it as is, and point the users having problems to the options they have.

The ideal solution would be to be able to store the labels as they are, or with some kind of escaping for the dots that still allows to make queries in an intuitive way. But I don't know if something like this is possible with the current Stack.

ruflin: The good news for me is that it was uncommented previously so everyone using the default config will not experience any change. I think that is also the reason we had it there to have a simpler migration path in 7.0.

The broken user experience here is definitively a very strong point. I'm in general on board with making the breaking change especially as we have a config to keep the old behaviour. If we go forward, lets make sure we mention it not only in the changelog but also in the breaking changelog and what users need to change and some explanation. 

As the processor is still in beta, could we bring it to GA? I think that would enforce our arguments on why we changed the defaults.

exekias: Sounds like a plan to me :+1:

jsoriano: So I prepare another PR to make `add_docker_metadata` GA? :smile:

jsoriano: #7497

exekias:
@jsoriano let's make sure we put a line in the breaking changes section of release notes

PR #7487, opened 2018-07-03
Add Filebeat modules to packages

andrewkroh:
This adds the Filebeat modules files to packages. They were missed during the refactor.

It also fixes a minor issue where files were not marked as config within .rpm files.

I opened https://github.com/elastic/beats/issues/7488 because this wasn't caught by any of the testing we do post packaging including beats-tester.


ruflin:
Code LGTM. Have not tested it locally but we should after the packaging job did run.

PR #7492, opened 2018-07-03
squid module init setup

BKS1989:


BKS1989:
grok pattern of access logs

filebeat/module/squid/access/_meta/fields.yml
```diff
@@ -0,0 +1,45 @@
+- key: access
+  type: group
+  description: >
+    squid Module
+  fields:
+    - name: duration
```
ruflin: This needs completion.

BKS1989: I have added test log

ruflin: Could you also update the description of the fields.

Looking at your ingest pipeline below it seems it does not put the fields into the same fields as defined here. You must be aware that all the fields should be prefixed by `squid.access` if they don't overlap with some global fields.

Randy-312:
Possible for this to be moved forward?

blakerouse:
@BKS1989 We are in the process of cleanup some old PR's, noticed this hasn't had any activity in a while. Are you still interested in getting this merged? Would you be able to rebase with latest master?

opoplawski:
I have a question about squid parsing status.  I notice that commit 6d0dc6260deda0df046c59d601885335f6c2c071 added a squid module under the x-pack tree.  Would beats still be amenable to an updated version of this pull request that would be freely available?  Or are we stuck with x-pack only now?

PR #7507, opened 2018-07-04
Add process monitor capabilities to flows (#541)

adriansr:
Flows are updated to include information about the local processes that generate them.

When one or both ends of a flow is identified as a local process, this feature will add the following fields:

* _"proc"_: Process name for the server-side of the flow, as set up in packetbeat.procs.monitored.

* _"cmdline"_: Command-line for the local process on the server-side of the flow.

* _"client_proc"_: Process name for the client-side of the flow, as set up in packetbeat.procs.monitored.

* _"client_cmdline"_: Command-line for the local process on the client-side of the flow.

Partially addresses #541

### DIFF
--- a/packetbeat/_meta/fields.yml
+++ b/packetbeat/_meta/fields.yml
@@ -79,9 +79,17 @@
       description: >
         The name of the process that served the transaction.
 
+    - name: cmdline
+      description: >
+        The command-line of the process that served the transaction.
+
     - name: client_proc
       description: >
         The name of the process that initiated the transaction.
+
+    - name: client_cmdline
+      description: >
+        The command-line of the process that initiated the transaction.
 
     - name: release
       description: >
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -306,6 +306,7 @@ https://github.com/elastic/beats/compare/v6.2.3...master[Check the HEAD diff]
 - HTTP publishes an Error event for unmatched requests or responses. {pull}6794[6794]
 - The process monitor now reports the command-line for all processes, under Linux and Windows. {pull}7135[7135]
 - Updated the TLS protocol parser with new cipher suites added to TLS 1.3. {issue}7455[7455]
+- Flows are enriched with process information using the process monitor. {pull}7507[7507]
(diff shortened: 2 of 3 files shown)


adriansr:
That makes complete sense, @ruflin. In this PR I just mimicked the fields names used in the other packetbeat protocols, we should create an issue to rename them all to conform to ECS

webmat:
I'm not terribly familiar with the Beats code base yet, so take this with the appropriate grain of salt :-) 

* Are we looping over all processes on the system for every flow that gets logged?
* Is packetbeat sampling the flows, or handling 100% of them?

These two questions have me a bit worried

adriansr:
Hi @webmat 

Currently, for every logged event that has a local IP as source or destination, the TCP port is looked up in a process cache. Only if not found in this cache, then all processes are scanned for its open ports. It is an expensive operation, usually ranging from 10 ms under Linux and about 100 ms in Windows. 

As processes monitoring is disabled by default, I think it makes sense to make it more precise when it's enabled, at the expense of some overhead. Previously it would only update the process list after a fixed timeout, causing some events to not be enriched with process information, especially the short-lived processes and sockets.

About the flows, yes, it's handling all of them.

webmat:
Got it, works for me if this is disabled by default. 👍

packetbeat/_meta/fields.yml
```diff
@@ -79,10 +79,18 @@
       description: >
         The name of the process that served the transaction.
 
+    - name: cmdline
```
andrewkroh: So these are not new, correct? They were just missing from fields.yml before?

ruflin: Do we have something like the `fields_check` we have for metricbeat in packetbeat to detect these?

adriansr: Yes, those were missing

andrewkroh: @ruflin The packetbeat system tests all appear to be checking that the fields are documented. But there are no tests with live data for the process enrichment feature (the feature is disabled when reading from pcaps since that isn't live).

PR #7508, opened 2018-07-04
Add host reporting to state API endpoint

ruflin:
Calling the `/state` API endpoint the data returned looks as following:

```
{
  "beat": {
    "name": "ruflin"
  },
  "host": {
    "architecture": "x86_64",
    "name": "ruflin",
    "os": {
      "build": "17E202",
      "family": "darwin",
      "platform": "darwin",
      "version": "10.13.4"
    }
  },
  "module": {
    "count": 3,
    "names": [
      "system"
    ]
  },
  "service": {
    "id": "ac7c8804-42da-4aa3-911c-eea5dd9a460b",
    "name": "metricbeat",
    "version": "7.0.0-alpha1"
  }
}
```

This is the same data as is added by the add_host_metadata processor. It is also reported as part of the monitoring endpoint and means it has to be added to the Elasticsearch monitoring template after merging.

The code for the reporter generation and MapStr was moved into the same file to simplify to keep them in sync. An idea would be to have a Visitor for MapStr which would make the additional code obsolete.

### DIFF
--- a/auditbeat/module/auditd/show_linux.go
+++ b/auditbeat/module/auditd/show_linux.go
@@ -1,3 +1,20 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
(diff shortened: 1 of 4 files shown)


ruflin:
@urso WDYT about a MapStr visitor that does not support slices and arrays? 
@brandonmensing FYI

PR #7510, opened 2018-07-04
Update go ucfg

urso:
- Update go-ucfg to 0.6.0 
- Update NewConfigFrom to treat raw string input as YAML contents
- Introduce MustNewConfigFrom, which throws a panic on error

### DIFF
--- a/libbeat/autodiscover/providers/jolokia/config.go
+++ b/libbeat/autodiscover/providers/jolokia/config.go
@@ -33,7 +33,7 @@ var (
 // Config for Jolokia Discovery autodiscover provider
 type Config struct {
 	// List of network interfaces to use for discovery probes
-	Interfaces []InterfaceConfig `config:"interfaces" validate:"nonzero"`
+	Interfaces []InterfaceConfig `config:"interfaces,replace" validate:"nonzero"`
 
 	Builders  []*common.Config        `config:"builders"`
 	Appenders []*common.Config        `config:"appenders"`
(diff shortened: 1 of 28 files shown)


exekias:
Uhm, it seems some test failures are legit, this may be affecting some existing Unpacks

urso:
Checking the error and default config I'd say the behavior is correct. One can use the `replace` field option to fix the test.

Given the test, the 'wanted' behavior is not clear to me. Without any settings, array values would be merge. e.g.

with default config
```
interfaces:
- a: 1
```

and user configs:
```
interfaces:
- b: 2
- c: 3
```

one would get:

```
interfaces:
- a: 1
  b: 2
- c: 3
```

Is this what we want? Or do we rather want the default array be replaced, so we only get the user settings?

In the former (merge-case), the validator should not fail, cause the resulting array is of length 1, due to the default entry has length=1.

libbeat/autodiscover/providers/jolokia/config.go
```diff
@@ -33,7 +33,7 @@ var (
 // Config for Jolokia Discovery autodiscover provider
 type Config struct {
 	// List of network interfaces to use for discovery probes
-	Interfaces []InterfaceConfig `config:"interfaces" validate:"nonzero"`
+	Interfaces []InterfaceConfig `config:"interfaces,replace" validate:"nonzero"`
```
ruflin: @jsoriano Could you check if this is the expected behaviour here?

PR #7511, opened 2018-07-04
Disable colors on tests

exekias:
It seems this is given problems on some terminals, this change
disables any kind of colors on tests


exekias:
@ruflin as I don't have a clear way to reproduce your issue, could you please test if this fixes your issue before merging? I think it should do

PR #7513, opened 2018-07-04
Separate Line into a readers

kvch:
This PR separates processing objects in `Line` which reads and decodes lines read from an input file. It consist of a reader which reads from an `io.Reader` and returns encoded lines. Then a decoder scanner reads the scanner and decodes the line returned. Both of the new structs follow the interfaces we want to introduce.

I also added few more unit tests and removed the E2E test regarding decoding errors. Mostly because we haven't caught decoding errors before correctly. Also, these kind of test are more effective on unit test levels and it's not future-proof to assert for messages provided by a third-party library.

First I wanted to do the decoding first and then scan for new lines. But it lead to a quite complex code which decodes each character one by one and calculates encoded line sizes using a rather complicated function. I would not introduce code which is not straightforward and easy to misunderstand.

Closing https://github.com/elastic/beats/pull/7337 in favor of this PR.

### DIFF
--- a/filebeat/reader/line/line.go
+++ b/filebeat/reader/line/line.go
@@ -124,11 +124,10 @@ func (d *decoderScanner) Scan() ([]byte, int, error) {
 	return transform.Bytes(d.decoder, buf[:n])
 }
 
+// Next returns a new decoded line from the input file
 func (r *Reader) Next() ([]byte, int, error) {
 	// This loop is need in case advance detects an line ending which turns out
 	// not to be one when decoded. If that is the case, reading continues.
-	var buf []byte
-	var n int
 	for {
 		// read next 'potential' line from input buffer/reader
 		buf, n, err := r.in.Scan()
@@ -138,10 +137,8 @@ func (r *Reader) Next() ([]byte, int, error) {
 
 		if buf[len(buf)-1] == '\n' {
 			return buf, n, err
-		} else {
-			logp.Debug("line", "Line ending char found which wasn't one: %s", buf[len(buf)-1])
 		}
-	}
 
-	return buf, n, nil
+		logp.Debug("line", "Line ending char found which wasn't one: %s", buf[len(buf)-1])
+	}
 }


filebeat/tests/system/test_harvester.py
```diff
@@ -778,47 +778,3 @@ def test_truncate(self):
...
         data = self.get_registry()
         assert len(data) == 1
-
-    def test_decode_error(self):
-        """
-        Tests that in case of a decoding error it is handled gracefully
-        """
-        self.render_config_template(
-            path=os.path.abspath(self.working_dir) + "/log/*",
-            encoding="utf-16be",
```
ruflin: I would keep these tests as they not only tests if the decoding happens correctly but also if the config options work which can't be covered in the unit tests.

ruflin:
Seems like the test failures here is related: https://beats-ci.elastic.co/job/elastic+beats+pull-request+multijob-linux/5346/beat=filebeat,label=ubuntu/testReport/junit/test_harvester/Test/test_exceed_buffer/

filebeat/reader/line/line.go
```diff
@@ -52,143 +64,81 @@ func New(input io.Reader, codec encoding.Encoding, bufferSize int) (*Reader, err
...
+			return 0, err
 		}
+		l.buffer.Append(buf[:n])
 
-		if buf[len(buf)-1] == '\n' {
-			break
-		} else {
-			logp.Debug("line", "Line ending char found which wasn't one: %s", buf[len(buf)-1])
-		}
+		idx = l.buffer.Index(l.nl)
```
urso: what if `idx` is < 0 here? Old solution did loop until newline was detected.

filebeat/reader/line/line.go
```diff
@@ -52,143 +64,81 @@ func New(input io.Reader, codec encoding.Encoding, bufferSize int) (*Reader, err
...
-	// This loop is need in case advance detects an line ending which turns out
-	// not to be one when decoded. If that is the case, reading continues.
-	for {
-		// read next 'potential' line from input buffer/reader
-		err := r.advance()
-		if err != nil {
-			return nil, 0, err
-		}
+func (l *lineReader) Read(buf []byte) (int, error) {
+	idx := l.buffer.Index(l.nl)
```
urso: This `Index` searches by byte sequence. The risk of this is, that the returned index is no actual newline (e.g. if `len(nl) == 1`), but might point in the middle of an encoded symbol. The old code did try to decode up to the `idx` and double check we actually hit a newline symbol.

filebeat/reader/line/line.go
```diff
@@ -52,143 +64,81 @@ func New(input io.Reader, codec encoding.Encoding, bufferSize int) (*Reader, err
...
-
-	// continue scanning input buffer from last position + 1
-	r.inOffset = idx + 1 - sz
-	if r.inOffset < 0 {
-		// fix inOffset if '\n' has encoding > 8bits + firl line has been decoded
-		r.inOffset = 0
-	}
-
-	return err
+	return transform.Bytes(d.decoder, buf[:n])
```
urso: The true scanner is in `d.reader`. This method assume the underlying reader did already decode and 'split' the contents correctly. That is, the `lineReader` is still the scanner and the `decoderScanner` lifts the `lineReader` into a scanner + byte transformation.

filebeat/tests/system/test_harvester.py
urso: This test was actually testing a bug... The error message the test looks for should never have been printed in first place.

ruflin: not sure I fully understand this comment. And I think there are 2 things here:

* Testing the config option
* Testing the conversion. 

What I think we should keep is the first point.

urso: The test didn't do any of these correctly. We tripped on it a few times, wondering why/how it actually works. After fixing the bug, we just couldn't make sense of the test anymore.

PR #7525, opened 2018-07-06
Create (X-Pack Monitoring) stats metricset for Kibana module

ycombinator:
This PR takes the `stats` metricset of the `kibana` Metricbeat module and makes it ship documents to `.monitoring-kibana-6-mb-%{YYYY.MM.DD}` indices, while preserving the current format/mapping expected by docs in these indices. This will ensure that current consumers of the data in these indices, viz. the X-Pack Monitoring UI and the Telemetry shipping module in Kibana, will continue to work as-is.

### To test this PR
The idea is that metricbeat (specifically the kibana/stats metricset with `xpack.enabled: true`) will create exactly the same documents in `.monitoring-kibana-*` indices as Kibana's bulk uploader does today. 

1. Start up Elasticsearch (with the default Basic license is okay)
2. Start up Kibana (with the default Basic license is okay), checked out from `master`.
3. Enable Monitoring in Elasticsearch (via the cluster setting `xpack.monitoring.collection.enabled: true`). You can do this via [Elasticsearch's Cluster Update Settings API](https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-update-settings.html) or by clicking the "Turn on Monitoring" button in the Monitoring UI in Kibana.
4. Let Kibana run for ~20 seconds so a few documents are indexed into `.monitoring-kibana-*`.
5. Stop Kibana
6. From `.monitoring-kibana-*`, retrieve a document each for `type = kibana_stats`
7. Delete the `.monitoring-kibana-*` indices.
8. In `kibana.yml`, set `xpack.monitoring.kibana.collection.enabled: false`. This will ensure that Kibana is no longer directly shipping its monitoring metrics to Elasticsearch.
9. Start up Kibana again, this time checked out from https://github.com/elastic/kibana/pull/20577.
10. Enable the `kibana` module in metricbeat: `./metricbeat modules enable kibana`.
11. In `modules.d/kibana.yml`, add the `stats` metricset and set `xpack.enabled: true`. Concretely, your `modules.d/kibana.yml` should look something like this:

      ```yaml
      - module: kibana
         metricsets:
         - stats
         period: 10s
         hosts: ["localhost:5601"]
         #basepath: ""
         #username: "user"
         #password: "secret"
         xpack.enabled: true
       ```

     Additionally, if your Kibana URLs have a basepath prefix, say `foo`, make sure to un-comment the `basepath` setting in the `kibana` module YAML file and set it's value to `"foo"`.

12. Start metricbeat.
13. Let metricbeat run for ~20 seconds so a few documents are indexed into `.monitoring-kibana-*`.
14. Stop metricbeat
15. From `.monitoring-kibana-*`, retrieve a document each for `type = kibana_stats`.
16. Using a tool such as http://www.jsondiff.com, compare the documents indexed by Kibana's with those indexed by metricbeat. Verify that their structures are identical (same fields, not necessarily same values), except for these known and expected differences:
    1. Only Metricbeat-indexed documents are expected to contain the fields `@timestamp`, `beat`, `host`, and `metricset`. These are "standard" fields added by beats and metricbeat and don't have an adverse impact since they are additive.
    3. Only Kibana-indexed documents are expected to contain the field `source_node`. This field is used for debugging purposes only and not actually consumed by either the Monitoring UI or Telemetry feature in Kibana.
    4. Only Kibana-indexed  documents are expected to contain the field `requests.status_codes`. This field is not currently being used by the Monitoring UI and is, in fact, not being indexed into Elasticsearch with the right mapping. So we're not going to index it going forward (via Metricbeat), until we need it in the UI.

### Out of scope for this PR
These items are related to this PR but will be covered in follow up PRs:
* This PR takes care of `kibana_stats` documents in `.monitoring-kibana-6-*` indices. Similarly, a future PR needs to take care of `kibana_settings` documents.
* Once both `kibana_stats` and `kibana_settings` documents are being indexed by Metricbeat's kibana module (with `xpack.enabled: true`), we need to document how users can switch over from Kibana to Metricbeat for collecting and shipping Kibana monitoring metrics into ES.


metricbeat/module/kibana/stats/data_xpack.go
```diff
@@ -0,0 +1,103 @@
...
+	c "github.com/elastic/beats/libbeat/common/schema/mapstriface"
+	"github.com/elastic/beats/metricbeat/helper"
+	"github.com/elastic/beats/metricbeat/mb"
+)
+
+var (
+	schemaXPackMonitoring = s.Schema{
+		"concurrent_connections": c.Int("concurrent_connections"),
+		"os": c.Dict("os", s.Schema{
+			"load": c.Dict("cpu.load_average", s.Schema{
```
ycombinator: @ruflin @jsoriano The unit test for this file is failing because of the usage of `cpu.load_average` here. The failure  message says:

```
1 error: key `os.cpu.load_average` not found
```

 I think it doesn't like the fact that they key is nested (`cpu.load_average`).

Any ideas how I could achieve what I'm trying to over here? I want the resulting schema to look like this:

```json
"os": {
  "load": {
    "1m": 12345,
    "5m": 23456,
    "15m": 34567
  }
}
```

But the corresponding source data looks like this:

```json
"os": {
  "cpu": {
    "load_average": {
      "1m": 12345,
      "5m": 23456,
      "15m": 34567,
    }
  }
}
```

jsoriano: @ycombinator try to use the nested notation in `c.Float()`, something like this:
```
"load": s.Object{
  "1m": c.Float("cpu.load_average.1m"),
  ... 
}
```

ycombinator: Thanks @jsoriano. I tried that but it didn't work either:

```
=== RUN   TestEventMappingXPack
--- FAIL: TestEventMappingXPack (0.00s)
        Error Trace:    data_xpack_test.go:43
        Error:          Received unexpected error:
                        3 errors: key `os.cpu.load_average.1m` not found; key `os.cpu.load_average.5m` not found; key `os.cpu.load_average.15m` not found
        Test:           TestEventMappingXPack
        Messages:       _meta/test/stats.700.json
        Error Trace:    data_xpack_test.go:44
        Error:          Should be true
        Test:           TestEventMappingXPack
        Messages:       _meta/test/stats.700.json
        Error Trace:    data_xpack_test.go:45
        Error:          Not equal:
                        expected: 0
                        actual  : 1
        Test:           TestEventMappingXPack
        Messages:       _meta/test/stats.700.json
```

jsoriano: I have been taking a look to the behaviour of nested fields in schema and I have opened a PR for discussion (#7583). 

The thing is that for some types (`string` and `interface{}`) the map is converted to `common.MapStr`, and then the value is obtained with `GetValue()`, what works with nested field paths. For the rest of types, as dicts and floats the map is used as a normal map and then it cannot work with nested values.  I'm not sure if this is the intended behaviour... but we should probably unify behaviours.

metricbeat/module/kibana/stats/data.go
```diff
@@ -28,61 +28,143 @@ import (
...
-		"event_loop_delay": c.Float("event_loop_delay"),
 		"process": c.Dict("process", s.Schema{
-			"memory": c.Dict("mem", s.Schema{
-				"heap": s.Object{
-					"max": s.Object{
-						"bytes": c.Int("heap_max_in_bytes"),
+			"event_loop_delay": s.Object{
+				"ms": c.Float("event_loop_delay"),
+			},
+			"memory": c.Dict("memory", s.Schema{
```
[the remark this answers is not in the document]
ycombinator: I think `system` module. I can remove it from the non-xpack-monitoring event here. WDYT?

ycombinator:
> I wonder if we really should add all these metrics or start with a subset and check what we really need. It's easy to add metrics but hard to remove them.

I'm +1 to starting with a minimal set for the reason you mentioned. Question is how to decide what we really need?

If I had to take a guess at this, I'd say we remove the usage section altogether. That would leave things like process metrics (node.js heap, etc.), request counts, response times, and kibana instance metadata.

WDYT?

> Could you also add service.name: kibana to the non xpack event?

Yes, will do. Thanks for catching.

metricbeat/module/kibana/stats/data.go
```diff
@@ -28,61 +28,143 @@ import (
...
+		}),
+		"kibana": KibanaDict,
+		"usage": c.Dict("usage", s.Schema{
+			"kibana":    kibanaUsageDict,
+			"reporting": ReportingUsageDict,
+		}),
+	}
+
+	// RequestsDict defines how to convert the requests field
+	RequestsDict = c.Dict("requests", s.Schema{
```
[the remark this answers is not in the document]
ycombinator: Done in 6357e1090766a02d6e5360c483b3128bb5804f04.

metricbeat/module/kibana/stats/data.go
```diff
@@ -28,61 +28,143 @@ import (
...
+					},
+				}),
 			}),
-			"pid": c.Int("pid"),
 			"uptime": s.Object{
 				"ms": c.Int("uptime_ms"),
 			},
 		}),
+		"requests": RequestsDict,
+		"response_times": c.Dict("response_times", s.Schema{
```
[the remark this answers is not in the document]
ycombinator: Done in 6357e1090766a02d6e5360c483b3128bb5804f04.

metricbeat/module/kibana/stats/data.go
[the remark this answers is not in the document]
ruflin: +1

ruflin:
What to add and what not is the tricky question that bugs us since day one of Metricbeat. The good news is that by now with processors we allow the users to also drop fields they don't need.

For the metrics I would ask the following questions to answer if we should keep them or not:
* If something is wrong with Kibana, will these metrics be valuable?
* Are these metrics of interest for the users when everything is green?

There is a lot of gray area and I suggest just pick the ones that you think are best for now.

ycombinator:
> For the metrics I would ask the following questions to answer if we should keep them or not:
> * If something is wrong with Kibana, will these metrics be valuable?
> * Are these metrics of interest for the users when everything is green?

Yeah, using similar guidelines I made my earlier suggestion:

>  I'd say we remove the usage section altogether. That would leave things like process metrics (node.js heap, etc.), request counts, response times, and kibana instance metadata.

I'll make the necessary changes to this PR. Thanks!

ycombinator:
> Could you generate a new data.json? This should make it easier to comment on the generated structure.

Done in 64903e7.

@ruflin I believe I have now addressed all the points from your review. Please take another 👀 whenever you get a chance. Thanks!

ycombinator:
jenkins, test this

metricbeat/module/kibana/stats/stats.go
```diff
@@ -44,20 +45,36 @@ var (
...
+		cfgwarn.Experimental("The experimental xpack.enabled flag in kibana/stats metricset is enabled.")
+	}
+
 	http, err := helper.NewHTTP(base)
 	if err != nil {
 		return nil, err
 	}
+
+	if config.BasePath != "" {
+		http.AddBasePath(config.BasePath)
```
ycombinator: @ruflin Since Kibana URIs can have an optional basepath, I'm allowing users of the `kibana` module to set it in their config YAML. If it is set, I insert it into the URI using `http.AddBasePath` as shown here. Please see my implementation of the `AddBasePath` method (elsewhere in this PR) as well.

WDYT - is this config setting the right approach? If yes, is the implementation of the `AddBasePath` method alright?

metricbeat/metricbeat.reference.yml
```diff
@@ -351,6 +351,7 @@ metricbeat.modules:
   metricsets: ["status"]
   period: 10s
   hosts: ["localhost:5601"]
+  basepath: ""
```
ruflin: Is the basepath mainly used for development? Would it also work the specify the host as `localhost:5601/abc`

metricbeat/module/kibana/stats/_meta/data.json
```diff
@@ -6,64 +6,54 @@
...
+            "id": "NkfU5AinRnyFnqBD36zhEw"
         }
     },
     "kibana": {
         "stats": {
-            "concurrent_connections": 17,
-            "event_loop_delay": 718.5184001922607,
-            "name": "kibana",
+            "concurrent_connections": 3,
+            "kibana": {
```
ruflin: What if we put this data up one level outside `stats`?

So it would be for example `kibana.host.name`, `kibana.name`, `kibana.id` etc.? I would assume in case we add other metricsets they would reuse this data.

It bugs me that we have `kibana.*.kibana`.

metricbeat/module/kibana/stats/_meta/fields.yml
```diff
@@ -8,118 +8,103 @@
...
-              type: long
-              format: bytes
-              description: >
-                The amount of space occupied in main memory for the process. Includes heap, code segment, and stack.
-            - name: external.bytes
-              type: long
-              format: bytes
-              description: >
-                Memory usage of C++ objects bound to JavaScript objects managed by V8.
-        - name: pid
```
ruflin: I'm tempted to say we should keep the `pid` and put it under `process.pid` for correlation reasons.

metricbeat/module/kibana/stats/_meta/fields.yml
```diff
@@ -8,118 +8,103 @@
...
-          type: long
+        - name: avg.ms
+          type: scaled_float
           description: >
-            Total number of requests sent by the server.
-        - name: disconnects
-          type: long
+            Average response time in milliseconds
+        - name: max.ms
+          type: scaled_float
```
ruflin: Should this be a long?

metricbeat/module/kibana/stats/data_xpack.go
```diff
@@ -0,0 +1,167 @@
...
+		return err
+	}
+
+	kibanaStatsFields, err := schemaXPackMonitoring.Apply(data)
+	if err != nil {
+		r.Error(err)
+		return err
+	}
+
+	process := data["process"].(map[string]interface{})
```
ruflin: We should validate all these type conversions to not have a panic. Also 3 more below.

metricbeat/module/kibana/stats/stats.go
ruflin: In general LGTM. See my comment above if this is only for dev purpose and if yes, if the host config change could work.

metricbeat/module/kibana/_meta/config.yml
```diff
@@ -3,5 +3,6 @@
   #  - status
   period: 10s
   hosts: ["localhost:5601"]
+  #basepath: ""
```
ruflin: Is that a config that non dev users will use? If yes, is it common? If not, lets skip it in the short config.

metricbeat/metricbeat.reference.yml
ycombinator: Basepath gets used in production scenarios too, where users deploy Kibana and also other applications behind the same proxy and need a unique base path to route to each application.

ruflin: Ok, then I think a separate config makes sense. I assume including in the host would also work?

ycombinator: >  I assume including in the host would also work?

You mean something like `hosts: ["localhost:5601/foo"]` in `modules.d/kibana.yml`, right?

Heh, I swear I tried that first and it didn't work at the time, so I went with a separate setting. However, I just tried it again, and it works! So now I think we should get rid of the extra `basepath` setting.

ycombinator: Ugh, spoke too soon. It doesn't work when I add it to the `hosts` setting. 😞 

What seems to happen is that the URL becomes `http://localhost:5601/foo?extended=true` instead of `http://localhost:5601/foo/api/stats?extended=true`.

So now I'm back to thinking we should give `basepath` its own setting.

ycombinator: BTW, even if we decide to have a separate `basepath` setting, I think there's a better implementation for utilizing it. I've made this alternate implementation in 1379dd89acd21cc2e7be23570e20f5c27891d5fa. Let me know what you think of it.

metricbeat/module/kibana/stats/_meta/fields.yml
ycombinator: Added back in 332c83e.

metricbeat/metricbeat.reference.yml
ruflin: I like the idea of having the basepath in our http parser directly as I remember also other modules use it. Not 100% about the name basepath or something like prefix at the same time does not matter much.

@andrewkroh WDYT about having basepath as part of the http parser?

metricbeat/mb/parse/url.go
```diff
@@ -77,8 +78,25 @@ func (b URLHostParserBuilder) Build() mb.HostParser {
...
+		path = strings.Trim(path, "/")
 
-		return ParseURL(host, b.DefaultScheme, user, pass, path, b.QueryParams)
+		t, ok = conf[b.BasePathConfigKey]
+		if ok {
+			basePath, ok = t.(string)
+			if !ok {
+				return mb.HostData{}, errors.Errorf("'%v' config for module %v is not a string", b.BasePathConfigKey, module.Name())
+			}
+		} else {
```
ruflin: I don't think this else is needed as the default for a string is already ""

metricbeat/mb/parse/url.go
```diff
@@ -77,8 +78,25 @@ func (b URLHostParserBuilder) Build() mb.HostParser {
...
+			basePath = ""
+		}
+		// Normalize basepath
+		basePath = strings.Trim(basePath, "/")
+
+		// Combine paths and normalize
+		fullPath := strings.Trim(basePath+"/"+path, "/")
```
ruflin: Want to use path.Join here? https://golang.org/pkg/path/#Join

metricbeat/mb/parse/url.go
```diff
@@ -77,8 +78,25 @@ func (b URLHostParserBuilder) Build() mb.HostParser {
 		} else {
 			path = b.DefaultPath
 		}
+		// Normalize path
+		path = strings.Trim(path, "/")
```
ruflin: I'm never sure if we should try to be smart about / prefix and post fixes or just leave them.

metricbeat/mb/parse/url_test.go
```diff
@@ -114,6 +114,7 @@ func TestURLHostParserBuilder(t *testing.T) {
 		{map[string]interface{}{}, URLHostParserBuilder{DefaultPath: "/default"}, "http://example.com/default"},
 		{map[string]interface{}{"username": "guest"}, URLHostParserBuilder{}, "http://guest@example.com"},
 		{map[string]interface{}{"username": "guest", "password": "secret"}, URLHostParserBuilder{}, "http://guest:secret@example.com"},
+		{map[string]interface{}{"basepath": "/foo"}, URLHostParserBuilder{BasePathConfigKey: "basepath", DefaultPath: "/default"}, "http://example.com/foo/default"},
```
ruflin: If we strip the / before and after lets also add some tests here to see what happens with the different varians.

metricbeat/module/kibana/config.go
```diff
@@ -0,0 +1,32 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package kibana
+
+// Config defines the structure for the Kibana module configuration options
+type Config struct {
+	XPackEnabled bool   `config:"xpack.enabled"`
+	BasePath     string `config:"basepath"`
```
ruflin: BasePath here becomes obsolete if we have it in the parser I assume?

metricbeat/module/kibana/stats/_meta/fields.yml
```diff
@@ -4,122 +4,106 @@
...
+      type: text
+      description: >
+        Kibana instance name
+    - name: index
       type: keyword
       description: >
-        Kibana instance name.
-    - name: uuid
+        Name of Kibana's internal index
+    - name: host
```
ruflin: Can we make this `host.name` to be consistent with ECS?

metricbeat/module/kibana/stats/_meta/fields.yml
```diff
@@ -4,122 +4,106 @@
...
+        Kibana version
+    - name: snapshot
+      type: boolean
+      description: >
+        Whether the Kibana build is a snapshot build
+    - name: status
+      type: keyword
+      description: >
+        Kibana instance's health status
+    - name: cluster_uuid
```
ruflin: In the future this should go under `elasticsearch.cluster.id` (not needed now)

metricbeat/module/kibana/stats/data_xpack.go
```diff
@@ -0,0 +1,195 @@
...
+		return err
+	}
+
+	kibanaStatsFields, err := schemaXPackMonitoring.Apply(data)
+	if err != nil {
+		r.Error(err)
+		return err
+	}
+
+	process, ok := data["process"].(map[string]interface{})
```
ruflin: @jsoriano I think this is an interesting example where we could potentially use "required" fields and skip this part if some required fields are missing. Not part of this PR.

metricbeat/module/kibana/stats/data_xpack.go
```diff
@@ -0,0 +1,195 @@
...
+	if !ok {
+		return reportErrorForMissingField("process", r)
+	}
+	memory, ok := process["memory"].(map[string]interface{})
+	if !ok {
+		return reportErrorForMissingField("process.memory", r)
+	}
+
+	rss, ok := memory["resident_set_size_bytes"].(float64)
```
ruflin: why do you use `float64`  here?

metricbeat/mb/parse/url.go
ycombinator: I added this because, without it, some existing tests were failing. I didn't want to change the tests because they might represent expectations of actual usage (i.e. the existing contract).

metricbeat/module/kibana/stats/_meta/fields.yml
ycombinator: I can easily make the change now itself. Unless there is a reason to hold off until later?

metricbeat/module/kibana/config.go
ycombinator: Yes, good catch. Forgot to remove `BasePath` from this file.

metricbeat/mb/parse/url_test.go
ycombinator: Added in a644e62.

metricbeat/module/kibana/stats/data_xpack.go
ycombinator: Originally I tried to use `int` here but the type assertion failed. The value in the JSON response from the Kibana stats API looks like an integer to me but for some reason the [JSON parser thinks its a float because perhaps it's too large](https://stackoverflow.com/questions/22343083/json-marshaling-with-long-numbers-in-golang-gives-floating-point-number)?.

ycombinator:
@ruflin I've addressed all the comments in your latest round of review. Most of your feedback I fixed up in the code but for some of it I made some comments/questions. Please take a look when you have a chance. Thanks!

metricbeat/module/kibana/stats/data_xpack.go
ruflin: Perhaps you should try `int64`? But I would be surprised if the value exceeds int TBH.

metricbeat/module/kibana/stats/_meta/fields.yml
ruflin: Feel free to do it now.

metricbeat/module/kibana/stats/data_xpack.go
ycombinator: Yeah, it didn't like `int64` either :(.

FWIW, if I do a `fmt.Println(memory["resident_set_size_bytes"])` (right before the type assertion), I get `6.2808064e+07`.

ruflin: Ah, I see. Kibana does report it in this format. float and later converstion to int SGTM. Perhaps make it int64 there.

ycombinator: Done in b3fbd5f.

metricbeat/module/kibana/stats/_meta/data.json
```diff
@@ -4,45 +4,52 @@
...
-    "elasticsearch": {
-        "cluster": {
-            "id": "njqU4EQaTROIDlWPeUMQyw"
-        }
-    },
     "kibana": {
         "stats": {
-            "concurrent_connections": 0,
+            "concurrent_connections": 3,
+            "elasticsearch": {
```
ruflin: This should be `elasticsearch.cluster.id` on the Root level. Same for `process.pid`

ycombinator: Ah doh. Fixing!

metricbeat/module/kibana/stats/_meta/data.json
```diff
@@ -73,6 +61,9 @@
         "name": "stats",
         "rtt": 115
     },
+    "process": {
```
ruflin: Nice

metricbeat/metricbeat.reference.yml
andrewkroh: I would prefer to see if we can change the behavior of the http parser such that if path is included in the `hosts` URL that it still does the right thing. It seems like that was the expected behavior. Fewer config options and an explicit URL sounds simpler to me, but I may not have the whole picture.

> What seems to happen is that the URL becomes http://localhost:5601/foo?extended=true instead of http://localhost:5601/foo/api/stats?extended=true.

ycombinator: That makes sense @andrewkroh. Thanks for weighing in. I'll give this a shot in a new PR this coming week.

ruflin: @ycombinator Could you set it a priority so we get it into 6.4?

ycombinator: Yep, I plan to have a PR up today!

ycombinator: PR is up and I've requested a review from both of you: https://github.com/elastic/beats/pull/7682

PR #7527, opened 2018-07-06
Add `bearer_token_file` paramter to HTTP helper

### LINKED ISSUE #7518: Implement service account token based HTTP authorization
Metricbeat uses HTTP helper to gather metrics from kubelet + many Prometheus endpoints. In order to use the safe port (HTTPS) we need to implement a way to fetch service account token from the metricbeat container. A request like this works:

```
TOKEN="$(cat /var/run/secrets/kubernetes.io/serviceaccount/token)"
curl --cacert /var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt https://localhost:10250/stats/summary -H "Authorization: Bearer $TOKEN"
```

more on service account tokens:
https://kubernetes.io/docs/concepts/storage/volumes/#projected
https://kubernetes.io/docs/reference/access-authn-authz/authentication/

exekias:
This change allows to load bearer tokens from files in modules using
the HTTP helper. This is especially useful for Kubernetes and Prometheus, as some deployments enforce SSL access (like OpenShift):

```
- module: kubernetes
  metricsets:
    - pod
  bearer_token_file: /var/run/secrets/kubernetes.io/serviceaccount/token
  ssl.certificate_authorities:
    - /var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt
```

Closes #7518


exekias:
@vjsamuel, I think this should deprecate the `kubernetes.token` appender

metricbeat/module/prometheus/_meta/config.yml
```diff
@@ -7,3 +7,8 @@
   #namespace: example
   #username: "user"
   #password: "secret"
+
+  # This can be used for service account based authorization:
```
andrewkroh: I think it would be useful to say what it does. Something like 

> It reads the contents of the file once at initialization and then uses the value in an HTTP Authorization header.

ruflin:
Needs a changelog entry?

exekias:
Thanks for the reviews, I have added more docs & changelog entry

metricbeat/module/prometheus/_meta/config.yml
exekias: Makes sense, I updated docs to include this longer explanation

metricbeat/helper/http_test.go
```diff
@@ -0,0 +1,54 @@
+package helper
```
exekias: I realize now this is a new file without license headers, but CI is not complaining. Are we checking that somewhere @andrewkroh?

ruflin: @exekias Was also looking into this this morning. We check but return 0 after the check. Not sure if that was on purpose (probably not).

exekias: In the meanwhile, I've pushed the header to this file

metricbeat/autodiscover/appender/kubernetes/token/token.go
```diff
@@ -41,7 +41,7 @@ type tokenAppender struct {
 // NewTokenAppender creates a token appender that can append a bearer token required to authenticate with
 // protected endpoints
 func NewTokenAppender(cfg *common.Config) (autodiscover.Appender, error) {
-	cfgwarn.Beta("The token appender is beta")
+	cfgwarn.Deprecate("7.0.0", "token appender is deprecated in favor of bearer_token_file config parameter")
```
vjsamuel: Do you suggest moving to `config` appender for use cases that require a bearer token then? I wouldnt do a blanket token path on all configs with the hints builder.

exekias: Yeah, my thought was to use the `config` appender to set `bearer_token_file` wherever you were using the `token` appender before. No need to create new hints for this.
Would that work for you?

vjsamuel: im ok with that. we will do the needful once this is merged.

metricbeat/module/prometheus/_meta/config.yml
ruflin: Should there also be an update to the reference config files? Also applies to the k8s one.

For the paths here for the prometheus config: Should we have k8s as default ones?

exekias: I think Prometheus is mostly used in k8s scenarios, it should not harm other use cases anyway? I've pushed a commit to include it in

exekias:
ups, my bad, should be ok now, sorry

PR #7533, opened 2018-07-06
Improve fields.yml generator of modules

### LINKED ISSUE #7472: [Filebeat] Fields for module fileset are not parsed when Grok pattern define also type of the field
Hi!
When generating the base `field.yml` file from pipeline definition in module's fileset, fields with type definition are ignored or not parsed. They do not appear in `fields.yml` file. All others are just fine.

Example definition: `%{NUMBER:elasticsearch.slowlog.total_hits:int}`
Example command for file generation: `make create-fields MODULE=elasticsearch FILESET=slowlog`

Thanks

kvch:
From now on when a user provides a type hint in an Ingest pipeline, it's added to the generated `fields.yml` instead of guessing.

Closes #7472

### DIFF
--- a/filebeat/scripts/generator/fields/main.go
+++ b/filebeat/scripts/generator/fields/main.go
@@ -33,6 +33,10 @@ import (
 const (
 	pipelinePath  = "%s/module/%s/%s/ingest/pipeline.json"
 	fieldsYmlPath = "%s/module/%s/%s/_meta/fields.yml"
+
+	typeIdx     = 0
+	elementsIdx = 1
+	hintIdx     = 2
 )
 
 var (
@@ -96,19 +100,32 @@ func newFieldYml(name, typeName string, noDoc bool) *fieldYml {
 func newField(lp string) field {
 	lp = lp[1 : len(lp)-1]
 	ee := strings.Split(lp, ":")
-	if 2 != len(ee) && len(ee) != 3 {
+	if !isValidFormat(ee) {
 		return field{}
 	}
 
 	hint := ""
-	if len(ee) == 3 {
-		hint = ee[2]
+	if containsHint(ee) {
+		hint = ee[hintIdx]
 	}
+
 	return field{
-		Type:     ee[0],
-		Elements: strings.Split(ee[1], "."),
+		Type:     ee[typeIdx],
+		Elements: strings.Split(ee[elementsIdx], "."),
 		Hint:     hint,
 	}
+}
+
+// isValidFormat checks if the input can be split correctly
+// 1. if lenght is 2, the format is {type}:{field.elements}
+// 2. if the lenght is 3, the format is {type}:{field.elements}:{hint}
+func isValidFormat(ee []string) bool {
+	return len(ee) == 2 || len(ee) == 3
+}
+
+// the last element is the type hint
+func containsHint(ee []string) bool {
+	return len(ee) == 3
 }
 
 func readPipeline(beatsPath, module, fileset string) (*pipeline, error) {
--- a/filebeat/scripts/generator/fields/main.go
+++ b/filebeat/scripts/generator/fields/main.go
@@ -68,9 +68,9 @@ type pipeline struct {
 }
 
 type field struct {
-	Type     string
-	Elements []string
-	Hint     string
+	Syntax           string
+	SemanticElements []string
+	Type             string
 }
 
 type fieldYml struct {
@@ -98,21 +98,25 @@ func newFieldYml(name, typeName string, noDoc bool) *fieldYml {
 }
 
 func newField(lp string) field {
+	if len(lp) <= 2 {
+		return field{}
+	}
 	lp = lp[1 : len(lp)-1]
+
 	ee := strings.Split(lp, ":")
 	if !isValidFormat(ee) {
 		return field{}
 	}
 
 	hint := ""
-	if containsHint(ee) {
+	if containsType(ee) {
 		hint = ee[hintIdx]
 	}
 
 	return field{
-		Type:     ee[typeIdx],
-		Elements: strings.Split(ee[elementsIdx], "."),
-		Hint:     hint,
+		Syntax:           ee[typeIdx],
+		SemanticElements: strings.Split(ee[elementsIdx], "."),
+		Type:             hint,
 	}
 }
 
@@ -124,7 +128,7 @@ func isValidFormat(ee []string) bool {
 }
 
 // the last element is the type hint
-func containsHint(ee []string) bool {
+func containsType(ee []string) bool {
 	return len(ee) == 3
 }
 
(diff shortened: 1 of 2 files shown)


filebeat/scripts/generator/fields/main.go
```diff
@@ -95,17 +96,18 @@ func newFieldYml(name, typeName string, noDoc bool) *fieldYml {
 func newField(lp string) field {
 	lp = lp[1 : len(lp)-1]
 	ee := strings.Split(lp, ":")
-	if len(ee) != 2 {
-		return field{
-			Type:     ee[0],
-			Elements: nil,
-		}
+	if 2 != len(ee) && len(ee) != 3 {
```
ruflin: Could you add a comment here on why we check for these 2 values? Should make reading the code easier.

filebeat/scripts/generator/fields/main.go
```diff
@@ -95,17 +96,18 @@ func newFieldYml(name, typeName string, noDoc bool) *fieldYml {
 func newField(lp string) field {
 	lp = lp[1 : len(lp)-1]
```
ruflin: This code is fine as long as we use it just as generator on the dev side. But as recently we started to move some code also in the Beats binary we should probably be more careful with validation checks. What happens here in case an empty string is passed to it?

kvch: This function never gets an empty string as a param, because it parses the strings returned by `FindAllString` which does not match for empty lines. But just to be sure, if some tries to use it incorrectly in the future, I added additional error checking.

filebeat/scripts/generator/fields/main.go
kvch: Done.

kvch:
I added additional comments and telling names to functions. Furthermore, I renamed the struct attributes to reflect the notions of the Grok processor.

filebeat/scripts/generator/fields/main.go
ruflin: I think the key here is we can't make assumption about future usage of this method as you stated so checks should be in place to make sure the method never panics.

Seems the checks didn't make it in yet? If I run `newField("")` the code would panic?

filebeat/scripts/generator/fields/main.go
```diff
@@ -95,20 +100,34 @@ func newFieldYml(name, typeName string, noDoc bool) *fieldYml {
...
 	ee := strings.Split(lp, ":")
```
ruflin: What stands `ee` for?

kvch: Well, it's "elements". I don't really have a good name for elements of `SYNTAX:SEMANTIC:TYPE` split at `:`s. I can rename it to `elements`, but I don't think it would help. :(

ruflin: I quite like `elements`, sounds better to me then `ee`

filebeat/scripts/generator/fields/main.go
```diff
@@ -92,23 +97,41 @@ func newFieldYml(name, typeName string, noDoc bool) *fieldYml {
...
-	lp = lp[1 : len(lp)-1]
-	ee := strings.Split(lp, ":")
-	if len(ee) != 2 {
-		return field{
-			Type:     ee[0],
-			Elements: nil,
-		}
+func newField(pattern string) field {
+	if len(pattern) <= 2 {
+		return field{}
```
ruflin: Should we log errors instead of return empty fields? This would make debugging much easier in the future.

filebeat/scripts/generator/fields/main.go
```diff
@@ -92,23 +97,41 @@ func newFieldYml(name, typeName string, noDoc bool) *fieldYml {
...
+	}
+	pattern = pattern[1 : len(pattern)-1]
+
+	elements := strings.Split(pattern, ":")
+	if !isValidFormat(elements) {
+		return field{}
```
ruflin: Return error?

filebeat/scripts/generator/fields/main.go
```diff
@@ -92,23 +97,41 @@ func newFieldYml(name, typeName string, noDoc bool) *fieldYml {
...
-	if len(ee) != 2 {
-		return field{
-			Type:     ee[0],
-			Elements: nil,
-		}
+func newField(pattern string) field {
+	if len(pattern) <= 2 {
+		return field{}
+	}
+	pattern = pattern[1 : len(pattern)-1]
```
ruflin: Alternative here would `strings.Trim("pattern", "{}")`

filebeat/scripts/generator/fields/main.go
```diff
@@ -92,23 +97,41 @@ func newFieldYml(name, typeName string, noDoc bool) *fieldYml {
...
 
+// isValidFormat checks if the input can be split correctly
+// 1. if lenght is 2, the format is {type}:{field.elements}
+// 2. if the lenght is 3, the format is {type}:{field.elements}:{hint}
+func isValidFormat(ee []string) bool {
+	return len(ee) == 2 || len(ee) == 3
+}
+
+// the last element is the type hint
+func containsType(ee []string) bool {
```
ruflin: Should this be `containsHint`? I understand the the hint is for the type but this checks if it contains hint?

filebeat/scripts/generator/fields/main.go
```diff
@@ -92,23 +97,41 @@ func newFieldYml(name, typeName string, noDoc bool) *fieldYml {
...
+		return field{}
+	}
+	pattern = pattern[1 : len(pattern)-1]
+
+	elements := strings.Split(pattern, ":")
+	if !isValidFormat(elements) {
+		return field{}
+	}
+
+	hint := ""
```
ruflin: What about having a function `getType(elements)` that does the below? Would also make it easier to test I think.

filebeat/scripts/generator/fields/main.go
```diff
@@ -92,23 +97,41 @@ func newFieldYml(name, typeName string, noDoc bool) *fieldYml {
...
+		hint = elements[hintIdx]
 	}
 
-	e := strings.Split(ee[1], ".")
 	return field{
-		Type:     ee[0],
-		Elements: e,
+		Syntax:           elements[typeIdx],
+		SemanticElements: strings.Split(elements[elementsIdx], "."),
+		Type:             hint,
```
ruflin: The variable should be called `type` probably, see the above.

filebeat/scripts/generator/fields/main.go
```diff
@@ -33,6 +33,10 @@ import (
 const (
 	pipelinePath  = "%s/module/%s/%s/ingest/pipeline.json"
 	fieldsYmlPath = "%s/module/%s/%s/_meta/fields.yml"
+
+	typeIdx     = 0
```
ruflin: Isn't that the syntax index?

kvch:
I would rather not address all your review comments in this PR as I already have another one specifically for refactoring and making this script ready to be used as a subcommand. I intended this PR to be a quick fix orthogonal to refactoring, so I can rebase my other PR on top of it with minimal conflicts.

So what do you think if we merge this PR and I do the refactoring you requested in https://github.com/elastic/beats/pull/7506?

ruflin:
Merged as we discussed that refactoring of the comments will happen in https://github.com/elastic/beats/pull/7506 or a separate PR.

PR #7537, opened 2018-07-09
Add queue name to state reporting

ruflin:
The queue name is reported as part of the state event. An event looks as following:

```
{
  "beat": {
    "name": "ruflin"
  },
  "host": {
    "architecture": "x86_64",
    "name": "ruflin",
    "os": {
      "build": "17E202",
      "family": "darwin",
      "platform": "darwin",
      "version": "10.13.4"
    }
  },
  "module": {
    "count": 3,
    "names": [
      "system"
    ]
  },
  "output": {
    "name": "elasticsearch"
  },
  "queue": {
    "name": "mem"
  },
  "service": {
    "id": "0e1b1913-73db-4512-b17a-8209fa45eaa4",
    "name": "metricbeat",
    "version": "7.0.0-alpha1"
  }
}
```

### DIFF
--- a/libbeat/publisher/pipeline/module.go
+++ b/libbeat/publisher/pipeline/module.go
@@ -157,6 +157,10 @@ func createQueueBuilder(config common.ConfigNamespace) (func(queue.Eventer) (que
 		queueConfig = common.NewConfig()
 	}
 
+	stateRegistry := monitoring.GetNamespace("state").GetRegistry()
+	queueRegistry := stateRegistry.NewRegistry("queue")
+	monitoring.NewString(queueRegistry, "name").Set(queueType)
+
 	return func(eventer queue.Eventer) (queue.Queue, error) {
 		return queueFactory(eventer, queueConfig)
 	}, nil


libbeat/publisher/pipeline/module.go
```diff
@@ -157,6 +157,10 @@ func createQueueBuilder(config common.ConfigNamespace) (func(queue.Eventer) (que
 		queueConfig = common.NewConfig()
 	}
 
+	stateRegistry := monitoring.GetNamespace("state").GetRegistry()
+	queueRegistry := stateRegistry.NewRegistry("queue")
+	monitoring.NewString(queueRegistry, "name").Set(queueType)
```
urso: The pipeline should not use any globals at all, so one can safely instantiate multiple pipelines via Load. The Load function even gets the metrics registry passed as parameter (optional, can be nil).

ruflin: The registry passed is not the same into which this data has to go.

If there can be multiple queues in the future we would need a variable inside this package to keep track of it like we do for modules. Would that be ok for you?

urso: I understand it's a different regsitry. In this case we should pass a 'telemetry' registry or some kind of callback to the loader. But no globals. It's one of the few packages with no globals in mind from the very beginning.

Me might have other places where we need to pass a telemetry and a metrics registry to a feature/module/plugin. Maybe we want to pass a struct or interface of type:

```
type monitoring interface {
  Telemetry() *monitoring.Registry
  Metrics() *monitoring.Registry
  Logs() logp.Logger
}
```

urso: @ruflin I see some globals have been sneaked in... see #8091 for cleanups + introduction of queue type telemetry

urso: Yes.

ruflin:
Closing in favor of https://github.com/elastic/beats/pull/8091

PR #7541, opened 2018-07-09
Add support for customising the Grok pattern in the nginx module

tsg:


filebeat/module/nginx/_meta/config.reference.yml
```diff
@@ -11,6 +11,11 @@
     # can be added under this section.
     #input:
 
+    # Customised Grok pattern. You can provide your own customised
+    # grok pattern here
+    #grok_pattern: >
```
ruflin: We should prefix this by `access` as it's the grok pattern for the access log.

ruflin:
In general +1 on allowing the user to modify it. The part I'm most worried about is that the user will use different fields or miss some fields so the dashboard breaks and create potential naming conflicts. But I don't see an easy way around it.

I wonder if we should allow the user to only change the grok patter or change the pipeline. If a user uses a different grok pattern he potentially also needs a different pipeline to rename fields etc. 

If the user uses his own grok pattern, we should find a way to at log a warning message that things might break in the future. 

It's a problem we partially already have now since we introduce dissect in some modules. With dissect the user can already today change the way things are processed.

In general I wonder if we should start to version pipelines and allow a user to configure if he wants to keep this version or the new one on upgrade. I expect most users on upgrade only replace the binary but keep everything else in place?

kvch:
I like the idea of customizing pipelines of Filebeat modules. This could make everyone's life much easier.

I think `grok_pattern` should be a list of Grok patterns. So users can define more than one custom pattern. Thus, I would rename the option to `grok_patterns`. The configuration would look quite similar:
```yml
grok_patterns:
 - "?%{IP_LIST:nginx.access.remote_ip_list} - %{DATA:nginx.access.user_name} \\[%{HTTPDATE:nginx.access.time}\\] "%{GREEDYDATA:nginx.access.info}" %{NUMBER:nginx.access.response_code} %{NUMBER:nginx.access.body_sent.bytes} "%{DATA:nginx.access.referrer}" "%{DATA:nginx.access.agent}"
 - %{NUMBER:nginx.access.response_code} %{NUMBER:nginx.access.body_sent.bytes} "%{DATA:nginx.access.referrer}" "%{DATA:nginx.access.agent}"
```
I suppose this way we could get rid of the additional `"`s at the beginning and end. So escaping `"` could be skipped.

I am thinking about adding a list of required field names to the configuration file as a comment. The list could be generated based on the Ingest pipeline and the Kibana objects that comes with the module.
```yml
# If the following fields are not present in the event after applying
# the grok processor, the pipeline and/or visualizations provided
# by Filebeat might break.
# List of required fields: nginx.access.remote_ip_list, nginx.access.user_name, ...
```

@tsg Do you want to introduce customization of processors in pipelines in the future?

@ruflin I would like to avoid versioning of pipelines as long as it's not block us.

ruflin:
I like the suggestion made above by @kvch 

For dissect: Here is an example on where we use it: https://github.com/elastic/beats/blob/master/filebeat/module/traefik/access/config/traefik-access.yml#L8 The main different from grok is where the processing happens. An other difference is on where the change applies. For the grok at the moment it will apply to all nginx modules ingesting data. So if there are 2 independent nginx services with 2 different log files, things will break. But that issue we have today.

For the versioning of pipelines: It's not blocking us yet but it holds us back in using the most recent features. For example current `@timestamp` cannot be used in grok patterns because of the `@` and we have have extra renaming steps inside to work around it. This issue will be fixed in upcoming Elasticsearch versions. What will we do now? Keep the old more complex pipeline start using the new features? I personally think we should find an option to keep old pipeline files around for compatiblity reason but at the same time be able to use most recent features. Only indirectly related to the above but the solution to it could be similar.

Having said all the above, +1 on moving forward here.

kvch:
@ruflin You got me. I won't protest against versioning pipelines.

ruflin:
An interesting feature that could be of use here is that we can have `min_elasticsearch_version` config option: https://github.com/elastic/beats/blob/master/filebeat/fileset/fileset.go#L195 

Instead of having multiple versions of the pipelines I would prefer to have something like fields.yml that we can add the additional parts which are ES version specific (@urso  ;-) ).

ruflin:
@tsg Pretty sure it will bubble up again and then I will remember :-)

PR #7544, opened 2018-07-09
Add tests to heartbeat's look.go

andrewvc:
Add a few tests to heartbeat's look.go

### DIFF
--- a/heartbeat/look/look_test.go
+++ b/heartbeat/look/look_test.go
@@ -1,0 +1,60 @@
+package look
+
+import (
+	"testing"
+	"time"
+
+	"fmt"
+
+	"github.com/stretchr/testify/assert"
+
+	reason2 "github.com/elastic/beats/heartbeat/reason"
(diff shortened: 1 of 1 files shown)


heartbeat/look/look_test.go
```diff
@@ -0,0 +1,60 @@
+package look
+
+import (
+	"testing"
+	"time"
+
```
ruflin: Nit: Remove the newline here. Sometime goimport does this :-(

heartbeat/look/look_test.go
```diff
@@ -0,0 +1,60 @@
...
+	"fmt"
+
+	"github.com/stretchr/testify/assert"
+
+	reason2 "github.com/elastic/beats/heartbeat/reason"
```
ruflin: I would normally rename the variable instead of the package but purely personal preference.

ruflin:
Merged as the above are mainly comments about presonal preferences.

Issue #3741, opened 2017-03-13
Heartbeat report TLS details


urso:
When heartbeat is connecting via TLS, report information on server certificate and connection details if possible:

- TLS version used
- cipher being used
- server certificate expiration date

RobertoFlores:
There is a date to include this feature on Heartbeat?

ruflin:
@RobertoFlores There is no specific date.

andrewvc:
As an update we now support the expiration dates. I'm on the fence WRT the other info as that crosses into security scanning tools when we're thinking about auditing things like cipher suites and TLS versions.

On the other hand, these are simple tiny fields.

andrewvc:
I propose we close this for now unless there are any objections. TLS version and cipher suite are usually in the realm of vulnerability scanning tools. I'm not sure they belong in heartbeat.

PR #7545, opened 2018-07-09
Heartbeat TLS metadata

### LINKED ISSUE #6300: [heartbeat] tls cert event data proposal
In https://github.com/elastic/beats/issues/3406#issuecomment-292488522 there was a suggestion for adding extra metadata including:

```
"tls": {
  "rtt": { "handshake": ... },
  "status": {
    "version": "TLS 3.0",
    "cipher": ...
  }
  "cert": {
    "expires": ...,
    ... // more certificate meta data?
  }
}
```

Specifically I am mostly interested in sending the `tls.cert.expires` key so that we can easily alert on certificates that will be expiring soon instead of the monitor failing once it has already expired. 

## Proposal

Ideally I would like to implement the following extra fields

1. `tls.cert.expires.notAfter: 2016-01-03T00:00:00.000Z`
2. `tls.cert.expires.notBefore: 2019-02-06T00:00:00.000Z`
3. `tls.status.version: TLS 3.0`
4. `tls.status.cipher: ECDHE-RSA-AES256-GCM-SHA384`

There are a lot of other fields that could be added however I see these being the most interesting when being combined with alerting.

andrewvc:
WIP Work toward adding extra TLS metadata. Would resolve #6300 and #3741 once complete. 

The PR is based off of #7784 , hence the large diff. We should wait for that to merge before reviewing this further.

I'm submitting this PR early to get feedback from @urso on this approach to injecting the TLS metadata into the dialchain. The code is very rough, but I wanted to get feedback early before I start adding tests and rounding this out.

Another goal of this PR would be to add tests to the namespaces and types this PR touches. Heartbeat needs more tests and the strategy here is to add a feature, and retroactively add tests alongside it.

One thing we might also want to consider is that there is shared code for the dialchain in `libbeat/outputs/transport` that probably should just move up a level to `libbeat/transport`.


heartbeat/monitors/active/dialchain/tls.go
```diff
@@ -35,24 +35,39 @@ import (
...
-
 			timer.stop()
 			event.Put("tls.rtt.handshake", look.RTT(timer.duration()))
+
+			connState := meta.Tls.ConnectionState()
+			certs := make([]common.MapStr, len(connState.VerifiedChains))
+			for chainIndex, chain := range connState.VerifiedChains {
+				for certIndex, cert := range chain {
+					j := common.MapStr{
+						"chain_index":      chainIndex,
```
andrewvc: This is only a subset of the fields for now, but I would expand this to the full field list.

ruflin: Also take into account what @ph contributed to ECS: https://github.com/elastic/ecs#tls Perhaps we can extend this further.

andrewvc:
This is much improved, but now needs some modifications to work with [ECS](https://github.com/elastic/ecs#tls). We have much richer data than the few fields ECS asks for, but I'll examine this in more detail tomorrow.

heartbeat/monitors/active/dialchain/tls.go
```diff
@@ -35,24 +36,36 @@ import (
...
 			return nil, err
 		}
 
 		return afterDial(dialer, func(conn net.Conn) (net.Conn, error) {
 			// TODO: extract TLS connection parameters from connection object.
-
 			timer.stop()
 			event.Put("tls.rtt.handshake", look.RTT(timer.duration()))
+
+			connState := meta.TLS.ConnectionState()
```
urso: We really need all the wiring up? There is quite some indirection here, I'd rather keep out if possible. The `conn` connection object should be a tls connection (as is being produced by TLSDialer). We should read the connection state from the `conn` parameter if possible.

andrewvc:
I'm pausing this a bit because we need better ECS schema definitions for TLS. I've opened up a GH issue to start the discussion here: https://github.com/elastic/ecs/issues/64

heartbeat/monitors/active/dialchain/tls.go
andrewvc: Hmmm, what exactly are you thinking here? `tls.Conn` in [crypto/tls/conn.go](https://golang.org/src/crypto/tls/conn.go) is not a super-type (is that the go lingo?) of `net.Conn`, so cannot be cast. It simply nests a `net.Conn` as a property.

AFAICT this means we have to use this sort of indirection. Would love a simpler alternative though. Do I misunderstand?

heartbeat/monitors/active/dialchain/tls.go
```diff
@@ -35,24 +36,36 @@ import (
...
+			certs := make([]common.MapStr, 0, len(connState.VerifiedChains))
+			for chainIndex, chain := range connState.VerifiedChains {
+				for certIndex, cert := range chain {
+					j := tls.CertToMap(cert, true)
+					j["chain_index"] = chainIndex
+					j["cert_index"] = certIndex
+					certs = append(certs, j)
+				}
+			}
+			event.Put("tls.certificates", certs)
```
ruflin: I'm a bit concerned that certificates here is going to be a nested object in Elasticsearch it seems.

What are the expected queries the user will run against these fields?

heartbeat/monitors/active/dialchain/tls.go
urso: tls.Conn wraps another connection, because TLS sits on top of another network layer. But tls.Conn implements `net.Conn` itself. In go we do not have inheritance or subtype relationships. But `tls.Conn` structurally satisfies the `net.Conn` interface. When building layers of dialers, we have to fix the return values of dialers, due to lack of generics and sub-typing (no co-/contra-variance on methods). The common denominator for dialers is given by:

```
type Dialer interface {
	Dial(network, address string) (net.Conn, error)
}
```

As `net.Conn` is an interface, each layer will wrap it's input layer + use a concrete type which is finally promoted to the interface value being returned.

This is reflected in the DialerChain type:

```
type DialerChain struct {
	Net    NetDialer
	Layers []Layer
}
```

The NetDialer produces the most low-level connection object. Each layer will wrap the current connection object, returning a new connection object.

As we want to do some measurements before and after a network layer becomes active, I introduced the `beforeDial` and `afterDial` wrappers, so to reduce common boilerplate and amount of wrapper functions needed. Again, lack of generics and sub-type relationships we do enforce the use of `net.Conn`. But in general, once a dialer-chain is setup, we have this layout:

```
+----+-----------------------+----+ net.Conn
     |  After Layer II       |
     +-----------------------+ Layer II Conn type
     |  Dial Layer II        |
     +-----------------------+
     |  Before Layer II      |
+---------------------------------+  net.Conn
     |  After Layer I        |
     +-----------------------+ Layer I Conn type
     |  Dial Layer I         |
     +-----------------------+
     |  Before Layer I       |
+---------------------------------+ net.Conn
     |  Net Layer (e.g. TCP) |
     +-----------------------+
```

A Layer must make no assumptions about the underlying layer. A layer must assume the connection object to wrap to be of type `net.Conn`. But within a layer (dial + after steps), we know the concrete type returned by the dialer and to be consumed by the 'after'-handler. Due to limitations of the type-system, we are forced to cast the interface value to the concrete type, though.

In case of the TLS layer, we use transport.TLSDialer. The concrete type returned  by `(*TLSDialer).Dial` is `*tls.Conn`. => we can cast the interface value to `*tls.Conn` in the 'after'-handler. As we know the concrete type in the 'after'-handler, we don't need another behind-the-scenes type capturing the concrete type. Admittedly, it's more type safe, but it's somewhat more involved and somewhat over-engineered for something that should be simple (as the after-handler is the only consumer that requires to access the tls state).

andrewvc: Thanks for the very detailed writeup. I'm realizing that my confusion was in typing `conn.(tls.Conn)` instead of `conn.(*tls.Conn)`. That was the root of my assumption that the interface wasn't implemented.

Thanks for the detailed diagram of the dialer stack, that's very helpful!

andrewvc:
OK, waiting for the build to finish, but this is looking mostly done.

This includes the following changes:

1. We now write certificate PEM data, and chain validation times to events
2. Enhanced test helpers (including mapval IsDefs) to handle new testing scenarios
3. Improved mapval handling of type mismatches.
4. Removed indirection around dialer metadata. Was able to simply cast the `afterDial` arg `conn` to a `*tls.Conn`.

heartbeat/monitors/active/dialchain/builder.go
```diff
@@ -37,6 +37,7 @@ type Builder struct {
 	template         *DialerChain
 	addrIndex        int
 	resolveViaSocks5 bool
+	meta             common.MapStr
```
urso: Is this used anywhere? I'd prefer to not use common.MapStr if it's not really required.

heartbeat/monitors/active/dialchain/tls.go
ruflin: We should move that diagram to the code docs somewhere.

heartbeat/monitors/active/dialchain/builder.go
andrewvc: Good catch, removing

andrewvc:
Added a bunch of specs for `mapval.Path` since code coverage checks were failing.

I added a little to this file, so it's fair to add more coverage in this PR IMHO.

andrewvc:
OK, should be ready for final review.

heartbeat/monitors/active/dialchain/tls.go
```diff
@@ -21,9 +21,14 @@ import (
 	"net"
 	"time"
 
+	cryptoTLS "crypto/tls"
```
ruflin: as usual goimports messes with you :-)

heartbeat/monitors/active/dialchain/tls.go
```diff
@@ -21,9 +21,14 @@ import (
...
+
+	"fmt"
+
 	"github.com/elastic/beats/heartbeat/look"
 	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/outputs/transport"
+	"github.com/elastic/beats/packetbeat/protos/tls"
```
ruflin: I'm a bit concerned about this import. Beats should not have dependencies with each other. Share code like this be moved into libbeat.

heartbeat/monitors/active/dialchain/tls.go
```diff
@@ -49,10 +54,38 @@ func TLSLayer(cfg *transport.TLSConfig, to time.Duration) Layer {
 		}
 
 		return afterDial(dialer, func(conn net.Conn) (net.Conn, error) {
-			// TODO: extract TLS connection parameters from connection object.
+			tlsConn, ok := conn.(*cryptoTLS.Conn)
+			if !ok {
+				panic(fmt.Sprintf("TLS afterDial received a non-tls connection %t. This should never happen", conn))
```
ruflin: Do we really want to panic here or rather return / log a critical error?

heartbeat/monitors/active/dialchain/tls.go
```diff
@@ -49,10 +54,38 @@ func TLSLayer(cfg *transport.TLSConfig, to time.Duration) Layer {
...
+			// TODO: extract TLS connection parameters from connection object.
 			timer.stop()
 			event.Put("tls.rtt.handshake", look.RTT(timer.duration()))
+
+			var certs []string
+
+			// Pointers because we need a nil value
+			var chainNotValidBefore *time.Time
+			var chainNotValidAfter *time.Time
+			for _, chain := range tlsConn.ConnectionState().VerifiedChains {
```
ruflin: Would be nice to have a code comment here that states we check all certs and take the oldest / newest timestamp. for before and after and not sure from the last certificate in the chain.

heartbeat/hbtest/hbtestutil.go
```diff
@@ -18,14 +18,20 @@
...
 	"net/url"
+	"os"
 	"strconv"
+	"testing"
 
-	"net/http/httptest"
+	"github.com/stretchr/testify/require"
 
 	"github.com/elastic/beats/libbeat/common/mapval"
+	"github.com/elastic/beats/packetbeat/protos/tls"
```
ruflin: See comment below about packetbeat dependency.

heartbeat/monitors/active/tcp/tcp_test.go
```diff
@@ -19,12 +19,17 @@ package tcp
 
 import (
 	"fmt"
+	"net/http"
 	"net/http/httptest"
 	"testing"
 
-	"github.com/stretchr/testify/require"
+	"crypto/x509"
```
ruflin: goimports messing with you.

libbeat/common/mapval/path.go
```diff
@@ -32,11 +32,21 @@ type PathComponentType int
...
 	PCSliceIdx
 )
 
+func (pct PathComponentType) String() string {
+	if pct == PCMapKey {
+		return "map"
+	} else if pct == PCSliceIdx {
+		return "slice"
+	} else {
+		panic(fmt.Sprintf("Unknown path component type! %d", int(pct)))
```
ruflin: Let's use an error here instead of a panic. What if someone uses this for production code and is not aware this can apnice?

packetbeat/protos/tls/parse.go
```diff
@@ -543,7 +543,9 @@ func getKeySize(key interface{}) int {
 	return 0
 }
 
-func certToMap(cert *x509.Certificate, includeRaw bool) common.MapStr {
+// CertToMap takes an x509 cert and converts it into a map. If includeRaw is set
+// to true a PEM encoded copy of the cert is encoded into the map as well.
+func CertToMap(cert *x509.Certificate, includeRaw bool) common.MapStr {
```
ruflin: Better move this libbeat.

libbeat/common/mapval/is_defs.go
```diff
@@ -115,20 +115,43 @@ var IsDuration = Is("is a duration", func(path Path, v interface{}) *Results {
...
+				return ValidResult(path)
+			}
+			return SimpleResult(
+				path,
+				false,
+				fmt.Sprintf("objects not equal: actual(%v) != expected(%v)", v, to),
+			)
+		})
+	}
+}
```
urso: As you are already a heavy user of reflection, how about this:

```
var equalChecks = map[reflect.Type]reflect.Value

// accepts and register equal function for function signatures:
// func(a, b T) bool
func RegisterEqual(fn interface{}) {
  v := reflect.ValueOf(fn)
  t := v.Type()
  if t.Kind() != reflect.Function { ... }
  if t.NumIn() != 2 { ...}
  if t.NumOut() != 1 { ... }
  if t.In(0) != t.In(1) { ... }
  if t.Out(0) != <bool> { ... }

  equalChecks[t.In(0)] = v
}

func IsEqual(to interface{}) IsDef {
  vTo := reflect.ValueOf(to)
  return Is("equals", func(path Path, in interface{}) *Results {
    vIn := reflect.ValueOf(in)
    if vTo.Type() != vIn.Type() { ... }

    // we always query, in case someone did add a comparison function after
    checker := equalChecks(vTo.Type())
    if checker == nil {
      // use DeepEqual
    }

    equal := checker.Call([]reflect.Value{vTo, VIn}).Bool()
    if !equal {
      ...
    }
  }
}
```

heartbeat/monitors/active/dialchain/tls.go
urso: Lets keep the panic here. If this fails it's definitely a programming error. Given we have TLS tests, this panic should only occur during testing.
If we hide this behind an error message being logged, we might also have test failures (or a bug at worst), but then we will have to search for the root cause. Being a little more offensive can help.

libbeat/common/mapval/path.go
andrewvc: Well, this practically cannot panic. The only way that would happen is if someone was messing deep in the guts of mapval and added a new type without overriding this.

I would say then, that if this did panic, it'd be a bug in mapval itself, not in user-facing code. Checking invariants like that seems like the right place for panic no?

packetbeat/protos/tls/parse.go
andrewvc: Cool, will move that whole package.

libbeat/common/mapval/is_defs.go
andrewvc: This is pretty nifty, out of curiosity I wonder if using reflect is faster than a type switch? I assume they must be pretty similar perf-wise under the covers?

I am however worried about YAGNI. How many special equals cases are there? `reflect.DeepEquals` works pretty well, this is the first case it hasn't.

If there are only a few, I personally prefer the switch because its very concise, however, if we can foresee adding 10 of these cases a map sounds better.

WDYT?

libbeat/common/mapval/path.go
andrewvc: I'm realizing this is similar to the other panic described earlier. Any case where another type is encountered would definitely be a programming error. You would need to create a new `PathComponentType` with a different (manual) value. 

@urso I see you +1'd this change, but opposed removing the other panic. Maybe you can help me understand the difference?

urso: Assuming mapval package could be used in other places in beats, the mapval package might be used to validate any kind of external input. We might even consider users to configure validation (based on mapval) in the heartbeat config file. When working with external input, mapval should not panic.

If the panic can only indicate programming errors, I'm fine with the panic. If the panic can be triggered by external (yet unchecked) input then we should go with an error. mapval has to check input, so we must assume any kind of invalid input.

It's common in String methods to not create a panic, but return an 'Unknown'-string.

Btw. consider using stringer like:

```

//go:generate stringer -type=PathComponentType -linecomment=true

const (
	PCMapKey PathComponentType = 1 + iota // map

	// PCSliceIdx is the Type for slice indices.
	PCSliceIdx                            // slice
)
```

Notice how I set `PCMapKey` to 1. That is, the zero-value is invalid. Oftentimes one defines an 'invalid'-const to be 0 like:

```

//go:generate stringer -type=PathComponentType -linecomment=true

const (
    PCMapInvalid PathComponentType = iota // invalid
	PCMapKey                              // map
	PCSliceIdx                            // slice
)
```

libbeat/common/mapval/is_defs.go
urso: reflection is somewhat smaller.

In go an `interface{}` value has the internal representation:

```
type ifcType struct {
  typ *typeDescr
  value *unsafe.Pointer
}
```

The reflection package defines Value to be:

```
type Value struct {
  typ *typeDescr
  value *unsafe.Pointer
  flags flags
}
```

when using a type-switch, the generated code checks if the 'typ' matches and casts the ifc pointer to your appropriate type. If your target type is also a pointer, its a comparison + copy pointer into new variable.

When using reflection, each call to any Value method checks flags, types and so on, checking if your call is actually valid. The `Value.Call` method is super involved. Reflection is definitely slower, but gives you some flexibility here.

reflect.DeepEquals has it's limits. It compares each field (including private fields) in a struct. Private fields might be different, while 2 objects might still compare equal. E.g. when comparing regular expressions. DeepEquals compares public and private fields. But the regex package introduced an object pool pointer in tip. Each regexp.Regex type holds a unique pointer => comparison fails, if one does not compare the same pointer, even if two regular expression are generated from the same string.

andrewvc: Cool, thanks for the detailed walkthrough. I think you're right, we'll probably add more of these, I'll add the registry.

libbeat/common/mapval/path.go
andrewvc: Thanks for the feedback, I'll take your advice and return an unknown string. I think that's a great compromise.

Also, thanks for the tip on the 1+, makes a lot of sense.

andrewvc:
@urso I've incorporated both of your suggestions,

for using reflection in equality in c2740e0f and for path panic-ing in 5a0a42e

andrewvc:
Squashed

Issue #7534, opened 2018-07-06
Filebeat - iis module - iis.access.hostname should be case-insensitive


mhunsber:
The iis-access-default pipeline should have either "lowercase" or "uppercase" processors for the field "iis.access.hostname" because RFC 4343 says DNS names are case-insensitive.

You could have a site hosted at mysite.com, and if the client connects to MySite.Com, filebeat will log these two hostnames as separate.

Most sites also have case-insensitive urls, but it is not standard.

ruflin:
@mhunsber Makes sense. Interested to open a PR?

knrt10:
@ruflin I would like to work on this.

PR #7558, opened 2018-07-10
Filebeat - iis module - iis.access.hostname should be case-insensitive

mhunsber:
Issue: #7534 
[RFC 4343](https://tools.ietf.org/html/rfc4343)

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -188,6 +188,7 @@ https://github.com/elastic/beats/compare/v6.2.3...master[Check the HEAD diff]
 
 *Filebeat*
 
+- Add lowercase processor to iis access pipeline {issue}7534[7534]
 - Add IIS module to parse access log and error log. {pull}6127[6127]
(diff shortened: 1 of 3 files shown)


CHANGELOG.asciidoc
```diff
@@ -188,6 +188,7 @@ https://github.com/elastic/beats/compare/v6.2.3...master[Check the HEAD diff]
 
 *Filebeat*
 
+- Add lowercase processor to iis access pipeline {issue}7534[7534]
```
ruflin: Could you changes to something like `Change field "iis.access.hostname" to always be lower case`. Otherwise users reading this don't really know what changed in the user facing part.

mhunsber: That makes sense, thank you.

jsoriano:
Not sure about this... should we keep the original value in another field in case someone wants to query things like how many users use `MySite.Com` vs `mysite.com`? :thinking:

Issue #7562, opened 2018-07-10
[Beats Monitoring] Monitoring data payload has breaking changes in schema from 6.3 to 7.0


tsullivan:
In 6.3.x, the `beats_stats` field in `.monitoring-beats` data looks like:
```
"beats_stats": {
  "timestamp": ...
    "beat": {
      "type": "filebeat",
      "version": "6.3.1",
      ...
    },
    "metrics": {
      "libbeat": {
        "config": {
          ...
        },
        "output": {
          "events": {
            ...
          },
          "write": {
            ...
          },
          "read": {
            ...
          },
          "type": ...
        },
        "pipeline": {
          "clients": ...,
          "events": {
            ...
          },
          "queue": ...
        }
      },
      "system": {
        ...
      },
      "beat": {
        ...
      },
      ...
    }
}
```

In 7.0/master branch, the `beats_stats` schema has changed and now looks like:
```
"beats_stats": {
  "timestamp": ...,
    "beat": {
      "type": "filebeat",
      "version": "7.0.0-alpha1",
      ...
    },
    "stats": {     # WAS: metrics
      "libbeat": {
        "config": {
          ...
        },
        "output": {
          "events": {
            ...
          },
          "write": {
            ...
          },
          "read": {
            ...
          },
          "type": ...
        },
        "pipeline": {
          "clients": ...,
          "events": {
            ...
          },
          "queue": ...
        }
      },
      "system": {
        ...
      },
      "beat": {
        ...
      },
      ...
    }
}
```

(Ellipses added for length)

This is a breaking change for the Monitoring UI, since `beats_stats.metrics` is used for a lot of info about Beats in the cluster.

tsullivan:
If the name change for the `metrics` field changed in 6.4, then this would be a blocker for 6.4.0

ruflin:
I opened a PR with a fix here: https://github.com/elastic/beats/pull/7563 Thanks for bringing this up.

PR #7563, opened 2018-07-11
Fix breaking change in monitoring data

ruflin:
The prefix for the stats metrics was metrics but renamed to `stats` by accident as the name is now auto generated. This reverts this change.

Closes https://github.com/elastic/beats/issues/7562

### DIFF
--- a/libbeat/monitoring/report/elasticsearch/elasticsearch.go
+++ b/libbeat/monitoring/report/elasticsearch/elasticsearch.go
@@ -198,11 +198,12 @@ func (r *reporter) initLoop(c config) {
 	logp.Info("Successfully connected to X-Pack Monitoring endpoint.")
 
 	// Start collector and send loop if monitoring endpoint has been found.
-	go r.snapshotLoop("state", c.StatePeriod)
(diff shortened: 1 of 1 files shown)


libbeat/monitoring/report/elasticsearch/elasticsearch.go
```diff
@@ -224,9 +224,15 @@ func (r *reporter) snapshotLoop(namespace string, period time.Duration) {
 			continue
 		}
 
+		// Prefix for the inside the event. For backward compatibility stats is renamed to metrics.
+		prefix := namespace
+		if namespace == "stats" {
+			prefix = "metrics"
+		}
```
jsoriano: What about passing the prefix as a parameter to this function to keep it generic? Then the calls would be:
```
go r.snapshotLoop("state", "state", c.StatePeriod)
go r.snapshotLoop("stats", "metrics", c.MetricsPeriod)
```

ruflin: good idea, changed.

jsoriano:

PR #7456, opened 2018-06-28
Fix Spelling errors

jsoref:
Generated by https://github.com/jsoref/spelling `f`; to maintain your repo, please consider `fchurn`

PR #7565, opened 2018-07-11
Correct typo

ritwik12:
multiline was spelled wrong

### DIFF
--- a/filebeat/filebeat.yml
+++ b/filebeat/filebeat.yml
@@ -48,7 +48,7 @@ filebeat.inputs:
 
   ### Multiline options
 
-  # Mutiline can be used for log messages spanning multiple lines. This is common
+  # Multiline can be used for log messages spanning multiple lines. This is common
   # for Java Stack Traces or C-Line Continuation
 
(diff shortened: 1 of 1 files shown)


ruflin:
jenkins, test this

ritwik12:
It's just a simple typo correction, I don't know why is it failing. :(

andrewkroh:
It's failing because the file you changed gets generated and checked in. The input is at

https://github.com/elastic/beats/blob/master/filebeat/_meta/common.p2.yml#L42

After fixing `filebeat/_meta/common.p2.yml` then running `make update` from within the filebeat dir will regenerate the config files. And CI should be happy. 👍 

The error from CI is "needs update". https://travis-ci.org/elastic/beats/jobs/402553451#L832

andrewkroh:
But perhaps this was already fixed by https://github.com/elastic/beats/pull/7456?

ruflin:
Yes, I think it's also fixed there. Will close the PR.

@ritwik12 Thanks for opening the PR, coincidence that it was fixed in 2 places at the same.

PR #7569, opened 2018-07-11
Add envoyproxy module

berfinsari:
The envoyproxy module collects metrics.


metricbeat/module/envoyproxy/server/data.go
```diff
@@ -0,0 +1,187 @@
...
+		"stats.*",
+		"server.*",
+		"listener.admin.*",
+		"listener.admin.http.admin.*",
+		"http.admin.*",
+		"http.async-client.*",
+	}
+	for i := 0; i < len(stats); i++ {
+		event := findStats(data, response, stats[i])
+		events, err = schema.Apply(event)
```
exekias: Isn't this overriding the `events` var on every pass? It sounds like you could call it just once after the loop (using the loop to populate `data` only)

exekias: Also, have a look to Apply options, the default behavior is: all fields are required, you will get an error if any of them is missing. I'm wondering if some fields may change between releases. A more relaxed check would be: `schema.Apply(event, schema.FailOnRequired)`. Pinging @jsoriano on this

metricbeat/module/envoyproxy/server/data.go
```diff
@@ -0,0 +1,187 @@
...
+		if err != nil && i == (len(stats)-1) {
+			return nil, err
+		}
+	}
+	return events, nil
+}
+
+func findStats(data common.MapStr, response []byte, reName string) common.MapStr {
+	re := regexp.MustCompile(reName)
```
exekias: you can avoid recompiling this on every fetch by moving from a local `stats := []string` to a static var `var stats []*regexp.Regexp`

metricbeat/module/envoyproxy/server/data.go
```diff
@@ -0,0 +1,187 @@
...
+			"memory_allocated":   c.Int("memory_allocated"),
+			"memory_heap_size":   c.Int("memory_heap_size"),
+			"parent_connections": c.Int("parent_connections"),
+			"total_connections":  c.Int("total_connections"),
+			"uptime":             c.Int("uptime"),
+			"version":            c.Int("version"),
+			"watchdog_mega_miss": c.Int("watchdog_mega_miss"),
+			"watchdog_miss":      c.Int("watchdog_miss"),
+		},
+		"listener": s.Object{
```
jsoriano: Is the list of `listeners` and `http` objects dynamic? If it is maybe `Schema` (as is now) is not the best option to parse these fields.

metricbeat/module/envoyproxy/server/data.go
jsoriano: There is an example of required fields [here](https://github.com/elastic/beats/blob/14da4086cab07a522b4d5db2e5c9ed2ee906df73/libbeat/autodiscover/providers/jolokia/discovery.go#L59).

berfinsari: Thanks for your review. I think now overriding issue is solved but I am not sure which fields are required. I am handle with this.

metricbeat/module/envoyproxy/server/data.go
berfinsari: What do you recommend to parse these fields? Do you have an example for this?

berfinsari: Thank you for the reply! I removed dynamic fields and updated docs for the non-dynamic fields.  In the meantime, I will look to other modules for dynamic fields.

berfinsari:
@exekias Did you have any chance to check latest changes?

ruflin:
jenkins, test this

exekias:
Thank you so much for this contribution @berfinsari! Do you happen to have dashboards for this module? It would be awesome to include them in the distribution, so all users benefit from them!

PR #7571, opened 2018-07-11
Add UDP support to packetbeat's process monitor

### LINKED ISSUE #541: Packetbeat - Associate a process with generic TCP/UDP flows

adriansr:
This updates the process monitor with UDP support.

Until now, it would lookup all ports as TCP. This could result in the wrong process being reported when two different processes used the same port number, one for TCP and one for UDP.

The Linux and Windows implementations have been updated to fetch information for UDP ports.

Closes #541

### DIFF
--- a/packetbeat/docs/packetbeat-options.asciidoc
+++ b/packetbeat/docs/packetbeat-options.asciidoc
@@ -1283,15 +1283,15 @@ process matching is disabled.
 When Packetbeat starts, and then periodically afterwards, it scans the process table for
 processes that match the configuration file. For each of these processes, it
 monitors which file descriptors it has opened. When a new packet is captured,
-it reads the list of active TCP connections and matches the corresponding one
+it reads the list of active TCP and UDP connections and matches the corresponding one
 with the list of file descriptors.
 
-On a Linux system, all this information is available via the `/proc`
-file system, so Packetbeat doesn't need a kernel module.
-
+All this information is available via system interfaces: The `/proc` file system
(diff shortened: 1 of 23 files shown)


andrewkroh:
Related to this page: https://www.elastic.co/guide/en/beats/packetbeat/master/configuration-processes.html

It needs to be updated to mention UDP. And the note about Linux only should be expanded to include Windows.

adriansr:
Thanks @andrewkroh, I've added a new commit with updated docs.

packetbeat/docs/packetbeat-options.asciidoc
```diff
@@ -1314,6 +1314,14 @@ packetbeat.procs:
       cmdline_grep: gunicorn
 ------------------------------------------------------------------------------
 
+When the process monitor is enabled, it will enrich all the events whose source
+or destination is a local process. The `cmdline` and/or `client_cmdline` fields
+will be added to an event, when the server side or client side of the connection
+belong to a local process, respectively. Additionally, you can specify a pattern
+using the `cmdline_grep` option, to also name those processes. This will cause
```
andrewkroh: This totally helps clarify how it works for me.

PR #7572, opened 2018-07-11
add conditionals to support apm-server docs update

graphaelli:
This is to support https://github.com/elastic/apm-server/pull/1103.  While that's still in review, it's moving along and I'd like to know if there are any objections here that need to be factored in there.

### DIFF
--- a/libbeat/docs/command-reference.asciidoc
+++ b/libbeat/docs/command-reference.asciidoc
@@ -20,11 +20,11 @@
 :modules-command-short-desc: Manages configured modules
 :run-command-short-desc: Runs {beatname_uc}. This command is used by default if you start {beatname_uc} without specifying a command
 
-ifeval::["{has_ml_jobs}"=="yes"]
+ifdef::has_ml_jobs[]
 :setup-command-short-desc: Sets up the initial environment, including the index template, Kibana dashboards (when available), and machine learning jobs (when available)
 endif::[]
 
-ifeval::["{has_ml_jobs}"!="yes"]
+ifndef::has_ml_jobs[]
 :setup-command-short-desc: Sets up the initial environment, including the index template, Kibana dashboards (when available)
 endif::[]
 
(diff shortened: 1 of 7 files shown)


libbeat/docs/command-reference.asciidoc
```diff
@@ -20,11 +20,11 @@
 :modules-command-short-desc: Manages configured modules
 :run-command-short-desc: Runs {beatname_uc}. This command is used by default if you start {beatname_uc} without specifying a command
 
-ifeval::["{has_ml_jobs}"=="yes"]
+ifdef::has_ml_jobs[]
```
ruflin: Can you share some background on how this is different from the previous true / false?

ruflin:
@graphaelli In general I don't see any reason we shouldn't have that.
@dedemorton Would be good if you could have a look at this one.

libbeat/docs/command-reference.asciidoc
graphaelli: In `libbeat/docs/security/basic-auth.asciidoc` `has_ml_jobs` is used as `ifdef`.  It seems we typically use `ifdef` over `ifeval` with a value so I sync'd in that direction.

graphaelli:
elastic/apm-server#1103 is now ready for merge.  Waiting on approval of these changes for a bit before merging.

ruflin:
@graphaelli Merging as I definitively don't want to block you on this one and change looks fine.

@dedemorton Please still have a look and let me know if we need to change / revert some things.

graphaelli:
thank you!

dedemorton:
Looks good! The only problem I noticed was one that I introduced. See https://github.com/elastic/beats/pull/7602.

@graphaelli You'll probably want to pick up the changes in https://github.com/elastic/beats/pull/7602 after it's merged.

graphaelli:
@dedemorton thanks for the heads up

PR #7573, opened 2018-07-11
Ensure metadata updates don't replace existing pod metrics

exekias:
Before things change, `add_kubernetes_metadata` would replace everything
under `kubernetes.pod` when enriching events.

This is not what you want, especially if you are using it in conjunction
with the Kubernetes module, where you could smash pod metrics when
enriching.


ruflin:
Probably you need to rebase on master to also fix the generator header error.

exekias:
rebased, tests should pass now

PR #7582, opened 2018-07-12
Separate Line reader into a reader and a scanner

kvch:
### Architecture
The existing `Line` reader is separated into two smaller units: a reader and a scanner. The reader decodes and converts the input bytes into UTF-8. Then the scanner looks for new lines in the output of the decoder. If it finds one, returns the new line along with its size.

### Challenges
1. Size of the encoded line to store the correct state
    As the scanner is encoding-agnostic, the length of the encoded bytes is unknown when scanning for new lines. However, it's required otherwise the state becomes corrupt. To eliminate this problem I added a helper slice to track bytes sizes. See more in the section "Size calculation".
2. Error handling
    Error detection during converting bytes is hard. Our previous solution was incorrect and hadn't caught every error. No error is reported if a users misconfigures the `encoding` of an `input`. We could use external libraries which guess the correct encoding, but it would not add as much value as it complicates the code and degrades performance. Also, we haven't heard about many problems related to selecting the correct encoding for an input file, so I think we can live without it.
3. Performance
    See more in the dedicated section.

### Size calculation
To calculate the original size of the encoded string I introduced an array. The length of the array equals the number of bytes of the converted string. The i-th element of the array shows how many bytes long in UTF-8 is the original i-th byte. In order to calculate the length of an n bytes long converted string, sum the numbers in the array from the left to right until it equals the `len` of the converted string. The original size is the number of elements that have to be added.
#### Example
String: `"tél"` (It means winter in Hungarian. Congrats, you've just learnt something new!)
Len in UTF-8: 5
```
len("tél") # equals 5
```
Latin1
```
[1 2 1 1]
count: 4
```
Original length of the encoded string: 4 bytes

UTF-8
```
[1 0 2 1 1]
count: 5
```
Original length of the encoded string: 5 bytes

UTF-16
```
[0 1 0 2 0 1 0 1]
count: 8
```
Original length of the encoded string: 8 bytes

### Possible optimizations
This is a general converter which transforms everything byte-by-byte. This could add a significant performance penalty. To avoid bad performance, specialized decoders could be introduced in later PRs.
1. No conversion and skip size calculation when encoding is UTF-8.
2. Convert the whole byte slice and skip size calculation when encoding is ASCII.
3. Convert two bytes at a time when encoding is UTF-16

### DIFF
--- a/filebeat/reader/readfile/decoder.go
+++ b/filebeat/reader/readfile/decoder.go
@@ -1,0 +1,166 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package readfile
+
+import (
+	"fmt"
+	"io"
+
+	"golang.org/x/text/encoding"
+	"golang.org/x/text/transform"
+
+	"github.com/elastic/beats/libbeat/common/streambuf"
+)
+
+type decoderReader struct {
+	in         io.Reader
+	decoder    transform.Transformer
+	buf        *streambuf.Buffer
+	encodedBuf *streambuf.Buffer
+	bufferSize int
+	symlen     []uint8
+}
+
+func newDecoderReader(in io.Reader, codec encoding.Encoding, bufferSize int) *decoderReader {
+	return &decoderReader{
+		in:         in,
+		decoder:    codec.NewDecoder(),
+		buf:        streambuf.New(nil),
+		encodedBuf: streambuf.New(nil),
+		bufferSize: bufferSize,
+	}
+}
+
+func (r *decoderReader) read(buf []byte) (int, error) {
+	b := make([]byte, r.bufferSize)
+
+	if r.buf.Len() != 0 {
+		return r.copyToOut(buf)
+	}
+
+	for {
+		start := r.encodedBuf.Len()
+		n, err := r.in.Read(b[start:])
+		if err != nil {
+			return len(buf[:n]), err
+		}
+
+		if start > 0 {
+			enc, err := r.encodedBuf.Collect(start)
+			if err != nil {
+				return 0, err
+			}
+			r.encodedBuf.Reset()
+			b = append(enc, b[start:]...)
+		}
+
+		nBytes, nProcessed, err := r.conv(b[:start+n], buf)
+		if err != nil {
+			if err == transform.ErrShortSrc {
+				r.encodedBuf.Append(b[nProcessed:])
+				return nBytes, nil
+			}
+			return 0, err
+		}
+		return nBytes, nil
+	}
+}
+
+// msgSize returns the size of the encoded message on the disk
+func (r *decoderReader) msgSize(symlen []uint8, size int) (int, []uint8, error) {
+	n := 0
+	for size > 0 {
+		if len(symlen) <= n {
+			return 0, symlen, fmt.Errorf("error calculating size: too short symlen")
+		}
+
+		size -= int(symlen[n])
+		n++
+	}
+
+	symlen = symlen[n:]
+
+	return n, symlen, nil
+}
+
+func (r *decoderReader) symbolsLen() []uint8 {
+	s := r.symlen
+	r.symlen = []uint8{}
+	return s
+}
+
+// conv converts encoded bytes into UTF-8 and produces a symlen array which
+// records the size of the encoded bytes and its converted size
+func (r *decoderReader) conv(in []byte, out []byte) (int, int, error) {
+	var err error
+	nProcessed := 0
+	decodedChar := make([]byte, 64)
+	r.symlen = make([]uint8, len(in))
+
+	i := 0
+	srcLen := len(in)
+	for i < srcLen {
+		j := i + 1
+
+		for j <= srcLen {
+			nDst, nSrc, err := r.decoder.Transform(decodedChar, in[i:j], false)
+			if err != nil {
+				// if no char is decoded, try increasing the input buffer
+				if err == transform.ErrShortSrc {
+					j++
+
+					// if the buffer size cannot be increased, return what's been decoded and an error
+					if srcLen < j {
+						n, _ := r.copyToOut(out)
+						r.symlen = r.symlen[:nProcessed]
+						return n, nProcessed, err
+					}
+				}
+				err = nil
+			}
+
+			// move in the symlen buffer if no char is decoded
+			if nDst == 0 && nSrc == 0 {
+				nProcessed++
+				continue
+			}
+
+			r.symlen[nProcessed] = uint8(nDst)
+			r.buf.Write(decodedChar[:nDst])
+			nProcessed++
+			break
+		}
+		i = j
+	}
+
+	n, err := r.copyToOut(out)
+	return n, nProcessed, err
+}
+
+func (r *decoderReader) copyToOut(out []byte) (int, error) {
+	until := len(out)
+	if r.buf.Len() < until {
+		until = r.buf.Len()
+	}
+	b, err := r.buf.Collect(until)
+	if err != nil {
+		return 0, err
+	}
+	r.buf.Reset()
+	return copy(out, b), nil
+}
--- a/filebeat/reader/readfile/decoder.go
+++ b/filebeat/reader/readfile/decoder.go
@@ -27,6 +27,9 @@ import (
 	"github.com/elastic/beats/libbeat/common/streambuf"
 )
 
+// decoderReader reads from an input file and converts its content into UTF-8.
+// It stores the size of the original bytes, so offset can be calculated later in
+// the line scanner.
 type decoderReader struct {
 	in         io.Reader
 	decoder    transform.Transformer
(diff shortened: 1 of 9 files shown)


kvch:
Still fighting with the CI.

filebeat/reader/line/line.go
```diff
@@ -18,177 +18,258 @@
...
-		// Check last decoded byte really being '\n' also unencoded
-		// if not, continue reading
-		buf := r.outBuffer.Bytes()
+type decoderReader struct {
+	in         io.Reader
+	decoder    transform.Transformer
+	buf        *streambuf.Buffer
+	encodedBuf *streambuf.Buffer
+	bufferSize int
+	symlen     []int
```
urso: let's safe some space and use `[]uint8`. On 64bit architectures an int takes 8bytes.

filebeat/reader/line/line.go
```diff
@@ -18,177 +18,258 @@
...
+	if r.buf.Len() != 0 {
+		return r.copyToOut(buf)
+	}
 
-		buf := make([]byte, r.bufferSize)
+	for {
+		start := r.encodedBuf.Len()
+		n, err := r.in.Read(b[start:])
+		if err != nil {
+			return len(buf[:n]), err
```
urso: this is equivalent to `return n, err`. But why return `n` here? No contents has been copied to buf yet.

urso: This return does not look correct. Its equivalent to `return n, err`. We will need a global variable tracking the actual number of bytes copied into buf.

filebeat/reader/line/line.go
```diff
@@ -18,177 +18,258 @@
...
+		}
 
-		// try to read more bytes into buffer
-		n, err := r.reader.Read(buf)
+		if start > 0 {
+			enc, _ := r.encodedBuf.Collect(start)
+			b = append(enc, b[start:]...)
```
urso: this is somewhat unfortunate, as we will always allocate and copy if encodedBuf is not empty.

filebeat/reader/line/line.go
```diff
@@ -18,177 +18,258 @@
...
+			if err == transform.ErrShortSrc {
+				r.encodedBuf.Append(b[nProcessed:])
 
-		// empty read => return buffer error (more bytes required error)
-		if n == 0 {
-			return streambuf.ErrNoMoreBytes
+				r.offset += nBytes
+				r.bytesOffset += nProcessed
+				return nBytes, nil
+			}
```
urso: can we hist ErrShortDst here as well?

filebeat/reader/line/line.go
```diff
@@ -18,177 +18,258 @@
...
+	in         *decoderReader
+	separator  []byte
+	bufferSize int
+
+	symlen         []int
+	buf            *streambuf.Buffer
+	offset         int
+	bytesOffset    int
+	lastMessageLen int
+}
```
urso: types should be at the top of the source file. Given both readers are somewhat non-trivial consider splitting them into separate files.

kvch: yes, i kind of expected this request, so i put the types like this, so it would be easier to copy them into different files.

kvch:
The failing tests in Jenkins are unrelated. In Travis everything is green.

kvch:
I am keeping the label in progress on this PR until FF. I would rather not release this refactoring without the additional optimizations.

urso:
The change is not that trivial. Can you have filebeat running a few hours with this PR and monitor memory usage? Do not look at RSS, but check for `Alloc`, `HeapInuse` as provided by the go runtime (see runtime.Memstats). When plotting these, the graphs will be quite jittery. Apply some smoothing filter (e.g. moving average) and check that we did not introduce a leak (it can take a while to discover a leak, due to gc related noise).

filebeat/reader/line/decoder.go
```diff
@@ -0,0 +1,166 @@
...
+	"fmt"
+	"io"
+
+	"golang.org/x/text/encoding"
+	"golang.org/x/text/transform"
+
+	"github.com/elastic/beats/libbeat/common/streambuf"
+)
+
+type decoderReader struct {
```
urso: some docstring about decoderReader transforming the input stream into UTF-8, by using a configurable character encoding would be nice here.

filebeat/reader/line/decoder.go
```diff
@@ -0,0 +1,166 @@
...
+		encodedBuf: streambuf.New(nil),
+		bufferSize: bufferSize,
+	}
+}
+
+func (r *decoderReader) read(buf []byte) (int, error) {
+	b := make([]byte, r.bufferSize)
+
+	if r.buf.Len() != 0 {
+		return r.copyToOut(buf)
```
urso: nit: what if `buf.Len()` == 1? In this case we would copy only one byte and return. It's not incorrect, but a little inefficient. Let's add a `TODO` comment at least, noting that we should try to collect at least `len(buf) - r.buf.Len()` bytes, if `len(buf) > r.buf.Len()`.

filebeat/reader/line/decoder.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	for {
+		start := r.encodedBuf.Len()
+		n, err := r.in.Read(b[start:])
+		if err != nil {
+			return len(buf[:n]), err
+		}
```
urso: this still looks like a potential bug. 

Let's assume before entering the loop we have this initial state:

```
encodedBuf.Len = 0
file-read-offset: 0
r.buffer_size: 1000 // len(b) == 1000
b[0:1000] = "\0..."
buf[0:1000] = "\0..."
```

The file state will contain 26 charatecters:
```
"abcdefghijklmnopqrstuvwxyz"
```

Given this state we already 

Now the read fails, but some bytes have been copied into the "internal" buffer b. Return values:
```
n = 10
err = "ups"
```
The new state is:

```
encodedBuf.Len = 0
file-read-offset: 10
b[0:10] = "abcdefghij"
b[10:1000] = "\0..."
buf[0:1000] = "\0..."
```

Next calling 'return' with this state we tell the caller `n` bytes have been written to buf -> return values are:

```
n = 10
err = "ups"
buf[0:10] = "\0..." // <- bug 1, we return contents, without having copied anythin
```

That is no content has been copied at all, yet we tell the next reader in the chain to process 10 characters.

Worse, if the next reader tries to read more, we have the initial state for the current function call at:

```
encodedBuf.Len = 0
file-read-offset: 10
r.buffer_size: 1000 // len(b) == 1000
b[0:1000] = "\0..."
buf[0:1000] = "\0..."
```

This time the read succeeds. As the filter read pointer is already at offset 16, we will read 16 bytes only: ->

```
n = 16
err = 0
b[0:16] = "klmnopqrstuvwxyz"
```

Assuming we can convert by just copying (all characters in lower ASCII range) the function return will be:

```
n = 16
err = nil
buf[0:16] = "klmnopqrstuvwxyz"
```

Due to the error handling we made the callee process the result byte sequence:
`"\0\0\0\0\0\0\0\0\0\0klmnopqrstuvwxyz"`, but the expected byte sequence should have been `"abcdefghijklmnopqrstuvwxyz"`.


Another potential bug arises we run the loop twice, with the first read succeeding and the second failing. Assuming `n=5` in the first run and `n=2` in the second run (plus error), then we will return `n=2, err="ups", buf[0:2]="ab"`. The other 5 bytes that have been writte from the file are ignored.

filebeat/reader/line/decoder.go
```diff
@@ -0,0 +1,166 @@
...
+
+		nBytes, nProcessed, err := r.conv(b[:start+n], buf)
+		if err != nil {
+			if err == transform.ErrShortSrc {
+				r.encodedBuf.Append(b[nProcessed:])
+				return nBytes, nil
+			}
+			return 0, err
+		}
+		return nBytes, nil
```
urso: The for loop always returns. In this case the for loop can be removed, no?

The idea of the for-loop was to continue reading if there is still some more space in `buf`.

filebeat/reader/line/decoder.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+}
+
+// msgSize returns the size of the encoded message on the disk
+func (r *decoderReader) msgSize(symlen []uint8, size int) (int, []uint8, error) {
```
urso: what does `size` mean? Is `size` the input in number of `rune`s or number of `byte`s? Given the loop itself, I expect `size` to be the number of bytes and the return value being the number of `rune`s.

What if `size` becomes negative?

filebeat/reader/line/decoder.go
```diff
@@ -0,0 +1,166 @@
...
+	return s
+}
+
+// conv converts encoded bytes into UTF-8 and produces a symlen array which
+// records the size of the encoded bytes and its converted size
+func (r *decoderReader) conv(in []byte, out []byte) (int, int, error) {
+	var err error
+	nProcessed := 0
+	decodedChar := make([]byte, 64)
+	r.symlen = make([]uint8, len(in))
```
urso: is there a chance `len(r.symlen) > 0` when entering `conv`?

filebeat/reader/line/decoder.go
```diff
@@ -0,0 +1,166 @@
...
+
+	i := 0
+	srcLen := len(in)
+	for i < srcLen {
+		j := i + 1
+
+		for j <= srcLen {
+			nDst, nSrc, err := r.decoder.Transform(decodedChar, in[i:j], false)
```
urso: ouch... checking `x/text`, there doesn't seem to be a `decodeRune` for any encoding. All encodings have the decode and copy functionality built right into the loop implemented by Transform :(
 
E.g. with `decodeRune` we could write the copy loop like this:

```
...
for i := 0; i < len(in); i++ {
  r, width, err := encoding.DecodeRune(in)
  if err != nil {
    ... // e.g. ErrShortSrc
  }

  uft8Width := utf8.RuneLen(r)
  if len(to) < utf8Width {
    ... // output buffer full, return
  }

  r.symLen[nProcessed] = width
  nProcessed++
  utf8.EncodeRune(to[nProcessed:], r)
}
...
```

We should extract the inner loop to decodeEncRune:

```
func decEncRune(decoder ..., in []byte) (rune, int, err) {
  for j := 0 ; j < len(in); j++ {
}
```

filebeat/reader/line/decoder.go
kvch: It is the number of converted bytes. I renamed the variables and added more comments.

filebeat/reader/line/decoder.go
kvch: Yes, this is the part of the code I am most worried about. It's complex and does not perform well. I would like to drop this algorithm in favour of the newer idea we discussed previously.

filebeat/reader/line/decoder.go
kvch: No, it's always retrieved by the scanner.

kvch:
I rebased the PR and addressed the review notes. Now I am focusing on the alternative "segmenting" algorithm we discussed. After that's complete I would like to compare the two solutions regarding complexity and performance.

filebeat/reader/readfile/decoder.go
```diff
@@ -0,0 +1,169 @@
...
+	// TODO add to the beginning of buf and then start decoding
+	// instead of returning
+	if r.buf.Len() != 0 {
+		return r.copyToOut(buf)
+	}
+
+	b := make([]byte, r.bufferSize)
+	start := r.encodedBuf.Len()
+	n, err := r.in.Read(b[start:])
+	if err != nil {
```
urso: If `n > 0`, then we must process all bytes from `b[start:start+n]`. Independent of the actual value of err.

filebeat/reader/readfile/decoder.go
```diff
@@ -0,0 +1,169 @@
...
+		return 0, err
+	}
+
+	if start > 0 {
+		enc, err := r.encodedBuf.Collect(start)
+		if err != nil {
+			return 0, err
+		}
+		r.encodedBuf.Reset()
+		b = append(enc, b[start:]...)
```
urso: more correct/efficient would be `b = append(enc, b[start:start+n]...)`. If `n < len(b[start:])`, then this append operation will add loads of '\0' to the end of b.

filebeat/reader/readfile/decoder.go
```diff
@@ -0,0 +1,169 @@
...
+	nProcessed := 0
+	decodedChar := make([]byte, 64)
+	r.symlen = make([]uint8, len(in))
+
+	i := 0
+	srcLen := len(in)
+	for i < srcLen {
+		j := i + 1
+
+		for j <= srcLen {
```
urso: if we use this approach, then this inner loop should be refactored into a `decodeRune` function/method (somewhere burried in the comments). Reading and understand this loop takes time. Code should be more skimmable at best.

filebeat/reader/readfile/scanner.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+}
+
+// Scan reads from the underlying decoder reader and returns decoded lines.
+func (s *lineScanner) scan() ([]byte, int, error) {
+	idx := s.buf.Index(s.separator)
+	for !separatorFound(idx) {
+		b := make([]byte, s.bufferSize)
+		n, err := s.in.read(b)
+		s.buf.Append(b[:n])
```
urso: another place where support for pre-allocating and accessing the available write-buffer in streambuf would be nice. Like:

```
s.buf.Ensure(s.bufferSize)
n, err := s.in.inread(s.buf.WriteBuffer())
```

Btw. we shoud not use streambuf here. The Reset method does set the read/write pointer to 0, but the underlying slice is not reset (buffer is reused). On big line/allocation can grow the buffer and waste memory we are never able to return.

kvch:
Closing as the solution in https://github.com/elastic/beats/pull/7940 performs better.

PR #7583, opened 2018-07-12
Nested fields for all types in common.Schema

jsoriano:
Allow to get nested fields from other types in `mapstriface` in common.Schema. It was already allowed on `string` and `interface{}`, added a test also to check the behaviour.

Not sure however if this is the intended behaviour.


jsoriano:
@ruflin wdyt?

ruflin:
+1 on this change. Tried to think of some edge cases that it could break but couldn't come up with any.

PR #7585, opened 2018-07-12
Update Filebeat module dev guide

webmat:
- Replace export_dashboard example to use the recommended `-yml` option
- Add a link to export_dashboard's doc

### DIFF
--- a/docs/devguide/modules-dev-guide.asciidoc
+++ b/docs/devguide/modules-dev-guide.asciidoc
@@ -74,16 +74,35 @@ is used as a title in the docs, so it's best to capitalize it.
 ==== _meta/kibana
 
 This folder contains the sample Kibana dashboards for this module. To create
-them, you can build them visually in Kibana and then run the following command:
+them, you can build them visually in Kibana and then export them with `export_dashboards`.
+
+The tool will export all of the dashboard dependencies (visualizations,
+saved searches) automatically.
+
+You can see various ways of using `export_dashboards` at <<export-dashboards>>.
+The recommended way to export them is to list your dashboards in your module's
+`module.yml` file:
+
+[source,yaml]
+----
+dashboards:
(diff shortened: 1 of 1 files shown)


kvch:
Could you please rebase this PR? The header problems should be fixed on master.

docs/devguide/modules-dev-guide.asciidoc
```diff
@@ -74,16 +74,35 @@ is used as a title in the docs, so it's best to capitalize it.
...
+
+Then run `export_dashboards` like this:
 
 [source,shell]
 ----
 $ cd dev-tools/cmd/dashboards
-$ make # if export_dashboard is not built
-$ ./export_dashboards -dashboard '{dashboard-id}' -output '../../../filebeat/module/{module}/_meta/kibana/default/dashboard'
+$ make # if export_dashboard is not built yet
+$ ./export_dashboards -yml '../../../filebeat/module/{module}/module.yml'
```
ruflin: I wonder if we should instead recommend to use `go run dev-tools/cmd/dashboards/export_dashboards ...` so make is not needed and it always uses the local platform.

kvch: At this point it is acceptable here, as `make` is not yet eliminated from module generation. Also, I think calling `./export_dashboards ...` is more elegant in the documentation.

webmat:
Rebase done and pushed. So we're good with keeping the `make` step and using the compiled `export_dashboards`? I also think it's nicer to have this much less to type ;-)

docs/devguide/modules-dev-guide.asciidoc
ruflin: Both works for me. The command is shorter but you need to get all the `../` right :-)

ruflin:
Both options work for me as long as it's working :-)

webmat:
Ok thanks. Merge at will when ready :-)

PR #7588, opened 2018-07-13
Note that python 2 is required for `generate.py`

jordansissel:
The `generate.py` script uses `raw_input` which is not available in Python 3 and fails if you try to run it. It works in Python 2 :)


ruflin:
Thanks for the fix. We are "trying" to keep everything also python 3 compatible but we often forget pieces :-(

PR #7599, opened 2018-07-13
Update go-ucfg to 0.6.1

urso:
Update fixes config unpacking if users overwrite settings from CLI, with
missing values. When using `-E key=` (e.g. in scripts defining potential
empty defaults via env variables like `-E key=${MYVALUE}`), an untyped
`nil`-values was inserted into the config. This untyped value will make
Unpack fail for most typed settings.


exekias:
jenkins, test this please

PR #7601, opened 2018-07-13
add kafkabeat2

arkady-emelyanov:
justsocialapps/kafkabeat is generally ok, here is [updated version of kafkabeat](https://github.com/arkady-emelyanov/kafkabeat) which supports two ways of handling Kafka messages: json and plain.

Published documentation describes both ways to deal with messages. Beat build on top of stable `6.3.1` release.


ruflin:
@arkady-emelyanov Thanks for the contribution.

PR #7608, opened 2018-07-16
Filebeat: Add patterns to support Kafka 1.1 logs

jsoriano:
Add patterns to support kafka 1.1 logs.

Existing patterns work pretty well for 1.1, but I found an issue with component parsing, closing bracket is not followed by comma, dot or colon, this causes that component is almost always set to `unknown`. It also caused in some scenarios that other bracket in the message is taken as the closing bracket of the component, incorrectly splitting the component and the log message.

This change adds a pattern for components so they cannot include a closing bracket, it also make optional the symbols after the component. Added also a test file that reproduced problems with logs in 1.1 version.

Documentation line about testing versions is fixed too.


jsoriano:

acchen97:
Thanks for this! @jsoriano for accuracy, should we also edit the 2.11 to 0.9 for Filebeat 6.1-6.3 docs as well?

jsoriano:
@acchen97 started #7617 to backport documentation changes to 6.x branches.

PR #7609, opened 2018-07-16
Cherry-pick #7573 to 6.3: Ensure metadata updates don't replace existing pod metrics

exekias:
Cherry-pick of PR #7573 to 6.3 branch. Original message: 

Before things change, `add_kubernetes_metadata` would replace everything
under `kubernetes.pod` when enriching events.

This is not what you want, especially if you are using it in conjunction
with the Kubernetes module, where you could smash pod metrics when
enriching.


exekias:
Tests are failing because `common.MustNewConfigFrom` is not available on this branch, will patch this to use `common.NewConfigFrom`

PR #7615, opened 2018-07-17
Add more detail topic about downloading docker images

dedemorton:
I wasn't completely satisfied with the language that we added in https://github.com/elastic/logstash/pull/9831, so I've made some changes. The original text was a bit verbose, overly passive, and  ambiguous in places. If you like the changes, we can get them into the other repos. 

* Removed the passive voice from the second paragraph and tightened up the language just a little.
* Removed the following sentence because I don't think it adds anything new: "For example, the Docker image can be retrieved with the following command."  If you think this is necessary, I can add it back in...but IMO, it is redundant.
* Broke the sentence that begins with "Alternatively, you can download..." into two sentences. The original sentence has a dangling modifier. Grammatically, it's saying that the Apache license is available from the docker page. Passed it by Gail, and she agreed that it was hard to parse.

I've also added the license content that wasn't in the Beats topic originally. Let me know if it's not completely accurate for Beats.

### DIFF
--- a/libbeat/docs/shared-docker.asciidoc
+++ b/libbeat/docs/shared-docker.asciidoc
@@ -2,7 +2,23 @@
 === Running {beatname_uc} on Docker
 
 Docker images for {beatname_uc} are available from the Elastic Docker
-registry. You can retrieve an image with a `docker pull` command.
+registry. The base image is https://hub.docker.com/_/centos/[centos:7].
+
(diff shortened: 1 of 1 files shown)


tbragin:
Thanks for taking a pass @dedemorton! LGTM.

As you suggest, it'd be great to get these same changes into the other places in the stack we are listing `docker pull` commands.

PR #7628, opened 2018-07-18
Basic support of ES GC metrics for jvm9

immon:
GC log format for JVM9 is more detailed than for JVM8.

Differences and possible improvements:
* To get cpu_times.* a corellation between log lines is required.
* Some GC metrics are available in jvm8 are not in jvm9
  (class_unload_time_sec, weak_refs_processing_time_sec, ...)
* heap.used_kb is empty, but it can be calculated as young_gen.used_kb +
  old_gen.size_kb
* GC phase times are logged in miliseconds vs seconds in jvm8


ruflin:
@immon Could you rebase on a recent version of master? CI fails because of some header issues which should have been fixed in the last days.

immon:
rebased, but CI failed on metricbeat part `Timeout while waiting for healthy docker-compose services: kubernetes`. @ruflin

immon:
jenkins test it

immon:
jenkins, test this

immon:
it's green now @ruflin

ruflin:
@immon Thanks. One thing I'm wondering is how this will work under load as we now have quite a list of grok patterns.

immon:
If we are going to visualize jvm memory usage we need to extract metrics from logs. Is there a better way to do it?
system/auth ingest pipeline use quite a few patterns too. Was it ever a performance issue?
/cc @inqueue

ruflin:
Performance testing is something we should add for all modules and don't have yet. We had an issue with the postgres grok patterns which was luckly discovered by the demo.elastic.co project.

I think it's not the number of pattern necessarly more if all the patterns are efficient and the only way to tell is the load tests I think. The other thing is that I hope we can use `dissect` much more in the future to do the first step of the parsing and only leave the heavy parts to grok.

Nothing you should worry about at the moment as I hope we can take care of it in generic way by having load tests for all modules so we see if one is standing out.

PR #7629, opened 2018-07-18

tsg:
There were no significant conflicts.


ruflin:
It will need a make update because of some versioned links in the docs.

PR #7631, opened 2018-07-18
Fix log message for Kibana beta state

ruflin:
From copy paste Kafka was in the log message instead of Kibana.


ycombinator:
jenkins, test this

PR #7635, opened 2018-07-18
Only fetch shard metrics from master node

ycombinator:
This PR makes it so that the `elasticsearch/shard` metricset only fetches information from the Elasticsearch node if that node is the master node.


ycombinator:
jenkins, test this

ycombinator:
@ruflin CI is green now.

PR #7636, opened 2018-07-18
Add null output for performance benchmarking purposes

vjsamuel:


vjsamuel:
@urso i made `bulk_max_size` configurable. i didnt quite get what the second request was. when we start up beats using `-e -v` the output plugin metrics are dumped to console. is that sufficient or do you need a way in which the metrics are always printed out once in say 30s?

urso:
I'd say bulk_max_size is good enough for now. I used to have an implementation (can't find it anymore) that writes the current rate and average event rate every second. When supressing any other logs, it used to overwrite the last stats on console. E.g. what `pv -Warl >/dev/null` does when using the console output.

libbeat/outputs/null/null.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+// Implement Outputer
+func (n *null) Close() error {
+	return nil
+}
+
+func (n *null) Publish(
+	batch publisher.Batch,
+) error {
```
urso: nit: we don't have many parameters here. No need for newlines for the parameter list.

jsoriano:
@vjsamuel would you like to follow up with latest proposal by @urso about adding the missing features to the pipeline used for stress testing?

PR #7638, opened 2018-07-18
Adding cluster_stats metricset for elasticsearch module

ycombinator:
This PR adds a `cluster_stats` metricset for the `elasticsearch` Metricbeat module. Only an **initial set** of fields are added by this PR.

### For follow up PRs
* Adding event generation for X-Pack Monitoring indices.
* Adding more fields, as necessary.


metricbeat/module/elasticsearch/cluster_stats/_meta/data.json
```diff
@@ -0,0 +1,55 @@
...
+    "beat": {
+        "hostname": "host.example.com",
+        "name": "host.example.com"
+    },
+    "elasticsearch": {
+        "cluster": {
+            "id": "njqU4EQaTROIDlWPeUMQyw",
+            "name": "elasticsearch",
+            "stats": {
+                "indices": {
```
ruflin: How does the cluster index data overlap with the `index` and `node_stats` metricset data?

metricbeat/module/elasticsearch/cluster_stats/data.go
```diff
@@ -0,0 +1,92 @@
...
+		return err
+	}
+
+	var event mb.Event
+	event.RootFields = common.MapStr{}
+	event.RootFields.Put("service.name", "elasticsearch")
+
+	event.ModuleFields = common.MapStr{}
+	event.ModuleFields.Put("cluster.name", info.ClusterName) // TODO cluster_name from cluster stats response could be used here?
+	event.ModuleFields.Put("cluster.id", info.ClusterID)     // TODO: Can we get this returned in cluster stats response itself?
```
ruflin: ++

I think both values are a must to be returned y cluster stats API

ycombinator: Created https://github.com/elastic/elasticsearch/issues/32205 to request cluster UUID from cluster stats API itself.

ycombinator: ... and put up PR as well: https://github.com/elastic/elasticsearch/pull/32206

metricbeat/module/elasticsearch/cluster_stats/_meta/data.json
ycombinator: * `index` metricset has `docs`, `segments`, and `store` which are also present in this metricset here.

* `node_stats` metricset has `docs`, `segments`, and `store`, which are also present in this metricset here.

* `node_stats` metricset also has `fs` and `jvm` stats, which are currently not present in this metricset here but are available from the Elasticsearch `GET _cluster/stats` API.

* This metricset here has `fielddata` which is absent in the `index` metricset but is available from the Elasticsearch `GET _stats` API. 

* This metricset here has `fielddata` which is absent in the `node_stats` metricset and also not available from the Elasticsearch `GET _nodes/_local` API. However it is available from the Elasticsearch `GET _nodes/_local/stats` API. Of course, the latter API is missing a bunch of _other_ fields that are present in the former API so either we call both or we ask the ES team to enhance the latter API's response to include all the fields we use from the former API's response as well.

ycombinator: I do need to fix the field names in this metricset here so they match up with the ones in other `elasticsearch` metricsets (e.g. I need to use `count` instead of `total`, etc.)

metricbeat/module/elasticsearch/cluster_stats/data.go
ruflin: Keep in mind that older versions of ES will not have it in, so in these cases we need to skip it.

ycombinator: Good point! Should we skip it or make the extra API call in that case (I realize that complicates the logic/organization of the code a fair bit)?

ruflin: Lets skip it to not add more complexity. New versions, new features :-)

metricbeat/module/elasticsearch/cluster_stats/_meta/data.json
ycombinator: Field names fixed in 6662787.

metricbeat/module/elasticsearch/cluster_stats/data.go
ycombinator: Fixed in ed74319b4.

metricbeat/module/elasticsearch/cluster_stats/_meta/data.json
ruflin: Should we report all the data we have in the index and node_stats metricset here too? Or should we focus on the data points that other metricsets don't have?

ycombinator: I think to start off we can only report the stats that are not already covered by other metricsets. Concretely, those would be `fielddata` and `shards` fields.

ruflin: ++

ycombinator: Fixed in 27ef3fb.

metricbeat/module/elasticsearch/cluster_stats/_meta/fields.yml
```diff
@@ -0,0 +1,53 @@
...
+          Total number of indices in cluster
+      - name: shards
+        type: group
+        description: >
+          Shard statistics
+        fields:
+        - name: count
+          type: long
+          description: >
+            Total number of shards in cluster
```
ruflin: Apply to all lines: I think normally we put a dot at the end for the docs.

metricbeat/module/elasticsearch/cluster_stats/_meta/test/cluster_stats.700.json
```diff
@@ -0,0 +1,147 @@
+{
```
ruflin: Can we also have one for 630?

ycombinator: Added in a5953bd.

ruflin:
@ycombinator Looks like this could be merged before the ES PR is merged as we do check if the field exists?

Could you rebase and directly squash the commits?

ycombinator:
Yes, good point @ruflin. I will fix up this PR now so it can be reviewed/merged soon.

ycombinator:
Alright, all squashed up into 69ee642 and ready for review/merge. Thanks @ruflin!

ycombinator:
jenkins, test this

Issue #6215, opened 2018-01-29
Update beats documentation on Kafka 1.0 compatibility


ppf2:
The current documentation does not mention Kafka 1.0 in the compatibility section.  From a quick chat with @urso , Kafka 1.0 should work (even though we currently still test with 0.10 images).   With Kafka 1.0 being out for few months now, it will be helpful to start testing with 1.0 and/or update the compatibility section of the beats doc accordingly.

yfoelling:
Kafka 1.0 is out for a while now. 
Filebat should support it too!

PR #7639, opened 2018-07-18
Update compatibility statement in Kafka docs

dedemorton:
@urso Should this say, "This output works with Kafka 0.8.2.0 through 1.1.0" (instead of 1.0.0). It wasn't clear to me whether we are saying that we support (and have tested) against 1.1.0. I think it looks a little weird to have 1.1.0 as a valid setting if we don't test against it. WDYT?

Closes https://github.com/elastic/beats/issues/6215

### DIFF
--- a/libbeat/docs/outputconfig.asciidoc
+++ b/libbeat/docs/outputconfig.asciidoc
@@ -667,7 +667,8 @@ NOTE: Events bigger than <<kafka-max_message_bytes,`max_message_bytes`>> will be
 
 ==== Compatibility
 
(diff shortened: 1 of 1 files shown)


urso:
> I think it looks a little weird to have 1.1.0 as a valid setting if we don't test against it

TBH, the same goes for kafka 0.8.2 and any other version ut 1.0.0. For testing we somewhat rely on saram testing. Docs and API states 0.8.2 is still "somewhat supported". But checking their travis CI, only 0.11 upwards are tested: https://travis-ci.org/Shopify/sarama

PR #7645, opened 2018-07-19
Add support for alias field type to fields.yml

ruflin:
In https://github.com/elastic/elasticsearch/issues/23714 Elasticsearch implemented the alias field type. This can be used in fields.yml as following:

```
- name: a.b
  type: alias
  path: a.c
```

`a.b` will be the alias for `a.c`.

### DIFF
--- a/libbeat/common/field.go
+++ b/libbeat/common/field.go
@@ -52,6 +52,7 @@ type Field struct {
 	DocValues             *bool       `config:"doc_values"`
 	CopyTo                string      `config:"copy_to"`
 	IgnoreAbove           int         `config:"ignore_above"`
+	AliasPath             string      `config:"path"`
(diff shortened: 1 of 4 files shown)


andrewkroh:
Would collocating the alias with the original field make sense?

```
- name: a.c
  type: keyword
  alias: a.b
```

`a.b.` will be an alias of `a.c`

ruflin:
If we do that, `alias` would have to be an array as multiple alias can point to the same field.

I would prefer to keep the alias definition local as the alias definition is likely going to be inside a metricset / dataset and the original field is defined in libbeat.

Also in case we add it to the original field and alias supports additional features in the future, not sure how we would deal with it if we have all alias defined in one place.

PR #7648, opened 2018-07-20
Additional HB Tests + Improved mapval

andrewvc:
This adds a number of new tests to heartbeat

1. Testing the full range of HTTP codes
2. Testing both connection refused *and* unroutable endpoints for both TCP and HTTP
3. Makes various ergonomic and functional improvements to mapval discovered in the course of writing the above tests.
4. Adds the `freeport` dep via govendor. This nifty little lib will spit out a free local port you can use for testing.
5. Changes the signature of `mapvaltest.Test` to take `Schema` and expected `MapStr` as separate arguments. This reduces a lot of typos and makes tests easier to read.

I attempted to add some ICMP tests, but that will be a little tricky since you need escalated privileges. There's certainly a good way to add those, but I think that's deserving of its own PR.

### DIFF
--- a/heartbeat/monitors/active/http/http_test.go
+++ b/heartbeat/monitors/active/http/http_test.go
@@ -24,6 +24,8 @@ import (
 
 	"github.com/stretchr/testify/require"
 
+	"fmt"
+
 	"github.com/elastic/beats/heartbeat/hbtest"
 	"github.com/elastic/beats/heartbeat/monitors"
 	"github.com/elastic/beats/libbeat/beat"
@@ -32,12 +34,10 @@ import (
 	"github.com/elastic/beats/libbeat/testing/mapvaltest"
 )
 
-func checkServer(t *testing.T, handlerFunc http.HandlerFunc) (*httptest.Server, beat.Event) {
-	server := httptest.NewServer(handlerFunc)
-	defer server.Close()
-
+func testRequest(t *testing.T, url string) beat.Event {
 	config := common.NewConfig()
-	config.SetString("urls", 0, server.URL)
+	config.SetString("timeout", 0, "1")
+	config.SetString("urls", 0, url)
 
 	jobs, err := create(monitors.Info{}, config)
 	require.NoError(t, err)
@@ -47,30 +47,40 @@ func checkServer(t *testing.T, handlerFunc http.HandlerFunc) (*httptest.Server, 
 	event, _, err := job.Run()
 	require.NoError(t, err)
 
+	return event
+}
+
+func checkServer(t *testing.T, handlerFunc http.HandlerFunc) (*httptest.Server, beat.Event) {
+	server := httptest.NewServer(handlerFunc)
+	defer server.Close()
+
+	event := testRequest(t, server.URL)
+
 	return server, event
 }
 
-func httpChecks(urlStr string, statusCode int) mapval.Validator {
+// The minimum response is just the URL. Only to be used for unreachable server
+// tests.
+func httpBaseChecks(url string) mapval.Validator {
 	return mapval.Schema(mapval.Map{
-		"http": mapval.Map{
-			"url": urlStr,
-			"response.status_code":   statusCode,
-			"rtt.content.us":         mapval.IsDuration,
-			"rtt.response_header.us": mapval.IsDuration,
-			"rtt.total.us":           mapval.IsDuration,
-			"rtt.validate.us":        mapval.IsDuration,
-			"rtt.write_request.us":   mapval.IsDuration,
-		},
+		"http.url": url,
 	})
 }
 
-func badGatewayChecks() mapval.Validator {
-	return mapval.Schema(mapval.Map{
-		"error": mapval.Map{
-			"message": "502 Bad Gateway",
-			"type":    "validate",
-		},
-	})
+func respondingHTTPChecks(url string, statusCode int) mapval.Validator {
+	return mapval.Compose(
+		httpBaseChecks(url),
+		mapval.Schema(mapval.Map{
+			"http": mapval.Map{
+				"response.status_code":   statusCode,
+				"rtt.content.us":         mapval.IsDuration,
+				"rtt.response_header.us": mapval.IsDuration,
+				"rtt.total.us":           mapval.IsDuration,
+				"rtt.validate.us":        mapval.IsDuration,
+				"rtt.write_request.us":   mapval.IsDuration,
+			},
+		}),
+	)
 }
 
 func TestOKJob(t *testing.T) {
@@ -82,8 +92,8 @@ func TestOKJob(t *testing.T) {
 		t,
 		mapval.Strict(mapval.Compose(
 			hbtest.MonitorChecks("http@"+server.URL, server.URL, "127.0.0.1", "http", "up"),
-			hbtest.TCPChecks(port),
-			httpChecks(server.URL, http.StatusOK),
+			hbtest.RespondingTCPChecks(port),
+			respondingHTTPChecks(server.URL, http.StatusOK),
 		))(event.Fields),
 	)
 }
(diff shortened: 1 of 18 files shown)


heartbeat/monitors/active/http/http_test.go
```diff
@@ -24,6 +24,10 @@ import (
 
 	"github.com/stretchr/testify/require"
 
+	"fmt"
```
ruflin: goimport fun

vendor/vendor.json
```diff
@@ -1222,6 +1222,12 @@
 			"revision": "653207bc29a6d2d62b5d4f55b596467cb715a128",
 			"revisionTime": "2017-03-27T18:58:03Z"
 		},
+		{
+			"checksumSHA1": "kZRhErakejBG0U2e8D+Ap/Djje8=",
```
ruflin: I assume you will need to run `make update` on the top level so the NOTICE.txt file is updated.

heartbeat/monitors/active/http/http_test.go
```diff
@@ -47,59 +48,198 @@ func checkServer(t *testing.T, handlerFunc http.HandlerFunc) (*httptest.Server,
...
+				)),
+				event.Fields,
+			)
+		})
+	}
+}
+
+func TestConnRefusedJob(t *testing.T) {
+	ip := "127.0.0.1"
+	port := uint16(freeport.GetPort())
```
ruflin: I wonder if you need a library here or you could just set :0 as port and it would automatically give you an available port.

heartbeat/monitors/active/tcp/tcp_test.go
```diff
@@ -60,14 +74,47 @@ func TestUpEndpoint(t *testing.T) {
...
 			}),
-		))(event.Fields),
+		)),
+		event.Fields,
+	)
+}
+
+func TestConnectionRefusedEndpointJob(t *testing.T) {
+	ip := "127.0.0.1"
+	port := uint16(freeport.GetPort())
```
ruflin: See comment above.

libbeat/common/mapval/results.go
```diff
@@ -19,24 +19,40 @@ package mapval
...
 // They are a flattened map (using dotted paths) of all the values []ValueResult representing the results
 // of the IsDefs.
-type Results map[string][]ValueResult
+type Results struct {
+	Fields map[string][]ValueResult
+	// Lazy cache to prevent repeat computation
+	cache *resultsCache
+}
+
+func MakeResults() Results {
```
ruflin: Normally we call such methods `NewResults`

libbeat/common/mapval/is_defs.go
```diff
@@ -21,6 +21,8 @@ import (
 	"fmt"
 	"time"
 
+	"strings"
```
ruflin: goimports fun

libbeat/common/mapval/results.go
```diff
@@ -71,23 +87,34 @@ func (vre ValueResultError) Error() string {
 
 // Errors returns a list of error objects, one per failed value validation.
 func (r Results) Errors() []error {
-	var errors []error
+	if r.cache.errors != nil {
```
ruflin: Could it be that between to `Errors` call the data inside changed?

libbeat/common/mapval/results.go
```diff
@@ -19,24 +19,40 @@ package mapval
...
+}
+
 // Results the results of executing a schema.
 // They are a flattened map (using dotted paths) of all the values []ValueResult representing the results
 // of the IsDefs.
-type Results map[string][]ValueResult
+type Results struct {
+	Fields map[string][]ValueResult
+	// Lazy cache to prevent repeat computation
+	cache *resultsCache
```
ruflin: Do we really need this cache?

libbeat/testing/mapvaltest/mapvaltest.go
```diff
@@ -26,12 +26,25 @@ import (
 
 	"github.com/stretchr/testify/assert"
 
+	"github.com/davecgh/go-spew/spew"
```
ruflin: goimports fun

libbeat/common/mapval/is_defs.go
andrewvc: Gah! I need to add a pre-commit hook. That's the only way I'll learn.

libbeat/common/mapval/results.go
andrewvc: ++

vendor/vendor.json
andrewvc: Yes! Done in latest push

heartbeat/monitors/active/http/http_test.go
andrewvc: So, yes and no. Yes, that's exactly what freeport does, but if we didn't use it, we'd have to write it ourselves. The logic is pretty simple if you look at [freeport.go](https://github.com/phayes/freeport/blob/master/freeport.go), but I guess I don't see the point in rewriting it ourselves.

As a reminder, I'm essentially doing that in most tests, freeport is useful because I want an *unused* port, that will give me a TCP RST packet so I can see what the response is to a conn refused.

In the regular, server is up case, I leave the port unspecified and retrieve from the server object the port it decided to use.

andrewvc:
I *think* I've addressed all PR comments.

I also removed the dump `Cache` object from `mapval.Results`. I *did* convert Results from a value object to a reference object, added the field `Valid bool`, and dropped caching the `[]error` of errors. It makes sense to just default `Valid` to `true`, and flip it to `false` the first time someone records an error from an efficiency standpoint. The `Results` are only used in an immutable way, though they are built up from mutable pieces. There is no use case for editing that object that I can think of.

heartbeat/monitors/active/http/http_test.go
urso: freeport only works because the net package uses `SO_REUSEADDR`. It has potential races, but for testing the approach is good enough I'd say.

Still, I'm no big fan if single-short-function dependencies ;)
It would be nice to have a more low-level common test support lib, providing this kind of function itself.

heartbeat/monitors/active/tcp/tcp_test.go
```diff
@@ -60,14 +76,47 @@ func TestUpEndpoint(t *testing.T) {
...
+func TestUnreachableEndpointJob(t *testing.T) {
+	ip := "203.0.113.1"
+	port := uint16(1234)
+	event := testTCPCheck(t, ip, port)
+
+	mapvaltest.Test(
+		t,
+		mapval.Strict(mapval.Compose(
+			tcpMonitorChecks(ip, ip, port, "down"),
+			hbtest.ErrorChecks(fmt.Sprintf("dial tcp %s:%d: i/o timeout", ip, port), "io"),
```
urso: have you considered to introduce `ErrorChecksf`?

```
func ErrorCheckfs(str strings, vs ...interface{}) {
```

libbeat/common/mapval/is_defs.go
```diff
@@ -30,6 +31,29 @@ var KeyPresent = IsDef{name: "check key present"}
 // KeyMissing checks that the given key is not present defined.
 var KeyMissing = IsDef{name: "check key not present", checkKeyMissing: true}
 
+// IsStringContaining validates that the the actual value contains the specified substring.
+func IsStringContaining(needle string) IsDef {
```
urso: how about `IsStringMatching`?

Using the FP approach we can also provide predicates:

```
func IsStringWith(pred func(str string) bool) IsDef {
```

heartbeat/monitors/active/tcp/tcp_test.go
andrewvc: That would clearly be better in this test, but I'd like to avoid that at the moment. Currently the signature is:

`func ErrorChecks(msg interface{}, errType string) mapval.Validator `

`msg` is not `string`, but rather `interface()` so that a user can specify either a string value or a `mapval.IsDef`. While in this specific case `ErrorChecksfs` would be better, I think it's better to have fewer methods that are more generic here. I don't think it'd be used enough to justify the extra surface area.

libbeat/common/mapval/is_defs.go
andrewvc: I think this is a great idea, but we don't need it yet. I'd prefer to hold off until we have running code that uses this. If we were releasing `mapval` as a separate library I'd agree in a heartbeat *rimshot*, but I prefer to have features come from real-world extraction rather than prediction.

It's not a big deal though, since we will probably use it sooner rather than later, so if you feel it's important I'm glad to add it :)

heartbeat/monitors/active/http/http_test.go
ruflin: +1 on not adding a dependency for it. Had a look at Freeport and it is only 15 lines of code: https://github.com/phayes/freeport/blob/master/freeport.go#L8

To move forward: we could keep it in this PR and remove it in a follow up PR where we introduce our own helper?

Issue #7643, opened 2018-07-19
Duplicate entries for Metricbeat module docs


acchen97:
In the Metricbeat module docs, it looks like kvm, munin, and uwsgi have duplicates. kvm and munin also have duplicate metricset docs.

https://www.elastic.co/guide/en/beats/metricbeat/6.3/_munin_node_metricset.html

/cc @dedemorton

ruflin:
PR to fix this opened here: https://github.com/elastic/beats/pull/7650

PR #7650, opened 2018-07-20
Fix duplicated module headers

ruflin:
Closes https://github.com/elastic/beats/issues/7643

### DIFF
--- a/metricbeat/module/kubernetes/apiserver/_meta/docs.asciidoc
+++ b/metricbeat/module/kubernetes/apiserver/_meta/docs.asciidoc
@@ -1,0 +1,1 @@
+This is the `apiserver` metricset of the Kubernetes module.
--- a/metricbeat/docs/modules/kubernetes.asciidoc
+++ b/metricbeat/docs/modules/kubernetes.asciidoc
@@ -89,6 +89,8 @@ This module supports TLS connection when using `ssl` config field, as described 
 
 The following metricsets are available:
 
+* <<metricbeat-metricset-kubernetes-apiserver,apiserver>>
+
 * <<metricbeat-metricset-kubernetes-container,container>>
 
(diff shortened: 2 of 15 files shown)


ruflin:
jenkins, test this

ycombinator:
This fixes the duplication of the **module** doc. But it does not yet fix the duplication of the **metricset** docs. For example I now only see one doc for "Munin module" but two (duplicated) docs under it for the node metricset.

screen shot 2018-07-20 at 6 46 47 am

To fix the duplication in the metricset docs I think you also need to edit `{module}/{metricset}/_meta/docs.asciidoc` and remove the `== {module} {metricset} MetricSet` line from the top.

metricbeat/module/kubernetes/apiserver/_meta/docs.asciidoc
```diff
@@ -0,0 +1 @@
+This is the `apiserver` metricset of the Kubernetes module.
```
ruflin: @exekias I wonder if you left this out on purpose?

exekias: probably not, :+1: to having it

ycombinator:
Sorry @ruflin but I just happened to notice while clicking around that the `traefik` module has the same duplication issue with its metricset. Mind fixing that one in this PR too?

ruflin:
@ycombinator Fixed. I also started to wonder why the new modules / metricsets have this issue and realised it's a left over in the generator we forgot to remove when we started to auto generate the titles. I now also cleaned up the generator.

ycombinator:
LGTM.

dedemorton:
Removing the `needs_backport` label because the changes are already in 6.x and 6.4 (I guess they were part of the auto merge.) I've created a separate backport PR for 6.3 (#7650), so I think we're covered.

PR #7651, opened 2018-07-20
Filebeat character decoding optimizations

kvch:
`Converter` interface is introduced to handle conversions from one encoding to UTF-8.

```golang
type Converter interface {
    Collect(out []byte) (int, error)                        // collects bytes converted to UTF-8
    Convert(in, out []byte) (int, int, error)               // converts the bytes in in and returns them in out
    MsgSize(symlen []uint8, size int) (int, []uint8, error) // calculates the lenght of the original message
    GetSymLen() []uint8                                     // returns the array storing size info
}
```

The existing general converter is extracted as a `DefaultConverter` which takes care of most encodings. However, where possible, optimizations were added.
Encodings which can be optimized:
- [x] plain
- [x] UTF-8
- [ ] UTF-16

TODO
- [ ] add more comments
- [ ] revisit dir structure

Depends on #7582


kvch:
Closing as the solution in https://github.com/elastic/beats/pull/7940 performs better.

PR #7655, opened 2018-07-20
Increase kafka version in tests to 1.1.1

urso:
Increase kafka version in test environment.


urso:
The metricbeat failure is unrelated to this change.

PR #7658, opened 2018-07-20
Release raid and socket metricset from system module as GA

ruflin:


metricbeat/module/system/raid/_meta/docs.asciidoc
```diff
@@ -1,3 +1,5 @@
 === system raid MetricSet
```
ycombinator: This line seems to be the cause of the failing doc generation task in CI. I think it should be removed.

ruflin: agree. will remove it.

Issue #7649, opened 2018-07-20
Support OpenStack in add_cloud_metadata Processor


herver:
Filebeat version: 6.3.1
Environment: OpenStack Queens

We run filebeat in our Openstack infrastructure and it'd be nice to add support for Openstack in the `cloud_metadata` processor.

I thought it would work out of the box with the EC2 provider because Openstack provides an AWS-compatible metadata service (https://docs.openstack.org/nova/latest/user/metadata-service.html) but it appears that the EC2 processor uses a different endpoint to get the data (https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-identity-documents.html)

I can contribute the provider for OpenStack, but would it make more sense to "fix" the EC2 provider ?

ruflin:
I would suggest to add an openstack provider that probably reuses 99% of the code from the AWS provider. Otherwise I worry users that don't know the two are almost identical they get confused that AWS is loaded but they use open stack.

Would be great if you contribute it.

If we build a separate provider, does AWS still need some fixing?

PR #7663, opened 2018-07-20
Add Openstack Nova cloud_metadata Processor (#7649)

herver:
Using the EC2 compatible metadata endpoint because it exposes more useful information that the native "openstack" endpoint.

### DIFF
--- a/libbeat/processors/add_cloud_metadata/provider_openstack_nova_test.go
+++ b/libbeat/processors/add_cloud_metadata/provider_openstack_nova_test.go
@@ -1,0 +1,91 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package add_cloud_metadata
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/elastic/beats/libbeat/beat"
+	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/logp"
+)
+
+func initOpenstackNovaTestServer() *httptest.Server {
+	return httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.RequestURI == osMetadataInstanceIDURI {
+			w.Write([]byte("i-0000ffac"))
+			return
+		}
+		if r.RequestURI == osMetadataInstanceTypeURI {
+			w.Write([]byte("m1.xlarge"))
+			return
+		}
+		if r.RequestURI == osMetadataHostnameURI {
+			w.Write([]byte("testvm01.stack.cloud"))
+			return
+		}
+		if r.RequestURI == osMetadataZoneURI {
+			w.Write([]byte("az-test-2"))
+			return
+		}
+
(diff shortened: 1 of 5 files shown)


andrewkroh:
> Using the EC2 compatible metadata endpoint because it exposes more useful information that the native "openstack" endpoint.

So if both metadata services (AWS and OpenStack) are available what happens? IIRC the first one to return a response wins. Perhaps it now needs a way to configure what providers are enabled.

herver:
> So if both metadata services (AWS and OpenStack) are available what happens? IIRC the first one to return a response wins. Perhaps it now needs a way to configure what providers are enabled.

Indeed... I can use the native Openstack data if preferred but this endpoint isn't on par with the EC2-compatible one (it doesn't provide the machine-type for instance)

ruflin:
I assume the above would happen if OpenStack is run on AWS? Would we expect both to return the same data?

Could you also add a changelog entry?

herver:
I don't really the point of running Openstack on AWS, but if that was the case, I assume that the network component of Openstack would provide some overlay network and thus the metadata endpoint would be the Openstack one.

I think that what @andrewkroh thinks about is if we run on AWS, then both metadata providers will be "active" (as both endpoint will return valid responses).

For the changelog entry, shall I add them to `libbeat/docs/release-notes/6.0.0.asciidoc` or `CHANGELOG.asciidoc` a new file/somewhere else ?

ruflin:
For the changelog, please add it to `CHANGELOG.asciidoc`.

Thanks for the explanation of AWS / OpenStack mix. I assume that the data format was the same but that the url to the API endpoints would be different so only 1 would work. But that's not the case?

herver:
Indeed, both providers use the exact same endpoints :crying_cat_face: 

So this morning I spun an instance on EC2 just to see which metadata is added and it appears that only the first matching provider is used (I didn't dig in the code to find where)

 * On EC2
```
2018-07-23T08:53:31.821Z INFO instance/beat.go:273 Setup Beat: filebeat; Version: 7.0.0-alpha1
2018-07-23T08:53:31.825Z INFO add_cloud_metadata/add_cloud_metadata.go:323 add_cloud_metadata: hosting provider type detected as ec2, metadata={"availability_zone":"eu-west-1a","instance_id":"i-0bbe64bec87a6044e","machine_type":"t2.micro","provider":"ec2","region":"eu-west-1"}
```

 * On Openstack
```
2018-07-23T10:51:50.544+0200    INFO    instance/beat.go:273    Setup Beat: filebeat; Version: 7.0.0-alpha1
2018-07-23T10:51:51.958+0200    INFO    add_cloud_metadata/add_cloud_metadata.go:323    add_cloud_metadata: hosting provider type detected as openstack, metadata={"availability_zone":"xxx-az-c","instance_id":"i-00011a84","instance_name":"test--998d932195.mycloud.tld","machine_type":"m2.large","provider":"openstack"}
```

libbeat/processors/add_cloud_metadata/provider_openstack_nova_test.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/stretchr/testify/assert"
+
+	"github.com/elastic/beats/libbeat/beat"
+	"github.com/elastic/beats/libbeat/common"
+	"github.com/elastic/beats/libbeat/logp"
+)
+
+func initOpenstackNovaTestServer() *httptest.Server {
+	return httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.RequestURI == "/2009-04-04/meta-data/instance-id" {
```
ruflin: You could use here and below the defined constants from above I think

andrewkroh:
I thought of one more thingthat needs updated. In 
https://github.com/herver/beats/blob/8c69672ad222ccdefb0af53f68390fde31786717/libbeat/docs/processors-using.asciidoc#add-cloud-metadata there is a list of cloud providers and a list of sample JSON output.

herver:
@andrewkroh thanks for the pointer, I've added the example to the doc as well !

andrewkroh:
jenkins test this please

kmbulebu:
I'm considering opening an issue and accompanying merge request. I believe the AWS compatible instance id should have been stored under a key identifying it as such. OpenStack users do not work with the EC2 compatible instance ID and will find it foreign. It doesn't match the ID of the API, CLI, or Web Console. 

I'm proposing moving the EC2 compatible instance id to a key named ec2_instance_id, and exposing the OpenStack instance_id under instance_id. However, I realize this would be a potentially breaking change for existing users. Any suggestions for handling the naming conflict? I could name the OpenStack instance id, 'id'. 

While I'm at it, I'll add some other fields such as project id.

PR #7664, opened 2018-07-20
Report kibana_settings to X-Pack Monitoring

### LINKED ISSUE #7621: Index `kibana_settings` docs for Kibana X-Pack Monitoring
As a follow up to https://github.com/elastic/kibana/pull/20577 and as a part of making progress on https://github.com/elastic/beats/issues/7035, we need a `settings` metricset in the `kibana` metricbeat module. This metricset would poll the [Kibana settings API](https://github.com/elastic/kibana/issues/20903) and ship its results into `.kibana-monitoring-6-*` indices provided `xpack.enabled: true` is set in the module's configuration.

ycombinator:
Resolves #7621.

Depends on https://github.com/elastic/kibana/pull/21100.

X-Pack Monitoring of Kibana requires two types of documents in the `.monitoring-kibana-*` indices: `kibana_stats` and `kibana_settings`. We made Metricbeat's `kibana/stats` metricset index `kibana_stats` documents into `.monitoring-kibana-*` in #7525. This PR makes the same metricset index `kibana_settings` documents into `.monitoring-kibana-*`.

### To test this PR
The idea is that metricbeat (specifically the kibana/stats metricset with `xpack.enabled: true`) will create exactly the same documents in `.monitoring-kibana-*` indices as Kibana's bulk uploader does today. 

1. Start up Elasticsearch (with the default Basic license is okay)
2. Start up Kibana (with the default Basic license is okay), checked out from `master`.
3. Enable Monitoring in Elasticsearch (via the cluster setting `xpack.monitoring.collection.enabled: true`). You can do this via [Elasticsearch's Cluster Update Settings API](https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-update-settings.html) or by clicking the "Turn on Monitoring" button in the Monitoring UI in Kibana.
4. Let Kibana run for ~20 seconds so a few documents are indexed into `.monitoring-kibana-*`.
5. Stop Kibana
6. From `.monitoring-kibana-*`, retrieve a document for `type = kibana_settings`
7. Delete the `.monitoring-kibana-*` indices.
8. In `kibana.yml`, set `xpack.monitoring.kibana.collection.enabled: false`. This will ensure that Kibana is no longer directly shipping its monitoring metrics to Elasticsearch.
9. Start up Kibana again, this time checked out from https://github.com/elastic/kibana/pull/21100.
10. Enable the `kibana` module in metricbeat: `./metricbeat modules enable kibana`.
11. In `modules.d/kibana.yml`, add the `stats` metricset and set `xpack.enabled: true`. Concretely, your `modules.d/kibana.yml` should look something like this:

      ```yaml
      - module: kibana
         metricsets:
         - stats
         period: 10s
         hosts: ["localhost:5601"]
         #basepath: ""
         #username: "user"
         #password: "secret"
         xpack.enabled: true
       ```

     Additionally, if your Kibana URLs have a basepath prefix, say `foo`, make sure to un-comment the `basepath` setting in the `kibana` module YAML file and set it's value to `"foo"`.

12. Start metricbeat.
13. Let metricbeat run for ~20 seconds so a few documents are indexed into `.monitoring-kibana-*`.
14. Stop metricbeat
15. From `.monitoring-kibana-*`, retrieve a document for `type = kibana_settings`.
16. Using a tool such as http://www.jsondiff.com, compare the documents indexed by Kibana's with those indexed by metricbeat. Verify that their structures are identical (same fields, not necessarily same values), except for these known and expected differences:
    1. Only Metricbeat-indexed documents are expected to contain the fields `@timestamp`, `beat`, `host`, and `metricset`. These are "standard" fields added by beats and metricbeat and don't have an adverse impact since they are additive.
    3. Only Kibana-indexed documents are expected to contain the field `source_node`. This field is used for debugging purposes only and not actually consumed by either the Monitoring UI or Telemetry feature in Kibana.


metricbeat/module/kibana/stats/_meta/test/settings.700.json
```diff
@@ -0,0 +1,18 @@
+{
```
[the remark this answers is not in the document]
ycombinator: The settings API in Kibana is not expected to exist until 6.5. Once it has been implemented for 6.5, I can add a `settings.650.json` file here. Should I do that?

Is your concern about how Metricbeat >= 6.5.0 will react when running with Kibana < 6.5.0?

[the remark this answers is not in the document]
ruflin: Yes. Seems like we should get the version number of KB on the first request and make decisions based on it.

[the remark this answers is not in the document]
ycombinator: Actually I have a question about this. Prior to 6.4.0 there is no `GET /api/stats` in Kibana either. Does that mean we should be checking the version of Kibana before making _that_ call too?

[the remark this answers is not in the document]
ruflin: Yes, I think we will need such a check or at least provide a good error message.

I think we could check the version once during `New` and just assume it stays the same over the lifetime of the metricset for now. So we would only do it once.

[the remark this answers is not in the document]
ycombinator: I'm going to add this version check in `New` in a separate PR since it affects code that has already been merged into `master` (#7525). I will rebase this PR on top of my new PR.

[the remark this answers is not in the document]
ycombinator: Version check PR: https://github.com/elastic/beats/pull/7697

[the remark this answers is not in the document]
ycombinator: Implemented version check for Kibana settings API in 2eaa84a.

ycombinator:
@ruflin The Kibana PR that this PR was dependent upon has now been merged. So this PR is ready for a final review/merge.

NOTICE.txt
```diff
@@ -171,6 +171,21 @@ COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
 IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
 CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
+--------------------------------------------------------------------
+Dependency: github.com/coreos/go-semver
```
ruflin: Is this package still somewhere used?

metricbeat/module/kibana/stats/data_xpack.go
```diff
@@ -105,61 +105,91 @@ var (
...
+	}
+
+	return "kibana_settings", clusterUUID, kibanaSettingsFields.(map[string]interface{}), nil
+}
+
+func eventMappingXPack(r mb.ReporterV2, intervalMs int64, now time.Time, content []byte, dataParserFunc dataParser) error {
+	var data map[string]interface{}
+	err := json.Unmarshal(content, &data)
+	if err != nil {
+		r.Error(err)
```
ruflin: Should we report an error in x-pack or just ignore it? I would opt for ignoring for now as it ends up in an other index and we will not make use of it.

metricbeat/module/kibana/stats/stats.go
```diff
@@ -63,22 +67,22 @@ func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
...
 	}
 
-	http, err := helper.NewHTTP(base)
+	statsHTTP, err := helper.NewHTTP(base)
 	if err != nil {
 		return nil, err
 	}
 
-	kibanaVersion, err := kibana.GetVersion(http, statsPath)
+	kibanaVersion, err := kibana.GetVersion(statsHTTP, statsPath)
```
ruflin: One thing I just realised that bugs me a bit with this implementation is that it requires Kibana to be up and running normally when Metricbeat is started. Assuming Kibana is flaky during the moment the check happens, the metricset will not be running but error out even though Kibana could be totally fine again 1 sec later.

Not for this PR but thought worth mentioning.

metricbeat/module/kibana/stats/stats.go
```diff
@@ -101,17 +131,35 @@ func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
 // It returns the event which is then forward to the output. In case of an error, a
 // descriptive error must be returned.
 func (m *MetricSet) Fetch(r mb.ReporterV2) {
-	content, err := m.http.FetchContent()
+	intervalMs := m.Module().Config().Period.Nanoseconds() / 1000 / 1000
```
ruflin: Could we have this code inside the `x-pack` part as it's only used there I think? Perhaps we can find a way to only if that xPackEnabled check once and run instead also in fetchStats. Why not move this to line 151 with the fetchSettings?

metricbeat/module/kibana/stats/stats.go
```diff
@@ -101,17 +131,35 @@ func New(base mb.BaseMetricSet) (mb.MetricSet, error) {
...
+		eventMappingStatsXPack(r, intervalMs, now, content)
 	} else {
 		eventMapping(r, content)
 	}
+}
+
+func (m *MetricSet) fetchSettings(r mb.ReporterV2, intervalMs int64, now time.Time) {
+	content, err := m.settingsHTTP.FetchContent()
+	if err != nil {
+		r.Error(err)
```
ruflin: Should we report an error? I think we will not make use of it.

NOTICE.txt
ycombinator: It shouldn't be any more. This line probably got re-introduced during a rebase or something. I will remove it. Thanks for catching!

metricbeat/module/kibana/stats/stats.go
ycombinator: I'm not sure I fully understood your comment but I took a stab at it in 3afa1a2a7. Take a look and let me know what you think. Thanks!

metricbeat/module/kibana/stats/stats.go
ycombinator: Good point. I suppose we could move this check (and the code dependent on it) into `Fetch` but that might be too expensive to run every fetch cycle?

ycombinator: Maybe we need another metricset lifecycle method in addition to `New()` and `Fetch()`, say `Ping()` (I know, not a great name), which runs periodically but less frequently than `Fetch()`. The period can be defined per-metricset in that metricset's `New()` function. The code in `Ping()` can add/overwrite info in the metricset's struct so the code in `Fetch()` can use it. Thoughts?

ruflin: What about using fetch but do it only once? I remember we do something like that in other cases.

metricbeat/module/kibana/stats/stats.go
ruflin: LGTM

metricbeat/module/kibana/stats/stats.go
ruflin: Merged as it's something we should discuss as a follow up.

PR #7670, opened 2018-07-21
Generate fields.yml using Mage

andrewkroh:
Make will now delegate to mage for generating fields.yml. Make will check if the mage command exists and go install it if not. The FIELDS_FILE_PATH make variable is not longer used because the path(s) are specified in magefile.go.

This allows fields.yml to be generated on Windows using Mage. The CI scripts for Windows have been updated so that fields.yml is generated for all Beats during testing.

This also adds a make.bat in each directory where building occurs to give Windows
users a starting point.

Some fixes were made to the generators because:
- rsync was excluding important source files contained in a directory
  named "build"
- the generated project needed to be `git init` before running certain
  magefile targets that detect project's root dir and import path.


generator/beat/{beat}/Makefile
```diff
@@ -23,21 +23,18 @@ copy-vendor:
 	rm -rf vendor/github.com/elastic/beats/.git vendor/github.com/elastic/beats/x-pack
 	mkdir -p vendor/github.com/magefile
 	cp -R ${BEAT_GOPATH}/src/github.com/elastic/beats/vendor/github.com/magefile/mage vendor/github.com/magefile
+	ls -la vendor/github.com/magefile
```
ruflin: Was this for testing?

dev-tools/jenkins_ci.ps1
```diff
@@ -35,20 +43,11 @@ New-Item -ItemType directory -Path build\coverage | Out-Null
...
-# always build the libbeat fields
-cp ..\libbeat\_meta\fields.common.yml ..\libbeat\_meta\fields.generated.yml
-cat ..\libbeat\processors\*\_meta\fields.yml | Out-File -append -encoding UTF8 -filepath ..\libbeat\_meta\fields.generated.yml
-cp ..\libbeat\_meta\fields.generated.yml ..\libbeat\fields.yml
-
-if ($env:beat -eq "metricbeat") {
-    cp .\_meta\fields.common.yml .\_meta\fields.generated.yml
-    python .\scripts\fields_collector.py | out-file -append -encoding UTF8 -filepath .\_meta\fields.generated.yml
-}
+exec { mage build } "Build FAILURE"
```
ruflin: Nice. This should have been simplified before already when we move to fields collection to golang.

generator/common/Makefile
```diff
@@ -11,23 +11,36 @@ PREPARE_COMMAND?=
 test: prepare-test
 	. ${PYTHON_ENV}/bin/activate; \
 	export GOPATH=${PWD}/build ; \
-	export PATH=${PATH}:${PWD}/build/bin; \
+	export PATH=$${GOPATH}/bin:${PATH}; \
```
ruflin: `$$` on purpose?

generator/metricbeat/Makefile
```diff
@@ -1,16 +1,14 @@
-BEAT_TYPE=metricbeat
-PREPARE_COMMAND=MODULE=elastic METRICSET=test make create-metricset ; FIELDS_FILE_PATH=module
-override FIELDS_FILE_PATH=
-export FIELDS_FILE_PATH
-
+BEAT_TYPE       = metricbeat
```
ruflin: I like the new indentation but thing I wonder is if the ` ` after the `=` could cause some issues? I remember we had a postfix space when it was something like  `BEAT_TYPE=metricbeat #is a comment`, not sure if it also applies to prefix.

libbeat/scripts/Makefile
```diff
@@ -304,9 +306,8 @@ coverage-report:
 
 
 .PHONY: fields
-fields:
-	@go run ${ES_BEATS}/libbeat/scripts/cmd/global_fields/main.go --es_beats_path $(ES_BEATS) --beat_path $(PWD) $(FIELDS_FILE_PATH)
-
+fields: mage
+	mage -v fields
```
ruflin: Replacing the Makefile step by step 👍

generator/beat/{beat}/Makefile
andrewkroh: Yeah, I did a bunch of debugging to solve the case of the missing github.com/magefile/mage/build directory. It was the rsync exclude.

I'll fix this.

generator/common/Makefile
andrewkroh: Yes. It's an escape so that GOPATH is resolve in the shell command rather than by Make. If you don't do this then you get the original GOPATH value and not the one set 2 lines above.

generator/metricbeat/Makefile
andrewkroh: I'm pretty sure the spaces are dropped by make (in bash they would not be). I've used this syntax before without issues (https://github.com/elastic/golang-crossbuild/blob/master/Makefile.common and several other files in this repo).

dev-tools/jenkins_ci.ps1
```diff
@@ -21,6 +21,14 @@ $env:PATH = "$env:GOPATH\bin;C:\tools\mingw64\bin;$env:PATH"
 # each run starts from a clean slate.
 $env:MAGEFILE_CACHE = "$env:WORKSPACE\.magefile"
 
+exec { go install github.com/elastic/beats/vendor/github.com/magefile/mage }
+exec { go get -u github.com/jstemmer/go-junit-report }
+
+echo "Building libbeat fields.yml"
```
andrewkroh: This command will go away once the field generation is updated to automatically get the libbeat field values.

dev-tools/jenkins_ci.ps1
andrewkroh: Does this also mean that .\scripts\fields_collector.py should have been removed?

ruflin: I would say yes. I remember something like we wanted to keep it around in the PR that introduced the golang for comparing the two and removing it afterwards but seems that was forgotten.

PR #7672, opened 2018-07-22
fix log suffix can only be .log in kubernetes

yu-yang2:
Since the kubernetes emptyDir log can only have suffix with `.log`, then we can not collect other log formats in the filebeat `type` log. So this piece of code should be removed.


exekias:
jenkins, test this please

yu-yang2:
@exekias changelog has been modified.

exekias:
It seems we lost track of this PR, I'm closing it as I don't see a clear need for the change. Please reopen if you think this still makes sense :pray:

PR #7674, opened 2018-07-23
6.3

whitezevi:


ruflin:
Closing this PR as it seems this is a PR backporting master to 6.3 which should not happen.

@dilchenko Could we move the discussion around the monitoring issue to a new Github issue instead or a PR with only the commit needed?

dilchenko:
I accidentally committed the doc change to the branch directly instead of creating a PR. Believe @dedemorton has fixed that, as well as the documentation text itself, so not much to discuss from my end.

dedemorton:
Right, the docs containing changes by @dilchenko are fixed now and should be in all relevant branches, so there's no need to back or forward port anything related to those docs at this point.

PR #7675, opened 2018-07-23
Docs: Add deprecation check for dashboard loading.

simitt:
For APM Server the recommended way of loading dashboards and Kibana index pattern will be through the Kibana UI from 6.4 on. Since the docs are based on the libbeat docs we need to add a deprecation flag for dashboard and index pattern related documentation.

relates to https://github.com/elastic/apm-server/pull/1142

### DIFF
--- a/libbeat/docs/command-reference.asciidoc
+++ b/libbeat/docs/command-reference.asciidoc
@@ -20,12 +20,22 @@
 :modules-command-short-desc: Manages configured modules
 :run-command-short-desc: Runs {beatname_uc}. This command is used by default if you start {beatname_uc} without specifying a command
 
+ifndef::deprecate_dashboard_loading[]
+
 ifdef::has_ml_jobs[]
 :setup-command-short-desc: Sets up the initial environment, including the index template, Kibana dashboards (when available), and machine learning jobs (when available)
 endif::[]
 
 ifndef::has_ml_jobs[]
-:setup-command-short-desc: Sets up the initial environment, including the index template, Kibana dashboards (when available)
+:setup-command-short-desc: Sets up the initial environment, including the index template and Kibana dashboards (when available)
+endif::[]
+
+endif::[]
+
+ifdef::deprecate_dashboard_loading[]
+
+:setup-command-short-desc: Sets up the initial environment, including the ES index template and Kibana dashboards (deprecated).
+
 endif::[]
 
 :test-command-short-desc: Tests the configuration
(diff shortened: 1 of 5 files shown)


libbeat/docs/command-reference.asciidoc
```diff
@@ -20,12 +20,23 @@
...
 ifdef::has_ml_jobs[]
 :setup-command-short-desc: Sets up the initial environment, including the index template, Kibana dashboards (when available), and machine learning jobs (when available)
 endif::[]
 
 ifndef::has_ml_jobs[]
-:setup-command-short-desc: Sets up the initial environment, including the index template, Kibana dashboards (when available)
+:setup-command-short-desc: Sets up the initial environment, including the index template and Kibana dashboards (when available)
+endif::[]
+
+endif::[]
```
ruflin: Could this be combined to a larger if statement? Also below you frequently have multiple if statements per paragraph which could be combined to a larger one. Is that on purpose?

libbeat/docs/dashboardsconfig.asciidoc
```diff
@@ -10,7 +10,7 @@
 //////////////////////////////////////////////////////////////////////////
 
 [[configuration-dashboards]]
-== Load the Kibana dashboards
+== Load the Kibana dashboards (deprecated)
```
ruflin: This would mean on our end it shows up as deprecated too? Not sure if we should put this into the title?

Instead you could use `[]deprecated` flag with an if statement around?

libbeat/docs/shared-kibana-config.asciidoc
```diff
@@ -10,7 +10,7 @@
 //////////////////////////////////////////////////////////////////////////
 
 [[setup-kibana-endpoint]]
-== Set up the Kibana endpoint
+== Set up the Kibana endpoint (deprecated)
```
ruflin: Same as above.

libbeat/docs/command-reference.asciidoc
simitt: The reason is that using larger blocks here would mean more duplication, as in between the blocks we have parts valid for both cases, including even headlines. 

If you prefer larger code blocks with more duplication I'll change it.

libbeat/docs/dashboardsconfig.asciidoc
simitt: Sure, I hadn't pushed the latest changes.

libbeat/docs/command-reference.asciidoc
ruflin: Definitively prefer less duplication. All good.

libbeat/docs/dashboardsconfig.asciidoc
```diff
@@ -10,7 +10,12 @@
 //////////////////////////////////////////////////////////////////////////
 
 [[configuration-dashboards]]
+ifndef::deprecate_dashboard_loading[]
 == Load the Kibana dashboards
+endif::[]
+ifdef::deprecate_dashboard_loading[]
+== Load the Kibana dashboards (deprecated)
```
ruflin: I still wonder if you really want to have this in the title instead of a deprecate box below as we do normally?

simitt:
@ruflin I updated to use the `deprecation` flag now.

PR #7677, opened 2018-07-23
Report k8s pct metrics from enrichment process

exekias:
Instead of doing it from the `state_container`. Problem with the
previous approach is that `state_container` metricset is not running in all
nodes, but from a single point. Making performance metrics not available
in all cases (in the DaemonSet).
    
With this new approach, the enriching process will also collect
performance metrics, so they should be available everywhere where the
module is run.


metricbeat/vendor/vendor.json
```diff
@@ -2,12 +2,31 @@
 	"comment": "",
 	"ignore": "test github.com/elastic/beats",
 	"package": [
+		{
+			"checksumSHA1": "GYschMPw/Zi5C4QnGKRfgljnx/c=",
+			"path": "github.com/gogo/protobuf/proto",
```
ruflin: @exekias Can you check if some of these we also have in the global vendor directory?  If yes, we should bring the versions in sync.

exekias: It seems we have `github.com/golang/protobuf` under base `vendor`, which differs from this one

ruflin: Should we sync them up?

PR #7691, opened 2018-07-24
Exclude IOWait time from the CPU total

### LINKED ISSUE #7627: [Metricbeat] system.cpu.total.*.pct includes IOWait
The total CPU% is calculated as the amount of time spent in the non-idle state:

https://github.com/elastic/beats/blob/494620259c81c1a724828afa57e2a7fcd2cc18b7/libbeat/metric/system/cpu/cpu.go#L119

I would expect that IOWait time to be excluded also.

tsg:
This affects the following two metrics:

* `system.cpu.total.pct`
* `system.cpu.total.norm.pct`

Fixes #7627.


libbeat/metric/system/cpu/cpu.go
```diff
@@ -116,7 +116,8 @@ func cpuPercentages(s0, s1 *sigar.Cpu, numCPU int) Percentages {
 	}
 
 	calculateTotalPct := func() float64 {
-		return common.Round(float64(numCPU)-calculatePct(s0.Idle, s1.Idle), common.DefaultDecimalPlacesCount)
+		idle := calculatePct(s0.Idle, s1.Idle) + calculatePct(s0.Wait, s1.Wait)
```
ruflin: Could you add a note to this line to explain why we calculate it like this? Can be later.

tsg: Left a comment.

andrewkroh:
LGTM

PR #7696, opened 2018-07-24
Add document for beat export dashboard

ruflin:
Follow up from https://github.com/elastic/beats/pull/7239

### DIFF
--- a/libbeat/docs/command-reference.asciidoc
+++ b/libbeat/docs/command-reference.asciidoc
@@ -14,7 +14,7 @@
 
 :global-flags: Also see <<global-flags,Global flags>>.
 
-:export-command-short-desc: Exports the configuration or index template to stdout
+:export-command-short-desc: Exports the configuration,index template or a dashboard to stdout
 :help-command-short-desc: Shows help for any command
 :keystore-command-short-desc: Manages the <<keystore,secrets keystore>>
 :modules-command-short-desc: Manages configured modules
@@ -77,7 +77,7 @@ endif::[]
 
 [options="header"]
 |=======================
-|Commands | 
+|Commands |
 |<<export-command,`export`>> |{export-command-short-desc}.
 |<<help-command,`help`>> |{help-command-short-desc}.
 |<<keystore-command,`keystore`>> |{keystore-command-short-desc}.
@@ -96,8 +96,8 @@ Also see <<global-flags,Global flags>>.
 ==== `export` command
 
 {export-command-short-desc}. You can use this
-command to quickly view your configuration or the contents of the index
-template.
+command to quickly view your configuration, see the contents of the index
+template or export a dashboard from Kibana.
 
 *SYNOPSIS*
 
@@ -112,6 +112,30 @@ template.
 *`config`*::
 Exports the current configuration to stdout. If you use the `-c` flag, this
 command exports the configuration that's defined in the specified file.
+
+
+
+*`dashboard`*::
+Exporting a dashboard allows to to store a dashboard on disk in a
+module and load it automatically. The following command can be used:
+
+[source,shell]
+----
+$ {beatname_lc} export dashboard --id="dashboard-id" > dashboard.json
+----
+
+The `dashboard-id` can be found in the Kibana URL. By default `export dashboard`
+will write the dashboard to stdout. Above it's written into `dashboard.json` so
+it can later imported again. The file contains the dashboard with all
+visualizations and searches. The index pattern is removed as it is
+expected to be loaded separately for a Beat.
+
+The generated `dashboard.json` file can be copied into the `kibana/6/dashboard`
+directory of {beatname_lc} and next time `{beatname_lc} setup dashboards` is
+run the dashboard will be imported.
+
+In case Kibana is not running on `localhost:5061` the {beatname_lc}
+configuration under `setup.kibana` must be adjusted.
 
 [[template-subcommand]]
 *`template`*::
@@ -176,7 +200,7 @@ Specifies the name of the command to show help for.
 [[keystore-command]]
 ==== `keystore` command
 
-{keystore-command-short-desc}. 
+{keystore-command-short-desc}.
 
 *SYNOPSIS*
 
@@ -690,4 +714,3 @@ the _Beats Platform Reference_ for more information.
 
 *`-v, --v`*::
 Logs INFO-level messages.
-
--- a/libbeat/docs/command-reference.asciidoc
+++ b/libbeat/docs/command-reference.asciidoc
@@ -14,7 +14,7 @@
 
 :global-flags: Also see <<global-flags,Global flags>>.
 
-:export-command-short-desc: Exports the configuration,index template or a dashboard to stdout
+:export-command-short-desc: Exports the configuration, index template or a dashboard to stdout
 :help-command-short-desc: Shows help for any command
 :keystore-command-short-desc: Manages the <<keystore,secrets keystore>>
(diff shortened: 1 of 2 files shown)


ruflin:
@dedemorton TBH was not 100% sure where to put it. This is basically a middle ground that allows people to export dashboards also without having to check out the Beats source code and build it. We don't really have this yet in the dev guide.

Alternative would be to put it in all guides where we list other commands?

dedemorton:
@ruflin I'll take a closer look at this PR tomorrow. I do think we should add this to the command reference, too. I can take that on in a separate PR. I keep thinking I should bump the command reference up one level in the TOC. It won't change any of the URLs, but might make it more visible. WDYT?

ruflin:
After your comment I realised this should definitively go into the command reference and not the dev guide. I moved it.

The reason is that the export methods we use in the dev guide produce a different format of dashboards which is used for contributing. The format created by export dashboard is to be used directly if the loading mechanism in Beat.

As commands become more important +1 on moving them up. I had a hard time to find it under `Setting up ...`.

libbeat/docs/command-reference.asciidoc
```diff
@@ -113,6 +113,30 @@ template.
 Exports the current configuration to stdout. If you use the `-c` flag, this
 command exports the configuration that's defined in the specified file.
 
+
+
+*`dashboard`*::
+Exporting a dashboard allows to to store a dashboard on disk in a
```
dedemorton: typo "to to"

libbeat/docs/command-reference.asciidoc
```diff
@@ -113,6 +113,30 @@ template.
...
+[source,shell]
+----
+$ {beatname_lc} export dashboard --id="dashboard-id" > dashboard.json
+----
+
+The `dashboard-id` can be found in the Kibana URL. By default `export dashboard`
+will write the dashboard to stdout. Above it's written into `dashboard.json` so
+it can later imported again. The file contains the dashboard with all
+visualizations and searches. The index pattern is removed as it is
+expected to be loaded separately for a Beat.
```
dedemorton: Change "a Beat" to {beatname_uc} (since this doc appears in each book).

ruflin:
@dedemorton Comments addressed. Looking forward to a bigger refactoring here.

libbeat/docs/command-reference.asciidoc
```diff
@@ -14,7 +14,7 @@
 
 :global-flags: Also see <<global-flags,Global flags>>.
 
-:export-command-short-desc: Exports the configuration or index template to stdout
+:export-command-short-desc: Exports the configuration,index template or a dashboard to stdout
```
dedemorton: Sorry, missed this during my first review. You need to add a space after `configuration,`

libbeat/docs/command-reference.asciidoc
```diff
@@ -113,6 +113,30 @@ template.
 Exports the current configuration to stdout. If you use the `-c` flag, this
 command exports the configuration that's defined in the specified file.
 
+
+
+*`dashboard`*::
+Exporting a dashboard allows to store a dashboard on disk in a
+module and load it automatically. The following command can be used:
+
+[source,shell]
```
dedemorton: Since the example contains an asciidoc reference, you need to replace line 122 with:
```
["source","yaml",subs="attributes"]
```

libbeat/docs/command-reference.asciidoc
```diff
@@ -113,6 +113,30 @@ template.
 Exports the current configuration to stdout. If you use the `-c` flag, this
 command exports the configuration that's defined in the specified file.
 
+
+
+*`dashboard`*::
+Exporting a dashboard allows to store a dashboard on disk in a
+module and load it automatically. The following command can be used:
+
```
dedemorton: Add a plus `+` here so the example gets indented correctly.

libbeat/docs/command-reference.asciidoc
```diff
@@ -113,6 +113,30 @@ template.
...
+[source,shell]
+----
+$ {beatname_lc} export dashboard --id="dashboard-id" > dashboard.json
+----
+
```
dedemorton: Add a plus `+` here so the text below gets indented correctly.

libbeat/docs/command-reference.asciidoc
```diff
@@ -113,6 +113,30 @@ template.
...
+The `dashboard-id` can be found in the Kibana URL. By default `export dashboard`
+will write the dashboard to stdout. Above it's written into `dashboard.json` so
+it can later imported again. The file contains the dashboard with all
+visualizations and searches. The index pattern is removed as it is
+expected to be loaded separately for {beatname_uc}.
+
```
dedemorton: Add a plus `+` here so the text below gets indented correctly.

libbeat/docs/command-reference.asciidoc
```diff
@@ -113,6 +113,30 @@ template.
...
+The generated `dashboard.json` file can be copied into the `kibana/6/dashboard`
+directory of {beatname_lc} and next time `{beatname_lc} setup dashboards` is
+run the dashboard will be imported.
+
```
dedemorton: Add a plus `+` here so the text below gets indented correctly.

libbeat/docs/command-reference.asciidoc
ruflin: done

libbeat/docs/command-reference.asciidoc
```diff
@@ -113,6 +113,30 @@ template.
...
+----
++
+The `dashboard-id` can be found in the Kibana URL. By default `export dashboard`
+will write the dashboard to stdout. Above it's written into `dashboard.json` so
+it can later imported again. The file contains the dashboard with all
+visualizations and searches. The index pattern is removed as it is
+expected to be loaded separately for {beatname_uc}.
++
+The generated `dashboard.json` file can be copied into the `kibana/6/dashboard`
+directory of {beatname_lc} and next time `{beatname_lc} setup dashboards` is
```
dedemorton: @ruflin  in order for the attribute to resolve
```
`{beatname_lc} setup dashboards`
```
 needs to be
````
+{beatname_lc} setup dashboards+
```
Sorry for missing *another* problem. Sigh.

libbeat/docs/command-reference.asciidoc
```diff
@@ -113,6 +113,30 @@ template.
...
 
+
+
+*`dashboard`*::
+Exporting a dashboard allows to store a dashboard on disk in a
+module and load it automatically. The following command can be used:
++
+["source","shell",subs="attributes"]
+----
+$ {beatname_lc} export dashboard --id="dashboard-id" > dashboard.json
```
dedemorton: We don't show $ in the other examples (makes copy/pasting harder).

libbeat/docs/command-reference.asciidoc
ruflin: Changed. 

Why does the tick not work here and have to use `+` instead?

libbeat/docs/command-reference.asciidoc
ruflin: makes sense. I wonder why I have it here. Pretty sure i copied from somewhere.

libbeat/docs/command-reference.asciidoc
dedemorton: asciidoc treats everything within back ticks as text. Using plus signs forces asciidoc to evaluate the asciidoc references first.

PR #7699, opened 2018-07-24
Update state_node metricset

### LINKED ISSUE #6499: Missing node_status data when using kube-state-metrics > v0.5.0
In version v1.0.0-rc.1 of kube-state-metrics included a number of changes to metric names including the removal of kube_node_status_ready and the addition new metric kube_node_status_condition. 

Full details can of changes can be found here. 
(https://github.com/kubernetes/kube-state-metrics/releases/tag/v1.0.0-rc.1)

Metricbeats needs to be updated to deal with the changes in data from newer versions of kube-state-metrics.

exekias:
`kube_node_status_ready` was deprecated in favor of `kube_node_status_condition`. This PR updates `state_node` metricset to support both old and new fields.

More details in: closes #6499

### DIFF
--- a/metricbeat/helper/prometheus/metric.go
+++ b/metricbeat/helper/prometheus/metric.go
@@ -29,6 +29,9 @@ import (
 
 // MetricMap defines the mapping from Prometheus metric to a Metricbeat field
 type MetricMap interface {
+	// GetOptions returns the list of metric options
+	GetOptions() []MetricOption
+
 	// GetField returns the resulting field name
 	GetField() string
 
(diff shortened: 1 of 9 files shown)


metricbeat/helper/prometheus/metric.go
```diff
@@ -212,3 +242,25 @@ func (m *infoMetric) GetValue(metric *dto.Metric) interface{} {
 func (m *infoMetric) GetField() string {
 	return ""
 }
+
+type opFilter struct {
+	labels map[string]string
+}
+
+func (o opFilter) Process(field string, value interface{}, labels common.MapStr) (string, interface{}, common.MapStr) {
```
ruflin: I wonder why `Hound` is not complaining here? Asking because this could use a comment.

PR #7700, opened 2018-07-24
Make URLHostParserBuilder support a basepath setting

ycombinator:
This will allow modules using URLHostParserBuilder to support a setting named "basepath" in their configuration. If users set this setting to some value, the value will be inserted into the HTTP URI as the first path segment.


ycombinator:
Done and done @andrewkroh. Also squashed commits into single one.

andrewkroh:
@ycombinator I'm thinking that `basepath` deserves a mention in https://www.elastic.co/guide/en/beats/metricbeat/master/configuration-metricbeat.html#module-http-config-options.

ycombinator:
Aha, didn't know about that doc. Thanks @andrewkroh! Added.

PR #7701, opened 2018-07-24
Adding more flexibility to describe a feature

ph:
**Motivation:**

Only adding a stability to describe a feature is often not enough,
instead we now allow to add the following information.

**Fullname:**
This is a more human readable version of the feature.
Example: Jolokia autodiscovery

**Doc:**
This is a one liner describing what you can do with the feature.
Example: The dissect processor allows to extract useful parts of the
original string.

**Stability:**
Describe how stable is the current feature.

From the conversation in https://github.com/elastic/beats/pull/7443#discussion_r199139263 to make the feature concept a bit more future proof.

**Note:** I've decided to use an interface in that case because it will be easier to give an upgrade path for feature developers.

### DIFF
--- a/libbeat/feature/feature.go
+++ b/libbeat/feature/feature.go
@@ -42,20 +42,18 @@ type Featurable interface {
 	// of the method is type checked by the 'FindFactory' of each namespace.
 	Factory() interface{}
 
-	// Stability is the stability of the Feature, this allow the user to filter embedded functionality
-	// by their maturity at runtime.
-	// Example: Beta, Experimental, Stable or Undefined.
-	Stability() Stability
+	// Description return the avaiable information for a specific feature.
+	Description() Describable
 
 	String() string
 }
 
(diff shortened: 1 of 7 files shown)


libbeat/feature/feature.go
```diff
@@ -42,20 +42,18 @@ type Featurable interface {
 	// of the method is type checked by the 'FindFactory' of each namespace.
 	Factory() interface{}
 
-	// Stability is the stability of the Feature, this allow the user to filter embedded functionality
-	// by their maturity at runtime.
-	// Example: Beta, Experimental, Stable or Undefined.
-	Stability() Stability
+	// Description return the avaiable information for a specific feature.
+	Description() Describable
```
ph: I had to make a choice here, I've decided to have a bigger interface for the description of a feature.
I could have created a small interface named 'Filterable' which would only require to have a 'Stability()' method defined.

Issue #7683, opened 2018-07-23
Failing metricbeat test: couchbase


urso:
Beats-CI recently fails on couchbase tests from time to time:

Failing test: `test_couchbase.Test.test_couchbase_0_bucket`

test log output:
```
15:49:22 ======================================================================
15:49:22 FAIL: couchbase metricsets tests [with metricset='bucket']
15:49:22 ----------------------------------------------------------------------
15:49:22 Traceback (most recent call last):
15:49:22   File "/go/src/github.com/elastic/beats/metricbeat/build/python-env/local/lib/python2.7/site-packages/parameterized/parameterized.py", line 392, in standalone_func
15:49:22     return func(*(a + p.args), **p.kwargs)
15:49:22   File "/go/src/github.com/elastic/beats/metricbeat/tests/system/test_couchbase.py", line 22, in test_couchbase
15:49:22     self.check_metricset("couchbase", metricset, self.get_hosts(), self.FIELDS)
15:49:22   File "/go/src/github.com/elastic/beats/metricbeat/module/traefik/../../tests/system/metricbeat.py", line 93, in check_metricset
15:49:22     self.assertItemsEqual(self.de_dot(fields), evt.keys())
15:49:22 AssertionError: Element counts were not equal:
15:49:22 First has 1, Second has 0:  'couchbase'
15:49:22 First has 0, Second has 1:  u'error'
15:49:22 -------------------- >> begin captured stdout << ---------------------
15:49:22 {u'beat': {u'hostname': u'18c49625f4e6', u'name': u'18c49625f4e6', u'version': u'7.0.0-alpha1'}, u'@timestamp': u'2018-07-23T15:42:16.582Z', u'host': {u'name': u'18c49625f4e6'}, u'error': {u'message': u'HTTP error 401 in bucket: 401 Unauthorized'}, u'metricset': {u'rtt': 4648, u'host': u'couchbase:8091', u'name': u'bucket', u'module': u'couchbase'}}
15:49:22 
```

metricbeat log:
```
[...]

PR #7703, opened 2018-07-24
Adding more flexibility to describe a feature

ph:
**Motivation:**

Only adding a stability to describe a feature is often not enough,
instead we now allow to add the following information.

**Fullname:**
This is a more human readable version of the feature.
Example: Jolokia autodiscovery

**Doc:**
This is a one liner describing what you can do with the feature.
Example: The dissect processor allows to extract useful parts of the
original string.

**Stability:**
Describe how stable is the current feature.

From the conversation in https://github.com/elastic/beats/pull/7443#discussion_r199139263 to make the feature concept a bit more future proof.

**Note:** I've decided to use an interface in that case because it will be easier to give an upgrade path for feature developers.

### DIFF
--- a/libbeat/feature/bundle.go
+++ b/libbeat/feature/bundle.go
@@ -38,7 +38,7 @@ func (b *Bundle) Filter(stabilities ...Stability) *Bundle {
 
 	for _, feature := range b.features {
 		for _, stability := range stabilities {
-			if feature.Stability() == stability {
+			if feature.Description().Stability() == stability {
 				filtered = append(filtered, feature)
 				break
 			}
--- a/libbeat/feature/bundle.go
+++ b/libbeat/feature/bundle.go
@@ -17,6 +17,9 @@
 
 package feature
 
+// FilterFunc is the function use to filter elements from a bundle.
+type FilterFunc = func(Featurable) bool
+
 // Bundleable merges featurable and bundle interface together.
 type bundleable interface {
 	Features() []Featurable
@@ -32,19 +35,21 @@ func NewBundle(features ...Featurable) *Bundle {
 	return &Bundle{features: features}
 }
 
-// Filter creates a new bundle with only the feature matching the requested stability.
-func (b *Bundle) Filter(stabilities ...Stability) *Bundle {
+// FilterWith takes a predicate and return a list of filtered bundle matching the predicate.
+func (b *Bundle) FilterWith(pred FilterFunc) *Bundle {
 	var filtered []Featurable
 
 	for _, feature := range b.features {
-		for _, stability := range stabilities {
-			if feature.Description().Stability() == stability {
-				filtered = append(filtered, feature)
-				break
-			}
+		if pred(feature) {
+			filtered = append(filtered, feature)
 		}
 	}
 	return NewBundle(filtered...)
+}
+
+// Filter creates a new bundle with only the feature matching the requested stability.
+func (b *Bundle) Filter(stabilities ...Stability) *Bundle {
+	return b.FilterWith(HasStabilityPred(stabilities...))
 }
 
 // Features returns the interface features slice so
@@ -59,4 +64,16 @@ func MustBundle(bundle ...bundleable) *Bundle {
 		merged = append(merged, feature.Features()...)
 	}
 	return NewBundle(merged...)
+}
+
+// HasStabilityPred returns true if the feature match any of the provided stabilities.
+func HasStabilityPred(stabilities ...Stability) FilterFunc {
+	return func(f Featurable) bool {
+		for _, s := range stabilities {
+			if s == f.Description().Stability() {
+				return true
+			}
+		}
+		return false
+	}
 }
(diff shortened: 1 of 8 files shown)


libbeat/feature/feature.go
```diff
@@ -42,20 +42,18 @@ type Featurable interface {
 	// of the method is type checked by the 'FindFactory' of each namespace.
 	Factory() interface{}
 
-	// Stability is the stability of the Feature, this allow the user to filter embedded functionality
-	// by their maturity at runtime.
-	// Example: Beta, Experimental, Stable or Undefined.
-	Stability() Stability
+	// Description return the avaiable information for a specific feature.
+	Description() Describable
```
ph: I had to make a choice here, I've decided to have a bigger interface for the description of a feature.
I could have created a small interface named 'Filterable' which would only require to have a 'Stability()' method defined.

libbeat/feature/details.go
```diff
@@ -0,0 +1,79 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+package feature
+
+import "fmt"
+
+// Describable contains general information for a specific feature, the fields will be used to report
+// useful information by the factories or any future CLI.
+type Describable interface {
```
urso: nit: it's common in go to end interface names with `er`. E.g. Describer

libbeat/feature/details.go
```diff
@@ -0,0 +1,79 @@
...
+
+	// Doc is a one liner describing the current feature.
+	// Example: Dissect allows to define patterns to extract useful information from a string.
+	Doc() string
+
+	// FullName is the human readable name of the feature.
+	// Example: Jolokia Discovery
+	FullName() string
+
+	String() string
```
urso: What do we need `String` for? Is it supposed to be the short name?

libbeat/publisher/queue/memqueue/broker.go
```diff
@@ -28,7 +28,13 @@ import (
 )
 
 // Feature exposes a memory queue.
-var Feature = queue.Feature("mem", create, feature.Stable)
+var Feature = queue.Feature("mem",
+	create,
+	feature.NewDetails(
+		"Memory queue",
+		"Buffer events in memory before sending to the output.",
+		feature.Stable),
```
urso: I'm a little undecided here if we should rather use a struct instead of an interface+constructor.

Pro:
- Especially with go-plugins I see the advantage of using an interface for the details. One can use custom ones. Even vendoring hickups will be no problem.
- We can provide alternative constructors, e.g. filling in `feature.Stable` or have constructors adding 'class/type' information in the future. E.g. the actual type-safe feature constructor or package can provide support for constructing details via parameters or alternative functions.

Pro and cons:
- Providing a constructor will always force anyone to update feature definitions once the constructor changes. Pro: always up to date meta-data, Cons: developer adding a field must update all features.

Cons:
- if we add more information to the details, it might become more difficult to see which fields are actually set. By using a struct it's always clear if one writes `Details{FullName: ..., Doc: ...}`
- By having `Describable` and interface makes the code a little more bloated. Plus it enforces the use of the constructor, so to work around symbol-name reuse.

libbeat/feature/bundle.go
```diff
@@ -38,7 +38,7 @@ func (b *Bundle) Filter(stabilities ...Stability) *Bundle {
 
 	for _, feature := range b.features {
 		for _, stability := range stabilities {
-			if feature.Stability() == stability {
+			if feature.Description().Stability() == stability {
```
urso: How about updating the function to accept a predicate.

e.g.

```
func (b *Bundle) Filter(pred func(Describer) bool) *Bundle {
  var filtered []Featurable
  for _, feature := range b.features {
    if pred(feature) {
      filtered = append(filtered, feature)
    }
  }
  return NewBundle(filtered...)
}
```

Filtering for stability can be provided using a predicat constructor:

```
func HasStabilityPred(stabilities ...Stability) func(Describer) bool {
  return func(d Describer) bool {
}
```

libbeat/publisher/queue/memqueue/broker.go
ph: I had the same indecision and a similar list.

My main reason I've opted for an interface is, is because data and need will surely change in a minor release and I wanted to make sure we reduce the changes needed by external developers.

libbeat/feature/bundle.go
ph: Im OK with this change, Not sure why this pattern never come to my mind.

libbeat/feature/details.go
ph: I've removed but details will still implement the stringer interface.

libbeat/feature/bundle.go
ph: I like the idea, the following chain will be possible.

```golang
bundle.FilterWith(NamespaceMatchPred("processor"))
   .FilterWith(HasStabilityPred(feature.Beta))
```

ph:
@urso I've updated the PR with the initial review.

Concerning the concrete type vs the interface for details is debatable.
I still think we should keep the interface,  this potentially gives us more flexibility especially in the transition of 6.x vs 7.0.

ph:
Failure is unrelated to this PR changes, Related issue https://github.com/elastic/beats/issues/7683

PR #7710, opened 2018-07-24
Add custom unpack to log hints config to avoid env resolution

vjsamuel:
Raising this as per @exekias. After numerous attempts, using a hints builder in a custom builder was not possible due to template variable being resolved when passing them to the log hints builder.


exekias:
jenkins, test this please

PR #7713, opened 2018-07-24
Simplify fields.yml generator

### LINKED ISSUE #7671: Refactor fields.yml generation
This is a request to have tooling around fields.yml modified to make it easier to use. 

- Do not depend on `libbeat/fields.yml`. Instead, always directly use `libbeat/_meta/fields.common.yml` and `libbeat/processors/**/fields.yml` when building a Beat's fields.yml. This greatly simplifies the build logic and reduces the possibilities of a race condition during concurrent builds. It also stops the build from generated new files inside of the vendor directory for community beats.
- Do not write intermediary files to disk such as `_meta/fields.generated.yml`. Only produce `fields.yml`.

Relates to https://github.com/elastic/beats/pull/7670.

kvch:
This PR simplifies the process of generating a new `fields.yml` file for a Beat.
* `libbeat/fields.yml` is not a dependency anymore
* only `fields.yml` is written to disk

Changes in the asset files are due to different number of newlines in generated `fields.yml` files.

### Interface changes of generator tools
Both `dev-tools/cmd/asset/asset.go` and `libbeat/scripts/cmd/global_fields/main.go` is changed. The first one has two new flags: `-in`, and `-out`. By default it reads from stdin and writes to stdout. Using the flags it's possible to read and write from other files. The second one has one new flag `-out` which specifies the output file. If it's not set, the output is written to stdout.

This makes it possible to generate asset files for Metricbeat without writing `libbeat/fields.yml` to disk. It will be useful in the future when more Beats have modular assets.

Closes #7671


libbeat/Makefile
```diff
@@ -6,4 +6,4 @@ include scripts/Makefile
 
 # Collects all dependencies and then calls update
 .PHONY: collect
-collect: libbeat_fields
+collect: fields
```
andrewkroh: Is there any reason not to give `update` a direct dependency on `fields`?

libbeat/scripts/Makefile
```diff
@@ -309,13 +309,9 @@ coverage-report:
 fields: mage
 	mage -v fields
```
andrewkroh: Can you set this line to `@mage fields`. I was debugging when changed it to its current state and didn't undo my changes.

andrewkroh:
I notice a missing newline from one of the commands. Maybe you can track it down.

```
ERROR: Field <beat.name> is duplicated. Please update and try again.exit status 1
                                                                   ^
```

libbeat/scripts/cmd/global_fields/main.go
```diff
@@ -39,40 +39,33 @@ func main() {
...
-			fmt.Fprintf(os.Stderr, "Cannot collect fields.yml files: %+v\n", err)
-			os.Exit(2)
-		}
-
-		fieldsFiles = append(fieldsFiles, fieldsFile...)
+		fieldsFiles = collectFilesFromDir(*beatPath, fieldsFilePath, fieldsFiles)
 	}
 
-	err = fields.Generate(*esBeatsPath, *beatPath, fieldsFiles)
+	err := fields.Generate(*esBeatsPath, *beatPath, fieldsFiles)
```
andrewkroh: How about internalizing the collection of module files to the `fields.Generate()`. Make it accept a list of module directories like `Generate(elasticBeatsDir, beatDir string, moduleDirs ...string) ([]byte, error)`.

Note that I added a `[]byte` return which is the actual content. Then the caller becomes responsible for writing the content which makes it a bit more reusable. I think the YAML content is small enough to be rendered to memory without any issues.

libbeat/generator/fields/fields.go
```diff
@@ -39,27 +32,54 @@ type YmlFile struct {
...
-	files := []*YmlFile{
-		{
-			Path:   commonFields,
+	var libbeatFieldFiles []*YmlFile
+	var err error
+	if !isLibbeat(beatPath) {
+		commonFields = append(
+			[]string{
+				filepath.Join(esBeatsPath, "libbeat", "_meta", "fields.common.yml"),
+				filepath.Join(beatPath, "_meta", "fields.yml"),
```
andrewkroh: I think _meta/fields.yml is not used anymore.

andrewkroh: I found a place that has some old references to fields files that needs updated to point to plain old fields.yml.

https://github.com/elastic/beats/blob/5443c8ac37ca333342611605690df13a214b128f/libbeat/tests/system/beat/beat.py#L462-L496

andrewkroh:
These lines should be removed now:

https://github.com/elastic/beats/blob/86dc97396f01d0dc49cd5bafa2820f59c53be4f1/dev-tools/jenkins_ci.ps1#L27-L30

metricbeat/Makefile
```diff
@@ -73,7 +73,7 @@ test-module: python-env update metricbeat.test
 	. ${PYTHON_ENV}/bin/activate && INTEGRATION_TESTS=1 nosetests tests/system/test_${MODULE}.py
 
 .PHONY: assets
-assets: libbeat_fields
```
andrewkroh: Unfortunately I don't think Metricbeat is free from this dependency yet due to the way that the fields.go files are composed.

One solution would be to include both metricbeat common fields and libbeat fields in `metricbeat/include/fields.go`. This could be generated by running `global_fields` without passing in the `module` dir as an argument.

It's be nice if we could pipe the output of one command into another like:

`go run ${ES_BEATS}/libbeat/scripts/cmd/global_fields/main.go -es_beats_path ${ES_BEATS} -beat_path ${BEAT_NAME} | go run ${ES_BEATS}/dev-tools/cmd/asset/asset.go -pkg include -in - -out include/fields.go`

ruflin:
@elastic/apm-server FYI, this change is happening in master.

libbeat/scripts/cmd/global_fields/main.go
```diff
@@ -39,18 +40,30 @@ func main() {
...
 
-	if len(beatFieldsPaths) == 0 {
-		fmt.Println("No field files to collect")
-		err = fields.AppendFromLibbeat(*esBeatsPath, *beatPath)
-		if err != nil {
-			fmt.Fprintf(os.Stderr, "Cannot generate global fields.yml for %s: %+v\n", name, err)
-			os.Exit(2)
+	if len(beatFieldsPaths) == 0 && os.SameFile(esBeatsInfo, beatInfo) {
+		if !*toStdout {
+			fmt.Println("No field files to collect")
```
andrewkroh: You might consider using the stdlib `log` package in the file and the util packages. Just initialize it with `log.SetFlags()` to remove the timestamps. All messages will go to stderr all the time.

And all the places where it logs and exits can use a `log.Fatal*`.

libbeat/scripts/cmd/global_fields/main.go
```diff
@@ -39,18 +40,30 @@ func main() {
...
+		os.Exit(1)
+	}
+	beat, err := os.Open(*beatPath)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error opening target Beat: %+v\n", err)
+		os.Exit(1)
+	}
+	esBeatsInfo, err := esBeats.Stat()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error getting file info of elastic/beat: %+v\n", err)
```
andrewkroh: `elastic/beats` (It's missing the `s`.)

libbeat/scripts/cmd/global_fields/main.go
```diff
@@ -29,6 +29,7 @@ import (
 func main() {
 	esBeatsPath := flag.String("es_beats_path", "..", "Path to elastic/beats")
 	beatPath := flag.String("beat_path", ".", "Path to your Beat")
+	toStdout := flag.Bool("stdout", false, "Write output to stdout")
```
andrewkroh: Personally I'd prefer to standardize on using `-out` to specify the output file in our CLI tools. And make the default be `-` mean output to `os.Stdout` like most unix tools. Then you get the same behavior across all tools. The downside being that you have to write `-out fields.yml` in the common case, but the common case is almost always done via `make` so it's not actually a problem.

libbeat/generator/fields/fields.go
```diff
@@ -39,37 +32,67 @@ type YmlFile struct {
...
+func collectCommonFiles(esBeatsPath, beatPath string, fieldFiles []*YmlFile) ([]*YmlFile, error) {
+	commonFields := []string{
+		filepath.Join(beatPath, "_meta", "fields.common.yml"),
+	}
+
+	var libbeatFieldFiles []*YmlFile
+	var err error
+	if !isLibbeat(beatPath) {
+		commonFields = append(commonFields,
+			filepath.Join(esBeatsPath, "libbeat", "_meta", "fields.common.yml"),
```
andrewkroh: BTW `filepath.Join` always calls `filepath.Clean` on the result. https://golang.org/pkg/path/filepath/#Join

This means you can write `filepath.Join(esBeatsPath, "libbeat/_meta/fields.common.yml")` which is more clear IMO.

libbeat/scripts/cmd/global_fields/main.go
kvch: I would rather keep it as it is right now, because this is the "idiomatic" Beats way when printing to console during scripts or commands. I see that you started to use `log` in mage files, but apart from that `fmt` package is used all over the codebase.

kvch:
Rebased and refactored.

libbeat/generator/fields/module_fields_collector.go
```diff
@@ -71,7 +71,7 @@ func CollectModuleFiles(modulesDir string) ([]*YmlFile, error) {
 func CollectFiles(module string, modulesPath string) ([]*YmlFile, error) {
 
 	var files []*YmlFile
-	fieldsYmlPath := filepath.Join(modulesPath, module, "_meta", "fields.yml")
+	fieldsYmlPath := filepath.Join(modulesPath, module, "_meta/fields.yml")
```
ruflin: What is the reason you introduced the `/` here? Could break on Windows. Any issues with the previous option?

libbeat/generator/fields/module_fields_collector.go
```diff
@@ -91,7 +91,7 @@ func CollectFiles(module string, modulesPath string) ([]*YmlFile, error) {
 		if !s.IsDir() {
 			continue
 		}
-		fieldsYmlPath = filepath.Join(modulesPath, module, s.Name(), "_meta", "fields.yml")
+		fieldsYmlPath = filepath.Join(modulesPath, module, s.Name(), "_meta/fields.yml")
```
ruflin: Same question as above.

libbeat/scripts/cmd/global_fields/main.go
```diff
@@ -27,37 +27,55 @@ import (
...
 		os.Exit(1)
 	}
 
-	if len(beatFieldsPaths) == 0 {
-		fmt.Println("No field files to collect")
-		err = fields.AppendFromLibbeat(*esBeatsPath, *beatPath)
-		if err != nil {
-			fmt.Fprintf(os.Stderr, "Cannot generate global fields.yml for %s: %+v\n", name, err)
-			os.Exit(2)
+	if len(beatFieldsPaths) == 0 && os.SameFile(esBeatsInfo, beatInfo) {
```
ruflin: Trying to follow what this does. Could you quickly elaborate?

metricbeat/Makefile
ruflin: Let's hope this becomes obsolete again in the near future :-)

packetbeat/_meta/fields.yml
```diff
@@ -1,2083 +0,0 @@
-- key: common
```
ruflin: Seems like this was a left over file that should already have been deleted before?

dev-tools/cmd/asset/asset.go
```diff
@@ -31,32 +32,51 @@ import (
 )
 
 var pkg *string
+var input *string
+var output *string
 
 func init() {
 	pkg = flag.String("pkg", "", "Package name")
+	input = flag.String("in", "-", "Source of input. \"-\" means reading from stdin")
```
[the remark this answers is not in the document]
kvch: Done. I forgot to push this commit.

libbeat/generator/fields/module_fields_collector.go
kvch: Andrew pointed out that it is working everywhere in https://github.com/elastic/beats/pull/7713#discussion_r205152643

packetbeat/_meta/fields.yml
kvch: Yes. It is a duplication of `packetbeat/_meta/fields.common.yml` with fields which coming from `protos`. Nothing used this file at all.

libbeat/scripts/cmd/global_fields/main.go
kvch: If a community Beat does not have its own`fields.yml` file, it still requires the fields coming from libbeat to generate e.g assets. In case of Elastic Beats, it's not a problem because all of them has unique fields.yml files somewhere.

ruflin: Thanks for the details. Could you perhaps add a comment to the line to make sure if we refactor this in the future we don't get rid of it?

libbeat/scripts/cmd/global_fields/main.go
ruflin: @andrewkroh Wasn't aware of this feature. We definitively should start using this in the future.

ruflin:
@kvch Thanks, could you follow up with the @elastic/apm-server team?

jalvz:
already aware, thnks ;)

PR #7722, opened 2018-07-25
Reading thrift file globs in Packetbeat

### LINKED ISSUE #7654: [Thrift-rpc] Add support to provide file globs to specify thrift files in packetbeats.yml.
Hi

We are trying to use Packetbeat to monitor our TCP thrift service layer.
 
Can you provide support to give file globs which contains all my thrift files instead of providing individual file names in packetbeat.yml? We have around 100 thrift files. Our thrift file repo follows a directory structure for different types.

yml config can be a combination of thrift file names, directory name or a regex or glob:
idl_files: [ 'my_dir1/*.thrift', 'my_dir2', 'my_dir3/my_dir4' ]

Discuss post for reference:
https://discuss.elastic.co/t/monitor-thrift-transport-using-packetbeat/140844

-
Thanks,
Pallavi

npallavi1988:
Fixes #7654

### DIFF
--- a/packetbeat/protos/thrift/thrift_idl.go
+++ b/packetbeat/protos/thrift/thrift_idl.go
@@ -19,7 +19,8 @@ package thrift
 
 import (
 	"fmt"
-
+	"path/filepath"
+	
 	"github.com/elastic/beats/libbeat/logp"
 
 	"github.com/samuel/go-thrift/parser"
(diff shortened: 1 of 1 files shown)


adriansr:
Thanks for your contribution!

I think it's better to do the glob expansion in `thrift.go`, so `newThriftIdl` is called with the list of expanded files. What do you think?

Also, a couple of couple of requisites before merging:
- Update the docs for the `idl_files` option in `packetbeat/docs/packetbeat-options.asciidoc`.
- Add an entry to `CHANGELOG.asciidoc`

PR #7724, opened 2018-07-25
Make oid, id and port optional in Postgresql activity metricset

ruflin:
It seems these fields do not exist in all setups and environments.


exekias:
Ups, this may need a changelog? If yes, please open a new PR to include it :)

ruflin:
I added it because in 6.3 with Postgres 10.4 it seems in some setups these errors are printed. What I remember in 6.4 this will not be the case anymore thanks to a change made by @jsoriano in the schema package, so it is only on a debug level. For this reason I would leave out the changelog entry.

PR #7728, opened 2018-07-25
Use go run in tests to export dashboards

ruflin:
* This removes the requirement for a makefile command to build the binary before the test is run
* Make sure always the right binary for the platform is executed (if the binary is build inside docker and then run outside on mac, error is returned)
* -dashboard flag renamed to -id to be consistent with binary export command.

### DIFF
--- a/dev-tools/cmd/dashboards/export_dashboards.go
+++ b/dev-tools/cmd/dashboards/export_dashboards.go
@@ -130,7 +130,7 @@ var quiet = false
 
 func main() {
 	kibanaURL := flag.String("kibana", "http://localhost:5601", "Kibana URL")
-	dashboard := flag.String("dashboard", "", "Dashboard ID")
+	id := flag.String("id", "", "Dashboard ID")
 	fileOutput := flag.String("output", "output.json", "Output file")
 	ymlFile := flag.String("yml", "", "Path to the module.yml file containing the dashboards")
 	flag.BoolVar(&indexPattern, "indexPattern", false, "include index-pattern in output")
@@ -144,7 +144,7 @@ func main() {
 
 	client := &http.Client{Transport: transCfg}
 
-	if len(*ymlFile) == 0 && len(*dashboard) == 0 {
+	if len(*ymlFile) == 0 && len(*id) == 0 {
 		fmt.Printf("Please specify a dashboard ID (-dashboard) or a manifest file (-yml)\n\n")
 		flag.Usage()
 		os.Exit(0)
@@ -172,8 +172,8 @@ func main() {
 		os.Exit(0)
 	}
 
-	if len(*dashboard) > 0 {
-		err := Export(client, *kibanaURL, *dashboard, *fileOutput)
+	if len(*id) > 0 {
+		err := Export(client, *kibanaURL, *id, *fileOutput)
 		if err != nil {
 			fmt.Printf("ERROR: fail to export the dashboards: %s\n", err)
 		}
--- a/dev-tools/cmd/dashboards/export_dashboards.go
+++ b/dev-tools/cmd/dashboards/export_dashboards.go
@@ -29,6 +29,8 @@ import (
 	"path"
 	"strings"
 
+	"path/filepath"
+
 	"github.com/elastic/beats/libbeat/common"
 	"github.com/elastic/beats/libbeat/kibana"
 )
@@ -94,7 +96,7 @@ func Export(client *http.Client, conn string, dashboard string, out string) erro
 	data["objects"] = objects
 
 	// Create all missing directories
-	err = os.MkdirAll(path.Dir(out), os.ModePerm)
+	err = os.MkdirAll(path.Dir(out), 0755)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


dev-tools/cmd/dashboards/export_dashboards.go
```diff
@@ -172,8 +172,8 @@ func main() {
 		os.Exit(0)
 	}
 
-	if len(*dashboard) > 0 {
-		err := Export(client, *kibanaURL, *dashboard, *fileOutput)
+	if len(*id) > 0 {
+		err := Export(client, *kibanaURL, *id, *fileOutput)
 		if err != nil {
 			fmt.Printf("ERROR: fail to export the dashboards: %s\n", err)
```
andrewkroh: Can you double check all the exit codes used in this file to ensure that when there is an error that the value is non-zero.

And check that os.Stderr is used to log errors.

libbeat/tests/system/test_dashboard.py
```diff
@@ -71,14 +71,14 @@ def test_export_dashboard(self):
 
         self.test_load_dashboard()
 
-        command = self.beat_path + "/../dev-tools/cmd/dashboards/export_dashboards -kibana http://" + \
+        command = self.beat_path + "/../dev-tools/cmd/dashboards/export_dashboards.go -kibana http://" + \
             self.get_kibana_host() + ":" + self.get_kibana_port()
 
         if os.name == "nt":
-            command = self.beat_path + "\..\dev-tools\cmd\dashboards\export_dashboards -kibana http://" + \
+            command = self.beat_path + "\..\dev-tools\cmd\dashboards\export_dashboards.go -kibana http://" + \
```
andrewkroh: There could be just one command for all OSes if os.path.join and os.path.normpath were used to construct path part.

dev-tools/cmd/dashboards/export_dashboards.go
ruflin: Did a quick "cleanup" I think we have to do a full cleanup to this file at one stage.

Is `os.Stderr` because of `mage`?

andrewkroh: Thanks. It will help integrating it with other tools if it follows the convention of writing errors to [stderr](https://en.wikipedia.org/wiki/Standard_streams#Standard_error_(stderr)).

And yes, if your follow these conventions if will integrate nicely with `mage` which by default passes through the stderr output. Stdout is only passed through when executing in verbose mode (`mage -v`).

dev-tools/cmd/dashboards/export_dashboards.go
```diff
@@ -162,11 +169,13 @@ func main() {
 			directory := path.Join(path.Dir(*ymlFile), "_meta/kibana/6/dashboard")
 			err := os.MkdirAll(directory, 0755)
 			if err != nil {
-				fmt.Printf("ERROR: fail to create directory %s: %v", directory, err)
+				fmt.Fprintf(os.Stderr, "ERROR: fail to create directory %s: %v", directory, err)
```
andrewkroh: This is missing a `\n` which cause some ugliness on the terminal.

dev-tools/cmd/dashboards/export_dashboards.go
```diff
@@ -176,6 +185,7 @@ func main() {
 		err := Export(client, *kibanaURL, *dashboard, *fileOutput)
 		if err != nil {
 			fmt.Printf("ERROR: fail to export the dashboards: %s\n", err)
```
andrewkroh: This is not logging to `stderr`.

dev-tools/cmd/dashboards/export_dashboards.go
```diff
@@ -92,6 +92,13 @@ func Export(client *http.Client, conn string, dashboard string, out string) erro
 	}
 
 	data["objects"] = objects
+
+	// Create all missing directories
+	err = os.MkdirAll(path.Dir(out), os.ModePerm)
```
andrewkroh: `os.ModePerm` is too permissive. Use `0755` for dirs.

andrewkroh: Using `filepath.Dir`. `path` is for URLs IIRC and does not work for Windows.

dev-tools/cmd/dashboards/export_dashboards.go
ruflin: I replaced all calls by a function now that adds the `\n` at the end.

dev-tools/cmd/dashboards/export_dashboards.go
ruflin: changed

dev-tools/cmd/dashboards/export_dashboards.go
```diff
@@ -125,6 +134,11 @@ func ReadManifest(file string) ([]map[string]string, error) {
 	return manifest.Dashboards, nil
 }
 
+func exitWithError(format string, a ...interface{}) {
```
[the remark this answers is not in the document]
ruflin: Have a look below if that is what you expected.

ruflin:
@andrewkroh Have a look again.

PR #7730, opened 2018-07-25
Simplify exporting dashboards for Devs

ruflin:
A new `mage` command is introduced to simplify exporting dashboards and contributing them to Beats. Currently a contributor is required to know about folder hierarchy under which dashboards are stored and are versioned. To simply this a simple `mage` command is provided:

```
MODULE=mysql ID=the-dashboard-id mage ExportDashboard
```

The Dev must provide the module he wants to export the dashboard for and the dashboard id. The script will take care of extracting it from Kibana, creating directories and files. The current script only supports to export dashboards from localhost:5061 as no params are provided to change the host at the moment. Also this change is only for Metricbeat for now. Filebeat should also be added.

Additional changes:

* Create directories
* Pass module and dashboard id
* Add exit code 1 in case of export failure
* Improve error handling

### DIFF
--- a/dev-tools/cmd/dashboards/export_dashboards.go
+++ b/dev-tools/cmd/dashboards/export_dashboards.go
@@ -89,6 +89,7 @@ func Export(client *http.Client, conn string, spaceID string, dashboard string, 
 	for _, obj := range objects {
 		o := obj.(common.MapStr)
 
+		// All fields are optional, so errors are not catched
 		decodeValue(o, "attributes.uiStateJSON")
 		decodeValue(o, "attributes.visState")
 		decodeValue(o, "attributes.optionsJSON")
@@ -104,23 +105,27 @@ func Export(client *http.Client, conn string, spaceID string, dashboard string, 
 		return err
 	}
 
-	err = ioutil.WriteFile(out, []byte(data.StringToPrint()), 0666)
+	err = ioutil.WriteFile(out, []byte(data.StringToPrint()), 0644)
 	if !quiet {
 		log.Printf("The dashboard %s was exported under the %s file\n", dashboard, out)
 	}
 	return err
 }
 
-func decodeValue(data common.MapStr, key string) {
+func decodeValue(data common.MapStr, key string) error {
 	v, err := data.GetValue(key)
 	if err != nil {
-		return
+		return err
 	}
 	s := v.(string)
 	var d interface{}
-	json.Unmarshal([]byte(s), &d)
+	err = json.Unmarshal([]byte(s), &d)
+	if err != nil {
+		return fmt.Errorf("error decoding %s: %v", key, err)
+	}
 
 	data.Put(key, d)
+	return nil
 }
 
 func ReadManifest(file string) ([]map[string]string, error) {
--- a/dev-tools/mage/dashboard.go
+++ b/dev-tools/mage/dashboard.go
@@ -1,0 +1,53 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package mage
+
+import (
+	"fmt"
+	"path/filepath"
+
+	"github.com/magefile/mage/sh"
+)
+
+// ExportDashboard exports a dashboard from Kibana and writes it into the given module.
+func ExportDashboard() error {
+	module := EnvOr("MODULE", "")
+	if module == "" {
+		return fmt.Errorf("MODULE must be specified")
+	}
+
+	id := EnvOr("ID", "")
+	if id == "" {
+		return fmt.Errorf("Dashboad ID must be specified")
+	}
+
+	beatsDir, err := ElasticBeatsDir()
+	if err != nil {
+		return err
+	}
+
+	// TODO: This is currently hardcoded for KB 6, we need to figure out what we do for KB 7
+	file := filepath.Join(beatsDir, BeatName, "module", module, "_meta/kibana/6/dashboard", id+".json")
+
+	dashboardCmd := sh.RunCmd("go", "run",
+		filepath.Join(beatsDir, "dev-tools/cmd/dashboards/export_dashboards.go"),
+		"-output", file, "-dashboard", id,
+	)
+
+	return dashboardCmd()
+}
--- a/dev-tools/mage/dashboard.go
+++ b/dev-tools/mage/dashboard.go
@@ -42,7 +42,7 @@ func ExportDashboard() error {
 	}
 
 	// TODO: This is currently hardcoded for KB 6, we need to figure out what we do for KB 7
-	file := filepath.Join(beatsDir, BeatName, "module", module, "_meta/kibana/6/dashboard", id+".json")
+	file := CWD("module", module, "_meta/kibana/6/dashboard", id+".json")
 
 	dashboardCmd := sh.RunCmd("go", "run",
 		filepath.Join(beatsDir, "dev-tools/cmd/dashboards/export_dashboards.go"),
--- a/docs/devguide/newdashboards.asciidoc
+++ b/docs/devguide/newdashboards.asciidoc
@@ -213,44 +213,24 @@ https://github.com/elastic/beats/tree/master/dev-tools/cmd/dashboards[dev-tools]
 for exporting Kibana 5.x dashboards. See the dev-tools
 https://github.com/elastic/beats/tree/master/dev-tools/README.md[readme] for more info.
 
-/////////////////////
-NOT YET IMPLEMENTED
-NOTE: You can make use of the Makefile from the Beat GitHub repository to export all the Kibana dashboards for a Beat
-from your Elasticsearch. If Elasticsearch is running on localhost, then you just need to run the following command from the Beat repository:
-
-[source,shell]
------------------------------
-make export-dashboards
------------------------------
-
-If Elasticsearch is running on a different host, then you can use the `ES_URL` variable:
-
-[source,shell]
-----------------------------
-ES_URL="http://192.168.3.206:9200" make export-dashboards
-----------------------------
-
-/////////////////////
-
 ==== Exporting Kibana 6.0 dashboards and newer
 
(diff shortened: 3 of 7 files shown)


ruflin:
@andrewkroh My first time creating a `mage` target. Questions:

* Should there also be a `make` command for this or is it ok to only have `mage`?
* Is the way I execute the command the recommended way?

dev-tools/cmd/dashboards/export_dashboards.go
```diff
@@ -92,6 +92,13 @@ func Export(client *http.Client, conn string, dashboard string, out string) erro
...
 
 	data["objects"] = objects
+
+	// Create all missing directories
+	err = os.MkdirAll(path.Dir(out), os.ModePerm)
+	if err != nil {
+		return err
+	}
+
 	err = ioutil.WriteFile(out, []byte(data.StringToPrint()), 0666)
```
andrewkroh: 0666 is too open. Use 0644.

dev-tools/cmd/dashboards/export_dashboards.go
```diff
@@ -92,6 +92,13 @@ func Export(client *http.Client, conn string, dashboard string, out string) erro
 	}
 
 	data["objects"] = objects
+
+	// Create all missing directories
+	err = os.MkdirAll(path.Dir(out), os.ModePerm)
```
andrewkroh: I think you fixed this in the other PR. `s/os.ModePerm/0755/`

metricbeat/magefile.go
```diff
@@ -96,6 +98,27 @@ func Fields() error {
 	return mage.GenerateFieldsYAML("module")
 }
 
+func ExportDashboard() error {
+	module := os.Getenv("MODULE")
```
andrewkroh: Do all of this work from the shared code in dev-tools/mage/dashboard.go so that you don't have to duplicate it in every magefile.go. Hopefully in the future there will be an [import mechanism](https://github.com/magefile/mage/issues/108) in mage to avoid duplication in an even better way.

metricbeat/magefile.go
```diff
@@ -96,6 +98,27 @@ func Fields() error {
...
+	if module == "" {
+		return fmt.Errorf("Dashboad ID must be specified")
+	}
+
+	beatsDir, err := mage.ElasticBeatsDir()
+	if err != nil {
+		return err
+	}
+
+	file := path.Join(beatsDir, "metricbeat", "module", module, "_meta", "kibana", "6", "dashboard", id+".json")
```
andrewkroh: Use `filepath.Join`. You can write `filepath.Join(beatsDir, "metricbeat/module", module, "_meta/kibana/6/dashboard", id+".json")` and it will work fine because it always does a `filepath.Clean`.

metricbeat/magefile.go
```diff
@@ -96,6 +98,27 @@ func Fields() error {
 	return mage.GenerateFieldsYAML("module")
 }
 
+func ExportDashboard() error {
```
andrewkroh: When you are done, document all the env vars that influence the behavior so that a user can run `mage -h exportDashboards` to see this information.

ruflin:
Unfortunately in this case I think there are no defaults. Do we have a way to make certain ENV variables required? Similar to EnvOr but EnvRequired?

Do you add all magefile commands to the Makefile or would it be ok to have new commands only in mage?

andrewkroh:
> Do you add all magefile commands to the Makefile or would it be ok to have new commands only in mage?

No, my goal is just to not break the existing Makefile targets and interface that people depend on. I think we can add mage targets and iterate on them.

> Do we have a way to make certain ENV variables required? Similar to EnvOr but EnvRequired?

Not at the moment. And I'm not sure we should. The underlying function that consumes the params would still need to validate them any way. So I'd opt for using `EnvOr("MY_VAR", "")` then validating the incoming args and returning an informative error message.

ruflin:
@andrewkroh Could you have a look again?

docs/devguide/newdashboards.asciidoc
```diff
@@ -213,44 +213,24 @@ https://github.com/elastic/beats/tree/master/dev-tools/cmd/dashboards[dev-tools]
...
-
-NOTE: The dashboard ID is available in the dashboard URL. For example, in case the dashboard URL is
-`app/kibana#/dashboard/7fea2930-478e-11e7-b1f0-cb29bac6bf8b?_g=()&_a=(description:'Overview%2...`, the dashboard ID is `7fea2930-478e-11e7-b1f0-cb29bac6bf8b`.
+To export a single dashboard for a module you can use the following command inside a Beat with modules:
 
 [source,shell]
 ---------------
-cd filebeat/module/redis/_meta/kibana/default/dashboard
-go run ../../../../../../../dev-tools/cmd/dashboards/export_dashboards.go -id 7fea2930-478e-11e7-b1f0-cb29bac6bf8b -output Filebeat-redis.json
+MODULE=redis ID=AV4REOpp5NkDleZmzKkE mage ExportDashboard
```
andrewkroh: This works, but if you do a `mage -l` it will show `exportDashboard` so I'd use that here.

dev-tools/mage/dashboard.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"fmt"
+	"path/filepath"
+
+	"github.com/magefile/mage/sh"
+)
+
+// ExportDashboard exports a dashboard from Kibana and writes it into the given module.
+func ExportDashboard(beat string) error {
```
andrewkroh: There should already be a BeatName variable defined in settings. So this won't need to be passed in.

dev-tools/mage/dashboard.go
```diff
@@ -0,0 +1,53 @@
...
+
+	beatsDir, err := ElasticBeatsDir()
+	if err != nil {
+		return err
+	}
+
+	file := filepath.Join(beatsDir, beat, "module", module, "_meta/kibana/6/dashboard", id+".json")
+
+	dashboardCmd := sh.RunCmd("go", "run",
+		filepath.Join(beatsDir, "dev-tools", "cmd", "dashboards", "export_dashboards.go"),
```
andrewkroh: Can you join the static paths here to make it more readable, like: `filepath.Join(beatsDir, "dev-tools/cmd/dashboards/export_dashboards.go")`

dev-tools/mage/dashboard.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"fmt"
+	"path/filepath"
+
+	"github.com/magefile/mage/sh"
+)
+
+// ExportDashboard exports a dashboard from Kibana and writes it into the given module.
+func ExportDashboard(beat string) error {
+
```
andrewkroh: Can you please remove this blank line. I know you put this here just for my sake. 😆

dev-tools/mage/dashboard.go
```diff
@@ -0,0 +1,53 @@
...
+	if id == "" {
+		return fmt.Errorf("Dashboad ID must be specified")
+	}
+
+	beatsDir, err := ElasticBeatsDir()
+	if err != nil {
+		return err
+	}
+
+	file := filepath.Join(beatsDir, beat, "module", module, "_meta/kibana/6/dashboard", id+".json")
```
andrewkroh: Is this command tied to K6? I think a comment would be useful here to remind us when this should be changed to 7.

metricbeat/magefile.go
```diff
@@ -116,6 +116,15 @@ func GoTestIntegration(ctx context.Context) error {
 	return mage.GoTest(ctx, mage.DefaultGoTestIntegrationArgs())
 }
 
+// ExportDashboard exports a dashboard and writes it into the correct directory
```
andrewkroh: Auditbeat has module dashboards too.

docs/devguide/newdashboards.asciidoc
ruflin: changed.

dev-tools/mage/dashboard.go
ruflin: I assume you mean that I can just use `BeatName` below?

dev-tools/mage/dashboard.go
ruflin: Adding a comment. Agree it's something we should now address that we start to work against 7.0. Added a TODO on my end.

metricbeat/magefile.go
ruflin: added

dev-tools/mage/dashboard.go
```diff
@@ -0,0 +1,53 @@
...
+		return fmt.Errorf("Dashboad ID must be specified")
+	}
+
+	beatsDir, err := ElasticBeatsDir()
+	if err != nil {
+		return err
+	}
+
+	// TODO: This is currently hardcoded for KB 6, we need to figure out what we do for KB 7
+	file := filepath.Join(beatsDir, BeatName, "module", module, "_meta/kibana/6/dashboard", id+".json")
```
andrewvc: How does this work with dashboards at locations like https://github.com/elastic/beats/tree/master/heartbeat/monitors/active/http/_meta/kibana/6/dashboard ?

Should those move to this new simpler pattern?

ruflin: It doesn't at the moment. We should probably make the path a config param which each beat can set, so it also work here (follow up PR).

ruflin:
Did rebase and quickly run it locally but exited with 1. Need to investigate what changed.

ruflin:
@andrewkroh Should be ready for an other round of review. Tested now locally and works.

ruflin:
@andrewkroh I tried to add it to x-pack filebeat and it the export worked. But it seems as BeatsDir it takes the one from the OSS Filebeat and puts the data there. I wonder if this is a general issue with the magefile?

I suggest for now we skip this and get this in to then iterate on top of it.

dev-tools/mage/dashboard.go
andrewkroh: For the output file you don’t want to use a path that is relative to the `ElasticBeatsDir` because this won’t work for community beats or x-pack beats. Instead use `CWD()` or just leave it as a relative path assuming the tool this is being passed to does the write thing and writes the output based on the CWD that it was executed from.

So in short, use `file := CWD(“module”, module, “_meta/kibana/6/dashboard”, id+”.json”)`.

andrewkroh:
@ruflin I see the problem that would prevent you from using it in x-pack and left a comment on how to fix it. Should be a simple fix.

kvch:
If this is merged my PR needs an rebase: https://github.com/elastic/beats/pull/9097
This is a somewhat othogonal change in functionality, but we touch the same files.

ruflin:
@andrewkroh I tried it with your adjustement in the code. It keeps working in OSS but for x-pack it still creates the data in OSS. I would prefer to rather get this PR in as is and get it improved in a follow up PR, especially as this conflicts with the work from @kvch .

ruflin:
@kvch Merged. You can rebase your PR on top of it now. In case you plan to backport your PR, should we backport this one too to 6.x?

kvch:
I think we can backport it along with the follow-up PR with the fix. But TBH I don't see as much value of this improvement on 6.x than on master. Devs would run it on master or on a branch from master, then backport the dashboards they have exported to 6.x.

ruflin:
@kvch Ok, leave it up to you to decide if you will need this in 6.x or not.

PR #7737, opened 2018-07-25
Make heartbeat test error checks handle platform differences better

andrewvc:
Different platforms commonly report different error messages. The good news is that the key substrings of errors remain the same. This refactor makes error checks substring based instead of exact matches. This should be just as good practically, and more robust to boot.

### DIFF
--- a/heartbeat/hbtest/hbtestutil.go
+++ b/heartbeat/hbtest/hbtestutil.go
@@ -83,10 +83,11 @@ func TCPBaseChecks(port uint16) mapval.Validator {
 
 // ErrorChecks checks the standard heartbeat error hierarchy, which should
 // consist of a message (or a mapval isdef that can match the message) and a type under the error key.
-func ErrorChecks(msg interface{}, errType string) mapval.Validator {
+// The message is checked only as a substring since exact string matches can be fragile due to platform differences.
+func ErrorChecks(msgSubstr string, errType string) mapval.Validator {
 	return mapval.Schema(mapval.Map{
 		"error": mapval.Map{
-			"message": msg,
+			"message": mapval.IsStringContaining(msgSubstr),
 			"type":    errType,
 		},
 	})
--- a/heartbeat/monitors/active/http/http_test.go
+++ b/heartbeat/monitors/active/http/http_test.go
@@ -182,15 +182,13 @@ func TestDownStatuses(t *testing.T) {
 			port, err := hbtest.ServerPort(server)
 			require.NoError(t, err)
 
-			errMsgCheck := mapval.IsStringContaining(fmt.Sprintf("%d", status))
-
 			mapvaltest.Test(
 				t,
 				mapval.Strict(mapval.Compose(
 					hbtest.MonitorChecks("http@"+server.URL, server.URL, "127.0.0.1", "http", "down"),
 					hbtest.RespondingTCPChecks(port),
 					respondingHTTPChecks(server.URL, status),
-					hbtest.ErrorChecks(errMsgCheck, "validate"),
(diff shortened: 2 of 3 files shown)


andrewvc:
Waiting on a green build FYI

andrewvc:
Build is now green!

heartbeat/hbtest/hbtestutil.go
```diff
@@ -83,10 +83,11 @@ func TCPBaseChecks(port uint16) mapval.Validator {
 
 // ErrorChecks checks the standard heartbeat error hierarchy, which should
 // consist of a message (or a mapval isdef that can match the message) and a type under the error key.
-func ErrorChecks(msg interface{}, errType string) mapval.Validator {
+// The message is checked only as a substring since exact string matches can be fragile due to platform differences.
+func ErrorChecks(msgSubstr string, errType string) mapval.Validator {
```
ycombinator: Although not absolutely required, I think its golang convention to start the function header comment with the same name as the function: https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

andrewvc: Ah, actually I *did* do this correctly, but the diff is formatted weirdly on the github UI. the green line is actually the third line of a multi-line comment.

urso: Uff.... do not compare errors by string. It's some very bad practice, prone to break every now and then. Bad practice often ncouraged by `type error interface{ Error() string }`.

Unfortunately there is no common way of how errors are defined, implemented and handled. But errors should be part of the domain of a package. Oftentimes people get lazy by using `errors.New` and `fmt.Errorf`, making it unnecessarily hard and error prone for library users to correctly handle errors.

E.g. in the stdlib `net` package you will find a many error types: net.AddrError, net.DNSConfigError, net.DNSError, net.InvalidAddrError, ...

Error types often add additional context one should use check against. E.g. errors can be temporary or have a different source/root cause. This kind of information is sometimes transported by the type name + some field in the struct (kinda bad practice as well), or a method (+1 for methods). Again, one rarely finds common practices :(
A 'common' workaround is to define predicates extracting common 'behavior'. E.g. the `os` package provides some `IsX` predicates which implement some OS-dependent checks on actual error codes.

Also check out:
- https://dave.cheney.net/2014/12/24/inspecting-errors
- https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully

andrewvc: @urso I read through those blog posts, and they were quite useful to a new go programmer such as myself.

However, there may be a bit of confusion here, and I don't think that advice applies in this situation. We aren't comparing actual errors types at this point, we're comparing event fields in a pseudo-json structure of nested maps. We don't have interfaces at that point, so we can't do the things Dave Cheney suggests.

On a practical level I think the tests as-is are suitable for this specific test. The utility of this test is answering the question "What happens when I try and ping an unreachable host?". The expected answer is quite broad as I discovered, which is "some type of i/o error". It's actually harder to be more specific than that because the output can vary from platform to platform, and even network to network.

It is arguable that we should have lower level tests that would intercept an error value directly and perform that extra validation, but I don't know what extra utility that would have in validating the correct functioning of our code.

urso: Well, as we validate the generated event structure, I guess it can't be helped. It's kinda funny, that we try to test on a higher level, but therefore have to deal with lower-level OS specific error messages.

This makes me wonder if we want to try to 'normalize' at least some error types in heartbeat, so to generate consistent platform-independent error messages.

PR #7745, opened 2018-07-25
Add slowlog metricset to redis

a3dho3yn:
Collects slow queries from Redis


ruflin:
This seems to overlap with https://www.elastic.co/guide/en/beats/filebeat/6.3/filebeat-module-redis.html

a3dho3yn:
I was looking for 'slowlog len' and then I found that it's helpful to have details about slowlog. So I added `slowlog.len` to the `info` metricset (#7695), and decided to add this new metricset.
But now I think this looks more relevant to _logs_ (and filebeaat), and it's better to just add some missing fields to that filebeat module.

PR #7747, opened 2018-07-25
fix --strict.perms option

caub:


ruflin:
@caub Thanks for the fix
@dedemorton Seems like we have this in all versions up to master. Which other versions should we fix it in?

dedemorton:
@ruflin I believe the command works with either one or two dashes, right? If so, it's good to make this change for the sake of consistency, but I don't think it's worth backporting to all the releases.

dedemorton:
@ruflin Actually, this won't backport very easily because I've refactored the content. I can take this over and make the change in the necessary branches after this PR is merged.

ruflin:
@dedemorton You are of course right and both options work. In this case I think we should only adjust it in master as it doesn't break anything.

@caub Did you hit any issues with the option or is this mainly for consistency?

caub:
@ruflin oh you mean `-strict.perms=false` is also valid? I was a bit struggling to run filebeat in docker, with the permissions checks. I ended up doing `filebeat --help` and noticed it shows `--strict.perms` so I thought it was a typo in the docs. Sorry if not

ruflin:
@caub Yes, both work. But we have `--` in the `-h` part.

caub:
then we could close this PR, sorry again

dedemorton:
I think the change is worth making because other users might be confused by inconsistent info. I have a PR in progress right now and will add this change so it gets into master and the 6.x branches.

ruflin:
Ok, lets close this PR as otherwise we have it in the 5.3 release and not the others. But with the changes from @dedemorton we will have it in all 6.x release.

@caub Thanks for bringing this up.

